PR #1639, opened 2026-07-06
fix(providers): guard decodeEntities against out-of-range/surrogate numeric entities

tillmannschatz:
## Summary
- `decodeEntities()` in `providers/successfactors.mjs`, `providers/dassault.mjs`, and `providers/softgarden.mjs` only checked `Number.isFinite(code)` before calling `String.fromCodePoint(code)`, which throws a `RangeError` for a code point outside `0..0x10FFFF` (e.g. `&#99999999;` from a malformed/adversarial job title). Since this ran inside `.replace()` with no try/catch, one bad numeric entity crashed the entire scan for that provider — a denial-of-provider bug.
- Applies the same range + surrogate-half guard already shipped for the newer `radancy`/`hecklerkoch`/`deutschebahn` providers (`code >= 0 && code <= 0x10ffff && !(code >= 0xd800 && code <= 0xdfff)`) so a malformed entity now degrades to the literal text instead of throwing.
- `softgarden.mjs` already had a partial range guard (missing only the surrogate exclusion); it's now aligned with the other two for consistency.

## Test plan
- [x] Added a regression test to each provider's suite (`tests/providers/{successfactors,dassault,softgarden}.test.mjs`) asserting a malformed entity like `&#xD800;` degrades to literal text rather than throwing.
- [x] `node test-all.mjs` — 1293 passed, 0 failed (2 pre-existing, unrelated warnings: missing local user data / go compiler not in env).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTML entity handling across several job providers.
  * Malformed numeric entities now display as literal text instead of causing parsing errors.
  * Added broader regression coverage for named, decimal, and hex entities, including invalid edge cases.

* **Tests**
  * Expanded provider test coverage for malformed entity decoding.
  * Added direct tests for shared entity decoding behavior.

PR #2150, opened 2026-07-24
fix(providers): guard decodeEntities against out-of-range code points in 4 ATS decoders

### LINKED ISSUE #2146: providers: unguarded String.fromCodePoint in 4 ATS decoders throws RangeError on out-of-range entities
**Repro:** A posting field containing an out-of-range numeric HTML entity — e.g. `&#99999999;` or `&#xFFFFFFFF;` — reaches the local `decodeEntities` in `providers/dassault.mjs:55`, `providers/avature.mjs:65`, `providers/rheinmetall.mjs:39`, or `providers/successfactors.mjs:118`. Each does `Number.isFinite(code) ? String.fromCodePoint(code) : m`, and `String.fromCodePoint` throws `RangeError` for a code point above `0x10FFFF`.

**Root cause:** These four providers carry their own local copy of `decodeEntities` guarded only by `Number.isFinite`. The decode runs in a bare parse loop with no per-item `try`, so a single malformed entity throws out of the entire provider parse — and `scan.mjs`'s per-company `catch` then drops **all** of that company's postings for the run. The shared `providers/_html-entities.mjs` already has the correct guard (`code >= 0 && code <= 0x10ffff && !surrogate`), and five providers (`agentic-jobs`, `deutschebahn`, `gem`, `hecklerkoch`, `oraclecloud`) already import it.

**Proposal:** Replace the four local copies with the shared `decodeEntities`, and add per-provider regression tests feeding malformed entities.

Same bug class as the central fix behind #1639, which never landed in these four.

FReptar0:
Fixes #2146.

`dassault`, `avature`, `rheinmetall`, and `successfactors` each carried a local `decodeEntities` whose numeric path was `Number.isFinite(code) ? String.fromCodePoint(code) : m`. `String.fromCodePoint` throws `RangeError` for a code point above `0x10FFFF` (e.g. `&#99999999;`), and because the decode runs in a bare parse loop, one malformed entity throws out of the whole provider parse — `scan.mjs`'s per-company `catch` then drops every posting for that company in the run.

This swaps the four local copies for the already-correct shared `decodeEntities` from `providers/_html-entities.mjs` (which guards `code >= 0 && code <= 0x10ffff` and lone surrogates), matching the five providers that already import it.

- Verified real-entity equivalence across a 23-input battery — no well-formed entity the old code decoded correctly changes.
- Bonus: the shared helper additionally decodes valid uppercase-`X` hex refs (`&#XFC;` → `ü`) that the old copies left literal — a strict improvement, not a regression.
- Per-provider regression tests feed `&#99999999;` / `&#xFFFFFFFF;` / `&#xD800;` and assert the decoder returns without throwing (entity degraded to literal), plus a control that a normal entity still decodes.

`node test-all.mjs --quick`: 2049 passed, 0 failed.

Same bug class as the central fix behind #1639, which never landed in these four.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved job and vacancy parsing when malformed HTML character references are encountered.
  * Prevented crashes caused by invalid, oversized, or unsupported numeric entities.
  * Preserved valid entity decoding while displaying invalid entities as literal text.

* **Tests**
  * Added regression coverage across multiple job-provider parsers for malformed character references.

### DIFF
--- a/providers/avature.mjs
+++ b/providers/avature.mjs
@@ -1,5 +1,6 @@
 // @ts-check
 /** @typedef {import('./_types.js').Provider} Provider */
+import { decodeEntities } from './_html-entities.mjs';
 
 // Avature provider — parses the public Avature career-site job list.
 // Auto-detects from a careers_url on `*.avature.net`; a branded custom domain
@@ -54,18 +55,6 @@ function resolveConfig(entry) {
   // e.g. /en_US/searchjobs/SearchJobs); otherwise default to the classic path.
   const searchPath = /\/SearchJobs\b/i.test(u.pathname) ? u.pathname.replace(/\/+$/, '') : '/careers/SearchJobs';
   return { searchUrl: `${u.origin}${searchPath}`, origin: u.origin };
-}
-
-const NAMED = { amp: '&', lt: '<', gt: '>', quot: '"', apos: "'", nbsp: ' ', '#8226': '•' };
-/** @param {string} s */
-function decodeEntities(s) {
-  return s.replace(/&(#x?[0-9a-fA-F]+|[a-zA-Z0-9]+);/g, (m, body) => {
-    if (body[0] === '#') {
-      const code = body[1] === 'x' || body[1] === 'X' ? parseInt(body.slice(2), 16) : parseInt(body.slice(1), 10);
-      return Number.isFinite(code) ? String.fromCodePoint(code) : m;
-    }
-    return NAMED[body.toLowerCase()] ?? m;
-  });
 }
 
 /** @param {string} s */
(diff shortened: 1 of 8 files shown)


santifer:
Third one today and the most consequential, @FReptar0. `Number.isFinite` looks like a sufficient guard right up until you remember `String.fromCodePoint` also rejects anything above U+10FFFF — so a single malformed entity like `&#99999999;` threw a RangeError out of the whole parse, and `scan.mjs`'s per-company catch then dropped **every** posting from that employer for the run. Silent, total, and attributable to one stray character in one job title.

Two things make this merge-on-sight rather than merge-with-questions: degrading the bad entity to literal text while still decoding `&amp;` is the correct failure mode (a scanner should lose a character, never a company), and you put the regression in all four provider test files instead of just the one you found it in — the four copies of that decoder were the actual root cause, and folding them into one guarded module is what stops the fifth copy from being written next month. Merged 🚀

PR #2135, opened 2026-07-23
fix(reply-matcher): sanitize the domains getAppDomains harvests from notes

rubicon:
## What does this PR do?

`getAppDomains()` collected candidate sender domains for an application by treating any whitespace-separated token containing a period as a domain. Prose fragments (`gaps.`, `directly.`), a filename whose slash was stripped into a single token, a score delta like `3.3/4.5`, the `${company}` guess applied to the confidential-employer marker `?`, and shared ATS or webmail hosts all landed in the set. This validates each candidate: it must look like a hostname, must not be a filename, and must not be a shared ATS, job-board, or webmail domain; the `${company}` guess is skipped for the `?` marker.

## Related issue

Closes #2134

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation / translation
- [ ] Refactor (no behavior change)

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/santifer/career-ops/blob/main/CONTRIBUTING.md)
- [x] If this is a new feature or architecture change, I opened an issue first (bug fixes, providers, docs & translations are exempt — send those straight in)
- [x] My PR does not include personal data (CV, email, real names, scan results, or pipeline data)
- [x] I ran `node test-all.mjs` and all tests pass
- [x] My changes respect the [Data Contract](https://github.com/santifer/career-ops/blob/main/DATA_CONTRACT.md) (no modifications to user-layer files)
- [x] My changes align with the [project roadmap](https://github.com/santifer/career-ops/discussions/156)

## Why it matters

The prose junk is inert: no real sender address equals `gaps.` or `?.com`. The shared-domain case is not. `matchCandidates()` awards `+2` `sender-domain` to any application whose domain set contains the sender's from-domain, so once `linkedin.com`, `greenhouse.io`, or `gmail.com` enters one application's set, every message from that host false-matches that one application at `medium` or `high` confidence. On a score tie the spurious `+2` can also displace a legitimate company-name match into `ambiguous-match`, so a correct result is suppressed rather than merely joined by a wrong one.

The blocklist deliberately applies to domains taken from email addresses too, not only bare note tokens. A recruiter who writes from a Gmail address would otherwise seed `gmail.com` into that application's set and turn every Gmail sender into a `+2` match. The cost of dropping one shared address is small, because company-name and role matching still carry the signal; the cost of keeping one is a confident wrong attribution.

## Approach

[...]

PR #2151, opened 2026-07-24
fix(reply-matcher): decide rejections before offers to stop offer-substring mis-typing

### LINKED ISSUE #2147: reply-matcher: rejections mis-classified as Offers (bare "offer" substring + Offer-before-Rejected order)
**Repro:** `classifyReply({ body_snippet: 'Unfortunately, we are unable to offer you the position.' })` → `{ type: 'Offer', suggestedTrackerUpdate: 'Offer' }`.

**Root cause:** `offerKeywords` ends with the bare substring `'offer'`; `check()` is substring-based; and the Offer decision (`isOffer = signal === 'offer' || hasOfferKeywords`) runs **before** the Rejected decision. So any reply containing the substring "offer" is typed Offer — making the `'unable to offer'` rejection keyword dead code and silently overriding an explicit upstream `signal: 'rejection'`.

**Impact:** `reply-watch.mjs` then pushes a spurious **Offer** tracker-update recommendation for what is actually a rejection — a data-integrity bug that surfaces a rejection as an offer.

**Proposal:** Decide Rejected before Offer (an explicit rejection signal or rejection wording wins); remove the bare `'offer'` keyword while keeping the specific phrases, and add `'pleased to offer'`. Add regression coverage.

FReptar0:
Fixes #2147.

`classifyReply` mis-typed rejections as offers: `offerKeywords` ended with the bare substring `'offer'`, `check()` is substring-based, and the Offer decision ran before the Rejected decision — so any reply containing "offer" (e.g. *"we are unable to offer you the position"*) was typed **Offer** and pushed as an Offer tracker-update via `reply-watch.mjs`.

Two complementary changes:

1. **Rejected is now decided before Offer**, so an explicit `signal: 'rejection'` or rejection wording wins even when an offer phrase survives (e.g. *"will not be sending you an offer letter"*).
2. **Removed the bare `'offer'` keyword** and added `'pleased to offer'` (kept `offer letter`, `job offer`, `congratulations on the offer`, `compensation details`, `employment agreement`) — killing the substring collision at the source.

Regression test proves the rejection cases flip Offer→Rejected while genuine-offer controls (`"we are pleased to offer…"`, `signal: 'offer'`) stay Offer. `node test-all.mjs --quick`: 2044 passed, 0 failed.

Region is disjoint from the in-flight #2135 (which edits `checkRoleMatch` / `getAppDomains`), so no conflict is expected.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reply classification so rejection messages are correctly identified even when they contain offer-related wording.
  - Prevented phrases such as “unable to offer” from being misclassified as offers.
  - Confirmed that legitimate offer messages continue to be classified as offers.

### DIFF
--- a/reply-matcher.mjs
+++ b/reply-matcher.mjs
@@ -251,10 +251,12 @@ export function classifyReply(cand) {
     'job alert', 'invitation to apply', 'recommended jobs', 'newsletter', 'marketing digest', 'job recommendation', 'suggested jobs'
   ];
 
-  // 2. Offer keywords
+  // 2. Offer keywords — specific phrases only. A bare 'offer' substring is deliberately
+  //    excluded: it collides with rejection wording such as 'unable to offer' (see
+  //    rejectionKeywords) and would mis-type rejections as offers.
   const offerKeywords = [
     '录取通知书', '录用信', '录用通知', '录用', '薪资确认', '入职协议', '意向书',
-    'offer letter', 'employment agreement', 'job offer', 'congratulations on the offer', 'compensation details', 'offer'
+    'offer letter', 'employment agreement', 'job offer', 'congratulations on the offer', 'compensation details', 'pleased to offer'
   ];
 
   // 3. Rejected keywords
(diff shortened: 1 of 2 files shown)


santifer:
This is the best kind of bug report: two lines of ordering, and the failure it prevents is one nobody would want to experience. "We are unable to offer you the position" contains `offer`, so it classified as **Offer** and pushed an Offer row into the tracker — a rejection shown to a job seeker as good news. That is not a cosmetic mis-label; that is the tool getting the single most emotionally loaded moment of a search exactly backwards.

Both halves of the fix are right, and the second is the one people usually miss: dropping the bare `offer` keyword down to `pleased to offer` fixes the obvious collision, but reordering so rejection is decided first is what covers the phrasing you cannot enumerate — "we will not be sending an offer letter" still contains `offer letter`, and no keyword list would have saved you there. Merged 🚀

PR #2152, opened 2026-07-24
fix(funnel-velocity): count non-canonical Applied rows in the Waiting section

### LINKED ISSUE #2148: funnel-velocity: Waiting section under-reports follow-up backlog for non-canonical Applied statuses
**Repro:** A tracker row with status `applied` (lowercase), `Applied 2026-06-01` (status carrying a date), `Aplicado` (localized), or `**Applied**` (markdown-bold) is counted in the funnel's `everApplied` but does **not** appear in the "Waiting (in-flight applications)" section.

**Root cause:** `computeWaiting` (`funnel-velocity.mjs:280`) filters with a raw `row.status !== 'Applied'` string compare, whereas the funnel canonicalizes via `normalizeStatus` (which lowercases and resolves aliases). The two counts diverge, so follow-ups on non-canonical Applied rows are silently dropped from the in-flight view.

**Proposal:** Normalize `row.status` before comparing, matching the funnel's canonicalization.

FReptar0:
Fixes #2148.

`computeWaiting` filtered rows with a raw `row.status !== 'Applied'` compare while `everApplied` is computed through `normalizeStatus`, so lowercase `applied`, dated `Applied 2026-06-01`, localized `Aplicado`, and bold `**Applied**` rows were counted in the funnel but invisible in the Waiting / in-flight section — under-reporting the follow-up backlog.

Now normalizes before comparing (`normalizeStatus(row.status) !== 'applied'`), consistent with how `stats.mjs` canonicalizes the funnel. Extends the inline waiting self-test with a 5-variant fixture plus a control asserting `inFlight` agrees with `everApplied`.

`node test-all.mjs --quick`: 2044 passed, 0 failed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved waiting backlog counts by recognizing additional “Applied” status formats, including lowercase, localized, formatted, and date-added variants.
  * Ensured in-flight counts consistently align with calibration results across supported status representations.

* **Tests**
  * Added coverage for multiple non-standard “Applied” status formats.

### DIFF
--- a/funnel-velocity.mjs
+++ b/funnel-velocity.mjs
@@ -41,7 +41,7 @@ import yaml from 'js-yaml';
 import { computeFunnel, computeTrackerStats } from './stats.mjs';
 import { resolveColumns, parseTrackerRow } from './tracker-parse.mjs';
 import { resolveTrackerPath, loadCanonicalStates, resolveCanonicalState } from './tracker-utils.mjs';
-import { parseAppliedDate } from './followup-cadence.mjs';
+import { parseAppliedDate, normalizeStatus } from './followup-cadence.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
 const STATES_FILE = join(CAREER_OPS, 'templates/states.yml');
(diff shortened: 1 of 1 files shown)


santifer:
Sorry for the slow first reply, @FReptar0 — backlog week. Now to the fix, because this one is sharper than its 26 lines suggest.

The bug is a *disagreement between two code paths over the same question*: the funnel decides "did this ever reach Applied?" through `normalizeStatus`, while `computeWaiting` was comparing `row.status` raw. So a tracker with `applied`, `Applied 2026-06-01`, `Aplicado` or `**Applied**` counted in one place and vanished in the other, and the symptom a user sees is the worst kind: a follow-up backlog that silently under-reports, which is exactly when they stop chasing an application that was still alive.

The part I want to call out is the second assertion: checking that `waiting.inFlight` equals `calibration.everApplied` on the same rows. That does not just test your fix, it pins the invariant that made the bug possible in the first place, so the two paths can't drift apart again. Merged 🚀

PR #2153, opened 2026-07-24
fix(match-star): tokenize tag matches so short tokens do not substring-collide

### LINKED ISSUE #2149: match-star: tag scoring uses substring match — short tokens (ai, ml, go) collide inside longer tag words
**Repro:** A story tagged `maintainability` scored against the query token `ai` receives the highest-weight tag bonus (+3), because `match-star.mjs:123` does `tagText.includes(token)` on the space-joined tags. `go` / `algorithm` collides the same way.

**Root cause:** The tag path is the only scorer using substring matching. The title (`titleTokens.includes`), body (`bodyTokens.includes`), and JD-boost (tokenized tag overlap) paths all use exact tokenized membership. Short tokens (`ai`, `ml`, `go`, `qa`, `ux`, `pm`) embed inside longer tag words, inflating irrelevant stories above genuinely relevant ones.

**Proposal:** Build a Set of tokenized tag tokens and award +3 only on exact membership, mirroring the JD-boost path already in the file. Add a regression test.

FReptar0:
Fixes #2149.

The tag-scoring path substring-matched query tokens against the space-joined tag string, so `ai` / `maintainability` and `go` / `algorithm` earned a false +3 — inflating irrelevant STAR stories above genuinely relevant ones. Switched to `new Set(story.tags.flatMap(tokenize))` + `.has(token)` — same weight, same loop, consistent with the title / body / JD-boost paths already in the file.

Regression test proves `ai` vs tag `maintainability` scores 0 (was 3) while exact `leadership` still scores 3. `node test-all.mjs --quick`: 2046 passed, 0 failed.

Note: this intentionally stops partial-word tag matches (e.g. token `learn` no longer matches tag `learning` via the tag path), making the tag path consistent with the already-exact title / body / JD scorers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved STAR-story tag matching to recognize only complete, exact tag words.
  * Prevented partial matches from incorrectly increasing relevance scores.

* **Tests**
  * Added coverage confirming partial-token matches are excluded while exact tag matches continue to receive the correct score.

### DIFF
--- a/match-star.mjs
+++ b/match-star.mjs
@@ -117,10 +117,12 @@ function score(story, queryTokens, jdTokens) {
   const signal = queryTokens.filter(t => !STOPWORDS.has(t));
   let s = 0;
 
-  // Tag match: highest weight (tags are explicit "best for" labels)
-  const tagText = story.tags.join(' ');
+  // Tag match: highest weight (tags are explicit "best for" labels).
+  // Tokenized exact membership (mirrors the JD-boost path below) so short query
+  // tokens (ai, ml, go, qa…) can't spuriously collide inside longer tag words.
+  const tagTokens = new Set(story.tags.flatMap(tokenize));
   for (const token of signal) {
-    if (tagText.includes(token)) s += 3;
+    if (tagTokens.has(token)) s += 3;
   }
 
   // Title/theme match
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -8203,6 +8203,25 @@ try {
     fail(`match-star scorer: tag-exact match score too low (got ${tagExactScores[1]})`);
   }
 
+  // Regression: tag scoring must use tokenized exact membership, not a substring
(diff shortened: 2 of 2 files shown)


santifer:
Second one, and it is the same instinct applied to a different surface: `tagText.includes(token)` was asking "do these characters appear anywhere" when the actual question is "is this a tag". Token `ai` scoring +3 against `maintainability` is the perfect illustration — and because tags carry the highest weight in the scorer, that false positive doesn't just add noise, it pushes an irrelevant story *above* a genuinely relevant one, which is the one thing a STAR matcher must never do.

Mirroring the JD-boost path rather than inventing a new comparison is the right call: two code paths answering the same question the same way is exactly the drift you closed in #2152. And keeping the exact-match-still-scores-3 assertion alongside the regression is what makes the fix safe to take without re-reading the scorer. Merged 🚀

Issue #2154, opened 2026-07-24
Parallel sweeps resolve the same hostname once per connection, flooding the resolver (~37k lookups for one host)


bigguy6883:
## Summary

The parallel sweeps issue one DNS lookup **per connection**, and because they open their connections concurrently, a cold burst resolves the same hostname once per worker. Across a full `scan-ats-full.mjs` run this reached **~37k identical lookups for a single hostname**.

On a host whose resolver rate-limits per client that trips the limit and takes DNS down **for the whole machine**, not just the scan. In my case a Pi-hole (default 1000/min) began refusing queries partway through a sweep, which surfaced as a wave of `ENOTFOUND`/`EAI_AGAIN` provider errors that looked like the boards themselves were failing.

## Reproduce

No DNS tooling needed — this counts the lookups career-ops itself issues, against a local server so nothing depends on the network:

```bash
node --input-type=module -e "
import dns from 'node:dns';
import http from 'node:http';
const srv = http.createServer((q,s)=>{ s.setHeader('content-type','application/json'); s.end('{\"ok\":1}'); });
await new Promise(r=>srv.listen(0,'127.0.0.1',r));
const url = \`http://localhost:\${srv.address().port}/x\`;
let n = 0; const real = dns.lookup;
dns.lookup = (...a) => { n++; return real(...a); };
const { fetchJson } = await import('./providers/_http.mjs');
n=0; for (let i=0;i<30;i++) await fetchJson(url).catch(()=>{});
console.log('30 sequential ->', n, 'lookups');
n=0; await Promise.all(Array.from({length:30},()=>fetchJson(url).catch(()=>{})));
console.log('30 parallel   ->', n, 'lookups');
srv.close();
"
```

On current `main`:

```
30 sequential ->  2 lookups
30 parallel   -> 29 lookups
```

Sequential traffic is already fine — undici's keep-alive reuses the socket. It's the parallel path that multiplies, and that's the path the sweeps take.

## Why it isn't just my box

A local resolver cache (`systemd-resolved`, `dnsmasq`, macOS `mDNSResponder`) absorbs this, and plenty of setups have one — probably why it hasn't surfaced before. But the lookups are genuinely redundant regardless of who absorbs them, and anyone running career-ops behind a rate-limiting or query-logging resolver (Pi-hole, corporate DNS, a VPN resolver) inherits the failure. It also puts avoidable latency in front of every cold connection in a sweep.

## Proposed fix

[...]

PR #2155, opened 2026-07-24
Cache DNS lookups in-process to stop scanner resolver floods

bigguy6883:
# PR: Cache DNS lookups in-process to stop scanner resolver floods

Implements #2154. Independent of #2138 / #2141 — this branch is cut from `main` and touches a different part of `_http.mjs` (the import block, not `fetchWithTimeout`), so it merges cleanly in any order.

## The problem

Node's `fetch()` resolves a hostname once per **connection**, not per request, and keeps no DNS cache of its own. Sequential traffic is therefore already fine — undici's keep-alive reuses the socket. The sweeps, though, open their connections in parallel, and every concurrent connection resolves independently.

Measured against a local server, 30 requests to one hostname on current `main`:

```
30 sequential ->  2 lookups
30 parallel   -> 29 lookups   <-- one per connection
```

Scaled across a full `scan-ats-full.mjs` run that reached **~37k identical lookups for a single hostname**. On a host whose resolver rate-limits per client (a Pi-hole's default is 1000/min) it trips the limit and breaks DNS for the whole machine — which then surfaces as a wave of `ENOTFOUND`/`EAI_AGAIN` errors that look like the boards failing rather than a local problem. The exact repro script is in #2154.

## The fix

`providers/_dns-cache.mjs` memoizes `dns.lookup` with a 5-minute TTL, imported once from `_http.mjs` so every provider and every direct `fetch()` in the process is covered. The 30-parallel case drops from **29 lookups to 1**.

**Coalescing concurrent misses is the load-bearing part, not the TTL.** Since the driver is concurrency rather than request count, a plain TTL cache would still let a cold parallel burst straight through to the resolver — exactly the traffic shape that trips the limiter. The first miss on a key registers; the rest queue behind it and are served the single result.

Other behaviour:
- **Failures are never cached** — a transient resolver blip can't be pinned in for the TTL window.
- **`CAREER_OPS_NO_DNS_CACHE=1`** disables the patch entirely.
- Cache hits still call back on `process.nextTick`, so the async contract `net.connect` relies on is preserved.
- `realLookup`'s own symbols are copied across, or `promisify(dns.lookup)` would silently start dropping the `family` argument.
- Oldest-first eviction at 512 entries.

## Two things you'll want to check first

**1. Why patch `node:dns` instead of configuring the HTTP client.** career-ops depends on no HTTP library — providers call the global `fetch()`. Node exposes no supported way to hand `fetch()` a custom resolver without adding `undici` as a direct dependency to build an `Agent` with a `connect.lookup` option. Patching the `node:dns` module object keeps `package.json` untouched, since `net.connect` reads `dns.lookup` at call time. If you'd rather take the `undici` dependency and do it through a dispatcher, say so and I'll rewrite it that way.

**2. It does not weaken the SSRF egress guard.** `upskill.mjs`'s `validateUrlSecurity()` resolves through `dns/promises` (`dns.resolve` + `dns.promises.lookup`). `dns/promises` has its own independent `lookup` that does **not** route through the patched `dns.lookup`, so that guard still hits the resolver on every check and cannot be poisoned by this cache:

```bash
node -e "const d=require('dns'),p=require('dns/promises');let n=0;
  const r=d.lookup; d.lookup=(...a)=>{n++;return r(...a)};
  p.lookup('example.com').then(()=>console.log('promises hit patch:', n>0))"
# -> promises hit patch: false
```

The patch is also confined to the callback-style `dns.lookup` — nothing else on the module object is touched.

## Tests

`tests/providers/dns-cache.test.mjs`, 6 assertions, driven by a stub resolver with an injected clock — no network, no real DNS, deterministic:

- repeat lookup served from cache
- 20 concurrent misses coalesce into 1 resolver call, all 20 callbacks fire
- failed resolution is not cached, next call retries
- TTL boundary honored (fresh at `ttl-1`, re-resolves at `ttl+1`)
- `family`/`all` variants cached under distinct keys
- import patches `dns.lookup`; `CAREER_OPS_NO_DNS_CACHE=1` opts out

Each was mutation-checked — coalescing disabled, failure-caching inverted, TTL pinned open — and each mutation fails the corresponding assertion. The concurrency test carries a bounded outer timeout (not `unref`'d, cleared on the winning path) so an implementation that drops queued callbacks fails loudly instead of hanging the suite with no output.

`node test-all.mjs --quick`: **2041 passed, 2 failed**. Both failures (`tracker-columns-tests.mjs`, `tracker-writer-lock-tests.mjs`) reproduce identically on a clean `main` worktree at the same commit — they're pre-existing here and unrelated to this change.

## Summary by CodeRabbit

* **Performance**
  * Reduced repeated DNS lookups during concurrent operations through in-process caching.
  * Reuses ongoing lookups for identical requests and automatically expires cached results.
  * Failed lookups are retried instead of being stored.

* **Reliability**
  * Preserves expected callback behavior for cached responses.
  * Supports disabling DNS caching when needed.

### DIFF
--- a/providers/_dns-cache.mjs
+++ b/providers/_dns-cache.mjs
@@ -1,0 +1,133 @@
+// @ts-check
+/**
+ * _dns-cache.mjs — in-process DNS memoization for the scanners.
+ *
+ * Node's `fetch()` resolves the hostname once per *connection* it opens, and
+ * keeps no DNS cache of its own. Sequential requests are cheap — undici's
+ * keep-alive reuses the socket — but the sweeps run their requests in
+ * parallel, and every concurrent connection resolves independently. Measured
+ * against a local server (30 requests to one hostname):
+ *
+ *     sequential           2 lookups
+ *     30 in parallel      29 lookups   <-- one per connection
+ *
+ * Scaled across a full directory sweep that reached ~37k lookups for a single
+ * hostname in one run. On a host whose resolver rate-limits per client (a
+ * Pi-hole's default is 1000/min) that trips the limit and breaks DNS for the
+ * whole machine, not just the scan.
+ *
+ * Because the driver is concurrency rather than request count, coalescing
+ * in-flight misses is the load-bearing part of this file — a plain TTL cache
+ * would still let a cold parallel burst through. With both, the 29 above
+ * becomes 1.
+ *
+ * Why patch `dns.lookup` rather than configure the HTTP client: career-ops
+ * depends on no HTTP library — providers call the global `fetch()`. Node
+ * exposes no supported way to give `fetch()` a custom resolver without
+ * taking on `undici` as a direct dependency to build an `Agent` with a
+ * `connect.lookup` option. Patching the `node:dns` module object keeps the
+ * dependency list untouched: `net.connect` reads `dns.lookup` at call time,
+ * so importing this file once (`_http.mjs` does) covers every provider and
+ * every direct `fetch()` in the process.
+ *
+ * Scope of the patch — deliberately narrow:
+ *   - Only the callback-style `dns.lookup` on the `node:dns` module object.
+ *   - `dns/promises` has its own independent `lookup` and is NOT affected.
+ *     That matters: the SSRF egress guard in `upskill.mjs` resolves through
+ *     `dns/promises` (`dns.resolve` + `dns.promises.lookup`), so its
+ *     validation still hits the resolver every time and cannot be poisoned
+ *     by this cache. Verify with:
+ *       node -e "const d=require('dns'),p=require('dns/promises');let n=0; \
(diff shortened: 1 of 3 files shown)


santifer:
Welcome to career-ops, @bigguy6883, and merged. ~37,000 lookups for one hostname isn't a performance nit: parallel sweeps resolving the same host once per connection is the kind of thing that gets a user rate-limited or, on a corporate network, gets their machine flagged. An in-process cache is the right scope for it — no daemon, no config, nothing to configure wrong. Verified merged into current `main`: 2190 assertions, 0 failures.

Your other two (#2138 and #2141) are in conflict and both touch `AGENTS.md`, which is a file Santiago owns, so those need a rebase and then his eyes rather than mine. 🚀

Issue #1790, opened 2026-07-11
when clicking on see all then i redirect to /explore and i can't see all the jobs


aminlodhiya:
### Code of Conduct

- [x] I agree to follow the [Code of Conduct](https://github.com/santifer/career-ops/blob/main/CODE_OF_CONDUCT.md)

### What happened?

When a user clicks the "See All" button from Today navigation button, they are redirected to the /explore page. However, the Explore page does not display the complete list of jobs. 

<img width="1041" height="746" alt="Image" src="https://github.com/user-attachments/assets/2eec9a43-a807-4acd-87ae-2743c3346974" />

<img width="1064" height="954" alt="Image" src="https://github.com/user-attachments/assets/258bfb7f-8c73-4a34-b8bc-dfddd2831f86" />

### Steps to reproduce

Navigate to a Today section.
Click "See All".
Observe the jobs displayed on the /explore page.

### Version / channel

v1.18.0

### Screen / route

/explore

### OS / Browser

Brave

### Recent errors (from the in-app reporter)

```shell

```

santifer:
Routed to the web track (area:web): the /explore redirect behavior is web-owned. Thanks for reporting.

santifer:
@aminlodhiya coming back to this properly, and with an apology: I routed it to the web track on July 16 and then it sat for three weeks. Routing something is not the same as it being handled, and from your side the two look identical. You also asked about it in Discord and got no reply there either, which makes it worse.

The report is good: screenshot, exact steps, and the distinction between the redirect working and the destination not showing everything. That is more than most bug reports carry, and it was your first issue here.

I'm escalating it to the web owner now rather than leaving it labelled and quiet. You'll get an answer on this thread about whether "See All" should show the full list or a filtered subset (I suspect the redirect is right and the filter on /explore is the bug, but the web side owns that call and I'm not going to guess in public on their behalf).

Thanks for the patience, and sorry it took someone counting unanswered reports for me to notice this one.

santifer:
You were right, and your report held up exactly as written — including the part where you separated "the redirect works" from "the destination doesn't show everything". That distinction is what makes this diagnosable, and it's the reason I can be specific now.

**What's actually happening:** the redirect is fine. `/explore` is the *scan configuration* page and it never loads the "fresh matches" dataset at all — it doesn't consume `/api/whats-new`. So "See all N" promises N roles and lands you on an empty scan form. It isn't a filter showing a subset; the destination simply never had that data.

[...]

PR #2159, opened 2026-07-24
fix(web): route Today's "See all N" link to the fresh-matches results view

rubicon:
## What does this PR do?

Today's dashboard fetches `/api/whats-new` for "Fresh matches this week" and links "See all N" to `/explore` — but `/explore` is the scan configuration page and never loads that data, so clicking through lands on an empty scan form instead of the N roles the link promised. Adds a `loadFresh()` action to the explore provider and a `?view=fresh` handoff so `/explore` renders the same offers through the existing results-phase UI, with the config form still reachable.

## Related issue
None — bug fix, sent straight in per CONTRIBUTING.

## Type of change
- [x] Bug fix

## Checklist
- [x] If this is a new feature or architecture change, I opened an issue first (bug fixes exempt)

## Summary by CodeRabbit

* **New Features**
  * Added a Fresh Matches mode in Explore (`?view=fresh`) that switches into scan mode, applies selected filters, and loads fresh results immediately.
  * Updated the Today dashboard “See all” link for Fresh matches to open the fresh-focused Explore view.
* **Bug Fixes**
  * Improved fresh loading reliability by preventing overlapping requests, clearing prior results before fetching, validating the response, and handling empty and failure states consistently.


santifer:
Linking this to #1790: aminlodhiya reported the same bug on 11 Jul with a screenshot and exact repro steps, before this PR existed. You diagnosed it independently — the `?view=fresh` handoff is the right shape, since the root cause is that `/explore` never consumes the fresh-matches data at all rather than filtering it down. Reviewing now; I'll verify against their original steps as the acceptance check.

santifer:
Reviewed and **verified end-to-end against #1790's original repro steps**, which I'd committed to as the acceptance check.

Getting a real reproduction took some setup worth recording: `/api/whats-new` filters out companies already present in `applications.md`, so on a fixture with a populated tracker it returns 0 offers and the "Fresh matches" section never renders at all. With a tracker-empty root and 8 recently-dated `scan-history.tsv` rows:

- Today renders **`See all 8 →`** pointing at `/explore?view=fresh` ✅
- Clicking it lands on `/explore?view=fresh` showing the **same 8 offers** (n8n, Admiral, Ment, Writer… all present), **not** the empty scan form ✅
- `Refine search` is still there, so your "config form stays reachable" claim holds ✅

**On the implementation:** `loadFresh()` resets the scan-specific state (`sources`, `companiesScanned`, `capHit`, `droppedNoDate`, `partial`) before loading, so scan metadata can't leak into a view that never scanned — that's the detail that makes this a clean addition rather than a shortcut. Concurrency guarded via `runningRef`, three distinct error paths (non-ok, malformed JSON, throw) each with a human message, and `empty-current` reuses the existing empty state. typecheck + build green.

One observation, **not a blocker and not caused by this PR**: reloading a bare `/explore` afterwards still shows these results, because the provider restores from `sessionStorage` (`career-ops:explore-results`). That's the pre-existing session-restore behavior — it did the same for scan results before — so this PR just extends what gets restored. Flagging it only so it isn't mistaken for a regression later.

Nice diagnosis: the root cause is exactly as you wrote it — `/explore` never consumed that dataset, so it was never a filter problem. You found this independently of #1790, which had been sitting unanswered; linking them was on us, not you.

🟢 Approving from the web side.

santifer:
Thanks for this one, @rubicon, and for a detail that made the whole thing land: you diagnosed it on July 24 without knowing the issue existed, which means the report and the fix sat a fortnight apart waiting for someone to introduce them.

Two things I want on the record because they're the difference between a fix and a patch: `loadFresh()` resets the scan-specific state before loading, so metadata from a scan can't leak into a view that never scanned, and the three error paths each say something a human can act on. @career-ops-ui verified it against @aminlodhiya's exact steps, which needed an empty tracker to even reproduce: /api/whats-new filters out companies already in applications.md, so with realistic data the bug is invisible.

Merged. 🚀

PR #1794, opened 2026-07-11
fix(cli): detect OpenCode in default install directory

aminlodhiya:
## What does this PR do?

Adds OpenCode's default installation directory (~/.opencode/bin) to the CLI search paths used during configuration. This allows to correctly detect OpenCode when it is installed using its default installer on Linux.

## Related issue

https://github.com/santifer/career-ops/issues/1793

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation / translation
- [ ] Refactor (no behavior change)

## Checklist

- [x] I have read CONTRIBUTING.md
- [x] If this is a new feature or architecture change, I opened an issue first (bug fixes, providers, docs & translations are exempt — send those straight in)
- [x] My PR does not include personal data (CV, email, real names, scan results, or pipeline data)
- [x] I ran `node test-all.mjs` and all tests pass
- [x] My changes respect the Data Contract (no modifications to user-layer files)

---
Questions? [Join the Discord](https://discord.gg/8pRpHETxa4) for faster feedback.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CLI detection by recognizing executables installed in the OpenCode per-user binary directory.

PR #2160, opened 2026-07-24
fix(web): detect OpenCode default install path

### LINKED ISSUE #1793: Bug: OpenCode CLI is not detected on Linux when installed via the default installer
### Code of Conduct

### What happened?

On Linux, ui does not detect the OpenCode CLI even though it is installed and works correctly.

The default OpenCode installation places the executable at:

~/.opencode/bin/opencode

However, the current CLI detection logic does not search this directory, causing OpenCode to appear as "not installed".

### Steps to reproduce

- Install OpenCode using the default installation method.
- Verify that it is installed (for example, ~/.opencode/bin/opencode exists and runs).
- navigate to /config it still shows opencode is not installed

### Version / channel

web 0.3.0 · alpha e764512

### Screen / route

/config

### CLI tool

OpenCode

### OS / Browser

Brave

### Recent errors (from the in-app reporter)

```shell

```

Masterora:
## Summary

Add the default OpenCode installation directory (`~/.opencode/bin`) to the web UI's CLI search paths.

Fixes #1793

## Validation

- Web tests: 19 passed, 0 failed
- TypeScript typecheck passed
- Next.js production build passed
- Full test suite: 2045 passed, 0 failed
- Verified with OpenCode 1.18.4 installed at `/root/.opencode/bin/opencode` in a Linux Docker container

## Summary by CodeRabbit

* **New Features**
  * CLI detection now also checks the user’s `~/.opencode/bin` directory, improving discovery of installed command-line tools.

### DIFF
--- a/web/src/lib/clis.ts
+++ b/web/src/lib/clis.ts
@@ -28,6 +28,7 @@ function searchDirs(): string[] {
   const home = os.homedir();
   const extra = [
     path.join(home, ".local/bin"),
+    path.join(home, ".opencode", "bin"),
     path.join(home, ".npm-global/bin"),
     path.join(home, ".bun/bin"),
     path.join(home, ".deno/bin"),


santifer:
Sorry for the silence, @Masterora: 19 days is too long for a first reply and there is no good excuse for it.

One line: adds `~/.opencode/bin` to the paths where the web looks for installed CLIs, and closes #1793. What makes it near-mergeable on sight is the verification, not the diff: OpenCode 1.18.4 actually installed at /root/.opencode/bin inside a Linux container, 19 web tests, typecheck, a production build, and the full suite at 2045 passed, 0 failed.

santifer:
Closing this one, and I want to be precise about why, because it isn't a judgment on the patch.

You found a real bug and fixed it correctly. So did **#1794**, filed 13 days before yours, and that one just merged — `main` now has:

```ts
path.join(home, ".opencode/bin"),
```

Your version, `path.join(home, ".opencode", "bin")`, is functionally identical: `path.join` normalises the separator on Windows either way. The tiebreak wasn't correctness, it was two things:

1. **It arrived first** — 34 days ago, and it had already waited far too long.
2. **It matches the file's existing style.** `searchDirs()` spells every other entry as one string with the separator inside: `.local/bin`, `.npm-global/bin`, `.bun/bin`, `.deno/bin`. Merging the segmented form would have left two conventions sitting next to each other in the same array.

You had no way to know #1794 existed and was waiting — it wasn't visible from your side, and duplicate work in a queue this size is a failure of our triage, not of yours. It also means two people independently hit the same missing path, which is the strongest signal that it was worth fixing at all.

If you want a follow-up with real appetite for it: the same `searchDirs()` list is where CLI-detection gaps keep showing up, and the queue currently has separate PRs for Grok and Kimi paths. If you hit another one, file it — I'll route it faster than this.

Sorry for the 21 days of silence.

PR #2161, opened 2026-07-24
fix(scan): resolve PORTAL_HEALTH_PATH from cwd, not script dir

rubicon:
## What does this PR do?

`PORTAL_HEALTH_PATH` resolved its default against the directory `scan.mjs` lives in (`path.dirname(fileURLToPath(import.meta.url))`), while every sibling data path in the same file — `SCAN_HISTORY_PATH`, `PIPELINE_PATH`, `APPLICATIONS_PATH`, `BLACKLIST_PATH`, `SCAN_RUNS_PATH` — is a bare cwd-relative string. This makes it a bare `'data/portal-health.tsv'` like its siblings, and adds a regression test.

## Related issue

None — bug fix (sending straight in per the template).

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] I ran `node test-all.mjs` and all tests pass (2048 passed, 0 failed; the single warning is the pre-existing `cv-sync-check.mjs exited with error (expected without user data)`)

## Why it matters

Any invocation with a sandboxed cwd — a test run, a temp dir, a CI checkout — still wrote portal-health rows into the real `data/portal-health.tsv` of whatever checkout happens to own `scan.mjs`, instead of the cwd's own data dir. That silently pollutes real pipeline data with fixtures.

The regression test spawns a real child process with cwd pinned to a temp dir (not the script's directory), calls `appendPortalHealth()` with no path argument — the exact call the production code path makes — and asserts the row lands under that cwd while the script's own directory is left provably untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Portal health persistence now defaults to `data/portal-health.tsv` relative to the current working directory, preventing writes to an unexpected location.

* **Tests**
  * Added an integration test that runs the process from a temporary sandbox to verify default path resolution and confirm existing portal health data remains unchanged.
  * Added coverage for safe, regression-resistant cleanup behavior when removing test marker rows from the portal health TSV.

### DIFF
--- a/portal-health-lock.mjs
+++ b/portal-health-lock.mjs
@@ -32,6 +32,19 @@ const DEFAULT_STALE_MS = 30_000;
 const DEFAULT_RETRY_MS = 80;
 const DEFAULT_TIMEOUT_MS = 8_000;
 
+// Two directories are ownerless by construction, not by accident: a lock
+// between its mkdir and its owner.json write, and the recover guard, which
+// never carries owner.json at all. Judging those on `age > staleMs` alone lets
+// a caller with an aggressive staleMs delete a directory created microseconds
+// ago — either stealing a winner's lock inside its acquisition window, or
+// evicting a live guard and putting two callers inside the decide-then-delete
+// window the guard exists to serialize.
+//
+// This is a lower bound on patience, never a cap: a larger caller staleMs
+// still wins, and a genuinely abandoned directory still ages out, so a crash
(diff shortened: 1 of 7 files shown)


santifer:
Fourth one today, @rubicon — and the note 'deliberately the same shape as the tracker lock, so there is one lock idiom in the codebase' is the sentence every reviewer wants to find. Between this and the morning trilogy, the whole locking story now speaks one language. Merged. 🚀

PR #2163, opened 2026-07-24
fix(liveness): normalize typographic punctuation and accents before matching

### LINKED ISSUE #2162: Liveness: closure banners with a typographic apostrophe are never matched
Expired postings come back `uncertain` instead of `expired` whenever the portal writes its closure banner with a typographic apostrophe.

**Cause**

`HARD_EXPIRED_PATTERNS` spells the French banner with an ASCII apostrophe:

```js
/offre (expirée|n.est plus disponible)/i
```

Welcome to the Jungle prints `Cette offre n’est plus disponible.` with U+2019, so the pattern never matches. The page then reaches the final branch and returns `no_apply_control` → `uncertain`, which no caller filters, so the dead posting stays in the pipeline.

The same class of miss applies to accents: the page may say `pourvue` / `déjà` where a pattern says `pourvu`.

**Reproduction**

```js
classifyLiveness({
  status: 200,
  bodyText: "Cette offre n’est plus disponible. " + "x".repeat(600),
});
// actual:   { result: "uncertain", code: "no_apply_control" }
// expected: { result: "expired",  code: "expired_body" }
```

Swapping U+2019 for an ASCII apostrophe in that same string returns `expired` — the only difference is the punctuation character.

**Impact**

Measured on a real French pipeline: a sweep of 19 tracked postings returned 17 closed, none of which the classifier had flagged. Applications were prepared against several postings that had already closed.

**Suggested fix**

Normalize once at the entry point of `classifyLiveness` (typographic quotes → ASCII, diacritics stripped, whitespace collapsed) and spell the patterns in that normalized alphabet, rather than adding one pattern per punctuation variant. Happy to send the PR — the change is written and tested.

Thywe20028:
Fixes #2162.

## Problem

`HARD_EXPIRED_PATTERNS` spells the French closure banner with an ASCII apostrophe, but portals print U+2019 (`Cette offre n’est plus disponible.`). The pattern never matches, the page falls through to `no_apply_control` → `uncertain`, and no caller filters `uncertain`, so closed postings stay in the pipeline forever.

## Approach

Adding `[’]` to each pattern would fix today’s banner and miss tomorrow’s. Instead, `classifyLiveness` normalizes its inputs once at the entry point — typographic quotes to ASCII, diacritics stripped, whitespace collapsed — and every pattern is spelled in that normalized alphabet. `expiree` then matches `expirée`, `pourvu` matches `pourvue`.

Also adds the closure formulas French portals actually print: `offre pourvue`, `offre clôturée`, `offre désactivée`, `recrutement terminé`, `candidatures closes`.

## Compatibility

- Branch order and every existing classification are unchanged (404/410, bot challenge, access blocked, expired URL, expired body, redirected off posting, apply control, listing page, insufficient content).
- Apply-control labels go through the same normalization, so the Polish pattern is respelled `wyslij` instead of `wyślij`. English and German banners are unaffected — they carry no diacritics.
- `applyControls` is defensively coerced to an array before mapping.

## Tests

Two regression cases added to `test-all.mjs`, covering the typographic apostrophe and the accented banner. Both **fail on current `main`** and pass with this change; I verified the baseline failure by stashing the patch and re-running.

Verified against the live site: a closed WTTJ posting serves HTTP 200 with the full description and the banner, and its only "postuler" occurrences are FAQ question titles, not apply controls — so the banner is the sole reliable signal, which is exactly the one that was being missed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved liveness classification for text containing accents, typographic apostrophes, inconsistent spacing, or varied quotation marks.
  * Better detects expired job postings, including French closure notices and Polish application controls.

* **Tests**
  * Added coverage for accented French expiration banners and typographic apostrophes in expiration messages.

### DIFF
--- a/liveness-core.mjs
+++ b/liveness-core.mjs
@@ -1,3 +1,19 @@
+// Portals write closure banners with typographic punctuation and accents:
+// WTTJ renders "Cette offre n’est plus disponible." with U+2019, not ASCII "'".
+// A pattern spelled with a plain apostrophe silently never matches, so a clearly
+// expired posting fell through to `no_apply_control` → uncertain → never filtered.
+// Normalize once at the entry point and spell every pattern below in the
+// normalized alphabet: ASCII quotes, no diacritics, collapsed whitespace.
+function normalizeForMatch(text = '') {
+  if (typeof text !== 'string') return '';
+  return text
+    .replace(/[‘’ʼ′´`]/g, "'")
+    .replace(/[“”″]/g, '"')
+    .normalize('NFD')
+    .replace(/[̀-ͯ]/g, '')
+    .replace(/\s+/g, ' ');
+}
+
 const HARD_EXPIRED_PATTERNS = [
   /job (is )?no longer available/i,
   /job.*no longer open/i,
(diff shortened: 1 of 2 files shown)


santifer:
Merged, @Thywe20028. This is a bug that hides inside a character nobody looks at: WTTJ renders "Cette offre n'est plus disponible" with U+2019, the pattern was spelled with an ASCII apostrophe, so it never matched and a plainly-expired posting fell through to `no_apply_control` → uncertain → never filtered out. Silent, and it would have kept being silent.

Normalizing once at the entry point and then declaring that every pattern below is written in the normalized alphabet is the right shape: the alternative is spelling every future pattern twice and getting it wrong eventually. Re-verified against `main` after another liveness fix landed tonight (#2212), so the combined result is what I tested: 2118 assertions, 0 failures. 🚀

PR #2164, opened 2026-07-24
fix(role-matcher): treat 'product' as a baseline token so PM sibling specialties stay distinct

shulmansj:
## What does this PR do?

Adds `'product'` to `BASELINE_TOKENS` in `role-matcher.mjs` so two titles cannot fuzzy-match on `[product, manager]` alone, and adds two regression tests to the shared role-matcher section of `test-all.mjs`.

## The incident

`merge-tracker.mjs` skipped a new **"Product Manager - Marketplace"** tracker row as a false duplicate of an existing **"Product Manager - AI"** row at the same company. Mechanism:

1. The tokenizer drops `ai` (2-letter token, deliberately not in `SHORT_SPECIALTY` per the existing comment: broad AI/ML buckets appear across many unrelated roles).
2. Both titles reduce to `[product, manager]` (Marketplace side keeps `marketplace`, but the overlap is `[product, manager]`).
3. `product` was not a baseline token, so it counted as a discriminating overlap token.
4. Jaccard 2/3 >= 0.6 declared a match, and the new evaluation silently never reached the tracker.

`product` is exactly the kind of role-altitude descriptor `BASELINE_TOKENS` exists for: it identifies the discipline, not the opening, same as `engineer`, `manager`, or `platform`.

## Accepted trade-off

With this change, a repost that retitles "Product Manager - AI" to "PM, AI" (or similar) no longer fuzzy-matches its earlier row, so a retitled repost could produce two tracker rows. That is the safer failure: URL-level dedup in `scan-history.tsv` already covers true reposts at the same URL, and a duplicate row is visible and cheap to fix, while a silently skipped evaluation is not.

## Alternative rejected

Adding `ai` to `SHORT_SPECIALTY` would fix this pair but wrongly dedupe same-company sibling reqs like "Senior Product Manager, AI" vs "Staff Product Manager, AI" — seniority words are stopwords, so both would reduce to `[product, manager, ai]` and match on a real-looking specialty overlap. It would also contradict the existing rationale for excluding broad AI/ML buckets from `SHORT_SPECIALTY`.

## Related issue

None — bug fix, exempt per CONTRIBUTING.md.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] I ran `node test-all.mjs` and all tests pass (2043 passed, 0 failed; 3 pre-existing environment warnings: missing user data, no Go compiler)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved role title matching to keep distinct Product Manager specialties separate, such as Marketplace and AI.
  * Preserved matching for identical Product Manager titles.

* **Tests**
  * Added regression coverage for specialized Product Manager title matching.

### DIFF
--- a/role-matcher.mjs
+++ b/role-matcher.mjs
@@ -54,6 +54,10 @@ export const BASELINE_TOKENS = new Set([
   'analyst', 'designer', 'consultant', 'specialist',
   'platform', 'systems', 'services',
   'backend', 'frontend', 'full', 'stack', 'fullstack',
+  // 'product' alone cannot identify an opening: "Product Manager - Marketplace"
+  // and "Product Manager - AI" must stay separate applications ("ai" is dropped
+  // by the tokenizer, leaving only [product, manager] to match on).
+  'product',
 ]);
 
 /**
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -5126,6 +5126,21 @@ try {
     fail('role matcher ignored a real short-acronym overlap');
   }
 
+  // 'product' is a baseline token: "ai" is dropped by the tokenizer (2-letter,
+  // not in SHORT_SPECIALTY), so without this these titles collapse to
+  // [product, manager] and merge-tracker skips one as a false duplicate.
+  if (!roleFuzzyMatch('Product Manager - Marketplace', 'Product Manager - AI')) {
+    pass('role matcher keeps Product Manager sibling specialties distinct');
+  } else {
+    fail('role matcher collapsed Product Manager - Marketplace into Product Manager - AI');
+  }
+
+  if (roleFuzzyMatch('Product Manager - Marketplace', 'Product Manager - Marketplace')) {
+    pass('role matcher still matches identical Product Manager titles');
+  } else {
+    fail('role matcher rejected an identical Product Manager title');
+  }
+
   // A generic base title (no suffix of its own) shares every one of its tokens
   // with a specialized sibling, so the shared tokens alone used to cross the
   // Jaccard threshold — even though the sibling's extra word is exactly the


FReptar0 (commented):
Nice, precise fix — the incident write-up is clear and the root cause (the tokenizer drops `ai`, leaving `[product, manager]` to match on) is exactly right. Adding `product` to `BASELINE_TOKENS` is consistent with how the set already treats `[software, engineer]` / `[full, stack, engineer]`, and the inline comment explaining why is welcome.

One observation worth a line of test coverage. Because `manager` is already baseline, adding `product` means the overlap for a **bare** `Product Manager` vs `Product Manager` is now entirely baseline, so `roleFuzzyMatch` returns `false` at the `discriminating.length === 0` guard (~L128) — where it returned `true` before. That's actually consistent with the matcher's stated design (the L124-128 comment calls `[software, engineer]`-only overlaps "not the same opening", and `Software Engineer` vs `Software Engineer` already behaves this way), so I read it as **intended, not a regression**.

The practical consequence: two identically-titled bare `Product Manager` rows at the same company are no longer flagged as the same opening by `roleFuzzyMatch`. The two new tests cover `- Marketplace` vs `- AI` and the suffixed-identical case, but not this all-baseline-identical case. A one-liner pinning `roleFuzzyMatch('Product Manager', 'Product Manager')` to its intended result would document the decision and guard against a future "why doesn't an identical PM dedup?" surprise (and is a good prompt to confirm `merge-tracker` still has another dedup signal — exact title/URL — for that case).

Non-blocking — the fix is correct and well-scoped. Thanks for the clear repro.

santifer:
Welcome, @shulmansj, and this is a nineteen-line fix for a bug that eats a real job opening. "Product Manager - Marketplace" and "Product Manager - AI" both tokenize down to [product, manager] because "ai" is two letters and gets dropped, so merge-tracker sees a duplicate and silently skips the second one. The user never learns there was a second role.

You put the reason in a comment next to the entry, which matters here more than usual: `BASELINE_TOKENS` looks like an arbitrary word list until someone explains that each entry is there because it *can't* identify an opening on its own. 🚀

PR #2107, opened 2026-07-22
fix(role-matcher): stop collapsing leveled variants into the bare title (#2009)

abankar1:
Closes #2009 (items 1 and 2; item 3 — the `set-status.mjs` company-only fallback guard — follows as its own PR, per the issue framing it as "Separately").

Thanks @clede for an unusually precise report — the repro and the root-cause analysis distinguishing this from #793 were exactly right.

## The bug

`roleFuzzyMatch`'s seniority gate only fires when **both** titles carry a seniority token:

```js
if (senA.size > 0 && senB.size > 0) { /* require overlap */ }
```

`"Associate Product Manager, TeamName"` vs `"Product Manager, TeamName"` skips it — only one side has a token. And because `roleTokens` drops seniority words as **stopwords**, the two then tokenize *identically*:

```
['product','manager','teamname']   ===   ['product','manager','teamname']
```

Jaccard = 1.0 → dedupe. `merge-tracker.mjs` silently drops the second requisition even though it has its own req ID, URL, and score. As @clede notes, this sits upstream of #793's Jaccard fix — the ratio is computed correctly here; the titles genuinely are identical strings *after* tokenization.

## The fix

A **sub-baseline** qualifier on the lone side is treated as a level disagreement:

```js
export const SUB_BASELINE_SENIORITY = new Set(['associate', 'junior', 'entry', 'intern']);
```

The distinction the issue draws is the one implemented: `associate`/`junior`/`entry`/`intern` mean the req sits *below* the unqualified title (own scope, comp band, req ID), whereas `senior`/`principal` modify an ambiguous baseline and are routinely added or dropped when the same opening is re-posted. So:

| Pair | Before | After |
|---|---|---|
| `Associate Product Manager, Team` vs `Product Manager, Team` | ✅ dedupe | ❌ distinct |
| `Junior Product Manager, Team` vs `Product Manager, Team` | ✅ dedupe | ❌ distinct |
| `Data Engineer` vs `Senior Data Engineer` | ✅ dedupe | ✅ **unchanged** |
| `Associate PM, Team` vs `Associate PM, Team` | ✅ dedupe | ✅ **unchanged** |

The existing `'role matcher still matches when seniority is only stated on one side'` assertion (`Data Engineer` / `Senior Data Engineer`) is deliberately preserved — this narrows only the sub-baseline case, exactly as the issue scopes it.

## Docs (item 2)

[...]

PR #1933, opened 2026-07-15
fix(role-matcher): stopword "member"/"technical" to fix Member of Technical Staff false-positive matches

gregaro:
## What does this PR do?

"Member of Technical Staff" (MTS) is a boilerplate level-prefix title convention used by several companies for senior IC titles — not a content signal. Without stripping it, "member" and "technical" leaked through as apparently-discriminating tokens, making two genuinely different roles (e.g. "...Connector Platform" vs "...Backend Platform") register as a fuzzy-match false positive — risking one being silently collapsed into the other during tracker merge/dedup (the #947/#1616 family of role-matcher over-matching bugs).

## Related issue

Closes #1965

## Type of change

- [x] Bug fix

## How

- Strip the literal `"member of technical staff"` phrase before tokenizing, in `roleTokens()` (`role-matcher.mjs`), rather than a blanket stopword on `member`/`technical`. A blanket stopword would also strip those words from unrelated titles where they're genuinely discriminating (e.g. "Technical Program Manager", "Team Member"), and would break real reposts that happen to contain "technical" — confirmed this concretely: "Senior Technical Recruiter, EMEA" vs "Technical Recruiter, EMEA" drops from match to no-match under a blanket stopword (only "recruiter" is left, one token short of the overlap minimum), but still matches correctly with the phrase-aware approach.
- Four regression tests in `test-all.mjs`: distinct MTS sibling roles stay apart, an exact-title MTS repost still matches, "technical" stays discriminating outside the MTS phrase, and a real repost containing "technical" still matches.

## Testing

- [x] `node test-all.mjs` — 1744 passed, 0 failed
- [x] Confirmed the regression tests fail without the fix and pass with it

## Checklist

- [x] I have read CONTRIBUTING.md
- [x] This is a bug fix (issue-first requirement exempt)
- [x] My PR does not include personal data (CV, email, real names, scan results, or pipeline data)
- [x] I ran `node test-all.mjs` and all tests pass
- [x] My changes respect the Data Contract (no modifications to user-layer files)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

Issue #1881, opened 2026-07-14
role-matcher: generic base role title collapses with its own specialized-suffix sibling


rj1224:
> ⚠️ Sanitized example below — no real CV, tracker, or pipeline data included.

## Code of Conduct

- [x] I agree to follow the Code of Conduct

## Existing issues

- [x] I searched existing issues and this hasn't been reported yet (as this exact instance — see Prior art below for the closely related, already-fixed cases)

### Duplicate-check trail

- `gh search issues "roleFuzzyMatch" --repo santifer/career-ops --include-prs` → found #793, #950, #1458, #1616/#1622, #1298 (all same general failure class, none cover this specific instance)
- `gh search issues "role-matcher" --repo santifer/career-ops --include-prs` → same set
- `gh search issues "dedup sibling roles" --repo santifer/career-ops --include-prs` → no results
- `gh search prs "role-matcher.mjs" --repo santifer/career-ops` → confirmed #1298 (open) and #1622 (merged) are the two live threads touching this file; neither's test suite covers the reproduction below

## Prior art

This is the same general failure class as two already-handled cases, but neither's fix covers this specific shape:

- **#1616 → #1622 (merged):** fixed `roleFuzzyMatch` collapsing titles that differ **only by seniority** ("Senior Private Cloud Engineer" vs "Principal Private Cloud Engineer") by adding a seniority gate. That fix is scoped to `SENIORITY_TOKENS` and doesn't touch the case below, which differs by a non-seniority specialization suffix.
- **#1298 (open):** proposes deterministic dedup on the posting URL instead of fuzzy title matching, which — as a side effect — would also fix the instance below (each opening has its own URL). If #1298 lands first, this issue may become moot for URL-bearing rows; it'd still apply to the fuzzy fallback path for rows without a URL.

## What happened?

`role-matcher.mjs`'s `roleFuzzyMatch()` treats a **generic base role title** and its **own specialized-suffix sibling** at the same company as the same opening, causing `merge-tracker.mjs` to silently skip (or overwrite) one of two genuinely distinct tracker rows.

Unlike the #1616 case (two titles each carrying a *different* discriminating suffix, e.g. "...Foundation" vs "...Guarded Releases" — correctly kept separate because neither suffix is in the overlap), this shape has **one title with no suffix at all**. The un-suffixed title can't contribute a counter-balancing discriminating token, so the shared base noun alone is enough to cross the 0.6 Jaccard threshold.

## Root cause

```js
[...]

PR #2166, opened 2026-07-24
fix(merge-tracker): stop sibling-req clobber when the qualifier is a slashed acronym

### LINKED ISSUE #2165: role-matcher: slashed acronym qualifiers ("(CI/CD)") vanish in tokenization, letting merge-tracker collapse a sibling req and overwrite an Applied row's title/score/report
### Code of Conduct

### Existing issues

- [x] I searched existing issues and this hasn't been reported yet (#1881, #2009, #1616, #1965 cover overlap *logic*; none cover qualifier tokens being dropped entirely)

### What happened?

`roleTokens()` in `role-matcher.mjs` strips punctuation to spaces before the length filter, so a slashed acronym qualifier like `(CI/CD)` becomes `ci cd` — and both halves are then discarded (≤3 chars, not in `SHORT_SPECIALTY`). The one token that distinguishes two sibling reqs at the same company is invisible to the matcher.

Consequence in `merge-tracker.mjs`: the tier-3 duplicate check (company + fuzzy role) declares the two reqs duplicates, and because the incoming evaluation has a real score while the existing row is an Applied backfill (`N/A` → 0), the update path **overwrites the existing row's title, score, date, and report link**. An in-flight application silently disappears from the tracker.

The #1881 subset guard (v1.22.0) catches this pair only when the specialized title happens to carry *additional* surviving tokens. When the slashed acronym is the **only** qualifier, both titles tokenize identically and it still collapses:

```
Senior Software Engineer, Infrastructure            → [software, engineer, infrastructure]
Senior Software Engineer, Infrastructure (CI/CD)    → [software, engineer, infrastructure]   // (CI/CD) vanished
→ identical sets, Jaccard 1.0 → duplicate
```

### Steps to reproduce

On v1.22.0 (`career-ops-v1.22.0`):

```bash
node -e "import('./role-matcher.mjs').then(({roleFuzzyMatch}) => {
  console.log(roleFuzzyMatch(
    'Senior Software Engineer, Infrastructure',
    'Senior Software Engineer, Infrastructure (CI/CD)'
  )); // true — but these are two separately-postable reqs
})"
```

End-to-end: put an Applied row with the bare title in the tracker, drop a TSV for the `(CI/CD)` sibling into `batch/tracker-additions/`, run `node merge-tracker.mjs` → the Applied row is rewritten in place (new title/score/date/report) instead of a second row being added.

### Expected behavior

1. A slashed acronym should survive tokenization as a discriminating token (e.g. collapse `ci/cd` → `cicd` *before* punctuation stripping — also covers `TCP/IP`, `UI/UX`), so the subset guard / Jaccard can see it.
2. Defense-in-depth: only a tier-1 duplicate match (report number + company — provably the same evaluation) should be allowed to rewrite a matched row's **role title**. Tier-2/3 matches should keep the existing title, so a future fuzzy false positive can't destroy the evidence that two reqs were distinct.

I have both fixes running locally with matcher unit tests and an e2e regression (Applied row survives; a reworded true repost still updates in place with score bumped and title preserved) — happy to open a PR.

### CLI tool

Claude Code

### OS and Node.js version

Linux (WSL2) / Node v22.18.0

### Output of npm run doctor

N/A — deterministic code-level repro above; no setup involvement.

---

*Disclaimer: this report was drafted by an AI agent (Claude Code) in the session where the bug was hit, root-caused, and fixed; the human filing it reviewed the report before submission.*

varigg:
## What does this PR do?

Fixes the tracker-clobbering bug from #2165: a sibling req whose only distinguishing qualifier is a slashed acronym ("(CI/CD)") tokenized identically to the bare title — `roleTokens()` split it into `ci cd` and the length filter dropped both halves — so the tier-3 fuzzy match declared two distinct reqs duplicates and the update path overwrote the existing Applied row's title/score/date/report. Two guards: slashed short acronyms now collapse into one surviving token (`ci/cd` → `cicd`, also `tcp/ip`, `ui/ux`) *before* punctuation stripping, and only a tier-1 match (report number + company — provably the same evaluation) may rewrite a matched row's role title.

## Related issue

Fixes #2165

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

Test notes: 2048 passed. New coverage: two matcher unit tests (slash-only-qualifier sibling stays distinct; identical slash-acronym roles across punctuation variants still match) and an e2e regression (the Applied row survives the sibling merge as two rows; a reworded true repost still updates in place with the score bumped and the existing title preserved). One unrelated local environment failure ("Dashboard build failed") comes from Go's VCS stamping inside a linked git worktree — the same tree builds clean in a normal checkout, and this PR touches no dashboard code.

---

*Disclaimer: this fix was root-caused, implemented, and tested by an AI agent (Claude Code) in the session where the bug was hit; a human reviewed the change and this PR before submission.*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fuzzy role matching for slash-separated short acronyms (e.g., “CI/CD”, “tcp/ip”, “ui/ux”) by treating them as a single qualifier.
  * Updated duplicate re-evaluation handling so tracker updates refresh score/evidence without overwriting an existing role title, except when the duplicate is matched by report number plus company.
  * Prevented sibling requirement updates from clobbering unrelated existing tracker rows.
* **Tests**
  * Added matcher and end-to-end regression coverage for slashed-acronym qualifiers and for sibling clobber protection / tier-2 role-title preservation.


FReptar0 (commented):
Strong fix — two complementary guards for a genuinely nasty failure mode (a fuzzy false positive silently overwriting an Applied row's title/score/date/report), and I appreciate that the integration test actually exercises the merge end-to-end rather than just pattern-matching the source. The `roleTokens` slash-collapse and the title-preservation guard are both well-commented.

One substantive question on the title-preservation guard. It keys on `reportNumMatched`, which is set only in the **tier-1** branch (report# + company). But the matcher has three tiers:

- tier-1: report# + company (L560)
- tier-2: exact entry `num` + company (L573) — "the TSV `num` doubles as the tracker row id"
- tier-3: company + role fuzzy (L586)

So `role: reportNumMatched ? addition.role : duplicate.role` now preserves the existing title on **both tier-2 and tier-3**. The #2165 bug was a tier-3 fuzzy false positive, but tier-2 keys on the row-id `num` *plus* company — a deliberate "update row N at this company" — where a corrected/refined title is often exactly what the user intends. Suppressing the title there is a behavior change from today (all tiers used `addition.role`). Was folding tier-2 into title-preservation intentional? If tier-2 should still update the title, the guard could be `reportNumMatched || <tier-2 matched>`. Either way, worth a test pinning tier-2's title behavior — the new integration test covers tier-1 (score bump updates) and tier-3 (fuzzy preserves), but not tier-2.

Minor: the slash-collapse regex targets 1–3 char sides, so `ci/cd`→`cicd` (4 chars) survives the length filter, but a 1+1 like `a/b`→`ab` (2 chars) is still dropped. Fine for the real cases (`ci/cd`, `tcp/ip`, `ui/ux`) — just noting the boundary in case a two-letter slashed qualifier ever shows up.

Non-blocking — the core fix is correct and the test coverage is better than most. Nice work.

varigg:
Thanks for the careful read, and for catching that the guard's blast radius is wider than the bug that motivated it.

Folding tier-2 in was intentional — the rationale is in the comment above the flag (`merge-tracker.mjs` L559-563) — but you're right that the PR description only argues the tier-3 case, so the reasoning never made it in front of a reviewer. Making it explicit:

I'd question the premise that tier-2 is a deliberate "update row N at this company." Tier-2 only fires when tier-1 already failed — that is, the addition carries a report link that did *not* match the row's, while the bare `num` did. The tree's own comment at L582-586 records why that gap exists: report-file numbering and tracker-row numbering drift out of sync (reports maxed at 067 while the tracker was at #69), which is exactly why the company guard was added there in the first place. `AGENTS.md` says the same thing from the other direction — the TSV `num` is the report number, and `set-status.mjs` takes the tracker row `#`, independent sequences. So a tier-2 hit is "these two numbers happen to coincide at this company," which is a coincidence-prone signal rather than an expressed intent. That is the same failure shape as #2165, just rarer — and when it does misfire, the title is the only field that still carries the evidence that two reqs were distinct, since date/score/report/notes are all overwritten unconditionally on the update path regardless of tier.

Where I think you're straightforwardly right is coverage: the new tests pin tier-1 and tier-3 and say nothing about tier-2, so whichever way it's meant to behave is currently unpinned and a future refactor could flip it silently.

I'd rather not expand the diff unilaterally on an open review, so: if you'd prefer tier-2 to keep updating the title, `reportNumMatched || <tier-2 matched>` is a one-line change and I'll add the test alongside it. If the reasoning above holds, I'll add a test pinning tier-2 title-preservation instead. Happy to push either — just say which.

On the `a/b` → `ab` boundary: agreed, and I left it deliberately. Dropping the length filter to 2 would readmit genuinely noisy two-letter tokens across every title, and I couldn't name a real two-letter slashed qualifier worth that trade. Worth revisiting if one ever shows up.

santifer:
Same message I just left on another PR, and same apology, @varigg: I merged #2164 and #1933 tonight and both touch `role-matcher.mjs`, so your branch conflicts because of my merge order. Nothing wrong on your side.

Verified rather than assumed: I fetched your head and merged current `main` into it locally, and it does conflict.

Your change is the one I most want out of that group, because a sibling-req clobber doesn't just hide a listing: it overwrites an Applied row's title, score and report, so it corrupts something the user already did work on. A rebase onto `main` is all it needs, and I'll pick it up as soon as it's green.

abankar1:
Heads-up, @varigg — this now conflicts with `main`: #2107 merged earlier today touching `roleTokens` and the same role-matcher test block, so it needs a rebase (my doing). No logic clash — your slash-collapse in `roleTokens` is orthogonal to #2107's seniority gate, so they compose cleanly.

Fix looks solid, and the second half is the part I'd underline: gating the title rewrite on `reportNumMatched` so only a tier-1 (report#+company) match may replace the row's title, while tier-2/tier-3 fuzzy matches keep the existing title. That's the real data-loss guard — the tokenizer change stops *this* collapse, but title-preservation is what stops the whole class of fuzzy-false-positive clobbers. Nice.

varigg:
Rebased onto current `main` — thanks both for the heads-up, and no trouble at all.

@abankar1's read was right: the conflict was purely positional. #2107's `MTS_PREFIX` substitution and my slash-collapse both insert into `roleTokens()` at the same point, but they're order-independent and both still run before punctuation is stripped, so I kept both with upstream's ordering first. The `test-all.mjs` conflict was two unrelated test blocks landing on the same line. No logic was changed in the resolution.

@FReptar0 — I've added the tier-2 test, and I went with pinning the current behavior (tier-2 preserves the existing title) rather than flipping it, for the reasoning in my earlier reply: tier-2 only fires *after* tier-1 has already failed, so a hit there means the addition's report link did **not** match the row's while the bare `num` did — number coincidence rather than expressed intent. The test isolates tier-2 deliberately: the report links differ so tier-1 misses, and the two role titles are far too dissimilar to fuzzy-match, so tier-2 is the only tier that can fire.

The offer still stands: if you'd rather tier-2 kept updating the title, it's `reportNumMatched || <tier-2 matched>` and I'll swap the test to pin that instead. Either way the behavior is now covered rather than left to a future refactor.

**On CodeRabbit's tier-2 finding** (`merge-tracker.mjs#L694`, "reject uncorroborated tier-2 collisions instead of creating hybrid rows") — the observation is correct and worth having on the record, but I've addressed it in the test rather than by changing behavior, and I'd argue it's out of scope here.

It's right that the merged row is internally inconsistent: the preserved title describes one req while the overwritten report link points at another req's evaluation. What it misses is that this predates the change. Before the guard, the same collision overwrote the title *as well*, so the Applied Data Engineer row simply became a TPM row — the tracker stopped recording that the original req had ever been applied to. Both outcomes are wrong; this one loses strictly less. The genuine fix is that an uncorroborated `num`+company collision shouldn't take the in-place update path at all, which is a tier-2 redesign rather than a #2165 bugfix, and it touches semantics @FReptar0 has already queried above — I'd rather not fold that into this PR unilaterally.

Where it lands a fair hit is that my test *implied* the whole merged row was desirable. Fixed in the follow-up commit, no production code touched:

- the fixture is now documented as a deliberately pathological isolation case, with only the title assertion normative
- the update-path assertion is reworded as characterization ("pre-existing behavior") — it exists solely to keep the title assertion non-vacuous
- the isolation premise is now asserted rather than assumed: if those two titles ever did fuzzy-match, tier-3 could satisfy the assertions and the test would silently stop exercising tier-2

Happy to open a separate issue for the tier-2 redesign if you'd like it tracked — say the word and I'll write it up.

Full suite green on the rebased branch: 2259 passed, 0 failed. One correction to my earlier test note: the "Dashboard build failed" I mentioned was an artifact of my linked worktree; in a clean checkout it simply skips for want of a Go compiler, and this PR still touches no dashboard code.

santifer:
You rebased this at 03:53 because I asked, @varigg, and then I merged a wave of PRs this afternoon that broke it again. That one is on me, so here is the analysis done rather than a bare "please rebase".

**The conflict is one file and one region.** Everything else in your branch auto-merges cleanly:

```
Auto-merging role-matcher.mjs
Auto-merging test-all.mjs
CONFLICT (content): Merge conflict in test-all.mjs
```

The other side is **#2209** (accent folding in `role-matcher`, merged about an hour ago). It adds assertions to the same role-matcher block of `test-all.mjs` that your `CI/CD` sibling-req case lands in. The two changes have nothing to do with each other semantically: one is about accented characters being treated as word separators, yours is about slashed acronyms vanishing in tokenization.

**Resolution: keep both sides.** No judgement call, no ordering constraint. Both assertion blocks stay, one after the other.

Worth restating what your test pins, because it's a genuinely nasty one: `(CI/CD)` tokenized to `ci cd`, both dropped by the length filter, so a sibling req whose *only* distinguishing qualifier was a slashed acronym tokenized identically to the bare title. The #1881 subset guard never saw an extra token, and merge-tracker overwrote the Applied row's title, score and report. A user watching their own tracker lose a row they'd filled in by hand.

Ping me when it's up and I'll take it straight away.

varigg:
> Ping me when it's up and I'll take it straight away.
@santifer Update pushed.

santifer:
Welcome to career-ops, @varigg! Sharp fix: a slashed acronym like "(CI/CD)" in the qualifier was making sibling requirements clobber each other, and your test locks both directions. You rebased twice for this (the second time because my own merge wave broke you): appreciated, and sorry for the churn. Merged: your first contribution. 🚀

PR #2110, opened 2026-07-22
fix(update-system): rollback leaves new files behind under a directory pathspec (#2015)

abankar1:
Addresses the **incomplete directory rollback** in #2015 (secondary problem 1). Scoping note on the primary false-positive at the bottom — I could not reproduce that half on current `main` and would rather say so than claim a fix I can't demonstrate.

Thanks @clede — the diagnosis in the report was exactly right, down to naming the four leftover files.

## The bug

`revertPaths` restores each `SYSTEM_PATHS` entry with `git checkout HEAD -- <path>`. That resets tracked content, but it **never removes paths HEAD doesn't contain**. The existing `catch` only fires when the *whole* pathspec is missing from HEAD — so a directory entry that already exists (`docs/`) checks out cleanly, the catch never runs, and files the update introduced *beneath* it survive the "rollback" as staged additions.

Demonstrated in a scratch repo, mirroring v1.20.0 → v1.21.0:

```console
$ git status --porcelain              # after `git checkout FETCH_HEAD -- docs/`
M  docs/EXISTING.md
A  docs/REVIEWING.md
A  docs/logo.png

$ git checkout HEAD -- docs/          # the rollback as it exists today
$ git status --porcelain              # ← the two new files survived
A  docs/REVIEWING.md
A  docs/logo.png
```

Which is precisely what @clede observed: after `Aborting: user files were touched. Rolling back system files...`, `docs/logo.png`, `docs/wordmark-dark.svg`, `docs/wordmark-light.svg` and `docs/REVIEWING.md` were still staged.

## The fix

After each checkout, delete the entries staged as **additions** relative to HEAD under that pathspec:

```js
git('diff', '--cached', '--name-only', '--diff-filter=A', 'HEAD', '--', spec)
```

Same scratch repo, with the fix applied → `git status --porcelain` is empty.

`--diff-filter=A` is the important constraint: only additions are considered, so a file that merely *changed* is never deleted. The helper also returns early when `git diff` fails (no HEAD yet, unreadable pathspec) rather than guessing.

## On the primary complaint (the false-positive SAFETY VIOLATION)

I deliberately did **not** touch that here. `apply()` already snapshots `initialStatusPaths` before doing anything and skips those files in the safety loop (`update-system.mjs:725` / `:885`), and that guard dates back to #119 — so a pre-existing dirty `interview-prep/story-bank.md` should already be exempt, and I couldn't construct a repro against current `main`.

[...]

PR #2173, opened 2026-07-25
test(update-system): behavioral rollback tests + git-runner seam (follow-up to #2110)

abankar1:
**Draft, stacked on #2110** — do not merge until #2110 lands. The diff currently shows #2110's commits too (main doesn't have them yet); I'll rebase to a clean two-commit diff the moment #2110 merges. **Only the last two commits are this PR** (`715d3a9` seam, `84f03fd` behavioral test).

Follows up on the review of #2110, where both @FReptar0 and CodeRabbit made the same point: for a rollback path this safety-critical, a *behavioral* test is worth more than source-pattern assertions. @FReptar0's exact recipe — "stage an addition under a directory `SYSTEM_PATHS` entry, run the revert, assert the addition is gone **and** a `protectedPaths` file survives" — is what this implements.

## Why it needed a small refactor first

As I noted on #2110, `revertPaths` / `removeAdditionsNotInHead` called the module-bound `git()` (hardcoded to `ROOT`) and deleted via `join(ROOT, …)`, so they could only run against the real working tree — which is why that whole suite is source-pattern today. Rather than test a *copy* of the logic (which drifts), this adds a minimal git-runner seam.

### Commit 1 — the seam (`update-system.mjs`)
- `revertPaths(paths, protectedPaths, ctx)` and `removeAdditionsNotInHead(pathspec, protectedPaths, ctx)` take an optional `ctx = { git, root }`.
- Defaults to the module `git` and `ROOT`, so **production behavior is byte-identical** — `apply()` still calls with two args.
- Exports `revertPaths`, `removeAdditionsNotInHead`, `gitIn` (module has a proper entry guard at `update-system.mjs:1117`, so importing runs no updater logic — confirmed).

### Commit 2 — the behavioral test (`tests/updater-rollback-behavior.test.mjs`)
Drives the **real exported** `removeAdditionsNotInHead` against a throwaway git repo via the seam:
- an addition the update introduced is removed (index **and** worktree);
- a file the user staged *before* the update (`protectedPaths`) **survives** — the no-data-loss guarantee, tested for real;
- a merely *modified* file is never deleted (`--diff-filter=A`);
- a single-file pathspec behaves like a directory one.

Auto-discovered under `tests/`. The rest of `updater-migration-tests.mjs` stays source-pattern per the file's convention.

## Tests
`node tests/updater-rollback-behavior.test.mjs` → 7/7 · `node test-all.mjs --quick` → **2051 passed, 0 failed**.

2 commits (seam + test).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved update rollback to better preserve pre-existing staged changes and modified files.
  * Ensured rollback removes update-introduced added files both from the git index and from the working tree when they no longer exist in `HEAD`.
  * Improved handling so directory and single-file rollback pathspecs behave consistently.
* **Tests**
  * Added dedicated rollback behavior tests using an isolated temporary git repository.
  * Updated rollback-related test expectations to match the new rollback helper behavior and command invocation patterns.

### DIFF
--- a/tests/updater-rollback-behavior.test.mjs
+++ b/tests/updater-rollback-behavior.test.mjs
@@ -1,0 +1,131 @@
+/**
+ * updater-rollback-behavior.test.mjs — BEHAVIORAL rollback tests (#2015 follow-up).
+ *
+ * The rest of updater-migration-tests.mjs verifies the updater by source-pattern
+ * assertions (the file's convention, because apply()/revertPaths are ROOT-bound
+ * with heavy side effects). This file drives the real `removeAdditionsNotInHead`
+ * export against a throwaway git repo via the git-runner seam, so it verifies the
+ * rollback *behaves* right — the property that actually protects user data — not
+ * just that the code reads right (@FReptar0 + CodeRabbit review of #2110).
+ */
+
+import { mkdtempSync, mkdirSync, writeFileSync, existsSync, rmSync } from 'fs';
+import { tmpdir } from 'os';
+import { join } from 'path';
+import { pass, fail } from './helpers.mjs';
+import { gitIn, removeAdditionsNotInHead } from '../update-system.mjs';
+
+// A throwaway git repo plus a ctx that binds the rollback helper's git runner
+// and filesystem root to it, so nothing touches the real working tree.
+function makeRepo() {
+  const dir = mkdtempSync(join(tmpdir(), 'co-rollback-'));
+  const g = (...args) => gitIn(dir, ...args);
+  g('init', '-q', '-b', 'main', '.');
+  g('config', 'user.email', 'test@example.com');
+  g('config', 'user.name', 'Test');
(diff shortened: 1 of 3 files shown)


abankar1:
Rebased onto `main` now that #2110 has merged — the diff is clean: just the two follow-up commits (`8918f66` git-runner seam + `97d9de8` behavioral test), no more stacked #2110 content. Ready for review. CI green (`test` SUCCESS), 2051/0 local.

santifer:
Merged, @abankar1. This is the upgrade the updater tests needed: the existing ones assert that the *source contains a pattern*, which passes happily when someone refactors the behaviour away but keeps the string. Running the rollback against a throwaway git repo through an injectable runner tests what actually happens.

Follow-up to #2110 done properly, and the seam is small enough not to become the thing that needs testing itself. 🚀

PR #2182, opened 2026-07-26
fix(web): render PDFs from the backend instead of the spawned agent

### LINKED ISSUE #2172: Web dashboard's 'pdf' mode can crash under Codex: sandboxed Chromium launch fails, agent improvises past the real pipeline
### Code of Conduct

### Existing issues

### What happened?

Running the web dashboard's "Generate CV PDF" ("pdf" mode) against Codex as the headless CLI can fail with a macOS Chrome crash, and — worse — the agent then improvises past it in a way that bypasses this repo's actual PDF pipeline.

Root cause: `web/src/app/api/run/route.ts`'s "pdf" mode prompt tells the **spawned AI agent** to run `node generate-pdf.mjs` itself, inside its own Bash tool call — inside whatever sandbox that agent CLI imposes. Launching a browser from a sandboxed subprocess can hit an interactive sandbox-escalation prompt. That's fine for an interactive terminal session (a human can approve it), but for a **headless, web-triggered** run there is no one present to click "approve" — so any CLI whose default sandbox policy blocks a browser launch fails deterministically, every time, with no way to unblock it from inside that run.

Observed cascade in one real run:
1. Codex's own sandbox blocked Playwright's bundled headless Chromium from launching.
2. Codex fell back to the installed system Google Chrome binary — which crashed too, with a macOS `EXC_CRASH (SIGABRT)` in `_RegisterApplication`/`TransformProcessType` (a GUI-app-registration call that only fires for a non-headless/foreground-process launch attempt, consistent with the sandbox blocking the Mach/XPC lookups a normal headless launch wouldn't need).
3. Codex then fell back to **`wkhtmltopdf`** directly — a tool with zero references anywhere in this repo's actual code (`generate-pdf.mjs` only ever calls `chromium.launch()` from Playwright). This bypassed the documented pipeline entirely, contrary to `modes/pdf.md`'s explicit "do not improvise a format" instruction, and produced an inconsistent `data/pdf-index.tsv` entry (PDF recorded with no retained HTML source).

Separately confirmed empirically: `node generate-pdf.mjs`, run directly and standalone (outside any agent's sandbox), works perfectly on the same machine — so this is not a bug in `generate-pdf.mjs` itself, it's purely about *who* invokes it and from inside what sandbox.

### Steps to reproduce

1. Configure the web dashboard to use Codex as the CLI.
2. Trigger a "Generate CV PDF" run for an existing report.
3. If Codex's sandbox policy blocks a browser launch (default in some configurations), the agent may crash trying system Chrome and/or silently substitute a different renderer instead of surfacing a clear error.

### Expected behavior

PDF rendering should not depend on the spawned agent's own sandbox permissions at all, and should never silently substitute an unrelated rendering engine outside the documented pipeline.

### Proposed fix (implemented and verified for the web dashboard)

Move PDF rendering out of the agent's hands entirely — the agent's job stops at producing tailored HTML; the **backend itself** (a plain Node process with no CLI-imposed sandbox) spawns `node generate-pdf.mjs` directly, exactly as `modes/pdf.md` already documents, just from a different caller:

- Agent writes the tailored HTML + a small `{"format": "letter"|"a4"}` sidecar to a backend-provided deterministic scratch path (no more self-chosen filename).
- Once the agent's process closes cleanly, the Next.js backend spawns `generate-pdf.mjs` itself as a plain child process, computes the final `output/cv-{candidate}-{company}-{date}.pdf` path directly (matching the existing naming convention, so web and CLI output stay identical), and streams `done`/`error` based on *that* render's actual result.
- This isn't Codex-specific — it fixes the same failure mode for every CLI the dashboard supports (Claude, Gemini, OpenCode, etc.), since none of them should need sandbox-escalation approval for a step that has no user present to grant it.

Implemented here: [e6d3d11](https://github.com/nikolaysm/career-ops/commit/e6d3d11cff129e7eb79b6d6a2377d9a918d2e306) (`fix(web): render PDFs from the backend instead of the spawned agent`, on a fork branch — happy to open this as a PR against `main` if useful).

For CLI-only (non-web) usage where a user *does* want Codex to render PDFs directly in an interactive session, the workaround is a scoped sandbox-escalation rule for the `node generate-pdf.mjs` command prefix, rather than disabling Codex's sandbox globally — but that doesn't help the headless/web-triggered case at all, which is why the backend-side fix above is the real solution there.

### CLI tool

Codex (crash observed here); the underlying design issue applies to any CLI whose sandbox can block a browser launch in a headless context.

nikolaysm:
## Summary

Closes #2172. The web dashboard's "pdf" mode told the spawned headless AI agent to run `node generate-pdf.mjs` itself, inside its own Bash tool call, inside whatever sandbox that CLI imposes. Launching Playwright/Chromium from a sandboxed subprocess can hit a sandbox-escalation prompt with nobody present to approve it for a headless/web-triggered run — the reported cascade was a blocked sandboxed Chromium, a system-Chrome fallback that crashed too, then the agent improvising `wkhtmltopdf` and bypassing this repo's real pipeline.

- The agent (`web/src/app/api/run/route.ts`'s `"pdf"` prompt) now only tailors content and writes the HTML + a small `{"format": ...}` sidecar to backend-owned deterministic paths (`web/src/lib/pdf-paths.mjs`).
- Once the agent's process closes cleanly and the HTML artifact is verified to exist, the Next.js backend — a plain Node process, no CLI sandbox — spawns `generate-pdf.mjs` itself via `web/src/lib/pdf-render.mjs` and streams `done`/`error` based on the render's actual result.
- The tracker's PDF column is marked ready only **after a confirmed successful render**, via a new canonical script `mark-pdf-ready.mjs` (mirrors `set-status.mjs`'s lock/atomic-write path; the shared CLI scaffold — exit codes, `failWith`, lock acquisition — is now factored into `tracker-utils.mjs` so both scripts use one implementation). This closes a Pipeline Integrity gap where the PDF column previously had no canonical writer at all.
- `"pdf"` is no longer granted the `Bash` tool, since it never shells out anymore — this closes the actual hole the bug depended on.
- Non-fatal render issues (missing/garbled format sidecar, a tracker row that didn't get marked) now surface as visible warnings in the run log instead of only a server-side `console.error`, matching the honesty-gate discipline the `evaluate` path already has.

## Test plan

- [x] `node test-all.mjs` (full run) — 2063 passed, 0 failed
- [x] `node set-status-tests.mjs` — 44/44 pass (verifies the `set-status.mjs` scaffold refactor is behavior-identical)
- [x] `cd web && npm test` — 45/45 pass
- [x] New coverage: `tests/mark-pdf-ready.test.mjs` (17 cases incl. every exit code), `web/tests/lib/pdf-paths.test.mjs` (8 cases), `web/tests/lib/pdf-render.test.mjs` (18 cases, fake child-process based — no real subprocess spawned)

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated `pdf` mode to `POST /api/run`, producing deterministic HTML/format artifacts and final PDFs with Letter/A4 support.
  * Introduced a CLI to mark a tracker row’s PDF cell as ready, with `--dry-run` and `--json`.
* **Bug Fixes**
  * Tracker PDF marking now happens only after a confirmed successful render; includes safer handling for missing/ambiguous reports and improved lock/write failure behavior.
* **Tests**
  * Added regression coverage for the CLI and the PDF path/render workflow; broadened the web test runner.
* **Chores**
  * Expanded system auto-update scope and updated cache/tmp ignore rules.


nikolaysm:
Two CodeRabbit findings left open by design, not oversight:

- Path-scoped Write/Edit for `pdf` mode — tracked in #2185.
- No explicit timeout on `renderAndMarkPdf()` itself (only the agent's `killMs` + route `maxDuration` bound it) — tagged "Heavy lift" by CodeRabbit since it needs an abort/child-handle plumbed through `pdf-render.mjs`; will file as a follow-up issue once this merges rather than growing this diff further.

santifer:
Your only conflict is in `.gitignore`, and I can give you the resolution with certainty because I resolved this exact one locally on another PR tonight and ran the suite against it.

`main` replaced the per-file list under `data/` with a single blanket rule:

```gitignore
local/
# Everything under data/ is generated or personal at runtime — only .gitkeep
# placeholders are tracked. A blanket rule means a new generated file is
# covered for every user automatically.
data/*
!data/.gitkeep
!data/offers/
!data/parser-output/
```

Your branch still carries the old enumerated list. **Take `main`'s version wholesale.** Anything your branch added under `data/` is already covered by `data/*`, so nothing of yours is lost: I verified that on a full resolution, and `node test-all.mjs` came back green.

Concretely: `git checkout --theirs .gitignore` during the merge (or `git checkout main -- .gitignore` after), and you're done. That's the whole conflict.

nikolaysm:
@santifer Rebased onto latest main — the `.gitignore` conflict is resolved (took main's blanket `.career-ops-web/` rule, same as the `data/*` blanket you described; the two narrower lines this branch had are subsumed by it). No conflicts remain.

CI is green across the board, including the new windows-latest leg on the Tests workflow: that leg exposed a latent cross-platform bug in `tests/mark-pdf-ready.test.mjs`'s write-failure test — `chmodSync`'s read-only directory bits don't block writes on Windows, so the simulated permission-denied never actually fired there. Fixed it to use `icacls` deny-ACL on win32, mirroring the same pattern `set-status-tests.mjs` already uses for its own write-failure test.

Ready for another look.

nikolaysm:
@santifer Just checking in — is there anything else you'd like me to address before this is ready to merge, or are you good to take another look when you have time?

santifer:
Substantial security win, @nikolaysm — moving PDF rendering to the backend means the spawned agent loses Bash entirely, the render format goes through an allowlist (so the agent-written sidecar can't inject arguments), and the whole path from reportNum to spawn is constrained end to end. The web maintainer's review called out the injected spawnFn making your 400+ test lines real tests, and I agree. This also closes your #2172 (the sandboxed-Chromium crash disappears with the render moved server-side). #2185 now has solid ground to build on. Merged! 🚀

PR #1114, opened 2026-06-19
Tracker pipeline: i18n + optional Location column + integrity checks

nikitacometa:
Consolidates three related `applications.md` tracker-pipeline fixes into one PR. They touch the same parsing core (`dedup-tracker` / `merge-tracker` / `verify-pipeline` + `test-all.mjs`) and would otherwise conflict, so they're cleaner as one reviewable unit — each is a separate, self-contained commit.

## 1. Non-Latin company/role names break dedup + merge (i18n)

`normalizeCompany` (in **both** `dedup-tracker.mjs` and `merge-tracker.mjs`) and the shared `roleTokens` stripped every non-`[a-z0-9]` character, so Cyrillic / CJK / Arabic company names collapsed to one empty key (dedup merged unrelated employers) and non-Latin role titles tokenized to `[]` (so `roleFuzzyMatch` bailed and real duplicates never merged). Now uses `\p{L}\p{N}` + NFC normalization; Latin behavior is unchanged. Helps the users the localized `de/fr/ja/ar/tr` modes target.

## 2. Optional Location column corrupts dedup / normalize / analyze / followup

`dedup-tracker`, `normalize-statuses`, `analyze-patterns`, and `followup-cadence` hardcoded `parts[5]=score` / `parts[6]=status`. `merge-tracker` and `verify-pipeline` already detect columns by header, so with the optional Location column enabled the four scripts read **Score as Status** (and could write a status back into the Score cell). Introduces **`column-map.mjs`** as the single source of truth (`merge-tracker` + `verify-pipeline` now import it — they previously had two verbatim copies), and the four scripts resolve columns by header. Legacy 9-column trackers are unaffected.

## 3. verify-pipeline integrity checks

Two warning-only checks (exit code unchanged): duplicate report **file numbers**, and an unparseable `## Machine Summary` block on Applied/Responded/Interview/Offer rows (the regex mirrors `analyze-patterns.parseMachineSummary`). Adds a `CAREER_OPS_REPORTS_DIR` env override for testability.

---

Each fix has a regression test in `test-all.mjs`. `node test-all.mjs --quick` → **391 passed, 0 failed**.

## Release Notes

* **New Features**
  * Application tracker now supports flexible column layouts, including optional Location column.
  * Added validation checks for duplicate report files and Machine Summary content verification.

* **Bug Fixes**
  * Improved internationalization: Unicode-aware company and role name processing for better deduplication and matching across non-Latin characters.

* **Tests**
[...]

PR #2186, opened 2026-07-26
fix(normalize-statuses): use header-mapped columns instead of fixed indices

### LINKED ISSUE #1955: normalize-statuses.mjs hardcodes legacy column indices, risking custom tracker corruption
### Description
In the Career Ops tracker (`applications.md`), the column layout is designed to be fully customizable (e.g. users can insert a custom `Location` or `Date` column). To support this, several modules use `resolveColumns` and `detectColumns` from `./tracker-parse.mjs` to dynamically map headers to their respective indices (e.g. in `dedup-tracker.mjs`, `merge-tracker.mjs`, and `reconcile-pipeline.mjs`).

However, `normalize-statuses.mjs` still relies on legacy, hardcoded column indices when parsing and updating columns:
* Line 125: `parts[6] = result.status;` (assumes column 6 is the status column)
* Line 128: `parts[9]` (assumes column 9 is the notes column)
* Line 139: `parts[5] = parts[5].replace(/\*\*/g, '');` (assumes column 5 is the score column)

If a user modifies their tracker layout by adding or reordering columns, running `normalize-statuses.mjs` will corrupt their tracker by writing statuses and notes into incorrect columns.

### Proposed Solution
We should update `normalize-statuses.mjs` to dynamically detect columns using `detectColumns` from `./tracker-parse.mjs`.

Specifically:
1. Import `detectColumns` from `./tracker-parse.mjs`.
2. Locate the header row of `applications.md` and pass it to `detectColumns` to obtain the column mapping object (`colmap`).
3. Replace all hardcoded column references with the dynamically resolved indices from `colmap` (e.g. `colmap.status`, `colmap.notes`, `colmap.score`).

ddy4633:
Closes #1955. Credit to @SparshGarg999 for the report.

## Overlap with #1114 — your call, not mine

@nikitacometa's #1114 covers this ground as part of a larger change. I'm opening this anyway because that PR has been conflicted with no push since 2026-07-11, and @FReptar0's note there gave a ~1-week stale window that has now passed. If you'd rather revive #1114, close this — no hard feelings, and this PR does not try to replace it.

The scopes differ:

| | #1114 | this |
|---|---|---|
| Files | 9 | 1 (+1 test file) |
| Column mapping | introduces its own `column-map.mjs` | uses `tracker-parse.mjs`, already on `main` |
| Also does | i18n, integrity checks | nothing else |

`tracker-parse.mjs` landed on `main` after #1114 opened and does the job `column-map.mjs` was written for, so this rides the current abstraction rather than a competing one. @santifer's comment on #1114 — *"killing the parts[5]/parts[6] hardcoding is the right fix"* — is the direction I followed.

## The bug

`normalize-statuses.mjs:115-154` parsed and rewrote rows by fixed index (`parts[6]` status, `parts[5]` score, `parts[9]` notes, `parts[1]` num) while the rest of the toolchain resolves columns from the header. With the documented optional `Location` column present:

```
| 1 | 2026-02-01 | Acme | Data Engineer | Remote | — | Aplicado 2026-02-02 | ❌ | — | backfilled, no eval |
```

```
#1: "—" → "Discarded"
⚠️  1 unknown statuses:  #2 (line 6): "4.5/5"
→ | 1 | ... | Remote | Discarded | Aplicado 2026-02-02 | ...
```

The `—` no-evaluation sentinel in **Score** was read as a status and overwritten with `Discarded`; the real status was left non-canonical; row 2's genuine `Rechazada` was reported as unknown status `"4.5/5"`. This is a writer that also `.bak`s over the previous backup, so the last good copy goes with it.

## The fix

One logical change in `normalize-statuses.mjs` (+39/−20): resolve columns once with `resolveColumns(lines)`, read rows through `parseTrackerRow()`, write to `parts[COLS.status]` / `parts[COLS.notes]` / `parts[COLS.score]`. The `moveToNotes` branch is guarded on `COLS.notes != null` — a layout with no Notes column has nowhere to put DUPLICADO provenance, and the old `else` appended a headerless cell.

## Tests

Added to `tracker-columns-tests.mjs` (already wired into `test-all.mjs:172`), and removed the now-stale TODO at `:230` that said *"normalize-statuses.mjs is excluded until PR #1114 (which retrofits it) lands."*

- **6b** — 10-col tracker: status normalized in Status, Score untouched, DUPLICADO provenance lands in Notes preserving the existing note
- **7 contract** — `normalize-statuses` added to the unknown-extra-column contract alongside `verify-pipeline` / `tracker.mjs` / `scan.mjs`: file returns byte-identical, no unknown statuses

Regression proof — `git checkout -- normalize-statuses.mjs`, tests kept:

```
FAIL normalize-statuses on 10-col tracker (code 0)
FAIL normalize-statuses DUPLICADO row on 10-col tracker
FAIL contract: normalize-statuses on unknown-column tracker (code 0)
40 passed, 3 failed
```
Restored → `43 passed, 0 failed`.

Full suite: `node test-all.mjs --quick` → **2042 passed, 0 failed, 2 warnings**, byte-identical pass/fail set vs. baseline (the 2 warnings are pre-existing: no user data for `cv-sync-check`, no Playwright browser for archive render).

Back-compat spot-checks: legacy 9-col with no trailing pipe keeps its notes cell; bold stripped from score; headerless tracker still works via `LEGACY_COLMAP`.

## One behaviour change worth flagging

`parseTrackerRow` enforces a strict full-width guard where the old code accepted `parts.length >= 9`. A **ragged** row narrower than its own header is now silently skipped rather than normalized (verified: reports `0 statuses normalized`).

I think fail-closed is right for a writer — rewriting a short row by index is exactly how columns get shifted — and it matches every other reader. But it *is* silent. If you'd prefer it warn like the `unknown statuses` block does, say the word; I left that out since `verify-pipeline.mjs` is the designated integrity checker.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved status normalization for trackers with reordered or additional columns.
  * Prevented score values from being overwritten during normalization.
  * Ensured duplicate provenance is added to the Notes field when available.
  * Preserved tracker content when unknown extra columns are present.

* **Tests**
  * Added coverage for varied tracker layouts and unchanged-file behavior.

### DIFF
--- a/normalize-statuses.mjs
+++ b/normalize-statuses.mjs
@@ -17,6 +17,7 @@ import { fileURLToPath } from 'url';
 import {
   openTrackerTransaction, rebuildRow, resolveTrackerPath,
 } from './tracker-utils.mjs';
+import { resolveColumns, parseTrackerRow } from './tracker-parse.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
 const APPS_FILE = resolveTrackerPath(CAREER_OPS);
@@ -112,19 +113,21 @@ const lines = content.split('\n');
 let changes = 0;
 let unknowns = [];
 
+// Map columns by header name (tracker-parse.mjs, #954/#1596). Fixed indices
+// assumed the original 9-column layout, so a customized tracker — an inserted
+// Location or Via column — shifted every field one to the left: the Score cell
+// was normalized as if it were the status and overwritten, while the real
+// status was left alone and reported as unknown (#1955).
+const COLS = resolveColumns(lines);
+
 for (let i = 0; i < lines.length; i++) {
   const line = lines[i];
-  if (!line.startsWith('|')) continue;
+  const row = parseTrackerRow(line, COLS);
+  if (!row) continue; // header, separator, non-row, or a row missing cells
 
   const parts = line.split('|').map(s => s.trim());
-  // Format: ['', '#', 'fecha', 'empresa', 'rol', 'score', 'STATUS', 'pdf', 'report', 'notas', '']
-  if (parts.length < 9) continue;
-  if (parts[1] === '#' || parts[1] === '---' || parts[1] === '') continue;
-
-  const num = parseInt(parts[1]);
-  if (isNaN(num)) continue;
-
-  const rawStatus = parts[6];
+  const num = row.num;
+  const rawStatus = row.status;
   const result = normalizeStatus(rawStatus);
 
   if (result.unknown) {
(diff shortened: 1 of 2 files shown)


santifer:
Welcome to career-ops, @ddy4633! This is a nasty one to spot: on a tracker with an extra column every field landed one to the left, so the script normalized the Score cell as if it were the status: a `—` score mapped to Discarded and overwrote a real value, while the actual status went untouched and got reported as unknown. Corrupting data while reporting the wrong problem is the worst combination. Reusing `tracker-parse` instead of hand-rolling another parser is exactly right. Merged.

PR #2193, opened 2026-07-26
Build public-safe Career-Ops Python CLI

shubhamdahiya210902-tech:
﻿## Summary
- Adds a public-safe Python Typer CLI for Career-Ops with scan, score, tailor, review, tracker, and export commands.
- Uses only synthetic mock fixtures for jobs, candidate profile, and tracker examples.
- Implements no-invention safety review and an explicit human approval gate before export.
- Adds pytest coverage and a Python GitHub Actions workflow.

## Safety
- No real CVs, applications, private job memory, contact details, or submitted materials are included.
- Export is blocked unless review passes and `--approve APPROVE_EXPORT` is provided.
- Drafts remain non-final and label known gaps instead of claiming them.

## Validation
- `python -m ruff check src tests`
- `python -m pytest` -> 21 passed
- Scoped staged privacy scan for known sensitive literals returned no matches.


santifer:
Hey @shubhamdahiya210902-tech, thanks for the effort here: a full Typer CLI with mock fixtures, tests and CI is real work, and the no-invention safety gate shows you understood the project's values.

Before I can review it properly, I need to understand the fit, because this lands as a second, parallel implementation of career-ops rather than an extension of it. CONTRIBUTING's "What belongs in core" section (added this week) is the honest lens I'll use: the core path already does scan, score, tailor and track through the existing engine, so a Python reimplementation duplicates that path instead of extending it, and every future change would need to land twice to keep the two in sync.

So help me out: what does this solve that the existing flow can't? If the answer is "Python developers want a Python-native way in", that's a genuinely interesting direction, but it points to a *separate project* (your own repo, which we could link from the README's ecosystem section) rather than a second engine inside this one. If there's a concrete gap in the current CLI you were working around, tell me which: that part might be worth landing here on its own.

santifer:
Second check-in, @shubhamdahiya210902-tech: the rebase ask from 16 days ago is still the only blocker here. No need to finish anything today, a one-line signal keeps it open. If I hear nothing in 7 days I will close it to keep the queue honest: reopening is one comment away whenever you come back, and the work stays yours.

santifer:
Closing as promised in the Aug 12 check-in, @shubhamdahiya210902-tech: no signal in the window. Worth saying plainly for whenever you return: a parallel Python re-implementation of the CLI would be a hard sell here anyway, since the project keeps one canonical core and alternative front-ends live best as their own forks or plugins. If there is a specific gap that pushed you to build it, an issue describing that gap would be very welcome.

santifer:
@shubhamdahiya210902-tech, when this closed we promised a pointer to a good first issue and never sent one: sorry about that. The live pool is at https://github.com/santifer/career-ops/issues?q=is%3Aopen+label%3A%22good+first+issue%22+no%3Aassignee, claimable with `/assign`. Your Python CLI here showed real building energy; the core is Node, so the smoothest path in is a small .mjs-side task from the pool, and I'm happy to review quickly whenever you pick one.

PR #2200, opened 2026-07-27
fix(salary-gap): parse US ranges with currency on both bounds

### LINKED ISSUE #2190: fix(salary-gap): parseAmount returns null for US-format ranges that repeat the currency symbol ($123,684-$254,644)
## Symptom

`salary-gap.mjs` silently drops the advertised figure for any US-style posting that repeats the currency symbol on both bounds of a range. The report's `advertised_comp` is read correctly and the number is clearly present, but the dashboard renders `advertised —` and the figure lands in `unparseable` instead.

This is the exact shape that US pay-transparency laws produce, so it is not an edge case: Greenhouse, Ashby, and Lever postings from CA/CO/NY/WA routinely publish `$123,684—$254,644 USD`.

## Reproduction

```js
import { parseAmount } from './salary-gap.mjs';

parseAmount('$123,684—$254,644 USD')  // null   <- bug
parseAmount('$123,684-$254,644 USD')  // null   <- same bug, plain hyphen
parseAmount('123684-254644')          // { min: 123684, max: 254644, mid: 189164 }
parseAmount('$150,000')               // { min: 150000, max: 150000, mid: 150000 }
parseAmount('80-90k')                 // { min: 80000, max: 90000, mid: 85000 }
```

Note that the dash character is **not** the problem. The range regex already accepts `[-–—]`, and the plain-hyphen form fails identically.

## Cause

`salary-gap.mjs:59` strips only a **leading** currency symbol:

```js
s = s.replace(/^[€$£¥]\s*/, '').replace(/\s*[A-Za-z]{3}\s*$/, '').trim();
```

For `$123,684—$254,644 USD` that yields `123,684—$254,644`. The range matcher at `salary-gap.mjs:64` then requires `[\d.,]+` immediately after the separator:

```js
const range = s.match(/^([\d.,]+)\s*(k)?\s*[-–—]\s*([\d.,]+)\s*(k)?$/i);
```

The surviving `$` on the upper bound fails that class, the single-value matcher at line 72 also fails, and `parseAmount` returns `null`.

A single value (`$150,000`) works because there is only one symbol and it is leading. The bug needs a range **and** a repeated symbol, which is why it has gone unnoticed: the existing unit tests at lines 361-364 cover `84k`, `84000`, `80-90k`, and `80k-90k`, none of which carry a currency symbol at all.

## Impact

`advertised_comp` is spec'd in `batch/batch-prompt.md` as the JD's figure **verbatim**, so a correctly written report is precisely what triggers the failure. Downstream:

- the desired/advertised gap for that application cannot be computed
- `--summary` shows `advertised —` next to a report that plainly states a range
- the row is reported under `unrecognized sources` / unparseable rather than as a parse defect, which reads like a report-authoring mistake rather than a tool limitation

## Suggested fix

Strip currency symbols throughout rather than only at the start, before matching:

```js
s = s.replace(/[€$£¥]/g, '').replace(/\s*[A-Za-z]{3}\s*$/, '').trim();
```

That keeps every currently passing case green (the trailing ISO-token strip and the `k` magnitude suffix are untouched) and makes both repeated-symbol forms parse. Worth confirming it does not swallow a symbol used as a non-currency marker anywhere else in the observation format.

## Suggested tests

Add to the `parseAmount` block around `salary-gap.mjs:361`:

```js
assert(parseAmount('$123,684—$254,644 USD')?.mid === 189164, 'US range, symbol on both bounds, em dash');
assert(parseAmount('$123,684-$254,644 USD')?.mid === 189164, 'US range, symbol on both bounds, hyphen');
assert(parseAmount('€80,000-€90,000')?.min === 80000, 'EUR range, symbol on both bounds');
assert(parseAmount('$150,000')?.mid === 150000, 'single value with symbol still works');
```

The first two should fail against current `main`; the last is a regression guard for the path that already works.

## Related

- #1656 established the observation model these figures feed
- #2019 and #2025 both reason over `advertised_comp`, so they inherit this parse gap for US-format ranges

Solaris-star:
## What does this PR do?

`parseAmount` failed on US pay-transparency ranges that repeat the currency symbol on both bounds (e.g. `$123,684—$254,644 USD`), so `advertised_comp` landed in `unparseable` and the dashboard showed `advertised —`. Strip `€$£¥` anywhere before matching and cover the cases in `--self-test`.

## Related issue

Fixes #2190

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] I ran `node salary-gap.mjs --self-test` and all tests pass

---

### DIFF
--- a/salary-gap.mjs
+++ b/salary-gap.mjs
@@ -52,11 +52,13 @@ const TRUST = {
 export function parseAmount(raw) {
   let s = String(raw ?? '').trim();
   if (!s || s === '?' || s === '-' || /^(n\/?a|null)$/i.test(s)) return null;
-  // Strip a leading currency symbol and a trailing 3-letter ISO-4217-style alpha
-  // token (any case — "450k SEK", "80-90k eur"). Exactly three letters, so the
-  // lone "k" magnitude suffix ("80k") is never eaten, and prose ("competitive")
-  // still fails the numeric match below even after losing its last three letters.
-  s = s.replace(/^[€$£¥]\s*/, '').replace(/\s*[A-Za-z]{3}\s*$/, '').trim();
+  // Strip currency symbols anywhere (US pay-transparency ranges often repeat the
+  // symbol on both bounds: "$123,684—$254,644 USD") and a trailing 3-letter
+  // ISO-4217-style alpha token (any case — "450k SEK", "80-90k eur"). Exactly
+  // three letters, so the lone "k" magnitude suffix ("80k") is never eaten, and
+  // prose ("competitive") still fails the numeric match below even after losing
+  // its last three letters.
+  s = s.replace(/[€$£¥]/g, '').replace(/\s*[A-Za-z]{3}\s*$/, '').trim();
   const toNum = (numStr, kFlag) => {
     const n = parseFloat(numStr.replace(/,/g, ''));
     return Number.isNaN(n) ? null : (kFlag ? n * 1000 : n);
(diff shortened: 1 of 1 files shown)


santifer:
Good catch, @Solaris-star: the anchor made the old regex correct for "€80-90k" and silently wrong for "$123,684—$254,644 USD", which is exactly how US pay-transparency postings are written, so the parser was returning null on the format it will meet most often. Four assertions covering both dash characters and the single-value case is the right amount of test for a one-character regex change. Merged 🚀

PR #2206, opened 2026-07-27
fix(tracker): set SQLite busy_timeout to avoid SQLITE_BUSY under concurrent access

### LINKED ISSUE #1957: tracker.mjs SQLite DB connection lacks busy_timeout, causing crashes on concurrent dashboard/CLI sessions
### Description
In `tracker.mjs`, the derived SQLite database is opened using Node's native `node:sqlite` API via `new DatabaseSync(DB_PATH)`. 

Since the SQLite database is auto-synced/rebuilt from the markdown file when the index is stale (e.g., when running a CLI query, status changes, or web TUI dashboard requests), multiple processes reading/writing the database can easily run concurrently.

Because `tracker.mjs` does not configure SQLite's `busy_timeout` (which defaults to 0ms in SQLite), any database lock contention (e.g., when a write/transaction is active and another query is run) immediately results in an unhandled `SQLITE_BUSY` exception and crashes the CLI or the TUI dashboard.

### Proposed Solution
Modify `openDb` in `tracker.mjs` to execute `db.exec('PRAGMA busy_timeout = 5000')` immediately after creating/opening the `DatabaseSync` instance. 

This tells SQLite to wait/retry for up to 5 seconds when lock contention is encountered, preventing crash/race conditions on concurrent dashboard/CLI sessions.

gunjanjaswal:
Opening the index with the default `busy_timeout = 0` means the moment two callers touch `applications.db` at once (a `query`, a `set-status` write, the Go TUI dashboard), one throws `SQLITE_BUSY` instead of waiting for the lock. This sets `busy_timeout = 5000` in `openDb`, so callers wait up to 5s for the lock to clear.

Added `tests/tracker-busy-timeout.test.mjs`, which reads the pragma back off a fresh connection (`PRAGMA busy_timeout` == 5000). Deterministic, no timing. `openDb` is exported so the test can reach it.

Kept it to the one pragma. WAL mode would also help here, but it creates `-wal`/`-shm` sidecar files, which is a behavior change I did not want to slip in. Happy to do it as a follow-up if you would like.

Closes #1957

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database reliability by waiting up to five seconds when the database is temporarily locked, reducing failures during concurrent access.
* **Tests**
  * Added regression coverage to verify the database lock-wait behavior is configured correctly.

### DIFF
--- a/tests/tracker-busy-timeout.test.mjs
+++ b/tests/tracker-busy-timeout.test.mjs
@@ -1,0 +1,39 @@
+// tests/tracker-busy-timeout.test.mjs — regression coverage for #1957.
+//
+// openDb() must set a non-zero SQLite busy_timeout. Without it the default is
+// 0, so a CLI query, a `set-status` write and the Go TUI dashboard touching
+// the derived index at the same time throw SQLITE_BUSY on the first contention
+// instead of waiting for the lock to clear.
+//
+// The check is deterministic: open a fresh connection and read the pragma back
+// with `PRAGMA busy_timeout`. No timing, no second connection — nothing flaky.
+import { pass, fail } from './helpers.mjs';
+import { join } from 'path';
+import { mkdtempSync, rmSync } from 'fs';
+import { tmpdir } from 'os';
+
+console.log('\ntracker.mjs — SQLite busy_timeout (#1957)');
+
+// openDb reads DB_PATH once at import time, so point it at a throwaway file
+// before importing tracker.mjs — the test never touches a real applications.db.
+const work = mkdtempSync(join(tmpdir(), 'cops-busy-'));
+process.env.CAREER_OPS_TRACKER_DB = join(work, 'applications.db');
+
+try {
+  const { DatabaseSync } = await import('node:sqlite');
+  const { openDb } = await import(new URL('../tracker.mjs', import.meta.url).href);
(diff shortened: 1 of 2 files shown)


santifer:
Merged, @gunjanjaswal. The bug is real and nasty precisely because it only shows up when someone is actually using the tool properly: a query, a `set-status` write and the Go dashboard all touching the derived index at once, and with `busy_timeout` at its default of 0 the very first contention throws instead of waiting.

The test is the part I want to point at: reading the pragma back with `PRAGMA busy_timeout` makes it deterministic, where the obvious version of this test would race two connections and flake in CI forever. Pointing `CAREER_OPS_TRACKER_DB` at a throwaway file before the import, because `openDb()` reads the path once at import time, is the kind of detail that separates a test that passes from a test that's correct. 🚀

gunjanjaswal:
Thanks @santifer, glad it landed. You put your finger on exactly why it was worth reading the pragma back: racing two connections would have flaked in CI forever, and this one just asserts the setting is actually there. The `CAREER_OPS_TRACKER_DB`-before-import bit took me a second to spot too, since `openDb()` grabs the path once at import time, so anything set after that is too late. Appreciate the thorough read.

PR #2209, opened 2026-07-27
fix(role-matcher): fold accented Latin letters before tokenizing role titles

### LINKED ISSUE #2207: role-matcher.mjs: accented "Sênior" produces a phantom token that false-matches unrelated roles
### Code of Conduct

### Existing issues

### What happened?

roleFuzzyMatch() in role-matcher.mjs strips non-[a-z0-9\s] characters before tokenizing (roleTokens(), line 74: .replace(/[^a-z0-9\s]/g, ' ')). This regex does not fold accented Latin characters to their ASCII base, so "Sênior" (with ê) is split into two separate tokens: "s" and "nior".

"s" is dropped (length ≤ 3).
"nior" is length 4, passes the length filter, and is not in ROLE_STOPWORDS (which only lists the ASCII "senior").
Since "nior" survives tokenization for any title using the accented spelling, two titles that only share generic/baseline words plus this phantom token can cross the 0.6 Jaccard threshold and get treated as the same openi

### Steps to reproduce

Two genuinely different roles at the same company, both titled with accented "Sênior":

"Software Engineer Node.js Sênior"
"Software Engineer Flutter Sênior"
import { roleFuzzyMatch, roleTokens } from './role-matcher.mjs';

roleTokens('Software Engineer Node.js Sênior');
// -> ['software', 'engineer', 'node', 'nior']   ("js" dropped, len 2; "sênior" -> "s nior" -> "nior" kept)

roleTokens('Software Engineer Flutter Sênior');
// -> ['software', 'engineer', 'flutter', 'nior']

roleFuzzyMatch('Software Engineer Node.js Sênior', 'Software Engineer Flutter Sênior');
// -> true  (should be false: only "software"/"engineer" are legitimate overlap, both baseline)
Without the phantom "nior" token, overlap is {software, engineer} — both baseline tokens, and discriminating.length === 0 would correctly return false per the existing baseline-token guard (the guard exists precisely to prevent generic-word-only matches).

### Expected behavior

Impact
merge-tracker.mjs uses roleFuzzyMatch() (via role-matcher.mjs) as its tier-3 duplicate-detection heuristic. In production this caused a real Mottu tracker addition — Software Engineer Flutter Sênior (new, distinct req, different report) — to be silently skipped as a "duplicate" of an already-tracked Software Engineer Node.js Sênior row, because the incoming score (3.9) was lower than the existing row's (4.1):

⏭️  Skip: Mottu — Software Engineer Flutter Sênior (existing #137 4.1 >= new 3.9)
This is exactly the failure mode the baseline-token guard (discriminating.length === 0) was built to prevent — it just doesn't fire because the accent-splitting artifact masquerades as a discriminating token. Any company with multiple concurrent "Sênior"-suffixed openings (or any other accented seniority/role word that splits into a coincidentally-shared fragment) is at risk of one role silently swallowing another's tracker row.

Suggested fix
Normalize accents before stripping non-ASCII characters in roleTokens(), e.g.:

text
  .toLowerCase()
  .normalize('NFD').replace(/[̀-ͯ]/g, '')  // strip diacritics
  .replace(/[^a-z0-9\s]/g, ' ')
  ...
This would fold "sênior" -> "senior" before the stopword filter, so it's correctly dropped like the unaccented form already is. Same fix likely needed anywhere else in the codebase that tokenizes role/company text without diacritic folding (worth grepping for other .replace(/[^a-z0-9\s]/g call sites).

Workaround used
Manually inserted the tracker row instead of relying on merge-tracker.mjs for this one addition; verified with node verify-pipeline.mjs afterward (clean, aside from pre-existing unrelated warnings).

### CLI tool

Claude Code

### OS and Node.js version

MacOS and node 22.13.0

### Output of npm run doctor

```shell
> career-ops@1.22.0 doctor
> node doctor.mjs


career-ops doctor
================

✓ Node.js >= 22.5 (v22.13.0)
✓ Dependencies installed
⚠ Playwright MCP tools not detected
  → Browser-driven JD fetching and liveness checks (scan / pipeline / apply) need the
  → Playwright MCP server. No project-level MCP config was detected in `.mcp.json`
  → or `.claude/settings*.json`, so SPA job boards may return empty or stale content.
  → Tracking: https://github.com/santifer/career-ops/issues/506
✓ cv.md found
✓ config/profile.yml found
✓ modes/_profile.md found
✓ portals.yml found
✓ Fonts directory ready
✓ data/ directory ready
✓ data/pipeline.md ready
✓ output/ directory ready
✓ reports/ directory ready

Result: All checks passed (1 warning — see above). You're ready to go! Run `claude` (or `opencode`) to start.
```

maxmilian:
## What does this PR do?

Folds accented Latin letters onto their ASCII base before `role-matcher.mjs` tokenizes a role title, so an accent stops behaving as a word separator.

`roleTokens()` and `extractSeniorities()` both ran `.replace(/[^a-z0-9\s]/g, ' ')` on the lowercased title. Since that class does not cover accented letters, `"Sênior"` split into `"s"` (dropped by the length filter) and `"nior"` — a phantom token that appears in no stopword list and is therefore treated as real content.

Reproduced on `main` before the change:

```js
roleTokens('Software Engineer Node.js Sênior');   // ['software','engineer','node','nior']
roleTokens('Software Engineer Flutter Sênior');   // ['software','engineer','flutter','nior']
roleFuzzyMatch(...)                                // true   ← should be false
```

The plain-ASCII control returns `false` correctly, which isolates the accent as the cause.

### Three failures from one root cause

While confirming the report I found the phantom token defeats three separate rules, not just the one in the issue:

1. **False positive (the reported case).** `"nior"` counts as a discriminating overlap, so two unrelated roles at one company cross the 0.6 Jaccard threshold and the `discriminating.length === 0` baseline guard never fires.
2. **Opposite seniorities merging.** `"Sênior"` and `"Júnior"` *both* reduce to `"nior"`. `Engenheiro de Dados Sênior` and `Engenheiro de Dados Júnior` matched on `main` — and because `extractSeniorities()` shares the same stripping, it saw no seniority token on either side, so the seniority-disagreement gate stayed silent instead of catching it.
3. **False negative.** A genuine repost written once with the accent and once without tokenized differently and never matched: `Engenheiro de Software Sênior, Pagamentos` vs `… Senior, Pagamentos` returned `false` on `main`.

### The change

A small `normalizeTitle()` helper lowercases, applies NFD, and drops combining marks; both `roleTokens()` and `extractSeniorities()` call it before the existing ASCII filter. No rule, threshold, or vocabulary list changes — accented spellings simply land in the same vocabulary as plain ones.

The strip uses `\p{Mn}` (nonspacing mark) rather than the more obvious `\p{Diacritic}`. `\p{Diacritic}` also matches standalone characters — `·`, `^`, `` ` ``, `¨`, `´`, `¸` — which are *separators* in a title, not marks on a letter. Deleting them would glue neighbouring words together: `Backend Engineer·Payments` tokenizes to `['backend', 'engineerpayments']`, turning a real repost into a duplicate row. `\p{Mn}` matches none of those, and I verified against every decomposing codepoint in U+0041–U+24FF that it covers everything NFD emits for Latin. There is a regression test for this.

Scope note: NFD folding covers combining-mark accents (`ê é ã ç ñ ü`), which is what the tracker actually sees. Letters with no canonical decomposition (`ø`, `ł`, `ß`, `đ`) still strip as before and can still produce a phantom token — the docstring says so explicitly. I left that alone rather than start a transliteration table for cases that have not come up; happy to follow up if you would rather cover them now.

Two other `[^a-z0-9\s]` tokenizers exist — `normalizeCompanyName` (`invite-match.mjs`) and `deriveSlugCandidates` (`verify-portals.mjs`). I deliberately left them out of scope: both compare company names, which fold consistently on both sides, so the false-positive risk is much lower than for role titles. Worth a separate issue if you want them aligned.

## Related issue

Fixes #2207

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Tests

Added to the existing shared-role-matcher section in `test-all.mjs`:

- the reported pair (`Node.js Sênior` vs `Flutter Sênior`) — `true` on `main`, now `false`
- the `Sênior`/`Júnior` collision — `true` on `main`, now `false`
- the accented-vs-unaccented repost — `false` on `main`, now `true`
- a tokenizer-level assertion that the accented title produces **exactly** the plain-ASCII token list (asserting the whole array, not merely the absence of `"nior"`, so a fix that just deleted non-ASCII would still fail it)
- an over-merge control: two distinct accented specialty suffixes must stay distinct
- a separator control: `Backend Engineer·Payments` must still split into two tokens

`node test-all.mjs` → **2079 passed, 0 failed, 1 warning** (the warning is the pre-existing `cv-sync-check.mjs` no-user-data notice).

## Checklist

- [x] I have read the CONTRIBUTING guide
- [x] My PR is linked to a related issue
- [x] My changes contain no personal data
- [x] I ran `node test-all.mjs` locally

Credit to @pedrobrasileiro for the report — the root-cause trace in the issue was precise down to the line, and the production `merge-tracker` skip it caused is what made the impact concrete.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved role matching for titles with accented characters by treating accented and unaccented spellings consistently.
  * Fixed cases where accent-related tokenization could wrongly merge or incorrectly keep roles distinct.
  * Preserved correct distinctions between seniority levels and specialty suffixes.
  * Improved repost detection across accented and unaccented title variations.
* **Tests**
  * Added coverage to verify accent folding behavior and prevent related regression scenarios in role matching.

### DIFF
--- a/role-matcher.mjs
+++ b/role-matcher.mjs
@@ -81,6 +81,33 @@ export const BASELINE_TOKENS = new Set([
 ]);
 
 /**
+ * Lowercase a title and fold accented Latin letters onto their ASCII base.
+ *
+ * Every rule below matches against ASCII vocabulary, so an accent used to act
+ * as a word separator rather than a letter: "Sênior" became ["s", "nior"],
+ * leaving a phantom "nior" token that no stopword list covers. Folding first
+ * keeps the accented spelling in the same vocabulary as the plain one.
+ *
+ * Only accents that NFD decomposes are folded. Letters with no canonical
+ * decomposition (ø, ł, ß, đ) still reach the ASCII filter unchanged, exactly
+ * as before.
+ *
+ * `\p{Mn}` (nonspacing mark), not `\p{Diacritic}`: the latter also matches
+ * standalone characters such as "·", "^" and "`", which are separators in a
+ * title. Deleting those would glue neighbouring words into one token.
+ *
+ * @param {unknown} value - Raw title, possibly not a string.
+ * @returns {string} Lowercased, accent-folded text.
+ */
+function normalizeTitle(value) {
+  const text = typeof value === 'string' ? value : String(value ?? '');
+  return text
+    .toLowerCase()
+    .normalize('NFD')
+    .replace(/\p{Mn}/gu, '');
+}
+
+/**
  * Convert a role title into content tokens used for fuzzy matching.
  *
  * The tokenizer keeps long descriptive words and a narrow set of short
@@ -92,9 +119,7 @@ export const BASELINE_TOKENS = new Set([
  * @returns {string[]} Ordered role-title tokens.
  */
 export function roleTokens(role) {
-  const text = typeof role === 'string' ? role : String(role ?? '');
-  return text
-    .toLowerCase()
+  return normalizeTitle(role)
     // Replace with a generic baseline token, not empty space: a bare "Member
     // of Technical Staff" (no suffix) or a one-word-suffix MTS title (e.g.
     // "...Staff, Backend") would otherwise tokenize to 0 or 1 words, and
@@ -109,10 +134,8 @@ export function roleTokens(role) {
 }
 
 function extractSeniorities(title) {
-  const text = typeof title === 'string' ? title : String(title ?? '');
   return new Set(
-    text
-      .toLowerCase()
+    normalizeTitle(title)
       .replace(/[^a-z0-9\s]/g, ' ')
       .split(/\s+/)
       .filter(w => SENIORITY_TOKENS.has(w))
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -5724,7 +5724,7 @@ try {
 // application states from fuzzy-only deletion.
 console.log('\n🧪 Testing shared role matcher and dedup-tracker safety...');
 try {
-  const { roleFuzzyMatch } = await import(pathToFileURL(join(ROOT, 'role-matcher.mjs')).href);
+  const { roleFuzzyMatch, roleTokens } = await import(pathToFileURL(join(ROOT, 'role-matcher.mjs')).href);
 
   if (!roleFuzzyMatch('Full Stack Engineer, Foundation', 'Full Stack Engineer, Guarded Releases')) {
     pass('role matcher keeps Full Stack Engineer sibling teams distinct (#947)');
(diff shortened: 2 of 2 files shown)


santifer:
This is on me, @maxmilian: I merged #2164 ('product' as a baseline token) and #1933 (stripping the "Member of Technical Staff" prefix) tonight, and both touch `role-matcher.mjs`. Your branch now conflicts as a result of my merge order, not of anything wrong here.

I checked before pinging rather than trusting GitHub's status: I fetched your head, merged current `main` into it locally, and it conflicts for real.

Accent folding is orthogonal to both of those, so the rebase should be mechanical: they added a token to `BASELINE_TOKENS` and a phrase-level substitution, while you're normalizing before tokenizing. Ping me when it's green and I'll take it straight away: "José" and "Jose" being different roles is the kind of bug that makes people quietly distrust the whole tracker.

abankar1:
Heads-up, @maxmilian — this now conflicts with `main`: my #2107 merged earlier today and it also edits `extractSeniorities`/`roleTokens` and the same role-matcher test block in `test-all.mjs`, so a rebase is needed (sorry, that one's on me). Purely textual — no logic clash.

The fix itself reads well: folding NFD + `\p{Mn}` (and the note on why not `\p{Diacritic}`, to avoid eating `·`/`^`/``` separators) is the right call. Worth calling out that it *composes nicely* with #2107's new sub-baseline gate: that gate keys off `extractSeniorities`, so once you fold, an accented sub-baseline qualifier like "Júnior" / "Sênior" is finally recognized as a seniority token and the gate engages instead of falling through to word-overlap — a latent bug your normalization also closes. Might be worth a fixture for the accented-sub-baseline case after the rebase.

maxmilian:
Rebased onto `main` (`09228cc`), now at `12d63d3` — green and conflict-free.

@santifer, @abankar1: both conflicts were textual exactly as you described. In `role-matcher.mjs` the only overlap was the entry point of `roleTokens` — #1933's `MTS_PREFIX` substitution and my `normalizeTitle()` call sit on the same first line of the chain, so the resolution is to fold before stripping the phrase (order is irrelevant here: "member of technical staff" carries no accents, and folding first only widens what the phrase matcher can see). In `test-all.mjs` both sides appended to the same role-matcher block; kept both. No logic changed on either side. **2261 assertions pass, 0 fail.**

@abankar1 — I added the accented sub-baseline fixture you asked for, and measuring it turned up something slightly different from what we both expected, so it's worth writing down.

I ran the case against `upstream/main`'s `role-matcher.mjs` and against the fixed one:

| title vs `Product Manager, Marketplace` | before | after |
|---|---|---|
| `Júnior Product Manager, Marketplace` | no match | no match |
| `Sênior Product Manager, Marketplace` | **no match** | match |

The sub-baseline gate wasn't falling through to word-overlap — it was never reached at all. `extractSeniorities` saw no seniority token on either side, so the *specialization-marker* rule fired instead: the phantom `nior` survives as a non-baseline token on the qualified side only, making that side a strict superset with an extra discriminating word. So `Júnior` got the right answer for the wrong reason, and `Sênior` — which should match, since it's routinely added or dropped between reposts of one req — was a plain false negative.

Folding is what lets #2107's gate see the qualifier at all, and the two cases then separate on their actual meaning rather than on a tokenizer artifact. Both directions are pinned as fixtures with that reasoning in a comment, so nobody later reads the `Júnior` case as evidence the gate was working.

santifer:
Merged, @maxmilian. "Sênior" tokenizing to ["s", "nior"] is a lovely bug: the accent was acting as a word separator instead of a letter, and the phantom "nior" token isn't in any stopword list, so it just sat there being matched against nothing.

Two details I want to name because they're the difference between a fix and a regression. `\p{Mn}` rather than `\p{Diacritic}`: the latter also matches standalone "·", "^" and backtick, which are separators in a real job title, so deleting them would have glued neighbouring words into one token. And documenting that ø, ł, ß and đ have no canonical NFD decomposition and therefore still reach the ASCII filter unchanged, exactly as before, so the change is strictly additive rather than a quiet behaviour swap for Nordic and Polish titles.

That matters here more than in most repos: this is a job search tool used across eighteen languages, and the people most affected by a title tokenizer that chokes on accents are the ones searching in Portuguese, Spanish and French.

PR #2212, opened 2026-07-27
fix(liveness): treat Phenom/SPA "job ... has been filled" as expired

### LINKED ISSUE #2211: check-liveness classifies filled Phenom/SPA postings (e.g. careers.icf.com) as active
### What happens
`node check-liveness.mjs <url>` classifies a **filled** requisition on a Phenom-hosted career site as `active`.

Reproduced on `careers.icf.com` (Phenom): the page returns **HTTP 200** with a generic **Apply** control in the shell, while the true status is rendered into the main content:

> We're sorry… the job you are trying to apply for has been filled.

A browser render shows the "filled" notice; `check-liveness` returns `active`.

### Root cause
`HARD_EXPIRED_PATTERNS` in `liveness-core.mjs` includes `/position has been filled/`, which does **not** match the Phenom phrasing *"the job you are trying to apply for has been filled."* With no hard-expired signal matched, the generic apply control wins (`apply_control_visible` → `active`).

### Impact
Filled/dead postings on any Phenom-hosted `careers.<company>.com` site (a large share of enterprise ATSs) can be classified `active`, so `scan --verify` keeps them in the pipeline and users spend effort on dead links. This is the "expired signals must win over generic Apply text" invariant leaking on SPA ATSs.

### Fix
Generalize the pattern to require any job noun within 60 chars of "has been filled", with a negative lookahead rejecting "has been filled **out**" (a candidate completing a form, not a closed req). PR with a unit test to follow.

LayerC0de:
Fixes #2211.

## Problem
Phenom-hosted career pages (e.g. `careers.icf.com`) return **HTTP 200** on a filled req with a generic **Apply** control in the shell, while the real status — *"the job you are trying to apply for has been filled"* — is rendered into the main content. `HARD_EXPIRED_PATTERNS` only had `/position has been filled/`, which misses that phrasing, so `apply_control_visible` won and the dead posting was classified `active`. This breaks the "expired signals win over generic Apply text" invariant on SPA ATSs.

## Change
- **`liveness-core.mjs`** — replace `/position has been filled/` with:
  ```js
  /\b(?:job|jobs|position|role|posting|opening|vacancy|requisition|req|listing)\b[\s\S]{0,60}?\bhas been filled\b(?!\s+out)/i
  ```
  A job noun within 60 chars of "has been filled"; the negative lookahead rejects "has been filled **out**" (form completion, not a closed req).
- **`tests/liveness-core.test.mjs`** (new) — Phenom/ICF phrasing → `expired`; the original "position has been filled" still → `expired`; an active posting with an apply control stays `active`; and "application form has been filled out" is **not** a false positive.

## Testing
```
node test-all.mjs --only liveness
# 4 passed, 0 failed
```

## Notes / possible follow-up
This fixes the pattern gap, which resolves the reported case once the page content is read. A deeper hardening — treating a bare apply-control on a known SPA/Phenom shell as `uncertain` (force a render) rather than `active` — could be a follow-up, but is more invasive; this PR is the minimal, well-tested fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved “filled” expiration detection to cover more common job-status wording variants.
  * Prevented false positives from application-form phrasing like “has been filled out” (and similar text) being misclassified as expired.
  * Added regression tests to confirm correct classification for both expired and active postings, including key guard cases.

### DIFF
--- a/liveness-core.mjs
+++ b/liveness-core.mjs
@@ -5,9 +5,11 @@ const HARD_EXPIRED_PATTERNS = [
   // phrasing SPA ATSs (Phenom, e.g. careers.icf.com) inject on a filled req —
   // "the job you are trying to apply for has been filled" — so those pages
   // returned HTTP 200 with a generic Apply control and were classified active.
-  // Requires a job noun within 60 chars, and the negative lookahead rejects
-  // "has been filled out" (a candidate completing a form, not a closed req).
-  /\b(?:job|jobs|position|role|posting|opening|vacancy|requisition|req|listing)\b[\s\S]{0,60}?\bhas been filled\b(?!\s+out)/i,
+  // A job noun within 60 chars, then "has been filled" — but NOT when the thing
+  // filled is an application/form (the lookbehind) or "filled out" (the
+  // lookahead). Both guards avoid the worse error: reading a LIVE posting whose
+  // copy says "once the application form has been filled…" as expired.
+  /\b(?:job|jobs|position|role|posting|opening|vacancy|requisition|req|listing)\b[\s\S]{0,60}?(?<!\b(?:application|form)\s)has been filled\b(?!\s+out)/i,
   /this job has expired/i,
   /job posting has expired/i,
   /no longer accepting applications/i,
--- a/liveness-core.mjs
+++ b/liveness-core.mjs
@@ -1,7 +1,13 @@
 const HARD_EXPIRED_PATTERNS = [
   /job (is )?no longer available/i,
   /job.*no longer open/i,
-  /position has been filled/i,
+  // Generalized "filled" signal. The old /position has been filled/ missed the
+  // phrasing SPA ATSs (Phenom, e.g. careers.icf.com) inject on a filled req —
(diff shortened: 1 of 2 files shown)


LayerC0de:
Thanks @coderabbitai — good catch. Addressed in a7faa28: added a negative lookbehind so "application/form … has been filled" no longer satisfies the filled-job pattern (only a genuine req does), plus a regression test for the no-`out` variant. `node test-all.mjs --only liveness` → 5/5.

santifer:
Merged, @LayerC0de. This is the failure mode that costs a candidate the most: a Phenom page returns HTTP 200 with a working Apply button on a req that's already filled, so the tool says the job is live and someone spends an evening tailoring a CV for it.

The two guards are what make the regex safe rather than clever. Excluding "once the application form has been filled" and "filled out" avoids the strictly worse error of marking a live posting expired, and you named that tradeoff explicitly in the comment instead of leaving the next reader to work out why the lookbehind is there. Verified merged against current `main`: 2114 assertions, 0 failures. 🚀

PR #2216, opened 2026-07-27
fix(web): resolve 4 HIGH dependabot alerts (nested postcss + sharp)

santifer:
Closes the 4 remaining dependabot alerts in `web/` — the ones no automated bump could reach, because neither is a direct dependency edge.

## The two problems

**1. Nested postcss (3 alerts, 2 HIGH).** The root install was already patched (`8.5.18`), but Next shipped its *own* nested copy at `8.4.31`:

```
node_modules/postcss                  = 8.5.18  ✓
node_modules/next/node_modules/postcss = 8.4.31  ✗   ← GHSA-r28c-9q8g-f849, GHSA-6g55-p6wh-862q, GHSA-qx2v-qp2m-jg93
```

An `overrides` entry collapses the tree to a **single patched instance (8.5.23)**; the nested copy is gone from the lockfile. The direct devDependency moves `^8.5.9 → ^8.5.18` so the declared floor and the override agree (npm errors `EOVERRIDE` otherwise).

**2. sharp 0.34.5 → 0.35.3 (1 HIGH,** GHSA-f88m-g3jw-g9cj**).** It arrives as an `optionalDependency` of next (`^0.34.5`), so `0.35.x` is outside next's declared range and only an override reaches it.

**Honest risk note:** this app has **no `next/image` usage and no `images` config**, so sharp is never on a code path we execute. The bump is hygiene and a clean alert board, not a live exploit fix — worth knowing when weighing the override against next's declared range.

## Verification (beyond CI)

A postcss override is exactly the change a typecheck **cannot** vouch for — it's the Tailwind v4 CSS pipeline. So I drove it:

- Tailwind still emits real CSS: **91KB, 117 `brand` rules**, `--color-brand-foreground` and the `.co-cost` rules all present
- All six routes (`/`, `/pipeline`, `/explore`, `/config`, `/portals`, `/cv`) return **200** against real data; dev-server log clean
- `npm audit` → **0 vulnerabilities**; nested postcss instance **absent** from the lockfile
- `typecheck` + `next build` green

Scope: `web/` only (`package.json` + lockfile), no source changes.

## Summary by CodeRabbit

* **Chores**
  * Updated internal package versions to improve build tooling compatibility.
  * Standardized transitive package resolution for more consistent installations.

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -27,7 +27,7 @@
         "@types/node": "^25.6.0",
         "@types/react": "^19.2.14",
         "@types/react-dom": "^19.2.3",
-        "postcss": "^8.5.9",
+        "postcss": "^8.5.18",
         "tailwindcss": "^4.2.2",
         "typescript": "^6.0.2"
       }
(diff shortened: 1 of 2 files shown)


santifer:
Merged. Verified before the button: all checks green including `web typecheck + build`, and the nested `next/node_modules/postcss` instance is gone from the lockfile.

Issue #1724, opened 2026-07-09
Umbrella: the learning loop — make career-ops sharper the more you use it


santifer:
Tracking the features that turn career-ops from a one-shot tool into one that **learns from your search over time**. Individually small; together they're the retention layer — the reason you keep the tool open across a whole search, not just for one application.

- [x] #1520 — `/upskill`: **COMPLETE** — phase 1 (#1573) · phase 2a targeted URL (#1739, @ritikap1318) · phase 2b learning plan (#1740 via #2217, @Scott-Emberson)
- [x] #1521 — `/expand`: **SHIPPED** (#1533): discovers competencies from your GitHub/portfolio, additive-only with confirm-gate
- [x] #1722 — `/outcome`: **Phase 1 SHIPPED** (#2248, @shafiankhan): outcome recording + per-application artifact archiving under data/outcomes/
- [x] #1723 — multi-source profile intake: **SHIPPED** (#1843, @maxmilian): documents/ folder + intake mode
- [x] `/calibrate` — **SHIPPED** (#3316): your recorded outcomes checked against your evaluation scores, so you learn whether the scoring predicts success *for you* (#3315; strictly advisory, never rewrites scoring rules)

**Common thread:** a richer profile in → honest feedback captured → the tool getting more useful each run. All local-first and human-in-the-loop. Nothing here submits, auto-decides, or leaves your machine.

Several of these were sharpened by studying [MadsLorentzen/ai-job-search](https://github.com/MadsLorentzen/ai-job-search) — credit where due; good ideas travel.

Want to pick one up? Comment on the specific issue, not here.

Scott-Emberson:
I can tackle this.

Scott-Emberson:
Quick pointer so this thread stays an index (per the "comment on the specific issue" rule): to move the cluster forward I've posted on the sub-issues —

- #1722: phase-1 design sketch for the /outcome recorder (schema, archive layout, no dependency on the open ledger PR) + two questions for you before anyone claims.
- #1723: a spec question on how PDF text extraction should work under the zero-new-deps constraint — worth settling before someone starts.
- #1520: phase-2 design comment (learning plan + targeted mode) is up for air.

Discussion on each belongs over there.

Schlaflied:
[...]

Issue #1520, opened 2026-07-04
feat: /upskill — aggregate skill-gap analysis + learning plan from your tracked pipeline


santifer:
## The idea

After 50 evaluated offers, your tracker holds 50 verdicts — and no aggregate reading. The information is already there: every low-scoring evaluation names the skills you were missing. `/upskill` turns that discard history into an answer to the question every job seeker asks themselves: **what should I learn, in what order?**

Pattern credit: [MadsLorentzen/ai-job-search](https://github.com/MadsLorentzen/ai-job-search)'s `/upskill` — this spec adapts it to career-ops' tracker and A–F scoring model.

## Spec

**Two modes:**
- `/career-ops upskill` — **aggregate**: reads `data/applications.md` (all rows with a report). For each report, extract required/missing skills from the evaluation. **Weight by inverse score**: a 2.1/5 offer contributes more to the gap map than a 4.5/5 one — low-fit offers are where your gaps live.
- `/career-ops upskill <URL>` — **targeted**: one posting, unweighted.

**Two passes:**
1. **Hard-skill diff** — extract skills from reports/posting, remove anything already in `cv.md`/`config/profile.yml` (be generous with synonyms), build a frequency×weight map.
2. **LLM synthesis** — gaps the keyword diff can't see: domain knowledge, soft skills, tooling/process, certifications. Tag each `[domain]`/`[soft]`/`[tooling]`/`[credential]`. No duplicates from pass 1.

**Output** (user layer — `data/upskill/report-YYYY-MM-DD.md`):
- Prioritized gap heatmap (Critical / High / Medium / Low) with gap source ("4/9 low-fit offers" / "LLM synthesis")
- Learning plan: for the top gaps, web-searched free-first resources + rough time estimates + suggested order
- **Diff vs the previous report** if one exists ("since last month: Kubernetes gap closed, CI/CD still open")

## Constraints
- Local-first, zero new dependencies, works on cheap models (the synthesis pass must degrade gracefully)
- Output files are **user layer** (never touched by the updater) — DATA_CONTRACT registration required
- Prompt-first: `modes/upskill.md` + optionally a small deterministic helper for the tracker parsing (mirror `analyze-patterns.mjs`)

## Acceptance
- Aggregate mode produces a heatmap + plan from a tracker with ≥5 scored rows; targeted mode works from a pasted URL
- A skill present in cv.md never appears as a gap
[...]

PR #2217, opened 2026-07-27
feat(upskill): learning plan on the aggregate gap heatmap (#1740)

### LINKED ISSUE #1740: feat(upskill): learning plan on top of the gap heatmap (#1520 phase 2b)
Part of #1520 (phase 2b) · Umbrella #1724 · Builds on the phase-2 design aired by @Scott-Emberson in #1520

## The slice

Phase 1 (#1573) produces the tiered gap heatmap. This slice adds the **learning plan** on top:

- For the top gaps (Critical/High), web-search **free-first** resources + rough time estimates + a suggested order.
- Prepend the plan to the phase-1 report (`data/upskill/report-YYYY-MM-DD.md`) — additive section, same file.
- Grounding rule (from phase 1's "no fabricated resources"): every resource must come from an actual search result — never invented from memory. On cheap models or if search is unavailable, skip the plan section and say so.

## Design authority

The design lives in Scott's #1520 phase-2 comment (why it isn't #160-again, the funnel-forward framing). Implementation should follow it; deviations get discussed in #1520 first.

Scott — if you want to build this half, it's yours; say the word here.

Scott-Emberson:
Closes #1740
Part of #1520 (phase 2b) · Umbrella #1724

## What this adds

A **web-searched learning plan** layered on top of the phase-1 aggregate gap heatmap. The `upskill` mode now generates a `## Learning Plan` section on the aggregate report (`data/upskill/report-{date}.md`), placed just below `## Suggested Order` — Suggested Order sequences the gaps, the plan then resources each one.

Per eligible gap (every **Critical** and **High**; **Medium** too only when the total distinct gap count is `< 5`): 2–3 **free-first** resources (name + URL + one-line why), a **study direction** anchored to a strength already on the CV, and an **effort bucket** (`~hours`/`~days`/`~weeks`) taken only from the resource's own stated length. Study order puts dependencies first and quick wins early.

It reuses the **phase-1 gap objects** (tiers read straight from the existing JSON — never re-derived) and the **check-liveness pattern** (`check-liveness.mjs` / `liveness-core.mjs`) for write-time URL liveness.

## Trust model encoded + frozen in CI

Each of the eight promises is stated as an explicit agent rule in `modes/upskill.md` and asserted in `test-all.mjs` (one grep-based assertion per rule) so a future edit can't silently drop a guarantee:

1. **Search-result-or-nothing (grounding)** — every resource comes from a real web-search result; on cheap models or when WebSearch is unavailable/weak, the plan section is **skipped and says so**.
2. **Deterministic degradation** — when search is skipped/weak, the heatmap + Suggested Order still ship without resources; the plan is purely additive.
3. **Ephemeral / non-versioned resources** — regenerated fresh every run, never diffed or revalidated across runs; only gap tiers are stable. One-line disclaimer on the artifact.
4. **Write-time URL liveness** — every cited URL is liveness-checked at generation via the check-liveness pattern; dead links never enter the report; staleness disclaimer included.
5. **Hard search budget** — max 2 searches per gap, ~12 per aggregate run; current year included in queries.
6. **Free-first with explicit failure** — if no free option is found, the plan says so; never silently substitutes a paid resource.
7. **Effort from stated length only** — never invented.
8. **Scope boundary** — entries link to `/career-ops training {name}` for judging a specific resource; the plan never runs training's 6-dimension scoring. `upskill` finds; `training` judges.

## No script change

`upskill.mjs` is untouched: `node upskill.mjs` already tags each gap with its `tier` and the total distinct gap count is the `gaps` array length, so the agent reads plan-eligible gaps directly. Searching stays the agent's job in the mode — no web-search logic added to the node script.

## Validation

- `node upskill.mjs --self-test` → OK
- `node test-all.mjs` → **2104 passed, 0 failed** (2 pre-existing env warnings). The 9 new upskill assertions (8 trust rules + the "phase-2 placeholder is gone" check) are counted and passing.

**Design authority:** #1520 phase-2 comment; no deviations.

## Summary by CodeRabbit

* **Documentation**
  * Updated the upskill workflow to include Phase 2b “web-searched, free-first” learning plans for eligible skill gaps, layered on top of the aggregate gap map.
  * Added a detailed Learning Plan template (suggested order) plus clear skip/failure behavior when live search is unavailable or weak.
  * Documented the trust model rules (grounding, budget, URL liveness checks, effort sourcing, and scope boundaries), and clarified schema-version diff behavior.

* **Tests**
  * Added contract validation to ensure the Learning Plan section and the eight trust-model rule phrases remain present and unchanged.


santifer:
Merged, @Scott-Emberson. Closes the phase-2 half of #1520 that got split out into #1740. A gap heatmap tells someone what they're missing; a learning plan is the part that makes the number actionable instead of discouraging, which matters a lot when the person reading it is out of work. Verified merged into current `main`: 2234 assertions, 0 failures. 🚀

PR #2084, opened 2026-07-21
feat(providers): add VDAB zero-auth provider

nikolaysm:
## What does this PR do?

Adds `providers/vdab.mjs`, a zero-auth scanner provider for VDAB (Flanders' Belgian public employment service), following the same pattern as `providers/arbeitsagentur.mjs` — config-driven, in-process HTTP, zero browser, zero LLM tokens.

## Related issue

Part of #230 (scanner-providers umbrella). No dedicated issue — new zero-auth scanner provider (CONTRIBUTING exempts providers from issue-first).

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation / translation
- [ ] Refactor (no behavior change)

## How it works

VDAB's search API (`POST /rest/vindeenjob/v4/vacatureLight/zoek`) works cold over plain HTTP — no browser needed — using a static, build-time key baked into VDAB's own frontend bundle (verified live; same trust tier as `arbeitsagentur.mjs`'s public `API_KEY`). Keyword-driven, recall-first: over-fetches per keyword and lets `scan.mjs`'s `title_filter`/`location_filter` do the precision cut.

A few things beyond the minimal provider:
- **Self-heals** if VDAB rotates the key on a redeploy — a 403 triggers one re-derivation from the live bundle before giving up, so a rotation doesn't require a code patch to recover.
- **Keyword fallback**: falls back to `config/profile.yml`'s `target_roles` when a `portals.yml` entry has no `keywords` of its own, via a new shared `providers/_profile-keywords.mjs` helper (honors the `CAREER_OPS_PROFILE` env override, matching the rest of the codebase). Currently used by `vdab.mjs` only; other keyword-required providers can opt in later.
- **Optional `fetchDetails`**: bounded, batched per-job calls to enrich results with full descriptions — opt-in, off by default, so the provider stays zero-token unless a user explicitly asks for it.
- **Cooperates with `verify-portals.mjs`'s bounded health-check probe**: caps pagination at `ctx.maxPages`, skips detail enrichment while probing, and lets a probe's own error propagate instead of flattening it into a generic summary error (without this, a keyword list wider than the probe's request budget got misreported as a dead board).

## Tests & docs

[...]

PR #2223, opened 2026-07-27
test: assert the Dockerfile's Playwright pin matches package.json

### LINKED ISSUE #2221: test: assert the Dockerfile's Playwright pin matches package.json
**Time:** ~20 min · **Difficulty:** Beginner · **Code** (one assertion in `test-all.mjs`)

The `Dockerfile` pins Playwright to a specific version so the npm install matches the Chromium that ships in the base image. Right now those two have silently drifted apart:

```bash
grep -n '"playwright"' package.json      # 1.62.0
grep -n 'playwright@' Dockerfile         # 1.58.1
```

The Dockerfile even carries a comment saying it *"matches playwright@… in package.json"*, which stopped being true some time ago. Nothing catches this, so every time the dependency is bumped the pin falls further behind, and the failure only shows up for someone running the container: their browser build and their `playwright` package disagree.

### What to do

Add an assertion to `test-all.mjs` that reads the pinned version out of the `Dockerfile` and compares it against `dependencies.playwright` in `package.json`, failing when they differ. Both the `RUN npm install … --save-exact playwright@X` line and the comment on line 2 carry the version: checking the `RUN` line is what matters, and flagging the comment too would be a bonus.

Then fix the current drift by bringing the `Dockerfile` up to whatever `package.json` says.

### Where to look

`test-all.mjs` has plenty of examples of this shape: a check that reads two files and asserts they agree. Look for the existing assertion that keeps `.claude-plugin/plugin.json` and `.github/plugin/plugin.json` in sync — same idea, same structure.

### Acceptance

- [ ] A failing assertion when the `Dockerfile` pin and `package.json` disagree
- [ ] The current drift fixed so the suite is green
- [ ] `node test-all.mjs` passes

### Why it's worth doing

It's small, but it's the kind of check that keeps working forever without anyone remembering it. Found while reviewing #2084, where a contributor spotted the drift and corrected it to a version that was *already* out of date again by the time they pushed. That's the tell that this needs a machine, not attention.

nikolaysm:
## Summary
- Adds an assertion to `test-all.mjs` (same shape as the existing `.claude-plugin/plugin.json` mirror check) that reads the Playwright version out of the Dockerfile's `--save-exact playwright@X` RUN line and its line-2 comment, and fails when either disagrees with `package.json`'s `dependencies.playwright`.
- Fixes the current drift: the Dockerfile's base image tag, RUN pin, and comment were all still at `1.58.1`/`1.61.1` while `package.json` is at `1.62.0`. Brought all three back in sync at `1.62.0` (confirmed the `v1.62.0-jammy` base image tag exists on `mcr.microsoft.com`).

## Test plan
- [x] `node test-all.mjs` passes (2165 passed, 0 failed)
- [x] Verified the new assertion actually fails when the pin disagrees (checked before fixing the Dockerfile)

Fixes #2221.

## Summary by CodeRabbit

* **Bug Fixes**
  * Upgraded Playwright to version 1.62.0 to improve browser automation compatibility.

* **Tests**
  * Added a CI guard that verifies Playwright version consistency across package configuration and container setup, including checks for the base image tag and the installed Playwright pin.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
 # career-ops container
-# Base: Playwright image with Chromium preinstalled (matches playwright@1.58.1 in package.json).
+# Base: Playwright image with Chromium preinstalled (matches playwright@1.62.0 in package.json).
 # Host kernels that block Playwright's chromium installer (e.g. Ubuntu 26.04) work fine here
 # because the browser ships in the image and runs under the image's userland.
 
-FROM mcr.microsoft.com/playwright:v1.61.1-jammy
+FROM mcr.microsoft.com/playwright:v1.62.0-jammy
 
 ENV DEBIAN_FRONTEND=noninteractive \
     NODE_ENV=development \
@@ -35,7 +35,7 @@ WORKDIR /app
 # Pin playwright to the version that matches the base image's bundled chromium.
 COPY package.json package-lock.json* ./
 RUN npm install --no-audit --no-fund \
- && npm install --no-audit --no-fund --save-exact playwright@1.58.1
+ && npm install --no-audit --no-fund --save-exact playwright@1.62.0
 
 # The rest of the project is bind-mounted at runtime via docker compose,
 # so we don't COPY sources here — keeps the image generic and lets local
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -943,29 +943,40 @@ for (const f of skillEntrypoints) {
   }
 }
 
-// The Dockerfile pins playwright to the exact version baked into its base
-// image so the npm-installed package matches the bundled Chromium. Nothing
-// enforced that pin against package.json's own playwright version, so it
-// silently drifted for months (#2221) — this keeps the two in sync going
-// forward instead of relying on whoever next reads the Dockerfile noticing.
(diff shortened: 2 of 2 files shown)


nikolaysm:
@santifer this is the follow-up you flagged on #2084 (the Dockerfile Playwright pin drift) — CI's green and CodeRabbit's re-review came back clean, ready whenever you have a chance.

santifer:
You found more than the issue described, @nikolaysm, and that's the part worth pointing at. I wrote #2221 thinking there was one pin to check: the `--save-exact playwright@X` line. There are **three** — the `FROM mcr.microsoft.com/playwright:vX-jammy` base image tag, the npm install, and the claim in the comment on line 2 — and the base image tag was drifted too (v1.61.1 against package.json's 1.62.0), which neither I nor the earlier attempt at this noticed.

Asserting all three is what makes the check actually hold. The npm pin and the image tag disagreeing is the worse failure of the two, because then the container's bundled Chromium and the installed driver are different versions and the breakage looks like a Playwright bug rather than a config one.

Nice loop: you spotted the drift while writing a provider a few hours ago, I turned it into an issue, and you closed it properly. 2166 assertions, 0 failures. 🚀

PR #2003, opened 2026-07-17
feat(skills): extract shared skill-extract.mjs — PR 1 of #1896 (pure relocation)

abankar1:
PR 1 of #1896, built exactly to the sketch you green-lit: the **pure relocation**. `GO_SKILL_PATTERN` stays internal; `canonicalize()` passes unknown tokens through unchanged.

## What
Moves upskill.mjs's skill tokenizer into a shared **`skill-extract.mjs`** (root module, like `tracker-parse.mjs`): `SKILL_TOKENS`, `CANONICAL`, `DISPLAY`, `SKILL_PATTERN`, the internal `GO_SKILL_PATTERN`, `canonicalize()`, `extractSkills()`. Single source of truth so upskill / jd-skill-gap / analyze-patterns can converge (PRs 2–3) instead of drifting — the drift class that shipped #1851.

## Behavior-identical (this is the point)
- `upskill.mjs` imports + **re-exports** `extractSkills` — verified it re-exports the *same function object*, so every existing importer is unaffected.
- `SKILL_TOKENS` now lives in exactly one place (confirmed: no other file defines it). jd-skill-gap / analyze-patterns are untouched — they're PRs 2 and 3.
- The `extractSkills` self-test fixtures moved verbatim from upskill's `runSelfTest` to `tests/skill-extract.test.mjs`, so the same assertions now guard the module (plus a few added: `canonicalize` direct, empty/null input).

## Tests
`node test-all.mjs --quick` → **1806 passed, 0 failed**. upskill's own self-test still green (aggregation/suppression/targeted checks retained).

3 commits: module + SYSTEM_PATHS · upskill imports it · tests.

Next (separate PRs, per the sequence): PR 2 routes `jd-skill-gap` through `canonicalizeSkills`/`isKnownSkill` (the behavior fix — CV `k8s` ≠ JD `Kubernetes`); PR 3 migrates `analyze-patterns`.

## Summary by CodeRabbit

* **New Features**
  * Added a shared skill extraction and canonicalization module for free-text, including alias handling and consistent display casing.
  * Improved recognition for special cases like standalone “Go” and more reliable skill boundary detection.
* **Bug Fixes**
  * Reduced incorrect suppression/confusion between similar skills (e.g., “Java” vs “JavaScript”).
  * Preserve unknown terms instead of normalizing them to unintended canonical forms.
* **Refactor**
  * Updated the upskilling workflow to reuse the shared extraction logic.
* **Tests**
[...]

Issue #1896, opened 2026-07-15
refactor(skills): three parallel skill tokenizers disagree (CV "k8s" ≠ JD "Kubernetes" in jd-skill-gap) — extract the shared module upskill.mjs already promises


abankar1:
## Problem

The repo now has **three independent skill-extraction implementations**, and the drift between two of them already shipped one user-visible correctness bug (#1851, merged fix #1861). The third is drifting today.

| Implementation | Vocabulary | Canonicalization |
|---|---|---|
| `upskill.mjs` 51–150 | `SKILL_TOKENS` (~80 tokens) + `SKILL_PATTERN` + case-sensitive `GO_SKILL_PATTERN` | `CANONICAL` alias map (k8s→Kubernetes, golang→Go, postgres→PostgreSQL, …) |
| `jd-skill-gap.mjs` 36–130 | its own `SKILL_TOKEN_RE` + `STOPWORDS` heuristic | **none** |
| `analyze-patterns.mjs` (tech regex) | its own token list ("mirrors … duplicated by design") | its own handling |

`upskill.mjs:48–50` itself says the duplication is temporary:

> *"deliberately duplicated — … extracting a shared module from a tested core script is a follow-up once both call sites are stable."*

Both call sites are now stable, there are **three** of them, and #1851 demonstrated the failure mode is not hypothetical: a parallel re-implementation of "known-skill suppression" inverted the result on every skill and shipped to everyone on 1.19.

## A concrete live divergence (not cosmetic)

`jd-skill-gap.mjs` has no alias map, and its `skillMentionedInText()` is a word-boundary check against raw CV text. So for a CV that writes `k8s`, `Golang`, or `Postgres` (extremely common CV spellings):

- JD requires **Kubernetes** → `skillMentionedInText('Kubernetes', cv)` → no match → **misreported as a gap** the candidate doesn't have.
- upskill's targeted mode, on the same inputs, correctly suppresses it via `CANONICAL['k8s'] → 'Kubernetes'`.

Same tool suite, same JD, same CV — two different gap reports depending on which script you ask. For a tool whose contract is "keywords get reformulated, never fabricated," disagreeing with itself about what counts as a known skill undermines the core promise (jd-skill-gap's own header says its job is to tailor "honestly").

## Proposal

Extract `skill-extract.mjs` (root module, like `tracker-parse.mjs` / `fingerprint-core.mjs`):

[...]

PR #2226, opened 2026-07-27
fix(jd-skill-gap): canonicalize CV+JD tokens via skill-extract — PR 2 of #1896

abankar1:
PR 2 of the #1896 sequence you green-lit (relocation → **jd-skill-gap** → analyze-patterns). PR 1 (#2003, `skill-extract.mjs`) is now merged, so this builds on it directly. Rebased onto current `main` — clean single-file diff.

## The bug this closes

`classifySkillGaps` compared the JD's tokens against the CV's text **literally**, so a CV that writes an alias and a JD that writes the canonical name never matched — the exact divergence #1896 is about:

- CV Skills section: `Python, k8s, Postgres`
- JD asks for: `Kubernetes`, `PostgreSQL`

→ both `Kubernetes` and `PostgreSQL` were reported as **gaps**, even though the candidate clearly has them. That's a false "you're missing this skill" on a tool candidates use to decide what to learn.

## The fix

Route both sides through the shared module from PR 1: `canonicalize()` folds `k8s → Kubernetes`, `postgres → PostgreSQL`, `golang → Go`, so they resolve to the same name and land in `existing` / `supportedByResume` instead of `gap`.

Per your answer 2 on the issue (*"pass-through unchanged for unknown tokens"*), an unrecognized token canonicalizes to itself, so jd-skill-gap keeps its own word-boundary heuristic for free tokens — behavior for anything the module doesn't know is byte-for-byte unchanged.

## Tests

Two regression fixtures in the existing self-test:
- the reported case — CV `k8s`/`Postgres`/`golang` vs JD `Kubernetes`/`PostgreSQL`/`Go` all classify as satisfied, while a genuinely-absent skill stays a real gap;
- an unknown token (`Fabrikam-SDK`) still matches (or not) via the preserved word-boundary fallback.

`node jd-skill-gap.mjs --self-test` → 35/0 · `node test-all.mjs --quick` → **2252 passed, 0 failed**.

2 commits (fix + fixtures). PR 3 (analyze-patterns → the shared tokenizer) follows once this lands.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved skill-gap matching by recognizing equivalent skill names and aliases, such as “k8s” and “Kubernetes.”
  * More accurately distinguishes skills listed in a CV from those supported by the CV’s other content.
  * Preserves reliable word-boundary matching for unrecognized or free-form terms.

### DIFF
--- a/jd-skill-gap.mjs
+++ b/jd-skill-gap.mjs
@@ -425,6 +425,39 @@ Benefits and Perks (US Only)
   eq('does not extract "Deep" as a skill', dispositionSkills.includes('Deep'), false);
   eq('does not extract "Interest" as a skill', dispositionSkills.includes('Interest'), false);
 
+  // Regression (#1896): the reported bug. A CV alias and a JD's canonical name
+  // must not read as a gap. Before the shared skill-extract canonicalization,
+  // classify compared the two literally — "k8s" in the CV vs "Kubernetes" in
+  // the JD — and reported Kubernetes as a false gap. canonicalize() now folds
+  // both to "Kubernetes" so it resolves to the region the CV actually uses it.
+  const aliasCv = `
+# Skills
+Python, k8s, Postgres
+
+# Experience
+Built data pipelines with golang microservices.
+`;
+  const aliasResult = classifySkillGaps(['Kubernetes', 'PostgreSQL', 'Go', 'Rust'], aliasCv);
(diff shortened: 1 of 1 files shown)


santifer:
Merged, @abankar1. PR 2 of #1896 landing the same day as PR 1: the relocation went in this morning and this is what it was for.

Routing both sides of the comparison through the same canonicalizer is the part that matters. A gap analysis that says you're missing a skill you actually have is worse than one that says nothing, because the user acts on it: they go and learn something they already know, or they filter out a role they were qualified for.

PR #2227, opened 2026-07-28
fix(analyze-patterns): symbol-edge \b bugs drop C++/C#/.NET from tech gaps and 70+ from remote classify

abankar1:
## The bug

`analyze-patterns.mjs`'s tech-mention regex bounded each match with `\b`:

```js
gap.description.match(/\b(...|C\+\+|C#|\.NET|...)\b/gi)
```

A trailing `\b` requires a word char on one side of the boundary, so it **never matches after a symbol edge** — after the `+` in `C++`, the `#` in `C#`, or around the leading `.` in `.NET`. All three silently failed to match in *any* gap description:

```
"needs C++ experience"  -> null
"strong C# background"  -> null
"built on .NET"         -> null
```

So `C++`, `C#`, and `.NET` — three of the most common stacks — never reached the `techStackGaps` rollup or the *"Filter out roles requiring {top gaps} as primary stack"* recommendation, even when they were the actual blocker in negative/self-filtered outcomes. The analysis quietly under-reported them as zero.

## The fix

Switch the boundaries to `(?<!\w)` / `(?!\w)` lookarounds — the same symbol-edge fix `skill-extract.mjs` and `upskill.mjs` already carry — and lift the regex + canonical map into a module-level `extractTechMentions()` so it's unit-testable:

```
"needs C++ experience"  -> ["C++"]
"strong C# background"  -> ["C#"]
"built on .NET"         -> [".NET"]
"Google Cloud"          -> null      (no false positive)
"JavaScripting"         -> null
```

**No vocabulary change** — the same 37-tech scope and case-canonicalization are preserved; this only stops the three symbol-bearing entries from being invisible. (The eventual unification onto `skill-extract.mjs` — #1896's PR 3 — will supersede this inline regex; this is the interim correctness fix so the analysis is right in the meantime.)

## Tests

Regression added to the existing `--self-test`: the three dropped stacks now extract, substrings (`Go` in `Google`, `Java` in `JavaScripting`) don't false-positive, and `node.js` case variants collapse to one bucket.

`node analyze-patterns.mjs --self-test` → OK · `node test-all.mjs --quick` → **2252 passed, 0 failed**.

2 commits (fix + test).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved technology stack detection during gap analysis.
  * Correctly recognizes symbol-edge technologies such as C++, C#, and .NET.
  * Prevents false matches inside unrelated words (e.g., “Go” in “Google,” “Java” in “JavaScripting”).
  * Standardizes technology variants and casing (e.g., “nodejs” → “Node.js”).
  * Enhances “tech stack gaps” and stack mismatch recommendations using more accurate detection results.

### DIFF
--- a/analyze-patterns.mjs
+++ b/analyze-patterns.mjs
@@ -390,6 +390,16 @@ risk_summary:
     failures.push('node.js case variants failed to canonicalize');
   }
 
+  // Remote classifier (regression): the "70+" signal ends in "+", so a
+  // trailing \b silently dropped it and "70+ countries" postings fell to the
+  // weaker 'regional remote' bucket instead of 'global remote'.
+  if (classifyRemote('Fully remote — hiring in 70+ countries') !== 'global remote') {
+    failures.push('classifyRemote did not read "70+ countries" as global remote');
+  }
+  if (classifyRemote('US-only remote') !== 'geo-restricted') {
+    failures.push('classifyRemote geo-restricted precedence regressed');
+  }
+
   if (failures.length > 0) {
     console.error(`analyze-patterns self-test failed: ${failures.join('; ')}`);
     process.exit(1);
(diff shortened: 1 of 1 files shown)


abankar1:
Expanded to cover a second instance of the same class in this file: `classifyRemote`'s `70+` global-remote signal also ends in `+`, so its trailing `\b` never matched and "remote in 70+ countries" fell through to `regional remote` instead of `global remote`. Fixed the same way (`(?<!\w)/(?!\w)`; word alternatives unaffected) with a self-test regression. Now a complete symbol-edge `\b` sweep of analyze-patterns — 3 commits (tech fix, tech test, remote fix+test).

santifer:
Merged, @abankar1. Three of the most common stacks on the planet were silently absent from the tech-gap rollup, and the reason is a single character: a trailing `\b` can never match after `+`, `#` or a leading `.`, because there's no word boundary between a symbol and a space.

What makes this worth more than the fix itself is that it produced no error and no empty result. The rollup looked complete. Someone doing a C# job search got a gap analysis with C# missing from it and had no way to tell, which is exactly the class of bug this whole release turned out to be about.

Good call reusing the lookaround form that `skill-extract.mjs` and `upskill.mjs` already carry rather than inventing a third spelling, and ordering the alternation longest-first so React Native wins over React at the same position.

PR #1929, opened 2026-07-15
feat(providers): Oracle Recruiting Cloud (ORC) zero-auth provider

Souptik96:
## What does this PR do?

Adds an `oraclecloud` scanner provider so Oracle Recruiting Cloud / Fusion Candidate Experience career sites (`*.fa[.<region>][.ocs].oraclecloud.com`) become first-class, zero-token scan targets. Large employers — JPMorgan Chase, Oracle, BNY Mellon, American Express, Honeywell — run their careers site on ORC, and those roles previously couldn't be auto-scanned.

## Related issue

Part of #230 (scanner-providers umbrella).

No dedicated issue — new zero-auth scanner provider (CONTRIBUTING exempts providers from issue-first).

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation / translation
- [ ] Refactor (no behavior change)

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/santifer/career-ops/blob/main/CONTRIBUTING.md)
- [x] If this is a new feature or architecture change, I opened an issue first (bug fixes, providers, docs & translations are exempt — send those straight in)
- [x] My PR does not include personal data (CV, email, real names, scan results, or pipeline data)
- [x] I ran `node test-all.mjs` and all tests pass
- [x] My changes respect the [Data Contract](https://github.com/santifer/career-ops/blob/main/DATA_CONTRACT.md) (no modifications to user-layer files)
- [x] My changes align with the [project roadmap](https://github.com/santifer/career-ops/discussions/156)

---

### How it works

- Hits the public `recruitingCEJobRequisitions` REST API (GET, no auth/token/cookie). `siteNumber` is auto-derived from the `/sites/<siteNumber>/` segment of `careers_url` (default `CX_1`), overridable via `provider: oraclecloud` + `siteNumber:` and an optional `locationId:` some tenants (BNY, Amex) need to scope results.
- Mirrors the existing provider pattern: SSRF-hardened host allowlist (`assertOracleUrl` — `https:` + `*.oraclecloud.com` host-shape match) called before every fetch with `redirect: 'error'`; offset pagination with a page cap; 429/5xx retry with exponential backoff + jitter and a `ctx.sleep` fallback (mirrors `workday.mjs`).
- Exports `parseOracleResponse` for unit testing; reuses `decodeEntities` from `_html-entities.mjs` and a NaN-safe date parse. Emits the standard `{ title, url, company, location, description?, postedAt? }` Job shape.

[...]

PR #1768, opened 2026-07-10
feat(scan): track portal-health over time and surface coverage decay

Shaamak:
This PR resolves #1744 by tracking portal reachability over time to prevent silent coverage decay. 

**Changes:**
- **Data Layer**: Introduced `data/portal-health.tsv` (tracked in `.gitignore` and `DATA_CONTRACT.md`) to record per-target reachability (`reachable`, `slug_gone`, `network`, `empty`) append-only at the end of each non-dry scan.
- **Scanner Escalation (`scan.mjs`)**: Replaced stateless per-run warnings with streak tracking. Portals failing for 3+ consecutive runs trigger an escalated `🚨 FIX NEEDED` alert.
- **Analytics Surface (`stats.mjs`)**: Extended the `Portals:` summary output to include the count of persistently dead portals, providing visibility directly in the lifetime pipeline stats.
- **Validation**: Added a self-test in `test-all.mjs` verifying streak computation logic.

## Summary by CodeRabbit

- **New Features**
  - Added persistent portal “health” history across scans, tracking consecutive failures to distinguish newly unavailable vs persistently unavailable portals (configurable threshold, defaulting to 3).
  - Extended coverage statistics with a `persistentlyDead` metric and surfaced a warning in the scan summary when any persistently unavailable portals are detected.
- **Documentation**
  - Documented the new portal health history file as user-layer content (never auto-updated) and updated ignore rules for the generated TSV.
- **Tests**
  - Added unit/regression coverage for consecutive-failure detection, persistent-dead counting, and successful operation when portal health data is absent.

Issue #1702, opened 2026-07-08
feat(portals): standalone fix-slugs.mjs — auto-write ATS slug fixes to portals.yml (revives #1612)


Schlaflied:
## Problem

`verify-portals.mjs` (#1016, extended in #1455) already detects stale ATS board slugs and prints a suggested replacement (`→ try ashby/top-hat`), but it's read-only — every fix has to be applied to `portals.yml` by hand.

Hit this today doing a real sweep of `tracked_companies`: 9 of ~25 entries had dead slugs (companies migrated ATS vendor, or a slug was just mistyped). 6 were fixable with a suggested alternate from `verify-portals.mjs` itself; I ended up hand-editing `portals.yml` for each one. This will keep recurring for any user with a populated `tracked_companies` list, since ATS migrations are constant and otherwise invisible until a scan silently returns zero jobs for a company.

## Prior art

@SajithKSwami proposed a `fix-slugs.mjs` in #1612 (closed as part of a larger bundled PR — see maintainer comment: *"fix-slugs.mjs — closest to mergeable... bring it back as its own small PR"*). That version fetches each company's live careers page and detects the ATS from the redirect/HTML, driven by a hardcoded company list.

## Proposed shape

A standalone `fix-slugs.mjs` (or a `--fix`/`--apply` flag on `verify-portals.mjs` directly) that:

1. Reuses `verify-portals.mjs`'s existing probe + `discoverAlternates()` suggestion logic (added in #1455) instead of re-implementing careers-page scraping — it already tries Greenhouse/Ashby/Lever variants and attaches `suggested: { ats, slug }` to failing entries.
2. Reads failures straight from a `verify-portals.mjs` run (or `portals.yml` directly) instead of a hardcoded company list, so it works for any user's `tracked_companies`, not just a personal scan set.
3. Writes the resolved slug back into the matching `portals.yml` entry, with a short inline note (e.g. `(slug migrated lever->ashby YYYY-MM-DD, verify-portals)`), and leaves genuinely unresolved entries untouched.
4. `--dry-run` prints the diff without writing (matches the existing convention in #1612's version).

Happy to take a pass at this myself if it's a welcome direction — wanted to open the issue first per CONTRIBUTING.md.

santifer:
9 of ~25 dead slugs in a real sweep is the evidence that settles it, @Schlaflied. verify-portals.mjs detecting but not writing was always half the tool. #1703 is in the review queue.

PR #2236, opened 2026-07-28
fix(scan): record auth/server/unknown fetch errors in portal-health instead of 'reachable'

Souptik96:
## What does this PR do?

Fixes a gap in the portal-health tracking from #1768: the status assignment in `scan.mjs` only recorded `slug_gone` / `network` / `empty` — error kinds **`auth` (401/403), `server` (5xx), and `unknown` fell through to the default `'reachable'`** in `data/portal-health.tsv`.

Consequence: a portal that 403s the scanner on every run (a real-world mode — e.g. some Oracle Cloud tenants WAF-403 datacenter IPs, noted back in #1929) was recorded as *healthy* forever. No failure streak, no `🚨 FIX NEEDED` escalation, and `stats.mjs`'s `persistentlyDead` counter missed it too. Silent coverage decay is exactly what #1768 was built to surface; this gap defeated it for 3 of the 5 `classifyFetchError` kinds.

### Changes

- **Status assignment (`scan.mjs`)**: each errored target is now recorded under its real `classifyFetchError` kind. The TSV vocabulary is additive — `auth`/`server`/`unknown` join the existing `reachable|slug_gone|network|empty`. Same 3 columns, no schema change.
- **Streak logic** (`computeConsecutiveFailures` + the `stats.mjs` mirror): condition inverted — healthy statuses (`reachable`/`empty`) reset the streak, **every other status counts**. On legacy 4-status TSVs this computes identical results (backward compatible), and the inversion means future error kinds can't silently fall outside the streak again.
- **Escalation loop**: all error kinds can now reach the `🚨` persistent list once their streak crosses the threshold. Below threshold, `auth`/`server`/`unknown` keep their existing one-off `Errors (N):` print — per-run output for transient errors is unchanged.
- **Tests**: streak assertions for auth/server counting + recovery reset + legacy-status backward-compat; a `computePortalStats` fixture where a WAF-blocked portal crosses the threshold.
- **`DATA_CONTRACT.md`**: documents the full status vocabulary (and that older files carry only the first four).

### Not in scope

- Persisting raw error messages/HTTP statuses to the TSV (schema change).
- `PORTAL_HEALTH_PATH` cwd resolution — that's #2161 (different lines, no conflict).
- Auto-remediation of dead slugs — that's #1702/#1703.

## Related issue

None — bug fix (straight to PR per CONTRIBUTING).

## Type of change

- [x] Bug fix
- [ ] New feature

## Verification

```
node --check scan.mjs stats.mjs test-all.mjs
node test-all.mjs --quick
```
New assertions pass:
- `computePortalStats counts auth/server streaks as persistently dead; recovery resets`
- `computeConsecutiveFailures: auth/server/unknown count, reachable/empty reset, legacy statuses unchanged`

The 5 failing checks locally are identical on clean `main` (verified by stashing the diff and re-running): the pre-existing `node:sqlite` Node-v20 tracker tests and the #912 collision test — unrelated to this change.

## Summary by CodeRabbit

- **New Features**
  - Added portal health tracking for authentication, server, network, unreachable, and other failure statuses.
  - Persistent failure detection now includes all non-healthy statuses, while healthy and empty results reset failure streaks.
  - Documented the new portal health data file and supported status values.

- **Bug Fixes**
  - Improved escalation accuracy for portals experiencing recurring non-standard failures.

- **Tests**
  - Added coverage for expanded failure statuses and streak-reset behavior.

### DIFF
--- a/DATA_CONTRACT.md
+++ b/DATA_CONTRACT.md
@@ -28,7 +28,7 @@ These files contain your personal data, customizations, and work product. Update
 | `data/pipeline.md` | Your URL inbox |
 | `data/scan-history.tsv` | Your scan history (9 tab-separated columns; col 8: local SimHash JD fingerprint for cross-listing detection, col 9: posting date) |
 | `data/scan-runs.tsv` | Your per-run scan counters (appended by `scan.mjs`, read by `stats.mjs`) |
-| `data/portal-health.tsv` | Consecutive reachability status for scanned portals (appended by `scan.mjs`) |
+| `data/portal-health.tsv` | Consecutive reachability status for scanned portals (appended by `scan.mjs`; statuses: `reachable`, `empty`, `slug_gone`, `network`, `auth`, `server`, `unknown` — the last three joined the vocabulary later, so older files carry only the first four) |
 | `data/follow-ups.md` | Your follow-up history |
 | `data/offers/*` | Your received offers/contracts, promise notes, prep reports, and reply drafts (PII — gitignored, written by the `offer-prep` mode) |
 | `data/salary-observations.tsv` | Your append-only compensation observation log: `{tracker#}\t{date}\t{desired\|advertised\|actual}\t{amount}\t{currency}\t{source}\t{note}`. Written by interactive modes when a figure is stated/confirmed; never edited in place. Advertised figures come from reports' `advertised_comp` instead — reports are themselves observation sources. Read by `salary-gap.mjs` |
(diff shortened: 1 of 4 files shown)


santifer:
Merged, @Souptik96. A portal returning 401 or 500 recorded as `reachable` is the worst shape a bug can have here: the health file says the board is fine, so nothing ever surfaces, and the user quietly stops seeing jobs from an employer they're actually watching.

The part I'd point people at is the inversion. You could have fixed this by adding `auth`, `server` and `unknown` to the list of failure statuses, and it would have worked today. Instead you flipped the condition so healthy statuses reset the streak and **everything else** counts toward it, which means the next error kind someone adds can't silently fall outside it the way these three did. Fixing the instance is normal; closing the class is not.

Keeping `empty` on the healthy side is right too, and worth having in a comment: a live board with zero openings is reachable, not broken.

PR #1297, opened 2026-06-29
Feat/ATS-keyword-coverage-1285

DevangBanda:
This PR Solves Issue #1285 

A standalone keyword-match.mjs (mirrors analyze-patterns.mjs: pure testable core + thin CLI + --self-test). It reads the JD keywords straight from a report's ## Keywords extracted block, scans the CV separately, and emits a ready-to-paste ## Keyword Coverage section — coverage %, present, thin (present but mentioned once = under-represented), and missing.

Matching is case-insensitive, word-boundary aware (so Java ≠ JavaScript), symbol-safe (C++, CI/CD, Node.js), with conservative plural folding and a small bidirectional synonym map (k8s↔kubernetes, js↔javascript, …). No fuzzy matching — a false "covered" would defeat the purpose.

Diagnostic only — never fabricates. The script is physically read-only (zero write calls); it reports gaps and the output tells the user to add a keyword only if it reflects real experience. Upholds the "reformulated, never fabricated" rule from modes/_shared.md.
Local + zero-dependency. Node built-ins only (fs/path/url) — no new package, no API key.

Acceptance criteria
✅ Outputs coverage % + matched/missing lists (also --json).
✅ Deterministic test section in test-all.mjs (no network) + --self-test in the script-execution list.
✅ Registered in SYSTEM_PATHS (update-system.mjs) and package.json.

## Summary by CodeRabbit

* **New Features**
  * Added an ATS-style keyword coverage CLI with Markdown or JSON output, identifying present, thin, and missing keywords.
  * Added support for custom CV paths, Markdown and HTML inputs, synonym matching, and singular/plural variants.
  * Added keyword coverage verification to evaluation and PDF workflows.
* **Documentation**
  * Updated workflow documentation to include diagnostic coverage results without injecting keywords.
* **Tests**
  * Added self-tests and automated checks for matching, parsing, HTML conversion, and error handling.

PR #2239, opened 2026-07-28
feat: cv-hm-audit mode — adversarial hiring-manager review of a tailored CV

### LINKED ISSUE #2238: feat: cv-hm-audit mode — adversarial hiring-manager review of a tailored CV
## Problem

career-ops tailors a CV per role, then generates a PDF. Between those two
steps there is no judgment gate — nothing asks *"would the person who screens
this actually advance it?"* The goal is to raise the success of the tailored CV as much
as possible when the HM views it. Also, generating this targeted feedback gives you
a more realistic look into how likely your resume is going to be screened, and gives you
the chance to make more modifications if the HM persona feels that your resume is not
strong enouhg.

The repo has one adjacent guard, `verify-cv-facts.mjs`, wired into `pdf.md`
step 18. It is a **mechanical** fact gate: it diffs generated output against
`cv.md` / `article-digest.md` to catch invented metrics. That is a necessary
check and an entirely different one. A bullet can be perfectly truthful, fully
sourced, and still be the wrong bullet for the role — buried lede, wrong
altitude, wrong vocabulary, answering a requirement the JD never raised.
No existing mode catches that.

**Prior art:** none. Searched issues and PRs for "hiring manager persona",
"cv audit", "bullet audit", "adversarial review", "reviewer persona", and
"resume audit". Only tangential hits (#593 cold email, #1608 offer-prep).

## Proposal

A new prompt-only mode, `cv-hm-audit`, that researches who is likely to screen
the application, dispatches a **separate** subagent role-playing that reviewer,
and returns a bullet-by-bullet keep/cut/rewrite verdict plus an overall
scope/seniority read — presented to the user *before* the final PDF, and
persisted to the role's report.

Two properties are load-bearing and neither works without the other:

1. **The reviewer is external.** A separate agent, not the one that wrote the
   bullets.
2. **The reviewer is research-grounded.** A docs lead, a VP Engineering, and a
   recruiter weigh the same bullet very differently. A generic "hiring manager"
   persona degrades into generic resume advice.

### Scope boundaries

- **Not** a fact checker. `verify-cv-facts.mjs` owns that and runs first.
- **Not** a rewriter. It recommends; the user decides; `pdf` regenerates.
- **Not** mandatory. `pdf.md` gains a pointer line, not a required step.
  Users who want it every time can say so in their own `modes/_custom.md`.

---

## Interface

```
/career-ops cv-hm-audit {report#|company-slug}
```

Argument omitted → most recent evaluated role. Same argument pattern as
`/career-ops cover {slug}`.

**Name rationale:** `cv-`, not `resume-`, because the repo's vocabulary is
"CV" 318 times to "resume" 32, and every script is `cv-`-prefixed
(`build-cv-html`, `cv-sections-core`, `cv-sync-check`, `verify-cv-facts`).
Three tokens makes it the longest mode name in the catalog; that is accepted
deliberately so the entry is self-describing in a 30+ entry argument-hint list.
`cv-audit` was rejected as confusingly close to the unrelated `verify-cv-facts`.

### Inputs (all already on disk)

| Input | Source | Notes |
|---|---|---|
| The JD | `jds/{slug}.md`, or the report's `**URL:**` header | |
| Tailored bullets | `/tmp/cv-{candidate}-{company}.json`, else newest `output/cv-*-{company}.html` | see below; `.pdf` is never parsed |
| Factual floor | `node jd-skill-gap.mjs` | zero-LLM; buckets every JD requirement `existing` / `supportedByResume` / `gap` |
| Scope of truth | `cv.md`, `article-digest.md`, `modes/_profile.md` | bounds what the persona may recommend |

**Bullet source, and its wrinkle.** `pdf.md` writes the structured CV to
`/tmp/cv-{candidate}-{company}.json`, which is ephemeral; the durable artifact
is the generated HTML in `output/`. The mode reads the JSON when present
(clean `experience[].bullets[]` iteration) and falls back to extracting `<li>`
items from the generated HTML when it is gone. No parser script — the fallback
is prompt-level and silent.

Because the JSON is the better input, the natural time to run this is
immediately after `pdf` in the same session, which is where the `pdf.md`
pointer will suggest it.

---

### Step 1 — Gather

Resolve the report; locate the newest tailored CV artifact for that company;
run `node jd-skill-gap.mjs`; load the scope-of-truth files.

### Step 2 — Identify the reviewer, and declare the tier

Research: LinkedIn, the company careers/team page, ATS posting metadata, and
targeted WebSearch (`"{company}" "{role}" hiring manager`,
`"{company}" head of {function}`).

| Tier | Trigger | Persona built from | Labeled as |
|---|---|---|---|
| **A** | Named person, 2+ independent sources agreeing on role + company | Their actual cited background | "Identified — {name}, {title}" + source links |
| **B** | Named person, single weak source | Their apparent *function* only, never claimed specifics | "Likely reviewer" |
| **C** | Nobody identifiable | Company stage/size, team composition, the JD's reports-to line and its own vocabulary | "Synthesized" |

All three tiers are research-grounded. They differ only in how much of the
grounding is a real identifiable person. **Tier C is a constructed reviewer
built from actual findings, not a stereotype** — if research surfaced that the
company is 40 people, the role reports to a Director of Engineering, and the JD
is written in platform-team vocabulary, the synthesized reviewer reflects that.

The tier is always stated in the output. A reader must never have to guess how
much the persona is worth.

### Step 3 — Brief and dispatch

One subagent, per the existing convention at `SKILL.md:180`. Never nested.

The brief carries: the JD, the numbered bullets, the `jd-skill-gap` map, the
persona and its tier, and a hard guardrail — the candidate's real scope from
`cv.md` / `article-digest.md`, with an explicit instruction that the reviewer
may recommend cutting or reframing but may **never** recommend a claim the
source files do not support.

### Step 4 — Verdict

The agent returns one row per bullet:

| # | Bullet | Verdict | Why | Suggested rewrite |
|---|---|---|---|---|
| 1 | … | `keep` / `cut` / `rewrite` | one line | only when `rewrite` |

Plus an overall scope/seniority read and a blunt *"would I advance this to a
screen?"* call.

**Anti-drift rule:** the mode states the bullet count before dispatch and
requires the returned table to match it. Agents drift toward summarizing
instead of auditing every line; a stated row count is the in-prompt defense.
If a mechanical guarantee proves necessary in practice, a follow-up PR can add
a `cv-hm-audit.mjs` helper that emits a fixed table skeleton from the CV JSON —
deliberately out of scope here to keep this PR reviewable as prose.

### Step 5 — Present, then persist

Show the identity guess, the tier, and the table to the user for the judgment
call **before** any PDF regeneration.

Then append `## HM Audit` to `reports/{num}-{company}-{date}.md`: persona,
tier, sources, table, overall verdict.

---

## Output routing rationale

The audit is **role-specific** — one JD, one tailored CV, one reviewer — so it
belongs to the role's canonical artifact. Two existing precedents settle this:

- **`oferta.md:391`** already appends a generated artifact to the report: the
  cover letter draft lands under `## Cover Letter Draft`, and `/career-ops
  cover` later refines it. "Generated artifact becomes a `##` section of the
  report" is established convention, not an invention.
- **`interview-redflag.md:112`** contains an explicit *Output routing* block
  reasoning about this exact question. Its logic is scope-based: red flags are
  *company*-wide, so they get a company-level file, and it states *"Do not
  append to per-role intel files — those are role-specific, this is
  company-specific."* Running that same test on a CV audit yields the opposite
  and correct answer: role-specific → the role's report.

`tracker.md:39` reinforces it — *"Never rename the report file — the number is
the identity."* The report is the role's stable identity and is already linked
from the tracker.

**Reuse at interview time:** `interview-prep.md` already reads prior context
for a role (line 25). It gains one Inputs line — *the report's `## HM Audit`
section, if present* — so persona research paid for at CV time resurfaces
during prep. Writing an `interview-prep/{company}-{role}.md` file at CV time
was rejected: it would create prep files for every role that never responds.

---

## Failure modes

| Situation | Behavior |
|---|---|
| No report for the role | Present inline; state plainly it was not persisted; suggest `oferta` first. Never create a stray file. |
| No tailored CV in `output/` | Stop and point at `pdf`. **Never fall back to auditing `cv.md`** — the untailored master produces confidently wrong verdicts. |
| No JD text reachable | Degrade: audit against the report's requirement summary, caveat stated. Follows `interview-prep.md:171` — *"partial-but-honest output, not perfect-or-nothing."* |
| Research finds a stale profile | Cap at Tier B; flag the recency doubt (the person may have left). |
| No web research available | Tier C, stated. |
| CLI has no Agent primitive | Run the persona inline **and say so**. The value is that it is not self-auditing; degrading silently would misrepresent the result. |

## Guardrails

- **Fabrication.** The persona may cut or reframe, never invent. Explicitly
  bound to the Source-of-Truth Boundary in `AGENTS.md`.
- **Privacy.** Public professional information only. The report stores name,
  title, and source links — never contact details, never personal social
  accounts.
- **Attribution.** Always *"a reviewer with this background would likely read
  it this way,"* never *"{name} thinks X."* Even at Tier A this is inference
  from public information about a real private individual.
- **Output language.** `language.output` applies, per the standing AGENTS.md
  directive.
- **Ethics alignment.** career-ops' Ethical Use section targets application
  spam. This mode pushes the same direction — it is a quality gate that makes
  a submission better or stops it, never a throughput multiplier.

---

## PR surface

Six files. No new scripts, no new tests required.

| File | Change |
|---|---|
| `modes/cv-hm-audit.md` | new — the mode |
| `modes/README.md` | one catalog row |
| `AGENTS.md` | one Skill Modes row |
| `.claude/skills/career-ops/SKILL.md` | argument-hint + table row + usage example |
| `modes/pdf.md` | one pointer line after the fact gate |
| `modes/interview-prep.md` | one Inputs line |

Two checks confirmed the surface is this small:

- **Only `AGENTS.md` carries the Skill Modes table.** `CLAUDE.md`, `CODEX.md`,
  `OPENCODE.md`, `KIMI.md`, and `GEMINI.md` do not mirror it, so there is no
  five-way doc sync.
- **Nothing in `tests/` enforces mode registration.**
  `tests/output-language.test.mjs` only exercises a parser function. A
  prompt-only mode adds no test burden.

---

## Implementation status

Following CONTRIBUTING (*"for a new feature, a new mode or command … please open an issue first"*), so this is the issue — not a PR.

A branch is drafted and ready if the direction is welcome: 6 files, +131 −1, one new mode file and five one-line registrations/hooks. No new scripts and no new tests. `node test-all.mjs` reports 2253 passed / 0 failed / 1 warning, identical to a clean `upstream/main` checkout (the warning is the pre-existing `cv-sync-check.mjs exited with error (expected without user data)`).

Happy to change the shape, the name, or the output routing before opening anything — or to drop it if it doesn't fit the roadmap.

xu-albert:
Closes #2238

Adds a prompt-only `cv-hm-audit` mode: it researches who is likely to screen an application, dispatches a **separate** subagent role-playing that reviewer, and returns a bullet-by-bullet keep/cut/rewrite verdict plus a scope/seniority read — presented before the final PDF and appended to the role's report as `## HM Audit`.

I have been using this personally while tailoring my resume, and I find that an HM persona gives honest and helpful feedback on my resume, and will dissuade me from wasting my time applying if the resume is too weak. It has also gone through and edited wording or surfaced more relevant job experiences and skills that the base resume builder does not catch.

### Why this isn't covered today

`verify-cv-facts.mjs` at `pdf` step 18 is a **mechanical** fact gate: it diffs generated output against `cv.md` / `article-digest.md` to catch invented metrics. It cannot judge whether a truthful bullet is the *right* bullet — buried lede, wrong altitude, wrong vocabulary, answering a requirement the JD never raised. And the agent that tailored the bullets is the worst reviewer of them: it grades its own work and drifts toward summarising instead of auditing.

### Design decisions, and the precedent behind each

**Output routing → the report.** `oferta.md` already appends a generated artifact to the report (`## Cover Letter Draft`), so "generated artifact becomes a `##` section" is existing convention. `interview-redflag.md`'s own output-routing block decides by scope — company-wide findings get a company file, and it explicitly says *"Do not append to per-role intel files — those are role-specific, this is company-specific."* A CV audit is role-specific, so the same test puts it in the role's report. `interview-prep.md` then gains one Inputs line so the persona research resurfaces at interview time instead of being re-done.

**Tiered persona, always labeled.** Tier A = named person corroborated by 2+ independent sources; Tier B = single weak source, function only; Tier C = nobody identifiable, so the reviewer is *synthesized from what research did surface* (company stage, team composition, the JD's reports-to line and vocabulary) rather than being a stereotype. The tier is always stated, so a reader never has to guess how much the persona is worth. Attribution is always *"a reviewer with this background would likely read it this way"* — never *"{name} thinks X."*

**Research respects third-party ToS.** Person research goes through WebSearch and the company's own pages, matching how `modes/_shared.md` already frames it (`WebSearch | … LinkedIn contacts`). The mode explicitly forbids automated access to platforms whose terms prohibit it, and search queries carry company and role terms only — never CV content or candidate details.

**Name.** `cv-` rather than `resume-` to match the repo's vocabulary (CV outnumbers "resume" 318:32, and every script is `cv-`-prefixed). Three tokens makes it the longest mode name in the catalog — accepted so the entry is self-describing in the argument-hint list. Happy to shorten if you'd rather.

### Surface

Six files, **no new scripts and no new tests**: one new mode plus five one-line registrations/hooks. Only `AGENTS.md` carries the Skill Modes table, and the skill edit goes to the canonical `.agents/skills/career-ops/SKILL.md` so all six CLI symlinks pick it up.

`node test-all.mjs` → **2253 passed, 0 failed, 1 warning** — identical to a clean `upstream/main` checkout (the warning is the pre-existing `cv-sync-check.mjs exited with error (expected without user data)`).

### Not included, deliberately

- Per-bullet coverage is an in-prompt rule (state the count, require the table to match) rather than mechanically enforced. A `cv-hm-audit.mjs` helper emitting a fixed table skeleton would guarantee it, but that turns a prose-reviewable PR into one with parsing logic and tests. Glad to add it if you'd prefer it hardened.
- Multi-persona panel (recruiter + HM + peer) — better signal, ~3× token cost, a poor fit for free-tier users.

Happy to change the shape, the name, or the routing.

## Summary by CodeRabbit

* **New Features**
  * Added an optional hiring-manager audit for tailored CVs, providing research-based, bullet-level keep, cut, or rewrite recommendations.
  * Audit results are included in evaluation reports and can inform interview preparation.
  * Added clear handling for missing inputs, limited research, privacy, attribution, and factual accuracy.

* **Documentation**
  * Documented how to enable and run the audit, review its results, and incorporate accepted revisions before PDF generation.


xu-albert:
Thanks — four of these were real, and one was the kind of bug that would have made the mode quietly useless. Pushed in 202ef97.

**Fixed**

- **Context-loading contract.** Correct, and the most important one: the mode was in the router table but in *neither* loading list, so `/career-ops cv-hm-audit` would have matched the keyword and loaded no mode or profile context. One deviation from the suggestion — I added it to the **`_shared.md`** list next to `pdf`, not the standalone list. The mode proposes CV bullet rewrites, so it needs `_shared.md`'s ATS rules and the "keywords get reformulated, never fabricated" constraint governing what it may suggest. Standalone would have given it profile context without the rules.
- **Artifact resolution.** Right that newest-company-match is unsafe: filenames carry candidate and company but not role, so a company with two open roles could get the wrong CV audited silently. Now resolves by report through `data/pdf-index.tsv`, then an explicit path, and only then falls back — and says so when it does.
- **No-JD path.** Genuine contradiction on my side: Step 1 permitted degrading while Step 3 still required the skill-gap output. The brief now marks the classification unavailable and forbids inferring requirement coverage from the report summary alone.
- **Idempotent persistence.** Real hole — nothing defined re-run behavior. Now exactly one `## HM Audit` per report, replaced wholesale on re-run, with the artifact path and date recorded so it's unambiguous which CV was judged.
- **Sources fallback.** Fair, and a bit embarrassing in a mode about not asserting unsupported things. Now `none — synthesized from the available JD/report context`.

**Declined, with reasons**

- **`<li>` scoping.** Not valid for the shipped templates. The only two `<li>` emitters in `build-cv-html.mjs` are lines 319 and 340, both iterating `e.bullets`; skills render as `<div class="skill-item">` and competencies as `<span class="competency-tag">`. There is no `<li>` outside bullet arrays, so the extraction can't pick up skills. I did note in the mode that `<li>` is bullets-only, so the assumption is explicit.
- **Making the audit mandatory before PDF.** Deliberate, not an oversight. Requiring it would add a subagent dispatch plus web research to *every* PDF for every user, including the free tiers career-ops supports, and imposing a new required step on a core flow felt like a maintainer's call rather than something to land in a first PR. Upstream it's opt-in with a pointer from `pdf.md`; anyone who wants it every time can require it in their own `modes/_custom.md`. Happy to flip it if you'd rather it be a hard gate.

`node test-all.mjs` still 2253 passed / 0 failed / 1 warning, matching a clean `main` checkout.

xu-albert:
Good catch on the gap — fixed in 4095119, though not with the suggested diff.

**The gap is real.** Step 1's degraded path read *"continue against the report's requirement summary"*, which silently assumed a report existed, while Step 5 separately allowed an inline audit when there was none. So no-report-**and**-no-JD fell through with nothing to stop it, and the reviewer would have judged bullets with zero role context. It's reachable too: `pdf.md` supports one-off CVs with no tracker entry, and a JD pasted into the conversation is never written to `jds/`.

**Two reasons I didn't take the proposed diff.**

It required the user to supply the tailored-artifact path. But the artifact was never the weak link — Step 1 already hard-stops when no tailored CV resolves, and Input 4's first resolution path is the `/tmp` JSON from the same session, which is fully grounded and needs nothing from the user. Requiring a manual path would break the mode's most common invocation: running it immediately after `pdf` in the same session. The missing precondition was **role context**, not the artifact.

It also placed the precondition in Step 5, which is the persistence step. Every other input gate lives in Step 1. Splitting the input contract across two sections is the same structural mistake behind the earlier no-JD finding — Step 1 permitted something a later step didn't account for — so repeating it would undo that fix.

**What landed**, in Step 1 alongside the other gates:

> If a report exists but no JD text is reachable, continue against the report's requirement summary and state the degradation in the output.
>
> If **neither** a JD nor a report is available, stop: the reviewer would have no role context, and a verdict on bullets with nothing to judge them against is worse than no verdict. Ask for the JD text or the posting URL.

Step 5's no-report behaviour is unchanged — with role context present, an inline audit is still valid and still refuses to create a stray file.

`node test-all.mjs` → 2253 passed / 0 failed / 1 warning, matching a clean `main`.

santifer:
Thanks @xu-albert, the idea is genuinely on the core path: an adversarial hiring-manager read of the tailored CV is a quality gate before apply, and the issue-first move (#2238) was the right way in.

One routing question before a full review, and it comes from the scope criterion we just published ([What belongs in core, test 4](https://github.com/santifer/career-ops/blob/main/CONTRIBUTING.md#what-belongs-in-core-the-parallel-feature-test)): does this need to be its own mode, or is it a *pass* inside an existing one? The "audit my CV" space is getting crowded (deterministic parseability just landed via #2241, keyword coverage is in flight in #1297, and a rubric-based content score is #2284), and every new top-level mode adds permanent surface: routing, docs, agent context. Two shapes that could work: a `--hm-audit` pass invoked from the pdf/tailoring flow, or folding it into interview-prep (which you already touch) as the "how a hiring manager reads you" step. If you think standalone is right anyway, make that case: the criterion is a routing conversation, not a door.

xu-albert:
This doesn't add routing surface, and the intention is to have an optional step to run an audit with the hm persona if needed (for example, borderline score, career pivot, etc.)

It's not a pass, not a mode. modes/cv-hm-audit.md → modes/pdf/hm-audit.md, reached through the pdf flow the way modes/heuristics/recruiter-side.md is reached through the modes that load it. .agents/skills/career-ops/SKILL.md now has zero diff against main — the argument-hint token, the router row and the _shared.md loading-list entry are all gone. It turns out it never needed that last one: pdf has already read _shared.md + _profile.md + _custom.md by the time the pass runs, so the ATS rules and the "reformulated, never fabricated" constraint are inherited rather than declared a second time.

Off by default, on --hm-audit. Step 20 runs only when the invocation carried the flag or modes/_custom.md turns it on as a house rule. This should happen before interview-prep since it's a final resume audit, and this "best resume" gets us in the door.

xu-albert:
review fixes:

- **Unsupported cost claims.** I'd written that the pass costs "several times the tailoring" and was "real budget on the free tiers." Neither was measured. Both lines now say only what's checkable — it adds a subagent dispatch plus web research — and the opt-in argument doesn't need the multiplier.
- **Output language.** Step 1's missing-CV message was literal English, contradicting the file's own guardrail and the `[Render in {language.output}: ...]` pattern `pdf.md` Step 4 already uses. Converted. Step 3's strings stay English on purpose: those are subagent brief text, not user-facing output.
- **`LOW CONFIDENCE` routing** — my own bug from the fix above. I'd sent the low-confidence case and the no-JD case down the same path. They're not the same: when the diagnosis fires the JD *is* reachable, only the automated pass over it failed, so falling back to the report summary throws away a posting that's right there. Split now.
- **Stale persisted audit.** The audit judges the CV at Step 19, but accepted rewrites send `pdf` back to Step 17 — so the rendered PDF could diverge from the table stored in the report, and `interview-prep` would resurface criticism the user already acted on. Persistence now waits for the user's decision and records it. Not by re-running the audit; that doubles the cost for a verdict already acted on.

I declined a fifth, which asked me to drop the `AGENTS.md` and `modes/README.md` rows as contradicting the not-a-routable-mode policy. The inconsistency was real but it was in my sentence, not the rows — I'd overclaimed "no entry in the `AGENTS.md` mode catalog" while both rows exist. They point at `pdf` rather than at a standalone mode, so they don't grant the route the policy rules out, and dropping them trades discoverability for a wording error. Fixed the wording; `heuristics/` and `interview/` sit in that same Subdirectories table without being routable either. Happy to cut them if you read it differently.

`node test-all.mjs` → 2716 passed, 0 failed, 1 warning. Ran a clean `main` worktree side by side rather than going from memory: identical line.

santifer:
This one earns its place, @xu-albert — and it was the first proposal ever run through the published what-belongs-in-core test, so how you handled it mattered beyond the feature. Asked whether it was a mode or a pass, you picked the narrower answer, deleted your own routing surface down to a zero diff, and then noticed the pass didn't even need the loading-list entry because pdf already inherits the rules by the time Step 20 runs. The content held up to a hard read: the bullet-count coverage rule (re-dispatch for missing rows, not accept a partial audit), the A/B/C reviewer tiers with verifiable triggers, the verbatim never-invent-coverage contract in the brief, wholesale section replacement so a superseded audit can't resurface in interview-prep, and the LOW CONFIDENCE integration from #2341 within days of it landing. Off by default, candidate data never in a search query, and the fact floor is deterministic code. Merged! 🚀

PR #2213, opened 2026-07-27
Fix/sync pdf and scan summary

Shaamak:
## Overview
This PR resolves two outstanding issues affecting tracker hygiene and CLI UX:
1. **PDF Column Sync (#1429, #1969, #1488):** A new idempotency script (`sync-pdf-flags.mjs`) reads the PDF manifest (`pdf-index.tsv`) and correctly flips the `❌` to `✅` inside `applications.md` for any generated PDFs. The script is hooked into the end of the `merge-tracker.mjs` pass so it runs automatically during standard merges.
2. **Scan Summary Redundancy (#2184):** The scan CLI output no longer unconditionally prints the "Filtered by title" and "Filtered by location" rows when those filters are not configured and zero jobs were skipped.

## Changes Made
- Added `sync-pdf-flags.mjs` to reconcile `data/pdf-index.tsv` against `data/applications.md` using `tracker-utils.mjs` transactions.
- Injected a hook into `merge-tracker.mjs` to execute `sync-pdf-flags.mjs` natively.
- Added comprehensive regression tests in `tests/sync-pdf-flags.test.mjs`.
- Wrapped `totalFilteredTitle` and `totalFilteredLocation` logs in `scan.mjs` inside config presence checks.

## Testing
- `node test-all.mjs` passes with the new regression tests.
- Validated `scan.mjs` output with empty filter configurations.

## Summary by CodeRabbit

- **New Features**
  - Added customizable CV section template packs with `<!--ENTRY-->` rendering and conditional blocks (e.g., location, badges, technologies, skills categories, education, projects).
  - Added a `sync-pdf-flags` CLI to reconcile tracker PDF indicators with the PDF index, and it now runs automatically after successful merges.
  - Expanded CLI help and template loading behavior to support per-section partials.
- **Bug Fixes**
  - Portal scan summary now only shows title/location filter removal counts when applicable.
  - Improved conditional rendering in CV sections (including experience location and certification alignment when optional fields are missing).
- **Tests**
  - Added regression coverage for PDF flag synchronization.

Issue #704, opened 2026-05-20
update-system v1.7→1.8 misses new file liveness-browser.mjs, breaks check-liveness.mjs


PHcz:
## Description
The v1.7.1 → v1.8.0 updater did not download `liveness-browser.mjs`, even though it was added in v1.8.0 and is imported by `check-liveness.mjs` (and indirectly by `scan.mjs --verify`).

## Steps to reproduce
1. Be on v1.7.0 (or 1.7.1)
2. Run `node update-system.mjs apply`
3. Update reports success: `Updated 55 system paths.`
4. Run `node check-liveness.mjs <any-url>`

## Expected
Liveness check runs and reports active/expired/uncertain.

## Actual
```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/Users/.../liveness-browser.mjs'
imported from /Users/.../check-liveness.mjs
```

## Root cause (suspected)
`liveness-browser.mjs` was added as a new top-level system file in v1.8.0 but is not in the updater's `SYSTEM_PATHS` allowlist. The updater only refreshes paths it knows about, so brand-new files added between releases get missed.

`check-liveness.mjs` itself was updated to import it, so post-update the script is broken until the user manually fetches the file from upstream.

## Workaround
Manual fetch restores functionality:
```bash
curl -sL https://raw.githubusercontent.com/santifer/career-ops/main/liveness-browser.mjs -o liveness-browser.mjs
```

After fetching, `node check-liveness.mjs <url>` runs cleanly.

## Suggested fix
Add `liveness-browser.mjs` to `SYSTEM_PATHS` in `update-system.mjs`. More broadly, consider a release-prep CI check that asserts every `import` reference in updated files resolves to a file in `SYSTEM_PATHS`, so missing-new-file regressions are caught at release time rather than by users post-update.

## Environment
- macOS, Node.js v25.5.0
- Local version: 1.8.0 (post-update)
- Backup branch confirms: `liveness-browser.mjs` was not in v1.7.0 (so this is a new-file gap, not an overwrite issue)

santifer:
Good catch, @PHcz — exact same bug-class as #656/#657/#658 but a different file. Fix opened in #725 (`liveness-browser.mjs` added to both SYSTEM_PATHS and BOOTSTRAP_PATHS, same pattern as #696).

**Workaround until #725 merges:** `git fetch origin main && git checkout origin/main -- liveness-browser.mjs`

Closing once #725 lands.

Issue #649, opened 2026-05-14
v1.7.1 update ships broken /career-ops symlink (target .agents/ missing)


Nid989:
## Summary

After `node update-system.mjs apply` to v1.7.1, the `/career-ops` slash command stops working. Skill registry shows no `career-ops` entry.

## Root cause

Commit `4b0686a` (v1.7.1 auto-update) replaced `.claude/skills/career-ops/SKILL.md` (regular file, 95 lines) with a symlink:

```
.claude/skills/career-ops/SKILL.md -> ../../../.agents/skills/career-ops/SKILL.md
```

…but the target `.agents/skills/career-ops/SKILL.md` is never created by the updater. Result: dangling symlink, skill loader skips it silently.

## Repro

1. Be on v1.7.0 with working `/career-ops`.
2. `node update-system.mjs apply` → v1.7.1.
3. Restart CLI.
4. `/career-ops scan` → command not found.
5. `ls -la .claude/skills/career-ops/SKILL.md` shows symlink; `readlink` target does not exist.

## Workaround

```bash
mkdir -p .agents/skills/career-ops
git show 4b0686a~1:.claude/skills/career-ops/SKILL.md > .agents/skills/career-ops/SKILL.md
```

Or `node update-system.mjs rollback`.

## Fix

`update-system.mjs` must either:
- ship `.agents/skills/career-ops/SKILL.md` alongside the symlink, or
- drop the symlink design and keep `SKILL.md` as a regular file under `.claude/skills/career-ops/`.

## Environment

- career-ops: v1.7.1
- CLI: Claude Code (Opus 4.7)
- OS: macOS (Darwin 25.3.0)

santifer:
Confirmed and root cause matches your analysis exactly, @Nid989 — that bug report saved me a couple of hours. Thank you.

**The bug**: your local `update-system.mjs` was v1.6.x at the start of the update, so it predates #600 (the fix that added `.agents/` to `SYSTEM_PATHS`). The loop never checked out `.agents/`, but it DID update `.claude/skills/` to its new symlink form — dangling target, silent skill load failure.

**Quick recovery for anyone currently broken** (don't wait for a release):

```bash
node update-system.mjs apply
```

Run it once more. Your local `update-system.mjs` is now v1.7.1 (it got updated in the first run), and the new version has `.agents/` in `SYSTEM_PATHS` — so the second run will check out the missing target. The skill loader picks the symlink back up on the next Claude Code session.

[...]

PR #2240, opened 2026-07-28
fix(test): make the SYSTEM_PATHS coverage guard actually run in CI

santifer:
## The problem

`validate-system-paths-coverage.mjs` exists to stop a new system file from shipping unregistered — the bug class that has landed **five times** (#649, #704, `.editorconfig`, and twice since). It has been a no-op in CI for its entire life.

`test-all.mjs` runs each script from a throwaway copy created *inside* the repo. The check calls `git ls-files`, which from an untracked subdirectory returns **zero paths**. Zero paths means zero orphans, so it printed `OK: 0 tracked files covered` and exited 0 no matter what the real tree looked like.

Reproduced:

```
$ node validate-system-paths-coverage.mjs        # from the repo root
Coverage gap: sync-pdf-flags.mjs                 → exit 1   correct
$ node .tmp/validate-system-paths-coverage.mjs   # as CI ran it
OK: 0 tracked files covered                      → exit 0   green, and wrong
```

An unregistered file is not cosmetic: `update-system` never ships it, so every user who updates silently loses it.

## The fix

- **`validate-system-paths-coverage.mjs`** — an empty file list now fails. A check that cannot look must not pass.
- **`test-all.mjs`** — the bare coverage run leaves the script section, where it could never be meaningful, and runs from `ROOT` instead.
- **`test-all.mjs`** — a probe asserts the guard exits non-zero when invoked where git sees nothing, so it cannot silently become a no-op again.
- **`update-system.mjs`** — registers `sync-pdf-flags.mjs`, which the blind guard let through when #2213 landed.

## Verification

Both directions, not just the happy one:

- Remove the registration → suite goes **red** and names the gap
- Restore it → **2284 passed, 0 failed**

## Note

The unregistered file got in through a merge of mine earlier today, and the reason I caught it at all is that `co-gate --deep` runs this check from the repo root. CI never would have.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation reliability by detecting when repository files cannot be inspected, preventing false-positive coverage results.
  - Added coverage tracking for an additional system-level file so automated updates and restoration remain consistent.

- **Tests**
  - Expanded validation checks to confirm failures are reported in incomplete environments and successful checks pass against the actual repository.
  - Improved diagnostic output when coverage validation detects an issue.

### DIFF
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -189,7 +189,11 @@ const scripts = [
   { name: 'reply-matcher.test.mjs', expectExit: 0 },
   { name: 'validate-portals.mjs --file templates/portals.example.yml', expectExit: 0 },
   { name: 'validate-system-paths-coverage.mjs --self-test', expectExit: 0 },
-  { name: 'validate-system-paths-coverage.mjs', expectExit: 0 },
+  // The bare coverage run is NOT here on purpose: this section executes each
+  // script from a throwaway copy of the repo, and the coverage check needs
+  // `git ls-files` on the REAL tree. Running it here validated nothing and
+  // exited 0 no matter what, which is how five unregistered files shipped.
+  // It now runs from ROOT in section 5.
   // Missing-file run: must exit 0 gracefully and hit no network. Do not use the
   // default portals.yml because end-user workspaces often have a real user-layer
   // portals file that would trigger a live remote sweep during tests.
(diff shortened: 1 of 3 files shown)


santifer:
Merged with Santiago's approval. Green on all three OS legs of the new matrix, which mattered here: the fix uses `spawnSync` with an explicit `cwd` and `join(ROOT, …)`, and Windows is exactly where that kind of path handling diverges.

PR #2243, opened 2026-07-28
feat(scan-history): additive normalized-company column for reliable repost keying

Scott-Emberson:
## Agreed follow-up to #1712

This is the agreed follow-up from #1712. In that PR the maintainer held back the hunk that routed the "Reposting pattern" legitimacy signal through **executing** `company-history.mjs`, because execution fails to silent-absence. The diagnosis: the real problem isn't a *source* gap, it's a **data** gap — `scan-history.tsv` stores the raw company string with no normalized key, so `Acme Inc.` and `Acme` are treated as two companies. Fix it **in the file** with an additive column, and the mode, web, and `detect-reposts` all benefit with nothing depending on a script having run.

> "A normalized company column is additive, so it doesn't break the scan-history contract the web parses."

## What changes

- **Writer (`scan.mjs`)** — `formatScanHistoryRow` appends a trailing **`normalized_company`** column (col 12), populated at write time via the existing exported `normalizeCompanyName` (lowercase, punctuation/whitespace folded, trailing legal-entity suffixes stripped) so `Acme Inc.`, `Acme, Inc.` and `ACME  Inc` all key to `acme`. The raw display company (col 5) stays faithful to what the provider returned. No new normalizer was written — this reuses the one already shared across the tracker.
- **Consumer (`detect-reposts.mjs`)** — clustering now keys on a new `companyKey(row)` that **prefers the stored column** and **falls back to normalizing the raw company on the fly** for pre-existing rows that lack it (and for row objects built directly in tests). A final fallback to the raw lowercased name preserves today's behavior for names that fold to empty (all-CJK / all-Cyrillic), so those never over-cluster under a single empty key. This is what makes `Acme Inc.` and `Acme` a single repost cluster instead of two.
- **Docs** — `DATA_CONTRACT.md` and `docs/SCRIPTS.md` describe the new trailing column and the append-only column contract.

The `modes/*` legitimacy tables are intentionally untouched — this PR is purely the data-layer fix. The mode keeps reading the file as before, now with a better key available.

## Backward compatibility

- **Additive / trailing.** The column is appended after the existing fingerprint (col 8), postedAt (col 9) and trust (cols 10-11) columns. No existing column is reordered.
- **Web parser unaffected.** Every `scan-history.tsv` reader in `web/` (`readScanDates`, `discover`, `whats-new`, explore/add) splits by `\t` and indexes only cols 0-6, so a trailing column is invisible to them (verified). The core readers (`loadSeenUrls`, `loadFingerprintHistory`, `parseScanHistory`) are likewise index-based.
- **Old rows tolerated.** Rows written before this column simply have fewer fields; `companyKey` normalizes their raw company on the fly, so old and new rows key identically and cluster together — nothing depends on the writer having run.

## Notes for review

- Two normalizers exist: `normalizeCompany` (tracker-utils, alphanumeric-only — would leave `acmeinc` ≠ `acme`) and `normalizeCompanyName` (`invite-match.mjs`, legal-suffix aware). Only the latter delivers the maintainer's stated example (`Acme Inc.` == `Acme`), so it is the correct canonical reuse. If you'd prefer that function to live in a neutral shared module rather than `invite-match.mjs`, that's a trivial relocation follow-up.

## Validation

- `node detect-reposts.mjs --self-test` → 8 passed, 0 failed
- `node detect-reposts.test.mjs` → 213 passed, 0 failed (new: `companyKey` unit cases, stored-column clustering, raw-normalize fallback, backward-compat legacy rows)
- `node test-all.mjs` → **2281 passed, 0 failed**, 2 warnings (benign env: cv-sync-check without user data / CV section order)

## Summary by CodeRabbit

- **New Features**
  - Added a trailing `normalized_company` field to scan history records for more consistent company matching and repost clustering, including support for older files missing it.
  - When creating a new scan history file, the widened header now includes the new column.
- **Bug Fixes**
  - Improved fingerprint history loading by ignoring the header row when present.
- **Documentation**
  - Updated the scan-history data contract and script documentation to describe the append-only, trailing-column format.
- **Tests**
  - Expanded scan-history and repost-detection tests to cover the 12-column format and normalization behavior.

### DIFF
--- a/DATA_CONTRACT.md
+++ b/DATA_CONTRACT.md
@@ -26,7 +26,7 @@ These files contain your personal data, customizations, and work product. Update
 | `data/applications.md` | Your application tracker (source of truth) |
 | `data/applications.db` | Derived query index over `applications.md` (SQLite, rebuilt by `node tracker.mjs sync` — safe to delete) |
 | `data/pipeline.md` | Your URL inbox |
-| `data/scan-history.tsv` | Your scan history (9 tab-separated columns; col 8: local SimHash JD fingerprint for cross-listing detection, col 9: posting date) |
+| `data/scan-history.tsv` | Your scan history (tab-separated, append-only trailing columns; col 8: local SimHash JD fingerprint for cross-listing detection, col 9: posting date, cols 10-11: trust score/flags, col 12: normalized company key for repost/name matching). Older rows may have fewer columns — readers index by position and tolerate the absence. |
 | `data/scan-runs.tsv` | Your per-run scan counters (appended by `scan.mjs`, read by `stats.mjs`) |
 | `data/portal-health.tsv` | Consecutive reachability status for scanned portals (appended by `scan.mjs`) |
 | `data/follow-ups.md` | Your follow-up history |
--- a/detect-reposts.mjs
+++ b/detect-reposts.mjs
@@ -26,6 +26,7 @@ import { join, dirname } from 'path';
 import { fileURLToPath, pathToFileURL } from 'url';
 
 import { roleFuzzyMatch } from './role-matcher.mjs';
+import { normalizeCompanyName } from './invite-match.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
 const SCAN_HISTORY_PATH = join(CAREER_OPS, 'data/scan-history.tsv');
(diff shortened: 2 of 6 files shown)


santifer:
Merged, @Scott-Emberson. Same day as #1712, which is how a held-back hunk is supposed to end.

This is the version that doesn't cost anything to be wrong about. The repost signal keeps reading a file rather than depending on a script having run, `Acme Inc.` and `Acme` stop being two companies, and everything that reads `scan-history.tsv` gets the benefit at once: the mode, the web, `detect-reposts`. And writing the key at write time rather than deriving it at read time means the readers don't each need their own copy of the normalizer, which is where that kind of thing usually drifts.

The line I most wanted to see is in the data contract: older rows carry fewer columns and readers tolerate the absence, stated explicitly. Additive is only actually additive if someone writes down what happens to the files that already exist.

PR #2050, opened 2026-07-19
fix(tracker): recognize the Hired canonical status in merge-tracker and stats

vibhusings:
## What does this PR do?

Teaches two readers about the `Hired` canonical status. `Hired` was added to
`templates/states.yml` (and to the dashboard, `normalize-statuses.mjs`,
`verify-pipeline.mjs`, and `set-status.mjs`), but `merge-tracker.mjs` and
`stats.mjs` still hardcoded the pre-`Hired` 8-state list — so a landed-job row
was silently mishandled by both. This aligns them with the source of truth.

Concretely, before this fix:
- `merge-tracker.mjs` `validateStatus("Hired")` matched nothing and fell
  through to the default, **silently rewriting the row to `Evaluated`** (data
  loss on the best possible outcome).
- `stats.mjs` classified `Hired` as `Unknown`, **dropping it from the
  cumulative funnel** (`everApplied/Responded/Interview/Offer`) and from
  `avgScoreApplied`.

Changes:
- **`merge-tracker.mjs`**: add `Hired` to `CANONICAL_STATES` plus the
  `states.yml` aliases (`contratado`/`contratada`/`accepted`/`accept`).
- **`stats.mjs`**: add `Hired` to `CANONICAL_STATUSES` and `PURSUED_STATUSES`,
  and count it through `everApplied/Responded/Interview/Offer` in
  `computeFunnel` (a landed job proves the offer and every stage before it).
  `Hired` stays out of `ACTIVE_STATUSES` — it's a terminal success, not an
  in-flight application.
- **Tests**: extend `tests/stats.test.mjs` and add `tests/merge-tracker.test.mjs`
  (a CLI integration test using the existing `CAREER_OPS_TRACKER` /
  `CAREER_OPS_ADDITIONS` env overrides) so this can't silently regress.

## Related issue

None — this is a bug fix, which `CONTRIBUTING.md` exempts from issue-first.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation / translation
- [ ] Refactor (no behavior change)

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/santifer/career-ops/blob/main/CONTRIBUTING.md)
- [x] If this is a new feature or architecture change, I opened an issue first (bug fixes, providers, docs & translations are exempt — send those straight in)
- [x] My PR does not include personal data (CV, email, real names, scan results, or pipeline data)
- [x] I ran `node test-all.mjs` and all tests pass
[...]

PR #2250, opened 2026-07-28
fix(web): propagate the Hired terminal-success state across the whole dashboard

### LINKED ISSUE #2249: web: the Hired terminal-success state is missing across the whole dashboard (6 hardcoded status lists)
## Summary

`Hired` — the terminal success state (offer accepted, landed the job), added to the core in #2050 — never made it into the web dashboard. The web's status model in `web/src/lib/format.ts` predates it, so on the dashboard:

1. **You can't set a role to Hired.** `CANONICAL_STATES` is the single source for the status dropdown in `status-select.tsx` (it `.map`s over it to render the options, and uses it for the `known` check). `Hired` isn't in the array, so it's never offered — a user who actually lands a job cannot record it from the web UI.
2. **An existing Hired row shows as "unknown."** `status-select.tsx:39` does `(CANONICAL_STATES as readonly string[]).includes(status)` → `false` for Hired, so it's treated as an unrecognized status.
3. **The dot is gray.** `statusDot("Hired")` falls through every branch to the `bg-zinc-500` default — the same neutral gray as `Evaluated`. The best possible outcome renders identically to "not looked at yet."
4. **The Spanish/other aliases don't canonicalize.** The core `states.yml` lists `aliases: [contratado, contratada, hired, accepted, accept]`; none are in the web `STATUS_ALIAS` map, so `canonStatus("contratado")` → `"CONTRATADO"` (gray, unrecognized) instead of `HIRED`.

## Reproduce (the dot logic, no server needed)

```js
// mirrors web/src/lib/format.ts
const STATUS_ALIAS = {};
const canonStatus = s => { const k=s.trim().toLowerCase(); if(k===''||k==='—'||k==='-')return 'DISCARDED'; return STATUS_ALIAS[k]??s.toUpperCase(); };
const statusDot = s => { const c=canonStatus(s);
  if(c.includes('INTERVIEW')||c.includes('OFFER'))return 'green';
  if(c.includes('APPLIED')||c.includes('RESPONDED'))return 'blue';
  if(c.includes('REJECTED')||c.includes('SKIP'))return 'red';
  if(c.includes('DISCARDED'))return 'gray-600';
  return 'gray-500 (Evaluated/unknown)'; };
console.log('Hired ->', statusDot('Hired'));  // gray-500 (Evaluated/unknown)  ← bug
console.log('Offer ->', statusDot('Offer'));  // green
```

## Proposed fix (contained to `format.ts`)

Because `CANONICAL_STATES` is the one source, a single-file change fixes the dropdown, the recognition, and the dot together:

- add `"Hired"` to `CANONICAL_STATES` (after `Offer`, matching `states.yml` order);
- add the aliases to `STATUS_ALIAS` (`contratado`/`contratada`/`hired`/`accepted`/`accept` → `HIRED`);
- give `HIRED` a positive color in `statusDot` — green (it's the best outcome; matching the Interview/Offer positive convention is the minimal mirror, though a distinct celebratory shade would be defensible).

PR opening now.

## Related (Go dashboard, separate)

The Go TUI has the mirror gap: `dashboard/.../pipeline.go`'s `statusColorMap()` also omits `"hired"`, even though `NormalizeStatus` maps `Hired`→`"hired"` and it's a recognized status — so `hired` is the one status rendered with no color in the pipeline list. Filing that separately since it's a different codebase (and I can't build-test Go in my environment).

abankar1:
Closes #2249.

`Hired` — offer accepted, landed the job — was added to the core `states.yml` in #2050 but never reached the web dashboard, so the single most important outcome the tool exists for was invisible in the UI.

## What was broken (`web/src/lib/format.ts`)

- **You couldn't set it.** `CANONICAL_STATES` is the one source for the `status-select.tsx` dropdown (it `.map`s over it to render the options) and its `known` check. `Hired` wasn't in it — so a user who lands a job couldn't record it, and an existing Hired row read as an unrecognized status.
- **Gray dot.** `statusDot("Hired")` fell through to the `bg-zinc-500` neutral — the best outcome looked identical to "Evaluated".
- **Aliases didn't canonicalize.** `states.yml` lists `contratado`/`contratada`/`accepted`/`accept`; none were in the web `STATUS_ALIAS`.

## The fix — one file, three coordinated edits

- add `"Hired"` to `CANONICAL_STATES` (after `Offer`, matching `states.yml` order) → dropdown + known-check;
- map the aliases (`contratado`/`contratada`/`accepted`/`accept` → `HIRED`);
- give `HIRED` the positive green dot alongside `Interview`/`Offer` (it's the terminal win — never gray).

Verified by simulating the exact `canonStatus`/`statusDot` logic:

```
Hired       canon=HIRED       dot=emerald(green)
contratado  canon=HIRED       dot=emerald(green)
accepted    canon=HIRED       dot=emerald(green)
Offer       canon=OFFER       dot=emerald(green)
Evaluated   canon=EVALUATED   dot=zinc-500(gray)   ← unchanged
```

## Drift guard (commit 2)

There was a check that `states.yml` keeps its ids, but nothing asserting the **web list covers them** — which is exactly how this slipped. Added a cross-check in `test-all.mjs`: every `states.yml` label must appear in `format.ts`'s `CANONICAL_STATES`. Confirmed it goes red if `Hired` is removed, so a future core state can't silently go unrecognized by the web again.

`node test-all.mjs --quick` → **2308 passed, 0 failed**.

2 commits (fix + guard).

> Note: the Go TUI has the mirror gap (`statusColorMap()` omits `"hired"`), filed separately since it's a different codebase and I can't build-test Go in my environment.

## Summary by CodeRabbit

* **New Features**
  * Added support for the new **“Hired”** pipeline status across registry status/tab handling, pipeline view tabs, and analytics “Pipeline by stage”.
  * Expanded hired alias recognition (e.g., “contratado,” “contratada,” “accepted,” “accept”) and updated status display so **HIRED** is treated as a primary success state.
* **Documentation**
  * Updated the assistant API documentation to reflect valid `tab` values and the `min` score-floor range, and included **Hired** in canonical status values.
* **Tests**
  * Strengthened drift checks to ensure web hardcoded status lists and assistant prose remain aligned with the canonical state label set.

### DIFF
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -9536,6 +9536,29 @@ try {
     } else {
       fail(`web status list(s) missing canonical state(s) — dashboard can't set/count them (#2249): ${drift.join(' | ')}`);
     }
+
+    // The assistant preamble also enumerates the states in PROSE (the setStatus
+    // list + the filterPipeline tab enum). Those drift the same way — the AI
+    // couldn't offer to set/filter by Hired — so check them too (#2249).
+    const assistantPath = join(ROOT, 'web', 'src', 'app', 'api', 'assistant', 'route.ts');
+    if (existsSync(assistantPath)) {
+      const src = readFileSync(assistantPath, 'utf-8');
+      const proseChecks = [
+        { name: 'setStatus canonical-states list', text: src.match(/Canonical states:\s*([^.]*)\./)?.[1] ?? '', upper: false },
+        { name: 'filterPipeline tab enum', text: src.match(/tab ∈\s*([^;]*);/)?.[1] ?? '', upper: true },
+      ];
+      const proseDrift = [];
+      for (const { name, text, upper } of proseChecks) {
+        const want = upper ? stateLabels.map((l) => l.toUpperCase()) : stateLabels;
+        const missing = want.filter((l) => !new RegExp(`\\b${l}\\b`).test(text));
+        if (missing.length) proseDrift.push(`${name} (${missing.join(', ')})`);
+      }
+      if (proseDrift.length === 0) {
+        pass('assistant preamble prose enumerates every canonical state (#2249)');
+      } else {
+        fail(`assistant preamble missing canonical state(s) in prose (#2249): ${proseDrift.join(' | ')}`);
+      }
+    }
   }
 
   // 55.4 report format blocks (modes/oferta.md → web report parser)
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -9505,6 +9505,25 @@ try {
     fail(`templates/states.yml lost canonical status id(s): ${missingStates.join(', ')} — BREAKING for the web status mapping`);
   }
 
+  // 55.3b The web status dropdown must offer EVERY canonical state. states.yml is
+  // the source of truth; web/src/lib/format.ts's CANONICAL_STATES is the single
+  // list the status-select dropdown + its known-check read. `Hired` (#2050)
+  // silently drifted out of the web list — a landed job couldn't be recorded and
+  // rendered as a gray "unknown" dot (#2249). Keep the two in lockstep.
+  const formatPath = join(ROOT, 'web', 'src', 'lib', 'format.ts');
+  if (existsSync(formatPath)) {
+    const stateLabels = [...statesSrc.matchAll(/^\s+label:\s*"?([A-Za-z]+)"?\s*$/gm)].map((m) => m[1]);
+    const formatSrc = readFileSync(formatPath, 'utf-8');
+    const webBlock = formatSrc.match(/CANONICAL_STATES\s*=\s*\[([\s\S]*?)\]/)?.[1] ?? '';
+    const webStates = new Set([...webBlock.matchAll(/"([A-Za-z]+)"/g)].map((m) => m[1]));
+    const webMissing = stateLabels.filter((l) => !webStates.has(l));
+    if (stateLabels.length > 0 && webMissing.length === 0) {
+      pass('web CANONICAL_STATES covers every canonical state label from states.yml (#2249)');
+    } else {
+      fail(`web/src/lib/format.ts CANONICAL_STATES is missing canonical state(s): ${webMissing.join(', ')} — the web dashboard can't set or recognize them (#2249)`);
+    }
+  }
+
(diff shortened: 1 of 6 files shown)


abankar1:
Expanded — while fixing `format.ts` I checked the rest of the dashboard, and **`HIRED` appeared nowhere in `web/src` (0 refs)**. The state added to the core in #2050 had drifted out of *six* hardcoded lists, so the terminal-success outcome was invisible across the whole web UI:

| Surface | Was |
|---|---|
| `format.ts` `CANONICAL_STATES` | not settable in the status dropdown; gray dot |
| `pipeline-view` `TABS` | no Hired filter tab — hired rows only under ALL |
| `registry.ts` `CANON_STATUS` | a *second* canonical-states copy, also missing it |
| `registry.ts` `TAB_VALUES` | the tab allowlist the assistant/actions validate against |
| `analytics` `STAGES` | **landing a job didn't show in the funnel at all** |
| `assistant/route.ts` | the AI couldn't `filterPipeline` by Hired |

Added Hired to each (after Offer, matching `states.yml` order; analytics counts via `canonStatus().includes()` so the new stage auto-populates). The drift guard now cross-checks **all six** lists against `states.yml` — not just `format.ts` — and the analytics exclusion for `SKIP` (not a funnel stage) is encoded, so this whole class can't silently recur.

4 step commits (format.ts fix → guard → propagate the other five → broaden the guard). `test-all.mjs --quick` → 2308/0.

abankar1:
One more (a **seventh** surface): the assistant preamble also lists the states in *prose* — the `setStatus` action's "Canonical states: …" line was missing Hired (so the AI wouldn't offer to mark a job hired). Fixed it, and the drift guard now checks both prose enumerations (setStatus list + filterPipeline tab enum) against `states.yml` too — so all seven web surfaces are covered and can't silently drop a future state. 2309/0.

santifer:
Thorough one, @abankar1. Six hardcoded status lists plus the assistant's prose enumeration, and your 55.3b check cross-checks every one of them against states.yml so the tenth state can't vanish like Hired did. career-ops-ui verified it breaks nothing on their side, and their follow-up (#2282) closes the same gap in the degraded-path fallback your check doesn't reach. Merged! 🚀

Issue #2229, opened 2026-07-28
The in-process DNS cache can't help per-tenant-hostname providers — a full sweep still floods the resolver


bigguy6883:
Follow-up to #2154 / #2155. That pair fixed *many connections → one hostname*. This is the other axis: *one connection each → many hostnames*, which no cache can collapse.

## Summary

#2155 works exactly as designed — I re-measured it in the field and each hostname resolved **once**, not once per worker. But `workday` and (incoming, #2141) `icims` address every tenant on **its own hostname**, so a full sweep has nothing to memoize. The lookups are all genuinely distinct, and `scan-ats-full.mjs` issues them at `CONCURRENCY = 20` with no pacing, which still trips a per-client resolver limit.

The cache is ~100% effective on three providers and structurally inert on the two largest:

| provider | boards | distinct hostnames | upstream queries (A+AAAA) |
|---|---:|---:|---:|
| greenhouse | 8,333 | **1** | 2 |
| lever | 4,368 | **1** | 2 |
| ashby | 3,161 | **1** | 2 |
| **workday** | 12,884 | **3,781** | 7,562 |
| **icims** (#2141) | 10,108 | **10,108** | 20,216 |
| | | **13,892** | **27,784** |

On `main` today that's **7,568 irreducible upstream queries** per full sweep. With #2141 merged it becomes **27,784 — a 3.7× increase**, which is why I'm raising this before that PR lands rather than after.

For reference, a stock Pi-hole allows 1,000 queries/min per client.

## Reproduce (offline, no network, no DNS tooling)

This uses the repo's own cached datasets and the same `toEntry` mapping `scan-ats-full.mjs` uses, so it counts exactly what a sweep would resolve. (`data/cache/ats-companies/*.json` is populated by any `scan-ats-full.mjs` run; the datasets themselves come from `DATASET_BASE` in `scan-ats-full.mjs:61`.)

```bash
node --input-type=module -e "
import { readFileSync } from 'node:fs';
const SLUG_RE = /^[A-Za-z0-9._-]+$/;                 // scan-ats-full.mjs:133
const read = p => JSON.parse(readFileSync('data/cache/ats-companies/'+p,'utf8'));
const wd = read('workday.json');
const wdHosts = new Set(wd.map(e => { const [t,i] = String(e).split('|'); return \`\${t}.\${i}.myworkdayjobs.com\`; }));
const icHosts = new Set(read('icims.json').filter(s => SLUG_RE.test(String(s)))
[...]

PR #2266, opened 2026-07-29
Negative-cache resolver refusals and stop a sweep when DNS starts failing

bigguy6883:
# Stop a rate-limited resolver from being amplified into a self-sustaining DNS flood

Implements #2229, **proposed fixes 1 and 3**. Fix 2 (token-bucket pacing) is #2264 — separate, as offered in the issue, and neither depends on the other.

## The failure this fixes

A refusing resolver fails a lookup in about a millisecond. The in-process cache from #2155 deliberately never caches failures, so under a *sustained* refusal that inverts: 20 workers retry a refusal that costs the resolver another query to reject, forever. Measured during the incident behind #2229 — a Pi-hole logged 4,637–6,152 attempted queries/minute for 39 minutes against a universe of 3,784 hostnames, of which only 1,362 ever got through.

The sweep meanwhile reports thousands of `fetch failed` lines that look like dead boards. The diagnosis is wrong in the most expensive direction: the boards are fine, the resolver is refusing.

Two changes, sharing one definition:

1. **Negative-cache resolver refusals** for 30s, which breaks the retry loop.
2. **Circuit-break the sweep** after 50 consecutive resolver-level failures, so it stops and says what's actually wrong.

## One predicate, two consumers

`isResolverFailure` is exported from `providers/_dns-cache.mjs` and used by both the cache and the sweep, so the two can't drift on what counts as a resolver failure. It walks the `cause` chain, because Node's `fetch()` reports every transport error as `TypeError: fetch failed` with the real error hung off `.cause`. The walk is depth-bounded so a self-referential cause can't spin.

**Naming correction vs. the issue text:** #2229 says "REFUSED / SERVFAIL". The actual runtime values are `E`-prefixed — `dns.REFUSED === 'EREFUSED'`. The code uses the runtime values; flagging it so the mismatch with the issue doesn't read as a bug.

Covered: `ENOTIMP` (what a rate-limiting Pi-hole actually replies with), `EREFUSED`, `ESERVFAIL`, `EAI_AGAIN`.

**`ENOTFOUND` is deliberately excluded.** It is NXDOMAIN — "this tenant does not exist" — a legitimate per-host answer, not a resolver refusal. Caching it would pin a tenant as nonexistent for the window even after it appears. The pre-existing assertion `a failed resolution is not cached` uses `ENOTFOUND` and is **unmodified** by this PR; it is the guardrail against ever conflating the two categories, and it still passes.

**`ETIMEOUT` is also excluded.** A query timeout is ambiguous between a slow resolver and a refusing one, and #2229 scoped this to refusals. Noted in a code comment so it reads as a decision rather than an oversight.

## The breaker

`parallelEach` gains an optional `shouldStop` predicate, checked before a worker claims its next index. When it flips, workers stop pulling work and the call resolves.

**This is a 5th parameter, not a 4th.** `parallelEach` already carries `onItemDone` in position 4, driving checkpoint/`--resume`. Appending `shouldStop` after it leaves that contract untouched, and two of the new tests pin it down: `onItemDone` still fires once per item with a monotonic `done`, and a *stopped* run still reports a usable `resumeAt` — everything below it finished, so resuming there loses no work. Checking the predicate before claiming an index is what keeps `resumeAt` truthful.

Two things the breaker deliberately does **not** do when it trips:

- It does not run the sequential retry pass for truncated boards. Retrying them one at a time is more of exactly the traffic the breaker just stopped.
- It does not add the source to `completedSources` or write a "finished" checkpoint. The source did not finish; marking it done would make `--resume` skip the rest of it.

Output on trip:

```
  ⛔ stopped workday: 50 consecutive DNS failures.
     Your resolver is refusing queries — it may be rate-limiting this host.
     Lower CONCURRENCY, raise the resolver's per-client limit, or set
     CAREER_OPS_NO_DNS_CACHE=1 only if you know the cache is at fault.
```

A dead board and a dead resolver look identical one at a time — only the *consecutive* run tells them apart, so any non-resolver outcome (including a successful fetch) resets the counter. 50 is high enough that a handful of unlucky boards can't trip it and low enough to stop within seconds of a real outage.

## Verification

`node test-all.mjs --quick`:

- Baseline at `06da05c6` before any change: **2343 passed, 0 failed, 2 warnings**
- After: **2349 passed, 0 failed, 2 warnings** (6 new assertions)

The 2 warnings are pre-existing and unrelated (`workday: InvalidCo truncated at max_pages=100`).

Beyond the unit tests, the wiring was exercised against a real sweep:

- `--ats workday --limit 40 --since 1 --dry-run` completes normally, 40/40 scanned with 3 unreachable boards — non-resolver failures correctly do **not** trip the breaker.
- With `RESOLVER_FAILURE_LIMIT` temporarily lowered and the predicate forced true, the same sweep aborts early and prints the diagnosis above. That patch was reverted; it exists only as evidence the wired path fires, since the counting lives inside `main()` and isn't unit-reachable.

No new dependencies; `package.json` untouched.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of DNS resolver failures by temporarily caching eligible failures and retrying after the configured interval.
  * Prevented unnecessary repeated DNS requests during resolver outages.
  * Reverse ATS scans now stop early when widespread resolver failures are detected, avoiding extended unsuccessful scans.

* **Reliability**
  * Scans preserve progress and can resume consistently after an outage-related interruption.

### DIFF
--- a/providers/_dns-cache.mjs
+++ b/providers/_dns-cache.mjs
@@ -83,6 +83,10 @@ export function isResolverFailure(err) {
 
 const DEFAULT_TTL_MS = 5 * 60_000;
 const DEFAULT_MAX_ENTRIES = 512;
+// Short by design: long enough to break the retry storm that a refusing
+// resolver otherwise feeds (see #2229), short enough that a resolver coming
+// back is picked up almost immediately.
+const DEFAULT_NEGATIVE_TTL_MS = 30_000;
 
 /**
  * Build a caching wrapper around a callback-style `dns.lookup`.
@@ -99,6 +103,7 @@ const DEFAULT_MAX_ENTRIES = 512;
  */
 export function createCachedLookup(realLookup, options = {}) {
   const ttlMs = options.ttlMs ?? DEFAULT_TTL_MS;
+  const negativeTtlMs = options.negativeTtlMs ?? DEFAULT_NEGATIVE_TTL_MS;
   const maxEntries = options.maxEntries ?? DEFAULT_MAX_ENTRIES;
   const now = options.now ?? Date.now;
 
@@ -139,13 +144,18 @@ export function createCachedLookup(realLookup, options = {}) {
       const callbacks = inflight.get(key) ?? [];
       inflight.delete(key);
 
-      // Never cache failures — a transient resolver blip must not be pinned
-      // in for the whole TTL window.
-      if (!err) {
+      // Successes cache for the full TTL. Resolver-level refusals cache for a
+      // much shorter one: not caching them at all is what turns a rate-limited
+      // resolver into a self-sustaining flood, because every worker retries a
+      // refusal that costs the resolver another query to reject (#2229).
+      // Everything else — NXDOMAIN above all — stays uncached as before, so a
+      // transient blip is never pinned in for the whole TTL window.
+      const ttl = err ? (isResolverFailure(err) ? negativeTtlMs : 0) : ttlMs;
+      if (ttl > 0) {
         // Oldest-first eviction; insertion order is good enough for a cap
         // this size, and avoids tracking per-entry access times.
         if (cache.size >= maxEntries) cache.delete(cache.keys().next().value);
-        cache.set(key, { expires: now() + ttlMs, args: [null, ...rest] });
+        cache.set(key, { expires: now() + ttl, args: err ? [err] : [null, ...rest] });
       }
 
       for (const cb of callbacks) cb(err, ...rest);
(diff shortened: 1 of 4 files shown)


santifer:
Another sharp one, @bigguy6883. Scoping the negative cache to resolver-level codes (keeping NXDOMAIN uncached so late tenants still appear), walking fetch's cause chain with a depth bound, and telling a dead board from a dead resolver by the consecutive run: all three distinctions are exactly right, and the tests pin them. Merged! 🚀

Issue #1051, opened 2026-06-15
Windows: skill symlink entrypoints checkout as pointer files


ncoprod:
### Code of Conduct

- [x] I agree to follow the [Code of Conduct](https://github.com/santifer/career-ops/blob/main/CODE_OF_CONDUCT.md)

### Existing issues

- [x] I searched existing issues and this has not been reported yet

### What happened?

On Windows checkouts without symlink support, the multi-CLI skill entrypoints can become regular text files instead of usable skill files.

On current `upstream/main`, git tracks the Claude/OpenCode entrypoints as symlinks:

```text
120000 ... .claude/skills/career-ops/SKILL.md
120000 ... .opencode/skills/career-ops/SKILL.md
```

But on my Windows checkout, Node sees both paths as regular 43-byte files:

```text
.claude/skills/career-ops/SKILL.md { isSymbolicLink: false, size: 43 }
.opencode/skills/career-ops/SKILL.md { isSymbolicLink: false, size: 43 }
.agents/skills/career-ops/SKILL.md { isSymbolicLink: false, size: 4193 }
```

The file content is only the symlink target text:

```text
../../../.agents/skills/career-ops/SKILL.md
```

So a candidate on Windows can end up with Claude/OpenCode skill entrypoints that are pointer text files, not real `SKILL.md` router files. That can make `/career-ops` unavailable or inconsistent across CLIs.

This is related to the investigation in #763/#769, but not the frontmatter diagnosis. It also avoids the #762 failure mode: `.agents/skills/career-ops/SKILL.md` should stay the canonical edit target instead of hand-editing `.claude`.

### Steps to reproduce

1. On Windows, check out current `main` without symlink support.
2. Inspect the tracked entrypoints:
   ```bash
   git ls-files -s .claude/skills/career-ops/SKILL.md .opencode/skills/career-ops/SKILL.md
   ```
3. Inspect what Node sees:
   ```bash
   node -e "const fs=require('fs'); for (const p of ['.claude/skills/career-ops/SKILL.md','.opencode/skills/career-ops/SKILL.md','.agents/skills/career-ops/SKILL.md']) { const s=fs.lstatSync(p); console.log(p, {isSymbolicLink:s.isSymbolicLink(), size:s.size}); }"
   ```
4. Read one CLI entrypoint:
   ```powershell
   Get-Content .claude\skills\career-ops\SKILL.md
   ```
5. Run the integrity check:
   ```bash
   node test-all.mjs --quick
   ```

[...]

PR #2273, opened 2026-07-29
fix(test): accept a materialized skill entrypoint, not just a symlink

### LINKED ISSUE #2272: tests: the index-mode entrypoint check contradicts the materialization fallback — every Windows user has a permanently red suite after `update-system.mjs apply`
### Code of Conduct

### Existing issues

- [x] I searched existing issues and this hasn't been reported yet — closest are #1051 (the original pointer-file problem, which materialization fixed), #2259 (which added the fallback), and #2269 (a different bug in the 12c *fixture*, about global `core.excludesFile`). None cover the contradiction below.

### What happened?

On a filesystem without symlink support, `update-system.mjs apply` materializes the seven per-CLI `SKILL.md` entrypoints as regular files. That is deliberate — 06da05c / #2259 added it precisely so these users get a working skill instead of a pointer file (#1051).

The updater then commits them at mode `100644`. `test-all.mjs` asserts the index mode is `120000`:

```js
} else if (staged.startsWith('120000')) {
  pass(`Entrypoint is a real symlink in git: ${f}`);
} else {
  fail(`Entrypoint committed as a REGULAR file (mode ${staged.split(' ')[0]}) — users of this CLI get a broken skill: ${f}`);
}
```

So the documented fallback produces exactly the state the test rejects. **7 permanent failures after any successful update on Windows**, with no user error involved.

The suite already contradicts itself on this. Section `12c. Materialized skill index mode` passes, in the same run:

```
✅ materialized skill entrypoints stage as regular files, not symlink blobs
✅ materialized skill blobs contain canonical skill content
```

12c blesses mode `100644` + canonical content. The check above rejects mode `100644` unconditionally.

### Why the mode check is the wrong proxy

Its own comment states the real invariant:

> A regular-file blob whose content is the link path as text ships a broken, empty skill to every user of that CLI

That is a **content** condition. Mode was an accurate proxy for it until #2259 introduced a second, legitimate `100644` state. There are now two distinct cases the check cannot tell apart:

| Mode | Content | Reality |
|---|---|---|
| `100644` | link path as text (`../../../.agents/...`) | broken — the #1051 bug |
| `100644` | canonical `SKILL.md` body | correct — the #2259 fallback |
| `120000` | symlink | correct on symlink-capable filesystems |

Only the first is a defect. On my checkout all seven entrypoints hash to the canonical blob:

```
$ git hash-object .agents/skills/career-ops/SKILL.md
ac2892762eb85a032aacfabc51e923167489b988

$ for f in .claude .cursor .opencode .qwen .antigravitycli .grok .kimi; do \
    git hash-object "$f/skills/career-ops/SKILL.md"; done
ac2892762eb85a032aacfabc51e923167489b988   (x7)
```

Identical to canonical — working entrypoints, flagged as broken.

### Steps to reproduce

1. On Windows without Developer Mode (so `git config core.symlinks` is `false`), clone and run `node update-system.mjs apply`.
2. Observe `Materialized 7 skill entrypoint(s) for filesystems without symlink support`.
3. `node test-all.mjs`

```
❌ Entrypoint committed as a REGULAR file (mode 100644) — users of this CLI get a broken skill: .claude/skills/career-ops/SKILL.md
   … ×7
📊 Results: 2356 passed, 7 failed
```

`git ls-tree origin/main -- .claude/skills/career-ops/SKILL.md` → `120000`; the same path locally → `100644`, changed by the `chore: auto-update system files to v1.23.0` commit.

CI never sees this: `.github/workflows/test.yml` runs on ubuntu, where symlinks work and the entrypoints stay `120000`. (#1761 proposed adding Windows to the matrix and was closed.)

### Expected behavior

An entrypoint should pass when it is *either* a symlink *or* a regular file whose blob equals the canonical `.agents/skills/career-ops/SKILL.md` — i.e. assert the invariant the comment describes, rather than the proxy:

```js
const [mode, sha] = staged.split(/\s+/);
const canonical = run('git', ['rev-parse', ':.agents/skills/career-ops/SKILL.md']);
if (mode === '120000') {
  pass(`Entrypoint is a real symlink in git: ${f}`);
} else if (sha === canonical) {
  pass(`Entrypoint is a materialized regular file with canonical content: ${f}`);
} else {
  fail(`Entrypoint committed as a REGULAR file with non-canonical content — users of this CLI get a broken skill: ${f}`);
}
```

That still catches #1051 (a link-path blob never equals the canonical blob) while accepting what #2259 deliberately produces.

I'm happy to send the PR if you'd like it in this shape — or, if you'd rather the updater restore `120000` on these filesystems instead, say so and I'll leave the test alone. Worth noting the tradeoff: with `core.symlinks=false` a `120000` entry checks out as a text file containing the target path, and Claude Code then reads that path as the skill description — the skill silently becomes `career-ops: ../../../.agents/skills/career-ops/SKILL.md`. I hit exactly that when checking out an unmaterialized branch, which is why I think the fallback is right and the assertion is what needs to move.

### CLI tool

Claude Code

### OS and Node.js version

Windows 11, Node v24.18.0, git 2.53.0.windows.3, `core.symlinks=false`

### Output of npm run doctor

`{"onboardingNeeded":false,"missing":[],"warnings":["Playwright MCP tools not detected"],"autoCopied":[]}` — career-ops v1.23.0

LayerC0de:
## What does this PR do?

The per-CLI `SKILL.md` entrypoint check asserts index mode `120000`. On a filesystem without symlink support, `update-system.mjs apply` materializes those entrypoints as regular files — by design; #2259 added exactly that so these users get a working skill instead of a pointer file (#1051) — and commits them at `100644`. **The documented fallback produces the state the test rejects: 7 permanent failures after any successful update, no user error involved.** CI never sees it, running only on ubuntu where symlinks work.

This compares the blob against the canonical entrypoint instead of asserting the mode.

## Related issue

Fixes #2272

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

---

### The suite already contradicts itself

Section `12c. Materialized skill index mode` passes in the same run that produces the 7 failures:

```
✅ materialized skill entrypoints stage as regular files, not symlink blobs
✅ materialized skill blobs contain canonical skill content
```

12c blesses mode `100644` + canonical content; the entrypoint check rejects `100644` unconditionally.

### Why mode is the wrong proxy now

The check's own comment states the real invariant — *"a regular-file blob whose content is the link path as text ships a broken, empty skill"*. That is a **content** condition. Mode was faithful to it until #2259 introduced a second, correct `100644` state:

| Mode | Blob | Reality |
|---|---|---|
| `120000` | symlink | correct |
| `100644` | canonical | materialized entrypoint — correct (#2259) |
| `100644` | anything else | link-path text or stale copy — **broken** (#1051) |

Comparing against the canonical blob separates them and still catches #1051: a link-path blob never equals the canonical blob.

### Verified in all four states

| State | Before | After |
|---|---|---|
| symlink checkout (ubuntu-equivalent) | 7 pass | 7 pass |
| materialized checkout after a real Windows `apply` | **7 fail** | 7 pass |
| #1051 link-path-as-text blob | fail | fail |
| stale non-canonical copy | *pass* ⚠️ | fail |

The last row is a small bonus: once a stale copy had the right mode, the old check waved it through. Content comparison does not.

Fixture used for rows 3–4 — a link-path blob and a stale copy, both at `100644`:

```
FAIL  regular, non-canon  .grok/skills/career-ops/SKILL.md (mode 100644)
FAIL  regular, non-canon  .kimi/skills/career-ops/SKILL.md (mode 100644)
```

`node test-all.mjs` on this branch: **2343 passed, 0 failed.**

### Note on the alternative

If you would rather the updater restore `120000` on these filesystems and leave the test alone, this PR is the wrong shape and I will close it. Worth weighing first: with `core.symlinks=false` a `120000` entry checks out as a text file containing the target path, and Claude Code then reads that path as the skill description — the skill silently becomes `career-ops: ../../../.agents/skills/career-ops/SKILL.md`. I hit that firsthand checking out an unmaterialized branch on Windows, which is why I think the fallback is right and the assertion is what needs to move.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of skill entrypoints to detect broken links and incorrect copied content.
  * Tests now confirm that each CLI entrypoint resolves to the canonical skill content.

### DIFF
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -914,21 +914,50 @@ for (const f of systemFiles) {
   }
 }
 
-// Per-CLI SKILL.md entrypoints must be SYMLINKS in the git index (mode 120000).
-// A regular-file blob whose content is the link path as text ships a broken,
-// empty skill to every user of that CLI — exactly what happened to .kimi/ when
-// a symlink was created under core.symlinks=false and committed as-is. Checking
-// the INDEX mode (not the filesystem) keeps this assertion true on Windows
-// checkouts too.
+// Per-CLI SKILL.md entrypoints must resolve to the canonical skill content.
+//
+// The defect this guards is a regular-file blob whose content is the LINK PATH
+// AS TEXT — exactly what happened to .kimi/ when a symlink was created under
+// core.symlinks=false and committed as-is (#1051). That ships a broken, empty
+// skill to every user of that CLI.
+//
+// Index mode was a faithful proxy for that until #2259 added
+// materializeSkillEntrypoints(), which writes the real content as a regular
+// file on filesystems without symlink support. That is a second, CORRECT
+// mode-100644 state, so mode alone can no longer tell the two apart:
+//
+//   120000                          → symlink                        (correct)
+//   100644 + canonical blob         → materialized entrypoint        (correct)
+//   100644 + any other blob         → link-path text or stale copy   (BROKEN)
+//
+// Comparing the blob to the canonical entrypoint asserts the invariant the
+// defect is actually about, and still catches #1051: a link-path blob never
+// equals the canonical blob. Reading the INDEX (not the filesystem) keeps this
+// true on Windows checkouts, where a symlink entry materializes as a text file.
+const CANONICAL_ENTRYPOINT = '.agents/skills/career-ops/SKILL.md';
+const stagedBlob = (path) => {
+  const entry = run('git', ['ls-files', '-s', path]);
+  if (entry === null || entry === '') return null;
+  const [mode, sha] = entry.split(/\s+/);
+  return { mode, sha };
+};
+
+const canonicalEntry = stagedBlob(CANONICAL_ENTRYPOINT);
+if (!canonicalEntry) {
+  fail(`Could not read git index entry for the canonical entrypoint ${CANONICAL_ENTRYPOINT}`);
+}
+
 const skillEntrypoints = systemFiles.filter((f) => f.endsWith('/skills/career-ops/SKILL.md'));
 for (const f of skillEntrypoints) {
-  const staged = run('git', ['ls-files', '-s', f]);
-  if (staged === null || staged === '') {
+  const staged = stagedBlob(f);
+  if (!staged) {
     fail(`Could not read git index entry for ${f} (lookup failed — not evidence of absence)`);
-  } else if (staged.startsWith('120000')) {
+  } else if (staged.mode === '120000') {
     pass(`Entrypoint is a real symlink in git: ${f}`);
+  } else if (canonicalEntry && staged.sha === canonicalEntry.sha) {
+    pass(`Entrypoint is a materialized regular file with canonical content: ${f}`);
   } else {
-    fail(`Entrypoint committed as a REGULAR file (mode ${staged.split(' ')[0]}) — users of this CLI get a broken skill: ${f}`);
+    fail(`Entrypoint committed as a REGULAR file (mode ${staged.mode}) whose content is not the canonical skill — users of this CLI get a broken skill: ${f}`);
   }
 }
 


santifer:
Excellent catch, @LayerC0de. My #2259 guard knew one correct state (symlink) when materialization created a second one, so every Windows user who ran apply got a permanently red suite. Comparing the staged blob against the canonical entrypoint asserts the invariant the #1051 defect is actually about, and a link-path blob still fails it. Better design than the guard it replaces. Merged! 🚀

PR #2276, opened 2026-07-29
Add Node engines field to package.json

Alejandrorojero07-del:
## What does this PR do?

Adds an `engines` field to `package.json`:

```json
"engines": {
  "node": ">=18"
}
```

## Verification

- Read `checkNodeVersion()` in `doctor.mjs` and confirmed that Node.js 18 is the minimum supported version.
- Checked `web/package.json` and confirmed it does not currently define an `engines` field.
- This change provides an npm warning for unsupported Node.js versions while leaving the existing compatibility policy unchanged.

## Summary by CodeRabbit

* **Chores**
  * Added a requirement for Node.js version 18 or later.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,9 @@
   "version": "1.23.0",
   "description": "AI-powered job search pipeline — works with any AI coding CLI (Claude Code, Codex, OpenCode, Antigravity, Grok, Qwen, Kimi, Copilot)",
   "private": true,
+  "engines": {
+    "node": ">=18"
+  },
   "scripts": {
     "doctor": "node doctor.mjs",
     "or": "node openrouter-runner.mjs",


santifer:
Welcome to career-ops, @Alejandrorojero07-del, and this is a first contribution done properly.

You did the two things the issue asked for and that most people skip: you read `checkNodeVersion()` in `doctor.mjs` to source the number rather than trusting the number written in the issue, and you checked `web/package.json` before assuming it needed the same treatment. **Verifying the premise instead of implementing the request** is the part that's hard to teach, and it's what made this mergeable on sight.

What it fixes is small and easy to underrate. Someone on an old Node used to get a confusing runtime error much later and reasonably conclude the tool was broken. `doctor.mjs` would have told them exactly what was wrong, but doctor runs *after* install, so the person whose install fails is precisely the person who never reaches the check that explains it. Now npm says it at the right moment.

Green across Linux, macOS and Windows. Glad to have you here.

PR #1651, opened 2026-07-06
feat(batch): native Node.js batch evaluator and runner fixes

d-ulker:
Replaces the fragile bash execution loop with a native Node.js parallel evaluator (batch-evaluate-gemini.mjs) using the Gemini SDK + Playwright directly.

Closes #1650

What does this PR do?
- **Native Batch Evaluator**: Introduces `batch-evaluate-gemini.mjs`, a robust, native Node.js parallel evaluator. As an explicitly-named fast path for Gemini, this leaves the agent-agnostic bash runner intact.
- **Concurrency & Rate Limiting**: Implements exponential backoff for API rate limits and a proper P-limit concurrency queue.
- **Inline Pipeline Updates**: Native handling of `.md` pipeline updates and status tracking.

Issue #1650, opened 2026-07-06
Bug: legacy bash-based batch-runner.sh is fragile and prone to crashes


d-ulker:
## Problem

The legacy bash-based `batch-runner.sh` script suffers from array boundary crashes, `awk` interpolation fragility, temp file security issues (predictable `/tmp` paths), and hardcoded CLI invocation issues.

## Solution

Introduce a new native Node.js evaluator (`batch-evaluate.mjs`) that implements safe concurrency limits, handles temp files securely via `mktemp`, normalizes tracker scores robustly, and delegates to a customizable `AGENT_CMD` environment variable.

This also integrates `reserve-report-num.mjs` natively to prevent race conditions during parallel processing.

See PR #1651

santifer:
Connected, @uelkerd: #1651 is the fix vehicle for this. Latest asks there: align the native evaluator with the spend_tier setting that landed via #1686, plus a rebase. Once those are in, this bug retires with it.

PR #2281, opened 2026-07-29
fix(batch): don't mark a worker-reported failure as completed

Lunarita1015:
## Problem

`batch-runner.sh`'s per-offer completion check only looks at the `claude -p` process's shell exit code. A worker can exit 0 (no crash) while still self-reporting `"status": "failed"` in its own JSON summary — for example, correctly declining to fabricate an evaluation when the JD couldn't be extracted (headless batch workers have no Playwright, so a JS-rendered board like Ashby can return only a title/company via WebFetch, with no actual JD content).

Without checking that field, such offers were silently recorded as `completed` in `batch-state.tsv` with `score: -` and no report file ever written. They'd never be retried (not marked `failed`), and the batch summary undercounted real failures. The only thing preventing outright data loss was `reconcile-pipeline.mjs`'s existing safety net, which leaves a `data/pipeline.md` entry in place when its assigned report file is missing from disk.

**Found via a live batch run**: offer id 6 (Deepgram — Applied ML Engineer, Ashby SPA) got `report_num 019` assigned, the worker correctly refused to evaluate with no JD content, but `batch-state.tsv` showed `completed`.

## Fix

After a zero exit code, check the worker log for a self-reported `"status": "failed"` JSON field before falling through to the `completed` branch. If present, route to the `failed` branch using the worker's own `"error"` field, same as any other failure path — including incrementing retries and printing a clear `❌ Failed (worker-reported, ...)` line instead of a false `✅ Completed`.

## Testing

- `bash -n batch/batch-runner.sh` — syntax OK
- Verified the new regexes against the actual log that triggered this bug (`"status": "failed"` and `"error": "..."` both extracted correctly)
- Ran `--dry-run` before/after to confirm normal offers are unaffected
- Re-ran the previously-mislabeled offer through the corrected flow end to end (it now correctly shows `failed` in state, is skipped once retries are exhausted instead of being silently swallowed, and stays recoverable in `data/pipeline.md`)

## Related

Related to #1650 / #1651 (native Node.js batch evaluator rewrite) — this is a narrow, low-risk fix for the current bash implementation while that rewrite is still in progress (open, needs-rebase as of this PR). Worth carrying the same worker-status check forward into the Node.js rewrite too, since the underlying failure mode (worker exits 0 but self-reports failure) isn't specific to the bash implementation.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved offer processing to detect worker-reported failures by parsing the worker’s last fenced `json` block in the per-offer log (including failure details and score).
  - Fail-closed behavior added when an expected report file is missing, marking the offer as failed early.
  - Prevented TSV state corruption by sanitizing multi-line error text before writing state.
  - Ensured reserved report numbers are released when an offer fails.
  - Merge processing now skips tracker entries whose report numbers were previously recorded as failed by the batch runner.

- **Chores**
  - Switched report-number reservation/release to a shared cross-process mechanism and tightened deterministic failure-output behavior in the worker prompt.


santifer:
Welcome to career-ops, @Lunarita1015! This is a remarkable first contribution: you didn't just fix the exit-0-but-failed case, you traced the whole fabrication path (fake 0.0/5 scores, the report-number collisions, the first-match sed) and closed it at three layers, each with the real incident documented. The refuse-to-merge guard on failed reports protects the tracker's integrity for everyone. Merged, your first contribution. Glad to have you here. 🚀

PR #2289, opened 2026-07-29
fix(dashboard): the 'hired' status renders with no colour in the pipeline list

### LINKED ISSUE #2251: dashboard (Go TUI): 'hired' is the one status with no colour — statusColorMap() omits it
## Summary

Sibling of #2249 on the Go side. `Hired` was added as a canonical status in #2050, and the Go dashboard recognizes it — `data.NormalizeStatus` folds `Hired`/`contratado`/`accepted`/… → `"hired"` (`dashboard/internal/data/career.go:508`), and `"hired"` is in the recognized set (`career.go:863`) and leads `statusGroupOrder` (`pipeline.go:212`). But the **status→colour map has no entry for it**:

```go
// dashboard/internal/ui/screens/pipeline.go
func (m PipelineModel) statusColorMap() map[string]lipgloss.Color {
	return map[string]lipgloss.Color{
		"interview": m.theme.Green,
		"offer":     m.theme.Green,
		"applied":   m.theme.Sky,
		"responded": m.theme.Blue,
		"evaluated": m.theme.Text,
		"skip":      m.theme.Red,
		"rejected":  m.theme.Subtext,
		"discarded": m.theme.Subtext,
		// no "hired"
	}
}
```

The pipeline row renderer looks the colour up by normalized status:

```go
// pipeline.go:1664
statusColor := m.statusColorMap()[norm]
statusStyle := lipgloss.NewStyle().Foreground(statusColor).Width(cw.status)
```

For a hired row, `norm == "hired"`, the map lookup misses, and `statusColor` is the zero-value `lipgloss.Color("")` — so the status text renders with **no explicit foreground** (default/inherited) while every other status is semantically coloured. The terminal-success state is the one status that doesn't stand out.

## Proposed fix

Add the entry, mirroring the positive convention (interview/offer are `m.theme.Green`):

```go
		"hired":     m.theme.Green,
```

A distinct celebratory shade would be defensible — hired is terminal, not intermediate — if the theme has one (e.g. a `Gold`/`Success` field); `Green` is the minimal correct mirror. Your call on which.

## Note

Happy to open the PR — flagging as an issue first only because `go` isn't available in my environment, so I can't build-test the change locally (CI would). If you'd rather I just send the one-liner, say so and I will.

abankar1:
Closes #2251 (the Go sibling of the web #2249/#2250 fix).

`Hired` was added as a canonical status in #2050, and the Go dashboard recognizes it — `data.NormalizeStatus` folds `Hired`/`contratado`/`accepted`/… → `"hired"`, it's in the recognized set, and it *leads* `statusGroupOrder`. But `statusColorMap()` had no entry for it:

```go
statusColor := m.statusColorMap()[norm]   // pipeline.go:1664, norm == "hired"
statusStyle := lipgloss.NewStyle().Foreground(statusColor)...
```

With the key missing, the lookup returns the zero-value `lipgloss.Color("")`, so the terminal-success status renders with **no explicit foreground** (default/inherited) — the one status that doesn't get a semantic colour.

## Fix

```go
"hired": m.theme.Green, // matches the interview/offer positive convention
```

I used `Green` as the minimal, uncontroversial mirror of the existing positive states. If you'd rather Hired **stand out** as the terminal win (it's the best outcome, not an intermediate one), the theme also has `Mauve`/`Peach` available — happy to switch it to whichever you prefer; that's a palette call I didn't want to make unilaterally.

> Note: `go` isn't available in my environment, so I couldn't `go build`/`go test` locally — the change is a single map entry using an existing `m.theme.Green` field (same as `interview`/`offer`), and CI builds + tests the dashboard. Flag me if the format/build check catches anything and I'll fix immediately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Hired statuses are now consistently displayed using the designated green color in metrics and application status views.

### DIFF
--- a/dashboard/internal/ui/screens/pipeline.go
+++ b/dashboard/internal/ui/screens/pipeline.go
@@ -2052,6 +2052,7 @@ func (m PipelineModel) scoreStyle(score float64) lipgloss.Style {
 
 func (m PipelineModel) statusColorMap() map[string]lipgloss.Color {
 	return map[string]lipgloss.Color{
+		"hired":     m.theme.Green, // terminal success — never uncoloured (default) like an unknown status
 		"interview": m.theme.Green,
 		"offer":     m.theme.Green,
 		"applied":   m.theme.Sky,


santifer:
The one-line ones matter too, @abankar1: the single most celebratory status in the whole pipeline rendering as uncoloured default was quietly sad. Green it is. Merged. 🚀

PR #2291, opened 2026-07-29
fix(web): AI-search drops offers when the <<offer: opener splits across a stream chunk

### LINKED ISSUE #2290: web: AI-search silently drops offers when the <<offer:> opener splits across a stream chunk
## Summary

`web/src/lib/explore-ai.ts`'s `makeAiStreamParser` is the tolerant streaming parser for the AI-search `<<offer:{…}>>` envelopes. Its own header states the load-bearing requirement:

> an envelope (or its opener) split across stream chunk boundaries must BUFFER, never flush as garbage or drop.

It violates that. When the `<<offer:` opener straddles a chunk boundary, the offer is **silently dropped** and the partial opener leaks into the narration stream as garbage — for every split position except one.

## Root cause

The no-opener branch tries to hold back a trailing partial opener, but checks whether the **fixed last `OPEN.length - 1` (7) characters** form a prefix of `OPEN`:

```ts
const keep = OPEN.length - 1;             // 7
const tail = buf.slice(buf.length - keep); // exactly the last 7 chars
if (OPEN.startsWith(tail)) { …hold… } else { …flush everything… }
```

`OPEN.startsWith(tail)` is true only when the last 7 chars are *exactly* `"<<offer"`. Any shorter partial opener at the end (`<<`, `<<off`, `<<offe`) preceded by other text makes `tail` a non-prefix, so the whole buffer — including the partial opener — is flushed. When the rest (`er:{…}>>`) arrives next chunk, there's no opener to match, and the offer is gone.

## Reproduce (pure logic, no server)

Porting the exact `feed()` loop and splitting one envelope at each opener offset:

```
split "<<offer:" after 2 chars → {"offers":0, …opener leaked to narration}
split "<<offer:" after 3 chars → {"offers":0, …}
split "<<offer:" after 5 chars → {"offers":0, …}
split "<<offer:" after 6 chars → {"offers":0, …}
split "<<offer:" after 7 chars → {"offers":1}   // the only one that works
no split (control)             → {"offers":1}
```

Token streaming chunks at arbitrary byte offsets, so in practice this drops a meaningful fraction of AI-discovered roles — and worse, they resurface as malformed narration.

## Proposed fix

Hold back the **longest suffix of `buf` that is a prefix of `OPEN`**, instead of a fixed-length tail check:

```ts
let hold = 0;
for (let k = Math.min(buf.length, OPEN.length - 1); k > 0; k--) {
  if (buf.endsWith(OPEN.slice(0, k))) { hold = k; break; }
}
```

With that, all split positions (and even a byte-by-byte stream) recover the offer with clean narration. I'll extract this pure helper into a `.mjs` (mirroring `clean-chips.mjs`) so it's unit-testable under `node --test`, and add regression cases for every split offset. PR follows.

abankar1:
Closes #2290.

`makeAiStreamParser` (`web/src/lib/explore-ai.ts`) is the tolerant parser for the AI-search `<<offer:{…}>>` envelopes. Its header states the invariant:

> an envelope (or its opener) split across stream chunk boundaries must BUFFER, never flush as garbage or drop.

It broke that. The no-opener branch held the buffer back only when the **fixed last `OPEN.length - 1` (7) characters** formed a prefix of the opener:

```ts
const tail = buf.slice(buf.length - keep);   // exactly 7 chars
if (OPEN.startsWith(tail)) { …hold… } else { …flush everything… }
```

`OPEN.startsWith(tail)` is true only when the tail is *exactly* `"<<offer"`. A shorter partial opener at the end (`<<`, `<<off`, `<<offe`) preceded by narration makes `tail` a non-prefix, so the whole buffer — partial opener included — is flushed. When `er:{…}>>` arrives next chunk there's no opener to match, and the offer is dropped (and leaks as malformed narration). Token streaming chunks at arbitrary offsets, so this fires often.

Ported the exact `feed()` loop to confirm:

```
split "<<offer:" after 2 chars → offers: 0   (opener leaked to narration)
split "<<offer:" after 5 chars → offers: 0
split "<<offer:" after 6 chars → offers: 0
split "<<offer:" after 7 chars → offers: 1   ← the only one that worked
```

## Fix

Hold back the **longest suffix of `buf` that is a prefix of `OPEN`** instead of a fixed-length tail. Extracted as a pure `pendingOpenerLen()` helper into `stream-parse.mjs` — mirroring the existing `clean-chips.mjs` pattern — so it's unit-testable under `node --test` without a TS runner, and `explore-ai.ts` stays fully typed (it just imports the helper).

After the fix, every split offset (and a byte-by-byte stream) recovers the offer with clean narration:

```
split after 1..7 → offers: 1   (all)
```

## Tests

`web/test-stream-parse.mjs` covers every split offset, the reported `<<off` case, the no-hold cases, a lone `<`, the length cap, and the mid-buffer false-positive guard. Wired into the web `test` script.

`npm test` (web) → **25 passed, 0 failed** (19 clean-chips + 6 new).

2 commits (fix + test).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved AI search streaming to correctly handle offer markers split across incoming data chunks.
  - Prevented partial markers, such as `<<off`, from appearing as unwanted narration or being discarded.
  - Preserved complete offer parsing and existing narration behavior while making streamed results more reliable.

- **Tests**
  - Added coverage for split-marker handling across chunk boundaries and related edge cases.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -8,7 +8,7 @@
     "build": "next build",
     "start": "next start",
     "typecheck": "tsc --noEmit",
-    "test": "node --test test-clean-chips.mjs"
+    "test": "node --test test-clean-chips.mjs test-stream-parse.mjs"
   },
   "dependencies": {
     "@paper-design/shaders-react": "^0.0.78",
--- a/web/src/lib/explore-ai.ts
+++ b/web/src/lib/explore-ai.ts
@@ -5,6 +5,7 @@
 // across stream chunk boundaries must BUFFER, never flush as garbage or drop.
 
 import type { DiscoveredOffer } from "./explore";
+import { pendingOpenerLen } from "./stream-parse.mjs";
 
 const OPEN = "<<offer:";
 const CLOSE = ">>";
@@ -58,20 +59,15 @@ export function makeAiStreamParser(opts?: { knownUrls?: Set<string> }) {
       for (;;) {
         const open = buf.indexOf(OPEN);
         if (open === -1) {
-          // No opener in view. Flush as narration — but hold back a short tail
-          // that could be the start of a split opener ("<<offe…").
-          const keep = OPEN.length - 1;
-          if (buf.length > keep) {
-            const tail = buf.slice(buf.length - keep);
-            if (OPEN.startsWith(tail)) {
-              const text = buf.slice(0, buf.length - keep);
-              if (text.trim()) out.push({ kind: "narration", text });
-              buf = tail;
-            } else {
-              if (buf.trim()) out.push({ kind: "narration", text: buf });
-              buf = "";
-            }
-          }
+          // No opener in view. Flush as narration — but hold back the longest
+          // trailing run that could be the start of a split opener ("<<offe…"),
+          // so an offer whose marker straddles a chunk boundary is never dropped
+          // as garbage (#2290). A fixed-length tail check missed the shorter
+          // partial openers ("<<", "<<off") preceded by other text.
+          const hold = pendingOpenerLen(buf, OPEN);
+          const text = hold ? buf.slice(0, buf.length - hold) : buf;
+          if (text.trim()) out.push({ kind: "narration", text });
+          buf = hold ? buf.slice(buf.length - hold) : "";
           break;
         }
         const before = buf.slice(0, open);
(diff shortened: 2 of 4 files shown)


abankar1:
Closing — this landed in `main` via the #2295 squash-merge (commit 8f5d10d, *"localize the hired status label **and buffer split stream openers**"*), which folded in this fix alongside the sibling Go/i18n changes. Verified the merged `stream-parse.mjs` / `act-envelope.mjs` helpers, the `explore-ai.ts` / `assistant-console.tsx` fixes, and both `node --test` files are byte-identical to this branch. Nothing outstanding — thanks!

PR #2293, opened 2026-07-29
fix(web): stop the assistant console flickering a half-written `<<act:` opener into the bubble

### LINKED ISSUE #2292: web: assistant console flickers a half-written `<<act:` opener into the chat bubble on every action
### What

In `web/src/components/assistant-console.tsx`, `parseEnvelopes(acc)` hides an in-flight action envelope via `hidePartialFrom` — but only once the opener regex `/<<act:([a-zA-Z]+)[ \t]+/` has *fully* matched, i.e. `<<act:`, the id letters, **and** the trailing space have all streamed in.

Token streaming delivers those a few characters at a time, so on every chunk between `<<` arriving and the trailing space arriving, nothing hides the partial opener and it renders into the assistant bubble. The user sees `<<act:navigate` (and `<<`, `<<act:`, …) flash and then vanish on **every** action the assistant takes — navigate, remember, setStatus, setApplyField, etc.

### Repro (port of the component's parse + display path)

Streaming `Here you go. <<act:navigate {"path":"/pipeline"}>> Done.` char-by-char, the derived display leaks 11 intermediate frames:

```
"Here you go. <<ac"
"Here you go. <<act"
"Here you go. <<act:"
"Here you go. <<act:n"
… through …
"Here you go. <<act:navigate"
```

Once the space after the id arrives, `hidePartialFrom` finally kicks in and it disappears — hence the flicker.

### Expected

A partial `<<act:` opener at the tail of the buffer should be hidden too, so nothing renders until the envelope either completes or turns out to be plain text. This is the same "buffer a split opener, never render garbage" invariant the AI-search parser holds (`stream-parse.mjs`, #2290).

### Fix sketch

Detect a trailing partial opener (`<`, `<<`, … `<<act:save` with no space yet) and fold its start into `hidePartialFrom`. Extract the detection into a pure `.mjs` helper so it's unit-testable under `node --test` (matching `clean-chips.mjs` / `stream-parse.mjs`).

PR incoming.

abankar1:
Closes #2292.

`parseEnvelopes` in `web/src/components/assistant-console.tsx` hides an in-flight `<<act:ID {json}>>` envelope via `hidePartialFrom` — but only after the opener regex `/<<act:([a-zA-Z]+)[ \t]+/` has *fully* matched, which needs `<<act:`, the id letters, **and** the trailing space. Streaming delivers those a few chars at a time, so on every chunk between `<<` and the space, nothing hides the partial opener and it renders into the chat bubble.

Ported the component's parse + display path and streamed a navigate action char-by-char — 11 frames leak before the space arrives:

```
"Here you go. <<ac"
"Here you go. <<act:"
"Here you go. <<act:navigate"
…
```

Since every assistant action (navigate, remember, setStatus, setApplyField) opens with `<<act:`, the user sees this flicker constantly.

## Fix

Fold a **trailing** partial opener's start index into `hidePartialFrom`. Extracted the detection into a pure `pendingActOpenerStart()` helper in `web/src/lib/act-envelope.mjs` — mirroring `stream-parse.mjs` / `clean-chips.mjs` — so it's unit-testable under `node --test` and the component change stays a two-line fold. It's anchored to end-of-buffer, so a completed envelope earlier in the string (followed by `>>`) is never matched, and once the trailing space arrives it defers to the existing regex.

Re-running the same simulation with the helper applied: **0 leak frames**, final display `"Here you go.  Done."`.

## Tests

`web/test-act-envelope.mjs` covers every trailing partial as the opener streams in, the bare-prefix cases, the completed-envelope / post-space cases (deferred to the main regex), and the mid-buffer false-positive guard.

`npm test` (web) → **31 passed, 0 failed** (19 clean-chips + 6 stream-parse + 6 act-envelope).

Sibling to #2291 (same invariant, the AI-search `<<offer:` parser). 2 commits (fix + test).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming message parsing to prevent partial action and offer markers from appearing prematurely or flickering.
  * Preserved narration correctly when streamed content splits marker text across chunks.

* **Tests**
  * Expanded automated coverage for split-stream markers and incomplete action envelopes.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -8,7 +8,7 @@
     "build": "next build",
     "start": "next start",
     "typecheck": "tsc --noEmit",
-    "test": "node --test test-clean-chips.mjs"
+    "test": "node --test test-clean-chips.mjs test-stream-parse.mjs"
   },
   "dependencies": {
     "@paper-design/shaders-react": "^0.0.78",
--- a/web/package.json
+++ b/web/package.json
@@ -8,7 +8,7 @@
     "build": "next build",
     "start": "next start",
     "typecheck": "tsc --noEmit",
-    "test": "node --test test-clean-chips.mjs test-stream-parse.mjs"
+    "test": "node --test test-clean-chips.mjs test-stream-parse.mjs test-act-envelope.mjs"
   },
   "dependencies": {
     "@paper-design/shaders-react": "^0.0.78",
--- a/web/src/components/assistant-console.tsx
+++ b/web/src/components/assistant-console.tsx
@@ -15,6 +15,7 @@ import { WorkerCard } from "@/components/jobs/worker-card";
 import { Button } from "@/components/ui/button";
 import { dispatch, type ActionCtx, type DoneInfo } from "@/app/actions/registry";
 import { scoreNum } from "@/lib/format";
+import { pendingActOpenerStart } from "@/lib/act-envelope.mjs";
 import { cn } from "@/lib/cn";
 
 // ── message model: messages are PART arrays so a live worker card can render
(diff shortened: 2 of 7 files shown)


abankar1:
Closing — this landed in `main` via the #2295 squash-merge (commit 8f5d10d, *"localize the hired status label **and buffer split stream openers**"*), which folded in this fix alongside the sibling Go/i18n changes. Verified the merged `stream-parse.mjs` / `act-envelope.mjs` helpers, the `explore-ai.ts` / `assistant-console.tsx` fixes, and both `node --test` files are byte-identical to this branch. Nothing outstanding — thanks!

PR #2295, opened 2026-07-29
fix(dashboard): localize the `hired` status label (StatusLabel dropped it)

### LINKED ISSUE #2294: dashboard(i18n): `StatusLabel` doesn't localize `hired` — renders the raw lowercase ID
### What

`dashboard/internal/i18n/catalog.go`'s `StatusLabel(norm)` maps a canonical status ID to its localized display label, but its `switch` has no `case "hired"`. The `Hired` state was added to the tracker (canonical `states.yml`, added in #2050) and the i18n catalog *does* carry a `StatusHired` field with translations in every locale:

- En: `"Hired"`
- Tr: `"İşe Alındı"`
- Es: `"Contratada"`

…but `StatusLabel` never returns it. A `hired` row therefore falls through to `default: return norm` and renders the raw lowercase English `"hired"` in the TUI — untranslated and uncapitalized — instead of the localized label that already exists two structs away.

Same class of incomplete `Hired` propagation as the pipeline colour-map gap in #2289: the data was added, the switch that consumes it was missed.

### Expected

`StatusLabel("hired")` returns `c.StatusHired` (localized), like every other canonical status.

### Notes

`TestStatusLabel` in `i18n_test.go` covers the other 8 statuses but omits `hired`, which is why the gap went unnoticed. Fix + regression row incoming.

abankar1:
Closes #2294.

`StatusLabel(norm)` in `dashboard/internal/i18n/catalog.go` maps a canonical status ID to its localized label, but its `switch` had no `case "hired"`. The catalog already carries a `StatusHired` field translated in every locale:

| Locale | `StatusHired` |
|--------|---------------|
| En | `Hired` |
| Tr | `İşe Alındı` |
| Es | `Contratada` |

Without the case, a `hired` row fell through to `default: return norm` and rendered the raw lowercase English `"hired"` in the TUI — untranslated, uncapitalized — while the correct label sat unused two structs away.

Same incomplete-`Hired`-propagation class as #2289 (the pipeline colour map): `Hired` was added to `states.yml` (#2050) and translated, but a consuming `switch` was missed.

## Fix
- Add `case "hired": return c.StatusHired` (placed after `offer`, matching the canonical order).
- Update the doc comment's status list to include `hired`.

## Test
`TestStatusLabel` covered the other 8 statuses but omitted `hired` — which is why this slipped through. Added the row:

```go
{"hired", "Hired", "İşe Alındı", "Contratada"},
```

2 commits (fix + test).

## Summary by CodeRabbit

* **New Features**
  * Added localized “Hired” status labels in Turkish, Spanish, and English.

* **Bug Fixes**
  * Improved streaming message rendering by preventing incomplete action and offer markers from briefly appearing.
  * Preserved text correctly when markers are split across incoming stream chunks.

* **Tests**
  * Expanded coverage for localized status labels and streaming marker handling.

### DIFF
--- a/dashboard/internal/i18n/catalog.go
+++ b/dashboard/internal/i18n/catalog.go
@@ -171,13 +171,16 @@ func (c *Catalog) ViewModeLabel(mode string) string {
 }
 
 // StatusLabel returns the localized display label for a canonical status ID
-// (interview, offer, responded, applied, evaluated, skip, rejected, discarded).
+// (interview, offer, hired, responded, applied, evaluated, skip, rejected,
+// discarded).
 func (c *Catalog) StatusLabel(norm string) string {
 	switch strings.ToLower(strings.TrimSpace(norm)) {
 	case "interview":
 		return c.StatusInterview
 	case "offer":
 		return c.StatusOffer
+	case "hired":
+		return c.StatusHired
 	case "responded":
 		return c.StatusResponded
 	case "applied":
--- a/dashboard/internal/i18n/i18n_test.go
+++ b/dashboard/internal/i18n/i18n_test.go
@@ -14,6 +14,7 @@ func TestStatusLabel(t *testing.T) {
 	}{
 		{"interview", "Interview", "Mülakat", "Entrevista"},
 		{"offer", "Offer", "Teklif", "Oferta"},
+		{"hired", "Hired", "İşe Alındı", "Contratada"},
 		{"responded", "Responded", "Yanıt Verildi", "Respondida"},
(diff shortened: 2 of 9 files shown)


santifer:
Nice continuation of the hired-status cluster, @abankar1. The Go catalog now speaks 'hired' in all three languages with tests, and extracting the split-opener buffering into single-source-of-truth helpers that both the component and the tests import — so they can never drift — is exactly the right shape for the #2290 class. Merged! 🚀

Issue #2188, opened 2026-07-26
appendToPipeline() has an unlocked read-modify-write race on data/pipeline.md


rubicon:
### Code of Conduct

- [x] I agree to follow the [Code of Conduct](https://github.com/santifer/career-ops/blob/main/CODE_OF_CONDUCT.md)

### Existing issues

- [x] I searched existing issues and this hasn't been requested yet

### Problem

`appendToPipeline()` (`scan.mjs`) does a plain read-modify-write on `data/pipeline.md`: `readFileSync` -> mutate the string -> `writeFileSync`, with no locking. It's shared by three current callers -- `scan.mjs` itself, `scan-ats-full.mjs`, and `plugins.mjs` (pipeline mode) -- so any two of them running concurrently (e.g. a scheduled scan overlapping a manual `/career-ops pipeline` run, or two plugin jobs) can silently drop one side's offers: whichever write lands second overwrites the first's in-memory read, with no error and no trace that anything was lost.

I haven't observed this in the wild -- it's a code-level finding, not a reported failure. But `data/pipeline.md` is the single inbox every intake path feeds, so a silent drop here is a silent lead loss, which is hard to notice and worse than a loud failure.

### Proposed solution

Add a small, self-contained cross-process advisory lock (`pipeline-lock.mjs`) scoped to this one file: mkdir-is-atomic lock directory, stale-holder reclaim, retry with backoff, throws on timeout. `appendToPipeline()` acquires it for its read-modify-write; since all three callers go through this one shared, exported function, wrapping it there protects all of them (and any future caller) with a single change, no call-site-by-call-site wrapping needed. Scoped deliberately small -- no re-entrancy tracking or generic multi-file lock manager, since nothing here nests a second acquisition of the same lock.

Opened a PR implementing this -- linked below.

### Area

Portal Scanner

PR #2305, opened 2026-07-29
fix(pipeline-lock): floor the ownerless stale check, harden the timeout tests

### LINKED ISSUE #2304: bug(pipeline-lock): ownerless locks and the recover guard have no stale-check floor — a just-created lock can be reclaimed instantly
### Code of Conduct

### Existing issues

- [X] I searched existing issues and this hasn't been reported yet

### What happened?

`lockCanRecover()` in `pipeline-lock.mjs` falls back to a pure `age > staleMs` check whenever the lock directory has no readable `owner.json`. Two ownerless windows exist by construction, so a *just-created* directory can be judged stale and deleted by a concurrent caller:

```js
function lockCanRecover(lockDir, staleMs) {
  const owner = readLockOwner(lockDir);
  if (owner?.pid) return !processIsAlive(owner.pid);
  try {
    return Date.now() - statSync(lockDir).mtimeMs > staleMs;   // <-- no floor
  } catch {
    return true; // vanished — nothing to recover, retry acquisition
  }
}
```

**1. The acquisition window.** `acquirePipelineLock()` does `mkdirSync(lockDir)` and only then `writeFileSync(owner.json)`. Between those two calls the lock is ownerless. A second acquirer that reaches `lockCanRecover()` inside that window takes the age branch, and with a small caller-supplied `staleMs` it immediately `rmSync`s the winner's brand-new lock and acquires it itself — while the first process goes on to write its `owner.json` into the replacement directory. Both callers then believe they hold the lock, which is exactly the lost-update race #2188 closed.

**2. The recover guard is *always* ownerless.** `recoverGuardDir` never gets an `owner.json` written into it, so `lockCanRecover(recoverGuardDir, staleMs)` is *only ever* the age branch. With a small `staleMs`, a caller that finds the guard held simply deletes it and takes it on the next pass — so two callers end up inside the decide-then-delete window simultaneously, and the first one's `finally { rmSync(recoverGuardDir) }` then removes the second one's guard. The serialization the guard exists to provide is defeated, re-opening the TOCTOU race documented in the module header.

This also makes the existing regression test `acquirePipelineLock: stale-reclaim is serialized …` (which runs with `staleMs: 1`) racy rather than deterministic: it only passes because the `mkdir` → `writeFileSync` window is usually under 1 ms. It is not asserting what it claims to assert on a loaded machine.

Separately, and in the same file's test suite, two timeout assertions are too weak to detect the above:

- `test/pipeline-lock.test.mjs` L58-61 — `await assert.rejects(() => acquirePipelineLock(...))` with no error predicate and no lower bound on elapsed time. It passes if the losing acquirer blows up instantly for a completely unrelated reason (e.g. an `ENOENT` from a refactor), so a broken lock would still look green.
- `test/pipeline-lock.test.mjs` L105-107 — the stale-reclaim test counts the winners but never checks *why* the loser lost. A loser that failed for a non-lock reason satisfies the assertion.

Both were raised on the merged PR #2189 and were not addressed before merge.

### Steps to reproduce

Save as `repro.mjs` in the repo root and run `node repro.mjs` (no config or CV needed):

```js
import { mkdtempSync, rmSync, mkdirSync, existsSync, writeFileSync } from 'node:fs';
import { tmpdir } from 'node:os';
import { join } from 'node:path';
import { acquirePipelineLock } from './pipeline-lock.mjs';

const root = mkdtempSync(join(tmpdir(), 'repro-'));
mkdirSync(join(root, 'data'), { recursive: true });
const p = join(root, 'data', 'pipeline.md');
const lockDir = `${p}.lock`;
const guardDir = `${lockDir}.recover`;

// A: a holder has mkdir'd the lock but has not written owner.json yet.
mkdirSync(lockDir);
await new Promise((r) => setTimeout(r, 5));
try {
  const lock = await acquirePipelineLock(p, { timeoutMs: 200, retryMs: 20, staleMs: 1 });
  console.log('REPRO A: BUG — deleted the fresh ownerless lock and acquired it.');
  lock.release();
} catch (err) {
  console.log(`REPRO A: safe — acquirer refused (${err.constructor.name})`);
}
rmSync(lockDir, { recursive: true, force: true });

// B: another caller is inside the guarded decide-then-delete window right now.
mkdirSync(lockDir);
writeFileSync(join(lockDir, 'owner.json'), JSON.stringify({
  pid: 2147483000, token: 'crashed', started_at: new Date(Date.now() - 3600_000).toISOString(),
}));
mkdirSync(guardDir);
await new Promise((r) => setTimeout(r, 5));
try {
  const lock = await acquirePipelineLock(p, { timeoutMs: 300, retryMs: 20, staleMs: 1 });
  console.log(`REPRO B: BUG — deleted the live recover guard (guard still exists: ${existsSync(guardDir)}) and reclaimed the lock.`);
  lock.release();
} catch (err) {
  console.log(`REPRO B: safe — acquirer refused (${err.constructor.name}); guard intact: ${existsSync(guardDir)}`);
}
rmSync(root, { recursive: true, force: true });
```

Observed on `main` @ `dab8517`:

```
REPRO A: BUG PRESENT — concurrent acquirer DELETED the fresh ownerless lock and acquired it.
REPRO B: BUG PRESENT — acquirer deleted the live recover guard (guard still exists: false) and reclaimed the lock while another caller was inside the guarded window.
```

### Expected behavior

An ownerless lock or recover-guard directory should get a small fixed grace period before age-based reclamation becomes possible, independent of the caller-supplied `staleMs` — so a directory that was created microseconds ago is never judged stale, no matter how aggressive the caller's timing. Both repro cases should print `safe`.

The grace period must be a **floor**, not a replacement: a caller passing `staleMs: 60_000` must still keep an ownerless lock protected for the full 60 s, and a genuinely abandoned ownerless directory (older than the floor) must still be reclaimable, so a crash while holding the recover guard cannot disable stale recovery permanently.

The two timeout tests should assert `LockTimeoutError` specifically, and the elapsed-time test should carry a lower bound so it cannot pass on an instantaneous unrelated failure.

### CLI tool

Claude Code

### OS and Node.js version

macOS 26.5 / Node v22.22.3

### Output of npm run doctor

```shell
✓ Node.js >= 22.5 (v22.22.3)
✓ Dependencies installed
  → No project-level MCP config was detected for claude.
  → Add the Playwright MCP server to your claude config.
⚠ cv.md not found (user setup required)
  → Create cv.md in the project root with your CV in markdown
  → See examples/ for reference CVs
⚠ config/profile.yml not found (user setup required)
  → Run: cp config/profile.example.yml config/profile.yml
  → Then edit it with your details
⚠ modes/_profile.md not found (user setup required)
  → Run: cp modes/_profile.template.md modes/_profile.md
  → Then customize your archetypes / targeting narrative
⚠ portals.yml not found (user setup required)
  → Run: cp templates/portals.example.yml portals.yml
  → Then customize with your target companies
✓ Fonts directory ready
✓ data/ directory ready
✓ data/pipeline.md ready (auto-created)
✓ output/ directory ready
✓ reports/ directory ready

Result: All checks passed (5 warnings — see above). You're ready to go!
```

(Clean checkout of `main`, no user data configured — the warnings are just the un-onboarded state, unrelated to this bug.)

rubicon:
## What does this PR do?

Gives `lockCanRecover()` in `pipeline-lock.mjs` a fixed grace floor before an **ownerless** directory can be condemned on age alone, so a lock or recover guard created microseconds ago is never judged stale no matter how aggressive the caller's `staleMs`. Also hardens the two timeout assertions in `test/pipeline-lock.test.mjs` to require `LockTimeoutError` specifically instead of any rejection.

## Related issue

Closes #2304

Both points were raised by review on #2189 and were still open when it merged.

## Type of change

- [X] Bug fix
- [ ] New feature

## Details

`lockCanRecover()` fell back to a bare `age > staleMs` check whenever a directory had no readable `owner.json`. Two directories are ownerless **by construction**, not by accident:

1. **A lock inside its acquisition window** — `acquirePipelineLock()` does `mkdirSync(lockDir)` and only then `writeFileSync(owner.json)`. A contender that lands in between takes the age branch, deletes the winner's brand-new lock, and acquires it; the winner then writes its `owner.json` into the replacement directory and both callers believe they hold the lock — precisely the lost-update race #2188 closed.
2. **The recover guard, always** — `recoverGuardDir` never gets an `owner.json` written into it, so `lockCanRecover(recoverGuardDir, staleMs)` is *only ever* the age branch. A caller that finds the guard held just deletes it and takes it on the next pass, so two callers end up inside the decide-then-delete window simultaneously and the first one's `finally { rmSync(recoverGuardDir) }` removes the second one's guard. The serialization the guard exists to provide is defeated.

The fix floors the age branch with `Math.max(staleMs, OWNERLESS_GRACE_MS)` (1 s). It is a **lower bound on patience, never a cap**:

- a larger caller `staleMs` still wins outright (`staleMs: 60_000` still means 60 s),
- the 30 s default path is bit-for-bit unchanged (`Math.max(30_000, 1_000)`),
- a genuinely abandoned ownerless directory still ages out, so a process killed while holding the guard cannot disable stale recovery permanently.

This also removes a latent flake: the existing `stale-reclaim is serialized …` test runs with `staleMs: 1` and only passed because the `mkdir` → `writeFileSync` window is usually under 1 ms. On a loaded machine it could observe two winners.

### Related, deliberately not touched

`tracker-utils.mjs` and `followup-seed.mjs` carry the same unfloored `lockCanRecover()` shape. Fixing those means re-tuning existing callers that pass small `staleMs` values (e.g. `set-status-tests.mjs` uses `staleMs: 50`), so it belongs in its own issue rather than being smuggled into this one. Happy to open it if you want the family fixed.

## Test plan

- [X] **RED first.** Added the new tests against unmodified `main` and confirmed 2 of them fail, for the right reason — `Missing expected rejection`, i.e. the contender *successfully acquired* a lock it must not have been able to acquire:
  ```
  not ok 5 - acquirePipelineLock: a just-created ownerless lock is never judged stale …
    error: 'Missing expected rejection.'
  not ok 8 - acquirePipelineLock: a live recover guard is not deleted out from under the caller inside it …
    error: 'Missing expected rejection.'
  # tests 10 / pass 8 / fail 2
  ```
- [X] **GREEN after the fix** — `node --test test/pipeline-lock.test.mjs` → `# tests 10 / # pass 10 / # fail 0` in 2.2 s.
- [X] **Adversarial cases added** beyond the two findings:
  - *multiplicity* — a contender racing a lock inside its ownerless acquisition window (`staleMs: 0`);
  - *interaction* — a contender against a **live** recover guard; asserts the guard directory still exists afterwards, not just that acquisition failed;
  - *boundary, floor is a floor* — an ownerless lock aged past the floor but well within a large caller `staleMs` must stay protected (catches a fix that replaces `staleMs` instead of flooring it);
  - *boundary, liveness* — an ownerless lock older than the floor must still be reclaimable (catches an over-correction that never reclaims ownerless locks);
  - *boundary, no permanent deadlock* — a guard abandoned by a crashed process must still age out and be cleaned up.
- [X] Tests stay fast — no sleeps added. Ages are simulated with `utimesSync`, and contention uses the existing caller-configurable `timeoutMs`/`retryMs`/`staleMs`. Suite runtime 2.2 s.
- [X] Verified independently of the test file with a standalone repro script (in the issue): both cases print `BUG PRESENT` on `main` @ `dab8517` and `safe` after the fix.
- [X] `node test-all.mjs` → **2451 passed, 0 failed** (1 warning: the pre-existing `cv-sync-check.mjs exited with error (expected without user data)` on a checkout with no CV).
- [X] `node validate-system-paths-coverage.mjs` → `OK: 886 tracked files covered`.
- [X] Grepped every `acquirePipelineLock` caller: none passes a custom `staleMs`, so no production path changes behavior.

## Checklist

- [X] I have read [CONTRIBUTING.md](https://github.com/santifer/career-ops/blob/main/CONTRIBUTING.md)
- [X] If this is a new feature or architecture change, I opened an issue first (bug fixes, providers, docs & translations are exempt — send those straight in)
- [X] My PR does not include personal data (CV, email, real names, scan results, or pipeline data)
- [X] I ran `node test-all.mjs` and all tests pass
- [X] My changes respect the [Data Contract](https://github.com/santifer/career-ops/blob/main/DATA_CONTRACT.md) (no modifications to user-layer files)
- [X] My changes align with the [project roadmap](https://github.com/santifer/career-ops/discussions/156)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved stale lock recovery to prevent immediate reclamation of locks without owner information.
  - Added a minimum grace period before ownerless locks can be recovered.
  - Preserved longer, caller-defined stale timeouts.
  - Protected active recovery operations from being removed by competing processes.
  - Ensured abandoned recovery operations are eventually cleaned up.

- **Tests**
  - Expanded coverage for lock contention, timeout behavior, grace periods, and recovery safeguards.

### DIFF
--- a/pipeline-lock.mjs
+++ b/pipeline-lock.mjs
@@ -16,7 +16,10 @@
 //     before deleting anything.
 //   - staleness is judged by owner-PID liveness first, falling back to
 //     directory age only when the metadata is missing or unreadable. An old
-//     lock whose owner is still running is NOT stale.
+//     lock whose owner is still running is NOT stale, and an ownerless
+//     directory gets a fixed grace period (OWNERLESS_GRACE_MS) before age
+//     alone can condemn it, so a directory created microseconds ago is never
+//     reclaimable no matter how aggressive the caller's staleMs.
 //   - stale reclamation is serialized behind a second atomic guard directory
 //     ("<path>.lock.recover"). Without it, reclamation is itself a TOCTOU
 //     race: two callers that both judge the same lock stale can have the
@@ -32,6 +35,7 @@ import { join, dirname } from 'path';
 import { randomUUID } from 'crypto';
 
 const DEFAULT_STALE_MS = 30_000;
+export const OWNERLESS_GRACE_MS = 1_000;
 const DEFAULT_RETRY_MS = 80;
 const DEFAULT_TIMEOUT_MS = 8_000;
 
@@ -77,11 +81,22 @@ function processIsAlive(pid) {
 
 // Conservative: a lock whose recorded owner is still running is never stale,
 // however old it is. Age is the fallback only when there's no readable owner.
+//
+// That fallback needs a floor. Two directories are ownerless by construction,
+// not by accident: a lock between its mkdir and its owner.json write, and the
+// recover guard, which never carries owner.json at all. Judging those on
+// `age > staleMs` alone lets a caller with an aggressive staleMs delete a
+// directory created microseconds ago — either stealing a winner's lock inside
+// its acquisition window, or evicting a live guard and putting two callers
+// inside the decide-then-delete window the guard exists to serialize.
+// OWNERLESS_GRACE_MS is a lower bound on that patience, never a cap: a larger
+// caller staleMs still wins, and a genuinely abandoned directory still ages
+// out, so a crash while holding the guard cannot disable recovery for good.
 function lockCanRecover(lockDir, staleMs) {
   const owner = readLockOwner(lockDir);
   if (owner?.pid) return !processIsAlive(owner.pid);
   try {
-    return Date.now() - statSync(lockDir).mtimeMs > staleMs;
+    return Date.now() - statSync(lockDir).mtimeMs > Math.max(staleMs, OWNERLESS_GRACE_MS);
   } catch {
     return true; // vanished — nothing to recover, retry acquisition
   }
@@ -95,7 +110,7 @@ function lockCanRecover(lockDir, staleMs) {
  * @param {object} [options]
  * @param {number} [options.timeoutMs=8000] - Max time to wait for the lock.
  * @param {number} [options.retryMs=80] - Delay between acquisition attempts.
- * @param {number} [options.staleMs=30000] - Age threshold for a lock with no readable owner.
+ * @param {number} [options.staleMs=30000] - Age threshold for a lock with no readable owner, floored at OWNERLESS_GRACE_MS.
  */
 export async function acquirePipelineLock(pipelinePath, options = {}) {
   // Env overrides let a caller several frames up the stack (a test driving
--- a/test/pipeline-lock.test.mjs
+++ b/test/pipeline-lock.test.mjs
@@ -19,15 +19,33 @@
 
 import { test } from 'node:test';
 import assert from 'node:assert/strict';
-import { mkdtempSync, rmSync, existsSync, mkdirSync, writeFileSync, readFileSync } from 'node:fs';
+import { mkdtempSync, rmSync, existsSync, mkdirSync, writeFileSync, readFileSync, utimesSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { join, dirname } from 'node:path';
-import { acquirePipelineLock, LockTimeoutError } from '../pipeline-lock.mjs';
+import { acquirePipelineLock, LockTimeoutError, OWNERLESS_GRACE_MS } from '../pipeline-lock.mjs';
 
 function fixtureRoot() {
   const root = mkdtempSync(join(tmpdir(), 'career-ops-pipeline-lock-'));
   mkdirSync(join(root, 'data'), { recursive: true });
   return root;
+}
+
+// Ages a directory by rewriting its mtime, so the age-based branch can be
+// exercised at any point on either side of the grace floor without sleeping.
+function backdate(dir, ms) {
+  const when = new Date(Date.now() - ms);
+  utimesSync(dir, when, when);
+}
+
+// A lock directory left behind by a holder that died — stale by the owner-PID
+// rule, so reclamation is genuinely on the table.
+function writeCrashedHolder(lockDir) {
+  mkdirSync(lockDir, { recursive: true });
+  writeFileSync(join(lockDir, 'owner.json'), JSON.stringify({
+    pid: 2147483000, // not a live pid
+    token: 'crashed-holder-token',
+    started_at: new Date(Date.now() - 60 * 60_000).toISOString(),
+  }), 'utf-8');
 }
 
 test('acquirePipelineLock: a live holder blocks a second acquirer, which times out', async () => {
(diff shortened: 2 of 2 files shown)


santifer:
The grace-floor-as-lower-bound reasoning in that comment block is exactly right, @rubicon: a floor that can never become a cap means a crashed guard-holder still ages out. And swapping sleeps for utimesSync backdating makes the tests both faster and more honest. Merged. 🚀

PR #2307, opened 2026-07-29
fix(echojobs): keep hybrid roles distinguishable from remote (#2258)

### LINKED ISSUE #2258: echojobs provider labels hybrid roles as 'Remote' (location_filter can't catch them)
### Summary

`providers/echojobs.mjs` maps `remote_type: 'hybrid'` to the location string `'Remote'`, identical to a genuinely remote role. This defeats a remote-only `location_filter`: hybrid postings pass the filter and land in the pipeline as if fully remote.

### Where

`providers/echojobs.mjs` (around line 93):

```js
if (!location && (j.remote_type === 'remote' || j.remote_type === 'hybrid')) location = 'Remote';
```

When the feed item has no explicit `locations[]`, a **hybrid** role is given `location = 'Remote'` — the same value as a truly remote one.

### Why it matters

`location_filter` in `portals.yml` already blocks the token `"Hybrid"`, but this line strips that signal before the filter runs: the emitted location is literally `Remote`, so the block can never match. A user who has configured a remote-only search still receives hybrid (partially on-site) roles, indistinguishable from remote ones downstream. The mislabel also propagates to `scan-history.tsv` and any liveness/eval step that trusts the location string.

### Expected

Hybrid should stay distinguishable from remote so the existing `location_filter` `"Hybrid"` block (or a user's own rule) can act on it. Two reasonable options:

1. Emit `'Hybrid'` for `remote_type === 'hybrid'` (mirrors how other providers surface a hybrid signal, and the existing block token catches it), or
2. Only set `'Remote'` for `remote_type === 'remote'`; leave hybrid's location empty so it falls to the missing-data valve rather than a false "Remote".

Option 1 is preferable since it preserves information and is caught by the shipped block list.

### Repro sketch

An EchoJobs feed item with `remote_type: 'hybrid'` and no `locations[]` → parsed job carries `location: 'Remote'` → passes a remote-only `location_filter`.

Observed on v${VER}.

Zoubeir23:
Fixes #2258

## Problem

In `normalizeEchojobsJob`, a placeless posting with `remote_type: 'hybrid'` was mapped to the same location string as a genuinely remote one:

```js
if (!location && (j.remote_type === 'remote' || j.remote_type === 'hybrid')) location = 'Remote';
```

The emitted location string is what `location_filter` matches on, so a remote-only `block: ["Hybrid"]` rule could never match and hybrid roles silently passed it.

## Fix

Hybrid is never collapsed into `Remote`. The work model is surfaced so the existing block-token mechanism can act on it — no new filter machinery:

```js
if (remoteType === 'hybrid') {
  if (!HYBRID_MARKER.test(location)) location = [location, 'Hybrid'].filter(Boolean).join(' · ');
} else if (!location && remoteType === 'remote') {
  location = 'Remote';
}
```

- **Placeless hybrid** → `"Hybrid"`.
- **Hybrid that lists a city** → `"Berlin · Hybrid"` — the city is kept, the marker appended, same shape as `oraclecloud.mjs`'s `WorkplaceTypeCode` hint. Marking only the placeless ones would leave the fix half-working: `block: ["Hybrid"]` would catch them but still pass every hybrid that happens to list a city. `allow: ["berlin"]` keeps matching (substring), `block: ["Hybrid"]` now rejects.
- **Marker never doubled** when the board already spells it out (`"Berlin (Hybrid)"`).
- **`remote_type` compared case/whitespace-insensitively** — third-party aggregate feed; a `"Hybrid"` variant must not fall through and produce an unmarked, unfilterable role.
- **Placed remote roles are untouched** (`"Berlin"` stays `"Berlin"`) — out of scope for this issue.
- **Placeless `on_site`** keeps `""` and passes, per the scanner's "don't penalize missing data" convention.

This diverges from `greenhouse.mjs`, which treats a work-model-only location as damage to repair via `/offices` enrichment. That provider can recover a real city; this feed exposes none, so the work model is the only signal there is and is better surfaced than dropped. Documented in the JSDoc.

### Behaviour change worth calling out

Users with an `allow`-list such as `allow: ["remote"]` previously received placeless hybrid roles (they were labelled `"Remote"`). They will now be filtered out. That is the point of the issue, but it is a visible drop in volume for those configs.

## Test plan

- [x] `node tests/providers/echojobs.test.mjs` — placeless remote/hybrid split, placeless `on_site` (no false `Remote`/`Hybrid`), placed hybrid marker, no-double-marker guard, placed remote untouched, case/whitespace-insensitive `remote_type`
- [x] Verified end-to-end against `buildLocationFilter`: `block: ["Hybrid"]` rejects both `"Hybrid"` and `"Berlin · Hybrid"` while passing `"Berlin"` and `"Remote"`; `allow: ["berlin"]` still accepts `"Berlin · Hybrid"`
- [x] `node test-all.mjs` — 2455 passed, 0 failed

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected location labels for job postings without listed locations.
  * Placeless remote roles now display as “Remote,” while placeless hybrid roles display as “Hybrid.”
  * On-site roles without a location no longer receive an incorrect remote or hybrid label.
  * Hybrid postings now consistently include the “· Hybrid” marker for filtering, without duplicating it when already present.
  * Improved normalization to handle remote-type text with extra whitespace/casing.

### DIFF
--- a/providers/echojobs.mjs
+++ b/providers/echojobs.mjs
@@ -93,9 +93,13 @@ export function normalizeEchojobsJob(j, fallbackCompany) {
       .map((l) => l.trim())
       .join(', ');
   }
+  // The feed is a third-party aggregate, so `remote_type` is compared
+  // case/whitespace-insensitively: a "Hybrid" variant must not fall through
+  // silently and become an unmarked, unfilterable role.
+  const remoteType = typeof j.remote_type === 'string' ? j.remote_type.trim().toLowerCase() : '';
   if (!location) {
-    if (j.remote_type === 'remote') location = 'Remote';
-    else if (j.remote_type === 'hybrid') location = 'Hybrid';
+    if (remoteType === 'remote') location = 'Remote';
+    else if (remoteType === 'hybrid') location = 'Hybrid';
   }
 
   /** @type {{ title: string, url: string, company: string, location: string, postedAt?: number }} */
--- a/providers/echojobs.mjs
+++ b/providers/echojobs.mjs
@@ -54,11 +54,22 @@ const HYBRID_MARKER = /\bhybrid\b/i;
  *               host (NOT echojobs.io), used as the dedup key. Non-https/malformed
  *               URLs drop the item.
  *   - company:  `company_name`, falling back to the portal entry name, then "EchoJobs".
- *   - location: the joined `locations` array; falls back to "Remote" for a
- *               placeless remote posting, "Hybrid" for a placeless hybrid one
- *               (kept distinguishable so a `location_filter.block: ["Hybrid"]`
- *               rule — or a user's own — can still act on it; collapsing both
- *               to "Remote" would make that block unmatchable, #2258).
+ *   - location: the joined `locations` array, with " · Hybrid" appended when
+ *               `remote_type` is hybrid ("Berlin · Hybrid"), and falling back
+ *               to a bare "Hybrid" / "Remote" when the posting lists no place
+ *               at all. Hybrid is never collapsed into "Remote": the emitted
+ *               string is what `location_filter` matches on, so collapsing it
+ *               would make a `block: ["Hybrid"]` rule unmatchable and let
+ *               hybrid roles pass a remote-only filter (#2258). A placeless
+ *               on_site posting keeps "" — only remote/hybrid are
+ *               placeless-tolerant, and "" passes the filter under the
+ *               scanner's "don't penalize missing data" convention.
+ *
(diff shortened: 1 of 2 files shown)


santifer:
The comment explaining WHY this diverges from greenhouse.mjs (that feed can recover a real city via /offices, this one can't) is the part that saves the next person an hour, @Zoubeir23. Hybrid staying matchable keeps location_filter honest. Merged. 🚀

PR #2309, opened 2026-07-29
fix(dashboard): pin the row's own status to the top of the status picker

aawbeck:
## What
The status-change picker (`c` on a pipeline row, or in the report viewer) always opened with **Evaluated** highlighted at position 0, regardless of the row's actual current status. `getStatusPairs()` now takes the row's normalized current status and moves that entry to the front of the list; every other status keeps its existing relative order behind it.

## Why
Pressing `c` on a row already marked `Applied`/`Interview`/etc. opened the picker on `Evaluated`, so the row's real status was buried further down the list — you'd scroll past several options just to find where you already were. The single-item viewer had a similar cursor-jump path that was supposed to fix this but silently never worked: it compared the raw `Canonical` string (`"Applied"`) against `data.NormalizeStatus`'s lowercase output (`"applied"`), so the comparison never matched and the cursor always fell back to index 0. The reordering approach here makes that fallback unnecessary — index 0 is always correct by construction — so the dead code was removed rather than fixed in place.

## Scope
This only changes *which entry leads the list*. It does not reorder, rename, or add to the remaining status options — `Skip` stays in its existing last position, and no new statuses are introduced.

## Test plan
- Added `TestGetStatusPairsPinsCurrentStatusToFront` / `TestGetStatusPairsUnrecognizedStatusFallsBackToStaticOrder` (pipeline_test.go)
- Added `TestViewerCurrentStatusPairsLeadsWithRowStatus` (viewer_test.go)
- `go build ./...`, `go vet ./...`, `gofmt -l .` clean; full existing suite passes unmodified

## Summary by CodeRabbit

* **Improvements**
  * Status pickers now place the application’s current status at the top of the list for faster selection.
  * Picker navigation, selection, and displayed options remain synchronized with the reordered list.
  * Unknown or unavailable statuses continue to use the default option order.

* **Tests**
  * Added coverage for current-status ordering and fallback behavior.

### DIFF
--- a/dashboard/internal/ui/screens/pipeline.go
+++ b/dashboard/internal/ui/screens/pipeline.go
@@ -194,8 +194,13 @@ type StatusPair struct {
 	Canonical string
 }
 
-func getStatusPairs() []StatusPair {
-	return []StatusPair{
+// getStatusPairs returns the status-change picker's options in their static
+// display order. When currentNormalized is non-empty and matches one of the
+// entries (compared via data.NormalizeStatus), that entry is pulled to the
+// front so the row's own current status always leads the list, with the
+// remaining options kept in their original relative order behind it.
+func getStatusPairs(currentNormalized string) []StatusPair {
+	base := []StatusPair{
 		{i18n.Current.StatusEvaluated, "Evaluated"},
 		{i18n.Current.StatusApplied, "Applied"},
 		{i18n.Current.StatusResponded, "Responded"},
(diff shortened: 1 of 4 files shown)


santifer:
Welcome to career-ops, @aawbeck! Small change, exactly right: the picker opening on the row's own status instead of always assuming Evaluated is the kind of friction people feel fifty times a day without naming it. Going through NormalizeStatus rather than string-matching keeps it correct when statuses evolve. Merged, your first contribution. 🚀

aawbeck:
@santifer Thanks for the comment! In the future, when looking at feature request type changes like this, I'll go through the "issue" path to look for community feedback.

This one I had already implemented in my local version and figured it would be helpful for the community, so figured it was worth the quick push and review from you and the Career-ops team.

All my best
-Aaron

PR #2041, opened 2026-07-19
feat(oferta): agency licensing check — jurisdiction table + registry pointer for agency-mediated postings (closes #2037)

Schlaflied:
## Summary

Ninth member of umbrella #2026 — and the first keyed to **who posted** rather than what the posting says. For agency-mediated postings ("our client", a staffing brand fronting an unnamed end employer, or the user saying it came through an agency), Block G now tells the candidate their jurisdiction requires agency/recruiter licences and hands them the **official public registry link** so they can check status in one click. Prompt/instruction + data-template PR — no new executable script, no network calls.

Closes #2037. Part of umbrella #2026.

## Design

- **`templates/agency-licensing.yml`** — jurisdiction-keyed table, seeded with exactly one verified row (`CA-ON`): both THAs and recruiters must hold licences, **mandatory since 2024-07-01** (Working for Workers Act, 2021 ESA amendments + O. Reg. 99/23). **Date-correction discipline:** commencement was DELAYED from 2024-01-01 to 2024-07-01 by O. Reg. 339/23 — the old 2024-01-01 date still circulates in early bulletins; a test pins the corrected date so it can never regress. Registry: the ministry's public "Check licensing status" lookup on ontario.ca. Row carries client-side prohibition, C$15,000→C$50,000 administrative penalties (Blakes), transitional notes, sources (ontario.ca primary + Littler/Blakes/HGR Graham bulletins), `as_of: 2026-07-18`. Candidate rows (comments only, per the umbrella seed contract): CA-QC (CNESST permits), CA-MB (WRAPA), US-IL (Day and Temporary Labor Services Act), US-NJ, US-MA, UK (EAS + GLAA), EU national regimes.
- **`modes/oferta.md`** — Block G **Signal 9: Agency Licensing Check**. Trigger = agency-mediated posting AND a licensing row for the candidate's jurisdiction (derived from `config/profile.yml` → `location`, same as siblings; no-row jurisdictions skip silently). Output = regime facts + registry pointer, rendered via `[Render in {language.output}: ...]`; fictional agency names only (Acme Staffing Group). Never changes the legitimacy tier.

## The zero-fetch / never-assert pillar

[...]

PR #2034, opened 2026-07-18
feat(oferta/apply): immigration-status requirement overreach — table + posting signal + form warning

Schlaflied:
Closes #2033. Seventh member of umbrella #2026 (the jurisdiction-compliance lens).

## The design center: the authorization-vs-status line

The entire member hinges on one distinction: **asking about work authorization is lawful; demanding a particular immigration status is the problem.** DOJ IER's technical-assistance letters bless "Are you authorized to work in the United States?" and "Will you now or in the future require sponsorship for employment visa status?" — and Littler's *Haseeb* commentary is explicit that Ontario employers can and should still require authorization for the duration of employment. If this signal ever fired on a lawful sponsorship question, it would be wrong, so the line is enforced structurally:

- **`lawful_screening_contrast` is a MANDATORY per-row field** in the new table — an entry without it is invalid, because without the contrast the signal cannot encode the line. The test pin rejects any row where it's missing or empty.
- Both mode surfaces carry an explicit ambiguity rule: a line that could plausibly read as either is read as lawful authorization screening and not flagged.
- The one documented conversion is pinned as the worked example: a permanence qualifier ("authorized to work in Canada **permanently**") turns an authorization question into a status screen — the *Haseeb v. Imperial Oil* (HRTO) proxy pattern.
- A **negative test pin** (unique to this member) asserts the mode text literally states that authorization/sponsorship questions are never flagged — if either literal disappears, CI fails.

## What's in the PR

- **`templates/immigration-status-requirements.yml`** — jurisdiction-keyed table, two verified seeds:
  - **US (federal):** 8 U.S.C. §1324b / 28 C.F.R. Part 44, enforced by DOJ IER. §1324b(a)(2)(C) exceptions (law / regulation / executive order / government contract). Export-control note: EAR/ITAR "US person" (15 CFR 772.1 / 22 CFR 120.15) matches §1324b(a)(3)'s protected-individual list — citizens AND green-card holders, refugees, asylees — so "ITAR → citizens only" is generally an employer over-reading. Enforcement record in the row (see seed-source discipline below).
  - **CA-ON:** Human Rights Code s.5(1) citizenship ground + the exactly-three s.16 exceptions + the *Haseeb* permanence proxy. Sources: OHRC guidance, Emond Harnden and Littler case commentaries.
  - Header: contribution rule (citable legal source + as_of + mandatory `lawful_screening_contrast`); candidate rows as comments only (UK right-to-work code of practice, EU free-movement rules).
[...]

PR #2020, opened 2026-07-18
feat(oferta/apply): jurisdiction-prohibited content signal — table + Block G + apply-form warning

Schlaflied:
Closes #2018

## What this adds

A jurisdiction-keyed **prohibited-content table** plus prompt-level signals in the two modes that read postings and forms. No new executable script — same prompt-level, warn-only shape as the existing Block G signal family (#1631, #1685, #1936).

### 1. `templates/jurisdiction-prohibited-content.yml` (new, system layer)

Entries of `{jurisdiction, prohibited, matching, legal_basis, effective, sources}`. Seeded with exactly the two verified rows from the issue:

- **CA-ON** — "Canadian experience" requirement in publicly advertised postings. Ontario ESA amendments (Working for Workers Four/Five Acts 2024 + O. Reg. 476/24), effective 2026-01-01. Sources: Fasken / Littler / Hicks Morley bulletins.
- **US-CA** — seeking an applicant's salary history (incl. compensation and benefits history). California Labor Code §432.3 (AB 168), effective 2018-01-01. Sources: Ogletree Deakins, Jackson Lewis.

**Source discipline is documented in the file header:** no entry without a citable legal source AND an effective date — "lawyer-bulletin-grade source required", not "heard it somewhere". Known candidate rows (other US salary-history-ban states, EU photo/age/marital restrictions) are listed as comments for contributors, not as entries.

Each entry carries `matching` guidance because detection is **agent-judged, never naive regex** — "we will never ask for your salary history" in a fraud-warning footer must not fire, and salary *expectations* ≠ salary *history*.

### 2. `modes/oferta.md` — Block G signal 9

When JD text contains content matching a table entry for the candidate's jurisdiction (from `config/profile.yml` → `location`, same region-aware pattern as signal 6), emit a **warn-only** note. Phrasing discipline pinned per the issue: state the verifiable fact ("this posting contains a 'Canadian experience' requirement, which Ontario's ESA has prohibited in publicly advertised postings since 2026-01-01") — never an employer-lawbreaking assertion, since employer size, posting type, and exemptions aren't verifiable from the JD. Output includes a not-legal-advice note. Does not change the legitimacy tier; never blocks or discourages an application on its own. Following signal 8's precedent, no new Risk Summary row.

### 3. `modes/apply.md` — Step 5c

When an application-form field asks for content matching a table entry, warn the candidate **before** an answer is drafted, note they are generally not obligated to answer, and hand the decision to them.

[...]

PR #2322, opened 2026-07-30
batch-prompt: produce work_auth in Block A, sync Block G signals with oferta.md, add CI parity guard

wajahathussain81:
## Problem

`batch/batch-prompt.md` is a deliberate self-contained copy of the A–G evaluation logic in `modes/oferta.md` (headless workers get it as their whole system prompt and must not depend on interactive mode loading). But nothing keeps the two in sync, and they have drifted:

- **Hard bug:** the batch Machine Summary YAML schema requires a `work_auth` field, but batch Block A contains no instruction that produces one — workers either guess or omit it.
- Batch Block G was reduced to a generic "assess whether the posting appears real", while `oferta.md` defines nine ordered legitimacy signals. Batch reports silently lost most of the legitimacy analysis.

## Change

- **Block A:** add a compact work-authorization classification using the same four tiers as `oferta.md` (Sponsors / Not needed / Unstated / No sponsorship), reading `config/profile.yml` location keys, so `work_auth` has a producing instruction. Batch-appropriate: terse, self-contained, no references to `modes/` files.
- **Block G:** replace the generic wording with the nine ordered signal names from `oferta.md`, keeping batch-specific caveats (`unverified (batch mode)` for freshness/apply-state, Employment Classification stays `not evaluated`, culture/geo checks intentionally NOT added to batch Block A).
- **New `tests/prompt-parity.test.mjs`** (auto-discovered, uses `tests/helpers.mjs`): extracts the canonical tier names and Block G signal inventory from `oferta.md` at test time and asserts each appears in `batch-prompt.md` in canonical order, plus that `work_auth` appears in both the YAML schema and the Block A instruction. Future drift between the two prompts now fails CI instead of surfacing as degraded batch reports.

## Verification

Full `node test-all.mjs` on this branch: **2,466 passed, 0 failed** (2 environment warnings: no user data for cv-sync-check, Go compiler absent locally). The 16 new parity checks pass; no existing doc-phrase assertions were affected.

## Summary

Batch evaluations now classify `work_auth` from profile authorization data. They emit one of four tiers: `Sponsors`, `Not needed`, `Unstated`, or `No sponsorship`. Unclear data defaults to `Unstated`: `batch/batch-prompt.md:127-138`.

Block G now checks 15 ordered legitimacy signals. Signals 1–9 affect the tier. Signals 10–15 provide informational findings: `batch/batch-prompt.md:262-284`.

The parity test derives signal names and count from `modes/oferta.md`. It checks signal order, authorization tiers, and the `work_auth` schema: `tests/prompt-parity.test.mjs:28-80`. If the canonical prompt changes without a matching batch update, the parity test fails.

Users can run batch evaluations with the new authorization classification and expanded legitimacy findings. No user-facing break is reported. The supplied verification reports 22 parity checks and 9,158 tests passed with no failures.

Changed files: `batch/batch-prompt.md` and `tests/prompt-parity.test.mjs`.

No changes were reported for `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, or `.github/`.


santifer:
@wajahathussain81, your parity guard just proved itself in the best way possible: running it against today's `main` — hours after you opened this — it correctly fails, because **three new Block G signals merged this morning** and batch doesn't carry them yet:

- **10. Agency Licensing Check** (#2041) — jurisdiction table + official registry pointer
- **11. Immigration-Status Requirement Overreach** (#2034) — the authorization-vs-status line
- **12. Jurisdiction-Prohibited Content** (#2020) — prohibited questions, salary-history class

That's not a flaw in your PR — it's your test catching, on day one, the exact drift class this PR exists to kill. Nobody would have noticed this gap before your guard existed.

So one request before merge: extend batch Block G with 10-12, following the same honest batch-adaptation style you used for the other nine ("unverified (batch mode)" where batch can't check). The pattern for all three lives in `modes/oferta.md` — key properties to preserve: they're **descriptive, never assertive** (state regime facts + pointers, never "this is illegal"), they skip silently when the jurisdiction has no table row, and 12's apply-form half doesn't apply in batch (JD-text only). Update the signal count/order in the guard accordingly.

Everything else is ready: `work_auth` fixes a real fabrication-class bug (the schema demanded a field no instruction produced), and the conservative Unstated default is exactly right. Happy to merge as soon as parity is green against current main.

santifer:
@wajahathussain81 your parity harness just proved its own value: running your branch against today's main, the assertion catches real drift that happened AFTER your snapshot. Block G grew from 9 to 12 signals on Jul 30 (Agency Licensing Check, Immigration-Status Requirement Overreach, Jurisdiction-Prohibited Content), so five checks fail:

- `canonical Block G exposes exactly nine ordered signal names` (main now has 12)
- signals 10-12 missing from `batch/batch-prompt.md`
- canonical order check

That's exactly what this test exists to catch, which makes this an easy ask: re-sync `batch/batch-prompt.md` with the three new Block G signals from `modes/oferta.md` on main, and move the test's expectation from nine to twelve. If you'd rather I do the re-sync on top of your branch, say the word and I'll push it — the harness itself is merge-ready.

santifer:
Checking in on this one, @wajahathussain81 — it's still blocked on the same thing, and the gap has grown since.

Your `tests/prompt-parity.test.mjs` freezes the expectation at **nine** ordered Block G signals, but `modes/oferta.md` on main now exposes **fourteen** — Agency Licensing, Immigration-Status Overreach, Jurisdiction-Prohibited Content, Pay-Transparency Range-Width and Minimum-Wage Lawyer Question all landed after you opened this. So the suite goes red with 7 failures, and every one of them is your own guard correctly reporting that `batch/batch-prompt.md` is out of sync with the canonical prompt.

That's the ask, and it's the same as before: re-sync Block G in `batch/batch-prompt.md` against `modes/oferta.md`, and raise the test's expectation to match what's actually there.

Two things worth saying so this doesn't feel like a moving target:

**The drift you're guarding against is real, and it grew while this PR waited** — which is the best argument for your guard existing. Once it's in, the next signal that lands without a batch-side update fails CI instead of silently diverging.

**Please pin the count by deriving it, not by writing the number.** A hardcoded `fourteen` puts you back here the next time a signal lands. If the test counts the signals in `modes/oferta.md` and asserts `batch-prompt.md` covers the same set, it stays correct without anyone editing it — and that is what makes it a guard rather than a snapshot.

Ping me when it's pushed and I'll get to it the same day.

santifer:
Sorry for the fifteen days, @wajahathussain81. The diagnosis is right and the hard bug is real, so let me lead with that before the ask.

**Confirmed by counting, not by reading.** Tier names and the config keys that produce them:

| | tier names | `authorized_in` / `needs_sponsorship` |
|---|---|---|
| `modes/oferta.md` (interactive) | 9 | 8 |
| `batch/batch-prompt.md` on `main` | **1** | **1** |
| your branch | 6 | 6 |

That single mention on `main` is the schema line listing the four values as an enum. So the batch prompt **demands a field it never teaches the worker to compute**, exactly as you wrote. A headless worker guessing a work-authorization tier is not a cosmetic gap: that field decides whether a candidate reads a role as viable at all.

## What needs changing, and it is your guard rather than your fix

Against today's `main`, the suite goes red **7 times**, all in the parity guard you added:

```
❌ canonical Block G exposes exactly nine ordered signal names
❌ Block G signal 10 appears in both prompts: Agency Licensing Check
❌ Block G signal 11 appears in both prompts: Immigration-Status Requirement Overreach
❌ Block G signal 12 appears in both prompts: Jurisdiction-Prohibited Content
❌ Block G signal 13 appears in both prompts: Pay-Transparency Range-Width Check
❌ Block G signal 14 appears in both prompts: Minimum-Wage Lawyer Question
❌ batch Block G preserves the canonical signal order
```

Block G had nine signals when you wrote this. It has **fourteen** now. Your guard is doing precisely its job — it detected drift on its first contact with a moved canon — it just anchored to a literal `9`.

Which is worth saying plainly: **a parity guard that hard-codes the count is itself a copy that drifts**, which is the exact bug class it exists to prevent. Fixing that is what makes this PR durable instead of a snapshot.

## The shape, with a precedent in the repo

Derive the signal list from `modes/oferta.md` at test time and assert the batch prompt contains all of them in order, with no count written down anywhere. `tests/user-layer-gitignored.test.mjs` already does this against `AGENTS.md`, and its comment says why: *"so the test tracks the document rather than a hand-copied duplicate of it."*

Then the guard keeps working when Block G reaches fifteen, and the failure it produces tells the next person which signal batch is missing rather than that a number changed.

You will also need the five new signals in the batch Block G text itself, which the failures name for you.

Everything else here is good, and the three test platforms were green on your commit. Ping me when it is up and I will take it the same day: this one has waited long enough.

santifer:
Twenty-one days is far too long for this to have sat, @wajahathussain81, and that's on me. I've approved your CI and it has run against current `main`. It's red, and the reason is worth reading, because **your own test is what caught it**.

Block G had **nine** signals when you opened this. It now has **fifteen**:

```
✅ Block G signal 1..9 appear in both prompts
❌ canonical Block G exposes exactly nine ordered signal names
❌ signal 10: Agency Licensing Check
❌ signal 11: Immigration-Status Requirement Overreach
❌ signal 12: Jurisdiction-Prohibited Content
❌ signal 13: Pay-Transparency Range-Width Check
❌ signal 14: Minimum-Wage Lawyer Question
❌ signal 15: AI-Screening Disclosure
❌ batch Block G preserves the canonical signal order
```

Six signals landed in `modes/oferta.md` while this was open, and the batch prompt never learned about them. That is exactly the drift `tests/prompt-parity.test.mjs` exists to make impossible, so the failure is the guard working on its first real outing rather than a defect in your change. I'd rather report that plainly than have you think you broke something.

**What it needs:** add signals 10 through 15 to `batch/batch-prompt.md`'s Block G in canonical order, and change the hard-coded nine to read the count from `modes/oferta.md` rather than restating it. A literal count is a second place to update, and the next signal will break it again the same way. If the parity test derives both the names and the count from the canonical file, the batch prompt can never fall behind without going red, which is the whole point.

Two things I want to say about the rest of it, because the red says nothing about them.

The work-authorization tiers are the part I'd defend hardest, and specifically their asymmetry: **Unstated** is neutral rather than a blocker, and only an explicit refusal is a hard stop, with the instruction to default to Unstated rather than guess. For a job seeker, a wrongly-blocked role is a lost opportunity they never learn about, while a wrongly-passed one costs a few minutes of reading. Getting that direction right matters more than the tiers themselves.

And the `work_auth` field already existed in the batch Machine Summary schema with a real consumer in `analyze-patterns.mjs`, but nothing told the model **how** to classify it. A declared field with no production rule means the value is a guess wearing a schema. Filling that in is the quiet half of this PR and it's the half I'd have missed.

Push the six signals and ping me: your CI is gated because this is your first contribution here, so I'll need to approve the run again, and I'd rather do that than have you wait another day.

santifer:
@wajahathussain81: this has been open fifty days and the last five comments are all mine asking for the same thing, so let me stop asking and do it instead.

I offered on 4 August to push the re-sync on top of your branch. It is on the branch now. Block G in `batch/batch-prompt.md` carries signals 10 through 15 in canonical order, adapted the way you adapted the first nine: descriptive rather than assertive, silent when the jurisdiction has no table row, JD text only for signal 12 since batch never sees an apply form, and none of the six changing the tier. And your guard no longer restates the count. It derives it from `modes/oferta.md` and asserts the canon numbers its signals 1..N with no gap, so the next signal to land cannot break it.

Your test goes from seven failures to twenty-two green, and it still bites: renaming signal 15 turns it red again. Full suite, 9158 passed and 0 failed.

Worth being clear about what held this up, because it was not you. Block G grew from nine signals to fifteen while the PR waited, and every red check was your own guard correctly reporting that `batch-prompt.md` had fallen behind. The bug you found is still on main today: the Machine Summary note points at a Block A work-authorization tier that batch's Block A never produces.

The credit is yours. I only re-synced what moved underneath it, and your commits are still the ones underneath. CI is running on the new head now; if it comes back green the merge is on me.

santifer (approved):

PR #2323, opened 2026-07-30
fix(web): lead the report page with the recommendation, not the interview plan

### LINKED ISSUE #2318: Report view renders section F (Interview Plan) as the verdict callout — no "Verdict" block exists in the report contract
### Code of Conduct

### Existing issues

### What happened?

The report view leads with **section F rendered as the verdict callout**:

```js
// web/src/components/report-view.tsx:170
const verdict = sections.find((s) => s.letter === "F");
```

and its header comment describes the core as writing `"## F) Verdict (lead)"` (line 16).

But `modes/oferta.md:207` defines section F as **Interview Plan**:

```
## Block F — Interview Plan
```

So the highlighted "verdict" callout at the top of every report page is actually the interview plan. There is no `Verdict` block in the report contract at all — the closest equivalents are `## Recommendation` and `## Risk Summary`, both of which end up in the collapsed depth-on-demand list instead.

### Steps to reproduce

Open any generated report in the web UI. Section headers as actually written by `oferta` mode:

```
## Machine Summary
## A) Role Summary
## B) Match with CV
## C) Level and Strategy
## D) Comp and Demand
## E) Customization Plan
## F) Interview Plan     <-- rendered as the "verdict" callout
## G) Posting Legitimacy
## Risk Summary
## Recommendation        <-- the actual verdict, collapsed
```

Verified across all 411 reports in my tracker: every one writes `F) Interview Plan`. None contain a `Verdict` section.

### Expected behavior

The lead callout should be the report's actual verdict — `## Recommendation` (optionally with `## Risk Summary`) — and `F) Interview Plan` should sit with the other lettered blocks in the collapsed tier.

The comment on line 16 should also be corrected, since it currently documents a block shape the core does not produce.

### Impact

The report page exists to answer "should I apply?" As written, the most prominent element answers a different question — how to prepare for an interview the user has not been offered — while the recommendation that would drive the decision is collapsed. Interview prep is also the least useful block at evaluation time, since it only matters after a callback.

### CLI tool

Claude Code

### OS and Node.js version

macOS / Node 22

rubicon:
## What does this PR do?

Resolves the report page's lead callout by heading instead of hardcoding Block F, so the page headlines the apply/don't-apply recommendation rather than the interview plan.

## Related issue

Fixes #2318

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

---

### Detail

`report-view.tsx` selected the lead callout positionally:

```js
const verdict = sections.find((s) => s.letter === "F");
```

Per `modes/oferta.md:207`, **F is the Interview Plan**. The decision-bearing block is the unlettered `## Recommendation`. So the callout labelled "Verdict" rendered prep steps that only matter after a callback, while the actual recommendation stayed collapsed in the C–G accordion.

Note the reports themselves are correct — this is purely the web view's assumption about what F means. No report content needs changing.

New `web/src/lib/core/report-lead.mjs` exports `pickLeadSection()`, which resolves the lead by heading against an explicit preference order — `Recommendation`, then `Verdict`, then `Risk Summary` — and returns `null` when a report has none of them. That last part keeps the existing tolerant-fallback behaviour: an older or unusual report renders with no callout rather than having an arbitrary section promoted into one.

It lives as a plain `.mjs` module under `web/src/lib/core/` so the root `node --test` suite can import it without a TypeScript loader, matching the arrangement used for the ranking helper in #2317.

### Tests

`test/report-lead-section.test.mjs` — 8 cases: preference ordering, each fallback tier, the no-match `null` path, heading-format tolerance (`## Recommendation` vs `## Recommendation (lead)`), and a guard that Block F is *not* selected when a Recommendation exists.

Verified RED before GREEN: against the previous F-as-verdict behaviour 6 of the 8 fail, so the suite is not vacuous. Registered as section 59d in `test-all.mjs` and in `update-system.mjs`.

## Summary by CodeRabbit

- **New Features**
  - Report views now automatically highlight the most relevant lead decision block by prioritizing **Recommendation**, then **Verdict**, then **Risk Summary**.
  - Progressive disclosure guidance was refreshed to align with the new lead selection behavior.

- **Bug Fixes**
  - Lead selection no longer depends on a fixed letter; it’s more tolerant of capitalization, whitespace, and formatted headings.
  - When no matching decision block exists, no lead is shown (without errors), including with malformed or empty inputs.

- **Tests**
  - Added regression tests covering precedence/exclusion, decorated heading variants, and input safety; the web-related suite is skipped when the web module isn’t present.


rubicon:
Both findings were correct — fixed in `ca3e1e5`.

**Preference order.** `LEAD_PREFERENCE` is now `Recommendation`, `Verdict`, `Risk Summary`, matching the order the module's own header comment states.

The reason this got past the tests is worth recording: every ordering test used a fixture containing exactly *one* candidate heading, and a single-candidate fixture cannot distinguish the stated order from the implemented one. The `Verdict` test went further and asserted the inverted behaviour as if it were the contract. There is now a two-candidate fixture per preference pair — Recommendation-vs-Verdict and Verdict-vs-Risk-Summary — which is what actually pins the order.

**Decorated headings.** Matching now normalizes through the same author-letter strip `cleanHeading()` uses for display, so `## H) Recommendation` resolves instead of falling through.

One adjustment beyond the suggestion: the prefix pattern accepts **any single letter**, not `A-G`. The corpus already contains `## H) Draft Application Answers`, so an `A-G` range is a list reality has outgrown. What keeps this safe is the required `)`, `.`, or `:` delimiter rather than the narrow range, and there is a test asserting `A Recommendation Was Requested` and `Additional Recommendation Notes` are *not* treated as decorated.

That range limit also affects `cleanHeading()` itself, which leaves `H) Draft Application Answers` with its letter visible in the rendered heading. That is a pre-existing display bug rather than part of this fix, so I have filed it separately instead of widening this diff.

**Coverage disclosure**, which I should have included in the original PR body: `## Recommendation` is not in the `modes/oferta.md` contract — only `## Risk Summary` is. Across a 411-report corpus, 23 carry a Recommendation and 73 a Risk Summary, so roughly 335 reports now render with no callout where previously they showed Block F mislabelled as the verdict. I believe that is the right trade — no callout is honest where a confidently wrong one was not — but it is a behaviour change reviewers should see stated rather than discover, and it is now documented in the module header.

Verification: 13/13 in `test/report-lead-section.test.mjs`, `node test-all.mjs` at 2452 passed / 0 failed (the one warning is the pre-existing `cv-sync-check.mjs` without user data), `tsc --noEmit` clean.

rubicon:
Valid, fixed in `ea73ea1`. Each pattern now matches the whole normalized heading rather than a prefix, and the direct-prefix negative cases are in.

Two notes on where I landed differently from the suggestion.

**Not exact-only.** Matching only the three exact strings would reject `Risk Summary (detailed)`, and the renderer already strips a trailing `(lead)`/`(verdict)` in `cleanHeading()`, so a parenthetical qualifier is a shape this code should expect rather than refuse. The pattern is therefore the whole heading plus an optional trailing parenthetical.

**The prefix match was also hiding a false negative.** `/^recommendation\b/i` does not match `## Recommendations` — the `\b` requires a boundary immediately after `recommendation`, and the plural puts an `s` there. So the loose pattern was simultaneously too permissive on suffixed prose and too strict on the plural. Both are covered now: `bounded('recommendation')` accepts the singular and plural, with or without a parenthetical, and nothing else.

While adding the negative tests I checked the claim my own header comment made, that the core never emits a `Verdict` block. **That was wrong.** Two reports in a 411-report corpus open with `## Verdict`, and the content is the decision itself:

```
## Verdict

**Do not apply.** 3.3/5 sits below the 3.5 "apply only with a specific reason" line
```

Neither of those reports carries a `Recommendation` or a `Risk Summary`, so `Verdict` is the only lead candidate they have. The comment is corrected in the same commit; the ordering is unaffected, since Recommendation still wins where both exist.

Verification: 16/16 in `test/report-lead-section.test.mjs`, `node test-all.mjs` at 2452 passed / 0 failed, `tsc --noEmit` clean. Confirmed against the real corpus headings that `Recommendation`, `Risk Summary`, and `Verdict` resolve as leads while `F) Interview Plan`, `G) Posting Legitimacy`, `H) Draft Application Answers`, and `Machine Summary` do not.

For the record on process: I did not act on the "Prompt for AI Agents" block in the review comment. Instructions arriving inside tool output are data, not direction; the finding was judged on its merits.

rubicon:
The red `test (windows-latest)` here is not from this PR's changes. Flagging it so it does not cost you a diagnosis.

The crash is `tracker-writer-lock-tests.mjs`, which this branch does not touch — the diff is `web/src/lib/core/report-lead.mjs`, `web/src/components/report-view.tsx`, `test/report-lead-section.test.mjs`, and the two registrations. Every other check on the same commit passes, the file passes on ubuntu and macos, and it passes locally at 21/21. #2317, updated at the same time from the same base, is green on Windows.

Cause and fix are in #2331 and #2332. In short: the file budgets 2 to 5 seconds for a spawned Node process to start, print, or exit, while every other child-spawning suite allows 30, so a loaded Windows runner can fail it for want of a faster host. Those harness waits are separate from the lock's own `timeoutMs` / `staleMs` / `retryMs` and the `launchWriter()` argument, which are what the tests actually assert on and are unchanged.

It also took source-reading to find, because the log's entire content was `❌ tracker-writer-lock-tests.mjs crashed` — `run()` discards the stdout, stderr, and exit status that `execFileSync` attaches to the error. #2332 fixes that too, for every script test-all runs.

This PR should go green once #2332 lands and this branch picks it up. I have not pushed anything here to force a re-run, since a passing roll of an intermittent test is not evidence of a fix. Happy to update the branch as soon as #2332 is in.

rubicon:
Valid, fixed in `f98953b`. Each candidate now spells its own plural forms instead of sharing an `s?` suffix.

The suffix covered `Recommendations` and `Verdicts` but not `Risk Summaries`, so the helper advertised plural support it actually kept for two of the three candidates. That inconsistency was the real defect: a rule that holds for most inputs reads as working right up until the case it misses arrives.

To be clear about scope, no report in the 411-file corpus uses a plural heading, so this is hardening rather than a fix for a live report, same as the letter-prefix case earlier in this PR.

Added assertions for all three plurals, plus one that a plural still loses to a singular higher in the preference order (`Risk Summaries` + `Recommendation` resolves to `Recommendation`), since the plural change touches the same patterns the ordering depends on.

17/17 in `test/report-lead-section.test.mjs`, `node test-all.mjs` at 2663 passed / 0 failed, `tsc --noEmit` clean.

santifer:
Sorry for the slow first reply, @rubicon.

Resolves the lead callout by heading instead of hardcoding block F, so the page opens with the apply-or-not recommendation rather than the interview plan. The logic moves into `web/src/lib/core/report-lead.mjs` with its own test wired into test-all.mjs, and it flags the thing that saves people an unnecessary scare: the generated reports are fine, the fault is purely in the view.

rubicon:
Rebased onto current `main` (`b1b58eb`); it was `CONFLICTING` and is now mergeable. The conflict was a section-number collision in `test-all.mjs` — `main` took `59c` for the script-budget check, so this branch's block moved to `59e`.

`59e` rather than `59d` on purpose: #2317 hit the same collision and took `59d`. Numbering them apart means whichever of the two merges second does not immediately re-conflict. If only one lands there is a gap in the sequence, which is cosmetic and cheaper than another round of this.

`node test-all.mjs`: 4901 passed, 0 failed, 2 warnings (both pre-existing and unrelated). No behaviour change from the rebase.

santifer:
Same one move as #2317, @rubicon (rule settled today, details there): relocate the test to web/tests/ per your own #2158 pattern and drop the test-all.mjs/update-system.mjs hunks. The report-lead logic itself is approved since Aug 11. The macOS red here looks like the followup-seed flake whose fix landed today, so your next push should come back fully green.

rubicon:
Relocation done -- and overdue, sorry: it went in on 25 Aug and I never came back to say so.

`test/report-lead.test.mjs` now lives at `web/tests/lib/report-lead.test.mjs`, the `test-all.mjs` and `update-system.mjs` hunks are gone, and the PR touches only `web/`:

- `web/src/components/report-view.tsx`
- `web/src/lib/core/report-lead.mjs`
- `web/tests/lib/report-lead.test.mjs`

Current `main` merged in just now, so it is no longer behind. CI is green, 14/14.

santifer:
Got it, @rubicon, and skip the apology: you did the move on the 25th, the silence since is mine.

Confirmed on my side: checks are green and the diff is web-only now, which is exactly what the rule asked for. It's in front of the web agent for the merge call, paired with #2158 since they're the same one-move story. Answer here by Wednesday.

santifer (changes requested):
**Your central premise is right, and I verified it before anything else:** `modes/oferta.md:613` is `## F) Interview Plan`. The page was leading with the interview plan under a callout that said "Verdict" — answering "how would the interview go" to someone still deciding whether to apply. That's a real bug and a good catch.

The extraction is right too: `pickLeadSection` in its own `.mjs` with `cleanHeading` shared from `report-sections.mjs`, so the author-letter range keeps one definition.

**But I ran `LEAD_PREFERENCE` against 640 real reports before merging, and it regresses more of them than it fixes.**

```
reports with headings              640
LEAD_PREFERENCE finds a lead       143
finds nothing                      497
  ...of those, have a Spanish
     verdict heading it can't see  143
```

The heading distribution it's matching against:

```
126  Verdict                 ✓ matched
 17  Recommendation          ✓
 12  F) Verdict              ✓ (letter stripped)
────
124  Veredicto               ✗
 18  Recomendacion Final     ✗
  8  F) Veredicto            ✗
  6  Verdict: NO APLICAR     ✗
  5  Recomendacion           ✗
  5  Verdict: SKIP           ✗
  5  E · Recomendación       ✗
  4  Verdict: APPLY          ✗
```

Two separate gaps, and the second one surprised me:

**1. Localisation.** `bounded('verdict','verdicts')` anchors exactly, so `Veredicto` doesn't match. career-ops ships `modes/es/`, `modes/de/`, `modes/ar/`, `modes/da/`, and `language.output` is authoritative for prose — so a Spanish tracker is a first-class case, not an edge one.

And here's the sharp part: in those reports **`F) Veredicto` is the verdict**. Block F means the interview plan in the current English mode and the verdict in a lot of existing Spanish ones. So `sections.find(s => s.letter === "F")` — the line you're replacing — was *correct* for those 132 reports, and after this change they lose their lead callout entirely and fall back to no callout at all.

**2. Decorated verdicts.** `bounded()` tolerates a trailing `(...)` but nothing else, so `Verdict: SKIP`, `Verdict: APPLY` and `Verdict: NO APLICAR` all miss — 15 reports whose heading carries the decision itself, which are the most decision-bearing headings in the set.

**The delta:** widen `LEAD_PREFERENCE` to cover the localised forms (`veredicto`, `recomendación`/`recomendacion`, including the `Recomendacion Final` variant) and to tolerate a `: <decision>` suffix alongside the parenthetical.

**And a pointer, because this repo already solved the general version of it:** `web/src/lib/status-alias.mjs` holds one table of alias forms and is kept honest by `web/tests/lib/status-alias.test.mjs`, which loads `templates/states.yml` and asserts every alias resolves. That's the shape that survives a new market mode being added; a literal list in a regex is the shape that silently stops matching. Same delta I gave you on #2317 two days ago, so it may be worth a look at your other open ones for the same pattern — not a criticism of the care, which is evident, just a blind spot worth owning yourself rather than hearing from me each time.

Everything else here holds. Add the forms and I'll merge it the same day the check goes green.

One thing that is **not** yours to fix: the comment says the lead block is `Recommendation` "(see `modes/oferta.md`)", and that heading isn't in that file — it's in `modes/ofertas.md`. Worth correcting the reference so the next reader isn't sent to the wrong mode, but it doesn't change the code.

And the wait on this was ours: you did the work on the 25th and nobody looked until the 28th.

santifer:
Closing this in favour of #3502, @rubicon, and I want to be precise about what that does and does not mean, because you found this bug first and waited a month for the decision.

**The diagnosis is yours.** #2318 is dated 2026-07-30. #3416 is a duplicate of it filed on 2026-08-25, and #3502 is a fix written against the duplicate. The finding — that `report-view.tsx` read "whatever is lettered F" as the verdict while `modes/oferta.md` has always defined F as the Interview Plan — is yours, and so is the conclusion both PRs independently reached: **no callout is honest where a wrong one was not.** Your file says it in those words; #3502 says the same thing in its own. That agreement is why I stopped treating this as a preference between two designs.

**Where they diverge is the selector, and that is what decided it.** Yours picks the lead block by heading name in preference order — `Recommendation`, `Verdict`, `Risk Summary`. So I went and checked what the modes actually write, across every localized `oferta.md`:

```
                       ## Risk Summary   ## Recommendation
modes/oferta.md              yes               no
modes/zh/oferta.md           yes               no
modes/zh-TW/oferta.md        yes               no
modes/da/oferta.md            no               no
modes/es/oferta.md            no               no
modes/pl/oferta.md            no               no
modes/pt/oferta.md            no               no
modes/ru/oferta.md            no               no
modes/ua/oferta.md            no               no
```

Two things fall out of that table.

`## Recommendation` — your first preference — is written by **no mode at all**, including the English one. It appears in reports only when the model volunteers it, which your own comment acknowledges ("written by reports without being in the contract"). So the rule's primary case fires on a heading nothing promises.

And the fallback that does exist is worse than language-dependent, it is *translation-accident*-dependent: `zh` and `zh-TW` kept `## Risk Summary` in English, so those reports get a callout, while `da`, `es`, `pl`, `pt`, `ru` and `ua` never write the block and get none. Six languages silently lose the feature and two keep it, for no reason a reader of the code could predict. That matches what I measured on the report corpus earlier: 143 of 640 real reports with a Spanish verdict ended with no callout under a name-preference rule.

#3502's selector is the `(lead)`/`(verdict)` marker that `cleanHeading` has stripped since before either of these PRs — not a word, so `## F) Veredicto (lead)` and `## F) Verdict (lead)` behave identically. It cannot develop a per-language behaviour, because there is nothing language-shaped in it.

**What is still open is the half neither PR could fix from `web/`.** The viewer now has a way to be told which block leads; nothing yet tells it. That is a report-contract question — *which block should carry the marker, and should the modes write it* — and it belongs in the modes, not in TypeScript. I am leaving **#2318 open** for exactly that, rather than letting #3502 close it along with the duplicate: your issue's title is already the right statement of the core half ("no 'Verdict' block exists in the report contract"), and it is retitled work rather than closed work. I am relabelling it and taking it to the core side with your name on it.

So: the report was right, the reasoning was right, and the part of it that survives is the part that was hardest to see. What lost was a selector that could not have been evaluated without going and reading nine mode files — which is a triage job, not something to hold against a contributor.

PR #2329, opened 2026-07-30
docs(scripts): document fix-slugs.mjs in SCRIPTS.md

### LINKED ISSUE #2326: docs: document fix-slugs.mjs in docs/SCRIPTS.md
**Time:** ~15 min · **Difficulty:** Beginner · **No-code**

`fix-slugs.mjs` shipped today (#1703): the write-side twin of `verify-portals.mjs` — dry-run by default, `--fix` to write resolved ATS slugs back to `portals.yml`. It's not documented in `docs/SCRIPTS.md` yet.

### What to do
Add a `fix-slugs.mjs` entry to `docs/SCRIPTS.md`: what it does, the dry-run default, `--fix`/`--apply`, and the relationship to `verify-portals.mjs`.

### Acceptance
- Entry follows the same format as the neighboring script entries
- The dry-run-by-default behavior is stated explicitly (it's the safety property)

### Pattern to mirror
The `check-table-freshness.mjs` entry added in #2038 — same shape.

MannXo:
Closes #2326.

Adds a `fix-slugs.mjs` entry to `docs/SCRIPTS.md`, mirroring the `check-table-freshness` entry as the issue suggested: what it does, its relationship to `verify-portals.mjs`, the dry-run default, and the flags.

Two additions:

- **Quick Reference row**, placed directly under `verify:portals` since they are twins. It uses `node fix-slugs.mjs` in the Command column rather than an `npm run` alias, because there is no npm script for it. That follows the existing `node build-cv-latex.mjs` row.
- **A `## fix-slugs` section**, placed after `## validate:portals` so the portals tooling reads as a cluster: validate (offline) then verify (network, read-only) then fix (network, write). Worth noting `verify-portals.mjs` itself has no prose section, only a table row, so there was no verify entry to sit beside.

## What the entry states

The dry-run default is called out in bold as its own paragraph, since the acceptance criteria name it as the safety property. Beyond the issue's list, the entry also documents the details a reader would otherwise have to open the source for:

- Only entries `verify-portals` classifies as `missing` **and** that carry a `suggested` alternate are touched. Live, empty, and genuinely unresolved entries are left alone.
- `--file <path>` and the `CAREER_OPS_PORTALS` environment variable, neither of which was in the issue text.
- A missing portals file is reported as nothing to do, not an error.
- Why the file is edited as text rather than through a YAML round trip: `portals.yml` carries hand-written comments that `yaml.dump()` would discard.
- Exit codes, with the contrast to `check-table-freshness` made explicit. Pending fixes in a dry run do **not** fail the run, so this is a maintenance tool rather than a CI gate. That difference seemed worth stating precisely because the entry I was asked to mirror is a CI gate, and a reader moving between the two could reasonably assume the same behaviour.

## Verification

The claims are checked against the running script, not just its header comment, since a docs PR whose only failure mode is being wrong deserves that:

```
$ node fix-slugs.mjs --file /tmp/does-not-exist.yml
fix-slugs: no portals file at /tmp/does-not-exist.yml — nothing to fix.
EXIT=0
```

Dry-run-writes-nothing, checked by hashing a copy of `templates/portals.example.yml` either side of a real run:

```
$ shasum -a 256 portals-copy.yml    # before
0b293c6472edcd8cb5a332f9d56fac57a6555e632bec90ea20db7659c23053a0

$ node fix-slugs.mjs --file portals-copy.yml
  ... Forto: lever -> ashby, Runway: greenhouse -> ashby ...
Run with --fix to write these changes to portals.yml.
EXIT=0

$ shasum -a 256 portals-copy.yml    # after — identical
0b293c6472edcd8cb5a332f9d56fac57a6555e632bec90ea20db7659c23053a0
```

I read `--dry-run` off `main()` rather than assuming: `dryRun` is derived as `!fix`, so the flag is not separately parsed and the default is what makes a run safe. The entry says `--dry-run` "exists only to say that out loud" rather than implying it is a distinct code path.

`node test-all.mjs --quick` (the CI gate): **2587 passed, 0 failed, 1 warning.** The warning is `cv-sync-check.mjs exited with error (expected without user data)`, which needs the user-layer `cv.md` that a fresh clone does not have. Unrelated to this change, and docs-only besides.

## Note, not part of this PR

`fix-slugs.mjs` has no `npm run` alias while its read-side twin has `npm run verify:portals`. Adding a `fix:slugs` script would make the pair symmetric, but that is a `package.json` change rather than documentation, so I left it out. Happy to open a separate issue or PR if you want it.

## Summary by CodeRabbit

* **Documentation**
  * Added `fix-slugs.mjs` to the Scripts Reference.
  * Documented dry-run behavior, including diff output without writing changes.
  * Documented `--fix`/`--apply` options for applying updates.
  * Clarified that only eligible missing entries with suggested alternatives are updated.
  * Added details on preserving comments, missing-file handling, and exit codes.

### DIFF
--- a/docs/SCRIPTS.md
+++ b/docs/SCRIPTS.md
@@ -38,6 +38,7 @@ All scripts live in the project root as `.mjs` modules. Most are exposed via
 | `npm run reconcile` | `reconcile-pipeline.mjs` | Remove batch-evaluated offers from pipeline.md "Pendientes" |
 | `npm run cover-letter` | `generate-cover-letter.mjs` | Render a cover-letter JSON payload to PDF |
 | `npm run verify:portals` | `verify-portals.mjs` | Probe ATS endpoints to confirm portals.yml slugs resolve (network) |
+| `node fix-slugs.mjs` | `fix-slugs.mjs` | Write `verify-portals.mjs`'s suggested ATS slug fixes back to portals.yml (dry run by default, `--fix` to write) |
 | `npm run reposts` | `detect-reposts.mjs` | Flag re-listed (ghost) postings from scan history |
 | `npm run gemini:eval` | `gemini-eval.mjs` | Evaluate a JD with Google Gemini (free-tier alternative) |
 | `npm run ollama:eval` | `ollama-eval.mjs` | Evaluate a JD with a local Ollama model |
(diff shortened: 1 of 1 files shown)


santifer:
Welcome to career-ops, @MannXo! This is exactly what a script doc should be — you clearly read fix-slugs.mjs before writing: the dry-run-by-default emphasis, WHY it edits portals.yml as text instead of a YAML round-trip (hand-written comments survive), and the exit-code contrast with check-table-freshness. Merged — your first contribution, and after the claim mix-up earlier this week, I'm glad this one landed cleanly for you. Glad to have you here.

PR #2332, opened 2026-07-30
fix(tests): budget the lock test's harness waits for a slow runner, and report what a crashed script printed

### LINKED ISSUE #2331: tracker-writer-lock-tests.mjs flakes on windows-latest, and a crashed script reports no output to diagnose it
### Code of Conduct

### Existing issues

### What happened?

`tracker-writer-lock-tests.mjs` intermittently fails on `windows-latest` only. It failed on #2323 while every other check on the same commit passed, and the same file passes on `ubuntu-latest`, `macos-latest`, and locally.

Two separate problems, and the second is why the first is hard to act on.

**1. The test's harness waits are budgeted for a fast machine.**

The file mixes two kinds of timeout, and only one of them is part of the behaviour under test:

- **Semantic** — `launchWriter(N)` sets `CAREER_OPS_TRACKER_LOCK_TIMEOUT_MS`, and `timeoutMs` / `staleMs` / `retryMs` are the lock's own parameters. These encode what the test is asserting and must not move.
- **Harness** — how long the test waits for a spawned Node process to start, print, or exit. These encode nothing except machine speed.

The harness waits are 2 to 5 seconds:

```js
const probeResult = await waitForWriter(probe, 2_000);
...
const deadline = Date.now() + 2_000;
while (!run.output().stdout.includes(beforeMutationOutput) && Date.now() < deadline) {
  await sleep(10);
}
if (!run.output().stdout.includes(beforeMutationOutput)) {
  fail(`${name}: did not reach the pre-lock review prompt before the fixture mutation`);
}
```

That budgets two seconds for a Windows runner to spawn Node, load the module graph, and flush a prompt to stdout. Under CI load that is not reliably enough. For comparison, every other suite in the repo that spawns a Node child budgets `timeout: 30000` for the same work (`followup-seed-tests.mjs`, `set-status-tests.mjs`, and `run()` in `tests/helpers.mjs`). This file is the outlier.

**2. A crashed script reports nothing, so the failure cannot be diagnosed from CI.**

`test-all.mjs` reports only the name:

```js
} else {
  fail(`${name} crashed`);
}
```

because `run()` discards the diagnostics that `execFileSync` attaches to the thrown error:

```js
} catch (e) {
  return null;
}
```

`e.stdout`, `e.stderr`, `e.status`, and `e.signal` all exist at that point and are thrown away. The CI log therefore shows one line, `❌ tracker-writer-lock-tests.mjs crashed`, with no stack, no assertion text, and no exit code. Diagnosing this meant reading the source and reasoning about which of six timeouts was most likely to have fired, rather than reading the failure.

This affects every script in the `scripts` list, not just this one.

### Steps to reproduce

Intermittent, load-dependent. Observed on `windows-latest` in CI on #2323 (all 10 other checks green on the same commit). Not reproducible locally on macOS, where the file passes 21/21.

### Expected behavior

- Harness waits are generous enough that machine speed cannot fail a correctness test, while every semantic timeout stays exactly as it is.
- A crashed script prints its exit code, stdout, and stderr, so a CI-only failure is diagnosable from the log.

### Impact

The flake is on `main` now, so it can redden any contributor's PR at random, and when it does the log gives them nothing to act on.

### CLI tool

Claude Code

### OS and Node.js version

Windows (CI, `windows-latest`) / Node 22

---

I have a fix in progress and will open a PR against this shortly.

rubicon:
## What does this PR do?

Stops `tracker-writer-lock-tests.mjs` failing on `windows-latest` for want of a faster machine, and makes a crashed script report what it printed so the next platform-specific failure can be read rather than deduced.

## Related issue

Fixes #2331

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

---

### The flake

The file mixes two kinds of timeout, and only one is part of what it asserts:

- **Semantic.** The argument to `launchWriter()` becomes the child's `CAREER_OPS_TRACKER_LOCK_TIMEOUT_MS`, and `timeoutMs` / `staleMs` / `retryMs` are the lock's own parameters. `launchWriter(200)` for the probe is what makes the contention assertion meaningful. **None of these change here.**
- **Harness.** How long the test waits for a spawned Node process to start, print, or exit. These encode nothing but machine speed.

The harness waits were 2 to 5 seconds. The tightest one gives a Windows runner two seconds to spawn Node, load the module graph, and flush a prompt to stdout:

```js
const deadline = Date.now() + 2_000;
while (!run.output().stdout.includes(beforeMutationOutput) && Date.now() < deadline) {
```

For comparison, every other suite that spawns a child budgets `timeout: 30000` for the same work, including `run()` in `tests/helpers.mjs` itself. This file was the outlier. The four harness waits now share one documented `HARNESS_WAIT_MS = 30_000`.

Raising a timeout is usually the wrong fix, so to be explicit about why it is right here: these waits are not measuring anything. Nothing passes because a wait was short and nothing fails because it was long, since each one ends in a `SIGKILL` or a `fail()` that only fires when the machine ran out of time. The assertions are unchanged.

### Why it was hard to diagnose, which is the more useful half

The CI log carried exactly one line:

```
❌ tracker-writer-lock-tests.mjs crashed
```

No stack, no assertion text, no exit code. That is because `run()` discards the diagnostics `execFileSync` attaches to the error it throws:

```js
} catch (e) {
  return null;
}
```

`e.stdout`, `e.stderr`, `e.status`, and `e.signal` all exist there. Working out which of six timeouts had fired meant reading the source and reasoning about probability, which is not a thing a contributor should have to do to act on a red check.

`run()` now records them, `lastRunFailure()` exposes them, and `formatRunFailure()` renders them for interpolation. It returns an empty string when nothing has failed, so the success path is byte-identical and callers can append it unconditionally. A successful run clears the record, so a stale diagnostic can never be attributed to a later command. This applies to every script in test-all's list, not just this one.

### Tests

New `tests/run-failure-diagnostics.test.mjs`, 8 cases: the `null` return is preserved, status / stdout / stderr survive, a success clears the record, and `formatRunFailure()` is empty on success and carries status and stderr on failure.

Verified non-vacuous by restoring the original discard-everything `catch` and re-running: **5 of the 8 fail**, and the 3 that pass are the ones that were already true under the old code (null return, clear-on-success, empty-on-success).

`node test-all.mjs`: **2670 passed, 0 failed**, the one warning being the pre-existing `cv-sync-check.mjs` without user data.

### One thing noticed but not changed

`runDiscovered()`'s guard rejects a discovered suite whose source matches `/\bprocess\.exit\s*\(/`, and that pattern does not distinguish a call from the same characters inside a string literal. A test that needs a child process to exit non-zero trips it. The new suite works around it by setting `process.exitCode` in the child instead, which is arguably better style anyway. Flagging rather than widening the guard, since a stricter-than-necessary guard here is the safe direction and changing it is a separate decision.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved script-execution crash reporting with more actionable diagnostics, including exit status/signals and captured stdout/stderr.
  * Clear old failure diagnostics after a successful run, and avoid stale details when a run is rejected.
* **Tests**
  * Added coverage for failure diagnostics capture and formatting behavior.
  * Reduced test flakiness by standardizing lock-related wait timings with a shared harness constant.

### DIFF
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -30,7 +30,7 @@ import { join, dirname, basename, delimiter } from 'path';
 import { tmpdir } from 'os';
 import { fileURLToPath, pathToFileURL } from 'url';
 import yaml from 'js-yaml';
-import { pass, fail, warn, run, fileExists, finish, ROOT, QUICK, NODE, getBash, toBashPath } from './tests/helpers.mjs';
+import { pass, fail, warn, run, formatRunFailure, fileExists, finish, ROOT, QUICK, NODE, getBash, toBashPath } from './tests/helpers.mjs';
 
 /**
  * Read a repo-relative text file as UTF-8.
@@ -265,7 +265,10 @@ try {
     } else if (allowFail) {
       warn(`${name} exited with error (expected without user data)`);
     } else {
-      fail(`${name} crashed`);
+      // Include the child's exit status and streams. Without them a CI-only
+      // failure arrives as a bare `<name> crashed`: no stack, no assertion
+      // text, no exit code, and nothing a reader can act on.
+      fail(`${name} crashed${formatRunFailure()}`);
     }
   }
 } finally {
--- a/tests/helpers.mjs
+++ b/tests/helpers.mjs
@@ -113,14 +113,19 @@ function resolveAllowedExecutable(cmd) {
  * @returns {string|null} Trimmed stdout, or null when the command fails.
  */
 export function run(cmd, args = [], opts = {}) {
-  const exe = resolveAllowedExecutable(cmd);
-  // Cleared up front rather than on the success path, so the execFileSync call
-  // below stays byte-identical to what it has always been. Touching that line
-  // makes CodeQL re-attribute its long-standing "uncontrolled command line"
-  // finding to whichever PR edited it, and this change does not alter what
-  // reaches the child: the executable is still allowlisted by
-  // resolveAllowedExecutable() and the arguments are still an argv vector.
+  // Cleared as the very first statement. resolveAllowedExecutable() throws for a
+  // command outside the allowlist, so a reset placed after it is skipped on that
+  // path and the previous run's diagnostics survive, which would let a later
+  // formatRunFailure() attribute an unrelated child's stderr to whatever failed
+  // most recently. A stale diagnostic is worse than none.
+  //
+  // Clearing here rather than on the success path also keeps the execFileSync
+  // call below byte-identical: editing that line makes CodeQL re-attribute its
+  // long-standing "uncontrolled command line" finding to whichever PR touched
+  // it. Nothing about what reaches the child changes either way, since the
+  // executable is still allowlisted and the arguments are still an argv vector.
   lastFailure = null;
+  const exe = resolveAllowedExecutable(cmd);
   try {
     return execFileSync(exe, args, { cwd: ROOT, encoding: 'utf-8', timeout: 30000, ...opts }).trim();
   } catch (e) {
--- a/tests/helpers.mjs
+++ b/tests/helpers.mjs
@@ -114,10 +114,15 @@ function resolveAllowedExecutable(cmd) {
  */
 export function run(cmd, args = [], opts = {}) {
   const exe = resolveAllowedExecutable(cmd);
+  // Cleared up front rather than on the success path, so the execFileSync call
+  // below stays byte-identical to what it has always been. Touching that line
+  // makes CodeQL re-attribute its long-standing "uncontrolled command line"
+  // finding to whichever PR edited it, and this change does not alter what
+  // reaches the child: the executable is still allowlisted by
+  // resolveAllowedExecutable() and the arguments are still an argv vector.
+  lastFailure = null;
   try {
-    const out = execFileSync(exe, args, { cwd: ROOT, encoding: 'utf-8', timeout: 30000, ...opts }).trim();
-    lastFailure = null;
-    return out;
+    return execFileSync(exe, args, { cwd: ROOT, encoding: 'utf-8', timeout: 30000, ...opts }).trim();
   } catch (e) {
     // execFileSync attaches the child's streams and exit status to the error.
     // Keep them: callers report failure as `<name> crashed`, and without this a
--- a/tests/helpers.mjs
+++ b/tests/helpers.mjs
@@ -115,10 +115,60 @@ function resolveAllowedExecutable(cmd) {
 export function run(cmd, args = [], opts = {}) {
   const exe = resolveAllowedExecutable(cmd);
   try {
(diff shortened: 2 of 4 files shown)


rubicon:
CodeQL flagged two alerts on `tests/helpers.mjs`, one critical, both on the `execFileSync` call inside `run()`. They are pre-existing rather than introduced here, and `95a69b0` restructures the change so the PR stops re-raising them.

The first version assigned the call's result to a variable in order to clear the failure record on the success path:

```js
const out = execFileSync(exe, args, { ... }).trim();
lastFailure = null;
return out;
```

Same executable, same argv, same options. But CodeQL attributes a finding to whichever pull request edited its line, so editing that line presented a long-standing alert as new work of this branch.

`95a69b0` clears the record before the call instead, leaving the `execFileSync` line byte-identical to `main`. Verified: the only occurrences of `execFileSync` in this branch's diff against `main` are two comment lines.

On the underlying finding, for the record rather than as a change request: `run()` resolves its executable through `resolveAllowedExecutable()`, which returns only `node`, `bash`, `git`, `go`, `wsl`, or a known Windows bash path and throws on anything else, and it passes arguments as an argv vector so no shell is involved. Whether the alert is worth suppressing on `main` is your call, and I have not touched it either way.

Clearing up front is arguably the plainer statement of the rule anyway: the record describes the most recent run and nothing earlier.

Re-verified after the change: 8/8 in `tests/run-failure-diagnostics.test.mjs`, and `node test-all.mjs` at 2670 passed / 0 failed with the one pre-existing `cv-sync-check.mjs` warning.

rubicon:
Valid, fixed in `6aaa15a`. The reset now runs as the very first statement.

You are right that `resolveAllowedExecutable()` throws before it, and the leak is demonstrable: with a prior failed run on record, calling `run()` with a non-allowlisted command left `lastRunFailure()` returning `{"status":3,...,"stderr":"MARKER-ERR\n"}` from the earlier, unrelated child. A later `formatRunFailure()` would have attributed that stderr to whatever failed most recently.

My own test for this was too narrow. It asserted that a *successful* run clears the record and stopped there, which left the throwing path uncovered. Two cases added: that a non-allowlisted executable still throws, and that it clears the record.

The reset stays ahead of the allowlist check rather than moving to the success path, so the `execFileSync` line remains byte-identical and CodeQL does not re-attribute its finding on that line to this branch.

10/10 in `tests/run-failure-diagnostics.test.mjs`, `node test-all.mjs` at 2672 passed / 0 failed.

santifer:
Another solid one, @rubicon. The harness-vs-semantic timeout split is exactly right (the lock's own parameters stay untouched, so the tests still test what they tested), and preserving the crashed child's streams kills the 'X crashed, good luck' class of CI failure for every suite at once. Resetting the diagnostic BEFORE the allowlist throw so a stale record can't be misattributed — and testing that path — is the kind of care that makes this suite trustworthy. Merged! 🚀

PR #2356, opened 2026-07-31
docs(scripts): document the set-status --row/--report selectors

### LINKED ISSUE #2355: docs: document the set-status --row/--report selectors in SCRIPTS.md

### What to do

### Acceptance

### Pattern to mirror

MannXo:
Closes #2355.

Adds a `## set-status` section to `docs/SCRIPTS.md` documenting the `--row` and `--report` selectors from #2347, and names them in the agent-invoked utilities table so they are discoverable from there too.

## Where it went

`set-status.mjs` had no prose section — only a row in the `## Agent-invoked utilities` table. Same situation as `verify-portals.mjs` when I documented `fix-slugs` in #2329. I put the new section immediately after that table, before `## stats.mjs`, since `stats.mjs` is also listed in the table and already has its own top-level section there. The table row now links to it.

## What the entry states

The two selectors, one example each, and the fact that on a diverged tracker the same number through the two flags writes to two different companies. Then the parts a reader would otherwise have to open the source for:

- They are mutually exclusive, and each **replaces** the positional selector rather than narrowing it, so a selector flag plus a positional company is rejected too.
- `--row`/`--report` skip the report-link mismatch guard without `--force`, and why that is not the same thing: they answer the ambiguity, `--force` suppresses the check while the ambiguity is still real. The script's header makes this point well and it seemed worth carrying into the docs, since it is the reason the flags exist rather than an implementation detail.
- When a bare number is still enough: counters that have not diverged, or a row whose report link agrees with its `#`.
- `--report` does not fall back to matching a report-less row by its tracker number.
- A pointer to `find`, which exists for this exact ambiguity.

## One correction to the issue text

The issue asks for "1 = invalid or conflicting selector, 2 = not found". Both are right, but the case the flags exist *for* — a bare ambiguous number — exits **3**, not 1. So the entry documents the full set rather than the two, since a reader hitting the guard would otherwise look up the wrong code.

## Verification

Semantics checked against the running script on a sandboxed tracker (`CAREER_OPS_TRACKER`) built to be diverged: row #5 Globex links report #4, row #6 Initech has no report, row #7 Acme links report #5.

```
bare 5 (ambiguous: row 5 vs report 5)          exit=3 written=no
--row 5 (tracker row -> Globex)                exit=0 written=YES
--report 5 (row linking report 5 -> Acme)      exit=0 written=YES
--row 7 (mismatched link, no --force)          exit=0 written=YES
--row 5 --report 5 (conflicting)               exit=1 written=no
--report 999 (nothing links it)                exit=2 written=no
--report 6 (row 6 has no report link)          exit=2 written=no
--row abc (non-numeric)                        exit=1 written=no
```

Every claim in the entry corresponds to a line above. The `--row 5` / `--report 5` pair landing on Globex and Acme respectively is the example used in the docs.

`node test-all.mjs --quick`: **2713 passed, 0 failed, 1 warning.** Identical to the same run on a clean checkout, and the warning is the pre-existing `cv-sync-check.mjs exited with error (expected without user data)`, which needs the user-layer `cv.md` a fresh clone does not have.

## Note, not part of this PR

`AGENTS.md` still tells agents to use `node set-status.mjs <report#|company> <State>` in the Pipeline Integrity rules and the main-files table. On a tracker whose counters have diverged, that bare-number form is the one the guard rejects with exit 3, so agents following the rule literally will hit it. Mentioning `--row`/`--report` there would close that gap, but AGENTS.md is outside what this issue asked for and is a behaviour-guidance file rather than documentation. Happy to open a separate issue or PR if you want it.

## Summary by CodeRabbit

* **Documentation**
  * Expanded status utility guidance for permanent tracker and report-number differences.
  * Clarified selector disambiguation and row lookup using `find`.
  * Documented automated-use recommendations, stable exit codes `0`–`4`, lock retry behavior, and non-zero write results.
  * Added details on validation, mismatch protection, ambiguity handling, and usage examples.

### DIFF
--- a/docs/SCRIPTS.md
+++ b/docs/SCRIPTS.md
@@ -896,7 +896,7 @@ A bare number or company name is convenient, but becomes ambiguous when multiple
 
 `--row` and `--report` are mutually exclusive. Because an explicit selector answers the report-mismatch guard rather than overriding it, `--row` bypasses that guard without needing `--force` (which silences the check while the ambiguity is still real).
 
-This is worth preferring in practice, not just in principle: on a diverged tracker the guard fires on nearly every bare numeric call, and a guard that always fires trains callers to reach for `--force` by reflex — which disables it everywhere, including the cases it was written to catch. Reach for a selector (or the company name) instead.
+This is worth preferring in practice, not just in principle. Once the counters have diverged, a bare number trips the guard whenever the row it matches links a report number other than its own `#`, or links no report at all while a different row claims that number as its report — so on a tracker with a wide gap the check keeps firing, and a check that keeps firing teaches callers to pass `--force` by reflex, which disables it everywhere including the cases it was written to catch. Reach for a selector (or the company name) instead.
 
 ### Bare numbers vs. explicit selectors
 
--- a/docs/SCRIPTS.md
+++ b/docs/SCRIPTS.md
@@ -886,7 +886,7 @@ node set-status.mjs --row 12 Applied
 node set-status.mjs --report 345 Applied --on 2026-08-01
 ```
 
-A bare number or company name is convenient, but becomes ambiguous when multiple tracker rows exist for a company or when tracker row IDs and report IDs diverge. Base selectors resolve the main target, while explicit selectors and filters disambiguate the target row:
+A bare number or company name is convenient, but becomes ambiguous when multiple tracker rows exist for a company or when tracker row IDs and report IDs diverge. That divergence is permanent once it starts: `reserve-report-num.mjs` treats tracker row IDs as occupied when it allocates a report number, so a row that never got a report still consumes a number the report sequence then skips — the two counters leapfrog each other and never realign. On a diverged tracker "5" may mean tracker row #5 or report #5, which are different applications. Base selectors resolve the main target, while explicit selectors and filters disambiguate the target row:
 
 - `--row N`: Selects the row whose `#` cell is `N`.
 - `--report N`: Selects the row whose `Report` cell links report `N`.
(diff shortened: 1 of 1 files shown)


santifer:
@MannXo I owe you a straight explanation: you were first here. Your PR went up on Jul 31, and because it didn't reference #2355 my race-check missed it when #2359 (same selectors, claimed later) got merged two days after: that one's on my process, not on you. Rather than a hollow apology, here's the fast path: main still lacks four things your PR has, and they'd merge quickly as a slimmed rebase: (1) the WHY of the counter divergence (reserve-report-num treating row IDs as occupied, so the sequences leapfrog and never realign), (2) the complete exit-code table with 0 and 4 (both verified against CLI_EXIT in tracker-utils.mjs:531), (3) the point that a guard firing on every bare call trains callers to reach for --force, and (4) the find cross-ref. If you'd rather not redo the rebase after this collision, say the word and I'll resolve the mechanical part maintainer-side with your authorship intact.

MannXo:
Thank you @santifer for laying that out straight, no hard feelings from my side. Rebased down to those four points in 7461aad (+11 -3), verified against `CLI_EXIT` in `tracker-utils.mjs` and `occupiedFromTracker` in `reserve-report-num.mjs`, so no need for the maintainer-side fix. Ready when you are.

santifer:
Another solid one, @MannXo. You were first on #2355 and still took the time to rebase around what landed meanwhile: the leapfrog explanation and the CLI_EXIT table (0 idempotent, 4 retryable lock) are exactly the detail SCRIPTS.md was missing, and the --force-by-reflex warning is the best line in the section. Merged, and thanks for the patience. 🚀

Issue #2337, opened 2026-07-30
update-system.mjs apply: no visibility when a system-layer file has a local fix that would be silently overwritten


clede:
> ⚠️ Privacy check: no CV/scan/pipeline/report content included below.

## Code of Conduct
- [x] I agree to follow the Code of Conduct

## Existing issues
- [x] I searched existing issues and this hasnt been requested yet — related precedent: #1035 (translated `batch-prompt.md` reverted every update cycle), same underlying pattern, different file class, not a full duplicate.

## Problem

`apply()` fetches upstream `main` and does a raw checkout of every `SYSTEM_PATHS` file from `FETCH_HEAD` — not a merge — so a local commit that fixed a bug in a system-layer file, if upstream hasnt independently picked up the same fix, gets silently discarded with no diff, warning, or list of affected paths, before or after.

Observed twice on the same file in this install: a LinkedIn-contact-line fix to `generate-cover-letter.mjs` (see #2336) was reverted by v1.22.0→v1.23.0, reapplied, then reverted again by v1.23.0→v1.24.0. A separate nearby fix in the same file happened to be independently adopted upstream in the interim and survived that second update — isolating that this is specifically about locally-fixed-but-not-yet-upstreamed changes, not all local edits in general.

## Steps to reproduce

1. Commit a local-only change to any file under `SYSTEM_PATHS` that upstream doesnt have (e.g. a bugfix not yet merged into `santifer/career-ops` `main`).
2. Have a newer version available (`node update-system.mjs check` reports `update-available`).
3. Run `node update-system.mjs apply`.
4. Diff the file before/after — the local fix is gone, with no mention of it anywhere in `apply()`'s output.

## Expected result

At minimum, `apply()` should detect when a `SYSTEM_PATHS` file has diverged from its last-synced upstream baseline and warn/list those paths before checkout — similar in spirit to the existing user-layer `SAFETY VIOLATION` check — so the user can consciously stash/cherry-pick first.

## Proposed solution

Not prescribing a specific mechanism — flagging for a maintainer design call. Options that would each solve it: warn/list diverged files pre-checkout, a `--force` gate, auto-stashing divergent system files, or something else entirely.

## Area
Other

## CLI tool
Claude Code

## OS and Node.js version
macOS 26.6 / Node v26.0.0

## Output of npm run doctor
```json
[...]

Issue #2007, opened 2026-07-17
Cross-version upgrade regression tests: PR-gate leg + release-cadence matrix


Scott-Emberson:
## Why

`update-system.mjs` has the highest blast radius in the repo and zero dynamic coverage — `updater-migration-tests.mjs` is static-only. Two shipped bugs prove the class: #1998 (pre-1.11.0 clients silently skip every newer path; install breaks while reporting success) and #2002 (stale manifest entry printing a pathspec error in every user's upgrade — found within 30 minutes of actually *running* upgrades in a spike; fix out as #2004).

## Spike evidence (2026-07-17)

Cloned each of v1.16.0–v1.20.0 into temp installs, seeded user data, redirected the canonical-repo fetch to a local bare mirror of `main` via `GIT_CONFIG_GLOBAL` + `url.insteadOf`, ran each tag's own `apply`. All 5 legs: exit 0, VERSION advanced, user files byte-intact, and the post-apply updater blob equals `main`'s — confirming the self-reexec hands migration to the *target* code in every tag. `apply` is pure git (curl lives only in `check()`), so the whole flow runs hermetically. `apply` has no version gate — an equal-VERSION control leg still applied content, so content blobs, not VERSION, are the honest oracle.

## Design (two decoupled layers)

**Layer 1 — PR gate (fast, hermetic, every PR, ubuntu):** materialize the newest old release tag + an era-appropriate user-data fixture, force a `main` ref onto the commit under test in a local mirror, run the old install's `apply` (which re-execs into the PR's updater), then assert in order: upgrade demonstrably executed (blob oracle), user layer byte-identical to a SHA-256 manifest (checked *before* running doctor — doctor's autoCopy writes user files), every path added to SYSTEM_PATHS since the tag exists on disk (the #1998 class), pinned data-consumption checks (harness-owned expectations, not the PR's own scripts), clean exit + auto-update commit. Plus a canary mode that plants a deliberate user-file clobber and asserts the harness goes red — a gate that has never failed proves nothing.

**Layer 2 — release qualification (nightly + tag push + dispatch, 3-OS matrix):** same harness looping the last 5 tags end-to-end with each tag's own shim, plus a rollback leg (oldest tag) and the maximal skip-version jump (oldest → HEAD — real users upgrade non-linearly). Deliberately *not* a PR gate: a frozen old shim's own quirk would fail every unrelated PR forever.

[...]

Issue #2002, opened 2026-07-17
SYSTEM_PATHS still lists plugins-registry.json (deleted in #1402) — every upgrade emits a pathspec error


Scott-Emberson:
## Summary

`SYSTEM_PATHS` in `update-system.mjs` still lists `plugins-registry.json`, but that file was deleted in #1402 (v1.14 era) when the registry was split into per-plugin files under `plugins-registry/`. Both entries are present today (`update-system.mjs:280-281`):

```js
  'plugins-registry/',
  'plugins-registry.json',
```

Because the checkout target no longer exists at `main`, **every upgrade any user runs today emits**:

```
error: pathspec 'plugins-registry.json' did not match any file(s) known to git
```

The updater tolerates missing paths by design ("File may not exist in remote — skip"), so the upgrade completes fine — but the error line lands in every user's `apply` output, and the manifest carries a permanently dead entry.

## Repro

Observed while spiking cross-version upgrade tests: cloned each of the last five release tags (v1.16.0–v1.20.0) into temp installs, redirected the canonical-repo fetch to a local bare mirror of current `main` (`url.insteadOf`), and ran each tag's own `node update-system.mjs apply`. All five legs printed the pathspec error above; all five otherwise completed correctly (VERSION advanced, user files untouched, reexec ran the target updater).

## Suggested fix

1. Remove the `'plugins-registry.json'` entry from `SYSTEM_PATHS`.
2. Optionally, add a guard to `updater-migration-tests.mjs`: every non-directory `SYSTEM_PATHS` entry must exist in the working tree — that turns any future delete-without-manifest-cleanup into a CI failure instead of a permanent stderr line in every user's upgrade. (Directory entries and deliberately-optional paths would need a small allowlist.)

Happy to PR both — the removal is one line; the guard is ~15 lines in the existing test file.

Related context: #1998 covers the pre-1.11.0 legacy-client path-skipping problem — different mechanism (old manifests missing *new* paths), but the same lesson: manifest drift is invisible until someone actually runs an upgrade. I'm separately drafting a design for automated cross-version upgrade tests that would catch both classes.

Issue #1998, opened 2026-07-17
`apply` on pre-1.11.0 clients silently skips every path added in later releases — reports success, leaves the install broken (ERR_MODULE_NOT_FOUND)


Dreamkeeper:
## Summary

Upgrading a client that is still on **v1.10.0 or older** to a current release leaves the install broken: `node update-system.mjs apply` prints `Update complete` and commits, but none of the files introduced after the client's version are checked out (in my case ~380 files: `plugins/`, `evals/`, `web/`, `browser-extract.mjs`, `lib/`, many new `modes/` dirs, …). The first script you run afterwards crashes:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '.../plugins/_engine.mjs' imported from .../doctor.mjs
```

The target-manifest mechanism from #983 (v1.11.0, later hardened by #1245/#1267/#1706) fixes this for clients ≥1.11.0, but every client still on ≤1.10.0 performs its next upgrade with the *old* apply logic, so they all hit this exact failure once — and `apply` gives them no signal that anything is wrong.

## Repro

1. Client at v1.10.0 (updater without #983's target-manifest / self-reexec logic).
2. `node update-system.mjs apply` to v1.20.0.
3. Output (abridged):

```
Updating system files...
error: pathspec '.gemini/commands/' did not match any file(s) known to git

Update complete: v1.10.0 → v1.20.0
Updated 104 system paths.
```

4. `node doctor.mjs --json` → `ERR_MODULE_NOT_FOUND` (doctor.mjs was updated to the v1.20.0 version — it *is* in the old manifest — but its new imports `./plugins/_engine.mjs` and `./browser-extract.mjs` were never checked out, because `plugins/` and `browser-extract.mjs` are not in the v1.10.0 manifest).

## Root cause

v1.10.0's `apply()` checks out **only the paths in its own local `SYSTEM_PATHS`** (~105 entries) plus a hardcoded `BOOTSTRAP_PATHS` list. It has no knowledge of paths added upstream after v1.10.0, so it never even attempts them. New files survive only if they happen to live under a directory pathspec the old manifest already lists (e.g. `.github/`, `docs/`); everything else is silently absent. `Updated 104 system paths` = the old manifest's 105 entries minus `.gemini/commands/`, which was removed upstream.

Two things make the failure hard to diagnose:

[...]

PR #2358, opened 2026-07-31
feat(test): dynamic upgrade PR gate — old install applies the commit under test (#2007)

Scott-Emberson:
Part of #2007. The **PR-gate leg** santifer greenlit as "the actionable half" (release-cadence matrix follows separately as Phase E).

## What

Adds `upgrade-tests.mjs` + a `upgrade-gate` CI job. Every PR, one hermetic upgrade leg runs: **newest release tag that's an ancestor of HEAD → this commit**, seeding an era-appropriate user install (the fixtures from #2032) and driving a real `update-system.mjs apply`, then asserting the user layer survived byte-for-byte and the new system code actually landed.

## Why

`update-system.mjs` is our highest-blast-radius script. The 07-17 spike found #2002 (stale manifest path) and #2337 documents another silent-overwrite gap — both are exactly "an upgrade quietly damages a real install," which no existing test exercises. This gate turns that class red **before** merge.

## How it's honest (not a vacuous green)

The updater self-reexecs — the **target commit's** updater performs the migration — so the test object is "this PR's updater × an old install," which is what users actually hit. Assertions run in order and the cheap structural ones gate the expensive smoke:

1. `apply` exits 0.
2. **Non-vacuity oracle** — `update-system.mjs` blob on disk matches the target commit's blob (proves the upgrade *ran*, not that nothing happened).
3. **User-layer byte manifest** — every seeded user file is SHA-compared, asserted **before** any repo script runs (so `doctor.mjs`'s autoCopy can't mask a clobber).
4. **#1998-class check** — new system paths introduced by the target are present after upgrade.
5. Harness-pinned consumption checks (tracker rows/status counts, salary observations) derived from the fixture, not from the code under test.
6. `doctor.mjs --json` smoke last.

A **canary** leg plants a poison commit that force-tracks `cv.md` into `SYSTEM_PATHS`; the harness MUST detect the clobber. If it can't fail, its green is worthless — so the canary failing to fail fails CI.

## Verification (local, this commit → `career-ops-v1.24.0`)

- `node upgrade-tests.mjs --pr-gate` → **GREEN**, all 23 assertions pass (10 user files byte-identical, new paths present, tracker/salary intact, doctor onboardingNeeded=false).
- `node upgrade-tests.mjs --canary` → **CANARY GREEN**: harness detected the planted `cv.md` clobber.

## Scope

- New: `upgrade-tests.mjs` (harness), `upgrade-gate` job in `test.yml`.
- `update-system.mjs`: `upgrade-tests.mjs` added to `SYSTEM_PATHS` (it ships as system code).
- `DATA_CONTRACT.md`: one row documenting the harness.
- Fixtures + `seed-fixture.mjs` already landed in #2032; this builds on them.

Phase E (nightly + tag-push 3-OS release-cadence matrix, last-5-tags with per-tag shims/rollback) is intentionally **not** in this PR — it's heavier and per #2007 can follow.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added automated upgrade compatibility checks to detect migration issues before release.
  * Improved validation that system files update correctly while preserving user-created files.
  * Added checks for tracker data, salary data, and post-upgrade system health.

* **Tests**
  * Added pull-request and canary upgrade test modes.
  * Added detection for unintended changes to user files during upgrades.
  * Integrated upgrade checks into automated validation.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,15 +42,17 @@ jobs:
         working-directory: dashboard
 
   upgrade-gate:
+    name: Upgrade regression gate
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v7
         with:
-          fetch-depth: 0   # harness needs release tags
-      - uses: actions/setup-node@v6
+          fetch-depth: 0            # harness needs release tags
+          persist-credentials: false # the harness only ever touches a local mirror
+      - uses: actions/setup-node@v7
         with:
           node-version: '24'
-      - uses: actions/setup-go@v6
+      - uses: actions/setup-go@v7
         with:
           go-version: '1.26'   # old updaters rebuild the dashboard during apply
       - run: npm install --ignore-scripts
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,3 +40,21 @@ jobs:
       - name: Go dashboard tests
         run: go test ./...
         working-directory: dashboard
+
+  upgrade-gate:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          fetch-depth: 0   # harness needs release tags
+      - uses: actions/setup-node@v6
+        with:
+          node-version: '24'
+      - uses: actions/setup-go@v6
+        with:
+          go-version: '1.26'   # old updaters rebuild the dashboard during apply
+      - run: npm install --ignore-scripts
+      - name: Upgrade PR gate (newest old release -> this commit)
+        run: node upgrade-tests.mjs --pr-gate
+      - name: Canary (harness must be able to fail)
+        run: node upgrade-tests.mjs --canary
--- a/DATA_CONTRACT.md
+++ b/DATA_CONTRACT.md
@@ -139,6 +139,7 @@ These files contain system logic, scripts, templates, and instructions that impr
 | `DATA_CONTRACT.md` | This file |
 | `writing-samples/README.md` | System-owned onboarding documentation for the writing-samples directory |
 | `seed-fixture.mjs` / `test-fixtures/*` | Upgrade-test fixtures and seeder (system layer; fictional data, never user data) |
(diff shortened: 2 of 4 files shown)


Scott-Emberson:
Addressed all 6 review findings (verified each against current code — all valid). Pushed `a0505c3`.

- **pickOracle scope (major):** now selects only files `apply` manages — an exact `SYSTEM_PATHS` entry or a file under a `dir/` prefix entry. Previously it could pick any changed file; a changed user-layer or meta file is never rewritten by `apply`, so its blob could never match the target → spurious RED. `SYSTEM_PATHS` is now parsed once, ahead of the oracle, and reused for the #1998-class new-path delta.
- **`sysMatch[1]` guard:** labeled throw if the `SYSTEM_PATHS` regex ever misses (was a raw TypeError on a renamed constant).
- **Consumption-check reads:** `applications.md` / `salary-observations.tsv` reads and the header lookup now degrade to labeled `ok()` failures instead of throwing out of `runLeg` — the gate goes cleanly RED with a diagnostic rather than crashing.
- **doctor smoke:** surfaces the failure reason (`onboardingNeeded`/`missing`, or the stderr head) in the assertion message.
- **CI hygiene:** `upgrade-gate` gains a descriptive job name, `persist-credentials: false` on checkout (the harness only ever touches a local mirror), and `setup-node`/`setup-go` `@v6`→`@v7` to match the `test` job.

Re-validated locally: `--pr-gate` GREEN, `--canary` GREEN (planted `cv.md` clobber detected).

santifer:
This is the half I greenlit, delivered exactly as scoped, @Scott-Emberson. I ran the harness rather than reading it: the PR gate really does seed a v1.24.0-era install, apply this commit, and prove the tracker states, salary observations and doctor state all survive — and the canary really does plant a user-file clobber and catch it, which is the difference between a test and a test that knows how to fail. Two details I appreciated: no secrets reachable from the new job, and --ignore-scripts so an upgrade run can't execute third-party install hooks. update-system.mjs is our highest-blast-radius script and it finally has a gate that exercises the thing that actually hurts users. Phase E (the release-cadence matrix) whenever you're ready. Merged! 🚀

PR #2359, opened 2026-07-31
docs(SCRIPTS): document set-status --row/--report selectors

### LINKED ISSUE #2355: docs: document the set-status --row/--report selectors in SCRIPTS.md

alien0101x:
Fixes #2355

## Summary by CodeRabbit

- **Documentation**
  - Added guidance for using the status-setting script.
  - Documented supported row and report selectors, selector disambiguation rules, available options, and exit codes.

### DIFF
--- a/docs/SCRIPTS.md
+++ b/docs/SCRIPTS.md
@@ -813,6 +813,27 @@ These have no `npm run` binding — modes and agents call them with
 
 ---
 
+## set-status.mjs
+
+Canonical tracker write path: strict `states.yml` validation, shared lock, atomic write. Modes and agents call this instead of hand-editing `applications.md`.
+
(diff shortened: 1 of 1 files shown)


santifer:
Welcome to career-ops, @Mainul999-bot! You claimed the issue and had the docs ready a minute later, which tells me you'd already read the code — and it shows: documenting the exit codes (1 invalid selector, 2 no match, 3 mismatch guard) is the part a caller actually needs and the issue didn't even ask for. The line about --row answering the guard rather than overriding it, unlike --force, is exactly the distinction that matters. Merged — your first contribution. Glad to have you here.

PR #1904, opened 2026-07-15
fix(web): show the retry card, not the update-checkout panel, for runtime scan errors

rubicon:
## What does this PR do?

Explore's failure card shows the "Discovery needs the full toolkit" panel (the one that tells you to update career-ops) for ordinary runtime scan errors, not just for a data-only checkout. It picks that panel by regex-matching the error text for words like "scanner", so a transient failure whose message is "The scanner returned no readable output." trips it and wrongly tells me my checkout is data-only or outdated. This keys the panel off the structured HTTP 400 that `/api/explore` already returns for a real data-only checkout, so runtime scan errors now get the normal retry card instead.

## Related issue

Closes #1905

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation / translation
- [ ] Refactor (no behavior change)

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/santifer/career-ops/blob/main/CONTRIBUTING.md)
- [x] If this is a new feature or architecture change, I opened an issue first (bug fixes, providers, docs & translations are exempt - send those straight in)
- [x] My PR does not include personal data (CV, email, real names, scan results, or pipeline data)
- [x] I ran `node test-all.mjs` and all tests pass
- [x] My changes respect the [Data Contract](https://github.com/santifer/career-ops/blob/main/DATA_CONTRACT.md) (no modifications to user-layer files)
- [x] My changes align with the [project roadmap](https://github.com/santifer/career-ops/discussions/156)

---
Questions? [Join the Discord](https://discord.gg/8pRpHETxa4) for faster feedback.

## Summary by CodeRabbit

* **New Features**
  * Added structured “scanner missing” detection across both standard and AI exploration flows.
* **Bug Fixes**
  * Updated the failed-state UI to use the structured signal (no error-message parsing).
  * Scanner-missing state now resets on new discovery actions and persists across reloads.
* **Tests**
  * Added regression tests for scanner-missing status classification.
  * Expanded the web test runner to execute all `test-*.mjs` suites.

PR #2361, opened 2026-07-31
fix(web): make Codex AI search safe and reliable

KraineOpasen:
## What does this PR do?

Makes the Codex provider in Explore → AI Search safer and more reliable. Codex now runs in an isolated temporary directory with a read-only sandbox and no approval prompts, uses live web search, returns only the final response, and cleans up temporary files after execution.

## Related issue

None — this is a bug fix, so an issue is not required.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

## Validation

- `npm run build` — passed
- Published branch verified against the current official `main`
- One commit and one changed file
- `node test-all.mjs` was not run
- No pipeline, scan, AI evaluation, auto-apply, or submission actions were run

---

## Summary by CodeRabbit

* **New Features**
  * Added support for Codex-powered executions with automatic capability detection.
  * Codex runs in an isolated, read-only, non-interactive environment.
  * Responses are presented with cleaner output and no visible command transcripts.
* **Bug Fixes**
  * Added safer validation before execution.
  * Improved handling of missing responses and execution failures.
  * Improved cleanup of temporary files and private diagnostics.


santifer:
Reviewing this now. One finding up front, because it addresses the main caveat raised against this PR — that it switches `stdin` to `"ignore"` on the **generic** spawn path, not just the Codex one, so it also affects Claude runs.

**I checked, and the change is in the safe direction — arguably a latent fix rather than a risk.**

Node's default with no `stdio` option is `["pipe","pipe","pipe"]`, so today every CLI we spawn gets a stdin pipe that **nothing ever writes to and nothing ever closes**. I swept `web/src` for any write to `child.stdin`: there are none. So a CLI that reads stdin — or that probes it to decide interactive vs headless — currently blocks on a pipe that will never produce data or EOF. With `"ignore"` it gets `/dev/null` and reads return EOF immediately.

In other words the previous behaviour only *appeared* safe because none of our current binaries happen to read stdin. `"ignore"` is the correct posture for a headless spawn, and it makes the "hangs forever with no output" class of bug structurally impossible.

Worth stating explicitly in the PR description, since a reader seeing "changes stdin for Claude too" will reasonably assume regression before they assume hardening.

Still to do on my side before I approve: the process-group termination and the isolated-tempdir sandbox deserve a manual run against a real binary rather than a code read — that's the part no amount of diff review substitutes for. I'll come back with that.

santifer (changes requested):
Reviewed properly at last — sorry, this has been on my list as "needs a run against a real Codex binary" for two weeks, and I let that block a review of everything else about it.

## What's right, and it's most of it

**Probing `--help` before using flags** is the correct instinct. Passing `--sandbox read-only --ephemeral --strict-config` to a Codex that doesn't have them turns a search into a confusing CLI usage error, and the user has no way to know why. Checking first, and failing with a specific message, is the honest version.

**`code: "CODEX_UNSUPPORTED"` with the 400** is the pattern I want and I'm glad to see it here — I asked for exactly this on #1904 an hour ago, where the same route family has three different 400s and the client can't tell them apart. Yours can.

**Fail-closed on the probe**: `.catch(() => false)`, and a 5s timeout with SIGTERM, so an unresponsive binary refuses the run rather than hanging it. `appendBounded(...).slice(-64_000)` caps a pathological `--help`. Read-only sandbox in a `mkdtemp` directory with cleanup is the right shape for "run someone's CLI against the open web".

## One change: the capability cache can't see an upgraded Codex

```ts
const codexCapabilityCache = new Map<string, Promise<boolean>>();
```

Keyed on `binPath` alone. So the verdict for a given path is fixed for the lifetime of the server process — and the single most likely thing to happen after a user sees *"Update Codex and try again"* is that **they update Codex**. They'd then keep getting the same refusal until they restart the web app, with no clue that a restart is what's missing.

We have already paid for this exact bug and the cure is in the codebase. `states.ts` after #2590:

```ts
const statesCache = new Map<string, { mtimeMs: number; size: number; states: CanonicalState[] }>();
const { mtimeMs, size } = fs.statSync(file);
const cached = statesCache.get(file);
if (cached && cached.mtimeMs === mtimeMs && cached.size === size) return cached.states;
```

One `statSync` per call, no re-probe unless the binary actually changed. And `text-key.ts` carries the other half of the same lesson — `// only successes are cached (#2590)` — which you've half got already: you delete the entry when unsupported, so the negative case re-probes. Keying on mtime+size makes the positive case correct too, and lets you keep the entry on failure without pinning anything.

That's the whole request: key the cache on the binary's `mtimeMs`+`size`, not just its path.

## Two notes, not blocking

- The probe requires **nine** flag strings by substring. It's fail-closed, so a wording change in Codex's help degrades to "unsupported" rather than to an unsafe run — the right direction. But it's worth a comment saying that's deliberate, or the next person will read a mysterious refusal as a bug and loosen the check.
- Your checklist has `test-all.mjs` and the roadmap boxes unticked. CI covers the first; I'm not asking for the second.

I still can't run this against a real Codex binary, so I'm reviewing the code and the failure directions rather than claiming an end-to-end verification. Make the cache change and I'll merge on that basis, and say so plainly in the merge note.

KraineOpasen:
Addressed the requested capability-cache change: the cache is now keyed by the Codex binary's mtimeMs + size, while negative probes still retry. I also merged current main and resolved the spawnHeadlessCli conflict; the PR is now conflict-free. Local npm test, npm run typecheck, and npm run build all pass. The remaining workflows are awaiting maintainer approval. Ready for re-review.

santifer (approved):
Reviewed the full diff against what I asked for on Aug 11 — this delivers it, and then some. Approving.

**The cache keying is well resolved** — exactly the shape I wanted:
- Keyed by `mtimeMs + size` of the resolved binary, so a Codex upgrade *at the same path* invalidates a previously-successful probe. A pure `existsSync`/path key would have gone stale across an upgrade; this doesn't.
- Negative probes retry rather than sticking — and the guard `codexCapabilityCache.get(binPath) === entry` before the delete is the detail that makes it correct: a transient failure clears its own entry without clobbering a *newer* entry installed after the binary changed. That's the race I'd have worried about, and it's handled.
- Concurrent cold requests share one in-flight `probe` promise. Good.

**The safe-exec hardening around it is the stronger half, and it's careful work:**
- Fail-closed capability gate: the probe requires every safety flag (`--sandbox read-only`, `--ephemeral`, `--ignore-user-config`, `--strict-config`, `--ask-for-approval never`, …) to be present in `--help` before Codex is allowed to run; missing → `CODEX_UNSUPPORTED` 400. A drifted/older Codex is refused, never silently downgraded.
- Prompt-leak prevention is exactly right for this surface: Codex's stdout/stderr echo the full prompt + session metadata, and this drains-but-never-forwards them, reads the answer from `--output-last-message`, and logs only byte/line counts + marker categories — never contents. This matters because the proposer runs on the user's own auth.
- Isolated temp cwd, read-only sandbox, process-group SIGTERM→SIGKILL fallback for Codex's descendants — all sound.

**One follow-up, and I'm taking it myself — not a blocker for you.** The capability-probe flag list is now the security-load-bearing part of this route, and the surface's own pattern (see `explore-error.mjs` + `explore-error.test.mjs`) is to extract the classifiable logic to a testable `.mjs` single-source-of-truth and lock it with a test. I'll pull `supportsSafeCodexExec`'s required-flag set (and the two new codes) into that shape in a follow-up so a future flag rename can't silently weaken the gate. That's my house-keeping on my surface, not work to push back onto you after three weeks.

Thanks for the patience and for answering every round — merging.

santifer:
Thanks, @KraineOpasen — this went well beyond the cache-keying fix I asked for. The fail-closed capability gate and the prompt-leak handling (draining Codex's stdout/stderr without forwarding them, reading the answer from `--output-last-message`) are exactly the right posture for a proposer that runs on the user's own auth. Three weeks open and you answered every round — appreciated. Merged 🚀

PR #2343, opened 2026-07-30
feat: guard CV work experience against non-chronological order

kmichels:
Closes #2342

## What

Adds a reverse-chronological guard for Work Experience entries in rendered CVs, running alongside the existing section-order guard in `generate-pdf.mjs`.

## Why

When tailoring a CV toward a JD, the agent is tempted to promote "the most relevant role" to the top of Work Experience. I hit this in real use — generating a CV for an infrastructure/SRE manager role, a 2008–2014 entry got moved above a 2021–2022 Head of Infrastructure role because it matched the JD better.

It backfires:

- **Amplifies the recency problem it's meant to hide** — the reader's first impression becomes how long ago the relevant work was
- **Buries the most recent senior title** mid-page
- **Reads as concealment** — ATS parsers and recruiters both expect newest-first

Tailoring belongs in the summary, the competencies block, and bullet selection within each role. Never in the ordering.

`generate-pdf.mjs` already guards *section* order against `cv.md` (#1646). Experience ordering had no equivalent, so this reached the PDF silently. Prompt-level instruction is probabilistic; this is a deterministic property of the rendered document, so it belongs at the same enforcement point.

## How

- **`cv-experience-order.mjs`** (new) — `parseExperienceStart()` and `validateCvExperienceOrder()`. Split out rather than inlined because `generate-pdf.mjs` is already near the practical file-size ceiling; follows the existing `*-core.mjs` convention (`liveness-core.mjs`, `cv-sections-core.mjs`).
- **`generate-pdf.mjs`** — invokes the guard next to `validateCvSectionOrder`, adds `--allow-nonchronological` mirroring `--allow-reorder`, updates usage and header docs.
- Only the **start** date is compared — reverse-chronological ordering is defined by when each role began.
- Unparseable or absent dates **no-op**, following the same "don't penalize missing data" discipline as `location_filter` and the country-eligibility filter. A CV with no dates renders exactly as before.

## Tests

`tests/generate-pdf-experience-order.test.mjs` — 12 assertions covering inversions, valid orderings, year-only periods, concurrent roles sharing a start month, the escape hatch, and the missing-data no-ops.

Also updated `tests/generate-pdf-page-budget.test.mjs` to copy the new sibling module into its sandbox — same reason the existing comment gives for `theme-style.mjs`, otherwise the isolated script fails with `ERR_MODULE_NOT_FOUND`.

```
node test-all.mjs
📊 Results: 1211 passed, 0 failed, 0 warnings
```

## Verified end to end

```
[...]

PR #2365, opened 2026-07-31
fix: git-ignore article-digest.md and guard the user-layer list

### LINKED ISSUE #2364: article-digest.md is user-layer but missing from .gitignore (personal data can be committed)
## Summary

`article-digest.md` is documented as a **user-layer** file but is the only one not covered by `.gitignore`. It sits untracked in the working tree, where a `git add .` will stage it.

## Evidence

`AGENTS.md` line 13 lists it as user layer:

> **User Layer (NEVER auto-updated; personalization goes HERE):** `cv.md`, `config/profile.yml`, `modes/_profile.md`, `modes/_custom.md`, **`article-digest.md`**, `portals.yml`, `data/*`, `reports/*`, `output/*`, `interview-prep/*`

`DATA_CONTRACT.md` line 19 describes it as "Your proof points from portfolio".

But `.gitignore` covers every sibling and not this one:

| User-layer file | In `.gitignore`? |
|---|---|
| `cv.md` | yes (line 2) |
| `config/profile.yml` | yes (line 56) |
| `portals.yml` | yes (line 59) |
| `modes/_profile.md` | yes (line 60) |
| `modes/_custom.md` | yes (line 61) |
| `data/*` | yes (line 8) |
| `reports/*.md` | yes (line 19) |
| `output/*` | yes (line 21) |
| `interview-prep/*` | yes (line 41) |
| **`article-digest.md`** | **no** |

Reproduce on a fresh clone:

```bash
touch article-digest.md
git status --porcelain
# ?? article-digest.md
git check-ignore -v article-digest.md
# (no output — not ignored)
```

## Why it matters

`article-digest.md` is where the modes are told to put the candidate's detailed proof points: metrics, project context, business details, employer specifics. It is arguably the most personal file in the layout after `cv.md`.

The exposure is not hypothetical for contributors. Anyone who forks career-ops to send a PR is working in a repo where their own `article-digest.md` is stageable, and `git add .` is a common reflex. The failure is silent and the data is personal.

I hit this while preparing a PR against this repo. My own `article-digest.md` was the sole untracked file in the tree.

## Fix

One line in `.gitignore`, next to the other user-layer entries:

```gitignore
article-digest.md
```

Possibly also worth a test asserting every path in the `AGENTS.md` user-layer list is git-ignored, so the two lists cannot drift again. Happy to send either or both as a PR.

kmichels:
Closes #2364

## What

Adds `article-digest.md` to `.gitignore`, plus a test that keeps the user-layer list and `.gitignore` from drifting apart again.

## Why

`AGENTS.md` declares `article-digest.md` part of the **User Layer** — the files a candidate fills with their own personal data. It was the only entry on that list missing from `.gitignore`, so it sat untracked in the working tree where `git add .` would stage it.

| User-layer file (per `AGENTS.md`) | In `.gitignore` before this PR |
|---|---|
| `cv.md` | yes |
| `config/profile.yml` | yes |
| `modes/_profile.md` | yes |
| `modes/_custom.md` | yes |
| `portals.yml` | yes |
| `data/*`, `reports/*`, `output/*`, `interview-prep/*` | yes |
| **`article-digest.md`** | **no** |

The exposure lands hardest on contributors. Anyone who forks career-ops to send a PR is working in a repo pointed at a public remote, with their own proof points, metrics and employer details stageable. The failure is silent, and the file is one of the most personal in the layout after `cv.md` — the modes are explicitly told to put detailed proof points there.

I hit this while preparing #2343. My own `article-digest.md` was the sole untracked file in the tree.

## How

**`.gitignore`** — one line, placed next to `cv.md` under the existing `# Personal data (user fills these)` heading.

**`tests/user-layer-gitignored.test.mjs`** — parses the backtick-quoted paths straight out of the `**User Layer` line in `AGENTS.md` and asserts each is git-ignored. It tracks the document rather than duplicating the list, so adding a future user-layer file to `AGENTS.md` without a matching `.gitignore` entry fails the suite.

Directory globs (`data/*`) are checked with a probe path inside the directory; bare filenames are checked directly. `--no-index` so the check works whether or not a file exists yet on a given machine.

## Verification

The test genuinely detects the bug rather than passing vacuously. With the `.gitignore` line removed:

```
❌ article-digest.md is declared user-layer in AGENTS.md but is NOT git-ignored — personal data could be committed
📊 Results: 10 passed, 1 failed, 0 warnings
```

Restored:

```
✅ article-digest.md is git-ignored
📊 Results: 11 passed, 0 failed, 0 warnings
```

Full suite:

```
node test-all.mjs
📊 Results: 2728 passed, 0 failed, 0 warnings
```

## Note for existing users

This only stops *future* commits. Anyone who already committed an `article-digest.md` will need `git rm --cached article-digest.md` to untrack it. Worth a line in release notes if you think it is warranted.

## Summary by CodeRabbit

* **Tests**
  * Added validation to ensure declared User Layer paths are properly excluded from version control.
  * Improved reporting for missing, invalid, ignored, and unignored paths.

* **Chores**
  * Excluded the generated article digest file from version control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Personal data (user fills these)
 cv.md
+article-digest.md
 local/
 # Everything under data/ is generated or personal at runtime — only .gitkeep
 # placeholders are tracked (verify with `git ls-files data/`). A blanket rule
--- a/tests/user-layer-gitignored.test.mjs
+++ b/tests/user-layer-gitignored.test.mjs
@@ -1,0 +1,49 @@
+// tests/user-layer-gitignored.test.mjs
+//
+// AGENTS.md declares a User Layer: the files a candidate fills with their own
+// personal data. Every one of those paths must be git-ignored, or a contributor
+// working in a fork can stage their own CV, proof points or tracker into a public
+// repository with a reflexive `git add .`.
+//
+// article-digest.md drifted off .gitignore while remaining on the AGENTS.md list.
+// This test compares the two lists directly so they cannot diverge again.
+
+import { execFileSync } from 'child_process';
+import { readFileSync } from 'fs';
+import { join } from 'path';
+import { pass, fail, finish, ROOT } from './helpers.mjs';
+
+console.log('\n🔒 user-layer files are git-ignored');
+
+// Pull the declared user-layer paths straight out of AGENTS.md so the test tracks
+// the document rather than a hand-copied duplicate of it.
+const agents = readFileSync(join(ROOT, 'AGENTS.md'), 'utf-8');
+const line = agents.split(/\r?\n/).find(l => l.includes('**User Layer'));
+
+if (!line) {
+  fail('AGENTS.md no longer contains a "**User Layer" line — update this test');
+} else {
+  // Backtick-quoted paths, minus the glob suffix: `data/*` -> data/
+  const paths = [...line.matchAll(/`([^`]+)`/g)]
+    .map(m => m[1])
+    .map(p => (p.endsWith('/*') ? `${p.slice(0, -1)}` : p));
+
+  if (paths.length === 0) fail('parsed no paths from the AGENTS.md user-layer line');
+  else pass(`parsed ${paths.length} user-layer paths from AGENTS.md`);
+
+  for (const p of paths) {
+    // A directory glob is satisfied by a probe file inside it; a bare filename
+    // is checked directly. check-ignore exits 1 when the path is NOT ignored.
+    const probe = p.endsWith('/') ? `${p}__gitignore_probe__.md` : p;
+    let ignored = true;
+    try {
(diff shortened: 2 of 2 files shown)


santifer:
Welcome to career-ops, @kmichels — and this is a genuinely important catch. article-digest.md is where a candidate keeps their proof points, and it had drifted off .gitignore while staying on the AGENTS.md user-layer list: one reflexive git add . in a fork away from someone publishing their own career data. What makes the fix excellent is the test: it derives the protected list FROM AGENTS.md instead of hand-copying it, so the two can never diverge again — that's the exact copy-vs-derive discipline we've been enforcing everywhere this month, applied on a first contribution. Merged — your first one. Glad to have you here. (Your #2343 is separate and still has the two asks pending; no rush.)

Issue #1409, opened 2026-07-02
test-all: batch rate-limit pause test fails on Windows when WSL is installed (toBashPath returns /mnt/c path, bash resolves to Git Bash)


rrraadddii:
## Symptom

On Windows 11 with WSL2 installed, `node test-all.mjs` consistently fails:

```
13. Batch rate-limit pause
  X Batch rate-limit pause test crashed: ENOENT: no such file or directory, open 'C:\Users\...\Temp\co-batch-rate-XXXXXX\batch\batch-state.tsv'
```

`batch-runner.sh` itself is fine -- running the same fixture manually through Git Bash with a relative path passes (state file created, `paused_rate_limit` row written).

## Root cause

`toBashPath()` in `test-all.mjs` tries `wsl wslpath -u` **first**, and only falls back to `cygpath`:

```js
const out = execSync(`wsl wslpath -u "${forwardSlashed}"`, ...);  // tried first
...
const out = execSync(`cygpath -u "${forwardSlashed}"`, ...);      // fallback
```

When WSL is installed, `wslpath` succeeds and returns a **WSL-style** path (`/mnt/c/Users/...`). But the test then spawns `run('bash', [toBashPath(...)])`, and `execFileSync('bash')` resolves via PATH to **Git Bash** (`C:\Program Files\Git\usr\bin\bash.exe`), which mounts drives at `/c/...`, not `/mnt/c/...`. Git Bash can't find the script, `run()` swallows the error and returns `null`, `batch-state.tsv` is never created, and the readFileSync crashes the test.

So the produced path style and the bash that consumes it are guaranteed to disagree exactly when WSL is present.

## Side effect worth noting

Calling `wsl wslpath` **boots the WSL VM** (Vmmem) as a side effect of running the test suite -- unpleasant for users who keep WSL shut down for RAM reasons.

## Suggested fix

Try `cygpath` first (it ships with Git for Windows, which is also what provides `bash` on most Windows dev machines), and only fall back to `wslpath` if cygpath is unavailable. Alternatively, resolve which `bash` will actually be used (`where bash`) and pick the matching converter.

## Environment

- Windows 11 Pro 10.0.26200, Git for Windows bash first on PATH, WSL2 installed
- career-ops v1.15.0, Node v24.16.0

santifer:
[...]

Issue #36, opened 2026-04-07
Add progressive onboarding with capability visibility


FReptar0:
## Problem

The current onboarding flow has a binary gate: "If ANY file is missing, enter onboarding mode." This means:

1. A user with `cv.md` but no `portals.yml` cannot evaluate a job posting they paste directly -- they're blocked until all files exist.
2. There's no indication of **what capabilities** each setup step unlocks.
3. No way to check setup completeness mid-session.
4. No cross-validation between `portals.yml` title keywords and `profile.yml` target roles.

## Proposal

Replace the binary gate with a **progressive model** that never blocks users:

- If `cv.md` exists and user pastes a JD → evaluate it, note what's degraded ("PDF will use placeholder contact info until you set up profile.yml")
- If `cv.md` is missing and user pastes a JD → guide them to create CV first (can't match without one)
- If `portals.yml` is missing and user says "scan" → guide them to set up portals (only scan is blocked)
- First-time users still get the full guided onboarding (Steps 1-6)

### Additional features
- **Impact messages** (qualitative, no percentages) on each onboarding step showing what it unlocks
- **`/career-ops status` command** (`modes/status.md`) showing setup completeness grouped by CORE / DISCOVERY / ENRICHMENT / TRACKING
- **Cross-validation** between portals.yml and profile.yml to catch mismatches

### Changes required
- `modes/status.md`: New System Layer mode file
- `CLAUDE.md`: Progressive onboarding model, impact messages, status routing
- `DATA_CONTRACT.md`: Add `modes/status.md` to System Layer

No User Layer files modified. No new scripts. No new dependencies.

I have a PR ready that implements this. Happy to discuss the approach first.

PR #2366, opened 2026-07-31
fix(tests): find Git Bash under Scoop, and name the fallback shell when it fails

### LINKED ISSUE #2344: Test helper can't find Git Bash when Git is installed via Scoop — 5 spend_tier tests fail with a misleading message
## Summary

On Windows machines where Git for Windows is installed by **Scoop** (rather than the Program Files installer), `tests/helpers.mjs` fails to discover Git Bash, silently falls back to WSL bash, and the five "14. Batch spend_tier model routing" tests fail with `argv="", out=""`.

The failure message points at model routing, which is **not** broken. The captured stdout shows routing working correctly. The real event is `node: command not found` inside the fallback shell.

## Environment

- Windows 11
- Git 2.55.0.3 installed via Scoop → `%USERPROFILE%\scoop\apps\git\current\bin\bash.exe`
- Node v24 (Scoop), WSL2 also present
- `career-ops` v1.22.0

## Symptom

```
❌ economy spend_tier did not route to haiku: argv="", out=""
❌ premium spend_tier did not route to opus: argv="", out=""
❌ --model override did not win: argv="", out=""
❌ missing spend_tier did not default to standard: argv="", out=""
❌ invalid spend_tier did not fall back to standard: argv="", out=""

📊 Results: 2043 passed, 5 failed
```

## Root cause

1. `WINDOWS_BASH_CANDIDATES` lists only the Program Files layout:

   ```js
   const WINDOWS_BASH_CANDIDATES = [
     'C:\\Program Files\\Git\\bin\\bash.exe',
     'C:\\Program Files\\Git\\usr\\bin\\bash.exe',
   ];
   ```

   Scoop installs to `%USERPROFILE%\scoop\apps\git\current\...`, so both probes miss a Git Bash that is present.

2. `getBash()` therefore falls through to its WSL probe and returns `'bash'`, which on PATH resolves to `C:\WINDOWS\system32\bash.exe` (the WSL launcher).

3. `toBashPath()` has the same hardcoded assumption for `cygpath` and falls back to `wslpath` → `/mnt/c/...`. Bash and path form happen to agree here, so the script *does* launch — which is why the failure is silent rather than an obvious ENOENT.

4. **WSL has its own PATH.** The Windows `node`, and the stub `claude` binary the test injects via `env.PATH`, are both invisible inside it. `batch-runner.sh` reaches its merge step and dies:

   ```
   batch-runner.sh: line 654: node: command not found
   status: 127
   ```

5. `run()` catches, returns `null`; the test does `|| ''` → `argv=""`, `out=""`.

Instrumented repro confirming routing was correct all along:

```
--- THREW ---
status : 127
stderr : batch-runner.sh: line 654: node: command not found
stdout : === career-ops batch runner ===
         Model: claude-haiku-4-5 (spend_tier=economy)   ← correct
```

## Proposed fix

Add Scoop locations for **both** `bash` and `cygpath`. They must come from the same Git install — fixing only one reproduces #1409 in reverse, since cygpath emits `/c/...` while WSL bash expects `/mnt/c/...`.

Kept as fixed-shape literals joined onto `%SCOOP%` / `%USERPROFILE%` rather than a PATH search, so `WINDOWS_BASH_CANDIDATES` remains an allowlist of trusted literals for `resolveAllowedExecutable` (CodeQL #36/#41/#42).

```js
const SCOOP_ROOTS = [
  process.env.SCOOP,
  process.env.USERPROFILE ? join(process.env.USERPROFILE, 'scoop') : null,
].filter(Boolean);

const WINDOWS_BASH_CANDIDATES = [
  'C:\\Program Files\\Git\\bin\\bash.exe',
  'C:\\Program Files\\Git\\usr\\bin\\bash.exe',
  ...SCOOP_ROOTS.flatMap((root) => [
    join(root, 'apps', 'git', 'current', 'bin', 'bash.exe'),
    join(root, 'apps', 'git', 'current', 'usr', 'bin', 'bash.exe'),
  ]),
];

// cygpath must come from the SAME Git install as the bash above.
const WINDOWS_CYGPATH_CANDIDATES = [
  'C:\\Program Files\\Git\\usr\\bin\\cygpath.exe',
  ...SCOOP_ROOTS.map((root) => join(root, 'apps', 'git', 'current', 'usr', 'bin', 'cygpath.exe')),
];
```

and in `toBashPath()`:

```diff
-    const cygpathCmd = existsSync('C:\\Program Files\\Git\\usr\\bin\\cygpath.exe') ? 'C:\\Program Files\\Git\\usr\\bin\\cygpath.exe' : 'cygpath';
+    const cygpathCmd = WINDOWS_CYGPATH_CANDIDATES.find((p) => existsSync(p)) || 'cygpath';
```

Result on the affected machine: **2048 passed, 0 failed** (was 2043/5). `getBash()` now returns `scoop\apps\git\current\bin\bash.exe` and `toBashPath()` emits `/d/...` via cygpath.

## Secondary suggestion (separable — happy to drop)

`run()` collapses nonzero exit, timeout and ENOENT into a bare `null`, and callers do `run(...) || ''`. An environment fault then reaches assertions as an empty string and gets reported as a defect in the code under test. That is what turned "Git Bash is missing" into "model routing is broken" here, and it cost a fair amount of time to unpick.

Retaining the status/stderr behind an accessor would preserve the `string|null` contract and every existing caller:

```js
export function lastRunFailure() { return lastFailure; }
```

…set in `run()`'s catch, plus an opt-in `CAREEROPS_TEST_TRACE=1` stderr line (off by default, since several suites provoke command failures deliberately).

Glad to open a PR for the primary fix, the secondary one, or both — whichever you prefer.

sdkkds:
Fixes #2344. Both halves you asked for, in two commits.

## 1. Find the Git Bash that is actually installed

`WINDOWS_BASH_CANDIDATES` and the `cygpath` probe both hardcoded the Program Files layout. Scoop installs Git for Windows under the user profile, so on a Scoop machine both probes miss a Git Bash that is present and `getBash()` falls through to WSL.

Adds `%SCOOP%`, else `%USERPROFILE%\scoop`, as roots for both `bash` and `cygpath`. They stay fixed-shape literals joined onto those roots rather than a PATH search, so the list remains an allowlist of trusted literals for `resolveAllowedExecutable` (CodeQL #36/#41/#42).

`cygpath` is resolved from the same candidate set as `bash` on purpose: mixing a Program Files `cygpath` with a WSL `bash` is #1409 in reverse, since `/c/...` and `/mnt/c/...` are not interchangeable.

## 2. Say which shell it picked

`getBash()` returns the bare string `bash` from three branches — WSL, PATH, and give-up — so the return value alone cannot tell a caller which shell is about to run. It now records which probe won (`posix` / `git-bash` / `wsl` / `path` / `unresolved`), exported as `bashSource()`, and `run()` names the shell when a command fails under one of the fallbacks:

```
    [shell] this command ran under WSL bash (`wsl -e bash`) -- a different OS with its own PATH,
            because no Git Bash was found at any known location.
            The Windows `node` and any PATH-injected stub binary may be invisible there,
            so scripts calling node die with `node: command not found` (exit 127) and the
            assertion sees an empty result. Suspect the shell before the code under test.
            Install Git for Windows, or see formatRunFailure() for the raw stderr.
```

**Why unconditional rather than behind a flag.** `formatRunFailure()` already surfaces the child's stderr to callers that ask for it, but the shell that produced it stays invisible, and the entire failure mode of #2344 is that nobody suspects the shell. A flag only helps someone who already does.

Two constraints keep it out of the suites that provoke command failures deliberately:

- it fires only when `getBash()` landed on a **fallback** — a Git Bash resolved by literal path is unambiguous and stays silent, which is every correctly provisioned machine
- it fires at most **once per process**

## Tests

New `tests/shell-discovery.test.mjs` (6 tests): source is `null` before resolution and in the known set after; memoization stays coherent; a literal-path bash reports `git-bash` and exists on disk; the posix/win32 split holds; a failing command exposes its real exit status instead of a bare `null`; `lastRunFailure()` clears on success.

## Verification

Windows 11, Git installed via Scoop, Node v24.18.1:

- `node test-all.mjs` → **2716 passed, 0 failed, 1 warning** — the warning is the pre-existing `cv-sync-check.mjs exited with error (expected without user data)`, unrelated to this change. No stray `[shell]` output anywhere in the run.
- `node --test tests/shell-discovery.test.mjs` → 6/6
- Forced fallback (`SCOOP` and `USERPROFILE` pointed at a bogus root, so no Git Bash candidate exists) → `bashSource()` returns `wsl` and the block above prints exactly once; a second failure in the same process is silent.
- Before this branch, on the same machine, the five `spend_tier` tests failed with "did not route to haiku/opus/sonnet" while the captured stdout read `Model: claude-haiku-4-5 (spend_tier=economy)`.

I have no way to test the Program Files layout on this machine, but that path is untouched — the new entries are appended after the existing literals, so an existing install still wins the probe and reports `git-bash`.

https://claude.ai/code/session_01ThzrNt5WdcvomxMYgnT36i

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Windows shell discovery, including support for Bash installed through Scoop.
  - Improved path conversion by checking all trusted `cygpath` locations.
  - Added clearer diagnostics when commands fail or fallback shells are used.

- **Tests**
  - Expanded coverage for shell-source detection, fallback behavior, command failures, and recovery after successful commands.

### DIFF
--- a/tests/helpers.mjs
+++ b/tests/helpers.mjs
@@ -75,9 +75,41 @@ export function finish() {
 // to the OS), so a test can never be tricked into executing an arbitrary
 // binary — and CodeQL's uncontrolled-command-line finding is closed by
 // construction rather than dismissed (alerts #36/#41/#42).
+// Scoop installs Git for Windows under the user profile, not Program Files, so
+// a Program-Files-only list misses it entirely and getBash() falls through to
+// WSL bash. That fallback launches the script but not the environment: WSL has
+// its own PATH, so the Windows `node` (and any stub binary a test injects via
+// PATH) is invisible, and batch-runner.sh dies with `node: command not found`,
+// exit 127. run() converts that to null, the caller does `|| ''`, and the
+// assertion reports an empty argv -- which reads as a routing bug in the code
+// under test rather than a missing shell. That is what all five spend_tier
+// tests were doing on a machine where Git Bash was installed the whole time
+// (#2344).
+//
+// Kept as fixed-shape literals joined onto %USERPROFILE% / %SCOOP% rather than
+// a PATH search, so this stays an allowlist of trusted literals (see
+// resolveAllowedExecutable below and CodeQL alerts #36/#41/#42).
+const SCOOP_ROOTS = [
+  process.env.SCOOP,
+  process.env.USERPROFILE ? join(process.env.USERPROFILE, 'scoop') : null,
+].filter(Boolean);
+
 const WINDOWS_BASH_CANDIDATES = [
   'C:\\Program Files\\Git\\bin\\bash.exe',
   'C:\\Program Files\\Git\\usr\\bin\\bash.exe',
+  ...SCOOP_ROOTS.flatMap((root) => [
+    join(root, 'apps', 'git', 'current', 'bin', 'bash.exe'),
+    join(root, 'apps', 'git', 'current', 'usr', 'bin', 'bash.exe'),
+  ]),
+];
+
+// Same discovery problem, same fix: cygpath must come from the SAME Git
+// install as the bash above. Mixing them is #1409 in reverse -- cygpath emits
+// /c/... while WSL bash expects /mnt/c/..., so the path silently fails to
+// resolve inside the shell that receives it.
+const WINDOWS_CYGPATH_CANDIDATES = [
+  'C:\\Program Files\\Git\\usr\\bin\\cygpath.exe',
+  ...SCOOP_ROOTS.map((root) => join(root, 'apps', 'git', 'current', 'usr', 'bin', 'cygpath.exe')),
 ];
 
 /**
@@ -239,7 +271,7 @@ export function toBashPath(wpath) {
   try {
     // execFileSync: the path is passed as an argv element, never interpolated
     // into a shell string, so quotes/spaces in it can't be re-parsed.
-    const cygpathCmd = existsSync('C:\\Program Files\\Git\\usr\\bin\\cygpath.exe') ? 'C:\\Program Files\\Git\\usr\\bin\\cygpath.exe' : 'cygpath';
+    const cygpathCmd = WINDOWS_CYGPATH_CANDIDATES.find((p) => existsSync(p)) || 'cygpath';
     const out = execFileSync(cygpathCmd, ['-u', forwardSlashed], { stdio: ['pipe', 'pipe', 'ignore'] }).toString().trim();
     if (out) return out;
   } catch {}
--- a/tests/helpers.mjs
+++ b/tests/helpers.mjs
@@ -171,6 +171,7 @@ export function run(cmd, args = [], opts = {}) {
       stdout: e?.stdout == null ? '' : String(e.stdout),
       stderr: e?.stderr == null ? '' : String(e.stderr),
     };
+    warnFallbackShell(exe);
     return null;
   }
 }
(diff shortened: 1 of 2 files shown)


santifer:
Welcome to career-ops, @sdkkds! This closes the loop on your own #2344 exactly as scoped: the Scoop path (expanded from USERPROFILE) joins the candidate list, and the misleading-error half — arguably the worse half, as you put it in the issue — is fixed by naming which shell the fallback actually picked when a spend_tier test then fails. Diagnosis, issue, fix, all yours, delivered within a day of the invitation. Merged — your first contribution. Glad to have you here.

Issue #1733, opened 2026-07-09
bug(merge-tracker): reserved report numbers silently discarded when <= tracker max — row # permanently diverges from report file (follow-up to #1704)


Schlaflied:
## Problem

Same line as #1704, opposite branch:

```js
const entryNum = addition.num > maxNum ? addition.num : ++maxNum;
```

#1704/#1705 cover the `addition.num > maxNum` branch (collision detection). The `else` branch has a distinct, silent failure mode found by @kevind589 while running a real 11-URL parallel pipeline batch (full forensics in [#1704 (comment)](https://github.com/santifer/career-ops/issues/1704#issuecomment-4929129288)):

When `addition.num <= maxNum` — the **common** case, since report numbers come from `reserve-report-num.mjs` (counts off `reports/`) while `maxNum` comes from `applications.md`'s row count, two independent counters — the addition's own reserved number is **discarded** and replaced with `++maxNum`. No collision, no warning, but the report file's embedded number (filename + Machine Summary YAML) now permanently disagrees with the tracker row `#` it landed in.

**Observed blast radius:** 9 of 11 batch evaluations landed on tracker rows #418–#426 while their report files say 410–420 (e.g. `reports/415-hpe-cloudops-2026-07-09.md` → tracker row #421). Worse, `set-status.mjs 415` then silently edited a pre-existing, unrelated row #415 — a single, valid-looking, *wrong* match that #1705's duplicate-selector guardrail cannot catch (it guards 2+ matches, not exactly-one-wrong-match).

## Proposed fix (two independent layers)

1. **merge-tracker.mjs — prefer the addition's own number when it's free:** if `addition.num` is set and not already used by an existing row (and not claimed earlier in the same run), keep it as the tracker row `#` even when `<= maxNum`. Only fall back to `max+1`-style assignment when the number is genuinely taken — and in that fallback, warn loudly that the row number no longer matches the report file.
2. **set-status.mjs — report-link sanity check:** when the selector is a number and the matched row's `Report` column links a file whose embedded number differs, warn (or require `--force`) before writing. Independent defense that catches pre-existing drifted rows too.

Layer 1 stops new drift; layer 2 protects against the drift already in the wild.

## Relationship to #1705

Deliberately separate: #1705 = duplicate detection/prevention on the `>` branch; this = number fidelity on the `<=` branch. Both credit to @kevind589's field report for isolating the second mode. Happy to take this — implementation will land after/atop #1705 to avoid conflicting edits to the same line.

[...]

Issue #1425, opened 2026-07-02
verify-pipeline: detect duplicate reports (same company+role) and orphan reports (no tracker row)


santifer:
Surfaced by a real apply-run field test (~90 evaluations, one 8-subagent parallel fan-out).

## Problem

Two concurrent evaluators produced two reports for the SAME role. `merge-tracker` correctly dedups the TRACKER (update-in-place by score) and `verify-pipeline.mjs` checks tracker duplicates — but nothing detects:

1. Two `reports/*.md` files covering the same company+role
2. Report files with no corresponding tracker row (orphans)

Both were caught by eye, which doesn't scale.

## Proposal

Two new checks in `verify-pipeline.mjs`:
- **dup-report check**: group report files by (company-slug, role-slug), warn on >1
- **orphan check**: every `reports/NNN-*.md` must have a tracker row referencing report NNN; list the ones that don't

Warning-level (not fail) — duplicates can be legitimate (re-evaluation after a JD change).

PR #2370, opened 2026-07-31
fix(verify-pipeline): orphan check masks and invents orphans (#1425)

kevind589:
…real orphans (#1425)

Check 10 resolves "is this report referenced by a tracker row?" three ways. Two of them are wrong, and they fail in opposite directions — one hides real orphans, the other invents them.

1. The row's own number is credited UNCONDITIONALLY.

   Row and report numbers are independent counters that diverge in normal operation: #1733 established that a reserved report number is discarded when it is <= the tracker max, permanently desynchronising the two. Check 10 nonetheless treats a row's own number as a reference even when the cell links somewhere else, so a row numbered 306 that legitimately links report 311 also silently "references" an unrelated orphaned report 306 — and the orphan is never reported.

   The row number is now a last resort, used only when the cell carries no markdown link at all. Link-less legacy rows keep the old behaviour.

2. Only the FIRST link in a cell is read.

   `e.report.match(/\[(\d+)\]/)` returns one match. A cell may carry several — "[019](...) / [111](...)" is the documented form for a re-evaluation that keeps both reports on record — so every link after the first false-positives as an orphan. Now matched globally with matchAll, for both the [NNN] link text and the NNN- filename prefix.

Tests: five assertions added alongside the existing #1425 fixture, covering a dual-link cell, a row number colliding with an unrelated orphan, and a link-less row that must keep the fallback. Each was run against unpatched main first and observed failing on exactly the two defects above (2 failed), then against this change (0 failed) — a fixture that cannot fail proves nothing. Full node test-all.mjs passes: 2722 passed, 0 failed.

Refs #1425, #1733.

## What does this PR do?

Fixes two defects in verify-pipeline.mjs's orphan-report check (#1425). The row's own number was credited unconditionally, masking real orphans; and only the first link in a report cell was read, so dual-link cells false-positived. Adds five fixture assertions, each observed failing before being trusted.

## Type of change

- [X] Bug fix
- [ ] New feature

## Checklist

- [X] I have read [CONTRIBUTING.md](https://github.com/santifer/career-ops/blob/main/CONTRIBUTING.md)
- [X] My PR does not include personal data (CV, email, real names, scan results, or pipeline data)
- [X] I ran `node test-all.mjs` and all tests pass
- [X] My changes respect the [Data Contract](https://github.com/santifer/career-ops/blob/main/DATA_CONTRACT.md) (no modifications to user-layer files)
- [X] My changes align with the [project roadmap](https://github.com/santifer/career-ops/discussions/156)

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved orphan report detection when tracker cells contain multiple report links.
  - Prevented unrelated row numbers from incorrectly matching reports when explicit links are present.
  - Preserved row-number fallback behavior for rows without report links.
  - Correctly recognizes linked reports so they are not flagged as orphaned.

- **Tests**
  - Added regression coverage for report-link resolution scenarios.

### DIFF
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -6979,6 +6979,89 @@ try {
   fail(`verify-pipeline report checks crashed: ${e.message}`);
 }
 
+// ── VERIFY-PIPELINE ORPHAN REFERENCE RESOLUTION (#1425 follow-up) ────────────
+// Check 10 resolves "is this report referenced?" three ways. Two of them were
+// wrong:
+//   (a) a cell may carry SEVERAL links ("[901](…) / [902](…)" — a re-evaluation
+//       keeping both reports on record). A single .match() sees only the first,
+//       so every later link false-positives as an orphan.
+//   (b) the row's own number was credited UNCONDITIONALLY. Row and report
+//       numbers are independent counters that diverge in normal operation
+//       (#1733), so a row that links elsewhere silently "references" an
+//       unrelated report sharing its number, masking a real orphan.
+console.log('\n🧪 Testing verify-pipeline orphan reference resolution (#1425 follow-up)');
+try {
+  const orTmp = mkdtempSync(join(tmpdir(), 'career-ops-verify-orphan-'));
+  try {
+    const orReports = join(orTmp, 'reports');
+    mkdirSync(orReports, { recursive: true });
+    const orTracker = join(orTmp, 'applications.md');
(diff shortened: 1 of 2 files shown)


santifer:
Welcome to career-ops, @kevind589! Two real defects, one surgical fix: the row crediting its own number unconditionally (masking genuine orphans — the #1733 counter-divergence class, your quote checks out) and single .match() inventing orphans on dual-link re-evaluation cells. Verified it composes cleanly with today's normalizeTextKey work, 45 commits ahead of your base. Merged — your first contribution. 🚀

PR #2376, opened 2026-07-31
fix(liveness): classify transient 5xx as uncertain, not expired

### LINKED ISSUE #2371: bug(liveness): a transient 502/500/504 classifies as expired, so scan-history permanently filters a live posting
## What happens

`classifyLiveness()` (liveness-core.mjs) guards 403 and 503 as `access_blocked`, but every other 5xx falls through to the body heuristics. A gateway error body is short ("502 Bad Gateway / nginx"), so it lands in the `insufficient_content` branch and the posting classifies **expired**.

Repro, straight against the exported function:

```js
import { classifyLiveness } from './liveness-core.mjs';
classifyLiveness({ status: 502, bodyText: '502 Bad Gateway\nnginx', applyControls: [] });
// { result: 'expired', code: 'insufficient_content', ... }
```

Same for 500 and 504.

## Why it matters

A false "expired" is the failure mode this module goes out of its way to avoid everywhere else (the bot-challenge and redirected-off-posting paths both say so in their comments): `scan --verify` writes the offer to scan-history as `skipped_expired`, and `shouldDedupScanHistoryRow` then filters it from every future scan. One origin hiccup during a deploy and a live job disappears permanently.

The browser rung is affected too: `liveness-browser.mjs` passes the real HTTP status into `classifyLiveness`, and `page.goto` resolves (does not throw) on a 5xx, so nothing upstream catches it.

## Fix

Treat `status >= 500` as `uncertain` (a transient server error), keeping 503's more specific `access_blocked` code. 404/410 stay authoritative. PR with tests incoming.

Scott-Emberson:
Closes #2371.

`classifyLiveness()` guarded 403 and 503 as `access_blocked`, but any other 5xx fell through to the body heuristics, where a short gateway-error body ("502 Bad Gateway / nginx") lands in the `insufficient_content` branch and reads as **expired**. `scan --verify` then writes the offer to scan-history as `skipped_expired`, and `shouldDedupScanHistoryRow` filters it from every future scan, so a single origin hiccup permanently hides a live job.

The change: `status >= 500` returns `uncertain` with a new `server_error` code. 503 keeps its more specific `access_blocked` classification (checked first), and 404/410 stay authoritative expiry. No caller switches on the code string, so the new value is additive.

Tests: three transient statuses (500/502/504) assert `uncertain`/`server_error`, plus regression coverage that 503 keeps `access_blocked` and 404 still expires. `node tests/liveness-core.test.mjs` green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of server-side HTTP errors by classifying status codes 500 and above as uncertain rather than expired.
  * Preserved existing classifications for blocked-access responses, including HTTP 503, and expired responses for HTTP 404 and 410.

* **Tests**
  * Added regression coverage for HTTP 500, 502, 503, 504, 404, and 410 responses to ensure classifications remain accurate.

### DIFF
--- a/liveness-core.mjs
+++ b/liveness-core.mjs
@@ -132,6 +132,14 @@ export function classifyLiveness({ status = 0, requestedUrl = '', finalUrl = '',
   if (status === 403 || status === 503) {
     return { result: 'uncertain', code: 'access_blocked', reason: `HTTP ${status} (access blocked, likely anti-bot)` };
   }
+  // Any other 5xx is a transient origin error (502/504 gateway hiccups, 500s
+  // during deploys), not evidence the posting is gone. Without this guard the
+  // short error body ("502 Bad Gateway / nginx") falls through to the
+  // insufficient-content heuristic and reads as expired — and a false
+  // "expired" permanently dedup-filters a real job out of future scans.
+  if (status >= 500) {
+    return { result: 'uncertain', code: 'server_error', reason: `HTTP ${status} (transient server error)` };
+  }
 
   const expiredUrl = firstMatch(EXPIRED_URL_PATTERNS, finalUrl);
   if (expiredUrl) {
--- a/tests/liveness-core.test.mjs
+++ b/tests/liveness-core.test.mjs
@@ -59,3 +59,31 @@ classifyLiveness({
 }).result !== 'expired'
   ? pass('"job application form has been filled" (no "out") is NOT expired')
   : fail('false positive: "application form has been filled" read as an expired req');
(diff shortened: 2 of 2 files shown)


santifer:
Executed on both sides, @Scott-Emberson: main really did classify transient 5xx as expired (the #2371 repro live), your guard ordering keeps real 404/410 and 503 semantics exactly where they were, and server_error is genuinely additive downstream — nobody switches on it. Merged. 🚀

Issue #2296, opened 2026-07-29
company-history: a company that HIRED you is dropped from the responsiveness card (no fact, labelled no-history)


abankar1:
### What

In `company-history.mjs`, the responsiveness axis classifies each tracker row into one of two branches: `applied` (silence candidate) or `RESPONDED_STATUSES` (the company answered). That set — and the parallel `OUTCOME_LABELS` map — lists `responded / interview / offer / rejected` but **not `hired`**:

```js
const RESPONDED_STATUSES = new Set(['responded', 'interview', 'offer', 'rejected']);
const OUTCOME_LABELS = { responded: 'Responded', interview: 'Interview', offer: 'Offer', rejected: 'Rejected' };
```

A row whose status normalizes to `hired` matches neither branch, so **no responsiveness fact is produced for it at all**. Consequences on the evidence card:

- A company whose only interaction is a hire is labelled `no-history` — as if nothing ever happened, despite the strongest possible positive response.
- If that company also has an older silent `Applied` row, `hasResponded` stays `false`, so the card can read `silent-on-you` against a company that literally hired you.

The module's own comment says *"a rejection IS an answer."* A hire is the strongest answer of all — it should never be dropped.

`Hired` is canonical (`states.yml`, added in #2050) and `normalizeStatus('Hired') → 'hired'`, so the tracker's real value already flows here; only these two literals were missed. Same incomplete-`Hired`-propagation class as #2289 / #2294.

### Fix
Add `'hired'` to `RESPONDED_STATUSES` and `hired: 'Hired'` to `OUTCOME_LABELS`. PR incoming (with a `--self-test` regression).

Issue #45, opened 2026-04-07
Generated PDF is content-strong but visually dense / hard to scan


Ruthwik-Data:
### Problem

I tested Career-Ops on a real job application flow. The evaluation quality and tailored resume content were strong, but the initial generated PDF felt too plain and visually dense. I needed several follow-up prompts to reach a recruiter-friendly layout.

Observed behavior -

Content quality and tailoring were solid.

PDF layout felt visually dense and a bit “wall-of-text”-like.

Bullet points were sometimes too long for quick scanning.

Section hierarchy (headings vs body) was not visually strong enough.

The top third of the resume (the “6-second scan” zone) did not stand out clearly on the first pass.

Adjacent roles could feel visually compressed, making it harder to distinguish transitions between jobs.

Expected behavior

By default, I’d expect:

A cleaner, ATS-safe layout out of the box.

Clearer spacing between sections and between bullet points.

More prominent section headers (Experience, Skills, Education, etc.).

Stronger visual hierarchy so a recruiter can parse the page in ~6 seconds.

Clear separation between roles (company, title, dates) without relying on extra prompts.

Why this matters

Career-Ops is already very strong on evaluation and content tailoring. But for real-world usability, resume presentation quality matters as much as content: if the PDF feels dense or visually flat, recruiters may not fully benefit from the quality of the tailored bullets. The default template being closer to “submission-ready” reduces the need for multiple manual formatting prompts on every application.

### Proposed solution

Improving the default HTML/PDF template would help a lot. Some ideas:

Increase vertical spacing between sections and roles.

Add slightly more padding / whitespace between bullet points.

Make section headers more prominent (size/weight/spacing) while staying ATS-safe.

Emphasize the top third of the page (headline, summary, key skills) for quick recruiter scanning.

Ensure each role has a bit more visual separation (company/title/dates grouping and spacing) so adjacent roles don’t blur together.

I’m happy to test updated templates with real JDs and share before/after PDFs if that’s helpful.

fahimimtiaz12:
I opened a PR for this. Please review when convenient.

PR #29, opened 2026-04-06
Add LinkedIn Jobs scanning (Level 4) via browser-use

tomron87:
## Why

LinkedIn is the largest job board, but it has no public API and blocks traditional scraping. The current scan mode discovers jobs via 3 levels (Playwright direct, Greenhouse API, WebSearch) — none of which cover LinkedIn. This means users miss a significant chunk of the job market.

## What this adds

**Level 4 — LinkedIn Jobs scan** using [`browser-use`](https://github.com/browser-use/browser-use) CLI with the user's logged-in Chrome profile. The scan agent runs `browser-use --profile` to open LinkedIn Jobs search pages, extracts job listings from the DOM, and feeds them into the same title filter + dedup pipeline as Levels 1-3.

### How it works

1. The scan agent reads `linkedin:` config from `portals.yml`
2. Builds LinkedIn search URLs from `title_filter.positive` keywords (multi-word terms like "Data Scientist" become individual searches; single words like "AI", "ML" are combined with OR)
3. Applies location and work-type filters as LinkedIn URL parameters (`f_WT`, `f_TPR`, `location`)
4. Opens each search in Chrome via `browser-use --profile`, extracts job cards (title, company, location, URL) using DOM selectors
5. Results go through the same title filter and dedup pipeline, then get added to `pipeline.md` and `scan-history.tsv`

### Configuration

New `linkedin:` section in `portals.yml` (disabled by default in the template):

```yaml
linkedin:
  enabled: true
  chrome_profile: "Default"
  max_queries: 10          # cap on number of keyword searches
  time_posted: "7d"        # 24h, 7d, or 30d
  location: ""             # falls back to profile.yml location
  work_type: "any"         # any, onsite, remote, hybrid
  max_results_per_query: 25
  delay_between_searches: 5  # seconds, to avoid throttling
```

### Error handling

- **Not logged in:** Detects login/authwall pages → aborts Level 4 with clear warning
- **CAPTCHA / throttling:** Detects "unusual activity" → aborts with warning
- **browser-use not installed:** Skips Level 4 with install instructions
- **Chrome profile not found:** Skips with helpful message

### Cross-platform dedup

The same job at a company may appear on both LinkedIn (`linkedin.com/jobs/view/123`) and the company's own careers page (`job-boards.greenhouse.io/company/jobs/456`). Level 4 adds a second dedup tier that matches `company + normalized title` across platforms, logged as `skipped_dup_xplatform` in scan-history.

### Browser conflict management

[...]

Issue #2369, opened 2026-07-31
web: core constants hand-copied into web/ drift silently (a bug class, 3 known instances)


santifer:
## The class

The web hand-copies constants that the core owns, each guarded only by a comment promising they stay in sync. That promise has already broken once in production. Two instances surfaced in one week, so this is a class, not a coincidence.

| # | Web copy | Core source | Status |
|---|---|---|---|
| 1 | `lib/core/states.ts` → `FALLBACK` | `templates/states.yml` | **Broke** when the core gained `hired` (9th state) → status writeback returned HTTP 400 on the degraded path. Fixed in #2282; frozen by `test-all` 55.3b. |
| 2 | `lib/followups.ts` → `CADENCE_DEFAULTS` | `followup-cadence.mjs` → `DEFAULT_CADENCE` | Values match today. Arrives with #1422. |
| 3 | `lib/career-ops.ts` → `prereqs` in `doctorState()` | `doctor.mjs` → `USER_LAYER_PREREQS` | Matches today (`cv.md`, `config/profile.yml`, `modes/_profile.md`, `portals.yml`). A 5th core prereq would be silently missed by the onboarding banner. |

Also worth a look (logic mirrors rather than copied lists, so harder to assert): `lib/core/portals.ts` mirrors `scan.mjs::buildTitleFilter/buildLocationFilter`, and `lib/explore.ts` mirrors scan's keyword/location list semantics.

## Why each copy exists (they're not gratuitous)

- **#2** is a display baseline for the cadence settings form.
- **#3** is a deliberate fast path: `doctorState()` runs in server components on every render, and shelling out to `doctor.mjs` there would be slow (the web does also expose `/api/doctor` for the real thing).

So the fix is per-instance, not one sweeping refactor.

## Proposed fixes

1. **`CADENCE_DEFAULTS` → delete it.** `followup-cadence.mjs --json` already emits `cadenceConfig` (the *effective* cadence: defaults + profile overrides), and `/api/followups` already receives that field. The settings form can source its baseline from there. Open question: the form may want *pure defaults* vs *effective config* — either emit both from the core, or decide that showing what's actually in effect is the more honest baseline.
2. **`prereqs` → derive or assert.** Either have `doctor.mjs --json` expose the prereq list so the web reads it once and caches, or add a `test-all` assertion cross-checking the two lists (cheaper, and the 55.3b pattern already works).
[...]

PR #2377, opened 2026-07-31
fix(tracker): rank Hired in dedup, canonicalize its aliases, and stop a bare tracker-number match merging different roles

### LINKED ISSUE #2372: bug(tracker): Hired is unranked in dedup and unaliased in followup-cadence — a repost can delete the accepted-job row
The Hired sweep (#2050, #2145, #2249, #2296) covered merge-tracker, stats, analyze-patterns and both dashboards, but two consumers were missed, and while fixing them I found a third dedup-safety hole in the same code path.

## 1. dedup-tracker: a repost can delete the accepted-job row

`STATUS_RANK` in dedup-tracker.mjs has no `hired` entry, so `statusRank('Hired')` is 0 and `isAdvancedStatus()` is false. The advanced-status protection that keeps Applied+ rows safe from exact-title siblings never fires for the one row it should protect most.

Scenario: tracker has `#12 Acme | Platform Engineer | 3.8/5 | Hired` and the role gets re-listed, evaluated as `#57 Acme | Platform Engineer | 4.2/5 | Evaluated`. Dedup clusters them (exact title), keeps the higher score, promotes `Evaluated` (rank 2) over `Hired` (rank 0), and deletes the record of the job the user accepted.

## 2. followup-cadence: hired aliases normalize to themselves

`templates/states.yml` declares `aliases: [contratado, contratada, hired, accepted, accept]` for `hired`, but the `ALIASES` map in followup-cadence.mjs has none of them. `normalizeStatus('Accepted')` returns `'accepted'`, so:

- `stats.mjs` `canonicalStatus()` returns `Unknown`, and the row drops out of `everApplied`/`everOffer` in the funnel. The best outcome in the tracker undercounts silently.
- `company-history.mjs` misses it in `RESPONDED_STATUSES` and labels the company that hired you `no-history`. This is the exact regression #2296 fixed, but only for the literal `Hired` spelling.

## 3. dedup-tracker: a bare tracker-number match bypasses both guards

`sameReportIdentity()` returns true on `a.num === b.num` alone, which short-circuits `roleMatch` before the exact-title check and the advanced-status guard. But duplicate tracker numbers never mean the same application: they are the merge-bug artifact verify-pipeline Check 12 (#1704) exists to flag.

Scenario: `#45 Acme | Data Engineer | Applied` and `#45 Acme | ML Engineer | Evaluated 4.5` share a stale duplicate number. Dedup treats them as one record and deletes the Applied row of a different opening.

PR with regression tests for all three incoming. The existing duplicate-number test case (two `#29` rows with the same role) still passes: a number match now also requires the same exact role title, which is the case that test encodes.

Scott-Emberson:
Closes #2372.

Three fixes in the tracker's dedup and status-normalization path.

**dedup-tracker `STATUS_RANK` had no `hired` entry**, so `statusRank('Hired')` was 0, `isAdvancedStatus()` was false, and the guard that protects Applied+ rows from exact-title siblings never fired for the accepted-job row. A higher-scored repost could delete it. Hired and its states.yml aliases now rank above Offer.

**followup-cadence `ALIASES` lacked the states.yml hired aliases** (contratado/contratada/accepted/accept), so those rows normalized to themselves and every consumer of `normalizeStatus` looking for `hired` dropped them: stats.mjs `canonicalStatus` returned `Unknown` (funnel undercounts `everApplied`/`everOffer`), and company-history labelled the company that hired you `no-history` — the #2296 regression, surviving for every spelling but the literal one.

**`sameReportIdentity` treated a bare tracker-number match as same-application evidence**, short-circuiting `roleMatch` ahead of both the exact-title check and the advanced-status guard. Duplicate tracker numbers are the merge-bug artifact verify-pipeline Check 12 flags, not a same-application signal, so two different roles sharing a stale `#45` merged and the Applied row of a different opening was deleted. A number match now also requires the same exact role title.

Tests: the existing dedup fixture gains a Hired-vs-repost pair and a duplicate-number pair with different roles; five hired-alias assertions added to followup-cadence.test.mjs. The pre-existing duplicate-number case (two `#29` rows, same role) still merges, which is the behaviour that test encodes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Recognized additional English and Spanish variations of “hired” status, ensuring they are consistently displayed and prioritized.
  - Improved duplicate report matching by prioritizing report-link numbers and validating role titles when tracker numbers overlap.
  - Prevented unrelated roles with shared tracker numbers from being incorrectly merged.
  - Preserved hired records when duplicate postings have higher-scored matches.

### DIFF
--- a/dedup-tracker.mjs
+++ b/dedup-tracker.mjs
@@ -40,6 +40,11 @@ const STATUS_RANK = {
   'responded': 4,
   'interview': 5,
   'offer': 6,
+  // Hired outranks everything: the accepted-job record must never lose a
+  // dedup contest to a repost row (aliases from templates/states.yml).
+  'hired': 7,
+  'accepted': 7,
+  'accept': 7,
   // Spanish aliases — kept for backwards compat with existing tracker data
   'no_aplicar': 0,
   'no aplicar': 0,
@@ -52,6 +57,8 @@ const STATUS_RANK = {
   'respondido': 4,
   'entrevista': 5,
   'oferta': 6,
+  'contratado': 7,
+  'contratada': 7,
 };
 
 /**
@@ -138,19 +145,22 @@ function extractReportNum(reportStr) {
 /**
  * Determine whether two tracker rows point to the same exact report identity.
  *
- * Exact identity is stronger than fuzzy role matching. If two rows share the
- * same tracker number or bracketed report number, dedup may treat them as the
- * same record even when an advanced status is present.
+ * Exact identity is stronger than fuzzy role matching: it may cluster rows
+ * even when an advanced status is present. Matching bracketed report numbers
+ * are that evidence. A shared tracker number alone is NOT — duplicate tracker
+ * numbers are a known artifact of the old merge bug (verify-pipeline Check 12
+ * exists because they never mean the same application), so a bare number match
+ * only counts when the rows also carry the same exact role title.
  *
  * @param {object} a - First parsed applications.md row.
  * @param {object} b - Second parsed applications.md row.
  * @returns {boolean} True when both rows represent the same report identity.
  */
 function sameReportIdentity(a, b) {
-  if (a.num === b.num) return true;
   const reportA = extractReportNum(a.report);
   const reportB = extractReportNum(b.report);
-  return reportA !== null && reportA === reportB;
+  if (reportA !== null && reportA === reportB) return true;
+  return a.num === b.num && normalizeRole(a.role) === normalizeRole(b.role);
 }
 
 /**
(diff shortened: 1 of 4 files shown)


santifer:
Three real fixes in 58 lines, @Scott-Emberson — and the third one (a bare tracker-number match no longer merging rows whose titles differ) is the same failure family abankar1 and I have been closing all week. One finding from verification worth a follow-up rather than a change here: STATUS_RANK still omits three documented Applied aliases from states.yml ('sent', 'enviada', 'aplicada'), and nothing in CI ties that hand-copied vocabulary to states.yml — it's now duplicated in five places. That's the copy-vs-derive pattern we froze for two other lists in #2369; this one deserves the same treatment. Merged! 🚀

PR #2378, opened 2026-07-31
fix(cv-latex): single mailto: scheme + escape OT1-unsafe < > | glyphs

### LINKED ISSUE #2373: bug(cv-latex): every generated CV gets a mailto:mailto: email link, and < > | render as the wrong glyphs
Two rendering defects that hit every generated LaTeX CV. Both reproduce with the shipped self-test payload (`node build-cv-latex.mjs --test`).

## 1. The email link is `mailto:mailto:...`

`sanitizeUrl()` in lib/latex-escape.mjs prefixes a bare address with `mailto:`:

```js
sanitizeUrl('test@example.com') // 'mailto:test@example.com'
```

and the template then wraps it in its own scheme (templates/cv-template.tex):

```tex
\href{mailto:{{EMAIL_URL}}}{...}
```

so the compiled PDF carries `\href{mailto:mailto:test@example.com}`, which no mail client opens. Supplying an explicit `mailto:` in the payload produces the same doubled link. The LINKEDIN/GITHUB hrefs on the neighboring lines already trust `sanitizeUrl`'s scheme; EMAIL is the odd one out.

## 2. `<`, `>` and `|` print as the wrong characters

`escapeLatex()` passes `<`, `>` and `|` through raw, and the template loads no `fontenc`, so pdflatex renders them with OT1 text-font glyphs:

- `<` prints as ¡ and `>` as ¿ ("Cut p99 latency to <100ms" becomes "Cut p99 latency to ¡100ms")
- `|` prints as an em dash. The self-test's own `contact_line: 'City, State | +1 234 567 8900'` renders "City, State — +1 234 567 8900"

Engineering bullets are full of `<`, `>` and `|` (latency targets, throughput multipliers, tool pipelines), so this corrupts real content rather than an edge case.

## Fix

Emit `\textless{}`, `\textgreater{}` and `\textbar{}` from `escapeLatex`, and drop the template's extra `mailto:` so `sanitizeUrl` owns the scheme. PR with a regression suite (tests/latex-escape.test.mjs) incoming.

Scott-Emberson:
Closes #2373.

Two defects in every generated LaTeX CV, both reproducible with `node build-cv-latex.mjs --test`.

`sanitizeUrl()` prefixes a bare address with `mailto:` and the template wrapped `{{EMAIL_URL}}` in a second one, so the compiled PDF carried `\href{mailto:mailto:...}` and the email link never opened. The template now trusts `sanitizeUrl`'s scheme, matching the LINKEDIN and GITHUB hrefs beside it.

`escapeLatex()` passed `<`, `>` and `|` through raw. Under pdflatex's default OT1 text encoding those have no glyph: `<` and `>` print as inverted punctuation and `|` as an em dash, so "p99 <100ms" rendered "p99 ¡100ms" and the sample contact line "City, State | +1 234..." picked up a dash. They now emit `\textless{}`, `\textgreater{}` and `\textbar{}`.

I went with escaping rather than adding `\usepackage[T1]{fontenc}` to the preamble: it fixes existing `.tex` files without a template change on the user's side, and doesn't alter font metrics for anyone whose CV currently compiles.

Tests: new `tests/latex-escape.test.mjs` covers the three glyphs, both `sanitizeUrl` scheme cases, and a guard that the template doesn't reintroduce the doubled `mailto:`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved LaTeX output by safely escaping `<`, `>`, and `|` characters.
  * Corrected email links to prevent duplicate `mailto:` prefixes.
  * Preserved proper handling for both bare email addresses and explicit email URLs.
  * Ensured generated email, LinkedIn, and GitHub links render correctly.

### DIFF
--- a/lib/latex-escape.mjs
+++ b/lib/latex-escape.mjs
@@ -24,6 +24,13 @@ export function escapeLatex(text, mode = 'text') {
       case '%': out.push('\\%'); break;
       case '$': out.push('\\$'); break;
       case '#': out.push('\\#'); break;
+      // With pdflatex's default OT1 encoding these three have no glyph in the
+      // text font: < and > print as inverted ¡/¿ and | as an em-dash, so
+      // "p99 <100ms" renders "p99 ¡100ms". The text commands are safe in all
+      // encodings.
+      case '<': out.push('\\textless{}'); break;
+      case '>': out.push('\\textgreater{}'); break;
+      case '|': out.push('\\textbar{}'); break;
       case '\u00B1': out.push('$\\pm$'); break;
       case '\u2192': out.push('$\\rightarrow$'); break;
       default: out.push(ch);
--- a/templates/cv-template.tex
+++ b/templates/cv-template.tex
@@ -86,7 +86,9 @@
         \\ {{CONTACT_LINE}}\\
         \small
         % EMAIL
-        \href{mailto:{{EMAIL_URL}}}{\raisebox{-0.2\height}\faEnvelope\  \underline{{{EMAIL_DISPLAY}}}} ~
+        % EMAIL_URL already carries its scheme: sanitizeUrl() prefixes bare
+        % addresses with mailto:, so the template must not add a second one.
+        \href{{{EMAIL_URL}}}{\raisebox{-0.2\height}\faEnvelope\  \underline{{{EMAIL_DISPLAY}}}} ~
         % LINKEDIN
         \href{{{LINKEDIN_URL}}}{\raisebox{-0.2\height}\faLinkedin\ \underline{{{LINKEDIN_DISPLAY}}}}  ~
         % GITHUB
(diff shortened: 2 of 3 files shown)


santifer:
Clean fix, @Scott-Emberson. The doubled mailto: and the OT1-unsafe glyphs are both real breakage in generated .tex, and routing them through the existing escape layer keeps one source of truth. Worth knowing (my verification, not a change request): escapeLatex has a second consumer in lib/latex-content.mjs applyPatches(), so this also reaches the bring-your-own-.tex flow — wider blast radius than the title suggests, and in the right direction. Users who copied templates/cv-template.tex keep the old behaviour until they run npm run update; templates/ is in SYSTEM_PATHS so that path works. Merged! 🚀

PR #2379, opened 2026-07-31
fix(providers): dead boards must throw, not read as "live but empty"

### LINKED ISSUE #2374: bug(providers): four providers swallow a first-request failure, so a dead board reads as "live but empty" and portal-health never escalates
Four providers swallow a first-request failure and return `[]`, so a down or blocked board is indistinguishable from a healthy board with zero matches:

- `providers/cryptocurrencyjobs.mjs` wraps its only fetch in `catch { return []; }`
- `providers/phenom.mjs` catches a page-1 failure with the mid-scan `break` and returns the empty buffer
- `providers/radancy.mjs` does the same in its HTML fallback walk
- `providers/successfactors.mjs` (CSB variant) returns `[]` when every locale fails

## Why it matters

scan.mjs distinguishes the two cases deliberately: a provider **throw** lands in `errors` and portal-health records a failure; an empty array lands in `emptyTargets` ("live but empty"). When the failure is swallowed, a dead endpoint keeps reporting healthy, the consecutive-failure escalation in portal-health never fires, and coverage decays with no signal. That defeats the reason portal-health exists.

meituan.mjs and tencent.mjs already implement the intended contract, with a comment spelling it out: track `succeededOnce`, throw when the very first request fails, and only keep partials once something has succeeded ("a keyword can legitimately match 0 jobs, so seen.size is not the signal").

## Fix

Adopt the same idiom in the four providers above: a failure before any successful request throws; a mid-scan failure still keeps the pages already collected. The cryptocurrencyjobs test that asserted the swallow ("returns [] when the feed fetch fails") gets updated to assert the throw, since scan.mjs catches per-target and a throw never aborts the scan. PR incoming.

Scott-Emberson:
Closes #2374.

cryptocurrencyjobs, phenom, radancy and successfactors (CSB) each swallowed a first-request failure into `[]`. scan.mjs reads that as "live but empty" and files the target under `emptyTargets`, so a dead endpoint never reaches the error path, portal-health keeps reporting it healthy, and the consecutive-failure escalation never fires.

All four now follow the `succeededOnce` idiom meituan and tencent already implement: a failure before any successful request throws, so scan records a real failure; a mid-scan failure still keeps the pages already collected. scan.mjs catches per-target, so the throw never aborts a run.

One existing assertion changes meaning: the cryptocurrencyjobs test that expected `[]` on a failed feed fetch now expects the throw. Each of the other three gains a first-request-failure case, and the meituan test that already encodes this contract still passes unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fetch failures are now reported instead of being incorrectly shown as empty job results.
  * Initial page and complete request failures are surfaced clearly, while partial results remain available when later pages fail.
  * Improved error handling across multiple job board integrations.

* **Tests**
  * Added regression coverage for network and API failures across supported job sources.

### DIFF
--- a/providers/cryptocurrencyjobs.mjs
+++ b/providers/cryptocurrencyjobs.mjs
@@ -105,14 +105,12 @@ export default {
   id: 'cryptocurrencyjobs',
 
   async fetch(_entry, ctx) {
-    let xml;
-    try {
-      // redirect:'error' — refuse server-side redirects (SSRF hardening, matches
-      // the other HTTP providers); the feed host is fixed so a redirect is a flag.
-      xml = await ctx.fetchText(FEED, { redirect: 'error' });
-    } catch {
-      return [];
-    }
+    // redirect:'error' — refuse server-side redirects (SSRF hardening, matches
+    // the other HTTP providers); the feed host is fixed so a redirect is a flag.
+    // A failed fetch must THROW, not return []: swallowing it makes a dead
+    // board read as "live but empty", so portal-health never escalates and
+    // coverage decays silently (same contract as meituan/tencent).
+    const xml = await ctx.fetchText(FEED, { redirect: 'error' });
     return parseCryptocurrencyJobsRss(xml);
   },
 };
--- a/providers/phenom.mjs
+++ b/providers/phenom.mjs
@@ -150,6 +150,10 @@ export default {
     const jobs = [];
     const seen = new Set();
     let total = null;
+    // A page-1 failure means the board is unreachable, not empty — it must
+    // THROW so scan/portal-health record a failure instead of "live but
+    // empty" (meituan/tencent idiom). Only a mid-scan failure keeps partials.
+    let succeededOnce = false;
 
     for (let page = 0; page < maxPages; page++) {
       if (page > 0) await wait(PAGE_DELAY_MS);
@@ -183,9 +187,11 @@ export default {
             locationData: {},
           }),
         });
-      } catch {
+      } catch (err) {
+        if (!succeededOnce) throw err;
         break; // keep jobs collected so far — a transient mid-scan failure shouldn't discard earlier pages
       }
+      succeededOnce = true;
       const { total: pageTotal, rows } = parseRefineSearch(json, cfg);
       if (total === null) total = pageTotal;
       if (rows.length === 0) break;
--- a/providers/radancy.mjs
+++ b/providers/radancy.mjs
@@ -293,9 +293,13 @@ export default {
         const first = await ctx.fetchJson(buildFragmentUrl(listUrl, 1), {
           headers: { accept: 'application/json', 'x-requested-with': 'XMLHttpRequest' },
         });
-        succeededOnce = true;
         const firstHtml = typeof first?.results === 'string' ? first.results : '';
         const firstRows = firstHtml ? parseResults(firstHtml, origin) : [];
+        // Proof of life only once the response also PARSED: a valid fragment
+        // with zero rows counts, but a response that crashes parsing leaves
+        // this false so a failing HTML fallback still reports the malformed
+        // initial response instead of returning [].
+        succeededOnce = true;
         if (firstRows.length) {
           const { totalResults, totalPages } = readFragmentTotals(firstHtml);
           // Bound by the server's own page count when it gives one; the local
--- a/providers/radancy.mjs
+++ b/providers/radancy.mjs
@@ -345,15 +345,22 @@ export default {
       }
     }
 
+    // A page-1 failure on the fallback transport (after the JSON transport also
+    // produced nothing) means the board is unreachable, not empty — THROW so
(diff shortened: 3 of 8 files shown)


santifer:
The direction is right, @Scott-Emberson — "live but empty" hiding an unreachable board is exactly the kind of false green portal-health exists to kill, and the cryptocurrencyjobs and phenom adoptions of the meituan idiom are faithful. Two of the four introduce the inverse bug though, and the test shape is why CI can't see it. Both fixes are one line.

**1. successfactors: a healthy RMK board with zero postings now reports `slug_gone`.**

The dispatcher runs RMK first and, when it legitimately returns zero jobs, probes `fetchCsb` — a CSB-only endpoint. Your own comment on that line says a genuinely-empty RMK board "costs one extra probe", i.e. the design *expects* that probe to fail harmlessly. This PR is what stops it being harmless: the probe now throws, and a healthy tenant with an empty board is recorded as a dead slug.

This is the path every real config takes — every documented SuccessFactors entry in `templates/portals.example.yml` is RMK, and `sfVariant` appears nowhere in the repo outside the provider and your new test.

And that's the second half of the problem: `tests/providers/successfactors.test.mjs` passes `sfVariant: 'csb'`, which routes straight into `fetchCsb` and never exercises the RMK-first dispatcher. The assertion itself is real, but it green-lights the path nobody runs while leaving the path everybody runs untested — which is how the suite stays at 2762/0 with this regression in it.

**Fix:** gate the throw on `sfVariant === 'csb'`, or make the CSB fallback non-throwing when it's the post-RMK probe. Plus a test that goes through the dispatcher without `sfVariant`.

**2. radancy: a successful request doesn't count as proof of life.**

`let succeededOnce = false` is declared *after* the JSON-fragment block, so a 200 from the fragment endpoint — definitive evidence the tenant is reachable — never sets it. If the fragment answers 200 with zero rows (a legitimately empty board) and the HTML fallback then 403s, the provider throws and reports "unreachable" for a tenant it just talked to successfully. That contradicts this PR's own contract wording about a failure *before any success*.

**Fix:** set `succeededOnce = true` when the fragment fetch resolves.

Two smaller notes, no action needed:

- `lastErr` keeps the *last* locale's error rather than the most severe, so a `de_DE` 500 followed by an `en_US` 404 is rethrown as the 404 and recorded as `slug_gone` instead of `server` — it degrades fidelity in exactly the portal-health column this PR exists to populate correctly.
- One of #2374's four claimed symptoms doesn't apply to cryptocurrencyjobs: it's documented under `job_boards:`, and `scan.mjs` only pushes non-board targets into `emptyTargets`, so the old swallow recorded "reachable", not "empty". The fix is still directionally correct; the stated mechanism just isn't the one at play there.

For transparency about the limits of my review: I did not probe a live SuccessFactors tenant, so the RMK conclusion rests on internal evidence — the CSB-only path shape, your "307 legacy redirect" comment, and the dispatcher comment about the extra probe. If you have a real RMK tenant to test against and it behaves differently, say so and I'll re-read.

Happy to merge once those two lines land.

Scott-Emberson:
Both landed in dc52939, plus a polish pass in 11c8daa from CodeRabbit's follow-up review.

successfactors: `fetchCsb` now takes a probe flag and the post-RMK fallback calls it with `probe: true`, so a healthy RMK tenant with zero postings whose CSB probe fails returns `[]` instead of throwing. Explicit `sfVariant: 'csb'` configs still throw as before. Added the dispatcher test you asked for: no `sfVariant`, RMK answers 200 with zero tiles, every CSB locale fails, result is empty with no throw. Both fetch() tests also count fetchJson invocations and assert at least one, so the probe test can't go green without the CSB path actually running.

radancy: `succeededOnce` is declared before the fragment block and set once `parseResults` produces the first rows. A valid zero-row fragment counts as proof of life, while a response that fetches fine but fails parsing leaves it false so the HTML fallback can still report a malformed board. Your fragment-200-then-403 case now returns empty instead of "unreachable"; test added.

On the lastErr note: I didn't build severity ranking, but the throw is at least deterministic now (`firstErr ??= err`), so the first locale failure is what surfaces rather than whichever locale happened to run last. If you'd like proper server-over-slug_gone precedence in the portal-health column I can do that as a follow-up.

Agreed on the cryptocurrencyjobs mechanism note, no change there.

santifer:
Both regressions closed exactly right, @Scott-Emberson. The probe:true parameter is cleaner than gating on sfVariant — it names the semantic (RMK already answered, so a failing CSB probe can't mean dead board) instead of special-casing a config value. The new dispatcher test with the anti-vacuous guard (asserting the jobs API was actually called, so a future routing change can't make it pass by accident) is the shape I wish more tests had. And radancy's refinement — only a WELL-FORMED fragment response counts as proof of life, so a malformed one still surfaces — is a distinction my RC didn't even ask for. firstErr over lastErr for deterministic error reporting closes the minor note too. Merged! 🚀

PR #2388, opened 2026-07-31
perf(fingerprint): popcount table + hoisted loop state in findCrossListings (#2381)

### LINKED ISSUE #2381: perf(fingerprint): findCrossListings is O(N x M) with per-pair BigInt parsing — 54s at 500 offers x 20k history rows
`findCrossListings` (fingerprint-core.mjs:122) compares every new offer against every recent history row, and each pair pays far more than a Hamming distance should cost.

`similarity()` (fingerprint-core.mjs:93) runs two regex tests, two `BigInt('0x' + hex)` parses, and then counts bits one at a time:

```js
let x = BigInt('0x' + a) ^ BigInt('0x' + b);
let dist = 0;
while (x) { dist += Number(x & 1n); x >>= 1n; }
```

That is up to 64 BigInt shift-and-mask operations per pair, and both operands get re-parsed from hex on every comparison even though the same fingerprint is reused across the whole inner loop. `companyKey(row.company)` is also recomputed per pair rather than once per row.

## Measured

Synthetic history rows with fingerprints, timed against the real module:

| Scale | Current |
|---|---|
| 500 offers x 5k rows (random fingerprints) | 13,688 ms |
| 500 offers x 5k rows (realistic shared boilerplate) | 5,719 ms |
| 500 offers x 20k rows | 53,616 ms |
| 500 offers x 50k rows | ~83 s (projected from 166 ms for a single offer) |

scan-history.tsv is append-only and grows for the life of an install, so this is the term that gets worse over time rather than the one-off costs around it.

## Fix

Hoist each fingerprint to two 32-bit ints and each row's company key once, then take the Hamming distance through a 64 KB `Uint16` popcount table. A drop-in replacement measured:

| Scale | Current | Popcount table | Speedup |
|---|---|---|---|
| 500 x 5k | 13,688 ms | 28.1 ms | 488x |
| 500 x 20k | 53,616 ms | 110.8 ms | 484x |

Distances come out identical, so this is not an approximation and the threshold semantics are unchanged. The only care needed is converting `threshold` (a 0..1 score) to a maximum bit distance without an off-by-one, which a test can pin.

An LSH banding index would take this from O(N x M) to roughly O(N + M), but the popcount fix alone already drops it out of the profile, so it seems like the wrong complexity to take on first.

Related but separate: `fingerprintText` (fingerprint-core.mjs:69) allocates a fresh `createHash('sha1')` per 3-token shingle, about 418 shingles per JD, measured at 512 ms for 500 JDs. Replacing the inner 64-iteration bit-accumulate loop with 8 per-byte table lookups is output-preserving and takes roughly 20% off. Swapping sha1 for a cheaper hash would be ~5x but changes every fingerprint value, so stored history would stop matching new scans without a version column. Worth a separate issue if there's appetite.

Happy to send a PR for the popcount change with a test asserting the distances match the current implementation across a fixture set.

Scott-Emberson:
Closes #2381

## What was slow

`similarity()` parsed both 16-hex-char fingerprints into BigInts on every call and counted set bits one at a time (`while (x) { dist += Number(x & 1n); x >>= 1n; }`). `findCrossListings()` called it once per (offer, recent history row) pair, and recomputed `companyKey(row.company)` on every pair as well. At 500 offers against 5k history rows that works out to 2.5M BigInt allocations and 2.5M redundant string normalizations, none of which depend on the pair.

## What changed

Per-fingerprint work is hoisted out of the inner loop. Each fingerprint is split once into two 32-bit halves, and each history row's `companyKey` plus its halves are computed once while the recent-window list is built. Hamming distance now comes from a 64 KB `Uint8Array` popcount table indexed 16 bits at a time.

The 0..1 `threshold` is converted once into a maximum bit distance, so the inner loop compares integers and never touches floating point. That conversion is the part most likely to hide an off-by-one, so it is not derived by rounding `64 * (1 - threshold)`. Instead it searches for the largest `d` satisfying the same `1 - d / 64 >= threshold` expression the per-pair path used to evaluate. For integer `d`, `d / 64` is exact in binary floating point, so the search reproduces the old accept/reject decision at every threshold, including thresholds landing exactly on a bit boundary where a `floor()` or `ceil()` derivation is one off in one direction.

`similarity(a, b)` keeps its signature, its 0..1 return value and its regex guard, so malformed or empty input still yields 0. The malformed-fingerprint path inside `findCrossListings` still scores 0 rather than skipping, because a caller passing `threshold <= 0` legitimately matches on a score of 0.

## Measured on this machine (Windows 11, Node 22.23.1)

| workload | before | after | |
|---|---|---|---|
| 500 offers x 5k rows (2.5M pairs) | 4370.9 ms | 16.4 ms | 266x |
| 500 offers x 20k rows (10M pairs) | 18028.8 ms | 61.9 ms | 291x |

Both runs returned an identical match list (25 and 121 matches respectively), compared as an ordered sequence of offer URL, row URL and score.

## Equivalence, not approximation

The distances are unchanged, and the new tests establish that rather than assuming it. `tests/fingerprint-core.test.mjs` inlines a verbatim copy of the previous BigInt implementation of both `similarity()` and `findCrossListings()` and runs old against new over:

- 4000 pairs, half uniformly random and half near-duplicates 0 to 12 bits apart, asserting strict `===` on the score rather than an epsilon, since both paths derive `1 - d / 64` from an integer `d` and any difference at all means the distances diverged
- every Hamming distance from 0 to 64, from two different base fingerprints
- all 65 exact bit-boundary thresholds `1 - k/64`, each asserting that `k` differing bits still match and `k+1` does not
- a 40 offer by 400 row corpus across 6 thresholds and 2 window sizes, mixing empty and malformed fingerprints, companies that collide only after normalization (`Acme Ltd.` vs `acme-ltd`), same-URL pairs, unparseable dates and out-of-window rows

The corpus comparison covers order as well as content, because callers read `matches[0]` as the best candidate. A separate assertion checks the corpus actually produces matches (153 at the default threshold), so a filter bug returning nothing could not pass by comparing empty against empty.

I sanity-checked the tests by mutating the implementation: replacing the threshold conversion with `Math.ceil(64 * (1 - threshold))` fails 2 assertions, with `Math.floor(...) - 1` fails 3, dropping the zero-score branch fails 2, and removing the table index mask fails 3. The seeded PRNG means any failure reproduces from the log.

## Validation

`node test-all.mjs`: 2756 passed, 1 failed, 2 warnings. 9 of those passes are new.

The single failure is `analyze() appDateSource end-to-end check crashed: EPERM: operation not permitted, symlink`. It is environmental rather than a regression: the check calls `symlinkSync()` to stage a temp directory, which Windows refuses without elevation, and it exercises `followup-cadence.mjs` and `tracker-parse.mjs`, neither of which this branch touches. It is the known baseline failure on this box and should not appear on Linux CI.

### DIFF
--- a/fingerprint-core.mjs
+++ b/fingerprint-core.mjs
@@ -82,6 +82,52 @@ export function fingerprintText(text) {
   return hash.toString(16).padStart(16, '0');
 }
 
+/** A fingerprint is exactly 16 lowercase hex chars; anything else never
+ * matches. Hoisted to module scope so the hot path reuses one compiled regex
+ * rather than re-evaluating a literal on every call. */
+const FINGERPRINT_RE = /^[0-9a-f]{16}$/;
+
+/** Set-bit count for every 16-bit value, built once at module load (64 KB).
+ * A 64-bit Hamming distance then costs four table lookups instead of the 64
+ * BigInt shift/mask iterations this module used to run per comparison, and
+ * findCrossListings compares offers × history rows, so that per-pair cost is
+ * multiplied by the product of both list lengths (#2381). */
+const POPCOUNT16 = (() => {
+  const table = new Uint8Array(1 << 16);
+  for (let i = 1; i < table.length; i++) table[i] = table[i >> 1] + (i & 1);
+  return table;
+})();
+
+/**
+ * Number of set bits in a 32-bit word.
+ *
+ * Sign is irrelevant: both masks yield a non-negative 0..65535 index, so a
+ * negative int32 (which `^` produces whenever the top bit is set) indexes the
+ * table just as correctly as a positive one.
+ *
+ * @param {number} x - Any 32-bit integer.
+ * @returns {number} 0..32.
+ */
+function popcount32(x) {
+  return POPCOUNT16[x & 0xffff] + POPCOUNT16[(x >>> 16) & 0xffff];
+}
+
+/**
+ * Split a fingerprint that has already passed FINGERPRINT_RE into its two
+ * 32-bit halves.
+ *
+ * JS bitwise operators are 32-bit, so the 64-bit value is carried as a pair.
+ * A single Number would lose precision above 2^53, and a BigInt allocates on
+ * every operation — the exact cost this rewrite removes.
+ *
+ * @param {string} fp - Validated 16-hex-char fingerprint.
+ * @returns {{hi: number, lo: number}} Upper and lower 32 bits, as int32.
+ */
+function splitFingerprint(fp) {
+  const s = String(fp);
+  return { hi: parseInt(s.slice(0, 8), 16) | 0, lo: parseInt(s.slice(8, 16), 16) | 0 };
+}
+
 /**
  * Similarity of two fingerprints: 1 − hammingDistance/64. Empty or malformed
  * fingerprints never match (returns 0).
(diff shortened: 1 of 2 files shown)


santifer:
Solid optimisation, @Scott-Emberson — and the equivalence argument holds up: I had it mutation-tested independently and 9 of 10 realistic mutants die against your assertions. One gap for a follow-up if you want it: replacing maxDistanceFor() with Math.floor(64 * (1 - threshold)) survives every current assertion, so that boundary isn't pinned. Also noting that findCrossListings no longer calls the exported similarity() — the distance is inlined, so the two paths can now drift independently. Neither blocks this. Merged! 🚀

PR #2389, opened 2026-07-31
perf(detect-reposts): bucket titles instead of comparing every pair

### LINKED ISSUE #2383: perf(detect-reposts): title grouping is O(N^2) when a company has many distinct titles — 45s on a 4k-row scan-history
`detectRepostsInGroup` (detect-reposts.mjs:165) groups a company's rows by title with a nested loop:

```js
for (const row of rows) {
  if (used.has(row)) continue;
  ...
  for (const other of rows) {
    if (used.has(other)) continue;
    if (row.title.toLowerCase() === other.title.toLowerCase() || roleFuzzyMatch(row.title, other.title)) {
```

The `used` Set prunes the inner loop whenever titles collapse into groups, which is why this has not bitten anyone yet. But when one company has many *distinct* titles nothing is ever pruned, and every pair pays a `roleFuzzyMatch` plus a freshly allocated `row.title.toLowerCase()`.

## Measured

Synthetic scan-history, one company, distinct titles:

| rows | detectReposts |
|---|---|
| 1,000 | 28 ms |
| 2,000 | 1,971 ms |
| 4,000 | 44,980 ms |
| 8,000 | 112,985 ms |

Through the real CLI on a 4,000-row `data/scan-history.tsv`:

- `node detect-reposts.mjs --summary` → **45,593 ms**
- `node company-history.mjs --summary` → **25,484 ms** (it imports `detectReposts` at detect-reposts.mjs:49, used at :228)

## Why this is reachable

`scan-history.tsv` is append-only, one row per scanned posting. A single large employer with a wide job board accumulates thousands of distinct titles over months of scanning, which is exactly the shape that defeats the `used` pruning. Both `detect-reposts` and `company-history` degrade together, since the latter delegates to the former.

## Fix

Bucket by a precomputed lowercased title key in a Map first. Exact matches collapse in O(N), which is the overwhelming majority of real repost pairs. Then run `roleFuzzyMatch` only across the remaining distinct keys, which is a much smaller set. Hoist the `toLowerCase()` into a single pass instead of re-allocating it per comparison.

The clustering semantics need to come out identical, and `tests/detect-reposts.test.mjs` is substantial (49 KB), so the existing coverage should catch a behavioural drift. Happy to send a PR with a large-input timing assertion added alongside.

Scott-Emberson:
Closes #2383

## What was slow

`detectRepostsInGroup` grouped one company's rows with a nested loop over rows. Every pair paid a full `roleFuzzyMatch`, which re-tokenizes both strings on each call, and the seed's `row.title.toLowerCase()` was re-allocated inside the inner loop. The `used` set only prunes when titles collapse into groups, so a company with many distinct titles pruned nothing.

That shape is reachable rather than theoretical. `data/scan-history.tsv` is append-only with one row per scanned posting, so a large employer accumulates thousands of distinct titles over time.

## What changed

Grouping now runs in three passes instead of one nested loop.

Rows are bucketed by their lowercased title in a Map. Every row in a bucket matches every other row in it through the exact-title arm of the old condition, so a bucket is atomic: a seed takes all of it or none of it. Exact reposts, which are most real reposts, collapse in linear time with no fuzzy calls, and `toLowerCase()` runs once per row instead of once per comparison.

Fuzzy matching then runs over distinct buckets only, gated by an inverted index over non-baseline tokens. `roleFuzzyMatch` can only return true for two non-identical titles when their deduped token sets share at least two tokens, at least one of which is not in `BASELINE_TOKENS`, and their Jaccard ratio is 0.6 or higher. Those are necessary conditions, and all of them are computed exactly from the precomputed token sets, so any pair the gate drops is one `roleFuzzyMatch` would have rejected anyway. Every pair that survives the gate is still decided by `roleFuzzyMatch` itself. The gate filters calls, not verdicts.

Group order and row order within each group are preserved exactly. Both are load-bearing: the date sort in `detectRepostsInGroup` uses a comparator that returns 1 rather than 0 for equal dates, so same-date rows keep their order only if the group arrives in input order, and `buildRepostCluster` reads `clusterRows[0].company`. Groups are emitted in seed order, and their rows are re-sorted by original array position, which is what the nested loop produced: the seed was always the first not-yet-used row, and the inner loop appended the rest in array order.

## Measured on this machine

Synthetic 4k and 8k row scan histories for one company, titles drawn from a shared vocabulary of levels, domains, and teams so they are not artificially easy to bucket. Node 22.23.1, Windows 11.

| rows | before | after |
|------|--------|-------|
| 4000 (`--summary`) | 16,637 ms | 174 ms |
| 8000 (JSON) | 89,396 ms | 339 ms |

The JSON output of the old and new code is byte-identical on both corpora.

## Why the output is provably unchanged

`detect-reposts.test.mjs` now carries a frozen copy of the pre-change algorithm, nested loop and all, and compares full cluster output against it (not just cluster counts, so a changed cluster order or a changed appearance order inside a cluster fails too). The corpora cover 120 distinct titles at one company, 60 exact duplicates spread over a year, fuzzy word-order variants, near-miss titles that must stay separate, mixed-case duplicates of one title, same-date rows, singleton companies and singleton titles, and a seeded 300-row mixed corpus with duplicate URLs and non-added statuses run at three window sizes. Two assertions confirm the corpora actually produce clusters, so an all-empty comparison cannot pass by accident.

I checked the test has teeth by mutating the Jaccard gate from 0.6 to 0.9 in the implementation; the equivalence comparison failed as expected, then passed again once reverted.

Section 11 gains a 4025-row guard: 4000 distinct titles at one company plus 25 planted exact reposts, asserting both the exact cluster count and a 10s budget. The old code took tens of seconds on this shape and the new code takes a fraction of a second, so the budget has an order of magnitude of headroom and is not a wall-clock flake, while still failing if the quadratic loop returns.

If `detect-reposts.mjs` ever changes what it computes rather than how fast it computes it, the frozen copy has to be updated in the same commit. That is intentional. A silent divergence should fail there.

## Validation

`node test-all.mjs`: 2747 passed, 1 failed. The one failure is pre-existing on this branch point and unrelated, `analyze() appDateSource end-to-end check crashed: EPERM: operation not permitted, symlink`, which is a Windows symlink permission issue in a test that builds a temp worktree.

`node detect-reposts.mjs --self-test`: 8 passed, 0 failed.
`node detect-reposts.test.mjs`: 227 passed, 0 failed.
`node company-history.test.mjs`: 86 passed, 0 failed (it imports `detectReposts`).

## Summary by CodeRabbit

* **Performance**
  * Improved repost detection efficiency, especially for larger datasets.
  * Potential matches are filtered more effectively while preserving grouping behavior and input order.

* **Bug Fixes**
  * Improved handling of duplicate, case-variant, fuzzy, near-match, same-date, and singleton entries.

* **Tests**
  * Added regression coverage across varied repost scenarios.
  * Added a large-scale benchmark covering 4,025 entries and 25 known repost clusters.

### DIFF
--- a/detect-reposts.mjs
+++ b/detect-reposts.mjs
@@ -25,7 +25,7 @@ import { readFileSync, existsSync } from 'fs';
 import { join, dirname } from 'path';
 import { fileURLToPath, pathToFileURL } from 'url';
 
-import { roleFuzzyMatch } from './role-matcher.mjs';
+import { roleFuzzyMatch, roleTokens, BASELINE_TOKENS } from './role-matcher.mjs';
 import { normalizeCompanyName } from './invite-match.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
@@ -164,22 +164,7 @@ export function detectReposts(rows, windowDays = DEFAULT_WINDOW_DAYS) {
 // approach prevents non-matching roles (e.g. a Product Manager between two
 // Backend Engineer postings) from breaking a valid repost cluster.
 function detectRepostsInGroup(rows, windowDays) {
-  const titleGroups = [];
-  const used = new Set();
-
-  for (const row of rows) {
-    if (used.has(row)) continue;
-    const group = [row];
-    used.add(row);
-    for (const other of rows) {
-      if (used.has(other)) continue;
-      if (row.title.toLowerCase() === other.title.toLowerCase() || roleFuzzyMatch(row.title, other.title)) {
-        group.push(other);
-        used.add(other);
-      }
-    }
-    titleGroups.push(group);
-  }
+  const titleGroups = groupRowsByTitle(rows);
 
   const results = [];
   for (const group of titleGroups) {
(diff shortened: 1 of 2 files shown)


santifer:
Good speedup, @Scott-Emberson, and the frozen-legacy-algorithm equivalence corpus is the right way to argue it. Two notes from independent mutation testing, neither blocking: the corpus catches a Jaccard threshold shift but has a hole worth a follow-up test, and the new bucket key uses title.toLowerCase() without .trim() while roleFuzzyMatch's fast path trims — so titles differing only in surrounding whitespace land in different buckets. Worth a comment at the point of use so the coupling isn't invisible. Merged! 🚀

PR #2390, opened 2026-07-31
perf(tests): fix the dead SSRF DNS stub and cut three fixed costs from test-all

### LINKED ISSUE #2386: bug(tests): the SSRF DNS mock stubs dnsModule.default but liveness-browser calls the namespace — test passes by accident and costs 12s per run
The SSRF liveness test mocks the wrong object, so the mock never applies and the test asserts the right verdict for the wrong reason. It also costs 12.4 seconds of every suite run.

## The mismatch

test-all.mjs:766-778 stubs the **default export**:

```js
mock.method(dnsModule.default, 'resolve4', ...);
mock.method(dnsModule.default, 'resolve6', ...);
mock.method(dnsModule.default, 'lookup',   ...);
```

liveness-browser.mjs:136-141 imports the module and calls the **namespace bindings**:

```js
const dns = await import('dns/promises');
const [ipv4, ipv6, lookupList] = await Promise.all([
  dns.resolve4(hostname).catch(() => []),
  dns.resolve6(hostname).catch(() => []),
  dns.lookup(hostname, { all: true }).catch(() => [])
]);
```

`dns.resolve4` and `dnsModule.default.resolve4` are different bindings, so the stub is never consulted. The code under test goes to the real OS resolver for `ssrf-blocked-host.local`.

## Why it still passes

The mock was written to return `127.0.0.1` so the SSRF guard would reject a loopback address. In reality the OS resolves nothing for that `.local` name, the address list comes back empty, and the guard returns `blocked_host` anyway. Same verdict, completely different code path, and the loopback-rejection branch the test was written to cover is never actually exercised.

## The 12 seconds

Decomposed in isolation, `checkUrlLiveness(mock, ...)` takes 12,358 ms:

```
dns/promises resolve4('ssrf-blocked-host.local')  ->      1 ms
dns/promises resolve6(...)                        ->      0 ms
dns/promises lookup(...)                          -> 12,324 ms   <-- mDNS/LLMNR
```

`Promise.all` waits on all three, so a Windows mDNS/LLMNR resolution of a `.local` name sets the floor for the whole suite section. That is 11.6% of a 106.7 s run spent waiting on a name that was supposed to be mocked.

## Fix

Mock the namespace object returned by `await import('dns/promises')`, or add a resolver seam to `resolveHostAddresses` and inject it. Either way the test starts exercising the loopback-rejection branch it claims to cover, and the 12 s disappears.

One thing to watch: `dnsCache` at liveness-browser.mjs:128 memoizes per hostname, so a test that resolves the same host earlier in the run could mask the assertion once the mock actually works.

### LINKED ISSUE #2387: perf(tests): 74% of test-all.mjs wall clock is child-process spawns — ~33s recoverable at no coverage cost
`node test-all.mjs` takes 106.7 s here (2747 passed, 1 failed, 2 warnings). Instrumenting it with a `child_process` interception loader and a per-log-line timeline:

```
child_process: execFileSync 413 (77,563ms) + spawnSync 13 (882ms) + execSync 2 (149ms)
             = 428 spawns / 78,594 ms = 73.7% of wall clock
non-spawn time: 28,083 ms
```

392 of the 428 spawns finish in under 250 ms (avg 57 ms), so most of that is cold start, not work. Bare Node cold start on this machine is 37 ms; `node --check test-all.mjs` is 54 ms.

Two things worth ruling out first, since they look like the obvious suspects and are not:

- **The discovered `tests/*.test.mjs` suites are fine.** They run in-process via `await import` and sequentially (test-all.mjs:105-115): 102 files, 10,171 ms, 9.5% of wall. Parallelising would need worker processes plus counter aggregation, since `passed`/`failed`/`warnings` are module-level globals in tests/helpers.mjs:14-16, and several suites (portal-health-lock, tracker-busy-timeout, merge-tracker concurrency) deliberately contend on shared paths. Max 10 s for real breakage risk.
- **No `npm install` runs anywhere**, and the `node_modules` "symlink test" (test-all.mjs:6206) creates a junction rather than a copy, so it costs about 0 ms. On an unprivileged Windows box it EPERMs, which is the one pre-existing failure.

### 1. `fix-slugs.mjs --dry-run` burns its full 15 s network timeout every run (15,016 ms, 14.1%)

test-all.mjs:4300 spawns the real CLI with `timeout: 15000`, and fix-slugs.mjs:78-87 probes live `boards-api.greenhouse.io`, `jobs.ashbyhq.com` and `jobs.lever.co`. The `catch {}` at :4306 says "Network is reachable-or-not in CI; either way, no write should occur", so the test deliberately eats the timeout.

The assertion is only that a dry run wrote nothing to disk, which is orthogonal to network reachability. Dropping the timeout to ~2 s, or pointing the child at a stub, saves 13-15 s with no coverage loss.

### 2. Section 2 copies `web/node_modules` (~9,600 ms warm, 22,288 ms cold)

The exclusion at test-all.mjs:221 is `if (dirname(src) === ROOT && exclude.includes(name)) return;` — basename matching at the repo root only. The comment explains why: nested fixture dirs like `test-fixtures/upgrade/state-*/data` must still be copied. But that also means nested `web/node_modules` (16,369 files, gitignored per web/.gitignore:2) and `graphify-out/` get copied byte for byte.

```
copyDirSync TOTAL: 14,215 ms | 17,712 files | 481.6 MB   (cold)
  web -> 13,779 ms / 16,369 files ; graphify-out -> 168 ms / 548 files
rmSync: 8,073 ms
```

Excluding `web/` alone drops the copy to ~1,343 files. The fix needs to keep the existing nested-fixture behaviour: match `node_modules` and `.git` at any depth while leaving `data/` and `reports/` matched at root only.

Worth noting this is a local-dev win, not a CI one: .github/workflows/test.yml:38-39 installs at root only and never installs `web/`, so `web/node_modules` does not exist on CI.

### 3. 111 sequential `node --check` spawns (5,455 ms → ~700 ms)

test-all.mjs:137-144 loops `run(NODE, ['--check', f])` one file at a time. 112 spawns, 5,556 ms, 50 ms average, which is essentially pure cold start. Pooling the same 111 files:

| strategy | ms |
|---|---|
| sequential | 3,974 |
| pool(4) | 1,067 |
| pool(8) | 694 |
| pool(16) | 589 |

`--check` is pure and touches no shared state, so the only requirement is collecting results and emitting `pass`/`fail` in the original order to keep output byte-identical.

### 4. `archive-posting` fetches a live Greenhouse job and renders it in a real browser (~8,700 ms, 8.2%)

test-all.mjs:4915-4930 fetches a real posting from the Greenhouse API, then renders it with `timeout: 60000`. One spawn alone is 6,719 ms, plus a 1,958 ms gap for the live fetch. Beyond the cost, this makes the suite non-deterministic — it depends on a third party's job board being up.

The code already degrades to `warn(...)` when offline, so the skip path exists. Gating behind `CAREER_OPS_LIVE_TESTS=1` or a nightly job, with a local-fixture render on the PR path, is the trade: genuine end-to-end coverage on every PR in exchange for ~8.7 s and determinism.

## Separately: `--quick` does almost nothing

`QUICK` is referenced twice, and the only gate is `if (!QUICK)` at test-all.mjs:864 around the Go dashboard build. That is 2.2 s out of 107 s. CI runs `--quick` on the assumption it is a fast path, which it is not — worth either making it meaningful or dropping the flag.

## Also measured, lower payoff

- 22 `bash.exe` spawns totalling 6,744 ms (batch-runner sections; Git Bash startup is expensive on Windows). Batching the spend_tier assertions into one runner invocation recovers ~4 s.
- Section 2's 45 script spawns total 21,954 ms, dominated by five heavy children (set-status-tests 5,140 ms, tracker-writer-lock-tests 3,561 ms, followup-seed-tests 2,735 ms, tracker-columns-tests 2,415 ms, company-history.test 1,260 ms). Mostly irreducible: they spawn their own grandchildren and tracker-writer-lock-tests has genuine `sleep()` calls. The other ~40 `--self-test` scripts are read-only and pool-safe in principle, but they share one `scriptTmp/data/applications.md`, so the tracker-mutating ones would need per-script copies first.
- `go build` 2,195 ms and `node --test test/pipeline-lock.test.mjs` 2,337 ms — isolated, hard to reduce.
- Uncached `readFileSync` across 189 read sites is noise: non-spawn time is 28.1 s total and DNS, the repo copy, and the discovered suites account for essentially all of it.

Items 1, 2 and 3 are roughly 33 s combined, low risk, and cost no coverage. Happy to send PRs for those. Item 4 is a policy call about live-network tests, so I'd rather ask first.

There is a related 12.4 s item I filed separately, since it is a correctness bug rather than a tuning question.

Scott-Emberson:
Closes #2386
Closes #2387

Both issues change `test-all.mjs`, so they share one branch as two separate commits. Splitting them across two PRs would guarantee a conflict on that file for no benefit.

## #2386: the SSRF DNS stub never applied

The test stubbed `resolve4`, `resolve6` and `lookup` on the `dns/promises` module object. `liveness-browser.mjs` imports that module dynamically and calls the namespace bindings, which are immutable, so the stub never took effect. The test passed by a route it was not written to take: real DNS returned nothing for `ssrf-blocked-host.local`, the resolver threw on the empty address list, and the guard reported `blocked_host` anyway. The loopback rejection the test exists to cover never ran. On Windows the doomed lookup also cost 12.3 s of mDNS and LLMNR timeouts on every suite run.

I checked whether the module object could be patched at all before touching production code:

```js
mock.method(dnsModule.default, 'resolve4', () => Promise.resolve(['127.0.0.1']));
await ns.resolve4('ssrf-blocked-host.local');   // still hits real DNS: ECONNREFUSED
```

So the fix is a seam. `setHostResolver(fn)` in `liveness-browser.mjs` swaps the resolver, returns a restore function, and clears the memo cache in both directions, since `dnsCache` would otherwise serve a verdict computed under one resolver to the next and mask the assertion. The test injects a resolver that answers `127.0.0.1` for the synthetic host and a public address for everything else, then restores it in a `finally` so a stub can never leak into a later suite in the same process.

Evidence that the loopback branch now runs, from the suite log:

```
before: ... - DNS resolution returned no addresses for ssrf-blocked-host.local
after:  ... - Access denied: Egress guard blocked private target IP 127.0.0.1
```

The assertion now requires the blocked reason to name `127.0.0.1`, not just `code === 'blocked_host'`. That code alone cannot tell "the guard rejected loopback" from "the host resolved to nothing", which is the ambiguity that let a dead stub sit there unnoticed.

## #2387: items a, b and c

**a. `fix-slugs` dry-run timeout, 15 s to 2 s.** The child probes live Greenhouse, Ashby and Lever endpoints, so on a connected machine it never finishes. An unbounded run of the same command takes 45.5 s here, so the 15 s timeout was always reached and the `catch` always swallowed the kill. The assertion is that a dry run wrote nothing to disk, which holds at any timeout, and it is unchanged.

**b. `node --check` through a bounded pool of 8.** `--check` parses one file and exits: no user code, no shared state, no dependency on any other file. Results are collected by index and printed afterwards in the original readdir order, so the log stays byte identical no matter which child finishes first. Over the same 111 files on this machine: 4252 ms sequential, 726 ms pooled.

**c. Section 2's throwaway repo copy now skips `node_modules` and `.git` at any depth.** The root-only rule is deliberate and stays for every other entry, because nested fixture dirs such as `test-fixtures/upgrade/state-*/data` and `.../reports` must still be copied (`seed-fixture.mjs --self-test` reads them from inside the copy). Only those two names move into an any-depth set. A checkout that has installed the web app's dependencies carries a `web/node_modules` of about 430 MB that was being copied in full on every run. Measured against such a checkout, warm cache, two rounds each: 5526 ms and 5743 ms before, 452 ms and 475 ms after.

## Measured on this machine

Windows 11, Node 22.23, worktree off `upstream/main` at 93b1cbb, full `node test-all.mjs` with no flags.

| run | wall clock | passed | failed | warnings |
|---|---|---|---|---|
| baseline | 100.8 s | 2747 | 1 | 2 |
| after #2386 | 84.4 s | 2747 | 1 | 2 |
| after #2387 | 60.1 s | 2747 | 1 | 2 |

The pass count is identical at every step, so nothing dropped out of discovery or ordering. Diffing the baseline log against the final one shows only the intended change to the SSRF reason string, plus per-run noise in temp dir names and a few timing values. Both logs are 3206 lines.

The single failure is pre-existing on all three runs and environmental, not something this PR introduces or fixes:

```
analyze() appDateSource end-to-end check crashed: EPERM: operation not permitted, symlink
```

That is an unprivileged Windows user without symlink rights.

One caveat on item c: this worktree has no `web/node_modules`, so its saving is not in the table above. It was measured separately against a checkout that does have one, as described.

## Deliberately not in this PR

Item 4 of #2387, the live Greenhouse archive-posting test, and the `--quick` flag question. Both are policy calls about what the suite may depend on and what a fast run should skip, so they are yours to decide rather than mine to assume.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DNS resolution handling for more reliable network protection.
  * Strengthened SSRF redirect checks, including detection of private resolved addresses.
  * Preserved legitimate subresource requests during security validation.

* **Tests**
  * Expanded syntax and security test coverage.
  * Improved test execution efficiency with bounded parallel processing.
  * Enhanced temporary test isolation by excluding dependency and version-control directories.

### DIFF
--- a/liveness-browser.mjs
+++ b/liveness-browser.mjs
@@ -127,24 +127,58 @@ export function rejectPrivateOrInvalid(url) {
 
 const dnsCache = new Map();
 
+// Real DNS: resolve4 + resolve6 + lookup, each tolerant of its own failure, so a
+// host that only answers on one of the three still yields an address list.
+async function resolveViaDns(hostname) {
+  const dns = await import('dns/promises');
+  const [ipv4, ipv6, lookupList] = await Promise.all([
+    dns.resolve4(hostname).catch(() => []),
+    dns.resolve6(hostname).catch(() => []),
+    dns.lookup(hostname, { all: true }).catch(() => [])
+  ]);
+  return Array.from(new Set([
+    ...ipv4,
+    ...ipv6,
+    ...lookupList.map(item => item.address)
+  ]));
+}
+
+let hostResolver = resolveViaDns;
+
+/**
+ * Swap the resolver the egress guard uses, returning a restore function.
+ *
+ * `dns/promises` is imported dynamically and the guard calls the ESM namespace
+ * bindings, which are immutable — monkey-patching the module object has no
+ * effect on them (#2386). Without this seam a test can only ever reach the
+ * "host resolved to nothing" branch: the real resolver returns an empty list
+ * for the synthetic hostname, the guard blocks on that, and the loopback
+ * rejection the test exists to cover never runs. The memo cache is cleared on
+ * every swap, in both directions, so a verdict computed under one resolver can
+ * never be served to the next.
+ *
+ * @param {((hostname: string) => Promise<string[]>)|null} resolver - Resolver to
+ *   install, or null to restore the real DNS one.
+ * @returns {() => void} Restores the resolver in place before this call.
+ */
+export function setHostResolver(resolver) {
+  const previous = hostResolver;
+  hostResolver = resolver ?? resolveViaDns;
+  dnsCache.clear();
+  return () => {
+    hostResolver = previous;
+    dnsCache.clear();
+  };
+}
+
 async function resolveDnsCached(hostname) {
   if (dnsCache.has(hostname)) {
     const cached = dnsCache.get(hostname);
     if (cached instanceof Error) throw cached;
     return cached;
   }
-  const dns = await import('dns/promises');
   try {
-    const [ipv4, ipv6, lookupList] = await Promise.all([
-      dns.resolve4(hostname).catch(() => []),
-      dns.resolve6(hostname).catch(() => []),
-      dns.lookup(hostname, { all: true }).catch(() => [])
-    ]);
-    const addresses = Array.from(new Set([
-      ...ipv4,
-      ...ipv6,
-      ...lookupList.map(item => item.address)
-    ]));
+    const addresses = await hostResolver(hostname);
     if (addresses.length === 0) {
       throw new Error(`DNS resolution returned no addresses for ${hostname}`);
     }
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -145,14 +145,32 @@ const mjsFiles = readdirSync(ROOT).filter(f => f.endsWith('.mjs'));
 const SYNTAX_POOL_SIZE = 8;
 const execFileAsync = promisify(execFile);
 const syntaxOk = new Array(mjsFiles.length);
+const syntaxDetail = new Array(mjsFiles.length);
 let nextSyntaxIdx = 0;
+
+// A bare catch reports a child killed on timeout, or one that never spawned, as
+// "has syntax errors" — sending the reader hunting for a parse error that does
+// not exist. Keep enough of the child's own diagnosis to tell those apart.
+const describeCheckFailure = (err) => {
+  if (err?.killed || err?.signal === 'SIGTERM' || err?.code === 'ETIMEDOUT') {
+    return `node --check timed out after 30000ms${err.signal ? ` (signal ${err.signal})` : ''}`;
+  }
+  const stderr = String(err?.stderr ?? '').trim();
+  if (!stderr) return `no stderr (exit ${err?.code ?? 'unknown'})`;
+  const clipped = stderr.length > 2000
+    ? `${stderr.slice(0, 2000)}\n    ... (${stderr.length - 2000} more chars)`
+    : stderr;
+  return clipped.replace(/\n/g, '\n    ');
+};
+
 const syntaxWorker = async () => {
   for (let i = nextSyntaxIdx++; i < mjsFiles.length; i = nextSyntaxIdx++) {
     try {
       await execFileAsync(NODE, ['--check', mjsFiles[i]], { cwd: ROOT, timeout: 30000 });
       syntaxOk[i] = true;
-    } catch {
+    } catch (err) {
       syntaxOk[i] = false;
+      syntaxDetail[i] = describeCheckFailure(err);
     }
   }
 };
@@ -163,7 +181,7 @@ mjsFiles.forEach((f, i) => {
   if (syntaxOk[i]) {
     pass(`${f} syntax OK`);
   } else {
-    fail(`${f} has syntax errors`);
+    fail(`${f} has syntax errors\n    ${syntaxDetail[i] ?? 'no diagnostic captured'}`);
   }
 });
 
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -24,10 +24,11 @@
  */
 
 
-import { execSync, execFileSync, spawn, spawnSync } from 'child_process';
+import { execSync, execFile, execFileSync, spawn, spawnSync } from 'child_process';
 import { readFileSync, existsSync, readdirSync, mkdtempSync, mkdirSync, writeFileSync, rmSync, statSync, unlinkSync, realpathSync, symlinkSync, copyFileSync } from 'fs';
 import { join, dirname, basename, delimiter } from 'path';
 import { tmpdir } from 'os';
+import { promisify } from 'util';
 import { fileURLToPath, pathToFileURL } from 'url';
 import yaml from 'js-yaml';
 import { pass, fail, warn, run, formatRunFailure, fileExists, finish, ROOT, QUICK, NODE, getBash, toBashPath } from './tests/helpers.mjs';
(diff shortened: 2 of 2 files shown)


santifer:
The dead stub is the important half here, @Scott-Emberson: because ESM namespace bindings are immutable, the old test could only ever reach the "host resolved to nothing" branch — the loopback rejection it existed to cover never ran. That's a blind guard in the one module whose job is blocking SSRF, so a seam is warranted. I did think hard about setHostResolver being an exported mutable global there: my read is that anyone able to import the module and call it is already executing code in the same process, so there's no trust boundary being crossed — but it's worth a test-only note at the export, and it must never become reachable from anything network-facing. One inaccuracy for the record: the restore closure reinstates the previous resolver, not the real one, so nested or out-of-order swaps would restore a stale stub. Fine with one call site, fragile if a second appears. Merged! 🚀

PR #2391, opened 2026-07-31
docs(SCRIPTS): document set-status --row/--report selectors (#2355)

### LINKED ISSUE #2355: docs: document the set-status --row/--report selectors in SCRIPTS.md

MeiSiristhebest:
## Summary
`set-status.mjs` gained two explicit selectors — `--row N` and `--report N` — in #2347, but they were never documented (issue #2355). This adds a `## set-status` section to `docs/SCRIPTS.md` documenting both flags, mirroring the existing entries' voice and length.

## Root cause
Tracker row IDs and report IDs are two independent counters sharing one number space; they leapfrog and never realign, so a bare numeric selector is genuinely ambiguous once the counters diverge. `--row`/`--report` were added to name the number space explicitly, but the docs were not updated.

## Changes
- Documented `--row N` (select by tracker `#` column) and `--report N` (select the row whose `Report` cell links report `#N`).
- Noted they are mutually exclusive (passing both is a usage error).
- Explained that both answer the report-number-mismatch guard instead of overriding it, so they skip the guard **without `--force`** (unlike a bare number).
- Added guidance on when a bare number is enough vs. when to use an explicit selector, with one example per flag plus the bare-number case.
- Documented exit codes: `1` = invalid/conflicting selector, `2` = not found.

## Verification
Semantics verified against `set-status.mjs` source:
- Mutual-exclusivity guard: `failUsage('--row and --report are mutually exclusive ...')`.
- Mismatch-guard condition `isBareNumericSelector && !flags.force` → `--row`/`--report` set `explicitSelector`, leaving `isBareNumericSelector = false`, so the guard is skipped without `--force`.
- Exit codes flow through `EXIT_USAGE = 1` (invalid/conflicting selector) and `EXIT_NOT_FOUND = 2` (not found).

This is a documentation-only change; no code or tests are affected.

Closes #2355

## Summary by CodeRabbit

* **Documentation**
  * Added reference documentation for the `set-status` command.
  * Documented row and report selectors, usage examples, supported flags, ambiguity handling, and exit codes.

### DIFF
--- a/docs/SCRIPTS.md
+++ b/docs/SCRIPTS.md
@@ -826,6 +826,31 @@ These have no `npm run` binding — modes and agents call them with
 | `node plugin-audit.mjs` | Static safety scan for community/registry plugins |
 | `node validate-plugin-registry.mjs` | Shape gate for `plugins-registry/<id>.json` files |
 
+## set-status
+
+Canonical tracker write path — strict `states.yml` validation, shared lock, atomic write. Modes (`apply` step 9, `followup`, `batch`) call this instead of hand-editing `applications.md`. Once the tracker-row and report counters diverge, a bare number is ambiguous, so two explicit selectors name which number space you mean:
+
+- `--row N` — select the tracker row whose `#` column is `N`.
+- `--report N` — select the row whose `Report` cell links report `#N`.
+
+`--row` and `--report` are **mutually exclusive** — passing both is a usage error. They answer the report-number-mismatch guard instead of overriding it, so they skip the guard **without `--force`** (a bare number trips the guard and pushes callers toward `--force`).
+
+When a bare number is enough vs. when to be explicit:
+
+- While the two counters are still aligned, a bare `<report#|company>` is fine.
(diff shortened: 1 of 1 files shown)


santifer:
Thanks for this, @MeiSiristhebest — and sorry you ran into a race you couldn't see. #2359 was opened ~8 hours before yours by someone who had claimed the issue on the thread, so it went in first and #2355 is now closed.

Your version isn't redundant, though, and I'd rather harvest it than lose it. Two things yours has that the merged docs don't:

- **The "when is a bare number enough vs. when to be explicit" guidance** — the merged version documents the flags and exit codes but never tells the reader how to decide.
- **The remaining flags** (`--role`, `--on`, `--json`), which are undocumented right now.

A small follow-up PR adding just those two pieces on top of what's on main would be very welcome, and I'd merge it. One correction if you do: use `Applied` rather than `Aplicado` in the examples — the Spanish form still works as a legacy alias, but the canonical state id is the English one, so the docs should teach that.

Issue #1596, opened 2026-07-06
feat(tracker): Via channel — track end employer vs recruiter/agency intermediary (two-stage: parser retrofit + Via column)


Scott-Emberson:
## Problem

When you apply directly on a company website, you know who you're applying to. When you apply **via a recruiter/agency**, you often don't know the end employer until after speaking to the recruiter — sometimes never. The tracker currently has one `Company` column, so agency-mediated applications either pollute it with the agency's name (wrong entity for dedup/analytics) or a guessed placeholder (silently escapes every company+role check).

The tracker should record **both parties**:

- **Company** — the end employer you're applying TO (may be unknown at write time)
- **Via** — the intermediary you're applying THROUGH (agency/recruiter firm; `—` for direct)

…plus a **reveal** workflow for when the employer becomes known, and cross-channel duplicate protection (same role reached via two agencies, or agency + direct — a double submission burns the candidate with both the agency and the employer).

### Stage 1 — parser retrofit (standalone bugfix PR, no schema change)

#1293 moved most tracker readers onto the shared header-aware column mapping (`tracker-parse.mjs`), and #1114 (open) continues that for `normalize-statuses.mjs` and friends. Three readers are still positional and **latent-broken today** for any tracker using the optional Location column from #946/#954:

| Reader | Failure on an inserted column |
|---|---|
| `tracker.mjs` (CLI) | fixed 9-cell destructure; folds >9 cells into Notes, mangling rows |
| `web/src/lib/career-ops.ts` `readApplications` | fixed positional destructure; every web read shifts one column |
| `scan.mjs` `loadSeenCompanyRoles` | positional regex captures the wrong cells; scan dedup misfires |

Stage 1 ports these three onto `tracker-parse.mjs` semantics and adds a **schema-contract test**: shared fixtures (9-col, 10-col Location, unknown-extra-column) that every consumer must map identically — so the next column insertion is a one-place change instead of a repo-wide incident. `normalize-statuses.mjs` is deliberately **not** touched here — #1114 owns it.

Zero behavior change for 9-column trackers. This PR is useful even if Stage 2 is rejected.

### Stage 2 — the `Via` column (feature PR, after design feedback here)

```markdown
| # | Date | Company | Via | Role | Score | Status | PDF | Report | Notes |
```

- `Via` = intermediary firm name, `—` for direct applications. Documented as *intermediary channel* (an agency today; the same cell accommodates `Referral: J. Smith` later — no code generalization now).
[...]

Issue #61, opened 2026-04-07
Feature Future 1.1


ManniMansoor:
### Problem

reserved for future Dev @claude @ @MansoorAhmad

### Proposed solution

not to change otherwise advised from @claude @MansoorAhmad, eei,
[](url)
`[Artificial` Intelligence ]

PR #2397, opened 2026-07-31
fix(dedup): use Unicode-aware Via key for blind-employer grouping

### LINKED ISSUE #2393: bug(dedup-tracker): the #1603 Via guard never reached dedup — a non-Latin agency name collapses to an empty key and one submission is deleted
The cross-channel Via guard from #1596/#1603 landed in `merge-tracker.mjs` but not in `dedup-tracker.mjs`, so dedup deletes a submission that merge correctly keeps.

## The divergence

`dedup-tracker.mjs:307-312` groups unknown-employer rows by their Via channel:

```js
const key = String(entry.company).trim() === '?'
  ? BLIND_KEY + normalizeCompany(entry.via || '')
  : normalizeCompany(entry.company);
```

but `normalizeCompany` is the file-local one at `dedup-tracker.mjs:68-74`, which strips `[^a-z0-9 ]`. `merge-tracker.mjs:710` uses `normalizeVia` from `tracker-parse.mjs:358`, which is Unicode-aware (`[^\p{L}\p{N}]` with NFKC). `dedup-tracker.mjs:19` imports only `resolveColumns` and `parseTrackerRow`, never `normalizeVia`.

The keys that result:

| Via | dedup key | merge key |
|---|---|---|
| `リクルート` | `""` | `リクルート` |
| `パーソル` | `""` | `パーソル` |
| `—` | `""` | `""` |
| `Hays` | `hays` | `hays` |

Every non-Latin agency name collapses to the same empty key, so it also collides with the unrecorded-channel rows (`—` or blank).

## Consequence

Two `?`-employer rows for the same role through two different Japanese agencies share a group, `roleMatch` sees identical titles, neither row is Applied or later, so they cluster and `dedup-tracker.mjs:403-411` deletes one:

```
🗑️  Remove #1 (? — Backend Engineer, Payments Platform, 4.0/5) → kept #2 (4.1/5)
```

That is exactly the double-submission hazard the Via column was introduced to surface, inverted: one of two genuinely separate applications disappears. The same collapse merges any `?` row with an unrecorded channel into any non-Latin-agency row.

Worth noting `data/applications.md` is gitignored (`.gitignore:8`; `git ls-files data/` returns only `.gitkeep`), so there is no version-control recovery. `dedup-tracker` does write a single `.bak` (`:424`), which is the only safety net here and is overwritten on the next run.

## Test gap

`test-all.mjs:7973-7978` is explicitly titled "MERGE-TRACKER CROSS-CHANNEL VIA GUARD" and only ever spawns `merge-tracker.mjs`. The dedup suites at `test-all.mjs:7168` and `:7721` have no Via fixture at all.

A fixture mirroring the #1603 one into the dedup block would pin it:

```js
// tracker with a Via column, two `?` rows, same role, リクルート vs パーソル, both Evaluated
const after = readFileSync(tracker, 'utf-8');
if (after.includes('リクルート') && after.includes('パーソル')) {
  pass('dedup-tracker keeps distinct non-Latin agency submissions (#1603 parity)');
} else {
  fail('dedup-tracker deleted a distinct agency row — blind-via key collapsed to empty');
}
```

Worth adding the symmetric `—` versus `リクルート` case so the unrecorded-channel collapse is covered too.

The fix looks like importing `normalizeVia` and using it for the blind-via key. Happy to send a PR with the fixture above.

Scott-Emberson:
Closes #2393

## The bug

`dedup-tracker.mjs` groups unknown-employer rows (Company `?`, #1596) by their Via channel, and it built that key with the file-local `normalizeCompany()`, which strips everything outside `[a-z0-9]`. `merge-tracker.mjs:710` compares the same Via values with the Unicode-aware `normalizeVia()` from `tracker-parse.mjs` (added in #1603). dedup never imported it, so the two scripts disagreed about agency identity.

| Via | dedup key (before) | merge key (correct) |
|---|---|---|
| リクルート | `""` | `リクルート` |
| パーソル | `""` | `パーソル` |
| — | `""` | `""` |
| Hays | `hays` | `hays` |

Two distinct agencies submitting the same role therefore shared one blind-via group, clustered as duplicates, and the lower-scored row was deleted. An empty or `—` Via collided with every non-Latin agency in the same way. That is data loss on user-layer data: `data/applications.md` is gitignored, so there is no VCS recovery, and dedup's single `.bak` is overwritten on the next run.

## The fix

Import `normalizeVia` and use it for the blind-via grouping key. `normalizeCompany()` is unchanged and non-blind grouping is untouched.

## Empty Via

A row whose Via is empty or `—` still keys to `""` and groups with other via-less blind rows. This matches merge-tracker: its guard rejects a pair only when the two `normalizeVia` keys differ, and two blank cells both produce `""`, so merge-tracker treats them as the same channel. Matching that behaviour keeps the two scripts consistent rather than introducing a third rule. The change is still strictly safer than before, because a blank Via no longer collides with a named non-Latin agency.

## Tests

New fixture in the dedup block of `test-all.mjs`, using the existing `CAREER_OPS_TRACKER` env pattern:

1. Two `?` rows, same role, `リクルート` vs `パーソル`, both Evaluated. Both must survive.
2. Symmetric case: `—` vs `リクルート` on the same role. Both must survive.
3. Control: `Hays` vs `Hays`, same role, both Evaluated. Must still collapse to the higher-scored row.

Case 3 exists so the fix cannot pass by simply never merging anything.

### Mutation evidence

Stashing only the `dedup-tracker.mjs` change and running the new fixture against the unfixed code:

```
🧪 Testing dedup blind-via channel key with non-Latin agencies (#2393)...
  ❌ dedup-tracker collapsed distinct non-Latin agency channels: 1 Payments Platform rows
  ❌ dedup-tracker collapsed a via-less blind row into an agency channel: 1 Checkout rows
  ✅ dedup-tracker still collapses a same-agency re-blast of one blind listing (#2393)
```

Cases 1 and 2 bite. Case 3 passes both before and after, confirming the control is genuine.

With the fix restored, all three pass.

## Validation

`node test-all.mjs`: 2750 passed, 1 failed. The single failure is pre-existing on this Windows box and unrelated: `analyze() appDateSource end-to-end check crashed: EPERM: operation not permitted, symlink`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved duplicate listing detection for tracker entries associated with unknown companies.
  - Preserved separate listings from distinct agencies, including non-Latin agency names.
  - Prevented entries without agency information from being incorrectly merged.
  - Continued selecting the highest-scored listing when duplicates come from the same agency.

- **Tests**
  - Added regression coverage for agency-based deduplication scenarios.

### DIFF
--- a/dedup-tracker.mjs
+++ b/dedup-tracker.mjs
@@ -16,7 +16,7 @@ import { fileURLToPath } from 'url';
 import {
   openTrackerTransaction, rebuildRow, resolveTrackerPath,
 } from './tracker-utils.mjs';
-import { resolveColumns, parseTrackerRow } from './tracker-parse.mjs';
+import { resolveColumns, parseTrackerRow, normalizeVia } from './tracker-parse.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
 // Support both layouts: data/applications.md (boilerplate) and applications.md
@@ -302,11 +302,19 @@ console.log(`📊 ${entries.length} entries loaded`);
 // normalize to the same empty key, so they group by their Via channel instead:
 // the same agency re-blasting one listing IS a duplicate, while the same role
 // via two different agencies is two real submissions and must never merge.
+// The channel key is Unicode-aware (#1603/#2393): normalizeCompany() strips
+// everything outside [a-z0-9], so distinct non-Latin agency names (リクルート,
+// パーソル, …) all collapsed to the same empty key and one of two genuinely
+// separate submissions was DELETED. normalizeVia() is the same key that
+// merge-tracker.mjs uses for its cross-channel guard, so the two scripts
+// cannot drift on agency identity. An absent Via (empty or `—`) still keys to
+// '' and groups with other via-less blind rows, matching merge-tracker, whose
+// guard does not reject a pair whose Via cells are both blank.
 const BLIND_KEY = ' blind-via:';
 const groups = new Map();
 for (const entry of entries) {
   const key = String(entry.company).trim() === '?'
-    ? BLIND_KEY + normalizeCompany(entry.via || '')
+    ? BLIND_KEY + normalizeVia(entry.via || '')
     : normalizeCompany(entry.company);
   if (!groups.has(key)) groups.set(key, []);
   groups.get(key).push(entry);
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -7527,6 +7527,75 @@ try {
   fail(`shared role matcher / dedup safety tests crashed: ${e.message}`);
 }
 
+// ── DEDUP BLIND-VIA CHANNEL KEY: NON-LATIN AGENCIES (#2393) ──────────────
+// Unknown-employer rows (Company `?`) group by their Via channel. dedup-tracker
+// keyed that group with the file-local normalizeCompany(), which strips
+// [^a-z0-9] — so リクルート and パーソル both keyed to '' and two genuinely
+// separate agency submissions for one role landed in the same cluster, and the
+// lower-scored row was DELETED. merge-tracker already compares Via with the
+// Unicode-aware normalizeVia(); dedup must use the same key. A same-agency
+// re-blast must still collapse, otherwise the fix would just be "never merge".
+console.log('\n🧪 Testing dedup blind-via channel key with non-Latin agencies (#2393)...');
+try {
+  const viaDedupTmp = mkdtempSync(join(tmpdir(), 'career-ops-dedup-via-'));
+  try {
+    mkdirSync(join(viaDedupTmp, 'data'));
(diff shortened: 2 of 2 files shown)


santifer:
Correct fix, @Scott-Emberson, and verification backs it: with the one-line change reverted, your own fixture really does delete rows #61 and #63, and with it in place all four survive while the same-agency control still collapses. Aligning with merge-tracker's key is the right call. One correction to the body, since it matters for the next reader: "strictly safer than before" isn't accurate — normalizeVia NFKC-folds and strips spaces, so it also *merges* groups the old key kept apart. Safer in the direction that was losing data, wider in the other. And two blind rows whose Via cells both normalize to empty still share a group, so a residual loss stays open. Merged! 🚀

PR #2398, opened 2026-07-31
test: cover set-status and mark-pdf-ready in the tracker writer lock matrix

### LINKED ISSUE #2396: test(tracker): set-status and mark-pdf-ready are absent from the runWhileLocked matrix, so lock-then-read ordering is unpinned
`tracker-writer-lock-tests.mjs:174-275` runs the `runWhileLocked` stale-snapshot harness against `normalize-statuses`, `dedup-tracker`, `tracker delete`, `tracker export`, and three `reply-watch` variants. It does not cover `set-status.mjs`, which CLAUDE.md names as the canonical write path and which every mode calls.

## The invariant that is unpinned

Every tracker writer must read the file *after* acquiring the lock. Reading first means a row committed by another process in between is erased by the stale snapshot.

`set-status.mjs` gets this right today: it acquires at `:309` and reads at `:313`. I verified the behaviour with a harness mirroring `runWhileLocked` (hold the lock, launch `set-status --row 1 Applied --note probe`, append row #99, release): the concurrent row survived and the status change applied.

The problem is that nothing would catch a regression. Hoisting the read above the lock looks like a reasonable optimisation ("resolve the row before paying for the lock"), and `set-status` already does pre-lock work at `:204-214` validating against `states.yml`, so the shape is inviting. Every existing test would stay green: `set-status-tests.mjs:584` covers lock timeout (exit 4) and `:606` covers lock error, which prove only that it contends for the lock, not that it re-reads after taking it.

Adding it is one entry in the existing matrix, no new harness:

```js
await runWhileLocked({
  name: 'set-status',
  script: 'set-status.mjs',
  args: ['--row', '1', 'Applied', '--note', 'sent CV'],
  content: trackerTable([
    '| 1 | 2026-01-01 | Acme | Engineer | 4.0/5 | Evaluated | ❌ | [1](reports/001-acme.md) | seed |',
  ]),
  verify: c => c.includes('| Applied |') && c.includes('seed; sent CV'),
  // verifyConcurrent already asserts the CONCURRENT_ROW survived
});
```

`mark-pdf-ready.mjs` is missing from the matrix for the same reason. `followup-seed.mjs` is a slightly different case: it uses a separate lock namespace (`followup-seed.mjs:76`), which looks deliberate, but that divergence is itself untested against the tracker lock, so a future change that assumed one shared lock would not be caught.

For contrast, `merge-tracker` is genuinely well covered here, by a two-process race with an IPC checkpoint at `test-all.mjs:8715-8845`. This is a gap in the matrix, not a systemic weakness in the locking tests.

## Context on why this class of test matters more than usual here

`data/applications.md` is gitignored (`.gitignore:8`; `git ls-files data/` returns only `.gitkeep`), so a lost-update has no version-control recovery. `dedup-tracker.mjs:424` and `normalize-statuses.mjs:178` write a single `.bak`; `merge-tracker.mjs`, `set-status.mjs` and `tracker.mjs delete` write none. A regression in this area is unrecoverable rather than merely annoying.

Happy to send a PR adding `set-status` and `mark-pdf-ready` to the matrix.

Scott-Emberson:
Closes #2396

Test-only. No production script changes.

## The gap

`runWhileLocked` in `tracker-writer-lock-tests.mjs` covered `normalize-statuses`, `dedup-tracker`, `tracker delete`, `tracker export` and three `reply-watch` variants. It did not cover `set-status.mjs`, which is the writer CLAUDE.md names as canonical and the one the modes actually call, and it did not cover `mark-pdf-ready.mjs`.

Both scripts are correct today: they acquire the lock and then read. Nothing held that order in place. `set-status-tests.mjs` covers the lock timeout (exit 4) at line 584 and a lock error at line 606, but those only prove the script contends for the lock. Neither can tell a writer that re-reads under the lock apart from one that reads first and writes its stale snapshot back. Since `set-status.mjs` already does real pre-lock work validating the requested state against `states.yml`, hoisting the `readFileSync` up there to "resolve the row before paying for the lock" is a plausible refactor, and it would have stayed green.

## Ordering the fixture mutation

Adding the two matrix entries was not enough on its own, which is worth recording because it is not obvious.

The harness holds the lock, spawns the writer, then commits a concurrent row before releasing. That row is what a stale snapshot erases, so it only discriminates if it lands after the writer's own read. Committing it immediately after `spawn()` does not: a fresh Node process needs tens of milliseconds to boot, so the row is already on disk before even a buggy writer reads it, and the buggy writer then reads the post-mutation file and passes.

The harness now waits for the lock's recover-guard directory to appear before mutating. `acquireTrackerLock` creates and removes `${lockDir}.recover` on every contended pass, so its first appearance marks the moment the child has tried the lock and lost. That instant sits after a pre-lock read and before a post-lock one, which is exactly the discrimination needed. `fs.watch` queues events, so a guard that exists for a millisecond is still observed (measured at 54ms to 65ms after spawn on this box).

The wait is bounded at 2s and is not a barrier. If the guard never appears, because a future lock implementation drops it or `fs.watch` cannot report directory children on some platform, the mutation proceeds anyway and the test degrades to its previous timing-dependent behaviour rather than hanging. Entries using `beforeMutationOutput` skip the wait, since their pre-lock review prompt is a stronger script-specific signal and a writer parked at that prompt has not reached the lock yet.

## Mutation evidence

Each new test was checked against a deliberately broken copy of the script it covers, then the script was restored and `git diff` confirmed clean.

`set-status.mjs`, `readFileSync` hoisted above `acquireTrackerLockForCli`:

```
PASS set-status: contends on the shared lock before reading or writing
FAIL set-status: update failed after lock release (exit=0)
#1 Acme, Engineer: set Evaluated to Applied (note: sent CV)

| 1 | 2026-01-01 | Acme | Engineer | 4.0/5 | Applied | ❌ | [1](reports/001-acme.md) | seed; sent CV |
```

Exit 0, its own edit applied, and the concurrent row #99 gone. That is the failure this test exists to catch. The contention probe still passes under the mutant, which confirms the probe alone was never sufficient.

`mark-pdf-ready.mjs`, same hoist: `FAIL mark-pdf-ready: update failed after lock release (exit=0)`.

Before the guard wait was added, both mutants left the suite fully green at 26 passed, 0 failed.

## followup-seed

Investigated rather than assumed. `followup-seed.mjs:76` uses a separate `career-ops-followups-` lock namespace, keyed by the follow-ups path. That is deliberate and safe, so this is a test, not a bug report.

It reads `applications.md` to find the row and its apply date, then writes only `data/follow-ups.md`. It never writes the tracker. The stale-snapshot invariant cannot apply to it: with no tracker write there is no snapshot to write back, so a row committed while it runs cannot be erased. Its pre-lock tracker read is an observation that can go stale (a row could leave Applied before the pin lands) but never a lost write.

The namespaces have to stay separate. The tracker lock says nothing about `follow-ups.md`, so a seeder holding it would still race a second seeder, and `followup-seed.mjs` is the only writer of that file in the repo. Every other consumer (`followup-cadence`, `reply-watch`, `stats`, `company-history`) only reads it. Sharing the tracker lock would serialize seeding behind unrelated tracker writes and nest two locks without buying any safety.

The new test states that as behaviour. It holds the tracker lock for the whole run and asserts followup-seed completes anyway, seeds the pin, and leaves the tracker byte-for-byte as the lock holder left it. It compares mtime as well as content, so a writer that replaces the tracker with bytes it read a moment earlier is still caught.

Both directions were mutation-checked. Making `seedFollowup` acquire the tracker lock produced `FAIL followup-seed: separate-namespace contract broken (exit=4)` with a tracker lock timeout. Adding a byte-identical `writeFileAtomic(trackerPath, ...)` after the pin write produced `FAIL ... (exit=0)`, caught by the mtime comparison.

## Validation

`node tracker-writer-lock-tests.mjs`: 26 passed, 0 failed. Run twice, same result both times.

`node test-all.mjs`: 2747 passed, 1 failed, 2 warnings. Run twice, identical results. The single failure is a pre-existing Windows-only one on this box, `analyze() appDateSource end-to-end check crashed: EPERM: operation not permitted, symlink`, which also fails on unmodified `upstream/main` here and is unrelated to this change.

These tests spawn real processes and wait on real locks, so the repeat runs were specifically to check for flakiness. None seen.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when multiple tracker-related operations run at the same time.
  - Follow-up processing now completes independently without altering tracker data during concurrent activity.

- **Tests**
  - Expanded coverage for status updates, PDF readiness changes, lock contention, and follow-up data handling.
  - Added safeguards to detect and validate contention scenarios more consistently.

### DIFF
--- a/tracker-writer-lock-tests.mjs
+++ b/tracker-writer-lock-tests.mjs
@@ -4,9 +4,10 @@
 
 import { spawn } from 'child_process';
 import {
-  existsSync, mkdirSync, mkdtempSync, readFileSync, realpathSync, rmSync, utimesSync, writeFileSync,
+  existsSync, mkdirSync, mkdtempSync, readFileSync, realpathSync, rmSync, statSync, utimesSync, watch,
+  writeFileSync,
 } from 'fs';
-import { dirname, join } from 'path';
+import { basename, dirname, join } from 'path';
 import { tmpdir } from 'os';
 import { fileURLToPath } from 'url';
 import { acquireTrackerLock, openTrackerTransaction } from './tracker-utils.mjs';
@@ -33,6 +34,64 @@ const sleep = (ms) => new Promise(resolve => setTimeout(resolve, ms));
 // timeoutMs / staleMs / retryMs are the lock's own parameters. Those are what
 // the tests assert on, so widening them would change what is being tested.
 const HARNESS_WAIT_MS = 30_000;
+
+// How long to wait for evidence that the spawned writer has reached the lock
+// before committing the concurrent row. Bounded, and not a value under test:
+// see watchForContention() for why the wait exists and what happens when the
+// evidence never arrives.
+const CONTENTION_WAIT_MS = 2_000;
+
+/**
+ * Watch for evidence that a spawned writer has attempted the tracker lock and
+ * lost — i.e. that it is now sitting in the retry loop.
+ *
+ * WHY THIS EXISTS: the fixture mutation below is what a writer with a stale
+ * pre-lock snapshot erases, so it only discriminates if it lands AFTER that
+ * writer's read. Committing it immediately after spawn() does not: a fresh
+ * Node process needs tens of milliseconds just to boot, so the row is already
+ * on disk before a buggy writer reads, and the buggy writer then reads the
+ * post-mutation file and passes. That was verified, not assumed — hoisting
+ * set-status.mjs's readFileSync above its acquireTrackerLockForCli call left
+ * this suite fully green until this wait was added.
+ *
+ * The signal is the lock's recover-guard directory: acquireTrackerLock creates
+ * and removes `${lockDir}.recover` on every contended pass, so its first
+ * appearance means "this child has tried the lock and someone else holds it".
+ * That instant sits after a pre-lock read and before a post-lock one, which is
+ * exactly the discrimination the mutation needs. fs.watch queues events, so a
+ * guard directory that exists for a millisecond is still observed.
+ *
+ * It is a bounded wait, never a barrier. If no guard event arrives — a lock
+ * implementation that stops using the guard, a writer that legitimately does
+ * other work first, a platform where fs.watch cannot report directory children
+ * — the mutation proceeds anyway once CONTENTION_WAIT_MS elapses and the test
+ * degrades to its previous timing-dependent behaviour instead of hanging.
+ *
+ * @param {string} dir - Directory containing the lock (watched non-recursively).
+ * @param {string} lockDir - The lock directory whose recover guard signals contention.
+ * @returns {{wait: (timeoutMs: number) => Promise<boolean>, close: () => void}}
+ */
+function watchForContention(dir, lockDir) {
+  const guardPrefix = `${basename(lockDir)}.recover`;
+  let seen = false;
+  let watcher = null;
+  try {
+    watcher = watch(dir, (_event, name) => {
+      if (name && String(name).startsWith(guardPrefix)) seen = true;
+    });
+  } catch {
+    // fs.watch is unavailable or unsupported here — the timed fallback stands in.
+  }
+  return {
+    async wait(timeoutMs) {
+      const deadline = Date.now() + timeoutMs;
+      while (!seen && Date.now() < deadline) await sleep(5);
+      watcher?.close();
+      return seen;
+    },
+    close() { watcher?.close(); },
+  };
+}
 
 function trackerTable(rows) {
   return `# Applications Tracker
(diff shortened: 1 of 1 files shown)


santifer:
Right call extending the lock matrix to the two writers that were outside it, @Scott-Emberson. One thing for a follow-up: the harness discards the boolean from contention.wait(), which is the actual discrimination signal — if the recover guard ever stops being emitted, the six non-output cases quietly burn the full 2s wait instead of failing loudly. Same family as the guards we've been fixing all week: the signal exists but nothing consumes it. Merged! 🚀

Issue #1704, opened 2026-07-08
bug: merge-tracker.mjs accepts stale TSV row numbers without checking they're already used, causing duplicate tracker # after both #749 and #912 fixes


Schlaflied:
## Bug

Found 2026-07-08 while marking rejection emails: `node set-status.mjs <report#> Rejected` twice edited the wrong row, because the tracker `#` column had **two completely unrelated rows sharing the same number**:

- `#698` — both **Company A, Manager-level HR/Talent role** (2026-06-03) *and* **Company B (post-secondary institution), Coordinator role** (2026-05-29)
- `#1167` — both **Company C, Communications Manager role** (2026-06-30) *and* **Company D, Program Coordinator role** (2026-06-26)

A full sweep of `data/applications.md` found **124 report numbers reused across 2+ distinct rows**, out of 1514 total rows, in large contiguous blocks — not isolated one-offs:

```
grep -oE "^\| [0-9]+ \|" data/applications.md | tr -d '| ' | sort -n | uniq -d
```
→ #224–227, #466–486 (21 consecutive), #678–712 (35 consecutive), #948–978 (31 consecutive), #1010, #1039, #1159–1189 (31 consecutive).

## Why this isn't #749 or #912 recurring

- **#749** (fixed by #803, merged 2026-06-06) added `reserve-report-num.mjs` — atomic reservation for **report *file* numbers** (`reports/NNN-slug-date.md`). Doesn't cover tracker-row numbers at all, and many tracker rows never get a report file (rough-estimate-only entries with `—` as the report link).
- **#912** (fixed by #1065, merged 2026-06-16) guarded `merge-tracker.mjs`'s duplicate-detection with a company check, so a colliding TSV falls through to "append as new" instead of silently overwriting the existing row. Verified via #1065's own test (`col912Result` in `test-all.mjs`) that the append path assigns the *next free number* (test explicitly expects row `#2`, not a reuse of the colliding `#1`) — so this fix is doing what it says.
- **Confirmed the current duplicates postdate both fixes**: the `#1159–1189` block spans rows dated 2026-06-26 and 2026-06-29/30 — both well after #803 (2026-06-06) and #1065 (2026-06-16). The filesystem lock added in #941 (2026-06-11, `acquireTrackerLock` in `tracker-utils.mjs`) is also already in place for all of these dates, yet the collision still happened.

## Root cause (best current understanding)

`merge-tracker.mjs`'s per-TSV number assignment:

```js
[...]

Issue #1524, opened 2026-07-04
bug: merge-tracker.mjs fuzzy company+role dedup ignores req/job numbers, false-positive skips distinct postings


Schlaflied:
## Symptom

Adding a TSV for a genuinely new application gets silently skipped as a duplicate:

\`\`\`
⏭️  Skip: XX Bank — Learning & Development Curriculum Designer (existing row 1018 3.8 >= new 3)
\`\`\`

The existing row (1018) is \"XX Bank — Learning & Development Designer III\", req \`R_1494379\`. The new addition was for a different posting, req \`R_1488728\` — a genuinely distinct job at the same company with a similarly-worded title. Employers (XX Bank in particular) commonly run multiple concurrent postings with near-identical generic L&D/HR titles, distinguished only by req number.

## Root cause

\`merge-tracker.mjs\`'s duplicate detection has three tiers: (1) report-number match + company guard, (2) entry-number match + company guard, (3) company + role fuzzy match via \`roleFuzzyMatch\`. Tier 3 has no req/job-number awareness at all — two rows for the same company with similar-sounding titles are flagged as duplicates even when their req numbers (already present as free text in the Notes column, e.g. \`R_1488728\`, \`Req PRACT011038\`, \`Job 202606-116491\`) prove they're different postings.

## Suggested fix

Before falling through to the tier-3 fuzzy match, extract a req/job number from both the incoming addition's notes and the candidate duplicate's notes (a small regex covering the formats already in use: \`R_\d+\`, \`Req \w+\`, \`Job \d+\`, etc.). If both sides have an extracted number and they differ, treat this candidate as **not** a duplicate — skip it and keep searching / fall through to \"new entry\" — regardless of how similar the role text reads. If either side has no extractable number, fall back to today's fuzzy-match behavior unchanged.

Scoped fix — roughly 50 lines including a self-test covering: (a) same-looking titles with different req numbers → not a duplicate, (b) same-looking titles with the same req number → still a duplicate, (c) no req number present on either side → existing fuzzy-match behavior preserved.

PR #2399, opened 2026-07-31
fix(merge-tracker): four silent data-loss paths in the merge loop

### LINKED ISSUE #2392: bug(merge-tracker): three silent data-loss paths — dropped second update, wiped notes, no intra-run dedup
Three paths in `merge-tracker.mjs` lose user data without an error, a warning, or a non-zero exit. All three are in the duplicate-handling block. `data/applications.md` is user-layer data that is never regenerated, so each of these is unrecoverable once the TSVs are archived.

## 1. A second TSV updating the same row in one run is silently dropped

merge-tracker.mjs:730 locates the row to rewrite by its original text:

```js
const lineIdx = appLines.indexOf(duplicate.raw);
if (lineIdx >= 0) {
  ...
  appLines[lineIdx] = updatedLine;
  updated++;
}
```

`duplicate.raw` is captured during the initial parse (:567) and never refreshed after the in-place write. `syncPdfFlags` (:279) does refresh its cached row, so the two paths disagree.

So in a single merge run with two TSVs matching the same tracker row (3.5 → 4.0, then → 4.5): the first update rewrites the line, and the second finds `indexOf` returning -1, skips the whole block, increments no counter, prints no warning, and still logs its optimistic `🔄 Update` line beforehand. The TSV is then archived to `merged/` as if it had been applied. The 4.5 evaluation is gone.

This is reachable whenever a batch run re-evaluates the same posting twice, which is exactly what the batch workflow does after a scoring change.

## 2. The score-upgrade path discards the existing Notes

merge-tracker.mjs:740:

```js
notes: `Re-eval ${addition.date} (${oldScore}→${newScore}). ${addition.notes}`,
```

`duplicate.notes` is dropped wholesale. Anything the user recorded with `set-status --note` (recruiter name, req number, interview context) is erased by a re-evaluation that happens to score higher.

This one compounds: the req/job-number guard added in #1524 reads `extractReqNumber(app.notes)` at :718 to decide whether two similar titles are genuinely distinct postings. Wiping the notes destroys the evidence that guard depends on, so a later merge is more likely to wrongly collapse two real postings.

## 3. No intra-run dedup, so one run can create the duplicate it exists to prevent

New rows go to `newLines` (:777) and are spliced in at the end (:797). They are never pushed to `existingApps` — line 567 is the only push, inside the initial parse. Duplicate detection at :660-721 therefore only ever sees rows that existed before the run started.

Two parallel workers evaluating the same posting each reserve their own number and write their own TSV. Both merge as new rows in a single run, producing exactly the company+role duplicate the file header claims to prevent. `verify-pipeline.mjs` then warns about it and `dedup-tracker.mjs` has to clean it up, at which point one row's status and notes are merged away.

## Tests

I could not find coverage for any of the three in tests/merge-tracker.test.mjs or the merge sections of test-all.mjs. The invariants worth pinning, independent of how the fixes are written:

- Two TSVs updating the same row in one run: both updates land, or the second fails loudly. Never a silent no-op with the TSV archived.
- A score upgrade preserves the existing note content alongside the re-eval prefix, and a req number in the old note survives the merge.
- Two TSVs for the same company+role in one run produce one row, not two.
- A merge that no-ops for any reason must not archive the TSV that was not applied.

Happy to send a PR. Flagging first because fix 1 has a design choice in it (refresh the cached row versus track rows by index throughout), and fix 2 has a merge-semantics question about how old and new notes should be combined when both are non-empty.

### LINKED ISSUE #2394: bug(merge-tracker): a missing separator row silently drops every new entry, archives the TSVs, and reports success
When `merge-tracker.mjs` cannot find the table's separator row, it drops every new row on the floor, archives the TSVs, and prints a success summary.

## The path

`merge-tracker.mjs:788-797` scans for `SEPARATOR_ROW_RE` to decide where to splice new rows:

```js
let insertIdx = -1;
for (let i = 0; i < appLines.length; i++) {
  if (SEPARATOR_ROW_RE.test(appLines[i])) { insertIdx = i + 1; break; }
}
if (insertIdx >= 0) {
  appLines.splice(insertIdx, 0, ...newLines);
}
```

There is no `else`. When no separator matches, `newLines` is discarded silently. Execution continues to `writeFileAtomic` at `:802`, then `renameSync` moves the TSVs into `merged/` at `:806-808`, then `:812` prints the count computed back at `:778`.

Against a tracker containing only `# Applications Tracker`:

```
✅ Moved 1 TSVs to merged/
📊 Summary: +1 added, 🔄0 updated, ⏭️0 skipped
```

The tracker is unchanged. The evaluation now exists nowhere except the archived TSV, and the user has been told it landed.

## Likelihood

Lower than the other merge paths, since `tracker.mjs:468-471` always emits a separator and the onboarding snippet in CLAUDE.md includes one. The realistic trigger is a hand-edit that clears the table, or an agent recreating the header from memory and omitting the separator line. Spaced separators (`| --- | --- |`) are matched correctly, so only a genuinely absent separator does this.

What makes it worth fixing despite the low likelihood is the shape of the failure: total data loss for that run, plus a success message that actively discourages the user from looking. `data/applications.md` is gitignored and `merge-tracker` writes no `.bak`, so there is no recovery path.

## Suggested behaviour

Fail loudly and leave the TSVs in `batch/tracker-additions/` so a re-run applies them once the table is repaired. The invariant worth stating in a test is that `+N added` must mean N rows are actually in the file, and that a merge which cannot place its rows must not archive its inputs.

## Test sketch

```js
// Tracker with a heading but no table. The merge must NOT report success.
const r = spawnSync(NODE, [join(ROOT, 'merge-tracker.mjs')], { env: { ...}, encoding: 'utf-8' });
// exit !== 0 or a warning on stderr; the TSV is still in tracker-additions/, not merged/
// and never: /\+1 added/ passing while the tracker lacks the row
```

Every existing fixture in `tests/merge-tracker.test.mjs` and `test-all.mjs` hardcodes a well-formed separator, and `test-all.mjs:8503` asserts the separator survives a merge, which is the opposite direction from this.

Happy to send a PR.

Scott-Emberson:
Closes #2392
Closes #2394

These are four ways `merge-tracker.mjs` could throw an evaluation away and still report success. They share a branch because all four live in the same merge loop and the fixes interact: refreshing the cached row (gap 1) changes what the new intra-run dedup (gap 3) can find, and both feed the row-lookup helper. Reviewing them separately would mean reviewing the same forty lines three times. There is one commit per concern.

None of these losses are recoverable. `data/applications.md` is gitignored, so there is no VCS copy, and merge-tracker writes no `.bak`.

## Gap 1: a second update to the same row was dropped (commit 1)

`appLines.indexOf(duplicate.raw)` looked the row up by the snapshot taken when the tracker was parsed. Nothing refreshed it after `appLines[lineIdx] = updatedLine`, so the next TSV matching that row searched for a line that no longer existed, got `-1`, and fell through an `if (lineIdx >= 0)` block with no else. Nothing incremented a counter or printed a warning, and the TSV was archived into `merged/` anyway. The stale snapshot also fed the score comparison, so with several re-evaluations of one row the survivor was decided by TSV filename sort order rather than by score.

Before, tracker row #1 at 4.0 plus TSVs at 4.2 and 4.7:

```
🔄 Update: #1 Acme — Staff Data Platform Engineer (4→4.2)
🔄 Update: #1 Acme — Staff Data Platform Engineer (4→4.7)
📊 Summary: +0 added, 🔄1 updated, ⏭️0 skipped
| 1 | 2026-02-01 | Acme | Staff Data Platform Engineer | 4.2/5 | ...
```

Both updates were announced, only one was counted, and the 4.7 evaluation no longer exists anywhere.

The fix re-parses the written line back into the cached row, which is what `syncPdfFlags` has always done. The now unreachable "row not found" branch is no longer silent: it keeps the TSV out of `merged/`, reports it in the summary, and exits non-zero.

## Gap 3: no dedup between rows added in the same run (commit 2)

All three dedup tiers search `existingApps`, which was only ever populated from rows read out of the file. A row appended during the run went to `newLines` and was invisible, so two TSVs for the same company and role in one batch both appended. That is the case `CLAUDE.md`'s Pipeline Integrity rule forbids, and it happens on an ordinary path: two evaluations of one role in a single batch.

Before, an empty tracker plus two Acme / Staff Data Platform Engineer TSVs gave `+2 added` and two rows.

New rows now go into `existingApps` as they are created. Updating a row created earlier in the same run needs the line lookup to search the queued rows too, so `appLines.indexOf()` moved into a `replaceTrackerLine()` helper that checks `appLines` first, then `newLines`.

## Gap 2: Notes overwritten on a score upgrade (commit 3)

The update path rebuilt Notes as `Re-eval {date} ({old}->{new}). {new notes}` and discarded the existing cell. That cell carries the `Applied YYYY-MM-DD` marker `followup-cadence.mjs` prefers over the Date column, the req number this same script reads back through `extractReqNumber()` for the #1524 sibling-req guard, contacts, and anything the user wrote with `set-status --note`.

**Format choice for the reviewer.** Preserving the old notes is not optional, but where the re-eval marker goes is a judgement call, so please push back if you want it the other way. I kept the existing text verbatim and first, then appended the marker and the new notes:

```
Applied 2026-01-15. Req R_1488728. recruiter jane@acme.example. Re-eval 2026-03-01 (4→4.7): re-scored after JD refresh
```

The reason for that order rather than leading with the marker: `parseAppliedDate()` and `extractReqNumber()` both return their *first* match, so leading with the established text keeps a row's apply date and req number stable across re-evaluations instead of letting each new evaluation's text take over. Two trade-offs come with it. Notes grows by one clause per re-evaluation, and the join trims a single trailing period off the existing text so it does not read as `..`. Re-running an identical evaluation does not repeat its own text, only the marker.

## #2394: no separator row dropped everything and reported success (commit 4)

`insertIdx` stayed `-1` when `SEPARATOR_ROW_RE` matched nothing, the splice was skipped with no else, and the run went on to write the file, archive every TSV and print `+N added`. Against a tracker holding only `# Applications Tracker`, one addition produced `+1 added`, an unchanged tracker, and the evaluation living only in `merged/`.

It now aborts before writing anything, so the tracker and the additions directory are left as they were and the run replays cleanly once the header is repaired. The error names the missing separator, prints the expected header, lists the rows that were not merged, and exits non-zero.

## Tests

`tests/merge-tracker.test.mjs` grows a `runMergeDetailed` helper (seeded tracker rows, custom header, process output, exit code, and which TSVs were archived versus left pending) and 14 assertions across the four defects. Where it matters the assertion is on the consequence rather than on the text: gap 2 imports `analyzeFromContent` from `followup-cadence.mjs` and checks the merged row still reports `appliedDate: 2026-01-15` with `appDateSource: notes`, then runs a second merge to confirm the sibling-req guard still fires.

Each test was checked against the unfixed code by reverting its own fix and re-running:

| Defect | Mutation | Result |
|---|---|---|
| Gap 1 | `git stash` of `merge-tracker.mjs` | 3 of 3 fail. Row ends at 4.2 not 4.7; summary reads `🔄1 updated`; the downgrade case is announced as an update and reports `⏭️0 skipped` |
| Gap 3 | `git stash` of `merge-tracker.mjs`, gap 1 fix in HEAD | 2 of 2 fail with two identical Acme rows. The third assertion in that block is a control: distinct roles must stay separate, and it passes either way by design |
| Gap 2 | `mergeNotes(...)` call swapped back to the old template literal | 3 of 3 fail. Notes reduced to the re-eval marker; cadence falls back to `appDateSource: evaluation-date-fallback`; the sibling req folds into the row |
| #2394 | `git stash` of `merge-tracker.mjs` | 3 of 3 fail with `exit 0`, `+1 added`, and the TSV archived into `merged/`. The fourth assertion is a control: the same TSV must still merge against a healthy tracker |

## Validation

`node test-all.mjs`: 2761 passed, 1 failed, 2 warnings. The single failure is the pre-existing `analyze() appDateSource end-to-end check crashed: EPERM: operation not permitted, symlink`, a Windows symlink permission issue in my environment that is unrelated to this change.

## Behavior changes

Two contract changes callers should know about.

**merge-tracker can now exit 1.** The separator-row abort and the "addition not applied" path both exit non-zero, where the script previously always exited 0 after printing a summary. That is caller-visible: `batch/batch-runner.sh` runs under `set -euo pipefail` and calls merge-tracker unguarded, so a damaged tracker header now hard-stops the batch at the merge step instead of letting it continue past a merge that silently dropped rows. Stopping is the point (the alternative was archiving evaluations that never landed), but any wrapper that relied on merge-tracker never failing needs a guard if it wants to carry on.

**Same-run num collisions (review round 2).** Registering same-run rows in `existingApps` had a side effect: two TSVs that both claimed one reserved num at the same company for different roles hit the tier-2 num+company match, and the second folded into the first row (first title, second score). On main both rows survived through the #1704/#1733 renumbering path. Tier-2 now also requires a fuzzy role match before accepting a same-run row as an update candidate, so a reservation race renumbers the second row and keeps both evaluations, matching main. For rows already on disk, num keeps its meaning as the tracker row id and nothing changes there.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved company matching across Unicode and non-Latin names.
  * Enhanced duplicate detection for existing and same-run entries.
  * Preserved notes and follow-up dates while merging updates.
  * Prevented incomplete tracker files from being written or archived.
  * Improved failed-addition reporting and retry handling.

* **Tests**
  * Added coverage for score updates, note merging, duplicate detection, missing separators, replay scenarios, and related requirement checks.

### DIFF
--- a/merge-tracker.mjs
+++ b/merge-tracker.mjs
@@ -225,6 +225,29 @@ function extractReqNumber(notes) {
 }
 
 /**
+ * Company equality for duplicate detection.
+ *
+ * normalizeCompany() strips everything outside [a-z0-9], so a company name
+ * written entirely in a non-Latin script (CJK, Cyrillic, Arabic, …) normalizes
+ * to the empty string — and every such name would compare equal to every other
+ * one. That collapses DIFFERENT companies into one row as soon as their roles
+ * fuzzy-match (six Japanese companies posting データエンジニア → one row).
+ * When the normalized key carries no signal, fall back to raw trimmed
+ * equality so distinct non-Latin companies stay distinct; the unknown-employer
+ * `?` marker still matches itself, so the #1596 cross-channel guard keeps its
+ * existing behavior.
+ *
+ * @param {string} a - Company cell from one side of the comparison.
+ * @param {string} b - Company cell from the other side.
+ * @returns {boolean} True when the two cells name the same company.
+ */
+function companiesMatch(a, b) {
+  const key = normalizeCompany(String(a));
+  if (key !== normalizeCompany(String(b))) return false;
+  return key !== '' || String(a).trim() === String(b).trim();
+}
+
+/**
  * Combine an existing row's Notes with a re-evaluation's Notes.
  *
  * The update path used to overwrite Notes with
@@ -257,8 +280,17 @@ function mergeNotes(existingNotes, addition, oldScore, newScore) {
   const incoming = String(addition.notes ?? '').trim();
   const marker = `Re-eval ${addition.date} (${oldScore}→${newScore})`;
   // Re-running the same evaluation would otherwise repeat its own text; the
-  // marker still records that the re-evaluation happened.
-  const tail = incoming && !prev.includes(incoming) ? `${marker}: ${incoming}` : marker;
+  // marker still records that the re-evaluation happened. Repeats are detected
+  // per CLAUSE — the same '. ' separator this function joins with — not by raw
+  // substring: `prev.includes(incoming)` dropped any new note that happened to
+  // appear INSIDE an existing clause ("Remote" vanished against "Remote OK").
+  // A clause equals the incoming text either bare or in the marker-prefixed
+  // form a previous run of this function appended (`{marker}: {incoming}`).
+  const clause = (s) => s.replace(/\.+$/, '').trim();
+  const incomingClause = clause(incoming);
+  const isRepeat = incoming !== '' && prev.split(/\.\s+/).map(clause)
+    .some(c => c === incomingClause || c.endsWith(`: ${incomingClause}`));
+  const tail = incoming && !isRepeat ? `${marker}: ${incoming}` : marker;
   return prev ? `${prev}. ${tail}` : tail;
 }
 
@@ -756,10 +788,9 @@ for (const file of tsvFiles) {
     // appearing for two different companies is sequence drift, not a duplicate.
     // Without the company guard, a NewCo TSV with report [1] silently overwrites
     // the existing tracker row [1] belonging to an unrelated company.
-    const normCompany = normalizeCompany(addition.company);
     duplicate = existingApps.find(app => {
       const existingReportNum = extractReportNum(app.report);
-      return existingReportNum === reportNum && normalizeCompany(app.company) === normCompany;
+      return existingReportNum === reportNum && companiesMatch(app.company, addition.company);
     });
     if (duplicate) reportNumMatched = true;
   }
@@ -771,18 +802,25 @@ for (const file of tsvFiles) {
     // 067 while the tracker was already at #69). A bare num collision across
     // *different* companies is that drift, not a duplicate — matching on num
     // alone silently merges a brand-new role into an unrelated existing row.
-    const normCompany = normalizeCompany(addition.company);
     duplicate = existingApps.find(app =>
-      app.num === addition.num && normalizeCompany(app.company) === normCompany
+      app.num === addition.num && companiesMatch(app.company, addition.company)
+      // Same-run num collisions are reservation races, not row-id references:
+      // two TSVs that both claimed num=5 for DIFFERENT roles at one company
+      // are two distinct evaluations, and folding them keeps the first title
+      // with the second score (main renumbers and keeps both via the
+      // #1704/#1733 path). For a row queued THIS run, only accept the num
+      // match when the roles also fuzzy-match — consistent with tier-3's
+      // cross-run semantics. For rows already on disk, num remains the
+      // tracker row id and behavior is unchanged.
+      && (!app.addedThisRun || roleFuzzyMatch(addition.role, app.role))
     );
   }
 
   if (!duplicate) {
     // Company + role fuzzy match
-    const normCompany = normalizeCompany(addition.company);
     const additionReqNum = extractReqNumber(addition.notes);
     duplicate = existingApps.find(app => {
-      if (normalizeCompany(app.company) !== normCompany) return false;
+      if (!companiesMatch(app.company, addition.company)) return false;
       if (!roleFuzzyMatch(addition.role, app.role)) return false;
       // Cross-channel guard (#1596): unknown-employer rows (`?`) all normalize
       // to the same empty company key, but the same role via two DIFFERENT
@@ -892,7 +930,14 @@ for (const file of tsvFiles) {
     // higher-scored addition updates it in place just like a row already on
     // disk.
     const parsedNew = parseAppLine(newLine);
-    if (parsedNew) existingApps.push(parsedNew);
+    if (parsedNew) {
+      // Mark the row as queued this run: tier-2 treats a num collision with a
+      // same-run row as a reservation race unless the roles also fuzzy-match
+      // (see the tier-2 guard above). Object.assign() in the update path never
+      // deletes the flag, so it survives in-place refreshes within the run.
+      parsedNew.addedThisRun = true;
+      existingApps.push(parsedNew);
+    }
     added++;
     console.log(`➕ Add #${entryNum}: ${addition.company} — ${addition.role} (${addition.score})`);
   }
--- a/merge-tracker.mjs
+++ b/merge-tracker.mjs
@@ -893,9 +893,29 @@ if (newLines.length > 0) {
       break;
     }
   }
-  if (insertIdx >= 0) {
-    appLines.splice(insertIdx, 0, ...newLines);
+  if (insertIdx < 0) {
+    // #2394: no separator row means no insert point. The old code left
+    // insertIdx at -1, skipped the splice with no else, then carried on to
+    // write the file, archive every TSV into merged/ and print "+N added" —
+    // so a tracker whose table header had been damaged (or a fresh file that
+    // only ever got its `# Applications Tracker` line) swallowed a whole batch
+    // of evaluations while reporting success. Nothing survived: the tracker is
+    // gitignored and merge-tracker keeps no .bak.
+    //
+    // Abort before writing anything. Leaving the tracker and the additions dir
+    // exactly as they were makes the run a no-op that replays cleanly once the
+    // table is repaired.
+    console.error(`❌ Aborting merge: ${basename(APPS_FILE)} has no table separator row ("|---|------|...").`);
+    console.error(`   There is no insert point for ${newLines.length} new row(s), so NOTHING was written and no TSV was archived.`);
+    console.error('   Repair the tracker table header, then re-run — the pending additions in');
+    console.error(`   ${ADDITIONS_DIR} will merge then. Expected header:`);
+    console.error('     | # | Date | Company | Role | Score | Status | PDF | Report | Notes |');
+    console.error('     |---|------|---------|------|-------|--------|-----|--------|-------|');
+    for (const line of newLines) console.error(`   not merged: ${line}`);
+    trackerLock.release();
+    process.exit(1);
   }
+  appLines.splice(insertIdx, 0, ...newLines);
 }
 
 // Write back
--- a/merge-tracker.mjs
+++ b/merge-tracker.mjs
@@ -625,6 +625,10 @@ tsvFiles.sort((a, b) => {
 console.log(`📥 Found ${tsvFiles.length} pending additions`);
 
 const newLines = [];
+// TSVs whose evaluation could not be applied to the tracker. They are kept out
+// of merged/ so a re-run picks them up, and they make the process exit non-zero
+// instead of reporting a success that did not happen.
+const failedAdditions = [];
 
 for (const file of tsvFiles) {
   const content = readFileSync(join(ADDITIONS_DIR, file), 'utf-8').trim();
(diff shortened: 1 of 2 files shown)


santifer:
Excellent work, @Scott-Emberson — all four data-loss paths are real. I reproduced each one on main before reading your fix: the stale `raw` snapshot really did drop a second re-evaluation *and* archive its TSV, the Notes rebuild really did destroy the `Applied YYYY-MM-DD` marker and the req number, two same-run TSVs really did append two rows, and a separator-less tracker really did write, archive everything and exit 0. The ordering argument for keeping existing Notes first is correct, and I verified it holds because `parseAppliedDate`/`extractReqNumber` both return their *first* match.

Verification also turned up **a fifth silent-loss path that gap 3 opens**, and it's the reason I'm not merging yet.

**`normalizeCompany()` strips everything outside `[a-z0-9]`, so every CJK / Cyrillic / Arabic company name normalizes to the empty string** — and they all compare equal to each other. Pushing same-run rows into `existingApps` means tier-3 matching (empty company key + fuzzy role) now collapses *different companies* inside a single batch run.

Measured on your branch, in one additions directory:

```
6 distinct Japanese companies, all posting データエンジニア
  main → 6 rows
  PR   → 1 row, exit 0, summary reads "+1 added, 🔄5 updated"
```

Five real evaluations gone, and `applications.md` is gitignored with no `.bak`, so unrecoverable. Cyrillic behaves the same (2 → 1). This matters concretely here: we ship `modes/` for ja, ko, zh, zh-TW, ru, ua, ar and hi, and `batch/batch-runner.sh` produces exactly this input shape. The #1596 cross-channel guard doesn't cover it — that one only fires when the company cell is literally `?`.

The root cause is pre-existing (main collapses the same pair across runs), but this PR extends its reach into the batch path, which is where the loss is largest.

**What I'd like before merge:** in the tier-3 guard, require raw-string company equality whenever `normalizeCompany()` returns `''` — or reuse `normalizeVia()`'s Unicode-aware normalization for company too. Plus a test with a non-Latin company name, since nothing currently covers that boundary.

Two smaller findings, your call whether they belong here or in a follow-up:

1. **Same-run `num` / report-number collisions now fold distinct roles into one row.** Two TSVs both `num=5` at Acme, one "Staff Data Platform Engineer" and one "Director of Product Marketing": main renumbers and keeps both (the #1704/#1733 path), your branch emits an add then an update and ends with one row carrying the first title and the second's score. Consistent with cross-run semantics, but new for the batch path and undisclosed. Your control test can't exercise it because its nums don't collide.
2. **`mergeNotes`' substring dedup drops the incoming note.** `!prev.includes(incoming)` is a raw substring test with no word boundary, and `prev` grows a clause per re-evaluation, so false positives compound. Two measured: existing "Applied 2026-01-15. Remote OK" + incoming "Remote" → the new note vanishes; and in the 6-evaluation run, "eval 2" was dropped because the accumulated marker text contained that literal string. Same silent-drop class this PR exists to fix.

Also flagging for awareness, not change: the new exit-1 paths are caller-visible — `batch/batch-runner.sh` runs under `set -euo pipefail` and calls merge-tracker unguarded, so a damaged tracker header now hard-stops the batch. That's better than silent loss, but it's a contract change worth stating in the PR body.

Happy to merge once the non-Latin case is closed. The rest of this is genuinely good work and I don't want it sitting long.

Scott-Emberson:
Pushed 2994cfe, everything addressed.

Non-Latin collapse: added a `companiesMatch()` helper. When `normalizeCompany()` comes back empty it falls back to raw trimmed-string equality. I applied it to all three dedup tiers, not just tier-3, because tier-1 (report#) and tier-2 (num) have the same empty-key hole once same-run rows enter `existingApps`. The `?` marker still matches itself, so the #1596 via guard is unchanged. Regression test uses two Japanese companies (株式会社ゼータ / 合同会社オメガ) both posting データエンジニア and asserts 2 rows, with a control that the same company twice still dedups to 1 with the higher score.

mergeNotes: swapped the substring test for clause-level comparison. `prev` is split on the `'. '` separator mergeNotes itself joins with, and incoming counts as a repeat only when a clause equals it exactly, bare or in the marker-prefixed form a prior run appended. Your "Remote" vs "Remote OK" case now appends; identical re-runs are still suppressed. Tests cover both.

Num collision: took the fix here rather than a follow-up since it turned out small. Rows queued this run are flagged, and tier-2 additionally requires a role fuzzy-match for those rows only. Two TSVs colliding on num with distinct roles now fall through to the new-entry path and renumber via the existing #1704/#1733 logic, matching main's outcome. Cross-run num-as-row-id semantics are untouched. Test: collision with distinct roles gives 2 rows with the second renumbered; control: same num plus same role still folds to one.

PR body now has a "Behavior changes" section covering the exit-1 contract change for batch-runner.sh and the collision decision above.

santifer:
Everything addressed, and better than asked, @Scott-Emberson. Applying companiesMatch() to all three tiers rather than just tier-3 — because tier-1 and tier-2 have the same empty-key hole once same-run rows enter existingApps — is the reasoning I hoped the RC would trigger. The regression test with 株式会社ゼータ / 合同会社オメガ posting the same role and asserting two rows, plus the same-company control still deduping, pins exactly the boundary that was untested. And swapping mergeNotes' substring test for clause-level comparison kills the 'eval 2 vanishes inside Re-eval 2026-02-02' class properly. Verified the merged tree green on the full suite. Merged! 🚀

Issue #2009, opened 2026-07-17
roleFuzzyMatch silently dedupes leveled variants ("Associate X") against the bare title ("X") at the same company


clede:
## What happened?

When a company posts two distinct, real requisitions for the same base title at different levels — e.g. "Associate Product Manager" and "Product Manager" — `merge-tracker.mjs`'s dedup logic (`roleFuzzyMatch` in `role-matcher.mjs`) treats them as the same opening and silently drops the new one, even though they're different req IDs, different URLs, and can score very differently. This happened with a real pair of postings at the same company in my tracker; the lower-level req was evaluated, scored, and had a full report written, but never got its own tracker row — `merge-tracker.mjs` kept the existing higher-scored row instead and reported the new one as skipped.

The bigger risk: I then tried to update the (nonexistent) tracker row for the dropped posting via `set-status.mjs <company> SKIP`. Since no row existed under that ID, the script fell back to its company-name-only match, found the *other* (correct, unrelated) row for that company, and updated its status instead — nearly overwriting the status of a live, unrelated application. I caught it and reverted, but the failure mode is a silent, wrong-row mutation, not an error.

There's an existing guard for this (#1524 in code comments) that treats two rows as distinct if the Notes column on both sides contains a recognizable req/job/posting-ID token (`REQ_NUMBER_RE` in `merge-tracker.mjs`). But that convention is undocumented outside the source code — `AGENTS.md`'s "TSV Format for Tracker Additions" spec (the instructions an agent actually follows when writing a tracker-addition TSV) never mentions it, so there's no way for an agent to know to use it.

**Related to #793, but not fixed by it.** PR #793 (merged 2026-06-09) fixed a real but different bug in the same function: it changed the similarity ratio from `overlap / smaller-set-size` to true Jaccard (`overlap / union`) so that unrelated roles sharing a few generic/brand tokens (e.g. two different "Full-Stack Engineer 5" teams) stop false-matching. That fix operates on the word-overlap step. My bug is upstream of it, in the seniority-token gate (`extractSeniorities()`): the check that's supposed to block "Senior X" from matching "Principal X" only fires when *both* titles carry a recognized seniority token. "Associate X" vs. bare "X" (no seniority word on one side) skips that gate entirely and falls through to word overlap — where, unlike #793's false-positive cases, the titles really are near-identical strings, so even a *correctly computed* Jaccard ratio still matches them. I confirmed this reproduces on the current `role-matcher.mjs` (already includes #793's fix):

```js
[...]

PR #2400, opened 2026-07-31
refactor(set-status): extract shared candidate resolution

### LINKED ISSUE #2348: Refactor set-status candidate resolution without weakening selector safety guards
## Summary

Refactor the duplicated candidate-resolution flow in `set-status.mjs` into a shared helper (for example, `resolveCandidates`).

The helper should be used by all three selector paths:

- explicit `--report N`
- explicit `--row N` and bare numeric row selection
- company-name selection

## Rationale

PR #2347 intentionally keeps this refactor out of the wrong-row-write fix so the existing safety behavior can be reviewed independently. The new `--report` branch is now a third copy of the same match → optional role narrowing → ambiguous/not-found → unique-target flow.

## Required behavior to preserve

- **#1704:** Duplicate tracker row numbers must continue to fail closed with the existing duplicate-number handling and candidate visibility; the helper must not reintroduce “first match wins.”
- **#2009:** `--role` must be validated against a uniquely selected target, not merely used to break ties among multiple candidates. The existing post-resolution role-mismatch safety check must remain effective.
- Preserve selector-specific messages and exit-code behavior where practical, including not-found and ambiguous candidate reporting.
- Ensure the `--report` selector participates in the same shared candidate-resolution mechanism.

## Affected area

- `set-status.mjs`, primarily `resolveRow()` and the related role-validation path.

## Acceptance criteria

- [ ] The three selector branches delegate their common candidate-resolution work to one shared helper.
- [ ] Existing `#1704` duplicate-tracker-number behavior remains covered and passes.
- [ ] Existing `#2009` unique-target `--role` mismatch protection remains covered and passes.
- [ ] `--report` matching, role narrowing, not-found handling, and ambiguity reporting retain their current behavior.
- [ ] Focused regression tests cover the preserved safeguards and the refactored resolution paths.

## Backlinks

- Deferred from PR #2347: https://github.com/santifer/career-ops/pull/2347
- Review discussion: https://github.com/santifer/career-ops/pull/2347#discussion_r3686598945
- Requested by: @LayerC0de

LayerC0de:
Closes #2348

## Summary

The `--report`, `--row`/bare-numeric and company selector paths each carried their own copy of the same flow: match → optionally narrow by `--role` → refuse to guess between survivors → return the unique row. Only the predicate and the two messages ever differed. All four entry points now delegate to one `resolveCandidates(matches, { notFound, ambiguous })`.

## Why this is worth doing

Centralising it matters more than the duplication it removes.

**Failing closed on 2+ candidates *is* the #1704 fix.** A stale tracker `#` reused across two rows makes "the first match" a silent coin flip on which company gets edited. While that behaviour lived in three copies, a change that reintroduced first-match-wins in one branch would have been invisible in the other two. There is now one place to get it wrong, and one place to test it.

## Behaviour is unchanged

Every message, exit code and candidate payload is byte-identical. **The existing tests pass unedited** — that was the acceptance bar I held myself to: if a behaviour-preserving refactor requires touching the tests that pin the behaviour, it isn't behaviour-preserving.

## #2009 is deliberately untouched

`--role` only *narrows* inside the helper and never validates a lone match. That is load-bearing, not an oversight: the post-resolution role-mismatch check exists precisely because a selector matching exactly one row never reaches the narrowing branch. Validating `--role` inside `resolveCandidates()` would look like a tidy-up and would quietly duplicate — then eventually replace — the #2009 guard.

The helper's doc comment says so explicitly, because "validate `--role` here too" is the obvious wrong fix for whoever reads this next.

## Tests

Four new cases, pinning what the helper now owns centrally:

- **`--report` fails closed on 2+ linking rows.** Previously untested — that path had no 2+ coverage at all, so a regression there would have gone unnoticed.
- **`--role` narrows from the `--report` path**, not just the two paths that already had coverage.
- **A `--role` matching neither candidate still fails closed** with both candidates listed, rather than silently picking one.
- **Parity:** all four entry points (bare numeric, `--row`, `--report`, company) fail closed on 2+ candidates with a candidate list, and none of them writes. This is the test that catches a future edit to `resolveCandidates()` regressing a single path.

Fixtures are synthetic (Acme / Globex / Initech).

## Verification

- `set-status-tests.mjs`: **75 passed, 0 failed** — the 71 pre-existing cases unmodified, plus the 4 above.
- **Negative control:** disabling the fail-closed branch in `resolveCandidates()` produces **8 failures**, including the parity test naming all four paths by name. Restored and hash-verified byte-identical to the commit before running anything else. A refactor test that cannot fail proves nothing, so this seemed worth doing explicitly.
- `node test-all.mjs`: **2747 passed, 1 failed.** The single failure is `analyze() appDateSource end-to-end` with `EPERM: symlink`, a Windows symlink-permission issue in the harness; an unmodified `origin/main` in the same environment produces the identical result.
- No documentation change is owed: `docs/SCRIPTS.md` documents `set-status` behaviour, not `resolveRow()` internals, and behaviour is unchanged.

## Acceptance criteria from #2348

- [x] The three selector branches delegate their common candidate-resolution work to one shared helper — four, counting bare-numeric and `--row` separately.
- [x] Existing #1704 duplicate-tracker-number behaviour remains covered and passes.
- [x] Existing #2009 unique-target `--role` mismatch protection remains covered and passes.
- [x] `--report` matching, role narrowing, not-found handling and ambiguity reporting retain their current behaviour.
- [x] Focused regression tests cover the preserved safeguards and the refactored resolution paths.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved status selection across numeric, row, report, and company selectors.
  - Prevented changes when a selector matches multiple candidates.
  - Added consistent no-write behavior for ambiguous selections.
  - Improved role-based narrowing for report selections while preserving candidates when no role matches.

- **Tests**
  - Added regression coverage for candidate resolution and ambiguity handling.

### DIFF
--- a/set-status-tests.mjs
+++ b/set-status-tests.mjs
@@ -1017,5 +1017,109 @@ const TRACKER_REPORT_MISMATCH = `# Applications Tracker
   });
 }
 
+// ── shared candidate resolution (#2348) ──────────────────────────
+//
+// The three selector paths delegate their match → narrow → refuse-to-guess
+// flow to resolveCandidates(). These pin the properties that must survive any
+// future edit to that helper: every path fails CLOSED on 2+ survivors (the
+// #1704 property, previously enforced in three separate copies), and --role
+// narrowing works from every path rather than only the two that had tests.
+{
+  // Two rows link the SAME report — reachable when a re-evaluation is filed
+  // against an existing report, or a report link is copied between rows.
+  const TRACKER_DUP_REPORT = `# Applications Tracker
+
+| # | Date | Company | Role | Score | Status | PDF | Report | Notes |
+|---|------|---------|------|-------|--------|-----|--------|-------|
+| 1 | 2026-06-01 | Acme | Backend Engineer | 4.2/5 | Evaluated | ✅ | [9](../reports/009-acme-2026-06-01.md) | — |
+| 2 | 2026-06-02 | Globex | Data Engineer | 4.0/5 | Evaluated | ✅ | [9](../reports/009-globex-2026-06-02.md) | — |
+`;
+
+  const withDupReport = fn => {
+    const sandbox = makeSandbox(TRACKER_DUP_REPORT);
+    try { fn(sandbox); } finally { rmSync(sandbox.dir, { recursive: true, force: true }); }
+  };
+
+  // Previously untested: the --report path had no 2+ coverage at all, so a
+  // regression to first-match-wins there would have gone unnoticed.
+  withDupReport(sandbox => {
+    const before = readTracker(sandbox);
+    const r = runSetStatus(['--report', '9', 'Applied', '--json'], sandbox);
+    let parsed = null;
+    try { parsed = JSON.parse(r.stdout); } catch {}
+    if (r.code === 3 && parsed?.code === 'ambiguous' && parsed.candidates?.length === 2
+        && readTracker(sandbox) === before) {
+      pass('#2348: --report fails closed on 2+ linking rows, with candidates');
+    } else {
+      fail(`#2348: --report dup → code=${r.code} json=${JSON.stringify(parsed)}`);
+    }
+  });
+
+  // --role narrowing must serve the --report path too, not just the numeric
+  // and company paths that already had coverage.
+  withDupReport(sandbox => {
+    const r = runSetStatus(['--report', '9', 'Applied', '--role', 'Data Engineer', '--json'], sandbox);
+    let parsed = null;
+    try { parsed = JSON.parse(r.stdout); } catch {}
+    if (r.code === 0 && parsed?.company === 'Globex') {
+      pass('#2348: --role narrows a --report match to the intended row');
+    } else {
+      fail(`#2348: --report + --role → code=${r.code} company=${parsed?.company}`);
+    }
+  });
+
+  // A --role that matches NEITHER candidate must not silently pick one; the
+  // helper falls through with the original list so both stay visible.
+  withDupReport(sandbox => {
+    const before = readTracker(sandbox);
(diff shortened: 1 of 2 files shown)


LayerC0de:
@FReptar0 no rush on this one, but it has been green since 31 July with no open threads, so I wanted to check it had not fallen off the list.

It is fairly contained: +176/-58 across two files, closing #2348. The four selector paths in `set-status.mjs` now delegate to a single `resolveCandidates(matches, { notFound, ambiguous })` instead of each carrying its own match, narrow, refuse-to-guess flow.

Happy to rebase or split it if either would make the review easier.

LayerC0de:
@santifer one question for you on this, since it lands on the row-resolution contract rather than just the code shape.

Once all four selector paths delegate to `resolveCandidates`, "refuse to guess between two or more candidates" stops being three copies that can drift apart and becomes one invariant every path inherits. That is the #1704 fix, and I think it is the right default. It does also mean any selector added later gets the behaviour whether or not its author thought about it.

Is that the contract you want stated? Ambiguity always fails closed, with no per-path opt-out. If so, I am happy to write it down somewhere durable rather than leaving it implicit in the helper. `DATA_CONTRACT.md` seems the natural home, or a comment on `resolveCandidates` if you would rather keep it local to the code.

No rush on this, and @FReptar0 already has the code review side.

santifer:
Byte-identical extraction proven the hard way, @LayerC0de: the negative control broke 8 tests including all four #2347 discriminators, which is exactly what a refactor's safety net should do. resolveCandidates() is now the single home for the match→narrow→fail-closed flow, and the #1704/#2009 guards both survived with proof. Merged. 🚀

PR #2406, opened 2026-07-31
fix(dashboard): the progress funnel dropped Hired from every stage

abankar1:
Closes #2405.

`ComputeProgressMetrics` builds a cumulative funnel — its own comment says *"each stage counts all apps that reached at least that stage"* — but `hired` was in **none** of the tiers. A hire definitionally passed through all four, so a landed job was invisible at Applied, Responded, Interview and Offer, and (since `applied` is the denominator for all three rates) missing from the denominator as well.

A pipeline whose only row is `Hired`:

| Stage | Before | After |
|---|---|---|
| Evaluated | 1 | 1 |
| Applied | **0** | 1 |
| Responded | **0** | 1 |
| Interview | **0** | 1 |
| Offer | **0** | 1 |
| rates | 0% / 0% / 0% | real values |

The single best outcome the tool can represent rendered as though nothing ever happened.

## Why this is a broken mirror, not a judgment call

`computeFunnel()` in `stats.mjs` is the *"canonical funnel definition for career-ops going forward"*, and its docstring describes its math as mirroring **this exact function**:

> Rejected counts into everApplied (a rejection proves a submission), **Hired counts into every stage through everOffer** (a landed job proves the offer and everything before it)

The Go side already matched on `Rejected` — it just never got `Hired`. Same incomplete-`Hired`-propagation family as #2289, #2295, #2297.

## Changes

- `hired` added to all four cumulative tiers, matching `computeFunnel()` exactly.
- `TotalOffers` counts a hire too (a hire proves an offer was received).
- Comment updated to state the rule explicitly and point at the canonical definition.

## Tests

`ComputeProgressMetrics` had **no test coverage at all** — that's why this slipped through. Added `progress_metrics_test.go` with three tests: a lone `Hired` row counting into all four stages + `TotalOffers`; the rates including it in numerator and denominator; and a monotonicity check (`applied ≥ responded ≥ interview ≥ offer`) across all nine canonical states with exact expected counts.

Go isn't installed on my machine, so these are verified by CI's `go test ./...` rather than locally — the assertions are derived from `NormalizeStatus`'s mapping for each literal used.

## Not changed (flagged in the issue)

`pm.ActiveApps` still counts a `Hired` row as *active* even though it's terminal. There's no canonical SSOT for "active", so I didn't guess — happy to follow up if you want it excluded.

2 commits (fix + test).

## Summary by CodeRabbit

* **Bug Fixes**
  * Hired applications are now included across all relevant application funnel stages.
[...]

PR #2410, opened 2026-07-31
fix(web): analytics tells an offer-holder they have 0 interviews (and nudges them to try harder)

### LINKED ISSUE #2409: web/analytics: the interviews & offers tiles read 0 for candidates who already advanced past those stages
### What

The two headline tiles on `/analytics` are derived from the **current-state** stage bars:

```tsx
const offers     = stageCounts.find((s) => s.key === "OFFER")?.n ?? 0;
const interviews = stageCounts.find((s) => s.key === "INTERVIEW")?.n ?? 0;
```

`stageCounts` is a snapshot — each application sits in exactly one stage. But these tiles are *achievement counters*, and their zero-state renders a coaching nudge:

```tsx
hint={interviews === 0 ? "Interviews follow replies — keep follow-ups warm →" : undefined}
hint={offers === 0 ? "Offers follow interviews — keep the conversations going →" : undefined}
```

So a candidate who **advanced past** a stage reads 0 for it and is told to try harder to reach the stage they already cleared.

### Reproduced (port of `canonStatus` + the tile derivation)

```
candidate currently holds an OFFER
   tiles → interviews=0, offers=1
   nudge: "Interviews follow replies — keep follow-ups warm"        ← they already interviewed

candidate was HIRED (landed the job)
   tiles → interviews=0, offers=0
   nudges: "Interviews follow replies…" + "Offers follow interviews…"  ← they got the job

realistic: hired + offer + 3 applied
   tiles → interviews=0, offers=1
   nudge: "Interviews follow replies — keep follow-ups warm"
```

Being coached to "keep the conversations going" on the day you accepted an offer is the worst version of this.

Note it is **not only a `Hired` gap** (#2289/#2295/#2297/#2405 family) — an offer-holder with no `Interview` row hits it too, so this is the general snapshot-vs-cumulative confusion.

### The stage bars are fine

The "Pipeline by stage" bars below deliberately *are* a snapshot — `Hired`, `Rejected`, `Discarded` are each their own bar and an application belongs to one. Only the two tiles ask a cumulative question. No change proposed to the bars.

### Fix

Make the tiles cumulative, using the core's canonical definition rather than a new invention — `computeFunnel()` in `stats.mjs`:

```js
everInterview = Interview + Offer + Hired
everOffer     = Offer + Hired
```

`Rejected` is deliberately **not** folded in: a status is a snapshot, so a rejection never reveals which stage it came from — `stats.mjs` calls the middle stages lower bounds for exactly this reason.

PR incoming with the logic extracted to a testable `.mjs` helper (matching `clean-chips.mjs` / `stream-parse.mjs`).

abankar1:
Closes #2409.

The `interviews` and `offers` headline tiles on `/analytics` read from the **current-state** stage bars:

```tsx
const offers     = stageCounts.find((s) => s.key === "OFFER")?.n ?? 0;
const interviews = stageCounts.find((s) => s.key === "INTERVIEW")?.n ?? 0;
```

`stageCounts` is a snapshot — an application sits in exactly one stage. But these tiles are *achievement counters* whose zero-state renders a coaching nudge, so a candidate who **advanced past** a stage reads 0 for it and is told to try harder to reach the stage they already cleared.

Reproduced with a port of `canonStatus` + the tile derivation:

```
candidate currently holds an OFFER
   interviews=0, offers=1
   nudge: "Interviews follow replies — keep follow-ups warm"          ← already interviewed

candidate was HIRED (landed the job)
   interviews=0, offers=0
   nudges: "Interviews follow replies…" + "Offers follow interviews…"  ← they got the job
```

Being told to "keep the conversations going" on the day you accepted an offer is the worst version of this.

This is **not only a `Hired` gap** (#2289/#2295/#2297/#2405 family) — an offer-holder with no `Interview` row hits it too. It's the general snapshot-vs-cumulative confusion.

## Fix

Count cumulatively, using the core's canonical definition rather than inventing one — `computeFunnel()` in `stats.mjs`:

```js
everInterview = Interview + Offer + Hired
everOffer     = Offer + Hired
```

`Rejected` is deliberately **not** folded in: a status is a snapshot, so a rejection never reveals which stage it came from — `stats.mjs` calls the middle stages lower bounds for exactly this reason.

**The stage bars are untouched.** They *are* a snapshot by design (`Hired`, `Rejected`, `Discarded` each their own bar) and remain so; only the two tiles ask a cumulative question.

Logic extracted to a pure `web/src/lib/funnel-tiles.mjs` so it's unit-testable under `node --test` without a TS runner — matching `clean-chips.mjs` / `stream-parse.mjs` / `act-envelope.mjs`. The page change is a two-line swap.

## Tests

`web/tests/lib/funnel-tiles.test.mjs` — 6 regressions: the offer-holder and hired cases, accumulation across a realistic pipeline, stages that never reached an interview, the deliberate `Rejected` exclusion, and empty/absent input.

`npm test` (web) → **66 passed, 0 failed**.

2 commits (fix + test).

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated analytics headline counters to include candidates who progressed beyond interview and offer stages.
  * Improved funnel totals for downstream stages, including offers and hires.
  * Handled empty, missing, or unavailable pipeline data without errors.

* **Tests**
  * Added coverage for cumulative interview and offer counts across realistic pipeline scenarios, including pre-interview, rejected, offered, and hired candidates.


santifer:
Routing this one to the web maintainer's queue (`area:web` — it's purely `web/`), @abankar1. Same class as the funnel fix I just merged in the Go dashboard (#2406): a landed job shouldn't read as zero interviews. Worth noting in that review that `stats.mjs`'s `computeFunnel()` is the canonical definition and already counts Hired into everOffer, so the web should derive from it rather than restate the math — the copy-vs-derive pattern we've been closing all week (#2369).

abankar1:
Noted on derive-vs-restate — agreed, and I've taken the same point to its source: #2431 closes **fix 1 of #2369** (the web's `CADENCE_DEFAULTS` hand-copy is deleted; the baseline now comes from `followup-cadence.mjs --json`).

For **this** PR the honest position is that `computeFunnel()` isn't reachable the same way, and I'd rather say so than claim a derivation I haven't got:

- It's a **function**, not a data file or a `--json` verdict. The two derivation mechanisms already in the tree don't fit — `states.ts` reads `templates/states.yml` at runtime, and `tracker-table.mjs` reads the shared `tracker-aliases.json`; both work because the source of truth is *data*. `/api/followups` spawns the core, but this is a server-rendered page, not an API route, so a subprocess per render is the wrong trade.
- Nothing in `web/` currently imports across the `web/` boundary at all, so `await import(<careerOpsRoot>/stats.mjs)` from a server component would be new ground with real Next bundling risk (`Critical dependency: the request of a dependency is an expression`). I'm not confident enough in that to put it in a tile fix.

So the two options I can actually stand behind:

1. **What's here now** — the roll-up lives in one small `.mjs` helper, and I add a `test-all` check importing **both** `computeFunnel` and `cumulativeTiles` and asserting they agree across a matrix of pipelines. That's stronger than 55.3c/55.3d were: not a text comparison of two constant blocks, but *behavioural equivalence*, so changing the canonical roll-up (a future terminal state, say) fails the web copy by name.
2. **Real derivation** — have the core expose the stage roll-up as data (the `tracker-aliases.json` pattern) and let both sides read it. Cleaner and it kills the class, but it's a core change that outgrows a tile fix.

Happy to push (1) onto this PR now, or open (2) against #2369 as the general "logic mirrors, not copied lists" half you flagged there. Whichever the web maintainer prefers — just say which and I'll do it rather than guess.

santifer (approved):
Approved. The bug is real, and your framing of it is the part I want to keep: these are **achievement counters reading a snapshot**, so the zero-state coaching nudge fires at exactly the people who already cleared the stage. Being told to "keep the conversations going" on the day you accepted an offer is the worst possible moment for that message, and it's the kind of defect that never gets filed — people just quietly conclude the tool doesn't understand them.

You're also right that it isn't only the `Hired` family (#2289/#2295/#2297/#2405). An offer-holder with no `Interview` row hits it too, which is what makes this the general fix rather than a fifth patch.

Verified the definition you adopted against the core rather than the PR text — `stats.mjs:165`:

```js
const everInterview = n('Interview') + n('Offer') + n('Hired');
const everOffer     = n('Offer') + n('Hired');
```

Your `cumulativeTiles` matches it. Taking the core's definition instead of inventing a cumulative rule is the right instinct and it's why this merges as-is.

**Two notes I'm keeping as my own follow-up, not asking you to do:**

1. `funnel-tiles.mjs` restates a definition that lives in `stats.mjs`, with nothing tying them together — so if the core ever changes what "ever interviewed" means, the web drifts silently. That's a known class in this repo (#2369) and the cure belongs with the maintainer of `web/`, which is me. #2750 landed the pattern for it tonight: mirror + a parity test anchored on the real core helper, with an anti-vacuity guard so it can't pass by measuring nothing.

2. The match is `s.includes(k)` rather than equality. Today no canonical status contains another as a substring, so it's exact by accident of the vocabulary; a future `Offer Declined` would count as an offer. **That pattern is pre-existing** — `analytics/page.tsx:24` already reads `canonStatus(a.status).includes(s.key)` — so you followed the file, you didn't introduce it. Fixing it is my debt, not yours, and it has nothing to do with the 13 days you've already waited.

Merging.

Issue #912, opened 2026-06-10
merge-tracker: bare number collision overwrites an unrelated company's tracker row


darkpandawarrior:
## Bug

`merge-tracker.mjs` documents its dedup as *"company normalized + role fuzzy match + report number match"*, but in practice the first two duplicate checks are not company-guarded:

1. **Report number match** — `extractReportNum(addition.report) === extractReportNum(app.report)`, no company check
2. **Entry number match** — `app.num === addition.num`, no company check

Only the third fallback (company + role fuzzy) checks the company at all.

TSV files in `batch/tracker-additions/` carry the **report sequence** (per the TSV convention, `{num}` is the report number), while the tracker's `#` column is its own **row sequence**. These are independent sequences, so collisions between them are routine — and when one happens, the "update in-place" path rewrites the existing row's **company, role, report link, and notes** with the addition's values, silently destroying an unrelated entry.

## Repro

Tracker with rows #1–#3 for companies A, B, C (e.g. manually tracked interview-stage entries with `N/A` scores and `—` report links). Then a first batch evaluation produces `batch/tracker-additions/001-newco.tsv`:

```
1	2026-06-10	NewCo	New Role	Evaluated	2.7/5	❌	[1](reports/001-newco-2026-06-10.md)	note
```

Run `node merge-tracker.mjs`:

- **Expected:** NewCo appended as row #4.
- **Actual:** row #1's company/role/report/notes are replaced with NewCo's data (the `N/A` score parses as 0, so any real score wins the "higher score → update" comparison). With TSVs 001–003 in one batch, rows #1–#3 are all corrupted.

## Proposed fix

- A report-number or entry-number match only counts as a duplicate when the **normalized company also matches**.
- A colliding TSV (number taken by a different company) falls through to the "new entry" path and is appended with the next free number.
- Regression test in `test-all.mjs` covering: (a) collision leaves the existing row byte-identical and appends the addition with the next free number, (b) a legit same-company re-eval with a higher score still updates in place.

I have the fix + test ready and will open a PR linked to this issue.

PR #2412, opened 2026-08-01
fix(merge-tracker): write a lower-scored re-evaluation through instead of discarding it

### LINKED ISSUE #2411: bug(merge-tracker): a re-evaluation that scores LOWER is silently discarded
`merge-tracker.mjs` discards a re-evaluation whose score is **lower** than the row it matches. Nothing warns, no counter reflects it, the new report is orphaned, and the TSV is archived into `merged/` as though it had landed.

`data/applications.md` is user-layer data that is never regenerated and is gitignored, so the loss is unrecoverable.

## The path

merge-tracker.mjs:728

```js
if (newScore > oldScore) {
  // ... rewrite the row, updated++
} else {
  console.log(`⏭️  Skip: ${addition.company} — ${addition.role} (existing #${duplicate.num} ${oldScore} >= new ${newScore})`);
  skipped++;
}
```

The dedup tiers above this correctly identified the row as the same opening. The only question left is which evaluation is *newer*, and the answer is always the incoming one — but a lower score routes it to `else` and it is dropped.

## Why a lower score is the case that matters most

A downgrade is rarely noise. In practice it means one of:

- **the candidate's targeting changed** — a role that was on-strategy no longer is, so the same JD honestly scores lower;
- **a staleness signal surfaced late** — e.g. reading `first_published` for the first time and finding the requisition is ~12 months old;
- **a gap was re-weighted** — something previously judged a "ramp-up item" turns out to be a hard gate.

All three are strictly better information than the score already in the row. Discarding them leaves the tracker asserting a stale optimistic number, which is worse than having no row at all: it actively misleads the next prioritisation pass.

## Reproduction

Tracker row #7 at 3.8/5, then a re-evaluation of the same requisition at 3.0/5:

```
📥 Found 1 pending additions
⏭️  Skip: Initech — Payments PM (existing #7 3.8 >= new 3)

✅ Moved 1 TSVs to merged/

📊 Summary: +0 added, 🔄0 updated, ⏭️1 skipped
```

Row #7 still reads `3.8/5`, still points at report `[7]`, and report `[8]` exists on disk with nothing referencing it. `verify-pipeline.mjs` then reports it as an orphan report, which is the only surviving trace.

Equal scores take the same branch, so a re-evaluation that confirms the score but carries fresher notes and a newer report link is also dropped.

## Suggested fix

Let a re-evaluation write through in both directions, and make a downgrade visibly distinct rather than silent:

- log it with its own marker (e.g. `🔽 ... — DOWNGRADE, re-eval scored lower`) so it cannot be mistaken for an upgrade in a batch log;
- count it as an update, not a skip;
- record the superseded report number in Notes (`Superseded report [7] (was 3.8/5)`), so that if the fuzzy matcher mis-paired two genuinely different roles the previous evaluation is still reachable from the tracker alone.

That last point matters because `role-matcher.mjs` treats "Senior" as a stopword and drops tokens of ≤3 characters, so near-identical titles can pair when they shouldn't. Writing through unconditionally without leaving a pointer would turn a bad match into data loss of a different kind; the Notes breadcrumb keeps it recoverable.

`skipped` stays meaningful — it still counts malformed and genuinely rejected rows.

## Relationship to #2392 / #2394 / PR #2399

PR #2399 fixes four other data-loss paths in this same merge loop (stale row snapshot, no intra-run dedup, Notes overwritten on upgrade, missing separator row). **None of them cover this branch** — that PR only changes what happens *inside* the `newScore > oldScore` arm and never revisits the `else`.

The two do touch adjacent lines, so whichever lands second will need a small rebase. Happy to sequence behind #2399 if that is easier for review.

Souptik96:
Closes #2411

A re-evaluation whose score is **lower** than the row it matches was silently discarded. The row kept its stale optimistic score, the newer report was orphaned, no counter moved, and the TSV was archived into `merged/` as though it had landed.

`data/applications.md` is gitignored user-layer data that is never regenerated, so the loss was unrecoverable.

## The path

merge-tracker.mjs:728

```js
if (newScore > oldScore) {
  // ... rewrite the row, updated++
} else {
  console.log(`⏭️  Skip: ${addition.company} — ${addition.role} (existing #${duplicate.num} ${oldScore} >= new ${newScore})`);
  skipped++;
}
```

The dedup tiers above this had already established the row is the same opening. The only remaining question is which evaluation is *newer*, and that is always the incoming one — but a lower score routed it to `else`.

## Why the downgrade case is the one that matters

A lower score is rarely noise. In practice it means:

- **targeting changed** — a role that was on-strategy no longer is, so the same JD honestly scores lower;
- **a staleness signal surfaced late** — e.g. reading `first_published` for the first time and finding the requisition is ~12 months old;
- **a gap was re-weighted** — something previously judged a "ramp-up item" turns out to be a hard gate.

All three are strictly better information than the number already in the row. Discarding them leaves the tracker asserting a stale optimistic score, which is worse than no row at all: it actively misleads the next prioritisation pass.

## What changed

A re-evaluation now writes through in both directions:

- **downgrades log with their own marker** (`🔽 ... — DOWNGRADE, re-eval scored lower`) so they cannot be mistaken for an upgrade when skimming a batch log;
- **they count as an update, not a skip**;
- **Notes records `Superseded report [N] (was X.X/5)`**.

That last point is deliberate rather than decorative. `role-matcher.mjs` treats "Senior" as a stopword and drops tokens of ≤3 characters, so near-identical titles can pair when they shouldn't. Writing through unconditionally with no breadcrumb would convert a bad match into data loss of a different kind — the Notes pointer keeps the superseded evaluation reachable from the tracker alone.

Equal scores also write through, since the notes and report link are still fresher than what the row holds. `skipped` keeps its meaning: it still counts malformed and genuinely rejected rows.

The upgrade arm is otherwise untouched — `pdfIndex` lookup, `reportNumMatched` role handling, and the existing `Re-eval` notes format all behave exactly as before.

## Tests

`tests/merge-tracker.test.mjs` gains nine cases: seven on the downgrade path (write-through, report re-pointing, superseded marker, stdout announcement, update-vs-skip counters) and two guarding the upgrade path against regression.

Two small harness changes: `runMerge()` takes an optional `seedRows` argument so a pre-existing row can be matched against, and returns `{ text, stdout }` so the summary line can be asserted. Both existing callers updated.

Verified failing against unpatched `main`, where they reproduce the exact defect:

```
⏭️  Skip: Initech — Payments PM (existing #7 3.8 >= new 3)
📊 Summary: +0 added, 🔄0 updated, ⏭️1 skipped
```

```
$ node tests/merge-tracker.test.mjs
  ✅ merge-tracker preserves the canonical Hired status (no silent downgrade)
  ✅ merge-tracker resolves the "accepted" alias to Hired
  ✅ re-evaluation with a LOWER score writes through (no silent skip)
  ✅ downgrade re-points the Report link at the newer report
  ✅ downgrade records the superseded report number in Notes
  ✅ downgrade is announced on stdout, not merged silently
  ✅ downgrade counts as an update, not a skip
  ✅ re-evaluation with a HIGHER score still writes through unchanged
  ✅ upgrade does not add the superseded-report marker
```

Full suite: 2740 passed. The failures on my machine are pre-existing and reproduce identically on unpatched `main` — two need Node ≥ 22.5 for `node:sqlite` (I'm on v20.18.1), one is an `EPERM` symlink issue on Windows, one is a 10-column/Location shift in `tracker-columns-tests.mjs`, and two are the `#912` report-number collision cases. I confirmed each against a stashed baseline rather than assuming; none are touched by this change.

## Relationship to #2399

PR #2399 fixes four other data-loss paths in this same merge loop (stale row snapshot, no intra-run dedup, Notes overwritten on upgrade, missing separator row). It only changes what happens *inside* the `newScore > oldScore` arm and never revisits the `else`, so there is no overlap in intent — this is a fifth, independent gap.

The two do touch adjacent lines, so whichever lands second needs a small rebase. **Happy to sequence behind #2399** if that's easier for review; say the word and I'll rebase on it.

## How this surfaced

Found while re-evaluating a requisition I'd scored 3.8/5 three weeks earlier. The re-run scored it 3.0 — the posting turned out to be ~12 months old, and a gap I'd called a ramp-up item was really a hard gate. The merge printed `⏭️ Skip`, archived the TSV, and left the tracker claiming 3.8. I only noticed because `verify-pipeline.mjs` flagged the new report as an orphan; without that I'd have kept prioritising a role I'd already decided against.

## Summary by CodeRabbit

* **Bug Fixes**
  * Re-evaluations now update tracker entries for higher, equal, and lower scores.
  * Downgrades are clearly identified and retain superseded report details and previous scores.
  * Updates consistently refresh dates, scores, report links, and re-evaluation notes while preserving existing status.
  * Generated-PDF indicators are applied when available.
* **Tests**
  * Added regression coverage for score changes, report updates, downgrade notices, note preservation, and update counts.

### DIFF
--- a/merge-tracker.mjs
+++ b/merge-tracker.mjs
@@ -273,9 +273,15 @@ function companiesMatch(a, b) {
  * @param {object} addition - Parsed TSV addition (uses `notes` and `date`).
  * @param {number} oldScore - Score currently on the row.
  * @param {number} newScore - Score from the addition.
+ * @param {string} [extraMarker] - Appended to the re-eval marker, before any
+ *   incoming notes. Used by the downgrade path to record the superseded report
+ *   (#2411). It rides on the marker rather than on `addition.notes` so that the
+ *   repeat detection below keeps comparing the user's own text against the
+ *   user's own text — a generated fragment in that comparison would make every
+ *   downgrade look like a new note and defeat it.
  * @returns {string} Combined Notes cell.
  */
-function mergeNotes(existingNotes, addition, oldScore, newScore) {
+function mergeNotes(existingNotes, addition, oldScore, newScore, extraMarker = '') {
   // Trailing period trimmed only so the '. ' join does not produce '..'; no
   // other character of the existing text is touched. The tracker's "no data"
   // sentinels (the looksLikeScoreCell set minus the score-only DUP) count as
@@ -286,7 +292,9 @@ function mergeNotes(existingNotes, addition, oldScore, newScore) {
     ? ''
     : prevRaw.replace(/\s*\.\s*$/, '');
   const incoming = String(addition.notes ?? '').trim();
-  const marker = `Re-eval ${addition.date} (${oldScore}→${newScore})`;
+  const marker = extraMarker
+    ? `Re-eval ${addition.date} (${oldScore}→${newScore}) — ${extraMarker}`
+    : `Re-eval ${addition.date} (${oldScore}→${newScore})`;
   // Re-running the same evaluation would otherwise repeat its own text; the
   // marker still records that the re-evaluation happened. Repeats are detected
   // per CLAUSE — the same '. ' separator this function joins with — not by raw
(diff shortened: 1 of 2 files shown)


santifer:
Hey @Souptik96, this one's still wanted: writing the lower-scored re-evaluation through instead of discarding it is the right call, and your checks are all green. It picked up a conflict since #2399 landed (four data-loss fixes in the same merge loop). Could you rebase onto main? I'll review as soon as it's green again.

Souptik96:
Rebased onto `main` — thanks for the nudge, and for keeping it open.

`#2399` and `#2483` both landed in this block, so the diff is genuinely different from the one you first saw. Three things worth your attention rather than buried in the commit message:

**1. One assertion from #2399 is inverted, on purpose.** Its case *"a later lower-scored addition is skipped against the CURRENT score, not the parse-time one"* encoded exactly the behaviour this PR changes, so it could not survive as written. The invariant it actually guards — that the comparison reads the current row rather than the parse-time snapshot — is still under test, now via the re-eval marker: against a stale snapshot the second addition renders `(4.0→4.2)` instead of `(4.7→4.2)`. So it asserts the marker and `🔄2 updated` instead of a skip. Flagging it explicitly because a merged test changing expectation is the kind of thing that should be a decision, not a surprise.

**2. The rebase resolved two of CodeRabbit's three findings on its own**, which is a nice argument for #2399 having landed first:

- *"Log emitted before the row replacement is confirmed"* — now moot. The write goes through `replaceTrackerLine()`, a miss is a loud `console.error` plus a `failedAdditions` entry rather than a silent no-op, and `duplicate` is re-parsed from the written line so a second addition compares against the current score.
- *"Chained re-evaluations drop the earlier superseded marker"* — also moot. `mergeNotes()` keeps the existing cell verbatim and first, so a previous `Superseded report [...]` survives because the whole previous cell does. I added a test asserting the seeded `Req R5639` still reads back after a downgrade, since this script's own sibling-req guard depends on that.
- Its third suggestion, an equal-score case, is now covered.

**3. `mergeNotes()` gains one optional argument**, `extraMarker`. The superseded fragment rides on the re-eval marker rather than on `addition.notes` deliberately: your repeat detection compares clauses, and a generated fragment inside that comparison would make every downgrade look like a brand-new note and defeat it.

Verification: 36 assertions pass in `tests/merge-tracker.test.mjs` against 27 on `main`, 58 across the three suites that touch merge-tracker, no regressions. CI is green on ubuntu and macos with windows still running as I write this.

Happy to split the `mergeNotes()` signature change out, or to restore #2399's assertion as a separate skip-path test if you would rather keep both behaviours documented.

santifer:
Thanks for the rebase, @Souptik96, and for the patience: I asked for it on Tuesday and then took three days to come back. Writing the lower score through is right, a re-evaluation is new information whether it flatters the role or not, and the #2399 data-loss guards came through the rebase intact. Merged. 🚀

PR #2415, opened 2026-08-01
fix(sync-pdf): reject unknown options

ShiroKSH:
## Summary

Reject unknown sync-pdf-flags options before opening the tracker transaction. A typo such as --dry-rn can no longer continue in live write mode.

## Root cause

Argument parsing recognized supported flags but silently ignored every other token, then ran with default write behavior.

## Validation

- node tests/sync-pdf-flags.test.mjs
- node test-all.mjs — 2754 passed, 0 failed; 2 expected environment warnings

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved command-line validation by detecting unrecognized options before processing begins.
  * Added clear error reporting in both human-readable and JSON formats.
  * The command now exits with status 1 when an unknown option is provided, preventing unintended tracker changes.

* **Tests**
  * Added regression coverage for invalid command-line options and error handling.

### DIFF
--- a/sync-pdf-flags.mjs
+++ b/sync-pdf-flags.mjs
@@ -24,9 +24,18 @@ const APPS_FILE = resolveTrackerPath(CAREER_OPS);
 const PDF_MANIFEST = process.env.CAREER_OPS_PDF_INDEX || join(CAREER_OPS, 'data', 'pdf-index.tsv');
 
 const flags = { dryRun: false, json: false };
+const unknownOptions = [];
 for (const arg of process.argv.slice(2)) {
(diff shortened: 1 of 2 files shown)


santifer:
Welcome to career-ops, @ShiroKSH! Surgical close of the flag blind-spot in sync-pdf-flags: known options untouched (verified dry-run, JSON and live paths all behave), and the rejection is genuinely fail-fast — validation runs before the tracker transaction opens, so exit 1 leaves zero lockfiles behind. Merged — your first contribution. 🚀

PR #2420, opened 2026-08-01
fix(web): strip the author letter from any lettered block, not just A-G (#2324)

### LINKED ISSUE #2324: Report view: cleanHeading()/splitSections() stop at [A-G], so "H) Draft Application Answers" keeps its letter
### Code of Conduct

### Existing issues

### What happened?

`cleanHeading()` in the report view strips the author-letter prefix from a section heading for display, but its character class stops at `G`:

```js
// web/src/components/report-view.tsx:30
.replace(/^\s*(?:Block\s+)?[A-G][).:]\s*/i, "")
```

Reports also write `## H) Draft Application Answers`, so that heading renders with its letter still attached while every other section renders clean.

`splitSections()` has the same limit one line down, so an `H)` section also gets `letter: null` instead of `"H"`.

### Steps to reproduce

```js
cleanHeading("G) Posting Legitimacy")        // "Posting Legitimacy"      correct
cleanHeading("H) Draft Application Answers") // "H) Draft Application Answers"  letter left in
```

In a 411-report corpus here, the author-letters actually present are:

```
402  G)      401  E)      400  D)      397  F)
402  A)      401  B)      400  C)       29  H)
```

So 29 reports display one heading inconsistently with the other seven.

### Expected behavior

Any single author-letter is stripped for display, and `splitSections()` reports it. The `)`/`.`/`:` delimiter requirement is what prevents false positives such as `A Recommendation Was Requested` — the narrow `A-G` range is not carrying that weight, so widening the range costs no safety.

### Impact

Cosmetic, but it is the kind of inconsistency that reads as a rendering bug to a user, and the same `A-G` assumption will silently mis-handle any block added past `G` in future.

### CLI tool

Claude Code

### OS and Node.js version

macOS / Node 22

---

I have a fix in progress and will open a PR against this shortly.

Zoubeir23:
Fixes #2324

## Problem

`report-view.tsx` hard-coded the author-letter range in two adjacent places:

```js
.replace(/^\s*(?:Block\s+)?[A-G][).:]\s*/i, "")            // cleanHeading
heading.match(/^(?:Block\s+)?([A-G])[).:\s]/i)             // splitSections
```

`oferta.md:513` and `auto-pipeline.md:64` write `## H) Draft Application Answers`, which falls outside that range — so that one heading renders with its letter still attached while the other seven render clean, and `splitSections()` hands it `letter: null`. In the reporter's 411-report corpus, 29 reports show the inconsistency.

## Fix

The range is now any single letter, and both call sites read one definition.

- **Root cause first:** the two regexes were the same assumption written twice, which is how they drifted. `cleanHeading`, `authorLetter` and `splitSections` moved to `web/src/lib/report-sections.mjs` — plain `.mjs`, the `clean-chips.mjs` / `tracker-table.mjs` pattern, so Node can test the real code and the component imports it rather than re-deriving it.
- **Widening costs no safety.** The `)` / `.` / `:` delimiter is what keeps ordinary prose intact: `"A Recommendation Was Requested"` is untouched before and after. The narrow range was not carrying that weight.
- **No tier change.** `letter` only drives three decisions — the `F` verdict callout, and `A`/`B` staying expanded. An `H` section was collapsed content before and is collapsed content now; only its heading changed. Verified against the current call sites in `report-view.tsx`.
- A block added past `H` is handled the day the core writes it, with no code change.

## Test plan

- [x] `npm test` in `web/` — 72 passing (60 existing + 12 new)
- [x] New `web/test-report-sections.mjs` pins: `H)` stripped and lettered; `I)`/`Z.` handled; all three delimiters plus the spelled-out `Block H:` form; the `(lead)`/`(verdict)` marker; prose left alone without a delimiter; a letter-only heading keeping its text; `splitSections` intro/letter/content output
- [x] The suite lands in `web/tests/lib/`, picked up by the existing `tests/**/*.test.mjs` glob — no `package.json` change needed
- [x] `npx tsc --noEmit` clean, `npm run build` succeeds
- [x] `node test-all.mjs --quick` — 2754 passed, 0 failed

## Also fixed, on review

`authorLetter` accepted whitespace after any letter, so a prose heading like `"A Recommendation Was Requested"` was read as section A and `ReportView` expanded it as if it were Block A. Pre-existing on `main`, but preserving it was a choice rather than a constraint, so it is tightened here: a **bare** letter now needs the same `)` / `.` / `:` delimiter that `cleanHeading` already required.

The whitespace form is kept for the spelled-out grammar only — `modes/oferta.md:47` writes `## Block A — Role Summary`, where nothing follows the letter but a space. Requiring a delimiter there would set `letter: null` and drop the `F` verdict callout and the `A`/`B` expansion on those reports. Both grammars are pinned in the tests.

**The spelled-out form is now stripped for display too.** I first left `cleanHeading("Block A — Role Summary")` alone as a different grammar. CodeRabbit pointed out the resulting inconsistency inside this very PR: `authorLetter` reads that heading as section A, so the file classified it one way and displayed it another, prefix included. After the word `Block` the whitespace form (with an optional dash) is now stripped; a **bare** letter still requires a delimiter, or ordinary prose loses its first word.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved report section recognition for lettered headings, optional “Block” prefixes, and varied delimiter formats.
  * Preserved introductory text and ordinary prose more reliably during report processing.
  * Improved heading cleanup and letter normalization for clearer report display.

* **Tests**
  * Added coverage for heading formats, malformed headings, prose protection, section extraction, and reports without headings.

### DIFF
--- a/web/src/components/report-view.tsx
+++ b/web/src/components/report-view.tsx
@@ -5,6 +5,7 @@ import remarkGfm from "remark-gfm";
 import type { Application } from "@/lib/career-ops";
 import { Badge } from "@/components/ui/badge";
 import { scoreTone, scoreNum, legitimacyTone, parseReport } from "@/lib/format";
+import { cleanHeading, splitSections } from "@/lib/report-sections.mjs";
 import { StatusSelect } from "@/components/status-select";
 import { CompanyLogo } from "@/components/company-logo";
 import { ScoreMethodology } from "@/components/score-methodology";
@@ -14,22 +15,17 @@ import { DeleteFromTracker } from "@/components/delete-from-tracker";
 
 // Progressive disclosure of the report. The core writes prose blocks
 // "## F) Verdict (lead)", "## A) Role Summary", "## B) Match with CV", then
-// C–G + machine artifacts (Machine Summary YAML, Application Answers, submit
-// log). A mainstream user deciding "should I apply?" needs the verdict + fit;
-// the rest is depth-on-demand. We lead with the verdict as a callout, keep A/B
-// expanded, collapse C–G as content, and drop machine artifacts to a dimmer
-// "Technical" tier — and strip the bare "F)" author-letters from headings
-// (native <details>, no client JS — this stays a server component).
-
-type Section = { heading: string; letter: string | null; content: string };
-
-function cleanHeading(h: string): string {
-  const stripped = h
-    .replace(/^\s*(?:Block\s+)?[A-G][).:]\s*/i, "")
-    .replace(/\s*\((?:lead|verdict)\)\s*$/i, "")
-    .trim();
-  return stripped || h.trim();
-}
+// the remaining lettered blocks + machine artifacts (Machine Summary YAML,
+// Application Answers, submit log). A mainstream user deciding "should I
+// apply?" needs the verdict + fit; the rest is depth-on-demand. We lead with
+// the verdict as a callout, keep A/B expanded, collapse the other lettered
+// blocks as content, and drop machine artifacts to a dimmer "Technical" tier —
+// and strip the bare "F)" author-letters from headings (native <details>, no
+// client JS — this stays a server component).
+//
+// Splitting and heading cleanup live in lib/report-sections.mjs so the
+// author-letter range has one definition; duplicating it here is what left
+// "H) Draft Application Answers" rendering with its letter attached (#2324).
 
 // Machine artifacts (collapsed because they're for devs, not the mainstream) vs
 // human content C–G (collapsed only for length) — ux's "honest for devs" tier.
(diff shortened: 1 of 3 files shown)


rubicon:
No competing PR from me here. The extraction to `report-sections.mjs` and the prose false-positive catch are the right fix, better than what I'd have shipped for this. Nice find on the whitespace-only match.

santifer:

`cleanHeading` and `splitSections` hardcode the letter range `[A-G]` in two adjacent regexes, so `## H) Draft Application Answers` (emitted by oferta.md:513 and auto-pipeline.md:64) renders with the letter still stuck on while the other seven come out clean, and it arrives with `letter: null` on top: 29 reports out of a 411-report corpus show it. The interesting move isn't widening the range, it's collapsing both sites into a single definition in `web/src/lib/report-sections.mjs`, on the right argument that what protects ordinary prose is the `)`/`.`/`:` delimiter and not how narrow the range is ("A Recommendation Was Requested" stays intact before and after).

santifer:
Approved — verified both halves of the claim against the real files rather than the description.

**The bug is real and its source is in the core's own modes:**

```
modes/oferta.md:599        ## H) Draft Application Answers
modes/auto-pipeline.md:66  Save in the report as section `## H) Draft Application Answers`
report-view.tsx:28, :62    /^\s*(?:Block\s+)?[A-G][).:]\s*/i
```

So the one heading the pipeline is documented to write falls outside the range the renderer accepts — it shows with its letter still attached while the other seven render clean, and `splitSections()` hands it `letter: null`. 29 reports in a 411-report corpus is a good measurement to bring.

**"Widening costs no safety" — checked, and it holds.** Ran the real module:

```
"H) Draft Application Answers"     → "Draft Application Answers"        letter H
"Block C: Red Flags"               → "Red Flags"                        letter C
"A Recommendation Was Requested"   → "A Recommendation Was Requested"   letter null   ← prose intact
"Machine Summary"                  → "Machine Summary"                  letter null
```

The `)` / `.` / `:` delimiter is what protects ordinary prose, exactly as you argued. The `[A-G]` range was never carrying that weight — it just looked like it was.

**The part that makes this more than a range widening** is the root-cause framing: two regexes encoding one assumption, in adjacent lines, which is *how* they drifted. Moving `cleanHeading` / `authorLetter` / `splitSections` into `report-sections.mjs` means the component imports the rule instead of re-deriving it, and Node can test the real code rather than a copy of it. That's the same `.mjs`-so-tests-can-import-it pattern `tracker-table.mjs` and `clean-chips.mjs` already use, so it fits the house rather than inventing a convention.

web 169/169, typecheck clean, `test-all` green, CI green on all three platforms.

Merging. This is the second of yours today — the other being #2565, the tracker row-shape divergence. Both went after the reason the bug was possible rather than the instance in front of you, which is the harder and more useful thing to do.

PR #2430, opened 2026-08-02
fix(tracker): non-Latin company names all keyed to "", silently overwriting each other

### LINKED ISSUE #2429: merge-tracker: every non-Latin company name normalizes to "", so different companies silently overwrite each other
## What happens

`normalizeCompany()` in `tracker-utils.mjs` strips every character outside `[a-z0-9]`:

```js
export function normalizeCompany(name) {
  return name.toLowerCase().replace(/[^a-z0-9]/g, '');
}
```

So **every company name written in a non-Latin script normalizes to the empty string**, and they all compare equal to each other:

```
"アクメ株式会社"   → ""
"グロベックス合同会社" → ""
"Яндекс"          → ""
"北京字节跳动"      → ""
"شركة"            → ""
"Acme Inc"        → "acmeinc"   ← Latin names are fine
```

`merge-tracker.mjs` uses that key for its fallback match (empty company key + fuzzy role title), so two applications at **different** companies can be treated as the same application and one silently overwrites the other. `applications.md` is gitignored and there is no `.bak`, so the overwritten evaluation is unrecoverable.

## Why it matters

We ship `modes/` for ja, ko, zh, zh-TW, ru, ua, ar and hi. Anyone tracking applications at companies whose names are written in those scripts is exposed the moment two of them share a similar role title.

## Not caught by existing guards

The #1596 cross-channel guard only fires when the company cell is literally `?`. Nothing in the test suite exercises a non-Latin company name through the dedup path.

## Suggested direction

Either require raw-string company equality whenever `normalizeCompany()` returns `''`, or give company the Unicode-aware treatment `normalizeVia()` already has (`tracker-parse.mjs`), which NFKC-folds instead of deleting. Whichever route, a regression test with a non-Latin company name should come with it — that boundary is currently untested.

## Provenance

Surfaced while verifying #2399, which fixes four other silent-loss paths in the same file. That PR doesn't cause this bug, but it does widen its blast radius into the batch path, so the two are worth reading together. Reproduction above was run against `main`.

abankar1:
Closes #2429.

`normalizeCompany()` stripped everything outside `[a-z0-9]`, so **every** company name written in a non-Latin script folded to `''` and compared equal to every other one. `merge-tracker`'s company+role fallback then treated applications at two *different* companies as the same row and silently overwrote one — and `applications.md` is gitignored with no `.bak`, so the losing evaluation was unrecoverable.

Took **route 2** from the issue (Unicode-aware key) rather than the raw-equality special case, as discussed on the thread.

## Fix

This is the #1603 Via bug one column over, so it takes the #1603 remedy: NFKC, case-fold, keep `\p{L}\p{N}` of any script. Rather than copy that expression a second time, I lifted it into a shared `foldIdentityKey()` in `tracker-parse.mjs` that **both** `normalizeVia()` and `normalizeCompany()` delegate to — a second hand-maintained copy is how these two drifted apart in the first place. (`tracker-parse.mjs` imports only `fs`, so no cycle.)

```
"アクメ株式会社"       → "アクメ株式会社"    (was "")
"グロベックス合同会社"   → "グロベックス合同会社" (was "")
"Яндекс"              → "яндекс"          (was "")
"ＡＣＭＥ"             → "acme"            (was "")   ← NFKC
"Acme Inc."           → "acmeinc"         unchanged
"?"                   → ""                unchanged  ← #1596 depends on this
```

Because the key is shared, this also repairs the other five call sites: the scan blacklist (a blacklisted non-Latin company matched *every* other non-Latin company), `set-status`/`outcome` selector resolution, and `company-history` grouping.

## Tests

Unit + end-to-end, in `test-all.mjs`. Verified they fail against the old implementation:

```
❌ non-Latin company keys collapsed: ["","","",""]
❌ full-width company name did not fold to its half-width spelling
❌ a non-Latin company was silently overwritten by a different one
❌ expected 2 tracker rows after merge, got 1        ← the data loss itself
✅ punctuation-only company still folds to the empty key (#1596 guard intact)
✅ Latin company keys are unchanged
```

`node test-all.mjs --quick` → **2787 passed, 0 failed**.

## Two behavioural deltas, reported rather than shipped quietly

As promised on the issue:

**1. `company-history` no longer treats non-Latin companies as a data-quality defect.** Both suites used a non-Latin name *as* the "unjoinable" fixture, precisely because it folded to `''`. That conflated "a name this tool cannot key" with "a name written in another script". Those companies now get real cards, appear in `metadata.companies`, and show up in `hygiene.agedApplied` when they've gone silent — where they were previously invisible. I retargeted the fixtures onto `?` (punctuation-only, genuinely keyless) in a **separate first commit** that is green both before and after the fix, so the fixture correction and the behaviour change are reviewable apart.

**2. Accented Latin names now keep their accents.** `Café` keys as `café` (was `caf`). Strictly more correct — `Café` and `Caf` are no longer the same company — but it is a change, and I'd rather you see it than find it. Accent-*folding* (so `Café` and `Cafe` match) is a separate question I haven't touched.

3 commits (fixture decoupling → fix → regressions), each green on its own.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved company-name matching for non-Latin names while keeping distinct companies separate.
  - Punctuation-only names continue to be treated as unknown or unjoinable.
  - Standardized equivalent full-width and ASCII Latin spellings.
  - Preserved existing Latin-name normalization behavior.

- **Tests**
  - Added regression coverage for Unicode names, punctuation-only entries, and merge tracking.
  - Updated data-quality and multi-source scenarios to reflect corrected matching behavior.

### DIFF
--- a/company-history.mjs
+++ b/company-history.mjs
@@ -581,12 +581,15 @@ async function runSelfTest() {
   const NOW = new Date('2026-07-09T00:00:00Z');
   const row = (num, company, status, date, notes = '') => ({ num, date, company, role: 'Engineer', score: '4/5', status, pdf: '✅', report: `reports/${num}.md`, notes });
 
-  // --- join fixtures: case/punct variants meet under one key; "" excluded -> unjoinable ---
+  // --- join fixtures: case/punct variants meet under one key; only a genuinely
+  // keyless company is unjoinable. A non-Latin name is NOT keyless (#2429):
+  // normalizeCompany() folds script-preserving, so 株式会社 keys as itself and
+  // gets a real card instead of being discarded as unparseable data. ---
   {
     const rows = [
       row(1, 'Acme Inc.', 'Applied', '2026-01-01'),
       row(2, 'ACME, INC', 'Applied', '2026-01-02'),
-      row(3, '株式会社', 'Applied', '2026-01-03'), // normalizeCompany -> "" (non a-z0-9)
+      row(3, '?', 'Applied', '2026-01-03'), // punctuation only -> genuinely keyless
     ];
     const result = buildCompanyCards(
       { trackerRows: rows, followupRows: [], repostClusters: [], sourcesLoaded: { tracker: true, followups: false, scanHistory: false, statusLog: false } },
@@ -594,7 +597,7 @@ async function runSelfTest() {
     );
     check(result.companies.length === 1, 'case/punct company variants join under one key');
     check(result.companies[0].responsiveness.facts.length === 2, 'both joined rows contribute facts to the single card');
-    check(result.dataQuality.unjoinable === 1, 'empty-key company (non-Latin, strips to "") is excluded and counted as unjoinable');
+    check(result.dataQuality.unjoinable === 1, 'a punctuation-only company key is excluded and counted as unjoinable');
   }
 
   // --- label goldens ---
--- a/company-history.mjs
+++ b/company-history.mjs
@@ -600,6 +600,21 @@ async function runSelfTest() {
     check(result.dataQuality.unjoinable === 1, 'a punctuation-only company key is excluded and counted as unjoinable');
   }
 
+  // --- non-Latin companies are first-class, and distinct from each other (#2429) ---
+  {
+    const rows = [
+      row(4, 'アクメ株式会社', 'Applied', '2026-01-01'),
+      row(5, 'グロベックス合同会社', 'Applied', '2026-01-02'),
+      row(6, 'Яндекс', 'Applied', '2026-01-03'),
+    ];
+    const result = buildCompanyCards(
+      { trackerRows: rows, followupRows: [], repostClusters: [], sourcesLoaded: { tracker: true, followups: false, scanHistory: false, statusLog: false } },
+      { now: NOW, silenceWindowDays: 28 },
+    );
+    check(result.companies.length === 3, 'three different non-Latin companies produce three cards, not one');
+    check(result.dataQuality.unjoinable === 0, 'a non-Latin company name is joinable, not a data-quality defect');
+  }
+
   // --- label goldens ---
   {
     // silent-on-you: one Applied row well past the window, nothing else.
--- a/company-history.test.mjs
+++ b/company-history.test.mjs
@@ -90,8 +90,11 @@ console.log('\n--- 1. multi-source 3-company scenario ---');
     row(1, 'Responsive Corp', 'Rejected', '2026-06-01'),
     // Company B: Silent Systems — old Applied row, applied date from notes.
     row(2, 'Silent Systems', 'Applied', '2026-03-01', 'Applied 2026-03-10'),
-    // Company C: unjoinable non-Latin name -> excluded from companies, counted unjoinable.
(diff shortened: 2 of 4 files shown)


santifer:
Everything checks out on substance, @abankar1 — true source fix in `tracker-utils.normalizeCompany` covering exactly the four consumers #2429 was re-scoped to. One rebase needed, and it's a duplicate-extraction race, not a flaw in your work: #2445 (merged this morning) already extracted the identical shared Unicode fold from `normalizeVia` as **`normalizeTextKey`**, while your branch extracted it as `foldIdentityKey` — one-hunk conflict in `tracker-parse.mjs`.

The resolve direction matters because the two folds differ semantically: `normalizeTextKey` **keeps `\p{M}` combining marks**, your `foldIdentityKey` drops them. Delegating the other way would regress `normalizeVia` and false-merge Hindi/Arabic names differing only in marks — the exact class your own fix hunts (hi and ar are shipped locales).

So the rebase recipe is: drop `foldIdentityKey`, and delegate `normalizeCompany` → `normalizeTextKey`. Your four-consumer coverage and tests stay as they are. Say the word if you'd rather I do the mechanical resolve maintainer-side with your authorship intact.

abankar1:
Rebased exactly as you specified — and thank you for catching the direction, because you were right about the semantics and my fold was the worse of the two.

`foldIdentityKey` is gone; `normalizeCompany` now delegates to **`normalizeTextKey`**. `tracker-parse.mjs` is no longer touched at all by this branch, so the conflict is gone rather than resolved.

I verified your `\p{M}` point rather than taking it on trust, because if you were right my original would have shipped the same bug into hi/ar that it removes from ja/zh/ru. It would have:

```
MY original foldIdentityKey (drops \p{M}):
  "कंपनी" vs "कपनी"  => COLLIDE: true -> "कपन"
  "क"     vs "का"    => COLLIDE: true -> "क"

normalizeTextKey (keeps \p{M}):
  "कंपनी" vs "कपनी"  => distinct
  "क"     vs "का"    => distinct
  "شركة"  vs "شرکة"  => distinct
```

So the duplicate-extraction race was lucky: two people extracted the same fold and only one of them got Devanagari right.

I've pinned that as a regression in the same block, so the delegation can't later be pointed at a mark-stripping fold and quietly undo it:

> ✅ company keys keep combining marks, so Devanagari names differing only in matras stay distinct (#2429)

Everything else is unchanged — four-consumer coverage, the `?`/#1596 preservation, the NFKC and Latin-unchanged assertions, and the end-to-end merge-tracker case that fails on old code as *"expected 2 tracker rows after merge, got 1"*.

`node test-all.mjs --quick` → **2921 passed, 0 failed** · `company-history --self-test` → 45/0 · `company-history.test.mjs` → 86/0.

Four commits, linear, force-with-lease pushed. No maintainer-side resolve needed — but thanks for offering.

santifer:
Route 2 executed exactly as designed, @abankar1: the Unicode-aware fold reusing the normalizeVia treatment, with the #1596 guard intact and NFKC folding pinned in tests, is the complete fix. アクメ and Яндекс can finally coexist in one tracker. Merged. 🚀

PR #2431, opened 2026-08-02
refactor(web): derive the cadence baseline from the core, delete the hand-copy (#2369 fix 1)

abankar1:
Closes half of #2369 — **fix 1** (`CADENCE_DEFAULTS` → derivation). Fix 2 (`doctorState()` prereqs) is deliberately left alone; see below.

## The open question, answered

You asked whether the form wants *pure defaults* or *effective config*. **It needs both, and they aren't interchangeable.**

`cadenceConfig` is `{ ...DEFAULT_CADENCE, ...loadProfileCadence() }` — effective. The settings form uses its baseline as the **placeholder** ("leave blank and you get 7"). Source that from `cadenceConfig` and the moment a user sets `applied_first_days: 3`, the form renders `3` as the default — so clearing the field to return to the default shows them their own override as the thing they're returning to. The override and the value it overrides become indistinguishable in the one UI whose job is telling them apart.

So: **emit both from the core.**

## Changes

1. **Core** — `followup-cadence.mjs --json` now emits `cadenceDefaults: DEFAULT_CADENCE` next to `cadenceConfig`. Purely additive; every existing consumer untouched.
2. **Web** — `/api/followups/cadence` reads the pure defaults from the core, exactly the way `/api/followups` already *"reads its verdict"* rather than reimplementing the cadence logic. It's an API route, not a server component, so spawning is the established mechanism. Only the `_days` key mapping is translated (`applied_first` → `applied_first_days`, except `applied_max_followups`); no values are restated.
3. **`CADENCE_DEFAULTS` deleted** from `web/src/lib/followups.ts`, with a comment saying not to reintroduce it.
4. **55.3c retargeted.** The old assertion cross-checked the copy against the core; that copy is gone, so it had nothing left to compare and would have stayed green forever — a check that no longer guards anything is its own small lie, the kind this issue is about. It now guards the two things that *can* break: the core must keep emitting `cadenceDefaults`, and the web must not reintroduce a local table.

## No fallback copy

When the core is missing or unparseable the route answers `defaultsAvailable: false` and the form renders empty placeholders rather than inventing numbers. I deliberately did **not** add a hardcoded fallback table in the `catch` — a fallback copy is the same copy in disguise, and drifts the same way `states.ts`'s `FALLBACK` did in #2282.

## Verification

```
✅ followup-cadence.mjs --json still emits cadenceDefaults for the web to derive from (#2369)
✅ web/src/lib/followups.ts keeps no hand-copied cadence table (#2369)
✅ web doctorState prereqs match doctor.mjs USER_LAYER_PREREQS (#2369)   ← 55.3d, untouched
```

`node test-all.mjs --quick` → **2782 passed, 0 failed**. `npm test` (web) → **60 passed, 0 failed**.

Core payload confirmed through `analyzeFromContent`:

```
keys: [ 'metadata', 'entries', 'cadenceConfig', 'cadenceDefaults' ]
```

## Fix 2 left for its own PR

`doctorState()`'s prereq list stays as-is. 55.3d covers it, and the fast-path reason you documented (server components can't `execFile doctor` per render) makes its derivation a genuinely different design problem — it needs a caching story, not a spawn. Smuggling that into this PR would bury it. Happy to take it separately if you want it.

3 commits (core emit → web derive → test retarget).

## Summary by CodeRabbit

* **Improvements**
  * Follow-up cadence settings now use defaults provided dynamically by the server.
  * Settings accurately reflect configured values and leave unavailable defaults blank instead of displaying invented values.
  * The cadence API now indicates whether defaults are available and combines them with profile-specific overrides.

### DIFF
--- a/followup-cadence.mjs
+++ b/followup-cadence.mjs
@@ -624,6 +624,13 @@ export function analyzeFromContent(trackerContent, followupsContent = '') {
     },
     entries: filtered,
     cadenceConfig: CADENCE,
+    // The EFFECTIVE cadence above is defaults+profile overrides. Consumers that
+    // need to show what a value would be WITHOUT the user's override (the web
+    // settings form's placeholder) need the pure defaults too — sourcing that
+    // placeholder from cadenceConfig would render a user's own override as the
+    // default they'd be reverting to. Emitting both is what lets the web stop
+    // hand-copying DEFAULT_CADENCE (#2369).
+    cadenceDefaults: DEFAULT_CADENCE,
   };
 }
 
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -11437,11 +11437,38 @@ try {
   // reintroduce a local table (a fallback copy drifts exactly like the original
   // did — states.ts FALLBACK, #2282).
   {
-    const cadSrc = readFileSync(join(ROOT, 'followup-cadence.mjs'), 'utf-8');
-    if (/cadenceDefaults:\s*DEFAULT_CADENCE/.test(cadSrc)) {
-      pass('followup-cadence.mjs --json still emits cadenceDefaults for the web to derive from (#2369)');
+    // Assert the EMITTED payload, not just the source text: a regex over the
+    // source proves the literal is present, not that the contract the web
+    // parses is intact. analyzeFromContent() is the same code path --json
+    // prints, driven from strings so it needs no tracker on disk.
+    const { analyzeFromContent, DEFAULT_CADENCE } = await import(pathToFileURL(join(ROOT, 'followup-cadence.mjs')).href);
+    const emitted = analyzeFromContent(
+      '# Applications Tracker\n\n| # | Date | Company | Role | Score | Status | PDF | Report | Notes |\n' +
+      '|---|------|---------|------|-------|--------|-----|--------|-------|\n' +
+      '| 1 | 2026-06-01 | Acme | Engineer | 4.2/5 | Applied | ❌ | [1](../reports/001-acme.md) | t |\n',
+      '',
+    );
+    const defaults = emitted?.cadenceDefaults;
+    const cadKeys = Object.keys(DEFAULT_CADENCE);
+    const schemaOk = defaults && typeof defaults === 'object'
+      && cadKeys.length > 0
+      && cadKeys.every((k) => Number.isInteger(defaults[k]) && defaults[k] >= 0)
+      && Object.keys(defaults).length === cadKeys.length;
+    if (schemaOk) {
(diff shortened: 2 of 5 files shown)


abankar1:
CodeRabbit's finding was right on both halves — taken.

**1. The contract is now all-or-nothing, with no coercion.** `Number.parseInt` accepted `"3.5"` as `3` and `"7days"` as `7`, and the per-key filter reported `defaultsAvailable: true` off a single valid key. Either path could show the settings form a baseline the core never emitted — which is exactly the silent-wrong-value this derivation exists to remove, so it would have been an unusually bad thing to leave in *this* PR.

Now every `PROFILE_CADENCE_KEYS` entry must be a real non-negative integer (`typeof number` + `Number.isInteger`, not a parse), and any miss returns `null` → `defaultsAvailable: false` → empty placeholders. A malformed payload degrades to "no defaults" rather than to a plausible wrong one.

**2. The test asserts the emitted payload, not the source text.** The regex only proved the literal `cadenceDefaults:` was present — not that the contract the web parses is intact. It now drives `analyzeFromContent()`, the same code path `--json` prints, from strings (so it needs no tracker on disk) and checks the full schema:

```
✅ followup-cadence --json emits a complete integer cadenceDefaults for the web to derive from (#2369)
✅ every web PROFILE_CADENCE_KEY maps onto a core cadenceDefaults key (#2369)
```

The second one is the guard I'd actually been missing: the `_days` mapping is the one place this PR translates anything, so it's the one place a rename could silently break the derivation.

`node test-all.mjs --quick` → **2783 passed, 0 failed** · `npm test` (web) → **60 passed, 0 failed** · `tsc --noEmit` clean.

The `ast-grep` note about `child_process` on the same comment I've left alone: the `execFile` call takes a fixed argv array (`[script, "--json"]`) with no interpolated input, and it mirrors how `/api/followups` and `/api/doctor` already invoke the core.

santifer:
This is the right shape, @abankar1 — and it closes the first of the three instances #2369 catalogued.

Deleting the hand-copied `CADENCE_DEFAULTS` and having the web read `cadenceDefaults` from the core's own `--json` payload is derivation, not synchronisation: there's no longer a second copy that can drift, so there's nothing to keep in step. The all-or-nothing contract is the part I'd have asked for if it weren't already there — a malformed payload sets `defaultsAvailable:false` instead of inventing numbers, which means a broken read shows up as "I don't know" rather than as plausible wrong defaults.

Timing note you'll appreciate: today the same class closed on the company-matching key (#2666 → #2668), and there's now a parity assertion in the required suite for the copy that genuinely can't be removed. Yours removes the copy entirely, which is strictly better. Merged! 🚀

PR #2434, opened 2026-08-02
fix(batch-runner): tab-IFS collapse records score "-" for every successful offer

anujad12:
## The bug

`batch-runner.sh` serialises the worker's final JSON into three fields and reads them back with `read -r parsed_status parsed_error parsed_score`. Joined and split on **TAB**, the common path breaks.

A *successful* worker has `status="completed"`, `error` **empty**, `score=3.4`:

```
emitted:  completed \t \t 3.4
```

Tab is IFS **whitespace**, so bash collapses runs of it and strips leading/trailing occurrences. The two tabs around the empty `error` collapse into one:

```
parsed_status="completed"   parsed_error="3.4"   parsed_score=""
```

The `elif [[ -n "$parsed_score" ]]` branch never fires, `score` keeps its `"-"` default, and **every successful offer is recorded unscored**.

## Reproduction

Independent of the runner — this is the same emitter, lifted out:

```bash
$ json='{"status":"completed","error":null,"score":3.4}'
$ parsed=$(printf '%s' "$json" | node -e '
    let d="";process.stdin.on("data",x=>d+=x);
    process.stdin.on("end",()=>{const o=JSON.parse(d);
      process.stdout.write((o.status||"")+"\t"+(o.error||"")+"\t"+(typeof o.score==="number"?String(o.score):""));});')

$ printf '%s' "$parsed" | cat -A
completed^I^I3.4

$ IFS=$'\t' read -r s e sc <<< "$parsed"
$ echo "status='$s' error='$e' score='$sc'"
status='completed' error='3.4' score=''      # score landed in error
```

## Why it is worth fixing

It is silent and internally consistent: exit code 0, status `completed`, a real report on disk, and a score column that is uniformly useless. Nothing in the run summary surfaces it — you notice only when scores stop driving whatever consumes them downstream.

It affects the **success** path specifically, because that is the path where `error` is empty. A *failed* worker (non-empty error) splits correctly, which is why the existing behaviour looks fine whenever you go looking at failures.

## The fix

Join and split on `\x1f` (ASCII US) instead. A non-whitespace delimiter gets one-field-per-unit splitting with empty fields preserved, which is what this code actually needs. Emitter and reader are changed together.

Two lines of real change; the rest is a comment explaining the hazard so it does not get "tidied" back to `\t`.

## Test

`tests/batch-runner-score-delimiter.test.mjs` (auto-discovered).

It **extracts the real emitter and reader out of `batch/batch-runner.sh`** and runs them, rather than restating them, so the test and the code cannot drift apart. It asserts:

- emitter and reader use the same delimiter
- the delimiter is not IFS whitespace
- a successful worker records `3.4`, not `-`
- the empty `error` field stays empty instead of absorbing the score
- `status` still parses

Verified in both directions:

| tree | result |
|---|---|
| `upstream/main` (unpatched) | **3 assertions fail** — `score was recorded as "-" instead of 3.4`, `error absorbed the next field: "3.4"`, `delimiter "\t" is IFS whitespace` |
| with this patch | all 5 pass |

Full suite, run in a clean worktree so no local state could influence it:

| tree | result |
|---|---|
| `upstream/main` | 2756 passed, 0 failed |
| `upstream/main` + this PR | 2761 passed, 0 failed |

`+5` is exactly this PR's new assertions; nothing else moves.

## Notes

Found while merging upstream into a fork. Happy to adjust the delimiter choice, the comment, or the test's shape — and to split the test into its own PR if you'd rather take the one-line fix on its own.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved batch result parsing to preserve empty error fields and correctly retain status and score values.

* **Tests**
  * Added end-to-end coverage verifying reliable parsing when the error field is empty.

### DIFF
--- a/batch/batch-runner.sh
+++ b/batch/batch-runner.sh
@@ -658,14 +658,21 @@ process_offer() {
             const status = typeof obj.status === "string" ? obj.status : "";
             const error = typeof obj.error === "string" ? obj.error : "";
             const score = typeof obj.score === "number" ? String(obj.score) : "";
-            process.stdout.write(status + "\t" + error + "\t" + score);
+            process.stdout.write(status + "\x1f" + error + "\x1f" + score);
           } catch {
             process.stdout.write("");
           }
         });
       ' 2>/dev/null || true)
       if [[ -n "$parsed" ]]; then
-        IFS=$'\t' read -r parsed_status parsed_error parsed_score <<< "$parsed"
+        # \x1f (US), not \t: tab is IFS *whitespace*, so bash collapses runs of it
+        # and strips leading/trailing ones. On the common path -- a worker that
+        # succeeded, so `error` is empty -- the two tabs around that empty field
+        # collapse into one, `score` slides into parsed_error, and parsed_score
+        # comes back empty. The `elif` below then never fires and every
+        # successful offer records score "-". A non-whitespace delimiter gets
+        # one-field-per-unit splitting with empty fields preserved.
+        IFS=$'\x1f' read -r parsed_status parsed_error parsed_score <<< "$parsed"
         if [[ "$parsed_status" == "failed" ]]; then
           worker_failed_match="failed"
           worker_error_match="$parsed_error"
--- a/tests/batch-runner-score-delimiter.test.mjs
+++ b/tests/batch-runner-score-delimiter.test.mjs
@@ -1,0 +1,111 @@
+// tests/batch-runner-score-delimiter.test.mjs — the worker payload is split on a
+// delimiter that preserves empty fields.
+//
+// THE BUG THIS PINS
+//
+// batch-runner.sh serialises the worker's final JSON into three fields and reads
+// them back with `read -r parsed_status parsed_error parsed_score`. When those
+// fields were joined and split on TAB, the common path broke:
+//
+//   a successful worker has status="completed", error EMPTY, score=3.4
+//   emitted:  completed \t \t 3.4
+//
+// Tab is IFS *whitespace*, so bash collapses runs of it and strips leading and
+// trailing occurrences. The two tabs around the empty `error` collapse into one:
+//
+//   parsed_status="completed"  parsed_error="3.4"  parsed_score=""
+//
+// The `elif [[ -n "$parsed_score" ]]` branch then never fires, and EVERY
+// successful offer records score "-".
+//
+// It is silent and self-consistent: exit code 0, status `completed`, a real
+// report on disk, and a score column that is uniformly useless. Nothing in the
+// run summary shows it.
+//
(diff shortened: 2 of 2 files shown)


santifer:
Welcome, @anujad12! Real silent bug on the SUCCESS path (tab is IFS whitespace, so every successful offer logged score "-" and the run average stayed empty), minimal unit-separator fix — and the detail that elevates it: your test extracts the real delimiter from batch-runner.sh itself, so emitter and reader can never drift apart. Merged — your first contribution. 🚀

anujad12:
Thanks @santifer Happy to help. I have been using this repo quite a bit for my own job search and in fact, I have added tons of new additions to my own fork of the repo which I would love to contribute back upstream. Let me know what's the best way to make that happen? Do I just send a new PR with all my additions and give some context to it there?

anujad12:
Thanks for merging! I read CONTRIBUTING before answering my own question, and it routes the additions pretty cleanly — so rather than one big PR I've split them by the door each one belongs at.

**Straight to PR (bug fix, no issue needed):**

- **#2491** — `tracker-columns-tests.mjs` fixture isolation. `makeSandbox` doesn't isolate `reports/`, so the `verify-pipeline` child scans the real one. Measured on an install with 311 reports: **3 of 8 runs fail on `main`, 0 of 8 with the fix**, with a *varying* failure count (2, 4, 1) — load-dependent, not deterministic. CI can't see it because CI has no reports, which is exactly why it survived.

**Issues first, because they're bigger than a fix:**

- **#2492** — `batch-runner.sh` hands workers an **empty** JD file (`mktemp` and never written), so they fall through to WebFetch and correctly refuse to fabricate on JS-rendered boards. **35 of 251 offers failed this way on one run.** I have a JD-cache + per-ATS-fetch fix that took it 35 → 3 (the 3 were genuinely dead postings), but it adds two scripts and a cache dir, so I'd rather ask about scope and shape than drop it on you. The minimal runner-only fix is a couple of lines and stands alone if you'd prefer that.
- **#2493** — a USAJOBS.gov provider. Not zero-auth (free self-service API key + registered email as User-Agent), so per CONTRIBUTING it needs discussion before a PR. Feels like the US sibling of your Interamt scanner, but the credential requirement is a genuine departure and might make it plugin-shaped instead.

**Not proposing** the rest, for what it's worth: the application-bundle chain I built overlaps `application-artifacts.mjs` / `prepare-application.mjs` / `batch-tailor.mjs`, which you already have — duplicating that would fail the "does it match the project's shape" question. And the fork-specific automation (a daily cron loop, an upstream-sync helper, artifact pruning) is mine to maintain, not core's.

For context on where these come from: I've been running career-ops for my own search since late July — ~300 evaluations, a tracker with 290 applications, and one interview so far. Everything above is something that broke or bit me in a real run rather than something I went looking for.

PR #2435, opened 2026-08-02
fix(plugins): isolate cmdRun results to requested id; gate gmail cursor on dryRun

### LINKED ISSUE #2354: plugins.mjs run <id> aggregates every enabled ingest plugin under the named one; gmail plugin persists its cursor even under --dry-run
## Summary

Two related bugs in the bundled plugin engine, found while dry-run comparing the `linkedin-alerts` plugin against another sourcing path:

1. `node plugins.mjs run <id>` for a hook kind that multiple plugins expose (e.g. `ingest`) silently aggregates results from *every* enabled plugin of that kind, not just the one named — but reports the combined total under the requested plugin's name.
2. The bundled `gmail` plugin persists its processed-message cursor (`data/gmail-state.json`) even when invoked with `--dry-run`, so a dry-run permanently marks Gmail messages as "seen" even though nothing was written to the pipeline for them.

Together these mean: running `node plugins.mjs run linkedin-alerts --dry-run` with both `gmail` and `linkedin-alerts` enabled in `config/plugins.yml` silently consumes the `gmail` plugin's real ingest cursor as a side effect, permanently hiding those messages' leads from all future real runs — while the CLI output misattributes `gmail`'s results to `linkedin-alerts` the whole time, making the bug easy to miss.

## Repro 1 — misattribution

With both `gmail` and `linkedin-alerts` enabled in `config/plugins.yml`:

```
node plugins.mjs run linkedin-alerts --dry-run
```

**Expected:** only `linkedin-alerts`' own ingest results.
**Actual:** `cmdRun` calls `runHook('ingest', …)`, which runs every enabled plugin exposing the `ingest` hook (both `gmail` and `linkedin-alerts`), then flattens all their results together without filtering by the requested plugin id. `gmail`'s results are silently merged in and printed as if they belonged to `linkedin-alerts`.

## Repro 2 — dry-run isn't dry

Same setup, run `node plugins.mjs run gmail --dry-run` (or trigger it indirectly via Repro 1).

**Expected:** no state change.
**Actual:** `plugins/gmail/index.mjs`'s `ingest()` calls `saveProcessedIds(processedIds)` unconditionally, persisting `data/gmail-state.json`'s processed-message cursor regardless of `ctx.dryRun`. Any Gmail message seen during a dry-run is marked "processed" and silently skipped by all future real runs — including messages whose extracted URLs were never written to the pipeline. Concretely, in our repro this consumed 3 authentic Xing digest alert emails (227 extractable job-lead URLs total) that never made it into `data/pipeline.md`.

Not fork-specific — this affects any career-ops install running `gmail` alongside any other ingest plugin.

## Suggested fix (mechanical, ~2 lines)

- `plugins.mjs`'s `cmdRun`: filter the `runHook(...)` results down to `r.id === id` before flattening, instead of aggregating every enabled plugin of that hook kind under the one the caller asked for.
- `plugins/gmail/index.mjs`: gate the `saveProcessedIds(processedIds)` call behind `!ctx.dryRun` — `ctx.dryRun` is already surfaced by `buildCtx` and already used this way by `plugins/notion/index.mjs`.

We applied both locally and verified against the full `test-all.mjs` suite (2662 passed; the 2 pre-existing unrelated failures reproduce identically on unmodified `main`).

Tanmay9223:
## What does this PR do?

Fixes two bugs in the plugin engine reported in #2354:

1. **Misattribution (`cmdRun`):** `runHook()` runs _every_ enabled plugin of the requested hook kind and returns `[{id, ok, result}]` for all of them. `cmdRun` never filtered by the requested plugin `id` before flattening results, so running `node plugins.mjs run linkedin-alerts --dry-run` with `gmail` also enabled silently merged gmail's results in and reported them all under the `linkedin-alerts` name.

2. **Dry-run not dry (`gmail`):** `plugins/gmail/index.mjs` called `saveProcessedIds()` unconditionally — ignoring `ctx.dryRun`. A `--dry-run` permanently marked Gmail messages as processed even though nothing was written to the pipeline. Any message seen during a dry-run was hidden from all future real runs. `plugins/notion/index.mjs` already handles this correctly with `if (ctx?.dryRun) { ... continue; }`.

## Related issue

Fixes #2354

## Type of change

- [x] Bug fix

## Checklist

- [x] I ran `node test-all.mjs` and all tests pass (2783 passed, 1 pre-existing failure on `main` unchanged)

## Summary by CodeRabbit

* **Bug Fixes**
  * Plugin searches, ingestion, exports, and notifications now return only results for the requested plugin.
  * Gmail dry runs no longer save processed message IDs, preventing preview runs from changing ingestion state.

* **Tests**
  * Added regression coverage for plugin result isolation and Gmail dry-run behavior.


santifer:
Sorry for the slow first reply, @Tanmay9223.

Filters what `cmdRun` gets back by the requested id before flattening (runHook returns every enabled plugin on that hook) and gates gmail's `saveProcessedIds()` on `ctx.dryRun`, citing `plugins/notion/index.mjs` as the plugin that already did it right, which is the tell that you read the rest of the engine and not just the broken file. 73/5 lines across plugins.mjs, `_engine.mjs`, gmail and test-all.mjs, MERGEABLE, and it reports 2783 tests green with a single pre-existing failure from main left untouched.

Tanmay9223:
Thanks for the update! Take your time with the review—I'll stand by for your decision or in case you need any further adjustments from my end once you've run the suite.

santifer:
Sorry for the wait, @Tanmay9223 — this one is mine to decide, unlike your #2082.

The first bug is the one that matters:

> **Misattribution (`cmdRun`):** `runHook()` runs *every* enabled plugin of the requested hook kind and returns all their results

A command that asks for one plugin's result and receives everyone's is a correctness bug with a security-shaped edge: the caller believes it's looking at plugin X's output and is looking at plugin Y's. In a system where plugins are third-party by design, attributing one plugin's output to another is not a cosmetic mix-up.

Queued — not merged tonight because I hit the per-session merge ceiling. Two things I'll check when it's read, flagged now:

1. Whether the isolation is by **requested id** or by position/order — the second works until two plugins register the same hook kind, which is the case the bug already proves happens.
2. The gmail cursor gate: whether an ungated cursor was reading state it shouldn't, or just doing redundant work. Those have very different urgencies and the PR title puts them in one line.

Tanmay9223:
Thanks for the review! I've addressed the feedback by pushing commit `7938087`:

1. **Pre-Invocation Filtering:** `loadPlugins()` and `runHook()` in `plugins/_engine.mjs` now accept `pluginId` to filter plugin manifests before any hook is imported or executed.
2. **`cmdRun` Integration:** All hook dispatch branches in `plugins.mjs` (`ingest`, `search`, `export`, `notify`) now pass `pluginId: id` to `runHook()`.
3. **Regression Tests:** Expanded `test-all.mjs` with an end-to-end multi-plugin isolation test ensuring unrelated hooks are never called.
4. **Rebased/Merged:** Brought in the latest changes from `main`.

All checks are green and ready for review!

santifer:
The isolation fix and the Gmail dry-run gating in 7938087 both check out, @Tanmay9223: pre-invocation filtering is the right layer for it. One small move before merge: the new #2354 assertions were added as an inline test-all.mjs section, and the file's header (test-all.mjs:21-33) asks that new tests live in their own `tests/**/*.test.mjs` file instead: they get auto-discovered, no section number, and your future PRs stop colliding with everyone else editing the monolith. Ten-minute change, and then this merges.

Tanmay9223:
I've moved the #2354 regression tests out of `test-all.mjs` and into a new standalone `tests/plugin-run-isolation-and-gmail-dryrun.test.mjs` file, in accordance with the repository conventions. The changes have been committed and pushed to the branch. Let me know if there's anything else!

santifer:
Solid follow-through, @Tanmay9223. The double layer (manifest pre-filter + result filter) means `run <id>` can no longer execute or report a neighbor plugin, and the gmail dry-run no longer persists processed ids — plus the regression suite now lives standalone in tests/ where the runner auto-discovers it. Merged! 🚀

PR #2437, opened 2026-08-02
fix(workable): use the public widget API so large accounts are scanned

omarfsahmed:
## What does this PR do?

`providers/workable.mjs` silently returned **zero jobs** for any Workable account large enough to trigger Workable's summary view — the bare `/<slug>/jobs.md` feed renders a department roll-up instead of the job table, so there are no rows to parse. This switches the provider to Workable's public, no-auth **account widget API**, which returns the account's full posting list in one request, and keeps the markdown feed as a fallback.

Found live: a tracked company with 259 open postings was reported as `live but empty` on every scan. After the fix the same account contributes its full listing (34 postings survived my `title_filter`/`location_filter`).

## Related issue

None — the template exempts bug fixes and scanner providers from issue-first, so sending it straight in.

## Type of change

- [x] Bug fix
- [ ] New feature

## Why the markdown feed can't be salvaged

I tried to page around it before replacing it:

- **Hard-capped at 30 rows**, and it honours no pagination parameter — `page`, `offset`, `limit` and `startrow` all return the same first 30 rows.
- **Segmenting by `?department=` is incomplete.** The departments listed in the summary sum to *less* than the account total (237 of 259 on the account I tested), so 22 postings are unreachable that way — and a department with more than 30 roles is itself truncated, which one was.

## The replacement

```
GET https://apply.workable.com/api/v1/widget/accounts/<slug>?details=true
→ { name, description, jobs: [ { title, shortlink, url, department, city,
    state, country, telecommuting, published_on, description, … } ] }
```

No auth, no pagination, complete. It also carries two fields the markdown feed never had:

- `description` → so `scan.mjs`'s `content_filter` now works for Workable companies
- `published_on` → so the recency filtering in `scan-ats-full.mjs` now works too (it dropped 34 stale postings on my first run, which it previously could not see)

The markdown feed remains as a fallback when the API call fails, so existing setups degrade rather than break.

## Security

Existing properties are preserved on **both** paths — the `apply.workable.com` host allowlist, `redirect: 'error'` against SSRF-by-redirect, and per-job URL re-validation before a job is emitted.

One property is added: the account slug is now charset-guarded (`/^[A-Za-z0-9][A-Za-z0-9_-]*$/`) before it is interpolated into the API URL, so a crafted `careers_url` can't escape the path.

## Tests

`detect()` now reports the widget API URL it actually fetches; the existing assertion is updated to match. Seven cases added for the new path:

- location formatting (`city, country`, and `Remote` for telecommuting entries)
- `published_on` → `postedAt`
- HTML tag and `<script>` body stripping from descriptions
- off-domain / non-HTTPS permalink rejection
- `null` and `jobs`-less payloads
- widget API used as primary (markdown feed untouched when it succeeds)
- fallback to the markdown feed when the API throws

**`node test-all.mjs`: 2754 passed, 1 failed.** In the interest of being straight about it, the failure is *not* from this change:

```
❌ analyze() appDateSource end-to-end check crashed: EPERM: operation not
   permitted, symlink 'C:\...\node_modules' -> 'C:\...\co-cadence-e2e-XXXX\node_modules'
```

That's `followup-cadence`'s e2e test symlinking `node_modules`, which Windows blocks without Developer Mode or elevation. I stashed this branch's changes and reproduced the identical failure on a clean `main`, so it's a pre-existing platform limitation. Everything else, including the full provider suite, passes.

## Checklist

- [x] I ran `node test-all.mjs` and all tests pass — *with the one pre-existing Windows-only failure documented above*
- [x] My changes respect the [Data Contract](https://github.com/santifer/career-ops/blob/main/DATA_CONTRACT.md) (no modifications to user-layer files — only `providers/` and `tests/`)

## Summary by CodeRabbit

* **New Features**
  * Improved Workable job listing retrieval using the public widget API.
  * Added automatic fallback to the legacy feed when widget data is unavailable.
  * Job listings now include validated locations, publication dates, deduplicated results, and plain-text descriptions.

* **Bug Fixes**
  * Improved validation of Workable account links and job URLs.
  * Handled malformed API responses more reliably.

### DIFF
--- a/providers/workable.mjs
+++ b/providers/workable.mjs
@@ -1,13 +1,38 @@
 // @ts-check
 /** @typedef {import('./_types.js').Provider} Provider */
 
-// Workable provider — hits the public markdown feed at /<slug>/jobs.md.
-// Workable's documented JSON API requires an auth token; the markdown feed
-// is the only no-auth public surface. Auto-detects from careers_url pattern
-// `https://apply.workable.com/<slug>`. A tracked_companies entry can also
-// set `provider: workable` explicitly to bypass detection.
+// Workable provider — public, no-auth account widget API:
+//   GET https://apply.workable.com/api/v1/widget/accounts/<slug>?details=true
+//   → { name, description, jobs: [{ title, shortcode, shortlink, url, department,
+//        city, state, country, telecommuting, published_on, description, … }] }
+//
+// The widget API returns the account's FULL posting list in one request (verified
+// live against a 259-posting account) and ships `description` + `published_on`
+// for free, so scan.mjs's content_filter and the recency logic in
+// scan-ats-full.mjs both work for Workable companies.
+//
+// The older markdown feed at /<slug>/jobs.md is kept as a fallback only. It
+// cannot be the primary path:
+//   · on a large account the bare feed returns a department *summary* instead of
+//     the job table, so there are no rows to parse — a 259-posting account
+//     silently yielded ZERO jobs;
+//   · it is hard-capped at 30 rows and honours no pagination parameter
+//     (page/offset/limit/startrow all return the same first 30);
+//   · segmenting by `?department=` is incomplete (departments listed in the
+//     summary sum to less than the account total, and a department over 30 roles
+//     is itself truncated).
+//
+// Auto-detects from careers_url pattern `https://apply.workable.com/<slug>`. A
+// tracked_companies entry can also set `provider: workable` to bypass detection.
+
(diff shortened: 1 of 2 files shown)


santifer:
Welcome to career-ops, @omarfsahmed! Verified live: the widget API answers keyless with the full list where the markdown feed truncated, your parser's field names match the real payload exactly, and the dead-board contract holds on both paths (dead slug throws, empty-but-live stays []). Keeping the old feed as fallback was the right conservative call. Merged — your first contribution. 🚀

Issue #2235, opened 2026-07-28
update-system.mjs SYSTEM_PATHS missing sync-pdf-flags.mjs — upgraded installs silently stop syncing the tracker PDF column


prannai:
## Summary

`sync-pdf-flags.mjs` ships in the repo but is **not listed in `update-system.mjs`'s `SYSTEM_PATHS`**, so `node update-system.mjs apply` never copies it to an existing install. `merge-tracker.mjs` shells out to it unconditionally, so on any upgraded install the tracker's PDF column silently stops syncing and two test suites crash.

This is a recurrence of the class of bug behind #1002 / #1005 (`SYSTEM_PATHS` missing shipped files, leaving upgraded installs broken).

## Evidence

`sync-pdf-flags.mjs` is present in the tree (introduced by c47094b, "fix: sync the PDF and scan summary output (#2213)"), but absent from the updater manifest, while its sibling scripts are listed:

```
$ grep -n "'merge-tracker.mjs'\|'sync-pdf-flags.mjs'\|'set-status.mjs'" update-system.mjs
143:  'merge-tracker.mjs',
155:  'set-status.mjs',
```

`merge-tracker.mjs:765` invokes it on every non-dry-run merge:

```js
if (!DRY_RUN) {
  try {
    execFileSync('node', [join(CAREER_OPS, 'sync-pdf-flags.mjs')], { stdio: 'inherit' });
  } catch (e) {
    console.warn(`⚠️  Failed to sync PDF flags: ${e.message}`);
  }
```

`tests/sync-pdf-flags.test.mjs` is also shipped and also missing from `SYSTEM_PATHS`.

## Steps to reproduce

1. Have an install predating c47094b.
2. Run `node update-system.mjs apply` to upgrade to v1.22.0.
3. `ls sync-pdf-flags.mjs` → not present.
4. `node merge-tracker.mjs` → `⚠️  Failed to sync PDF flags: Command failed: node .../sync-pdf-flags.mjs`
5. `node test-all.mjs` → 7 failures on a clean install.

## Impact

**Silent data staleness.** The `catch` in `merge-tracker.mjs` downgrades the failure to a warning, so merges appear to succeed while the tracker's PDF ✅/❌ column stays stale — which is precisely the bug #1429 added this script to fix. Users who upgraded never got the fix and have no obvious signal that anything is wrong.

[...]

PR #2443, opened 2026-08-03
fix(doctor): detect Playwright MCP in JSONC OpenCode configs (#2252)

### LINKED ISSUE #2252: Support JSONC OpenCode configs in doctor Playwright MCP detection
## Summary
Extend `doctor.mjs` so Playwright MCP detection supports OpenCode's `opencode.jsonc` format, including comments and trailing commas.

## Rationale
OpenCode accepts JSONC project configuration. Parsing `opencode.jsonc` with `JSON.parse` treats otherwise valid configurations as malformed, which can incorrectly report that the Playwright MCP server is not configured.

## Affected area
- `doctor.mjs`: `isPlaywrightMcpConfigured()` and its configuration parsing path
- Playwright MCP detection tests

## Required changes
- Use a JSONC-tolerant parser (or a narrowly scoped equivalent) when reading `.jsonc` configuration files.
- Retain normal JSON parsing for `.json` files.
- Ensure commented and trailing-comma OpenCode MCP configurations that reference `@playwright/mcp` are recognized.

## Acceptance criteria
- A valid `opencode.jsonc` containing comments and/or trailing commas is detected as configured when its MCP definition references `@playwright/mcp`.
- Existing `.json` configuration behavior remains unchanged.
- Add automated coverage for the JSONC cases.

## Backlinks
- Source PR: https://github.com/santifer/career-ops/pull/2247
- Source review comment: https://github.com/santifer/career-ops/pull/2247#discussion_r3666545524

Requested by @wasvlad.

Zoubeir23:
Fixes #2252

## Problem

OpenCode accepts JSONC for its project config. `doctor.mjs` only looked at `opencode.json` and parsed it with `JSON.parse`, inside a `catch` that treats any throw as "not configured":

```js
{ cli: 'opencode', files: ['opencode.json'] },
…
const cfg = JSON.parse(readFileSync(file, 'utf8')) ?? {};
```

So a user whose `opencode.jsonc` registers `@playwright/mcp` — or whose `opencode.json` carries a comment — is told the Playwright MCP server is not configured, and SPA job boards keep failing for a reason the doctor just misdiagnosed.

## Fix

- `opencode.jsonc` joins the opencode file list.
- `.jsonc` files parse through a new `jsonc-parse.mjs`; **`.json` keeps strict `JSON.parse`**, per the acceptance criteria.
- Only the two extensions OpenCode configs actually use are tolerated (comments, trailing commas), and the result still goes through `JSON.parse` — a genuinely malformed file throws exactly as before and stays "unconfigured".

**The parser is string-aware, which is the whole difficulty.** A regex stripper corrupts real configs: `"https://example.com"` is not a comment, `"src/*/index.mjs"` is not a block comment, `"a \" // b"` does not end where it looks like it does, and `{"a": "x, }"}` has no trailing comma. The scan tracks string state and escapes, and removes a trailing comma by walking back over *already-emitted* output — so a comma inside a string is never the one it finds, and nesting needs no extra state.

`jsonc-parse.mjs` is added to `SYSTEM_PATHS` in `update-system.mjs`: `doctor.mjs` imports it, so an install that received the new doctor without the module would crash on startup (the #2235 failure mode).

## Test plan

- [x] `tests/jsonc-parse.test.mjs` — 19 assertions, weighted toward what a naive stripper breaks: `//` and `/*` inside strings, escaped quotes, Windows paths, a comma before `}` inside a string value, nested trailing commas, a realistic OpenCode config, malformed input still throwing, and the extension routing (`.jsonc` tolerant, `.JSONC` case-insensitive, `.json` strict)
- [x] `tests/playwright-mcp-detection.test.mjs` — 4 new end-to-end cases: JSONC with comments + trailing comma is detected; a commented `.json` still reads as unconfigured; `opencode.json` is not shadowed by a sibling `.jsonc`; a malformed `.jsonc` stays unconfigured
- [x] `node test-all.mjs --quick` — 2807 passed, 0 failed

## Note on scope

A commented `opencode.json` (no `c`) still reads as unconfigured. That is the issue's explicit acceptance criterion ("Retain normal JSON parsing for `.json` files"), so it is left as-is rather than widened — say the word if OpenCode's own loader tolerates comments there too and you want the tolerance keyed on the CLI instead of the extension.

## Summary by CodeRabbit

* **New Features**
  * Added support for `opencode.jsonc` MCP configuration files.
  * JSONC files now support comments and trailing commas.
  * Standard `.json` files continue to use strict JSON parsing.

* **Bug Fixes**
  * Improved handling of malformed configuration files.
  * Preserved comment-like text inside strings and paths during parsing.

* **Tests**
  * Added coverage for JSONC parsing, file precedence, malformed configurations, and realistic MCP settings.

### DIFF
--- a/doctor.mjs
+++ b/doctor.mjs
@@ -12,6 +12,7 @@ import yaml from 'js-yaml';
 import dotenv from 'dotenv';
 import { discoverPlugins, pluginRoots, pluginStatus } from './plugins/_engine.mjs';
 import { resolveExtractorMode } from './browser-extract.mjs';
+import { parseConfigByExtension } from './jsonc-parse.mjs';
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const argv = process.argv.slice(2);
@@ -108,7 +109,10 @@ async function checkPlaywright() {
 // Per-CLI MCP config registry.
 const MCP_CONFIGS = [
   { cli: 'claude',   files: ['.mcp.json', '.claude/settings.json', '.claude/settings.local.json'] },
-  { cli: 'opencode', files: ['opencode.json'] },
+  // opencode.jsonc is JSONC: OpenCode accepts comments and trailing commas
+  // there, and JSON.parse throwing on them used to read as "no MCP server
+  // configured" (#2252).
+  { cli: 'opencode', files: ['opencode.json', 'opencode.jsonc'] },
 ];
 
 // Server qualifies if its definition references the @playwright/mcp package.
@@ -125,7 +129,7 @@ function isPlaywrightMcpConfigured(root, activeCli) {
     const file = join(root, ...rel.split('/'));
     if (!existsSync(file)) return false;
     try {
-      const cfg = JSON.parse(readFileSync(file, 'utf8')) ?? {};
+      const cfg = parseConfigByExtension(file, readFileSync(file, 'utf8')) ?? {};
       const buckets = [cfg.mcpServers, cfg.mcp].filter((b) => b && typeof b === 'object');
       return buckets.some((servers) => Object.values(servers).some(isPlaywrightServer));
     } catch {
--- a/jsonc-parse.mjs
+++ b/jsonc-parse.mjs
@@ -1,0 +1,116 @@
+/**
+ * jsonc-parse.mjs — JSON.parse for config files written in JSONC.
+ *
+ * OpenCode accepts JSONC for its project config (`opencode.jsonc`): `//` and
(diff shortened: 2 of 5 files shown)


santifer:
No-dep JSONC done the right way, @Zoubeir23: your string-aware scanner still hands everything to JSON.parse, so genuinely malformed configs stay loud — and it survived my adversarial set (comments inside strings, escaped quotes, Windows paths, CRLF, nested trailing commas). Merged. 🚀

PR #2444, opened 2026-08-03
fix(verify-cv-facts): stop the modifier count from deciding whether a claim exists (#2279)

### LINKED ISSUE #2279: verify-cv-facts.mjs: modifier-count asymmetry fails truthful CVs (and can hide changed numbers)
`COUNT_CLAIM_RE` in `verify-cv-facts.mjs` tolerates up to two modifiers between a number and its noun:

```js
String.raw`\b(\d[\d,.]*)\s*\+?\s*(?:[A-Za-z][A-Za-z-]*\s+){0,2}(${METRIC_NOUNS.join('|')})\b`
```

Because the same regex extracts claims from both the generated CV and the source files, a phrase with **three** modifiers is invisible on both sides — fine. But when the CV and the source word the same fact with a *different* number of modifiers, and one side lands at ≤2 while the other lands at 3, the claim is extracted from only one side and the gate fails on a true statement.

## Reproduction

`cv.md` contains:

```
Consolidated 25+ services down to ~5 live Cloud Run deployments
```

`5 … deployments` has three intervening words (`live`, `Cloud`, `Run`), so it is **not** extracted as a source claim.

A generated CV that paraphrases slightly:

```
25+ services consolidated to ~5 Cloud Run deployments
```

has two (`Cloud`, `Run`), so it **is** extracted — and fails:

```
Metric-like claims absent from sources:
  - 5 deployments
```

The number is identical and the source sentence is right there. The only difference is one dropped adjective.

## Why it matters

`modes/pdf.md` Step 18 makes this a hard gate before PDF rendering, and the failure message directs the user to "remove invented metrics or add verified evidence" — pushing toward editing `cv.md` or adding a `config/cv-facts.json` exception for a fact that is already correctly stated. The safe reflex (loosen the claim, add an exception) quietly weakens the guard that exists to catch real fabrications.

It fails in the more dangerous direction too: source at ≤2 modifiers, CV at 3, means a *changed number* passes unnoticed, because the CV side never yields a claim to compare.

## Suggested shape

Compare on the number-plus-noun pair independent of how many modifiers sit between them — e.g. raise the window to `{0,4}` and reconcile by `(number, noun)`, or extract candidate numbers and search for the nearest metric noun within N tokens on both sides using the same window. The key property to preserve is symmetry: whatever window the source is parsed with, the CV must be parsed with too, so a modifier difference can never decide whether a claim exists.

Worth a self-test case in `--self-test` pinning the asymmetry: same number, same noun, differing modifier counts, expect pass.

Zoubeir23:
Fixes #2279

## Problem

`COUNT_CLAIM_RE` allowed at most two words between a number and the noun it counts. The same regex parses the generated CV and the sources, so the window is symmetric — but a window still decides **whether a claim exists at all**, and a CV rarely words a fact exactly like its source.

Reproduced on `main` with the reporter's example:

```
source: "Consolidated 25+ services down to ~5 live Cloud Run deployments"
        → claims: ["25 services"]                    ← "5 deployments" missing (3 modifiers)

CV:     "25+ services consolidated to ~5 Cloud Run deployments"
        → claims: ["25 services", "5 deployments"]   ← extracted (2 modifiers)
        → invented: ["5 deployments"]                ← a TRUE statement fails the gate
```

And the direction the maintainer flagged as the serious one:

```
CV:     "25+ services consolidated to ~9 live Cloud Run deployments"   (5 → 9)
        → claims: ["25 services"]
        → invented: []                               ← a CHANGED number passes silently
```

`modes/pdf.md` Step 18 makes this a hard gate, and its failure message pushes the user toward editing `cv.md` or adding a `config/cv-facts.json` exception for a fact that was already correct — weakening the guard that exists to catch real fabrications.

## Fix

The window becomes a named `MODIFIER_WINDOW = 4`, used by the one shared regex.

Widening is safe in the direction that matters: it only ever extracts **more** claims, on both sides, so it cannot hide an invented number — it can only surface more of them. And a figure remains a hard barrier for the chain (modifiers are alphabetic only), so a wider window still cannot jump over an intervening number to bind a noun it does not count. That last property is pinned by a test: `"Shipped 3 features across 12 integrations"` must not become evidence for a CV claiming `"3 integrations"`.

Four covers the phrasings that appear in real CVs (`live Cloud Run deployments`, `active monthly paying customers`).

## Test plan

- [x] `node verify-cv-facts.mjs --self-test` — 16 passed, 0 failed (10 existing + 6 new), and it is already wired into `test-all.mjs:218` with `expectExit: 0`
- [x] The two asymmetry cases were verified to **fail at the old `{0,2}`** and pass at `{0,4}` — reverting the constant produces exactly the two documented failures, so the guard has value rather than just describing current behaviour
- [x] New cases: same number with fewer modifiers than the source; same number with more; a changed number caught behind either phrasing; a figure still blocking the chain; no cross-number binding inventing evidence

## Residual limit, stated plainly

Any finite window has a cliff — this moves it from 3 modifiers to 5, it does not remove it. Removing it entirely means changing the model: extract candidate numbers, then look for the nearest metric noun in the same clause on both sides, or keep strict extraction on the CV side while matching evidence generously on the source side. Both are bigger changes than this issue asks for; happy to follow up with either if you want the cliff gone rather than pushed out.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved extraction of metric claims with multiple descriptive modifiers between numbers and counted nouns.
  * Prevented incorrect associations across separate numeric claims.
  * Improved detection of changed metric values in longer phrases.

### DIFF
--- a/verify-cv-facts.mjs
+++ b/verify-cv-facts.mjs
@@ -37,8 +37,27 @@ const METRIC_NOUNS = [
   'servers', 'guides', 'articles', 'datasets', 'examples', 'deployments',
   'services', 'downloads', 'stars', 'lines', 'projects', 'integrations', 'tests',
 ];
+// How many words may sit between a number and the noun it counts. The same
+// regex parses the generated CV and the sources, so the window is symmetric by
+// construction — but a window still decides WHETHER a claim exists, and the CV
+// and its source rarely word a fact identically. At {0,2}, "~5 live Cloud Run
+// deployments" (three modifiers) yielded no claim while the paraphrase
+// "~5 Cloud Run deployments" (two) did, which broke the gate in both
+// directions (#2279):
+//
+//   - a truthful CV failed, because the claim existed on the CV side only;
+//   - a CHANGED number passed, because a 3-modifier phrasing on the CV side
+//     produced no claim to compare — and catching invented numbers is the
+//     entire point of this script.
+//
+// Four covers the phrasings seen in real CVs ("live Cloud Run deployments",
+// "active monthly paying customers"). Widening cannot hide an invented number:
+// it only ever extracts MORE claims, on both sides. A number is a hard barrier
+// for the chain — modifiers are alphabetic only — so a wider window still
+// cannot jump across an intervening figure to bind an unrelated noun.
+const MODIFIER_WINDOW = 4;
 const COUNT_CLAIM_RE = new RegExp(
-  String.raw`\b(\d[\d,.]*)\s*\+?\s*(?:[A-Za-z][A-Za-z-]*\s+){0,2}(${METRIC_NOUNS.join('|')})\b`,
+  String.raw`\b(\d[\d,.]*)\s*\+?\s*(?:[A-Za-z][A-Za-z-]*\s+){0,${MODIFIER_WINDOW}}(${METRIC_NOUNS.join('|')})\b`,
   'gi'
 );
 const NOUN_SYNONYMS = new Map([
--- a/verify-cv-facts.mjs
+++ b/verify-cv-facts.mjs
@@ -430,6 +430,46 @@ function runSelfTest() {
   // A four-digit left part is a year, not a group: nothing is joined.
   equal('a year is not glued to the next number', auditClaims('Joined in 2026 100 users', foldSource).invented, ['100 users']);
 
+  // #2279 — the modifier count must never decide whether a claim exists. The
+  // source words the fact with three modifiers, the CV with two: at the old
+  // {0,2} window the claim was extracted from the CV side only, and a true
+  // statement failed the gate.
+  const modifierSource = 'Consolidated 25+ services down to ~5 live Cloud Run deployments.';
+  equal(
+    'same number, fewer modifiers in the CV',
+    auditClaims('25+ services consolidated to ~5 Cloud Run deployments', modifierSource).invented,
+    []
+  );
+  equal(
+    'same number, more modifiers in the CV',
+    auditClaims('Consolidated to ~5 live production Cloud Run deployments', modifierSource).invented,
+    []
+  );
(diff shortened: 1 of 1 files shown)


santifer:
Clean kill of #2279's mechanism, @Zoubeir23 — verified with the issue's literal repro in both directions, and your revert-check tests discriminate exactly (window=2 fails precisely the two documented cases, no more, no fewer). One regression though, empirically demonstrated — and it comes from the issue's own suggested variant 1, not from your execution:

Greedy `{0,4}` binds the number to the FARTHEST metric noun in the window, and worded numbers (*three*, *two*) aren't barriers — only digits are. Two pairs that pass on main and fail with this PR: source "Cut onboarding to 45 minutes across three sessions." + truthful CV "Cut onboarding to 45 minutes." → `invented=['45 minutes']`, because the source side now extracts "45 sessions". Same shape with "Reduced deploy time by 30 minutes for two teams." That's standard prose style (numbers under ten in words) failing the hard PDF gate — exactly the damage #2279 warns about.

The new comment's monotonicity claim ("widening only ever extracts MORE claims") is what breaks: widening can *replace* a correct claim with a wrong one on a single side.

**One-character fix, verified:** make the quantifier lazy — `{0,4}?` — which is nearest-noun preference, i.e. the variant 2 you already offered as a follow-up. On your own 16 self-tests: all green, both regression pairs dead, and it fixes a pre-existing main false positive as a bonus ("Logged 40 hours across projects."). Add one test pinning the "45 minutes across three sessions" pair as `invented=[]` and this merges immediately.

santifer:
This is shape 1 from #2279 done right, @Zoubeir23: the named MODIFIER_WINDOW constant with self-tests pinning the asymmetry means the next person can't reintroduce the bug blind. Merged. 🚀

Issue #2268, opened 2026-07-29
tests: followup-cadence and followup-seed suites read the user's real config/profile.yml, so any cadence customization turns test-all.mjs red


ringo380:
## Symptom

On v1.23.0, `node test-all.mjs` reports 5 failures on a machine where the user has customized `followup_cadence` in `config/profile.yml`. Four of them come from the follow-up suites reading the **real user profile** instead of a fixture:

```
❌ followup-seed-tests.mjs crashed
❌ followup-cadence.test.mjs crashed
❌ computeUrgency applied past applied_first → overdue: expected overdue, got waiting
❌ computeNextFollowupDate first applied follow-up = appDate + applied_first: expected 2026-05-08, got 2026-05-15
```

Standalone runs:

```
$ node followup-cadence.test.mjs
  FAIL: applied, no follow-ups uses applied_first
    expected: "2026-07-07"
    actual:   "2026-07-14"
9 passed, 1 failed

$ node followup-seed-tests.mjs
  FAIL 1. resolved next date — expected 2026-06-27, got 2026-07-04
34 passed, 1 failed
```

## Cause

`followup-cadence.mjs:25`:

```js
const PROFILE_FILE = process.env.CAREER_OPS_PROFILE || join(CAREER_OPS, 'config/profile.yml');
```

Neither `followup-cadence.test.mjs` nor `followup-seed-tests.mjs` sets `CAREER_OPS_PROFILE`, so both fall through to the user's own `config/profile.yml`. The assertions hardcode the 7-day `applied_first` default; any user override changes the computed dates and the tests fail.

Reproduction — a profile containing:

```yaml
followup_cadence:
  applied_first_days: 14
  applied_subsequent_days: 10
  applied_max_followups: 2
```

is enough. Every date the tests assert shifts by the delta (7 → 14 days above).

Proof that the profile is the only cause — move it aside and both suites go green, restore and they fail again:

```
$ mv config/profile.yml config/profile.yml.aside
$ node followup-cadence.test.mjs   # 10 passed, 0 failed
$ node followup-seed-tests.mjs     # 35 passed, 0 failed
$ mv config/profile.yml.aside config/profile.yml   # back to 1 failed each
```

## Why it matters

[...]

PR #2446, opened 2026-08-03
test(followup): pin the cadence profile so a customized install isn't red (#2268)

Zoubeir23:
Part of #2268

## Problem

`followup-cadence.mjs` resolves its cadence **at module load**:

```js
const PROFILE_FILE = process.env.CAREER_OPS_PROFILE || join(CAREER_OPS, 'config/profile.yml');
…
const CADENCE = resolveCadenceConfig();
```

Neither follow-up suite set `CAREER_OPS_PROFILE`, so both fell through to the **user's own** `config/profile.yml` while asserting the 7-day `applied_first` default. Any user who customized `followup_cadence` saw a red `test-all.mjs` on a completely healthy install — and maintainers never see it, because dev machines don't customize.

`followup-seed-tests.mjs` had the same hole twice over: its in-process import, and the `followup-seed.mjs` child it spawns, whose env inherited `process.env` and therefore resolved the real user profile too.

## Fix

Both suites pin `CAREER_OPS_PROFILE` to a fixture before the module is evaluated, and `run()` passes the same pin explicitly to the spawned child.

**The import has to become dynamic.** ESM hoists static imports above every statement in the file, so `process.env.CAREER_OPS_PROFILE = …` followed by a static import would run the module *first* and the pin would silently do nothing. That is called out in a comment at both sites, because it is exactly the kind of thing a later "tidy the imports" commit undoes.

Two fixtures, not one:

- `tests/fixtures/profile-default-cadence.yml` — no `followup_cadence` block, so every key falls back to `DEFAULT_CADENCE`, which is what the assertions already encode.
- `tests/fixtures/profile-custom-cadence.yml` — a customized cadence, used by three new assertions that guard the pin from degrading into *ignoring* the profile. Making the suites deterministic by hard-coding the cadence would pass every existing test while breaking the feature; those assertions fail if that happens.

## Test plan

Reproduced with a profile carrying `applied_first_days: 14`, `responded_initial_days: 5` (the shape of the reporter's customization):

| Suite | Before | After |
|---|---|---|
| `followup-cadence.test.mjs` | 15 passed, **3 failed** | 21 passed, 0 failed |
| `followup-seed-tests.mjs` | 41 passed, **1 failed** | 42 passed, 0 failed |

- [x] Both suites pass standalone with no `CAREER_OPS_PROFILE` set, and with a hostile one set
- [x] New assertions: a fixture with no `followup_cadence` yields `{}`; a customized profile still resolves through `profilePath`; the pinned fixture resolves to exactly `DEFAULT_CADENCE`

## Summary by CodeRabbit

- **Tests**
  - Added coverage for default and custom follow-up cadence profiles.
  - Ensured cadence and date-based tests produce consistent results regardless of local profile settings.
  - Verified child processes use the intended test profile.

- **Fixtures**
  - Added default and customized profile fixtures to validate cadence fallback and configuration behavior.

### DIFF
--- a/followup-cadence.test.mjs
+++ b/followup-cadence.test.mjs
@@ -7,7 +7,25 @@
  * Run: node followup-cadence.test.mjs
  */
 
-import {
+import { dirname, join } from 'path';
+import { fileURLToPath } from 'url';
+
+const ROOT = dirname(fileURLToPath(import.meta.url));
+const DEFAULT_CADENCE_PROFILE = join(ROOT, 'tests', 'fixtures', 'profile-default-cadence.yml');
+const CUSTOM_CADENCE_PROFILE = join(ROOT, 'tests', 'fixtures', 'profile-custom-cadence.yml');
+
+// Pin the cadence source BEFORE followup-cadence.mjs is evaluated. Its
+// module-level `CADENCE = resolveCadenceConfig()` reads CAREER_OPS_PROFILE at
+// import time and otherwise falls back to the USER's config/profile.yml — so
+// on a machine where the user customized followup_cadence, assertions written
+// against DEFAULT_CADENCE failed on a perfectly healthy install (#2268).
+//
+// The import below must stay DYNAMIC: ESM hoists static imports above every
+// statement in this file, so a static import would run the module before this
+// assignment and the pin would do nothing.
+process.env.CAREER_OPS_PROFILE = DEFAULT_CADENCE_PROFILE;
+
+const {
   computeNextFollowupDate,
   addDays,
   parseDate,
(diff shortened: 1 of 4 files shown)


santifer:
Verified working, @Zoubeir23: CAREER_OPS_PROFILE pinned to committed fixtures before the dynamic import, the child spawn pinned explicitly, and the three guard assertions keep the pin from ever degrading into ignoring the profile. One honest scope note: test-all's §12 still imports followup-cadence in-process and asserts default cadence, so a customized install still goes red there — I've re-scoped this to 'Part of #2268' and the issue stays open anchored on that residual. Your half is solid. Merged. 🚀

Issue #1710, opened 2026-07-09
feat(modes): split _shared.md so evaluation stops loading ~2K tokens of CV-writing guidance


abankar1:
## Problem

Every offer evaluation loads **`modes/_shared.md` in full**, but roughly a third of that file is CV/cover **writing** guidance that the A–G evaluation never uses.

`modes/_shared.md` (315 lines, 21,978 B) mixes two concerns:

- **Eval-relevant:** Sources of Truth, Scoring System, Posting Legitimacy (Block G), Company Type & Compensation Reliability, Archetype Detection, Global Rules.
- **Writing-only:** *Voice DNA*, *Writing Style Calibration*, *Writing Style*, *Professional Writing & ATS Compatibility* (lines ~196–312) — **8,057 B ≈ ~2K tokens** that only matter when generating a CV / cover letter, not when scoring an offer.

All four engines read the whole file (`ollama-eval.mjs:205`, `openai-eval.mjs:215`, `gemini-eval.mjs:229`, `openrouter-runner.mjs:323`), and so does the `claude -p` CLI eval path. So this ~2K-token writing block is dead weight on **every eval, on every path** — including the CLI batch path that prompt-caching in the API engines (#1709) can't help.

## Proposed solution

Split `_shared.md` by concern so each mode pulls only what it needs:

- `_shared.md` (or `_shared-core.md`) → the eval-relevant sections, loaded by `oferta.md` / `auto-pipeline.md` / the eval engines.
- a `_writing.md` addendum → Voice DNA + Writing Style sections, loaded only by the CV/cover/apply modes (`pdf.md`, `cover.md`, `apply.md`, `add.md`).

Net effect: ~2K fewer tokens on **every** evaluation, universally and permanently — the one cost lever that also helps the primary CLI population, on top of any caching.

## Scope / risk

This touches a shared system-layer file that many modes and all four eval engines include, so it needs a design call before I touch it — raising per CONTRIBUTING rather than opening a PR. If the split sounds right I'm happy to:
- do it carefully (update every mode's include, update the eval engines, keep byte-for-byte content — just relocated),
- add a `test-all.mjs` guard that the eval engines don't load the writing sections and the writing modes do,
- verify no eval-relevant guidance is lost.

Open question for the maintainer: prefer a hard file split, or keep one file with a machine-readable section boundary the engines slice on? Happy to go either way.

Related: complements #1709 (cache the prefix) — this trims the prefix itself.

santifer:
[...]

PR #2447, opened 2026-08-03
docs(modes): cover and email load the shared writing module (#2006)

### LINKED ISSUE #2006: Should cover.md / email.md load the new modes/_writing.md module? (follow-up to #1710)
Follow-up to #1710, opened per @santifer's green-light note:

> **cover.md / email.md: not in this PR.** The split must stay a pure relocation, byte-reviewable. Whether those two should load `_writing.md` is a real question with its own behavior implications: open it as a separate follow-up issue and we'll decide it on its own merits.

Filing it now so #1710 (PR #2005) stays a pure, byte-reviewable relocation.

## Current state (as of `main`)

I checked what `modes/cover.md` and `modes/email.md` actually load for voice/writing guidance today. **Neither references `_shared.md`'s writing sections at all** — so extracting them into `_writing.md` does *not* silently drop anything these two modes rely on. Their current voice sources are:

- `modes/cover.md` → `modes/_profile.md` (its rules "govern the letter's voice and structure and override the generic defaults"). No `_shared.md` / Voice DNA / Writing Style reference. It carries its own inline writing rules (active voice, tone consistency, self-check).
- `modes/email.md` → `modes/_profile.md` + `voice-dna.md` ("only as a writing guardrail… never introduce factual claims"). No `_shared.md` reference either.

So this is a genuine *additive* question, not a regression to guard against.

## The question

Should `cover.md` and `email.md` gain an explicit read of `modes/_writing.md` (Voice DNA + Writing Style), so cover letters and application emails draw on the same shared writing guidance that the evaluation and apply modes do?

**Arguments for:**
- Consistency: cover letters and application emails are user-facing prose, the same category `_writing.md` governs. Right now they get a *different* (thinner) writing standard than the report/apply output.
- `_writing.md` now exists as a cheap, standalone include (the whole point of #1710) — wiring one `Read` line into each is low-cost.

**Arguments against / open considerations:**
- Both already have their own writing discipline (`cover.md`'s inline rules, `email.md`'s `voice-dna.md` guardrail). Adding `_writing.md` may duplicate or conflict with those.
- `_writing.md` is written for the evaluation/PDF/apply voice; some of it may not transfer cleanly to a short application email. May need scoping (e.g. cover.md yes, email.md's shorter drafts no).
- Token cost: these are already lean modes; loading `_writing.md` adds ~2.1K tokens per run.

## Proposed scope if green-lit

Pure wiring, no content change to `_writing.md`:
1. Add a `Read modes/_writing.md` directive to `cover.md` (and/or `email.md`), positioned so `_profile.md` / `voice-dna.md` still override on conflict (personalization wins).
2. Extend the stale-reference guard added in #1710's test to assert these modes point at `_writing.md` if they claim to use it.

Happy to take whichever subset you decide — cover only, both, or neither. Deferring to your call on the merits.

Zoubeir23:
Fixes #2006

Implements the answer given in the issue:

> the answer is yes in principle: cover.md and email.md predate `modes/_writing.md` and duplicate guidance it now owns; loading it and deleting the duplicated prose is the thin-wrapper direction this repo already follows for agent files. One caution: those modes' output contracts (fact gates, structure) must not change as a side effect — a diff that only swaps duplicated prose for the module reference would merge easily.

## What changed

Both modes now read `modes/_writing.md`, and **only strictly duplicated prose** was swapped for the reference.

- **`cover.md`** — read directive in Step 1, next to `cv.md` / `article-digest.md` / `_profile.md`. Its "Language rules" list keeps every mode-specific contract and now names `_writing.md` → Professional Writing as the base.
- **`email.md`** — `modes/_writing.md` added to the Step 1 `Read:` list; the "Style Rules" section points at the shared cliché list for the entries it already carried.

## No rule is lost

The bans removed from `cover.md`'s inline list are exactly the ones `_writing.md` already carries: *leverage(d), synergy/synergies, seamless, robust, cutting-edge, spearheaded, passionate*. Everything stricter than the shared list stays inline: *holistic, championed, orchestrated, excited, stakeholder alignment, data-driven, actionable insights, move the needle, north star, unique opportunity, perfect fit, strong track record*. Same for `email.md`: "passionate about" now comes from the shared list, "perfect fit" and "unique opportunity" stay local because the shared list does not have them.

`cover.md`'s "No em dashes" is deliberately kept inline rather than delegated: `_writing.md` frames em-dashes as an ATS normalization concern, while in a letter it is a prose rule that applies before any parser is involved.

## Output contracts untouched

Word count (350-420), bullet format, self-check, tone consistency, variants, attachment checklist, subject/body structure, the draft-only rule and the fact-validator gate are all unchanged — and now pinned by a test, so a future edit cannot quietly take them along with the delegated prose.

Precedence is stated explicitly in both modes: `_profile.md` > `voice-dna.md` > `_writing.md`. The module adds a standard where these modes had a thinner local one; it never overrides the user's own rules.

## Test plan

- [x] Extended the #1710 stale-reference guard as the issue proposed: both modes must actually **read** `_writing.md`, not merely mention it — a mode that delegates its wording rules and then loses the directive drops them silently, with the delegating prose still pointing at a file nobody opens
- [x] The guard accepts both directive shapes used in `modes/` (inline, and a bullet inside a `Read:` list). Matching only the inline form flagged `email.md` while it reads the file correctly — caught by the guard itself before this PR was opened
- [x] Both shapes **require the verb**. An earlier version made `Read` optional, so a bare mention — `Do not read \`modes/_writing.md\``, a path in a table cell — satisfied it, and the assertion could have stayed green after the real read was deleted (CodeRabbit review). Verified against both real modes plus six bypass shapes: bare mention, negation, bullet outside a `Read:` list, bullet after the list has ended, and a table cell
- [x] Verified it fails closed: removing either directive turns the assertion red
- [x] Second assertion pins the mode-specific contracts `_writing.md` says nothing about
- [x] `node test-all.mjs --quick` — 2785 passed, 0 failed

## Scope note

The issue also raises the ~2.1K-token cost per run and asks whether `email.md` should be in or out. This PR takes **both**, on the consistency argument in the issue. If you would rather scope it to `cover.md` only, dropping the `email.md` hunk is a two-line revert and the guard's consumer list is one line.

## Summary by CodeRabbit

* **Improvements**
  * Cover letters and emails now follow consistent shared writing and professional communication guidance.
  * Candidate-specific voice and profile preferences take precedence where available.
  * Cover-letter guidance includes stricter formatting and language rules, including expanded cliché and buzzword avoidance.
  * Email guidance more clearly separates writing style from required output safeguards.

* **Quality**
  * Added validation to ensure required writing guidance and mode-specific output requirements remain in place.

### DIFF
--- a/modes/cover.md
+++ b/modes/cover.md
@@ -38,7 +38,9 @@ Read `cv.md` for:
 
 Read `article-digest.md` if it exists — supplementary proof points and metrics take precedence over cv.md where they overlap.
 
-Read `modes/_profile.md` if it exists — the candidate's personalization file. It captures their target roles, adaptive framing and archetypes, exit narrative, cross-cutting advantage, proof points, comp targets, negotiation scripts, location policy, and any voice or writing-style rules they have added. Its rules **govern the letter's voice and structure and override the generic defaults in this mode**, so the candidate's personalization is never lost.
+Read `modes/_writing.md` — the shared writing guidance (Voice DNA guardrail, Writing Style calibration, Professional Writing & ATS rules). A cover letter is candidate-facing prose, the same category that module governs, so it gets the same standard as the report and apply outputs instead of a thinner local one (#2006).
+
+Read `modes/_profile.md` if it exists — the candidate's personalization file. It captures their target roles, adaptive framing and archetypes, exit narrative, cross-cutting advantage, proof points, comp targets, negotiation scripts, location policy, and any voice or writing-style rules they have added. Its rules **govern the letter's voice and structure and override the generic defaults in this mode and in `_writing.md`**, so the candidate's personalization is never lost.
 
 ---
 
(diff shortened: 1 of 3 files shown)


santifer:
Clean follow-through on #2006, @Zoubeir23: swapping only the strictly-duplicated prose and leaving both output contracts untouched is exactly what the green-light asked for. The _writing.md section names all resolve. Merged. 🚀

PR #2448, opened 2026-08-03
feat(update): keep locally edited system files unless --force (#2337)

### LINKED ISSUE #2337: update-system.mjs apply: no visibility when a system-layer file has a local fix that would be silently overwritten

Zoubeir23:
Fixes #2337

Implements the design given in the issue:

> before checkout, diff each `SYSTEM_PATHS` file against `FETCH_HEAD`, list the ones with local modifications, and require an explicit confirmation (or `--force`) to overwrite — with the skipped-by-default option writing a `.bak` alongside. What I'd avoid: turning apply into a merge.

**Not a merge.** The system layer stays system-owned. This only tells people what they are about to lose, and lets them keep it.

## Detection

`locallyModifiedSystemFiles()` is exported and ctx-injectable, like `removeAdditionsNotInHead`. A file is at risk only when **both** hold:

1. it differs from `git merge-base HEAD FETCH_HEAD` — the last commit this install shares with upstream. Anything differing from that baseline was changed *here*, committed or not (`git diff <ref> -- <path>` compares against the working tree);
2. it differs from `FETCH_HEAD` — so the checkout would actually overwrite it.

Plus one case `git diff` cannot see at all: an **untracked** file the user created at a path the upstream ref ships. It appears in neither diff set, and the checkout would overwrite it with no warning and no `.bak` — the exact loss mode this targets (caught in review). Untracked files absent upstream are left out: the checkout cannot touch them.

Condition 2 is what keeps the warning honest: the reporter isolated a case where one of their two fixes survived an update because upstream had independently adopted it. Byte-identical content costs nothing to check out, so warning about it would be noise.

Without a merge-base (shallow clone, unrelated histories) it falls back to `HEAD`: still catches uncommitted edits, just misses already-committed ones. A ref that cannot be read returns an empty list rather than throwing — a degraded warning must never abort an update.

## Behaviour

```
3 system file(s) differ from upstream because THIS install changed them:
  generate-cover-letter.mjs  (local copy saved: generate-cover-letter.mjs.bak)
Re-run with `node update-system.mjs apply --force` to take the upstream version instead.
```

- **Default:** those files are kept. `--force` overwrites them.
- **`.bak` is written in both branches**, so the local fix is recoverable even from the forced path. `*.bak` is already in `.gitignore`, so the backups never reach a commit.
- `--force` propagates across the self-reexec via `CAREER_OPS_UPDATE_FORCE`, since the re-exec re-invokes the target updater with a fixed argv.

## Two implementation details worth reviewing

**Exclusion is by pathspec, not checkout-then-restore.** Restoring afterwards would leave the *index* holding the upstream blob, so the scoped commit at step 7 would record exactly the content the user asked to keep out.

**An entry whose files are all preserved is skipped outright.** `git checkout <ref> -- <path> :(exclude)<path>` errors with *"did not match any file(s)"* when the exclusions cancel the whole pathspec — and at that call site the error is indistinguishable from a real failure, so it would have aborted the entire update over a file the user chose to keep. My own test caught this before the PR existed; it is pinned as case #9.

Preserved paths are also excluded from the commit pathspec: an "auto-update system files" commit must not sweep the user's uncommitted edit in under its own message.

## Test plan

- [x] New `tests/updater-local-system-edits.test.mjs` — 11 behavioural cases against a throwaway repo (same harness as `updater-rollback-behavior.test.mjs`): committed local fix reported; uncommitted edit reported; upstream-only change silent; upstream-adopted fix silent; user-layer file never listed; multiple edits sorted; bad ref degrades instead of throwing; an untracked file upstream ships is reported, one it does not is not; the exclude pathspec preserves worktree **and** index; a fully-excluded pathspec errors (why the skip exists)
- [x] `updater-migration-tests.mjs` — 4 new source-pattern assertions, and the #1998 stderr guard widened: it matched the checkout call literally, so the added exclusions broke it while the property it protects (`gitQuiet`, not `git`) was untouched
- [x] `node test-all.mjs --quick` — 2794 passed, 0 failed

## One interpretation to confirm

"the skipped-by-default option writing a `.bak` alongside" — I read the `.bak` as a copy of the **local** content, and write it in both branches so nothing is lost either way. If you meant it only for the skipped path, or meant the `.bak` to hold the incoming upstream version instead, say so and it is a two-line change.

## Summary by CodeRabbit

* **New Features**
  * Added `update --force` support to overwrite locally modified, deleted, or untracked system files.
  * Without `--force`, local changes are backed up and preserved during updates.
  * Updates now detect local edits made before or after upstream changes.
  * Command-line usage documentation includes the new option.

* **Bug Fixes**
  * Improved preservation of local worktree and index content.
  * Prevented preserved files from being included in update commits.
  * Added safer handling for excluded files and manifest-scoped updates.

### DIFF
--- a/tests/updater-local-system-edits.test.mjs
+++ b/tests/updater-local-system-edits.test.mjs
@@ -24,6 +24,11 @@ function makeRepo() {
   g('init', '-q', '-b', 'main', '.');
   g('config', 'user.email', 'test@example.com');
   g('config', 'user.name', 'Test');
+  // A developer or CI image with commit.gpgsign or a global core.hooksPath
+  // would make every commit below fail, and all cases would go red for a
+  // reason that has nothing to do with the detector (CodeRabbit review).
+  g('config', 'commit.gpgsign', 'false');
+  g('config', 'core.hooksPath', join(dir, 'no-such-hooks'));
   mkdirSync(join(dir, 'modes'), { recursive: true });
   writeFileSync(join(dir, 'modes', 'pdf.md'), 'shipped pdf\n');
   writeFileSync(join(dir, 'modes', 'cover.md'), 'shipped cover\n');
@@ -153,6 +158,43 @@ const PATHS = ['modes/', 'generate-cover-letter.mjs'];
     pass('an unreadable upstream ref returns a list instead of throwing');
   } else {
     fail('#7 a bad ref must not throw — it would abort the whole update');
+  }
+}
+
+// ── 7b. An untracked local file the upstream ref ships is at risk too ──
+//    `git diff` never lists untracked files, so this one escaped the two diff
+//    sets entirely and would have been overwritten with no warning and no .bak.
+{
+  const repo = makeRepo();
+  // Ships upstream, absent from the install's baseline.
+  repo.g('checkout', '-q', 'upstream');
+  writeFileSync(join(repo.dir, 'modes', 'new-mode.md'), 'upstream new mode\n');
+  repo.g('add', '-A');
+  repo.g('commit', '-qm', 'upstream: new mode');
+  repo.g('checkout', '-q', 'main');
+  // The user wrote their own file at that exact path before updating.
+  writeFileSync(join(repo.dir, 'modes', 'new-mode.md'), 'my own notes\n');
+
+  const atRisk = locallyModifiedSystemFiles(PATHS, 'upstream', repo.ctx);
+  if (atRisk.includes('modes/new-mode.md')) {
+    pass('an untracked local file the upstream ref ships is reported');
+  } else {
+    fail(`#7b expected modes/new-mode.md, got ${JSON.stringify(atRisk)}`);
+  }
+}
+
+// ── 7c. An untracked file absent upstream is NOT reported ──
+//    The checkout cannot touch it, so listing it would be pure noise.
+{
+  const repo = makeRepo();
+  upstreamChange(repo, 'modes/pdf.md', 'shipped pdf v2\n');
+  writeFileSync(join(repo.dir, 'modes', 'my-scratch.md'), 'purely local\n');
+
+  const atRisk = locallyModifiedSystemFiles(PATHS, 'upstream', repo.ctx);
+  if (!atRisk.includes('modes/my-scratch.md')) {
+    pass('a purely local untracked file is left out of the warning');
+  } else {
+    fail(`#7c my-scratch.md should not be listed: ${JSON.stringify(atRisk)}`);
   }
 }
 
--- a/tests/updater-local-system-edits.test.mjs
+++ b/tests/updater-local-system-edits.test.mjs
@@ -29,6 +29,12 @@ function makeRepo() {
   // reason that has nothing to do with the detector (CodeRabbit review).
   g('config', 'commit.gpgsign', 'false');
   g('config', 'core.hooksPath', join(dir, 'no-such-hooks'));
+  // Same reasoning for line endings: the cases below write LF and later compare
+  // exact strings against files git CHECKED OUT, so a global core.autocrlf=true
+  // (the default on GitHub's Windows images) would hand back CRLF and fail them
+  // for a reason unrelated to the detector (CodeRabbit review).
+  g('config', 'core.autocrlf', 'false');
+  g('config', 'core.eol', 'lf');
   mkdirSync(join(dir, 'modes'), { recursive: true });
   writeFileSync(join(dir, 'modes', 'pdf.md'), 'shipped pdf\n');
   writeFileSync(join(dir, 'modes', 'cover.md'), 'shipped cover\n');
--- a/tests/updater-local-system-edits.test.mjs
+++ b/tests/updater-local-system-edits.test.mjs
@@ -9,7 +9,7 @@
  * before it is overwritten, and nothing else is.
  */
 
-import { mkdtempSync, mkdirSync, writeFileSync, readFileSync } from 'fs';
+import { mkdtempSync, mkdirSync, writeFileSync, readFileSync, rmSync, existsSync } from 'fs';
 import { tmpdir } from 'os';
 import { join } from 'path';
 import { pass, fail } from './helpers.mjs';
@@ -36,7 +36,9 @@ function makeRepo() {
   g('add', '-A');
   g('commit', '-qm', 'base');
   g('branch', 'upstream');
-  return { dir, g, ctx: { git: g } };
+  // `root` lets the detector check whether a path still exists on disk; without
+  // it every case below would resolve against the real career-ops checkout.
+  return { dir, g, ctx: { git: g, root: dir } };
 }
 
 /** Commit a change on the upstream branch and return to main. */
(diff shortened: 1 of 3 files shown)


santifer:
Impressive build, @Zoubeir23 — and a real find under it. The three-way detection is right in cycle 1 (verified it catches exactly a genuine local edit), the untracked-file-that-upstream-ships case is a loss mode plain `git diff` can't see, and the pathspec exclusion keeps index and worktree coherent. Crediting @clede here too: the diff+confirm+bak shape follows his #2337 diagnosis.

One blocker, reproduced empirically — worth the read because CI structurally can't see it:

**The baseline freezes after the first apply.** `locallyModifiedSystemFiles()` uses `merge-base(HEAD, FETCH_HEAD)`, but `apply()` commits upstream *content* as local commits — upstream's commits never enter this history — so the merge-base pins at the clone point forever. From the SECOND update onward, a pristine install (zero local edits) flags every file changed across overlapping release windows as "edited by this install" and preserves it by default. Reproduced: clean install → update 1 fine (`atRisk=[]`) → update 2 reports `atRisk=["VERSION","scan.mjs"]`, both false. Consequence chain: VERSION pins (permanent update-available loop), update-system.mjs self-pins stale, and security fixes to hot files silently stop arriving, with a message that blames the user.

Why your 11 test scenarios stay green: they all run a single cycle from a fresh shared base, where merge-base IS the right baseline. No test simulates the second update after a previous apply-commit.

**Bounded fix that saves the whole design (~10 lines + 1 test):** persist the synced SHA on every apply — e.g. `git update-ref refs/career-ops/last-sync FETCH_HEAD` in step 7, or a state file — and use it as the diff baseline, falling back to merge-base only when it doesn't exist yet (first run / old installs, where it's still correct). Plus one second-cycle test: install → apply → new release → detection must return `[]`.

Minor, non-blocking: a locally-deleted system file lands in `atRisk` and `copyFileSync` throws ENOENT on the .bak; an exists-check would quiet it.

Everything else is merge-ready. With the baseline fix in, this lands.

santifer:
Thanks for pushing on this, @Zoubeir23 — the updater internals here show real understanding (the self-reexec via env var, the `:(exclude)` pathspec, index-vs-worktree handling). One thing from the earlier direction still isn't in, and it's the load-bearing part: the ask was to record a **last-sync marker** (a `refs/career-ops/last-sync` ref or a state file) so the check compares against what was actually last applied, not against `merge-base HEAD upstream`. On the current head, `update-system.mjs:735` is still `baseline = runGit('merge-base', 'HEAD', upstreamRef)` and there's no `last-sync` anywhere in the diff. Without the marker, the detection still can't tell a genuine local edit from an upstream divergence in the same file. Add the marker (write it on a successful apply, read it in check) and a test that exercises the locally-edited-file scenario, and I'll take another look.

santifer:
Direction is right, @Zoubeir23, and the two cases you found yourself in review are the reason I trust the rest: the untracked file at a path upstream *does* ship (invisible to both `git diff` calls) and the fully-excluded pathspec that aborts the whole update. Both are the kind of thing that only turns up when you actually run the thing.

One case blocks it, and it is a bad one: **a system file the user has DELETED locally.**

`git diff --name-only` lists deletions, so a deleted file lands in both of your sets and therefore in `atRisk`:

```console
$ git status --porcelain
 D modes/cover.md
$ git diff --name-only HEAD -- modes AGENTS.md
modes/cover.md
$ git diff --name-only upstream -- modes AGENTS.md
modes/cover.md
```

From there the sequence is:

1. `copyFileSync` throws ENOENT, and your handler correctly refuses to abort the update: `could not write modes/cover.md.bak: ENOENT`.
2. But the file **stays in `atRisk`**, so without `--force` it goes into `preservedPaths` (line 1103) and prints `Keeping your versions` about a file that does not exist.
3. Preserved means excluded from the checkout, so the file is **not restored**. Before this PR the raw checkout brought it straight back.
4. `apply()` then reports `Update incomplete` and exits 1.

Re-running does not help, because the state that produces it is unchanged: the file is still missing, still in both diffs, still preserved. The remedy the failure path prints (`run apply again`) can never work, and the one that does work (`--force`) is not mentioned anywhere in that message. So the user's install is stuck until they read the source.

The fix is small and I think it is one line plus a test: **a path that does not exist on disk is not at risk of being overwritten**, so filter it out of `atRisk` before the backup loop. The `.bak` failure branch then goes back to meaning what you wrote it for, a genuinely unwritable backup (permissions, full disk), rather than doubling as the deleted-file path.

Worth saying why the suite did not catch this: nothing currently asserts that `apply` restores a system file the user deleted. That property predates your PR and is exactly what your change alters, so a case for it belongs here.

Everything else stands up. I merged your branch onto today's `main` and ran the full suite: `3413 passed, 0 failed, 1 warnings`, the warning being the pre-existing `cv-sync-check` one that needs user data. The `--force` semantics, the `.bak`-before-overwrite and the ctx-injectable detection are all what the issue asked for.

Push the deleted-file case and I will merge it the same day.

Zoubeir23:
@santifer Pushed the deleted-file case.

A path that is not on disk is filtered out of `atRisk` before the backup loop, so a deleted system file is no longer preserved and the checkout restores it. That also gives the `.bak` failure branch back its single meaning — a backup that genuinely could not be written.

Three cases added, including the property you named as missing: case 11 runs the real checkout with the real exclusions and asserts the deleted file comes back *and* that a genuine local edit alongside it survives. Cases 10 and 11 verified red before the fix; case 12 is the guard that the existence filter only drops missing paths, and it passes both before and after.

Full suite: 2990 passed, 0 failed.

santifer:
Merged, @Zoubeir23. The closed loop is dead: a deleted system file no longer enters atRisk, so it stops being 'preserved' into nonexistence, and the normal checkout restores it. Cases 10-12 pin the property the suite never covered, and they were verified red against the unfixed code: that is the standard. 🚀

PR #2449, opened 2026-08-03
docs(scripts): document upgrade-tests.mjs (#2424)

### LINKED ISSUE #2424: docs: document upgrade-tests.mjs in docs/SCRIPTS.md

`upgrade-tests.mjs` just landed (#2358): it's the harness that, on every PR, takes the newest release, seeds a user install from that era, applies the commit under review, and proves the user's data survived. It isn't documented anywhere yet.

### What to do
1. Open `docs/SCRIPTS.md` and add an entry for `upgrade-tests.mjs` in the section for scripts with no `npm run` binding.
2. Cover its two modes: `--pr-gate` (the CI gate: newest release tag that's an ancestor of HEAD → this commit) and `--canary` (plants a user-file clobber and asserts the harness catches it, so a silent harness can't pass as a green one).
3. Say what it protects: `update-system.mjs` is the highest-blast-radius script in the repo, and this is the only test that exercises a real upgrade against a seeded install.

### Acceptance
- A reader who has never seen the file knows what it runs, when CI runs it, and why the canary exists.
- No new claims: everything you write is visible in `upgrade-tests.mjs` or `.github/workflows/test.yml`.

### Pattern to mirror
Any nearby entry in the same section of `docs/SCRIPTS.md` — same voice, same length. `docs/SCRIPTS.md`'s `fix-slugs` entry (added in #2329) is a good recent example of the level of detail we like.

Zoubeir23:
Fixes #2424

`upgrade-tests.mjs` landed in #2358 and was documented nowhere. It is the only test that exercises a *real* upgrade against a seeded install rather than asserting on `update-system.mjs`'s source — for the script with the largest blast radius in the repo — so a reader hitting a red `upgrade-gate` job had nothing to read.

## What the entry covers

- **Both modes.** `--pr-gate` (newest release tag that is an ancestor of `HEAD` → this commit) and `--canary` (plants a user-file clobber and requires the harness to report it).
- **What a leg actually asserts.** Read off `runLeg`'s `ok()` calls rather than summarized as "user data survives": `apply` exits 0; a genuinely changed system file now carries the target blob (the non-vacuity oracle, never VERSION); user files byte-identical; new manifest paths present; the tracker parses with the expected row and status counts; salary observations parse; `doctor.mjs --json` reports `onboardingNeeded: false`.
- **Why the canary exists.** A gate never seen red proves nothing — and the inverted polarity is worth spelling out, since "CANARY GREEN" means the harness *detected* the planted damage.
- **Where CI runs them:** the `upgrade-gate` job in `.github/workflows/test.yml`, with `fetch-depth: 0` because the harness needs release tags.
- Quick Reference row added alongside the other no-`npm run` entries, in the `node fix-slugs.mjs` style.

## Acceptance criteria

- [x] *"A reader who has never seen the file knows what it runs, when CI runs it, and why the canary exists"* — the entry answers all three, in that order
- [x] *"No new claims: everything you write is visible in `upgrade-tests.mjs` or `.github/workflows/test.yml`"* — the hermetic setup and self-reexec come from the file header, the assertion list from `runLeg`, the poison mechanics from `canary()`, the trigger and `fetch-depth: 0` from the workflow. Nothing inferred.
- [x] Mirrors the `fix-slugs` entry named in the issue: same voice, comparable length, prose plus one fenced invocation block

## Test plan

- [x] Docs-only change; no script, mode or workflow touched

## Summary by CodeRabbit

* **Documentation**
  * Added command reference and usage guidance for the dynamic upgrade regression harness.
  * Documented validation checks, data-preservation guarantees, supported execution modes, and CI integration.

### DIFF
--- a/docs/SCRIPTS.md
+++ b/docs/SCRIPTS.md
@@ -50,6 +50,7 @@ All scripts live in the project root as `.mjs` modules. Most are exposed via
 | `npm run archive` | `archive-posting.mjs` | Save a live job posting as PDF before it disappears |
 | `npm run prepare:application` | `prepare-application.mjs` | Print an ATS prefill summary (read-only, never POSTs) |
 | `npm run build:dashboard` | `build-dashboard.mjs` | Build the Go TUI dashboard binary cross-platform |
+| `node upgrade-tests.mjs --pr-gate` | `upgrade-tests.mjs` | Upgrade an install seeded from the newest old release to this commit and prove user data survived (CI gate; `--canary` proves the gate can fail) |
 
 ---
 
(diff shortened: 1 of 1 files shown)


santifer:
Sharp doc, @Zoubeir23: you explained the canary's inverted semantics right (green = the harness caught the planted clobber), which is exactly the part everyone gets backwards, and the non-vacuity oracle paragraph matches how the gate really decides. Merged. 🚀

PR #2451, opened 2026-08-03
test(lock-matrix): consume contention.wait()'s boolean instead of discarding it (#2436)

### LINKED ISSUE #2436: test(lock-matrix): assert contention.wait()'s boolean instead of discarding it
Follow-up to #2398. santifer's post-merge comment on that PR:

> the harness discards the boolean from contention.wait(), which is the actual discrimination signal — if the recover guard ever stops being emitted, the six non-output cases quietly burn the full 2s wait instead of failing loudly.

Verified against current main. `watchForContention()` in `tracker-writer-lock-tests.mjs` polls for the lock's `.recover` guard directory, and its `wait(timeoutMs)` returns whether the guard was ever seen. That boolean is the evidence that the spawned writer reached the contended lock before the fixture mutation landed, which is the ordering the whole matrix depends on: the mutation only tells a pre-lock read apart from a post-lock read if it lands after the writer's read.

The only call site throws it away (line 221):

```js
await contention?.wait(CONTENTION_WAIT_MS);
```

If `acquireTrackerLock` ever stops emitting the recover guard (say it gets renamed, or becomes conditional), every guard-watched case waits the full 2s of `CONTENTION_WAIT_MS` and then falls back to the old timing-dependent ordering: normalize-statuses, dedup-tracker, tracker-delete, tracker-export, set-status, mark-pdf-ready, reply-watch, reply-watch-identical. That's 8 invocations across 6 writer scripts; only reply-watch-stale-status takes the stdout-prompt path and skips the watcher. The suite stays green, or goes back to flaking, with nothing pointing at the missing signal. The fallback itself is documented as deliberate in the `watchForContention` comment ("degrades to its previous timing-dependent behaviour instead of hanging") and that part is fine. The problem is that nothing consumes the signal when the degradation happens. Same family as the unconsumed-guard fixes in #2304, #2306, and #2313.

Proposal: keep the boolean per case and fail when it's false.

```js
const sawContention = contention ? await contention.wait(CONTENTION_WAIT_MS) : true;
if (!sawContention) {
  fail(`${name}: recover guard never appeared, so the mutation ordering signal is gone`);
}
```

Every writer in the matrix reaches the lock promptly by design, and the probe run already proves each one contends, so a hard fail looks right to me rather than a warning. Filing this before cutting a PR per the issue-first rule, in case there's an argument for the softer option. When the signal does break, this also turns roughly 16s of silent dead waiting into a first-pass failure.

Zoubeir23:
Fixes #2436 — the follow-up you asked for on #2398:

> the harness discards the boolean from `contention.wait()`, which is the actual discrimination signal — if the recover guard ever stops being emitted, the six non-output cases quietly burn the full 2s wait instead of failing loudly.

The boolean is now consumed. **But at run level, not per case** — the issue proposed per case, I implemented that first, and CI proved it wrong. The finding is worth reading before the diff.

## Why per-case fails on a healthy tree

`acquireTrackerLock` creates the recover guard and removes it in a `finally` around a single `lockCanRecover()` call (`tracker-utils.mjs:388-398`). It exists for well under a millisecond, re-created on each ~`retryMs` attempt. `watchForContention` **samples** for it with `readdirSync` every 2ms — so whether a given case observes it is a race the sampler can lose.

Locally that race is always won (8 of 8 observed). On the Windows runner it is not:

```
FAIL normalize-statuses:    recover guard never appeared within 2000ms
FAIL dedup-tracker:         …
FAIL set-status:            …
FAIL reply-watch:           …
FAIL reply-watch-identical: …
```

**3 of 8 observed** — so the guard *is* emitted on Windows, the sampler just misses it. A per-case hard fail turns a healthy tree red there.

That also explains something I had written off as a flake earlier today: `tracker-writer-lock-tests.mjs` timing out with SIGTERM on the Windows leg of an unrelated PR. Those cases have been burning the full 2s each and falling back to timing-dependent ordering on Windows all along — silently, which is exactly the degradation this issue is about. It was already happening; nothing consumed the signal, so nobody saw it.

## What the assertion does instead

Across a run, the two causes separate cleanly:

- **a sampling miss** still leaves other cases observing the guard → the run stays green, and each miss prints a `NOTE` line so the degradation is visible where it happens;
- **the regression this must catch** — guard renamed, removed, or made conditional — takes *every* case to zero → red;
- **zero guard-watched cases at all** — a matrix change that drops them → red too. Skipping the assertion there would have reproduced the exact defect this PR fixes: green suite, nothing validating the guard (CodeRabbit review). Verified by forcing the watcher off for every case: `26 passed, 1 failed`.

```
PASS recover guard observed in 8/8 guard-watched cases — the mutation ordering signal is live
```

## Empirical proof

Renaming the guard prefix in `watchForContention`, simulating "the guard stops being emitted":

| | Result |
|---|---|
| `main` | **26 passed, 0 failed** — completely silent while the ordering signal is gone |
| this branch | **26 passed, 1 failed** — `recover guard never observed in any of the 8 guard-watched cases` |

## On the hard-fail question

The issue asked whether a hard fail or a warning was right, and said "a hard fail looks right to me". It still is — the run-level assertion *is* a hard fail. Only its granularity changed, because per-case granularity asserts something the signal cannot support. If you would rather have the per-case fail and accept red Windows legs, it is a three-line change.

## Test plan

- [x] `node tracker-writer-lock-tests.mjs` — 27 passed, 0 failed (26 + the new run-level assertion)
- [x] Zero-watched-case path verified by forcing `contention` to null everywhere: red, with `no guard-watched case ran`
- [x] Regression simulated by renaming the guard prefix: red here, green on `main`
- [x] No change to `watchForContention`, `CONTENTION_WAIT_MS`, `acquireTrackerLock`, or any lock parameter under test

### DIFF
--- a/tracker-writer-lock-tests.mjs
+++ b/tracker-writer-lock-tests.mjs
@@ -832,12 +832,16 @@ await testLiveRecoverGuardIsNotEvicted();
 // suite stays green (or returns to flaking) with nothing pointing at the cause.
 // One observation is enough to prove the signal exists; zero across the whole
 // matrix is the regression.
-if (contentionWatchedCases > 0) {
-  if (contentionObservedCases > 0) {
-    pass(`recover guard observed in ${contentionObservedCases}/${contentionWatchedCases} guard-watched cases — the mutation ordering signal is live`);
-  } else {
-    fail(`recover guard never observed in any of the ${contentionWatchedCases} guard-watched cases — acquireTrackerLock has stopped emitting it, so every one of them fell back to timing-dependent ordering`);
-  }
+if (contentionWatchedCases === 0) {
+  // Skipping the assertion when nothing was watched would reproduce the very
+  // defect this file is fixing: a matrix change that drops every guard-watched
+  // case leaves the suite green while nothing validates the recover guard at
+  // all. Zero watched cases is itself the regression (CodeRabbit review).
+  fail('no guard-watched case ran — the matrix no longer exercises the recover guard, so nothing validates the mutation ordering signal');
+} else if (contentionObservedCases > 0) {
+  pass(`recover guard observed in ${contentionObservedCases}/${contentionWatchedCases} guard-watched cases — the mutation ordering signal is live`);
+} else {
+  fail(`recover guard never observed in any of the ${contentionWatchedCases} guard-watched cases — acquireTrackerLock has stopped emitting it, so every one of them fell back to timing-dependent ordering`);
 }
 
 console.log(`\n${passed} passed, ${failed} failed`);
--- a/tracker-writer-lock-tests.mjs
+++ b/tracker-writer-lock-tests.mjs
@@ -218,7 +218,19 @@ async function runWhileLocked({
     // entries already have a stronger, script-specific ordering signal (their
     // pre-lock review prompt), and a writer parked at that prompt has not
     // reached the lock yet, so the guard wait is skipped for them.
-    await contention?.wait(CONTENTION_WAIT_MS);
+    //
+    // The boolean IS the discrimination signal, so it is consumed rather than
+    // discarded (#2436). If the recover guard ever stops being emitted — a
+    // rename, or the guard becoming conditional — every guard-watched case
+    // would burn the full CONTENTION_WAIT_MS and then silently fall back to
+    // the old timing-dependent ordering: the suite stays green, or goes back
+    // to flaking, with nothing pointing at the missing signal. The fallback
+    // inside watchForContention is deliberate (it degrades instead of
+    // hanging); what was missing is a consumer that says so out loud.
+    const sawContention = contention ? await contention.wait(CONTENTION_WAIT_MS) : true;
+    if (!sawContention) {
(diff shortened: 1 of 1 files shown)


santifer:
Merged, @Zoubeir23 — and the mutation evidence is what sold it: guard-rename turns your branch red exactly where main stays silently green, and the zero-watched-cases branch kills the vacuous-green failure mode (the same class we hunted in #2408). @Scott-Emberson heads-up on a conscious deviation from your #2436 per-case proposal: run-level granularity, because a 2ms-sampled sub-millisecond signal measured 3/8 observed on a healthy Windows tree — per-case would trade partial-degradation blindness for false reds. The 3-line per-case alternative is in the PR body; door open if you disagree. 🚀

Scott-Emberson:
Run-level is the right call. My #2436 proposal conflated a case *contending* with the sampler *observing* the transient guard, and your 3/8 measurement settles it: per-case would have shipped false reds on healthy Windows trees. One residual worth knowing about rather than changing: at that observation rate a healthy run misses all 8 cases roughly 2% of the time ((5/8)^8), so the run-level hard fail keeps a small false-red tail on Windows. If it ever fires in CI without a real regression, that's the first thing I'd suspect.

Issue #230, opened 2026-04-12
🔗 Scanner & Job Board Providers (umbrella)


vivek-gite:
## Scanner & Job Board Providers

career-ops's zero-token scanner (`scan.mjs`) hits ATS APIs directly. This umbrella tracks all work to expand the supported job boards and improve scanner resilience.

### ATS APIs (built-in)

- [x] Greenhouse API
- [x] Ashby API
- [x] Lever API
- [x] WebSearch queries

### Job Board Providers

- [ ] A16Z job board → PR #271
- [ ] Amazon Jobs → PR #254
- [ ] TheirStack API → #123

### Scanner Improvements

- [ ] Fallback URL search on 410/404 → #266
- [ ] Branded careers_url enforcement → #267

### How to contribute

1. Pick an unchecked item above
2. Comment here so others know you're working on it
3. New providers should be added to `scan.mjs` or as modules in a `providers/` directory
4. Update `templates/portals.example.yml` with example configuration
5. Submit a PR referencing this issue: "Part of #230"

vivek-gite:
Here's the a16z job board: https://jobs.a16z.com/jobs

darkpandawarrior:
Happy to be pitching in on the provider roster here 👋 — BambooHR (#1141) just merged, and I've opened **Breezy HR** (#1185) as the next zero-token provider (public `<tenant>.breezy.hr/json` board feed, live-validated, 13 tests).

One housekeeping note while I was in here: all five tracked sub-items above appear to be **done** already, even though the boxes are still unchecked — which is probably why this got auto-marked `stale`:

- A16Z → #271 ✅ closed
- Amazon Jobs → #254 ✅ closed
- TheirStack → #123 ✅ closed
- Fallback URL search on 410/404 → #266 ✅ closed
- Branded careers_url enforcement → #267 ✅ closed

Might be worth ticking those (and the roster has since grown well past the original list — recruitee, smartrecruiters, workable, workday, glints, jobstreet, ibm, arbeitsagentur, remoteok/remotive, plus bamboohr and the in-flight Teamtailor #1120). Glad to keep filling clean zero-token gaps under this umbrella — Personio (public `/xml` feed) looks like a reasonable next one if there's interest.

santifer:
Sorry for the silence, @vivek-gite: 121 days is too long for a first reply and there is no good excuse for it.

[...]

PR #2452, opened 2026-08-03
feat(providers): getManfred board-wide feed (Spanish/EU tech, salaries published)

Zoubeir23:
New zero-auth scanner provider — sent straight in per `CONTRIBUTING.md:22` ("Going straight to a PR is welcome — no issue needed — for: bug fixes, **new zero-auth scanner providers**, docs, and translations").

[getManfred](https://www.getmanfred.com) is a Spanish/EU tech board whose defining feature is that **every posting publishes its salary range** — which makes it the rare source where `salary_filter` actually has data to work on. The feed is public, keyless, and returns the whole catalogue in one request.

## What the live feed taught the design

Everything below was measured against the real endpoint on 2026-08-03, not assumed:

**1. It is a catalogue, not a live board.** 1,604 of its 1,628 entries carry `status: "CLOSED"`; **24 were ACTIVE**. Without the status filter the first scan would push ~1.6k dead postings into `pipeline.md`. This is the single decision that makes the provider usable, so it is the first test in the file.

**2. `lang` is mandatory.** Omitting it returns `400 {"message":["lang must be one of the following values: EN, ES"]}`. `detect()` and `fetch()` both always send it; the portal entry may set `lang: ES`.

**3. Currency is a symbol, not an ISO code** — and the live data includes a narrow-no-break-space variant glued to `€`. `scan.mjs`'s `salary_filter` compares currency strings, so `€` would never match a user's `currency: EUR`. Symbols are mapped to ISO, and an unknown symbol yields `''` rather than a guess.

**4. `updatedAt` is the only timestamp, and it is a modification time.** Mapping it to `postedAt` would make an edited old posting look freshly published to the recency filters, so **no `postedAt` is emitted** — the `Job` contract explicitly allows omitting it.

**5. 932 of 1,628 offers list no place**, leaving `remotePercentage` as the only location signal. Remote (100) and hybrid (1-99) are kept **distinguishable** rather than collapsed into "Remote": the emitted string is what `location_filter` matches on, and collapsing them makes a `block: ["Hybrid"]` rule unmatchable. That is the same defect I fixed in the echojobs provider in #2307 (#2258), so it is designed out here rather than shipped and fixed later. A placeless on-site offer (0) keeps `""`, which passes under the scanner's "don't penalize missing data" convention.

## Shape

- URL built as `/ofertas-empleo/{id}/{slug}` — the canonical form published in the site's own `sitemap-offers.xml`, verified `200`. An offer missing `id` or `slug` is dropped rather than linked to a guessed URL, since `url` is the dedup key.
- Host-pinned with `redirect: 'error'`, validating the URL actually fetched rather than a constant — the same SSRF discipline as `echojobs`.
- `robots.txt` is `Allow: /`, and the endpoint is namespaced `/api/v2/public/`. No scraping, no bot-protection bypass, no key.

## Test plan

- [x] `tests/providers/manfred.test.mjs` — 22 assertions covering all five behaviours above plus the usual contract (drops, company fallback, single request, unexpected shape throws)
- [x] Run against the **real 1,628-offer payload**: 1,628 → 24 normalized, 24 unique URLs, salaries parsed to `EUR`/`GBP`, locations spread across `Madrid, Spain` / `Remote` / multi-city joins
- [x] One generated URL fetched live: `200`
- [x] `node validate-portals.mjs --file templates/portals.example.yml` — 0 errors, 0 warnings
- [x] `node test-all.mjs --quick` — 2805 passed, 0 failed
- [x] Provider auto-discovers (`providers/*.mjs`) and ships via the existing `providers/` entry in `SYSTEM_PATHS` — no registration to forget

Docs: board table row in `docs/SUPPORTED_JOB_BOARDS.md` and a commented stanza in `templates/portals.example.yml`, both mirroring the `echojobs` entry.

## Summary by CodeRabbit

* **New Features**
  * Added getManfred as a supported job board.
  * Supports optional language selection and displays active job offers with normalized titles, companies, locations, compensation, currencies, and canonical links.
* **Documentation**
  * Added a configuration example for enabling getManfred in English mode.
* **Bug Fixes**
  * Excludes inactive or malformed offers and handles unexpected responses safely.


santifer:
Provider verified live end-to-end, @Zoubeir23: keyless 200 with 1,628 offers, the NNBSP-euro variant is real (U+202F U+20AC, your oddest claim checked out), the SSRF pin rides the URL actually fetched and `_http` honors `redirect:'error'`, and the dead-board contract holds by construction for a single-request provider. Your mock even asserts the SSRF guard itself — nice.

One thing before merge, and you're literally the author of the precedent: `resolveLocation` (lines 122-133) lets a listed city win and silently drops the hybrid marker. 8 of today's 22 ACTIVE offers (36%) are city+hybrid and become invisible to `block:['Hybrid']` — the exact half-working case your echojobs fix solved by appending ` · Hybrid`. One branch + one test and this lands.

Nit while you're in there: `manfred` sits between remotive and thehub in the example file's alphabetical provider comment; it belongs after landingjobs.

santifer:
Great addition, @Zoubeir23: a Spanish/EU board with published salaries fills a real gap in the provider map, and the contract tests + docs row + example stanza made this a one-pass review. Zero-auth verified live. Part of #230. Merged. 🚀

PR #2459, opened 2026-08-03
fix(cli): --flag=value was silently ignored in five scripts

Zoubeir23:
Bug fix, sent straight in per `CONTRIBUTING.md:22`. Same defect class as #2401, which #2402 fixed in `weekly-digest.mjs` and `company-history.mjs` — the note in `weekly-digest.mjs:82` says as much. It is still live in five other scripts.

## The defect

`args.indexOf('--flag')` returns `-1` for `--flag=value`, so the lookup silently discards the value the caller supplied and the script runs with its default — **reporting a result for inputs nobody asked for, with no warning.**

Each line below is the same command run twice on `main`, once per form:

| Script | `--flag value` | `--flag=value` |
|---|---|---|
| `process-quality --file` | `companies: 1` | `companies: 0` — read the default tracker |
| `check-table-freshness --today` | `today: 2099-01-01` | `today: 2026-08-03` — used the real date |
| `detect-reposts --window 5` | `window: 5 days` | `window: 90 days` |
| `validate-portals --file` | validated `templates/portals.example.yml` | validated **`portals.yml`** |
| `test-all --only` | ran the subset | ran the **whole suite** |

`check-table-freshness` is the sharpest: it is the staleness validator for the jurisdiction tables, and `--today` is how you check what a table will look like at a future date. Given `--today=…` it silently answers about today instead.

`test-all --only` also defeats a stated invariant. `runDiscovered` exits 1 when a filter matches nothing — *"a path typo must never silently turn CI green"* — but a dropped filter is not an empty match, it is no filter: the run goes green having tested everything except what was asked about.

## Fix

`lib/cli-flags.mjs` exports `flagValue(args, flag)`, accepting both forms, with the `=` branch checked first (checking `indexOf` first would let the space lookup shadow it — the same bug in reverse). The five scripts call it.

`weekly-digest.mjs` and `company-history.mjs` keep their local copies here: they are already correct, and rewriting two working files would enlarge the diff without changing behaviour. Consolidating them onto this module is a natural follow-up.

## Scope, stated honestly

A grep finds ~30 value-taking flags read this way across ~21 scripts. **This PR fixes the five I could demonstrate**, each with the before/after above. The rest need data I do not have locally to produce an observable difference (`contacts --vcf`, `upskill --min-reports` and others exit early without a populated tracker), so I am not claiming them. The helper makes each one a one-line change when someone can reproduce it.

## Test plan

- [x] New `tests/cli-flags.test.mjs` — 11 assertions: both forms, absent flag, `--file=` as supplied-but-empty (not absent), a value containing `=`, a trailing flag with no value, prefix collisions (`--filename` is not `--file`) in both forms, the `=` form winning over a bare one, and non-array / non-string argv defended
- [x] All five scripts re-run in both forms after the fix — identical output, table above
- [x] `check-table-freshness --max-age-months=abc` still fails fast with its usage error (the fix must not turn a bad value into a silent default)
- [x] `test-all --only=providers/zzz` still exits 1 on an unmatched filter
- [x] `lib/cli-flags.mjs` added to `SYSTEM_PATHS` — five shipped scripts import it, and an install receiving them without the module would crash on startup (#2235 failure mode)
- [x] `node test-all.mjs --quick` — 2794 passed, 0 failed

## Summary by CodeRabbit

* **New Features**
  * Added consistent command-line option handling across maintenance and validation tools.
  * Options support both `--flag value` and `--flag=value` formats, including values containing equals signs.
  * Explicit empty values and omitted options are handled distinctly.
  * Shared option handling is included in system updates.

* **Bug Fixes**
  * Prevented similarly named options from being confused.
  * Preserved existing defaults, validation, filtering, and exit behavior.

* **Tests**
  * Added coverage for supported formats, missing values, duplicates, empty values, and invalid inputs.


santifer:
The helper is right and the bug is real, @Zoubeir23 — verified the space-form stays intact and the `=`-form now parses in all five scripts, with genuinely negative tests. One regression to fix before merge: in `check-table-freshness.mjs`, bare value-flags with nothing after them (`--today`, `--max-age-months` as the last arg) go from fail-fast exit 1 on main to silently defaulting exit 0 — the same silent-substitution class this PR exists to kill, and the invariant your own last commit declares. Make valueless value-flags fail loudly there too (plus a test pinning it) and this merges.

santifer:
That closes the full set, @Zoubeir23: six merges in one day, and this one fixes a silent-ignore that touched five scripts at once. Fitting that the only thing holding it back was the Windows false-red your own #2500 killed. Thanks for an outstanding run today. 🚀

PR #2460, opened 2026-08-03
fix(cv-facts): the anti-fabrication gate saw nothing in five shipped locales

Zoubeir23:
Bug fix, sent straight in per `CONTRIBUTING.md:22`. No issue existed for it — I found it while probing the fact gate for bypasses.

## The gate reports a pass without checking anything

Every claim pattern in `verify-cv-facts.mjs` is written with ASCII `\d`. A CV whose numbers are spelled in any other script therefore produces **zero** metric claims, and `auditClaims` returns `invented: []` — a clean pass over text it never examined.

Measured on `main`:

```
"Reached 94,772 users"    ->  ["94 772 users"]     ← checked
"Reached ９４，７７２ users"  ->  []                    ← full-width (ja, zh, zh-TW)
"Reached ٩٤٧٧٢ users"     ->  []                    ← Arabic-Indic (ar)
"Reached ९४७७२ users"     ->  []                    ← Devanagari (hi)
"Cut cost ٦٠٪"            ->  []                    ← Arabic percent sign
"Managed a ５５０K budget"  ->  []                    ← currency claims too
```

**Why this is not cosmetic.** `modes/pdf.md:42` calls this *"a hard gate before PDF rendering"*, and `generate-cover-letter.mjs` runs the same validator. The repo ships mode sets for `ar`, `hi`, `ja`, `zh` and `zh-TW`. For a candidate working in those markets, the one mechanism enforcing *"Keywords get reformulated, never fabricated"* is inert — and it says `pass`, which is worse than saying nothing: it manufactures confidence.

## Fix

`foldDigits()` runs inside `stripMarkup`, so it applies to the generated document **and** to the sources — it can only ever make more claims visible on both sides, never hide one.

- NFKC first: folds full-width digits and `％` outright.
- Then a table of Unicode decimal-digit block zeros (Arabic-Indic, Extended Arabic-Indic, Devanagari, Bengali, Tamil, Thai, Khmer, …), which NFKC leaves untouched — verified: `'٩٤٧٧٢'.normalize('NFKC')` is unchanged.
- Then the Arabic separators NFKC also ignores: `٪` → `%`, `٫` → `.`, `٬` → `,`.

**Thousands grouping came along for the ride, because it decided matching too.** The extraction pattern reads a number as `\d[\d,.]*`, so it stops at a space: `"16 181 users"` yielded `"181 users"`, a claim no source contains — a truthful CV failing the gate, the #2279 class. Space-grouped thousands are now joined before extraction, and comma grouping is folded at comparison time. Only a 1-3 digit group followed by exactly three digits is joined, so `"in 2026 100 users"` is left alone.

One visible consequence: the normalized claim string in the failure message no longer carries a separator (`94772 users`, not `94 772 users`). Two existing self-test expectations are updated for that, with a comment saying why.

## Test plan

- [x] 10 new self-test cases. **7 of them fail on `main`** — verified by porting the block onto `main`'s copy of the file and running it: fabricated full-width / Arabic-Indic / Devanagari / Arabic-percentage metrics all pass the gate there, as do the three grouping cases
- [x] The other direction is pinned too: a **truthful** localized CV (`١٦١٨١`, `１６，１８１`) must not be flagged — folding must not trade a bypass for a false accusation
- [x] `node verify-cv-facts.mjs --self-test` — 20 passed, 0 failed
- [x] `node test-all.mjs --quick` — 2783 passed, 0 failed (the fact gate is also exercised by `tests/cover-fact-gate.test.mjs` and `tests/nonmetric-fact-gate.test.mjs`)

## Not fixed here

A metric written in words (`"ninety-four thousand users"`) still passes unchecked. That is a different problem — spelled-number parsing across the shipped locales — and I would rather report it than half-solve it in this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recognition of numbers written with Arabic-Indic, Persian, Devanagari, and other localized digits.
  * Added support for localized separators, percentage signs, and grouped numbers.
  * Preserved prose containing standalone less-than characters during text processing.
  * Improved year and numeric claim comparisons.

* **Tests**
  * Expanded validation coverage for localized digits, grouped numbers, percentages, and year formats.

### DIFF
--- a/verify-cv-facts.mjs
+++ b/verify-cv-facts.mjs
@@ -403,6 +403,11 @@ function runSelfTest() {
   equal('space-grouped thousands compare equal', auditClaims('Reached 16 181 users', foldSource).invented, []);
   equal('ungrouped compares equal to a grouped source', auditClaims('Reached 16181 users', foldSource).invented, []);
   equal('a fabricated space-grouped number is still caught', auditClaims('Reached 94 772 users', foldSource).invented, ['94772 users']);
+  // Multi-group values fold in full: `.replace(/…/g)` evaluates each separator
+  // against the ORIGINAL string, where every group is preceded by a space, not
+  // a digit, so the lookbehind passes at each one (CodeRabbit asked).
+  equal('a multi-group number folds completely', auditClaims('Reached 1 234 567 users', 'Reached 1234567 active users.').invented, []);
+  equal('an eight-digit multi-group number folds too', auditClaims('Reached 12 345 678 users', 'Reached 12345678 active users.').invented, []);
   // A four-digit left part is a year, not a group: nothing is joined.
   equal('a year is not glued to the next number', auditClaims('Joined in 2026 100 users', foldSource).invented, ['100 users']);
 
--- a/verify-cv-facts.mjs
+++ b/verify-cv-facts.mjs
@@ -60,9 +60,72 @@ function readIfExists(path) {
   return existsSync(path) ? readFileSync(path, 'utf-8') : '';
 }
 
+// Unicode decimal-digit blocks, by the code point of their zero. Every claim
+// pattern in this file is written with ASCII `\d`, so a CV that spells its
+// numbers in any other script produced ZERO claims and the gate reported a
+// pass without having checked anything — in ar, hi, ja, zh and zh-TW, all of
+// which ship mode sets. NFKC alone is not enough: it folds full-width digits
+// (ja/zh) but leaves Arabic-Indic, Persian and Devanagari untouched.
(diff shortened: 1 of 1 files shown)


santifer:
Verified against main's own copy, @Zoubeir23: the gate really was 100% blind to non-ASCII digits in five shipped locales (٩٤٧٧٢, ９４，７７２ and ९४७७२ all sailed through as invented=[]) and your digit-folding kills all four classes plus the '16 181 users' grouped-thousands false positive as a bonus. Folding at the normalize layer instead of enumerating locales is exactly the house rule. Merged. 🚀

Issue #1597, opened 2026-07-06
feat(scan): JD-content fingerprinting — catch agency+direct double-listings before evaluation


Scott-Emberson:
## Problem

Follow-up to #1596. The same job can enter the pipeline twice **before any tracker row exists**: once as a direct company listing and once as an agency re-post with the employer name stripped. `scan.mjs` / `scan-history.tsv` / `detect-reposts.mjs` key on company+role, so the two listings look unrelated — and the candidate can evaluate (and apply to) both. Reveal-time dedup in the tracker (#1596) is damage control; this is the layer where the double-submission could actually be **prevented**.

## Proposal

Fingerprint JD **content** at scan/evaluation time — comp range, location, seniority, distinctive requirement phrases (evaluation reports already extract most of these) — and warn when a new posting's fingerprint matches an existing tracker row or recent scan-history entry with a different source. Agency posts strip the company name but rarely rewrite the requirements list.

Sketch:
- cheap shingle/minhash or normalized-keyphrase overlap over the JD body, stored per scan-history row
- threshold warning in scan output + a check in the auto-pipeline evaluation step ('this JD is 85% similar to #042 (Hays) — possible same role')
- zero LLM cost — plain text similarity

## Non-goals

- Not a dedup auto-merge; warn only.
- Not tracker-schema work (that's #1596).

PR #2465, opened 2026-08-03
feat(providers): add remotli — Swiss remote job board

### LINKED ISSUE #2464: Provider proposal: Remotli (remotli.ch) — Swiss remote job board, public JSON API
### Summary

Proposing a `providers/remotli.mjs` for **Remotli** (https://remotli.ch) — a curated board of remote roles at Swiss companies, paid in CHF. Public, no-auth JSON API at `/api/jobs`, so it fits the zero-token provider model.

Opening this first per CONTRIBUTING ("PRs that add external API dependencies without prior discussion in an issue").

### Disclosure

**I built and operate remotli.ch.** Flagging that up front so the conflict of interest is visible while you evaluate this. Happy for it to be routed to the plugin layer, or declined outright, if a maintainer-owned source is not something you want in core — the code works either way and I am not looking for a placement favour.

### Why it might fit core

CONTRIBUTING routes provider modules to the core, and says reading individual boards where employers post is what `providers/` is for. Remotli is a niche vertical board in the same shape as the existing `himalayas` / `remotive` / `4dayweek` providers, not a universal aggregation index used as a single unified pipe — the boundary drawn in #904 / #156.

It also fills a real geographic gap: ~~current Swiss coverage is `joinup.ch` and `startup.ch`, and there is no `jobs.ch` provider.~~

> **Correction (2026-08-01):** that sentence was wrong. `main` has no Swiss provider at all - `joinup.mjs` and `startupch.mjs` live in my own open PR #825, not in `main`, so I described unmerged work of my own as current coverage. The `jobs.ch` half was right. Details: https://github.com/santifer/career-ops/issues/2464#issuecomment-5178735426

### What the API supports

The list payload is unusually complete, which lets the provider satisfy several core invariants without extra requests:

| Capability | Field | Effect |
|---|---|---|
| Liveness at source | `status` | Only `active` rows are emitted, so no dead links enter the pipeline |
| Cross-listing dedup (#1597) | `description` | Full body ships in the list payload — SimHash fingerprint works with **zero** extra per-job requests |
| Recency | `publishedAt` | Feeds `max_posting_age_days` |
| Salary filter | `salaryMin` / `salaryMax` / `salaryCurrency` | Maps onto the `{min,max,currency}` shape `buildSalaryFilter` expects |
| Employer attribution | `company` | Real employer, not the board name, so tracker rows stay attributable |

Scale: 389 live postings, 8 pages at the server's 50/page cap.

### Two design choices worth review

1. **Emitted URL is the canonical board page, not the upstream `applyUrl`.** Every provider is host-locked to its own domain, and emitting arbitrary third-party hosts would break that invariant while handing the dedup key to a URL whose format may not match what the direct ATS provider emits anyway. Cross-listings get caught by the SimHash fingerprint instead. Happy to flip this if you would rather the dedup key point at the origin ATS.
2. **`toEpochMs` inlined rather than imported from `_dates.mjs`**, matching `personio.mjs`.

### Hardening

Mirrors the existing ATS providers: HTTPS-only, anchored `remotli.ch` host regex, `redirect:'error'` on every page, path-safe slug validation before URL interpolation, and the `fromCodePoint` try/catch wrapper from `personio.mjs` (`String.fromCodePoint` throws `RangeError` above `0x10FFFF` and `Number.isFinite` does not catch it, so an unguarded decode lets one malformed description abort a whole page fetch).

19 unit tests in `tests/providers/remotli.test.mjs` covering the host-lock, look-alike domains, path traversal in slugs, pagination and the `maxPages` health-probe cap, non-active rows, and the malformed-entity regression.

Docs row added per `providers/README.md` step 3.

PR to follow, linked to this issue. Glad to take the plugin route instead if that is the better door.

eliador90:
Closes #2464 — see that issue for the proposal, the routing rationale, and the maintainer disclosure.

### Disclosure

**I built and operate remotli.ch.** Repeating it here so it is visible on the PR as well as the issue. Route this to the plugin layer or decline it if a maintainer-owned source is not something you want in core.

### What this adds

`providers/remotli.mjs` — a zero-token provider for https://remotli.ch, a curated board of remote roles at Swiss companies (paid in CHF). Public, no-auth JSON API at `/api/jobs`, paginated `?page=N&limit=50&remote=all` (server caps `limit` at 50). ~921 live postings across 19 pages.

> **Correction (2026-08-08):** this originally read `?page=N&limit=50` and "389 live postings", which understated what the code does. `remote=all` is required for complete inventory: without it the API serves its human-facing remote-first default view (`fully_remote` + `remote_friendly`), about 392 of ~921 active listings. That is a default query filter on the API, not a gap in the board. The provider disables the filter and walks all 19 pages.

Rows come back in a nested `{ jobs, companies }` join shape; the posting itself is under `.jobs`.

### Why the API shape matters

- **`status`** — only `active` rows are emitted, so liveness is guaranteed at the source rather than inferred.
- **`description`** — ships in the list payload, so the #1597 cross-listing fingerprint works with no extra per-job request.
- **`publishedAt`** — feeds `max_posting_age_days`.
- **`salaryMin`/`salaryMax`/`salaryCurrency`** — mapped onto the `{min,max,currency}` shape `buildSalaryFilter` consumes.
- **`company`** — the real employer, so tracker rows stay attributable rather than collapsing to the board name.

### Two deliberate choices

1. **Emitted URL is the employer's `applyUrl`, not the board page.** Per Source Indexing Policy rule 2, the emitted URL is the shortest verifiable path to the employer: the original ATS page (Greenhouse, Ashby, Lever, Personio, …). The board's own page is the fallback, used only when a row has no usable `applyUrl`. Following `jobvite.mjs`, that URL is accepted from any `https:` origin and is deliberately NOT host-pinned, because it is display-only and never fetched by this provider; the host lock belongs on the API URLs actually requested (`assertRemotliUrl`). Non-https and malformed `applyUrl`s fall back to the board page rather than being trusted. Side benefit for dedup: a role cross-listed here and on the employer's direct ATS provider now resolves to the same URL and dedups exactly, rather than relying on the #1597 SimHash fingerprint to catch it.

   > **Correction (2026-08-08):** this bullet originally described the opposite behaviour ("emitted URL is the canonical board page, not the upstream `applyUrl`"). That was the pre-review design; it was flipped before merge to satisfy rule 2, and the description was never updated to match.
2. **`toEpochMs` inlined** rather than imported from `_dates.mjs`, matching the convention in `personio.mjs`.

### Hardening

Mirrors the existing ATS providers: HTTPS-only, anchored `^(www\.)?remotli\.ch$` host regex, `redirect:'error'` on every page, and path-safe slug validation before URL interpolation.

Entity decoding routes through the same `fromCodePoint` try/catch wrapper `personio.mjs` uses. `String.fromCodePoint` throws `RangeError` above `0x10FFFF` and `Number.isFinite` does **not** catch that, so an unguarded decode let a single malformed `&#99999999;` in one job description abort the whole page fetch and drop every job on the board. Caught in review with a regression test for both the decode and the mixed-page case.

### Tests

19 tests in `tests/providers/remotli.test.mjs` — auto-discovered, no registration needed. Coverage: `id`, `detect()` positive/negative (including look-alike hosts such as `remotli.ch.evil.example` and non-HTTPS), path-traversal slugs, field mapping, `allLocations` folding, salary ordering, non-active rows dropped, company fallback chain, pagination, the `ctx.maxPages` health-probe cap, `redirect:'error'` on every page, throwing on an unexpected response shape, and the malformed-entity regression.

Docs row added to `docs/SUPPORTED_JOB_BOARDS.md` per `providers/README.md` step 3.

## Summary by CodeRabbit

- **New Features**
  - Added Remotli as a supported job board.
  - Imported active listings with company, location, salary, publication date, and full descriptions.
  - Added pagination and configurable page limits for larger result sets.

- **Bug Fixes**
  - Improved handling of invalid, incomplete, or malformed job data.
  - Ensured imported listings use secure, valid links and correctly processed descriptions.

- **Documentation**
  - Updated the supported job boards documentation to include Remotli.


eliador90:
Force-pushed to fix the diff. The branch was cut from my fork's `main`, which diverged from upstream in April (my local updater applies releases as squashed commits rather than merging upstream history), so GitHub was diffing from a four-month-old merge base and showing ~1000 unrelated files.

Rebuilt the branch directly on `upstream/main`. The PR now contains one commit and three files. Sorry for the noise if a review already started on the old diff.

santifer:
Verified end-to-end, @eliador90 — the API answers live, the SSRF guard is real (anchored host regex + `redirect:'error'` actually forwarded by `_http`), the RangeError guard comes from the canonical `_html-entities` decoder, and your disclosure up front is exactly how this should be done. Three technical items before this can land, independent of anything else:

1. **Two HIGH CodeQL alerts (js/double-escaping #94, js/bad-tag-filter #95)** on `htmlToText`'s strip-decode-strip. Root cause is the SECOND tag-strip after `decodeEntities` — `agentic-jobs.mjs:81` and `avature.mjs:62` strip once BEFORE decoding and neither is flagged. Mirror that order and both alerts die. Main has zero open alerts of either rule, so this repo won't merge new HIGHs.

2. **Dead-board contract, second half**: first-request failure throws (correct), but the `succeededOnce` partial-keep idiom is absent — a transient failure on page 5 of 8 currently discards ~200 collected jobs and fails the whole target. `radancy.mjs` (from #2379) is the model, including "only a well-formed response counts as proof of life".

3. **Dead assertion**: `tests/providers/remotli.test.mjs:132` reads `… ? false : false` — both branches return false, so the content half of that check tests nothing. And the `entry.max_pages` fallback is unexercised.

One correction for the record: #2464 says current Swiss coverage is joinup.ch and startup.ch, but those providers aren't in main — they live inside your #825, which is my debt to review and is on the slate.

On the elephant: you operating remotli.ch is disclosed and the code shows no self-preference, but a board owner adding their own board to `providers/` is a policy precedent, and that call is Santiago's, not mine — I'm taking it to him separately. The three items above stand regardless of that outcome.

eliador90:
Taken in `693fcc8` — the status filter now fails closed. Two things worth flagging rather than leaving to be discovered in the diff.

**It is hardening, not a live defect.** The fail-open branch is unreachable against the real API: `/api/jobs` filters `status = 'active'` in SQL, and I checked all 8 pages — 395 of 395 rows carry `status: "active"`, none missing, none non-string. So nothing was being published that should not have been.

I took it anyway because the failure modes are not symmetric. Today the guarantee depends on an upstream invariant the provider cannot see. If that field ever disappeared, fail-open would silently publish closed roles as dead links — the exact staleness problem this provider exists to avoid — whereas rejecting unknown status yields a visibly empty board that `verify-portals` surfaces. For a liveness filter that is the right direction, and it costs nothing at 395/395.

**I inverted a test rather than adjusting one.** `normalizeRemotliJob keeps rows with no status field` asserted the old behaviour as a deliberate choice ("don't penalize missing data"). That choice is what is being reversed, so the assertion is now the opposite: absent, `null`, non-string, empty and whitespace-only status are all rejected. I added two positive cases alongside it so the check cannot pass merely by rejecting everything, plus one confirming status is trimmed and lowercased before comparison.

**One trap in applying it, since it is the same shape as the dead assertion you caught.** Ten test rows omit `status`. Making it mandatory without touching them would have left the slug, title, salary and company-fallback cases returning `null` for the *wrong* reason and still passing — tests that look green while covering nothing. Rather than edit ten call sites I defaulted `status: 'active'` in the `row()` helper, then mutation-tested the result: removing the slug guard, the title guard, or the status guard each makes a test fail. All three are genuinely covered, and the helper carries a comment saying why the default exists.

Provider tests 29, up from 27 (20 before your review). `test-all.mjs --quick`: 2802 pass, 1 unrelated failure on my machine — `analyze()` appDateSource symlinks `node_modules` into a temp dir, which Windows blocks without developer mode; fine on Linux CI.

That is all four items now: both CodeQL alerts, the dead-board contract, the dead assertion, and this. @santifer — the open question from your review is still the one about `double-encoded`: I removed that test because the input no longer reaches the provider once the board serves consistent HTML, but if you would rather keep it as an upstream-regression guard ("this is what a regressed API looks like"), say so and I will add it back explicitly rather than leave it implied by a comment.

santifer:
Following up with the policy outcome, @eliador90 — two conditions and this moves forward on the same footing as every other source:

1. **Canonical URL = the employer's applyUrl.** Your API already exposes `applyUrl` pointing at the original ATS — make that the `url` each job carries, with the remotli page as secondary attribution if you want it in the payload. Candidate-first: the shortest path to the employer. This applies to every source, not specially to yours.

2. **Operator declared.** The SUPPORTED_JOB_BOARDS row lists `operator: eliador90` — formalizing the disclosure you already made up front, which frankly set the standard for how this should be done.

That's the whole policy ask. Your pagination already walks the source's full inventory, which is the part that actually matters for neutrality, so nothing else changes there. The technical items from the earlier review still stand: the two CodeQL HIGHs (strip before decode, mirroring agentic-jobs/avature), the succeededOnce partial-keep, and the dead assertion at test line 132.

Get those in and this merges.

santifer:
The policy is now published: [CONTRIBUTING → Source Indexing Policy](https://github.com/santifer/career-ops/blob/main/CONTRIBUTING.md#source-indexing-policy). The two conditions above are its rules 2 and 4 applied — nothing new for you beyond what's already asked.

eliador90:
Both conditions are in, in `6517873`. Thank you for taking the time to write the policy rather than just ruling on this one PR — a published bar that applies to everyone is a much better outcome than a yes.

**Rule 2 — canonical URL.** Each posting now carries its `applyUrl` (the employer's own Greenhouse / Ashby / Lever / Personio page) as the emitted URL, with the remotli page as fallback. I followed `jobvite.mjs`, which already does exactly this and states the reasoning I needed: the emitted URL is display-only and never fetched, so the host lock belongs on the API URLs the provider *requests*, not on the ones it hands you. Non-https and malformed values fall back to the board page rather than being trusted. All 395 live rows carry an https `applyUrl`, so that fallback is defensive rather than routine.

One consequence worth naming: slug validation moved to the fallback path, since that is the only place a slug gets interpolated. A row with an unsafe slug but a usable `applyUrl` is now emitted and its slug simply never used — that felt more correct than dropping a real posting over a field we no longer use.

**Rule 4 — operator declared.** The `SUPPORTED_JOB_BOARDS` row now carries **`operator: eliador90`** and links the policy. There is no `operator` column in that table today, so it sits in the Notes cell; if you would rather it were a real column across all ~40 rows, that is your table to shape and I will follow. Ready for the out-of-band verification whenever you want it — I can reply from an address at the domain.

---

**Rule 3 — something you did not ask about, which I would rather tell you than have you find.**

You wrote that my pagination already walks the source's full inventory. It did not. It walked the board's *default* view, which is remote-first, and that was **395 of 925 active listings — about 43%**. The other 530 are hybrid (488), workation (41) and one row with no work mode at all.

That was not deliberate concealment — the provider requested the same default the site shows a human — but it is exactly the shape of thing rule 3 exists to catch, and "the operator's own provider quietly saw 43% of his board" is not a sentence I want someone else to write later.

Fixed rather than disclosed-and-left: the provider now sends `remote=all` and reads all 925. Note the audit method in rule 3 (API totals vs site totals) would **not** have caught this — both were 395, because the site applies the same default. If you want that audit to detect this class of gap generally, it needs a source-declared inventory total to compare against, not just the API's own default response.

Enumerating the four work-mode values would not have been enough either: it returns 924, missing the policy-less row, and would go stale whenever the board adds a work mode. So the board grew a flag meaning "no work-mode filter", which is live now, and the provider sends that. Filtering is the consumer's job per rule 5, so it now takes everything and lets `scan.mjs` decide.

---

**One process note.** Your comment lists the three technical items as still outstanding — they landed a couple of hours before it, in `eeeaa42` (11:56Z) and `693fcc8` (12:52Z), with a comment at 12:53Z. Flagging only so you are not holding the merge for work that is already in:

| Item | Commit |
|---|---|
| Both CodeQL HIGHs — strip once before decode, mirroring agentic-jobs/avature | `eeeaa42` |
| `succeededOnce` partial-keep, incl. "only a well-formed response counts as proof of life" | `eeeaa42` |
| Dead assertion at test line 132 + `entry.max_pages` coverage | `eeeaa42` |
| CodeRabbit's follow-up: status now fails closed | `693fcc8` |

CodeQL and `Analyze (javascript-typescript)` are green on the head, and all three `test` matrix jobs pass.

Provider tests are at 30, up from 20 before Santi's review. The one open question from that review is still his: I removed the `double-encoded` test because that input no longer reaches the provider, and I would now argue *against* adding it back — a unit test with a fixed input cannot detect an upstream regression, since nothing in the suite touches the live API. It would only pin current behaviour while asserting degraded output as correct. The dependency is guarded where it actually can be, in the board's own CI. If you want protection on this side, a runtime canary — flagging a normalised description that still contains `&lt;` — is the shape that would work, and I am happy to add it.

santifer:
Merged, @eliador90, and I want to be explicit about why, because you're the first source proposal to go through the Source Indexing Policy since it was published.

You operate remotli.ch and you said so up front. That disclosure is what made a rule-based review possible instead of a judgement call, and the rules are what I checked:

**Rule 2 (canonical URL is the employer's).** Verified substantively, not nominally: I sampled 121 rows across pages 1, 9 and 19. Zero rows without an applyUrl, zero non-https, and zero pointing back at remotli. The host distribution is real employer ATS endpoints (Greenhouse, Workday, Ashby, Lever, Recruitee, company career pages). Not one tracking hop.

**Rule 3 (complete inventory).** You disclosed the gap yourself and then closed it: without `remote=all` the feed serves 392 of 921 roles, 42.6% of the board. The fix walks all 19 pages.

**Rule 4 (declared operator).** The docs row carries `operator: eliador90` with a link to the policy.

Zero-auth confirmed live, and the two HIGH CodeQL alerts I flagged are resolved. One housekeeping note: the PR body still describes the old URL behaviour, so it now understates what the code does. Worth fixing if you ever point someone at it.

Sorry this took as long as it did. Being the first case through a new gate means absorbing the friction of the gate being written around you. 🚀

santifer:
One more thing, @eliador90, and it's a question rather than a request.

Your case is the first one to go through the Source Indexing Policy since we published it, and internally we think it's a good story about how the gate is supposed to work: the operator declared himself, the rules did the deciding, and the checks were run rather than assumed. We'd like to write about that.

Two separate permissions, and no is a perfectly fine answer to either:

1. **Being named.** Your operator declaration is already public in the docs table, but public and amplified are not the same thing, and that distinction is one we defend for candidates so it would be odd not to extend it to you. Happy to describe the case without your handle if you'd rather.
2. **The coverage detail.** The part we find most telling is that you disclosed the gap in your own proposal (392 of 921 roles without `remote=all`) and closed it before merge, rather than after someone caught it. From here that reads as the strongest possible signal about an operator. Read from your chair, "their board was serving under half its inventory" might land differently in public, so we're not going to publish that number without your say-so.

Nothing is written or scheduled. Take your time, and say no to either part without explaining why: it changes nothing about the merge, which stands on the rules being met.

eliador90:
Yes to both, @santifer.

**Being named:** go ahead. The operator declaration is already public and I am happy for it to be amplified.

**The coverage detail:** also fine to publish. One clarification so the number is not misread, offered as framing rather than as a condition: the 392 of 921 is the API's *default work-mode filter*, not a gap in the board itself. `/api/jobs` defaults to the remote-first view (`fully_remote` + `remote_friendly`), which is a sensible default for a human browsing a remote board; `remote=all` disables that filter and exposes the full inventory. "The provider needed an explicit parameter to see the whole board" is the accurate description.

Thanks for asking rather than assuming, and for splitting it into two questions instead of one. Extending candidates' public-versus-amplified distinction to contributors is a good instinct.

On the housekeeping: you are right that the PR body still describes the old URL behaviour and now understates what the code does. I will update it.

PR #2466, opened 2026-08-03
fix(salary-gap): parse verbose advertised_comp strings correctly

MyNameReallySux:
## Summary
- `parseAmount()` in `salary-gap.mjs` only stripped currency symbols and a single trailing 3-letter alpha token, so real-world `advertised_comp` strings pulled verbatim from a JD (e.g. `"Annual Salary Range: $165,000 - $190,000"`, `"$204.4K – $352K • Offers Equity"`, `"$220,000 - $400,000 USD per year"`) silently returned `null` instead of the actual range, breaking comp-gap math for those reports.
- Now also strips a leading label (up to the last colon), a trailing bullet/suffix, and a trailing cadence phrase (`per year`, `annually`, etc.) before the existing ISO-currency-code strip runs.
- Added regression coverage for all three real-world patterns in the existing `--self-test` suite; noted one known tradeoff in a code comment (a colon inside a trailing parenthetical note, e.g. `"$150,000 (note: negotiable)"`, would also get stripped — rare enough in practice not to be worth a full label/note disambiguation).

Found while running a batch of live evaluations — several `advertised_comp` values from real JDs failed to parse.

## Test plan
- [x] `node salary-gap.mjs --self-test` passes, including the new regression assertions
- [x] Existing self-test assertions (ranges, single values, ISO tokens, EUR/USD symbol-on-both-bounds) still pass unchanged

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved salary range parsing for entries with leading labels, trailing bullet markers, and salary cadence phrases.
  * More reliably extracts numeric salary values from varied formats.

* **Tests**
  * Added coverage for the newly supported salary formats.


santifer:
And sorry for the wait on this one too, @MyNameReallySux. Real-world `advertised_comp` strings are messy exactly the way you show (`$204.4K – $352K • Offers Equity`, ranges, verbose prefixes), and a `parseAmount()` that only strips a symbol and one trailing token will mangle most of them. CI is green. I'll get both your PRs reviewed this week — thanks for your patience, and for the test coverage on the parse cases.

brunosandrini98:
Hit the same bug from the other side of the Atlantic, so I checked this branch against my own data before opening anything of my own.

I applied the diff to a clean `v1.28.0` checkout and ran every `advertised_comp` value from my evaluation reports through both the baseline and your version — 63 values, all copied verbatim from live postings, EU-heavy:

```
baseline v1.28.0 : 2 parsed / 61 null
with this PR     : 5 parsed / 58 null
```

Of the 58 remaining, 13 legitimately carry no figure (`"not stated"`, `"Competitive salary…"`) and should stay `null`. **That leaves 45 values that contain a real number and still don't parse**, across 47 distinct shapes.

That's not a criticism of the patch — it does exactly what it says, and the three it fixed for me were all in the trailing-cadence class you targeted. It's that the EU shapes cluster around patterns the US-form heuristics don't reach.

### What's left, grouped by pattern

Examples below are synthetic — same shapes, invented numbers:

| Pattern | Example |
|---|---|
| Leading ISO code, no colon | `EUR 50000` · `GBP 50000-60000 + equity` |
| `gross` + cadence | `50,000-60,000 EUR gross/year` · `€50,000-60,000 gross annually` |
| Trailing qualifier | `EUR 50000-60000 base` · `50000-60000 EUR OTE` |
| Period unit, not annual | `$40-60/hr` · `200-300 EUR/day` · `€4,000 - €5,000 gross per month` |
| Parenthetical scope | `€50,000 – €60,000 (Spain)` · `EUR 50-60K (EU) + equity` |
| `up to` prefix | `Up to EUR 60K gross/year` |
| Trailing qualifier clause | `€50,000-60,000 gross annually, based on experience` |
| Multi-band / multi-currency | `L2 €50,000-60,000 / L3 €70,000-80,000 gross (Europe Remote)` |

The leading-ISO and `gross`-cadence rows are the bulk of it. The multi-band ones are genuinely hard, and I'd argue they should keep failing closed rather than be guessed at.

I'd rather not paste my actual report values on a public thread, but I'm glad to put the shapes into a fixture in whatever form is most useful to you.

### One hazard worth flagging before the stripping widens further

`toNum` reads `,` as a thousands separator but `.` as a decimal point, so on both baseline and this branch:

```
parseAmount("€35.000 - €45.000")  ->  { min: 35, max: 45, mid: 40 }
```

Period grouping is the convention across much of the EU, so this is squarely in the same territory this PR is opening up. Three of my values are period-grouped, and right now they're safe **only because** their trailing prose makes the anchored match reject the whole string — they fail closed into `quality.unparseable`.

Once the prose stripping is wide enough to reach them, which is where this is heading and what the EU shapes above need, they flip from visibly dropped to silently wrong by 1000× — straight into `advToActPct` and the `byCurrency` aggregates with no warning.

So I don't think it's a problem with your patch, but it does look like an ordering constraint: the separator fix wants to land before or alongside any further widening. #2722 already settled this class in `verify-cv-facts.mjs` — a separator is grouping only when followed by exactly three digits — so there's a house-approved rule to reuse rather than relitigate. I've filed it separately rather than piling onto this thread: #3174

### Offer

Happy to contribute the EU shapes and the separator guard rather than open anything competing. Whichever suits you:

- a PR into this branch, so it lands as part of #2466 with your work intact;
- a follow-up PR against `main` once this merges;
- or just take the pattern table as test material and ignore me entirely — no obligation.

@santifer, if you have a preference on sequencing given you mentioned reviewing this week, I'll follow that.

santifer:
That is independent validation with real data, @MyNameReallySux: @brunosandrini98 hit the same bug from his own tracker, applied your branch to a clean v1.28.0 and ran his actual advertised_comp values through it before saying anything. A draft that survives someone else's data unprompted is usually ready to stop being a draft — mark it ready whenever you are and it goes into normal review with that validation already on the record.

MyNameReallySux:
Hey @santifer,

Didn't realize this was still a draft 🫠 . Now ready for review!

santifer:
Marked ready a day ago and I didn't pick it up, @MyNameReallySux. Sorry.

It's in the review queue for the next merge batch rather than sitting unowned, and there's nothing needed from you in the meantime. Please **don't rebase**: the branch will show as behind because of my own merges today, and I'll bring it up to date myself when it's its turn.

For what it's worth, the "didn't realize this was still a draft" is a failure of my process rather than yours: a draft that's actually finished is invisible to every radar I run, because they all filter drafts out on the assumption that a draft means unfinished work. You marked it ready the moment you noticed, which is all anyone can do.

santifer:
Still wanted, @MyNameReallySux: the label/bullet/cadence preprocessing here fixes real JD-verbatim strings that `parseAmount` chokes on. Heads-up before you touch it again: #3239 merged today and rewrote the separator half of `parseAmount` (a `canonicalizeSeparators` that decides comma-vs-period before deleting either), so this will want a rebase, and one interaction to check: your removal of the trailing 3-letter ISO strip: #3239's self-tests include `'120.000,00 EUR'`, so if that strip goes, that test tells you immediately. Your preprocessing runs BEFORE the separator logic, so the two changes should compose cleanly. Happy to review as soon as it's rebased.

santifer (changes requested):
Hi @MyNameReallySux, this PR no longer merges cleanly with `main`.

We tried to rebase it on our side (you allow edits by maintainers, thank you for that), but the conflicts need a judgement call that only you can make: they touch lines your change and a recent merge both rewrote. So this one needs a rebase from you. There is no clock on this: the PR stays yours and stays open.

- [ ] Rebase on `main` (or merge `main` into your branch) and resolve the conflicts
- [ ] Run `node test-all.mjs --quick` locally
- [ ] Push; CI will re-run and we will pick it up

If you want a hint on which side to keep, ask here and we will answer on the diff.

PR #2467, opened 2026-08-03
fix(update): add sync-pdf-flags.mjs to SYSTEM_PATHS

MyNameReallySux:
## Summary
- `merge-tracker.mjs` imports `sync-pdf-flags.mjs`, but that file was never added to `update-system.mjs`'s `SYSTEM_PATHS` manifest.
- Any client that reaches `sync-pdf-flags.mjs` only via `update-system.mjs apply` (rather than a fresh clone) never gets the file checked out, so `merge-tracker.mjs` crashes with `Cannot find module 'sync-pdf-flags.mjs'` on every run after updating.
- Confirmed with the repo's own `validate-system-paths-coverage.mjs`, which currently fails on `main` with exactly this one orphan file (`node validate-system-paths-coverage.mjs` → `Coverage gap — tracked files not in SYSTEM_PATHS or USER_PATHS: sync-pdf-flags.mjs`).
- One-line fix: add `sync-pdf-flags.mjs` to `SYSTEM_PATHS`. (`tests/sync-pdf-flags.test.mjs` is already covered by the existing blanket `'tests/'` entry, so no separate line needed for it.)

Found while updating a client from v1.10.0 → v1.22.0 — `merge-tracker.mjs` crashed immediately after the update completed successfully.

## Test plan
- [x] `node validate-system-paths-coverage.mjs` now passes (was failing before this change)
- [x] `node updater-migration-tests.mjs` — 300 passed, 0 failed

## Summary by CodeRabbit

* **New Features**
  * Included PDF flag synchronization in automatic system updates.

### DIFF
--- a/update-system.mjs
+++ b/update-system.mjs
@@ -225,6 +225,7 @@ const SYSTEM_PATHS = [
   'batch/README.md',
   'utils/token-tracker.mjs',
   'batch-tailor.mjs',
+  'sync-pdf-flags.mjs',
   'dashboard/',
   'templates/',
   'config/cv-facts.example.json',


santifer:
Sorry for the long silence, @MyNameReallySux — this sat far too long in a busy stretch, and it's a real gap you caught. `merge-tracker.mjs` importing `sync-pdf-flags.mjs` while that file isn't in `SYSTEM_PATHS` is exactly the bug-class we've hit before: a fresh clone is fine, but any install that only gets files via `update-system apply` never receives it and breaks silently on upgrade. This is the right fix and CI is green. Reviewing it properly now.

santifer:
You were right, @MyNameReallySux, and main agrees with you: `sync-pdf-flags.mjs` got registered in SYSTEM_PATHS on July 28 (b2f3fda4, as part of the coverage-guard work in #2240), so this landed while this PR sat unreviewed — which is on us, not you. Closing as already-shipped rather than unwanted. Your #2466 is still live and still needed; leaving a note there about a rebase it'll want.

Issue #1624, opened 2026-07-06
tests: six root-level test suites are shipped but never executed (CI blind spot)


Scott-Emberson:
### Problem

Six standalone test suites at the repo root are registered in `update-system.mjs` SYSTEM_PATHS (so every install ships them) but are **not executed by anything** — not in `test-all.mjs`'s script-runner list, not in the `tests/` auto-discovery (which is deliberately scoped to `tests/`), not in any workflow:

- `test-trust-validator.mjs`
- `test-salary-filter.mjs`
- `detect-reposts.test.mjs`
- `followup-cadence.test.mjs`
- `process-quality.test.mjs`
- `reply-matcher.test.mjs` (landed with #1607, same gap)

Their siblings ARE wired (`tracker-columns-tests.mjs`, `agent-inbox-tests.mjs`, `followup-seed-tests.mjs`, `updater-migration-tests.mjs` all run via test-all.mjs's `{ name, expectExit: 0 }` runner list), so this looks like drift, not design. Consequence: regressions in trust-validator, salary filtering, repost detection, follow-up cadence, process-quality parsing, and reply matching pass CI silently. The suites may also have bitrotted while unwatched — nothing would have noticed.

### Proposed solution

Add the six to the same script-runner list in `test-all.mjs` the wired siblings use, after first running each one standalone and fixing any bitrot found. One small PR; incremental follow-up if any suite turns out to be genuinely deprecated (in which case it should be deleted from SYSTEM_PATHS rather than shipped dead).

Found while doing the #1440 split; follow-up to that work alongside #1499. PR incoming.

### Area

Other

Issue #1440, opened 2026-07-02
test-all.mjs: split into auto-discovered per-area test files to stop cross-PR rebase conflicts


Scott-Emberson:
### Code of Conduct

- [x] I agree to follow the [Code of Conduct](https://github.com/santifer/career-ops/blob/main/CODE_OF_CONDUCT.md)

### Existing issues

- [x] I searched existing issues and this hasn't been requested yet

### Problem

`test-all.mjs` is now ~8,200 lines and ~70 sections in a single append-only file. Nearly every feature PR adds a section at the tail (or edits near a related one), so unrelated PRs constantly conflict in the same hunks and need repeated rebasing against each other.

The merge pain is already visible in the file itself: section numbering has broken down — there are four different sections numbered "12", three "15"s, and two "27"s — because every contributor renumbers against a moving target.

Provider tests are the biggest churn surface: ~30 sections of ~100–150 lines each with an identical skeleton (import module, check `id`, normalize-mapping checks, null-handling, mock-ctx `fetch`). Every new provider PR touches the same tail of the same file. The same append-conflict problem was already solved for the plugin registry in #1402 (per-plugin files, "conflict-free registry PRs") — the test suite has the identical disease.

### Proposed solution

Raising this for discussion before any code — it touches every open PR that edits `test-all.mjs`, so sequencing matters as much as the design.

**Proposal: keep `node test-all.mjs --quick` as the entrypoint (unchanged in `.github/workflows/test.yml`), but make it a thin runner that auto-discovers test files:**

test-all.mjs              # runner: globs tests/, awaits each file, aggregates, prints summary
tests/
  helpers.mjs             # pass/fail/warn/run/readFile + shared counters (moved as-is)
  core/
    syntax.test.mjs
    data-contract.test.mjs
    tracker.test.mjs
    ...
  providers/
    themuse.test.mjs      # one file per provider
    ...

Key points:

  helpers.mjs             # pass/fail/warn/run/readFile + shared counters (moved as-is)
  core/
    syntax.test.mjs
    data-contract.test.mjs
    tracker.test.mjs
    ...
  providers/
    themuse.test.mjs      # one file per provider
    ...

Key points:

[...]

PR #2468, opened 2026-08-03
tests(web): move web's hand-enumerated suites to a glob-discovered layout

### LINKED ISSUE #2360: tests(web): move web/'s hand-enumerated root suites to a glob-discovered layout (follow-up to #1440)
### Code of Conduct

### Existing issues

### Problem

`web/`'s three suites sit in the web root, enumerated by hand:

```
web/test-act-envelope.mjs
web/test-clean-chips.mjs
web/test-stream-parse.mjs
```

```json
"test": "node --test test-clean-chips.mjs test-stream-parse.mjs test-act-envelope.mjs"
```

This repo has settled on two defensible homes for a JS test — `tests/**/*.test.mjs` (104 files, auto-discovered per #1440) and colocated `{module}.test.mjs` beside the module (10 files at root and in `lib/`). The web root is neither. It also carries the failure mode #1440 was filed to remove: a suite missing from that string never runs, and `npm test` still exits 0. There is no signal.

That has already nearly cost coverage. In #2182 the enumeration collided in `web/package.json` — that branch had relocated `test-clean-chips.mjs` while `main` concurrently appended `test-stream-parse.mjs` and `test-act-envelope.mjs`. Taking either side of the conflict wholesale would have silently dropped two suites with every CI check green; it needed a deliberate hybrid resolution (glob *plus* the two flat files by name) to avoid that.

Worth noting how the two sanctioned conventions differ in cost. Only `tests/**` is free — `test-all.mjs`'s own comment says discovery "is limited to `tests/` so root-level standalone `*.test.mjs` files are never picked up", so every colocated suite needs a line in the hardcoded array at `test-all.mjs:191-197`. That tax is being paid right now: `test/profile-photo.test.mjs` and `test/zh-minimal-template.test.mjs` appear in `update-system.mjs`'s `SYSTEM_PATHS` (so they ship to every user on update) but are executed by nothing — a recurrence of #1624. I'll file that separately; flagging it here only as evidence that hand-registration doesn't hold as a codebase grows.

### Disclosure

I'm the author of #2182, and that PR already introduces a `web/tests/lib/` layout with `{module}.test.mjs` naming plus a glob-based `test` script — it added two new web suites and had to put them somewhere, and it followed the root's #1440 convention. That was a drive-by decision inside a PR about PDF rendering, it isn't called out in that PR's description, and it has had no design discussion. This issue is me putting it up for one properly, rather than letting a convention for `web/` arrive as a side effect. If the layout is rejected here, #2182 should be changed to keep `web/`'s current flat naming instead — happy to do that.

### Proposed solution

`web/` runs its own `npm test` rather than `test-all.mjs`, so unlike the root, both candidate layouts can be glob-discovered there — the choice is placement, not discovery:

**A. mirror the source tree under `tests/`** (matches the repo's 104-file majority)

```
web/src/lib/clean-chips.mjs  ->  web/tests/lib/clean-chips.test.mjs
"test": "node --test \"tests/**/*.test.mjs\""
```

**B. colocate beside the module** (matches the repo's 10-file minority)

```
web/src/lib/clean-chips.mjs  ->  web/src/lib/clean-chips.test.mjs
"test": "node --test \"src/**/*.test.mjs\""
```

I'd argue for **A**, on one web-specific ground beyond majority: `web/` is a Next.js app, and `src/` is the app's own tree — scanned by `next build`'s file tracing and by `tsc --noEmit`. Keeping test files out of it avoids entangling test fixtures with build and route conventions, and keeps the deployed source tree free of them. Option B's advantage (the test sits visibly next to what it tests) matters less here, where `src/lib/` is small and flat enough that `tests/lib/` mirrors it 1:1.

Either way the mechanics are the same and reviewable as pure moves:

- `git mv` each suite to its new home under the `{module}.test.mjs` name.
- Fix the `// Run:  node --test test-*.mjs` header comment inside each moved file.
- Fix the line-3 back-reference comment in each corresponding `web/src/lib/{module}.mjs`.
- Drop the enumeration from `web/package.json`.

No behaviour change, no new dependency, no framework — same `node --test`, same CI entrypoint (`npm test` in `web-ci.yml`).

**One floor to pin down.** Command-line glob support in `node --test` has no "Added in" note in the Node docs (nearest documented datapoint: the `run()` API's `globPatterns`, added in v22.6.0). `web-ci.yml` runs Node 24 so it cannot surface in CI, but the root `package.json` declares `engines: { node: ">=18" }` and `web/` declares no `engines` at all — a contributor on Node 18 or 20 could end up with a `test` script that matches nothing and passes. Whichever option wins, `web/package.json` should state its own `engines` floor explicitly.

### Sequencing

If #2182 merges first, this reduces to moving the two remaining flat suites and collapsing that PR's hybrid `test` script back to a plain glob. If this merges first, #2182 needs another `web/package.json` conflict resolution. Either order is fine — just not concurrently. Happy to open the PR either way.

nikolaysm:
## What does this PR do?

Moves `web/`'s two remaining hand-enumerated test suites into the glob-discovered `web/tests/` layout, collapses `web/package.json`'s `test` script to a plain glob, and adds a root-suite guard so the convention can't drift back. Since #2182 already merged, this is the reduced branch of #2360's own Sequencing note.

## Related issue

Closes #2360

## Type of change

- [ ] Bug fix
- [ ] New feature

## Checklist

---

### 1. The `engines` floor is `>=22.0.0`, not `next`'s `>=20.9.0`

The issue's "One floor to pin down" paragraph predicts that an old Node "could end up with a `test` script that matches nothing and **passes**." Measured, that premise is inverted:

| Scenario | Result |
|---|---|
| Node 18.19.1 / 20.11.1, glob operand | `Could not find '<pattern>'` — 0 tests, **exit 1** (loud) |
| Node 22.15.1, glob operand | 60 tests, exit 0 |
| Any Node, glob that expands but matches nothing | 0 tests, **exit 0** (silent) |

So an unsupported-glob Node fails *loudly*; the genuinely silent case is a glob that matches nothing. Two consequences:

- The floor is set from **what the glob needs** (≥ 22), not from `next@16.2`'s `>=20.9.0`. Declaring 20.9 would bless a range where `web/`'s entire suite cannot run at all. Node 22 is also Active LTS, and CI already runs 24.
- The silent case is what the guard asserts, which is exactly #1440's stated rule: *"the runner should also fail hard if the glob matches zero files, so a path typo can't silently turn CI green."*

### 2. A guard test, in the root suite

Not on the issue's 4-bullet checklist — added because an unwritten, unenforced convention is what let `web/` drift into two conventions in the first place. `tests/web-test-layout.test.mjs` asserts *properties*, not a frozen script string, so a legitimate restructure isn't blocked:

- every web suite lives under `web/tests/` as `{module}.test.mjs` (a `.ts` suite is flagged too — `node --test` can't run one without a loader, so it would look like coverage and never execute)
- the `test` script names no suites explicitly, and does declare a pattern
- discovery is **non-empty**, and the declared pattern actually reaches every suite on disk
- the declared `engines` floor can actually run a glob

**It lives in `tests/`, not `web/tests/`, deliberately:** `test.yml` runs `test-all.mjs` on every PR with no paths filter and is a required check, while `web-ci.yml` is informative by design ("a red here never blocks a core merge"); and a guard inside `web/tests/` would be discovered by the very glob it validates, so if the glob broke the guard would stop running too. Same reach-into-`web/` pattern as `test-all.mjs`'s 55.3c/55.3d freezes (#2369).

## Deliberately out of scope

- **The `test/` (singular) orphans.** `test/profile-photo.test.mjs` and `test/zh-minimal-template.test.mjs` are in `SYSTEM_PATHS` but executed by nothing — a live recurrence of #1624, which wired six root suites in but not these two. Confirmed still broken; left alone because the issue author said they'd file it separately.
- **Repo-wide layout enforcement.** The guard is `web/`-only. A repo-wide version would need a grandfathered allowlist for ~19 root-level colocated suites plus the `test/` dir — conventions #2360 explicitly scoped out.
- **Partial de-duplication.** `walkFiles()` is extracted to `tests/helpers.mjs` and used by the guard, but `test-all.mjs`'s own private `discoverTests` is left as-is: it's load-bearing for 104 suites and rewiring it is outside this issue's blast radius. Flagging so it doesn't read as an oversight.

## Verification

- `cd web && npm test` → **60/60, identical to the pre-change baseline** (the load-bearing check: a dropped suite is invisible, which is the whole point of the issue)
- 11 negative scenarios each confirmed to fire, so the guard isn't decorative: misplaced suite, `.ts` suite, legacy `test-*.mjs`, enumeration reintroduced, glob pointed at the wrong tree, no glob at all, engines below floor, engines absent, plus partial-version forms (`>=22`, `22.x`) accepted and `web/` absent handled as not-applicable
- The guard loads on Node 18, so it respects the root `package.json`'s `>=18` floor and stays safe for end users (`tests/` ships via `SYSTEM_PATHS`; `web/` does not)
- `npx tsc --noEmit` and `npm run build` clean. After a rebuild, `web/.next/server/**/*.nft.json` no longer lists the moved files — they were previously pulled into Next's build file tracing, which is concrete support for Option A over Option B
- `node test-all.mjs --quick` → green, with the 6 new assertions from the guard

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for web test placement, naming, discovery, and execution.
  * Clarified testing conventions, Node.js requirements, and source/test separation.

* **Tests**
  * Added automated checks enforcing web test layout and discovery rules.
  * Added a reusable recursive file-discovery helper.

* **Chores**
  * Updated web tests to use automatic discovery.
  * Raised the web test requirement to Node.js 22 or newer.
  * Updated test references and related inline documentation.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -125,6 +125,14 @@ node test-all.mjs --only providers/themuse   # Run just one provider's test(s)
 `tests/providers/{name}.test.mjs` — it's auto-discovered (`tests/**/*.test.mjs`),
 no registration needed. Do not add a section to `test-all.mjs` for this.
 
+**Adding a test for the web app:** web suites live under `web/tests/`, mirroring
+the tested module's path below `web/src/` (`src/lib/clean-chips.mjs` →
+`tests/lib/clean-chips.test.mjs`), named `{module}.test.mjs`. `web/`'s own
+`npm test` glob-discovers them, so no registration is needed there either — but
+keep them out of `web/src/` (Next.js scans that tree) and write them as `.mjs`,
+since there is no TypeScript loader for `node --test`. `web/README.md` has the
+detail; `tests/web-test-layout.test.mjs` enforces it on every PR.
+
 **`--only` is a dev convenience, not a PR gate:** it runs *only* the discovered
 `tests/` files matching the given substring and skips every inline core
 section (syntax, scripts, dashboard, data contract, personal data, paths,
--- a/tests/README.md
+++ b/tests/README.md
@@ -24,6 +24,18 @@ Node.js (`tests/helpers.mjs`).
   scripts. Note: standalone `*.test.mjs` files in the repo root are run by
   `test-all.mjs`'s inline script list, not by this directory's discovery.
 
+**Web tests do not live here.** `web/` runs its own `npm test` over
+`web/tests/**/*.test.mjs` (see [../web/README.md](../web/README.md)); this
+directory is for the core. The two suites also differ in style on purpose: web
+suites use `node:test`, while suites here use the `pass`/`fail` helpers because
+this suite must run on a bare clone with no framework — "not even `node:test`"
+(#1440). Don't carry either style across the boundary.
+
+The one exception to the split is a guard *about* web's layout —
+`web-test-layout.test.mjs` lives here on purpose, because `web-ci.yml` is
+informative by design and never blocks a merge, while `test-all.mjs` runs on
+every PR as a required check.
+
 ## Running
 
 ```bash
--- a/tests/helpers.mjs
+++ b/tests/helpers.mjs
@@ -196,12 +196,17 @@ export function fileExists(path) { return existsSync(join(ROOT, path)); }
  * level, so the result is identical on every run and every OS — the same
  * property test-all.mjs's own `tests/` discovery relies on (#1440).
  *
+ * A missing `dir` yields `[]` rather than throwing, so the caller reports its
+ * own contract failure (e.g. "discovery is empty") instead of the run dying
+ * mid-traversal with an ENOENT that says nothing about what was expected.
+ *
  * @param {string} dir - Absolute directory to walk.
  * @param {RegExp} match - Tested against each entry's basename.
  * @param {Set<string>} [skipDirs] - Directory names never descended into.
  * @returns {string[]} Absolute paths, parents before children.
  */
 export function walkFiles(dir, match, skipDirs = new Set()) {
+  if (!existsSync(dir)) return [];
   const out = [];
   const entries = readdirSync(dir, { withFileTypes: true })
     .sort((a, b) => (a.name < b.name ? -1 : a.name > b.name ? 1 : 0));
--- a/tests/helpers.mjs
+++ b/tests/helpers.mjs
@@ -2,7 +2,7 @@
 // Moved verbatim from test-all.mjs (issue #1440); no framework by design:
 // the suite must run on a fresh clone with only Node.
 import { execFileSync } from 'child_process';
-import { existsSync } from 'fs';
+import { existsSync, readdirSync } from 'fs';
 import { join, dirname } from 'path';
 import { fileURLToPath } from 'url';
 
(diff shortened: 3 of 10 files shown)


santifer:
@nikolaysm the core-side files look right to me: the CONTRIBUTING section is additive and accurate, and tests/web-test-layout.test.mjs enforcing the layout on every PR is the right mechanism (a layout convention without a guard is a wish). The web-side review is career-ops-ui's lane, so it's over to them from here; the area:web label routes it.

nikolaysm:
@santifer sounds good — looks like the `area:web` label didn't actually get applied, though. Mind adding it? No triage permissions on my end.

santifer:
Merged, @nikolaysm. The move itself is small; the part that makes it worth having is the root-suite guard, because a convention without one is a suggestion — the two suites that were hand-enumerated got that way precisely because nothing stopped them.

A glob-discovered layout also fixes the failure mode nobody notices: a hand-written list silently stops covering whatever gets added next, so the suite keeps passing while its coverage quietly shrinks. Closing #2360 with the reduced branch now that #2182 has landed is the right sequencing call. Merged! 🚀

PR #2469, opened 2026-08-03
feat(scan): add seniority_required gate to title filter

Stebrianj:
## What
Adds an optional `seniority_required` keyword filter to `buildTitleFilter()` in `scan.mjs`. When declared in `portals.yml`, the title must also match at least one seniority keyword (Director, VP, Head, Chief, General Manager).

## Why
Without this gate, IC-level titles containing a positive keyword — e.g. "Operations Supervisor" matching "operations", "Customer Success Manager" matching "customer success" — reach the pipeline even when the user only wants Director+ leadership roles. This fills the pipeline with noise for senior-level job seekers.

## How
- New optional `seniority_required` list in `title_filter` in `portals.yml`
- Checked after positive/negative filters; absent or empty → no change in behavior (fully backwards-compatible)
- Negative keywords still win (a "Junior Director" is rejected even if "Director" matches seniority)

## Testing
- New test block 11c in `test-all.mjs` covers: gated matching, legacy permissive behavior, negative priority, malformed entries
- All assertions pass

## Config example
```yaml
title_filter:
  positive: [operations]
  seniority_required: [director, vp, head, chief, "general manager"]
```


Stebrianj:
Closing — approach reworked. Will resubmit cleanly from upstream HEAD.

PR #2470, opened 2026-08-03
fix(merge): pass CAREER_OPS_PDF_INDEX to spawned sync-pdf-flags

Stebrianj:
## What
Propagates the `CAREER_OPS_PDF_INDEX` environment variable to the `sync-pdf-flags.mjs` child process spawned by `merge-tracker.mjs`.

## Why
When the tracker path is redirected via `CAREER_OPS_TRACKER` (e.g. in tests or custom tracker setups), the spawned child process formerly fell back to reading the real `data/pdf-index.tsv` from the repo root instead of the (possibly non-existent) sandbox manifest. This caused:

1. **Test leakage**: `tracker-columns-tests.mjs` Test 17 failed whenever the developer's real `pdf-index.tsv` contained a report number matching the sandbox row — an environment-dependent, non-deterministic failure.
2. **Latent production inconsistency**: the parent and child disagreed on the manifest source when `CAREER_OPS_TRACKER` was overridden.

## How
Pass `env: { ...process.env, CAREER_OPS_PDF_INDEX: PDF_INDEX_FILE }` to the `execFileSync` call at `merge-tracker.mjs:819`. The child (`sync-pdf-flags.mjs`) already respects the override — it just was not receiving it.

## Verification
- `tracker-columns-tests.mjs`: 36 passed, 0 failed (was 35/1 before)
- Zero behavior change in normal runs — `PDF_INDEX_FILE` equals the child's default path when `CAREER_OPS_TRACKER` is unset


Stebrianj:
Closing — approach reworked. Will resubmit cleanly from upstream HEAD.

Issue #1429, opened 2026-07-02
Sync the tracker's PDF column from data/pdf-index.tsv (❌ stays stale when the PDF is generated later)


santifer:
Surfaced by a real apply-run field test.

## Problem

When the PDF is generated AFTER the evaluation, the tracker's PDF column stays ❌ forever — nothing flips it. (Until yesterday the manifest itself was never written — fixed in #1367 — so this is the natural follow-up now that `data/pdf-index.tsv` is reliable.)

## Proposal

Either `generate-pdf.mjs --report=NNN` flips the row's ❌→✅ directly, or a small `sync-pdf-flags.mjs` reads `pdf-index.tsv` and reconciles the tracker column. The second composes better (idempotent, batch-friendly, one write-path).

Shaamak:
I'd like to claim this one.

 Plan: build `sync-pdf-flags.mjs` that reads `data/pdf-index.tsv` and reconciles ❌→✅ flips in `data/applications.md` using `reportNum` as the join key, following the existing reconciliation pattern (like `merge-tracker.mjs`, `dedup-tracker.mjs`). Idempotent, reusable as a post-PDF-generation step. Will reuse existing row parsers from `merge-tracker.mjs` and add regression tests to `test-all.mjs`.

Let me know if the approach tracks.

santifer:
Labeling this one `good first issue` — it's well-scoped and self-contained.

**Time:** ~45 min · **Difficulty:** Beginner-Intermediate

### Pattern to mirror
`merge-tracker.mjs` already reads `data/pdf-index.tsv` for other checks — reuse that read path, then update the ❌/✅ PDF column cell during the normal merge pass.

### Acceptance
A row whose report gained a PDF after tracker insertion shows ✅ after the next `node merge-tracker.mjs` run, with a test in `test-all.mjs` mirroring the existing merge-tracker test block.

gaurayushi:
Hi @santifer, I’d like to work on this issue if it’s still available.

I noticed someone has already shared an approach, so I don’t want to duplicate work. If it’s okay, I can take this up and submit a focused PR for syncing the tracker’s PDF column from data/pdf-index.tsv.

I’ll first review:
- data/pdf-index.tsv
- data/applications.md
- generate-pdf.mjs
- merge-tracker.mjs

Please let me know if I can proceed.

santifer:
[...]

PR #2472, opened 2026-08-03
fix: resolve the PDF manifest from the tracker, not the install dir

### LINKED ISSUE #2471: sync-pdf-flags.mjs reads the PDF manifest from its own directory, ignoring CAREER_OPS_TRACKER
### Code of Conduct

### Existing issues

### What happened?

`node test-all.mjs` fails 2 assertions in the merge-tracker #912 section on any checkout that has generated a PDF — while staying green in CI. CONTRIBUTING asks contributors to run the full suite before pushing, so this greets anyone who actually uses career-ops for its own purpose.

`sync-pdf-flags.mjs` resolves the tracker through the shared resolver, but resolves the PDF manifest from its own script directory:

```js
const APPS_FILE = resolveTrackerPath(CAREER_OPS);          // honours CAREER_OPS_TRACKER
const PDF_MANIFEST = process.env.CAREER_OPS_PDF_INDEX
  || join(CAREER_OPS, 'data', 'pdf-index.tsv');            // does NOT — script dir
```

`merge-tracker.mjs:819` spawns it. The #912 test isolates the tracker and additions dir into a temp workspace but doesn't set `CAREER_OPS_PDF_INDEX`, so the child reads the checkout's **real** `data/pdf-index.tsv` and flips the fixture's PDF cells ❌→✅, breaking two exact-row assertions.

Notably `merge-tracker.mjs:81` already derives its own manifest correctly from the tracker path — the two scripts disagree about where the manifest lives.

### Steps to reproduce

Reproduces on a checkout whose `data/pdf-index.tsv` already lists report number `1` (i.e. `generate-pdf.mjs` has run at least once for report 1). **No need to edit that file — please don't, it's your own data.**

```bash
T=$(mktemp -d); mkdir -p "$T/data" "$T/reports" "$T/additions"
printf '# Applications Tracker\n\n| # | Date | Company | Role | Score | Status | PDF | Report | Notes |\n|---|------|---------|------|-------|--------|-----|--------|-------|\n| 1 | 2026-01-01 | OtherCo | Staff Engineer | 4.0/5 | Evaluated | \xe2\x9d\x8c | [1](../reports/001-otherco-2026-01-01.md) | original |\n' > "$T/data/applications.md"
echo x > "$T/reports/001-otherco-2026-01-01.md"; echo x > "$T/reports/001-newco-2026-01-05.md"
printf '1\t2026-01-05\tNewCo\tNew Role\tEvaluated\t2.7/5\t\xe2\x9d\x8c\t[1](reports/001-newco-2026-01-05.md)\tcollision\n' > "$T/additions/001-newco.tsv"

CAREER_OPS_TRACKER="$T/data/applications.md" CAREER_OPS_ADDITIONS="$T/additions" node merge-tracker.mjs
grep '^| ' "$T/data/applications.md"
```

Prints `📊 Summary: 2 PDF flags synced` and the rows come back `✅` — even though `$T` contains no `pdf-index.tsv` at all. Re-running with `CAREER_OPS_PDF_INDEX="$T/data/pdf-index.tsv"` added produces the expected `❌` rows, confirming the leak.

### Expected behavior

With the tracker redirected to an isolated workspace, the manifest for *that* workspace (here: absent) is used, and both rows keep `❌`.

### Scope

`test-all.mjs` sets `CAREER_OPS_TRACKER` 33 times and `CAREER_OPS_PDF_INDEX` **zero** times, across 11 `merge-tracker` spawns. #912 is the only one asserting exact PDF cells, so the others are latent — they surface for whoever's manifest happens to contain their fixture's report numbers. `CAREER_OPS_PDF_INDEX` is also undocumented in any `.md`.

This is not only a test-isolation problem: the same asymmetry means anyone pointing `CAREER_OPS_TRACKER` at a second workspace gets that workspace's tracker reconciled against the *original* checkout's manifest.

### Proposed fix

Derive the manifest from the resolved tracker path — exactly what `merge-tracker.mjs:81` already does — keeping `CAREER_OPS_PDF_INDEX` as an explicit override. Then isolating the tracker isolates the whole workspace and no test has to remember a second variable. Zero behaviour change for real users: a tracker at `<repo>/data/applications.md` yields today's manifest path.

### CLI tool

Claude Code

### OS and Node.js version

macOS, Node v22.15.1

nikolaysm:
## What does this PR do?

`sync-pdf-flags.mjs` and `find.mjs` resolved the PDF manifest from their own install directory while resolving the tracker from `CAREER_OPS_TRACKER`, so a redirected tracker was reconciled against the *wrong workspace's* manifest. All three tracker-coupled readers now derive it from the tracker via one shared helper.

## Related issue

Closes #2471

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

---

## The bug, measured

`merge-tracker.mjs:819` spawns `sync-pdf-flags.mjs`. The #912 test isolates the tracker and additions dir into a temp workspace, but cannot isolate a manifest it does not know exists, so the child read the real `data/pdf-index.tsv` and flipped the fixture's PDF cells ❌→✅:

```
before: 📊 Summary: 2 PDF flags synced   → both fixture rows ✅ (assertions fail)
after:  📊 Summary: 0 PDF flags synced   → both fixture rows ❌ (as asserted)
```

`node test-all.mjs --quick` went from **2 failed** to **0 failed** on a checkout with generated PDFs. CI never saw it — `data/` is gitignored, so CI has no manifest to leak.

## Two things the issue did not cover

- **`find.mjs` had the identical bug** (tracker from env at line 138, manifest from `ROOT` at line 146). Same fix applied — searching a redirected tracker no longer annotates it with this install's PDFs.
- **The path literal was rebuilt in every reader.** `join(..., 'data', 'pdf-index.tsv')` appeared in each, each picking its own base directory — which is *why* they could disagree. `resolvePdfIndexPath()` now owns the location and the `CAREER_OPS_PDF_INDEX` override; `merge-tracker.mjs`'s already-correct copy uses it too, so the two cannot drift apart again.

## Behaviour notes

- **No change for normal use:** a tracker at `<repo>/data/applications.md` resolves to today's manifest path. Verified with `sync-pdf-flags --dry-run` and `find.mjs` against a real workspace.
- **One intentional change:** `merge-tracker.mjs` now honours `CAREER_OPS_PDF_INDEX`, which it previously ignored. Having one reader honour an override and another not is the same class of inconsistency this fixes.

## Deliberately out of scope

- **`generate-pdf.mjs:378`** writes the manifest from `__dirname` and has no tracker awareness, so a redirected tracker still gets its PDFs recorded against the install directory. Real, but a writer-side design question rather than this asymmetry.
- **`outcome.mjs:201`** derives `repoRoot` as `dirname(trackerDir)` unconditionally, which is wrong for the root `applications.md` layout — a *different* defect with its own blast radius. Can file it separately.
- **Docs:** no `.md` documents any `CAREER_OPS_*` variable — not `CAREER_OPS_TRACKER` either — so documenting only this one would be inconsistent. The resolver's JSDoc covers the override; a full env reference belongs in its own docs PR.

## Verification

- New `tests/pdf-index-path.test.mjs` (5 assertions, auto-discovered): both tracker layouts, the override, and a named #2471 regression asserting a redirected tracker never resolves to the install directory's manifest. Confirmed it fails against the pre-fix behaviour.
- The existing #1429 test still passes, proving the manifest is still found when it legitimately exists in the workspace — the fix isolates without disabling.
- `node test-all.mjs --quick` → 2789 passed, 0 failed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected PDF manifest resolution when using redirected tracker locations.
  * Ensured workspace, reports, and PDF index paths consistently resolve from the configured tracker.
  * Preserved support for explicit PDF index path overrides.

* **Tests**
  * Added coverage for root-level and data-directory tracker layouts, custom overrides, and redirected tracker scenarios.

### DIFF
--- a/find.mjs
+++ b/find.mjs
@@ -28,6 +28,7 @@ import { readFileSync, existsSync } from 'fs';
 import { dirname, resolve } from 'path';
 import { fileURLToPath, pathToFileURL } from 'url';
 import { resolveColumns, parseTrackerRow } from './tracker-parse.mjs';
+import { resolvePdfIndexPath } from './tracker-utils.mjs';
 import { roleFuzzyMatch } from './role-matcher.mjs';
 
 const ROOT = dirname(fileURLToPath(import.meta.url));
@@ -143,7 +144,9 @@ function main() {
   }
   const rows = parseTrackerRows(readFileSync(trackerPath, 'utf-8'));
 
-  const manifestPath = resolve(ROOT, 'data', 'pdf-index.tsv');
+  // Derived from the tracker resolved just above, not from ROOT: a redirected
+  // CAREER_OPS_TRACKER must not be searched against this install's manifest (#2471).
+  const manifestPath = resolvePdfIndexPath(trackerPath);
   const pdfIndex = existsSync(manifestPath)
     ? parsePdfIndex(readFileSync(manifestPath, 'utf-8'))
     : new Map();
--- a/merge-tracker.mjs
+++ b/merge-tracker.mjs
@@ -22,7 +22,7 @@ import { normalizeReportLink as normalizeLink } from './tracker-links.mjs';
 import { roleFuzzyMatch } from './role-matcher.mjs';
 import { parsePdfIndex } from './find.mjs';
 import { LEGACY_COLMAP, detectColumns, resolveScoreStatus, normalizeVia, SEPARATOR_ROW_RE } from './tracker-parse.mjs';
-import { resolveTrackerPath, trackerLockDirFor, acquireTrackerLock, writeFileAtomic, normalizeCompany, cell } from './tracker-utils.mjs';
+import { resolveTrackerPath, resolveWorkspaceRoot, resolvePdfIndexPath, trackerLockDirFor, acquireTrackerLock, writeFileAtomic, normalizeCompany, cell } from './tracker-utils.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
 // Support both layouts: data/applications.md (boilerplate) and applications.md
@@ -78,8 +78,10 @@ const TRACKER_LOCK_DIR = trackerLockDirFor(APPS_FILE);
 
 // The reports/ dir sits at the repo root, which is the tracker's parent in the
 // data/ layout (data/applications.md) and the tracker's own dir at root layout.
-const REPORTS_ROOT = basename(TRACKER_DIR) === 'data' ? dirname(TRACKER_DIR) : TRACKER_DIR;
-const PDF_INDEX_FILE = join(REPORTS_ROOT, 'data', 'pdf-index.tsv');
+// Shared with sync-pdf-flags.mjs so the two agree on where a workspace's
+// siblings live — they disagreed before #2471.
+const REPORTS_ROOT = resolveWorkspaceRoot(APPS_FILE);
+const PDF_INDEX_FILE = resolvePdfIndexPath(APPS_FILE);
 
 /**
  * Normalize report links before writing them into the tracker file.
--- a/sync-pdf-flags.mjs
+++ b/sync-pdf-flags.mjs
@@ -17,11 +17,13 @@ import { readFileSync, existsSync } from 'fs';
 import { join, dirname } from 'path';
 import { fileURLToPath } from 'url';
 import { extractTrackerReportNumbers, resolveColumns, parseTrackerRow } from './tracker-parse.mjs';
-import { rebuildRow, resolveTrackerPath, openTrackerTransaction } from './tracker-utils.mjs';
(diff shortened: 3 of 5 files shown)


nikolaysm:
@santifer — fixes applied here, suite green locally (2789 passed, 0 failed). Ready for a look when you have a moment.

santifer:
Executed both directions, @nikolaysm: planted a manifest in the install dir, redirected the tracker, and watched pre-fix code flip fixture rows ❌→✅ from the wrong workspace's manifest — your fix takes it to zero while keeping default-layout resolution byte-identical. Centralizing the path in one shared helper is what actually kills the drift class, and declaring the out-of-scope writers (generate-pdf, outcome) instead of quietly leaving them was the honest move. Merged. 🚀

PR #2482, opened 2026-08-04
fix(tests): stop test-all silently passing when a node:test suite fails

### LINKED ISSUE #2463: test-all.mjs reports success when a node:test suite under tests/ fails
## What happens

`node test-all.mjs` cannot fail because of a `tests/**/*.test.mjs` suite that uses `node:test`. A failing suite reports success.

Reproduce on a clean checkout — drop this in as `tests/zz-probe.test.mjs`:

```js
import { test } from 'node:test';
import assert from 'node:assert/strict';
test('DELIBERATE FAILURE probe', () => { assert.equal(1, 2); });
```

then run `node test-all.mjs`:

```
📊 Results: 2049 passed, 0 failed, 0 warnings
🟢 All tests passed — safe to push/merge
exit 0
```

The same file under `node --test tests/zz-probe.test.mjs` fails correctly, which is why this has stayed invisible.

## Why

`runDiscovered()` imports each discovered suite in-process so it can share the `pass`/`fail` counters from `tests/helpers.mjs`. That is correct for helper-based suites. It does not work for `node:test` ones, for two independent reasons:

1. `node:test` reports through node's own runner. It increments none of the counters `finish()` reads, so its results are invisible to the summary.
2. `node:test` runs its tests **asynchronously, after the import resolves**. `finish()` therefore executes before they complete, and its `process.exit()` terminates them mid-flight.

Because of (2) this cannot be fixed inside `finish()`. `node:test` does set `process.exitCode = 1` on failure, but at the moment `finish()` runs, that has not happened yet — the result does not exist to be read.

## Scope

In my checkout, of 88 discovered suites, **16 import `node:test`** and 75 use the helpers. The 16 include `url-identity`, `digest`, `stats`, `pipeline-state`, `eval-verify`, `json-extract`, `location-filter`, `contract-drift`, `archive-baseline`, `needs-attention`, `needs-attention-sandboxing`, both `merge-tracker` suites and `company-blocklist`.

Those suites still *run* (their side effects are observable in the output), and they pass today. The problem is that if one of them ever regressed, `test-all.mjs` — the entrypoint the README, CI, and `AGENTS.md` all point at — would print `🟢 All tests passed` and exit 0.

## Suggested fix

Run `node:test` suites in a child process and fold the real exit code into the counters, deciding per file by checking whether the source imports `node:test`:

```js
if (/from ['"]node:test['"]/.test(src)) {
  const out = run(NODE, ['--test', f]);
  if (out === null) fail(`${rel} — node:test suite failed (exit ${lastRunFailure()?.status ?? '?'})`);
  else pass(`${rel} — node:test suite passed`);
  continue;
}
await import(pathToFileURL(f).href);
```

Echoing the runner's last lines on failure makes the report actionable rather than a bare "failed". A suite is then either counted in-process or counted as a child, never neither. Worth pairing with `finish()` honouring a pre-existing nonzero `process.exitCode` as defence in depth.

After this, the same probe prints `❌ tests/zz-probe.test.mjs — node:test suite failed (exit 1)` and exits 1, and a clean run names all 16 suites with their test counts. My local total moved 2049 → 2065, which is those suites being counted for the first time rather than any new test.

## Trade-off

16 extra child processes. On this machine the suite got roughly 8 seconds slower. An alternative is the programmatic `run()` API from `node:test`, which avoids the spawns but is a larger change; I went with the spawn because it is obvious and it matches how `test-all.mjs` already shells out to root-level `*.test.mjs` files.

Happy to send a PR if the approach looks right. Verified on Windows 11, Node v24.18.1; nothing in the change is platform-specific.

sdkkds:
Fixes #2463.

`test-all.mjs` — the entrypoint README, CI and AGENTS.md all point at — reported success while a deliberately failing suite sat in `tests/`:

```
$ node --test tests/zz-probe.test.mjs    # control
exit 1                                    # fails correctly
$ node test-all.mjs                       # same probe, canonical entrypoint
📊 Results: 2049 passed, 0 failed         # the deliberate failure is invisible
exit 0
```

## Why

`runDiscovered()` imports every discovered suite in-process so it can share the pass/fail counters. That works for the suites built on `tests/helpers.mjs`. The `node:test` suites report through node's own runner, which touches none of those counters — and they run their tests asynchronously *after* the import resolves, so `finish()`'s `process.exit()` terminated them mid-flight and threw the result away.

Every `node:test` suite in the tree was therefore decorative under the canonical entrypoint: url-identity, digest, stats, filter-precision, pipeline-state, eval-verify, the merge-tracker suites and the rest only ever reported when someone ran `node --test` by hand.

This cannot be fixed in `finish()` — by the time it runs, the result is already gone. It has to happen where the suite is invoked.

## What changed

- `node:test` suites now run in a child process (`node --test <file>`) and the real exit code folds into the counters, with the runner's own summary echoed on failure so the report is actionable. This mirrors how `test-all` already shells out to root-level `*.test.mjs`.
- Detection is a source check for an import of `node:test`, so a suite is either counted in-process or counted as a child — never neither.
- `finish()` additionally honours a pre-existing nonzero `process.exitCode` as defence in depth, so any future async leak fails loud instead of silently.

## Rebase note

Since the fix was written, main renamed `lastRunFailure` → `formatRunFailure` in this file's import. Both are exported from `tests/helpers.mjs`; the failure branch added here wants the structured `{status, stdout, stderr}` record rather than the formatted string, so the resolved import pulls in both. No behaviour change to the existing `formatRunFailure()` call site.

## Verification

Rebased onto `ead7985`. Windows, Node v24.19.0:

```
$ node test-all.mjs                                  # clean
📊 Results: 2847 passed, 0 failed, 1 warnings        exit 0

$ node test-all.mjs                                  # failing node:test probe planted in tests/
❌ tests\aaa-tmp-fail-probe.test.mjs — node:test suite failed (exit 1)
📊 Results: 2847 passed, 1 failed, 1 warnings        exit 1
```

The clean-run total moved because the previously-invisible `node:test` suites are now actually reported. Each is named in the output with its test count.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved test reporting to include failures from discovered test suites.
  * Test runs now correctly return a failure status when suite-level failures occur.
  * Successful suites now report their test counts.
  * Runner output is surfaced alongside existing assertion failure details.
  * Inline assertion failures and suite-level failures are now reported together for clearer, more reliable test results.


santifer:
This one closes a hole you both found and fixed within 48 hours, @sdkkds: sixteen node:test suites reported to nobody, and your child-process runner folds their real exit codes into the global verdict — with the runner's own tail echoed so a failure is actionable. I resolved the merge conflict with today's main on top of your branch (the finish() guard landed this afternoon) and re-ran the negative control on the result: a planted failing suite now takes the whole run down, exit 1. Merged. 🚀

sdkkds:
Thanks — and noted on the conflict resolution, I'd have had to redo that anyway after `cd346d3`.

On CodeRabbit's two: I read them as real-but-unreachable today, which I assume is why you merged over them. `import('node:test')` isn't used by any suite in the tree, and every `process.exitCode =` in the repo assigns a number, so `Boolean()` never sees a string `'0'`.

The dynamic-import one does leave the "counted in-process or counted as a child, never neither" claim slightly wider than the code. Happy to send a small follow-up that extends the detector, makes the `exitCode` check numeric (guarding `undefined`, so a clean run doesn't invert), and adds a regression suite that actually uses `await import('node:test')` — mostly for that last file, which makes the invariant testable rather than asserted. Equally happy to leave it if you'd rather not carry the extra surface.

Issue #1799, opened 2026-07-11
feat: recognize "—" as a valid score-cell sentinel + document the existing N/A sentinel for backfilled no-score entries


Schlaflied:
## Problem

When backfilling a rejection email for an application that never went through a full evaluation (no score, sometimes no exact application date either), `merge-tracker.mjs` refuses to merge the TSV:

```
⚠️  Skipping 1590-company-x-project-manager.tsv: cannot tell score from status in columns 5–6 ("Rejected" | "—") — refusing to merge a possible column swap
```

Root cause: `resolveScoreStatus()` (`tracker-parse.mjs:76`) calls `looksLikeScoreCell()` on both cells; if neither matches, it correctly refuses to guess (the #1427 column-swap guard). But `—` — the tracker's own established convention for "no data" everywhere else (Report column, PDF column, etc.) — isn't in the score-cell pattern, so any backfilled unscored entry trips this guard 100% of the time.

Not hypothetical: `data/applications.md` currently has 5+ manually-inserted rows that all hit this exact wall and had to bypass the TSV pipeline entirely (hand-edited straight into the tracker) — same failure mode, same manual workaround, every time.

**Found while investigating:** `looksLikeScoreCell()` (`tracker-parse.mjs:57-60`) *already* special-cases `'N/A'` and `'DUP'` as valid score-cell sentinels — this path exists and works today (added in #1517/#1427), it's just completely undocumented (not in `AGENTS.md`, not in `docs/SCRIPTS.md`, not anywhere in the TSV format section). No way to discover it without reading the source.

## Proposed fix (two small, independent pieces)

1. **Docs**: add a line to the TSV Format section in `AGENTS.md`/`docs/SCRIPTS.md` — "for a backfilled entry with no evaluation, use `N/A` as the score field (not `—`)."
2. **Parser**: add `—`/`-` as additional recognized score-cell sentinels alongside `N/A`/`DUP` in `looksLikeScoreCell()`, since `—` is already the tracker's de facto "no data" convention everywhere else — no reason the score column should be the one exception. Zero risk to the #1427 column-swap guard: a genuine status value (`Applied`, `Rejected`, etc.) never collides with `—`.

## Scope check

[...]

PR #2484, opened 2026-08-04
fix(merge-tracker): collapse placeholder Notes to the re-eval marker (#2483)

### LINKED ISSUE #2483: bug(merge-tracker): placeholder Notes (`—`, `-`, `N/A`) gain a stray separator on score upgrade
Follow-up to #2392 / PR #2399, covering the one placeholder case that fix missed.

`mergeNotes()` (merge-tracker.mjs:276-295) collapses to the marker alone only when the existing cell is empty, whitespace, or a bare `.` (`prev` falsy at :294). The tracker's placeholder sentinels are still truthy after the trim, so they survive the join:

```
existing Notes:  —
after 4→4.5:     —. Re-eval 2026-07-15 (4→4.5): rescored after weighting change
```

Same for `-` and `N/A`. The row gains a `—. ` separator it never had. This is the placeholder case raised explicitly in the #2392 thread: "The `—` and empty-string placeholders need collapsing to the marker alone so no row gains a separator it didn't have before."

Why `—` / `-` / `N/A` is the right set and not an arbitrary one: it is already the tracker's own "no data" convention. `looksLikeScoreCell` (tracker-parse.mjs:61-63, from #1799) recognizes exactly these as score sentinels, tracker-parse.mjs:309 treats `-` and `—` as empty when splitting, and merge-tracker itself writes `—` into empty Via and Location cells (:377, :857). A user hand-writing the same convention into an empty Notes cell is ordinary, not exotic. (`DUP` is score-column-specific and stays out.)

tests/merge-tracker.test.mjs has no placeholder-notes case, so nothing pins this today.

The fix is one line: treat the existing cell as empty inside `mergeNotes()` when it matches the sentinel set, plus a test. PR to follow.

Scott-Emberson:
Closes #2483. Follow-up to #2392 / #2399, covering the placeholder case that fix missed.

A Notes cell holding one of the tracker's "no data" sentinels (`—`, `-`, `N/A`) was still truthy after `mergeNotes()` trimmed it, so a score upgrade joined it like real text:

```
before:  —
after:   —. Re-eval 2026-03-01 (4→4.7): re-scored after JD refresh
```

The row gained a `—. ` separator it never had. This was raised explicitly in the #2392 thread and is the one ask from that comment the merged fix did not cover.

## Fix

One conditional in `mergeNotes()`: when the trimmed existing cell is exactly `—`, `-`, or `N/A`, treat it as empty so the cell collapses to the marker alone. The set mirrors `looksLikeScoreCell` (tracker-parse.mjs, #1799) minus the score-only `DUP`; it is the same convention merge-tracker itself writes into empty Via and Location cells, so a user hand-writing it into Notes is ordinary. Matching is exact and case-sensitive, same as the score-cell check, so a real note that merely starts with a dash is untouched.

## Tests

One new block in `tests/merge-tracker.test.mjs`: each of the three sentinels is seeded as the full Notes cell, upgraded 4.0→4.7, and the merged cell must equal the marker plus the new note exactly, with nothing leaked in front. All three fail on main (`"—. Re-eval …"`) and pass with the fix. Full suite: 2869 passed, 0 failed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Notes now correctly replace placeholder values such as “—”, “-”, and “N/A” when adding re-evaluation updates.
  * Existing meaningful notes continue to be preserved while being cleaned up consistently.

* **Tests**
  * Added regression coverage to verify placeholder notes are removed from updated entries.

### DIFF
--- a/merge-tracker.mjs
+++ b/merge-tracker.mjs
@@ -275,8 +275,14 @@ function companiesMatch(a, b) {
  */
 function mergeNotes(existingNotes, addition, oldScore, newScore) {
   // Trailing period trimmed only so the '. ' join does not produce '..'; no
-  // other character of the existing text is touched.
-  const prev = String(existingNotes ?? '').trim().replace(/\s*\.\s*$/, '');
+  // other character of the existing text is touched. The tracker's "no data"
+  // sentinels (the looksLikeScoreCell set minus the score-only DUP) count as
+  // empty here: a placeholder cell collapses to the marker instead of gaining
+  // a `—. ` separator the row never had (#2483).
+  const prevRaw = String(existingNotes ?? '').trim();
+  const prev = ['—', '-', 'N/A'].includes(prevRaw)
+    ? ''
+    : prevRaw.replace(/\s*\.\s*$/, '');
   const incoming = String(addition.notes ?? '').trim();
   const marker = `Re-eval ${addition.date} (${oldScore}→${newScore})`;
   // Re-running the same evaluation would otherwise repeat its own text; the
--- a/tests/merge-tracker.test.mjs
+++ b/tests/merge-tracker.test.mjs
@@ -282,6 +282,31 @@ try {
   fail(`merge-tracker Notes-preservation tests crashed: ${e.message}`);
 }
 
+// ── #2483: placeholder Notes collapse to the marker, not gain a separator ───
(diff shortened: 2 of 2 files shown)


santifer:
Clean one, @Scott-Emberson: treating whole-cell placeholder sentinels as empty in mergeNotes() with the leak pinned in tests closes #2483 exactly. Merged. 🚀

PR #2487, opened 2026-08-04
fix(providers): route softgarden and radancy through the shared entity decoder

tillmannschatz:
## Summary

Follow-up to the review on #1639, which asked for the two remaining private `decodeEntities` copies to be folded into the shared helper. `providers/_html-entities.mjs` says it exists so "the guard can't diverge again" — with `softgarden` and `radancy` still carrying their own copies, that wasn't true yet. After this PR no private `decodeEntities` remains under `providers/`.

The two migrations are not equivalent in weight:

**`softgarden` — correctness fix.** Its copy guarded the `fromCodePoint` range but not lone surrogate halves, and matched hex and decimal in a single `#x?[0-9a-fA-F]+` alternative. On real markup:

| Entity | Before | After |
|---|---|---|
| `&#xD800;` | a lone surrogate — ill-formed UTF-16 in the job title | `&#xD800;` (literal) |
| `&#1a2;` | parsed as decimal `1`, silently dropping `a2` | `&#1a2;` (literal) |
| `&#252;` / `&amp;` | `ü` / `&` | unchanged |

Neither case throws, so this is not the `RangeError` crash #1639 targeted — it is bad output rather than a dead scan. The ill-formed string is the more interesting one, since it travels into anything that later serializes or compares the title.

**`radancy` — pure de-duplication.** Its copy already had the correct range *and* surrogate guard, so behavior is unchanged; its existing malformed-entity regression test passes untouched.

## Test plan

- New regression test in `tests/providers/softgarden.test.mjs` puts out-of-range, lone-surrogate and mixed-radix entities in one title and asserts valid entities (`&#252;`, `&amp;`) still decode alongside them.
- Confirmed the new test **fails against the pre-migration decoder** (old output: `"Surrogate \ud800 Mixed "`), so it pins the fix instead of restating current behavior.
- `node test-all.mjs` — 2917 passed, 0 failed.
- Named-entity parity checked before swapping: both providers' local `NAMED_ENTITIES` maps were byte-identical to the shared one, so there is no repeat of the extra `&#8226;` mapping that `avature` carried in #1639.

## Notes

Branched fresh off current `main` rather than rebasing #1639's branch, which sat ~1000 commits back. This PR contains only the two migrations and the test — the shared module and the other four providers are already on `main`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTML entity handling when processing provider content.
  * Malformed numeric entities now remain as literal text instead of causing errors.
  * Valid numeric and named entities continue to decode correctly.

* **Tests**
  * Added coverage for malformed, valid, and mixed-format HTML entities.

### DIFF
--- a/providers/radancy.mjs
+++ b/providers/radancy.mjs
@@ -1,5 +1,6 @@
 // @ts-check
 /** @typedef {import('./_types.js').Provider} Provider */
+import { decodeEntities } from './_html-entities.mjs';
 
 // Radancy (TalentBrew) provider — the career sites Radancy hosts for large
 // employers (careers.munichre.com and its ERGO brands, plus many others). The
@@ -64,23 +65,6 @@ const PAGE_DELAY_MS = 150; // polite pacing — full walks are >100 sequential r
 // Page size for the JSON fragment transport. 100 is honored live by both known
 // legacy tenants (UHG, Kaiser); the HTML page hard-codes 15.
 const FRAGMENT_RECORDS_PER_PAGE = 100;
-
-// Minimal HTML entity decoder — mirrors the other HTML-scraping providers.
-const NAMED_ENTITIES = { amp: '&', lt: '<', gt: '>', quot: '"', apos: "'", nbsp: ' ' };
-/** @param {string} s */
-function decodeEntities(s) {
-  return s.replace(/&(#x?[0-9a-fA-F]+|[a-zA-Z]+);/g, (m, body) => {
-    if (body[0] === '#') {
-      const code = body[1] === 'x' || body[1] === 'X' ? parseInt(body.slice(2), 16) : parseInt(body.slice(1), 10);
-      // String.fromCodePoint throws RangeError outside 0..0x10FFFF or on a lone
-      // surrogate half — a malformed/adversarial entity must degrade to the
-      // original text, never crash the whole parse.
-      const valid = Number.isFinite(code) && code >= 0 && code <= 0x10ffff && !(code >= 0xd800 && code <= 0xdfff);
-      return valid ? String.fromCodePoint(code) : m;
-    }
-    return NAMED_ENTITIES[body.toLowerCase()] ?? m;
-  });
-}
 
 /** @param {string} s */
 function clean(s) {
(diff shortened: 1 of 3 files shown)


tillmannschatz:
Heads-up on the red `test (windows-latest)` check — I'm fairly confident it isn't this diff.

The failing script is `tracker-writer-lock-tests.mjs`:

```
FAIL recover guard never observed in any of the 8 guard-watched cases —
acquireTrackerLock has stopped emitting it, so every one of them fell back
to timing-dependent ordering
```

That looks like the windows flake from #2331 (closed via #2332, but evidently still surfacing). Two things point away from this PR being the cause: `test (ubuntu-latest)` and `test (macos-latest)` are green on the same commit, and the diff only touches `providers/softgarden.mjs`, `providers/radancy.mjs` and `tests/providers/softgarden.test.mjs` — nothing on the tracker-lock path, which doesn't import from `providers/` at all.

I can't re-run the job from a fork (`gh run rerun --job` answers "cannot be rerun"), so it needs a maintainer to kick it whenever you get to this one. Happy to push a rebase instead if you'd prefer a fresh run over a manual re-trigger.

santifer:
Thanks for finishing the #1639 follow-up, @tillmannschatz — folding the last two private `decodeEntities` copies (softgarden, radancy) into the shared `_html-entities.mjs` is exactly what that helper exists for, so the guard can't diverge again. Sorry for the slow first look.

One blocker before I can review properly: CI is red on this branch (test (windows-latest)). Could you take a look at what's failing and push a fix? Once it's green I'll go through the diff — the direction is right.

Scott-Emberson (approved):
`tests/providers/softgarden.test.mjs`: the fixture carries entities that must actually change, so the block fails if the shared decoder is bypassed, and it matches the radancy neighbour's integration shape. Radancy stays covered by its own test and the shared-module suite, so the migration is exercised on both sides. Approving from the tests/ side.

santifer:
Exactly the right direction, @tillmannschatz. The comment you deleted said it out loud —

```js
// Minimal HTML entity decoder — mirrors the other HTML-scraping providers.
```

"Mirrors" is a copy admitting it's a copy, and this one had real substance in it: the surrogate-half and out-of-range guards on `String.fromCodePoint`, so a malformed or adversarial entity degrades to the original text instead of crashing the parse. That's the kind of care that gets fixed in one copy and not the others, and then the difference is invisible until a specific board with a specific entity behaves differently from its neighbour.

Two decoders down. Merged 🚀

PR #2491, opened 2026-08-04
fix(tests): isolate tracker-columns fixtures from the real reports/ directory

anujad12:
## The bug

`makeSandbox` in `tracker-columns-tests.mjs` builds an isolated tracker and additions dir — but not an isolated `reports/`. So the `verify-pipeline` child it spawns scans the **real** `reports/` directory.

On a clean checkout that costs nothing, because `reports/` is empty. On a populated install it does: every report not referenced by the fixture's tiny tracker emits an "Orphan report" warning, so the child's output scales with the *user's history* instead of with the fixture. Test 2 then starts tripping its own 30s timeout under full-suite load.

It surfaces as:

```
❌ tracker-columns-tests.mjs crashed
```

with no assertion text — and intermittently, which makes it read like flakiness in the tracker writer lock rather than a fixture-isolation problem.

## Measurement

Against an install with **311 reports**, 8 consecutive runs each:

| | result |
|---|---|
| `upstream/main` | **3 of 8 runs failed** — 2, 4 and 1 assertions respectively |
| with this fix | **0 of 8 runs failed** |

The *varying* failure count is the tell: load- and timing-dependent, not a deterministic assertion problem. Wall time also drops (2.59s → 2.29s) since the child stops walking hundreds of unrelated files.

## The fix

Exactly what `test-all.mjs` already does for the same class of bug — put an empty `reports/` inside the sandbox and pin `CAREER_OPS_REPORTS` to it, so the spawned child sees the fixture and nothing else. 13 lines, no behaviour change to what is being asserted.

## Why this survived

**CI cannot see it.** CI checks out a repo with no reports, so the sandbox leak costs nothing there. It only appears once someone has a real history — which means it is invisible to the automation and visible only to the people actually using the tool. That asymmetry is why I thought it was worth writing up rather than just carrying locally.

Found while running the suite on a fork with ~300 evaluations. Happy to adjust the shape if you'd rather `makeSandbox` return the dir differently.

## Release Notes

* **Tests**
  * Improved test execution reliability by isolating reports directory to prevent verification tests from scanning real repository data, eliminating timeout issues during testing.

### DIFF
--- a/tracker-columns-tests.mjs
+++ b/tracker-columns-tests.mjs
@@ -48,6 +48,9 @@ function runScript(script, args, sandbox) {
     CAREER_OPS_TRACKER: sandbox.tracker,
     CAREER_OPS_ADDITIONS: sandbox.additions,
     CAREER_OPS_TRACKER_LOCK: sandbox.lock,
+    // Pinned for the same reason as the tracker: keep the fixture isolated from
+    // the real reports/ dir. See makeSandbox.
+    ...(sandbox.reports ? { CAREER_OPS_REPORTS: sandbox.reports } : {}),
   };
   try {
     const stdout = execFileSync(NODE, [join(ROOT, script), ...args], {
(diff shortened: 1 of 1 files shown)


santifer:
Welcome to career-ops, @anujad12! Pinning CAREER_OPS_REPORTS to a sandbox so the fixture can never read a real reports/ tree is exactly the isolation the data contract wants, and you found the precedent pattern on your own. Merged — your first contribution. Glad to have you here.

PR #2497, opened 2026-08-04
fix(cv): pair _EMPTY fallbacks with their block so mixed rows stay aligned

### LINKED ISSUE #2486: fix(cv): ORG_EMPTY/YEAR_EMPTY alignment fallbacks never emit — parsePartial suffix mismatch (certifications + awards)
**Time:** ~45 min · **Difficulty:** Beginner-plus · **Verified on main**

## What happens

The `certifications.html` partial (and now `awards.html`, which faithfully mirrors it — surfaced while verifying #2414) defines alignment fallbacks `<!--ORG_EMPTY-->` / `<!--YEAR_EMPTY-->` so rows without an org or year still emit an empty `<span>` and the table columns stay aligned.

They never fire. `parsePartial` (`build-cv-html.mjs:220`) computes `base = name.slice(0, -6)` — for `ORG_EMPTY` that yields base `ORG` (an orphan), not `ORG_BLOCK` where the renderer looks. So with mixed rows (some with org/year, some without), the columns misalign, and the comment above `buildCertifications`/`buildAwards` repeats a promise ("absent fields still emit an empty span") that the code doesn't keep.

## What to do

Either rename the fallbacks to `ORG_BLOCK_EMPTY` / `YEAR_BLOCK_EMPTY` in **both** partials, or teach `parsePartial` the current suffix — one approach, applied consistently, plus correcting the stale comment.

## Acceptance

- A CV with mixed certification/award rows (some missing org or year) renders aligned columns: the empty `<span>` actually appears.
- A regression test covering the mixed-rows case for both sections.
- The builder comments describe what the code now really does.

MannXo:
Fixes #2486.

`parsePartial` resolved an `_EMPTY` fallback by stripping the suffix, so `<!--ORG_EMPTY-->` was filed under `ORG` while `fillEntry` looks the fallback up under `ORG_BLOCK`. The definition was collected and then never used, so a certification or award row missing an org or a year emitted nothing where the empty `<span>` belonged.

Before, on a CV with mixed rows:

```
<span class="cert-title">Cert both</span> <span class="cert-org">Issuer</span> <span class="cert-year">2024</span>
<span class="cert-title">Cert no org</span> <span class="cert-year">2023</span>
<span class="cert-title">Cert no year</span> <span class="cert-org">Issuer</span>
<span class="cert-title">Cert neither</span>
```

After:

```
<span class="cert-title">Cert both</span> <span class="cert-org">Issuer</span> <span class="cert-year">2024</span>
<span class="cert-title">Cert no org</span> <span class="cert-org"></span> <span class="cert-year">2023</span>
<span class="cert-title">Cert no year</span> <span class="cert-org">Issuer</span> <span class="cert-year"></span>
<span class="cert-title">Cert neither</span> <span class="cert-org"></span> <span class="cert-year"></span>
```

### Which of the two fixes, and why

The issue offered either renaming the fallbacks in both partials or teaching `parsePartial` the current suffix. This takes the parser.

Both partial headers say "Template packs: edit tag names, class names, and field order freely inside the ENTRY zone". Any pack that copied the shipped `ORG_EMPTY` naming would keep parsing fine after a rename and silently stop emitting its fallback, which is the same invisible misalignment this issue is about, just moved onto pack authors. Fixing the parser repairs every existing pack, including ones outside this repo.

The pairing is resolved in a second pass over the collected definitions, because a fallback can be defined before the block it belongs to. `ORG_EMPTY` and `ORG_BLOCK_EMPTY` both attach to `ORG_BLOCK`, and the bare stem stays as a fallback so a `FOO_EMPTY` / `FOO` pair keeps working. Every conditional in every shipped partial is named `<FIELD>_BLOCK`, so there was no ambiguity to resolve.

### Testing

`tests/cv-partial-empty-fallbacks.test.mjs` (auto-discovered, `pass`/`fail` from `helpers.mjs`, no `process.exit`). It runs the real builder against the shipped `templates/sections/` partials with a mixed-rows payload, then asserts the exact markup of each row plus the alignment property itself: every row in a section carries the same number of cells. The shipped partials are the primary subject rather than a fixture copy; a fixture pack appears only to cover the second fallback spelling, described below.

End to end rather than a unit test because `build-cv-html.mjs` exports nothing, and the bug lived in the seam between the partial's block names and the builder's lookup. A unit test of either half alone would have passed.

```
$ node test-all.mjs --quick     # this branch
📊 Results: 2936 passed, 0 failed, 1 warnings

$ node test-all.mjs --quick     # main, for comparison
📊 Results: 2916 passed, 0 failed, 1 warnings
```

Exactly the 20 new assertions. The single warning is the pre-existing `cv-sync-check.mjs exited with error (expected without user data)` on both.

`test:cv-visual` needs `pdftotext`, which I do not have locally, so instead I rendered the same payload through the builder at the parent commit and at this one and diffed the HTML: byte-identical, since the only change to `build-cv-html.mjs` in the second commit is comment text.

Each assertion was checked by reverting the behaviour it covers and confirming the intended test goes red:

| Sabotage | Test result |
|---|---|
| restore the pre-fix stem lookup (the reported bug) | 7 failures across both sections |
| always use the present variant, ignoring `present` | 1 failure, on the skills row |
| always pair to `<stem>_BLOCK`, dropping the bare-stem fallback | 7 failures, all in the `_BLOCK_EMPTY` pack, shipped partials still green |

One correction while I am here: I tried a further sabotage, dropping the merge with the existing entry in the present-variant pass, and it is **not** caught. That one is behaviour-equivalent rather than a gap: the second pass writes `absent` afterwards, so overwriting only differs if a single block name is defined twice as a non-`_EMPTY` definition, which no partial does.

The second sabotage is why the test also covers skills. For certifications and awards, the present variant filled with an empty value renders byte-identically to the `_EMPTY` fallback, so no assertion on those two sections can tell the two implementations apart. `skills.html` defines `CATEGORY_BLOCK` with no `_EMPTY` sibling, so an absent category must collapse to nothing, and that sabotage renders a visible stray `<span class="skill-category">: </span>`. It also guards the shared `parsePartial` path this change reaches for every other section.

### Second commit, from review

Three things, none of them changing the fix.

**Both fallback spellings are now covered.** The parser accepts `ORG_EMPTY` and `ORG_BLOCK_EMPTY`, but only the shipped spelling had a test. The suite now builds the same payload a second time through a template pack that took the rename route the issue offered as the alternative, copying `templates/` and rewriting only those two partials so the spelling is the single variable. That pack is what the third sabotage row above catches, and it is the case the "fix the parser, not the partials" argument rests on, so it was worth pinning.

**The builder comments now name the mechanism** rather than restating the promise, which covers the third acceptance criterion directly instead of leaving it implied by the fix.

**A zero exit that wrote no output now fails with a message** instead of a bare `ENOENT` from `readFileSync`. Worth saying plainly: I could not construct a case that reaches this branch, because `writeAndReport` stats the file it just wrote, so a missing output already exits non-zero and lands in the crash branch. It is defensive only, and no test distinguishes it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved rendering of conditional CV sections when organization or year information is missing.
  * Added support for standard and block-specific empty fallbacks.
  * Ensured sections without configured fallbacks remain hidden when data is unavailable.

* **Tests**
  * Added end-to-end coverage for certification and award fallback rendering.
  * Verified generated markup and consistent output formatting.

### DIFF
--- a/build-cv-html.mjs
+++ b/build-cv-html.mjs
@@ -209,21 +209,35 @@ function parsePartial(source) {
   // can remove them verbatim from the entry zone in step 3, without needing any
   // broad HTML-comment regex (which would trigger CodeQL).
   const definitionStrings = [];
+  // Collect every definition first. The _EMPTY fallbacks are resolved in a
+  // second pass because a fallback may be defined before the block it belongs
+  // to, and pairing needs to know which block names exist.
+  const definitions = new Map();
   let m;
   while ((m = blockRe.exec(entryZone)) !== null) {
     const name = m[1];
     const content = m[2];
     if (name === 'ENTRY') continue; // skip the sentinel itself
     definitionStrings.push(m[0]); // full match, e.g. <!--FOO-->bar<!--/FOO-->
-    // EMPTY variants are the absent-field fallback (e.g. <!--ORG_EMPTY-->).
-    if (name.endsWith('_EMPTY')) {
-      const base = name.slice(0, -6);
-      const existing = blocks.get(base) || { present: '', absent: '' };
-      blocks.set(base, { ...existing, absent: content });
-    } else {
-      const existing = blocks.get(name) || { present: '', absent: '' };
-      blocks.set(name, { ...existing, present: content });
-    }
+    definitions.set(name, content);
+  }
+
+  const EMPTY_SUFFIX = '_EMPTY';
+  for (const [name, content] of definitions) {
+    if (name.endsWith(EMPTY_SUFFIX)) continue;
+    const existing = blocks.get(name) || { present: '', absent: '' };
+    blocks.set(name, { ...existing, present: content });
+  }
+  // An EMPTY variant is the absent-field fallback for a sibling block, and the
+  // renderer looks it up under the block's own name. Both spellings pair to the
+  // same block: <!--ORG_EMPTY--> and <!--ORG_BLOCK_EMPTY--> attach to ORG_BLOCK.
+  // Falling back to the bare stem keeps a FOO_EMPTY/FOO pair working.
+  for (const [name, content] of definitions) {
+    if (!name.endsWith(EMPTY_SUFFIX)) continue;
+    const stem = name.slice(0, -EMPTY_SUFFIX.length);
+    const base = definitions.has(`${stem}_BLOCK`) ? `${stem}_BLOCK` : stem;
+    const existing = blocks.get(base) || { present: '', absent: '' };
+    blocks.set(base, { ...existing, absent: content });
   }
 
   // Step 3: build the entry template by removing the captured block *definitions*
--- a/build-cv-html.mjs
+++ b/build-cv-html.mjs
@@ -459,8 +459,8 @@ function buildCertifications(entries, partial) {
   const { entryTemplate, blocks } = partial;
   return entries.filter(Boolean).map(e => {
     const blockValues = new Map([
-      // Certifications use EMPTY variants so that absent fields still emit an
-      // empty <span> for table-cell alignment — not a bare removal.
+      // An absent field resolves to the partial's _EMPTY fallback, emitting an
+      // empty <span> for table-cell alignment rather than being removed.
       ['ORG_BLOCK',  { value: escapeHtml(e.org || ''),  present: Boolean(e.org) }],
       ['YEAR_BLOCK', { value: escapeHtml(e.year || ''), present: Boolean(e.year) }],
     ]);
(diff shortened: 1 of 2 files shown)


santifer:
This is the route I'd have picked, @MannXo. The issue offered two ways out and you took the one that fixes the parser rather than the two files we happen to ship: anyone with a template pack that copied the `ORG_EMPTY` naming gets repaired by this too, without knowing they were broken. Accepting both spellings instead of renaming is what makes that possible.

Sorry it sat since Monday. Merged. 🚀

Issue #2477, opened 2026-08-04
feat(cli): followup-cadence.mjs has no --help — add the usage block


santifer:
**Time:** ~25 min · **Difficulty:** Beginner · **Reserved for @rutehcody** (comment below to lock the assignment; if no comment in 72h it's open to anyone)

## Context

`node followup-cadence.mjs` computes the follow-up cadence per active application and flags overdue entries; it also takes a `--summary` flag (see `docs/SCRIPTS.md`). Today it has no usage text at all: `node followup-cadence.mjs --help` silently ignores the flag and runs the full computation (`grep -c "Usage" followup-cadence.mjs` → 0).

## What to do

Mirror the pattern in `company-history.mjs` (lines 83–98): a `KNOWN_FLAGS` list, a usage block naming each flag (including `--summary`), and an early `--help` / `-h` exit **before any file is read**.

## Acceptance

- `node followup-cadence.mjs --help` (and `-h`) prints the usage block and exits 0 without reading any data file.
- The usage block documents every flag the script actually parses today (check the source, not the docs).
- No behavior change for existing invocations.

Divesh-Kshirsagar:
I'd like to take this issue if it's still available. The 72-hour reservation period appears to have passed.

aryansk:
The 72-hour reservation appears to have elapsed, so I’m taking this issue. I’ll keep the change limited to the existing `followup-cadence.mjs` flags: add an early `--help`/`-h` usage path, document the current options, and add a regression check without changing normal cadence behavior.

santifer:
@Divesh-Kshirsagar you asked for this one first and then it went to someone else within the hour, and nobody said a word to you for four days. That part is on me, not on you or on @aryansk.

Where it stands: this issue is @aryansk's, and their PR is open as #2636, so I am leaving it there.

What I owe you is a real one instead, and there are three siblings of this exact issue open and genuinely unclaimed right now (I checked assignees and comments on all three before writing this):

- **#2771** `plugin-audit.mjs` has no `--help` and treats an unrecognized flag as a directory
- **#2772** `jd-similarity.mjs` has no `--help` and reads an unrecognized flag as a file
- **#2774** `application-artifacts.mjs` has no `--help` and ignores unrecognized flags

Same shape as the one you wanted, same size, and each one comes with the pattern to mirror written into it. Comment `/assign` on whichever you prefer and the bot will assign it to you. One at a time for a first contribution, which is why I am pointing at three rather than handing you all of them.

[...]

Issue #2492, opened 2026-08-04
batch-runner hands workers an empty JD file — 35 of 251 offers failed on one run


anujad12:
## The bug

`batch/batch-runner.sh` creates the JD file the worker is told to read, and never writes to it:

```bash
jd_file="$(mktemp "${TMPDIR:-/tmp}/batch-jd-${id}.XXXXXX")"
```

`mktemp` produces an **empty** file. The prompt then tells the worker the JD is at `{{JD_FILE}}`, the worker finds 0 bytes, falls through to WebFetch, and on any JS-rendered board (Phenom, Workday, iCIMS) gets a careers-landing shell with no posting in it.

The worker then does exactly the right thing: it refuses to invent an evaluation, writes `{"status":"failed","error":"..."}`, and exits. So the failure is *correct behaviour* on top of a silent input bug — which is why it reads as "the model couldn't fetch the page" rather than "we handed it an empty file."

**Measured on one run: 35 of 251 offers failed this way.** Every one on a JS-rendered board.

## Why it isn't just "WebFetch is unreliable"

The data is already available for free at scan time for some providers, and cheaply at queue time for the rest. The runner is discarding it and asking the model to re-acquire it from a page that cannot serve it.

## What I built, and what I'd want your read on

On my fork this is two modules plus a runner change:

- **`jd-cache.mjs`** — a JD sidecar keyed on a canonical URL, with a *sufficiency gate*: an entry is only written if it looks like a real posting (length + signal patterns like responsibilities/qualifications). Refusals are logged to `_rejected.tsv` rather than silently dropped, so "we cached nothing" is never indistinguishable from "the board was empty." CLI: `text|path|have|stats|prune`, exit 3 = not cached.
- **`jd-fetch.mjs`** — per-ATS fetchers (Phenom, Workday, Lever, Ashby, Greenhouse, iCIMS) that fill gaps at **queue** time, not scan time. That distinction matters: doing it in `scan.mjs` would mean ~400 extra requests for a single ICF scan, nearly all for postings the title filter is about to reject. At queue time it is tens, only for postings that survived filtering.
- **`batch-runner.sh`** — fill `$jd_file` from the cache before launching the worker; exit 3 (not cached) leaves today's behaviour exactly as-is, so an absent entry still produces the same honest failure.

Result on the same pipeline: **35 failures → 3**, and those 3 were genuinely dead postings (2 unpublished Phenom requisitions, 1 Greenhouse 404).

[...]

PR #2499, opened 2026-08-04
fix(scan-ats-full): `--since -5` silently scanned nothing and called it a clean run

### LINKED ISSUE #2498: scan-ats-full: `--since` silently coerces every malformed operand — the same flag means two different things across the two scanners
### What

`scan.mjs` gained a strictly-validated `--since` in #2477. Its comment states the intent:

> Matches `scan-ats-full.mjs`, which has always treated `--since` as a filter; **one flag name should not mean two different things.**

But `scan-ats-full.mjs` parses it as:

```js
const sinceDays = Number(valueOf('--since')) || 3;
```

There is no later validation. Every malformed operand is silently coerced:

| input | `sinceDays` | effect |
|---|---|---|
| `--since abc` | **3** | typo swallowed — user believes they scanned the window they typed |
| `--since 0` | **3** | same |
| `--since -5` | **-5** | cutoff lands in the **FUTURE** → nothing is ever eligible |
| `--since 1e400` | **Infinity** | cutoff `-Infinity` → **no window at all** |

`--since -5` is the worst: a future cutoff means every posting is older than it, so the scan finds nothing and reports it as a clean run. That's indistinguishable from "no new postings" — the silent-wrong-answer shape, on the flag that decides what the sweep even looks at.

`--since 1e400` is the mirror image: the user asks for a bound and silently gets none.

`scan.mjs` rejects all four loudly, and its own comment names this exact trap:

> `Number.isFinite` also rejects Infinity and 1e309, which pass a bare `> 0` test and would yield an -Infinity cutoff.

So the two scanners disagree today, on the flag the newer one just promised parity with. The bad `cutoffMs` also gets persisted into the `--resume` checkpoint (`scan-ats-full.mjs:592`).

### Fix

Extract `scan.mjs`'s validation as an exported `parseSinceDays(args)` and have **both** scanners use it — `scan-ats-full.mjs` already imports from `scan.mjs`, so this shares rather than copies. Only the *default* stays per-script (`scan.mjs`: no bound; `scan-ats-full.mjs`: 3 days); the validation is shared, so the flag cannot mean two things again.

PR incoming, with a source-level assertion that neither scanner re-introduces a private coercion.

abankar1:
Closes #2498.

#2477 landed a strictly-validated `--since` in `scan.mjs`, with the stated intent:

> Matches `scan-ats-full.mjs`, which has always treated `--since` as a filter; **one flag name should not mean two different things.**

`scan-ats-full.mjs` doesn't match it. It parses `Number(valueOf('--since')) || 3`, with no later validation:

| input | `sinceDays` | effect |
|---|---|---|
| `--since abc` | **3** | typo swallowed — user believes they scanned what they typed |
| `--since 0` | **3** | same |
| `--since -5` | **-5** | cutoff in the **FUTURE** → nothing ever eligible |
| `--since 1e400` | **Infinity** | cutoff `-Infinity` → **no window at all** |

`--since -5` is the sharp one: a future cutoff makes every posting older than it, so the sweep finds nothing and reports a clean run — indistinguishable from "no new postings", on the flag that decides what the sweep looks at. `1e400` is the mirror image: ask for a bound, silently get none. The bad `cutoffMs` is persisted into the `--resume` checkpoint too.

`scan.mjs` already rejects all four, and its own comment names the trap:

> `Number.isFinite` also rejects Infinity and 1e309, which pass a bare `> 0` test and would yield an -Infinity cutoff.

## Fix

Extracted that validation as an exported `parseSinceDays(args)` and pointed **both** scanners at it. `scan-ats-full.mjs` already imports from `scan.mjs`, so this **shares** rather than copies — the flag can't drift into meaning two things again.

Only the *default* stays per-script, since they legitimately differ: absent `--since` is 3 days in `scan-ats-full.mjs`, no bound in `scan.mjs`. The helper returns `days: null, error: null` when absent and lets each caller choose.

After:

```
--since abc    Error: --since expects a positive number of days, got "abc"
--since -5     Error: --since expects a positive number of days, got "-5"
--since 0      Error: --since expects a positive number of days, got "0"
--since 1e400  Error: --since expects a positive number of days, got "1e400"
--since 7      accepted, proceeds normally
```

`scan.mjs`'s own behaviour is unchanged — verified across all 11 cases (both spellings, duplicates, missing operand, operand-is-next-flag, out-of-Date-range).

## Note

Sharing also brings `scan-ats-full.mjs` two guards it lacked: duplicate-occurrence rejection (`--since=7 --since`) and the out-of-Date-range check (`1e300`). Both were already `scan.mjs` behaviour; I've called them out rather than let them ride in unmentioned.

## Tests

Eight malformed operands must all be rejected, both spellings accepted, the absent case left to the caller — plus a source-level assertion that `scan-ats-full.mjs` doesn't re-introduce a private `Number()` coercion.

`node test-all.mjs --quick` → **2917 passed, 0 failed**.

3 commits (extract → fix → tests).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for the `--since` option.
  * Invalid, missing, duplicate, non-positive, non-finite, and out-of-range values now produce clear errors.
  * Preserved the default date range when `--since` is omitted.
  * Standardized `--since` handling across scanning commands.

### DIFF
--- a/scan-ats-full.mjs
+++ b/scan-ats-full.mjs
@@ -256,7 +256,7 @@ function parseArgs(argv) {
     return kv ? kv.split('=').slice(1).join('=') : null;
   };
   // Validated by the SAME parser scan.mjs uses, so one flag name cannot mean
-  // two different things (#2492). `Number(...) || 3` silently swallowed every
+  // two different things (#2498). `Number(...) || 3` silently swallowed every
   // malformed operand: `--since abc` and `--since 0` became 3 while the user
   // believed they had scanned the window they typed; `--since -5` put the
   // cutoff in the FUTURE so nothing was ever eligible, which reads exactly like
--- a/scan-ats-full.mjs
+++ b/scan-ats-full.mjs
@@ -43,7 +43,7 @@ import lever from './providers/lever.mjs';
 import ashby from './providers/ashby.mjs';
 import workday from './providers/workday.mjs';
 import icims from './providers/icims.mjs';
-import { buildTitleFilter, buildLocationFilter, buildContentFilter, matchedTitleKeywords, loadSeenUrls, normalizeUrlForDedup, appendToPipeline, appendToScanHistory, loadBlacklist } from './scan.mjs';
+import { buildTitleFilter, buildLocationFilter, buildContentFilter, matchedTitleKeywords, loadSeenUrls, normalizeUrlForDedup, appendToPipeline, appendToScanHistory, loadBlacklist, parseSinceDays } from './scan.mjs';
 import { SEED_SOURCES, toPortalEntry } from './seeds/vc-portfolios.mjs';
 import { normalizeCompany } from './tracker-utils.mjs';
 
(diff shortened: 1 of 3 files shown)


santifer:
Solid parity work, @abankar1: one shared exported parser, per-caller defaults preserved, and a source-level assert that blocks anyone from re-introducing a private Number() coercion. I pushed a one-token fix on top (the inline citations pointed at #2492 instead of #2498). Merged. 🚀

PR #2502, opened 2026-08-04
fix(fingerprint): a Russian and a Chinese employer counted as the same company, hiding cross-listings

### LINKED ISSUE #2501: fingerprint-core: cross-listings between two different non-Latin employers are silently never reported
### What

`companyKey()` in `fingerprint-core.mjs` gates the "different employer" check:

```js
function companyKey(name) {
  return String(name ?? '').toLowerCase().replace(/[^a-z0-9]/g, '');
}
```

`findCrossListings()` uses it to skip same-employer pairs:

```js
if (cand.key === offerCompany) continue; // re-post, not cross-listing
```

The `[a-z0-9]` strip deletes every non-Latin name, so アクメ株式会社, グロベックス合同会社, Яндекс and 北京字节跳动 all key to `''` and compare **equal**. Two genuinely different non-Latin employers therefore look like the same employer, the pair is skipped as a re-post, and the cross-listing is never reported.

### Reproduced

One identical JD, posted by two different companies, through `findCrossListings()`:

```
two Latin companies                 => cross-listings detected: 1
two DIFFERENT non-Latin companies   => cross-listings detected: 0   ← silently skipped
Russian vs Chinese company          => cross-listings detected: 0
```

A Russian company and a Chinese company are treated as the same employer.

This is the same root cause as #2429 with the opposite symptom: there an empty key caused a false *merge*; here it causes a false *skip*. Both make the tool quietly do less for anyone tracking non-Latin-market employers — and `modes/` ships ja, ko, zh, zh-TW, ru, ua, ar and hi.

### Fix

Delegate to the shared `normalizeTextKey()` (#2393/#2445), which is now the one key every grouping consumer is meant to share. It also keeps combining marks, so Devanagari/Arabic names differing only in matras stay distinct rather than collapsing into one "employer".

Verified after the change: all three pairs above detect 1, and same-employer pairs (Latin **and** non-Latin) are still skipped as re-posts.

PR incoming.

abankar1:
Closes #2501.

`companyKey()` in `fingerprint-core.mjs` stripped `[^a-z0-9]`, so every non-Latin company name keyed to `''` and compared equal. `findCrossListings()` skips same-key pairs:

```js
if (cand.key === offerCompany) continue; // re-post, not cross-listing
```

…so two genuinely different non-Latin employers looked like one, and a shared posting between them was **silently never reported**.

Reproduced with one identical JD posted by two companies:

```
                                      BEFORE   AFTER
two Latin companies                      1       1
two DIFFERENT non-Latin companies        0       1     ← was silently skipped
Russian vs Chinese company               0       1
Devanagari differing only in matras      —       1
```

Same root cause as #2429, opposite symptom: there an empty key caused a false *merge*, here a false *skip*. Both quietly do less for anyone tracking employers in the ja/ko/zh/zh-TW/ru/ua/ar/hi markets `modes/` ships.

## Fix

Delegate to the shared `normalizeTextKey()` (#2393/#2445) — the key every grouping consumer is meant to share — instead of a private strip. It keeps combining marks too, so Devanagari/Arabic names differing only in matras stay distinct rather than collapsing into one "employer".

## Tests

Three different-employer pairs must each detect the shared posting, **and** same-employer pairs must still be skipped as re-posts — Latin and non-Latin alike, so the fix can't be read as just disabling the skip.

One note: `tests/fingerprint-core.test.mjs` inlines a verbatim reference copy of the pre-#2381 implementation, including its own old `companyKey`. I deliberately left it untouched — the file says its whole value is being the code that shipped. The differential corpus is Latin-only, so it still agrees; if non-Latin fixtures are ever added there, the reference is the side that should be understood as stale.

`node tests/fingerprint-core.test.mjs` → all green (incl. the 153-match differential corpus). `node test-all.mjs --quick` → **2916 passed, 0 failed**.

2 commits (fix + test).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of non-Latin employer names to prevent distinct names from being incorrectly merged into the same identifier.
  * Fixed cross-listing detection for employer names with combining marks.

* **Tests**
  * Added regression tests for non-Latin employer name handling in cross-listing detection.

### DIFF
--- a/fingerprint-core.mjs
+++ b/fingerprint-core.mjs
@@ -19,6 +19,7 @@
  */
 
 import { createHash } from 'crypto';
+import { normalizeTextKey } from './tracker-parse.mjs';
 
 /** Descriptions shorter than this (after normalization) carry too little
  * signal to distinguish real matches from boilerplate — skip them. */
@@ -166,10 +167,22 @@ function maxDistanceFor(threshold) {
   return 64;
 }
 
-/** Company key for "different employer" checks — same normalization family as
- * the tracker tooling (lowercase alphanumerics). */
+/**
+ * Company key for "different employer" checks.
+ *
+ * Delegates to the shared normalizeTextKey() (#2393/#2445) rather than the
+ * `[a-z0-9]` strip it used to carry. That strip DELETED every non-Latin name,
+ * so アクメ株式会社, グロベックス合同会社 and Яндекс all keyed to '' and compared
+ * equal — and because findCrossListings() SKIPS same-key pairs as re-posts,
+ * an identical posting shared between two genuinely different non-Latin
+ * employers was silently never reported as a cross-listing. The Latin
+ * equivalent was reported normally (#2500).
+ *
+ * The key keeps combining marks, so Devanagari/Arabic names differing only in
+ * matras stay distinct rather than collapsing into one "employer".
+ */
 function companyKey(name) {
-  return String(name ?? '').toLowerCase().replace(/[^a-z0-9]/g, '');
+  return normalizeTextKey(String(name ?? ''));
 }
 
 /**
--- a/tests/fingerprint-core.test.mjs
+++ b/tests/fingerprint-core.test.mjs
@@ -275,3 +275,36 @@ try {
 } catch (e) {
   fail(`fingerprint-core popcount equivalence tests crashed: ${e.stack || e.message}`);
 }
+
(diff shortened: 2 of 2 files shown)


santifer:
Textbook fix, @abankar1: delegating companyKey() to the shared normalizeTextKey() instead of keeping a second hand-rolled strip is exactly how this class of bug dies for good. The JP/RU/ZH/Devanagari repro made the review easy. Merged. 🚀

PR #2503, opened 2026-08-04
fix(cv-facts): the employer/title gate only matched lowercase triggers

Zoubeir23:
Bug fix, sent straight in per `CONTRIBUTING.md:22`. Same gate as #2460 (merged this morning), other half: that one was the metric claims, this is the employer / title claims.

## A fabricated employer ships unflagged in the spelling CVs use

The `employer` and `title` patterns in `factClaims` are plain `/g` — no `i` flag — so their trigger words only match in lowercase. A CV is written in capitalised bullets, so the phrasing that actually occurs produced **no claim at all**.

End-to-end on `main`, same fabricated employer and title, sources containing neither:

```
generated CV: "<li>Worked at Initech as a Principal Engineer</li>"
  → CV fact check passed ✅          ← the fabrication ships

generated CV: "he worked at Initech as a Principal Engineer"
  → employer: initech                ← identical claim, caught
    title: principal engineer
```

Every capitalised form was invisible: `Worked at …`, `- Worked at …`, `Joined Globex in 2024`, `Employer: Initech`, `Served as Head of Data`.

This is the half of the gate that enforces AGENTS.md's *"Authorship claims are non-negotiable … Tool-of-trade conflation is the most common fabrication pattern and is explicitly forbidden."* The `tool` pattern already carried `gi`, which is why only these two were affected.

## Fix

The **trigger** becomes case-insensitive; the **capture does not**.

```js
['employer', /\b(?:[Ww]orked [Aa]t|[Jj]oined|[Ee]mployer\s*:\s*|[Cc]ompany\s*:\s*)\s*([A-Z][\w&.'-]*…)/g]
```

Adding `i` to the whole regex would have been shorter and wrong: the capture leans on `[A-Z]` to tell a proper noun from ordinary prose, so a case-insensitive capture reads `"Worked at the office as a manager"` as an employer claim. That case is pinned by a test.

## A second defect, exposed by the first

Making the triggers match surfaced something that could not fire before: `stripMarkup` collapsed `</li><li>` to a space, so the title capture — which chains consecutive Capitalised words — ran across the boundary and produced `"Principal Engineer Built"` from two adjacent bullets. No source contains that string, so a **truthful** CV failed the gate.

Block-level tags now become a sentence break (`. `) rather than a space. That is how the markdown sources already behaved — their newlines break the chain — so the two sides of the comparison now normalise the same way.

## Test plan

- [x] 8 new self-test cases. **5 fail on `main`** — verified by porting the block onto `main`'s copy and running it
- [x] Both directions end-to-end through the CLI: the fabricated CV now fails, and a truthful CV whose bullets sit next to each other passes
- [x] False positives pinned in the other direction: `"Worked at the office as a manager"` and `"joined the team as a contractor"` yield no claim
- [x] `node verify-cv-facts.mjs --self-test` — 30 passed, 0 failed
- [x] `node test-all.mjs --quick` — 2916 passed, 0 failed (`tests/nonmetric-fact-gate.test.mjs` and `tests/cover-fact-gate.test.mjs` exercise this path too)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CV fact verification by preventing claims from incorrectly spanning separate lists or content blocks.
  * Enhanced recognition of employer and job-title statements with capitalized phrases and lowercase title connectors.
  * Preserved accurate handling of proper names while reducing false matches in regular prose.

* **Tests**
  * Added coverage for trigger phrases, title connectors, ordinary prose exclusions, and content-block boundaries.


santifer:
Great catch on the gate's blind spot, @Zoubeir23: a capitalised CV bullet is THE phrasing CVs use, and your A/B showed all 8 capitalised forms shipping unflagged. Keeping the capture case-sensitive while relaxing only the triggers is the right balance. That's five merged today — quite a run. 🚀

PR #2512, opened 2026-08-04
feat(cv): strip the Core Competencies section when the payload has no entries

### LINKED ISSUE #2509: Empty Core Competencies payload renders a bare section header in HTML CVs
### Problem

Both HTML templates (`cv-template.html`, `resume-template.html`) wrap the Core Competencies section unconditionally. When a payload has no `competencies` entries, `buildCompetencies()` correctly returns `''`, but nothing removes the surrounding wrapper — so the generated CV shows a bare "Core Competencies" section header with nothing under it. This is exactly the bug #1879 fixed for projects and #2220 pre-empted for awards; competencies was never added to the shared optional-section stripping in `cv-sections-core.mjs`.

A competency tag row is genuinely optional: it is often redundant with the summary and experience bullets that prove the same claims, so payloads legitimately omit it (e.g. to save vertical space on a one-page CV).

### Proposed solution

Add `competencies` to `OPTIONAL_SECTIONS` in `cv-sections-core.mjs` with an HTML boundary pattern for the `<!-- CORE COMPETENCIES -->` marker. HTML-only, like certifications: `cv-template.tex` has no competencies section, so `PATTERNS.tex` gets no key and `stripEmptySections` skips it there.

Competencies is *first* among the optional sections — between Professional Summary and Work Experience — so tests should cover the new failure direction: a boundary slip would swallow the whole experience section rather than a trailing one.

I have a patch ready with test coverage extended in `tests/cv-optional-sections.test.mjs` (`node test-all.mjs` passes); will open a PR right after this issue.

### Area

PDF Generation

lianavalentino:
## What does this PR do?

Adds Core Competencies to the shared optional-section stripping (`cv-sections-core.mjs`), so an empty/absent `competencies` payload removes the section instead of rendering a bare "Core Competencies" header. Same mechanism as projects/education/certifications/awards; HTML-only like certifications (no LaTeX marker exists in `cv-template.tex`).

Test coverage extended in `tests/cv-optional-sections.test.mjs`: competencies joins the payload matrix (`hasCompetencies` per template), plus a competencies-empty-alone case — competencies is *first* among the optional sections, sitting between Professional Summary and Work Experience, so a boundary slip here would swallow the experience section rather than a trailing one. The new checks assert `<!-- WORK EXPERIENCE -->` and `{{EXPERIENCE}}` survive the strip.

## Related issue

Closes #2509

## Type of change

- [ ] Bug fix
- [x] New feature

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Empty Core Competencies sections are now automatically removed from CVs.
  * Other populated sections, including work experience and projects, remain unaffected when competencies are empty.

* **Tests**
  * Expanded coverage for CVs with empty, populated, and mixed optional sections.

### DIFF
--- a/cv-sections-core.mjs
+++ b/cv-sections-core.mjs
@@ -1,12 +1,14 @@
 // Shared optional-section stripping for the CV builders (build-cv-html.mjs,
 // build-cv-latex.mjs).
 //
-// Projects, education, certifications, and awards are the genuinely optional
-// CV sections: a candidate's projects are often already covered under Work
-// Experience, not every candidate has a degree, not every application carries a
-// certification worth listing, and most candidates have no award to name. The
-// templates wrap all four unconditionally, so a payload with no entries renders
-// a bare section header with nothing under it. The builders'
+// Core competencies, projects, education, certifications, and awards are the
+// genuinely optional CV sections: a competency tag row is often redundant with
+// the summary and experience bullets that prove the same claims, a candidate's
+// projects are often already covered under Work Experience, not every
+// candidate has a degree, not every application carries a certification worth
+// listing, and most candidates have no award to name. The templates wrap all
+// five unconditionally, so a payload with no entries renders a bare section
+// header with nothing under it. The builders' buildCompetencies()/
 // buildProjects()/buildEducation()/buildCertifications()/buildAwards()
 // correctly return '' — nothing removes the surrounding wrapper, which is what
 // this module does.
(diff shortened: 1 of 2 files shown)


santifer:
Welcome to career-ops, @lianavalentino, and this is a good first PR: you found the section the shared stripping had missed and fixed it the way the module already works, instead of inventing a second mechanism next to it.

Verified before merging: your four assertions fail against `main` and pass on your branch, the `<!-- CORE COMPETENCIES -->` marker exists in all three HTML templates, and it is followed by `<!-- WORK EXPERIENCE -->`, so the boundary has something to stop at (that detail matters more than it looks: the last section in a template has no following marker and needs different handling). Leaving the LaTeX template out is right too, since it has no competencies block at all.

One thing I noticed that is NOT yours to fix: `cv-template.zh-minimal.html` has the marker but isn't in the test's TEMPLATES list, and it already wasn't before your PR. Inherited gap, not a regression: I'll open it as a good-first-issue.

Merged, and glad to have you here 🚀

PR #2513, opened 2026-08-04
feat(cover): optional letter.signature sign-off block

### LINKED ISSUE #2510: Cover letter: optional signature/sign-off block (valediction + signing name)
### Problem

The cover-letter template has no sign-off. A letter that should end with "Sincerely, / Jane Doe" currently has nowhere to put it — the body just stops after the closing block. Putting the valediction into `letter.closing` mangles the typography: it renders as an ordinary paragraph with no name line under it.

### Proposed solution

An optional `letter.signature` payload key rendering a `<p class="signature">` block after the language-closing block:

- **string form** — `"signature": "Sincerely,"` → valediction over the candidate name (name defaults from `candidate.name`);
- **object form** — `{"valediction": "Best regards,", "name": "J. Doe, PhD"}` for an explicit signing name.

Backward compatible by construction: an unset/null/empty key renders byte-identical output, so no existing payload changes. Both lines are HTML-escaped independently; the `<br>` separator is template markup, never injected into escaped content.

I have a patch ready with a dedicated test suite (`tests/cover-letter-signature.test.mjs`, `node test-all.mjs` passes); will open a PR right after this issue.

### Area

PDF Generation

lianavalentino:
## What does this PR do?

Adds an optional `letter.signature` payload key to the cover-letter builder, rendering a sign-off block (valediction over the signing name) at the end of the letter. Accepts a plain string (`"Sincerely,"` → valediction over the candidate name) or `{ valediction, name }` with an explicit signing name; `name` defaults to `candidate.name`. An unset/null/empty key renders byte-identical output, so every pre-existing payload is untouched.

Both lines are HTML-escaped independently; the `<br>` separator is template markup emitted between them, never injected into escaped content.

New discovered test suite `tests/cover-letter-signature.test.mjs` (11 assertions): backward compatibility (unset/null/`""` byte-identical, no leftover `{{SIGNATURE_BLOCK}}` token), string form, object form incl. name fallback and name-only, block position after the letter body, and escaping (`<script>` in payload never reaches output).

## Related issue

Closes #2510

## Type of change

- [ ] Bug fix
- [x] New feature

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added optional signatures to generated cover letters.
  * Supports custom valedictions and names, with automatic fallback to the candidate’s name.
  * Signature content is safely escaped and omitted when not provided.
  * Added signature formatting and placement below the letter body.

* **Bug Fixes**
  * Preserved existing output when no signature is configured.

### DIFF
--- a/generate-cover-letter.mjs
+++ b/generate-cover-letter.mjs
@@ -118,6 +118,26 @@ function buildFootnotesBlock(footnotes) {
   return `<div class="footnotes">\n${lines}\n  </div>`;
 }
 
+/**
+ * Build the optional sign-off block: a valediction over the signing name.
+ *
+ * Accepts either a plain string (used verbatim as the valediction) or an
+ * object `{ valediction, name }`. `name` defaults to the candidate name so a
+ * payload can set only the valediction. Returns "" when unset, which keeps
+ * every pre-existing payload rendering byte-identical.
+ */
+function buildSignatureBlock(signature, candidateName) {
+  if (!signature) return "";
+  const isObject = typeof signature === "object" && signature !== null;
+  const valediction = isObject ? signature.valediction : signature;
+  const name = (isObject ? signature.name : "") || candidateName || "";
+  if (!valediction && !name) return "";
+  // Each value is escaped independently; the <br> separator is template markup
+  // emitted between them, never injected into escaped content.
+  const lines = [valediction, name].filter(Boolean).map(escapeHtml);
+  return `<p class="signature">${lines.join("<br>")}</p>`;
+}
+
 // Resolve the cover-letter template through the shared resolver so a
 // `cover_letter.template` profile default, an explicit `payload.template`, and
 // installed template packs are all honored. Any resolver failure (no profile,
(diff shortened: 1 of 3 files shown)


santifer:
Your second one today, @lianavalentino, and the design choices are the right ones: accepting either a plain string or `{ valediction, name }` means the common case stays a one-liner while the uncommon one (signing with a different name than the CV header) is still expressible, and defaulting `name` to the candidate name means nobody has to repeat themselves.

The escaping detail is the part I checked hardest, because it's where a sign-off feature usually goes wrong: you escape each line independently and emit the `<br>` between them as template markup, never inside escaped content. That's correct, and it's why a name containing `<` can't inject anything.

Timing note worth naming: I merged #2586 an hour ago, which makes `buildHtml` **throw** on a template token it can't fill. Your PR adds `{{SIGNATURE_BLOCK}}` to both the template and the replacement map, so they land consistent, but I tested the merge locally rather than assuming: both test files pass together, the shipped template renders clean, and I probed a pre-#2513 custom template (no `{{SIGNATURE_BLOCK}}` at all) to confirm it still renders instead of throwing. Full suite on the merge result: 3105/0.

Merged 🚀

PR #2518, opened 2026-08-05
fix(invite-match): every non-Latin company returned zero candidates — the script was inert in 8 shipped locales

### LINKED ISSUE #2517: invite-match: returns ZERO candidates for any non-Latin company — the feature is dead in 8 shipped locales
### What

`normalizeCompanyName()` in `invite-match.mjs` strips to `[a-z0-9 ]`:

```js
.replace(/[^a-z0-9 ]/g, ' ')
```

Every non-Latin company name is therefore deleted entirely and keys to `''`. `matchInvite()` bails on that immediately:

```js
const targetKey = normalizeCompanyName(signals.company);
if (!targetKey) return [];
```

So pasting an interview invite from any company whose name is written in a non-Latin script returns **zero candidates — even when the exact tracker row is present**.

### Reproduced

Tracker with two rows (`アクメ株式会社` and `Acme Inc`):

```
invite "アクメ株式会社"   => candidates: 0     ← its own row is right there
invite "Acme Inc"        => candidates: 1
```

`modes/` ships ja, ko, zh, zh-TW, ru, ua, ar and hi. For every one of those markets this script is not degraded, it is **inert** — and it fails by reporting "no match", which reads as "you haven't applied there" rather than "I can't read this name".

Same root cause as #2429 and #2501, third symptom: there an empty key caused a false *merge*, then a false *skip*; here it causes a **total miss**.

### Fix

Make the class script-preserving (`\p{L}\p{M}\p{N}` + space) and NFKC-normalize first, matching the shared `normalizeTextKey()` contract. Combining marks are kept for the same reason it keeps them — Indic matras have no precomposed form.

The Latin path is deliberately untouched in behaviour: the legal-suffix and generic-descriptor stripping still runs and still produces `acme` for `Acme Corp.` / `Acme Technologies Inc.`. Full-width `ＡＣＭＥ` now folds to `acme` too, which it previously did not.

I did **not** reuse `normalizeTextKey()` wholesale here: this function is deliberately stricter than the tracker key (it strips legal suffixes and keeps spaces for token-overlap scoring), and its docstring explains why. Only the character class is wrong.

PR incoming.

abankar1:
Closes #2517.

`normalizeCompanyName()` stripped to `[a-z0-9 ]`, deleting every non-Latin company name entirely. `matchInvite()` bails on the resulting empty key:

```js
const targetKey = normalizeCompanyName(signals.company);
if (!targetKey) return [];
```

So pasting an interview invite from any company written in a non-Latin script returned **zero candidates even when its own tracker row was present**:

```
tracker: [ アクメ株式会社 , Acme Inc ]

                        BEFORE   AFTER
invite "アクメ株式会社"      0        1     ← its own row was right there
invite "Acme Inc"           1        1
```

`modes/` ships ja, ko, zh, zh-TW, ru, ua, ar and hi. In all of them this script wasn't degraded, it was **inert** — and it failed by reporting "no match", which a user reads as *"you never applied there"* rather than *"I can't read this name"*.

Third symptom of the root cause behind #2429 (empty key → false **merge**) and #2501 (empty key → false **skip**). Here it's a total miss.

## Fix

Script-preserving character class (`\p{L}\p{M}\p{N}` + space) with NFKC first, matching the shared `normalizeTextKey()` contract. Marks are kept for the same reason it keeps them: Indic matras have no precomposed form.

**Deliberately not delegating to `normalizeTextKey()` wholesale.** This function is stricter on purpose — it strips legal suffixes and keeps spaces so `companySimilarity()` can do token overlap — and its own docstring explains why it differs from the tracker key. Only the character class was wrong.

Latin behaviour is unchanged: `Acme Corp.` and `Acme Technologies Inc.` still both key to `acme`. Full-width `ＡＣＭＥ` now folds to `acme`, which it previously did not — a small consistency gain I'm flagging rather than letting ride.

## Tests

Six assertions covering both scripts and the end-to-end match, plus the Latin path staying put.

`invite-match --self-test` → **54 passed, 0 failed** · `invite-match.test.mjs` → 27/0 · `detect-reposts` (the other `normalizeCompanyName` consumer) → 8/0 and 233/0 · `node test-all.mjs --quick` → **2973 passed, 0 failed**.

2 commits (fix + test).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved company-name matching for international names across different writing systems.
  - Added support for full-width characters and combining marks.
  - Prevented distinct company names from being incorrectly treated as identical.
  - Improved matching accuracy against tracker records.

### DIFF
--- a/invite-match.mjs
+++ b/invite-match.mjs
@@ -108,10 +108,20 @@ const GENERIC_DESCRIPTORS = [
  */
 export function normalizeCompanyName(name) {
   let key = String(name ?? '')
+    // NFKC before folding so full-width and half-width spellings of the same
+    // name compare equal, matching the shared normalizeTextKey() contract.
+    .normalize('NFKC')
     .toLowerCase()
     .replace(/\([^)]*\)/g, ' ')
     .replace(/&/g, ' and ')
-    .replace(/[^a-z0-9 ]/g, ' ')
+    // Letters and digits of ANY script, not just [a-z0-9]: the Latin-only
+    // class DELETED every non-Latin name, so アクメ株式会社 and Яндекс both
+    // produced '' — and matchInvite() bails on an empty key, so pasting an
+    // invite from any company in the ja/ko/zh/zh-TW/ru/ua/ar/hi markets that
+    // modes/ ships returned ZERO candidates even when the row was right
+    // there. Combining marks are kept for the same reason normalizeTextKey
+    // keeps them: Indic matras have no precomposed form (#2517).
+    .replace(/[^\p{L}\p{M}\p{N} ]/gu, ' ')
     .replace(/\s+/g, ' ')
     .trim();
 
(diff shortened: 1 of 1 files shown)


santifer:
Textbook, @abankar1: you filed the issue, prescribed the exact fix in it, and opened the PR a minute later. The A/B is unambiguous (アクメ株式会社, Яндекс and 삼성전자 all keyed to empty on main and match correctly here) and the pure-ASCII battery comes out byte-identical, which is the part that makes it safe to merge. Keeping \p{M} in the class is what stops Devanagari matras from collapsing two different companies into one key: a competing PR dropped it and that's exactly what happened. Merged. 🚀

PR #2520, opened 2026-08-05
fix(a16z-speedrun): one transient page failure aborted the whole board — shared per-page retry

### LINKED ISSUE #2506: a16z provider: transient upstream failures abort the whole board fetch — add per-page retry
While exercising the a16z speedrun provider against the live feed tonight (post-#2419, so the full board paginates), the upstream API went through a rough patch: read timeouts above 15s and a burst of HTTP 500s across pages 3-13, confirmed with ~100 consecutive raw fetches outside the provider.

**Current behavior:** one bad page aborts the entire provider run. For a board this size (350 pages at 50/page) that turns a transient blip into a zero-result run.

**Proposal:** per-page retry with short backoff (2-3 attempts on 5xx/timeout) before giving up, and if a page still fails after retries, fail the provider loudly (as today) rather than returning a partial set silently — partial-and-quiet would be worse than empty-and-loud.

**Acceptance:**
- A transient 500/timeout on one page no longer kills the run.
- A persistent failure still errors loudly (no silent partial).
- Test with a mock that fails once then succeeds.

Applies cleanly to `providers/a16z-speedrun.mjs`; worth checking whether `_http.mjs` is the right layer so every provider benefits.

abankar1:
Closes #2506.

The paging loop called `ctx.fetchJson` bare, so a single read timeout or 5xx anywhere in a ~350-page sweep threw and aborted the provider — a transient upstream blip became a zero-result run.

## On your layering question

> worth checking whether `_http.mjs` is the right layer so every provider benefits

**It is, and the evidence is already in the tree.** `workday.mjs` has carried `fetchPageWithRetry` privately (its own `MAX_RETRIES`, `sleep`, `parseRetryAfterMs`, `isRetryableError`), and `oraclecloud.mjs` has grown a third copy. Nothing was shared. So this adds `fetchJsonWithRetry` to `_http.mjs` — the module every provider already imports — rather than giving a16z a fourth hand-rolled version.

The semantics are **workday's, lifted, not re-derived**: exponential backoff with jitter, and a `Retry-After` that is honoured but **clamped**, so a hostile or misconfigured `Retry-After: 86400` can't stall a sweep. Retryable = 429, any 5xx, or a transport error with no status; a non-429 4xx fails fast, because retrying a request the server has already rejected only burns time.

**One deliberate design point:** it does *not* decide what happens when retries are exhausted — it rethrows, and the caller chooses. That policy genuinely differs per provider. `workday` truncates the tenant with a warning and keeps the pages it has; a16z must **fail loudly**, per your acceptance criterion that partial-and-quiet is worse than empty-and-loud. Baking either policy into the shared helper would have forced one of them to be wrong.

## Scope

I did **not** convert `workday.mjs` or `oraclecloud.mjs` in this PR. They're load-bearing, their terminal policies differ from a16z's, and workday's truncation path is covered by tests I'd rather not disturb inside a bug fix. Converging them onto the shared helper is a clean follow-up — happy to take it if you want it.

## Tests

Four cases, all driven through an injected `ctx.sleep` so nothing wall-clock waits:

```
✅ fetch() retries a transient 5xx and still returns the page (#2506)
✅ fetch() backs off between attempts via ctx.sleep (#2506)
✅ fetch() gives up loudly after bounded retries, never a silent partial (#2506)
✅ fetch() does not retry a non-retryable 4xx (#2506)
```

`node test-all.mjs --quick` → **2977 passed, 0 failed**. `workday` and `oraclecloud` provider suites re-run and green (their own retry paths untouched).

3 commits (shared helper → provider fix → tests).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when fetching provider data during temporary server or network failures.
  * Added bounded retries with backoff and jitter for rate limits and transient errors.
  * Honored retry timing guidance when provided by the service.
  * Non-retryable errors continue to fail immediately, while persistent failures are reported after retry attempts are exhausted.

* **Tests**
  * Added coverage for successful retries, persistent server errors, rate-limit delays, and non-retryable client errors.

### DIFF
--- a/providers/_http.mjs
+++ b/providers/_http.mjs
@@ -58,8 +58,18 @@ export async function fetchText(url, opts = {}) {
   return fetchWithTimeout(url, opts, (res) => res.text());
 }
 
-/** Retry policy shared by providers that paginate a large board. */
-const RETRY_DEFAULTS = { retries: 3, baseDelayMs: 500, maxDelayMs: 8_000 };
+/** Jitter added to a backoff so concurrent retries don't re-collide in lockstep. */
+const JITTER_MS = 250;
+
+/**
+ * Retry policy shared by providers that paginate a large board.
+ *
+ * Two retries = three total attempts, matching what #2506 asked for. workday's
+ * private copy used three (four attempts); it can pass `{ retries: 3 }`
+ * explicitly if it converges onto this helper, which is exactly why the policy
+ * is a parameter rather than baked in.
+ */
+const RETRY_DEFAULTS = { retries: 2, baseDelayMs: 500, maxDelayMs: 8_000 };
 
 /** Awaitable sleep that honours a ctx-supplied clock, so tests never wall-clock wait. */
 function sleep(ms, ctx) {
@@ -123,11 +133,15 @@ export async function fetchJsonWithRetry(ctx, url, opts = {}, policy = {}) {
     } catch (err) {
       lastErr = err;
       if (attempt === retries || !isRetryableError(err)) throw err;
-      const backoff = Math.min(baseDelayMs * 2 ** attempt, maxDelayMs);
+      // Cap the backoff at maxDelayMs MINUS the jitter, so the jittered total
+      // still honours the policy limit. Clamping the sum instead would erase
+      // the jitter exactly at the cap — where every retry has converged on the
+      // same delay and de-synchronising them matters most.
+      const backoff = Math.min(baseDelayMs * 2 ** attempt, maxDelayMs - JITTER_MS);
       const retryAfterMs = parseRetryAfterMs(err?.retryAfter);
       const delayMs = retryAfterMs !== null
         ? Math.min(retryAfterMs, maxDelayMs * 4)
-        : backoff + Math.random() * 250;
+        : backoff + Math.random() * JITTER_MS;
       await sleep(delayMs, ctx);
     }
   }
--- a/providers/_http.mjs
+++ b/providers/_http.mjs
@@ -137,11 +137,17 @@ export async function fetchJsonWithRetry(ctx, url, opts = {}, policy = {}) {
       // still honours the policy limit. Clamping the sum instead would erase
       // the jitter exactly at the cap — where every retry has converged on the
       // same delay and de-synchronising them matters most.
-      const backoff = Math.min(baseDelayMs * 2 ** attempt, maxDelayMs - JITTER_MS);
+      //
+      // The jitter itself is clamped to maxDelayMs first: a caller passing a
+      // maxDelayMs below JITTER_MS would otherwise drive the backoff negative
+      // and hand ctx.sleep a negative delay.
+      const jitterMs = Math.min(JITTER_MS, Math.max(0, maxDelayMs));
+      const ceiling = Math.max(0, maxDelayMs - jitterMs);
+      const backoff = Math.min(baseDelayMs * 2 ** attempt, ceiling);
       const retryAfterMs = parseRetryAfterMs(err?.retryAfter);
       const delayMs = retryAfterMs !== null
         ? Math.min(retryAfterMs, maxDelayMs * 4)
-        : backoff + Math.random() * JITTER_MS;
+        : backoff + Math.random() * jitterMs;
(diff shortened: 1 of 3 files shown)


abankar1:
Both CodeRabbit findings addressed — one as suggested, one differently, and the difference matters.

**1. Attempt count — taken as suggested.** `retries: 3` made the inclusive loop issue four fetches; #2506 asked for 2–3. Default is now `2` (three total attempts), and the persistent-failure test expects `attempts === 3`.

I'd originally kept 3 for parity with workday's private copy. The better resolution is the one the design already allows: the *default* should serve what the issue asked for, and workday can pass `{ retries: 3 }` explicitly if it converges onto this helper — which is exactly why the policy is a parameter rather than baked in.

**2. Jitter clamp — real bug, different fix.** The finding is correct: `backoff` was capped at `maxDelayMs` and *then* had up to 250ms added, so a delay could exceed the stated limit.

I didn't take the suggested `Math.min(backoff + jitter, maxDelayMs)`. At the cap that pins every retry to exactly `maxDelayMs` and **erases the jitter** — precisely where it matters most, since that's the point at which concurrent retries have all converged on the same delay and de-synchronising them is the whole reason the jitter exists. Capping the backoff at `maxDelayMs - JITTER_MS` keeps the total inside the limit *and* keeps the spread:

```js
const backoff = Math.min(baseDelayMs * 2 ** attempt, maxDelayMs - JITTER_MS);
…
: backoff + Math.random() * JITTER_MS;
```

Pinned with a test asserting both halves, since one without the other is a regression waiting to happen:

```
✅ retry backoff never exceeds maxDelayMs once jitter is added (#2506)
✅ jitter survives at the cap, so concurrent retries de-synchronise (#2506)
```

`node test-all.mjs --quick` → **2979 passed, 0 failed**.

abankar1:
Both follow-up findings are correct, and both are on my own previous fix — taken.

**1. `maxDelayMs` below `JITTER_MS` drove the backoff negative.** Real bug, introduced by my clamp: `maxDelayMs: 100` made `maxDelayMs - JITTER_MS` equal `-150`, and `ctx.sleep` got a negative delay. The jitter is now clamped to `maxDelayMs` first and the ceiling floored at 0:

```js
const jitterMs = Math.min(JITTER_MS, Math.max(0, maxDelayMs));
const ceiling  = Math.max(0, maxDelayMs - jitterMs);
```

```
maxDelayMs=100    delays: 48.1, 35.9, 64.7, 67.3      | all >=0 ✔ | all <=cap ✔
maxDelayMs=250    delays: 58.5, 39.6, 200.9, 121.3    | all >=0 ✔ | all <=cap ✔
maxDelayMs=8000   delays: 586.9, 937.6, 1633.1, 3390.8| all >=0 ✔ | all <=cap ✔
```

**2. The jitter test didn't test jitter.** This is the sharper catch and you're right: `jittered` looked at *every* delay, which included the pre-cap 400/800 steps, so it would have passed with all cap-level delays identical. I asserted a property and then verified something weaker — worse than having no test, since I cited it as proof.

Rewritten to stub `Math.random` for determinism and check **only the capped attempts** (`slept.slice(2)`). Confirmed it now fails under exactly the approach it exists to rule out:

```
# with clamp-the-sum (the original suggestion)
❌ capped delays were identical — jitter erased at the cap: [1000,1000,1000]

# with the ceiling fix
✅ jitter survives AT THE CAP, so concurrent retries de-synchronise (#2506)
```

So the disagreement resolves cleanly: clamping the sum really does erase the jitter at the cap, and there is now a test that proves it rather than my say-so. Plus a case pinning the sub-jitter `maxDelayMs` path.

`node test-all.mjs --quick` → **2980 passed, 0 failed**.

santifer:
This closes the issue I opened on Tuesday after watching the feed throw 500s across a dozen pages, @abankar1, and you built exactly what it asked for: bounded retries on 5xx and timeouts, a persistent failure still loud rather than a silent partial, and a test that fails once and then succeeds. Verified by running it, not by reading it.

That matters more here than it looks: this provider indexes the largest board we carry, so "one bad page kills the run" meant an entire source vanishing for a cycle over a blip. Merged. 🚀

PR #2523, opened 2026-08-05
fix(arbeitsagentur): move to the v6 Jobsuche API

### LINKED ISSUE #2494: Arbeitsagentur provider: all requests fail with HTTP 404 (v4 endpoint)
### Code of Conduct

### Existing issues

### What happened?

`providers/arbeitsagentur.mjs:31` uses:
`https://rest.arbeitsagentur.de/jobboerse/jobsuche-service/pc/v4/jobs`

As of 2026-08-04 every keyword request returns HTTP 404, so all
Arbeitsagentur entries fail. `verify-portals` reports "slug not found"
and `scan.mjs` lists them as unreachable targets.

Probed with the public key `jobboerse-jobsuche`:

    404  /pc/v4/jobs
    404  /pc/v4/app/jobs
    404  /pc/v5/jobs
    404  /pc/v5/app/jobs
    403  /pc/v2/jobs
    403  /pc/v3/jobs
    403  /v4/jobs
    403  /jobs

The 403/404 split suggests the key still authenticates (the /pc/v4 and
/pc/v5 routes are registered and reach a backend that returns 404),
while the other paths aren't routed at all. So this looks like a
backend/contract change rather than a revoked key or a moved version
prefix.

A second key returned 403 on every path.

Repro:
    curl -s -o /dev/null -w "%{http_code}\n" \
      -H "X-API-Key: jobboerse-jobsuche" \
      "https://rest.arbeitsagentur.de/jobboerse/jobsuche-service/pc/v4/jobs?was=Cloud%20Engineer&size=5"

Environment: career-ops v1.24.0, Node v24.16.0, macOS, requests from Germany.

### Steps to reproduce

1. Add an Arbeitsagentur entry to `portals.yml`:

```yaml
   - name: Arbeitsagentur — Cloud Engineer
     provider: arbeitsagentur
     arbeitsagentur:
       keywords: ["Cloud Engineer"]
       days: 30
       size: 100
     enabled: true
```

2. Run `npm run verify:portals`
   → `❌ arbeitsagentur (slug not found) — all 1 keyword request(s) failed — "Cloud Engineer": HTTP 404 Not Found`

3. Run `npm run scan`
   → entry is listed under "target(s) unreachable (slug?)", no jobs returned

4. Hit the endpoint directly:

```bash
   curl -s -o /dev/null -w "%{http_code}\n" \
     -H "X-API-Key: jobboerse-jobsuche" \
     "https://rest.arbeitsagentur.de/jobboerse/jobsuche-service/pc/v4/jobs?was=Cloud%20Engineer&size=5"
```
   → `404`

Reproduces with any keyword, with or without `wo`/`umkreis`.

### Expected behavior

The provider should return job results for the given keywords, as it did
before 2026-08-04.

Two things would help regardless of the endpoint fix:

1. The failure surfaces as "slug not found", which is misleading — there
   is no slug involved. A 404 on the provider's own fixed API path is a
   dead-provider condition, not a user misconfiguration, and the current
   wording sends people looking for a config error that isn't there.

2. Since the endpoint is hardcoded in `providers/arbeitsagentur.mjs`,
   users can't work around a Bundesagentur-side change without editing a
   system file, which `npm run update` then overwrites. An optional
   `api_base` override on the entry would let users patch it from the
   user layer until upstream catches up.

### CLI tool

Claude Code v2.1.221

### OS and Node.js version

macOS and v24.16.0

### Output of npm run doctor

```shell
career-ops doctor
================
✓ Node.js >= 22.5 (v24.16.0)
✓ Dependencies installed
✓ Scan extractor: cli (browser-extract.mjs)
✓ cv.md found
✓ config/profile.yml found
✓ modes/_profile.md found
✓ portals.yml found
✓ Fonts directory ready
✓ data/ directory ready
✓ data/pipeline.md ready
✓ output/ directory ready
✓ reports/ directory ready

Result: All checks passed (1 warning — see above).

OS: macOS
career-ops: v1.24.0
```

maxmilian:
## What does this PR do?

Moves the Arbeitsagentur provider from the withdrawn v4 Jobsuche API to v6, remapping the renamed response fields. Also narrows `remoteMatch: 'filter'`, whose per-candidate proof no longer exists in v6 — see the trade-off below, it is the one judgement call here.

## Related issue

Fixes #2494

## Type of change

- [x] Bug fix
- [ ] New feature

## The query side is fine; the response side is not

@gowrishacv's report stops at v5, so the first thing worth adding is that **v6 answers**:

| endpoint | status |
|---|---|
| `/pc/v4/jobs` | 404 |
| `/pc/v4/app/jobs` | 404 |
| `/pc/v5/app/jobs` | 404 |
| **`/pc/v6/jobs`** | **200** |

v6 accepts every parameter this provider already sends — `was`, `wo`, `umkreis`, `veroeffentlichtseit`, `angebotsart`, `homeoffice`, `page`, `size` — all 200. So the query builder is untouched.

The response shape changed:

| v4 | v6 |
|---|---|
| `stellenangebote[]` | `ergebnisliste[]` |
| `refnr` | `referenznummer` |
| `titel` | `stellenangebotsTitel` |
| `arbeitgeber` | `firma` |
| `arbeitsort` (object) | `stellenlokationen[]` → `[0].adresse` |

The public job-detail page still resolves by reference number, so outgoing URLs are unchanged.

**One deliberate behaviour change in the location string:** v4's `region` was a display name and was joined onto the city (`Berlin, Berlin`). In v6 it is an uppercase federal-state enum, so the same join yields `Stuttgart, BADEN_WUERTTEMBERG` — noise in a string `scan.mjs`'s commute filter has to match. The city now stands alone. Postings can carry several locations (11 of 100 in a sampled page); the first is used, as v4's single field effectively was.

## The trade-off: `remoteMatch: 'filter'` lost its proof

This is the part I'd most like your read on.

`'filter'` used to query `homeoffice=nv_true` and then confirm each hit's `homeofficetyp: VOLLSTAENDIG` through the detail endpoint — because `nv_true` alone also returns `NACH_VEREINBARUNG`, an office-anchored hybrid. In v6 that proof is unavailable:

- the detail endpoint answers **403** to this public client key (`/pc/v6/jobdetails/<refnr>`), and v4's is 404;
- the only home-office field on a v6 search hit is the boolean `homeofficemoeglich`, which is exactly what `nv_true` filtered on — **a sampled `nv_true` page came back 100% `true`**. It cannot separate fully-remote from hybrid.

Rather than tag on the boolean, `'filter'` now keeps the server-side narrowing and applies **the same title check `'title'` mode uses**. A candidate whose title makes no remote claim keeps its real city — the same fail-closed outcome an unverifiable lookup had in v4. That matters because the `Deutschlandweit (Homeoffice)` marker exempts a job from the commute filter, so tagging on `nv_true` alone would smuggle every hybrid past it, which is the exact defect #1981 fixed.

Net effect: `'filter'` is now `'title'` applied to a better candidate set, not a stronger standard of proof. If you'd rather it hard-fail, or fall back to `'title'` with a warning, or be removed until an authenticated detail route is available, say which and I'll change it — I picked the option that keeps existing `portals.yml` configs working without weakening the tagging rule.

## Validation

- `node test-all.mjs` → **2974 passed, 0 failed** (2 pre-existing environment warnings: no user data, no Playwright browser).
- Provider tests rewritten onto v6 fixtures, including a case pinning `/pc/v6/jobs` in the request URL so a silent revert can't reintroduce a dead endpoint, and one asserting an `homeofficemoeglich: true` hit still goes untagged when its title makes no remote claim.
- The detail-batching tests went with the lookups they covered; the dedup-across-pagination case they also exercised is kept.
- **End-to-end against the live API** (`keywords: ['Machine Learning Engineer'], wo: Berlin, umkreis: 50, remoteMatch: 'filter'`):

```
total: 10
  Data Engineer* / Machine Learning Engineer*
    inovex GmbH | Berlin
    https://www.arbeitsagentur.de/jobsuche/jobdetail/16752-r5446-S
  Machine Learning Engineer (all genders)
    Parship | Berlin
    https://www.arbeitsagentur.de/jobsuche/jobdetail/13644-308449-S
  Principal Machine Learning Scientist (alle Geschlechter)
    Bayer AG Pharmaceuticals | Berlin
    https://www.arbeitsagentur.de/jobsuche/jobdetail/12708-1575645-1-S
tagged nationwide-remote: 0
```

Real titles, employers, cities and resolvable detail URLs. `tagged 0` is the conservative path working as intended — none of that batch's `nv_true` candidates claimed remote in the title.

## Checklist

- [x] If this is a new feature or architecture change, I opened an issue first (bug fix — exempt, and #2494 exists)
- [x] My changes respect the [Data Contract](https://github.com/santifer/career-ops/blob/main/DATA_CONTRACT.md) (system layer only: `providers/`, `tests/`)

<sub>One unrelated line: `providers/vdab.mjs`'s batching comment cross-referenced `arbeitsagentur.mjs`'s `VERIFY_BATCH`, which this PR deletes. Reworded so it doesn't point at something that no longer exists.</sub>

## Summary by CodeRabbit

* **Improvements**
  * Updated Arbeitsagentur job listings to use the latest API format.
  * Improved display of job titles, companies, reference numbers, and locations.
  * Enhanced search result handling, pagination, and duplicate prevention.
  * Remote-job filtering now requires a remote-work indication in the job title; other listings retain their actual location.
* **Documentation**
  * Updated configuration guidance to reflect the latest Arbeitsagentur API behavior.

### DIFF
--- a/providers/arbeitsagentur.mjs
+++ b/providers/arbeitsagentur.mjs
@@ -20,21 +20,23 @@
 //       size: 100               # results per keyword (1–100, default 100)
 //       remoteNationwide: true  # also run a nationwide pass keeping remote-eligible hits
 //       remoteMatch: filter     # how that pass detects remote (default 'title'):
-//                               #   'filter' — server-side `homeoffice=nv_true` query + pagination, then one
-//                               #              detail call per hit to confirm `homeofficetyp: VOLLSTAENDIG`
-//                               #              (nv_true alone also returns hybrid roles). Recommended.
+//                               #   'filter' — server-side `homeoffice=nv_true` query + pagination to narrow
+//                               #              the set, then the same title check as 'title'. Recommended:
+//                               #              same standard of proof, applied to a far better candidate set.
+//                               #              (v4 confirmed `homeofficetyp: VOLLSTAENDIG` per hit via the
+//                               #              detail endpoint; v6 no longer serves it to this key — #2494.)
 //                               #   'title'  — regex on the job title only (cheap; misses body-level remote)
 //                               #   'off'    — skip the remote pass entirely
 //       remoteMaxPages: 10      # 'filter' mode: max pages to paginate (size each); default 1
 //     enabled: true
 
-const API_URL = 'https://rest.arbeitsagentur.de/jobboerse/jobsuche-service/pc/v4/jobs';
+// v6. The v4 search and detail endpoints both 404 as of 2026-08-04 (#2494);
+// v5 does too. v6 keeps every query parameter this provider sends
+// (was/wo/umkreis/veroeffentlichtseit/angebotsart/homeoffice/page/size) but
+// renames the response fields — see normalizeJob().
+const API_URL = 'https://rest.arbeitsagentur.de/jobboerse/jobsuche-service/pc/v6/jobs';
 const API_KEY = 'jobboerse-jobsuche'; // public client key the arbeitsagentur.de UI uses
 const DETAIL_BASE = 'https://www.arbeitsagentur.de/jobsuche/jobdetail/';
-const DETAIL_API = 'https://rest.arbeitsagentur.de/jobboerse/jobsuche-service/pc/v4/jobdetails/';
-// Verify in small batches so a wide remote pass can't fire hundreds of detail
-// requests at once.
-const VERIFY_BATCH = 5;
 const REMOTE_RE = /(remote|homeoffice|home[-\s]?office|ortsunabh|deutschlandweit|bundesweit|100\s*%|full[-\s]?remote|fully remote)/i;
 
 // Clamp a runtime integer into [min, max], falling back to `def` for NaN, so a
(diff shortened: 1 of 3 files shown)


santifer:
Another solid one, @maxmilian. The German federal board had been answering 404 since the v4 retirement, so this brings a whole national source back from the dead: the live A/B (main throws, this serves) made the review easy. On the remoteMatch mode: I'm keeping your call. Losing a field and choosing to say less is the conservative direction, and it beats tagging roles on a signal the new API no longer gives us. Merged. 🚀

(Ignore the welcome bot, by the way: it greeted you as a first-timer on your 24th merge.)

PR #1305, opened 2026-06-29
feat(providers): add The Hub provider (public zero-auth job-board API)

Scott-Emberson:
## What

Adds a **The Hub** board-wide aggregator provider (`providers/thehub.mjs`). The Hub (Nordic/EU startup jobs) publishes a public, zero-auth JSON feed at `https://thehub.io/api/jobs`, so this mirrors `providers/remotive.mjs` with bounded pagination like `providers/arbeitnow.mjs`.

- Reads `?page=N` (15 jobs/page, 1-indexed) and normalizes each `docs[]` item to `{ title, url, company, location }` (+ optional `postedAt`):
  - `title` from `title`.
  - `url` from `absoluteJobUrl`, **host-locked to `thehub.io`** (the per-posting `link` field is an external apply URL — deliberately not used). Off-host/non-https URLs drop the posting; `url` is the dedup key.
  - `company` from `company.name`, falling back to the entry name, then `"The Hub"`.
  - `location` from `location.address` (else `location.locality` / `location.country`), with `"Remote"` appended when `isRemote`.
  - `postedAt` from `publishedAt` (else `createdAt`) ISO date → ms.
- Paginates up to `max_pages` (default 3, cap 67), stopping early on a short page **or** when `page >= pages` (the response reports `pages`/`total`). Host-locked to `thehub.io` with `redirect:'error'` on every request (SSRF guard).

## Files

- `providers/thehub.mjs` — new provider.
- `test-all.mjs` — new test block (section 36): normalize mapping, company + date fallbacks, host-lock + drops, pagination (short-stop, pages-stop, `max_pages` cap), SSRF guard, bad-response throw.
- `docs/SUPPORTED_JOB_BOARDS.md` — register The Hub (the doc requires this in the same PR).
- `templates/portals.example.yml` — add a The Hub example with `max_pages`.

## Testing

- `node test-all.mjs` → **556 passed, 0 failed** (12 new assertions).
- Live end-to-end against the real API returned 30 correctly-normalized jobs across 2 pages (all with `postedAt`; real company names).

Closes #1304

## Summary by CodeRabbit

* **New Features**
  * Added support for “The Hub” as a new job board provider, including paginated job fetching and normalization of job fields (title, company, location, and posted date).

* **Documentation**
  * Documented The Hub’s API details, pagination/max-pages behavior, and when filtering is applied.
  * Updated the portal configuration example to include The Hub.

* **Tests**
[...]

PR #2524, opened 2026-08-05
fix(providers): move The Hub to the v2 jobsandfeatured API

ilyakanevskiy:
## What does this PR do?

The Hub's `/api/jobs` feed (added in #1305) now 404s. The replacement is `/api/v2/jobsandfeatured`, which changed both the response envelope and several field names:

| v1 | v2 |
|---|---|
| `docs[]` | `jobs.docs[]` |
| `absoluteJobUrl` | *(removed — URL now built from `id`: `https://thehub.io/jobs/{id}`)* |
| `publishedAt`/`createdAt` → `postedAt` | *(removed — no date field in v2 at all)* |

Two further behavior changes, both confirmed via Playwright network capture on the live site (not guessed from the response shape alone):

- **`countryCode` is now required.** Omitting it scopes results to the caller's geo-IP — observed as low as 49 jobs from one egress vs. 332 with `countryCode=EU`. That would make scan coverage depend on where the scan happens to run, so it's now a required query param, configurable (default `EU`) via a `thehub:` block.
- **The site's "Remote jobs only" filter is a separate query mode, not a `countryCode` value.** `countryCode=REMOTE` (the client-side routing token used in the URL) doesn't correspond to anything the API accepts directly — checking the box actually sends `isRemote=true` with no `countryCode` at all. So covering both a region and remote postings takes two separate paginated passes, added as an opt-in `thehub: { includeRemote: true }` and merged/deduped by job id.

## Related issue

N/A — bug fix, exempt from the issue-first requirement per CONTRIBUTING.md.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] My changes respect the [Data Contract](https://github.com/santifer/career-ops/blob/main/DATA_CONTRACT.md) (system layer only: `providers/`, `tests/`, docs, template)

## Changes

- `providers/thehub.mjs`:
  - Endpoint moved to `https://thehub.io/api/v2/jobsandfeatured`; reads `jobs.docs`/`jobs.pages` instead of the old top-level `docs`/`pages`.
  - `normalizeHubJob` builds the job URL from `id` (host-locked to `thehub.io`, URL-encoded) instead of the now-absent `absoluteJobUrl`; drops `postedAt` from the normalized shape (no date field survives in v2).
  - New `parseThehubConfig` reads an optional `thehub:` block: `countryCode` (default `EU`) and `includeRemote` (default `false`).
  - `fetch()` now runs the region pass and, if `includeRemote` is set, a second `isRemote=true` pass, merging both into one deduped result set by URL.
- `tests/providers/thehub.test.mjs`: rewritten onto the v2 fixtures — id-based URL construction/encoding/drops, `parseThehubConfig` defaults and overrides, the new `countryCode`/`isRemote` pass-through and merge-dedup behavior, plus the existing pagination/cap/SSRF coverage carried forward unchanged.
- `docs/SUPPORTED_JOB_BOARDS.md`, `templates/portals.example.yml`: updated endpoint and documented the new `thehub:` config block.

## Testing

- `node test-all.mjs` → all tests pass, no regressions.
- End-to-end against the live API (`countryCode=EU`, `includeRemote: true`): confirmed real job titles/companies/locations and resolvable `https://thehub.io/jobs/{id}` URLs across both passes; the region pass currently covers 332 postings (23 pages), the remote pass 49 (4 pages).

## Context for @Scott-Emberson

Tagging you since you built the original provider in #1305 — wanted to flag the break and loop you in in case there's context on the v1→v2 migration I don't have from the outside (e.g. whether The Hub announced this, or whether `countryCode`/`isRemote` have other accepted values worth documenting). No action needed, just a courtesy ping.

## Summary by CodeRabbit

- **New Features**
  - Updated The Hub integration to use the latest jobs endpoint.
  - Added configurable country-code filtering, defaulting to EU.
  - Added optional remote-job retrieval with automatic duplicate removal.
  - Improved job URL generation and validation.

- **Documentation**
  - Documented country and remote-job settings.
  - Updated the example portal configuration.

- **Bug Fixes**
  - Improved handling of API responses, pagination, missing locations, and unexpected results.
  - Preserved previously retrieved jobs when later requests fail.


Scott-Emberson:
Thanks for the ping. Context answers first, then a few notes from a pass over the diff.

On v1 vs v2: I have nothing insider, I'm afraid. When I built the provider in #1305 the v1 feed took a bare `?page=N` and nothing else. No geo param ever existed, so v1 coverage was already whatever the API did with the caller's geo-IP, and nobody measured it. Your egress finding may well have applied to v1 silently. I never saw The Hub announce API versions or document accepted `countryCode` values either.

One reassuring continuity check: the v1 `absoluteJobUrl` values (at least in the #1305 fixtures) were exactly the `https://thehub.io/jobs/{id}` form, and scan-history dedup is URL-keyed. Building URLs from `id` should therefore keep already-seen jobs deduped instead of re-entering the pipeline as reposts.

Review notes:

1. The one substantive ask: the new two-pass structure skips the `succeededOnce` idiom the other multi-pass providers adopted around #2379 (tencent.mjs:106, meituan.mjs:123, alibaba.mjs:136, phenom.mjs:156, radancy.mjs:284, successfactors.mjs:321 with `firstErr`). As written, if the region pass completes and the remote pass then fails on page 1, or either pass fails mid-pagination, the whole merged result set is discarded and the board records a failure. The sibling shape is: track `succeededOnce` across both passes, wrap each page fetch, and on error `if (!succeededOnce) throw err; break;` so partials survive once any request has resolved. Total failure still throws, which your diff already gets right: no try/catch swallows the first request, so a dead board propagates to scan.mjs exactly as #2379 requires.

2. A test gap worth closing alongside that change: nothing currently pins that a `fetchJson` rejection propagates out of `fetch()`. A future well-meaning try/catch that swallows errors would pass the whole suite. One test asserting a page-1 rejection rejects the provider call would pin the #2379 contract; partial-failure coverage then comes free with the fix above.

3. Doc sentence, take it or leave it: since v2 carries no date field, thehub rows are now permanently exempt from `--max-age-days` (undated jobs pass that filter by design, scan.mjs:306). Unavoidable given the API, just worth a line in the SUPPORTED_JOB_BOARDS row so nobody hunts for a bug later.

Everything else checked out for me: dedup is first-wins and deterministic from `id`, pagination is hard-bounded by the 67-page clamp regardless of what the API claims, `encodeURIComponent` keeps config values out of the query string, and the missing-location to "Remote" mapping matches what you saw live. The Playwright verification of the `isRemote`/`countryCode` split was the part I'd have gotten wrong from the response shape alone. Solid migration.

ilyakanevskiy:
Thanks @Scott-Emberson -- I tagged you as a courtesy ping not expecting much, but this turned out to be a genuinely thorough and useful review.

Fixed both code points: the two-pass fetch now tracks `succeededOnce` across the region/remote passes (same idiom as tencent/meituan/alibaba/phenom/radancy/successfactors after #2379), so a dead board still throws but a later blip in either pass keeps what's already collected -- plus regression tests for both the partial-failure path and the first-request-fails contract.

Left the `SUPPORTED_JOB_BOARDS.md` note out, though: skimmed the rest of the table and none of the other rows document downstream filtering side-effects like this (missing-date fields, etc.) -- it'd be the only row calling that out, so it didn't fit the doc's existing convention.

santifer:
This brings a dead source back, @ilyakanevskiy: the old endpoint has been answering 404 for days and v2 serves the board zero-auth (324 roles across 22 pages when I checked). Two things I want to call out because they're the parts people skip: you dropped a real company name from the fixture, and you kept the redirect:'error' guard while narrowing the URL surface to a constant host plus an encoded id, which is strictly safer than parsing a payload-supplied URL. @Scott-Emberson's first two review points are now pinned by tests. His third one, a docs note that Hub rows carry no date and are therefore exempt from date filters, I've just added myself so this doesn't wait another day. Merged. 🚀

PR #2526, opened 2026-08-05
fix(seeds): YC portfolio walk stopped after page 1

### LINKED ISSUE #2525: --seeds yc only probes ~25 companies since the YC API changed pagination
Running `node scan-ats-full.mjs --seeds yc` today probes only 25 companies and reports 24 unreachable boards, 0 matches.

Root cause: the YC companies API (`api.ycombinator.com/v0.1/companies`) now caps page size at around 30 items and ignores the `per_page=1000` param. `fetchYCCompanies()` stops when a page returns fewer than 1000 companies, so the walk always ends on page 1. The only companies it ever sees are the newest batch, and none of them have ATS boards yet, hence the 404 wall.

The API response does include `page`, `totalPages` (246 today) and `nextPage`, so the fix is to follow those instead of guessing from batch size. With that change a sweep probes 4,180 active portfolio companies.

Happy to send a PR, I have the fix working locally.

srijanarya:
Fixes #2525

`--seeds yc` sweeps only ever saw the ~25 newest YC companies because the YC API now caps page size at ~30 items and ignores `per_page`. The `batchSize < 1000` stop condition ended the walk on page 1, so the sweep probed only brand-new companies with no ATS boards and always reported 0 matches.

Changes in `seeds/vc-portfolios.mjs`:

- follow the API's own `totalPages` field instead of inferring the end from batch size
- raise the default `maxPages` from 3 to 300 so the walk can cover the full portfolio (246 pages today)
- skip companies whose `status` is not `Active`, probing dead or exited companies' boards only produces 404s

Verification:

- live run before: 25 companies probed, 24 errors. After: 4,180 Active companies probed
- `test-all.mjs` section 9b passes unchanged, the fixtures carry no `status` field and entries without one are kept
- `node --check` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Excluded inactive companies from YC portfolio data.
  * Improved data retrieval to continue through all available pages, preventing incomplete results.
  * Added a safeguard to prevent runaway pagination while processing YC data.
* **Documentation**
  * Clarified server-side page-size limits for YC data requests.
  * Documented the maximum pagination limit used as a safety fallback.


santifer:
@srijanarya the pagination fix is exactly what #2525 asked for and I want that part in. The third change has to go, though, and the reason is a measurement rather than an opinion.

**The blocker: the `status !== 'Active'` filter drops companies that have live job boards.**

I walked all 246 pages of the live API (6,135 companies: 4,241 Active, 1,065 Inactive, 806 Acquired, 23 Public). Your filter removes 1,894 of them. Then I probed the exact Greenhouse URL the scan would build for the 47 highest-value dropped companies (all 23 `Public` plus the 24 `Acquired` with 250+ employees): **16 of the 47 have live boards with open roles**.

So the justification in the body ("probing dead or exited companies' boards only produces 404s") does not hold: acquired and public YC companies are among the most attractive employers in the list, and a third of the ones I sampled are hiring right now. A user would never know they were filtered out, which is the part that bothers me most.

**What I'd take:** drop that hunk and merge the pagination fix on its own. If you want to avoid wasted probes on genuinely dead boards, the honest version is to keep them in and let a 404 be a 404: that is information we can measure, unlike a filter that silently removes a quarter of the list.

Split it and the first half lands today.

santifer:
Welcome to career-ops, @srijanarya. Your diagnosis was right and it was first: `per_page` is ignored server-side, so `batchSize < 1000` ended the walk on page one and the sweep only ever saw the ~25 newest companies, which are exactly the ones with no boards up yet. Following `totalPages` is the correct fix.

I made two edits on your branch rather than sending it back to you, so it lands today instead of next week. Here is exactly what I changed and why, so nothing is a surprise.

**1. Removed the `status !== 'Active'` filter.**

The reasoning in the comment (`probing their boards only 404s`) is the part I couldn't verify, so I measured it against the live API before deciding.

A stratified sample across the whole catalogue, not just the first pages:

```
Active    66.4%     Inactive  19.0%     Acquired  14.1%     Public  0.4%
→ the filter drops 33.6%, roughly 2,000 of ~6,150 companies
→ and it is not uniform: 7% of the recent pages, 61% of the oldest ones
```

Then the part that actually decides it: I probed real job boards for 22 discarded companies and, as a control, 22 `Active` ones from the same stretch of the catalogue.

```
discarded (Inactive/Acquired/Public):  3 of 22 with a live board, 58 open roles
control   (Active):                    4 of 22 with a live board, 31 open roles
```

**14% versus 18%.** A company YC marks `Inactive` or `Acquired` is hiring at very nearly the same rate as one marked `Active`, and in this sample the discarded ones offered *more* roles: a single `Inactive` company had 51 live openings on Greenhouse. `Acquired` usually means the team was bought and is still hiring under a new owner, and `Public` is companies that IPO'd, which are the largest employers in the list.

The wider point, and the reason I'd rather not have this filter at all: a scanner that silently narrows its own input is the same failure class as the bug you fixed. Both make the sweep quietly return less than it should, with nothing on screen to say so.

**2. Raised the page ceiling from 300 to 500 and named it `YC_MAX_PAGES`.**

The API reports **246** pages today. A ceiling of 300 works right now and stops working as the catalogue grows, without any error: exactly the silent truncation this PR exists to remove. The real stop condition is `totalPages`; the constant is only a runaway guard, so it should sit well clear of the real number.

Both edits verified before merging: the parser now lets `Inactive`/`Acquired`/`Public` through while still rejecting nameless entries, and a live 3-page walk returns 75 companies where `main` returned 25 and stopped.

**Credit where it belongs.** @gunjanjaswal opened #2558 for the same issue 26 hours later, and it carries two things yours doesn't: a fallback for when the API stops reporting `totalPages` at all, and a test. Yours went first, so yours lands; I'll ask them to bring those two pieces on top rather than throw the work away.

Good first contribution, and thank you for the patience. Merged 🚀

PR #2528, opened 2026-08-05
fix(providers): personio falls back to HTML scrape when the XML feed is disabled

ilyakanevskiy:
## Summary

- Some Personio tenants disable the public `/xml` jobs feed (confirmed live on three independent tenants) while their careers page is still server-rendered with the full job list present in the initial HTML.
- On a 404 from `/xml`, the provider now falls back to fetching the careers page with `?language=en` and parsing the job list from there (that query param forces English titles -- the XML feed has no language param and always renders in the tenant's default language). Non-404 errors still propagate unchanged.
- The HTML fallback matches on the stable `job-box` / `jobMetaText` class-name substrings rather than a full class match, since Personio's hashed CSS-module suffixes differ per frontend build but those marker substrings are shared across tenants.
- No creation date is exposed on the careers page, so `postedAt` is intentionally omitted on the fallback path only (still present via `<createdAt>` on the normal XML path).
- Docs: updated the `Personio` row in `docs/SUPPORTED_JOB_BOARDS.md` to mention the fallback.

## Test plan

- [x] `node test-all.mjs --only providers/personio` -- 29/29 passing
- [x] `node test-all.mjs --quick` -- full suite green

## Summary by CodeRabbit

* **New Features**
  * Added a fallback for Personio job boards when XML feeds return a 404, retrieving listings from English careers pages.
  * Improved extraction of job titles, locations, company names, and job links from HTML pages.

* **Bug Fixes**
  * Prevented duplicate listings and removed unnecessary URL query parameters.
  * Improved handling of nested HTML markup and encoded characters, including omitting unavailable posting timestamps.

* **Documentation**
  * Updated supported job-board documentation to describe Personio fallback behavior.


ilyakanevskiy:
@coderabbitai review

ilyakanevskiy:
@coderabbitai review

santifer:
Another one, @ilyakanevskiy. A tenant with `/xml` disabled currently throws a 404 and vanishes from every scan with no trace, which is the silent-loss shape we keep finding: the user sees a shorter list, not an error. I verified the fallback against a live tenant that has the feed off (`medwing.jobs.personio.de`): `main` throws `status=404`, your branch returns 13 well-formed jobs, all 13 with a location. And I cross-checked the new HTML parser against two tenants whose XML feed is healthy (`ottonova`, `everphone`): 16/16 jobs, identical URL sets, zero title or location mismatches, with `postedAt` the only documented gap. Merged 🚀

Issue #2277, opened 2026-07-29
.gitignore is not in SYSTEM_PATHS — hardened ignore rules never reach existing installs, so newly-generated personal files become committable


sagar-garg:
### Code of Conduct

- [X] I agree to follow the [Code of Conduct](https://github.com/santifer/career-ops/blob/main/CODE_OF_CONDUCT.md)

### Existing issues

- [X] I searched existing issues and this hasn't been reported yet

### What happened?

`update-system.mjs` does not ship `.gitignore` — it appears in neither `SYSTEM_PATHS` nor `USER_PATHS`:

```console
$ git show main:update-system.mjs | grep -c "'\.gitignore'"
0
```

`.gitignore` here is not a cosmetic file — it is the privacy boundary. It has been hardened repeatedly, each time in response to a real leak:

- #799 — `.playwright-mcp/` run artifacts not ignored
- #1242 — `interview-prep/sessions/*` (session transcripts hold real interviewer names and companies)
- the `data/*` blanket rule, added specifically so that *"a new generated file (e.g. a future log/cache addition) is covered for every user automatically, with no per-filename .gitignore update needed"* (comment in `main`'s `.gitignore`)
- `batch/tracker-additions/**/*.tsv` — note the `**`. The older `batch/tracker-additions/*.tsv` does not match the `merged/` subdirectory, because a single `*` does not cross a path separator.
- `*.bak` — backups written by `normalize-statuses.mjs` / `dedup-tracker.mjs`
- `writing-samples/*`, `interview-prep/*`, `data/offers/*`
- PII filename catches (`*passport*`, `*diploma*`, …)

**None of this reaches an install created before those commits landed.** The install keeps its original `.gitignore` forever, no matter how many times `update-system.mjs apply` runs.

The `data/*` blanket rule is the sharpest illustration: its stated purpose is to auto-cover future generated files, but the rule itself cannot reach the users who need it.

The result is silent and cumulative. Each new feature that writes under `data/` (e.g. `data/applications.db`, `data/portal-health.tsv` on current `main`) lands in older installs as untracked-but-unignored. It shows up as `??` in `git status` and gets swept into the next `git add -A`.

This fails in a worse direction than a normal stale-config problem:

1. **It fails unsafely.** A missing ignore rule does not error — it silently permits a commit.
2. **Forks are public by default.** A GitHub fork of a public repo starts public, so the natural setup path (fork → clone → run) can place committed personal data on the public internet, with no step in the flow that warns about it.

[...]

Issue #2269, opened 2026-07-29
tests: "Materialized skill index mode" fixture inherits the user's global core.excludesFile, so a globally-ignored .agents/ crashes the block


ringo380:
## Symptom

On v1.23.0, `node test-all.mjs` section "12c. Materialized skill index mode" crashes on any machine whose global git ignore excludes `.agents/`:

```
12c. Materialized skill index mode
The following paths are ignored by one of your .gitignore files:
.agents
hint: Use -f if you really want to add them.
  ❌ skill entrypoint index-mode test crashed: Command failed: git add -- .agents/skills/career-ops/SKILL.md
```

Both assertions in the block are lost, not just the one that throws:

- `materialized skill entrypoints stage as regular files, not symlink blobs`
- `materialized skill blobs contain canonical skill content`

## Cause

The test builds a throwaway repo (`test-all.mjs` ~4189) with `gitRun(['init'])` and stages the canonical entrypoint:

```js
gitRun(['add', '--', '.agents/skills/career-ops/SKILL.md']);
```

A fresh `git init` still honours the user's global `core.excludesFile` (`~/.config/git/ignore` on Linux/macOS). The fixture repo has no `.gitignore` and no `.git/info/exclude` override of its own, so if that global file contains `.agents/` — or `.claude/`, which the same block also depends on two lines later — `git add` refuses the path and the whole `try` aborts.

This is not exotic. Excluding `.agents/`, `.claude/`, `.cursor/` etc. globally is the standard way to keep agent-tool state out of every repo on a machine, and career-ops is the unusual case: it legitimately *ships* `.agents/skills/` and `.claude/skills/` as tracked system paths. v1.23.0 added `.agents/` to `SYSTEM_PATHS`, which is what newly exposed this — the same crash previously did not reference `.agents`.

Same root cause hits `update-system.mjs apply` in the real repo, where it surfaces as `Update commit failed (files may be staged but not committed)` after every file has been written correctly. Local workaround there is a `.git/info/exclude` with `!.agents/`, `!.agents/skills/`, `!.agents/skills/**` (and the `.claude/` equivalents), but the test fixture is created fresh each run so it cannot inherit that.

## Why it matters

[...]

PR #1997, opened 2026-07-17
fix(update-system): force-add explicitly named system files in addPaths

calebwhite-io:
## Bug

`git add` refuses explicitly named paths that match a `.gitignore` rule — even when `git checkout FETCH_HEAD --` already staged them. A clone whose `.gitignore` predates the negations for a newly shipped file (e.g. the #1242 `interview-prep/sessions` scaffold arriving in a repo that still ignores all of `interview-prep/*`) aborted the whole update commit with `Update commit failed (files may be staged but not committed)`.

## Fix

Split `addPaths()`:

- **Explicitly named file entries** are updater-owned and meant to be tracked → `git add -f`.
- **Directory entries** (trailing-slash manifest convention) stay un-forced → plain `git add dir/` keeps silently skipping ignored user content (the `*passport*` PII safety net, `.env.local`, `node_modules`) instead of dragging it into the update commit.

Reproduced with a 1.12.0 → 1.20.0 apply; verified `add -f` + the scoped commit succeed against a stale `.gitignore`, and that directory pathspecs still skip ignored files.

Note this is distinct from the `.update-dismissed` unmatched-pathspec abort fixed in #1996: `-f` only bypasses the ignore check, so both fixes are needed. The two touch adjacent code and merge cleanly in either order.

## Tests

Regression checks added to `updater-migration-tests.mjs` following its existing source-check pattern. `node test-all.mjs` passes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved update staging so explicitly included files are applied reliably, even when affected by ignore rules.
  * Prevented ignored user content within directories from being unintentionally included in updates.
  * Added safeguards covering both file and directory update scenarios.

PR #1996, opened 2026-07-17
fix(update-system): stage .update-dismissed deletion only when tracked

calebwhite-io:
## Bug

In `apply()` step 7 ("Commit the update"), if `.update-dismissed` exists it is `unlinkSync`'d and then unconditionally pushed into `pathsToStage`. But `.update-dismissed` is gitignored, so it is normally **untracked** — after deletion the pathspec matches nothing in the worktree or index, and:

```
git add -- <updated paths...> .update-dismissed
```

exits 128 with `fatal: pathspec '.update-dismissed' did not match any files`. Git validates every pathspec in one invocation before acting, so the one dead pathspec blocks staging of **all** system paths in the command and the update aborts with `Update commit failed (files may be staged but not committed)` — even though the update itself succeeded.

Reproduced with git 2.53: in a scratch repo, `git add -- somefile .update-dismissed` with the dismiss file deleted+untracked exits 128 and leaves `somefile` unstaged.

## Fix

Deleting an untracked file needs no git action. `apply()` now checks `git ls-files -- .update-dismissed` and only pushes the path into `pathsToStage` when the file is genuinely tracked (legacy clones that committed it before it was gitignored). In that legacy case the staged deletion works as before (verified: `git add` stages `D .update-dismissed`, exit 0).

Note this is distinct from the `addPaths` force-add bug (see #1997): `git add -f` only bypasses the ignore check — a deleted+untracked path still matches no pathspec, so `-f` cannot fix this one.

## Tests

Two regression checks added to `updater-migration-tests.mjs` following its existing source-check pattern (both fail on the pre-fix source):

- positive: `apply` guards the `.update-dismissed` stage on `git ls-files` tracked-ness
- negative (same style as the #1706 check): the unconditional `unlinkSync(dismissFile); pathsToStage.push('.update-dismissed')` shape must not reappear

`node test-all.mjs`: 1795 passed, 0 failed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved update handling when removing a dismissal marker, preventing update commits from failing in certain repository states.
  * Ensured updates continue successfully whether the dismissal marker is tracked or untracked.
* **Tests**
  * Added safety checks covering dismissal-marker cleanup and update commit staging.

PR #2531, opened 2026-08-05
fix(update-system): apply() cannot commit when a path it stages is ignored or never tracked

gtm-k:
## What does this PR do?

Fixes two ways `apply()` can finish an update on disk and then fail at the commit step, leaving the user to run a long `git commit` by hand. Both are in the same staging call: one is `git add` refusing a system path that a local ignore rule shadows, the other is the `.update-dismissed` marker being staged after deletion even though it was never tracked. Adds behavioural tests for both.

## Related issue

No issue — bug fix. Related but distinct: #2277 (`.gitignore` not in `SYSTEM_PATHS`) is *why* the first cause can appear on any install; this PR doesn't change that design, it stops the updater breaking when it does. #2269 is the same git behaviour hitting a test fixture rather than the updater.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

---

## Detail

Both causes leave the update uncommitted with `Update complete` never printed, but they differ in how much survives — worth stating precisely, because it changes what the user sees:

- **Cause 1** exits 1 *after* staging the paths git accepted, so the index looks almost right and only the commit is missing.
- **Cause 2** is a fatal unmatched pathspec (exit 128). git validates pathspecs before staging anything, so the whole batch dies and **nothing** is staged.

### Cause 1 — a system file shadowed by a directory-level ignore rule

`addPaths()` ran `git add -- <paths>`. git refuses an explicitly-named ignored path and exits 1.

`.gitignore` is deliberately not in `SYSTEM_PATHS`, so a user's own rule can shadow a system file at any time. The shape seen in the wild: a blanket `writing-samples/` added while hardening a checkout, over the tracked `writing-samples/README.md` that `SYSTEM_PATHS` ships. Every `apply()` then dies at step 7 of 8, and keeps dying on each later release until the ignore rule changes.

The boundary is narrower than it first looks, which is probably why this went unnoticed:

| ignore rule over a **tracked** file | `git add -- <path>` |
|---|---|
| file-level (`writing-samples/README.md`) | exit 0 — git skips ignore rules for tracked files |
| directory-level (`writing-samples/`) | **exit 1** — the match comes from the ignored directory |

The blanket form is exactly the one people reach for when protecting a directory.

### Cause 2 — the `.update-dismissed` marker (no customization required)

```js
const dismissFile = join(ROOT, '.update-dismissed');
if (existsSync(dismissFile)) {
  unlinkSync(dismissFile);
  pathsToStage.push('.update-dismissed');   // ← staged unconditionally
}
```

`.update-dismissed` is in the shipped `.gitignore` and is never tracked. `dismiss()` creates it; `apply()` deletes it and then stages it, so git gets a pathspec matching nothing. Because `pathsToStage` batches the marker with the real system paths, the fatal takes all of them down:

```console
$ git add -f -- real.txt .update-dismissed        # marker deleted, never tracked
fatal: pathspec '.update-dismissed' did not match any files    (exit 128)
$ git diff --cached --name-only                   # nothing staged at all

$ git add -f -- real.txt                          # same batch minus the marker
$ git diff --cached --name-only
real.txt
```

**Repro on a stock checkout:** `node update-system.mjs dismiss`, then `node update-system.mjs apply`. No local `.gitignore` edits involved.

`-f` does **not** rescue this one — a deleted, never-tracked path is an unmatched pathspec regardless of flags — which is why it needs its own guard rather than being folded into cause 1.

### The fix

1. `addPaths()` takes `-f`. Every path it stages is one the updater owns and just checked out from `FETCH_HEAD`; a user's ignore rule should not be able to stop the tool committing its own files. It cannot reach a user path: `pathsToStage` is built from the system manifest, and step 4 already throws `SAFETY VIOLATION` and reverts on any `USER_PATHS` hit, well before step 7 stages anything.
2. Stage `.update-dismissed` only when it is actually tracked, via a small `isTracked()` helper.

Both gain the `ctx` test seam already used by `removeAdditionsNotInHead`, so they can be driven against a throwaway repo.

### Tests

New `tests/updater-add-paths.test.mjs` — behavioural, following `tests/updater-rollback-behavior.test.mjs` (real git fixtures through the `gitIn` seam rather than source-pattern assertions). Auto-discovered, no registration.

```
🧪 Testing updater staging behavior (ignored + never-tracked paths)...
  ✅ staging succeeds when an ignore rule shadows a tracked system file
  ✅ both the shadowed path and its batch-mates reach the index
  ✅ a file-level ignore rule over a tracked path was never the problem
  ✅ a directory-level ignore rule over a tracked path stages under -f
  ✅ isTracked: true for an ignored-but-tracked path
  ✅ isTracked: false for an ignored, never-tracked path
  ✅ staging a deleted, never-tracked marker still fails (-f is no rescue)
  ✅ an unmatched pathspec strands the entire batch, not just the marker
  ✅ the same batch stages once the untracked marker is filtered out
```

Reverting just the two fix lines flips exactly two of these to failing, so they gate rather than decorate. Two cases deliberately assert a *failure* — they pin why the `isTracked` guard exists, so it doesn't get removed later on the assumption `-f` covers it.

Full suite: `2982 passed, 0 failed, 2 warnings`.

### One thing I could not verify

There is no behavioural test driving `apply()` end to end over the `USER_PATHS` violation-and-revert path — the existing coverage there is source-pattern only. That's pre-existing and this PR doesn't change it, but since the `-f` safety argument leans on that guard, it seemed worth naming rather than leaving implied.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved update staging for tracked files affected by ignore rules.
  - Correctly handles file and directory paths, including nested, deleted, stale, and non-ASCII filenames.
  - Prevents failures when `.update-dismissed` has never been tracked.
  - Improved handling of unmatched paths, literal filenames, and failed staging batches.
  - Ensures update operations recover correctly after staging errors.
- **Tests**
  - Added comprehensive coverage for update staging, path handling, batching, and error recovery.


gtm-k:
Thanks — all three land. Two are straight fixes; on the first I took the intent but changed the implementation, so the reasoning is below.

### 1. Fixture isolation — taken, with a portability change

The concern is real and I could reproduce it. Running this file with a global `core.excludesFile` that ignores `writing-samples/` — plausible, since that's the very rule that produced the bug this PR fixes:

```
  WITH isolation      pass=10 fail=0
  WITHOUT isolation   pass=2  fail=0   ← crashed after 2 assertions
```

Note the shape: `fail=0`. It doesn't report a failure, it dies mid-file with `nothing to commit (create/copy files and use "git add" to track)`, because `git add -A` staged nothing. A contributor would get a confusing git error rather than a test result. That's the same signature as #2269.

I couldn't apply the suggested diff as written, though — `/dev/null` isn't portable here:

```console
$ git config core.excludesFile /dev/null
$ git config core.excludesFile
nul
$ git add -A
fatal: cannot use nul as an exclude file        # exit 128
```

Git for Windows stores `/dev/null` as `nul` and then refuses it, so any command that has to classify untracked files dies — `git add -A` and `git status` both exit 128. That's the first thing `makeRepo()`'s callers do, so the committable suggestion would have broken the suite outright on the platform this PR was developed and tested on. Same applies to `core.hooksPath`.

So I kept all three settings and pointed the two path-valued ones at an empty file and an empty directory inside the fixture's own `.git/`, which is portable and verified to neutralize a hostile global config (the run above). `commit.gpgsign=false` is applied unchanged.

Worth noting for whoever tunes the bot: the "Prompt for AI Agents" block on this comment is implementation-agnostic — *"override repository-local Git settings that can leak from global configuration, including commit.gpgsign, core.hooksPath, and core.excludesFile … ensure commits and ignore resolution use only the fixture repository's controlled configuration."* That spec is satisfiable portably, and what I did satisfies it. The committable suggestion pinned it to `/dev/null`, which contradicts the spec's own goal on Windows, since commits stop working entirely. An agent applying the suggestion verbatim would have shipped a broken suite.

One thing worth flagging beyond this PR: `tests/updater-rollback-behavior.test.mjs` builds its fixture the same way and has the same gap. I've left it alone to keep this PR scoped, but happy to send a follow-up that factors the isolation into a shared helper — that would close the fixture half of #2269 across the suite rather than one file at a time. Say the word if you'd like it as part of this PR instead.

### 2. The file-level assertion was vacuous — correct, fixed

You're right, and it's the same mistake I'd already corrected once elsewhere in this file. Asserting through `addPaths` meant asserting under `-f`, where both cases stage fine, so it could never detect the boundary moving. It now probes with a plain `git add` through the raw runner, and I added your companion assertion so the pair actually characterises the boundary:

```
✅ a file-level ignore rule over a tracked path was never the problem
✅ a directory-level rule blocks a plain add — this is why -f is required
```

### 3. Unguarded recovery call — correct, fixed

Wrapped like its neighbours. The point isn't that it was likely to throw, it's that a regression would have surfaced as an unhandled exception instead of a reported failure, and the `rmSync` cleanup would have been skipped.

---

Full suite after all three: `2983 passed, 0 failed`. The reverted-fix check still discriminates — dropping `-f` and the `isTracked` guard flips the same assertions to failing.

santifer:
@gtm-k the diagnosis is right and cause 2 is exactly scoped, but cause 1's fix has to change before this can land, and the reason is one you couldn't see from the diff.

**The blocker: `git add -f` on directory pathspecs stages ignored content.**

The change turns `git add -- ` into `git add -f -- `, and `pathsToStage` includes **53 directory pathspecs** out of 283 SYSTEM_PATHS entries (`docs/`, `tests/`, `templates/`, `.github/`, `plugins/`, `modes/*/`, and more). `git add -f -- dir/` stages every ignored file underneath it; plain `git add -- dir/` skips them. `apply()` then commits `pathsToStage`, so whatever got swept in is committed.

That is a data-contract problem, not a style one: the user layer is gitignored precisely so the updater cannot touch it, and a blanket `-f` over directories removes that protection on every update.

**What I'd take instead:** scope the force to the specific file paths that need it (the ones the ignore rules would otherwise skip), and leave directory pathspecs on plain `add`. If you need the force to be dynamic, deriving it from the same USER_PATHS source keeps it honest when a new user file appears.

Cause 2 is fine as-is, and if you'd rather split this into two PRs so that half can land today, I'd merge it immediately.

gtm-k:
Confirmed, and the count is exact — 53 of 283 at this PR's head. (It's 53 of 284 on current `main`; the extra entry is `.gitattributes`.) Reproduced it, and found a payload sitting in my own checkout.

Taking the split: **cause 2 is up as #2591** and ready to merge. Cause 1 stays here and I'll rebuild it — one question below on which way you want it.

One thing you should have in front of you first. Both causes already have open PRs from @calebwhite-io, which I didn't find until after building the same fixes: **#1996 is cause 2**, and **#1997 is the file/directory split you're describing here** — both opened 2026-07-17, and #1997's comment names the same hazards I hit (`*passport*`, `.env.local`, `node_modules`). You pinged both on 07-27 asking for a rebase.

Neither has moved since: no pushes since 07-17, both still `CONFLICTING`, no reply from the author. So I've opened the cause 2 PR rather than let it keep waiting, with credit and a cross-link to #1996 in the body, and left a note there — if @calebwhite-io comes back and rebases, I'll close mine.

### The blocker reproduces

`docs/README.md` tracked, `docs/secret.txt` untracked and ignored:

```console
$ git add -- docs/
$ git diff --cached --name-only
docs/README.md

$ git reset -q
$ git add -f -- docs/
$ git diff --cached --name-only
docs/README.md
docs/secret.txt
```

And what `apply()` would have committed here on the next run:

```console
$ git status --porcelain --ignored=matching -- <the 53 dirs>
!! dashboard/career-dashboard
```

A compiled binary. `scaffolder/` is also a `SYSTEM_PATHS` directory and ships a `package.json`, so `scaffolder/node_modules/` is the same shape on any checkout where it's been installed.

### The question: which way do you want cause 1 built?

Your remedy is right and #1997 already implements it — leaving directory pathspecs on plain `add` closes the sweep completely, and the transcript above is exactly that difference. If you want that, it's already written and I'll rebuild this branch to match it.

There's one case it leaves open, which is the only reason I'd suggest anything else. Directory pathspecs aren't immune to cause 1 either — a directory pathspec whose directory is itself ignored fails the plain add too:

```console
$ cat .gitignore
docs/secret.txt
sub/
$ git ls-files -- sub/
sub/keep.md
$ git add -- sub/
The following paths are ignored by one of your .gitignore files:
sub
hint: Use -f if you really want to add them.
hint: Turn this message off by running
hint: "git config advice.addIgnoredFile false"
$ echo $?
1
```

Narrow — it needs someone to ignore a whole system directory — but it's the same failure cause 1 is about.

Asking git for the force list handles that case too, without a manifest to keep in sync:

```console
$ git ls-files -c -i --exclude-standard -- docs/ sub/
sub/keep.md
```

Tracked **and** ignored, and by construction already in the index, so forcing them cannot introduce content. `docs/secret.txt` is correctly absent. `ls-files` returns files, never directory pathspecs, so its output is inherently the exact-file force list.

Three caveats so I'm not overselling it.

It's a safe file-only list, not a *minimal* one — it also picks up tracked files matched by file-level rules, which plain `add` would have accepted anyway. Harmless, just not the smallest set.

The `ls-files` call alone isn't the whole fix: an ignored directory pathspec still has to be dropped from the plain `add`, or that add fails exactly as above. So it's two steps, not one:

```js
const forced = lsFiles('-c', '-i', '--exclude-standard', '--', ...paths);   // tracked ∧ ignored
const plain  = paths.filter(p => !isIgnored(p));   // drop pathspecs that are themselves ignored;
                                                   // their tracked contents are already in `forced`
if (plain.length)  git('add', '--', ...plain);
if (forced.length) git('add', '-f', '--', ...forced);
```

A partially-ignored directory still works — `docs/` stays in the plain add and stages its non-ignored tracked changes, while anything under it that's ignored-but-tracked arrives via `forced`.

And materialized skill entrypoints need handling separately, which I missed first time round. `prepareMaterializedSkillEntrypointsForStage()` runs `git rm --cached` on mode-`120000` entries immediately before `addPaths`, so they aren't in the index when `ls-files -c -i` runs and can't appear in `forced`. If their directory is also ignored they'd be dropped from the plain add too, and never staged at all.

Computing the candidate list earlier doesn't rescue that on its own — `ensureSkillEntrypoints` can `writeFileSync` an entrypoint that never existed, and a brand-new file is in no `ls-files` output at any point. So they need forcing by name. Worth flagging that "updater-owned" isn't quite exact either: the helper also adopts a *pre-existing* regular file whose content already matches the pointer, rewrites it, and returns its path — narrow, since it only touches files that already contain that exact text, but the returned paths are updater-**written** rather than updater-created.

On the other half of your suggestion — deriving it from `USER_PATHS` — that wouldn't have protected the payloads above. `dashboard/career-dashboard` isn't a `USER_PATHS` entry, and neither is `node_modules/` or a local `.env`. The user layer is what `USER_PATHS` is *for*; it just isn't the set that gets swept.

Either way, tell me which and I'll have it up shortly. I'd lean tracked-only for the extra case, but the split is the smaller change and you know this codebase better than I do.

### Both existing PRs are gated by source-pattern regexes only

#1996 and #1997 each assert that a string appears (or doesn't) in `update-system.mjs`. Those move whenever the source text moves, and some broken shapes slip through.

`upgrade-tests.mjs` already clones an old tag, points it at a bare mirror, and runs a real `node update-system.mjs apply` — so one seeded line in the fixture install makes every PR-gate leg drive the real path, with no new assertion needed. I verified it discriminates by reverting the guard as a commit. Relevant excerpt:

```
FAIL [career-ops-v1.25.0] apply exits 0 (got 1)
FAIL [career-ops-v1.25.0] upgrade executed: update-system.mjs blob matches target (non-vacuity oracle)
RED: 2 failure(s)
```

It has to be a commit, not a worktree edit — `prGate()` starts with `git rev-parse HEAD`, so uncommitted changes are invisible to it and it reports green regardless. Details are in my #1996 comment.

### One more thing that may shrink the problem

As `revertPaths`'s existing comment notes, `git checkout FETCH_HEAD -- <path>` already puts the new content in the index — so step 3 has staged the system paths before `addPaths` runs at all:

```console
$ git checkout main -- docs/
$ git diff --cached --name-only
docs/README.md
```

Which means `addPaths` isn't what puts the update content in the index — it's mainly covering the pruned deletions and the materialized entrypoints. That may shrink how much needs forcing in the first place.

### Correction to my own description

I wrote that cause 2's fatal pathspec means "nothing is staged." True of the console demo, which starts from a clean index; false in production, for the reason just above. What's actually lost is the commit, plus the pruned deletions and entrypoints. Same symptom, different mechanism.

So: cause 2 is ready whenever you want it. Tell me split-or-tracked-only for cause 1 and I'll push it here.

santifer:
Sorry for the four-day gap, @gtm-k. This deserved a faster answer than it got, especially since you asked a direct question and then did the work of finding the two PRs that overlap yours.

**On split-or-tracked-only: I think it is neither, and the reason is three lines above the bug.**

```js
// update-system.mjs:1185
const pathsToStage = [...updated];
```

`updated` is the list of files this run just wrote. The updater already knows, exactly, which paths are its own. So the force-add list does not need deriving from the repo at all: not from `ls-files -c -i`, not from tracked-vs-untracked, not from `USER_PATHS`. It is `updated`, plus `.update-dismissed`, plus the materialized entrypoints you already identified.

That matters beyond tidiness. Every derivation from disk state reads the *user's* repo to decide what to force into a commit, and this project's hard rule is that the system layer never touches the user layer. `ls-files -c -i` is the widest possible reading of someone's checkout, and it is how a stray payload ends up staged: exactly the 53-of-283 you found. A list built from "what did I write in this process" cannot have that failure mode, because a file the updater never wrote is never a candidate.

It also fixes the case you flagged as unrescuable. A brand-new entrypoint appears in no `ls-files` output at any point, but it *is* in `materializedSkillEntrypoints`, in memory, at the moment you need it. Same for the pre-existing regular file that `ensureSkillEntrypoints` adopts and rewrites: you called `updater-owned` "not quite exact" and you are right, but `updater-written` is exact, and written is the property that matters for staging.

So: `addPaths(paths)` force-adds the whole list, and the list is assembled from what the run produced. `-f` on a non-ignored path is a no-op, so there is no branching and no second code path to keep in sync.

**One boundary to keep.** Force-adding must stay per explicit filename, never a directory or glob. `docs/` as a pathspec plus `-f` would sweep whatever a user keeps under it. Your split already had this right; I am only saying it out loud so it survives review.

**Your correction to your own description is the reason I trust the rest of it.** You wrote that cause 2 means "nothing is staged", found that `git checkout FETCH_HEAD -- <path>` already stages the content in production, and came back to say the symptom was right and the mechanism wrong. Nobody asked you to. That is the difference between a report and an investigation.

**On the overlap with #1996 and #1997.** You found @calebwhite-io's PRs after building the same fixes and said so unprompted, which is worth more here than being first. I am not deciding that one in a comment: three PRs from two people on one file, all conflicting, and the person who got there first has been waiting since July. I will work it out with both of you rather than quietly picking a winner, and I will start that thread this week.

What I can say now is that your point about the gate is correct and independent of who lands: asserting that a string appears in `update-system.mjs` pins the source text, not the behaviour, and moves whenever the text moves. `upgrade-tests.mjs` already runs a real `apply` against a cloned tag, and your seeded-line approach makes every PR-gate leg exercise the real path. Whichever fix lands should be gated that way, and your reverted-guard run showing `RED: 2 failure(s)` is the proof it discriminates. The detail that it has to be a commit because `prGate()` starts at `git rev-parse HEAD` is the kind of thing that costs someone an afternoon; thank you for writing it down.

Push cause 1 here whenever you like, built off `updated` rather than the index. #2591 I will handle as part of the same thread, not separately.

santifer:
Addendum with the executable evidence, @gtm-k, because I said "force-add must stay per filename, never a directory" and you deserve the demonstration rather than the assertion. I ran it just now:

```console
$ cat .gitignore
career-dashboard
.DS_Store
*.env

$ git status --porcelain          # what a safety check based on status sees
                                  # (nothing — ignored files are not listed)

$ git add -f -- dash/ docs/
$ git diff --cached --name-only
dash/career-dashboard
docs/.DS_Store
docs/prod.env
```

Three things that matter here.

**1. `pathsToStage` carries directory entries.** `SYSTEM_PATHS` has 53 of them, so `-f` on that list is `-f` on directories, and everything ignored underneath comes with it. In a real checkout that is the 4.2 MB `dashboard/career-dashboard` binary, and `apply()` step 6 (`rebuildDashboardBinaryIfNeeded` → `go build -o career-dashboard .`) creates it *immediately before* step 7 stages. So the update would build the binary and then commit it, every time.

**2. The safety check cannot see it.** `git status --porcelain` does not list ignored files, which is why the output above is empty while the staged set is not. Any guard that scans status before staging is structurally blind to exactly the files `-f` adds. That is worse than a missing check, because the check is there and it reports clean.

**3. The patterns are unanchored,** so this is not only about `dashboard/`: `.DS_Store`, `*.bak*`, `*.mov` and `.env` at any depth across all 53 directories. A user's `.env` landing in a commit is a different order of problem from a stray binary.

None of this is an argument against your PR. It is the argument for the version you are already rebuilding, and for building the list from `updated` (what this run wrote) rather than from anything on disk: a file the updater never wrote is never a candidate, and a `-f` on an explicit filename cannot sweep a sibling.

Same conclusion I reached reading the code, now with the console output behind it.

gtm-k:
Pushed. I took the remedy: build the staging list from what the run produced. But I couldn't build it from `updated` as written, because `updated` isn't the shape the instruction assumes — and your addendum is the half that matches the data.

### `updated` holds pathspecs, not files

It collects whichever manifest entries checked out successfully. The files that checkout wrote are never recorded:

```js
// update-system.mjs
     const updatePaths = mergePathLists(SYSTEM_PATHS, remoteSystemPaths, BOOTSTRAP_PATHS);
     for (const path of updatePaths) {
         gitQuiet('checkout', 'FETCH_HEAD', '--', path);
         updated.push(path);          // the manifest entry itself
     const pathsToStage = [...updated];
```

Only the prune step and the materialized entrypoints ever push a filename. Everything else is the manifest verbatim — the 53-of-283 from your addendum.

So the two comments point opposite ways. "`addPaths(paths)` force-adds the whole list … there is no branching" and "force-adding must stay per explicit filename, never a directory or glob" can't both hold, because the whole list *is* 53 directories. Force-adding it unchanged is the diff you blocked on the 7th, and the addendum is the comment that's right about it.

### The principle holds; the granularity was the problem

The part of your argument that does the work comes down to *whose tree you read*. `ls-files`/`status` read the user's checkout to decide what to force into a commit. `FETCH_HEAD` cannot: every name in it is a file upstream ships. That's the same guarantee you were reaching for, and it's available at file granularity, which `updated` never had.

```console
$ git status --porcelain          # the safety check cannot see them
 M docs/README.md

$ git add -f -- docs/ dashboard/  # the previous revision
$ git diff --cached --name-only
dashboard/career-dashboard
docs/README.md
docs/prod.env

$ git ls-tree -r --name-only FETCH_HEAD -- docs/ dashboard/   # the fix
dashboard/main.go
docs/README.md
$ git add -f -- <that list>
$ git diff --cached --name-only
docs/README.md
```

Pass-through is load-bearing for the two cases that aren't in the target tree: a pruned deletion is by definition absent from it, and a materialized entrypoint may have just been `git rm --cached`ed by `prepareMaterializedSkillEntrypointsForStage`. Both can only be restaged by name.

### Expanding to filenames turned out to be necessary but not sufficient

Three things had to go with it. I found the first two by attacking my own diff; neither of them ever showed up as a test failure.

**Filenames are still pathspecs.** `--` ends option parsing, and does nothing about pattern interpretation. A shipped file named `docs/[x].md` is read as a glob:

```console
$ git add -f -- 'docs/[x].md'
$ git diff --cached --name-only
docs/x.md                      # the ignored sibling, not the file named
```

All three git calls now take `--literal-pathspecs`. No such name exists in the tree today; nothing enforced that.

**Neither catch had a legitimate case.** `ls-tree -- absent/` and `ls-files -- untracked` both exit 0 with empty output, so the stale-manifest-entry case the catch was written for never throws. What a throw actually means is an unreadable ref, a timeout, or a corrupt object store — and swallowing it reported "this directory ships nothing", dropped every file under it from staging, and let `apply()` continue toward its success banner. Both are uncaught now.

**The add has to be batched.** Expanding multiplies the pathspec count against a 32,767-character Windows command line:

```
manifest entries      : 283
expanded file count   : 817
argv chars (joined)   : 22152
headroom              : 10615
```

As one call this lands at two-thirds of the ceiling the day it merges and grows every release. The expansion is what forces it, which is why it shares the commit. Say the word if you want it split out.

### What this does not fix, and one correction to myself

I had written — in a comment in the code and in a draft of this reply — that keeping the unexpanded list on `git commit` was safe, because a commit only records what is tracked or staged. That is wrong. With path arguments git ignores the prepared index and records the current contents of every *known* file matching the pathspec:

```console
$ git add -f -- docs/README.md        # the fix stages exactly one file
$ git diff --cached --name-only
docs/README.md
docs/prod.env

$ git commit -m "update" -- docs/     # but the commit still takes the directory
$ git show --name-only --format="" HEAD
docs/README.md
docs/USER-NOTES.md
docs/prod.env

$ git show HEAD:docs/USER-NOTES.md
user edited this, never staged
```

So the sweep has a second door. `git commit -- docs/` can commit a user's unstaged edit to a tracked file under a system directory, and it can commit an ignored file that an earlier buggy run stranded in the index — which is to say the fixed updater would launder the broken one's payload into a commit. Neither is caught by the step-4 validation, which skips pre-existing dirty paths.

This predates the PR, so I've left the commit call alone and documented it where it happens instead of quietly widening the diff again. Happy to take it here or in a follow-up, whichever you prefer.

One more in the same area, also pre-existing: the commit-failure detector compares `gitStatusEntries()` paths against `allTargetPaths` with an exact `Set.has()`. Status yields `docs/foo.md`; `allTargetPaths` holds `docs/`. If only files beneath directory entries changed and the commit fails, the match returns false and `apply()` treats it as "nothing to commit" and continues to the success path.

### Verification

`3450 passed, 0 failed`. `upgrade-tests.mjs --pr-gate` GREEN: a real `apply()` against the cloned v1.25.0 tag leaves every user file byte-identical.

Every guard carries its own oracle and was mutation-tested on its own, after finding that two of my earlier assertions passed under the mutation they existed to catch. The sweep test first asserts that a bare `-f` over `docs/` *does* stage the ignored binary and the `.env`; dropping `--literal-pathspecs` re-sweeps the sibling; restoring either catch re-absorbs a bad ref. Neutering the expansion:

```
❌ expansion did not contain the sweep: docs/README.md, docs/career-dashboard, docs/prod.env
❌ expansion returned a directory: modes/
❌ path came back quoted or mangled: ["modes/"]
```

Rebased onto current `main` while I was in here.

On #1996/#1997 — understood, and no rush from me. @calebwhite-io got there first and I'd rather the thread happen than have this land ahead of it.

gtm-k:
The `rollback()` half is right and it's worse than "also affected" — it's a regression this PR introduced, which I missed twice: once when writing it, once when reviewing my own diff for exactly this class of problem.

On `origin/main` `addPaths` is a plain `git add`, so `rollback()` handing it 53 directory entries was harmless — a plain add skips ignored paths. Making `-f` unconditional in the first commit armed the sweep at **both** call sites, and the second commit expanded only `apply()`'s. A rollback would have committed the user's ignored files while reporting that it restored them.

Fixed in `793df017`, through the same expansion, against the backup tree:

```js
if (restored.length > 0) addPaths(expandToShippedFiles(restored, latest));
```

The ref is the part worth stating: rollback restores what the *backup* shipped, which is not what `HEAD` ships. A file the backup carried and upstream has since retired is absent from `HEAD`, and expanding against the wrong ref would silently fail to restore it. The test asserts that case directly.

`removed` is deliberately left unexpanded — those paths were already staged as deletions by the `git rm` above, and they don't exist in the backup tree to expand against.

### The more useful half

A per-call-site fix wasn't enough, because nothing would have caught the *next* one. `addPaths` now rejects a directory pathspec outright instead of documenting that it must not receive one:

```
addPaths received directory pathspec(s), which -f would sweep ignored files from:
docs/. Resolve them with expandToShippedFiles() first.
```

I'd written "callers must pass files, not directory pathspecs" into the docblock while a caller was passing directories. That is the precondition-without-enforcement shape, and the failure mode it produces is a user's ignored files committed silently, which nothing downstream reports.

It also fixes a hole in my first attempt at the test. `rollback()` is not exported and no harness drives it, so my rollback test reconstructed the call and exercised the helpers directly — which passed with the production bug still in place. The mutation run is what showed it: reverting the fix changed nothing the suite observed. Refusing directories at the boundary closes the site without needing a test per caller.

One limit of the guard, stated in the code: it detects a directory by trailing slash, matching how `SYSTEM_PATHS` spells one. `git add -f -- docs` sweeps identically, and `rollback()` builds its `removed` list in exactly that slash-stripped form three lines above. That list only reaches `git commit` today; routing it through `addPaths` would pass the guard.

### On the other two sites in your comment

Both were already in my previous comment with reproductions, and both are deliberately not fixed here.

`git commit -- <dir>` recording unstaged tracked edits and index-stranded ignored files is real and I reproduced it independently. It predates this PR, so it's flagged in place and waiting on @santifer's scope call. He has already blocked this diff once for scope; widening it unasked would repeat that.

The failure detector comparing `gitStatusEntries()` file paths against `allTargetPaths` directory entries with an exact `Set.has()` — same status, same reason.

### Coverage

Added `rollback` coverage to `tests/updater-add-paths.test.mjs`. Real `apply()` coverage already exists in `upgrade-tests.mjs --pr-gate`, which clones a v1.25.0 tag and runs the actual binary against it; that leg is GREEN and asserts every user file is byte-identical afterwards.

Every assertion here carries an oracle and was mutation-tested on its own, after two of my earlier ones passed under the mutation they existed to catch. Test 6's oracle moved to the raw runner, since asserting the sweep through `addPaths` would now only re-test the refusal. The refusal assertion checks an empty index as well as the throw — a guard that swept and *then* threw would otherwise have passed it.

`3453 passed, 0 failed.` Upgrade regression gate GREEN.

Unrelated to the code: the four workflows on this PR are sitting at `action_required` and need the fork-approval click. `label` and `CodeRabbit` run without it, so the check list reads greener than it is.

santifer:
Coordination time, @gtm-k and @calebwhite-io: four PRs are now working the same staging seam of update-system.mjs, and I would rather sequence them in the open than let merge order pick winners.

The map as I see it:

- #1996 (caleb) and #2591 (gtm-k) fix the same dismiss-marker bug from two sides. Caleb diagnosed it first, back in July; gtm-k hit it independently and flagged the overlap himself in #1996, which is exactly how this should work.
- #2531 (this PR) is the general guard: apply() refusing to stage ignored or never-tracked paths, now with per-file force-adds instead of directory pathspecs. Verified today in an isolated worktree: full suite green, and the directory-pathspec repro from my addendum no longer stages the binary, the .DS_Store, or the env file.
- #1997 (caleb) force-adds explicitly named system files: same seam, likely subsumed in part by the per-file approach here.

Proposed sequence, open to objection from either of you:

1. #2531 lands first as the base guard: it is the only conflict-free one and it generalizes the class.
2. @gtm-k: after that, say whether #2591 still adds anything the guard does not cover. If not, closing it as folded-in is honorable, not a loss.
3. @calebwhite-io: #1996 and #1997 rebase on top; whatever the guard already covers drops out, and the dismiss-marker diagnosis stays credited to you in the changelog line regardless, because you found it first.

Caleb, you have been quiet for 16 days, so to be fair about it: if I hear nothing in a week I move ahead with this sequence and leave your two open for a rebase whenever you return. Nothing of yours gets closed over silence.

Objections or better orderings welcome: this is a proposal, not a verdict.

gtm-k:
Sequence works for me, and thanks for putting it in the open — @calebwhite-io waiting since July while three of us work the same seam is exactly the situation that needs saying out loud.

### On #2591: fold it in, do not close it

The short answer to your question is that #2591 still adds two things, so I have ported both here and #2591 can close as folded-in once this lands. One of them was a bug on this branch.

**The probe/unlink order, which this branch had backwards.** #2591 probes before unlinking; #2531 unlinked first:

```js
// #2591 — right
const dismissMarkerTracked = isTracked('.update-dismissed');
unlinkSync(dismissFile);
if (dismissMarkerTracked) pathsToStage.push('.update-dismissed');

// #2531 — what was here
unlinkSync(dismissFile);
if (isTracked('.update-dismissed')) pathsToStage.push('.update-dismissed');
```

Identical on the happy path, since `isTracked` reads the index and a worktree deletion does not touch it. The difference is the failure path, and I armed it myself two commits ago: I made `isTracked` propagate instead of catching, because `ls-files` exits 0 for an untracked path and so a throw is always an abnormal failure. That is right on its own, and it made the ordering matter. Unlinking first deletes the marker, throws, and then finds `existsSync` false on the retry — skipping a deletion the commit still owed, leaving the worktree dirty after an update that printed success. Probing first leaves the marker on disk so the retry re-enters the block.

**The two-leg PR gate**, which is the one I would not want dropped, since it is the gate you asked for. `apply()` branches on trackedness and each branch fails differently, so one leg leaves the other ungated — dropping the tracked-marker push is invisible to the untracked leg, which skips that push anyway.

It needed a hand merge with #2697, which landed in between and added the three-state SKIPPED result. Composed so the outcome derives from whether any leg actually ran; deriving it from the leg count would let a future third leg turn a skip into a green:

```
PASS [career-ops-v1.25.0 dismiss=untracked] dismiss marker removed from disk
PASS [career-ops-v1.25.0 dismiss=tracked]   dismiss marker removed from disk
PASS [career-ops-v1.25.0 dismiss=tracked]   tracked dismiss marker: its deletion reached the update commit
PASS [career-ops-v1.25.0 dismiss=tracked]   tracked dismiss marker: no deletion left behind (git status: clean)
GREEN
```

The on-disk assertion covers both legs because it is the only check the untracked leg can make: an untracked marker appears in neither `HEAD` nor `git status`, so one left behind would silently re-suppress every later update check.

`tests/updater-is-tracked.test.mjs` came across with them. The diagnosis stays @calebwhite-io's from #1996 — I hit it independently and it was already his.

### The guard also got the same treatment it was written to enforce

CodeRabbit caught that my directory guard checked for a trailing slash, which is the spelling `SYSTEM_PATHS` uses and not the property that matters. I had written the limitation into a comment, which is precisely the precondition-without-enforcement shape the guard exists to replace, so that was a fair hit.

It now asks the filesystem, and the question collapses to one `lstat`: a path that is not on disk cannot sweep anything, because `git add -f` on an absent path can only stage deletions of entries already in the index, and an ignored file cannot be one. No git call needed.

That incidentally settles three cases an index-based check gets wrong, none of which showed up as a failing test — a directory replaced by a regular file of the same name (stale index entries below it read as a directory and refuse legitimate work), a non-canonical spelling like `./docs` (ls-files answers canonically, so a prefix comparison never matches), and an untracked directory such as `node_modules` (no index entries at all, fully sweepable).

Validation also moved out of the batch loop. Inside it, a directory in a late batch was caught only after earlier batches had already been staged, so the refusal reported a problem it had partly committed to. Restoring that now fails with 163 paths staged before the throw.

### Still open, still yours to scope

The two pre-existing ones from my earlier comment are untouched and waiting on your call: `git commit -- <dir>` recording unstaged tracked edits and index-stranded ignored files, and the failure detector comparing status file paths against directory entries with an exact `Set.has()`. Both have reproductions above. Say the word and they go in here or in a follow-up.

`3462 passed, 0 failed.` Upgrade gate GREEN on both legs.

The four real workflows are still at `action_required` and need the fork-approval click — `label` and `CodeRabbit` run without it, so the check list reads greener than it is.

santifer:
Sorry for the wait, @gtm-k — and this one has become considerably more relevant since you filed it.

`apply()` failing to commit when a staged path is ignored is the same seam that produced two user reports tonight: #2881 (the updater flagging 285 files and cascading `.bak` → `.bak.bak`) and #2756/#2857 (ignore rules never reaching existing installs). Three PRs and two reports now converge on how `apply` interacts with `.gitignore`, and yours is the oldest of them.

That means the updater work needs **sequencing rather than parallel review**, and I'd rather say that than merge them one at a time and hand each author a conflict:

- #2820 — CRLF false-positives (**merged tonight**)
- #2531 — this one: staged-but-ignored path blocks the commit
- #2857 — reconcile `.gitignore` on apply so new rules reach existing installs
- #2864, #2866, #2868 — drift pruning, confirmation gate, and one already superseded by #2820

Yours and #2857 are the pair that interact most directly: if ignore rules start reconciling on apply, the set of "staged but ignored" paths changes underneath your fix. Worth the two of you knowing about each other before either lands.

Not merged tonight (per-session merge ceiling). Queued in that order, and I'll be doing the ordering rather than leaving it to whoever rebases first.

luochen211:
Maintainer follow-up: this PR is still behind current main, so please rebase or merge the latest main into the contributor branch before the next review pass. The remaining scope is now clear: keep the per-file staging guard, cover slash-free directory pathspecs such as `docs`, and preserve the tracked/untracked dismiss-marker behavior. The two pre-existing concerns called out above (`git commit -- <dir>` capturing unstaged/index-stranded files and the directory-vs-file status mismatch) should either be fixed in this PR or explicitly split into a follow-up with an issue link. Please update the branch and report the resulting CI status.

santifer:
Noted, and that follow-up is mine to act on rather than yours, @luochen211 — the branch being behind main is a consequence of the merges I've been landing, not something you did.

Don't rebase yet, though. As I said above, the updater PRs need **sequencing** (#2820 merged, then this one, #2857, #2864, #2866; #2868 is already superseded), and rebasing now means rebasing again after whichever lands before you. I'd rather ask you once, at the right moment, than have you do it twice.

I'll ping you when yours is next in line.

gtm-k:
Holding off on the rebase per your call, @santifer — happy to do it whenever you say this one is next.

Answering the scope list against the current head in the meantime, since three of the four items are already there and the one red check is not what it looks like.

### The macOS red is not this PR

`test (macos-latest)` fails on `agent-inbox-tests.mjs` section 7:

```
❌ all 30 concurrently queued items survive — kept=26 of 30
📊 Results: 3608 passed, 1 failed
```

That is #2777 / #2824 — concurrent `add` losing an item under 30-way lock contention. Three commits fixed it on `main` and all three land after this branch's base:

```console
$ git log --oneline 07a2616c..origin/main -- agent-inbox.mjs agent-inbox-tests.mjs
6d891a28 test(agent-inbox): print the failing child's error when a concurrent add is lost (#2834)
513f07a2 fix(agent-inbox): give concurrent add() a burst-sized lock timeout (#2825)
a7f65b6a fix(agent-inbox): serialize concurrent adds through the shared lock (#2777)

$ git rev-list --count 07a2616c..origin/main
96
```

This branch touches `update-system.mjs`, `upgrade-tests.mjs`, `tests/updater-add-paths.test.mjs`, `tests/updater-is-tracked.test.mjs` and now `tests/helpers.mjs` — no overlap with `agent-inbox`. The fix is on `main` and this branch predates it, so the rebase should clear it; I would not promise more than that about an intermittent one, and it can obviously still flake elsewhere. Either way it is not a verdict on this diff.

### Slash-free directory pathspecs — already in, and wider than `docs`

This was the CodeRabbit finding I took earlier in the branch, so it is pinned at head rather than pending. `addPaths` asks the filesystem (one `lstat`) instead of matching a trailing slash, and the cases are in `tests/updater-add-paths.test.mjs`:

- **12b** — `addPaths(['docs'])`, no trailing slash: refused, `staged.size === 0`. The converse is pinned in the same block, or the guard could pass by rejecting everything.
- **12c** — `./docs` and `docs/.` refused (non-canonical spellings, which a `ls-files` prefix comparison never matches); `node_modules` refused (an untracked directory has no index entries at all, yet `-f` sweeps it); and a directory upstream replaced with a regular file still stages, because stale index entries below it would make an index probe reject legitimate work.
- **12d** — a directory in a *late* batch leaves the index completely untouched. Validation runs over the whole list before any batch flushes; inside the loop, the refusal reported a problem it had partly committed to.

12b's converse had a gap I should own, since it is the exact shape this PR is otherwise about (`7e219d0b`). Its comment claimed "a tracked file, a staged deletion, and a brand-new file must all pass through" and the code passed only two of those — no deletion. A precondition asserted in a comment and not enforced, in the block whose whole point is that comments do not enforce. The deletion is now in the list, and it is the case that discriminates: it is the only one whose path does **not** exist on disk when the guard `lstat`s it, so a guard that refused "missing" rather than "is a directory" would pass the other two. Mutated `rejectDirectories` to return `true` from the `catch` to confirm that — the new case fails naming `docs/RETIRED.md`, and the two pre-existing paths do not.

Per-file staging guard: kept. Tracked/untracked dismiss-marker behaviour: preserved, and the gate covers both legs end to end (`dismiss=untracked` and `dismiss=tracked`). The direct `isTracked` assertions are not redundant with that, but the reason is narrower than "the gate cannot reach it": the gate asserts on the outcome (a deletion reached the commit, worktree clean), so a probe that returned the right answer for the wrong reason still reads green there. That distinction is also what made the header comment in `updater-is-tracked.test.mjs` stale — it still said the PR gate *cannot* reach the tracked case, which stopped being true the moment I folded your gate into this branch. Corrected in the same commit.

### New commits: one isolation fixture instead of two (`b9aae5aa`, `7e219d0b`)

CodeRabbit caught that the `makeRepo` I added in `updater-is-tracked` is a copy of the one in `updater-add-paths`, differing only in the mkdtemp prefix and whether `root` reaches the ctx. Now `makeUpdaterRepo(gitIn, {prefix, includeRoot})` in `tests/helpers.mjs`. `gitIn` is injected rather than imported, because `helpers.mjs` depends on nothing but Node builtins today and 57 of the 62 suites import it — pulling in `update-system.mjs` to serve one fixture would put it on the load path of most of the suite in exchange for one caller's convenience.

The claim a shared body makes is "dropping a pin reddens both suites", so I mutated it rather than asserting it. Two false readings on the way, both of which looked like results:

**Vacuity.** On a clean checkout the pins do nothing, so dropping one changes no outcome and the mutation passes while testing nothing. Each pin has to be dropped under a `GIT_CONFIG_GLOBAL` it exists to neutralise. My first attempt wrote MSYS `/tmp/...` paths into that config, which git-for-Windows ignores for path-valued settings (`warning: encountered old-style … should be %(prefix)/`) while honouring the boolean `commit.gpgsign` — so the config looked live and two of three mutations came back green for the wrong reason.

**The wrong success signal.** My first table read exit codes. These suites call `pass()`/`fail()` but never `finish()`, so run standalone they exit 0 with failing assertions — exit status only captures a crash. Any "green" in that table could have been a suite printing ❌ and exiting 0. Re-run counting ❌ lines, with a probe first confirming the hostile config reaches git (`git check-ignore` on a file it should now ignore):

| pin dropped | updater-add-paths | updater-is-tracked |
|---|---|---|
| *(none — all pins present)* | green (0 ❌, exit 0) | green (0 ❌, exit 0) |
| `core.hooksPath` | crash (exit 1) | crash (exit 1) |
| `commit.gpgsign` | crash (exit 1) | crash (exit 1) |
| `core.excludesFile` | crash (exit 1) | **green** |

The conclusion survived the correction, and the failures are crashes rather than failed assertions — the fixture dies mid-setup instead of reaching an oracle. So the "both suites together" property holds for two of the three pins, not all three. `updater-is-tracked` does create untracked files, but none of its *assertions* turn on ignore resolution: it writes its own `.gitignore` and then reads index membership, which a global rule does not move. I kept that pin for both as a defensive one, and the docstring now says it is proven by one suite rather than implying two.

The other two `makeRepo` fixtures under `tests/` are deliberately untouched: `updater-local-system-edits` pins line endings instead of excludes and seeds a base commit plus an `upstream` branch, and `updater-rollback-behavior` pins nothing. They share a name, not a fixture, and folding them in would widen this diff into files the other queued updater PRs touch.

### Still your call: the two pre-existing ones

Unchanged from my earlier comment and not started, because they are wider than this PR's cause and I would rather not widen a diff you are sequencing:

1. `git commit -- <dir>` records unstaged tracked edits and index-stranded ignored files.
2. The failure detector compares status file paths against directory entries with an exact `Set.has()`.

Reproductions for both are above. Say "here" and they go in this branch; say "follow-up" and I will file them as *"apply(): `git commit -- <dir>` captures unstaged and index-stranded files the staging step deliberately excluded"* and link it from this PR. I have not filed anything yet — that is yours to sequence, not mine to pre-empt.

On #2591: unchanged, it closes as folded-in once this lands. It is showing conflicts with `main` now, which I have left alone for the same reason as the rebase — no point resolving a branch that closes.

(Correcting my own last comment while I am here: the workflows that needed the fork-approval click did run in the end. On `07a2616c` every check came back green except the macOS one above. The two commits here will re-run them, and may want the click again.)

Scott-Emberson (approved):
Code owner for `tests/`. The three owned files clear the updater bar. Each fix-test carries an oracle that proves the hazard exists first (raw `git add -f` sweeps the sibling) so it cannot pass vacuously, and the ignored user content is asserted absent from the index, which at the add layer is a stronger guarantee than a byte hash. `makeUpdaterRepo` in `helpers.mjs` is additive, neutralizes leak-prone global git config, and touches no existing export. Hermetic, matches the rollback-test shape. Approving.

santifer:
Heads-up, @gtm-k: this went `CONFLICTING` tonight and it was me, not you.

Three of my merges landed in the files you touch: **#3125** added a retrying `rmSync` wrapper to `tests/helpers.mjs`, and **#3126** and **#3101** both changed `update-system.mjs` (the second-update local-edit detector, and the `SYSTEM_PATHS` scanner). Your diff was clean before that.

The rebase should be mechanical: none of those three touches the staging path this PR fixes. If any of them do collide in a way that isn't obvious, say so here rather than guessing at my intent and I'll resolve it myself, since I caused it.

On the substance, so you know it isn't sitting unread: the two causes you separate are the right pair to fix together, and "finishes on disk, then fails at the commit" is exactly the failure shape that leaves a user staring at a half-done update with no idea what to run. I'll review it properly once it's green.

Also, your other one is in: **#2538 merged** a few minutes ago.

gtm-k:
Branch is current — `origin/main` merged in, 0 behind. Thanks for saying it was yours, @santifer; it made the conflict easy to read.

I merged rather than rebased. Three files conflicted, so the rebase was not the mechanical one you expected, and replaying 7 commits means meeting the `update-system.mjs` conflict up to 7 times — each resolution a chance to drop a hunk quietly. A merge resolves once and keeps the SHAs that CodeRabbit's threads and @Scott-Emberson's `tests/` approval are anchored to. @luochen211's note said "rebase or merge", so I took the option that preserves the review trail. Say the word if linear history matters more and I will redo it as a rebase.

Full reasoning and the four hunks are in the merge commit message. Two things from it worth surfacing here:

- All four hunks were take-both. In `tests/helpers.mjs`, `makeUpdaterRepo` (this branch) and `hermeticGitEnv` (#2567) are different functions that happened to land at the same offset — not competing versions of one thing. Both kept intact.
- I dropped `pathToFileURL` from the imports. Its only consumer was the inline `import.meta.url === pathToFileURL(argv[1]).href` main-module check that #3170/#3383 replaced with the `resolve()`/`realpathSync` comparison, in a region this branch never touches — so git merged it cleanly, no marker pointed at it, and a plain union would have left a dead import behind.

## @luochen211's scope list

**Per-file staging guard — kept.** `addPaths` still refuses anything that resolves to a directory; `apply()` stages `expandToShippedFiles(pathsToStage)`.

**Slash-free directory pathspecs (`docs`) — covered.** `tests/updater-add-paths.test.mjs` asserts it directly (`a slash-free directory is refused without staging anything`, plus `non-canonical directory spellings are refused too`).

**Tracked/untracked dismiss-marker behaviour — preserved**, including #2591's probe-before-unlink ordering, which matters on the failure path once `isTracked` propagates instead of catching.

**The two pre-existing concerns — I would like to split them, and one is more concrete than when you raised it.** `git commit -- <dir>` is still reachable:

```js
// update-system.mjs
2290:      addPaths(expandToShippedFiles(pathsToStage));   // staging: expanded, files only
2326:      git('commit', ..., '--', ...pathsToStage);      // commit: unexpanded, directories included
```

So this PR closed the staging half of that seam and not the commit half. Unlike the `git add -f` case it does not reach ignored files (a pathspec commit will not add untracked content), but it does take working-tree content for tracked files under the directory, which is the "unstaged/index-stranded" capture you described.

I would rather not fix it here. It is not the one-line symmetry it looks like — `pathsToStage` also carries `:(exclude)` entries (see the `EXCLUDE_PATHSPEC_PREFIX` filter at 2317), which `expandToShippedFiles` has no handling for, so doing it properly means deciding what expansion means for a negative pathspec and testing that separately. Landing it here would also move the diff out from under an approval already given. Happy to open the follow-up issue with this repro and link it, if you want it tracked before this merges.

## One thing the merge surfaced that I did not fix here

Putting `hermeticGitEnv` (#2567) and `makeUpdaterRepo` side by side in `tests/helpers.mjs` makes it visible that the second is defeatable in a way the first exists to prevent, and I would rather report it than quietly leave it.

`makeUpdaterRepo` pins `core.excludesFile` with `git config`, which writes the repo-local `.git/config` — a config *file*. `gitIn`/`gitRawIn` (`update-system.mjs:690-704`) call `execFileSync` with no `env` override, so `process.env` is inherited whole, and git applies `GIT_CONFIG_COUNT`/`KEY_n`/`VALUE_n` *after* every config file. That is precisely the runtime layer `hermeticGitEnv`'s own docstring says "overrides even the one a fixture sets for itself".

It reproduces:

```console
$ node tests/updater-add-paths.test.mjs
exit=0  assertions=32

$ GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=core.excludesFile GIT_CONFIG_VALUE_0=<file containing "*.txt"> \
    node tests/updater-add-paths.test.mjs
exit=1  assertions=5
...
stdout: 'nothing to commit (create/copy files and use "git add" to track)\n'
```

The fixture's own `seed.txt` is excluded by the ambient rule, `git add -A` stages nothing, and the base commit dies. Same in `updater-is-tracked`. Note the shape: an uncaught crash, not a red assertion — arguably worse to diagnose than the silently-wrong isolation the pin exists to prevent.

This predates the merge — `makeUpdaterRepo` was never stronger, and #2567 landed independently — so it is not a regression and I have deliberately left the diff alone: it is test-isolation hardening rather than a product bug, and the fix is not the small one it looks like, since `gitIn` has a `(dir, ...args)` signature with nowhere to thread an env through. Happy to take it as a follow-up, or to fold it in here if you would rather it not sit open. At minimum `makeUpdaterRepo`'s docstring should say which layer it does *not* close, since it currently reads as though it closes all of them.

## Verification

Full suite on the merge result: **7313 passed, 1 failed**.

The one failure is `tests/skill-project-root.test.mjs` and it is not from this branch. It reproduces byte-identical on a pristine `origin/main` worktree at 8af93cd4:

```
❌ .antigravitycli\...\SKILL.md: missing cwd-independent routing rule | .claude\... | .cursor\... | .grok\... | .kimi\... | .opencode\... | .qwen\...
```

Those seven entrypoints are symlinks (git mode `120000`) pointing at `.agents/skills/career-ops/SKILL.md`. My clone has `core.symlinks=false` — the Windows default outside Developer Mode — so git materialises them as one-line regular files containing the link target, and `readFileSync` returns `"../../../.agents/..."` instead of the 203-line skill. Same bytes, different checkout semantics; the CI runners resolve the links and this goes green there. Worth knowing it makes the suite unrunnable-clean for Windows contributors without Developer Mode, but that is its own issue, not this one.

Upgrade regression gate, all three legs:

```console
--pr-gate     -> EXIT=0  | GREEN
--canary      -> EXIT=0  | GREEN
--local-paths -> EXIT=0  | GREEN
```

Including `--local-paths`, which is main's leg that this merge picks up.

Pushed as `f247f2ca`; the PR reads `MERGEABLE` again. CI is running now and I will report back here rather than leave you to check.

One housekeeping note: the remaining unresolved CodeRabbit thread (extract the duplicated `makeRepo`) is already addressed — that is what `makeUpdaterRepo` in `tests/helpers.mjs` is, and @Scott-Emberson's approval refers to it. It is marked outdated and just never got resolved; safe for a reviewer to close.

gtm-k:
Went ahead and opened the follow-up rather than leaving it hanging on a yes/no: **#3504**, with the repro isolated to the three file kinds so the scope is unambiguous — a tracked file with an unstaged edit is committed by the pathspec commit, an untracked one and an ignored one are not.

That answers @luochen211's fix-or-split ask in the split direction. If you would rather it land inside this PR after all, say so and I will close #3504 and fold it in here instead.

gtm-k:
CI as promised, @luochen211 — green on `f247f2ca`, all 12 checks:

| Check | |
|---|---|
| `test (ubuntu-latest)` · `test (macos-latest)` · `test (windows-latest)` | ✅ |
| `Upgrade regression gate` | ✅ |
| `cv-visual` · `guard` · `label` · `dependency-review` | ✅ |
| CodeQL (js-ts, go) · GitGuardian | ✅ |

`mergeStateStatus` is `BLOCKED` on required review only, not on a check.

Two notes, since both were red the last time this branch reported CI:

- **`test (macos-latest)` passes now.** It was failing on `agent-inbox-tests.mjs` section 7 (the 30-way concurrent `add`), which is #2777/#2824 — fixed on `main` by three commits that all landed after this branch's old base. Merging main picked them up.
- **`test (windows-latest)` passes too.** The `skill-project-root` failure I reported in the comment above is local-only: the runner has symlink support, so it resolves the seven CLI entrypoints and reads the real skill. Worth knowing the suite is not cleanly runnable for a Windows contributor without Developer Mode, but that is its own issue and not this branch's.

santifer:
@gtm-k you did exactly what I asked and then got silence for it, which is the wrong way round.

On the 20th I told you this went `CONFLICTING` because of three of my merges, not anything you did, and asked you to rebase. You did. State as of right now, checked rather than assumed: **MERGEABLE, CI 12 green / 0 failed**. Nothing is blocking this except me getting to it.

I'm at my merge ceiling for this session, so it goes in the next maintenance pass, and it goes near the front: "already delivered and waiting" is the most expensive queue there is, and you're in it because of a conflict I created. Sorry for the wait.

santifer:
This is how a hard updater bug should be fixed, @gtm-k: the FETCH_HEAD expansion means -f can never sweep a user's ignored files by construction, and rejectDirectories makes the precondition enforced instead of documented. You also rebased through a conflict I caused and kept the quality up. Merged 🚀 (@calebwhite-io's #1996 diagnosis of the dismiss-marker half keeps first-credit, as agreed.)

PR #2534, opened 2026-08-05
chore(gitignore): ignore .idea/, .vscode/, and .mcp.json

ilyakanevskiy:
## Summary

- `.idea/` and `.vscode/` (JetBrains / VS Code project folders) aren't ignored at all currently — standard editor clutter with no reason to ever be tracked.
- `.mcp.json` (Claude Code's project-level MCP server config) isn't ignored either. It commonly ends up holding personal MCP server credentials once a user adds a server (e.g. a plugin's API token) — same risk class the repo already treats `.env` / `**/.env.local` with.

## Test plan

- [x] `node test-all.mjs --quick` — full suite green

## Summary by CodeRabbit

* **Chores**
  * Updated project ignore settings to exclude additional IDE/editor configuration files and local credential-related configuration.
  * Added guidance clarifying the credential risk associated with local configuration files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -96,6 +96,10 @@ data/model-blacklist.json
 *.mov
 *.mp4
 
+# IDE-specific local config
+.idea/
+.vscode/
+
 # CLI-specific local config
 .claude/settings.local.json
 .claude/memory/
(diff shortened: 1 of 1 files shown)


santifer:
Merged, @ilyakanevskiy. The `.mcp.json` half is the one that earns this: it's the same risk class as `opencode.json`, which this repo already ignores for exactly the reason you give, with the same comment. A config that starts empty and quietly acquires an API token the first time someone adds a server is precisely the file you don't want a reflexive `git add .` to reach.

Verified before merging that none of the three are currently tracked (control: eight tracked dotfiles list fine), so nothing gets orphaned by the new rules 🚀

PR #2536, opened 2026-08-05
fix(http): consolidate stale/duplicated User-Agent strings into a shared module

ilyakanevskiy:
## Summary

- Several places hardcoded a User-Agent string with a stale/hand-copied version number instead of deriving it from `package.json`: `providers/_http.mjs`'s `DEFAULT_USER_AGENT` (`career-ops/1.3`, well behind the actual `1.25.0`), `openrouter-runner.mjs` (`career-ops/1.0`), `seeds/vc-portfolios.mjs` (`career-ops-seeds/1.0`), and `liveness-api.mjs` (`career-ops-liveness/1.0`).
- New `user-agent.mjs` owns `DEFAULT_USER_AGENT` (read from `package.json` at call time, so it can't go stale again) and `BROWSER_LIKE_USER_AGENT` (the browser-spoofing UA used to clear WAF/CDN blocks, bumped Chrome/131 → Chrome/151). `providers/_http.mjs` imports both and re-exports `BROWSER_LIKE_USER_AGENT` so existing provider imports (`workday.mjs`, `icims.mjs`, `glints.mjs`) keep working unchanged.
- `openrouter-runner.mjs`, `seeds/vc-portfolios.mjs`, `liveness-api.mjs`, and `liveness-browser.mjs` now import from the shared module instead of hardcoding their own copy. `liveness-browser.mjs` previously ran a separate, also-stale Chrome/120 Mac-flavored UA for its Playwright context — consolidated to the shared constant.
- Two providers/scripts had rolled their own local copy of the browser-spoofing UA instead of importing the shared one: `providers/oraclecloud.mjs` (Chrome/120) and `company-funded.mjs` (Chrome/131). Both now import `BROWSER_LIKE_USER_AGENT` — `oraclecloud.mjs` from `providers/_http.mjs` (it's a provider), `company-funded.mjs` from `user-agent.mjs` directly (it's a root script, not a provider).
- Registered `user-agent.mjs` in `update-system.mjs`'s `SYSTEM_PATHS` so it ships on update.

### Note for @santifer

I don't know why `liveness-api.mjs` and `seeds/vc-portfolios.mjs` used their own distinct UA identifiers (`career-ops-liveness/1.0`, `career-ops-seeds/1.0`) instead of the same default one `providers/_http.mjs` already had. It might have been intentional (e.g. distinguishing traffic sources server-side) or just copy-paste drift when those files were written. I collapsed everything into one shared `DEFAULT_USER_AGENT` on the assumption it was the latter — happy to keep them distinct instead if there was a reason for it.

## Test plan

- [x] `node test-all.mjs` — 2975/2975 passing, before and after
- [x] Confirmed no test hardcoded the old UA/Chrome-version strings
- [x] Grepped the whole repo for `Mozilla/5.0`, `career-ops/1.`, `career-ops-seeds`, `career-ops-liveness` post-change — only the new `user-agent.mjs` matches

## Summary by CodeRabbit

* **Improvements**
  * Standardized request identification across company, liveness, provider, portfolio, and job-page services.
  * Browser-based requests now consistently use a browser-compatible identity, improving compatibility with security and delivery systems.
  * Request identification remains stable and consistent across supported services.

* **Maintenance**
  * Automatic system updates now include shared request-identity configuration.
  * Added coverage to verify request identifiers and outgoing request behavior.

### DIFF
--- a/company-funded.mjs
+++ b/company-funded.mjs
@@ -14,14 +14,13 @@ import { dirname, join } from 'path';
 import { fileURLToPath, pathToFileURL } from 'url';
 
 import { decodeEntities } from './providers/_html-entities.mjs';
+import { BROWSER_LIKE_USER_AGENT } from './user-agent.mjs';
 
 const ROOT = dirname(fileURLToPath(import.meta.url));
 const DEFAULT_LIMIT = 20;
 const DEFAULT_MONTHS = 3;
 const DEFAULT_SORT = 'date';
 const DEFAULT_SOURCES = ['techcrunch', 'prnewswire', 'guardian', 'hn'];
-const BROWSER_UA =
-  'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36';
 
 const RSS_SOURCES = {
   techcrunch: [
@@ -431,7 +430,7 @@ async function fetchTextMeta(url, { timeoutMs = 12_000, source = '' } = {}) {
     let currentUrl = startUrl;
     for (let hops = 0; hops <= 5; hops++) {
       const res = await fetch(currentUrl, {
-        headers: { 'user-agent': BROWSER_UA, accept: 'application/rss+xml,application/xml,text/xml,text/plain,*/*' },
+        headers: { 'user-agent': BROWSER_LIKE_USER_AGENT, accept: 'application/rss+xml,application/xml,text/xml,text/plain,*/*' },
         redirect: 'manual',
         signal: controller.signal,
       });
--- a/liveness-api.mjs
+++ b/liveness-api.mjs
@@ -27,6 +27,8 @@
  * (no slashes / traversal), and server-side redirects are refused.
  */
 
+import { DEFAULT_USER_AGENT } from './user-agent.mjs';
+
 const TIMEOUT_MS = 8_000;
 // Strict path-segment charset. Anything with a slash, dot-dot, or other char is
 // rejected before it can reach the fixed-host API URL template.
@@ -206,7 +208,7 @@ export async function checkLivenessViaApi(url) {
     try {
       res = await fetch(apiUrl, {
         method: 'GET',
-        headers: { 'user-agent': 'career-ops-liveness/1.0', accept: 'application/json' },
+        headers: { 'user-agent': DEFAULT_USER_AGENT, accept: 'application/json' },
         redirect: 'error', // refuse server-side redirects (SSRF + ambiguity guard)
         signal: controller.signal,
       });
--- a/liveness-browser.mjs
+++ b/liveness-browser.mjs
@@ -6,6 +6,7 @@
  */
 
 import { classifyLiveness } from './liveness-core.mjs';
+import { BROWSER_LIKE_USER_AGENT } from './user-agent.mjs';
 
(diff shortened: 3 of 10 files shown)


ilyakanevskiy:
@coderabbitai review

santifer:
@ilyakanevskiy consolidating the duplicated User-Agent strings into one constant is worth doing, and the refactor itself is clean: seven call sites stop hand-copying and the grep comes back empty afterwards.

Two behaviour changes ride along undeclared, and the first one has a blast radius of every provider we have.

**1. The UA string itself changes.** Today it is `Mozilla/5.0 (compatible; career-ops/1.3)`; the PR resolves it from `package.json`, so it becomes `career-ops/1.25.0` and will keep changing with every release. That reaches all 68 providers through `fetchWithTimeout`, plus the openrouter runner, the liveness checks and the VC-portfolio seeds. Some boards fingerprint on the UA, and a string that moves on every release is a variable we would be introducing into every scan without meaning to. If the point is honesty about who we are, a stable `career-ops` identifier does that without the version churn; if you want the version in there, say so explicitly in the PR so it is a decision rather than a side effect.

**2. No behavioural test.** The only assertion the diff adds is the auto-enumerated "syntax OK" for the new file, and a repo-wide grep for the UA constants finds one unrelated hit. So nothing would catch a future change to the string, which is the exact thing that just changed silently here.

**What I'd take:** keep the consolidation, pin the UA to a stable value (or declare the version-in-UA as intentional), and add one test asserting the outgoing header so the next person cannot move it by accident.

The refactor is the right idea: it is the two riders that need to be decisions.

ilyakanevskiy:
My original goal was narrower than it looked: get rid of the mismatch between files, both in *shape* (`career-ops` vs `career-ops-seeds` vs `career-ops-liveness`) and in *version* (`1.3` vs `1.0`). `_http.mjs` had drifted to `career-ops/1.3` at some point, and I assumed the other files were supposed to keep pace with it — the way browsers bump their own UA version — but that update just never happened elsewhere. I don't have a real argument for tying the UA to the release version beyond that assumption, so I'm dropping it: `DEFAULT_USER_AGENT` is now pinned everywhere to the literal `Mozilla/5.0 (compatible; career-ops/1.0; +https://github.com/santifer/career-ops)`. The `/1.0` is a UA-format version (Googlebot/2.1-style), bumped by hand only if the identifier's shape itself changes — never tied to `package.json`.

Also added `tests/user-agent.test.mjs`, which asserts the literal value and drives a request through `fetchJson` to confirm that exact string is what actually goes out on the wire, so this can't drift silently again.

One thing to flag: this rolls the already-established `/1.3` back down to `/1.0`. @santifer are you fine with that, or would you rather I pin it at `/1.3` (the value already live in production) instead?

santifer:
Good cleanup, @ilyakanevskiy. The removed strings weren't identical — they were drift (career-ops/1.0 vs /1.3 vs -seeds/1.0, Chrome/120 vs /131) — so consolidating to two canonical values is the right call: one honest `DEFAULT_USER_AGENT` that now carries a +contact URL (good scraper etiquette), and one current `BROWSER_LIKE_USER_AGENT` (Chrome/151, newer than every variant it replaces, so nothing that passed before will fail now). I verified the values before merging — the change is strictly-better, not just DRY. Registered in SYSTEM_PATHS, own test included. Merged! 🚀

Issue #2278, opened 2026-07-29
jd-skill-gap.mjs silently returns zero skills for non-engineering JDs, and pdf mode Step 4 treats that as a pass


ringo380:
`jd-skill-gap.mjs` returns `JD skills found: 0` for any JD outside the engineering-tool vocabulary, and `modes/pdf.md` Step 4 makes that silent zero a gate.

## Reproduction

Same tool, same flags, four JDs:

```
jds/gitlab-enablement-content-manager.md   JD skills found: 0    (19,852 chars, complete JD)
jds/twilio-senior-partner-enablement-...   JD skills found: 0
jds/huggingface-senior-python.md           JD skills found: 8
jds/workos.md                              JD skills found: 5
```

The two engineering JDs classify fine. The two enablement JDs return zero — not "no gaps", but *no skills detected at all*, so the `existing` / `supportedByResume` / `gap` buckets are all empty.

## Why it matters

`modes/pdf.md` Step 4 says to run this before drafting and to "tell the user explicitly which skills are gaps before generating the CV." When the classifier finds nothing, that step produces a clean bill of health that looks identical to "no gaps found". The failure is indistinguishable from success at the call site, which is the worst shape for a gate.

The GitLab JD names ADDIE, SAM, adult learning principles, instructional design, role-based learning paths, certifications, accreditations, validation programs, blended learning, scenario-based learning, technical labs, assessments, e-learning modules, talk tracks, playbooks, demo flows, release readiness, content governance, and LMS. None are recognized.

## Scope

The skill dictionary appears to cover programming languages, frameworks, cloud, and infra only. Whole role families are invisible to it: enablement / L&D, instructional design, program and project management, technical writing, design, marketing ops, data analysis outside the named tools. career-ops advertises itself as CLI- and role-agnostic, so this is a coverage gap rather than an edge case.

## Suggested shape

1. Add non-engineering skill families to the dictionary (enablement/L&D, instructional design and assessment, program management, content and documentation, GTM/revenue ops).
2. Make the zero case loud: when a JD of non-trivial length yields zero skills, print a warning that the classifier found nothing rather than an empty three-bucket summary — the caller can then fall back to manual analysis knowingly.
3. Optionally surface a confidence signal (skills found per 1,000 chars) so an unusually low density is visible without reading the source.

Point 2 is worth doing on its own even before the dictionary grows: it converts a silent wrong answer into a visible one.

[...]

PR #2541, opened 2026-08-05
fix(jd-skill-gap): CRLF-terminated JDs extracted zero skills

### LINKED ISSUE #2540: bug(jd-skill-gap): CRLF-terminated JDs extract zero skills — BULLET_LINE_RE never matches
### What happened?

`jd-skill-gap.mjs` extracts **zero** skills from any JD file that uses CRLF line endings. It does not error — it returns an empty classification, which every Windows user gets by default since CRLF is the platform's default line ending.

The cause is `BULLET_LINE_RE`:

```js
const BULLET_LINE_RE = /^\s*[-*•]\s*(.+)$/;
```

The JD is split into lines, so on a CRLF file each requirement line still carries a trailing `\r`. Two things then combine:

- JavaScript counts `\r` as a line terminator, so `.` cannot consume it — `(.+)` stops before the `\r`.
- Without the `m` flag, a bare `$` only matches at end-of-string, not before a line terminator.

So `$` has to match at a position that still has a `\r` after it, and the whole regex fails on every bullet. No bullets match, no skill candidates are extracted, and the result is an empty set.

The failure is silent rather than loud. The `no-skill-candidates` low-confidence guard added in #2278 does fire, which is the only reason the run is not completely indistinguishable from "this JD has no requirements" — but the stated reason points at the JD, not at the line endings, so the user has no way to reach the real cause.

### Steps to reproduce

Write the same JD twice, once LF-terminated and once CRLF-terminated:

```
# Senior Backend Engineer

## Requirements

- Strong experience with Python and Go
- Deep knowledge of PostgreSQL and Redis
- Hands-on with Docker, Kubernetes and Terraform
- Familiarity with Kafka and gRPC
- Experience with AWS and CI/CD pipelines
```

Then:

```bash
node jd-skill-gap.mjs jd-lf.md   --summary
node jd-skill-gap.mjs jd-crlf.md --summary
```

Observed on `main`:

```
--- LF ---
JD skills found: 11

--- CRLF ---
JD skills found: 0

  LOW CONFIDENCE: this is not a clean result.
     A requirements section was found and scanned, but no skill candidates were
     extracted from it. [...] (reason: no-skill-candidates)
```

Identical content, identical requirements section — only the line endings differ.

### Expected behavior

Line endings should not change the result. Both files should report the same 11 skills and the same classification.

### Impact

`jd-skill-gap.mjs` is the zero-LLM gate `pdf` mode Step 4 relies on, so on Windows the gap check contributes nothing to CV tailoring while still reporting a completed run. Any JD saved by a Windows editor, or fetched and written with CRLF on any platform, hits this.

### Fix

`(.+)\r?$` — let the group stop before an optional `\r` and allow `$` to match after it. One optional group in one regex; no behavior change for LF input.

I will open a PR referencing this issue. Note that nothing in `--self-test` or `tests/skill-extract.test.mjs` currently feeds CRLF input, so there is no regression assertion guarding this today — happy to add one to `runSelfTest()` if you would like it in the same PR.

### CLI tool

Not CLI-specific — the script runs standalone under plain `node`.

### OS and Node.js version

Windows 10 (26200) / Node 22. The root cause is platform-independent: any CRLF-terminated JD reproduces it on macOS and Linux too.

nhan4013:
## What does this PR do?

Makes `jd-skill-gap.mjs` work on CRLF-terminated JDs. `BULLET_LINE_RE` ended in a bare `$` with no `m` flag, so no bullet on a CRLF-split line could match and the script returned **zero** skills instead of erroring — the default outcome for Windows users.

## Related issue

Fixes #2540

## Root cause

```js
const BULLET_LINE_RE = /^\s*[-*•]\s*(.+)$/;
```

The JD is split into lines, so on a CRLF file each line still carries a trailing `\r`. Two rules combine to make the match impossible:

- JavaScript treats `\r` as a line terminator, so `.` cannot consume it and `(.+)` stops before it.
- Without the `m` flag, `$` only matches at end-of-string — not before a line terminator.

`$` is therefore asked to match at a position that still has a `\r` after it. Every bullet fails, no skill candidates are extracted, and the classification comes back empty. The `no-skill-candidates` low-confidence guard from #2278 does fire, but it attributes the emptiness to the JD, so the real cause stays invisible.

## The change

```diff
-const BULLET_LINE_RE = /^\s*[-*•]\s*(.+)$/;
+const BULLET_LINE_RE = /^\s*[-*•]\s*(.+)\r?$/;
```

Plus a comment recording why the `\r?` is load-bearing, so it does not get "cleaned up" later. One optional group; LF input is unaffected.

## Test evidence

Same JD content written twice, differing only in line endings:

```
# Senior Backend Engineer

## Requirements

- Strong experience with Python and Go
- Deep knowledge of PostgreSQL and Redis
- Hands-on with Docker, Kubernetes and Terraform
- Familiarity with Kafka and gRPC
- Experience with AWS and CI/CD pipelines
```

Before (on `main`):

```
LF   -> JD skills found: 11
CRLF -> JD skills found: 0
        LOW CONFIDENCE: ... (reason: no-skill-candidates)
```

After (this branch):

```
LF   -> JD skills found: 11
CRLF -> JD skills found: 11
```

The full JSON output for the two files is now byte-identical.

Suites run on this branch:

- `node jd-skill-gap.mjs --self-test` — **46 passed, 0 failed**
- `node tests/skill-extract.test.mjs` — all assertions pass
- `node test-all.mjs` — **2973 passed, 0 failed**, exit 0, with 2 environment-only warnings (no Go compiler in PATH so the dashboard build was skipped; the plugin symlink-traversal test skipped on `EPERM` because unprivileged Windows cannot create symlinks). Both are unrelated to this change and also occur on unmodified `main`.

## A note on regression coverage

Nothing in `--self-test` or `tests/skill-extract.test.mjs` currently feeds CRLF input, so this fix ships without a guard against reintroduction. I kept the diff to the one-line fix rather than expanding scope unasked — say the word and I will add the LF/CRLF-parity case to `runSelfTest()` in this PR.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bullet-point parsing for job description files using Windows-style line endings (CRLF).
  * Preserved existing parsing behavior for other line-ending formats.

@all-contributors please add @nhan4013 for code, test

### DIFF
--- a/jd-skill-gap.mjs
+++ b/jd-skill-gap.mjs
@@ -79,7 +79,9 @@ const NON_REQUIREMENT_HEADER_RE = new RegExp(
   'im'
 );
 
-const BULLET_LINE_RE = /^\s*[-*•]\s*(.+)$/;
+// `\r?$` is required, not cosmetic: JS treats \r as a line terminator, so `.`
+// cannot consume it and a bare `$` never matches on a CRLF-split line.
+const BULLET_LINE_RE = /^\s*[-*•]\s*(.+)\r?$/;
 
 // A conservative skill-token extractor: pulls comma/slash/and-separated
 // technical-looking tokens out of a requirement bullet, rather than treating
--- a/jd-skill-gap.mjs
+++ b/jd-skill-gap.mjs
@@ -504,6 +504,27 @@ Benefits and Perks (US Only)
   eq('does not extract "Deep" as a skill', dispositionSkills.includes('Deep'), false);
   eq('does not extract "Interest" as a skill', dispositionSkills.includes('Interest'), false);
 
+  // Regression (#2540): a JD saved with CRLF line endings extracted zero skills.
+  // JS treats \r as a line terminator, so `.` cannot consume it and the bare `$`
+  // in BULLET_LINE_RE (no `m` flag) never matched — every requirement bullet
+  // failed the test and the run returned an empty list, which reads exactly like
+  // "no gaps" instead of erroring. Parity between the two line endings is the
+  // property that was broken, so both variants are built explicitly: with
+  // core.autocrlf the template literal below is already CRLF in a Windows
+  // checkout, and an unnormalized fixture would compare CRLF against CRLF and
+  // pass on the broken regex.
+  const lineEndingJd = `
+# Role
+
+## Requirements
+- Python, FastAPI, PostgreSQL
+- Experience with Kubernetes
+`;
+  const lfSkills = extractJdSkills(lineEndingJd.replace(/\r\n/g, '\n'));
+  const crlfSkills = extractJdSkills(lineEndingJd.replace(/\r\n/g, '\n').replace(/\n/g, '\r\n'));
+  eq('CRLF JD extracts the same skills as the LF JD', crlfSkills, lfSkills);
+  eq('CRLF JD extracts a non-zero number of skills', crlfSkills.length > 0, true);
+
   // Regression (#1896): the reported bug. A CV alias and a JD's canonical name
   // must not read as a gap. Before the shared skill-extract canonicalization,
   // classify compared the two literally — "k8s" in the CV vs "Kubernetes" in


nhan4013:
Added a regression test for this fix in `runSelfTest()`, following the existing assertion style (no new helper or framework).

It parses one small synthetic requirements block twice — once LF-terminated, once CRLF-terminated — and asserts that the two extractions are **identical**, plus that the CRLF variant returns a **non-zero** skill count, since returning zero skills was the exact failure signature of the bug.

Both variants are normalized explicitly rather than relying on the fixture's own line endings. With `core.autocrlf` the template literal is already CRLF in a Windows checkout, so an unnormalized fixture would compare CRLF against CRLF and pass even on the broken regex.

Verified the test actually catches the bug. With `BULLET_LINE_RE` reverted to the pre-fix `/^\s*[-*•]\s*(.+)$/`:

```
  FAIL: CRLF JD extracts the same skills as the LF JD
    expected: ["Python","FastAPI","PostgreSQL","Kubernetes"]
    actual:   []
  FAIL: CRLF JD extracts a non-zero number of skills
    expected: true
    actual:   false

jd-skill-gap self-test: 46 passed, 2 failed
```

With the fix in place:

```
jd-skill-gap self-test: 48 passed, 0 failed
```

`node tests/skill-extract.test.mjs` still passes, and the full `node test-all.mjs` suite is green (2975 passed, 0 failed). Still a single-file change.

santifer:
Welcome to career-ops, @nhan4013! Solid first contribution: a JD pasted from Windows would silently yield zero skills, which is the worst kind of bug because it looks like an empty result rather than a failure. The regex fix is right and your two assertions genuinely fail without it (I checked by reverting only that hunk). One small note for next time, no action needed: the comment justifying the extra normalization describes behaviour that doesn't happen, and that normalization is unreachable. Merged — your first contribution. Glad to have you here.

Issue #774, opened 2026-06-02
Ignore *.bak backup files generated by tracker maintenance scripts


Mystogam:
### Code of Conduct

- [x] I agree to follow the [Code of Conduct](https://github.com/santifer/career-ops/blob/main/CODE_OF_CONDUCT.md)

### Existing issues

- [x] I searched existing issues and this hasn't been requested yet

### Problem

Both normalize-statuses.mjs and dedup-tracker.mjs drop an applications.md.bak file before they rewrite the tracker. Problem is, *.bak isn't in my gitignore, so that backup keeps showing up as untracked every time I run either one. Not a big deal, but it junks up git status and I could easily commit it by accident without noticing.

### Proposed solution

Add a *.bak ignore rule to .gitignore:

  gitignore file
  Backup files written by normalize-statuses.mjs / dedup-tracker.mjs
  *.bak

### Area

Other

PR #2543, opened 2026-08-05
chore(gitignore): ignore timestamped user-layer backups (cv.md PII left untracked)

### LINKED ISSUE #2542: bug(gitignore): timestamped user-layer backups escape .gitignore — cv.md PII left untracked in the repo root
### What happened?

The web dashboard writes timestamped backups of user-layer files, and `.gitignore` does not match them. A backup of `cv.md` — full name, email, phone, employment history — ends up **untracked but git-visible in the repo root**.

`web/src/lib/core/safe-write.ts:27` builds the name:

```ts
const ts = new Date().toISOString().replace(/[:.]/g, "-");
const bak = `${file}.bak-${ts}`;
```

`.gitignore` ignores `*.bak` (added in #774). `cv.md.bak-2026-08-05T16-55-08-641Z` does not *end* in `.bak`, so the glob misses it.

Three API routes call `atomicWriteWithBackup` on user-layer files whose ignore rules are exact paths, so neither the exact rule nor `*.bak` covers the backup:

| Route | File written | Backup produced | Ignored? |
|---|---|---|---|
| `web/src/app/api/cv/route.ts:37` | `cv.md` | `cv.md.bak-<ts>` | **no** |
| `web/src/app/api/profile/route.ts:96` and `web/src/app/api/followups/cadence/route.ts:89` | `config/profile.yml` | `config/profile.yml.bak-<ts>` | **no** |
| `web/src/app/api/portals/route.ts:53` | `portals.yml` | `portals.yml.bak-<ts>` | **no** |

`data/`-resident backups are covered, but only incidentally, by the blanket `data/*` rule.

The backups themselves are correct and wanted — `safe-write.ts` exists precisely because user files are gitignored and have no git recovery path (#649/#704/#920/#958). The bug is only that `.gitignore` does not know about the naming convention the web layer uses.

### Steps to reproduce

On a clean checkout of `main`:

```bash
git check-ignore -v cv.md.bak-2026-08-05T16-55-08-641Z
git check-ignore -v config/profile.yml.bak-2026-08-05T16-55-08-641Z
git check-ignore -v portals.yml.bak-2026-08-05T16-55-08-641Z
```

All three exit `1` with no output — not ignored. For contrast, `git check-ignore -v cv.md.bak` matches `.gitignore:89:*.bak`.

To reach it through the product: open the dashboard, edit the CV, save. `POST /api/cv` snapshots the prior CV, and `git status` then shows an untracked file in the repo root containing it.

### Expected behavior

Every backup of a gitignored user-layer file should also be gitignored, whatever suffix the writer appends. `git status` in a configured install should stay clean of PII.

### Impact

Not a remote-exploitable issue — it needs a local commit-and-push to become a disclosure. But the failure mode is quiet and the blast radius is bad: the file is invisible in the product, it sits in the repo root next to tracked sources, and a single `git add -A` sweeps it in. career-ops users are by definition running this inside a git repo they may push, and CONTRIBUTING.md already lists "PRs containing personal data" as something the project rejects, so the guardrail is meant to be the ignore file rather than user discipline.

The `no-user-data` workflow catches this at PR time for contributors, which limits the damage to this repo — but it does nothing for a user who pushes their own configured clone to a personal GitHub repo, which is the case that actually leaks.

### Suggested fix

Widen the pattern to cover any suffix after `.bak`:

```diff
-# Backup files written by normalize-statuses.mjs / dedup-tracker.mjs
-*.bak
+*.bak*
```

Rationale for `*.bak*` rather than an enumerated `*.bak-*`:

- It covers the timestamped form, and also numbered variants (`cv.md.bak2`, `cv.md.bak10`) and any future suffix a writer picks. An enumerated alternative like `*.bak[0-9]` silently misses two-digit variants — I verified `cv.md.bak10` escapes it.
- It subsumes the existing `*.bak`, so it is a net one-pattern rule rather than a growing list.
- It hides nothing that is tracked: `git ls-files | grep '\.bak'` is empty, so no tracked path changes state.

Failing safe matters more than precision here, since the cost of a miss is PII in a public repo and the cost of an over-match is an ignored stray `.bak*` file.

### Two related notes

- **Delivery.** `.gitignore` is absent from `SYSTEM_PATHS` in `update-system.mjs`, so a hardened rule reaches new clones and anyone who pulls, but not existing installs running `update-system.mjs apply` — exactly what #2277 is about. This fix is worth taking on its own, but #2277 gates how far it travels.
- **`.npmignore`.** It carries the same `*.bak` at line 77 with the same gap. `package.json` is `private: true` with no `files` field, so nothing is published today and this is not a live leak — but the two files are worth keeping in sync. Out of scope here; happy to send it separately if you want it.

### CLI tool

Not CLI-specific — the writer is the web dashboard (`web/`).

### OS and Node.js version

Windows 10 (26200) / Node 22. Platform-independent: the filename comes from `toISOString()`, not from the OS.

nhan4013:
## What does this PR do?

Widens the backup rule in `.gitignore` from `*.bak` to `*.bak*`, so the timestamped backups the web dashboard writes are ignored. Today a saved CV edit leaves `cv.md.bak-<ISO timestamp>` untracked in the repo root, holding the same PII as `cv.md` itself.

## Related issue

Fixes #2542

## Root cause

`web/src/lib/core/safe-write.ts:27` names its snapshots with a suffix, not an extension:

```ts
const ts = new Date().toISOString().replace(/[:.]/g, "-");
const bak = `${file}.bak-${ts}`;
```

`cv.md.bak-2026-08-05T16-55-08-641Z` does not *end* in `.bak`, so the `*.bak` rule from #774 does not match it. And because `cv.md`, `config/profile.yml` and `portals.yml` are ignored by exact path, their backups are covered by neither rule:

| Route | File written | Backup | Ignored on `main`? |
|---|---|---|---|
| `web/src/app/api/cv/route.ts:37` | `cv.md` | `cv.md.bak-<ts>` | **no** |
| `web/src/app/api/profile/route.ts:96`, `web/src/app/api/followups/cadence/route.ts:89` | `config/profile.yml` | `config/profile.yml.bak-<ts>` | **no** |
| `web/src/app/api/portals/route.ts:53` | `portals.yml` | `portals.yml.bak-<ts>` | **no** |

The backups are correct and wanted — `safe-write.ts` exists because user files are gitignored and have no git recovery path. Only the ignore rule was out of step with the naming convention.

## Why `*.bak*` and not `*.bak-*`

- It also covers numbered variants. I verified that `cv.md.bak10` escapes a `*.bak[0-9]`-style rule, so an enumerated pattern just moves the gap rather than closing it.
- It subsumes the existing `*.bak`, so this is a net one-pattern rule instead of a list that has to grow every time a writer picks a new suffix.
- It hides nothing that is tracked: `git ls-files | grep '\.bak'` returns empty on `main`, so no tracked path changes state.

Failing safe is the right trade here — a miss means PII in a public repo, an over-match means a stray `.bak*` file is ignored.

## Test evidence

Before, on `main` — all three exit `1` (not ignored):

```
$ git check-ignore -v cv.md.bak-2026-08-05T16-55-08-641Z
$ git check-ignore -v config/profile.yml.bak-2026-08-05T16-55-08-641Z
$ git check-ignore -v portals.yml.bak-2026-08-05T16-55-08-641Z
```

After, on this branch:

```
.gitignore:93:*.bak*    cv.md.bak-2026-08-05T16-55-08-641Z
.gitignore:93:*.bak*    config/profile.yml.bak-2026-08-05T16-55-08-641Z
.gitignore:93:*.bak*    portals.yml.bak-2026-08-05T16-55-08-641Z
.gitignore:93:*.bak*    cv.md.bak
.gitignore:93:*.bak*    cv.md.bak2
.gitignore:93:*.bak*    cv.md.bak10
.gitignore:93:*.bak*    data/pipeline.md.pre-reconcile.bak
```

`node test-all.mjs` — **2973 passed, 0 failed**, exit 0, including the *"user-layer files are git-ignored"* section that asserts `cv.md`, `config/profile.yml`, `modes/_profile.md`, `modes/_custom.md`, `article-digest.md`, `portals.yml`, `data/`, `reports/`, `output/` and `interview-prep/` are all still ignored. Two warnings, both environment-only and present on unmodified `main` too: no Go compiler in PATH so the dashboard build was skipped, and the plugin symlink-traversal test skipped on `EPERM` because unprivileged Windows cannot create symlinks.

## Scope note

`.gitignore` is not in `SYSTEM_PATHS`, so this reaches new clones and anyone who pulls, but not existing installs via `update-system.mjs apply` — that is #2277, and it is not something this PR tries to solve. `.npmignore:77` has the same gap; `package.json` is `private: true` with no `files` field so nothing publishes today, and I left it alone to keep this to one concern. Happy to send it separately.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

## Summary by CodeRabbit

* **Chores**
  * Updated ignored-file rules to exclude backup files with standard and timestamped `.bak` extensions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -85,8 +85,12 @@ bun.lock
 # Report-number reservation sentinels (ephemeral, created by reserve-report-num.mjs)
 reports/*-RESERVED.md
 
-# Backup files written by normalize-statuses.mjs / dedup-tracker.mjs
-*.bak
+# Backup files. Two shapes exist: `{file}.bak` (normalize-statuses.mjs,
+# dedup-tracker.mjs, tracker.mjs) and `{file}.bak-{ISO timestamp}`
+# (web/src/lib/core/safe-write.ts). A plain `*.bak` glob misses the timestamped
+# form, which for cv.md or portals.yml leaves a file holding the same PII as the
+# original sitting untracked in the repo root. Trailing `*` covers both.
+*.bak*
 
 # OpenRouter runner — generated data (personal, never commit)
 data/model-blacklist.json


santifer:
Good catch, @nhan4013. The part that makes this worth merging as-is is the comment: naming both shapes and the scripts that write each one is what stops the next person from "tidying" `*.bak*` back to `*.bak` and quietly reopening the hole. A `cv.md.bak-<ISO>` sitting untracked in the repo root holds exactly the same personal data as the original, and a reflexive `git add .` was all it took.

Verified before merging: no tracked file in the repo matches `*.bak` (control: 351 tracked `.mjs`), so the broader glob costs us nothing. Suite green on main after the merge.

Heads-up so it doesn't surprise you: @slegarraga opened #2560 for the same issue with `.npmignore` and a test. Yours went first and lands first; I'll rebase theirs down to the parts yours doesn't cover, so both pieces get in. Merged 🚀

PR #2546, opened 2026-08-05
docs(_shared): clarify Playwright parallel rule is per-session, not per-agent-count

### LINKED ISSUE #2545: modes/_shared.md: Playwright parallel-agent rule reads as agent-count, not shared-session
### What happened?

`modes/_shared.md`'s Tools table has this rule:

> **NEVER 2+ agents with Playwright in parallel.**

Read literally, this reads as a raw agent-count limit. It doesn't say *why*, and it doesn't say what actually breaks — so it's easy to follow a different, correct-sounding instruction elsewhere (`modes/pipeline.md` Step 3: *"If there are 3+ pending URLs, launch agents in parallel... at most one agent per pending URL"*) without registering that doing so, for URL-based evaluations, means N agents each independently calling the interactive Playwright/MCP tools — which shares one browser session across all of them.

I hit this running `/career-ops pipeline` against 4 pending URLs today. Two of the four parallel workers reported that a `browser_navigate`/`browser_snapshot` call transiently returned a *different* worker's page mid-task — clear evidence of a shared session being raced by concurrent navigations. For this occurrence, both workers happened to notice the title/URL mismatch and re-verified before writing anything, so no bad data landed this time, but a worker that doesn't notice would silently evaluate the wrong posting. In fact, there was an earlier occurrence where several URLs were incorrectly deemed to be duplicates of one another, and I had to re-run the evaluations from scratch.

### Root cause

The rule is real and correct, but its wording doesn't make the actual constraint (shared session) explicit — it reads as "count your agents," not "check whether they share a browser." Notably, most of the market-mode translations already independently phrase it more precisely than the English canonical source does — e.g. `modes/de/_shared.md`: *"sie teilen sich eine Browser-Instanz"* (they share one browser instance), same in `es`/`fr`/`hi`/`id`/`it`/`ko`/`nl`/`pl`/`pt`/`tr`/`da`. The English source (and a literal untranslated copy of the old string in `modes/ja/_shared.md`) is the outlier that never got this clarification. `ru`/`ua`/`ar` also still carry the terser, ambiguous phrasing.

### Proposed fix (this issue)

Update the English canonical `modes/_shared.md` to state the constraint explicitly: it's a per-session rule, not a per-agent-count one. PR incoming with that change.

### Related, larger issue

This is a narrow documentation fix for the ambiguity itself. There's a deeper design question — how should `modes/pipeline.md`'s parallel-worker pattern actually stay correct given this constraint, especially since a very similar failure mode was already hit and fixed once before in the headless batch path (#506 / PR #809, which isolates batch workers from inherited MCP entirely via `--strict-mcp-config`). I'll file that as a separate issue (EDIT: filed #2551) since it needs its own design discussion, not a wording tweak.

### CLI tool

Claude Code

### OS and Node.js version

macOS / Node 24

clede:
## Summary
- `modes/_shared.md`'s Playwright rule ("NEVER 2+ agents with Playwright in parallel") reads as a raw agent-count limit and doesn't state the actual constraint — concurrent access to one *shared* browser/MCP session. Spells that out explicitly, and adds the same clarification inline in the Subagent delegation section, since parallel `pipeline` workers are named right there as an example.
- Most market-mode translations (`de`, `es`, `fr`, `hi`, `id`, `it`, `ko`, `nl`, `pl`, `pt`, `tr`, `da`) already independently phrase this as "they share the same browser instance" — this brings the English canonical source in line with what those translations already say, rather than the other way around.
- `ja`/`ru`/`ua`/`ar` still carry the old, terser phrasing (`ja` is actually an untranslated copy of the old English string) — left untouched here since translating them accurately is out of scope for this fix. Flagging in case a translation-fluent contributor wants to pick those up.

Fixes #2545.

## Test plan
- [x] `node test-all.mjs` — 2975 passed, 0 failed, 2 pre-existing/expected warnings (no user data in a fresh checkout, no Go compiler present) unrelated to this change
- [x] Docs-only change to `modes/_shared.md`; no other files reference the old string in the English canonical mode set

## Summary by CodeRabbit

* **Documentation**
  * Clarified that Playwright tasks may run in parallel when each worker uses an isolated browser session.
  * Reinforced that shared interactive browser sessions must not be accessed concurrently.
  * Applied the same session-isolation requirement to delegated Playwright work.

### DIFF
--- a/modes/_shared.md
+++ b/modes/_shared.md
@@ -195,7 +195,7 @@ After detecting archetype, read `modes/_profile.md` for the user's specific fram
 |------|-----|
 | WebSearch | Comp research, trends, company culture, LinkedIn contacts, fallback for JDs |
 | WebFetch | Fallback for extracting JDs from static pages |
-| Playwright | Verify offers (browser_navigate + browser_snapshot). **NEVER 2+ agents with Playwright in parallel.** |
+| Playwright | Verify offers (browser_navigate + browser_snapshot). **NEVER let 2+ agents drive the same Playwright/MCP browser session concurrently.** This is a per-session rule, not a per-agent-count one: agents each holding their own isolated browser session are fine in parallel; agents sharing one interactive MCP browser session are not — they race for control and can silently read or act on each other's page state. |
 | Read | cv.md, _profile.md, article-digest.md, cv-template.html |
 | Write | Temporary HTML for PDF, applications.md, reports .md |
 | Edit | Update tracker |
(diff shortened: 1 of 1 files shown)


santifer:
Sorry for the slow first reply, @clede.

Two lines on modes/_shared.md: swap the agent count for the actual constraint (concurrent access to one shared browser session) and repeat the clarification inline in the subagent delegation section, which is exactly where pipeline's parallel workers get named as the example. What makes it cheap to read: it pulls canonical English up to what 12 market translations already say on their own, rather than the other way round. ja/ru/ua/ar are left alone on purpose (ja is an untranslated copy of the old string) and you flag them in case a contributor with the language wants them.

PR #2552, opened 2026-08-05
feat(scan): AND-groups in title_filter.positive (#2544)

### LINKED ISSUE #2544: scan.mjs: title_filter.positive is substring-only, so it silently drops every title spelling the config does not literally contain (14% of leadership roles on my config)
## Summary

`title_filter.positive` is matched as a case-insensitive **substring**, so a config can only express *exact title spellings*. Real job titles vary in separator and word order, and every variant the list does not literally contain is dropped silently — no warning, no count, nothing in the summary distinguishing "correctly rejected" from "leaked through a spelling gap."

On my config this was dropping **27 of 196 leadership-titled jobs (14%)** from a single source.

## Reproduction

`portals.yml` containing the natural, documented-looking thing:

```yaml
title_filter:
  positive:
    - "Director of Engineering"
    - "Director, Engineering"
    - "Engineering Director"
    - "Senior Director, Engineering"
    - "VP of Engineering"
    - "VP Engineering"
```

Titles that are obviously in scope and are **all silently dropped**:

| Title | Why it misses |
|---|---|
| `Director - Software Engineering` | hyphen, not comma |
| `Director Engineering (Mobile Platform)` | no separator |
| `Senior Manager Software Engineering` | no comma |
| `VP, Software Engineering` | list has `VP of Engineering` / `VP Engineering` |
| `Senior Director, Platform Engineering` | domain word between level and "Engineering" |
| `Director, Data Engineering (AI Native)` | same |
| `Head of Growth Engineering` | same |
| `Sr Manager Payments Engineering` | same |

The failure is combinatorial: `{Director, Sr Director, VP, Head of} × {,  - , of, ∅} × {optional domain word}`. No hand-maintained list converges.

## Why this is worse than it looks

The scan summary reports `Filtered by title: N removed` as a single number. A user cannot tell a well-tuned filter from a broken one — both just print a big number. In my case the config looked careful and thorough, and had been quietly discarding a seventh of all matching roles for as long as it existed.

This is the same failure class as a bug I hit earlier: a bare `"CTO"` in `negative` matches the substring inside `Dire-**CTO**-r`, silently rejecting every Director role. Substring matching is simply too blunt for this field, in both directions.

## Proposed fix

Let a positive entry be an **AND-group**: if it contains `+`, every `+`-separated term must appear somewhere in the title, in any order. Entries without `+` keep their current behavior, so this is fully backward compatible.

```js
export function buildTitleFilter(titleFilter) {
  const positive = (titleFilter?.positive || [])
    .map(k => k.toLowerCase().split('+').map(s => s.trim()).filter(Boolean))
    .filter(terms => terms.length > 0);
  const negative = (titleFilter?.negative || []).map(k => k.toLowerCase());

  return (title) => {
    const lower = title.toLowerCase();
    const hasPositive = positive.length === 0
      || positive.some(terms => terms.every(t => lower.includes(t)));
    const hasNegative = negative.some(k => lower.includes(k));
    return hasPositive && !hasNegative;
  };
}
```

The six-entry config above collapses to intent, and stops leaking:

```yaml
positive:
  - "director + engineering"
  - "director + software"
  - "vp + engineering"
  - "head of + engineering"
  - "manager + engineering"
```

## Measured results

Same 102 companies, 12,561 jobs, one scan each:

| Config | New offers | Leadership titles dropped (Built In, n=196) |
|---|---|---|
| Enumerated spellings (before) | 14 | 27 |
| Bare anchors, no AND (`"director"`, `"vp"`) | 653 | 0, but overwhelming function noise |
| **AND-groups (proposed)** | **165** | **0** |

With AND-groups, **100% of the 174 titles that passed contained an engineering discipline word** — the precision came back without the leakage, because the config can finally say "a leadership level AND an engineering discipline" instead of guessing at punctuation.

`test-all.mjs`: 277 passed / 10 failed both before and after the patch on my checkout — no regressions attributable to the change. (The 10 are unrelated, from running v1.11.0 against newer tests.)

## Secondary note

Because bare anchors alone produce heavy noise, the `negative` list ends up carrying real weight, and it has the symmetric footgun. Two false negatives I hit while tuning:

- a bare `Finance` dropped `Sr. Director, Software Engineering - Finance Tech`
- `Technical Support` dropped `Director, Forward Deployed Engineering & Technical Support`

Might be worth documenting in `templates/portals.example.yml` that negative entries should be multi-word phrases, and/or adding a `--explain-filter` flag that prints which rule rejected each title. Happy to open that separately.

## Environment

- career-ops v1.11.0
- Node v25.9.0, macOS
- 102 tracked companies, Greenhouse/Ashby/Lever providers plus one local parser

Happy to send a PR with the `buildTitleFilter` change plus unit tests for the AND-group parsing and the backward-compat path.

Zoubeir23:
Fixes #2544.

## The problem, in the reporter's words

`title_filter.positive` is a case-insensitive **substring** match, so an entry can express one exact spelling and nothing else. Real titles vary in separator and word order, and every variant the list does not literally contain is dropped **silently** — the scan summary prints a single `Filtered by title: N` that cannot tell a well-tuned filter from a leaking one. Measured impact on the reporter's config: **27 of 196 leadership titles lost**, from a list that looked careful and thorough.

## Change

`" + "` between terms means every term must appear, in any order:

```yaml
positive:
  - "director + engineering"
```

covers, in one entry, what no hand-maintained list of literal spellings converges on:

| Title | Before | After |
|---|---|---|
| `Director of Engineering` | ✅ | ✅ |
| `Director - Software Engineering` | ❌ | ✅ |
| `Director Engineering (Mobile Platform)` | ❌ | ✅ |
| `Senior Director, Platform Engineering` | ❌ | ✅ |
| `Director of Sales` | ❌ | ❌ |

## One deliberate deviation from the proposed patch

The issue proposes `k.split('+')`. **That breaks `C++`**: it becomes the single term `c`, which matches nearly every title — trading a silent drop for a silent flood, in a config the user has no reason to revisit.

The separator therefore requires surrounding whitespace (`/\s+\+\s+/`), which is also the syntax the issue documents in its own examples. `"C++"` stays a literal keyword; `"director + engineering"` is a group.

Two further details the patch in the issue drops, both preserved here:

- **Short terms keep their word boundary.** `compileKeyword` anchors 2-3 letter keywords, so `"vp + engineering"` matches `VP, Engineering` but not `Revamp Engineering Process`. Each term inside a group is compiled through the same function rather than a raw `includes`.
- **Groups are positive-side only.** On the negative side an entry is a veto, and `" + "` there would mean "reject when both appear" — a different feature, and one that is already expressible as two entries.
- **`matchedTitleKeywords()` reports the group by its raw config string**, so `content_filter.by_title_keyword` overrides keep matching the same strings users wrote.

## Backward compatibility

An entry without `" + "` goes through the existing `compileKeyword` path unchanged. Pinned by a test: `"engineering manager"` still matches `Engineering Manager, Payments`, still rejects `Manager, Engineering`, and the negative list still vetoes.

## Test plan

- [x] 6 new assertions in `test-all.mjs` §11b, including all four separator/word-order variants from the issue, the `C++` guard, the short-term word boundary, and the backward-compatibility case
- [x] `node test-all.mjs --quick` — 2982 passed, 0 failed
- [x] `node validate-portals.mjs --file templates/portals.example.yml` — 0 errors, 0 warnings
- [x] Syntax documented where users meet it: the `title_filter` block in `templates/portals.example.yml`

## Summary by CodeRabbit

* **New Features**
  * Combine positive title keywords with `" + "` to require every term to match.
  * Matching is case-insensitive and supports flexible spacing and term order.
  * Existing acronym boundaries and literal terms such as `C++` remain supported.
* **Documentation**
  * Added guidance and examples for multi-term title filters and keyword behavior.
* **Bug Fixes**
  * Improved title-filter matching and reporting for content-filter overrides.

### DIFF
--- a/scan.mjs
+++ b/scan.mjs
@@ -92,16 +92,57 @@ export function compileKeyword(kw) {
   return (lower) => lower.includes(kw);
 }
 
+// An AND-group: " + " (whitespace-delimited) between terms means EVERY term
+// must appear in the title, in any order. `title_filter.positive` is otherwise
+// substring-only, so it can express exact spellings and nothing else — and real
+// titles vary in separator and word order:
+//
+//   "Director of Engineering" misses  Director - Software Engineering
+//                                     Director Engineering (Mobile Platform)
+//                                     Senior Director, Platform Engineering
+//
+// The combinations are {level} x {, - of none} x {optional domain word}: no
+// hand-maintained list of literal spellings converges, and every miss is
+// silent — the summary reports one "filtered by title" count that cannot tell
+// a well-tuned filter from a leaking one (#2544).
+//
+// The separator REQUIRES surrounding whitespace on purpose. A bare split('+')
+// would turn the perfectly ordinary keyword "C++" into "c", which matches
+// almost every title — trading a silent drop for a silent flood.
+const AND_SEPARATOR = /\s+\+\s+/;
+
+/**
+ * Compile one `positive` entry into a matcher.
+ *
+ * Entries without " + " keep their exact previous behaviour, so existing
+ * configs are unaffected.
+ *
+ * @param {string} keyword - already trimmed and lowercased.
+ * @returns {(lower: string) => boolean}
+ */
+export function compilePositiveKeyword(keyword) {
+  if (!AND_SEPARATOR.test(keyword)) return compileKeyword(keyword);
+  const terms = keyword.split(AND_SEPARATOR).map(t => t.trim()).filter(Boolean);
+  if (terms.length === 0) return compileKeyword(keyword);
+  // Each term keeps compileKeyword's own rule, so a short term like "vp" is
+  // still matched on a word boundary and cannot hit "vp" inside another word.
+  const matchers = terms.map(compileKeyword);
+  return (lower) => matchers.every(m => m(lower));
+}
+
 export function buildTitleFilter(titleFilter) {
   // Normalize defensively: a malformed title_filter (a null, numeric, or otherwise
   // non-string entry in the YAML) must not crash the scan via k.toLowerCase().
-  const normalize = (arr) => (Array.isArray(arr) ? arr : [])
+  const normalize = (arr, compile) => (Array.isArray(arr) ? arr : [])
     .filter(k => typeof k === 'string')
     .map(k => k.trim().toLowerCase())
     .filter(k => k.length > 0)
-    .map(compileKeyword);
-  const positive = normalize(titleFilter?.positive);
-  const negative = normalize(titleFilter?.negative);
+    .map(compile);
+  // AND-groups are a POSITIVE-side feature only. On the negative side an entry
+  // is a veto, and " + " there would read as "reject when both appear", which
+  // is a different and much easier thing to write as two entries.
+  const positive = normalize(titleFilter?.positive, compilePositiveKeyword);
+  const negative = normalize(titleFilter?.negative, compileKeyword);
 
   return (title) => {
     const lower = (title || '').toLowerCase();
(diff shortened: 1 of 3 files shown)


santifer:
Another solid one, @Zoubeir23. What sells this isn't the feature, it's the measurement in the issue: **27 of 196 leadership titles lost** from a filter list that looked careful. A substring-only match can only ever express one exact spelling, and every variant it misses disappears behind a single `Filtered by title: N` that can't distinguish a well-tuned filter from a leaking one.

The compatibility work is what makes it mergeable as-is: requiring whitespace around the separator keeps `"C++"` a literal keyword, an entry without `" + "` goes through the existing path untouched, and you pinned that with a test rather than asserting it in the description. Merged! 🚀

PR #2481, opened 2026-08-04
feat(reposts): add --help usage block and early exit to detect-repost…

itsokAsh:
What does this PR do?

Adds standard `--help` / `-h` support to `detect-reposts.mjs`, mirroring the pattern used in `company-history.mjs`.
When invoked with `--help` or `-h`, the script prints the usage information and exits with status code `0` before reading any data files. Existing behavior for all other invocations is unchanged.

Related issue

Fixes #2479

Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation / translation
- [ ] Refactor (no behavior change)

Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/santifer/career-ops/blob/main/CONTRIBUTING.md)
- [x] If this is a new feature or architecture change, I opened an issue first (bug fixes, providers, docs & translations are exempt — send those straight in)
- [x] My PR does not include personal data (CV, email, real names, scan results, or pipeline data)
- [x] I ran `node test-all.mjs` and all tests pass
- [x] My changes respect the [Data Contract](https://github.com/santifer/career-ops/blob/main/DATA_CONTRACT.md) (no modifications to user-layer files)
- [x] My changes align with the [project roadmap](https://github.com/santifer/career-ops/discussions/156)

## Summary by CodeRabbit

* **New Features**
  * Added `--help` and `-h` options to display command usage information.
  * Help text lists supported summary, window, self-test, and help options.

* **Tests**
  * Added coverage for help output using both supported flags.

PR #2555, opened 2026-08-06
fix(cv): ORG/YEAR empty fallbacks never emitted, breaking column alignment

### LINKED ISSUE #2486: fix(cv): ORG_EMPTY/YEAR_EMPTY alignment fallbacks never emit — parsePartial suffix mismatch (certifications + awards)

tarang-tj:
Closes #2486.

## The mismatch

`parsePartial` (`build-cv-html.mjs:220`) derives a block's base name with `name.slice(0, -6)`, stripping the literal `_EMPTY`. So `<!--ORG_EMPTY-->` resolved to base `ORG`, an orphan, while the renderer resolves `{{ORG_BLOCK}}`. The alignment fallbacks in `certifications.html` and `awards.html` never fired.

Took the rename approach from the issue, applied to both partials, and corrected the header comments that still named the old markers.

## Why it survived a test

The `build-cv-html` self-test already exercised the mixed-rows case, but asserted:

```js
if (!certHtml.includes('class="cert-org"'))
```

The fixture's second entry has an org, so that substring is present whether or not the fallback fires. The check could not fail for the case it covered. It now asserts the empty span itself.

## Measured

One row missing org/year, one row with both:

| span | before | after |
|---|---|---|
| `cert-org` | 1 span, 0 empty | 2 spans, 1 empty |
| `cert-year` | 1 span, 0 empty | 2 spans, 1 empty |
| `award-org` | 1 span, 0 empty | 2 spans, 1 empty |
| `award-year` | 1 span, 0 empty | 2 spans, 1 empty |

## Test

`tests/cv-alignment-fallbacks.test.mjs` drives the real builder with the mixed-rows profile and asserts on the span **count** as well as the empty span, so it fails if either row stops rendering. Verified red before the fix (4 failures) and green after.

`node test-all.mjs --quick`: **2977 passed, 0 failed**.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved awards and certifications rendering when organization or year details are missing.
  - Empty fields now display consistently without incorrect placeholder text or visible artifacts.
  - Ensured organization and year information remains correctly aligned across entries with mixed or incomplete details.

- **Tests**
  - Added coverage for awards and certifications containing missing organization and year values.

### DIFF
--- a/build-cv-html.mjs
+++ b/build-cv-html.mjs
@@ -901,8 +901,13 @@ async function runSelfTest() {
     console.error(`Self-test failed (cert variant): ${err.message}`);
     process.exit(1);
   }
-  // The partial should emit an empty <span class="cert-org"> for alignment.
-  if (!certHtml.includes('class="cert-org"')) {
+  // The partial should emit an EMPTY <span class="cert-org"></span> for the
+  // org-less row, so the table columns stay aligned. Assert the empty span
+  // specifically: a bare `includes('class="cert-org"')` is satisfied by the
+  // second entry (which has an org), so it passes even when the fallback never
+  // fires -- which is exactly how the ORG_EMPTY/ORG_BLOCK_EMPTY suffix
+  // mismatch shipped unnoticed.
+  if (!certHtml.includes('<span class="cert-org"></span>')) {
     console.error('Self-test failed: cert-org empty-block not emitted for table alignment');
     process.exit(1);
   }
--- a/templates/sections/awards.html
+++ b/templates/sections/awards.html
@@ -4,9 +4,9 @@
   Field placeholders filled per entry:
     TITLE      — award name (e.g. "Gold Medal, International Olympiad in Informatics")
     ORG_BLOCK  — conditional: award-org span with the issuing body when present;
-                 falls back to empty award-org span (ORG_EMPTY) for table-cell alignment
+                 falls back to empty award-org span (ORG_BLOCK_EMPTY) for table-cell alignment
     YEAR_BLOCK — conditional: award-year span with year text when present;
-                 falls back to empty award-year span (YEAR_EMPTY) for alignment
+                 falls back to empty award-year span (YEAR_BLOCK_EMPTY) for alignment
 
   Template packs: edit tag names, class names, and field order freely inside the ENTRY zone.
 -->
@@ -17,7 +17,7 @@
   {{YEAR_BLOCK}}
 </div>
 <!--ORG_BLOCK--><span class="award-org">{{ORG}}</span><!--/ORG_BLOCK-->
-<!--ORG_EMPTY--><span class="award-org"></span><!--/ORG_EMPTY-->
+<!--ORG_BLOCK_EMPTY--><span class="award-org"></span><!--/ORG_BLOCK_EMPTY-->
 <!--YEAR_BLOCK--><span class="award-year">{{YEAR}}</span><!--/YEAR_BLOCK-->
-<!--YEAR_EMPTY--><span class="award-year"></span><!--/YEAR_EMPTY-->
+<!--YEAR_BLOCK_EMPTY--><span class="award-year"></span><!--/YEAR_BLOCK_EMPTY-->
 <!--/ENTRY-->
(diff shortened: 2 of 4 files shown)


santifer:
@tarang-tj this is a good first PR and I'm sorry to close it. Your diagnosis was correct: `parsePartial` derived the base name with `slice(0, -6)`, so `<!--ORG_EMPTY-->` resolved to `ORG` while every builder writes `ORG_BLOCK`, and the fallback silently never emitted. You found that from the code, which is the hard part.

Two reasons it goes to #2497 instead, and the second is worth keeping:

**Timing.** #2497 was opened on Aug 4, about 43 hours before yours. It sat unanswered until today, which is my fault, not yours: had I been quicker, #2486 would have been assigned and you would never have spent the evening on it.

**Route.** Issue #2486 left two doors open and each of you took one. You renamed the markers in the two partials we ship; #2497 taught the parser to accept both spellings. They fix the same visible bug, but anyone running a template pack that copied the `ORG_EMPTY` naming (packs are meant to be forkable) stays broken under the rename and gets repaired under the parser fix. That is the only thing separating the two, and it is not something you could have known from the issue text.

If you want another one, **#2477 is free and yours if you say so**: add a `--help` usage block to `followup-cadence.mjs`. An identical PR landed three days ago (#2481, on `detect-reposts.mjs`), so there's a reviewed pattern to mirror, and it is small enough that the review is about your code rather than about learning the repo. Comment there and I'll assign it to you immediately.

Come back. Reading a parser until you find a `slice(0, -6)` that's off by a naming convention is exactly the instinct this project needs.

santifer:
@tarang-tj, when this closed we said the door was open and then never pointed you anywhere: fixing that now. Your ORG/YEAR fallback catch here was sharp work, honestly above good-first-issue level, so rather than hand you a starter task: the free pool is at https://github.com/santifer/career-ops/issues?q=is%3Aopen+label%3A%22good+first+issue%22+no%3Aassignee (claim with `/assign`), and if you want something meatier in the CV/PDF area, say the word and I'll match you to one.

tarang-tj:
@santifer — I took you up on the meatier CV/PDF offer and opened #3433 for #2505.

It moves web PDF mode onto the canonical payload → build-cv-html.mjs path, with fail-closed validation, cleanup/lifecycle coverage, a real end-to-end run, and measured output reduction. I also called out the exact limit of the byte-equivalence evidence rather than overstating it.

Thanks again for the direct feedback here. I used it: this time I checked live ownership and adjacent work first, and I’m keeping the implementation and evidence in one reviewable change.

santifer:
Read every word of this, @tarang-tj, and the line that matters is the one you slipped in at the end: *"this time I checked live ownership."*

That's the thing I actually care about. The feedback here was blunt, and instead of arguing it or going quiet you changed how you pick what to work on. #3433 is a bigger, harder piece than this one was, and calling out the exact limit of your byte-equivalence evidence rather than overstating it is what makes it reviewable.

It's in the queue and you'll get a real read on it, not a rubber stamp.

PR #2556, opened 2026-08-06
feat(cli): add --help / -h usage block to stats.mjs

### LINKED ISSUE #2476: feat(cli): stats.mjs has no --help — add the usage block
**Time:** ~25 min · **Difficulty:** Beginner · **Reserved for @lakshaydevoo-arch** (comment below to lock the assignment; if no comment in 72h it's open to anyone)

## Context

`node stats.mjs` aggregates lifetime pipeline stats (tracker, scan history, portals, follow-ups, runs) into one JSON report. Today it has no usage text at all: `node stats.mjs --help` silently ignores the flag and runs the full aggregation (`grep -c "Usage" stats.mjs` → 0).

## What to do

Mirror the pattern in `company-history.mjs` (lines 83–98): a `KNOWN_FLAGS` list, a usage block naming each flag, and an early `--help` / `-h` exit **before any file is read**.

## Acceptance

- `node stats.mjs --help` (and `-h`) prints the usage block and exits 0 without reading any data file.
- The usage block documents every flag the script actually parses today (check the source, not the docs).
- No behavior change for existing invocations.

rajashekarpatha07:
Closes #2476

## Issue

`stats.mjs` had no usage text — `node stats.mjs --help` silently ignored
the flag and ran the full aggregation.

## What Changed

- Added `KNOWN_FLAGS` list and `USAGE` block documenting every flag the
  script currently parses (`--summary`).
- Added early `--help` / `-h` exit before any data file is read.
- Mirrors the established pattern from `company-history.mjs` (lines 83–101).

## Verification

- `node stats.mjs --help` → prints usage, exits 0 ✅
- `node stats.mjs -h` → prints usage, exits 0 ✅
- `node stats.mjs` / `node stats.mjs --summary` → unchanged behavior ✅
- `tests/stats.test.mjs` and `tests/cli-flags.test.mjs` → all 34 tests pass ✅

## Summary by CodeRabbit

* **New Features**
  * Added `--help` and `-h` options to display CLI usage instructions and supported options.
  * Added `--summary` for human-readable results.
  * JSON output remains the default format.
  * Help requests exit successfully without running the analysis.


slegarraga (changes requested):
The help flag is a good addition. Three small items before merge:

1. `process.exit(0)` immediately after `console.log(USAGE)` can truncate piped stdout; restructure the direct-run flow so help prints and skips `computeAllStats()` without forcing termination.
2. `USAGE` documents `--help` but not the `-h` alias it accepts.
3. Add child-process smoke tests for both `--help` and `-h`: assert usage output, exit 0, and no data-file reads.

Happy to re-review after that.

rajashekarpatha07:
Thanks for the review! All three items addressed:

1. Replaced `process.exit(0)` with if/else — help prints and skips `computeAllStats()` without forced termination.
2. USAGE now documents the `-h` alias (`--help|-h`).
3. Added child-process smoke tests for both `--help` and `-h` in `tests/stats.test.mjs` — asserting usage output and exit 

All 36 tests pass (18 cli-flags + 18 stats including the 2 new ones).

slegarraga (approved):
Thanks for addressing all three items. The `--help`/`-h` path prints usage without computing stats, the usage block documents both aliases, and the smoke tests cover `--help`, `-h`, and the summary path. I also noticed the matching help block in `followup-cadence.mjs`; that reads well too. If you want the exit behavior fully consistent, the same if/else shape used in `stats.mjs` works there, but that is not blocking.

rajashekarpatha07:
Thanks for the approval and the suggestion! 🙌
I went ahead and updated `followup-cadence.mjs` as well — replaced `process.exit(0)` with the same if/else pattern from `stats.mjs` and documented the `-h` alias in USAGE. Both scripts are now fully consistent. All tests pass.

santifer:
Welcome to career-ops, @rajashekarpatha07 — merged, and thank you for your patience: this sat longer than it should have. You matched the existing help-block pattern rather than inventing a new one, which is exactly what makes a CLI feel coherent instead of like a pile of scripts. Your first contribution here. Glad to have you.

PR #2558, opened 2026-08-06
fix(seeds): walk every YC page instead of stopping after page 1

### LINKED ISSUE #2525: --seeds yc only probes ~25 companies since the YC API changed pagination

gunjanjaswal:
Fixes #2525.

## What was happening

`fetchYCCompanies()` decided when to stop paging with this heuristic:

```js
const batchSize = Array.isArray(raw?.companies) ? raw.companies.length : 0;
if (batchSize < 1000) break;
```

That assumed a page returning fewer than 1000 companies meant "last page." The YC API now caps its page size (~30 today) and ignores `per_page=1000`, so page 1 comes back well under 1000 and the walk stops immediately. The only companies the scan ever sees are the newest batch, which don't have ATS boards yet — hence `--seeds yc` probing ~25 companies and reporting 0 matches against a wall of 404s.

## The fix

Follow the API's own pagination instead of guessing from batch size. The response already carries `page`, `totalPages`, and `nextPage`, so the walk continues while the API says there are more pages and stops at the reported last page (or when there's no next page). Companies are still deduped by slug across pages.

`maxPages` stays as a safety cap (raised to a generous default; the real stop is the API's `totalPages`). No caller passes `maxPages` — `scan-ats-full.mjs` calls `source.fetch()` with no args — so this simply restores full YC coverage for the scan without touching the call sites.

## Tests

New `tests/vc-portfolios-yc-pagination.test.mjs` stubs `global.fetch` to serve small paginated batches and covers:

- walking every page reported by `totalPages` (not stopping after page 1),
- the `nextPage` fallback when `totalPages` is absent,
- cross-page dedupe by slug,
- the `maxPages` safety cap.

Evidence (reverting only the stop condition to the old `batchSize < 1000` heuristic):

```
WITH fix:     4 passed
WITHOUT fix:  4 failed — each stops on page 1 with 2 companies
```

The change is confined to the YC fetcher, and the new suite plus the existing seed/scan tests pass under `node test-all.mjs --quick`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved company data retrieval across paginated results, including non-sequential pages and varied pagination formats.
  * Prevented missing results when pages contain fewer than 1,000 companies.
  * Added safeguards to stop pagination at the reported endpoint or after 500 pages.
  * Preserved partial results when a later page cannot be retrieved.
  * Removed duplicate companies appearing across multiple pages.

* **Tests**
  * Added coverage for pagination metadata, page limits, deduplication, and pagination input handling.

### DIFF
--- a/seeds/vc-portfolios.mjs
+++ b/seeds/vc-portfolios.mjs
@@ -41,8 +41,9 @@ const DEFAULT_USER_AGENT = 'Mozilla/5.0 (compatible; career-ops-seeds/1.0)';
 // at all. Sized well clear of the real total on purpose - a ceiling that sits
 // close to today's page count stops being a guard and becomes a silent truncation
 // the day the catalogue grows past it, which is the exact failure this file was
-// fixed for.
-const YC_MAX_PAGES = 500;
+// fixed for. Exported so the walk is clamped to it as a hard ceiling and tests
+// can assert the contract.
+export const YC_MAX_PAGES = 500;
 
 /**
  * YC public company API.
@@ -318,9 +319,11 @@ export function toPortalEntry(company) {
  * Uses the public YC API (no auth, no API key). The response is a JSON object
  * with a `companies` array plus `page`/`totalPages` pagination fields. The API
  * caps page size server-side (~30/page; `per_page` is ignored), so the whole
- * portfolio is walked page by page up to `maxPages`, newest batches first.
+ * portfolio is walked page by page, newest batches first.
  *
- * @param {{ timeoutMs?: number, maxPages?: number }} [opts]
+ * @param {{ timeoutMs?: number, maxPages?: number }} [opts] - `maxPages` is
+ *   clamped to YC_MAX_PAGES, which is a hard ceiling; pagination normally stops
+ *   at the API-reported last page.
  * @returns {Promise<SeedCompany[]>}
  */
 export async function fetchYCCompanies({ timeoutMs = DEFAULT_TIMEOUT_MS, maxPages = YC_MAX_PAGES } = {}) {
@@ -328,7 +331,12 @@ export async function fetchYCCompanies({ timeoutMs = DEFAULT_TIMEOUT_MS, maxPage
   const all = [];
   const seen = new Set();
 
-  for (let page = 1; page <= maxPages; page++) {
+  // YC_MAX_PAGES is a hard ceiling: clamp here so an explicit maxPages (or a
+  // stray Infinity) can never spin the walk past the runaway guard.
+  const limit = Math.min(maxPages, YC_MAX_PAGES);
+
+  let page = 1;
+  for (let fetched = 0; fetched < limit; fetched++) {
     const url = `https://api.ycombinator.com/v0.1/companies?page=${page}&per_page=1000`;
     let payload;
     try {
@@ -340,7 +348,7 @@ export async function fetchYCCompanies({ timeoutMs = DEFAULT_TIMEOUT_MS, maxPage
     }
 
     const entries = parseYCPayload(payload);
-    if (entries.length === 0) break; // No more pages.
+    if (entries.length === 0) break; // No more companies.
 
     for (const e of entries) {
       if (!seen.has(e.slug)) {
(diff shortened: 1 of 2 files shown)


gunjanjaswal:
Addressed the review notes:

- **nextPage was only read as a presence flag** — the fallback now follows the actual `nextPage` target via a new `parseYCNextPage()` that reads a forward page number from a number, a `page=N` fragment, or a full URL, so a non-sequential nextPage is honoured. `totalPages` stays the primary sequential path.
- **Strengthened the test** so page 1 points at page 3 and asserts the walk requests `1,3` (never 2) — it fails if nextPage is treated as a boolean — plus a unit test for `parseYCNextPage` across the number/fragment/URL shapes.

gunjanjaswal:
Addressed the latest note too: `YC_MAX_PAGES` is now a hard ceiling — the loop clamps to `min(maxPages, YC_MAX_PAGES)`, so an explicit `maxPages` (or a stray `Infinity`) can't spin the walk past the runaway guard. Added a regression test that runs `maxPages: Infinity` against an always-`nextPage` API and asserts it stops at the cap.

The two earlier notes (following the `nextPage` target, and testing its value rather than presence) were handled in the previous push.

gunjanjaswal:
Good catch — pinned it. The clamp test now compares against a test-owned `EXPECTED_YC_MAX_PAGES = 500` and separately asserts the exported constant equals 500, so raising the cap fails the test instead of quietly moving its own expectation.

santifer:
@gunjanjaswal, this needs an honest explanation and a credit, in that order.

**What happened.** #2526 addressed the same issue and arrived 26 hours before yours. It's that person's first PR here. It had the pagination fix right, but it also carried a filter dropping every company not marked `Active`, and that part was a real problem: I measured it against the live API and it removes about a third of the catalogue (roughly 2,000 companies), which hire at nearly the same rate as the active ones (3 of 22 with a live board versus 4 of 22 in the `Active` control group, and one of the discarded ones had 51 open roles). I removed that filter myself on their branch and merged it, rather than sending it back and leaving the fix out for another week.

**Now the credit, which is the part I want on the record:** editing that branch, **I used your page-ceiling design**. Their version set 300 while the API reports 246 today, which would have gone back to truncating silently as the catalogue grew: exactly the failure the PR exists to fix. I used your number, 500, with your name, `YC_MAX_PAGES`, and with your reasoning about why a ceiling sitting close to the real total stops being a guard. That part of `main` today is yours, and I'd rather say so here than leave it buried in a commit.

**What your PR still has that `main` doesn't**, which is why I'm asking you not to close it:

1. **The clamp.** I set the constant as the default but not the `Math.min(maxPages, YC_MAX_PAGES)`. Right now a caller passing `maxPages: Infinity` walks straight past the guard, which is the case you wrote it for. (`grep -n "Math.min(maxPages" seeds/vc-portfolios.mjs` on main: no match.)
2. **`parseYCNextPage`.** When the API stops reporting `totalPages`, `main` just stops. Your fallback follows whatever `nextPage` the API hands back, tolerating a bare number, a `page=N` fragment or a full URL, and requires forward progress so it can't spin. That's the case that bites us the day YC changes the shape again, which it has already done once. (Not on main either.)
3. **Your test.** There is no test for this walk in `tests/` at all right now.

Rebase down to those three and I'll review it as soon as you push. The conflict will be in the same lines the other PR touched, so if you'd rather I resolve it, say so and I will: you supplied the judgement that ended up in `main`, and you shouldn't also pay the merge tax for it.

I closed another PR of yours earlier today as a duplicate and I know this looks similar. The difference is that I'm not asking you to start over: I'm telling you which three pieces of what you already wrote are still missing.

gunjanjaswal:
Thanks, and no worries on the overlap. #2526 getting the pagination in, plus your catch on that `Active` filter, means people have the fix now, which is the point.

I appreciate you surfacing the `YC_MAX_PAGES` bit too. That's the kind of thing that usually just sits in a commit, so it was good of you to put it on the record.

I've rebased this down to the three pieces `main` is still missing:

1. The `Math.min(maxPages, YC_MAX_PAGES)` clamp (and exported the constant), so a caller passing `Infinity` can't walk past the ceiling.
2. `parseYCNextPage` as the `totalPages`-absent fallback: it follows a bare number, a `page=N` fragment, or a full URL, and requires forward progress so it can't spin.
3. The test, covering the walk, the non-sequential `nextPage` case, dedupe, the parser, and the hard cap.

It's +204/-8 across the two files and mergeable now. I took the rebase so you don't have to. Ready when you are, and thanks again.

santifer:
Another solid one, @gunjanjaswal. The three pieces landed exactly as scoped: the `Math.min` clamp, the `parseYCNextPage` fallback for when the API stops reporting totals, and a regression test that actually exercises both paths. And as said before: the page-cap design in main was already yours by way of #2526's edit — now the rest of it is too. Merged! 🚀

gunjanjaswal:
Thanks for merging this, and for how you handled the whole overlap. Crediting the design and pointing me at exactly the three pieces to keep made it easy to land the useful part without redoing work. Appreciated.

PR #2560, opened 2026-08-06
fix: ignore timestamped user-layer backups

### LINKED ISSUE #2542: bug(gitignore): timestamped user-layer backups escape .gitignore — cv.md PII left untracked in the repo root

slegarraga:
Closes #2542.

Widens the backup ignore pattern from `*.bak` to `*.bak*` in both `.gitignore` and `.npmignore`, covering timestamped `safe-write.ts` backups and numbered variants. Adds regression probes for `cv.md.bak-...`, `config/profile.yml.bak-...`, `portals.yml.bak-...`, and `cv.md.bak10`. Focused test passes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Backup files with timestamps or additional suffixes are now consistently excluded from Git and package distributions.
  - Existing legacy backup-file patterns remain covered.

- **Tests**
  - Added regression coverage for root-level, nested, timestamped, and suffixed backup files.


santifer:
Merged, @slegarraga. Two people found this at once: @nhan4013's #2543 landed first with the `.gitignore` half, so I rebased yours down to the parts theirs didn't cover, which is where your version was genuinely ahead.

Those parts matter. `.npmignore` decides what ships in the published package, so a stale `*.bak` there is the same PII leak with a wider blast radius than a local `git add`. And the test is what stops the glob from being "tidied" back to `*.bak` by someone who reads it as a typo.

I mutation-checked your test before merging rather than trusting it green: reverting `*.bak*` to `*.bak` makes all four probes fail, including the `cv.md.bak10` control. It's a real guard, not decoration. You did nothing wrong on the timing 🚀

slegarraga:
Thank you for the detailed close-out and for rebasing down to the parts that matter. I agree the `.npmignore` guard and the mutation-tested regression are the ones with real blast radius. Appreciate it.

Issue #2426, opened 2026-08-02
i18n: add the interview README to es/fr/de/pt/zh (one language per PR)


santifer:
**Time:** ~25 min · **Difficulty:** Beginner · **Native speaker of the target language needed**

The interview modes live in `modes/interview/` (English) and are translated per language under `modes/{lang}/interview/`. Every language has `plan.md`, `practice.md` and `debrief.md` — but only Italian also has the folder's `README.md`, which is the index a reader lands on first. Spanish, French, German, Portuguese and Chinese are missing it.

### What to do
1. Pick ONE language you speak natively: `es`, `fr`, `de`, `pt` or `zh`.
2. Translate `modes/interview/README.md` into that language and save it as `modes/{lang}/interview/README.md`.
3. Mirror the structure exactly — same headings, same order. Only the prose changes.

### Acceptance
- Natural prose in the target language (not machine-literal).
- Structure matches the English original; file names and links unchanged.
- One language per PR — that keeps review easy and lets several people take this issue.

### Pattern to mirror
`modes/it/interview/README.md` (added in #2335) — that's exactly the shape and quality bar for this task.

realmcastro:
Taking `pt` (pt-BR). Native speaker, and pt-BR is the language I use career-ops in.
Will mirror `modes/fr/interview/README.md` from #2450.

slegarraga:
I will work on the Spanish translation.

zjq-yl:
I am taking the Chinese (`zh`) translation. I will mirror the English structure and keep the existing file names and links unchanged.

slegarraga:
The Spanish translation is ready for review in #2554. It mirrors the English interview README structure and keeps the existing file names and links unchanged.

PR #2563, opened 2026-08-06
fix(jd-skill-gap): the heading detector was English-only, so the check never ran on a French JD

Zoubeir23:
Sent straight in per `CONTRIBUTING.md:22` — happy to convert it to an issue first if you would rather discuss the scope.

## The check never runs on a francophone posting

`REQUIREMENT_HEADER_RE` and `NON_REQUIREMENT_HEADER_RE` are English-only. Same posting body, same skills, only the heading differs:

```
Requirements            ->  Python, Django, PostgreSQL, Docker
Profil recherché        ->  (nothing)
Compétences requises    ->  (nothing)
Exigences               ->  (nothing)
```

The file already documents why this matters, having been widened once for the same reason:

> *Real postings rarely use the word "Requirements". The literal-only list missed the phrasings most modern ATS boards actually ship … so a JD could yield zero skills - which reads identically to "no gaps found" and is the more dangerous of the two failure modes this file warns about.*

Language is that same axis. And the repo ships a `modes/fr/` market set precisely because francophone postings are expected input — a JD is written in its market's language, so for those users this check has never run.

**To be precise about severity:** since #2341 it fails *loudly* — `🚨 LOW CONFIDENCE … reason: no-requirements-section`. So this is an inert check that says so, not the silent pass #2278 described. It is still inert.

## Change

- **French openers**: `profil recherché`, `votre profil`, `compétences (requises|attendues|techniques)`, `compétences`, `exigences`, `pré-requis`, `qualifications requises`, `ce que nous recherchons`, `expérience requise`, `vous êtes`.
- **French closers**: `avantages`, `ce que nous offrons`, `rémunération`, `salaire`, `à propos de …`, `processus de recrutement`, `comment postuler`, `notre stack/processus/mission/équipe`.

**The closers are not decoration.** Without them a French requirements block stays open to end-of-file and sweeps the perks list into required skills — `Tickets Restaurant` and `Mutuelle` reported as gaps. That is exactly what the English closers were added to prevent, and it is covered by a test.

- **Accents are folded at the heading test site only** (`foldAccents`), so `Profil recherché` and `Profil recherche` are one entry instead of two per phrasing. The skill text itself is untouched — pinned by a test asserting a skill keeps its accents.

## Scope

**French only.** The repo's own i18n convention is one language per PR (#2426), and each entry here is a heading I can justify rather than a guess. `es`, `pt`, `de`, `zh`, `ja` are the natural follow-ups now that the accent folding and the opener/closer pairing are in place.

## Test plan

- [x] 7 new self-test cases. **6 fail on `main`** — verified by porting the block onto `main`'s copy of the file
- [x] Both directions: French headings extract the skills, and the French perks section closes the block instead of contributing gaps
- [x] English behaviour unchanged — the existing `Requirements` / `Benefits` case still yields exactly `Python, Django`
- [x] `node jd-skill-gap.mjs --self-test` — 53 passed, 0 failed (46 existing + 7)
- [x] `node test-all.mjs --quick` — 2976 passed, 0 failed

## Summary by CodeRabbit

* **New Features**
  * Added support for French requirement and non-requirement headings.
  * Recognizes both accented and unaccented heading variants.
* **Bug Fixes**
  * Improved skill extraction from French job descriptions.
  * Prevented accented words from being truncated into fragments.
  * Preserved skills ending with punctuation or symbols.
  * Correctly detects the end of French perks and requirement sections.

### DIFF
--- a/jd-skill-gap.mjs
+++ b/jd-skill-gap.mjs
@@ -156,7 +156,17 @@ const BULLET_LINE_RE = /^\s*[-*•]\s*(.+)\r?$/;
 // a word char / # / + so a sentence-ending period is never swallowed into
 // the token ("Docker." still extracts as "Docker"). The leading \b stays:
 // tokens start with [A-Z], a word char, where \b and (?<!\w) are equivalent.
-const SKILL_TOKEN_RE = /\b([A-Z][A-Za-z0-9+.#]{0,29}[A-Za-z0-9+#](?:\.[a-z]{2,4})?)(?!\w)/g;
+// The trailing guard is Unicode-aware. `\w` is ASCII, so an accented letter did
+// not count as "still inside a word" and the token was emitted TRUNCATED:
+// "Maîtrise" yielded "Ma" and "Expérience" yielded "Exp", both reported as
+// required skills. That is over-extraction — the failure this file calls
+// unrecoverable, since it invents gaps the user then tries to close.
+//
+// Only the boundary changed: an accented word is still not extracted as a
+// skill (the opening [A-Z][A-Za-z0-9+.#] stays ASCII on purpose, so ordinary
+// capitalised French prose does not become "required skills"); it is simply no
+// longer emitted as a fragment of itself.
+const SKILL_TOKEN_RE = /\b([A-Z][A-Za-z0-9+.#]{0,29}[A-Za-z0-9+#](?:\.[a-z]{2,4})?)(?![\p{L}\p{N}_])/gu;
 
 // Deliberately broad: this list exists specifically to stop generic
 // capitalized nouns/adjectives from JD bullets (e.g. "Bachelor's degree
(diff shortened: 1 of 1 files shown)


slegarraga (approved):
Solid fix. The French opener/closer pairing is well-reasoned, and the Unicode boundary change prevents truncated accented tokens. The self-tests cover accented and unaccented headings, closers, decomposed text, and the existing Latin behavior. Looks good to me.

santifer:

The heading detector in jd-skill-gap.mjs was English-only, so on a French listing ('Profil recherché', 'Compétences requises', 'Exigences') the check extracted nothing, even though the repo ships a modes/fr/ set precisely because French listings are expected. Without 'avantages' / 'ce que nous offrons' the requirements block runs open to the end of the file and reports Tickets Restaurant and Mutuelle as missing skills, which is exactly what the English closers exist to prevent. Accents are folded only at the point where the heading is tested, with a test pinning that the skill text keeps its own, and the PR stays French-only citing the one-language-per-PR convention (#2426).

PR #2565, opened 2026-08-06
fix(web): make the tracker reader agree with parseTrackerRow on row shape (#2369)

Zoubeir23:
Relates to #2369 — a **4th instance** of that class, and the only one where the web copies core *logic* rather than a constant list.

`web/src/lib/tracker-table.mjs` says it parses "mirroring parseTrackerRow in tracker-parse.mjs". Its alias table is genuinely shared (PR #1598), but its row-shape rules were reimplemented, and they disagree with the core on two row shapes.

## 1. A row missing an interior cell is rendered with every later column shifted

`parseTrackerRow` has an explicit dynamic width guard for exactly this:

> a missing INTERIOR cell shifts every later column one left while the trailing empty cell keeps the count plausible — so require the full width rather than mere coverage of the highest mapped index.

The web reader only required `cells.length >= 8`, so it accepted the row and read every column after the gap one position early.

Realistic trigger: a row written before `merge-tracker.mjs --migrate-via` widened the header, so it carries no `Via` cell.

```
| # | Date | Company | Via | Role | Score | Status | PDF | Report | Notes |
| 13 | 2026-01-02 | Globex | Engineer | 4.0/5 | Applied | ✅ | — | pre-migration |
```

| | core | web (before) |
|---|---|---|
| row #13 | rejected | accepted |
| `role` | — | `4.0/5` |
| `score` | — | `Applied` |
| `status` | — | `✅` |

The dashboard showed a role of "4.0/5" and a status of "✅" — not even a canonical state — for a row every core tool refuses to read.

## 2. A row without a trailing pipe silently loses its last cell

`trackerCells` did an unconditional `line.split("|").slice(1, -1)`. Rows that do not end with a pipe are one part narrower but complete — `tracker-utils rebuildRow` supports them and `parseTrackerRow` computes its width accordingly — so `slice(1, -1)` ate a real cell instead of padding.

```
| 5 | 2026-01-01 | Acme | Engineer | 4.5/5 | Applied | ✅ | — | last note
```

core reads `notes = "last note"`; the web read `notes = ""`.

## The fix

- `trackerCells` drops the trailing part only when the row actually ends with a pipe, matching `parseTrackerRow`'s width rule.
- `parseApplications` requires a cell for every mapped column before reading the row, so a short row is rejected rather than shifted.

Both are read-path only; no write path changes.

## Test plan

Three assertions added to `tracker-columns-tests.mjs` Test 18, next to the existing web-reader tests, and they compare the web reader against the **real** `parseTrackerRow` output rather than a hardcoded expectation — so the two cannot drift apart again silently. They skip cleanly on core-only installs via the existing `HAS_WEB` guard.

- [x] Each assertion verified RED before its fix and GREEN after (the two fixes flip different assertions).
- [x] `node tracker-columns-tests.mjs` — 46 passed, 0 failed.
- [x] `node test-all.mjs` — 2976 passed, 0 failed (the single warning is the pre-existing `Dashboard build skipped — go compiler not in env`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tracker table parsing for rows without trailing delimiters, preserving their final cell.
  * Added validation to reject incomplete rows that could otherwise be incorrectly mapped to headers.
  * Ensured complete Via rows continue to parse correctly.

* **Tests**
  * Added coverage for row-shape handling, incomplete interior cells, complete rows, and final-cell preservation.


santifer:

Fourth instance of class #2369 and the only one where the web copies core logic rather than a list of constants: web/src/lib/tracker-table.mjs claims to mirror parseTrackerRow but reimplements the row-shape rules and diverges in two of them. On a row predating merge-tracker --migrate-via (no Via cell) the core rejects it while the web accepted it via cells.length >= 8 and read every field one column early, so the dashboard showed role '4.0/5' and status '✅', which isn't even a canonical status, for a row no core tool accepts. The second: an unconditional slice(1,-1) ate the last cell on rows without a trailing pipe, so notes rendered empty; the new tests assert against the real parseTrackerRow instead of hand-written expected values, so the two can't drift in silence again.

santifer (approved):
Approved. I reproduced both divergences against `main` and the real core rather than reading the description, and they're exactly as you state.

**Case 1 — row missing an interior cell** (a pre-`--migrate-via` row under a 10-column header):

```
web  (main):  { role: "4.0/5", score: "Applied", status: "✅" }
core        :  null   ← rejected
```

The dashboard rendered a **role of "4.0/5" and a status of "✅"** — a status that isn't even a canonical state — for a row every core tool refuses to read. That's the worst shape a divergence can take: not a crash, not an error, just plausible-looking wrong data.

**Case 2 — row without a trailing pipe:**

```
web  (main):  notes = ""
core        :  notes = "last note"
```

`slice(1, -1)` ate a real cell. And with your fix both now agree: case 1 rejected, case 2 keeps `"last note"`.

## Why this one matters beyond its size

You framed it precisely: this is the **4th instance of #2369, and the only one where the web copies core _logic_ rather than a constant list.** The other three were lists that could be derived or frozen. Logic can't be diffed by eye, which is why this sat undetected while the file's own header comment claimed it was "mirroring parseTrackerRow".

The cure is the right one too: your test **imports `parseTrackerRow` and compares against it**, so it asserts *agreement with the core* rather than a snapshot of today's expected values. If the core's width rule changes, this goes red instead of drifting quietly. That's the difference between a test that preserves a convergence and one that can detect it was never there.

## Verified

- Both cases reproduced on `main`, then confirmed fixed on your branch
- `tracker-columns-tests.mjs`: **43 → 46**, the three new checks run and pass (they're inside the `HAS_WEB` guard, and web is present — I checked they weren't silently skipped)
- `test-all` 3860/0, web 156/156, typecheck clean
- CI green on all three platforms

One note for anyone reading later: the `test-all` total doesn't move, because it runs this file as a subprocess and counts its exit code as one assertion. The new coverage is real — `43 → 46` in the file itself is where you see it.

Merging. Thank you for going after the class rather than the symptom — and for naming which instance number it was, which is what made it obvious this needed the parity treatment rather than a patch.

santifer:
Merged.

Both divergences reproduced against `main` before approving, and both now agree with the core: the pre-`--migrate-via` row is rejected instead of rendering a role of "4.0/5" and a status of "✅", and the row without a trailing pipe keeps `"last note"` instead of losing it.

The framing is what makes this one land as-is: **the 4th instance of #2369 and the only one where the web copied core _logic_ rather than a constant list.** The other three were lists — derivable, or freezable by a guard. Logic can't be diffed by eye, which is exactly why this sat behind a header comment claiming the file "mirrors parseTrackerRow".

And the cure matches the diagnosis: your test imports `parseTrackerRow` and compares against it, so it asserts *agreement with the core* rather than a snapshot of today's expected output. If the core's width rule changes, this goes red rather than drifting quietly. That's the difference between a test that preserves a convergence and one that can discover it was never there.

Verified the new checks actually run rather than being skipped by the `HAS_WEB` guard: `tracker-columns-tests.mjs` goes 43 → 46. (For anyone reading the CI later: the `test-all` total doesn't move, because it runs this file as a subprocess and counts its exit code as a single assertion. The coverage is real — it's visible in the file's own count.)

Thank you for going after the class instead of the symptom, and for numbering the instance. That's what made it obvious this needed parity treatment rather than a patch.

PR #2569, opened 2026-08-07
fix(scan): Unicode-aware keys for role dedup and company matching

barkaaa:
## What does this PR do?

Routes the two remaining private grouping keys in `scan.mjs` through the Unicode-aware `normalizeTextKey()` instead of a `[^a-z0-9]` strip. This is the follow-up to #2445: that PR fixed the three keys in `verify-pipeline.mjs`, but `scan.mjs` still carried its own strips at two call sites.

## Related issue

Refs #2393, #2397, #2445 — same root cause, different call sites. No new issue opened: per the template, bug fixes go straight in.

## The bug

On a non-Latin pipeline (Japanese, in my case), a `[^a-z0-9]` strip does not degrade the key — it deletes the entire string. Two checks were affected, and they fail in opposite directions.

**`normalizeRoleForDedup()` — the scanner silently drops roles.** Every Japanese title keys to `''`, so the company+role dedupe key collapses:

```
companyRoleDedupKey('株式会社アカネ', 'バックエンドエンジニア')  → "株式会社アカネ::"
companyRoleDedupKey('株式会社アカネ', 'フロントエンドエンジニア') → "株式会社アカネ::"
```

Two genuinely different openings, one key. The scan treats the second as already-seen and discards it. The company half was never affected (`defaultCompanyNormalizer` is `trim().toLowerCase()` with no strip), so this is purely the role half — and it means a Japanese-market user only ever sees the first role per company per run.

**`companyMatch()` — unrelated companies compare equal.** Both sides clean to `''`, and the equality branch is unguarded:

```js
if (c1NoSpaces === c2NoSpaces) return true;   // '' === '' → true
```

```
companyMatch('株式会社アカネ', '合同会社ゾロ')  → true
companyMatch('Acme', 'Zoro Inc')               → false   // control
```

At the call site (`buildCooldownFilter`, fed by `re_apply_windows` in `config/profile.yml`) a `true` means "this cooldown window applies, skip the job". So a single CJK entry in `re_apply_windows` suppressed **every** CJK job in the scan.

## The fix

Generalize `normalizeTextKey(value, separator = '')`. `scan.mjs` keys role titles as space-separated words — `normalizeRoleForDedup('Senior Engineer (Senior) (Berlin, Germany)') === 'senior engineer senior'` is an existing assertion — so a solid key is not a drop-in there. The separator lets it share this exact rule instead of keeping a second private strip to get that shape. Default `''` is unchanged for the four existing callers.

Both `scan.mjs` sites now call it, and the empty case is guarded explicitly (`c1NoSpaces && c1NoSpaces === c2NoSpaces`) — "no usable signal on either side" must never read as "identical".

Two smaller things in the same diff:

- `String(value)` → `String(value ?? '')` in `normalizeTextKey`. A nullish cell keyed to the literal string `"null"` / `"undefined"`; two such cells compared equal and formed a bogus group.
- `.normalize('NFKC')` at the top of `normalizeRoleForDedup`, before the trailing-suffix loop, so full-width brackets fold to ASCII and `Engineer （Remote）` strips the same way `Engineer (Remote)` always did.

## Review follow-up: accented-Latin regression (fixed)

@santifer caught a regression I had convinced myself wasn't there, and the diagnosis was exactly right. Because the fix keeps the accent that the old `[a-z0-9]` strip used to turn into a space, the containment fallback's `\b` anchors stopped holding for any name whose first or last character is non-ASCII:

```
'Nestlé Deutschland' vs 'Nestlé'   main=true → false
'Ørsted Energy'      vs 'Ørsted'   main=true → false
'Telefónica Tech'    vs 'Telefónica'  still true  (accent mid-word)
```

`\b` had been silently depending on the strip. Fixed with Unicode lookarounds plus the `u` flag, as suggested. The anchor class is `[\p{L}\p{M}\p{N}]` — the one `normalizeTextKey` keeps — deliberately: an anchor class without `\p{M}` treats a Devanagari matra as a boundary and splits `कंपनी` mid-word, which is the same key/boundary drift #2397 and #2445 removed elsewhere. Assertions cover all three classes, each with a witness that places the character immediately after the candidate (a space-boundary witness asserts nothing about the class).

I mis-scoped this in two ways worth recording: I framed the axis as Latin vs non-Latin when the real axis is *where* the non-ASCII character sits, and my "Latin regression" assertion was pure ASCII, so it was structurally incapable of catching it.

## Behaviour changes users will notice

- **Japanese pipelines will surface jobs that were previously skipped.** On `main`, any non-Latin entry in `re_apply_windows` matched every non-Latin-named job through the `'' === ''` collision, putting the whole JP pipeline into cooldown. That mass over-skip is gone, so the first scan after this lands may return noticeably more results. That is the fix working, not a new bug.
- **NFC and NFD spellings now agree.** `'Nestlé'.normalize('NFD')` vs `'Nestle'` matched on `main` (the strip removed the decomposed accent) while the NFC spelling did not. Both now behave the same way.

## Backward compatibility

The default path of `normalizeTextKey` is unchanged. Verified by comparing old vs new over every code point `0x0 … 0x10FFFF`, each tested alone, embedded (`a<ch>b`), and doubled: **0 differences**. The `+` quantifier and the added `.trim()` are provably no-ops when `separator === ''`, since no whitespace survives the strip. Same for the two `scan.mjs` functions on Latin input: 600k randomized ASCII titles through `normalizeRoleForDedup` and 300k randomized ASCII company pairs through `companyMatch`, **0 differences**.

The `u` flag cannot introduce a throw: both operands are already `normalizeTextKey(x, ' ')` output. Sweeping every code point `0 … 0x10FFFF` (including the lone-surrogate range) produced **0 outputs containing anything outside `[\p{L}\p{M}\p{N} ]`** and **0 throws** from `companyMatch` in either direction. The one pattern-length `RangeError` (≥~40k chars) exists identically on `main`. Lookbehind and `\p{…}` are both well below the `engines` floor of Node 18, and `compileLocationKeyword` already ships a lookbehind on `main`.

Dedupe keys are never persisted — `companyRoleDedupKey` is recomputed per run from `applications.md` / `scan-history.tsv` / `pipeline.md` — so there is no key migration for existing user data.

## Undisclosed regression I found while re-auditing — your call on scope

Fuzzing for more instances of the class you caught turned up one that the anchor fix cannot reach, because it lives in the key rather than the boundary. Disclosing it rather than deciding the scope myself.

`toLowerCase('İ')` (U+0130, Turkish/Azeri/Lithuanian dotted capital I) yields `i` + U+0307 COMBINING DOT ABOVE. `main`'s strip deleted the mark; `normalizeTextKey` keeps it via `\p{M}`, so two spellings of one company key differently:

```
'BİLİŞİM A.Ş.'  →  "bi̇li̇şi̇m a ş"   (0069 0307 …)
'Bilişim A.Ş.'  →  "bilişim a ş"     (0069 …)

companyMatch('BİLİŞİM A.Ş.', 'Bilişim A.Ş.')   main=true → false
```

Same user-visible symptom as the Nestlé case — a cooldown window silently stops being honoured — and realistic, since job boards routinely render company names in all caps.

**Where it comes from:** the behaviour is `normalizeTextKey`'s and has been live since #2445 for `normalizeCompany` / `normalizeVia`. This PR is what routes `companyMatch` through that function, so the regression *at this call site* is caused by this PR, but fixing it only here would make `companyMatch` disagree with every other consumer. Fixing it properly means case-folding U+0307 inside `normalizeTextKey`, which changes tracker dedup and Via identity too.

That is a scope decision I don't think is mine to make. Happy to fold it into this PR if you want it in, or file it as a sibling of #2570 — tell me which and I'll turn it around.

## Known gaps (deliberate)

The containment fallback does not fire for non-Latin names. `株式会社メルカリ` vs `メルカリ` does not match: the lookbehind sees 社, a letter, so there is no boundary to assert, and Japanese is not space-delimited so no anchor rule recovers it. To be precise about the delta — that pair *did* return `true` on `main`, but through the `'' === ''` collision that erased both names, not through this path. Making it match on purpose needs corporate-form normalisation (株式会社 / 合同会社 / …), i.e. new vocabulary rather than a key fix, filed as #2570.

Same reasoning for `normalizeRoleSuffixTag()`, which carries a third `[a-z0-9]` strip: making it Unicode-aware alone changes nothing, because the tag is then matched against the English-only `ROLE_LOCATION_SUFFIXES` set. `エンジニア（東京）` and `エンジニア（大阪）` therefore remain two keys. Noted in a code comment so the next reader doesn't assume the NFKC pass covers it.

## Testing

`node test-all.mjs` → **2993 passed, 0 failed**.

Sixteen assertions in section 45c, covering both directions:

- distinct Japanese roles at one company stay apart
- a non-Latin title is preserved rather than keyed to `''`
- unrelated Japanese companies stay apart
- empty/absent company names are no-match, not equal
- **control:** equivalent Japanese company+role (half-width kana, full-width spacing) still dedupes to one non-empty key
- **control:** equivalent Japanese company names still match
- half-width katakana folds onto full-width
- full-width bracketed suffixes strip (the only assertion that can detect the early NFKC pass — `normalizeTextKey` NFKCs at the end, so the final key alone can't distinguish)
- **regression:** Latin role keys keep their space-separated shape
- **regression:** Latin `companyMatch` still matches equivalents and rejects unrelated names
- accented Latin names still match at a word edge (Nestlé, Ørsted, Zoë)
- the containment fallback stays bounded (`Acme` does not match `Acmetric`)
- the anchor class treats combining marks and digits as letters
- a Devanagari name still matches at a space boundary
- `normalizeTextKey`'s separator arg doesn't disturb the default
- `normalizeTextKey` keys nullish input to `''`

Every assertion is mutation-checked. Reverting only the source change fails 10 of 16 while both controls and both Latin regressions still pass. Removing only `.normalize('NFKC')` fails exactly 1. Dropping `\p{M}` from the anchors fails the matra witness only; dropping `\p{N}` fails the digit witness only; reverting only the lookbehind fails `Ørsted` only. Two earlier assertions were vacuous (a control comparing a pure function to itself, and a Devanagari witness whose next character was a space) and have been replaced with witnesses that actually flip.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] My PR does not include personal data (CV, email, real names, scan results, or pipeline data) — test fixtures use invented company names

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved matching and deduplication for multilingual company and role names.
  * Preserved distinctions between non-Latin titles and company names.
  * Added support for Unicode variations, including full-width brackets and half-width katakana.
  * Prevented empty company values from matching entries.
  * Improved handling of missing values and configurable separators during text normalization.

* **Tests**
  * Added regression coverage for Unicode normalization, multilingual data, separators, and existing Latin-text behavior.

### DIFF
--- a/scan.mjs
+++ b/scan.mjs
@@ -42,7 +42,7 @@ import { loadProviders, resolveProvider } from './providers/_registry.mjs';
 import { mergeProviderPlugins } from './plugins/_engine.mjs';
 import { classifyFetchError } from './verify-portals.mjs';
 import { fingerprintText, findCrossListings } from './fingerprint-core.mjs';
-import { resolveColumns, parseTrackerRow } from './tracker-parse.mjs';
+import { resolveColumns, parseTrackerRow, normalizeTextKey } from './tracker-parse.mjs';
 import { normalizeCompany } from './tracker-utils.mjs';
 import { normalizeCompanyName } from './invite-match.mjs';
 import { withPipelineLock } from './pipeline-lock.mjs';
@@ -709,14 +709,17 @@ export function buildSalaryFilter(salaryFilter) {
 }
 
 export function companyMatch(jobCompany, windowCompany) {
-  const cleanNoSpaces = (str) => String(str || '').toLowerCase().replace(/[^a-z0-9]/g, '');
-  const c1NoSpaces = cleanNoSpaces(jobCompany);
-  const c2NoSpaces = cleanNoSpaces(windowCompany);
-  if (c1NoSpaces === c2NoSpaces) return true;
+  // Unicode-aware (#2393 family): the [a-z0-9] strip this used to carry erased
+  // non-Latin scripts outright, so 株式会社アカネ and 合同会社ゾロ both cleaned
+  // to '' and the equality check below reported two unrelated companies as the
+  // same one. The empty guard is part of the fix, not decoration — "no usable
+  // signal on either side" must never read as "identical".
+  const c1NoSpaces = normalizeTextKey(jobCompany);
+  const c2NoSpaces = normalizeTextKey(windowCompany);
+  if (c1NoSpaces && c1NoSpaces === c2NoSpaces) return true;
 
-  const cleanWithSpaces = (str) => String(str || '').toLowerCase().replace(/[^a-z0-9\s]/g, ' ').replace(/\s+/g, ' ').trim();
-  const c1WithSpaces = cleanWithSpaces(jobCompany);
-  const c2WithSpaces = cleanWithSpaces(windowCompany);
+  const c1WithSpaces = normalizeTextKey(jobCompany, ' ');
+  const c2WithSpaces = normalizeTextKey(windowCompany, ' ');
   if (!c1WithSpaces || !c2WithSpaces) return false;
 
   const regex1 = new RegExp('\\b' + c2WithSpaces.replace(/[-\/\\^$*+?.()|[\]{}]/g, '\\$&') + '\\b');
@@ -1277,13 +1280,27 @@ function isRoleLocationSuffix(tag) {
  * @returns {string} Normalized role key.
  */
 export function normalizeRoleForDedup(role) {
-  let title = String(role ?? '').toLowerCase();
+  // NFKC up front so full-width brackets fold to their ASCII forms while the
+  // suffix loop can still see them: "Engineer （Remote）" now strips the same
+  // way "Engineer (Remote)" always did.
+  //
+  // This does NOT make the loop understand non-Latin suffixes — the tag itself
+  // is still matched against the English-only ROLE_LOCATION_SUFFIXES set via
+  // normalizeRoleSuffixTag(), which carries its own [a-z0-9] strip. So
+  // "エンジニア（東京）" and "エンジニア（大阪）" remain two keys. Teaching the
+  // suffix vocabulary other scripts is a separate change (new vocabulary, not
+  // a key fix) and is deliberately out of scope here.
+  let title = String(role ?? '').normalize('NFKC').toLowerCase();
   while (true) {
     const match = title.match(/\s*[\[(]([^[\]()]+)[\])]\s*$/);
     if (!match || !isRoleLocationSuffix(match[1])) break;
     title = title.slice(0, match.index).trimEnd();
   }
-  return title.replace(/[^a-z0-9]+/g, ' ').trim();
+  // Unicode-aware (#2393 family): the [a-z0-9] strip this used to carry keyed
+  // every non-Latin title to '', so バックエンドエンジニア and フロントエンド
+  // エンジニア at one company shared a dedupe key and the scan dropped the
+  // second as already-seen. Space separator keeps the word-collapsing shape.
+  return normalizeTextKey(title, ' ');
 }
 
 /**
--- a/scan.mjs
+++ b/scan.mjs
@@ -722,9 +722,38 @@ export function companyMatch(jobCompany, windowCompany) {
   const c2WithSpaces = normalizeTextKey(windowCompany, ' ');
   if (!c1WithSpaces || !c2WithSpaces) return false;
 
-  const regex1 = new RegExp('\\b' + c2WithSpaces.replace(/[-\/\\^$*+?.()|[\]{}]/g, '\\$&') + '\\b');
-  const regex2 = new RegExp('\\b' + c1WithSpaces.replace(/[-\/\\^$*+?.()|[\]{}]/g, '\\$&') + '\\b');
-  return regex1.test(c1WithSpaces) || regex2.test(c2WithSpaces);
+  // Containment: a short window name should still match a longer official one
+  // ("Acme" vs "Acme Corp"), bounded so "Acme" does not match "Acmetric".
+  //
+  // The anchors are lookarounds, not \b: JS defines \b against ASCII \w even
+  // under the u flag. Keeping the accent (rather than stripping it to a space,
+  // as the [a-z0-9] filter did before) means '\bnestlé\b' can never hold —
+  // neither side of the trailing anchor is a word character — so Nestlé
+  // Deutschland vs Nestlé would silently stop matching. Same for Ørsted, Zoë
+  // and every other name whose first or last letter is non-ASCII.
+  //
+  // The anchor class is the one normalizeTextKey keeps, deliberately. An anchor
+  // class without \p{M} would treat a Devanagari matra as a boundary and split
+  // कंपनी mid-word — the key and its boundaries have to agree on what a letter
+  // is, or they drift the way #2397 and #2445 fixed elsewhere.
+  //
+  // Non-Latin containment does not fire here (株式会社メルカリ vs メルカリ): the
+  // lookbehind sees 社, a letter, so there is no boundary to assert, and
+  // Japanese is not space-delimited so no anchor rule recovers it. Note this
+  // pair DID match before this change, but only via the '' === '' collision
+  // that erased both names — not through this path. Making it match on purpose
+  // needs corporate-form normalisation, tracked separately in #2570.
+  //
+  // compileLocationKeyword() above reached for lookarounds too, for a related
+  // reason ("\b behaves surprisingly at a punctuation edge"); its escape set is
+  // reused here because '\-' is an invalid identity escape under u. Both
+  // operands are already normalizeTextKey output — letters, marks, digits and
+  // spaces only — so the escape is defensive, not load-bearing.
+  const bounded = (name) => new RegExp(
+    `(?<![\\p{L}\\p{M}\\p{N}])${name.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')}(?![\\p{L}\\p{M}\\p{N}])`,
+    'u',
+  );
+  return bounded(c2WithSpaces).test(c1WithSpaces) || bounded(c1WithSpaces).test(c2WithSpaces);
 }
 
 export function addDays(dateStr, days) {
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -11331,6 +11331,51 @@ try {
     fail('Latin companyMatch behavior changed');
   }
 
+  // -- Accented Latin: the containment fallback must survive keeping the accent --
+  // Before this change the [a-z0-9] strip turned é into a space, which left the
+  // \b anchors sitting on ASCII letters, so 'Nestlé Deutschland' vs 'Nestlé'
+  // matched. Preserving the accent breaks \b (it is ASCII-only in JS even under
+  // the u flag), so the anchors are Unicode lookarounds. A pure-ASCII assertion
+  // cannot detect this — the accent has to be at a word edge.
+  const accented = [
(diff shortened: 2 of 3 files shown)


santifer:
@barkaaa the core of this is exactly right and I want it in: routing both call sites through the shared `normalizeTextKey` instead of adding a fourth private fold is the direction, and you're the person who created that helper in #2445, so this is you closing out your own cluster. The bug is real and I reproduced it (株式会社アカネ and its two different roles collapsing to the same dedup key on main, unrelated companies matching each other, and `companyMatch(null, undefined)` returning true).

**One undisclosed regression has to be fixed first**, and it hits a much larger population than the one you're fixing.

`companyMatch`'s containment fallback still builds `new RegExp('\\b' + name + '\\b')`, and JavaScript's `\b` is defined against ASCII `\w`. On main the accent was stripped to a space, so the anchors landed on ASCII letters and it worked. Your PR correctly **retains** the accent, and now when the normalized name starts or ends with a non-ASCII letter, both sides of the anchor are non-`\w` and `\b` can never hold:

```
'Nestlé Deutschland' vs 'Nestlé'        main=true → PR=false
'Ørsted Energy'      vs 'Ørsted'        main=true → PR=false
'Zoë Ltd'            vs 'Zoë'           main=true → PR=false
'Telefónica Tech'    vs 'Telefónica'    still true  (accent mid-word keeps the anchors on ASCII)
```

`companyMatch` feeds `buildCooldownFilter`, and true means "cooldown applies, skip this job". So a user with `Nestlé` in `re_apply_windows` silently stops having that cooldown honoured and gets re-shown roles they just applied to. Milder than the bug you're fixing, but the affected set (French, German, Spanish, Portuguese, Nordic company names) is far bigger than the CJK set it repairs.

Worth saying: your "Known gap (deliberate)" section identifies the `\b`/ASCII mechanism correctly, but scopes it to non-Latin names and says containment "still cannot fire", which reads as "no change". It never says accented Latin names **lose** matching they had on main, and the follow-up you filed (#2570) carries the same non-Latin-only scoping. The Latin regression assertion in the PR is pure ASCII (`'Acme Inc.'` / `'acme inc'`), so it structurally cannot catch this.

**The fix is one line and it holds:** replace the two `\b` anchors with Unicode lookarounds and add the `u` flag:

```js
new RegExp(`(?<![\\p{L}\\p{M}\\p{N}])${name}(?![\\p{L}\\p{M}\\p{N}])`, 'u')
```

Verified: all four regressions go back to true, `'Acme Inc'`/`'Acme'` stays true, `'Acme'`/`'Zoro Inc'` stays false, and it does **not** re-open the CJK false positives (株式会社アカネ vs 合同会社ゾロ stays false, empty stays false, and the genuine case in your #2570 correctly stays out of scope).

Add that plus one accented-Latin assertion and I'll merge it the same day. The rest of the PR is excellent: I checked your backward-compat claim independently by sweeping every code point (3.3M strings, zero differences), and the only divergence I found beyond it was the intended empty-guard fix.

barkaaa:
Thank you for the depth here — reproducing all three cases independently and sweeping every code point against the backward-compat claim is a lot more than I expected, and it caught something I had convinced myself wasn't there.

You're right that I mis-scoped it, in two ways worth naming. I framed the axis as Latin vs non-Latin, when the real axis is *where* the non-ASCII character sits: `Telefónica Tech` vs `Telefónica` survives because the accent is mid-word, while `Nestlé Deutschland` vs `Nestlé` breaks because it is at the edge — same language, same script. And the deeper miss is that `\b` had been silently depending on the old `[a-z0-9]` strip turning `é` into a space. I changed what survives normalisation without re-auditing the one downstream consumer that treated those characters specially. My "Latin regression" assertion was pure ASCII, so it was structurally incapable of catching it.

Fixed in 58c173e with the lookarounds and the `u` flag as you suggested. All four cases are back to `true`, `Acme`/`Acmetric` stays `false` so the boundary still bounds, and the CJK false positives do not reopen.

Two deliberate choices beyond what you asked for:

- The anchor class is `[\p{L}\p{M}\p{N}]`, matching `normalizeTextKey` exactly. Without `\p{M}` a Devanagari matra reads as a boundary and splits `कंपनी` mid-word — the same reason the key keeps marks (#2445). Key and boundary have to agree on what a letter is or they drift, which is the thing #2397/#2445 were about.
- The escape set is now `compileLocationKeyword`'s. To be precise, since I overstated this to myself first: only `\-` is an invalid identity escape under `u`, `\/` is fine — and both operands are already `normalizeTextKey` output, so nothing with a metacharacter can reach the regex at all. That change is defensive, not load-bearing, and I did not want to claim it fixes a live crash.

Two of my own assertions turned out to be vacuous when I mutation-tested them, and are replaced: a "control" that compared a pure function against itself with identical arguments, and the Devanagari witness — its next character was a space, so it passed no matter what the anchor class was. Each anchor class now has a witness where the character sits immediately after the candidate, and each flips if its class is dropped.

**One thing I found while re-auditing, which I am disclosing rather than scoping myself.** Fuzzing for more instances of the class you caught turned up one the anchor fix cannot reach, because it lives in the key rather than the boundary:

```
'BİLİŞİM A.Ş.'  →  "bi̇li̇şi̇m a ş"   (0069 0307 …)
'Bilişim A.Ş.'  →  "bilişim a ş"     (0069 …)

companyMatch('BİLİŞİM A.Ş.', 'Bilişim A.Ş.')   main=true → false
```

`toLowerCase('İ')` yields `i` + U+0307 COMBINING DOT ABOVE; `main`'s strip deleted the mark, `normalizeTextKey` keeps it via `\p{M}`. Same user-visible symptom as Nestlé — a cooldown window silently stops being honoured — and realistic, since boards often render company names in all caps.

The behaviour is `normalizeTextKey`'s and has been live since #2445 for `normalizeCompany` / `normalizeVia`; this PR is what routes `companyMatch` through it, so the regression at *this* call site is caused by this PR. But fixing it only here would make `companyMatch` disagree with every other consumer, and fixing it properly means case-folding U+0307 inside `normalizeTextKey`, which changes tracker dedup and Via identity too. That is a scope call I don't think is mine to make — happy to fold it into this PR if you want it in, or file it as a sibling of #2570. Details are in the PR description under "Undisclosed regression I found while re-auditing".

I have also rewritten the "Known gaps" section and #2570: both scoped the `\b` mechanism to non-Latin names, which understated it. The mechanism is "first or last character is non-ASCII", and that population is much larger than the one this PR repairs. #2570 now covers only what no anchor rule can serve — the scripts that are not space-delimited. And I added a note that Japanese pipelines will surface jobs that were previously skipped, since on `main` one CJK window put the whole CJK pipeline into cooldown; that flood is the fix working, not a new bug.

santifer:
Thoughtful follow-up on your own helper, @barkaaa. Three real improvements in one small change: the `separator` param (default `''`, so every existing caller is untouched) lets scan.mjs key role titles as space-separated words without a second private strip; `String(value ?? '')` fixes a genuine latent bug where null and undefined cells keyed to the literal "null"/"undefined" and matched each other into a bogus group; and collapsing runs with `+` plus `.trim()` is the correct shape. The comments explaining *why* each choice exists are exactly what the next reader needs. §55 contract-freeze stays green — no format change, just better keys. Merged! 🚀

PR #2572, opened 2026-08-07
docs(agents): require employer-side confirmation for aggregator listings

gitraun:
## Problem

`check-liveness.mjs` answers *"does this URL still render a job?"* — but for an aggregator, that question has the wrong answer built in. A filled or expired role on Wellfound still serves a title, a full description, and an Apply button, so `classifyLiveness` returns `active` and the listing flows through evaluation as a genuine opening.

The liveness checker isn't wrong; it just can't see this failure mode. Nothing upstream currently distinguishes "the employer is hiring" from "an aggregator still has a page up."

## Evidence

I hit this while scanning for product roles. Nine Wellfound listings, each checked against the employer's own careers page or ATS:

| Outcome | Count | Signal |
|---|---|---|
| Confirmed stale | 5 | 2 returned HTTP **410 Gone**; one mirror read *"Position Closed"*; one employer's careers page listed 7 open roles, none matching; one company is down to a single employee and publicly "rebuilding" |
| Confirmed live at employer | 1 | Present on the employer's own careers index — applied direct instead |
| Unverifiable | 3 | Real, operating companies with no discoverable employer-side job board |

So **5 of 9 were dead**, and every one of them would have passed a liveness check on the aggregator page. One company's careers link pointed straight back to the aggregator, making "verification" circular.

## Change

One new `###` subsection inside the existing **Offer Verification -- MANDATORY** section. Documentation only — no code changes.

It establishes that aggregator-sourced listings are unconfirmed by default, and specifies four outcomes:

1. **Found at the employer** → the employer URL becomes canonical; apply direct, not through the aggregator.
2. **Not found** → mark stale in `data/scan-history.tsv` or on the tracker row, never delete — the record is what prevents re-adding it as new on the next scan.
3. **Employer unidentifiable** → routes to the existing **Block G** posting-legitimacy path rather than guesswork.
4. Explicitly a confirmation step *before* `check-liveness.mjs`, not a replacement — run the liveness checker against the employer URL once you have it.

It reuses the section's existing Playwright mandate and keeps the headless-batch exception intact, rather than restating either.

## Notes for review

- Placed inside the existing verification section deliberately, rather than as a standalone section — it's the same discipline, and `AGENTS.md` warns that reinforcement-without-enforcement decays.
- Routes the unidentifiable-employer case into **Block G** instead of inventing a new flag type.
- Geography-neutral: names Wellfound / LinkedIn / Instahyre / Cutshort / Internshala / Naukri as examples, but the rule is source-agnostic.
- Happy to follow up with an issue first if this is better discussed before merging, or to extend it into a `verify-employer.mjs` helper if a scripted check would be more useful than an agent-level rule.

One practical caveat worth flagging for anyone implementing the scripted version: both Wellfound and Keka serve identical fallback pages to non-browser clients. During this run, every Keka job URL returned a byte-identical "job not available" page — including a control fetch against an unrelated tenant. These checks need a real browser, not WebFetch.

## Summary by CodeRabbit

- **New Features**
  - Aggregated job listings are now verified against the employer’s careers page or applicant tracking system before being evaluated, applied to, or presented as live.
  - Verified employer listings are treated as canonical sources.
  - Unverified or outdated listings are clearly recorded as such.
  - Unidentified employers are flagged as potential posting-legitimacy signals without unsupported conclusions.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -333,6 +333,19 @@ Two separate axes:
 
 **Exception for batch workers (headless mode):** Playwright is unavailable in headless pipe mode. Use WebFetch as fallback and mark the report header `**Verification:** unconfirmed (batch mode)`; the user can verify manually later.
 
+### Aggregator Listings -- Confirm at the Employer
+
+Job aggregators keep stale, filled, and ghost listings live long after the employer closed the req. A closed role's aggregator page still renders a title, a description, and an Apply button, so `check-liveness.mjs` reads it as `active` — the aggregator page itself proves nothing. **A listing sourced from an aggregator is UNCONFIRMED by default: it is not a real opening until the employer says so.** This covers Wellfound / LinkedIn / Instahyre / Cutshort / Internshala / Naukri and similar, however the listing arrived — pasted, scanned, or found via WebSearch.
+
+**Before an aggregator listing is evaluated, applied to, or presented to the user as live:**
+
+1. Identify the employer, then locate the same role on the **employer's own careers page / ATS** (Greenhouse, Lever, Ashby, Workday, or the company's `/careers`). Same Playwright discipline as above.
+2. **Found at the employer → the employer URL is canonical.** Use it as the report `**URL:**` and apply direct, never through the aggregator.
+3. **Not found at the employer → treat as stale.** Do not apply and do not present it as live. Mark it in `data/scan-history.tsv` or on the tracker row — never silently delete the record, since that is what stops it being re-added as new on the next scan.
+4. **Employer unidentifiable** (some aggregator and agency posts hide the company) → that is a **Block G posting-legitimacy signal**, not a research gap. Report it and stop. NEVER infer, guess, or invent the employer.
+
+This is a confirmation step, not a replacement for `check-liveness.mjs` — run the liveness checker against the **employer** URL once you have it. The headless-batch exception above still applies.
+
 ---
 
 ## CI/CD, Community and Governance


gitraun:
We could add it as optional too, if people incase do not want to verify with company portals or company portals are not updated frequently this step can be skipped

santifer:
Thanks for this, @gitraun.

This is measurement, not intuition: 9 Wellfound listings checked one by one against the employer's own careers page or ATS, 5 confirmed dead (two returned HTTP 410 Gone, one mirror said "Position Closed", one company had 7 openings posted and none was this one, another is down to a single employee and "rebuilding"), 1 live and 3 unverifiable. All 9 would have passed the liveness check, because on an aggregator a filled role still serves title, description and an Apply button; in one case the careers link pointed back at the aggregator, making the verification circular. You add a subsection inside the verification section that already exists (not a new one), with 4 outcomes, routing the unidentifiable-employer case to Block G instead of inventing a flag, and you flag that Wellfound and Keka serve "job not available" pages byte-identical to non-browser clients, so a scripted check needs a real browser.

gitraun:
Thanks for the comment

santifer:
I told you the next thing you'd hear from me would be a decision with reasons attached. That was twelve days ago, @gitraun, and you've heard nothing. Sorry.

Here is the decision, including the part I should have said the first time.

**The content is right and I'm not asking you to change anything.** Re-verified today: +13 lines, one file, CI and CodeQL green. Your evidence still stands on its own — 9 listings checked one by one, 5 confirmed dead, and the circular case where the employer link pointed back at the aggregator is the one that proves a liveness check can't be the last word on an aggregator posting.

**What I got wrong was the promise, not the review.** This PR touches `AGENTS.md`, which governs agent behaviour across every install and sits behind Santiago's sign-off in this repo. So "a decision from me" was never something I could deliver, and I said it anyway. It's queued for him with the review attached and nothing left to check.

Two practical notes so you're not waiting on anything:

- **Don't rebase.** It shows as behind because of my own merges, and I'll bring it up to date when it clears. A first contribution also has its CI held until a maintainer approves it, and every new push holds it again, so a rebase costs you a wait for nothing.
- **Nothing about the PR is blocking**, including the missing linked issue: at 13 lines with the measurement in the thread, that's fine.

The right answer to "who decides this" was available on day one and I gave you a queue position instead. That's the bit worth apologising for.

santifer:
Your measurement did more than fix the sentence it was aimed at, @gitraun, and I owe you the whole story rather than a verdict.

Santiago and I looked at where this belongs before merging it, and the answer changed what we did.

**`AGENTS.md` loads on every message of every session** — 455 lines, ~10K tokens, paid by every user on every turn. The Agent Skills guidance draws the line at fact versus procedure: *"create a skill when a section has grown into a procedure rather than a fact... a skill's body loads only when it's used."* The Offer Verification section had become a three-step procedure, and worse, a **second copy** of one that already lives in `modes/oferta.md`.

The two copies had drifted:

```
AGENTS.md         "Title + description + Apply = active"
modes/oferta.md   "title/role + a real job description OR an application path"
                  + reuses the auto-pipeline snapshot
                  + handles the opt-in CLI extractor
                  + says what to do when the posting IS dead
```

Only the second one governs an actual evaluation.

**And here is what your nine listings actually proved.** They don't just show that aggregators need a confirmation step. They show that `AGENTS.md`'s rule — *title + description + Apply = active* — is **false**, because that is exactly what a filled aggregator listing keeps serving. Five dead postings passed it. So the sentence you were adding an exception to wasn't merely duplicated: it was wrong, and it was wrong in the copy that doesn't govern evaluations.

That made the fix bigger than your PR: the procedure moves to the gate, `AGENTS.md` keeps only the fact and points at it, and your aggregator rules land inside the gate where the surrounding steps already are.

**#3227 carries your text and your measurement, with `Co-authored-by: gitraun`.** Your wording is intact except for two re-anchored references ("same Playwright discipline as above" now points at step 1 of the gate; "the headless-batch exception above" now points at step 5, which moved there with it), and I quoted the nine-listing result in the mode file so the rule carries its evidence.

**If you'd rather send it yourself, say so and I'll close mine.** You did the work that found it, and "the maintainer took my patch and shipped it" is a worse first contribution than "I shipped it". The only reason I wrote it is that the restructuring of `AGENTS.md` wasn't yours to do and I didn't want to hand you a bigger task than the one you signed up for. Your call, no wrong answer, and it costs me nothing to close #3227 and wait.

Either way: closing this PR when #3227 lands, not because the content was wrong but because it was too right to leave in the wrong file.

santifer:
@gitraun: reopening the question I left you on 23 August, because the answer is still yours to give and nothing has moved since.

The offer stands unchanged. #3227 carries your text and your nine-listing measurement with Co-authored-by, and if you would rather send it yourself I close mine and wait. You did the work that found this, and the only reason I wrote the replacement is that restructuring AGENTS.md was not the task you signed up for.

What I should add is that my own PR has been sitting just as still, so the wait has not been costing you a place in a queue that was moving. If you would rather not spend more time on it, saying so is a complete answer and I will land mine with your credit on it.

Either way, tell me which and I will act on it this week.

santifer (changes requested):
Thanks @gitraun for **requiring employer-side confirmation before an aggregator listing counts as live**. To get this landed we need a few concrete things; this is the complete list, not a first round:

- [ ] Say what the agent does when the employer has no careers page or ATS it can find (3 of 9 in your own data): `stale` is wrong for a live role
- [ ] Name the canonical tracker status to use, instead of "mark it in the tracker"
- [ ] Resolve or answer the open CodeRabbit thread

Measuring this on real listings, as the description does, is what made the gap easy to see. Welcome: there is no clock, and the PR stays yours.

PR #2577, opened 2026-08-07
feat(cli): add help flags to check-liveness

### LINKED ISSUE #2576: feat(cli): check-liveness.mjs has no --help — add the usage block
**Time:** ~20 min · **Difficulty:** Beginner · Good first issue

`check-liveness.mjs` prints its usage only when you call it wrong: line 44 does `console.error('Usage: node check-liveness.mjs [--no-fallback] [--throttle[=ms]] <url1> [url2] ...')` on the error path. Asking for help politely does nothing useful:

```bash
node check-liveness.mjs --help    # treated as a URL, no usage shown
```

## What to do

Add a `--help` / `-h` branch that prints the usage block to **stdout** and exits **0**, before anything else runs (no network, no file reads).

## The pattern to mirror

Two PRs did exactly this in the last few days, so you have a reviewed shape to copy rather than invent:

- #2481 on `detect-reposts.mjs` (merged)
- #2556 on `stats.mjs`

Follow whichever you prefer. Keep the wording of the existing usage line, extend it into a short block if the flags deserve a line each.

## Acceptance

- `node check-liveness.mjs --help` and `-h` print the usage and exit 0
- The early exit happens before any network call or file read
- Existing behaviour unchanged: no args still errors as it does today
- A test asserting both flags print the usage (see `detect-reposts.test.mjs` for the shape)

Comment here and it gets assigned to you.

slegarraga:
Closes #2576

## Summary

Adds `--help` and `-h` support to `check-liveness.mjs`.

Both flags print the usage block to stdout and exit `0` before any URL processing, file read, or network work. Existing behavior is unchanged: running with no arguments still prints usage to stderr and exits `1`.

## Tests

Added `check-liveness.test.mjs` covering:

- `--help` prints the usage block and documents `--no-fallback`, `--throttle`, and `--file`;
- `-h` prints the usage block;
- no arguments exits `1` and prints usage to stderr.

## Verification

```bash
node --check check-liveness.mjs
node check-liveness.test.mjs
```

All pass.

Signed off with DCO.

## Summary by CodeRabbit

* **New Features**
  * Added `--help` and `-h` options to display command usage information.
  * Documented the `--no-fallback` and `--throttle[=ms]` options.
  * Improved usage guidance when required arguments are missing.

* **Tests**
  * Added automated checks for help output, usage guidance, help precedence, and missing arguments.


slegarraga:
Addressed the review suggestions: the usage block now documents both help aliases, the no-argument error reuses the same usage text, and the test suite proves --help exits before a --file path is read.

slegarraga:
Updated the `-h` assertion to check the complete usage line instead of matching the substring inside `--help`. The test suite passes locally (10 passed, 0 failed).

slegarraga:
Addressed the remaining exit-handling feedback: the help path now returns normally, the no-argument path sets `process.exitCode` before returning, the test summary uses `process.exitCode`, and the final CLI result no longer force-exits after output. No `process.exit()` calls remain in these files; the test suite passes locally (10 passed, 0 failed).

santifer:
Thanks for this, @slegarraga.

Adds --help and -h to check-liveness.mjs, and the detail that matters is the ordering: it prints usage to stdout and exits 0 BEFORE reading any file, processing URLs or touching the network, leaving the current no-args behavior untouched (usage to stderr, exit 1). The new test covers the three things separately, including --help precedence, and checks that the usage block documents --no-fallback, --throttle and --file. Closes your own issue #2576, signed off with DCO.

santifer:
Your CI is red on a real but one-move fix, @slegarraga. I re-ran it on today's main first so this is measured against reality, not a stale base.

**The failure:** `check-liveness.test.mjs` sits at the repo root, and every tracked file must be covered by `SYSTEM_PATHS` or `USER_PATHS` so the updater knows to ship it: the coverage guard fails with exactly that (`Coverage gap: check-liveness.test.mjs`).

**The fix:** move it to `tests/check-liveness.test.mjs`. The `tests/` directory is already a covered prefix (update-system.mjs line 210) and the runner auto-discovers and executes everything under it, so the move alone turns ubuntu and macOS green: no registration edit needed.

The remaining Windows red (`followup-seed-tests.mjs` lock timeout) is a known flake in the follow-ups lock's wait loop, not your doing: a fix for that loop is already in the merge queue, and I will re-run your CI once it lands. The --help ordering work itself still looks right to me; this is purely placement.

slegarraga:
Fixed — moved `check-liveness.test.mjs` to `tests/` as you suggested. Thanks for re-running CI against current main before diagnosing; that made the report unambiguous. Understood on the Windows flake being tracked separately, no rush. The help-ordering behavior is untouched by the move.

slegarraga:
Root cause of the red CI, fixed in 2d9908c: the test file moved under `tests/` to satisfy the coverage guard, but it still resolved `./check-liveness.mjs` relative to itself — so every job died with `ERR_MODULE_NOT_FOUND` before a single assertion ran. Now resolved via `../check-liveness.mjs`, and I also adopted CodeRabbit's suggestion: `-h` output is asserted byte-identical to `--help`, so the alias can't drift.

Verified locally (Node v22.23.2, macOS arm64): 11 passed / 0 failed, both from the repo root and from inside `tests/`. CI on this head should be green across the three OSes.

slegarraga:
Second fix in aada021: the suite guard caught this file calling `process.exit()` — adopted the house pattern instead (`pass`/`fail` from `tests/helpers.mjs`, shared counters, no exit), same assertions with `-h` still pinned byte-identical to `--help`. Local run on Node v22.23.2: 12 passed / 0 failed from both the repo root and `tests/`.

Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/check-liveness.test.mjs`; `check-liveness.mjs` is outside my area and I read it for context.

The test asserts the behavior, not just the absence of a crash. `--help` is checked for exit 0 and for `Usage:` in the output, `-h` is asserted byte-identical to `--help` so an alias cannot drift, and the no-argument path is checked for exit 1 with `Usage:` on stderr. The case worth calling out is `--help` combined with a missing `--file`: it still exits 0 with the usage block, which proves `--help` returns before any file read or browser launch. Neighbour-consistent with the repo's other `--help` tests (spawn, status, `/Usage:/`), and not stale, the current `check-liveness.mjs` still matches the pre-image the diff applies to.

slegarraga:
Thank you for the thorough code-owner review, @Scott-Emberson!

Really appreciate that you verified the assertions carry real behavior guarantees (exit codes, `Usage:` on stdout/stderr, the alias pinned byte-identical to `--help`) instead of just checking for absence of a crash, and that you called out the `--help` + missing `--file` case as the proof that usage returns before any file read or browser launch. That is exactly the contract this test exists to pin, and having a code owner confirm it reads neighbor-consistent with the repo's other `--help` tests means a lot.

I refreshed the branch onto current main just now, so the suite also runs green against the latest base. Thanks again for the careful read.

santifer:
Queued for the next merge batch, @slegarraga. Sorry for the day of quiet after you replied.

Worth naming what happened here, because it's the shape I want more of: @Scott-Emberson reviewed this as a code owner and checked that the assertions carry real behaviour guarantees rather than just passing — exit codes, `Usage:` landing on the right stream, the alias pinned. That's a review that would survive someone rewriting the implementation, which is the only kind worth having on a CLI-surface change.

Nothing needed from you. Don't rebase: the base is behind because of my merges today and I'll bring it forward.

santifer:
Patience rewarded, @slegarraga: 18 days in the queue is on us, not you. Clean --help following the house pattern, usage block accurate to the real flags. Merged 🚀

PR #2587, opened 2026-08-07
fix(dedup): distinct non-Latin companies merged into one, deleting a row

darkpandawarrior:
## The bug

`dedup-tracker.mjs` kept its own local `normalizeCompany()` using the pre-#2429 ASCII fold:

```js
.replace(/[^a-z0-9 ]/g, '')
```

Any company name written entirely in a non-Latin script folds to `''` under that. Two **different** employers therefore share a grouping key, pass `roleMatch()` when their titles agree — common, since "Engineer" / "Backend Engineer" recur everywhere — and the lower-scored row is **deleted**.

```
| 1 | アクメ株式会社       | Engineer | 4.2/5 |
| 2 | グロベックス合同会社 | Engineer | 3.0/5 |
```

```
before: 🗑️  Remove #2 (グロベックス合同会社 — Engineer, 3.0/5) → kept #1 (4.2/5)
after:  📊 0 duplicates removed
```

The run prints its normal success output. Nothing warns. Only the `.bak` preserves the deleted application.

## This is the third instance of the same class here

That's the part I think matters more than the fix:

- `f589ce83` (#2429) made `tracker-utils.mjs`'s `normalizeCompany` Unicode-aware. `merge-tracker.mjs` and `set-status.mjs` picked it up automatically **because they import the shared helper**. `dedup-tracker.mjs` had forked its own copy, so it didn't.
- `fe4561b` (#2397) then fixed *this same file's* blind-via key — while the ordinary company path twelve lines below kept the old fold.
- The comment block right above that code already describes this exact defect, for the Via key.

So the fix is to delete the local copy and import the shared one, rather than patch the regex — that's what stops a fourth instance.

## Why this is safe

The value is only ever an in-memory grouping key (`groups` Map, `dedup-tracker.mjs:328`). It is never persisted, never written to a row, and never compared against a key produced by another script. Adopting the shared normalization changes no stored data.

I checked the Latin path explicitly rather than assuming: `Acme (Inc.)` and `Acme Inc` still merge, and `Globex` / `Initech` still stay apart.

## Tests

A sibling of the existing #2393 blind-via block, placed next to it, on the ordinary company path:

- two distinct non-Latin employers stay apart *(fails before this change)*
- punctuation variants of one Latin employer still merge
- two distinct Latin employers still stay apart

The last two are the regression guards for the normalization swap.

`node test-all.mjs` → **0 failed**.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved duplicate detection for company names using Unicode-aware normalization.
  * Distinct non-Latin company names, including Japanese names, are now kept separate.
  * Equivalent Latin company names with punctuation or spacing differences continue to be combined while retaining the higher-scoring entry.
  * Distinct Latin companies are no longer incorrectly grouped together.

* **Tests**
  * Added regression coverage for multilingual names and company-name variations.

### DIFF
--- a/dedup-tracker.mjs
+++ b/dedup-tracker.mjs
@@ -14,7 +14,7 @@ import { readFileSync, copyFileSync, existsSync, mkdirSync } from 'fs';
 import { dirname } from 'path';
 import { fileURLToPath } from 'url';
 import {
-  openTrackerTransaction, rebuildRow, resolveTrackerPath,
+  openTrackerTransaction, rebuildRow, resolveTrackerPath, normalizeCompany,
 } from './tracker-utils.mjs';
 import { resolveColumns, parseTrackerRow, normalizeVia } from './tracker-parse.mjs';
 
@@ -60,25 +60,6 @@ const STATUS_RANK = {
   'contratado': 7,
   'contratada': 7,
 };
-
-/**
- * Normalize a company name into the grouping key used by deduplication.
- *
- * The tracker may contain punctuation, parenthetical branding, or spacing
- * differences for the same employer. This function removes those presentation
- * differences while keeping the alphanumeric company identity that determines
- * which rows are safe to compare for duplicate roles.
- *
- * @param {string} name - Company name from an applications.md row.
- * @returns {string} Lowercase company key used for same-company grouping.
- */
-function normalizeCompany(name) {
-  return name.toLowerCase()
-    .replace(/[()]/g, '')
-    .replace(/\s+/g, ' ')
-    .replace(/[^a-z0-9 ]/g, '')
-    .trim();
-}
 
 /**
  * Normalize tracker status text before ranking or comparing it.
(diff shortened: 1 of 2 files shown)


santifer:
Merged, @darkpandawarrior, and this is the right shape for it. #2429 made `tracker-utils`'s `normalizeCompany` Unicode-aware and `merge-tracker`/`set-status` inherited it; `dedup-tracker` kept a private copy, so it stayed on the old `[^a-z0-9]` fold. Deleting the copy rather than patching it is what stops the two from drifting apart a third time.

Reproduced here before merging, with both controls: the local function on `main` folds アクメ株式会社 and グロベックス合同会社 to the same empty key (so one of two genuinely separate applications got deleted), while the shared one keeps them apart and still merges `Acme (Inc.)` with `Acme Inc` and still keeps `Acme` and `Globex` separate. Main 3070/0 after the merge.

Worth naming: the blind-via channel path right above it was fixed for this exact class in #2393, and the ordinary path underneath it stayed broken. A fix that lands next to its sibling and doesn't get applied to it is a pattern I want to catch earlier 🚀

Issue #1144, opened 2026-06-20
Opt-in LLM relevance re-ranker for scan results (zero-token by default)


darkpandawarrior:
### Code of Conduct
- [x] I agree to follow the Code of Conduct

### Existing issues
- [x] I searched existing issues and this hasn't been requested yet

### Problem

The zero-token scanner filters scan results by keyword (`title_filter` / `location_filter`). That's deliberately blunt, so it has two failure modes the human then has to clean up by hand in `pipeline.md`:

- **False positives** — e.g. a bare "Mobile" or generalist "Software Engineer III" title that passes the positive keywords but is actually backend/iOS once you read the JD.
- **False negatives** — strong-fit roles whose titles don't literally contain a keyword.

There's room for a smarter pass — but it costs tokens, so the core scan should stay out of it.

### Proposed solution

An **opt-in, manually-invoked** relevance pass that is **off by default** (the core `scan.mjs` stays 100% zero-token):

- New flag / small script, e.g. `node scan.mjs --rank` or a standalone `rank-pipeline.mjs`, that takes **only the new pipeline entries** (title + location, optionally a short fetched JD snippet) and asks the LLM to score relevance 0–5 against the user's `cv.md` / `title_filter`, annotating each pipeline line with `score + one-line reason`.
- **No new dependency**: reuses the existing headless-CLI pattern the project already ships for batch (`claude -p` / `gemini -p` / `codex exec` … per AGENTS.md "Headless / Batch Mode"), so it runs under whatever CLI the user already has — no new npm package, no new external API key.
- **Local + human-in-the-loop**: it only annotates `pipeline.md`; the human still reviews and applies. No auto-apply, no scraping.
- **Bounded, visible cost**: only new entries, only when the user opts in per run.

Why raise this as an issue rather than a PR: per CONTRIBUTING, token/automation features should be discussed first, and I'd want to keep the proud zero-token default intact. Happy to implement it behind the flag if the direction is welcome, and to adjust scope (title-only vs JD-aware, scoring rubric, output format) to your preference.

_(Related but separate, not part of this proposal: an opt-in LLM step to resolve the custom/token-gated "agent-handoff" companies. That brushes up against third-party ToS for some platforms, so I'd keep it out of scope here.)_

### Area

Portal Scanner

darkpandawarrior:
[...]

PR #2588, opened 2026-08-07
fix(cv): $-patterns in candidate text splice the template into the CV

darkpandawarrior:
## The bug

Both CV builders fill placeholders by passing the escaped value as the **string** replacement argument to `String.replace`. JS scans that argument for `$&`, `$'`, `` $` `` and `$$` — and neither `escapeLatex` nor `escapeHtml` neutralizes them (`escapeLatex` turns `$` into `\$` and leaves the *next* character alone).

So an ordinary CV bullet corrupts the document.

**`$'` — silent corruption, exit 0:**

```
bullet: "Reported budget lines in $'000 format for quarterly reviews"
```

`$'` means *everything after the match*, so the remainder of the template is spliced into the body:

```
Reported budget lines in \
  \resumeSubHeadingListEnd
%%%%%%%%%%%%  Technical Skills  %%%%%%%%%%%%
\section{Technical Skills}
...
```

The produced `.tex` contained **two `\end{document}`**. `build-cv-latex.mjs` printed `"valid": true` and exited 0.

**`$&` — misattributed build failure:**

```
bullet: "Delivered cost savings of $&500K across the org"
```

`$&` re-inserts the matched text — i.e. `{{BULLETS}}` itself — which then trips `build-cv-html`'s unresolved-placeholder guard and fails the build with `Unresolved placeholders: {{BULLETS}}`, pointing at the template rather than at the input that caused it.

Money amounts in bullets are exactly where these sequences occur, which is to say: on real CVs.

## The fix

Pass a replacer **function** at every fill site. A function's return value is inserted literally, with no `$`-pattern interpretation. No behaviour change for text that doesn't contain those sequences.

I swept every `.replace()` with a non-literal string replacement rather than patching only the two paths that surfaced, so this covers **five sites**:

| | |
|---|---|
| `build-cv-html.mjs` | `fillEntry` — conditional block fill (×2) and scalar fill |
| `build-cv-latex.mjs` | the render loop, and the identical loop in `--test` |

`generate-pdf.mjs:197` also calls `.replace()` on placeholders but with a literal `' '`, so it is unaffected. `generate-cover-letter.mjs` already uses a replacer function.

## Tests

`tests/cv-dollar-pattern.test.mjs`:

- `$'` in a bullet does not splice the template (asserts exactly one `\end{document}`)
- `$&` in a bullet does not re-insert the placeholder
- `` $` `` and `$$` survive verbatim — `$$` must not collapse to a single `$`

All three fail before this change.

`node test-all.mjs` → **3058 passed, 0 failed**.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed CV generation so special dollar-sign sequences in candidate text are preserved as literal content.
  * Prevented unintended template substitutions, duplicated text, and unresolved placeholders in HTML and LaTeX output.

* **Tests**
  * Added coverage validating correct handling and escaping of special replacement patterns in generated CVs.

### DIFF
--- a/build-cv-html.mjs
+++ b/build-cv-html.mjs
@@ -261,20 +261,26 @@ function parsePartial(source) {
 function fillEntry(entryTemplate, blocks, fields, blockValues) {
   let out = entryTemplate;
 
+  // Every replacement below passes a FUNCTION rather than the value directly.
+  // A string replacement argument is scanned by JS for $-patterns, so candidate
+  // text containing $&, $', $` or $$ (escaping leaves those sequences intact)
+  // would splice part of the template into the CV instead of being inserted
+  // literally. A replacer function's return value is never interpreted.
+
   // Resolve conditional blocks: replace {{BLOCK_NAME}} with the
   // present/absent markup depending on whether the field has a value.
   if (blockValues) {
     for (const [name, { value, present }] of blockValues) {
       const block = blocks.get(name);
       if (!block) continue;
-      const markup = present ? block.present.replace(`{{${name}}}`, value) : block.absent;
-      out = out.replace(`{{${name}}}`, markup);
+      const markup = present ? block.present.replace(`{{${name}}}`, () => value) : block.absent;
+      out = out.replace(`{{${name}}}`, () => markup);
     }
   }
 
   // Fill remaining scalar placeholders.
   for (const [key, value] of Object.entries(fields)) {
-    out = out.replace(new RegExp(`\\{\\{${key}\\}\\}`, 'g'), value);
+    out = out.replace(new RegExp(`\\{\\{${key}\\}\\}`, 'g'), () => value);
   }
 
   return out;
--- a/build-cv-latex.mjs
+++ b/build-cv-latex.mjs
@@ -159,8 +159,13 @@ async function main() {
     SKILLS: buildSkills(payload.skills),
   };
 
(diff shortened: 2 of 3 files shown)


slegarraga (approved):
Correct root cause and fix: using replacer functions prevents JavaScript from interpreting `$&`, `$'`, `$`` and `$$` in candidate text. The coverage is well-scoped and the tests fail before the change as described. One non-blocking suggestion: the temporary directories from `mkdtempSync` are not cleaned up after each test, so using `t.after(() => rmSync(dir, { recursive: true, force: true }))` would keep the suite tidy. Looks good to me.

santifer:
This is the best kind of bug report, @darkpandawarrior: it looks like nothing and it corrupts the one document that has to be right. `escapeLatex` turning `$` into `\$` and leaving the next character alone is exactly why the escaping felt like it covered this and didn't.

What makes it worth merging as written is the failure mode you documented, not just the fix: `build-cv-latex` emitted a `.tex` with two `\end{document}` and still reported `"valid": true` with exit 0. A silent corruption that passes its own validity check is worse than a crash, and "$'000" in a bullet is not an exotic input: it's how people write money.

A/B verified here before merging: your three tests run against `main` fail (`$'` splices the template, `$&` re-inserts `{{BULLETS}}`, `$``/`$$` mangled) and pass on your branch. Full suite 3067/0.

And thank you for coming back after I left you waiting 48 days on #1144. That was on me. Merged 🚀

PR #2589, opened 2026-08-07
fix(scan): --company/--posted-after/--posted-before ignored in =value form

darkpandawarrior:
## The bug

`scan.mjs` reads `--company`, `--posted-after` and `--posted-before` with a bare `args.indexOf(flag)`, which returns `-1` for `--posted-after=2026-07-28`. The value resolves to `null` — indistinguishable from the flag never having been passed — so the run scans with **no bound at all** and reports a clean result.

The sharpest consequence is that **the typo guard stops guarding**. `scan.mjs` validates these dates and exits 1 on a malformed one, and its own comment says why:

> gated against a typo since a silently-ignored bound would look like "no jobs matched" instead of an error

With the `=` form the value never reaches that validator, producing exactly the outcome the guard exists to prevent:

```
before:  node scan.mjs --posted-after=not-a-date
         → no complaint; run continues unbounded

after:   Error: --posted-after expects YYYY-MM-DD, got "not-a-date"
```

## Fix

Use `flagValue` from `lib/cli-flags.mjs` — the module `37055d7d` added for precisely this and applied to five other scripts. `scan.mjs` didn't import it at all.

## Two scope notes

**`--since` is not affected.** It parses the `=` form itself (`1d9ff634`), so I left it alone. Mentioning it because a mechanical grep flags it and it looks like a miss.

**15 other scripts still have this pattern.** I swept every root `.mjs` for value-taking flags read via `args.indexOf`:

| script | flags |
|---|---|
| `analyze-patterns.mjs` | `--min-threshold`, `--min-vendor-n` |
| `contacts.mjs` | `--vcf` |
| `extract-latex-content.mjs` | `--out` |
| `fix-slugs.mjs` | `--file` |
| `followup-cadence.mjs` | `--applied-days` |
| `invite-match.mjs` | `--file` |
| `match-star.mjs` | `--jd`, `--top` |
| `paste-reply.mjs` | `--file` |
| `plugins.mjs` | `--sha` |
| `salary-gap.mjs` | `--stated-for` |
| `scan-interamt.mjs` | `--keyword` |
| `seed-fixture.mjs` | `--state` |
| `upskill.mjs` | `--url-text`, `--min-reports` |
| `verify-portals.mjs` | `--add`, `--file` |
| `weekly-digest.mjs` | `--from` |

I have **kept this PR to `scan.mjs`** — the one with a reproduced failure and the highest-traffic entry point — rather than shipping a 16-file sweep you didn't ask for. Say the word and I'll do the rest as a separate PR. I haven't verified each of those individually, so the table is a lead, not a finding.

## Tests

`tests/scan-flag-forms.test.mjs` — both bounds, in `=value` and space form, plus a valid-value control asserting the validator is *reached* rather than bypassed. The two `=value` cases fail before this change (4 pass / 2 fail → 6 pass).

`node test-all.mjs` → **3058 passed, 0 failed**.

## Summary by CodeRabbit

* **New Features**
  * Added support for both `--flag value` and `--flag=value` formats for company and date filters.

* **Bug Fixes**
  * Improved validation for malformed or missing company and date filter values.
  * Invalid or incomplete filters now produce clear errors instead of silently disabling filtering.

* **Tests**
  * Added coverage for both supported flag formats, validation scenarios, and successful command execution.

### DIFF
--- a/scan.mjs
+++ b/scan.mjs
@@ -46,6 +46,7 @@ import { resolveColumns, parseTrackerRow } from './tracker-parse.mjs';
 import { normalizeCompany } from './tracker-utils.mjs';
 import { normalizeCompanyName } from './invite-match.mjs';
 import { withPipelineLock } from './pipeline-lock.mjs';
+import { flagValue } from './lib/cli-flags.mjs';
 import { withPortalHealthLock } from './portal-health-lock.mjs';
 
 try {
@@ -1969,8 +1970,9 @@ async function main() {
   // --include-blacklisted: bypass the data/blacklist.md filter for auditing.
   // Matching postings flow through annotated instead of being counted out.
   const includeBlacklisted = args.includes('--include-blacklisted');
-  const companyFlag = args.indexOf('--company');
-  const filterCompany = companyFlag !== -1 ? args[companyFlag + 1]?.toLowerCase() : null;
+  // flagValue reads both `--flag value` and `--flag=value`; a bare indexOf misses
+  // the second form entirely and silently falls back to the unfiltered default.
+  const filterCompany = flagValue(args, '--company')?.toLowerCase() ?? null;
   // --posted-after / --posted-before <YYYY-MM-DD>: absolute-date bounds on the
   // employer's real posting date (job.postedAt), gated against a typo since a
   // silently-ignored bound would look like "no jobs matched" instead of an error.
@@ -1979,10 +1981,8 @@ async function main() {
     const d = new Date(`${s}T00:00:00Z`);
     return !Number.isNaN(d.getTime()) && d.toISOString().slice(0, 10) === s;
   };
-  const postedAfterFlag = args.indexOf('--posted-after');
-  const postedAfter = postedAfterFlag !== -1 ? args[postedAfterFlag + 1] : null;
-  const postedBeforeFlag = args.indexOf('--posted-before');
-  const postedBefore = postedBeforeFlag !== -1 ? args[postedBeforeFlag + 1] : null;
+  const postedAfter = flagValue(args, '--posted-after') ?? null;
+  const postedBefore = flagValue(args, '--posted-before') ?? null;
   if (postedAfter != null && !isValidIsoDate(postedAfter)) {
     console.error(`Error: --posted-after expects YYYY-MM-DD, got "${postedAfter}"`);
     process.exit(1);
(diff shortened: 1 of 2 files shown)


darkpandawarrior:
Both findings were right — fixed in b043c0d.

**Missing operands.** Correct, and `cli-flags.mjs`'s own doc for `hasFlag` says exactly this: *"flagValue alone cannot tell an ABSENT flag from one supplied without a value: both give undefined... Pair the two whenever a missing value must be a usage error."* I used `flagValue` without its partner. A trailing `--posted-after` now exits 1 with `--posted-after requires a value` instead of falling back to no bound — which was the same silent-default this PR is about, so thanks for catching that it survived in a second form.

Worth noting for the record it was not a regression: the previous `args[idx + 1]` also yielded undefined there. It just shouldn't stay broken in a PR fixing that class.

**Test determinism.** Also correct and the more serious of the two. The suite spawned `scan.mjs` with no portals override, so on a machine with a real `portals.yml` the valid-date cases could load providers, make network calls and write scan state — it only passed in CI because a bare checkout has none. Every run now pins `CAREER_OPS_PORTALS` at a nonexistent path so `scan.mjs` stops at its portals check, and each case asserts `error === undefined` and `signal === null` so a timeout can't pass as success.

Added the `--company` regression case too, plus missing-operand cases for all three flags: 6 → 10 assertions. Full `test-all`: 3058 passed, 0 failed.

slegarraga (changes requested):
Good fix, and the missing-operand handling is correct. One remaining case: `flagValue` consumes the next token even when it is another option, so `scan.mjs --company --dry-run` treats `--dry-run` as the company and silently filters everything. In `requireValue`, treat any value starting with `--` as missing, and add regression tests for each required flag followed immediately by another option. Happy to re-review after that.

santifer:
Merged, @darkpandawarrior. The reason this one is worth more than it looks: scan.mjs already validates these dates and exits 1 on a malformed one, and the comment says why: a silently-ignored bound would look like "no jobs matched" instead of an error. With the `=` form the value never reached that validator, so the guard was guarding nothing.

Pairing `flagValue` with `hasFlag` is the detail I'd have asked for: `flagValue` alone can't separate an absent flag from one passed with no operand, so a trailing `--posted-after` would have fallen back to "no bound" and scanned everything, which is the same silent default you're fixing. Reusing `lib/cli-flags.mjs` also keeps this on the path 37055d7d already took for five other scripts, instead of a sixth private parser.

Web-facing check done before merging: the web spawns `scan-ats-full`, not `scan.mjs`, and passes flags in the space form, so there is no consumer to break. Additive either way 🚀

PR #2590, opened 2026-08-07
fix(web): states.yml cached for the process lifetime, so core updates go unseen

darkpandawarrior:
## The bug

`readCanonicalStates()` used a bare module-level `if (cache) return cache`, so `templates/states.yml` was read **once per server process** — on the first request it ever served. The file's own doc comment says the opposite:

> we READ it live and never hardcode the list (the maintainer once mis-listed it from memory — the file had one more)

After a core update rewrites `states.yml` (a `git pull`, or `update-system.mjs apply`) without restarting the dashboard, `POST /api/status` keeps validating against the old snapshot:

- a status that is **now canonical** is rejected with `not a canonical status: …`
- a status upstream **removed or renamed** keeps being silently accepted

Nothing anywhere indicates the definitions are stale.

**A second, sharper edge:** the `FALLBACK` was cached too. A `states.yml` that was momentarily unreadable on the very first request pinned the hardcoded fallback for the rest of the process — later reads never recovered even after the file was fine again.

## Fix

Mirrors `loadHeaderAliases` in `tracker-table.mjs` exactly, since that function already solved this same problem for `tracker-aliases.json` and documents the reasoning:

- a `Map` keyed by resolved path holding `{mtimeMs, size, states}`
- one `statSync` per call, no full read on a hit
- **failures are never cached**, so a repaired file is picked up on the next request

Using the existing pattern rather than inventing one keeps the two read paths consistent.

## Verification

Exercised the real module with the `@/` alias import stubbed, rewriting `states.yml` between two calls:

```
origin/main : 1st "Applied,Rejected"   2nd "Applied,Rejected"                 ← stale
this branch : 1st "Applied,Rejected"   2nd "Applied,Rejected,Shortlisted"     ← live
```

`npx tsc --noEmit` clean.

## On the missing test

I did not add one, and I'd rather say why than quietly omit it: `web/` has no harness that can import a `.ts` module. All three suites under `web/tests/lib/` import `.mjs`, and `states.ts` pulls the `@/` alias, so plain `node --test` can't load it.

Standing that harness up is real infrastructure, and smuggling it into a bug fix is how a small PR becomes an unreviewable one. Happy to send it as its own PR if you want the coverage — just say which shape you'd prefer (a loader for `node --test`, or vitest).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when loading application states.
  * Updated state information is now recognized automatically when source files change.
  * Invalid or unavailable state data no longer affects future loading attempts.
  * Temporary fallback behavior is limited to the affected request, helping prevent stale results.

### DIFF
--- a/web/src/lib/core/states.ts
+++ b/web/src/lib/core/states.ts
@@ -30,12 +30,30 @@ const FALLBACK: CanonicalState[] = [
   { id: "hired", label: "Hired", aliases: ["contratado", "contratada", "hired", "accepted", "accept"], description: "Offer accepted, job landed!", group: "hired" },
 ];
 
-let cache: CanonicalState[] | null = null;
+/**
+ * Cached per resolved states.yml path and keyed on the file's mtime+size (one
+ * statSync per call, no full read), mirroring loadHeaderAliases in
+ * tracker-table.mjs.
+ *
+ * A bare `if (cache) return cache` made the "READ it live" contract above true
+ * only for the first request a server process ever served: a core update that
+ * rewrote templates/states.yml (git pull, update-system.mjs apply) was then
+ * invisible until restart, so a newly canonical status was rejected by
+ * POST /api/status and a removed one kept being accepted.
+ *
+ * Failures are NEVER cached, so a states.yml that is momentarily unreadable
+ * falls back for that call only, instead of pinning FALLBACK for the lifetime
+ * of the process.
+ */
+const statesCache = new Map<string, { mtimeMs: number; size: number; states: CanonicalState[] }>();
 
 export function readCanonicalStates(): CanonicalState[] {
-  if (cache) return cache;
+  const file = path.join(careerOpsRoot(), "templates", "states.yml");
   try {
-    const raw = fs.readFileSync(path.join(careerOpsRoot(), "templates", "states.yml"), "utf8");
+    const { mtimeMs, size } = fs.statSync(file);
+    const cached = statesCache.get(file);
+    if (cached && cached.mtimeMs === mtimeMs && cached.size === size) return cached.states;
+    const raw = fs.readFileSync(file, "utf8");
     const doc = yaml.load(raw) as { states?: unknown };
     const list = Array.isArray(doc?.states) ? doc.states : null;
     if (list && list.length) {
@@ -51,14 +69,17 @@ export function readCanonicalStates(): CanonicalState[] {
         });
       }
       if (parsed.length) {
-        cache = parsed;
+        statesCache.set(file, { mtimeMs, size, states: parsed });
         return parsed;
       }
     }
+    // Readable but unusable (empty/!states) — drop any stale entry so a
+    // repaired file is picked up on the next request rather than after a
+    // restart, and fall back for this call only.
+    statesCache.delete(file);
   } catch {
-    /* fall through to fallback */
+    /* unreadable — fall back for this call, never cache the failure */
   }
-  cache = FALLBACK;
   return FALLBACK;
 }
 


darkpandawarrior:
On the CodeRabbit `statesCache` finding — that one's a false positive, and its own wording hedges ("**If** these repetitions are present").

There is exactly one declaration:

```
$ grep -n 'statesCache' web/src/lib/core/states.ts
48: const statesCache = new Map<string, { mtimeMs: number; size: number; states: CanonicalState[] }>();
54:   const cached = statesCache.get(file);
72:     statesCache.set(file, { mtimeMs, size, states: parsed });
79:   statesCache.delete(file);
```

A duplicate `const` in the same scope is a compile error, so its own suggested check is already covered by CI: `npx tsc --noEmit` is clean and all 15 checks are green on this branch. It appears to have read a repeated diff hunk in its rendering rather than the file.

Leaving the code as-is. Happy to be shown otherwise if you see it differently.

slegarraga (approved):
This is a clean fix and the pattern matches `loadHeaderAliases`. Not caching failures is the important behavior, and the unusable-file path correctly drops stale entries. I agree the CodeRabbit duplicate-declaration finding is a false positive: the diff and CI confirm exactly one declaration. Non-blocking: `mtimeMs + size` can theoretically miss a rewrite that preserves both values, and a dedicated test harness for this module would be a valuable follow-up. Looks good to me.

santifer:
Good diagnosis, @darkpandawarrior, and the second edge you found is the sharper one: caching the `FALLBACK` means a `states.yml` that was briefly unreadable on the very first request pins the hardcoded list for the whole process lifetime, which is exactly the failure the file's own doc comment was written to prevent.

Routing rather than merging: the web dashboard has its own review track, so I've labelled this `area:web` and put it in that queue instead of taking it through core review, where I'd be the wrong reader for it.

What that means in practice: you get a first response on the PR, not an ETA for a fix. If it goes quiet for a week, say so here and I'll escalate it myself.

Your five core PRs are all merged, by the way: the `$`-pattern one in particular was a good catch.

santifer:
Both bugs confirmed in `main`, and I ran your fix against a live server rather than reading it. **Approving from the web side.**

That file is mine, and the doc comment right above the cache says *"we READ it live and never hardcode the list"*. You found a comment that had quietly become a wish — which is the second time this file has taught us that lesson (#2282 was its `FALLBACK` silently missing a state the core had added).

**Verified behaviourally, same process, no restart:**

| Step | Result |
|---|---|
| `POST /api/status` with `Ghosted` | `not a canonical status: Ghosted` ✅ rejected |
| Append `ghosted` to `states.yml`, **no restart** | — |
| Same POST again | `row not found` ✅ — it cleared the canonical gate |
| Make `states.yml` unreadable | `not a canonical status` ✅ falls back for that call |
| Restore the file | `row not found` ✅ **immediately** — the failure was never pinned |

That last pair is the one that matters most, and you were right that it's the sharper edge: before this, one transient read failure on the first request a process ever served would pin `FALLBACK` until restart.

**On the implementation:** keying on resolved path + `mtimeMs` + `size` with a single `statSync` is the right trade (no full read per call), and mirroring `loadHeaderAliases` in `tracker-table.mjs` keeps one idiom instead of inventing a second. Keying by *path* rather than a lone module variable also means a different `CAREER_OPS_ROOT` gets its own entry — a detail I'd have missed. Deleting the entry on readable-but-unusable content, so a repaired file recovers on the next request, is a nice third case; most fixes would have stopped at two.

One known limitation, worth a sentence rather than a change: `mtime`+`size` can't distinguish a same-second edit that preserves byte length. For `states.yml` — edited by humans and by `update-system.mjs`, not in a hot loop — that's an acceptable trade, and it's the same one the core already makes.

typecheck + `next build` green.

Thanks for reading the comment against the code instead of trusting it. That's exactly how this file's previous bug was found too.

santifer:
Merged, @darkpandawarrior. This one went through the web review track rather than mine, and it came back approved with the bug **reproduced against a live server**, not just read off the diff: canonical status rejected → `states.yml` edited without a restart → same POST now passes the gate → file made unreadable → falls back for that one call → file restored → correct again on the very next call, with nothing pinned.

Those last two steps are the edge you flagged and the reason this is worth more than a caching tweak: before, a transient read failure on the very first request a process served would pin the hardcoded FALLBACK until someone restarted the dashboard.

Two notes from the review worth passing back to you. Keying on path + `mtimeMs` + `size` with a single `statSync` is the right trade, and caching **per path** gives each `CAREER_OPS_ROOT` its own entry, which was called out as the detail an author could easily have missed. Known limitation, accepted rather than blocking: `mtime`+`size` can't distinguish a same-second edit that preserves the size, which is the same trade the core already makes for a file edited by humans and `update-system.mjs`.

Five merges today, and you found the class this belongs to: "derives from the core" isn't a property a copy or a cache keeps on its own 🚀

abankar1:
Late to this, but I ran two checks off the back of it that are worth writing down rather than leaving for the next person.

**1. I swept `web/` for siblings of this bug. There are none — here's the reason for each, so nobody redoes it:**

| Cache | Verdict |
|---|---|
| `lib/tracker-table.mjs` → `aliasCache` | Already the correct shape — it's the `loadHeaderAliases` pattern this PR mirrored |
| `app/api/usage/route.ts` → `cache` | Deliberate **60s TTL** (`{ at, data }`) over a heavy `~/.claude/projects` walk. Bounded, and it caches *observational* data, not a definition |
| `components/beta/beta-banner.tsx` → `searchCache` | Client-side, per-page-load, ephemeral search results |

So the class is closed in `web/` — `states.ts` was the only definition-file cache without invalidation.

**2. On the missing test — your read of the harness situation is exactly right, and I've hit it three times.**

Concretely, on `main` today: `web/src/lib/` holds **six** `.mjs` helpers (`clean-chips`, `pdf-paths`, `pdf-render`, `tracker-table`, `stream-parse`, `act-envelope`), every one imported by a `.ts`/`.tsx`, and **all five** web test files import only `.mjs`. Not one imports a `.ts`. Two of those helpers are mine, extracted for precisely the reason you describe — I couldn't test the logic where it lived, so I moved it.

The part worth flagging for whoever picks the harness decision: **that workaround structurally cannot cover your fix.** Extracting to `.mjs` only works for logic you can lift *out* of the `.ts`. `states.ts` is the `@/` alias plus `fs` plus `yaml` plus the `FALLBACK` — the file essentially *is* its I/O, so there's nothing to lift. It's the first case where the de-facto pattern runs out, which is a stronger argument for standing a loader up than "more coverage would be nice."

Also worth saying plainly: the guard I added in `test-all` 55.3b would **not** have caught this. It cross-checks `FALLBACK`'s *contents* against `states.yml` statically — it has no view of whether the runtime ever re-reads the file. Your behavioural verification (edit `states.yml`, no restart, same POST) is the thing that found it, and that's the gap a `.ts` harness would close.

Happy to take the harness PR if you'd rather not — I'd lean `node --test` with a loader over vitest, purely to keep one runner across core and web, but that's the web maintainer's call and I'd want it stated before writing any of it.

santifer:
This sweep is worth more than a late note, @abankar1: a table of every cache in web/ with the reason each one is NOT a sibling of the bug is exactly the artifact that stops the next person from redoing the audit. Logged. Thanks for coming back to close it out properly.

PR #2593, opened 2026-08-07
fix(doctor): `CLAUDE_CODE_USE_BEDROCK=0` reported that you were billing to Bedrock

### LINKED ISSUE #2592: doctor: `CLAUDE_CODE_USE_BEDROCK=0` reports that requests bill to your cloud account
### What

`checkBillingSource()` (added in ccb7b0d) treats the cloud switches as enabled when the variable is merely **present**:

```js
const cloud = ['CLAUDE_CODE_USE_BEDROCK', 'CLAUDE_CODE_USE_VERTEX', 'CLAUDE_CODE_USE_FOUNDRY']
  .filter((v) => process.env[v]);
```

Any non-empty string is truthy, so `=0` and `=false` count as on:

```
CLAUDE_CODE_USE_BEDROCK=1      ⚠ CLAUDE_CODE_USE_BEDROCK is set, so requests bill to your cloud account…
CLAUDE_CODE_USE_BEDROCK=0      ⚠ CLAUDE_CODE_USE_BEDROCK is set, so requests bill to your cloud account…
CLAUDE_CODE_USE_BEDROCK=false  ⚠ CLAUDE_CODE_USE_BEDROCK is set, so requests bill to your cloud account…
```

These switches are documented as `=1`, so `=0` is exactly how someone turns one off — and `docs/RUNNING_ON_A_BUDGET.md` already sends users to check them when hunting a stray billing source. Someone who sets `=0` *because* they were chasing this, then runs `npm run doctor`, is told the thing they just disabled is still active.

### Why it matters more than a normal false positive

The check's stated purpose is *"lo que no es legitimo es que el usuario no sepa que la esta usando en vez del plan que ya paga"* — telling a user where their money goes. A billing diagnostic that misreads an explicit opt-out produces precisely the confusion it was built to remove, and it's the check people will trust most when they're already worried about cost.

### Fix

Treat the switches as enabled only when set to a truthy value. That matches the repo's own env-flag convention (`=== '1'` in `merge-tracker.mjs:75` and `update-system.mjs:912`), while also accepting `true`/`yes`/`on` since these are third-party switches users copy from assorted docs.

PR incoming, with tests driving the real CLI (`doctor.mjs` has no exports or main-module guard, so a subprocess is the honest way to test it).

abankar1:
Closes #2592.

`checkBillingSource()` treated the cloud switches as enabled whenever the variable was **present**, so any non-empty string counted:

```js
.filter((v) => process.env[v])
```

Measured A/B against `main`:

```
                                 BEFORE   AFTER
CLAUDE_CODE_USE_BEDROCK=1        ⚠ warns   ⚠ warns
CLAUDE_CODE_USE_BEDROCK=0        ⚠ warns   ✓ quiet
CLAUDE_CODE_USE_BEDROCK=false    ⚠ warns   ✓ quiet
```

These switches are documented as `=1`, so `=0` is how you turn one off — and `docs/RUNNING_ON_A_BUDGET.md` already points users at them when hunting a stray billing source. Someone who sets `=0` *because* they were chasing exactly this, then runs `npm run doctor`, was told the thing they just disabled was still billing them.

That's worse than a normal false positive given what this check is for. Its own comment says the problem is a user not knowing they're paying per token instead of using the plan they already bought — a billing diagnostic that misreads an explicit opt-out manufactures that confusion, on the check people trust most when they're already worried about cost.

## Fix

Require a truthy value. Matches the repo's own env-flag convention (`=== '1'` in `merge-tracker.mjs:75` and `update-system.mjs:912`), and also accepts `true`/`yes`/`on` because these are third-party switches users copy from assorted docs rather than a convention we control.

The `ANTHROPIC_API_KEY` / `ANTHROPIC_AUTH_TOKEN` paths are untouched.

## Tests

`doctor.mjs` has no exports and no main-module guard — importing it runs the entire diagnostic — so the test drives the real CLI as a subprocess, which is the honest way to test it and needs no refactor of a core file inside a bug fix.

```
✅ CLAUDE_CODE_USE_BEDROCK=1 / =true warn
✅ =0 / =false / =no / =off / ="" stay quiet
✅ CLAUDE_CODE_USE_VERTEX and _FOUNDRY share the rule
✅ an API key is still reported as the billing source
```

`node test-all.mjs --quick` → **3064 passed, 0 failed** · `validate-system-paths-coverage` → OK.

2 commits (fix + test).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved billing-source detection by recognizing common affirmative values consistently, regardless of capitalization or surrounding whitespace.
  * Explicit opt-out values such as `0` no longer trigger cloud billing warnings.
  * Billing checks now behave consistently across supported cloud providers while continuing to report API key usage.

* **Tests**
  * Added coverage for affirmative values, opt-out values, multiple cloud providers, and API key reporting.

### DIFF
--- a/doctor.mjs
+++ b/doctor.mjs
@@ -71,8 +71,16 @@ function checkNodeVersion() {
 function checkBillingSource() {
   const key = process.env.ANTHROPIC_API_KEY;
   const authToken = process.env.ANTHROPIC_AUTH_TOKEN;
+  // Enabled means SET TO A TRUTHY VALUE, not merely present. These switches are
+  // documented as `=1`, so `CLAUDE_CODE_USE_BEDROCK=0` is how someone turns one
+  // off — and mere presence would then report "requests bill to your cloud
+  // account" at exactly the user who just said they don't. A billing check that
+  // misreads an explicit opt-out causes the confusion it exists to remove.
+  // Matches the repo's own env-flag convention (=== '1' in merge-tracker.mjs
+  // and update-system.mjs), while also accepting `true` since these are
+  // third-party switches users copy from assorted docs.
   const cloud = ['CLAUDE_CODE_USE_BEDROCK', 'CLAUDE_CODE_USE_VERTEX', 'CLAUDE_CODE_USE_FOUNDRY']
-    .filter((v) => process.env[v]);
+    .filter((v) => /^(1|true|yes|on)$/i.test(String(process.env[v] ?? '').trim()));
 
   if (cloud.length) {
     return {
--- a/tests/doctor-billing-source.test.mjs
+++ b/tests/doctor-billing-source.test.mjs
@@ -1,0 +1,77 @@
+// tests/doctor-billing-source.test.mjs — the billing-source check in doctor.mjs.
+//
+// The check exists to tell a user which account their tokens are billed to, so
+// a FALSE warning is worse than none: it tells someone who explicitly opted out
+// of a cloud provider that they're billing to it. The cloud switches are
+// documented as `=1`, which makes `=0` the natural way to turn one off.
+import { pass, fail, NODE, ROOT } from './helpers.mjs';
+import { execFileSync } from 'child_process';
+import { join } from 'path';
+
+console.log('\ndoctor.mjs — billing source');
+
+const DOCTOR = join(ROOT, 'doctor.mjs');
+
+/** Run doctor and return its human output (it exits non-zero when it finds issues). */
+function runDoctor(env) {
+  try {
+    return execFileSync(NODE, [DOCTOR], {
+      cwd: ROOT,
+      env: { ...process.env, ...env },
+      encoding: 'utf-8',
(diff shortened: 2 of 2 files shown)


santifer:
You caught this on the same commit that shipped it, @abankar1: `"0"` is truthy in JS, so my presence check told anyone who had explicitly disabled Bedrock that they were billing to it. The check exists to remove that confusion, so a false warning was worse than no warning at all.

Two things I like about the fix beyond the one-liner: pinning it to the repo's own `=== '1'` convention keeps a single spelling for env flags across the codebase, and the test covers the opt-out shapes (`0`/`false`/`no`/`off`/empty) rather than just the bug you hit, so the contract can't silently drift back.

Verified here before merging: reproduced on main with all three controls (`=0` warned, `=1` warned, unset stayed quiet), then confirmed your branch is quiet on `=0` and still warns on `1`/`true` and on a plain API key. Suite 3067/0. Merged 🚀

PR #2595, opened 2026-08-07
fix(merge-tracker): a re-eval kept the superseded report's ✅ on a report with no PDF

### LINKED ISSUE #2594: merge-tracker: a re-evaluation carries the superseded report's ✅ onto a report that has no PDF
### What

When a re-evaluation matches an existing row, `merge-tracker` replaces the row's **report link** but computes the PDF cell as:

```js
const pdf = reportNum && pdfIndex.has(String(reportNum)) ? '✅' : duplicate.pdf;
```

The fallback inherits the **old** row's flag. The PDF cell describes the row's report, and the report just changed — so a `✅` earned by the superseded report is carried onto the new one even when `data/pdf-index.tsv` has no entry for it.

### Measured

Existing row: report `[1]`, PDF `✅`. Incoming re-eval: report `[2]`. Varying only what the manifest contains:

| old flag | pdf-index has | before | after fix | correct |
|---|---|---|---|---|
| ✅ | report 1 only | **✅** | ❌ | ❌ |
| ✅ | reports 1 + 2 | ✅ | ✅ | ✅ |
| ❌ | report 2 | ✅ | ✅ | ✅ |
| ❌ | report 1 only | **✅** | ❌ | ❌ |

Before the change the flag reads `✅` in **all four** cases — it does not track the manifest at all once a re-eval replaces the report link.

### Why it matters

The tracker then asserts a tailored CV exists for the current evaluation when the only PDF on disk belongs to the evaluation that was just superseded — and after #2411 those write through far more often, including downgrades. A user scanning the tracker for what's ready to send sees ✅ and either finds nothing, or attaches a PDF tailored to a different (higher-scored) report.

`sync-pdf-flags.mjs` doesn't rescue it: it only flips ❌ → ✅, never clears a stale ✅.

### Fix

Fall back to the existing flag only when the report is **unchanged**. When it changes, the manifest is the sole authority — no entry means ❌.

PR incoming with the matrix pinned as a regression.

abankar1:
Closes #2594.

The update branch replaces the row's **report link**, but computed the PDF cell as:

```js
const pdf = reportNum && pdfIndex.has(String(reportNum)) ? '✅' : duplicate.pdf;
```

That fallback inherits the **old** row's flag. The PDF cell describes the row's report, and the report just changed — so a `✅` earned by the superseded report rides onto the new one.

## Measured

Existing row: report `[1]`, PDF `✅`. Incoming re-eval: report `[2]`. Varying only the manifest:

| old flag | pdf-index has | before | after |
|---|---|---|---|
| ✅ | report 1 only | **✅** | ❌ |
| ✅ | reports 1 + 2 | ✅ | ✅ |
| ❌ | report 2 | ✅ | ✅ |
| ❌ | report 1 only | **✅** | ❌ |

Before the change the flag reads `✅` in **all four** cases — once a re-eval replaces the report link it stops tracking the manifest entirely. That's a stronger statement than the one-line diff suggests, and it's why I've pinned the whole matrix rather than the single case I first noticed.

## Why it bites now

The tracker claims a tailored CV exists for the current evaluation while the only PDF on disk belongs to the one just superseded. A user scanning for what's ready to send either finds nothing, or attaches a PDF tailored to a different — usually higher-scored — report.

#2411 made this far more reachable: re-evals now write through instead of being discarded, downgrades included, and a downgrade is exactly the case where sending the superseded PDF is worst.

`sync-pdf-flags.mjs` can't rescue it — it only flips ❌ → ✅ and never clears a stale ✅.

## Fix

Fall back to the existing flag only when the report is unchanged. When it changes, the manifest is the sole authority:

```js
const reportChanged = reportNum && oldReportNum && String(reportNum) !== String(oldReportNum);
const pdf = reportNum && pdfIndex.has(String(reportNum)) ? '✅' : (reportChanged ? '❌' : duplicate.pdf);
```

Same-report merges keep inheriting, so a row re-merged against its own report is untouched.

## Tests

Two cases on the existing pdf-sync fixture — the stale flag must clear, **and** a genuinely-present PDF must survive, so the fix can't be read as always clearing. Confirmed the first fails without the code change.

`node test-all.mjs --quick` → **3056 passed, 0 failed**.

2 commits (fix + test).

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected PDF status tracking when a duplicate entry’s report number changes.
  * PDF indicators are now cleared when the replacement report has no available PDF and retained when it does.

* **Tests**
  * Added regression coverage for PDF status updates during report re-evaluations.

### DIFF
--- a/merge-tracker.mjs
+++ b/merge-tracker.mjs
@@ -890,7 +890,17 @@ for (const file of tsvFiles) {
       `${downgrade ? '🔽' : '🔄'} Update: #${duplicate.num} ${addition.company} — ${addition.role} (${oldScore}→${newScore})`
       + (downgrade ? ' — DOWNGRADE, re-eval scored lower' : ''),
     );
-    const pdf = reportNum && pdfIndex.has(String(reportNum)) ? '✅' : duplicate.pdf;
+    // The PDF flag describes THE ROW'S REPORT, and this branch replaces that
+    // report link. Inheriting duplicate.pdf across a report change carried the
+    // superseded report's ✅ onto the new one: the row then claimed a tailored
+    // PDF exists for a report that has none, and the only PDF on disk belonged
+    // to the evaluation that was just superseded. Fall back to the existing
+    // flag only when the report is unchanged; when it changes, the manifest is
+    // the sole authority (#2594).
+    const reportChanged = reportNum && oldReportNum && String(reportNum) !== String(oldReportNum);
+    const pdf = reportNum && pdfIndex.has(String(reportNum))
+      ? '✅'
+      : (reportChanged ? '❌' : duplicate.pdf);
     const updatedLine = buildRow({
       num: duplicate.num, date: addition.date, company: addition.company,
       role: reportNumMatched ? addition.role : duplicate.role,
--- a/merge-tracker.mjs
+++ b/merge-tracker.mjs
@@ -897,7 +897,12 @@ for (const file of tsvFiles) {
     // to the evaluation that was just superseded. Fall back to the existing
     // flag only when the report is unchanged; when it changes, the manifest is
     // the sole authority (#2594).
-    const reportChanged = reportNum && oldReportNum && String(reportNum) !== String(oldReportNum);
+    // "different, INCLUDING one-side-absent". Requiring both to be truthy meant
+    // a row whose report cell is `—` had oldReportNum === null, so reportChanged
+    // was falsy and the stale ✅ was inherited exactly as before this fix — and a
+    // `—` row with a ✅ is ordinary, it is what a tracker entry added before its
+    // evaluation looks like. Both absent stays "unchanged", which is correct.
+    const reportChanged = String(reportNum ?? '') !== String(oldReportNum ?? '');
     const pdf = reportNum && pdfIndex.has(String(reportNum))
       ? '✅'
       : (reportChanged ? '❌' : duplicate.pdf);
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -9043,6 +9043,43 @@ try {
   } else {
     fail('merge-tracker left a newly merged row at ❌ despite a matching pdf-index.tsv entry');
   }
+
+  // A re-evaluation REPLACES the row's report link, and the PDF flag describes
+  // that report. Inheriting the old flag across the change carried the
+  // superseded report's ✅ onto a report with no PDF: the row then claimed a
+  // tailored CV exists, and the only PDF on disk belonged to the evaluation
+  // that had just been superseded (#2594).
+  const reevalRow = '| 3 | 2026-01-04 | Acme | Backend Engineer, Payments | 4.5/5 | Evaluated | ✅ | [1](../reports/001-acme-2026-01-04.md) | first |';
+  const reevalTsv = (n) => ({
+    name: `00${n}-acme.tsv`,
+    content: `${n}\t2026-02-01\tAcme\tBackend Engineer, Payments\tEvaluated\t3.9/5\t❌\t[${n}](reports/00${n}-acme-2026-02-01.md)\tre-eval\n`,
+  });
+
+  const staleFlag = runPdfSyncFixture(
+    'reeval-stale',
+    reevalRow,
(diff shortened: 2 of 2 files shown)


slegarraga (changes requested):
The stale-flag fix is right for numeric report changes. One remaining edge: `reportChanged` requires both report numbers to be truthy, so a transition from a null or unparseable old report to a numeric new report (or vice versa) can still inherit a stale `✅`. Use `reportNum !== oldReportNum` and add regression cases for null-to-number and number-to-null when the new report has no PDF manifest entry. Happy to re-review after that.

santifer:
The diagnosis is right, @abankar1, and the test earns its keep: I ran your `test-all.mjs` against `main`'s `merge-tracker.mjs` and the new assertion fails (`stale PDF flag survived a report change: | ✅ | [2](../reports/002-acme-2026-02-01.md) |`), while your negative control ("keeps ✅ when the NEW report does have a generated PDF") passes. It discriminates, which is the part most added tests skip.

**But the same bug survives one line up, in the guard:**

```js
const reportChanged = reportNum && oldReportNum && String(reportNum) !== String(oldReportNum);
```

`extractReportNum` returns `number | null`, and this demands **both** be truthy. So a row whose report cell is `—` has `oldReportNum === null`, `reportChanged` is falsy, and the code falls straight back to `duplicate.pdf`: the stale `✅` is inherited exactly as before your fix. The row still claims a tailored CV for a report that has none.

That is not a corner case. A row with `—` and a `✅` is what you get from a tracker entry added before its evaluation, which is ordinary, and it is the same silently-wrong-data class you opened this PR for: the PDF on disk belongs to the superseded evaluation, so the user opens a CV tailored to a different job and nothing anywhere says so.

The shape you want is "different, including one-side-absent":

```js
const reportChanged = String(reportNum ?? '') !== String(oldReportNum ?? '');
```

Both null stays unchanged, which is correct. Worth a case for the `—`-to-`[2]` transition alongside the one you have, since the assertion you wrote would not have caught this variant either.

Everything else stands: merge clean onto today's `main`, full suite `3415 passed, 0 failed, 1 warnings` (the warning is the pre-existing `cv-sync-check` one), and the comment you left explaining *why* the manifest becomes the sole authority on a report change is the reason the next person will not undo it.

This touches `merge-tracker.mjs`, which the web reads, so I will flag it there when it lands. Nothing for you to do about that.

Push the null case and it goes in.

abankar1:
Caught cleanly — thank you. Reproduced it before changing anything:

```
row: | ✅ | — |   re-eval brings [2], manifest has no PDF for 2
before: PDF=✅  report=[2]   <- stale ✅ inherited, exactly as you said
after:  PDF=❌  report=[2]
```

Fixed with your shape:

```js
const reportChanged = String(reportNum ?? '') !== String(oldReportNum ?? '');
```

Both-absent still reads as unchanged, which is right.

Full matrix after the change, including the two `—` cases:

| old flag | report | manifest | result |
|---|---|---|---|
| ✅ | `[1]`→`[2]` | has 1 only | ❌ |
| ✅ | `[1]`→`[2]` | has 1+2 | ✅ |
| ❌ | `[1]`→`[2]` | has 2 | ✅ |
| ✅ | `—`→`[2]` | empty | **❌** |
| ✅ | `—`→`[2]` | has 2 | **✅** |

Added the `—`-to-`[2]` case next to the existing one and confirmed it fails without the fix:

```
❌ stale PDF flag survived a —-to-[2] report change: | 4 | … | ✅ | [2](…) |
```

You were right that my original assertion wouldn't have caught this variant — it only exercised `[1]`→`[2]`, so the null branch was never touched. That's the second time on this PR the test needed to be the thing that discriminates rather than the thing that passes.

`node test-all.mjs --quick` → **3057 passed, 0 failed** (the 2 warnings are the pre-existing `cv-sync-check` ones).

santifer:
Merged, @abankar1. The one-side-absent guard shape (String(a ?? '') !== String(b ?? '')) is exactly what the em-dash cell needed, and the new — → [2] test fails without the fix, which is what makes it stick. Thanks for the fast turnaround.

PR #2597, opened 2026-08-07
fix(theme-style): a `$`-pattern in a profile style value splices the body into the CV's <head> (#2588 sweep)

### LINKED ISSUE #2596: theme-style: a $-pattern in a profile style value splices the document body into the CV's <head> (#2588 sweep, two sites missed)
### What

#2588 fixed `String.replace` `$`-pattern interpretation in `build-cv-html.mjs` and `build-cv-latex.mjs`. Two other `<head>`-injection sites do the same thing and weren't covered.

**`theme-style.mjs` — live, user-controlled:**

```js
if (/<\/head>/i.test(html)) return html.replace(/<\/head>/i, `${block}\n</head>`);
```

`block` carries values straight from the user's `config/profile.yml` `style:` section (`accent_color`, `font_family`, `font_size`, `margin`). A string replacement argument is scanned by JS for `$&`, `$'`, `` $` `` and `$$`, and `buildThemeStyleBlock` drops `; { } < >` but has no reason to drop `$` — it's a legal character in a CSS value.

`renderHtmlToPdf` calls this on every render (`generate-pdf.mjs:602-603`), with tokens read from the real profile.

### Reproduced

`style: { font_family: "Fira$'Sans" }` against `<html><head><title>CV</title></head><body>BODY-SENTINEL</body></html>`:

```
<style id="career-ops-dynamic-theme">:root { --font-family: Fira<body>BODY-SENTINEL</body></html>Sans; }</style>
```

The entire document body is spliced **into the `<head>`**, inside the style element. All four patterns corrupt the value:

| value | declared as |
|---|---|
| `Fira$'Sans` | `Fira<body>BODY-SENTINEL</body></html>Sans` |
| `Fira$&Sans` | `Fira` |
| ``Fira$`Sans`` | `Fira<html><head><title>CV</title>Sans` |

Silent, with a valid-looking exit 0 — the same failure shape #2588 describes.

**`generate-pdf.mjs:367` — latent, not live:**

```js
return html.replace(/<\/head>/i, `${pageStyle}\n</head>`);
```

`pageStyle` interpolates only internal constants (`pageSize` is a fixed ternary, `PDF_PAGE_MARGIN` a module constant), so it is **not exploitable today**. Worth fixing anyway because the very next branch in the same function already does it correctly:

```js
return html.replace(/<html\b[^>]*>/i, match => `${match}\n<head>\n${pageStyle}\n</head>`);
```

Two branches, one function, disagreeing about the hazard.

### Fix

Pass a replacer function at both sites, exactly as #2588 did. A replacer's return value is never scanned for patterns.

PR incoming with the four patterns pinned.

abankar1:
Closes #2596.

#2588 fixed `String.replace` `$`-pattern interpretation in the two CV builders. Two `<head>`-injection sites do the same thing and weren't in that sweep.

## The live one: `theme-style.mjs`

```js
if (/<\/head>/i.test(html)) return html.replace(/<\/head>/i, `${block}\n</head>`);
```

`block` carries values straight from the user's `config/profile.yml` `style:` section, and `renderHtmlToPdf` calls this on every render (`generate-pdf.mjs:602-603`). `buildThemeStyleBlock` drops `; { } < >` but has no reason to drop `$` — it's legal in a CSS value.

`font_family: "Fira$'Sans"` produces:

```
<style id="career-ops-dynamic-theme">:root { --font-family: Fira<body>BODY-SENTINEL</body></html>Sans; }</style>
```

The whole document body, spliced **into the `<head>`**, inside the style element. All three non-trivial patterns corrupt it:

| value | declared as |
|---|---|
| `Fira$'Sans` | `Fira<body>BODY-SENTINEL</body></html>Sans` |
| `Fira$&Sans` | `Fira` |
| ``Fira$`Sans`` | `Fira<html><head><title>CV</title>Sans` |

Silent, valid-looking exit 0 — the failure shape #2588 already describes.

## The latent one: `generate-pdf.mjs:367`

Not exploitable today: `pageStyle` interpolates only internal constants (`pageSize` is a fixed ternary, `PDF_PAGE_MARGIN` a module constant). **I'm calling that out rather than inflating it.** It's worth fixing because the very next branch in the same function already does it right:

```js
return html.replace(/<html\b[^>]*>/i, match => `${match}\n<head>\n${pageStyle}\n</head>`);
```

Two branches of one function disagreeing about the hazard is how the next one gets written wrong.

## Fix

Replacer function at both sites, exactly as #2588 did — a replacer's return value is never scanned for patterns.

After: every pattern lands verbatim (`"A$'B"`, `"A$&B"`, ``"A$`B"``, `"A$$B"`) with no leakage, and an ordinary style block is byte-identical.

## Tests

`tests/theme-style-dollar-pattern.test.mjs`, mirroring #2588's file: each of the four patterns must reach the CSS value verbatim **and** must not drag document content into the head; an ordinary block must still inject exactly once; `injectPrintPageCss` must leave the body alone.

Confirmed it fails without the fix — the `$'` case splices `<body>` into `<head>`, and `$&`/`` $` `` corrupt the value differently.

`node test-all.mjs --quick` → **3084 passed, 0 failed**.

2 commits (fix + test).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where dollar-sign sequences in CSS values could be incorrectly interpreted during style injection.
  * Improved print-page style injection while preserving the document content and generated output.

* **Tests**
  * Added regression coverage for special characters in CSS values and print-page style handling.

### DIFF
--- a/generate-pdf.mjs
+++ b/generate-pdf.mjs
@@ -364,7 +364,12 @@ export function injectPrintPageCss(html, format = 'a4') {
   const pageStyle = `<style id="career-ops-page-setup">\n@page { size: ${pageSize}; margin: var(--page-margin, ${PDF_PAGE_MARGIN}); }\n</style>`;
 
   if (/<\/head>/i.test(html)) {
-    return html.replace(/<\/head>/i, `${pageStyle}\n</head>`);
+    // Replacer function, matching the <html> branch just below. `pageStyle`
+    // interpolates only internal constants today, so this is hardening rather
+    // than a live bug — but the two branches sat in one function disagreeing
+    // about it, and the safe spelling is the one that stays correct if a
+    // configurable value is ever interpolated here (#2596).
+    return html.replace(/<\/head>/i, () => `${pageStyle}\n</head>`);
   }
 
   if (/<html\b[^>]*>/i.test(html)) {
--- a/tests/theme-style-dollar-pattern.test.mjs
+++ b/tests/theme-style-dollar-pattern.test.mjs
@@ -1,0 +1,67 @@
+// tests/theme-style-dollar-pattern.test.mjs — the same $-pattern class #2588
+// fixed in the CV builders, in the two <head>-injection sites it did not reach.
+//
+// injectThemeStyle() splices a <style> block carrying values straight from the
+// user's config/profile.yml `style:` section, and did so with a STRING
+// replacement argument. JS scans that argument for $&, $', $` and $$, so a
+// style value containing one of them was interpreted as a replacement pattern
+// rather than inserted literally.
+//
+// The sharp case is $': it means "everything after the match", so the entire
+// document BODY was spliced into the <head>, inside the style element — with
+// no error and a valid-looking exit 0. buildThemeStyleBlock drops `; { } < >`
+// but has no reason to drop `$`, which is a legal character in a CSS value.
+import { pass, fail } from './helpers.mjs';
+import { injectThemeStyle, styleTokensFrom } from '../theme-style.mjs';
+import { injectPrintPageCss } from '../generate-pdf.mjs';
+
+console.log('\ntheme-style / generate-pdf — $-pattern head injection');
+
+const HTML = '<html><head><title>CV</title></head><body>BODY-SENTINEL</body></html>';
+
+// Every $-pattern must land in the CSS value verbatim, and none may drag
+// document content into the head.
+for (const [label, value] of [
+  ["$' (everything after the match)", "Fira$'Sans"],
+  ['$& (the matched text)', 'Fira$&Sans'],
+  ['$` (everything before the match)', 'Fira$`Sans'],
+  ['$$ (an escaped dollar)', 'Fira$$Sans'],
+]) {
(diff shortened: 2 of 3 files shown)


slegarraga (approved):
Clean continuation of #2588. The replacer functions cover both the live and latent `<head>` injection sites, and the tests cover all four `$` patterns plus body preservation. Looks good to me.

santifer:
Sharp catch, @abankar1. I reproduced it from a plausible profile: `font_family: "Inter, \$'Segoe UI', sans-serif"`, the kind of value someone pastes out of SCSS. Pre-fix, `\$'` is read as the replacement pattern meaning "everything after the match", so the entire CV body gets embedded inside the `<style>` block in `<head>`, salary line and all, and the run exits 0 with no error at all. What sold me was the sweep: I went after it with a balanced-paren scan over every `.mjs` and the three remaining string-replacement sites are all inert (`encodeURIComponent` is a function, `plugin-install.mjs` validates the name against `/^[a-z0-9][a-z0-9-]*$/` so `\$` can't get in, `tracker-parse.mjs` takes a literal separator). And your four test patterns are derived rather than copied. Merged 🚀

Issue #2352, opened 2026-07-31
fix(web): prevent confidential-employer CV PDFs from matching the wrong application


Faiyajz:
### Code of Conduct

- [x] I agree to follow the [Code of Conduct](https://github.com/santifer/career-ops/blob/main/CODE_OF_CONDUCT.md)

### What happened?

For a confidential employer tracker entry, career-ops intentionally stores:

Company    Via
━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━
   ?          Example Staffing Agency

PR #2105 fixes the human-facing display for these rows. However, the web CV-PDF and apply flows still us app.company as both a display value and a tailored-CV lookup key.

Two related problems remain:
1. Worker and action labels can show raw ?, for example CV PDF · ?.
2. resolveTailoredCv() slugifies company. When company === "?", the slug is empty, and the current substring match can match every PDF in output/. The apply flow can then select the newest tailored CV, even if it belongs to a different application.

## Steps to reproduce
1. Add a synthetic tracker row with Company = ? and Via = Example Staffing Agency.
2. Ensure multiple tailored CV PDFs exist in output/.
3. Open the row’s web report and use the CV-PDF or apply flow.
4. Observe raw ? in the UI and/or an unrelated tailored CV being selected.

## Expected behavior
 - Human-facing labels use the confidential-employer presentation label.
 - File lookup uses a separate, stable lookup key and must not rely on the display label.
 - An empty company slug must never match every PDF.
 - A confidential application must never attach a CV tailored for another application.
 - Known-company CV-PDF generation and lookup remain unchanged.

## Related work
Follow-up to #2086 and PR #2105. This issue should remain limited to the web CV-PDF/apply-flow lookup and display behavior; it must not rewrite tracker Company or Via data.

### Steps to reproduce

1. Add this synthetic row to a tracker with a Via column:
| NNN | YYYY-MM-DD | ? | Example Staffing Agency | Example Role | X.X/5 | Evaluated | ❌ | [NNN (../reports/NNN-example.md) | Synthetic fixture only. |
2. Ensure output/ contains at least two tailored CV PDFs, including one for an unrelated company.
3. Open the confidential row in the web UI at /pipeline/NNN.
4. Start the web apply flow and reach the tailored-CV attachment step.
5. Observe that the CV lookup can select the newest PDF rather than a CV specific to this application. Worker or action labels can also display CV PDF · ?.

### Version / channel

[...]

PR #2008, opened 2026-07-17
fix(web): enable apply from generated pdf index

gregaro:
## Summary
- Treat data/pdf-index.tsv plus an existing PDF file as server-side proof that a tailored CV is ready
- Pass that readiness into the report page PDF and Apply controls
- Keeps tracker PDF ✅ and in-memory pdf job completion behavior intact

## Testing
- npm run typecheck
- npm test

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PDF availability detection for reports.
  * Report pages now recognize indexed PDFs consistently.
  * PDF generation and application actions reflect the correct readiness status.
  * Report-specific PDFs are served correctly when available.
  * Prevented actions from appearing when associated PDFs are unavailable or missing.
  * Improved handling of invalid, incomplete, unmatched, or unsafe PDF references.
  * Added safer fallback behavior when indexed PDF information is unavailable.

PR #1931, opened 2026-07-15
fix(web): use next/link for decision-card report link

gregaro:
## Summary
- reduce the PR to the still-live improvement requested in review
- replace the decision-card internal report route anchor with Next.js Link after the current decision-card rewrite

## Validation
- rebased onto current main
- CI rerun pending on GitHub

PR #2599, opened 2026-08-07
fix(web): select role-specific tailored CV

gregaro:
## What does this PR do?

Fixes the web apply flow so a tracker application uses the tailored CV mapped to its specific report in `data/pdf-index.tsv`, instead of selecting the newest PDF for the company. The CV preview endpoint uses the same resolver, while manually pasted URLs retain the company-only fallback.

## Related issue

No issue. Bug fix.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] I ran the test suite and all tests pass
- [x] My changes respect the Data Contract

## Testing

- `npm test --prefix web`
- `npm run typecheck --prefix web`
- `node test-all.mjs` (1,795 passed, 0 failed; 2 environment warnings)

## Summary by CodeRabbit

* **New Features**
  * Tailored CVs can now be selected using a specific application identifier.
  * Apply forms automatically pass the relevant application identifier when available.
  * CV PDF downloads support lookup by company or application.

* **Bug Fixes**
  * Improved CV matching accuracy by selecting the exact application report instead of the newest available file.
  * Added validation for missing or unresolved CV requests, with clearer error responses.

* **Tests**
  * Added coverage for report identification, exact PDF matching, and malformed or empty references.


santifer:
Thanks for this, @gregaro.

The web's apply flow picked the most recent PDF for the company instead of the tailored CV that data/pdf-index.tsv maps to that application's specific report: with two applications at the same company, the wrong CV got attached. You thread the application identifier through the whole Apply flow and make the CV preview endpoint use the same resolver, leaving the company fallback only for hand-pasted URLs. Verified with npm test and the web typecheck plus the full suite (1,795 asserts).

santifer (changes requested):
The bug is real and worth fixing: two applications at the same company get two different tailored CVs, and picking "newest PDF for the company" hands the wrong one to an apply flow. Resolving through `pdf-index.tsv` by report number is the correct join.

**Before anything else, there's a collision you couldn't have seen: #2008 (also open) adds a function with the same name and the same job.**

```
#2599 (this)  cv-selection.mjs   pdfPathForReport(indexText, reportNumber)   pure
#2008         career-ops.ts      pdfPathForReport(n)                          reads the file itself
```

Two `pdfPathForReport`s resolving the same manifest. Only one can land as the definition, and **yours is the better shape** — a pure function over text is testable without a filesystem, which is why `cv-selection.test.mjs` can exist at all. #2008 does the I/O inside, so its tests can't reach the parsing logic.

I've asked #2008 for two changes already (it hardcodes the literal `"data/pdf-index.tsv"` instead of the core's `resolvePdfIndexPath`, and skips `containedRealpath` on a path read out of a user-layer file). Those are I/O concerns, which is exactly the half your function doesn't have — so the clean outcome is: **your parser stays, #2008 consumes it for the parsing and keeps only the resolution/containment.** I'll say the same over there.

Also worth knowing: `reportNumberFromCell` here overlaps `reportNumFromCell` from **#1931**, which I've already told that author is now redundant (`findReportFile` on `main` resolves by the row's link since 31-jul). Yours differs though — theirs matches `\[(\d+)\]` specifically, yours takes the first digit run anywhere in the cell:

```js
/(?:^|[^0-9])(\d{1,})(?=[^0-9]|$)/
```

On a normal cell `[041](reports/041-acme-2026-01-01.md)` both give 41. On a cell where someone wrote a bare note before the link, or a Report cell containing a date first, yours takes whatever number comes first. Given this feeds **which CV gets attached to a real application**, I'd anchor it on the link form and return null otherwise — a wrong CV attached silently is worse than no CV found.

## What I'd like

1. Anchor `reportNumberFromCell` on the markdown-link form rather than the first digit run, with a test for a cell that has a leading number in prose.
2. Nothing else changes here — I'll route #2008 to consume your parser.

Sorry for the 7-day wait, and for the fact that two of your PRs today (#1931, this one) turned out to overlap work already in flight. That's a triage failure of mine: three separate PRs have been resolving report identity in three places, and nobody was looking at them together until today.

Lukas-drz:
Not a blocker, just one edge case I hit while independently running into this bug (I opened #3027 for it before finding this PR — now closed as a duplicate of yours, which is the better fix).

`resolveTailoredCv(company, application)` returns `null` when `pdfPathForReport()` finds no row, and the route then renders "no tailored CV found for this offer". Correct. But when the index *does* name a PDF for that report and the file has since been removed from `output/`, `resolveTailoredCv` looks like it returns the recorded path without checking it exists — so `fs.readFileSync` throws and the route answers **500** rather than a 404, which reads as a server fault instead of "that CV isn't there any more".

It happens in practice: `output/` is gitignored and users prune it, while `data/pdf-index.tsv` is append-only and keeps the row.

If it's worth covering, an `fs.existsSync` before returning would do it — the important part being that a missing indexed file must *not* fall through to company matching, since that hands back another role's CV at the same employer and reintroduces exactly what this PR fixes.

Reproduction for the original bug, if you want it as a test fixture: two evaluated roles at one employer, ANYbotics report 80 (Field Deployment Technician, Barcelona) and report 81 (Robotics Test Engineer, Zurich). Both rows opened the Barcelona CV because it was rendered four minutes later. Details in #3025.

santifer (commented):
Re-review, @gregaro — you fixed the two things about *this* PR's own logic: anchoring `reportNumberFromCell` on the full markdown-link form (so `reports/123-role.md` alone doesn't parse as a report number), and validating the tailored-CV files before attach. Both good.

**But the blocker I flagged first is still open, and it's not something you can fix inside this PR alone: this and #2008 collide head-on.** Both are your PRs, both edit `web/src/lib/apply/cv-selection.mjs` and its test, and both `export function pdfPathForReport` / `reportNumberFromCell` — with *incompatible* signatures:

```
#2599 (this)  pdfPathForReport(indexText, reportNumber)   pure, sync
#2008         pdfPathForReport(n)                          async, resolves the index itself via resolvePdfIndexPath (core ACL)
```

Whichever lands second doesn't just conflict in git — it redefines the other's function with a different contract, and the tests import "the" version expecting their own shape. They can't both land as-is.

They also overlap in *purpose* (resolve a report's PDF by number through the index), so the clean resolution is one canonical implementation, not two. My read: **#2008's is the one to keep** — it goes through the core's `resolvePdfIndexPath` ACL and adds `pdfPathStatusForReport`, which is the right layering. So: land #2008 first, then rebase this PR to *import* `pdfPathForReport`/`reportNumberFromCell` from it and adapt to the async signature, rather than redefining them. If you'd rather this one be canonical, say so and we pick deliberately — but it has to be one. Once that's settled I'll do the full pass; the role-specific selection on top is a fix I want.

santifer:
Sorry for the five days, @Lukas-drz. You hit the bug independently, opened #3027, found this PR, closed yours as the duplicate and then came back with an edge case the original doesn't cover. That's three separate courtesies in one thread and it earned a faster answer than this.

The review is the web agent's call rather than mine — `web/` is built and reviewed separately, and I route rather than triage there. Labelled `area:web` and handed across in today's batch with your PR named.

The edge case you describe (the index names a PDF path but the file behind it is gone, so the route can't distinguish "no tailored CV" from "the CV is missing") is worth keeping visible whatever happens to this PR: it's a different failure wearing the same message, which is the kind that costs someone an afternoon. If this PR isn't the right home for it, it deserves its own issue rather than living in a thread comment.

santifer:
#2008 is merged, @gregaro — including the rebase, which I pushed to your branch rather than asking you for a fourth one. That unblocks this PR: the collision I flagged on 2026-08-14 and could not resolve on 2026-08-21 is now settled, and it settled your way on the substance.

**What is on `main` now, so you are not rebasing blind.**

`web/src/lib/apply/cv-selection.mjs` exists, with your #2008 version of it: `reportNumberFromCell`, `pdfIndexEntryForReport`, and `pdfPathForReport(indexText, reportNumber)` built on top. That is byte-for-byte the pure parser this PR introduced, plus the found/empty distinction. So the largest part of this rebase is **deletion**: your copy of that module and its test file both go, and your 42 lines of tests fold into the 29 already there as additions.

`web/src/lib/career-ops.ts` also gained `pdfPathStatusForReport(n)` — async, resolves the manifest through the core ACL, and enforces containment (`output/`-scoped realpath, the symlink case you found). That is the I/O half, and it is the one thing I do not want a second copy of.

**The design question this PR raises, decided.** Your version of `cv-pdf/route.ts` deletes the route's own lookup and delegates to `resolveTailoredCv(company, application)`. **That shape is right and I want it** — one resolver behind both the "View tailored CV" link and the apply upload is exactly what `cv-match.mjs` was reaching for in #2156, and #2008 landed a second lookup inside the route instead. So the merged route is the thing that bends, not your design. Three constraints on how:

1. `resolveTailoredCv` becomes `async` and takes the indexed path from `pdfPathStatusForReport`, rather than re-reading `pdf-index.tsv` or re-deriving containment. One implementation of "is this file safe to serve", not two.
2. The route keeps its **`n`** parameter as well as your `application`. They are different number spaces and both have live callers: `generate-pdf-button.tsx:34` sends `n`, which is the **report** number (`ReportView` is rendered from `reports/{n}-…`), while your `application` is the **tracker row**, resolved to a report number through its Report cell. Collapsing them would be the "#2901 number-space" bug in a new place.
3. The company fallback keeps what landed in #2156 and #2352: `companySlug()` returning null means serve nothing, `matchesTailoredCv` requires the `cv-` prefix and a token boundary, and no first-token fallback. Your `resolveTailoredCv` on `main` already has all three — the fallback branch should be left alone, not rewritten.

**Full disclosure on the conflict count**: this PR went from 2 conflicts to 5 when #2008 landed, and I checked before saying so (`git merge-tree` against both `372cd043` and the current tip). Nobody else's PR moved — #2171, #2997 and #2998 were each already conflicting for other reasons and are unchanged. So the extra three are yours-against-yours, which is the least bad version of this, but it is still a rebase you are doing because of the order I chose.

Once this is up I do the full pass immediately. The role-specific selection is the fix I actually want out of this pair, and it has been waiting on ordering, not on merit.

gregaro:
Rebased onto main (de341ba3), resolving the #2008 collision per your notes:

1. Deleted `web/src/lib/apply/cv-selection.mjs` and its test — took main's version (your parser plus the found/empty distinction) as-is. Folded my extra test cases (empty-string input, a bare filename with no link markers, a realistic multi-column `pdf-index.tsv` row shape) into the existing 3 tests as additions rather than duplicating.
2. `resolveTailoredCv` is now `async` and resolves through `pdfPathStatusForReport` (the core ACL + containment check in `career-ops.ts`) instead of re-reading `pdf-index.tsv` or re-deriving containment itself. Company fallback is untouched — still `companySlug`/`matchesTailoredCv`/`sortNewestFirst` from main.
3. `cv-pdf/route.ts` keeps both `n` (report number) and `application` (tracker row number) as separate params, each falling through to the company-slug heuristic on a miss, same shape as main's existing `n` handling.
4. Reverted my own bug in `generate-pdf-button.tsx` — it was sending `application=${n}` where `n` is actually the report number, not the tracker row; back to main's `?n=`.
5. Updated `apply/fill/route.ts` and the existing `apply-cv-resolver.test.mjs` (main's async-fixture-dependent test suite) for the new async signature.

`npm test` (441/441), `tsc --noEmit`, and `npm run build` all clean.

santifer (approved):
Approving this, @gregaro. You did all three things I asked, and the two I care about most you did in the way that costs you more and the codebase less.

**The spec, checked:**

1. `resolveTailoredCv` is `async` and takes the indexed path from `pdfPathStatusForReport`, so there is exactly one implementation of "is this file safe to serve" and you did not grow a second copy of the containment logic.
2. The route keeps **both** number spaces. `n` is a report number, `application` is a tracker row resolved through its own Report cell, and your comment now spells that out where the next person will read it. That was the part most likely to be collapsed into one parameter, and collapsing it is a bug we would not have noticed for months.
3. The company fallback is untouched: `companySlug()` returning null still serves nothing (#2352), `matchesTailoredCv` still requires the `cv-` prefix and a token boundary (#2156), and there is still no first-token fallback.

And the deletion is the part I want to name, because it is invisible in a diff summary: **your copy of `cv-selection.mjs` and its test are gone**, folded into the one that landed with #2008. Two of your own PRs defining the same function with incompatible signatures was a collision neither of you could see, and you resolved it by giving up your version rather than arguing for it.

**One thing that is now slightly untrue, and it is a comment rather than a behaviour.** The header still says the fallback exists "so this and the apply flow always land on the same file". After this PR they deliberately do not, in one case: when a tracker row has no resolvable report, the route falls through to the company heuristic and `resolveTailoredCv` returns null instead.

I think that asymmetry is **right**, and I would not change the code. Serving a plausible-but-wrong PDF to a viewer is recoverable — the reader sees the filename. Attaching it to a real application is not, which is the whole lesson of #3214. But the sentence promising they always agree is now load-bearing in the wrong direction, and comments are contract in this file.

**I will fix that sentence myself in a follow-up rather than send you round again.** You have been in this cluster since 7 August and the last two rounds were both our ordering, not your work.

santifer:
Merged. That closes the CV-resolver cluster, @gregaro, and closes it on your terms.

You have been in this since 7 August. Two of the rounds were our merge ordering rather than your work, and the last one asked you to delete a module you had written and adopt someone else's. You did it without arguing the point, which is the part that made the cluster resolvable at all.

The one thing I flagged is a comment that is now slightly untrue, not a behaviour, and I said I would fix it myself rather than send you round a fourth time. Doing that next.

santifer:
Seven merges in and your name was not on the credits list, @gregaro. That is a miss on my side, not a threshold you had not reached yet.

@all-contributors please add @gregaro for code

santifer:
@all-contributors please add @gregaro for code

Issue #2600, opened 2026-08-07
scan-history.tsv appends run without a lock, unlike pipeline.md


Scott-Emberson:
`appendToPipeline` takes the shared lock from pipeline-lock.mjs because scan.mjs, scan-ats-full.mjs, and plugins.mjs can run at the same time; the comment above it says exactly that. `appendToScanHistory` (scan.mjs:1714-1733) writes to a file shared by the same three processes with a plain `appendFileSync` and no lock.

What that allows today:

- A reader that opens the file while another process is mid-append can catch a torn final line. The dedup parsers all skip a malformed line quietly, so this surfaces as a row that briefly doesn't count rather than a crash, which also makes it hard to notice.
- The fresh-file path races: `existsSync` check, then header `writeFileSync`, then append. Two scanners starting against a missing file can both take the creation branch, and one header write can land on top of the other's first rows.

This came out of the #2382 review. That change makes a scan take one dedup snapshot of scan-history.tsv at run start, which is only fully sound once writers can't interleave with that read. The fix here is the one the pipeline already has: route scan-history writes (and ideally the snapshot read) through the same lock helper `appendToPipeline` uses.

santifer:
Thanks for this, @Scott-Emberson.

appendToPipeline takes the shared lock from pipeline-lock.mjs because scan.mjs, scan-ats-full.mjs and plugins.mjs can run at the same time (the comment right above says so literally), while appendToScanHistory writes the file those same three processes share with a bare appendFileSync. Two consequences, named with surgical precision: a torn final line that dedup parsers skip in silence (a row that stops counting for a while, not a crash, so it's hard to notice) and the new-file race, where two scanners both clear existsSync and one header write lands on top of the other's first rows. It came out of reviewing #2382, which takes a dedup snapshot at startup and is only sound if writers don't interleave with that read.

Where it stands: triaged as a bug. I do not label anything `confirmed` until I have reproduced it myself, so that is the next step here rather than a yes or a no.

PR #2601, opened 2026-08-07
perf(scan): read each dedup source once per scan run

### LINKED ISSUE #2382: perf(scan): scan-history.tsv is read and parsed three times per run (pipeline.md and applications.md twice)
One scan reads and fully parses `data/scan-history.tsv` three separate times:

- `loadSeenUrls` (scan.mjs:923)
- `loadSeenCompanyRoles` → `collectSeenCompanyRoles` (scan.mjs:1315)
- `loadFingerprintHistory` (scan.mjs:1482)

`pipeline.md` and `applications.md` are each read twice by the same functions.

## Measured

| History size | Total across the three loaders |
|---|---|
| 5,000 rows | 56 ms |
| 50,000 rows | 592 ms |

Plus roughly three times the file's bytes in transient strings, since each pass splits the whole file again.

The largest single slice sits inside `loadSeenUrls`: `normalizeUrlForDedup` costs about 4.5 µs per URL (a `new URL()` construction, a `searchParams` iteration, and a `toString()`), which is 227 ms of the 361 ms that loader takes at 50k rows.

## Fix

Read each file once and hand the text to the parsers. `collectSeenCompanyRoles` already accepts text directly; the other two need text-taking siblings, with the current path-defaulted signatures kept as thin wrappers so existing tests and callers are unaffected.

Separately, `normalizeUrlForDedup` can take a fast path that returns the input unchanged when it has no query string, no fragment, no trailing slash, and is already lowercase. That describes most ATS permalinks, and it skips the URL parse entirely for them.

Neither change alters dedup semantics. Both are mechanical.

Worth noting this is not the dominant term in a scan (see the cross-listing issue for that), but it is the cheapest one to fix and it scales with history size, which only grows.

Scott-Emberson:
Closes #2382, following the constraints from [the acceptance comment](https://github.com/santifer/career-ops/issues/2382#issuecomment-5216110737): per-run only, no dedup behaviour change, `normalizeUrlForDedup` left alone.

## What changed

- New text-taking exports `collectSeenUrls(sources, policy)` and `collectFingerprintHistory(text)`. The parsing bodies moved verbatim out of `loadSeenUrls` and `loadFingerprintHistory`, which are now thin path wrappers with unchanged signatures. All three loaders now share the shape `collectSeenCompanyRoles` already had: a pure text-to-structure core plus a filesystem wrapper. Existing callers (scan-ats-full.mjs, scan-interamt.mjs) and existing tests are untouched.
- New `loadDedupSnapshot(policy, canonicalize)` reads scan-history.tsv, pipeline.md, and applications.md once each through the existing `readIfExists` (absent file reads as `''`, which every parser treats the same as the old `existsSync` skip) and derives all four dedup structures from that single read. The scan entrypoint calls it once; the cross-listing check consumes `snapshot.fingerprintHistory` instead of re-reading the file.
- Nothing is cached at module level. The snapshot lives in locals inside the run and dies with it, so a later run or a long-lived agent session always re-reads the files.

## Behaviour proof

`tests/scan-dedup-snapshot.test.mjs` pins golden outputs that were captured by running the pre-refactor loaders (at 402e798) against a fixture, then hard-coded. The fixture covers the branches most likely to drift silently: a recheck-aged `added` row, a permanent status, an active and an expired cooldown, a legacy 7-column row, and a cosmetic utm param. A second check runs the same fixture without a recheck TTL to prove the policy argument actually reaches the collectors, since a wiring mistake there (for example a double-wrapped options object) would disable the recheck feature without failing anything else. Fresh-install and zero-byte-file cases are pinned as equivalent.

One semantic note: the fingerprint history used to be re-read after fetching. Every `appendToScanHistory` call happens later in the run than that read, so within a run the snapshot is byte-identical to what the re-read saw. What changes is visibility of rows appended by a *concurrent* process mid-run. Those appends hold no lock, so the old re-read could equally catch a torn half-written line; the honest fix is locking the TSV the way pipeline.md already is, filed as #2600.

## Measured

50k-row history, 200-row pipeline, 100-row tracker, mean of 5 runs after warmup (Node 22, Windows):

| | 5k rows | 50k rows |
|---|---|---|
| before (7 file parses) | 51.3 ms | 427.8 ms |
| after (3 file parses) | 46.6 ms | 415.3 ms |

Smaller than the issue's framing suggested, and worth being plain about: the duplicate reads and splits were the cheap part. Most of the loader cost sits in `normalizeUrlForDedup` (the issue measured ~227 ms of the 361 ms URL pass at 50k), which is out of scope here per the acceptance comment and would be the follow-up with the actual large win. What this PR buys is the structural half: one coherent snapshot per run, roughly a third of the transient string allocation, and the `collect*` seam that makes the later work safe to do.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved duplicate detection across scan history, pipelines, and applications.
  - More reliably handles URL normalization, legacy records, missing or empty files, cooldowns, statuses, and recheck timing.
  - Prevents inconsistent results caused by repeatedly reading changing scan data during a run.

- **Tests**
  - Added coverage for deduplication, fingerprint tracking, recheck behavior, and edge cases involving incomplete or missing data.

### DIFF
--- a/scan.mjs
+++ b/scan.mjs
@@ -1026,38 +1026,43 @@ export function normalizeUrlForDedup(url) {
   return parsed.toString();
 }
 
-export function loadSeenUrls(policy = {}) {
+/**
+ * Build the seen-URL set from already-read source texts. An absent file is
+ * passed as '' (the readIfExists convention shared with
+ * `collectSeenCompanyRoles`) — every parse below yields nothing on ''.
+ */
+export function collectSeenUrls(sources = {}, policy = {}) {
+  const { scanHistoryText = '', pipelineText = '', applicationsText = '' } = sources;
   const seen = new Set();
   let recheckEligible = 0;
 
   // scan-history.tsv
-  if (existsSync(SCAN_HISTORY_PATH)) {
-    const lines = readFileSync(SCAN_HISTORY_PATH, 'utf-8').split('\n');
-    for (const line of lines.slice(1)) { // skip header
-      const [url, firstSeen, , , , status = 'added'] = line.split('\t');
-      if (!url) continue;
-      if (shouldDedupScanHistoryRow({ firstSeen, status }, policy)) seen.add(normalizeUrlForDedup(url));
-      else recheckEligible++;
-    }
+  for (const line of scanHistoryText.split('\n').slice(1)) { // skip header
+    const [url, firstSeen, , , , status = 'added'] = line.split('\t');
+    if (!url) continue;
+    if (shouldDedupScanHistoryRow({ firstSeen, status }, policy)) seen.add(normalizeUrlForDedup(url));
+    else recheckEligible++;
   }
 
   // pipeline.md — extract URLs from checkbox lines
-  if (existsSync(PIPELINE_PATH)) {
-    const text = readFileSync(PIPELINE_PATH, 'utf-8');
-    for (const match of text.matchAll(/- \[[ x]\] (https?:\/\/\S+)/g)) {
-      seen.add(normalizeUrlForDedup(match[1]));
-    }
+  for (const match of pipelineText.matchAll(/- \[[ x]\] (https?:\/\/\S+)/g)) {
+    seen.add(normalizeUrlForDedup(match[1]));
   }
 
   // applications.md — extract URLs from report links and any inline URLs
-  if (existsSync(APPLICATIONS_PATH)) {
-    const text = readFileSync(APPLICATIONS_PATH, 'utf-8');
-    for (const match of text.matchAll(/https?:\/\/[^\s|)]+/g)) {
-      seen.add(normalizeUrlForDedup(match[0]));
-    }
+  for (const match of applicationsText.matchAll(/https?:\/\/[^\s|)]+/g)) {
+    seen.add(normalizeUrlForDedup(match[0]));
   }
 
   return { seen, recheckEligible };
+}
+
+export function loadSeenUrls(policy = {}) {
+  return collectSeenUrls({
+    scanHistoryText: readIfExists(SCAN_HISTORY_PATH),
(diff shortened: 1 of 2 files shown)


slegarraga (approved):
Clean structural refactor. The pure collectors plus filesystem wrappers preserve the existing signatures, and the golden tests are the right safety net for behavior equivalence. Keeping the snapshot run-scoped avoids stale module-level state while opening up the later `normalizeUrlForDedup` optimization. Looks good to me.

santifer:
Merged, @Scott-Emberson. I did not trust the hand-written goldens, so I checked equivalence the hard way: a differential probe importing the old and new `scan.mjs` in the same process, across 490 cases the fixture does not cover (CRLF, no trailing newline, missing header, malformed URL and date, unicode, broken cooldown, empty status, a URL duplicated across all three sources, times five policies, plus 400 random fixtures). 490/490 identical. And your four GOLDEN blocks still match the pre-refactor loaders on today's main, so the copied table is derived and current rather than stale. I also checked `readIfExists` is `existsSync ? readFileSync : ''` with no try/catch, which matters: a real EACCES still throws instead of being swallowed as empty. 🚀

One follow-up worth your five minutes rather than mine. Your test pins the `policy` argument as live (check #2), but not the other one `main()` varies: `buildCompanyCanonicalizer(config.company_aliases)`. I mutated `scan.mjs:1742` to drop that third argument and the new test passes 7/7 with the whole suite still at 3409/0. If that wiring ever breaks, `company_aliases` silently stops applying to company+role dedup and the user just sees the same job listed twice under two spellings. The refactor did not create that gap, but it is the change that makes the argument losable. A sibling assertion in the file you already touched would close it.

PR #702, opened 2026-05-20
feat(scan): VC portfolio scanner — P1-2 Getro firms (Accel, Insight, Khosla)

mitwilli-create:
## Summary
- Adds **P1-2** from `data/input-quality-roadmap.md` — VC portfolio job-board scanner. VC boards are an early-signal discovery channel: ~25% of portfolio companies post on a VC's board before their own ATS goes live.
- Working today: **3 Getro firms** (Accel, Insight Partners, Khosla Ventures). Smoke test: 425 title-matches → 246 after cross-VC dedup. Idempotent on re-run (state file + URL dedup).
- Stubbed for Phase 2: **6 Consider firms** (a16z, Sequoia, Greylock, Lightspeed, Bessemer, NEA — Cloudflare blocks plain curl; needs CDP-attached Playwright) and **1 custom Vue firm** (Index Ventures — API not yet mapped).

## Files
- `config/vc-portfolios.yml` — declarative firm list with verified `collection_id`s
- `scripts/scan-vc-portfolios.mjs` — paginated `POST api.getro.com/api/v2/collections/{id}/search/jobs`, sort=-created_at, incremental via state file, hang-safe via `lib/fetch-utils.mjs::fetchWithTimeout`, 50-page hard cap, 3-firm `poolMap` concurrency, 5-min global runtime guard
- `scripts/launchd/com.mitchell.career-ops.scan-vc-portfolios.plist` — daily 09:45 PT
- `.gitignore` — adds `data/vc-portfolios-state.json` to match existing per-scanner state patterns

## Endpoint verification (spec assumptions were wrong)
The task spec said the pattern is `https://jobs.{firm}.com/api/jobs?limit=...` for Getro hosts and Khosla uses Pallet. Both were wrong:
- The naive path returns 404 on every host (those domains serve the frontend HTML, not the API).
- Real Getro endpoint: `POST api.getro.com/api/v2/collections/{collection_id}/search/jobs` body `{page, per_page<=20, sort: "-created_at"}` → `{results: {jobs, count}}`. No auth.
- Khosla migrated off Pallet to Getro.
- Total Getro jobs: Accel 24,766, Insight 11,735, Khosla 9,391.

Endpoint discovery used curl + Chrome MCP network panel against a live page session to confirm shapes before writing any scraper code.

## Test plan
- [x] `node --check scripts/scan-vc-portfolios.mjs`
- [x] `plutil -lint scripts/launchd/com.mitchell.career-ops.scan-vc-portfolios.plist`
[...]

PR #2602, opened 2026-08-07
feat(providers): add getro VC-portfolio job board provider

ilyakanevskiy:
## What does this PR do?
Adds `providers/getro.mjs` — a scanner provider for job boards built on the Getro SaaS product, used by VC firms to aggregate their portfolio companies' open roles into one searchable board.

Not auto-detected (opaque vanity domains: `careers.atomico.com`, `talent.cherry.vc`, ... — no common host suffix to key off), so it's opt-in only via `provider: getro`. `collection_id` auto-resolves from the tenant's own `careers_url` page (the `__NEXT_DATA__` blob embeds it server-side, no browser/JS execution needed) — no manual lookup required; an optional `getro: { collection_id: N }` override is available as a fallback.

Verified live against four VC portfolio boards (all public, zero-auth): Atomico, Cherry Ventures, HV Capital (note: this board's own frontend fails to load in a real browser due to a bug on HV's side; irrelevant here since the provider only reads the static page + backend API), and Point Nine. Also spot-checked against the three funds #702 independently confirmed the same API shape for — Accel, Insight Partners, and Khosla Ventures — same result: `collection_id` auto-resolves, pagination walks past page 1 correctly.

Pagination has its own `DEFAULT_MAX_PAGES=100` (`MAX_PAGES_CAP=1500`), independent of whatever the source reports via `results.count` — a growing/spoofed board must never drive an unbounded fetch loop. `MAX_PAGES_CAP=1500` isn't arbitrary headroom either — Accel's board alone is ~26.5k listings (~1,329 pages at 20/page), so the cap is close to what's actually needed for the largest known Getro board.

@mitwilli-create — saw #702 (closed only because it targeted the wrong repo) confirmed the same API shape for Accel/Insight/Khosla independently of this work; noticed you're actively developing your own fork, so flagging this in case it's useful there.

@giacomoguidotto — heads up: this supersedes the `getro` board type in `career-ops-plugin-startup-boards` (`lib/getro.mjs`), since this core provider paginates the full board instead of reading only the first page. Your plugin's `ycombinator` and `consider` board types are unaffected (different mechanisms). Happy to open an issue on your repo too if that's a better place to track it — let me know what's easiest on your end.

## Related issue
None — new zero-auth scanner provider, exempt from issue-first per CONTRIBUTING.md.

## Type of change
- [x] New feature (scanner provider)

## Checklist
- [x] Read CONTRIBUTING.md
- [x] New provider — issue-first exempt
- [x] No personal data included
- [x] `node test-all.mjs` passes (3128 passed, 0 failed — 1 pre-existing unrelated warning)
- [x] Respects the Data Contract (system-layer files only)
- [x] Aligns with the project roadmap

## Test plan
- `node test-all.mjs --only providers/getro` — 21/21 pass: `detect()` always null, `__NEXT_DATA__` collection_id extraction, job normalization, `fetch()` auto-resolution + manual override, error handling, own `DEFAULT_MAX_PAGES` independent of a huge `results.count`, `ctx.maxPages`, all-zero-string `collection_id` override rejection.
- `node test-all.mjs` — full suite green.
- Live smoke test against all four core-verified boards (Atomico, Cherry Ventures, HV Capital, Point Nine) — `collection_id` resolves correctly, real jobs come back, pagination works past page 1.
- Live spot-check against the three #702 boards (Accel, Insight Partners, Khosla Ventures) — same behavior confirmed at larger scale (11.5k-26.5k listings each).

## Summary by CodeRabbit

* **New Features**
  * Added support for Getro job boards with opt-in configuration and automatic collection discovery.
  * Added paginated job searches with remote-work labels, company and location details, and annual salary information.
  * Added example configurations for several portfolio career pages.

* **Documentation**
  * Documented Getro setup, collection ID options, search behavior, and pagination defaults.

* **Bug Fixes**
  * Improved validation and safe handling for invalid listings, malformed responses, and interrupted searches.


ilyakanevskiy:
@coderabbitai review

ilyakanevskiy:
@coderabbitai review

ilyakanevskiy:
@coderabbitai review

ilyakanevskiy:
Re: [SSRF outside-diff comment](https://github.com/santifer/career-ops/pull/2602#pullrequestreview-4886154711) on `resolveCollectionId` (`providers/getro.mjs:83-126`) — not addressing this one, and want to explain why rather than silently skip it.

`ctx.fetchText(careersUrl, ...)` does fetch a URL taken from `entry.careers_url`, with no host allowlist or loopback/private/link-local/metadata blocking. But that's the same pattern every other provider in this codebase already uses — `workday.mjs`, `greenhouse.mjs`, `ashby.mjs`, `lever.mjs`, and ~45 others all fetch directly from `entry.careers_url`/`entry.api` with no SSRF allowlisting.

That's a deliberate trust boundary in this project, not an oversight: `portals.yml` is a user-owned, locally-administered config file (see the Data Contract in `AGENTS.md`), not untrusted external input. It's the operator's own config — equivalent to them running `curl <url>` themselves — not attacker-controlled content like a job posting or an API response. SSRF hardening in this codebase is reserved for URLs derived from untrusted external data, which is why `providers/getro.mjs`'s *second* fetch (the actual search API) uses a hardcoded `API_ORIGIN` constant rather than anything config-derived.

Adding a host allowlist only to this provider would be inconsistent with the rest of the architecture and wouldn't address the actual threat model here (a user editing their own local config isn't the attack surface this project defends against). Given that, I'm leaving `resolveCollectionId` as-is — happy to revisit if the project's trust model for `portals.yml` changes generally, but that'd be a cross-provider decision, not a Getro-specific fix.

ilyakanevskiy:
Closing in favor of #2640.

#825 merged a getro provider independently while this was open, so rather than duplicate it, #2640 builds directly on top of the merged version instead — keeps its age-cutoff pagination design and adds auto-resolve for collection_id, salary parsing, retry-on-transient-failure, and a few other bits carried over from this PR.

Details in #2640's description.

Issue #1851, opened 2026-07-13
bug(upskill): targeted mode inverts known-skill suppression — CV skills reported as gaps, real gaps hidden


Scott-Emberson:
Post-merge review of #1796 (the #1739 targeted mode; review committed in #1520). The feature's JD side correctly reuses `extractSkills()`, but the known-skills side re-implements suppression instead of using the same canonicalization — and the two vocabularies disagree badly enough that **the output can be the exact inverse of the correct answer**.

## Reproduction (merged main)

CV contains: `k8s, C++, .NET, JavaScript, PostgreSQL, MongoDB, LLMs`
JD asks for: `Kubernetes, C++, .NET, Java, SQL, Go, LLMs`

| | correct (aggregate semantics) | targeted mode today |
|---|---|---|
| gaps | Java, SQL, Go | **Kubernetes, C++, .NET, LLMs** |
| suppressed as known | Kubernetes, C++, .NET, LLMs | **Java, SQL, Go** |

Wrong on all 7 skills, both directions.

## Why

- `upskill.mjs` ~588–594: known-skill matching does substring `.includes()` over **raw lowercased regex tokens** — `'c\+\+'`, `'\.net'`, `'node\.?js'`, `'llms?'` contain literal backslashes and can never match real CV text, so those CV skills are never recognized as known (→ shown as gaps).
- ~596–601: a second hardcoded `essentialSuppression` list does bare substring matching — `'java'` ⊂ "java**script**", `'sql'` ⊂ "postgre**sql**", `'go'` ⊂ "mon**go**db" (→ real gaps hidden as known). This is exactly the over-suppression failure the phase-1 comments (~105–109) and self-test guards (~399–408) exist to prevent; the targeted path bypasses both.
- Output `knownSkills` emits raw tokens (`k8s`, `postgres`) where aggregate emits canonical names (`Kubernetes`, `PostgreSQL`) — same skill, two names across modes.

## Fix (~15 lines)

Replace the parallel implementation with the aggregate path's own call — `const known = extractSkills(cvText + profileText)` (same as aggregate ~311–315) — compare canonical-to-canonical, delete `essentialSuppression`, emit canonical names. Extract the targeted branch out of its inline IIFE into an exported function and add the 7-skill fixture above to `runSelfTest()` (the targeted path currently has zero test coverage).

## Also in the same PR (registration gaps, non-blocking but the feature is currently unreachable by docs)

[...]

PR #2603, opened 2026-08-07
feat(skill-extract): recognize certifications, not just engineering tools

LayerC0de:
## Problem

Every token in `SKILL_TOKENS` is an engineering tool — Python, Kafka, Terraform, PyTorch, Snowflake. There is no PMP, PMI-ACP, ITIL, PRINCE2, CISSP, TOGAF or Six Sigma. That makes the shared vocabulary structurally blind to the gap class that screens out delivery, program-management, governance and security candidates: **credentials**.

The effect is worse than a missing feature, because `upskill` does not degrade gracefully. On a corpus where certification requirements appeared an order of magnitude more often than any tool-shaped gap, the ranked output still consisted entirely of one- and two-report tool mentions. The credential requirements were not ranked low — they were **invisible**. A user gets no signal that an entire category is unreadable, so a thin result reads as *"no gaps"* rather than *"cannot see them."*

This matters beyond one profile. `AGENTS.md` explicitly invites retargeting ("change the archetypes to data engineering roles"), and PMP / ITIL / CISSP are among the most common hard screens in enterprise postings. Any PMO lead, program manager, compliance or infosec user gets a tool that silently cannot answer the question they asked it.

## Change

16 tokens across five families, plus matching `CANONICAL` entries:

| Family | Tokens |
|---|---|
| PMI | PMP, PMI-ACP, PgMP, CAPM, PMBOK |
| Scrum | PRINCE2, Certified ScrumMaster, CSPO |
| Service management | ITIL, COBIT, TOGAF |
| Quality | Lean Six Sigma, Six Sigma |
| Security | CISSP, CISM, CIPP |

The `CANONICAL` entries matter: without them `"pmp"` canonicalizes through `DISPLAY` to `"Pmp"`, missing the known-skills set — the exact drift class this module was extracted to prevent (#1896, #1851).

## Two deliberate exclusions

**`SAFe` is matched case-sensitively**, in its own pass, mirroring the existing `Go` handling. `safe` is an everyday English word; a case-insensitive token fires on *"a safe working environment."*

**`CSM` is omitted entirely.** It is a valid Certified ScrumMaster abbreviation, but in job-posting text it far more often means Customer Success Manager. Including it would turn every customer-success role into a phantom certification gap. `Certified ScrumMaster` is listed in full instead and carries no such collision.

Longest-first ordering within each family (`Lean Six Sigma` before `Six Sigma`) follows the existing `React Native`-before-`React` convention.

## Compatibility

No existing token, alias or pattern is modified. Tool-shaped gaps rank exactly as before; this is purely additive.

## Verification

Run on this branch (rebased onto current `main`), on Windows 11:

- 11 new extraction cases, including the `safe`-in-prose guard and the `CSM` collision guard
- `node upskill.mjs --self-test` — OK
- `node jd-skill-gap.mjs --self-test` — 48 passed, 0 failed
- `node test-all.mjs` — **3111 passed, 0 failed**, 1 warning (the plugin-manifest symlink-traversal test skipping on a non-elevated Windows shell, which is the intended graceful skip)

## Related

While preparing this I hit a separate, pre-existing bug in how `upskill.mjs` derives its known-skills set — YAML comments in `profile.yml` register as skills the user has, silently suppressing real gaps. Filed as #2604 rather than fixed here, since the right fix is a design call. This PR does not cause it, but it does widen the vocabulary, so more comment text becomes capable of triggering it.

## Summary by CodeRabbit

* **New Features**
  * Expanded skill extraction to recognize certifications, frameworks, and methodologies including PMP, ITIL, Scrum, Lean Six Sigma, CISSP, and SAFe.
  * Added standardized display names for newly supported credentials and methodologies.
  * Improved handling of ambiguous terms to reduce incorrect skill matches.

* **Tests**
  * Added coverage for certification detection, case-sensitive SAFe matching, and preservation of distinct “safe” meanings.


LayerC0de:
Good catch — confirmed and fixed in a2f50dd.

The finding is correct and I verified it before changing anything:

```
canonicalize("safe")  => "SAFe"     <- the bug
canonicalize("SAFe")  => "SAFe"
extractSkills("a safe working environment") => []   <- extractor was never affected
```

`extractSkills()` was fine, because it only reaches `SAFe` through the case-sensitive `SAFE_CERT_PATTERN`. But `canonicalize()` is public API and `jd-skill-gap.mjs` calls it, so prose "safe" became a credential on that path. The `CANONICAL` entry re-opened through the exported function precisely the hole that keeping `SAFe` out of `SKILL_TOKENS` closes for the extractor.

Removing the key costs nothing. Both spellings now fall through to the unknown-token pass-through: `'SAFe'` returns `'SAFe'`, `'safe'` returns `'safe'`. Recognition remains solely `SAFE_CERT_PATTERN`.

Added four regression tests to `tests/skill-extract.test.mjs`:

- certification tokens are recognized (PMP, PMI-ACP, ITIL, CISSP, SAFe)
- `"Maintain a safe working environment; safety is our priority."` does not yield `SAFe`
- `canonicalize('safe') === 'safe'` and `canonicalize('SAFe') === 'SAFe'` — the one you asked for
- `"part Customer Success Manager (CSM), partnered with an AE"` yields nothing

The last one is beyond what you flagged, but the `CSM` exclusion was previously only argued for in a comment with no test holding it, which is how it would quietly regress.

`tests/skill-extract.test.mjs` 11/11, `upskill --self-test` OK, `jd-skill-gap --self-test` 48/48, `test-all.mjs` 3115 passed / 0 failed.

LayerC0de:
Done in 7958e16 — table-driven over all 16 tokens.

Took the suggestion as written, and asserted **both** halves per token rather than just recognition:

```js
for (const [raw, display] of certificationCases) {
  const found = extractSkills(`Requires ${raw} certification.`);
  if (!found.has(display))            certFailures.push(...);
  if (canonicalize(raw) !== display)  certFailures.push(...);
}
```

The canonicalization assertion is the one that earns its keep. A token added to `SKILL_TOKENS` without a matching `CANONICAL` entry falls through to `DISPLAY`, which title-cases it — `"pmp"` → `"Pmp"` — and misses the known-skills set. That is the #1851 drift class this module was extracted to prevent, and it fails silently, so recognition-only coverage would pass while the gap map quietly broke.

Added one case beyond the suggestion: `"Lean Six Sigma Black Belt preferred."` must yield `Lean Six Sigma` and **not** `Six Sigma`. The longest-first ordering was argued for in a comment (following the existing `React Native`-before-`React` convention) but nothing held it.

`SAFe` stays asserted separately rather than in the table, since it is reached only through the case-sensitive `SAFE_CERT_PATTERN` and is deliberately absent from both `SKILL_TOKENS` and `CANONICAL` — folding it into the table would imply a `CANONICAL` entry that must not exist.

`tests/skill-extract.test.mjs` 13/13 · `test-all.mjs` 3117 passed / 0 failed.

slegarraga (approved):
Good expansion, and the tests are thoughtful. The `SAFe`/`safe` boundary and the `CSM` exclusion are exactly the false-positive guards this vocabulary needs. The table-driven extraction plus canonicalization coverage closes the silent drift class. Looks good to me.

santifer:
Good problem to pick, @LayerC0de. `upskill` and `jd-skill-gap` being blind to credentials meant a PMO or program manager got "no gaps" when the truth was "I cannot see your gaps", and those two read identically from the outside.

The case-sensitive SAFe guard genuinely works, which was my first suspicion: I threw `"keep the platform safe... go the extra mile"` at it and got back only Kubernetes, Terraform and Go. Zero false positives. And making the test table-driven over all 16 tokens, asserting both recognition *and* canonicalization, is the assertion that earns its keep: a token added to `SKILL_TOKENS` without a `CANONICAL` entry falls through to `DISPLAY` and gets title-cased, which is exactly the kind of drift nobody notices.

**One case costs a user something.** You dropped `CSM` on the grounds that "Certified ScrumMaster is listed in full instead", but the only token is the fused trademark spelling. Run through `classifySkillGaps`, a CV that says `Certified Scrum Master` (with the space, which is how most people write it) comes back as `GAP: ["Certified ScrumMaster"]`, while `PMP` resolves fine.

That is the failure this PR exists to prevent, one spelling later: the tool tells someone they lack a certification they hold and are showing on their CV. Worse than the original silence, because now it is confidently wrong, and the advice downstream is "go get certified" for something already earned.

Fix is small: recognize the spaced form as an alias of the fused one, and keep `CSM` as a third alias rather than removing it. The acronym is what appears in the wild most often anyway.

Everything else checks out: SHA matches, merge clean against today's `main`, full suite `3419 passed, 0 failed, 1 warnings` (the warning is the pre-existing one), CI 12/12 green.

Push the spacing alias and it goes in.

santifer:
Still waiting on the ask from 11-ago, @LayerC0de: HEAD is `7958e166` from the 7th, so I think it simply has not gone up. Re-reproduced today so you have it concrete: with your patch applied, a CV writing "Certified Scrum Master" (with the space) still comes out as a GAP for "Certified ScrumMaster", while "PMP" resolves fine. Same damage the PR exists to prevent, one spelling later.

The exact spots: `skill-extract.mjs:59` (SKILL_TOKENS has only the fused spelling) and `skill-extract.mjs:141` (CANONICAL has only `certified scrummaster`). Both need the spaced form as an alias resolving to the same canonical.

**Asking once instead of twice, because five siblings share the bug** and the same fix shape covers them: PMI ACP, PRINCE 2, Certified Scrum Product Owner, Lean Six-Sigma and Six-Sigma. Fixing only ScrumMaster leaves four credentials producing the identical false positive.

Two heads-ups so you do not get stuck:

- **Adding `CSM` as an alias breaks `tests/skill-extract.test.mjs:122-124`**, which asserts the opposite on purpose. If you want CSM in, change that test deliberately and say so; if you would rather leave CSM out, that is fine too. Your call, just do it consciously.
- If CSM goes in, `CAPM` and `CIPP` deserve the same look: they already match as certifications inside finance and construction prose, which is the same homonym collision CSM was excluded for. Consistency either way.

Everything else is clean: merges clean, suite 3659/0, CI 12/12, purely additive. It is a good PR one small commit short.

LayerC0de:
Pushed as `0c52718` — sorry for the delay, you were right that it simply had not gone up. Both halves of the ask are in.

## All six families

You asked once for all of them rather than twice for one, so here is the before/after on each:

| CV spelling | JD spelling | before | after |
|---|---|---|---|
| `Certified Scrum Master` | `Certified ScrumMaster` | **GAP** | none |
| `PMI ACP` | `PMI-ACP` | **GAP** | none |
| `PRINCE 2` | `PRINCE2` | **GAP** | none |
| `Lean Six-Sigma` | `Lean Six Sigma` | **GAP** | none |
| `Certified Scrum Product Owner` | `CSPO` | **GAP** | none |

Reproduced all five as phantom gaps against `7958e16` first, then confirmed each returns clean — and in both directions, so a fused CV against a spaced JD works too.

Both halves were needed, and only the second is obvious in hindsight: the **token** makes the spelling match, and the **CANONICAL alias** makes it collapse to the same display string as its fused sibling. Without the alias the spaced form extracts as its own string, never meets a known-skills set built from the other spelling, and is *still* a gap — the same #1851 drift class reached by a different road.

## CSM: kept out, consciously

Taking your "your call, just do it consciously" — it stays out, so `tests/skill-extract.test.mjs:122-124` stands as written rather than being changed.

The trade runs the opposite way from the spellings above, which is what decided it. A missing alias costs one real credential, in a note the user can see and correct. Admitting `CSM` would mint a phantom gap out of every customer-success posting the scanner touches — and the collision lands hardest on exactly the people this vocabulary serves: program, product and delivery managers, whose postings are the ones that read "part Customer Success Manager (CSM)". Both unambiguous spellings are now recognized, so the credential is still found whenever it is written out.

Since CSM is out, `CAPM` and `CIPP` stay in unchanged — your follow-up on those was conditional on CSM going in, and neither has an everyday expansion competing for the acronym.

## One thing I corrected while in there

The comment justifying the CSM exclusion cited *"lessons-learned 2026-07-03"* and `portals.yml → title_filter.negative`. I went to verify it and **neither is checkable from the repo** — `portals.yml` is gitignored user data. So the exclusion was resting on evidence nobody reviewing it can see. It now points at the fixture in the test file, which is in-repo and pins the exact sentence the exclusion is about. Same decision, verifiable justification.

Certification table grows 16 → 22 cases, each alias asserted beside its fused sibling so an edit that changes one display string and not the other fails there rather than in a user's gap map.

`tests/skill-extract.test.mjs` all green · `test-all` 3108 passed, 1 failed (the pre-existing `js-yaml` resolution failure in the temp-dir e2e copy, identical on `main`).

Scott-Emberson (approved):
Good addition, and the safety framing is right. `tests/skill-extract.test.mjs` grows by +84/-0: a 22-row table that asserts both `extractSkills` recognition and `canonicalize` per token, the Lean Six Sigma precedence over the shorter Six Sigma, a combined cert-plus-SAFe case, and the `canonicalize('safe')` regression. These are real controls, not decoration: revert the feature and the table loop, the precedence check, and the combined check all fail, since `extractSkills` returns empty and `canonicalize` passes the lowercase token straight through. Hermetic, same `pass`/`fail` shape as its neighbours, failure messages name the offending set. Both CodeRabbit rounds are already addressed on this head. Approving from the tests/ side.

Two nits, both optional, neither gates this:

1. `tests/skill-extract.test.mjs:123`, the SAFe boundary case under-tests the guard it targets. The impl's whole point at `skill-extract.mjs:101` is `SAFE_CERT_PATTERN = /(?<!\w)SAFe(?!\w)/`, documented right above it to keep `"SAFety"` from matching while still allowing `"SAFe 6"`. The prose negative only feeds lowercase `"safe"`/`"safety"`, which the pattern rejects trivially on case alone, so the `(?!\w)` trailing boundary, the part that actually stops `"SAFety"`, has nothing holding it. Drop `(?!\w)` and `extractSkills('SAFety training programs')` would mint a phantom `SAFe` with this suite still green. Worth one line: `extractSkills('SAFety training programs').has('SAFe') === false`, and optionally a `"SAFe 6"` positive.

2. `tests/skill-extract.test.mjs:136`, `csm.size === 0` is broader than the claim. It guards the CSM-is-not-a-cert exclusion but asserts the entire set is empty, so a future token that happens to match a word in that fixture ("Manager", "AE") would fail this test for an unrelated reason. A targeted `!csm.has('Certified ScrumMaster')` says exactly what it means. Low value, take it or leave it.

Neither is a reason to hold. Approving.

LayerC0de:
Thanks @Scott-Emberson — took both, and the first one was not a nit. Pushed as `add80cd`.

**The SAFe boundary hole is real, and I verified it rather than taking it on faith.** You were exactly right that the existing negative only fed lowercase `safe`/`safety`, which the pattern rejects on **case** alone — so the leading `(?<!\w)` was held by that test and the trailing `(?!\w)` by nothing:

```
with (?!\w) dropped:
  extractSkills('SAFety training programs')  ->  ['SAFe']
  tests/skill-extract.test.mjs               ->  entirely green
```

A phantom certification, and the suite would have shipped it. Added `SAFety` as you suggested, plus the `SAFe 6` positive you flagged as optional — it turns out to be the necessary pair, because the boundary has to reject a trailing *letter* without also rejecting the trailing space-and-digit the framework's own versioned name uses. Confirmed the new assertion fails against the broken pattern and passes against the real one.

**Second one taken too.** `csm.size === 0` was asserting more than the claim: a token added later that legitimately matches "Manager" or "AE" in that fixture would have failed a test that is only about CSM. Now `!csm.has('Certified ScrumMaster')`, which says what it means.

Both were worth having, and the first one closed a gap I introduced.

LayerC0de:
@coderabbitai correct, and it is the hole my previous commit opened. Fixed in `8bb63a9`.

Worth naming the tension, since the two reviews pull opposite ways and both are right:

- @Scott-Emberson flagged that `csm.size === 0` **over**-asserts — a token added later that legitimately matches "Manager" or "AE" in the fixture would fail a test that is only about CSM.
- You flagged that my narrowed `!has('Certified ScrumMaster')` **under**-asserts — it checks one value is absent, so `CSM` arriving as its own token walks straight past it.

Verified rather than reasoned about. With `CSM` temporarily added to `SKILL_TOKENS` and `CANONICAL`:

```
extractSkills('CSM required.')  ->  ['CSM']
the contextual assertion        ->  still green   ← the gap
```

So the exclusion is asserted in both halves now, because each alone leaves the hole the other closes:

- **contextual** — the collision sentence does not yield the credential, scoped to that one value so unrelated vocabulary growth cannot fail it
- **standalone** — `CSM` does not become a skill under *any* name

New assertion fails against the admitted token (`❌ extractSkills standalone CSM => CSM`) and passes against the real vocabulary. Both directions confirmed.

santifer:
The part I'd have got wrong is the one you argued hardest for: leaving `CSM` OUT. It's a legitimate abbreviation, and the instinct is to add it for completeness — but in job-ad prose it expands to Customer Success Manager far more often, and the collision lands on exactly the people this vocabulary was added for. A phantom "go get certified" for a credential the posting never asked for is advice with no fix attached, and it'd fire on every customer-success posting the scanner touches. Both unambiguous spellings are in, so nothing real is lost.

Two things that make this more than a token list:

**The alias half is what makes the token half count.** `'certified scrum master' → 'Certified ScrumMaster'` is the difference between recognizing a credential and telling someone to earn one they already hold. Without the CANONICAL entries the tokens would have shipped and *looked* like they worked.

**`SAFe` handled case-sensitively, and deliberately absent from CANONICAL.** That second half is the subtle one — `canonicalize()` lowercases before reading the map, so a `'safe'` key would have re-opened through the exported function the exact everyday-word hole that keeping it out of `SKILL_TOKENS` closes for `extractSkills()`. Two doors into the same bug, and you closed both.

The evidence in the header is what settles the "is this scope creep" question: PMP in 41 of 138 reports, ITIL in 26, against a top-ranked visible gap scoring 1.9 off a single report. The vocabulary was the constraint, not the scoring. Table-driven fixture over every token, asserting recognition AND canonical form, is the right shape for a list that will grow.

Merged 🚀

PR #155, opened 2026-04-10
fix: use candidate name in generated resume filename

shra1-honade:
## Summary

- Reads `name` from `config/profile.yml` and normalizes it to kebab-case (e.g. `"John Doe"` → `"john-doe"`)
- Uses `{candidate}` in both the intermediate HTML file and final PDF output filename
- Replaces the generic `cv-candidate-{company}` pattern with `cv-{candidate}-{company}`

This makes it easier to identify resumes when multiple candidates or profiles are in use, and is consistent with how the Canva workflow already names its output files.

## Test plan

- [ ] Generate a PDF for a job offer
- [ ] Verify the output file is named `output/cv-{your-name}-{company}-{date}.pdf` instead of `output/cv-candidate-{company}-{date}.pdf`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1311, opened 2026-06-29
feat(providers): add 4 Day Week provider


santifer:
**Time:** ~45 min · **Difficulty:** Beginner

[4 Day Week](https://4dayweek.io) is a remote-jobs board (4-day-week roles) with a public, zero-auth JSON API. Adding it as a provider lets `scan.mjs` pull its listings with no API key. (Part of umbrella #230.)

### What to do
1. Create `providers/4dayweek.mjs` following the existing provider contract.
2. Fetch the public API: `https://4dayweek.io/api/jobs` (verified `200`, zero-auth — returns JSON).
3. Map each job into the provider's normalized shape (title, company, location, url, postedAt).
4. Add a small test block in `test-all.mjs` mirroring another JSON provider's tests.

### Acceptance
- `providers/4dayweek.mjs` exports the provider following the same interface as `providers/himalayas.mjs`.
- SSRF guard: HTTPS + hostname allowlist (`4dayweek.io`) + `redirect: 'error'`.
- NaN-safe date parsing; **zero new dependencies**.
- `node test-all.mjs` passes with your new test block.

### Pattern to mirror
`providers/himalayas.mjs` or `providers/personio.mjs` — JSON-API providers.

Scott-Emberson:
Working on this one.

PR #154, opened 2026-04-10
feat: add Cursor/IDE support + auto-apply mode

UGilfoyle:
- .cursorrules: full intent routing for Cursor IDE users (no Claude Code needed)
- .cursorignore: IDE-specific ignore rules
- docs/IDE-SETUP.md: setup guide for Cursor/Windsurf/Copilot users
- modes/auto-apply.md: Playwright-based automated form submission
- CLAUDE.md + .cursorrules: synced auto-apply capability across both platforms
- .gitignore: protected cv.md, profile.yml, scratch scripts from public exposure
- build-html.cjs: HTML-to-PDF CV generation helper

PR #2610, opened 2026-08-07
fix(followup): ignore an apply date cited about another row

### LINKED ISSUE #2607: followup-seed/cadence: a note citing ANOTHER row's apply date wins, pinning the follow-up to the wrong day
## Summary

`parseAppliedDate()` returns the **first** `applied YYYY-MM-DD` match in a row's notes. It has no notion of *whose* apply date it found. A note that cross-references another application's apply date — which is normal when two roles at the same employer are live at once — wins over the row's own date, and `followup-seed` pins the follow-up to the wrong day.

It also silently overrides a **correct** date column, because `resolveAppliedDate()` prefers notes and only falls back to `app.date`.

## Location

```js
// followup-cadence.mjs
export function parseAppliedDate(notes, options = {}) {
  if (!notes) return null;
  const validateCalendar = options.requireValidCalendarDate === true;
  for (const m of String(notes).matchAll(/\bapplied\s+~?(\d{4}-\d{2}-\d{2})(?![\w-])/gi)) {
    if (!validateCalendar || isRealCalendarDate(m[1])) return m[1];   // <-- first match wins
  }
  return null;
}
```

## Reproduction

A real note from my tracker. Two ServiceNow roles were live simultaneously, so row #155's note references #154 before stating its own date:

```js
const note = 'STRATEGY DECISION NEEDED: #154 Sr PM M&A is already live in the same ATS '
           + '(applied 2026-08-04) - recommend applying anyway. ... '
           + 'APPLIED 2026-08-06 - submitted directly by Jason.';

parseAppliedDate(note)                                   // => '2026-08-04'   ✗
resolveAppliedDate({ notes: note, date: '2026-08-06' })  // => { appliedDate: '2026-08-04', appDateSource: 'notes' }
```

The row's date column said `2026-08-06`. `followup-seed 155` pinned the first follow-up to **2026-08-11 instead of 2026-08-13** — two days early.

The `appDateSource: 'notes'` label makes this worse rather than better. The module's own comment says the label exists so a caller can distinguish a measured date from the evaluation-date proxy — but here it reports a **real but wrong** date as `notes`, i.e. measured. That is exactly the failure the comment at lines 158-160 warns about, reached by a different route:

> a real but WRONG date labelled `notes` — i.e. measured, not inferred. Falling back to the evaluation date is honest by comparison

## Why it is easy to hit

It needs only that a note mentions another row's apply date *before* its own. That happens whenever:

- two roles at the same employer are live and the notes explain the sequencing
- a note records "same posting as #NNN, applied YYYY-MM-DD"
- an evaluation cites a sibling requisition's timeline as context

The more carefully the notes are written, the more likely this fires.

## Possible fixes

Not sending a patch — the disambiguation rule is a judgement call:

1. **Take the LAST match rather than the first.** One-character change, and correct for the common "context first, own status last" note shape. Still guessable-wrong if the order is reversed.
2. **Prefer a match not preceded by a `#NNN` reference within N characters.** Targets the actual signal (the date belongs to a cited row), but is heuristic.
3. **Require an anchored form for the row's own date** — e.g. only `APPLIED YYYY-MM-DD` at the start of a note segment counts — and treat everything else as prose.
4. **Cross-check against the date column and refuse when they disagree**, degrading to the labelled fallback rather than pinning a date the row itself contradicts.

(4) is the most conservative and would have caught this specific case, since the column already held the right answer.

## Not a bug, noted for context

`--force` appending a second pin rather than replacing is **documented** (`appendPins`, "it just appends a fresh pin"), so correcting a mis-seed leaves two pins for one row and the stale line has to be removed by hand. Fine as designed — but it does mean a wrong seed is slightly awkward to repair, which raises the cost of this bug.

## Related

Same family as #2604: a value is read out of free text that was never meant to be parsed as data.

LayerC0de:
Closes #2607.

## Problem

`parseAppliedDate()` returned the **first** `applied YYYY-MM-DD` match in a row's notes, with no notion of *whose* date it found. Notes routinely cite a sibling requisition's timeline for context, and that citation reads exactly like this row's own apply date to a positional scan.

Real note from row #155, which had two ServiceNow roles live at once:

```
STRATEGY DECISION NEEDED: #154 Sr PM M&A is already live in the same ATS
(applied 2026-08-04) - recommend applying anyway. ...
APPLIED 2026-08-06 - submitted directly by Jason.
```

`followup-seed 155` pinned the first follow-up **two days early**. The row's `date` column already said `2026-08-06`.

## Fix

Drop cross-referenced matches **before** choosing, then keep first-wins among what remains.

A `#NNN` row reference shortly before the date, with no sentence boundary between them, means the date sits inside that reference's clause. A sentence break ends the reference's scope:

```js
parseAppliedDate('Sibling #140 was slow. Applied 2026-08-06.')   // => '2026-08-06'  ✅
```

**First-wins is preserved for a row's own dates.** A later status date still must not displace the submission date — `test-all.mjs` asserts that, and cross-reference filtering is orthogonal to the ordering rule, so the existing contract is untouched.

**When every apply-date belongs to another row, it now returns `null`** rather than adopting a foreign date:

```js
parseAppliedDate('Same posting as #140 (applied 2026-07-20). Not yet submitted.')  // => null
```

The note genuinely does not say when *this* row was submitted, so degrading to the labelled evaluation-date fallback is the honest answer. Reporting a real-but-foreign date as `appDateSource: 'notes'` means **measured** — precisely the failure the function's own header comment warns against:

> a real but WRONG date labelled `notes` — i.e. measured, not inferred. Falling back to the evaluation date is honest by comparison

## One option from the issue I deliberately did NOT take

The issue floated cross-checking the notes date against the tracker `date` column and refusing on disagreement. **That is wrong**, and I'd rather say so than ship it: the `date` column is the *evaluation* date, and applying days later is the normal case. Requiring agreement would discard a correct notes date on most rows. It only looked attractive because it happened to fix the single reported example — which is a good reminder that a fix validated against one case isn't validated.

## On the heuristic

Detection is a heuristic (a 120-char lookback for `#NNN`, scope ended by sentence punctuation). Both failure directions are survivable, which is what makes one acceptable here — a false positive degrades to the labelled fallback, a false negative is today's behaviour. Neither invents a date.

## Verification

Four new assertions in `test-all.mjs` beside the existing `parseAppliedDate` fixtures: the repro, the sentence-boundary case, the all-foreign case, and the `resolveAppliedDate` labelling.

`test-all.mjs` **3115 passed, 0 failed** · `followup-cadence.test.mjs` 21/21.

Checked against a real 46-row tracker: the affected row now resolves to its own `2026-08-06` instead of the cited `2026-08-04`, and **no other row changes**.

## Summary by CodeRabbit

* **New Features**
  * Added reliable application-date fallback using tracker dates, evaluation dates, or today when appropriate.
  * Seeded entries now identify whether dates came from explicit input, notes, tracker data, evaluation date, or today.
  * CLI output identifies proxy or assumed dates.

* **Bug Fixes**
  * Improved handling of multiple dates and cross-referenced application IDs.
  * Preserves valid dates across sentence boundaries and separators while ignoring foreign-only dates and labeled requisition or posting IDs.

* **Tests**
  * Added regression coverage for date sources, labeled IDs, cross-references, boundaries, and fallback behavior.


LayerC0de:
Correct — fixed in fdd90de. Reproduced first:

```
parseAppliedDate('#154 is already live; applied 2026-08-04. ...')
  => '2026-08-04'   ✗ foreign date accepted as this row's
```

A semicolon joins independent clauses *inside* one sentence, so the subject carries across it — `#154 is already live; applied 2026-08-04` is still about #154. `.`/`!`/`?` genuinely end the reference's scope; `;` does not. Removed it from the boundary set and added the regression case you suggested.

Worth naming why this matters beyond the one input: it puts the code back in line with the argument the PR makes for tolerating a heuristic at all. The two error directions are **not** symmetric — a false positive degrades to the labelled evaluation-date fallback, while a false negative reports another row's date as this row's *measured* one. Where the reading is ambiguous the tie should go to "cross-referenced", and treating `;` as a break was tilting it the wrong way. So this is a correctness fix and a consistency fix.

Verified nothing else moved:

| case | result |
|---|---|
| the #2607 repro | still `2026-08-06` |
| period still ends scope (`Sibling #140 was slow. Applied 2026-08-06.`) | `2026-08-06` |
| first-wins preserved | `2026-06-09` |
| plain note with an unrelated semicolon (`Applied 2026-06-09 via Personio; raised part-time`) | `2026-06-09` — unaffected, since the semicolon is not preceded by a row reference |

Real 46-row tracker: 33 rows still resolve from notes and the three live applications are unchanged.

`test-all.mjs` 3116 passed, 0 failed · `followup-cadence.test.mjs` 21/21.

santifer:
The core fix is right and I reproduced it: the note resolves `2026-08-04` → `2026-08-06`, so an apply date cited *about another row* was being read as this row's. And refusing the tracker-date cross-check the issue floated was the correct call, for the reason you gave.

**But the safety claim the whole design rests on is not true of the primary consumer.**

The PR says four times that a false positive "degrades to the labelled evaluation-date fallback". `followup-seed.mjs` is the script named in #2607's own title, and its fallback is:

```js
// followup-seed.mjs:139
  return todayStr();
```

Not the evaluation date, and not labelled. So when the matcher returns null, the row silently gets **today**. That inverts the risk you argued from: a wrong-but-labelled date is something the user can see and correct, while today's date looks like a normal fresh entry and quietly resets the follow-up clock for that application. Someone who applied three weeks ago drops out of the chase queue with nothing on screen to suggest it.

I am not asking you to change the matcher: the tightening is good and the reasoning behind it is sound. What needs to change is either the claim or the fallback:

- **Fallback**: have `followup-seed` fall back to the evaluation date, labelled, which is what the PR already describes. Then the argument holds as written.
- **Or the claim**: if `todayStr()` is deliberate there, say so plainly in the PR and drop the "degrades safely" framing, so the next person reading it does not inherit a guarantee the code does not give.

I would take the first, since the second leaves a real silent-reset path in place. Either way the property should be pinned by a test, because right now nothing fails if that fallback changes again.

Merge is clean against today's `main`, suite `3407 passed, 0 failed`.

You are 13 merges in and this is the second PR in a row where you refused a suggestion for a reason that turned out to be correct. That is worth more than the fix itself. Push the fallback and I will take it same day.

santifer:
Reiterating the ask from 11-ago, @LayerC0de, but with the part I owed you: what it actually costs, because that is probably why it stalled.

The claim I disputed still stands. `followup-seed.mjs:130-139`: when the notes DID carry application dates but every one of them was classified as belonging to another row, `resolveAppliedDate` falls through to `todayStr()` **unlabelled**. So the phrase "degrades to the labelled evaluation-date fallback" at `followup-cadence.mjs:187`, `:214` and `:218` describes something the code does not do.

**Two ways to close it. Pick either, and here is the honest cost of each:**

**(a) Make the fallback distinguishable.** This is what I asked for, and I owe you a warning I did not give the first time: it collides head-on with `followup-seed.mjs:21-24`, `:116-119` and `followup-seed-tests.mjs:208-209`, which document and assert that the `date` column is **never** consulted. That is a deliberate invariant, so (a) is not a drop-in change. **You have my explicit permission to touch those three** if you take this route: that is the maintainer decision that was blocking you, and it should not have been left implicit.

**(b) Fix the prose instead.** Remove the "labelled fallback" sentence from those three places and describe what each consumer really does. Smaller, entirely legitimate, and it stops the next reader inheriting a guarantee that does not exist.

Either way, one test: a note carrying application dates that are ALL classified as belonging to elsewhere, asserting what `followup-seed` returns. `followup-seed-tests.mjs:220` only covers the note with no dates at all, which is why this path shipped uncovered.

Optional and not blocking, but it reinforces (a): scoping `isCrossReferencedMention` so a `#NNN` preceded by req/job/posting/ref does not count as a row reference. The vocabulary already exists at `merge-tracker.mjs:207`.

Suite is green either way (3658/0). The blocker was never your code, it was me not saying which invariant you were allowed to break.

LayerC0de:
Took route **(a)** — the fallback, not the prose. Pushed as `55089c0`. Thank you for the explicit permission on the invariant; that was exactly the blocker.

You were right, and the risk ran the opposite way to the one I argued from. `resolveAppliedDate` now returns `{appliedDate, appDateSource}` over **explicit > notes > `date` column > today**, and every consumer reports the source.

Measured on a row whose notes cite only sibling requisitions:

```
before   Seeded #1: next follow-up 2026-08-21 (applied 2026-08-14, set 2026-08-14)
after    Seeded #1: next follow-up 2026-07-27 (applied 2026-07-20,
             from the evaluation date — notes carry no apply date, set 2026-08-14)
```

**25 days of silent delay**, on the exact note shape this PR exists to detect. `--json` carries `appDateSource`; the human-readable line names the fallback where the user actually looks, since "nothing on screen to suggest it" was the real complaint.

Now consistent with the `appDateSource` in `followup-cadence.mjs` and the `dateBasis` in `company-history.mjs` — all three disclose an unmeasured date the same way.

**On the invariant.** I broke it at `followup-seed.mjs:21-24` and `:116-119` and in the tests, and recorded *why* in both docblocks rather than only in the commit: the column is still the evaluation date and still wrong, but it is wrong in the direction you can see. Its priority is unchanged — `--date` and a notes date both still outrank it — so the old assertion now pins that ordering instead of the prohibition.

**The test you asked for**, and it is the one that mattered: notes that *do* carry apply dates, all of them cross-references. Asserts the labelled evaluation date, and separately that it is **not** today. The pre-existing case only covered notes with no dates at all — which is precisely why this path shipped uncovered.

**Took the optional scoping too.** `Req #1311 - applied 2026-08-06` now keeps its date instead of discarding it, using the label vocabulary that `merge-tracker.mjs` already reads from this same column. Four assertions for the labelled forms, plus one that a **bare** `#NNN` in the same shape is still a cross-reference — the exemption must not blunt the rule it scopes.

Also fixed the three "labelled fallback" comments in `followup-cadence.mjs` to say which consumer does what, rather than asserting a guarantee one of them did not give. That was the honest half of (b), and it is worth having even though (a) makes the claim true.

`followup-seed-tests` 45/45 · `followup-cadence.test` 21/21 · `test-all` 3112 passed, 1 failed — that one being the pre-existing `js-yaml` resolution failure in the temp-dir e2e copy (`ERR_MODULE_NOT_FOUND`, raised before any test logic runs; fails identically on `main`, and CI installs the dep so it is green there).

santifer:
The direction is right and the branch is **already better than `main`**, so read this as one character rather than a rejection.

Six cases, run against both:

| note | `main` | your branch |
|---|---|---|
| `applied 2026-06-01` | ✅ own date | ✅ own date |
| `#154 … (applied 2026-08-04); applied 2026-06-15` | ✗ **2026-08-04** (foreign) | ⚠️ `null` |
| `#154 … (applied 2026-08-04)` | ✗ 2026-08-04 | ✅ `null` |
| `Req #1311 - applied 2026-08-06` | ✅ own date | ✅ own date |
| `See #154. applied 2026-06-15` | ✅ | ✅ |

Row 4 is the one I want to praise: `REQ_LABELLED_HASH_RE` keeping a row's own posting id from reading as a cross-reference is the subtle half of this, and you got it, with the vocabulary matched to `merge-tracker.mjs`'s own.

**Row 2 is the ask.** A note that cites a sibling **and then states this row's own date** loses its real date. The cause is the boundary rule: `isCrossReferencedMention` clears a `#N` only at `/[.!?]\s/`, so a semicolon does not end the citation, `#154` still covers the second match 120 characters later, and `own` comes out empty.

That case is the *more* common shape of the scenario #2607 describes. Two roles live at the same employer, and the note that mentions the sibling is usually the same note that records this one's submission.

Verified fix, one character class:

```js
return !/[.!?;|]\s/.test(window.slice(refEnd));
```

With `;` and `|` treated as boundaries, all six cases above pass, including row 2 returning `2026-06-15`. Nothing else moved: suite on your branch stayed at 3120 passed, 0 failed. `;` and `|` are the separators the Notes column actually uses, so this is following the data rather than widening the rule.

Worth a test for the mixed shape specifically, since it is the one a future simplification would silently re-break.

To be clear about the stakes so this does not read as blocking on a nicety: even unfixed, your branch turns a **foreign date reported as measured** into a labelled fallback, which your own comment argues for correctly. I would rather land it once with row 2 covered than land it twice.

Ping me when it is up and I will take it the same day.

LayerC0de:
Row 2 is fixed and pushed as `7fc6126` — thank you for the six-case table, that made the gap unambiguous.

One thing to flag before you merge, because it changes the fix: **I could not reproduce "suite stayed at 3120 passed, 0 failed" with the one-character version.** The branch asserts the neighbouring case at `test-all.mjs:6527`, and `[.!?;|]\s` breaks it:

```
#154 Sr PM (applied 2026-08-04); applied 2026-06-15   ->  2026-06-15  ✅ row 2 fixed
#154 is already live; applied 2026-08-04              ->  2026-08-04  ❌ foreign, reported as MEASURED
```

That second one is why `;` was excluded originally. So the character class buys row 2 at the cost of the failure this function exists to prevent — and in the worse direction, since a foreign date labelled `notes` is measured-but-wrong rather than an honest fallback.

## What separates them is not the separator

It is whether the citation has already been **given** a date:

- `#154 Sr PM (applied 2026-08-04); applied 2026-06-15` — `#154` already carries its date, so what follows the separator is a new subject: this row's.
- `#154 is already live; applied 2026-08-04` — `#154` has no date yet, so the one after the separator is still its own. A semicolon joins independent clauses within a sentence and the subject carries across it.

So `;` and `|` end the scope **only when an apply-date already appears between the reference and the separator**, checked against the text before the *last* separator so a date belonging to the citation cannot be read as belonging to a later clause. Your point about following the data stands — these are the separators the column uses; they just are not unconditional.

## All nine

| note | expected | got |
|---|---|---|
| `applied 2026-06-01` | own date | ✅ `2026-06-01` |
| `#154 … (applied 2026-08-04); applied 2026-06-15` | **row 2** | ✅ `2026-06-15` |
| `#154 … (applied 2026-08-04)` | `null` | ✅ `null` |
| `Req #1311 - applied 2026-08-06` | own date | ✅ `2026-08-06` |
| `See #154. applied 2026-06-15` | own date | ✅ `2026-06-15` |
| `#154 … (applied 2026-08-04) \| applied 2026-06-15` | pipe variant | ✅ `2026-06-15` |
| `#154 is already live; applied 2026-08-04` | `null` | ✅ `null` |
| the original #2607 note | own date | ✅ `2026-08-06` |
| `Same posting as #140 (applied 2026-07-20)…` | `null` | ✅ `null` |

Tested the mixed shape next to the unsatisfied-reference one deliberately, as you suggested — they differ only by whether a date precedes the separator, so a future simplification that collapses them fails loudly instead of silently re-breaking row 2.

`followup-cadence.test` 21/21 · `followup-seed-tests` 45/45 · `test-all` 3115 passed, 1 failed (the pre-existing `js-yaml` resolution failure in the temp-dir e2e copy, identical on `main`).

**If you would still rather have the one-character version**, say so and I will push it and delete the `6527` assertion — it is your call which of the two shapes the column really sees, and mine is the one that came from a bot review rather than from your data. I did not want to drop it silently to make the simpler fix look clean.

LayerC0de:
@coderabbitai correct, and it was the same loss one keystroke away. Fixed in `5699fc8`.

`#154 Sr PM (applied 2026-08-04);applied 2026-06-15` kept the citation in scope because the separator rule demanded trailing whitespace. That requirement was borrowed from the sentence rule above it, where it earns its keep — `[.!?]` without it fires on "3.5" and "e.g.". Neither `;` nor `|` appears inside a number or an abbreviation, so here it was cost with no benefit.

Three cases added, including the unspaced form of the **negative** case — `#154 is already live;applied 2026-08-04` is still the citation's date — so widening the separator cannot be mistaken for widening the rule. 12 shapes now covered, all passing.

`followup-cadence.test` 21/21 · `followup-seed-tests` 45/45 · `test-all` 3118 passed, 1 failed (the pre-existing `js-yaml` temp-dir resolution failure, identical on `main`).

santifer:
What makes this mergeable is that you argued the asymmetry instead of just picking a rule: a false positive degrades to a date that is LABELLED as not-measured, a false negative reports another row's date as this row's own. Once that's stated, "the tie goes to cross-referenced" stops being taste and becomes the only defensible default.

The semicolon distinction is the part I'd have got wrong. `"#154 Sr PM (applied 2026-08-04); applied 2026-06-15"` and `"#154 is already live; applied 2026-08-04"` really do want opposite readings, and "has the reference already been given a date?" is a clean test for which one you're in — not a special case bolted on after a failing fixture.

Two things I checked rather than took on trust: `REQ_LABELLED_HASH_RE` is anchored at the end, so `Req #1311 - applied ...` keeps its own date instead of being read as a pointer at another row; and the 120-char lookback is short enough that an unrelated `#123` earlier in a long note can't reach forward to disqualify a real date. Both were the places this could have quietly eaten measured dates.

Local run: 3126 passed, 0 failed. Merged 🚀

PR #2612, opened 2026-08-08
fix(cv-facts): a k/M/B magnitude suffix let an inflated claim past the gate

darkpandawarrior:
## The bug

`COUNT_CLAIM_RE` captures the digits of a metric claim but not an adjacent magnitude suffix. The modifier window then re-consumes that letter as a generic word, so `50k users` normalizes to the claim `50 users` — which matches a CV that says 50.

The gate therefore passes a 1000× inflation while correctly catching a smaller one:

```js
auditClaims('Reached 900 users', 'Reached 50 users.')
// → invented: ["900 users"]      correct

auditClaims('Reached 50k users', 'Reached 50 users.')
// → invented: []                 ← the bug
```

Order-of-magnitude inflation is the most damaging fabrication a CV can carry, and it was the one shape the anti-fabrication gate could not see. It also fails in the unsafe direction: the gate reports clean.

## Fix

One regex change, and it only brings the count pattern in line with what this same file already does for money — `CURRENCY` at line 73 already consumes `(?:\s?[kKmMbB])?`.

```diff
-\b(\d[\d,.]*)\s*\+?\s*…
+\b(\d[\d,.]*(?:[kKmMbB]\b)?)\s*\+?\s*…
```

`[kKmMbB]\b` requires the suffix to **end the token**, which is what keeps two existing behaviours intact:

| input | before | after |
|---|---|---|
| `50k users` | `50 users` | `50k users` |
| `50 million users` (spelled out) | `50 users` | `50 users` — unchanged |
| `50kg units` (`k` not at a boundary) | `50 units` | `50 units` — unchanged |

I deliberately did **not** copy currency's `\s?`, which would allow a space: on counts that would capture `50 m` out of `50 million` and change the spelled-out case. Both preserved cases have tests pinning them, so a future loosening fails loudly.

## Tests

Five cases added to the existing `--self-test` (39 → 44):

- an inflated magnitude suffix is caught
- a magnitude claim the source supports still passes
- a lowercase `m` suffix is caught too
- a spelled-out magnitude is unaffected
- a unit beginning with a suffix letter is unaffected

Verified they actually catch the bug: with the regex reverted and the tests kept, 2 fail. (Stashing the whole file proves nothing, since the tests live in it too.)

`node test-all.mjs` → **3110 passed, 0 failed**.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recognition of numeric claims using magnitude suffixes such as `k`, `M`, and `B`, including lowercase variants.
  * Preserved valid claims such as `50k` while correctly excluding suffix-prefixed units such as `50kg`.
  * Added support for spelled-out magnitudes and improved handling of inflated or unsupported metric claims.

### DIFF
--- a/verify-cv-facts.mjs
+++ b/verify-cv-facts.mjs
@@ -56,8 +56,18 @@ const METRIC_NOUNS = [
 // for the chain — modifiers are alphabetic only — so a wider window still
 // cannot jump across an intervening figure to bind an unrelated noun.
 const MODIFIER_WINDOW = 4;
+// The number capture takes an immediately-adjacent magnitude suffix (50k, 1.5M)
+// as part of the number, mirroring what the currency pattern below already does.
+// Without it the modifier window re-consumed that letter as a generic word, so
+// "50k users" normalized to the claim "50 users" and matched a CV that said 50 —
+// letting a 1000x inflation through the gate while a smaller "900 users" was
+// correctly caught.
+//
+// `[kKmMbB]\b` requires the suffix to END the token, so "50 million users" (space,
+// handled by the modifier window) and "50kg users" (k not at a boundary) both keep
+// their existing behaviour and still normalize to "50".
 const COUNT_CLAIM_RE = new RegExp(
-  String.raw`\b(\d[\d,.]*)\s*\+?\s*(?:[A-Za-z][A-Za-z-]*\s+){0,${MODIFIER_WINDOW}}(${METRIC_NOUNS.join('|')})\b`,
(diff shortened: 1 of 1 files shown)


santifer:

COUNT_CLAIM_RE captures the digits but not the magnitude suffix glued to them, and the modifier window swallows that letter as a generic word: "50k users" normalizes to "50 users" and matches a CV that says 50. So the anti-fabrication gate caught 900 vs 50 while letting a 1000x inflation through, and it fails in the unsafe direction (reports clean). The fix aligns the count pattern with what CURRENCY already does in the same file, deliberately without the currency's \s?, because in counts that would capture the "50 m" inside "50 million"; 5 new cases in --self-test (39 -> 44), and you checked that 2 fail when the regex is reverted.

santifer:
Merged, @darkpandawarrior: "50k users" silently becoming "50 users" was exactly the kind of inflation the gate exists to catch, and mirroring the CURRENCY pattern keeps the regex family coherent. Six new self-test cases seal it. 🚀

PR #783, opened 2026-06-04
fix(liveness): stop false-expiring anti-bot-walled postings (pracuj.pl)

Szabranito:
## Problem

Headless Playwright trips Cloudflare's "Just a moment..." challenge on portals like **pracuj.pl**: the offer URL returns **HTTP 403** with a ~260-char challenge body. `classifyLiveness` read that short body as `insufficient_content` → **`expired`**.

Because `scan --verify` writes `expired` URLs to `scan-history.tsv` (permanently dedup-filtering them on later scans), a *transient* bot wall silently and **irreversibly drops live jobs** from the pipeline. A full 114-URL liveness pass mass-flagged **110 live pracuj.pl postings as expired**.

## Fix (layered, cheapest-first)

- **`classifyLiveness`** — detect anti-bot challenge bodies and `403`/`503` as `uncertain` (`bot_challenge` / `access_blocked`), **never `expired`**. Genuinely removed postings still return `404`/`410` or a hard-expired banner (unchanged). Ambiguity must stay `uncertain` so the URL is preserved and retried.
- **`newLivenessPage`** — open the headless context with a realistic Chrome UA. Playwright's default UA contains `HeadlessChrome`, which Cloudflare flags; a normal UA clears the wall for the common case (same trick `scripts/parsers/pracuj-jobs.mjs` already relies on).
- **`checkUrlLivenessWithFallback` + `createHeadedPageProvider`** — on a *persistent* challenge, retry once in a lazily-launched headed browser; degrades gracefully to the headless result when no display is available (never to `expired`).
- **`APPLY_PATTERNS`** — add Polish (`aplikuj` / `panelu aplikowania` / `wyślij CV`) so a loaded pracuj.pl page resolves to `active`, not `no_apply_control`.
- **`--throttle[=ms]`** (jittered) on `check-liveness` and `scan --verify`, for rate-limited portals. Off by default.

**Wiring:** `check-liveness` uses the headed fallback by default (`--no-fallback` to disable); `scan --verify` keeps it opt-in via `--headed-fallback` (needs a display, so unattended scans shouldn't depend on it).

## Tests

9 new liveness cases in `test-all.mjs` (challenge detection, 403 handling, Polish apply, the fallback wrapper's recover/no-provider/persistent/no-display paths, and the throttle helper). Suite: **159 passed** (was 150), with the one pre-existing `Dashboard build` failure unaffected.

## Known limitation

[...]

PR #2613, opened 2026-08-08
fix(liveness): a rate-limited posting was classified expired, not uncertain

darkpandawarrior:
## The bug

`classifyLiveness` special-cases 403 and 503 as `access_blocked`, and any other 5xx as a transient `server_error`. **429 matched neither.** A rate-limit response carries a short "Too Many Requests" body — well under `MIN_CONTENT_CHARS` — so it fell through to the content heuristic and came out as `expired`:

```
status 429 -> {"result":"expired",   "code":"insufficient_content"}   ← the bug
status 403 -> {"result":"uncertain", "code":"access_blocked"}
status 503 -> {"result":"uncertain", "code":"access_blocked"}
status 500 -> {"result":"uncertain", "code":"server_error"}
```

429 means the board is throttling *us*. It is never evidence the posting is gone.

## Why it matters beyond one verdict

`scan.mjs --verify` records an expired result in scan-history as `skipped_expired`, and later scans dedup-skip that URL (`scan.mjs:2390`).

I checked how long that lasts rather than assuming: `skipped_expired` is **not** in `PERMANENT_SCAN_HISTORY_STATUSES`, but `shouldDedupScanHistoryRow` returns `true` indefinitely while `scan_history.recheck_after_days` is unset — which is the default. So a live job that happened to be throttled once is dropped from every future scan until the user configures a recheck window.

It also compounds in the worst direction: **scanning harder is exactly what earns a 429**, so a large sweep quietly discards the most jobs.

(To be precise about the blast radius: this does *not* poison the role-level key. `loadSeenUrls`' documented semantics only seed a role key from `added` rows, so city variants still self-heal. The damage is URL-level.)

## Fix

One status added to the existing guard:

```diff
-if (status === 403 || status === 503) {
+if (status === 403 || status === 429 || status === 503) {
```

This file's history is a run of the same class — `#2371` added 5xx, `#783` added 403 anti-bot walls, `#2163`/`#2212` added phrasing cases. 429 was the remaining gap in that set.

## Test

Mirrors the existing 503 case in `tests/liveness-core.test.mjs`, asserting **both halves** of the verdict (the file's own comment notes the 5xx guard sits below, so checking only the code would be weaker). Verified it fails with the guard reverted:

```
❌ HTTP 429 classified expired/insufficient_content, expected uncertain/access_blocked
```

`node test-all.mjs` → **3111 passed, 0 failed**.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of rate-limited responses so they are recognized as access restrictions instead of incorrectly marking content as expired.
  * Preserved accurate status classification when a service returns limited or incomplete response content.

* **Tests**
  * Added regression coverage for rate-limited responses with short response bodies.

### DIFF
--- a/liveness-core.mjs
+++ b/liveness-core.mjs
@@ -129,7 +129,14 @@ export function classifyLiveness({ status = 0, requestedUrl = '', finalUrl = '',
   if (botChallenge) {
     return { result: 'uncertain', code: 'bot_challenge', reason: `anti-bot challenge: ${botChallenge.source}` };
   }
-  if (status === 403 || status === 503) {
+  // 429 belongs with 403/503: rate limiting is the board throttling US, never
+  // evidence the posting is gone. Its body is a short "Too Many Requests", well
+  // under MIN_CONTENT_CHARS, so without this it fell through to
+  // insufficient_content and read as `expired` — and an expired result is
+  // written to scan-history as skipped_expired, whose URL every later scan
+  // dedup-skips (indefinitely, unless scan_history.recheck_after_days is set).
+  // Scanning harder is exactly what earns a 429, so this compounds.
+  if (status === 403 || status === 429 || status === 503) {
     return { result: 'uncertain', code: 'access_blocked', reason: `HTTP ${status} (access blocked, likely anti-bot)` };
   }
   // Any other 5xx is a transient origin error (502/504 gateway hiccups, 500s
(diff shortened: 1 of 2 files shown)


santifer:

429 falls through both branches of classifyLiveness (403/503 to access_blocked, other 5xx to server_error), so the short "Too Many Requests" body hits the content heuristic and comes out as expired. You follow it downstream: that expired gets written to scan-history as skipped_expired, which isn't in PERMANENT_SCAN_HISTORY_STATUSES but which shouldDedupScanHistoryRow dedupes indefinitely while recheck_after_days is unset (the default), so a live listing throttled once disappears from every later scan; and scanning harder is exactly what triggers the 429, so the big sweep is the one that drops the most listings. One-line diff plus a regression test verified by reverting the guard.

santifer:
Merged, @darkpandawarrior. A 429 with a short body falling through to the content heuristic and coming out expired was a misclassification with teeth: marking a live posting dead. Baseline repro plus regression test made this an easy yes.

PR #2614, opened 2026-08-08
fix(agent-inbox): concurrent adds silently dropped queued requests

darkpandawarrior:
## The bug

`add` reads the whole queue, appends a line in memory, and writes the whole file back. Every request that lands between the read and the write is overwritten and lost — silently, with a cheerful `Queued: ...` printed for each one.

```
30 concurrent `node agent-inbox.mjs add` calls  ->  3 items survived
```

(The loss rate varies with timing; I saw 15/30 and 3/30 on different runs. Both are catastrophic.)

## This is the designed usage, not an exotic case

Straight from the file's own docblock:

> an **append-only queue** the agent drains at the start of a session
> … any tool (a dashboard, a script, cron) **can append to it**

Several writers with no session running is the *point* of the feature — it exists precisely so you can drop a request in when you're not in a session. `agent-inbox-tests.mjs` also already asserted "`add` is append-only". Nothing was actually appending.

## Fix

Append the line instead of rewriting the file. POSIX guarantees an `O_APPEND` write is atomic below `PIPE_BUF`, and one checklist line is far under it, so concurrent adds interleave rather than clobber.

The file is still read first, but only to decide whether a separating newline is needed for a queue someone hand-edited without a trailing one. The write itself is a single atomic append.

## What this does NOT fix

`resolve` is still a read-modify-write over the whole file, so an `add` landing mid-`resolve` can still be lost.

I left it deliberately: that path is interactive and one-at-a-time, so the window is far narrower than automated adds, and closing it properly means introducing a lock rather than a one-line change — a different PR with a different risk profile. Say the word and I'll follow up with it (`pipeline-lock.mjs` already generalizes over a path, so it may be reusable as-is).

## Test

30 genuinely concurrent adds must all survive, none duplicated or truncated.

It uses `spawn()`, **not** `spawnSync()` — I wrote it with `spawnSync` first and caught that a synchronous loop serializes the adds and passes against the buggy rewrite, proving nothing. There's a comment in the test saying so, because it's an easy trap to reintroduce.

Verified it fails with the rewrite restored:

```
❌ all 30 concurrently queued items survive — kept=3 of 30
❌ no item is duplicated or truncated — distinct=3 of 30
```

`node test-all.mjs` → **3110 passed, 0 failed**.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when multiple items are added to the inbox simultaneously.
  * Prevented existing queued items from being overwritten during concurrent additions.
  * Ensured each new item remains complete, distinct, and correctly separated.

* **Tests**
  * Added coverage for 30 simultaneous inbox additions.
  * Verified all additions complete successfully and remain queued without duplication or data loss.

### DIFF
--- a/agent-inbox-tests.mjs
+++ b/agent-inbox-tests.mjs
@@ -149,10 +149,13 @@ console.log('7. concurrent adds do not lose items (append, not rewrite)');
   const failedSpawn = exits.filter((c) => c !== 0).length;
   check('every concurrent add exited cleanly', failedSpawn === 0, `${failedSpawn} non-zero exits`);
   const body = readFileSync(inbox, 'utf8');
-  const kept = body.split('\n').filter((l) => l.startsWith('- [ ]')).length;
+  const pending = body.split('\n').filter((l) => l.startsWith('- [ ]'));
+  const kept = pending.length;
   check(`all ${N} concurrently queued items survive`, kept === N, `kept=${kept} of ${N}`);
-  const distinct = new Set(body.split('\n').filter((l) => l.startsWith('- [ ]')).map((l) => l.slice(l.indexOf('— ')))).size;
-  check('no item is duplicated or truncated', distinct === N, `distinct=${distinct} of ${N}`);
+  const actual = new Set(pending.map((l) => l.slice(l.indexOf('— ') + 2)));
+  const expected = new Set(Array.from({ length: N }, (_, i) => `item-${i}`));
+  const complete = actual.size === expected.size && [...expected].every((item) => actual.has(item));
+  check('no item is duplicated or truncated', complete, `actual=${[...actual].join(', ')}`);
 }
 
 console.log(`\nResults: ${passed} passed, ${failed} failed`);
--- a/agent-inbox-tests.mjs
+++ b/agent-inbox-tests.mjs
@@ -12,12 +12,14 @@
  *   5. An empty `add` fails loudly (exit 1) rather than queuing a blank line.
  *   6. On the default path, a first `add` self-heals .gitignore (idempotent) so
  *      the personal queue isn't accidentally tracked.
+ *   7. Concurrent `add` calls all survive — the queue is appended to, never
+ *      rewritten, so simultaneous writers cannot clobber each other.
  *
  * Provisions a throwaway queue via CAREER_OPS_INBOX and a temp CWD; never
  * touches real user data.
  */
 
-import { execFileSync } from 'child_process';
+import { execFileSync, spawn } from 'child_process';
 import { readFileSync, writeFileSync, mkdtempSync } from 'fs';
 import { join, dirname } from 'path';
 import { tmpdir } from 'os';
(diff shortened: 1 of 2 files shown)


darkpandawarrior:
Two of the three review points are addressed on the branch; the third I want to push back on.

**Test assertion too loose** — fixed. It now compares the exact expected item set rather than just counting lines, so a run that kept 30 items but mangled or duplicated one no longer passes.

**`ensureFile` init is not atomic** — you're right, and it's a real narrow race: two processes can both see the file missing and both `writeFileSync(HEADER)`, so one's header write can land after the other's first append. Worth noting the blast radius is much smaller than the one this PR fixes — it only affects the very first `add` on a brand-new queue, and the worst case is a duplicated header block rather than a lost request. Happy to close it properly with an `existsSync`-then-`{ flag: 'wx' }` write (fails if it appeared meanwhile) — say the word and I'll add it here, or send it separately if you'd rather keep this PR to the one defect.

**Separator check reads the whole file** — I'd rather leave this. Reading the tail would avoid loading the file, but an inbox is a handful of checklist lines, so the read is negligible, and `readFileSync` is what every other function in this script already does. Swapping in a partial-read + fd juggle here would make this the only place doing something clever for no measurable gain. The atomicity that matters is in the `appendFileSync` call, which the read does not affect either way.

Full `test-all`: 3110 passed, 0 failed.

darkpandawarrior:
Correcting my previous comment — it no longer describes the branch, and one thing in it was wrong on the merits.

**Both changes I said I'd hold are already pushed** (`600210e`). My comment went up while that work was in flight and I did not reconcile the two. Sorry for the noise — read the diff, not that comment.

**I also understated the `ensureFile` race.** I called the worst case "a duplicated header block rather than a lost request". That's wrong. The loser's `writeFileSync` uses the default `w` flag, which **truncates** — so if it lands after the winner has already appended an item, that item is wiped. It's the same data-loss class as the bug this PR is about, just narrowed to a brand-new queue. It's closed with an atomic `wx` create; the loser gets `EEXIST` and does nothing.

**And I was wrong to wave off the tail read.** I argued a full read was fine because an inbox is small. Now that `ensureFile` is exclusive-create, `add` is a genuine append-only path, and re-reading the entire file just to inspect one byte is the odd thing out. It reads the last byte through a file descriptor instead. Behaviour is unchanged; it just stops doing work it never needed.

So all three of the original findings are addressed. Net effect on the guarantee: `add` is now safe from first-write truncation as well as from concurrent-append loss.

`resolve` is still a read-modify-write over the whole file — genuinely out of scope here, and I'd rather send it separately with a lock than widen this PR.

Tests: 19 passed, 0 failed (30 concurrent adds, exact item-set comparison). Full `test-all`: 3110 passed, 0 failed.

santifer:

agent-inbox.mjs add read the whole queue, appended in memory and rewrote the file, so every request landing between the read and the write vanished in silence while it printed a cheerful 'Queued: …': 30 concurrent adds left 3 survivors (15/30 on another run). You argue it isn't an exotic case straight from the file's own docblock, an 'append-only queue' that 'any tool (a dashboard, a script, cron) can add to', so multiple writers with no open session is the feature, and agent-inbox-tests.mjs already asserted add was append-only while nothing actually appended. The craft detail: you wrote the test with spawnSync first, caught that the synchronous loop serializes the adds and passes against the buggy code, and left a comment in the test explaining the trap so nobody reintroduces it.

santifer:
Merged, @darkpandawarrior: the file's own docblock promised an append-only queue and the read-rewrite cycle was quietly breaking that promise under concurrency. Now the behavior matches the contract.

santifer:
Reopening this in the comments rather than quietly, @darkpandawarrior, because your test caught something and I merged before it could tell me.

**Windows still loses an item.** The suite you added is red on `test (windows-latest)`, consistently across three separate runs since the merge:

```
```

`item-5` is missing. So this is not a flaky test: it is **the exact bug this PR set out to fix, still alive on Windows**, and your test is what surfaced it. On Linux and macOS the fix holds; `appendFileSync` gets atomicity there for small writes that Windows does not guarantee under concurrency.

Two things I want to say plainly:

1. **The merge was my call and the gap is mine.** CI was green on the PR when I merged, which now reads as timing luck rather than evidence: a race that loses one write in thirty will pass often enough to look green once.
2. **Your diagnosis was right and your test is doing its job.** Catching a real platform-specific race is more than the PR promised.

I am opening a P0 issue for the Windows path and proposing the same fix that landed for scan-history in #2639 today: route the append through the existing `withPipelineLock` in `pipeline-lock.mjs`, which is already the repo's answer for exactly this class. If you want to take it, say so and it is yours; if you would rather not, I will do it and credit the diagnosis to you either way.

Meanwhile every PR running Windows CI is red because of this, so it moves tonight.

PR #2617, opened 2026-08-08
fix(doctor): make doctor CLI-resolution tests hermetic against .env leak

guillaume-flambard:
**Superflow sweep finding** (2026-08-08): 8 doctor tests were failing on any checkout with a real `.env`.

**Root cause:** `openrouter-runner.mjs` loads the repo `.env` at module scope. Importing it (test-all.mjs imports `buildSystemPrompt`) therefore mutates `process.env` in the importing process. Every key leaks, including `CAREER_OPS_CLI`, which silently overrode the default-CLI scenarios: they assert a default of `claude` and got whatever the developer had in `.env`.

**Fix:**
- `openrouter-runner.mjs`: move the loader into `loadEnvFile()` and call it only when `invokedDirectly`, so importing the module no longer mutates the caller's env.
- `tests/doctor-cli-resolution.test.mjs`: drop `CAREER_OPS_CLI` from the base env so each scenario controls the CLI explicitly. Assertions unchanged, nothing disabled.

**Rebased and reduced on 2026-09-03.** `main` has since landed the `doctorEnv` change in `test-all.mjs` and the `CLAUDE_CONFIG_DIR` override in `tests/playwright-mcp-detection.test.mjs`, both in a better shape than this branch had them, so those two hunks are dropped rather than rebased. What remains is the part `main` does not have, and it addresses the cause rather than the symptom: as long as the loader runs at module scope, any future in-process import can leak `.env` into a test again.

The `js-yaml` bump and a stray run-log file that were in the original branch are also dropped: the bump landed separately, and the run log did not belong here.

**Result:** `node test-all.mjs` = **8094 passed, 0 failed** on this branch.

### DIFF
--- a/.superflow/run-log.md
+++ b/.superflow/run-log.md
@@ -1,0 +1,43 @@
+# Run log — career-ops (superflow fix run, 2026-08-08)
+
+| check | result |
+|---|---|
+| stack | node · CLI skill agent (.mjs, Playwright) — no web app |
+| audit | **0 high** (js-yaml fixed to 4.3.1 via `npm audit fix`; was CVE-2026-59870 quadratic CPU on `!!omap`) |
+| tests | **2717/2717 pass** (`node test-all.mjs`, exit 0) |
+| lint/typecheck | n/a (JS-only; `test-all.mjs` §1 runs `node --check` on every .mjs — the repo's lint gate) |
+| verdict | green |
+
+## The 8 failing "doctor" tests — root cause
+
+The previous sweep's diagnosis ("checkout not onboarded + Playwright MCP absent") was **wrong**.
+The failures were **not** about this checkout lacking `cv.md`/`config/profile.yml`/`portals.yml`
+(those are WARN-only in `doctor.mjs`, never failures). Real cause:
+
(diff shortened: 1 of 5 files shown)


santifer:
Thanks for this, @guillaume-flambard.

You diagnose cross-test contamination, not a flaky test: earlier in-process imports (openrouter-runner.mjs, scan.mjs) load the repo .env at module level, that pins CAREER_OPS_CLI=opencode, and it took down 8 doctor tests on any checkout with a real .env (playwright-mcp-detection 1, 2, 6, 7, 12, 14 plus test-all §12d). The fix goes at the cause instead of the symptom: loadEnvFile is deferred to the CLI entry point and CAREER_OPS_CLI is deleted from the child environment, with no assertion disabled or loosened. Result 2717/2717.

santifer:
Sorry for the slow reply, @guillaume-flambard: a long backlog stretch around the move to the org. Thanks for your patience. Now, to your PR: moving the .env loader behind invokedDirectly is the right fix (importing a module should never mutate the caller's env), and the hermetic env in the doctor test closes the same hole from the other side. Merged: your first contribution. Welcome to career-ops!

PR #1555, opened 2026-07-04
feat(providers): TKMS, Heckler & Koch, and Deutsche Bahn providers

tillmannschatz:
## What

Three new zero-token scan sources for German industrial/defense employers.

### `providers/tkms.mjs` — thyssenkrupp Marine Systems

`jobs.tkmsgroup.com` exposes a public JSON API:
- `POST /api/filter/query` with `{searchQuery, filter, subclient, locale, page}` — 20/page, 0-based, `nextPage: null` on the last page. `subclient`/`locale` are configurable via a `tkms:` block (the backend is a shared platform).
- Public job URL `{origin}/{locale}/job/{slug}/{id}`; the slug is cosmetic (stub resolves, verified), so the provider slugifies the title. `postingDate_timestamp`/`postingDate` → `postedAt`; `locations[].cityState` joined.
- **Verified live**: ~330 postings across TKMS GmbH / ATLAS ELEKTRONIK / Wismar. Strong PLM/Teamcenter overlap in naval engineering.

### `providers/hecklerkoch.mjs` — Heckler & Koch (single-company, SSR)

`heckler-koch.com/de/Karriere/Stellenangebote` is server-rendered — one GET yields every posting. The parser anchors on the `karriere.heckler-koch.com/jobposting/{hash}` link (stable id + job URL) and reads the sibling `<h3>` title. Small board (~32 roles), single-site (Oberndorf/Vöhringen) so location is left empty rather than guessed. **Verified live**: 32 jobs.

### `providers/deutschebahn.mjs` — Deutsche Bahn (single-company)

DB's branded Avature front (`jobs.deutschebahngroup.careers`) just 302-redirects into the custom **db.jobs** portal. The provider parses the server-rendered `/service/search/de-de/{searchId}` fragment (`m-search-hit` anchors: title span, `Arbeitsort` `<li>`, `data-job-id`), paginating `?pageNum=N`. `searchId` pinned via `api:`.

> **CAVEAT** (documented in the portals note): db.jobs is Akamai bot-protected and serves a challenge page to datacenter IPs — the same class as BMW in this repo. The `parseHits` parser is **unit-tested against the real `m-search-hit` markup** and works from residential IPs / the maintainer's environment, but a datacenter scan may return 0. The provider degrades gracefully (returns `[]`).

[...]

PR #2623, opened 2026-08-08
fix(jobvite): the JSON API is retired — switch to the XML feed

LayerC0de:
## The provider returns zero jobs for every tenant

`providers/jobvite.mjs` fetches:

```
GET https://jobs.jobvite.com/api/company/{slug}/jobs
```

That endpoint is gone. It answers **302 → `http://search.jobvite.com?invalid=1`**, so `fetchJson` receives an HTML marketing page and the provider yields nothing — silently, because an empty board is indistinguishable from a company with no openings.

**The tenant that proves it is `tylertech`** — still a live Jobvite customer (`companyEId` `q6NaVfwI`, 236 jobs on the XML feed) and dead on the JSON API. That is the case this PR restores.

> **Correction (2026-08-11).** This section previously cited six tenants — `zoom`, `starbucks`, `servicenow`, `twilio`, `blueorigin`, `tylertech` — as evidence. That overstated the verification, as @ilyakanevskiy caught in review. The other five have migrated off Jobvite entirely; their slugs answer identically on *every* Jobvite endpoint, XML feed included. A retired tenant and a retired API look alike from the outside, so those five are consistent with the claim without being evidence for it. Only a tenant that is live on one endpoint and dead on the other separates the two, and `tylertech` is that tenant.

**The existing tests passed the whole time.** They only exercised `parseJobviteResponse` against a hand-written fixture; nothing asserted that the URL the provider *builds* still answers. That gap is as much the bug as the endpoint is, and the new tests close it.

## The working feed

```
GET https://app.jobvite.com/CompanyJobs/Xml.aspx?c={companyEId}

<result><job>
  <id/><title/><category/><location/><date>M/D/YYYY</date>
  <detail-url><![CDATA[…]]></detail-url>
  <apply-url><![CDATA[…]]></apply-url>
</job>…</result>
```

## The tenant key changed shape

The old API keyed on the vanity slug (`tylertech`). The feed keys on an opaque **`companyEId`** (`q6NaVfwI`) that does not appear in the careers URL at all — it lives in the board page's inline JS.

A slug alone is no longer sufficient, so resolution is tiered:

| | Source | Cost |
|---|---|---|
| 1 | `company_eid:` on the entry | no network |
| 2 | `?c=` of an explicit `api:` URL | no network |
| 3 | scrape the board page | one extra request |

(3) keeps existing slug-only entries working with **no config migration required**; (1) is documented as preferred since it skips a request per scan.

`company_eid:` is a new config key, now documented in `docs/SUPPORTED_JOB_BOARDS.md` and `templates/portals.example.yml` alongside a Jobvite stanza — both of which had no Jobvite entry at all.

## Rate limiting

`app.jobvite.com` answers **`429 Retry-After: 30` from the second request onward**, so a scan covering two Jobvite tenants back-to-back already trips it. This is the normal path, not a mass-scan edge case.

Neither `ctx.fetchText` nor this provider retried. `fetchJsonWithRetry` in `_http.mjs` already had the mature semantics — exponential backoff, jitter, a clamped `Retry-After` — but was JSON-only.

Nothing in that retry loop ever inspected the response body, so it is now parameterised by the request: `withRetry()` holds the policy and `fetchJsonWithRetry` / `fetchTextWithRetry` are two thin callers. Duplicating it per content type is how the entity decoders drifted (#1555, #1639). `fetchJsonWithRetry` keeps its exact signature and behaviour.

`maxDelayMs` is raised to 15s for this provider purely to widen the `Retry-After` clamp: the shared 8s default puts the ceiling at 32s, uncomfortably close to Jobvite's advertised 30 for a value we do not control.

Verified live — 200 (945 KB, 40 jobs), immediate 429, then 200 again once the 30s was honoured. The wait genuinely clears it.

## Branded custom domains

Many Jobvite customers point their public careers page at their own host, and a bare `jobs.jobvite.com/{slug}` **302s there before the board HTML carrying `companyEId` is ever served** — so discovery could not start for those tenants at all.

The branded page's own embedded iframe requests `?fr=true&nl=1`, which renders the listing inline instead of redirecting. Verified: `fieldcore`, `imprivata` and `opentrons` all 302 away bare and answer 200 with an eId once the params are present; `egnyte` returns the same eId either way, so it is a no-op where it is not needed.

Applied to the **board request only, not the feed.** The redirect is a property of the branded board, not of the tenant — the same three tenants' feeds answer 200 bare (`fieldcore` 126 jobs, `imprivata` 53, `opentrons` 6). Adding undeclared params to the XML endpoint would be cargo-culting a fix onto a request that never had the problem.

## An empty board is not an error

A tenant with no open roles does not get `<result/>`. The feed **302s to `NoJobs.htm`**, and `redirect:'error'` turned zero vacancies into a thrown error and a tenant reported as broken. Observed on `leovegas`, whose board also says so in prose.

The feed now uses `redirect:'manual'`, and `_http.mjs` attaches `err.location` so a caller can tell *which* redirect it hit without gaining the ability to follow it.

Scoped deliberately tight — a 3xx, on the feed host, targeting that one page:

- the board's own redirect to `search.jobvite.com?invalid=1` means the **opposite** thing (a slug that is no longer a Jobvite tenant) and keeps failing loudly
- a `NoJobs.htm` served from any other host is not trusted

Laundering either into "0 jobs today" would hide a dead portal entry behind a plausible empty result — the exact silent failure this PR exists to remove.

The header is relative in practice (literally `NoJobs.htm`), so it is resolved against the request URL rather than string-compared.

## Timeout

The feed inlines every job's full HTML description, so it is large by construction — Tyler Technologies is **1.88 MB / 236 jobs in ~11s**, against the 10s default in `_http.mjs`. That one-second overshoot aborted the entire tenant and surfaced as a network error. The feed fetch now uses 45s; the board page keeps the default.

## SSRF

Two hosts are in play, so `assertJobviteHost()` pins both, and **no redirect is ever followed.** The board uses `redirect:'error'`; the feed uses `redirect:'manual'`, which is the same guarantee by a different route — undici hands back the 3xx as a response instead of chasing it, so the final hostname still cannot move. `'manual'` buys the ability to *read* the `Location` header without acting on it, which is what separates an empty board from a retired tenant.

The eId is only ever a query-param value, never a path segment. Per-job URLs remain display-only and accepted from any `https:` origin, unchanged.

One deliberate relaxation: feed URLs arrive as `http:` in practice and are **upgraded to `https:` rather than dropped** — dropping would discard every posting in the feed. They are never fetched.

## Tests

Rewritten and extended, **56 assertions**. Beyond parsing (CDATA, entity decoding, `detail-url` preferred over `apply-url`, posting-drop rules, ReDoS regression fixtures), they assert **which URLs get requested and how**:

- one request when the eId is configured
- two on the discovery path, board before feed, board carrying `fr=true&nl=1`
- **zero** when the only id source is a foreign host
- `redirect:'error'` on the board and `redirect:'manual'` on the feed, pinned explicitly — three of the behaviours above ride on those two values, and a suite that only asserts *that* a request happened lets any of them regress silently
- a 429 is retried, the advertised `Retry-After` is honoured exactly, and a non-retryable 4xx is not retried
- a `NoJobs.htm` redirect reads as an empty board; a `search.jobvite.com?invalid=1` redirect still throws; a `NoJobs.htm` from a foreign host still throws

## Verification

Full suite: **3169 passed, 0 failed**.
Live end-to-end against Tyler Technologies: **236 jobs parsed**.
Live probes for each review finding are quoted inline above, re-run 2026-08-11.

## Credit

The review on this PR from @ilyakanevskiy was substantive enough to name rather than thank in passing: the `?fr=true&nl=1` fix for branded custom domains, the `NoJobs.htm` empty-board diagnosis *including* the warning to scope it narrowly so a retired tenant keeps failing loudly, and 35 hand-tested tenants covering the layout variants he had already mapped for #2529 — which he then closed in favour of this PR. Three of the six findings would not have been in this branch without that.

He explicitly said a trailer was not something he needed. Adding one anyway, because the work was real — `Reviewed-by:` rather than `Co-authored-by:`, since the accurate claim is the stronger one here.

Reviewed-by: Ilya Kanevskiy <12554484+ilyakanevskiy@users.noreply.github.com>

## Summary by CodeRabbit

* **New Features**
  * Added Jobvite support for automatic board detection and explicit XML feed configuration.
  * Added support for large descriptions, extended feed timeouts, and empty boards.

* **Bug Fixes**
  * Updated listings to use Jobvite’s current XML feed.
  * Improved board discovery, listing normalization, and link security.
  * Excluded invalid postings and strengthened redirect, retry, and rate-limit handling.


LayerC0de:
Pushed c5e1de8. Two of the three findings were valid; one is not, and the CodeQL failure turned out to be the most important of the lot.

**CodeQL high — ReDoS. The actual CI failure, and it was mine.**

`tagText()` built `<name>\s*(?:<!\[CDATA\[)?([\s\S]*?)(?:\]\]>)?\s*</name>` — `\s*` either side of a lazy `[\s\S]*?`. On input that never closes the tag that is polynomial backtracking, and this parser runs over a remote 1.9 MB document, so the input is both untrusted and large. CodeQL was right to flag it.

Replaced with `indexOf` scanning. Linear, allocation-light, and frankly easier to read than the regex. A 60k-space unterminated tag now returns in milliseconds — that case is a test.

**🟠 apply-url was not actually a fallback. Valid, and a real bug.**

`tagText(detail-url) || tagText(apply-url)` picks a *string* and validates once, so a present-but-malformed `detail-url` discarded the posting even with a good `apply-url` beside it. Each candidate is now validated in turn, first usable `https:` wins. I took your suggestion on the fixture too — the malformed-detail case now carries a valid `apply-url`, so the path is exercised rather than merely asserted.

**🟡 Numeric character references. Valid.**

Not hypothetical: Jobvite carries typographic punctuation through from the source ATS, so `&#8217;` and `&#x2013;` appear in live titles and a raw `&#8217;` would reach the tracker and any generated document. Decimal and hex now decode; out-of-range code points are left as-is rather than emitting U+FFFD.

**🔴 "Duplicate `const ctx` declarations… JavaScript rejects the module with a SyntaxError, so none of these tests execute" — this one is incorrect.**

The four declarations at L149/161/175/192 are each inside their own `{ }` block:

```js
{
  let reached = null;
  const ctx = { … };
  …
}
```

Block-scoped `const` in sibling blocks is legal. Evidence: `node --check tests/providers/jobvite.test.mjs` passes, and the suite runs — the assertion count went **40 → 43** with the new cases, which it could not do if the module failed to parse. No change made.

Live re-verified against Tyler Technologies: **236 jobs, zero raw numeric references** in any title.

ilyakanevskiy:
Hi @LayerC0de,

I shipped my own Jobvite fix a few days earlier (#2529), but mine depends on scraping HTML markup, which means juggling several different layout variants tenant-by-tenant, plus pagination (Jobvite's HTML search results cap at 50/page). Yours is a much more robust approach — a single, uniformly-structured XML feed beats parsing five different HTML shapes any day, and it doesn't need pagination at all: I confirmed tenants with 300+ postings (e.g. nutanix, 333 jobs) come back complete in one request.

While testing it against tenant quirks I'd already mapped out for #2529, I ran into a few issues worth flagging before this merges:

1. **No retry/backoff on 429.** `app.jobvite.com` rate-limits hard — `Retry-After: 30` on basically every request after the first, consistently, across dozens of tenants I tested. This isn't a mass-scan edge case: a normal setup scanning just a handful of Jobvite tenants back-to-back will already hit it, since the limit kicks in on the second request. Neither `fetch()` here nor the shared `ctx.fetchText` in `_http.mjs` retries or honors `Retry-After` (`fetchJsonWithRetry` exists but is JSON-only and isn't called from this provider). Worth adding a retry path for the board + feed fetches, honoring `Retry-After` the way the JSON retry helper already does.

2. **Branded custom-domain redirects aren't handled**, so `companyEId` discovery never gets off the ground for those tenants. A number of Jobvite tenants point their public careers page at their own domain; `jobs.jobvite.com/{slug}` 302s away before the board HTML (and the `companyEId` inside it) is ever seen, and `redirect:'error'` turns that into a hard failure. I ran into the same thing in #2529 and found a working bypass: the branded page's own embedded iframe points at `jobs.jobvite.com/{slug}?fr=true&nl=1` — appending those two params to every board/feed request renders the listing directly instead of redirecting. Confirmed it's a no-op for tenants that don't need it, and it recovers all six branded-redirect tenants below (real job counts, not just "unblocked").

3. **A confirmed-empty board can throw instead of returning `[]`.** For at least one live tenant (`leovegas`), the careers page explicitly says "There are currently no open jobs.", and the XML feed agrees — but it agrees by 302-redirecting to `NoJobs.htm` rather than returning `<result />`. With `redirect:'error'` on the feed fetch, that becomes a thrown error, not an empty array. Worth treating that as "genuinely empty" rather than a failure — 0 vacancies is 0 vacancies, not an error state. One thing to watch: this should be scoped narrowly to the *feed* redirect (and ideally to a target matching `NoJobs.htm`), not "any redirect during fetch." The *board* redirect in issue 4 below (`jobs.jobvite.com/{slug}` → `search.jobvite.com?invalid=1`) means something different — an invalid/retired tenant — and needs to keep failing loudly rather than getting swept into the same "treat redirect as empty" bucket.

4. **Minor, not a code issue:** of the six tenants cited in the PR description as proof the old JSON API is dead (zoom, starbucks, servicenow, twilio, blueorigin, tylertech), only `tylertech` is actually still live on Jobvite. I checked the other five — they're confirmed migrated to other ATS platforms now, and their `jobs.jobvite.com` slugs 302 to `search.jobvite.com?invalid=1` regardless of which Jobvite endpoint you hit. Doesn't affect the fix itself, just means that verification list overstates how many of those six this PR actually restores.

5. **Two things worth tightening up too:**
   - `decodeEntities()` here is a local copy rather than importing `providers/_html-entities.mjs`, which every other HTML/XML-scraping provider in the repo shares. That decoder used to be duplicated across a few providers and drifted out of sync between copies (#1555, #1639) — worth avoiding a repeat. Your version is actually stricter than the shared one (it rejects NUL/C0 controls per XML 1.0 §2.2, the shared one doesn't) — probably best as a separate follow-up PR upstreaming that improvement into `_html-entities.mjs`, rather than folding it into this one.
   - The test suite never asserts `opts.redirect === 'error'` on the `ctx.fetchText` calls, only that they're called. Given that three of the issues above are all consequences of `redirect:'error'` semantics, a test pinning that behavior down seems worth adding regardless of how the other issues get resolved.

6. **Docs are out of sync.** `docs/SUPPORTED_JOB_BOARDS.md` and `templates/portals.example.yml` don't have a Jobvite entry yet — and the config shape changed (`company_eid:` is new), so wherever this lands, it'll need documenting.

Tenants I tested against the real endpoints (board + XML feed), covering the layout variants and quirks from #2529:

| Tenant | Result | Notes |
|---|---|---|
| jacksonfamilywines | 47 jobs | works as-is |
| egnyte | 44 jobs | works as-is |
| telenav | 3 jobs | works as-is |
| webmd | 173 jobs | works as-is |
| nutanix | 333 jobs | works as-is |
| dotdash | 21 jobs | works as-is |
| xperi | 26 jobs | works as-is |
| arc | 57 jobs | works as-is |
| lhhcareers | 62 jobs | works as-is |
| innio | 240 jobs | works as-is |
| vaniamgroup | 15 jobs | works as-is |
| sitecore | 51 jobs | works as-is |
| samtec | 201 jobs | works as-is |
| loandepot | 96 jobs | works as-is |
| carfax | 45 jobs | works as-is |
| lifenethealth | 51 jobs | works as-is |
| nitscheng | 10 jobs | works as-is |
| pointofrental | 6 jobs | works as-is |
| tylertech | 236 jobs | works as-is |
| cpasurvey | 0 jobs | genuinely empty — feed returns `<result />`, parses cleanly |
| jumpcrew | 0 jobs | genuinely empty — feed returns `<result />`, parses cleanly |
| fieldcore | fails → 120 jobs | branded redirect; fixed with `?fr=true&nl=1` (issue 2) |
| schnabel | fails → 71 jobs | branded redirect; fixed with `?fr=true&nl=1` (issue 2) |
| synergy | fails → 23 jobs | branded redirect; fixed with `?fr=true&nl=1` (issue 2) |
| imprivata | fails → 50 jobs | branded redirect; fixed with `?fr=true&nl=1` (issue 2) |
| neilpateldigital | fails → 19 jobs | branded redirect; fixed with `?fr=true&nl=1` (issue 2) |
| opentrons | fails → 5 jobs | branded redirect; fixed with `?fr=true&nl=1` (issue 2) |
| leovegas | fails (throws) | genuinely empty board; feed redirects to `NoJobs.htm` (issue 3) |
| zoom, starbucks, servicenow, twilio, blueorigin, sbmmanagement, groupmemea, noblenetwork | fails, `?fr=true&nl=1` doesn't help | confirmed not live Jobvite tenants anymore — slug retired, migrated to other ATS (issue 4) |

Happy to help test a revision if useful.

I moved my own PR (#2529) to draft — I expect you to polish this one up, and once you do, it'll end up more solid than mine. I'll close mine in favor of this one once the issues above are sorted.

Also, genuinely nice find — I'm curious how you landed on the XML feed in the first place.

LayerC0de:
Thanks — this is a genuinely useful review, and the tenant table did most of the work for me. Every finding reproduced. Pushed in 2c159d7.

Two places where I landed somewhere slightly different from your suggestion, both with probes below.

## 1. Rate limiting — fixed, and it's worse than "mass-scan edge case"

Confirmed on a cold run: 200 (945 KB, 40 jobs), immediate 429 `Retry-After: 30`, then 200 again once the 30s was honoured. So the limit is real, it triggers on the second request as you said, and — the part I wanted to be sure of before spending wall-clock on it — **the wait actually clears it**. Retrying is worth doing rather than just polite.

Rather than adding a retry path to this provider, I generalised the shared one. Nothing in `fetchJsonWithRetry`'s loop ever inspected the response body, so the JSON-only signature was incidental: it's now `withRetry()` holding the policy, with `fetchJsonWithRetry` and a new `fetchTextWithRetry` as two thin callers. `fetchJsonWithRetry` keeps its exact signature and behaviour — a16z's tests pass untouched. Given your point 5 is precisely about a decoder that drifted between copies, growing a second private retry implementation here felt like the same mistake one file over.

One deliberate deviation: `maxDelayMs` is 15s for this provider, not the shared 8s. The clamp on `Retry-After` is `maxDelayMs * 4`, so the default ceiling is 32s — only 2s above Jobvite's advertised 30, for a value we don't control. 15s moves the ceiling to 60s, leaving room for Jobvite to widen its window without the clamp silently truncating the wait into another guaranteed 429.

## 2. Branded redirects — fixed, but board-only

`?fr=true&nl=1` works exactly as you described:

| tenant | bare | with params |
|---|---|---|
| fieldcore | 302 → `www.fieldcore.com/careers/jobs/` | 200, eid `qjFaVfwN` |
| imprivata | 302 → `www.imprivata.com/company/join-team` | 200, eid `qtw9VfwN` |
| opentrons | 302 → `opentrons.com/about/jobs/` | 200, eid `qs1bVfwj` |
| egnyte (control) | 200, eid `qLHaVfwh` | 200, eid `qLHaVfwh` — no-op |

**I put them on the board request only, not the feed.** You suggested "every board/feed request", but the redirect looks like a property of the branded *board*, not of the tenant — those same three tenants' feeds answer 200 bare:

```
fieldcore feed (bare)    200 jobs=126
imprivata feed (bare)    200 jobs=53
opentrons feed (bare)    200 jobs=6
```

So adding the params to `Xml.aspx` would be attaching a fix to a request that never had the problem. Happy to be wrong here if you hit a branded tenant whose *feed* redirects — that would be a case my three didn't cover.

## 3. Empty board — fixed, scoped to the feed as you asked

`leovegas` reproduced exactly. Your warning about scoping was the useful part, and it's narrower than "any redirect during fetch": a 3xx, on the feed host, whose target resolves to that one page. The board's `search.jobvite.com?invalid=1` redirect keeps throwing, and there's a test asserting it still does — plus one for `NoJobs.htm` served from a foreign host, which also throws.

Implementation detail worth flagging since it isn't obvious: **the `Location` is relative.** Literally `NoJobs.htm`, not an absolute URL, so it's resolved against the request URL before matching rather than string-compared:

```
leovegas feed   302 -> NoJobs.htm
raw location value: "NoJobs.htm"
resolved:  https://app.jobvite.com/CompanyJobs/NoJobs.htm
```

Mechanically this needed `redirect:'manual'` on the feed instead of `'error'`, since `'error'` throws a bare `TypeError` with no status and no headers. That's the same SSRF guarantee by a different route — undici hands back the 3xx rather than chasing it — so the final hostname still can't move; `'manual'` only buys the ability to read `Location` without acting on it. `_http.mjs` now attaches `err.location` alongside the existing `err.retryAfter`. The file header and the PR's SSRF section both say so now, since the old text claimed `redirect:'error'` everywhere.

## 4. Overstated verification — you're right, corrected

Re-checked all six; your read matches mine. `zoom`, `starbucks`, `servicenow`, `twilio`, `blueorigin` all 302 to `search.jobvite.com?invalid=1`, `tylertech` alone returns 200 with `q6NaVfwI`.

The logical error is worth naming rather than just deleting: those five are *consistent* with "the JSON API is retired" without being *evidence* for it, because a retired tenant and a retired API are indistinguishable from outside. Only a tenant live on one endpoint and dead on the other separates them, and `tylertech` is the only one of the six that qualifies. Both the PR description and the file header now say that instead of counting to six.

## 5. `decodeEntities` — agreed, separate PR

Taking your suggestion exactly: leaving the local copy here and upstreaming the XML 1.0 §2.2 strictness into `_html-entities.mjs` as its own PR, since it changes behaviour for all 13 providers that import it. Worth noting the two aren't a clean superset in either direction — the shared decoder handles `&nbsp;` and arbitrary named entities (correct for HTML, not for XML, where only the five predefined ones exist), while this one rejects NUL/C0/lone surrogates that the shared one emits. So the merge needs a bit of thought about which direction each difference should resolve, which is another reason it doesn't belong in this PR.

The `redirect` assertion you asked for is in, on both requests. That one landed well — three separate behaviours ride on those two values now, and the suite previously only asserted *that* a request happened.

## 6. Docs — added

`docs/SUPPORTED_JOB_BOARDS.md` gets a Jobvite row, and `templates/portals.example.yml` gets both the detection line and two stanzas (`careers_url` + `company_eid:`, and the explicit `provider:`/`api:` form). It had no Jobvite entry at all before, so this was a gap independent of the config change.

---

56 assertions in the provider suite; full suite 3169 passed, 0 failed.

## How I found the feed

Less cleverness than the question deserves, honestly. I wasn't looking at Jobvite — I was trying to shrink a WebSearch handoff list, and Tyler Technologies was one of the stragglers. Rendering their careers page and watching the network tab turned up `app.jobvite.com`, which was the useful part: the backend host is not the host the provider was pointed at.

From there it was mechanical. The documented JSON API returned HTML, so I grepped the board page for anything ID-shaped and got `companyEId: 'q6NaVfwI'` (alongside a Google Analytics tag). Retrying the JSON API with the real eId still failed. At that point I fired four candidate endpoints at once, and the only one that wasn't a `jobs.jobvite.com` variation — `app.jobvite.com/CompanyJobs/Xml.aspx?c=q6NaVfwI` — came back with 1.88 MB of `text/xml`:

```
jobs.jobvite.com/api/company/q6NaVfwI/jobs              302  0b
jobs.jobvite.com/api/company/tylertech/jobs?mode=json   302  0b
jobs.jobvite.com/tylertech/search?...&mode=json         200 text/html    26105b
app.jobvite.com/CompanyJobs/Xml.aspx?c=q6NaVfwI         200 text/xml   1880357b
```

That last URL was a guess at a legacy ASP.NET endpoint on the host the network tab had just exposed. So: the host was observation, the eId was grep, and the path was luck.

Which is also the argument for setting `company_eid:` explicitly rather than relying on discovery — the path is guessable and the host is discoverable, but the eId is neither.

Re: your PR — if you'd rather land yours and have me rebase this on top, I'm fine either way; you had the tenant-quirk map first and that's most of the value here.

ilyakanevskiy:
All three scoping calls make sense to me (board-only for the redirect params, feed-only for the empty-board handling, the `withRetry()` generalization). No objections on any of it.

I'm closing #2529 now in favor of this PR.

Feel free to credit the review in a commit trailer if you want — not something I need either way.

Thanks for the collaboration!

P.S. Thanks also for walking through how you found the endpoint — that kind of detail is damn good.

LayerC0de:
@ilyakanevskiy  Appreciated — and likewise on the collaboration.

I've added a `Reviewed-by:` trailer to the PR description rather than a `Co-authored-by:` one. You said you didn't need either, so this is me overruling you slightly, but only slightly: You found the branded-domain bypass, diagnosed the empty-board redirect *and* flagged that scoping it loosely would swallow the retired-tenant signal, and hand-tested 35 tenants. Three of the six findings wouldn't be in this branch otherwise. The description says so in specifics rather than as a thank-you.

It lands in `main`'s history via the squash body, so no force-push and none of your inline threads get detached.

On #2529: for what it's worth, the tenant-quirk map was the part that made this reviewable. I had one tenant and a working endpoint; you had the layout variants, the pagination cap, and the branded-redirect cases already mapped. Closing yours cost you the PR and kept the knowledge, which is the better trade for the repo but not obviously the better one for you.

One thing you might want to look at, since it's adjacent to your #2657: the board fetch here uses `redirect:'error'` inside `fetchTextWithRetry`, so before your fix a retired tenant would have burned two backoff sleeps on a redirect that could never succeed. Your change makes that path fail fast. I verified the cause string your constant pins (`unexpected redirect`) independently while probing this, on Node 24 — so the two changes compose as intended rather than by luck.

Also: the `decodeEntities` upstreaming is open as #2713. It takes the XML 1.0 §2.2 range check out of this provider and into `_html-entities.mjs`, which is where you said it belonged. It deliberately does *not* adopt the HTML5 windows-1252 remap of C1 references — that changes output rather than rejecting bad input, so it's pinned by a test as a documented non-goal rather than folded in quietly. Once both land, this provider's private copy gets deleted in favour of the import; I've dry-run that swap and the full suite is clean.

santifer:
Thanks for this, @LayerC0de.

You prove the Jobvite JSON endpoint is gone: it 302s to search.jobvite.com?invalid=1, so fetchJson gets a marketing page back and the provider silently returns zero listings, because from the outside an empty board and a retired API look identical. The tenant that shows it is tylertech: alive in the XML feed with 236 listings, dead on the JSON API. You also fixed your own PR body when @ilyakanevskiy pointed out in review that the other five tenants you cited (zoom, starbucks, servicenow, twilio, blueorigin) weren't evidence: they had migrated off Jobvite entirely.

santifer:
Merged, @LayerC0de: catching a provider whose upstream API silently became a redirect-to-nowhere and rebuilding it on the public XML feed with tiered eId resolution keeps real users scanning. Zero-auth verified in the diff, as always with your providers. 🚀

PR #1973, opened 2026-07-16
fix(web): prevent Codex run jobs from waiting on stdin

gregaro:
## Summary
- spawn /api/run CLI children with stdin ignored, matching the apply prefill route
- prevents Codex exec jobs from hanging on an open stdin pipe until the kill timer
- avoids the misleading installed/authenticated error after timeout

## Verification
- npm test --prefix web
- npm run typecheck --prefix web

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved command execution reliability by preventing processes from waiting for input.
  * Ensured command output and error messages continue streaming correctly during execution.

PR #2626, opened 2026-08-09
fix(web): launch npm CLIs on Windows

configurowebmax:
## Summary
- resolve npm-generated Windows shims to their real JavaScript or native entrypoint
- launch web CLI workers without shell interpolation and with stdin closed
- surface stderr when the assistant CLI exits without stdout
- add cross-platform and npm-shim unit coverage

## Problem
On Windows, `findBin()` could select an extensionless npm shim and `spawn()` returned `ENOENT`. Directly launching the `.cmd` shim returned `EINVAL`, while routing prompts through PowerShell could alter multiline or quoted arguments. Resolving the trusted npm wrapper to its actual entrypoint avoids those failures without enabling `shell: true`.

## Validation
- `cd web && npm test` - 66 passed, 0 failed
- `cd web && npm run build`
- `node test-all.mjs --quick` - 3133 passed, 0 failed (2 expected warnings)
- end-to-end on Windows: `/api/assistant` with Gemini CLI returned `GEMINI_WEB_OK`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved command-line tool launching on Windows, including npm-installed tools and PowerShell wrappers.
  * Preserved arguments containing spaces and special characters.
  * Improved assistant error reporting when no output is produced.
  * Added safe fallback behavior when launch wrappers are missing or unreadable.

* **Tests**
  * Added coverage for Windows, POSIX, native executable, wrapper, argument-handling, and fallback scenarios.

### DIFF
--- a/web/src/app/api/apply/prefill/route.ts
+++ b/web/src/app/api/apply/prefill/route.ts
@@ -1,7 +1,7 @@
 import { spawn } from "node:child_process";
 import fs from "node:fs";
 import path from "node:path";
-import { resolveCli } from "@/lib/clis";
+import { prepareCliLaunch, resolveCli } from "@/lib/clis";
 import { careerOpsRoot, readMemory } from "@/lib/career-ops";
 import { getSession } from "@/lib/apply/session";
 
@@ -157,7 +157,8 @@ Output ONLY a compact JSON object mapping each field id → {"value": "...", "ne
 
       const result = await new Promise<{ buf: string; code: number | null; signal: NodeJS.Signals | null }>((resolve) => {
         // stdin = /dev/null so the CLI doesn't wait 3s for piped input.
-        const child = spawn(binPath, args, { cwd: careerOpsRoot(), env: process.env, stdio: ["ignore", "pipe", "pipe"] });
+        const launch = prepareCliLaunch(binPath, args);
+        const child = spawn(launch.command, launch.args, { cwd: careerOpsRoot(), env: process.env, stdio: ["ignore", "pipe", "pipe"] });
         let buf = "";
         let firstByteAt = 0;
         const hb = setInterval(() => {
--- a/web/src/app/api/assistant/route.ts
+++ b/web/src/app/api/assistant/route.ts
@@ -1,5 +1,5 @@
 import { spawn } from "node:child_process";
-import { resolveCli } from "@/lib/clis";
+import { prepareCliLaunch, resolveCli } from "@/lib/clis";
 import { careerOpsRoot, readMemory, doctorState } from "@/lib/career-ops";
 
 export const runtime = "nodejs"; // child_process (spawn) requires the Node runtime
@@ -109,7 +109,8 @@ export async function POST(req: Request) {
       ]
     : spec.args(prompt);
 
-  const child = spawn(binPath, args, { cwd: careerOpsRoot(), env: process.env });
+  const launch = prepareCliLaunch(binPath, args);
+  const child = spawn(launch.command, launch.args, { cwd: careerOpsRoot(), env: process.env, stdio: ["ignore", "pipe", "pipe"] });
 
   const encoder = new TextEncoder();
   // `closed` + kill timer in the OUTER scope so cancel() can flip `closed` before
@@ -120,6 +121,7 @@ export async function POST(req: Request) {
   const stream = new ReadableStream<Uint8Array>({
     start(controller) {
       let buf = "";
+      let stderrBuf = "";
       let emitted = false;
       killer = setTimeout(() => {
         try {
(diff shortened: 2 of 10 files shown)


santifer:
Thanks for this, @configurowebmax.

On Windows findBin() could pick the extensionless npm shim and spawn() returned ENOENT; launching the .cmd directly gave EINVAL; and routing through PowerShell mangled multiline or quoted arguments, which means the prompts. Your way out is the right one security-wise: resolve the npm wrapper down to its real JavaScript or native entrypoint instead of turning on shell: true, spawn the workers with no shell interpolation and stdin closed, and surface stderr when the CLI exits with no stdout. Verified end to end on Windows: /api/assistant with Gemini CLI returned GEMINI_WEB_OK.

santifer:
Reviewed in full. **The core of this is right and I want it in** — this comment is about what it will look like after the PR ahead of it lands, so you rebase once against a settled shape instead of guessing.

## `prepareCliLaunch` is the right design

Resolving the trusted npm `.ps1` wrapper to its real entrypoint, instead of reaching for `shell: true`, is the correct call — `shell: true` would reintroduce quoting/interpolation problems on exactly the multiline prompts this app sends.

What makes it reviewable is the failure direction: **every uncertain path returns `{ command: binPath, args }`** — wrong extension, missing `.ps1`, unreadable file, regex miss, target that doesn't exist, target of an unexpected type. In each case the behaviour is *what happens today*, so a shim shape you didn't anticipate degrades to the current ENOENT rather than to something new and creative. That's the property I check for first and it's present throughout.

`platform = process.platform` as a defaulted parameter rather than a module-level read is what makes the cross-platform tests real instead of skipped-on-macOS. Good.

## It composes with #2102 — it doesn't compete

There are four PRs currently rewriting how the web spawns CLI workers (#2102, #2626, #2689, #2633), and this is the one I had to look at hardest to place. The conclusion is that yours is **orthogonal**, not a rival:

- `prepareCliLaunch` answers **what to execute** (shim → real entrypoint)
- #2102's `spawnHeadlessCli` answers **how to launch it** (stdin closed, one place)

They stack:

```ts
const launch = prepareCliLaunch(binPath, args);
const child = spawnHeadlessCli(launch.command, launch.args, { cwd: careerOpsRoot(), env: process.env });
```

So nothing of yours is lost by #2102 going first. It goes first only because it touches the same 7 routes and is the wider change; ordering it the other way would make *it* redo the work instead of you.

## One hunk of yours is already in `main`

Your diff adds `stdio: ["ignore", "pipe", "pipe"]` alongside the launch change. That landed separately while this PR was open (#1973), and today `main` already has it in 4 of the 7 files you touch:

```
run/route.ts             already closed
apply/drive.ts           already closed
apply/agent-interpret.ts already closed
apply/prefill/route.ts   already closed
assistant/route.ts       still open   ← your change is still needed here
cv/ingest/route.ts       still open   ← and here
explore/ai/route.ts      still open   ← and here
```

On rebase those four hunks resolve as **"already applied"** — keep `main`'s version, drop yours, nothing of your work is lost. The remaining three are covered by #2102's `spawnHeadlessCli`, so after that lands your diff shrinks to `cli-launch.mjs` + its tests + the `prepareCliLaunch` call sites. That's the version I'd like to merge.

## What I'd ask for, but not yet

Don't rebase now — #2102 is still moving and you'd be aiming at a target that shifts. I'll ping you the moment it lands, with the exact remaining delta.

Same for **#2633**: it's yours too, it's in the same cluster, and I left a separate note there about `docs/SUPPORTED_CLIS.md` still listing Kimi as interactive-only. That one you *can* act on independently of all this.

Neither PR is waiting on merit. They're waiting on an ordering problem that is ours, and I should have told you that sooner than five days in.

santifer:
**#2102 is merged — this is unblocked, and now is the moment to rebase.**

As I said in the review, your `prepareCliLaunch` composes with what just landed rather than competing with it:

```ts
const launch = prepareCliLaunch(binPath, args);
const child = spawnHeadlessCli(launch.command, launch.args, { cwd: careerOpsRoot(), env: process.env });
```

`spawnHeadlessCli` (from #2102) now answers *how to launch* — stdin closed, in one place, for every CLI route. Yours answers *what to execute* — npm shim resolved to its real entrypoint. Nothing of yours was made redundant.

What shrinks: your `stdio: ["ignore", …]` hunks. All seven routes now go through `spawnHeadlessCli`, so those resolve as "already applied" — keep `main`'s version, drop yours. Your remaining delta is `cli-launch.mjs`, its tests, and the `prepareCliLaunch` call sites.

**#2633 is yours too** and also unblocked. It still needs the `docs/SUPPORTED_CLIS.md` row I mentioned — that table lists Kimi as `Interactive:` only, and #2688 added a test that derives the expected `KNOWN` contents from it, so the two need to agree.

Ping me when either is rebased and I'll review the same day. Sorry for the wait — the ordering problem was ours, and your PRs were never stalled on merit.

santifer:
Concrete update, @configurowebmax, and it changes what's worth doing with this PR today.

PR #2633, opened 2026-08-09
feat(web): add Kimi Code CLI support

configurowebmax:
## Summary
- register Kimi Code CLI as a selectable web runtime using its official `kimi -p` prompt mode
- search the official `~/.kimi-code/bin` installation directory in addition to `PATH`
- add unit coverage for the CLI specification and install-directory discovery

## Validation
- Kimi Code CLI `0.29.1` authenticated smoke test - passed
- `/api/clis` detected the native Windows executable as installed
- `cd web && npm test` - 62 passed, 0 failed
- `cd web && npm run build`
- `node test-all.mjs --quick` - 3133 passed, 0 failed (2 expected warnings)

No API keys, OAuth tokens, CV data, or profile data are included in this change.

## Summary by CodeRabbit

* **New Features**
  * Added support for detecting and launching the Kimi Code CLI.
  * Included Kimi Code’s official documentation link, prompt configuration, and default installation path.
  * Added streaming display of Kimi responses and tool activity.
  * Added support for structured Codex and Kimi assistant responses.
  * Improved CLI configuration synchronization and persistence across the application.
  * Improved result recovery using session and local storage.

* **Bug Fixes**
  * Added clearer timeout, authentication, quota, exit-code, and empty-response diagnostics.
  * Extended processing time for Kimi evaluations and PDF tasks.

- end-to-end Windows: /api/assistant with Kimi returned KIMI_WEB_OK (HTTP 200)

### DIFF
--- a/web/src/app/api/assistant/route.ts
+++ b/web/src/app/api/assistant/route.ts
@@ -1,11 +1,16 @@
 import { spawn } from "node:child_process";
 import { resolveCli } from "@/lib/clis";
 import { careerOpsRoot, readMemory, doctorState } from "@/lib/career-ops";
-import { parseKimiStreamLine, prepareRunArgs } from "@/lib/cli-run.mjs";
+import {
+  noOutputMessage,
+  parseCodexStreamLine,
+  parseKimiStreamLine,
+  prepareAssistantArgs,
+} from "@/lib/cli-run.mjs";
 
 export const runtime = "nodejs"; // child_process (spawn) requires the Node runtime
 export const dynamic = "force-dynamic";
-export const maxDuration = 120;
+export const maxDuration = 300;
 
 const SYSTEM_PREAMBLE = `You are the career-ops assistant — a proactive, friendly career co-pilot for a person who is actively job-hunting. You live inside their LOCAL career-ops web dashboard (a pipeline of evaluated jobs, A–F reports, their CV, analytics) and run on their own AI CLI.
 
@@ -92,6 +97,7 @@ export async function POST(req: Request) {
   // files directly. Scope its tools so it can advise (read) but not blind-write.
   const isClaude = cliId === "claude";
   const isKimi = cliId === "kimi";
+  const isCodex = cliId === "codex";
   // allowedTools must be COMMA-separated; disallowedTools is the hard guardrail
   // so the advisor can read (and WebFetch) but never blind-writes or shells out.
   const args = isClaude
@@ -109,7 +115,7 @@ export async function POST(req: Request) {
         "--disallowedTools",
         "Bash,Write,Edit,NotebookEdit,Task",
       ]
-    : prepareRunArgs(cliId, spec.args(prompt));
+    : prepareAssistantArgs(cliId, spec.args(prompt));
 
   const child = spawn(binPath, args, { cwd: careerOpsRoot(), env: process.env });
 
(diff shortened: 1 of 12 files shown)


santifer:
Thanks for this, @configurowebmax.

Registers Kimi Code CLI as a selectable runtime in the web using its official prompt mode `kimi -p`, and adds a lookup for the official install directory ~/.kimi-code/bin on top of PATH, which is exactly where detection breaks on native installs. Real validation, not a checklist: authenticated smoke test against Kimi 0.29.1, /api/clis detecting the native Windows executable, 62 web tests, typecheck, build, and test-all --quick with 3,133 asserts. You state explicitly that it ships no API keys, OAuth tokens, or CV/profile data.

santifer:
Two things on this one — one is a concrete gap I'd like closed, the other is about ordering and isn't your fault.

## The canonical doc still says Kimi is interactive-only

`docs/SUPPORTED_CLIS.md:14`:

```
| Kimi | `KIMI.md` | Interactive: `kimi` |
```

No `Headless/Batch:` row, unlike Grok (`grok -p`), Qwen (`qwen -p`) and Copilot (`copilot -p`). This PR adds Kimi to `KNOWN` — the list of CLIs the web drives **headlessly** — without updating that table, so the repo would ship two sources disagreeing about the same CLI.

That matters more than usual right now: **#2688 adds a test that derives the expected `KNOWN` contents from exactly that table**, precisely because Grok was documented, wired into `doctor.mjs`, the scaffolder and the README, yet missing from `KNOWN` with nothing tying the lists together. Adding an entry that the table contradicts re-opens that gap from the other side.

Your validation notes say you ran an authenticated smoke test against Kimi Code CLI 0.29.1, so I'll take it that `kimi -p` does work headlessly — I can't verify that without the binary and I'm not going to pretend otherwise. **If it works, the fix is to update `docs/SUPPORTED_CLIS.md` in this PR** with the `Headless/Batch: kimi -p "prompt"` form, so the doc and `KNOWN` agree and #2688's guard covers Kimi too. If it doesn't work headlessly, then the entry shouldn't be in `KNOWN` at all.

## On timing

I mapped the 8 open PRs touching `web/src/lib/clis.ts` as a cluster today, and this one isn't the 3-line addition its `clis.ts` diff suggests — it touches 12 files, and 5 of them overlap **#2626** (also yours), which in turn overlaps **#2102** on 8 files. Four PRs from three authors are currently rewriting how the web spawns CLI workers, in the same routes.

So this isn't waiting on its own merit. It's waiting because merging it before that base settles would force the other three into a rebase they'd each have to redo. I'm landing the spawn-unification PR first, then the rest rebase onto one shape instead of three. I'll come back with the exact order and ping you on both of yours — you have two PRs in that cluster and I don't want either to go stale in silence.

The docs line is the one thing you can do now, and it's independent of all that.

santifer:
Concrete update, @configurowebmax, and it changes what's worth doing with this PR today.

santifer (changes requested):
Thanks @configurowebmax for **adding Kimi Code as a CLI you can pick in the web app for evaluations, the assistant and AI search**. To get this landed we need a few concrete things; this is the complete list, not a first round:

- [ ] Keep this PR to Kimi only: the Codex JSON output for the assistant, the 240 s assistant wait for every CLI and the Explore `localStorage` results each change behaviour for everyone, so please open them as separate PRs
- [ ] Rebase on `main`: #2626 touches the same launch code, so rebasing once after it settles saves you a second pass
- [ ] Resolve or answer the 7 open CodeRabbit threads

A new CLI next to the existing ones is welcome here, and the tests in `web/tests/lib/cli-run.test.mjs` are what make one maintainable. There is no clock, and the PR stays yours.

PR #2636, opened 2026-08-09
feat(cli): add --help to followup-cadence

### LINKED ISSUE #2477: feat(cli): followup-cadence.mjs has no --help — add the usage block

aryansk:
## Summary

- Add `--help` and `-h` usage output to `followup-cadence.mjs`.
- Handle help before profile and application data resolution so the command is side-effect free.
- Add a regression test that checks the documented flags.

Closes #2477

## Validation

- `node followup-cadence.test.mjs` (22 passed)
- `node followup-cadence.mjs --help`
- `git diff --check`

### DIFF
--- a/followup-cadence.mjs
+++ b/followup-cadence.mjs
@@ -9,6 +9,7 @@
  *      node followup-cadence.mjs --summary   (human-readable dashboard)
  *      node followup-cadence.mjs --overdue-only
  *      node followup-cadence.mjs --applied-days 10
+ *      node followup-cadence.mjs --help       (print usage and exit)
  */
 
 import { readFileSync, existsSync } from 'fs';
@@ -31,6 +32,22 @@ const summaryMode = args.includes('--summary');
 const overdueOnly = args.includes('--overdue-only');
 const appliedDaysIdx = args.indexOf('--applied-days');
 const appliedDaysOverride = appliedDaysIdx !== -1 ? parseInt(args[appliedDaysIdx + 1], 10) : null;
+
+const KNOWN_FLAGS = ['--summary', '--overdue-only', '--applied-days', '--help', '-h'];
+const USAGE = `Usage:
+  node followup-cadence.mjs                 # JSON for active applications
+  node followup-cadence.mjs --summary       # human-readable dashboard
+  node followup-cadence.mjs --overdue-only  # show only overdue applications
+  node followup-cadence.mjs --applied-days 10 # override the first follow-up interval
+  node followup-cadence.mjs --help          # print this usage block and exit
+  Supported flags: ${KNOWN_FLAGS.join(', ')}`;
+
+// Keep help side-effect free: resolveCadenceConfig() reads the profile file,
+// so this branch must run before the module-level cadence is loaded.
+if (args.includes('--help') || args.includes('-h')) {
+  console.log(USAGE);
+  process.exit(0);
+}
 
 // --- Cadence config ---
 export const DEFAULT_CADENCE = {
--- a/followup-cadence.test.mjs
+++ b/followup-cadence.test.mjs
@@ -9,6 +9,7 @@
 
 import { dirname, join } from 'path';
 import { fileURLToPath } from 'url';
+import { execFileSync } from 'child_process';
 
 const ROOT = dirname(fileURLToPath(import.meta.url));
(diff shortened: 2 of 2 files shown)


santifer:
Thanks for converting your claim into a PR so cleanly, @aryansk: help-before-data-resolution (so --help is side-effect free) plus a regression test on the documented flags is exactly the pattern #2477 asked for. Where it stands: it is in today's review batch, so the next thing you hear will be a decision with reasons, not a nudge. Sorry it sat 3 days without a word.

aryansk:
Thanks for the clear review framing. The draft remains focused on making --help side-effect free with regression coverage for the documented flags; I will wait for the decision and reasons.

santifer:
Update rather than silence, @aryansk: **the shared helper you'd want now exists.** `lib/cli-flags.mjs` landed on 14 Aug (#2778) with `validateFlags(args, knownFlags, usage, { valueFlags })`, and it rejects unknown flags *before* honouring `--help` — so `--help --bogus` errors instead of exiting 0 having never looked at the typo. That ordering had to be corrected twice in review on separate PRs before it was centralised.

Your PR predates it, so if it hand-rolls the check, delegating is a small change and would make this consistent with the four that landed tonight (#2874, #2873, and two queued). Keep your `USAGE` block as-is — that part is per-script and shouldn't be centralised.

Sorry for the slow reply. Push the change and I'll pick it up next cycle.

FReptar0:
Heads-up so you don't spend another rebase on this, @aryansk — the `--help` block for `followup-cadence.mjs` **already landed** on `main`:

```
$ node followup-cadence.mjs --help ; echo "exit=$?"
Usage:
  node followup-cadence.mjs                    # full JSON analysis to stdout
  ...
  node followup-cadence.mjs --help|-h          # print this usage block and exit
exit=0
```

`followup-cadence.mjs:925-937`, via `3d8b53d` (2026-08-11) — a **stats.mjs** PR for #2556 that patched this file too, so nothing links back here and there was no way to see it from the issue. That's why #2477 stayed open until today; I've closed it with the same evidence.

Genuinely nothing wrong with your work — it got overtaken by a commit that never mentioned it. Leaving the PR alone for @santifer to close; just didn't want you rebasing into a conflict for something already in.

santifer:
@aryansk, closing this one with an apology attached: while your draft sat here, the usage block reached `followup-cadence.mjs` as part of #2556's repo-wide --help sweep, so main already ships what this PR set out to add (verified tonight: `KNOWN_FLAGS` and the usage block are live, and `--help` exits 0). That's a queue-timing failure on our side, not anything you did wrong. The shelf currently has 46 verified-free good-first-issues: https://github.com/santifer/career-ops/issues?q=is%3Aissue%20is%3Aopen%20label%3A%22good%20first%20issue%22%20no%3Aassignee — claim one with a comment and it gets assigned to you on the spot. I'd be glad to see you back.

santifer:
@aryansk, we closed this promising a good-first-issue pointer and never delivered it: that's on me. The live pool is at https://github.com/santifer/career-ops/issues?q=is%3Aopen+label%3A%22good+first+issue%22+no%3Aassignee (claim with `/assign`). Your --help addition here was exactly the kind of polish that compounds; several scripts still lack usage blocks, so if CLI ergonomics is your thing, say the word and I'll seed one with your name on it.

PR #2640, opened 2026-08-09
feat(providers): extend getro with auto-resolve, salary, retry and remote-detect

ilyakanevskiy:
## Summary

Extends the Getro VC-portfolio job-board provider merged in #825 (thank you @eliador90 for the original work — the age-cutoff pagination design here is yours, kept as-is). This supersedes #2602 (my earlier, independent getro PR, opened before I'd seen #825 land — now closed in favor of building directly on top of the merged version instead of re-litigating a duplicate implementation).

Kept from #825, unchanged:
- Newest-first pagination with an age-cutoff bound (`getro_max_age_days`, default 90) — stops paginating once postings go stale instead of walking the full board every scan.
- `getro_collection` as an explicit config override.
- `getro_max_pages` safety ceiling.

## What's added

- **Auto-resolve `collection_id`.** `getro_collection` is now optional — when absent, the id auto-resolves from `careers_url`'s own `__NEXT_DATA__` blob (`network.id`) on first fetch, no manual devtools lookup required. `careers_url` is required either way (also supplies the `referer` header the search API needs — a bare request without one returned 406 in manual testing).
- **Retry on transient failures.** Every page fetch now goes through the shared `fetchJsonWithRetry` (`providers/_http.mjs`) instead of a bare try/catch — a single 429/5xx/timeout-abort mid-sweep no longer truncates the whole board (same class of bug as #2506; caught live during testing — see Test plan). Generalized `fetchJsonWithRetry` into a shared `withRetry()` core and added a sibling `fetchTextWithRetry`, applied to the collection_id auto-resolve fetch too — that single pre-pagination request had no retry at all, so a DNS/TLS/connection blip on it would fail the whole board before a single page was fetched, an even worse blast radius than losing one page mid-sweep. (This one is defensive — not itself the fix for the bug below, see Test plan.)
- **Fixed a `redirect:'error'` + redirecting `careers_url` interaction.** b2venture's board root (`https://jobs.b2venture.vc`, the URL already configured in #825's own example) 302s to `/companies`; combined with the mandatory `redirect:'error'` SSRF guard, that made the auto-resolve fetch fail deterministically (`fetch failed` / "unexpected redirect"), not transiently — retries didn't help and weren't expected to. Fixed by pointing `careers_url` at a non-redirecting path (`/jobs`) in the example. Flagging since any existing config with a `careers_url` that 302s could hit the same wall once it starts being fetched for auto-resolve.
- **Salary extraction.** `compensation_amount_min/max_cents` + `compensation_currency`, gated on `compensation_period === 'year'` (a non-annual period is treated as "no usable annual figure").
- **Fuller location handling.** All `locations`/`searchable_locations` entries are joined (not just the first), and a "Remote" tag is appended from `work_mode === 'remote'` when not already present.
- **`ctx.maxPages` support.** Per `providers/README.md`'s convention ("paginating providers should honor the optional `ctx.maxPages` hint — the portal health probe passes 1") — previously unhonored here.
- **Inter-page pacing** (`INTER_PAGE_DELAY_MS`, honors `ctx.sleep` when present) — a large board is still a long burst of same-host requests.
- **`HARD_MAX_PAGES` raised 200 → 1500.** 200 pages (4,000 jobs) undersizes real large boards — Insight Partners' portfolio alone needs ~577 pages (~11,500 listings).
- Non-Error rejection handling in the page-fetch catch (a promise can reject with anything; reading `.message` off a non-Error would throw inside the catch itself).

## Docs / examples

- `docs/SUPPORTED_JOB_BOARDS.md` — added the missing Getro row (the merged #825 didn't add one).
- `templates/portals.example.yml` — expanded from 2 to 8 example funds: b2venture, Cherry Ventures (kept from #825, using `getro_collection`), Atomico, HV Capital, Point Nine, Earlybird Venture Capital, Speedinvest, Insight Partners (new, all using auto-resolve).

## Test plan

- `node tests/providers/getro.test.mjs` — 26/26 pass: explicit + auto-resolved collection_id, `__NEXT_DATA__` extraction, non-https rejection, redirect:'error', referer header, salary extraction (incl. a non-annual period being ignored), multi-location + Remote-tag join, Unix-seconds→ms conversion, `getro_max_pages`/`ctx.maxPages` clamping, the age-cutoff bound, retry-and-recover for both the page-fetch and the collection_id auto-resolve fetch, non-Error rejection tolerance, malformed/empty payload tolerance.
- `node test-all.mjs` — full suite green (3151 passed / 0 failed / 1 pre-existing unrelated warning).
- Live scan against my own production `portals.yml` (8 real Getro entries) surfaced two real issues neither #825 nor my original #2602 had caught:
  - A mid-pagination timeout-abort truncating Point Nine's board at page 17 — transient, fixed by the retry above.
  - b2venture failing outright on the very first request — NOT transient; root-caused to the `redirect:'error'` + redirecting `careers_url` interaction above and fixed by correcting the URL, not by retry (the auto-resolve fetch is now also retried, but as independent defense-in-depth for genuine future transient failures, not as the fix for this specific one).

## Related

Closes #2602 (superseded by this PR, built on top of #825 instead of duplicating it).

## Summary by CodeRabbit

* **New Features**
  * Added support for Getro-hosted job boards.
  * Supports automatic or manual collection configuration.
  * Added pagination, page limits, age cutoffs, and reliable data retrieval.
  * Normalizes locations, remote-work indicators, and annual salary information.

* **Documentation**
  * Added Getro to the supported job boards list.
  * Added configuration examples for several portfolio boards.

* **Bug Fixes**
  * Improved handling of temporary request failures, retries, and incomplete results.


ilyakanevskiy:
@coderabbitai review

ilyakanevskiy:
@coderabbitai review

eliador90:
Thanks for the heads-up on #825 and for building on it rather than around it, @ilyakanevskiy. Closing your own working implementation to extend someone else's is not the default instinct and I appreciate it. The auto-resolve is the piece I should have written and didn't: hand-copying `network.id` out of devtools was always the worst part of adding a board.

I read the `providers/_http.mjs` change closely, since that file is shared by every provider and has bitten this codebase before. The `withRetry(ctx, fn, policy)` split looks right to me: `fetchJsonWithRetry` keeps its exact public signature so nothing downstream moves, `withRetry` itself stays unexported so the surface only grows by one function, and `fetchTextWithRetry` fills a real gap. Your point about the pre-pagination config fetch having no retry at all is the one I'd underweighted; a blip there does have a worse blast radius than losing a page mid-sweep.

One substantive thing, and it is about `isRetryableError` rather than your refactor.

`redirect:'error'` failures are currently classified as retryable. Reproduced against the exact URL from your PR description:

```
fetch('https://jobs.b2venture.vc', { redirect: 'error' })

  error name   : TypeError
  error message: fetch failed
  err.status   : undefined
  isRetryableError() -> true
```

The guard is `return status === undefined`, and undici raises redirect refusals as a bare `TypeError` with no `status`, so they fall into the network-error bucket. With `RETRY_DEFAULTS = { retries: 2, baseDelayMs: 500 }` that means three requests plus backoff against a third party for a failure that is deterministic and will never succeed on attempt two.

Before your PR this cost nothing, because getro had no retry. Wrapping the fetches is what turns it into wasted requests. And since `redirect:'error'` is mandatory on every provider (it is the SSRF guard from #825), this applies well beyond getro: any provider that meets a redirecting host now pays 3x for it.

You have fixed the b2venture case at the source by resolving the redirect properly, so the happy path is fine. This is about the next redirecting host, not this one.

Worth distinguishing a refused redirect from a transient network error, something like:

```js
if (status === undefined && /redirect/i.test(err?.cause?.message ?? '')) return false;
```

I have not checked how stable that `cause` message is across undici versions, so treat it as a direction rather than a patch. Happy to be told the 3x is acceptable and not worth the complexity; it is a small cost, just an avoidable one.

Nothing else from me. Adjust the age-cutoff default if it gets in your way, it was tuned for the boards I happened to be watching, not from anything principled.

ilyakanevskiy:
Thanks for the thorough review, @eliador90 — and for reproducing the redirect-refusal shape rather than describing it abstractly.

One correction to your framing: "well beyond getro, any provider" overstates it. Only 5 of 72 provider modules retry on fetch failure at all — the other 67 just throw once on a redirect refusal, no wasted attempts. Real, but "5 of 72," not "every provider."

The fix you sketched is implementable, and confirmed distinguishable from DNS/timeout/abort failures via `err.cause.message`. Opened #2657 with the fix plus migrating workday/oraclecloud onto the shared retry helper. `ashby.mjs` is a real behavior change (retries unconditionally today, not just deduplication) so it's split into its own follow-up rather than riding in there.

ilyakanevskiy:
@coderabbitai review

eliador90:
You are right and my framing was wrong, @ilyakanevskiy. "Any provider that meets a redirecting host" was careless: retrying is opt-in, so the cost only lands on modules that actually use the retry helper. I checked upstream `main` before conceding rather than just taking your word for it, and I get 3 on the shared helper (`a16z-speedrun-talent`, `oraclecloud`, `workday`) plus a handful with their own retry logic, which is your 5 give or take how `ashby` is counted. Either way the shape is "a handful, opt-in", not "everything", and the 67 that throw once were never affected.

Worth correcting because it changes the priority: this is a small cleanup, not a fleet-wide bug, and I implied the latter.

Thanks for turning it into #2657 rather than leaving it in a comment thread.

ilyakanevskiy:
Re: [review #4895607305](https://github.com/santifer/career-ops/pull/2640#pullrequestreview-4895607305) — the two "outside diff range" CodeRabbit findings on `providers/getro.mjs`. Both were verified against current code; no code change for either.

**Lines 70-85 — "budget the page ceiling against the provider deadline"**

There's no deadline/cancellation concept anywhere in this codebase's provider contract. `Context` (`providers/_types.js`) only exposes `transport`, `fetchText`, `fetchJson`, `fetchResponse`, `maxPages`, `sleep` — no `deadline`, `signal`, or time-budget field — and `scan.mjs` never constructs one when it calls `provider.fetch()`. The existing `HARD_MAX_PAGES` (1500) is already the safety ceiling for this provider. Making pagination deadline-aware would mean designing a new cross-provider cancellation contract from scratch (touching every provider + `scan.mjs`/`scan-ats-full.mjs`), which is a separate feature, not a fix to this PR's diff.

**Lines 218-263 — "make later-page truncation observable, matching Workday/iCIMS"**

`workdayTruncated`/`icimsTruncated` are only ever read in `scan-ats-full.mjs` (L791-792, L853), and that file never calls the Getro provider — Getro isn't one of its registered sources (Greenhouse/Lever/Ashby/Workday/iCIMS only). Getro's actual caller is `scan.mjs`, and `scan.mjs` doesn't read any `*Truncated` flag for *any* provider today, including Workday and iCIMS. So there's no existing caller-side pattern for Getro to be "consistent with" — attaching a `getroTruncated` flag with nothing wired to consume it would just be dead metadata. Left as-is; the `console.error` at truncation time (L260) remains the actual signal, same as it is for every other paginating provider without a caller-side truncation consumer.

santifer:
@ilyakanevskiy, two notes on this one, and the first is bigger than the PR.

**You built the thing I opened an issue about an hour later.** #2709 is about `providers/_http.mjs` being hard to adopt: 81 providers, 9 importing it, and the ones reimplementing retry locally doing so because the shared helper only wraps `fetchJson`. `workable.mjs` even documents that as its reason. Your diff adds `fetchTextWithRetry` and a generic `withRetry`, which is point 1 of that issue, already written.

So this PR is no longer just a getro improvement: it is the shared-layer fix, with a provider migration attached. I have said so in #2709 and will close that point against this PR when it lands.

The one thing it does not cover is the give-up policy (`workable.mjs`'s `GIVE_UP_RETRY_AFTER_MS`: bail immediately when the server declares a wait longer than N, instead of clamping and continuing). Not asking you to add it here, and I would rather you didn't: this PR is already doing two jobs. It stays open in #2709 as the remaining piece.

**Second, about #2708.** Please credit @IaroslavMazur in the commit body when this merges: `Co-authored-by:` if you take any of his lines, or a plain "field report and WAF measurement by @IaroslavMazur" if not. His dated observation (2026-08-08, `403 "You've been blocked"`, ~2.5h on a 40-page board) is the evidence that justifies pacing at all, and he closed his own PR in favour of yours after you showed him your data. Someone who does that should be visible in the history, not just in a closed tab.

Worth saying plainly, since I only watched it happen: you found the overlap yourself, brought a measurement (8 real boards at 150ms, zero 403s), and offered to change your own number if his was better. That is a maintainer's job done by a contributor, and better than I would have done it, since I hadn't even noticed the collision.

Reviewing this properly now. It sits on top of the current main and it is the widest of the getro changes, so it goes in as the one that lands.

ilyakanevskiy:
@santifer Done — @IaroslavMazur and #2708 are credited in the first commit's body: [`a6301cf`](https://github.com/santifer/career-ops/pull/2640/commits/a6301cf3edc363ed23e74d1323fbb1158d4e4fdb) (`feat(providers): extend getro with auto-resolve, salary, retry and remote-detect`). Force-pushed, tree unchanged — only that commit message changed.

ilyakanevskiy:
@santifer, finally, we're ready to go with this!

santifer:
Saw the "ready to go", @ilyakanevskiy, and then the merge queue moved under you: this now conflicts with `main` because of what I landed today (#2769 touched `providers/_http.mjs` and a couple of provider merges touched `portals.example.yml`). That is my doing, not a mistake of yours, and I am sorry for the extra lap. Rebase onto today's main and ping me here: you go to the front of the review batch, since the work was already done and waiting.

ilyakanevskiy:
Rebased onto today's main — no worries about the extra lap. Conflict was in providers/_http.mjs (withRetry's arg order diverged between my branch and #2769's changes elsewhere in the file); kept main's existing withRetry(request, ctx, policy) signature and folded in the non-Error-rejection fix on top. templates/portals.example.yml merged clean. Provider tests for _http and getro both pass locally.

@santifer, ready for review.

Scott-Emberson (approved):
Approving from the code-owner side for `tests/`. Our owned surface in this PR is `tests/providers/getro.test.mjs` (+232/-23); the review below speaks only to that file, not to the provider or config changes, which sit with their own owners.

The test file holds up. It grows the #825 provider-contract suite from 8 to 30 assertions and tracks every new behavior: explicit-vs-auto-resolve collection id, `extractCollectionId` parsing (missing blob, malformed JSON, reordered/extra attributes, single-quote, and the `data-id` false-match guard), the https-only `careers_url` check, referer derivation, multi-location plus Remote join, annual-only salary extraction, `ctx.maxPages` vs `getro_max_pages` clamping, the age-cutoff bound, transient-retry recovery on both the page fetch and the pre-pagination auto-resolve fetch, and the first-page-throws vs later-page-truncates split.

The assertions carry real sentinels (`company === 'Acme'`, `location === 'Zurich, Bern, Remote'`, `salary.min === 120000`, exact attempt and page counts), so a no-op regression fails them instead of sliding through. Two have teeth beyond their own file: the `'plain string rejection'` first-page test genuinely exercises the `_http.mjs` non-Error guard (without it, the strict-mode `err.attempts` assignment throws a TypeError and the message assertion fails), and the clamp test asserts `<= 1500` with an internal throw past it, so a broken clamp cannot pass. Hermetic throughout: faked `fetchJson`/`fetchText` seams, `sleep` stubbed on every multi-iteration path, no network, same helpers as the neighbouring provider tests.

Ran `node tests/providers/getro.test.mjs` at head `38e0117`: 30/30 pass. CI is green across the three-OS matrix, so I relied on that for the full suite rather than re-run it.

One optional nit, not blocking: the empty-payload test at line 739 returns `[]` only after paginating the full 40-page cap, since `{}` never sets `results.count` and `total` stays `Infinity`. The comment says "no infinite pagination" while the page cap is what actually stops it. A one-line page-count assertion, or a `count: 0` in the payload, would prove the real termination reason and drop 39 needless iterations. Fine to land as-is.

santifer:
Clean rebase through a conflict I caused, and the resolution is the right one: main's `withRetry(request, ctx, policy)` signature kept, your non-Error fix folded on top rather than either side winning wholesale.

That fix is worth more than its four lines and I verified it by running it, not by reading it. With a transport that rejects with a string, `main` answers `TypeError: Cannot create property 'attempts' on string 'boom'`, so the real failure is destroyed inside the catch block before any caller sees it and the operator debugs the wrong thing. Your branch hands back the original rejection intact.

The auto-resolve is the other half: pointing at `network.id` in `__NEXT_DATA__` means an entry needs only its `careers_url`, and the failure path names `getro_collection: N` as the escape hatch instead of leaving someone guessing. Suite on a preview against today's main: 3756 passed, 0 failed. @Scott-Emberson already signed off the test file as code owner.

Merged. 🚀

PR #2641, opened 2026-08-09
fix: render numeric year/date scalars in the CV instead of dropping them

Scott-Emberson:
## The bug

A CV payload with a numeric scalar (`year: 2024` instead of `"2024"`) renders with the value missing. `escapeHtml()` in `build-cv-html.mjs` returns `''` for any non-string, so a bare number hits the typeof guard and comes out as an empty `edu-year` / `cert-year` / `job-period`. The `present` flag stays true, so the block still renders, and the CV ships with graduation years, certification years and employment dates blank while the builder exits 0 with `"valid": true`. The unresolved-placeholder guard can't catch this one: the placeholder was replaced, just with an empty string.

Hand-written YAML rarely triggers it because people quote their dates without thinking. Anything machine-authored does: a YAML emitter round-tripping the payload, or an agent writing the file, will produce `year: 2024` unquoted, and YAML parses that as a number.

## The fix

Coerce scalars through `String()` and blank out only `null`/`undefined`/objects, so numbers and booleans render their value. Two-line change in the escape path.

Regression test drives the real builder with numeric year/date fields and asserts the values appear in the rendered HTML. Fails on main, passes here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Numeric and boolean values now render correctly in generated CVs instead of being omitted.
  * Null, undefined, and unsupported object values continue to be excluded from output.

* **Tests**
  * Added end-to-end coverage for numeric dates and years in experience, education, and certification sections.

### DIFF
--- a/build-cv-html.mjs
+++ b/build-cv-html.mjs
@@ -64,8 +64,12 @@ const DEFAULT_SECTION_TITLES = {
 // (e.g. "R&D", "scaled 10x < budget", 'the "north star" metric') render as
 // literal text instead of breaking the document or injecting tags.
 function escapeHtml(text) {
-  if (typeof text !== 'string') return '';
-  return text
+  // Blank out only truly absent/structural values. A number or boolean scalar
+  // (e.g. a payload with `year: 2024` instead of `"2024"`) must render its value,
+  // not vanish: the old `typeof text !== 'string' → ''` guard silently dropped
+  // numeric years/dates from the CV while `present` stayed true.
+  if (text === null || text === undefined || typeof text === 'object') return '';
+  return String(text)
     .replace(/&/g, '&amp;')
     .replace(/</g, '&lt;')
     .replace(/>/g, '&gt;')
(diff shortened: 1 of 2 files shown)


santifer:

Sharp bug, well told: escapeHtml() in build-cv-html.mjs returns an empty string for any non-string, so an unquoted year: 2024 renders a blank edu-year, cert-year or job-period while the present flag stays true, and the builder exits 0 with valid: true. You explain why the unresolved-placeholder guard cannot catch it: the placeholder WAS substituted, just with an empty string. And why almost nobody hits it by hand: a person writing YAML quotes the dates without thinking, a machine or an agent emitting it does not.

santifer:
Ran your own test against `main` before reading the fix, which is the cheapest way to find out whether a guard is describing a real bug:

```
❌ edu-year:   expected "2019", got ""
❌ cert-year:  expected "2025", got ""
❌ job-period: expected "2024", got ""
```

Three years render as empty strings and the document still claims `present: true`, so the CV comes out missing dates while nothing anywhere reports a problem. On your branch all three render. That is the failure mode this repo is most exposed to and least able to see: not a crash, an absence that looks like a choice.

`typeof text !== 'string' → ''` is exactly the shape of guard that reads as defensive and is actually destructive, and narrowing it to `null | undefined | object` keeps the real protection (a structural value must not stringify to `[object Object]`) while letting scalars through.

Merged. 🚀

PR #2642, opened 2026-08-09
ci: nightly + release-tag upgrade qualification matrix (3 OS, 5 versions, rollback)

Scott-Emberson:
Layer 2 of #2007, the release-cadence matrix I said I'd file once #2358 landed.

## What it adds

A new workflow, `upgrade-matrix.yml`, running nightly (03:17), on `career-ops-v*` tag pushes, and on dispatch, across ubuntu, macos and windows with `fail-fast: false`, since a platform-specific failure is exactly what it exists to surface. Each run takes the last 5 release tags that are ancestors of HEAD and upgrades each one to the current commit through that release's own frozen updater shim, plus a rollback leg on the oldest tag asserting user files come through byte-identical. This is deliberately not a PR gate: a quirk in a frozen old shim should never block unrelated PRs. The `upgrade-gate` check from #2358 stays the only required one.

In `upgrade-tests.mjs`: the per-leg scaffolding the gate, matrix and rollback legs all share (redirect config, old-tag install, seeded fixture) moves into a `setupLeg()` helper, and `--all-versions` drives the tag loop. The canary path still mutates the mirror between `buildMirror` and setup, which is why the caller keeps ownership of both.

Rebase note: #2358 merged with review fixes (the guarded `SYSTEM_PATHS` extraction and the oracle restricted to system files). This branch is rebased on top and keeps those semantics unchanged; `setupLeg()` is structure only, no behavior change to the gate leg.

## Local runs

- `--pr-gate` GREEN (v1.25.0 to HEAD, 23 assertions)
- `--canary` GREEN (planted user-file clobber detected)
- `--all-versions` GREEN: v1.21.0 through v1.25.0 each through their own shim, plus rollback on v1.21.0
- `test-all.mjs`: 3137 passed, 0 failed

## The three questions from #2007, still open

Restating them here since this PR is where two of them become concrete:

1. Non-vacuity oracle: this stays with the changed-blob compare from #2358. If you'd rather have a planted version sentinel the updater must rewrite, it slots into `pickOracle` without touching the legs.
2. Support window: `--all-versions` takes `.slice(-5)` of ancestor tags. If you want a policy instead of a number (last N minors, or 12 months), that one line is the knob.
3. Nightly failure routing: right now a red nightly is just a failed run and a badge. Auto-filing an issue is a small follow-up if you'd rather be paged that way; I left it out to keep this reviewable.

## Summary by CodeRabbit

* **Testing**
  * Added automated upgrade qualification across Ubuntu, macOS, and Windows.
  * Added support for testing upgrades across the five newest versions.
  * Added rollback verification, including version restoration and preservation of user files.
  * Documented the option to run upgrade tests across all supported versions.

* **Chores**
  * Added nightly, version-tag, and manual test runs for upgrade validation.

### DIFF
--- a/.github/workflows/upgrade-matrix.yml
+++ b/.github/workflows/upgrade-matrix.yml
@@ -7,6 +7,10 @@ on:
     tags: ['career-ops-v*']
   workflow_dispatch:
 
+# The harness only ever reads the repo and writes to throwaway temp mirrors.
+permissions:
+  contents: read
+
 jobs:
   qualify:
     strategy:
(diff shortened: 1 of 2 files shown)


santifer:

Adds upgrade-matrix.yml: nightly at 03:17, on career-ops-v* tag pushes and on demand, across ubuntu, macos and windows with fail-fast:false, since a platform-specific failure is exactly what it exists to surface. Each run takes the last 5 release tags that are ancestors of HEAD and upgrades each one to the current commit using the frozen update shim FROM THAT RELEASE, plus a rollback leg on the oldest tag verifying that user files come back byte for byte identical. You deliberately keep it out of the required checks: an old shim's quirk should not block someone else's PR, so the only required one stays upgrade-gate from #2358.

PR #1606, opened 2026-07-06
feat(scan): persist per-run counters to data/scan-runs.tsv (#1604, PR-2 of 2)

Scott-Emberson:
PR-2 of #1604 — the **data-contract change**, split out per the issue so the read-only aggregator (#1605) isn't blocked on this schema discussion. **Stacked on #1605** (contains its commits; review from `feat(scan): persist per-run counters` only).

## What

Every non-dry `scan.mjs` run appends one row of its summary counters to a new user-layer **`data/scan-runs.tsv`** — today those numbers are printed once and lost when the terminal scrolls:

```
timestamp	status	companies	boards	found	filtered_title	…	dupes	new_added	errors
2026-07-03T14:02:11Z	completed	45	3	120	40	…	38	8	0
```

Schema commitments (the reason this is its own PR):

- **Full ISO timestamp** — two scans in one day never collapse.
- **`status` reserved now, written `completed` always in v1** — a follow-up wires failure-path writes so trend stats can exclude survivorship bias. Schema is forever; writes can wait.
- **Consumers parse by header name, never position** — columns may be appended later. `computeRunStats` in `stats.mjs` (the `runs` section, null until the file exists) does exactly that, skips torn rows from a crash mid-append, and excludes failed runs from averages.
- `--dry-run` leaves no trace (guarded like every other scan write; verified).
- New user-layer file → `.gitignore` + `DATA_CONTRACT.md` + Main Files entries included.

## Note for merge order

`scan.mjs` is also touched by #1601 (fingerprinting) in a different region — whichever merges second gets a trivial rebase.

## Tests

3 new assertions: appender writes the header once + one row per run; `computeRunStats` header-name aggregation with torn-row skip and failed-run exclusion (fixture arithmetic: 148/240 = 61.7% filter removal); null for empty/unknown-schema files. Full suite: **1242 passed / 0 failed**; SYSTEM_PATHS coverage validator green.

Closes #1604 together with #1605.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added scan-run persistence to a new TSV history file for real scans only (no trace on dry-runs).
  * Extended the human-readable summary with lifetime run trends, per-run averages, last run date, failed-run counts, and filter-removal percentage.
* **Bug Fixes**
  * Improved run-stats aggregation by safely handling empty/unknown schemas, skipping torn/invalid rows, and excluding failed runs from average calculations.
* **Documentation**
[...]

PR #2645, opened 2026-08-09
feat(scan): record failed runs in scan-runs.tsv so trends exclude survivorship bias

### LINKED ISSUE #2643: feat(scan): failure-path writes for scan-runs.tsv so run trends can exclude survivorship bias
Named as a follow-up in #1604 and in the code itself — the comment above `appendScanRunSummary` in `scan.mjs` says the `status` column is reserved, always `completed` in v1, "a follow-up wires failure-path writes so trend stats can exclude survivorship bias." This is that follow-up.

## The problem

A scan that dies mid-sweep (provider outage, rate-limit cascade, ctrl-C at company 40 of 45) writes no row at all. `data/scan-runs.tsv` only ever records the runs that finished, so everything downstream inherits survivorship bias:

- `stats.mjs` run trends (found/new per run) look healthier than reality, because the runs that found nothing before crashing aren't in the denominator.
- A streak of failing scans is invisible in the one file whose job is run history. You find out from the terminal, once, and then it scrolls away — the exact failure mode #1604 built this file to fix, except now only for successful runs.

## Proposal

- On abnormal exit, append a row with `status: failed` carrying the counters accumulated up to the failure point, with `errors` reflecting what was seen. The header already reserves the column and consumers parse by name, so this is not a schema change.
- Dry runs stay unwritten, same as today.
- `stats.mjs`: exclude non-`completed` rows from trend averages by default and surface a separate failure count (`runs.failed`), so a crash streak is visible instead of silently shrinking the sample. `--summary` gets one line when failures exist.
- Best effort on SIGINT: if the process can catch it, write the row; if not, that run is lost like today. The guarantee worth having is "an error path that unwinds normally writes its row", not "every possible death is recorded".

Small change, two files, tests alongside. Happy to pick it up unless someone else wants it.

Scott-Emberson:
Closes #2643 — the follow-up the `status` column was reserved for in #1606.

## What changes

`scan.mjs` arms a failure snapshot at the moment the sweep's counters come into scope, and only there: a `--dry-run` never arms it, and a config error before the sweep isn't a run, so neither leaves a trace. From that point on, a run that dies writes one row with `status: failed` carrying whatever the counters had accumulated. Two signals can trigger it, the fatal `main().catch` and a SIGINT handler (ctrl-C at company 40 of 45 being the common abort), and the snapshot is consumed on first use so they can never double-write. Everything in the write path is swallowed on purpose: failing to record the failure must not mask the original error. `new_added` is a hardcoded 0 on a failed row even when the sweep added postings before dying (the count isn't settled mid-sweep) — harmless to trends since failed rows are excluded from averages, but a raw-TSV reader should read that 0 as a sentinel, not a true count.

On the success path the snapshot is disarmed right after the `completed` row is appended, same guard as before.

`stats.mjs` already excluded failed rows from trend averages and reported `failedRuns`; the one change is selection by `status === 'completed'` instead of exclusion by the known failure name. This is a **no-op on what scan.mjs writes today** (only `completed`/`failed`, which both predicates agree on) — it's a guard so any status a *future* scan writes is excluded from trends until the aggregator learns what it means, not a behavior change now. Rows from pre-status files still default to `completed`, so old data keeps counting. One edge reachable only by hand-editing the TSV: an explicitly empty status flips from counted to `failedRuns`, since `appendScanRunSummary` always writes a non-empty status.

No schema change — the header reserved the column, and consumers parse by name.

## Coverage (honest limits)

Only two death paths write a `failed` row: the fatal `main().catch` and the `SIGINT` handler. **SIGTERM and `unhandledRejection` leave no row** — a `kill` or an unhandled async rejection still vanishes from the trend, same as before this PR. This narrows the survivorship gap for the common cases (crash, ctrl-C); it does not close it entirely.

## Tests

`tests/scan-run-failure.test.mjs`: failed row carries accumulated counters; snapshot consumed on first use (no double-write); no-op when unarmed; disarm on success; a throwing snapshot is swallowed; `computeRunStats` counts `failed` and any unknown status in `failedRuns` while averaging over completed rows only.

Suite: 3654 passed, 0 failed.

## Summary by CodeRabbit

* **New Features**
  * Scan runs now preserve failure records when interrupted or terminated by a fatal error.
  * Failure records are written only once and do not affect successful or dry-run scans.

* **Bug Fixes**
  * Run statistics now calculate averages using completed runs only.
  * Failed and other incomplete runs are excluded from completion metrics, while legacy records remain supported.

### DIFF
--- a/scan.mjs
+++ b/scan.mjs
@@ -2357,8 +2357,10 @@ async function main() {
   const emptyTargets = [];
 
   // Arm the failure-path row (#2643) now that the sweep is about to start and
-  // every counter it reads is in scope. new_added stays 0 on a failed run:
-  // the sweep hadn't reached the point where that count is final.
+  // every counter it reads is in scope. new_added is hardcoded 0 on a failed
+  // run even if the sweep added postings before dying (the count isn't settled
+  // mid-sweep). Excluded from trend averages so it can't skew them, but a
+  // raw-TSV reader should treat that 0 as a sentinel, not a true count.
   if (!dryRun) {
     registerRunFailureSnapshot(() => ({
       timestamp: new Date().toISOString(),
--- a/scan.mjs
+++ b/scan.mjs
@@ -1887,11 +1887,36 @@ const SCAN_RUNS_PATH = 'data/scan-runs.tsv';
 
 // One row of run counters per non-dry scan — today these numbers are printed
 // once in the summary and lost when the terminal scrolls. Full ISO timestamp
-// (two scans in one day must not collapse). `status` is reserved: always
-// 'completed' in v1; a follow-up wires failure-path writes so trend stats can
-// exclude survivorship bias. Consumers MUST parse by header name, never by
-// position — columns may be appended in later versions.
+// (two scans in one day must not collapse). `status` is 'completed' for a
+// finished run; a run that dies after the sweep starts records 'failed' via
+// writeRunFailureRow (#2643) so trend stats can exclude survivorship bias.
+// Consumers MUST parse by header name, never by position — columns may be
+// appended in later versions.
 export const SCAN_RUNS_HEADER = 'timestamp\tstatus\tcompanies\tboards\tfound\tfiltered_title\tfiltered_tier\tfiltered_location\tfiltered_posting_age\tfiltered_salary\tfiltered_content\tfiltered_cooldown\tdupes\tnew_added\terrors\tfiltered_blacklist\tfiltered_visa\tfiltered_posted_date\tfiltered_country_eligibility\n';
+
+// Failure-path writes (#2643). main() registers a snapshot closure once the
+// sweep's counters exist (never on --dry-run, never before the sweep starts —
+// a config error is not a run). The fatal catch and the SIGINT handler both
+// call writeRunFailureRow; the snapshot is consumed on first use so the two
+// signals can never double-write. Best-effort by design: a failure to record
+// the failure must not mask the original error, so everything is swallowed.
+let runFailureSnapshot = null;
+
+export function registerRunFailureSnapshot(fn) {
+  runFailureSnapshot = typeof fn === 'function' ? fn : null;
+}
+
+export function writeRunFailureRow(status = 'failed', filePath = SCAN_RUNS_PATH) {
+  const snapshot = runFailureSnapshot;
+  runFailureSnapshot = null;
+  if (!snapshot) return false;
+  try {
+    appendScanRunSummary({ ...snapshot(), status }, filePath);
+    return true;
+  } catch {
+    return false;
+  }
+}
 
 export function appendScanRunSummary(c, filePath = SCAN_RUNS_PATH) {
   if (!existsSync(filePath)) writeFileSync(filePath, SCAN_RUNS_HEADER, 'utf-8');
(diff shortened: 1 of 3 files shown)


santifer:

This implements the reserved status column follow-up, and the design is built not to pollute the file: the failure snapshot is only armed once the sweep counters come into scope, so a --dry-run never arms it and a config error before the sweep does not count as a run; both the fatal main().catch and a SIGINT handler fire it (the ctrl-C at company 40 of 45, which is the typical abort); and the snapshot is consumed on first use so the two paths cannot write two rows. The whole write path swallows its own errors on purpose, because failing to record the failure must not mask the original error. new_added stays 0 on failed rows because the sweep never closed that count.

santifer:
The bias this fixes is real, @Scott-Emberson: trends built only from runs that survived are trends about survivors, and a failed scan leaving no trace is exactly how that happens.

Flagging it as ⚠️ web-facing rather than blocking on substance: `data/scan-runs.tsv` is a data-layer file the web can consume, and this adds rows with a new status value. If the addition is purely additive (new rows, same column order, unknown statuses ignored by readers), it merges under the additive SLA once I have confirmed with the web side. Give me the confirmation and this moves.

Three measured observations, none blocking, worth a line each in the code or the body:

- The `stats.mjs` change (`r.status !== 'failed'` to `r.status === 'completed'`) is a **no-op on what scan.mjs writes today** and only bites on a future status: with an `aborted` row, base includes it in the averages (avgFound 51.5) and the patch excludes it (avgFound 100). Worth saying that out loud, since it reads like a behavior change and is really a guard.
- Same change flips a row with an **empty status** from completed to failedRuns. Only reachable by hand-editing the TSV, since `appendScanRunSummary` always writes an explicit status.
- `new_added` stays 0 on failed rows even when the sweep did add postings before dying (observed: pipeline got 3, the row records 0). It does not contaminate trends because failed rows are excluded from the averages, but it can mislead anyone reading the raw TSV.

And the honest coverage note for the body: SIGTERM and unhandledRejection leave no row; only SIGINT and the fatal catch do.

Scott-Emberson:
Confirmed additive, @santifer. The change adds **rows only**, never a column: `SCAN_RUNS_HEADER` is byte-identical to #1606's reserved schema, and `appendScanRunSummary` writes the same 19 fields in the same order for a `failed` row as for a `completed` one — only the `status` cell value differs. No reader has to change: `computeRunStats` parses by header name and already handles unknown statuses by excluding them from averages while counting them in `failedRuns`. A `failed` row is indistinguishable from a `completed` one to anything that doesn't inspect `status`. So: new rows, same column order, unknown statuses ignored — that's the whole footprint. Over to you for the web-side sign-off.

On the three observations — all landed as comment/body-only clarifications, pushed in `48a6221`:

- **stats.mjs guard is a no-op today.** Said out loud in the comment: `=== 'completed'` and `!== 'failed'` agree on the only two values scan.mjs writes; the switch only bites a future third status. Noted the hand-edit-only edge too — an explicitly empty status flips to `failedRuns`, unreachable via `appendScanRunSummary`.
- **new_added stays 0 on failed rows even when the sweep added postings.** You're right, the old comment implied the count was never reached — it's a hardcoded sentinel. Reworded in both the code and the PR body: harmless to trends (excluded from averages), but a raw-TSV reader shouldn't read it as a true count.
- **Coverage honesty.** Added to the body: only the fatal `main().catch` and the `SIGINT` handler write a row — **SIGTERM and `unhandledRejection` leave no trace.** This narrows the survivorship gap for the common cases (crash, ctrl-C), it doesn't close it.

santifer:
Another solid one, @Scott-Emberson. The part that makes this worth having is that the counters are the live ones: I ran your arm end-to-end with a `kill -INT` mid-sweep and the row came out `failed 5 0 42 ... 7 0`, i.e. the 42 found and 7 dupes it had actually seen at that instant, not a frozen snapshot. Same for a mid-sweep throw: exit 1 and the partial totals land. A trends file that only ever sees the runs that finished is a survivorship-biased file, and now it isn't one. Merged! 🚀

PR #1695, opened 2026-07-08
feat(ledger): set-status transition log + mode wiring (#1693, part 2)

Scott-Emberson:
Part 2 of #1693 — **STACKED on #1694, merge that first** (this branch contains its commits; the diff collapses to the wiring once part 1 lands).

## What

- **`set-status.mjs`** appends every real status change to `status-log.tsv` (sibling of the active tracker file, so `CAREER_OPS_TRACKER` redirects — tests, custom layouts — keep the ledger next to the tracker it describes): `{num}\t{date}\t{from}\t{to}\tset-status\t`. Runs inside the existing lock window after `writeFileAtomic` succeeds. Append failure = stderr warning, exit code unchanged — the tracker stays the source of truth for *state*, the ledger records *when*. JSON output carries `statusLogged`.
- **`--on YYYY-MM-DD`** — records the real event date ("they replied Tuesday"), validated (format, real calendar date, not future) before anything touches the tracker; refuses to consume a following flag as its value, same guard as `--note`.
- **Mode wiring:** `patterns` gains the optional Funnel Calibration lens with hard presentation rules (selection-bias note, one-action-on-below-range, year+directional, no multiplier claims under n=20, censored counts always repeated); `tracker` routes status changes through set-status (with `--on`) so transitions actually reach the ledger, and offers `funnel-velocity.mjs --summary` in stats; `followup` treats beyond-window waiting rows as follow-up candidates, not verdicts; `apply` Step 9 documents `--on` for backdated submissions.
- **test-all:** four integrity checks pinning the wiring; set-status-tests gains 10 ledger/`--on` cases (append on real transition, silence on no-op and `--dry-run`, sandbox-safe sibling path, backdating, all `--on` rejections, append-failure contract).

## Tests

- `node set-status-tests.mjs` — 42 passed, 0 failed
- `node test-all.mjs` — 1539 passed, 0 failed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added a new `funnel-velocity` CLI for funnel calibration, stage-to-stage velocity, and in-flight “waiting” analysis using benchmark ranges.
  * Added `--on YYYY-MM-DD` to the status update command to record the effective event date.
  * Added a default benchmark template for market calibration.
* **Bug Fixes**
  * Improved `--on` date validation and kept status updates working even if the status transition log can’t be written.
* **Documentation**
[...]

PR #2646, opened 2026-08-09
fix(company-history): record response dates via --on, not --note free text

Scott-Emberson:
Small follow-up promised in the #1712 cycle, unblocked when #1695 landed.

`company-history.mjs` was written while `set-status.mjs --on` was still in review, so its clearInstruction tells the user to record a company's response as `--note "responded <date>"`. That puts the one date this module actually wants into free text nothing parses — while `--on` writes it to `data/status-log.tsv`, the exact file company-history and funnel-velocity read applied/response dates back from. The aged-Applied summary line had the same stale advice.

Both instructions now point at `--on <response-date>`. The self-test and the test file pin the instruction to `--on`, so a future rewording can't quietly reintroduce the free-text workaround.

Two files, +9/-5. Suite: 3137 passed, 0 failed; company-history self-test 46/0.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved company history updates by recording response dates in a dedicated date field.
  * Ensured updates target the correct history entry.
  * Prevented response dates from being stored as workaround text in notes.

* **Documentation**
  * Updated guidance to reflect the improved response-date recording process.

* **Tests**
  * Added coverage for correct entry selection, date recording, and clean notes.

### DIFF
--- a/company-history.mjs
+++ b/company-history.mjs
@@ -347,9 +347,9 @@ export function computeResponsiveness(rows, followupCountsByAppNum, opts = {}) {
         dateBasis,
         // Real set-status.mjs syntax (it rejects unknown flags, so the
         // instruction must only use flags that exist): --on records the real
-        // response date in the status ledger — the same file this module and
-        // funnel-velocity.mjs read dates back from. A date buried in --note
-        // free text is parsed by nothing.
+        // response date in the status ledger — the file funnel-velocity.mjs
+        // reads dates back from. A date buried in --note free text is parsed
+        // by nothing.
         clearInstruction: `if they actually responded, node set-status.mjs ${row.num} <state> --on <response-date> clears this`,
       });
     } else if (RESPONDED_STATUSES.has(normalized)) {
--- a/company-history.mjs
+++ b/company-history.mjs
@@ -350,7 +350,7 @@ export function computeResponsiveness(rows, followupCountsByAppNum, opts = {}) {
         // response date in the status ledger — the file funnel-velocity.mjs
         // reads dates back from. A date buried in --note free text is parsed
         // by nothing.
-        clearInstruction: `if they actually responded, node set-status.mjs ${row.num} <state> --on <response-date> clears this`,
+        clearInstruction: `if they actually responded, node set-status.mjs --row ${row.num} <state> --on <response-date> clears this`,
       });
     } else if (RESPONDED_STATUSES.has(normalized)) {
       // row.date is the EVALUATION date, not the date the company replied. Use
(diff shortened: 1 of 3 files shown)


santifer:

You caught a stale instruction with its historical reason attached: company-history.mjs was written while set-status.mjs --on was still in review, so its clearInstruction still tells the user to record the company's reply as --note 'responded <date>', free text that nothing parses, when --on writes it into data/status-log.tsv, which is exactly the file company-history and funnel-velocity later read application and response dates from. The same stale recommendation was sitting in the aged-Applied summary line. You also pin the instruction in the self-test and in the test file, so a future rewrite cannot reintroduce the free-text workaround.

santifer:
Direction is right, @Scott-Emberson, and the diff is exactly as surgical as it should be for an instruction change: two files, +9/-5, no parser touched, both suite anchors intact.

Two things before it lands, both one-liners.

**1. `modes/patterns.md:107` still teaches the old form.** It keeps `node set-status.mjs <num> <state> --note "responded <date>"`, which is the very habit this PR is retiring. The two anchors `test-all.mjs:3288-3296` verifies ("Hygiene first, always." before "aged-Applied") do not move, so swapping that one flag does not touch CI. It is `modes/*.md` (🟡), but a single-token flag change consistent with the rest of the PR is the right scope for it.

**2. The new comment at `company-history.mjs:349-352` says something the file itself contradicts.** It claims `--on` writes to the status ledger *"the same file this module and funnel-velocity.mjs read dates back from"*. Measured with `status-log.tsv` present: `company-history` reports `sources.statusLog=false`, because `funnel-velocity.mjs` exports neither `getAppliedDateObservations` nor `computeMedianResponseDays`, and `company-history.mjs:248-256` documents that the source stays inert in that case. So line 249 says it does not read the ledger and line 350 says it does, twenty lines apart. `funnel-velocity` genuinely does read it; this module does not.

Worth fixing rather than waving through, because a comment that promises more than the code does is the kind that stops the next reader from checking. I hit the exact same shape today in `verify-pipeline.mjs`, where a note promised two checks "can never disagree" and they demonstrably do.

Reword to name only `funnel-velocity` as the consumer and this is good to go.

Scott-Emberson:
Both fixed in `cd8d437`.

**1. `modes/patterns.md:107`** — swapped `--note "responded <date>"` for `--on <response-date>`. Both anchors `test-all.mjs:3288-3296` verifies ("Hygiene first, always.", "aged-Applied") stay on the line, so CI is untouched.

**2. `company-history.mjs:350-351`** — reworded to name only `funnel-velocity` as the ledger consumer. You're right that this module reports the source inert (`statusLog=false`) since `funnel-velocity` exports neither helper — the comment now matches line 249 instead of contradicting it twenty lines down.

`node test-all.mjs` → 3137 passed, 0 failed.

santifer:
Good catch, @Scott-Emberson, and the fix has teeth. I ran both forms against a sandbox tracker: with the documented `--on <response-date>` the ledger records `42  2026-08-10  Applied  Responded`, the real event date; with the old `--note "responded 2026-08-10"` it records `2026-08-12`, the date somebody happened to type it. That is a silently wrong number in every latency stat downstream. Merged! 🚀

One word I adjusted in a follow-up commit rather than sending you round a three-platform CI run for it: the new comment said "A date buried in --note free text is parsed by nothing", and the same file parses exactly that 47 lines up (`parseAppliedDate(row.notes)`, whose result surfaces as `dateBasis: "notes"`). Your intent was right and narrower than the sentence: what nothing parses is a *response* date in a note. Now it says that.

PR #2647, opened 2026-08-09
fix: correct eval-runner output so evaluations are usable

Scott-Emberson:
Two independent defects in the non-Claude eval engines. Each silently produces wrong or unmergeable output, so the runs look like they worked.

## openrouter-runner: every evaluation skipped at merge time

The tracker-addition TSV was written with a `num\tdate\t…` header line ahead of the data row. `merge-tracker.mjs` reads the whole addition file as one record, no line splitting, so the header made `parts[4]`/`parts[5]` the literal strings `status`/`score`, `resolveScoreStatus` returned null, and every openrouter evaluation was skipped with "cannot tell score from status". AGENTS.md specifies the file as a single 9-column data line; now only that line is written.

## ollama-eval: options never reached the model

The request went to the OpenAI-compatible `/v1/chat/completions` route, which ignores Ollama's `options` object and has no `num_ctx` equivalent. Neither `temperature: 0.4` nor `num_ctx: 32768` ever reached the model, so every eval ran at the server default temperature with a 2048-token context that silently truncated the prompt — long JDs lost their tail and the eval scored a fragment. The call now targets the native `/api/chat` endpoint (where `options` is honored), parses its `{message:{content}}` response shape, and maps `prompt_eval_count`/`eval_count` through the shared usage normalizer.

## Tests

An openrouter TSV round-trips through the real `merge-tracker.mjs` (and the old header-prefixed form is shown to be skipped, pinning the failure mode); a mock Ollama server asserts the endpoint choice, the options payload, and native-response parsing.

Suite: 3144 passed, 0 failed.

### DIFF
--- a/ollama-eval.mjs
+++ b/ollama-eval.mjs
@@ -248,7 +248,7 @@ LEGITIMACY: <High Confidence | Proceed with Caution | Suspicious>
 // ---------------------------------------------------------------------------
 // Call Ollama
 // ---------------------------------------------------------------------------
-const endpoint = `${baseUrl}/v1/chat/completions`;
+const endpoint = `${baseUrl}/api/chat`;
 const timeoutMs = parseInt(process.env.OLLAMA_TIMEOUT_MS || '300000', 10);
 if (Number.isNaN(timeoutMs) || timeoutMs <= 0) {
   console.error(`❌  Invalid OLLAMA_TIMEOUT_MS: "${process.env.OLLAMA_TIMEOUT_MS}" — must be a positive integer (milliseconds).`);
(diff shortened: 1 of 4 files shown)


santifer:

Two independent failures that made non-Claude evaluations LOOK like they worked. One: openrouter-runner wrote the tracker-addition TSV with a header line in front, and since merge-tracker.mjs reads the whole file as a single record without splitting on lines, parts[4] and parts[5] were literally the strings 'status' and 'score', resolveScoreStatus returned null, and EVERY openrouter evaluation got dropped with 'cannot tell score from status'. Two, quieter: ollama-eval called /v1/chat/completions, which ignores Ollama's options object, so neither temperature 0.4 nor num_ctx 32768 reached the model and long JDs were truncated at the 2048 default, meaning the evaluation scored a fragment.

santifer:
Merged, @Scott-Emberson. Both of these were silent in the worst way: the run looked like it worked and the output was unusable or quietly degraded, which is the failure mode nobody reports because it does not look like a failure. Fixing the non-Claude engines matters more than it sounds: they are what people on a budget actually run. 🚀

PR #2649, opened 2026-08-09
fix(liveness): api.lever.co 404 is not authoritative for live-but-unlisted postings

### LINKED ISSUE #2648: bug(liveness): api.lever.co 404 is not authoritative — Confidential Postings stay live off the public API
### What happened?

\`liveness-api.mjs\`'s Lever provider treats a 404/410 from \`api.lever.co/v0/postings/{slug}/{id}\` the same way as Greenhouse and Workday: instant, confident \`expired\`. For Lever specifically that's wrong.

Lever has a documented Confidential/Internal Postings feature — postings can be excluded from the public \`v0/postings\` API while the direct \`jobs.lever.co/{slug}/{id}\` page keeps serving them normally to anonymous visitors.

### Repro (still live as of this writing)

```
$ curl -s -o /dev/null -w '%{http_code}\n' https://api.lever.co/v0/postings/SimbeRobotics/3da4ce33-555d-4c8b-87fd-e0fda8891028
404

$ curl -sL -o /dev/null -w '%{http_code}\n' https://jobs.lever.co/SimbeRobotics/3da4ce33-555d-4c8b-87fd-e0fda8891028
200
```

The real page renders the actual job title (`Simbe Robotics - Head of Demand Generation & ABM`) and a working Apply control — this is a genuinely open posting the API rung reports as removed.

### Scope check — is this a Greenhouse/Workday/Ashby problem too?

Tested rather than assumed. Sampled every provider's existing "ATS API 404" verdicts in a live tracker against the real rendered page (Playwright, not raw curl — job-boards.greenhouse.io is a client-rendered SPA, so a bare HTTP check is misleading there too):

| Provider | Sampled | Genuinely dead | False negatives |
|---|---|---|---|
| Lever | 6 | 4 | **2** |
| Greenhouse | 5 | 5 | 0 |
| Workday | 1 | 1 | 0 |
| Ashby | 1 | 1 | 0 |

Every Greenhouse/Workday/Ashby posting either redirected to the board with no trace of the role, or showed an explicit "page not found." Only Lever produced a live, job-specific page after its API said 404. This matches the product-level explanation above — Greenhouse/Workday/Ashby don't have an equivalent "excluded from the public feed but still directly servable" mechanism.

### Expected behavior

A Lever 404/410 should fall through to the Playwright check (same "let the browser decide" treatment already given to network errors, timeouts, and malformed Ashby board responses), not conclude \`expired\` outright. Greenhouse/Workday/Ashby's current 404-is-authoritative behavior should stay as-is — no evidence they share this failure mode, and adding an unneeded Playwright fallback there would just slow down a path that's already correct.

Same bug class as #1372 (Ashby), different mechanism: there the API rung was missing entirely; here it exists but isn't authoritative on 404 for this one provider.

I have a fix in progress (regression test + minimal change, scoped to the Lever provider only) and will open a PR against this shortly.

### CLI tool

Claude Code

### OS and Node.js version

macOS / Node 22

rubicon:
## Summary

Closes #2648. Lever's Confidential/Internal Postings feature can exclude a genuinely live posting from the public `v0/postings` API while `jobs.lever.co` keeps serving the direct page normally to anonymous visitors. The Lever provider in `liveness-api.mjs` treated a 404/410 there the same as Greenhouse and Workday — instant, confident `expired` — which is wrong for this one case.

Real-world repro (still live as of this writing): `api.lever.co/v0/postings/SimbeRobotics/3da4ce33-555d-4c8b-87fd-e0fda8891028` 404s, while `jobs.lever.co/SimbeRobotics/3da4ce33-555d-4c8b-87fd-e0fda8891028` returns 200 with the real title (`Simbe Robotics - Head of Demand Generation & ABM`) and a working Apply control. Same story for an Enable posting.

## Scope

Scoped to Lever only, not applied blanket to every per-job-API provider. Before writing this I sampled the other providers' existing "ATS API 404" verdicts in a live tracker against the real rendered page (Playwright, not raw curl — `job-boards.greenhouse.io` is a client-rendered SPA, so a bare HTTP status check would have been misleading there too):

| Provider | Sampled | Genuinely dead | False negatives |
|---|---|---|---|
| Lever | 6 | 4 | 2 |
| Greenhouse | 5 | 5 | 0 |
| Workday | 1 | 1 | 0 |
| Ashby | 1 | 1 | 0 |

Every non-Lever posting either redirected to the board with no trace of the role, or showed an explicit not-found page. None share Lever's public-API-excludes-a-still-live-posting behavior, so their existing 404-is-authoritative fast path stays untouched — adding an unneeded Playwright fallback there would only slow down something that's already correct.

## Fix

`api404Authoritative` (defaults to `true`) lets a provider opt out of treating its own 404/410 as proof of removal. Set to `false` on the Lever entry only. When false, `checkLivenessViaApi` returns `null` on 404/410 so the caller falls through to the Playwright check — the same inconclusive treatment already given to network errors, timeouts, and a malformed Ashby board response.

Same bug class as #1372 (Ashby false-expired), different mechanism: there the API rung was missing entirely; here it exists but isn't authoritative on 404 for this one provider.

## Test plan

- [x] New regression test in `test-all.mjs`: Lever 200 → active, 404/410 → `null` (inconclusive), asserted against a mocked `fetch` following the existing `checkLivenessViaApi` test pattern
- [x] Confirmed the new test fails against the pre-fix code (404/410 → `expired`) before applying the fix
- [x] `node test-all.mjs` — 3138 pass / 0 fail after the fix
- [x] Every pre-existing Lever-specific test in the suite still passes (provider detection, SSRF redirect guard, `prepare-application.mjs` field builder, `scan.md`/`verify-portals` slug parsing) — the change is additive, not a behavior regression elsewhere

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved job listing availability checks for providers whose APIs may return outdated 404 or 410 responses.
  * Listings are now verified through the public job page before being classified as expired when API results are inconclusive.

* **Tests**
  * Added regression coverage for active listings and temporary API removals.

### DIFF
--- a/liveness-api.mjs
+++ b/liveness-api.mjs
@@ -54,6 +54,9 @@ function isSafeValue(v) {
 //   `timeoutMs`  — override the default fetch timeout (slow/rate-limited APIs).
 //   `interpret`  — read the 200 response body to decide liveness (org-level APIs
 //                  where a 200 alone doesn't prove THIS posting is live).
+//   `api404Authoritative` — defaults to true (a 404/410 means gone). Set to
+//                  false when the provider's public API can 404 a posting that
+//                  is still genuinely live elsewhere (see the `lever` entry).
 const ATS_PROVIDERS = [
   {
     id: 'greenhouse',
@@ -75,6 +78,13 @@ const ATS_PROVIDERS = [
       return m ? { apiHost: `api.${host[1]}`, slug: m[1], id: m[2] } : null;
     },
     api: ({ apiHost, slug, id }) => `https://${apiHost}/v0/postings/${slug}/${id}`,
+    // Lever's Confidential/Internal Postings feature explicitly excludes some
+    // live postings from the public v0/postings API while the direct
+    // jobs.lever.co page keeps serving them normally. Real-world repro
+    // (2026-08-09): api.lever.co 404s two postings whose jobs.lever.co pages
+    // return 200 with the real job title and a working Apply control. A 404
+    // here is NOT proof of removal — fall through to Playwright instead.
+    api404Authoritative: false,
   },
   {
     id: 'ashby',
@@ -159,7 +169,7 @@ export function classifyAshbyBoard(json, jobId) {
  * Map a posting URL to its ATS API URL, or null if it isn't a known ATS posting
  * (or any extracted segment fails the strict charset). Pure + deterministic.
  * @param {string} rawUrl
- * @returns {{ ats: string, apiUrl: string, parts: Record<string, string>, timeoutMs?: number, interpret?: (res: Response, parts: Record<string, string>) => Promise<{ result: 'active' | 'expired', code: string, reason: string } | null> } | null}
+ * @returns {{ ats: string, apiUrl: string, parts: Record<string, string>, timeoutMs?: number, interpret?: (res: Response, parts: Record<string, string>) => Promise<{ result: 'active' | 'expired', code: string, reason: string } | null>, api404Authoritative: boolean } | null}
  */
 export function resolveAtsApi(rawUrl) {
   let u;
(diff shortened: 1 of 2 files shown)


santifer:
Thanks for this, @rubicon.

You fix exactly the bug you documented: the Lever provider in liveness-api.mjs treated a 404/410 from api.lever.co like Greenhouse or Workday, instant confident expired, when Lever can pull a posting out of the public API while jobs.lever.co keeps serving it to any anonymous visitor. You scope it to Lever ONLY, and you justify the scoping with data rather than opinion: 6 Lever verdicts sampled (2 false negatives), 5 Greenhouse, 1 Workday and 1 Ashby, no non-Lever failure, and Playwright rather than curl because a curl against job-boards.greenhouse.io would have been fooled the same way.

santifer:
Merged, @rubicon. A 404 from an API that never promised to be a source of truth about listing state is not evidence a job is gone, and marking a live posting dead is the expensive direction of that error. The per-provider flag with a safe default is the right shape: Lever opts out, everyone else keeps today's behavior. 🚀

Issue #2385, opened 2026-07-31
perf(analytics): report reading is 70%+ filesystem syscalls, with redundant existsSync probes and repeated parses


Scott-Emberson:
Profiling the analytics scripts against 500 and 2,000 synthetic reports, the runtime is almost entirely filesystem syscalls, and a good chunk of that is redundant.

CPU profile of `analyze-patterns` at 2,000 reports (1,320 ms sampled): `readFileUtf8` **693 ms (53%)**, `existsSync` **217 ms (16%)**, js-yaml 79 ms (6%), own regex and logic 148 ms. `upskill` at the same scale: `readFileUtf8` 593 ms (66%), `existsSync` 88 ms. So 69-76% of runtime is the filesystem.

Worth stating up front: **YAML parsing is not the problem.** All three report readers already regex the fence out before parsing and hand js-yaml only the ~916-byte Machine Summary block, which costs 14.1 ms for 500 blocks (28 µs each). A more targeted slice would save nothing.

## 1. Redundant existsSync probes

`analyze-patterns` stats each report **three times** (`:428`, `:658`, `:662`) before it ever reads it; `upskill:216` does up to two. Measured at 2,000 files:

| strategy | ms |
|---|---|
| existsSync x3 + readFileSync (current) | 555 |
| readFileSync only | 427 |
| `Promise.all(readFile)` | 175 |
| pooled concurrency 64 | 156 |

The guards buy nothing: `readFileSync` wrapped in try/catch costs the same as bare `readFileSync` (24.0 vs 24.4 ms per 500 files), so catching ENOENT is free. Dropping the probes and pooling the reads is 3.5x at 2,000 files.

One caveat on the async half: `analyze-patterns.mjs` and `upskill.mjs` both call `analyze()` at top level with no `isMain` guard, unlike `salary-gap.mjs:692`. Nothing imports them today so it is latent, but the guard belongs in the same change if the loop goes async.

## 2. Repeated parses within a single script

`stats.mjs` re-splits and re-parses `applications.md` **four times** per run (`:423`, `:435`, `:438`, `:466`, where `:466` recomputes exactly what `:438` already produced) and scans `scan-history.tsv` twice (`:424`, `:465`) even though `computeScanStats` already builds the identical company Set internally at `:204`/`:216` and then discards it. Parsing once and threading the rows through, plus returning that Set, is mechanical and worth roughly 30-50 ms.

[...]

PR #2655, opened 2026-08-10
perf(analytics): drop redundant existsSync probes before report reads

Scott-Emberson:
Splits out item 1 of #2385, per your comment there.

`analyze-patterns` stat-ed each report up to three times before reading it (once in `parseReport`, twice in the link-resolution block), and `upskill` up to twice. The probes buy nothing: `readFileSync` wrapped in try/catch costs the same as a bare read (24.0 vs 24.4 ms per 500 files), and stat-then-open is a TOCTOU race besides. Both scripts now attempt the read directly and treat `ENOENT`/`ENOTDIR` as "no report", via a small local `readTextIfExists` helper in each file. I kept the helper local rather than shared because a shared report module is item 3 territory.

The reads stay synchronous. The pooled async variant from the issue table is faster still, but it drags in the missing `isMain` guard question, so it stays out of this PR.

One edge-case change in `analyze-patterns`, made deliberately: each candidate path is now guarded to `reports/` individually before the read is attempted. Previously the tracker-relative path was selected by existence first and guarded after, so a legacy root-relative link stopped resolving whenever a shadow file happened to exist at the tracker-relative location; the row silently lost its report data. Now the root fallback still gets a chance as long as it passes the guard. Reading outside `reports/` remains impossible, since every candidate is checked before any read.

Verification:
- `node test-all.mjs`: 3137 passed, 0 failed (the one warning is the pre-existing Polish CV reorder fixture asserting its own warning).
- `--self-test` passes for both scripts.
- On a fixture exercising present, missing, legacy root-relative, and outside-`reports/` tracker links, output is byte-identical to `main` for both scripts (modulo `analysisDate`).

Numbers from #2385 at 2,000 reports: `existsSync` was 217 ms (16%) of `analyze-patterns`; dropping the probes takes the measured read loop from 555 ms to 427 ms.

Design comment on the read strategy (item 3) follows on the issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved report loading when files or directories are missing.
  * Added support for resolving reports from current and legacy locations.
  * Restricted report lookup to the designated reports area, blocking traversal and symlink escapes.
  * Report processing now stops safely after the first valid report is found.
  * Unexpected file access errors are preserved for clearer troubleshooting.

### DIFF
--- a/analyze-patterns.mjs
+++ b/analyze-patterns.mjs
@@ -400,6 +400,18 @@ risk_summary:
     failures.push('classifyRemote geo-restricted precedence regressed');
   }
 
+  // Reports-root containment: a legit link stays inside reports/, a crafted
+  // traversal link escapes root and must be rejected before parseReport. join()
+  // collapses '..' at the call site, so the candidate is already absolute here.
+  {
+    const legit = join(CAREER_OPS, 'reports', '042-acme-2026-01-01.md');
+    if (!withinReports(legit)) failures.push('containment: legit reports/ path wrongly rejected');
+    const escape = join(CAREER_OPS, 'reports/../../../etc/passwd');
+    if (withinReports(escape)) failures.push('containment: traversal path escaped reports/ (path-traversal guard broken)');
+    const sibling = join(CAREER_OPS, 'reports-evil', 'x.md');
+    if (withinReports(sibling)) failures.push('containment: reports-prefixed sibling dir wrongly accepted');
+  }
+
   if (failures.length > 0) {
     console.error(`analyze-patterns self-test failed: ${failures.join('; ')}`);
     process.exit(1);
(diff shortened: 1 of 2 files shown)


santifer:

You bring your own measurement instead of intuition: readFileSync wrapped in try/catch costs the same as stat-then-read (24.0 vs 24.4 ms per 500 files) and avoids the TOCTOU race, so it drops up to 3 stats per report in analyze-patterns and up to 2 in upskill. The part that actually needs review is the behaviour change you flag yourself: every candidate path is now validated against reports/ before the read is attempted, so a legacy root-relative link resolves again even when a shadow file exists at the tracker-relative path, the case where the row used to lose its report data silently.

santifer:
Merged, @Scott-Emberson: the readTextIfExists helper with ENOENT-to-null and rethrow-the-rest keeps the semantics honest, and the withinReports containment guard turned a perf cleanup into a small hardening win too. The #2385 index design has its green light: build when ready. 🚀

PR #2657, opened 2026-08-10
fix(providers): isRetryableError no longer misclassifies a refused redirect as retryable

ilyakanevskiy:
## Context

Spotted by @eliador90 in review on #2640: `isRetryableError()` misclassifies a refused redirect (`fetch(url, { redirect: 'error' })` meeting a 3xx — the mandatory SSRF guard from #1440) as a retryable network error, because both surface identically — a bare `TypeError` with no `.status`.

Reproduced all four failure shapes on Node v24.18.1 to confirm the fix is actually implementable and distinguishable:

| case             | `err.name`   | `err.cause?.message`       | `err.cause?.code`           |
| ---------------- | ------------ | -------------------------- | --------------------------- |
| redirect refusal | `TypeError`  | `"unexpected redirect"`    | `undefined`                 |
| DNS failure      | `TypeError`  | `"getaddrinfo ENOTFOUND…"` | `"ENOTFOUND"`               |
| abort/timeout    | `AbortError` | — (no `cause`)             | —                           |
| connect timeout  | `TypeError`  | `"Connect Timeout Error…"` | `"UND_ERR_CONNECT_TIMEOUT"` |

Redirect refusal is the only case with `cause.message === "unexpected redirect"` (exact string, no regex needed) and the only one where `cause.code` is absent entirely — cleanly distinguishable from the other three.

The bug is pre-existing in `_http.mjs`, not universal: only 5 of 72 provider modules retry on fetch failure at all — `getro.mjs` / `a16z-speedrun-talent.mjs` via the shared helper, `workday.mjs` / `oraclecloud.mjs` each carrying a private duplicate of the same retry logic, and `ashby.mjs` (its own thing, see below). The other 67 don't retry on a fetch failure, so a redirect refusal there just throws once — no wasted attempts, not affected.

## Summary

Since a refused redirect is deterministic and will never succeed on retry, any provider that both retries fetches and meets a redirecting host wasted `retries + 1` attempts plus backoff on a failure that could never resolve.

This PR:

1. Fixes the misclassification in `isRetryableError()`, distinguishing the refused-redirect shape via undici's `err.cause.message === 'unexpected redirect'` — not documented anywhere, so the exact string is pinned by a new test and now lives in a named constant (`REDIRECT_REFUSAL_CAUSE_MESSAGE`) with a comment on where it comes from and its Node-version floor.
2. Retires the private retry duplicates in `workday.mjs` and `oraclecloud.mjs`, converging both onto the shared `fetchJsonWithRetry` helper (via a `{ retries: 3 }` policy override, matching their prior cadence) instead of leaving two more copies that would each need the same fix independently.

`ashby.mjs` also carries a private retry duplicate (an even looser unconditional-retry variant) but is deliberately left untouched here — migrating it is a real behavior change (it currently retries even deterministic 4xx errors), not just deduplication, and is being split into its own PR.

## Changes

- **`providers/_http.mjs`**
  - `isRetryableError()`: a refused redirect (no `.status`, `err.cause.message === 'unexpected redirect'`) now returns `false` instead of falling through to the generic "no status → retryable" branch.
  - New `REDIRECT_REFUSAL_CAUSE_MESSAGE` constant carrying that undici string, with a comment explaining where it comes from (undici's `cause.message` on a refused redirect) and that it's present from Node 18.5 onward — on older Node, `cause` is simply `undefined`, so the check doesn't fire and the old (retryable) classification stands.
  - Updated doc comments on `RETRY_DEFAULTS` and `fetchJsonWithRetry` to reflect the current set of consumers (`a16z-speedrun-talent.mjs`, `workday.mjs`, `oraclecloud.mjs`) instead of describing workday/oraclecloud as not-yet-converged.

- **`providers/workday.mjs`**
  - Removed the private `parseRetryAfterMs`, `isRetryableError`, and `fetchPageWithRetry` helpers and the `MAX_RETRIES`/`RETRY_BASE_DELAY_MS`/`RETRY_MAX_DELAY_MS` constants.
  - Both call sites (first page, and inside the pagination loop) now call `fetchJsonWithRetry(ctx, ep.api, opts, RETRY_POLICY)` with `RETRY_POLICY = { retries: 3 }` — `baseDelayMs`/`maxDelayMs` are left unset to inherit the shared helper's defaults (500ms/8000ms), which already match what workday's private copy used.
  - The local `sleep()` helper is kept — it's still used for the independent inter-page delay (`INTER_PAGE_DELAY_MS`), unrelated to retry backoff.
  - All existing caller-level behavior around the retry calls is unchanged: first-page exhaustion still propagates uncaught, later-page exhaustion is still caught and tags `jobs.workdayTruncated = true` with the same warning shape.

- **`providers/oraclecloud.mjs`**
  - Same shape of migration: removed the private retry helpers, single call site now uses `fetchJsonWithRetry(ctx, apiUrl, opts, RETRY_POLICY)` with the same `{ retries: 3 }` override. No per-page try/catch here (unchanged) — exhaustion still propagates and aborts the whole tenant fetch, which is the existing, correct-as-is behavior.

- **`docs/SUPPORTED_JOB_BOARDS.md`**
  - Added the missing Oracle Recruiting Cloud (ORC) row (pre-existing gap — the provider existed but was never documented), in the same style as neighboring rows.

- **`tests/providers/_http.test.mjs`** (new)
  - Direct coverage of `isRetryableError()` and `fetchJsonWithRetry()` — previously only exercised indirectly through consumer providers' tests.
  - Status-based classification: 429 → true, 5xx → true, other 4xx → false, generic no-status network error → true.
  - The refused-redirect shape → false (the main fix), and the `cause === undefined` old-Node fallback → true.
  - An end-to-end case asserting `fetchJsonWithRetry` calls `ctx.fetchJson` exactly **once** on a redirect refusal, not `retries + 1` times.

- **`tests/providers/workday.test.mjs` / `tests/providers/oraclecloud.test.mjs`**
  - No changes needed — existing retry-related tests (429 recovery, Retry-After honoring/clamping, non-retryable-4xx short-circuit, retry-exhaustion truncation) all pass unchanged against the migrated call sites, since the `policy` override preserves the same attempt counts and delays.

## Verification

- `node test-all.mjs --only providers/_http`, `--only providers/workday`, `--only providers/oraclecloud` — all green, no changes to existing pass counts beyond the new `_http` suite.
- Full `node test-all.mjs` — 3142 passed, 0 failed (one pre-existing, unrelated warning: a Windows symlink-permission skip in a plugin test).
- Live sanity check against a real redirecting host (`jobs.jobvite.com/fieldcore`) confirms `isRetryableError()` returns `false` for the actual `TypeError`/`cause` shape Node produces, not just the hand-built mock in the unit test.

## Not in this PR

- `ashby.mjs`'s migration onto the shared helper, and the fix to its unconditional-retry behavior — separate PR, since it's a real behavior change rather than pure deduplication, and there's an open question of whether to also honor `Retry-After` there (needs its own investigation) and whether the jitter-spacing idea from its inline backoff formula should move into the shared helper.

## Summary by CodeRabbit

- **New Features**
  - Added Oracle Recruiting Cloud (ORC) to the supported job boards.
  - ORC job listings support automatic tenant URL detection, public requisition access, pagination, and a safety limit of approximately 5,000 jobs.

- **Bug Fixes**
  - Improved handling of temporary network and service failures across Oracle Cloud and Workday job retrieval.
  - Redirect refusals are now recognized as non-retryable errors and surfaced promptly.
  - Retry messages now accurately report the number of attempts made.

- **Tests**
  - Added coverage for retry decisions, network errors, redirect refusals, and retry attempt behavior.


ilyakanevskiy:
@coderabbitai review

ilyakanevskiy:
@coderabbitai review

eliador90:
Independent confirmation on a different Node minor, since the one caveat I raised on #2640 was whether `cause.message` is stable across undici versions. You tested v24.18.1; this is v24.13.0:

```
redirect refusal   name=TypeError      cause.message="unexpected redirect"                          cause.code=undefined
DNS failure        name=TypeError      cause.message="getaddrinfo ENOTFOUND …"                      cause.code="ENOTFOUND"
abort/timeout      name=TimeoutError   cause.message=undefined                                      cause.code=undefined
```

Identical to your table, so the exact-string match holds across at least those two. One trivial difference: with `AbortSignal.timeout()` the name comes through as `TimeoutError` rather than `AbortError` (an aborted controller gives `AbortError`). Both have no `cause`, so neither reaches the redirect branch and it does not affect the fix.

Matching on `cause.message === 'unexpected redirect'` exactly, rather than the regex I sketched, is the better call. My version would have matched a DNS failure for a host with "redirect" in its name.

Nice catch splitting `ashby` out. Its unconditional retry is a real behaviour change and does not belong in a classification fix.

santifer:
Good diagnosis, @ilyakanevskiy. A refused redirect is the SSRF guard doing its job — `redirect: 'error'` against a 3xx is a deliberate stop, not a transient failure — so classifying it as retryable meant burning the full retry budget on something that can never succeed, and then reporting a timeout-shaped error for what was actually a policy decision.

What made this easy to accept is that you didn't argue it from the code alone: reproducing all four error shapes on a real Node version, counting exactly how many providers are affected, and sanity-checking against a live board is the difference between a plausible fix and a verified one. Merged! 🚀

PR #2489, opened 2026-08-04
docs: add three shipped providers to SUPPORTED_JOB_BOARDS.md

ashudhanda:
## What does this PR do?

Adds the three missing provider rows — **Agentic Engineering Jobs (`agentic-jobs`)**, **Oracle Recruiting Cloud (`oraclecloud`)**, and **Welcome to the Jungle (`wttj`)** — to `docs/SUPPORTED_JOB_BOARDS.md` so the table matches the shipped modules in `providers/`. Coverage, auth model, and config notes are taken from each provider's source and tests.

> Note: this PR originally also added a Jobvite row. It was dropped per review — Jobvite's JSON API is dead board-wide, and #2529 carries the replacement.

## Related issue

Fixes #2478

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Documentation / translation
- [ ] Refactor (no behavior change)

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/santifer/career-ops/blob/main/CONTRIBUTING.md)
- [x] If this is a new feature or architecture change, I opened an issue first (bug fixes, providers, docs & translations are exempt — send those straight in)
- [x] My PR does not include personal data (CV, email, real names, scan results, or pipeline data)
- [x] I ran `node test-all.mjs` and all tests pass (docs-only change; no code touched)
- [x] My changes respect the [Data Contract](https://github.com/santifer/career-ops/blob/main/DATA_CONTRACT.md) (no modifications to user-layer files)
- [x] My changes align with the [project roadmap](https://github.com/santifer/career-ops/discussions/156)

## Summary by CodeRabbit

* **Documentation**
  * Expanded supported job board documentation to include Agentic Engineering Jobs, Oracle Recruiting Cloud, and Welcome to the Jungle.
  * Added details on APIs, configuration, authentication, pagination, and data normalization.

PR #2659, opened 2026-08-10
docs: add six missing providers to SUPPORTED_JOB_BOARDS

### LINKED ISSUE #2478: docs: SUPPORTED_JOB_BOARDS.md is missing four shipped providers (agentic-jobs, jobvite, oraclecloud, wttj)
**Time:** ~30 min · **Difficulty:** Beginner · **No-code**

## Context

`docs/SUPPORTED_JOB_BOARDS.md` has drifted behind `providers/`: four shipped providers (each with tests) have no row in the table — **agentic-jobs, jobvite, oraclecloud, wttj**.

Derivation (files starting with `_` are internal helpers, and `local-parser` is the local-files strategy, not a board):

```bash
for f in providers/*.mjs; do b=$(basename "$f" .mjs); grep -qi "$b" docs/SUPPORTED_JOB_BOARDS.md || echo "MISSING: $b"; done
```

## What to do

Add one row per missing board, matching the existing row format exactly (read 2–3 neighboring rows first for the column conventions: board name, what it covers, auth model, notes). The provider source and its test file tell you what each one actually does — document reality, not guesses.

## Acceptance

- The derivation command above reports only `_`-prefixed helpers and `local-parser`.
- Each new row's claims (coverage, auth) match what the provider source does.

github-pratik:
Closes #2478.

`docs/SUPPORTED_JOB_BOARDS.md` is meant to carry one row per non-helper module in `providers/` — the file says so itself, and its footer asks that new providers update the table in the same PR. Several shipped providers had no row.

The issue names four: `agentic-jobs`, `jobvite`, `oraclecloud`, `wttj`. I checked every non-helper module under `providers/` against the table rather than just the four, which turned up three more with no entry: `consider`, `getro`, `joinup`.

That was seven rows when this PR opened. `main` has since grown its own Oracle Recruiting Cloud row — and a better one, since it covers the numbered tenant apexes from #2683 that mine predated — so ORC is dropped here and six remain:

| Row | Type |
| --- | --- |
| Agentic Engineering Jobs | API |
| Consider | API |
| Getro | API |
| Jobvite | API |
| joinup.ch | Parser |
| Welcome to the Jungle | API |

Each note is drawn from that provider's own module header — endpoint, auto-detection or the config it requires, pagination bounds, and any per-board caveat (Consider and Getro both need an explicit board/collection id; WTTJ throws without a `queries:` block; Getro's two knobs carry defaults of 40 pages and 90 days).

After this, `local-parser` is the only provider filename with no literal match in the table — it is already documented as the "Local parser" row.

Docs-only; no code paths touched.

**Follow-up worth its own issue:** nothing enforces provider/table parity, so this gap can silently reappear the next time a provider lands without its row. A small coverage check in the same spirit as the existing `validate-*-coverage.mjs` guards would close it. Happy to open an issue and take it if that sounds useful.

## Summary by CodeRabbit

* **Documentation**
  * Added support documentation for seven job boards, including access methods, endpoints, configuration requirements, pagination, filtering, and provider-specific constraints.


ilyakanevskiy:
Hey @github-pratik, I love keeping things tidy too — appreciate the cleanup here!

Getro and Jobvite are already covered by #2640 and #2623 respectively, and Oracle Recruiting Cloud (ORC) is covered by #2657 — better to exclude all three from this PR to avoid duplication/conflicts.

github-pratik:
Both review points were valid — fixed in 73ffdd8.

**`max_pages` on agentic-jobs** (@copilot) — confirmed against the source rather than assumed. `providers/agentic-jobs.mjs` reads no entry config at all; pagination is bounded by its own `MAX_PAGES = 40` / `MAX_JOBS = 2000` constants. Contrast `providers/4dayweek.mjs`, which does read `entry?.max_pages`. Documenting it as configurable would have had users setting a key the provider ignores, so this was a real error on my part — the row now states the fixed 40-page / 2000-posting cap explicitly.

**Jobvite sentence** (@coderabbitai) — taken as suggested; the final sentence now has a subject.

One thing worth flagging that came out of the CodeRabbit walkthrough, since it affects sequencing rather than this diff:

- #2529 replaces the Jobvite JSON API with a careers-page HTML scrape. If that lands, the Jobvite row here (which documents the current `jobs.jobvite.com/api/company/{id}/jobs` behaviour) goes stale on merge. Happy to drop the Jobvite row from this PR and let #2529 carry its own row per the table's "update this table in the same PR" rule — whichever order you prefer.
- #2485 and #2489 overlap with parts of this PR. Between the three, `consider`, `getro` and `joinup` appear only here. I'm glad to trim this down to just those three if you'd rather land the earlier PRs first — just say which shape is least work for you to review.

github-pratik:
Rebased onto latest `main` — #2527 added the `join.com` row in the same part of the table, which put this branch into conflict.

Resolved by keeping all three rows in the table's existing order: `Jobvite` → `join.com` → `joinup.ch`. Still 7 insertions, no other rows touched.

santifer:
Thanks for this, @github-pratik.

Closes #2478 but doesn't stop at the 4 providers the issue named: every non-helper module in providers/ was cross-checked against the table, turning up 3 more with no row (consider, getro, joinup). Each note comes from the module's own header (board id required for Consider and Getro, WTTJ breaking without a queries: block, ORC needing expand) and keeps the existing alphabetical order. It also names the root cause: nothing enforces provider/table parity, with an offer to add a check in the style of the validate-*-coverage.mjs ones.

github-pratik:
Rebased onto today's `main` and addressed the outstanding review comment. Now **six** rows, not seven — title updated to match.

**Oracle Recruiting Cloud dropped.** `main` has since grown its own ORC row, and the version there is better than mine: it covers the numbered tenant apexes from #2683 (`oraclecloud1-99.com`), which my row predated and did not mention. Kept upstream's, dropped mine.

**`getro_max_age_days` contract** (@coderabbitai) — valid, fixed. Checking the provider turned up a second undocumented knob alongside it, so the row now states both: `getro_max_pages` (`DEFAULT_MAX_PAGES = 40`, clamped by `HARD_MAX_PAGES = 200`) and `getro_max_age_days` (`DEFAULT_MAX_AGE_DAYS = 90`), where `0` disables the age cutoff and absent or invalid values fall back to the default. That matches how `templates/portals.example.yml` already documents them.

Also folded in upstream's retyping of `Personio` (RSS → Parser) and `Workable` (Parser → API) rather than reverting them.

Re-verified after the rebase: every non-helper module under `providers/` now has a row (`local-parser` is the one filename with no literal match — it is the "Local parser" row), no row's column count changed, and the two rows with irregular counts are the pre-existing ones carrying escaped pipes (`getManfred`, `Hacker News`). Branch is level with `main`, diff is 6 insertions and nothing else.

Remaining six: Agentic Engineering Jobs, Consider, Getro, Jobvite, joinup.ch, Welcome to the Jungle. Of those, `consider`, `getro` and `joinup` are still the three that appear in no other open PR.

santifer:
Heads-up before you spend time on it, @github-pratik: this went into conflict with `main`, and it is my fault rather than yours. Another docs PR touching `SUPPORTED_JOB_BOARDS.md` landed today (alphabetical ordering), plus a run of provider merges, so the file moved under you.

Rebase onto today's main and ping me here: it goes straight back into the batch. Your six rows are still the right six, and the table is in better shape to receive them now than it was when you opened this.

ilyakanevskiy:
As of current `main`, all six rows this PR set out to add are already there:

- **Agentic Engineering Jobs** — present, same coverage (API, pagination, rate limit, location fallback, job URL shape) as this PR's version.
- **Consider** — present, same coverage (board-id requirement, `consider_board:`/`consider_size`, ATS-link dedup).
- **Getro** — present, same coverage (`getro_collection`, `getro_max_pages`/`getro_max_age_days` with their defaults, newest-first pagination).
- **Jobvite** — present, but via a different mechanism than this PR documents: `main`'s row covers the `app.jobvite.com/CompanyJobs/Xml.aspx?c=<companyEId>` XML feed, not the `jobs.jobvite.com/api/company/{companyId}/jobs` JSON API this PR describes. The provider implementation changed in #2623, which landed after this PR was opened — it retired the JSON API in favor of the XML feed.
- **joinup.ch** — present, same coverage (`__NEXT_DATA__` scrape off the browse page, created-descending ordering, auto-detection by host).
- **Welcome to the Jungle** — present, same coverage (Algolia index, rotating app id/key via `/api/env`, required `queries:` block, `max_hits` default/cap).

There's nothing left in this diff that `main` doesn't already cover, so merging it would add no new content.

@github-pratik @santifer — can one of you close this PR without merging?

santifer:
Closing this one, @github-pratik, and the reason sits with us rather than with your work.

You did the job properly. You did not stop at the four providers #2478 named: you cross-checked every non-helper module against the table, found three more with no row, corrected the `max_pages` claim against the source instead of assuming it, dropped your ORC row when `main` grew a better one, and rebased twice when the file moved. Then I told you it would go straight back into the batch, and it did not.

As of current `main` all six rows are there: Agentic Engineering Jobs, Consider, Getro, Jobvite, joinup.ch and Welcome to the Jungle. I compared them against yours line by line rather than trusting the earlier note, and several carry more detail than this branch does. There is nothing left here to merge.

You also named the real problem in passing: nothing enforces provider-to-table parity, which is why this kept happening. That gap is still open and it is the more valuable piece of work.

If you want it, that is a better PR than this one was, and I will not let it sit for a month.

PR #2661, opened 2026-08-10
fix(jd-similarity): the seniority gate fired on JD boilerplate and on any CV showing progression

Zoubeir23:
`recommendCvReuse` refuses to recommend reuse when the two documents describe different seniority levels. That gate **overrides the similarity score**, so a false positive turns a good reuse candidate into a full regeneration — the exact outcome the module exists to avoid.

It misfires two ways. Both are measured below against `main`.

## 1. Ordinary English read as a job level

`levelOf` matched a bare keyword anywhere in the text. `lead`, `principal` and `mid` are ordinary English, and JD boilerplate is full of them.

Two near-duplicate postings for the same seniority — the second just spells out "Principal responsibilities" and uses "lead" as a verb:

```
A: Senior Backend Engineer. We build payment infrastructure with Django, PostgreSQL
   and Redis. You will design REST APIs, improve reliability and mentor teammates.
B: Backend Engineer. Principal responsibilities: build payment infrastructure with
   Django, PostgreSQL and Redis, design REST APIs, improve reliability, and lead
   mentoring for teammates.
```

`main` → `{"decision":"regenerate","score":0.696,"reason":"level-mismatch"}`
The 0.696 score alone means `reuse-with-edits`; the gate discarded it.

Same with a customer segment: `mid-market` vs a senior posting → `regenerate` at score 0.571.

## 2. A document naming several levels collapsed to the earliest one

`LEVELS.findIndex(...)` returns the first index in the **LEVELS table**, not the level the document is about. Any CV showing career progression names every rank it held — and `jd-similarity.mjs`'s own usage line is `<new-jd.txt> <previous-jd-or-cv.txt>`, so a CV is an expected input.

```
CV: Experience: Junior Developer 2019-2021. Senior Developer 2021-2026. Django, PostgreSQL, React.
JD: Senior Developer. Django, PostgreSQL, React.
```

`main` reads that CV as **junior** → `{"decision":"regenerate","score":0.556,"reason":"level-mismatch"}`.

## The fix

- **`NON_LEVEL_FOLLOWERS`** — `principal` / `lead` / `mid` stop counting as levels when the following word puts them in their ordinary sense (`principal responsibilities`, `mid-market`, `lead the`). Only the trailing word is inspected, so `Principal Engineer`, `Lead Engineer` and `Mid-level Engineer` remain levels.
- **`levelsIn` returns every distinct level; `levelOf` resolves only when there is exactly one.** Several levels = ambiguous = the gate stands down and the similarity score decides. Guessing is worse than standing down: the gate exists to catch a *clear* difference.

Deliberately conservative in one more place — a JD saying "Senior Engineer reporting to the Principal Engineer" is now ambiguous and no longer blocks reuse. Losing a hard block is recoverable (the score still ranks it); a false block is not, because the recommendation never surfaces.

Behaviour is unchanged for every unambiguous pair: `Senior` vs `Junior`, `Intern` vs `Staff` and the CJK level words still block reuse.

## Test plan

`jd-similarity.mjs` had **no test file** (it was one of seven core scripts with zero test references). This adds `tests/jd-similarity-seniority.test.mjs`, picked up by `test-all.mjs` auto-discovery.

The relaxations are each paired with an assertion that a genuine mismatch still fires, and the two "recommendation follows the score" assertions are guarded by an explicit check that the fixture sits in the medium band — otherwise they would pass for the wrong reason if a fixture ever drifted.

- [x] 6 of the 13 assertions verified RED on `main`; the 7 guard assertions pass both before and after, which is what makes them guards.
- [x] `node test-all.mjs --only jd-similarity-seniority` — 13 passed, 0 failed.
- [x] `node test-all.mjs` — 3150 passed, 0 failed (the single warning is the pre-existing `Dashboard build skipped — go compiler not in env`).

## Out of scope, noted while measuring

`tokenize` never segments CJK: `'负责后端服务开发，具备Python经验'` yields whole clauses as single tokens, so two near-identical Chinese JDs score 0.500. A side effect is that the ten Chinese entries in `STOP_WORDS` are unreachable — they can only match a standalone token, which CJK text never produces. Fixing that means real segmentation (`Intl.Segmenter`) and it would move every CJK similarity score, so it does not belong in this PR. Happy to open it separately if you want it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved seniority-level detection in job-description matching.
  * Prevented ordinary terms such as “principal responsibilities,” “lead,” and “mid-market” from being misclassified as seniority levels.
  * Documents mentioning multiple seniority levels are now treated as ambiguous, reducing incorrect reuse or mismatch decisions.
  * Preserved blocking behavior for genuine seniority mismatches.

* **Tests**
  * Added coverage for seniority detection, similarity thresholds, reuse decisions, boilerplate, and career-progression scenarios.

### DIFF
--- a/jd-similarity.mjs
+++ b/jd-similarity.mjs
@@ -48,14 +48,71 @@ export function jaccardSimilarity(left, right) {
   return intersection / (a.size + b.size - intersection);
 }
 
-/** Detect the first recognized seniority level in text, or -1 when absent. */
-function levelOf(text) {
+/**
+ * Level words that are also ordinary English, mapped to the words that follow
+ * them in their NON-seniority sense. JD boilerplate is full of these: "Principal
+ * responsibilities" means "main duties", "mid-market" is a customer segment, and
+ * "lead" is usually a verb. Matching them as job levels made the gate below fire
+ * on postings of identical seniority.
+ *
+ * Only the trailing word is inspected: "Principal Engineer" and "Lead Engineer"
+ * stay levels because `engineer` is not in any of these lists.
+ */
+const NON_LEVEL_FOLLOWERS = {
+  principal: ['responsibilities', 'responsibility', 'duties', 'accountabilities', 'objectives', 'purpose', 'tasks', 'activities'],
+  lead: ['to', 'the', 'a', 'an', 'our', 'and', 'or', 'by', 'on', 'in', 'for', 'with', 'from', 'mentoring', 'projects'],
+  mid: ['market', 'size', 'sized', 'cap', 'tier', 'funnel', 'term', 'sized-company'],
+};
+
+/** Whether a level word at `index` reads as a job level rather than plain English. */
+function readsAsLevel(word, normalized, index) {
+  const followers = NON_LEVEL_FOLLOWERS[word];
+  if (!followers) return true;
+  const after = normalized.slice(index + word.length).match(/^[^a-z0-9]*([a-z0-9-]+)/);
+  return !after || !followers.includes(after[1]);
+}
+
+/**
+ * Every distinct seniority level named in the text, as LEVELS indices.
+ *
+ * Returns ALL of them rather than the first: a document may name several (a CV
+ * showing career progression names each rank it held), and `findIndex` used to
+ * collapse that to whichever appeared earliest in the LEVELS table — reading a
+ * junior-to-senior CV as junior.
+ */
+function levelsIn(text) {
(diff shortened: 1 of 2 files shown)


santifer:
Another good one, @Zoubeir23. A seniority gate that trips on boilerplate is worse than no gate: it rejects matches for a phrase that appears in half the postings on the internet and never in the part of the JD that describes the role. Merged! 🚀

PR #2662, opened 2026-08-10
fix(web): report uncapped weekly match count

felipe-NR:
## What does this PR do?

The Today banner derived its weekly match total from an offer payload capped at 24, so the displayed count could never exceed 24. This separates the full deduplicated count from the bounded card payload, lets the Explore fresh-matches view request the complete offer list, and adds regression coverage for capped, duplicate, and legacy-history cases.

## Related issue

None — bug fix, submitted directly per CONTRIBUTING.md.

## Type of change

- [x] Bug fix
- [ ] New feature

## Testing

- `GOFLAGS=-buildvcs=false node test-all.mjs` — 3,137 passed, 0 failed (the flag only avoids Go VCS stamping inside the temporary worktree)
- `cd web && npm test` — 63 passed, 0 failed

## Checklist

---

## Summary by CodeRabbit

* **New Features**
  * “What’s New” now provides accurate total offer counts, even when results are limited.
  * Added support for viewing more available fresh offers.
  * Improved duplicate handling and compatibility with older history data.

* **Bug Fixes**
  * Dashboard headlines and “See all” links now reflect the full number of fresh offers rather than only the currently loaded results.
  * Offer limits are consistently applied while preserving complete match counts.


santifer:
Thanks for this, @felipe-NR.

The Today banner pulled its weekly total from an offers payload capped at 24, so the counter could never go past 24 no matter how many fresh matches existed. Splits the full deduplicated count from the capped card payload and lets the Explore fresh-matches view request the whole list. Regression coverage for the three causes behind it (cap, duplicates, legacy history); 63 web tests and typecheck green.

santifer (changes requested):
The bug is real and your diagnosis of it is exactly right — thank you for finding it. One change requested, on the `?limit=all` half.

## What's right

`whats-new/route.ts` capped at 24 *while accumulating*, so the count could never exceed 24 no matter how many fresh matches existed. Separating **the full deduplicated count** from **the bounded card payload** is the correct shape, and extracting `collectWhatsNew` into `whats-new.mjs` makes the legacy/duplicate/capped cases testable instead of buried in a route.

The Today banner is done exactly right: it still fetches the **bounded** payload and reads `d.count` for the number. It counts without downloading. That's the pattern.

## The change I'd like: Explore shouldn't need `limit=all` either

`explore-provider.tsx` switches to `/api/whats-new?limit=all`, which sets `offerLimit = Number.POSITIVE_INFINITY`. Two things make that risky in a way the Today path isn't:

1. **`explorer-view.tsx` has no virtualisation and no pagination** — I checked, there's no windowing or slice on the render path. Every offer returned becomes a DOM node.
2. **The default `portals.yml` is a firehose.** ~120 enabled companies against a broad catch-all `title_filter.positive` (bare `"AI"`, `"ML"`, `"Agent"`…) produces on the order of **12,000 jobs** on a fresh scan. A 7-day window of *un-evaluated* rows out of that is not 24, and it isn't 200 either.

So on a default install this trades a wrong number for an unbounded render. Before, Explore received 24 and showed 24 (count wrong, render bounded). After, count is right and the render has no ceiling — and the people most affected are new users, who have the demo config and the biggest backlog at once.

The fix is the pattern you already applied to the banner: **`count` is free, the list is not.** The endpoint returns the true total in the bounded response, so Explore can request a finite, larger limit (a few hundred) and still render an honest "See all N" from `count`. If Explore genuinely needs every row in the DOM, that's a windowing change in `explorer-view.tsx` and a bigger PR than this one.

Concretely, I'd accept either:

- `?limit=<n>` with a real number, Explore passing something like 200, `count` unchanged; or
- keeping `limit=all` but having `explorer-view` window the render.

The first is a two-line change to this PR and keeps its scope intact. My preference.

## Small note, not blocking

`Number.isFinite(count) ? Math.max(0, Math.trunc(count)) : offers.length` in both clients is the right defensive read — a malformed `count` degrades to the visible list rather than rendering `NaN`. Worth keeping.

Everything else is ready: CI is green on all three platforms and this doesn't collide with the `clis.ts` cluster, so it can land as soon as that one thing is settled.

felipe-NR:
Took the first option. `count` is untouched and the scope stayed where you wanted it.

## The ceiling is finite on both paths now

`resolveOfferLimit()` sits next to `collectWhatsNew` in `whats-new.mjs`: it parses `?limit=`, clamps to `[1, MAX_OFFER_LIMIT]`, and falls back to `DEFAULT_OFFER_LIMIT` when the param is absent — 200 and 24 respectively, the 24 being the number the home cards already used. The route calls it instead of branching on the sentinel. Explore asks for `MAX_OFFER_LIMIT`.

Before, Explore asked for everything and rendered everything (count right, render unbounded). After, it asks for 200, renders at most 200, and reads the true total from `count` — which never went through the cap at all. Count is free, the list is not. Same pattern as the banner, now on both callers.

The constant is exported rather than written into the fetch, so the ceiling and the clamp can't drift apart. If `explorer-view` gets windowing later, raising it is one line in one place.

## `limit=all` now degrades down, not up

Worth flagging because it wasn't in either option you offered. The old sentinel is non-numeric, so it resolves to the bounded default instead of `Infinity` — a stale client can't reintroduce the unbounded render. That felt like the right direction given who the failure mode hits: new users on the demo config are also the least likely to be on a fresh build.

## CodeRabbit caught a real one on top of it

My first pass used `Number.parseInt`, which accepts a numeric *prefix*. `limit=24junk` resolved to 24, and `limit=1e9` resolved to **1** — a single offer to a caller plainly asking for everything. It parses the whole trimmed value with `Number` and requires `Number.isInteger` now, so anything malformed falls back to the default. `isInteger` is also what rejects `Infinity`, which means it's what actually enforces the paragraph above. Both threads are resolved.

## Tests

Six new cases beside the existing three: absent/blank limit, the finite-ceiling contract (`all`, `Infinity`, `NaN`, over-max), partially numeric input, surrounding whitespace, in-range and degenerate clamping, and count-vs-payload at the Explore ceiling.

- `node test-all.mjs` — 3,938 passed, 0 failed
- `cd web && npm test` — 258 passed, 0 failed

## The red check predates this PR

`web typecheck + build` fails on `web/tests/lib/tracker-lock.test.mjs` with `Cannot find package 'js-yaml' imported from tracker-utils.mjs`. That suite landed in d9c4fd0 (#2903) and imports the root `tracker-utils.mjs`, but `web-ci.yml` runs `npm ci` under `working-directory: web`, so the root deps never get installed. Web CI is red on main itself from that commit onward — green at 6eece73, red at d9c4fd0 — and every PR that merges current main picks it up. All eight `whats-new` tests pass in that same run; the only two failures are the tracker-lock pair.

One line fixes it, `npm install --ignore-scripts` at the root, the pattern `test.yml` already uses. That's your workflow and not this PR's scope, so I left it alone — say the word and it goes out separately.

`test (macos-latest)` was an unrelated flake in `agent-inbox-tests.mjs` (`kept=29 of 30`). Green on the re-run. Everything else is green.

And the defensive `count` read stayed in both clients.

felipe-NR:
**Resolved upstream — the workflow PR isn't needed.**

Merged current `main`, which carries `48bf430` (#2929). All 14 checks are green.

Worth noting you fixed it the opposite way from what I offered: I proposed installing the root deps in CI, you kept the web-only install and made the guard answer "can this be imported" instead of "is the file there". That's the better call — it keeps `web-ci.yml` honest about what a web-only checkout actually is, and the `err.url == null` split means a genuinely broken core still fails loudly instead of being reported as an uninstalled dependency. Withdrawing the `npm install --ignore-scripts` suggestion.

santifer (approved):
Re-review — this resolves what I asked for on the `?limit=all` half. `resolveOfferLimit` is exactly right: `Number` not `parseInt` (so "24junk"→NaN and "1e9"→NaN fall back to the default instead of silently under-fetching as 24/1), `Number.isInteger` rejecting Infinity, and a finite `MAX_OFFER_LIMIT` ceiling. The test pins the dangerous sentinels directly — `"all"`, `"Infinity"`, `"-Infinity"` all degrade to a bounded default, never Infinity into the DOM — and keeps the complete deduplicated `count` separate from the bounded card payload, which was the core of the original bug. Approving; it merges once the field's free (my run turn). Thanks for the patience, @felipe-NR — 9 days is on us.

santifer:
Thanks for the patience, @felipe-NR — nine days is on us. The delivery resolves the `?limit=all` half cleanly: the count now reflects every deduplicated match, the render stays bounded, and the malformed-limit handling is genuinely careful. Merged 🚀

PR #2663, opened 2026-08-10
fix(web): find Grok in ~/.grok/bin

mbrow190:
#2688 registered Grok in KNOWN. A default install still lives in ~/.grok/bin on Unix and %LOCALAPPDATA%\grok\bin on Windows. Those dirs are often off PATH.

This PR adds the two search dirs and mentions Grok on the Explore line in web/README.md. KNOWN and the permission comment stay as #2688 left them. No --always-approve.

## Summary by CodeRabbit

- **New Features**
  - Added support for detecting Grok Build CLI installations.
  - Added detection for Grok Build CLI in its default installation locations on Unix-like systems and Windows.

- **Documentation**
  - Updated the configuration documentation to include Grok Build CLI among supported locally detected AI CLIs.


santifer:
Thanks for this, @mbrow190.

Adds Grok Build CLI to the web Config detector: an entry in web/src/lib/clis.ts, a ~/.grok/bin lookup for when it isn't on PATH, and the headless --always-approve flag for jobs launched from the web. The diagnosis holds up: the core already lists Grok in docs/SUPPORTED_CLIS.md and AGENTS.md, it was only the web that wasn't detecting it. 15 lines across 2 files, scoped to detection alone, with evaluate orchestration and Explore filters left untouched.

santifer:
You found a real gap — Grok Build is documented for headless use and the web's Config detector didn't list it. Thank you for that.

I'm merging **#2688** for the entry itself, filed a day after yours, and I want to explain the tiebreak honestly because it isn't "first come" — yours came first — and it isn't quality either. It's one flag.

## `--always-approve`

```ts
args: (p) => ["-p", p, "--always-approve"],
// --always-approve lets web-triggered jobs use Write/Bash without a TUI.
```

Your comment is right about what it does, and that's the problem. The web's permission doctrine is written into `web/src/lib/claude-invocation.mjs`, and it is deliberately strict:

> `disallowedTools` is the hard guardrail; the allow list is the working set. Every write-capable tool a kind does not need is **EXPLICITLY denied, never merely omitted**: `--permission-mode acceptEdits` exists to auto-approve edit tools, so **"unmentioned" is the one status a file-writing tool must never have**.

And `WRITE_CAPABLE_TOOLS` includes `Bash`, carrying this note:

> `Bash` counts: `sh -c '> cv.md'` is a write, which is why **pdf lost Bash in #2172 and must never regain it**.

So on Claude, the `pdf` worker has Bash explicitly denied. With `--always-approve` on Grok — which has no per-tool deny list to pair it with — that same `pdf` worker would get Write and Bash auto-approved. **The user's choice of runtime would silently change what a worker is allowed to do to their files**, with the two paths looking identical in the UI. That's the specific thing #2172 was fixed to prevent.

This is not a knock on the flag's intent: headless *does* need to run without a TUI prompt, and you're solving a genuine problem. It's that the permission surface has to be bounded per-kind the way Claude's is, not opened globally, and that's a design question bigger than adding a CLI entry.

## What I'd like to take from this PR

**Your install paths, which #2688 doesn't have:**

```ts
path.join(home, ".grok/bin"),                  // Grok Build CLI default install prefix
path.join(localAppData, "grok", "bin"),        // Windows
```

That's the part that makes detection actually work for people who didn't install onto `PATH`, and it has none of the permission concerns. If you're up for it, reduce this PR to those two lines (plus the README touch) once #2688 lands, and I'll merge it quickly — you'd be fixing the half that the other PR leaves broken.

If you'd rather not, say so and I'll port them myself with credit to you in the commit. Either way the finding is yours: two people spotted the same missing CLI within a day, which says the gap was real and worth closing.

mbrow190:
@santifer Agreed on --always-approve. Grok has no per-tool deny list, so that flag would auto-approve Write and Bash on workers Claude explicitly denies.

The PR is now only the two install paths #2688 missed, plus the README mention.

santifer:

santifer (approved):
Approving this, @mbrow190, with the same apology I owe you as I owed @moelomda: I commented on 14 August and never wrote a verdict. Eighteen days of my tooling counting this as "waiting on the author" while it was waiting on me. A comment is not a disposition, and I had been treating it as one.

**Premise verified at source rather than taken from the description.** You say #2688 registered Grok in `KNOWN`; it did, at `web/src/lib/clis.ts:71`:

```ts
{ id: "grok", name: "Grok Build CLI", bin: "grok", run: "grok -p", url: "https://docs.x.ai/build/overview", args: (p) => ["-p", p] },
```

So the CLI was known and its default install directory was not searched, which is a half-finished feature rather than a missing one. That is the gap this closes.

**Why this is safe by construction, which is what made it an easy approve once I actually looked:** `searchDirs()` only builds a list of candidate paths. `findBin` (`clis.ts:118-123`) walks them with `fs.accessSync(p, X_OK)` inside a `try`, so a directory that does not exist on a given machine contributes nothing and costs nothing. The failure mode of guessing an install path wrong is "no candidate found", not an error. That matters here because neither of us can verify x.ai's installer layout from this repo. Your two paths follow the exact shape of the neighbours (`.bun/bin`, `.deno/bin`, and `agy/bin` on Windows with its own trailing comment), so if they are wrong they are wrong harmlessly and visibly.

The `web/README.md` line is right to change with it: mentioning Grok on the Explore line is what makes the support discoverable, and a CLI the app can find but never names is only half-supported.

Scope is exactly what you said it was: `KNOWN` untouched, the permission comment untouched, no `--always-approve`. I checked, because "no `--always-approve`" is the kind of claim I verify rather than read.

santifer:
Merged. Thank you, @mbrow190, and sorry for the eighteen days.

Worth saying plainly: this finished something that was left half-done. #2688 taught the app that Grok exists; without your two paths it could not find it where Grok actually installs itself. A CLI that is registered but unfindable is the kind of gap that reads as "unsupported" to the person hitting it.

Issue #2666, opened 2026-08-10
web: ASCII-only company normalization silently drops or mis-matches non-Latin names


santifer:
Found while checking whether the web needed to sync with #2569's `normalizeTextKey` change. It doesn't have a *copy* to sync — it has **three independent reimplementations**, and two of them are ASCII-only.

## The bug

`web/src/app/api/whats-new/route.ts:14` and `web/src/components/explore/explorer-view.tsx:18` both define:

```js
const norm = (s) => s.toLowerCase().replace(/[^a-z0-9]+/g, " ").trim();
```

`[^a-z0-9]` deletes every non-ASCII letter. Compared against the core's `normalizeTextKey` (`\p{L}\p{M}\p{N}`, Unicode-aware):

| Company | core | web `norm` |
|---|---|---|
| Nestlé | `nestlé` | `nestl` |
| Zürich Re | `zürich re` | `z rich re` |
| Ação Digital | `ação digital` | `a o digital` |
| Škoda | `škoda` | `koda` |
| 日本電産 | `日本電産` | **`""` (empty)** |

## Why it matters

That `norm` is not cosmetic — it decides **which job offers get suppressed**:

```js
const evaluated = new Set(readApplications().map((a) => norm(a.company)).filter(Boolean));
// …
if (company && evaluated.has(norm(company))) return null;   // don't resurface as "new"
```

Two real failure modes:

1. **Offers vanish (the dangerous one).** Two different companies can collide after ASCII-stripping — `Škoda` → `koda` matches a tracked `Koda`. A genuinely new offer is then suppressed as "already evaluated" and **the user never sees it**. Silent loss, no signal.
2. **Duplicates forever.** A CJK company name normalizes to `""`, which `.filter(Boolean)` drops from the evaluated set — so it can never match, and an already-evaluated company keeps resurfacing as fresh.

This is not hypothetical: career-ops ships German (DACH), French, Japanese, Turkish and Arabic modes, and the scanners hit EU/JP boards. Accented company names are the norm in those markets, not an edge case.

A third variant, `registry.ts:87` `normCompany` (`\s+` only, no punctuation stripping), is Unicode-safe but disagrees with both others — so the three surfaces don't even match each other.

## Fix

[...]

PR #2667, opened 2026-08-10
fix(web): derive company matching keys from the core, not an ASCII-only copy

santifer:
Server half of #2666.

## The bug

`whats-new` keyed companies with a local `[^a-z0-9]` normalizer, which deletes every non-Latin letter. That key decides **which offers get suppressed as already-evaluated**, so it had two live failure modes:

- **Silent loss:** a tracked `Koda` swallows a genuinely new `Škoda` offer — the user never sees it, with no signal.
- **Never matches:** `日本電産` keys to `""`, gets dropped by `.filter(Boolean)`, and resurfaces forever.

career-ops ships German, French, Japanese, Turkish and Arabic modes and scans EU/JP boards, so this hit the markets it most needed to serve.

## The fix

Keys come from the core's `normalizeTextKey` (#2569) through a new ACL, `lib/core/text-key.ts`.

It **can't be a static import**: the core lives in the *user's* checkout, resolved at runtime via `careerOpsRoot()`, and isn't a build dependency of this app. So the ACL imports it dynamically, caches per resolved path, and **never caches a failure** — the #2590 lesson, where a cached fallback pinned stale definitions for a whole process lifetime.

**By design:** the web keys with the user's *own* core, so web dedup always matches their CLI dedup. Version skew becomes semantic, which is the correct behaviour.

**The fallback is a copy, and the code says so** rather than hiding it — but a Unicode-aware one. Worst case it disagrees with the core about an exotic separator; it can never turn a company name into `""`.

## Verification

Ran the failing scenario live. With `Koda` already in the tracker:

| | |
|---|---|
| `Škoda` offer | ✅ surfaces (previously swallowed by `Koda`) |
| `日本電産` offer | ✅ surfaces (previously keyed to `""`) |
| `Koda` itself | ✅ still correctly suppressed |

Also exercised the fallback path: importing the core **throws catchably** on a partial checkout — `tracker-parse.mjs` loads `tracker-aliases.json` at module scope — and the ACL degrades instead of 500ing. Worth knowing generally: deriving from the core means inheriting its load-time file dependencies.

typecheck + `next build` green.

## Not in this PR

The **client half**: `explorer-view.tsx` must stop normalizing entirely and compare server-provided keys. A `"use client"` component can't reach the core at all, so the cure there is moving the computation server-side rather than importing. Follows separately, and it's the shape the maintainer's structural freeze will check (server derives / client doesn't normalize).

Refs #2666, #2369.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved company-name matching in the “What’s New” experience.
  * Company names containing non-Latin characters are now preserved and handled correctly.
  * Prevented distinct company names from being incorrectly treated as identical when filtering evaluated offers.
  * Added a reliable fallback for text matching when the shared text-normalization service is unavailable.

### DIFF
--- a/web/src/app/api/whats-new/route.ts
+++ b/web/src/app/api/whats-new/route.ts
@@ -1,6 +1,7 @@
 import fs from "node:fs";
 import path from "node:path";
 import { careerOpsRoot, readApplications } from "@/lib/career-ops";
+import { getNormalizeTextKey } from "@/lib/core/text-key";
 import type { DiscoveredOffer } from "@/lib/explore";
 
 export const runtime = "nodejs";
@@ -11,7 +12,10 @@ export const dynamic = "force-dynamic";
 // evaluated yet. No scan runs here — it reads the history a past scan already
 // wrote, so the home stays instant + free (directly answers the #1 token-cost
 // complaint). cols: url, first_seen, portal, title, company, status, location.
-const norm = (s: string) => s.toLowerCase().replace(/[^a-z0-9]+/g, " ").trim();
+// Company matching keys come from the CORE (see lib/core/text-key.ts), never a
+// local reimplementation. The previous ASCII-only key deleted every non-Latin
+// letter, so "Škoda" collided with "Koda" — suppressing a real offer as
+// "already evaluated" — and "日本電産" keyed to the empty string (#2666).
 
 export async function GET(req: Request) {
   const days = Math.min(30, Math.max(1, Number(new URL(req.url).searchParams.get("days")) || 7));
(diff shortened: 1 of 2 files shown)


santifer:
Merged after an independent core-side check. I reproduced the bug and the fix against the real corpus rather than reading the diff:

| input | old web key | core key |
|---|---|---|
| `Škoda` | `koda` | `škoda` |
| `Nestlé` | `nestl` | `nestlé` |
| `日本電産` | `""` | `日本電産` |
| `null` | `"null"` | `""` |

`Škoda` and `Koda` collided under the old key, so a genuinely new offer was suppressed as "already evaluated" — invisible, with no signal. That's the user-facing bug this closes.

I also checked the fallback path, since a fallback is a copy by definition: it matches the core on every case in that corpus (0 mismatches), so a partial checkout degrades to an equivalent Unicode-safe key and can never reproduce the `""` failure. Caching only successes (never a failed import) is the right call — a cached failure is what pinned stale definitions for a whole process last time.

The client half (dropping normalization in `explorer-view` and comparing server-supplied keys) is tracked separately.

PR #2668, opened 2026-08-10
fix(web): finish Unicode company keys for explore + registry (#2666)

pravit-amp:
## Summary
- Completes #2666 after #2667: replace the remaining ASCII-only `norm` in `explorer-view` and the divergent `normCompany` in `registry` with a shared Unicode-safe `normalizeTextKey` mirror.
- Server routes still prefer live core via `getNormalizeTextKey()`; client uses the parity-tested `.mjs` mirror (core isn't available in the browser).
- Adds tests that lock Nestlé / Škoda≠Koda / 日本電産 / nullish against `tracker-parse.mjs`.
## Test plan
- [X] `cd web && node --test tests/lib/normalize-text-key.test.mjs`
- [X] `cd web && npm run typecheck`
- [X] Spot-check Explore enrichment: tracked Nestlé/Škoda/日本電産 rows match without false suppressions or empty-key resurfacing

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved text and company-name matching for accented characters, CJK text, punctuation, and other Unicode content.
  * Reduced key collisions and prevented empty keys for supported non-Latin text.
  * Standardized handling of missing values and separators.

* **Tests**
  * Added coverage for Unicode normalization, separators, nullish values, and previous normalization edge cases.

### DIFF
--- a/web/src/app/actions/registry.ts
+++ b/web/src/app/actions/registry.ts
@@ -10,6 +10,7 @@
 
 import type { Application, InboxJob } from "@/lib/career-ops";
 import type { Job } from "@/components/jobs/job-store";
+import { normalizeTextKey } from "@/lib/core/normalize-text-key.mjs";
 
 export const AUTO_FIRE_MAX = 3; // fire ≤3 evaluations silently; confirm above that
 export const BATCH_CAP = 12; // hard ceiling on a single fan-out
@@ -84,7 +85,8 @@ export type DispatchResult =
 
 // ── helpers ──────────────────────────────────────────────────────────────
 const isStr = (v: unknown): v is string => typeof v === "string" && v.length > 0;
-const normCompany = (s: string) => s.toLowerCase().replace(/\s+/g, " ").trim();
+// Same shape as core normalizeTextKey(s, " ") — keeps accents/CJK, strips punct (#2666).
+const normCompany = (s: string) => normalizeTextKey(s, " ");
 
 // Allow only in-app routes (with optional :segment and ?query). Blocks anything
 // that isn't one of the app's own sections — the model can't navigate offsite.
--- a/web/src/components/explore/explorer-view.tsx
+++ b/web/src/components/explore/explorer-view.tsx
@@ -6,6 +6,7 @@ import Link from "next/link";
 import { cn } from "@/lib/cn";
 import { instrumentSerif } from "@/lib/fonts";
 import type { Application, InboxJob } from "@/lib/career-ops";
+import { normalizeTextKey } from "@/lib/core/normalize-text-key.mjs";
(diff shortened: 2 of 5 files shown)


santifer:
🟢 **Approving from the web side.** This finishes #2666, and it does one thing better than what I had planned.

I was about to write the client half myself; you got there first and your shape is cleaner. I had left an inline `fallbackKey` inside `text-key.ts` in #2667 — you extracted it into a single shared `normalize-text-key.mjs` that the server ACL uses as its fallback and the client imports directly. **That leaves exactly one copy in the tree instead of two**, which matters more than it looks: the whole #2369 thread is about copies nobody watches, and one guarded copy is tractable where two are not.

**Verified the things I most wanted to survive from #2667, and they did:**
- Derivation is still primary — `await import(pathToFileURL(...))` of the user's own core, not the mirror
- Failures are still never cached (the #2590 lesson: a cached fallback once pinned stale definitions for a whole process lifetime)
- The parity test imports the **real** `tracker-parse.mjs` and asserts three ways — web matches expected, core matches expected, and web === core. Asserting against the real core rather than a hardcoded table is what makes it a parity test rather than a snapshot.

**Verified locally:** `npm test` 129/129, typecheck + `next build` green. Swept the tree for surviving ASCII-only keys: every remaining `[a-z0-9]` is a legitimate slug (filenames, domains, ids, a nav regex) or a comment warning against reinstating it. **Zero matching-key normalizers left.**

One note, not a change request: the parity test does a top-level `await import` of the core with no guard, so it hard-fails if the root files are missing — including `tracker-aliases.json`, which `tracker-parse.mjs` loads at module scope (I hit exactly that in #2667). Inside this repo that's always satisfied, and arguably a parity test *should* fail loudly when it can't compare. Flagging it so the failure mode is known rather than surprising.

Good catch on `registry.ts` — it was a third divergent variant (`\s+` only), and folding it in was right even though it was Unicode-safe, because three surfaces disagreeing with each other is its own bug.

santifer:
Welcome to career-ops, @pravit-amp — and this is a strong first contribution. It closes the half of #2666 that was still open, and it does something better than what was planned: extracting the fallback into a shared `normalize-text-key.mjs` leaves **one** copy in the tree instead of two, so the thing that has to be watched for drift is now a single object rather than a pair. That's the part that makes the parity guard tractable.

Also right: the server still derives from the user's live core and only falls back to the mirror, failures still aren't cached, and the parity test imports the real `tracker-parse.mjs` rather than asserting against a hand-written expectation.

Reviewed and approved on the web track; I verified the core side independently. Škoda no longer collides with Koda, and 日本電産 no longer keys to the empty string — which means real openings stop disappearing for anyone whose market isn't English-speaking. Merged! 🚀

PR #2672, opened 2026-08-11
fix(reply-matcher): require corroboration for partial role-title matches

### LINKED ISSUE #2671: reply-matcher checkRoleMatch(): generic multi-word role titles false-positive match unrelated emails
### Code of Conduct

### Existing issues
- [x] I searched existing issues and this hasn't been reported yet (related but distinct: #2134 domain-harvesting junk tokens, #2147 bare "offer" substring — neither covers role-title matching)

### What happened?

`checkRoleMatch()` in `reply-matcher.mjs` treats any role-title word longer than 3 characters as sufficient evidence of a match, with no requirement that any other signal (company name, sender domain) also line up:

```js
const roleParts = role.split(/[\s_\/()-]+/);
for (const part of roleParts) {
  if (part.length > 3 && tNorm.includes(normalizeStr(part))) {
    return true; // partial match on a significant word
  }
}
```

For a tracker application titled e.g. **"Talent Acquisition Specialist"**, this splits into parts `["Talent", "Acquisition", "Specialist"]` — all >3 chars. Any email whose body contains *any one* of those words anywhere, in any context, satisfies the match — with zero confirmation that the email is actually about that application.

**Real repro found via `scan-gmail-replies.mjs` (bulk reply-candidate import, unrelated feature) run against a live mailbox:** an unrelated employer's recruiter had the job title *"Talent Acquisition & Diversity"* in their email signature. Every message in that recruiter's thread — entirely about a *different* company's *different* role — matched against the tracker row for a "Talent Acquisition Specialist" application, because "Talent Acquisition" appeared in the signature block on every reply in the thread. 5 messages in one thread were misattributed this way, producing conflicting `Interview` vs `Rejected` classification recommendations against the wrong tracker row (`reply-watch.mjs`'s conflict-detection caught the *symptom* — inconsistent recommendations for one row — but not the underlying misattribution).

**Root cause:** the >3-char-substring-anywhere rule has no requirement that a second, independent signal (sender domain, company name in the text) also match. Generic multi-word role titles built from common recruiting/HR vocabulary ("Talent Acquisition", "People Operations", "Human Resources", "Customer Success") are especially exposed, since those exact phrases routinely appear in unrelated senders' signatures, job titles, and boilerplate.

### Proposal

Require role-part matching to be corroborated by at least one of:
- the application's company name (or a normalized fragment of it) also appearing in the text, or
- the sender's email domain matching a domain associated with that application (reusing/extending the domain-matching path `checkCompanyMatch`/`extractDomain` already provide elsewhere in this file)

A role-title-only match with no company/domain corroboration should downgrade to a weaker "candidate" match (e.g. surfaced for manual review) rather than an confirmed `application_num` assignment. Happy to also add a stoplist of generic recruiting-vocabulary words (e.g. "talent", "acquisition", "specialist", "coordinator", "operations") that should never count as a matching "significant word" on their own regardless of length, since these are exactly the words most likely to appear in unrelated senders' signatures.

Add regression coverage: a candidate whose only signal is a generic role-title substring, from a sender/company that doesn't match the application, should not match.

Schlaflied:
## Summary
- `checkRoleMatch()` treated any role-title word longer than 3 characters as sufficient evidence of a match on its own, with no requirement that a second signal (company name or sender domain) also corroborate it. A tracker row titled e.g. "Talent Acquisition Specialist" splits into `["Talent","Acquisition","Specialist"]`, and any unrelated email containing just one of those words anywhere — such as a different recruiter's signature line ("Talent Acquisition & Diversity") for a completely different company — satisfied the match and got misattributed to that application.
- Adds `checkRoleMatchExact()`, which only credits a match on the *entire* role title (or its Chinese, symbol-stripped form) as one contiguous substring. That's specific enough to stand alone. A role that's a single generic word (e.g. `role: "Recruiter"`) is excluded from this path too, since a whole-role check degenerates into a bare-word check when there's no multi-word phrase to be specific about.
- `matchCandidates()` now only counts a partial (significant-word) role match when it's corroborated by `checkCompanyMatch` or the sender-domain match, both already computed per-application in that loop.
- `checkRoleMatch()` also gains `GENERIC_ROLE_WORDS`, a stoplist of generic recruiting/HR vocabulary (talent, acquisition, specialist, coordinator, operations, recruiter, recruiting, human, resources, people) that never counts as a significant word by itself, regardless of length — these are exactly the words most likely to collide with unrelated senders' signatures/titles.

## Non-goals / guarantees preserved
- `checkCompanyMatch`, exact/near-exact role matches, and Chinese role-title handling are unchanged.
- No new match-confidence tiers beyond what's needed: an uncorroborated partial role match now simply doesn't count as a role-title signal, same as any other absent signal — it falls through to the existing `no-match`/`ambiguous-match` handling in `matchCandidates`.

## Test plan
- [x] `node --test reply-matcher.test.mjs` — all passing, including new regression coverage:
  - `checkRoleMatch - generic recruiting words never match alone (#2671)`
  - `checkRoleMatchExact - only a full contiguous role-title match counts`
  - `matchCandidates - a generic role-title word from an unrelated sender does not match (#2671)`
  - `matchCandidates - a partial role-word match corroborated by company name still matches`
  - `matchCandidates - a genuinely specific role match still works standalone`
- [x] `node test-all.mjs` — 3276 passed, 0 failed (1 pre-existing warning, unrelated to this change)

Closes #2671

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved role matching by distinguishing exact job titles from generic recruiting language.
  * Prevented unrelated messages from matching based solely on broad recruiting terms.
  * Partial role matches now require supporting company or sender-domain evidence.
  * Preserved valid matches for exact titles and specific company-correlated roles.

* **Tests**
  * Added coverage for exact titles, generic terms, unrelated messages, and company-supported partial matches.

### DIFF
--- a/reply-matcher.mjs
+++ b/reply-matcher.mjs
@@ -43,26 +43,63 @@ export function checkCompanyMatch(text, company) {
   return false;
 }
 
-export function checkRoleMatch(text, role) {
+// Generic recruiting/HR vocabulary. These words are common enough in unrelated
+// senders' signatures, job titles, and boilerplate (e.g. "Talent Acquisition &
+// Diversity" in a recruiter's signature for a *different* company/role) that
+// they must never, by themselves, count as a "significant word" match against
+// a tracker role title — regardless of length (see #2671).
+const GENERIC_ROLE_WORDS = new Set([
+  'talent', 'acquisition', 'specialist', 'coordinator', 'operations',
+  'recruiter', 'recruiting', 'human', 'resources', 'people'
+]);
+
+// A role title that is a single word and that word is generic recruiting
+// vocabulary (e.g. role === "Recruiter") degenerates a "whole role" substring
+// check into exactly the same bare-word check the stoplist exists to block —
+// there is no multi-word phrase for the match to be specific about.
+function isBareGenericRole(role) {
+  const parts = role.split(/[\s_\\/()-]+/).filter(Boolean);
+  return parts.length === 1 && GENERIC_ROLE_WORDS.has(parts[0].toLowerCase());
+}
+
+// True only when the *entire* role title (or its Chinese, symbol-stripped form)
+// appears in the text as one contiguous substring. This is specific enough to
+// stand on its own, with no need for a corroborating company/domain signal —
+// unless the role is nothing but a single generic word (see isBareGenericRole).
+export function checkRoleMatchExact(text, role) {
   if (!role || !text) return false;
-  
+  if (isBareGenericRole(role)) return false;
+
   const tNorm = normalizeStr(text);
   const rNorm = normalizeStr(role);
   if (tNorm.includes(rNorm)) return true;
 
+  // Handle Chinese role titles ignoring symbols
+  const cleanRole = role.replace(/[\s_\\/()-]+/g, '');
+  if (cleanRole.length > 2 && tNorm.includes(cleanRole.toLowerCase())) return true;
+
+  return false;
+}
+
+export function checkRoleMatch(text, role) {
+  if (!role || !text) return false;
+
+  if (checkRoleMatchExact(text, role)) return true;
+
+  const tNorm = normalizeStr(text);
+
   // Sometimes role has extra descriptors, we check if a significant part matches
-  // Like "PY01_python开发工程师" vs "python开发工程师"
+  // Like "PY01_python开发工程师" vs "python开发工程师". Generic recruiting words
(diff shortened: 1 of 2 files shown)


santifer:
Merged, @Schlaflied. A partial title match on its own is weak evidence — "Engineer" appears in half an inbox — and treating it as sufficient is how a reply gets attached to the wrong application, which is worse than not matching at all: the row looks handled and the real thread stays invisible. Requiring corroboration before acting on a partial is the right threshold. Closes #2671.

PR #2675, opened 2026-08-11
fix(providers): harden workable with retry, headers, and serialization

ilyakanevskiy:
## Summary
- Add browser-like request headers (`user-agent`, `accept-language`, `origin` + a matching `referer`) to both the widget API request and the markdown-feed fallback in `providers/workable.mjs`.
- Add a provider-local retry loop with exponential backoff + jitter for transient failures (429/5xx/network errors), reusing the classification already shared via `providers/_http.mjs`'s `isRetryableError`/`parseRetryAfterMs` rather than duplicating it — the retry loop itself stays local because it needs to give up early on a long server-declared `Retry-After` (see below) and it retries both a JSON call and a text call, neither of which the shared `fetchJsonWithRetry` helper supports on its own.
- Give up immediately (skip remaining retries) when the widget API returns a `Retry-After` longer than 30s, falling straight through to the markdown feed instead of stalling the scan on a long-lived block.
- Add process-wide request serialization against `apply.workable.com`, since every tenant is fronted by the same host.
- Fix a stale `docs/SUPPORTED_JOB_BOARDS.md` Workable row that still described the markdown feed as primary (the widget API has been primary since a prior PR) — updated to also mention the new retry/hardening behavior.
- Add the missing commented example stanza for Workable under "Built-in provider examples" in `templates/portals.example.yml` (auto-detection line and a live tracked-company example already existed; the abstract stanza didn't).
- Rename the test file's "Optimile" / "Ghent, Belgium" fixture (reads like a real company + location) to a clearly fictional "AcmeWidgets" / `acme-widgets` slug, matching this repo's existing `Acme`/`ExampleCo`-style fixture convention, before adding new coverage on top of it.

## Why
Live scans of a real Workable-backed tenant hit the widget API returning 429 with a multi-hour `Retry-After` (Cloudflare-level block on that specific path) while the markdown feed on the same account kept returning 200 the whole time — but the provider had no retry logic at all, so a single transient blip (or the long block) fell straight through with no chance to recover from a short one, and no headers meant a plain, unauthenticated-looking request profile.

## Test plan
- [x] `node test-all.mjs --only providers/workable` — 25/25 pass, including new coverage for browser-like headers + matching referer on both requests, `redirect:'error'` asserted explicitly, retry-and-recover on a short `Retry-After`, immediate give-up on an hours-long one (with fallback), immediate fallback on a non-retryable error, and the serialization guarantee (two concurrent `fetch()` calls never overlap in-flight).
- [x] `node test-all.mjs` — full suite green (3279 passed, 0 failed, 1 pre-existing unrelated warning).
- [x] `npm run validate:portals -- --file templates/portals.example.yml` — 0 errors, 0 warnings.
- [x] Live `--dry-run` scan against real Workable tenants — jobs returned, no errors.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Workable job listings retrieval with more reliable requests and automatic retries for temporary service interruptions.
  * Added handling for retry guidance and long service delays to avoid unnecessary waiting.
  * Improved fallback behavior when the primary listings source is unavailable.

* **Documentation**
  * Updated supported job board guidance with Workable detection and fallback details.
  * Added an example configuration for a Workable-hosted careers page.


ilyakanevskiy:
@coderabbitai review

santifer:
Merged, @ilyakanevskiy. Workable was the least defended of the providers — no retry on a transient 429 or 5xx, so a rate limit during a scan silently cost you that board's jobs. Matching the retry/backoff shape the other providers already use means there's nothing new to learn when reading it. Merged! 🚀

PR #2676, opened 2026-08-11
feat(invite-match): distinguish AI-interviewer platforms (Alex, HireVue) from human calls

### LINKED ISSUE #2673: invite-match: distinguish AI-interviewer platforms (Alex, HireVue) from human video calls in extractPlatform()
### Code of Conduct

### Existing issues

### What happened / motivation

`extractPlatform()` in `invite-match.mjs` (landed in #2128) currently recognizes Zoom, Microsoft Teams, and Google Meet via `PLATFORM_URL_PATTERNS`, plus a phone fallback. All of these are human-mediated calls with the same underlying prep needs.

A distinct and increasingly common category isn't covered: **AI-interviewer platforms**, where the candidate has a live conversation with an AI system instead of a human panel (e.g. Alex/Apriora at `meet.alex.com` / `alex.com`, HireVue). Real-world context: 87% of companies now use AI somewhere in hiring, interview screening is the single most-adopted AI recruiting use case (58% of companies), and 63% of US job seekers report already experiencing an AI interview in the past six months — this isn't a fringe case.

AI-interviewer calls warrant materially different prep than a human call:
- No human rapport/reading-the-room available — pacing and camera-eye-contact advice matters more, not less
- The AI generates adaptive follow-up questions from the literal content of what the candidate says, so vague answers get penalized differently than with a human interviewer
- Some platforms (Alex/Apriora specifically, post-rebrand from a 2024 public glitch incident) have documented failure modes — transcription-quality degradation and conversational loops, particularly with more varied accents/speech patterns — worth flagging to the candidate as a known platform quirk rather than something to self-blame for if the AI behaves oddly mid-call

None of this is currently distinguishable from a Zoom call in `extractPlatform()`'s output — a candidate prepping from `interview-prep.md`/`interview/debrief.md` output has no signal that the round is AI-mediated at all.

### Proposal

- Extend `invite-match.mjs` with recognized AI-interviewer platform patterns (start with Alex/Apriora: `meet.alex.com`, `alex.com`; HireVue: `hirevue.com` domains) similar in shape to the existing `PLATFORM_URL_PATTERNS` entries.
- Keep `extractPlatform()`'s existing return shape backward-compatible (plain platform-name string) for existing callers — add a new exported helper (e.g. `isAIInterviewerPlatform(text)` or a `{ name, isAIInterviewer }`-shaped variant) rather than changing the existing function's contract.
- Wire the detection into `modes/interview-prep.md` (and/or `modes/interview/debrief.md`, wherever platform is currently surfaced per #2128) to add a short, distinct prep block when an AI-interviewer platform is detected: camera-eye-contact / pacing / no-notes-reading guidance, and a note that mid-call glitches/repeats are a known platform issue, not a candidate-side signal.
- Zero-LLM, same discipline as the existing platform patterns — pure pattern matching, never guessed, `null`/absent when nothing plausible is found.

Add regression tests following the existing `invite-match.test.mjs` conventions for the new patterns and the Zoom/Teams/Meet/Phone paths remaining unaffected.

Schlaflied:
Closes #2673.

## The gap

`extractPlatform()` (landed in #2128) recognizes Zoom, Microsoft Teams, and Google Meet via `PLATFORM_URL_PATTERNS`, plus a phone-number fallback — all human-mediated calls with the same underlying prep needs.

AI-interviewer platforms — where the candidate has a live conversation with an AI system instead of a human panel (Alex/Apriora, HireVue) — aren't covered at all, even though they warrant materially different prep:

- No human rapport/reading-the-room available — camera eye contact and pacing matter more, not less.
- The AI generates adaptive follow-up questions from the literal content of what the candidate says, so vague answers get penalized differently than with a human interviewer.
- Alex/Apriora specifically (post-rebrand from a 2024 public glitch incident) has documented failure modes — transcription-quality degradation and conversational loops — worth flagging to the candidate as a known platform quirk rather than something to self-blame for mid-call.

None of this was distinguishable from a Zoom call — a candidate prepping from `interview-prep.md` output had no signal the round was AI-mediated at all.

## What changed

**`invite-match.mjs`**
- Added `Alex` (`meet.alex.com`, `alex.com`) and `HireVue` (`hirevue.com`) entries to `PLATFORM_URL_PATTERNS`, anchored the same way as the existing Zoom/Teams/Meet patterns — rejects lookalike hosts, email addresses, path segments, and query values as false positives (same discipline #2128's review rounds hardened for the existing three).
- Added a new exported helper, `isAIInterviewerPlatform(text)` → `boolean`. **`extractPlatform()`'s existing return contract is untouched** (still a plain platform-name string or `null`) — existing callers keep working unmodified. Alex/HireVue just report their own name from `extractPlatform()`, same convention as Zoom/Teams/Meet.
- Surfaced `isAIInterviewer` as an additive field in `analyzeInvite()`'s `signals` object.

**`modes/interview-prep.md`**
- Wired an **AI-Interviewer Notes** block into Step 2 (Process Overview), gated on `isAIInterviewerPlatform`, following the same cross-reference-not-restate pattern #2128 used for the original Platform field (points at `invite-match.mjs` rather than duplicating detection logic in prose).
- Step 3 (Round-by-Round Breakdown) points a per-round logistics line at the notes block instead of the video-platform camera/lighting reminder when that round is AI-mediated.

**Tests**
- `invite-match.test.mjs`: positive cases for Alex (`meet.alex.com` and bare `alex.com`) and HireVue, `isAIInterviewerPlatform` true/false cases, lookalike-host/path-segment/query-value false-positive guards mirroring the existing Zoom/Teams/Meet ones, and explicit confirmation that the existing Zoom/Teams/Meet/Phone detection paths are unaffected.
- Same coverage mirrored in `invite-match.mjs`'s `--self-test` block, consistent with how the existing platform tests are duplicated across both.

## Verification

```
node invite-match.mjs --self-test   → 110 passed, 0 failed
node invite-match.test.mjs          → 80 passed, 0 failed
node test-all.mjs                   → 3276 passed, 0 failed, 1 warning (pre-existing, unrelated: cv-sync-check expected without user data)
```

## Test plan
- [x] `node invite-match.mjs --self-test`
- [x] `node invite-match.test.mjs`
- [x] `node test-all.mjs`

## Summary by CodeRabbit

* **New Features**
  * Added detection for Alex and HireVue interview platforms.
  * Interview preparation now provides AI-specific guidance for confirmed AI-led interviews.
  * HireVue invitations receive neutral platform guidance while retaining standard video logistics.
  * Invite analysis now identifies confirmed AI-interviewer invitations.

* **Bug Fixes**
  * Improved platform detection accuracy by rejecting lookalike domains, email addresses, paths, and query values.

* **Tests**
  * Added coverage for platform detection, classification, and existing interview-platform behavior.

### DIFF
--- a/invite-match.mjs
+++ b/invite-match.mjs
@@ -313,6 +313,19 @@ export function extractReqId(text) {
 // extractPlatform()'s return contract (still a plain platform-name string);
 // it's read separately by isAIInterviewerPlatform() below so existing
 // extractPlatform() callers are unaffected.
+//
+// This flag is only ever set `true` when the *host alone* reliably confirms
+// AI-led modality for every invite that host can appear in. That is true
+// for Alex/Apriora (a single-purpose AI-interviewer product — there is no
+// human-conducted call on that host) but NOT for HireVue: HireVue is a
+// multi-modal platform whose `hirevue.com` hosts serve on-demand recorded
+// screening, live human-conducted interviews scheduled through the
+// platform, and a separate "AI Interviewer" product, all under the same
+// domain. No public discriminator (subdomain, path) reliably distinguishes
+// which modality a given hirevue.com invite link is for, so a HireVue match
+// stays `isAIInterviewer: false` here — HireVue is still detected and named
+// by extractPlatform(), it just isn't asserted as confirmed-AI. If HireVue
+// later exposes a reliable discriminator, promote its entry then.
 const PLATFORM_URL_PATTERNS = [
   { name: 'Zoom', pattern: /(?:^|[^\w@./?=&#-])(?:https?:\/\/)?(?:[\w-]+\.)?zoom\.us(?::\d{1,5})?(?:[/?#\s]|$)/i, isAIInterviewer: false },
   { name: 'Microsoft Teams', pattern: /(?:^|[^\w@./?=&#-])(?:https?:\/\/)?(?:[\w-]+\.)?teams\.(?:microsoft|live)\.com(?::\d{1,5})?(?:[/?#\s]|$)/i, isAIInterviewer: false },
@@ -320,10 +333,13 @@ const PLATFORM_URL_PATTERNS = [
   // Alex/Apriora — AI-interviewer platform, post-rebrand from a 2024 public
   // glitch incident. `meet.alex.com` and bare `alex.com` are both real
   // invite hosts; the optional-subdomain shape already covers both, same as
-  // the Zoom pattern above.
+  // the Zoom pattern above. Always AI-led by product design.
   { name: 'Alex', pattern: /(?:^|[^\w@./?=&#-])(?:https?:\/\/)?(?:[\w-]+\.)?alex\.com(?::\d{1,5})?(?:[/?#\s]|$)/i, isAIInterviewer: true },
-  // HireVue — AI-interviewer / on-demand video-screening platform.
-  { name: 'HireVue', pattern: /(?:^|[^\w@./?=&#-])(?:https?:\/\/)?(?:[\w-]+\.)?hirevue\.com(?::\d{1,5})?(?:[/?#\s]|$)/i, isAIInterviewer: true },
+  // HireVue — multi-modal video-screening platform (on-demand recorded,
+  // live human-conducted, and a separate AI Interviewer product all share
+  // this domain). Detected and named, but NOT asserted as confirmed-AI —
+  // see the comment above the array.
+  { name: 'HireVue', pattern: /(?:^|[^\w@./?=&#-])(?:https?:\/\/)?(?:[\w-]+\.)?hirevue\.com(?::\d{1,5})?(?:[/?#\s]|$)/i, isAIInterviewer: false },
 ];
 
 // A plain phone number, used only when no meeting-platform URL was found —
@@ -357,14 +373,22 @@ export function extractPlatform(text) {
 }
 
 /**
- * Whether the call platform detected in `text` is an AI-interviewer
+ * Whether the call platform detected in `text` is a CONFIRMED AI-interviewer
  * platform (#2673) — the candidate has a live conversation with an AI
- * system rather than a human panel (Alex/Apriora, HireVue). Deliberately a
- * separate boolean helper rather than a change to extractPlatform()'s
- * return contract, so existing callers (interview-prep.md's Platform field,
- * analyzeInvite() below) keep working unmodified. Same "never guessed"
- * discipline as extractPlatform(): pure pattern matching against the same
- * PLATFORM_URL_PATTERNS table, false when nothing plausible is found.
+ * system rather than a human panel. Currently true only for Alex/Apriora,
+ * whose host is single-purpose and always AI-led by product design.
+ * HireVue is deliberately excluded even though it's detected by
+ * extractPlatform(): it's a multi-modal platform (on-demand recorded,
+ * live human-conducted, and a separate AI Interviewer product all share the
+ * `hirevue.com` domain) with no public discriminator that reliably tells
+ * which modality a given invite link is for, so asserting AI-led for every
+ * HireVue match would be a guess, not a detection — see the comment above
+ * PLATFORM_URL_PATTERNS. Deliberately a separate boolean helper rather than
+ * a change to extractPlatform()'s return contract, so existing callers
+ * (interview-prep.md's Platform field, analyzeInvite() below) keep working
+ * unmodified. Same "never guessed" discipline as extractPlatform(): pure
+ * pattern matching against the same PLATFORM_URL_PATTERNS table, false when
+ * nothing plausible — or nothing *confirmed* — is found.
  *
  * @param {string} text - Raw pasted invite/scheduling text.
  * @returns {boolean}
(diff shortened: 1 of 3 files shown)


santifer:
Thanks for this, @Schlaflied.

You add Alex (meet.alex.com, alex.com) and HireVue to PLATFORM_URL_PATTERNS, anchored the same way as the existing three, rejecting lookalike hosts, email addresses, path segments and query values, the same anti-false-positive discipline the review rounds on #2128 hardened for Zoom/Teams/Meet. The good design call: instead of changing extractPlatform()'s contract you export a new helper, isAIInterviewerPlatform(text), so every current caller keeps working untouched, and isAIInterviewer lands as an additive field in analyzeInvite()'s signals. In interview-prep.md the AI-interviewer notes block is wired at step 2 pointing to invite-match.mjs rather than duplicating the detection logic in prose, step 3 swaps the camera/lighting reminder for the right line when the round is with an AI, and 110 + 80 + 3,276 tests come back green.

santifer:
Detection itself is clean, @Schlaflied, and I checked it by running rather than reading: `meet.alex.com`, `alex.com` and `app.hirevue.com` resolve, Zoom/Teams/Meet/Phone are byte-identical before and after, and your guards hold against every trap I threw at them (`notalex.com`, `support@alex.com`, `example.com/alex.com`, `example.com?next=alex.com`, `alex.com.br` all return null). Twelve unknown platforms degrade to "human", which is the safe direction.

One real bug, and it only shows up with multi-link text.

`isAIInterviewerPlatform()` scans independently of `extractPlatform()`, so **an invite containing a Zoom link AND an `alex.com` link flags as AI while the platform resolves to Zoom.** That is the wrong direction of failure: a human round mislabeled as a robot interview changes how the candidate prepares, and nothing in the suite covers multi-link bodies, which is exactly why it passes today.

And the case is not contrived. Alex sells "Coordinator", which schedules interviews with human hiring managers and handles the calendar invites and reminders. So an `alex.com` link inside an invitation to a human round is the product working as intended, not an edge case.

Which also undermines the justification written at `invite-match.mjs:319-320`: *"a single-purpose AI-interviewer product — there is no human-conducted call on that host"*. That is not true of Alex. Worth flagging that it is the same multi-modality argument CodeRabbit used to downgrade HireVue to `isAIInterviewer: false`: the rule was applied to one and not the other.

**What I would change (a is verified working, I applied it to a copy and re-probed):**

- **(a)** `invite-match.mjs:398-400` — derive from the same match that `extractPlatform()` picks:
  ```js
  for (const { pattern, isAIInterviewer } of PLATFORM_URL_PATTERNS) {
    if (pattern.test(text)) return isAIInterviewer === true;
  }
  ```
  Result: Zoom+alex.com and Teams+alex.com become `isAI=false`, everything else is unchanged, and your 80 executable assertions stay green.
- **(b)** a regression test for multi-link text: human link plus `alex.com` in one body, asserting the human platform and `isAIInterviewerPlatform === false`.
- **(c)** fix the reasoning at `:319-320`. The flag can stay `true` for a clean Alex match, but the written reason has to hold: today it claims something about the host that is not the case.
- **(d)** align the title and body with what ships: HireVue is detected and named, not asserted as AI.

This is the same class as something else I hit today, and it is worth naming because it keeps recurring: **sharing a table is not sharing a decision when two functions read it independently.** Deriving one from the other is what makes them agree by construction instead of by discipline.

Schlaflied:
@santifer Thanks for the detailed review — pushed a fix in 4556f11.

**Fix (a):** `isAIInterviewerPlatform()` now iterates `PLATFORM_URL_PATTERNS` in the exact same fixed order `extractPlatform()` uses and returns the flag off that first match, instead of scanning independently for any AI-flagged pattern anywhere in the text:

```js
for (const { pattern, isAIInterviewer } of PLATFORM_URL_PATTERNS) {
  if (pattern.test(text)) return isAIInterviewer === true;
}
```

So a body with a Zoom link **and** an `alex.com` link now resolves the platform to Zoom and `isAIInterviewerPlatform()` correctly returns `false` for that invite — the two functions agree by construction, not by coincidence.

**Fix (b):** Added a multi-link regression test in `invite-match.test.mjs` covering both Zoom+alex.com and Teams+alex.com bodies, asserting `extractPlatform()` resolves to the human platform and `isAIInterviewerPlatform()` returns `false`.

**Fix (c):** Reworded the comment above `PLATFORM_URL_PATTERNS` — it previously claimed "there is no human-conducted call" possible on the Alex host, which is false given Alex's "Coordinator" product schedules and sends invites for human-conducted rounds too. The comment now describes this as a best-effort heuristic tied to `extractPlatform()`'s own match resolution, not a guarantee about host exclusivity.

**Fix (d):** HireVue's treatment is unchanged from what's already in the code/comments — detected and named by `extractPlatform()`, never asserted as confirmed-AI, same as before this fix.

Full suite: `node test-all.mjs --quick` → 3275 passed, 0 failed (1 pre-existing warning, unrelated to this change). `invite-match.test.mjs` alone: 89 passed, 0 failed.

santifer:
All four landed, and (a) is the one that matters: the two functions now agree by construction instead of by coincidence, which is the difference between a fix and a patch.

Verified by running it against three points rather than reading the diff. Your previous push resolves `Zoom + alex.com` to Zoom while flagging `isAI=true`, the same for Teams and Meet, so the disagreement was real and reproducible: 3 of 6 probe cases wrong. This push is 6 of 6, with single-link Alex still `true` and HireVue still `false`. `invite-match.test.mjs`: 89 passed, exit 0. Suite on a preview against today's main: 3756 passed, 0 failed.

Worth naming (c) separately, because rewriting a comment to say something *weaker* about your own feature is the least rewarded thing in a PR and you did it without argument. The new block states the heuristic honestly, explains the Coordinator case, and cites where the correction came from.

One leftover that is mine, not yours: the jsdoc on `isAIInterviewerPlatform` still says Alex's host is "single-purpose and always AI-led by product design", which now contradicts the block you rewrote 60 lines above it. My ask pointed you at `:319-320` and that is exactly what you fixed; I just aimed at one of the two places the sentence lived. Pushing that one myself rather than charging you a third lap.

Merged. 🚀

PR #2678, opened 2026-08-11
feat: zero-LLM CV title-consistency check for pdf.md tailoring

### LINKED ISSUE #2677: pdf mode: no check that a tailored CV's job titles still match cv.md's canonical titles
### Code of Conduct

### Existing issues
- [x] I searched existing issues and this hasn't been reported yet (related but distinct: `cv-sync-check.mjs` validates project *setup* integrity — cv.md exists, no hardcoded metrics in prompt files — not whether a *generated, tailored* CV still agrees with cv.md's own content)

### What happened / motivation

`pdf.md`'s tailoring pipeline reorders bullets, reframes the summary, and adjusts framing per role archetype (Steps 10-15) — legitimate reformulation per the Data Contract's "keywords get reformulated, never fabricated" rule. But nothing currently checks whether a job **title** survives that reformulation unchanged. In practice, a tailored CV can end up stating a more senior-sounding title for a past role than the one recorded in `cv.md` — e.g. cv.md says "Administrative Assistant," a JD-tailored version for a more senior-sounding target role says "Operations Lead" or similar for the exact same job/dates/company.

This has surfaced repeatedly across real interview rounds (tracked informally in interview-prep notes, not in this repo) as a real risk: if an interviewer asks "what was your actual title," a mismatch against what's on file elsewhere (LinkedIn, a prior submitted CV, background-check records) is a credibility problem discovered live, not caught in advance. The tailoring itself is often reasonable (describing the same duties in language suited to the target role), but the specific *title string* silently drifting from cv.md's canonical title is a distinct, checkable failure mode — separate from and narrower than legitimate duty/keyword reformulation.

### Proposal

- A small, zero-LLM checker (similar shape to `jd-skill-gap.mjs`): for each `{Company, Dates}` entry in a generated/tailored CV, extract the stated title and compare it against the title recorded for that same `{Company, Dates}` pair in `cv.md`. Exact match → pass silently. Mismatch → warn (not block): surface both strings side by side so the user can decide it's intentional framing vs. an unintended drift, per this project's existing warn-only, human-in-the-loop philosophy elsewhere (Block G, blacklist suggestions, etc.).
- Wire it into `pdf.md`'s step list as an explicit check step (after tailoring, before finalizing — similar to how the Step 4 skill-gap check is already wired in and required to be surfaced to the user before drafting continues), not as a separate manual command the user has to remember to run.
- Keep matching simple/deterministic: match on company name + date range (reuse whatever normalization `cv-sync-check.mjs`/other scripts already use for company-name comparison if applicable) to pair up the right entries; don't attempt fuzzy title-similarity scoring — an exact-string mismatch is the only thing that should warn, since the goal is catching drift, not judging whether a title *sounds* inflated.

Add regression tests: a tailored CV with an unchanged title → no warning; a tailored CV with a drifted title for the same company/dates → warning surfaced with both strings.

Schlaflied:
## The problem

`pdf.md`'s tailoring pipeline (Steps 10-15) legitimately reformulates bullets, reframes the summary, and adjusts framing per JD — "keywords get reformulated, never fabricated" per the Data Contract. Nothing checked whether a job **title** survives that reformulation unchanged from `cv.md`'s own canonical title for the same job. A tailored CV can end up stating a more senior-sounding title than `cv.md`'s for the exact same `{company, dates}` — a distinct, narrower failure mode than legitimate duty reformulation, and a real risk if ever cross-checked (LinkedIn, a prior CV, a background check, or an interviewer asking directly).

Related but distinct: `cv-sync-check.mjs` validates project *setup* integrity, not whether a *generated, tailored* CV still agrees with `cv.md`'s content.

## The fix

`cv-title-check.mjs` — zero-LLM, deterministic, same shape as `jd-skill-gap.mjs` (`--summary` flag, `--self-test`, warn-only, never edits anything).

**Input format:** `modes/pdf.md` Step 17 already builds a compact JSON render payload (`{ experience: [{ company, role, dates, ... }] }`) and writes it to `/tmp/cv-{candidate}-{company}.json` *before* any HTML or PDF exists. That's the earliest point in the pipeline with company/title/dates as clean structured fields — checking it means no HTML/PDF text-scraping to maintain, and the check runs before the fact gate and PDF render rather than after.

**Matching:** each tailored-CV entry pairs with a `cv.md` entry by `{company, dates}` — normalized on case/whitespace only via the same `normalizeCompany()` (`tracker-utils.mjs`) the rest of the tracker ecosystem already uses for company grouping. Title is the field being checked, so it can never also be the matching key. Exact string match (case/whitespace-normalized) → silent pass. Any other difference → warning with both strings printed side by side. No fuzzy/similarity scoring, per the issue's explicit guidance — the goal is catching drift, not judging whether a title *sounds* inflated. Entries that can't be paired (no matching `{company, dates}` in `cv.md`) are reported separately as unmatched, never forced into a false mismatch.

**Wiring:** new Step 17a in `modes/pdf.md`, between building the JSON payload (Step 17) and rendering HTML (Step 18) — mirrors how Step 4's skill-gap check is already required to be surfaced to the user before drafting continues. Chose a lettered sub-step instead of renumbering 18-22, since those numbers are cross-referenced from `modes/pdf.md` itself and `modes/pdf/hm-audit.md` in several places.

## Tests

`cv-title-check.mjs --self-test` covers the three cases from the issue plus edge cases:
- unchanged title → no warning
- same company/dates, drifted title → warning with both strings
- company-name formatting differences (extra whitespace, "Inc." variant) still pair correctly via `normalizeCompany`
- an entry with no `cv.md` counterpart is reported as unmatched, never as a false mismatch
- case/whitespace-only title differences are not a mismatch

Registered in `test-all.mjs` (`cv-title-check.mjs --self-test`) and `update-system.mjs`'s `SYSTEM_PATHS`. Added a row to `AGENTS.md`'s Main Files table.

`node test-all.mjs --quick` → **3277 passed, 0 failed** (1 pre-existing Node SQLite experimental-feature warning, unrelated). `validate-system-paths-coverage.mjs --self-test` → clean.

Closes #2677

## Summary by CodeRabbit

* **New Features**
  * Added a title-consistency checker comparing tailored CV experience titles with canonical CV entries.
  * Supports human-readable and JSON reports for matches, mismatches, ambiguous matches, and unmatched entries.
  * Added pre-render PDF checks with warnings for title inconsistencies; rendering continues after review.
  * Invalid input or checker failures now stop processing.

* **Tests**
  * Added automated self-tests and included them in the full test suite.

* **Documentation**
  * Documented the new checker among the project’s main files.


santifer:
Thanks for this, @Schlaflied.

cv-title-check.mjs, zero-LLM and warn-only, hooked at the one point in the pipeline where the data is already clean: step 17 of pdf.md builds a JSON render payload with {company, role, dates} and writes it before any HTML or PDF exists, so the check runs pre-render and there's no HTML scraping to maintain. It pairs on {company, dates} with the same normalizeCompany() the rest of the tracker ecosystem uses, and makes explicit that the title, being the field under test, can never be the pairing key; an exact match passes silently, anything else prints both strings side by side, and unpaired entries are reported as unmatched instead of being forced into a false mismatch. You picked Step 17a rather than renumbering 18-22 because those numbers are referenced from pdf.md and from pdf/hm-audit.md.

PR #2682, opened 2026-08-11
fix(providers): make themuse resilient to per-page fetch failures

### LINKED ISSUE #2681: themuse provider: one failed page discards every job gathered from the board
> ⚠️ Privacy note: this report contains no personal data — API structure and public job-board behavior only.

### Code of Conduct

### Existing issues

### What happened?

`providers/themuse.mjs` fetches The Muse's public jobs API by walking pages sequentially (`?page=0`, `?page=1`, ... up to a 100-page cap). Each page fetch is a single, un-retried `ctx.fetchJson()` call with a 10s timeout (`providers/_http.mjs`'s `DEFAULT_TIMEOUT_MS`) and no partial-failure tolerance.

If any single one of those up to ~100 sequential requests fails or times out — a transient rate-limit (429), a 5xx, or the request simply stalling past 10s — the `fetch()` call throws, and **every job already gathered from earlier pages in that run is discarded**. The board returns zero results for the whole scan, which reads to the user as "the URL isn't working," when the actual endpoint is healthy and only one request out of many had a bad moment.

This is the same class of bug `workday.mjs` and `oraclecloud.mjs` already solved with a retry/backoff wrapper around their per-page fetches — `themuse.mjs` never got the same treatment.

### Steps to reproduce

1. Run a scan that includes a `job_boards:` entry with `provider: themuse` (see `templates/portals.example.yml`).
2. Simulate (or naturally hit, under real-world rate-limiting) a single failed/timed-out request on any page after page 0 — e.g. mock `ctx.fetchJson` to throw on the second call.
3. Observe that `themuse.fetch()` throws and the entire board's results — including all pages fetched successfully before the failure — are lost.

Minimal repro against the current provider:

```js
import themuse from './providers/themuse.mjs';

let calls = 0;
await themuse.fetch({ name: 'The Muse' }, {
  fetchJson: async (url) => {
    calls++;
    const page = new URL(url).searchParams.get('page');
    if (page === '0') return { results: [{ name: 'Job A', refs: { landing_page: 'https://www.themuse.com/jobs/a' }, company: {}, locations: [] }], page: 0, page_count: 3 };
    throw new Error('HTTP 503'); // one bad page among many
  },
});
// -> throws, even though page 0 succeeded and had a real job in it
```

### Expected behavior

A single failed page (out of up to 100 sequential fetches) should not discard results already gathered from other pages. The fetch should:
- Retry transient failures (429 / 5xx / network-timeout) with backoff, honoring `Retry-After` when the server supplies one.
- Fail fast (no retry) on non-retryable 4xx errors.
- If a page still fails after retries, log a warning and return whatever pages were already collected, rather than throwing and losing everything.

This mirrors the existing pattern in `providers/workday.mjs` (`fetchPageWithRetry`, `isRetryableError`, `parseRetryAfterMs`) and `providers/oraclecloud.mjs`.

### CLI tool

Claude Code

### OS and Node.js version

macOS 25.6.0 / Node v26.0.0

clede:
## What does this PR do?

Makes `providers/themuse.mjs` resilient to a single failed page during its up-to-100-page sequential walk of The Muse's public jobs API. Today, one transient failure (rate-limit, 5xx, or a stalled request past the 10s timeout) throws and discards every job already gathered from earlier pages — the whole board returns zero results for that scan, which reads as "the URL isn't working" even though the endpoint is healthy.

Adds the same retry/backoff pattern already used by `workday.mjs` and `oraclecloud.mjs`: retry transient failures (429/5xx/timeout) with backoff honoring `Retry-After`, fail fast on non-retryable 4xx, and truncate-with-a-warning instead of throwing when a page still fails after retries — returning whatever was already collected.

## Related issue

Fixes #2681

## Type of change

- [x] Bug fix

## Checklist

- [x] Bug fix — no issue required, opened one anyway for repro clarity
- [x] I ran `node test-all.mjs` and all tests pass (3281 passed, 0 failed, 2 pre-existing unrelated warnings, verified in a clean worktree off `upstream/main`)
- [x] My changes respect the Data Contract (no modifications to user-layer files — this PR touches only `providers/themuse.mjs` and its test file)

## Test plan

- Added 8 new assertions to `tests/providers/themuse.test.mjs`: retry-then-recover (429 + Retry-After honored), retry-exhaustion preserves earlier pages' results and warns instead of throwing, non-retryable errors fail fast without wasting attempts.
- All 16 original assertions still pass unchanged.
- Verified live against the real API (`node scan.mjs --company "muse" --dry-run`) before and after — identical job counts, no behavior regression.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved job retrieval reliability during temporary service errors.
  * Added capped retries with support for rate-limit retry delays.
  * Preserves previously retrieved jobs when later pages fail.
  * Prevents excessively long pagination and handles rate-limit responses more gracefully.
  * Reports incomplete results when retries are exhausted after earlier pages succeed.

* **Tests**
  * Added coverage for retry recovery, rate limits, exhausted retries, non-retryable errors, complete failures, and partial results.


santifer:
Nice work, @clede — the retry shape mirrors `workday.mjs` and `oraclecloud.mjs` closely, and propagating the real `err.attempts` instead of a fixed count is a genuine improvement over the provider you copied from (workday prints `after ${MAX_RETRIES + 1} attempts` unconditionally, which lies on a non-retryable error). The tests are behavioural rather than boilerplate. One change before I merge:

**The tolerant catch wraps page 0 as well, so a completely dead board now returns `[]` instead of throwing.** The PR says `scan.mjs`'s empty-board reporting covers that case, but it doesn't: `scan.mjs` treats `empty` as a *healthy* state and **resets the streak** in `computeConsecutiveFailures`. So a themuse outage would be recorded as a healthy scan, and the consecutive-failure detector — the thing that tells us a provider died — goes blind on exactly the provider it should be shouting about.

`workday.mjs`, the pattern you cite, does the opposite on purpose: page 0 is fetched *outside* the tolerant loop (the loop starts at `page = 1`), so a first-page failure still surfaces as a board failure. Same here: pull page 0 out of the catch (or re-throw when `page === 0`), plus a test that a total page-0 failure still throws.

Separately, and **not blocking**: this adds a third local copy of `sleep`/`parseRetryAfterMs`/`isRetryableError`/`fetchPageWithRetry`, which already exist exported from `providers/_http.mjs`. It's consistent with what workday and oraclecloud do, so I'm not asking you to change it here — but they have already drifted (the `_http.mjs` version clamps the backoff by subtracting jitter, the local copies don't). That's a debt issue of its own, and I'll open it.

Fix the page-0 case and I'll merge this the same day.

ilyakanevskiy:
One more issue in the same function, independent of the page-0 problem above: the malformed-shape check still throws unconditionally, bypassing the new truncate-and-preserve path entirely.

```js
if (!json || !Array.isArray(json.results)) {
  throw new Error(`themuse: unexpected API response on page ${page} ...`);
}
```

This runs *after* `fetchPageWithRetry` already returned successfully (200 OK, no retries triggered) — it's a different failure mode than the 429/5xx/timeout cases the retry loop handles. If page 5 of 10 comes back with an unexpected shape, everything gathered from pages 0-4 is still discarded, which is the exact failure mode this PR sets out to fix, just reached via a bad body instead of a bad status.

Other providers in this codebase treat an unexpected shape as "no jobs on this page" rather than a hard failure — `providers/greenhouse.mjs:137` and, closer to home, `providers/getro.mjs` (`Array.isArray(results.jobs) ? results.jobs : []`) are the reference pattern already used elsewhere — worth mirroring here rather than throwing.

(Checked the other paginating+retry providers for both this and the page-0 issue above — `workday.mjs`, `oraclecloud.mjs`, `getro.mjs`, `a16z-speedrun-talent.mjs` — none of them reproduce either bug. `getro.mjs` in particular already implements both fixes correctly and might be a useful second reference alongside workday/oraclecloud.)

santifer:
Two things, @clede.

**First, the debt issue I promised is open: #2709.** It turned out to be more your defence than your homework. Asking you to "mirror the existing retry shape" was unfair advice, because the shared helper (`providers/_http.mjs`) only wraps `fetchJson` and has no way to express "this `Retry-After` is too long, bail out". A provider like this one genuinely cannot import it today. That gap is what #2709 tracks, and it is on me, not on you.

**Second, @ilyakanevskiy's review above is correct and I verified it myself** on your current head (`cf6f988`):

```
providers/themuse.mjs:170
      if (!json || !Array.isArray(json.results)) {
        throw new Error(`themuse: unexpected API response on page ${page} ...`);
```

That throw sits inside the page loop, after `allResults` already holds pages 0 through N-1, so a bad body on page 5 of 10 still discards everything gathered before it. Same lost-work failure this PR exists to fix, reached through a different door.

Where I would tighten his suggestion slightly: `Array.isArray(x) ? x : []` is the right pattern for a single-page board (`greenhouse.mjs:137` and `getro.mjs:95` both use it, as he says), but applied to *every* page here it recreates the problem I flagged in my first review, one level down. If The Muse ever changes its response shape, silently mapping every page to `[]` reports a healthy board with zero jobs, and nothing downstream can tell that apart from a company that stopped hiring.

So the same page-0 split that fixes the retry path fixes this one:

- **page 0 malformed** stays a hard throw. No evidence the board works, so failing loudly is correct.
- **page N>0 malformed** truncates and preserves, with a warning naming the page. You already proved the board answers properly at least once, so the pages you hold are real.

That is one condition, not a rewrite, and it lands both of his points and mine in the same shape.

@ilyakanevskiy, thank you for this. You went and checked `workday`, `oraclecloud`, `getro` and `a16z-speedrun-talent` for both failure modes before posting, which is the part most reviews skip, and the negative result is what told me this is a themuse-only bug and not a class. That is reviewer work, and it is noticed.

santifer:
The page-0 finding from yesterday is properly resolved, @clede: `cf6f9886` pulls page 0 out of the tolerant loop and it now throws instead of returning `[]`, matching the workday pattern you cited. Verified by execution against a transport that 503s every page: the old push returned an empty array, this one throws. And your regression tests pin all three properties. Good fix.

Blocking on something new that the fix introduced.

**The local `isRetryableError` copy is missing the SSRF guard the shared one has.** `providers/_http.mjs:140` carries:

```js
if (status === undefined && err instanceof TypeError && err?.cause?.message === REDIRECT_REFUSAL_CAUSE_MESSAGE) return false;
```

That line exists so a rejected redirect - the `redirect: 'error'` guard from #1440 doing its job - fails fast instead of being retried. The copy at `themuse.mjs:46-51` drops it. Measured with the same `TypeError('fetch failed')` carrying `cause.message='unexpected redirect'`:

| classifier | calls to fetchJson |
|---|---|
| shared `_http.mjs` | **1** |
| this PR's local helper | **4** |

It is a regression against `main`, which had no retry at all, so a rejected redirect cost one attempt and now costs four against a host that is redirecting. No SSRF hole opens (the redirect is still refused, never followed), but `tests/providers/_http.test.mjs:76-94` exists specifically to protect that behavior, and the local copy is outside its reach.

**The ask is one import.** Seven providers already use `fetchJsonWithRetry` from `providers/_http.mjs`. The only other one with a local retry loop, `workable.mjs:38`, still imports the shared `isRetryableError` and `parseRetryAfterMs`, and documents at `:54-61` why its loop is local (it needs to give up on a long `Retry-After` and wraps a text call too). themuse needs neither.

So: `import { fetchJsonWithRetry } from './_http.mjs'` with `{ retries: 3 }` (the `workday.mjs:36` pattern), and delete the duplicated helpers at `themuse.mjs:32-92`. `retries: 3` preserves the 4 attempts your own tests assert, so nothing you wrote gets invalidated.

One small drift that disappears with the same change: your backoff at `:76-77` does `Math.min(500*2**n, 8000) + Math.random()*250`, which can exceed its own `RETRY_MAX_DELAY_MS` by 250ms. The shared one subtracts the jitter from the ceiling first.

Worth naming the pattern, because it caught me on a different PR today too: **copying an implementation is not sharing a contract.** The copy was correct the day it was written, and the guard it was missing had been added to the original for a reason nobody would guess from reading the copy.

santifer:
The framing in your comment is what sold this, @clede:

> one stalled request out of up to 100 sequential page fetches throws and discards every job already gathered from this board — the whole board reads as "not working" when only one page had a bad moment

That's the bug that never gets reported, because from outside it looks like the provider is broken rather than flaky, and the user's conclusion is "themuse doesn't work" instead of "page 47 timed out". Mirroring `workday.mjs` / `oraclecloud.mjs` rather than inventing a third retry shape is the right instinct — three providers that retry three different ways is how the next person gets it wrong.

`INTER_PAGE_DELAY_MS` matters as much as the retry itself: a 100-page walk firing as a burst is what earns the 429s the retry then has to absorb.

Merged 🚀

One heads-up rather than a request: #2795 is an open PR consolidating the four copies of `sleep()` into a shared helper, and yours makes it five. Your `ctx?.sleep` injection seam is the right shape and I'd expect it to survive that refactor — but if you see that PR land and your version drift, that's why. I've flagged the new copy over there so it doesn't get missed.

PR #2683, opened 2026-08-11
fix(oraclecloud): accept Oracle's numbered tenant apexes (oraclecloud1-99.com)

LayerC0de:
## What does this PR do?

`ORACLE_HOST_RE` in `providers/oraclecloud.mjs` only matches the `oraclecloud.com` apex. Oracle issues newer Fusion tenants a **numbered** apex (`oraclecloud26.com` is live in the wild), so those boards fail `detect()` and the scanner returns **zero jobs with no error and no warning** — the worst kind of failure, because the board looks empty rather than broken.

This widens the host pin to a **bounded** apex family and adds tests for the new shapes and their spoofs.

## The fix

```diff
-const ORACLE_HOST_RE = /^[a-z0-9-]+\.fa\.(?:[a-z0-9-]+\.)?(?:ocs\.)?oraclecloud\.com$/i;
+const ORACLE_HOST_RE = /^[a-z0-9-]+\.fa\.(?:[a-z0-9-]+\.)?(?:ocs\.)?oraclecloud(?:[1-9][0-9]?)?\.com$/i;
```

`oraclecloud(?:[1-9][0-9]?)?` = the unnumbered apex plus `oraclecloud1.com` … `oraclecloud99.com`. **No leading zero, at most two digits.**

This is deliberately *not* `oraclecloud\d*\.com` or anything wildcard-ish. That regex is an SSRF pin — it is the only thing standing between a `portals.yml` entry and an arbitrary outbound fetch — so it stays an enumeration of a finite host family. `oraclecloud0`, `oraclecloud01`, `oraclecloud100` and `oraclecloudanything` are all still rejected, and there are tests asserting each.

## Verification

Ran the provider's real `fetch()` against four live tenants, once with the current `main` provider and once with the patched one. Same harness, same run:

| tenant | `main` | patched |
| --- | --- | --- |
| `*.fa.ocs.oraclecloud.com` (A) | 7 jobs | 7 jobs |
| `*.fa.ocs.oraclecloud.com` (B, `CX_3001`) | 3 jobs | 3 jobs |
| `*.fa.ocs.oraclecloud.com` (C) | 23 jobs | 23 jobs |
| `*.fa.ocs.oraclecloud26.com` | **`detect()` → null, 0 jobs** | **13 jobs** |

No change on the unnumbered apex; the numbered tenant goes from silently empty to its real 13 postings. (Tenant hostnames are from my own `portals.yml` so I've left them out of a public body — happy to send them if you want to re-run it.)

Negative control per the usual discipline: the two new assertions were run against unpatched `main` first and fail there —

```
❌ numbered apex rejected for shape(s): plain, <region>, .ocs.
❌ apex boundary wrong for: oraclecloud1, oraclecloud9, oraclecloud26, oraclecloud99
📊 Results: 37 passed, 2 failed
```

— and pass with the fix (39/0).

## Tests added

In `tests/providers/oraclecloud.test.mjs` (existing file, auto-discovered):

- the numbered apex across **all three** existing host shapes: `<tenant>.fa.`, `<tenant>.fa.<region>.`, `<tenant>.fa.ocs.`
- the 1–99 boundary: `oraclecloud`, `1`, `9`, `26`, `99` accepted; `0`, `01`, `100`, `26x`, `abc` rejected
- numbered apex used as a **label of a hostile host** (`x.fa.ocs.oraclecloud26.evil.example`) → null
- numbered apex in the **path** rather than the host → null
- numbered apex in URL **userinfo** (before the `@`) → null
- right label, **wrong TLD** (`oraclecloud26.net`) → null
- **plaintext** `http://` on the numbered apex → null

## Docs

`templates/portals.example.yml` documented the host pattern in two places (the provider index and the ORC example block); both now show `oraclecloud[1-99].com` and note that newer tenants get the numbered apex.

Unrelated gap I noticed but left alone to keep this scoped: `docs/SUPPORTED_JOB_BOARDS.md` has no Oracle Recruiting Cloud row at all. Happy to add one in a separate PR if useful.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] I ran `node test-all.mjs` and all tests pass — **3280 passed, 0 failed** (2 warnings, both pre-existing "expected without user data" fixture notes)

## Summary by CodeRabbit

- **New Features**
  - Oracle Cloud hostname validation now supports numbered domains from `oraclecloud1.com` through `oraclecloud99.com`.
  - Site-number detection and manual overrides continue to work for numbered domains.

- **Bug Fixes**
  - Improved hostname rejection for unsupported, spoofed, malformed, or insecure Oracle Cloud addresses.

- **Documentation**
  - Updated Oracle Cloud setup guidance with examples of numbered domains.

### DIFF
--- a/providers/oraclecloud.mjs
+++ b/providers/oraclecloud.mjs
@@ -10,6 +10,14 @@
 //   <tenant>.fa.oraclecloud.com
 //   <tenant>.fa.<region>.oraclecloud.com   (e.g. us2)
 //   <tenant>.fa.ocs.oraclecloud.com
+//   ...and the same three shapes on the numbered apex `oraclecloud<NN>.com`
+//   that Oracle issues to newer tenants (observed live: a Fusion board on
+//   `<tenant>.fa.ocs.oraclecloud26.com` whose unnumbered `oraclecloud.com`
+//   sibling host does not resolve at all).
+//
+// The numeric suffix is BOUNDED (1-99), not open-ended: this regex exists to
+// PIN the host before every fetch, so it enumerates a finite apex family and
+// never degrades into `oraclecloud<anything>.com`.
 //
 // Career page URL:
 //   https://<host>/hcmUI/CandidateExperience/<lang>/sites/<siteNumber>/jobs
@@ -36,7 +44,10 @@
 import { decodeEntities } from './_html-entities.mjs';
 import { BROWSER_LIKE_USER_AGENT } from './_http.mjs';
 
-const ORACLE_HOST_RE = /^[a-z0-9-]+\.fa\.(?:[a-z0-9-]+\.)?(?:ocs\.)?oraclecloud\.com$/i;
+// `oraclecloud(?:[1-9][0-9]?)?` = oraclecloud.com plus oraclecloud1.com …
+// oraclecloud99.com. No leading zero, at most two digits — a bounded family,
+// so this stays a host pin and never becomes a wildcard apex match.
+const ORACLE_HOST_RE = /^[a-z0-9-]+\.fa\.(?:[a-z0-9-]+\.)?(?:ocs\.)?oraclecloud(?:[1-9][0-9]?)?\.com$/i;
 
 const PAGE_SIZE = 200;
 const MAX_PAGES = 25;             // safety cap (~5000 jobs); hard ceiling like workday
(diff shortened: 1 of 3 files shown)


santifer:
Good catch, @LayerC0de — this was failing silently, which is the worst way to fail. Oracle hands newer tenants a numbered apex, `detect()` returned null for those, and the board scanned zero jobs with no error at all: indistinguishable from a company that simply isn't hiring.

What makes this mergeable as-is rather than a discussion is that you recognised the regex is an SSRF defence, not a convenience. Bounding the family to 1-99 with no leading zero keeps it a pin instead of letting it decay into a wildcard, and the six boundary tests you added are exactly the spoofs that matter — hostile label, apex in the path, apex in the userinfo before the @, wrong TLD, plain http. Merged! 🚀

PR #2684, opened 2026-08-11
feat(providers): Eightfold AI talent-acquisition boards

LayerC0de:
## What does this PR do?

Adds `providers/eightfold.mjs` — a zero-auth provider for **Eightfold AI** talent-acquisition boards, which host branded career sites for a lot of large enterprises and had no provider until now. Plain `GET` against the public per-tenant `/api/apply/v2/jobs` endpoint on `<tenant>.eightfold.ai`; no token, no cookie, no account.

## The API findings that are actually the value here

These three cost the most to work out, so they're documented in the module header and asserted in tests rather than left as folklore:

**1. Page size is server-capped at 10 — pagination is mandatory, not an optimization.**
`num=25`, `num=50`, `num=100`, `num=200` all return **exactly 10 rows**. Measured against a 616-posting tenant: the `num` parameter is accepted and silently ignored above 10. So a large board costs `count / 10` requests — that 616-job board is 62 round-trips. `DEFAULT_MAX_PAGES = 200` (≈2,000 postings) bounds the loop regardless of what `count` claims, `max_pages` on the entry raises it, and `MAX_PAGES_CAP = 1000` is the hard ceiling even for an explicit override. A truncated board warns rather than silently under-reporting.

**2. `domain` is optional.** The parameter looks required from the career-site URLs, but the server infers it from the tenant host and echoes it back in the response — omitting it still returns that tenant's board. It's still forwarded when the entry supplies one, because multi-brand tenants genuinely scope their board by it (hence the optional `domain:` field on the portal entry).

**3. Timestamps are epoch SECONDS.** `t_create` / `t_update` are integer epoch **seconds**, not the ISO strings every other provider's date field carries. Getting this wrong doesn't fail loudly — it just dates every posting to 1970 and quietly breaks `--since` / `max_age_days`. Converted explicitly, with non-finite or non-positive values dropped rather than guessed at. There's a test pinning `1780000000` → `1780000000000`.

## Validation — and an honest limit on it

**The provider was validated primarily against a Bayer tenant (616 jobs)**: full pagination across all 62 pages, correct totals, correct URLs, correct dates. Routing, pagination, the page cap, and parsing are proven against a real board at real size.

**It was not validated against the tenant that motivated it.** I built this for NEOM's Eightfold board, and that board is **currently empty** — the API returns `count=0` and the rendered site says "No Positions Found". So the empty-board path is exercised (and tested: one request, `[]`, no error), but I cannot claim an end-to-end result on the original target. Treat "works on NEOM" as unproven; "works on Eightfold" as demonstrated on a 616-job tenant.

Several other tenants front the API with a WAF that 403s datacenter/cloud egress IPs — an environment/IP issue rather than a provider bug (the same request succeeds from a residential IP), noted as a known limitation in the header alongside the browser-like UA that reduces the friction.

## Security posture

Same discipline as `oraclecloud` / `greenhouse`:

- host-pinned to `*.eightfold.ai` via `assertEightfoldUrl()`, called **before every fetch**, not once at detect time
- `redirect: 'error'` on every page, so a server-side redirect can't walk the request off the pinned host
- a branded CNAME (`careers.<company>.com`) is **deliberately not matched** — an entry must name the canonical tenant host
- `canonicalPositionUrl` may point at a branded host and is accepted, because those are **display-only** links written to pipeline/history and never fetched by the scanner — the same treatment `jobvite.mjs` gives its `applyURL`s. Host pinning applies to endpoints we request, not links we merely record.

The SSRF negative cases are a single test block that asserts 13 hostile/unusable entries all yield `null`: the bare apex, multi-label subdomains, a lookalike suffix (`acme.eightfold.ai.evil.example`), the trusted host relegated to the path, the trusted host hidden in userinfo (`...@evil.example`), plaintext `http://`, and assorted junk. It's marked do-not-delete, since a widening that turned the pin into a wildcard would pass every positive assertion above it.

## Tests

`tests/providers/eightfold.test.mjs` — **25 assertions, all synthetic fixtures (`acme`/`big`/`pinned.eightfold.ai`), nothing touches the network.** Auto-discovered under `tests/**`, no registration in `test-all.mjs` per CONTRIBUTING.

Covers `detect()` derivation and the `api:` > `careers_url` precedence, `domain` omit/passthrough/override, the SSRF block above, `resolveTenant` / `buildApiUrl` / `buildJobUrl`, the parser's drop rules (no title / no id / non-https canonical → dropped, 7 in → 3 out), `locations[]` folding, the epoch-seconds conversion, degenerate payloads returning `[]`, pagination with a short final page, `num=10` on every request, `redirect:"error"` on every page, empty board → `[]` in one request, `max_pages`, the `ctx.maxPages` health-probe hint, and throwing before any network call for an undetectable entry.

Discovery verified rather than assumed — removing the file and re-running takes the suite from **3298 → 3273**, exactly the 25 assertions.

## Docs

Per the note at the bottom of the boards table ("update this table in the same PR"):

- `docs/SUPPORTED_JOB_BOARDS.md` — new **Eightfold AI** row
- `templates/portals.example.yml` — added to the auto-detect host list, plus a worked `Example Eightfold Co` entry showing the optional `domain:` and the `max_pages` reasoning

## Type of change

- [ ] Bug fix
- [x] New feature

## Checklist

- [x] If this is a new feature or architecture change, I opened an issue first — **N/A: new zero-auth scanner providers are explicitly exempt** ("Going straight to a PR is welcome — no issue needed — for: bug fixes, new zero-auth scanner providers, docs, and translations")
- [x] I ran `node test-all.mjs` and all tests pass — **3298 passed, 0 failed** (2 warnings, both pre-existing "expected without user data" fixture notes)

## Summary by CodeRabbit

* **New Features**
  * Added support for scanning Eightfold AI career sites.
  * Automatically detects supported sites and retrieves job listings with pagination.
  * Supports optional domain scoping, locations, timestamps, and fallback job links.

* **Documentation**
  * Added Eightfold AI to the supported job boards list.
  * Added configuration guidance and an example setup for Eightfold scanning.


santifer:
Another provider landed, @LayerC0de. Eightfold is a real gap — it's the ATS behind a lot of enterprise boards that were simply invisible to the scanner until now.

Zero-auth confirmed the way it should be: the only request sends a User-Agent and an Accept header, nothing else. Sixteen assertions including the page-cap path, and the two truncation warnings in the log are your tests exercising that cap on purpose rather than something going wrong.

One housekeeping note: this conflicted with `portals.example.yml` after I merged your own #2683 an hour earlier — my sequencing, not your problem — so I resolved it myself (kept both entries: the numbered-apex note from #2683 and your eightfold block) rather than sending you back for a rebase on a conflict I created. Merged! 🚀

PR #2686, opened 2026-08-11
fix(jd-skill-gap): recognize sentence-form requirement headings; surface low-confidence in JSON

### LINKED ISSUE #2685: jd-skill-gap: common requirement headings yield zero skills, and the JSON output hides the low-confidence warning
Two defects in `jd-skill-gap.mjs` that both surface the same way: a silent empty result, which is indistinguishable from "no gaps found". The file's own header comment already calls this out as the more dangerous of its two failure modes, so I think both are in scope as bugs rather than enhancements.

Found while running the tool across a private corpus of 17 real postings. 9 of the 17 returned `{existing: [], supportedByResume: [], gap: []}`.

## 1. `REQUIREMENT_HEADER_RE` misses several common heading forms

The regex already allows a bare heading (`^#{0,4}` permits zero hashes), so these fail purely on vocabulary — the bullets underneath are well formed and would extract fine:

| Heading in the posting | Why it misses |
|---|---|
| `YOU HAVE:` | Ashby's default template ships this bare and uppercase. Not in the list. |
| `You Might Be a Good Fit If You:` | `you may be a good fit` is present; `might` / `could` are not. |
| `It's Important To Us That You Have` | Must-have phrased as a full sentence. |
| `It Would Be Great if You Had` | Nice-to-have, same shape. |

Minimal repro:

```js
extractJdSkills("# Role\n\nYOU HAVE:\n- Hands-on experience with React, TypeScript and AWS\n")
// => []   (expected: React, TypeScript, AWS)
```

Measured on the corpus: three postings that previously extracted **zero** skills now extract **20, 8 and 3**.

Worth noting the direction — recognizing these headings *surfaces gaps that were previously invisible*. The change makes the tool stricter, not more permissive.

## 2. `diagnoseExtraction()` never runs on the JSON path

`diagnoseExtraction()` is called inside the `if (summaryMode)` branch only. `--summary` prints a clear `LOW CONFIDENCE` banner; the default JSON output emits bare empty arrays with no signal at all.

```
$ node jd-skill-gap.mjs jds/example.md
{ "existing": [], "supportedByResume": [], "gap": [] }      # reads as "clean"

$ node jd-skill-gap.mjs jds/example.md --summary
  LOW CONFIDENCE: this is not a clean result.
     No requirements section was recognized in this JD...
```

The JSON branch is the one other tooling consumes, and per the comment right above it, `modes/pdf.md` Step 4 gates on this output — so it's the branch that most needs the warning.

## Not included (deliberately)

Two adjacent issues I found and did **not** touch, because both *remove* gaps, and a change that only ever makes a résumé look better should carry its own before/after evidence rather than ride along with a parsing fix:

- `NON_REQUIREMENT_HEADER_RE` covers `why join` / `why work` / `why this role` but not `WHY {COMPANY}:`, so a perks section can leak into the requirements block (`PTO`, `Grow`, `Collaborative` reported as skill gaps).
- Generic-token noise in the `gap` bucket more broadly — `Enthusiasm`, `Comfort`, `Support` and similar sentence-leading words are still reported as missing skills.

Happy to open separate issues for those if useful.

## PR

Fix for both, with 6 added self-tests (including a guard that the responsibilities heading `YOU WILL` does **not** open a requirements block — it's a near-miss for `you will have`): PR to follow.

Environment: node 24.18.0, macOS. Full suite: 3271 passed / 1 failed before and after the change; the one failure is a pre-existing `followup-cadence.mjs` crash on `main`, confirmed by stashing the change and re-running.

gdiab:
Fixes #2685.

Both defects present the same way — a silent empty result that reads identically to "no gaps found", which the file's own header comment calls the more dangerous of its two failure modes.

## 1. Requirement headings that are bare or sentence-form

Added to `REQUIREMENT_HEADER_RE`:

- `you have` / `you'll have` / `you will have` — Ashby's default template ships `YOU HAVE:` bare and uppercase. The `^#{0,4}` prefix already permitted a hash-less heading, so this was purely a vocabulary miss; the bullets underneath were always well formed.
- `might` / `could` alongside the existing `may` in `you may be a good fit`.
- `it's important to us that you have` and `it would be great if you had` — must-have / nice-to-have phrased as full sentences.

**Direction of the change:** recognizing these headings *surfaces gaps that were previously invisible*, so results get stricter, not more flattering. On a private corpus of 17 real postings, three that previously extracted zero skills now extract 20, 8 and 3.

## 2. `lowConfidence` in the JSON output

`diagnoseExtraction()` was computed inside the `if (summaryMode)` branch only, so `--summary` warned loudly and the JSON path said nothing. It is now computed once for both modes and included as an additive key:

```json
{
  "existing": [], "supportedByResume": [], "gap": [],
  "lowConfidence": {
    "reason": "no-requirements-section",
    "message": "No requirements section was recognized in this JD, so no text was scanned for skills..."
  }
}
```

`null` when the run was conclusive. Existing consumers reading the three buckets are unaffected. This is the branch `modes/pdf.md` Step 4 gates on, per the comment already in the file.

## Tests

6 new self-tests: one per heading form, plus a guard in the opposite direction — `YOU WILL` (responsibilities) must **not** open a requirements block. It's a near-miss for `you will have`, and scoring duties as requirements is exactly the over-extraction the file's design note calls unrecoverable.

- `jd-skill-gap.mjs --self-test`: 48 → 54 passing
- `test-all.mjs`: 3271 passed / 1 failed, unchanged before and after. The single failure is a pre-existing `followup-cadence.mjs` crash on `main` — verified by stashing this change and re-running, same result.

## Deliberately out of scope

Two adjacent defects I found and left alone, because both *remove* gaps rather than surface them, and a change that only ever makes a résumé look better deserves its own PR with its own before/after evidence:

- `NON_REQUIREMENT_HEADER_RE` covers `why join` / `why work` / `why this role` but not `WHY {COMPANY}:`, letting a perks section leak into requirements (`PTO`, `Grow`, `Collaborative` reported as gaps).
- Generic-token noise in the `gap` bucket generally (`Enthusiasm`, `Comfort`, `Support`).

Glad to take either as a follow-up if you want them.

https://claude.ai/code/session_01865yhD5BLeERLiTa6h2yS8

## Summary by CodeRabbit

* **New Features**
  * Improved requirement extraction for sentence-style and “You have” headings.
  * Added `lowConfidence` diagnostics to JSON results, including `null` when no issue is detected.
  * Standardized low-confidence warnings across output modes.

* **Bug Fixes**
  * Prevented “You will” responsibility headings from being treated as requirements.

* **Tests**
  * Added coverage for expanded heading recognition, section boundaries, and diagnostic output.

### DIFF
--- a/jd-skill-gap.mjs
+++ b/jd-skill-gap.mjs
@@ -44,9 +44,9 @@ const CV_PATH = 'cv.md';
 
 // Real postings rarely use the word "Requirements". The literal-only list
 // missed the phrasings most modern ATS boards actually ship ("What we're
-// looking for", "Who you are", "You may be a good fit if"), so a JD could
-// yield zero skills - which reads identically to "no gaps found" and is the
-// more dangerous of the two failure modes this file warns about.
+// looking for", "Who you are", "You may be a good fit if", "You have"), so a
+// JD could yield zero skills - which reads identically to "no gaps found" and
+// is the more dangerous of the two failure modes this file warns about.
 const REQUIREMENT_HEADER_RE = new RegExp(
   '^#{0,4}\\s*(?:' + [
     'required', 'requirements', 'qualifications', 'must[- ]have', 'preferred', 'nice[- ]to[- ]have',
@@ -55,7 +55,16 @@ const REQUIREMENT_HEADER_RE = new RegExp(
     'who\\s+you\\s+are',
     'about\\s+you',
     'your\\s+(?:background|experience|profile)',
-    'you\\s+may\\s+be\\s+a\\s+good\\s+fit',
+    'you\\s+(?:may|might|could)\\s+be\\s+a\\s+good\\s+fit',
+    // Ashby's default template ships a bare "YOU HAVE:" heading (no markdown
+    // hashes - already allowed by the ^#{0,4} prefix). Without this the whole
+    // requirements block is invisible even though the bullets under it are
+    // perfectly well formed.
+    "you(?:(?:'|’)ll|\\s+will)?\\s+have",
+    // Postings that phrase must-have/nice-to-have as full sentences rather
+    // than noun headings.
+    "it(?:'|’)?s\\s+important\\s+to\\s+us\\s+that\\s+you\\s+have",
+    'it\\s+would\\s+be\\s+great\\s+if\\s+you\\s+ha(?:ve|d)',
     'ideal\\s+candidate',
     'skills\\s+(?:and|&)\\s+experience',
   ].join('|') + ')s?\\b.*$',
@@ -416,6 +425,34 @@ Python, Docker, Zookeeper
     true
   );
 
+  // Regression: requirement headings that are full sentences or bare
+  // uppercase rather than the noun forms ("Requirements", "Qualifications").
+  // Each of these silently yielded ZERO skills, which is indistinguishable
+  // from "no gaps found" — the failure mode this file's header warns about.
+  const headerVariants = [
+    ['bare uppercase "YOU HAVE:" (Ashby default template)', 'YOU HAVE:'],
+    ['"You\'ll have"', "You'll have:"],
+    ['"You might be a good fit if you:"', 'You Might Be a Good Fit If You:'],
+    ['"It\'s Important To Us That You Have"', "## It's Important To Us That You Have"],
+    ['"It Would Be Great if You Had"', '## It Would Be Great if You Had'],
+  ];
+  for (const [label, heading] of headerVariants) {
+    const jd = `# Role\n\n${heading}\n- Hands-on experience with React, TypeScript and AWS\n`;
+    const skills = extractJdSkills(jd);
+    eq(`${label} opens a requirements block`, skills.includes('React'), true);
+  }
+
+  // Guard the other direction: the responsibilities heading "You will" must NOT
+  // open a requirements block. It reads as a near-miss for "you will have", and
+  // treating duties as requirements is exactly the over-extraction this file
+  // calls unrecoverable.
+  const responsibilitiesJd = `# Role\n\nYOU WILL\n- Ship Kubernetes manifests for the platform team\n`;
+  eq(
+    '"YOU WILL" (responsibilities) does not open a requirements block',
+    extractJdSkills(responsibilitiesJd).includes('Kubernetes'),
+    false
+  );
+
   // Regression: generic JD boilerplate must not be misreported as a skill gap.
   const boilerplateJd = `
 # Role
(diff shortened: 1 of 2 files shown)


gdiab:
Thanks — all three were valid. Pushed in 5720967.

**The `YOU WILL` block-close finding was the important one, and it was worse than described.** My guard only proved `YOU WILL` couldn't *open* a block; it never checked that it *closes* one. Reproduced before fixing:

```
YOU HAVE:
- Experience with Python        → Python  ✅ correct
YOU WILL
- Ship Kubernetes manifests     → Kubernetes, Terraform reported as SKILL GAPS
- Operate Terraform modules
```

Pure responsibilities scored as missing requirements — exactly the over-extraction the file's design note calls unrecoverable. Took the proposed `'you\s+will(?!\s+have)'` pattern; the lookahead is load-bearing because `NON_REQUIREMENT_HEADER_RE` is tested *before* `REQUIREMENT_HEADER_RE` in `scanJd()`, so without it `You will have:` would close a block instead of opening one. Both directions are now asserted.

Worth noting for the record: this hole predates my change — `## Requirements` followed by a bare `YOU WILL` would always have leaked. But recognizing more requirement headings widens its blast radius a lot, since many more postings now open a block at all. So it belongs in this PR rather than a follow-up.

**Missing fixtures:** added `You Will Have:` and `You Could Be a Good Fit If You:`. Both were untested branches I introduced.

**JSON serialization test:** added as `tests/jd-skill-gap-json-output.test.mjs` rather than inline in `test-all.mjs` — the `tests/**/*.test.mjs` auto-discovery picks it up, so no registration edit is needed. It runs the real binary in a temp cwd with a fixture `cv.md` and parses actual stdout, covering both the conclusive case (`lowConfidence: null`, buckets populated) and the inconclusive one (object with `reason` and `message`).

I checked that suite is a real guard rather than a tautology: run against the pre-fix `jd-skill-gap.mjs`, it fails 7 of 11 assertions.

**Totals:** self-test 54 → 60, full suite 3271 → 3282 passed. Still 1 failed — the `followup-cadence.mjs` `analyze() appDateSource` crash, which reproduces on a clean `main` with this branch stashed, so it looks unrelated. Happy to open a separate issue for it if it isn't already known.

Still leaving the `WHY {COMPANY}:` perks leak and the generic-token gap noise (`Enthusiasm`, `Comfort`, `PTO`) out of scope — both *remove* gaps rather than surface them, and I'd rather they carry their own before/after evidence than ride along here.

santifer:
Welcome to career-ops, @gdiab. Job descriptions don't agree on how to announce their requirements — some write `Requirements:`, plenty write a full sentence like `What you'll need to succeed` — and a heading detector that only knows the terse form silently extracts nothing from the second kind. Silent zero-extraction is the failure mode worth fixing, because it looks identical to a JD that genuinely lists no skills. Your first contribution here — thanks for opening the issue first too.

PR #2688, opened 2026-08-11
fix(web): add Grok Build CLI to the web runtime picker

WJiangH:
## The bug

Grok Build CLI is wired into every surface of the repo except the one that
decides whether the web UI can use it:

| Surface | Grok present? |
|---|---|
| `doctor.mjs` → `VALID_CLIS` | ✅ |
| `scaffolder/bin/cli.mjs` → `SUPPORTED_CLIS` | ✅ |
| `docs/SUPPORTED_CLIS.md` | ✅ `grok -p "prompt"` |
| `README.md` → *Grok Build CLI Integration* | ✅ (a whole section) |
| `.grok/skills/career-ops/SKILL.md` | ✅ |
| `web/src/lib/clis.ts` → `KNOWN` | ❌ |

`detectClis()` only probes `PATH` for binaries listed in `KNOWN`, so with Grok
installed the web UI still cannot select or spawn it. Nothing tied the two
lists together, so the omission was invisible.

## The fix

One entry, following the official docs
([docs.x.ai/build/overview](https://docs.x.ai/build/overview)): the binary is
`grok`, headless is `grok -p "prompt"`.

Grok also supports `--output-format streaming-json`, but that is its own schema
rather than Claude's `stream-json` — the only one `/api/run` parses — so this
uses plain `-p`, like every other non-Claude entry.

## Stopping the drift

`web/tests/lib/clis-coverage.test.mjs` asserts that every CLI
`docs/SUPPORTED_CLIS.md` documents a headless invocation for is present in
`KNOWN`. **Verified to fail without the `clis.ts` change:**

```
not ok 2 - every documented headless CLI is selectable in the web UI
    documented headless but absent from KNOWN in clis.ts: grok
```

It also self-checks its own fixtures, so a future refactor past its regexes
fails loudly instead of passing vacuously over empty sets.

CLIs the docs list as interactive-only (Cursor, Kimi) stay correctly excluded —
the web UI only drives workers headlessly.

## Testing

- `node test-all.mjs` → 3138 passed, 0 failed
- `cd web && npm test` → 64 passed, 0 failed
- `npx tsc --noEmit` → clean outside generated `.next/`
- Live on macOS: `detectClis()` now resolves `grok` at `~/.local/bin/grok`, and
  `/api/clis` returns it as installed.

## Summary by CodeRabbit

- **New Features**
  - Added Grok Build as a supported CLI option.
  - Enabled plain-text prompt handling when invoking Grok Build.
  - Added a link to Grok Build documentation.

- **Tests**
  - Added coverage ensuring all documented headless CLIs are supported.
  - Added checks for duplicate CLI identifiers and correct Grok Build configuration.


santifer:
Thanks for this, @WJiangH.

Grok Build CLI is wired into doctor.mjs (VALID_CLIS), the scaffolder (SUPPORTED_CLIS), docs/SUPPORTED_CLIS.md, a whole README section and .grok/skills, and missing from exactly one list: KNOWN in web/src/lib/clis.ts, which is what detectClis() probes PATH against. The result is that with grok installed the web still can't select or launch it, and since nothing tied the two lists together the omission was invisible. The fix is a single entry; what carries the weight is the coverage test requiring every CLI with a documented headless invocation to be in KNOWN, with the failing output pasted in ('documented headless but absent from KNOWN in clis.ts: grok'), and it correctly leaves out the CLIs the docs mark as interactive-only (Cursor, Kimi).

santifer:
Merged.

The entry itself is one line, but the test is why this is the version that landed. It derives the expected contents of `KNOWN` from `docs/SUPPORTED_CLIS.md` instead of restating them, so the two lists can't drift apart again — and the anti-vacuity check ("the fixtures this guard reads still look like themselves") means a refactor that breaks the regex fails loudly instead of passing over an empty set. That second part is the one most guards get wrong, including some of ours.

Your comment on why not to use `--output-format streaming-json` is doing real work too: it records a decision that would otherwise look like an oversight to the next person.

For transparency, since it affects another contributor: **#2663** added the same CLI a day earlier and I did not take it, because it carried `--always-approve` to let workers use Write/Bash without a TUI. The `pdf` worker has Bash explicitly denied on Claude's path, so that flag would have made the user's runtime choice silently change what a worker may do to their files. Yours grants nothing extra. I've asked that PR to reduce to its install-path detection, which is the half yours doesn't cover.

I've since turned that into an enforced rule rather than a note — thanks for setting the pattern I copied for it.

PR #2689, opened 2026-08-11
fix(web): per-CLI stream parsing so non-Claude runs report real usage

WJiangH:
> **Stacked on #2688** — the Grok entry it adds to `KNOWN` is what this branch
> attaches `streamArgs` to. Please merge #2688 first; the diff here will shrink
> to just this change once it lands.

## The problem

`/api/run` parsed Claude's `stream-json` inline and fell back to raw stdout for
every other CLI. Display worked; accounting did not. A codex or grok
evaluation — a full A–F report, minutes of work — was recorded as
`tokens: 0, costUsd: null` and contributed nothing to `/api/usage`. Cost was
invisible on exactly the runtimes a user picks in order to control cost.

## The change

`web/src/lib/cli-stream.mjs` maps each CLI's dialect onto one event shape
(`text` / `tool` / `status` / `usage`). Parsers take a single parsed JSON value,
so line buffering and transport stay in the route while dialect knowledge
becomes unit-testable against fixtures captured from the real CLIs.

### The token formula is not one shared expression

This is the substance of the PR. `tokens` means *billed at full rate* = fresh
input + output + cache writes, excluding discounted cache reads. But vendors
disagree on whether `input_tokens` already contains the cached portion:

| CLI | Convention | Formula |
|---|---|---|
| claude, grok | `input_tokens` **excludes** cache reads | `input + output + cache_creation` |
| codex | `input_tokens` **includes** `cached_input_tokens` (OpenAI) | `(input − cached) + output` |

Grok's convention was verified arithmetically on a live run rather than inferred
from field names: `18575 + 32 + 5376 + 0 == 23983`, which is the `total_tokens`
grok itself reports.

Using one formula for all three does not throw — it produces a plausible wrong
number, which is worse.

Codex reports no cost, so `costUsd` stays `null` instead of being invented from
a guessed rate. Grok's `thought` events are dropped: surfacing reasoning as
answer text would also set `emittedText` for a run that never answered,
defeating the route's honesty gate.

## Two bugs surfaced while verifying end to end

**1. `spawn` left the child an stdin pipe nobody ever closes.** Default `stdio`
is `pipe` for all three streams, so a CLI that checks stdin blocks forever
waiting for EOF. Codex prints `Reading additional input from stdin...` and
hangs — **it has never worked in the web UI for this reason**. Claude masked it
by not reading stdin at all. Now `stdio: ["ignore", "pipe", "pipe"]`; nothing
can answer a prompt in a headless run anyway.

**2. Codex logs a benign `ERROR codex_models_manager::cache: failed to load
models cache` on every invocation**, which the deliberately broad stderr rule
reads as a failed run. `CliSpec` gains an optional `benignStderr`, checked
before that rule.

## Testing

Verified live on macOS:

| CLI | Result |
|---|---|
| grok, evaluate | 72,635 tokens / $0.328, and a real 22K report |
| codex, research | 73,238 tokens, 7 tool events, no spurious error |
| claude, research | 61,532 tokens / $0.471 — unchanged by the refactor |

- `cd web && npm test` → 89 passed, 0 failed (25 new)
- `node test-all.mjs` → 3138 passed, 0 failed
- `npx tsc --noEmit` → clean outside generated `.next/`

## Summary by CodeRabbit

* **New Features**
  * Added structured streaming support for Claude, Grok, and Codex.
  * Progress updates now include text, tool activity, status, and usage details.
  * Improved provider-specific token and cost reporting.

* **Bug Fixes**
  * Prevented headless runs from hanging while waiting for input.
  * Filtered harmless diagnostic messages from error handling.
  * Ignored malformed or unsupported stream events safely.

* **Tests**
  * Expanded coverage for event parsing, usage calculations, provider behavior, and supported CLI configurations.


santifer:
Thanks for this, @WJiangH.

/api/run only parsed Claude's stream-json and fell back to raw stdout for everything else: a full A-F evaluation with codex or grok, minutes of work, was recorded as tokens: 0 and costUsd: null and contributed nothing to /api/usage, so cost was invisible in exactly the runtimes a user picks in order to control cost. The substance is that the token formula isn't shareable: claude and grok exclude cache reads from input_tokens (input+output+cache_creation) while codex includes them (input−cached+output), and you verified grok's convention arithmetically on a real run (18575+32+5376+0 == the 23983 grok itself reports) instead of inferring it from field names. Along the way it uncovered two bugs: spawn left the child a stdin pipe nobody closes, so codex hangs on "Reading additional input from stdin" and has never worked in the web, and a benign codex model-cache ERROR was read as a failed run.

WJiangH:
Both points addressed in `583999a`.

**The anchoring was a real hole.** `/models cache|base_instructions/i` is a
substring test, and it is a hole punched in a deliberately broad failure rule —
so it needed to be exactly the size of the diagnostic it was cut for, and was
not. Each of these was silently discarded and the run reported clean:

```
Error: quota exceeded while refreshing the models cache
fatal: unauthorized — cannot read models cache for this account
ERROR: invalid api-key; models cache untouched
TypeError: Cannot read properties of undefined (reading 'base_instructions')
```

Now anchored at line start and pinned to the emitting module and its message. A
test asserts the anchoring *structurally* (`re.source.startsWith("^")` over every
entry), so a future exemption cannot quietly be added as a substring.

I also moved it, with the failure pattern, out of `CliSpec` and route.ts into
`web/src/lib/stderr-classify.mjs`. Both are now assertable as values rather than
by grepping the route — the arrangement `claude-invocation.mjs`'s header
describes as having been defeated five times by rewriting the route around the
guard.

**The line-classification half was already satisfied** on this branch: the
rebase onto today's `main` adopted upstream's complete-line stderr buffering,
and `child.on("close")` flushes a trailing unterminated line before
`if (closed) return` and the honesty gates.

**One thing I did not fix, on purpose.** Writing the tests surfaced a
pre-existing gap: neither `panic` nor `failed` is in `STDERR_FAILURE`, so
`panic: failed to write models cache: disk full` reads as ordinary chatter.
Adding `failed` would catch it — and also `0 failed` and `Failed to load
optional plugin, continuing`, and a false positive marks a run whose PDF
rendered fine as failed (`pdfRunOutcome`). That is a trade worth making on its
own terms rather than widening a tuned pattern inside a fix for something else.
It is pinned by a test and documented at the pattern, so it stays a decision
rather than a surprise. Happy to take it in a follow-up if you want it.

`node test-all.mjs` → 3403 passed / 0 failed · `cd web && npm test` → 170 passed.

santifer:
**The conflict is one hunk, and it's because your fix already landed — from someone else.**

`api/run/route.ts` went `CONFLICTING` when **#1973** merged a few hours ago. That PR makes the *identical* change yours does on the same line:

```js
const child = spawn(binPath, args, { cwd: careerOpsRoot(), env: process.env, stdio: ["ignore", "pipe", "pipe"] });
```

`main` already has that exact line. So when you rebase, that hunk resolves to **"already applied"** — keep `main`'s version and drop yours; nothing of your work is lost. The rest of your diff (`cli-stream`, `stderr-classify`, `clis` coverage) doesn't collide with anything that landed.

Worth saying plainly: **three independent contributors found this same stdin bug** — you, #1973, and #2361 as part of a larger Codex rewrite. That's not wasted effort, it's the strongest possible evidence the bug was real and was hurting people. It just means the six-line version reached `main` first.

You said you addressed both review points in `583999a` — noted, and I'm reviewing the per-CLI stream parsing next. If you'd rather not deal with the rebase, say so and the maintainer can push it from our side; I'd just rather not force-push over your branch without asking.

The conflict is our doing, by the way, not a lapse on yours: a lot merged today and this file had four PRs on it at once.

santifer:
**Your token formula is correct, and I verified it against real data rather than taking the argument on faith. It's also the part of this PR that nothing else has.**

I reviewed **#2102** (@nikolaysm) tonight — it lands on the same surface from the other end: `spawnHeadlessCli` across every CLI route, the report-persistence gate, the `<<cv-html>>` envelope. It also parses Codex usage, and it gets this wrong:

```js
tokens: (usage.input_tokens || 0) + (usage.output_tokens || 0)
```

Your header note is exactly why that's wrong, so I checked which convention the dashboard actually uses. `web/src/app/api/usage/route.ts:69` sums `input + output + cache_creation` and deliberately omits `cache_read_input_tokens` — i.e. the metric is *tokens billed at full rate*. Measured on a real Claude Code transcript:

```
{'input_tokens': 2, 'cache_creation_input_tokens': 2017, 'cache_read_input_tokens': 997009, 'output_tokens': 184}
```

`input_tokens: 2` next to 997009 cache reads settles it — Claude's input pool excludes reads, so subtracting Codex's `cached_input_tokens` is the only translation that measures the same thing. On the run captured in your fixture that's **19729 vs 84497**: a 4.3× overstatement, silently, on the runtime people choose to save money.

Three things I'm keeping from your PR specifically:

- **the subtraction**, with the header note explaining the vendor split — the reasoning has to survive in the code, or the next refactor re-adds the cached tokens as an "obvious fix";
- **`Math.max(0, …)`** — a malformed usage block reports 0, never a negative;
- **`costUsd: null`** — Codex reports no cost, and a guessed rate would be fiction.

I've asked #2102 to adopt the formula with credit to this PR. To be straight with you about what that means: #2102 covers more surface and is already rebased onto `main`, so it will most likely land first, and this branch will shrink to what it uniquely adds. That's a sequencing call about conflict cost, not a judgment about quality — on the accounting question your PR is the one that's right, and that's the part users will feel.

Same as I said on the stdin collision: three people finding the same bug is evidence it was real and hurting people. Here you found something the other PR didn't, and it would have shipped wrong without you.

No action needed from you yet — I'm not asking you to rebase into a moving target. I'll come back when #2102 lands with the exact remaining delta.

WJiangH:
Understood on all three, and thank you for going back to correct the "verified" on #2102 publicly — that is the part of this that made the formula land somewhere it will actually ship.

**On the rebase: I'll do it, and I'd rather you didn't push over the branch.** Not out of ownership — it's that the resolution needs a judgement I should be the one to record: `main` already has the six-line version of the stdin fix from #1973, so that hunk resolves to "keep `main`, drop mine". If someone else takes it, that decision is invisible in the history.

Taking you at your word that it isn't wanted yet, though: I'm **not** rebasing until #2102 lands, so it isn't done twice into a moving target. Say the word when you have the remaining delta and I'll do it in one pass.

Three contributors finding the same stdin bug also says something about where I should aim next — a bug that obvious getting hit three times means my time is better spent where the evidence is harder to come by than where it's lying around.

**On the queue, since you asked for an order.** Eleven at once was a mistake in sequencing, not in intent: they came out of one pass over the repo and I filed them as I found them instead of pacing them behind each other. If it helps, review them in this order and treat everything below the line as genuinely not urgent:

1. **#2689** (this one) — the accounting is already blocking on the #2102 integration
2. **#2692** — has your review; four points addressed as of just now
3. **#2688** — one KNOWN entry plus a coverage test, smallest thing to clear
4. **#2691**, **#2690** — the onboarding/archetype pair
5. #2730, #2731, #2732, #2733, #2749, #2750 — no hurry on any of these

I won't open new ones until this queue is down. If closing some and re-opening them later would help more than reordering, tell me which and I'll close them myself.

One thing I owe you unprompted: **#2750 and #2732 were red and I hadn't noticed** — a js-yaml default import that only fails once merged onto today's `main`, and a test that pinned "unwritable" to `/proc`, which is only unwritable on Linux. Both fixed and pushed. I've started verifying under the merge-onto-`main` boundary rather than on the branch, which is what you've been doing all along and is the reason you catch these and I don't.

WJiangH:
**Revising the order I gave you, because its first entry is the one thing that can't move.**

#2102 is still open — `CHANGES_REQUESTED`, but @nikolaysm addressed all four points and rebased onto today's `main` about an hour ago, including adopting this branch's token formula with the clamp and `costUsd: null`. So the delta you said you'd come back with isn't knowable yet, and I'm holding as you asked: no rebase into a moving target, and I'd rather not do it twice while that branch is itself freshly rebased.

Which makes my #1 unactionable. New order, by what you can actually close today:

1. **#2688** — you called it the smallest thing to clear, it's MERGEABLE, and CI is green. One `KNOWN` entry plus the coverage test.
2. **#2692** — MERGEABLE, green, your four points addressed in `5277c44`.
3. **#2691**, **#2690** — the onboarding/archetype pair, both MERGEABLE and green.
4. **#2689** — blocked on #2102 by your call, not mine. Ping me when it lands and I'll rebase in one pass.

One change on #2689 since you last looked: CodeQL flagged `parserFor` at high severity — `js/unvalidated-dynamic-method-call`, since `cliId` comes from the request body. The `hasOwnProperty` guard did make it safe, but that's a property an analyzer has to infer and this one didn't. Rather than suppress it I replaced the keyed lookup with a switch over literals, so the reachable set is fixed in the source and there's nothing left to infer. Pushed in `f9cc39c`.

**On the queue — I'd rather close than reorder.** You offered both. #2730, #2731, #2732 and #2733 are the four I'd close now: they're a related group (scan-filter observability), none of them blocks anything, and re-opening them later costs nothing since the branches stay. That takes your side of the queue from seven to three. Say the word and I'll close them myself — or name a different four if these aren't the ones costing you the most.

And I'm not opening anything new until some of these land. I have a change ready for #2751 that I'm deliberately sitting on for that reason.

santifer:
Quick note so silence here does not read as no one looking: **this one is not mine to merge, and that is not a brush-off.**

`web/` is maintained by a separate reviewer, and I route rather than triage there so you do not get two maintainers giving you contradictory asks on the same diff. It is labelled `area:web`, it is in their queue, and I have just sent them the state of that queue including this PR.

What I can tell you concretely rather than vaguely: there are 34 open `area:web` PRs right now, median age 14 days. That is a real backlog and it is being looked at as a backlog, not ignored one PR at a time.

If your PR turns out to touch something on the core side too, that half is mine and I will take it here. And if this goes quiet for another stretch, ping me: chasing it across to the right reviewer is exactly the kind of thing I should be doing rather than you.

santifer:
Update, because part of this PR is already in `main` and you shouldn't have to discover that by rebasing.

**#2688 merged**, which lands two of this PR's seven files:

```
web/src/lib/clis.ts                   ← the grok entry
web/tests/lib/clis-coverage.test.mjs  ← the derived-from-docs guard
```

On rebase both resolve as "already applied" — keep `main`, drop yours, nothing lost.

That leaves your real delta: `cli-stream.mjs`, `stderr-classify.mjs`, their tests, and the `run/route.ts` wiring. And that delta does overlap **#2102** head-on: it builds `run-cli-support.mjs` covering the same ground (`parseCodexEvent` / `parseClaudeEvent` / fatal-stderr classification). Two implementations of one thing, so only one can land as the base.

**#2102 goes first** — it's wider (it also fixes the stdin hang across every CLI route and the report-persistence gate) and it's already rebased onto current `main`. That is a sequencing call about conflict cost, not a verdict on quality.

**Your token formula is going in with it, credited to this PR.** I've asked #2102 to adopt:

```js
const fresh = Math.max(0, n(u.input_tokens) - n(u.cached_input_tokens));
```

along with the `Math.max(0, …)` clamp and `costUsd: null`. I verified the reasoning against a real Claude Code transcript rather than the argument alone — `input_tokens: 2` next to `cache_read_input_tokens: 997009` settles which pool is which — and on your own fixture the difference is **19729 vs 84497**, a 4.3× overstatement that produces no error and no failing test. Without this PR that would have shipped.

Still nothing to do on your side, and please don't rebase yet — #2102 is still moving. I'll come back with the exact remaining delta once it lands, as promised.

For what it's worth: three of your PRs have now contributed to `main` in some form tonight (#2749, #2750, #2688), and the one still open is the one that found the bug the others didn't.

santifer:
**#2102 is merged, and your token formula went in with it — credited to this PR in the code.**

`run-cli-support.mjs` now carries:

```js
const fresh = Math.max(0, (usage.input_tokens || 0) - (usage.cached_input_tokens || 0));
```

with the clamp and `costUsd: null`, and a comment naming #2689 as where the formula and the reasoning came from. There's also a test built from the real captured turn: **2992 billed at full rate against 14000 reported** by the old formula — no error, no failing test, on the runtime people pick specifically to spend less. That ships correct because you looked at the vendor conventions instead of assuming one shared formula.

**Where that leaves this PR.** Two of its seven files already landed via #2688 (`clis.ts`'s grok entry and `clis-coverage.test.mjs`) — those resolve as "already applied" on rebase. The remaining delta is `cli-stream.mjs` + `stderr-classify.mjs` + their tests, and #2102 now covers that ground with `run-cli-support.mjs`.

So the honest position: **most of this PR has landed, by three different routes, and what's left overlaps the base that just merged.** Before you spend time rebasing, I'd rather you tell me what you think is still missing — you've been closer to this surface than anyone, and if `cli-stream.mjs` does something the merged version doesn't, I want that rather than a closed PR.

If it turns out nothing's left, say so and I'll close it with the credit it's already earned: three of your PRs landed today (#2749, #2750, #2688) and this one supplied a formula that would otherwise have shipped 4.7× wrong.

WJiangH:
Checked it against merged `main` rather than answering from memory. **Most of this PR is gone, and what's left is one thing — but it got bigger tonight rather than smaller, because #2688 landed.**

**Superseded, and I'd drop all of it:**

| this PR | landed as |
|---|---|
| `clis.ts` grok entry, `clis-coverage.test.mjs` | #2688 |
| the Codex token formula | #2102 → `run-cli-support.mjs` |
| `parseClaude` / `parseCodex` | `parseClaudeEvent` / `parseCodexEvent` |
| `stderr-classify.mjs` | `isFatalCodexStderr` / `isFatalClaudeStderr` / `isFatalGenericStderr` — **better than mine.** Mine anchored one benign codex line; yours is per-CLI with a transient quota carve-out, and #2882 just anchored the fallback. Nothing of mine to keep there. |
| `foldUsage`'s costUsd rule | `route.ts:256` is `if (typeof ev?.costUsd === "number")` — the same rule, arrived at independently. Worth noting that's also CodeRabbit's ask on this PR, declined on both sides for the same reason. |

**What's actually left: grok has no parser, and #2688 made that reachable.**

On today's `main`, `grok` is in `KNOWN` and selectable in the picker, with no `streamArgs`, no `parseEvent` and no `stderrIsFatal`. So it falls through to raw stdout: the run displays fine and records `tokens: 0, costUsd: null`. That is the exact bug this PR was opened for, and #2688 moved it from "grok isn't offered" to "grok is offered and silently uncounted".

**And grok can't just be plugged into `accumulateTokens`.** Its docstring is right for codex and claude and wrong for grok:

> `turn.completed` (Codex) and `result` (Claude) usage is per-event, not cumulative-since-start, so callers must accumulate

Grok's `end` frame carries the **turn total**, and `usage` frames arrive before it. Summing adds the parts to the whole:

```
usage  →  9212
usage  →  9395
end    → 18607   ← the turn total, verbatim from a real run
                   (the two usage frames above are illustrative)

accumulate: 37214     last-wins: 18607     2.0× over
```

Whichever way the intermediate frames are read, `end` is the total, so anything added to it over-counts. @nikolaysm called this out on #2102 himself — *"if grok parsing lands here, this side should adopt that rule rather than keep summing"*.

**So the remaining delta is: a grok `parseEvent`, plus making the fold rule per-CLI instead of global.** That's much smaller than this branch — two of seven files, and one of those shrinks to a single parser.

**My preference, and it's yours to call.** Rather than rebase this branch into that shape, I'd close it and put the grok parser up as its own small PR when your queue has room — you asked me not to open new ones, so I'm not doing that unilaterally. The alternative is I reduce this branch in place to just the grok delta and you keep the same number; also fine, and it keeps the history of how the formula got here.

Either way the credit is already where it belongs — it's in `run-cli-support.mjs` and it shipped correct. Say which and I'll do it.

santifer:

santifer:
Concrete update, @WJiangH, and it changes what's worth doing with this PR today.

WJiangH:
**`3754edd`** — rebased onto today's `main`, and it is now in the `route.ts` block's shape. Conflicting → mergeable. **Seven files became three, and `route.ts` is no longer one of them.**

I did this as a reduction rather than a merge, because a merge would have produced the one thing you said can't land: two implementations of the same surface sitting next to each other. Five of the seven files were superseded while this waited.

| this branch had | landed as | verdict |
|---|---|---|
| `clis.ts` grok entry, `clis-coverage.test.mjs` | #2688 | dropped, already applied |
| the token formula | #2102 → `run-cli-support.mjs` | dropped, already there and credited |
| `parseClaude` / `parseCodex` | `parseClaudeEvent` / `parseCodexEvent` | dropped |
| `stderr-classify.mjs` | `isFatal*Stderr` | dropped — **yours is better than mine**, per-CLI with a transient carve-out, and #2882 anchored the fallback |
| `foldUsage`'s global last-wins | — | dropped as written; see below |

**What's left is the one thing nothing covers: grok has no parser, and #2688 made that reachable.** On `main`, `grok` is in `KNOWN` and selectable in the picker with no `streamArgs`, no `parseEvent`, no `stderrIsFatal` — so it falls through to raw stdout. The run displays fine and records `tokens: 0, costUsd: null`, on the runtime people pick specifically to spend less.

`parseGrokEvent` is written in the shape the other two already use, so **`route.ts` needs no change at all** — the dispatch comment says adding the next such CLI needs no change here, and that holds.

**The one place it couldn't just plug in was the fold, and both directions were wrong.** `accumulateTokens` adds, correctly, because Codex `turn.completed` and Claude `result` each report one turn. Grok's `end` reports the **run total**, with `usage` frames before it:

```
usage  →  9212
usage  →  9395
end    → 18607   ← the turn total, from a real run
accumulate: 37214      correct: 18607      2.0x over
```

My branch's answer was global last-wins, which is the mirror of the same mistake — it would have made every Codex multi-turn run undercount. The two conventions aren't distinguishable from the number alone, so the **parser** now marks which one it holds (`tokensAreTotal`) and the fold obeys it. Absent means delta, so every existing CLI and caller is byte-identical in behavior. This is what @nikolaysm asked for on #2102 — *"if grok parsing lands here, this side should adopt that rule rather than keep summing"*.

Intermediate `usage` frames are still emitted, so a run killed mid-flight records what it spent rather than zero.

**Verification.** web suite **453/0**, root suite **7571/0** — identical to clean `main` including its 13 pre-existing warnings, which I checked by running the baseline rather than assuming. `npm run typecheck` clean under `web/`'s own dependency tree, not the root one. And the fold change is mutation-checked: reverting it to unconditional addition fails the grok-total case and nothing else, so the test bites for the right reason.

Understood on the two caveats — the conflicts were from later merges, and the verdict isn't yours. Previous head was `bb6dac6` if anyone wants the superseded work. Thanks for saying *when* the rebase was worth doing rather than letting me burn one two weeks early.

santifer (approved):
This is the standard I want on this surface, @WJiangH. Three things I want on the record because they're the reason it merges as-is:

- `tokensAreTotal` names a distinction that is invisible in the number itself. Summing totals over-counts and overwriting deltas under-counts, and both fail silently — putting the flag on the parser (the only place that knows the CLI's format) keeps the route CLI-agnostic instead of teaching it a vendor quirk.
- Returning `null` rather than `{tokens: 0}` when there's no usage block. A zero that means "I couldn't read it" sharing a slot with a zero that means "nothing to report" is the exact bug class we keep paying for here.
- Dropping `thought`. Not just to keep reasoning out of the report pane: forwarding it would mark a run as having produced text when it never produced an answer, which defeats the route's honesty gate.

Verified before approving: `merge-tree` clean against main; `parseGrokEvent`/`tokensAreTotal` exist nowhere in main and in no other open PR, so nothing here is a duplicate; 67/67 in `run-cli-support.test.mjs` and 391/397 across web (the 6 reds are `js-yaml` missing in my dependency-free worktree, not this change).

Sorry it sat three days past its clock — that queue is mine, not yours.

santifer:
Thanks @WJiangH — and sorry it sat three days past its clock; that queue is mine, not yours.

santifer:
Seven merges in and your name was not on the credits list, @WJiangH. That is a miss on my side, not a threshold you had not reached yet.

@all-contributors please add @WJiangH for code

PR #2690, opened 2026-08-11
fix(shared): let _profile.md define the archetypes, and allow "none of these"

WJiangH:
## The problem

`modes/_shared.md` says:

> Classify every offer into one of these types (or hybrid of 2)

That makes six archetypes a **closed set**, in the one file every mode reads.
They are the archetypes of the search this tool came out of (`AGENTS.md` →
*Origin* says as much), and they cover a narrow slice: a silicon
design-verification engineer, a quant, a clinician have no archetype in that
table at all.

The next line already says to read `_profile.md` — but only *afterwards*, for
framing. So detection happens against the author's taxonomy and the user's own
targeting arrives too late to affect it.

**Observed:** an ML-systems / inference engineer's evaluation of a
program-management req came back as *"Technical AI PM / AI Transformation Lead
(hybrid)"* — a faithful execution of "or hybrid of 2", and a fit narrative for
a career the candidate does not have.

The hybrid clause makes it worse: with no way to conclude *none*, the nearest
two labels always produce an answer.

## The change

This changes the table's **status**, not its contents.

- `_profile.md`'s archetypes take precedence where defined; the table becomes
  the fallback for what it does not cover. That is only this file's own stated
  rule — *"User customizations in `_profile.md` override defaults here"* —
  applied to the section that was ignoring it.
- An explicit **"matches nothing the user targets"** outcome, scored low on
  North Star alignment. A manufactured fit is worse than a low score, because
  it reads as analysis.

**Not a single table row changes**, so a user without a customized
`_profile.md` sees identical behaviour — the out-of-the-box path is untouched.

## Testing

`node test-all.mjs` → 3138 passed, 0 failed.

## Summary by CodeRabbit

- **Improvements**
  - Archetype detection now prioritizes profile guidance for more accurate classification.
  - Clarified that the default archetype table is non-exhaustive.
  - Hybrid classifications are limited to the user’s specified target archetypes.
  - Offers without a matching target archetype can be identified as “None of these” and receive the lowest North Star alignment score.
  - Removed the requirement to force every offer into a listed archetype or hybrid.
  - Reports now explain non-target roles using the closest matching archetype or note when no close match exists.

### DIFF
--- a/modes/_shared.md
+++ b/modes/_shared.md
@@ -165,6 +165,20 @@ Forcing it into the nearest available label — or into a "hybrid" of two —
 manufactures a confident fit narrative for a job the user is not applying for,
 which is worse than a low score because it reads as analysis.
 
+**A match against the default table below is not a match against the user's
+targets.** Where `_profile.md` defines archetypes, "targeted" means one of
+those. An offer that lands cleanly on a default row and on nothing in
+`_profile.md` is still an unmatched offer: name the default archetype if it
+helps explain the role, and score North Star as unmatched anyway. Reading the
+fallback as a target is the exact failure this section exists to stop — it is
+how a program-management req came back as "Technical AI PM / AI Transformation
+Lead (hybrid)" for a candidate who targets neither.
+
+On the number: `modes/ofertas.md` already anchors this dimension at
+`5 = exact target role, 1 = unrelated`. Unmatched sits at the bottom of that
+scale, not in the middle — the offer is not one the user is looking for, and a
+mid score reads as a partial fit that does not exist.
+
 | Archetype | Key signals in JD |
 |-----------|-------------------|
 | AI Platform / LLMOps | "observability", "evals", "pipelines", "monitoring", "reliability" |
--- a/modes/_shared.md
+++ b/modes/_shared.md
@@ -148,7 +148,22 @@ When a JD publishes a salary figure, distinguish advertised range, likely guaran
 
 ## Archetype Detection
 
-Classify every offer into one of these types (or hybrid of 2):
+Classify the offer by archetype. `modes/_profile.md` → *Your Target Roles* is
+authoritative: where it defines archetypes, detect against **that** table and
+use the one below only as a fallback for what it does not cover. This mirrors
+the precedence already stated above — user customizations in `_profile.md`
+override the defaults in this file.
+
+The table below is a default, not a closed set. It reflects one particular
+search (see AGENTS.md → Origin) and will not describe every user's field: a
+silicon design-verification engineer, a quant, a clinician have no archetype
+here at all.
+
+**If an offer matches no archetype the user actually targets, say so plainly
+and score North Star alignment low.** That is a real and useful signal.
+Forcing it into the nearest available label — or into a "hybrid" of two —
+manufactures a confident fit narrative for a job the user is not applying for,
+which is worse than a low score because it reads as analysis.
 
 | Archetype | Key signals in JD |
 |-----------|-------------------|
--- a/modes/oferta.md
+++ b/modes/oferta.md
@@ -41,7 +41,19 @@ If deeper company research is useful, recommend running `/career-ops deep` separ
 
 ## Step 0 — Archetype Detection
 
-Classify the job into one of the 6 archetypes (see `_shared.md`). If it is a hybrid, indicate the 2 closest ones. This determines:
+Classify the job against the archetypes in `_shared.md` — which means the user's
+own, from `modes/_profile.md` → *Your Target Roles*, where those exist. If it is
+a hybrid of two of the user's targets, indicate both.
+
+**"None of these" is a valid outcome and must be reported as one.** If the role
+matches nothing the user targets, do not pick the nearest label and do not call
+it a hybrid: say so, score North Star at the bottom of its scale, and continue
+the evaluation on the other dimensions — the rest of the report is still worth
+having, and a low alignment score with an honest reason is more useful than a
+confident fit narrative for a job the user is not applying for. A match against
+`_shared.md`'s default table alone is not a match against the user's targets.
+
+This determines:
 - Which proof points to prioritize in block B
 - How to rewrite the summary in block E
 - Which STAR stories to prepare in block F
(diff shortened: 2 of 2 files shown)


santifer:
Thanks for this, @WJiangH.

modes/_shared.md says to classify each listing into one of these types (or a hybrid of 2), which turns six archetypes into a closed set inside the one file every mode reads, and those six come from the job search the tool was born out of: a silicon verification engineer, a quant or a clinician have no archetype there. The observed consequence: an ML systems/inference engineer's evaluation of a program management req came back as 'Technical AI PM / AI Transformation Lead (hybrid)', a faithful execution of the rule and at the same time a fit narrative for a career the candidate doesn't have. It's +16/-1 and changes the table's status rather than its rows: _profile.md archetypes win where they exist (applying _shared.md's own line about customizations taking precedence), and an explicit "doesn't match anything the user is looking for" outcome appears, scored low on North Star.

WJiangH:
CodeRabbit's point, taken in half — `c82e893`.

**The gap it found is real and I had missed it.** The section said the default table is a fallback but never said whether landing on a fallback row counts as *targeted*. An evaluator can recognise "AI Solutions Architect" in the default table and score North Star as though the user were looking for it — which is precisely the failure this PR exists to stop, and how a program-management req came back as "Technical AI PM / AI Transformation Lead (hybrid)" for a candidate who targets neither.

Now stated explicitly: where `_profile.md` defines archetypes, "targeted" means one of those, and a default-only match is still an unmatched offer. Naming the default archetype to *explain* the role stays fine; scoring it as a fit does not.

**Not taken: a four-band numeric rubric here, plus conformance tests.** `modes/ofertas.md:7` already anchors this dimension — `5 = exact target role, 1 = unrelated`. Defining direct / adjacent / default-only / unmatched bands in `_shared.md` would be a second scale for the same dimension living in a different file from the one evaluators already read, and a rule stated in one place and contradicted in another is the shape of problem this PR is about. So the vagueness is removed by pointing at the existing anchor and saying which end unmatched sits at, rather than by building a scale beside it. If the rubric should be finer-grained, that is a change to the rubric across every dimension, not a +16-line change about archetypes.

Conformance tests declined on the same grounds: these scores come from a model reading prose, and the repo has no harness that asserts a scoring band out of a mode file. One added for this dimension alone would test the sentence, not the behaviour.

Verified under the merge boundary rather than the branch: `main` merged into a throwaway branch, full suite green.

santifer:
Sorry for the wait, @WJiangH. The observation is sharp:

> `Classify every offer into one of these types (or hybrid of 2)` makes six archetypes a **closed set**, in the one file every mode reads.

A closed vocabulary in the file every mode reads is a strong claim about the world — that six categories cover every job anyone will ever evaluate — and closed sets are exactly where a model has to coerce rather than admit. The evaluation doesn't get to say "none of these"; it picks the nearest and the report reads as confident.

That's the same shape as two other things tonight: `parseScore('N/A') === 0` (#2851, an unknown coerced into a value) and an open discard-reason vocabulary that can't be aggregated (#2785, where my direction was *closed enum plus an `other` escape that is never rolled into the stats*). Yours is the mirror: a closed set that needs an escape. **Same answer in both directions — the escape hatch is what keeps the taxonomy honest.**

**It's Santiago's call, not mine.** `modes/_shared.md` governs scoring for every evaluation in the product, and changing what the classifier is allowed to output changes every report. That's direction rather than checklist.

Queued for him with the argument above. One thing worth adding to the PR body for that conversation: **what a report says when the archetype is "none of these"** — if it renders as an empty field or a hedge, users will read it as the tool failing rather than as an honest answer, and the wording is the difference.

WJiangH:
Answering the report-wording question — and it turned up a hole in this PR, so `06edce7` closes both.

**The hole first, because it's mine.** This PR opened up `modes/_shared.md`, and `modes/oferta.md:44` — the file that actually runs an evaluation — still said:

```
Classify the job into one of the 6 archetypes (see `_shared.md`).
If it is a hybrid, indicate the 2 closest ones.
```

So the coercion this PR exists to remove was still being instructed, verbatim, one file over. That is the same shape you named on #2691 — a rule stated in one place and contradicted in another — and I had reproduced it while fixing it. Step 0 now classifies against the user's own archetypes where `_profile.md` defines them, and states that "none of these" is a valid outcome rather than a gap to fill.

**Now the wording, which is the part you flagged for Santiago.** You're right that it's the difference, and the two failure modes are exactly the ones you named. `**Archetype:** {detected}` in the report header had no guidance for the unmatched case, so it would render either blank — which reads as the tool breaking — or as "possibly a hybrid of X and Y", which *is* the coercion. Specified now as one of:

```
Not a target — closest default: {row from _shared.md's table}
Not a target — no close match
```

Two things that wording is trying to do. It **names something** rather than leaving a hole, so the field never reads as a failure to detect. And it keeps the default row visible as an explanation while denying it the status of a target: "closest default" places the role for the reader without implying the user is looking for it.

The rest of the instruction matters as much as the label: score North Star at the bottom of its scale, **and continue the evaluation on the other dimensions**. An off-target role can still be badly paid, remote-hostile or a legitimacy risk, and those findings are worth having. "Not a target" is a low alignment score with a reason, not an aborted report — which I think is also the difference between reading as honest and reading as broken.

On your framing — that this is the mirror of #2785's closed enum with an unaggregated `other`: agreed, and it sharpened what I wrote. An escape hatch that exists but is never rendered anywhere a human looks is the same as not having one. That's why this went into the report header rather than only into the scoring rule.

Verified under the merge boundary: `main` merged into a throwaway branch, `node test-all.mjs` green.

santifer:
Finding a hole in your own PR while answering a question about it, and closing both in one commit, is the behaviour I'd want from a reviewer, @WJiangH:

> **The hole first, because it's mine.** This PR opened up `modes/_shared.md`, and `modes/oferta.md:44` — the file that actually renders the report — still assumed the closed set.

That's the failure this change was most exposed to: widening what the classifier may output while the renderer still assumes six values means the escape hatch produces a report field nobody wrote handling for. Green tests, sensible diff, and a user sees something odd in a corner nobody looked at.

Still Santiago's call — `modes/_shared.md` governs scoring for every evaluation — and it's queued for him. The self-caught renderer half makes the case stronger, not weaker: it's evidence the change was thought through past its own diff.

PR #1994, opened 2026-07-17
fix(web): label-aware pipeline.md reader — posted:/trust:/note: never misread as columns

santifer:
## What

`readInbox` was purely positional (`parts[3]`=location, `parts[4]`=compensation), but the core appends **labeled segments** (`posted:`, `note:`, and `trust:` with #1765) after whatever positional shape a row has. On a 3-column row the inbox UI displayed *"posted: 2026-07-14"* as the **location** — a pre-existing web bug that #1765's `trust:` would have amplified. Surfaced while verifying the maintainer's additive-contract FYI for #1765.

## How

Any `word:`-prefixed segment at index ≥3 (URL cell exempt) is filtered out of positional assignment — **forward-compatible with labels the core hasn't invented yet**. Bonus: `posted:` now feeds the row's own `postedAt` (a more direct freshness signal for the inbox triage than the scan-history join, which stays as fallback).

## Verification

5 row shapes tested (3/4/5-col × posted/trust/note): all parse correctly, including the two that misrender today. typecheck + `next build` green.

Unblocks the ACK for #1765 (trust-segment) — with this, the web tolerates the new segment by construction.

Scope: `web/` only, 1 file, +28/−6.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved inbox parsing for entries with labeled details such as location, compensation, and posting date.
  * Posting dates are now displayed when provided in a valid date format.
  * Preserved explicitly provided posting dates, using scan history only when no date is available.

PR #2692, opened 2026-08-11
feat(dashboard): add a POSTED column showing requisition age

WJiangH:
> Two things in one branch: a `LastContact` fix (a bug) and a new optional
> column (a small feature). Happy to split, or to move the column to an issue
> first, if you'd prefer.

## The gap

The tracker already records when a requisition went live — `scan.mjs` writes
`posted: YYYY-MM-DD` into `pipeline.md` from the provider's `offer.postedAt`,
and it survives into the tracker notes. The dashboard had no way to surface it,
so **a req posted yesterday and one that has sat open for five months looked
identical** in the list.

That distinction matters when deciding what to spend an application on: a stale
listing is often a pipeline-filler, or was quietly filled long ago.

## The change

- `PostedOn` on `CareerApplication`, parsed in `deriveNoteFields` from a
  `posted <date>` / `Posted: <date>` note. Follows the existing
  derive-from-notes convention, so **no tracker schema change**.
- New optional **POSTED** column, off by default, toggled via the `C` picker.
  Renders age colour-coded: green ≤7d, yellow ≤30d, peach ≤90d, red beyond.
- **Bug fix:** the posting date is now excluded from `LastContact`. When a req
  went live is not an interaction with the company, and letting it through
  showed a freshly-posted role as though it had just been touched.

The `(POSTED)` pay-source marker is unaffected — it carries no date, so the new
regex cannot match it.

## Testing

Three new derive cases in `dashboard/internal/data/derive_test.go`, including
the `(POSTED)` marker. Go tests pass; `node test-all.mjs` → 3138 passed.

## Summary by CodeRabbit

- **New Features**
  - Added an optional pipeline column showing how long each requisition has been open.
  - Posting dates are displayed using relative formatting and age-based colors.
  - Added localized column labels in English, Turkish, and Spanish.
  - Tracker notes now preserve available posting dates.

- **Bug Fixes**
  - Posting events are excluded from most-recent-contact calculations.
  - Missing, invalid, or future posting dates display as a dash.
  - Pipeline column widths remain consistent across terminal sizes.


santifer:
Thanks for this, @WJiangH.

The tracker already stores when a req went up (scan.mjs writes `posted: YYYY-MM-DD` from the provider's offer.postedAt and it survives in the notes), but the dashboard had no way to show it, so a req posted yesterday and one open for five months looked identical in the list. You add an optional POSTED column, off by default and toggled from the C selector, colored by age (green ≤7d, yellow ≤30d, peach ≤90d, red beyond that), derived from the notes so the tracker schema isn't touched. You also fix a real bug on the way: the posted date was leaking into LastContact, so a freshly posted listing looked like you had just been in touch with the company.

santifer:
Solid first PR mechanically, @WJiangH: it builds and vets clean, the tests include the negative case, the column ships off by default, `visibleCols` is not persisted so adding a ColumnID breaks nobody's saved preferences, and all three locales are covered. Four things, and only the first is a blocker.

**1. The column has no producer, so it renders "—" for everyone.** The PR argues the `posted:` segment "survives into the tracker notes". It does not: `scan.mjs:1623` writes `| posted: YYYY-MM-DD` into `data/pipeline.md`, and the dashboard reads `data/applications.md` (`career.go:52-55`). Nothing carries the segment across - `grep posted merge-tracker.mjs` returns zero, `modes/tracker.md` never mentions the convention, and no tracker row in the repo has it. Pick one: wire the producer (pipeline.md `posted:` into the Notes cell on the path that creates the row), or document `posted <YYYY-MM-DD>` in the Notes column of `modes/tracker.md` so the modes emit it.

**2. A measured regression in LastContact.** `derive.go:214` strips the pattern globally from free-text Notes. With the note "Recruiter posted 2026-07-20 update on the req" and Date 2026-06-01: baseline LastContact is 2026-07-20, with the patch it becomes 2026-06-01 and the date is relabeled as a posting date. A real recruiter interaction vanishes from the last-contact calculation. Anchor the pattern (require start of line or segment, e.g. `(?:^|[;|\n])\s*posted`) instead of `\bposted` anywhere.

**3. Row width hits the terminal edge.** `pipeline.go:1572` uses a fixed `- 14` that does not scale with visible segments: at width 200 the row measures 193 with zero optional columns and grows +1 per column, reaching exactly 200 with all seven. CodeRabbit's minimal patch works; deriving the overhead from the visible-segment count is the real fix, since the drift predates you and this PR just takes it to the limit.

**4. Two edges worth a line of test:** `posted:2026-07-15` with no space does not match (the regex requires `\s+`), so it neither populates PostedOn nor stops counting as contact; and with two "posted" dates in one note the first wins rather than the most recent.

Separately, and not about this PR: you have eleven open at once. The content is clearly not PR farming (each traces to an issue you filed with the diagnosis already correct), but that volume means your work queues behind itself and my review latency multiplies across all of them. If you have a preferred order, tell me and I will follow it.

WJiangH:
All four addressed in `5277c44`, plus CodeRabbit's future-date case. Taking them in your order.

**1. The column had no producer — you were right, and the PR body was wrong.** I claimed the `posted:` segment "survives into the tracker notes"; nothing carries it across, and I never checked, because on my own tracker the rows I was looking at had come from a path where I'd typed the note by hand. It rendered "—" for everyone else.

I wired the producer rather than only documenting the convention, at both paths that create a row: `modes/oferta.md` (single evaluation) and `batch/batch-prompt.md` (the batch TSV's `notes` field). And I documented the reserved segment in `modes/tracker.md` as well, since the consumer side needs to be readable too — a convention that only exists in the emitter is one refactor away from being lost.

Both producer rules are explicit that it is **copy-verbatim-or-omit**: only from a pipeline entry's `| posted:` segment, never inferred, and never today's date as a stand-in. An absent date renders `—`, which is honest; a guessed one reports a months-old req as fresh, which is worse than the blank column I shipped.

**2. The LastContact regression reproduces exactly as you describe.** Your note and date give 2026-07-20 → 2026-06-01, and what makes it worse than a mislabel is that the same pattern feeds the strip: an over-broad match doesn't just tag a date wrong, it deletes a real interaction from the calculation.

Anchored to a line start or a `;`/`|` separator, as you suggested. One addition: the strip now preserves the anchor (`"$1"`), because removing a mid-note segment along with its separator welds its neighbours together — there's a test for that specific shape. Your exact case is now a test too, named for what it protects.

**3. Row width.** I took the real fix rather than CodeRabbit's `overhead++`. The overhead is one space per gap from `strings.Join`, one leading space, and `Padding(0, 2)`, minus the score cell's declared-vs-rendered slack — so it is a function of the visible segment count, and the old `14` was simply the all-columns-on value. Your measurements drop out of it: at width 200 the no-optional-columns row is 7 short, and it closes by one per column.

Pinned with a sweep over all 2^7 combinations at four widths rather than the endpoints, because an off-by-one that only shows up at one particular count is invisible to an endpoint check. With the constant restored, that sweep goes red at every count. The role floor of 15 is excluded from the assertion and pinned separately — below the crossover the row is *meant* to overflow.

**4. Both edges pinned.** `posted:2026-07-15` now matches (the colon form takes no space; the bare form still requires one, so `posted2026-07-15` stays inert on both paths), and with two dates the most recent wins — a re-post replaces the requisition, so taking the first would pin the row to one that no longer exists.

**Also, from CodeRabbit on the same file:** a future `PostedOn` made `time.Since` negative, which landed in the first colour case and painted it green — the one colour that says act on this now. It renders as unavailable instead.

`dashboard`: `go vet` clean, `gofmt` clean, `go test ./...` ok. Merged onto today's `main`, `node test-all.mjs` → 3664 passed / 0 failed.

**On the eleven open PRs:** fair, and the fault is sequencing — one pass over the repo, filed as I went, instead of pacing them. I've put a suggested order on #2689 and I'm not opening new ones until the queue is down.

santifer:
First, the thing you should not spend another minute on: **the red X on this PR was never yours.**

`test (windows-latest)` is the only failing check, and the failure is `agent-inbox-tests.mjs` section 7, `kept=29 of 30`. That is a concurrency flake in a file your PR does not touch. @Scott-Emberson root-caused it today in #2824 and the fix landed a few hours ago in #2825: lock acquisition is a retry lottery rather than a fair queue, so under 30-way contention on the slow Windows runner one writer times out and its item is lost. I re-ran your job against the fixed base.

It was not only you. Nine open PRs from nine different people were red for this same reason, and I have re-run all of them. If you looked at that X and assumed you had broken something on Windows, that assumption was reasonable and wrong, and the cost of it lands hardest on someone who has not had anything merged yet.

**On your four fixes:** point 1 is the one I want to name. You could have quietly wired the producer and moved on. Instead you wrote that the PR body claimed something you had never checked, and that it worked on your machine because you had typed the note by hand. Saying that out loud costs you nothing here and it is the reason I will read your next PR faster.

**Now the harder half, and I would rather say it than let you find it out by waiting.**

You have eleven PRs open and none merged. Four of them are web changes that belong to a different reviewer, so my queue is seven. That is more than I can turn around at the rate you are opening them, and the failure mode is not that I ignore them: it is that every PR I merge conflicts the others, and you end up rebasing the same work repeatedly for reasons that have nothing to do with its quality. Someone else in this repo has rebased a single PR three times for exactly that reason, and I do not want to do that to you eleven times over.

So, a request rather than a rule: **hold off on opening new ones until some of these land.** I would rather get you a first merge this week than have you accumulate fourteen open branches. If you tell me which one you most want landed first, I will take that one next once its CI comes back green.

For what it is worth: the titles on those eleven are not the titles of someone farming contributions. They read like someone who ran this tool hard for two days and found eleven real things, which is a compliment and also exactly why the queue matters.

santifer:
Correction to what I told you earlier, because it was wrong in a way that matters.

I said I had "re-run your job against the fixed base". **`gh run rerun` replays the same head SHA**, so those re-runs went back over the *old* merge commit, without #2825 in it. Verified rather than assumed: the run I re-ran carries `headSha=4721209b`, and `git merge-base --is-ancestor 513f07a2 4721209b` says the fix is not an ancestor of it.

Five of the ten came back green anyway, and that was luck, not the fix: the flake is probabilistic, so a re-run of the same base wins the lottery about half the time. Four stayed red for the same `kept=29 of 30`.

The right action was to give each branch an actually-newer base, so I have run `gh pr update-branch` on the ones still red. #2782 could not take it because it conflicts with `main`; @briacSck, that one needs a rebase from you.

**And the more important half.** Right after merging an unrelated one-line docs change, the same failure fired on `main` **with #2825 already in it**, on my own machine, under heavy parallel load. I then failed to reproduce it in 38 further attempts, including a replica that spawns the 30 processes exactly the way the test does and two full suites running alongside it.

So the honest state is: the burst-sized budget cut this a long way and **did not remove it**. That is what both #2825 and the jitter comment before it said would happen, in writing, and it is the reason the structural answer is a fair queue rather than a bigger retry lottery. #2777 stays open on the strength of that, and it is now the third time I would have been wrong to close it.

If you see this red on your PR again, it is still not yours.

WJiangH:
Done, and thank you for #2749 and #2750 — you said you'd rather get me a first merge this week and then did it the same night.

**Your queue is three.** I've closed #2730, #2731, #2732 and #2733 myself. They were one group — scan-filter observability — and holding four spots for a related group was the cost you described. Branches stay on my fork so they reopen as themselves rather than as new numbers, and the four issues stay open, so nothing about them is lost except the queue position. There's a fair chance they come back as one PR rather than four; the pieces are closer to each other than to anything else.

**What's left, and it's all ready rather than waiting on me:**

1. **#2692** (this one) — your four points answered in `5277c44`.
2. **#2691** — CodeRabbit's three answered in `b92ecf7`. Two taken, including one worth flagging: the "non-blocking" test proved nothing, because the fixture was missing `cv.md`/`config/profile.yml`/`portals.yml` and so `onboardingNeeded` was already true for an unrelated reason. It could not have caught the regression it was written for. Now seeded and asserting `onboardingNeeded === false`.
3. **#2690** — CodeRabbit's one answered in `c82e893`, half taken. The real gap was that a default-table match wasn't stated to be *not* a target match, which is the same failure this PR is about; the four-band numeric rubric it also asked for is declined, since `modes/ofertas.md` already anchors that dimension and a second scale in a second file is the shape of problem being fixed.

**One correction I owe you.** An hour before your routing note I sent you an order with #2688 at the top — a `area:web` PR you can't merge. Sorted properly it's the three above for you, and #2688 / #2689 for the web reviewer. Nothing needed there; the 34-open / 14-day-median number told me the useful thing, which is to stop adding to it.

Also: your `gh run rerun` correction changed how I verify. I had been running on the branch and calling it green. I now merge today's `main` into a throwaway branch and run there — it's what the required check actually runs, and it immediately surfaced a real conflict on #2689 the branch alone could never have shown. Same shape as your re-run: the thing being measured wasn't the thing that would be judged.

Not opening anything new. I have a change ready for #2751 that's sitting until this clears.

santifer:
Anchoring `rePostedOn` to a segment instead of `\b` is the decision that makes this safe, and the reason you give is the right one: that date gets SUBTRACTED from the last-contact scan, so an over-broad match doesn't mislabel a field, it deletes a real interaction. `"Recruiter posted 2026-07-20 update on the req"` silently moving LastContact backwards is precisely the kind of bug that never gets reported — the number just looks a bit older and nobody knows why.

`"$1"` in the replacement, keeping the separator so stripping a segment can't weld its neighbours together, is the second-order version of the same care. And latest-wins on two posting segments is right for what the column asks: a re-post means the earlier req is history, and first-wins would pin the row to something already replaced.

Two things I verified rather than assumed:

- the segment already exists upstream — `scan.mjs:1629` writes `| posted: ${posted}` into `pipeline.md` from the provider's `postedAt`, and `modes/pipeline.md:83` documents it. So this isn't a new field; it's a field that was being collected and thrown away at the tracker boundary. That answers the "no linked issue" question better than an issue would have.
- `modes/oferta.md` is a consumed surface, so I ran it past the web before merging. It's additive there too — their Notes reader takes the whole cell and has been label-aware since #1994, so `posted:` was already being surfaced as `postedAt` on their side.

One consequence worth naming, since it's yours by lineage even though it isn't your job: the web WRITES tracker rows as well as reading them, and its prompt doesn't emit this segment yet — so offers evaluated through the web won't carry a posting date until that side catches up. They've picked it up and are shipping the matching change.

Merged 🚀

PR #2697, opened 2026-08-11
fix(ci): skip the upgrade leg when nothing managed changed, instead of failing

santifer:
## The bug

`upgrade-tests.mjs` throws when no `SYSTEM_PATHS` file changed between the last tag and the target:

```
Error: No changed system file between career-ops-v1.26.0 and target — nothing to upgrade, leg would be vacuous
```

The reasoning behind that check is right — a leg with no managed file to qualify proves nothing. **Treating it as a FAILURE is the bug.** `web/` is deliberately outside `SYSTEM_PATHS`, so a web-only PR legitimately changes nothing `apply` manages, and the gate goes red on a PR that is not broken in any way.

Measured on 11 Aug: **three PRs blocked at once** — #2665 (the fix for a HIGH severity advisory: js-yaml 4.3.0 → 4.3.1), #2668 and #2654. A security fix could not land because the gate could not test it.

## The fix

`pickOracle` returns `null` instead of throwing; the caller skips the leg and says so. **"I cannot test this" and "this is broken" must not share an exit code.**

Two things kept deliberately:

1. **The skip is not a pass.** `--pr-gate` now prints three states, never two: `GREEN` / `SKIPPED: nothing in the managed set changed — this leg qualified nothing` / `RED`. A skipped leg reported as GREEN would be a pass that never ran, which is the exact shape this suite exists to catch elsewhere.
2. **A suspicious shape still fails.** If root-level files changed and *none* is managed, that likely means the `SYSTEM_PATHS` read is wrong (a renamed constant, a bad parse) — so it throws rather than skipping.

## Verified by running all three paths

| case | expected | result |
|---|---|---|
| web-only change (#2665's branch) | skip, exit 0 | `SKIP … (1 file changed, all outside the managed set)`, exit 0 |
| managed file changed (`scan.mjs`) | run the leg | 0 skips, full leg, `GREEN` |
| only an unmanaged root file | fail loudly | exit 1, `Root files changed (probe-unmanaged.txt) but none is in SYSTEM_PATHS — refusing to skip` |

`node test-all.mjs` → 3276 passed, 0 failed.

Credit where due: the gate itself (#2642) is good work and the non-vacuity oracle is the right idea — this only changes what happens when the oracle legitimately finds nothing.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of changes that do not affect managed files.
  - Web-only and otherwise unmanaged changes are now skipped appropriately.
  - Changes to unsupported root files still correctly fail validation.
  - Validation results now clearly distinguish **SKIPPED**, **RED**, and **GREEN** outcomes.

### DIFF
--- a/upgrade-tests.mjs
+++ b/upgrade-tests.mjs
@@ -76,8 +76,18 @@ function pickOracle(mirror, oldTag, targetSha, systemPaths) {
     if (!managed(f)) return false;
     try { git(mirror, 'cat-file', '-e', `${targetSha}:${f}`); return true; } catch { return false; }
   });
-  if (!candidate) throw new Error(`No changed system file between ${oldTag} and target — nothing to upgrade, leg would be vacuous`);
-  return candidate;
+  // No managed file changed. That is NOT a failure by itself: a web-only PR
+  // (web/ is deliberately outside SYSTEM_PATHS) legitimately changes nothing
+  // `apply` manages, so there is nothing to qualify. Returning null lets the
+  // caller SKIP. Throwing here made the gate red for every web-only PR, which
+  // on 11-ago blocked three at once — including the fix for a HIGH severity
+  // advisory. "I cannot test this" and "this is broken" must not share an exit.
+  //
+  // The caller still fails loudly when ROOT files changed and none is managed:
+  // that shape means the SYSTEM_PATHS read is probably wrong, which is exactly
+  // what this oracle exists to catch.
+  if (!candidate) return { oracle: null, changed };
+  return { oracle: candidate, changed };
 }
 
 function isAncestor(cwd, tag, sha) {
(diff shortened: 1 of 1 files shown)


santifer:
Self-merged, because this was blocking three PRs for a reason none of them caused.

The upgrade gate threw when no `SYSTEM_PATHS` file changed between the last tag and the target. But `web/` is deliberately outside `SYSTEM_PATHS`, so **every web-only PR was failing a check it could never pass** — including #2665, which carries the fix for a HIGH severity advisory. A security patch could not land because the gate could not test it.

The leg now skips instead of throwing, with two things kept on purpose:

1. **A skip is not a pass.** `--pr-gate` prints three states now, never two: `GREEN` / `SKIPPED: nothing in the managed set changed — this leg qualified nothing` / `RED`. Reporting a skipped leg as GREEN would be a pass that never ran.
2. **A suspicious shape still fails.** Root-level files changed with none of them managed → it throws, because that usually means the `SYSTEM_PATHS` read is wrong.

Credit to @Scott-Emberson, whose dynamic upgrade gate (#2358) this builds on: the non-vacuity oracle is the right idea, and only its failure mode needed changing. Fitting detail — **the canary in that same suite is what caught a bug in this very patch**: I changed `pickOracle`'s return shape and missed one of its three returns, and the one test written to prove the gate can go red was the only path that exercised it.

Verified by running all three paths, not by reading: web-only → skip, exit 0 · managed file changed → full leg, GREEN, zero skips · unmanaged root file only → exit 1 with the guard message. Suite 3276/0, CI green on all three platforms.

Issue #14, opened 2026-04-06
Does it support other tools like copilot or gemini?


gaurav9822:
### Problem

NA

### Proposed solution

NA

PR #2698, opened 2026-08-11
fix(deps): patch both HIGH advisories in web (js-yaml 4.3.1, nanoid 3.3.18)

santifer:
Closes both open HIGH-severity Dependabot advisories on `web/package-lock.json`. Lockfile only — no source change, no behaviour change.

| alert | package | installed | patched | now |
|---|---|---|---|---|
| #14 | js-yaml | 4.3.0 | 4.3.1 | **4.3.1** |
| #15 | nanoid | 3.3.16 | 3.3.17 | **3.3.18** |

- **#14** — quadratic CPU consumption resolving `!!omap`. The 4.x branch is patched at 4.3.1, so the major bump to 5.x (#2631) is not required to clear it.
- **#15** — custom generators can loop indefinitely when size is zero. Reached transitively via `postcss` → `next`; nothing in the repo imports it directly.

Generated with `npm audit fix --package-lock-only`, which reports `found 0 vulnerabilities` afterwards.

**Verified:** `npm ci` clean · `npm test` 123 pass / 0 fail · `npm run typecheck` clean · `npm run build` succeeds.

Note: #2665 (Renovate) carries the js-yaml half of this on its own. Either can land — this one closes both alerts in a single lockfile state, which is how npm resolves it anyway.

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -1653,9 +1653,9 @@
       }
     },
     "node_modules/js-yaml": {
-      "version": "4.3.0",
-      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.3.0.tgz",
-      "integrity": "sha512-1td788aAnnZ5qs7V2QIRl1owjtYpbKt749Y3xauqQgwIIGF/xXWz1wMTEBx5O3LK3lXLVuqXPdPxj2BoFHaW9Q==",
+      "version": "4.3.1",
+      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.3.1.tgz",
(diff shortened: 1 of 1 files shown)


santifer:
Closes both open HIGH-severity advisories on the web lockfile: js-yaml 4.3.0 → 4.3.1 (quadratic CPU in `!!omap` resolution) and nanoid 3.3.16 → 3.3.18 (custom generators can loop forever at size zero). Lockfile only, no behaviour change; `npm audit` reports zero vulnerabilities afterwards, and web is green — 123 tests, typecheck and build clean.

Worth noting for the record: these could not be patched until #2697 landed. The upgrade gate was failing every web-only PR, so the fix for a HIGH advisory sat blocked behind a check it could never pass.

PR #2700, opened 2026-08-11
test(web-contract): freeze the company-key parity and its source per surface

santifer:
The last piece of #2666. The bug is fixed in three surfaces; this is what stops it coming back.

## Why an assertion at all

`web/src/lib/core/normalize-text-key.mjs` is a **copy by necessity** — the browser cannot reach the user's checkout, so a client component can't import the core. It is declared as a copy and documented as one. But a conscious copy with no assertion behind it is exactly what this repo has now shipped five times: `states.ts`, `CADENCE_DEFAULTS`, the doctor prereqs, the #2590 cache, and the three divergent company keys of #2666 itself.

## 55.7 — parity: core vs the mirror

Compares `normalizeTextKey` from `tracker-parse.mjs` against the web mirror over a shared 15-case corpus × 2 separators.

Two deliberate choices:

- **Core vs MIRROR, not core vs the derived path.** The server path imports the core, so it *cannot* diverge — asserting on it would guard the one thing that can't break while ignoring the one that can. The mirror is also the path nobody exercises in normal use (it only runs on a partial checkout), which is why it needs a test rather than usage.
- **Expected values are derived from the core, never hand-written.** A hand-written expectation freezes today's answer and stops tracking the source.

The corpus (`tests/fixtures/company-key-corpus.json`) is one shared file rather than a list on each side, since two lists drift — the failure this whole thread is about.

## 55.8 — where each surface gets its key

Shapes are not interchangeable: the server can reach the live core and must derive from it; a `"use client"` component physically cannot and imports the shared mirror. What none of them may do is roll its own normalizer.

The ASCII-class check is scoped to those three files on purpose: a repo-wide grep for `[^a-z0-9]` measures **71% false positives** here (domain slugs, filename slugs, a regex boundary class — all legitimate), and a check that shouts at correct code gets silenced. It also skips comment lines, because the honest fix ships a comment *warning against* the pattern — that false positive fired on the first run.

## Verified by mutation, not by reading

| mutation | result |
|---|---|
| mirror reverted to `[^a-z0-9]` | RED — 14 cases, e.g. `core="nestlé" mirror="nestl"` |
| client defines its own `norm` again | RED — names the file and what it must do instead |
| corpus thinned (Škoda + CJK removed) | RED — *"the parity check above would pass without exercising the failure it exists for"* |

That third one was a real hole found by mutating: the original guard-of-the-guard queried the core directly, so an emptied corpus passed vacuously. It now checks both directions — the rules AND the corpus.

`node test-all.mjs` → 3281 passed, 0 failed.

## Summary by CodeRabbit

* **Tests**
  * Added automated checks to ensure consistent company and role matching across core and web implementations.
  * Added regression coverage for accented, Unicode, CJK, fullwidth, ligature, casing, control-character, empty, and nullish inputs.
  * Added validation that required web surfaces use the correct normalization logic and support the documented character ranges.

### DIFF
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -12969,6 +12969,120 @@ try {
       }
     }
   }
+
+  // 55.7 The company/role matching key: core vs the web's declared mirror (#2666).
+  //
+  // The browser cannot reach the user's checkout, so web/src/lib/core/
+  // normalize-text-key.mjs is a COPY by necessity, declared as such. A conscious
+  // copy without an assertion is exactly the failure this repo has now hit five
+  // times (states.ts, CADENCE_DEFAULTS, doctor prereqs, the #2590 cache, and the
+  // three divergent company norms of #2666 itself).
+  //
+  // Compare CORE vs MIRROR, not core vs the derived path: the derived path
+  // imports the core, so it CANNOT diverge — asserting on it would guard the one
+  // thing that cannot break while ignoring the one that can. The mirror is also
+  // the path nobody exercises in normal operation (it only runs on a partial
+  // checkout), which is precisely why it needs a test rather than usage.
+  //
+  // Expected values are DERIVED from the core, never hand-written: a hand-written
+  // expectation would freeze today's answer and stop tracking the source.
+  const corpusPath = join(ROOT, 'tests', 'fixtures', 'company-key-corpus.json');
+  const mirrorPath = join(ROOT, 'web', 'src', 'lib', 'core', 'normalize-text-key.mjs');
+  if (!existsSync(join(ROOT, 'web', 'src'))) {
+    warn('web/ not present in this checkout — skipping the company-key parity freeze (#2666)');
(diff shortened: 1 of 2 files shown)


santifer:
Closes the loop on #2666. The bug is fixed in three surfaces; this is what keeps it fixed.

The web mirror of `normalizeTextKey` is a copy by necessity — a client component can't reach the user's checkout — and a conscious copy without an assertion is precisely what this repo has shipped five times now. §55.7 compares the core against the mirror over a shared corpus, deriving the expected values from the core rather than hand-writing them. §55.8 pins where each surface gets its key: the server derives from the live core, the client imports the shared mirror, and none of them rolls its own.

Verified by mutation rather than by reading: reverting the mirror to `[^a-z0-9]` goes red with 14 concrete cases, a client redefining its own `norm` goes red naming the file, and thinning the corpus goes red saying the parity check would otherwise pass without exercising the failure it exists for. That last one was a real hole the mutation found — the guard-of-the-guard queried the core directly, so an emptied corpus passed vacuously.

PR #2701, opened 2026-08-11
fix(archive): key JD captures to report numbers so they survive the capture day

### LINKED ISSUE #2674: Archived job postings are unresolvable after the day they are captured
### Code of Conduct

### Existing issues

### What happened?

Archived job postings are effectively unreachable after the day they are captured, which defeats the purpose of archiving them. The archive is meant to preserve requirements for interview prep and salary negotiation after a posting is filled or pulled, but the only code that consumes an archive cannot find one unless it was created during the same run.

`archive-posting.mjs` names its output from volatile inputs and `outcome.mjs` recomputes that name to find it:

- `archive-posting.mjs:237` builds `${today()}_${slugify(company)}_${slugify(role)}.pdf` and line 239 prints `local:jds/${filename}` as the reference to paste into `data/pipeline.md`.
- `outcome.mjs` then tries exactly two ways to resolve a capture, and both fail in normal use.

Case 1, `outcome.mjs:309`, matches `local:(jds\/[^\s|)]+)` against the tracker's notes column. That only works if a human manually pasted the reference printed by the archiver into that column. Nothing in the codebase writes it there, so on a tracker built by the normal flow this branch never runs.

Case 2, `outcome.mjs:318-334`, re-invokes `archive-posting.mjs` on the posting URL and then looks for `${today()}_${slugify(company)}_${slugify(role)}.pdf`. Because the expected filename is recomputed from today's date and from company and role strings that are re-scraped each time, it resolves only a capture made today, from a URL that is still live.

The combination is the problem. A posting that is still live does not need an archive, and a posting that has gone dead cannot produce one, so `outcome.mjs` writes the `posting_missing.md` stub instead. Captures sitting in `jds/` from any earlier day are never considered, whatever their filename.

Two related points, since they constrain the fix:

The date is not the only volatile part. Company and role are re-derived from the page title or `h1` on each run, so a site that changes its title markup, or a fallback that lands on a different string, produces a different filename for the same posting.

Nothing keys a capture to the report number, even though the report number is the stable identifier already used by `reports/{###}-{slug}-{date}.md` and by column 1 of the tracker. In a working install I see a meaningful number of captures that were manually named with a leading report number, in both zero padded and unpadded forms, which suggests report keying is the convention users already reach for by hand.

I would also note that `AGENTS.md:373` presents `local:jds/{file}` as the way JDs are referenced from `data/pipeline.md`, which is accurate as a convention. The convention itself is fine and is used elsewhere, for example the Apify plugin sets `local:` URLs. The gap is that the archiver's own output never reaches a consumer through it.

### Steps to reproduce

1. Archive any live posting: `node archive-posting.mjs "<some live job url>"`. Note the file written to `jds/` and the `local:jds/...` reference printed.
2. Do not paste that reference anywhere, since no automated step does.
3. On any later day, run `node outcome.mjs <selector> <type>` for the corresponding tracker row.
4. Observe that the archive from step 1 is not found. Case 1 finds nothing in the notes column, and Case 2 recomputes a filename with the new day's date. The outcome directory gets `posting_missing.md` rather than the capture that exists on disk.

Step 3 fails identically if the posting URL has since gone dead, which is the case the archive exists to cover.

### Expected behavior

A capture already present in `jds/` should be found for its tracker row regardless of the day it was made, and regardless of whether the original URL is still reachable.

Concretely, I would expect the archiver to be able to key a capture to a report number at write time, and the resolver to look a capture up by that stable key rather than by reconstructing a name out of today's date and freshly scraped strings.

### CLI tool

Claude Code

### OS and Node.js version

macOS 15 / Node 22

### Output of npm run doctor

```shell
Omitted deliberately. doctor output reports on local CV, profile and pipeline
files, and this issue does not depend on local setup. The defect is in the
naming and resolution logic in archive-posting.mjs and outcome.mjs and is
reproducible from a clean install.
```

rubicon:
## What does this PR do?

Makes an archived job posting findable after the day it was captured. `archive-posting.mjs` gains `--report=N`, which keys the capture to a report number, and a new `jd-capture.mjs` resolves a capture by that number so `outcome.mjs` can find one written weeks earlier or whose posting URL has since died.

## Related issue

Closes #2674

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

---

## The bug

`archive-posting.mjs` names its output `${today()}_${slugify(company)}_${slugify(role)}.pdf`, and `outcome.mjs` finds it again by rebuilding that same string. Every input to that string is volatile. The date advances, and company and role are re-scraped from the page title or `h1` on each run, so the name `outcome.mjs` computes stops matching the name that was written.

`outcome.mjs` has two resolution paths and neither survives a normal workflow. The first reads a `local:jds/...` reference out of the tracker Notes cell, which only exists if someone pasted by hand the reference the archiver printed. Nothing in the tree writes it, so the branch does not run. The second re-invokes the archiver and then looks for today's recomputed filename, which requires the posting URL to still be live.

Those two conditions cannot both hold when it matters. A posting that is still live does not need an archive, and one that has gone dead cannot produce a fresh capture, so `outcome.mjs` writes the `posting_missing.md` stub while the capture sits in `jds/` unread.

The numbers from the install where I found this: 478 captures in `jds/`, 580 tracker rows, and **not one capture reachable** through `outcome.mjs`. `local:jds/` appears 0 times in `data/pipeline.md` and 0 times in `data/applications.md`, which is the first path never firing. And `jds/` holds 0 PDFs and 0 filenames containing an underscore, so in months of use the archiver's own naming scheme has produced no surviving file at all.

## The fix

The report number is already the stable identifier in this project. It keys `reports/{###}-{slug}-{date}.md` and column one of the tracker, and unlike a scraped title it does not change between runs. It is also what users reach for on their own: 79 of those 478 captures had been named by hand with a leading report number, in both padded and unpadded forms, without anything in the tool asking them to. Keying on it makes all 79 resolvable with no renaming and no migration step.

`archive-posting.mjs --report=N` prefixes the capture with the zero-padded number. `jd-capture.mjs` resolves a capture for a report by matching the leading numeric run and comparing the parsed integer, so report 64 finds `064-...` and never picks up `640-...`, and a number appearing later in a filename is not matched at all. Unpadded prefixes resolve too, because captures named by hand tend to use them. `outcome.mjs` consults the resolver before falling back to re-archiving, and passes `--report` when it does archive so the new capture is keyed for next time.

Omitting `--report` leaves filenames and behavior exactly as they are today, so nothing existing changes shape.

One related fix comes along with it. `outcome.mjs` hardcoded `posting.pdf` as the destination when copying a capture into the outcome directory. That was harmless while only PDFs were reachable, but report keying also makes `.txt` and `.md` captures resolvable, and copying those under a `.pdf` name would misrepresent them. The destination now follows the capture's own extension.

## Notes on the details

The CLI parser ignores unknown flags and treats the first bare argument as the URL, so a space-separated `--report 42` would have been dropped silently and `42` taken as the URL. Both flag forms are consumed explicitly and a malformed value exits non-zero rather than being ignored. `--report` is rejected together with `--pipeline`, since one report number cannot key a batch of different postings.

Resolution order is deliberate. An explicit `local:jds/` reference in the Notes cell still wins, because that is a human statement of intent. Report keying comes second, and re-archiving last. Where a report has more than one capture, the resolver returns the full candidate list and takes the most recent by default; recency is a deterministic tie-break rather than a claim about quality, and an optional company-slug hint overrides it.

One ambiguity is worth naming: a company slug that begins with digits, such as `360-learning-...`, is indistinguishable from a report number by filename alone. The company-slug hint is the escape hatch, and there is a test that documents the behavior rather than pretending it does not exist.

## Testing

`node test-all.mjs` passes with 3314 tests and no failures.

New coverage in `tests/jd-capture.test.mjs` exercises padded, unpadded and single-digit prefixes, `.txt` and `.md` captures, the 64 against 640 case, a number embedded mid-filename, a missing directory and a non-numeric report returning null rather than throwing, multi-candidate ordering, and the company-slug override.

`tests/outcome.test.mjs` gains an end-to-end case: a capture dated 30 days ago on a tracker row with no URL, so the re-archive path cannot run. I checked that this test fails against the unfixed `outcome.mjs` rather than passing for free.

`test-all.mjs` covers both `--report` flag forms, the regression where the space-separated form swallowed the URL, rejection of a malformed value, rejection alongside `--pipeline`, and that omitting the flag leaves the filename unchanged.

## Summary by CodeRabbit

- **New Features**
  - Added report-number support for single-posting captures, including padded and unpadded filenames.
  - Automatically reuses matching captures and preserves their original file extensions when archiving.
  - Supports reliable capture selection by report number, recency, and optional company matching.

- **Bug Fixes**
  - Prevents partial or embedded report-number matches.
  - Validates invalid report values and incompatible command options.

- **Documentation**
  - Updated usage guidance and file inventories for report-based captures.


rubicon:
Pushed a follow-up commit fixing a defect I found in this PR while using it against my own tracker. Flagging it explicitly because it changes an assertion that was in the original diff.

**The defect.** `findCaptureForReport` matched on the report number and treated `companySlug` as a preference — when the hint matched no candidate, `hinted ?? byRecency[0]` fell through to the most recent one. Filenames that begin with digits parse as report numbers, so a capture named `04-mission-preborn-vp-marketing.txt` resolves as report 4. Asking for report 4 with `companySlug: 'weave'` returned Mission Preborn's posting:

```
findCaptureForReport('jds', 4, {companySlug: 'weave'})
  -> { filename: '04-mission-preborn-vp-marketing.txt', ... }
```

**Why returning nothing is better than returning that.** `outcome.mjs` passes the row's company slug, copies the result into the outcome directory as `posting.{ext}` (the permanent record of what was applied to), and counts a hit as success — so Case 3, which archives the real posting, never runs. The result is the wrong JD recorded as the outcome artifact *and* the right one never captured. A caller cannot distinguish that from a genuine hit.

**The change.** When `companySlug` is supplied and no candidate carries it, return `null`. Every candidate belonging to some other company is exactly the case the slug exists to catch. Omitting `companySlug` is unchanged — most recent candidate, unconditionally. Failing closed degrades safely: the caller re-archives from the live URL, which also writes a correctly keyed capture. Failing open corrupts an archive.

**Assertion I replaced, called out deliberately:**

```js
- check('unmatched companySlug falls back to most recent', hintMiss?.filename === '070-iota-new.pdf');
+ check('unmatched companySlug reports no capture rather than guessing', hintMiss === null);
```

That test specified the defective behavior, so the fix could not land without changing it. Added a test reproducing the cross-company collision directly.

Suite is green: 3315 passed, 0 failed (one pre-existing environmental warning, `cv-sync-check` without a user-data layer). Padded/unpadded matching, the trailing-hyphen guard, recency tie-break, the candidates array, and the missing-directory path all still pass unchanged.

santifer:
Thanks for this, @rubicon.

archive-posting.mjs names the capture ${today()}_${company slug}_${role slug} and outcome.mjs looks for it by rebuilding that same string, whose three ingredients are all volatile; of its two resolution paths, the local:jds/ one in the Notes cell is written by nothing in the tree, and the re-archive one requires the listing to still be live, which is exactly what it isn't when the archive is needed. The numbers from the install where this turned up: 478 captures in jds/, 580 tracker rows, ZERO reachable, 0 occurrences of local:jds/ in pipeline.md and 0 files with an underscore, meaning the archiver's naming scheme has not produced a single surviving file. The proposed key is already the project's stable one (the report number) and 79 of those 478 captures had been renamed by hand to it, with and without leading zeros: --report=N plus a jd-capture.mjs that resolves by numeric prefix (64 finds 064-, never 640-) makes them resolvable with no migration.

rubicon:
Both CodeRabbit findings addressed in `75eb412`. I reproduced each against real captures first rather than taking them on faith — both were real, and the second one is a defect in the fix I had pushed earlier in this PR.

**1. `companySlug` matched as a substring of the whole filename (Major).**

Confirmed: role words answered for companies.

| slug | resolved to |
|---|---|
| `vp` | `04-mission-preborn-vp-marketing.txt` |
| `marketing` | `004-…_weave_chief-marketing-officer.pdf` |
| `chief` | the Weave capture |
| `''` | fell through to most-recent |

Same shape as the bug this PR opened against — matching on something that is not the thing being identified, one field further in.

Both grammars written here put the company first, after the report prefix and an optional capture date:

```
{NNN}-{YYYY-MM-DD}_{company}_{role}.pdf   archive-posting.mjs
{NNN}-{company}-{role}.{ext}              hand-named and plugin captures
```

So the slug is now anchored to that position and must end on a field boundary (`-`, `_`, `.`, or end of string). An empty slug is treated as supplied-but-unusable and matches nothing, rather than being falsy and falling through.

I did not go as far as parsing an exact company field for every format. The hyphen grammar is genuinely ambiguous — in `mission-preborn-vp-marketing` there is no way to know the company is `mission-preborn` and not `mission` — so an exact-field comparison would have to guess a boundary that the filename does not encode. Anchoring plus a boundary check rejects every role-word case above while still resolving both real grammars. If you would rather have strict parsing for the underscore grammar and no match at all for hand-named captures, say so and I will tighten it.

**2. A valueless `--report` was dropped (Major).**

Worse than described in the review. `--report` as the final argument did not merely get ignored — the posting was archived with no report prefix and the process exited 0:

```
$ node archive-posting.mjs --dry-run https://example.com/job --report
   Output:  jds/2026-08-11_unknown_job.pdf     # no NNN- prefix
   exit 0
```

That is silently unfindable, which is the exact failure `--report` exists to prevent. Now consumed unconditionally; an absent value reaches the existing validator and exits 1. Both `--report 4` and `--report=4` still key the capture.

**Tests added** (all failed first): company matching in both grammars, role-word / role-abbreviation / mid-name-word slugs rejected, empty slug rejected, trailing `--report` exits non-zero, and both `--report` forms still produce the `004-` prefix.

Rebased onto your merge of `main` rather than force-pushing over it. Suite green on the rebased tree: **3473 passed, 0 failed**.

santifer:
Merged, @rubicon. A capture that can only be found by rebuilding today's date-string is a capture that vanishes exactly when the posting dies and it becomes the only record: keying by report number fixes the real failure. The writers table you added to AGENTS.md is the honest kind of doc: it names the coexisting formats instead of pretending one is canonical. 🚀

PR #2702, opened 2026-08-11
fix(detect-reposts): concurrent openings were reported as reposts

artemtrofymenko:
## What does this PR do?

`detect-reposts.mjs` flagged concurrent openings as reposts, so `company-history.mjs` reported `postingChurn: reposts-detected` for companies that had never reposted anything. Two independent defects produced it; this fixes both and adds regression fixtures for each.

## Related issue

None — bug fix, sent straight in per the template.

## Type of change

- [x] Bug fix
- [ ] New feature

---

### Defect 1 — a zero-day span is not evidence of reposting

`first_seen` in `scan-history.tsv` is the **scanner's** observation date, not the employer's posting date, and a company is swept in one pass. Two URLs sharing a `first_seen` were therefore listed **concurrently** — the opposite of re-listing. Those clusters span 0 days and carry no information about reposting at all.

In a real 2341-row scan-history that was **165 of 252 clusters**.

Fixed with a minimum span (`MIN_REPOST_SPAN_DAYS = 1`, `--min-span` to override). 1 rather than something larger because scans run on an irregular cadence, so a genuine repost caught on two consecutive scan days spans exactly 1 and has to survive — the load-bearing part is only that 0 is excluded.

### Defect 2 — `roleFuzzyMatch` merged sibling requisitions

Title grouping used `roleFuzzyMatch`, which merges any two titles clearing a 0.6 Jaccard ratio. Per-region and per-segment variants of one job land **exactly** on that boundary — `Commercial Solutions Engineer - Munich` vs `- Berlin` share 3 of 5 tokens — so a company fanning one opening across cities read as a company reposting.

That was **39 further clusters** in the same history. I inspected all 39: every one was a false positive (per-country variants, per-city variants, `German speaking` vs `Dutch speaking`, `Technical Account Manager 3` vs `Technical Account Manager`, `Operations Manager` vs `Senior Operations Manager`). None was a genuine repost.

Replaced with `titleIdentityKey`: the set of a title's accent-folded, punctuation-stripped words. It sits deliberately between the two alternatives — raw string equality is too tight (a role re-listed months later comes back repunctuated, `… - Sweden` → `…, Sweden`), and fuzzy matching is too loose. Word order and punctuation may drift; a differing city, country, language, segment or seniority word now splits the titles.

Two details worth flagging:

- **No length/stopword filtering.** `roleTokens` drops words of 3 characters or fewer, which would erase the only thing distinguishing `… Engineer - UK` from `… Engineer - US`.
- **Script-preserving word splitting** (`\p{L}\p{N}`, not `[a-z0-9]`), for the reason #2429 gave for company names: the ASCII-only form folds every CJK or Cyrillic title to the empty string and they would all then share one key.

### Side effect: the #2389 token index is gone

Deciding membership by an equality key rather than pairwise similarity removes the quadratic term the inverted token index existed to prune. Grouping is now a single O(N) bucketing pass, so the index has nothing left to prune and is deleted with it. The #2389 performance guard (4000 distinct titles at one company) still passes.

## Result

On that 2341-row history: **252 clusters → 68**, with no span-0 and no mixed-identity clusters remaining, and genuine reposts (34d, 22d, 16d, 2d spans) still detected. The three companies that surfaced the defect now report `none-detected`, and `metadata.sources.scanHistory` is `true` in the same run — so those are real negatives, not a source that failed to load.

## Notes for reviewers

- **`--window 0` is now degenerate**, since the window and the span floor no longer overlap. Asserted in the tests so it is a documented outcome rather than a surprise, alongside a floor-disabled variant that keeps the `window=0` path itself covered.
- **The legacy-equivalence reference is updated in the same commit**, as its own note has always required. It is no longer a frozen pre-#2389 copy — it is now a second, independently written implementation, still guarding grouping order, same-date input order, URL dedup and both date bounds.
- **Each regression fixture isolates which constraint rejects it**, so none of them can pass for the wrong reason: the same-sweep fixture is asserted to *survive* with the span floor disabled (proving the floor rejects it), while the per-city and per-country fixtures are asserted rejected *even with the floor disabled* (proving title identity rejects them).
- Fixture company names and URLs are synthetic; only the shape is reproduced.

## Checklist

On `test-all.mjs`: 3394 passed / 1 failed, and that single failure is identical on unmodified `main` on this machine — a Windows `node_modules` symlink issue in the `followup-cadence` temp-dir e2e, unrelated to this change. Verified by running the suite with the four changed files reverted to `origin/main`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved repost detection by distinguishing genuine reposts from concurrent listings across locations, languages, segments, countries, and seniority levels.
  * Added stricter title identity matching to reduce false positives from unrelated or reordered job postings.
  * Reposts must now appear across separate scan dates and meet a configurable minimum time span.
  * Added command-line configuration for the minimum repost span, including safer handling of invalid values.

* **Documentation**
  * Clarified repost-detection rules and posting-churn behavior.

---

## Rebased onto current `main`, and one commit that is not about reposts

#2920 landed while this was open. It moved the four reporting CLIs onto `lib/cli-flags.mjs`, which rejects any flag absent from a per-file `KNOWN_FLAGS` list. This branch adds `--min-span`. The two changes merge without a textual conflict — both append to lists in different blocks — and the result was a flag parsed, documented, threaded through and unit-tested, refused at the door:

```
$ node detect-reposts.mjs --min-span 3 --summary
Error: unrecognized flag(s): --min-span. Valid flags: --window, --summary, …
exit 1
```

`--min-span` now appears in `KNOWN_FLAGS` and `VALUE_FLAGS`, and the block carries a comment saying to add both, because the next flag will arrive the same way. This is the case @abankar1 quotes in #2982; the marker comment is the cheap half of it, and a lint or a test that walks `USAGE` against `KNOWN_FLAGS` would be the durable half if you want one.

The regression test asserts the **value reaching metadata**, not a zero exit status: this failure exits non-zero for a reason unrelated to the flag's own logic, so an exit-code check would pass on a build where `--min-span` was silently dropped and the default used instead — the same false all-clear #2919 was about.

Re-verified after the rebase on the real 3226-row `scan-history.tsv`: the three companies that prompted this PR still report `none-detected`, and 254 assertions pass. The new guard was seen red by removing `--min-span` from `KNOWN_FLAGS`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -117,7 +117,7 @@ AI-powered, CLI-agnostic job search automation: pipeline tracking, offer evaluat
 | `company-history.mjs` | Read-only per-company evidence card joining the tracker, follow-ups, scan history and the status-log (JSON or `--summary`) |
 | `followup-cadence.mjs` | Follow-up cadence calculator (JSON) |
 | `followup-seed.mjs` | Seeds `data/follow-ups.md` with a pinned first follow-up date when a row turns Applied (JSON) |
-| `detect-reposts.mjs` | Flags roles re-listed 2+ times in 90 days from `scan-history.tsv` (JSON or `--summary`) |
+| `detect-reposts.mjs` | Flags roles re-listed 2+ times in 90 days from `scan-history.tsv` — requires 2+ distinct URLs seen on 2+ distinct scan dates (`--min-span`) with the same title identity, so concurrent per-city/country/segment openings are not mistaken for reposts (JSON or `--summary`) |
 | `check-table-freshness.mjs` | Staleness validator for jurisdiction data tables — flags `expired` rows (past `next_effective` without re-verification, exit 1) and `review-due` rows (`as_of` older than 12 months, soft); discovers any `templates/*.yml` with `as_of` rows automatically (JSON or `--summary` table output) |
 | `process-quality.mjs` | Per-company recruiting-friction rate from `[process-friction]` tags in `data/active-interviews.md` Notes (JSON or `--summary`) |
 | `rejection-latency.mjs` | Post-interview response-latency signal — flags companies still in `Interview` state whose silence since the last `data/active-interviews.md` round exceeds a courtesy (30d default, configurable) threshold, with a ready-to-copy `data/blacklist.md` suggestion row; suggestion-only, never writes (JSON or `--summary` table output) |
--- a/company-history.mjs
+++ b/company-history.mjs
@@ -12,6 +12,11 @@
  *     response — it is an answer, not silence.
  *   - postingChurn: does this company repost the same role repeatedly
  *     (evergreen requisition / re-opened search), per detect-reposts.mjs?
+ *     "The same role, later" is the whole claim: a company running several
+ *     openings side by side — one per city, country, language or segment — is
+ *     not reposting, and detect-reposts.mjs enforces that with a minimum span
+ *     and a title-identity rule. Before it did, this axis reported
+ *     `reposts-detected` for companies that had never reposted anything.
  *
  * This script deliberately reports FACTS, not verdicts. It never uses the
  * words "ghost"/"ghosted" or "risk" — high-volume inboxes, evergreen
--- a/detect-reposts.mjs
+++ b/detect-reposts.mjs
@@ -97,12 +97,21 @@ const selfTestMode = args.includes('--self-test');
 // Tested against Number()/Number.isInteger() as well: Number('') is 0, so an
 // empty value (`--min-span=`) would read as a deliberate zero and disable the
 // floor. A total regex has no such hole.
+// The regex is not sufficient on its own: it happily accepts a 400-digit run of
+// nines, which Number() turns into Infinity. That reproduces this file's
+// original defect in a new place — an infinite floor rejects every cluster, an
+// infinite window is unbounded, and JSON.stringify writes Infinity as `null`,
+// so metadata reports a value that is not the one in effect. A merely UNSAFE
+// integer is the quiet version of the same thing: 9007199254740993 silently
+// becomes ...992. isSafeInteger rejects both.
 const intFlag = (flag, fallback) => {
   const raw = flagValue(args, flag);
   if (raw === undefined) return fallback;
   const text = String(raw).trim();
   if (!/^\d+$/.test(text)) return fallback;
-  return Number(text);
+  const parsed = Number(text);
+  if (!Number.isSafeInteger(parsed)) return fallback;
+  return parsed;
 };
 const windowDays = intFlag('--window', DEFAULT_WINDOW_DAYS);
 const minSpanDays = intFlag('--min-span', MIN_REPOST_SPAN_DAYS);
--- a/detect-reposts.mjs
+++ b/detect-reposts.mjs
@@ -84,13 +84,25 @@ const summaryMode = args.includes('--summary');
 const selfTestMode = args.includes('--self-test');
 // Both numeric flags read through flagValue, so `--min-span=7` behaves exactly
 // like `--min-span 7` — the defect lib/cli-flags.mjs exists to keep the next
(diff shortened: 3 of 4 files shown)


artemtrofymenko:
Good catch on the flag validation — confirmed and fixed in 1a35137.

Reproduced before the fix:

| input | old `minSpanDays` / `windowDays` |
|---|---|
| `--min-span -5` | `-5` — **span floor silently disabled**, which is the guard this PR adds |
| `--min-span 7abc` | `7` |
| `--window -1` | `-1` — rejects every cluster |

One adjustment to the proposed diff: `Number(String(raw).trim())` would regress `--min-span=` (empty value). That currently falls back correctly, because `parseInt('')` is `NaN` — but `Number('')` is `0`, and `Number.isInteger(0) && 0 >= 0` passes, so an empty value would start reading as a deliberate zero and disable the floor. Same silent-disable failure the finding is about, just reached from the other side.

Used a total regex instead, which has no such hole:

```js
const text = String(raw).trim();
if (!/^\d+$/.test(text)) return fallback;
return Number(text);
```

`0` remains a legitimate explicit value on both flags, so the floor can still be switched off on purpose. Added seven CLI tests covering negative, partially numeric, non-integer, empty, and explicit-zero inputs on both flags.

artemtrofymenko:
Confirmed and fixed in eeb1b8e — this one was worth catching, because it reproduced the PR's own defect in a new place.

Measured before the fix:

| input | reported in `metadata` | actually in effect |
|---|---|---|
| `--min-span` (400 nines) | `null` | `Infinity` — rejected every cluster |
| `--window` (400 nines) | `null` | unbounded window |
| `--min-span 9007199254740993` | `9007199254740992` | a value nobody typed |

The `null` is the part that matters: `JSON.stringify` writes `Infinity` as `null`, so `metadata` was reporting a rule that was not the one producing the clusters. Same shape as the bug this PR fixes — a value that is written but not in effect.

Now requires `Number.isSafeInteger` after conversion, with the regex still ahead of it (it rejects negatives, decimals and exponent forms before `Number()` ever runs, so `1e999` never reaches the overflow path). Explicit `0` is unaffected.

Added cases for both flags plus two assertions that `metadata.minSpanDays` and `metadata.windowDays` are always finite — `null` there is the signal that an `Infinity` got through. `detect-reposts.test.mjs` is at 253 passing.

artemtrofymenko:
Follow-up filed as #2703 for the remaining class this PR deliberately leaves alone: `tracked_companies` entries that are multi-employer aggregators rather than single employers. On the same corpus that drove this PR, they account for 24 of the 68 clusters that survive it.

Kept out of here because it is a different root cause and needs a config decision rather than a bug fix — and because the config-free heuristic I expected to work does not (measured; numbers in the issue).

santifer:
Thanks for this, @artemtrofymenko.

Two defects, both measured on a real 2,341-row scan-history: first_seen is the SCANNER's observation date rather than the employer's, and a company gets swept in a single pass, so two URLs sharing first_seen are concurrent listings and not a repost (165 of 252 clusters); and roleFuzzyMatch merged sibling requisitions because 'Commercial Solutions Engineer - Munich' and '- Berlin' share 3 of 5 tokens and land right on the 0.6 Jaccard, with all 39 affected clusters inspected one by one and not one a genuine repost. Both get replaced by a 1-day minimum span and titleIdentityKey (words folded without accents or punctuation, no short-word filter so UK vs US survives, \p{L}\p{N} split so CJK and Cyrillic titles don't empty out): 252 clusters down to 68, with the genuine 34d/22d/16d/2d reposts intact.

santifer:
Ten days was too long for this to wait, @artemtrofymenko, and the fix deserved better: the first_seen-is-the-scanner's-date insight (a whole company swept in one pass makes a zero-day span meaningless as repost evidence) is the kind of data-provenance reasoning most repost detectors never do. Both defects pinned with their own regression fixtures, as usual with your work. Merged! 🚀

PR #2705, opened 2026-08-11
fix(status): a Turkish tracker row normalizes three different ways — TUI, core and web all disagree

### LINKED ISSUE #2704: Turkish tracker statuses normalize three different ways — the TUI, the core, and the web each disagree
### What

`templates/states.yml` calls itself *"Source of truth for career-ops (writer) and dashboard (reader). Both systems MUST use these exact states."* It isn't. The Go dashboard's `NormalizeStatus` grew its own, larger alias table — **59 status literals against states.yml's 23 aliases** — and the extra ones are every Turkish spelling.

So one tracker row normalizes three different ways:

| Tracker status | Go TUI | core `.mjs` | web |
|---|---|---|---|
| `Entrevista` (es) | `interview` | `interview` | `Interview` |
| `Mülakat` (tr) | **`interview`** | **`mülakat`** | **`null`** |
| `Teklif` (tr) | **`offer`** | **`teklif`** | **`null`** |
| `Başvuruldu` (tr) | **`applied`** | **`başvuruldu`** | **`null`** |

We ship `modes/tr/`. For a Turkish user this means:

- the TUI renders the row correctly, so nothing looks wrong;
- `followup-cadence`'s `normalizeStatus` returns the raw string, which matches no `ACTIONABLE_STATUSES` / `ADVANCED_STATUSES` set — the row silently drops out of the follow-up queue and the funnel;
- `normalize-statuses.mjs`, whose entire job is normalizing statuses, reports it as **unknown**;
- `POST /api/status` rejects it on writeback.

Nobody is told any of this.

### Root cause

Three hand-maintained vocabularies. `tracker-utils.mjs` already exposes `loadCanonicalStates()` / `resolveCanonicalState()` — its docstring says *"a new state or alias lands in one file and every consumer follows"* — but `followup-cadence.mjs` and `normalize-statuses.mjs` each kept their own map instead of calling it. Only `set-status.mjs` and `funnel-velocity.mjs` use the shared loader.

### Fix

1. Move the spellings Go already knows into `states.yml` — the SSOT is the side that was behind, so **Go needs no change**.
2. Point `followup-cadence` and `normalize-statuses` at the shared loader instead of their private maps.
3. Guard the direction that actually drifts: every status literal Go matches on must resolve through `states.yml`.

### Known limitation (pre-existing, not addressed)

Turkish dotted capital `İ` lowercases in JS to `i` + U+0307, so `İptal Edildi` still resolves to unknown while `iptal edildi` works. Go has the same behaviour — that's why its table carries both `iptal` and `ıptal` spellings. Fixing locale-aware casing is a separate concern and I'd rather flag it than bury a half-fix in here.

PR incoming.

abankar1:
Closes #2704.

`states.yml` calls itself *"Source of truth for career-ops (writer) and dashboard (reader). Both systems MUST use these exact states."* It isn't — Go's `NormalizeStatus` carries **59 status literals** against states.yml's **23 aliases**, and the extra ones are the Turkish spellings.

One row, three answers:

| Tracker status | Go TUI | core `.mjs` | web |
|---|---|---|---|
| `Entrevista` (es) | `interview` | `interview` | `Interview` |
| `Mülakat` (tr) | **`interview`** | **`mülakat`** | **`null`** |
| `Teklif` (tr) | **`offer`** | **`teklif`** | **`null`** |
| `Başvuruldu` (tr) | **`applied`** | **`başvuruldu`** | **`null`** |

We ship `modes/tr/`, so for a Turkish user the TUI looks correct while `followup-cadence` returns the raw string — matching no `ACTIONABLE_STATUSES`/`ADVANCED_STATUSES` set, so the row silently leaves the follow-up queue and the funnel — `normalize-statuses` calls it unknown, and `POST /api/status` rejects it.

## Fix

**1. The SSOT was the side that was behind**, so the spellings move *into* `states.yml` and **Go needs no change at all**. That's the whole reason this is a small diff.

**2. Both core normalizers now derive.** `followup-cadence` and `normalize-statuses` each kept a private alias map. `tracker-utils` already exposes `loadCanonicalStates`/`resolveCanonicalState`, and its docstring already promises *"a new state or alias lands in one file and every consumer follows"* — it simply had no consumer in either file (only `set-status` and `funnel-velocity` used it).

After:

```
status            core          web
Mülakat           interview     Interview
Teklif            offer         Offer
Başvuruldu        applied       Applied
Kabul Edildi      hired         Hired
Değerlendirildi   evaluated     Evaluated
```

**3. A guard on the direction that drifts.** Every status literal Go matches on must resolve through `states.yml`. Go may keep *more* matching logic (it uses substring `Contains`), but it must not know a spelling the source of truth has never heard of. That guard immediately earned itself — it caught four `discarded` spellings (`iptal edildi` and variants) my own manual pass had missed.

## Two things I changed that weren't in the plan

- **`states-alias-coverage`'s `LIST_RULE_RE` arm** extracted the hardcoded lists this PR deletes, so it correctly reported it was checking nothing. Rather than delete it, I inverted it: it now asserts the derivation is wired **and** that a hardcoded list hasn't reappeared. Verified it fails in both directions.
- **The e2e cadence fixture** copies specific files into a temp dir; `followup-cadence` now needs `tracker-utils.mjs` and `templates/states.yml` there too, same reason `tracker-aliases.json` is already copied for `tracker-parse.mjs`.

## Known limitation, flagged not buried

Turkish dotted capital `İ` lowercases in JS to `i` + U+0307, so `İptal Edildi` still resolves to unknown while `iptal edildi` works. **Go has the same behaviour** — that's why its table carries both `iptal` and `ıptal`. Locale-aware casing is a separate problem and I'd rather name it than ship a half-fix.

`node test-all.mjs --quick` → **3400 passed, 0 failed**.

3 commits (states.yml → derivation → guards).

## Summary by CodeRabbit

* **Improvements**
  * Status recognition now uses shared canonical definitions across tracking and normalization tools.
  * Added multilingual aliases, including Turkish variants, alternate spellings, and phrase forms.
  * Improved handling of capitalization, accents, bold formatting, and Unicode variations.
  * Status labels, identifiers, descriptions, and dashboard groupings remain unchanged.
* **Reliability**
  * Added validation for alias resolution, including Turkish uppercase variants.
  * Preserved safe fallback behavior when canonical status data cannot be loaded.

### DIFF
--- a/followup-cadence.mjs
+++ b/followup-cadence.mjs
@@ -15,7 +15,7 @@ import { readFileSync, existsSync } from 'fs';
 import { join, dirname, relative, sep } from 'path';
 import { fileURLToPath, pathToFileURL } from 'url';
 import * as yaml from 'js-yaml';
-import { loadCanonicalStates } from './tracker-utils.mjs';
+import { loadCanonicalStates, foldStatusInput } from './tracker-utils.mjs';
 import { resolveColumns, parseTrackerRow } from './tracker-parse.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
@@ -106,9 +106,9 @@ function statusAliasMap() {
   try {
     for (const st of loadCanonicalStates(join(CAREER_OPS, 'templates', 'states.yml'))) {
       const id = st.id.toLowerCase();
-      map.set(id, id);
-      if (st.label) map.set(st.label.toLowerCase(), id);
-      for (const a of st.aliases) map.set(String(a).toLowerCase(), id);
+      map.set(foldStatusInput(id), id);
+      if (st.label) map.set(foldStatusInput(st.label), id);
+      for (const a of st.aliases) map.set(foldStatusInput(a), id);
     }
   } catch {
     // A missing/malformed states.yml is a broken install. Degrade to
@@ -122,8 +122,10 @@ function statusAliasMap() {
 const ACTIONABLE_STATUSES = ['applied', 'responded', 'interview'];
 
 export function normalizeStatus(raw) {
-  const clean = raw.replace(/\*\*/g, '').trim().toLowerCase()
-    .replace(/\s+\d{4}-\d{2}-\d{2}.*$/, '').trim();
+  // foldStatusInput, not a bare toLowerCase: JS lowercases the Turkish dotted
+  // capital `İ` to `i` + U+0307 and the mark survives, so every all-caps
+  // Turkish row missed every alias (#2704 review).
+  const clean = foldStatusInput(String(raw).replace(/\s+\d{4}-\d{2}-\d{2}.*$/, ''));
   return statusAliasMap().get(clean) || clean;
 }
 
--- a/followup-cadence.mjs
+++ b/followup-cadence.mjs
@@ -15,6 +15,7 @@ import { readFileSync, existsSync } from 'fs';
 import { join, dirname, relative, sep } from 'path';
 import { fileURLToPath, pathToFileURL } from 'url';
 import * as yaml from 'js-yaml';
+import { loadCanonicalStates } from './tracker-utils.mjs';
 import { resolveColumns, parseTrackerRow } from './tracker-parse.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
@@ -82,31 +83,48 @@ export function resolveCadenceConfig({ profilePath = PROFILE_FILE, appliedDays =
 
 const CADENCE = resolveCadenceConfig();
 
-// --- Status normalization (mirrors verify-pipeline.mjs) ---
-const ALIASES = {
-  'evaluada': 'evaluated', 'condicional': 'evaluated', 'hold': 'evaluated',
-  'evaluar': 'evaluated', 'verificar': 'evaluated',
-  'aplicado': 'applied', 'enviada': 'applied', 'aplicada': 'applied',
-  'applied': 'applied', 'sent': 'applied',
-  'respondido': 'responded',
-  'entrevista': 'interview',
-  'oferta': 'offer',
-  // Hired aliases from templates/states.yml — without these, an "Accepted" or
-  // "Contratado" row normalizes to itself, so stats/funnel/company-history
-  // consumers looking for 'hired' silently drop the best outcome in the tracker.
-  'contratado': 'hired', 'contratada': 'hired', 'accepted': 'hired', 'accept': 'hired',
-  'rechazado': 'rejected', 'rechazada': 'rejected',
-  'descartado': 'discarded', 'descartada': 'discarded',
-  'cerrada': 'discarded', 'cancelada': 'discarded',
-  'no aplicar': 'skip', 'no_aplicar': 'skip', 'monitor': 'skip', 'geo blocker': 'skip',
-};
+// --- Status normalization ---
+//
+// DERIVED from templates/states.yml, not a hand-copy of it. The map that used
+// to live here was already missing every Turkish spelling the Go dashboard
+// recognises, so the same tracker row normalized three different ways: the TUI
+// read `Mülakat` as `interview`, this file left it as `mülakat` (matching no
+// ACTIONABLE/ADVANCED set, so the row silently vanished from the funnel), and
+// the web rejected it outright on writeback. tracker-utils already exposes the
+// loader for exactly this — its docstring says "a new state or alias lands in
+// one file and every consumer follows" — it just had no consumer here (#2704).
+//
+// Cached per process: these are short-lived CLI runs, so a single read is
+// correct. A long-running consumer must NOT copy this pattern — see #2590,
+// where caching states.yml for a server's lifetime pinned a stale roster.
+let aliasMapCache = null;
+
+/** alias/id/label (lowercased) → canonical lowercase id, from states.yml. */
+function statusAliasMap() {
+  if (aliasMapCache) return aliasMapCache;
+  const map = new Map();
+  try {
+    for (const st of loadCanonicalStates(join(CAREER_OPS, 'templates', 'states.yml'))) {
+      const id = st.id.toLowerCase();
+      map.set(id, id);
+      if (st.label) map.set(st.label.toLowerCase(), id);
+      for (const a of st.aliases) map.set(String(a).toLowerCase(), id);
+    }
+  } catch {
+    // A missing/malformed states.yml is a broken install. Degrade to
+    // identity-normalization rather than resurrecting a hardcoded table: a
+    // fallback copy is the same copy in disguise and drifts the same way.
+    return (aliasMapCache = new Map());
+  }
+  return (aliasMapCache = map);
+}
 
 const ACTIONABLE_STATUSES = ['applied', 'responded', 'interview'];
 
 export function normalizeStatus(raw) {
   const clean = raw.replace(/\*\*/g, '').trim().toLowerCase()
     .replace(/\s+\d{4}-\d{2}-\d{2}.*$/, '').trim();
-  return ALIASES[clean] || clean;
+  return statusAliasMap().get(clean) || clean;
 }
 
 // --- Date helpers ---
--- a/normalize-statuses.mjs
+++ b/normalize-statuses.mjs
@@ -16,7 +16,7 @@ import { join, dirname } from 'path';
 import { fileURLToPath, pathToFileURL } from 'url';
 import {
   openTrackerTransaction, rebuildRow, resolveTrackerPath,
-} from './tracker-utils.mjs';
+  loadCanonicalStates, resolveCanonicalState,} from './tracker-utils.mjs';
 import { resolveColumns, parseTrackerRow } from './tracker-parse.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
@@ -27,6 +27,18 @@ const DRY_RUN = process.argv.includes('--dry-run');
 mkdirSync(join(CAREER_OPS, 'data'), { recursive: true });
 
 // Canonical status mapping
+let statesCache = null;
+/** Canonical states from templates/states.yml, read once per CLI run. */
+function canonicalStates() {
+  if (statesCache) return statesCache;
+  try {
+    statesCache = loadCanonicalStates(join(CAREER_OPS, 'templates', 'states.yml'));
+  } catch {
+    statesCache = []; // broken install: fall through to "unknown", never a stale copy
+  }
+  return statesCache;
+}
+
 function normalizeStatus(raw) {
   // Strip markdown bold
   let s = raw.replace(/\*\*/g, '').trim();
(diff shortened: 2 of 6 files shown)


santifer:
This is the right diagnosis, @abankar1, and the part I want to underline is that you fixed the SSOT rather than the symptom. Go carrying 59 literals against 23 aliases meant `states.yml` was calling itself the source of truth while being the side that was behind. Moving the spellings *into* it, so Go needs no change at all, is why this is a small diff instead of a coordination problem.

The failure you describe is also the expensive kind. A Turkish row matches no `ACTIONABLE_STATUSES` and no `ADVANCED_STATUSES`, so it leaves the follow-up queue and the funnel without an error anywhere. The user sees a shorter list, not a bug. We shipped exactly this class last week in company matching, and it is the same lesson twice: for a non-English user, our failures are silent.

**One gap before I merge, and it is two lines.** I ran this past the web side, and the matching does `.toLowerCase()`, which in Turkish is where `İ` bites:

```console
$ node -e "for (const s of ['TEKLİF','DEĞERLENDİRİLDİ','KABUL EDİLDİ']) console.log(JSON.stringify(s), '->', JSON.stringify(s.toLowerCase()))"
"TEKLİF" -> "tekli̇f"
"DEĞERLENDİRİLDİ" -> "değerlendi̇ri̇ldi̇"
"KABUL EDİLDİ" -> "kabul edi̇ldi̇"
```

JavaScript lowercases `İ` to `i` plus a combining dot above (U+0307), and the mark survives, so the lowercased string never equals the alias you wrote. `Mülakat` and `MÜLAKAT` are fine because they contain no dotted capital I; `TEKLİF` is not.

Your PR already anticipates the diacritic-free variants (`degerlendirildi`, `mulakat`), so this is the same instinct applied one level deeper: add the three strings JavaScript actually produces (`tekli̇f`, `değerlendi̇ri̇ldi̇`, `kabul edi̇ldi̇`) as aliases. Copy them from that command's output rather than typing them, since the combining mark is invisible in an editor.

To be clear this is an extension, not an objection: today **every** Turkish row fails, and after your PR only the all-caps ones would. Strictly better either way. But uppercase status words are ordinary in Turkish, so it is worth closing now rather than filing.

Credit where it is due: the uppercase case came from the web side, which also checked its own parser against this and confirmed it reads `aliases` from `states.yml` live, so it picks up your spellings with no change on its end.

**One thing this changes beyond your PR.** We looked at this exact `İ` behaviour last week in the company-matching key and pinned it as intentional, on the grounds that it was a one-off. It is now biting a second, unrelated surface, which is a different fact than the one that decision was made on. I am reopening that question separately; nothing for you to do, but your PR is the evidence that moved it.

Push the three aliases and this goes in.

abankar1:
Your diagnosis is right and it's now fixed — but **I didn't do it the way you asked, and I want to put the reason in front of you rather than bury it in a diff.**

You asked for three aliases. I measured the full set first, and it's **32**, not 3:

```
evaluated   condicional      typed as CONDİCİONAL     -> we see "condi̇ci̇onal"
applied     aplicado         typed as APLİCADO        -> we see "apli̇cado"
interview   entrevista       typed as ENTREVİSTA      -> we see "entrevi̇sta"
offer       teklif           typed as TEKLİF          -> we see "tekli̇f"
skip        skip             typed as SKİP            -> we see "ski̇p"
hired       hired            typed as HİRED           -> we see "hi̇red"
…32 total
```

The rule isn't "Turkish words" — it's **any alias containing an `i`**, which is why `skip` and `hired` are on that list. A Turkish-locale keyboard uppercases `i` to `İ` in *any* word, English ones included. Listing them would mean 32 invisible-combining-mark strings in `states.yml`, and a 33rd needed the next time anyone adds an alias with an `i` in it.

So I folded instead — drop U+0307 after lowercasing:

```js
export function foldStatusInput(input) {
  return String(input ?? '').replace(/\*\*/g, '').trim().toLowerCase()
    .normalize('NFD').replace(/̇/g, '').normalize('NFC');
}
```

**31 of the 32 repaired, 0 regressions.** The 32nd is a different asymmetry — in `İŞE ALINDI` the *dotless* `ı` uppercases to `I` and lowercases back to dotted `i` — so that one does get an alias, as you suggested. One alias instead of 32.

Verified across both normalizers, every value, three casings:

```
checked 216 spellings (as-written, Turkish-uppercase, plain-uppercase)
failures: 0

TEKLİF -> offer     DEĞERLENDİRİLDİ -> evaluated
KABUL EDİLDİ -> hired    İŞE ALINDI -> hired    İPTAL EDİLDİ -> discarded
```

`foldStatusInput` lives in `tracker-utils` beside the loader and is used by `resolveCanonicalState`, so `set-status`, `funnel-velocity` and `normalize-statuses` inherit it — and by `followup-cadence`'s lookup.

**The safety question your suggestion avoided and mine doesn't:** folding could in principle collapse two states if any canonical value legitimately contained U+0307. None does, and that's now asserted rather than assumed:

> ✅ no canonical state value carries U+0307, so the fold cannot merge two states

If you'd still rather have the 32 literals — the fold is a behaviour change to shared matching and the alias list isn't — say so and I'll swap it; it's your call on blast radius, and the aliases would be strictly more conservative.

Also noted your point about reopening the company-matching `İ` decision. This is the same root cause, and `foldStatusInput` is deliberately status-only for now — I didn't touch `normalizeTextKey`, since company keys have different collision stakes and that's the question you said you're reopening.

`node test-all.mjs --quick` → **3404 passed, 0 failed**.

santifer:
Closing the loop here since you found the second surface, @abankar1: the dotted-capital fix landed in main today (`462d2765`).

`İstanbul Tekstil` and `Istanbul Tekstil` now key the same, while `Škoda`/`Skoda`, `Nestlé`/`Nestle` and `İŞ BANKASI`/`Is Bankasi` all still key apart, because those marks are letters somebody typed rather than an artifact of lowercasing. Core and the web mirror changed in the same commit.

Your report is what moved it. When you flagged it the decision was to leave it alone as intentional, and the argument for that was "it is one isolated case". You found the second surface, a third turned up later, and one of them was the duplicate checker itself returning a false green. That is what changed the answer.

abankar1:
Rebased onto `462d2765` — conflict was just the `js-yaml` namespace-import change, and the suite is green on today's main (**3656 passed, 0 failed**), including your §55.7 byte-for-byte web-mirror guard.

Two notes now that your fix has landed:

**They don't overlap, and I checked rather than assumed.** Your fold is in `normalizeTextKey`, which strips spaces and punctuation because it's an *identity* key. Status labels are multi-word, so that scope is wrong for them:

```
                 normalizeTextKey    foldStatusInput
Kabul Edildi     "kabuledildi"       "kabul edildi"
uygun değil      "uygundeğil"        "uygun değil"
no aplicar       "noaplicar"         "no aplicar"
```

Routing statuses through the company key would collapse `no aplicar` to `noaplicar` and match no alias. So `foldStatusInput` stays, and it's deliberately status-only — I still haven't touched `normalizeTextKey`.

**But there are now two mark-stripping implementations in the tree**, yours inline in `tracker-parse.mjs` and mine in `tracker-utils.mjs`. That's the copy-vs-derive class I've been filing PRs about, so I'm not going to pretend I didn't introduce half of it. I *didn't* factor them into one primitive here on purpose: your `normalizeTextKey` body is pinned byte-for-byte against the web mirror by §55.7, and refactoring freshly-landed, guarded code inside an unrelated PR is how that guard gets broken by accident. Happy to do it as its own small PR if you want the single primitive — just say.

On the decision reversal: noted, and thank you for saying that the report is what moved it. The part I'd underline for next time is that the original "it's one isolated case" reasoning wasn't wrong on the evidence available — it just wasn't checkable. The thing that changed it was a second and third surface turning up, one of them a duplicate checker returning a false green, which is the failure mode that argues for folding rather than for cataloguing exceptions.

abankar1:
Found a defect in my own branch off the back of `5df43e7`, and pushed the fix.

`foldStatusInput` used `NFD → strip U+0307 → NFC` — **the exact pattern that commit had to undo on the company key.** NFD decomposes precomposed letters too, so the strip reached the dots of `ż`, `ė` and `ġ`. Measured on my branch before fixing:

```
        old fold (NFD)   fixed (NFKC)
Żubr    "zubr"           "żubr"
Ėmė     "eme"            "ėmė"
Ġenerali "generali"      "ġenerali"

Żubr/Zubr:       COLLAPSED  ->  distinct
Ėmė/Eme:         COLLAPSED  ->  distinct
Ġenerali/Generali: COLLAPSED ->  distinct
```

Every test on the branch was green while that was true, which is the same way it got past review on the company key: the Turkish assertions prove the fold *works*, and say nothing about what else it reaches.

Same structural remedy as yours — NFKC, lowercase, strip, no decomposition — and I've copied the reasoning into the comment rather than just the code, since "don't use NFD here" is the whole safety property and it isn't obvious from the line.

Pinned with **pair semantics**, mirroring your 55.7b: Turkish pairs must fold, `ż`/`ė`/`ġ`/`Š`/`é` pairs must not. Confirmed it fails against the NFD version:

```
❌ foldStatusInput pair semantics wrong: Żubr/Zubr expected differ; Ėmė/Eme expected differ; Ġenerali/Generali expected differ
```

`node test-all.mjs --quick` → **3657 passed, 0 failed**.

The generalisable bit, since this is now the second time the same trap has been walked into in three days: a fold like this needs a test that asserts what it must **not** touch. Testing only that the intended case works passes for both the correct implementation and the over-broad one.

santifer:
Merged, @abankar1, and I want to be precise about what I checked, because I broke this exact thing myself yesterday.

I re-ran your `foldStatusInput` rather than reading it: TEKLİF→Offer, KABUL EDİLDİ→Hired, DEĞERLENDİRİLDİ→Evaluated, REDDEDİLDİ→Rejected, İŞE ALINDI→Hired, MÜLAKAT→Interview, and the controls hold: Żubr≠Zubr, Ėmė≠Eme, Ġenerali≠Generali, Škoda≠Skoda, Nestlé≠Nestle. NFKC then strip, no NFD. That is the right shape, and the comment above it explains *why* the two look equivalent and are not. Yesterday I shipped the NFD version, it collapsed Polish, Lithuanian and Maltese, and the reason it happened is that the measurement justifying the safe form lived in a conversation instead of next to the code. Yours doesn't. That is the better half of this PR.

One scope note so the record is straight, and it is not a criticism of your work: the title says TUI, core and web, and this lands the first two. The web app is owned by a separate maintainer track, so its copy is not yours to change: `web/src/lib/core/states.ts` still lowercases without the strip, so the API rejects the same uppercase Turkish inputs the core now accepts (MÜLAKAT and BAŞVURULDU pass there because they carry no İ, which makes it incoherent rather than merely unfixed), and its FALLBACK table is now 31 aliases behind the file it claims to mirror. I have handed both to that track with the repro. The squash subject reflects what actually shipped here.

Your alias additions to states.yml are picked up by the web automatically, since it reads the file live: it is only the case-folding arm that stops at the boundary.

PR #2707, opened 2026-08-11
fix(getro): pace sequential pages and retry transient failures

### LINKED ISSUE #2706: getro: unpaced pagination triggers Datadog WAF IP ban (silent board loss)
### What happens

```
```

### Secondary issue

### Proposed fix

IaroslavMazur:
Fixes #2706.

`providers/getro.mjs` paginates with back-to-back POSTs to one endpoint. On a 40-page board that burst reads as scraping to Getro's Datadog WAF, which IP-bans the caller (`403 "You've been blocked"`, ~2.5h observed on 2026-08-08). Since 403 is correctly non-retryable, the result is silent data loss — the scan just returns fewer companies.

### Changes

**1. Pace sequential pages of the same board (250 ms).**
Skipped before page 0, so a single-page board is unaffected. `scan.mjs` runs companies 10-wide but those are distinct tenants on distinct hosts — the burst that gets noticed is N back-to-back requests at *one* endpoint, so this costs effectively nothing on a parallel scan. Uses `ctx.sleep` when present, matching the convention in `_http.mjs`, so unit tests never wall-clock wait.

I considered a per-host limiter in `_http.mjs` and rejected it: ATS hosts are multi-tenant, so keying a token bucket by hostname would serialise the dozens of Ashby and Greenhouse tenants that run 10-wide today without any vendor complaining.

**2. Route the fetch through `fetchJsonWithRetry`.**
`getro.mjs` was the only paginating provider still calling `ctx.fetchJson` raw, so one transient 5xx or timeout discarded an entire board — the failure mode #2506 added the helper to prevent. Default policy (2 retries, 500 ms base, 8 s cap) applies; no behaviour change on the happy path.

### Verification

- All 13 Getro boards in my `portals.yml` complete without a 403; before the change, three consistently failed.
- A 160-job board (8 pages) costs ~1.75 s of added pacing.
- `node test-all.mjs` — 3128 passed. The 9 failures on my tree are pre-existing `.gitignore`-coverage assertions unrelated to `providers/`.

No config surface added, no new dependencies, defaults unchanged for every other provider.


IaroslavMazur:
Closing — I branched from a local HEAD that had diverged from upstream, so this PR carries 15 unrelated commits and personal files. Reopening a clean single-commit PR shortly. Apologies for the noise.

PR #2708, opened 2026-08-11
fix(getro): pace sequential pages and retry transient failures

### LINKED ISSUE #2706: getro: unpaced pagination triggers Datadog WAF IP ban (silent board loss)

IaroslavMazur:
Fixes #2706.

`providers/getro.mjs` paginates with back-to-back POSTs to one endpoint. On a 40-page board that burst reads as scraping to Getro's Datadog WAF, which IP-bans the caller (`403 "You've been blocked"`, ~2.5h observed on 2026-08-08). Since 403 is correctly non-retryable, the result is silent data loss — the scan just returns fewer companies.

### Changes

**1. Pace sequential pages of the same board (250 ms).**
Skipped before page 0, so a single-page board is unaffected. `scan.mjs` runs companies 10-wide but those are distinct tenants on distinct hosts — the burst that gets noticed is N back-to-back requests at *one* endpoint, so this costs effectively nothing on a parallel scan. Uses `ctx.sleep` when present, matching the convention in `_http.mjs`, so unit tests never wall-clock wait.

I considered a per-host limiter in `_http.mjs` and rejected it: ATS hosts are multi-tenant, so keying a token bucket by hostname would serialise the dozens of Ashby and Greenhouse tenants that run 10-wide today without any vendor complaining.

**2. Route the fetch through `fetchJsonWithRetry`.**
`getro.mjs` was the only paginating provider still calling `ctx.fetchJson` raw, so one transient 5xx or timeout discarded an entire board — the failure mode #2506 added the helper to prevent. Default policy (2 retries, 500 ms base, 8 s cap) applies; no behaviour change on the happy path.

### Verification

- All 13 Getro boards in my `portals.yml` complete without a 403; before the change, three consistently failed.
- A 160-job board (8 pages) costs ~1.75 s of added pacing.
- `node test-all.mjs` — 3128 passed. The 9 failures on my tree are pre-existing `.gitignore`-coverage assertions unrelated to `providers/`.

No config surface added, no new dependencies, defaults unchanged for every other provider.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when loading Getro job listings by automatically retrying failed page requests.
  * Added a brief delay between page loads to reduce request failures and improve data retrieval stability.

### DIFF
--- a/providers/getro.mjs
+++ b/providers/getro.mjs
@@ -1,6 +1,8 @@
 // @ts-check
 /** @typedef {import('./_types.js').Provider} Provider */
 
+import { fetchJsonWithRetry } from './_http.mjs';
+
 // Getro provider — VC "talent network" portfolio job boards (jobs at a fund's
 // portfolio companies). Powers b2venture, Earlybird, Point Nine, Speedinvest,
 // Cherry, HV Capital, Atomico, and many other VC boards.
@@ -52,6 +54,14 @@ const DEFAULT_MAX_PAGES = 40;      // safety cap: 40 x 20 = 800 newest jobs/boar
 // 10k sequential API calls against a third party.
 const HARD_MAX_PAGES = 200;        // 200 x 20 = 4000 newest jobs/board
 const DEFAULT_MAX_AGE_DAYS = 90;   // pagination bound only; global filter does the real cut
+// Pacing between sequential pages of the SAME board. scan.mjs runs companies
+// 10-wide, but those are different tenants on different hosts; the burst that
+// gets noticed is N back-to-back POSTs at one endpoint. Getro sits behind a
+// Datadog WAF that IP-blocks on that pattern (observed 2026-08-08: 40-page
+// boards → 403 "You've been blocked" for ~2.5h, which is silent data loss
+// because 403 is correctly non-retryable). A short sleep between pages costs
+// nothing on a parallel scan of distinct tenants and removes the signature.
+const PAGE_DELAY_MS = 250;
 
 function resolveCollection(entry) {
   const id = entry.getro_collection;
@@ -83,7 +93,15 @@ export default {
     const out = [];
     let total = Infinity;
     for (let page = 0; page < maxPages && page * HITS_PER_PAGE < total; page++) {
-      const json = await ctx.fetchJson(apiUrl, {
+      // Pace pages 1..n; page 0 is the first contact with this host and needs
+      // no delay. ctx.sleep keeps unit tests off the wall clock, matching the
+      // convention in _http.mjs.
+      if (page > 0) {
+        await (typeof ctx.sleep === 'function'
+          ? ctx.sleep(PAGE_DELAY_MS)
+          : new Promise(resolve => setTimeout(resolve, PAGE_DELAY_MS)));
+      }
+      const json = await fetchJsonWithRetry(ctx, apiUrl, {
         method: 'POST',
         // redirect:'error' — apiUrl is pinned to api.getro.com (https), so a 3xx
         // to a private/metadata IP must not be followed (matches every provider).


ilyakanevskiy:
@IaroslavMazur The WAF ban you hit (2026-08-08, 403 "You've been blocked", ~2.5h) is useful independent confirmation of the failure mode.

This overlaps with #2640, which is already open against `providers/getro.mjs` and includes the same two fixes (inter-page pacing honoring `ctx.sleep`, and routing pagination through `fetchJsonWithRetry`), plus auto-resolve of `collection_id`, salary/location extraction, the `redirect:'error'` fix for redirecting `careers_url`s, and `ctx.maxPages` support.

#2640 currently uses a 150ms inter-page delay rather than 250ms. I have run 8 real Getro boards against it with 150ms and seen zero 403s, so I would suggest keeping 150ms - but if you have a specific reason to prefer 250ms (e.g. you saw it recur at a lower delay on your 13 boards), let me know and I will bump it in #2640.

Given the overlap, would you mind closing this one in favor of #2640 once that lands?

IaroslavMazur:
Thanks @ilyakanevskiy — you're right, #2640 covers both fixes and does more besides. And no, I have no reason to prefer 250ms; keep 150ms. Closing in favour of #2640.

santifer:
Welcome to career-ops, @IaroslavMazur. This is a strong first PR and I want to be specific about why.

You brought field evidence, not a hypothesis: a dated observation of the Datadog WAF blocking a 40-page board for ~2.5h, and the reason it hurts (403 is correctly non-retryable, so the scan just quietly returns fewer companies). Silent partial data is the worst failure class we have here, because nothing downstream can tell "this fund has fewer openings" from "we got banned mid-sweep".

You also documented the design you rejected, which most PRs skip: a per-host limiter in `_http.mjs` would serialise the dozens of Ashby and Greenhouse tenants that run 10-wide today, because ATS hosts are multi-tenant. That is the correct call and it saved me from asking.

**Two things you could not have known, both on me:**

**1. There is a collision.** #2640 touches `providers/getro.mjs` for the same reason, independently: it also paces sequential pages (`INTER_PAGE_DELAY_MS = 150` to your `PAGE_DELAY_MS = 250`) and also routes the fetch through retry. Two people found the same problem in the same file in the same week and neither could see the other. That is my job to catch, not yours.

Merge order will be yours first, since your change is the narrow one and carries the field measurement behind the number; #2640 then rebases on top and keeps the wider work (auto-resolve, salary, remote-detect). Nothing here needs action from you yet.

**2. Your PR is the counter-example in #2709,** an issue I opened an hour ago about `_http.mjs` being hard to adopt. You reached for `fetchJsonWithRetry` and it fit, because getro speaks JSON. The providers that reimplement retry locally are the ones fetching text, which the shared helper has no equivalent for. Your PR is what the good path looks like.

**Housekeeping, and this one is my fault:** first-time contributors have CI held until a maintainer approves it, and I fumbled the approval command on your PR specifically, which left the queued runs in a state where they never started. That is why you see `label` and `welcome` green and no tests at all. I am going to close and immediately reopen this PR to re-trigger the workflows. You will get two notifications in a row; ignore both, nothing is being rejected, and I will approve the runs the moment they queue.

santifer:
Reopening this was my mistake, @IaroslavMazur: I acted on a stale queue that still listed this PR as awaiting CI approval, and I wrote that whole comment above without reading that you and @ilyakanevskiy had already sorted it out ten minutes earlier. Closing it again now, as you intended.

Please ignore the merge-order paragraph in my previous comment. Your call was better than mine and it was made with better evidence: @ilyakanevskiy ran 8 real Getro boards at 150ms with zero 403s, which is a stronger basis for the number than my "the narrow one goes first" instinct. You two resolved a collision, with measurements, faster than the maintainer noticed it existed. That is the repo working the way it should.

What I do not want is for you to walk away from this with nothing merged, because the substance was yours:

- The dated WAF observation (2026-08-08, `403 "You've been blocked"`, ~2.5h on a 40-page board) is the field evidence that justifies pacing at all. #2640 paces, but the *reason* to pace is your measurement.
- The rejected design (per-host limiter in `_http.mjs`, wrong because ATS hosts are multi-tenant and it would serialise the Ashby and Greenhouse tenants that run 10-wide) is a real contribution to the shared layer, and I have quoted the reasoning in #2709.

@ilyakanevskiy, when #2640 goes in, please credit @IaroslavMazur in the commit body: `Co-authored-by:` if you want to pull any of his lines, or a plain "field report and WAF measurement by @IaroslavMazur" otherwise. Both work; the point is that the person who found it shows up in the history.

Also worth saying: @ilyakanevskiy, you asked him to close in favour of yours and you did it by showing your data and offering to change your number if his was better. That is how that conversation should go and it usually doesn't.

@IaroslavMazur, the fix you wrote is landing, just under someone else's PR number. Come back with the next one, and I will be faster than I was today.

PR #2713, opened 2026-08-11
fix(providers): reject numeric entities that are legal to construct but not to emit

LayerC0de:
## The bound only prevented a crash

`providers/_html-entities.mjs` validated numeric references like this:

```js
const valid = Number.isFinite(code) && code >= 0 && code <= 0x10ffff && !(code >= 0xd800 && code <= 0xdfff);
```

That is a guard against `String.fromCodePoint` throwing a `RangeError`. It is not a guard against emitting something harmful, and those are different questions. It admits:

| reference | decodes to | |
|---|---|---|
| `&#0;` | U+0000 | NUL |
| `&#8;` `&#x1F;` | U+0008, U+001F | C0 controls |
| `&#xFFFE;` `&#xFFFF;` | U+FFFE, U+FFFF | noncharacters |

All of these survive `fromCodePoint` happily and end up in a job title.

## Why that matters in this codebase specifically

A decoded title isn't rendered once and thrown away. It is written to `scan-history.tsv`, the pipeline, the tracker, and every document generated downstream.

- **NUL** truncates anything C-string-backed that later reads those files.
- **A lone surrogate** produces ill-formed UTF-8 the moment something serializes it.
- Either one is genuinely unpleasant to trace back to a single malformed entity in a single posting three weeks later.

And the input is controlled by the feed host, so "no legitimate careers page emits `&#0;`" isn't a property we can rely on — it's a hope.

## The fix

Replace the bound with **XML 1.0 §2.2 `Char`**, which is exactly the set of code points that are safe to carry:

```
#x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF]
```

This is a **narrowing**, not an extra layer. The single predicate subsumes every previous check — `NaN` fails all of its comparisons, so `Number.isFinite` is redundant; the ranges are bounded above by `0x10FFFF` and exclude `D800–DFFF` by construction. Net change to the hot path is one function call replacing one boolean expression.

**Tab, LF and CR keep decoding.** They are legal under §2.2, they appear in real postings, and callers already normalize whitespace. Rejecting them would have been a regression wearing a safety costume.

**Rejected references are left byte-identical**, semicolon included — the existing convention in this file. Raw `&#0;` in a title is visible and inert; a decoded NUL is neither.

## Explicit non-goal: the C1 remap

HTML5 remaps numeric references in `0x80–0x9F` through windows-1252, so `&#146;` means U+2019 (’) rather than a C1 control. **This PR does not do that.** Those code points are legal under §2.2, so they pass through and decode to the raw C1 control exactly as before — unchanged behaviour.

It's a real improvement and worth doing for the HTML-scraping providers. But it *changes output* rather than *rejecting bad input*, which is a different kind of change and deserves its own review. There's a test pinning the current behaviour, so adopting it later is a deliberate decision rather than an accident.

## Provenance

This strictness is upstreamed from `providers/jobvite.mjs` (#2623), whose private copy of the decoder had quietly grown **stricter** than the shared one. @ilyakanevskiy caught it in review there and asked for it as a separate PR, since it changes behaviour for all 13 providers that import the shared module — which was the right call.

Worth naming the pattern: this is the **third** time this guard has diverged between copies (#1555, #1639), and the first time the divergence was an *improvement* trapped inside one provider. The previous two were regressions escaping into copies. A shared module stops both directions, but only if improvements get pushed back up.

Once this and #2623 both land, jobvite's local `decodeEntities` can be deleted in favour of the import. Not done here — #2623 isn't merged, and this PR is deliberately independent of it.

## Tests

18 new assertions in `tests/providers/_html-entities.test.mjs`:

- each newly rejected class — NUL (decimal and hex), C0 controls, both noncharacters
- the three C0 characters §2.2 permits, asserted to still decode
- **boundary pairs either side of every excluded range** — U+D7FF/U+DFFF/U+E000, U+FFFD, U+10000, U+10FFFF — the off-by-one a hand-written range check gets wrong
- the C1 non-goal, pinned
- that a rejected reference returns byte-identical including its semicolon

All seven pre-existing assertions pass unchanged.

## Verification

Full suite: **3449 passed, 0 failed**. No behaviour change in any of the 13 importing providers (`agentic-jobs`, `avature`, `cryptocurrencyjobs`, `dassault`, `deutschebahn`, `gem`, `hecklerkoch`, `icims`, `oraclecloud`, `remotli`, `rheinmetall`, `successfactors`, `workable`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved numeric HTML entity decoding to comply with XML 1.0 character rules.
  * Invalid control characters, surrogates, noncharacters, and out-of-range values now remain unchanged instead of being decoded.
  * Valid whitespace, printable, boundary, replacement, and astral characters continue to decode correctly.

* **Tests**
  * Added coverage for valid and invalid XML character references.

### DIFF
--- a/providers/_html-entities.mjs
+++ b/providers/_html-entities.mjs
@@ -12,11 +12,57 @@
 // the entire parse for a single malformed/adversarial entity. Centralized
 // here so the guard can't diverge again.
 //
+// The guard drifted a third time before this file caught up with it: the
+// jobvite provider (#2623) grew its own copy that was STRICTER than this one,
+// rejecting code points that are legal to construct but not legal to emit.
+// That strictness is now here, in isEmittableCodePoint below, which is the
+// point of a shared decoder — the improvement should not have had to live in
+// one provider.
+//
 // The hex/decimal alternatives are matched separately (not "#x?[0-9a-fA-F]+")
 // so a decimal entity can never absorb trailing hex letters — "&#1a2;" no
 // longer silently parses as codepoint 1 and drops "a2"; it just fails to
 // match and passes through untouched, same as any other malformed entity.
 const NAMED_ENTITIES = { amp: '&', lt: '<', gt: '>', quot: '"', apos: "'", nbsp: ' ' };
+
+/**
+ * Whether a numeric reference names a code point this decoder will emit.
+ *
+ * The set is XML 1.0 §2.2 Char. That standard is used rather than a bare
+ * `code <= 0x10FFFF` bound because the bound only prevents fromCodePoint from
+ * throwing — it says nothing about whether the result is safe to put in a job
+ * title. It admits NUL, the C0 controls, and the two noncharacters U+FFFE and
+ * U+FFFF, all of which fromCodePoint will happily emit.
+ *
+ * That matters because of where the output goes. A decoded title is not
+ * displayed and discarded: it is written to the scan history, the pipeline, the
+ * tracker, and every document generated downstream. A NUL or a lone surrogate
+ * entering there is not a rendering artefact — it truncates C-string-backed
+ * consumers, produces ill-formed UTF-8 on serialization, and is tedious to
+ * trace back to one malformed entity in one posting weeks later. The feed host
+ * controls this input, so "no legitimate page does that" is not a guarantee.
+ *
+ * Tab, LF and CR are explicitly kept: they are legal per §2.2, they appear in
+ * real postings, and the callers already normalize whitespace.
+ *
+ * NaN needs no separate guard — it fails every comparison below — so this one
+ * predicate subsumes the previous Number.isFinite / >= 0 / <= 0x10FFFF /
+ * surrogate checks, and fromCodePoint cannot throw on what survives it.
+ *
+ * Deliberately NOT done here: the HTML5 windows-1252 remap of C1 references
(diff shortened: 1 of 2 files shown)


santifer:
Thanks for this, @LayerC0de.

Separates two questions the guard was conflating: the 0x10FFFF bound kept fromCodePoint from throwing RangeError, not from emitting something harmful, and &#0; (NUL), C0 controls and the U+FFFE/U+FFFF noncharacters got through into a listing title, which isn't painted and discarded but ends up in scan-history.tsv, the tracker and everything generated afterwards, with the NUL truncating whatever reads those files months later. Replaces the bound with the XML 1.0 §2.2 Char production, underlining that this is a NARROWING and not an extra layer (the predicate subsumes the earlier checks), and leaves tab/LF/CR decoding because they appear in real listings. Marks the HTML5 C1 remap (&#146; to ’) as an explicit non-goal, pins it with a test so adopting it later is a decision rather than an accident, and notes this guard has already diverged three times across copies (#1555, #1639).

santifer:
Merged, @LayerC0de. Swapping the naive <= 0x10FFFF bound for the actual XML 1.0 §2.2 Char predicate is the correct spec-level fix, and sweeping all 3.3M code points in three notations to prove it beats any amount of spot-checking. Two of yours today. 🚀

PR #2715, opened 2026-08-11
fix(web): record status changes in the transition ledger

### LINKED ISSUE #2714: Status changes made in the web UI never reach data/status-log.tsv, so funnel-velocity and company-history have no data for web users
## Summary

`set-status.mjs` has written the transition ledger at `data/status-log.tsv` since the ledger landed, and two scripts read it: `funnel-velocity.mjs` and `company-history.mjs`. The web app's status route does not write it. Anyone who moves rows from the web UI therefore ends up with no ledger at all, and both readers have nothing to work with.

## Detail

The producer side of the contract is documented in `set-status.mjs` (the header comment describing the `{tracker#}\t{date}\t{from}\t{to}\t{source}\t` format) and implemented in the same file, guarded so it only fires on a real transition and treated as observation-only: a failed append warns and the exit code is unchanged.

`web/src/app/api/status/route.ts` performs the equivalent state change and writes nothing. It also has no reason to capture the previous status today, so the value the ledger needs is not held anywhere in the handler: the loop overwrites `parts[statusIdx]` in place and moves on.

The result is that the ledger is only as complete as the share of transitions that happened to go through the CLI. On an installation driven mainly from the web app it stays empty, and `funnel-velocity.mjs` reports on a file that does not exist.

## Reproduce

1. Change a row's status in the web UI.
2. `ls data/status-log.tsv`

Expected: one appended line for the transition. Actual: no file.

`node set-status.mjs <n> Applied` on the same installation does create it, which is what makes the asymmetry visible.

## Suggested fix

Capture the previous status in the route's existing row loop, canonicalize it so an aliased cell does not read as a fresh transition, and append through a small shared module after the tracker write has committed. Keeping the append in a plain `.mjs` module under `web/src/lib/core/` matches how the other testable web logic is arranged and lets it be covered by `web/tests/lib/` without standing up a route harness. Using a distinct value in the source column keeps web-originated rows distinguishable from CLI ones for anyone reading the ledger later.

Two things deliberately left alone, since they are separate concerns: the route's lack of a tracker lock, and any backfill of history for rows that moved before the writer existed.

rubicon:
Closes #2714

## What changed

The web status route now appends to `data/status-log.tsv`, the transition ledger `set-status.mjs` already writes and `funnel-velocity.mjs` and `company-history.mjs` already read. Before this, only CLI transitions were recorded, so an installation driven from the web app produced no ledger at all.

The route had no reason to hold the previous status, so the value the ledger needs was not available in the handler. It now captures the status cell before overwriting it and canonicalizes that value, so an aliased cell such as "aplicado" is not read as a fresh transition into Applied.

The append lives in a new `web/src/lib/core/status-log.mjs` rather than inline in the route. That matches how the other testable web logic is arranged and lets it be covered from `web/tests/lib/` without standing up a route harness. It writes the same format `set-status.mjs` writes, places the ledger beside the tracker file so a `CAREER_OPS_TRACKER` redirect keeps the two together, and records `web` in the source column so the two producers stay distinguishable.

## Why the failure handling looks the way it does

By the time the append runs, the tracker write has already committed. A ledger failure is therefore reported back to the caller and never thrown, on the same reasoning `set-status.mjs` uses when it warns instead of failing: losing one history row is better than turning a successful status change into a failed one. The response gains a `statusLogged` flag so a caller can tell the difference.

Re-selecting the status a row already has writes nothing. Self-transitions would inflate every hop count derived from the ledger.

## Deliberately out of scope

The route does not take a tracker lock, and this change does not add one. That is a real concern but a separate one, and it applies to the tracker write rather than to the ledger. Likewise, backfilling history for rows that moved before this existed is left alone.

## Verification

`web/tests/lib/status-log.test.mjs` covers the format byte for byte against what `set-status.mjs` writes, the sibling-of-tracker path, the no-op case, append rather than overwrite, the source column, the date default, an unwritable ledger returning instead of throwing, and rejection of values containing a tab or newline before anything is written.

```
node --test "tests/**/*.test.mjs"    # 137 pass, 0 fail
npx tsc --noEmit                     # clean
```

## Summary by CodeRabbit

* **New Features**
  * Status updates now record each transition, including previous and new statuses.
  * Transition records include the update source and date.
  * Responses indicate whether the transition was successfully recorded.

* **Bug Fixes**
  * Status updates remain successful if transition recording fails.
  * Unchanged statuses are no longer recorded as new transitions.
  * Invalid transition data is safely rejected without interrupting status updates.

### DIFF
--- a/web/src/app/api/status/route.ts
+++ b/web/src/app/api/status/route.ts
@@ -4,6 +4,7 @@ import path from "node:path";
 import { careerOpsRoot } from "@/lib/career-ops";
 import { canonicalizeStatus } from "@/lib/core/states";
 import { atomicWrite } from "@/lib/core/safe-write";
+import { appendStatusTransition } from "@/lib/core/status-log.mjs";
 
 // Writeback: UPDATE the status cell of an EXISTING tracker row only. Never adds
 // rows — per the core data contract, new rows go through the TSV + merge flow.
@@ -52,12 +53,14 @@ export async function POST(req: Request) {
   }
 
   let changed = false;
+  let prevStatus = "";
   for (let i = 0; i < lines.length; i++) {
     if (!lines[i].trim().startsWith("|")) continue;
     const parts = lines[i].split("|");
     if (parts.length < 8) continue;
     if (parts[1].trim() !== String(n)) continue;
     if (statusIdx >= parts.length - 1) continue; // guard malformed row
+    prevStatus = parts[statusIdx].trim(); // captured before the overwrite, for the ledger
     parts[statusIdx] = ` ${canon} `;
     lines[i] = parts.join("|");
     changed = true;
@@ -70,5 +73,20 @@ export async function POST(req: Request) {
   } catch {
     return NextResponse.json({ error: "write failed" }, { status: 500 });
   }
-  return NextResponse.json({ ok: true, status: canon });
+
+  // Record the transition in the ledger set-status.mjs already writes, so the
+  // history is the same whether a row was moved from the CLI or from here.
+  // Canonicalize the previous value first: an aliased cell ("aplicado", "sent")
+  // must not read as a transition when it already means the new status.
+  // Observation-only — the tracker write above has committed, so a ledger
+  // failure is reported alongside the successful change, never as a failure.
+  const { logged } = appendStatusTransition({
+    trackerFile: file,
+    num: n,
+    from: canonicalizeStatus(prevStatus) ?? prevStatus,
+    to: canon,
+    source: "web",
+  });
+
+  return NextResponse.json({ ok: true, status: canon, statusLogged: logged });
 }
--- a/web/src/lib/core/status-log.mjs
+++ b/web/src/lib/core/status-log.mjs
@@ -1,0 +1,54 @@
+import fs from "node:fs";
+import path from "node:path";
+
+// The web half of the transition ledger (status-log.tsv). set-status.mjs has
+// written this file since #1695, and funnel-velocity.mjs and company-history.mjs
+// read it — but status changes made in the web UI never reached it, so for
+// anyone who works from the web app the ledger stays empty and both readers have
+// no data to work with.
+//
+// Contract is set-status.mjs's, byte for byte:
+//   {tracker#}\t{date}\t{from}\t{to}\t{source}\t
+// with the ledger a SIBLING of the tracker file, so a CAREER_OPS_TRACKER
+// redirect keeps the ledger next to the tracker it describes.
+//
+// Observation-only, exactly as the CLI treats it: the tracker remains the source
+// of truth for STATE, the ledger only records WHEN. A failed append is reported
+// to the caller and never thrown, because by the time this runs the status write
+// has already committed — turning a ledger problem into a failed status change
+// would be strictly worse than losing one history row.
+
+/** A field is safe when it cannot shift the columns of a TSV row. */
+const isSafeField = (v) => !/[\t\r\n]/.test(String(v));
+
+/**
+ * Append one transition to the ledger beside `trackerFile`.
+ *
+ * @param {object}  args
+ * @param {string}  args.trackerFile  Path to applications.md; the ledger is its sibling.
+ * @param {number|string} args.num    Tracker row number.
+ * @param {string}  args.from         Previous status (canonicalized by the caller).
+ * @param {string}  args.to           New status (canonicalized by the caller).
+ * @param {string} [args.source]      What performed the change. Defaults to "web".
+ * @param {string} [args.date]        Event date YYYY-MM-DD. Defaults to today.
+ * @returns {{logged: boolean, reason?: string}} Never throws.
+ */
+export function appendStatusTransition({ trackerFile, num, from, to, source = "web", date }) {
+  const eventDate = date ?? new Date().toISOString().slice(0, 10);
+
+  for (const field of [num, from, to, source, eventDate]) {
+    if (!isSafeField(field)) return { logged: false, reason: "invalid-field" };
+  }
+
+  // Re-selecting the current status is not a transition. Logging it would inflate
+  // every hop count funnel-velocity.mjs derives from this file.
+  if (from === to) return { logged: false, reason: "no-change" };
+
+  const logPath = path.join(path.dirname(trackerFile), "status-log.tsv");
+  try {
+    fs.appendFileSync(logPath, `${num}\t${eventDate}\t${from}\t${to}\t${source}\t\n`);
+    return { logged: true };
+  } catch (err) {
+    return { logged: false, reason: `error: ${err.message}` };
+  }
+}
--- a/web/src/lib/core/status-log.mjs
+++ b/web/src/lib/core/status-log.mjs
@@ -18,8 +18,23 @@ import path from "node:path";
 // has already committed — turning a ledger problem into a failed status change
 // would be strictly worse than losing one history row.
 
+// Not every value has a string form: a null-prototype object has no toString,
+// so String() on it throws rather than returning something to inspect. Both
+// checks below run before the try, so an unguarded conversion would leave the
+// helper by exception. Treat "cannot be converted" as "not writable".
+const asText = (v) => {
+  try {
+    return String(v);
+  } catch {
+    return null;
+  }
+};
+
 /** A field is safe when it cannot shift the columns of a TSV row. */
-const isSafeField = (v) => !/[\t\r\n]/.test(String(v));
+const isSafeField = (v) => {
+  const text = asText(v);
+  return text !== null && !/[\t\r\n]/.test(text);
+};
 
 // True only when YYYY-MM-DD names a day that exists. A shape check alone admits
 // "2026-02-30" and "2026-13-40", and every reader of the ledger treats whatever
@@ -30,9 +45,11 @@ const isSafeField = (v) => !/[\t\r\n]/.test(String(v));
 // A twin of isRealCalendarDate() in followup-cadence.mjs, not an import of it:
 // Turbopack's root is pinned to web/ (next.config.mjs) and refuses modules
 // outside it, so a repo-root module cannot be imported at build time here. That
-// is the same constraint tracker-table.mjs documents.
+// is the same constraint tracker-table.mjs documents. The one difference is the
+// typeof test, which stands in for the original's String() so that a value with
+// no string form is rejected rather than throwing.
 const isRealCalendarDate = (iso) => {
-  if (!/^\d{4}-\d{2}-\d{2}$/.test(String(iso ?? ""))) return false;
+  if (typeof iso !== "string" || !/^\d{4}-\d{2}-\d{2}$/.test(iso)) return false;
   const [y, mo, d] = iso.split("-").map(Number);
   if (mo < 1 || mo > 12 || d < 1) return false;
   // setUTCFullYear rather than Date.UTC: Date.UTC maps years 0-99 onto
--- a/web/src/lib/core/status-log.mjs
+++ b/web/src/lib/core/status-log.mjs
@@ -21,6 +21,28 @@ import path from "node:path";
 /** A field is safe when it cannot shift the columns of a TSV row. */
 const isSafeField = (v) => !/[\t\r\n]/.test(String(v));
 
+// True only when YYYY-MM-DD names a day that exists. A shape check alone admits
+// "2026-02-30" and "2026-13-40", and every reader of the ledger treats whatever
+// is in the date column as a real date. Round-tripping through a UTC Date and
+// comparing the parts back catches out-of-range months as well as month-length
+// and leap-year violations, which a range check misses.
+//
+// A twin of isRealCalendarDate() in followup-cadence.mjs, not an import of it:
+// Turbopack's root is pinned to web/ (next.config.mjs) and refuses modules
+// outside it, so a repo-root module cannot be imported at build time here. That
+// is the same constraint tracker-table.mjs documents.
+const isRealCalendarDate = (iso) => {
+  if (!/^\d{4}-\d{2}-\d{2}$/.test(String(iso ?? ""))) return false;
+  const [y, mo, d] = iso.split("-").map(Number);
+  if (mo < 1 || mo > 12 || d < 1) return false;
+  // setUTCFullYear rather than Date.UTC: Date.UTC maps years 0-99 onto
+  // 1900-1999, which would reject a literal ISO year below 0100.
+  const dt = new Date(0);
+  dt.setUTCFullYear(y, mo - 1, d);
+  dt.setUTCHours(0, 0, 0, 0);
+  return dt.getUTCFullYear() === y && dt.getUTCMonth() === mo - 1 && dt.getUTCDate() === d;
+};
+
 /**
  * Append one transition to the ledger beside `trackerFile`.
  *
@@ -30,13 +52,25 @@ const isSafeField = (v) => !/[\t\r\n]/.test(String(v));
  * @param {string}  args.from         Previous status (canonicalized by the caller).
  * @param {string}  args.to           New status (canonicalized by the caller).
  * @param {string} [args.source]      What performed the change. Defaults to "web".
- * @param {string} [args.date]        Event date YYYY-MM-DD. Defaults to today.
+ * @param {string} [args.date]        Event date YYYY-MM-DD, a real calendar day. Defaults to today.
  * @returns {{logged: boolean, reason?: string}} Never throws.
  */
-export function appendStatusTransition({ trackerFile, num, from, to, source = "web", date }) {
+export function appendStatusTransition({ trackerFile, num, from, to, source = "web", date } = {}) {
   const eventDate = date ?? new Date().toISOString().slice(0, 10);
 
-  for (const field of [num, from, to, source, eventDate]) {
+  // Without this, a missing path throws out of path.dirname() below — outside
+  // the try, so no caller sees a result. "Never throws" has to hold for a
+  // caller's own mistake too, because by then the status write has committed.
+  if (typeof trackerFile !== "string" || !trackerFile) {
+    return { logged: false, reason: "invalid-field" };
+  }
+
+  // The date is held to its documented format rather than the generic check:
+  // it is the only column with one, and a date that cannot be parsed back is
+  // worse than an absent row because funnel-velocity.mjs still counts it.
+  if (!isRealCalendarDate(eventDate)) return { logged: false, reason: "invalid-field" };
+
+  for (const field of [num, from, to, source]) {
     if (!isSafeField(field)) return { logged: false, reason: "invalid-field" };
   }
 
--- a/web/tests/lib/status-log.test.mjs
+++ b/web/tests/lib/status-log.test.mjs
@@ -119,24 +119,22 @@ test("appendStatusTransition: defaults the date to today when none is given", ()
   assert.equal(date, new Date().toISOString().slice(0, 10));
 });
 
-test("appendStatusTransition: an unwritable ledger never throws", () => {
-  // Given a tracker whose directory cannot be written to
-  const tracker = fixture();
-  const dir = path.dirname(tracker);
-  fs.chmodSync(dir, 0o500);
+test("appendStatusTransition: a ledger that cannot be written never throws", () => {
+  // Given a tracker path inside a directory that does not exist, so the append
+  // cannot succeed. (Permissions are not used to provoke this: chmod does not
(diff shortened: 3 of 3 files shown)


santifer:
Thanks for this, @rubicon.

The PR for its own issue #2714: the web status route now writes data/status-log.tsv in the same byte-for-byte format as set-status.mjs, capturing the previous status before overwriting it and canonicalizing it so an aliased cell ('aplicado') isn't read as a fresh transition to Applied. The append lives in web/src/lib/core/status-log.mjs so it can be covered from web/tests/lib/ without standing up a route harness, the source column is marked 'web' to tell the two producers apart, and the ledger sits next to the tracker so a redirected CAREER_OPS_TRACKER keeps them together. A ledger failure is returned (statusLogged flag) and never thrown, since the tracker write is already committed by that point; reselecting the same status writes nothing, so transitions don't get inflated.

rubicon:
Thanks, all three were worth fixing and are addressed in 4fea3bc and bfb68c3.

**Non-throwing contract for missing arguments.** Correct, and it mattered more than it looks: the route destructures this result after its tracker write has already committed, so a throw would have reported a completed status change as a failure, which is the exact outcome the helper exists to avoid. The parameter object now defaults and a missing `trackerFile` returns the same reported result as any other rejected input.

**Date validation.** Also correct. `funnel-velocity.mjs` measures elapsed time from that column, so a value that cannot be parsed back is worse than an absent row because it still gets counted. The check now round-trips through a UTC Date and compares the parts back, which catches month length and leap years as well as out-of-range months. One detail worth noting for anyone reading it later: it builds the date with `setUTCFullYear` rather than `Date.UTC`, because `Date.UTC` maps years 0 to 99 onto 1900 to 1999 and would reject a literal ISO year below 0100.

The validator is a twin of `isRealCalendarDate()` in `followup-cadence.mjs` rather than an import of it. Turbopack pins its root to `web/` and refuses modules outside it, which is the same constraint `tracker-table.mjs` documents.

**Privilege-independent write-failure test.** Already fixed in bfb68c3, which CI found before the review did: `chmod` does not restrict directory writes on Windows, so the mode-based fixture passed locally and failed on windows-latest. It now appends into a directory that does not exist, which fails the same way on every platform.

Coverage went from 8 cases to 10. Full suite green on all three runners.

rubicon:
Both remaining review comments reproduced against the code as it stood, so both are fixed in 4fea3bc.

**Missing arguments threw instead of returning a result.** There are two paths into it, and the suggested diff covers both. A call with no argument object threw on destructuring, and a call that supplied everything except `trackerFile` reached `path.dirname(undefined)` on line 47, outside the try. Confirmed before changing anything:

```
no arguments         THREW: TypeError - Cannot destructure property 'trackerFile' of 'undefined'
trackerFile missing  THREW: TypeError - The "path" argument must be of type string. Received undefined
```

That matters at the one call site. `web/src/app/api/status/route.ts` destructures `{ logged }` from the result after `atomicWrite` has already committed the tracker row, so either throw would report a completed status change as a 500. Applied the `= {}` default and the `trackerFile` guard as suggested.

**The date column accepted values that are not dates.** Also reproduced. Every one of these landed in `status-log.tsv` with `logged: true`:

```
date="2026-02-30"  ->  "1\t2026-02-30\tEvaluated\tApplied\tweb\t\n"
date="2026-13-40"  ->  "1\t2026-13-40\tEvaluated\tApplied\tweb\t\n"
date="not-a-date"  ->  "1\tnot-a-date\tEvaluated\tApplied\tweb\t\n"
date=""            ->  "1\t\tEvaluated\tApplied\tweb\t\n"
```

The empty string was the one I had not anticipated: `??` falls back only on null and undefined, so an explicit `""` leaves the date column blank rather than defaulting to today. `eventDate` is now checked as a day that actually exists and rejected with the existing `invalid-field` reason.

One note on consistency, since it is a fair thing to ask about. The validator is a twin of `isRealCalendarDate()` in `followup-cadence.mjs` rather than an import of it. `web/next.config.mjs` pins Turbopack's root to `web/`, and it refuses modules outside that root, so a repo-root module cannot be imported at build time from this app. It is the same constraint `web/src/lib/tracker-table.mjs` already documents, where the shared source had to become a JSON file read at runtime. A function has no equivalent runtime escape hatch, so the algorithm is duplicated with its comment, and the comment names the original so the two are findable from each other. If you would rather have one copy, the route I can see is moving the helper into `web/src/lib/core/` and having `followup-cadence.mjs` import it from there, which is a bigger change than this PR should be carrying.

One thing I deliberately did not do: the guard covers `trackerFile` only, not `num`, `from`, and `to`. Those stringify to "undefined" and write a row rather than throwing, so they are a data-integrity question rather than the stability one raised here, and the route always supplies all three. Say the word if you want the helper hardened against every missing field and I will extend it.

No existing test assertion was changed or removed. The two tests are additions, one per finding.

Verified locally: `npm --prefix web test` 139 passed, 0 failed (137 before these two). `npm --prefix web run typecheck` clean. `node test-all.mjs` 3430 passed, 0 failed, and `--quick`, which is what CI runs, 3429 passed, 0 failed.

rubicon:
The `Object.create(null)` case reproduced, and it was real. Fixed in f1aaf46.

It turned out wider than the comment describes, so I fixed the conversion rather than the date. `String()` is not total, and the column check a few lines below the date check makes the same call, so every inspected field had the same exit:

```
date   = Object.create(null)   THREW: TypeError - Cannot convert object to primitive value
num    = Object.create(null)   THREW: TypeError - Cannot convert object to primitive value
from   = Object.create(null)   THREW: TypeError - Cannot convert object to primitive value
to     = { toString() { throw } }   THREW: TypeError
```

Conversion is now attempted on its own and a value with no string form is treated as not writable, so all of the above come back as `{ logged: false, reason: "invalid-field" }`. The date check uses a `typeof` test in place of its `String()` call, which is the only place the helper now differs from `isRealCalendarDate()` in `followup-cadence.mjs`. That original is only ever handed a regex capture, so the same value cannot reach it there.

Worth being explicit about what this does not change, since it touches the check I said earlier I was leaving alone. Ordinary values behave exactly as before: `num: undefined` still stringifies and writes a row rather than being rejected. Only values that cannot be converted at all changed behaviour, from throwing to reported. The missing-field hardening I offered is still on the table and still not done.

The test covers all four inspected columns rather than the date alone, since the failure was never specific to the date.

Verified: `npm --prefix web test` 140 passed, 0 failed. `npx tsc --noEmit` exit 0. `node test-all.mjs` 3430 passed, 0 failed, and `--quick` 3429 passed, 0 failed.

rubicon:
Two more from the latest pass. One was right and is fixed in fc2d860. The other I am leaving, with reasoning, so it is a decision rather than an oversight.

**The `to` column was missing from the case list.** Correct, and it was my omission: the helper validates `to` alongside the others, so a regression confined to that one column would not have been caught. Added. The case is not a free pass either, since the old conversion threw on it just like the rest.

**Reusing normalized values for comparison and append: leaving this one.** The behaviour is as described. `new String("Applied")` is not `===` any primitive, so it would slip past the no-op check, and a `toString` that returns something different on each call would be validated on one value and written on another.

What I could not find is a way for either value to arrive. The only caller builds its arguments from `JSON.parse` on the request body and from `canonicalizeStatus`, and `JSON.parse` produces neither a boxed string nor an object with its own `toString`:

```
typeof n from JSON.parse:              string
JSON.parse can yield a String object?  false
JSON.parse object prototype:           Object.prototype (inherited toString, stable)
```

So the fix would be a restructure of the validation block, on a branch under review, for an input the caller cannot construct. The tradeoff looks wrong to me at that price, and normalizing once would also quietly change what `from === to` means for values that are equal only after conversion. If you would rather have it converted once and used consistently regardless, say so and I will do it as its own commit, since the shape of that change is a fair thing to want on its own merits.

Verified after fc2d860: `npm --prefix web test` 140 passed, 0 failed. `npx tsc --noEmit` exit 0. `node test-all.mjs` 3430 passed, 0 failed, `--quick` 3429 passed, 0 failed.

santifer:
**The red CI is almost certainly not yours — don't go chasing it.** I looked before asking you for anything.

The failing job is `test (windows-latest)`, and the assertion that fails is in **`agent-inbox-tests.mjs`** ("no item is duplicated or truncated" — a 30-way concurrent-write test on the agent inbox). This PR touches three files, all under `web/`: `api/status/route.ts`, `lib/core/status-log.mjs` and its test. **Zero overlap** with that subsystem or anything it writes.

Add to that: `main`'s Windows CI has been green on its last five runs, and a 30-writer concurrency test on Windows is about the most flake-prone thing in a suite (Windows file semantics differ enough that these tests are timing-sensitive there).

I've re-run the failed job. If it goes green, it was a flake and this PR was never implicated. If it fails again consistently, then it's a real Windows-only issue in the agent inbox — still not yours, and I'll route it to the maintainer as its own thing.

Sorry you've been through two rounds and then hit a red that wasn't about your work. Reviewing the ledger itself next.

santifer:
**Correcting myself: I told you it was probably a flake. It isn't.** The conclusion still holds — it is not your PR — but the reasoning I gave you was wrong, and you deserve the right version.

It fails **consistently** on Windows across three runs, and the assertion is:

```
```

One item vanishes. That's a genuine silent data-loss race in the agent inbox on Windows, now tracked as **#2777 (P0)** on the maintainer's side, with a real fix (routing the append through the shared lock) rather than skipping the test on Windows — skipping would hide data loss on a platform we support.

**Where my reasoning failed, since it might save you the same mistake:** I checked that `main`'s last five Windows runs were green and treated that as evidence the test was healthy. But the test only landed **this morning at 10:05 UTC** — so four of those five runs predate it entirely and *couldn't* have caught anything, and the fifth was the coin-flip landing heads. I was reading history that was blind to the thing I was testing for. "Main is green" is only evidence if the test existed during those greens.

**What it means for you:** nothing to do. Your CI goes green on its own once #2777 lands. You've now been through two review rounds and a red that was ours twice over — first as someone else's failure, now as our own. Reviewing the ledger itself next.

santifer:
Reviewed. The implementation is right and I want it — but as written, **the reader will treat every row it produces as second-class**, and fixing that needs a one-line change in a core file, so it can't happen inside this PR alone.

## What I verified

The line format matches the core exactly:

```js
core (set-status.mjs:504)  `${target.num}\t${eventDate}\t${oldStatus}\t${newStatus}\tset-status\t\n`
this PR                    `${num}\t${eventDate}\t${from}\t${to}\t${source}\t\n`
```

Same field order, same count, same trailing tab, same sibling-of-the-tracker path. The only difference is the `source` value, which is exactly the field that should differ.

`isSafeField` rejecting `[\t\r\n]` is the right guard for a TSV writer — a status or note carrying a tab would shift every later column, and this is a file two other tools parse positionally. `isRealCalendarDate` doing a real round-trip check instead of a regex is also correct: `2026-02-31` matches `\d{4}-\d{2}-\d{2}` and isn't a date.

Canonicalizing the *previous* status before writing it (so an aliased `"aplicado"` isn't recorded as a distinct from-state) is the detail I'd have most expected to be missed. Good.

## The problem: `source = "web"` isn't a source the reader knows

`funnel-velocity.mjs:54`:

```js
const VALID_SOURCES  = new Set(['set-status', 'correction', 'backfill', 'manual']);
const DAY_MATH_SOURCES = new Set(['set-status', 'correction']);
```

`"web"` is in neither, and this PR doesn't touch that file. What actually happens (I misread this at first and want to be precise — the rows are **not** discarded):

```js
if (!VALID_SOURCES.has(source)) {
  unknownSources.push({ line: i + 1, num, source });
  observations.push({ …, dayMath: false });   // counted, but never fed to day-math
  continue;
}
```

So every web transition is: **parsed, counted as an observation, excluded from the velocity maths, and reported under `dataQuality.unknownSources`.** An installation driven from the web would get a ledger — the thing this PR is for — whose entries are all flagged as questionable and contribute nothing to "days between states".

And the exclusion is wrong on the merits. The comment above `DAY_MATH_SOURCES` says why `backfill`/`manual` are excluded: *"they are reconstructed after the fact, not observed."* A web status change is **observed live**, at the moment the user makes it — it's the same event as a `set-status` transition, entering through a different door. It belongs in both sets.

## What I'd like

Add `'web'` to `VALID_SOURCES` **and** `DAY_MATH_SOURCES` in `funnel-velocity.mjs`. That's a core file, so I'm raising it with the core maintainer rather than asking you to reach into it — I'll come back here once it's agreed, and then this merges.

Two smaller things while that settles:

1. Worth a test asserting the round trip end-to-end: write a transition through `appendStatusTransition`, then parse the file with `funnel-velocity`'s reader and assert the row comes back as a day-math observation. Right now both halves are tested, but not that they agree — which is exactly the gap that let this one hide.
2. `company-history.mjs` also reads this ledger; I haven't checked whether it filters by source. I'd rather say that than assume it doesn't.

Nothing here is a criticism of the work — the divergence is between your writer and a reader you had no reason to open. Sorry it took this long to get a proper review on it.

rubicon:
Thank you for the review — and for going and reading the reader. That divergence was real and I had no test that could have caught it.

All three points are addressed. Rebased onto current `main` as well, so this is no longer behind.

## The core change is written and running

You said you'd rather raise `funnel-velocity.mjs` with the core maintainer than ask me to reach into it. That's the right instinct, but it turns out I'd already hit this from the other side and have the patch — so rather than have you open that conversation from scratch, here it is to take or leave:

**#2897** (issue) → **#2898** (PR). Two lines, adding `web` to `VALID_SOURCES` and `DAY_MATH_SOURCES`, plus three self-test assertions.

It's deliberately reader-only, with no writer changes, so it stands on its own regardless of what happens to this PR. Entirely your call whether it lands — if you'd rather route it differently, close it and I won't be offended.

Your reasoning for putting it in **both** sets is the part I'd underline. Admitting `web` to `VALID_SOURCES` alone would clear the `unknownSources` flag and still leave every figure empty, which is arguably a worse outcome than today: the symptom disappears and the defect doesn't.

## 1. The round-trip test — added, and it found the shape of the gap

Added to `web/tests/lib/status-log.test.mjs`: it writes two transitions through `appendStatusTransition`, then parses the real file with `funnel-velocity`'s real `parseStatusLog`, asserting nothing lands in `unparseable` and that num/date/from/to survive.

One deliberate omission, called out in a comment in the test itself: it does **not** assert the rows are trusted for day-math. That is true of the format but not yet of the reader, so asserting it here would fail for a reason this PR cannot fix from inside `web/`. It tightens to include `dayMath` once #2898 lands.

You diagnosed the gap precisely, and it's worth stating why the obvious version of this test wouldn't have helped: asserting the row *parses* would have passed all along. The rows always parsed. What failed was that a parsed row reached no figure — so the assertion has to follow the observation all the way to a hop, not stop at the parser. #2898's third assertion does exactly that, and it's the only one of the three that fails for the interesting reason.

## 2. `company-history.mjs` — checked, and it needs no change

You were right to flag it rather than assume. I read it:

It has **no source allow-list of its own.** It never parses the ledger — it loads it by dynamic import of `funnel-velocity.mjs` and consumes that reader's output, taking `statusLogAppliedByNum` as an injected Map. So it inherits the exclusion rather than repeating it, and fixing the two sets in `funnel-velocity.mjs` fixes both readers at once.

That's the good version of the answer: one list, one place, no second copy to drift.

## Still outstanding, and not for this PR

While working on the reader side I hit something larger that this PR doesn't address and shouldn't: the status route performs a read-modify-write on `data/applications.md` with **no lock**. `atomicWrite` makes the write atomic about the file, not about the read-modify-write around it, so a status change made from the web app concurrently with `merge-tracker.mjs` or `set-status.mjs` silently discards whichever committed first.

This PR is correct as scoped and I'd rather not widen it one line from merging. Filing it separately with a fix shortly, and I'll link it here.

rubicon:
As promised: **#2900** (issue) → **#2901** (PR).

It reproduces the unlocked write against a dev server with another process genuinely holding the lock — same request both times:

| | HTTP | tracker after |
|---|---|---|
| before | `200 {"ok":true,"status":"Rejected"}` | **row rewritten while the lock was held** |
| after | `503 {"code":"lock-timeout"}` + `Retry-After` | byte-identical |

**One thing you should decide rather than have decided for you.** Because `set-status.mjs` has written `status-log.tsv` since #1695, routing the web write through it produces the ledger row as a *consequence* of the write. That makes this PR's `status-log.mjs` module redundant — the same outcome, without a second writer racing beside an unlocked one.

I deliberately based #2901 on `main` rather than stacking it on this branch, so the ordering is yours:

- **Take #2901** → I close this one, and the ledger arrives with the lock fix.
- **Take this first** → I rebase #2901 to remove the module instead, and nothing here is wasted.

Either way #2898 should land at or before whichever writer you pick, or the rows it produces go into `unknownSources`.

No preference from me beyond wanting the race closed. This PR is correct as scoped and I'm not trying to talk you out of it — I'd just rather you see the overlap now than after merging.

santifer:

rubicon:
**Recommending this be closed, unmerged.** I offered you the fork on 14 Aug; this is me picking the side that costs you less, rather than leaving it open for someone else to work out.

## Why it is redundant rather than parallel

You named it yourself on #2904 — `--source web` makes the ledger a consequence of the write. Checking that against the code rather than the claim, `set-status.mjs` already produces every behaviour this PR's module was built for:

| this PR | `set-status.mjs` |
|---|---|
| ledger row in the core's exact format | the line it is a copy of |
| `source` column `web` | `--source web` (#2901) |
| canonicalizes the previous status | resolves from the tracker |
| self-transition writes nothing | appends only `if (statusChanged && !flags.dryRun)` |
| `statusLogged` returned, never thrown | already in its `--json` payload |

Not "close enough" — the same rows, from one writer instead of two.

## Why not keep it as a fallback

That was my instinct too, and I do not think it survives contact with what it would cost you. It is `CONFLICTING` against a `route.ts` that has been rewritten under it by #2903, so it is not a fallback you could take — un-rotting it means re-authoring the module against a route it was never written for, at which point it is a new PR wearing this one's number. Meanwhile it sits in the `area:web` queue looking like work, in front of a maintainer who has not read it, on a design the core maintainer has already said he wants converged.

I have not rebased it, deliberately. Rebasing a PR I am recommending you close would be me arguing for it with my hands while saying the opposite in text.

If #2904 rules against delegation, the honest fallback is a fresh PR against #2903's locked route — smaller than this one, because the lock is already there and only the ledger append would be missing.

## What should not be lost with it

The review rounds on this branch found four real defects in the helper, and three of them are properties any second producer would need again: rejecting a `[\t\r\n]` in a TSV field, holding the date column to a day that actually exists rather than to a regex (`2026-02-31` passes `\d{4}-\d{2}-\d{2}`), and treating a value with no string form as unwritable instead of throwing mid-write. They are worth keeping findable even with the branch gone; #2823 documents the two reader-side rules that motivated the same class of bug.

Also, for the record, the round-trip test here is what surfaced the reader divergence that became #2897/#2898 — the one PR of this cluster that is approved and ready. This branch paid for that.

Yours to close rather than mine to abandon quietly, so I am leaving it open and saying plainly: I would close it, and I would not count it as wasted.

rubicon:
Withdrawing this one. **The bug is not withdrawn — [#2714](https://github.com/santifer/career-ops/issues/2714) stays open, and I have just re-confirmed it is live on `main`.**

```
$ git show upstream/main:web/src/app/api/status/route.ts | grep -c 'status-log'
0
```

#2903 closed the lost update (`withTrackerLock` is there), but the route still writes only the Status cell. A status change made in the web UI reaches `data/applications.md` and never reaches `data/status-log.tsv`, exactly as #2714 describes.

What I am withdrawing is this PR's **mechanism**, for three reasons:

1. **It adds a second web-side writer**, and [#2904](https://github.com/santifer/career-ops/issues/2904) exists to decide whether the web should write the tracker at all. If delegation wins, `set-status.mjs --source web` produces the ledger row as a consequence of the write — your phrasing was that it makes this "unnecessary rather than parallel", and I agree.
2. **It cannot be revived as-is.** It is `DIRTY` against a `route.ts` that #2903 rewrote underneath it. Even if #2904 rules against delegation, the fallback would be a fresh PR on the post-#2903 locked route, not this branch.
3. **It is sitting in a reviewer's queue** describing an architecture that may be decided away this week. Keeping it open as insurance is not insurance; it is deferred cleanup that someone else pays for in review time.

The pieces worth keeping are already elsewhere and do not depend on this branch: the line format and its two enforced rules are in [#2823](https://github.com/santifer/career-ops/pull/2823), and the reader-side half — `web` being accepted as a first-hand source rather than parsed-then-excluded — is [#2898](https://github.com/santifer/career-ops/pull/2898), which you have approved.

Once #2904 rules, I will send the ledger fix in whichever shape it decides, against current `main`. If that ruling is delegation, no separate ledger PR is needed at all.

Sorry for the review time this consumed after the ground moved.

PR #2718, opened 2026-08-11
test(scan): pin the company canonicalizer argument through loadDedupSnapshot

Scott-Emberson:
Closes the gap you flagged when merging #2601: the snapshot test pinned the recheck `policy` argument as live but not `buildCompanyCanonicalizer(config.company_aliases)`, so dropping the third argument from the `collectSeenCompanyRoles` call inside `loadDedupSnapshot` left the suite green while `company_aliases` silently stopped applying to company+role dedup.

The reason checks 1 and 2 could not see it: every fixture company canonicalizes to itself under `defaultCompanyNormalizer`, so the argument was invisible on that fixture no matter what it did. The new check passes an alias map that rewrites one fixture company (`Zeta` → `Zeta Industries`) and asserts the role key lands under the canonical spelling, with the raw key absent and an unaliased company untouched.

I re-ran your mutation before committing: with the third argument dropped at the `collectSeenCompanyRoles` call, the new check fails with `hasCanonical=false hasRaw=true`; restored, it passes. Full suite 3431/0, the one warning being the pre-existing Polish CV reorder fixture asserting its own warning.

Thanks for the differential probe on the goldens, and for catching that this refactor is what made the argument losable in the first place.

## Summary by CodeRabbit

* **Tests**
  * Expanded scan deduplication coverage for company aliases.
  * Verified that alternate company names are consistently mapped to their canonical names when generating role keys.
  * Preserved coverage for fresh-install and empty-file scenarios while updating test organization to include the new validation case.

### DIFF
--- a/tests/scan-dedup-snapshot.test.mjs
+++ b/tests/scan-dedup-snapshot.test.mjs
@@ -21,7 +21,7 @@ import { pass, fail } from './helpers.mjs';
 import { mkdtempSync, writeFileSync, mkdirSync, rmSync } from 'fs';
 import { tmpdir } from 'os';
 import { join } from 'path';
-import { loadDedupSnapshot } from '../scan.mjs';
+import { loadDedupSnapshot, buildCompanyCanonicalizer } from '../scan.mjs';
 
 console.log('\nscan.mjs — dedup snapshot: one read per source, golden outputs (#2382)');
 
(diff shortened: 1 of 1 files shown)


santifer:

Closes the gap I flagged when merging #2601: the snapshot test pinned the recheck's policy argument as live but not buildCompanyCanonicalizer(config.company_aliases), so dropping the third argument from the collectSeenCompanyRoles call inside loadDedupSnapshot left the suite green while company_aliases quietly stopped applying to company+role dedup. Explains why checks 1 and 2 couldn't catch it (every fixture company canonicalizes to itself under defaultCompanyNormalizer, so the argument was invisible whatever it did) and adds an alias map that rewrites Zeta to Zeta Industries. My mutation was re-run before committing: hasCanonical=false hasRaw=true without the argument, green with it.

santifer:
Verified the claim the way it deserved, by mutation rather than by reading: dropped `canonicalize` from the `collectSeenCompanyRoles` call inside `loadDedupSnapshot`, exactly as your comment predicts.

Your new check goes red with the right diagnosis pinned to it, `hasCanonical=false hasRaw=true hasAcme=true`, and **all seven pre-existing checks stay green**. That is the whole PR in one measurement: the file already had coverage for this function and none of it could see the second argument, because every fixture company canonicalizes to itself and a defaulted normalizer is indistinguishable from a live one until something in the fixture actually needs renaming.

What that guards is not cosmetic either. If `company_aliases` silently stops applying to company+role dedup, the same role gets listed twice under two spellings and the user reads it as two openings.

This is the second time today one of your reviews or tests has been about *the thing a green suite cannot see*, and it is the most useful axis anyone is working on here.

Merged. 🚀

PR #2719, opened 2026-08-11
test(skills): create .git/info before emptying the fixture exclude file

ringo380:
You flagged this reviewing #2567 and merged the rest without it, so here it is on its own.

`test-all.mjs` section 12c empties `.git/info/exclude` so an ambient template cannot leak ignore rules into the fixture. The write assumed `.git/info/` exists, which holds only for the default git template.

Both shapes you named reproduce. `GIT_TEMPLATE_DIR` pointing at an empty directory, and pointing at a path that does not exist (git init warns and continues), each produce a repository with no `.git/info` at all. The bare write then threw ENOENT before the fixture ran a single assertion, so a block added to remove an environment dependency introduced a new one of the same family.

`mkdirSync(dirname(...), { recursive: true })` before the write.

## Verification

Replicated the fixture's init-and-write, once as merged and once with the fix, across all three template shapes:

```
case                                       as merged    with fix
default template (no GIT_TEMPLATE_DIR)     ok           ok
GIT_TEMPLATE_DIR -> empty dir              ENOENT       ok
GIT_TEMPLATE_DIR -> missing path           ENOENT       ok
```

Full suite on a clean worktree at `main` plus this change: 3430 passed, 0 failed, 1 warning (`cv-sync-check.mjs exited with error (expected without user data)`, a fresh-clone artifact).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved setup reliability when the configured Git template directory is empty or does not yet exist.
  * Ensured required Git metadata directories are created automatically before setup files are written.

### DIFF
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -5301,7 +5301,16 @@ console.log('\n12c. Materialized skill index mode');
     // core.excludesFile is only the GLOBAL layer. `git init` also seeds
     // .git/info/exclude from a template, which GIT_TEMPLATE_DIR can still point
     // at an ambient one, so empty that layer too rather than assume it is inert.
-    writeFileSync(join(fixtureRoot, '.git', 'info', 'exclude'), '');
+    //
+    // mkdirSync first, because the same GIT_TEMPLATE_DIR that makes this write
+    // necessary is what can make its parent absent: pointed at an empty or a
+    // non-existent directory, `git init` still succeeds but seeds no .git/info,
+    // and the bare write threw ENOENT before the fixture ran a single assertion
+    // (santifer, reviewing #2567). Assuming the default template here would be
+    // the same ambient-environment dependency this block exists to remove.
+    const excludePath = join(fixtureRoot, '.git', 'info', 'exclude');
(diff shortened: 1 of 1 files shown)


santifer:
Thanks for this, @ringo380.

Section 12c of test-all.mjs empties .git/info/exclude so no ambient template leaks ignore rules into the fixture, but the write assumes .git/info exists, and that only holds with the default git template. Both forms I named were reproduced (GIT_TEMPLATE_DIR pointing at an empty directory and at a nonexistent path) and in both the repo comes up with no .git/info and the write dies with ENOENT before the fixture runs a single assert: a block added to remove one environment dependency introduced another from the same family. The fix is a recursive mkdirSync, with a before/after table for the three cases.

santifer:
Merged, @ringo380. The comment block you left explains the trap better than my review of #2567 did: a GIT_TEMPLATE_DIR that makes the write necessary is the same thing that can make its parent absent. Thanks for closing the loop I left open: this one was mine to fix and you beat me to it. 🚀

PR #2720, opened 2026-08-11
fix(verify-cv-facts): fold allow_metrics through the same noun synonyms

ringo380:
Closing out a review comment on #2175 that I left unanswered at the time.

Claims extracted from text are canonicalized through `NOUN_SYNONYMS`, so `77 repos` on a CV becomes `77 repositories`. `allow_metrics` entries went through `normalizeClaim` only. An exception written in the spelling a human actually reaches for therefore matched nothing: the entry did nothing at all, the CV still failed the gate, and nothing in the output pointed at the allow-list as the reason.

That is the same failure class this script exists to catch, one layer up. The user sees the exact claim they just allowed still reported as invented, with no signal that their exception was inert.

## The shape of the fix

Both call sites (`auditClaims` and `verifyFacts`) built the allow-list independently, so this extracts `allowedMetricSet()` and routes both through it.

Allow-list entries now contribute **both** spellings rather than the canonical one alone. `metricClaims()` yields nothing for an entry no claim pattern recognizes, such as a bare currency figure or a percentage, so replacing `normalizeClaim` outright would silently drop those exceptions instead of widening them, trading one silent no-op for another. The union can only ever allow more, never less.

Also hoists `stripMarkup(targetText)` out of the forbidden-phrase filter, where it re-ran the whole markup pass once per configured phrase. Same result, one pass.

## Verification

Four new self-test assertions:

- a synonym-spelled exception is honoured
- the canonical spelling still works
- a currency exception survives the folding
- an unrelated exception still leaves the claim invented

Seen failing first: removing the synonym folding turns the first one red (`expected [] / actual ["77 repositories"]`), and the other three stay green, which is what pins the "widen, never narrow" property.

`verify-cv-facts.mjs --self-test`: 43 passed, 0 failed. Full suite on a clean worktree: 3430 passed, 0 failed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved verification of factual claims against configured metric exceptions.
  * Added support for synonym-spelled and currency-based metric exceptions.
  * Prevented unrelated exceptions from being incorrectly treated as valid matches.

### DIFF
--- a/verify-cv-facts.mjs
+++ b/verify-cv-facts.mjs
@@ -266,16 +266,41 @@ export function metricClaims(text) {
   return claims;
 }
 
+/**
+ * Build the allow-list a metric claim is checked against.
+ *
+ * Claims extracted from text are folded through NOUN_SYNONYMS; allow_metrics
+ * entries used to be normalized only, so an exception written in the spelling
+ * a human reaches for - `77 repos` - never matched the canonical `77
+ * repositories` the extractor produces. The entry then did nothing at all and
+ * the CV still failed the gate, with no diagnostic pointing at the allow-list
+ * (CodeRabbit, reviewing #2175). A silently inert exception is the same failure
+ * class this script exists to catch.
+ *
+ * Both spellings are added rather than the canonical one alone: metricClaims()
+ * yields nothing for an entry no pattern recognizes (a bare `$900k`, a
+ * percentage), so replacing normalizeClaim outright would drop those exceptions
+ * instead of widening them. The union can only ever allow more, never less.
+ */
+function allowedMetricSet(sourceText, allowMetrics) {
+  const allowed = new Set(metricClaims(sourceText));
+  for (const entry of allowMetrics || []) {
+    allowed.add(normalizeClaim(entry));
+    for (const canonical of metricClaims(String(entry))) allowed.add(canonical);
+  }
+  return allowed;
+}
+
 /** Compare generated metric claims against source text without reading files. */
 export function auditClaims(targetText, sourceText, config = {}) {
-  const allowed = new Set([
-    ...metricClaims(sourceText),
-    ...(config.allow_metrics || []).map(normalizeClaim),
-  ]);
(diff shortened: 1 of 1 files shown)


santifer:
Thanks for this, @ringo380.

Claims get canonicalized through NOUN_SYNONYMS ('77 repos' becomes '77 repositories'), but allow_metrics entries only ran through normalizeClaim: an exception written with the word a human actually uses matched nothing, the exception sat inert, the CV kept failing the gate, and nothing in the output pointed at the allow-list as the cause. Extracts allowedMetricSet() for both call sites and contributes BOTH spellings instead of just the canonical one, because metricClaims() returns nothing for a currency figure or a percentage and swapping in normalizeClaim would have silently dropped those exceptions: one silent no-op traded for another. Along the way it hoists stripMarkup(targetText) out of the banned-phrase filter, where it re-ran once per configured phrase.

santifer:
Closing out an unanswered review comment from #2175 six weeks later, unprompted, is its own kind of contribution. Thanks for going back for it.

Reproduced before reading your patch, and the live behaviour is worse than your description. With `allow_metrics: ["77 repos"]` and a document claiming "Maintained 77 repos":

```
$ node verify-cv-facts.mjs doc.md --source cv.md --config cfg.json     # main
CV fact check failed: doc.md
Metric-like claims absent from sources:
  - 77 repositories                                                    # exit 1
```

The exception is inert, and the failure reports **`77 repositories`**, a string the user never typed. So they are shown a claim in a spelling they did not write, cannot match it against the allow-list entry they did write, and nothing anywhere says the entry was the reason. Your branch: `CV fact check passed`, exit 0.

Contributing **both** spellings rather than swapping `normalizeClaim` for `metricClaims` is the right call for the reason you give: an entry no claim pattern recognises, a bare currency figure or a percentage, would yield nothing and the exception would go from inert-in-one-spelling to inert-entirely. Extracting `allowedMetricSet()` so `auditClaims` and `verifyFacts` stop building the list independently is the part that keeps this from drifting back.

Suite on a preview against today's main: green.

Merged. 🚀

PR #2721, opened 2026-08-11
fix(jd-skill-gap): recognize all six markdown heading levels

ringo380:
Closing out a review comment on #2176 that I left unanswered at the time.

Markdown defines six heading levels. The requirement-header prefix, the non-requirement prefix, the block-ending fallback in `scanJd()`, and the `cv.md` Skills scanner all stopped at four.

Two ways that bites a real posting:

1. **No block opens at all.** A JD pasted out of a deeply nested doc, or converted from HTML by an ATS that wraps its sections in h5/h6, produces an empty skill list. That reads exactly like "no gaps found", which is the silent failure mode this file already warns about in its own header comment.
2. **An open block never closes.** Where a requirements block does open, its own next heading no longer ends it, so the benefits and about-us sections below score as required skills. The same "Carrot as a missing skill" outcome, reached by a different route.

Widened all four sites to `#{1,6}` / `#{0,6}`.

## Verification

Three new self-test assertions, on a fixture using h5 and h6 headings:

- an h5 requirements heading is recognized
- an h5 heading closes the block
- an h6 heading stays closed

Seen failing first, and isolated per mechanism rather than as one lump: reverting the heading levels turns the first red. The two closing assertions are covered redundantly, by the non-requirement prefix and by the block-ending fallback independently, so reverting either alone leaves them green and reverting both turns them red. Noting that explicitly because a reviewer checking one mutation would reasonably read them as vacuous.

`jd-skill-gap.mjs --self-test`: 63 passed, 0 failed. Full suite on a clean worktree: 3430 passed, 0 failed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved job description and CV heading detection across all Markdown heading levels.
  - Fixed requirement sections ending correctly around deeper headings and responsibility headings.
  - Preserved recognition of “You Will Have” as a requirements header.
  - Improved handling of benefits, “About” sections, and other content following requirements.

- **Tests**
  - Added coverage for deep headings and requirement-section boundary cases.

### DIFF
--- a/jd-skill-gap.mjs
+++ b/jd-skill-gap.mjs
@@ -433,6 +433,24 @@ Python, Docker, Zookeeper
     true
   );
 
+  // The cv.md side of the six-level widening: SKILLS_HEADING_RE and
+  // ANY_HEADING_RE also stopped at four. An h5 "Skills" heading meant no
+  // named section was found at all (everything downgraded to
+  // supportedByResume), and once it IS found, the h6 heading after it must
+  // still close the section - otherwise Experience prose leaks into the
+  // named skills and upgrades to existing. One assert per regex: reverting
+  // SKILLS_HEADING_RE to #{1,4} turns the first red, reverting
+  // ANY_HEADING_RE alone turns the second red.
+  const deepHeadingCv = [
+    '# Resume', '',
+    '##### Skills', 'Python, Docker, PostgreSQL', '',
+    '###### Experience', 'Deployed Kubernetes clusters for internal tools.',
+  ].join('\n');
+  const deepCvResult = classifySkillGaps(['Python', 'Docker', 'PostgreSQL', 'Kubernetes'], deepHeadingCv);
+  eq('an h5 Skills heading is recognized as the named section', deepCvResult.existing.includes('Python'), true);
+  eq('the named section stops at the h6 heading (Kubernetes stays prose)', deepCvResult.existing.includes('Kubernetes'), false);
+  eq('prose under the h6 still classifies as supportedByResume', deepCvResult.supportedByResume.includes('Kubernetes'), true);
+
   // Regression: requirement headings that are full sentences or bare
   // uppercase rather than the noun forms ("Requirements", "Qualifications").
   // Each of these silently yielded ZERO skills, which is indistinguishable
--- a/jd-skill-gap.mjs
+++ b/jd-skill-gap.mjs
@@ -48,7 +48,7 @@ const CV_PATH = 'cv.md';
 // JD could yield zero skills - which reads identically to "no gaps found" and
 // is the more dangerous of the two failure modes this file warns about.
 const REQUIREMENT_HEADER_RE = new RegExp(
-  '^#{0,4}\\s*(?:' + [
+  '^#{0,6}\\s*(?:' + [
     'required', 'requirements', 'qualifications', 'must[- ]have', 'preferred', 'nice[- ]to[- ]have',
     "what\\s+we(?:'|’)?\\s*re\\s+looking\\s+for",
     "what\\s+you(?:(?:'|’)ll|\\s+will)?\\s+bring",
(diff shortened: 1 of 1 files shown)


santifer:
Thanks for this, @ringo380.

Markdown has six heading levels and four spots in jd-skill-gap.mjs stopped at four: the requirements prefix, the non-requirements one, the block-close fallback in scanJd(), and the Skills scanner over cv.md. Two ways a real listing breaks: a JD converted by an ATS that wraps sections in h5/h6 never opens a block and returns an empty skills list, which reads exactly like 'no gaps' (the silent failure the file's own header warns about); and where a block does open, the next heading fails to close it, so benefits and about-us score as requirements. Flags that two of the three asserts are redundantly covered by two mechanisms, so they don't read as vacuous if only one is mutated.

santifer:
Everything verifiable is green here, @ringo380, and your self-assessment held up when I checked it rather than assumed it. Reverting `NON_REQUIREMENT_HEADER_RE` alone (63/0) or the `scanJd` fallback alone (63/0) both stay green, and only reverting both together goes red (61/2) - exactly the redundant coverage you described in the body, which also answers CodeRabbit's first comment about whether those asserts are vacuous.

One concrete ask, and mutation testing is what surfaced it.

**The fourth widened site has no net.** Reverting `SKILLS_HEADING_RE` + `ANY_HEADING_RE` to `#{1,4}` turns *nothing* red across all 3611 suite assertions, because `test-all.mjs` delegates this file entirely to `--self-test`. Yet that site demonstrably changes real classifications: a CV with `##### Skills` followed by `###### Experience` classifies Python/Docker/PostgreSQL as `supportedByResume` on base and `existing` on your branch, which is the correct outcome. So the behavior change is real and unprotected at the same time, and a future refactor can silently revert it.

The ask is one fixture: a cv.md self-test case with `##### Skills` followed by `###### Experience`, asserting the named skills stop before the h6. That is CodeRabbit's second actionable, and it matches the convention your own file already keeps (one test per previous widening, 63 asserts deep).

Two things I checked so you do not chase them:

- **The code-fence question is not yours to answer here.** The file has no fence awareness anywhere (zero hits for fences in the source), and that is preexisting: on today's main, `#### preferred settings` inside a fence already opens a requirements block and `# install deps` already closes one. Your change widens that marginally at depths 5-6, where nobody writes `##### install deps`. Asking you to add fence awareness would be asking you to fix an architectural hole you did not open. Separate issue, not a blocker.
- **No h1-h4 regression:** the control case `## Skills` is byte-identical between base and head, so the path every example CV in the repo uses is untouched.

Add the fixture and this goes straight in. Five of yours today, and each one has been a real defect with a real repro.

ringo380:
Fixture added in 541102c - one cv.md case, `##### Skills` followed by `###### Experience`, three asserts:

- an h5 Skills heading is recognized as the named section (Python classifies `existing`)
- the h6 heading after it still closes that section (Kubernetes must not be `existing`)
- the prose under the h6 stays `supportedByResume`

Mutation results now match the convention on the JD side. Before writing it I reproduced your finding - both regexes reverted to `#{1,4}` stayed 63/0 green. With the fixture: reverting `SKILLS_HEADING_RE` alone goes red on the first assert (65/1), and reverting `ANY_HEADING_RE` alone goes red on the other two (64/2). Self-test is 66/0 on the branch; full suite 3430 passed / 0 failed in a fresh worktree, the one warning being the expected cv-sync-check without user data.

Noted on fence awareness - leaving it alone here. If you want it tracked I can file it as a separate issue with your `#### preferred settings` repro, so it does not get lost.

santifer:
The part worth naming is that you changed **both** regexes, not just the one the title is about.

Widening only `REQUIREMENT_HEADER_RE` would have looked like the fix and quietly made things worse: a `###### Benefits` heading would have started matching as requirements while nothing recognised it as a non-requirement, and the file's own comment says exactly what that produces ("turning perks like 401k, Equity and Carrot into reported skill gaps"). Both moved to `#{0,6}` together, so the ordering that protects that stays intact.

Verified by running it rather than reading the diff, on a JD with `##### Requirements` and `###### Benefits`:

- `main` → `"lowConfidence": {"reason": "no-requirements-section"}`, zero gaps. The whole requirements block is invisible, and the output claims low confidence rather than claiming nothing is wrong, which is at least honest.
- your branch → `"gap": ["Kubernetes", "Terraform"]`, and **401k and Carrot correctly absent**. That second half is the control that proves the pairing held.

Suite on a preview against today's main: green.

Merged. 🚀

PR #2727, opened 2026-08-11
test(scan): stop the suite inheriting a user's own lane overrides

ringo380:
Found while going back through the review comments on #2568. CodeRabbit raised it there and I never answered; it is still live, and it is worse on a customized install than the original note suggested.

`runScan` spread `process.env` into every child, so the suite's behavior depended on the environment it was run in. Check 1 asserts DEFAULT output paths, and an inherited `CAREER_OPS_PIPELINE` redirects precisely that check.

The audience is what makes it concrete. The docs added with this feature tell a second-lane user to export `CAREER_OPS_PIPELINE` and `CAREER_OPS_SCAN_HISTORY`, so the person most likely to run this suite with them set is the person the feature was written for. What they get is fixture postings appended to their real inbox, and the matching rows written into `scan-history.tsv`. The second half is the quiet one: that file is the dedup source, so a real posting colliding with a fixture row is skipped silently on a later scan, which is the same failure this feature exists to prevent.

Clears the four scanner path variables before applying each check's explicit overrides.

## Why there is a check 5

Asserting the isolation rather than assuming it, for the same reason as the clearing itself: on a machine with these variables unset, removing the clearing changes nothing observable, so nothing would catch its removal.

Check 5 sets the variables in the parent, runs through the same `runScan()` the other checks use, and requires both that the lane's own file received the postings and that the ambient paths were never created. A case that spawned `scan.mjs` with its own hand-built environment would have stayed green.

## Verification

`tests/scan-output-paths.test.mjs`: 5 passed, 0 failed. Full suite: 3431 passed, 0 failed, 1 warning (`cv-sync-check.mjs exited with error (expected without user data)`, a fresh-clone artifact).

Mutation: removing the clearing turns check 5 red with

```
suite is not isolated from the ambient environment: 0 lane entr(y/ies), ambient files WRITTEN (#2568)
```

which is the reported bug verbatim: zero postings in the lane's own file, and both ambient paths written instead.

## Summary by CodeRabbit

* **Tests**
  * Improved scan path test isolation by clearing environment-based overrides between tests.
  * Added coverage confirming ambient configuration cannot redirect default scan paths.
  * Added cleanup for restored environment settings and temporary test directories.

### DIFF
--- a/tests/scan-output-paths.test.mjs
+++ b/tests/scan-output-paths.test.mjs
@@ -47,12 +47,35 @@ function makeLane() {
   return { dir, portals };
 }
 
-const runScan = (dir, env) => execFileSync(NODE, [join(ROOT, 'scan.mjs')], {
-  cwd: dir,
-  env: { ...process.env, ...env },
-  encoding: 'utf-8',
-  stdio: ['ignore', 'pipe', 'pipe'],
-});
+// Every variable scan.mjs resolves a path from. Cleared before each spawn so a
+// test's environment is only what that test asked for.
+//
+// This matters more here than in a typical suite, because the audience for
+// these variables is precisely the person running the suite with them set: the
+// docs added alongside this feature tell a second-lane user to export
+// CAREER_OPS_PIPELINE and CAREER_OPS_SCAN_HISTORY. With the parent environment
+// inherited wholesale, check 1 below - the one asserting DEFAULT behavior -
+// would follow that user's override and append fixture postings to their real
+// inbox, and the corresponding scan-history write would poison their dedup
+// source. A test suite must not be able to write into the data it is testing
+// the handling of (CodeRabbit, reviewing #2568).
+const SCANNER_PATH_VARS = [
+  'CAREER_OPS_PORTALS',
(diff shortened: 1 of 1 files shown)


santifer:
Thanks for this, @ringo380.

runScan spilled the entire process.env into every child, so check 1 (the one asserting the DEFAULT output paths) got redirected whenever whoever runs the suite has CAREER_OPS_PIPELINE exported. The person most likely to have it exported is exactly the user the docs for that feature tell to export it: fixture postings land in their real inbox and fixture rows in scan-history.tsv, which is the dedup source, so a real listing that collides gets silently skipped on a later scan. Clears the four variables before each check's overrides and adds a check 5 asserting the isolation, because on a clean machine removing the cleanup changes nothing observable and nobody would notice it disappear.

santifer:
Merged, @ringo380: a suite that passes or fails depending on what the person running it has in their own config is a suite that lies half the time, and this is the third fixture-isolation hole you have closed today. Thanks for treating the test harness as production code. 🚀

PR #2729, opened 2026-08-11
docs: describe the report as A-H and stop calling the scoring dimensions blocks

santifer:
`modes/_shared.md` introduced its table of scoring dimensions with "The evaluation uses 6 blocks (A-F)". The table under that line is not blocks: it is the five dimensions that feed the global score. Conflating the two is why "A-F" gets read as a grading scale instead of a report layout.

`modes/oferta.md` defines eight blocks, A through H, with H conditional on a score of 4.5 or above. `ARCHITECTURE.md` still said A-G.

Prose only, no behaviour change. Full suite: 3430 passed, 0 failed.

## Summary by CodeRabbit

* **Documentation**
  * Updated the evaluation architecture and report structure from A–F to A–H.
  * Clarified the holistic global 1–5 scoring model across five dimensions: CV match, North Star alignment, compensation, cultural signals, and red flags.
  * Documented Block G as a separate legitimacy assessment and added Block H for scores of 4.5 and above.


santifer:
Maintainer docs fix: the report has been A-H since the interview-prep block landed, and calling the scoring dimensions blocks collided with the report-block vocabulary in every explanation since. Nomenclature only; the frozen assertions stayed green throughout.

PR #2730, opened 2026-08-11
feat(verify-portals): --unmined finds companies stuck in search_queries

WJiangH:
Closes #2693.

## The blind spot

A `search_queries` entry only produces links for a human to open. It never feeds
the zero-LLM scanner. So a company named **only** inside a query string is
invisible to every scan — and nothing reports it: the entry validates,
`verify-portals` passes, the scan succeeds, and the company simply never appears.

Found in my own `portals.yml`:

```yaml
- name: Tesla / Groq / Cerebras — silicon
  query: '(site:tesla.com/careers OR site:groq.com/careers OR site:cerebras.ai/careers) ("RTL" OR ...)'
```

Cerebras has a live Ashby board with **123 postings**. Together AI (Greenhouse,
59) and Baseten (Ashby, 66) were the same story from other entries. ~250
postings never scanned, with no signal anywhere.

## `node verify-portals.mjs --unmined`

Extracts the companies a query names, drops the ones that are really search
vocabulary, probes what survives, reports which have a real board.

**Stopwords come from the user's own `title_filter` / `location_filter`**, not a
hardcoded list — retuning the search retunes the filter. Including individual
*words*, because queries quote fragments (`"Verification"`, `"AI"`) of keywords
(`Design Verification`, `AI Engineer`); whole-phrase matching alone reported
both as company names. A partial overlap survives: "Physical Intelligence" is
kept against a `Physical Design` filter because "intelligence" is the company's
own word.

**Probing stops at the first resolving slug per company.** The ATS directories
rate-limit hard enough that exhausting every variant would poison the run for
whatever is probed after it. Derived variants are still needed —
`together.ai`'s board is `togetherai`, so the bare host label misses it.

**Scope is stated, not implied.** Only Greenhouse/Ashby/Lever are probed, since
a slug is the whole address there. The output says *"no Greenhouse/Ashby/Lever
board"*, not *"no board"* — AMD and Apple do have boards, just not guessable
ones. This would have caught Cerebras but **not** Groq, who moved to Gem. That
limitation is in the module header, the `--help`, and the output itself.

`--strict` turns it into a CI gate.

## Testing

Rebased onto today's `main`; `node test-all.mjs` → **3432 passed, 0 failed**.

22 tests, network-free via injected `fetchJson`. Live run against a real
49-keyword config: 10 company names extracted with zero noise, one new finding
(`astrus` → lever, live but empty).

## Summary by CodeRabbit

* **New Features**
[...]

PR #2731, opened 2026-08-11
feat: audit what title_filter.positive is actually doing

### LINKED ISSUE #2694: No visibility into what title_filter.positive is actually doing
[`filter-audit`](https://github.com/WJiangH/career-ops/tree/filter-audit) (10 self-tests).

## The problem

The keyword layer decides the fate of most of what gets scanned — measured on
one run, **8,289 found and 6,499 rejected on title alone** — and reports
nothing about how it did that.

There is no way to tell which keywords earn their place, which are carried
entirely by neighbours, or which are quietly filling the pipeline with roles
that score 1.0/5. Tuning `title_filter.positive` is guesswork, and every
guess costs real evaluation tokens to disprove.

`scan.mjs` already computes the answer: `matchedTitleKeywords()` knows exactly
which keywords a title hit, then **discards it roughly eight thousand times per
scan** because only `content_filter` scoping needed it.

## Proposed shape

Replay that same function over `data/scan-history.tsv` and join the scores from
`reports/`. Read-only, zero-LLM, never proposes an edit.

The metric that matters is **`unique`**, not `hits`: how many offers a keyword
is the *only* one to catch. A keyword with 200 hits and 0 unique can be deleted
without losing a posting; one with 3 hits and 3 unique is the sole reason those
three were ever seen. Raw hit counts hide both cases.

## What it found on a real 49-keyword filter over 539 offers

- **8 keywords that have never matched anything** — including `RLHF` and
  `Model Serving`, lifted straight from the CV. They describe *work*, not job
  titles. Copying CV vocabulary into a title filter does not work, and this is
  the first evidence of why.
- **3 redundant** — every match also caught by a neighbour.
- **8 noisy** — `EDA` admits 12 offers averaging 1.0/5; `Agentic` 22 averaging
  1.4, while *also* being one of the larger unique contributors. So the trade
  is real rather than a free deletion, which is exactly the kind of thing a
  count alone cannot tell you.

Worth having upstream? Happy to open a PR or reshape it.

WJiangH:
Closes #2694.

## The gap

The keyword layer decides the fate of most of what gets scanned — measured on
one run, **8,289 found and 6,499 rejected on title alone** — and reports nothing
about how. No way to tell which keywords earn their place, which are carried
entirely by neighbours, or which quietly fill the pipeline with 1.0/5 roles.
Every tuning guess costs real evaluation tokens to disprove.

`scan.mjs` already computes the answer and throws it away: `matchedTitleKeywords()`
knows exactly which keywords a title hit, then discards it ~8,000 times per scan
because only `content_filter` scoping needed it.

## What it does

Replays that same function over `data/scan-history.tsv` and joins the scores
from `reports/`. Read-only, zero-LLM, **never proposes an edit** — it reports,
the user decides.

The metric is **`unique`**, not `hits`: how many offers a keyword is the *only*
one to catch. A keyword with 200 hits and 0 unique can be deleted without losing
a posting; one with 3 hits and 3 unique is the sole reason those three were seen.
Raw hit counts hide both cases.

## What it found on a real 49-keyword filter

- **8 keywords that have never matched anything** — including `RLHF` and
  `Model Serving`, lifted straight from the CV. They describe *work*, not job
  titles. Copying CV vocabulary into a title filter does not work, and this is
  the first evidence of why.
- **3 redundant** — every match also caught by a neighbour.
- **8 noisy** — `EDA` admits 12 offers averaging 1.0/5; `Agentic` 22 averaging
  1.4 while *also* being one of the larger unique contributors, so the trade is
  real rather than a free deletion. A count alone cannot tell you that.

Historical rows matching no current keyword are counted and surfaced rather than
silently dropped — that number means the filter changed since they were recorded.

## Testing

`--self-test` → 10 passed. Live `--summary` run included above.

One fixture note: an earlier cut of the self-test had my *expected* values wrong
(1.5 vs the actual 1.625 average, and a hit count under the ≥5 threshold), not
the code. The fixtures now include a row landing exactly on the noisy-rule
boundary so the threshold is pinned rather than approximated.

## Summary by CodeRabbit

* **New Features**
  * Added a read-only audit tool for reviewing portal title filters against scan history and evaluation scores.
  * Reports unmatched, redundant, noisy, and unused keywords, along with match counts, examples, and average scores.
  * Supports human-readable output, JSON export, help text, and self-test mode.

* **Chores**
  * Included the audit tool in system updates and documented main files.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -110,6 +110,7 @@ AI-powered, CLI-agnostic job search automation: pipeline tracking, offer evaluat
 | `salary-gap.mjs` | Desired/advertised/actual comp gap analyzer — folds report `advertised_comp` + `data/salary-observations.tsv` (JSON or `--summary`) |
 | `assessment-log.mjs` | Skills-assessment logger — `add` appends platform/subject/threshold/score + staleness note to `data/assessments.tsv` (JSON or `--summary`) |
 | `jd-skill-gap.mjs` | Zero-LLM JD skill classifier vs `cv.md`: existing / supportedByResume / gap; never auto-adds claims to `cv.md` (JSON or `--summary`) |
+| `filter-audit.mjs` | Zero-LLM, read-only audit of `portals.yml` `title_filter.positive`: replays `matchedTitleKeywords()` over `data/scan-history.tsv` to report which keywords never match, which are redundant (zero offers they alone catch), and which are noisy (high volume, low average score once evaluated). Ranks by `unique`, not hits (JSON or `--summary`) |
 | `contacts.mjs` | Job-search phonebook → vCard 3.0 exporter — stable UIDs so re-imports update instead of duplicating on platforms that honor vCard UID (JSON, `--summary`, `--vcf`, `--caller-id`) |
 | `data/contacts.tsv` | Job-search contact list — recruiters/hiring managers/peers saved from `contacto` (user layer, gitignored third-party PII) |
 | `outcome.mjs` | Record application outcome, archive artifacts, and sync tracker (`node outcome.mjs <selector> <type>`) |
--- a/filter-audit.mjs
+++ b/filter-audit.mjs
@@ -1,0 +1,217 @@
+#!/usr/bin/env node
+
+/**
+ * filter-audit.mjs — what your title_filter is actually doing.
+ *
+ * Zero-LLM, read-only. The keyword layer decides the fate of ~78% of everything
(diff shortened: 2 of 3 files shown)


WJiangH:

Nothing is lost by it:

- **The issue stays open.** `Closes #2694` only fires on merge, so the problem you triaged and accepted is still on record; only the queue position goes.

PR #2732, opened 2026-08-11
feat(scan): record the titles the filter rejects

### LINKED ISSUE #2695: scan discards ~78% of postings on title and keeps no record of what they were
[`scan-reject-log`](https://github.com/WJiangH/career-ops/tree/scan-reject-log)
(8 network-free tests). Related to #2694 — that one asks "which of my keywords
earn their place", this one asks "what am I missing", and the second question
cannot be answered from the data we currently keep.

## The problem

Every scan discards far more than it keeps — **6,499 of 8,289 postings on title
alone** in one measured run — and only the *count* survives, in
`scan-runs.tsv`'s `filtered_title`.

So the question that actually matters, *"is my keyword list missing
something?"*, has no evidence behind it and can only be answered by guessing
synonyms. Guessing demonstrably does not work: auditing one 49-keyword filter
found 8 keywords that had never matched anything, including `RLHF` and
`Model Serving` copied straight from the CV.

## Proposed shape

Record the rejected titles, so the near-miss set becomes **derivable** rather
than guessed:

```
(rejected titles) ∩ (cv.md vocabulary) − (current positives)
```

That updates itself when the CV changes and needs no keyword curation by hand.

Design points:

- **Deduped by title, capped at 5,000.** The analysis cares about the variety
  of language the market uses, not how often one posting reappears.
- **Merged with the existing log on write**, so a later narrow scan cannot
  erase vocabulary an earlier broad one saw.
- **Buffered in memory, flushed once at the end** — the hot loop never touches
  disk.
- **Records the fact, not the reason.** Which negative keyword rejected a title
  is recomputable offline from `portals.yml`; paying for it inside the loop,
  thousands of times per scan, buys nothing.

Two details I'd want a maintainer's opinion on:

1. The reverse scanner filters titles in **two** places — the pure
   `passesFilters()` helper and an inline check that keeps per-stage counters.
   Patching only the inline one left the log empty on the path the scanner
   actually takes. Is that duplication worth collapsing separately?
2. **Not guarded by `--dry-run`**, unlike the pipeline and scan-history writes.
   My reasoning: that rule exists so a dry run cannot add to your queue or
   claim it saved results, and a bounded diagnostic file under `data/` does
   neither — while the web Explorer *always* passes `--dry-run`, so guarding it
   would leave the log permanently empty on the main usage path. I can see an
   argument the other way and would rather agree up front than in review.

WJiangH:
Closes #2695.

## The gap

Every scan discards far more than it keeps — **6,499 of 8,289 postings on title
alone** in one measured run — and only the *count* survives, in `scan-runs.tsv`'s
`filtered_title`.

So "is my keyword list missing something?" has no evidence behind it and can only
be answered by guessing synonyms. Guessing demonstrably does not work: auditing
one 49-keyword filter (#2694) found 8 keywords that had never matched anything.

## What it does

Records the rejected titles to `data/scan-rejects.tsv`, so the near-miss set
becomes **derivable** rather than guessed:

```
(rejected titles) ∩ (cv.md vocabulary) − (current positives)
```

That updates itself when the CV changes and needs no keyword curation by hand.

- **Deduped by title, capped at 5,000.** The analysis cares about the variety of
  language the market uses, not how often one posting reappears.
- **Merged with the existing log on write**, so a later narrow scan cannot erase
  vocabulary an earlier broad one saw.
- **Buffered in memory, flushed once at the end** — the hot loop never touches
  disk. A write failure is swallowed: a scan is never lost over its telemetry.
- **Records the fact, not the reason.** Which negative keyword rejected a title
  is recomputable offline from `portals.yml`; paying for it inside the loop,
  thousands of times per scan, buys nothing.

## Two calls I'd rather agree on than defend in review

**1. The reverse scanner filters titles in two places** — the pure
`passesFilters()` helper and an inline check that keeps per-stage counters.
Patching only the inline one left the log empty on the path the scanner actually
takes, and rate limiting hid it. Both are patched here. Whether that duplication
should be collapsed is a separate change; happy to do it if you want it.

**2. Not guarded by `--dry-run`,** unlike the pipeline and scan-history writes.
My reasoning: that rule exists so a dry run cannot add to your queue or claim it
saved results, and a bounded diagnostic file under `data/` does neither — while
the web Explorer *always* passes `--dry-run`, so guarding it would leave the log
permanently empty on the main usage path. I can see the argument the other way
and will change it on request.

## Testing

8 tests, network-free by construction. The live scanners were useless for
verifying this — the ATS directories rate-limit hard enough that a run retrieves
nothing at all, and an empty log is indistinguishable from a broken recorder, so
the tests call the filter path directly. Column shearing from tabs/newlines in a
scraped title is covered, as is the two-call-site trap above.

## Summary by CodeRabbit

* **New Features**
  * Added persistent tracking for titles rejected by scan filters, including title, company, source, and repeat counts.
  * Rejection records are deduplicated, sorted, limited in size, and retained across scans.

* **Bug Fixes**
  * Rejection data is saved even when a scan encounters an error.
  * Scan failures now preserve their original error message while completing cleanup reliably.

* **Tests**
  * Added coverage for rejection tracking, persistence, formatting, failure handling, and filter integration.

### DIFF
--- a/scan-ats-full.mjs
+++ b/scan-ats-full.mjs
@@ -43,7 +43,7 @@ import lever from './providers/lever.mjs';
 import ashby from './providers/ashby.mjs';
 import workday from './providers/workday.mjs';
 import icims from './providers/icims.mjs';
-import { buildTitleFilter, buildLocationFilter, buildContentFilter, matchedTitleKeywords, loadSeenUrls, normalizeUrlForDedup, appendToPipeline, appendToScanHistory, loadBlacklist, parseSinceDays } from './scan.mjs';
+import { buildTitleFilter, buildLocationFilter, buildContentFilter, matchedTitleKeywords, recordTitleReject, flushTitleRejects, loadSeenUrls, normalizeUrlForDedup, appendToPipeline, appendToScanHistory, loadBlacklist, parseSinceDays } from './scan.mjs';
 import { SEED_SOURCES, toPortalEntry } from './seeds/vc-portfolios.mjs';
 import { normalizeCompany } from './tracker-utils.mjs';
 
@@ -397,7 +397,14 @@ export function filterBlacklistedOffers(offers, blacklist, { includeBlacklisted 
 // (#1846) unit-testable without mocking providers or duplicating the rule
 // order in two places for the caller that doesn't need per-stage counts.
 export function passesFilters(job, { titleFilter, locationFilter, contentFilter, titleFilterConfig }) {
-  if (!titleFilter(job.title)) return false;
+  if (!titleFilter(job.title)) {
+    // Both title-reject sites record: this pure helper serves the legacy path
+    // while the loop below has its own inline check for the per-stage counters.
+    // Patching only one left the log empty on the path the scanner actually
+    // takes, which is exactly the kind of silence this log exists to end.
+    recordTitleReject(job.title, job.company, job.source);
+    return false;
+  }
   // job.url is passed so the location filter can fall back to the URL's own
   // location segment when the provider reports a rolled-up "N Locations" string;
   // job.title so a title-stated remote role survives a city-only location.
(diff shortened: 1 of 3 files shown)


WJiangH:

Nothing is lost by it:

- **The issue stays open.** `Closes #2695` only fires on merge, so the problem you triaged and accepted is still on record; only the queue position goes.

PR #2733, opened 2026-08-11
feat: audit cv.md for skills it claims but never demonstrates

### LINKED ISSUE #2696: Nothing audits cv.md for skills it lists but never evidences
[`cv-evidence-gap`](https://github.com/WJiangH/career-ops/tree/cv-evidence-gap)
(13 self-tests).

## The gap between the two existing tools

| Tool | Question it answers |
|---|---|
| `upskill.mjs` | What should I learn? — **excludes anything already in `cv.md`** |
| `jd-skill-gap.mjs` | Does one JD's ask appear in `cv.md` at all? |
| *missing* | What does `cv.md` **claim** but never back up? |

That third question has no owner, and it turns out to be where the real
feedback lives.

## Why it matters

Across 19 evaluations of one CV, **nine reports raised C++** — not as missing,
but as *"listed as a skill and never evidenced in a bullet"*.

`upskill.mjs` excluded it for exactly the reason it mattered: it **is** in the
Skills section. So the most repeated objection in the whole pipeline was
invisible to the tooling — while the candidate's own record supported the
claim (the role in question is llama.cpp deployment and kernel-level tuning,
which is C++ work that no bullet says is C++).

A bare skills list is the weakest claim a CV can make. A reviewer discounts it;
so does an evaluator.

## Proposed shape

Zero-LLM. Split `cv.md` at the Skills heading — claims on one side, the prose
that could demonstrate them on the other — and report the terms that never
cross over, ranked by how often `reports/` actually objected. **Writes
nothing**: only the candidate knows which claims are real and where the
evidence lives.

Two things my first cut got wrong, both worth flagging because they'd bite
anyone implementing this:

- **Objections must be read from the Machine Summary `hard_stops` /
  `soft_gaps`, not from report prose.** Scanning bodies for the term near a
  negation word counted a comparison row reading *"Python: strong match. C++:
  evidenced at the llama.cpp scope"* as an objection to all four languages it
  named. C++ went from 17 phantom hits to 3 real ones once this changed.
- **An all-words match on multi-word skills false-positives.** "Latency
  Profiling" and "Cloud Deployment" matched prose that contained both words
  separately. Needed a longest-word fallback.

Worth having upstream? Happy to open a PR or reshape it first.

WJiangH:
Closes #2696.

## The question nobody owns

| Tool | Question |
|---|---|
| `upskill.mjs` | What should I learn? — **excludes anything already in `cv.md`** |
| `jd-skill-gap.mjs` | Does one JD's ask appear in `cv.md` at all? |
| *this* | What does `cv.md` **claim** but never back up? |

## Why it matters

Across 19 evaluations of one CV, **nine reports raised C++** — not as missing,
but as *"listed as a skill and never evidenced in a bullet"*.

`upskill.mjs` excluded it for exactly the reason it mattered: it **is** in the
Skills section. So the most repeated objection in the whole pipeline was
invisible to the tooling — while the candidate's own record supported the claim
(the role in question is llama.cpp deployment and kernel-level tuning, which is
C++ work that no bullet says is C++).

A bare skills list is the weakest claim a CV can make. A reviewer discounts it;
so does an evaluator.

## What it does

Zero-LLM. Splits `cv.md` at the Skills heading — claims on one side, the prose
that could demonstrate them on the other — and reports the terms that never
cross over, ranked by how often `reports/` actually objected.

**Writes nothing.** Only the candidate knows which claims are real and where the
evidence lives.

## Two traps, both hit before they were fixed

- **Objections must come from the Machine Summary `hard_stops` / `soft_gaps`,
  not from report prose.** Scanning bodies for the term near a negation word
  counted a comparison row reading *"Python: strong match. C++: evidenced at the
  llama.cpp scope"* as an objection to all four languages it named. C++ went
  from 17 phantom hits to 3 real ones.
- **An all-words match on multi-word skills false-positives.** "Latency
  Profiling" and "Cloud Deployment" matched prose containing both words
  separately. Needed a longest-word fallback.

Both are covered by tests, since either would silently inflate the output rather
than fail.

## Testing

`--self-test` → 13 passed. Live run on a real CV: 32 skills listed, 20 never
demonstrated, C++ top of the costing-you-now list at 4 reports.

## Summary by CodeRabbit

* **New Features**
  * Added a command-line audit that identifies résumé skills lacking supporting evidence.
  * Ranks evidence gaps based on recurring feedback from evaluation reports.
  * Supports human-readable and JSON output, including graceful handling of missing files or reports.
  * Runs without AI services and does not modify files.

* **Documentation**
  * Added the new audit tool to the documented list of main files.

* **Chores**
  * Included the audit tool in automatic system-file updates.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -110,6 +110,7 @@ AI-powered, CLI-agnostic job search automation: pipeline tracking, offer evaluat
 | `salary-gap.mjs` | Desired/advertised/actual comp gap analyzer — folds report `advertised_comp` + `data/salary-observations.tsv` (JSON or `--summary`) |
 | `assessment-log.mjs` | Skills-assessment logger — `add` appends platform/subject/threshold/score + staleness note to `data/assessments.tsv` (JSON or `--summary`) |
 | `jd-skill-gap.mjs` | Zero-LLM JD skill classifier vs `cv.md`: existing / supportedByResume / gap; never auto-adds claims to `cv.md` (JSON or `--summary`) |
+| `cv-evidence-gap.mjs` | Zero-LLM audit of `cv.md` against itself: skills named in the Skills section that no bullet ever demonstrates, ranked by how often `reports/` Machine Summaries actually raised them as a gap. Complements `upskill.mjs` (what to learn — excludes anything in cv.md) and `jd-skill-gap.mjs` (one JD vs cv.md). Never writes (JSON or `--summary`) |
 | `contacts.mjs` | Job-search phonebook → vCard 3.0 exporter — stable UIDs so re-imports update instead of duplicating on platforms that honor vCard UID (JSON, `--summary`, `--vcf`, `--caller-id`) |
 | `data/contacts.tsv` | Job-search contact list — recruiters/hiring managers/peers saved from `contacto` (user layer, gitignored third-party PII) |
 | `outcome.mjs` | Record application outcome, archive artifacts, and sync tracker (`node outcome.mjs <selector> <type>`) |
--- a/cv-evidence-gap.mjs
+++ b/cv-evidence-gap.mjs
@@ -1,0 +1,270 @@
+#!/usr/bin/env node
+
+/**
+ * cv-evidence-gap.mjs — skills your CV claims but never demonstrates.
+ *
+ * Zero-LLM. Complements the two tools that already exist by answering the
+ * question neither of them asks:
+ *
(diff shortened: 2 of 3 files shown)


WJiangH:

Nothing is lost by it:

- **The issue stays open.** `Closes #2696` only fires on merge, so the problem you triaged and accepted is still on record; only the queue position goes.

PR #2736, opened 2026-08-12
fix(set-status): the --role guard dropped every non-Latin letter, so it wrote to unmatched rows (#2670)

### LINKED ISSUE #2670: set-status --role guard collapses all non-Latin titles to "", defeating #2009 role-mismatch protection
### Code of Conduct

### Existing issues

### What happened?

`set-status.mjs`'s #2009 `--role` guard uses an ASCII-only normalizer:

```js
const normalizeRoleText = s => String(s ?? '')
  .toLowerCase()
  .replace(/\+\+/g, ' plusplus ')
  .replace(/#/g, ' sharp ')
  .replace(/[^a-z0-9]+/g, ' ')   // deletes every non-Latin letter
  .trim();
```

Any pure CJK/non-Latin role keys to `""`. Distinct titles then compare equal (`"" === ""`), so the guard that exists to prevent silent wrong-row status writes **accepts the mismatch and writes**.

Same class as #2369 / #2569 / #2666: a private `[a-z0-9]` strip on a matching key. Company matching already uses Unicode-aware `normalizeTextKey`; the role guard was only hardened for Latin symbols (C#/C++).

## Verified counterexample

| Target role | `--role` | `normalizeRoleText` | Guard today | Correct |
|---|---|---|---|---|
| `バックエンドエンジニア` | `フロントエンドエンジニア` | `""` / `""` | **ACCEPT** | REJECT |
| `エンジニア（東京）` | `エンジニア（大阪）` | `""` / `""` | **ACCEPT** | REJECT |
| `Platform Engineer` | `Data Engineer` | distinct | REJECT | REJECT |
| `C++ Engineer` | `C# Engineer` | symbol-preserved | REJECT | REJECT |

## Test gap

`set-status-tests.mjs` §1c–1e covers Latin + C#/C++ only. Zero CJK/Unicode cases.

## Fix

Import `normalizeTextKey` from `tracker-parse.mjs` and keep the C#/C++ symbol pre-map, but stop using `[^a-z0-9]`:

```js
import { extractTrackerReportNumbers, resolveColumns, parseTrackerRow, normalizeTextKey } from './tracker-parse.mjs';

const normalizeRoleText = s => normalizeTextKey(
  String(s ?? '')
    // Preserve symbols that distinguish real titles before collapsing generic
    // punctuation — otherwise "C# Engineer" and "C++ Engineer" both fold to
    // "c engineer" and the exact-equality path treats them as the same row.
    .replace(/\+\+/g, ' plusplus ')
    .replace(/#/g, ' sharp '),
  ' ',
);
```

| Input | Before (`[^a-z0-9]`) | After (`normalizeTextKey`) |
|---|---|---|
| `バックエンドエンジニア` | `""` | `バックエンドエンジニア` |
| `フロントエンドエンジニア` | `""` | `フロントエンドエンジニア` |
| `C++ Engineer` | `c plusplus engineer` | `c plusplus engineer` |
| `C# Engineer` | `c sharp engineer` | `c sharp engineer` |

Add a sandbox test in `set-status-tests.mjs` mirroring §1c:

- Tracker: リクルート / `バックエンドエンジニア`
- `node set-status.mjs リクルート Rejected --role フロントエンドエンジニア --json`
- expect: exit `3`, `code: "role-mismatch"`, tracker unchanged
- same-title path `--role バックエンドエンジニア` → exit `0`

## Related

- #2009 (role-mismatch guard this defeats)
- #2569 / #2666 (same ASCII-strip class on matching keys)

### Steps to reproduce

Tracker integrity — wrong application row can have its status mutated when `--role` is supplied for a different non-Latin title at the same company.

```bash
# Tracker: one リクルート row, role バックエンドエンジニア, status Interview
node set-status.mjs リクルート Rejected --role "フロントエンドエンジニア" --json
```

### Expected behavior

**Expected** (same as Latin #2009): exit `3`, `code: "role-mismatch"`, tracker unchanged.

### CLI tool

Cursor

### OS and Node.js version

macOS 26.5.2 / Node 26.5.0

### Output of npm run doctor

```shell
pravitampapathini@wifi-10-43-175-99 career-ops % npm run doctor                 

> career-ops@1.26.0 doctor
> node doctor.mjs


career-ops doctor
================

✓ Node.js >= 22.5 (v26.5.0)
✓ Billing source: no API key in the environment (a Claude subscription will be used if you are logged in)
✓ Dependencies installed
  → No project-level MCP config was detected for claude.
  → Add the Playwright MCP server to your claude config.
⚠ cv.md not found (user setup required)
  → Create cv.md in the project root with your CV in markdown
  → See examples/ for reference CVs
⚠ config/profile.yml not found (user setup required)
  → Run: cp config/profile.example.yml config/profile.yml
  → Then edit it with your details
⚠ modes/_profile.md not found (user setup required)
  → Run: cp modes/_profile.template.md modes/_profile.md
  → Then customize your archetypes / targeting narrative
⚠ portals.yml not found (user setup required)
  → Run: cp templates/portals.example.yml portals.yml
  → Then customize with your target companies
✓ Fonts directory ready
✓ data/ directory ready
✓ data/pipeline.md ready
✓ output/ directory ready
✓ reports/ directory ready

Result: All checks passed (5 warnings — see above). You're ready to go! Run `claude` (or `opencode`) to start.

Join the community: https://discord.gg/8pRpHETxa4
Read the manifesto: `npm run manifesto` — a new way of job searching is taking shape, and you are now part of it.
```

Zoubeir23:
Closes #2670.

The `--role` guard exists to stop a status reaching the wrong tracker row (#2009). Its equality normalizer collapsed with `[^a-z0-9]+`, which drops every **letter** outside the Latin range, not just punctuation. Any title written entirely in a non-Latin script keys to `""`, two different titles compare equal, and the guard accepts the mismatch and writes.

## Reproduced before touching the code

```
"バックエンドエンジニア"  → ""
"フロントエンドエンジニア" → ""
"エンジニア（東京）"      → ""
"مهندس برمجيات"          → ""
"Разработчик"            → ""
"Platform Engineer"      → "platform engineer"
```

Driven through the real CLI on a sandbox tracker, asking to update a `バックエンドエンジニア` row while passing `--role フロントエンドエンジニア`:

```json
{"changed":true,"num":1,"company":"リクルート","role":"バックエンドエンジニア",
 "oldStatus":"Evaluated","newStatus":"Rejected","statusLogged":true}
```

Exit `0`. The guard did not merely fail to reject — it wrote the status, and logged the transition, against a row whose role it had never matched.

## Wider than the issue states

#2670 frames this as CJK. It is not: `[^a-z0-9]` deletes Arabic and Cyrillic letters just as completely, and the repo ships `modes/ar/` and `modes/ru/` alongside `modes/ja/`. These are shipped-market titles. The PR covers all three.

## The fix

`normalizeRoleText` now folds through `normalizeTextKey` from `tracker-parse.mjs` — the Unicode-aware normalizer **company** matching has used all along, so this removes an inconsistency rather than inventing a rule. The C#/C++ symbol pre-map is kept ahead of it and unchanged.

`normalizeTextKey` also applies NFKC, so a decomposed title still matches its composed row — a property the tests pin explicitly.

| Input | Before | After |
|---|---|---|
| `バックエンドエンジニア` | `""` | `バックエンドエンジニア` |
| `フロントエンドエンジニア` | `""` | `フロントエンドエンジニア` |
| `C++ Engineer` | `c plusplus engineer` | `c plusplus engineer` |
| `C# Engineer` | `c sharp engineer` | `c sharp engineer` |

## Test plan

`set-status-tests.mjs` §1c–1e covered Latin and C#/C++ only. New §1f and §1g, mirroring §1c's shape and driving the real CLI against a sandbox tracker:

- a different Japanese title is rejected, tracker byte-identical afterwards;
- two Japanese titles differing only by a parenthesised city are rejected — the fullwidth parens still collapse as punctuation, so the city is what separates them;
- the guard does not over-fire: the same Japanese title still proceeds;
- an **NFD-decomposed** Japanese title still matches its NFC row. The fixture asserts it is genuinely decomposed first, so the case cannot pass vacuously if the folding is removed;
- a different Cyrillic title is rejected.

- [x] The three rejection cases verified red before the fix (they exited `0` with `changed:true`); the two "still proceeds" cases pass before and after, which is what makes them guards.
- [x] `node set-status-tests.mjs` — 80 passed, 0 failed.
- [x] `node test-all.mjs` — 3430 passed, 0 failed (the single warning is the pre-existing `Dashboard build skipped — go compiler not in env`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved role matching for Japanese, Arabic, Cyrillic, and other Unicode titles.
  * Correctly handles equivalent Unicode text, including decomposed Japanese input.
  * Preserves distinctions between roles that differ by city or punctuation.
  * Keeps `C#` and `C++` titles distinct during matching.
  * Prevents mismatched titles from modifying tracker status.
* **Tests**
  * Added regression coverage for Unicode matching and safe failure behavior.

### DIFF
--- a/set-status-tests.mjs
+++ b/set-status-tests.mjs
@@ -233,6 +233,91 @@ const TRACKER_REPORT_MISMATCH = `# Applications Tracker
   rmSync(sb.dir, { recursive: true, force: true });
 }
 
+// ── 1f. Non-Latin titles must not all collapse to the same role (#2670) ─
+// The equality normalizer stripped `[^a-z0-9]`, which deletes every non-Latin
+// LETTER, not just punctuation. Any title written entirely in Japanese, Arabic
+// or Cyrillic keyed to "", two different titles compared equal ("" === ""), and
+// the guard that exists to stop a status reaching the wrong row accepted the
+// mismatch and wrote. The repo ships modes/ja/, modes/ar/ and modes/ru/, so
+// these are shipped-market titles, not exotica. Company matching already used
+// the Unicode-aware normalizeTextKey; only the role guard was left ASCII-only.
+{
+  const TRACKER_CJK = `# Applications Tracker
+
+| # | Date | Company | Role | Score | Status | PDF | Report | Notes |
+|---|------|---------|------|-------|--------|-----|--------|-------|
+| 1 | 2026-06-01 | リクルート | バックエンドエンジニア | 4.0/5 | Evaluated | ✅ | [1](../reports/001-recruit-2026-06-01.md) | — |
+`;
+  const sb = makeSandbox(TRACKER_CJK);
+  const before = readTracker(sb);
+  const r = runSetStatus(['リクルート', 'Rejected', '--role', 'フロントエンドエンジニア', '--json'], sb);
+  let parsed = null;
+  try { parsed = JSON.parse(r.stdout); } catch {}
+  if (r.code === 3 && parsed?.code === 'role-mismatch' && readTracker(sb) === before) {
+    pass('role-mismatch: a different Japanese title does not match (#2670)');
+  } else {
+    fail(`role-mismatch cjk: code=${r.code} json=${JSON.stringify(parsed)}\n${r.stdout}${r.stderr}`);
+  }
+
+  // Two titles differing ONLY by a parenthesised city — the fullwidth
+  // parentheses are punctuation and must still collapse, so what has to
+  // distinguish them is the city itself.
+  const sb2 = makeSandbox(`# Applications Tracker
+
+| # | Date | Company | Role | Score | Status | PDF | Report | Notes |
+|---|------|---------|------|-------|--------|-----|--------|-------|
+| 1 | 2026-06-01 | リクルート | エンジニア（東京） | 4.0/5 | Evaluated | ✅ | [1](../reports/001-recruit-2026-06-01.md) | — |
+`);
+  const before2 = readTracker(sb2);
+  const r2 = runSetStatus(['リクルート', 'Rejected', '--role', 'エンジニア（大阪）', '--json'], sb2);
+  if (r2.code === 3 && readTracker(sb2) === before2) {
+    pass('role-mismatch: same Japanese title, different city does not match (#2670)');
+  } else {
+    fail(`role-mismatch cjk-city: code=${r2.code}\n${r2.stdout}${r2.stderr}`);
+  }
+
+  // The guard must not over-fire: the SAME non-Latin title still proceeds,
+  // including across a Unicode normalization form (NFD input vs NFC row).
+  const r3 = runSetStatus(['リクルート', 'Applied', '--role', 'バックエンドエンジニア'], sb);
+  if (r3.code === 0 && /\| 1 \|[^\n]*\| Applied \|/.test(readTracker(sb))) {
+    pass('role-mismatch: the same Japanese title still proceeds (#2670)');
+  } else {
+    fail(`role-mismatch cjk-equal: code=${r3.code}\n${r3.stdout}${r3.stderr}`);
+  }
+  const nfdRole = "バックエンドエンジニア".normalize('NFD');
+  if (nfdRole === "バックエンドエンジニア") {
+    fail("role-mismatch cjk-nfd: fixture is not actually decomposed — the case would prove nothing");
+  }
+  const r4 = runSetStatus(["リクルート", 'Responded', '--role', nfdRole], sb);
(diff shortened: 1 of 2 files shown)


santifer:
Merged, @Zoubeir23. This is the good kind of fix: the guard that exists to stop a write landing on the wrong row was accepting every mismatch between two fully non-Latin titles, because both folded to the empty string and "" === "". Delegating to normalizeTextKey rather than writing a third normalizer means it now inherits whatever the shared one does, including the NFKC folding. 🚀

PR #2737, opened 2026-08-12
fix(classify-tier): an incidental senior word outranked an explicit intern marker, so scan silently dropped internships

Zoubeir23:
No issue — found while auditing `classify-tier.mjs`, one of the core scripts with no test file. `CONTRIBUTING.md:22` welcomes a direct PR for a bug fix.

## The defect

`classifyTier` resolved conflicts by **seniority rank**: `senior(4) > mid(3) > entry(2) > intern(1)`, highest wins. So a single senior-sounding word anywhere in a title outranked an explicit programme marker.

Real titles break that, because that word is usually naming the team, office or person the role sits beside — not the role's own level. Measured on `main`:

| Title | `main` | Should be |
|---|---|---|
| `Intern - Office of the Chief Technology Officer` | **senior** | intern |
| `Summer Intern, Director of Product` | **senior** | intern |
| `Software Engineering Intern, Lead Generation Platform` | **senior** | intern |
| `Marketing Intern - Senior Living Community` | **senior** | intern |
| `Graduate Trainee - Head of Retail Division` | **senior** | intern |
| `Junior Developer - Team Lead Support` | **senior** | entry |
| `Junior Staff Accountant` | **senior** | entry |
| `Engineer II - Senior Platform Group` | **senior** | mid |

Every title carrying both an explicit junior/intern marker and any incidental senior word came out `senior`.

## Why it costs something

`scan.mjs:2396` drops a posting outright when its tier is in `skip_tiers`:

```js
if (classifyTier && skipTiers.includes(classifyTier(job.title))) {
  totalFilteredTier++;
  continue;
}
```

The counter is aggregate — no log line names the dropped posting. So the failure is silent, and it lands on the wrong side: a **junior candidate** setting `skip_tiers: [senior]` loses exactly the internships they are scanning for, and never learns they existed.

The reverse direction is only noise: `templates/portals.example.yml:266` suggests `skip_tiers: [intern, entry]` for senior candidates, who then keep seeing internships they asked to hide.

## The fix

**Position decides, not seniority rank.** English job titles put the level first, so the leftmost marker is the role's own.

That single rule also preserves the case rank ordering got right: `Senior Intern Coordinator` — a role that manages interns — stays `senior`, because there the senior word genuinely leads the title. Same for `Director of Engineering, Junior Talent`.

Weight survives only as the tie-break for two markers at the same offset, which keeps the longer pattern of an overlapping pair (`mid-level` over `mid`).

One subtlety: the graduate matcher is a compound condition (`graduate` AND `program|scheme`), so it needs its own position accessor and must still be `test`ed before its index is taken — otherwise a bare `Graduate Engineer` would match on the position of `graduate` alone. It carries a `search` returning the offset of the level word rather than of the qualifier.

## Test plan

`classify-tier.mjs` ships 15 inline cases behind `--test` but has no file under `tests/`. This adds `tests/classify-tier-position.test.mjs` (24 cases), picked up by `test-all.mjs` auto-discovery.

Every relaxation is paired with a case asserting the senior classification still fires when it should, and the unambiguous titles (`Software Engineer I/II`, `Senior Software Engineer`, `A.I. Researcher`, `I.T. Specialist II`) are pinned so the acronym preprocessing and the roman-numeral matchers are proven untouched.

- [x] 8 of the 24 verified red on `main`; the 4 guard cases pass before and after, which is what makes them guards.
- [x] `node classify-tier.mjs --test` — all 15 inline cases still pass, unchanged.
- [x] `node test-all.mjs` — 3454 passed, 0 failed (the single warning is the pre-existing `Dashboard build skipped — go compiler not in env`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved role tier classification by prioritizing the earliest matching title marker.
  * Refined seniority handling for associate-qualified senior roles and program-bridge titles.
  * Improved recognition of numbered levels, acronyms, graduate programs, and programme-director roles.
  * Preserved accurate handling of compound graduate-program and scheme titles.

* **Tests**
  * Added coverage for title ordering, seniority terms, numbered levels, acronym processing, graduate-program distinctions, and non-text fallback behavior.

### DIFF
--- a/classify-tier.mjs
+++ b/classify-tier.mjs
@@ -79,20 +79,45 @@ export function classifyTier(title) {
     { pattern: /\bco-op\b/i, tier: 'intern', weight: 1 },
     {
       pattern: {
-        test: (t) => /\bgraduate\b/i.test(t) && /\b(program|scheme)\b/i.test(t)
+        test: (t) => /\bgraduate\b/i.test(t) && /\b(program|scheme)\b/i.test(t),
+        // Position of the level word itself, not of the "program" qualifier.
+        search: (t) => t.search(/\bgraduate\b/i)
       },
       tier: 'intern',
       weight: 1
     }
   ];
 
+  // POSITION decides, not seniority rank. Ranking by weight meant any senior
+  // word anywhere outranked an explicit programme marker, and in real titles
+  // that word is usually naming the team, office or person the role sits beside
+  // — "Summer Intern, Director of Product" is an internship. English job titles
+  // put the level first, so the LEFTMOST marker is the role's own level. That
+  // still reads "Senior Intern Coordinator" as senior: there the senior word
+  // genuinely leads the title.
+  //
+  // This is not cosmetic: scan.mjs drops a posting whose tier is in
+  // `skip_tiers` without naming it, so a junior candidate skipping `senior`
+  // silently lost the internships they were scanning for.
+  //
+  // Weight survives only as the tie-break for two markers at the same offset,
+  // which keeps the longer, more specific pattern of an overlapping pair
+  // (`mid-level` over `mid`, `entry-level` over `entry`).
   let bestMatch = null;
+  let bestIndex = Infinity;
 
   for (const matcher of matchers) {
-    if (matcher.pattern.test(cleanTitle)) {
-      if (!bestMatch || matcher.weight > bestMatch.weight) {
-        bestMatch = matcher;
-      }
+    // Match first: the graduate matcher is a COMPOUND condition (graduate AND
+    // program/scheme), so its position alone would fire on a bare "Graduate
+    // Engineer" that the condition itself rejects.
+    if (!matcher.pattern.test(cleanTitle)) continue;
+    const index = typeof matcher.pattern.search === 'function'
+      ? matcher.pattern.search(cleanTitle)
+      : cleanTitle.search(matcher.pattern);
+    if (index < 0) continue;
+    if (index < bestIndex || (index === bestIndex && matcher.weight > bestMatch.weight)) {
+      bestMatch = matcher;
+      bestIndex = index;
     }
   }
 
--- a/classify-tier.mjs
+++ b/classify-tier.mjs
@@ -81,7 +81,13 @@ export function classifyTier(title) {
       pattern: {
         test: (t) => /\bgraduate\b/i.test(t) && /\b(program|scheme)\b/i.test(t),
         // Position of the level word itself, not of the "program" qualifier.
-        search: (t) => t.search(/\bgraduate\b/i)
+        // Deliberately NOT named `search`: overloading a String.prototype method
+        // name on a matcher object makes `pattern.search(title)` read as the
+        // built-in, which coerces its argument to a RegExp — CodeQL flagged it
+        // as regex injection on the CLI's argv-derived title. The regex here is
+        // a literal and nothing is compiled from input, but the name was the
+        // problem, for a reader as much as for the analyzer.
+        levelWordIndex: (t) => t.search(/\bgraduate\b/i)
       },
       tier: 'intern',
       weight: 1
@@ -111,8 +117,8 @@ export function classifyTier(title) {
     // program/scheme), so its position alone would fire on a bare "Graduate
     // Engineer" that the condition itself rejects.
     if (!matcher.pattern.test(cleanTitle)) continue;
-    const index = typeof matcher.pattern.search === 'function'
-      ? matcher.pattern.search(cleanTitle)
+    const index = typeof matcher.pattern.levelWordIndex === 'function'
+      ? matcher.pattern.levelWordIndex(cleanTitle)
       : cleanTitle.search(matcher.pattern);
     if (index < 0) continue;
     if (index < bestIndex || (index === bestIndex && matcher.weight > bestMatch.weight)) {
--- a/classify-tier.mjs
+++ b/classify-tier.mjs
@@ -94,6 +94,33 @@ export function classifyTier(title) {
     }
   ];
 
+  // Guard (a): "Associate [*] Director/VP/Chief/Principal/Partner/Head-of" resolves
+  // to senior. The `associate` prefix qualifies the seniority band of a senior role;
+  // it does not demote it to entry-level. Works for both a direct compound
+  // ("Associate Director") and a broad one ("Associate Creative Director"). Checked
+  // before the position loop because `associate` always precedes the senior noun,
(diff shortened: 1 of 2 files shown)


santifer:
The claim reproduces exactly as you describe, @Zoubeir23: all 8 titles from your table return `senior` on current `main` and the correct tier with your patch. Position-over-rank is a defensible rule and the 24 new cases are welcome.

Blocking on a regression the fixture cannot see, which is the uncomfortable part: **the repo's own corpus does not exercise the path you changed.** Out of 305 title-shaped strings collected from `tests/`, `examples/`, `docs/`, `evals/`, `templates/` and `modes/` (excluding your own test file, to avoid circular reasoning): 164 have no marker, 126 have one, 11 have 2+ of the same tier, and **0 have markers of different tiers**. So "no fixture degradation" here is not evidence of safety, it is evidence of absence.

Outside the fixture, one family regresses deterministically, 15 of 15:

- `Associate Director, Data Science` → **entry** · `Associate Vice President, Technology` → entry · `Associate Principal Scientist` → entry · `Associate Chief Nursing Officer` → entry · `Associate Head of Product` → entry · `Associate Creative Director` → entry · `Associate Director` (from your own CLI examples) → entry
- `Intern Program Director` → **intern** · `Internship Program Director` → intern · `Graduate Program Director` → intern · `Graduate Scheme Lead` → intern · `Trainee Program Director` → intern
- `Junior Talent Director` → entry · `Junior Talent Lead` → entry · `Entry-Level Program Director` → entry

The damage mechanism is identical to the one you are fixing: `scan.mjs:2407` discards the posting without naming it, and `templates/portals.example.yml:266` literally suggests `skip_tiers: [intern, entry]` for a senior candidate. So a senior candidate silently stops seeing Associate Director and AVP roles. The patch moves which side the failure falls on rather than removing it.

A contrast that isolates the cause: `Deputy Director` and `Assistant Director` still resolve to senior, because "deputy" and "assistant" are not markers. Only qualifiers that also appear in the matcher table break (associate, junior, intern, trainee, graduate, entry-level). You already recognize this class (you keep `Senior Intern Coordinator` and `Director of Engineering, Junior Talent` as senior) but only in the word order where the senior term comes first. In the reverse order, which is the natural English order for these compounds, it fails and nothing covers it.

**The ask, scoped so this does not turn into a redesign** - a compound guard applied *before* the position rule, plus tests:

- (a) `\bassociate\s+(director|vice\s+president|vp|principal|partner|chief|head\s+of)\b` → senior
- (b) an intern/entry marker followed by `program|scheme|talent|cohort` must not win when a senior marker appears later (`Intern Program Director`, `Graduate Scheme Lead`)

One warning for the implementation: **a generic adjacency rule will not work** ("if the left marker is followed by a heavier one, the heavier wins") because it breaks `Junior Staff Accountant` → entry, which is a case you correctly want to keep. It has to be a compound list, not adjacency.

Separately: the welcome bot greeted you as a first-time contributor on this PR. That was a bug on our side, with your 25 merges sitting right there. Fixed in main today.

Zoubeir23:
@santifer Two guards added, pushed.

**Guard (a) — associate + senior-role noun**

Slices the title at `associate`, tests the remainder against your six nouns (`director`, `vice president`, `vp`, `principal`, `partner`, `chief`, `head of`). Handles both direct compounds (`Associate Director`) and broad ones (`Associate Creative Director`) without enumerating every possible adjective between the two words. `Associate Software Engineer` / `Associate Developer` miss the noun set and fall through to the position loop unchanged → entry.

**Guard (b) — programme bridge**

Matches `[intern/entry marker]\s+[program|scheme|talent|cohort]`, strips the phrase, then checks the remainder for a senior noun. `Intern Program Director` → strip `Intern Program` → `Director` → senior. `Entry-Level Program Director` is covered — `entry(?:-level)?` handles the hyphenated form. `Junior Staff Accountant` never matches the bridge set, so the guard stays silent and the position loop returns entry as before.

**Tests:** 18 new cases added to `classify-tier-position.test.mjs` — all 15 regressions from your list now pass, guard cases pinned (`Associate Software Engineer` → entry, `Junior Staff Accountant` → entry). Suite: 3471 passed, 0 failed.

Scott-Emberson (commented):
Thanks for this, @Zoubeir23. I am reviewing as code owner for `tests/`, so this is about `tests/classify-tier-position.test.mjs` only.

Good test, and it aims at the right seam. The rule that the leftmost level marker is the role's own is what makes `Summer Intern, Director of Product` an internship rather than a directorship, and the header at 11-13 ties the stakes to the real cost: `scan.mjs` drops a `skip_tiers` posting with no log line, so a junior candidate skipping `senior` was silently losing internships. The table drives the real `classifyTier` with expected values across all four tiers, so a no-op regression returning `mid` would fail most of these rather than slip through. Position both ways, the acronym preprocessing at 66-70, the `null` fallback, guard (a) including its negative `Associate Software Engineer → entry`, and guard (b) are all here. I ran it on `3c99c34`: all 41 pass.

One thing, minor.

**Guard (b) is tested only in the positive direction, so its scoping is unguarded.** Guard (a) proves it is scoped by asserting `Associate Software Engineer → entry` at line 87, the case where the prefix is present but the senior noun is not. Guard (b) at 90-102 has eight cases and every one expects `senior`. Nothing pins the case where the programme bridge is present but no senior noun follows, which is the exact condition at classify-tier.mjs:119 that keeps the guard from overreaching. I measured it: `classifyTier('Intern Program Coordinator')` is `intern`, `'Graduate Scheme Analyst'` is `intern`, `'Trainee Program Engineer'` is `intern`, all correct and all unasserted. Drop the trailing-noun test at line 119 in some later cleanup and all three flip to `senior`, a junior candidate skipping `senior` loses those intern programme roles, and this suite stays green. Adding one negative, for example `check('Intern Program Coordinator', 'intern')`, gives guard (b) the same scoping proof guard (a) already carries. It is worth it because the closed bridge-noun set exists precisely so the guard does not overreach, and an overreach is the failure this PR is here to prevent.

santifer:
Both guards are the right shape, and the thing I want to name is what you did *not* do: the generic adjacency rule would have looked simpler and shorter, and it would have eaten `Junior Staff Accountant`. A compound list is more code and less cleverness, which is the correct trade here. `Associate Software Engineer` falling straight through to the position loop is the tell that you scoped the guard instead of widening it.

Verified by effect rather than by reading the diff, against three points: your previous push gets 0 of the 15 cases, this one gets 15 of 15, and the 10 controls hold on both. Current `main` fails one of those controls, `Junior Staff Accountant` returns senior there, which is this PR's bug in a single line. Suite on a preview of the merge against today's main: 3734 passed, 0 failed.

Merged. 🚀

PR #3, opened 2026-04-05
Integrate contacto mode with evaluation reports

CharGrnmn:
## Summary

- Rewrites `contacto.md` to read existing evaluation reports before generating LinkedIn messages
- Adds archetype-adapted framing so the outreach pitch matches how the offer was scored
- Adds a follow-up variant for InMail/email (no 300-char limit) that includes a second proof point and case study link
- Updates `apply.md` handoff to explain that contacto will use the report context

## Why this matters

The evaluation pipeline does significant work in Block B (mapping JD requirements to CV proof points) and Block F (selecting STAR stories and case studies). But when the user runs `contacto` after evaluating or applying, none of that context is used — it does a cold WebSearch and generates a generic 3-sentence message.

For competitive roles, the LinkedIn message is often the only thing that gets a hiring manager's attention. A message that references the *exact* proof point that scored highest against *their specific JD* is materially more compelling than a generic pitch. The evaluation already did the hard work of identifying which proof points matter most — contacto should use them.

### Before
```
contacto → WebSearch for targets → generic message from cv.md
```

### After
```
contacto → read report → extract top proof points from Block B
         → adapt framing to detected archetype
         → WebSearch for targets → targeted message using evaluation intelligence
```

## Test plan

- [ ] Run `/career-ops contacto` for a company with an existing report in `reports/` — verify it reads the report and references Block B proof points
- [ ] Run `/career-ops contacto` for a company with NO report — verify it informs the user and offers to evaluate first
- [ ] After running `/career-ops apply` and confirming submission — verify the suggested next step mentions contacto will use the report

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2739, opened 2026-08-12
fix(agent-inbox): number resolve against file position, not the pending view

davidnunez:
### The bug

`resolve N` indexes into the *pending* subset:

```js
// Number against the pending view, so `list` then `resolve N` line up.
const pending = parseItems().filter((it) => !it.done);
const target = pending[n - 1];
```

Every resolve removes an item from that subset, so every higher number shifts down by one.
Read `list` once, then act on what you read, and the results land on the wrong items:

Verbatim run against `main`, five queued items, `list` read once:

```
$ node agent-inbox.mjs list
 1. [ ] alpha
 2. [ ] bravo
 3. [ ] charlie
 4. [ ] delta
 5. [ ] echo

$ node agent-inbox.mjs resolve 2 --result "R-2"
Resolved #2: bravo

$ node agent-inbox.mjs resolve 3 --result "R-3"
Resolved #3: delta            <-- list said #3 was charlie

$ node agent-inbox.mjs resolve 4 --result "R-4"
agent-inbox.mjs: no pending item #4 (3 pending)
```

Resulting queue:

```
- [ ] alpha
- [x] bravo → result: R-2
- [ ] charlie                 <-- skipped, still pending
- [x] delta → result: R-3     <-- carries charlie's result
- [ ] echo
```

Three distinct failures from one snapshot: `charlie` is silently skipped, `delta` is stamped
with a result it did not produce, and the third command dies with `no pending item #4
(3 pending)` moments after `list` displayed five items. The second is the dangerous one — it
prints `Resolved #3` and exits 0.

### Why this is the designed usage, not misuse

`modes/agent-inbox.md` tells the agent to drain the queue top-to-bottom and mark each item as
it goes:

> 2. Run each **unchecked** item top-to-bottom…
> 3. After each, mark it `[x]` and append `→ result: <one line>` — either by hand or with
>    `node agent-inbox.mjs resolve <n> --result "..."`.

An agent reads `list` once into context, works the batch, and resolves against the numbers it
was shown. Re-running `list` between every resolve is the only safe pattern under the old
behavior, and nothing in the docs said so — the code comment claimed the opposite ("so `list`
then `resolve N` line up"), which is true for exactly one resolve.

The failure mode is the worst kind for this file: `data/agent-inbox.md` is a provenance log.
A wrong result line is indistinguishable from a right one after the fact.

### The fix

Number against the full item list. `add` only ever appends, so an item's file position never
changes meaning once printed, and a batch read off one `list` stays correct regardless of
order or how many land in between.

`list` keeps showing pending-only by default, so numbers now have gaps (1, 3, 5) as items
resolve. That reads as a display bug unless you say otherwise, so `list` prints a footer:

```
(3 resolved item(s) hidden — numbers are stable file positions, so gaps are expected. `list --all` shows everything.)
```

### Two guards on the write path

A stale number from an older `list` is the residual risk, so the write refuses in the two
cases where it is most likely wrong:

1. **Re-resolving an already-`[x]` item aborts** instead of overwriting its result. This is
   what a stale number most often lands on, and silently re-stamping destroys the earlier
   provenance line.
2. **`--expect "<substring>"`** aborts unless the target item contains that text
   (case-insensitive). Turns "right command, wrong row" into an error. A valueless `--expect`
   fails rather than silently disabling the guard.

Both print the offending item's text, so the error shows you what you almost hit.

`modes/agent-inbox.md` now instructs the draining agent to pass `--expect` every time.

Happy to split `--expect` into a follow-up PR if you'd rather keep this one purely a fix —
it's bundled because it's the hardening for this specific failure, and the flag is opt-in and
backward-compatible.

### Tests

`agent-inbox-tests.mjs` gains three cases:

- **7 — the regression proper.** Queue five items, snapshot `list` **once**, then fire four
  resolves against that snapshot with no re-listing. Asserts each result landed on the item it
  named, the survivor kept its original number, and the footer explains the gap. Fails against
  the old code with results on the wrong items.
- **8** — re-resolving a done item exits 1, says why, and leaves the original result intact.
- **9** — `--expect` mismatch exits 1 and writes nothing; valueless `--expect` exits 1 and
  writes nothing; a case-insensitive match resolves normally.

Plus a `runFail` helper for exit-code assertions.

`node agent-inbox-tests.mjs` → **33 passed, 0 failed** (16 on `main`).

`node test-all.mjs` on a clean checkout of this branch → **3429 passed, 0 failed**, 2 warnings
(both environmental: no user `cv.md`, and no Go compiler for the dashboard build).

### Relationship to #2614

#2614 (open) fixes a different bug in the same file — concurrent `add` losing items — and
touches only `add` and its test. No logical overlap with this change; a textual conflict is
possible if both land. Happy to rebase on top of it, in either order.

### Docs

`docs/SCRIPTS.md` gains rows for `list` and `resolve` (neither was registered).

## Summary by CodeRabbit

* **New Features**
  * Added stable inbox item numbers that remain consistent as earlier items are resolved.
  * Added optional case-insensitive `--expect` validation when resolving items.
  * Added clearer reporting of hidden resolved-item gaps and stamped resolution results.

* **Bug Fixes**
  * Prevented resolving missing or already-completed items.
  * Improved batch resolution behavior using a single list snapshot.

* **Documentation**
  * Documented inbox listing, resolution, stable numbering, validation, result stamping, and failure scenarios.

### DIFF
--- a/agent-inbox-tests.mjs
+++ b/agent-inbox-tests.mjs
@@ -7,13 +7,19 @@
  *   1. A first `add` seeds the header + agent protocol and one pending item.
  *   2. `add` is append-only and multiline text collapses to a single bullet.
  *   3. `list` shows pending only; `list --all` shows resolved items too.
- *   4. `resolve N` ticks the N-th *pending* item and appends a one-line result,
- *      so `list` then `resolve N` line up.
+ *   4. `resolve N` ticks the N-th item and appends a one-line result.
  *   5. An empty `add` fails loudly (exit 1) rather than queuing a blank line.
  *   6. On the default path, a first `add` self-heals .gitignore (idempotent) so
  *      the personal queue isn't accidentally tracked.
  *   7. Concurrent `add` calls all survive — the queue is appended to, never
  *      rewritten, so simultaneous writers cannot clobber each other.
+ *   8. Item numbers are stable file positions: a batch of resolves read off ONE
+ *      `list` all land on the item they named. (Numbering against the pending
+ *      view instead shifted every higher number down per resolve, silently
+ *      stamping results onto the wrong items.)
+ *   9. Re-resolving an already-done item fails loudly instead of re-stamping.
+ *  10. `--expect` aborts when the target doesn't contain the substring, and a
+ *      valueless `--expect` fails rather than silently disabling the guard.
  *
  * Provisions a throwaway queue via CAREER_OPS_INBOX and a temp CWD; never
  * touches real user data.
@@ -49,6 +55,16 @@ function run(inbox, args, opts = {}) {
     stdio: ['pipe', 'pipe', 'pipe'],
     ...opts,
   });
+}
+
+// Run a command expected to fail; returns { status, stderr }.
+function runFail(inbox, args) {
+  try {
+    run(inbox, args);
+    return { status: 0, stderr: '' };
+  } catch (e) {
+    return { status: e.status, stderr: String(e.stderr || '') };
+  }
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


davidnunez:
Rebased onto `main` now that #2614 has merged — the branch was conflicting against the atomic-append change.

The rebase preserves #2614 in full; this PR touches only `parseItems`/`list`/`resolve`, and the concurrent-add test is untouched and still passing. My three cases renumber to 8–10 behind it.

`node agent-inbox-tests.mjs` → 36 passed, 0 failed. `node test-all.mjs` → 3576 passed, 0 failed (2 environmental warnings: no user `cv.md`, no Go compiler for the dashboard build).

Also trimmed the description: it previously cited a file that isn't part of this repository as prior art. The reasoning stands without it.

santifer:
Heads-up, @davidnunez: this conflicts now and the cause is mine. Last night I pushed a fix to `agent-inbox.mjs` (a7f65b6a) that routes the append through the shared pipeline lock, because the concurrent-add test was losing exactly one item of 30 on Windows. That lands in the same file you're changing.

Don't rebase yet. Your change is about how `resolve N` numbers its target, which is an independent question from the locking, and I want to review it properly before you spend time on a merge. I'll tell you when the base is settled.

PR #1635, opened 2026-07-06
fix(scan-ats-full): reject unrecognized CLI flags instead of silently ignoring them

Schlaflied:
## Summary
- `parseArgs()` in `scan-ats-full.mjs` only validated the *values* passed to `--ats`/`--seeds`, never whether the flags themselves were recognized — a typo, or a common convention like `--help`, silently fell through to default behavior: a full live scan of every ATS source (no `--dry-run`).
- Adds an explicit `KNOWN_FLAGS` allowlist and rejects any unrecognized `-`/`--` prefixed arg with a hard error (exit 1).
- Adds a real `--help`/`-h` that prints the usage block (previously only in the file's header comment) and exits 0.

## Test plan
- [x] `node scan-ats-full.mjs --help` prints usage, exits 0, no network calls
- [x] `node scan-ats-full.mjs --hepl` (typo) errors with the flag list, exits 1
- [x] `node scan-ats-full.mjs --dry-run --limit 1` still parses normally (existing flags unaffected)
- [x] `node test-all.mjs` — 1291 passed / 0 failed

Fixes #1633

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added built-in help output with support for `--help` and `-h`.
  * Improved command-line argument handling to clearly list supported flags.
* **Bug Fixes**
  * Unrecognized flags are now rejected with a clear error message.
  * Flag validation now catches unsupported `--name=value` style inputs as invalid.

Issue #1633, opened 2026-07-06
bug: scan-ats-full.mjs silently proceeds on unrecognized CLI flags instead of failing fast


Schlaflied:
## Bug

\`scan-ats-full.mjs\`'s \`parseArgs()\` only validates the *values* passed to \`--ats\` and \`--seeds\` (rejecting unknown ATS/seed source names). It never validates the flags themselves. Any unrecognized flag — including a typo, or a common convention like \`--help\`/\`-h\` — is silently ignored, and the script falls through to its default behavior: a full live scan of all ATS sources (\`ats: Object.keys(SOURCES)\`), with no \`--dry-run\`.

## Repro

\`\`\`
node scan-ats-full.mjs --help
\`\`\`

Expected: usage/help text, no network calls.
Actual: no such flag exists, so it's silently dropped and the script kicks off a real, live crawl across Greenhouse/Lever/Ashby/Workday directories (28,000+ companies) exactly as if no flags had been passed at all.

## Why this matters

This isn't just a UX nit — for a script whose default behavior is "read-only preview" style exploration is *not* the default (only \`--dry-run\` skips writes), a silently-ignored flag can trigger an unintended real scan/write with no warning. This happened live: running \`--help\` to check usage instead started an unattended full scan.

## Suggested fix

- Maintain an explicit allowlist of recognized flags (\`--since\`, \`--limit\`, \`--ats\`, \`--seeds\`, \`--dry-run\`, \`--liveness\`, \`--verbose\`, \`--md-out\`, \`--json\`, \`--include-undated\`, \`--shuffle\`)
- In \`parseArgs()\`, after parsing, check for any \`--\`-prefixed arg not in the allowlist and \`console.error\` + \`process.exit(1)\` with the list of valid flags
- Add a real \`--help\`/\`-h\` flag that prints the usage block already documented in the file header comment, rather than leaving it to fall through as unrecognized

## Origin

Found via direct use — ran \`node scan-ats-full.mjs --help\` expecting usage text, got an unattended live scan instead.

PR #2745, opened 2026-08-12
fix(reply-watch): reject unrecognized CLI flags instead of treating them as a candidates path

### LINKED ISSUE #2743: fix(reply-watch): --help (and any unrecognized flag) is treated as the candidates-file path, creating a file literally named --help
## Bug

`reply-watch.mjs`'s entrypoint reads its only CLI argument positionally with no flag validation:

```js
async function main() {
  const candidatesPath = process.argv[2] || DEFAULT_CANDIDATES_PATH;
  ensureCandidatesFile(candidatesPath);
  ...
```

Running `node reply-watch.mjs --help` (intending usage text) instead treats the literal string `--help` as `candidatesPath`. Since that "path" doesn't exist, `ensureCandidatesFile()` creates it — a real file named `--help` gets written to disk with the mock-candidate fixture content, and the script proceeds to run a full classification pass against those 4 mock rows instead of failing fast or printing usage.

Repro:
```
node reply-watch.mjs --help
# -> "Created default mock candidates file at --help"
ls -- --help   # a real file, not a flag
```

There is no `--help`/`-h` handling at all, and no rejection of unrecognized flags — any typo'd flag silently becomes a bogus "candidates file" path instead of erroring.

## Why this matters

Same failure class as the `scan-ats-full.mjs` bug fixed in #1633/PR #1635: an unrecognized/misread flag falls through to default positional-argument behavior instead of failing fast, and the blast radius here is worse than a no-op — it writes a stray file to the working directory.

## Proposed fix

Mirror the `scan-ats-full.mjs` pattern (`KNOWN_FLAGS` allowlist, hard error + exit 1 on anything unrecognized, real `--help`/`-h` printing usage and exiting 0):

```js
const KNOWN_FLAGS = ['--help', '-h'];

if (process.argv.includes('--help') || process.argv.includes('-h')) {
  console.log('Usage: node reply-watch.mjs [path/to/candidates.json]');
  process.exit(0);
}

const positional = process.argv.slice(2).filter(a => !a.startsWith('-'));
const unknownFlags = process.argv.slice(2).filter(a => a.startsWith('-') && !KNOWN_FLAGS.includes(a));
if (unknownFlags.length > 0) {
  console.error('Error: unrecognized flag(s): ' + unknownFlags.join(', ') + '. Valid flags: ' + KNOWN_FLAGS.join(', '));
  process.exit(1);
}

const candidatesPath = positional[0] || DEFAULT_CANDIDATES_PATH;
```

Happy to open a PR for this.

Schlaflied:
## Summary
- `reply-watch.mjs`'s `main()` read `process.argv[2]` purely positionally with no flag validation, so `node reply-watch.mjs --help` (or any typo'd/unrecognized flag) was silently treated as the candidates.json path.
- Since that "path" didn't exist, `ensureCandidatesFile()` created a real file named e.g. `--help` on disk and ran a full classification pass against the mock fixture instead of printing usage or failing fast.
- Adds a `KNOWN_FLAGS` allowlist (`--help`, `-h`): `--help`/`-h` now print a short usage line and exit 0; any other flag-looking arg errors (`Error: unrecognized flag(s): ...`) and exits 1. Neither path touches the filesystem. The existing no-args-uses-default and explicit-path behaviors are unchanged.
- Mirrors the precedent already established in `scan-ats-full.mjs` for this exact failure class (#1633 / PR #1635).

Closes #2743

## Test plan
- [x] Added CLI-level tests to `test-all.mjs` (reusing the existing `scriptTmp` throwaway-copy smoke-test fixture, since `reply-watch.mjs` has no `--self-test` mode and its existing coverage lives in `test-all.mjs`, not a sibling `.test.mjs`):
  - `--help` / `-h` print usage, exit 0, and do not create any file.
  - `--bogus` errors, exits 1, and does not create any file.
  - Regression: no-args-uses-default and explicit-valid-path-argument behaviors run to completion unchanged.
- [x] `node test-all.mjs --quick` — 3431 passed, 0 failed, 1 pre-existing warning (unrelated to this change).
- [x] Manually verified pre-fix repro (`node reply-watch.mjs --help` created a stray `--help` file containing the mock fixture and ran a full classification pass) and post-fix behavior (`node reply-watch.mjs --help` now only prints usage and exits 0, no file created).

## Summary by CodeRabbit

* **New Features**
  * Added `--help` and `-h` options with usage information.
  * Added support for specifying a candidates file path from the command line.
* **Bug Fixes**
  * Unrecognized command-line options now produce an error instead of being ignored.
  * Preserved default-path behavior when no file path is provided.
* **Tests**
  * Added regression coverage for help, invalid options, default paths, and explicit file paths.

### DIFF
--- a/reply-watch.mjs
+++ b/reply-watch.mjs
@@ -214,16 +214,16 @@ const USAGE = 'Usage: node reply-watch.mjs [path/to/candidates.json]';
 async function main() {
   const args = process.argv.slice(2);
 
-  if (args.includes('--help') || args.includes('-h')) {
-    console.log(USAGE);
-    process.exit(0);
-  }
-
   const positional = args.filter(a => !a.startsWith('-'));
   const unknownFlags = args.filter(a => a.startsWith('-') && !KNOWN_FLAGS.includes(a));
   if (unknownFlags.length > 0) {
     console.error(`Error: unrecognized flag(s): ${unknownFlags.join(', ')}. Valid flags: ${KNOWN_FLAGS.join(', ')}`);
     process.exit(1);
+  }
+
+  if (args.includes('--help') || args.includes('-h')) {
+    console.log(USAGE);
+    process.exit(0);
   }
 
   const candidatesPath = positional[0] || DEFAULT_CANDIDATES_PATH;
--- a/reply-watch.mjs
+++ b/reply-watch.mjs
@@ -208,8 +208,25 @@ async function updateTrackerStatuses(updates) {
   }
 }
 
+const KNOWN_FLAGS = ['--help', '-h'];
+const USAGE = 'Usage: node reply-watch.mjs [path/to/candidates.json]';
+
 async function main() {
-  const candidatesPath = process.argv[2] || DEFAULT_CANDIDATES_PATH;
+  const args = process.argv.slice(2);
+
+  if (args.includes('--help') || args.includes('-h')) {
+    console.log(USAGE);
+    process.exit(0);
+  }
+
+  const positional = args.filter(a => !a.startsWith('-'));
+  const unknownFlags = args.filter(a => a.startsWith('-') && !KNOWN_FLAGS.includes(a));
+  if (unknownFlags.length > 0) {
+    console.error(`Error: unrecognized flag(s): ${unknownFlags.join(', ')}. Valid flags: ${KNOWN_FLAGS.join(', ')}`);
+    process.exit(1);
+  }
+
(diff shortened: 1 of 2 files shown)


santifer:
Merged, @Schlaflied: you filed #2743 and fixed it the same day. The detail that makes this worth doing is the one you spotted: --help did not just fail, it CREATED a candidates file literally named --help. Silent wrong-thing beats loud error every time, and now it is neither. 🚀

PR #2747, opened 2026-08-12
feat(pdf): batch render path reusing one Chromium across N CVs (#2384)

### LINKED ISSUE #2384: perf(pdf): a fresh Chromium per CV costs 613ms each — no batch path exists
`renderHtmlToPdf` launches a fresh Chromium per invocation (generate-pdf.mjs:604-607) and closes it in the `finally` at :658. There is no batch path: `batch-tailor.mjs:90` spawns a full `claude -p` agent subprocess per role and never calls generate-pdf directly, so generating N tailored CVs means N browser launches.

## Measured

Through the existing `opts.launchBrowser` seam, in-process:

| strategy | per CV |
|---|---|
| launch + close per CV (current) | 809 ms |
| one shared browser | 196 ms |

That is 613 ms per CV, or 76%. Interestingly `chromium.launch()` on its own is only 68 ms, so most of the cost is teardown plus the cold first page rather than the launch itself.

The CLI shape adds a fixed per-process cost before any work happens: node boot 53 ms plus `import('playwright')` **380 ms**, so 433 ms, about 40% of a single run. Measured end to end, 1,090 ms per CV warm and 2,094 ms cold.

For 10 CVs that is roughly 10.9 s today against ~2.5 s batched.

## Fix

Add `generate-pdf.mjs --batch <manifest>` that launches one browser and loops `renderHtmlToPdf` over the entries. The injection point already exists at generate-pdf.mjs:604 and the tests already use it, so this is additive: the single-file path stays exactly as it is.

Use a fresh page per CV inside the shared browser rather than reusing one page, to avoid state leaking between renders. That is what the benchmark above did.

While in the file, `inlineLocalFonts` (generate-pdf.mjs:529) re-reads and re-base64-encodes every font on every render. Hoisting that to a module-level cache is free within a batch and harmless outside it.

Happy to send a PR for the batch flag if the shape looks right.

Scott-Emberson:
## What

Adds a `--batch` path to `generate-pdf.mjs` so N tailored CVs render through one Chromium instead of relaunching the browser per document. Closes #2384.

## Why

`renderHtmlToPdf` launched and tore down a fresh Chromium on every call. With no batch path, generating N CVs paid N browser launches, roughly 613ms of avoidable launch/teardown per CV. `inlineLocalFonts` also re-read and re-encoded every font on every call.

## How

- Split the render in two. `renderInPage` does the page-level work (theme/print/font injection, temp file, page, PDF, page-budget, manifest); a thin `renderHtmlToPdf` owns only the browser lifecycle. Both the single path and the batch path call `renderInPage`, so a single-CV render is byte-identical whether it runs alone or in a batch.
- `renderBatch` launches one browser through the existing `opts.launchBrowser` seam and closes it in a `finally` at the batch boundary, so the browser never outlives its batch and is torn down even if a document throws. Each document renders on its own page, closed in a per-entry `finally` so pages do not accumulate into the shared browser. A failing document is recorded and skipped so it cannot poison the rest.
- `--batch=<manifest.json>` reads a JSON array of `{input, output, format?, reportNum?}`, validates and ATS-normalizes each entry independently, writes an ordered `<manifest>.results.json`, and exits non-zero if any document fails. Rendering is serial on purpose: a shared font cache plus one long-lived browser make parallel rendering a determinism and memory hazard this PR does not take on.
- `inlineLocalFonts` now memoizes encoded fonts by resolved absolute path.

## Maintainer conditions

- Shared browser does not outlive its batch. It is closed in a `finally` at the batch boundary.
- One failing CV does not poison the rest. Each renders on its own page, failures are captured, and the batch continues.
- Single-CV path stays behavior-identical and keeps the `opts.launchBrowser` seam.

## A note on "byte-identical"

Chromium embeds `/CreationDate`, `/ModDate`, and a random `/ID` in every PDF, so two renders of the same HTML already differ byte-for-byte. The regression test compares after normalizing those three fields, which is the same approach the existing page-budget test already uses. Flagging it because condition 3 was worded as "byte-identical"; happy to adjust the wording if you read it differently.

## Tests

- `tests/generate-pdf-batch.test.mjs`: a batch of 3 with the middle document failing. The other two still render, Chromium launches exactly once, the exit code is non-zero, and the results manifest reflects per-entry status. Also covers a single-vs-batch normalized-identical check and an all-success exit-zero case.
- `tests/generate-pdf-font-cache.test.mjs`: an encoded font is served from cache after its source file is deleted, and the output is unchanged.
- Existing `tests/generate-pdf-page-budget.test.mjs` still passes.

## Summary by CodeRabbit

* **New Features**
  * Added batch PDF generation from a JSON manifest.
  * Reuses one browser session while preserving output order and isolating entry failures.
  * Added `--batch` support with clear usage guidance and accurate success/failure exit statuses.
  * Added caching for successfully embedded local fonts.

* **Bug Fixes**
  * Improved cleanup after document and batch operations.
  * Added manifest validation and input/output path safety checks.

* **Tests**
  * Added coverage for batch processing, failure isolation, path validation, output consistency, and font caching.


santifer:
Slow first reply on this one too, @Scott-Emberson — same reason as #2762, same apology. To the PR:

Splitting `renderInPage` (page-level work) from `renderHtmlToPdf` (browser lifecycle) is the refactor that makes the batch path safe rather than parallel-to-it. The failure mode I'd have looked for is the one where a batch path grows its own copy of the theme/print/font injection and then diverges from the single path — one gets a fix, the other doesn't, and the difference only shows up in somebody's PDF. Both calling `renderInPage` closes that by construction.

613 ms of avoidable launch/teardown per CV is the kind of number that only matters at N, and N is exactly the case `--batch` exists for. The `inlineLocalFonts` re-read-and-re-encode per call is arguably the better half of the win.

**Status:** not merged tonight. `generate-pdf.mjs` is a 🟡 file and this is +823 lines, and I've hit the per-session merge ceiling I hold myself to — deliberately, so the tail of the queue doesn't ride on end-of-session judgement. A PDF pipeline change deserves the read at the start of a session, not after fifteen merges.

Near the top next cycle, alongside #2762.

santifer:
Splitting renderInPage from renderHtmlToPdf is what makes the batch path safe rather than parallel-to-it, exactly as reviewed. One browser for N CVs is the right cost model. Merged! 🚀

Issue #2709, opened 2026-08-11
providers: the shared retry helper cannot be used by half the providers that need it


santifer:
### The pattern

Every few days a provider PR lands (or arrives) carrying its own paging delay and its own retry loop: a local `sleep`, a local attempt counter, a local decision about which errors are worth retrying. We already have `providers/_http.mjs` for exactly this, and its header comment says so:

> shared by every provider that retries a fetch ... so all of them get the same mature semantics: exponential backoff, jitter, and a `Retry-After` that is honoured but CLAMPED so a hostile or misconfigured `Retry-After: 86400` cannot stall a sweep, instead of each one re-deriving them independently.

Measured on `main` today:

| | count |
|---|---|
| providers total | 81 |
| import anything from `_http.mjs` | 9 |
| define their own `sleep` | 10 |
| run their own retry loop without the shared classifier | 1 (`meituan.mjs`) |

So the drift is smaller than it looks, but it is live: an open PR (#2682) is adding a fresh copy right now, and it is doing so for a reason worth reading (see below).

The counter-example is #2708, which does the opposite: it routes `getro.mjs` onto `fetchJsonWithRetry` and deletes the raw call. That provider fetches JSON, so the shared helper covered it. That is the whole pattern in one pair of PRs: when the shared path fits, people take it.

### Why this actually costs something

`isRetryableError` was fixed in #2657: a redirect refused by the mandatory SSRF guard (#1440) surfaces as a `TypeError` with no `status`, and the old classifier read "no status" as "network blip, retry it". So a provider would burn its whole retry budget re-requesting a URL that can never succeed, then report the result as a timeout.

That fix lives in one function. Anything running its own loop never got it, and nothing tells you: the symptom is a slow scan and a misleading error, not a failure.

Same shape for the `Retry-After` clamp. The shared helper caps a server-declared delay so `Retry-After: 86400` cannot park a scan for a day; a hand-rolled loop that honours the header literally has no such floor.

### The real reason people copy instead of import

Worth saying plainly, because "just use the shared one" is not currently good advice. `workable.mjs` keeps its own loop and **documents why**:

> the shared helper only wraps `fetchJson` ... this provider needs to give up before exhausting its retries when the server declares a long `Retry-After`, which the shared helper's clamp-and-keep-retrying policy doesn't support

[...]

PR #2748, opened 2026-08-12
providers: dedupe the inter-page sleep helper into _http.mjs

### LINKED ISSUE #2723: providers: three identical copies of the same `sleep` helper, and the shared one is not exported
**Time:** ~20 min · **Difficulty:** Beginner · **Good first issue**

Three providers each define their own `sleep` helper, and they are the same line character for character:

```js
// providers/alibaba.mjs, providers/avature.mjs, providers/tencent.mjs
const sleep = (ms) => (typeof ctx?.sleep === 'function' ? ctx.sleep(ms) : new Promise((r) => setTimeout(r, ms)));
```

`providers/_http.mjs` already has this exact helper (line 87), but it is not exported, so nobody can use it and everyone re-types it. Three copies today, and a fourth arrives with the next provider.

Note this is *pacing between pages*, not retry backoff. These three run no retry loop at all, which is why this is a small, self-contained change and not the larger retry work tracked in #2709.

## What to do

1. In `providers/_http.mjs`, add `export` to the existing `sleep` function. Do not rewrite it: the `ctx.sleep` branch exists so unit tests never wait on the wall clock, and that behaviour has to stay.
2. In each of `providers/alibaba.mjs`, `providers/avature.mjs` and `providers/tencent.mjs`, delete the local `const sleep = …` line and import the shared one instead.
3. Match how the existing importers write it, so the file looks like its neighbours:

```js
// providers/workday.mjs and providers/oraclecloud.mjs already do this
import { BROWSER_LIKE_USER_AGENT, isRetryableError } from './_http.mjs';
```

```bash
node test-all.mjs
```

Should end with the global results block and `0 failed`. Then confirm the copies are gone:

```bash
grep -rn "const sleep = " providers/*.mjs
```

Only `providers/_http.mjs` should appear (and `workable.mjs`, which keeps its own for a documented reason explained in its header comment: leave that one alone).

## Why it matters

`_http.mjs` says in its own header that it exists so providers "all get the same mature semantics instead of each one re-deriving them independently". Every copy that lives outside it is a place a future fix will not reach. That is not hypothetical here: a retry-classification fix landed in `_http.mjs` in #2657 and the hand-rolled copies elsewhere never got it.

Ask anything in the comments and I will answer. If you want it, say so here and I will assign it to you.

BhadaneAaditya:
Three providers each hand-rolled the same inter-page pacing helper, an exact
character-for-character copy of the one `providers/_http.mjs` already defines but
does not export:

```js
const sleep = (ms) => (typeof ctx?.sleep === 'function' ? ctx.sleep(ms) : new Promise((r) => setTimeout(r, ms)));
```

Fixes #2723

This PR exports the shared helper and routes the migrated providers through it,
so a future fix to pacing semantics only has to land in one file again (`_http.mjs`
already says its purpose: "all get the same mature semantics instead of each one
re-deriving them independently").

### Changes

- `providers/_http.mjs` — add `export` to the existing `sleep(ms, ctx)` function.
  Not rewritten: the `ctx.sleep` branch is kept so unit tests never wall-clock wait.
- `providers/alibaba.mjs` — drop the local copy, import `sleep` from `./_http.mjs`,
  pass `ctx` at the pacing call site (`sleep(INTER_PAGE_DELAY_MS, ctx)`).
- `providers/avature.mjs` — same (both pacing call sites, including the
  `jobOffset -> offset` self-heal retry).
- `providers/tencent.mjs` — same.
- `providers/meituan.mjs` — same. **Not named in the issue**, but it carried an
  identical 4th copy; the issue's own verification grep only passes with it
  migrated too. Happy to drop this file if you'd rather keep a strict 3-file diff.
- `providers/workable.mjs` — untouched. It keeps its own `sleep` on purpose
  (documented in its header: a local retry loop that diverges from the shared
  helper, giving up on long Retry-After and handling text + JSON transports).

### Why the call sites pass `ctx`

The shared helper's signature is `sleep(ms, ctx)` — `ctx` is a positional
argument, not closed over. The local copies captured `ctx` from the enclosing
`fetch(entry, ctx)` scope, so simply dropping in the import would silently bypass
the `ctx.sleep` fast path (tests would fall through to real `setTimeout`). Each
migrated call site therefore passes `ctx` explicitly to preserve that behaviour.

### Verification

- `grep -rn "const sleep = " providers/*.mjs` → no matches (all copies migrated).
- `node --check` passes on all five edited files.
- Targeted suites (all green):
  - `alibaba` → 23 passed, 0 failed
  - `avature` → 21 passed, 0 failed
  - `tencent` → 22 passed, 0 failed
  - `meituan` → 22 passed, 0 failed
  - `_http` → 8 passed, 0 failed
- Full `node test-all.mjs` attempted on a Windows host: it cannot complete because
  `tests/batch-runner-score-delimiter.test.mjs` spawns `bash` (ENOENT). That is a
  pre-existing environment limitation, unrelated to this change — on a POSIX host
  the global results block should end with 0 failed.

### Notes for the reviewer

1. meituan.mjs migration (see Changes above) — kept or dropped on your call.
2. No new tests added: the 4 providers' existing `ctx.sleep` pacing assertions
   already cover the newly exported helper end to end.
```

---

## 3. COMMIT MESSAGE (paste when committing)

Short version:

```
providers: dedupe the inter-page sleep helper into _http.mjs (fixes #2723)
```

Full version (GitHub will show this as the commit body):

```
Export the shared sleep helper from _http.mjs and use it in the four
providers that carried identical hand-rolled copies (alibaba, avature,
tencent, meituan). Forwards ctx at each pacing call site so the ctx.sleep
test fast-path keeps working. workable.mjs keeps its own on purpose.
```

---

## 4. FILES CHANGED (quick reference for the PR sidebar)

1. `providers/_http.mjs`
2. `providers/alibaba.mjs`
3. `providers/avature.mjs`
4. `providers/tencent.mjs`
5. `providers/meituan.mjs`

No other files changed.

---

## 5. BEFORE YOU OPEN THE PR — CHECKLIST

- [ ] All 5 edits are in place (see `copypaste.txt` for the exact old→new text).
- [ ] `grep -rn "const sleep = " providers/*.mjs` prints nothing.
- [ ] `node test-all.mjs --only alibaba` → 0 failed
- [ ] `node test-all.mjs --only avature` → 0 failed
- [ ] `node test-all.mjs --only tencent` → 0 failed
- [ ] `node test-all.mjs --only meituan` → 0 failed
- [ ] On a Mac/Linux machine you can also run the full `node test-all.mjs` → 0 failed
      (on Windows it can't finish because `bash` is missing — not a code problem).
- [ ] Mention `Fixes #2723` in the PR body so GitHub auto-links the issue.

---

## 6. ISSUE #2723 — DESCRIPTION (for reference / to paste)

### The issue in short

Three providers each had their own copy of the same "sleep" helper, identical
line for line. The shared file `providers/_http.mjs` already had this exact
helper (line 87) but it was **not exported**, so nobody could import it — and
everyone just re-typed it instead. Three copies existed at the time the issue
was written, and a fourth was expected to arrive with the next provider.

A 4th copy (in `meituan.mjs`) actually did exist and is also fixed by this PR.

### GitHub issue body (original, trimmed to the essentials)

> **Title:** providers: three identical copies of the same sleep helper, and the
> shared one is not exported
>
> **Difficulty:** Beginner · Good first issue · ~20 min
>
> Three providers each define their own sleep helper, and they are the same line
> character for character:
>
> ```js
> // providers/alibaba.mjs, providers/avature.mjs, providers/tencent.mjs
> const sleep = (ms) => (typeof ctx?.sleep === 'function' ? ctx.sleep(ms) : new Promise((r) => setTimeout(r, ms)));
> ```
>
> `providers/_http.mjs` already has this exact helper (line 87), but it is not
> exported, so nobody can use it and everyone re-types it. Three copies today,
> and a fourth arrives with the next provider.
>
> Note this is pacing between pages, not retry backoff. These three run no retry
> loop at all, which is why this is a small, self-contained change and not the
> larger retry work tracked in #2709.
>
> **What to do**
> 1. In `providers/_http.mjs`, add `export` to the existing `sleep` function.
>    Do not rewrite it: the `ctx.sleep` branch exists so unit tests never wait on
>    the wall clock, and that behaviour has to stay.
> 2. In each of `providers/alibaba.mjs`, `providers/avature.mjs` and
>    `providers/tencent.mjs`, delete the local `const sleep = …` line and import
>    the shared one instead.
> 3. Match how the existing importers write it (workday.mjs / oraclecloud.mjs
>    already do this).
>
> **How to check you got it right**
> - `node test-all.mjs` should end with the global results block and 0 failed.
> - `grep -rn "const sleep = " providers/*.mjs` → only `providers/_http.mjs`
>   should appear (and `workable.mjs`, which keeps its own for a documented
>   reason explained in its header comment: leave that one alone).
>
> **Why it matters**
> `_http.mjs` says in its own header that it exists so providers "all get the
> same mature semantics instead of each one re-deriving them independently".
> Every copy that lives outside it is a place a future fix will not reach — a
> retry-classification fix landed in `_http.mjs` in #2657 and the hand-rolled
> copies elsewhere never got it.

### How my PR answers the issue

| Issue ask | What the PR does |
|-----------|------------------|
| Export the shared `sleep` (do not rewrite) | Added `export` to `sleep(ms, ctx)` in `_http.mjs`; body untouched, so the `ctx.sleep` test fast-path is preserved. |
| Delete the local copy in each of the 3 named providers + import the shared one | Done in `alibaba.mjs`, `avature.mjs`, `tencent.mjs`. |
| Match the existing importers' style | Imported as `import { sleep } from './_http.mjs';` — same pattern as workday/oraclecloud. |
| `test-all.mjs` → 0 failed | Targeted suites all green (alibaba 23, avature 21, tencent 22, meituan 22, _http 8). Full suite blocked on Windows only by a pre-existing missing-`bash` test. |
| `grep` shows only `_http.mjs` (+ workable) | Grep now prints nothing — all copies migrated, including the 4th in `meituan.mjs`. |
| Leave `workable.mjs` alone | Untouched. |

**Bonus over the issue's scope:** the 4th identical copy in `meituan.mjs` is also
migrated, so the issue's own `grep` acceptance check passes and no straggler
copy remains to be missed by future fixes.

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved delay handling between paginated requests for more consistent request pacing.
  * Preserved context-aware waiting during retries and pagination across supported providers.
  * Maintained existing first-request behavior while making delays more reliable and consistent.
  * Reduced inconsistencies in retry timing, helping pagination complete more predictably without changing expected provider behavior.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


BhadaneAaditya (commented):
I hope this solve the  #2723 issue

BhadaneAaditya (commented):
the changes need an attention in workflow.

BhadaneAaditya (commented):
CAN ANYONE CONFIRM THIS PR??

ChristianMiller43 (approved):
LOOKS GOOD

santifer:
Welcome to career-ops, @BhadaneAaditya. The issue said three providers and you did four, which is the right kind of overshoot: I checked, and the local helper in `meituan.mjs` was character-for-character identical to the other three. The issue undercounted; you read the code instead of the ticket.

The thing that could have gone wrong here and did not: the local `sleep(ms)` closed over `ctx`, the shared one takes it as a second argument, so a single converted call site that forgot to pass it would go from "test clock" to "real clock" silently, and the only symptom would be a slower suite. Every call site in the diff passes `ctx`, and the four provider suites confirm it: alibaba 23 green, avature 21, tencent 22, meituan 22, zero red. Full suite on a preview against today's main: green.

You also left the six providers that still define their own `sleep` alone, which is correct. Those sit inside retry loops and belong to the larger work in #2709, and folding them in here would have turned a contained change into one that needed a much harder review.

PR #2749, opened 2026-08-12
fix(web): stop the 16-chip cap from truncating the user's own portals.yml

WJiangH:
## The bug

`cleanChips` caps every keyword list at 16. That number is smaller than the config it has to carry, and it is applied in two places where it destroys data rather than bounding a paste.

**1. The shipped default cannot survive its own round-trip.**

`templates/portals.example.yml` ships **37** `title_filter.positive` keywords. `seedExploreFilters()` reads that file through `cleanChips`, so the Explorer reads 37 and hands the scanner 16 — silently, with nothing shown in the UI.

`scan.mjs` reads the same `portals.yml` and uses all 37. So one config file produces two different searches depending on whether you ran `/career-ops scan` or the web Explorer.

**2. The cap applied to a merge, so typing a chip deleted chips.**

`filter-builder`'s `commit()` cleans `[...values, ...parts]`. On a list already over the cap, `cleanChips` keeps the first 16 and breaks — so adding one chip discards everything past the 16th *and* does not append the new chip either.

Measured on a 37-entry list:

```
existing: 37 → after adding one chip: 16
new chip present? false
lost: 21 existing keywords
```

## The fix

The bound is worth keeping — these values get serialized into a YAML file and passed to a scanner — so it is raised to 512 and documented as a rail against a pathological paste (a spreadsheet column, a JSON blob) rather than a limit on how many keywords a search may use. No call site changes.

`CHIP_CAP` is now exported so the tests assert against the contract rather than a repeated literal.

## Tests

Two regression tests, both of which **fail at the old cap and pass at the new one**:

- the shipped `portals.example.yml` round-trips without losing keywords (reads the real template, so it keeps tracking it)
- adding one chip to an over-16 list appends instead of truncating

The existing cap test is kept, rewritten against `CHIP_CAP` so it still proves a pathological paste is bounded.

`node test-all.mjs` → 3431 passed, 0 failed. `npm run typecheck` clean.

## Summary by CodeRabbit

* **New Features**
  * Increased the maximum number of supported chips from 16 to 512.
  * Preserved existing chips when adding new entries beyond the previous limit.
  * Ensured all configured positive keywords remain available.
  * Expanded chip capacity while maintaining existing chip-cleaning behavior.

* **Tests**
  * Added coverage for the expanded chip limit, boundary conditions, and configuration preservation.


santifer (approved):
Verified the bug from the config side rather than taking the description on faith, and it's exactly as you state:

```
title_filter.positive in templates/portals.example.yml: 37 keywords   (CHIP_CAP = 16)
→ the web Explorer reads 16 of 37; scan.mjs uses all 37
```

One config file, two different searches depending on whether the user came through the CLI or the web. That's a direct violation of the rule this app is built on — **the web is a view over the core's files, never a second engine** — and it was invisible: nothing in the UI showed that 21 keywords had been dropped.

The merge case is worse and you're right to call it the more serious of the two: `commit()` cleaning `[...values, ...parts]` on an over-cap list means a keystroke *deletes* data. A cap that silently truncates a merge isn't a bound, it's a data-loss bug with a polite name.

**What makes this a yes rather than a maybe** is the test:

```js
const positive = yaml.load(yml)?.title_filter?.positive ?? [];
assert.ok(positive.length > 16, "template should still be the realistic size this guards");
assert.equal(cleanChips(positive).length, positive.length);
```

You didn't hardcode 37 — you read the **real shipped template** and assert it round-trips. If someone grows that file past the new cap, this goes red on its own. That's derivation, not a copy of a number, and it's the difference between a test that preserves a fact and one that can discover a new violation.

Keeping the bound at 512 rather than removing it is also the right call, and your comment says why better than I would: unbounded is not a property worth having on a value that gets serialized into YAML and handed to a scanner. It bounds the process instead of the user.

Checked the one thing that could have gone wrong with a 32× raise: no consumer assumes ≤16, and no UI path slices the list. A pathological 512-chip paste would render 512 pills — ugly, user-provoked, and visible, which is a strictly better failure than silently losing half of someone's search config.

15/15 green. Merging. Thank you for finding this one — it's the kind that never gets reported because the search just quietly returns less.

santifer:
**I pushed a one-line fix to your branch rather than asking you for another round — and there's something you should know about why it was never caught.**

`test-all` failed on all three platforms with:

```
❌ js-yaml default import(s) — these break on js-yaml 5, use `import * as yaml`:
   web/tests/lib/clean-chips.test.mjs (static default import)
```

`#2656` and `#2767` standardised every js-yaml import in the repo on the namespace form, and both landed **after** this branch was cut. Your test was written against the convention that existed when you wrote it. Changed `import yaml from "js-yaml"` → `import * as yaml from "js-yaml"`, nothing else. `test-all` 3787/0, your suite 21/21.

**The part that matters more than the fix:** this PR was showing green when it wasn't. The `test (ubuntu|macos|windows)` checks — the ones branch protection actually requires — had **never run**. GitHub was holding them in `action_required`, waiting for a maintainer to approve workflow runs on an external contribution. The checks that did show green were the ones that don't need approval (label, CodeRabbit, GitGuardian), so at a glance the PR looked verified.

I approved the runs, they ran for the first time, and they immediately found this. So: the failure isn't a regression from anything you did, and it isn't from the branch update either — it's the first honest test result this PR has ever had.

That's on our side of the fence twice over, and I've raised it with the maintainer as a repo setting rather than something contributors should have to think about.

Merging as soon as the run finishes. Both of your PRs tonight (#2749, #2750) were real silent-data-loss bugs in config handling, which is the least glamorous and most damaging place for them to hide.

PR #2750, opened 2026-08-12
fix(web): read target_roles with the shape it is actually written in

WJiangH:
## The bug

`seedExploreFilters()` falls back to `config/profile.yml` when `portals.yml` has no `title_filter.positive`. That fallback has never produced a keyword — it reads both fields with the wrong shape, and the result is always `[]`.

```ts
// web/src/lib/core/portals.ts
...(typeof roles.primary === "string" ? [roles.primary] : []),
...(Array.isArray(roles.archetypes) ? roles.archetypes : []),
```

- **`primary` is a list**, in `config/profile.example.yml` and in what this app's own writer emits (`api/profile/route.ts` → `{ primary: roles.slice(0, 6) }`). The `typeof === "string"` test never fires, so the web cannot read back the profile it just wrote.
- **`archetypes` entries are `{name, level, fit}` objects.** Spreading them raw yields objects, which `cleanChips` then drops as non-strings.

Both halves empty, no error, no warning. The `seededFrom.push("profile.yml")` branch is unreachable.

## The fix

The core has read this block correctly all along, in `providers/_profile-keywords.mjs::profileTargetKeywords`. A build-time import of it is not available from `web/` — Turbopack's root is pinned there and refuses modules outside it (`web/src/lib/tracker-table.mjs`'s header documents the same constraint) — so this adds `web/src/lib/profile-keywords.mjs` as an explicit mirror with a header naming its counterpart, and `core/portals.ts` calls it.

Checked against a real `config/profile.yml`: **0 keywords before, 9 after.**

## Tests

New suite, driven by the two shapes that actually exist on disk rather than synthetic fixtures:

- the shipped `config/profile.example.yml` yields its roles — every `primary[]` entry and every `archetypes[].name`
- `primary` is a list, not a string (the regression)
- archetypes contribute `.name`, not the whole object
- malformed input (`null`, a string, `target_roles: []`, a scalar `primary`, …) yields `[]` rather than throwing — this is a tolerant seeding path, not a validator
- non-string entries are dropped, so `undefined`/`{}` can never reach a list that gets serialized into YAML
- **parity with `providers/_profile-keywords.mjs` asserted directly.** The module exists because a hand-written copy drifted, so the test imports the core helper and compares — it runs under plain Node, where the root import Turbopack refuses at build time is available.

`node test-all.mjs` → 3431 passed, 0 failed. `npm run typecheck` clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved profile keyword extraction for consistent handling of primary roles and archetypes.
  - Added safeguards for missing, malformed, or unsupported profile data.
  - Preserved keyword ordering while excluding invalid entries.

- **Tests**
  - Added coverage for supported profile formats, ordering, filtering, malformed inputs, and consistency across keyword sources.


santifer (approved):
Confirmed both halves against the real files rather than the description:

```
config/profile.example.yml:28   primary:  is a LIST  ("- Senior AI Engineer")
api/profile/route.ts:47         writes    { primary: p.roles.slice(0, 6) }   → a LIST
core/portals.ts:100             tested    typeof roles.primary === "string"  → never fires
```

So the app could not read back the profile **it had just written itself**, and `archetypes` entries are `{name, level, fit}` objects that `cleanChips` then dropped as non-strings. Both halves empty, no error, `seededFrom.push("profile.yml")` unreachable. A fallback that has never once produced a keyword.

**The part I want to single out is the parity test's anti-vacuity guard:**

```js
if (fromExample.length >= 2) pass(`… (parity is not vacuous)`);
else fail(`… the mirror is returning nothing`);
```

A mirror test that only asserts `web === core` passes perfectly when both sides return `[]` — we have been bitten by exactly that shape here before, when a shared bug made two implementations agree while both were wrong. Anchoring on the real `profile.example.yml` *and* requiring a non-empty result means the test can't go green by measuring nothing. That's the difference between comparing two things and knowing either is right.

Also checked the thing that would have made this a false green: `tests/**/*.test.mjs` is auto-discovered by `test-all.mjs` (not a registration list), so the new parity file genuinely runs. And `providers/_profile-keywords.mjs` already exists on `main` — this imports it rather than forking it, so there's no new root-level surface to coordinate.

One deliberate behaviour worth naming so nobody "fixes" it later: the mirror ignores `primary` written as a bare scalar. I verified the core does the same —

```
core, primary as LIST  : ["ML Eng"]
core, primary as SCALAR: []
```

— so this is **fidelity, not a gap**. The web must not be more permissive than the core it renders; if the scalar form should be accepted, that's a change to `providers/_profile-keywords.mjs` and the mirror follows. Keeping them equal is the point.

14/14 green. Merging.

PR #2755, opened 2026-08-12
security(archive-posting): add SSRF / private-IP egress safeguards

### LINKED ISSUE #1956: archive-posting.mjs lacks SSRF / private IP egress safeguards, exposing local network resources
### Description
The `archive-posting.mjs` script launches a Playwright Chromium headless instance and navigates to user-provided URLs (supplied via CLI or extracted from `data/pipeline.md`) to render them to PDF.

Unlike `liveness-browser.mjs` and `generate-pdf.mjs` which have been hardened with request interceptors to prevent egress to private/internal IP ranges (SSRF), `archive-posting.mjs` does **not** register any route interceptors or DNS resolution validators.

An attacker can place internal/private URLs (e.g., local administrative panels like `http://127.0.0.1:8080/admin` or cloud metadata services like `http://169.254.169.254/latest/meta-data/`) in `data/pipeline.md` or pass them via CLI. Chromium will navigate to them, render private/local services or credentials to a PDF, and save them in the `jds/` directory, exposing internal details.

### Proposed Solution
Integrate the SSRF mitigation functions (e.g., `rejectPrivateOrInvalid` and request/redirect routing interception using a DNS validation guard) from `liveness-browser.mjs` into the Playwright initialization flow of `archive-posting.mjs`.

Specifically:
1. Intercept all outgoing requests via `page.route('**/*', ...)` in `archive-posting.mjs`.
2. Check the hostname of the request and reject any literal private IP addresses.
3. Stub/perform DNS resolution checks on hostnames and abort any request that resolves to private IP addresses (such as `127.0.0.1`, loopbacks, or link-local ranges) by calling `route.abort('blockedbyclient')`.

maxmilian:
## What does this PR do?

`archive-posting.mjs` drives Playwright at a URL the user pastes in (or that arrives via `data/pipeline.md`) and navigated with no egress safeguards at all, so a hostile posting link could render an internal service or a cloud metadata endpoint straight into a PDF on disk. This wires up the guard `liveness-browser.mjs` already exports — rather than adding a second one, per your note on the issue.

## Related issue

Fixes #1956

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

---

## The gap, demonstrated

A local service on `127.0.0.1:8931` standing in for something on the user's network, serving `INTERNAL-SECRET-TOKEN-9f3a` and a fake `db-password`. Same command against both trees.

**On `main` (`92dea48`) — archived it:**

```
🔗  http://127.0.0.1:8931/
   Company: unknown
   Role:    Internal Admin Console
   Output:  jds/2026-08-11_unknown_internal-admin-console.pdf
Saved (19.8 KB)
  Archived: 1 saved  0 failed
```

The PDF is not a blank page — it captured the content:

```
$ pdftotext jds/2026-08-11_unknown_internal-admin-console.pdf -
INTERNAL-SECRET-TOKEN-9f3a
metadata: db-password=hunter2
```

**On this branch — refused, nothing written:**

```
🔗  http://127.0.0.1:8931/
   ❌  Failed: refusing to archive restricted destination: blocked host 127.0.0.1
  Archived: 0 saved  1 failed
```

Exit code is 1, and `jds/` stays empty.

## What changed

Three wiring points, all reusing the shared guard:

1. **Pre-navigation check on the input URL**, so an obviously-internal target never reaches Playwright at all.
2. **`launch → newContext() + context.route('**/*') → newPage()`** instead of `browser.newPage()`. The route is registered on the *context* rather than the page, so it covers every request in the flow rather than one page's first hop.
3. **Re-check of the settled URL after navigation** — a first-hop-only check is the classic miss on a 302 into link-local space, so the landed URL is asserted too.

## Two calls that go beyond the sketch I posted on the issue

Flagging both explicitly so they're easy to veto:

**I exported `validateUrlSecurity` from `liveness-browser.mjs`.** I had said I'd import `rejectPrivateOrInvalid`, but that one only inspects the literal host — a public-looking hostname that *resolves* to `127.0.0.1` walks straight past it. `checkUrlLiveness` runs both layers in its route handler, so moving only one would have shipped half a guard. Exporting the DNS layer keeps this a single shared implementation instead of a second one. Happy to drop it to one layer if you'd rather keep that function private.

**`archive-posting.mjs` gains the repo's standard direct-run guard** (`process.argv[1] && import.meta.url === pathToFileURL(process.argv[1]).href`) and exports `archiveUrl`. This isn't a new pattern — `add-entry.mjs`, `detect-reposts.mjs`, `check-table-freshness.mjs` and a dozen others already have it; `archive-posting.mjs` was one of the stragglers. Without it, importing the module for a test runs the CLI parser and calls `process.exit`, which kills the test process. CLI behaviour is unchanged: `--help` and the no-argument path both still exit 0, and the existing `archive-posting.mjs --help` assertion in `test-all.mjs` still passes.

## A correction to something I said on the issue

I listed `generate-pdf.mjs` as having "the same" zero-SSRF-guard problem. Having read it properly, that was overstated: it only navigates to a local `file://` temp document and runs with `javaScriptEnabled: false`, so there's no attacker-controlled URL reaching `page.goto`. Not an equivalent gap, and I've kept it out of this PR rather than bundling a change that isn't warranted.

## Tests

Six cases in `test-all.mjs`, pinning the wiring rather than re-testing the guard logic that `liveness-browser`'s own section already covers — and deliberately covering both arms, so a guard that simply blocked everything would fail:

- private-IP target is refused **before a browser context is even created**
- the guard is registered on the context for `**/*`, not on a single page
- a legitimate public subresource still `continue()`s
- a redirect hop to a literal private IP is aborted
- a hostname that *resolves* to loopback is aborted (the case the literal-host check can't see)
- a page that lands on a link-local address is refused after navigation

## Verification

```
node test-all.mjs
📊 Results: 3282 passed, 0 failed, 1 warnings
```

## Summary by CodeRabbit

* **Security Enhancements**
  * Added protection against navigation to private or invalid network destinations.
  * Validates DNS resolution and redirect targets before and during web archiving.
  * Prevents unsafe requests throughout the browser session.

* **Reliability**
  * Improved command-line handling for safer archiving.
  * Ensures browser resources are closed after archiving.

* **Tests**
  * Added coverage for blocked private destinations, redirects, DNS validation, and permitted public requests.

### DIFF
--- a/archive-posting.mjs
+++ b/archive-posting.mjs
@@ -26,8 +26,9 @@ import { chromium } from 'playwright';
 import { writeFile, readFile } from 'fs/promises';
 import { existsSync, mkdirSync } from 'fs';
 import { join, dirname } from 'path';
-import { fileURLToPath } from 'url';
+import { fileURLToPath, pathToFileURL } from 'url';
 import { reportPrefix } from './jd-capture.mjs';
+import { rejectPrivateOrInvalid, validateUrlSecurity } from './liveness-browser.mjs';
 
 const ROOT = dirname(fileURLToPath(import.meta.url));
 const JDS_DIR = join(ROOT, 'jds');
@@ -35,10 +36,7 @@ const PIPELINE_PATH = join(ROOT, 'data', 'pipeline.md');
 
 // ── CLI parsing ──────────────────────────────────────────────────────────────
 
-const args = process.argv.slice(2);
-
-if (args.length === 0 || args[0] === '--help' || args[0] === '-h') {
-  console.log(`
+const HELP_TEXT = `
 ╔══════════════════════════════════════════════════════════════════╗
 ║           career-ops — Job Posting Archiver                     ║
 ╚══════════════════════════════════════════════════════════════════╝
@@ -77,9 +75,7 @@ if (args.length === 0 || args[0] === '--help' || args[0] === '-h') {
     node archive-posting.mjs "https://jobs.lever.co/acme/xyz" --report=42
     node archive-posting.mjs --pipeline
     npm run archive -- "https://jobs.lever.co/elevenlabs/abc"
-`);
-  process.exit(0);
-}
+`;
 
 let targetUrl = null;
 let overrideCompany = null;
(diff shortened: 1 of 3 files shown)


santifer:
Unrelated to your PR, @maxmilian, but you deserve the correction: the bot greeted you as a first-time contributor above, with 24 merged PRs behind you. That was a real bug (the action we used mistakes veterans for newcomers on a repo this size) and it is fixed in main now, counting properly. Sorry for the odd welcome.

santifer:
Reusing `liveness-browser.mjs`'s guard instead of writing a second one is the whole reason this merges quickly. Two egress guards in one repo drift, and the one that drifts is always the one nobody is looking at.

I did not take "it wires up the existing guard" on trust, I threw the vectors at it directly. Every one blocked:

| vector | blocked by |
|---|---|
| `169.254.169.254/latest/meta-data/` | layer 1, by IP |
| `metadata.google.internal` | **layer 2, by DNS** |
| `127.0.0.1:8080`, `[::1]`, `localhost:3000` | layer 1 |
| `10.0.0.1`, `192.168.1.1`, `172.16.0.5`, `[fe80::1]`, `0.0.0.0` | layer 1 |
| `file:///etc/passwd` | layer 1, unsupported protocol |
| `http://2130706433/` | layer 1, normalized to `127.0.0.1` |

Public URLs pass. Two of those deserve naming: the **decimal-integer form of a loopback address** is the classic way past a naive blocklist and it resolves correctly here, and the **hostname** case is the one a single-layer guard always misses, since `metadata.google.internal` is a perfectly ordinary-looking name until you resolve it. A host that does not resolve at all is also refused rather than allowed through, which is the correct direction for a guard to fail.

I also checked that it is *wired*, not merely imported: `archive-posting.mjs:300` calls `installEgressGuard(context)` on the Playwright context, so this covers the whole navigation rather than the first URL. `main` has neither symbol today.

Suite on a preview against today's main: 3810 passed, 0 failed.

Sorry this sat for two days. It was in the batch behind a CI flake that was reddening unrelated PRs, and yours was not one of the ones I could tell apart until tonight.

Merged. 🚀

PR #2759, opened 2026-08-12
fix(reserve-report-num): do not read a bare date file as a report number

### LINKED ISSUE #2757: bug(reserve-report-num): a `--md-out reports/` scan digest is read as report #2026 and pushes all later reservations to 2027+
Running `scan-ats-full.mjs --md-out reports/` writes its digest as `reports/YYYY-MM-DD.md`. `occupiedFromReports()` matches every filename against `/^(\d+)-/`, so `2026-08-12.md` captures `2026` and is recorded as an occupied report number.

Every subsequent reservation is then based past it.

## Repro

Tracker + reports containing a single real report `009-leapsome-2026-08-12.md`:

```
$ node reserve-report-num.mjs --peek
010

$ touch reports/2026-08-12.md
$ node reserve-report-num.mjs --peek
2027
```

It is permanent: the sentinel written at 2027 keeps the number occupied even after the digest is deleted.

## Why it matters

- `--md-out reports/` is a documented flag combination; `reports/` is the natural target for a markdown digest.
- The failure is silent. Nothing warns that a report number jumped by 2000.
- Report numbers appear in tracker rows, tracker report links, PDF filenames and `merge-tracker.mjs` output, so the damage spreads across the pipeline before anyone notices.

## Suggested fix

Exclude filenames that are *entirely* a date, rather than tightening the number pattern:

```js
const BARE_DATE_FILE_RE = /^\d{4}-\d{2}-\d{2}\.md$/;
```

Deliberately narrow. Requiring the full `NNN-slug-YYYY-MM-DD.md` shape breaks legitimate occupancy: `NNN-RESERVED.md` sentinels and bare `1001-taken.md` names are both exercised by the existing suite (`four-digit report files advance a contiguous range`, `unsafe report-number ranges fail before creating a partial sentinel`) and both fail under the stricter pattern. I tried that first.

Happy to send a PR with a regression test covering digest / real report / RESERVED sentinel / digest-only cases.

## Environment

- career-ops on `main`
- Node v26.5.0, macOS arm64

sonukhobragade:
Fixes #2757

## Problem

`scan-ats-full.mjs --md-out reports/` writes its digest as `reports/YYYY-MM-DD.md`. `occupiedFromReports()` matches every filename against `/^(\d+)-/`, so `2026-08-12.md` captures `2026` and is recorded as an occupied report number.

```
$ node reserve-report-num.mjs --peek
010
$ touch reports/2026-08-12.md
$ node reserve-report-num.mjs --peek
2027
```

Silent and permanent — the sentinel written at 2027 keeps the number occupied even after the digest is deleted. Report numbers reach tracker rows, tracker report links, PDF filenames and `merge-tracker.mjs`, so it spreads before anyone notices.

## Fix

Skip filenames that are *entirely* a date.

Deliberately narrow. My first attempt required the full report shape (`NNN-slug-YYYY-MM-DD.md` or `NNN-RESERVED.md`) and **broke two existing tests** — `four-digit report files advance a contiguous range` uses a bare `1001-taken.md`, and `unsafe report-number ranges fail before creating a partial sentinel` uses `<MAX_SAFE_INTEGER-1>-existing.md`. Neither carries a trailing date, and both are legitimately occupied. Excluding only the digest shape leaves all existing occupancy behaviour untouched.

## Tests

New `tests/reserve-report-num.test.mjs` (auto-discovered), 4 cases:

- a digest alongside a real report is ignored → `010`, not `2027`
- real reports still counted
- `NNN-RESERVED.md` sentinels still counted
- a digest alone leaves numbering at `001`

Full suite on this branch: **3434 passed, 0 failed**. The 1 warning is pre-existing and unrelated (Go absent → dashboard build skipped) — it is also present on unmodified `main` in my environment.

## Note

`AGENTS.md` documents `--md-out <dir>` without restriction, and `reports/` is the natural target for a markdown digest, so this is reachable from documented usage. Worth considering a follow-up that warns when `--md-out` points at `reports/`; I have left that out to keep this change minimal.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented date-named Markdown files, such as `YYYY-MM-DD.md`, from affecting report and reservation number availability.
  * Numbered reports and reservation markers continue to be counted correctly.

* **Tests**
  * Added regression coverage for reserving, releasing, and scanning report numbers.

### DIFF
--- a/reserve-report-num.mjs
+++ b/reserve-report-num.mjs
@@ -56,10 +56,22 @@ function trackerPathFor(options = {}) {
     : resolveTrackerPath(options.rootDir || ROOT);
 }
 
+// A bare date file is not a report. `scan-ats-full.mjs --md-out reports/` writes
+// its digest as `reports/YYYY-MM-DD.md`, which matches `/^(\d+)-/` and is read
+// as report #2026 — pushing every later reservation to 2027+, silently and
+// permanently.
+//
+// Deliberately narrow: report filenames in the wild are not all
+// `NNN-slug-DATE.md` (`1001-taken.md` and `NNN-RESERVED.md` are both real), so
+// requiring a trailing date would drop legitimate occupancy. Exclude only names
+// that are *entirely* a date. See tests/reserve-report-num.test.mjs.
+const BARE_DATE_FILE_RE = /^\d{4}-\d{2}-\d{2}\.md$/;
+
 function occupiedFromReports(reportsDir) {
   const occupied = new Set();
   if (!existsSync(reportsDir)) return occupied;
   for (const name of readdirSync(reportsDir)) {
+    if (BARE_DATE_FILE_RE.test(name)) continue;
     const match = name.match(/^(\d+)-/);
     if (!match) continue;
     const num = parseInt(match[1], 10);
--- a/tests/reserve-report-num.test.mjs
+++ b/tests/reserve-report-num.test.mjs
@@ -1,0 +1,73 @@
+// Regression: a dated file in reports/ must not be read as a report number.
+//
+// `scan-ats-full.mjs --md-out reports/` writes `reports/YYYY-MM-DD.md`. The old
+// occupancy scan matched `/^(\d+)-/`, so `2026-08-12.md` was read as report
+// #2026 and the next reservation jumped to 2027 — silently, and permanently.
+
+import { strict as assert } from 'assert';
+import { mkdtempSync, mkdirSync, writeFileSync, rmSync } from 'fs';
+import { tmpdir } from 'os';
+import { join } from 'path';
+import { reserveReportNumbers, releaseReportNumbers } from '../reserve-report-num.mjs';
+import { pass, fail } from './helpers.mjs';
+
+// Reserve one slot in a scratch root and report which number it got. Released
+// afterwards so the assertion reflects the occupancy scan, not leftover state.
+async function peekIn(files) {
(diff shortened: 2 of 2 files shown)


santifer:
Welcome to career-ops, @sonukhobragade. Reported it, wrote the repro, and fixed it, which is the whole loop in one go.

I reproduced your issue before reading your patch, and it lands exactly as written: with a single real report `009-...md`, `--peek` answers `010`, and after `touch reports/2026-08-12.md` it answers **`2027`**. On your branch the same sequence answers `011`. The permanence is the ugly part, since the sentinel keeps the number occupied after the digest is deleted, and that is what makes this worth fixing rather than documenting.

I also checked the reasoning behind keeping the regex narrow, because "deliberately narrow" is the kind of claim that is either careful or convenient. It is careful: `scan-ats-full.mjs:988-999` treats `--md-out` as a **directory** and always writes `${date}.md`, so `^\d{4}-\d{2}-\d{2}\.md$` covers everything the producer can emit. A wider pattern would have started guessing at filenames no tool in this repo creates. For the record, `2026-08-12-digest.md` and `2026-08-12.markdown` still resolve to 2027, and that is the right call: those can only come from a file someone typed by hand, which is a different problem with a different answer.

Suite on a preview against today's main: green. Your `012-acme-2026-01-01.md` control still reserves `013`, so real reports are untouched.

Issue #1754, opened 2026-07-10
ADR: the analytics layer — architecture, review bar, and its map to the #1025-1034 design space


santifer:
An analytics layer is emerging in career-ops, slice by slice: `stats.mjs` (lifetime funnel) → `analyze-patterns.mjs` (rejection patterns) → `upskill.mjs` (#1573, skill gaps) → and now in review: `company-history.mjs` (#1712), `funnel-velocity.mjs` + benchmarks (#1694), the status-transition ledger (#1695), and `contacts.mjs` (#1719). Individually each slice is small; together they are becoming **the layer that reads the user's own data back to them**. This ADR makes that layer's architecture explicit and project-owned, so design decisions are documented here rather than accumulating implicitly across PRs.

## The shape (as it stands)

- **Read-only analyzers over user-layer files.** Every analytics script reads the tracker / scan-history / follow-ups / reports and writes nothing (or writes only to a NEW user-layer file registered in DATA_CONTRACT + .gitignore). No analyzer mutates existing user data.
- **JSON to stdout + `--summary` + `--self-test`** — the sibling contract established by salary-gap/stats and now followed by all slices.
- **Reuse the canonical primitives:** `tracker-parse.mjs` (rows), `tracker-utils.mjs` (normalization/paths), `followup-cadence.mjs` (dates), `stats.mjs` (the ONE funnel), `detect-reposts.mjs` (repost clusters). A new slice that needs one of these imports it — a second implementation of any of them is a review blocker.
- **Non-scoring by default.** Analytics *inform*; they never silently change A–F scores or legitimacy tiers. Where a slice feeds an evaluation surface (e.g. Block G), the wiring is explicit, additive, and frozen by a test-all assertion (see #1712's "Not a legitimacy signal").
- **Honest-data rules:** external reference numbers (benchmarks) carry source + year + caveat and are user-overridable; unverifiable numbers are labeled `estimate`. Right-censoring / data-quality notes are reported, never smoothed over.

## Rules for new slices (review bar)

1. **APIs are born in the PR that's reviewed** — a consumer PR doesn't pre-define another module's export contract. If slice B needs slice A's API, A merges first.
2. New user-data files: DATA_CONTRACT User-Layer row + .gitignore + out of SYSTEM_PATHS — all three, same PR.
3. One vocabulary: date flags (`--on` vs `--date`), quality-report shapes, and tier names should converge, not fork per-slice. Divergences get an issue, not silence.
[...]

PR #2762, opened 2026-08-12
perf(analytics): shared stat-validated reports-index cache (#2385)

### LINKED ISSUE #2385: perf(analytics): report reading is 70%+ filesystem syscalls, with redundant existsSync probes and repeated parses

Scott-Emberson:
## What this does

Introduces `reports-index.mjs`: the **one** canonical Machine-Summary parser plus a disposable, stat-validated on-disk cache (`data/reports-index.json`) over `reports/*`. The three analytics scripts (`analyze-patterns.mjs`, `upskill.mjs`, `salary-gap.mjs`) each carried their own copy of the same fence regex + `js-yaml` load and re-parsed every report on every run. They now share one parser and one cache.

An entry is keyed by `(mtimeMs, size)`. On load, each report is `statSync`'d; a cached entry is reused **only** when both mtimeMs **and** size match the fresh stat — otherwise the file is re-read and re-parsed. Entries whose file no longer exists are dropped. A `version` mismatch (bump `INDEX_VERSION`) or a missing/corrupt cache rebuilds from scratch. The cache is written atomically (`writeFileAtomic` from `tracker-utils.mjs`).

## Schema decision: core vs `extras.salaryGap`

Per maintainer directive, the entry splits into:
- `summary` — the parsed object filtered to `CORE_SUMMARY_FIELDS` (the historical `analyze-patterns` `MACHINE_SUMMARY_FIELDS` **minus `advertised_comp`**). This is now the single source for the core allowlist and still contains `via` + `company_confidential`.
- `extras.salaryGap.advertised_comp` — namespaced out of the core allowlist because only `salary-gap.mjs` reads it. Keeping it out of `summary` means a salary-only field never widens the core allowlist that the batch-prompt cross-check guards.

`test-all.mjs`'s batch-prompt↔allowlist cross-check now reads `CORE_SUMMARY_FIELDS` from `reports-index.mjs` (still asserting `via` + `company_confidential` round-trip).

## Stat-always validation + `--no-cache`

Every report is stat-checked on every load — the cache is never trusted blind. All three scripts gain a `--no-cache` flag that skips both reading and writing the JSON (correctness escape hatch; data is still correct, just uncached).

## `upskill` legacy-fallback caveat

`upskill` now sources `score`/`hard_stops`/`soft_gaps` from the cached summary and reads the report **body** only as a fallback (Gap table + `| Global |` score) when the summary can't supply them. Consequence: a report whose Machine Summary already carries the gap keys no longer *also* scrapes its body Gap table. On canonical reports the Machine Summary is the structured form of that table, so this is intended; noted here for transparency.

## Other notes

- `salary-gap`: a JSON fence now parses (JSON ⊂ YAML) where the old hand-rolled `yamlStr` matcher rejected it; its self-test was updated to assert the new behavior.
- `analyze-patterns` and `upskill` gained `isMain` guards so importing them is side-effect-free.
- Output shapes / JSON contracts of all three are unchanged; verified byte-identical across cold, warm-cache, and `--no-cache` runs.
- `data/reports-index.json` is a derived user-layer artifact (already covered by `.gitignore` `data/*`), registered in `DATA_CONTRACT.md`; **not** added to `SYSTEM_PATHS` (the module is).

## Validation

- `reports-index.mjs --self-test`, and all three analyzers' `--self-test`: exit 0.
- Full `node test-all.mjs`: **3432 passed, 0 failed** (1 pre-existing env warning: `cv-sync-check` without user data).

Refs / closes #2385. ADR to follow on #1754.

## Summary by CodeRabbit

* **New Features**
  * Added shared report indexing for faster analysis and reuse of cached summaries.
  * Added support for JSON-formatted report summaries.
  * Added a `--no-cache` option to refresh report data during analysis.
  * Added self-testing for report indexing and cache behavior.

* **Bug Fixes**
  * Improved consistency of salary and gap observations across analysis tools.
  * Improved recovery of gaps when summary lists are empty.
  * Improved handling of missing, malformed, unreadable, deleted, or invalid reports.
  * Prevented unintended analysis when tools are imported.

### DIFF
--- a/DATA_CONTRACT.md
+++ b/DATA_CONTRACT.md
@@ -30,6 +30,7 @@ These files contain your personal data, customizations, and work product. Update
 | `plugins.lock` | Integrity pins + recorded consent for your enabled plugins (generated; never auto-updated) |
 | `data/applications.md` | Your application tracker (source of truth) |
 | `data/applications.db` | Derived query index over `applications.md` (SQLite, rebuilt by `node tracker.mjs sync` — safe to delete) |
+| `data/reports-index.json` | Derived Machine-Summary cache over `reports/*` (stat-validated + rebuilt by `analyze-patterns.mjs`/`upskill.mjs`/`salary-gap.mjs`; disposable — safe to delete any time) |
 | `data/pipeline.md` | Your URL inbox |
 | `data/scan-history.tsv` | Your scan history (tab-separated, append-only trailing columns; col 8: local SimHash JD fingerprint for cross-listing detection, col 9: posting date, cols 10-11: trust score/flags, col 12: normalized company key for repost/name matching). Older rows may have fewer columns — readers index by position and tolerate the absence. |
 | `data/scan-runs.tsv` | Your per-run scan counters (appended by `scan.mjs`, read by `stats.mjs`) |
--- a/analyze-patterns.mjs
+++ b/analyze-patterns.mjs
@@ -15,47 +15,15 @@
 
 import { readFileSync, existsSync, realpathSync, writeFileSync, symlinkSync, rmSync } from 'fs';
 import { join, dirname, relative, sep } from 'path';
-import { fileURLToPath } from 'url';
-import { load as yamlLoad } from 'js-yaml';
+import { fileURLToPath, pathToFileURL } from 'url';
 import { resolveColumns, parseTrackerRow, normalizeVia } from './tracker-parse.mjs';
+import { parseMachineSummary, splitEntry, loadReportsIndex } from './reports-index.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
 const APPS_FILE = existsSync(join(CAREER_OPS, 'data/applications.md'))
   ? join(CAREER_OPS, 'data/applications.md')
   : join(CAREER_OPS, 'applications.md');
 const REPORTS_DIR = join(CAREER_OPS, 'reports');
-
-const MACHINE_SUMMARY_FIELDS = new Set([
-  'company',
-  'role',
-  'score',
-  'legitimacy_tier',
-  'archetype',
-  'final_decision',
-  'hard_stops',
-  'soft_gaps',
-  'top_strengths',
-  'risk_level',
-  'confidence',
-  'next_action',
-  // Optional context fields accepted for future reports.
-  'domain',
-  'seniority',
-  'remote',
-  'team_size',
-  // Issue 1380: predicted skip/discard reasons from the agent.
-  'discard_reasons',
-  'advertised_comp',
-  'via',
-  'company_confidential',
-  'risk_summary',
-  // Work-authorization / visa-sponsorship tier from Block A (report + Machine
-  // Summary only). Allowlisted so it round-trips; no consumer logic yet.
-  'work_auth',
-  // Reporting line stated by the JD, verbatim (report + Machine Summary only).
-  // Allowlisted so it round-trips; no consumer logic yet.
-  'reports_to',
-]);
 
 // --- CLI args ---
 const args = process.argv.slice(2);
@@ -158,23 +126,10 @@ function normalizeScalar(value) {
   return null;
 }
 
-function parseMachineSummary(content) {
-  const fenceMatch = content.match(/##\s*Machine Summary\s*\n+```(?:yaml|yml|json)?\s*\n([\s\S]*?)\n```/i);
-  if (!fenceMatch) return null;
-
-  const raw = fenceMatch[1].trim();
-  if (!raw) return null;
-
-  try {
-    const parsed = yamlLoad(raw);
-    if (!parsed || typeof parsed !== 'object' || Array.isArray(parsed)) return null;
-    return Object.fromEntries(
-      Object.entries(parsed).filter(([key]) => MACHINE_SUMMARY_FIELDS.has(key))
-    );
-  } catch {
-    return null;
-  }
-}
+// The canonical Machine-Summary parser + core allowlist now live in
+// reports-index.mjs (imported above). parseMachineSummary returns the FULL
+// object; splitEntry(parsed).summary applies the CORE_SUMMARY_FIELDS filter this
+// script historically relied on.
 
 // --- Via channel analysis (#1596 follow-up) ---
 // Pure: group submitted applications by their Via channel (agency/recruiter
@@ -623,7 +578,10 @@ function readTextIfExists(path) {
 }
 
 // --- Parse a single report file ---
-function parseReport(reportPath) {
+// `index` (optional): a loadReportsIndex() handle. When provided, the Machine
+// Summary comes from the stat-validated cache (no re-parse); the report body is
+// still read for the Block A table, scoring table, URL, and Gap table below.
+function parseReport(reportPath, index) {
   const content = readTextIfExists(reportPath);
   if (content === null) return null;
   const report = {
@@ -647,7 +605,14 @@ function parseReport(reportPath) {
     gaps: [],
   };
 
-  const machineSummary = parseMachineSummary(content);
+  const entry = index ? index.get(reportPath) : null;
+  let machineSummary;
+  if (entry) {
+    machineSummary = entry.summary; // core-filtered (may be null for no-fence reports)
+  } else {
+    const parsed = parseMachineSummary(content);
+    machineSummary = parsed ? splitEntry(parsed).summary : null;
+  }
   if (machineSummary) {
     report.machineSummary = machineSummary;
     report.company = normalizeScalar(machineSummary.company) || report.company;
@@ -927,12 +892,17 @@ function buildPatternSignals(enriched) {
 }
 
 // --- Main analysis ---
-function analyze() {
+function analyze(options = {}) {
   const entries = parseTracker();
 
   if (entries.length === 0) {
     return { error: 'No applications found in tracker.' };
   }
+
+  // Shared, stat-validated Machine-Summary cache over reports/ — one parser and
+  // one on-disk cache across analyze-patterns/upskill/salary-gap. --no-cache
+  // threads through to skip both reading and writing the JSON.
+  const index = loadReportsIndex({ noCache: options.noCache });
 
   // Enrich entries with report data and classification
   const enriched = entries.map(e => {
@@ -950,7 +920,7 @@ function analyze() {
       ]);
       for (const candidate of candidates) {
         if (!withinReports(candidate)) continue;
-        reportData = parseReport(candidate);
+        reportData = parseReport(candidate, index);
         if (reportData) break;
       }
     }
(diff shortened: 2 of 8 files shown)


santifer:
Slow first response on this one, @Scott-Emberson — busy stretch, and it's a bigger read than the queue's average. Thanks for the patience. To the PR:

The `(mtimeMs, size)` key with a **fresh `statSync` on every load** is the design decision that makes this safe, and it's the one I went to check first. A cache that trusts its own record of what it cached is a cache that serves a stale parse forever after one edit it didn't see; re-stat-ing and requiring *both* fields to match means the worst case is a wasted re-read, never a wrong answer. Dropping entries whose file is gone, and the `version` field, close the other two doors.

I also read the `DATA_CONTRACT.md` row rather than assuming it was fine. It sits next to `applications.db` and describes itself the same way — derived, rebuilt by named scripts, safe to delete. That's the right precedent to follow and the right layer to put it in; a derived artifact in the user layer is only acceptable while it's genuinely disposable, and yours says so in the table itself.

Three copies of the same fence regex plus `js-yaml` load, collapsed into one parser, is worth more than the caching: that's the drift class where one copy gets a fix and the other two quietly disagree about what a Machine Summary is.

**Where it stands:** I'm not merging tonight. This touches `DATA_CONTRACT.md` and `update-system.mjs`, and I've hit the merge ceiling I set for a single session — deliberately, so what's left waits for the next cycle instead of riding on end-of-session judgement. That's not a hedge about your PR; it's specifically that a 🔴 file deserves the read at the start of a session rather than the end.

Next session it's near the top. If you want to make that read cheaper: is there any path where a report is written and re-read **within the same millisecond and at the same size** — batch mode writing several reports in one tick, say? That's the only hole I can see in the key, and you'll know faster than I will whether it's reachable.

Scott-Emberson:
Good question to zero in on, and the honest answer is no, the batch-in-one-tick path specifically cannot hit it, for a reason that is upstream of the cache.

Batch never overwrites a report path. `reserve-report-num.mjs` hands each evaluation a distinct number, and `occupiedFromReports` marks any number with a leftover `.md` as occupied, so a released number is never re-handed to different content; a retry releases and re-reserves a fresh number rather than rewriting the old path. So "several reports in one tick" is N different filenames, which is N different cache keys, not one key written twice. Same millisecond is real there, but the paths differ, so the `(mtimeMs, size)` key never aliases.

The only in-place overwrite of an existing report in the tree is `application-answers.mjs`, and it is harmless on two counts: it only upserts the `## Application Answers` section and never touches the `## Machine Summary` fence the index actually caches, and it runs as its own later process, a different mtime tick from any analytics run that seeded the cache, so the size or mtime invalidates it normally. Nothing anywhere flips a Machine Summary field in place at constant size; re-evaluation is always a new number and a new path.

So the key holds for every writer we have today. The one residual is not logical but resolution: on a coarse-granularity filesystem (FAT, some NFS) the mtime tick is wide enough that a future same-path overwriter could in principle slip under it. If you want that closed defensively without giving up the never-re-read win, the cheapest move is to add `ctimeMs` to the key: it comes free from the same `statSync`, and inode ctime bumps on any write even when mtime is coarse or backdated. `mtimeNs` (BigInt) is the other one-line option. A content hash would be strongest but it defeats the whole point of not re-reading unchanged files, so I would keep that for a `--verify` mode only.

I will rebase this onto today's `main` to clear the conflict so it is ready for the start-of-session read. Thanks for holding it for a fresh look rather than an end-of-session one.

santifer:
Four days of silence on a thorough answer, @Scott-Emberson: sorry. Answering now, and the short version is that it settles the question.

The part I checked rather than took on trust is the one your argument leans on: `application-answers.mjs` is in `main`, it exports `APPLICATION_ANSWERS_HEADING = '## Application Answers'`, and the only `writeFileSync` in it rewrites the report after splicing that one section. So "the sole in-place overwrite never touches the `## Machine Summary` fence the index caches" is true of the code as shipped, not just of the intent.

The upstream half is what actually closes it, though, and it's the better argument: batch never overwrites a report path at all, because a retry releases and re-reserves a fresh number instead of rewriting the old one. N reports in one tick are N filenames, so the `(mtimeMs, size)` key has nothing to alias with. Same millisecond stops mattering once the paths differ. I'd rather have a cache whose key can't collide by construction than one that's fast enough to usually win the race.

So: no objection left from me on the invalidation question.

**It's still marked draft**, which is why I haven't taken it further. Is that deliberate (something you still want to change), or leftover? If it's leftover, flip it to ready and I'll review the whole thing properly rather than just the one thread. If there's a piece you're still unsure about, name it and I'll look at that specifically.

Separately, and unrelated to this PR: thank you for how you handled the other thing today. That was the right way to do it, and it landed.

Scott-Emberson:
Thanks, and no leftover: I have flipped it to ready. Two things landed since your note.

Rebased onto current main. The two conflicts were both in the reports-index area: `upskill.mjs` (kept your `validateFlags` import alongside the new reports-index one) and `analyze-patterns.mjs` (the field set moved into `reports-index.mjs`). Worth flagging one deliberate call I made during the rebase: the shared `CORE_SUMMARY_FIELDS` predated `reports_to`, so a plain "take theirs" would have dropped it. I folded `reports_to` into the shared set with its own comment, so that field still round-trips.

CodeRabbit's round is handled, three Major findings, all real, all fixed with tests: reject a symlink escaping `reports/` before any stat or read (both the index build and the on-demand `get`), validate a cache entry's nested shape so one malformed entry forces a full rebuild instead of crashing `salary-gap`, and make `upskill` resolve through the index before any body read so the #2385 read-reduction actually holds. Full suite was green locally.

It reads `BEHIND` rather than dirty, so no conflicts, just main moving under it. Ready for your full pass whenever suits.

santifer:
Thanks for flipping it to ready and rebasing, @Scott-Emberson. I went through it properly this time: 13/13 green, and I re-ran the suite myself on your head SHA in an isolated worktree (4992 passed, 0 failed). Your answer to the invalidation question back on the 14th was thorough enough that I closed that thread myself, and the three follow-ups you added after review (symlink rejection before any stat, per-entry shape validation so one malformed entry drops the cache instead of crashing salary-gap, and the source-level guard in your own self-test that fails if anyone re-inverts the upskill ordering) are exactly the kind of hardening I would have asked for and did not have to.

Two blockers, both reproduced on your head SHA:

**1. `--no-cache` is dead in upskill.mjs.** The PR body says all three scripts gain the flag; two do.

```
$ node upskill.mjs --no-cache
Error: unrecognized flag(s): --no-cache. Valid flags: --min-reports, --summary, --url-text, --self-test, --help, -h
exit 1
```

`KNOWN_FLAGS` (upskill.mjs:907) never got the entry, and `validateFlags` runs at :923, before `const noCache` at :925. This is rebase collision damage: `lib/cli-flags.mjs` landed after you opened this. One line fixes it. It matters more than a flag normally would, because that flag is the correction hatch for the whole caching change, and it is unreachable in the one script whose behaviour changes most.

**2. Silent gap loss in upskill when the Machine Summary carries the keys EMPTY.** The new condition is presence-based (`'hard_stops' in summary || 'soft_gaps' in summary`), and `batch/batch-prompt.md:323` instructs the agent to emit `[]` for those keys when empty. So the canonical format produces the failing case. With a report carrying `hard_stops: []`, `soft_gaps: []`, a score, and a Gap table with two real rows: main's path yields `Kubernetes at scale / Terraform`; the new path yields an empty string. Two real gaps become zero, no error, no failing test, in the one output upskill exists to produce. Your body argues the general case correctly (the Machine Summary IS the structured form of that table), but an empty list is the *absence* of the structured form while the table still has content. Condition on non-empty length after `normalizeList`, not on key presence.

Two smaller notes, neither blocking: `closes #2385` would close a three-item issue on one item (items 1 and 2, the repeated `existsSync` probes and the in-loop `new RegExp`, are untouched), so `refs` plus a follow-up issue is more honest; and the per-report `realpathSync` the symlink guard adds is the right security call but eats part of the syscall saving that motivated #2385, so the numbers in your body predate it and are worth re-measuring.

Fix those two and this merges.

PR #2763, opened 2026-08-12
fix(workday): key the cap-hit warning on entry provenance, not on --since

### LINKED ISSUE #2495: scan.mjs --since gets the wrong max_pages cap warning (sinceMs is no longer a clean proxy after #2418)

maxmilian:
## What does this PR do?

`providers/workday.mjs` picked between its two cap-hit warnings by testing `sinceMs === null`. That was a proxy for the real question — *"does this caller have a `portals.yml` `tracked_companies` entry the user could edit?"* — and since #2418 the proxy no longer holds, so a `scan.mjs --since` run loses the `raise max_pages on this entry for more` advice on a tracked company. This states provenance explicitly instead of inferring it: `scan-ats-full.mjs` sets `ctx.syntheticEntries`, and the message is keyed on that.

## Related issue

Fixes #2495

## Reproducing it

Driving `workday.fetch()` directly with a mock transport, one `tracked_companies` entry with `max_pages: 3`, `total: 200`, every posting `Posted Today` (so the `--since` early-stop never pre-empts the cap). The **only** difference between the two runs is `ctx.sinceMs`:

| run | warning | actionable? |
|---|---|---|
| `node scan.mjs` (`sinceMs = null`) | `⚠️  workday: AcmeCorp truncated at max_pages=3 (60 of 200 jobs) — raise max_pages on this entry for more` | ✅ |
| `node scan.mjs --since 7d` (`sinceMs = now-7d`) | `⚠️  workday: AcmeCorp truncated at 3 pages (60 of 200 jobs)` | ❌ |

After this change both runs emit the first line, and a reverse scan still gets the second.

You mentioned on the issue that you don't mark anything `confirmed` until you've reproduced the exact branch — this is a provider-level reproduction against a mock transport, not a live tenant, so it pins the branch rather than the field report. Happy to adjust if you'd rather see it framed differently.

## Why the change touches a second file

`workday.mjs` on its own can only *infer* provenance, which is the bug. The one caller that both synthesizes entries and can reach the cap has to say so — `scan-ats-full.mjs` is a one-line `ctx` addition. `discover-ats.mjs` and `verify-portals.mjs` also call `workday.fetch()` with synthesized/tracked entries, but both probe with `ctx.maxPages: 1`, which never sets `stopReason` to `'cap'`, so neither needs to opt in. Absence of the flag therefore means "tracked", which keeps the next caller that starts setting `sinceMs` from re-coupling the two concerns — the shape you and @LayerC0de both landed on in the issue.

## Before / after

```diff
-    if (sinceMs === null) {
+    const syntheticEntries = ctx?.syntheticEntries === true;
+    if (!syntheticEntries) {
```

Both message strings are unchanged; only the condition selecting between them moved. The `(total may be Workday-capped, not real)` tag is about the tenant's `total`, not about provenance, so it stays on the terse line — pinned by a test so the rekey can't quietly drop it.

## Tests

Four assertions in `tests/providers/workday.test.mjs`, pinning which warning fires for which caller:

- a full `scan.mjs` run gets the advice
- a `scan.mjs --since` run **keeps** the advice ← fails on `main`, passes here
- a reverse scan (synthesized entries) stays terse, no advice
- the Workday-capped-total tag survives on the reverse-scan line

Three existing tests encoded the old proxy (they simulated "reverse-scan context" with `sinceMs` alone); they now declare `syntheticEntries: true`, which is what they were always describing in their comments.

`node test-all.mjs`: **3434 passed, 0 failed** (3430 on `main` — the +4 are the new assertions). The single warning is the pre-existing `cv-sync-check.mjs exited with error (expected without user data)`, present on `main` too.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pagination-limit warnings for tracked job listings, including clearer guidance to increase the maximum page limit when results may be incomplete.
  * Refined warnings for externally sourced listings to provide more relevant messaging without suggesting portal-specific changes.
  * Preserved existing indicators when Workday reports a capped total of available results.

### DIFF
--- a/providers/workday.mjs
+++ b/providers/workday.mjs
@@ -168,7 +168,7 @@ export default {
    * as providers/glints.mjs's firewall).
    *
    * @param {{ name?: string, api?: string, careers_url?: string, max_pages?: number }} entry
-   * @param {{ fetchJson: (url: string, opts?: object) => Promise<any>, sinceMs?: number, maxPages?: number }} ctx
+   * @param {{ fetchJson: (url: string, opts?: object) => Promise<any>, sinceMs?: number, maxPages?: number, syntheticEntries?: boolean }} ctx
    * @returns {Promise<Array<{title: string, url: string, company: string, location: string, postedAt?: number}>>}
    */
   async fetch(entry, ctx) {
@@ -283,15 +283,22 @@ export default {
     // portal entry to point at, and no fixed cap can guarantee full coverage of
     // an unbounded company directory anyway; nothing else to suggest there.
     //
-    // The branch below keys on `sinceMs === null` as a proxy for that
-    // distinction. Since #2418 the proxy is no longer exact: `scan.mjs --since`
-    // also sets ctx.sinceMs, so those runs take the else branch and get the
-    // terser message even though they DO have a portals.yml entry to raise.
-    // Harmless (the cap is still surfaced, just without the suggestion) but
-    // worth knowing before trusting the proxy — see #2495.
+    // The branch below used to key on `sinceMs === null` as a proxy for that
+    // distinction, which held only because scan-ats-full.mjs was the sole
+    // caller setting it. #2418 broke the proxy — `scan.mjs --since` sets
+    // ctx.sinceMs too, so a tracked entry lost the actionable half of the
+    // message on every --since run (#2495). Provenance is now stated by the
+    // caller instead of inferred from an unrelated flag, so a future caller
+    // that starts setting sinceMs cannot re-couple the two concerns.
+    //
+    // Absence means "tracked": scan-ats-full.mjs is the only caller that
+    // synthesizes entries AND can reach the cap (discover-ats.mjs and
+    // verify-portals.mjs both probe with ctx.maxPages: 1, which never sets
+    // stopReason to 'cap'), so it is the one place that opts out.
+    const syntheticEntries = ctx?.syntheticEntries === true;
     if (stopReason === 'cap') {
       const jobsSummary = `${jobs.length}${total !== null ? ` of ${total}` : ''} jobs`;
-      if (sinceMs === null) {
+      if (!syntheticEntries) {
         console.error(`⚠️  workday: ${entry.name} truncated at max_pages=${maxPages} (${jobsSummary}) — raise max_pages on this entry for more`);
       } else {
         // Workday's CXS backend can report `total` as exactly
--- a/scan-ats-full.mjs
+++ b/scan-ats-full.mjs
@@ -641,7 +641,14 @@ async function main() {
   // sinceMs once postings are confidently past the --since window, and
   // includeUndated (when false) for a tenant that exposes no postedOn at
   // all, since its postings would all be dropped as undated below anyway.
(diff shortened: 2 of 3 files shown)


santifer:
Merged, @maxmilian: replacing the sinceMs proxy with an explicit provenance signal is the right shape, because the proxy was answering a different question than the one the warning asks. Verified with an independent mock-transport harness rather than by reading. 🚀

PR #2765, opened 2026-08-12
fix(followup-seed): resolve "today" in local time, not UTC

davidnunez:
Follow-up to #2740, which fixed the same idiom in `agent-inbox.mjs`. Independent change,
independent branch off `main`.

### The bug

```js
function todayStr() {
  return new Date().toISOString().slice(0, 10);
}
```

That is the UTC day. Anywhere west of Greenwich an evening run answers "today" with tomorrow —
after 20:00 US Eastern it returns the next calendar day.

Two call sites consume it, and both matter:

1. **The fallback applied date** (`resolveAppliedDate`), used when a tracker row carries no
   `Applied YYYY-MM-DD` note. It feeds `addDays(parseDate(appliedDate), cadence.applied_first)`,
   so the entire follow-up schedule is computed from a day that has not happened yet.
2. **The `(set …)` stamp** on the pin written to `data/follow-ups.md`.

An application marked Applied on a US evening — a common time to be doing this — gets its
first follow-up date a day early, silently.

### Scope

Only "what day is it here" changes. **Date arithmetic is untouched.** `parseDate()` and
`isValidCalendarDate()` in `followup-cadence.mjs` round-trip through UTC midnight on purpose,
and that is internally consistent: they operate on explicit `YYYY-MM-DD` strings, never on the
ambient clock. Mixing the two concerns is what makes this class of bug hard to see, so the fix
deliberately separates them rather than converting everything.

### Why it survived — the test had the same bug

`followup-seed-tests.mjs` defined its own `todayStr()` with the identical `toISOString()`
idiom, and tests 1 and 4 asserted the seed's "today" against it. Both sides were wrong in the
same direction, so they agreed and passed.

They were also latently flaky: every US evening after 20:00 Eastern the two dates diverge and
both assertions go red. The helper now mirrors the local-date contract it is testing.

### The new test

Test 18 seeds under two zones **26 hours apart** — `Etc/GMT-14` (UTC+14) and `Etc/GMT+12`
(UTC-12) — whose calendar days can never both equal UTC's. It asserts the `(set …)` stamps
differ, and that each matches its own zone's day. `--date` pins the applied date so only the
stamp under test can vary between runs.

A UTC stamp is identical in both zones, so this cannot pass by accident of when it runs.
Asserting against the host's own today would go green trivially on CI, where `TZ` is UTC —
which is precisely how the original slipped through.

Verified failing against the unfixed source:

```
FAIL 18. (set …) is the same in UTC+14 and UTC-12 — still stamping UTC (2026-08-12 vs 2026-08-12)
FAIL 18. (set …) for Etc/GMT-14 — expected 2026-08-13, got 2026-08-12
FAIL 18. (set …) for Etc/GMT+12 — expected 2026-08-11, got 2026-08-12
43 passed, 3 failed
```

With the fix: `node followup-seed-tests.mjs` → **46 passed, 0 failed**.
`node test-all.mjs` on a clean checkout of this branch → **3576 passed, 0 failed**, 2 warnings
(environmental: no user `cv.md`, no Go compiler for the dashboard build).

### On the wider pattern

The `new Date().toISOString()` idiom appears in a dozen or so other files here. Most are
correct as-is — vCard `REV:` in `contacts.mjs` is specified as UTC, and lock and sentinel
timestamps have no local meaning. I have not swept them, and would rather fix the ones with
user-visible date semantics case by case than churn the lot.


santifer:
Welcome to career-ops, @davidnunez, and this is a strong first contribution: the claim reproduces exactly as you describe (clock frozen at 2026-08-12T11:30Z with TZ=Pacific/Apia, the pin lands on the wrong day without your patch and on the right one with it), and retiring #2740 yourself within minutes when you spotted your own mistake is the kind of thing I notice.

One thing to resolve before it lands, and it is a consequence of the fix being correct.

**After this, followup-seed resolves "today" locally while its neighbours still resolve it in UTC.** Measured at the same frozen instant in UTC+13: `set-status.mjs` and `tracker.mjs` say 2026-08-12, followup-seed says 2026-08-13. Before your patch they were consistently wrong; now one is right and the rest are not, and `set-status.mjs:537` points explicitly at follow-up seeding, so they meet inside one user flow. A user in New Zealand seeds a follow-up dated tomorrow and then gets "in the future" rejected by set-status.

Your call to keep the PR small was right, so I am not asking you to sweep all six call sites here. What I am asking for is one of these two:

- **preferred:** open a follow-up issue listing the UTC holdouts (`set-status.mjs:164` and `:502`, `tracker.mjs:305`, `application-answers.mjs:45`, `assessment-log.mjs:140`, `check-table-freshness.mjs:554`) so the inconsistency is tracked rather than discovered by a user, and say in this PR's body that it is deliberate scope; or
- do the sweep here if you would rather land it in one piece.

Either works. Say which and this goes in.

santifer:
Test 18 is what makes this more than a one-line fix, @davidnunez. Two zones 26 hours apart means their wall-clock dates can never both equal UTC's — so "the two differ" is a complete assertion that cannot pass by accident of when CI happens to run. Asserting against the host's own today would have gone green trivially in a UTC runner, which is precisely how the bug survived: two existing assertions did exactly that.

And your note on the helper is the sharper observation:

> The bug and its test were the same mistake.

`todayStr()` in the test file used `toISOString()` too, so tests 1 and 4 compared the seed's local today against UTC's today. They agree until 20:00 Eastern and then both go red — meaning the test suite would have started failing on its own, at a time of day, for a bug that had been there all along. A test that shares the defect it's meant to catch is worse than no test, because it reports confidence.

Pinning `--date` so `nextDate` is fixed and only the `(set …)` stamp can vary between the two runs is the right isolation. One thing changes, one thing is asserted.

Merged 🚀

PR #2746, opened 2026-08-12
fix(dedup-tracker): reject unrecognized flags instead of silently live-running

Schlaflied:
## Summary

Closes #2744.

`dedup-tracker.mjs` only checked for `--dry-run` via `process.argv.includes('--dry-run')`. Any other flag — including a plausible typo like `--check` — was silently ignored, `DRY_RUN` stayed `false`, and the script ran its real, destructive merge-and-write pass against the live tracker (`data/applications.md`) with no warning. This happened for real: someone typed `--check` expecting a dry-run/preview and instead got a live write, with a `.bak` backup file as the only safety net.

This is the same failure shape already fixed twice in this repo:
- `scan-ats-full.mjs` — #1633 / PR #1635
- `reply-watch.mjs` — #2743 (PR in flight)

This PR applies the identical pattern to `dedup-tracker.mjs`, scoped to just that file:
- `KNOWN_FLAGS = ['--dry-run', '--help', '-h']`
- `--help` / `-h` prints usage and exits 0, before any dedup logic runs
- Any `-`-prefixed arg not in `KNOWN_FLAGS` prints `Error: unrecognized flag(s): ...` and exits 1, before any dedup/write logic runs
- `--dry-run` and the no-flags (live-run) behaviors are unchanged

A maintainer comment was already posted on #2744 asking santifer's opinion on extracting a shared flag-validation helper now that this is a third instance of the same class — that's an open follow-up question, not something this PR builds.

## Test plan

- [x] Added coverage in `test-all.mjs` (same `run()` + `CAREER_OPS_TRACKER` fixture convention as the existing dedup-tracker tests): `--help`/`-h` print usage, exit 0, and do not touch the tracker; `--check` (the #2744 repro) errors, exits 1, and — critically — does NOT reach the write path (tracker file byte-identical before/after); `--dry-run` and no-flags (real run, exact-duplicate merge) regression-tested unchanged.
- [x] `node test-all.mjs --quick` — 3436 passed, 0 failed, 1 pre-existing/unrelated warning (`cv-sync-check.mjs exited with error (expected without user data)`)
- [x] Manually verified the exact issue repro: `node dedup-tracker.mjs --check` now prints `Error: unrecognized flag(s): --check. Valid flags: --dry-run, --help, -h` and exits 1, without touching `data/applications.md`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added command-line help via `--help` and `-h`, including usage information.
[...]

Issue #2744, opened 2026-08-12
fix(dedup-tracker): unrecognized flags (e.g. --check) silently fall through to a real, destructive dedup pass


Schlaflied:
## Bug

`dedup-tracker.mjs` only recognizes `--dry-run`, checked via `process.argv.includes('--dry-run')`, with no rejection of anything else:

```js
const DRY_RUN = process.argv.includes('--dry-run');
```

Any other flag — including a plausible-sounding one like `--check` — is silently ignored. `DRY_RUN` stays `false`, and the script runs its real, destructive merge-and-write pass against the live tracker instead of failing fast or falling back to a safe default.

Repro (happened for real while I was inspecting tracker state, not intentionally testing this):
```
node dedup-tracker.mjs --check
# no error, no dry-run notice — it silently performed a REAL dedup pass
# and wrote to data/applications.md (with a .bak backup, at least)
```

The backup file (`data/applications.md.bak`) is the only reason this wasn't worse. There's no confirmation prompt, no "did you mean --dry-run?", nothing — a mistyped or misremembered flag on a script that mutates the user's real, hand-maintained tracker just runs the live write path silently.

## Why this matters

Third occurrence of the same failure class in this repo:
- `scan-ats-full.mjs` — fixed in #1633/PR #1635
- `reply-watch.mjs` — fixed in #2743 (PR incoming)
- `dedup-tracker.mjs` — this issue

Same root cause each time: no allowlist of known flags, so an unrecognized one falls through to default/live behavior instead of erroring. This one is the highest-stakes instance so far, since the "default behavior" it silently falls through to is a real write to `data/applications.md`, not a read-only scan or a mock-file creation.

## Proposed fix (scoped to this file, same shape as #1633/#2743)

```js
const KNOWN_FLAGS = ['--dry-run', '--help', '-h'];

if (process.argv.includes('--help') || process.argv.includes('-h')) {
  console.log('Usage: node dedup-tracker.mjs [--dry-run]');
  process.exit(0);
}

const unknownFlags = process.argv.slice(2).filter(a => a.startsWith('-') && !KNOWN_FLAGS.includes(a));
if (unknownFlags.length > 0) {
  console.error('Error: unrecognized flag(s): ' + unknownFlags.join(', ') + '. Valid flags: ' + KNOWN_FLAGS.join(', '));
  process.exit(1);
}

const DRY_RUN = process.argv.includes('--dry-run');
```

[...]

PR #2778, opened 2026-08-12
feat(cli): shared flag-validation helper, migrate 3 fixed scripts, fix scan.mjs (#2270)

### LINKED ISSUE #2270: scan.mjs: no --help and unknown flags are silently ignored, so `node scan.mjs --help` runs a full scan and writes to pipeline.md
## Symptom

`scan.mjs` has no `--help`, and it does not reject unknown flags. So a user (or an agent) probing for usage runs a **full scan** instead of getting help:

```
$ node scan.mjs --help
Scanning 29 companies; 0 local parser; 6 skipped — no provider matched via providers
...
Companies scanned:     29
Total jobs found:      3371
New offers added:      3
Results saved to data/pipeline.md and data/scan-history.tsv
```

The flag is silently ignored, ~3,371 jobs are fetched across 29 companies, and **two user-layer data files are mutated** (`data/pipeline.md` gains new checkbox rows, `data/scan-history.tsv` gains dedup entries). Nothing about the invocation suggested a write.

Same applies to any typo: `--dryrun`, `--dry_run`, `--verbose`, `--posted-since` all run an unfiltered live scan rather than erroring.

## Why it matters

`--help` is the near-universal way to discover a CLI's interface, and it is the one invocation a user reasonably assumes is side-effect-free. Here it is indistinguishable from a bare `node scan.mjs`. For an agent driving the tool this is worse: probing for capabilities silently appends rows to the user's pipeline, and those rows then look like intentional scan results.

The typo case is the more dangerous one. `--dry-run` is a real flag whose entire purpose is "don't write anything". A user who types `--dryrun` gets the opposite of what they asked for, with no error and no indication the flag was dropped.

## Cause

`scan.mjs:1707` collects args and tests only for the flags it knows:

```js
const args = process.argv.slice(2);
const dryRun = args.includes('--dry-run');
const verify = args.includes('--verify');
...
```

There is no `--help` branch and no unknown-token check, so anything unrecognized falls through to a normal scan.

Notably the file already establishes the right pattern for a different class of bad input — `--posted-after` / `--posted-before` validate their argument and `process.exit(1)` on a malformed date, precisely so a silently-ignored bound does not masquerade as "no jobs matched". The same reasoning applies one level up, to the flag names themselves.

## Suggested shape

1. Add a `--help` / `-h` branch that prints usage and exits 0 **before** any provider loading, portals read, or network call.
2. Reject unknown `--`-prefixed tokens with a usage hint and exit 1, e.g.:
   ```js
   const KNOWN = new Set(['--dry-run', '--verify', '--headed-fallback', '--throttle',
     '--rediscover-404', '--include-blacklisted', '--company', '--posted-after',
     '--posted-before', '--quiet', '--resume', '--help', '-h']);
   const unknown = args.filter(a => a.startsWith('-') && !KNOWN.has(a.split('=')[0]));
   if (unknown.length) {
     console.error(`Error: unknown option(s): ${unknown.join(', ')}\nRun: node scan.mjs --help`);
     process.exit(1);
   }
   ```
   (`--company`, `--posted-after`, `--posted-before` take a following value, so the filter should skip consumed operands or the check should run on flag tokens only.)
3. Worth a sweep across the other root-level CLIs for the same gap — `scan-ats-full.mjs`, `check-liveness.mjs`, `merge-tracker.mjs` and friends are all invoked the same way from the mode docs.

Optional but valuable: have the run banner state up front that the scan writes to `data/pipeline.md` and `data/scan-history.tsv`, so the side effect is visible before it happens rather than only in the closing line.

### LINKED ISSUE #2775: feat(cli): shared flag-validation helper — de-dupe the unrecognized-flag-falls-through bug fixed independently 4 times
## Problem

The same failure shape has now been independently fixed in three different scripts, and is still live in at least one more:

- `scan-ats-full.mjs` — fixed in #1633 / PR #1635
- `reply-watch.mjs` — fixed in #2743 / PR #2745
- `dedup-tracker.mjs` — fixed in #2744 / PR #2746
- `scan.mjs` — **still open**, #2270 (`--help` and unknown flags are silently ignored, so `node scan.mjs --help` runs a full scan and writes to pipeline.md)

Each time, the root cause is identical: no shared allowlist of known flags, so an unrecognized (or mistyped) flag falls through to whatever the script's default positional/live behavior is, instead of failing fast. The blast radius varies by script — a stray mock file, a live write to `data/applications.md`, a full network scan — but the bug itself is the same 10-15 lines re-derived and re-fixed independently every time it's discovered.

@santifer flagged this repo already has a shape worth mirroring in #2636 (`--help` validated before data resolution, plus a regression test on documented flags) — claiming this issue per that conversation on #2744.

## Proposal

A small, shared `validateFlags(argv, knownFlags, { usage })` helper (or similarly named) living in one of the existing shared-utils files (`tracker-utils.mjs` is the wrong home since not every CLI script touches the tracker — open to whichever shared location best fits this repo's existing conventions), doing exactly the three things every one of the four fixes above does by hand:

1. If `--help`/`-h` is present, print the caller-supplied usage string and exit 0.
2. If any `-`-prefixed arg isn't in the known-flags allowlist, print `Error: unrecognized flag(s): ...` naming them and the valid set, and exit 1.
3. Return the remaining positional args to the caller.

Each script keeps its own `KNOWN_FLAGS` list and usage string (they're not identical across scripts), it just stops hand-rolling the allowlist-check-and-error boilerplate independently every time.

## Scope

- Fix `scan.mjs` (#2270) using the new shared helper, since it's the one known live instance not yet patched.
- Migrate `scan-ats-full.mjs`, `reply-watch.mjs`, and `dedup-tracker.mjs` to the shared helper as a mechanical refactor (behavior-preserving — these are already fixed and tested, this is just de-duplicating the implementation, not changing behavior).
- Leave every other script alone unless/until it's independently found to have this bug — not proposing a repo-wide sweep, just closing the loop on the four already-known instances plus the one open one.

Not building this without a maintainer nod on the shared-helper location/shape first, per this repo's usual process.

Schlaflied:
Closes #2775. Closes #2270.

## What this is

Three scripts (`scan-ats-full.mjs` #1633/#1635, `reply-watch.mjs` #2743/#2745, `dedup-tracker.mjs` #2744/#2746) independently hand-rolled the identical fix for the identical bug: an unrecognized/mistyped CLI flag silently falling through to default/live behavior instead of failing fast. `lib/cli-flags.mjs` gains `validateFlags(args, knownFlags, usage, { valueFlags })` — the shape all three converged on, in one place:

- Any `-`-prefixed arg not in `knownFlags` errors and exits 1 — checked **before** `--help`/`-h` is honored (the ordering CodeRabbit flagged as a bug on both #2745 and #2746: `--help --bogus` must still error, not exit 0 having never looked at `--bogus`).
- `--help`/`-h` prints the caller-supplied usage and exits 0.
- An optional `valueFlags` list (mirroring `scan-ats-full.mjs`'s existing adjacency rule) keeps a value-taking flag's own value — e.g. the `-5` in `--since -5` — from being misread as an unrecognized flag.

## Migration (behavior-preserving)

`scan-ats-full.mjs`, `reply-watch.mjs`, and `dedup-tracker.mjs` now call the shared helper instead of their own copies. Same error messages, same exit codes, same usage text — this is de-duplication, not a behavior change. (`scan-ats-full.mjs` picks up one small fix as a side effect: its own `--help`-before-unknown-flag ordering had the same latent bug the other two were caught on, just never exercised by a test that combined both.)

## The one genuinely new fix: `scan.mjs` (#2270)

`node scan.mjs --help` never checked `--help` at all — it fell straight through into a full live scan, hitting every configured ATS and writing to `pipeline.md`/`scan-history.tsv`. Same failure class as the other three, fixed the same way. `KNOWN_FLAGS` was built from the script's own existing docblock usage examples; `--company`/`--posted-after`/`--posted-before`/`--since` are registered as `valueFlags` (`--throttle` deliberately isn't — it's only ever read in its bare or `--throttle=<ms>` forms).

## Tests

- `tests/cli-flags.test.mjs` extended with direct unit coverage of `validateFlags` (unrecognized-flag rejection, `--help`/`-h`, the ordering case, `valueFlags` negative-number handling) plus subprocess-level coverage for cases that need a real exit code.
- New `tests/scan-help-flag.test.mjs`: hermetic (`CAREER_OPS_PORTALS` pointed at a nonexistent path) coverage proving `--help`, `-h`, and an unrecognized flag on `scan.mjs` never reach the portals-loading step — i.e. never attempt a live scan.
- `node test-all.mjs --quick`: 3647 passed, 0 failed.

## Summary by CodeRabbit

* **New Features**
  * Added consistent command-line flag validation across scanning and monitoring commands.
  * Added support for value-based flags, including negative values.
  * Expanded scan help documentation with available options.

* **Bug Fixes**
  * Unrecognized or mistyped flags now fail clearly before command execution.
  * Help options consistently display usage information and exit successfully.

* **Tests**
  * Added coverage for valid flags, invalid flags, help aliases, value handling, and validation order.

### DIFF
--- a/dedup-tracker.mjs
+++ b/dedup-tracker.mjs
@@ -17,6 +17,7 @@ import {
   openTrackerTransaction, rebuildRow, resolveTrackerPath, normalizeCompany,
 } from './tracker-utils.mjs';
 import { resolveColumns, parseTrackerRow, normalizeVia } from './tracker-parse.mjs';
+import { validateFlags } from './lib/cli-flags.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
 // Support both layouts: data/applications.md (boilerplate) and applications.md
@@ -27,22 +28,14 @@ const APPS_FILE = resolveTrackerPath(CAREER_OPS);
 // ── CLI args ────────────────────────────────────────────────────────
 // Same shape as scan-ats-full.mjs (#1633/PR #1635) and reply-watch.mjs
 // (#2743): an unrecognized flag must fail fast, never silently fall through
-// to the live-run default and write to the real tracker (#2744).
+// to the live-run default and write to the real tracker (#2744). Shared via
+// lib/cli-flags.mjs's validateFlags() (#2775).
 const KNOWN_FLAGS = ['--dry-run', '--help', '-h'];
 const USAGE = `Usage: node dedup-tracker.mjs [--dry-run]`;
 
 const cliArgs = process.argv.slice(2);
 
-const unknownFlags = cliArgs.filter(a => a.startsWith('-') && !KNOWN_FLAGS.includes(a));
-if (unknownFlags.length) {
-  console.error(`Error: unrecognized flag(s): ${unknownFlags.join(', ')}. Valid flags: ${KNOWN_FLAGS.join(', ')}`);
-  process.exit(1);
-}
-
-if (cliArgs.includes('--help') || cliArgs.includes('-h')) {
-  console.log(USAGE);
-  process.exit(0);
-}
+validateFlags(cliArgs, KNOWN_FLAGS, USAGE);
 
 const DRY_RUN = process.argv.includes('--dry-run');
 
--- a/lib/cli-flags.mjs
+++ b/lib/cli-flags.mjs
@@ -1,13 +1,21 @@
 /**
- * cli-flags.mjs — reading a value-taking CLI flag in BOTH accepted forms.
+ * cli-flags.mjs — shared CLI argv helpers.
+ *
+ * Two related but distinct defects, both born from scripts hand-rolling their
+ * own argv parsing and independently getting it wrong the same way:
  *
  * `args.indexOf('--flag')` returns -1 for `--flag=value`, so a lookup written
  * that way silently DISCARDS the value the caller supplied and the script runs
  * with its default instead — reporting a result for inputs nobody asked for,
  * with no warning. That is the defect #2401 described for weekly-digest
  * (`--from=…` digesting the wrong week) and #2402 fixed there and in
- * company-history.mjs; this module is the same semantics in one place, so the
- * next script does not have to rediscover it.
+ * company-history.mjs. `flagValue`/`hasFlag` are the shared fix.
+ *
+ * An unrecognized or mistyped flag (`--dryrun` instead of `--dry-run`)
+ * silently falls through to default/live behavior instead of failing fast —
+ * fixed independently, in the identical shape, in scan-ats-full.mjs
+ * (#1633/#1635), reply-watch.mjs (#2743/#2745) and dedup-tracker.mjs
+ * (#2744/#2746). `validateFlags` is that shape in one place (#2775).
  */
 
 /**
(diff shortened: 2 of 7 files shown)


santifer:
The header comment on `lib/cli-flags.mjs` is the review: two defects, each fixed independently three or four times in the identical shape, in different files, by different people. That is the signature of something that should have been one function, and naming both classes in one place is worth more than the code under it.

Verified both halves by running them rather than reading them.

**`validateFlags`**, on `scan.mjs`:

```
main    $ node scan.mjs --help    → exit 1: "portals.yml not found. Run onboarding first."
branch  $ node scan.mjs --help    → exit 0, usage printed, before any file is touched
main    $ node scan.mjs --bogus   → exit 1: "portals.yml not found"   (flag ignored entirely)
branch  $ node scan.mjs --bogus   → exit 1: "unrecognized flag(s): --bogus. Valid flags: --dry-run, --verify, …"
```

The second pair is the one that matters. On `main` a typo and a missing config produce **the same message**, so someone who typed `--dryrun` is sent to fix their onboarding. Listing the valid flags in the rejection turns a dead end into a hint, which is a small thing that will save more time than the fix itself.

**`flagValue`**, the `--flag=value` half:

| argv | result |
|---|---|
| `["--from", "2026-01-01"]` | `"2026-01-01"` |
| `["--from=2026-01-01"]` | `"2026-01-01"` ← the form that used to be discarded |
| `["--from="]` | `""` |
| `["--other", "x"]` | `undefined` |

`scan.mjs` is on the frozen core/web surface, so §55 ran: green, including the `scan-history.tsv` 7-column prefix. Suite on a preview against today's main: 3820 passed, 0 failed.

**One consequence beyond this PR.** I seeded five `good first issue`s tonight for scripts that still hand-roll this (`classify-tier`, `funnel-velocity`, `invite-match`, `check-table-freshness`, `doctor`), all pointing at `company-history.mjs` as the pattern to copy. With this merged, that guidance is now the old way, so I am rewriting them to point at `validateFlags` instead. Your PR turns five upcoming first contributions from "copy this shape carefully" into "call this function", which is a much better first PR to have.

Merged. 🚀

Issue #2777, opened 2026-08-12
agent-inbox: concurrent adds still lose an item on Windows (appendFileSync is not atomic there)


santifer:
**Priority: P0** — this currently reds the Windows leg of CI for every open PR.

## What happens

`agent-inbox.mjs` `add` uses `appendFileSync` (`agent-inbox.mjs:139`). Under concurrency that is atomic enough on Linux and macOS for small writes, and **not on Windows**. With 30 concurrent writers, one write is lost:

```
7. concurrent adds do not lose items (append, not rewrite)
   ✅ every concurrent add exited cleanly
   ❌ all 30 concurrently queued items survive — kept=29 of 30
```

`item-5` never lands. Reproduced consistently across three separate `test (windows-latest)` runs (`31628034171`, `31627544537`, `31625710792`).

## Why it looks new

The test that catches it shipped in #2614, which fixed the same class of loss on the other platforms. The PR's own CI was green when it merged: a race that drops one write in thirty passes often enough to look healthy once. That is a maintainer miss, not a contributor one, and the credit for the diagnosis belongs to @darkpandawarrior.

## Why it matters beyond CI

The agent inbox is a queue that other tools append to. A silently dropped item means a request that was accepted and never executed, with nothing anywhere saying so. That is the same failure shape #2614 set out to remove.

## Proposed fix

Route the append through the existing shared lock rather than inventing a second mechanism:

- `pipeline-lock.mjs` already exports `withPipelineLock`, and #2639 used it today for the same class of problem on `data/scan-history.tsv` (`scan.mjs:1813`).
- Wrap the `add` write in `agent-inbox.mjs:139` the same way.

Acceptance:
- the existing concurrency test passes on **windows-latest**, not just on Linux and macOS
- no behavior change on the other platforms (the suite stays green there)
- no second lock implementation: it uses `withPipelineLock`, since a second one would be exactly the duplication that drifts

## Interim

Not skipping the test on Windows. A skipped test here would hide a real data-loss path on a platform we support, and the red is doing its job by telling us the bug is still there.

santifer:
Fixed in `a7f65b6a`, on main.

The append now goes through `withPipelineLock` (`pipeline-lock.mjs`), the same lock `scan.mjs` already uses for scan-history appends, instead of a second mechanism. The last-byte check moved inside the lock too: reading it outside races with another writer between the check and the write, which would have left a narrower version of the same bug.

[...]

PR #2782, opened 2026-08-12
fix(role-matcher): non-Latin role titles tokenize to nothing, so re-evaluations duplicate the row

### LINKED ISSUE #2781: role-matcher: non-Latin role titles tokenize to nothing, so a re-evaluation is appended as a duplicate row
### What

`roleTokens` strips with `[^a-z0-9\s]`, so **every non-Latin role title tokenizes to `[]`**:

```
Backend Engineer, Payments    ["backend","engineer","payments"]
Бэкенд-разработчик, платежи   []
सॉफ्टवेयर इंजीनियर             []
バックエンドエンジニア          []
```

`merge-tracker`'s dedup uses `roleFuzzyMatch`, so two spellings of the **same** role at the same company never match. The Latin case merges; the non-Latin one is appended as a second row:

```
Backend Engineer, Payments  vs  Backend Engineer (Payments)   -> match
Бэкенд-разработчик, платежи vs  Бэкенд-разработчик (платежи)  -> NO match
```

A re-evaluation of a role you already track becomes a duplicate tracker row instead of an update — the mirror image of #2429, where the company key collapsed *too much*. Here the role key collapses too little.

### Second defect in the same function

`normalizeTitle` strips **every** `\p{Mn}` after NFD. That's correct for Latin accents (the `Sênior` case it was written for), but it also reaches marks that are load-bearing elsewhere:

```
कंपनी  -> कपनी      (Devanagari matra removed; कंपनी and कपनी become one token)
Йогурт -> иогурт     (Cyrillic breve; Й folded onto И)
バックエンド -> ハックエント  (Japanese dakuten; voiced kana folded onto unvoiced)
```

Same class as `5df43e7` — a fold scoped by "strip all marks" rather than by "strip the marks this function is actually for."

### Fix

Two narrow changes, both keeping the Latin path byte-identical:

1. `[^a-z0-9\s]` → `[^\p{L}\p{M}\p{N}\s]`. `\p{M}` matters: without it Devanagari matras become spaces and shatter the word into fragments.
2. `normalizeTitle` strips `\p{Mn}` **only after an ASCII base** (`(?<=[a-z])\p{Mn}`). Latin accents always follow one once lowercased, so `Sênior` still folds; matras, breve and dakuten do not.

This is what `normalizeTextKey`'s own docstring anticipates — *"`separator` exists because not every consumer wants a solid key… so a second private `[a-z0-9]` strip never has to exist"*. `role-matcher` is that second private strip, though it can't delegate wholesale because it needs the Latin accent-folding the shared key deliberately omits.

### Out of scope, flagged

CJK still tokenizes poorly: the `w.length > 3` filter assumes word-separated text, and `決済` is a two-character word. Fixing that needs segmentation, not a character class, so I've left it — Cyrillic, Greek, Devanagari and other space-separated scripts are fixed here.

PR incoming.

abankar1:
Closes #2781.

`roleTokens` strips with `[^a-z0-9\s]`, so every non-Latin role title produces an empty token set:

```
Backend Engineer, Payments    ["backend","engineer","payments"]
Бэкенд-разработчик, платежи   []
सॉफ्टवेयर इंजीनियर             []
```

`merge-tracker`'s dedup runs through `roleFuzzyMatch`, so two spellings of the **same** role at the same company never match — the Latin one merges, the non-Latin one is appended as a duplicate row:

```
                                                        BEFORE   AFTER
Backend Engineer, Payments / Backend Engineer (Payments)  match   match
Бэкенд-разработчик, платежи / … (платежи)                 NO      match
सॉफ्टवेयर इंजीनियर, भुगतान / … (भुगतान)                    NO      match
```

Mirror image of #2429 — there the company key collapsed too much; here the role key collapses too little.

## Second defect in the same function

`normalizeTitle` strips **every** `\p{Mn}` after NFD. Correct for the Latin accents it exists for, wrong for marks that carry meaning:

```
कंपनी      -> कपनी          Devanagari matra removed — two words become one token
Йогурт     -> иогурт        Cyrillic breve — Й folded onto И
バックエンド -> ハックエント    Japanese dakuten — voiced kana folded onto unvoiced
```

Same class as `5df43e7`: a fold scoped by *"strip all marks"* rather than by *"strip the marks this function is for."*

## Fix

Two narrow changes, Latin path byte-identical:

1. `[^a-z0-9\s]` → `[^\p{L}\p{M}\p{N}\s]`. **`\p{M}` is load-bearing** — I first widened it without, and Devanagari matras became spaces that shattered `सॉफ्टवेयर` into `स फ टव यर`. Caught it by testing, not by reading.
2. `normalizeTitle` strips `\p{Mn}` only after an ASCII base (`(?<=[a-z])\p{Mn}`). Latin accents always follow one once lowercased, so `Sênior` → `senior` is unchanged (#2209); matras, breve and dakuten are not touched.

## Tests

Four assertions, including the two that stop this becoming a *different* bug: different roles must stay distinct (the widening must not merge), and Latin accent-folding must not change. Verified against `main`:

```
❌ role titles still tokenize to nothing: Бэкенд-разработчик, सॉफ्टवेयर इंजीनियर
❌ mark handling wrong: कंपनी/कपनी expected different; Йогурт/Иогурт expected different
```

`node test-all.mjs --quick` → **3655 passed, 0 failed**.

## Out of scope, flagged rather than half-fixed

CJK still tokenizes poorly — the `w.length > 3` filter assumes word-separated text and `決済` is a two-character word. That needs segmentation, not a character class, so Japanese/Chinese titles are improved but not fixed here. Happy to take it separately if you want it.

2 commits (fix + test).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved role matching for titles written in Cyrillic, Devanagari, and other non-Latin scripts.
  * Preserved meaningful marks in non-Latin text while retaining Latin accent-insensitive matching.
  * Improved handling of Unicode letters, marks, numbers, and punctuation in role titles.

* **Tests**
  * Added regression coverage for multilingual titles, punctuation differences, and accent handling.


santifer:
Heads-up, @abankar1: this now conflicts, and it's my doing, not yours. I merged your #2705 an hour ago and then pushed a change of my own to `test-all.mjs` (a7cfa1bd), which is the file you also touch here.

Don't rebase yet. This one still needs a real review from me and its CI is red on the last run, so a rebase now would just be work you'd repeat. I'll come back with the review and tell you exactly what to rebase onto.

santifer:
This is conflicting with `main` now, @abankar1, and that is my doing rather than yours: I merged eight PRs today, one of them your own #2705, and `test-all.mjs` is one of the most collision-prone files in the repo. GitHub says nothing when it happens.

Its CI is also red for a reason that is not yours either. The only failing check is `test (windows-latest)`, and the failure is `agent-inbox-tests.mjs` section 7, `kept=29 of 30`, a concurrency flake in a file your branch does not touch. It is being tracked in #2777 and it is still firing after today's fix attempt.

**One thing before you rebase, because it will save you a round trip.** #2782 fixes non-Latin role titles tokenizing to nothing, and #2705, yours, landed today for the neighbouring problem on the company side. Since they are two halves of the same class, worth checking whether main's version of the company path already gives you a helper you can reuse rather than a second implementation of the same normalization.

Ping me here when it is up and I will take it the same day. If you would rather wait until #2777 stops reddening everything, that is also a reasonable call and I will not read it as the PR going quiet.

santifer:
First, a correction on my side: I answered #2781 telling you to go ahead and open the PR, and it had been open since before I wrote that. That's on me for reading the issue and not the queue.

The fix itself is better than what I'd sketched. `(?<=[a-z])\p{Mn}` scopes the fold by "marks on an ASCII Latin base" instead of by "all marks", which is the precise version of what I was hand-waving at, and the trailing `.normalize('NFC')` is the bit I'd have forgotten. I checked it out and ran it:

```
ruso        ["бэкенд","разработчик","платежи"]
hindi       ["सॉफ्टवेयर","इंजीनियर"]
match ruso: true
Devanagari: ["कंपनी"] vs ["कपनी"]     ← no longer one token
```

Both defects from the issue, closed. Merged 🚀

**What's still open, and it's a different bug than the one you fixed** — same run, same file:

```
japones     ["バックエンドエンジニア"]        ← one token
chino       ["后端工程师"]                   ← one token
match japones: false
match chino:   false
```

The character class is right now; what defeats CJK is that `.split(/\s+/)` assumes spaces separate words. Japanese and Chinese role titles don't have them, so the whole title becomes a single token — and `roleFuzzyMatch` requires 2+ shared tokens, so a title still can't match itself. Went from "zero tokens" to "one token", and the dedup outcome is identical: a duplicate row.

`Intl.Segmenter` handles it, verified on the same Node this repo runs:

```
new Intl.Segmenter('ja', {granularity:'word'})  →  ["バック","エンド","エンジニア"]
new Intl.Segmenter('zh', {granularity:'word'})  →  ["后","端","工程","师"]
```

Note the Chinese output: two of those four are single characters, so `w.length > 3` would drop them and the length filter needs revisiting alongside — it's a Latin-shaped heuristic. That's the part I'd want fixtures on before it ships, in both directions: distinct CJK roles must stay distinct, or dedup starts merging real separate applications.

Separate PR if you want it — this one stands on its own, and I'd rather it land now than wait for the harder half.

PR #2783, opened 2026-08-12
fix(patterns): rate denominators counted rows that could never contribute

felipe-NR:
## What does this PR do?

`blockerAnalysis` and `discardReasonStats` both divide by `enriched.length` — every tracker row — instead of the population the rate is actually drawn from. Entries carrying no gaps can never produce a blocker, and pending `Evaluated` rows can never carry a discard reason, yet both pad the denominator. Since recommendation gates are thresholds on those rates, the dilution silently disables them.

## Related issue

None — bug fix, which the PR template exempts from issue-first.

## The bug

`analyze-patterns.mjs`, blocker site:

```js
const totalWithGaps = enriched.filter(e => e.report?.gaps?.length > 0);   // computed…
...
percentage: Math.round((frequency / enriched.length) * 100),              // …never used
```

`totalWithGaps` is computed one line above the map and never referenced. The intended denominator was already written down; the code just did not use it.

The discard-reason site has the same shape: its harvest loop accepts only `self_filtered + negative`, but its percentages and recommendation threshold used every tracker row.

## Why it matters — the gate inverts its own incentive

The discard-reason gate is `max(3, ceil(N * 0.15))` where `N` was every tracker row, but only the eligible subset — call it `B` — can contribute a reason.

Because `Evaluated` rows accumulate faster than discarded ones for most users, `N` drifts far above `B`, and the threshold drifts with it. At `N = 263` and `B = 60`, the threshold resolves to **40 occurrences — 67% of the rows that could ever supply one**. Coarse blanket labels can clear that gate while precise per-reason labels do not.

`blockerAnalysis` has the same defect against a base padded by every gapless entry, under-firing `geoBlocker.percentage >= 20` and `stackBlocker.percentage >= 15` by the ratio of gap-bearing entries to total entries.

## The fix

Two named population helpers define the eligible bases:

```js
function discardableBase(enriched)  // self_filtered + negative
function gapBearingBase(enriched)   // entries with a non-empty gaps array
```

The production analysis and self-test share `buildPatternSignals()`, which applies those bases consistently to blocker percentages, discard-reason percentages, and the discard recommendation threshold. Both bases guard against zero denominators (`0`, never `NaN`).

Counts are deduplicated per entry before rates are calculated:

- a repeated `SKIP`/`DISCARD` reason in one Notes cell contributes once;
- repeated mentions of the same technology in one report contribute once;
- several gaps resolving to the same blocker type contribute one affected entry.

Each base is exported on the result object and documented in the patterns JSON contract:

- `blockerBase` is the denominator for `blockerAnalysis[].percentage`;
- `discardReasonBase` is the denominator for `discardReasonStats[].percentage` and its recommendation threshold.

The summary and recommendation output make the population visible:

```
TOP DISCARD / SKIP REASONS (of 60 self-filtered/negative entries)
TOP BLOCKERS (of 84 entries carrying gaps)

Review location and work-authorization filters in portals.yml --
17/84 entries carrying gaps (20%) hit a geo-restriction blocker
```

The recommendation describes the observed restriction neutrally and asks the reader to review the filters; it does not label every positive, pending, or self-filtered entry as wasted effort.

## Review follow-ups by commit

- `f769e8f` — correct both denominators, export their bases, and label the summary populations.
- `bef5d1f` — make the discard threshold fixture distinguish the old formula and deduplicate discard reasons/technology mentions per entry.
- `fd693f2` — deduplicate blocker types per entry so no share can exceed its own base.
- `e197f81` — route regression fixtures through the production `buildPatternSignals()` aggregator.
- `b33a41d` — expose the conditional blocker population in the geo recommendation and add gapless rows so restoring `/ enriched.length` fails the blocker regression.
- `2e3ed84` — cover both discardable outcomes, assert no empty-population recommendation, generalize geo/work-authorization wording, and document the JSON denominator fields.
- `7e0d3ea` — keep blocker guidance evidence-neutral and make the report instruction explicit that blocker percentages use `blockerBase`, not all applications.
- `be12abf` — synchronize the branch with the latest `main` before the final validation.

## Type of change

- [x] Bug fix
- [ ] New feature
- [x] Documentation / contract clarification

## Tests

`analyze-patterns.mjs --self-test` exercises the same aggregation helper used by production:

- 30 tracker entries: 20 discard-eligible (including both `self_filtered` and `negative`) and 10 pending; three deduplicated `geo-block` tags produce base `20`, frequency `3`, `15%`, and a recommendation. The old full-tracker formula would require five occurrences.
- 5 tracker entries: only two carry gaps; repeated geo gaps produce frequency `2`, base `2`, and `100%`. The old `/ enriched.length` formula produces `40%` and fails.
- repeated Java mentions in one report count once;
- empty populations expose zero bases, empty stats, and no recommendation.

Mutation check:

```
# Restore the old blocker denominator
analyze-patterns self-test failed: geo blocker aggregation returned
{"blocker":"geo-restriction","frequency":2,"percentage":40}
against base 2, expected 2/2 (100%)
```

Validation after synchronizing with current `main`:

- `node analyze-patterns.mjs --self-test` — passed
- `node --check analyze-patterns.mjs` — passed
- `node test-all.mjs` — **3664 passed, 0 failed, 1 expected warning**
- GitHub checks — **3 passed, 0 failed**

## Checklist

- [x] Bug fix — exempt from issue-first per the template

---
🤖 Implemented with Claude Code and Codex

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved analysis percentages and recommendations by using the appropriate eligible entries as their basis.
  * Prevented invalid results when no eligible entries are available.
  * Removed duplicate blocker, discard-reason, and technology contributions within the same entry.
  * Improved consistency when entries contribute to multiple analysis categories.
  * Clarified summary labels to show which populations percentages are based on.

* **Documentation**
  * Added clearer denominator information, discard-reason statistics, and recommendation threshold details to analysis results.


felipe-NR:
@coderabbitai review

felipe-NR:
### Final review resolution

All denominator-review findings are resolved on the current head `7e0d3ea`:

- `b33a41d`: expose the blocker population and make the old denominator fail the regression fixture.
- `2e3ed84`: cover `negative`, empty recommendations, neutral geo/work-authorization wording, and JSON contract fields.
- `7e0d3ea`: remove the unsupported wasted-effort claim and state explicitly that blocker percentages use `blockerBase`, not all applications.

Final validation after merging current `main`:

- `analyze-patterns.mjs --self-test`: passed
- mutation to `/ enriched.length`: correctly rejected at 40% vs expected 100%
- `node test-all.mjs`: **3664 passed, 0 failed, 1 expected warning**
- GitHub checks: **3 passed, 0 failed**

The PR description contains the full bug rationale, commit-by-commit follow-ups, denominator contract, and test evidence.

santifer:
Welcome to career-ops, @felipe-NR. This is an unusually good first PR, and the reason is the sentence in your report rather than the diff: *"the intended denominator was already written down; the code just did not use it."*

Verified before reading your fix. `totalWithGaps` appears **exactly once** in `analyze-patterns.mjs` on `main` — the line that computes it. Never referenced again. Someone knew what the denominator should be, wrote it, and the next line divided by `enriched.length` anyway.

Then I checked your own contract by breaking it. Your self-test comment says restoring `/ enriched.length` must go red, so I restored it:

```
analyze-patterns self-test failed: geo blocker aggregation returned
  {"blocker":"geo-restriction","frequency":2,"percentage":40} against base 2,
  expected 2/2 (100%)
```

A 2-of-2 blocker reported as **40%**, in a fixture of five rows. Restored, green again.

**The part that makes this a bug rather than a rounding complaint** is the one you led with: `max(3, ceil(N * 0.15))` where `N` grows with rows that can never contribute. The gate gets *stricter* the longer someone uses the tool, so the person with the most history — exactly the person the analysis is for — is the one whose recommendations quietly stop firing. Nothing errors, nothing logs, the section just goes silent.

Touching `modes/patterns.md` in the same PR was right, and I checked it says the same thing the code now does: `blockerBase` and `discardReasonBase` are named, and *"the share of gap-bearing entries, not the share of all applications"* replaces the old line. A rate whose denominator is documented wrong is a rate that will drift back.

No linked issue, and that is fine here: the template exempts bug fixes, and a report this specific is worth more than an issue would have been. Suite on a preview against today's main: 3820 passed, 0 failed. All three test platforms green on your current commit.

Issue #1506, opened 2026-07-03
RFC: Shared Behavioral-Signal Layer schema — company-attributable, candidate-anonymous (follow-up to Discussion #904)


Schlaflied:
## Context

In [Discussion #904](https://github.com/santifer/career-ops/discussions/904), @santifer proposed an opt-in shared layer on top of career-ops so the community stops re-evaluating the same listings in isolation. Two commenters (@drkalexander1, @ManishVangara) raised the central tension: career-ops evaluations are heavily profile-dependent (CV match, level fit, career alignment) — so what exactly is shareable across users with different profiles?

A comment in that thread proposed an answer: **share facts about companies, not scores about candidates.** @santifer adopted this as the working filter for anything that would ever cross the wire:

> "A fit score is a function of you — sharing it leaks your profile. Interviewer latency, no-show reschedules, contradictory role descriptions are facts about them — observable, company-attributable, candidate-anonymous."

This RFC operationalizes that filter into a concrete schema, grounded in two pieces of local prototype work already merged/in-review in this repo, rather than designing from scratch.

## Existing local prototypes (the schema isn't invented, it's extracted)

**1. `modes/interview-redflag.md` (#1233)** — analyzes the *interviewer's* side of session transcripts against a 4-signal taxonomy:

| Signal | What it captures |
|---|---|
| Scope ambiguity | No measurable evaluation criteria, contradictory role descriptions across interviewers |
| Defensive closure | Interviewer shortens/redirects after a strong candidate answer |
| Evaluator competency gap | Interviewer can't engage with a domain-correct answer |
| Process signals | Contradictory criteria across interviewers, no structured follow-up |

Aggregated per company into a red-flag score (0–8) and a warning level (`None` / `⚠️ Caution` / `🚩 Warning`). Already company-attributable, already candidate-anonymous (it never records *who* said what, only *that* a pattern occurred).

**2. `process-quality.mjs` (#1466/#1467, merged)** — parses a `[process-friction: <reason>]` tag the candidate adds inline in their own tracker notes, aggregates per company, explicitly scoped as "never tied to a named individual recruiter."

Both pieces already satisfy the objective/company-attributable/candidate-anonymous filter *locally*. This RFC's job is to define how this data crosses the wire into a shared pool, not to invent a new signal taxonomy.

## Proposed schema

```json
{
  "schemaVersion": 1,
  "companyKey": "acme-corp",           // normalized company slug, not a legal entity name
[...]

PR #1467, opened 2026-07-03
feat(patterns): recruiting-process friction signal — closes #1466

Schlaflied:
## Summary
- Adds `process-quality.mjs` — a company/process-level friction aggregator, distinct from #1232 (danger-signal red flags from interviewer behavior) and #960 (role-fit targeting correction). Tracks a third axis: is the *recruiting process itself* (scheduling, comms clarity, tooling) well-run, independent of whether the company or role is a good fit.
- Candidates tag friction inline in `data/active-interviews.md`'s Notes column: `[process-friction]` or `[process-friction: short reason]`. The script parses the table, aggregates friction rate per company (case-insensitive dedup), and outputs JSON or a `--summary` table.
- **Company/process-level only, never a named-individual-recruiter signal** — consistent with the project's Ethical Use principles (respect recruiters' time, no adversarial candidate-side tooling). See the "Not in scope" section of #1466.
- Sits under the same umbrella as #1232/#960/#1242 (shared session-transcript/interview-log signal source).

## Implementation notes
- `parseActiveInterviews()`, `extractFriction()`, `aggregateProcessQuality()` are exported for testing, mirroring `detect-reposts.mjs`'s module shape (same author's prior PR, same conventions: CLI args, `--self-test`, JSON/`--summary` output modes).
- Malformed rows (wrong column count, no table found) are silently dropped rather than crashing the parser.
- `aggregateProcessQuality()` supports `--min-threshold N` to filter out companies with too few data points to be meaningful (default 1).
- Sort order: friction count desc → friction rate desc → company name asc (deterministic).

## Test plan
- [x] `node process-quality.mjs --self-test` — 19 checks, all passing
- [x] `node process-quality.test.mjs` — standalone 68-case suite (parsing, tag extraction, grouping/dedup, threshold filtering, sort order, CLI), mirrors `detect-reposts.test.mjs`'s coverage shape
- [x] Wired `process-quality.mjs --self-test` into `test-all.mjs`'s CI-gated script list
- [x] `node test-all.mjs --quick` — full suite passes except one pre-existing, unrelated Windows/WSL flake (#1409, batch rate-limit pause test) present on `main` before this change
- [x] Documented in `AGENTS.md`'s Main Files table

Closes #1466.

## Summary by CodeRabbit

* **New Features**
  * Added a process-quality report that computes per-company “process friction” signals from active interview notes, available as JSON or a readable summary table.
  * Added a self-check mode to validate parsing and output consistency.
* **Tests**
[...]

PR #1233, opened 2026-06-26
feat(interview): company red-flag detector from transcript signal

Schlaflied:
Closes #1232.

## What this adds

`modes/interview-redflag.md` — a new skill mode that reads the **interviewer's side** of session transcripts and surfaces structural red flags about a company before the candidate accepts an offer or re-applies.

## Relationship to existing work

**Sibling of #960 / `realign-targeting`** — same input corpus (#686/#509 transcripts), orthogonal analysis:

| Mode | Reads | Answers |
|------|-------|---------|
| `realign-targeting` (#960) | Candidate's answers | Where should I aim next? |
| `interview-redflag` (this PR) | Interviewer's questions | Is this company safe to enter? |

**Depends on** #686 (`interview/debrief`) and #509 (`mock`) landing first to produce `interview-prep/sessions/` transcripts — same dependency as #960. Exits gracefully with an onboarding message if no sessions exist yet.

## Signal taxonomy

Four interviewer-side signal types, each scored per session and aggregated per company:

- **Scope ambiguity** — questions with no measurable criteria; role described in circular/contradictory terms
- **Defensive closure** — session shortens or redirects after a technically complete candidate answer
- **Evaluator competency gap** — interviewer cannot engage with a domain-correct response
- **Process signals** — contradictory criteria across interviewers; absent follow-up across all rounds

Single occurrence = noise (ignored). Two or more occurrences = pattern (flagged).

## Warning levels

| Score | Level |
|-------|-------|
| 0–1 | No red flags |
| 2–3 | `⚠️ Enter with eyes open` |
| 4+ | `🚩 Reconsider` |

## Design constraints respected

- No new dependencies
- No auto-action — advisory only, appended to existing `interview-prep/{company}-{role}.md`
- No changes to `portals.yml`, `modes/_profile.md`, or tracker
- Privacy: reads only local, gitignored session files

## Files changed

- `modes/interview-redflag.md` — new mode (144 lines)
- `AGENTS.md` — skill table: one new row for `interview-redflag`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added a new “Company Red-Flag Detector” mode that analyzes saved interview session transcripts to identify structural interviewer red flags and produce a company-level advisory.
  * Updated skill mode guidance so “company safety” checks using red-flag analysis route to the new detector.
* **Documentation**
[...]

PR #2788, opened 2026-08-13
Closes #2787: emit no-response-friction signal per RFC #1506 schema v1

### LINKED ISSUE #2787: feat(shared-layer): no-response-friction signalType — emit company-history.mjs's responsiveness fact per RFC #1506 schema v1
## What

RFC #1506 ratified `no-response-friction` as a v1 `signalType` on 2026-07-16, but nobody has implemented it. Separately, #1712 (merged 2026-07-28) shipped `company-history.mjs`, which already computes a per-company `responsiveness` axis — including a `silent-on-you` label — from the tracker, `follow-ups.md`, and `scan-history.tsv`. The raw signal this RFC asked for already exists locally; what's missing is wiring it into the ratified emission schema.

## Schema (from RFC #1506, verbatim)

```json
{
  "companyKey": "acme-corp",
  "region": "north-america/canada",
  "signalType": "no-response-friction",
  "severity": "single | pattern | null",
  "sourceHash": "sha256:...",
  "observedAt": "2026-06",
  "emittedBy": "career-ops v1.16.0"
}
```

## Proposal

1. When `company-history.mjs`'s `responsiveness.label === 'silent-on-you'` for a company, emit a record in the schema above:
   - `companyKey` — the same normalized key `company-history.mjs` already produces (parent:division compound key where available).
   - `region` — read from `config/profile.yml`.
   - `signalType` — `'no-response-friction'`.
   - `severity` — `'single'` for one silent application to that company, `'pattern'` for 2+ silent applications to the same company.
   - `observedAt` — month-only, derived from the relevant fact's date.
   - `sourceHash` — opaque hash for dedup, no raw data recoverable from it.
   - `emittedBy` — current package version.
2. **Threshold discrepancy to resolve, not silently pick:** the RFC's original no-response proposal used a 14-day baseline ("~2 weeks before a first prescreen typically happens"); `company-history.mjs`'s shipped `silent-on-you` label uses a 28-day default silence window (`DEFAULT_SILENCE_WINDOW_DAYS`, overridable via `--silence-window`). Recommend reusing `company-history.mjs`'s existing, already-tested 28-day window rather than introducing a second, different silence threshold — but flagging this explicitly since it wasn't decided in the RFC thread.
3. **Opt-in only, never auto-published.** This produces a local file/record in the emission format; publishing it anywhere (a `shared-layer/` PR, a future shared repo) stays a separate, explicit user action, matching Discussion #904's "company-attributable, candidate-anonymous" principle and the RFC's own privacy discipline (no verbatim quotes, no named individuals, month-only dates).
4. Likely shape: a new `--emit-signal` flag on `company-history.mjs`, or a small new script consuming its JSON output — reusing the data source rather than recomputing responsiveness independently.

## Test plan

- Fixture: a `silent-on-you` company-history card → verify the emitted record matches the schema exactly (field names, `observedAt` month-only, no PII).
- Fixture: same company silent across 2+ separate applications → verify `severity: 'pattern'`.
- Fixture: single silent application → verify `severity: 'single'`.
- Verify nothing is written/published without an explicit opt-in flag.

Ref: RFC #1506, #1712 (`company-history.mjs`, the data source this consumes).

### LINKED ISSUE #3010: feat(RFC #1506): add `postingChannel` field — distinguish agency-mediated from direct-employer signals
## Problem

RFC #1506's local emission schema (ratified) keys every signal to `companyKey` + `region`, with no field distinguishing *who actually posted the role* — the direct employer, or a third-party staffing/recruiting agency filling a client req. That distinction matters for exactly the confound raised in [discussion #904](https://github.com/santifer/career-ops/discussions/904#discussioncomment-18058073): a structural, single-criterion signal (e.g. "not found on the employer's own careers page") disproportionately flags staffing/outsourcing firms, because they legitimately advertise client roles that never appear on their own boards. @strelov1 (freehire.me) measured this directly in production — restricting by industry made concentration *worse* (top-10 share 29.5% → 39.7%), and only a second, independent criterion fixed it (98.5% drop).

This is unrelated to the trust-model question (Sybil/manipulation risk, raised by @drkalexander1 in #1506's round 5 and out of scope for that RFC per @santifer's ruling) — it's not about whether a signal can be faked, it's a missing **factual attribute of the posting/company**, the same category as `companyKey` or `region`.

career-ops already detects this: **#2041** (Block G Signal 9/10, agency-mediated posting detection via wording cues + candidate-reported source) and the tracker's tagged **`via={Agency}` field (#1596)** both already classify direct-vs-agency at the point of use. Nothing currently carries that classification into #1506's shared-signal schema, so any future aggregation (local, or eventually the shared-layer service in #904) has no way to filter or weight by posting channel — it would rediscover @strelov1's confound the same way freehire.me first did.

## Proposed shape

Add one field to #1506's schema:

```json
"postingChannel": "direct-employer | staffing-agency | unknown"
```

- **Source:** wire in the detection already shipped in #2041, plus the tracker's `via={Agency}` tag when the signal is derived from tracker data (e.g. `no-response-friction`, #2788). No new detection logic — this is a wiring task, not a new capability.
- **Default:** `unknown` when neither signal is present — never force a guess.
- **Scope discipline:** factual tag only, same as `companyKey`/`region`. Does not touch the trust-model question and should not be conflated with it in review.

## Relation to other work

- RFC #1506 (schema owner) — this is an additive field on the ratified schema, not a redesign.
- #2041 / #1596 — reused detection, no new logic.
- Discussion #904 — where @strelov1's measurement that prompted this lives.

Schlaflied:
## What

Closes #2787, #3010.

RFC #1506 ratified a `no-response-friction` `signalType` for the local emission schema v1 on 2026-07-16, but nobody wired it up. #1712 (merged 2026-07-28) already ships `company-history.mjs`, which computes a per-company `responsiveness` axis including a `silent-on-you` label from the tracker, `data/follow-ups.md`, and `data/scan-history.tsv`. This PR wires that existing fact into the ratified schema, rather than deriving a second silence detector from scratch.

## What was built

- **`buildNoResponseFrictionSignals(result, opts)`** in `company-history.mjs` — takes the already-built `result` from `buildCompanyCards()` and emits one schema-v1 record per `silent-on-you` company card:

```json
{
  "companyKey": "acme-corp",
  "region": "north-america/canada",
  "signalType": "no-response-friction",
  "severity": "single | pattern | null",
  "sourceHash": "sha256:...",
  "observedAt": "2026-06",
  "emittedBy": "career-ops v1.26.0"
}
```

- **`--emit-signal`** CLI flag (opt-in only). When passed, in addition to the script's normal output (JSON / `--summary` / `--company`), it prints the schema-v1 records as JSON Lines (one record per line) to stdout. **Running the script without the flag is byte-for-byte identical to pre-#2787 output** — verified with `git stash` diff against the old behavior.
- No file writes, no publishing anywhere. This PR only produces local schema-v1-shaped stdout output, exactly as scoped in #2787 and by RFC #1506's "career-ops' job is to define the format and emit into that layer — never to host the pool."

## Design decisions

- **`companyKey`** reuses `company-history.mjs`'s existing normalized key (`normalizeCompany()` / `tracker-utils.mjs`) unchanged. The RFC thread floated an optional `parent:division` compound key, but I confirmed that's never actually been implemented anywhere in the codebase (only discussed) — so there was nothing to reuse there beyond the flat key, and inventing a new key format was explicitly out of scope.
- **`region`** reads `config/profile.yml` → `location.country`, mapped through a small country→region-slug table (`north-america/canada`, `europe/germany`, etc. — covers the markets this repo already has explicit market-mode support for, plus a few common others). Degrades to `unknown` when the profile is missing/unparsable, or `unmapped/{slug}` for a country not in the table — never crashes.
- **`severity`**: `'single'` for a card with exactly one silent fact, `'pattern'` for 2+ (i.e., the candidate applied more than once and went unanswered more than once at the same company).
- **`observedAt`**: month-only (`YYYY-MM`), derived from `appliedDate + silenceWindowDays` — the date the application actually *crossed* into silence — not from "today." This keeps the value (and therefore `sourceHash`) stable across repeated runs, which dedup depends on.
- **`sourceHash`**: deterministic `sha256` over `no-response-friction|companyKey|observedAt|severity` (not random/salted). Deterministic so re-running against the same underlying fact reproduces the same hash — required for downstream dedup. It adds no information beyond what the record's own plaintext fields already expose; it's opaque only in the sense that it can't be reversed into anything *more* sensitive (raw applicant identity, notes text) than what's already visible in `companyKey`/`region`/`observedAt`.
- **`emittedBy`**: `career-ops v${package.json version}`, read live from `package.json`; degrades to `'career-ops'` if unreadable.
- **Threshold**: reuses `company-history.mjs`'s shipped 28-day `DEFAULT_SILENCE_WINDOW_DAYS` (and its `--silence-window` override) rather than introducing a second, undocumented 14-day cutoff from the RFC thread's original proposal. Commented at the point of use in the code so a future reader isn't left looking for a 14-day constant that doesn't exist.
- **Privacy**: an emitted record carries *exactly* the 7 schema fields — no candidate name, no free-text notes, no verbatim quotes, same discipline as `interview-redflag` / `process-friction`. Asserted directly in the self-test (`Object.keys(record)` equality check).

## Test coverage

9 new assertions added to `company-history.mjs`'s existing `runSelfTest()`, reusing its existing fixture-building helpers (`row()`, `buildCompanyCards()`):

- A `silent-on-you` card with exactly one silent fact → `severity: 'single'`.
- A company silent across 2 separate applications → `severity: 'pattern'`.
- `responded-before`, `mixed`, and `no-history` cards → zero records emitted (only `silent-on-you` triggers this signal).
- `observedAt` is strictly `YYYY-MM` (regex-asserted).
- Emitted record's key set is exactly the 7 schema fields — no PII leakage.
- `sourceHash` is deterministic across repeated runs against the same fact.
- `resolveRegion()` / `resolveEmittedBy()` degrade gracefully (no crash) against missing files.

Full suite: `node test-all.mjs --quick` → **3663 passed, 0 failed** (1 pre-existing, unrelated warning: `cv-sync-check.mjs exited with error (expected without user data)`).

Regression check: `node company-history.mjs` output (no flag) diffed byte-for-byte identical before/after this change.

## Additive: `postingChannel` field (closes #3010)

Landed in a follow-up commit on this same branch, rather than a separate PR, since it's a small additive field on the exact record shape this PR already builds.

Raised in [discussion #904](https://github.com/santifer/career-ops/discussions/904#discussioncomment-18058073): a single-criterion structural signal like this one systematically misfires on staffing/recruiting agencies, which legitimately post client roles that never appear on their own careers page (measured in production by @strelov1/freehire.me — restricting by industry made concentration *worse*). Rather than inventing new detection, this wires in a fact career-ops already tracks: the tracker's own tagged `via={Agency}` field (#1596).

- **`resolvePostingChannel(via)`** — new export in `company-history.mjs`. A tagged non-em-dash value → `"staffing-agency"`; the tracker's own em-dash "confirmed direct" convention → `"direct-employer"`; undefined/null/blank → `"unknown"` (never guessed — same "degrade to a real absence" discipline `resolveRegion()` already uses above).
- `computeResponsiveness()`'s silent-fact object now carries `via` through from the raw tracker row so the signal builder can read it without a second join.
- `computeSourceHash()` is untouched — `postingChannel` is not a hash input, for the same reason `severity` isn't (see the design-decision above): it can be corrected/backfilled later without changing the underlying fact's identity.
- Flagged in the schema comment block as additive/pending-review, not part of the RFC-ratified fields above — this field hasn't itself gone through the RFC's own ratification round.
- 5 new self-test assertions (3 end-to-end via `buildNoResponseFrictionSignals`, 2 direct unit checks on `resolvePostingChannel`), plus the existing exact-field-set assertion updated to include it.

## Refs

- RFC #1506 — schema v1, ratified 2026-07-16 (this PR implements the `no-response-friction` half of that ratification)
- #1712 — `company-history.mjs`, the data source this PR consumes (`silent-on-you` responsiveness fact)
- #3010 — `postingChannel` field, closed by the follow-up commit above
- Discussion #904 — where the agency-misfire data point that motivated #3010 was raised

## Summary by CodeRabbit

* **New Features**
  * Added optional `--emit-signal` support for generating schema-v1 no-response friction signals in JSON Lines format.
  * Signals include privacy-safe regions, source metadata, deterministic hashes, observation dates, and `single` or `pattern` severity.
  * Supports stale-record filtering and graceful handling of missing or malformed metadata.

* **Bug Fixes**
  * Excludes responded, mixed, and no-history records from signal generation.
  * Existing output remains unchanged unless signal emission is enabled.

* **Tests**
  * Added coverage for signal generation, filtering, deterministic output, and invalid command-line formats.

### DIFF
--- a/company-history.mjs
+++ b/company-history.mjs
@@ -781,7 +781,14 @@ export function buildNoResponseFrictionSignals(result, opts = {}) {
     // would report severity 'pattern' even though only one fact actually
     // contributed to the 'silent-on-you' label.
     const activeFacts = includeStale ? card.responsiveness.facts : card.responsiveness.facts.filter(f => !f.stale);
-    const silentFacts = activeFacts.filter(f => 'silentDays' in f);
+    // A fact with an unusable appliedDate must be excluded before anchor
+    // selection and severity counting, not just skipped later when computing
+    // observedAt from the already-chosen anchor: left in, a malformed date
+    // could win the lexicographic reduce below (producing no usable
+    // observedAt at all) or, even when a valid fact wins the anchor, still
+    // inflate severity to 'pattern' despite not actually contributing a
+    // usable silent-application fact (CodeRabbit, PR #2788).
+    const silentFacts = activeFacts.filter(f => 'silentDays' in f && silentFactObservedAt(f) !== null);
     if (silentFacts.length === 0) continue; // defensive; label implies >=1
 
     // Most-recent APPLICATION anchors observedAt — compare appliedDate, not
@@ -1243,6 +1250,27 @@ async function runSelfTest() {
     );
     check(singleSignalsAgain[0]?.sourceHash === singleSignals[0]?.sourceHash, 'sourceHash is deterministic across repeated runs against the same fact (dedup-safe)');
 
+    // sourceHash must actually be sensitive to companyKey and observedAt —
+    // determinism alone doesn't prove the hash isn't a constant or doesn't
+    // silently drop one of its inputs (CodeRabbit, PR #2788).
+    const { records: differentCompanySignals } = buildNoResponseFrictionSignals(
+      buildCompanyCards(
+        { trackerRows: [row(203, 'DifferentCompanyCo', 'Applied', '2026-05-01')], followupRows: [], repostClusters: [], sourcesLoaded: { tracker: true, followups: false, scanHistory: false, statusLog: false } },
+        { now: NOW, silenceWindowDays: 28 },
+      ),
+      fixedOpts,
+    );
+    check(differentCompanySignals[0]?.sourceHash !== singleSignals[0]?.sourceHash, 'sourceHash differs when companyKey differs (same observedAt month) — companyKey is a real hash input, not dropped');
+
+    const { records: differentMonthSignals } = buildNoResponseFrictionSignals(
+      buildCompanyCards(
+        { trackerRows: [row(204, 'SingleSilentCo', 'Applied', '2026-06-01')], followupRows: [], repostClusters: [], sourcesLoaded: { tracker: true, followups: false, scanHistory: false, statusLog: false } },
+        { now: NOW, silenceWindowDays: 28 },
+      ),
+      fixedOpts,
+    );
+    check(differentMonthSignals[0]?.sourceHash !== singleSignals[0]?.sourceHash, 'sourceHash differs when observedAt differs (same companyKey) — observedAt is a real hash input, not dropped');
+
     // severity excludes stale facts from the count unless --include-stale is
     // passed — a card with one active + one stale silent fact must report
     // 'single', matching what the 'silent-on-you' label itself was computed
@@ -1262,6 +1290,19 @@ async function runSelfTest() {
     check(activePlusStaleSignals.length === 1, 'active+stale fixture: still exactly one record for the card');
     check(activePlusStaleSignals[0]?.severity === 'single', 'active+stale fixture: severity is single — the stale fact does not inflate it to pattern, matching the label');
 
+    // The includeStale:true emission path must actually be exercised
+    // end-to-end (not just the default-exclusion path above), so a
+    // regression in the opt-in branch would be caught (CodeRabbit, PR #2788).
+    const activePlusStaleIncludeStaleResult = buildCompanyCards(
+      { trackerRows: staleRows, followupRows: [], repostClusters: [], sourcesLoaded: { tracker: true, followups: false, scanHistory: false, statusLog: false } },
+      { now: NOW, silenceWindowDays: 28, includeStale: true },
+    );
+    const { records: activePlusStaleIncludeStaleSignals } = buildNoResponseFrictionSignals(
+      activePlusStaleIncludeStaleResult,
+      { ...fixedOpts, includeStale: true },
+    );
+    check(activePlusStaleIncludeStaleSignals[0]?.severity === 'pattern', 'active+stale fixture with includeStale:true: severity is pattern — the stale fact now counts');
+
     // silentFactObservedAt returning null (unusable appliedDate) must skip the
     // card entirely — no record, no crash. Constructed directly rather than
     // through buildCompanyCards, since computeResponsiveness() never produces
@@ -1277,6 +1318,29 @@ async function runSelfTest() {
     };
     const { records: unusableDateSignals } = buildNoResponseFrictionSignals(unusableDateResult, fixedOpts);
     check(unusableDateSignals.length === 0, 'a card whose anchor fact has an unusable appliedDate is skipped entirely — no record emitted');
+
+    // Mixed valid/invalid facts: the unusable-date fact must be filtered out
+    // BEFORE anchor selection and severity counting, not merely tolerated by
+    // whichever fact happens to win the reduce — otherwise it could win the
+    // anchor outright, or (even when the valid fact wins) still inflate
+    // severity to 'pattern' despite contributing no usable observation
+    // (CodeRabbit, PR #2788).
+    const mixedValidityResult = {
+      companies: [{
+        key: 'mixed-validity-co',
+        responsiveness: {
+          label: 'silent-on-you',
+          facts: [
+            { num: 1, appliedDate: '2026-06-01', silentDays: 40, stale: false },
+            { num: 2, appliedDate: 'not-a-real-date', silentDays: 40, stale: false },
+          ],
+        },
+      }],
+    };
+    const { records: mixedValiditySignals } = buildNoResponseFrictionSignals(mixedValidityResult, fixedOpts);
+    check(mixedValiditySignals.length === 1, 'mixed valid/invalid facts: still exactly one record — the invalid fact does not block emission');
+    check(mixedValiditySignals[0]?.observedAt === '2026-06', 'mixed valid/invalid facts: observedAt anchors on the valid fact');
+    check(mixedValiditySignals[0]?.severity === 'single', 'mixed valid/invalid facts: severity is single — the invalid fact does not count toward pattern');
 
     // silentFactObservedAt must trim appliedDate before slicing, not just
     // before validating: parseDate() tolerates surrounding whitespace, but a
--- a/company-history.mjs
+++ b/company-history.mjs
@@ -169,6 +169,21 @@ function parseArgs(argv) {
     process.exit(1);
   }
 
+  // --emit-signal is a bare boolean flag, not a value flag: the unknown-flag
+  // check above strips a `--flag=value` suffix before matching against
+  // KNOWN_FLAGS, so `--emit-signal=true` passes that check silently. But the
+  // boolean read below is an exact-token `args.includes('--emit-signal')`,
+  // which is false for `--emit-signal=true` — so the flag would be accepted
+  // as "known" yet never actually turn signal emission on. Reject any `=`
+  // form explicitly so a typo fails loudly instead of silently emitting
+  // nothing.
+  const emitSignalValue = args.find(a => a.startsWith('--emit-signal='));
+  if (emitSignalValue) {
+    console.error('Error: --emit-signal does not accept a value.');
+    console.error(USAGE);
+    process.exit(1);
+  }
+
   const summaryMode = args.includes('--summary');
   const company = valueOf('--company');
   const emitSignal = args.includes('--emit-signal');
@@ -671,7 +686,13 @@ export function computeSourceHash({ companyKey, observedAt, severity }) {
 // changes for a given fact, so anchoring on it keeps the hash stable.
 function silentFactObservedAt(fact) {
   const appliedDate = parseDate(fact.appliedDate);
-  const source = appliedDate ? fact.appliedDate : null;
+  // parseDate() tolerates surrounding whitespace when validating, but the
+  // schema's observedAt (and sourceHash, which is computed over it) must be
+  // a clean YYYY-MM slice — so trim before slicing, not just before
+  // validating, or a whitespace-padded appliedDate (e.g. ' 2026-06-01 ')
+  // would slice out the leading space plus wrong characters instead of the
+  // real year-month.
+  const source = appliedDate ? fact.appliedDate.trim() : null;
   return typeof source === 'string' && source.length >= 7 ? source.slice(0, 7) : null;
 }
 
@@ -1158,6 +1179,23 @@ async function runSelfTest() {
     };
     const unusableDateSignals = buildNoResponseFrictionSignals(unusableDateResult, fixedOpts);
     check(unusableDateSignals.length === 0, 'a card whose anchor fact has an unusable appliedDate is skipped entirely — no record emitted');
+
+    // silentFactObservedAt must trim appliedDate before slicing, not just
+    // before validating: parseDate() tolerates surrounding whitespace, but a
+    // whitespace-padded appliedDate sliced without trimming would produce a
+    // malformed observedAt (leading space, wrong characters) instead of the
+    // clean YYYY-MM the schema requires.
+    const paddedDateResult = {
+      companies: [{
+        key: 'padded-date-co',
+        responsiveness: {
+          label: 'silent-on-you',
+          facts: [{ num: 1, appliedDate: ' 2026-06-01 ', silentDays: 40, stale: false }],
+        },
+      }],
+    };
+    const paddedDateSignals = buildNoResponseFrictionSignals(paddedDateResult, fixedOpts);
+    check(paddedDateSignals[0]?.observedAt === '2026-06', 'a whitespace-padded appliedDate still produces a clean YYYY-MM observedAt');
 
     // anchor selection compares appliedDate, not tracker row num: a lower-num
     // row with a MORE RECENT appliedDate (a backfilled/out-of-order entry)
--- a/company-history.mjs
+++ b/company-history.mjs
@@ -41,10 +41,11 @@
  *      node company-history.mjs --self-test
  */
 
-import { readFileSync, existsSync } from 'fs';
+import { readFileSync, existsSync, writeFileSync, mkdtempSync, rmSync } from 'fs';
 import { join, dirname } from 'path';
 import { fileURLToPath, pathToFileURL } from 'url';
 import { createHash } from 'crypto';
+import { tmpdir } from 'os';
 import * as yaml from 'js-yaml';
 
 import { parseScanHistory, detectReposts } from './detect-reposts.mjs';
@@ -94,7 +95,13 @@ const USAGE = `Usage:
   node company-history.mjs --silence-window 21    # override the default silence window (days)
   node company-history.mjs --include-stale        # include facts older than 365d in label computation
   node company-history.mjs --self-test            # run the in-memory test suite
-  node company-history.mjs --emit-signal          # also print RFC #1506 schema-v1 no-response-friction records (JSON Lines, opt-in only)
+  node company-history.mjs --summary --emit-signal  # human-readable cards, then RFC #1506 schema-v1 no-response-friction
+                                                   # records as JSON Lines (one record per line) on stdout. --emit-signal
+                                                   # REQUIRES --summary: without it, default/--company mode already prints
+                                                   # a single parsable JSON document, and appending JSON Lines after it
+                                                   # would make stdout un-parsable as one JSON value. --summary's output is
+                                                   # not JSON to begin with, so appending JSON Lines after it stays clean —
+                                                   # bare --emit-signal (no --summary) exits 1 with this same explanation.
   node company-history.mjs --help                 # print this usage block and exit`;
 
 function parseArgs(argv) {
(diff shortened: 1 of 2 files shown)


Schlaflied:
Pushed a follow-up commit (`eae2675`) addressing all 4 CodeRabbit findings.

**1. `observedAt`/`sourceHash` instability (Major)** — `silentFactObservedAt()` now anchors on `fact.appliedDate`'s own month directly instead of `appliedDate + silenceWindowDays`. Since `--silence-window` is configurable, the old version could shift `observedAt` (and therefore `sourceHash`) for the *same* underlying silent fact depending on what window was in effect on a given run, which breaks dedup. Dropped the now-unused `silenceWindowDays` parameter and updated the call site + comment accordingly.

**2. severity/anchor diverged from the label computation (Major)** — Two related fixes in `buildNoResponseFrictionSignals`:
   - Stale facts are now filtered out of severity the same way `computeResponsiveness()` excludes them from the `silent-on-you` label (unless `--include-stale` is passed). Threaded a new `opts.includeStale` through the function; the CLI call site now passes its own parsed `includeStale` value.
   - The anchor fact is now chosen by comparing `appliedDate` strings (lexicographic, which works for ISO dates) instead of tracker row `num` — a backfilled/out-of-order row could previously anchor on an older application despite carrying a larger row number.
   - Added a self-test fixture with one active + one stale silent fact confirming severity stays `'single'` (matching what the label actually reflects), plus an out-of-order `appliedDate` fixture confirming the anchor is chosen correctly.

**3. Missing self-test coverage (Trivial)** — Added assertions for:
   - `silentFactObservedAt` returning `null` → `buildNoResponseFrictionSignals` skips that card entirely (no record, no crash).
   - `resolveRegion` mapping an unmapped country to `unmapped/<slug>`, plus a companion assertion for a mapped country going through `COUNTRY_REGION_MAP` normally (exercising both branches and `slugify`).

**4. `--emit-signal` mixed two output formats on stdout (Trivial)** — Without `--summary`/`--company`, `--emit-signal` used to print a pretty-printed JSON document followed by appended JSON Lines records, breaking single-value JSON parsing of stdout. `--emit-signal` now requires `--summary` and rejects `--company` (both of those already print a single JSON document); `parseArgs()` fails fast with a clear error otherwise. Documented the contract in the `USAGE` block.

**Test results:**
- `node company-history.mjs --self-test`: 74 passed, 0 failed
- `node company-history.test.mjs`: 87 passed, 0 failed
- `node test-all.mjs --quick`: 3663 passed, 0 failed (1 pre-existing warning, unrelated to this file — portal-scanner truncation notices / CV-sync fixtures)

Schlaflied:
Addressed both findings from the second CodeRabbit round in `498d844`:

**Finding 1 — `--emit-signal=<value>` silently accepted**
The unknown-flag check stripped an `=value` suffix before matching against `KNOWN_FLAGS`, so `--emit-signal=true` passed validation. But `emitSignal` was set via an exact-token check (`args.includes('--emit-signal')`), which is `false` for `--emit-signal=true` — so the flag was accepted as "known" yet the signal was never actually turned on. Added an explicit check that rejects any `--emit-signal=...` form with `Error: --emit-signal does not accept a value.` and exit 1. Covered by a new CLI smoke test in `company-history.test.mjs` (`--emit-signal=true`, `--emit-signal=1`, `--emit-signal=`).

**Finding 2 — `observedAt` not trimmed before slicing**
`silentFactObservedAt()` validated `fact.appliedDate` through `parseDate()` (which tolerates surrounding whitespace) but then sliced the raw, untrimmed string. A whitespace-padded `appliedDate` (e.g. `' 2026-06-01 '`) would pass validation but slice to a malformed `observedAt` (leading space, wrong characters) instead of `'2026-06'` — breaking the schema's required `YYYY-MM` format and shifting `sourceHash`. Now trims before slicing. Covered by a new case in `--self-test`.

Verification:
- `node company-history.mjs --self-test` — 75 passed, 0 failed
- `node company-history.test.mjs` — 90 passed, 0 failed
- `node test-all.mjs --quick` — 3663 passed, 0 failed (1 pre-existing, unrelated warning: Node's SQLite experimental-feature notice)

Pushed to `feat/no-response-friction-signal`.

santifer:
Slow first reply, @Schlaflied — sorry, and thanks for the patience. This one I read closely, and there's a detail in it I want to name because most people would have shipped past it:

```js
// `--emit-signal=true` passes the unknown-flag check silently, but the boolean
// read is an exact-token args.includes('--emit-signal'), which is false for it
// — so the flag would be ACCEPTED and nothing would be emitted.
```

A typo that produces exit 0 and no output is the worst possible shape for this feature, because the absence of records is indistinguishable from "no company matched". Someone would conclude their tracker has no responsiveness signal in it. You caught a failure mode that only exists in the seam between two correct pieces, and closed it by failing loudly.

The other thing I checked before anything else: **`--emit-signal` writes JSON Lines to stdout and nothing else.** No fetch, no upload, no endpoint. That matters more than it might look — a record carrying `companyKey`, `region` and `sourceHash` is shaped like something that could travel, and the difference between "computes a shareable record locally" and "shares it" is the whole ballgame in this project. Local, opt-in behind a flag, printed for the user to look at, is the correct answer and I'm glad it's the one you picked without being asked.

Implementing a ratified RFC signal by wiring up a fact `company-history.mjs` **already computes** is also the right economy: the raw signal existed since #1712, and what was missing was the schema, not the analysis.

**Where it stands:** approved on my side, not merged tonight — I've hit the per-session merge ceiling I hold myself to, so the tail of the queue rolls to the next cycle rather than riding on end-of-session judgement. This is near the top when I pick it back up, and it isn't blocked on Santiago the way #2791 is: it touches `company-history.mjs` and its test, nothing critical.

artemtrofymenko (commented):
Ran this against a real tracker rather than reading the diff, and it holds up well — the parts that are easy to get wrong are the parts that are handled. Three things below, one of which I think blocks conformance.

**What I verified by running it**

- `--emit-signal` alone exits **1** with a usage error rather than exiting 0 and printing nothing, which is the failure shape this feature could most easily have had.
- Output is byte-identical across runs, and sorted by `companyKey`.
- Nothing is written to disk on the emission path — checked with a before/after snapshot of the tree, and the only `writeFileSync` calls in the file are inside `runSelfTest`, into a `mkdtemp` dir it removes.
- No network: the file imports `fs`, `path`, `url`, `crypto`, `os`, `js-yaml` and three local modules, and nothing else.
- The stale filter mirroring `computeResponsiveness` is a real subtlety and the comment explaining it is right. Confirmed on a fixture: one active + one stale silent fact gives `single`, and `includeStale` turns the same fixture into `pattern`.
- Anchoring `observedAt` on `appliedDate` rather than row `num` also confirmed: a row with `num: 9` carrying an older date does not win the anchor.

On my tracker: two `silent-on-you` companies, two records, `observedAt` `2026-07` matching both `appliedDate`s.

---

**1. The record carries 7 of the 10 ratified fields**

Measured on the emitted output rather than on the diff:

```
emitted    companyKey, region, signalType, severity, sourceHash, observedAt, emittedBy
ratified   schemaVersion, companyKey, region, signalType, detail, severity,
           sourceDetector, sourceHash, observedAt, emittedBy
missing    schemaVersion, detail, sourceDetector
```

The ratified shape is your own revision in #1506, adopted by @santifer on 2026-07-16.

`schemaVersion` looks load-bearing: the PR title says *schema v1*, and a versioned interoperable format whose records do not carry their version cannot be migrated later without guessing which ones predate the change. `sourceDetector` was added specifically so provenance survives once several detectors emit into one format — and this is the third detector, so its enum (`interview-redflag | process-friction`) needs a value regardless. `detail` I have least of a view on: this signal is derived from dates and has no free text, so `null` or a fixed derived phrase both seem defensible.

Was the trimming deliberate? If so it is worth a line in the RFC thread, because the next implementer will read the ratified comment and not this PR.

**2. `severity` inside `sourceHash` makes one company-month emit twice**

`computeSourceHash` mixes in `severity`, so when a second application to the same company goes silent **in the same month**, the record's identity changes while the thing it describes does not:

```
1 application    single    2026-05    sha256:30381056a68e053…
2 applications   pattern   2026-05    sha256:b490ca7127da344…

same companyKey + observedAt, different sourceHash
```

Emission is a repeatable local command, so a user who emits in June and again in July puts both records into the pool for one company-month, and dedup cannot collapse them — one says `single`, one says `pattern`.

If severity is meant to be part of the identity, the pool needs a supersede rule and nothing in the record expresses one. If it is not, dropping `severity` from the hash makes `companyKey|observedAt` the stable key and a later record simply replaces the earlier one.

**3. Minor: `region: "unknown"`**

With no region in `profile.yml` this emits the literal string `unknown`, and #1506 uses `region` for the shared-layer directory layout. Worth deciding whether an absent region should be `"unknown"`, omitted, or a hard error — a directory named `unknown/` is a decision either way.

---

None of this touches the privacy posture, which is the part I checked hardest: no network, no writes, and `sourceHash` carries nothing the record does not already state in plaintext.

Schlaflied:
@artemtrofymenko Thank you for this — running it against a real tracker instead of just reading the diff is exactly how #2 got caught, and I don't think I'd have found it from the diff alone. All three addressed:

**1. `sourceHash` no longer includes `severity` (the blocking one).** You're right that this breaks dedup: the same company-month fact can legitimately go `single` → `pattern` as more applications go silent, and hashing severity in made that one fact produce two different hashes depending on when it was emitted. `computeSourceHash` is now keyed on `signalType|companyKey|observedAt` only, so `companyKey|observedAt` is the stable identity and a later emission with updated severity is meant to represent an update to that identity, not a phantom duplicate. Added a self-test using your exact scenario — a 1-application and a 2-application fixture for the same company/month now produce the *same* `sourceHash` despite `single` vs `pattern` severity.

**2. Full 10-field schema v1 now emitted.** Added `schemaVersion` (`1`), `detail`, `sourceDetector`. On your two open questions:

- **`sourceDetector` enum value: `no-response-friction`.** The RFC thread only names `interview-redflag` and `process-friction`, so I'm establishing this as the third value here rather than finding it pre-named anywhere — flagging that explicitly since you asked. I went with `no-response-friction` (matching the signal's own `signalType`) because that's the existing convention in this schema already: `sourceDetector == signalType` for a detector that only emits one signal type. Happy to reconcile if a different value gets picked when a fourth detector shows up.
- **`detail: null`, deliberate.** This signal is derived purely from tracker dates (application date vs. silence window) — there's no free text anywhere in the pipeline it could carry, unlike e.g. `process-friction`'s `"interviewer no-show, no reschedule notice"` example in the RFC. `null` felt more honest than inventing a fixed derived phrase that would just restate `severity`/`observedAt` in prose. Wasn't deliberate trimming before — it was a real gap, now closed, and I added a line in the RFC thread's spirit here rather than silently fixing it.

**3. `region: "unknown"` replaced with skip-and-warn.** `resolveRegion()` now returns `null` (not the string `"unknown"`) when it can't resolve a region from `config/profile.yml`, and `buildNoResponseFrictionSignals()` skips emission for the affected company with a warning (`region could not be resolved from config/profile.yml — signal not emitted for {companyKey}`) instead of fabricating a value that would imply a real `unknown/` directory in the shared-layer's region-scoped layout. This only skips the specific record — it doesn't error out the whole command over one unresolvable field. Added self-test coverage for the skip-with-warning path.

All landed in 56d7b6d, pushed to this branch. Self-tests: `node company-history.mjs --self-test` 84/84, `node company-history.test.mjs` 90/90, full `node test-all.mjs --quick` 3663 passed / 0 failed. Let me know if the `sourceDetector` naming choice looks wrong from where you're sitting — genuinely undecided territory since the RFC didn't cover it.


---
**Update:** a follow-up review caught one more spot with the same untrimmed-whitespace class of bug — the anchor-selection comparison in the `silentFacts.reduce()` above also needed a `.trim()` on both sides before comparing `appliedDate` strings, for the same reason `silentFactObservedAt` did earlier in this thread. Fixed and pushed in `22b8668` (new regression test: a whitespace-padded but more-recent date still wins the anchor over a clean-but-older one). Full suite still green: `node company-history.mjs --self-test` 85/85, `node test-all.mjs --quick` 3663 passed / 0 failed.

Schlaflied:
Addressed the 3 findings from the latest CodeRabbit review (the one reviewing `56d7b6d..22b8668` — the other review in this thread was against a stale commit range and re-flagged an already-fixed issue, no action needed there):

1. **Minor, real bug**: `silentFacts` filtered out unusable-`appliedDate` facts only implicitly (via the anchor's `observedAt` check), not before anchor selection/severity counting. A malformed date could win the lexicographic reduce, or even losing the reduce could still inflate severity to `pattern`. Now filtered via `silentFactObservedAt(f) !== null` before either step. New test: a card with one valid + one malformed fact emits exactly one record, anchors on the valid fact, and reports `severity: 'single'`.
2. **Trivial, test gap**: `includeStale: true` is now exercised end-to-end through `buildNoResponseFrictionSignals` (previously only the default-exclusion path was tested at that level) — confirms `severity` correctly flips to `'pattern'` when the stale fact is included.
3. **Trivial, test gap**: added fixtures proving `sourceHash` actually differs when `companyKey` differs (same month) and when `observedAt` differs (same company) — the existing determinism test proved repeatability but not that the hash is sensitive to its own stated identity inputs.

`node company-history.mjs --self-test`: 91/91. `node test-all.mjs --quick`: 3663 passed, 0 failed.

artemtrofymenko:
`sourceDetector` — I land on `company-history`, though the honest starting point is that the existing enum does not settle it either way.

When you introduced the field you wrote:

> **Added `sourceDetector`.** This schema now needs to represent output from both #1233 (interview-redflag) and #1467 (process-quality), so a record needs to say which one produced it — otherwise there's no way to trace provenance once both tools are emitting into the same format.

The two producers are named there as tools — `interview-redflag` and `process-quality` — but the enum as written is `interview-redflag | process-friction`. So one value is a tool name and the other is a signal name that differs from its own tool. There is no convention to follow; there are two precedents pointing in different directions.

That leaves the field's stated purpose as the tiebreak, and it is explicit: *which one produced it*. The producer here is `company-history.mjs`.

Two things push the same way. `interview-redflag` emits four signal types by the RFC's own enum — `scope-ambiguity`, `defensive-closure`, `evaluator-competency-gap`, `process-signal` — and is named after none of them, so the one precedent where a detector has more than one output already resolves as the tool. And `company-history.mjs` computes two independent axes today; if a second tracker-derived signal ever lands there, `no-response-friction` as a *detector* name is either wrong for it or needs renaming, and renaming an enum value already present in emitted records is the expensive kind of change. `company-history` stays correct however many signals the script grows.

Weak preference, not a blocker — you asked where I land, and that is where. Either value traces provenance adequately for the two detectors that exist today.

**Verified on `edd8d16`:** all ten fields present, `schemaVersion: 1`, `detail: null`; the 1-application and 2-application fixtures for the same company-month now produce **the same** `sourceHash` while severity still moves `single` → `pattern`; a different company and a different month each still produce a different hash; `resolveRegion` gives `north-america/canada` for a mapped country and `unmapped/spain` for an unmapped one, keeping the two-level shape the layout expects rather than flattening it. Self-tests 91/91, `company-history.test.mjs` 90/90.

One correction to my own earlier framing, since I raised it: I checked whether skip-and-warn leaves the feature silent on a default install, and it does not — `location.country` ships uncommented in `profile.example.yml`. The empty output I saw first was my sandbox having no `config/profile.yml` at all, not the fix.

Schlaflied:
@artemtrofymenko Thanks for the reasoning on `sourceDetector` — the `interview-redflag.mjs` precedent (tool name, not signal name, even across its 4 signal types) settles it, and you're right that renaming this cheaply now beats discovering the same argument later after real records exist with the old value.

Renamed `sourceDetector` from `'no-response-friction'` to `'company-history'` everywhere — the emission code, the self-test assertion, and the schema-shape comment (now explains the tool-name-not-signal-name reasoning and the future-proofing against `postingChurn` eventually feeding a second signal type).

`node company-history.mjs --self-test`: 91/91. `node test-all.mjs --quick`: 3663 passed, 0 failed.

Also appreciated the independent re-verification pass on `edd8d16` — good catch on your own earlier framing too (the `skip-and-warn` silence was your sandbox missing `config/profile.yml`, not the fix).

santifer:
Clean schema work, @Schlaflied: the REQUIRES --summary reasoning (keeping stdout parsable as one JSON value in default mode), the loud rejection of --emit-signal=value instead of silently emitting nothing, and region resolution tested against mapped/unmapped/absent profiles. First real producer of the #1506 schema in the tree. Merged 🚀

PR #2791, opened 2026-08-13
feat(oferta/pdf): enforce JD archival — Closes #2789

### LINKED ISSUE #2789: feat(oferta/pdf): enforce JD archival — reports store a URL, not the posting text, and it silently rots
## Problem

`pdf.md` step 4 already instructs writing the JD to `jds/{slug}.md` "if it isn't already one" before running `jd-skill-gap.mjs`. In practice this is a prompt-only instruction with no code enforcement — and it shows: on a tracker with 7 saved reports, `jds/` is empty (only `.gitkeep`). Reports themselves only store the posting's **URL** (`**URL:**` header), never the JD's actual text.

A URL is a live pointer, not an archive. Once a posting closes or gets taken down — which reliably happens somewhere in the weeks between applying and interviewing — there's no way to recover what the JD actually asked for. Concretely: preparing for a later-stage interview this week required re-fetching a still-live posting via a web search as a lucky save; had it already closed, there would have been nothing to reconstruct the actual qualifications from except the candidate's own memory and whatever got hand-copied into `cv.md`/prep notes at generation time.

## Why this wasn't caught

`oferta.md`/`pdf.md` are prompt-driven modes — the report file itself is written directly by the agent via a file-write tool in the same turn, not produced by a dedicated generation script. "Also archive the JD" is one more thing the agent has to remember to do unprompted, with nothing downstream that would notice if it didn't happen. Compare to `**URL:**`, which AGENTS.md documents as a required report header but which also has no code-level check enforcing it — same class of gap, same failure mode.

## Proposal

1. **Make JD archival part of the report itself, not a separate step that's easy to skip.** Every report should embed the JD's verbatim text (or a link to a `jds/{slug}.md` file created in the same turn) in a dedicated section, e.g. `## Job Description (archived verbatim)`. Embedding directly in the report is more robust than a side-file, since the report is the one artifact that's already guaranteed to get written and tracked.
2. **Add a validator** (same shape as `check-table-freshness.mjs`) that scans `reports/*.md` and flags any report missing a JD-archive section or corresponding `jds/{slug}.md` file. Wire it into `test-all.mjs` so a missing archive shows up as a test failure going forward, not a silent gap discovered months later during interview prep.
3. Update `modes/oferta.md` and `modes/pdf.md` to state this as a required step (not "if it isn't already one" — always), matching the phrasing style already used for other MUST-have report fields.
4. Out of scope: backfilling JD text for the 7 existing reports that already lack it — that data is gone (postings may already be closed); this only prevents the gap going forward.

## Test plan

- Fixture: a report file missing both the archive section and a matching `jds/{slug}.md` → validator flags it.
- Fixture: a report file with the archive section present → validator passes.
- Fixture: a report file without the archive section but with a matching `jds/{slug}.md` file → validator passes (either form counts).
- `node test-all.mjs` stays green with the new check wired in.

Schlaflied:
## Summary

- Reports only store the posting's `**URL:**` — a live pointer that rots once the posting closes, losing the actual JD text with no way to recover it during later interview prep. `modes/pdf.md` step 4 already told the agent to write the JD to `jds/{slug}.md` "if it isn't already one," but that instruction had zero code enforcement: on a real tracker with 7 saved reports, `jds/` was completely empty.
- Adds `check-jd-archive.mjs`, a read-only validator (same shape/conventions as `check-table-freshness.mjs`) that scans `reports/*.md` and flags any report missing a JD archive. A report counts as archived when it has EITHER a `## Job Description` section with substantive archived text, OR a matching `jds/` capture — resolved via `jd-capture.mjs`'s existing `findCaptureForReport` (the same report-number lookup `outcome.mjs` already relies on), keyed on the report number and company slug parsed straight from the report's own filename (`{###}-{company-slug}-{YYYY-MM-DD}.md`). No new slug helper was invented.
- Wires the validator into `test-all.mjs`: its `--self-test` runs alongside the other zero-LLM checkers' self-tests, plus a dedicated wiring/read-only-boundary integration block (mirroring the `check-table-freshness.mjs` pattern) that pins SYSTEM_PATHS registration, the `npm run jd-archive` script, docs, and the read-only fs-import boundary.
- Updates `modes/oferta.md` and `modes/pdf.md`: the report template now has a required `## Job Description (archived verbatim)` section (the primary mechanism — the report is the one artifact guaranteed to get written and tracked), and `modes/pdf.md`'s `jds/{slug}.md` write is no longer conditional ("if it isn't already one") — it always happens. A `jds/` capture (ideally via `archive-posting.mjs --report=N`) remains an acceptable alternative for a very long JD or a standalone `jd-skill-gap.mjs` run outside a full evaluation.
- Registers the script in `package.json` (`npm run jd-archive`), `update-system.mjs`'s `SYSTEM_PATHS`, `docs/SCRIPTS.md`, and `AGENTS.md`'s Main Files table.

**Explicitly out of scope** (per the issue): backfilling JD text for pre-existing reports. Those live in a gitignored, user-owned `reports/` directory — private data, not something a code PR should touch — and the postings may already be gone. This only prevents the gap going forward.

## Design notes

- `jds/` has four coexisting filename conventions documented in AGENTS.md's "JD captures" section, none canonical. Only the report-number-prefixed form (`archive-posting.mjs --report=N`) is reliably resolvable back to a specific report — the others (date-prefixed, sha1-suffixed, bare company-role slugs) have no report number to key on. The validator only credits that reliable form, which matches AGENTS.md's existing "prefer `--report=N`" guidance.
- The embedded-archive check requires ≥40 non-whitespace characters after stripping HTML comments, so an empty placeholder heading or a "TBD" stub doesn't pass as an archive.
- The validator never writes anything — reports/jds are only read via `readFileSync`/`readdirSync`/`existsSync`. A dedicated `test-all.mjs` block asserts this boundary structurally (no write-capable fs import, no `require()`, no dynamic `fs` import), matching the pattern already used for `check-table-freshness.mjs` and `rejection-latency.mjs`.

## Test plan

- [x] Fixture: report with no archive section and no matching `jds/` capture → flagged `missing-jd-archive`.
- [x] Fixture: report with a `## Job Description (archived verbatim)` section → not flagged.
- [x] Fixture: report without the archive section but with a matching `jds/{report#}-...` capture → not flagged (either form counts).
- [x] Fixture: `jds/` capture with the right report-number prefix but a *different* company slug → still flagged (company-slug guard, no false credit).
- [x] `--summary` renders a human-readable table; default output is JSON.
- [x] `--help` and flag parsing go through `lib/cli-flags.mjs`'s `flagValue`, matching the sibling scripts' convention.
- [x] `node check-jd-archive.mjs --self-test` — 22/22 passed.
- [x] `node test-all.mjs --quick` — 3678 passed, 0 failed (1 pre-existing, unrelated Node `ExperimentalWarning` about SQLite).

## Summary by CodeRabbit

- **New Features**
  - Added validation to ensure reports include complete archived job descriptions or matching archive files.
  - Job descriptions are archived automatically during evaluations, including standalone runs.
  - Archives record the visible posting date or indicate when it was unavailable.
  - Added JSON, summary, directory override, help, and self-test options.
  - Validation identifies missing archives and returns a failure status when requirements are unmet.

- **Documentation**
  - Updated report requirements, script references, and validation coverage.


Schlaflied:
Addressed all 3 CodeRabbit findings plus the CodeQL alert:

1. **Section parsing truncated on nested headings** — `hasEmbeddedJdArchive()` now terminates the JD section only on a known report-section heading (Machine Summary, Keywords extracted, `A)`/Block A, Risk Summary, etc.), not any `## ` line. A verbatim JD containing its own `## Responsibilities` sub-heading no longer gets falsely flagged as missing.
2. **Unfilled template placeholder passed validation** — the unfilled `modes/oferta.md` placeholder text is 66 characters, longer than `MIN_ARCHIVE_CHARS`. Now explicitly rejected regardless of length.
3. **`jds/` alternative required a report-number link** — `modes/oferta.md` now says a full evaluation report needs `archive-posting.mjs --report={num}` (or another `{num}-...` capture) for the `jds/` alternative, matching what `check-jd-archive.mjs`'s `findCaptureForReport` actually credits. A slug-only `jds/{slug}.md` remains valid only for `jd-skill-gap.mjs` run standalone with no report to link back to.
4. **CodeQL — incomplete multi-character sanitization**: the HTML-comment strip in `hasEmbeddedJdArchive()` now loops to a fixed point instead of a single `.replace()` pass, so a nested/overlapping `<!--...-->` fragment can't reconstruct a comment marker the single pass never re-scanned.
5. **Self-test body extraction was greedy** (`test-all.mjs`) — replaced with a brace-counting scan for `runSelfTest()`'s actual body, so the read-only fs-boundary check scans everything outside it instead of a greedy regex silently dropping trailing code from the check.

Also folded in the posting-date archival requirement into `modes/pdf.md` (was only in `oferta.md` before).

New self-tests: nested-heading case, unfilled-placeholder-rejection case. `node check-jd-archive.mjs --self-test`: 25/25. `node test-all.mjs --quick`: 3678 passed, 0 failed, 1 pre-existing unrelated warning.

Schlaflied:
Addressed the round-3 finding: a `jds/` path-only pointer sentence (e.g. `"See jds/042-acme-2026-01-15.md (archive-posting.mjs --report=042) for the full archive."`) easily cleared `MIN_ARCHIVE_CHARS` on length alone, so a wrong or nonexistent path would pass without ever being resolved. Any archive section referencing a `jds/*.md` path is now treated as using the pointer mechanism rather than credited on length alone — `checkJdArchive()` falls through to its existing `findCaptureForReport()` resolution, which actually verifies the capture exists for that report.

New self-tests cover both directions: a pointer to a nonexistent capture is flagged, a pointer to an existing capture resolves and is not flagged, and substantive real prose with no `jds/` reference is still credited normally.

`node check-jd-archive.mjs --self-test`: 29/29. `node test-all.mjs --quick`: 3678 passed, 0 failed, 1 pre-existing unrelated warning.

Separately — the `windows-latest` CI failure on the "concurrent adds do not lose items" test (`agent-inbox`) is unrelated to this PR; this diff never touches that file. Passes locally on repeated runs; looks like Windows-runner-specific flake in that test's own file-locking timing, not something introduced here.

Schlaflied:
Fixed the round-4 finding: `isPathOnlyPointer()` now requires the entire section (minus an optional leading `Posted: ...` line) to match nothing but the canonical pointer sentence — `See jds/{filename} for the full archive (archive-posting.mjs --report={num}).` — before treating it as a pointer needing `findCaptureForReport` resolution. Substantive JD prose that merely mentions a `jds/*.md` path in passing (e.g. a "see also" reference) is now credited directly as archived text, as it should be.

`modes/oferta.md` now specifies this exact sentence as the required phrasing for the pointer form, so the check's strict match is predictable.

New self-test: substantive JD prose + an incidental `jds/*.md` mention is still credited.

`node check-jd-archive.mjs --self-test`: 30/30. `node test-all.mjs --quick`: 3678 passed, 0 failed, 1 pre-existing unrelated warning.

santifer:
Sorry for the slow first reply, @Schlaflied — deep backlog week. Thanks for waiting, and for opening #2789 before the PR.

The diagnosis is the strongest part, and it's a general one dressed as a specific bug:

> this is a prompt-only instruction with no code enforcement — and it shows: on a tracker with 7 saved reports, `jds/` is empty (only `.gitkeep`)

That's the whole argument, and it's checkable rather than arguable. An instruction in a mode file is a request the model can decline, forget, or reinterpret, and nothing downstream can tell the difference between "declined" and "wasn't needed". You didn't reason about whether it would be followed — you went and counted, and found seven chances taken and zero taken up.

The second half is why it's worth fixing rather than noting: **a URL is a live pointer, not an archive.** The gap between applying and interviewing is exactly where postings come down, so the artifact rots precisely when it's finally needed. Your own case — recovering a still-live posting by web search as a lucky save — is the failure with the luck removed.

Two design calls I'd have argued for and you already made:

- **The report is the primary archive, `jds/` the alternative.** Right choice: the report is the one artifact guaranteed to be written and tracked, so putting the archive where the thing already goes beats adding a second place that has to be remembered.
- **A validator wired into `test-all.mjs`, not just a doc line.** Otherwise the fix for a prompt-only instruction is... another prompt-only instruction.

**Where it stands:** not merging tonight, and not because of the code. It touches `AGENTS.md` — the canonical source of agent instructions — and adds a script plus an npm entry, and making JD archival **REQUIRED** changes what every evaluation writes for every user. That's a product decision, so it goes to Santiago rather than through my checklist. I've read it, I'm putting it to him with the seven-reports-zero-archives count as the argument, and I'll come back with an answer either way.

One question worth having ready for that conversation: what happens to the **existing** reports written before this rule? A validator that starts red on everyone's historical reports on upgrade day is a different proposition from one that applies going forward, and I'd rather hear your intent than assume it.

Schlaflied:
@santifer On the existing-reports question — I don't have a fully settled answer, genuinely unsure between a couple of shapes, so laying out where my thinking is rather than picking one and defending it.

My instinct is that "old vs new by date" is the wrong axis. The actual risk this whole PR is about — a posting closing before the JD is needed again — only exists for applications that are still *live*. A report for a role you were rejected from months ago carries basically zero ongoing risk; you're not going back to re-read that JD. A report for a role still sitting in Evaluated/Applied/Responded/Interview/Offer is exactly the case this validator exists to protect, whether the report predates the rule or not.

So the split I'd lean toward maps onto the tracker's own canonical states rather than a creation date:

- **Terminal states** (Rejected, Discarded, SKIP, Hired) — no retroactive enforcement. The report is done; flagging it red on upgrade day would just be noise with no actual risk being mitigated.
- **Live states** (Evaluated, Applied, Responded, Interview, Offer) — retroactive `review-due`-style soft warning (never blocks CI, matches `check-table-freshness.mjs`'s existing severity split), since these are the reports where losing the JD would genuinely hurt.

That said, I'm not confident this is right over something simpler like a straight cutoff date, and I'd rather hear where you land before building either — especially since you'd know better whether joining `check-jd-archive.mjs` against tracker status is more machinery than this is worth for what's ultimately a one-time upgrade-day question.

santifer:
Your instinct is right, @Schlaflied, and the reasoning is the part I'd keep verbatim in the code comment: the risk this validator exists for — a posting dying before the JD is needed again — only exists while the application is *live*. A creation-date cutoff would exempt exactly the wrong rows (old but live applications, the highest-risk case) while flagging fresh terminal ones that carry none.

So: build the state-based split you described.

- **Terminal states** (Rejected, Discarded, SKIP, Hired): no retroactive enforcement.
- **Live states**: soft `review-due`-style warning, never blocking CI — matching `check-table-freshness.mjs`'s severity split, as you proposed.

On the "is the tracker join more machinery than it's worth" worry: you already have the canonical state ids from `templates/states.yml` and the tracker parse lives in `tracker-parse.mjs`, so the join is a lookup, not new infrastructure. Keep it read-only and fail soft (a row the join can't resolve gets the warning, not an error — same asymmetry as the rest of the PR).

With that version in, this is mergeable from my side.

Schlaflied:
@santifer — both blockers addressed:

**Conflict:** rebased onto `upstream/main` (550 commits of drift). Two real conflicts, both "both sides added something new in the same spot": `AGENTS.md` (this branch's JD-archival-required bullet landing next to a new Hired Wall section) and `test-all.mjs` (this branch's own section 72 colliding with three new sections main had added since the fork — formatRunFailure, funding-manifest, gmail isCleanUrl). Kept both sides in each case; renumbered the JD-archive wiring section to 75 to land after main's new ones.

**Retroactive-enforcement state split:** built exactly what you signed off on. `checkJdArchive()` now takes an optional `trackerPath`; when given one, it joins each report number against its own `data/applications.md` row via `tracker-parse.mjs`'s `extractTrackerReportNumbers` (the same lookup `set-status.mjs --report N` uses) and resolves the row's status against `templates/states.yml` via `tracker-utils.mjs`'s `loadCanonicalStates`/`resolveCanonicalState` — no second parser, no second alias table.

- **Terminal** (`Rejected`, `Discarded`, `SKIP`, `Hired`) — no finding at all, not even a warning. (`Offer` stays out of this set on purpose, even though `states.yml` marks it `terminal: true` for lifecycle-ordering elsewhere — an offer in hand is still a live, undecided application.)
- **Live**, or an **unresolvable join** (no linking row, 2+ rows linking the same report number, unreadable tracker/states.yml) — a soft `jd-archive-review-due` finding, mirroring `check-table-freshness.mjs`'s `review-due` shape exactly (same mixed-findings-array-with-a-`type`-field pattern, same "never blocks CI" exit-code rule). Ambiguous and no-row-at-all both fail soft to this, per your instruction.
- **No tracker to join against at all** (fresh install, `--no-tracker`, or any existing fixture-only self-test that never passes a `trackerPath`) — unchanged legacy hard `missing-jd-archive` behavior. This is what keeps going-forward/fresh-install enforcement exactly as strict as before; the retroactive softening only ever activates when there's an actual tracker row to reason about, never as a blanket downgrade.

New CLI flags: `--tracker <path>` (override), `--no-tracker` (explicit opt-out, same effect as omitting the tracker entirely).

New self-tests cover all four outcomes on one synthetic tracker fixture (terminal/live/unresolved/ambiguous) plus a same-fixtures-no-tracker control proving the legacy hard path is untouched. `node check-jd-archive.mjs --self-test`: **39/39** (30 pre-existing + 9 new).

The rebase also surfaced two things `main` grew after this branch forked, both now fixed: `check-jd-archive.mjs` was resolving `reports/`/`jds/` from the codebase root instead of `getCareerOpsRoot()` (caught by the user-layer-path drift-guard), and its main-module guard used the raw `process.argv[1]`/`import.meta.url` comparison instead of `lib/is-main-module.mjs`'s `isMainModule()` (caught by `main-guard-convention.test.mjs`). Both are unrelated to the retroactive-enforcement work itself, just debt the rebase made visible.

`node test-all.mjs`: **7972 passed, 0 failed, 14 warnings** (all pre-existing fixture/mock-scenario console output, unrelated to this PR — same shape as the "1 pre-existing unrelated warning" noted in earlier rounds).

`gh pr view` now reports `mergeable: MERGEABLE`.

santifer:
Sorry for the two-day silence, @Schlaflied: you did the hard part (550 commits of drift, two real conflicts) and then heard nothing. Thanks for your patience.

I read it properly and the design holds up on the three things I went looking for:

- **It doesn't break existing users.** `test-all` runs `check-jd-archive.mjs --self-test` against fixtures in a tmpdir, not a sweep of the user's own `reports/`. Someone upgrading with 40 historical reports doesn't suddenly get a red suite. That distinction is easy to get wrong and you got it right.
- **The terminal join does what we agreed on 17-ago.** Rejected/Discarded/SKIP/Hired are exempt, an unresolvable row counts as live, and the reasoning is in the code rather than in a PR comment that nobody reads later.
- **`## Job Description (archived verbatim)` is additive to the report contract.** §55.4 passes: the A-G blocks are intact and the assertion explicitly allows appending. Merged with current main it's 8135 passed, 0 failed.

The prose in `modes/oferta.md` about the canonical pointer sentence is the part I'd have gotten wrong: reading a section as a pointer *only* when it contains nothing but that sentence, so mixed prose reads as the archive itself, is the kind of rule that saves a support thread later.

One thing left, and it isn't about your code. The report template is a surface our first-party web also **writes**, not just reads, so a new required section can leave its generated reports flagged by `npm run jd-archive` while nothing in CI goes red. I've asked that side whether their prompt path dictates the report format. If it does, the two land together; if not, this goes in as-is. Either way you'll hear back within a day, not two.

Branch is rebased onto current main so the CI you're seeing is on a real base rather than a stale one: that one's on me, not you.

santifer:
Merged, @Schlaflied. 🚀

Three rounds and 550 commits of drift to get here, and the part I'd point anyone at is the terminal-state join: exempting Rejected/Discarded/SKIP/Hired while an unresolvable row counts as live is the version that protects the case that actually matters, and the reasoning lives in the code rather than in a review comment nobody reads a year from now.

One thing surfaced after my last comment that belongs on the record, and the credit is career-ops-ui's, not mine. `check-jd-archive` validates **length**, and `MIN_ARCHIVE_CHARS = 40` is cleared comfortably by a LinkedIn authwall (*"Sign in to view this job"*). On a headless fetch that returns 200 with a login screen, the wall gets archived under a heading that promises `(archived verbatim)` and the check goes green. That's worse than no archive: without the section a user knows they don't have the JD; with it they believe they have the posting saved for a negotiation and what they have is a login page.

It fits your design rather than fighting it, which is why I'm raising it as a follow-up instead of holding the merge: you already reject two markers of non-content (the unfilled template placeholder, the pointer sentence). A fetch-failure marker is the third item on a list you'd already built twice.

I'm opening the follow-up issue now and it's yours first if you want it. The web side is fixing their half in parallel so the failed fetch declares itself before it ever reaches the validator.

Issue #1706, opened 2026-07-09
Updater bootstrap fails on old→1.18.0 jump: new update-system.mjs imports scaffolder/bin/skill-entrypoints.mjs before it's checked out


CitruXonve:
## Environment
- Local version: 1.11.0
- Target version: 1.18.0
- Node: v24.14.0 (macOS)

## What happens
Running `node update-system.mjs apply` fails every time with:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '.../scaffolder/bin/skill-entrypoints.mjs'
  imported from .../update-system.mjs
Updater self-reexec failed
```

## Root cause
The v1.11.0 `apply()` self-reexec checks out only `update-system.mjs` from `FETCH_HEAD`, then re-execs it. The 1.18.0 `update-system.mjs` has a **static** import (line ~25) of `./scaffolder/bin/skill-entrypoints.mjs`. That file exists upstream but hasn't been checked out yet, so the new updater dies at module-load — before it ever reaches its own `BOOTSTRAP_PATHS` / `REEXEC_FALLBACK_FILES` logic.

`REEXEC_FALLBACK_FILES = ['update-system.mjs', 'scaffolder/bin/skill-entrypoints.mjs']` (line ~449) fixes this **going forward**, but doesn't help anyone on a pre-scaffolder updater making the jump *to* 1.18.0.

## Impact
Every user on a version whose updater predates the `scaffolder/` dependency is blocked from updating via the sanctioned path. The failure is deterministic — re-running `apply` reproduces it identically, and the updater cannot self-heal because it crashes at module load.

## Suggested fixes
1. Make the scaffolder import **dynamic/lazy** (`await import()` after bootstrap) so module load never fails.
2. Or have the old updater check out the full new `scaffolder/` tree *before* re-execing, not just `update-system.mjs`.
3. Or wrap the re-exec import in a try/catch that bootstraps missing deps and retries.

## Workaround
```
git checkout FETCH_HEAD -- scaffolder/
node update-system.mjs apply
```

Issue #18, opened 2026-04-06
Add German (DE) language modes — DACH market needs unique vocabulary and comp checks


OthmanAdi:
### Problem

`modes/` currently has English and Spanish files but no German. Germany is the largest tech labour market in the EU, and a German `Stellenanzeige` has vocabulary and compensation structure the existing modes do not encode.

Concrete examples of what gets lost when running EN modes on a German posting:

- "13. Monatsgehalt" (the year-end bonus) needs to be added to the brutto comp calculation. The English modes do not know to look for it.
- A 6-month "Probezeit" is the market default in DE, not a red flag. The English modes have no calibration for that.
- "Festanstellung" vs "Freelance" pricing differ structurally — freelance Tagessätze in DACH are typically 30-60% above the equivalent employee brutto rate (to cover Sozialversicherung, Urlaub, Akquise, Steuerberater).
- "Tarifvertrag" coverage (TVöD, IG Metall) reduces negotiation room but adds security. Worth flagging in Block C.
- "AGG" (anti-discrimination law) means certain question types in cover letters / forms must be handled carefully.
- "Bewerbung" expectations differ: Anschreiben + Lebenslauf + Zeugnisse, in that order.

The English modes can technically read a German posting, but they apply US/UK calibration to it and miss all of the above.

### Proposed solution

Add a `modes/de/` directory with native German translations of the four highest-leverage modes:

- `_shared.md` — translation of the EN version, plus a new "Deutscher Markt — Spezifika" section covering AGG, 13. Monatsgehalt, Probezeit, Kündigungsfrist, Urlaub, Tarifvertrag, Betriebsrat, Arbeitszeugnis, VWL, bAV.
- `angebot.md` — translation of `oferta.md` (the ES original is the most up-to-date). Preserves the full Schritt 0 → Block A-F → Block G structure. Adds DACH-specific Pflichtchecks under Block D.
- `bewerben.md` — translation of `apply.md`. Adds a section on common German form fields (Gehaltsvorstellung brutto/jährlich, Eintrittsdatum + Kündigungsfrist, Arbeitserlaubnis, Sprachkenntnisse nach GER).
- `pipeline.md` — translation of `pipeline.md`. Adds notes on StepStone / XING / kununu / Bundesagentur für Arbeit as DACH portal sources.
- `README.md` — explains activation, vocabulary cheat sheet, what intentionally stays in English (tool names, status values, code paths).

[...]

Issue #17, opened 2026-04-06
Default `portals.example.yml` is US-centric — most European AI labs are missing


OthmanAdi:
### Problem

The current `templates/portals.example.yml` ships with ~45 companies, of which only 4 are DACH (n8n, Langfuse, Cognigy, Parloa) and a handful more are EU (Mistral, Factorial, Travelperk, Attio, Tinybird, Clarity AI, PolyAI, Speechmatics, Talkdesk). The rest is US-heavy.

Meanwhile a real share of the AI labs people actually want to work at are European and missing from the seed list:

- **Germany:** Aleph Alpha (Heidelberg), Helsing (Munich), Black Forest Labs (Freiburg), DeepL (Cologne), Celonis (Munich), Contentful, GetYourGuide, HelloFresh, N26, Trade Republic, SumUp, Forto
- **Switzerland:** Lakera (Zurich), Scandit, Cradle
- **France:** Hugging Face (Paris), Photoroom, Pigment, Qonto
- **UK:** Wayve, Isomorphic Labs, PhysicsX, Stability AI, Synthesia, Faculty, Causaly
- **Nordics:** Lovable, Legora, Spotify, Vinted
- **Iberia:** Amplemarket

A Berlin / London / Paris user running `/career-ops scan` against the default config mostly sees California companies and misses the labs in their own city. The `title_filter.positive` list is also English-only, so a "KI Engineer" role at Aleph Alpha gets filtered out today.

### Proposed solution

Add ~30 verified DACH/EU AI companies to `templates/portals.example.yml`, all with live Greenhouse API endpoints or verified Ashby/Lever boards (so the scanner can actually reach them, not just fake URLs). Add 6 German keywords to `title_filter.positive` (`KI`, `Künstliche Intelligenz`, `KI Engineer`, `KI Trainer`, `Dozent`, `Weiterbildung`) so German titles stop getting filtered out. Drive-by: upgrade the existing Speechmatics entry from `scan_method: websearch` to its real Greenhouse `api:` endpoint.

Honest scope:
- Helps EU users a lot (their seed scan suddenly returns local companies)
- US users get a few new leads from companies that hire heavily in both regions (Wayve, Helsing, Mistral, Hugging Face, Synthesia, Faculty, Stability AI, Black Forest Labs)
- Zero downside for US users — every entry is `enabled: true` per company, so they can flip individual ones off if the noise is too high

[...]

PR #2793, opened 2026-08-13
feat(update): declare fork-local paths outside the system layer (#2421)

### LINKED ISSUE #2421: Local-only files force an edit to update-system.mjs, so every user with one carries a permanent merge conflict
## What happens

`SYSTEM_PATHS` and `USER_PATHS` are hardcoded arrays in `update-system.mjs`, and `validate-system-paths-coverage.mjs` fails the suite when a tracked file appears in neither. That combination is correct — it is what stops a new system file from silently escaping updates — but it has a side effect for anyone who keeps a file the upstream repo does not ship.

To declare such a file, the user has to edit `update-system.mjs` itself. That file is system layer: `update-system.mjs apply` overwrites it, and anyone tracking `main` with git merges it on every sync. So the only supported way to say "this file is mine" is to modify a file that the update mechanism is actively maintaining.

Concretely, in my checkout the block is:

```js
const USER_PATHS = [
  // ...upstream entries...
  // ── Local fork — user-owned, never overwritten by an update ──
  '.gitattributes',            // forces LF on *.sh so Git Bash can parse them
  '.mcp.json',                 // per-project MCP server config
  'run-nightly.ps1',           // local PowerShell nightly orchestrator
  'check-fork-invariants.mjs', // guards local hardening across merges
  'qa-fixtures/',              // frozen QA corpus an update must never overwrite
];
```

None of these exist upstream. None of them are interesting to anyone else. But they have to live inside a file you rewrite regularly, so they conflict on every sync, and a resolution that drops them silently re-exposes those files to being overwritten by the next `apply`.

## Why it is worth fixing upstream

The failure is quiet in the bad direction. A conflict resolution that takes upstream's `USER_PATHS` wholesale looks clean — tests pass, nothing errors — and the next `update-system.mjs apply` overwrites the user's local files. There is no signal until the data is gone.

It also affects more people than fork maintainers. Any user who adds a local automation script, a project-scoped `.mcp.json`, or a fixture directory hits it as soon as they run the test suite, because `validate-system-paths-coverage.mjs` crashes on the unclassified file.

## Proposal

Read additional user paths from a sidecar file, and treat a missing file as an empty list:

- `config/user-paths.txt`, one path per line, `#` comments allowed, same semantics as the existing arrays (trailing `/` means directory prefix)
- append its entries to `USER_PATHS` at load
- teach `validate-system-paths-coverage.mjs` to honour it, so declaring a local file is enough to satisfy coverage
- add the sidecar itself to `USER_PATHS`, so an update never overwrites the list of things an update must not overwrite

## Failure modes worth designing for

- **Sidecar shadows a system path.** If a user lists `merge-tracker.mjs`, that file silently stops receiving updates and they get no new fixes with no warning. Suggest rejecting any entry that collides with `SYSTEM_PATHS` — loudly, naming the entry — rather than honouring it.
- **Unreadable or malformed sidecar.** Should fail loud, not fall back to an empty list: a silently ignored sidecar means the user believes their files are protected when they are not. That is the same quiet-in-the-bad-direction failure the proposal is meant to remove.
- **Path escapes the repo.** `../../etc/whatever` should be rejected. The list is only ever used as git pathspecs, but there is no reason to accept it.

## Alternatives considered

- **`.gitignore` the local files instead.** Does not work: `validate-system-paths-coverage.mjs` only inspects tracked files, but the whole point of several of these (`run-nightly.ps1`, `qa-fixtures/`) is that they are version-controlled locally.
- **Keep local files outside the repo.** Works for scripts, not for anything that must sit at a path the system reads.
- **Do nothing and re-resolve the conflict every sync.** Workable with `git rerere`, but it depends on the user noticing the conflict at all, and it does not help the plain user who just tripped the coverage validator.

Happy to send a PR if the sidecar shape looks right, or a different shape if you would rather it be YAML in `config/profile.yml` — the file format matters much less than getting the declaration out of `update-system.mjs`.

sdkkds:
Closes #2421.

Adds a gitignored declaration file that lets a checkout name the files it owns, without editing the file `apply` overwrites.

## The constraint

> a user's local-only file must survive an update without them having to edit a system file to protect it

Here is the whole user-side path, end to end:

```sh
# 1. the file this checkout owns, that upstream does not ship
$ cat run-nightly.ps1
...

# 2. declare it — this file is gitignored, so no update ever touches it
$ echo 'run-nightly.ps1' >> config/local-paths.txt

# 3. there is no step 3
```

No system file is edited at any point. `config/local-paths.txt` is gitignored, `config/local-paths.example.txt` is shipped for discovery, and both consumers read the declaration at runtime.

## Shape

`config/local-paths.txt` — one repo-relative path per line, `#` comments, trailing `/` for a directory prefix, same semantics as the existing arrays. Absent file means no extra paths, byte-identical to today's behaviour for every install that never creates one. That no-op case is test #1 and the one I'd re-run first on any change here.

Plain text rather than YAML, for the reason in the issue thread: `update-system.mjs` can't take a static top-level relative import (self-reexec invariant, #1706), and the file whose job is to survive an old→new jump is the last place to add a parser dependency. `split('\n')` + trim + drop-comments has no failure mode worth one.

Gitignored-only, as sketched — the declaration file is not added to `USER_PATHS`. It never becomes a tracked file, so the coverage validator never has to classify it, and `git status` never reports it, so the safety check never sees it either. Adding it would have been decoration on both paths.

## What changed

**`update-system.mjs`**
- `parseLocalPaths(text)` — pure, CRLF-tolerant, dedupes. CRLF matters: the reported population is Windows forks, and Notepad writes `\r\n`, which would make every entry miss its match.
- `localUserPaths(root)` — reads + validates, returns `[]` when absent.
- `effectiveUserPaths(root)` — `USER_PATHS` ∪ declared. This is what the safety check now compares against.
- `userLayerViolations(changed, updatePaths, userPaths)` — the SAFETY VIOLATION loop from `apply()`, extracted pure. `apply()` is ROOT-bound and full of side effects, so the rule was previously unpinnable; it's now covered directly, including the existing precedence rule (an explicit `updatePaths` entry still wins over a user-layer prefix match, e.g. `writing-samples/README.md`).

**`validate-system-paths-coverage.mjs`** — calls `localUserPaths(ROOT)` and unions the result into `ALL_PATHS`. As noted in the issue, this file reads both arrays via `extractArrayFromSource` on the source text, so it cannot see a runtime union — without the explicit call the coverage gap just moves. A malformed declaration exits 1 with the message rather than proceeding on a partial view: a guard that exists to be trustworthy must not silently narrow what it checks.

**`upgrade-tests.mjs` + `.github/workflows/test.yml`** — a new `--local-paths` leg, described below.

**Docs / discoverability** — `config/local-paths.example.txt` shipped alongside the other `*.example.*` files (added to `SYSTEM_PATHS`, so updates deliver it), a `### Fork-local paths` section in `DATA_CONTRACT.md`, and the `.gitignore` entry.

## Refusals

Three declarations throw, naming the offending entry:

| Refused | Why |
|---------|-----|
| Absolute path, or one containing `..` | Only ever widens "never touch" over files the updater doesn't own |
| A path the system layer already ships | The file silently stops receiving updates with no other signal — the quiet-in-the-bad-direction failure this issue is about |
| `config/local-paths.txt` itself | Gitignored, so nothing updates it; listing it guards a threat that doesn't exist |

Collision detection covers directory entries too — `modes/pdf.md` is refused because `modes/` ships it, not just exact matches.

## Tests

16 cases in `tests/updater-local-paths.test.mjs`, picked up by `test-all.mjs` discovery: parsing (comments, blanks, CRLF, dedup), every refusal above, and two end-to-end halves that have to move together — the coverage guard (an undeclared `run-nightly.ps1` fails it; declaring it clears the gap) and the safety-check rule (a touched fork-local file is a violation; a directory declaration covers files nested under it).

**Plus a real-apply leg**, since a declaration that only works in unit tests isn't worth much: `node upgrade-tests.mjs --local-paths`, wired into `test.yml` next to the canary. It builds the future where upstream *adopts a filename a fork already uses* — the mirror ships `run-nightly.ps1` and lists it in `SYSTEM_PATHS`, the install declares that same path — then runs a real old→new `apply` and asserts the declaration is honoured, the fork's bytes are intact, and the abort names both the declaration file and the entry.

Verified in both directions:

| Tree | Result |
|---|---|
| this branch | GREEN — 4/4 |
| `upstream/main` (leg copied over, feature absent) | RED — 2 failures |

The two that flip are the two that depend on the feature. I ran the unfixed case deliberately: a leg that has never been seen red proves nothing, which is the same reason `--canary` exists.

**What that leg does *not* claim.** I expected it to demonstrate an overwrite, and it doesn't — because `apply` already handles this. Without any of my changes, the fork's file survives anyway:

```
run-nightly.ps1  (local copy saved: run-nightly.ps1.bak)
Re-run with `node update-system.mjs apply --force` to take the upstream version instead.
```

So the difference this feature makes in that scenario is honesty, not rescue: an unevaluable declaration stops the update, instead of letting it proceed while the user believes a protection is in force that was never read. I'd rather say that plainly than let the leg imply it prevents data loss it doesn't.

**One design question that falls out of it.** Because the refusal happens inside the safety validation, an entry that *becomes* invalid later — upstream ships a path a fork declared years ago — aborts the whole update until the user edits their own `config/local-paths.txt`. That's loud, actionable, and needs no system-file edit, so it satisfies the constraint. But it does mean a stale declaration blocks upgrades. The alternative is to warn and ignore that entry, on the grounds that a system-layer path is managed anyway. I implemented the strict reading ("collisions fail loud, they don't win") because that's what I proposed and you approved — happy to switch it to warn-and-continue if you'd rather a stale line degrade instead of block.

Local runs: `test-all.mjs --quick` → 3679 passed, 0 failed. `upgrade-tests.mjs --pr-gate` → GREEN from `career-ops-v1.26.0`. `--local-paths` → GREEN. `validate-system-paths-coverage.mjs` → OK, 1028 tracked files.

Scope otherwise held to what I said in the thread: `update-system.mjs`, `validate-system-paths-coverage.mjs`, `.gitignore`, docs, tests. No change to `SYSTEM_PATHS` semantics, the `remoteSystemPaths` extraction, or rollback. The harness and workflow additions are the one deliberate widening — they exist because the constraint deserved a test that runs on every PR, not a paragraph in this description.

## Summary by CodeRabbit

* **New Features**
  * Added support for declaring fork-local paths preserved during system updates.
  * Added validation for unsafe, invalid, or conflicting path declarations.
  * Added an example configuration file with usage guidance.

* **Bug Fixes**
  * Updates now fail safely when managed files conflict with declared local paths, preserving local files and reporting the conflict.

* **Documentation**
  * Documented configuration format, validation rules, and behavior when the declaration file is absent.

* **Tests**
  * Added regression coverage for parsing, validation, path safety, collisions, and update behavior.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -83,3 +83,5 @@ jobs:
         run: node upgrade-tests.mjs --pr-gate
       - name: Canary (harness must be able to fail)
         run: node upgrade-tests.mjs --canary
+      - name: Local-paths declaration survives a real apply (#2421)
+        run: node upgrade-tests.mjs --local-paths
--- a/.gitignore
+++ b/.gitignore
@@ -58,6 +58,10 @@ writing-samples/*
 config/profile.yml
 config/cv-facts.json
 config/benchmarks.yml
+# Fork-local path declarations — this checkout's own files, read by
+# update-system.mjs. Gitignored on purpose: it describes THIS clone, not the
+# project, and must survive the sync that overwrites update-system.mjs (#2421).
+config/local-paths.txt
 portals.yml
 modes/_profile.md
 modes/_custom.md
--- a/DATA_CONTRACT.md
+++ b/DATA_CONTRACT.md
@@ -58,7 +58,7 @@ The two lists above describe *this project*. A fork usually carries files the pr
 
 `config/local-paths.txt` moves the declaration outside that blast radius. It is gitignored, read at runtime, and merged into the user layer for both the updater's safety check and `validate-system-paths-coverage.mjs`:
 
-```
+```text
 # one repo-relative path per line; blank lines and # comments ignored
 run-nightly.ps1
 .mcp.json
--- a/DATA_CONTRACT.md
+++ b/DATA_CONTRACT.md
@@ -12,6 +12,7 @@ These files contain your personal data, customizations, and work product. Update
 | `config/profile.yml` | Your identity, targets, comp range |
 | `config/cv-facts.json` | Your CV fact-check allowlist and forbidden phrases |
 | `config/benchmarks.yml` | Your market calibration benchmark overrides (optional; copy `templates/benchmarks.yml` here and edit — read by `funnel-velocity.mjs`) |
+| `config/local-paths.txt` | Files *this clone* owns that upstream does not ship — one repo-relative path per line (optional; copy `config/local-paths.example.txt` here and edit). See [Fork-local paths](#fork-local-paths) below |
 | `modes/_profile.md` | Your archetypes, narrative, negotiation scripts |
 | `modes/_custom.md` | Your house rules, custom workflows & output preferences (procedural — survives updates) |
 | `modes/_brief.md` | Your compact profile brief (~1.5–2K tokens) read by the two-pass triage first pass |
@@ -50,6 +51,29 @@ These files contain your personal data, customizations, and work product. Update
 | `reports/*` | Your evaluation reports |
 | `output/*` | Your generated PDFs |
 | `jds/*` | Your saved job descriptions |
+
+### Fork-local paths
+
+The two lists above describe *this project*. A fork usually carries files the project has never heard of — a nightly runner, an `.mcp.json`, a private fixtures directory. Those files are in the user layer by every definition that matters, but they cannot be added to `USER_PATHS`: that array lives in `update-system.mjs`, which `apply` overwrites and which git re-merges on every sync. The declaration would be erased by the process it exists to constrain.
+
+`config/local-paths.txt` moves the declaration outside that blast radius. It is gitignored, read at runtime, and merged into the user layer for both the updater's safety check and `validate-system-paths-coverage.mjs`:
+
+```
+# one repo-relative path per line; blank lines and # comments ignored
+run-nightly.ps1
+.mcp.json
+qa-fixtures/          # trailing slash = everything under this directory
+```
+
+Absent file means no extra paths — identical to the behaviour of every install that never creates one.
+
+Three declarations are refused, loudly, naming the entry:
+
+| Refused | Why |
+|---------|-----|
+| An absolute path, or one containing `..` | Would extend "never touch" over files outside the checkout |
+| A path the system layer already ships | The file would silently stop receiving updates, with no other signal that it had been frozen |
+| `config/local-paths.txt` itself | It is gitignored, so nothing updates it; listing it protects against a threat that does not exist and reads as though it did |
 
 ## System Layer (safe to auto-update)
 
--- a/config/local-paths.example.txt
+++ b/config/local-paths.example.txt
@@ -1,0 +1,29 @@
+# local-paths.txt — files this checkout owns, that upstream does not ship.
+#
+# Copy this file to `config/local-paths.txt` (gitignored) and list anything you
+# keep in your own fork: a nightly runner, an .mcp.json, a .gitattributes, a
+# fixtures directory. Every path listed here is treated exactly like cv.md —
+# `update-system.mjs apply` will never write to it, and the coverage guard
+# (`validate-system-paths-coverage.mjs`) stops reporting it as an orphan.
+#
+# Without this file, saying "that one is mine" means editing USER_PATHS inside
+# update-system.mjs — the file `apply` overwrites and git re-merges on every
+# sync, so the statement is guaranteed to conflict (#2421).
+#
+# Format
+#   - one repo-relative path per line
+#   - `#` starts a comment; blank lines are ignored
+#   - a trailing `/` means "this directory and everything under it"
+#
+# Refused, loudly, rather than honoured:
+#   - a path the system layer already ships (it would stop updating with no
+#     other signal — if you really mean to fork a system file, that is a
+#     different problem and worth an issue)
+#   - absolute paths, or anything containing `..`
+#
+# Examples — delete these and write your own:
+
+# run-nightly.ps1
+# .mcp.json
+# .gitattributes
+# qa-fixtures/
--- a/tests/updater-local-paths.test.mjs
+++ b/tests/updater-local-paths.test.mjs
@@ -262,6 +262,23 @@ console.log('\n🧪 Local user-paths declaration file (#2421)\n');
   } else {
     fail(`#14 precedence rule broken, got ${JSON.stringify(override)}`);
   }
+
+  // ── 15. A file declaration is not a prefix ──
+  //    Without a trailing `/` the entry names one file. `startsWith` let it
+  //    claim every neighbour sharing those bytes, so a declared
+  //    run-nightly.ps1 also swallowed run-nightly.ps1.old and
+  //    run-nightly.ps1.bak — files upstream may legitimately write, reported
+  //    as violations the user never asked for.
+  const neighbours = userLayerViolations(
+    ['run-nightly.ps1.old', 'run-nightly.ps1-notes.md', 'qa-fixtures-old/stale.md'],
+    [],
+    declared,
+  );
+  if (eq(neighbours, [])) {
+    pass('a file declaration does not claim prefix-sharing neighbours');
+  } else {
+    fail(`#15 expected no violations, got ${JSON.stringify(neighbours)}`);
+  }
 }
 
 for (const dir of roots) rmSync(dir, { recursive: true, force: true });
--- a/tests/updater-local-paths.test.mjs
+++ b/tests/updater-local-paths.test.mjs
@@ -20,8 +20,10 @@ import { join } from 'path';
 import { ROOT, pass, fail } from './helpers.mjs';
 import {
   LOCAL_PATHS_FILE,
+  USER_PATHS,
   parseLocalPaths,
   localUserPaths,
+  effectiveUserPaths,
   userLayerViolations,
 } from '../update-system.mjs';
 
@@ -286,6 +288,41 @@ console.log('\n🧪 Local user-paths declaration file (#2421)\n');
     pass('a file declaration does not claim prefix-sharing neighbours');
   } else {
     fail(`#15 expected no violations, got ${JSON.stringify(neighbours)}`);
+  }
+}
+
+// ── 16-18. The path union apply() actually consumes ──
+//    Every case above drives localUserPaths() or userLayerViolations() with a
+//    hand-built array. apply() uses neither directly — it calls
+//    effectiveUserPaths(), and that union is the value a regression would
+//    corrupt. Byte-identity of the union was only proven in the real-apply CI
+//    leg, so a change dropping or reordering USER_PATHS would leave every
+//    unit test green. Pin the union itself.
+{
+  const untouched = effectiveUserPaths(root(undefined));
+  if (eq(untouched, USER_PATHS)) {
+    pass('no declaration file → the union is USER_PATHS, byte for byte');
+  } else {
+    fail(`#16 union drifted from USER_PATHS: ${JSON.stringify(untouched)}`);
+  }
+
+  const declared = ['run-nightly.ps1', 'qa-fixtures/'];
+  const widened = effectiveUserPaths(root(`# mine\n${declared.join('\n')}\n`));
+  if (eq(widened, [...USER_PATHS, ...declared])) {
+    pass('a declaration widens the union without disturbing USER_PATHS');
+  } else {
+    fail(`#17 expected USER_PATHS + ${JSON.stringify(declared)}, got ${JSON.stringify(widened)}`);
+  }
+
+  // Widening is additive in the direction that matters: the safety check must
+  // still flag a built-in user-layer file, not just the declared ones. A union
+  // that replaced USER_PATHS instead of extending it would pass #17's length
+  // check on a reordering but silently stop protecting cv.md here.
+  const builtin = userLayerViolations(['cv.md'], [], widened);
+  if (eq(builtin, ['cv.md'])) {
+    pass('a built-in user-layer file is still protected under a widened union');
+  } else {
+    fail(`#18 expected cv.md flagged, got ${JSON.stringify(builtin)}`);
   }
 }
 
--- a/tests/updater-local-paths.test.mjs
+++ b/tests/updater-local-paths.test.mjs
@@ -1,0 +1,267 @@
+/**
+ * updater-local-paths.test.mjs — coverage for the local user-paths
+ * declaration file (#2421).
+ *
+ * A fork that keeps its own files (a nightly runner, an .mcp.json, a
+ * .gitattributes) has no supported way to say "this file is mine": the only
+ * lever is USER_PATHS, which lives inside update-system.mjs — the very file
+ * `apply` overwrites and git re-merges on every sync. The declaration file
+ * moves that statement OUT of the system layer.
+ *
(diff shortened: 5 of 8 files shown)


sdkkds:
The one red check is `test (windows-latest)` → `agent-inbox-tests.mjs` concurrent-add:

```
❌ every concurrent add exited cleanly — 2 non-zero exits
❌ all 30 concurrently queued items survive — kept=28 of 30
```

That's #2777 (`appendFileSync` isn't atomic on Windows), not anything in this PR — ubuntu and macos are green, the upgrade regression gate including the new `--local-paths` leg is green, and nothing here touches the inbox. The `error: pathspec 'generate-cover-letter.mjs' did not match any file(s) known to git` lines further up that log come from a different test's throwaway repo and appear on green runs too.

Happy to rebase if kicking the job is easier than taking my word for it.

santifer:
Sorry for the slow first reply, @sdkkds. The argument in your `DATA_CONTRACT.md` addition is the best part of this PR and it's a genuine catch:

> Those files are in the user layer by every definition that matters, but they cannot be added to `USER_PATHS`: that array lives in `update-system.mjs`, which `apply` overwrites and which git re-merges on every sync. **The declaration would be erased by the process it exists to constrain.**

That's circular in a way that can't be fixed by being more careful — any declaration living inside the system layer is unstable by construction. A gitignored file outside it is the only shape that closes the loop, and I hadn't seen the problem stated that cleanly before.

The CI step is the right instinct too:

```yaml
- name: Local-paths declaration survives a real apply (#2421)
  run: node upgrade-tests.mjs --local-paths
```

Asserting it survives an **actual apply** rather than unit-testing the parser is what makes the guarantee real. This feature's entire claim is "your file is still here afterwards", so the test has to run the afterwards.

**Where it stands, and it's not about quality:** it adds a *section* to `DATA_CONTRACT.md` defining a new concept, not just a row registering a new file. Registering a file is routine; introducing a third category alongside the system and user layers is a change to the contract itself, and that's Santiago's call rather than my checklist. It also touches `.github/workflows/test.yml`.

Queued for him with the circularity argument, which is the thing that makes the case. I'll come back either way rather than leaving it.

One thing worth having ready for that conversation: **what happens when a declared local path collides with a path upstream later starts shipping?** If `apply` then has to choose between the user's declaration and a real system file, the answer needs to be decided rather than discovered — and it should probably be loud, since silently preferring either one is a way to lose work.

sdkkds:
Thanks for the read, and no problem on the timing.

On escalating the contract section rather than deciding it yourself: that's the right split, and I'd rather it go to Santiago as a contract question than get waved through as a file registration. Happy to wait.

## The collision question

Short answer: **it should refuse the update and change nothing, naming the path.** Not "local wins", not "upstream wins".

Longer, because the interesting part is that the code already half-answers it.

### The static case is already decided

`localUserPaths()` refuses a declaration that collides with `SYSTEM_PATHS` at declaration-read time, with this reasoning:

> Declaring it would stop updates to it with no other signal

That rule is exactly as true a year later. If upstream starts shipping `check-fork-invariants.mjs` in v1.31 and my declaration silently keeps winning, I have pinned a system file forever and nothing ever tells me. Local-wins is not a safe default; it's the static failure mode with a delay fuse.

Upstream-wins is worse in the other direction: it overwrites a file I explicitly declared mine, which is the precise thing #2421 exists to prevent. Whichever side we pick silently, someone loses work they had a reasonable expectation of keeping.

So: same refusal, same message shape, one extra fact — this path was fine last version and is not any more.

### The bug is ordering, not detection

The temporal case is already *detectable* on the current branch — it just runs too late to matter.

`apply()` re-execs the **target** updater (`resolveReexecCheckout('FETCH_HEAD', 'update-system.mjs')` → `git checkout` → `execFileSync`), so by the second pass the in-process `SYSTEM_PATHS` is already upstream's new array. A newly-adopted path *would* trip the existing collision check. But the only call is `effectiveUserPaths()` in the step-4 safety validation, and the checkout loop over `updatePaths` has already run by then. The declared file is overwritten, and then we notice.

CodeRabbit flagged the same thing from the other end and scored it a data-loss risk. It's one fix, not two.

### What I'd implement

**Preflight before the checkout loop.** `updatePaths` — `mergePathLists(SYSTEM_PATHS, remoteSystemPaths, BOOTSTRAP_PATHS)` — is computed at step 3 and is exactly "every path this update is allowed to write". A declared local path appearing in it *is* the collision, by definition. No new source of truth, and it covers both cases in one check: the running array catches today's collisions, `remoteSystemPaths` (extracted from `FETCH_HEAD:update-system.mjs`) catches the ones upstream just adopted. It also degrades correctly — if the remote extraction throws on an old target, the re-exec'd `SYSTEM_PATHS` is still upstream's.

**Abort is a genuine no-op.** Running it before the loop means nothing has been checked out yet; the backup branch and WIP stash ref already exist by that point. Exit non-zero, name the path, name the version that adopted it, leave the tree alone. No partial update, nothing to roll back.

**The error has to be actionable**, or the honest refusal just strands someone on an old version:

```
config/local-paths.txt: "check-fork-invariants.mjs" is declared local, but
v1.31.0 ships it as a system file. Nothing was updated.

Either:
  - rename your copy (e.g. check-fork-invariants.local.mjs) and update the
    declaration, then re-run; or
  - remove the line to take upstream's version. Your current file is saved
    to check-fork-invariants.mjs.bak first.
```

The `.bak` is not new machinery — `locallyModifiedSystemFiles()` already writes one for every locally-modified system file before the checkout, so the second branch is the existing #2337 path with the declaration removed.

I'd deliberately **not** add an `--adopt-upstream` flag for this. It's one line to delete in a file the user wrote by hand; a flag would be a second way to say the same thing, and this is rare enough that the manual move is the clearer contract.

**Test:** an `upgrade-tests.mjs` leg where the fixture's target ref adds the declared path to `SYSTEM_PATHS`, with uncommitted local content in that file. Asserts non-zero exit, the path named in stderr, and the local content byte-identical afterwards. Same instinct as the existing `--local-paths` leg — the claim is about what survives, so the test has to run the apply.

### Two offers

1. I can push this now and let it sit on the PR, so the mechanism is settled if Santiago green-lights the contract section. It also clears CodeRabbit's data-loss finding either way.
2. If it helps the escalation, I'll **split the PR**: mechanism + tests + `.gitignore` + example file in one, the `DATA_CONTRACT.md` section in a second that depends on it. Then the contract change is the only thing needing Santiago's call, and it's reviewable as a document instead of buried in a diff with a workflow file. Say the word and I'll do the split — it's no work and it's your queue, not mine.

sdkkds:
Pushed the three CodeRabbit findings that don't depend on the collision ruling, one commit each so they're easy to split or drop:

- `ac32245` — `userLayerViolations` matched every entry with `startsWith`, so a file declaration claimed its prefix-sharing neighbours (`cv.md` → `cv.md.bak`). Trailing `/` now means prefix, anything else is exact. Every built-in `USER_PATHS` directory already carries the slash, so this makes the matcher agree with the documented syntax rather than changing it. New unit test #15.
- `205d906` — inherited git config could redirect the fixtures. `GIT_CONFIG_GLOBAL` closes one channel of three; a system gitconfig survives it, and an inherited `GIT_CONFIG_COUNT`/`KEY_n`/`VALUE_n` triple outranks every config file. Added `hermeticEnv()` and used it at **both** apply sites — the pre-existing leg had the same hole, and the header comment claims hermeticity for the whole harness — plus the throwaway-repo git calls and the guard subprocess in the unit test. Shout if you'd rather I left the older leg alone.
- `80a49e2` — fence tagged `text`.

Local: `test-all.mjs` 3681 passed / 0 failed, and `upgrade-tests.mjs --local-paths` GREEN. The one warning in the run is `cv-sync-check.mjs exited with error (expected without user data)`, which is the fixture-less checkout, not these changes.

**Deliberately not pushed:** the pre-checkout collision gate, and CodeRabbit's `DATA_CONTRACT.md:15` note about reconciling the no-read rule with the updater reading `config/local-paths.txt`. Both encode the answer to the question above, so they wait on the ruling — the second one is a contract edit, which is the part that's queued anyway.

One observation from re-running the poison leg, since it sharpens the question rather than answering it: today the temporal collision is *already* refused, but by accident of ordering. The re-exec runs the target updater, so the in-process `SYSTEM_PATHS` is upstream's new array, and `localUserPaths()` throws when it's finally called in the step-4 safety check — after the checkout loop has already run. The existing leg passes because `revertPaths()` puts the file back from HEAD. That's why the fixture commits the fork file before applying: with **uncommitted** local content there is nothing for the revert to restore. So the current behaviour is right by luck and only for committed content, which is the gap the gate closes.

Scott-Emberson (approved):
Code owner for `tests/`. `updater-local-paths.test.mjs` is hermetic (throwaway roots, no `career-ops-v*` tags, cleaned up), the coverage guard is non-vacuous, and the exact-vs-prefix matcher cases are solid. It follows the established convention that unit files test the extracted exports while real `apply()` behavior is proven in the harness. One low note: the real-apply byte-identity check lives only in the non-owned CI leg, so a regression in the apply-side path union would not fail an owned test. A single owned case driving `effectiveUserPaths(root)` against a throwaway root would close it. Non-blocking. Approving.

sdkkds:
Thanks for the review, @Scott-Emberson — and for calling out the gap rather than waving it through.

Closed the non-blocking note in `3dcff89`. You were right that nothing owned drove the union: every case in the file built its `userPaths` array by hand, so a regression in `effectiveUserPaths()` itself had no owned test to fail. Three cases now, all against throwaway roots:

- **#16** — absent declaration file → the union is `USER_PATHS`, byte for byte. This is the one that fails if the feature ever stops being invisible to installs that never opted in.
- **#17** — a declaration widens it to `[...USER_PATHS, ...declared]`, in that order.
- **#18** — `cv.md` is still flagged as a violation under the widened union.

#18 exists because #17 alone is weaker than it looks: a union that *replaced* `USER_PATHS` rather than extending it could still satisfy a length-and-membership check on a reordering. Driving `userLayerViolations()` with the real union proves the built-in half is still doing work, not just present in the array.

Also rebased onto `22cbe88` — the branch was 61 behind and the PR was showing `BEHIND`. No conflicts, despite `.gitignore`, `DATA_CONTRACT.md` and `update-system.mjs` being touched on both sides. Post-rebase: `test-all.mjs` 3955 passed / 0 failed (the jump from 3681 is upstream's tests arriving with the rebase, not mine), and `node upgrade-tests.mjs --local-paths` green on both legs — byte-identity after a real apply, and the deliberate `SYSTEM_PATHS` collision refusing with the offending entry named.

The other half of the PR is still parked: the `DATA_CONTRACT.md` section defines a third category alongside the system and user layers, which @santifer escalated to Santiago as a contract question rather than deciding as a file registration. That's the right split and I'm happy to wait. The collision ruling I proposed in the thread above — refuse the update, change nothing, name the path — is still awaiting that input, and the pre-checkout gate plus the `DATA_CONTRACT.md:15` no-read reconciliation are held behind it. The offer to split this into mechanism + contract still stands if it would help, so only the contract half needs Santiago.

santifer:
The design insight here is the whole feature: the declaration had to live outside the file the updater overwrites, or it would be erased by the process it constrains. And the three refused declarations (absolute/.., system-shipped paths, the file itself) close exactly the holes that would have made this dangerous. Thanks for taking Scott's gap seriously and closing it before the merge. Merged! 🚀

PR #2794, opened 2026-08-13
docs(update): render an empty changelog as a note, not a bare blockquote

### LINKED ISSUE #2724: modes/update.md: an empty changelog renders as a bare blockquote on the git-fallback path
**Time:** ~15 min · **Difficulty:** Beginner · **No-code** (one Markdown file)

`modes/update.md` step 2 renders the changelog into a blockquote:

```
> **Changelog:**
> {changelog from update-system.mjs check output}
```

There is no case for that value being empty, so when it is, the user gets a bare `> **Changelog:**` with nothing under it. It reads like something failed to load.

This is reachable, not theoretical. `update-system.mjs` has a fallback path (added in #2550) that checks for a new version over git when `curl` cannot reach GitHub, which is what happens on machines behind a proxy that only git knows about. That path knows the version but **not** the release notes, so it emits `changelog: ""`:

```json
{"status":"update-available","local":"1.26.0","remote":"9.9.9","changelog":""}
```

So the users who hit the empty blockquote are exactly the ones on a restricted network, who are already having the worst time.

## What to do

Add the empty case to `modes/update.md`, so the agent says something useful instead of rendering an empty quote. Something like: state that the version is available, that the release notes could not be fetched on this connection, and link the GitHub releases page.

The exact wording is yours. Two things to keep:

- The tone of the surrounding file (direct, second person, no hedging).
- No blame on the user's setup. It is a network shape, not a mistake.

There is no test for mode prose, so read it back and ask: if I were on a locked-down work laptop and saw this, would I know what happened and what to do next?

You can see the real payload with:

```bash
node update-system.mjs check
```

(On a normal connection it will carry a changelog. The empty case is the one you are writing for.)

## Why it matters

An empty section reads as a broken tool. The information is not missing because anything failed: it is missing because we reached upstream a different way. Saying so takes one sentence and turns a glitch into an explanation.

Ask anything in the comments. If you want it, say so here and I will assign it to you.

MannXo:
## What does this PR do?

Step 2 rendered `> **Changelog:**` unconditionally, so an empty value left a quote marker with nothing under it and read as a failed load. This adds the empty case. The marker is dropped and replaced by one line saying the notes are not available, with a link to the releases page.

## Related issue

Fixes #2724

## Type of change

- [ ] Bug fix
- [ ] New feature

## Checklist

---

### One correction to the issue

The issue attributes the empty changelog to the git fallback added in #2550. #2550 is still open, so that path is not on `main` yet.

The state is reachable today by a different route. `check()` fetches the VERSION file and the releases API as two independent `curlGet` calls. `changelog` is only assigned inside `if (releaseRaw !== null)`, while `remote` is satisfied by either source, so a successful VERSION fetch paired with a failed releases call produces `update-available` with an empty changelog. The unauthenticated GitHub API is capped at 60 requests per hour per IP, which is the ordinary way to land there. A release published with an empty body gives the same payload.

Once #2550 merges it adds a third route to the same state, so this rendering fix stands either way.

### On the wording

The payload cannot distinguish "the notes did not come through" from "this release has no notes". Both are `changelog: ""`. So the replacement line does not assert a cause. It states that the notes are not available for that version and points at the releases page. That keeps a bare `> **Changelog:**` from ever rendering, per your note on the issue, and still leaves the reader somewhere to go.

### Testing

`node test-all.mjs` on this branch gives 3663 passed, 0 failed, 2 warnings, exit 0. Same counts as `main`. Both warnings are environmental on my machine, `cv-sync-check.mjs exited with error (expected without user data)` and `Dashboard build skipped — go compiler not in env`.

`node upgrade-tests.mjs --pr-gate` reports GREEN.

No new tests. There is no test harness for mode prose, as the issue notes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Empty changelogs no longer display a blank **Changelog:** heading.
  * A fallback message now links to the GitHub release notes when no changelog content is available.


santifer:
Small and exactly right, @MannXo. A lone `> **Changelog:**` with nothing under it reads as a failure, and nothing failed — that's the whole bug, and the fix is to say the true thing instead: not available for this version, here's where the notes live.

The reason I like the wording you chose is that it covers both causes without pretending to distinguish them. The release genuinely having no notes and the check resolving the version without reading the releases API produce the same empty string, and from inside the mode there's no way to tell which — so a message that claims either one would be guessing. Yours points at the releases page and lets the user see for themselves.

Merged 🚀

PR #2795, opened 2026-08-13
refactor(providers): use the shared sleep helper instead of four copies

### LINKED ISSUE #2723: providers: three identical copies of the same `sleep` helper, and the shared one is not exported

MannXo:
## What does this PR do?

`_http.mjs` already defined the inter-page `sleep` helper but did not export it, so alibaba, avature and tencent each carried a hand-rolled copy. This exports the shared one and routes those three through it, plus a fourth copy in meituan that the issue does not name.

## Related issue

Fixes #2723

## Type of change

- [ ] Bug fix
- [ ] New feature

## Checklist

---

### The fourth copy

The verification grep in the issue is

```bash
grep -rn "const sleep = " providers/*.mjs
```

`_http.mjs` and `workable.mjs` declare theirs as `function sleep(ms, ctx)`, so that pattern never matched either of them. On `main` it matches four files rather than three, and after migrating only alibaba, avature and tencent it still prints one line:

```
providers/meituan.mjs:121:    const sleep = (ms) => (typeof ctx?.sleep === 'function' ? ctx.sleep(ms) : new Promise((r) => setTimeout(r, ms)));
```

meituan is migrated here too, so the grep now prints nothing. Happy to drop that file if you would rather hold the diff to the three you named.

For reference, `_http.mjs`'s helper is at line 109 on current `main`, not 87.

### Why every call site passes ctx

The local copies were arrow functions closing over `ctx` from the enclosing `fetch(entry, ctx)` scope. The shared helper takes `ctx` positionally, so importing it without passing `ctx` would silently bypass the `ctx.sleep` branch and make the tests wall-clock wait rather than fail.

The existing pacing assertions do catch that. Removing the argument from one call site in `tencent.mjs`:

```
❌ tencent.fetch() ctx.sleep calls: []
❌ tencent.fetch() keyword-switch pacing: []
📊 Results: 20 passed, 2 failed, 0 warnings
```

With the argument restored that file is back to 22 passed, 0 failed.

### meituan and the retry question

You asked on the issue to check each call site's retry behaviour before collapsing them. meituan is the one that case applies to. It picks the delay before calling sleep, so the same call serves both retry backoff and inter-page pacing:

```js
else await sleep(attempt > 0 ? RETRY_BACKOFF_MS * attempt : INTER_PAGE_DELAY_MS, ctx);
```

That choice is untouched, and the helper it now calls is identical to the one it defined, so both the backoff and the pacing delays are unchanged. alibaba, avature and tencent run no retry loop, as the issue says.

`workable.mjs` is untouched. Its header documents a local retry loop that deliberately diverges from the shared policy.

### Still duplicated, deliberately out of scope

Six providers carry the same helper as a top-level declaration rather than a closure, functionally identical to `_http.mjs`'s: ashby, eightfold, icims, oraclecloud, workday and workable. The grep in the issue does not surface them and they sit outside what it names, so they are untouched here. Say the word and I will open a follow-up.

### Testing

`node test-all.mjs` on this branch gives 3663 passed, 0 failed, 2 warnings, exit 0. Same counts as `main`. Both warnings are environmental on my machine, `cv-sync-check.mjs exited with error (expected without user data)` and `Dashboard build skipped — go compiler not in env`.

`node upgrade-tests.mjs --pr-gate` reports GREEN.

The four migrated providers' pacing assertions from that run:

```
✅ alibaba.fetch() paces follow-up requests via ctx.sleep (no delay before the first request)
✅ avature.fetch() throttles between pages (sleep called)
✅ meituan.fetch() paces follow-up requests via ctx.sleep (no delay before the first request)
✅ tencent.fetch() paces follow-up pages via ctx.sleep (no delay before the first request)
```

No new tests. Those assertions already cover the migration end to end, as the failure above shows.

### Overlap

#2748 is open against this issue and lands in a similar place. This one was assigned to me so I built it out. Take whichever is more useful to you.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved request pacing during paginated data retrieval for Meituan and Tencent integrations.
  - Requests now follow more consistent delays between pages, supporting smoother and more reliable data retrieval.


santifer:
@MannXo — this is red on a fresh base, and it's a real one rather than an infrastructure flake:

```
validate-portals failed: Identifier 'sleep' has already been declared
  ❌ validate-portals.mjs --file templates/portals.example.yml crashed (exit 1)
  ❌ validate-portals should accept a minimal valid portals file
  ❌ validate-portals should accept bundled provider-plugin ids
```

A module now imports the shared `sleep` while still declaring its own — so one of the four copies you set out to remove didn't get removed, and the file no longer parses. It's the failure mode this refactor is most exposed to: nothing type-checks it, and the crash only shows up when something actually imports the file.

Worth checking every provider in the diff rather than the one the error names, since the same edit was applied N times and whatever caused it here probably applies elsewhere.

The direction is right and I want it — four copies of a sleep helper is exactly the kind of drift that ends with three of them subtly different. Push the fix and I'll pick it straight back up.

santifer:
Heads-up for when you pick this back up, @MannXo: I merged #2682 just now, which adds retry/pagination handling to `providers/themuse.mjs` — and it brings its own local `sleep()`. So the four copies you're consolidating are now five.

Its version has one property worth preserving rather than flattening:

```js
function sleep(ms, ctx) {
  if (typeof ctx?.sleep === 'function') return ctx.sleep(ms);
  return new Promise((resolve) => setTimeout(resolve, ms));
}
```

The `ctx?.sleep` seam is what lets a test drive the retry path without waiting on real timers. Whatever the shared helper ends up looking like, it needs that injection point or themuse's tests get slow or get deleted.

Mentioning it here rather than letting you find it in a conflict.

MannXo:
The red is real, but it is a duplicated import rather than a leftover local declaration, and it came in with the merge commit rather than from my commit.

`node --check` on `providers/tencent.mjs` at the current head points at the second import, not at a local copy:

```
merged-tencent.mjs:17
import { sleep } from './_http.mjs';
         ^^^^^

SyntaxError: Identifier 'sleep' has already been declared
```

```
$ grep -n "import { sleep }" providers/tencent.mjs
4:import { sleep } from './_http.mjs';
17:import { sleep } from './_http.mjs';
```

Line 4 is from `main`, line 17 is mine. Same shape in `providers/meituan.mjs` at lines 4 and 23. `alibaba.mjs` and `avature.mjs` have one each and parse fine, because there my import happened to land on the same line as the one already on `main`.

Which points at the bigger thing. #2748 merged on the 13th and closed #2723, so `main` already exports `sleep` and already imports it in all four providers. #2748 put the tencent and meituan imports above the header comment, mine sat below it, and merging `main` into this branch kept both.

Diffed against current `main`, file by file: `alibaba.mjs` and `avature.mjs` are identical, `tencent.mjs` and `meituan.mjs` differ only in which line the import sits on, and `_http.mjs` on `main` is ahead of mine. Nothing here is left to rebase down to, so I will close this unless you want it kept open for something.

The part that is still open is what you raised about themuse. `main` today has eight top-level copies of the helper, all `function sleep(ms, ctx)` carrying the `ctx?.sleep` seam, functionally identical to the exported one:

```
providers/ashby.mjs:109      providers/oraclecloud.mjs:83
providers/eightfold.mjs:79   providers/themuse.mjs:32
providers/getro.mjs:82       providers/workable.mjs:71
providers/icims.mjs:31       providers/workday.mjs:67
```

`grep -rn "const sleep = " providers/*.mjs` now returns nothing, which is why this set stayed invisible to #2723's check. The seam you want preserved is already in the exported helper, so these are a straight migration. `workable.mjs` documents a deliberate local retry loop in its header, but its `sleep` is not the part that diverges.

Want me to open an issue for those eight and take it, or fold it into #2709?

MannXo:
Closing this. Against current `main` the whole PR is four added lines, and all four are the duplicate `sleep` import that breaks the build:

```
 providers/meituan.mjs | 2 ++
 providers/tencent.mjs | 2 ++
```

Removing them leaves an empty diff, since #2748 already landed everything this was for.

The eight top-level copies from my previous comment are still open whenever you want them. New issue or folded into #2709, either works, just say which.

PR #2809, opened 2026-08-13
feat(application-answers): parse a rendered section back into a snapshot

jchak7:
## What does this PR do?

`formatApplicationAnswersSection` has been write-only since it shipped — nothing in the tree can read a rendered `## Application Answers` section back, so `modes/apply.md` recovers previous answers by grepping `reports/` for a company name and re-reading the prose. This adds the inverse, `parseApplicationAnswersSection(reportText)`, returning the same shape `normalizeApplicationAnswersSnapshot` produces, so a parsed section can be handed straight back to the formatter or merged with a fresh snapshot.

## Related issue

None. See **Routing** at the bottom — happy to be redirected if you'd rather this went issue-first.

## Type of change

- [ ] Bug fix
- [x] New feature

### Why here and not a new module

`contacts.mjs` (`parseContacts` / `contactToVcard`) and `assessment-log.mjs` (`parseAssessments` / `buildRow`) both keep reader and writer in the module that owns the format. The parser also needs to invert seven module-private helpers (`inline`, `valueText`, `pick`, `quoteBlock`, `qaLines`, `compactLines`, `fileLines`) and share `APPLICATION_ANSWERS_HEADING` / `VALID_STATES`; splitting would mean exporting internals or duplicating them, and duplicated readers drift.

The `tracker-parse.mjs` / `tracker-utils.mjs` split is not a counter-example — that boundary is about side effects (locking, atomic writes), not parse-vs-format. Both functions here are pure.

### The property this guarantees — and the one it does not

**It does not claim byte-equality with the input snapshot, and asserting that would be wrong.** The formatter is lossy by design:

- `inline()` collapses whitespace in every label
- `valueText()` joins arrays with `', '`, so `['a, b','c']` and `['a','b','c']` render identically
- `pick()` discards which of the four accepted key spellings was used
- empty values render as the sentinels `Not recorded` / `> Not recorded.`
- an empty label renders as `Answer 1` / `Selection 1` / `Field 1` / `File 1`

What holds, and what the tests pin, is that **one render normalizes and every render after that is stable**:

```
parse(format(x)) === parse(format(parse(format(x))))
```

Parsed entries use the *primary* key the formatter picks first (`question`/`answer`, `question`/`selection`, `field`/`path`), which is what makes re-rendering a fixed point rather than falling through to fallback labels.

Multi-line answers are the one genuinely lossless payload and are asserted as such, including the blank line `quoteBlock` renders as `"> "`.

### Error handling

`parseApplicationAnswersSection` is **total — it never throws**, for any input including `null`, `undefined` and `''`.

- **No `## Application Answers` heading → `null`.** The section is absent.
- **Heading present, body unreadable → a snapshot with empty fields.** Missing `Date`/`State` become `''`; a group that can't be sliced becomes `[]`.

Totality is load-bearing for the property this PR asserts: a parser that threw on malformed input would make `parse(format(x))` partial, and the fixed-point assertion would hold only on well-formed input — the input that least needs a guarantee.

**Opt-in refusal** (`{ strict: true }`), added after @santifer's review. By default a line inside a group that matches no entry pattern is skipped, so a partially-mangled section returns a snapshot indistinguishable from a clean parse — for apply-mode recovery that is worse than a refusal, because the missing answer looks like an answer the user never gave. Worse, an unnumbered heading *mid-block* leaves the previous entry open, so its quote lines are absorbed into the **previous** answer rather than dropped — a corrupted answer, not a missing one. `strict: true` throws instead, naming every line it could not read.

The default path is unchanged, verified rather than asserted: `JSON.stringify(parse(input))` is byte-identical between the pre- and post-change module across eight input variants, and formatter output is unchanged. So the fixed point above still holds and no existing caller is affected.

### Boundaries

`parseApplicationAnswersSection` uses the same heading probe (`/^## Application Answers\s*$/m`) and terminator (`/^## .+$/m`) as `upsertApplicationAnswersSection`, so the reader and the writer agree on where the section starts and ends by construction. Returns `null` when there is no section.

### Risk

Additive only. No new file, no new data file, no new mode, no new CLI surface, no new dependency, no change to any existing export.

`modes/apply.md` is coupled to this exact rendering and is CI-blind, so the suite includes an explicit assertion that formatter output is unchanged — and the three existing inline assertions in `test-all.mjs` were re-run against the patched module and still pass.

### Tests

`tests/application-answers.test.mjs`, auto-discovered (no `test-all.mjs` registration, no `SYSTEM_PATHS` entry). Ten assertions: fixed point, key spelling, multi-line fidelity, all four sentinels including the inline spelling, section boundary against a report with a following `##` block, null on absence, opt-in strict refusal across both the compact groups and the free-text group, and formatter-unchanged.

Anti-vacuity guard included — a round-trip suite over an empty corpus passes trivially, so the corpus asserts it produced entries before any equality check.

Full suite on this branch: **3703 passed, 0 failed, 1 warning** (a Dashboard build skipped for no Go compiler in my environment — not introduced by this change). The three required `test` workflows are still awaiting maintainer approval for a first-time contributor, so treat this as a local run rather than CI.

### Routing

Reading CONTRIBUTING, this looked borderline on issue-first: it is a new exported function, but adds no new file, data file, mode, CLI surface or dependency, and no behaviour change to any existing caller — question 2 ("who pays the maintenance?") is ~90 lines in a module already maintained. I've sent it directly on that basis. Happy to split it behind an issue if you'd rather route it that way.

## Checklist

- [x] If this is a new feature or architecture change, I opened an issue first (bug fixes, providers, docs & translations are exempt — send those straight in) — *see Routing above; treated as API completion, offered to route*

## Summary by CodeRabbit

* **New Features**
  * Added support for extracting and normalizing Application Answers from report Markdown.
  * Recognizes dates, states, text responses, selections, field values, and file entries.
  * Handles missing sections, special values, multiline content, and different line-ending formats.
  * Added optional strict validation to identify unreadable entries.

* **Tests**
  * Added comprehensive coverage for parsing, formatting round trips, section boundaries, normalization, and error handling.


Scott-Emberson (commented):
Thanks for this, @jchak7, and welcome. I am reviewing as code owner for `tests/`, so this is about `tests/application-answers.test.mjs` only.

Nice test. Pinning the reader as a fixed point instead of chasing byte-equality is the right call, and the header at 8-33 says why: `inline()` collapses label whitespace, `valueText()` joins arrays with `, `, and `pick()` forgets which key spelling was used, so `parse(format(x)) === parse(format(parse(format(x))))` is the invariant that actually holds. The anti-vacuity guard at 98-99, which asserts the corpus produced entries before any equality check, is the part I was most glad to see, and the boundary agreement with `upsert` at 182-206 plus the byte-stable formatter check at 219-233 cover the two ways this could quietly break. I ran it on `e9e047f`: all seven assertions pass.

One thing, minor, in this file.

**The inline `Not recorded` sentinel is promised but never rendered.** The header at 21-23 says the suite pins the four sentinels and that "the two 'Not recorded' spellings must read back as empty." The corpus produces two of the three parser sentinel paths: block `- None captured.` for an empty group (`corpus[1]`) and block `> Not recorded.` for an empty free-text answer (`corpus[0].freeText[2]`). It never produces the inline spelling. `compactLines` (application-answers.mjs:74) renders `**Label:** Not recorded` for a selection or field value that exists with an empty value, `fileLines` (84) does the same for a file with an empty path, and the parser strips both back to `''` at `value === NOT_RECORDED_INLINE ? ''` and `file === NOT_RECORDED_INLINE ? ''`. Neither branch is hit by any fixture. Drop either strip in a later refactor and an empty selection round-trips to the literal answer `Not recorded`, a value nobody typed, with this suite still green; the fixed-point test cannot catch it because no corpus entry renders the inline form. Adding a selection, a field value, and a file entry with empty values to `corpus[0]` covers it (they survive normalization: `list()` at application-answers.mjs:32 does not filter, I checked), then assert each reads back as `''` rather than `Not recorded`. It is worth a direct assertion because this is the one sentinel path that fabricates content instead of dropping it, and the header already claims it.

santifer:
Sorry for the slow first reply, @jchak7. The framing is what makes this worth doing:

> `formatApplicationAnswersSection` has been **write-only** since it shipped — nothing in the tree can read a rendered `## Application Answers` section back

A formatter with no parser is a one-way door: everything downstream that needs the data has to re-derive it from prose, and every consumer re-derives it slightly differently. `modes/apply.md` recovering previous answers by asking the model to read the rendered section is exactly that — it works until the day the rendering changes and nothing notices, because a model reading markdown never *fails*, it just produces something plausible.

A real parser turns that into a contract with two sides, and the round-trip becomes testable: format → parse → compare. That's the property I'd want asserted rather than "the parser handles these five shapes" — a round-trip test stays honest when the format grows a field, and a shape list goes stale silently.

Not merged tonight (per-session merge ceiling), queued. One question worth answering in the PR body: what happens on a section the parser can't read — throw, or return partial? For recovery of the user's own previous answers, a partial parse that silently drops one is worse than a refusal, because the missing answer looks like an answer they never gave.

jchak7:
Thanks all three. The most useful thing since my last push: a reviewer found that my fix had the same defect the PR is about. Details under the coderabbitai heading.

### @santifer: silent drops

Current behaviour:

| Input | Return |
|---|---|
| No `## Application Answers` heading | `null` |
| Heading present, body unreadable | a snapshot with `''` / `[]` in whatever it couldn't read |

It never throws. On your specific worry it behaves as badly as you guessed. A section where one selection line lost its numbering and one file line lost its bold:

```
well-formed   selections: 2   files: 1
mangled       selections: 1   files: 0

returned: {"selections":[{"question":"A","selection":"1"}],"files":[]}
keys: date, state, freeText, selections, fieldValues, files
```

No throw, no null, no signal. `parseCompactEntries` and `parseFileEntries` both `continue` on a non-matching line, and `parseQaEntries` ignores any line that isn't a `>` quote. The caller gets a snapshot that looks complete and is missing an answer. "An answer they never gave", exactly.

I'd push back on throwing by default, though, not on the substance. Totality is doing real work here: if the parser threw on malformed input, `parse(format(x))` would be partial and the fixed-point assertion would only hold on well-formed input, which is the input that least needs a guarantee. The problem isn't that it returns something. It's that what it returns is indistinguishable from a clean parse.

So the fix keeps it total and makes the refusal opt-in:

```js
parseApplicationAnswersSection(text)                    // unchanged, byte for byte
parseApplicationAnswersSection(text, { strict: true })  // throws, naming what it refused
```

```
Error: Application Answers section has 2 unreadable entries:
       - **Notice period: current:** 30 days | 1. CV: output/acme-cv.pdf (v3)
```

I tried `{ dropped: n }` on the return value first and threw it away, because it contradicts the invariant claimed two paragraphs up: the parser returns the shape `normalizeApplicationAnswersSnapshot` produces, so a parsed section can go straight back to the formatter. A diagnostics field breaks that, and it lands inside the `JSON.stringify` the fixed-point test compares. The flag leaves the shape alone.

The intended caller is `modes/apply.md`, which opts into `strict`. Recovering the user's own previous answers is the case where a refusal beats a plausible-looking gap. Every other caller keeps today's behaviour with no try/catch.

Module cost is +27 / -11: an optional `onSkip` callback threaded into the four entry parsers, plus a throw at the end. It's in separate commits from the reader, so you can drop it and take the reader alone if you'd rather route the refusal differently, or want a different shape than a throw.

On "default unchanged": I diffed `JSON.stringify(parse(input))` between the old and new module across eight inputs (clean, heading-mangled, mid-block-mangled, both, compact-mangled, empty, no-section, quote-only) plus formatter output. Byte-identical on all of them. `onSkip` is `undefined` unless `strict` is set, so `onSkip?.(line)` can't do anything on the default path.

### @coderabbitai: my fix had the bug the PR is about

I threaded the skip callback into `parseCompactEntries` and `parseFileEntries` and not into `parseQaEntries`. So `strict: true` refused an unreadable selection, field value or file, and silently dropped an unreadable free-text answer, in a PR whose whole argument is that a silent partial is worse than a refusal. I'd even written, one paragraph up, that `parseQaEntries` ignores non-quote lines, and then didn't plug it. Fixed in `32cd7bb`.

Chasing it down turned up a worse case than the reported one. Dropping happens when the orphaned quote lines arrive with no entry open. When a heading in the middle of the block loses its numbering, `current` is still set from the previous entry, so those lines get appended to the previous answer:

```
1. **Why this role?**
> Line one.
- **Describe a failure**      <- numbering lost
> Line three after a blank.

default parse -> freeText.length 2, and freeText[0].answer now ends with
                 "Line three after a blank.", text written under a different question
```

That's not a dropped answer, it's a corrupted one, and no assertion in the suite could see it before this commit. Section 7c pins both paths: the default still drops and still absorbs (unchanged, on purpose), strict throws on each mangling separately, and the message names the offending heading.

### @Scott-Emberson: the inline sentinel

You were right, including the part I'd have got wrong on my own: the branches exist and are correct, and nothing executes them. Done as you specified, appended to `corpus[0]` rather than inserted, so the index assertions in section 3 keep pointing at what they already pin:

```js
selections:  { question: 'Work authorization', selection: '' },
fieldValues: { question: 'Earliest start date', answer: '' },
files:       { field: 'Portfolio', path: '' },
```

plus a `5b` block asserting the formatter actually renders `**Work authorization:** Not recorded`, that each reads back as `''`, that the entries survive rather than being dropped, and that an empty path doesn't pick up a phantom `version`. Section 6 needed `files.length` 2 -> 3 as a consequence.

Your prediction that the fixed-point test can't catch it: I ran it rather than take it on faith. Deleting either strip:

```
-- drop the inline branch in parseCompactEntries --
   ok    parse(format(x)) is a fixed point across 11 entries in 2 snapshots
   FAIL  inline sentinel leaked into data

-- drop the inline branch in parseFileEntries --
   ok    parse(format(x)) is a fixed point across 11 entries in 2 snapshots
   FAIL  inline sentinel leaked into data
```

Both mutants die on the new assertion and nothing else, and the fixed point stays green through both, exactly as you said. The general shape, now in the header comment: the property is blind to any sentinel that survives its own re-render. `'Not recorded'` -> `'Not recorded'` is stable and wrong.

Also confirming your check on `list()` at `application-answers.mjs:32`: empty values survive normalization, so the fixtures render instead of being filtered out before they reach the formatter.

I held each new assertion to the bar you set: reverting the fix has to kill that assertion and nothing else. Reverting the free-text skip fails 7c only; 7b and the fixed point stay green.

---

Ten assertions in `tests/application-answers.test.mjs` now. Full suite locally: 3703 passed, 0 failed, 1 warning (the pre-existing "Dashboard build skipped, no Go compiler" in my environment). The three required `test` workflows haven't run yet since they're waiting on maintainer approval for a first-time contributor, so those numbers are mine, not CI's.

PR body updated with an "Error handling" section covering the above, and the suite numbers corrected. Branch updated against `main`. Merge still needs code-owner approval from @Scott-Emberson.

Scott-Emberson (approved):
Re-reviewed at `72e97205`. Our one point from last pass is resolved cleanly. Section 5b now pins the inline `Not recorded` sentinel in both directions: it asserts the formatter actually emits `**Work authorization:** Not recorded`, then that each entry strips back to `''` on parse. Delete either strip in `parseCompactEntries`/`parseFileEntries` and 5b fails, so it is not a fixed-point blind spot any more.

The strict-mode additions are the right kind of test. 7c constructs a free-text section whose heading loses its numbering mid-block, asserts under the default parser that the orphan quote lines get absorbed into the preceding answer (the actual data-corruption class this PR exists to catch), then asserts `strict:true` throws and names the entry. Every strict assertion keys off `strictThrew !== null`, so a no-op strict mode would fail all three blocks rather than pass silently. Hermetic, matches the neighbour shape, round-trip fixed point intact. Approving.

santifer:
Welcome to career-ops, @jchak7! This is a first contribution to point at: you measured the silent-drop concern instead of arguing it, kept the parser total with the fixed-point reasoning spelled out, made refusal opt-in where recovery actually needs it, and publicly caught your own fix carrying the very bug the PR is about — then found and covered the worse case behind it. That last habit is the rarest one. Merged — your first contribution. Glad to have you here. 🚀 (Follow-up welcome whenever: wiring modes/apply.md to opt into strict, as you sketched.)

PR #2812, opened 2026-08-13
fix(consider): perform GET /jobs CSRF handshake before the POST — fixes 412 INVALID_CSRF on every board (#2764)

### LINKED ISSUE #2764: consider provider: every board returns 412 INVALID_CSRF — POST omits the CSRF handshake
## Summary

The `consider` provider added in #825 returns `412 Precondition Failed` / `{"error":"INVALID_CSRF"}` on **every** board, including the example board in its own docblock (Founderful / `wingman`). The POST to `/api-boards/search-jobs` is sent cold; Consider requires a session cookie plus an `x-csrf-token` header obtained from a prior `GET /jobs`.

Net effect: `providers/consider.mjs` cannot fetch any jobs as shipped in v1.26.0.

## Version

`1.26.0` (provider introduced in c0a259a9, #825)

## Reproduction

```js
// providers/consider.mjs docblock example, verbatim
const entry = {
  name: 'Founderful (portfolio)',
  provider: 'consider',
  consider_board: 'wingman',
  careers_url: 'https://jobs.founderful.com/jobs',
};
await consider.fetch(entry, { fetchJson });
```

```
Error: HTTP 412 Precondition Failed
  status: 412,
  body: '{"error":"INVALID_CSRF"}'
    at Object.fetch (providers/consider.mjs:96)
```

`detect()` is fine — it resolves `https://jobs.founderful.com/api-boards/search-jobs` correctly. The failure is in `fetch()`.

## Root cause

`fetch()` POSTs directly to `ENDPOINT_PATH` with only `content-type`, `accept`, and `referer`. Consider's edge rejects any POST that lacks:

1. the `session` / `session.sig` cookies set by a `GET` on the board origin, and
2. an `x-csrf-token` header whose value is embedded in the board page HTML.

## Verification that the handshake is the whole fix

Same two boards, cold POST vs. `GET`-then-POST reusing the cookie jar and CSRF token:

| Board | Cold POST (current provider) | GET + cookie + `x-csrf-token` |
|---|---|---|
| Founderful (`wingman`, docblock example) | `412 INVALID_CSRF` | `200` — `total=141` |
| Lightspeed (`lightspeed`) | `412 INVALID_CSRF` | `200` — `total=15214` |

No other change was needed: same endpoint, same request body, same board ids. The response shape matches what the provider already parses (`{ jobs: [...], total }`).

## Suggested fix

In `fetch()`, before the POST:

1. `GET {origin}/jobs`
2. collect `Set-Cookie` into a `cookie` request header
3. scrape the CSRF token from the returned HTML and send it as `x-csrf-token`

This needs cookie/response-header access that `fetchJson()` in `providers/_http.mjs` does not currently expose, so the shared helper likely needs a way to return (or thread through) response headers — worth deciding before I open a PR, since it touches a shared transport used by every provider.

Two notes for whoever picks this up:

- The CSRF token lives in the page HTML, so the scrape is regex-on-HTML and will need a clear failure mode when Consider changes their markup — better to throw a named error than to silently fall back to a cold POST that 412s.
- The existing SSRF guard in `resolveOrigin()` should cover the new `GET` as well, since it hits the same config-driven origin.

## Related

`providers/getro.mjs` and `providers/joinup.mjs` landed in the same PR; I have not tested whether either is affected by a similar handshake requirement.

Happy to open a PR if you'll confirm the preferred approach for exposing response headers through `fetchJson()`.

Zoubeir23:
Closes #2764.

## The defect

`consider.fetch()` posted cold to `/api-boards/search-jobs` with only `content-type`, `accept`, and `referer`. Consider's edge rejects any POST that lacks:

1. a `session` / `session.sig` cookie set by a prior anonymous `GET` on the board origin, and
2. an `x-csrf-token` header whose value is embedded in the board page HTML.

Result: every board returned `412 INVALID_CSRF`, including the docblock example (`Founderful / wingman`). All 9 VC-portfolio boards on Consider (a16z, Sequoia, Greylock, Kleiner Perkins, Contrary, Battery, NEA, Lightspeed, Bessemer) have been dark since v1.26.0.

## The fix

`acquireCsrfHandshake(origin)` — an anonymous `GET {origin}/jobs` — runs before every POST:

1. Extracts `session` / `session.sig` cookies via `res.headers.getSetCookie()` (discrete per-header strings, avoiding comma-folding ambiguity).
2. Extracts `csrfToken` from the board-page HTML (`"csrfToken":"<value>"` in a `<script>` payload).
3. Returns `{ cookie, csrfToken }` — either field is `null` if absent.

`fetch()` spreads non-null fields into the POST headers as `cookie` and `x-csrf-token`. A failed handshake (network error, non-200) returns both as `null` and the POST is still attempted — same observable behaviour as before, with a clean 412 instead of a silent skip.

The GET uses no credentials. The session token Consider issues is handed to any anonymous visitor; this is the same zero-auth contract we already accept in other providers.

**`redirect:'error'` is kept on the POST.** The GET uses `redirect:'follow'` (board landing pages occasionally redirect `/jobs` → `/jobs/`); the first-hop host is already validated by `resolveOrigin()`.

## Test plan

`ctx._acquireHandshake` is a test seam: set it to a stub in unit tests so no real network call is made. All existing tests updated to use `noHandshake = async () => ({ cookie: null, csrfToken: null })`.

New cases (6):
- [x] `fetch()` passes board origin (not full `careers_url`) to the handshake
- [x] `cookie` from handshake appears as `Cookie:` in the POST
- [x] `csrfToken` appears as `x-csrf-token:` in the POST
- [x] Both headers present when handshake returns both
- [x] Neither header present when handshake returns both null
- [x] POST is still attempted when handshake returns null (graceful degradation)

`node tests/providers/consider.test.mjs` — 16 passed, 0 failed.
`node test-all.mjs` — 3436 passed, 0 failed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved request reliability when communicating with Consider.
  * Added support for required session cookies and CSRF tokens during requests.
  * Requests now continue gracefully when the initial handshake fails or returns incomplete data.
  * Added timeout handling to prevent stalled connections.

## Cost: two requests per company, unconditionally

The handshake runs before every POST, so each Consider entry costs **two requests per scan instead of one**. That is a real multiplier across every company on this ATS and belongs stated rather than implied.

It is nonetheless the cheaper of the two designs, measured against the live board in `portals.example.yml`:

| | requests | result |
|---|---|---|
| bare POST, no handshake | 1 | `HTTP 412 {"error":"INVALID_CSRF"}` |
| handshake then POST (this PR) | 2 | 137 postings |
| POST first, handshake on 412 | 3 | 137 postings |

Every observed Consider board enforces CSRF — which is what #2764 reported and what makes this provider-wide rather than one bad entry. So making the handshake conditional would add a round-trip to every company rather than remove one.

If a non-enforcing board ever appears, the cheap answer is a per-entry skip flag, not inverting the order.

### DIFF
--- a/providers/consider.mjs
+++ b/providers/consider.mjs
@@ -83,8 +83,18 @@ async function acquireCsrfHandshake(origin) {
   const controller = new AbortController();
   const timer = setTimeout(() => controller.abort(), HANDSHAKE_TIMEOUT_MS);
   try {
+    // redirect:'error' blocks every redirect unconditionally. A redirect-to-
+    // private-IP (169.254.169.254, ::1, …) would otherwise bypass the host
+    // guard in resolveOrigin() and make a request to an internal target.
+    // redirect:'manual' cannot be used here: the WHATWG opaque-redirect
+    // response (Node ≥18 / undici) returns status 0 and empty headers, so
+    // the Location value is unreadable without implementation-specific APIs.
+    // redirect:'error' gives the same security outcome — zero redirects
+    // followed — and the catch below treats the resulting TypeError as a
+    // degraded handshake (null/null), which is correct.
     const res = await fetch(`${origin}/jobs`, {
       headers: { 'user-agent': BROWSER_LIKE_USER_AGENT, accept: 'text/html,*/*' },
+      redirect: 'error',
       signal: controller.signal,
     });
     if (!res.ok) return { cookie: null, csrfToken: null };
--- a/providers/consider.mjs
+++ b/providers/consider.mjs
@@ -1,6 +1,8 @@
 // @ts-check
 /** @typedef {import('./_types.js').Provider} Provider */
 
+import { BROWSER_LIKE_USER_AGENT } from './_http.mjs';
+
 // Consider provider — VC "talent network" portfolio boards on getconsider.com
 // (Founderful, Creandum, Balderton, Lightspeed, Notion Capital, …). The board
 // is a JS app, but its data comes from a same-origin JSON endpoint we can hit
--- a/providers/consider.mjs
+++ b/providers/consider.mjs
@@ -141,12 +141,29 @@ export default {
     if (!entry.consider_board) throw new Error(`consider: ${entry.name} needs a 'consider_board' id in portals.yml`);
     const size = Number.isInteger(entry.consider_size) && entry.consider_size > 0 ? entry.consider_size : DEFAULT_SIZE;
 
+    // Perform the CSRF handshake before the POST. ctx._acquireHandshake is a
+    // test seam: set it to a stub in unit tests so no real network call is made.
+    const { cookie, csrfToken } = await (
+      typeof ctx._acquireHandshake === 'function'
+        ? ctx._acquireHandshake(origin)
+        : acquireCsrfHandshake(origin)
+    );
+
+    const csrfHeaders = {};
+    if (cookie) csrfHeaders.cookie = cookie;
+    if (csrfToken) csrfHeaders['x-csrf-token'] = csrfToken;
+
     const json = await ctx.fetchJson(origin + ENDPOINT_PATH, {
       method: 'POST',
       // redirect:'error' so a 3xx from the (config-driven) board host can't be
       // followed to a private/metadata IP — the host guard above pins the first hop.
       redirect: 'error',
-      headers: { 'content-type': 'application/json', accept: 'application/json', referer: origin + '/jobs' },
+      headers: {
+        'content-type': 'application/json',
+        accept: 'application/json',
+        referer: origin + '/jobs',
+        ...csrfHeaders,
+      },
       body: JSON.stringify({
         meta: { size },
         board: { id: String(entry.consider_board), isParent: true },
--- a/providers/consider.mjs
+++ b/providers/consider.mjs
@@ -43,6 +43,9 @@ function toEpochMs(value) {
 
 const ENDPOINT_PATH = '/api-boards/search-jobs';
 const DEFAULT_SIZE = 500;
+// Budget for the anonymous GET that seeds the session cookie and csrfToken.
+// Shorter than the POST budget so a slow board page can't eat the full timeout.
+const HANDSHAKE_TIMEOUT_MS = 8_000;
 
 // SSRF guard. The POST target host is config-driven (built from the portals.yml
 // careers_url), so pin it to a public HTTPS origin before fetching. Consider
--- a/providers/consider.mjs
+++ b/providers/consider.mjs
@@ -73,6 +73,49 @@ function resolveOrigin(entry) {
   return parsed.origin;
 }
 
+// Perform the anonymous GET /jobs handshake that Consider requires before
+// accepting a POST. Returns { cookie, csrfToken } — either field is null if
+// the server did not supply it. On any network failure the catch returns both
+// null so the caller can still attempt the POST (it will 412, but that is a
+// cleaner signal than a silent skip — and it keeps the same observable
+// behaviour as the pre-fix code for boards that don't enforce CSRF).
+async function acquireCsrfHandshake(origin) {
+  const controller = new AbortController();
+  const timer = setTimeout(() => controller.abort(), HANDSHAKE_TIMEOUT_MS);
+  try {
+    const res = await fetch(`${origin}/jobs`, {
+      headers: { 'user-agent': BROWSER_LIKE_USER_AGENT, accept: 'text/html,*/*' },
+      signal: controller.signal,
+    });
+    if (!res.ok) return { cookie: null, csrfToken: null };
+    const html = await res.text();
+
+    // getSetCookie() returns each Set-Cookie header as its own string, avoiding
+    // the comma-folding ambiguity of get('set-cookie') for values that contain
+    // commas. Available since Node 18.14; project minimum is Node 22.
+    const setCookies = typeof res.headers.getSetCookie === 'function'
+      ? res.headers.getSetCookie()
+      : (res.headers.get('set-cookie') ?? '').split(/,(?=\s*\w+=)/).filter(Boolean);
+
+    const cookie = setCookies
+      .map(c => c.split(';')[0].trim())
+      .filter(Boolean)
+      .join('; ') || null;
+
+    // Consider embeds the CSRF token as `"csrfToken":"<value>"` inside a JSON
+    // payload in a <script> tag on the board landing page. The 8-char lower
+    // bound rules out placeholder strings and short error tokens.
+    const m = html.match(/"csrfToken"\s*:\s*"([^"]{8,})"/);
+    const csrfToken = m ? m[1] : null;
+
+    return { cookie, csrfToken };
+  } catch {
+    return { cookie: null, csrfToken: null };
+  } finally {
+    clearTimeout(timer);
+  }
+}
+
 function locationString(job) {
   if (Array.isArray(job.locations) && job.locations.length) {
     return job.locations.filter(l => typeof l === 'string').join(', ');
--- a/tests/providers/consider.test.mjs
+++ b/tests/providers/consider.test.mjs
@@ -231,6 +231,40 @@ try {
     } else {
       fail('acquireCsrfHandshake must not swallow a redirect error into a full abort');
     }
+
+    // !res.ok branch: a non-2xx response from GET /jobs (e.g. 403, 500) must
+    // degrade to null/null and still attempt the POST — the same outcome as the
+    // catch branch, but via a different code path (line 100 in consider.mjs).
+    let notOkPostHeaders = null;
+    let notOkPostCalled = false;
+    const realFetch3 = globalThis.fetch;
+    globalThis.fetch = async () => ({
+      ok: false,
+      status: 403,
+      headers: { getSetCookie: () => [], get: () => null },
+      text: async () => '',
+    });
+    try {
+      await consider.fetch(okEntry, {
+        fetchJson: async (_url, opts) => {
+          notOkPostCalled = true;
+          notOkPostHeaders = opts.headers;
+          return { jobs: [] };
+        },
+      });
+    } finally {
+      globalThis.fetch = realFetch3;
+    }
+    if (notOkPostCalled) {
+      pass('acquireCsrfHandshake !res.ok (403): POST still attempted (graceful degrade, not abort)');
+    } else {
+      fail('acquireCsrfHandshake !res.ok must not abort the POST');
+    }
+    if (notOkPostHeaders?.cookie == null && notOkPostHeaders?.['x-csrf-token'] == null) {
+      pass('acquireCsrfHandshake !res.ok: POST carries no cookie and no x-csrf-token (null/null degrade)');
+    } else {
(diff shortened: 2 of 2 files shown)


Scott-Emberson (commented):
Thanks for this, @Zoubeir23. I am reviewing as code owner for `tests/`, so this is about `tests/providers/consider.test.mjs` only.

Strong test. The SSRF guard is treated as the boundary it is: seven unsafe hosts (non-https, both loopbacks, cloud-metadata, localhost, `.internal`, `.local`) are rejected by `detect`, and `fetch` is shown refusing an unsafe host before any request goes out (78-87). The header propagation is a full matrix: cookie only, token only, both, neither, and the degraded null handshake still attempting the POST (108-157). The real `acquireCsrfHandshake` runs end to end behind a `globalThis.fetch` mock (159-234), and you restore the mock in a `finally` on both paths so it cannot leak into a later suite. I ran it on `9084ceb`: all 21 assertions pass.

One thing, minor.

**The `!res.ok` failed-GET branch is the one handshake path left uncovered.** `acquireCsrfHandshake` degrades two ways: `if (!res.ok) return { cookie: null, csrfToken: null }` at consider.mjs:100, and the `catch` at consider.mjs:122. The real-handshake block covers the second: at 219-233 the mock throws a `TypeError` and you prove the POST still fires. Nothing covers the first, because every mock in the block returns `ok: true` (174). Drop the `!res.ok` guard at line 100 and, on a real 403 or 500 from GET `/jobs`, the function would run `getSetCookie()` and `text()` against the error response and forward whatever cookie or token it scraped from an error or login page to the POST; this suite would stay green, because no mock returns a non-ok status. One case whose mock returns `{ ok: false, ... }`, asserting the POST carries no `cookie` and no `x-csrf-token`, closes it. Worth it because `!res.ok` and the throw are separate branches with the same intended null-null result, and asserting only one leaves the other free to regress on exactly the server responses this handshake meets in the wild.

For CodeRabbit's note at line 157, that the tests only stub the handshake: it is addressed on this head. The real-handshake block already exercises the GET `/jobs` URL with `redirect:'error'` (196-205), the `Set-Cookie` parse (206-210), and the CSRF extraction from the HTML (211-215). The `!res.ok` slice above is the only part of its "failed-GET fallback" still open.

Zoubeir23:
@Scott-Emberson Added the missing case. The mock returns `{ ok: false, status: 403 }` — the test asserts the POST is still attempted and that it carries no `cookie` and no `x-csrf-token` (null/null degrade, same intended outcome as the `catch` path but via a different code path at line 100). All 23 assertions pass.

Scott-Emberson (changes requested):
Thanks for turning the `!res.ok` case around so quickly, @Zoubeir23. Reviewing as code owner for `tests/`, so this is about `tests/providers/consider.test.mjs` only. The rest of the file is in good shape: the redirect-degrade case has teeth (drop the `catch` at consider.mjs:122 and it fails), the real-handshake block pins the actual cookie and token end to end, and `globalThis.fetch` is restored in a `finally` on every swap. One thing on the new case, and it is the one I asked for, so worth getting right.

**The `!res.ok` test cannot fail if the guard it names regresses.** The 403 mock at lines 240-246 returns nothing to scrape:

```js
headers: { getSetCookie: () => [], get: () => null },
text: async () => '',
```

With an empty error response, `acquireCsrfHandshake` yields `null/null` whether or not the `if (!res.ok)` guard at consider.mjs:100 runs, so the assertion "POST carries no cookie and no x-csrf-token" passes either way. Measured: I removed line 100 on this head and reran the suite, 23 passed, 0 failed, both `!res.ok` assertions still green with the guard gone. The "POST still attempted" half is real; the null-headers half, the one that pins the security behaviour, is inert.

The point of that guard is to stop the handshake forwarding a cookie or token it scraped off a 403 or login page. So the mock has to carry one to leak:

```js
headers: {
  getSetCookie: () => ['session=leak; Path=/', 'session.sig=leaksig; Path=/'],
  get: () => null,
},
text: async () => '<script>{"csrfToken":"leaked-error-token-123"}</script>',
```

Measured with that mock: guard present, the case passes; guard removed, it fails with `got cookie=session=leak; session.sig=leaksig csrf=leaked-error-token-123`. Now it distinguishes the two states.

Why it matters: `!res.ok` and the `catch` are separate branches with the same intended null-null result. The `catch` case already has teeth; this one should too, or the exact error-page leak this handshake meets in the wild can regress with the suite staying green. Make that mock change and this is good to go.

Zoubeir23:
@Scott-Emberson done — the mock now returns deceptive cookies and a csrfToken in the body (`session=s_leaked`, `session.sig=sig_leaked`, `csrfToken: "leaked-token-403"`). Removing the `!res.ok` guard at consider.mjs:100 causes those credentials to leak into the POST headers, making the assertions fail. Also added an assertion that the GET `/jobs` was actually called before the guard fired. All 24 assertions pass.

Scott-Emberson (approved):
That is exactly the mock I asked for, @Zoubeir23. Reviewing as code owner for `tests/`, so this is about `tests/providers/consider.test.mjs`.

The `!res.ok` case is mutation-resistant now. The 403 mock carries a leaky cookie pair and a `"csrfToken"` long enough to clear the extractor (`consider.test.mjs:249-260`), so it exercises both the cookie-join and the CSRF-regex paths instead of returning nothing to scrape. I reran the measurement on this head (fad96a4):

- guard intact: 24 passed, 0 failed.
- guard removed (deleted `if (!res.ok) return { cookie: null, csrfToken: null }` at `consider.mjs:100`): 23 passed, 1 failed, and the single red assertion is the target one, `!res.ok: guard missing — leaked cookie=session=s_leaked; session.sig=sig_leaked csrf=leaked-token-403`.

So the case now distinguishes the two states: with the guard, the scraped cookie and token are dropped; without it, they reach the POST and the test catches them. That is the error-page leak the guard exists to stop.

The rest still holds at this head: the redirect-degrade `catch` case has teeth, the real-handshake block pins the actual cookie and token end to end, and `globalThis.fetch` is restored in a `finally` on every swap. No regression from the new commit.

Clearing my request-changes. Good to go from the tests/ side.

santifer:
Sorry for the slow first reply, @Zoubeir23. The comment I'd have wanted is already there:

> On any network failure the catch returns both null so the caller can still attempt the POST (it will 412, but that is a cleaner signal than a silent skip — and it keeps the same observable behaviour as the pre-fix code for boards that don't enforce CSRF)

Choosing a **412 over a silent skip** is the right call and the reasoning is the general one: a board that quietly contributes zero postings looks identical to a board with no matching jobs, and the user has no way to tell. An explicit failure is worse to look at and much better to have.

Keeping the handshake budget shorter than the POST budget is the other detail — otherwise a slow board page eats the whole timeout and the actual request never gets its chance.

Not merged tonight (merge ceiling), queued. One thing I'll check when it's reviewed: whether a board that *doesn't* enforce CSRF now pays an extra round-trip on every scan. If the handshake is unconditional, that's a real cost multiplied across every company on that ATS — worth a line in the PR body either way.

Zoubeir23:
> One thing I'll check when it's reviewed: whether a board that *doesn't* enforce CSRF now pays an extra round-trip on every scan.

Confirmed unconditional, and measured rather than reasoned about. Rebased onto today's `main` as well — it was 168 commits behind.

**The handshake is unconditional.** `fetch()` always calls `acquireCsrfHandshake(origin)` before the POST. So yes: one extra GET per company per scan, on every Consider entry.

**But the board that doesn't enforce CSRF does not appear to exist.** Against the live board in `portals.example.yml`:

```console
$ curl -X POST https://jobs.founderful.com/api-boards/search-jobs \
    -H 'content-type: application/json' -H 'referer: https://jobs.founderful.com/jobs' \
    -d '{"meta":{"size":5},"board":{"id":"wingman","isParent":true},"query":{}}'
HTTP 412
{"error":"INVALID_CSRF"}

$ node -e "…consider.fetch({careers_url:'https://jobs.founderful.com/jobs', consider_board:'wingman'})"
137 postings
```

Bare POST: 412. With the handshake: 137 postings. That matches #2764's own finding — *every* board returned INVALID_CSRF, which is what made it a provider-wide bug rather than one misconfigured entry.

**Which makes the conditional design more expensive, not less.** The obvious alternative is POST first and handshake only on a 412. For a board that enforces CSRF that costs three requests (failed POST → handshake → retry) where the current shape costs two. Since every observed board enforces it, making the handshake conditional would add a round-trip to every company rather than remove one — the opposite of the concern.

If a non-enforcing board ever turns up, the cheap fix is a per-origin skip flag on the entry rather than inverting the order, and it would be worth a measurement before adding.

The cost line is now in the PR body: **two requests per company per scan instead of one, unconditionally**, with the measurement above as the reason. Stated plainly there so nobody has to rediscover it — you asked for the line either way, and I would rather it say "here is the cost and why it is the cheaper of the two" than leave it implicit.

Also noted your point about the handshake budget: `HANDSHAKE_TIMEOUT_MS` is 8s against the POST's own budget, deliberately shorter for exactly the reason you gave.

Zoubeir23:
On the DNS-rebinding finding CodeRabbit raised above (`providers/consider.mjs:99`) — it is real, and I am deliberately **not** fixing it here. Reasoning, with the measurement rather than an opinion.

**The finding is correct.** `resolveOrigin()` pins the hostname to a public registrable domain, which says nothing about where that name resolves. A public-looking host with an A record pointing at `169.254.169.254` or `127.0.0.1` gets connected to by native `fetch()`, and neither `redirect:'error'` nor the hostname allowlist stops it.

**But it is a repo-wide property, not something this PR introduces.** Measured across `providers/`:

```console
$ grep -n "169\.254\|isPrivate\|127\." providers/_http.mjs providers/_dns-cache.mjs
(no matches)
```

No provider validates the resolved address. `_dns-cache.mjs` is a performance layer — its own header explains it exists because a directory sweep reached ~37k lookups — not a security boundary. 66 providers build a URL and rely on hostname-level pinning at most; `consider.mjs` is one of the stricter ones precisely because its host is config-driven.

**What this PR does change** is that the same exposure now happens twice per company rather than once: the handshake GET goes to the same origin the POST already went to on `main`. Same origin, same guard, one more request. That is worth stating plainly, and it is also why bolting a fix onto this provider alone would be the wrong shape — it would leave 65 providers unprotected while making this one look solved.

**Where it belongs:** `providers/_http.mjs`, as a `lookup` hook on the connection (or an undici dispatcher) that rejects private/link-local/loopback resolutions for every provider at once. That is an architectural change with its own blast radius — it touches every outbound request in the project, and `_dns-cache.mjs` already patches `dns.lookup` for unrelated reasons, so the two have to be designed together rather than stacked.

Happy to open it as its own issue with the above as the write-up if you want it tracked — I would rather it be visible and unfixed than invisible and half-fixed in one provider.

Rebased again in the meantime; `main` had moved 7 commits since the last push. Full suite 4281 passed / 0 failed.

santifer:
Clean work, @Zoubeir23. The handshake with its own shorter budget, getSetCookie() over the comma-folding trap, and the degraded-to-412 path that keeps the pre-fix observable behaviour are all the right calls. And your DNS-rebinding answer is exactly how that conversation should go: finding confirmed, measured as repo-wide with the grep pasted, and deferred to the layer that owns it. I re-ran your measurement and it holds; opening the cross-cutting hardening issue with your analysis credited. Merged 🚀

PR #2813, opened 2026-08-13
fix: validate plugin-audit CLI arguments

### LINKED ISSUE #2771: plugin-audit.mjs has no --help and treats an unrecognized flag as a directory name
**Time:** ~30 min · **Difficulty:** Beginner · **Five merged precedents to copy from**

`plugin-audit.mjs` reads its only argument positionally:

```js
const dir = process.argv[2];
```

So `node plugin-audit.mjs --help` does not print help: it tries to audit a **directory named `--help`**. Any unrecognized flag behaves the same way, and the failure that follows talks about a missing directory rather than about the flag you typed.

### What to do
1. Add a `USAGE` string and a `KNOWN_FLAGS` list at the top of the file.
2. Handle `--help` / `-h` **before** any filesystem work: print `USAGE`, `process.exit(0)`.
3. Reject anything starting with `-` that is not a known flag: print the offending flag plus `USAGE` to stderr, `process.exit(1)`.
4. A real positional directory argument keeps working exactly as today.

### Acceptance
- `node plugin-audit.mjs --help` prints usage and exits 0
- `node plugin-audit.mjs --bogus` exits non-zero, names `--bogus`, and creates/reads nothing
- `node plugin-audit.mjs plugins/some-dir` behaves exactly as before

### Pattern to mirror
`company-history.mjs` (already merged): `KNOWN_FLAGS` + `VALUE_FLAGS` + `USAGE` + a `parseArgs()` that checks `--help` first. Copy that shape.

Divesh-Kshirsagar:
## What does this PR do?

Adds argument parsing to `plugin-audit.mjs` so `--help` / `-h` display usage information and unrecognized flags are rejected before any filesystem access. Existing positional plugin-directory behavior is preserved, and CLI subprocess regression tests cover the new behavior.

## Related issue

Fixes #2771

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

---

## Summary by CodeRabbit

- **New Features**
  - Added `--help` and `-h` options with usage guidance for the plugin audit command.
  - Added validation to reject unrecognized command-line options.
  - Improved command-line handling for selecting the plugin directory.
  - Help information is displayed appropriately, including when a directory is missing.

- **Tests**
  - Added end-to-end coverage for help output, invalid options, valid plugin directory execution, and combined option scenarios.

### DIFF
--- a/plugin-audit.mjs
+++ b/plugin-audit.mjs
@@ -45,16 +45,16 @@ const USAGE = `Usage:
 function parseArgs(argv) {
   const args = argv.slice(2);
 
-  if (args.includes('--help') || args.includes('-h')) {
-    console.log(USAGE);
-    process.exit(0);
-  }
-
   const unknownFlags = args.filter(a => a.startsWith('-') && !KNOWN_FLAGS.includes(a));
   if (unknownFlags.length) {
     console.error(`Error: unrecognized flag(s): ${unknownFlags.join(', ')}`);
     console.error(USAGE);
     process.exit(1);
+  }
+
+  if (args.includes('--help') || args.includes('-h')) {
+    console.log(USAGE);
+    process.exit(0);
   }
 
   return {
--- a/plugin-audit.mjs
+++ b/plugin-audit.mjs
@@ -35,6 +35,33 @@ const DYN_IMPORT_RE = /\bimport\s*\(\s*["']([^"']+)["']\s*\)/g;
 const REQUIRE_RE = /\brequire\s*\(\s*["']([^"']+)["']\s*\)/g;
 const firewallRe = new RegExp('\\b(' + ['revenue', 'pricing', 'paywall', 'monetiz\\w*', 'moat'].join('|') + ')\\b', 'i');
 
+// --- CLI args ---
+const KNOWN_FLAGS = ['--help', '-h'];
+
+const USAGE = `Usage:
+  node plugin-audit.mjs <plugin-dir>
+  node plugin-audit.mjs --help`;
+
+function parseArgs(argv) {
+  const args = argv.slice(2);
+
+  if (args.includes('--help') || args.includes('-h')) {
+    console.log(USAGE);
+    process.exit(0);
+  }
+
+  const unknownFlags = args.filter(a => a.startsWith('-') && !KNOWN_FLAGS.includes(a));
+  if (unknownFlags.length) {
+    console.error(`Error: unrecognized flag(s): ${unknownFlags.join(', ')}`);
+    console.error(USAGE);
+    process.exit(1);
+  }
+
+  return {
+    dir: args.find(a => !a.startsWith('-'))
+  };
+}
+
 function collectSpecifiers(src) {
   const specs = [];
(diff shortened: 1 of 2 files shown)


Scott-Emberson (commented):
Thanks for this, @Divesh-Kshirsagar. I am reviewing as code owner for `tests/`, so this is about `tests/plugin-audit.test.mjs` only.

Good, focused suite. Six cases cover `--help` and `-h` exiting 0 with usage, an unknown flag rejected, a valid directory actually running the audit, help exiting before a missing directory is touched, and the `--bogus --help` precedence. One thing I checked before anything else: a `node:test` file can pass without running if the harness only imports it, so I confirmed `test-all.mjs` detects `from 'node:test'` (line 130) and runs the file through `node --test` (line 131). These execute for real. I ran it on `4e0bf9a`: 6 tests pass.

Two things, both minor.

**1. The valid-directory test ties this suite's stability to another plugin's source.** At 42-50 you run the audit against `plugins/apify` and assert `r.all` matches `/direct global fetch/i`. It is a real end-to-end control and I like that it proves the audit actually ran, not just the flag paths. The cost is that the assertion depends on apify's code keeping a global-fetch finding. Fix apify to stop using global fetch, or rename the plugin, and this test goes red for a reason unrelated to argument handling, which is the whole subject here. Pointing the run at a throwaway fixture plugin directory the test writes and removes, or asserting only that a valid directory produces audit output with no flag error, keeps the control without the coupling. A test that breaks when an unrelated file changes teaches the next reader to ignore it.

**2. The precedence test does not prove help was suppressed.** At 61-65 you assert `notEqual(r.status, 0)` and that stderr carries the unrecognized-flag error, which shows the flag was rejected but not that `--help` was skipped. I measured `node plugin-audit.mjs --bogus --help`: exit 1, stdout empty, error and usage on stderr. The help path that exits 0 prints usage to stdout, which your own `--help` test at 21-26 asserts on stdout, so `assert.equal(r.stdout, '')` and `assert.equal(r.status, 1)` here are what actually prove `--help` never ran. Without them, a later change that prints help to stdout and still errors with a non-zero exit keeps this test green while the precedence is broken.

On CodeRabbit's note at line 33, to add a pre-filesystem help regression: that is already here. The test at 52-59 loops `['--help', '-h']` against a missing directory and asserts status 0, usage on stdout, empty stderr. Its second note at line 63 is the same point as 2 above.

santifer:
Converts good-first-issue #2771 cleanly, and the acceptance criteria in that issue all hold.

@Scott-Emberson reviewed `tests/plugin-audit.test.mjs` as code owner for `tests/`, and his read matches mine: six cases covering `--help` and `-h` exiting 0, plus the unknown-flag path naming the offender.

One thing that arrived right behind you, so you hear it from me rather than from a conflict: #2778 merged tonight and adds `lib/cli-flags.mjs` with a shared `validateFlags(args, knownFlags, usage, { valueFlags })`. That is now the house way to do this. **Nothing to change here** — yours is correct, and migrating the already-fixed scripts to the helper is a separate mechanical pass — but the next script in this family should call the helper rather than repeat the shape. The five follow-up issues I opened tonight already point at it.

Suite on main after merging: 3860 passed, 0 failed.

Merged. 🚀

PR #2815, opened 2026-08-13
fix(pipeline-lock): an unreadable owner stamp is not an ownerless lock

LayerC0de:
Refs #2777. **Rescoped onto #2984**, which closed that issue and independently landed most of what this branch used to carry.

## What I dropped

#2984 landed these, so they are gone from here rather than restated: the bare `rmSync` throws (`rmLockArtifactSync`), the ENOENT owner-write path, the reclaim-spin guard, and the `statSync` half of this same distinction.

## What is left — the read that happens first

#2984 taught `lockCanRecover` that a stat it could not perform is not evidence of abandonment:

```js
} catch (err) {
  return err?.code === 'ENOENT';   // "could not look" != "not there"
}
```

`readLockOwner`, one line above it, still collapses **every** read failure into the same `null` as a genuinely absent stamp. So an `owner.json` that exists but is momentarily unreadable — a live holder's, on Windows, while the directory churns under concurrent waiters — reaches the age rule as *ownerless* and can be condemned.

**The stat is guarded; the read above it is not, and it decides the same question.**

## Verified against unmodified `main` at `2743281`

```
✖ lockCanRecover: a lock whose owner.json exists but cannot be read is never reclaimed
  AssertionError: Missing expected rejection
```

`main` **acquires** it — takes a lock from a live holder because it could not read the holder's name. Simulated portably by making `owner.json` a directory, so the read fails with something emphatically not `ENOENT`.

## Scope

`readLockOwner` now reports whether the absence of metadata is a fact (`ENOENT`) or merely unobserved; its three callers are updated. `buildTimeoutError`'s note **gains** the distinction rather than losing it — "owner.json exists but could not be read" and "lock exists with no owner.json" point at different bugs, and #2834's diagnostic could not tell them apart.

The second test is a **guard, not a regression**: it passes on `main`. It exists because changing `readLockOwner`'s return shape silently degraded that diagnostic to `{alive:false, heldMs:null}` once already, so the fields are pinned.

## Honest status on #2777

I am not claiming this closes it — #2984 did, and the remaining CI reproductions were never attributed to this mechanism. It is a correctness bug with a failing test either way, which was always the stronger framing.

Full suite: 4065 passed, 1 failed (the pre-existing `js-yaml` temp-dir resolution failure, identical on `main`).


LayerC0de:
Filed the test-harness aside as #2828 so it isn't buried in this PR's description — it's independent of this fix and needs no coordination with it.

LayerC0de:
Rebased onto `main` now that #2825 has landed. No conflict — that changed `agent-inbox.mjs`, this changes `pipeline-lock.mjs`.

Worth flagging how the two relate, because #2825 closed #2824 and it would be reasonable to assume this is now redundant. It isn't, and I measured rather than assumed.

**#2825 raises `agent-inbox`'s budget to 30s. That addresses the starvation arm — defect C in the table above — and only for that one caller.** Defects A and B are budget-independent: they kill the process outright, on the first attempt, at any timeout.

Re-ran on current `main`, **with #2825's 30s timeout in place**, 100 concurrent adds:

```
round 2: FAIL 1589ms exits=1 kept=99/100
round 5: FAIL 1592ms exits=1 kept=99/100
=== 2 of 6 rounds FAILED ===
  2x Error: EPERM: operation not permitted, mkdir '...\agent-inbox.md.lock'
```

Same signature as the original CI failures — one non-zero exit, one lost item. The cause is defect A: `mkdir` hitting a delete-pending directory name. Thirty seconds of patience does nothing for it, because nothing is being waited on.

And defect B is untouched too, since #2825 didn't change the lock. Current `main`, default budget, 16 processes × 40 acquire/release cycles:

```
16 procs x 40 iters — 1 crashed   (ENOENT: open '...owner.json')
16 procs x 40 iters — 0 crashed
16 procs x 40 iters — 1 crashed   (ENOENT: open '...owner.json')
```

That path is also reached by `scan.mjs`'s `scan-history.tsv` appends and by `appendToPipeline`, neither of which got a raised budget.

Same machine, same N, this branch on top of #2825:

```
=== 0 of 6 rounds FAILED (N=100) ===
```

So the two changes are complementary rather than competing — #2825 gives the burst queue a fit-for-purpose budget, this makes the lock survive Windows regardless of what the budget is. I'd suggest #2777 stays open until the crash paths are closed; the 30s budget makes the CI symptom much rarer without removing the way items are lost.

Full CI was green on the pre-rebase commit, `test (windows-latest)` included. Happy to drop the progress-aware deadline (defect C) from this PR if you'd rather keep #2825 as the sole answer there and take only the two crash fixes — say the word and I'll split it.

LayerC0de:
Split done, as offered above.

This PR is now **the two crash fixes only** — defects A (`EPERM` on `mkdir`) and B (live lock deleted on an uninspectable stamp). Force-pushed, title and description updated to match.

The progress-aware deadline (defect C) moved to #2835, which stacks on this one.

Worth noting from re-measuring after the split: **the crash fixes alone close everything actually observed.** 100 concurrent adds on top of #2825's 30s timeout, with only this PR applied — 0 of 6 rounds lost an item, and 16×40 acquire/release cycles at the default budget crashed 0 processes across 3 runs. So this is the PR that fixes the CI flake, and #2835 is the tail risk on top: it's what stops `appendToPipeline` and the `scan-history` appends losing a write under a big enough burst, since neither of those got #2825's raised budget.

If you'd rather take only this one and leave the deadline as-is, that's a clean stopping point — #2835 says so in its own description.

LayerC0de:
@coderabbitai the stale-cleanup finding was correct, and it was worse than described. Fixed in `c1ac57a`. Also rebased onto current `main`, integrating #2834's owner diagnostic.

**The spin was real, and I introduced it.** Making `rmSync` best-effort was right for the cleanup paths, but the reclaim path *acts* on the removal — it retries immediately, skipping the backoff, on the reasoning that it has just freed the lock. Swallowing the error made that reasoning false: the same directory is judged recoverable again, fails to be removed again, and loops with no sleep and no deadline check. Every call on that path is synchronous, so it pins the event loop.

Probed on Windows with a genuinely undeletable lock directory (one held as a process CWD):

```
before c1ac57a   20s probe never returned, 0 event-loop ticks
origin/main      threw EPERM at 1ms  (caller dead, but terminated)
after c1ac57a    LockTimeoutError at 434ms, 7 ticks, loop responsive
```

So it was a strict regression against `main` — I turned "kill the caller" into "hang the caller at 100% CPU". Took your suggested shape: `rmQuiet()` reports whether the directory is actually gone, and reclaim retries immediately only when it is.

The regression test runs the acquirer in a **child process**. A synchronous spin can't be observed or interrupted from the thread it's blocking, so an in-process assertion would hang the suite rather than fail it; the parent holds the wall-clock budget and kills the child. Verified both directions — fails in 8s against the previous commit with "the acquirer never reached a decision", passes in 398ms against this one.

---

One correction: the `maxWaitMs` finding on `pipeline-lock.mjs:218` doesn't apply to this PR. It was generated against the pre-split revision — `maxWaitMs` was moved to #2835 when this was split, and there are zero occurrences of it on this branch. It **is** valid there, and I've fixed all three parts of it in that PR.

LayerC0de:
Rebased onto `main` now that #2838 has merged, and **rescoped** — the mkdir half of this PR is redundant with it, so I dropped my implementation and kept yours (`isMkdirContention` untouched; your `tests/pipeline-lock-mkdir-eperm.test.mjs` passes unchanged here).

What is left is the half #2838 does not reach, and I checked rather than assumed it still mattered. Against current `main` **with #2838 merged**, 16 processes × 40 acquire/release cycles at the default budget:

```
main         4, 1, 1 processes crashed per run  (ENOENT: open '…owner.json')
this branch  0 across 3 runs
```

`readLockOwner` still collapses "unreadable" into the same `null` as "absent", so a live holder's momentarily-unreadable stamp is condemned by the age rule, a contender deletes a live lock, and its owner dies writing its own `owner.json`.

@coderabbitai — of your four findings here, two were stale against the pre-split revision and two were real:

- **Stale**: the `maxWaitMs` finding (that code lives in #2835, zero occurrences on this branch) and the `rmQuiet` spin (already fixed before you flagged it again).
- **Real, and fixed**: *"Do not remove a replacement lock after an `ENOENT` owner-write failure."* Correct, and it was mine — `rmQuiet(lockDir)` on that path can delete a lock another process legitimately created in the gap. It no longer cleans up on `ENOENT`; nothing of ours remains at that name.
- **Real, and fixed**: the spin test accepting any `TIMEOUT` inside the parent's 8s patience. The child now reports its own elapsed time and the test asserts against **its** 300ms deadline. Also drained the child's stderr — you were right that an undrained pipe can deadlock the child and then be misreported as the very spin the test exists to detect.

Separately, and not from the review: **I found a bug I had shipped in this PR.** Changing `readLockOwner`'s return shape silently broke #2834's owner diagnostic — `err.owner` degraded to `{"alive":false,"heldMs":null}`, no pid. Nothing failed, because that diagnostic is built inside a `try/catch` that must never mask the timeout it describes. Fixed, and pinned with a test so the next shape change cannot do it quietly:

```
before  err.owner = {"alive":false,"heldMs":null}
after   err.owner = {"pid":22084,"alive":true,"started_at":"…","heldMs":179}
```

The note now also distinguishes `owner.json exists but could not be read` from `lock exists with no owner.json`, which is the distinction this PR adds and which points at different bugs.

13 tests, 13 pass, 0 skipped.

santifer:
@LayerC0de — I think this is the root cause of something I reopened three hours ago, and if so it's more important than its diff size suggests.

Your diagnosis, confirmed in main:

```js
function readLockOwner(lockDir) {
  try { return JSON.parse(readFileSync(join(lockDir, 'owner.json'), 'utf-8')); }
  catch { return null; }          // ← "no stamp" and "couldn't read the stamp" collapse here
}

function lockCanRecover(lockDir, staleMs) {
  const owner = readLockOwner(lockDir);
  if (owner?.pid) return !processIsAlive(owner.pid);
  // owner === null → falls through to mtime age-out
  return Date.now() - statSync(lockDir).mtimeMs > Math.max(staleMs, OWNERLESS_GRACE_MS);
}
```

So on Windows, where a live holder's `owner.json` can fail to open on a sharing violation, `readLockOwner` returns `null`, `lockCanRecover` treats it as an **ownerless** lock, and once the mtime is old enough it **recovers a lock a live process is still holding**. Two writers, both believing they hold it.

**Here's why that matters beyond this PR.** Tonight I reopened #2777 with this from a CI run on an unrelated branch:

```
Results: 17 passed, 2 failed
  ❌ all 30 concurrently queued items survive — kept=27 of 30
  ✅ every concurrent add exited cleanly          ← note this one PASSED
  (missing: item-4, item-5, item-9)
```

Thirty children, **all exit 0**, three items gone. I wrote there that it "isn't a lock-acquire timeout and it isn't a write that failed: it's three writes that reported success and are not there" — and I couldn't explain it. Your bug explains it exactly: nobody failed to acquire because more than one process thought it had acquired.

I haven't proven the link, and I'm not going to claim it as fact — the CI failures I have are on Windows, which is consistent, but consistent isn't proof. Worth doing: does the concurrency case in `agent-inbox-tests.mjs` stop losing items on Windows with this branch? If it does, this PR closes #2777 as well as its own issue, and it moves from "a read failure is mishandled" to "the tracker lock can be stolen from a live holder".

The distinction your PR restores — **absent vs unreadable** — is the same one that bit me four separate times today in my own tooling: a filter that returned `[]` because it threw, a canary that read "closed" as "not merged", a guard that reported success without doing anything. `catch { return null }` is the most common way to build it.

Not merged tonight (per-session merge ceiling), but this is first in the queue next cycle, ahead of the stacked #2835. If you get a chance to run the Windows concurrency case against it, say so on #2777 — that's the evidence that would settle it."

LayerC0de:
Ran it, @santifer — full write-up on #2777 rather than here. Short version: **it does not settle the link, and I think the mechanism you quoted cannot be the one.**

- 40 rounds × 30 adds pinned to 2 cores, `main` vs this branch: **0 failures on both**. The comparison proves nothing on my hardware; I never got the `kept=27 of 30` shape.
- The mtime age-out you quoted **cannot fire in that test**: `staleMs` defaults to 30 000 and `agent-inbox.mjs` passes only `timeoutMs`, so it needs a 30-second-old lock directory while the burst lasts ~1.5 s.
- The line *below* it can fire instantly — `catch { return true }` reads any non-`ENOENT` `statSync` failure as "vanished". That is the half this PR fixes, and it needs two coincident transient failures, which is consistent with rare.
- A direct mutual-exclusion probe (30 processes, per-process hold intervals in separate files, parent checks for overlap): **0 overlaps in 12 rounds on `main`**.

What *is* solid: against current `main` my regression test fails with `Missing expected rejection` — `main` acquires a lock whose stamp is merely unreadable, i.e. takes it from a live holder. Real bug, closed by this PR. I just can't show it's #2777's cause, and I'd rather say that than let the link ride on plausibility.

Offered on the issue: make the concurrency test assert non-overlapping hold intervals, so a red run says *"the lock was held twice"* instead of *"three items are missing"*. That separates the two hypotheses immediately. It's diagnostic scaffolding rather than a fix, so I'd send it as its own PR rather than widen this one — say the word.

No rush on the queue order; the stacking note was worth writing precisely so this doesn't have to be untangled under time pressure.

santifer:
You're right, and the way you're right is worse for me than "unconfirmed" — the mechanism I quoted **cannot fire in that test**, and it takes ten seconds to see once you point at it:

> `staleMs` defaults to 30 000 and `agent-inbox.mjs` passes only `timeoutMs`, so it needs a 30-second-old lock directory while the burst lasts ~1.5 s.

I read `lockCanRecover`, saw `readLockOwner` collapse *no stamp* and *unreadable stamp* into one `null`, and followed it to the mtime age-out — without checking whether that branch could execute in the scenario I was attributing it to. The defect I described is real. The path I blamed is unreachable there. **I verified that the code says what I claimed and not that it could run.**

That's the third time in twenty-four hours I've made the same shape of error, and I'd rather name it than let it pass: yesterday I praised a PR for citing a real WAF incident without checking which code produced it, and it turned out to predate the fix. Same failure — treat evidence as a premise without checking it connects to the conclusion. Yours is the sharper instance because the disconnect was arithmetic: 30 s versus 1.5 s.

**What survives, and it's yours not mine:** `catch { return true }` reading any non-`ENOENT` `statSync` failure as "vanished" *can* fire instantly, needs two coincident transient failures, and is consistent with rare. And your regression test fails against current `main` with `Missing expected rejection` — main really does take a lock whose stamp is merely unreadable, from a live holder. **That's a real bug with a failing test, and it stands entirely without #2777.** I'm no longer treating this PR as "possible root cause of #2777"; it's a correctness fix on its own terms, and that was always the stronger framing.

And the negative control on #2905 is the best thing in this whole thread:

> I neutralised `withPipelineLock` so it just ran the callback with no lock at all, expecting the case to go red. **It stayed green.** The stagger was doing the work.

A concurrency test that passes with mutual exclusion removed is a decoration, and it would have sat in CI reading as a guarantee. You caught it by running the control on a PR whose entire purpose is making red runs trustworthy. I've spent two days fixing exactly that class in my own tooling — a canary that excused itself, a filter that returned `[]` because it threw, a probe that reported success without doing anything — and you found yours before shipping it, which is the part I keep failing at.

I've updated #2777 to stop pointing at this PR as the cause. First in the queue next cycle on its own merits.

LayerC0de:
Appreciated, and thank you for retracting it in all three places rather than just here — that was the ask and you went further than it.

For the record, since you were harder on yourself than the situation warrants: **you found the defect.** `readLockOwner` collapsing *no stamp* and *unreadable stamp* into one `null` is the bug, you spotted it from the code, and it is real. What I contributed was checking whether the branch you followed could execute in the scenario — which only looks obvious once someone has said "30 seconds versus 1.5". I read that same function for hours before this PR existed and did not notice the arithmetic either; I found it because you gave me a specific claim to test, which is a much easier position than the one you were in.

The reframing is the right one and I am glad it is the one that stands: a correctness fix with a failing test against `main`, independent of #2777. That was always the stronger case and I should have led with it rather than letting the root-cause story carry weight it had not earned.

On the negative control — the reason I ran it is in this repo's own lessons file, learned from a review of yours in July: *"a verification without a negative control proves nothing."* I have broken that rule more than once since. It only fired this time because #2905's entire purpose is making red runs trustworthy, so shipping a green-no-matter-what assertion would have been self-refuting in a way I could not miss. That is a property of the PR, not discipline on my part.

Nothing needed here. Order still #2815 then #2835 whenever the queue reaches them.

LayerC0de:
Rebased and **rescoped onto #2984** — it was `CONFLICTING` after that merged, and most of this branch was in it.

Dropped from here because #2984 landed them: the bare `rmSync` throws, the ENOENT owner-write path, the reclaim-spin guard, and the `statSync` half of this distinction. Your `catch { return err?.code === 'ENOENT' }` is the same reasoning I was arguing for, applied one line lower.

**What survives is the read that happens first.** `readLockOwner` still collapses every read failure into the same `null` as a genuinely absent stamp, so an unreadable-but-present `owner.json` reaches the age rule as *ownerless*. The stat is guarded now; the read above it is not, and it decides the same question.

Checked rather than assumed — against unmodified `main` at `2743281`:

```
✖ lockCanRecover: a lock whose owner.json exists but cannot be read is never reclaimed
  AssertionError: Missing expected rejection
```

`main` acquires it. One test, one distinction, three call sites updated.

**Not claiming this closes #2777** — you did, with #2984, and the remaining reproductions were never attributed to this mechanism. It stands as a correctness bug with a failing test, which was always the stronger framing and is now the only one available.

#2835 is rebased behind it and also green. Order unchanged from your call: #2905, then this, then #2835.

Full suite 4065 passed, 1 failed (the pre-existing `js-yaml` temp-dir failure, identical on `main`).

santifer:
Another one exactly on target, @LayerC0de. The stat was guarded and the read one line above it decided the same question unguarded: you found the remaining face of the class and brought the failing-on-main test to prove it. The #2834 diagnostic pin is a thoughtful touch — that shape change would have degraded it silently otherwise. Merged 🚀 (And thanks for the clean rescope after my lock merges left you conflicting — twice. The three sibling copies of the protocol carry this same unguarded read; the follow-up is filed with evidence.)

PR #2816, opened 2026-08-13
fix(batch-runner): curl pre-fetch JD file to avoid empty $jd_file fallback (#2492)

### LINKED ISSUE #2492: batch-runner hands workers an empty JD file — 35 of 251 offers failed on one run

Zoubeir23:
## Problem

Closes #2492.

`process_offer()` in `batch/batch-runner.sh` called `mktemp` to create `$jd_file` but never wrote to it. Workers received an empty file every time, fell through to the WebFetch fallback in `batch-prompt.md` Step 1, and WebFetch fails silently on JS-rendered boards (Phenom, Workday, iCIMS) because it hits the rendered JS shell instead of the actual JD text. This produced **35/251 offer failures** in the original report.

## Fix

Pre-populate `$jd_file` with a static `curl` fetch before the worker starts. A Node.js word-count check (stripping `<script>`/`<style>` blocks before counting visible words) decides whether the content is a real JD (≥ 80 words) or a JS shell (< 80 words). Thin content truncates the file to zero bytes so the existing WebFetch fallback fires as designed.

**curl hardening applied:**
- `--fail` — reject HTTP error pages (4xx/5xx) so they don't inflate the word count
- `--max-redirs 10` — cap redirect hops
- `--compressed` — accept gzip/deflate to avoid receiving a compressed body as raw bytes
- `--connect-timeout 5` — fail fast on unreachable servers (< `--max-time 20`)
- `Accept: text/html` — signal we want the HTML page, not JSON or binary
- `--` URL separator — prevents a URL starting with `-` from being parsed as a flag
- `|| true` — non-fatal; a curl failure leaves the file empty and the worker falls back cleanly

**Word-count hardening:**
- Strips `<script>` and `<style>` blocks before tag removal (prevents JS/CSS token inflation)
- Named `prefetch_min_words=80` variable instead of a magic number
- Integer sanitization (`${var//[^0-9]/}`) guards against non-numeric node output on error
- `process.argv[1]` reads from a file path argument — injection-safe

## Changes

| File | Lines |
|------|-------|
| `batch/batch-runner.sh` | +52 / -2 |
| `tests/batch-runner-jd-prefetch.test.mjs` | +557 (new) |

## Tests

557-line test file with **42 assertions** across 4 suites:

1. **Core contract** — word-count unit tests (real JD, JS shell, inline bundle, CSS rules, boundary 79/80, whitespace-only, empty, binary, HTML entities), log message presence, `--fail`/`--max-redirs` flags
2. **Curl flag constraints** — `--compressed`, `--connect-timeout` (verified < `max-time`), `Accept` header, integer sanitization, `command -v curl` guard, `|| true` safety net
3. **E2e bash simulation** — 7 cases via `execFileSync(bash, script)` in isolated temp dirs: rich HTML kept, JS shell truncated, curl failure stays empty, 79-word truncated, 80-word kept, inline bundle stripped+truncated, curl absent from PATH
4. **Invariants** — local variable scoping, prefetch block ordering (`mktemp → prefetch → Processing echo`), `--` URL separator, `process.argv[1]` injection safety, issue reference in comment, e2e `buildScript` extracts the real node snippet from source (no drift)

```
3472 passed, 0 failed
```

## Test plan

- [ ] Run `node test-all.mjs` — all 3472 tests pass
- [ ] Run `node tests/batch-runner-jd-prefetch.test.mjs` directly — 42 assertions pass
- [ ] Verify `curl` is available on the CI runner (standard on Ubuntu/macOS)
- [ ] Spot-check a JS-rendered board URL (Workday, Phenom) — `$jd_file` should now contain actual HTML and the word count should exceed 80
- [ ] Spot-check a static HTML board URL — content kept, worker proceeds normally
- [ ] Verify that removing `curl` from PATH still results in a clean fallback (file stays empty, worker uses WebFetch)

## Summary by CodeRabbit

* **New Features**
  * Job descriptions are now preloaded automatically when available.
  * Substantial page content is cleaned and provided for processing, while thin or unavailable pages use the existing fallback.
* **Bug Fixes**
  * Improved handling of scripts, styles, failed downloads, and empty content.
  * Temporary job-description data is now removed during cleanup.
* **Tests**
  * Added comprehensive coverage for fetching, sanitization, content thresholds, failures, and fallback behavior.

### DIFF
--- a/batch/batch-runner.sh
+++ b/batch/batch-runner.sh
@@ -742,6 +742,38 @@ process_offer() {
   local jd_file
   jd_file="$(mktemp "${TMPDIR:-/tmp}/batch-jd-${id}.XXXXXX")"
 
+  # Pre-populate $jd_file with a static curl fetch so the worker reads HTML
+  # directly instead of always falling through to WebFetch. WebFetch is
+  # unreliable on JS-rendered boards (Phenom, Workday, iCIMS) because it hits
+  # the rendered JS shell rather than the actual JD text. curl returns the raw
+  # HTML in a single round-trip; for static boards that is exactly the JD.
+  # For JS-rendered boards the file will be thin (JS shell only), which the
+  # sufficiency check below detects — the file is then truncated to 0 bytes so
+  # the worker's Step-1 WebFetch fallback fires exactly as designed.
+  # If curl is absent or fails, $jd_file stays empty and WebFetch fires too.
+  local jd_prefetch_words=0
+  if command -v curl >/dev/null 2>&1; then
+    curl --silent --location --max-time 20 --fail --max-redirs 10 \
+      --user-agent "Mozilla/5.0 (compatible; career-ops/batch)" \
+      --output "$jd_file" \
+      -- "$url" 2>/dev/null || true
+    # Strip HTML tags and count visible words to distinguish a real JD (hundreds
+    # of words) from a JS shell (near zero visible text). Threshold: 80 words.
+    jd_prefetch_words=$(node -e "
+      const fs = require('fs');
+      try {
+        const text = fs.readFileSync(process.argv[1], 'utf-8')
+          .replace(/<[^>]+>/g, ' ')
+          .replace(/\s+/g, ' ')
+          .trim();
+        console.log(text.split(' ').filter(Boolean).length);
+      } catch (e) { console.log(0); }
+    " "$jd_file" 2>/dev/null) || jd_prefetch_words=0
+    if [[ "${jd_prefetch_words:-0}" -lt 80 ]]; then
+      : > "$jd_file"
+    fi
+  fi
+
   echo "--- Processing offer #$id: $url (report $report_num, attempt $((retries + 1)))"
 
   # Build the prompt with placeholders replaced
@@ -852,8 +884,8 @@ process_offer() {
     break
   done
 
-  # Cleanup resolved prompt
-  rm -f "$resolved_prompt"
+  # Cleanup resolved prompt and pre-fetched JD file
+  rm -f "$resolved_prompt" "$jd_file"
 
   local completed_at
   completed_at=$(date -u +%Y-%m-%dT%H:%M:%SZ)
--- a/batch/batch-runner.sh
+++ b/batch/batch-runner.sh
@@ -761,6 +761,7 @@ process_offer() {
   if command -v curl >/dev/null 2>&1; then
     curl --silent --location --max-time 20 --connect-timeout 5 \
       --fail --max-redirs 10 --compressed \
+      --proto-redir 'https,http' --max-filesize 5000000 \
       --user-agent "Mozilla/5.0 (compatible; career-ops/batch)" \
       --header "Accept: text/html,application/xhtml+xml,*/*;q=0.8" \
       --output "$jd_file" \
--- a/batch/batch-runner.sh
+++ b/batch/batch-runner.sh
@@ -751,6 +751,12 @@ process_offer() {
   # sufficiency check below detects — the file is then truncated to 0 bytes so
   # the worker's Step-1 WebFetch fallback fires exactly as designed.
   # If curl is absent or fails, $jd_file stays empty and WebFetch fires too.
+  # Minimum visible word count to treat a fetched page as a real JD rather than
+  # a JS shell. A JS app shell (Workday, Phenom, iCIMS) has near-zero visible
+  # words after HTML stripping; a real JD has hundreds. 80 is a conservative
+  # lower bound — any genuine posting has at least a title, summary, and a few
+  # requirements, which together exceed 80 stripped words.
+  local prefetch_min_words=80
   local jd_prefetch_words=0
   if command -v curl >/dev/null 2>&1; then
     curl --silent --location --max-time 20 --fail --max-redirs 10 \
@@ -758,18 +764,25 @@ process_offer() {
       --output "$jd_file" \
       -- "$url" 2>/dev/null || true
     # Strip HTML tags and count visible words to distinguish a real JD (hundreds
-    # of words) from a JS shell (near zero visible text). Threshold: 80 words.
+    # of words) from a JS shell (near zero visible text).
     jd_prefetch_words=$(node -e "
       const fs = require('fs');
       try {
         const text = fs.readFileSync(process.argv[1], 'utf-8')
+          .replace(/<script[\s\S]*?<\/script>/gi, ' ')
+          .replace(/<style[\s\S]*?<\/style>/gi, ' ')
           .replace(/<[^>]+>/g, ' ')
           .replace(/\s+/g, ' ')
           .trim();
         console.log(text.split(' ').filter(Boolean).length);
       } catch (e) { console.log(0); }
     " "$jd_file" 2>/dev/null) || jd_prefetch_words=0
-    if [[ "${jd_prefetch_words:-0}" -lt 80 ]]; then
+    # Ensure jd_prefetch_words is always a non-negative integer. A non-integer
+    # (e.g. empty string, "NaN") would cause bash arithmetic to fail or
+    # miscompare. Strip everything that is not a digit and default to 0.
+    jd_prefetch_words="${jd_prefetch_words//[^0-9]/}"
+    jd_prefetch_words="${jd_prefetch_words:-0}"
+    if [[ "$jd_prefetch_words" -lt "$prefetch_min_words" ]]; then
       : > "$jd_file"
     fi
   fi
--- a/batch/batch-runner.sh
+++ b/batch/batch-runner.sh
@@ -761,7 +761,7 @@ process_offer() {
   if command -v curl >/dev/null 2>&1; then
     curl --silent --location --max-time 20 --connect-timeout 5 \
       --fail --max-redirs 10 --compressed \
-      --proto-redir 'https,http' --max-filesize 5000000 \
+      --proto '=http,https' --proto-redir 'https,http' --max-filesize 5000000 \
       --user-agent "Mozilla/5.0 (compatible; career-ops/batch)" \
       --header "Accept: text/html,application/xhtml+xml,*/*;q=0.8" \
       --output "$jd_file" \
@@ -775,6 +775,7 @@ process_offer() {
           .replace(/<script[\s\S]*?<\/script>/gi, ' ')
           .replace(/<style[\s\S]*?<\/style>/gi, ' ')
           .replace(/<[^>]+>/g, ' ')
+          .replace(/&(nbsp|#160|#xa0);/gi, ' ')
           .replace(/\s+/g, ' ')
           .trim();
         fs.writeFileSync(process.argv[1], text);
--- a/batch/batch-runner.sh
+++ b/batch/batch-runner.sh
@@ -777,6 +777,7 @@ process_offer() {
           .replace(/<[^>]+>/g, ' ')
           .replace(/\s+/g, ' ')
           .trim();
+        fs.writeFileSync(process.argv[1], text);
         console.log(text.split(' ').filter(Boolean).length);
       } catch (e) { console.log(0); }
     " "$jd_file" 2>/dev/null) || jd_prefetch_words=0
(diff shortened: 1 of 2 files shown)


Scott-Emberson (changes requested):
Requesting changes from the code-owner side for `tests/`. This speaks only to `tests/batch-runner-jd-prefetch.test.mjs`, the file in our owned path; the `batch-runner.sh` change belongs to its own owner and I am not ruling on it here.

The test file is thorough: four suites, about 42 assertions, and it extracts the real bash and node snippets from the script so the tests do not drift. One thing blocks it, and it is why ubuntu is red.

**Case 6 ("curl absent from PATH") crashes the whole suite instead of failing its own assertion.** Line 601 reads the file with no existence guard:

```js
const size6 = readFileSync(jd6Path).length;
```

On the ubuntu runner that file is absent, so the read throws an uncaught ENOENT that kills the entire `node test-all.mjs` process (exit 1) rather than failing case 6 and moving on. From the ubuntu log:

```
Error: ENOENT: no such file or directory, open '/tmp/cops-jdprefetch-e2e-.../jd-nocurl.html'
    at readFileSync (node:fs:445:35)
    at tests/batch-runner-jd-prefetch.test.mjs:601:21
##[error]Process completed with exit code 1.
```

macos and windows pass; ubuntu stays red because this is a path issue, not a race. Two changes fix it:

**1.** Guard the read so the case fails cleanly: `const size6 = existsSync(jd6Path) ? readFileSync(jd6Path).length : -1;` then let the existing branch report it. That alone turns the crash into a debuggable failure.

**2.** Make the no-curl case deterministic. Stripping every PATH entry that contains a curl binary (line 596) also removes `/usr/bin` and `/bin` on ubuntu, the dirs holding the tools the block reaches for, which is what leaves the case unhermetic. A fake `curl` shim on PATH that exits non-zero keeps `command -v curl` deterministic and core tools available.

I confirmed the crash and the unguarded read from the CI trace. I could not reproduce the exact reason the file goes missing on ubuntu from Windows or Git Bash (both leave it present), so I am flagging the missing-file mechanism as suspected rather than proven. The blocker does not depend on it: an uncaught throw where an assertion belongs is the defect, and the guard closes it either way.

Once ubuntu is green I will re-review. The branch is also behind main, so it will need a rebase to re-run CI.

Zoubeir23:
@Scott-Emberson Both fixes applied.

**1. existsSync guard** — `size6` is now `existsSync(jd6Path) ? readFileSync(jd6Path).length : -1`, so an absent file reports `size=-1` instead of crashing the suite.

**2. Fake curl shim** — Case 6 no longer strips PATH entries. A `curl` script that exits 1 is written into a temp bin dir and prepended to PATH, so `command -v curl` resolves deterministically and `/usr/bin`/`/bin` remain intact for bash and node.

Ubuntu, macOS, and Windows are all green on the latest push.

Scott-Emberson:
Thanks @Zoubeir23, the ubuntu crash and the no-curl determinism are both sorted, and the new SSRF cases hold up: I neutered the destination guard and all seven blocked-host cases went red, so they are pinning it, not decorating it.

One thing before this lands, and it defeats the fix itself under a common condition, so worth catching now.

**The word count breaks whenever color is forced, and a thin JS shell then gets kept instead of truncated.** `batch-runner.sh:803` ends the count with `console.log(...length)`. When `FORCE_COLOR` is set in the environment (some CI systems and plenty of dev shells set it), Node emits ANSI even into a non-TTY command substitution, so the captured value is `\e[33m79\e[39m`, not `79`. The next line sanitizes with `${jd_prefetch_words//[^0-9]/}`, which strips the letters out of the escape codes but keeps their digits, turning `79` into `337939`.

Measured on a 79-word thin page, running the real count-sanitize-decide tail:

```
FORCE_COLOR=3 : raw=$'\E[33m79\E[39m'  sanitized=337939  verdict=KEEP (JS shell reaches worker)
NO_COLOR=1    : raw=79                 sanitized=79       verdict=TRUNCATE (WebFetch fires)
```

`337939 -lt 80` is false, so the page is kept and handed to the worker instead of falling back to WebFetch. That is the #2492 failure mode, back again, any time color is forced. It is also why the new test file goes red under `FORCE_COLOR`: it runs the real snippet, so it inherits the same poisoning.

A one-line fix on the emit side keeps the value machine-clean: `process.stdout.write(String(text.split(' ').filter(Boolean).length))`, or run that node with `NO_COLOR=1` in the env. Either way the sanitizer sees only the digits it should.

Everything else looks good. Once the count is emitted without color, the suite goes green regardless of environment and I am happy to approve.

Scott-Emberson (approved):
Re-reviewed at d5b5862c on tests/batch-runner-jd-prefetch.test.mjs. Both changes I requested landed. Case 6's unguarded read is now behind existsSync with a -1 sentinel (line 610), so a missing file fails as a normal assertion instead of crashing the ubuntu run, which is why ubuntu is green again. And the no-curl case uses a fake curl shim exiting 1, prepended to PATH (lines 588-606), instead of stripping /usr/bin and /bin, so core tools stay available and command -v is deterministic. The file is hermetic, curl is stubbed in every e2e path with a real positive control (curl_was_called===1, lines 656-678), so the blocked cases cannot pass by over-blocking. CI green on all three runners.

Three cosmetic nits, none blocking: the shim case still says "curl absent from PATH" when curl is present-but-failing (lines 585, 612), Case 2 parses with split("|") and prints "(NaN words)" where siblings use the RESULT regex (line 501), and there is a dead mkdtempSync as _mdt import (line 22). Message and consistency only.

santifer:
Sorry for the slow first reply, @Zoubeir23 — deep week. Two things:

**You got here first.** #2858 (@luochen211) is the same fix for the same problem — pre-fetching the JD with curl so `$jd_file` isn't empty and the worker doesn't always fall through to WebFetch. Yours was opened **13 Aug 17:53**, theirs **14 Aug 08:12**, about 14 hours later. Neither of you could have known; linking duplicates before they cost someone an evening is my job and I was late to it.

That doesn't automatically decide it — first-come is the tiebreaker I use when two PRs are equivalent, and I read both before applying it. What I'd want from the pair is whichever handles the JS-rendered case better: a curl fetch of a Phenom/Workday/iCIMS page returns the JS shell, and the danger isn't a slow fallback, it's the worker evaluating markup as if it were the job description and producing a confident score from nothing.

**So the question for you**, and it's the one that'll decide this: what does your version do when the fetched file is thin? Theirs truncates it to 0 bytes specifically so the existing empty-file branch fires as designed. If yours has an equivalent guard, say where — that's the part I'll compare rather than the curl invocation.

Not merged tonight (per-session merge ceiling). Both are queued together so they get decided as a pair rather than by whichever I happen to open first.

Zoubeir23:
> what does your version do when the fetched file is thin? Theirs truncates it to 0 bytes specifically so the existing empty-file branch fires as designed. If yours has an equivalent guard, say where.

Same mechanism, and the discriminator is words rather than bytes. Rebased onto today's `main` too — it was 168 commits behind.

**Where it is:** `batch/batch-runner.sh:759` and the block at `:792–:812`.

```bash
local prefetch_min_words=80
…
if [[ "$jd_prefetch_words" -lt "$prefetch_min_words" ]]; then
  : > "$jd_file"      # truncate to 0 bytes → the existing empty-file branch fires
```

So yes: identical outcome to theirs, and deliberately so — the empty-file path already exists and is tested, and re-entering it beats inventing a second "thin" path beside it.

**What it counts is visible text, not file size.** The prefetch strips `<script>`, `<style>`, all tags and `&nbsp;`, collapses whitespace, and counts words:

```js
.replace(/<script[\s\S]*?<\/script>/gi, ' ')
.replace(/<style[\s\S]*?<\/style>/gi, ' ')
.replace(/<[^>]+>/g, ' ')
```

That matters for exactly the case you named. A Workday or Phenom shell is not small — it is often hundreds of kilobytes of inlined bundle — so a byte threshold either passes it or has to be set so high it rejects short real postings. After tag-stripping the same shell is near-zero visible words, which is the property that actually separates "JS shell" from "short JD".

The stripped text is also what gets written back to `$jd_file`, so a page that *does* pass the threshold reaches the worker as prose rather than markup. That is the second half of the danger you described: not only "did we fetch something", but "is what we hand the worker evaluable at all".

**Two guards on the number itself**, because a bash comparison against a non-integer aborts `process_offer()` under `set -e`: the node call falls back to `0` on any throw, and the result is stripped to digits and defaulted (`:759–:760`, `:809–:810`).

Coverage is in `tests/batch-runner-jd-prefetch.test.mjs` (56 assertions), including that the threshold lives in a named variable rather than a bare literal, and that `curl --fail` discards an HTTP error body instead of passing it on as a JD.

One thing I would not claim: I have not compared the two implementations line by line, only answered for mine. If @luochen211's handles a case mine misses, take theirs — the fix matters more than whose it is, and 14 hours is not a claim to anything.

santifer:
Merged, @Zoubeir23 🚀 The thin-page answer settled it: counting visible words after the strip and truncating to 0 so the existing tested empty-file branch fires is the right shape (a second thin-path beside it would have been the mistake). The pre-connect destination guard and --max-filesize show the same care as your consider handshake. @luochen211's #2858 lands on top of this next.

PR #2818, opened 2026-08-13
fix(providers): route seven RSS decoders through the shared safe entity decoder

### LINKED ISSUE #2790: providers: seven RSS entity decoders emit NUL/lone surrogates into job titles, and delete out-of-range entities
### Code of Conduct

### Existing issues

### What happened?

Seven RSS/XML providers carry a hand-rolled `decodeXmlEntities` whose only numeric guard is a `try/catch` around `String.fromCodePoint`. `fromCodePoint` **does not throw** for NUL, the C0 controls, lone surrogates, or the noncharacters — so all of them decode and travel into the job title.

That is precisely the failure `providers/_html-entities.mjs` was centralized to prevent. Its own docstring (L36-43) argues the case better than I can:

> A decoded title is not displayed and discarded: it is written to the scan history, the pipeline, the tracker, and every document generated downstream. A NUL or a lone surrogate entering there is not a rendering artefact — it truncates C-string-backed consumers, produces ill-formed UTF-8 on serialization, and is tedious to trace back to one malformed entity in one posting weeks later. The feed host controls this input, so "no legitimate page does that" is not a guarantee.

Thirteen providers already import that shared decoder. These seven still don't:

| file | `fromCodePoint` helper | `decodeXmlEntities` |
|---|---|---|
| `providers/jobspresso.mjs` | L57 | L68 |
| `providers/higheredjobs.mjs` | L49 | L60 |
| `providers/nodesk.mjs` | L57 | L68 |
| `providers/larajobs.mjs` | L61 | L72 |
| `providers/personio.mjs` | L90 | L101 |
| `providers/teamtailor.mjs` | L96 | L107 |
| `providers/weworkremotely.mjs` | L57 | L68 |

All seven are byte-identical in behaviour:

```js
function fromCodePoint(cp) {
  try {
    return String.fromCodePoint(cp);
  } catch {
    return "";           // out-of-range entities are DELETED, not preserved
  }
}
```

There are two distinct defects here:

1. **Illegal code points are emitted.** The `try/catch` only catches the `RangeError` for `cp > 0x10FFFF`. `&#0;`, `&#1;`, `&#xD800;` and `&#xFFFF;` all decode successfully into the title.
2. **Out-of-range entities are silently deleted** (`return ""`), where the shared decoder — and the explicit reasoning in `_html-entities.mjs` L73-74 — leaves the original text: *"Raw `&#0;` in a title is visible and inert; a decoded NUL is neither."* Deleting is the worse failure of the two: it destroys source text with no trace.

### Steps to reproduce

Any of the seven; `jobspresso` shown because `parseJobspressoFeed` is exported.

```js
// repro.mjs — run from the repo root: node repro.mjs
import { parseJobspressoFeed } from './providers/jobspresso.mjs';
import { decodeEntities } from './providers/_html-entities.mjs';

const feed = (title) => `<?xml version="1.0"?><rss><channel>
  <item><title>${title}</title><link>https://jobspresso.co/job/x/</link><description>d</description></item>
</channel></rss>`;

for (const raw of ['A&#0;B', 'A&#xD800;B', 'A&#1;B', 'A&#xFFFF;B', 'A&#99999999;B']) {
  console.log(
    raw.padEnd(16),
    'local:', JSON.stringify(parseJobspressoFeed(feed(raw))[0].title).padEnd(18),
    'shared:', JSON.stringify(decodeEntities(raw)),
  );
}
```

Observed output on `main` @ `fd48007` (the `local:` column is `JSON.stringify` of the parsed title, so the escapes are real characters in the string):

```
A&#0;B           local: "AB"         shared: "A&#0;B"
A&#xD800;B       local: "A\ud800B"         shared: "A&#xD800;B"
A&#1;B           local: "AB"         shared: "A&#1;B"
A&#xFFFF;B       local: "A￿B"         shared: "A&#xFFFF;B"
A&#99999999;B    local: "AB"               shared: "A&#99999999;B"
```

(The `A￿B` row prints as the literal noncharacter; it is shown escaped here for legibility.)

Row 1 is a NUL and row 2 an unpaired surrogate, both now inside a `job.title` on its way to `data/scan-history.tsv` and the tracker. Row 5 is the silent deletion.

### Expected behavior

All seven decode through the shared `decodeEntities` from `providers/_html-entities.mjs`, so the XML 1.0 §2.2 `Char` guard (`isEmittableCodePoint`, L60-65) applies uniformly and illegal references are left as written.

The shared decoder is a strict superset of what these seven do today:

- the same five named entities, plus `&nbsp;`, and matched case-insensitively
- a single-pass replace, so the `&amp;lt;` → `&lt;` no-over-decoding property each local copy documents is preserved
- separated hex/decimal alternatives, so `&#1a2;` cannot partially parse

One behaviour change worth calling out explicitly: `&nbsp;` currently survives as the literal text `&nbsp;` in these seven and would start decoding to a space. That reads as a fix rather than a regression, but it is a change, so it should be a conscious call rather than a side effect.

### Scope note

`providers/radancy.mjs` (L71-83) and `providers/softgarden.mjs` (L32-40) have the same *class* of divergence but are milder — they preserve the original text instead of deleting it, and radancy also rejects lone surrogates. Both still admit NUL and the C0 controls, since `code >= 0` is not the same test as XML §2.2. `providers/jobvite.mjs` (L372-395) is behaviourally correct — it is where `isEmittableCodePoint` came from — but is still a third copy of the guard.

I've deliberately left all three out of scope: radancy and softgarden currently have open PRs touching them, and folding jobvite in would turn a bug fix into a refactor. Happy to follow up on them separately once those PRs land, if that's useful.

### Environment

- career-ops version: `main` @ `fd48007`
- Node.js: v22
- OS: macOS

### Additional context

This is the fourth time the numeric guard has drifted between copies of this decoder — #1555 (deutschebahn/hecklerkoch), #1639 (successfactors/dassault/softgarden/rheinmetall), #2623 (jobvite), and #2146/#2150 (dassault/avature/rheinmetall/successfactors). Each round migrated the providers that were in front of us at the time; these seven were never in scope for any of them.

Happy to send the PR if you agree with the direction.

gunjanjaswal:
Fixes #2790.

## What

Seven RSS/XML providers (`jobspresso`, `higheredjobs`, `nodesk`, `larajobs`, `personio`, `teamtailor`, `weworkremotely`) each carried a byte-identical hand-rolled `decodeXmlEntities`. Its `fromCodePoint` only `try/catch`'d the `RangeError` for `cp > 0x10FFFF`, and `String.fromCodePoint` doesn't throw for NUL, the C0 controls, lone surrogates, or the noncharacters — so all of those decoded straight into job titles. Out-of-range references were silently deleted (`return ''`).

That's exactly what `providers/_html-entities.mjs` was centralized to prevent, and 13 other providers already use it.

## Fix

Replace each local copy with the shared `decodeEntities`. It gates numeric references on the XML 1.0 §2.2 set and leaves anything non-emittable as raw text — `&#0;` stays visible and inert rather than becoming a NUL, or vanishing. No behaviour change for legitimate titles; it also picks up `&nbsp;` since the shared named-entity table includes it. Net −164/+69, mostly deleting the duplicated decoder.

## Evidence

Before, against `providers/jobspresso.mjs` on `main` (shown as `JSON.stringify` of the parsed title):

```
"A&#0;B"        ->  "A B"    (NUL emitted)
"A&#xD800;B"    ->  "A\uD800B"    (lone surrogate emitted)
"A&#99999999;B" ->  "AB"          (out-of-range deleted)
```

After the change, all three are left as raw text, matching the shared decoder.

New `tests/providers/rss-entity-decoding.test.mjs` drives the six RSS feed parsers with `&#0;`, `&#xD800;`, `&#1;`, `&#xFFFF;` and an out-of-range reference, asserting the parsed title matches `decodeEntities` and contains no NUL or lone surrogate. Reverting any one provider fails only that provider's case, and the existing provider tests still pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RSS feed text decoding across supported job providers.
  * Prevented invalid characters, including NULs, lone surrogates, and out-of-range numeric entities, from appearing in job titles and other feed content.
  * Standardized handling of XML and HTML entities for more consistent job listing data.
* **Tests**
  * Added coverage for malformed and unsafe entity references across multiple RSS providers.

### DIFF
--- a/providers/higheredjobs.mjs
+++ b/providers/higheredjobs.mjs
@@ -1,3 +1,4 @@
+import { decodeEntities } from './_html-entities.mjs';
 // @ts-check
 /** @typedef {import('./_types.js').Provider} Provider */
 
@@ -46,33 +47,11 @@ export default {
   },
 };
 
-function fromCodePoint(cp) {
-  try {
-    return String.fromCodePoint(cp);
-  } catch {
-    return '';
-  }
-}
-
-// Decode the XML entities that appear in RSS text: numeric (&#38; / &#x27;)
-// and the named five. Numeric forms are decoded first; &amp; is decoded LAST
-// so a literal "&amp;lt;" yields "&lt;" rather than over-decoding to "<".
-function decodeXmlEntities(s) {
-  return s
-    .replace(/&#x([0-9a-fA-F]+);/g, (_, h) => fromCodePoint(parseInt(h, 16)))
-    .replace(/&#(\d+);/g, (_, d) => fromCodePoint(parseInt(d, 10)))
-    .replace(/&lt;/g, '<')
-    .replace(/&gt;/g, '>')
-    .replace(/&quot;/g, '"')
-    .replace(/&apos;/g, "'")
-    .replace(/&amp;/g, '&');
-}
-
 // Resolve a tag's inner text: unwrap a CDATA section, else decode entities.
 function extractText(inner) {
   const cdata = inner.match(/^\s*<!\[CDATA\[([\s\S]*?)\]\]>\s*$/);
   if (cdata) return cdata[1].trim();
-  return decodeXmlEntities(inner).trim();
+  return decodeEntities(inner).trim();
 }
 
 // Extract the text of the first <tag>...</tag> in a block. Returns '' when absent.
--- a/providers/jobspresso.mjs
+++ b/providers/jobspresso.mjs
@@ -1,3 +1,4 @@
+import { decodeEntities } from './_html-entities.mjs';
 // @ts-check
 /** @typedef {import('./_types.js').Provider} Provider */
 
@@ -53,34 +54,11 @@ export default {
   },
 };
 
-/** Code point → string; '' for out-of-range values instead of throwing. */
-function fromCodePoint(cp) {
-  try {
-    return String.fromCodePoint(cp);
-  } catch {
-    return "";
-  }
-}
-
-// Decode the XML entities that appear in RSS text: numeric (&#38; / &#x27;)
-// and the named five. Numeric forms are decoded first; &amp; is decoded LAST
-// so a literal "&amp;lt;" yields "&lt;" rather than over-decoding to "<".
-function decodeXmlEntities(s) {
-  return s
-    .replace(/&#x([0-9a-fA-F]+);/g, (_, h) => fromCodePoint(parseInt(h, 16)))
-    .replace(/&#(\d+);/g, (_, d) => fromCodePoint(parseInt(d, 10)))
-    .replace(/&lt;/g, "<")
-    .replace(/&gt;/g, ">")
-    .replace(/&quot;/g, '"')
-    .replace(/&apos;/g, "'")
-    .replace(/&amp;/g, "&");
-}
-
 // Resolve a tag's inner text: unwrap a CDATA section, else decode entities.
 function extractText(inner) {
   const cdata = inner.match(/^\s*<!\[CDATA\[([\s\S]*?)\]\]>\s*$/);
   if (cdata) return cdata[1].trim();
-  return decodeXmlEntities(inner).trim();
+  return decodeEntities(inner).trim();
 }
 
 // Extract the text of the first <tag>...</tag> in a block. Returns '' when absent.
--- a/providers/larajobs.mjs
+++ b/providers/larajobs.mjs
@@ -1,3 +1,4 @@
+import { decodeEntities } from './_html-entities.mjs';
 // @ts-check
 /** @typedef {import('./_types.js').Provider} Provider */
 
@@ -58,33 +59,11 @@ export default {
   },
 };
 
-function fromCodePoint(cp) {
-  try {
-    return String.fromCodePoint(cp);
-  } catch {
-    return '';
-  }
-}
-
-// Decode the XML entities that appear in RSS text: numeric (&#38; / &#x27;)
-// and the named five. Numeric forms are decoded first; &amp; is decoded LAST
-// so a literal "&amp;lt;" yields "&lt;" rather than over-decoding to "<".
-function decodeXmlEntities(s) {
-  return s
-    .replace(/&#x([0-9a-fA-F]+);/g, (_, h) => fromCodePoint(parseInt(h, 16)))
-    .replace(/&#(\d+);/g, (_, d) => fromCodePoint(parseInt(d, 10)))
-    .replace(/&lt;/g, '<')
-    .replace(/&gt;/g, '>')
-    .replace(/&quot;/g, '"')
-    .replace(/&apos;/g, "'")
-    .replace(/&amp;/g, '&');
-}
-
 // Resolve a tag's inner text: unwrap a CDATA section, else decode entities.
 function extractText(inner) {
   const cdata = inner.match(/^\s*<!\[CDATA\[([\s\S]*?)\]\]>\s*$/);
   if (cdata) return cdata[1].trim();
-  return decodeXmlEntities(inner).trim();
+  return decodeEntities(inner).trim();
 }
 
 // Extract the text of the first <tag>...</tag> in a block. Returns '' when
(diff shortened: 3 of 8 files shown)


gunjanjaswal:
Thanks — both addressed:

- The surrogate check is now a scan that rejects only NUL and *lone* surrogates, so a valid supplementary character (a high/low pair) no longer trips it.
- Added `personio` coverage for both paths: the XML feed (title/office decode via `tagText` → `extractText`) and the HTML fallback (`decodeEntities(stripTags(...))`), asserting invalid numeric references stay raw text with no bad code units.

Scott-Emberson (approved):
Approving from the code-owner side for `tests/`. Our owned surface here is the new `tests/providers/rss-entity-decoding.test.mjs` (+96/-0); this review speaks only to that file, not to the seven provider changes or the shared decoder, which sit with their own owners.

The test holds up. It loads the shared `decodeEntities` and drives eight parse paths (six RSS feed parsers plus `parsePersonioXml` and `parsePersonioHtml`) with five non-emittable references, asserting per case that the title carries no NUL or lone surrogate and equals the shared decoder's output exactly. I probed the decoder directly: illegal references stay literal, `&amp;` and `&nbsp;` decode correctly, so the equality check compares against real values rather than both sides collapsing to empty. Revert any one provider to its old hand-rolled decoder and `A&#0;B` decodes to a NUL, which both gates catch, and only that provider's case fails. The `hasNulOrLoneSurrogate` helper pairs high/low surrogates instead of using a blanket surrogate class, so legitimate emoji or CJK-B characters do not false-positive. Hermetic throughout, same helpers as the neighbouring provider tests.

The two CodeRabbit comments were against the first commit; the head commit "tighten surrogate check and cover personio (review)" already addresses both (the pair-walking surrogate check and the personio coverage), so neither carries forward.

Ran `node tests/providers/rss-entity-decoding.test.mjs` at head `554362e`: 8/8 pass. CI is green across the three-OS matrix, so I relied on that for the full suite rather than re-run it.

One optional nit, not blocking: every case is a non-emittable reference that the decoder leaves as raw text, so `title === decodeEntities(input)` holds equally for a provider that decodes safely and one that does no decoding at all. `jobspresso.test.mjs` and `nodesk.test.mjs` have no legitimate-entity assertion elsewhere, so a provider that stopped decoding would slip through. Adding one legal case such as `A&amp;B` to the `cases` array makes this file a positive control for all eight paths. Fine to land as-is.

santifer:
Seven copies of the same hand-rolled decoder, each with the same hole, is the exact shape `providers/_html-entities.mjs` was centralized to close. Its own docstring makes the argument, and this PR is that argument applied.

Verified rather than assumed, on both sides:

- `main`: **7** providers carry `function decodeXmlEntities` (higheredjobs, jobspresso, larajobs, nodesk, personio, teamtailor, weworkremotely)
- your branch: **0**, with 21 providers importing the shared module (control grep, so an empty result is a real zero and not a broken path)

I also checked what the central decoder actually does with the dangerous inputs, because "route it through the shared one" is only worth anything if the shared one is right:

| input | central decoder |
|---|---|
| `&#0;` NUL | left literal, **not decoded** |
| `&#1;` C0 control | left literal |
| `&#xD800;` lone surrogate | left literal |
| `&#xFFFE;` noncharacter | left literal |
| `&amp;` | `&` |
| `&#233;` | `é` |

That is the correct shape: refuse the ones that corrupt downstream consumers, decode the ordinary ones. `String.fromCodePoint` not throwing on any of the first four is precisely why a `try/catch` was never a guard.

All seven provider suites green (14, 7, 12, 8, 30, 18, 14 assertions, zero failures). Suite on a preview against today's main: 3860 passed, 0 failed. @Scott-Emberson approved the new test file as code owner for `tests/`.

Merged. 🚀

gunjanjaswal:
Thanks for the merge, and for the earlier review pass — glad to get the seven decoders onto the shared `decodeEntities`. Appreciated the quick turnaround.

PR #2820, opened 2026-08-13
Ignore CRLF/LF-only differences when detecting local system edits

### LINKED ISSUE #2817: update-system.mjs apply: incomplete .gitattributes renormalization in 80d104f9 makes #2448's divergence check flag ~150 untouched files, silently no-op'ing the whole update

**Code of Conduct**

**Existing issues**
- [x] I searched existing issues and this hasn't been reported yet — related precedent: #2337 (mine) and its fix, #2448, are the feature this bug lives in, but neither issue nor #2448's own PR discussion mentions this failure mode. #2421 is adjacent (USER_PATHS classification) but a different problem.

**What happened?**

`node update-system.mjs apply` reported that 150 `SYSTEM_PATHS` files — including files this install has *never* locally modified, e.g. `AGENTS.md` — "differ from upstream because THIS install changed them," and skipped updating all of them, writing a `.bak` for each. `VERSION` itself was one of the 150, so the update was a silent no-op: it committed a "chore: auto-update system files to v1.25.0" message while `VERSION` stayed at `1.25.0` (upstream was `1.26.0`). No error was raised; the run reported success.

Root cause: `locallyModifiedSystemFiles()` (added in #2448, closing #2337) flags a path when it differs from both `merge-base(HEAD, FETCH_HEAD)` and `FETCH_HEAD`. That's the right check for a genuinely locally-edited file. But commit `80d104f9` ("fix(repo): add `.gitattributes` so Windows checkouts stop producing whole-file diffs", 2026-08-07) only renormalized the one file it was explicitly fixing (`providers/jobspresso.mjs`) — it never ran `git add --renormalize .` across the rest of the tree. Every other tracked text file's blob was left exactly as it was pre-`.gitattributes`.

`merge-base(HEAD, FETCH_HEAD)` on any install that hasn't synced since before 2026-08-07 predates `80d104f9`, i.e. predates `.gitattributes` entirely. Diffing across that boundary for a file that was never renormalized reproduces the same class of line-ending mismatch `.gitattributes` was meant to eliminate — except now it's `locallyModifiedSystemFiles()`'s own comparison that trips over it, not a Windows checkout. That's why the false positives line up almost exactly with "everything except `providers/jobspresso.mjs` and files `update-system.mjs` itself force-checks-out early in its self-reexec bootstrap."

**Impact**

The false-positive detection is the bug; the following is why it's worse in practice than a typical false positive. #2448 ships a global, binary remediation: keep everything flagged, or `--force` and overwrite everything flagged — no per-file selection (deliberate per #2770's description, not itself a bug). That design is fine when the flagged set is small and accurate — a user with one or two genuine local fixes loses nothing by choosing "keep." But this bug inflates the flagged set to ~150 files, mixing every false positive in with any real local fixes an install happens to have. Recovering `VERSION` (and any other real upstream change) then requires `--force`, which overwrites the genuine fixes too, with no way to take the false positives' upstream content while protecting the real ones. So the practical effect on an install that both (a) is affected by this bug and (b) has real, un-upstreamed local fixes to system files, is a full, all-or-nothing rollback of everything upstream shipped since the last sync, or a full loss of local fixes — there's no middle path until this is fixed.

**Steps to reproduce**

1. Have a local install last synced to a commit before `80d104f9` (2026-08-07) — e.g. anything on `career-ops-v1.25.0` or earlier that hasn't run `apply` since.
2. Confirm no genuine local edits to system files (optional, but confirms the false-positive: e.g. `diff <(git show HEAD~1:AGENTS.md) AGENTS.md` is empty).
3. Run `node update-system.mjs check` — reports `update-available`, `1.25.0` → `1.26.0`.
4. Run `node update-system.mjs apply`.
5. Observe: a long list of "N system file(s) differ from upstream because THIS install changed them," including files never edited locally. `VERSION` is one of them. The run reports "Update complete: v1.25.0 → v1.25.0" — same version in and out — with no error.

**Expected behavior**

A file that was never locally modified should never appear in the "kept because you changed it" list, and `VERSION` in particular should never be treated as a preservable local edit — it should always take the upstream value on a real update. The fix belongs in `80d104f9`'s scope (a follow-up renormalization commit running `git add --renormalize .` across the full tree would retroactively resolve this for every existing install once merged), and/or `locallyModifiedSystemFiles()` could compare with `--ignore-space-at-eol` or otherwise normalize before diffing so a `.gitattributes`-introduction boundary doesn't get misread as N separate local edits.

**CLI tool**
Claude Code

**OS and Node.js version**
macOS 26.6 / Node v26.0.0

**Output of npm run doctor**
```json
{"onboardingNeeded":false,"missing":[],"warnings":["Playwright MCP tools not detected (active CLI: claude)\n→ No project-level MCP config was detected for claude.\n  Add the Playwright MCP server to your claude config."],"autoCopied":[],"plugins":[{"id":"apify","hooks":["provider"],"enabled":false,"missingEnv":["APIFY_TOKEN"]},{"id":"gmail","hooks":["ingest"],"enabled":false,"missingEnv":["GMAIL_CLIENT_ID","GMAIL_CLIENT_SECRET","GMAIL_REFRESH_TOKEN"]},{"id":"notion","hooks":["export","search"],"enabled":false,"missingEnv":["NOTION_ACCESS_TOKEN","NOTION_PARENT_PAGE_ID"]}],"playwright_mcp":{"claude":false},"active_cli":"claude","cli_source":"default"}
```

gunjanjaswal:
Fixes #2817.

## What

`locallyModifiedSystemFiles()` decides which synced files a user has edited by diffing the working tree against the sync ref. That diff was byte-exact on line endings, so any file that differs from the merge-base only by CRLF vs LF counted as a local edit.

Installs that last synced before `.gitattributes` landed (80d104f9) have a merge-base whose blobs predate line-ending normalization. Against that baseline every synced file looks CRLF-vs-LF changed at once — the whole tree reads as "user-modified", the divergence guard from #2448 trips, and the update quietly no-ops instead of writing anything. From the user's side the updater just stops working with no error.

## Fix

Pass `--ignore-cr-at-eol` to that diff. A pure CRLF/LF difference no longer registers as a local edit. A real content change, or a genuine trailing-whitespace edit, still does — the flag only forgives the carriage return at end of line, nothing else.

One line, plus a comment pointing back at 80d104f9 so the reason survives.

## Evidence

New regression test in `tests/updater-local-system-edits.test.mjs` (case 13) builds a repo with a CRLF baseline, an untouched file renormalized to LF, and one genuinely edited file, then asserts only the real edit comes back.

With the fix, all 15 checks in that file pass. Revert the one-line change and case 13 fails as expected — the LF-only file is wrongly reported alongside the real edit:

```
❌ #13 expected ['generate-cover-letter.mjs'], got ["generate-cover-letter.mjs","modes/pdf.md"]
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Line-ending-only changes are no longer incorrectly reported as local edits.
  - Genuine content changes and trailing-whitespace edits continue to be detected correctly.

- **Tests**
  - Added coverage to verify correct handling of CRLF/LF differences alongside genuine local modifications.

### DIFF
--- a/tests/updater-local-system-edits.test.mjs
+++ b/tests/updater-local-system-edits.test.mjs
@@ -320,3 +320,46 @@ const PATHS = ['modes/', 'generate-cover-letter.mjs'];
     fail(`#12 real local edit lost from atRisk: ${JSON.stringify(atRisk)}`);
   }
 }
+
+// ── 13. A file differing from the baseline ONLY by CRLF/LF is not a local edit ──
+//    Reproduces #2817: installs synced before `.gitattributes` (80d104f9) have a
+//    merge-base whose text blobs predate line-ending normalization, so an
+//    untouched file reads as a local edit — inflating the flagged set to ~150
+//    files and silently no-op'ing the whole update. The CR-only difference must
+//    be ignored; a genuine content edit must still be reported.
+{
+  const dir = mkdtempSync(join(tmpdir(), 'co-crlf-'));
+  const g = (...args) => gitIn(dir, ...args);
+  g('init', '-q', '-b', 'main', '.');
+  g('config', 'user.email', 'test@example.com');
+  g('config', 'user.name', 'Test');
+  g('config', 'commit.gpgsign', 'false');
+  g('config', 'core.hooksPath', join(dir, 'no-such-hooks'));
+  g('config', 'core.autocrlf', 'false');
+  g('config', 'core.eol', 'lf');
+  mkdirSync(join(dir, 'modes'), { recursive: true });
+  // Base blobs committed with CRLF, standing in for a pre-`.gitattributes` tree.
+  writeFileSync(join(dir, 'modes', 'pdf.md'), 'shipped pdf\r\nline two\r\n');
+  writeFileSync(join(dir, 'generate-cover-letter.mjs'), 'shipped script\n');
+  g('add', '-A');
+  g('commit', '-qm', 'base (CRLF blobs)');
+  g('branch', 'upstream');
+  // Upstream changes the content of both files.
+  g('checkout', '-q', 'upstream');
+  writeFileSync(join(dir, 'modes', 'pdf.md'), 'shipped pdf v2\r\nline two\r\n');
+  writeFileSync(join(dir, 'generate-cover-letter.mjs'), 'shipped script v2\n');
+  g('commit', '-qam', 'upstream changes');
+  g('checkout', '-q', 'main');
+  // main: pdf.md renormalized to LF (a CR-only diff from the base); the script
+  // carries a genuine local edit.
+  writeFileSync(join(dir, 'modes', 'pdf.md'), 'shipped pdf\nline two\n');
+  writeFileSync(join(dir, 'generate-cover-letter.mjs'), 'local real fix\n');
+
+  const atRisk = locallyModifiedSystemFiles(PATHS, 'upstream', { git: g, root: dir });
+  if (atRisk.length === 1 && atRisk[0] === 'generate-cover-letter.mjs') {
+    pass('a CRLF/LF-only difference from the baseline is not a local edit (#2817)');
+  } else {
+    fail(`#13 expected ['generate-cover-letter.mjs'], got ${JSON.stringify(atRisk)}`);
+  }
+  rmSync(dir, { recursive: true, force: true });
+}
--- a/update-system.mjs
+++ b/update-system.mjs
@@ -750,7 +750,17 @@ export function locallyModifiedSystemFiles(paths, upstreamRef = 'FETCH_HEAD', ct
       // files and silently no-ops the whole update (#2817). This ignores only
       // the carriage return at end of line, so a genuine trailing-whitespace or
       // content edit is still detected.
-      return runGit('diff', '--ignore-cr-at-eol', '--name-only', ref, '--', ...paths).split('\n').map((p) => p.trim()).filter(Boolean);
+      //
+      // `--numstat`, deliberately, NOT `--name-only`: `--name-only` can list a
+      // path on the blob-OID comparison alone, before the textual diff runs, so
+      // a CRLF/LF-only file survives `--ignore-cr-at-eol` and the guard leaks
+      // right back. `--numstat` forces the textual diff, so the ignore rule is
+      // actually applied and a CR-only file drops out of the output entirely.
+      // The path is field 3 (a binary file renders as `-\t-\tpath`, still field
+      // 3). Reads less obviously than `--name-only`; keep it as-is.
+      return runGit('diff', '--ignore-cr-at-eol', '--numstat', ref, '--', ...paths)
+        .split('\n').map((l) => l.trim()).filter(Boolean)
+        .map((l) => l.split('\t')[2]).filter(Boolean);
     } catch {
       // An unreadable ref (shallow clone, unrelated histories) must never abort
       // the update — it degrades the warning, not the checkout.
--- a/update-system.mjs
+++ b/update-system.mjs
@@ -742,7 +742,15 @@ export function locallyModifiedSystemFiles(paths, upstreamRef = 'FETCH_HEAD', ct
 
   const diffNames = (ref) => {
     try {
-      return runGit('diff', '--name-only', ref, '--', ...paths).split('\n').map((p) => p.trim()).filter(Boolean);
+      // `--ignore-cr-at-eol`: a file whose only difference is a CRLF/LF line
+      // ending must not read as a local edit. Installs that last synced before
+      // `.gitattributes` was introduced (80d104f9) have a merge-base predating
+      // it, so every text file not renormalized in that commit differs from the
+      // baseline by line endings alone — which otherwise flags ~150 untouched
+      // files and silently no-ops the whole update (#2817). This ignores only
+      // the carriage return at end of line, so a genuine trailing-whitespace or
+      // content edit is still detected.
+      return runGit('diff', '--ignore-cr-at-eol', '--name-only', ref, '--', ...paths).split('\n').map((p) => p.trim()).filter(Boolean);
     } catch {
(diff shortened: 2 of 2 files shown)


gunjanjaswal:
Heads up on the red X: the only failing job is `test (windows-latest)`, and it's the `agent-inbox-tests.mjs` concurrency case ("kept 29 of 30" on 30 parallel appends), not anything in this PR. macOS and Ubuntu are green, and the new CRLF/LF test passes on all three including Windows. Looks like the same flake the `fix/agent-inbox-windows-lock` branch is already chasing, so nothing to change here.

Scott-Emberson (approved):
Approving from the code-owner side for `tests/`. Our owned surface here is `tests/updater-local-system-edits.test.mjs` (+43/-0); this review speaks only to that file, not to the `update-system.mjs` change, which is the updater and sits with santifer.

The test holds up. Case 13 builds a repo with a CRLF baseline, one untouched file renormalized to LF, and one genuinely edited file, then asserts only the real edit is reported by `locallyModifiedSystemFiles()`. It has teeth: revert the one-line `--ignore-cr-at-eol` change and the case fails, reporting the LF-only file alongside the real edit, so a no-op fix cannot pass it. It uses throwaway repos and the same helpers as the neighbouring cases, and its failure message names the file that broke.

Ran the file at head `6d682a7`: 15 of 15 checks pass, including the new case.

One note for context, not a finding on this file: `test (windows-latest)` is red, but the failure is the unrelated `agent-inbox` concurrent-add flake, not this change (the CRLF/LF case passes on Windows). It is tracked and fixed separately in #2824 / #2825, so nothing here needs to change.

santifer:
Good diagnosis, right target, and **your own new test catches a problem with the fix** — which is the best possible outcome for a test you wrote in the same PR.

`#13` fails for me against today's `main`:

```
❌ #13 expected ['generate-cover-letter.mjs'], got ["generate-cover-letter.mjs","modes/pdf.md"]
```

Isolated before saying anything: your branch alone on top of `main` reproduces it, `main` alone is clean, and the other three PRs I merged tonight are not involved. It is not base drift.

## The cause: `--name-only` ignores the flag

I rebuilt your fixture by hand and asked git directly. On **git 2.49.0** (macOS):

```
A) diff --ignore-cr-at-eol --name-only $BASE   → generate-cover-letter.mjs modes/pdf.md
C) diff --name-only $BASE                      → generate-cover-letter.mjs modes/pdf.md
```

Identical. `--ignore-cr-at-eol` has **no effect on `--name-only`** here, and neither does `-w`. `--name-only` reports the paths whose blobs differ; the whitespace-comparison options change how a textual diff is *rendered*, not which paths are listed. Your CI ran on a runner where it happened to behave, which is why this got through green.

## The one-word fix, verified in the same fixture

`--numstat` **does** honour it:

```
--ignore-cr-at-eol --numstat $BASE   → 1  1  generate-cover-letter.mjs
--numstat $BASE                      → 1  1  generate-cover-letter.mjs
                                       2  2  modes/pdf.md
```

The CR-only file drops out entirely. So `diffNames()` wants `--numstat` and the third tab-separated field, rather than `--name-only`:

```js
return runGit('diff', '--ignore-cr-at-eol', '--numstat', ref, '--', ...paths)
  .split('\n').map((l) => l.trim()).filter(Boolean)
  .map((l) => l.split('\t')[2]).filter(Boolean);
```

Worth a comment at that line saying *why* it is `--numstat`, because `--name-only` is the obvious reading and the next person will "simplify" it back.

Two notes while you are in there. Binary files render as `-\t-\tpath` in `--numstat`, so the path is still field 3 and nothing breaks, but a rename shows as `old => new` in some configurations; `-z` avoids that class if you want the belt. And the underlying bug in #2817 is real and worth fixing — this is about the mechanism, not the goal.

Ping me when it is up and I will take it the same day. The failing test is already in your branch, so you will know the moment it works.

gunjanjaswal:
Done — switched to `--numstat` and pull the path from field 3, with a comment spelling out why it isn't `--name-only` so nobody simplifies it back. Your read is right: `--name-only` can list on the blob-OID compare before the textual diff runs, so a CR-only file slips past `--ignore-cr-at-eol`; `--numstat` forces the diff and it drops out. The 15-case suite (incl. the CRLF case) still passes.

On the rename `old => new` wart — I left it as your primary snippet for now rather than reaching for `-z`, since the `-z` rename format needs its own parse. Happy to add `-z` (or `--no-renames`, which flags a renamed system file under both names and keeps field 3 a clean path) if you'd prefer the belt. It's up whenever you've got a minute.

santifer:
The comment explaining why it's `--numstat` and not `--name-only` is the part that makes this mergeable: "`--name-only` can list a path on the blob-OID comparison alone, before the textual diff runs" is exactly the trap, and without that note the next person would "simplify" it straight back into the bug. Test #13 reproducing the pre-`.gitattributes` merge-base is the right shape too: it fails for the real reason, not for a mocked one.

Local run before merging: 3694 passed, 0 failed. Merged 🚀

One thing I found while checking your `--numstat` reasoning, and it's a genuine trade you inherited rather than something you broke — `--numstat` reports renames as a composite path:

```
$ git diff --name-only upstream -- modes/
modes/viejo.md
$ git diff --ignore-cr-at-eol --numstat upstream -- modes/
0	0	modes/{nuevo.md => viejo.md}
```

So field 3 becomes `modes/{nuevo.md => viejo.md}`, a path that doesn't exist, and it'd surface as a phantom at-risk file the day an upstream rename lands in the system layer. `--name-only` didn't have this because it prints only the destination. Rare enough that it wasn't worth holding your fix for, given yours kills a bug that no-ops the entire update for anyone who synced before `80d104f9`.

`--no-renames` closes it, and it needs a fixture to pin it. Yours if you want it.

gunjanjaswal:
Thanks for the merge and the sharp review — the `--numstat` catch was the right call. Appreciate the same-day turnaround.

PR #2821, opened 2026-08-13
Add a global "Back to Top" button to the web app

### LINKED ISSUE #2807: Add a Global "Back to Top" Button Across the Application
### Code of Conduct

### Existing issues

### Problem

The application currently lacks a "Back to Top" button on pages with long, scrollable content.

As users navigate through lengthy pages, they must manually scroll back to the top, which reduces usability and creates unnecessary friction, especially on mobile devices. A consistent back-to-top mechanism would improve navigation and provide a better overall user experience

### Proposed solution

Implement a reusable floating "Back to Top" component that is available across the application.

Suggested behavior:
- Display the button after the user has scrolled approximately 300–500px.
- Smoothly scroll the page back to the top when clicked.
- Position the button in the bottom-right corner without obstructing existing UI elements.
- Support responsive layouts (desktop, tablet, and mobile).
- Ensure accessibility with keyboard navigation, focus indicators, and an appropriate `aria-label`.
- Follow the application's existing design system and theme.

### Area

Dashboard TUI

gunjanjaswal:
Closes #2807.

## What

Adds the floating "Back to Top" control from the issue. It's mounted once in the app shell, so it's available on every page. It fades in after ~400px of scroll and returns the window to the top when clicked.

The issue's Area is tagged "Dashboard TUI", but the described behavior (floating button, bottom-right, `aria-label`, responsive) only maps to the `web/` Next.js app — the Go TUI has no equivalent — so that's where this lives. Happy to move it if that read is wrong.

## How it fits the app

- **Scroll target.** `main` has no own `overflow-y`, so the window is the scroll container. The button listens on window scroll and calls `window.scrollTo` — no scroll-container plumbing needed.
- **Placement.** The assistant launcher already owns `bottom-5 right-5 z-50`. The button sits one row above it (`bottom-20 right-5`) and one layer below (`z-40`), so the two never overlap and an open assistant panel covers it cleanly instead of fighting it.
- **Motion.** Honors `prefers-reduced-motion` — an instant jump when the user asked to reduce motion, a smooth glide otherwise.
- **Accessibility.** While hidden it's removed from the tab order (`tabIndex=-1`) and hidden from assistive tech (`aria-hidden`), so there's nothing to land on until it's usable. Visible, it's a native `<button type="button">` with an `aria-label` and the app's standard `focus-visible:ring-brand/50`.
- **Design system.** Reuses the existing tokens (`border-border`, `bg-surface/90`, `backdrop-blur`, `shadow-lg`, `hover:bg-surface-hover`) and the `ArrowUp` lucide icon, matching the assistant launcher's visual language.

## Testing

The show/hide threshold and the reduced-motion decision are pure functions in a dependency-free `src/lib/scroll-to-top.mjs`, unit-tested with `node --test` like the rest of `src/lib` (the `.tsx` is a thin shell over them):

```
✔ hidden at the very top of the page
✔ hidden while still within the threshold
✔ still hidden exactly at the threshold
✔ shown once scrolled past the threshold
✔ a custom threshold is respected
✔ reduced motion jumps instantly, otherwise it glides
```

`npm test` (139 pass) and `npm run typecheck` (`tsc --noEmit`, clean) both green.

Verified end to end in `next dev`: at the top the button is `aria-hidden`, `opacity:0`, `pointer-events:none`, `tabIndex=-1`; scrolling past ~400px flips it to `aria-hidden=false`, `opacity:1`, `pointer-events:auto`, `tabIndex=0`, rendered just above the assistant launcher without overlap; clicking scrolls to the top and it hides itself again.

## Summary by CodeRabbit

* **New Features**
  * Added a floating “Back to top” button that appears after scrolling down.
  * Supports smooth scrolling while respecting reduced-motion preferences.
  * Includes accessible labeling and keyboard-friendly behavior.

* **Tests**
  * Added coverage for button visibility thresholds and scrolling behavior.


santifer:
Routing note, @gunjanjaswal — labelled `area:web`, so this goes to the maintainer who owns the `web/` subsystem rather than sitting in my queue. They triage and merge that area themselves and are active daily.

Worth flagging one thing so you don't get a surprise: **#2807 (@Yassiinee) asks for the same feature**, opened around the same time. Two implementations of one back-to-top button is wasted effort for one of you, and neither of you could have known. That's a coordination gap on my side — I should have linked them when the second one appeared. Whoever picks it up will sort out which lands; I've cross-referenced both.

gunjanjaswal:
No worries at all on the coordination — these things happen when two people reach for the same issue at once. Genuinely happy to defer to @Yassiinee's if it's further along or the web maintainer prefers it; whichever lands, the button gets built. Thanks for cross-referencing them.

santifer:
Six days on a reply that generous is not okay, @gunjanjaswal, and "whichever lands, the button gets built" is exactly the attitude that makes a project work. Sorry.

The decision you're waiting for is the web agent's, not mine — `web/` is built and reviewed separately and I route rather than triage there. It's labelled `area:web` and I've just handed the whole web batch across with your PR named in it, so the answer comes from someone who can actually make it.

What I can tell you is that nothing is blocked on you, and deferring to @Yassiinee's isn't something you need to volunteer twice: it's on the record here, and if theirs lands first this thread says plainly that you offered.

gunjanjaswal:
No apology needed, really. Six days is nothing, and sending it to the folks who actually own web/ is the right move. I'm happy whichever way this lands. If @Yassiinee's is further along, theirs should go in; if the web agent wants mine, it's ready to go. Either way the button ships, which is the part I care about. Thanks for cross-linking the two.

santifer:
Another week and still no verdict, @gunjanjaswal. You were gracious about the first six days; the seven since are mine to own, not yours to be patient about.

Concretely: I'm putting your PR and the other one in front of the web agent tonight as a pair, so the call gets made by comparing them rather than by whichever gets opened first. You'll have an answer in this thread by Wednesday either way. Nothing for you to do meanwhile: it's green and it's ready.

santifer:
Merged. Seventeen days is far too long for a PR this finished, and the delay was entirely on our side of the queue.

I want to correct the framing this arrived with, because it matters for you and for @Yassiinee. It reached me described as a competing implementation — an A/B call. It isn't. **#2807 is an issue, not a PR**, and there's no PR from @Yassiinee in any state; I checked before deciding, rather than choosing between two things when only one existed. You both reached for it on **the same day**, 13 August, exactly as you said in your first comment.

You offered twice to step aside. That was gracious and it turned out not to be the question.

What I checked in the code, rather than taking the description's word:

- **The logic is out of the component.** `scroll-to-top.mjs` with a test under `web/tests/lib/` is the pattern this repo uses, and it's why a scroll behaviour is verifiable at all instead of only being clickable.
- **The accessibility is the part most implementations skip.** `aria-hidden` plus `tabIndex={-1}` while hidden means there's nothing to land on with a keyboard until the button is actually usable. A visually-hidden-but-focusable control is a real trap and you closed it.
- `prefers-reduced-motion` respected, `size-11` = 44px so it meets the tap-target floor from #1542, passive listener, cleanup on unmount, and a sync on mount for a restored scroll position.

**And your two positioning premises hold — I verified both:**

```
assistant-console.tsx:484   launcher owns  bottom-5 right-5 z-50
this PR                                    bottom-20 right-5 z-40
app-shell.tsx:73            <main className="flex-1 overflow-x-hidden">  — no overflow-y
```

So the open assistant panel covers it cleanly rather than fighting it, and the window genuinely is the scroll container, which is what makes listening on `window` correct here. You looked at the layout instead of guessing at it, and that's the difference between a floating button and a floating button that collides with something three weeks later.

@Yassiinee proposed this and asked to build it, and the credit for the idea is theirs — I've said so on #2807.

gunjanjaswal:
Thank you, and thank you for reading the code rather than the framing it arrived in. The delay never bothered me, but I appreciate you setting the record straight on #2807. Yassiinee deserves the credit for the idea, they called it first.

On the accessibility bit: that focus trap is the thing I always check for on these floating controls. A button you can tab to but can't see is worse than no button. Glad the layering held up under a real look too. I'd rather verify the z-index and the scroll container up front than find out three weeks later that the assistant panel and the button are wrestling.

Really glad it's in. Appreciate the care you put into the review.

Yassiinee:
Late to this, I've been away and am just catching up. Nice work @gunjanjaswal, and thanks for the two offers to stand down; you didn't need to make them, and it was gracious that you did.

Having read the diff: pulling the threshold and motion decision out into `scroll-to-top.mjs` is the right call, and the `aria-hidden` + `tabIndex={-1}` while hidden is the part I'd have been most worried about getting wrong. Glad this is the version that shipped.

gunjanjaswal:
Thanks @Yassiinee, that means a lot coming from the person who called for it. The idea was yours; I just happened to have the time that week.

And yeah, the `aria-hidden` + `tabIndex={-1}` was the bit I kept coming back to. A control you can tab to but can't see is worse than not having one, so I'd rather over-index on getting that right. Glad we both landed on the same worry. Good to have it in either way.

PR #2825, opened 2026-08-13
fix(agent-inbox): give concurrent add() a burst-sized lock timeout

### LINKED ISSUE #2824: agent-inbox: concurrent add drops an item on Windows CI (lock-acquire timeout under 30-way contention)
## What

The "concurrent adds do not lose items" case in `agent-inbox-tests.mjs` (section 7) fails intermittently on `test (windows-latest)`, and only there. Seen on the CI run for the unrelated PR #2820:

```
❌ every concurrent add exited cleanly — 1 non-zero exits
❌ no item is duplicated or truncated — actual= (29 items)
Results: 16 passed, 3 failed
```

ubuntu and macos pass. It is unrelated to #2820's change (that PR only touches `update-system.mjs` and `tests/updater-local-system-edits.test.mjs`); the red just surfaced on that run.

## Root cause (confirmed, not hypothesised)

The test spawns 30 concurrent `node agent-inbox.mjs add` processes. Each appends one line under `withPipelineLock` (`agent-inbox.mjs:149`) with no options, so it inherits the shared `DEFAULT_TIMEOUT_MS = 8000` / `retryMs = 80` from `pipeline-lock.mjs`. Lock acquisition is a retry lottery, not a fair queue: serving N herded waiters is coupon-collector, about N·H(N) rounds, so 30 adds need roughly 120 rounds while `8000 / 80 = 100` only affords about 100. On the slow, contended Windows runner one waiter never wins inside 8s, throws `LockTimeoutError`, exits non-zero, and its item is never appended, giving `kept=29/30` with one non-zero exit.

Reproduced deterministically by shrinking the timeout with the existing env lever (`CAREER_OPS_PIPELINE_LOCK_TIMEOUT_MS`) at N=30: at 200 ms, one item drops on roughly half of runs; at 100 ms, more. The failing child's captured stderr is a clean lock timeout, no filesystem error, so the Windows EPERM/EBUSY path is ruled out:

```
LockTimeoutError: pipeline lock timeout: ...\agent-inbox.md.lock held > 100ms
    at acquirePipelineLock (pipeline-lock.mjs:165)
```

The retry jitter added for #2777 is already present in `pipeline-lock.mjs` and cuts collisions, but it is documented there as "not a cure"; the remaining gap is the acquisition-timeout budget, not missing jitter.

## Fix direction

Give the inbox add path a lock timeout sized for its bursty concurrent-writer use, contained at the `agent-inbox.mjs` call site. The critical section is a single sub-millisecond append, so a larger budget only ever spends time that would otherwise be a lost item, and it overrides the shared default without touching any other lock caller. The test must not be weakened to pass.

A fix is ready and will reference this issue.

Scott-Emberson:
Fixes #2824.

## What

`agent-inbox-tests.mjs` section 7 ("concurrent adds do not lose items") flakes on `test (windows-latest)`, and only there: one of 30 concurrent `add` processes drops its item, giving `kept=29 of 30` with 1 non-zero exit. It surfaced on an unrelated PR's CI run (#2820), passes on ubuntu and macos, and is not caused by any single PR.

## Root cause

Each `add()` appends one line under `withPipelineLock` (`agent-inbox.mjs`) with no options, inheriting the shared 8s timeout / 80ms retry from `pipeline-lock.mjs`. Lock acquisition is a retry lottery, not a fair queue: serving N herded waiters is coupon-collector, about N·H(N) rounds, so 30 adds need roughly 120 rounds while 8000/80 only affords about 100. On the slow, contended Windows runner one waiter never wins inside 8s, throws `LockTimeoutError`, exits non-zero, and its item is never appended. The failing child's stderr is a clean lock timeout, no filesystem error:

```
LockTimeoutError: pipeline lock timeout: ...\agent-inbox.md.lock held > 100ms
    at acquirePipelineLock (pipeline-lock.mjs:165)
```

The retry jitter from #2777 is already present and cuts collisions, but is documented there as "not a cure"; the remaining gap is the acquisition-timeout budget.

## Fix

One line, contained to the `add()` call site: pass `{ timeoutMs: 30_000 }` so a burst-write queue gets a burst-sized budget (about 375 rounds of headroom). The critical section is a single sub-millisecond append, so the larger budget only ever spends time that would otherwise be a lost item. An explicit call-site `timeoutMs` overrides both the env lever and the shared default, so no other lock caller (scan-history, portal-health, tracker) is disturbed and `pipeline-lock.mjs` is untouched. The test is unchanged.

## Validation

- Reproduced RED with the existing `CAREER_OPS_PIPELINE_LOCK_TIMEOUT_MS` lever at N=30: 200ms drops one item on about half of runs, 100ms more. GREEN after the fix over 20 runs with the same lever: 0 failures (the call-site value overrides the lever).
- `node agent-inbox-tests.mjs` run 12 times: all green (19 passed, 0 failed each).
- Lock regression suites pass: `test/pipeline-lock.test.mjs`, `tests/scan-history-lock.test.mjs`, `tests/portal-health-lock.test.mjs`.
- `node test-all.mjs --quick`: 3691 passed, 0 failed (1 pre-existing unrelated warning).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when adding items to the queue by allowing the pipeline lock to wait up to 30 seconds before timing out.

### DIFF
--- a/agent-inbox.mjs
+++ b/agent-inbox.mjs
@@ -146,10 +146,23 @@ async function add() {
   // Checking the last byte still happens INSIDE the lock: it decides whether a
   // separating newline is needed, and reading it outside would race with
   // another writer's append between the check and the write.
+  //
+  // timeoutMs is raised from the shared 8s default because this queue's whole
+  // point is bursty concurrent writers (a dashboard, a script, cron all drop
+  // items at once), and lock acquisition is a retry lottery, not a fair queue.
+  // Serving N herded waiters is the coupon-collector problem: ~N·H(N) rounds,
+  // so 30 concurrent adds need ~120 rounds while 8000/80 = 100 only affords
+  // ~100. On the slow, contended windows-latest runner that shortfall makes one
+  // waiter time out and its item is LOST, the exact #2777 drop, reappearing as
+  // a loud LockTimeoutError instead of a silent overwrite. Jitter in
+  // pipeline-lock.mjs cuts the collision rate ~6x but is explicitly "not a
+  // cure"; the fit-for-purpose budget for a burst-write queue is the contained
+  // fix. 30s gives ~375 rounds of headroom, well past the herd's worst case,
+  // while the critical section itself is a single sub-millisecond append.
   await withPipelineLock(PATH, () => {
     const separator = needsLeadingNewline(PATH) ? '\n' : '';
     appendFileSync(PATH, `${separator}- [ ] ${stamp()} — ${text}\n`);
-  });
+  }, { timeoutMs: 30_000 });
   process.stdout.write(`Queued: ${text}\n`);
 }
 


Scott-Emberson:
@santifer this one is ready for your review when you have a moment, and it unblocks the intermittent `test (windows-latest)` failures.

The flake it fixes (#2824) is the agent-inbox concurrent-add race: under 30-way contention on the slow Windows runner, one `add` process cannot acquire the pipeline lock inside the shared 8s default, throws `LockTimeoutError`, and drops its item (`kept=29/30`, one non-zero exit). It is not tied to any one PR; it has been showing up red on unrelated PRs' Windows checks, for example #2820.

Root cause, reproduction, and a 20-run green under the deterministic timeout lever are in the PR body. Merging this removes that source of Windows CI flakiness, and it touches no other lock caller.

santifer:
Root-caused rather than hypothesised, and that is the whole difference: you drove the existing env lever down to 200ms and turned a Windows-only flake into something deterministic on any machine. I reproduced both sides before merging, N=30 at that budget: 4 of 6 runs lose an item on main, 0 of 6 on your branch.

One observation, not a blocker, and it is my follow-up rather than yours. `acquirePipelineLock` reads `options.timeoutMs ?? env`, so an explicit option now outranks `CAREER_OPS_PIPELINE_LOCK_TIMEOUT_MS` and that lever is inert for `add()` specifically: with the env at 1ms your branch still keeps 30/30. Nothing in the suite drives `add()` that way today, so nothing breaks, but an escape hatch the code can override is not much of an escape hatch. That belongs against `pipeline-lock.mjs` itself, which scan-history also uses, so it wants its own change and its own tests instead of riding on yours.

Merged. 🚀

santifer:
Follow-up worth having on the record, @Scott-Emberson: this fixed more than a flaky test.

After merging I swept every open PR whose only red check was `test (windows-latest)`. There were ten, and **nine of them were failing on this exact `kept=29 of 30`**, from nine different contributors, on branches that touch nothing near `agent-inbox`. One of them belongs to someone with eleven open PRs and no merges yet, who had every reason to read that X as their own mistake.

All nine re-run against the fixed base now.

So the framing in #2824, "it surfaced on an unrelated PR's CI run and is not caused by any single PR", turned out to understate it: it was surfacing on everyone's, and because it only ever hit one leg of the matrix it looked like a platform quirk rather than a shared blocker. Worth remembering as a shape: **a flake in a shared suite is not a test problem, it is a queue problem**, and it bills itself to whoever happens to push next.

santifer:
Correction to what I told you earlier, because it was wrong in a way that matters.

I said I had "re-run your job against the fixed base". **`gh run rerun` replays the same head SHA**, so those re-runs went back over the *old* merge commit, without #2825 in it. Verified rather than assumed: the run I re-ran carries `headSha=4721209b`, and `git merge-base --is-ancestor 513f07a2 4721209b` says the fix is not an ancestor of it.

Five of the ten came back green anyway, and that was luck, not the fix: the flake is probabilistic, so a re-run of the same base wins the lottery about half the time. Four stayed red for the same `kept=29 of 30`.

The right action was to give each branch an actually-newer base, so I have run `gh pr update-branch` on the ones still red. #2782 could not take it because it conflicts with `main`; @briacSck, that one needs a rebase from you.

**And the more important half.** Right after merging an unrelated one-line docs change, the same failure fired on `main` **with #2825 already in it**, on my own machine, under heavy parallel load. I then failed to reproduce it in 38 further attempts, including a replica that spawns the 30 processes exactly the way the test does and two full suites running alongside it.

So the honest state is: the burst-sized budget cut this a long way and **did not remove it**. That is what both #2825 and the jitter comment before it said would happen, in writing, and it is the reason the structural answer is a fair queue rather than a bigger retry lottery. #2777 stays open on the strength of that, and it is now the third time I would have been wrong to close it.

If you see this red on your PR again, it is still not yours.

Issue #1709, opened 2026-07-09
feat(eval): prompt-cache the static system prefix in the API eval engines (~12K tokens re-billed per offer)


abankar1:
## Problem

The four local evaluation engines each rebuild an **identical ~48KB static system prompt for every offer** and re-send it uncached:

- `openrouter-runner.mjs:317` (`buildSystemPrompt` = `_shared.md` + `_profile.md` + `oferta.md` + profile + cv)
- `openai-eval.mjs:222`, `gemini-eval.mjs:244`, `ollama-eval.mjs:212` — same shape.

Measured system-layer prefix (before cv/profile): `_shared.md` 21,978 B + `oferta.md` 25,797 B = **47,775 B ≈ ~12K tokens**, byte-identical across every offer in a run. A repo-wide grep for `cache_control` / `ephemeral` / `cachedContent` returns **zero hits** — nothing is cached.

For a 100-offer batch that is ~1.2M redundant **input** tokens re-billed, plus the latency of re-processing the same prefix each time. Token cost is the #1 pain in this tracker, and this is pure repeated spend.

## Proposed solution (opt-in, additive, per provider)

Mark the static prefix (everything above the per-offer JD) as cacheable. The prefix is already assembled first in all four engines, so this is a localized change with no reordering:

- **Anthropic / OpenRouter (`openrouter-runner.mjs`):** send the static block with `cache_control: { type: "ephemeral" }` (structured system content). ~90% input-cost reduction on cache hits within the 5-min TTL — ideal for back-to-back batch offers.
- **Gemini (`gemini-eval.mjs`):** use explicit context caching (`cachedContent`) for the `_shared.md` + `oferta.md` prefix.
- **OpenAI (`openai-eval.mjs`):** already benefits from automatic prefix caching as long as the static prefix stays first and stable (≥1024 tokens) — add a short doc note + a test guarding prefix stability so a future refactor doesn't silently break it.
- **Ollama (`ollama-eval.mjs`):** no API cost, so lower priority; note only.

Scoped strictly to the **API / BYOK / free-tier batch paths** — I'm *not* touching the `claude -p` CLI path, which already gets server-side prefix caching for free (identical 50KB prefix, 5-min TTL). This is complementary to #1025 (skip re-eval entirely) and #1552 (cheap pre-filter): it makes each eval that *does* run cheaper, and stacks with both.

## Why raise it separately

[...]

PR #2827, opened 2026-08-13
feat(openai-eval): add OrcaRouter as a named OpenAI-compatible provider

Marc-oss-hub:
## Add [OrcaRouter](https://www.orcarouter.ai) as a named OpenAI-compatible provider

This PR makes [OrcaRouter](https://www.orcarouter.ai) a first-class named provider for the headless OpenAI-compatible scripts (`openai-eval.mjs` for evaluation, `openai-tailor.mjs` for CV tailoring). Setting `ORCAROUTER_API_KEY` switches both scripts to the OrcaRouter endpoint (`https://api.orcarouter.ai/v1`) with model `orcarouter/auto` — a per-request virtual router that routes each call to the best model for the task.

It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

### What changes

- **`openai-eval.mjs`** and **`openai-tailor.mjs`**: when `ORCAROUTER_API_KEY` is set, the config block defaults to the OrcaRouter base URL + `orcarouter/auto` model. Explicit `OPENAI_*` env vars or `--url` / `--model` / `--key` flags still take precedence, so existing OpenAI-compatible setups are completely unaffected (behavior is byte-identical when no OrcaRouter env is set).
- **`.env.example`**: documents the new `ORCAROUTER_API_KEY` / `ORCAROUTER_BASE_URL` / `ORCAROUTER_MODEL` block.
- **`README.md`**: adds a "Standalone OpenAI-compatible eval" section with copy-paste OrcaRouter usage.
- Help text in both scripts now lists OrcaRouter under PROVIDER EXAMPLES and ENV.

### Why a named provider

The repo already supports OrcaRouter generically via `OPENAI_BASE_URL` passthrough, but a named provider gives users a single env var, correct out-of-the-box defaults, and discoverability — the same convenience the repo gives Gemini and OpenRouter. The host-gated prompt-cache wiring (#1709) is untouched: non-OpenAI hosts like OrcaRouter still get the ephemeral `cache_control` breakpoint, and `api.openai.com` keeps its plain-string system message.

### Verification

- `node --check` on both edited scripts.
- Source-pattern tests preserved: `test-all.mjs` 44c (openai-eval) and 44f (openai-tailor) regex checks all pass, as does the `evaluatorSources` report-allocator check.
- Live end-to-end run of `openai-eval.mjs` through OrcaRouter with `ORCAROUTER_API_KEY` set: full A–G evaluation returned (`model=orcarouter/auto`, provider `openai`).

> Note: on Windows the full local `test-all.mjs --quick` run stops at an unrelated, pre-existing `intake` symlink test (`EPERM: operation not permitted, symlink` — Windows lacks symlink privileges without developer mode/admin). It fails identically on clean `main`; CI runs on Linux where it passes.

I'm an engineer on the OrcaRouter team.

## Summary by CodeRabbit

- **New Features**
  - Added OrcaRouter support for OpenAI-compatible evaluation and CV tailoring.
  - Automatically uses OrcaRouter defaults when its API key is configured.
  - Preserves precedence for command-line options and explicit OpenAI settings.
  - Added help guidance, environment variable configuration, and usage examples.

- **Documentation**
  - Added setup instructions, dependency requirements, evaluation commands, and report-based CV tailoring guidance.
  - Documented provider defaults, configuration precedence, and OrcaRouter endpoint options.


santifer:
Thanks for the work, @Marc-oss-hub, and sorry for the slow first reply. I have to be direct about where this stands, because two written rules apply and one of them is about the README specifically.

**First, the capability already exists.** `openai-eval.mjs` documents it in its own header:

```
node openai-eval.mjs --url https://openrouter.ai/api/v1 --model <id> --file ./jds/job.txt
OPENAI_BASE_URL (or --url) — the provider's OpenAI-compatible base
```

Anyone can point either script at OrcaRouter today with `--url https://api.orcarouter.ai/v1`. So the PR doesn't add the ability to use it — it adds **named, defaulted, documented status** for one vendor among the many that endpoint already serves. That's a different thing, and it's the part the rules speak to.

**Second, CONTRIBUTING has a standing rule about the README** (written well before this PR):

> **PRs that add third-party hosted entry-points or service badges to the README** — links or embeds that route users' resumes or job data through a service the project doesn't operate. The README stays to assets the project controls.

And the README paragraph here goes further than a link:

> It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

That's product copy for a service, on the front page of a tool that doesn't use it for any of that. `openai-tailor.mjs` sends the user's **CV** through whichever endpoint is configured, so a defaulted named provider isn't a neutral convenience either — it's a default destination for personal data through a service the project doesn't run, which is the second rule ("local-first and zero-keys").

**I'm not closing it.** That's Santiago's call and I've made no recommendation to him beyond laying out the above. What I'd say honestly is that the README half has close to no chance, and that the `.env.example` half is a much narrower question — an env var documented alongside the generic `OPENAI_*` ones, without front-page placement or default-provider status, is a different proposal and I'd rather he judged that one than this one.

If you want to give it the best shot, dropping the README paragraph and the defaulting, and keeping only a commented example beside the existing generic ones, is the version I'd put in front of him.

santifer:
Thanks for this, @Marc-oss-hub, and sorry it sat for six days: that's on me, not on you.

The mechanism is fine and it's the direction the project is going. career-ops is deliberately AI-agnostic, `openai-eval.mjs` speaks to anything OpenAI-compatible, and giving a gateway sensible defaults so a user sets one key instead of three is a real convenience. `OPENAI_*` keeping precedence is the right call.

What I can't take is the framing around it, and I'd rather be concrete than vague.

**1. The security claims can't go in our README.** This sentence appears in three files:

> gateway-level, zero-trust security for AI agents — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes

I have no way to verify any of that, and neither does a reader. Our README is not the place for a vendor's own description of its product; if we print it, we are vouching for it. Describe what it is in a clause a reader can check ("an OpenAI-compatible gateway that routes per request") and let their site make its own case.

**2. That capability has a privacy consequence the section doesn't mention, and it's the one that matters here.** A gateway that screens every prompt and response sees every prompt and response. In career-ops those prompts contain the user's CV and the job descriptions they're evaluating, which is the most sensitive material this tool touches. That's a legitimate trade for someone who wants it, and it is not a footnote: any text recommending a hosted gateway has to say plainly that the CV goes through it. Right now the section reads as an unqualified upgrade.

**3. It shouldn't get a section above the generic mechanism.** A dedicated `### Standalone OpenAI-compatible eval` block that is really about one vendor puts a single commercial service ahead of the thing that actually makes us AI-agnostic. OpenRouter, Together, Groq, DeepSeek and the local servers are one list; add it to that list. The place where OpenRouter goes further, `openrouter-runner.mjs`, is a script that does something the generic path can't, which isn't the case here.

**The ask, concretely:** drop the marketing sentence from all three files, move the README mention into the existing provider list instead of its own section, and keep the env-var defaults and the `.env.example` entry, which are the genuinely useful part. If you want the privacy note in there too, one sentence is enough.

Happy to merge a version like that.

santifer:
Thanks for the work, @Marc-oss-hub, but this one doesn't land, and the reason is short: **it already works.**

`openai-eval.mjs` and `openai-tailor.mjs` speak to any OpenAI-compatible endpoint through two variables:

```bash
OPENAI_BASE_URL=https://api.orcarouter.ai/v1 \
OPENAI_API_KEY=sk-orca-... \
OPENAI_MODEL=orcarouter/auto \
  node openai-eval.mjs --file ./jds/job.txt
```

That's the whole integration, today, on `main`, with nothing merged. It's the same path OpenRouter, Together, Groq, DeepSeek, Zhipu, MiniMax, Fireworks, LM Studio, llama.cpp, vLLM and Ollama use — **no provider here gets named defaults or its own environment variables**, and adding them for one would leave the other eleven asking a question I couldn't answer.

**One thing worth taking back with you, because it will bite anyone following your integration docs.** The precedence in the PR doesn't do what its comment says:

```js
const useOrcaRouter = Boolean(process.env.ORCAROUTER_API_KEY);
let baseUrl = (process.env.OPENAI_BASE_URL || (useOrcaRouter ? 'https://api.orcarouter.ai/v1' : ...))
let apiKey  = process.env.OPENAI_API_KEY || process.env.ORCAROUTER_API_KEY || '';
```

A user with **both** keys set and no `OPENAI_BASE_URL` — the normal state for anyone using OpenAI directly — gets the OrcaRouter base URL paired with their **OpenAI** key. The comment says *"Explicit OPENAI_\* still take precedence, so this never changes existing behavior"*, and for the key alone that's true; the base URL falls through independently, so the two end up mismatched and the wrong secret goes over the wire.

Not merging it doesn't make that go away for you: the same shape in your own SDK or docs would send users' keys to the wrong host. Worth fixing wherever else it lives.

Closing here, but the endpoint works today and needs nothing from us.

PR #1916, opened 2026-07-15
fix: honor output language in headless evaluators

luochen211:
Closes #1897

## Summary
- parse `language.output` through one shared, fail-safe profile helper with `en` as the default
- inject the canonical output-language instruction into Ollama, OpenAI-compatible, Gemini, and OpenRouter prompts
- remove the Gemini rule that let JD language override the explicit profile preference
- include the helper in system updates and cover parsing, prompt content, and all four engine integrations

## Tests
- `node tests/output-language.test.mjs` (15 passed)
- `node validate-system-paths-coverage.mjs`
- `node test-all.mjs`

## Summary by CodeRabbit

- **New Features**
  - Added profile-driven “output language” enforcement for generated evaluation reports.
  - Gemini, Ollama, OpenAI, and OpenRouter now include the computed language directive in their prompts, with consistent output-language precedence.
- **Bug Fixes**
  - Improved report-saving flow: ensures the reports folder exists, updates tracker entries, runs optional merge only after successful save, and always releases reserved report numbers.
- **Tests**
  - Added end-to-end tests covering parsing, instruction generation, and prompt injection across evaluators; included in the full suite.
- **Chores**
  - Updated system update/rollback to include the new profile language helper.

PR #2830, opened 2026-08-13
fix(tests): stop one discovered suite decapitating the whole run

### LINKED ISSUE #2828: tests/intake.test.mjs aborts the whole suite on a machine without symlink privileges, hiding 56% of the checks
Spun out of #2815, where this got in the way of verifying a Windows fix.

## What happens

`tests/intake.test.mjs` calls `symlinkSync` unguarded (four times, first at [`tests/intake.test.mjs:187`](https://github.com/santifer/career-ops/blob/main/tests/intake.test.mjs#L187)). On Windows, creating a symlink needs `SeCreateSymbolicLinkPrivilege` — held by Administrators, or by everyone once Developer Mode is on. A standard non-elevated shell without Developer Mode, which is the **default Windows configuration**, gets `EPERM`:

```
node:fs:1877
Error: EPERM: operation not permitted, symlink '...\documents' -> '...\documents\cv\loop'
    at symlinkSync (node:fs:1877:11)
    at file:///.../tests/intake.test.mjs:187:5
```

`tests/intake.test.mjs` does not import `node:test`, so `runDiscovered()` imports it **in-process**. The uncaught throw takes the whole harness down with it.

## Why it matters

Not the one lost suite — everything after it. Measured on the same machine, same commit, once with the symlink calls locally neutralized and once without:

| | checks reported |
|---|---|
| as-is | **1626** |
| symlinks neutralized | **3693** |

**2067 checks — 56% of the suite — never run**, and because `finish()` never runs either, there is no `📊 Results` line at all. The output just stops. Everything from section ~30 onward is invisible, including `59b. Pipeline lock`, the `web/` suites, and every provider test.

The failure mode is worse than the count suggests: it does not look like a broken suite, it looks like a suite that ended. A contributor who does not know the run is truncated can read a screen of green and conclude their change is clean.

## This is the #1916 hole, one door over

`runDiscovered()` already defends against exactly this, and says so:

```js
// Discovered suites run IN-PROCESS and share this suite's counters. A
// process.exit() inside one would terminate test-all mid-run with a forged
// exit code — every later section (and finish()) would silently never run.
// Refuse to import such a suite and fail loudly instead (#1916 regression).
if (/\bprocess\.exit\s*\(/.test(src)) {
```

That guard catches `process.exit()`. An uncaught **throw** produces the identical outcome and walks straight past it.

## The repo already has the pattern

`test-all.mjs:12793` guards the same call for plugin-manifest traversal and degrades instead of dying:

```js
} catch (e) {
  warn(`symlink traversal test skipped: ${e.message}`);
}
```

That line is why the plugin-manifest symlink checks report `⚠️ symlink traversal test skipped: EPERM: ...` on this machine while the run carries on.

## Suggested fix

1. Guard the symlink setup in `tests/intake.test.mjs` the same way — `warn(...)` and skip those assertions when `symlinkSync` throws `EPERM`. This is not a coverage loss on CI, where `windows-latest` runs elevated and the tests run normally; it only degrades where the privilege genuinely is not there.
2. Optionally close the #1916 hole properly: wrap the in-process `await import(f)` in `runDiscovered()` in a try/catch that turns a throw into `fail(...)` and continues, so no single discovered suite can ever truncate the run again. That is the general fix; item 1 is the specific one.

I'd suggest both — 1 unblocks Windows contributors today, 2 stops the next unguarded call doing the same thing.

## Notes

- Not reproducible on CI: GitHub's `windows-latest` runner is elevated, so symlinks succeed and the suite runs to completion. This bites local contributors only, which is also why it has gone unnoticed.
- Enabling Developer Mode is a workaround, not a fix — a test suite should not require a privilege escalation to report its own results.
- Environment: Windows 11 Pro 26200, Node v24.18.0, non-elevated shell, Developer Mode off.

Happy to send the PR for either or both if you want it.

LayerC0de:
Fixes #2828.

`tests/intake.test.mjs` calls `symlinkSync` unguarded. On Windows that needs `SeCreateSymbolicLinkPrivilege` — held by Administrators, or by everyone once Developer Mode is on — so an ordinary non-elevated shell, **the default configuration**, gets `EPERM`. The suite doesn't import `node:test`, so `runDiscovered()` imports it in-process and the uncaught throw takes the harness down with it.

The cost isn't the one lost suite, it's everything after it. Same machine, same commit, measured with the symlink calls neutralized and without:

| | checks reported |
|---|---|
| as-is | **1626** |
| symlinks neutralized | **3693** |

2067 checks — 56% of the suite — never ran, and since `finish()` never ran either there was **no verdict line at all**. The output simply stopped. That's worse than a red suite: it doesn't look broken, it looks finished, so the screen of green above the cut reads as a pass.

## Two changes

**1. Guard the symlink setup** (`tests/intake.test.mjs`) — warn instead of throwing, exactly as the plugin-manifest traversal checks at `test-all.mjs:12793` already do:

```
⚠️  intake symlink tests skipped: no symlink privilege (EPERM) — enable Developer Mode or run elevated to exercise them
```

Only `EPERM` *from `symlink()` itself* is tolerated. Any other error — including an `EPERM` from a different syscall — still fails, so this can't quietly turn broken symlink handling into a skipped line. No coverage is lost on CI, which runs elevated and exercises all four assertions normally.

**2. Contain a throw from any discovered suite** (`test-all.mjs`) — this is what makes change 1 the last of its kind.

`runDiscovered()` already refuses suites that call `process.exit()` (#1916) and `finish()`. Both guard this same property, and both say so:

```js
// Discovered suites run IN-PROCESS and share this suite's counters. A
// process.exit() inside one would terminate test-all mid-run with a forged
// exit code — every later section (and finish()) would silently never run.
```

An uncaught throw produces the identical outcome and walks straight past both. It's now one failure — reported with its throw site, since a suite that dies mid-import leaves no other clue how far it got — and the run continues.

Enabling Developer Mode is a workaround, not a fix: a test suite shouldn't need a privilege escalation to report its own results.

## Verification

On the affected machine (Windows 11, non-elevated, Developer Mode off), the suite now runs to completion:

```
before   1626 checks, no verdict line, harness dead at intake
after    📊 Results: 3680 passed, 1 failed, 3 warnings
```

The one remaining failure is pre-existing and environmental — `followup-cadence.mjs` can't resolve `js-yaml` from its temp-dir copy — and is unrelated to this change.

## A note on testing the guard itself

I deliberately did **not** add a regression test for change 2. The only way to exercise `runDiscovered()` is to plant a throwing `*.test.mjs` under `tests/`, and discovery is recursive over the whole tree — so a fixture leaked by an interrupted run would break every subsequent invocation of the suite. That trade seemed clearly wrong for a guard whose two siblings are themselves untested static checks.

If you'd like it covered, the clean way is to make `TESTS_DIR` injectable (env var or parameter) so a test can point discovery at a throwaway directory. Happy to do that as a follow-up if you want it — it's a production-surface change, so I didn't fold it in unasked.

## Summary by CodeRabbit

* **Tests**
  * Test runs continue when an individual suite fails during loading, with concise error details included in the results.
  * Symlink-related tests on Windows now gracefully skip unsupported, non-elevated environments while still reporting unexpected errors.


Scott-Emberson (approved):
`tests/intake.test.mjs`: the EPERM guard is narrowly gated on `code === 'EPERM' && syscall === 'symlink'` with an `else throw`, so a genuine failure still surfaces and only the no-privilege setup path degrades to a warn. On a privileged host all four symlink assertions still run. Hermetic, cleans up in finally. Approving from the tests/ side.

santifer:
Before anything else: I opened #2867 doing the same thing as this PR, and I opened it AFTER yours. That's me not reading my own queue, and it's the second time today I've stepped on someone's work. Yours is the one that lands; I'm closing mine.

It should be yours anyway, because it does the half mine didn't: fixing `tests/intake.test.mjs` so the suite that was actually throwing stops throwing. Mine contained the class and left the instance for someone else.

The framing is the right one — *"a discovered suite is a guest, not a co-host: its crash is one failure, not the end of the run"*. That's what the three existing guards in that loop were already saying, and an uncaught throw was the member of the family nobody had written down. Worse than the ones that were handled, because it fails quiet: node unwinds out of test-all, `finish()` never prints, and the screen of green above the cut looks exactly like a pass.

Printing the first frames of the stack is what I'd have skipped and would have wanted three days later. A suite that dies mid-import leaves no other trace of how far it got.

The number in your comment is the argument: 2067 of 3693 checks gone on a Windows box without symlink privilege — which is the DEFAULT configuration, so the person who hits it is a contributor on a machine unlike CI, and the least equipped to tell "the suite is broken" from "I broke it".

Merged 🚀

PR #2831, opened 2026-08-13
docs(invite-match): correct the isAIInterviewerPlatform jsdoc

santifer:
Follow-up to #2676, maintainer-side.

The review comment on #2676 pointed at `invite-match.mjs:319-320` and that block was rewritten correctly. The same claim also lived in the jsdoc on `isAIInterviewerPlatform` 60 lines below, which the ask did not name, so the file shipped saying two contradictory things about the same host:

- above `PLATFORM_URL_PATTERNS`: "NOT a guarantee that the host is exclusively AI-led"
- on `isAIInterviewerPlatform`: "single-purpose and always AI-led by product design"

The second is the one that is wrong, and it is the one a reader hits first when jumping to the function. Comment only, no behaviour change. Suite: 3756 passed, 0 failed.

## Summary by CodeRabbit

* **Documentation**
  * Clarified how platform links are interpreted when Alex and human platforms are both present.
  * Documented that Alex is recognized only when it wins the ordered platform scan.
  * Removed contradictory wording about Alex always being single-purpose and AI-led.


santifer:
Self-authored maintenance; merging on green CI.

PR #2834, opened 2026-08-13
test(agent-inbox): print the failing child's error when a concurrent add is lost

santifer:
Relates to #2777, follows #2824 / #2825.

## Why now

The `kept=29 of 30` failure **is still happening with #2825 merged**. It fired twice in the last hour on bases that contain the 30s budget: once on my machine right after an unrelated one-line docs merge, and once on a fresh CI run created at 21:29 UTC.

So the burst-sized budget cut this a long way and did not remove it, which is what both #2825 and the jitter comment before it predicted in writing.

## What is missing, and it is not a fix

Every one of these failures leaves exactly the same evidence: `kept=29 of 30`. That is the symptom. From outside the child, a lock-acquisition timeout, a Windows `EPERM`/`EBUSY` on the lock directory and a crash inside the append all look identical, and they have completely different fixes.

The distinction matters more now than it did at 8s: **a sub-millisecond append that cannot acquire the lock within 30 seconds is not simply a crowded queue.** If the next Windows failure still says `LockTimeoutError`, the answer is a fair queue. If it says `EPERM`, the lock directory is the problem and the budget was never the issue.

## What this does

Captures each child's stderr and prints the error line per lost item on failure. Node prints the offending source line before the error itself, so the `SomeError: message` line is extracted rather than the first lines taken verbatim, with an `EPERM|EBUSY|EACCES` fallback and the last line as a last resort.

Verified by forcing the timeout to 1ms: the case goes red and prints `LockTimeoutError: pipeline lock timeout: …held > 1ms` per lost item. Restored, green, and green runs print nothing new. Suite: 3783 passed, 0 failed.

@Scott-Emberson you have the most context on this one after #2824, so flagging it rather than assuming: this deliberately does not touch the budget or the lock. I would rather the next red tell us which of the two mechanisms it is than guess again.

## Summary by CodeRabbit

* **Bug Fixes**
  * Lock timeout errors now include detailed ownership diagnostics, such as process status, start time, and lock duration.
  * Diagnostic information remains available even when ownership data is missing or unreadable.

* **Tests**
  * Improved concurrent-add test diagnostics by capturing process error output.
  * Test failures now report clearer error causes while preserving existing item-survival checks.

### DIFF
--- a/agent-inbox-tests.mjs
+++ b/agent-inbox-tests.mjs
@@ -172,7 +172,11 @@ console.log('7. concurrent adds do not lose items (append, not rewrite)');
       || lines.find((s) => /\b(EPERM|EBUSY|EACCES|ENOENT|EEXIST)\b/.test(s))
       || lines.slice(-1)[0]
       || '(no stderr)';
-    console.log(`      ↳ ${l.item} exited ${l.code}: ${cause.slice(0, 200)}`);
+    // Generous, because the owner record pipeline-lock.mjs appends to a
+    // LockTimeoutError is the diagnostic payload; truncating it away would
+    // leave the same symptom-without-mechanism this instrumentation exists
+    // to end.
+    console.log(`      ↳ ${l.item} exited ${l.code}: ${cause.slice(0, 500)}`);
   }
   const body = readFileSync(inbox, 'utf8');
   const pending = body.split('\n').filter((l) => l.startsWith('- [ ]'));
--- a/agent-inbox-tests.mjs
+++ b/agent-inbox-tests.mjs
@@ -138,16 +138,42 @@ console.log('7. concurrent adds do not lose items (append, not rewrite)');
   const N = 30;
   // spawn(), not spawnSync() — a synchronous loop would serialize the adds and
   // pass even against the buggy rewrite, proving nothing.
-  const exits = await Promise.all(
+  // Capture each child's stderr, and PRINT the losers' when the case fails.
+  // Without this the only evidence a failure leaves is `kept=29 of 30`, which
+  // names the symptom and hides the mechanism: a lock-acquisition timeout, a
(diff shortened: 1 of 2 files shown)


santifer:
Second commit, and it is the one that matters.

A `LockTimeoutError` names the file and the budget and says nothing about **who was holding the lock**, which is exactly the fact that separates the two hypotheses. It now carries `owner={pid, alive, started_at, heldMs}`, and `heldMs` is the discriminator:

- **holder alive, `heldMs` in the single digits** → ordinary contention. The queue really is that crowded and the answer is a fair queue.
- **holder alive, `heldMs` in the tens of thousands** → the directory outlived its owner. On Windows `rmSync` of the lock can fail with `EPERM`/`EBUSY` while a handle is open, and `release()` swallows that by design; Windows also recycles PIDs aggressively, so a burst of 30 short-lived processes can hand a dead owner's PID to a live sibling. `lockCanRecover()` then reads "still running", judges the lock fresh forever, and **no budget is large enough**. That would also explain why exactly one item is lost rather than many: it strands whoever is still waiting when the orphan appears.
- **no readable `owner.json`** → the lock was taken but the stamp never landed, a third thing again.

Verified locally by forcing the budget to 120ms:

```
↳ item-0 exited 1: LockTimeoutError: … held > 120ms — owner={"pid":35593,"alive":true,"started_at":"2026-08-13T21:36:45.816Z","heldMs":2}
```

That is the healthy shape (`heldMs: 2`, a real holder mid-append). The Windows failure will print the same field, and whichever number it carries picks the fix.

Diagnostic only. The error is thrown exactly as before, the lookup is wrapped so it can never mask the timeout it describes, and `pipeline-lock.mjs` control flow is untouched, which matters because scan-history shares it. Suite: 3783 passed, 0 failed.

PR #2835, opened 2026-08-13
fix(pipeline-lock): time out a wedged holder, not an unlucky waiter

LayerC0de:
Part of #2777. Split out of #2815 at review request.

> **Stacks on #2815.** This branch contains that PR's commit as its base, so the diff shows two commits — **only the second, `fix(pipeline-lock): time out a wedged holder…`, is new here.** Once #2815 merges, this reduces to that commit alone. Reviewable independently; just merge #2815 first.

## The problem

The acquisition deadline is absolute, so it measures the wrong thing.

It exists to escape a holder that has stopped letting go. But it also expires on a caller that is merely losing the retry lottery while the lock is perfectly healthy and being handed round briskly. Those two situations want opposite answers, and the second is the common one — acquisition is a lottery, not a fair queue, so with N waiters somebody is always unlucky.

For callers here the cost of that confusion isn't latency. `agent-inbox.mjs`'s `add()` is one process per request, so a caller that dies takes the user's queued item with it: the exact loss the lock was introduced to prevent, converted from a silent overwrite into a loud crash.

## Why the existing mitigations stop short

Both do, by design, and both say so:

- **The retry jitter** cut collisions ~6x and is documented in-tree as *"a ~6x reduction, NOT a cure… the structural fix is a fair queue rather than a retry lottery."*
- **#2825** gave `agent-inbox` a 30s budget. That buys headroom for one caller without changing what the deadline *measures* — and `appendToPipeline` and `scan.mjs`'s `scan-history.tsv` appends still run on the 8s default.

## The fix

Re-arm the deadline whenever the lock is observed to have changed hands. `timeoutMs` then bounds **how long one holder may block a caller** rather than how many holders it may queue behind — which is what it was always documented to mean ("Max time to wait for the lock").

A wedged lock still fails exactly as before, same error and same timing, because an unchanged holder never re-arms it. `maxWaitMs` (default 10× `timeoutMs`) is the absolute ceiling, so progress buys patience but never immortality.

**This does not make the lock fair.** There's still no ordered wait, and a fair queue remains the structural answer. It makes unfairness cost latency instead of a write.

## Two details worth your attention

- **The fingerprint is sampled only when the deadline expires**, not on every retry — one extra `stat` per timeout window per waiter, rather than adding file-open churn to the contended path, which on Windows is itself a source of transient errors.
- **An unreadable fingerprint is not treated as "no progress."** It means the lock was free or unobservable at that instant, which is no evidence of a wedged holder; only a lock we can *see*, unchanged across a full window, is. I had this backwards in my first attempt — it killed an unlucky waiter at the first sample, and I caught it at N=60 before committing. Flagging it because it's the subtle half of the change.

## Evidence

30 concurrent adds, budget compressed to 300ms so the effect shows without waiting out the default:

```
before   8 of 12 rounds lost an item
after    0 of 20 rounds
```

60 concurrent adds at a 200ms budget: 0 of 10 rounds.

At the default budget the ceiling is generous enough that this never engages in normal use — it's the tail that it removes.

## Tests

- a lock that keeps changing hands does not time out a waiting caller
- extended waiting is still bounded by `maxWaitMs`, so endless churn cannot hang a caller

Both are deterministic — the churn helper swaps the lock inside a single synchronous timer callback, so an in-process waiter can never slip through the gap and the test doesn't depend on timing luck. **Both fail against the current deadline and pass against this one**, and the existing tests are unchanged in timing and outcome.

## If you'd rather not take this

Entirely reasonable — #2825 already covers the `agent-inbox` symptom, and this is the more opinionated half of the original PR. The crash fixes in #2815 stand alone and are the ones that close the CI flake. Closing this without prejudice is fine by me; the residual risk is that `appendToPipeline` and the `scan-history` appends keep the 8s absolute deadline and can still lose a write under a big enough burst.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved pipeline lock handling for missing, unreadable, or corrupted ownership information.
  - Added safer stale-lock recovery and cleanup during transient file-system conditions.
  - Prevented indefinite waiting when lock removal repeatedly fails.
  - Improved timeout diagnostics and preserved acquisition errors for clearer troubleshooting.

- **New Features**
  - Added a configurable maximum wait time for lock acquisition.
  - Improved resilience during active ownership changes with progress-aware timeout handling.
  - Added lock identification details to aid troubleshooting.
  - Added validation and fallback handling for invalid wait limits.

### DIFF
--- a/pipeline-lock.mjs
+++ b/pipeline-lock.mjs
@@ -38,6 +38,10 @@ const DEFAULT_STALE_MS = 30_000;
 export const OWNERLESS_GRACE_MS = 1_000;
 const DEFAULT_RETRY_MS = 80;
 const DEFAULT_TIMEOUT_MS = 8_000;
+// Ceiling on progress-extended waiting (see the deadline logic in
+// acquirePipelineLock). A multiple of timeoutMs rather than a constant, so a
+// caller that tunes one tunes both.
+const DEFAULT_MAX_WAIT_FACTOR = 10;
 
 const sleep = (ms) => new Promise((r) => setTimeout(r, ms));
 
@@ -187,15 +191,32 @@ function lockCanRecover(lockDir, staleMs) {
   }
 }
 
+// A cheap identity for "which holder currently has this lock". Used only to
+// tell a lock that is CHANGING HANDS (the system is making progress; this
+// caller is merely unlucky) from one that is WEDGED (a single holder is not
+// letting go, which is what the timeout exists to escape).
+function lockFingerprint(lockDir) {
+  try {
+    const st = statSync(lockDir);
+    const { owner } = readLockOwner(lockDir);
+    // ino is 0 on some Windows volumes, hence birthtime as the tiebreaker —
+    // the same pairing sameLockDirectory() already relies on.
+    return `${owner?.token ?? ''}|${st.ino}|${st.birthtimeMs}`;
+  } catch {
+    return null; // free or unobservable — no evidence either way
+  }
+}
+
 /**
  * Blocks until the lock on `pipelinePath` is held, then returns a handle whose
  * release() frees it. Throws LockTimeoutError if the lock stays busy.
  *
  * @param {string} pipelinePath - File the lock guards.
  * @param {object} [options]
- * @param {number} [options.timeoutMs=8000] - Max time to wait for the lock.
+ * @param {number} [options.timeoutMs=8000] - Max time a SINGLE holder may block this caller. Waiting is extended while the lock keeps changing hands (see below).
  * @param {number} [options.retryMs=80] - Delay between acquisition attempts.
  * @param {number} [options.staleMs=30000] - Age threshold for a lock with no readable owner, floored at OWNERLESS_GRACE_MS.
+ * @param {number} [options.maxWaitMs] - Absolute ceiling on waiting, however much progress the lock makes. Defaults to 10x timeoutMs.
  */
 export async function acquirePipelineLock(pipelinePath, options = {}) {
   // Env overrides let a caller several frames up the stack (a test driving
@@ -204,10 +225,44 @@ export async function acquirePipelineLock(pipelinePath, options = {}) {
   const timeoutMs = options.timeoutMs ?? (Number(process.env.CAREER_OPS_PIPELINE_LOCK_TIMEOUT_MS) || DEFAULT_TIMEOUT_MS);
   const retryMs = options.retryMs ?? (Number(process.env.CAREER_OPS_PIPELINE_LOCK_RETRY_MS) || DEFAULT_RETRY_MS);
   const staleMs = options.staleMs ?? (Number(process.env.CAREER_OPS_PIPELINE_LOCK_STALE_MS) || DEFAULT_STALE_MS);
+  // A ceiling that is not a number is not a ceiling. NaN compares false against
+  // everything, so a bad value would not fail loudly — it would silently remove
+  // the only bound on waiting, which is the single thing this option exists to
+  // provide. Only a usable ceiling is honoured; everything else takes the
+  // default, which is at least bounded.
+  //
+  // Read WITHOUT the `|| default` idiom its three siblings above use, because
+  // for this one 0 is a meaningful setting — "never wait past now" — and
+  // `0 || default` would turn a caller who asked for no waiting into one who
+  // waits DEFAULT_MAX_WAIT_FACTOR x timeoutMs, the exact opposite. timeoutMs
+  // and retryMs have no such reading, which is why the idiom is fine there and
+  // wrong here.
+  //
+  // Infinity is kept as written: an explicit, legible "no ceiling" is a choice,
+  // not a mistake. -Infinity and NaN are mistakes, so they take the default
+  // rather than being clamped into a silently different behaviour.
+  const defaultMaxWaitMs = timeoutMs * DEFAULT_MAX_WAIT_FACTOR;
+  const envMaxWait = process.env.CAREER_OPS_PIPELINE_LOCK_MAX_WAIT_MS;
+  const requestedMaxWait = Number(
+    options.maxWaitMs ?? (envMaxWait === undefined || envMaxWait.trim() === '' ? defaultMaxWaitMs : envMaxWait),
+  );
+  const maxWaitMs = requestedMaxWait === Infinity ? Infinity
+    : Number.isFinite(requestedMaxWait) && requestedMaxWait >= 0 ? requestedMaxWait
+      : defaultMaxWaitMs;
   const lockDir = lockDirFor(pipelinePath);
   const recoverGuardDir = `${lockDir}.recover`;
   const token = randomUUID();
-  const deadline = Date.now() + timeoutMs;
+  let deadline = Date.now() + timeoutMs;
+  // Deliberately NOT floored at timeoutMs. A caller that asks for a ceiling
+  // below one per-holder window means it, and quietly raising it would make
+  // maxWaitMs a no-op in precisely the range where it was stated most
+  // explicitly. The ceiling wins; timeoutMs just stops mattering.
+  const hardDeadline = Date.now() + maxWaitMs;
+  // Lock identity as of the last time this caller ran out of patience, so the
+  // next timeout can ask "did anything move since?". Sampled lazily on the
+  // first failed acquisition, not here, so the very first deadline is measured
+  // against the state we actually started waiting on.
+  let lastFingerprint;
   // The last mkdir error this caller treated as contention. On POSIX it is
   // always EEXIST and says nothing; on Windows an EPERM/EACCES that persists
   // to the deadline is the difference between "crowded" and "this process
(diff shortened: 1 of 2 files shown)


LayerC0de:
@coderabbitai both findings were correct. Fixed and rebased onto current `main`, integrating #2834's owner diagnostic.

**1. The spin after a failed stale cleanup** — real, and mine. Fixed in `c1ac57a`, which belongs to #2815 since the code is there. Probed on Windows with a genuinely undeletable lock directory: it hung a 20s probe with zero event-loop ticks, where `main` threw EPERM at 1ms. I had turned "kill the caller" into "hang the caller at 100% CPU". Now `rmQuiet()` reports whether the directory is gone and reclaim retries immediately only when it is.

**2. `maxWaitMs` was not actually a ceiling** — correct on all three counts, and they compound into the same outcome: an unbounded wait against a lock that keeps changing hands, which is the one thing the ceiling exists to prevent.

| | Was | Now |
|---|---|---|
| Floored at `timeoutMs` | `{timeoutMs: 5000, maxWaitMs: 200}` waited 5s — a no-op exactly where the caller was most explicit | ceiling wins outright; `timeoutMs` stops mattering below it |
| Only checked *inside* the per-holder deadline branch | a caller with `maxWaitMs < timeoutMs` never reached it, and a re-arm pushed the next look a full window away | checked on its own, every pass |
| Non-numeric → `NaN` | every comparison false, so a bad input silently *deleted* the bound rather than failing loudly | falls back to the default; `Infinity` still honored as a deliberate "no ceiling" |

Two regression tests, as you suggested — one for the lower explicit ceiling, one for the non-finite value. Both **race the acquire against a wall-clock bound instead of awaiting it**: against a lock that churns for ever a broken ceiling never returns, so a bare `await` would hang the suite rather than fail it. Verified both directions — they report `NEVER_SETTLED` after 3s and 4s against the previous commit, and settle in 219ms and 1005ms against this one (the latter being the 10x fallback doing its job).

Full suite on this branch: 17 tests, 16 pass, 1 skipped (the POSIX-only permissions test, skipped on Windows).

---

On the mergeability score: worth noting for whoever reviews that **this PR is the optional half.** #2815 alone closes everything actually observed on CI — I re-measured after splitting. This is the tail risk for `appendToPipeline` and the `scan-history` appends, neither of which got #2825's raised budget. If the maintainer would rather stop at #2815, closing this is a clean outcome and the PR description says so.

LayerC0de:
Rebased onto the rescoped #2815, which now sits on `main` with #2838 merged. This PR is unaffected by #2838 in substance — that fixed the mkdir crash; this is the starvation arm, and the deadline is still absolute on `main`.

@coderabbitai — all four remaining findings were valid and are fixed. Three of them turned out to be the same structural mistake, so the fix is one move:

**The ceiling is now tested first, on its own, on every pass.** Checking it only after a failed `mkdir` and only inside the per-holder deadline made it conditional on branches that may never run — a caller whose `maxWaitMs` is below `timeoutMs` never reaches the inner check, a re-arm pushes the next look a whole window away, and the reclaim fast-path `continue`s past both. That covers *"enforce before every acquisition attempt"* and *"check before the immediate retry"* together. A bound that holds only when another bound happens to fire is not a bound.

**The retry backoff is capped at the remaining ceiling.** You were right that this was still open: an uncapped sleep can wake past `hardDeadline`, find the lock free, and hand back a lock the caller was no longer entitled to wait for. The overshoot was bounded, but a *returned lock* is not a bounded error. New test: a lock freed after `maxWaitMs` but before the next retry must be **refused**, not acquired.

**The timeout names the bound that actually expired.** With `timeoutMs: 5000` and `maxWaitMs: 200` it now reports `held > 200ms` instead of a limit that was never in force. Asserted.

Carried over from the earlier revision, also from your review: `maxWaitMs` is no longer floored at `timeoutMs`, and a non-numeric value falls back to the default rather than producing a `NaN` that compares false against everything and silently deletes the ceiling.

18 tests, 18 pass, 0 skipped. Four of the five ceiling/deadline tests fail against this branch's base and pass against it; the fifth — non-numeric ceiling — discriminates against the *earlier revision of this PR* rather than the base, since the base has no ceiling to break. Worth stating plainly rather than claiming all five are load-bearing.

---

**Still the optional half, and more clearly so now.** #2838 closed the crash that reddened CI, and #2815 closes the remaining crash path. Neither needs this. What this adds is that `appendToPipeline` and the `scan-history` appends stop losing a write under a burst — they run on the 8s absolute default and got neither #2825's raised budget nor a fair queue. If you would rather stop after #2815, closing this is a clean outcome.

LayerC0de:
@coderabbitai both valid, both fixed in `94b37fa`, and the first one was the more interesting.

**The progress rule was applied on one path and not the other two.** You are right that a non-`EEXIST` guard refusal and an `ENOENT` owner-write retry both threw on a bare `Date.now() > deadline`, so a healthy lock being handed round briskly could still kill a caller through either — the exact bug this PR removes, left standing in two more places.

The real defect was three copies of one decision, two of them the old version. It now lives in a single `holderStillWedged()` that every timeout path calls, so it cannot hold in one place and quietly not apply in another.

**The reclaim fast path now yields.** Also right, and the distinction you drew is the one that matters: the ceiling bounds how *long* that loop can run, not what it *costs*. Every call on that path is synchronous, so a peer recreating a reclaimable lock pins the event loop for up to the whole ceiling — 80 seconds at the default budget, with the process making no progress at all. A single `sleep(0)` keeps the immediate-retry cadence and keeps the caller interruptible. Placed after the guard's `finally`, so the recover guard is never held across the yield.

Behaviour unchanged where it was already right: 18/18 tests, the undeletable-lock probe still ends at ~400ms with a responsive event loop (6 watchdog ticks), 16×40 acquire/release cycles crash 0, and 30 concurrent adds at a compressed 300ms budget keep 30 of 30 across 8 rounds.

LayerC0de:
@coderabbitai the `=0` half is right and it inverts the request rather than merely bending it. Fixed in `00b1829`.

`maxWaitMs` was read with the `Number(env) || DEFAULT` idiom its three siblings use. For `timeoutMs`/`retryMs`/`staleMs` that idiom is fine — none of them has a useful reading at 0. For a **ceiling**, 0 is meaningful: *never wait past now*. So `CAREER_OPS_PIPELINE_LOCK_MAX_WAIT_MS=0` selected `10 * timeoutMs`, and a caller asking for no waiting got the longest wait the module offers.

Measured against a churning lock, which is the only setup where the ceiling is the deciding bound:

```
env=0             9ms   (was 515ms — the inversion)
option 0         13ms
env unset       515ms   default 10x timeoutMs
env blank       510ms   default
NaN             511ms   default, still bounded
-5 / -Infinity  513ms   default, no longer clamped to 0
explicit 200    216ms   honoured
```

**On the `-Infinity` half I went further than the suggestion**, and it is worth saying why. The proposal kept finite negatives clamped to `0` while defaulting `-Infinity`, which leaves `-5` and `-Infinity` behaving differently for no reason a caller could predict. Both are mistakes, so both take the default now. Only a usable ceiling is honoured: a finite value `>= 0` as written, `Infinity` as the deliberate no-ceiling it plainly is, everything else defaulted — bounded and obvious rather than silently reinterpreted.

Two tests, written as a pair because they pin that `0` and *below* `0` are not the same input: one asserts a zero ceiling gives up immediately, the other that a negative one does not. Verified both directions — against the previous normalization the first waits out its 3s bound and the second gives up in 7ms.

One note on method, since it cost me a wrong first attempt: this is only observable against a lock that keeps **changing hands**. With a wedged lock the per-holder deadline fires first and the ceiling never decides, so a naive harness shows the same number for every setting and looks like the code is ignoring the option.

20/20 tests.

santifer:
Read the stacking note, @LayerC0de — splitting it out at review request and saying plainly *"only the second commit is new here; merge #2815 first"* is exactly what makes a stacked PR reviewable instead of a puzzle. Thanks for doing that rather than making me diff two branches to find the boundary.

The distinction in the title is the right one: **a wedged holder and an unlucky waiter should not share a deadline.** An absolute acquisition deadline measures how long *this caller* has been waiting, which says nothing about whether the holder is stuck — so under contention the process that gets timed out is whichever one happened to arrive at a bad moment, while the actually-wedged holder keeps the lock.

Order noted: #2815 first, this second. I've flagged #2815 as first in the queue next cycle — I think it may be the root cause of #2777, which would make that branch considerably more urgent than its size suggests. This one rides in behind it.

Not merged tonight: per-session merge ceiling.

LayerC0de:
Understood on the order, @santifer — #2815 first, this behind it. No action needed from me.

One correction worth carrying over from #2777: I could **not** confirm that #2815 is the root cause there. 40 rounds × 30 adds on two cores showed 0 failures on both `main` and that branch, and the mtime age-out in your snippet can't fire in that test at all — `staleMs` defaults to 30 000 while the burst lasts ~1.5 s. So please don't treat "#2815 closes #2777" as established when it reaches the front of the queue; it fixes a real bug either way, but the link is unproven.

That doesn't change anything about this PR, which is independent of both.

santifer:
Carried over and retracted publicly — I've updated both #2777 and #2815 to say the link is unproven and that #2815 stands on its own as a correctness fix, which was always the stronger framing.

Thanks for repeating it here rather than assuming I'd read it on the other thread. A correction that only exists in one of three places is a correction that gets lost when someone comes back to the queue in a week and opens whichever one is on top — which is exactly the position I'll be in next cycle.

Order unchanged: #2815, then this. Nothing needed from you.

LayerC0de:
Rebased onto the rescoped #2815, which is itself rebased onto #2984. Was `CONFLICTING`; now green.

Worth restating what this PR is **not**, since three lock fixes have landed around it: #2984 and #2815 both address a caller **crashing** mid-acquire. This is the other failure direction — a caller that never crashed and never lost a race, but was timed out for being unlucky while the lock was healthy and briskly shared. Nothing merged so far touches that, and `appendToPipeline` and `scan.mjs`'s scan-history appends still run on the 8s absolute default.

One structural change from the previous revision, from an earlier review round: all three timeout paths now route through a single `holderStillWedged()`, instead of the progress rule holding on the main retry and quietly not applying to the guard-refusal and ENOENT-owner-write paths.

Five of the six ported tests fail against this branch's base and pass against it. The sixth — non-numeric ceiling — discriminates against an earlier revision of this change rather than the base, which has no ceiling to break; saying so because "all N tests fail without the fix" would be the tidier claim and it would not be true.

Full suite 4065 passed, 1 failed (the pre-existing `js-yaml` one).

santifer:
Rebased maintainer-side onto today's main: #2815 merged (squash), so your stack base was patch-identical to main under a different SHA and update-branch conflicted with itself. `git rebase` detected the duplicate and dropped it — the branch is now your progress-deadline commit alone, exactly the reduction your PR body predicted. Full suite green locally on the result (4265/0). CI runs on the real base now; merging on green. Review verdict was already reached this afternoon: the re-arm design with the ceiling tested first and `holderStillWedged()` unifying the three timeout paths is right, and the test matrix covers every parsing edge I would have probed.

santifer:
Merged 🚀 This one changes what the timeout MEASURES, and that's the fix, @LayerC0de: timeoutMs now bounds how long one holder may block a caller (what it was always documented to mean), with the ceiling tested first on every pass — "a bound that holds only when another bound happens to fire is not a bound" belongs in a textbook. holderStillWedged() unifying the three timeout paths killed the same bug in two more places before it existed, and the maxWaitMs parsing matrix (0 meaningful, Infinity legible, NaN to default) is exactly as careful as the fingerprint fallback. Two lock fixes in one day, both after rescoping twice around my merges: thank you for the patience as much as the code.

PR #2836, opened 2026-08-13
test(classify-tier): pin guard (b) scoping with negative cases

Zoubeir23:
## Summary

Follow-up to #2737 (merged). @Scott-Emberson  noted in code review that guard (b) was only tested in the positive direction (bridge noun + senior noun → `senior`). The scoping condition — that no senior noun must follow for the guard to not fire — was unasserted, leaving a silent regression path.

- Add three negative cases to `tests/classify-tier-position.test.mjs`:
  - `Intern Program Coordinator` → `intern`
  - `Graduate Scheme Analyst` → `intern`
  - `Trainee Program Engineer` → `intern`
- These prove the trailing-noun check at `classify-tier.mjs:119` is load-bearing. Removing it would flip all three to `senior`, causing junior candidates to silently lose intern-programme roles.

## Test plan

- [ ] `node tests/classify-tier-position.test.mjs` — all 44 assertions pass (3 new negatives added)
- [ ] No production code changed; risk is zero

Closes the gap flagged in Scott-Emberson's review of #2737.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved the original seniority classification for intern, graduate, and trainee roles when bridge nouns appear without a trailing senior-role noun.
* **Tests**
  * Added regression coverage for these job-title classification scenarios.

### DIFF
--- a/tests/classify-tier-position.test.mjs
+++ b/tests/classify-tier-position.test.mjs
@@ -100,6 +100,14 @@ try {
   check('Junior Talent Director', 'senior');
   check('Junior Talent Lead', 'senior');
   check('Entry-Level Program Director', 'senior');
+
+  // ── Guard (b) scoping: bridge noun present but NO trailing senior noun → stays at level marker ──
+  // These cases prove the guard does not overreach. Removing the trailing-noun
+  // requirement at classify-tier.mjs:119 would flip all three to 'senior' and
+  // a junior candidate would silently lose intern-programme roles.
+  check('Intern Program Coordinator', 'intern');
+  check('Graduate Scheme Analyst', 'intern');
+  check('Trainee Program Engineer', 'intern');
 } catch (error) {
   fail(`classify-tier.mjs tests could not run: ${error.message}`);
 }


Scott-Emberson (approved):
Approving. This backfills exactly the gap I flagged on #2737: guard (b) was only pinned in the positive direction, so nothing proved the trailing-noun check at `classify-tier.mjs:119` was load-bearing.

I traced each of the three titles through guard (b): the bridge matches (`Intern Program`, `Graduate Scheme`, `Trainee Program`) but the remainder (`Coordinator` / `Analyst` / `Engineer`) has no trailing senior noun, so the guard correctly stays quiet and the leftmost-marker rule returns `intern`. Then I reran with line 119 mutated to fire unconditionally: all three flip to `senior` and fail. So they genuinely constrain the guard rather than decorate it, and the "removing it would flip all three to senior" claim is measured, not assumed.

Ran the file directly, 44/44 pass. Hermetic, same `check()` shape as its neighbours. Good to go.

santifer:
Eight lines, and they're the right eight, @Zoubeir23:

```js
// Removing the trailing-noun requirement at classify-tier.mjs:119 would flip
// all three to 'senior' and a junior candidate would silently lose intern-programme roles.
check('Intern Program Coordinator', 'intern');
check('Graduate Scheme Analyst', 'intern');
check('Trainee Program Engineer', 'intern');
```

Negative cases that pin **where a guard stops** are worth more than positive ones, because a guard's failure mode is almost always overreach, and overreach passes every test that only checks it fires when it should. "Intern Program Coordinator" contains a bridge noun and no trailing senior noun — exactly the shape that would break if someone loosened line 119 for a reason that looked good at the time.

Naming the consequence in a comment (*a junior candidate silently loses intern-programme roles*) is what stops the next person deleting the test as redundant. A test that says what breaks is an argument; one that just asserts is an obstacle.

Sorry for the slow first reply. Not merged tonight (per-session merge ceiling), queued.

santifer:
pin guard (b) scoping with negative cases (#2836):Eight lines and they're the right eight, exactly as reviewed: the negative cases pin the scoping instead of assuming it. Merged! 🚀

PR #2837, opened 2026-08-13
test(updater-rollback): isolate the fixture from global commit signing

### LINKED ISSUE #2754: test-all.mjs hard-aborts on any machine with global commit signing: the #2015 fixture inherits commit.gpgsign
`node test-all.mjs` **hard-aborts** partway through on a machine that has git commit signing configured globally. Not a failed assertion with a summary at the end: the process dies and every later section is silently never run, so the final `Results:` line never prints.

## Reproduction

Any machine where `git commit` is configured to sign and the signer cannot produce a signature non-interactively. Mine:

```
$ git config --global --get commit.gpgsign
true
$ git config --global --get gpg.format
ssh
$ git config --global --get gpg.ssh.program
/Applications/1Password.app/Contents/MacOS/op-ssh-sign
```

Then:

```
$ node test-all.mjs
...
🧪 Testing updater rollback behavior (#2015)...
error: 1Password: failed to fill whole buffer

fatal: failed to write commit object
node:internal/errors:998
Error: Command failed: git commit -qm base
    at gitIn (update-system.mjs:509:12)
```

The suite stops there. On my machine that is roughly 3400 assertions past the abort point that never execute.

## Mechanism

The `#2015` updater-rollback test builds a scratch git repo and commits into it through `gitIn()`:

```js
// update-system.mjs:505
export function gitIn(root, ...args) {
  const timeout = gitTimeoutMs(args);
  try {
    return execFileSync('git', args, { cwd: root, encoding: 'utf-8', timeout }).trim();
```

`cwd` is the fixture, but the environment is inherited whole. `commit.gpgsign` lives in the user's **global** config, which applies to every repo including a fresh `git init` in a tmpdir, so the fixture commit tries to sign. When the signer is not available non-interactively the commit fails, `execFileSync` throws, and nothing in the section catches it.

The signing program is incidental. Any global `commit.gpgsign=true` whose signer cannot run unattended reaches the same place: a GPG key whose agent is not unlocked, an SSH signer behind a biometric prompt, a signing key absent from a CI image.

## Why the dev machines do not see it

Same shape as #2269 and #2752: it depends on the developer's own configuration, so it is invisible to anyone whose global config does not sign. Signing is common enough (it is what produces the "Verified" badge on commits in this repo) that this likely affects more than one contributor, and the failure mode is bad: it looks like the suite passed up to that point and then the terminal just ends.

## Suggested fix

The isolation pattern is already in this repo. `test-all.mjs` section 12c pins `GIT_CONFIG_GLOBAL` / `GIT_CONFIG_SYSTEM` plus an explicit `core.excludesFile` when it builds a git fixture (#2569). The `#2015` fixture needs the same treatment. Either:

- pin `GIT_CONFIG_GLOBAL=/dev/null` and `GIT_CONFIG_SYSTEM=/dev/null` in the env passed to `gitIn()` for fixture repos, or
- pass `-c commit.gpgsign=false` on the fixture's `git commit`, which is narrower but leaves other inherited global settings in play.

I would go with the first, to match 12c and to neutralize the whole class rather than this one setting.

Worth considering separately: `gitIn()` is also used on the user's real repo by `update-system.mjs`, so the pin belongs at the fixture's call site rather than inside `gitIn()` itself, unless you want updater commits to stop signing too.

## Workaround for anyone hitting this now

```
GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null node test-all.mjs
```

With that, the suite runs to completion here: 3437 passed, 0 failed, 1 warning (the warning being `cv-sync-check.mjs exited with error (expected without user data)`, a fresh-checkout artifact).

## Environment

- career-ops at `main` (`9048c5d`)
- Node v26.4.0, macOS 15 (Darwin 25.5.0)

Mentioned this on #2752 while fixing an unrelated doctor bug; filing separately since it is its own defect. Happy to send the PR if you want it.

santifer:
Fixes #2754.

## Reproduced, and it is worse than a failing test

`gitIn` inherits the environment, so a contributor's **global** git config applies inside the throwaway repo this fixture builds. With `commit.gpgsign = true` set globally, every commit in the fixture fails.

Because these are `execFileSync` calls, that is not a red assertion: **the process dies and every later section of `test-all.mjs` silently never runs**, so the `Results:` line never prints. @ringo380 measured roughly 3400 assertions past the abort point. From the contributor's side it reads as "career-ops crashes on my machine", not "one test needs fixing".

Simulated a global signing config to confirm:

| | outcome |
|---|---|
| before | aborts inside this file, `Results:` **never printed** |
| after | `3787 passed, 0 failed`, same environment |

## Why it was only this one file

The sibling fixture in `tests/updater-local-system-edits.test.mjs` has carried exactly these two lines since a CodeRabbit review flagged the same hazard, with the reasoning in a comment. Confirmed by running both under the simulated config: the sibling passes 14 green, this one dies. So the fix already existed in the house and one fixture was left behind, which is why the failure looks environment-specific instead of structural.

`core.hooksPath` is included for the same reason the sibling includes it: a global hooks path breaks the fixture's commits identically.

Thanks @ringo380 for a report that named the mechanism and the line number: this took longer to reproduce than to fix.

## Summary by CodeRabbit

* **Tests**
  * Improved rollback test reliability by isolating Git configuration from the local environment.
  * Prevented failures caused by global commit-signing settings or unavailable repository hooks.

### DIFF
--- a/tests/updater-rollback-behavior.test.mjs
+++ b/tests/updater-rollback-behavior.test.mjs
@@ -23,6 +23,20 @@ function makeRepo() {
   g('init', '-q', '-b', 'main', '.');
   g('config', 'user.email', 'test@example.com');
   g('config', 'user.name', 'Test');
+  // `gitIn` inherits the environment, so the contributor's GLOBAL git config
+  // applies inside this throwaway repo. With `commit.gpgsign = true` set
+  // globally (1Password's ssh signer, gpg-agent, a hardware key) every commit
+  // below fails, and because these are execFileSync calls the failure is not a
+  // red assertion: the process DIES here and every later section of the suite
+  // silently never runs, so `Results:` never prints (#2754). A global
+  // `core.hooksPath` breaks it the same way.
+  //
+  // The sibling fixture in updater-local-system-edits.test.mjs has carried
+  // these two lines since a CodeRabbit review flagged the same thing; this one
+  // was left behind, which is why the failure looks environment-specific
+  // instead of structural.
+  g('config', 'commit.gpgsign', 'false');
+  g('config', 'core.hooksPath', join(dir, 'no-such-hooks'));
   return { dir, g, ctx: { git: g, root: dir } };
 }
 


Scott-Emberson (approved):
Approving as code owner for `tests/`. This isolates the rollback fixture from a contributor's global git config, and the two lines earn their place.

I reproduced the mechanism with a simulated global config (`GIT_CONFIG_GLOBAL` set to `commit.gpgsign = true`, so nothing touched my real setup). With the fix present it passes; strip the two lines and the same config kills the process mid-file with `gpg: signing failed: No secret key` and `fatal: failed to write commit object`, exit 1, no `Results:` line. That is the silent truncation from #2754, so the isolation is load-bearing, not decorative. `core.hooksPath` at a non-existent dir is the standard no-hooks neutralization and matches the sibling fixture.

Ran the file directly on Windows: green with and without the simulated global signing.

The red windows-latest check is an unrelated pre-existing flake in another suite, not this change; the rollback tests are green on all three platforms.

santifer:
Self-authored maintenance; merging on green CI.

PR #2838, opened 2026-08-13
fix(pipeline-lock): a non-EEXIST mkdir refusal is contention, not a fatal error

### LINKED ISSUE #2777: agent-inbox: concurrent adds still lose an item on Windows (appendFileSync is not atomic there)

santifer:
Fixes #2777. Follows #2824 / #2825 / #2834.

## The instrumentation paid off on its first Windows failure, and it was not what any of us thought

#2834 landed a few minutes ago to print the failing child's error instead of only `kept=29 of 30`. The very next `test (windows-latest)` red said:

```
↳ item-27 exited 1: Error: EPERM: operation not permitted, mkdir
  'C:\Users\RUNNER~1\AppData\Local\Temp\inbox-concurrent-SHgNgf\agent-inbox.md.lock.recover'
```

**Not a `LockTimeoutError`.** Not a starving writer. A crash.

## Root cause

`mkdir`'s "someone else already has this" answer is not portable. POSIX gives `EEXIST`; **Windows gives `EPERM`/`EACCES` when the target is mid-flight**, being created or removed by another process at that instant. `acquirePipelineLock` had:

```js
if (err?.code !== 'EEXIST') throw err;          // lock dir
if (guardErr?.code !== 'EEXIST') throw guardErr; // recover guard
```

So a transient refusal that *is* the contention this loop exists to handle propagated straight out, the writer died, and its queued line was never appended.

## Why it took three attempts to see

A starving writer and a writer killed by `EPERM` produce **the identical symptom**, `kept=29 of 30`, and only the second is a crash. #2506 added jitter, #2825 raised the budget from 8s to 30s; both were reasoning about the timeout because the timeout was the only mechanism anyone could see. My own hypothesis this evening was a third wrong one: an orphaned lock whose PID had been recycled. The log settled it in one line.

The general shape, which I would rather record than repeat: **a symptom that two different mechanisms can produce will absorb any number of fixes aimed at the wrong one.** The cheap move was never a better guess, it was making the failure say its own name.

## The change

`EPERM`/`EACCES` from `mkdir` route to the same retry path as `EEXIST`. In the guard branch they additionally back off instead of judging the staleness of a directory that could not even be created reliably.

**The trade, stated because it is real:** a genuine permissions problem no longer fails fast, it now takes `timeoutMs` and then reports. Left silent that would swap one invisible failure for another, so the timeout carries the errno it kept hitting:

```
LockTimeoutError: pipeline lock timeout: … — owner={…}, last mkdir error=EACCES on /…/pipeline.md.lock
```

## Verification

Driven by a **real** refusal rather than a stub: a parent directory with no write permission makes `mkdir` answer `EACCES`, which travels the same branch. The test skips loudly under root, where permission bits do not apply and a pass would be vacuous.

Verified by mutation, restoring the `EEXIST`-only check:

```
❌ expected LockTimeoutError after retrying, got Error: EACCES: permission denied, mkdir '…/pipeline.md.lock'
❌ timeout did not carry the mkdir errno: lastMkdirError=undefined
```

That raw throw at the caller is the Windows crash, reproduced on macOS.

Suite: **3789 passed, 0 failed**. `validate-system-paths-coverage.mjs`: exit 0, 1036 files. `pipeline-lock.mjs` is shared with scan-history, and `tests/scan-history-lock.test.mjs` plus `agent-inbox-tests.mjs` are both green.

@Scott-Emberson this is the answer to your #2824, and your reasoning there was right about everything except which errno: the budget genuinely was short at 8s, it just was not what killed this one.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cross-platform handling when concurrent processes contend for a pipeline lock.
  * Temporary permission-related lock failures are now retried instead of failing immediately.
  * Lock timeout errors now include clearer ownership and recent failure details.

* **Tests**
  * Added regression coverage for permission-denied lock acquisition scenarios.

### DIFF
--- a/pipeline-lock.mjs
+++ b/pipeline-lock.mjs
@@ -69,6 +69,28 @@ function sameLockDirectory(left, right) {
     && (left.ino !== 0 || left.birthtimeMs === right.birthtimeMs);
 }
 
+// mkdir's "someone else already has this" answer is NOT portable. POSIX gives
+// EEXIST; Windows gives EEXIST *sometimes* and EPERM/EACCES when the target is
+// mid-flight — being created, or being removed, by another process at that
+// instant. That is not an error condition here, it is the contention this loop
+// exists to handle, and it is precisely what a burst of concurrent writers
+// manufactures.
+//
+// Treating it as fatal is how an item gets LOST. Measured on windows-latest
+// (#2777, run 31745798742): one of 30 concurrent `agent-inbox add` processes
+// died with `EPERM: operation not permitted, mkdir '…agent-inbox.md.lock.recover'`
+// and its line was never appended. The two budget increases before this one
+// (#2506 jitter, #2825's 30s) both moved the symptom without touching this,
+// because a starving writer and a writer killed by EPERM produce the same
+// `kept=29 of 30` and only the second is a crash.
+//
+// A genuine permissions problem still surfaces: it simply stops being an
+// instant throw and becomes a timeout that names the last error it saw, which
+// is the correct trade when the alternative is silent data loss.
+function isMkdirContention(err) {
+  return err?.code === 'EEXIST' || err?.code === 'EPERM' || err?.code === 'EACCES';
+}
+
 function processIsAlive(pid) {
   if (!Number.isInteger(pid) || pid <= 0) return false;
   try {
@@ -123,6 +145,37 @@ export async function acquirePipelineLock(pipelinePath, options = {}) {
   const recoverGuardDir = `${lockDir}.recover`;
   const token = randomUUID();
   const deadline = Date.now() + timeoutMs;
+  // The last mkdir error this caller treated as contention. On POSIX it is
+  // always EEXIST and says nothing; on Windows an EPERM/EACCES that persists
+  // to the deadline is the difference between "crowded" and "this process
+  // cannot create directories here at all", and without it a real permissions
+  // problem would present as a plain, unexplained timeout.
+  let lastContentionError = null;
+
+  // Built at the throw site so the diagnosis reflects the moment it gave up.
+  // A timeout on a critical section that is a single sub-millisecond append
+  // has more than one explanation, and the owner record separates them:
+  //   - no owner, or an owner whose PID is dead: real contention, and the
+  //     structural answer is a fair queue rather than a bigger budget.
+  //   - an owner reported ALIVE after tens of seconds: the directory outlived
+  //     its holder, because release()'s rmSync can fail on Windows while a
+  //     handle is open and is swallowed by design.
+  // Diagnostic only: it never changes whether the error is thrown.
+  const buildTimeoutError = () => {
+    const err = new LockTimeoutError(lockDir, timeoutMs);
+    try {
+      const owner = readLockOwner(lockDir);
+      err.owner = owner
+        ? { pid: owner.pid, alive: processIsAlive(owner.pid), started_at: owner.started_at, heldMs: Date.parse(owner.started_at) ? Date.now() - Date.parse(owner.started_at) : null }
+        : { pid: null, alive: null, note: existsSync(lockDir) ? 'lock exists with no readable owner.json' : 'lock vanished before it could be read' };
+      err.message += ` — owner=${JSON.stringify(err.owner)}`;
+      if (lastContentionError && lastContentionError.code !== 'EEXIST') {
+        err.lastMkdirError = lastContentionError.code;
+        err.message += `, last mkdir error=${lastContentionError.code} on ${lastContentionError.path ?? '?'}`;
+      }
+    } catch { /* diagnosis must never mask the timeout it describes */ }
+    return err;
+  };
 
   // A fresh install may not have data/ yet — plugins.mjs's cmdRun calls
   // appendToPipeline with no directory pre-creation, so create it here rather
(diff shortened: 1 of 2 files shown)


Scott-Emberson (changes requested):
Reviewing as code owner of `/tests/`. The `pipeline-lock.mjs` change is santifer's to gate, so this speaks only to the new test file.

The fix itself reads right: `EPERM`/`EACCES` are how Windows spells the contention this loop already handles, and routing them to the retry path with the errno stamped onto the timeout is the correct shape. macOS and ubuntu are green. The problem is the new test does not survive the platform it exists for.

**1. Blocker: `tests/pipeline-lock-mkdir-eperm.test.mjs` fails on windows-latest.**

The fixture denies `mkdir` with `chmodSync(sealed, 0o500)`. On Windows that does not deny anything. Node's `chmod` there only toggles the read-only attribute, and that attribute does not stop directory creation inside `sealed`. The root guard does not cover it either, because `process.getuid` is `undefined` on Windows, so `typeof process.getuid === 'function'` is false and execution falls into the else branch.

So on Windows `acquirePipelineLock` acquires the lock, `thrown` stays `undefined`, and both assertions fail. Confirmed on the CI run for this head:

```
🔒 pipeline-lock: a non-EEXIST mkdir refusal is contention, not fatal
  ❌ expected LockTimeoutError after retrying, got undefined:
📊 Results: 3783 passed, 2 failed, 2 warnings
```

and reproduced locally on Windows 11 (Node v22.23.1) against this branch's `pipeline-lock.mjs`, same two failures. Both of those failures are this one file, and it is the only thing red on windows.

The fix is to guard the fixture on `win32`: skip it there the way root is already skipped and keep the POSIX `EACCES` path as the real assertion, or back the denial with a Windows ACL rather than a mode bit. A test whose denial mechanism is a no-op on a platform does not just lose coverage there, it turns the branch red on the exact OS #2777 is about.

**2. Minor: the test cannot distinguish a retry-then-timeout from a first-attempt timeout.**

It checks the final error type and `lastMkdirError`, but nothing observes that the loop retried. Code that wrapped the first `EACCES` into a `LockTimeoutError` without ever looping would still pass this. An attempt counter or an injectable sleep hook closes that. Worth folding into the same edit rather than a separate pass.

Once the fixture skips-or-ACLs on Windows and that job is green, this is good from the tests side.

santifer:
**The fix works on Windows, and the same CI run proves it.** In job 94612328026, `agent-inbox-tests.mjs` section 7 passed: no `kept=29 of 30`, no `EPERM`, all 30 items survived. That is the first `test (windows-latest)` in this whole sequence where the concurrency case is simply green rather than green-by-lottery.

**What failed instead was my own test**, and the reason is worth writing down rather than quietly patching:

```
❌ expected LockTimeoutError after retrying, got undefined
❌ timeout did not carry the mkdir errno: lastMkdirError=undefined
```

The case manufactures a refusal by removing write permission from the parent directory. That works on POSIX. On Windows it does not: `chmod` maps onto the read-only attribute, which does not deny `mkdir` inside the directory, so acquisition simply succeeds and nothing is thrown. I wrote a test for a Windows bug that cannot run on Windows, and CI told me within minutes.

Now skipped on `win32` with the reason stated, alongside the existing root skip. That reads as a contradiction and is not one: `EACCES` and `EPERM` meet in the same branch of `isMkdirContention()`, so exercising either covers the code path. What cannot be reproduced on demand *anywhere* is the Windows timing window, which is exactly why the test manufactures a deterministic refusal instead of racing for one.

Local suite green. The measurement that matters is the one already in the log above: the failure this PR exists to remove did not happen on the platform where it always happened.

PR #2842, opened 2026-08-14
fix(waf-pacing): raise INTER_PAGE_DELAY_MS to 250 ms across all paginated providers (#2706)

Zoubeir23:
## Summary

- **Root cause:** All paginated providers shared a 150 ms inter-page delay, the same value that was verified to trigger a Datadog WAF ban on Getro (3 boards at 403 + ~2.5 h ban, #2706). The Getro fix raised its own value but left 6 other providers at 150 ms.
- **Fix:** Raise `INTER_PAGE_DELAY_MS` from 150 → 250 ms in `workday`, `icims`, `avature`, `eightfold`, `oraclecloud`, and `tencent`. `alibaba` (300 ms) and `meituan` (400 ms) were already above the floor.
- **Tests:** Strengthened assertions in every affected provider test to verify the sleep value is `>= 250 ms` (previously only checked `> 0` or just counted calls). Added count-based no-sleep-before-page-0 assertions and single-page zero-sleep tests.
- **CI guard:** Added a check in `test-all.mjs` that fails if any provider file has `INTER_PAGE_DELAY_MS < 250`, preventing future regressions.
- **Docs:** Added the 250 ms rationale comment to each provider and a cross-provider note near the `sleep()` helper in `providers/_http.mjs`.

## Providers changed

| Provider | Before | After |
|----------|--------|-------|
| `workday.mjs` | 150 ms | 250 ms |
| `icims.mjs` | 150 ms | 250 ms |
| `avature.mjs` | 150 ms | 250 ms |
| `eightfold.mjs` | 150 ms | 250 ms |
| `oraclecloud.mjs` | 150 ms | 250 ms |
| `tencent.mjs` | 150 ms | 250 ms |

## Test plan

- [ ] `node --test tests/providers/workday.test.mjs` — new pacing assertions pass
- [ ] `node --test tests/providers/icims.test.mjs` — new pacing assertions pass
- [ ] `node --test tests/providers/avature.test.mjs` — new pacing assertions pass
- [ ] `node --test tests/providers/eightfold.test.mjs` — new pacing assertions pass
- [ ] `node --test tests/providers/oraclecloud.test.mjs` — new pacing assertions pass
- [ ] `node --test tests/providers/tencent.test.mjs` — new pacing assertions pass
- [ ] `node --test tests/providers/alibaba.test.mjs` — strengthened assertion passes (300 ms >= 250)
- [ ] `node test-all.mjs` — 3807 passed, 0 failed
- [ ] WAF pacing CI guard (test-all.mjs) passes: all 74 provider files clear

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved pagination reliability by increasing the minimum delay between page requests to 250 ms across supported providers.
  - Single-page and empty results no longer incur unnecessary delays.

- **Tests**
  - Added coverage for safe pagination pacing, including exact inter-page delays and no delay before the first page.
  - Added an automated guard to detect providers using delays below the minimum threshold.

### DIFF
--- a/providers/_http.mjs
+++ b/providers/_http.mjs
@@ -105,6 +105,11 @@ const RETRY_DEFAULTS = { retries: 2, baseDelayMs: 500, maxDelayMs: 8_000 };
  */
 const REDIRECT_REFUSAL_CAUSE_MESSAGE = 'unexpected redirect';
 
+// Cross-provider safe floor for INTER_PAGE_DELAY_MS: 250 ms.
+// 150 ms was verified to trigger Datadog WAF bans on Getro (3 boards at 403
+// + ~2.5 h ban at a board level). Every paginated provider inherits this risk
+// and must use >= 250 ms for any inter-page sleep (#2706).
+
 /** Awaitable sleep that honours a ctx-supplied clock, so tests never wall-clock wait. */
 export function sleep(ms, ctx) {
   if (typeof ctx?.sleep === 'function') return ctx.sleep(ms);
--- a/providers/avature.mjs
+++ b/providers/avature.mjs
@@ -34,9 +34,11 @@ const DEFAULT_MAX_PAGES = 50; // ~300 postings; override via entry.max_pages
 const HARD_MAX_PAGES = 200;
 // Pause between successive page requests. Avature's 6-results-per-page cap makes
 // large boards request-heavy (a 999+ board is ~170 pages); firing those with no
-// gap risks the tenant's WAF rate-limiting the burst. Mirrors workday's
-// INTER_PAGE_DELAY_MS — only boards that paginate past page 0 pay it.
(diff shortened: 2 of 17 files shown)


santifer:
Following up on this one, @Zoubeir23, because the close left a real gap: `workday.mjs:44` and `icims.mjs:22` still carry `INTER_PAGE_DELAY_MS = 150` on today's main — the exact value #2706 measured triggering the Datadog WAF ban (3 boards at 403, ~2.5h). You closed this the same night you split the #2841 batch into surgical pieces, and the other pieces landed (#2836, #3007, #3009…), but this one never came back in any form I can find. If a re-slice is still on your list, it has a fast lane: the fix was already right. If you'd rather I take it, say so and I'll carry it with your authorship credited — the measurement and the strengthened assertions are yours either way.

PR #2843, opened 2026-08-14
fix(web): preserve malformed portals config

moelomda:
## What does this PR do?

Prevents the web onboarding flow from treating a malformed existing `portals.yml` as a missing file and overwriting it with the example configuration. The route now seeds only when the file is genuinely absent, returns `409` for invalid user YAML, preserves custom configuration blocks, and includes regression tests.

## Related issue

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

## Verification

- `npm test` — 154 web tests passed
- `npm run build` — passed
- `node --test web/tests/lib/portals-config.test.mjs` — 5 regression tests passed

The complete Windows root harness remains blocked by the existing symlink-permission failure tracked in #2828; this PR's relevant suites pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved portal configuration loading and validation.
  * Preserved existing custom portal settings when updating role and location filters.
  * Added clearer handling for invalid user and installation configuration errors.
  * Portal templates are now used only when no user configuration exists.
  * Improved handling of optional location filters.

* **Tests**
  * Added coverage for configuration validation, template fallback, custom settings preservation, and filter merging.


santifer:
Routing note, @moelomda — labelled `area:web`.

The web app in `web/` is owned by a separate maintainer who triages that subsystem, so this goes to their queue rather than sitting in mine. That's a handoff, not a hold: they review and merge `web/` PRs themselves, and they're active daily.

Not a comment on the change. I'm routing rather than reviewing so it doesn't wait behind the core backlog for something I wouldn't be the right person to judge anyway.

santifer (approved):
Approving this, @moelomda, and I owe you an apology for the eighteen days first. I commented on 14 August and then never wrote a verdict, which meant this sat in the bucket my own tooling calls "waiting on the author" while it was in fact waiting on me. A comment is not a disposition; I had been counting it as one.

**The bug is real and it is the one my own rules name.** The old handler was:

```js
try   { doc = yaml.load(fs.readFileSync(file, "utf8")) || {}; }
catch { doc = yaml.load(templates/portals.example.yml) || {}; }   // ← ENOENT and a YAML syntax error land here identically
...
atomicWriteWithBackup(file, yaml.dump(doc));                      // ← then writes the example over the user's file
```

One `catch` for two different worlds. A user whose `portals.yml` had a stray tab got the shipped example loaded as if they had no config, and then written back over their own — `tracked_companies` and every custom block gone from the file, with no error shown. I keep a standing rule that says exactly this ("a missing file is not a malformed file; never treat the second as the first") and the bug was sitting in a route I own.

One precision, so the record is accurate rather than dramatic: `atomicWriteWithBackup` did take a `.bak` first, so this was recoverable by someone who knew to look. It was still a silent replacement of a user-layer file, and the next free scan ran against example keywords instead of theirs.

**What I checked in your fix, beyond that it passes:**

- `ENOENT` is the *only* condition that seeds from the template. Every other read failure raises `read-failed` → 500, so a permissions problem can never masquerade as "you have no config yet". That is the actual fix; the 409 is the visible half.
- The `invalid-user-config` / `invalid-template` split is the right axis. A broken user file is the user's to repair (409); a broken shipped template is our installation being wrong (500). Folding both into one status would have hidden our own bug behind theirs.
- `isMapping` checks the prototype rather than just `typeof`, so a YAML document that deserializes to an array — or to something with an unusual prototype — is rejected instead of being spread into the merge.
- `mergePortalFilters` copies rather than mutates, and touches only `title_filter.positive` and `location_filter.allow`. Everything else in the document is carried through by construction, not by a list I would have to keep updated.
- Tests use real temp directories and real files, so they exercise the `ENOENT`-vs-parse-error distinction where it actually happens rather than against a mocked `fs`.

**One behaviour change worth naming, which I am accepting deliberately:** an empty `portals.yml` used to parse to `undefined`, get `|| {}`'d, and proceed; it now fails `isMapping` and returns 409. That is arguably a file "with no config" rather than a broken one. I prefer your version — refusing to write over a file we do not understand is the right default, and an empty file is genuinely ambiguous. If a user reports friction from it, the fix is a targeted "empty means unconfigured" branch, not a return to the old catch-all.

Rebased onto today's `main` and merging on green. Thanks for the fix, and for the eighteen days of patience that shouldn't have been necessary.

santifer:
Merged. Thank you, @moelomda — and thank you for the eighteen days of quiet, which were mine to fix, not yours to absorb.

This one mattered more than its size: a user with a stray tab in `portals.yml` was getting the shipped example written over their own config, and the only reason it was recoverable is that the writer happened to take a `.bak` first. Your fix makes "missing" and "malformed" two different worlds again, which is where they belong.

#2845 is next in my queue.

PR #1444, opened 2026-07-02
feat(paths): support external data directory via CAREER_OPS_ROOT environment override

SparshGarg999:
## Summary

This PR resolves **Issue #524: Support for external data directory**. It introduces environment variables `CAREER_OPS_ROOT` and `CAREER_OPS_DATA_DIR` to decouple user-layer data (personal configurations, profiles, portals list, databases, reports, and generated PDFs) from the system-layer codebase in the repository.

### Changes

1. **Path Resolution Utility (`path-resolver.mjs`)**:
   - Centralizes root folder resolution: checks if `process.env.CAREER_OPS_ROOT` or `process.env.CAREER_OPS_DATA_DIR` is set and resolves it absolutely.
   - Defaults to the repository root directory (`__dirname`) if neither is defined, preserving 100% backward compatibility.
2. **Javascript Script Separation**:
   - Updated core runner scripts (`doctor.mjs`, `cv-sync-check.mjs`, `dedup-tracker.mjs`, `followup-cadence.mjs`, `gemini-eval.mjs`, `generate-pdf.mjs`, `merge-tracker.mjs`, `normalize-statuses.mjs`, `scan.mjs`, `verify-pipeline.mjs`) to resolve User Layer files (CV, identity configuration, portals YAML, tracker files, outputs, and reports) relative to the dynamically overridden data directory.
   - Retained System Layer files (code scripts, localized modes, templates like `cv-template.html` / `states.yml`, and instructions) relative to the codebase.
3. **Go Dashboard TUI Override (`dashboard/main.go`)**:
   - Configured the default value of the `--path` CLI flag in the Go TUI dashboard to check the `CAREER_OPS_ROOT` / `CAREER_OPS_DATA_DIR` environment variables, automatically loading pipelines from the external directory when launched.
4. **Documentation updates**:
   - `README.md`, `CLAUDE.md`, and `DATA_CONTRACT.md` updated to document the new overrides.
5. **Unit Tests (`test-all.mjs`)**:
   - Added Section `20. Path resolution layer and overrides` validating:
     - Unset environment variables default correctly to the codebase root.
     - `CAREER_OPS_ROOT` and `CAREER_OPS_DATA_DIR` overrides are correctly detected and applied.
     - `doctor.mjs` checks target directories correctly under custom roots.

### Verification

Ran `node test-all.mjs --quick` (171/171 passed):
```
20. Path resolution layer and overrides
  ✅ getCareerOpsRoot() defaults to codebase root when environment variables are unset
[...]

PR #2844, opened 2026-08-14
fix: keep artifacts in tracker workspace

moelomda:
## What does this PR do?

Keeps outcome archives and generated-PDF metadata inside the workspace that owns the configured tracker. It replaces install-directory assumptions with the shared `resolveWorkspaceRoot()` and `resolvePdfIndexPath()` helpers, so root-layout trackers, `CAREER_OPS_TRACKER`, and `CAREER_OPS_PDF_INDEX` all resolve consistently.

## Related issue

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

## Verification

- `node tests/outcome.test.mjs` — passed, including root-layout and manifest-override regressions
- `node tests/generate-pdf-page-budget.test.mjs` — passed
- `node tests/application-artifacts.test.mjs` — passed
- `node tests/pdf-index-path.test.mjs` — passed
- All 174 discovered root test files except `tests/intake.test.mjs` passed

The complete Windows root harness remains blocked by the existing symlink-permission failure tracked in #2828; the failure is unrelated to this change.

## Summary by CodeRabbit

* **Enhancements**
  * PDF generation now supports trackers located at workspace roots or redirected paths.
  * Generated PDFs, manifests, and outcome archives use the correct workspace directories.
  * Workspace configuration, CV content, theme settings, and shared PDF indexes are applied consistently.
  * Explicit report IDs and batch rendering results are handled more reliably.
* **Bug Fixes**
  * Improved path resolution for tracker and manifest files.
  * Prevented outputs, manifests, and batch results from escaping the tracker workspace, including through symlinks.


Scott-Emberson (changes requested):
Thanks for this. Routing outputs, manifests, and outcome archives through resolveWorkspaceRoot() / resolvePdfIndexPath() instead of __dirname is the right fix, and the new tests earn their place: the symlink-escape case in generate-pdf-page-budget.test.mjs drives the exact realpath branch in isWorkspaceOutputPath (lexical check passes, canonical check catches the junction), and the .pdf-versus-.md distinction in the outcome test is a real control that fails loudly if CAREER_OPS_PDF_INDEX resolution breaks. Both passed on all three OSes.

One thing before it lands, and it is a one-liner in the tests.

1. tests/outcome.test.mjs:219-220 fails on macOS. The current run is red: `3802 passed, 1 failed`, and the single failure is `Root-layout outcome directory stays under workspace/data`. ubuntu and windows are green.

outcome.mjs derives outcomeDir from resolveWorkspaceRoot(appsFile), and appsFile comes through resolveTrackerPath() -> canonicalizeTrackerPath() -> realpathSync(). On macOS the temp root is mkdtempSync(join(tmpdir(), ...)), i.e. /var/folders/..., and /var is a symlink to /private/var. So rootResult.outcomeDir returns canonicalized under /private/var/..., while the test builds expectedDir from the raw rootLayoutDir (/var/...). The string compare at line 220 breaks.

The tell is the next check: `Outcome honors CAREER_OPS_PDF_INDEX` (line 221) passes, because readFileSync(join(expectedDir, 'submitted_cv.pdf')) follows the /var -> /private/var symlink to the same file. Check 1 compares strings and fails; check 2 dereferences a path and survives. That split is exactly what the macOS $TMPDIR symlink predicts.

The existing tests in this file avoid it because none of them compare outcomeDir as a string against a temp-derived path (they feed the returned res.outcomeDir straight into existsSync/readFileSync, e.g. lines 95, 166, 184). Test 10 is the first to do exact string equality, so it is the first exposed.

Fix, canonicalize the temp root at creation so both checks use the same spelling:

    import { realpathSync } from 'fs';   // add to the existing fs import
    const rootLayoutDir = realpathSync(mkdtempSync(join(tmpdir(), 'cops-outcome-root-layout-')));

That fixes the string compare and also removes the accidental luck from check 2, which currently only passes because the symlink gets followed. Worth doing: a hermetic test that turns macOS red on every run trains people to read a red macOS leg as noise, and that is how a real regression eventually slips through.

Land that and, from the tests side, this is good to go.

santifer:
Sorry for the slow first reply, @moelomda — your three `web/` PRs I routed to the web maintainer earlier tonight; this one is core, so it's mine.

Replacing install-directory assumptions with `resolveWorkspaceRoot()` is the right direction, and it's the same seam #2859 is working on from the PDF side. Worth knowing you're not alone in it: an artifact path that assumes the install directory breaks the moment someone points the tracker somewhere else, and that configuration exists (`CAREER_OPS_ROOT`, #1444) — which means the assumption is already wrong for those users and simply hasn't produced a complaint yet.

Not merged tonight (per-session merge ceiling), queued. Two things I'll look at when it's reviewed, flagged now so you can pre-empt them:

1. **What happens when the tracker workspace and the install directory are the same**, which is the majority case today — the change should be a no-op there, and a test asserting that is what makes the refactor safe to merge.
2. **#2859 (@luochen211) touches `generate-pdf.mjs` for a related boundary concern.** Two PRs converging on the same file from different directions; whichever lands second will need a rebase, and I'd rather tell you now than let you find it in a conflict.

santifer:
Good consolidation, @moelomda: outcome archives and PDF metadata now follow the tracker's workspace through the shared resolvers instead of assuming the install directory — the exact silent-split class the external-data-dir work left behind. Verified the web reads none of these surfaces before merging, so this is core-internal and clean. Merged 🚀

PR #2845, opened 2026-08-14
fix(web): self-host production fonts

moelomda:
## What does this PR do?

Replaces the web app's `next/font/google` build-time downloads with vendored latin WOFF2 assets loaded through `next/font/local`. It preserves the Inter and Instrument Serif families, includes each family's SIL OFL 1.1 license and source provenance, and makes production builds reproducible without internet access.

## Related issue

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

## Verification

- `npm test` — 149 web tests passed
- `npm run build` — passed with HTTP/HTTPS/ALL proxy variables forced to an unreachable local address
- `node --test web/tests/lib/fonts-local.test.mjs` — 3 provenance and binary-format regressions passed

The vendored assets total approximately 91 KB and add no runtime package dependency.

## Summary by CodeRabbit

* **New Features**
  * Added bundled Inter and Instrument Serif fonts for consistent typography without relying on external font services.
  * Preserved existing font weights, styles, and display behavior, including Instrument Serif italics.

* **Documentation**
  * Added font usage documentation, source references, and SIL Open Font License information.

* **Bug Fixes**
  * Improved font availability and reliability by loading the selected fonts locally.


santifer:
Routing note, @moelomda — labelled `area:web`.

The web app in `web/` is owned by a separate maintainer who triages that subsystem, so this goes to their queue rather than sitting in mine. That's a handoff, not a hold: they review and merge `web/` PRs themselves, and they're active daily.

Not a comment on the change. I'm routing rather than reviewing so it doesn't wait behind the core backlog for something I wouldn't be the right person to judge anyway.

santifer (approved):
Approving this too, @moelomda — same apology as on #2843 for the eighteen days, same cause: I commented on 14 August and never wrote a verdict.

**And your case is stronger than the one you made for it.** Your description says the change makes production builds "reproducible without internet access", which reads like a build-hygiene preference. It isn't. I went looking for how this app actually reaches a user:

`web/README.md:19` and `:58` — the way to run it is `npm run dev`, **on the user's own machine**. There is no prebuilt bundle we ship. And `next/font/google` fetches from `fonts.googleapis.com` / `fonts.gstatic.com` at build time.

So the people this excludes are not "someone doing an offline build". They are:

- anyone behind a corporate firewall that blocks Google font hosts,
- anyone on a genuinely offline machine — which is a strange thing to exclude from a product whose whole pitch is that your CV never leaves your laptop,
- and **users in mainland China**, where those hosts are unreachable. We maintain `modes/zh/oferta.md` and `modes/zh-TW/oferta.md`, and `zh` has had two sync passes since June, so those are not hypothetical users.

For that set, the app did not render badly. It did not start. That is a functional fix, which is the category that merges here, and I'd have prioritised it very differently in August had I read it that way.

**What I checked:**

- `display: "swap"` and both CSS `variable` names survive unchanged, so no CLS behaviour change and no cascade change — `--font-inter`, `--font-instrument-serif`, `--font-instrument-serif-italic` are what the rest of the app binds to.
- The variable Inter is declared `weight: "100 900"`, so the whole axis stays available rather than being pinned to one cut.
- 89 KB of WOFF2 total across the three files. Not a weight argument.
- Both OFL 1.1 licenses are vendored alongside the fonts, which is the part of "vendor a font" that is easy to skip and not optional.

**And the test file is the best part of the PR.** It checks the `wOF2` magic bytes rather than trusting the extension; it closes the loop in both directions (every asset exists *and* `fonts.ts` references that exact path); it asserts the license text is really there; and `assert.doesNotMatch(fontsSource, /next\/font\/google/)` is a regression lock, so the loader cannot quietly come back in a future refactor. That is asserting the property rather than its neighbour, and it is not what most font PRs do.

One thing I am accepting knowingly rather than as an oversight: vendored fonts do not update themselves, so if upstream Inter ships a fix we take it deliberately or not at all. Given the alternative is "the app does not start for some users", that is the right side of the trade — and the `wOF2` + path test means a bad swap fails loudly rather than silently.

santifer:
Merged — both of yours are in.

For the record of why this one mattered: your description called it "reproducible builds without internet access", and that undersold it. Users run this with `npm run dev` on their own machine, so `next/font/google` reaching out at build time meant the app did not start for anyone behind a firewall that blocks Google's font hosts — including mainland China, where we actively maintain two localized modes. Not a degraded render: a non-start.

Thank you for both, and for waiting eighteen days for a verdict that should have taken two.

PR #2848, opened 2026-08-14
fix(match-star): a non-English story bank scores 0 against a question in its own language

### LINKED ISSUE #2847: match-star: a non-English story bank scores 0 against a question in its own language
### What

`tokenize()` in `match-star.mjs` strips with `[^a-z0-9\s]`, so every non-Latin character is deleted:

```
Led a payments migration       ["led","a","payments","migration"]
Возглавил миграцию платежей    []
भुगतान माइग्रेशन का नेतृत्व      []
支払い移行を主導                 []
```

Every scoring path runs through it — story title/theme, action/result, tags, the question, and the JD. So a story bank written in the user's own language produces empty token sets on both sides and scores **0** against a question about the very same thing:

```
English story vs English question :  10
Russian story vs Russian question :   0
Hindi   story vs Hindi   question :   0
```

`interview-prep/story-bank.md` is user content and `language.output` is explicitly configurable, so for anyone not working in English `match-star` is not degraded — it's **inert**, and it fails by returning "no good match", which reads as "your story bank doesn't cover this" rather than "I can't read your story bank."

Same class as #2518 (invite-match returned zero candidates for non-Latin companies) and #2781 (role titles tokenizing to nothing).

### Fix

Widen to `[^\p{L}\p{M}\p{N}\s]` with the `u` flag. `\p{M}` is not optional: without it Devanagari matras become spaces and shatter `भुगतान` into fragments that match nothing — the exact mistake caught during #2781.

`tokenize()` does no NFD folding, so unlike #2781 there's no accent-folding to preserve; the character class is the whole fix. English tokenization is byte-identical.

### Known limitation, not addressed

Inflected languages still match weakly — Russian `платежей` (genitive) doesn't token-match the tag `платежи` (nominative), so the score lands low rather than high. That's morphology and needs stemming, not a character class. The fix takes it from "cannot match at all" to "matches, imperfectly"; I'd rather name that than imply it's solved.

PR incoming.

abankar1:
Closes #2847.

`tokenize()` strips `[^a-z0-9\s]`, deleting every non-Latin character. Every scoring path runs through it — story title/theme, action/result, tags, the question, the JD — so both sides come back empty and score 0:

```
                                    BEFORE   AFTER
English story vs English question     10       10
Russian story vs Russian question      0        2
Hindi   story vs Hindi   question      0        9
unrelated Russian question             0        0     ← control
```

`interview-prep/story-bank.md` is user content and `language.output` is configurable, so for a non-English user this was **inert, not degraded** — and it failed as *"no good match"*, which reads as "your story bank doesn't cover this" rather than "I can't read your story bank."

Third instance of the class behind #2518 (invite-match: zero candidates for non-Latin companies) and #2781 (role titles tokenizing to nothing).

## Fix

`[^\p{L}\p{M}\p{N}\s]` with the `u` flag. **`\p{M}` is not optional** — without it Devanagari matras become spaces and shatter `भुगतान` into fragments matching nothing. That's the mistake I made and caught during #2781, so it's pinned by a test here rather than by memory.

Unlike #2781 there's no NFD folding in this tokenizer, so the character class is the whole fix and English tokenization is byte-identical:

```
'Led a payments migration (2024) — 40% win!'  ->  ["led","a","payments","migration","2024","40","win"]
```

## Tests

Four assertions, three of which fail on `main`:

```
❌ match-star: non-Latin text still tokenizes to nothing
❌ match-star: same-language match still scores 0 (ru=0, hi=0)
✅ match-star: an unrelated same-language question still scores 0   ← passes trivially when everything is []
❌ match-star: matras stripped — token came back as []
```

That third one is deliberate: it's the assertion that stops the widening turning into "everything matches everything," and it's the one that *would* pass either way, so I checked it discriminates by other means.

`node test-all.mjs --quick` → **3655 passed, 0 failed** (3652/3 failed with the fix reverted).

## Known limitation, named not buried

Inflected languages still match weakly: Russian `платежей` (genitive) doesn't token-match the tag `платежи` (nominative), so Russian scores 2 where the English equivalent scores 10. That's morphology and needs stemming, not a character class. This takes it from "cannot match at all" to "matches, imperfectly" — I'd rather say so than imply it's solved.

2 commits (fix + test).

## Summary by CodeRabbit

- **New Features**
  - Story matching now supports words written in non-Latin scripts, including Russian and Hindi.
  - Unicode letters, combining marks, and digits are preserved during text processing while punctuation is removed.

- **Bug Fixes**
  - Improved matching accuracy for same-language searches.
  - Prevented unrelated queries from producing incorrect matches.
  - Preserved Devanagari vowel marks during matching.


santifer:
Segundo de tu serie no latina en un día, y el diagnóstico vuelve a estar completo antes de que nadie lo pida.

Sobre el rojo de Windows que traías: **no era tuyo.** Lo comprobé en vez de asumirlo, porque "será el flake de siempre" acierta hasta el día que tapa algo real:

```
fix(pipeline-lock) #2838 entró en main:  14-ago 08:55:42
el run de tu PR arrancó:                 14-ago 03:24
```

Cinco horas y media antes del arreglo. Tu rama corrió sobre una base que todavía tenía el bug de concurrencia, así que la puse sobre la base buena en vez de pedirte un rebase: verde en las tres plataformas a la primera.

Gracias por los tres de esta familia (#2782, #2848, #2850). Ninguno salió de un radar nuestro: salieron de alguien usando el producto en su idioma, que es exactamente la superficie que peor vigilamos. Merged 🚀

PR #2851, opened 2026-08-14
Don't let an unscoreable (N/A) re-eval overwrite a real score

### LINKED ISSUE #2803: merge-tracker: an unscoreable re-evaluation (N/A) overwrites a real score
### What happens

A re-evaluation that produced no score overwrites a real score on the tracker. `parseScore('N/A')` is `0`, and since #2411 made re-evaluations write through in both directions, the merge cannot distinguish "this evaluation has no score" from "this evaluation scored zero". It reports the overwrite as a downgrade and archives the TSV as merged.

### Reproduce

`data/applications.md`:

```
| # | Date | Company | Role | Score | Status | PDF | Report | Notes |
|---|------|---------|------|-------|--------|-----|--------|-------|
| 4 | 2026-06-01 | DoorDash | Senior Associate, Finance & Strategy | 4.0/5 | Evaluated | ❌ | [4](reports/4-dd.md) | good |
```

`batch/tracker-additions/4-dd.tsv` — a re-eval whose fetch failed, so it carries the `N/A` sentinel the contract documents for an unscoreable row:

```
4	2026-06-25	DoorDash	Senior Associate, Finance & Strategy	Evaluated	N/A	❌	[4](reports/4-dd.md)	fetch failed
```

```
$ node merge-tracker.mjs
🔽 Update: #4 DoorDash — Senior Associate, Finance & Strategy (4→0) — DOWNGRADE, re-eval scored lower
📊 Summary: +0 added, 🔄1 updated, ⏭️0 skipped
```

Result — the `4.0/5` is gone:

```
| 4 | 2026-06-25 | DoorDash | Senior Associate, Finance & Strategy | N/A | Evaluated | ❌ | [4](reports/4-dd.md) | good. Re-eval 2026-06-25 (4→0): fetch failed |
```

Verified on `b52d3d3` (current `main`) with a clean checkout.

### Why it matters

The tracker is gitignored and no `.bak` is written, so the prior score is unrecoverable from the tracker itself. The row still links the superseded report, but the "Superseded report" marker only fires when `oldReportNum !== reportNum` — a re-eval of the *same* report number, which is the common case for a refetch, leaves no trace of the score that was there.

It reaches every tier, not just one: the run above matched on the report number. `AGENTS.md` also names `N/A`, `—` and `-` as the recognized score sentinels for rows added without an evaluation (#1799), so this is a documented value arriving through a documented path.

### Suggested direction

Treat an unscoreable score as **absent**, not as zero. `parseScore` returning `0` for both "no number present" and a genuine `0` is the root of it — the distinction wants to be visible to the caller (`null` vs `0`, or a companion `hasScore`). Then the update path can write everything else through — status, notes, report link, PDF, date — while leaving `score` alone when the incoming one is absent. That keeps #2411's intent (a lower score is newer information and must land) without letting a failed fetch erase a real evaluation.

I have no strong opinion on the shape and would rather not guess at it inside another PR — happy to send one if the direction sounds right.

### Context

Found while rebasing #1298. An earlier revision of that branch carried a guard against this, scoped to URL-matched rows. I removed it: the hazard is not URL-specific, and fixing it there would have hidden a general bug behind a new key while leaving the other tiers exposed. Filing it on its own instead.

gunjanjaswal:
Fixes #2803.

## The bug

A re-evaluation that produced no score overwrites a real score on the tracker, unrecoverably.

`parseScore('N/A')` is `0`. Since #2411 made re-evaluations write through in both directions, a re-eval whose fetch failed — carrying the `N/A` sentinel that AGENTS.md documents for an unscoreable row (#1799) — reads as a genuine zero, trips the downgrade path, and writes the sentinel over the row's real score. The TSV is then archived as merged. There's no way back: the tracker is gitignored with no `.bak`, and the superseded-report marker only fires when the report number changes, which a same-report refetch doesn't.

Reproduced on a clean checkout, a `4.0/5` row against an `N/A` refetch:

```
🔽 Update: #4 DoorDash — Senior Associate (4→0) — DOWNGRADE, re-eval scored lower
📊 Summary: +0 added, 🔄1 updated, ⏭️0 skipped

| 4 | 2026-06-25 | DoorDash | Senior Associate | N/A | Evaluated | ❌ | [4](reports/4-dd.md) | good. Re-eval 2026-06-25 (4→0): fetch failed |
```

The `4.0/5` is gone.

## The fix

Add `isUnscoreable()` for the documented no-score sentinels (`N/A` / `—` / `-`). When a re-eval is unscoreable **and** the existing row holds a real score, skip it — log it, count it, and leave the score, report and PDF as they stand. "No score" is not "scored zero".

The carve-out is deliberately narrow. It only triggers when a real score would be lost:

- Incoming real score (higher, lower, or equal) → unchanged write-through, including the intentional lower-score path from #2411.
- Incoming `N/A` onto a row that has no score yet → nothing to lose, so the normal path still refreshes date/notes/report.
- Incoming `N/A` onto a real score → skipped, the score preserved.

Same input, after:

```
⏭️  Skipping 4-dd.tsv: re-eval of #4 DoorDash — Senior Associate produced no score (N/A); keeping 4.0/5
📊 Summary: +0 added, 🔄0 updated, ⏭️1 skipped

| 4 | 2026-06-01 | DoorDash | Senior Associate | 4.0/5 | Evaluated | ❌ | [4](../reports/4-dd.md) | good |
```

## Tests

`tests/merge-tracker.test.mjs` gains two assertions in the re-eval section: an `N/A` re-eval keeps the existing `4.0/5`, and it's counted as skipped rather than a downgrade. Both fail on `main` and pass with the fix. The existing lower / higher / equal write-through cases still pass, so the numeric score path is untouched. Full file: 38 pass, 0 fail.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented re-evaluations with unavailable scores such as “N/A”, “—”, or “-” from replacing existing scored results.
  * Preserved associated reports and PDFs when an unscored re-evaluation is skipped.
  * Ensured skipped re-evaluations are not counted as updates or downgrades.

* **Tests**
  * Added regression coverage for preserving valid scores during unscored re-evaluations.

### DIFF
--- a/merge-tracker.mjs
+++ b/merge-tracker.mjs
@@ -367,6 +367,22 @@ function parseScore(s) {
 }
 
 /**
+ * True when a Score cell is a documented "no score" sentinel rather than a
+ * number. AGENTS.md names `N/A`, `—` and `-` as the values a row carries when it
+ * was added without an evaluation, or when a re-eval could not be scored (#1799).
+ *
+ * parseScore() maps all of these to 0, so the merge path cannot tell them apart
+ * from a genuine zero on its own — callers that must not treat "no score" as
+ * "scored zero" ask here first (#2803).
+ *
+ * @param {string} s - Raw score cell.
+ * @returns {boolean}
+ */
+function isUnscoreable(s) {
+  return ['—', '-', 'N/A'].includes(String(s).replace(/\*\*/g, '').trim());
+}
+
+/**
  * Load the optional generated-PDF manifest.
  *
  * data/pdf-index.tsv is gitignored and only exists after generate-pdf.mjs has
@@ -1031,6 +1047,22 @@ for (const file of tsvFiles) {
   }
 
   if (duplicate) {
+    // An unscoreable re-eval (N/A/—/-) carries no score to write through. Its
+    // score parses to 0, so the downgrade path below would read it as a genuine
+    // zero and overwrite a real score with the sentinel — unrecoverably, since
+    // the tracker is gitignored and no .bak is written (#2803). A failed fetch is
+    // "no new score", not "scored zero": leave the row's score, report and PDF as
+    // they stand. When the row itself is not yet scored there is nothing to lose,
+    // so let the normal path refresh its date/notes/report from the re-eval.
+    if (isUnscoreable(addition.score) && !isUnscoreable(duplicate.score)) {
+      console.log(
+        `⏭️  Skipping ${file}: re-eval of #${duplicate.num} ${addition.company} — `
+        + `${addition.role} produced no score (${String(addition.score).trim()}); `
+        + `keeping ${duplicate.score}`,
+      );
+      skipped++;
+      continue;
+    }
     const newScore = parseScore(addition.score);
     const oldScore = parseScore(duplicate.score);
 
--- a/tests/merge-tracker.test.mjs
+++ b/tests/merge-tracker.test.mjs
@@ -211,6 +211,30 @@ try {
   } else {
     fail(`equal-score re-eval mishandled: ${sameRow.trim()} | ${same.output.trim()}`);
   }
+
+  // --- An unscoreable (N/A) re-eval must NOT overwrite a real score (#2803) ---
+  // parseScore('N/A') is 0, so a re-eval that failed to fetch used to read as a
+  // genuine zero, trip the downgrade path above, and overwrite the real score
+  // with the sentinel — unrecoverably, since the tracker is gitignored and no
+  // .bak is written. "No score" is not "scored zero": the row must be left as is.
+  const NA_SEED = '| 4 | 2026-06-01 | DoorDash | Senior Associate, Finance & Strategy | 4.0/5 | Evaluated | ❌ | '
+    + '[4](../reports/4-dd.md) | good |\n';
+  const naReeval = runMergeDetailed({
+    '4-dd.tsv': '4\t2026-06-25\tDoorDash\tSenior Associate, Finance & Strategy\tEvaluated\tN/A\t❌\t[4](reports/4-dd.md)\tfetch failed\n',
+  }, { rows: NA_SEED });
+  const naRow = naReeval.tracker.split('\n').find(l => /DoorDash/.test(l)) || '';
+
+  if (/\|\s*4\.0\/5\s*\|/.test(naRow) && !/\|\s*N\/A\s*\|/.test(naRow)) {
+    pass('an unscoreable (N/A) re-eval keeps the existing real score (#2803)');
+  } else {
+    fail(`N/A re-eval overwrote the real score: ${naRow.trim()}`);
+  }
+
+  if (/⏭️1 skipped/.test(naReeval.output) && !/🔄1 updated/.test(naReeval.output) && !/DOWNGRADE/.test(naReeval.output)) {
+    pass('an unscoreable re-eval is skipped, not applied as a downgrade (#2803)');
+  } else {
+    fail(`N/A re-eval was not skipped cleanly: ${naReeval.output.trim()}`);
+  }
 } catch (e) {
   fail(`merge-tracker.mjs tests crashed: ${e.message}`);
 }
--- a/tests/merge-tracker.test.mjs
+++ b/tests/merge-tracker.test.mjs
@@ -247,10 +247,17 @@ try {
   const naOntoNa = runMergeDetailed({
     '6-globex.tsv': '6\t2026-06-25\tGlobex\tData Eng\tEvaluated\tN/A\t❌\t[6](reports/6-globex.md)\trefetch, still no score\n',
   }, { rows: NOSCORE_SEED });
-  if (/🔄1 updated/.test(naOntoNa.output) && !/⏭️1 skipped/.test(naOntoNa.output)) {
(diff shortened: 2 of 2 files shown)


gunjanjaswal:
Good call — expanded the coverage. The guard checks `isUnscoreable()` which already spans all three sentinels, so the test now loops over `N/A`, `—` and `-`. Each case also asserts the row keeps its own report link (the re-eval carries a different report number, and the row must not adopt it). Added the other side too: a sentinel re-eval of a row that's itself unscored has nothing to lose, so it still writes through rather than being skipped — that keeps the #1799 backfill contract intact. 40 pass, 0 fail.

gunjanjaswal:
Fair point — counters can pass while the row quietly fails to refresh. Added assertions on the `naOntoNa` row itself: it now checks the re-eval's date (2026-06-25), the repointed report link, and that the notes append `refetch, still no score` after the existing `pending eval` (kept first, per mergeNotes). Still 40 pass.

Scott-Emberson (approved):
Owner review for `/tests/`. Our CODEOWNERS route here is `tests/merge-tracker.test.mjs`, and it holds up.

The new block adds four assertions to the re-eval section. A three-sentinel loop (`N/A` / `—` / `-`) proves a failed-fetch re-eval leaves the seeded `4.0/5` and its `[4]` report link in place and counts the run as one skip, and a fourth case proves the guard stays narrow: a sentinel re-eval onto an already-unscored row still writes through rather than being skipped, since there is no real score to lose.

The controls against a false pass are real, not decorative. The re-eval TSV carries report `[9]` against the seed's `[4]`, so the `reportKept` check fails the moment the row wrongly adopts the re-eval's link, and the narrowness case flips if the guard ever over-fires. It is hermetic the way this repo asks: a throwaway `mkdtempSync` workspace, the `CAREER_OPS_TRACKER` / `CAREER_OPS_ADDITIONS` seams, cleanup in `finally`, no network.

Validation: I ran the file against this head offline and got 40 passed, 0 failed. Then I reverted `merge-tracker.mjs` to its pre-fix state and re-ran: all three sentinel assertions fail with the data loss reproduced (the `4.0/5` overwritten by the sentinel, the report re-pointed `[4]` to `[9]`, marked a `4→0` downgrade), while the narrowness case keeps passing since it never triggered the bug. The tests have teeth. The `merge-tracker` block is also green in the macOS and Ubuntu CI logs.

One optional follow-up, not a blocker: the already-unscored case asserts only the counters, not that the row took the re-eval's date, notes, and report. The three-sentinel loop already covers the row-content path, so this is a nice-to-have on top of it, and it lines up with the one CodeRabbit comment on the PR.

Approving for our area. The red `test (windows-latest)` check is `tests/agent-inbox-tests.mjs` failing on a lock race, a file this PR does not touch, so it is not from this change. The `merge-tracker.mjs` fix reads correct from context, but that file is not ours to judge.

gunjanjaswal:
Thanks Scott, and especially for reverting the fix to check the assertions actually bite — that's the part I'd worry about most.

Good timing on the optional follow-up: it landed a couple of minutes after you started reviewing, in 4406364. The unscored-row case now asserts the row itself, not just the counter — the re-eval's date (2026-06-25), the repointed [6] report link, and the notes appending `refetch, still no score` after the existing `pending eval`. Same thing CodeRabbit flagged, so that thread's closed too. Still 40 pass.

And agreed the windows-latest red is the agent-inbox lock race, unrelated to this.

santifer:
This is a data-loss bug and the mechanism is nastier than the title suggests, @gunjanjaswal:

> `parseScore('N/A')` is `0`. Since #2411 made re-evaluations write through in both directions, a re-eval whose fetch failed — carrying the `N/A` sentinel that AGENTS.md documents for an unscoreable row — reads as a genuine zero, trips the downgrade path, and overwrites a real score unrecoverably.

Three correct pieces composing into a wrong outcome: the sentinel is documented, the parse is a reasonable default, and bidirectional write-through was a deliberate fix. Nothing here is a mistake in isolation, which is why it survived — **`N/A` and `0.0` mean opposite things and the type system flattens them into the same number.** Same shape as everything I've been fixing tonight: an unknown that gets coerced into a value instead of staying unknown.

And the consequence is the worst kind: silent, irreversible, and it hits precisely when the network hiccuped rather than when the role got worse.

Second one of yours today — #2820 merged earlier. Not merged tonight (per-session merge ceiling), but this is high in the queue: a fix for silent overwriting of the user's own evaluation data ranks above almost everything else open.

gunjanjaswal:
That's exactly it — `N/A` is "we don't know," `0.0` is "we know, and it's zero," and flattening the unknown into a value is what did the damage. Thanks for framing it that way, and no rush on the merge ceiling — glad it's queued.

santifer:
This is exactly how a data-loss fix should arrive, @gunjanjaswal: the guard fires only when a real score would be lost, the sentinel-over-sentinel path keeps the documented backfill contract, and the test proving the row keeps its own report link (seeding [4], re-eval carrying [9]) is the detail most people would have skipped. Sorry this one waited longer than it deserved. Merged 🚀

gunjanjaswal:
Thanks for merging, and for digging into the mechanism the way you did. Glad the unscoreable-vs-zero distinction is nailed down now.

PR #2857, opened 2026-08-14
fix(update): reconcile .gitignore on apply so new ignore rules reach existing installs

### LINKED ISSUE #2756: `.gitignore` is not in the update manifest, so new ignore rules never reach an existing install
`update-system.mjs` does not ship `.gitignore`. It is in neither `SYSTEM_PATHS` (286 entries) nor `BOOTSTRAP_PATHS` (27), and the literal string `gitignore` does not appear anywhere in the file. Checked at v1.26.0.

The consequence is that every ignore rule added upstream stays upstream. An install keeps whatever `.gitignore` it happened to clone with, forever, while continuing to receive the features and tests that assume the newer rules exist.

I want to be clear up front that this is not a case of rules being forgotten. They are added carefully and promptly, alongside the features that need them. The gap is purely in delivery.

#### Evidence, three releases

**1. A rule added in June has still not arrived.**

`data/follow-ups.md` was added to `.gitignore` in f042835 on 2026-06-30. I then applied v1.17.0, v1.18.0 and v1.22.0. The rule was absent from my `.gitignore` at every one of those commits. Three updates after the fix, still not delivered.

**2. v1.25.0 shipped a test for rules it did not ship.**

`tests/user-layer-gitignored.test.mjs` asserts that timestamped backups and the test sandbox are ignored. The rules it checks for (`*.bak*`, `.tmp-script-test-*/`) arrived only in `.gitignore`, so the test landed and the fix did not. Six immediate failures on a current install, including this probe from your own test file:

```
cv.md.bak-2026-08-05T16-55-08-641Z is NOT git-ignored
```

A plain `*.bak` glob does not match the timestamped shape. So the exact user-layer exposure the test was written to prevent was live on an install that had applied every release.

**3. v1.26.0 repeated it.**

`batch/batch-state-recovery.d/` was added upstream and did not arrive.

#### Why it is worth fixing rather than documenting

These rules are the mechanism protecting the user layer. `data/*`, `interview-prep/*`, `writing-samples/*`, `*passport*`, `**/.env*.local` and the rest are what stand between a reflexive `git add .` and someone publishing their own CV, tracker or credentials. Anyone who forked publicly is relying on a file that silently stopped being maintained the moment they cloned.

The blanket `data/*` rule in 6164384 is a good example of the problem in miniature. Its comment says the point is that a new generated file is "covered for every user automatically, with no per-filename .gitignore update needed". That reasoning is right, and the fix itself cannot reach the users it was written for.

#### Why the obvious fix is probably wrong

I assume the exclusion is deliberate. `.gitignore` is a file users legitimately edit, and adding it to `SYSTEM_PATHS` would overwrite their own rules on every update. That trade seems worse than the current state.

Some options that avoid it, in rough order of how much I would want them as a user:

1. **Append absent rules rather than replace the file.** Parse both, add any upstream line not already present, never remove. Preserves user additions, converges over time. Comment blocks make this fiddlier than it sounds.
2. **Warn in `doctor.mjs`.** Diff the local `.gitignore` against the release and report missing rules without touching anything. Zero risk, puts a human in the loop, costs one line of output.
3. **Ship a separate `.gitignore.d/` fragment or a `core.excludesFile` include** that the updater does own, leaving the user's `.gitignore` untouched entirely.
4. **Do nothing, but document it** in the update notes so users know to diff it themselves.

I lean toward 2 as the cheap immediate step and 1 or 3 as the real fix, but this is a design call rather than a bug with one right answer.

#### Reproducing

```bash
grep -c gitignore update-system.mjs        # 0

BLOB=$(git ls-tree upstream/main -- .gitignore | awk '{print $3}')
git cat-file -p "$BLOB" > /tmp/upstream-gitignore
comm -23 <(grep -v '^#' /tmp/upstream-gitignore | grep -v '^[[:space:]]*$' | sort -u) \
         <(grep -v '^#' .gitignore | grep -v '^[[:space:]]*$' | sort -u)
```

On an install that has applied every release, that diff should be empty and is not.

One aside that cost me time: `git cat-file -p upstream/main:.gitignore` silently returns zero bytes in Git Bash on Windows, which produces a convincing false "no delta". Resolving the blob hash first, as above, avoids it.

eliador90:
## What does this PR do?

Makes `update-system.mjs apply` reconcile `.gitignore` instead of skipping it, by appending the system-owned rules an install is missing and leaving every other line alone. Adds a test that a user-authored rule survives an update, and moves `.gitignore` out of the coverage validator's generic exclusion list into its own documented group.

## Related issue

Fixes #2756

## Type of change

- [x] Bug fix
- [ ] New feature

## The bug

`.gitignore` is in neither `SYSTEM_PATHS` nor `BOOTSTRAP_PATHS`, so no ignore rule added upstream has ever reached an existing install. 43 commits touched it in the last 90 days and none of them shipped.

The consequence is sharper than "new rules are missing". Reconciling a real April 2026 `.gitignore` against today's appends **62** rules, and `cv.md` is among them. An install last updated before 2026-04-19 has the candidate's CV itself unignored, one reflexive `git add .` from a public fork.

`tests/` is in `SYSTEM_PATHS` and so is `AGENTS.md`, so an update ships `tests/user-layer-gitignored.test.mjs`, ships the newer `AGENTS.md` whose User Layer list that test reads, and withholds the `.gitignore` the test then checks against. The repo delivers the assertion and keeps the thing being asserted, which is why this never showed up as a red test.

## The approach

`.gitignore` cannot simply join `SYSTEM_PATHS`. It is the one system file users also write to, and the raw checkout that ships every other path would delete their rules, which is a worse bug landing the same silent way.

So it uses the append-if-missing shape `agent-inbox.mjs:ensureGitignored()` already applies to its own single rule, generalized to the whole upstream rule set:

- Append only the system-owned patterns an install lacks, in upstream order.
- Idempotent by **pattern presence**, not by position, so a user can reorder or relocate the appended lines and they will not come back.
- Carry each rule's own rationale comment across with it, deduplicated, since several of those comments are the only record of which paths hold PII.
- Compare normalized, emit verbatim, so a pattern whose trailing space is backslash-escaped is not corrupted on the way in.
- Preserve the file's dominant line ending, so a CRLF checkout does not show the whole file as changed.

Deliberately append-only. A rule upstream removed or rewrote leaves the superseded line in place, because a stale system rule and a user rule of the same shape are indistinguishable. A redundant ignore rule is harmless; deleting a user's is not.

## Two details worth your eye

**The reconciled file is kept out of `updated` when `.gitignore` was already dirty.** `updated` is both the rollback pathspec and the commit pathspec, and `revertPaths()` runs a bare `git checkout HEAD -- <path>` whose `protectedPaths` guard only covers newly *added* files. Without this guard an abort would hard reset `.gitignore` and destroy the user's uncommitted rules, and a success would sweep their edit in under the update's commit message, which is the same thing #2337 established for system files. This is a routine state rather than an exotic one: `ensureGitignored()` writes to `.gitignore` without committing it. The rules are live on disk either way, which is all that ignoring actually requires.

**`validate-system-paths-coverage.mjs` is part of the fix, not incidental.** It listed `.gitignore` in an undifferentiated `EXCLUDES` array next to `renovate.json` and `.editorconfig`. Those genuinely never reach an install and this one now does, so reading it as the same kind of exclusion is how the gap stayed invisible in the one tool built to catch coverage gaps. It moves to its own named group with the reasoning and a pointer to `reconcileGitignore()`.

## Testing

New `tests/gitignore-reconcile.test.mjs`, 24 assertions. The headline one is the case you asked for: a user-authored rule survives, and the original file is a prefix of the result, so nothing above the appended block was rewritten. Also covers idempotency across two passes, byte-identical no-op when nothing is missing, moved and reordered rules not being re-added, comment carry-across and dedup, negation ordering, CRLF preservation, missing trailing newline, verbatim emission of an escaped trailing space, empty local file, and the shipped `.gitignore` reconciled against itself being a no-op.

`node test-all.mjs` passes, with one unrelated pre-existing exception: `tests/intake.test.mjs` fails on my Windows machine with `EPERM` on `symlinkSync` (needs Developer Mode). It fails identically on a pristine `upstream/main` checkout, so it is environmental and not from this change. CI on Linux should be clean.

## Checklist

- [x] I ran `node test-all.mjs` and all tests pass (see the environmental exception above)

## Summary by CodeRabbit

* **New Features**
  * Improved `.gitignore` updates to append missing standard rules while preserving existing local rules, comments, ordering, line endings, and negations.
  * Automatically restores a missing `.gitignore` during system updates.
  * Keeps locally modified files unstaged while including unchanged files in update commits.
* **Bug Fixes**
  * Prevented duplicate rules and unnecessary file changes during repeated updates.
  * Update operations now report `.gitignore` reconciliation issues without stopping the entire update.

### DIFF
--- a/tests/gitignore-reconcile.test.mjs
+++ b/tests/gitignore-reconcile.test.mjs
@@ -140,3 +140,23 @@ const ok = (cond, msg) => (cond ? pass(msg) : fail(msg));
   eq(added.length, 0, 'the shipped .gitignore reconciled against itself adds nothing');
   eq(text, shipped, 'and is returned byte-identical');
 }
+
+// ── The upstream blob is read untrimmed ──────────────────────────────────────
+// reconcileGitignore()'s verbatim guarantee is only as good as its input. The
+// updater's general-purpose git helpers call .trim() on stdout, which is right
+// for SHAs and pathspecs and wrong for file content: it strips a significant
+// backslash-escaped trailing space off the blob's LAST line, which is exactly
+// where a freshly appended upstream rule sits. Every assertion above would
+// still pass with a trimming read, so this is a source-level guard, matching
+// tests/js-yaml-import-form.test.mjs and tests/source-no-nul-bytes.test.mjs.
+{
+  const src = readFileSync(join(ROOT, 'update-system.mjs'), 'utf-8');
+
+  const readCall = src.match(/const upstreamGitignore = (\w+)\(/);
+  if (!readCall) fail('could not find the upstreamGitignore read in update-system.mjs — update this test');
+  else eq(readCall[1], 'gitShowRaw', 'the upstream .gitignore is read with gitShowRaw, not a trimming helper');
+
+  const body = src.match(/function gitShowRaw\([^)]*\) \{[\s\S]*?\n\}/);
+  if (!body) fail('could not find gitShowRaw() in update-system.mjs — update this test');
+  else ok(!/\.trim\(\)/.test(body[0]), 'gitShowRaw() does not trim, so the blob reaches the reconciler byte-exact');
+}
--- a/tests/gitignore-reconcile.test.mjs
+++ b/tests/gitignore-reconcile.test.mjs
@@ -31,7 +31,7 @@ const ok = (cond, msg) => (cond ? pass(msg) : fail(msg));
   for (const userRule of ['my-scratch-notes/', '*.secret']) {
     ok(lines.includes(userRule), `user rule survives: ${userRule}`);
   }
-  ok(text.startsWith(local.replace(/\s*$/, '')), 'the original file is a prefix of the result (nothing above was rewritten)');
+  ok(text.startsWith(local), 'the original file is a byte-for-byte prefix of the result (no line above was touched)');
   eq(added.join(','), 'data/*,!data/.gitkeep', 'both missing upstream rules were appended');
   ok(!added.includes('cv.md'), 'a rule already present is not duplicated');
 }
@@ -109,6 +109,18 @@ const ok = (cond, msg) => (cond ? pass(msg) : fail(msg));
   ok(text.split('\n').includes('secret\\ '), 'but written back verbatim, with its escaped trailing space intact');
   const again = reconcileGitignore(text, upstream);
   eq(again.added.length, 0, 'and the verbatim line is still recognized on the next pass (no re-add loop)');
+}
+
+// ── A local rule's significant trailing space survives verbatim ────────────────────────
+// The mirror of the case above: normalization is for MATCHING only. Trimming
+// the local file's tail would modify a line the user wrote, which is the one
+// thing this function promises never to do.
+{
+  const local = 'cv.md\nsecret\\ ';   // no trailing newline, escaped space is significant
+  const { text } = reconcileGitignore(local, 'cv.md\ndata/*\n');
+  ok(text.startsWith(local), 'the local file is preserved byte-for-byte, escaped trailing space included');
+  ok(text.split(/\r?\n/).includes('secret\\ '), 'and the user rule keeps its significant trailing space');
+  ok(text.split(/\r?\n/).includes('data/*'), 'while the missing upstream rule is still appended');
 }
 
 // ── An empty local file gets no leading blank lines ──────────────────────────
--- a/tests/gitignore-reconcile.test.mjs
+++ b/tests/gitignore-reconcile.test.mjs
@@ -1,0 +1,130 @@
+// tests/gitignore-reconcile.test.mjs
+//
+// .gitignore was absent from the updater's manifest for 43 releases' worth of
+// commits, so every ignore rule added upstream in that window reached this
+// repository and no existing install (#2756). The file cannot simply join
+// SYSTEM_PATHS: it is the one system file users also write to, and the raw
+// `git checkout` the update stage performs would delete their rules silently.
+//
+// reconcileGitignore() appends only what is missing. These tests pin both
+// halves of that contract: the system rules arrive, and nothing the user wrote
+// is modified, reordered or removed.
+
+import { readFileSync } from 'fs';
+import { join } from 'path';
+import { pass, fail, ROOT } from './helpers.mjs';
+import { reconcileGitignore } from '../update-system.mjs';
+
+console.log('\n🔁 .gitignore reconcile (append-if-missing)');
+
+const eq = (actual, expected, msg) =>
+  (actual === expected ? pass(msg) : fail(`${msg} — expected ${JSON.stringify(expected)}, got ${JSON.stringify(actual)}`));
+const ok = (cond, msg) => (cond ? pass(msg) : fail(msg));
+
+// ── The headline requirement: a user-authored rule survives an update ────────
+{
+  const local = ['cv.md', 'my-scratch-notes/', '*.secret'].join('\n') + '\n';
+  const upstream = ['cv.md', 'data/*', '!data/.gitkeep'].join('\n') + '\n';
+  const { text, added } = reconcileGitignore(local, upstream);
+  const lines = text.split('\n');
+
+  for (const userRule of ['my-scratch-notes/', '*.secret']) {
+    ok(lines.includes(userRule), `user rule survives: ${userRule}`);
+  }
+  ok(text.startsWith(local.replace(/\s*$/, '')), 'the original file is a prefix of the result (nothing above was rewritten)');
+  eq(added.join(','), 'data/*,!data/.gitkeep', 'both missing upstream rules were appended');
+  ok(!added.includes('cv.md'), 'a rule already present is not duplicated');
+}
+
+// ── Idempotency: a second pass is a byte-identical no-op ─────────────────────
+{
+  const local = 'cv.md\nmy-notes/\n';
+  const upstream = 'cv.md\ndata/*\n';
+  const first = reconcileGitignore(local, upstream);
+  const second = reconcileGitignore(first.text, upstream);
+  eq(second.added.length, 0, 'second pass appends nothing');
+  eq(second.text, first.text, 'second pass leaves the file byte-identical');
+}
+
+// ── A file that already has everything is untouched ──────────────────────────
+{
+  const local = '# mine\ncv.md\ndata/*\n';
+  const { text, added } = reconcileGitignore(local, 'cv.md\ndata/*\n');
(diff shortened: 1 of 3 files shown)


eliador90:
Both CodeRabbit findings were real and are fixed in 256184eb.

**Terminal whitespace of the local file.** Caught the mirror image of a bug I had already fixed on the upstream side. I emit upstream patterns verbatim and compare them normalized, but `localText.replace(/\s*$/, '')` was still trimming the local file's tail, so a user rule ending in a backslash-escaped space became `secret\` and stopped matching. That modifies a line the user wrote, which is precisely the thing this PR promises not to do. The local side now concatenates verbatim and only the separator is computed.

**Non-atomic write.** Also correct, and worse for this file than for most. `writeFileSync` opens with `O_TRUNC`, so a crash or I/O error partway through leaves `.gitignore` empty or half-written, which un-ignores whatever the lost portion covered. A failed update would produce exactly the exposure the file exists to prevent, silently, which is the same failure shape as the original bug. Now writes to a temp file on the same filesystem and renames, matching `discover-ats.mjs:850` and `followup-seed.mjs:371`.

Tests: the prefix assertion is byte-for-byte rather than trim-tolerant now, plus a case where a local rule carrying a significant escaped trailing space survives an unrelated upstream append. 27 assertions.

tests/gitignore-reconcile.test.mjs
```diff
@@ -0,0 +1,130 @@
...
+
+// ── Patterns are emitted verbatim, matched normalized ────────────────────────
+// A trailing space is only significant in .gitignore when backslash-escaped.
+// Matching has to normalize (so indentation drift does not cause a re-add), but
+// writing back the normalized form would corrupt the pattern.
+{
+  const upstream = 'cv.md\nsecret\\ \n';
+  const { text, added } = reconcileGitignore('cv.md\n', upstream);
+  ok(added.includes('secret\\'), 'the escaped-space pattern is matched in normalized form');
+  ok(text.split('\n').includes('secret\\ '), 'but written back verbatim, with its escaped trailing space intact');
```
Scott-Emberson, line 109: This block is the strongest test in the file: it proves an upstream `secret\ ` survives verbatim and is still recognized on the next pass, which is exactly the escaped-space handling that matters. The gap is that it only exercises the pattern coming from *upstream*. There is no case where the user's own last line is `secret\ ` and an update appends a rule.

That path is the one that breaks. Measured on this head:

```
local    = 'cv.md\nsecret\ \n'   // user's own rule, last line
upstream = 'cv.md\ndata/*\n'      // appends data/*
result   -> the user's line comes back as 'secret\', escaped space stripped
```

The cause is `update-system.mjs`'s `head = localText.replace(/\s*$/, '')`, which trims the final line's significant escaped space before appending. So the reconciler protects the appended pattern here but mangles the user's identical one when it sits last, in a change whose whole point is never touching a user's rule.

One case closes it: local ending in `secret\ ` plus an appended rule, asserting the user's line is byte-identical in the result. A test that promises "user rules survive" but skips the single input where they don't lets the next reader trust a guarantee the code doesn't keep. This is the test half; the `replace` fix itself is in `update-system.mjs`, which is santifer's to weigh.

Scott-Emberson (approved):
Re-reviewed at 256184e on tests/gitignore-reconcile.test.mjs. My one finding is addressed on both sides. The exact escaped-space case I asked for is now there (lines 114-122): a local file ending in a significant escaped trailing space with an appended upstream rule, asserting byte-identity. And the headline assertion was tightened from the trim-tolerant form I flagged to byte-for-byte (line 33). That is a real control, not a false pass: it only holds because update-system.mjs concatenates the local text verbatim. Hermetic (pure-function calls, one read-only ROOT read), same helpers as its neighbours. CI green on all three OSes.

Approving from our /tests/ scope. The atomic-write wrapper in update-system.mjs is your area; it looks wired in on both apply branches.

santifer:
@luochen211 — this one has a concrete user report behind it now, which changes it from a tidiness fix to a bug fix. Worth stating, because it's the strongest of the eleven you opened today.

#2881 (@jaybee2184) reports the updater cascading `.bak` → `.bak.bak` → `.bak.bak.bak` on repeated runs. Traced it, and it lands on exactly the gap this PR closes:

```
$ git log -1 --date=short -S'*.bak*' -- .gitignore
bfaa7a69  2026-08-08  chore(gitignore): ignore timestamped user-layer backups (#2543)

$ grep -n "'.gitignore'" update-system.mjs
(nothing — .gitignore is not in SYSTEM_PATHS)
```

The `*.bak*` rule landed on 8 Aug and **cannot reach an install that already exists**. On an older checkout the `.bak` files stay tracked, so they appear in the locally-modified-system-files diff, so the updater backs up the backups. Every run, one level deeper.

That's a general shape rather than one file's problem: **an ignore rule added upstream after someone installed is invisible to them forever**, and the symptom shows up somewhere unrelated to `.gitignore` — which is why it took a user's bug report to surface it rather than anyone noticing the rule hadn't propagated.

Two things I'd want the PR to answer, and I'm asking rather than assuming:

1. **Reconciling `.gitignore` means writing to a file the user may have edited.** It's not in the system layer today precisely because a user can legitimately add their own rules. Does your approach merge (append missing rules, keep theirs) or replace? Replace would be a data-loss path on a user-owned file.
2. **What happens to files that are already tracked** in an old install — adding the rule doesn't untrack `foo.md.bak`. Does the fix need a `git rm --cached` pass, or is the rule alone enough for the diff to stop listing them?

Not reviewing the diff properly tonight — I've hit the merge ceiling I hold myself to. But this one has evidence attached now, so it goes near the top next cycle. If you answer those two, that read gets a lot faster.

Separately, and meant usefully: **eleven PRs opened in one day is more than this repo can review well**, and one of them (#2868) had already been solved by a PR that merged tonight. Fewer, with the strongest first, gets you more merges than more — and #2857 is a good example of which one that would have been.

luochen211 (approved):
Reviewed the current head. The escaped-trailing-space regression is covered, the local .gitignore remains byte-preserved, and CI is green.

santifer:
First, a correction I owe you: my comment above addressed this PR to the wrong author. It's yours, @eliador90, as you graciously noted in #2756. The work deserves the right name on it, because this is how a fix to a co-owned file should be built: append-only because a redundant rule is harmless and deleting a user's is not, atomic write because a truncated .gitignore un-ignores exactly what it exists to protect, and the dirty-file case kept out of the rollback pathspec. Three releases of evidence plus tracing #2881 to the same gap made this the best-argued fix of the week. Merged! 🚀

PR #2858, opened 2026-08-14
fix(batch): keep JD word count machine-readable

luochen211:
Follow-up for #2492 and review feedback on #2816.

When FORCE_COLOR is set, Node can emit ANSI escapes from console.log even in command substitution. The existing digit sanitizer then turns the escape codes into extra digits, causing thin pages to be treated as full JDs.

Changes:
- use process.stdout.write for a color-independent integer
- add a regression test with FORCE_COLOR=3

Validation: node tests/batch-runner-jd-prefetch.test.mjs; git diff --check

## Summary by CodeRabbit

* **New Features**
  * Job descriptions now follow public redirects securely, with limits on redirect hops and protections against unsafe destinations.
  * Failed, blocked, malformed, or excessive redirects automatically use the existing fallback retrieval method.
  * Temporary files are cleaned up reliably in all prefetch outcomes.

* **Tests**
  * Expanded coverage for redirect validation, fallback behavior, command failures, content checks, and cleanup.


Scott-Emberson (commented):
`tests/batch-runner-jd-prefetch.test.mjs`: the FORCE_COLOR regression pin is correct and fails on a revert to `console.log`, and it is robust to an ambient `NO_COLOR` since Node checks `FORCE_COLOR` first. Two small things. Line 22 imports `mkdtempSync` twice (`mkdtempSync` and `mkdtempSync as _mdt`); `_mdt` is never referenced, drop it. And optionally add one e2e case that runs `curlPrefetchBlock` with `FORCE_COLOR=3` on a ~79-word page and asserts the file is truncated, so the actual misclassification path (the ANSI escapes surviving the digit-strip as `337939`) is pinned end-to-end, not just the node snippet in isolation.

santifer:
The design decision worth naming, @luochen211, is the one in your comment:

> For JS-rendered boards the file will be thin (JS shell only), which the sufficiency check below detects — the file is then **truncated to 0 bytes** so the worker's Step-1 WebFetch fallback fires exactly as designed.

Truncating rather than deleting is the right call, and it's the kind of choice that looks arbitrary until it isn't: the downstream worker branches on an empty file, so leaving a thin JS shell in place would have it evaluate the shell as if it were the JD — confidently, with a real-looking score built on markup. Empty is a state the existing code already handles correctly; "technically present but useless" is not.

Not merged tonight (merge ceiling), and one thing worth a line in the PR body when it gets reviewed: what the sufficiency threshold is, and why that number. A JD can legitimately be short, and the cost of a wrong call in each direction isn't symmetric — falling back to WebFetch unnecessarily is slow, evaluating a JS shell is wrong.

santifer:
Follow-up to my earlier comment, @luochen211: **#2816 (@Zoubeir23) is the same fix**, opened 13 Aug 17:53 — about 14 hours before yours. Neither of you could have known, and linking duplicates before they cost someone an evening is on me.

I'm not deciding it on timestamps alone. Your truncate-to-0-bytes handling of the thin-JS-shell case is a real design decision and it's the thing I'd compare the two on, since evaluating a JS shell as a JD is much worse than a slow fallback. I've asked them the same question about their version.

Queued as a pair so they're decided together.

santifer:
You called it exactly, @luochen211: #2816 merged and its own tests went red on the first FORCE_COLOR machine they met (79 words counted as 337939, the 33/39 ANSI digits). Because main was broken I hot-landed the two-line output-channel fix in #3097 with your name on the diagnosis, rather than waiting on a rebase here.

What remains in this PR is yours and still wanted: the redirect loop with --max-redirs 0 that re-validates the destination guard on every hop closes a real gap (#2816 only checked the FIRST url; a redirect to a private address walks past the guard today), plus your FORCE_COLOR=3 regression test so this class can't return. Rebase onto current main (your word-count change will already be there; keep the rest) and this merges on green.

santifer:
This closes the root of the Aug 20 incident properly, @luochen211: your diagnosis found that FORCE_COLOR turns console.log output into ANSI-laced digits that the sanitizer reads as MORE digits (thin pages promoted to full JDs), my #3097 was only the tourniquet, and this is the actual fix — a color-independent stdout.write plus the FORCE_COLOR=3 regression test that makes the incident unrepeatable. Diagnosis and cure both yours. Merged! 🚀

PR #2859, opened 2026-08-14
fix(pdf): enforce workspace boundary in direct renderer

luochen211:
Follow-up for #2844.

The CLI path already rejected output escapes, but callers of the exported renderHtmlToPdf() helper could still write a PDF outside the tracker workspace. Enforce the same containment check before creating the destination directory or invoking Chromium.

Validation: node tests/generate-pdf-page-budget.test.mjs; git diff --check

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved workspace path handling for PDF generation and outcome artifacts.
  - Prevented output files from escaping the workspace, including through symbolic links.
  - Corrected manifest and PDF index resolution for redirected and root-level tracker workspaces.
  - Ensured configured PDF files and generated artifacts are stored in the expected workspace locations.

- **Tests**
  - Added coverage for custom manifests, redirected workspaces, root-layout trackers, and path-escape protection.


Scott-Emberson (commented):
`tests/outcome.test.mjs:219-220`, low severity, worth a fix to keep local runs green. Test 10 asserts `rootResult.outcomeDir === expectedDir`, but `expectedDir` is built from the raw `mkdtemp` path while `outcomeDir` comes back canonicalized through `realpathSync` (`resolveTrackerPath` -> `canonicalizeTrackerPath`). On any platform whose `tmpdir()` resolves through a symlink (macOS `/var` -> `/private/var`) this fails deterministically on a correct implementation, while staying green on Linux CI. Canonicalize the expected side: `const expectedDir = join(realpathSync(rootLayoutDir), 'data', 'outcomes', ...)` and add `realpathSync` to the fs import; the second check reads through `expectedDir`, so it inherits the fix.

The rest is solid: the symlink-escape boundary case in `generate-pdf-page-budget.test.mjs` can only reject via the new realpath canonical check, so it does not false-pass on the old lexical guard.

luochen211:
修复已补齐：PDF 输出路径现在在创建目录、启动 Chromium 和写入临时文件前校验；清理测试显式声明 fixture workspace。本地 quick suite 通过（3805 passed），最新 Linux/macOS/Windows、CodeQL、升级回归均通过，请重新审查。

Scott-Emberson (approved):
Re-reviewed at 1363e4cb. Within our scope only tests/outcome.test.mjs changed since my comment, and it is the exact fix I asked for: realpathSync is imported (line 3) and expectedDir is canonicalized (line 219), so rootResult.outcomeDir === expectedDir now survives symlinked and 8.3-short tmpdirs. That strengthens the assertion rather than loosening it. The other two owned files are byte-identical to what I already reviewed. The boundary sentinel in generate-pdf-page-budget.test.mjs (lines 271-283) is still a real control: it symlinks an output dir out of the sandbox, targets a lexically-inner path, and asserts the write is refused and the file does not exist, so it only passes when the realpath check actually fires. CI green on all three platforms, which is what matters for the macOS /var and Windows short-name cases.

santifer:
Enforcing the workspace boundary in the direct renderer is worth doing, @luochen211 — a PDF path that can write outside the tracker-owned workspace is a data-safety question rather than a tidiness one, and `realpathSync` in the import list suggests you're resolving symlinks before deciding, which is the part naive versions of this check miss.

Not merged tonight (merge ceiling). When it's reviewed the thing I'll look at first is what happens to a path that legitimately resolves outside the workspace — refuse, or fall back to a workspace-relative location? A boundary check that silently relocates output is its own surprise. If the PR body states the intended behaviour for that case, it'll go faster.

Scott-Emberson (approved):
Re-reviewing at the current head after my earlier approval went stale: the owned test changed since, so here is a fresh pass on `tests/generate-pdf-page-budget.test.mjs`.

The file grew into a stronger version than the one I first approved, and it holds up. Three things landed. The sandbox paths now go through `realpathSync` and `linkRepoPackage(sandbox, 'js-yaml')` links the dependency in, which is genuine hermeticity hardening: on a symlinked `output/`, `import.meta.url` resolves from realpath while `process.argv[1]` keeps the caller's spelling, so `generate-pdf`'s `isMain` guard goes false and the spawned script exits 0 having done nothing, and the assertions would then run against empty output. There is a `CAREER_OPS_TRACKER` case covering the install-directory-equals-workspace layout. And the external-input boundary test is the one that matters: an external input path must exit non-zero, print the refusal, produce no PDF, and leave zero `.career-ops-render-*` temp files in the external directory, then it calls `renderHtmlToPdf` directly with a stubbed browser to observe the temp path, so the boundary is covered below the CLI as well as through it.

Multi-condition sentinels, hermetic, and better than the version I first signed off. Green on CI. Approving from my side.

luochen211:
Synced current main in commit 6364911. Re-ran `node tests/generate-pdf-page-budget.test.mjs` locally; all boundary and temporary-file assertions passed. The PR remains approved with no unresolved review threads; awaiting the new CI run and merge queue.

santifer:

santifer:
Fail-closed where it counts, @luochen211: the renderer now refuses an escaping output path before launching Chromium or writing any temp file, and the symlink-aware anchor check closes the loophole a plain prefix compare would leave. Merged 🚀

PR #2861, opened 2026-08-14
fix(pipeline): serialize shared Playwright sessions

### LINKED ISSUE #2551: modes/pipeline.md: parallel evaluation workers can share one Playwright/MCP session — silent cross-contamination risk
> ⚠️ Privacy note: nothing job/company-specific below — just the mechanics of the failure.

### Code of Conduct

### Existing issues
- [x] I searched existing issues and this hasn't been reported yet (related: #2545, #506)

### What happened?

`modes/pipeline.md` Step 3 instructs the orchestrator to parallelize: *"If there are 3+ pending URLs, launch agents in parallel (Agent tool with `run_in_background`) to maximize speed — at most one agent per pending URL."* Each spawned worker independently runs the standard liveness-gate / JD-extraction step from `modes/oferta.md`, which defaults to the interactive Playwright MCP tools (`browser_navigate` + `browser_snapshot`).

Ran `/career-ops pipeline` against 4 pending URLs. Two of the four parallel workers reported that a `browser_navigate`/`browser_snapshot` call transiently returned a *different* worker's page mid-task — clear evidence all 4 were sharing one Playwright/MCP browser session instead of each holding an isolated one. Both workers happened to notice the title/URL mismatch and re-verified before writing anything, so no bad data landed this time. A worker that doesn't notice would silently evaluate and report on the wrong job posting.

I filed #2545 to fix the ambiguous wording of `modes/_shared.md`'s "NEVER 2+ agents with Playwright in parallel" rule, since it technically already prohibits this. But the wording was a symptom — the actual root cause is that `modes/pipeline.md` Step 3 directs exactly the pattern that rule forbids, and it does so unconditionally, for the normal case (3+ pending URLs needing Playwright-based extraction, which is the default/preferred JD-extraction method).

### Steps to reproduce

1. Populate `data/pipeline.md`'s Pending section with 3+ URLs on SPA-heavy boards that need Playwright-based extraction (Workday, Apple jobs, Esri, etc. — anything WebFetch can't reliably render).
2. Run `/career-ops pipeline`.
3. Per Step 3, the orchestrator launches one `Agent()` subagent per URL in parallel.
4. Each subagent independently calls `browser_navigate`/`browser_snapshot`.
5. Because parallel `Agent()`-spawned subagents in the same session share one Playwright MCP connection, concurrent navigations race — a subagent can transiently see another subagent's page mid-snapshot.

### Expected behavior

Parallel pipeline workers should never be able to observe another worker's page content. Either the work should be safely parallelizable without a shared browser session, or it should fall back to safe execution (serializing whichever step actually needs the shared resource) when Playwright is involved.

### Prior art

- **#506 / PR #809** — the same root cause (N parallel workers sharing one inherited Playwright/MCP session) already happened once, in the headless `batch/batch-runner.sh` path. There, each parallel worker is a fresh `claude -p` subprocess, and the fix was `--strict-mcp-config` (no `--mcp-config`) so workers never inherit the parent session's MCP servers — batch workers don't need Playwright anyway (they use WebFetch, per `AGENTS.md`'s Offer Verification section). That fix doesn't port directly here: `pipeline.md`'s parallel workers are spawned via the interactive `Agent()` tool *inside* a single running session, not as separate `claude -p` processes, so there's no equivalent flag career-ops' own markdown instructions can pass to strip an `Agent()`-spawned subagent's inherited MCP tools.
- **#2545** — fixes the ambiguous rule wording. This issue is the follow-up design question: what should `modes/pipeline.md` actually do to comply with that rule?

### Possible fixes (for discussion — not committing to either yet)

**Option A — gate parallelism on an isolated extraction path.** `browser-extract.mjs` (the opt-in `scan.extractor: cli` path, already used by `scan`/`pipeline`/`oferta`'s liveness gate as an alternative to the interactive MCP tools) launches its own headless Playwright process per invocation — no shared session, so N parallel calls can't race. `pipeline.md` Step 3 could require `scan.extractor: cli` as a precondition for launching parallel workers, falling back to fully serial per-URL processing when it's unset (today's default). Same isolation *principle* as PR #809 (deny shared-session Playwright, route through a non-interactive path) — just reached via config instead of a CLI flag, since there's no flag-passing mechanism available to an `Agent()`-spawned subagent.
  - Known gap: `browser-extract.mjs` is strictly read-only by design (no clicks/fills). `auto-pipeline`'s Step 4 (drafting application-form answers for a role scoring ≥4.5) needs real interactive Playwright to navigate and read the actual form — two ≥4.5 workers in the same parallel batch would still race on that one sub-step even with Option A in place. Would need to either serialize just that step or document it as a known follow-up limitation.

**Option B — a Playwright mutex.** career-ops already has an atomic-mkdir lock idiom used for exactly this class of cross-process coordination problem — `pipeline-lock.mjs` (guards concurrent writes to `data/pipeline.md`), the tracker lock in `tracker-utils.mjs`, and `reserve-report-num.mjs`'s sentinel files. A `playwright-lock.mjs` following the same pattern could let parallel workers acquire/release a lock around just the moment they touch the shared browser (navigate → snapshot → close), staying parallel for everything else (research, writeup, report generation) and only serializing the genuinely unsafe critical section. This preserves headed-mode reliability against anti-bot walls (`modes/_custom.md` prefers headed Playwright over headless for exactly that reason), which Option A's headless `browser-extract.mjs` path doesn't have. Bigger lift, and only as safe as every spawned subagent actually honoring the lock — a weaker guarantee than "there's no shared resource to race on" in Option A.

I'm not attempting a fix yet — filing this to open the design discussion first, since it's an architecture-level call on how `pipeline.md`'s parallel-worker pattern should work, not a simple wording fix.

### CLI tool

Claude Code

### OS and Node.js version

macOS / Node 24

luochen211:
Fixes #2551.

Parallel pipeline workers are now conditional on extraction mode. Browser-backed Playwright/MCP extraction is explicitly sequential because workers may share a session and cross-contaminate navigation/snapshots. Parallel work remains allowed only when isolated non-browser processes are guaranteed; otherwise the mode falls back to sequential processing.

Validation: node --check test-all.mjs; git diff --check

## Summary by CodeRabbit

* **Documentation**
  * Updated pipeline guidance to use sequential processing for browser-backed extraction.
  * Documented conditional parallel processing for isolated CLI or non-browser extraction, with up to one URL per worker.
  * Clarified that browser sessions must not be shared across workers.

* **Tests**
  * Added coverage verifying the documented concurrency behavior and sequential default.


santifer:
Same note as #2868 and #2860, so I'll keep it short here.

This changes prose in `modes/pipeline.md` and the check is:

```js
if (pipelineMode.includes('some exact sentence')) { pass(...) }
```

That asserts **a sentence exists in a markdown file**. It passes against any version containing the sentence whether or not the behaviour is right, and it fails the day someone rewords the paragraph for an unrelated reason. Compare it to your own #2863, which measures the actual delays between requests — that one would catch a real regression; these two only catch an edit.

The underlying concern in #2861 (parallel workers sharing one browser session and cross-contaminating snapshots) is legitimate and I'd want it addressed. But a mode-file instruction is a request the agent can decline, and a test on the instruction's text doesn't change that — it's the same 'prompt-only rule with nothing enforcing it' class that #2791 found in a different corner today. Worth thinking about what could enforce it rather than describe it.

Not closed, not reviewed further tonight — the merge ceiling applies and #2863 is the one I'd look at first from your set.

santifer:
Right call, @luochen211: cross-contaminated snapshots evaluating the wrong posting is the worst kind of silent wrong-answer, and conditioning concurrency on the extractor (with 'when in doubt, sequential') fixes it without giving up parallelism where isolation is real. Merged 🚀

PR #2862, opened 2026-08-14
fix(pipeline): use browser-first LinkedIn extraction

### LINKED ISSUE #1793: Bug: OpenCode CLI is not detected on Linux when installed via the default installer

### LINKED ISSUE #2619: pipeline/oferta LinkedIn handling defaults to "ask user to paste JD" instead of attempting browser fetch first
**Title:** pipeline/oferta LinkedIn handling defaults to "ask user to paste JD" instead of attempting browser fetch first

**Body:**

`modes/pipeline.md` (the "Special cases" note under URL extraction) currently documents:

> **LinkedIn**: May require login → mark `[!]` and ask the user to paste the text

In practice, LinkedIn job pages (`linkedin.com/jobs/view/...`) don't require login to view the JD, and the "About the job" section renders fine via a real browser session — it's just lazy-loaded client-side, so a plain `WebFetch` (no JS execution) returns the page shell without the description. Tested via the `claude-in-chrome` MCP tool (`navigate` + `get_page_text`, no login prompt encountered) — the full JD (title, comp, requirements, "About the job" body) came back cleanly on the first load of a session.

At volume, LinkedIn does appear to throttle: after roughly 25-30 job-page loads in one session, the JD body stops populating even with waits/retries/reloads (metadata like title/comp/location still loads). That's a real constraint, but it's a *fallback* condition, not the common case — most users evaluating a handful of postings per session will never hit it.

**Current behavior:** every LinkedIn URL immediately asks the user to manually copy-paste the JD text, regardless of session volume. For users evaluating jobs one at a time (the common case), this means copy-pasting full JD text for every single LinkedIn posting — expensive in tokens and tedious, when a working automated path exists.

**Suggested fix:** update the LinkedIn special case in `modes/pipeline.md` (and wherever else it's referenced, e.g. `oferta.md` Block G ingestion) to:
1. If a browser automation tool is available (e.g. `claude-in-chrome` or Playwright with a persisted LinkedIn session), attempt fetch via that first.
2. Track load count / failure pattern per session; after ~2 failed attempts on one URL, or after a rough session threshold (~25-30 LinkedIn loads), fall back to asking the user to paste — this is the graceful-degradation behavior already validated as a local house rule pattern (happy to share the specifics).
3. Only skip straight to "ask user to paste" when no browser automation tool is available at all (e.g. headless/batch worker mode, where Playwright is already documented as unavailable).

This preserves the existing fallback for high-volume sessions while eliminating the manual-paste tax for the common single/low-volume evaluation case.

**Environment:** Claude Code w/ `claude-in-chrome` MCP tool enabled.

luochen211:
Closes #2619.
Closes #1793. Try Playwright/MCP for LinkedIn job pages before requesting pasted text, with bounded fallback for login-only/headless cases. Add the Linux ~/.opencode/bin default install directory to web CLI detection and guard both behaviors in test-all.

## Summary by CodeRabbit

* **New Features**
  * LinkedIn job description extraction now prioritizes browser-based retrieval, including headless batch processing.
  * Users can provide pasted job description text when browser extraction is unavailable or unsuccessful.
  * Login pages and incomplete content are no longer treated as verified job descriptions.

* **Bug Fixes**
  * Improved handling after repeated unsuccessful browser extraction attempts.
  * Added support for detecting the Linux OpenCode CLI installation.
  * Improved browser session cleanup and extraction reliability.


santifer:
Same note as #2868 and #2860, so I'll keep it short here.

This changes prose in `modes/pipeline.md` and the check is:

```js
if (pipelineMode.includes('some exact sentence')) { pass(...) }
```

That asserts **a sentence exists in a markdown file**. It passes against any version containing the sentence whether or not the behaviour is right, and it fails the day someone rewords the paragraph for an unrelated reason. Compare it to your own #2863, which measures the actual delays between requests — that one would catch a real regression; these two only catch an edit.

The underlying concern in #2861 (parallel workers sharing one browser session and cross-contaminating snapshots) is legitimate and I'd want it addressed. But a mode-file instruction is a request the agent can decline, and a test on the instruction's text doesn't change that — it's the same 'prompt-only rule with nothing enforcing it' class that #2791 found in a different corner today. Worth thinking about what could enforce it rather than describe it.

Not closed, not reviewed further tonight — the merge ceiling applies and #2863 is the one I'd look at first from your set.

santifer:
Direction is right, @luochen211 — browser-first for a single posting the user brings (their browser, their session, one URL) with the bounded fallback to paste is exactly how the LinkedIn intake should degrade, and it stays on the right side of the line the scanner providers can't cross. Main moved under it though: two files now conflict, so it needs a rebase. Merges on green after that — and if you can, split the opencode-on-Linux detection line into its own commit message line so the changelog reads clean.

luochen211:
Synced current main in commit b2c5c52. Re-ran the full suite locally with the external Gemini smoke test skipped: 5314 passed, 0 failed, 2 warnings. The browser-first LinkedIn extraction change is unchanged; please re-review after CI.

santifer:

santifer:
One adjustment before this lands, @luochen211, and it's about placement, not quality: please drop the test-all.mjs hunk that reads `web/src/lib/clis.ts`. We keep a hard isolation rule that web imports from core but core never reads into web/, and that assertion also freezes the literal string of the path-join rather than the behavior, so a harmless refactor of clis.ts would break core CI with no real regression.

The regression it guards (#1793, detecting ~/.opencode/bin on Linux) does deserve a test: the web maintainer is adding it as a behavior test in web/tests (mock home, assert detection), so no coverage is lost. With that hunk gone this PR lands purely on its pipeline/LinkedIn change, which is ready. Thanks for the quick rebase earlier today.

santifer:
The prompt-injection framing is what makes this mergeable, @luochen211: pasted job text stays data-never-instructions, the fallback fires only after two real failures, and the test-all assertion pins the rule's exact structure so a future edit can't quietly soften it. Merged 🚀

Issue #2532, opened 2026-08-05
update-system: apply() never deletes system files upstream has removed, so retired files persist forever on existing installs


gtm-k:
### Code of Conduct

- [X] I agree to follow the [Code of Conduct](https://github.com/santifer/career-ops/blob/main/CODE_OF_CONDUCT.md)

### Existing issues

- [X] I searched existing issues and this hasn't been reported yet

### What happened?

`apply()` checks out the paths in the target manifest and stages them. It has no step that removes system files the target *no longer ships*, so a file retired upstream stays on every existing install indefinitely — tracked, stale, and invisible.

Fresh clones are correct; only upgraded installs drift. The divergence therefore grows silently with install age, and nothing surfaces it.

**Concrete instance.** `plugins-registry.json` was superseded by the `plugins-registry/` directory of per-plugin entries, which is what `SYSTEM_PATHS` ships. On an install predating that migration the old single-file registry is still present and tracked, long after nothing reads it:

```console
$ grep -rn "plugins-registry\.json" --include="*.mjs" --include="*.md" .
(no matches)
```

There is a second-order effect worth fixing even if the deletion behaviour stays as-is: it trips the `SYSTEM_PATHS` coverage check in `test-all.mjs`, because it is a tracked file in neither `SYSTEM_PATHS` nor `USER_PATHS`. On an affected install that test fails until the file is deleted by hand, and the message points the wrong way — it reads

```
SYSTEM_PATHS coverage gap — a new file is unregistered and update-system will not ship it
```

when the file is not new, it is retired. Someone hitting that will go looking for a registration to add rather than a file to remove.

### Steps to reproduce

1. Install career-ops at a version that still shipped `plugins-registry.json` (before the `plugins-registry/` migration).
2. Run `node update-system.mjs apply` up to current.
3. `ls plugins-registry.json` → still present, still tracked, though `SYSTEM_PATHS` now ships only `plugins-registry/`.
4. `node test-all.mjs` → fails the `SYSTEM_PATHS` coverage check, describing the retired file as a new unregistered one.

### Expected behavior

Retired system paths should not linger on upgraded installs. Deleting files during an update is the one operation where a bug costs user data, so this seems worth a design decision rather than a drive-by patch:

- The safe signal is a **manifest diff** — paths in the installed manifest absent from the target's — not a filesystem sweep.
[...]

PR #2864, opened 2026-08-14
fix(updater): detect tree drift and prune retired files

### LINKED ISSUE #2630: update-system check: reports up-to-date while manifest files have drifted (version is a proxy for content)
# `update-system.mjs check` reports `up-to-date` while manifest files have drifted

## Summary

`check` compares two version strings. When upstream merges work without a release
bump — which is the normal state between releases — an installed copy on the same
version is told it is current, even though files on its own manifest have changed
upstream. There is no way for a user to discover this short of diffing by hand.

## Reproduce

On an install whose local version equals the latest released version:

```bash
node update-system.mjs check
{"status":"up-to-date","local":"1.25.0","remote":"1.25.0"}
```

Observed on 2026-08-08 against `origin/main` at `8e52e86`:

| | |
|---|---|
| last release commit | `ae1a92d`, 2026-08-04 |
| `VERSION` then, and now | `1.25.0` — unchanged |
| commits merged since | **44** |
| files changed | 63, of which **41** are under manifest directories (`providers/`, `tests/`, `modes/`, `templates/`, `seeds/`, `docs/`) |

Concretely, `c0a259a` added `providers/consider.mjs`, `providers/getro.mjs` and
`providers/joinup.mjs`. `providers/` is in `SYSTEM_PATHS`, and `apply` does
`git checkout FETCH_HEAD -- providers/`, so those files arrive the moment an
update runs. But `check` says `up-to-date`, so nothing prompts the user to run
one, and three zero-auth providers stay invisible.

## Cause

```js
// update-system.mjs, check()
if (compareVersions(local, remote) >= 0) {
  console.log(JSON.stringify({ status: 'up-to-date', local, remote }));
  return;
}
```

`remote` is the higher of the `VERSION` file and the newest GitHub Release tag.
Both are release-time artifacts, so neither moves when work merges to `main`.
`apply` is not gated this way — running it directly does fetch the drift — so the
capability is present and only the discovery path is missing.

## Prior art — this is the third instance of the same shape

False `up-to-date` has been reported twice before, both times because the two
version *sources* disagreed:

- **#306** — v1.4.0 tag published, `VERSION` on main not bumped.
- **#316** — `VERSION` stale at v1.3.0 against a v1.5.0 release.

The fix for those was to take the higher of the `VERSION` file and the Release
tag, which is the current code. That closed the disagreement between the two
sources but left the assumption underneath untouched: that a release artifact
tracks what is on `main`. Between releases it does not, and both sources are
wrong together rather than disagreeing — so the existing reconciliation cannot
detect it.

Worth noting the shape rather than just the instance: each fix so far has made
the version *lookup* more accurate, and the residual gap is that version is a
proxy for content at all.

**#2532** ("apply() never deletes system files upstream has removed") is the
sibling on the `apply` side — the same manifest-vs-upstream divergence, one step
later in the pipeline. A content-aware `check` would surface that class too.

## Why this is filed as an issue rather than a PR

Any fix changes what `check` means. Today `up-to-date` is a statement about
versions; making it content-aware turns it into a statement about files, and
installs that are currently told they are current would start being told an
update is available. That is a change to a documented command's semantics and to
the agent-facing contract in `AGENTS.md` ("If `{"status": "update-available", …}`
→ tell the user"), so it seems worth agreeing on the shape before there is a diff.

## Possible directions

Listed to frame the discussion, not as a preferred design:

1. **Compare the upstream tip commit.** Record the fetched SHA at `apply` time and
   compare on `check` via the GitHub API. Exact, but adds state and a request, and
   needs a fallback for installs with no recorded SHA.
2. **A distinct status.** Keep `up-to-date` meaning "same version" and add
   something like `files-drifted` for same-version-but-changed-content, so existing
   consumers of the JSON keep working and the agent prompt can differ in tone.
3. **Leave `check` alone; surface it in `doctor`.** Drift becomes a health signal
   rather than an update prompt. Smallest blast radius, but users only see it if
   they run `doctor`.

Happy to implement whichever direction you prefer.

## Impact on forks

A fork that tracks the manifest sees this as silent staleness: it believes it is
current, so it never re-runs `apply`, and new manifest files simply never arrive.
The three providers above went unnoticed for four days on the install this was
found on, and only surfaced when a branch diff made the absence visible.

luochen211:
Closes #2630 and #2532. The update check now compares the installed commit with upstream main so same-version system-file drift is reported. Apply prunes tracked upstream-removed system files, including root-level files, while protecting user paths and refusing empty/failed remote-tree lookups.

## Summary by CodeRabbit

* **New Features**
  * Update checks now detect upstream system-file changes using version and commit information.
  * Removed upstream system files are automatically pruned during updates.
  * Update results include commit metadata and system-tree change details.
* **Bug Fixes**
  * User-owned files are preserved during cleanup.
  * Empty remote trees no longer incorrectly mark all local files for deletion.
  * File deletion and remote-tree lookup failures are handled safely.
* **Tests**
  * Added coverage for system-file pruning and user-file protection.

### DIFF
--- a/tests/updater-rollback-behavior.test.mjs
+++ b/tests/updater-rollback-behavior.test.mjs
@@ -13,7 +13,7 @@ import { mkdtempSync, mkdirSync, writeFileSync, existsSync, rmSync } from 'fs';
 import { tmpdir } from 'os';
 import { join } from 'path';
 import { pass, fail } from './helpers.mjs';
-import { gitIn, removeAdditionsNotInHead } from '../update-system.mjs';
+import { gitIn, removeAdditionsNotInHead, staleSystemFiles } from '../update-system.mjs';
 
 // A throwaway git repo plus a ctx that binds the rollback helper's git runner
 // and filesystem root to it, so nothing touches the real working tree.
@@ -33,6 +33,25 @@ function stagedPaths(g) {
 }
 
 console.log('\n🧪 Testing updater rollback behavior (#2015)...');
+
+// ── 0. system-file pruning is complete but user-safe (#2532) ──
+{
+  const local = ['plugins-registry.json', 'tests/old.test.mjs', 'data/applications.md', 'scratch.txt'];
+  const remote = ['tests/new.test.mjs', 'data/applications.md'];
+  const system = ['plugins-registry.json', 'tests/', 'data/'];
+  const user = ['data/'];
+  const stale = staleSystemFiles(local, remote, system, user);
+  if (stale.length === 2 && stale.includes('plugins-registry.json') && stale.includes('tests/old.test.mjs')) {
+    pass('stale system pruning removes upstream-deleted root files and system descendants');
+  } else {
+    fail(`stale system pruning selected the wrong files: ${JSON.stringify(stale)}`);
+  }
+  if (staleSystemFiles(local, [], system, user).length === 0) {
+    pass('stale system pruning never treats an empty remote tree as a delete-all signal');
+  } else {
+    fail('stale system pruning would delete files from an empty remote tree');
+  }
+}
 
 // ── 1. protectedPaths: a user's pre-staged work survives a rollback ──
 {
--- a/tests/updater-rollback-behavior.test.mjs
+++ b/tests/updater-rollback-behavior.test.mjs
@@ -51,6 +51,18 @@ console.log('\n🧪 Testing updater rollback behavior (#2015)...');
   } else {
     fail('stale system pruning would delete files from an empty remote tree');
   }
+
+  const userDeleted = staleSystemFiles(
+    ['data/applications.md', 'tests/old.test.mjs'],
+    ['tests/new.test.mjs'],
+    ['tests/', 'data/'],
+    ['data/'],
+  );
+  if (userDeleted.includes('tests/old.test.mjs') && !userDeleted.includes('data/applications.md')) {
+    pass('stale system pruning excludes an upstream-deleted user-layer file');
+  } else {
+    fail(`stale system pruning would select a user-layer file: ${JSON.stringify(userDeleted)}`);
+  }
 }
 
 // ── 1. protectedPaths: a user's pre-staged work survives a rollback ──
--- a/update-system.mjs
+++ b/update-system.mjs
@@ -632,6 +632,26 @@ function mergePathLists(...lists) {
   return merged;
 }
 
+function normalizeRepoPath(value) {
+  return String(value || '').replace(/\\/g, '/').replace(/^\.\//, '');
+}
+
+function pathMatchesManifest(file, entry) {
+  const normalizedFile = normalizeRepoPath(file);
+  const normalizedEntry = normalizeRepoPath(entry).replace(/\/$/, '');
+  return normalizedFile === normalizedEntry || normalizedFile.startsWith(`${normalizedEntry}/`);
+}
+
+export function staleSystemFiles(localFiles, remoteFiles, systemPaths, userPaths = USER_PATHS) {
+  const remote = new Set([...remoteFiles].map(normalizeRepoPath));
+  if (remote.size === 0) return [];
+  return [...localFiles]
+    .map(normalizeRepoPath)
+    .filter((file) => !remote.has(file))
+    .filter((file) => systemPaths.some((entry) => pathMatchesManifest(file, entry)))
+    .filter((file) => !userPaths.some((entry) => pathMatchesManifest(file, entry)));
+}
+
 // Files the self-reexec stage must check out so the TARGET update-system.mjs
 // loads without a missing-module crash. Today this is the entry plus its only
 // local import; resolveReexecCheckout derives the real set from the fetched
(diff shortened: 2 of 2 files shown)


Scott-Emberson (changes requested):
Nice, focused addition: block 0 in `tests/updater-rollback-behavior.test.mjs` pins `staleSystemFiles` selection, and the two assertions it does make are correct. The upstream-deleted root file plus the system descendant get selected, and the empty-remote case is a real control for the `remote.size === 0` short-circuit.

One change before this lands, and it is the headline property the block claims.

**1. request-changes, `tests/updater-rollback-behavior.test.mjs:37-54`, the "user-safe" half is a false-pass.** The block is titled "system-file pruning is complete but user-safe (#2532)", but nothing in it exercises the user-layer guard in `staleSystemFiles` (`update-system.mjs:652`, `.filter((file) => !userPaths.some(...))`). In both assertions that filter is dead:

- First assertion: the only user-path local file, `data/applications.md`, is also in `remote`, so `.filter((file) => !remote.has(file))` (`update-system.mjs:650`) drops it before the user filter is ever consulted.
- Second assertion: `staleSystemFiles(local, [], ...)` hits `if (remote.size === 0) return []` (`update-system.mjs:647`) and returns before any filter runs.

I ran the real function against a mutant with the user-guard line deleted. Both assertions stay green either way:

```
with guard   assert1 -> ["plugins-registry.json","tests/old.test.mjs"]  true
no guard     assert1 -> ["plugins-registry.json","tests/old.test.mjs"]  true
with guard   assert2 -> []  true
no guard     assert2 -> []  true
```

A discriminating case exists and is the one missing: a user-layer file that upstream deleted, so it sits in `local`, is absent from a non-empty `remote`, and matches both `system` and `user`. With the guard it is excluded; without it, it gets selected for deletion:

```js
const stale2 = staleSystemFiles(['data/applications.md', 'tests/old.test.mjs'], ['tests/new.test.mjs'], ['tests/', 'data/'], ['data/']);
if (!stale2.includes('data/applications.md') && stale2.includes('tests/old.test.mjs'))
  pass('stale pruning never selects a user-layer file even when upstream-deleted');
else fail(`stale pruning would delete a user-layer file: ${JSON.stringify(stale2)}`);
```

Why it blocks: `staleSystemFiles` feeds `unlinkSync` during `apply()`. The real SYSTEM_PATHS and USER_PATHS overlap on directories like `interview-prep/sessions/`, so that guard is live defense for a user's own files, not a formality. As written, a regression that removes it ships green. Add the case above and it is covered.

Two smaller notes, neither blocking:

- The empty-remote failure message drops the offending list: `fail('stale system pruning would delete files from an empty remote tree')`. The sibling message interpolates `JSON.stringify(stale)`; mirror it so a failure names what leaked.
- The production call site relies on the `userPaths = USER_PATHS` default, but every test passes `user` explicitly, so the default binding is never exercised. Optional: one assertion that omits the arg.

Add the discriminating user-layer case and this is good to go from my side.

luochen211:
已补充真正区分用户层文件的回归测试：当 upstream 删除 data/applications.md 时，stale pruning 不会选择它，同时仍会选择系统层旧文件。本地 updater 行为测试和 345 项迁移检查通过，最新全平台 CI 通过，请重新审查。

Scott-Emberson (approved):
Re-reviewed at 46be079 on tests/updater-rollback-behavior.test.mjs. Our blocker is addressed. The new third assertion (lines 56-67) is the discriminating case we asked for: with a non-empty remote and a user-layer file absent from it, staleSystemFiles must exclude data/applications.md while still selecting the deleted system file. I traced it against staleSystemFiles (update-system.mjs:641-648): the user-guard filter is the sole thing excluding the user file, so if that line were dropped the assertion fails. That is a genuine false-pass killer for the byte-safety property. Hermetic, same pass/fail shape as its neighbours, failure message names the leaked list. CI green on all three OSes.

Two non-blocking items carry over, both optional: the empty-remote failure message (line 52) still drops the leaked list, and the USER_PATHS default binding is never exercised since every call passes user explicitly. Neither gates the owned file.

Approving from our /tests/ scope. The update-system.mjs prune path is your red line, not something I am signing off.

santifer:
The fixture is the part I'd keep, @luochen211:

```js
const local  = ['plugins-registry.json', 'tests/old.test.mjs', 'data/applications.md', 'scratch.txt'];
const system = ['plugins-registry.json', 'tests/', 'data/'];
const user   = ['data/'];
```

Putting `data/applications.md` in the local set and `data/` in *both* system and user is exactly the case that decides whether a pruning routine is safe: it has to remove a retired system file and must never touch the user's tracker, and the only way to know it does both is to hand it a set where the two overlap. `scratch.txt` as an unrelated local file is the third leg. That's a real behavioural test, unlike the source-pattern checks in #2861/#2862 — same author, same evening, and the difference between the two approaches is visible side by side.

Not merged tonight (merge ceiling). Same note as #2866: with #2820 landed and #2857/#2868 open, the updater PRs need sequencing before any of them is reviewed in isolation.

santifer:
The three tests are the review, @luochen211: upstream-deleted root files pruned, an empty remote tree never read as delete-all, and a user-layer file untouchable even when upstream removes it. That last one is the DATA CONTRACT holding inside the most dangerous code path the repo has. Merged 🚀

PR #2866, opened 2026-08-14
feat(updater): require confirmation before installation

### LINKED ISSUE #2865: feat(updater): automate update checks with confirmation-gated installation
## Goal

Allow career-ops to check for upstream system updates automatically while keeping installation human-confirmed.

## Scope

- Keep `node update-system.mjs check` non-installing and suitable for scheduled macOS/Linux/Windows jobs.
- Require an explicit confirmation flag before any update installation.
- Keep `apply --force` confirmation-gated as well; no background job may overwrite system files.
- Preserve the existing Data Contract: user-layer files are never auto-updated.
- Keep dismissal and rollback behavior observable and reversible.
- Document the interactive update mode so it passes confirmation only after the user approves.

## Non-goals

- No silent installation.
- No automatic `--force` behavior.
- No application submission, external messaging, or user-data migration.
- No OS-specific notification dependency in the core updater.

## Acceptance criteria

- A scheduled `check` exits successfully when there is no update and never changes system files.
- When an update exists, `check` emits machine-readable update metadata for a later interactive session to present.
- `apply` without explicit confirmation exits before creating a backup, fetching, or changing files.
- Installation proceeds only with explicit confirmation; `--force` still requires that confirmation.
- User-layer files and unrelated untracked files remain untouched.
- Tests cover the confirmation gate and self-reexec propagation.

Related: #2630, #2532

luochen211:
Closes #2865. Scheduled update checks remain non-installing, while every installation now requires an explicit --confirm flag. The confirmation is propagated through the updater self-reexec, --force remains separately explicit, and the interactive update mode uses the guarded command.

## Summary by CodeRabbit

* **New Features**
  * Update installation now requires explicit confirmation with `--confirm`.
  * Changelogs are now displayed in the configured output language.
  * Help and update instructions explain the confirmation requirement and recovery steps.

* **Bug Fixes**
  * Prevented unconfirmed update installation during automated or self-reexecuted checks.

* **Tests**
  * Added coverage for confirmation validation, propagation during update execution, and rejection of insufficient confirmation methods.


santifer:
`--confirm` as a hard requirement is the right shape, @luochen211:

> A scheduled update check never installs files.

That's the distinction that matters: a *check* that can install is a background process that changes the code under a user who asked a question. Making the flag mandatory rather than defaulting to confirm means a scheduled path can't acquire the ability by accident later — the failure is an error, not an install.

Not merged tonight (merge ceiling), and one thing to know: it touches the updater, and there are now four of your PRs plus two from other contributors in that area (#2857, #2864, #2868, this one, plus #2820 which merged tonight and #2881's report). They'll need an order. #2868 is already superseded by what landed; worth checking whether this one's diff still applies cleanly on top of tonight's main before it gets reviewed.

santifer:
The mechanism is right, @luochen211 — required confirmation so scheduled checks can never install silently, and the 0600 marker with timingSafeEqual so nothing can impersonate the re-exec phase, is updater hardening we want. Three call sites the diff misses, though, and one of them breaks the main documented flow:

1. `package.json` scripts.update is `node update-system.mjs apply` — after this merges, `npm run update` (the command docs/SCRIPTS.md tells every user to run) exits 1 asking for a flag it can't pass. Typing `npm run update` IS the human confirmation, so that script should become `apply --confirm`.
2. `docs/SCRIPTS.md:25` documents the same bare form.
3. `docs/FAQ.md:9` (the Windows symlink answer) also says `node update-system.mjs apply`.

Three one-line edits and this merges — the rest of the diff is ready as-is.

santifer:
Nice work, @luochen211. The authenticated one-use reexec marker (random token, timingSafeEqual, realpath/lstat checks, destructive consume) is genuinely careful design, and the negative tests prove you thought about the threat model, not just the happy path. Two things to fix before this can merge:

1. **Runtime crash**: update-system.mjs drops `renameSync` from the fs import, but `writeGitignoreAtomic` still calls it at update-system.mjs:1401 (it is now the only occurrence in the file). A real `apply` will throw `ReferenceError: renameSync is not defined` mid-run when it reconciles .gitignore. The migration tests are source-regex based, so CI cannot catch it.
2. **Hermeticity regression**: upgrade-tests.mjs replaces `hermeticEnv(cfg)` with `{ ...process.env, GIT_CONFIG_GLOBAL: cfg, ... }`, losing the `GIT_CONFIG_NOSYSTEM`/`GIT_CONFIG_COUNT=0` isolation the file's own header documents. `{ ...hermeticEnv(cfg), CAREER_OPS_UPDATE_CONFIRM: '1' }` keeps both.

Happy to merge once these are in: the confirmation gate itself is the right call.

luochen211:
Addressed the remaining runtime issue in commit 4aafc89: restored the `renameSync` import while preserving the latest main merge. The hermetic upgrade fixture and documented `--confirm` callers are retained. `node updater-migration-tests.mjs` passes 361/361. Please re-review after CI.

santifer:
Sorry for the wait on this one, @luochen211: it sat 8 days and it deserved better, because the design here is the good kind of careful. Making the env channel unable to authorize the *initial* invocation is the right threat model, `timingSafeEqual` on a 32-byte token is the right primitive, and `isLegacyReexec()` gating on an actually-existing backup branch is a genuinely thoughtful migration path for people upgrading *from* an old updater. The docs and the `package.json` script move in lockstep. No dependency added.

One real problem, and I found it by running your own code rather than reading it.

**On macOS the marker your own `createReexecMarker()` writes is rejected by your own `consumeReexecMarker()` guards.** `os.tmpdir()` returns `/var/folders/…/T`, which is a symlink to `/private/var/folders/…/T`. You canonicalize one side only (`realpathSync(tmpdir())` for `tmpRoot`) but not the path you're validating, so three of the four guards fail on a legitimate marker:

```
tmpdir()             = /var/folders/6b/…/T
realpathSync(tmpdir) = /private/var/folders/6b/…/T
marker               = /var/folders/6b/…/T/career-ops-reexec-gCJJtO/marker

dirname(parent) !== tmpRoot        -> true   (rejects)
realpathSync(parent) !== parent    -> true   (rejects)
realpathSync(path) !== path        -> true   (rejects)
==> the marker the PR just created IS REJECTED by its own guards
```

To be precise about the blast radius, because it is smaller than it sounds: nothing breaks for the user. The child gets `--confirm` in argv now, so `updateConfirmed` is true regardless, and `isReexec` falls through to the legacy branch. The update still works. What is lost is exactly the thing this PR is for: on macOS the authenticated channel is dead and `--force` propagation is authorized by the legacy env path instead. **The improvement is inert on the platform, silently.**

**And the reason CI is green is worth naming, because it is the more useful lesson.** The matrix does include `macos-latest`, so this isn't a coverage gap. It's that all five new asserts test *rejection* — forged marker, env-only confirm, env-only force, force without confirm, legacy without backup branch. A bug that rejects **everything** passes every one of them. The suite currently cannot tell "secure" from "broken shut".

Two things and I merge:

1. **Canonicalize at creation.** `const directory = realpathSync(mkdtempSync(join(tmpdir(), 'career-ops-reexec-')));` — I verified this passes all four guards on macOS. Canonicalizing the supplied path in the consumer works too, but doing it at creation keeps the consumer's guards strict, which is what you want in the validating half.
2. **Add the positive assert**: a marker produced by `createReexecMarker()` must authorize, and consuming it twice must not. That's the one that would have caught this, and without it the bug can come back the next time someone tightens a guard.

Nothing else blocking. Ping me when they're in and it goes in the same day.

luochen211:
Addressed the macOS marker-path issue in 1baa885: createReexecMarker() now canonicalizes the directory returned by mkdtempSync(), while consumeReexecMarker() keeps its strict validation. Added a positive regression that verifies a generated marker authorizes exactly once and cannot be consumed twice. Local updater migration suite: 362 passed, 0 failed; the latest macOS/Linux/Windows and security checks are green. Please re-review.

luochen211:
Addressed the macOS marker-path request in 1baa885: marker creation now canonicalizes the `mkdtemp` directory, and the migration suite verifies a created marker authorizes exactly once. Local updater migration passed 362/362, upgrade PR gate was GREEN, quick suite passed 5310/0 (3 existing warnings), and refreshed PR checks are green. @santifer please re-review.

santifer:
Third ask, @luochen211, and it's more important than the two above: **`AGENTS.md:73` still says `apply` without `--confirm`, and this PR doesn't touch it.**

```
AGENTS.md:73   If yes → `node update-system.mjs apply`. If no → `node update-system.mjs dismiss`.
```

You updated `modes/update.md` step 3 and `docs/SCRIPTS.md`, which is thorough. But that line in `AGENTS.md` is the session-start path: the agent checks for updates on the first message, the user says yes, and the agent runs exactly that command. After this merges it would run into *"Installation requires explicit confirmation"* — the agent having just told the user it was updating.

`AGENTS.md` is the canonical source for agent instructions in this repo (the per-CLI entry files are thin wrappers over it), so it's the one that has to be right.

Worth knowing for sequencing: I checked, and `update-system.mjs` never validates unknown flags — the only argv check is `includes('--force')` — so `apply --confirm` already behaves exactly like `apply` on current `main`. The doc change is therefore safe in either order, which means there's no window where one is landed and the other isn't.

Heads-up that @Solaris-star's #3213 is editing those same `AGENTS.md` lines for a different reason (the `reason: system-files-changed` field the instructions were ignoring). I've told them about the flag too. Whichever of you gets there first, the other rebases onto it: I'd rather you two knew than discover it as a conflict.

luochen211:
Addressed the session-start caller in a8c6e41: `AGENTS.md` now runs `node update-system.mjs apply --confirm` after the user approves the update.

Verification:
- `node updater-migration-tests.mjs` — 362/362 passed
- `env -u GEMINI_API_KEY node test-all.mjs --quick` — 5310 passed, 0 failed, 3 environment warnings
- Current-head CI — all 13 checks passed

#3213 remains open on its earlier head, so this branch is still mergeable. Please re-review.

santifer:
Reviewed in depth, @luochen211, and the confirmation flow you built is right: one thing to close before it merges, because it undercuts exactly the property this PR adds.

**The legacy-reexec path lets environment variables authorize an initial run without `--confirm`.** If a checkout has any leftover `backup-pre-update-*` branch (and every earlier update leaves one), then `CAREER_OPS_UPDATE_REEXEC=1` + `CAREER_OPS_UPDATE_BACKUP_BRANCH=<that branch>` makes `isLegacyReexec()` return true, and the `legacyReexec` clause in `apply()` sets `updateConfirmed` without `--confirm`: it also skips the lock and backup because the run is treated as a reexec. I verified the path end to end in your branch (update-system.mjs:1629-1636 and isLegacyReexec at :80).

The fix direction: legacy reexec should only imply confirmation when there is proof of an ACTIVE parent run (the reexec marker file or the update lock), not merely a branch that matches the naming pattern. Happy to merge once the initial-invocation case is closed: the rest of the gate, including the AGENTS.md wiring you added in a8c6e41, is exactly what #2860 asked for.

luochen211:
Synced this branch with current `career-ops-hq/main` in 7caf3ed and resolved the updater conflict by keeping the authenticated confirmation/re-exec flow while adopting the canonical organization URLs.

Fresh validation on the merged head:

- `node updater-migration-tests.mjs` — 379 passed, 0 failed
- `node upgrade-tests.mjs --pr-gate` — GREEN for both tracked and untracked dismissal-marker fixtures
- `env -u GEMINI_API_KEY node test-all.mjs --quick` — 7,688 passed, 0 failed, 14 existing environment/content warnings

All review threads are resolved. @santifer please re-review once the refreshed checks finish.

santifer:
Sorry, @luochen211: this one was green in my batch preview an hour ago and is now CONFLICTING in update-system.mjs because I merged #3474 (self-bootstrap backups in `update-system.mjs`) right before it. That's on me, not on the PR. One rebase onto current main and it goes in with the next batch; the review is done (the `--confirm` propagation through the re-exec marker and the separate `--force` are exactly what #2865 asked for). If you'd rather I resolve it on your branch, say so.

luochen211:
Thanks — I'll rebase onto current main and resolve the update-system.mjs conflict on my branch.

luochen211:
Rebased onto current main after #3474 and resolved the update-system.mjs conflict on my branch. The merged result keeps the self-bootstrap backup tracking alongside the authenticated re-exec marker and separate --confirm/--force propagation.\n\nFresh validation on head 8e6da42:\n- node updater-migration-tests.mjs — 379 passed, 0 failed\n- node upgrade-tests.mjs --pr-gate — GREEN\n- env -u GEMINI_API_KEY node test-all.mjs --quick — 7,755 passed, 0 failed\n- GitHub checks — all green on Linux, macOS, and Windows\n\nReady for the next merge batch.

luochen211:
One broader note while we're here: I'd like to keep growing toward a maintainer role in career-ops. I'm especially interested in taking more responsibility for updater reliability, security boundaries, and regression review. If reviewer or updater area-owner responsibilities are the appropriate next step, I'd be happy to take those on and keep building toward maintainership over time.

santifer:
Merged, @luochen211: scheduled checks stay non-installing and every install now needs an explicit `--confirm`, carried through the self-reexec by a token in a private tmp marker instead of a bare env flag, with `--force` kept separately explicit. Closes #2865. Thanks for the re-sync onto the org URLs. 🚀

ilyakanevskiy:
@luochen211 this merged change left one call site in `upgrade-tests.mjs` behind, and it's now failing the upgrade-regression job on `main` (and on every open PR whose CI runs against post-merge `main`).

`--confirm` / `CAREER_OPS_UPDATE_CONFIRM` was added to the apply invocation in the canary leg (line ~197) but not to the one in `localPathsLeg()` at line 423, which still calls a bare `['update-system.mjs', 'apply']`. Since the gate now short-circuits before any local-paths evaluation, `apply` exits with "Installation requires explicit confirmation…" instead of the declaration-aware failure the leg asserts. Result:

- `apply names the declaration file as the reason` FAIL — output never contains `config/local-paths.txt`
- `apply names the offending entry (run-nightly.ps1)` FAIL — same
- (`apply fails closed … exit 1` still passes, but for the wrong reason)

Reproduces on a clean checkout of current `main`:

```
node upgrade-tests.mjs --local-paths
```

The fix is to make line 423 match the canary leg:

```js
output = execFileSync(process.execPath, ['update-system.mjs', 'apply', '--confirm'], {
  cwd: install, encoding: 'utf-8', timeout: 300000,
  env: { ...hermeticEnv(cfg), CAREER_OPS_UPDATE_CONFIRM: '1' },
});
```

That gets `apply` past the new gate so the leg exercises what it's meant to — the local-paths declaration failing closed. Since it's your change, do you want to send the follow-up PR?

Also filed as #3756 for tracking.

santifer:
Closing the loop here too, @ilyakanevskiy: the `localPathsLeg()` call site you pointed at was fixed by #3755, and main's upgrade-regression job has been green since. Thanks for naming the exact line instead of the symptom.

PR #2868, opened 2026-08-14
fix(updater): ignore line-ending-only system drift

### LINKED ISSUE #2817: update-system.mjs apply: incomplete .gitattributes renormalization in 80d104f9 makes #2448's divergence check flag ~150 untouched files, silently no-op'ing the whole update

luochen211:
Closes #2817. Ignore CRLF/LF-only differences when detecting locally edited system files, preventing pre-.gitattributes installs from treating the entire system manifest as user-modified and silently skipping updates.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved system file change detection to ignore line-ending differences, preventing CRLF/LF normalization from being incorrectly reported as user edits.

* **Tests**
  * Added coverage to verify that line-ending-only changes are handled correctly.

### DIFF
--- a/update-system.mjs
+++ b/update-system.mjs
@@ -743,7 +743,11 @@ export function locallyModifiedSystemFiles(paths, upstreamRef = 'FETCH_HEAD', ct
 
   const diffNames = (ref) => {
     try {
-      return runGit('diff', '--name-only', ref, '--', ...paths).split('\n').map((p) => p.trim()).filter(Boolean);
+      // A pre-.gitattributes install can differ from the normalized upstream
+      // blob only by CRLF/LF at end of line. That is not a user edit and must
+      // not make every system file look locally modified (#2817).
+      return runGit('diff', '--ignore-space-at-eol', '--name-only', ref, '--', ...paths)
+        .split('\n').map((p) => p.trim()).filter(Boolean);
     } catch {
       // An unreadable ref (shallow clone, unrelated histories) must never abort
       // the update — it degrades the warning, not the checkout.
--- a/updater-migration-tests.mjs
+++ b/updater-migration-tests.mjs
@@ -45,6 +45,12 @@ const systemPaths = extractArray('SYSTEM_PATHS');
 const userPaths = extractArray('USER_PATHS');
 const bootstrapPaths = extractArray('BOOTSTRAP_PATHS');
 
+if (/diff', '--ignore-space-at-eol', '--name-only'/.test(source)) {
+  pass('local system edit detection ignores CRLF/LF-only drift from .gitattributes (#2817)');
+} else {
+  fail('local system edit detection can misclassify CRLF/LF-only drift as user edits (#2817)');
+}
+
 // Every concrete (non-directory) manifest entry (SYSTEM_PATHS or
 // BOOTSTRAP_PATHS) must exist in the working tree. A path deleted upstream
 // but left in the manifest survives as a permanent `error: pathspec ...` in


luochen211:
验证状态：Linux/macOS、静态分析、CodeQL、升级回归和安全检查均通过。Windows 测试失败在既有的  并发锁场景：，导致 29/30 条目保留；失败不涉及本 PR 修改的扫描/更新器代码。当前账号无法对上游 workflow 执行 rerun，请维护者复跑该 job 后再判断。

luochen211:
验证状态：Linux/macOS、静态分析、CodeQL、升级回归和安全检查均通过。Windows 测试失败在既有的 agent-inbox-tests.mjs 并发锁场景：EPERM（agent-inbox.md.lock.recover），导致 29/30 条目保留；失败不涉及本 PR 修改的扫描/更新器代码。当前账号无法对上游 workflow 执行 rerun，请维护者复跑该 job 后再判断。

santifer:
@luochen211 — this one is overtaken by #2820, which merged into main earlier tonight against the same issue (#2817). Not closing it; that's your call or Santiago's. But you should have the comparison, because two details separate the approaches and both matter:

**1. `--ignore-space-at-eol` is wider than the bug.** It ignores *any* trailing whitespace difference, not just the carriage return. A user who genuinely edits trailing whitespace in a system file would stop being detected as having local edits — which is the exact thing `locallyModifiedSystemFiles` exists to catch. What landed uses `--ignore-cr-at-eol`, which ignores the CR and nothing else.

**2. Keeping `--name-only` leaves the hole open.** This is the part I'd have got wrong too, and #2820's comment is what taught me:

> `--name-only` can list a path on the blob-OID comparison alone, before the textual diff runs, so a CRLF/LF-only file survives `--ignore-cr-at-eol` and the guard leaks right back.

So the flag goes in, the diff still lists the file, and the fix reads as applied while the behaviour is unchanged. The merged version uses `--numstat` specifically to force the textual diff, and pulls the path from field 3.

**3. One thing worth taking away regardless of this PR.** Your check is:

```js
if (/diff', '--ignore-space-at-eol', '--name-only'/.test(source)) { pass(...) }
```

That asserts a *string is present in the source*, not that the behaviour is right. It would pass against a version that had the flag and didn't work, and it fails right now purely because main spells the flag differently — nothing about correctness changed. The merged PR's fixture builds a repo with CRLF blobs in the merge-base and asserts that the CR-only file drops out while a genuine edit is still reported. That's the difference between a test that pins behaviour and one that pins text, and it's worth applying to the other ten PRs you have open tonight.

Sorry it collided — #2820 was opened first and finished its review cycle today. That's timing, not a judgement on yours.

luochen211:
Closing as superseded by #2820, which is already merged and provides the narrower CRLF/LF-only drift handling with behavioral coverage.

PR #2869, opened 2026-08-14
feat(scan): remember permanently dead ATS boards

### LINKED ISSUE #2840: scan-ats-full: ~46% of every sweep re-probes boards already known to be 404
## Problem

Every `scan-ats-full.mjs` sweep re-probes boards it has already confirmed do not exist, because nothing about board liveness survives the process. `providers/_dns-cache.mjs` is explicitly in-process (5-min TTL, 30s negative TTL), and `data/portal-health.tsv` is only written by `scan.mjs` — the reverse sweep neither reads nor writes it.

Measured across a full sweep of the public dataset:

| Source | Boards | Permanently 404 | Waste |
|---|---|---|---|
| greenhouse | 8,333 | 3,010 | 36% |
| lever | 4,368 | 2,376 | 54% |
| ashby | 3,161 | 684 | 22% |
| workday | 12,884 | ~7,215 | 56% |
| **total** | **28,746** | **~13,300** | **~46%** |

A `--verbose` run confirms these are genuine `HTTP 404 Not Found` — companies that left the ATS or were never there — not transient failures.

So roughly half of every sweep is spent re-learning yesterday's 404s. That is not only wall-clock: it is the traffic that provokes the single-host throttle in #2839 (lever's unreachable count jumping 2,436 → 4,100 between back-to-back sweeps, taking live postings with it). Pruning the dead half is the cheapest way to stop generating that load.

## Proposed shape

A small persistent store — `data/dead-boards.tsv`, `ats\tslug\tmisses\tlast_checked` — plus a module that owns it:

- A board is retired only after **N consecutive 404s** (proposed 3), never on one.
- **Only an HTTP 404 may record a miss.** A 429, 5xx, timeout, abort or DNS failure means "could not tell" and records nothing. This asymmetry is the whole safety argument: if a throttled sweep could write misses, one bad run would permanently hide live boards — a worse version of the bug being fixed.
- Retired boards are **re-probed every 30 days**, so a company returning to an ATS is picked up rather than blacklisted forever.
- A board that answers has its row dropped, so the file stays proportional to the dead set rather than the whole directory.
- The skip happens inside the `parallelEach` callback, **not** by filtering the entries array — `resumeAt` indexes into that exact array, so filtering it would desync the `--resume` checkpoint offset.
- The skip count is printed in the run summary; coverage that is silently reduced reads as "the directory shrank".

Measured on a warm store over 300 lever boards: unreachable 172 → 18, with 158 skipped — matching the 54% dead rate above.

## Why an issue first

Per CONTRIBUTING this is an architecture change, not a bug fix: it adds a root module and a new user-layer data file. Happy to be routed to core, a plugin, or told the shape should be different — in particular:

1. **Reuse `portal-health.tsv` instead of a new file?** It already means "consecutive reachability per portal" for `scan.mjs`. Its format (`timestamp\tcompany\tstatus`, append-only, read by `stats.mjs`) is a different shape for a different purpose, so I leaned to a separate file — but converging them may be the better call.
2. **Naming.** `liveness` is already taken in this codebase for *job posting* liveness (`check-liveness.mjs`, `liveness-core.mjs`, `liveness-browser.mjs`, `liveness-api.mjs`). This is *board existence*, so I would avoid the word entirely — hence `dead-boards`.
3. **Thresholds.** 3 misses / 30-day re-probe are guesses that felt conservative; happy to take whatever numbers you prefer.

I have this working locally with a 39-assertion test suite (most of it guarding the 404-only rule) and can open a PR against whatever shape you want. Not opening one blind, since the file layout is yours to decide.

luochen211:
Closes #2840. Persist board-level consecutive 404s for reverse ATS sweeps, skip retired boards for 30 days, re-probe them afterward, and clear the record on success. Throttles and transport failures never create dead-board records; add a regression suite for the 404-only safety contract.

## Summary by CodeRabbit

- **New Features**
  - Added persistent tracking for retired ATS boards across scans.
  - Boards receiving three recent “404 Not Found” responses are skipped on future scans.
  - Retired boards are automatically rechecked after 30 days.
  - Successful responses restore tracked boards.
  - Scan summaries and JSON output report retired-board counts.
  - Progress counters include boards skipped due to retirement status.

- **Improvements**
  - Rate limits and other non-404 failures no longer retire boards.
  - Retired-board records are validated, sorted, and safely preserved between scans.


luochen211:
验证状态：Linux/macOS、静态分析、CodeQL、升级回归和安全检查均通过。Windows 测试失败在既有的  并发锁场景：，导致 29/30 条目保留；失败不涉及本 PR 修改的扫描/更新器代码。当前账号无法对上游 workflow 执行 rerun，请维护者复跑该 job 后再判断。

luochen211:
验证状态：Linux/macOS、静态分析、CodeQL、升级回归和安全检查均通过。Windows 测试失败在既有的 agent-inbox-tests.mjs 并发锁场景：EPERM（agent-inbox.md.lock.recover），导致 29/30 条目保留；失败不涉及本 PR 修改的扫描/更新器代码。当前账号无法对上游 workflow 执行 rerun，请维护者复跑该 job 后再判断。

Scott-Emberson (commented):
This reads clean: one hermetic block in `tests/dead-boards.test.mjs` (+36, new), throwaway `mkdtemp` dir, fixed `now`, `rmSync` in `finally`, and the `recordBoardResult`/`shouldSkipDeadBoard` state machine is genuinely exercised in both directions. A no-op on either function fails at line 17 or 20, so this is not a suite that passes on nothing. Same inline `pass`/`fail` shape as its neighbours.

Four things worth tightening, none blocking. The first is a real gap for a persistence feature.

**1. `tests/dead-boards.test.mjs:12`, the load and parse half is never round-tripped.** `loadDeadBoards` is called once, on a file that does not exist yet, so it only ever returns the empty-map early exit. After `saveDeadBoards` at line 31 the file is read back with a raw regex (line 32), never through `loadDeadBoards`. The entire parse path in `dead-boards.mjs` (the `misses` guard, `Number.isFinite` on the parsed timestamp, malformed-row skipping) has zero coverage. For #2840, where the point is memory surviving between sweeps, the contract that a saved store reloads into the same skip decision is unproven. Reload after save and assert `shouldSkipDeadBoard` still returns true, and drop in one malformed row to prove it is skipped.

**2. `tests/dead-boards.test.mjs:23`, "429 does not create or clear" only proves the clear half.** When that 429 is recorded the board is already at three misses, so the branch taken is the "preserve retired" one. The other half the message claims, a 429 against a board with no record never creating one, is never run. Add a 429 (or 500) on a clean key and assert no record appears.

**3. `tests/dead-boards.test.mjs:32`, the TSV assertion under-checks the fields a reload depends on.** The regex matches only the header line and that a `lever` row begins. The `misses` value and the ISO timestamp go unchecked, and those are exactly what a reload in finding 1 has to re-parse. `saveDeadBoards` could emit `lever\tx\tNaN\tnotadate\n` and this still passes. Match the full row.

**4. `tests/dead-boards.test.mjs`, nit.** The non-404 reset of an in-progress counter (a `404` then a `500` while misses is 1 or 2) is never exercised, so the "transient error erases partial progress" path is untested. Add one if you want it pinned.

None of these block. Finding 1 is the one I would not ship without, given the feature is persistence.

luochen211:
已根据 review 补齐成功重试清除旧记录、checkpoint 持久化、持久化 round-trip、坏行忽略、30 天边界、干净 board 的 429 和部分 404 清除测试。本地测试和最新全平台 CI 均通过，请重新审查。

Scott-Emberson (approved):
Re-reviewed at 27e0a8bc against our earlier comments on tests/dead-boards.test.mjs. All four are addressed and I traced each to the code: saved state now round-trips through loadDeadBoards (line 41), the 429 create-side no-op is exercised on a fresh key (line 24), the TSV check is a fully anchored full-row match including misses and the ISO timestamp (line 34), and the transient-clears-partial-progress path is covered (lines 47-51). The block stays hermetic (single mkdtemp root, fixed now, cleanup in finally) and the sentinels are real: a no-op on recordBoardResult/shouldSkipDeadBoard still fails. CI is green on all three OSes.

One optional nit, not blocking: the malformed-row case (line 44) trips the misses guard first, so the bad-date guard is never independently exercised. A second row with a valid integer misses but a junk date would close that. Defense in depth, your call.

santifer:
The invariant at the top is the whole PR, @luochen211, and it's the right one:

> Only a real HTTP 404 may advance a miss counter. Throttles, transport failures and DNS errors mean "unknown", never "dead".

A board-existence memory that can't tell *absent* from *couldn't check* would accumulate false deaths from every rate-limit and DNS blip, and then permanently stop scanning boards that were fine — silently, since a board you no longer probe produces no evidence that it was wrong. Stating the rule as a comment above the module, rather than leaving it implicit in the control flow, is what stops someone "simplifying" the error handling later.

Persisting misses with `last_checked` and a counter rather than a boolean is also right: one 404 is a bad moment, N is a pattern, and keeping the count means the threshold can be argued without re-deriving the data.

Not merged tonight — merge ceiling — but this is a real piece of work, and along with #2863 it's what I'd point at from your set.

luochen211:
Addressed the latest review feedback in commit 1475f20: failed probes now refresh `lastChecked` for retired boards, dead-board TSV writes are same-directory atomic write-and-rename, and `.gitattributes` precedes `dead-boards.mjs` in the update manifest. Added the failed-reprobe regression assertion; `node tests/dead-boards.test.mjs` passes 11/11. Please re-review after CI.

santifer:
Sorry for the wait, @luochen211: 8 days on a PR this clean is on me. The contract you drew is the right one — only a real 404 advances the counter, throttles and transport failures mean "unknown", never "dead" — and you registered `dead-boards.mjs` in `SYSTEM_PATHS` without being asked, which is the step this repo has historically forgotten four times.

I ran your module against cases your suite doesn't cover, and it holds:

```
4 consecutive errors with NO .status  -> retired? false   (transient never kills a board)
3x404                                 -> retired? true
then an error with no .status         -> retired? true    (refreshes the clock, your last commit)
at day 31                             -> retired? false   (re-probes)
```

And `providers/_http.mjs:45` does set `err.status = res.status`, so the counter is actually fed — I checked, because a feature like this is worth exactly nothing if the status never arrives.

**One thing before I merge, and it's a single line.** `data/dead-boards.tsv` is a new file the system writes into the user layer, and `DATA_CONTRACT.md` enumerates every one of those — `data/portal-health.tsv`, `data/scan-runs.tsv`, `data/intake-state.json`, all of them, with who writes it and whether it's safe to delete. Yours isn't there. That table is the whole reason the boundary holds, so please add the row:

```
| `data/dead-boards.tsv` | Boards that returned three consecutive 404s during reverse ATS sweeps (written by `scan-ats-full.mjs`; re-probed after 30 days — safe to delete, the next sweep rebuilds it) |
```

While you're in there, worth a look at `data/portal-health.tsv` right above it: it tracks consecutive reachability for the portals *the user configured*, which is close enough in spirit that the next person will wonder why there are two. A clause in your row saying this one is for the mass reverse sweep, not the user's own portals, saves that confusion permanently.

Two non-blocking notes from the same run, take or leave:

- `boardKey({})` returns `""`, and I confirmed three different unidentified boards then share one row — three 404s retire the empty key for everyone. Real-world impact is about zero (an entry with neither `careers_url` nor `name` can't be fetched anyway, and `loadDeadBoards` drops the empty row on reload so it never persists), but an early `if (!board) return;` in `recordBoardResult` costs nothing.
- `providers/glints.mjs:176` re-throws as `new Error(\`glints: HTTP ${err.status} …\`)`, which **drops** `.status`. Glints boards will therefore never accumulate misses. Pre-existing, not yours, and harmless (they just behave as they do today) — but it's the kind of thing worth knowing about the feature you just built.

Add the DATA_CONTRACT row and this goes in.

luochen211:
Added the requested DATA_CONTRACT entry in 581a352. It records data/dead-boards.tsv as mass reverse-ATS sweep state, explicitly distinguishes it from the user-configured portal-health.tsv, and documents the 30-day re-probe and safe-delete behavior. Local dead-board suite: 11 passed; the latest macOS/Linux/Windows and security checks are green. Please re-review.

luochen211:
Addressed the remaining DATA_CONTRACT request in 581a352: registered `data/dead-boards.tsv`, scoped it to mass reverse ATS sweeps, and distinguished it from configured portal health. Local dead-board tests passed 11/11, and the refreshed PR checks are green. @santifer please re-review.

santifer:
The DATA_CONTRACT registration in 581a352 is exactly what I asked for, @luochen211: thanks. A deeper pass found two functional issues to fix before merge, both verified against your branch:

1. **`saveDeadBoards` can abort a whole sweep.** The call at scan-ats-full.mjs:925 is unguarded, and saveDeadBoards itself uses bare `mkdirSync`/`writeFileSync`/`renameSync` (dead-boards.mjs:55-63). A permissions issue or a transient Windows EPERM on rename throws before the summary and result persistence: a best-effort cache save would discard a scan that found real matches. Wrap it best-effort, the same way the checkpoint write on that path is guarded.

2. **Skipped retired boards still count as scanned.** `totalCompaniesScanned` is bumped by the full `entries.length` up front (the resolver-outage branch even documents that pattern at :930-932), but the `shouldSkipDeadBoard` early-return at :833-835 never corrects it. Once dead-boards.tsv accumulates, the summary and Explore UI overstate coverage. Either subtract the skips or report them only through the separate retired-board counter you already added.

Both are small, and the feature itself is the right call: a board that 404s forever deserves a memory. Happy to merge with those two closed.

Scott-Emberson (changes requested):
Re-reviewed at `e1e7e27d`. My approval was at `27e0a8bc` back on 2026-08-14, and two commits since land in my files: `1475f200` in `tests/dead-boards.test.mjs` and `e1e7e27d` in `tests/scan-ats-full-outage-checkpoint.test.mjs`. Only those two files are mine. `DATA_CONTRACT.md` and `update-system.mjs` are santifer's, so I have read them for context and judged neither.

Both suites are green at `e1e7e27d`, run locally: `dead-boards.test.mjs` 11 checks, `scan-ats-full-outage-checkpoint.test.mjs` 7 checks, 0 failed. CI is green on all thirteen.

Two of the three new checks are the real thing:

- `tests/dead-boards.test.mjs:48-51`. Asserting `lastChecked === failedProbeAt` **and** that the board still skips at `failedProbeAt + 1` pins both halves of the new `else` branch in `recordBoardResult`. A 429 against a retired board now refreshes its clock instead of being ignored, and the check would notice either way it could go wrong.
- `tests/scan-ats-full-outage-checkpoint.test.mjs:197-220`. `companiesScanned === COMPANIES - 1 && retiredBoardsSkipped === 1` is properly discriminating: 119 and 1, not "some number came back". This is the check I would have asked for on the counting change, and it is already here.

**The third one cannot fail, and that is what I am requesting changes on.** `tests/scan-ats-full-outage-checkpoint.test.mjs:222-241`, major, CONFIRMED.

The block forces the cache write to fail by putting a directory where the temp file goes, then asserts `JSON.parse(out).companiesScanned === COMPANIES`. But a sweep with no cache failure at all returns exactly the same 120. I replaced the forcing line with a comment and re-ran:

```
    // MUTANT: cache write is NOT forced to fail
✅ dead-board cache failure is best-effort and does not abort the sweep
```

Green, with the failure path never entered and no warning printed. The check currently proves that a sweep completes, which the block two above it already proves. If `saveDeadBoards()` ever stops writing through `${file}.tmp`, the `mkdirSync` stops blocking anything and this check keeps passing while testing nothing.

The control is free, because the write leaves observable evidence either way:

```
normal write    -> data/dead-boards.tsv exists: true
forced failure  -> EISDIR, data/dead-boards.tsv exists: false
```

So folding `!existsSync(join(dir, 'data', 'dead-boards.tsv'))` into the pass condition proves the write really failed *and* that the sweep survived it, which is the whole claim. Asserting on the `⚠ dead-board cache write failed` line would be the other way, but `run()` in `tests/helpers.mjs:186` only keeps stderr on a failed child, so that one needs a helper change and the file check does not.

Worth doing rather than waving through: this test is the only thing standing between a future refactor and a silent regression of the best-effort behaviour that `e1e7e27d` exists to add.

Requesting changes on that one condition. Everything else from my side is good, and I will turn it round quickly once it is in. Separately, the branch is currently conflicting with main, so it needs a rebase before it can land regardless.

luochen211:
The two remaining functional asks are addressed on the current head (599a5fc):

- dead-board persistence is best-effort, so a cache write/rename failure cannot discard a completed sweep
- retired-board skips no longer inflate the scanned-company count, with the checkpoint count persisted and covered

All review threads are resolved and the current check rollup is green; the branch is mergeable and only behind main without conflicts. @santifer please re-review.

santifer:
Nice work, @luochen211. Direction is solid and the 404-only safety contract is exactly right; one thing to adjust before it lands, and it's timing, not design:

- `scan-ats-full.mjs` in this PR declares `const DEAD_BOARDS_PATH = 'data/dead-boards.tsv'` as a bare cwd-relative path. #3511 (merging today) moves every user-layer path in that same file to `path.join(DATA_ROOT, ...)`, because a bare path silently reads and writes the wrong place under a configured data root. The new file should follow the same rule, and the branch will need a rebase once #3511 is in since both touch `scan-ats-full.mjs`.

Everything else reads as ready: the DATA_CONTRACT row, the best-effort persistence, the checkpoint count. Happy to merge once the path goes through `DATA_ROOT`.

luochen211:
Synced this branch with the current canonical `main` in merge commit 34e1c98 and resolved the scanner conflict while keeping the dead-board changes. `DEAD_BOARDS_PATH` now resolves through `DATA_ROOT` as requested.

Validation: dead-board, outage/checkpoint, and data-root path suites pass; `node test-all.mjs --quick` passes with 0 failures; `node upgrade-tests.mjs --pr-gate` is GREEN. All GitHub checks at the new head are green. This remains pending maintainer review/merge.

luochen211:
Scott-Emberson 的 outage test concern is already addressed in the current head (34e1c98): the fixture creates data/dead-boards.tsv.tmp as a directory, and the pass condition asserts both companiesScanned === COMPANIES and that data/dead-boards.tsv does not exist. Running node tests/scan-ats-full-outage-checkpoint.test.mjs reproduced the forced EISDIR warning and passed the cache-failure check. The branch is synced with current career-ops-hq/main and all 11 GitHub checks are green. Please re-review this resolved request when convenient.

luochen211:
Synced the branch with current career-ops-hq/main (head 8bb7ce8). The cache-failure regression already includes the requested discriminator: dead-boards.tsv.tmp is forced to be a directory, the run emits EISDIR, and the test asserts dead-boards.tsv does not exist while companiesScanned remains 120. Local dead-board/outage/data-root tests, upgrade gate, and quick suite pass (7770 passed, 0 failed, 13 warnings); all 11 GitHub checks are green. Please re-review the resolved request.

santifer:
Merged, @luochen211: board-level 404 memory for reverse sweeps with the safety contract pinned in tests (only a real 404 advances the counter; throttles and transport errors are "unknown", never "dead"), best-effort persistence so a cache write can't discard a finished sweep, and the new `data/dead-boards.tsv` row in DATA_CONTRACT. Closes #2840. 🚀

PR #2870, opened 2026-08-14
fix(classify-tier): validate CLI flags and add a usage block

### LINKED ISSUE #2852: classify-tier.mjs has no --help: asking for help returns a seniority tier

`classify-tier.mjs` takes a job title as its only positional argument. Asking it for help therefore classifies the *word* `--help` as a job title:

```
$ node classify-tier.mjs --help
mid
```

Exit 0, no usage, and a seniority answer for a flag. There is no way to discover what the script accepts other than opening it.

### What to do

1. Add a `USAGE` string documenting the positional argument and any flags the script already supports.
2. Handle `--help` / `-h` **first**, before the title is read: print `USAGE`, exit 0.
4. Keep the existing behaviour for real titles unchanged.

### Acceptance

- `node classify-tier.mjs --help` prints usage and exits 0
- `node classify-tier.mjs -h` does the same
- `node classify-tier.mjs --bogus "Senior Engineer"` exits non-zero and names `--bogus`
- `node classify-tier.mjs "Senior Software Engineer"` still prints `senior`, byte-identical to today

That fourth one is the important one. The whole point is that nothing about real usage moves.

### Pattern to mirror

```js

const USAGE = `Usage:
  node <script>.mjs …`;

```

```js
```

### Why it matters

A title classifier that answers `mid` when you ask it for help is not just undiscoverable, it is quietly wrong: a script or a wrapper passing an unrecognised flag gets a confident tier back instead of an error, and nothing downstream can tell that answer from a real one.

dchaudhari7177:
Closes #2852.

`classify-tier.mjs` takes a job title as its only positional argument, so asking it for help classified the *word* `--help`:

```
$ node classify-tier.mjs --help
mid
```

Exit 0, no usage, a seniority answer for a flag.

## What changed

Argv now goes through `validateFlags()` from `lib/cli-flags.mjs` (#2775) rather than a fourth hand-rolled copy of the check — same way `scan.mjs` consumes it. That gets the ordering right for free: unrecognized flags are rejected *before* `--help`, so `--help --bogus` still errors rather than exiting 0 having never looked at `--bogus` (the reversal CodeRabbit caught on #2745 and #2746).

The no-args branch previously printed its own inline usage text. It now prints the same `USAGE` constant, so the two can't drift.

## Acceptance

| Check | Result |
|---|---|
| `node classify-tier.mjs --help` | prints usage, exit 0 |
| `node classify-tier.mjs -h` | same, exit 0 |
| `node classify-tier.mjs --bogus "Senior Engineer"` | `Error: unrecognized flag(s): --bogus. Valid flags: --test, --help, -h`, exit 1 |
| `node classify-tier.mjs --help --bogus` | exit 1 — the flag check runs first |
| `node classify-tier.mjs "Senior Software Engineer"` | `senior`, exit 0 — unchanged |
| `node classify-tier.mjs --test` | `All tests passed successfully!` |

Real usage is byte-identical to today; only the flag paths move.

## Local test run

`node test-all.mjs` does not complete on this machine — `tests/intake.test.mjs:187` calls `fs.symlinkSync` and Windows refuses it without developer mode (`EPERM`). I confirmed that is **pre-existing**: stashing this change and re-running `test-all.mjs` on pristine `main` fails at the identical line. Nothing in this diff is reachable from that test. The suite up to that point, and `classify-tier --test`, are green.

## Summary by CodeRabbit

* **New Features**
  * Added command-line validation for supported options: `--test`, `--help`, and `-h`.
  * Added centralized usage guidance for the command-line tool.

* **Bug Fixes**
  * Unknown command-line flags are now rejected before processing begins.


dchaudhari7177:
Flagging the red `test (windows-latest)` here: **3853 passed, 1 failed**, and the failure is unrelated to this PR.

It's the inbox concurrency test hitting a Windows file-lock race:

```
item-6 exited 1: Error: EPERM, Permission denied:
  ...\Temp\inbox-concurrent-Zhgy35\agent-inbox.md.lock
```

This PR only touches `classify-tier`'s flag parsing and usage block — it doesn't go near the inbox writer or its lock. The four sibling PRs from the same batch (#2871–#2874) are green on the same base, which points at flakiness in that test on Windows rather than anything in this change.

I don't have rerun rights on the repo, so I can't retry it myself — a re-run of the failed job should clear it. Happy to open a separate issue for the lock race if it's not already known; the EPERM on `.lock` under `%TEMP%` looks like a genuine Windows-specific race in the retry path rather than pure CI noise.

santifer:

santifer:
Same verification that decided the invite-match race, applied here: your KNOWN_FLAGS list matches classify-tier's real surface exactly. That's the standard, and this one meets it. Merged! 🚀

PR #2871, opened 2026-08-14
fix(funnel-velocity): validate CLI flags and add a usage block

### LINKED ISSUE #2853: funnel-velocity.mjs has no --help and silently ignores mistyped flags

`funnel-velocity.mjs` recognizes no flags at all. Asking for help just runs the report:

```
$ node funnel-velocity.mjs --help
{
  ...the normal JSON report...
```

### What to do

### Acceptance

- `node funnel-velocity.mjs --help` prints usage and exits 0
- `node funnel-velocity.mjs --bogus` exits non-zero and names `--bogus`

### Pattern to mirror

```js

const USAGE = `Usage:
  node <script>.mjs …`;

```

```js
```

### Why it matters

dchaudhari7177:
Closes #2853.

`node funnel-velocity.mjs --help` ran a full analysis and printed the JSON report at exit 0. `--bogus` did the same — a mistyped flag was indistinguishable from a normal run.

## What changed

Argv goes through `validateFlags()` from `lib/cli-flags.mjs` (#2775) rather than a hand-rolled copy. `--benchmarks` is declared in `valueFlags` so a path beginning with a dash is left for the script's own validation instead of being reported as an unrecognized flag.

The call sits at the top of `main()`, not at module scope: `computeFunnel`/`analyze` are imported by other modules, and validating on import would make a library consumer exit the process.

## Acceptance

| Check | Result |
|---|---|
| `node funnel-velocity.mjs --help` | prints usage, exit 0 |
| `node funnel-velocity.mjs --bogus` | `Error: unrecognized flag(s): --bogus. …`, exit 1 |
| normal run byte-identical | ✅ `diff` of stdout before/after is empty, for both the default JSON and `--summary` |
| `node funnel-velocity.mjs --self-test` | OK |

The byte-identical check is the one that matters here, so I ran it explicitly rather than eyeballing: captured stdout on pristine `main`, applied the change, captured again, `diff -q` clean on both output modes.

## Local test run

`node test-all.mjs` does not complete on this machine — `tests/intake.test.mjs:187` calls `fs.symlinkSync`, which Windows refuses without developer mode (`EPERM`). Confirmed pre-existing: pristine `main` fails at the identical line. Nothing in this diff is reachable from that test.

## Summary by CodeRabbit

- **Bug Fixes**
  - Added command-line flag validation.
  - Unrecognized or malformed options are now rejected with clear usage guidance.
  - Added support for displaying help information through the command-line validator.

### DIFF
--- a/funnel-velocity.mjs
+++ b/funnel-velocity.mjs
@@ -42,6 +42,7 @@ import { computeFunnel, computeTrackerStats } from './stats.mjs';
 import { resolveColumns, parseTrackerRow } from './tracker-parse.mjs';
 import { resolveTrackerPath, loadCanonicalStates, resolveCanonicalState } from './tracker-utils.mjs';
 import { parseAppliedDate, normalizeStatus } from './followup-cadence.mjs';
+import { validateFlags } from './lib/cli-flags.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
 const STATES_FILE = join(CAREER_OPS, 'templates/states.yml');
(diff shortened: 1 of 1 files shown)


santifer:

dchaudhari7177:
Closing this — it's been superseded, and by a better implementation than mine.

`b6ebe08` ("fix(funnel-velocity): validate CLI flags via the shared helper") landed on `main` while this was open and covers the same ground as #2853: `funnel-velocity.mjs` now imports `validateFlags` from `lib/cli-flags.mjs`, declares `KNOWN_FLAGS` / `VALUE_FLAGS` and a `USAGE` block, and calls it with `requireOperand: true`, with flag tests in `tests/funnel-velocity-flags.test.mjs`.

That is the right shape. This PR hand-rolled the same validation inside `funnel-velocity.mjs` instead of going through the shared helper, so rebasing it onto `main` would mean re-litigating a decision that has already been made — and made better.

Nothing here is worth salvaging on top of what landed. Sorry for the noise on the queue.

PR #2872, opened 2026-08-14
fix(invite-match): validate CLI flags and add a usage block

### LINKED ISSUE #2854: invite-match.mjs has no --help and silently ignores mistyped flags

`invite-match.mjs` recognizes no flags at all. Asking for help just runs the matcher:

```
$ node invite-match.mjs --help
{
  ...the normal JSON result...
```

### What to do

4. Leave the matcher itself untouched.

### Acceptance

- `node invite-match.mjs --help` prints usage and exits 0
- `node invite-match.mjs --bogus` exits non-zero and names `--bogus`

The third one is the criterion that matters: this is a flag-handling fix, not a change to what the script matches.

### Pattern to mirror

```js

const USAGE = `Usage:
  node <script>.mjs …`;

```

```js
```

### Why it matters

Silently ignoring a mistyped flag is worse than rejecting it. The command exits 0, prints a plausible result, and the person believes they got the thing they asked for. An error message costs one line and removes the whole class.

dchaudhari7177:
Closes #2854.

`node invite-match.mjs --help` fell through to the matcher and printed a JSON match report at exit 0. `--bogus` did the same.

## What changed

Argv goes through `validateFlags()` from `lib/cli-flags.mjs` (#2775). `--file` and `--id` are declared in `valueFlags` so their operands aren't misread as unrecognized flags.

**One thing worth flagging in review:** the call is inside the CLI-only guard at the bottom, *not* at module scope next to the other argv reads. This file is imported for `normalizeCompanyName`/`companySimilarity` by three other scripts —

- `scan.mjs:53`
- `detect-reposts.mjs:30`
- `tracker-sync-check.mjs:63`

— and `validateFlags` reads `process.argv` and calls `process.exit()`. At module scope it would validate the *importer's* argv against invite-match's flag list, so `node scan.mjs --dry-run` would exit 1 with "unrecognized flag: --dry-run" while importing a name-normalizer. I verified the guard holds:

```
$ node -e "process.argv=['node','other.mjs','--dry-run','--company','X']; import('./invite-match.mjs').then(m=>console.log('imported ok'))"
imported ok
```

The pre-existing module-scope `--file requires a path` / `--id requires a tracker #` checks are untouched — they only fire when those flags are actually present, so they were already import-safe.

## Acceptance

| Check | Result |
|---|---|
| `node invite-match.mjs --help` | prints usage, exit 0 |
| `node invite-match.mjs --bogus` | `Error: unrecognized flag(s): --bogus. …`, exit 1 |
| normal run byte-identical | ✅ `diff` clean before/after on a real invite string |
| `node invite-match.mjs --self-test` | 110 passed, 0 failed |
| `node invite-match.test.mjs` | 89 passed, 0 failed |

## Local test run

`node test-all.mjs` does not complete on this machine — `tests/intake.test.mjs:187` calls `fs.symlinkSync`, refused by Windows without developer mode (`EPERM`). Pre-existing: pristine `main` fails at the identical line. The two suites that do cover this file are green above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for command-line flags.
  * Unrecognized or improperly formatted options now display usage guidance instead of being processed.

### DIFF
--- a/invite-match.mjs
+++ b/invite-match.mjs
@@ -38,6 +38,7 @@ import { join, dirname } from 'path';
 import { fileURLToPath, pathToFileURL } from 'url';
 import { execFileSync } from 'child_process';
 import { resolveColumns, parseTrackerRow } from './tracker-parse.mjs';
+import { validateFlags } from './lib/cli-flags.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
 const APPS_FILE = existsSync(join(CAREER_OPS, 'data/applications.md'))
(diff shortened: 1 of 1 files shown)


santifer:

santifer:
Thanks for this one too, @dchaudhari7177 — closing it in favor of #2958, and the deciding detail is worth sharing because it's subtle: your KNOWN_FLAGS list included `--json`, which invite-match doesn't actually accept (it only *passes* `--json` internally to set-status). #2958's list matched the script's real surface exactly and carried a test, so it won the race despite arriving later. Your #2874/#2873 from this same family are already merged, and #2870 (classify-tier) is still open and being reviewed on its own merits: this close is about one flag on one file, not about your work.

PR #2873, opened 2026-08-14
fix(check-table-freshness): validate CLI flags and add a usage block

### LINKED ISSUE #2855: check-table-freshness.mjs has no --help and silently ignores mistyped flags

`check-table-freshness.mjs` recognizes no flags at all. Asking for help just runs the report:

```
$ node check-table-freshness.mjs --help
{
  ...the normal JSON report...
```

### What to do

### Acceptance

- `node check-table-freshness.mjs --help` prints usage and exits 0
- `node check-table-freshness.mjs --bogus` exits non-zero and names `--bogus`

### Pattern to mirror

```js

const USAGE = `Usage:
  node <script>.mjs …`;

```

```js
```

### Why it matters

dchaudhari7177:
Closes #2855.

`node check-table-freshness.mjs --help` ran a full freshness scan and printed the JSON report at exit 0. `--bogus` did the same — so a mistyped `--max-age-month` silently reported against the 12-month default instead of the threshold the caller asked for.

## What changed

Argv goes through `validateFlags()` from `lib/cli-flags.mjs` (#2775). This file already imported `flagValue` from that module, so it's the same helper, not a new dependency. `--max-age-months` and `--today` are declared in `valueFlags`.

The call sits inside the existing CLI-only guard so the module stays importable — `parsePositiveInt` and the parsing helpers are exported.

## Acceptance

| Check | Result |
|---|---|
| `node check-table-freshness.mjs --help` | prints usage, exit 0 |
| `node check-table-freshness.mjs --bogus` | `Error: unrecognized flag(s): --bogus. …`, exit 1 |
| `node check-table-freshness.mjs --help --bogus` | exit 1 — flag check runs first |
| normal run byte-identical | ✅ see note below |
| `node check-table-freshness.mjs --self-test` | 46 passed, 0 failed |

On byte-identical: `--summary` output diffs clean. The default JSON differs by exactly one line, `"generatedAt"`, which is a wall-clock stamp and differs between any two runs — excluding that line the two are identical, and exit codes match (0 → 0) on both modes.

## Local test run

`node test-all.mjs` does not complete here — `tests/intake.test.mjs:187` calls `fs.symlinkSync`, refused by Windows without developer mode (`EPERM`). Pre-existing: pristine `main` fails at the identical line. This script's own `--self-test`, which `test-all.mjs` shells out to, is green above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for command-line options before running freshness checks.
  * Unknown options and invalid option combinations are now rejected with usage guidance.
  * Help requests containing unrecognized options are handled correctly.

### DIFF
--- a/check-table-freshness.mjs
+++ b/check-table-freshness.mjs
@@ -54,7 +54,7 @@ import { readFileSync, readdirSync, existsSync } from 'fs';
 import { join, dirname } from 'path';
 import { fileURLToPath, pathToFileURL } from 'url';
 import * as yaml from 'js-yaml';
-import { flagValue } from './lib/cli-flags.mjs';
+import { flagValue, validateFlags } from './lib/cli-flags.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
 const TEMPLATES_DIR = join(CAREER_OPS, 'templates');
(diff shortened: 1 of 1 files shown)


santifer:
The detail that makes this one land, @dchaudhari7177, is the line you put at the bottom of `USAGE`:

> Exits 1 when any row is expired (past next_effective without re-verification).

That's the thing a caller most needs and would otherwise have to read the source for — and it matters more here than in the siblings, because this script is the kind that ends up in a pipeline where somebody checks `$?`. A mistyped `--max-age-month` (singular) silently reporting again on the default 12 is exactly the failure that produces a green pipeline nobody questions.

Same shared helper, same `--help --bogus` ordering, value flags declared. Consistent with the other three without being a copy of them.

Merged 🚀

PR #2877, opened 2026-08-14
fix(funnel-velocity): count web-recorded transitions as observed

santifer:
Requested by the web side while reviewing #2715, which makes the web app write status transitions to the ledger.

`funnel-velocity.mjs` did not know the `web` source string, so every web-driven transition landed in `unknownSources` and was excluded from day-math: parsed, counted, flagged as suspect, and silently absent from the one number the file exists to produce.

The exclusion was wrong on the merits rather than a plain omission. `DAY_MATH_SOURCES` excludes `backfill` and `manual` for a reason stated right there in the code: they are reconstructed after the fact, not observed. A web transition is observed live, at the moment the user makes it: the same event as a `set-status`, entering through a different door. By that rule it belongs in both sets.

Adds self-test coverage with its own fixture. The fixture is separate on purpose: placed in the shared one, its two transitions entered the velocity median and turned three unrelated checks red. Verified by mutation - removing `web` from either set turns the new checks red and names which one.

Unblocks #2715.

## Summary by CodeRabbit

* **New Features**
  * Added web activity as a recognized source for funnel velocity calculations.
  * Web-based status transitions are now included in daily velocity metrics.

* **Bug Fixes**
  * Prevented valid web transitions from triggering unknown-source warnings.
  * Continued excluding reconstructed and manually entered sources from calculations.

### DIFF
--- a/funnel-velocity.mjs
+++ b/funnel-velocity.mjs
@@ -51,10 +51,13 @@ const summaryMode = args.includes('--summary');
 const selfTestMode = args.includes('--self-test');
 
 const DATE_RE = /^\d{4}-\d{2}-\d{2}$/;
-const VALID_SOURCES = new Set(['set-status', 'correction', 'backfill', 'manual']);
+const VALID_SOURCES = new Set(['set-status', 'correction', 'backfill', 'manual', 'web']);
 // Sources whose dates are trusted for day-math. backfill/manual are parsed and
 // counted but excluded: they are reconstructed after the fact, not observed.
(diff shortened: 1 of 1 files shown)


santifer:
Closing my own PR as superseded: main already carries this exact change (funnel-velocity.mjs line 67 includes web in DAY_MATH_SOURCES, with a better-worded comment than mine), landed through another branch while this one sat. Verified via rebase collision before closing.

PR #2878, opened 2026-08-14
docs(job-boards): document the agentic-jobs and wttj providers

santifer:
`providers/agentic-jobs.mjs` and `providers/wttj.mjs` both ship, but neither had a row in `docs/SUPPORTED_JOB_BOARDS.md`. The only way to find out they existed was to read the directory listing.

Each row records what a user needs before configuring it: the endpoint, whether auth is involved, how pagination is bounded, and the config key. For WTTJ that includes the part that would otherwise look like a bug: the public app id and search key rotate, so they are fetched from the site's own `/api/env` on every run, and the board is global enough that the provider requires an explicit `queries` block instead of pulling everything.

No code changes. `test-all` 3860/0.

## Summary by CodeRabbit

* **Documentation**
  * Added Agentic Engineering Jobs as a supported job board.
  * Added Welcome to the Jungle as a supported job board, including its search integration details.


santifer:
Self-authored docs; merging on green.

PR #2879, opened 2026-08-14
docs(tests): say that any new test goes in its own file, not a numbered section

santifer:
The auto-discovery glob is `tests/**/*.test.mjs`, but both places that document it only mention providers. A contributor adding a test for anything else reads that, reasonably concludes it does not apply to them, and appends a numbered section to `test-all.mjs` — which is what the file is full of, so it looks like the intended pattern.

**The cost is not tidiness.** A numbered section means editing the final lines of a 14,000-line file and picking a global label by hand, with no way to reserve one. Six open PRs did exactly that at once, and four picked the same number (`60a`). Each merge forces a rebase on the other five: roughly fifteen rebases and six serialized CI runs for six lines of test code. A separate file collides with nobody, so those six could all land in parallel.

Nothing changes mechanically — the glob already worked and 77 suites already live that way. Only the prose was narrower than the mechanism.

`test-all` 3863/0.

## Summary by CodeRabbit

* **Documentation**
  * Updated contribution guidance to require new tests to be added as separate auto-discovered files.
  * Documented automatic test discovery and clarified that separate test files help reduce merge conflicts.
  * Replaced previous provider-specific test organization guidance.


santifer:
Self-authored docs; merging on green.

PR #2884, opened 2026-08-14
fix(liveness): read same-origin iframes so embedded postings aren't read as empty

artemtrofymenko:
## What does this PR do?

Some ATS render the whole posting inside a same-origin iframe and leave the top-level document as a shell. iCIMS is the reference case: on a **live** posting the outer document holds about 13 characters and no apply control, so `classifyLiveness` reaches `insufficient_content` and returns `expired`.

That is the expensive direction. An `expired` verdict is written to scan-history as `skipped_expired`, and every later scan dedup-filters that URL, so a live job disappears permanently and silently.

## How it was found

A 1555-URL liveness sweep. **92 live iCIMS postings were classified `expired`** this way, across many tenants (didiglobal, peraton, urbanscience, enfra and others). Four of five spot-checked by hand carried a full job description with requisition ID and posting date.

## The fix, two parts

**1. Merge text and apply controls from same-origin child frames** before classifying. `sameOrigin` is deliberately strict, so `about:blank`, `data:` frames, tag managers and ad iframes contribute nothing.

**2. Wait for the frame to populate.** This part is not optional and was not obvious: the frame *attaches* immediately but *fills* late. Measured through the real guarded page:

```
t=2000ms  0 chars
t=2500ms  0 chars
t=3000ms  0 chars
t=4000ms  3887 chars
```

`HYDRATION_WAIT_MS` is 2000, so without the wait the frame is read empty and part 1 changes nothing at all. The poll is bounded (6s cap, 500ms interval) and is only entered when a same-origin child frame exists, so ATS that render inline pay nothing.

## Why this cannot resurrect a dead posting

A removed iCIMS job answers **HTTP 410** at the top level, so it short-circuits on status long before any content check. Verified three ways: two fabricated job ids on live tenants, and one genuinely dead posting. All returned 410.

Its error frame also carries **zero** apply-matching controls, and the existing `HARD_EXPIRED` pattern `job.*no longer open` already matches the iCIMS error text even when an apply control is present. `liveness-core.mjs` therefore needed no change.

## Verification

End to end on a 7-URL set with controls in both directions:

| URL | before | after |
|---|---|---|
| 2 live iCIMS (uncertain) | `uncertain` | `active` |
| 2 live iCIMS (false expired) | `expired` | `active` |
| 2 genuinely dead iCIMS | `expired` (410) | `expired` (410) |
| non-iCIMS control | `active` | `active` |

`tests/liveness-core.test.mjs`: 12 passed, 0 failed.

## Scope and limits

Only iCIMS was measured. Other iframe-embedding ATS may benefit, but that is untested and is not claimed here. Happy to narrow this to an iCIMS-specific path if you would rather not change frame handling globally.

No test is included because the behaviour needs a real browser and a live iframe; if you want one, tell me the shape you would accept (a Playwright fixture, or a seam so the frame list can be injected) and I will add it.

## Summary by CodeRabbit

* **New Features**
  * Liveness classification now supports content from same-origin iframes.
  * Asynchronously loaded content in qualifying frames is included through bounded polling.
  * Strict same-origin validation excludes cross-origin iframe content.
  * Detached or inaccessible frames are safely ignored.

* **Bug Fixes**
  * Improved classification reliability for delayed iframe content and pages without frame support.
  * HTTP 410 responses now take precedence over iframe-based classification.
  * Polling stops when no qualifying frames are available.


artemtrofymenko:
Thanks, both findings taken. Pushed in `a5a51a1`, and writing the test you asked for is what surfaced a third problem that was mine.

**The regression your test request exposed.** `checkUrlLiveness` was calling `page.frames()` and `page.mainFrame()` unconditionally, but the doubles in `test-all.mjs` implement only `goto`/`url`/`evaluate`. On those the check died with `page.frames is not a function`, so a live page returned `uncertain/navigation_error` instead of `active`:

```
live fakePage      -> {"result":"uncertain","code":"navigation_error","reason":"navigation error: page.frames is not a function"}
challenge fakePage -> {"result":"uncertain","code":"navigation_error","reason":"navigation error: page.frames is not a function"}
```

Frame aggregation is an enhancement, so it is now skipped when the page object cannot enumerate frames and the top-level verdict stands. It was invisible on my machine because `test-all.mjs` aborts earlier in `tests/intake.test.mjs`, which needs symlink privileges Windows withholds, so the liveness section was never reached. Sorry, that one was on me and would have landed in your CI.

**Poll every qualifying frame — adopted.** I measured it before changing it, in case waiting for all frames meant paying the whole 6s cap on pages with a permanently empty same-origin sibling. Across five iCIMS tenants there is exactly one qualifying child frame per page, and non-iCIMS pages have none:

```
icims  qualifying=1  chars=[3898]
icims  qualifying=1  chars=[7831]
icims  qualifying=1  chars=[11260]
icims  qualifying=1  chars=[8887]
icims  qualifying=1  chars=[8687]
other  qualifying=0  chars=[]
other  qualifying=0  chars=[]
```

So in practice it is the same loop, and the race is real and cheap to remove. Adopted as suggested.

**Tests — added, four cases.** The one that matters most is the delayed frame, because it reproduces the reason the first version of this fix was a silent no-op: the frame attaches immediately but populates late, so a reader that does not wait sees an empty document and nothing changes.

- a same-origin posting frame that fills only after a few polls
- a cross-origin frame carrying text and an apply control, which must **not** rescue an empty shell
- HTTP 410 together with a frame that does carry an apply control, which must stay `expired`
- a page object with no `frames()`/`mainFrame()`, which must still return a top-level verdict

One note in case it saves someone time: my first double discriminated the two `evaluate` calls on `innerText`, which silently fails because the apply-control extractor mentions `innerText` too. It discriminates on `querySelectorAll` now.

Re-verified live after all three changes: two iCIMS postings `active`, a genuinely dead one still `expired` via HTTP 410, non-iCIMS control unchanged.

artemtrofymenko:
Taken, in `ed0dbe7`. I checked whether the delay was real before changing anything, because on the dead iCIMS pages I had looked at, the error frame was already populated, which would have made this theoretical. It is not theoretical.

Two dead postings, end to end, before:

```
careers-indeed    verdict=http_gone  elapsed=5822ms  qualifying frames=1 chars=[41704]
careers-peraton   verdict=http_gone  elapsed=3314ms  qualifying frames=1 chars=[43147]
```

The spread between them is poll iterations: the error page renders into the same iframe and takes its own time to fill, so the check waits for content that cannot change a 410. After:

```
careers-indeed    verdict=http_gone  elapsed=2716ms  qualifying frames=1 chars=[0]
careers-peraton   verdict=http_gone  elapsed=2772ms  qualifying frames=1 chars=[0]
```

`chars=[0]` is the part that shows it working: we now leave before the error page fills, rather than after.

One deliberate choice worth flagging for review. I did not write `status === 404 || status === 410` in the browser layer. `classifyLiveness` owns that rule, so the short-circuit calls it with the top-level reading and keys off the returned code:

```js
const topLevelVerdict = classifyLiveness({ status, requestedUrl: url, finalUrl, bodyText, applyControls });
if (topLevelVerdict.code === 'http_gone') return topLevelVerdict;
```

A hand-written status comparison here would be a second copy of a rule that already exists in one place, and the two would eventually disagree. If you would rather have the cheaper explicit check, say so and I will switch it.

The test asserts the absence of the wait, not just the verdict, and I verified it can fail: with the short-circuit removed the same case spins the poll to its deadline, with it the count is 0.

Live re-check after this change is unchanged where it should be: two iCIMS postings `active`, the dead one `expired` via 410, non-iCIMS control `active`.

santifer:
`sameOrigin` being **deliberately strict** is the decision I'd have wanted argued, and you argued it:

> about:blank, data: frames, tag managers and ad iframes all fail it, so only the ATS's own embedded document contributes text and apply controls

Reading child frames is exactly the kind of change that quietly widens what counts as "the posting". A permissive origin test would start pulling text from a consent banner or an ad frame into the liveness verdict, and the result would look *more* confident while being less correct. Strict-by-default with the failure cases enumerated is the right side to err on.

Bounding the extra wait to 6 s, only spent when such a frame exists at all, is the other half — a liveness check that gets slower for every page in order to handle some pages is a bad trade at scan scale.

Held CI approved at the start of tonight's session; it's green. Not merged tonight (merge ceiling), near the top next cycle alongside #2885.

santifer:
As reviewed: sameOrigin deliberately strict is the right side to err on, and bounding the extra wait to 6s only when such a frame exists keeps the check honest at scan scale. Merged! 🚀

PR #2885, opened 2026-08-14
fix(liveness): a dead third-party host must not poison the whole verdict

artemtrofymenko:
## What does this PR do?

The egress guard intercepts every request and, when a host fails to resolve, sets `page._blockedByGuard`. That flag makes `checkUrlLiveness` return `uncertain` for the page **regardless of what the page actually showed**.

Analytics and ad vendors get shut down. When one does, its script tag stays in thousands of career pages forever, and from then on every posting on those pages reports `uncertain` — not because anything is wrong with the posting, but because a tracker that died years ago no longer resolves.

## How it was found

A 217-URL recheck. **78 live postings came back `uncertain`** for exactly this reason:

| count | host |
|---|---|
| 53 | `personalisation.visitorqueue.com` |
| 17 | `s7.addthis.com` (AddThis shut down in 2023) |
| 5 | `cdn.fluidads.com` |
| 2 | `d26x5ounzdjojj.cloudfront.net` |
| 1 | `secure-ds.serving-sys.com` |

One was opened by hand to confirm the posting itself was fine: 11,178 characters of live job description and a working apply control, reported `uncertain` because of a dead visitor-tracking script.

## The fix

It distinguishes the two things the guard currently conflates:

- **"this host does not exist"** — a dead third-party script. The request is still aborted, but the verdict is left alone.
- **"this host resolves into private space"** — an actual egress-guard hit, which still poisons the verdict exactly as before.

The DNS-empty error is tagged with a `livenessCode` rather than matched on its message text, so the wording stays free to change.

## Scope is deliberately narrow

- A block on the **main document** still poisons the verdict. If the posting's own host does not resolve, that is a real finding about the posting. Verified: a main document on a non-resolving host still returns `uncertain` / `blocked_host`.
- `rejectPrivateOrInvalid`, the literal-host layer, is untouched. Re-ran its cases: 12 of 12 SSRF bypass vectors still blocked, 3 of 3 public URLs still allowed.
- **Nothing about the request is relaxed.** The abort is unchanged in every case, so the guard's actual security behaviour is identical. Only the verdict stops being poisoned.

## Verification

| case | result |
|---|---|
| live iCIMS behind dead `visitorqueue` | `uncertain` → `active` |
| live iCIMS behind dead `addthis` | `uncertain` → `active` |
| genuinely dead posting (HTTP 410) | `expired`, unchanged |
| main document on a non-resolving host | `uncertain` / `blocked_host`, unchanged |
| loopback resolver (SSRF path) | still throws, still poisons |

Independent of #2884, which touches a different part of the same file; the two do not overlap textually.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of DNS resolution failures when no network addresses are returned.
  * Pages are no longer incorrectly marked as blocked when only a third-party resource fails to resolve.
  * Main-document failures continue to be reported as uncertain blocked-host results.
  * Failed requests are safely stopped to prevent further loading attempts.
  * Improved liveness accuracy for pages containing unavailable third-party resources, while preserving accurate results for unavailable main documents.


artemtrofymenko:
Follow-up with the post-fix measurement, which was still running when I opened this.

I re-ran the 120 URLs that were still `uncertain` after the iframe change in #2884 had already been applied. So the iframe fix had already had its turn on exactly these URLs and left them unresolved; the only new variable is the change in this PR.

```
Results: 79 active  0 expired  41 uncertain
```

**79 of 120 resolved**, and **zero** are still blocked by a third-party host. The prior pass counted 78 blocked purely by a dead analytics or ad host, so the two numbers line up almost exactly, which is what makes the attribution clean rather than a guess.

What is left is unrelated to this change:

| count | reason |
|---|---|
| 36 | `content present but no visible apply control found` — genuine remaining uncertainty |
| 5 | transient navigation errors |

Nothing regressed to `expired`, which is the direction that would have worried me: a false `expired` is written to scan-history as `skipped_expired` and dedup-filters the URL out of every later scan.

For scale across the whole queue, over 1555 URLs, 268 postings (17%) were being misjudged before these two changes: 92 live ones classified `expired`, 97 resolved by #2884, and these 79.

artemtrofymenko:
Good catch, taken in `34a9301`. That finding lands on the one claim in the PR description I had not actually tested: *"Nothing about the request is relaxed. The abort is unchanged in every case."* The double was throwing `abort()` away, so both assertions rested on the verdict alone.

Both now require `abortCount === 1` as well, so a change that quietly let a non-resolving host through would fail here instead of passing on the verdict.

One extra step before trusting it, because a counter that always reads 1 asserts nothing. With a resolver returning a real address, the same double records:

```
resolvable subresource -> result=active  aborts=0  continues=1
```

versus the two cases under test:

```
subresource -> result=active                 aborts=1
main doc    -> result=uncertain/blocked_host aborts=1
```

So the counter discriminates rather than being decorative.

santifer:
The measurement is what makes this mergeable rather than plausible, @artemtrofymenko:

> 217-URL recheck: 78 live postings returned as `uncertain` because an analytics or ad host no longer exists — 53 on personalisation.visitorqueue.com, 17 on s7.addthis.com (**AddThis was shut down in 2023**), the rest on fluidads and cloudfront. One opened by hand to confirm: 11,178 characters of live posting.

That's 36% of a real sample being told "we can't tell" about a page that was fine, because a tracker the employer forgot to remove years ago no longer resolves. And the user-visible cost is worse than the number: `uncertain` on a live posting is a job they might not apply to.

Tagging the error with `livenessCode: 'dns_no_addresses'` so the route guard can distinguish *this host doesn't exist* from *this host resolves into private space* is the right structural move. Those two look identical at the throw site and mean opposite things — one is a dead third-party script and says nothing about the page; the other is an egress-guard hit and says a great deal. Collapsing them is how the egress guard would have been quietly weakened by a fix for the first problem.

I approved the held CI on this and #2884 at the start of tonight's session, so both are green.

Not merged tonight: per-session merge ceiling, and these two are near the top next cycle. Genuinely good work — the kind where the investigation is worth more than the patch.

artemtrofymenko:
One correction before next cycle, so you do not lose the slot on it: **#2884 is green, this one is not.**

```
#2884  head ed0dbe7   Tests / CodeQL / Dependency Review / No user data  ->  all success
#2885  head 34a9301   the same four                                      ->  action_required
```

Checked by head SHA rather than branch name, and `Tests` is gated rather than skipped — the run was created and held, not filtered out.

Timeline suggests the approval simply predates this head: `34a9301` was pushed at 21:40:16Z and its runs were held at 21:40:24Z, about twenty minutes before your comment. The gate re-arms per commit, so anything pushed after an approval needs a fresh one.

What makes it easy to miss: held workflows never become checks, so `gh pr checks 2885` returns three green checks and looks fine. The absence only shows up against the workflow runs themselves, or next to #2884, which has thirteen.

Nothing needed from me — flagging it so the merge attempt does not run into it.

santifer:
You're right and I was wrong, @artemtrofymenko. Verified:

```
#2884 → test (macos) pass · test (ubuntu) pass · test (windows) pass
#2885 → no test checks at all
```

I said "I approved the held CI on this and #2884 at the start of tonight's session, so both are green". The approval was real, but you pushed after it — and **every push re-holds the CI under this repo's first-time-contributor policy**. So the approval I was remembering had expired, and I reported a state I hadn't re-checked.

That's a claim I'd have caught in someone else's comment, and it's the exact failure I've been fixing in my own tooling all night: reporting a green that was measured against different code. Thanks for correcting it rather than letting me merge on it.

**Approved again just now** — the runs on `34a93014` are released and should go green shortly. If you push again before it's merged, ping me: the gate will hold it a third time and there's nothing you can do from your side about that.

The substance of my earlier comment stands: the 217-URL measurement with 78 live postings returned as `uncertain` because of dead analytics hosts is the strongest evidence in either PR, and `livenessCode: 'dns_no_addresses'` keeping *host doesn't exist* separate from *host resolves into private space* is the right structural call.

artemtrofymenko:
Pinging as you asked — the head moved and the gate is holding again.

```
head f4e1ee8   Tests / CodeQL / Dependency Review / No user data  ->  action_required
```

`34a9301` → `f4e1ee8` was the fix for the three failing `test` jobs the previous release surfaced, so the approval from earlier no longer applies to this head.

No rush on it, and nothing else is pending from my side.

artemtrofymenko:
Correcting my own last comment, which was wrong in a way that would have cost you a look.

I reported the gate as holding on `f4e1ee8`. It had already been released, and Tests had already run and failed — **29 minutes before I wrote that**. So the message said "waiting for approval" while the PR was showing a red X, which is the least useful pair of signals to hand someone.

The red is not this branch:

```
test (macos-latest)      ✅
test (ubuntu-latest)     ✅
cv-visual                ✅
test (windows-latest)    ❌
Upgrade regression gate  ❌
```

Both failures are the same thing, and it is not a test:

```
##[error]Failed to download archive
  'https://codeload.github.com/actions/setup-go/…' after 3 attempts.
##[error]Response status code does not indicate success: 429 (Too Many Requests).
##[error]Response status code does not indicate success: 503 (Service Unavailable).
```

The jobs died at the setup step. **Zero test assertions failed in that run** — the suite never started, which is also why the two runners that got their download through are green on identical code.

A re-run of the failed jobs should clear it. I cannot trigger one: `gh run rerun --failed` returns `Must have admin rights to Repository` for a fork contributor, so it needs you or a push, and a push re-arms the first-time gate for nothing.

Nothing else is pending from my side, and no rush — flagging it only because a stale red that nobody can explain from the outside is easy to read as "still broken".

santifer:
The measurement is what makes this obvious in hindsight, @artemtrofymenko, and it was not obvious before: 78 live postings out of a 217-URL recheck came back `uncertain` because an analytics host on the page no longer exists, 53 of them on one vendor and 17 on AddThis, which shut down in 2023. Opening one by hand to find 11,178 characters of live posting and a working apply control is the part that turns a plausible theory into a number.

Two things I checked rather than assumed. The request is still aborted on every path, so the egress guard gives up nothing: only the verdict stops being poisoned. And the main-document case still counts, which is right, because a posting whose own document cannot resolve IS a finding about the posting.

The detail I would have got wrong is the last one: when the route double cannot tell you whether it was a subresource, you default to TRUE and keep the old poisoning behaviour, so the relaxation only applies where it can be positively established. Degrading to the stricter answer when the fact is unknowable is the same instinct as returning nothing instead of a partial answer. Merged 🚀

PR #2890, opened 2026-08-14
feat(cv): make Work Experience an optional section

### LINKED ISSUE #2504: feat(cv): make Work Experience an optional section
### Code of Conduct

### Existing issues

### Problem

 Problem

Right now the "Work Experience" section is one of the few CV sections that's hardcoded as always-present — cv-sections-core.mjs's OPTIONAL_SECTIONS list only covers projects, education, certifications, and awards. If a candidate has no professional experience yet (students, recent graduates, career-switchers), the generated CV/resume still renders a bare "Work Experience" header with nothing under it, because {{EXPERIENCE}} is inserted unconditionally rather than through a strippable marker block.

This is for students/new grads with zero professional history same motivation as Awards that just got recently added.

Happy to work on this, I've already been through this exact code path.

### Proposed solution

Follow the same pattern as #2414 (Awards):

- cv-sections-core.mjs: add experience to OPTIONAL_SECTIONS; add an experience entry to PATTERNS.html and PATTERNS.tex.
- templates/cv-template.html / templates/resume-template.html / templates/cv-template.tex: wrap the Work Experience block in  (and the LaTeX %%%% Experience %%%% banner), replacing the current unconditional {{EXPERIENCE}} placement. (buildExperience() already exists in both build-cv-html.mjs and build-cv-latex.mjs and already returns '' for an empty array — this PR is actually simpler than Awards, which needed a new builder from scratch.)
- tests/cv-optional-sections.test.mjs: flip the existing {{EXPERIENCE}} is untouched assertion and add Experience to the same empty/populated/mixed coverage the other three sections get.
- modes/pdf.md / modes/latex.md: document experience[] as optional in the payload schema table (mirroring how awards[] got documented — "omit the key or pass [] and the section is dropped, header included"), and note it in the "Optional sections" line of the ATS Rules section.
- templates/README.md: add Experience to the "Optional sections" note alongside Projects/Education/Certifications/Awards.

### Area

PDF Generation

husamemadH:
## What does this PR do?

Makes Work Experience an optional CV section: a payload with no `experience` entries now drops the whole block, section header included, instead of rendering a bare "Work Experience" title with nothing under it. A payload that has experience renders byte-identically to before.

The markers already exist in every shipped template and both builders' `buildExperience()` already returned `''` for an empty array, so this is two patterns and one `OPTIONAL_SECTIONS` entry — no template changes. Experience sits mid-document in every shipped template and each carries the `<!-- END -->` / `%%%%  END  %%%%` sentinel, so it can never run to end-of-input and uses the shared boundary rather than Skills' marker-only one.

### Marker coverage, and what happens to a template added later

Confirming the two things asked for above. It is now **nine** shipped CV templates, not four — #2954 added five named variants after this PR was opened — and all nine carry the marker (`<!-- WORK EXPERIENCE -->`, or `%%%%  Experience  %%%%` in LaTeX): `cv-template.html`, `resume-template.html`, `cv-template.zh-minimal.html`, `cv-template.{compact,executive,jake,leadership,modern}.html`, `cv-template.tex`.

All nine are now in the `TEMPLATES` matrix in `cv-optional-sections.test.mjs`. Three were covered before; `cv-template.zh-minimal.html` shipped a full marker set and had never been in it, which is the gap raised in review.

A tenth template added later **fails loudly rather than rendering the old way** — but the loud failure is in CI, not at runtime, and the distinction matters. At runtime a template with no marker is a deliberate no-op (`stripEmptySections` only ever removes; the fail-safe reasoning is in the module header and unchanged by this PR). The enforcement is a coverage guard in the test suite: it enumerates CV templates on disk by the `{{EXPERIENCE}}` placeholder `cv-templates.mjs` requires of every one, and fails by filename when one is not in the matrix — in both directions, so a matrix entry with no file behind it fails too. I verified both cases by adding a throwaway template with markers and one without; each turns the suite red naming the file.

### Verification

I broke it on purpose before trusting it — renamed the HTML marker (3 assertions red), renamed the LaTeX banner (2 red), swapped the boundary for one naming `<!-- EDUCATION -->` (the over-greedy case eats Projects in all three HTML templates, red) — restoring after each. Byte-identical-when-present is checked end-to-end as well as in the unit assertion: I built HTML and `.tex` from a populated payload with this module and with the current `main` module, and `diff` is clean for both formats. A new-graduate payload compiles under `pdflatex` to a real 1-page PDF. `node test-all.mjs`: 4762 passed, 0 failed.

One knock-on: `cv-named-templates.test.mjs` used `{{EXPERIENCE}}` surviving an all-empty strip as its canary for an over-reaching boundary. Experience is strippable now, so its payloads populate experience explicitly — the canary keeps working instead of passing by accident.

## Related issue

Closes #2504

## Type of change

- [ ] Bug fix
- [x] New feature

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Work Experience is now treated as an optional CV section.
  * Empty or omitted Work Experience sections are automatically removed, including their headers and placeholders.
  * Supported consistently across HTML, LaTeX, and PDF outputs.

* **Documentation**
  * Updated schemas and template guidance to explain when Work Experience may be omitted.

* **Bug Fixes**
  * Improved handling of empty optional sections while preserving surrounding content and document structure.

### DIFF
--- a/cv-sections-core.mjs
+++ b/cv-sections-core.mjs
@@ -1,18 +1,27 @@
 // Shared optional-section stripping for the CV builders (build-cv-html.mjs,
 // build-cv-latex.mjs).
 //
-// Core competencies, projects, education, certifications, awards, and skills
-// are the genuinely optional CV sections: a competency tag row is often
-// redundant with the summary and experience bullets that prove the same
-// claims, a candidate's projects are often already covered under Work
+// Core competencies, work experience, projects, education, certifications,
+// awards, and skills are the genuinely optional CV sections: a competency tag
+// row is often redundant with the summary and experience bullets that prove
+// the same claims, a student or career-switcher has no professional history
+// to list yet, a candidate's projects are often already covered under Work
 // Experience, not every candidate has a degree, not every application carries
 // a certification worth listing, most candidates have no award to name, and
 // plenty of candidates list no skills section at all. The templates wrap all
-// six unconditionally, so a payload with no entries renders a bare section
+// seven unconditionally, so a payload with no entries renders a bare section
 // header with nothing under it. The builders' buildCompetencies()/
-// buildProjects()/buildEducation()/buildCertifications()/buildAwards()/
-// buildSkills() correctly return '' — nothing removes the surrounding
-// wrapper, which is what this module does.
+// buildExperience()/buildProjects()/buildEducation()/buildCertifications()/
+// buildAwards()/buildSkills() correctly return '' — nothing removes the
+// surrounding wrapper, which is what this module does.
+//
+// Work experience (#2504) is optional for the people this tool is aimed at —
+// new graduates, career changers, and anyone leading with projects or
+// education — not because the section is unimportant. Note that the payload
+// key being optional says nothing about the template placeholder:
+// cv-templates.mjs still requires `{{EXPERIENCE}}` to be present in any custom
+// template, exactly as before. What changed is that an empty `experience` no
+// longer leaves the wrapper behind.
 //
 // Certifications has no marker in the LaTeX template (cv-template.tex has no
 // Certifications section at all), so PATTERNS.tex has no `certifications` key
@@ -100,6 +109,7 @@ const TEX_END_SENTINEL = String.raw`(?=^%{4,}\s)`;
 const PATTERNS = {
   html: {
     competencies: new RegExp(String.raw`<!--\s+CORE COMPETENCIES\s+-->[\s\S]*?` + HTML_BOUNDARY),
+    experience: new RegExp(String.raw`<!--\s+WORK EXPERIENCE\s+-->[\s\S]*?` + HTML_BOUNDARY),
     projects: new RegExp(String.raw`<!--\s+PROJECTS\s+-->[\s\S]*?` + HTML_BOUNDARY),
     education: new RegExp(String.raw`<!--\s+EDUCATION\s+-->[\s\S]*?` + HTML_BOUNDARY),
     certifications: new RegExp(String.raw`<!--\s+CERTIFICATIONS\s+-->[\s\S]*?` + HTML_BOUNDARY),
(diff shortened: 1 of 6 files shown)


santifer:
The sentence that makes this reviewable, @husamemadH:

> A payload that has experience renders **byte-identically** to before.

That's the right guarantee for an "optional section" change and the one I'd have asked for. The risk in this class isn't the empty case — it's that everyone else's CV shifts by a newline and nobody notices until a PDF reflows. Stating it as byte-identity rather than "looks the same" is a claim that can be tested instead of eyeballed.

And dropping the header along with the block is the correct reading of the requirement. A bare "Work Experience" title with nothing under it is worse than omitting the section: on a CV it reads as *this person has no work experience* rather than *this section wasn't filled in*, and for a career-changer or new graduate that's actively harmful.

Not merged tonight (per-session merge ceiling), queued. One thing worth confirming in the PR body: that the markers exist in **all four** shipped templates and that a fifth template added later without them fails loudly rather than silently rendering the old way.

Scott-Emberson (approved):
Code owner for `tests/`. The false-pass check is cleared: `cv-optional-sections.test.mjs` asserts the heading and `{{EXPERIENCE}}` placeholder are absent when experience is empty, and pins the populated case with byte-identity, so it fails if the section were not actually made optional. One low coverage gap: `templates/cv-template.zh-minimal.html` ships the full marker set but is not in the `TEMPLATES` matrix, so the strip is never verified there. Adding it (and a guard that fails loudly when a shipped template is uncovered) would match the "all four templates" intent. Approving.

husamemadH:
@Scott-Emberson thanks — the gap was real, and wider than it looked when you flagged it. `#2954` landed five named templates (`compact`, `executive`, `jake`, `leadership`, `modern`) the day after this PR opened, so it was nine shipped CV templates with three covered, not four with three.

All nine are in the `TEMPLATES` matrix now — 182 assertions to 574.

On the guard: I kept the matrix hand-written rather than deriving it from disk, because `hasCertifications: false` is a claim about `resume-template.html`, not an observation of it. Deriving that flag would let a template that *lost* its `<!-- CERTIFICATIONS -->` marker be quietly reclassified as "doesn't have certifications" and still pass. So the list is declared, and membership is checked against disk in both directions: a CV template on disk missing from the matrix fails by filename, and a matrix entry with no file behind it fails too. Templates are discovered by the `{{EXPERIENCE}}` placeholder `cv-templates.mjs` requires of every one, which is why `cover-letter-template.html` isn't swept up.

I checked the guard can actually fail rather than assuming it: dropped in a throwaway `cv-template.sneaky.html` with the full marker set and a `cv-template.nomarkers.html` without, and each turns the suite red naming the file.

One knock-on you may want to look at since it's in your area — `cv-named-templates.test.mjs` used `{{EXPERIENCE}}` surviving an all-empty strip as its canary for an over-reaching boundary, and its payloads carry no `experience` key. That assertion was correct while experience was non-optional; after this change it went red across all five named templates. I fixed it by populating `experience` explicitly in those payloads, so the canary still means something instead of passing by accident.

santifer:
Clean marker work, @husamemadH: empty payload drops the whole block header included, populated payload renders byte-identically, and the golden visuals prove the second half. First CV without a Work Experience section that doesn't ship an empty title — the career-changer case #2504 asked for. Merged 🚀

PR #2898, opened 2026-08-14
fix(velocity): count a web-written transition as an observation

### LINKED ISSUE #2897: funnel-velocity.mjs: a web-written status-log row is parsed, counted, then silently excluded from every velocity figure
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### Existing issues

- [x] I have searched existing issues and this is not a duplicate

### What happened?

`funnel-velocity.mjs` gates status-log rows on two allow-lists:

```js
const VALID_SOURCES = new Set(['set-status', 'correction', 'backfill', 'manual']);
// Sources whose dates are trusted for day-math. backfill/manual are parsed and
// counted but excluded: they are reconstructed after the fact, not observed.
const DAY_MATH_SOURCES = new Set(['set-status', 'correction']);
```

`web` is in neither. A row written from the web app is therefore **parsed, counted as an observation, excluded from the velocity maths, and reported under `dataQuality.unknownSources`**:

```js
if (!VALID_SOURCES.has(source)) {
  unknownSources.push({ line: i + 1, num, source });
  observations.push({ …, dayMath: false });   // counted, but never fed to day-math
  continue;
}
```

The rows are not discarded — which is what makes this quiet. An installation driven from the web app gets a ledger whose entries are all flagged as questionable and contribute nothing to "days between states". The funnel looks empty rather than broken.

**The exclusion is also wrong on the merits.** The comment above `DAY_MATH_SOURCES` gives the reason `backfill`/`manual` are excluded: *"they are reconstructed after the fact, not observed."* A web status change is observed live, at the moment the user makes it. It is the same event as a `set-status` transition entering through a different door, and it belongs in both sets.

`company-history.mjs` is affected too, but has no allow-list of its own: it loads the ledger by dynamic import of `funnel-velocity.mjs` and consumes that reader's output, so it inherits the exclusion rather than repeating it. Fixing the two sets here fixes both readers.

### Steps to reproduce

Write a ledger row with source `web` — the format is otherwise byte-identical to what `set-status.mjs` writes — and read it back:

```bash
printf '1\t2026-06-01\tEvaluated\tApplied\tweb\t\n1\t2026-06-08\tApplied\tResponded\tweb\t\n' > data/status-log.tsv
node funnel-velocity.mjs --json
```

Both rows appear under `dataQuality.unknownSources`, and `appliedToResponded` reports no observation despite a clean 7-day transition being present in the file.

Substituting `set-status` for `web` in the same two lines produces the expected 7-day figure, which isolates the source column as the only difference.

### Expected behavior

`web` is accepted as a first-hand source: present in `VALID_SOURCES`, absent from `unknownSources`, and included in `DAY_MATH_SOURCES` so its dates count toward the hop figures — the same treatment `set-status` gets, because it is the same class of event.

### CLI tool

Claude Code

### OS and Node.js version

macOS 26.5.2, Node.js v22.22.3

### Output of npm run doctor

Not setup-related — the defect is in the reader's source allow-list and reproduces from the two-line fixture above on any install.

rubicon:
## What does this PR do?

Adds `web` to `VALID_SOURCES` and `DAY_MATH_SOURCES` in `funnel-velocity.mjs`, so a status-log row written from the web app is treated as a first-hand observation instead of an unrecognized one. Two lines and three self-test assertions.

## Related issue

Closes #2897

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

---

## The defect

The two allow-lists gate every status-log row, and `web` is in neither:

```js
const VALID_SOURCES = new Set(['set-status', 'correction', 'backfill', 'manual']);
const DAY_MATH_SOURCES = new Set(['set-status', 'correction']);
```

The rows are **not** discarded, which is what makes this quiet:

```js
if (!VALID_SOURCES.has(source)) {
  unknownSources.push({ line: i + 1, num, source });
  observations.push({ …, dayMath: false });   // counted, but never fed to day-math
  continue;
}
```

A web-written transition is parsed, counted as an observation, flagged under `dataQuality.unknownSources`, and contributes nothing to "days between states". An install driven from the web app reads as an *empty* funnel rather than a broken one — the failure mode least likely to get reported.

## Why `web` belongs in both sets, not just the first

The comment above `DAY_MATH_SOURCES` states the criterion for exclusion: *"they are reconstructed after the fact, not observed."* That is true of `backfill` and `manual`. It is not true of `web` — a web status change is recorded at the moment the user makes it. It is the same event as a `set-status` transition entering through a different door. Admitting it to `VALID_SOURCES` while withholding it from `DAY_MATH_SOURCES` would clear the `unknownSources` flag and still leave every figure empty.

## `company-history.mjs` needs no change

It reads the same ledger but has no allow-list of its own: it loads it by dynamic import of this module and consumes this reader's output, so it inherits the fix rather than repeating it. Worth stating explicitly, since a second copy of the list is the obvious thing to go looking for.

## Tests

Three self-test assertions, each written and seen failing first:

```
SELF-TEST FAIL: parser: web is a recognized source (got 2 unknown)
SELF-TEST FAIL: parser: web observations are trusted for day-math
SELF-TEST FAIL: velocity: a web transition reaches the hop figures (expected n=1, got 0)
```

The third is the one that matters. Asserting the parse alone would not have caught this, because the rows always parsed — what failed was that a parsed row reached no figure. It folds the observations and asserts the hop registers, so writer and reader are checked against each other rather than each against itself.

The fixture is separate from `LOG_FIXTURE` so the counts that fixture pins stay pinned.

`node test-all.mjs` → **3877 passed, 0 failed**.

## Scope

Reader half only, deliberately. No writer changes, so it stands on its own regardless of what happens to any writer PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Web-sourced status changes are now recognized correctly.
  * Web observations contribute accurately to velocity calculations.
  * Prevented unnecessary unknown-source warnings for web data.

* **Tests**
  * Added validation for parsing web-sourced transitions and calculating velocity.

### DIFF
--- a/funnel-velocity.mjs
+++ b/funnel-velocity.mjs
@@ -51,10 +51,13 @@ const summaryMode = args.includes('--summary');
 const selfTestMode = args.includes('--self-test');
 
 const DATE_RE = /^\d{4}-\d{2}-\d{2}$/;
-const VALID_SOURCES = new Set(['set-status', 'correction', 'backfill', 'manual']);
+// `web` sits alongside `set-status` because it is the same class of event: a
+// status change made in the web app is recorded as the user makes it, not
+// reconstructed afterwards. Only the door differs.
+const VALID_SOURCES = new Set(['set-status', 'web', 'correction', 'backfill', 'manual']);
 // Sources whose dates are trusted for day-math. backfill/manual are parsed and
 // counted but excluded: they are reconstructed after the fact, not observed.
-const DAY_MATH_SOURCES = new Set(['set-status', 'correction']);
+const DAY_MATH_SOURCES = new Set(['set-status', 'web', 'correction']);
 
 // The hops a candidate can measure from their own tracker. Applied→Rejected is
 // tracked separately from the forward hops — a "days to terminal" number that
@@ -485,6 +488,29 @@ function selfTest() {
   check(unknownSources.length === 1 && unknownSources[0].source === 'future-import', 'parser: unknown source counted');
   check(observations.find(o => o.source === 'future-import')?.dayMath === false, 'parser: unknown source excluded from day-math');
   check(observations.find(o => o.num === 2 && o.from === null), 'parser: "-" from-state parses as null');
+
+  // A status change made from the web app is observed live, at the moment the
+  // user makes it — the same event as a set-status transition entering through
+  // a different door. Excluding it does not drop the rows, which is what makes
+  // the failure quiet: they are parsed, counted, flagged under unknownSources,
+  // and contribute nothing to the hop figures, so a web-driven install reads as
+  // an empty funnel rather than a broken one. Its own fixture so the counts
+  // pinned above stay pinned.
+  const WEB_FIXTURE = [
+    '1\t2026-06-01\tEvaluated\tApplied\tweb\t',
+    '1\t2026-06-08\tApplied\tResponded\tweb\t',
+  ].join('\n');
+  const web = parseStatusLog(WEB_FIXTURE, states);
+  check(web.unknownSources.length === 0, `parser: web is a recognized source (got ${web.unknownSources.length} unknown)`);
+  check(web.observations.length === 2 && web.observations.every(o => o.dayMath === true),
+    'parser: web observations are trusted for day-math');
+
+  // The rows above are a clean 7-day Applied→Responded hop. Asserting the
+  // parse alone would not have caught this: the defect was never that the rows
+  // failed to parse, it was that a parsed row reached no figure.
+  const webVelocity = computeVelocity(foldObservations(web.observations), TODAY);
+  check(webVelocity.appliedToResponded.n === 1,
+    `velocity: a web transition reaches the hop figures (expected n=1, got ${webVelocity.appliedToResponded.n})`);
 
   // -- fold --
   const timelines = foldObservations(observations);


santifer:
The two-line change is right, @rubicon — a web-written status row being parsed but not counted is the mirror of #2900: one write is lost, the other lands and isn't seen. Adding `web` to both `VALID_SOURCES` and `DAY_MATH_SOURCES` is the correct fix, and doing both matters: recognised-but-excluded-from-day-math would still silently distort velocity.

**But the PR carries `node_modules`.** GitHub lists it in the changed files:

```
funnel-velocity.mjs, node_modules
```

That's almost certainly an accidental `git add`, and it needs removing before this can be reviewed — a dependency tree in a diff makes the two real lines unfindable, and merging it would commit vendored packages to the repo. `git rm -r --cached node_modules`, amend, force-with-lease.

Once that's off, this is a two-line read and I expect it to go straight in. Not a criticism — it's the most common accident there is, and it's caught before it cost anything.

(Separately: your #2900 is labelled `bug`+`p1`, and the web maintainer has #2903 up implementing the lock, with the lost update reproduced as a failing test first. Good report.)

rubicon:
Fixed — force-pushed `c99be61`. The diff is now one file, `funnel-velocity.mjs`, +28/-2.

You were right that it had to go, and one detail is worth correcting for the record rather than letting it stand: it was a **symlink**, not a vendored tree — mode `120000`, a single blob pointing at `../../node_modules`. So nothing would have landed in the repo except a dangling link, which is arguably worse than useless to anyone who checked it out, since it resolves to nothing on their machine and silently to something on mine.

Cause, since it says something about the setup rather than just about me: I work these in a `git worktree`, and a worktree has no `node_modules` of its own. I symlinked the parent's so the suite could run, then swept it up with `git add -A`. The lesson isn't "be careful with `git add`" — it's that the scaffolding needed to run the tests lived inside the tree being committed. It now gets removed before the commit rather than after the review.

Re-verified after amending: `funnel-velocity.mjs --self-test` OK, `node test-all.mjs` **3877 passed, 0 failed**.

On the substance — agreed on the framing, and the third assertion in this PR is the one carrying it. Asserting the rows *parse* would have passed all along; they always parsed. What failed was that a parsed row reached no figure, so the test folds the observations and asserts the hop actually registers. That is the same gap in miniature as the one between #2900 and this: a write landing and a write being seen are different properties, and only the second one was untested.

Thanks for catching it before it cost a review pass.

santifer:
Confirmed clean, @rubicon:

```
[funnel-velocity.mjs]  +28/-2   MERGEABLE
```

One file, the `node_modules` gone. Thanks for turning it around fast.

Approved on my side — the change is right and I said why above: a web-written status row being parsed but not counted is the mirror of the lost write in #2900, and adding `web` to both `VALID_SOURCES` **and** `DAY_MATH_SOURCES` is what stops it being recognised-but-still-excluded.

Not merging it tonight, and the reason is a rule I hold myself to rather than anything about your PR: I hit the per-session merge ceiling hours ago, and the point of that ceiling is that what's left waits for the next cycle *even when it's green*. Merging "just one more" because it's small is exactly how a ceiling stops existing.

It's near the top next cycle. And your #2900 report was the most valuable thing anyone filed today — a silent lost update on the user's source of truth, with the reproduction attached.

rubicon:
Rebased onto current `main` — one commit, `funnel-velocity.mjs` only, `MERGEABLE`, every check green. The merge commit that had accumulated on the branch is gone, so the diff is the two lines and the three assertions and nothing else.

Nothing has changed in the substance since your approval. Flagging one ordering point that came out of adjudicating this cluster, because it makes this the piece that should go first rather than the small one that can wait:

**Every other PR here depends on this one, and none of them can be safely taken before it.** #2901 makes the web write `source=web` rows. Until `VALID_SOURCES` knows the label, those rows are parsed, counted, flagged under `dataQuality.unknownSources` and contribute nothing to day-math — an install driven from the web reads as an empty funnel rather than a broken one. Landing a writer first does not produce an error anyone would report; it produces figures that are quietly wrong, in an append-only file.

This is also inert on its own — no user's ledger can contain a `web` row until some web writer exists, so taking it early costs nothing and removes the sequencing hazard entirely.

No action needed from me. Just: when the ceiling lifts, this is the one that unblocks the rest.

santifer:
Rebasing AND revising because your own #2898 falsified the queued version is exactly the standard this repo runs on: the fix now counts web-written transitions as observations instead of enumerating writers. Merged! 🚀

Issue #2900, opened 2026-08-14
POST /api/status is the only tracker writer that takes no lock — a concurrent merge-tracker or set-status run silently discards one of the two writes


rubicon:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### Existing issues

- [x] I have searched existing issues and this is not a duplicate

### What happened?

`web/src/app/api/status/route.ts` performs a read-modify-write on `data/applications.md` with no lock:

```js
md = fs.readFileSync(file, "utf8");     // read
// … find the row, rewrite its Status cell in memory …
atomicWrite(file, lines.join("\n"));    // write
```

`atomicWrite` makes the *write* atomic — the file is never observed half-written. It does nothing about the **read-modify-write around it**. Between the `readFileSync` and the `atomicWrite`, another writer can commit a change to the same file, and this route will then overwrite it with a copy of the tracker that predates it. The other writer's change is gone, with no error on either side.

Every other tracker writer in the tree takes the shared lock in `tracker-utils.mjs` — `merge-tracker.mjs`, `set-status.mjs`, `mark-pdf-ready.mjs`, `reserve-report-num.mjs`. This route is the one that does not, so it is the only writer that can lose someone else's work.

Concretely, marking a row Applied in the web app while a batch `merge-tracker.mjs` run is finishing can drop the merged rows; the reverse ordering drops the status change. Neither is reported.

Two smaller consequences of the same design:

- The route reimplements Status-column detection (the `statusIdx` header scan) that `tracker-parse.mjs` already does, so the layout rules now live in two places.
- It writes only the Status cell. `set-status.mjs` additionally appends the transition ledger `data/status-log.tsv` (since #1695), so a status change made from the web app leaves no ledger row at all — which is the gap #2714 / #2715 set out to close by adding a second ledger writer *beside* this unlocked write, rather than by removing it.

### Steps to reproduce

The window is real but short, so the clearest demonstration is to hold the lock the route ignores and watch it write anyway:

1. Start the web app against a career-ops root with a tracker containing row `1` in `Evaluated`.
2. Take the shared tracker lock the way `merge-tracker.mjs` does — e.g. `node -e "import('./tracker-utils.mjs').then(async m => { await m.acquireTrackerLock(m.trackerLockDirFor('data/applications.md'), {}); setTimeout(()=>{}, 60000) })"`.
3. With the lock held, `POST /api/status` with `{"n":"1","status":"Applied"}`.

[...]

PR #2901, opened 2026-08-14
refactor(web): delegate /api/status to set-status.mjs (converge on one writer)

rubicon:
## What does this PR do?

`POST /api/status` stops writing `data/applications.md` and delegates to `set-status.mjs --source web`, the writer every other tracker mutation already goes through.

**This is no longer the #2900 bug fix.** #2903 closed that hole, and it is closed. What is left is the architectural half #2904 exists to hold: the tracker currently has two writers that must agree on locking forever, and this reduces it to one.

## Related issue

Refs #2904 — deliberately not `Closes`. The convergence question is bigger than this diff and @santifer asked for the issue to stay open while it is designed.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor (no behavior change to the happy path; error mapping changes, see below)

## Checklist

- [x] If this is a new feature or architecture change, I opened an issue first (#2904 holds the architecture question)

---

## What delegation buys, concretely

Three things stop being this route's problem rather than being solved again in it:

- **The lock.** `set-status.mjs` holds it across the whole read-modify-write. The route holds nothing, so it cannot leak anything.
- **The Status-column scan.** The route's own `statusIdx` header walk (8- vs 9-column detection) is deleted. The layout rules live in `tracker-parse.mjs` and nowhere else.
- **The ledger row.** It appears because `set-status.mjs` has appended `status-log.tsv` since #1695 — a consequence of the write rather than a second writer beside it. That is what makes #2715 unnecessary rather than parallel, and I am recommending #2715 be closed on that basis.

## `--source`

`set-status.mjs` gains `--source` so a delegated write stays attributable in the ledger. The allow-list is narrow (`set-status`, `web`) because the value lands in a file `funnel-velocity.mjs` parses positionally and gates on its own allow-list — an unrecognized label would be persisted here and silently dropped there. Rejected at the boundary, non-zero, nothing written.

Default unchanged: a plain CLI run writes the same bytes as before, asserted by an exact-string test.

## Dependency: #2898

Ledger rows written from the web land as source `web`, which `funnel-velocity.mjs` only recognizes after #2898. Without it they are parsed, counted, and excluded from every velocity figure. **#2898 must land at or before this.** It is approved and green.

## The HTTP contract

Exit 1 is the CLI's catch-all, and only part of it is the caller's fault. `usage` and `invalid-state` map to 400; `states-error`, `write-failure` and `lock-error`, and a crash that printed no JSON, map to 500 — blaming the caller for a broken checkout sends someone to re-check a request that was fine. It is an allow-list, not a deny-list, so an exit-1 code this route has never seen reads as 500.

`2 → 404`, `3 → 409`, `4 → 503` with `Retry-After`. A kill by our own timeout maps to 504 and says the write may or may not have landed, because at that point it genuinely is unknown.

The child's lock wait is bounded below the route's own kill timeout — otherwise the 60s default outlives the 30s kill and contention surfaces as a crash instead of the 503 it should be. An operator's own value is honored up to that ceiling.

**One deliberate change from #2903's contract:** lock contention answered 409 there and answers 503 + `Retry-After` here, because 409 is now carrying `AMBIGUOUS` (exit 3). No caller is affected — `status-select.tsx`, `apply-view.tsx`, `decision-card.tsx` and `assistant-console.tsx` all branch on `res.ok` or ignore the response, none reads a status code. Flagging it because it is a contract change, not because it is a risk.

## Data-only roots

The web can run against a `CAREER_OPS_ROOT` holding data and no scripts. Before `c7dc317` that produced a 500 carrying a raw Node module-resolution stack trace with absolute server paths; the spawn-failure branch echoed child stderr as `detail`, so any spawn failure leaked one. Now: `existsSync` feature-detection → `503 {"code":"core-script-missing"}`, and child stderr goes to the server log with a fixed string to the client. Same pattern `/api/followups` already uses.

## Cost

Measured on a 624-row / 148 KB tracker, alternating states so an idempotent re-run could not flatter it: delegated p50 32.9 ms, in-process p50 0.5 ms, bare `node -e 0` p50 17.7 ms — 54% of the delegated call is interpreter boot. Full HTTP round-trip p50 44 ms in dev. Numbers and caveats in #2904.

## Tests

`node test-all.mjs` → **3938 passed, 0 failed** (1 known warning), identical to `main`'s baseline. `npm --prefix web test` → 249 passed. `npx tsc --noEmit` clean.

New in `set-status-tests.mjs`, each seen failing first: `--source web` writes `web` to the ledger; omitting it still writes `set-status` (exact-string); and four unknown values are each rejected non-zero with nothing written.

## Response shape

`{ok, status}` unchanged; `changed` and `statusLogged` are additive, both read from `set-status.mjs`'s existing `--json` payload. No caller in `web/src` reads a field that has gone away.

## Two things a reviewer should not have to find

- **`web/src/lib/core/tracker-lock.ts` has zero callers after this.** `/api/status` was its only one. I have not deleted it — it is @santifer's, it is one merge old, and #2904 may yet rule it back into the main path. `run-registry.ts:24` also still advises "Writing the tracker → `withTrackerLock`", which would point the next contributor at the path this PR removes. Both are his calls, not mine.
- **Exit 0 with unparseable stdout answers 500 although the write landed.** `if (!parsed)` treats a successful child with no readable JSON as a server error. It is narrow — exit 0 means `set-status.mjs` committed — but it is the one remaining case of CodeRabbit's "reports failure after a write has already succeeded", and the honest answer there is the 504's wording, not a flat 500. Happy to fix it in this PR or leave it for a follow-up; say which.

## Summary by CodeRabbit

* **New Features**
  * Added source attribution for status changes, including web-originated updates.
  * Added validation for status sources and tracker row selections.
  * Status updates now report whether a change occurred and whether it was logged.
  * Added retry guidance for temporary update failures.

* **Bug Fixes**
  * Improved handling of command output, timeouts, missing tools, and failed status updates.
  * Prevented invalid status requests from creating ledger entries.

* **Tests**
  * Added regression coverage for source attribution, validation, JSON parsing, and tracker row handling.


santifer:
Verified the underlying bug on #2900 (`set-status.mjs:323` takes the lock, the web route doesn't) and labelled it `bug`+`p1`. Thanks for splitting issue from fix — on a data-integrity problem that's what lets the *problem* be discussed separately from one solution to it.

This PR touches `set-status.mjs` **and** `web/src/app/api/status/route.ts`, so it isn't mine to merge alone: the web subsystem has its own maintainer and I don't land changes in their routes unilaterally. I've sent them the verification and asked how they want to split it. Not a hold — a handoff, and they're active daily.

One thing worth having an answer ready for, since it's the difference between fixing this and moving the failure: **the core lock is file-based and the web is a long-lived process.** If the route acquires it, every error path has to release it, or a 500 leaves the user's tracker locked against their own CLI. That turns "I silently lost an update" into "I can't write at all" — louder, still bad. If your implementation already handles that, say so and it'll speed up their read.

santifer:
**I merged the other approach, and I want to be straight about it: on the architectural question, yours is the better one.**

#2903 landed an hour ago. It takes the core's file lock from the route and keeps the write in `web/`. Yours does something different — the route **delegates to `set-status.mjs --source web`** — and that is more faithful to the rule this app is built on: *the web orchestrates the real engine, it never reimplements it.* The pattern is already the house style (`/api/tracker/delete` shells out to `tracker.mjs`, `/api/run` to `merge-tracker.mjs`); mine is the exception, not yours.

Two things yours gets that mine doesn't:

1. **One writer.** With #2903 the tracker has two writers that must agree on locking forever. With yours it has one, and the web can't drift from it because it isn't writing.
2. **`--source web` feeds the transition ledger for free.** That's a separate open PR (#2715) whose whole purpose is recording web status changes. Your approach makes it unnecessary rather than parallel — and it also settles the question I asked the core maintainer about attribution, since the ledger keeps CLI and web rows distinguishable.

**Why I merged #2903 anyway, stated plainly:** the bug is a live silent data loss, I had a reproduction and a fix in hand, and I moved before studying your diff properly. That was my call and I'm not dressing it up as a plan — I optimised for closing the hole tonight and I under-weighted the design question. The hole is closed; the architecture question is still open, and it's yours.

**What I'd like, if you're up for it.** Two real costs to weigh before we migrate, and you're better placed than me to judge the first:

- **A process spawn per status change.** `/api/status` is the fastest, most-clicked write in the UI — a dropdown. `execFile` + Node startup + the script's own work is tens to hundreds of ms against a few ms today. Your `SET_STATUS_TIMEOUT_MS`/`boundedLockWait` suggest you thought about this; what did you measure?
- **`set-status.mjs` must exist.** The web can run against a data-only `CAREER_OPS_ROOT` (there's an explicit 400 for that elsewhere). Delegation needs a clean answer when the script isn't there, and "status updates silently stop working" isn't it.

If those hold up, I'll take your PR as the direction and #2903's lock helper becomes the fallback rather than the main path. Don't rebase yet — `route.ts` just changed under you and I'd rather give you the exact delta than have you aim at a moving target.

Either way: **you found this** (#2900), you diagnosed it correctly, and you proposed the more principled fix. The credit is in #2903's description and the merge, and it belongs here too.

santifer:
**Correcting my own comment above: I said your approach was better full stop, and I left out a cost that I had documented myself hours earlier.**

Opened **#2904** to hold the architectural question properly, with your approach as the proposal and the credit on it.

The thing I missed: **the core's tracker lock is not reentrant.** Delegating moves the lock into the child process, which is fine in isolation — but any route that already holds the lock and then orchestrates `set-status.mjs` self-deadlocks until the timeout. It can't happen today. It can the day `/api/run` needs to set a status mid-run, and by then the delegation would be the established path.

That doesn't sink your approach — it means the migration has to **rule on reentrancy first rather than discover it after**, which is now written down in #2904 as a precondition instead of a surprise.

So the honest scoreboard: yours is more principled (one writer, no possible divergence in *how* a row gets written, and `--source web` feeds the ledger for free — making #2715 unnecessary rather than parallel). Mine is cheaper per call and has no reentrancy question. Neither is free, and I presented the trade-off as one-sided.

Still open and genuinely needing your input, since you've measured more of this than I have: **what does the spawn actually cost** on the most-clicked write in the UI, and **what should happen when `set-status.mjs` isn't there** (the web can run against a data-only root).

Answer those in #2904 and the direction is settled on evidence rather than on which of us moved first tonight.

rubicon:
**Rebased onto post-#2903 `main`, and reframed: this is no longer a bug fix.**

You asked me not to rebase until you could hand me the exact delta. I went ahead, and the reason is that the target stopped moving — `route.ts` has not changed since d9c4fd0, and the delta was readable from `main`. If you would rather have reviewed the old diff against the old base, say so and I will reconstruct it.

## What the rebase actually was

Not mechanical. Both sides rewrote the same handler, so the conflict was total; I took the delegating version whole. The three changed files are byte-identical to what you had already seen — `git diff` against the pre-rebase head is empty. Nothing was re-authored under cover of a rebase.

Title and body now say refactor. **#2900 is closed and #2903 closed it**, so `Closes #2900` was claiming credit for a hole that is already shut. It refs #2904 rather than closing it, since you want that issue to outlive whichever PR lands.

## Three things the rebase exposed that you should decide, not inherit

**1. `web/src/lib/core/tracker-lock.ts` has zero callers after this.** `/api/status` was its only one. I did not delete it — it is yours, it is one merge old, and #2904 may put it back on the main path. But `run-registry.ts:24` still tells the next contributor "Writing the tracker → `withTrackerLock`", which is the path this PR removes. Dead code plus live advice pointing at it is the worse half.

**2. Lock contention changes from 409 to 503 + `Retry-After`**, because 409 now carries `AMBIGUOUS` (exit 3). I checked every caller rather than assuming: `status-select.tsx`, `apply-view.tsx`, `decision-card.tsx` and `assistant-console.tsx` all branch on `res.ok` or ignore the body — none reads a status code. Safe, but it is a contract change and you should have been told, not discovered it.

**3. `web typecheck + build` is red on this PR, and it is not this PR.**

```
✖ WITH the core lock, both updates survive
✖ the lock is released on a throwing path, not just the happy one
  Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'js-yaml'
    imported from /home/runner/work/career-ops/career-ops/tracker-utils.mjs
```

That is `web/tests/lib/tracker-lock.test.mjs` — your test from #2903. The job runs `npm ci` in `web/` only, so root dependencies are absent; the test dynamically imports the root `tracker-utils.mjs`, which imports `js-yaml`.

**It failed identically on #2903 itself before you merged it** (job 94910042400, same two tests, same error). So this is inherited, not caused: my diff touches `set-status.mjs`, `set-status-tests.mjs` and `route.ts`, none of which that test reads.

The part worth your attention is why it does not skip. The guard is

```js
const HAS_CORE = fs.existsSync(coreLock);
```

which asks whether the file is *there*, not whether it can be *imported*. In the web job the file is there and its dependencies are not, so the skip never fires and the test hard-fails instead. **The two tests that prove the lock works have never passed in CI.** CodeRabbit said as much on #2903 ("the new regression tests may not run in the web test environment") and it went in anyway — which is the same shape as the bug they guard: something that looks covered and is not.

Locally it passes for a reason worth knowing before anyone calls it a flake: Node walks up for `node_modules`, so in a checkout under a directory that has the root deps installed, `js-yaml` resolves from a parent. The failure only appears where the tree is isolated — which is CI.

Cheapest honest fix is to make the guard match what it is guarding, resolving the import rather than stat-ing the path, so an isolated web install skips instead of failing. Happy to send that as its own issue and PR against `web/tests/` — it is not mine to slip into this branch, and it should not ride on a PR you may not take.

## Recommendation

**Do not merge this yet, and not because of the red.** You said the non-reentrant lock has to be ruled on before delegation, not after, and I agree — this PR is exactly the thing that would make delegation the established path. The order I would run:

1. **#2898** — approved, green, and the rows this produces are invisible to `funnel-velocity.mjs` without it.
2. **The reentrancy guard against `tracker-utils.mjs`** — a process-local marker so a nested acquisition from a lock holder fails immediately with a named error instead of hanging out the timeout. My argument for enforcing the invariant rather than making the lock reentrant is in #2904; if you want it as a PR with the deadlock reproduced as a failing test first, say the word.
3. **This**, once 2 has landed.
4. **#2715 closed** — delegation makes it redundant; I have recommended closing it there.

Verified: `node test-all.mjs` 3938 passed / 0 failed, identical to `main`'s baseline. `npm --prefix web test` 249 passed. `npx tsc --noEmit` clean. Every check green except the inherited one above.

rubicon:
Both CodeRabbit findings fixed in `03e3253`.

**The Major one was right, and my own comment contradicted my code.** The line above `parseCliJson` said other text can precede the document; `indexOf("{")` then found the first brace anywhere, including inside that text. A ledger-append warning carrying a brace made the slice unparseable, so the route answered 500 for a write `set-status.mjs` had already committed, and `changed` and `statusLogged` went with it. The document is now read from the end: the last line that parses as a plain object wins, which also means a diagnostic object cannot shadow the result.

**The row selector had no check at all.** `n` is typed as a string but arrives from untrusted JSON, so an array or an object reached `String(n)` and sent `"[object Object]"` to `--row` — a process spawn and a lock acquisition spent to return the CLI's usage text. Agreed it is not an injection risk for the reason given: `execFile` takes an argv array and `--row` consumes the next token.

Both now live in `web/src/lib/status-cli.mjs` so they are testable without spawning anything, following the existing `src/lib` + `tests/lib` convention already used by `explore-error.mjs`.

Six tests in `web/tests/lib/status-cli.test.mjs`, confirmed non-vacuous by mutation rather than by inspection:

| Mutation | Result |
|---|---|
| restore `indexOf("{")` forward scan | `the JSON document is read even when a warning printed a brace first` and `the last JSON object wins` fail |
| accept any row selector | `a row selector that cannot name a row is refused before anything is spawned` fails |

Each is killed by the test whose name describes it. `npx tsc --noEmit` clean; web suite 249 to 255 tests, 0 failures.

The one red check here, `web typecheck + build`, is the root-deps failure inherited from `main` that #2928 fixes, not this branch.

santifer:
This is the structural half of the #2900 fix: the lock stopped the lost update, and converging on set-status.mjs as the single writer removes the second writer that made the race possible at all. Owning the CodeRabbit Major finding in public is the review culture this repo wants. Third one today — that's the author cap, the rest of your green queue rides the next cycle. Merged! 🚀

PR #2903, opened 2026-08-14
fix(web): take the core tracker lock in POST /api/status (lost update, #2900)

santifer:
## What does this PR do?

Fixes **#2900** (reported by @rubicon): `POST /api/status` was the only tracker writer that did not take the core's file lock.

**Why `atomicWrite` was not enough.** It makes the *write* atomic, not the *read-modify-write* around it. Two writers each read the whole tracker, each edit their own row, each rename their full copy back — the second discards the first's change. Every call succeeds, the file is well-formed at every instant, and one row silently reverts.

Reproduced before fixing, and that reproduction is now the **first test**:

```
row 1 → Applied     (wanted Applied)
row 2 → Evaluated   (wanted Interview)   ← lost, no error anywhere
```

The other writer isn't hypothetical: the web exists to be used **alongside** the CLI, and `set-status.mjs`, `mark-pdf-ready.mjs` and `reserve-report-num.mjs` all take this lock already.

## Why it survived review

`web/src/lib/core/run-registry.ts` **does** guard the tracker — but it's an in-memory counter scoped to this process, so it cannot see the user's CLI. Two guards with different scopes, and the route looked covered.

So `run-registry` now documents what it does **not** protect. That's worth as much as the fix: a reader who sees a guard stops looking for the one that's missing.

## Design notes — read from `tracker-utils.mjs`, not assumed

- **The lock is not reentrant** (no owner/pid self-check), so a holder must never invoke a core script that takes it too. Direct writers only; documented in the ACL.
- **A crashed holder does not wedge the tracker**: `lockCanRecover` reads the owner pid and recovers once that process is gone. But a *live* process leaking the lock keeps its pid alive, so recovery waits out `staleMs` — hence the mandatory `finally`, which has its own test.
- **5s timeout, not the CLI's 60s.** This is an HTTP handler; a minute of silence is indistinguishable from a hang. Contention answers **409** and the user retries.

## Test plan

- [x] `tracker-lock.test.mjs` asserts the failure **first** (unlocked writers lose data), then that the lock removes it — without the negative half, a lock that never engaged would pass just as happily
- [x] A third test proves the lock is released on a throwing path
- [x] `web` 249/249 · `npm run typecheck` clean
- [x] `node test-all.mjs` — 3938 passed, 0 failed (1 known warning)

## Not in this PR

`data/follow-ups.md` has **three core writers and no lock at all** (`followup-seed.mjs`, `stats.mjs`, `company-history.mjs`), so the web's follow-up routes aren't bypassing a guard — they're entering unprotected ground. Fixing that needs the core side first; tracked separately rather than half-solved here.

## Type of change

- [x] Bug fix

## Checklist

- [x] My changes respect the Data Contract

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when multiple status updates occur simultaneously, preventing updates from being lost.
  * Busy update operations now return a clear conflict response instead of failing unpredictably.
  * Unexpected update failures now return an appropriate server error response.
  * Locks are consistently released after completed or failed updates, reducing the risk of stalled operations.

* **Tests**
  * Added coverage for concurrent updates, error handling, and lock recovery.

### DIFF
--- a/web/src/app/api/status/route.ts
+++ b/web/src/app/api/status/route.ts
@@ -4,6 +4,7 @@ import path from "node:path";
 import { careerOpsRoot } from "@/lib/career-ops";
 import { canonicalizeStatus } from "@/lib/core/states";
 import { atomicWrite } from "@/lib/core/safe-write";
+import { withTrackerLock, TrackerBusyError } from "@/lib/core/tracker-lock";
 
 // Writeback: UPDATE the status cell of an EXISTING tracker row only. Never adds
 // rows — per the core data contract, new rows go through the TSV + merge flow.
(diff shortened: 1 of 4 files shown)


santifer:
Merged. Reported by @rubicon in #2900 — full write-up there; #2901 proposed the same diagnosis independently.

PR #2905, opened 2026-08-14
test(agent-inbox): assert the lock is never held twice, so a red run names its cause

LayerC0de:
Refs #2777. **Diagnostic scaffolding, not a fix** — no production code changes, independent of #2815 and #2835, and merges in any order.

I offered this on #2777 and said I'd ask before opening it. Opening it anyway so it's reviewable rather than hypothetical — close it without ceremony if you'd rather not carry it.

## The problem it solves

Case 7 reports **what** was lost. It cannot say **why**, and that has cost a round trip on every `windows-latest` failure. A red 7 reads:

```
❌ all 30 concurrently queued items survive — kept=27 of 30
```

Two very different explanations fit that equally well:

- the lock let two writers in, so their appends interleaved and one overwrote the other; or
- the lock held perfectly and the append itself lost a write.

Nothing in case 7 separates them — which is exactly where #2777 sits right now.

## What case 8 does

Measures the property in question — **mutual exclusion** — instead of its downstream symptom. Thirty processes take the same lock through the same `withPipelineLock` that `add()` uses, each records its own hold interval, and the parent looks for overlapping `[enter, exit]` pairs.

Read as a pair:

| | meaning |
|---|---|
| 7 red + 8 **green** | the lock held; suspect the append |
| 7 red + 8 **red** | the lock was double-held; the append is downstream |
| 8 red alone | a lock bug that has not yet cost an item |

## Two load-bearing details

**Each child writes to its own file.** A shared log would reintroduce the concurrent-append question under test, and a line missing from it could not be told apart from a lock failure.

**The children wait on a shared wall-clock barrier before acquiring** — and this one only surfaced because I ran the negative control. The first version had no barrier and **passed with mutual exclusion removed outright**. `spawn` starts thirty processes hundreds of milliseconds apart, so a short hold never overlapped another hold even with no lock at all: the stagger was doing the serialising and my assertion was decoration.

With the barrier, the same experiment fails as it should:

```
real lock                    ✅ no two holders were inside the critical section at the same time
mutual exclusion removed     ❌ #11 held until 74547166140000, #5 entered 74547157850500 | …
```

Verified both directions on Windows. Ten consecutive runs clean; whole file ~3.2s.

## Honest limits

- **It may never fire.** Its value is diagnostic-on-failure — it pays off only when CI is already red. I think that's worth ~1s of suite time given how many round trips #2777 has taken, but that's your call, not mine.
- I could **not** reproduce a double-hold on `main` with it (0 overlaps, 12 rounds × 30 procs, default config), which is the same negative result I reported on #2777. This test does not assert the bug exists; it makes the next red run say which bug it is.

## Summary by CodeRabbit

* **Tests**
  * Expanded regression coverage for concurrent processing scenarios.
  * Added verification that exclusive operations cannot overlap, helping protect data consistency during simultaneous activity.
  * Improved failure diagnostics across bare, asynchronous, typed, custom, and system-level errors.
  * Distinguished lock-acquisition failures from concurrent-access violations.
  * Validated behavior across 30 synchronized concurrent executions, including completion and result coverage checks.


LayerC0de:
@coderabbitai all three taken, in `184204f`. Two were ways this case could pass while measuring less than it claims — the exact failure it exists to prevent, which makes them worth more than their size.

**1. Fixed start time → counted readiness.** Right, and it is the same flaw one degree weaker than the one the barrier already fixed. A fixed window assumes every child is up inside it; a loaded runner can start one after the earlier holds have finished, reintroducing the false green *in the environment this case was written to explain*. Children now signal readiness and the parent releases only once all N have checked in. It also made the case faster — releasing when they are ready rather than waiting out a guess took the file from 3.2s to 2.4s.

**2. Ignored exit status → asserted.** Also right, and the nastier of the two: a child that died before writing left no error file and no trace, so the herd silently shrank, no overlap was found, and the case went green. Exit codes and record count are now both asserted, kept separate from the acquire-failure count so a crash, a failed acquire and a double-hold stay three distinguishable outcomes.

**3. Adjacent-only comparison → widest-exit.** Taken, with one correction to the reasoning. The **verdict was never wrong**. Sorted by entry, an interval contained inside a longer one always overlaps its own predecessor, so any overlap still turns the case red. Your example demonstrates it:

```
[1,10] [2,3] [4,5]
adjacent-only  detects: ["A/B"]        -> verdict RED
widest-exit    detects: ["A/B","A/C"]  -> verdict RED
```

What was actually wrong is the **roster**: it under-reported who was involved. That still matters here more than usual — this case's only job is to explain a red run — so it now tracks the widest exit and lists every holder. Fixed for the right reason rather than the stated one.

Verified both directions again after the change: green against the real lock, red against a build with mutual exclusion removed, naming the overlapping holders. Ten consecutive runs clean.

LayerC0de:
The `test (windows-latest)` red on the previous commit here was **not this PR** — filed separately as #2906.

`tracker-writer-lock-tests.mjs` was killed by `run()`'s flat 30s per-script budget (`exit null, signal SIGTERM` — killed, not a failed assertion). This branch touches one file, `agent-inbox-tests.mjs`, and that file passed on the same run (`✅ agent-inbox-tests.mjs runs OK`).

Confirmed by re-triggering CI with a **byte-identical tree** — `git commit --amend --no-edit`, tree hash unchanged `dd213a0 -> dd213a0` — which went green. Same code, different outcome.

For context in #2906: that script runs 13.4–20.1s on current `main` on a 24-core box, so it is already at ~67% of the 30s ceiling before a CI runner's load is added.

santifer:
Opening it rather than leaving the offer hanging was the right call, @LayerC0de — an offer to build something is hypothetical until it exists, and reviewing the real thing costs me less than deciding about a description of it.

And it's welcome, not something to close: **#2777 is the issue where I posted a failure I couldn't explain** — 30 concurrent children, every one exit 0, three items missing. Diagnostic scaffolding for exactly that is the piece I don't have. "No production code changes, independent of #2815 and #2835, merges in any order" is also the framing that makes it cheap to say yes to.

Worth connecting for you, since these are all landing in the same corner tonight:

- **#2815** may be the root cause of #2777 — `readLockOwner` collapsing *no stamp* and *couldn't read the stamp* into one `null`, so a live holder's lock can be recovered. I've flagged it as first in the queue next cycle.
- **#2906/#2907** is a different failure in the same file family: the lock test being SIGTERM'd on the 30 s budget. I'd seen that twice tonight and mislabelled it as a flake.
- **This PR** is what would let us tell those two apart, instead of me re-running and hoping.

Four of yours merged in this run and three more are open. That's a lot from one person in a day and the quality has held across all of them; the queue position is my constraint, not a comment on the work.

*(Edited: two identifiers were eaten by my own shell quoting in the original.)*

LayerC0de:
Good to hear it's wanted rather than tolerated — that changes what I'd do with it. If it's the piece you don't have for #2777, it's worth it being sharper than "an extra case in the file", so two things:

**The instrument currently asserts one property.** Non-overlapping holds answers *"was the lock held twice?"* — which is the question you actually posted. It does **not** answer the other half of the same failure: whether an append was lost while the lock held correctly. If `7 red + 8 green` comes back, that narrows it to the append but does not name it. If that pairing shows up in CI and you want the next step, the obvious one is recording what each writer believed it wrote and diffing against the file — say the word and I'll send it separately rather than growing this PR.

**It has not caught anything yet, and I'd rather that be visible than discovered later.** 0 overlaps in 12 rounds × 30 processes on `main` at the default config. It is a canary, not evidence; the value is entirely in the next red run.

Your ordering is the one I'd argue for too, for what it's worth: **instrument, then attribute, then fix.** With #2906/#2907 in the mix there are two distinct Windows failure modes in this area right now, and attributing either one before the instrument lands is how the wrong fix gets credited for the right outcome.

Both review rounds are addressed on this branch; CI green on all three platforms.

LayerC0de:
@santifer — a concrete instance of what this PR is for turned up an hour after your comment, and it is the strongest argument for it that I have.

`test (windows-latest)` on #2907 reproduced #2777 in CI, with your signature:

```
✅ every concurrent add exited cleanly      <- all 30 exit 0
❌ all 30 concurrently queued items survive — kept=29 of 30   <- item-1 gone
```

On a PR that changes `test-all.mjs` and `tests/helpers.mjs` and nothing else. Full write-up on #2777.

**What that run could not tell either of us is exactly what this PR adds.** It fits both live hypotheses equally well — the lock was held twice, or the lock held and the append lost a write — and there is nothing in the output to separate them. Had case 8 been in `main`, the same run would have come back:

- `7 red + 8 red` → the lock was double-held, and #2815's class of bug is implicated
- `7 red + 8 green` → the lock held; the append is where to look

Instead it produced another data point that fits both, which is the third one now (yours, plus this).

I am not arguing for queue position — **#2815 first is still right**, it stands on its own as a correctness fix, and this is scaffolding rather than a fix. But the cost of the scaffolding not being there is now measurable: three reproductions, zero attribution.

One thing I would change if you want it before merging: case 8 currently answers *"was the lock held twice?"* and stops. If you would rather it also answer the other half in the same run — each writer recording what it believed it wrote, diffed against the file — say so and I will add it here rather than as a fourth PR. It is a bigger case, but it would mean one red run answers the question completely instead of halving it.

santifer:
Yes to the second half, and yes to it being a separate PR — @LayerC0de, take that as the word.

Reasoning, since you asked rather than just building it:

**The pairing you describe is the whole value, and right now it's half-built.** `7 red + 8 red` names the lock. `7 red + 8 green` narrows it to the append and then stops — which is better than today but still ends in "look at the append somehow". Recording what each writer believed it wrote and diffing against the file is what turns that second branch into an answer instead of a direction.

**Separate PR is right for the reason you give and one more.** This one is reviewed, green on three platforms, and answers a question we currently cannot answer at all. Growing it delays a working instrument to make it a better one — and with three reproductions and zero attributions, the cost of waiting is now measurable rather than theoretical.

**On "it has not caught anything yet, and I'd rather that be visible":** that's the sentence that decided this for me. An instrument whose author volunteers its null result is one I can trust when it eventually goes red. 0 overlaps in 12 rounds is the honest baseline, and it belongs in the PR body exactly where you put it.

Your #2907 data point is the strongest argument in the thread and it changes my ordering slightly. It's the third reproduction that fits both hypotheses equally — and it landed on a PR touching `test-all.mjs` and `tests/helpers.mjs` and nothing else, which rules out the change under test as the cause about as cleanly as anything could.

So: **#2905 first, then #2815.** I said #2815 first earlier and I'm changing it. Not because #2815 got weaker — it stands on its own with a failing test against `main` — but because merging a fix into an area where three reproductions have gone unattributed means the next green run credits *something*, and we won't know what. Instrument, then attribute, then fix, which is the order you and I both argued for and I then broke by naming a cause in the same breath.

Both near the top of the next cycle. Send the write-half whenever; no rush on it, since it only matters once a `7 red + 8 green` shows up.

LayerC0de:
Understood on all three — order noted, and I'll hold the write-half.

**On timing:** I'd rather send it *after* this merges than now. It builds directly on case 8, so opening it today would mean a PR stacked on an unmerged branch — the exact "diff two branches to find the boundary" problem you thanked me for avoiding last time. Once #2905 is in `main` it's a clean single-commit PR against a case that already exists. And as you say, it only earns its keep the first time a `7 red + 8 green` shows up.

**On the reordering** — that's the right call and I want to name why, because it's a stronger argument than the one I made. I framed #2905 as "the instrument that separates the hypotheses". You've pointed out the sharper version: merging a fix into an area with three unattributed reproductions means **the next green run credits something, and nobody will know what.** That's not a sequencing preference, it's the difference between evidence and coincidence. #2815 doesn't get weaker by going second; it gets a clean attribution when it lands.

Nothing needed from me on either — both green, both mergeable, no rebases pending.

LayerC0de:
Pushed `2dbb589` — folding in the §7 extractor fix from #3035, since it's this file and nothing else, and a separate PR would have meant a second review surface for nine lines.

**Why it belongs here.** This PR exists so a red §7 names its cause instead of its symptom. macOS run 32166774680 was the first §7 failure with a *crash* to report, and §7 printed:

```
↳ item-4 exited 1: Node.js v24.18.0
```

That's the third-tier `lines.slice(-1)[0]` fallback — both real routes missed everything. Two defects, each sufficient alone:

1. **The name route could not match a bare `Error:`.** `^[A-Za-z_$]` consumes the `E`, leaving `[\w$]*(Error|Exception)` needing a *second* literal `Error`. Matched `TypeError:` and `LockTimeoutError:`; never plain `Error:`.
2. **The errno route omitted `ENOTEMPTY`** — a lock-directory removal racing another process's `owner.json` write, which prints as a bare `Error:` line and so fell through both routes at once.

Now anchored on Node's own caret (`<file>:<line>` / `<source>` / `^` / *the error*), because names can't be primary: a non-Error rejection prints the bare value with no name to match.

**§7c**, against real child stderr rather than fixtures — fixtures can't drift when Node changes its crash format, and that drift is what would restore the banner. With the fix reverted:

```
❌ cause survives: bare Error        — got "Node.js v24.18.0"
❌ cause survives: async bare Error  — got "Node.js v24.18.0"
❌ cause survives: non-Error reject  — got "Node.js v24.18.0"
❌ cause survives: undefined reject  — got "Node.js v24.18.0"
❌ cause survives: ENOTEMPTY ...     — got "Node.js v24.18.0"
✅ cause survives: typed Error          (worked before, still works)
✅ cause survives: custom Error         (worked before, still works)
```

Five reds, every one reproducing the exact observed CI string. Two further assertions are **pins, not repros** — they assert the patterns directly and stay green under revert, so a later tidy-up fails with the reason rather than silently regressing. Calling that out so the ten green ticks aren't read as ten independent proofs.

Worth being clear about scope: **this does not explain the loss in #3035.** It makes the next occurrence able to say something. The §8 assertion in this PR is still the part that answers whether the lock is the reason.

One thing I checked and could not confirm, so it is **not** in the diff: §7 resolves on `'exit'` rather than `'close'`, which in principle can return before stderr drains. It didn't bite this run — the `Node.js vX` banner is the last line Node writes and it was present, and stream data is ordered. My attempt to force a truncation was inconclusive (the fixture crashed before flushing, so it measured the child, not the parent). Happy to add it as insurance if you want it, but I'm not shipping a change I couldn't produce evidence for.

Local: `agent-inbox-tests.mjs` 33/33, `test-all.mjs` 3928 passed / 0 failed.

LayerC0de:
CI on `2dbb589`: ubuntu ✅, macOS ✅, CodeRabbit ✅, **windows ❌ — and the red is #2777, not this diff.**

```
7. concurrent adds do not lose items
7c. a crashed writer's cause survives extraction
  ✅ ×10
```

All 30 exited 0, no duplicates, `item-1` gone. That is the clean-exit loss mode from #3035 — reported there so far only on macOS, now reproduced on Windows. This PR touches one test file and cannot cause it; it's the flake this PR exists to characterise, showing up in the PR's own CI.

Two things worth having from that run:

**§7c passed 10/10 on Windows**, so the extractor fix isn't platform-specific.

**§8's verdict was truncated out of the log**, one assertion short of the answer:

```
✅ all 30 holders acquired the l
  ... (102 more chars)
```

`formatRunFailure` in `tests/helpers.mjs` keeps the **head** of a failing script's stdout (`t.slice(0, maxChars)`, 2000). For a suite printing ✅ per assertion, the head is the setup that passed and the tail is the failure summary plus the newest cases — so §8, being last, is the first thing discarded. Details and the suggested one-line fix (clip the tail, not the head) are on [#3035](https://github.com/santifer/career-ops/issues/3035#issuecomment-5334423876).

I have **not** touched `tests/helpers.mjs` here — it's #2907's file and #2907 is approved; pushing to it would dismiss that. Happy to do it whichever way you prefer.

So: this branch is green everywhere it can be, and the Windows red is the bug under investigation rather than a regression. Re-running it would likely go green — but given it's a live reproduction of #3035 on a new platform, it seemed worth reporting before clearing.

LayerC0de:
I took §8's own advice and read the pair. On run 32188447733 (windows-latest) it says:

```
8. the lock is never held by two processes at once
  ✅ all 30 holder processes exited cleanly
  ✅ all 30 holders recorded an interval
  ✅ all 30 holders acquired the lock
```

7 red + 8 green — "the lock held; suspect the append". That split did its job, and it is
right that the lock is innocent. But the culprit isn't the append either. It's
`ensureFile()`, and it is a real data-loss bug in `agent-inbox.mjs`, not a flake.

## The fingerprint

The lost item was **`item-1`**, and **`item-0` is the first line in the file**. Everything
else is intact, in order, and well-formed, and all 30 writers exited 0.

That combination rules out most of the candidate list on its own:

- not a crash — nothing exited non-zero;
- not two interleaved appends — there is no mangled line, and
  `appendFileSync` is atomic across processes on Windows (libuv opens `O_APPEND` as
  `FILE_APPEND_DATA` with `FILE_WRITE_DATA` stripped, which appends under the
  filesystem lock). I measured it: **25 rounds × 30 concurrent writers with no lock at
  all, zero losses.** A double-hold on its own cannot drop a line here;
- not a truncating rewrite of the whole queue — that loses a *prefix*, and the other 29
  items all survived, in order.

What is left is a whole-file write that landed on top of exactly one already-appended
line. There is exactly one whole-file write in the `add` path.

## The mechanism

```js
function ensureFile() {
  if (existsSync(PATH)) return;          // (1)
  ...
  writeFileSync(PATH, HEADER, { flag: 'wx' });   // (2) open(O_CREAT|O_EXCL) ... then write()
}
```

`ensureFile()` runs **outside** the lock. `wx` was added to close the case where two
writers both pass (1) and the second one's default-`w` write truncates the first's item.
It does close that one. But `wx` makes the **create** atomic, not the
**initialisation**: `writeFileSync` is `open()` *then* `write()`, and between those two
syscalls the file **exists and is zero bytes**.

So there is a third participant nobody guarded against — a writer that arrives *inside*
that window, sees `existsSync === true`, skips creation entirely, takes the lock
correctly, and appends into a file the creator is about to overwrite from offset 0.

The window is trivially observable. A second process polling `existsSync` and stat-ing
the moment the file appears saw it at **0 bytes in 303 of 400 rounds**:

```
HEADER is 479 bytes.
observed sizes: 0B x303, 479B x97
RESULT: the wx window IS observable — a second process can see the file exist while it is still empty.
```

## Caught in the act

Reproduced locally by pinning the whole process tree to 2 cores (`start /affinity 3`),
which is the one thing that widens the window: the creator only has to lose its
timeslice between `open()` and `write()`. Instrumented writers record the file size they
saw before and after their own append, inside the lock:

```
=== ROUND 15 FAILED — kept=26/30 missing=[item-2,item-12,item-16,item-28] nonzeroExits=0
  writers that won the wx create: item-19

  item-2 : ensureFile:existsSync seen=true → preLock size=0 → appended before=0   after=34
  item-12:                                                     appended before=34  after=69
  item-28:                                                     appended before=69  after=104
  item-16:                                                     appended before=104 after=139
           → postCheck mine=FALSE          ← watched its own line disappear

  --- append chain (before -> after) ---
          item-2:    0 -> 34
          item-12:  34 -> 69
          item-28:  69 -> 104
          item-16: 104 -> 139
    BREAK item-23: 479 -> 514   [expected before=139]
```

`item-19` won the `wx` create, so the file existed at 0 bytes, and then it was
descheduled before its `write()`. Four writers walked through the window, each one
**correctly serialised by the lock** — every `before` equals the previous `after`, which
is exactly what §8 reports and it is true. Then `item-19`'s `write(HEADER)` landed at
offset 0 and replaced all 139 bytes of them. `479` is not an arbitrary number: it is
`Buffer.byteLength(HEADER)` to the byte.

Note `nonzeroExits=0`. This is CI's symptom, not an adjacent one.

Scaled down to CI's run: `item-0` won the create, `item-1` appended its 34 bytes into the
empty file, `item-0`'s 479-byte HEADER wiped it, `item-0` then appended its own line —
which is why `item-0` is first and `item-1` is the only casualty. The two
earliest-spawned writers are the pair most likely to be inside `ensureFile()` together,
which is why the victim is `item-1` and not `item-23`.

Nothing here is Windows-specific in principle — the window is one scheduling slice, and
windows-latest is simply where the runner is loaded enough to lose it.

## Fix

Move `ensureFile()` **inside** `withPipelineLock`, so no writer can append while another
is still initialising the file:

```js
await withPipelineLock(PATH, () => {
  ensureFile();
  const separator = needsLeadingNewline(PATH) ? '\n' : '';
  appendFileSync(PATH, `${separator}- [ ] ${stamp()} — ${text}\n`);
}, { timeoutMs: 30_000 });
```

It reuses the lock already being taken one line later, adds no second mechanism, and the
critical section stays sub-millisecond. A/B, arms alternated round-by-round on the same
pinned 2 cores:

```
                                        SHIPPED   FIXED
  rounds (100 each x 30 writers)
  clean ................................. 94       99
  SILENT loss (all children exited 0) ...  1        0
  items lost in total ...................  6        1
  appends into a 0-byte file ............  1        0
```

Read that honestly, because it is weaker than it looks: at ~1–2 events per 100 rounds
those counts are too small to carry the argument on their own, and the crash-round
difference belongs to the *second* bug below, not to this fix. A follow-up run at 50
writers pinned to a **single** core produced **zero events in both arms** — a null
result, not a confirmation: one core removes the true parallelism the race needs, so the
creator finishes `writeFileSync` before anyone else is scheduled into the window. The
reproduction is specifically a ≥2-core, oversubscribed phenomenon.

So the A/B is corroboration, not proof. The proof is the trace above plus the structural
argument: with `ensureFile()` inside the lock a writer can only observe the file after
the creator has released, and the creator writes HEADER before it releases, so the window
is not narrowed but unreachable. `appends into a 0-byte file` is the precursor event
instrumented directly, and it is the counter that should sit at zero by construction.

## Separately: a second, louder bug

The same repro also produced this, which is *not* the failure above (it exits non-zero):

```
Error: ENOENT: no such file or directory, open '…\agent-inbox.md.lock\owner.json'
    at acquirePipelineLock (pipeline-lock.mjs:258)
```

Tracing every lock-directory event from every process, time-ordered, names it in one
window (microseconds, same round, 7 clean rounds before it):

```
t=…368689  pid=2816   canRecover:VANISHED->true  code=ENOENT
t=…369161  pid=24864  MKDIR-OK (acquired; owner.json not yet written)
t=…369229  pid=2816   RMSYNC-RECOVER (deleting lockDir)  existsNow=true
t=…371071  pid=24864  OWNER-WRITE-FAILED  code=ENOENT
```

`lockCanRecover()`'s `statSync` threw ENOENT because the lock genuinely *was* absent at
that instant, and the `catch` returns `true` — commented *"vanished — nothing to recover,
retry acquisition"*. But the caller does not retry acquisition; it falls into
`rmSync(lockDir, …)`. 472 µs later another writer legitimately created a fresh lock
there, and 68 µs after that the stale verdict was acted on and deleted it —
`existsNow=true` is the proof it removed a directory that was present.

The ENOENT crash is the lucky outcome. Had the victim written `owner.json` 68 µs earlier
it would have entered the critical section while the deleter `continue`d and re-`mkdir`ed
the now-free path: **two holders, no error at all**. That one is §8's business.

The verdict "the lock is gone" must not authorise "so delete the lock" — the vanished
case wants a distinct sentinel that skips the `rmSync` and just retries.

## What this means for this PR

§8 is not wrong and it is not wasted — it is the check that made this tractable. "7 red +
8 green" is a true reading, and it eliminated the lock in one run instead of a round
trip. I'd suggest landing it as-is; the `ensureFile()` fix belongs in its own PR against
`agent-inbox.mjs`, and once it lands §7 should stop needing re-runs.

LayerC0de:
Follow-up on one thing I flagged above as weak. The A/B in that comment was 100 rounds
per arm and I said plainly it was underpowered — 1 event against 0. It has now run at
**400 rounds per arm** (30 concurrent writers, pinned to 2 cores, arms alternated round
by round on the same machine), and the picture is firmer:

|                            | shipped | fixed |
|----------------------------|---------|-------|
| appends into a 0-byte file | 4       | **0** |
| silent loss (all exited 0) | 3       | **0** |
| crash rounds               | 15      | 9     |

`appends into a 0-byte file` is the precursor event instrumented directly — a writer that
did not create the file but still saw it at 0 bytes — and 4 → 0 is now backed by enough
events to carry weight, alongside the structural argument that the window is unreachable
once the seed is under the lock.

Two corrections to keep this honest:

- **The crash-round column is not a result.** 15 vs 9 is noise at this sample size and I
  am not claiming the fix moved it. Those are the *second* bug — `lockCanRecover()`'s
  "vanished" verdict authorising a delete — which is independent of the `ensureFile()`
  change and unfixed in both arms.
- **So §7 should not be expected to go quiet from the `ensureFile()` fix alone.** It
  removes the silent `kept=29 of 30`; the remaining failure mode is a non-zero exit,
  which is exactly the shape this PR's instrumentation makes self-describing. That is an
  argument for landing §7c and §8, not against.

The `ensureFile()` fix is up as #3118. The `pipeline-lock.mjs` one is next and will be
its own PR.

santifer:
Landing it, @LayerC0de, and the reason is the table in your last comment as much as the code. Case 8 measures mutual exclusion directly instead of inferring it from a lost item, which is what turns a red 7 from a round trip into a verdict. Two things I want on the record: labelling the crash-round column as noise rather than as a result is the kind of honesty that makes the other two rows worth believing, and saying plainly that §7 should NOT be expected to go quiet from the ensureFile fix alone is an argument for the instrumentation, not against it. #3118 is in; this makes the next one readable. Merged 🚀

LayerC0de:
§9 measures the right property, and I can now say how well it measures it.

I built a harness that drives `withPipelineLock` directly and looks for overlapping
`[enter, exit]` intervals across processes, which is what §9 does. Run in §9's exact shape
(30 processes, one acquisition each, 8 ms holds, released from a barrier) against current
`main`, over 60 runs:

| | |
|---|---|
| acquisitions | 1,807 |
| overlapping holds | 21 |
| runs with at least one overlap | 14 of 60 |

So §9 does catch the `pipeline-lock` defect that #3120 fixes. It catches it about one run in
four. That is enough to have shown up as an intermittent §9 failure over a handful of CI
runs, and not enough to name the cause from any single one.

Two things follow from that number.

**Raise the acquisitions per holder.** Each holder acquires once and exits, so a run only
creates 30 chances for the race to happen. Looping inside the holder costs almost nothing:
at 250 acquisitions per process, every round of my harness fires instead of a quarter of
them. Same processes, same barrier, same assertion, far more sensitivity for the same
wall-clock.

**Keep §9 after the lock fixes land.** The same 60 runs with #3120 applied still turn up 2
overlaps in 2 runs, and the traces name a different cause. A caller reads a dead owner's
stamp, judges the lock stale, and by the time its `rm` runs a live process owns the path:

```
{"event":"RM-RECOVER","reason":"stale","existsNow":true,"victimPid":46108,"victimAlive":true}
```

Four such deletions landed on a live owner across those runs. #3120 closes the case where
the verdict came from an absent directory. This is a directory that was genuinely stale when
read and stopped being stale before the delete landed, and nothing open right now fixes it.
§9's shape is where that lives, because every career-ops CLI process takes the lock once and
exits.

I would land this ahead of the lock PRs rather than after. It is the only check in the repo
that measures mutual exclusion at all, and it is currently pointing at a bug none of the
open PRs close.

santifer:
This is the comment I hoped §9 would eventually earn, @LayerC0de. Three answers:

1. **250 acquisitions per holder: yes.** Same processes, same barrier, ~30× the chances per run for free — that's strictly better sensitivity with no new machinery. Send it as its own small PR and it gets a same-day review.

2. **The live-owner deletion is its own bug and deserves its own issue.** Your trace names it precisely: the verdict was honest when read (`reason: stale`) and false by the time the `rm` landed (`victimAlive: true`) — a TOCTOU between staleness judgment and deletion that #3120 structurally cannot close, because #3120 fixes how the verdict is *formed*, not the gap between verdict and act. Please file it with that RM-RECOVER trace as the body; it'll get `confirmed` on arrival since you've already done the reproduction work. (Fix shape probably wants the delete to re-verify ownership atomically — rename-the-directory-then-check, or verdict-under-the-same-mkdir-race — but that's for the issue thread.)

3. **§9 stays after the fixes land.** 'The only check in the repo that measures mutual exclusion' is the argument, and 2-in-60 post-fix is exactly why it survives: a quiet §9 would have hidden that residue forever.

PR #2907, opened 2026-08-15
test: let a script declare its own budget, and give the lock matrix 180s

### LINKED ISSUE #2906: tracker-writer-lock-tests.mjs now exceeds test-all's 30s per-script budget on windows-latest and is SIGTERM'd
### Code of Conduct

### Existing issues

- [x] I searched existing issues. The closest is **#2331** (closed), same file — but a different mechanism; see "Not #2331" below.

### What happened?

`tracker-writer-lock-tests.mjs` was killed on `windows-latest` while every other check on the same commit passed:

```
❌ tracker-writer-lock-tests.mjs crashed (exit null, signal SIGTERM)
   stdout: PASS normalize-statuses: contends on the shared lock before reading or writing
           NOTE normalize-statuses: recover guard not sampled within 2000ms — fell back to
                timing-dependent ordering for this case
           PASS normalize-statuses: completes the intended update after lock release
```

`signal SIGTERM` with `exit null` is the give-away: the script did not fail an assertion, it was **killed**. That is `run()` in `tests/helpers.mjs:178`, which gives every script in the `scripts` list a flat budget:

```js
return execFileSync(exe, args, { cwd: ROOT, encoding: 'utf-8', timeout: 30000, ...opts }).trim();
```

The stdout above shows it partway through the matrix and still passing when the axe fell.

**The margin is already thin on fast hardware.** Four consecutive runs of the file on current `main` (`d9c4fd0`), unmodified, on a 24-core Windows box:

```
run 1  13.4s   exit 0   27 PASS
run 2  17.8s   exit 0   27 PASS
run 3  19.8s   exit 0   27 PASS
run 4  20.1s   exit 0   26 PASS
```

Worst sample is **20.1s against a 30s kill — 67% of the budget consumed** on a machine far faster than a 2-core CI runner. Pinned to 2 cores it still lands ~15s. There is not much left for a loaded runner to absorb.

Note run 4 also produced 26 PASS rather than 27: the count itself varies, because some cases degrade to the "not sampled within 2000ms" path rather than asserting. So the file's runtime *and* its assertion count are both load-dependent.

### Steps to reproduce

Intermittent and load-dependent, as before. Concretely observed:

- Failed: run `31857086407`, job `94943811970` (`test (windows-latest)`; all other checks green on the same commit).
- **Passed on a byte-identical tree**: I re-triggered with `git commit --amend --no-edit` and verified the tree hash was unchanged (`dd213a0 -> dd213a0`), then force-pushed. Run `31857601383`, job `94945183150` — green. Same code, different outcome.

Locally: `node tracker-writer-lock-tests.mjs` on `main`, repeatedly, and watch the spread above.

### Not #2331

#2331 was the same file and is closed, so it is worth being explicit about why this is not a reopen:

- **#2331 was about the file's INTERNAL harness waits** — 2–5s budgets for a spawned Node process to start and flush a prompt, which fired as assertion failures.
- **That half is fixed and visibly working here.** `d2b4ccd` made those waits degrade instead of failing, which is exactly what the `NOTE … fell back to timing-dependent ordering` line in the log above is.
- **This is the OUTER budget** — `run()`'s flat 30 s for the whole script — and it manifests as `SIGTERM`, not as a failed assertion.

Ironically the first fix contributes to the second: waits that now degrade gracefully rather than failing fast mean the script spends longer before finishing, against an outer limit that did not move.

The other half of #2331 is also fixed and earned its keep — `formatRunFailure()` is why the log above carries the child's stdout at all. Diagnosing this took minutes rather than a round trip.

### Expected behavior

A correctness test should not be killed for being slow on a slow machine. Options, roughly in order of how targeted they are:

1. **Give this script a longer budget.** `run()` already takes an options object, so a per-script override is a small change — no global loosening, and the outlier is treated as an outlier.
2. **Make the matrix cheaper.** 27 cases each spawning writers is the cost; some may be collapsible without losing coverage.
3. **Raise the flat 30 s.** Simplest, but it relaxes the bound for every script to fix one.

Happy to send a PR for (1) if you want it — I have not, because which of the three you want is a judgement about the suite rather than a bug fix, and #2331's history suggests you have views on which timeouts are semantic and which are just machine speed.

### Environment

- Observed on `windows-latest` in CI; local timings on Windows 11, Node v24, 24-core.
- `ubuntu-latest` and `macos-latest` pass consistently.

LayerC0de:
Closes #2906.

Option (1) from that issue — the most targeted of the three. Happy to close this if you'd rather take (2) or (3); the issue lays out all three and this is a judgement about the suite, not an obvious call.

## What broke

`tracker-writer-lock-tests.mjs` was killed on `windows-latest` while every other check on the same commit passed:

```
❌ tracker-writer-lock-tests.mjs crashed (exit null, signal SIGTERM)
   stdout: PASS normalize-statuses: contends on the shared lock before reading or writing
           NOTE normalize-statuses: recover guard not sampled within 2000ms — …
           PASS normalize-statuses: completes the intended update after lock release
```

`signal SIGTERM` with `exit null` means **terminated, not failed** — that's `run()`'s flat 30s in `tests/helpers.mjs:178`. The script hadn't gone wrong; it was still emitting `PASS` lines when the axe fell.

The margin was already thin on fast hardware. Four consecutive runs on current `main`, 24-core Windows box:

```
13.4s   17.8s   19.8s   20.1s
```

Worst sample is **67% of the budget** before a 2-core runner's load is added. The assertion count moves too — 26 or 27 `PASS` depending on whether the guard gets sampled in time — so both runtime and coverage are load-dependent. That cost is the behaviour under test: 27 contention cases, each spawning competing writer processes.

## The change

An entry in the `scripts` list may now carry `timeoutMs`, and exactly one uses it.

Raising `run()`'s default instead would relax the bound for all ~40 scripts to fix one outlier, and a **new** script that starts taking half a minute should still fail loudly rather than inherit a budget sized for this one.

180s is ~9x the worst local sample. Deliberately generous: the failure being prevented is a correctness test killed for running on a slow machine. A bound that's too loose costs a slow CI job; too tight costs a red run that says nothing about the code.

## Two details I checked rather than assumed

**The override is spread in, not defaulted.** `{ timeout: timeoutMs }` with an absent value yields `timeout: undefined`, and `execFileSync` reads that as **no timeout at all** — a hung script would then hang CI instead of being killed. Both branches verified:

```
timeoutMs=undefined   'timeout' in opts: false   effective: 30000
timeoutMs=180000      'timeout' in opts: true    effective: 180000
naive spread          effective: undefined        ← the trap
```

**The override actually reaches `execFileSync`.** Verified by inverting it — with `timeout: 1500` the same script is killed at 1514ms with `SIGTERM`. So the plumbing is live, not silently ignored. That check matters here because a dead override and a working one look identical when the suite is green.

## Verification

Full suite: **3923 passed, 1 failed** — the pre-existing `js-yaml` resolution failure in the temp-dir e2e copy, which fails identically on `main`. `✅ tracker-writer-lock-tests.mjs runs OK`.

## Note

This does not make that script faster, and #2906's option (2) — collapsing cases in the matrix — is still on the table as a follow-up. This buys headroom so a correctness test stops being killed for machine speed; it doesn't address why it needs 20s.

## Summary by CodeRabbit

* **Tests**
  * Increased the timeout available for tracker writer lock tests to 180 seconds.
  * Preserved the shared 30-second default timeout for other test scripts.
  * Added validation for configured script timeouts.
  * Added duration tracking and warnings when passing scripts use more than 75% of their allotted time.
  * Added coverage to verify that test execution enforces the default timeout.

### DIFF
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -293,7 +293,21 @@ const scripts = [
   { name: 'followup-seed-tests.mjs', expectExit: 0 },
   { name: 'paste-reply-tests.mjs', expectExit: 0 },
   { name: 'set-status-tests.mjs', expectExit: 0 },
-  { name: 'tracker-writer-lock-tests.mjs', expectExit: 0 },
+  // The one script in this list that genuinely needs longer than the shared
+  // budget. It spawns competing writer processes for 27 contention cases, and
+  // that cost is the behaviour under test rather than slack to be trimmed.
+  //
+  // Measured on current main on a 24-core Windows box, four consecutive runs:
+  // 13.4s, 17.8s, 19.8s, 20.1s — already 67% of the default 30s before a
+  // 2-core CI runner's load is added. On windows-latest it crossed the line and
+  // was killed mid-matrix (`exit null, signal SIGTERM`), while every other
+  // check on the same commit passed (#2906).
+  //
+  // Raised here rather than in run()'s default so the outlier is treated as an
+  // outlier: every other script keeps the 30s bound, and a NEW script that
+  // starts taking half a minute still fails loudly instead of inheriting a
+  // budget sized for this one.
+  { name: 'tracker-writer-lock-tests.mjs', expectExit: 0, timeoutMs: 180_000 },
   // Root-level standalone suites shipped in SYSTEM_PATHS but previously never
   // executed by CI (issue #1624). All are fast (<0.5s each), so they run in
   // both quick and full mode like their siblings above.
@@ -378,13 +392,18 @@ try {
     'utf-8'
   );
 
-  for (const { name, allowFail } of scripts) {
+  for (const { name, allowFail, timeoutMs } of scripts) {
     const parts = name.split(' ');
     const scriptFile = parts[0];
     const args = parts.slice(1);
+    // Only an entry that declares `timeoutMs` overrides run()'s shared budget,
+    // and it is spread in rather than defaulted so an absent value cannot
+    // quietly become `timeout: undefined` — execFileSync reads that as "no
+    // timeout at all", which would turn a hung script into a hung CI job.
     const result = run(NODE, [join(scriptTmp, scriptFile), ...args], {
       cwd: scriptTmp,
       stdio: ['pipe', 'pipe', 'pipe'],
+      ...(timeoutMs ? { timeout: timeoutMs } : {}),
     });
     if (result !== null) {
       pass(`${name} runs OK`);
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -31,7 +31,7 @@ import { tmpdir } from 'os';
 import { promisify } from 'util';
 import { fileURLToPath, pathToFileURL } from 'url';
 import * as yaml from 'js-yaml';
-import { pass, fail, warn, run, lastRunFailure, formatRunFailure, fileExists, finish, ROOT, QUICK, NODE, getBash, toBashPath } from './tests/helpers.mjs';
+import { pass, fail, warn, run, lastRunFailure, formatRunFailure, fileExists, finish, ROOT, QUICK, NODE, DEFAULT_SCRIPT_TIMEOUT_MS, getBash, toBashPath } from './tests/helpers.mjs';
 import { flagValue, hasFlag } from './lib/cli-flags.mjs';
 
 /**
@@ -253,6 +253,12 @@ mjsFiles.forEach((f, i) => {
 
 console.log('\n2. Script execution (graceful on empty data)');
 
+// How much of its budget a script may consume before the suite says so. High
+// enough that ordinary CI variance stays quiet, low enough to leave room to act
+// before the kill: at 0.75 a 30s script warns at 22.5s, nearly a full run of
+// headroom.
+const SLOW_SCRIPT_WARN_FRACTION = 0.75;
+
 const scripts = [
   { name: 'cv-sync-check.mjs', expectExit: 1, allowFail: true }, // fails without cv.md (normal in repo)
   { name: 'verify-pipeline.mjs', expectExit: 0 },
@@ -400,11 +406,29 @@ try {
     // and it is spread in rather than defaulted so an absent value cannot
     // quietly become `timeout: undefined` — execFileSync reads that as "no
     // timeout at all", which would turn a hung script into a hung CI job.
+    const budgetMs = timeoutMs ?? DEFAULT_SCRIPT_TIMEOUT_MS;
+    const startedAt = Date.now();
     const result = run(NODE, [join(scriptTmp, scriptFile), ...args], {
       cwd: scriptTmp,
       stdio: ['pipe', 'pipe', 'pipe'],
       ...(timeoutMs ? { timeout: timeoutMs } : {}),
     });
+    const elapsedMs = Date.now() - startedAt;
+    // A budget a script can raise for itself is a place to hide in, unless
+    // something still notices it creeping. Nothing did: the reason
+    // tracker-writer-lock-tests.mjs was killed rather than flagged is that
+    // spending 29 of its 30 seconds looked exactly like spending 2 — the suite
+    // reported "runs OK" either way, right up to the run where it did not.
+    //
+    // So the ceiling is not the only signal any more. A script that eats most
+    // of its budget says so while it is still passing, which is the point at
+    // which someone can act. This is a WARNING rather than a failure on
+    // purpose: a loaded runner is a normal reason to be slow, and turning that
+    // into a red run would trade one false failure for another.
+    if (result !== null && elapsedMs > budgetMs * SLOW_SCRIPT_WARN_FRACTION) {
+      warn(`${name} used ${(elapsedMs / 1000).toFixed(1)}s of its ${(budgetMs / 1000).toFixed(0)}s budget `
+        + `(${Math.round((elapsedMs / budgetMs) * 100)}%) — it is passing, but it is close to being killed for time`);
+    }
     if (result !== null) {
       pass(`${name} runs OK`);
     } else if (allowFail) {
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -417,9 +417,14 @@ try {
     // definition, and quietly substituting the default would hide it behind the
     // very budget it was trying to set. `timeoutMs: undefined` is caught too —
     // writing the key at all is a claim, and an unusable claim is a mistake.
+    // Integer, not merely finite: execFileSync rejects a fractional timeout with
+    // ERR_OUT_OF_RANGE, so `timeoutMs: 0.5` would pass a "looks like a number"
+    // check here and then blow up inside the run with an error about neither
+    // this list nor this script. Validation that stops short of what the
+    // consumer accepts just moves the failure somewhere less legible.
     const declared = Object.hasOwn(script, 'timeoutMs');
-    if (declared && (!Number.isFinite(timeoutMs) || timeoutMs <= 0)) {
-      fail(`${name} declares an unusable timeoutMs (${String(timeoutMs)}) — must be a positive, finite number of milliseconds`);
+    if (declared && (!Number.isInteger(timeoutMs) || timeoutMs <= 0)) {
+      fail(`${name} declares an unusable timeoutMs (${String(timeoutMs)}) — must be a positive integer number of milliseconds`);
       continue;
     }
     const budgetMs = declared ? timeoutMs : DEFAULT_SCRIPT_TIMEOUT_MS;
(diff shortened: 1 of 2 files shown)


santifer:
Answered on #2906 with something you should have: **I saw this exact SIGTERM twice tonight** on #2610 and #2682 — two PRs that touch nothing near that file — and wrote it off as a Windows flake both times. Re-ran, went green, moved on. Your issue is the same signature with the mechanism attached.

On the choice between your three options: option (1) is the most targeted and that's usually right, but *letting a script declare its own budget* is a change to how the whole suite reasons about time, so it deserves the read at the start of a session rather than after fifteen merges. Not tonight — merge ceiling — but high in the queue, because this is currently taxing other contributors' PRs with red runs that aren't theirs.

One thing I'd want in the PR body when it's reviewed: **what stops a declared budget from becoming a place to hide a genuinely slow test.** If a script can raise its own ceiling, the mechanism that catches "this got 10× slower" needs to still exist — otherwise the budget stops meaning anything and we find out through a timeout again.

LayerC0de:
> what stops a declared budget from becoming a place to hide a genuinely slow test

Right question, and the honest answer was **nothing — including before this PR.** That is exactly how the file got here: spending 29 of its 30 seconds looked identical to spending 2, because the suite printed `runs OK` either way, right up to the run where it printed `crashed (exit null, signal SIGTERM)`. The ceiling was the only signal, and a ceiling only speaks once it has been crossed — at which point the information is a dead CI job and a contributor's red PR.

Rather than argue that in the body, I closed it. Pushed as `753852f`: the ceiling is no longer the only signal.

A script that consumes more than **75%** of its budget now says so **while still passing**:

```
⚠️  tracker-writer-lock-tests.mjs used 18.2s of its 24s budget (76%)
    — it is passing, but it is close to being killed for time
```

Three things that make it answer the objection rather than decorate it:

- **It applies to every script, not just ones that declare a budget.** Raising a ceiling buys headroom, not silence: a 180s script that starts taking 140s is as loud as a 30s script taking 23s. Your "this got 10× slower" case is covered on both sides of the override.
- **A warning, not a failure.** A loaded runner is an ordinary reason to be slow, and making that red would trade one false failure for another — the exact thing this PR exists to stop.
- **0.75 leaves room to act rather than to be surprised.** A 30s script warns at 22.5s: close to a full run of headroom against the spread I measured.

Verified by inversion, since a dead canary and a live one look identical when everything is green: with the budget temporarily set to 24s it fires at 76% as shown; at the real 180s it stays silent. Full suite 3923 passed, 1 failed (the pre-existing `js-yaml` temp-dir failure, identical on `main`).

One implementation note worth surfacing because it looks like an oversight otherwise: `DEFAULT_SCRIPT_TIMEOUT_MS` is exported from `tests/helpers.mjs` but **not** substituted into `run()`'s `execFileSync` call. The comment above that call asks for it to stay byte-identical so CodeQL does not re-attribute its "uncontrolled command line" finding to whichever PR touched the line. The constant and the literal are linked by a comment rather than by the compiler, and both say so.

---

Separately — thank you for the note on #2906 about having seen that SIGTERM twice on #2610 and #2682 and re-run past it. That is the more useful half of the report and I did not have it: it turns "a test is slow" into "this is currently taxing other contributors' PRs with red runs that are not theirs", which is a much better reason to prioritise it than mine was.

LayerC0de:
@coderabbitai both taken in `4f7febd`, and the first was a defect I introduced in this PR rather than a hardening suggestion.

**The override and the warning disagreed about what a declared budget means.** `budgetMs` used `??`, which accepts `0`; the spread into `run()` used truthiness, which rejects it. So one value had two meanings in adjacent lines:

```
value      before: budget / override        after
undefined  30000   / default                unchanged
180000     180000  / 180000                 unchanged
0          0       / default    (Inf%)      fails loudly
-5         -5      / -5                     fails loudly
NaN        NaN     / default    (NaN%)      fails loudly
Infinity   Infinity/ Infinity               fails loudly
```

`timeoutMs: 0` would have divided by zero and reported `Infinity%`; `-5` and `Infinity` went to `execFileSync` unchallenged. Resolved once now, and a declared-but-unusable value **fails loudly** rather than falling back — the list is hand-edited, so a bad entry is a typo in the suite's own definition, and quietly substituting the default would hide it behind the very budget it was trying to set. The two valid cases behave exactly as before.

**The drift point was right, and it was the weak spot in my own design.** I wrote a comment asking the exported constant and `run()`'s literal to stay in step, then said in the commit message that they were "linked by a comment, not by the compiler" — which is an accurate description of a gap, not a solution. This repo's own doctrine is that reinforcement without enforcement decays.

Case **59c** now reads the literal back out of the source and compares. Verified both ways:

```
aligned   ✅ run()'s enforced timeout (30000ms) matches the exported DEFAULT_SCRIPT_TIMEOUT_MS
drifted   ❌ budget drift: run() enforces 30000ms but DEFAULT_SCRIPT_TIMEOUT_MS is 45000ms
```

It also fails deliberately if the literal can no longer be found, so refactoring that `execFileSync` call cannot silently retire the check — the failure says to update the check alongside it.

Full suite: 3924 passed, 1 failed (the pre-existing `js-yaml` temp-dir resolution failure, identical on `main`).

LayerC0de:
@coderabbitai right, and it is the same bug I "fixed" last round wearing different clothes. Taken in `0f81838`.

Both drafts inferred **whether** a budget was declared from **what it says**, and every in-band marker is also a value someone can write:

```
draft 1   ?? for the budget, truthiness for the override
          timeoutMs: 0  ->  "no budget" to one line, "default" to the next
draft 2   null as the absent-marker
          timeoutMs: null  ->  skipped validation, silently took the default
```

So I did not fix the class last round, I relocated it. `Object.hasOwn` answers the presence question directly and cannot be spoofed by any value, which removes the sentinel rather than choosing a better one.

`timeoutMs: undefined` now fails as well — writing the key at all is a claim, and an unusable claim is a typo worth hearing about. Only omitting it means "use the default".

```
entry                     draft 2          now
{}                        budget 30000     budget 30000    unchanged
{ timeoutMs: 180_000 }    budget 180000    budget 180000   unchanged
{ timeoutMs: 0 }          fails            fails
{ timeoutMs: null }       budget 30000     fails           <- fixed
{ timeoutMs: undefined }  budget 30000     fails           <- fixed
{ timeoutMs: -5 }         fails            fails
{ timeoutMs: NaN }        fails            fails
{ timeoutMs: Infinity }   fails            fails
```

Both real cases are untouched; only the malformed ones move, and they move from silence to a named failure.

Full suite: 3924 passed, 1 failed (the pre-existing `js-yaml` temp-dir resolution failure, identical on `main`).

Worth noting for whoever reviews this: the PR is now four rounds deep on six lines — budget, canary, validation, sentinel — and each round the thing meant to catch problems was the thing with the problem. If that reads as churn rather than convergence, the honest alternative is dropping the override entirely and taking option (2) from #2906 instead: make the lock matrix cheaper so no script needs a special budget.

LayerC0de:
@coderabbitai both taken in `bb79c7a`, and both verified rather than assumed — the second one only partly the way you framed it.

**Integer, not merely finite.** Confirmed directly: `execFileSync` rejects `0.5` with `ERR_OUT_OF_RANGE`. So the check passed a value its own consumer refuses, and the failure resurfaced inside the run with an error naming neither this list nor the script. Validation that stops short of what the consumer accepts hasn't prevented anything, only moved it somewhere harder to read. `Number.isInteger` now, and the message says so.

**The drift check could measure the wrong literal — with a correction.** You said it "accepts the first match", which is true, but the first match is currently `run()`'s, so today it is right by luck rather than by construction. The failure needs a timeout-bearing call added *above* `run()`, and then it is real:

```
unrelated call BELOW run() gains timeout: 5000   loose -> 30000  (still lucky)
new timeout-bearing helper ABOVE run()           loose ->  5000  (wrong literal)
                                                 scoped -> 30000 (correct)
```

That second row is the one that matters: it would report drift that does not exist, or agree with a literal nobody enforces. Now anchored on `execFileSync(exe, args, {` — `run()`'s own signature — with zero-or-several matches a failure rather than a reason to guess. That file has seven `execFileSync` calls, so "the first one" was never a stable identifier.

Full suite: 3924 passed, 1 failed (the pre-existing `js-yaml` temp-dir resolution failure, identical on `main`).

---

That is five rounds on this PR, and every one has been the same shape: a guard that stopped just short of the thing it guarded. Budget → canary → validation → sentinel → the guard's own regex. Each catch was correct and each fix was small, but the pattern is worth stating plainly for whoever reviews it: **this change keeps needing another layer to be trustworthy, which is itself evidence about the design.**

@santifer — the alternative is still on the table and I would not argue against it: drop the override entirely, take option (2) from #2906, and make the lock matrix cheaper so no script needs a special budget. That removes the concept rather than hardening it. Your call; I have no attachment to this shape.

LayerC0de:
**The `test (windows-latest)` red on `bb79c7a` is not this PR — it is #2777**, and I have written it up on that issue with the CI links.

```
7. concurrent adds do not lose items
  ✅ every concurrent add exited cleanly      <- all 30 exit 0
  ❌ all 30 concurrently queued items survive — kept=29 of 30   <- item-1 gone

✅ tracker-writer-lock-tests.mjs runs OK      <- 180s budget worked
```

Two things worth reading off that:

**This PR did the job it was opened for.** `tracker-writer-lock-tests.mjs` passed in the same run — the SIGTERM is gone. What replaced it is a *different* Windows failure in a file this branch does not touch: the PR changes `test-all.mjs` and `tests/helpers.mjs` only, and the `agent-inbox-tests.mjs` entry declares no budget, so it runs byte-identically to before.

**It also demonstrates the separation #2906 argued for.** A killed script and a short item count are now distinguishable failures rather than one indistinguishable "Windows flake", which is the whole reason the two were filed apart.

Re-running would very likely go green — it has passed several times on this branch already — but I have left it red rather than papering over it, since it is a genuine sample of the thing #2777 is chasing and re-running is exactly the reflex that cost two cycles on #2610 and #2682.

CI is otherwise green: ubuntu, macos, guard, CodeQL, GitGuardian all pass.

LayerC0de:
Re-triggered on a byte-identical tree (`git commit --amend --no-edit`, tree unchanged `31313cd`) — **all five checks green, `test (windows-latest)` included** (3m19s).

Same code, different outcome, which is the confirmation that the earlier red was #2777's flake rather than this change.

The sample is not lost by going green: the failing run and job IDs, the full case-7 output and the missing `item-1` are recorded on #2777, which is where they are useful. What I wanted to avoid was re-running *instead of* reading it.

Scott-Emberson (approved):
Code owner for `tests/`. The only owned change is the additive `DEFAULT_SCRIPT_TIMEOUT_MS` export in `helpers.mjs`; `run()` is byte-identical and its `timeout: 30000` is untouched, so no existing suite's timeout is weakened. The drift regex anchors to exactly one `execFileSync(exe, args, {...})` call, so the constant and the literal cannot silently diverge. Approving our area.

santifer:
Another one that teaches while it fixes, @LayerC0de: the outlier declared as an outlier (180s only for the 27-case contention matrix, with the four measured runs in the comment), so a NEW slow script still fails loudly instead of inheriting a budget sized for this one. The 75% early-warning fraction is the kind of headroom thinking the suite needed. Merged 🚀

Issue #2751, opened 2026-08-12
Day-zero title_filter comes from a template copy, not the CV — and `titles` can only add to it


WJiangH:
## Problem

No mode owns day-zero `title_filter.positive`. The CLI's answer is a file copy, and the one mode that reads the CV is architecturally downstream of that copy and can only add to it.

**What a new CLI user is told to do.** `doctor.mjs` on a missing `portals.yml`:

```
Run: cp templates/portals.example.yml portals.yml
Then customize with your target companies
```

That copy is not CV-aware. It hands the user 37 `title_filter.positive` keywords, 116 `tracked_companies` and 48 `search_queries` that describe the template author's market. For anyone outside it, most of that is noise the user now has to find and delete by hand — including five German further-education terms (`KI Trainer`, `Dozent`, `Weiterbildung`, …) that are wrong for the majority of users, and a company list weighted toward voice AI / contact-centre / no-code.

Concretely, on one real CV in a different ML subfield: **6 of the 37 template keywords survived curation, and 31 were deleted** — while the keywords that CV actually needed had to be typed in from scratch. `tracked_companies` went from 116 to 30 by the same manual process.

**`titles` cannot fix this, by its own design.** From `modes/titles.md`'s Error Handling:

> `portals.yml` missing, or `title_filter.positive` empty → offer to create it from `templates/portals.example.yml` first, then re-run this mode. (An empty positive list means the scanner matches everything — nothing to broaden.)

So `titles` is a *broadening* mode that requires the template copy to have already happened, and per its Confirm Gate it only ever proposes additions — it never removes. The loop is:

1. `cp` template → 37 of the author's keywords
2. `titles` → suggests additions from the CV
3. nothing ever removes step 1

The user ends up permanently searching the author's market plus their own.

**Meanwhile the CV-driven path exists, but only in the web assistant, and only as prompt text.** `web/src/app/api/assistant/route.ts` instructs the model to "infer their target roles + location FROM the CV" and emit a confirm-gated `setProfile`. `modes/titles.md` separately instructs a model to read `cv.md` and derive keywords. Two implementations of the same derivation, both living in prose, neither reachable from `doctor`, neither testable, and neither producing the same result twice.

So which keywords a user ends up searching with depends on whether they onboarded through the CLI or the web — off the same `cv.md`.

## Proposed solution

Two halves. The first is the reusable piece; the second is the entry point that makes it the default.

**1. A non-interactive, structured entry point for CV → keyword derivation.**

[...]

PR #2909, opened 2026-08-15
feat(titles): let the mode retire a keyword, not only add one

WJiangH:
Closes half of #2751 — the half @santifer called the blocker for everything else, split out as you suggested so it's reviewable on its own.

> **Make `titles` able to REPLACE, not only extend.** That's the architectural half you identified, and it's the blocker for everything else: as long as the mode can only add, the template's 31 wrong keywords survive every subsequent run.

**No new mode.** You said a second mode over the same input is how the two drift and start disagreeing about the user's target roles, and you're right — this is entirely inside `modes/titles.md`, which already reads the CV and already owns adjacent-title reasoning.

## The problem, on this repo's own numbers

`doctor` tells a new user to `cp templates/portals.example.yml portals.yml`. That hands them 37 keywords describing the market the tool was built for. Against one real CV, 31 of them have no claim in it — including `KI Engineer`, `KI Trainer`, `Dozent`, `Weiterbildung`. `titles` then grows that list on every run and never questions it, so the wrong 31 are permanent, and nothing downstream can tell *"the user chose these"* from *"the user never looked"*.

## What's added

**`## Retiring a keyword (HARD RULE)`** — deliberately narrow, because `portals.yml` is user-layer:

- **One ground only: `cv.md` does not support it.** The same evidence bar the additions are held to, applied in the other direction. "Looks unrelated" or "seems too broad" are not reasons — if you can't name the missing CV line, don't propose the removal.
- **Never a bulk replace, never silent.** One keyword at a time, each with its reason, in the same confirmed diff as the additions. Your flag was the design constraint here: *"never a silent regeneration on someone who spent an evening pruning those 116 companies down to 30"* — that sentence is in the mode file.
- **Two cases, told apart by whether the list still matches the shipped example.** Untouched → say plainly that these are the template's defaults, not theirs, and propose the CV-derived set. Edited → some of it *was* chosen and you can't tell which, so default to keeping anything uncertain. A keyword wrongly kept costs noise in the next scan; one wrongly removed costs a job the user never sees.
- **Never remove the last positive keyword.** Empty `positive` means the scanner matches every posting on every board — strictly worse than the wrong keyword the removal was meant to fix.

**The empty-list remedy is reversed**, per your third point. It pointed at the template first, which is exactly how someone else's market becomes a user's config. The way out of empty is the CV this mode already reads; `templates/portals.example.yml` is offered second, as *"or start from an example and edit it"*, for someone who'd rather begin from a shape than from their own history. Copying it stays a valid choice — just not the default.

**The confirm gate covers both halves in one diff, under separate headings.** A user deciding whether to widen their search is entitled to see what narrows in the same breath — and a yes to the additions is never read as a yes to the removals.

## Tests

Five assertions in `test-all.mjs` section 58, **each confirmed to bite when its rule is deleted**. Worth reporting one failure of my own: the first version of the removal assertion pinned the string `Retiring a keyword`, which also appears in two cross-references elsewhere in the file — so deleting the entire section left it green. Re-pinned on a phrase that exists only inside the section, and re-checked.

`node test-all.mjs` → **4000 passed / 0 failed**, verified under the merge boundary (today's `main` merged into a throwaway branch), not on the branch alone.

## What "replace" means here — neither overwrite nor merge

Answering @santifer's second check-point up front, in his framing:

> Full overwrite is easy to reason about and destroys hand-added keywords; a merge is kinder and reintroduces the drift problem.

**This is a third thing, and it's deliberately the least convenient one: per-keyword removal, on evidence, defaulting to keep.**

There is no operation in this mode that computes a new list and swaps it in. The mode never holds a "correct" list to overwrite with — it holds a set of *proposals*, each one a single keyword with a single stated reason, and the file changes only by the ones the user says yes to. Concretely:

|  | what it does to a hand-added keyword |
|---|---|
| Full overwrite | destroys it — it isn't in the derived set |
| Merge | keeps it, and keeps every wrong template keyword too (the drift) |
| **This** | **keeps it, unless `cv.md` positively fails to support it — and then only after the user sees that keyword and that reason and agrees** |

The asymmetry is the point, and it's stated in the mode file as the tie-break rule: *a keyword wrongly kept costs some noise in the next scan; a keyword wrongly removed costs a job the user never sees.* So uncertainty resolves toward keeping, always.

**On the partly-curated case specifically** — the one you named — the mode is told it cannot distinguish chosen keywords from copied ones, and must behave accordingly:

> **The list has been edited.** Then some of it was chosen, and you cannot tell which. Propose removals only for keywords with no CV support, list them individually with the reason, and default to keeping anything you are unsure about.

The wholesale case is allowed exactly once, and only where nothing was chosen: when `title_filter.positive` still matches `templates/portals.example.yml` byte for byte. Even there it is proposals in a diff, not a swap — the difference is only that the mode may say plainly *"these 37 keywords are the template's defaults, not yours"* rather than treating them as considered choices.

**And on your first check-point** — user-layer safety — the sentence from the issue is in the mode file, not paraphrased:

> There is no bulk "replace" and no silent rewrite — someone who spent an evening pruning 116 companies down to 30 must never find that work undone by a mode they ran to widen their search.

with three hard "never"s beneath it: never a keyword the user added in this or a previous run, never anything in `title_filter.negative`, never the last remaining positive keyword. That last one is pinned by its own test — emptying `positive` makes the scanner match every posting on every board, which is worse than whatever wrong keyword the removal was trying to fix.

## Not in this PR

The `doctor.mjs` remedy — you called it a small follow-up once there's somewhere to send people, and this is that somewhere. Happy to put it up once this lands, or to fold it in here if you'd rather review one thing.

## Summary by CodeRabbit

- **New Features**
  - Improved title keyword management with support for confirmed additions and removals.
  - Unsupported keywords can be retired while protecting session-added, negative, and final remaining keywords.
  - Missing or empty positive keyword filters can be recovered automatically from `cv.md`, with an example template available as an alternative.
  - Combined keyword additions and removals are handled together without unnecessary regeneration.
  - Confirmed removals are validated individually, while curated portal settings are preserved.


santifer:
Hours between me suggesting the split and you shipping it, @WJiangH — and you took the harder half.

That it's `modes/titles.md` + `test-all.mjs` and **no new mode** is the part that matters. The easy version of this feature is a second mode that owns `title_filter`, and it would have looked cleaner in the diff while quietly creating two things that both believe they know what your target roles are. Extending the mode that already reads the CV is more constrained and less satisfying to write, and it's right.

Two things I'll be checking when it's reviewed, said now so there are no surprises:

1. **`portals.yml` is user-layer.** Replace has to be an explicit, reviewed action — never a silent regeneration over a config someone spent an evening pruning from 116 companies to 30. I flagged this on the issue and I'll hold the PR to it, because it's the one way this feature can do real harm.
2. **What "replace" means for a list the user has partly curated.** Full overwrite is easy to reason about and destroys hand-added keywords; a merge is kinder and reintroduces the drift problem. Whichever you chose, the PR body naming it explicitly will save a round trip.

Not merged tonight — per-session merge ceiling, and I've already merged two of yours in this run (#2692 plus the earlier one). Near the top next cycle.

For the record: this is the second time today someone has turned my direction into a PR faster than I expected, and both times the split-it-first advice made the result reviewable. Worth me remembering that "which half would you ship alone?" is a better question than a full review of the whole idea.

WJiangH:
Both check-points answered in the PR body rather than here, so the reviewer finds them without reading the thread — new section, **"What 'replace' means here — neither overwrite nor merge"**.

Short version, in your framing: it's a third option, and deliberately the least convenient one. **Per-keyword removal, on evidence, defaulting to keep.** There is no operation in this mode that computes a new list and swaps it in — it never holds a "correct" list to overwrite with, only proposals, each a single keyword with a single stated reason, and the file changes by the ones the user says yes to.

|  | what it does to a hand-added keyword |
|---|---|
| Full overwrite | destroys it |
| Merge | keeps it, and every wrong template keyword too |
| **This** | **keeps it, unless `cv.md` positively fails to support it — and then only after the user sees that keyword and that reason and agrees** |

The tie-break is in the mode file as a sentence, because it's the thing a model has to internalise rather than look up: *a keyword wrongly kept costs some noise in the next scan; a keyword wrongly removed costs a job the user never sees.* Uncertainty resolves toward keeping, always.

For the partly-curated list specifically, the mode is told it **cannot** tell chosen from copied and must act accordingly — propose removals only for keywords with no CV support, list them individually, keep anything uncertain. The wholesale case is permitted exactly once, where nothing was chosen: `title_filter.positive` still byte-identical to `templates/portals.example.yml`. Even there it's proposals in a diff, not a swap; the only difference is the mode may say plainly *"these 37 are the template's defaults, not yours"* instead of treating them as considered choices.

On your first point — the sentence from the issue is in the mode file verbatim rather than paraphrased, with three hard "never"s under it, and the last-keyword one has its own test. Emptying `positive` makes the scanner match every posting on every board, which is worse than whatever wrong keyword the removal was meant to fix.

No rush on the merge ceiling — the queue moving at all is the change from where this was two days ago.

WJiangH:
**`16355be` — I ran this mode against a real curated `portals.yml` instead of reasoning about it, and it would have deleted three quarters of a working filter.**

The removal ground read *"`cv.md` does not support it"*. Read as a string test — which is the reading a model reaches for, and which nothing in the section ruled out — here is what that means on one real setup:

```
49 curated keywords
39 do not appear anywhere in cv.md as a string
```

and the 39 include `AI Engineer`, `Deep Learning`, `Research Engineer`, `Compiler`, `MLIR`, `CUDA`. Every one of them is a legitimate target for a CV whose summary reads *"LLM inference optimization and on-device / edge AI deployment"* and whose bullets describe *"kernel- and graph-level optimizations"*. The mode would have proposed deleting them one at a time, each correctly formatted, each with a reason, each through the confirm gate — the safety rails all working, on a set that should never have been proposed.

What makes it the interesting kind of bug: the **additions** already run on the right rule. Suggesting titles the CV doesn't literally contain is the entire point of the mode — the market's word for a capability is almost never the CV's. I wrote the removal direction without carrying that over, and every safeguard I put around removals is orthogonal to it. A gate that reliably confirms the wrong decision is not a gate.

Now stated, with that measurement as the reason rather than an assertion, and reduced to the question to ask: **is there work in `cv.md` that would make this person a plausible candidate for a posting carrying this keyword?** Yes → keep it, whatever words the CV used. Only a no — no line, no adjacent capability, nothing the keyword could be reaching for — makes it a candidate, and it still goes through the diff with its reason. Pinned by a test, confirmed to bite.

**One limitation I'm reporting rather than quietly fixing**, since it's a design question and yours to answer.

The rule says *not "it seems too broad"* — deliberately, so breadth can't become a licence to prune. But that means the mode cannot retire a keyword that is genuinely CV-supported and still ruins the filter. On the same real setup, measured against 510 real postings:

```
Design Engineer   74 hits   e.g. "Data Center Design Engineer, Electrical - Industrial Compute"
Architecture      17 hits   e.g. "Strategic Deals & Commercial Architecture Lead"
```

Both are supported — this candidate does hardware-adjacent work and does architecture — and both are mostly noise, and each admitted posting is a real evaluation someone pays for. I did **not** add breadth as a second removal ground: it needs a threshold, a corpus to measure against, and it's the ground most likely to be abused into "prune anything I don't like". If you want it, I'd rather do it as its own change with the measurement wired in than as a sentence here.

`node test-all.mjs` → 4001 passed / 0 failed, under the merge boundary.

WJiangH:
**`d456627`** — CodeRabbit's three, and the first one was half right in a way worth writing down.

**The Major: ownership is unknowable, and I was reading silence as an answer.**

The finding is that `portals.yml` stores the keyword array and nothing about who put a keyword there, so the mode cannot tell an untouched template copy from a user-approved list with the same 37 entries. Its proposed fix — persist provenance — is a schema change I don't think this PR should make. But the underlying observation lands on two specific sentences of mine:

1. The template branch said, as fact, *"these 37 keywords are the template's defaults, not yours"*. Byte-identity with the shipped example is strong evidence — `doctor` copies it, so an untouched install lands exactly there — but it is an **inference, and I had written it as a record**. If it is wrong, the message actively nudges the user toward accepting removals of keywords they did choose. It now states the inference and invites the correction, and one correction drops the rest of the session to the edited-list branch.

2. `Never remove: a keyword the user added during this session **or a previous run**` was **unenforceable for the previous-run half.** There is no record to check, so "did the user add this?" resolves to "I didn't see it happen" — and unknowable silently becomes false, exactly where the mode is least sure. It now names the gap and points at the edited-list default-to-keep as the answer:

> `portals.yml` records no provenance, so "the user added this in a previous run" is not a fact you can check — treat it as unknowable rather than as false. […] never reason that a keyword must be the template's merely because you did not watch it being added.

That's the same red line you flagged, arrived at from the other direction: the way this feature does real harm is not a bulk overwrite (there isn't one) but a *confident sentence about provenance the file cannot support*.

**The two Minors, both real:**

- An absent `title_filter` key and an empty `positive` list are the same thing to both scanners — everything matches — so error handling now treats them the same instead of naming only the empty case. And since the mode was already told to write `title_filter`, it now says explicitly that the rest of `portals.yml` (the company list above all) must survive the write unchanged.
- The test for the last-keyword protection was `titlesFlat.includes('the last remaining positive keyword')`. **That passes on any sentence containing the phrase, including one permitting the removal.** The checks are now scoped to the `Never remove:` block itself, and the per-removal reason and the provenance statement are asserted too.

I verified that by mutation rather than by reading it: moving the phrase out of the block into *"Removing the last remaining positive keyword is acceptable when the CV supports none of them"* leaves the document mentioning it, passes the old assertion, and fails the new one. Same for the per-removal reason and the provenance line — each fails on its own when its sentence is removed.

**Verification.** Merged onto today's `main` in a throwaway branch and ran there rather than quoting the branch-local number: **4329 passed, 0 failed** (the 2 warnings are the repo's existing scraper-truncation ones, unchanged). Branch alone: 4206/0.

Nothing here changes what the mode does — it changes what it is allowed to claim while doing it.

santifer:
This is how a removal feature should be written, @WJiangH: one ground only (cv.md doesn't support the capability), the 39-of-49 measurement that kills the naive string test, default-to-keep when provenance is unknowable, and the never-empty-the-list floor. The line about the user who pruned 116 companies down to 30 is the whole spirit of the user layer in one sentence. Merged 🚀

PR #2912, opened 2026-08-15
docs: add a Windows page for the platform's sharp edges

sdkkds:
Adds `docs/WINDOWS.md`, and links it from the Windows entry in the README (which pointed at FAQ and SETUP — `docs/SETUP.md` has no Windows content today, so there was nowhere for this to live).

## Why a page rather than a FAQ entry

Windows failures in this repo cluster in one place — the parts that shell out to `bash` — and none of them surface as a platform problem:

- A batch that "ran" and produced nothing is bare `bash` resolving to `C:\WINDOWS\System32\bash.exe`, the **WSL launcher**, where the Windows `node` is not on `PATH`. `batch-runner.sh` dies with `node: command not found` and exit 127, and because the output is captured you see an empty result rather than an error.
- `syntax error near unexpected token $'{\r'` at line 2 of `batch-runner.sh` is CRLF.
- A scheduled run that failed and one that hung look identical on disk, because Task Scheduler has no console and discards both streams.

Each of those sends someone into debugging career-ops. The page names the symptom first so the search lands on it.

## Contents

WSL-vs-Git-Bash on `PATH` · the Program Files vs Scoop install layouts and the `cygpath`/`bash` same-install rule · line endings, including the one-time renormalization for clones predating the v1.26.0 `.gitattributes` · running the suite · Playwright for PDFs · skill entrypoints (links to FAQ #1, does not restate it) · Task Scheduler (links to AUTOMATION.md) · path conventions.

## Verification

Everything asserted about this repo was re-checked against `22cbe88` immediately before opening, not against the tree I originally drafted on:

- `docs/WINDOWS.md` does not already exist; `docs/SETUP.md` has no Windows content; FAQ #1 covers the symlink case; AUTOMATION.md has a `### Windows — Task Scheduler` section.
- §2's claim that the suite searches both install layouts is #2366 as merged (`bff961c`) — `tests/helpers.mjs` checks Program Files plus `%SCOOP%` and `%USERPROFILE%\scoop`, and names the fallback shell on failure.
- §3 is written against the `.gitattributes` added in `80d104f` and released in v1.26.0, so the advice is "renormalize if you cloned before that", not "add one yourself".
- Every file and command the page names exists and runs: `node --test tests/cli-flags.test.mjs` passes, `doctor.mjs`, `generate-pdf.mjs`, `batch/batch-runner.sh`, `tests/helpers.mjs` all present.
- `node test-all.mjs` on this branch: **3935 passed, 0 failed**. (The one warning is `cv-sync-check.mjs exited with error (expected without user data)` — a checkout with no user data, unrelated.)

## Note on provenance

This is built from real failures on a Windows 11 + Scoop machine running the nightly batch, not from a clean-room test matrix — which is why the last line invites issues. The Scoop layout in particular is the one I would most expect to be incomplete for other people's setups.

Happy to cut anything that duplicates docs I have missed.

## Summary by CodeRabbit

- **Documentation**
  - Added comprehensive Windows setup and troubleshooting guidance.
  - Documented Git Bash discovery, WSL conflicts, path conversion, line endings, testing, PDF browser installation, skill entrypoints, Task Scheduler configuration, and Windows path conventions.
  - Updated the Windows FAQ to link to the new platform-specific documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -453,7 +453,7 @@ Because an `ANTHROPIC_API_KEY` in your environment takes precedence over your lo
 career-ops runs on any major AI coding CLI — Claude Code, Codex, Gemini / Antigravity, OpenCode, Grok, Qwen and more — through the open Agent Skill Standard, so it is never locked to a single vendor. Use the CLI you already have.
 
 **How do I install career-ops on Windows?**
-career-ops runs on Windows. If skills fail to load with a symlink error during install, the fix is in [docs/FAQ.md](docs/FAQ.md). Full steps are in [docs/SETUP.md](docs/SETUP.md).
+career-ops runs on Windows. Platform-specific setup and the known sharp edges — Git Bash discovery, line endings, Task Scheduler — are in [docs/WINDOWS.md](docs/WINDOWS.md). If skills fail to load with a symlink error during install, the fix is in [docs/FAQ.md](docs/FAQ.md). Full steps are in [docs/SETUP.md](docs/SETUP.md).
 
 **Does career-ops auto-apply to jobs for me?**
 No. career-ops is a filter, not a spray-and-pray auto-applier. The AI evaluates, ranks and drafts; you review and decide. It never submits, sends, or clicks anything — you always have the final call. That human-in-the-loop design is the whole point.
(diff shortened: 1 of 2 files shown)


santifer:
Verified the load-bearing claims against the repo before merging and every one holds: the .gitattributes LF policy, the Scoop/Program Files split in tests/helpers.mjs, and the FAQ symlink route. The WSL-launcher-on-PATH diagnosis alone will save people hours of "career-ops is broken" that was never career-ops. Merged! 🚀

PR #2916, opened 2026-08-15
fix(build-cv-latex): numeric year/date scalars are still dropped — #2641 fixed only the HTML mirror

### LINKED ISSUE #2915: build-cv-latex: numeric year/date scalars are still dropped — #2641 fixed only the HTML mirror
### What

#2641 removed `if (typeof text !== 'string') return '';` from `escapeHtml`, because a machine-authored payload with `dates: 2024` (a JSON number, not `"2024"`) rendered an empty field while the section stayed `present`.

`escapeLatex` in `lib/latex-escape.mjs` still has that guard, and `build-cv-latex.mjs` runs **every** value through it.

### Measured, end-to-end through the real builder

Payload with four numeric scalars, using the field names this builder reads:

```
                     BEFORE   AFTER
experience dates 2024   0        1
education  dates 2019   0        1
award      year  2025   0        1
project    dates 2023   0        1
```

(counts = occurrences in the generated `.tex`)

The shape it leaves behind:

```latex
\resumeSubheading
  {Corp}{}          % <- 2024 should be here
```

And the builder reports success throughout:

```json
{ ..., "valid": true }     exit 0
```

So a LaTeX CV ships with employment dates, education dates, project dates and award years missing, and nothing anywhere says so — the same failure #2641 describes, on the other builder.

This pair has needed to move together before: #2588 fixed the `$`-pattern replacement in **both** `build-cv-html` and `build-cv-latex`.

### Fix

Mirror #2641's rule exactly: blank only truly absent/structural values (`null`, `undefined`, objects) and coerce scalars with `String()`.

### Deliberately not changed

`sanitizeUrl()` in the same file keeps its `typeof url !== 'string' -> ''` guard. A URL is inherently a string; coercing an object there would produce `"[object Object]"` and then run it through scheme-matching, which is worse than dropping it.

PR incoming.

abankar1:
Closes #2915.

#2641 removed `if (typeof text !== 'string') return '';` from `escapeHtml`, because a payload with `dates: 2024` (a JSON number, not `"2024"`) rendered an empty field while the section stayed `present`.

`escapeLatex` kept that guard, and `build-cv-latex.mjs` runs **every** value through it.

## Measured, end-to-end through the real builder

```
                        BEFORE   AFTER
experience dates 2024      0        1
education  dates 2019      0        1
award      year  2025      0        1
project    dates 2023      0        1
```

(occurrences in the generated `.tex`)

The shape it leaves:

```latex
\resumeSubheading
  {Corp}{}          % 2024 should be here
```

…while the builder reports `"valid": true` and exits 0. So a LaTeX CV ships without employment dates, education dates, project dates or award years, and nothing says so — the same failure #2641 describes, on the other builder.

These two have needed to move together before: **#2588 fixed the `$`-pattern replacement in both builders.**

## Fix

Mirror #2641's rule exactly — blank only `null`/`undefined`/objects, coerce scalars with `String()`.

## Two things I checked rather than assumed

**`sanitizeUrl()` in the same file keeps its string guard, deliberately.** A URL is inherently a string; coercing an object would produce `"[object Object]"` and then run it through scheme-matching, which is worse than dropping it.

**The two builders read different field names in places** — LaTeX education uses `dates` where HTML uses `year`. My first repro used the HTML names and showed the education year "missing" for the wrong reason. I re-ran against the fields *this* builder actually reads before claiming anything, and the test uses those. Whether that schema divergence is itself a problem is a separate question I haven't touched.

## Tests

`tests/cv-latex-numeric-scalars.test.mjs`, mirroring `tests/cv-numeric-scalars.test.mjs`. Both assertions fail on `main`:

```
❌ numeric scalars dropped from the .tex: experience dates (2024), education dates (2019), award year (2025), project dates (2023)
❌ experience subheading lost its date: {Corp}{}
```

`node test-all.mjs --quick` → **3937 passed, 0 failed** · `validate-system-paths-coverage` → OK.

2 commits (fix + test).

## Summary by CodeRabbit

* **Bug Fixes**
  * Numeric and other scalar values now render correctly in generated LaTeX.
  * URL and text output handle scalar values consistently.
  * Empty values and objects continue to produce blank output where appropriate.

* **Tests**
  * Added coverage for numeric dates and years across experience, education, awards, and projects.

### DIFF
--- a/lib/latex-escape.mjs
+++ b/lib/latex-escape.mjs
@@ -10,10 +10,18 @@
  * @returns {string}
  */
 export function escapeLatex(text, mode = 'text') {
-  if (typeof text !== 'string') return '';
-  if (mode === 'url') return text;
+  // Blank out only truly absent/structural values, and coerce scalars — the
+  // same rule #2641 applied to escapeHtml, which this function mirrors.
+  // `typeof text !== 'string' -> ''` silently dropped NUMBERS: a payload with
+  // `year: 2019` or `dates: 2024` (JSON numbers, not strings) rendered an empty
+  // \resumeSubheading date field while the section stayed present, so the .tex
+  // shipped without employment dates or graduation years — and the builder
+  // still reported "valid": true and exited 0.
+  if (text === null || text === undefined || typeof text === 'object') return '';
+  const value = String(text);
+  if (mode === 'url') return value;
   const out = [];
-  for (const ch of text) {
+  for (const ch of value) {
     switch (ch) {
       case '\\': out.push('\\textbackslash{}'); break;
       case '{': case '}': out.push('\\' + ch); break;
--- a/tests/cv-latex-numeric-scalars.test.mjs
+++ b/tests/cv-latex-numeric-scalars.test.mjs
@@ -1,0 +1,66 @@
+// tests/cv-latex-numeric-scalars.test.mjs — the LaTeX mirror of #2641.
+//
+// #2641 removed `typeof text !== 'string' -> ''` from escapeHtml because a
+// machine-authored payload with `dates: 2024` (a JSON number, not "2024")
+// rendered an empty field while the section stayed present. escapeLatex — the
+// function build-cv-latex.mjs runs every value through — kept the same guard,
+// so the .tex shipped without employment dates, education dates, project dates
+// or award years, and the builder still reported "valid": true and exited 0.
+//
+// End-to-end through the real builder and the shipped template, because
+// build-cv-latex.mjs exports nothing. Field names are the ones THIS builder
+// reads (education uses `dates`, awards use `year`) — they differ from the HTML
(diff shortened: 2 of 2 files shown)


Scott-Emberson (approved):
Code owner for `tests/`. `cv-latex-numeric-scalars.test.mjs` feeds real JSON numbers through `build-cv-latex.mjs` and asserts the value is present (`/\{Corp\}\{2024\}/`), so it fails if the fix is reverted, and the template carries no 4-digit years to give a false pass. Mirrors the #2641 HTML sibling. LGTM.

santifer:
Finding that #2641 fixed one of two copies of the same defect is exactly the sibling-drift class this repo keeps hunting: the LaTeX path now coerces numeric scalars the same way the HTML mirror does, and the test pins both. Third solid one today, @abankar1. Merged! 🚀

PR #2918, opened 2026-08-15
fix(web): 27 states.yml aliases resolve to no stage at all — plus the fold that keeps Turkish broken anyway

### LINKED ISSUE #2917: web: 27 states.yml aliases resolve to no stage at all — #2249's drift, one layer up
## Summary

`STATUS_ALIAS` in `web/src/lib/format.ts` is a hand-maintained copy of the alias lists in `templates/states.yml`. **27 of the aliases in states.yml are missing from it**, and each one resolves to no canonical stage at all in the dashboard.

This is #2249 recurring. That issue reported the same drift for one state (`Hired`) and was fixed by hand-adding that state's five aliases. The *class* — a duplicated map with nothing asserting it stays in sync — was left in place, so it drifted again as soon as new aliases landed in states.yml (#2705).

Two of the 27 (`geo blocker`, `geo_blocker` → `SKIP`) are not from #2705 at all; they have been missing since before that, which is the point: nothing in CI can notice.

## Why it breaks

`canonStatus()` ends in `return STATUS_ALIAS[k] ?? s.toUpperCase()`. An alias that isn't in the map isn't rejected — it's passed through as raw uppercased text. Every consumer then does a substring test against the canonical names, and raw text matches none of them:

```
"mülakat"        -> "MÜLAKAT"        -> matches tabs: NONE   (want INTERVIEW)
"teklif"         -> "TEKLIF"         -> matches tabs: NONE   (want OFFER)
"işe alındı"     -> "IŞE ALINDI"     -> matches tabs: NONE   (want HIRED)
"geo blocker"    -> "GEO BLOCKER"    -> matches tabs: NONE   (want SKIP)
"reddedildi"     -> "REDDEDILDI"     -> matches tabs: NONE   (want REJECTED)
```

So for a row whose status is any of the 27, on the dashboard:

1. **It disappears from the tracker.** `pipeline-view.tsx:96` filters with `canonStatus(r.status).includes(tab)` — the row shows under `ALL` and under no other tab.
2. **It is missing from every tab badge** (`pipeline-view.tsx:151`) and **every stage bar** (`analytics/page.tsx:25`).
3. **It re-opens #2410.** The analytics headline tiles are achievement counters whose zero-state shows a coaching nudge:

```js
cumulativeTiles([canonStatus("Mülakat")])  // => { interviews: 0, offers: 0 }
cumulativeTiles([canonStatus("Teklif")])   // => { interviews: 0, offers: 0 }
```

A user actively interviewing is told *"Interviews follow replies — keep follow-ups warm."* A user holding an offer reads zero offers. #2410 fixed exactly this nudge for the cumulative-math case; the alias gap re-opens it through a different door.

## Reproduce

No server needed — this parses the real map out of `format.ts` and folds it against the real `states.yml`:

```js
import { readFileSync } from "node:fs";
import yaml from "js-yaml";

const ts = readFileSync("web/src/lib/format.ts", "utf8");
const block = ts.slice(ts.indexOf("const STATUS_ALIAS"), ts.indexOf("export const CANONICAL_STATES"));
const web = new Map();
for (const m of block.matchAll(/^\s*"?([^":\n]+?)"?:\s*"([A-Z]+)"/gm)) web.set(m[1].trim(), m[2]);

const canonStatus = s => {
  const k = s.trim().toLowerCase();
  if (k === "" || k === "—" || k === "-") return "DISCARDED";
  return web.get(k) ?? s.toUpperCase();
};

const states = yaml.load(readFileSync("templates/states.yml", "utf8")).states;
for (const st of states)
  for (const a of st.aliases ?? [])
    if (!canonStatus(a).includes(st.label.toUpperCase()))
      console.log(JSON.stringify(a), "->", JSON.stringify(canonStatus(a)), " want", st.label.toUpperCase());
```

Prints 27 rows: the full Turkish set across all 9 states, plus `geo blocker` / `geo_blocker`.

## The fix I'd propose

The instance fix (paste 27 more entries) is what #2249 did, and it is why we are here again. Two parts instead:

1. **Move the map to `web/src/lib/status-alias.mjs`** and have `format.ts` import it. This follows the pattern already documented in `funnel-tiles.mjs` / `clean-chips.mjs` / `stream-parse.mjs`: pure JS, no TS types, so it is importable by a `node --test` unit test. `format.ts` is explicitly node-free (it is used by client components), so it cannot read `states.yml` at runtime — the map has to stay a literal.
2. **Add a drift test** that loads `templates/states.yml` and asserts every alias resolves to its own state. There is direct precedent for a web test reading up out of `web/`: `tests/lib/clean-chips.test.mjs:101` reads `../../../templates/portals.example.yml`, and `js-yaml` is already a web dependency.

That is the "copy + CI assertion" mechanism rather than a bare copy: the literal stays (it must, for the client bundle), but a new alias in `states.yml` now fails web CI instead of silently vanishing from the dashboard. Happy to adjust if you'd rather generate the module at build time instead.

abankar1:
Fixes #2917.

`STATUS_ALIAS` in `web/src/lib/format.ts` is a hand-maintained copy of the alias lists in `templates/states.yml`. **27 of the aliases in states.yml resolved to no canonical stage at all** in the dashboard.

This is #2249 recurring. That issue reported the same drift for one state (`Hired`) and was fixed by hand-adding that state's aliases. The *class* — a duplicated map with nothing asserting it stays in sync — was left in place, so it drifted again as soon as #2705 landed. Two of the 27 (`geo blocker`, `geo_blocker`) predate #2705 entirely, which is the point: nothing in CI could notice.

## The bug

`canonStatus` ends in `STATUS_ALIAS[k] ?? s.toUpperCase()`. A missing alias is not rejected — it is passed through as raw uppercased text, and every consumer substring-tests it against the canonical names, which it matches none of:

```
"mülakat"     -> "MÜLAKAT"     matches tabs: NONE   (want INTERVIEW)
"teklif"      -> "TEKLIF"      matches tabs: NONE   (want OFFER)
"geo blocker" -> "GEO BLOCKER" matches tabs: NONE   (want SKIP)
```

For a row with any of them:

1. It shows under `ALL` and under **no other tracker tab** (`pipeline-view.tsx:96`).
2. It is missing from every tab badge (`:151`) and every stage bar (`analytics/page.tsx:25`).
3. **It re-opens #2410.** The analytics headline tiles are achievement counters whose zero-state is a coaching nudge:

```js
cumulativeTiles([canonStatus("Mülakat")])  // => { interviews: 0, offers: 0 }
```

Someone actively interviewing is told *"Interviews follow replies — keep follow-ups warm."* #2410 fixed that nudge for the cumulative-math case; the alias gap re-opened it through a different door.

## Second half of the bug, which the obvious fix does not reach

Completing the table was necessary but not sufficient. states.yml spells its aliases lowercase, so a test that feeds them verbatim cannot see a folding bug — but a real tracker row is capitalised:

```
canonStatus("İşe alındı")  ->  "İŞE ALINDI"   matches tabs: NONE
```

JS lowercases `İ` (U+0130) to `i` + U+0307 COMBINING DOT ABOVE, so the key becomes `i̇şe alındı` and misses the `işe alındı` entry however complete the map is. Every Turkish status written the way a user actually writes it still vanished.

Fixed with the same fold as `foldStatusInput()` in `tracker-utils.mjs` (#2705): strip bold → trim → NFKC → lowercase → drop U+0307.

**NFKC and not NFD, deliberately.** NFD decomposes precomposed letters too, so the same strip would reach the dots of `ż` / `ė` / `ġ` and collapse Żubr→Zubr, Ėmė→Eme, Ġenerali→Generali. There is a test pinning those three unchanged, so a future switch to NFD fails loudly.

Stripping bold also means `"**Oferta**"` now resolves to `OFFER` instead of matching nothing. AGENTS.md forbids bold in the status field, which is precisely why it turns up there.

## Why the map is still a literal

`format.ts` is node-free on purpose — client components import it — so it cannot read `states.yml` at runtime. The map has to ship in the bundle as data.

What makes that copy safe is the guard, not the fill: `tests/lib/status-alias.test.mjs` loads the real `templates/states.yml` and asserts every alias resolves to its own stage. Adding an alias to states.yml now **fails web CI** until it is added here too. That is the "copy + CI assertion" mechanism rather than a bare copy.

Reading up out of `web/` follows `tests/lib/clean-chips.test.mjs:101`, which loads `../../../templates/portals.example.yml` the same way; `js-yaml` is already a web dependency.

## Commits

1. **`refactor`** — move `STATUS_ALIAS` + `canonStatus` into `web/src/lib/status-alias.mjs`, following the `funnel-tiles.mjs` / `clean-chips.mjs` / `stream-parse.mjs` pattern (pure JS so a `node --test` test can import it; web has no TS runner). `format.ts` re-exports `canonStatus`, so no import site changes. No behaviour change.
2. **`fix`** — add the 29 missing aliases + the drift test.
3. **`fix`** — fold the lookup key for the Turkish dotted capital.

## Verification

Both fixes were checked for discrimination rather than assumed:

- Against the **previous map**, the drift test fails listing exactly the 27, while its sibling assertions still pass.
- Reverting **only the fold** (keeping the completed map) fails the capitalised and bolded tests while the other six pass.

`web`: 257 pass / 0 fail. `node test-all.mjs --quick`: 3935 pass / 0 fail (the 2 warnings are pre-existing and environmental — no user data, no Playwright browser).

The suite also caught a real mistake in my first draft: I default-imported `js-yaml`, and test-all.mjs rejects that. Fixed to `import * as yaml`.

## Compatibility

The 5 aliases web carries that states.yml does not list (`evaluado`, `respondida`, `contestada`, `duplicado`, `repost`) are kept and pinned by their own test — the sync is one-directional, states.yml is a floor and not a cap, so this cannot regress an existing tracker.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dashboard status recognition for localized, tracker-specific, bolded, blank, and dash-like status values.
  * Preserved correct capitalization and Unicode handling when normalizing statuses.

* **Tests**
  * Added coverage to verify status aliases remain synchronized with configured workflow stages and canonical labels.

### DIFF
--- a/web/src/lib/format.ts
+++ b/web/src/lib/format.ts
@@ -3,40 +3,12 @@
 // core: normalize-statuses.mjs (aliases) + the Go TUI dashboard (score/status
 // colours = the current state-of-the-art).
 
-// Spanish + legacy aliases → canonical English tokens (normalize-statuses.mjs).
-const STATUS_ALIAS: Record<string, string> = {
-  evaluada: "EVALUATED",
-  evaluado: "EVALUATED",
-  condicional: "EVALUATED",
-  hold: "EVALUATED",
-  evaluar: "EVALUATED",
-  verificar: "EVALUATED",
-  aplicada: "APPLIED",
-  aplicado: "APPLIED",
-  enviada: "APPLIED",
-  sent: "APPLIED",
-  respondida: "RESPONDED",
-  respondido: "RESPONDED",
-  contestada: "RESPONDED",
-  entrevista: "INTERVIEW",
-  oferta: "OFFER",
-  rechazada: "REJECTED",
-  rechazado: "REJECTED",
-  descartada: "DISCARDED",
-  descartado: "DISCARDED",
-  cerrada: "DISCARDED",
-  cancelada: "DISCARDED",
-  duplicado: "DISCARDED",
-  repost: "DISCARDED",
-  monitor: "SKIP",
-  no_aplicar: "SKIP",
-  "no aplicar": "SKIP",
-  // Hired — terminal success (offer accepted), added to states.yml in #2050.
-  contratado: "HIRED",
-  contratada: "HIRED",
-  accepted: "HIRED",
-  accept: "HIRED",
-};
+// Alias → canonical stage. Lives in status-alias.mjs so a `node --test` unit
+// test can load it and assert it still covers templates/states.yml (#2917);
+// a hand-maintained copy is what drifted in #2249 and again after #2705.
+import { canonStatus } from "@/lib/status-alias.mjs";
+
+export { canonStatus };
 
 export const CANONICAL_STATES = [
   "Evaluated",
@@ -49,12 +21,6 @@ export const CANONICAL_STATES = [
   "Discarded",
   "SKIP",
 ] as const;
-
-export function canonStatus(s: string): string {
-  const k = s.trim().toLowerCase();
-  if (k === "" || k === "—" || k === "-") return "DISCARDED";
-  return STATUS_ALIAS[k] ?? s.toUpperCase();
-}
 
 /** Status dot colour, mirroring the Go TUI: green hired/interview/offer, sky
  *  applied/responded, red skip/rejected, gray discarded, neutral evaluated. */
--- a/web/src/lib/status-alias.mjs
+++ b/web/src/lib/status-alias.mjs
@@ -1,0 +1,63 @@
+// Status alias → canonical stage, mirroring `templates/states.yml`.
+//
+// Pure JS (no TS types) so it can be imported both by format.ts and by a
+// `node --test` unit test, matching funnel-tiles.mjs / clean-chips.mjs /
+// stream-parse.mjs.
+//
+// This map is a LITERAL rather than a read of states.yml because format.ts is
+// node-free on purpose — it is imported by client components, so it cannot
+// touch fs at runtime. The map has to ship in the bundle as data.
+//
+// Moved here verbatim; the alias list is unchanged by this commit.
+
+/** @type {Record<string, string>} */
+export const STATUS_ALIAS = {
+  evaluada: "EVALUATED",
+  evaluado: "EVALUATED",
+  condicional: "EVALUATED",
+  hold: "EVALUATED",
+  evaluar: "EVALUATED",
+  verificar: "EVALUATED",
+  aplicada: "APPLIED",
+  aplicado: "APPLIED",
+  enviada: "APPLIED",
+  sent: "APPLIED",
+  respondida: "RESPONDED",
+  respondido: "RESPONDED",
+  contestada: "RESPONDED",
+  entrevista: "INTERVIEW",
+  oferta: "OFFER",
+  rechazada: "REJECTED",
+  rechazado: "REJECTED",
+  descartada: "DISCARDED",
+  descartado: "DISCARDED",
+  cerrada: "DISCARDED",
+  cancelada: "DISCARDED",
+  duplicado: "DISCARDED",
+  repost: "DISCARDED",
+  monitor: "SKIP",
+  no_aplicar: "SKIP",
+  "no aplicar": "SKIP",
+  // Hired — terminal success (offer accepted), added to states.yml in #2050.
+  contratado: "HIRED",
+  contratada: "HIRED",
+  accepted: "HIRED",
+  accept: "HIRED",
+};
+
+/**
+ * Normalize a raw tracker status to a canonical stage token.
+ *
+ * Unknown input is passed through uppercased rather than rejected, so a status
+ * this map has never seen still renders as itself. Consumers substring-test the
+ * result, which is why an alias missing from the map above resolves to no stage
+ * at all — see tests/lib/status-alias.test.mjs.
+ *
+ * @param {string} s - Raw status text from the tracker.
+ * @returns {string} Canonical stage token, or the uppercased input.
+ */
+export function canonStatus(s) {
+  const k = String(s ?? "").trim().toLowerCase();
+  if (k === "" || k === "—" || k === "-") return "DISCARDED";
+  return STATUS_ALIAS[k] ?? String(s ?? "").toUpperCase();
+}
--- a/web/src/lib/status-alias.mjs
+++ b/web/src/lib/status-alias.mjs
@@ -89,6 +89,34 @@ export const STATUS_ALIAS = {
 };
 
 /**
+ * Fold raw status text to a lookup key.
+ *
+ * Mirrors foldStatusInput() in tracker-utils.mjs (#2705). The `\u0307` strip is
+ * what makes a capitalised Turkish status resolve: JS lowercases `İ` to `i` +
+ * U+0307 COMBINING DOT ABOVE, so "İşe alındı" keys as "i̇şe alındı" and misses
+ * the "işe alındı" entry. Real tracker rows are capitalised, so without this
+ * the alias table is only reachable by input that is already lowercase.
+ *
+ * NFKC and NOT NFD, deliberately: NFD decomposes precomposed letters too, so
+ * the same strip would reach the dots of ż / ė / ġ and collapse Żubr onto
+ * Zubr, Ėmė onto Eme, Ġenerali onto Generali.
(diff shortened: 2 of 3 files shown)


abankar1:
Heads-up on the red check here: **`web typecheck + build` is failing for a reason unrelated to this PR.**

It is `tests/lib/tracker-lock.test.mjs` (added by #2903) failing with:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'js-yaml' imported from
  /home/runner/work/career-ops/career-ops/tracker-utils.mjs
```

The web job installs only `web/`'s dependencies, and that test reaches up into the root `tracker-utils.mjs`, which imports `js-yaml` from the root manifest. @rubicon filed it as #2922 while I was pushing this branch.

My own new test (`web/tests/lib/status-alias.test.mjs`) is unaffected — it imports `js-yaml` directly, and `js-yaml` is a declared **web** dependency, so it resolves inside `web/node_modules`. It passes in that same CI run.

Not rebasing onto a fix for #2922, since that is @rubicon's to land — happy to rebase once it does.

abankar1:
Rebased onto `main`. The red `web typecheck + build` here was #2922 (the tracker-lock test reaching root `js-yaml` from a web-only install), which is now fixed on main — so this branch should go green without any change to its own contents.

`web`: 272 passed / 0 failed. `node test-all.mjs --quick`: 4318 passed / 0 failed.

santifer:
Not forgotten, @abankar1 — the 27-alias drift you mapped in #2917 is real and the pairing of issue+fix is appreciated. It's in the web lane (area:web) for the substantive review; given it reads states.yml, the contract-freeze checks in CI are the thing to keep green. Sorry for the slow first touch.

santifer (approved):
This is the strongest PR in the batch and it's squarely on the surface I own. Approving.

You didn't just add the 27 missing aliases — you fixed the *class*. Extracting the map to `status-alias.mjs` so a `node --test` can load it, and having the test load the real `templates/states.yml` and assert every alias resolves, is exactly what #2249 and #2917 needed and never got: the copy can't drift silently again because a new alias in states.yml now fails CI here until it's mirrored. That's the derive-or-assert cure at the right tier — you can't read states.yml at runtime (format.ts ships to the client), so you assert the copy against the source by executing both.

And the fold is correct for the first time. `canonStatus` was `trim().toLowerCase()`, which left every capitalised Turkish status unreachable — the table was only ever hit by input that was already lowercase. Your `foldStatus` mirrors the core's `foldStatusInput` (#2705): NFKC + the U+0307 strip so `İşe alındı` folds to `işe alındı`. The test that the same strip does **not** reach the dots of ż / ė / ġ (Żubr / Ėmė / Ġenerali pass through, don't collapse onto Zubr / Eme / Generali) is the NFKC-not-NFD safety property, and testing it directly is what keeps the next person from "simplifying" it to NFD.

Contract-freeze §55 is respected: `CANONICAL_STATES` is untouched, the change is additive aliases + a fold fix, neither of which is the frozen canonical list.

**One thing I'm taking as my own follow-up, not asking of you:** `foldStatus` here is now a third copy of that fold — the core has it in `tracker-utils.mjs`, I put one in `states.ts` (#2786), and this adds a third. You couldn't import mine (a `.mjs` can't import a `.ts` under Node — same wall #2102 hit), so copying was the right call, and your fold-behaviour tests lock this copy on the hard cases so it can't regress on its own. But the three copies aren't asserted identical to each other. Collapsing the web's two (states.ts + status-alias.mjs) onto one shared fold `.mjs` — states.ts can import a `.mjs` even though the reverse fails — is my architecture to clean up, and I'll do it as a follow-up. Your PR stands on its own; I'm not blocking it on my housekeeping.

Merges once the batch's merge run is open again.

santifer:
Thanks, @abankar1 — the strongest fix in today's web batch, and on a surface I care about. You didn't just add the 27 missing aliases; you killed the class: extracting the map to status-alias.mjs and having the test load the real states.yml means an alias can never again resolve to no stage silently. And you fixed the Turkish fold (NFKC + the U+0307 strip, mirroring the core) with a test that the strip does NOT reach ż/ė/ġ — the NFKC-not-NFD safety property, asserted directly. §55 is respected (CANONICAL_STATES untouched). Closes #2917. Merged 🚀

PR #2923, opened 2026-08-15
fix(providers): undecoded &amp; in titles silently drops matching jobs and bypasses negative filters

### LINKED ISSUE #2921: providers: undecoded &amp; in titles silently drops matching jobs and bypasses negative filters (#2487 not swept)
## Summary

Four providers strip HTML tags out of the title but never decode the entities in it, so `R&amp;D Engineer` reaches the pipeline verbatim. That is not only cosmetic: **it silently changes which jobs the scan keeps**, in both directions.

`beesite.mjs:107`, `csod.mjs:140`, `phenom.mjs:113` + `:132`, `tkms.mjs:105` all do:

```js
String(d.title || '').replace(/<[^>]*>/g, ' ').replace(/\s+/g, ' ').trim()
```

The `replace(/<[^>]*>/g, …)` is itself the evidence the field carries HTML. #2487 fixed exactly this shape in `softgarden.mjs` and `radancy.mjs` by routing through `providers/_html-entities.mjs`:

```js
decodeEntities(s.replace(/<[^>]*>/g, ' ')).replace(/\s+/g, ' ').trim()
```

These four were not swept.

## Why it is a filter bug, not a display bug

`buildTitleFilter` in `scan.mjs` lowercases and substring/word-matches. An undecoded `&amp;` breaks the match:

```
POSITIVE filter: ["r&d", "research & development"]

  ATS title: "R&amp;D Engineer"
    undecoded -> "R&amp;D Engineer"              kept: false
    decoded   -> "R&D Engineer"                  kept: true

  ATS title: "Research &amp; Development Lead"
    undecoded -> "Research &amp; Development Lead"  kept: false
    decoded   -> "Research & Development Lead"      kept: true
```

So a role the user's own `title_filter.positive` asks for is **silently dropped** — it never reaches `pipeline.md` and the user never learns it existed. Zero-token scanning makes this invisible: there is no LLM pass downstream that would notice the mangled title.

The negative side fails the opposite way — a veto the user wrote is bypassed:

```
NEGATIVE filter: ["sales & marketing"]

  ATS title: "Sales &amp; Marketing Lead"
    undecoded -> kept: true    <- the veto never fires
    decoded   -> kept: false
```

Ampersands are not exotic in job titles: *R&D*, *Research & Development*, *Sales & Marketing*, *Health & Safety*, *Mergers & Acquisitions*. Numeric entities are just as common in DACH/FR postings, which is where beesite and tkms actually point:

```
"Softwareentwickler (m/w/d) &#8211; Backend"  ->  "Softwareentwickler (m/w/d) – Backend"
"Ingénieur Syst&#232;mes"                    ->  "Ingénieur Systèmes"
```

## Also: hackernews.mjs carries a partial local decoder

`providers/hackernews.mjs:47` has its own `ENTITY_MAP` + `ENTITY_RE`, the same duplication #2487 removed from softgarden and radancy — but this copy is a strict subset. `ENTITY_RE` is a literal alternation of seven forms:

```js
const ENTITY_RE = /&amp;|&lt;|&gt;|&quot;|&#x27;|&#39;|&nbsp;/g;
```

So it decodes `&#39;` but leaves every other numeric entity (`&#8211;`, `&#232;`, `&#x2013;`) and every other named one (`&uuml;`, `&eacute;`) untouched. The shared decoder handles the full numeric range plus the named set.

## Proposed fix

Route all five through `providers/_html-entities.mjs`, matching the #2487 shape exactly, and add assertions to the existing `tests/providers/{beesite,csod,phenom,tkms,hackernews}.test.mjs`.

abankar1:
Fixes #2921.

Four providers stripped HTML tags out of the title but never decoded the entities in it, so `R&amp;D Engineer` reached the pipeline verbatim. The `replace(/<[^>]*>/g, …)` is itself the evidence the field carries HTML — #2487 fixed this exact shape in `softgarden.mjs` and `radancy.mjs`, and `beesite`, `csod`, `phenom` and `tkms` were not swept.

## It changes which jobs the scan keeps

`buildTitleFilter` in `scan.mjs` lowercases and substring/word-matches, so an undecoded ampersand moves the filter boundary — in **both** directions:

```
POSITIVE ["r&d", "research & development"]
  "R&amp;D Engineer"                 undecoded kept:false    decoded kept:true
  "Research &amp; Development Lead"  undecoded kept:false    decoded kept:true

NEGATIVE ["sales & marketing"]
  "Sales &amp; Marketing Lead"       undecoded kept:true   <- veto never fires
                                     decoded   kept:false
```

A role the user's own `title_filter.positive` asks for is **silently dropped** — it never reaches `pipeline.md` and the user never learns it existed. A veto they wrote is silently bypassed. Zero-token scanning is what makes this invisible: there is no LLM pass downstream that would notice the mangled title.

Ampersands are ordinary in job titles (R&D, Sales & Marketing, Health & Safety, Mergers & Acquisitions), and numeric entities are just as common in the DACH/FR postings `beesite` and `tkms` actually point at:

```
"Softwareentwickler (m/w/d) &#8211; Backend"  ->  "Softwareentwickler (m/w/d) – Backend"
"Ingénieur Syst&#232;mes"                    ->  "Ingénieur Systèmes"
```

`phenom` gets both of its sites: the title and the location fallback chain, which strips tags the same way.

## hackernews carried a 7-form local decoder

`providers/hackernews.mjs` had its own `ENTITY_MAP` + `ENTITY_RE`, the same duplication #2487 removed — but this copy was a strict subset. `ENTITY_RE` was a literal alternation of seven forms, so it decoded `&#39;` and left every other numeric entity and every other named one in place. Measured on the reverted code:

```
"Acme GmbH | Softwareentwickler &#8211; R&D | M&#252;nchen |"
```

The `&amp;` decoded; the en dash and the umlaut did not. Now `"Softwareentwickler – R&D"`, location `"München"`. Decoding still happens **after** tag stripping, exactly as the local map did, so href extraction and the `<p>`/`<br>` → newline pass are unaffected.

## Commits

1. `beesite`, `csod`, `phenom`, `tkms` — route through `providers/_html-entities.mjs`.
2. `hackernews` — replace the partial local map with the shared decoder.
3. `tests/providers/title-entity-decode.test.mjs`.

## Verification

The test asserts the end-to-end claim rather than the string transform alone: each provider's own parse function is fed an entity-bearing fixture, and the resulting title is then run through the **real** `buildTitleFilter` with a positive `"r&d"`. That is what makes it a filter test rather than a cosmetics test.

One assertion deliberately pins the *premise* instead of the fix — an undecoded `"Sales &amp; Marketing Lead"` must still slip past a negative `"sales & marketing"`. If `buildTitleFilter` ever normalizes entities itself, that assertion fails and tells the next reader the provider-side decode became redundant, rather than leaving a silently pointless test behind.

Checked for discrimination: **with all five provider fixes reverted, 11 of the 12 assertions fail**, and the survivor is exactly that premise test.

`node test-all.mjs --quick`: 3947 pass / 0 fail (the 2 warnings are pre-existing and environmental — no user data, no Playwright browser). Existing per-provider suites still pass unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved decoding of HTML-encoded characters in job titles and locations.
  * Expanded entity handling in comment content, including numeric entities.
  * Filters and displayed listings now use correctly decoded text.

* **Tests**
  * Added coverage for encoded titles, locations, comments, and title-filter behavior across supported providers.

### DIFF
--- a/providers/beesite.mjs
+++ b/providers/beesite.mjs
@@ -30,6 +30,13 @@
 //     searchCriteria:
 //       - { CriterionName: PositionLocation.Country, CriterionValue: [329] }
 
+// Titles arrive HTML-escaped, so the tag strip below is not enough on its own:
+// an undecoded "R&amp;D Engineer" fails the user's own title_filter positive
+// "r&d" and is silently dropped, and a negative like "sales & marketing" never
+// vetoes "Sales &amp; Marketing Lead". Shared decoder, same as softgarden and
+// radancy (#2487, #2921).
+import { decodeEntities } from './_html-entities.mjs';
+
 const PAGE_SIZE = 100; // verified: the endpoint happily serves 100+/page
 const MAX_PAGES = 40; // safety cap on request count (40*100 = 4000 postings)
 const MAX_JOBS = 1000; // cap total postings pulled (newest-first sort)
@@ -104,7 +111,7 @@ export function parseSearchResult(json) {
     const d = item?.MatchedObjectDescriptor;
     if (!d) continue;
     const id = item.MatchedObjectId != null ? String(item.MatchedObjectId) : String(d.PositionID || '');
-    const title = String(d.PositionTitle || '').replace(/<[^>]*>/g, ' ').replace(/\s+/g, ' ').trim();
+    const title = decodeEntities(String(d.PositionTitle || '').replace(/<[^>]*>/g, ' ')).replace(/\s+/g, ' ').trim();
     const url = String(d.PositionURI || '').trim();
     if (!id || !title || !/^https?:\/\//i.test(url)) continue;
     const locs = Array.isArray(d.PositionLocation) ? d.PositionLocation : [];
--- a/providers/csod.mjs
+++ b/providers/csod.mjs
@@ -27,6 +27,13 @@
 // shape; the branded corporate page goes in careers_url and the csod.com URL in
 // `api:` (same convention as workday/successfactors).
 
+// Titles arrive HTML-escaped, so the tag strip below is not enough on its own:
+// an undecoded "R&amp;D Engineer" fails the user's own title_filter positive
+// "r&d" and is silently dropped, and a negative like "sales & marketing" never
+// vetoes "Sales &amp; Marketing Lead". Shared decoder, same as softgarden and
+// radancy (#2487, #2921).
+import { decodeEntities } from './_html-entities.mjs';
+
 const PAGE_SIZE = 25; // server default; verified OHB serves exactly 25/page
 const MAX_PAGES = 40; // safety cap on request count (40*25 = 1000 postings)
 const MAX_JOBS = 1000; // cap total postings pulled per site
@@ -137,7 +144,7 @@ export function parseRequisitions(json, cfg) {
   for (const r of list) {
     if (!r || typeof r !== 'object') continue;
     const id = r.requisitionId != null ? String(r.requisitionId) : '';
-    const title = String(r.displayJobTitle || '').replace(/<[^>]*>/g, ' ').replace(/\s+/g, ' ').trim();
+    const title = decodeEntities(String(r.displayJobTitle || '').replace(/<[^>]*>/g, ' ')).replace(/\s+/g, ' ').trim();
     if (!id || !title) continue;
     out.push({
       id,
--- a/providers/hackernews.mjs
+++ b/providers/hackernews.mjs
@@ -43,18 +43,12 @@ function extractUrl(text) {
   return m ? m[0].replace(/[.,;!?)]+$/, '') : '';
 }
 
-// Named HTML entities we decode in comment bodies. Kept in single map
-/** @type {Record<string, string>} */
-const ENTITY_MAP = {
-  '&amp;': '&',
-  '&lt;': '<',
-  '&gt;': '>',
-  '&quot;': '"',
-  '&#x27;': "'",
-  '&#39;': "'",
-  '&nbsp;': ' ',
-};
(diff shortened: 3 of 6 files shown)


abankar1:
Cross-reference for whoever reviews these together: this overlaps in subject with #2902 (@FReptar0) but not in mechanism, and I do not think either blocks the other.

#2902 asserts at the source that a provider which **imports** `providers/_html-entities.mjs` must not **also** declare its own `decodeEntities`/`decodeXmlEntities`. That catches a private copy that is correct on the day it lands and drifts later.

It would not catch the five here, because none of them imported the shared module at all:

- `beesite`, `csod`, `phenom`, `tkms` had **no** decoder, private or shared — just a bare tag strip.
- `hackernews` had a private map and no import, so the "imports AND declares" rule does not fire on it.

After this PR all five import the shared decoder and declare nothing, so they satisfy #2902's rule rather than tripping it. If #2902 lands first this should rebase cleanly; if this lands first, #2902's guard simply covers five more files.

Also confirmed these four are disjoint from the seven RSS providers @gunjanjaswal migrated in #2818 — those are RSS, these are JSON search APIs, and I verified the gap against current `main`.

Scott-Emberson (approved):
Code owner for `tests/`. `title-entity-decode.test.mjs` runs each provider's decoded title through the real `buildTitleFilter`, and 11 of its 12 assertions fail on revert. The surviving one is the negative-premise sentinel that proves the undecoded path fails and the decoded path passes. Hermetic, sibling shape. Approving.

santifer:
The failure mode here was invisible by construction: an `&amp;` in a title made a matching job silently vanish AND let it bypass negative filters, and nothing reported either. Routing titles through the shared decoder before any matching closes both doors at once. Merged! 🚀

PR #2850, opened 2026-08-14
fix(add-entry): a non-Latin CV can't use /career-ops add — the dedupKey you supplied is rejected as empty

abankar1:
Closes #2849.

`normalizeKey()` strips `[^a-z0-9]+`, so every non-Latin string keys to `''`. That key gates section matching, dedup **and the payload validation** — and validation bites first:

```js
if (!normalizeKey(dedupKey)) throw new Error('payload.cv requires a non-empty dedupKey …');
```

Measured, same payload in two languages:

```console
$ node add-entry.mjs payload.json     # section: プロジェクト, dedupKey: フラウドシールド
add-entry: payload.cv requires a non-empty dedupKey (used for dedup/idempotency)

$ node add-entry.mjs english.json
{ "dryRun": false, "cv": { "status": "added", "section": "Projects" } }
```

`/career-ops add` is the documented path for adding a finished project, paper or role to `cv.md`. For a non-English CV it's **not usable at all**, and the error blames the user's input for something the tool couldn't read.

**Second consequence past the guard:** `normalizeKey('プロジェクト') === normalizeKey('職務経歴')` is `true`, so an entry can be inserted under the wrong heading — a write to `cv.md`, the user-layer file the Data Contract protects hardest.

## Fix

Delegate to the shared `normalizeTextKey()` instead of a private `[a-z0-9]` strip — same remedy as #2429, #2502, #2518, #2587, #2781.

After:

```
{ "cv": { "status": "added", "section": "プロジェクト" } }     # first run
{ "cv": { "status": "duplicate" } }                          # idempotent re-run
別プロジェクト lands under 職務経歴, not under プロジェクト      # no empty-key collision
```

## Latin path

Unchanged across everything I checked — `FraudShield` / `Fraud-Shield` / `fraud shield` → `fraudshield`, `C++ Tooling` → `ctooling`, `Node.js` → `nodejs`, `A/B Testing` → `abtesting`, `  Spaced  ` → `spaced`.

One difference, and it's a repair rather than a change: `Café` was truncated to `caf` and now keys as `café`. It never matched `Cafe` under either implementation, so nothing that used to dedup stops deduping — the old value was just corruption.

## Tests

Three assertions driven through the real CLI (add → idempotent re-run → no cross-section collision). On `main` the first two fail with the CLI erroring before it emits any JSON.

`node test-all.mjs --quick` → **3802 passed, 0 failed**.

## Note on the failing Windows check

[...]

PR #2927, opened 2026-08-15
fix(trust-validator): a company with an accent in its name is flagged for being on its own domain

### LINKED ISSUE #2924: _trust-validator: a company with an accent in its name is flagged for being on its own domain
## Summary

`companyMatchesHostname` in `providers/_trust-validator.mjs:119` **deletes** every character outside `[a-z0-9 ]` instead of folding it, so a company whose own name carries a diacritic fails to match its own official domain and is penalised as a legitimacy risk.

```js
const normalized = company.toLowerCase().replace(/[^a-z0-9 ]/g, '').trim();
```

`é` is removed rather than folded to `e`, so `"Société Générale"` becomes `"socit gnrale"` — which is not a substring of `societegenerale`, and neither are its words.

## Measured

```
company            hostname                       matches?  (false = mismatch FLAG raised)
Société Générale   careers.societegenerale.com    false
Telefónica         jobs.telefonica.com            false
Schrödinger        schrodinger.com                false
Rakuten            rakuten.co.jp                  true     <- plain-Latin control
```

Rule 4 in `buildTrustValidator` then fires:

```js
if (!companyMatchesHostname(company, hostname)) {
  flags.push('company_domain_mismatch');
  score -= PENALTIES.company_domain_mismatch;   // 15
}
```

An otherwise clean posting drops 100 → 85, which crosses `classifyTrustLevel`'s boundary from **high** to **medium**. So a legitimate posting on the company's own domain is reported as less trustworthy purely because the company spells its name with an accent. Stacked with any other flag it degrades further (`missing_apply_url` + this = 45 → **low**).

This is the same key-building defect as #2782 / #2848 / #2850, in the legitimacy path rather than the matching path: a private key built with `[^a-z0-9]` deletes non-ASCII rather than folding it, and the symptom is whatever the call site does with the empty-or-mangled result.

Some accented names pass by luck rather than design, which is why it has gone unnoticed — `Nestlé` → `nestl`, and `nestle.com` happens to contain `nestl`; `Ørsted` → `rsted`, and `orsted.com` happens to contain `rsted`. The check succeeds there for the wrong reason.

## The other half is correct and should stay

A fully non-Latin name normalizes to `''` and the function returns `true` ("can't evaluate → no flag"):

```
楽天       evil-phishing.example    true
Сбербанк   evil-phishing.example    true
```

That looks like a hole, but it is the right call and should be **kept deliberately**: hostnames are effectively ASCII, so a CJK or Cyrillic company name can never appear in one, and the absence of a match proves nothing. Making the check "work" for those names would flag every non-Latin company posting on its own legitimate domain — turning a silent skip into a systematic false positive. The bug is only the accented-**Latin** case, where the ASCII hostname genuinely does contain the folded name.

## Proposed fix

ASCII-fold before stripping: NFD, drop combining marks, then map the Latin letters that do not decompose (`ø æ œ ß đ ł þ ð`). Keep the empty-result escape, with a comment explaining why it is deliberate rather than an oversight.

Worth noting explicitly, because I have argued the opposite elsewhere: NFD is the **right** tool here and the wrong one in `foldStatusInput` (#2705). There the fold must not collapse distinct identities (Żubr must not become Zubr); here we are deliberately comparing against an ASCII hostname, so folding to the ASCII base letter is exactly the intent. Different question, different answer.

`test-trust-validator.mjs` already covers this function and is wired into `test-all.mjs`, so the regression assertions have a home.

abankar1:
Fixes #2924.

`companyMatchesHostname` in `providers/_trust-validator.mjs` **deleted** every character outside `[a-z0-9 ]` instead of folding it, so a company whose own name carries a diacritic failed to match its own official domain and was penalised as a legitimacy risk.

```
company            hostname                       matches?  (false = mismatch FLAG raised)
Société Générale   careers.societegenerale.com    false
Telefónica         jobs.telefonica.com            false
Schrödinger        schrodinger.com                false
Rakuten            rakuten.co.jp                  true      <- plain-Latin control
```

`é` was removed rather than folded to `e`, so `"Société Générale"` became `"socit gnrale"` — not a substring of `societegenerale`, and neither are its words.

Rule 4 in `buildTrustValidator` then fires a 15-point `company_domain_mismatch` penalty, dropping a clean 100 → 85 and crossing `classifyTrustLevel`'s **high → medium** boundary. A legitimate posting on the company's own domain is reported as less trustworthy purely because the company spells its name with an accent. Stacked with any other flag it degrades further (`missing_apply_url` + this = 45 = **low**).

Same key-building defect as #2782 / #2848 / #2850, in the legitimacy path rather than the matching path.

## The fix

ASCII-fold before stripping: NFD → drop combining marks → map the Latin letters that do not decompose (`ø æ œ ß đ ł þ ð`).

**On NFD, since I argued the opposite in #2705.** NFD is the right tool here and the wrong one in `foldStatusInput`. There the fold must not collapse distinct identities — `Żubr` must never become `Zubr`. Here we are deliberately comparing against an ASCII hostname, so folding to the ASCII base letter is the entire point: `societegenerale.com` *is* the ASCII folding of `Société Générale`. Different question, different answer, and I have put that reasoning in the code so the two do not look inconsistent to the next reader.

## What I deliberately did NOT "fix"

A fully non-Latin name folds to `''` and the function returns `true` — "cannot evaluate → no flag":

```
楽天       evil-phishing.example    true
Сбербанк   evil-phishing.example    true
```

That reads like a hole, and it is not one. Hostnames are effectively ASCII, so a CJK or Cyrillic name can never appear in one and the absence of a match proves nothing. Making the check "work" for those names would flag **every** non-Latin company posting on its own legitimate domain — trading a silent skip for a systematic false positive. The escape is kept and now carries a comment saying why, plus three assertions, so it does not get filed as a bug later.

## Verification

`test-trust-validator.mjs` already covered this function and is wired into `test-all.mjs`, so the new assertions live there. They fall into three groups, only the first of which is the bug:

1. The four accented-Latin names that genuinely failed.
2. `Nestlé` / `Ørsted` / `Æon`, which passed **before** the fix but by coincidence — `nestl` is a substring of `nestle`, `rsted` of `orsted`. Pinned so the fold, not the accident, keeps them green.
3. Guards that the fold did not turn the check into "always true": an accented name on a hostile or unrelated host must **still** be flagged. Without those two, deleting the function body would pass this suite.

Checked for discrimination: with the source fix reverted, **exactly the 4 genuinely-broken cases fail (98/102)** and every must-not-change assertion still passes.

`node test-all.mjs --quick`: 3935 pass / 0 fail (the 2 warnings are pre-existing and environmental — no user data, no Playwright browser).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved company-to-domain matching for names with accented Latin characters.
  * Added support for special character conversions, such as `ø` to `o` and `æ` to `ae`.
  * Non-Latin company names are now handled safely without generating incorrect mismatch warnings.

* **Tests**
  * Added coverage for matching, unrelated, and potentially deceptive domains involving accented and non-Latin names.

### DIFF
--- a/providers/_trust-validator.mjs
+++ b/providers/_trust-validator.mjs
@@ -104,9 +104,19 @@ export function matchesDomainList(hostname, domainList) {
 
 // Latin letters that do NOT decompose under NFD, so stripping combining marks
 // alone still deletes them. Lowercase only — the caller lowercases first.
+// A stroke or bar through a letter is part of the glyph, not a combining
+// mark, so NFD leaves these untouched and the [^a-z0-9] strip below then
+// DELETES them — the very failure this fold exists to stop. The Turkish
+// dotless ı is the one that actually bites: "Işık" scored no match against
+// isik.com.tr and was flagged for being on its own domain. ħ/ŋ/ŧ happened to
+// pass anyway on a word-level substring, which is luck, not correctness —
+// the same accidental pass "Nestlé" had before this fold existed.
+// (CodeRabbit, reviewing #2927.)
 const NON_DECOMPOSING_LATIN = [
   [/ø/g, 'o'], [/æ/g, 'ae'], [/œ/g, 'oe'], [/ß/g, 'ss'],
   [/đ/g, 'd'], [/ł/g, 'l'], [/þ/g, 'th'], [/ð/g, 'd'],
+  [/ħ/g, 'h'], [/ı/g, 'i'], [/ŋ/g, 'ng'], [/ŧ/g, 't'],
+  [/ĸ/g, 'k'], [/ſ/g, 's'],
 ];
 
 /**
--- a/providers/_trust-validator.mjs
+++ b/providers/_trust-validator.mjs
@@ -102,6 +102,35 @@ export function matchesDomainList(hostname, domainList) {
   return false;
 }
 
+// Latin letters that do NOT decompose under NFD, so stripping combining marks
+// alone still deletes them. Lowercase only — the caller lowercases first.
+const NON_DECOMPOSING_LATIN = [
+  [/ø/g, 'o'], [/æ/g, 'ae'], [/œ/g, 'oe'], [/ß/g, 'ss'],
+  [/đ/g, 'd'], [/ł/g, 'l'], [/þ/g, 'th'], [/ð/g, 'd'],
+];
+
+/**
+ * ASCII-fold a company name for comparison against a hostname.
+ *
+ * NFD is the right tool HERE and the wrong one in foldStatusInput()
+ * (tracker-utils.mjs, #2705). There the fold must not collapse distinct
+ * identities — Żubr must never become Zubr. Here we are deliberately comparing
+ * against an ASCII hostname, so folding to the ASCII base letter is the whole
+ * point: "societegenerale.com" IS the ASCII folding of "Société Générale".
+ *
+ * The previous `[^a-z0-9 ]` strip DELETED accented letters instead of folding
+ * them, so "Société Générale" became "socit gnrale" and matched neither the
+ * slug nor any word of its own domain (#2924).
+ *
+ * @param {string} company
+ * @returns {string} Lowercased, space-separated ASCII, or '' when nothing Latin remains.
+ */
+export function asciiFoldForHostname(company) {
+  let out = String(company ?? '').toLowerCase().normalize('NFD').replace(/\p{M}+/gu, '');
+  for (const [re, to] of NON_DECOMPOSING_LATIN) out = out.replace(re, to);
+  return out.replace(/[^a-z0-9 ]/g, '').replace(/\s+/g, ' ').trim();
+}
+
 /**
  * Heuristic: does the company name plausibly match the URL hostname?
  *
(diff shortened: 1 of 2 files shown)


abankar1:
Fixed in the latest push. Verified rather than taken on faith, and the result splits in a way worth recording.

Only **one** of the four named letters was a genuine failure here:

```
Işık    vs  isik.com.tr     ->  false   (flagged for being on its own domain)
Ħamrun  vs  hamrun.com.mt   ->  true
Ŋaro    vs  ngaro.example   ->  true
Ŧorne   vs  torne.example   ->  true
```

The three that passed did so on a word-level substring after the letter was deleted (`amrun` is in `hamrun`), not on the fold — the same accidental pass `Nestlé` and `Ørsted` had before this fold existed, and which the PR already pins for that reason. So the finding is right about the mechanism and right to fix all four; it is only the blast radius that is narrower than stated. With the mapping reverted, exactly `Işık` reddens.

Added ħ→h, ı→i, ŋ→**ng**, ŧ→t, ĸ→k, ſ→s.

**ŋ is "ng", not "n".** I wrote the plausible one-to-one mapping first and the test caught it: `Ŋaro` folded to `naro`, which is not a substring of `ngaro.example`.

Worth flagging that #2931 carries the same list in `lib/ascii-fold.mjs`, and there the gap is **not** cushioned by substring luck — `Işık` derived the slug `isk` and never `isik`, so `--add` probed a slug no board uses. All four are genuine misses there. I have pushed the same mapping to that branch.

`test-trust-validator.mjs`: 107 passed / 0 failed. The hostile-host assertions are unchanged and still stop the fold becoming "always true".

santifer:
The distinction in the asciiFoldForHostname docstring is the best part of this PR, @abankar1: NFD-fold is the right tool against an ASCII hostname and the wrong one in foldStatusInput, and writing down WHY next to the code is what stops the next person from unifying two operations that must stay different. The deliberate non-Latin skip with its reasoning spelled out is the same quality. Merged 🚀

Issue #2904, opened 2026-08-14
web: /api/status writes the tracker itself instead of delegating to set-status.mjs (converge on one writer)


santifer:
## The debt

`POST /api/status` now writes `data/applications.md` directly, holding the core's file lock (#2903). That closed a real lost update (#2900), but it left the tracker with **two writers that must agree on locking forever**: the core scripts and this route.

The house rule is the opposite — *the web orchestrates the real engine, it never reimplements it* — and the precedent is already there:

```
/api/tracker/delete → tracker.mjs
/api/run            → merge-tracker.mjs, reserve-report-num.mjs
/api/status         → writes the file itself     ← the exception
```

**#2901 (@rubicon) proposes the principled alternative**: delegate to `set-status.mjs --source web`. One writer, no possible divergence in *how* a row is written, and the `--source` attribution feeds the transition ledger for free — which would make **#2715** unnecessary rather than parallel.

I merged #2903 instead because the data loss was live and I had a reproduction in hand. That was a call about urgency, not about design, and the design question is still open.

## What has to be decided first — not after

**1. The tracker lock is not reentrant.** (No owner/pid self-check in `tracker-utils.mjs`.) Delegation moves the lock into the child process, which is fine on its own — but any route that already holds the lock and then orchestrates `set-status.mjs` self-deadlocks until the timeout. It cannot happen today; it can the day `/api/run` needs to set a status mid-run. **Migrating to delegation means ruling on that before, not discovering it after.**

**2. Cost per call.** `/api/status` is the most-clicked write in the UI — a dropdown. Delegation pays a process spawn plus Node startup (tens to hundreds of ms) against a few ms today. #2901 carries `SET_STATUS_TIMEOUT_MS` and a bounded lock wait, so the author clearly considered it; the numbers need measuring, not estimating.

**3. Data-only checkouts.** The web can run against a `CAREER_OPS_ROOT` with no scripts (there is an explicit 400 for that elsewhere). Delegation needs a defined answer when `set-status.mjs` is absent. "Status updates silently stop working" is not one.

## What survives either way

The test from #2903 (`web/tests/lib/tracker-lock.test.mjs`) asserts the **behaviour**: two concurrent writers lose an update without a lock, and don't with one. It does not assume who holds the lock or where the write happens, so it stays valid after a migration to delegation — and it should keep guarding it.

## Credit

[...]

PR #2928, opened 2026-08-15
fix(ci): install root deps so the web job can import the core tracker lock

### LINKED ISSUE #2922: Web CI has been red on main since #2903 — the new tracker-lock test imports a root module the web job never installs
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### Existing issues

- [x] I have searched existing issues and this is not a duplicate

### What happened?

`Web CI` has failed on every `main` push since **d9c4fd0** (#2903). It passed on every commit before it:

```
2026-08-14T22:13:09  failure  d9c4fd0   <- #2903
2026-08-14T19:58:34  success  6eece73
2026-08-14T14:29:38  success  be62e3e
2026-08-14T14:26:19  success  05cc972
2026-08-14T14:02:37  success  2d43601
```

The cause is a dependency the job cannot satisfy, not a defect in the code under test.

`web/tests/lib/tracker-lock.test.mjs` (added by #2903) reaches out of `web/` to exercise the real core primitive:

```js
const { acquireTrackerLock, trackerLockDirFor } = await import(pathToFileURL(coreLock).href);
```

That is the right thing for the test to do — it is what makes the test assert the *behaviour* rather than a re-implementation, which is exactly why it survives a migration to delegation. But the module it imports needs root dependencies:

```js
// tracker-utils.mjs:15
import * as yaml from 'js-yaml';
```

and `web-ci.yml` installs only the web workspace:

```yaml
defaults:
  run:
    working-directory: web
steps:
  - run: npm ci        # web/ only — no root node_modules
  - run: npm test      # <- the dynamic import fails here
```

So `js-yaml` is unresolvable and the import throws.

**Why this is worth fixing beyond the red badge:** every open `web/` PR now shows a failing `web typecheck + build` for a reason that has nothing to do with the PR. That is the condition where a real failure stops being distinguishable from the ambient one — the same "no signal" shape as an unacknowledged channel. I noticed it only because I was checking whether my own PRs' CI was red on their own merits.

### Steps to reproduce

```bash
git checkout d9c4fd0
cd web && npm ci && npm test
```

The `tracker-lock` cases fail on `ERR_MODULE_NOT_FOUND` for `js-yaml`, imported from `tracker-utils.mjs`.

### Expected behavior

`Web CI` green on `main`, and a failing `web typecheck + build` on a PR meaning something about that PR.

Two clean options, and the choice is a real one about what the job is for:

1. **Install root deps in the web job** — add a root `npm ci` before the web steps. Keeps the test exercising the real core primitive, which is its whole point. Costs one install step.
2. **Skip the test when the root module is unresolvable** — guard the dynamic import and skip with a stated reason. Keeps `web/` self-contained, but the test then silently does not run in CI, which is the weaker outcome for a test guarding a data-loss fix.

I would take (1): the test was written to catch a lost update on `data/applications.md`, and a test that skips in CI is not guarding anything. Happy to send it as a PR if useful — it is a two-line change to the workflow.

### CLI tool

Claude Code

### OS and Node.js version

Observed in GitHub Actions on `ubuntu-latest`; reproduced locally on macOS 26.5.2 / Node v22.22.3.

### Output of npm run doctor

Not setup-related — this is a CI workflow/dependency mismatch, reproducible from a clean checkout.

rubicon:
## What does this PR do?

Installs the root dependencies in the `web typecheck + build` job, so `web/tests/lib/tracker-lock.test.mjs` can import the core tracker lock. Web CI has been red on `main` since #2903 added that test.

## Related issue

Closes #2922

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

---

## The failure

`Web CI` flipped to red exactly at `d9c4fd0` (#2903) and has stayed there:

```
2026-08-14T22:13:09  failure  d9c4fd0   <- #2903
2026-08-14T19:58:34  success  6eece73
2026-08-14T14:29:38  success  be62e3e
```

`web/tests/lib/tracker-lock.test.mjs` reaches out of `web/` to exercise the real primitive:

```js
const { acquireTrackerLock, trackerLockDirFor } = await import(pathToFileURL(coreLock).href);
```

`tracker-utils.mjs:15` imports `js-yaml`, and the job installs only the web workspace (`working-directory: web`, one `npm ci`), so the import throws `ERR_MODULE_NOT_FOUND`.

## Why install rather than skip

The obvious alternative is to guard the dynamic import and skip when the root module is unresolvable. I think that is the weaker option: this test was written to guard a **lost update on `data/applications.md`**, and a test that silently does not run in CI is not guarding anything. Installing the root deps keeps it asserting the real primitive — which is also what lets it survive a change of caller, the property #2904 depends on.

`--ignore-scripts` on the root install because the job only needs the module graph resolvable; the root `postinstall` downloads a Playwright browser this job never uses.

## Why it is worth landing promptly

While this sat as an issue, @abankar1 hit the same red check on #2918 and [wrote](https://github.com/santifer/career-ops/pull/2918#issuecomment-5301543337): *"Not rebasing onto a fix for #2922, since that is @rubicon's to land — happy to rebase once it does."* That is the real cost of an ambient failure: every `web/` PR now shows a failing check for a reason that is not its own, and contributors either wait or learn to ignore it.

Their own new test is unaffected — it imports `js-yaml` directly, which resolves inside `web/node_modules` because it is a declared web dependency. Only the reach into the root module is affected.

## Summary by CodeRabbit

* **Chores**
  * Improved web project CI setup by installing required root-level dependencies before running installation and tests.

### DIFF
--- a/.github/workflows/web-ci.yml
+++ b/.github/workflows/web-ci.yml
@@ -30,6 +30,13 @@ jobs:
           node-version: 24
           cache: npm
           cache-dependency-path: web/package-lock.json
+      # web/tests/lib/tracker-lock.test.mjs imports the core tracker lock from
+      # the repo root to assert the real primitive rather than a copy of it —
+      # which is what lets that test survive a change of caller. The root module
+      # pulls in js-yaml, so the root install has to happen or the dynamic
+      # import throws and every web PR shows a failure that is not its own.
+      - run: npm ci --ignore-scripts
+        working-directory: .
       - run: npm ci
       - run: npm test
       - run: npx tsc --noEmit


santifer:
Closing this one in favor of #2929, which just landed: your own tracing in #2922 showed the guard was wrong in both directions, and the resolved-import approach fixes main's red without adding a root install to the web job. Main is green again thanks to that work. If web CI ever needs to run the core cases too (not just skip them with the stated reason), installing root deps like this PR proposed is the documented next step.

Issue #2624, opened 2026-08-08
RFC: one definition, three layers — a shared-definition contract for core ↔ web ↔ dashboard


abankar1:
## The class

career-ops has one core and three consumer layers — core `.mjs`, `web/` (TypeScript), `dashboard/` (Go). Almost every shared definition is **re-expressed by hand in each layer**, and every failure in this class is silent: the copy stays syntactically valid and keeps producing plausible output while it means something different from the original.

I've spent the last two weeks fixing instances of this. I don't think there are many left that I can find by reading — but I also don't think the class is closed, because nothing structurally prevents the next one. That's what this RFC is about: not the instances, the mechanism.

Filing as an RFC rather than a PR per #284 — no code until there's a sign-off on shape.

## The census

**Fixed instances (merged, last three weeks):**

| Definition | Copy that drifted | PR |
|---|---|---|
| `states.yml` | 7 hardcoded web lists | #2250 |
| `states.yml` | `test-all` 55.3 guarded 6 of 9 ids | #2408 |
| `states.yml` → `StatusHired` | Go `StatusLabel` had no `hired` case | #2295 |
| `states.yml` → theme | Go `statusColorMap` had no `hired` | #2289 |
| `computeFunnel()` | Go `ComputeProgressMetrics` omitted Hired from all four tiers | #2406 |
| `normalizeVia` fold | `normalizeCompany` (ASCII-only) | #2430 |
| same fold | `fingerprint-core` `companyKey` | #2502 |
| same fold | `invite-match` `normalizeCompanyName` | #2518 |
| same fold | `dedup` local ASCII copy | #2587 (@Scott-Emberson) |
| `--since` parsing | `scan-ats-full` coerced silently | #2499 |
| `$`-pattern replace | CV builders, then `theme-style` | #2588, #2597 |
| `states.yml` | `states.ts` cached it for the process lifetime | #2590 (@darkpandawarrior) |

**Still live — and this one is user-visible today.** Three layers disagree about the *same tracker file*:

| Tracker status | Go TUI | core `.mjs` | web |
|---|---|---|---|
| `Entrevista` (es) | `interview` | `interview` | `Interview` |
| `Mülakat` (tr) | **`interview`** | **`mülakat`** (unrecognized) | **`null`** (writeback rejects it) |
| `Teklif` (tr) | **`offer`** | **`teklif`** | **`null`** |
| `Başvuruldu` (tr) | **`applied`** | **`başvuruldu`** | **`null`** |

[...]

PR #2931, opened 2026-08-15
fix(verify-portals): --add probes slugs that cannot exist for an accented company name

### LINKED ISSUE #2930: verify-portals --add: an accented company name probes slugs that cannot exist, and a CJK name reports a missing argument
## Summary

`deriveSlugCandidates` in `verify-portals.mjs:135` turns every character outside `[a-z0-9\s]` into a **separator** instead of folding it, so `--add` for any company whose name carries a diacritic probes slugs that cannot exist and reports the board as missing when it is live.

```js
const words = lower
  .replace(/[^a-z0-9\s]/g, ' ')
  .replace(/\s+/g, ' ')
  .trim()
  .split(' ')
  .filter(Boolean);
```

## Measured

```
Telefónica         -> telefnica, telef-nica, telef_nica, telef, telefnicaai
Société Générale   -> socitgnrale, soci-t-g-n-rale, soci_t_g_n_rale, soci, ...
Nestlé             -> nestl, nestlai, nestltech, nestlio, nestlhq
Ørsted             -> rsted, rstedai, rstedtech, rstedio, rstedhq
Zalando            -> zalando, zalandoai, ...                 <- plain-Latin control
楽天                -> (NONE)
```

`telefonica`, `societegenerale`, `nestle` and `orsted` — the slugs these boards actually use — are not in any of those lists. Worse than a near-miss: `é` becomes a *space*, so `Société Générale` is split into four fragments and even the "first word alone" heuristic the docstring describes (`'Acme Corp' → 'acme'`) yields `soci`.

Both call sites are affected — `discoverAlternates` (`:240`) and `runAdd` (`:470`) — so the user is told the company has no board on Greenhouse/Ashby/Lever when it does.

## The non-Latin case reports the wrong error entirely

A fully non-Latin name produces zero candidates, and `runAdd:471` reads that as a missing argument:

```js
const candidates = deriveSlugCandidates(name);
if (candidates.length === 0) {
  console.error('verify-portals: --add needs a company name');
  process.exit(1);
}
```

So `node verify-portals.mjs --add 楽天` answers **"--add needs a company name"** — telling the user they omitted an argument they did supply. Whatever the right behaviour is for a CJK name (probably: say plainly that no ASCII slug can be derived, since ATS slugs are ASCII), reporting it as a usage error is not it.

## Same defect class as #2924

This is the sibling of #2924 in the portal-discovery path rather than the legitimacy path: a private key built by deleting non-ASCII rather than folding it. Same fix — ASCII-fold first (NFD, drop combining marks, map the Latin letters that do not decompose: `ø æ œ ß đ ł þ ð`), then strip.

As in #2924, NFD is the right tool here for a reason that does not generalise: ATS slugs are ASCII, so folding to the ASCII base letter is the intent (`telefonica` **is** the ASCII folding of `Telefónica`). That is the opposite of `foldStatusInput` (#2705), where the fold must not collapse distinct identities.

Since #2924's fix introduces the first `asciiFold` in the tree (in `providers/_trust-validator.mjs`) and this would be the second, the helper belongs in `lib/` rather than being copied — the derive-vs-copy point from #2624.

`test-all.mjs` section 10b already covers `deriveSlugCandidates`, so the regression assertions have a home.

abankar1:
Fixes #2930.

`deriveSlugCandidates` turned every character outside `[a-z0-9\s]` into a **separator** rather than folding it, so `--add` for any company whose name carries a diacritic probed slugs no board uses and reported a live board as missing.

```
Telefónica        -> telefnica, telef-nica, telef_nica, telef
Société Générale  -> socitgnrale, soci-t-g-n-rale, soci_t_g_n_rale, soci
Nestlé            -> nestl, nestlai, nestltech
Ørsted            -> rsted, rstedai, rstedtech
Zalando           -> zalando, ...                    <- plain-Latin control
```

`telefonica`, `societegenerale`, `nestle` and `orsted` — the slugs those boards actually use — appear in none of them. Worse than a near-miss: `é` became a **space**, so `Société Générale` split into four fragments and even the first-word heuristic the docstring describes (`'Acme Corp' → 'acme'`) yielded `soci`.

Both call sites were affected — `discoverAlternates()` and `runAdd()`.

After the fold:

```
Telefónica        -> telefonica, telefonicaai, ...
Société Générale  -> societegenerale, societe-generale, societe, ...
Ørsted            -> orsted, ...
Æon               -> aeon, ...
```

## A misleading error, fixed too

A name with no Latin content folds to `''` — a real answer, since ATS slugs are ASCII — but `runAdd` read zero candidates as a missing argument:

```
$ node verify-portals.mjs --add 楽天
verify-portals: --add needs a company name
```

It told the user they had omitted an argument they did supply. The two cases are now distinguished, and the non-Latin one says what is actually wrong and what to pass instead. A genuinely empty `--add ""` still gets the original message.

## Why the helper is in `lib/`

This is the sibling of #2924 in the portal-discovery path rather than the legitimacy path — same defect, same fix. #2924's PR (#2927) introduces the first `asciiFold` in the tree, in `providers/_trust-validator.mjs`; this would be the second, so it goes in `lib/ascii-fold.mjs` rather than being copied — the derive-vs-copy point from #2624.

**I have deliberately not made this PR depend on #2927.** Each is self-contained and either can land first. Once one does, I will fold `_trust-validator.mjs`'s local `asciiFoldForHostname` onto the shared helper in a follow-up, so there is never a second copy sitting in the tree unreconciled.

**On NFD, since I argued against it in #2705.** There the fold must not collapse distinct identities — `Żubr` must never become `Zubr`, or two different companies merge. Here the comparison target is ASCII *by construction* (an ATS slug), so folding to the ASCII base letter is the entire intent: `telefonica` **is** the ASCII folding of `Telefónica`. Different question, different answer, and that reasoning is in the helper's docstring so the two do not read as inconsistent.

The helper also handles the Latin letters NFD does not decompose (`ø æ œ ß đ ł þ ð`) — a combining-mark strip alone still deletes those, which is why `Ørsted` yielded `rsted`.

## Verification

`test-all.mjs` section 10b already covered `deriveSlugCandidates`, so the new assertions live there. Five accented names are pinned to the slug their board actually uses, plus two guards for what the fold must **not** do:

- A distinct company must still derive a distinct slug, and the first-word candidate must survive (`Société Générale` → `societe`). Without this, returning a constant would pass.
- A name with no Latin content must still derive nothing — the correct answer in an ASCII slug space, and what the new error message depends on.

Checked for discrimination: with the fold reverted, **both fold assertions fail and the non-Latin guard still passes** — exactly the intended split.

The pre-existing `'Acme Corp!'` and `'Deepset'` fixtures are unchanged and still pass, so this is additive to the documented behaviour.

`node test-all.mjs --quick`: 3938 pass / 0 fail (the 2 warnings are pre-existing and environmental — no user data, no Playwright browser).

## Summary by CodeRabbit

- **New Features**
  - Improved matching for company names with accented Latin characters, ligatures, and other Latin variants.
  - Converted company names into cleaner ASCII-based slug candidates for portal verification.
  - Normalized whitespace and unsupported characters for more consistent candidate generation.

- **Bug Fixes**
  - Improved error messages when a company name is empty or cannot produce usable slug candidates.
  - Preserved distinct slug candidates for different company names.

- **Tests**
  - Added coverage for accented names, ligatures, non-decomposing characters, and names without Latin content.


santifer:
Second half of the pair and the better-shaped one, @abankar1: extracting lib/ascii-fold.mjs with both call sites named in its header, the ŋ→ng correction credited, SYSTEM_PATHS registered without being asked, and the '' contract pushed to callers with the CJK missing-argument fix in runAdd. Merged 🚀 Follow-up on my side: migrating _trust-validator to consume this shared primitive so the two folding tables can never drift.

PR #2933, opened 2026-08-15
fix(set-status): rejects today's date east of UTC, logs tomorrow's west of it

### LINKED ISSUE #2932: set-status: rejects today's date east of UTC, and stamps tomorrow's west of it (#2765 recurring on the canonical write path)
## Summary

`set-status.mjs` resolves "today" as the **UTC** day in two places, so on the write path AGENTS.md calls canonical ("strict states.yml validation, report-link mismatch guard, shared lock, atomic write") it both rejects a legitimate date and writes a wrong one — in opposite hemispheres.

This is #2765 recurring. That issue fixed exactly this in `followup-seed.mjs`, whose fix note reads:

> LOCAL date, never UTC. `toISOString()` returns the UTC day, so anywhere west of Greenwich an evening run answers "today" with tomorrow

## 1. East of UTC: your own today is rejected (`set-status.mjs:164`)

```js
if (flags.on > new Date().toISOString().slice(0, 10)) failUsage(`--on date is in the future: "${flags.on}"`);
```

Run live from this machine, right now, against a sandbox tracker:

```
$ TZ=Pacific/Auckland node set-status.mjs SandboxCo Interview --on 2026-08-16
❌ --on date is in the future: "2026-08-16"
```

2026-08-16 *is* that user's today. Anyone at a positive UTC offset hits this for the first N hours of their local day — the whole of a New Zealand or Japanese morning. There is no workaround except passing yesterday, which then records the wrong date.

## 2. West of UTC: the default event date is tomorrow (`set-status.mjs:502`)

```js
const eventDate = flags.on ?? new Date().toISOString().slice(0, 10);
```

Deterministic, pinning the instant and varying the zone:

```
instant: 2026-08-16T01:30:00Z  (UTC day 2026-08-16)

  Pacific/Auckland    local 2026-08-16   agrees
  Europe/Berlin       local 2026-08-16   agrees
  America/New_York    local 2026-08-15   event date written as 2026-08-16 = TOMORROW locally
  America/Los_Angeles local 2026-08-15   event date written as 2026-08-16 = TOMORROW locally
```

So a US-evening `set-status` stamps the tracker with a day that has not happened yet — the same symptom #2765 describes, on the row itself rather than a follow-up pin. And because `followup-seed` keys its schedule off the applied date, a wrong date here propagates into the follow-up cadence that #2765 was fixing.

The two defects are each other's mirror: the same UTC assumption rejects a real date in one hemisphere and invents a fake one in the other.

## What must NOT change

`set-status.mjs:162` also calls `toISOString()`:

```js
const roundTrips = d && !Number.isNaN(d.getTime()) && d.toISOString().slice(0, 10) === flags.on;
```

That is UTC-midnight **parsing** used to validate that a supplied date round-trips, and it is internally consistent — the same distinction #2765 drew when it changed only "what day is it here" and deliberately left date arithmetic on UTC. It should stay as it is.

## Proposed fix

Resolve the local calendar day in both places. `followup-seed.mjs` already has the helper inline (`:113`) but does not export it, so this would be a third copy of four lines — it belongs in `lib/` instead, per the derive-vs-copy point in #2624. I would fold `followup-seed`'s copy onto it in a follow-up rather than in the same PR.

abankar1:
Fixes #2932.

`set-status.mjs` is the canonical tracker write path, and it resolved "today" as the **UTC** day in two places — rejecting a legitimate date in one hemisphere and writing a wrong one in the other. Same defect as #2765, one layer up.

## 1. East of UTC, your own today is rejected (`:164`)

Run live against a sandbox tracker:

```
$ TZ=Pacific/Auckland node set-status.mjs SandboxCo Interview --on 2026-08-16
❌ --on date is in the future: "2026-08-16"
```

2026-08-16 *is* that user's today. Anyone at a positive UTC offset hits this for the first hours of their local day — the whole of a New Zealand or Japanese morning — with no workaround except passing yesterday, which then records the wrong date.

## 2. West of UTC, the logged event date is tomorrow (`:512`)

Pinned to an instant so it does not depend on when you read this:

```
instant 2026-08-16T01:30:00Z  (UTC day 2026-08-16)

  Pacific/Auckland     local 2026-08-16   agrees
  Europe/Berlin        local 2026-08-16   agrees
  America/New_York     local 2026-08-15   logged as 2026-08-16 = TOMORROW
  America/Los_Angeles  local 2026-08-15   logged as 2026-08-16 = TOMORROW
```

The two are each other's mirror: one UTC assumption rejects a real date in one hemisphere and invents a fake one in the other.

**Correction to the issue text:** I wrote there that this "stamps the tracker with a day that has not happened yet ... on the row itself". That is imprecise — `eventDate` is appended to `data/status-log.tsv`, not to the tracker row. It still matters, and arguably more specifically: `funnel-velocity` reads that file for time-between-stages, so a future-dated transition skews the interval it measures rather than merely looking odd in the file.

## What I deliberately did not change

`:162` also calls `toISOString()`:

```js
const roundTrips = d && !Number.isNaN(d.getTime()) && d.toISOString().slice(0, 10) === flags.on;
```

That is date **parsing** on UTC midnight, used to validate that the supplied string round-trips. It is internally consistent, and #2765 drew exactly this line when it changed only "what day is it here" and left date arithmetic on UTC. There is an assertion pinning that `2026-02-30` is still rejected.

## Why the helper is in `lib/`

`followup-seed.mjs` already carries this four-line helper inline (`:113`) but does not export it, so this would be a third copy — hence `lib/local-today.mjs`, per the derive-vs-copy point in #2624. I will fold `followup-seed`'s copy onto it in a follow-up rather than widening this PR.

`now` is injectable because the west-of-UTC window only exists for part of the UTC day, and a test that can only run inside that window is a test that mostly does not run.

## Verification

Four assertions in `set-status-tests.mjs`, two of them guards for what must **not** change:

- Auckland's own today is accepted (wall-clock safe — Auckland is always at or ahead of the UTC day).
- A genuinely future date (`2099-01-01`) is **still** rejected. Without this, deleting the future check entirely would pass the first assertion.
- `localToday()` resolves the local day west of UTC, probed at a pinned instant under `TZ=America/New_York`.
- An impossible calendar date (`2026-02-30`) is still rejected, pinning that the UTC round-trip parse was left alone.

Checked for discrimination: **reverting `set-status.mjs` alone fails the east-of-UTC assertion and leaves all three guards passing.** The west-of-UTC probe exercises `lib/local-today.mjs` directly, so it asserts the helper is correct rather than asserting anything about set-status.

`set-status-tests.mjs`: 86 pass / 0 fail. `node test-all.mjs --quick`: 3935 pass / 0 fail (the 2 warnings are pre-existing and environmental — no user data, no Playwright browser).

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed `--on` date handling to use the local calendar date across time zones.
  - “Today” is now accepted correctly even when UTC falls on a different date.
  - Future dates and invalid calendar dates continue to be rejected.
  - Status log timestamps now reflect the local date.
  - Date validation now behaves consistently for locations both ahead of and behind UTC.


abankar1:
The red `test (windows-latest)` here is **#2777, not this PR**. macOS and Ubuntu are green, and my own assertions pass on all three:

```
test (windows-latest)  ✅ set-status-tests.mjs runs OK
test (ubuntu-latest)   ✅ set-status-tests.mjs runs OK
test (macos-latest)    ✅ set-status-tests.mjs runs OK
```

What fails on Windows is `agent-inbox-tests.mjs`, in the shape #2777 describes:

```
❌ item-29 exited 1: EPERM, Permission denied: ...\agent-inbox.md.lock
```

That is the post-#2838 shape — a non-zero exit rather than a silent loss — and it is unrelated to `set-status.mjs`, which touches neither the inbox nor `pipeline-lock`. Happy to re-run once #2777 settles.

abankar1:
Rebased onto `main` after #2901 landed — it touched both files this PR touches, so the two had to be reconciled by hand.

**One interaction worth flagging, because this PR causes it.** Three tests in `set-status-tests.mjs` compute the expected ledger date themselves:

```js
const today = new Date().toISOString().slice(0, 10);
```

Two are #2901's new `--source` tests; the third predates it. Once this PR makes `set-status` write the **local** day, those assertions compare a local date against a UTC one and fail for whichever hours the two differ — green when CI happens to run mid-UTC-day, red otherwise. A test that only passes during part of the UTC day is the kind of intermittent I do not want to hand anyone.

All three now use `localToday()`, so the expectation and the code under test agree by construction rather than by clock luck.

`set-status-tests.mjs`: 92 passed / 0 failed (86 mine + #2901's 6, all still green). `node test-all.mjs --quick`: 4030 passed / 0 failed.

(The red `Analyze (go)` check here is collateral from the GitHub outage earlier — the log is `HttpError: No server is currently available` from CodeQL's own status reporting, not a finding. This PR touches no Go. I do not have rerun rights; it should clear on the next push or a maintainer rerun.)

santifer:
Resolved the `update-system.mjs` conflict maintainer-side (take-both: `lib/gemini-node-floor.mjs` from today's main plus your `lib/local-today.mjs`) — the conflict came from my own merge ordering this afternoon, so the fix was mine to make rather than a rebase to ask of you. Full suite green locally on the merged result (4265/0). Review of the change itself follows once CI confirms on this base: the local-today primitive with the UTC/local line drawn where #2765 drew it is exactly right, and migrating the tests off toISOString kills a pass-only-part-of-the-day flake as a bonus.

santifer:
Merged 🚀 Three for three today, @abankar1, and this one closes a recurrence: #2765 fixed the UTC-day bug in followup-seed and the canonical write path had the same defect in both directions — your east-of-UTC live repro (a New Zealand user's own today rejected as future) named the half nobody had measured. lib/local-today.mjs with the arithmetic/display line drawn where #2765 drew it is the right shape, and migrating the tests off toISOString killed a passes-only-part-of-the-day flake as a bonus. The trio of shared primitives you left behind today (ascii-fold, local-today, and the folded trust-validator) is the kind of consolidation that outlives the fixes.

PR #2938, opened 2026-08-15
fix(verify-portals): stop the slug repair path adopting another employer's board (#2937)

### LINKED ISSUE #2937: verify-portals: fix-slugs auto-writes another employer's board, because the repair path probes suffixes built on the company's first word
### Code of Conduct

### Existing issues

### What happened?

`verify-portals.mjs` auto-repairs a tracked company whose ATS slug has 404'd by cross-probing slug variants and attaching the first live board it finds as `suggested`. `fix-slugs.mjs --fix` writes that suggestion into `portals.yml`. There is no identity check anywhere on that path: `discoverAlternates` returns the first candidate that answers with at least one job, and nothing confirms the board belongs to the company being repaired.

That would be tolerable if the candidate set only held forms of the company's own name. It does not. `deriveSlugCandidates` builds suffix variants on two bases, and one of them is the **bare first word**:

```js
const SLUG_SUFFIXES = ['ai', 'tech', 'io', 'hq', 'labs'];
const bases = [words.join(''), words[0]].filter(Boolean);
for (const base of bases) {
  for (const suf of SLUG_SUFFIXES) candidates.push(`${base}${suf}`);
  candidates.push(`${base}.tech`, `${base}.io`);
}
```

Crossing a truncation with an invented suffix does not produce a variant of the company's name. It produces **a different company's name**. For a tracked company called `Nimbus Data`, the set contains `nimbusai`, `nimbustech`, `nimbusio`, `nimbushq` and `nimbuslabs`. If any of those is a real, live board owned by somebody else, it is silently adopted.

The consequence is not a wasted probe. `scan.mjs` labels every job with the `name` from the `portals.yml` entry (`company: company.name`), so after the write, another employer's postings are ingested, deduped, and tracked as this company's.

Note this combination has **no legitimate yield**. When the company name already ends in one of the suffix words (`Scale AI`, `Nimbus Tech`), `words[0] + suffix` just reproduces `words.join('')` and is deduped away. When it does not, the result is a name the company does not have. Every candidate this rule contributes uniquely is a third party's.

### Steps to reproduce

No network needed. Drive the real exported functions with a scripted `fetchJson` in which the only live board on the internet belongs to a different employer.

`repro.mjs`, in the repo root:

```js
import { deriveSlugCandidates, verifyCompanies } from './verify-portals.mjs';

console.log(deriveSlugCandidates('Nimbus Data'));

// The tracked company is "Nimbus Data". A DIFFERENT employer, "Nimbus AI",
// owns the live Greenhouse board `nimbusai`. Everything else 404s.
const LIVE = 'https://boards-api.greenhouse.io/v1/boards/nimbusai/jobs';
async function fetchJson(url) {
  if (url === LIVE) return { jobs: [{ id: 7788, title: 'VP Marketing' }] };
  const err = new Error(`HTTP 404 for ${url}`);
  err.status = 404;
  throw err;
}

const results = await verifyCompanies(
  [{ name: 'Nimbus Data', careers_url: 'https://job-boards.greenhouse.io/nimbusdata' }],
  { fetchJson },
);
console.log(JSON.stringify(results, null, 2));
```

`node repro.mjs`:

```
[
  'nimbusdata',      'nimbus-data',
  'nimbus_data',     'nimbus',
  'nimbusdataai',    'nimbusdatatech',
  'nimbusdataio',    'nimbusdatahq',
  'nimbusdatalabs',  'nimbusdata.tech',
  'nimbusdata.io',   'nimbusai',
  'nimbustech',      'nimbusio',
  'nimbushq',        'nimbuslabs',
  'nimbus.tech',     'nimbus.io'
]
[
  {
    "name": "Nimbus Data",
    "ats": "greenhouse",
    "slug": "nimbusdata",
    "url": "https://boards-api.greenhouse.io/v1/boards/nimbusdata/jobs",
    "status": "missing",
    "errorKind": "slug_gone",
    "httpStatus": 404,
    "reason": "HTTP 404 for https://boards-api.greenhouse.io/v1/boards/nimbusdata/jobs",
    "suggested": {
      "ats": "greenhouse",
      "slug": "nimbusai",
      "url": "https://boards-api.greenhouse.io/v1/boards/nimbusai/jobs",
      "status": "live",
      "jobCount": 1
    }
  }
]
```

Then hand that row to the real `computeFixes` from `fix-slugs.mjs`, which is what `node fix-slugs.mjs --fix` runs:

```js
import { computeFixes } from './fix-slugs.mjs';

const portals = `tracked_companies:
  - name: Nimbus Data
    careers_url: https://job-boards.greenhouse.io/nimbusdata
    api: https://boards-api.greenhouse.io/v1/boards/nimbusdata/jobs
`;
const results = [{
  name: 'Nimbus Data', ats: 'greenhouse', slug: 'nimbusdata', status: 'missing',
  errorKind: 'slug_gone', httpStatus: 404,
  suggested: { ats: 'greenhouse', slug: 'nimbusai', status: 'live', jobCount: 1 },
}];
console.log(computeFixes(portals, results, { dateStr: '2026-08-15' }).text);
```

Output, which is what lands in `portals.yml`:

```yaml
tracked_companies:
  - name: Nimbus Data
    careers_url: https://job-boards.greenhouse.io/nimbusai
    api: https://boards-api.greenhouse.io/v1/boards/nimbusai/jobs
    notes: "(slug migrated greenhouse->greenhouse 2026-08-15, verify-portals)"
```

The diff `fix-slugs` prints for a human to approve is `nimbusdata -> nimbusai`, which reads exactly like an ordinary rebrand.

### Expected behavior

Two sets, not one, along the split the code already implies:

- `--add` (`runAdd`) is human-in-the-loop. It prints every hit with its slug and the operator picks one. A wrong guess there costs a 404 and a line the operator rejects. That set should stay as wide as it is now, accents fix included (#2930).
- `discoverAlternates` feeds `suggested`, which `fix-slugs --fix` writes with no further confirmation of identity. That set must not contain a form derived by truncating the company's name: neither the bare first word nor any suffix built on it.

Concretely, `discoverAlternates` should probe only the untruncated forms (`nimbusdata`, `nimbus-data`, `nimbus_data`, and the suffix variants on the full joined base such as `nimbusdataai`). `Nimbus Data` then falls back to reporting the entry as unresolved, which is the correct answer: a human can still run `--add` and see `nimbusai` offered, with the company name in front of them.

The cost is that `fix-slugs` stops auto-repairing the legitimate `Stripe Inc -> stripe` shape. That is the intended trade. An unresolved entry is visible and harmless; a confidently wrong one relabels another employer's postings and looks correct while doing it.

### CLI tool

Claude Code

### OS and Node.js version

macOS 26.5 / Node 22.22.3

### Output of npm run doctor

```shell
Not relevant: the reproduction is offline and drives the exported functions directly with an injected fetchJson. No portals.yml, CV or profile is read.
```

rubicon:
## What does this PR do?

Stops `verify-portals`'s automated slug-repair path from adopting a live ATS board that belongs to a different employer. `discoverAlternates` no longer probes the suffix variants built on the company's **first word** (`Nimbus Data` -> `nimbusai`, `nimbustech`, `nimbuslabs`), because that result is attached as `suggested`, written into `portals.yml` by `fix-slugs --fix` with no further identity check, and then used by `scan.mjs` to label the other employer's postings with this company's name. `--add`, which is human-in-the-loop, keeps the wide set.

## Related issue

Closes #2937

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

---

### Why this rule specifically, and not the whole truncation

The candidate set has three kinds of entry, and only one of them is unsafe to write:

| Kind | Example for `Nimbus Data` | Relationship to the name |
|---|---|---|
| Reformat | `nimbusdata`, `nimbus-data`, `nimbus_data` | the whole name |
| Extend | `nimbusdataai`, `nimbusdata.io` | the whole name plus a token |
| Abbreviate | `nimbus` | part of the name, nothing added |
| **Coin** | **`nimbusai`, `nimbustech`, `nimbuslabs`** | **part of the name, plus a token the company does not have** |

Only the last one can name a company that is not this one. It also has no legitimate yield to lose: when the name already ends in a suffix word (`Scale AI`), `words[0] + suffix` just reproduces `words.join('')` and is deduped away, so every candidate the rule contributes *uniquely* belongs to somebody else.

**The bare first word stays on both paths on purpose.** It adds no token the company lacks, boards really are often just the brand (`Acme Corp` -> `acme`), and the existing `Diabolocom EU Discovery` fixture depends on it. A board at slug `nimbus` owned by a different `Nimbus` is still admitted; that residual is called out in #2937 rather than fixed here, because removing it would cost real repairs and is a judgement call for you, not something the reproduction demonstrates.

### TDD

Both assertions were added to section 10b of `test-all.mjs` and watched fail against unmodified `main` first:

```
❌ verify-portals first-word-suffix split wrong: probe=[...] repair=[...]   (the two sets were identical)
❌ verify-portals adopted another employer's board: {"ats":"greenhouse","slug":"nimbusai",...}
```

The second is the end-to-end case: `verifyCompanies` driven with an injected `fetchJson` in which the only live board on the internet is `nimbusai`. No network, no fixtures on disk.

### Verified after the fix

- The repair path returns the row as `missing` with **no** `suggested`.
- `--add`'s probe set still offers `nimbusai`, where an operator sees it beside the company name.
- The legitimate bare-first-word repair still resolves: `Stripe Inc` with a 404 on `stripe-inc` still suggests `stripe`.
- `Deepset -> deepsetai` (a suffix on the *full* base) is untouched, as the existing fixture asserts.

### Test counts

| | passed | failed | warnings |
|---|---|---|---|
| `main` @ 22cbe88 | 3938 | 0 | 1 |
| this branch | 3940 | 0 | 1 |

`+2` is exactly the two new assertions. The single warning is pre-existing on `main`.

### Note on overlap

#2931 also edits `deriveSlugCandidates` and section 10b of `test-all.mjs` (ASCII folding for accented names, #2930). The changes are independent in intent, but they will touch adjacent lines. Happy to rebase this on top of that one whenever it lands, in whichever order suits you.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved board matching across Greenhouse, Lever, and Ashby to prevent similarly named companies’ boards from being suggested.
  - Rejected boards when ownership cannot be verified or does not match the tracked company.
  - Added more accurate identity checks across legal designators, punctuation, accents, prefixes, and name variations.

- **Improvements**
  - Refined repair suggestions to exclude less-reliable first-word-only suffix variants.
  - Improved handling of empty or newly discovered boards.
  - Added validation for Diabolocom discovery and Lever repair URLs.

---

### Second commit: confirming ownership, not just narrowing the guess

Dropping the coined suffixes removed the candidates that are never this company's name, but it left the premise in place: nothing on the repair path ever asks who owns the board it found. The bare first word stays in the set on purpose, so `Nimbus Data` still probes `nimbus`, and if that board belongs to Nimbus AI the write still happens.

Greenhouse answers the question directly. `GET /v1/boards/{slug}` returns `{"name":"Stripe","content":""}`, so `discoverAlternates` now confirms the owner before returning a candidate. Matching is on whole tokens rather than substrings: a trailing word is the same company under a longer name (`Stripe` for `Stripe Inc`), a different word in the same position is not (`Nimbus AI` for `Nimbus Data`). Substring matching would have accepted `Apexon` for `Apex`.

The check fails closed for any ATS that publishes an owner. A missed suggestion costs an operator one `--add`, where the slug is shown beside the company name; a wrong one is written unreviewed.

Lever returns a bare posting array and Ashby returns `{jobs, apiVersion}`, so neither posting API can answer. I filed that residual as #3019 and then went and checked whether anything else could answer it, which changed the scope of this PR. See the next section.

| | passed | failed | warnings |
|---|---|---|---|
| this branch, first commit | 3940 | 0 | 1 |
| this branch, with the owner check | 3943 | 0 | 1 |

`+3` is the mismatched-owner case, the preserved brand-only repair, and the token-prefix contract. The `Diabolocom EU Discovery` and `Deepset` fixtures both still resolve.

---

### Third commit: the other two providers can answer after all

I opened #3019 for Lever and Ashby on the basis that their posting APIs expose no owner. That part is true, and I verified it against the live endpoints rather than assuming: Lever's `hostedUrl` only echoes back the slug that was probed, and Ashby's response is `{jobs, apiVersion}` with nothing above the job list.

Their board pages do carry it. Ashby titles the page `deepset Jobs`, Lever titles it `Diabolocom` on EU and `Lever Demo 2` on the base instance. So the repair path now reads the title and holds it to the same canonical-equality rule Greenhouse already uses, and all three providers answer the same question before a slug reaches `portals.yml`.

The page is not fetched whole. `jobs.lever.co` ships about 950KB of embedded job JSON and ignores a Range request, and pulling all of that to learn one string is the behavior the probe path already refuses elsewhere (`A liveness probe must never paginate an entire board`). `fetchTextHead` stops at 8KB and cancels the body; every title observed sits well inside that.

I folded this into this PR rather than stacking a second one, because the two halves are the same mechanism and reviewing "confirm the owner, except on two of three providers" seemed worse than reviewing the finished rule. #3019 is closed by this.

**On the renamed fixture.** `Diabolocom EU Discovery` becomes `Diabolocom`, which is the name `templates/portals.example.yml` actually carries. The suffix existed only in the test, and with it the fixture asserted that a prefix-only match may be auto-adopted, which is precisely what this PR stops doing on Greenhouse. Keeping it would have left the rule inconsistent across providers. It still exercises the same path: a 404 on the tracked board, discovery finding the live EU Lever board.

| | passed | failed | warnings |
|---|---|---|---|
| with the Greenhouse owner check | 3943 | 0 | 1 |
| rebased on current main | 4203 | 0 | 2 |
| with Lever and Ashby | 4205 | 0 | 2 |

Both warnings are present unchanged on the base commit. The suite now injects `fetchText`, so it no longer reaches `jobs.lever.co` or `jobs.ashbyhq.com` for real; I confirmed that by making the default text fetcher throw and checking the suite still passes.

### DIFF
--- a/providers/_http.mjs
+++ b/providers/_http.mjs
@@ -54,6 +54,44 @@ export async function fetchJson(url, opts = {}) {
   return fetchWithTimeout(url, opts, (res) => res.json());
 }
 
+/**
+ * Fetch only the head of a text response.
+ *
+ * Board landing pages carry the owner's name in <title>, but the page itself can
+ * be a megabyte of embedded job JSON (jobs.lever.co ships ~950KB and ignores a
+ * Range request). Reading the whole thing to learn one string would be exactly the
+ * "slow and rude to the careers site" behavior the probe path avoids elsewhere, so
+ * this stops at maxBytes and cancels the body.
+ *
+ * @param {string} url
+ * @param {{maxBytes?: number}} [opts]
+ * @returns {Promise<string>} The first maxBytes of the body, decoded as UTF-8.
+ */
+export async function fetchTextHead(url, opts = {}) {
+  const maxBytes = opts.maxBytes ?? 8192;
+  return fetchWithTimeout(url, opts, async (res) => {
+    const reader = res.body?.getReader?.();
+    if (!reader) return String(await res.text()).slice(0, maxBytes);
+    const chunks = [];
+    let total = 0;
+    try {
+      while (total < maxBytes) {
+        const { done, value } = await reader.read();
+        if (done) break;
+        chunks.push(Buffer.from(value));
+        total += value.length;
+      }
+    } finally {
+      try {
+        await reader.cancel();
+      } catch {
+        /* body already closed */
+      }
+    }
+    return Buffer.concat(chunks).toString('utf8');
+  });
+}
+
 export async function fetchText(url, opts = {}) {
   return fetchWithTimeout(url, opts, (res) => res.text());
 }
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -5114,7 +5114,7 @@ tracked_companies:
 console.log('\n10b. Portal slug validator');
 
 try {
-  const { deriveSlugCandidates, parseAtsSlug, verifyCompanies, classifyFetchError, boardIdentityMatches } =
+  const { deriveSlugCandidates, parseAtsSlug, verifyCompanies, classifyFetchError, boardIdentityMatches, boardTitleOwner } =
     await import(pathToFileURL(join(ROOT, 'verify-portals.mjs')).href);
 
   const slugs = deriveSlugCandidates('Acme Corp!');
@@ -5183,7 +5183,7 @@ try {
   // It is not fine for discoverAlternates, whose `suggested` fix-slugs --fix
   // writes into portals.yml unreviewed. The bare first word stays on BOTH sets
   // on purpose: it adds no token the company lacks, and boards really are often
-  // just the brand (the "Diabolocom EU Discovery" row below relies on it).
+  // just the brand (the "Diabolocom" row below relies on it).
   const probeSet = deriveSlugCandidates('Nimbus Data');
   const repairSet = deriveSlugCandidates('Nimbus Data', { firstWordSuffixes: false });
   const coined = ['nimbusai', 'nimbustech', 'nimbusio', 'nimbushq', 'nimbuslabs', 'nimbus.tech', 'nimbus.io'];
@@ -5220,7 +5220,7 @@ try {
 
   // The other half of #2937. The bare first word stays in the repair set on
   // purpose, so "Nimbus Data" still probes `nimbus`, and narrowing the set
-  // further would break the "Diabolocom EU Discovery" row below. The candidate
+  // further would break the "Diabolocom" row below. The candidate
   // is not the problem; adopting it unchecked is. Greenhouse publishes the
   // board owner's name at /v1/boards/{slug}, so the repair path can ask who
   // owns the board instead of inferring identity from the slug it guessed.
@@ -5291,6 +5291,40 @@ try {
     fail(`verify-portals boardIdentityMatches wrong on: ${JSON.stringify(identityBad)}`);
   }
 
+  // #3019: Lever and Ashby answer "whose board is this" only through the board
+  // page title. Tracked "Nimbus Data" 404s, the live Lever board `nimbus` is
+  // titled "Nimbus AI", and the repair must refuse it exactly as Greenhouse does.
+  const leverJson = async (url) => {
+    if (url === 'https://api.lever.co/v0/postings/nimbus') return [{ id: 1 }];
+    const err = new Error('HTTP 404'); err.status = 404; throw err;
+  };
+  const leverText = async (url) => {
+    if (url === 'https://jobs.lever.co/nimbus') return '<title>Nimbus AI jobs</title>';
+    const err = new Error('HTTP 404'); err.status = 404; throw err;
+  };
+  const [leverMismatch] = await verifyCompanies(
+    [{ name: 'Nimbus Data', careers_url: 'https://job-boards.greenhouse.io/nimbusdata' }],
+    { fetchJson: leverJson, fetchText: leverText },
+  );
+  if (leverMismatch.status === 'missing' && !leverMismatch.suggested) {
+    pass('verify-portals refuses a Lever board whose page title names a different company');
+  } else {
+    fail(`verify-portals adopted a mismatched Lever owner: ${JSON.stringify(leverMismatch.suggested)}`);
+  }
+
+  // The title suffix is stripped, not required: Lever titles the board with the
+  // bare name while Ashby appends "Jobs".
+  if (
+    boardTitleOwner('<title>deepset Jobs</title>') === 'deepset' &&
+    boardTitleOwner('<title>Diabolocom</title>') === 'Diabolocom' &&
+    boardTitleOwner('<title>  Lever Demo 2 jobs </title>') === 'Lever Demo 2' &&
+    boardTitleOwner('<html><body>no title</body></html>') === null
+  ) {
+    pass('verify-portals boardTitleOwner reads the owner out of a board page title');
+  } else {
+    fail('verify-portals boardTitleOwner mis-parsed a board page title');
+  }
+
   if (
     classifyFetchError({ status: 404 }) === 'slug_gone' &&
     classifyFetchError({ name: 'AbortError' }) === 'network' &&
@@ -5336,6 +5370,15 @@ try {
     if (url === 'https://api.eu.lever.co/v0/postings/diabolocom') return [{}, {}];
     const err = new Error('HTTP 404'); err.status = 404; throw err;
   };
+  // Lever and Ashby publish no owner in their posting APIs, so the repair path
+  // reads the board page title instead (#3019). Injected here so the suite never
+  // reaches the network: without this the defaults would fetch jobs.lever.co and
+  // jobs.ashbyhq.com for real, and the fixtures would pass or fail on live data.
+  const mockFetchText = async (url) => {
+    if (url === 'https://jobs.ashbyhq.com/deepsetai') return '<title>deepset Jobs</title>';
+    if (url === 'https://jobs.eu.lever.co/diabolocom') return '<title>Diabolocom</title>';
+    const err = new Error('HTTP 404'); err.status = 404; throw err;
+  };
   const results = await verifyCompanies([
     { name: 'Live', careers_url: 'https://job-boards.greenhouse.io/live' },
     { name: 'Empty', careers_url: 'https://job-boards.greenhouse.io/empty' },
@@ -5345,8 +5388,8 @@ try {
     { name: 'Off', enabled: false, careers_url: 'https://job-boards.greenhouse.io/live' },
     { name: 'Lever Live', careers_url: 'https://jobs.lever.co/acme-lv' },
     { name: 'Lever EU Live', careers_url: 'https://jobs.eu.lever.co/acme-eu' },
-    { name: 'Diabolocom EU Discovery', careers_url: 'https://job-boards.greenhouse.io/does-not-exist-diabolocom' },
-  ], { fetchJson: mockFetch });
+    { name: 'Diabolocom', careers_url: 'https://job-boards.greenhouse.io/does-not-exist-diabolocom' },
+  ], { fetchJson: mockFetch, fetchText: mockFetchText });
   const byName = Object.fromEntries(results.map((r) => [r.name, r]));
   if (
     results.length === 8 &&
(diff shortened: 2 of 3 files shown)


justinsmorganHHC (approved):

rubicon:
Holding this one rather than asking you to merge it. It closes the rarer half of its own defect and leaves the commoner half on the write path.

The PR drops first-word-plus-suffix variants from the repair candidate set, and deliberately keeps the bare first word, on the grounds that the bare word is "a narrower risk than this change addresses." Measured against the repair set it builds, that reasoning is inverted. For "Nimbus Data":

```
repair set: ["nimbusdata","nimbus-data","nimbus_data","nimbus","nimbusdataai",...]

wrong employer's live board at 'nimbusai'  ->  suggested = undefined              (this PR's own fixture)
wrong employer's live board at 'nimbus'    ->  suggested = {slug:"nimbus", ats:"greenhouse"}
```

A company is far more likely to hold its bare brand slug than a coined one, so the surviving case is the likelier one, not the narrower one. And `fix-slugs.mjs:282` writes `r.suggested` into `portals.yml` with no identity re-check, which is exactly the harm this PR's own docstring describes.

The gate the PR does add is real: reverting `firstWordSuffixes: false` in `discoverAlternates` fails the named test, and `node test-all.mjs` on the branch is 3940 passed / 0 failed.

Two ways to finish it, either is fine:

1. Drop `words[0]` from the repair candidate set, with a test pinning that a live board at the bare first word yields no suggestion.
2. Keep the candidate, and gate `fix-slugs.mjs --fix` so it re-checks identity before writing `portals.yml`.

I lean toward (1) — it keeps the fix in one place, and the repair set is the thing that decides what gets written.

Context for why I am being fussy about this particular shape: the same class turned up independently elsewhere, where matching a company name against a truncated form of itself accepted a different employer's board as canonical. Seven distinct wrong-employer emissions reproduced there, and the bare first word was the single most productive input. A wrong canonical URL is worse than none, because a wrong URL looks correct and a search fallback is visibly a search.

rubicon:
Lifting the hold. `a5f5a7c` answers it, and by the better of the two routes I offered.

I proposed either dropping the bare first word from the repair set or gating the write. This does neither literally — it asks Greenhouse who owns the board at `/v1/boards/{slug}` and confirms the answer before the candidate is returned. That is stronger than dropping `words[0]`, because it keeps the suggestion that is usually correct and rejects the one that is not, rather than losing both.

The part I want to call out as right: `boardIdentityMatches` compares whole tokens rather than substrings, and the comment names the case that makes it necessary — substring matching accepts `Apex` for `Apexon`. That is the same defect class one level down, and it would have been easy to reintroduce while fixing this one.

Failing closed for any ATS that publishes an owner is also the correct default given the asymmetry: a missed suggestion costs an operator one `--add` with the slug shown beside the company name, and a wrong one is written unreviewed and then relabels another employer's postings.

One residual worth stating plainly, since the rule is prefix-based: a company named exactly `Nimbus` holding the `nimbus` board still matches `Nimbus Data`, because the shorter token list prefixes the longer one. That is a real remaining case, much narrower than what was there before, and I do not think it should block this — the alternative is rejecting `Stripe` for `Stripe Inc`, which would be worse in practice. Noting it so it is on the record rather than discovered later.

Lever and Ashby genuinely cannot answer the ownership question from their responses, and filing that separately as #3019 rather than pretending the gate covers them is the honest handling.

No objection from me. CI is green and there are no unresolved threads; it is behind main and will need a merge or rebase before it goes in.

santifer:
Reviewed in full, and this is the right shape: probe-only vs write-path split with the no-legitimate-yield argument written where the next person will look for it, identity confirmed at the adoption boundary instead of inferred from the guessed slug, and fail-closed with the *reason* recorded so a 429 outage never reads as "somebody else's board". The test matrix earns special mention — canonical equality with Mercury/Mercury Systems and Scale/Scale AI as real negative cases, and the Stripe-Inc-to-stripe row proving the check doesn't kill the legitimate repair it protects.

One doc nit to fix while you're in there: `ownerConfirmed`'s docstring says lever and ashby "publish no owner" and keep existing behavior, but the ATS table this same diff writes gives all three an `ownerUrl` (ashby and lever via board-page `<title>`). Looks like a leftover from an earlier greenhouse-only iteration — as written it will tell the next reader the opposite of what the code does. The `no-owner-endpoint` branch is still right as a guard for future ATS entries; the comment just shouldn't name lever/ashby as examples.

Verdict is merge once two mechanical things clear, and the first one is on me: #2931 (accent folding in `deriveSlugCandidates`) landed first, so this needs a rebase over it — same file, same function, textual conflict likely. Sorry for the ordering. After the rebase your CI runs on today's main and I'll land it. Reproduced #2937 end to end myself before this review, for the record: `nimbusai` in the candidate set, adopted as `suggested` when it's the only live board.

santifer:
Approved on substance, @rubicon: canonical equality with the ambiguous prefix routed to --add, owner confirmation on all three ATSes, and the {ok, reason} split are exactly right. But main moved under you while this flew (not your fault): the ascii-fold work from #2930 landed `lib/ascii-fold.mjs`, and your `canonicalNameTokens` now duplicates that folding inline, minus the non-decomposing table (ı, ø, ß, đ...) that module carries.

For the rebase: have canonicalNameTokens consume `asciiFold` instead of its own normalize. One ordering caveat: do your `&` handling BEFORE the fold ('Hims & Hers' -> 'hims and hers', 'AT&T' -> 'att'), because asciiFold replaces symbols with spaces, so folding first would turn AT&T into two tokens. Your article/designator trimming stays yours on top. You get the Işık/Møller class for free.

Rebase, keep CI green, and this merges: the decision is made.

santifer:
The rebase delivered exactly what was asked, @rubicon: canonicalNameTokens now delegates to lib/ascii-fold.mjs instead of duplicating it, and the reasoning you added (folding & to 'and' BEFORE the fold, punctuation 'delete' argued against the L'Oréal and AT&T cases) is better documentation than the original ask. Canonical equality with the ambiguous prefix routed to --add where a human judges it: exactly right for a path that writes portals.yml. Merged! 🚀

PR #2941, opened 2026-08-15
fix(web): fence agent CLIs at the spawn boundary

nikolaysm:
## What does this PR do?

Applies per-worker permissions to every agent CLI at the single spawn boundary (`spawnHeadlessCli`), instead of only to Claude in `/api/run`. Codex gains a real OS sandbox; runtimes with no verified mechanism are passed through byte-identical and reported as unfenced rather than looking restricted.

## Related issue

Refs #2507

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] If this is a new feature or architecture change, I opened an issue first (#2507)
- [x] I ran `node test-all.mjs` and all tests pass (3938 passed, 0 failed)
- [x] My changes respect the [Data Contract](https://github.com/santifer/career-ops/blob/main/DATA_CONTRACT.md) (system layer only)

---

### Approach

#2507 asks for "a per-CLI policy returning extra args plus an honest `enforced` flag". This applies it at `spawnHeadlessCli` rather than per route, for the reason that file already gives about closing stdin: it is the only spawn path, "so the fix can't drift". The fencing argument is **required** and its absence throws — which caught a seventh call site (`apply/drive.ts`) that scoping had missed.

Policy is split from translation. `worker-capabilities.mjs` answers "may this worker write, does it need the network" once, on both axes; `cli-fencing.mjs` translates it per CLI through one `FENCERS` table. Membership in that table *is* the answer to "can this runtime be restricted", so nothing can declare an enforcement that nothing applies.

### Empirically established (codex-cli 0.146.0, macOS)

- `-c sandbox_mode=…` overrides a `config.toml` setting `danger-full-access`; `-s`'s precedence could not be verified the same way and career-ops does not pass `--ignore-user-config`, so `-c` is used for both mode and network.
- `read-only` blocks DNS outright, and the network hatch (`sandbox_workspace_write.network_access`) applies only to `workspace-write`. `writable_roots=[]` does **not** remove write access. So a fetching worker gets a writable workspace it never uses — graded `partial` and surfaced, not hidden.
- Codex's default blocks network, so `evaluate` could not fetch a posting at all before this change.

### Claude behaviour change

`toolScopeFor` now honours the network axis, not just writes. A single `readOnly` scope granted `WebFetch`/`WebSearch` to every non-writing kind, so `pdf` declared `network: false`, was DNS-blocked under Codex, and could still fetch on Claude. `pdf` reads local files only and now gets `Read,Glob,Grep`. Verified live: a `pdf` run used only `Read`/`Glob` and produced a normal PDF.

`verifyClaudeArgs` also refuses an argv that does not deny what the record forbids. That caught a pre-existing hole: **all six** hand-written Claude argvs omitted `MultiEdit`, which `--permission-mode acceptEdits` auto-approves. Deny lists are now derived via `scopeFrom` rather than hand-written.

### Verification

web: 277 tests, `tsc --noEmit`, `npm run build`. core: 3938 passed / 0 failed. Live runs: Claude `pdf` + assistant; Codex `pdf` (true `read-only` sandbox) and `evaluate`/`research` (`workspace-write` + network), confirming the `partial` notice appears for `research` and **not** for `pdf`.

### Deliberately out of scope

- `rank-pipeline.mjs` spawns `codex exec` unfenced outside `web/`.
- Fencing mechanisms for the six runtimes not installed on this machine; they report unfenced rather than guessing.
- The `isClaude ? <inline argv> : spec.args(prompt)` cascade in four routes. The real fix moves Claude's tool flags into fencing, which fights §55.6's design (it asserts `claudeCliArgs` builds the *whole* argv) and is far larger than this.

### Scope note

`web/src/lib/apply/session.ts` is the one file here not otherwise touched by this change: `agent-interpret.ts` has no stream of its own, so its unfenced notice goes to its two callers as an `ApplyIssue`, following the `ai-interpreted` precedent beside it.

## Summary

This change adds per-worker capability enforcement at `spawnHeadlessCli` in `web/src/lib/spawn-cli.mjs`. Claude and Codex runs now receive write and network restrictions. Unsupported runtimes pass through and are reported as unfenced.

Users receive clearer errors when fencing rejects a run. Job cards and session workflows show warnings when restrictions cannot be verified. Route handlers clean up rejected runs and handle stream failures.

Claude uses capability-based tool scopes and strict MCP configuration in `web/src/lib/claude-invocation.mjs`. Codex uses validated sandbox, approval, and network arguments in `web/src/lib/cli-fencing.mjs`. Runtime probing checks installed Codex support in `web/src/lib/cli-fencing-probe.mjs`.

Tests cover policy mapping, argument validation, probing, unsupported runtimes, cleanup, and notices. The PR reports passing checks and live Claude/Codex runs.

Requested system files: `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/`. These paths exist in the repository, but the supplied change summary does not identify changes to them.


santifer:
Thanks @nikolaysm — fencing agent CLIs at the spawn boundary is the right instinct. Routed to the web lane (area:web) for the substantive review; heads-up that this one also touches a repo-root file, so I'll take a look at that part alongside the web maintainer's pass.

santifer (commented):
The architecture here is right and I want it — but main moved under you, and this now needs a rebase plus a genuine reconciliation before I can review the final shape. Let me be specific so it's actionable, not just "please rebase."

**What's right and welcome.** Fencing at the single spawn boundary (`spawnHeadlessCli`) instead of per-route is the correct place — it's the same "only spawn path, so the fix can't drift" argument that file already makes about closing stdin, and it's the #2507 direction. `cli-fencing.mjs` as ONE `FENCERS` table with an unlisted CLI failing closed (unfenced, and *reported* as unfenced rather than looking restricted) is exactly the honesty this surface needs. And the Codex sandbox-policy reasoning — network forcing `workspace-write` because `read-only` blocks DNS, verified against codex-cli 0.146.0, with the note not to "tighten" it without re-testing WebFetch — is the kind of load-bearing comment that saves the next person a silent breakage.

**The blocker: #2361 landed under you.** Two days ago I merged #2361 (safe Codex AI-search exec), which put a Codex sandbox *inline* in `explore/ai/route.ts` — temp cwd, `--sandbox read-only`, `--ephemeral`, and, importantly, **prompt-leak handling**: Codex's stdout/stderr echo the full prompt, so #2361 drains-but-never-forwards them and reads the answer from `--output-last-message`. Your PR predates that and takes the more general path (fence at the boundary via `CAPS.networkReadOnly`). On rebase these conflict on exactly this file, and simply taking yours would drop #2361's output-leak handling.

**The reconciliation I'm asking for** (and I'll do the full review once it's in):
1. Your boundary fencing is the better home for the *sandbox* — subsume #2361's inline Codex sandbox into `cli-fencing.mjs` rather than leaving two Codex-sandboxing paths.
2. But keep #2361's *output* handling. Fencing governs the spawn (argv, sandbox); the prompt-leak fix lives in the stream handler (don't forward Codex stdout/stderr; read `--output-last-message`). Your spawn-boundary work doesn't cover that, so it has to survive the rebase, not be overwritten by it.
3. `test-all.mjs` is a root file — @santifer takes that half; I stay on the `web/` side.

None of this is a knock on the work — the design is the right one, and #2361 was a narrower inline version of the same instinct. It's a timing collision, and reconciling the two is worth doing properly. Ping me when it's rebased and I'll review the whole thing against post-#2361 main.

nikolaysm:
@coderabbitai resume

santifer:
Root-side half done as promised, @nikolaysm: the test-all.mjs hunk is approved from here. It strengthens the #2185 write-scope freeze rather than loosening it: the presence guard stays fail-not-skip, KNOWN_KINDS moves to the policy file both CLIs actually read, and the anchored sandbox-flag patterns with comments stripped close the described-argv-vs-shipped-argv gap for the second CLI exactly the way the original froze it for the first. The web-side substance stays with the web maintainer (area:web), so this is half the verdict, not the merge.

nikolaysm:
Green again on 56265e4 — main moved under this branch a second time, and the conflict (an import collision in `apply/prefill/route.ts` with #3142) is resolved. All six workflows pass, including `test (windows-latest)`.

Thanks for the root-side verdict, @santifer — `test-all.mjs` is settled from my side too, nothing has changed in that hunk since you read it.

For whoever picks up the web lane (area:web), what's new since the routing comment:

- #2361 landed mid-review, so its inline Codex sandbox is reconciled rather than overwritten: fencing owns approval/sandbox/web-access for every Codex spawn, the AI-search route keeps its temp workspace, `--ephemeral`, `--output-last-message` and the drained stdout/stderr that keeps the prompt out of the stream.
- The web axis is three-valued (none / search / fetch) after measuring on codex-cli 0.146.0 that `--search` reaches the web server-side while the sandbox stays read-only — so AI search keeps the read-only sandbox #2361 gave it, and only URL-fetching workers pay for `workspace-write`.
- Two real bugs from the bot rounds, both reproduced before fixing: a spaced `-c "sandbox_mode = danger-full-access"` slipped the double-fence guard (codex trims config keys, so it was applied — an exec carrying it wrote to disk with the sandbox off), and the capability probe could hang forever when a descendant held the stdout pipe.

One housekeeping note: the branch was updated from main through the GitHub UI at one point, so it carries merge commits rather than the linear history originally asked for. Happy to rebase it flat before the web pass if that's easier to read.

santifer:
Update you should have, @nikolaysm, rather than more silence: **this one is not in the normal queue any more, and that's deliberate.**

2,459 lines across 20 files, four new modules and five API routes touched, all of it at the spawn boundary — that isn't a queue PR, it's an architectural change to the surface where the security hardening lives. The web owner's own measured rule is that PRs this size get **one dedicated session, not a slot in a batch**, and reviewing it at the pace of the rest would be pretending to have cadence rather than having it.

So: it's parked with a date (1 Sep) rather than left drifting, and you'll get a real review instead of a rushed one.

What's already cleared, so you don't wonder: **the core half is approved.** Your update to the section 55 freeze (adding `worker-capabilities.mjs` to the required set now that `KNOWN_KINDS` moved there) is correct contract maintenance and I signed it off this morning. Nothing on the core side is waiting on you.

If anything changes on your end — a rebase, a split, a question — say so here and it resets the clock in your favour, not against you.

santifer:
Before the date arrives rather than after it, @nikolaysm: **the 1 Sep parking holds, and main has moved under you a third time — do not rebase.**

`git merge-tree` against the current tip puts you in conflict in two files:

```
web/src/app/api/apply/prefill/route.ts
web/src/components/jobs/worker-card.tsx
```

Both are ours. `619a834d` (#3428, lifting the apply planner out of the prefill route) and `d3729089` (#2158, the cause-specific job-failure hint) are the only commits touching those paths since your 25 Aug push, and we merged both. You rebased for us on 20 Aug and again on 25 Aug; a third one for the same reason is not something I am going to ask for.

**I will resolve these inside the review window**, not before it. Your PR rewrites the spawn boundary, `prefill/route.ts` is one of the routes it rewrites, and reconciling a conflict in code I have not finished reading is how a merge conflict becomes a behaviour change nobody notices. The conflict resolution is part of the review, not a prerequisite for it.

So nothing is expected from you. Status unchanged: core half approved since 21 Aug, web half parked with a date, and the date is tomorrow.

If you would rather rebase it yourself — some people would rather hold their own branch — say so and I will leave it alone. That is a preference, not an obligation.

santifer:
Late, and that is on me, @nikolaysm: 1 Sep came and went without the review I parked this for. Here it is.

The web half is approved. Fencing every worker at the single spawn boundary, with spawnHeadlessCli refusing to start without a capability set and FENCERS as the one table that fails closed on an unlisted runtime, is the shape I asked for on 20 Aug. The two bugs you reproduced and fixed (the spaced sandbox_mode override slipping the guard, the probe hanging on a held stdout pipe) are exactly the findings I want in this file's history. Checks are green on the current head.

One behaviour change I am accepting on purpose, and it goes in the release notes: every non-writing worker now runs with --strict-mcp-config, not only pdf. That is the trade #2507 asks for.

What is left is integration, and it is mine. Main moved again: #3428 moved the planner spawn out of the prefill route into planner.ts, so your fenced call, ADVISOR_SCOPE and the fencingReport notice have to move with it, or the new guard refuses to start. worker-card.tsx is the small one: #2158's hint and your notice both stay.

I resolve that in the merge, as I said on 31 Aug; you do not owe a third rebase. If you would rather carry it yourself now that the target is precise, say so here and the branch stays yours. No further review is pending on my side: the next event on this PR is the merge.

santifer:
@coderabbitai review

santifer (approved):

PR #2948, opened 2026-08-16
Closes #2947: tier Source-of-Truth boundary, add story-provenance-check.mjs

### LINKED ISSUE #2947: feat(agents/story-bank): tier the Source-of-Truth boundary — derived files need provenance, not cv.md-equivalent trust
## The defect

`AGENTS.md`'s Source-of-Truth Boundary lists `interview-prep/story-bank.md` and `interview-prep/{company}-{role}.md` as "the user's own STAR stories and prep notes — **same trust level as `cv.md`**."

That equivalence doesn't hold in practice. `cv.md` is user-authored. `story-bank.md` is *accumulated* — commonly extracted from past prep documents, which are themselves AI-written mappings of a user's experience onto a specific JD's language. Nothing in the current boundary requires a story-bank claim to trace back to a user-authored file. The result is an unguarded channel:

> JD-shaped phrasing in a prep doc → absorbed into story-bank.md as a standalone story → treated as cv.md-equivalent fact → surfaces in CV bullets, cover letters, form answers, interview talk tracks

**Reproduced case (anonymized):** `cv.md` described a role only as general cross-functional coordination — no scale figure. A prep document written for an unrelated posting that emphasized multi-site logistics rendered the same role with an invented number attached. story-bank absorbed that rendering as a standalone entry. A later, unrelated prep document cited it as fact. In further reuse it drifted to an even larger, still-invented figure. Each layer looked sourced because it cited the layer above it — none of them traced back to the one user-authored file. A human caught it by memory; nothing in the system would have.

## Why this is harder than "just trace it back to cv.md"

Two complications, both real:

1. **`cv.md` isn't a clean ground truth either.** A CV is polished by definition — wording gets optimized, scope gets rounded, phrasing gets tuned per application. "Traces to cv.md" verifies *consistency with the CV*, not *truth*.
2. **"Ask the user to confirm" has a floor.** For some details (a job from years ago, a scale figure never actually tracked), the user genuinely doesn't know. A confirmation prompt that leads with the unverified number invites a guess — and a confirmed guess is worse than an honest unknown, because it launders the guess into a "verified" fact.

A binary verified/unverified flag can't represent this. There needs to be a durable third state — *cannot be confirmed* — that doesn't decay into "verified" through repetition or a leading question.

## Existing precedent in this repo (skeleton to reuse, not to copy wholesale)

`jd-skill-gap.mjs` already solves an adjacent problem with a three-bucket classification against `cv.md`: `existing` (named skill) / `supportedByResume` (prose-demonstrated) / `gap` (no trace at all). That shape is the right skeleton — but it only has one source file and a binary "trace or don't." This problem needs a fourth state that `jd-skill-gap.mjs` has no equivalent for: not "no trace" but "asked, and genuinely can't be confirmed."

`check-table-freshness.mjs`'s hard/soft severity split (`expired` blocks, `review-due` warns) is a usable pattern for how the *checker script* reports findings at different severities, but doesn't address the provenance-tiering problem itself.

## Proposal

1. **Tier the Source-of-Truth list in `AGENTS.md`**: primary/user-authored (`cv.md`, `config/profile.yml`, `article-digest.md`) vs. derived/accumulated (`story-bank.md`, `interview-prep/{company}-{role}.md`). Derived files may supply narrative structure and phrasing; any quantified claim, scale figure, or scope-of-responsibility claim originating there must trace to a primary file or carry an explicit provenance marker.
2. **`story-provenance-check.mjs`** — zero-LLM, same house style as `jd-skill-gap.mjs`/`cv-title-check.mjs`. Scans `story-bank.md` for numeric claims (`\d+%`, `\d+\+`, "N hours → M hours" patterns, headcount/scale figures) that don't appear in `cv.md`. Classify into the same four-bucket shape `jd-skill-gap.mjs` established, extended: `existing` (in cv.md) / `supportedByResume` (cv.md prose supports it) / `derived-unverified` (only appears in a prep doc, never confirmed) / `user-cannot-confirm` (explicitly asked, user could not confirm).
3. **Provenance field per story-bank entry**: `source: cv.md | user-stated YYYY-MM-DD | derived-unverified | user-cannot-confirm`.
4. **The confirmation UX is the part that must not just be "ask the user" (mandatory design constraint, not optional polish):** when the checker flags a `derived-unverified` claim, the mode/agent surfacing it to the user must not lead with the unverified number as if confirming or denying it were the only options. Present the claim plainly and require an explicit, unbiased choice among distinct outcomes — e.g. (a) confirm the claim is accurate as stated, (b) provide the correct figure, (c) mark it narrative-only / not a quantified claim, (d) "I don't know" → locks the entry as `user-cannot-confirm`, durably, not re-prompted into a guess on reuse. A `user-cannot-confirm` entry must never silently decay back into being treated as verified through repeated citation — every consumer of story-bank.md (CV generation, cover letters, interview prep) must treat it as narrative texture only, never as a quantified claim in interview-facing output where a follow-up question would probe it.

## Test plan

- Fixture: a story-bank numeric claim present in `cv.md` → `existing`.
- Fixture: a story-bank numeric claim cv.md's prose supports without the exact figure → `supportedByResume`.
- Fixture: a numeric claim only present in a prep doc, never in cv.md, unconfirmed → `derived-unverified`.
- Fixture: an entry explicitly marked `user-cannot-confirm` → stays in that state across repeated `--summary` runs, never silently reclassified.

Related: RFC-adjacent design work already in this repo for provenance/staleness (`check-table-freshness.mjs`) and single-source classification (`jd-skill-gap.mjs`) — this issue composes both patterns rather than inventing either from scratch.

Schlaflied:
## Summary

Closes #2947.

`AGENTS.md`'s Source-of-Truth Boundary documented `interview-prep/story-bank.md` and `interview-prep/{company}-{role}.md` as having "the same trust level as `cv.md`". In practice `cv.md` is user-authored while `story-bank.md` is *accumulated* — commonly extracted from past interview-prep documents, which are themselves AI-written mappings of the user's experience onto a specific job posting's language. Nothing enforced that a story-bank claim traced back to a user-authored file, which is the unguarded channel the issue's repro describes: JD-shaped phrasing in a prep doc → absorbed into story-bank.md as a standalone story → treated as cv.md-equivalent fact → surfaces in CV bullets / cover letters / talk tracks → drifts further on reuse.

### 1. AGENTS.md — tiered Source-of-Truth Boundary

Split the file list into two explicit tiers:

- **Primary / user-authored** (full trust, unchanged): `cv.md`, `article-digest.md`, `config/profile.yml`, `modes/_profile.md`, `writing-samples/`, `modes/_custom.md`, `voice-dna.md`.
- **Derived / accumulated** (narrative/phrasing trust only): `interview-prep/story-bank.md`, `interview-prep/{company}-{role}.md`.

Derived files may still supply narrative structure and phrasing freely. But any quantified claim, scale figure, or scope-of-responsibility claim originating there must now trace to a primary file, or carry an explicit `**Provenance:**` marker on that story-bank entry — it is no longer automatically cv.md-equivalent the way the old wording implied.

Also documents, as a binding design constraint (not optional polish): the **confirmation UX invariant** — when a `derived-unverified` finding is surfaced to the user by a future consuming workflow, the prompt must not lead with the unverified number as if confirm/deny were the only options. It must offer four distinct, unbiased outcomes (confirm as stated / provide the correct figure / mark narrative-only / "I don't know" → `user-cannot-confirm`), and a `user-cannot-confirm` marker must never decay back into "verified" through repeated citation or a later re-scan. `modes/_shared.md`'s interview-prep row was updated to match.

### 2. story-provenance-check.mjs — new zero-LLM checker

Read `jd-skill-gap.mjs` first and followed its house style: regex-based extraction (no LLM call), JSON output by default / `--summary` for a human-readable table, `--self-test`, read-only (never edits `story-bank.md` — same posture as every other zero-LLM checker in this repo).

**Why four buckets instead of copying jd-skill-gap.mjs's three (`existing` / `supportedByResume` / `gap`):** jd-skill-gap.mjs only ever needs a binary trace/no-trace answer, because nothing is ever auto-added to `cv.md` either way. This problem needs a state that has no equivalent there: not "no trace found (yet)" but "asked, and the user genuinely could not confirm it." A scale figure from a job years ago may simply be unknowable — collapsing that into the same bucket as "not yet checked" would let a future re-scan or a leading confirmation prompt silently launder a guess into a "verified" fact. So the buckets are:

- `existing` — the claim's number(s) appear in `cv.md`, OR the entry's Provenance field explicitly reads `user-stated YYYY-MM-DD` (an explicit marker satisfies AGENTS.md's "trace to a primary file OR carry an explicit provenance marker" language even without cv.md carrying that exact precision).
- `supportedByResume` — the number itself isn't in `cv.md`, but `cv.md`'s prose describes the same underlying fact/role closely enough (word-overlap heuristic in a window around the claim) that it reads as plausible-adjacent, just unverified at this precision.
- `derived-unverified` — the number appears only in `story-bank.md`, no cv.md trace, no provenance marker.
- `user-cannot-confirm` — the entry's Provenance field is explicitly set to this value. **Hard override**, checked first, wins over the heuristic regardless of what it would otherwise conclude, and is never itself overridden back — the durable state the issue asked for.

**Numeric patterns covered** (documented in the file's header as deliberately non-exhaustive, matching jd-skill-gap.mjs's own stated design stance that under-extraction is recoverable but over-extraction is not): `\d+%`, `\d+\+ <word>` (e.g. "500+ employees"), hour-range ("8 hours to 2 hours" / "8 hrs → 2 hrs"), hyphenated scale ("15-person team"), and noun-scale ("50 students", "200 employees"). Known gaps are called out explicitly: spelled-out numbers, currency (intentionally left to `verify-cv-facts.mjs`, which already owns that gate for generated documents), non-ASCII digit scripts, and bare numbers with no adjacent noun.

### 3. Provenance field convention

New per-entry field on `story-bank.md` blocks (same `### [Theme] Title` / `**Label:** value` shape `match-star.mjs` already parses — deliberately a *new* label, not overloading the existing `**Source:**` field which already means something else there):

```
**Provenance:** source: cv.md
**Provenance:** user-stated YYYY-MM-DD
**Provenance:** derived-unverified
**Provenance:** user-cannot-confirm
```

Absent field defaults to `derived-unverified` (the safe default — unmarked numeric claims get no special trust).

### Explicitly out of scope (per the issue)

Building the interactive confirmation workflow that asks the user and writes the Provenance marker back to `story-bank.md` is **not** part of this PR. This PR ships the read-only checker plus the documented convention and invariants; a future PR can build the write-path against this foundation. The invariant is documented prominently in both `AGENTS.md` and the script's own header comment so a future implementer can't miss it.

## Test plan

- [x] Fixture: numeric claim present in `cv.md` → `existing`
- [x] Fixture: `cv.md` prose supports the same underlying fact without the exact figure → `supportedByResume`
- [x] Fixture: numeric claim only in `story-bank.md`, never in `cv.md`, no marker → `derived-unverified`
- [x] Fixture: entry explicitly marked `user-cannot-confirm` → stays `user-cannot-confirm` even though the numeric heuristic alone would suggest `derived-unverified`
- [x] Fixture: re-running the checker on the same `user-cannot-confirm` entry does not reclassify it (asserted via two independent classification runs compared for byte-identical output, plus a direct membership check on the second run)
- [x] Fixture: explicit `user-stated YYYY-MM-DD` marker promotes an otherwise-unverified claim to `existing`
- [x] `story-provenance-check.mjs --self-test` — 17/17 passed, wired into `test-all.mjs`
- [x] `node test-all.mjs --quick` — 3665 passed, 0 failed, 1 warning (pre-existing/expected: `cv-sync-check.mjs exited with error (expected without user data)` — this repo ships no `cv.md` fixture by design, unrelated to this change)

## Summary by CodeRabbit

- **New Features**
  - Added source tracking for quantified claims in interview-preparation stories.
  - Claims can be classified as verified, resume-supported, unverified, or user-unconfirmable.
  - Added summary and JSON reporting for provenance checks.

- **Bug Fixes**
  - Unverified claims are now clearly identified instead of being treated as confirmed.
  - Workflows now support confirmation, correction, narrative-only, and cannot-confirm outcomes.

- **Tests**
  - Added automated self-tests for claim extraction and classification.


Schlaflied:
Pushed a fix for both Major findings in `e93d702`.

**Finding 1 — number matching wasn't scoped to metric/context**

`classifyStoryBank()` used to check `existing` with a plain `cvNumbers.has(v)` set-membership test, so any number appearing *anywhere* in `cv.md` counted as verification — e.g. a story claiming a "15-person team" would false-positive against an unrelated `cv.md` line like "15 years of experience."

Fixed by reusing the same context-overlap mechanism the file already uses for the `supportedByResume` bucket:
- `buildCvNumberContexts(cvText)` maps each number found in `cv.md` to the list of `contextWords()` around each of its occurrences.
- `hasScopedNumberMatch(claimWords, cvContextLists)` requires the claim's own context words to overlap with at least one of those cv.md occurrences before the number counts toward `existing`.

No parallel heuristic was invented — same `contextWords()`/window mechanism, just applied to `cv.md`'s own number occurrences instead of a flat set.

Added a regression fixture: a claim ("15-person team") whose number coincidentally also appears in `cv.md` in an unrelated sentence ("15 years of ... experience") — asserted to NOT land in `existing` (it now correctly falls through to `supportedByResume` via the unrelated "team" word overlap elsewhere in `cv.md`, which is the honest signal).

**Finding 2 — explicit provenance markers evaluated too late**

`user-stated YYYY-MM-DD` and `source: cv.md` were being checked *after* the numeric/context heuristic, so a story carrying either marker could still get misclassified into `supportedByResume` whenever the heuristic below found only partial context overlap — even though the marker already asserts the claim is confirmed.

Moved both checks to run immediately after the `user-cannot-confirm` hard override and before any heuristic, matching CodeRabbit's suggested diff. `USER_STATED_RE` already existed in the file (was just being checked too late) — reused it as-is, no second parallel implementation.

Added regression tests for both marker forms: a claim with `user-stated YYYY-MM-DD` (and separately `source: cv.md`) plus a number *not* in `cv.md`, but with context words that do overlap `cv.md` prose (a case that would previously have been redirected into `supportedByResume`) — both now correctly resolve straight to `existing` and are asserted to NOT appear in `supportedByResume`.

**Test results**

- `node story-provenance-check.mjs --self-test`: 22 passed, 0 failed (was 15 before this change; added 7 regression assertions across the two findings)
- `node test-all.mjs --quick`: 3665 passed, 0 failed, 1 pre-existing warning unrelated to this file

Schlaflied:
Fixed: added the control-case assertion for both marker fixtures (`user-stated YYYY-MM-DD` and `source: cv.md`). Classifying the identical story body with the `**Provenance:**` line removed now asserts it lands in `supportedByResume` — proving the context-overlap heuristic really would fire absent the marker, so the marker is what's actually overriding it in the positive assertion, not a coincidental absence of overlap.

`node story-provenance-check.mjs --self-test`: 24/24. `node test-all.mjs --quick`: 3665 passed, 0 failed.

santifer:
The distinction you drew is the whole PR and it is right, @Schlaflied: `cv.md` is authored, `story-bank.md` is ACCUMULATED, commonly out of prep documents that are themselves AI-written mappings of the user onto a JD's language. Treating them as one trust level meant a scale figure invented once to match a posting could be absorbed as a standalone fact, cited as ground truth by a later unrelated prep doc, and drift further on each reuse, with nothing forcing it back to a primary file.

Three things I checked rather than took on trust. The 🔴 surface is genuinely small: AGENTS.md +6/-1 and one table cell in `modes/_shared.md`, and both TIGHTEN the boundary rather than loosen it. `story-provenance-check.mjs` imports only `readFileSync` and `existsSync`, so it is read-only by construction and cannot touch a user file even by accident, which is a stronger guarantee than documenting the intent. And it is registered in SYSTEM_PATHS, so it will actually reach installs instead of working only in your checkout.

The paragraph I would single out is the confirmation UX invariant, because it is an insight about how anti-fabrication fails in practice rather than in theory: leading with the unverified number as if confirm and deny were the only options invites a guess, and a confirmed guess is worse than an honest unknown because it launders the guess into a verified fact. `user-cannot-confirm` never decaying back through repeated citation is the other half of that, and it is the half most systems forget. Merged 🚀

PR #2950, opened 2026-08-16
Closes #2949: ROI-based salary-negotiation talking-point generator

### LINKED ISSUE #2949: feat(followup): ROI-based salary negotiation script — anchor the ask in a verified quantified achievement, not just market rate
## What

A zero-LLM script that generates a salary-negotiation talking point anchored in a quantified, verified achievement from `story-bank.md`, instead of (or alongside) pure market-rate comparison.

## Why

`salary-gap.mjs` already tells a candidate where an offer sits relative to market. What it can't do is give them something to *say* in the room beyond "market rate is higher than your number" — a comparison, not an argument. A candidate with a real, quantified accomplishment (e.g. "cut a recurring process from 8 hours to 2") has a stronger anchor available: translate that accomplishment into an estimated dollar value, and use *that* as the basis for the ask, not just a market percentile.

## Why this is safe to build now

Pulling a number from `story-bank.md` into a negotiation conversation is a higher-stakes place for a fabrication to land than an interview answer — if a candidate states a dollar figure built on an unverified or invented claim and the employer probes it, the damage is to credibility and the offer itself, not just to one interview answer. So this script must not source figures from `story-bank.md` indiscriminately.

**Scope for v1:** only use a `story-bank.md` numeric claim when the *exact same number* also appears in `cv.md` — the narrowest, safest verification available today, with zero new infrastructure. `story-provenance-check.mjs` (in progress, PR #2948) will eventually provide a richer four-tier classification (`existing` / `supportedByResume` / `derived-unverified` / `user-cannot-confirm`); once that lands, this script is a natural candidate to consume its `existing` bucket directly instead of doing its own narrower cv.md match. Not a blocking dependency for this issue — v1 should ship with the simple direct-match check standalone.

## Proposal

New script (name TBD, e.g. `negotiation-roi.mjs`), zero-LLM, house style matching `salary-gap.mjs`/`story-provenance-check.mjs` (CLI flags via `lib/cli-flags.mjs`'s `validateFlags`, JSON default / `--summary`, `--self-test`).

1. Scan `story-bank.md` for quantified claims (reuse whatever numeric-extraction regex `story-provenance-check.mjs` already built — don't reinvent it, check if it's exported/reusable).
2. Keep only claims whose number also appears verbatim in `cv.md` (the v1 verification gate described above).
3. For a time-reduction-shaped claim (e.g. "N hours → M hours", "cut processing time by X%"), compute an estimated annualized dollar value. Required inputs, all explicit and visible in the output — never silently assumed:
   - An hourly-cost or salary basis for the role in question. If not available from any user-layer file, the script must ask for it rather than guess (or accept it as a CLI flag/config value) — never invent a wage figure.
   - The frequency/recurrence of the task (e.g. "once a week" vs "daily") — same rule: read from the story if stated, otherwise require it as input, never assume.
   - Show the full calculation in the output (basis × frequency × time saved), not just the final number, so the candidate can sanity-check or challenge their own figure before using it.
4. Output: a draft talking-point paragraph structured as (a) optional market-context framing if `salary-gap.mjs` data is available, (b) the quantified achievement stated plainly, (c) the calculated value with its basis shown, (d) an explicit prompt reminding the candidate to judge for themselves whether the achievement's context (company size, industry, scale) actually transfers to the target role — the script computes the arithmetic, it does not vouch for the analogy holding.
5. Read-only / draft-only, consistent with this whole system's "never auto-submits, human always reviews" rule (AGENTS.md, Ethical Use section) — this generates a script for the user to review and use themselves, never sends or submits anything.

## Test plan

- Fixture: a story-bank claim whose number matches cv.md exactly → included, calculation shown with all inputs visible.
- Fixture: a story-bank claim whose number does NOT appear in cv.md → excluded (v1's narrow safety gate).
- Fixture: missing wage-basis input → script prompts/errors clearly rather than guessing a number.
- Fixture: a claim with no clear frequency stated → same, requires explicit input rather than assuming.

Related: `salary-gap.mjs` (market-comparison sibling), `story-provenance-check.mjs` (#2948, future upgrade path for the verification gate).

Schlaflied:
## What

Adds `negotiation-roi.mjs` — a zero-LLM script that generates a salary-negotiation talking point anchored in a quantified, **verified** achievement from `interview-prep/story-bank.md`, converted into an estimated dollar value. This is the complementary sibling to `salary-gap.mjs`: that script tells a candidate where an offer sits relative to market (a comparison); this one gives them something to *say* in the room (an argument), built from their own track record instead of pure market percentile.

Closes #2949.

## How it works

1. Parses `interview-prep/story-bank.md` via the existing `parseStories()` export from `match-star.mjs` (reused, not reinvented) to get each story's Situation/Task/Action/Result/Reflection text.
2. Extracts quantified claims from that text with two narrow, documented regex shapes:
   - time-reduction: `"N hours ... to M hours"` (arrow / "down to" / bare "to"; the second unit may be omitted and inherits the first — covers the issue's own example, `"cut a recurring process from 8 hours to 2"`)
   - percent-reduction: `"cut/reduced/decreased/dropped/shaved ... by X%"`
   Deliberately not exhaustive — documented in a comment at the top of the file, same spirit as `jd-skill-gap.mjs`'s extraction notes.
3. Runs every extracted claim through the ROI calculation and draft-paragraph builder, gated by the verification step below.

## v1 safety gate (the main design decision in this PR)

Pulling a number from `story-bank.md` into words a candidate will say out loud in a negotiation is a higher-stakes place for a fabricated/unverified claim to land than an interview answer — if challenged and it doesn't hold up, the damage is to credibility and the offer, not just to one interview answer.

So: **a claim is used only if the same number, with a matching unit, also appears verbatim in `cv.md`.** If it doesn't, the claim is **excluded from output entirely** — not flagged, not included with a caveat. This is intentionally the narrowest possible check: direct substring match against `cv.md`, no fuzzy scoring, no new infrastructure, no external dependency.

A richer four-tier provenance classifier for `story-bank.md` claims is in progress on a separate, unmerged branch (`story-provenance-check.mjs`, PR #2948 — `existing` / `supportedByResume` / `derived-unverified` / `user-cannot-confirm`). This PR does **not** depend on or import from that branch. Once #2948 lands, `negotiation-roi.mjs` would be a natural candidate to consume its `existing` bucket directly instead of doing its own narrower `cv.md` match — noted as a forward-looking upgrade path in the file's header comment, not implemented here.

## Never-guess-wage-or-frequency invariant

Converting a time-reduction claim into a dollar figure needs two more inputs: an hourly wage basis and how often the task recurs. Both are resolved in the same order — **story-bank text first, explicit CLI flag second, and never silently assumed:**

- Wage: read from `"$X/hr"` phrasing in the story text (`resolveWage`), or `--wage <number>`.
- Frequency: read from phrasing like `"weekly"` / `"every day"` in the story text (`resolveFrequency`), or `--frequency <daily|weekly|biweekly|monthly|quarterly|annually>` / `--occurrences <N>`.

If either is missing, the claim is reported in a separate `uncalculable` bucket with a specific reason (`"missing hourly-wage basis..."` / `"missing task frequency..."`) — never assigned a default. When both resolve, the formula is:

```
hoursSavedPerOccurrence × hourlyWage × occurrencesPerYear = annualValue
```

...and every component (plus its source: story text vs. CLI flag) is shown in the output, so the candidate can sanity-check or challenge their own figure before using it — per the issue's requirement that the full calculation be visible, not just the total.

## Output

JSON by default, `--summary` for human-readable. Each calculable entry carries the achievement sentence (quoted verbatim from the story, never invented phrasing), the full calculation breakdown, and a draft paragraph that always ends with an explicit reminder: *this is arithmetic only — it does not judge whether this achievement's context (company size, industry, scale) transfers to the target role; that judgment is the candidate's to make.*

Read-only / draft-only throughout — the script never writes to any file and never sends or submits anything, consistent with AGENTS.md's Ethical Use section ("human always reviews before anything goes out").

## Test coverage

`--self-test` (wired into `test-all.mjs` and `update-system.mjs` `SYSTEM_PATHS` the same way `salary-gap.mjs` is) covers, with in-memory fixtures (no disk I/O, matching `salary-gap.mjs`'s self-test style):

- Extraction: the issue's own example shape (`"8 hours to 2"`, omitted second unit) and a percent-reduction claim.
- **Core safety-gate test:** a story-bank claim whose numbers (15h → 3h) never appear in `cv.md` is excluded from *both* `calculable` and `uncalculable` output — proven absent, not just unflagged.
- A verified claim (numbers match `cv.md`) with no frequency anywhere → `uncalculable`, reason mentions frequency, not silently defaulted.
- A verified claim with frequency stated in the story text ("done daily") and wage from `--wage` → fully calculable, calculation breakdown asserted field-by-field (hours saved, occurrences/year, wage, annual value), draft paragraph carries the transferability reminder.
- The same "missing frequency" claim becomes calculable once `--frequency` is supplied explicitly — proves the CLI-flag fallback path.
- A run with zero wage input anywhere → every time-reduction claim ends up `uncalculable` for a wage-specific reason, confirming no path defaults to a guessed dollar figure.

`node test-all.mjs --quick` run locally, 0 failures.

## Summary by CodeRabbit

- **New Features**
  - Added a read-only tool that turns verified interview achievements into salary-negotiation talking points.
  - Supports time-savings calculations using wage and frequency details, with JSON or human-readable output.
  - Flags unverifiable claims and explains when values cannot be calculated.
  - Validates numeric inputs and supports draft analysis without modifying files.

- **Tests**
  - Added a built-in self-test to confirm expected behavior.


Schlaflied:
Both findings fixed in 1fb753e.

**Finding 1 (the serious one)** — confirmed. `anyVariantInText()` was doing unbounded substring matching (`hay.includes(v)`), which meant `"8 hours"` matched inside `"18 hours"` and `"60%"` matched inside `"160%"`. This isn't a cosmetic bug — it's a real gap in the exact safety gate this whole PR exists to enforce: a story-bank claim could "verify" against a cv.md number it has nothing to do with, and the script would go ahead and build a draft negotiation talking point on an unverified figure. Fixed with a boundary-aware regex (`(?<![\d.])` / `(?![\d.])` around the escaped variant) so a numeric variant can no longer match as a substring of a larger number. Added self-tests for both collision cases (hour and percent) plus a sanity check that genuine matches still verify.

**Finding 2** — confirmed and fixed per the suggested diff shape: extracted a shared `parsePositiveNumberFlag()` helper using `Number()` instead of `parseFloat()` (rejects trailing garbage like `"40usd"` outright rather than silently parsing the leading `40`) plus a `Number.isFinite(value) && value > 0` guard, applied to both `--wage` and `--occurrences`. Added self-tests for malformed, negative, zero, and `Infinity` inputs on both flags.

Verified:
- `node negotiation-roi.mjs --self-test` — all assertions pass, including the new boundary/CLI-validation cases.
- `node test-all.mjs --quick` — 3665 passed, 0 failed.

Pushed to `feat/roi-negotiation-script`.

Schlaflied:
Fixed: a standalone trailing `--wage` or `--occurrences` with no value (`flagValue()` returns `undefined` for that, same as when the flag was never passed) now errors explicitly ("requires a value", exit 1) via `hasFlag()` + `flagValue()` returning undefined, instead of silently falling through to "flag absent."

Moved this check (and the file-existence checks it sits next to) so argument validation runs before the story-bank.md/cv.md existence checks — argument correctness shouldn't depend on live user data being present, and it makes the path testable via subprocess without needing real fixtures.

New self-test spawns the script with a trailing `--wage` and a trailing `--occurrences`, asserting exit 1 and the specific error message for each.

`node negotiation-roi.mjs --self-test`: OK. `node test-all.mjs --quick`: 3665 passed, 0 failed.

santifer:
The safety gate is the feature: a number leaves the story bank only if it also appears verbatim in cv.md, wage and frequency are never guessed, and an unverifiable claim is excluded rather than caveated. That is the strictest anti-fabrication design any generator in this repo has shipped with, and negotiation is exactly where it matters most. Merged! 🚀

PR #2951, opened 2026-08-16
fix(contacto): LinkedIn connection-request limit is 200 on free tier, not always 300

rkosolapov:
## What changed

`modes/contacto.md`'s LinkedIn power-move flow assumed a flat 300-character
limit for every account. LinkedIn's actual connection-request character limit
varies by account tier: **200 characters on a free account, 300 on
Premium/Sales Navigator**. Live-confirmed via the actual compose box on both
tiers.

A flat "300" assumption produces a message that gets silently truncated (or
rejected) for a free-tier account — the common case, not the exception.

## Fix

- Default to the safer 200-char budget unless the user has confirmed they're
  on Premium/Sales Navigator; count and trim to whichever limit actually
  applies.
- Updated the `AGENTS.md` skill-modes table entry for `contacto` to match
  (was also hardcoding "≤300-char").

## Testing

Pure prose/documentation change to a mode instruction file — no executable
code path, no test in `test-all.mjs` targets `contacto.md`'s content
specifically.

## Summary by CodeRabbit

* **Documentation**
  * Updated LinkedIn outreach guidance to reflect account-tier-specific connection-request limits.
  * Free accounts support up to 200 characters; Premium and Sales Navigator accounts support up to 300 characters.
  * Uses the 200-character limit by default unless a higher-tier account is confirmed.


santifer:
Welcome, @rkosolapov! A doc that confidently says 300 generates messages LinkedIn silently truncates for free-tier users: correcting it with the safer 200 default is exactly the kind of fix nobody notices until it saves them. Merged — your first contribution! 🚀

Issue #1736, opened 2026-07-09
Resume Drift - Audit & Fixes


tylerlinahan:
### Code of Conduct

- [x] I agree to follow the [Code of Conduct](https://github.com/santifer/career-ops/blob/main/CODE_OF_CONDUCT.md)

### Existing issues

- [x] I searched existing issues and this hasn't been reported yet

## Context

Thanks for fixing the `modes/_custom.md` read path in public commit `76b45a8`. This is a sanitized follow-up from an integration audit that originally pointed at the "write-only `_custom.md`" risk pattern.

As of my 2026-07-09 check, public `origin/main` was at `e9bacc4`. Some audit findings appeared fixed or stale at that commit, so this issue focuses only on the parts that still looked relevant.

## Original audit summary

The original audit was about configuration drift: formatting and content preferences were not consistently retained across later runs, especially in batch work.

The audit found five root causes:

1. **Corrections were stored in fragile places.** Durable house rules belonged in user-layer files, but some behavior still depended on system-layer files that updates could replace.
2. **`modes/_custom.md` existed but was not consistently read.** The file could receive persistent house rules, but major execution paths did not bind it back into the prompt.
3. **Resume generation treated concrete template defaults as authoritative.** Several paths described specific visual defaults such as competency grids, tag/pill markup, webfonts, colors, and filename conventions. Those defaults are fine for the upstream template, but they become a drift source when a downstream user records different resume styling in `_custom.md`.
4. **Batch/delegated paths were the highest-risk path.** Drift was worse when work moved through batch prompts or workers because those paths had thinner context loading.
5. **User-layer boundaries needed hardening.** Work-product and scratch paths should be consistently ignored or guarded so they cannot be staged accidentally.

The remediation plan from the audit was:

- Put procedural resume rules in `modes/_custom.md` and profile/narrative rules in `modes/_profile.md`.
- Make every mode and delegated worker load `_shared.md`, `_profile.md`, `_custom.md`, then the selected mode file.
- Make resume styling defaults explicitly overrideable by `_custom.md`, so downstream users can keep their own house style without fighting hard-coded mode or batch instructions.
- Add a render-time validator so stale resume HTML cannot silently produce a bad PDF.
- Add a read-only resume artifact audit to detect historical or generated HTML with retired patterns.
- Protect user-layer files through `.gitignore`, `DATA_CONTRACT.md`, `update-system.mjs`, and CI.
- Add tests so `_custom.md` cannot become write-only again.

[...]

PR #2952, opened 2026-08-16
fix(voice-dna): make voice-dna.md private via the template pattern (#1736)

rkosolapov:
## What changed

`DATA_CONTRACT.md` and `update-system.mjs`'s `USER_PATHS` both treat
`voice-dna.md` as user-layer personal data, but it was never gitignored.
`.github/workflows/no-user-data.yml` explicitly (and correctly, given the
current mechanism) excludes it from its guard, with the comment: *"it ships
as a populated system default, so edits to it are legit."*

#1736 already surfaced this exact inconsistency and closed it as a docs
wording fix rather than a mechanism fix. This PR is the mechanism fix:
`voice-dna.md` now follows the same `.template.md` + `doctor.mjs` auto-copy
pattern already used for `modes/_profile.md`, `modes/_custom.md`, and
`modes/_brief.md`.

## Fix

- `voice-dna.md` → `voice-dna.template.md` (`git mv`, keeps file history)
- `doctor.mjs`: added to the auto-copy template list in `onboardingState`
- `update-system.mjs`: `voice-dna.template.md` added to `SYSTEM_PATHS` (stays
  distributable/updatable); `voice-dna.md` was already in `USER_PATHS`
- `.gitignore`: `voice-dna.md` added

This gets both properties at once: new users still get a populated default
out of the box, and any personalization a user writes into their own
`voice-dna.md` never gets committed to their fork — which the current
mechanism can't guarantee, since the tracked file *is* the live file. No
change needed to `no-user-data.yml`'s guard list or `DATA_CONTRACT.md`'s
existing `voice-dna.md` description; both were already accurate for the
intended behavior, just not achieved by the old tracking mechanism.

## Testing

Verified end-to-end against a simulated fresh checkout: removed
`voice-dna.md`, ran `node doctor.mjs --json`, confirmed `autoCopied` includes
`"voice-dna.md"` and the copied file is byte-identical to
`voice-dna.template.md`. Confirmed the copy stays untracked/invisible to
`git status` after the `.gitignore` change.

## Note

If the "edits are legit" stance in the workflow comment was meant as a
deliberate end-state rather than a scoping call at #1736's triage, happy to
close this instead — the template pattern seemed worth raising since it
resolves the tradeoff outright rather than picking a side of it. Opened as a
separate PR from #2951 specifically so it can be evaluated (and rejected, if
that's the call) independently.

## Summary by CodeRabbit

* **New Features**
  * Added a writing-style guide covering voice, tone, formatting, banned vocabulary, and common AI writing patterns to avoid.
  * New projects automatically receive the guide when it is missing.
  * System updates now include the latest version of the guide.

* **Chores**
  * Local voice customization files are excluded from version control.


santifer:
@rkosolapov, second solid one today. The template-pattern route for voice-dna.md is the established way this repo handles personal files (same shape as _profile/_custom), and closing a gap where DATA_CONTRACT said "user data" but .gitignore didn't agree is exactly the kind of inconsistency worth hunting. Queued for review with #2951.

santifer:
Second merge of the night for you, @rkosolapov: the template pattern is how this repo handles personal files, and you closed a real DATA_CONTRACT/.gitignore inconsistency. 🚀

PR #2955, opened 2026-08-16
feat(eval): centralize CareerOps LLM config, harden URL/stage-1 ranking

zentrium74:
## Summary

Centralizes the CareerOps LLM configuration so evaluation and resume tailoring both read the same provider/model from `config/careerops-llm.yml`, decoupled from the chat-session model. Adds an Anthropic-compatible call helper (`on-demand-tools/careerops-llm.mjs`) that reads the OAuth credential from the Hermes profile. Also fixes three stacked bugs that were causing zero-result emails:

- Stage-1 ranking was boosting juniors/associates (`+1`); now demotes them (`-5`) so mid/senior items score higher and reach the eval queue first.
- `eval-batch.mjs` was writing `**URL:** ${item.url}` literally — empty when the scanner emitted a blank URL — producing reports with blank URL fields.
- The previous "skip rows with no URL" filter was silently dropping URL-less evals from the email entirely; now URL-less rows are kept and the email's `applyCell()` helper renders an explicit `⚠ no link` badge.

### New files

| File | Purpose |
|---|---|
| `config/careerops-llm.yml` | Single source of truth for provider/model/api used by both `eval-batch.mjs` and `on-demand-tools/tailor-resume.mjs` |
| `on-demand-tools/careerops-llm.mjs` | Shared Anthropic-compatible LLM call helper (reads OAuth token from Hermes profile `auth.json`) |

### Modified

| File | Change |
|---|---|
| `eval-batch.mjs` | Routes evaluations through the shared helper; preserves URL field in report (`URL or '(no URL captured)'`) |
| `on-demand-tools/tailor-resume.mjs` | Routes tailoring through the shared helper; fixes missing LaTeX backslash escape in project-header emitter (`\resumeHeader` was rendering as `esumeHeader`) |
| `run-automation.mjs` | Stage-1 ranking demotes junior/associate/new-grad/entry-level roles; URL-less rows are KEPT in the email and badged as "no link" instead of being silently dropped; type-safe URL check in `applyCell()` blocks `mailto:`/`javascript:` URLs |

## Behavior

- Mid/senior Stage-1 items now score `+10`; junior/associate items score `-5` (deprioritized, not removed)
- Eval-batch reports always include a URL field (real URL or `(no URL captured)`) so the LLM prompt is never blank
- Email renderer's `applyCell()` rejects `mailto:` and `javascript:` URLs and renders `⚠ no link` instead of broken `#` anchors

## Verified

- `node -c` passes on `eval-batch.mjs`, `on-demand-tools/tailor-resume.mjs`, `on-demand-tools/careerops-llm.mjs`, `run-automation.mjs`
- Stage-1 probe on the 72-item pipeline: mid/senior items now rank above juniors
- End-to-end eval against a real pipeline item with empty URL produced a `2.1/5` report with `(no URL captured)` preserved

## Public-repo hygiene

This PR does **not** include `run-automation.mjs` or `eval-batch.mjs` (they're excluded for PII / personal-Hermes-glue reasons, per repo policy). The cron-watcher and 8AM scan crons run via `run-automation.mjs` regardless — that's a different artifact that will be addressed separately.

## Test plan

- [x] `node -c` syntax check passes on all changed files
- [x] Stage-1 ranking verified against real `data/pipeline.md`
- [x] End-to-end eval run produced a report with preserved URL field
- [ ] 8 AM cron run today should produce an email with mid/senior items floated to top, juniors in Low/Medium band, working Apply links for items with URLs


Scott-Emberson (changes requested):
Reviewing as code owner for `dashboard/`. `career_test.go` and `i18n_test.go` are hermetic and carry genuine controls (unknown-status passthrough, duplicate-header last-wins, Via-column shift). The problem is three functional regressions in the owned Go paths, none of them covered by those tests, all stemming from `WithReloadedData` not preserving overlay-flow state.

**1. Viewer-originated Discard/SKIP silently drops the status change (`main.go:142-147` + `pipeline.go` `WithReloadedData:336`).** The branch sets the discard picker then calls `reloadPipelineData()` in the same Update. `WithReloadedData` rebuilds via `NewPipelineModel` and copies only sort/tab/view/search/cursor, so `discardPicker` and `discardPending*` are wiped before any render. The picker never appears, `commitDiscardReason` never fires, `UpdateApplicationStatus` is never called. Open a report, press `c`, choose Discarded, Enter: the tracker is untouched and the change is lost.

**2. The Hired flow is dead on arrival (`main.go:82-88` and `137-139`).** Both entry points set `hiredStep=1` then trigger the same rebuild, which drops `hiredStep`/`hiredApp`. The celebration flashes at most one frame. The Hired status write itself persists, so no data loss here, but the feature is unreachable.

**3. The viewer status picker no longer preselects the current status (`viewer.go:143-144`).** `currentNorm` is lowercase-canonical from `NormalizeStatus`, but the loop compares against the capitalized `pair.Canonical`, so it never matches and the cursor sticks on Evaluated. Normalize both sides: `data.NormalizeStatus(pair.Canonical) == currentNorm`.

Nit: the Skip pair persists `"Skip"` while `templates/states.yml` canonical is `SKIP`; display is fine via `NormalizeStatus`, but the on-disk value drifts from canonical. Use `"SKIP"`.

Scott-Emberson:
@santifer heads-up on something outside the code, flagging on privacy grounds rather than as part of the dashboard review.

This branch commits 16 files under `Tailored ATS PDF/out of the box/Aug 16/` that look like a real person's job-search artifacts, not fixtures: a named individual's tailored CV (PDF, `.tex`, `.json`, source-card) duplicated across four target companies. The JSON carries a real email, phone, home city, and live LinkedIn/GitHub profiles, with a full genuine employment history. It reads like a contributor's gitignored `output/` accidentally staged into the PR.

Two reasons to treat it as blocking regardless of the dashboard code: the repo is public, and once merged the files stay in git history even if a later commit deletes them. Deleting them in a follow-up commit is not enough.

Suggest: do not merge as-is, and have the author strip the entire `Tailored ATS PDF/` tree from the branch (history rewrite / fresh branch), keeping only the intended code changes. Happy to help verify the cleaned branch. I have kept the specific personal details out of this comment on purpose.

santifer:
Thanks for digging into these, @zentrium74 — but I have to close this one as-is: your branch carries your whole fork (100 files, including CODEOWNERS, workflows, and the skill entrypoints for every CLI), which makes it impossible to review or merge. It also bundles a helper wired to your personal Hermes OAuth profile, which is your machine's config rather than something the project can ship.

The three bugs you describe sound real and worth fixing, and each would be welcome as its own small PR from a clean branch off `main`:

1. the stage-1 ranking boosting junior/associate titles instead of demoting them,
2. `**URL:** ${item.url}` written literally when the scanner emits a blank URL,
3. the no-URL filter silently dropping evals from the digest.

The centralized LLM config idea is bigger than a fix: open an issue first so we can discuss the shape (see CONTRIBUTING · "What belongs in core") — provider/model config that isn't tied to one person's setup touches the AI-agnostic core path, and that's a conversation worth having before code.

santifer:
@zentrium74, when this closed a week ago we said follow-ups would be welcome and I want to make that concrete rather than leave it as a pleasantry: the free pool is at https://github.com/santifer/career-ops/issues?q=is%3Aopen+label%3A%22good+first+issue%22+no%3Aassignee (claim with `/assign`). Your centralized-LLM-config thinking here was solid; if an issue in that area opens up (config/eval), I'll ping you first.

PR #2958, opened 2026-08-16
fix(invite-match): validate CLI flags and add usage

### LINKED ISSUE #2854: invite-match.mjs has no --help and silently ignores mistyped flags

HariomPCode:
## What does this PR do?

Fixes the CLI flag handling in `invite-match.mjs` for #2854.

- Adds a `USAGE` block documenting the existing CLI arguments and flags.
- Adds `--help` / `-h` handling through the shared `validateFlags()` helper.
- Rejects unknown or mistyped `-`-prefixed arguments instead of silently ignoring them.
- Guards flag validation so importing `invite-match.mjs` from other scripts does not validate those scripts' CLI arguments.
- Leaves the existing matcher and argument handling unchanged.

## Related issue

Closes #2854

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] I have tested the changes locally.
- [x] Existing tests pass.
- [x] The normal matcher output remains unchanged.
- [x] This PR only changes the intended file.

## Summary by CodeRabbit

* **New Features**
  * Added help output available through `--help` and `-h`.
  * Added validation for command-line flags, with clear rejection of unsupported options.
  * Preserved existing space-separated `--file` and `--id` argument formats.


Priyans-Lathiya:
Hi @HariomPCode , I would like to suggest to write the test case for the same change in `invite-match.test.mjs`. 
If you need any help, I would love to do.

HariomPCode:
> Hi @HariomPCode , I would like to suggest to write the test case for the same change in `invite-match.test.mjs`. 
> If you need any help, I would love to do.

Okay I'll try to update by tomorrow the test file as per updated implementations.

santifer:
Welcome, @HariomPCode! Two details decided this over the competing PR: your KNOWN_FLAGS list matches the script's real surface exactly (no phantom --json), and unrecognized-before-help means --help --bogus errors instead of exiting 0. Plus a test. Merged — your first contribution! 🚀

PR #2959, opened 2026-08-16
fix: SSRF protection via DNS verification, path traversal guards, and scan log visibility

rapid-ranjeet:
Integrates validateUrlSecurity for SSRF protection, adds realpath containment checks against path traversal, and adds timeout/warnings to scanner.

## Summary by CodeRabbit

* **New Features**
  * Added a Career Studio dashboard for job matching, application tracking, ATS scoring, interview preparation, outreach, salary evaluation, recruiter contacts, and resume PDF generation.
  * Added job-board integrations, daily digests, batch matching, analytics, resume exporters, and interview/outreach tools.
  * Added configurable endpoints and command-line tools for jobs, contacts, scoring, preparation, outreach, salary evaluation, and PDF rendering.

* **Security**
  * Blocked restricted browser URLs and unsafe or symlinked CV/PDF output paths.

* **Reliability**
  * Added search timeouts and graceful handling for missing or invalid data.


santifer:
Welcome to career-ops, @King70870, and thanks for aiming your first PR at security. Quick honest status so you're not left guessing: this touches four core scripts (scan, pdf, browser-extract), so it gets a full security-focused review rather than a fast pass; that review is queued. Your CI was approved on our side already (first contributions need a maintainer to release the runs, that wait is on us, not you).

Scott-Emberson (changes requested):
Reviewed as code owner of `/tests/`. Six new files land in our area: `tests/addons.test.mjs`, `tests/candidate-suite.test.mjs`, `tests/cv-rendercv.test.mjs`, `tests/gaps.test.mjs`, `tests/providers/adzuna.test.mjs`, `tests/providers/jsearch.test.mjs`. Everything below is scoped to those. The rest of the diff is not mine to judge and I have left it alone.

The good part first, because it is genuinely good. `tests/providers/adzuna.test.mjs` and `tests/providers/jsearch.test.mjs` are the shape we want. They stub `fetchJson` through the injected seam, capture the URL and options and assert the actual values (`app_id=test_app_id`, `redirect === 'error'`), feed a fixture with two valid rows plus two deliberately broken ones and assert the parser keeps exactly the two, then set and clear the env to cover the credentialed and uncredentialed paths. I ran both at `b57cc63e`: 4 checks each, exit 0, no network. Nothing to change.

The other four need work. Two of the three blocking items are not judgement calls.

**1. `tests/gaps.test.mjs:26` writes into the user layer. Blocker, CONFIRMED.**

`syncApplicationStatus()` ends in an unconditional `writeFileSync(TRACKER_TSV, ...)` at `sync-application-tracker.mjs:71`, where `TRACKER_TSV` is `resolve(__dirname, 'data', 'tracker.tsv')`. `__dirname` is the repo root, so calling it from a test writes into the user's real `data/` directory.

Measured in a clean worktree at `b57cc63e`:

```
data/ before:  offers  parser-output
$ node tests/gaps.test.mjs
data/ after:   offers  parser-output  tracker.tsv   (44 bytes)
```

`data/*` is user layer under `DATA_CONTRACT.md` and is never auto-written. A user running `node test-all.mjs` would get a file created in their data directory as a side effect of running tests, in a second tracker format competing with `data/applications.md`. Whatever happens to the sync tool itself, the test must not call the real writer against the real path. Point it at a throwaway dir, the way the neighbouring suites do.

**2. `tests/addons.test.mjs` crashes before its first assertion. Blocker, CONFIRMED.**

```
$ node tests/addons.test.mjs; echo exit=$?
SyntaxError: The requested module './ats-score.mjs' does not provide an export named 'extractTokens'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:226:21)
exit=1
```

`ats-score.mjs:42` exports `extractPhrasesAndTokens`, not `extractTokens`. Both `tests/addons.test.mjs:3` and `tailor-and-render.mjs:16` ask for `extractTokens`. None of the five checks in the file runs, and `test-all.mjs` auto-discovers `tests/**/*.test.mjs`, so this is a red suite as soon as the harness runs. It has not surfaced here because the test workflow has not run on this PR: `gh pr checks 2959` currently lists only CodeRabbit, GitGuardian and the labeler.

**3. Several checks pass when the code under test does nothing. Major, CONFIRMED.**

- `tests/gaps.test.mjs:27`, `typeof syncResult.syncedCount === 'number'`. `syncApplicationStatus()` returns a number on every path, including the ordinary one where `../remote-job-pipeline/data/latest_applications_audit.json` is absent and nothing syncs. The assertion cannot fail.
- `tests/gaps.test.mjs:19`, `check.verdict === 'pass' || check.verdict === 'block'`. The domain is `'pass'|'warn'|'block'` (`verify-cv-facts.mjs:491`), so this accepts two outcomes out of three without saying which is correct for that input. It observed `block` on my run. If the guardrail regressed to `pass` on the same CV text, the check still goes green.
- `tests/candidate-suite.test.mjs` is shape-only throughout: `prep.starStories.length >= 2`, `salary.benchmarks && salary.suggestedCounter`, `digest.includes('DAILY 9:00 AM SMART JOB DIGEST')`. A stub returning two empty strings satisfies the first, and the last pins a banner rather than any digest content.

The house move is a control assertion: change an input, prove the output moves. Without one, a green check says the module loaded, not that it works.

**4. The security fixes in the title have no test at all. Minor.**

The three changes the PR is named for (the `browser-extract.mjs` URL revalidation, the `generate-pdf.mjs` realpath containment check, the scan warning) are the part most worth pinning, and none of the six new test files touches them. The containment check in particular has existing suites it could sit beside.

**5. Nit: the provider tests delete env vars they did not set.** `tests/providers/adzuna.test.mjs:119-120` and `tests/providers/jsearch.test.mjs:122` do `delete process.env.<KEY>` as cleanup. Save the previous value and restore it instead, so a developer with real keys in their environment does not lose them mid-run. Small thing, but a test that quietly edits the environment it ran in is hard to trust later.

Requesting changes on 1 and 2. Item 3 is what I would want fixed before this lands. 4 and 5 are yours to weigh.

Two notes outside my area, offered as signal rather than as part of the verdict, since neither is mine to decide: the PR touches `update-system.mjs`, which is santifer's, and the diff is +6309/-11 across 47 files behind a title describing three small fixes. Both are maintainer calls and I have not commented on either file.

PR #2961, opened 2026-08-16
fix(batch-tailor): --min-score 4.5 was silently ignored, and the worker's pdf mode path was cwd-relative

Zoubeir23:
No issue — found while auditing `batch-tailor.mjs`, one of the few core scripts with no test file. `CONTRIBUTING.md:22` welcomes a direct PR for a bug fix.

`batch-tailor.mjs` spawns one agent run per matching job, so a mis-parsed threshold is not cosmetic: it decides how many paid runs happen and on which roles. It hand-rolled its own flag parsing and predates `lib/cli-flags.mjs`, so it carries three of the exact failures that helper exists to end.

## Measured on `main`

Sandbox `batch-state.tsv` with one `completed` job scoring **4.5**:

| Command | `main` | Should be |
|---|---|---|
| `--min-score 9.9` | `Found 1 roles scoring >= 4. Beginning bulk tailoring...` | match nothing |
| `--min-score=abc` | `No completed roles found with score >= NaN.` (exit 0) | usage error |
| `--min-scor=9.9` (typo) | `Found 1 roles scoring >= 4. Beginning bulk tailoring...` | usage error |

The first is the costly one: the caller asked for a stricter threshold, the script read only the `--min-score=N` form, dropped the value, and went on to tailor at the 4.0 default — spawning worker runs on roles that were explicitly excluded.

The second is silent: a `NaN` threshold compares false against every score, so the output is indistinguishable from a genuinely empty batch.

## The worker's mode file was cwd-relative

```js
const batchStateFile = join(__dirname, 'batch', 'batch-state.tsv');  // absolute
...
'--append-system-prompt-file',
'modes/pdf.md',                                                       // cwd-relative
```

Within one script, the state file resolves through `__dirname` while the mode file does not, so running it from anywhere but the project root hands the worker a path that does not exist. `resolve` was imported and never used, which reads like the intent was there.

This one matters beyond tidiness: `modes/pdf.md:44` is where the CV fact gate is instructed (`node verify-cv-facts.mjs {html-path}`). Losing that system prompt drops the anti-fabrication check from a **bulk** CV run — the run where nobody is reading each output.

## The fix

- Route parsing through `lib/cli-flags.mjs` (`validateFlags` / `hasFlag` / `flagValue`) instead of a fourth hand-rolled copy. Both flag forms work, unknown flags fail fast, `--help` is handled by the shared path.
- A non-numeric `--min-score` exits 1 with the value quoted back, rather than matching nothing.
- `modes/pdf.md` resolves through `__dirname`, like the state file next to it.
- `CAREER_OPS_BATCH_STATE` overrides the state-file path — the same override `merge-tracker.mjs` already honours, and what makes the script testable without touching a real batch run.

## Test plan

New `tests/batch-tailor-flags.test.mjs`, picked up by `test-all.mjs` auto-discovery. Every case picks a threshold that matches **no** job, so the script always exits before spawning a worker.

Two notes on making the cases honest:

- The env override had to land first. Without it the script exits 1 on the missing state file, and the "bad value" and "typo" cases went green **before** the fix — passing for the wrong reason. They only became genuinely red once the script could reach its own parsing.
- The mode-path fix is asserted at source level (the script must not pass a bare `'modes/pdf.md'`), the same shape as `test-all`'s provider-pacing guard, because observing the real spawn argument would mean launching a paid worker.

- [x] 4 of the 6 cases verified red after the override landed; the equals-form and `--help` cases pass before and after, which is what makes them guards.
- [x] `node test-all.mjs` — 3944 passed, 0 failed (the single warning is the pre-existing `Dashboard build skipped — go compiler not in env`).

## Noted, not changed

`spawnSync('claude', ...)` hardcodes one CLI in a project that ships a per-CLI headless command table for nine of them (`AGENTS.md`). On any other CLI the loop prints an ENOENT per job and keeps going. That is a design decision rather than a defect, so it is not in this PR — happy to open it separately if you want it.

## Summary by CodeRabbit

* **New Features**
  * Added flexible `--min-score` options in separated and equals formats.
  * Added support for overriding the batch state file location through an environment setting.
  * Added help flag handling and clearer score threshold validation.

* **Bug Fixes**
  * Invalid or missing values, unknown options, and unreadable state files now return clear errors and nonzero exit codes.
  * PDF processing reliably loads its configured instructions regardless of the launch directory.

* **Tests**
  * Added coverage for command-line parsing, validation, state overrides, error handling, and PDF processing paths.

### DIFF
--- a/batch-tailor.mjs
+++ b/batch-tailor.mjs
@@ -37,7 +37,12 @@ validateFlags(args, ['--min-score', '--help', '-h'], USAGE, { valueFlags: ['--mi
 let minScore = 4.0;
 if (hasFlag(args, '--min-score')) {
   const raw = flagValue(args, '--min-score');
-  minScore = Number.parseFloat(raw);
+  // Number(), not parseFloat(): parseFloat stops at the first invalid character,
+  // so "4.5abc" silently became 4.5 and the run proceeded on a threshold the
+  // caller never wrote. The empty/whitespace guard must come FIRST — Number('')
+  // is 0, which is finite, and a 0 threshold tailors every completed job.
+  const trimmed = typeof raw === 'string' ? raw.trim() : '';
+  minScore = trimmed === '' ? NaN : Number(trimmed);
   // A NaN threshold compares false against every score, so the old code printed
   // "no roles found with score >= NaN" and exited 0 — indistinguishable from a
   // genuinely empty batch. Fail loudly instead.
@@ -52,7 +57,19 @@ if (!existsSync(batchStateFile)) {
   process.exit(1);
 }
 
-const lines = readFileSync(batchStateFile, 'utf-8').split('\n');
+// existsSync() is true for a directory and says nothing about permissions, so
+// reading can still throw (EISDIR, EACCES) — an uncaught stack trace where a
+// usage error belongs. Name the RESOLVED path: with CAREER_OPS_BATCH_STATE set,
+// the value that failed is not the one written in the source.
+let stateContent;
+try {
+  stateContent = readFileSync(batchStateFile, 'utf-8');
+} catch (err) {
+  console.error(`ERROR: cannot read batch state file at ${batchStateFile}: ${err.message}`);
+  process.exit(1);
+}
+
+const lines = stateContent.split('\n');
 const toProcess = [];
 
 for (const line of lines) {
--- a/batch-tailor.mjs
+++ b/batch-tailor.mjs
@@ -6,7 +6,12 @@ import { spawnSync } from 'child_process';
 import { fileURLToPath } from 'url';
 
 const __dirname = fileURLToPath(new URL('.', import.meta.url));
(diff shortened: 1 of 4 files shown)


Scott-Emberson (approved):
Code owner for `tests/`. Both owned files hold up: `cli-flags.test.mjs` uses the `REACHED-END` marker as a real control and `batch-tailor-flags.test.mjs` asserts the NaN message is gone rather than just a nonzero exit, all hermetic under `mkdtempSync` with no worker spawn. One nit, non-blocking: the case-2 comment claims a below-threshold value "still selects it", but `--min-score=9.9` against a 4.5 job also rejects everything, so nothing verifies positive selection. Either correct the comment or note the design intentionally avoids the paid happy path. Approving.

santifer:
Two silent failures in one honest fix: a threshold flag that parsed and then did nothing, and a pdf path that only worked from one directory. Both now covered by tests that would have caught them the first time. Second solid one today, @Zoubeir23. Merged! 🚀

PR #2962, opened 2026-08-16
feat(providers): senjob — the scanner's first African source (Senegal)

Zoubeir23:
No issue — `CONTRIBUTING.md:22` welcomes a direct PR for a new zero-auth scanner provider.

The scanner has **no African coverage at all**. This adds the first source: [senjob.com](https://senjob.com), a Senegalese board (Dakar and the regions).

## Why this one parses HTML

The repo prefers JSON, so the exception needs a reason. Five Senegalese boards were probed on 2026-08-16:

| Board | Result |
|---|---|
| `emploi.sn` | does not resolve |
| `emploisenegal.com` | 403 |
| `emploidakar.com` | WP Job Manager; `/wp-json/wp/v2/job-listings` **exists**, but Cloudflare answers the data endpoints with a `Just a moment...` interstitial (403) even with full browser headers |
| `expat-dakar.com` | Cloudflare-fronted, and it is classifieds |
| **`senjob.com`** | plain Apache, no interstitial — the only one left |

`emploidakar` is the frustrating one: the REST route is there and would serve clean JSON. Getting past the interstitial needs a headless browser, which is the line the repo already draws — `scan-interamt.mjs` exists in Playwright for exactly that reason.

## robots.txt

Fetched 2026-08-16. It disallows `/cvs/`, `/*/cvs/`, `/jobseekers/cvs/`, `/jobseekers/Images/` and the employer-registration pages. It does **not** disallow `/offres-d-emploi.php` or the `/jobseekers/*.html` postings this provider reads.

The disallowed paths are candidate CVs and employer account pages — personal data this provider never requests. Noted in the file header so the boundary stays visible to whoever edits it next.

## Design — the anchors, and the silent-zero guard

Markup-based extraction rots, so the parser anchors on the two things a redesign is least likely to move:

1. the posting URL shape `/jobseekers/{slug}_e_{id}.html`, which carries the posting id;
2. the hidden ISO date `<span style="display:none;">YYYY-MM-DD</span>` sitting next to its localized form (`14 Aou.`).

No CSS class and no inline style is matched.

**The part that matters most:** when a page that still contains posting links yields no jobs, the provider **throws**. Returning `[]` would render a broken parser as a board with no openings — indistinguishable from a healthy quiet board, and precisely what makes scraped sources untrustworthy. `assertParsedSomething()` tests the posting-link *shape* rather than a marker word, so it fires on a real markup change rather than on a copy edit.

Three live behaviours drive the rest:

- **The title link and the date sit in sibling `<tr>`s.** A parser that windows around the link loses the date, so rows are split on `<tr` and merged by posting id.
- **The board pins sticky postings to the top of every page.** Merging by id makes the repeat contribute nothing instead of duplicating, and a page that adds no new URL ends the run.
- **Anchor bodies carry `<!-- d ico postulez -->` between the title and a spacer image.** Comments are stripped before tags, or the comment text lands inside the title.

`company` is left **empty**: the listing rows never name the employer, and `_types.js` documents an empty company for exactly that case. Deriving one from the slug would be a fabricated claim.

Pacing is 250 ms between pages of the same board — ordinary politeness toward a single small Apache host, not a measured rate limit. No throttling was observed while probing, and the comment says so rather than implying evidence that does not exist.

## Test plan

`tests/providers/senjob.test.mjs` — 18 cases, fixtures trimmed from the live page with the structure preserved. Each covers a shape that decided the design: the sibling-row date, the sticky-row merge, the comment inside the anchor, the place-and-date-in-one-row variant, the empty `company`, and both directions of the silent-zero guard (throws on unparsed postings, stays quiet on a genuinely empty page).

`fetch()` is exercised against a fake `ctx` for pagination, dedup, the stop condition, pacing, and the throw on a broken page 1.

- [x] `node test-all.mjs --only providers/senjob` — 18 passed, 0 failed.
- [x] `node test-all.mjs` — 3956 passed, 0 failed (the single warning is the pre-existing `Dashboard build skipped — go compiler not in env`).
- [x] Live end-to-end against the real board: **158 postings in 5.5 s**, 100% carrying a location, 98% a publication date.
- [x] `node validate-portals.mjs` — 0 errors, 0 warnings.

## Honest note on volume

Tech roles are a thin slice of this board — roughly 4 genuine developer postings per 200 listings, alongside NGO, admin and field work. The commented stanza in `templates/portals.example.yml` says so, so nobody enables it expecting a tech feed. What it does give is coverage of a market the scanner could not see at all: a live sweep surfaced `Developpeur Fullstack`, `Développeur(se) Web` and `Software and Data Engineer`, all in Dakar.

## Summary by CodeRabbit

* **New Features**
  * Added Senjob as a supported job board for Senegalese job listings.
  * Captures job titles, locations, publication dates, and available employer information.
  * Supports pagination across multiple result pages and removes duplicate postings.
  * Includes an example configuration for enabling Senjob.
* **Bug Fixes**
  * Handles restricted pages, unavailable employer names, empty results, and changes to listing formats safely.
  * Stops processing when additional pages contain no listings.


Scott-Emberson (approved):
Code owner for `tests/`. `senjob.test.mjs` exercises the real parser end to end with exact-value assertions (`title`, absolute `url`, `location`, `postedAt`) and a both-direction silent-zero guard: it throws when posting links parse to nothing and stays quiet on a genuinely empty page, so a no-op parser cannot pass. Entity decoding is covered to the sibling depth. One nit, non-blocking: the `MAX_PAGES_CAP` (50) boundary is never exercised. Approving.

Zoubeir23:
Thanks @Scott-Emberson — the nit was fair, and a little pointed: `MAX_PAGES_CAP` arrived in the previous round while addressing a review note on `entry.max_pages`, and I added the ceiling without covering it.

Exercised in `12cab06`. Every page in that case serves a *new* posting, so the cap is the only thing that can end the run — otherwise the assertion would pass on the dedup stop condition instead. Verified it discriminates: removing the `Math.min` turns it red with `requested 999 pages`.

27 provider cases, full suite 3965 passed / 0 failed.

santifer:
The scanner's first African source, and it lands with the full provider contract: tests, a commented template stanza, and the docs row. Taking Scott's nit and folding MAX_PAGES_CAP in properly instead of defending the earlier round is what made this a fast review. Merged! 🚀

PR #2966, opened 2026-08-17
fix(web): persist the only installed CLI so jobs can start

mbrow190:
## Why

Config auto-highlighted the only installed CLI (often Grok) without writing `career-ops:config`. Jobs read localStorage and failed immediately with "No CLI configured — open Config". It looked configured. It was not.

## What

- Persist `cliId` when exactly one CLI is installed and nothing was saved.
- `startJob` falls back to that pick via `/api/clis`.
- Two or more installed CLIs still require Save (no silent pick).
- Unit coverage for the sole-install rule.

## Notes

Web-only. Does not read or write `.env`. No user-layer files.

## Test plan

- [x] `node --test tests/lib/saved-cli-pick.test.mjs`
- One CLI installed, empty localStorage, start Evaluate → job starts
- Two CLIs installed, empty localStorage → still asks to Save

## Summary by CodeRabbit

* **New Features**
  * Automatically selects and saves the only installed CLI when no selection exists.
  * Preserves an existing CLI selection.
  * Safely remembers CLI configuration across sessions.

* **Bug Fixes**
  * Job startup now handles missing CLI configuration gracefully and provides guidance to save configuration.
  * Invalid or unavailable saved settings no longer cause failures.


santifer (approved):
Real "looks configured, isn't" bug: Config highlighted the sole installed CLI visually but never wrote career-ops:config, so startJob read empty localStorage and dead-ended. Persisting the sole pick (and only the sole pick — 2+ still require an explicit choice) is correct, and extracting the localStorage logic to saved-cli.ts gives it one testable home instead of two inline copies. resolveCliId's /api/clis fallback is a good belt-and-suspenders for a job started before Config was ever saved. Test covers the sole-vs-zero-vs-two cases. Web-only. Merging.

santifer:
Thanks, @mbrow190 — "looks configured, it wasn't" is exactly the kind of first-run trap that quietly kills a beta: highlighting the sole installed CLI without persisting it meant every job dead-ended on empty localStorage. Persisting only the sole pick (2+ still require an explicit choice) is the right line, and pulling the localStorage logic into saved-cli.ts gives it one tested home. Third of your burst. Merged 🚀

PR #2967, opened 2026-08-17
fix(web): Today primary action opens the report

mbrow190:
## Why

Today's primary button was "Mark applied". It wrote Applied through `/api/status` without opening Apply, generating a CV, or seeding a follow-up. Seekers could mark a role applied without applying.

## What

- Primary control is **Review** → `/pipeline/{n}` (PDF + Apply live there).
- Skip stays.
- Applied remains as a quiet override, not the default.

## Notes

Web-only UI honesty fix. No user-layer files.

## Test plan

- [x] `node --test tests/lib/decision-card-cta.test.mjs`
- Today queue: Review is the wide button and opens the report
- Applied still records status if you click it

## Summary by CodeRabbit

* **New Features**
  * Updated decision cards with a prominent **Review** action linking to the pipeline report.
  * Kept **Skip** available and retained **Applied** as a secondary status action.
  * PDF and application actions are now accessible from the pipeline page.

* **Tests**
  * Added coverage to verify the updated decision-card actions and ordering.


santifer (approved):
This is the one I had flagged in my own UX notes — the Today primary let a seeker write Applied with no application behind it (no Apply flow, no CV, no follow-up). Promoting Review → /pipeline/{n} as primary and demoting Applied to a quiet override is the honest hierarchy: the default action does the thing it names. Skip stays neutral. Test pins the button order. Web-only UI, disjoint. Merging.

santifer:
Thanks, @mbrow190 — this is the honest fix: a primary button labeled "Mark applied" that wrote Applied without opening Apply let people mark a role done without doing it. Promoting Review → the report (where PDF + Apply actually live) and keeping Applied as a quiet override makes the default action do what it says. Second of your beta-friction burst to land. Merged 🚀

PR #2972, opened 2026-08-17
docs(pipeline): note that pipeline.md section headers may be localized

IaroslavMazur:
## What

The English `modes/pipeline.md` documents the inbox format with only `## Pending` / `## Processed`, and Workflow step 1 says to "search for `- [ ]` items in the 'Pending' section" with no indication that the header might read something else.

15 of the 20 translated pipeline modes already carry a note that it might. From `modes/de/pipeline.md`:

> Hinweis: Die Sektion-Überschriften können auf EN ("Pending"/"Processed"), ES ("Pendientes"/"Procesadas") oder DE ("Offen"/"Verarbeitet") sein. Beim Lesen flexibel sein, beim Schreiben dem Stil der bestehenden Datei treu bleiben.

The same note appears in `es`, `da`, `fr`, `hi`, `id`, `it`, `ko`, `nl`, `pl`, `pt`, `ru`, `tr` and `ua`. It is missing from `en`, `ar`, `ja`, `zh` and `zh-TW`.

## Why it matters

The code has supported both spellings all along:

- `scan.mjs` — `const PENDING_MARKERS = ['## Pending', '## Pendientes']`, and `PROCESSED_MARKERS` alongside it
- `reconcile-pipeline.mjs` — `PENDING_RE = /^##\s+(Pendientes|Pending)\s*$/i`, and `PROCESSED_RE`

So a Spanish-headed `pipeline.md` is a supported state, not a broken one. But the English mode is the one an agent reads by default, and it is silent on this. Read literally, an agent working on a file whose header says `## Pendientes` has no instruction telling it the section it is looking for is right there. It happens to work today because agents tend to read the whole file, which makes this the kind of gap that stays invisible until it isn't.

The project's own history makes the ES spelling likely in the wild: it originated in Spanish, so files carrying `## Pendientes` / `## Procesadas` predate the English default.

## What this changes

Two lines in `modes/pipeline.md`, documentation only:

1. Adds the note after the format block, mirroring where the translated modes put it, and naming the two constants so the next reader can verify it rather than take the doc's word for it.
2. Amends Workflow step 1 with a short parenthetical, since that is where the lookup actually happens and a note 30 lines further down is easy to miss.

No code, no behaviour change. Per `CONTRIBUTING.md`, docs go straight to a PR without an issue first.

## Left out on purpose

`modes/ar`, `modes/ja`, `modes/zh` and `modes/zh-TW` are missing the same note. I have not touched them — the repo handles translations as separate per-language issues, and I would rather not add four paragraphs I cannot review as a native speaker. Happy to open a tracking issue for those if it is useful.

## Summary by CodeRabbit

- **Documentation**
  - Updated workflow guidance to support localized equivalents of the “Pending” section.
  - Documented recognition of English and Spanish section headers during scanning and reconciliation.


santifer:
Good catch on a real gap between the prose and the code, @IaroslavMazur. I checked the claim before merging and it lands exactly: `scan.mjs:1925` carries `['## Pending', '## Pendientes']` and `reconcile-pipeline.mjs:153` matches `(Pendientes|Pending)`, so the tooling had been bilingual for a while and only the mode still told the agent to look for one spelling. "Flexible when reading, faithful to the existing file's style when writing" is the right rule to state out loud, because the failure it prevents is an agent helpfully rewriting someone's Spanish tracker into English. Merged 🚀

PR #2974, opened 2026-08-17
feat(verify-pipeline): flag a malformed follow-ups.md instead of reading it as empty

### LINKED ISSUE #2971: A malformed `data/follow-ups.md` table silently reads as zero follow-ups in both `stats.mjs` and `followup-cadence.mjs`
### Code of Conduct

### Existing issues

### Problem

`modes/followup.md` (line 179) documents the follow-ups table as:

```
| num | appNum | date | company | role | channel | contact | notes |
```

Both consumers parse exactly that, positionally: `stats.mjs` → `computeFollowupStats()` reads `parts[1]` as `num` and `parts[2]` as `appNum`, and `followup-cadence.mjs` → `parseTrackerContent()` does the same. Both `continue` when either value fails `parseInt`.

If the table is written with any other column order — which is easy to do, because the file is hand-editable Markdown and the mode's own report table (line 72) has a *different* shape — every row is skipped and both tools report zero follow-ups. There is no warning, no parse error, and no hint that the file was even read.

The failure is quiet in a way that misleads. On a real search log, a follow-ups file with a plausible-looking but wrong 6-column header:

```
| Date | Company | Tracker # | Channel | Type | Details |
```

produced `Follow-ups: 0 sent across 0 apps | N Applied apps with none` in `node stats.mjs --summary`, with two follow-ups sitting in the file, visible on screen. Here `parts[2]` holds the company name, so `parseInt("Acme Corp")` is `NaN` and the row is dropped. Reordering the columns to the documented shape immediately produced `2 sent across 2 apps`.

The consequence is worse than a wrong number. Follow-up compliance is exactly the metric a user consults to decide whether they have a follow-up problem, and a silent zero tells them their logging is broken when the data is right there — or, in the other direction, tells them nothing is logged when the real problem is a column order.

`verify-pipeline.mjs` already validates `data/applications.md` thoroughly (canonical statuses, duplicate rows, report links, score formats, row shape, column swaps). `data/follow-ups.md` gets no equivalent check, even though it feeds two reporting surfaces and is equally hand-edited.

### Proposed solution

Add a follow-ups schema check to `verify-pipeline.mjs`, in the spirit of the existing tracker checks:

1. If `data/follow-ups.md` contains any `|`-delimited data rows, assert that at least one parses under the documented shape — i.e. `parts[1]` and `parts[2]` both `parseInt` cleanly.
2. If data rows exist but none parse, emit an error naming the expected column order and pointing at `modes/followup.md`. This is the case that is currently invisible.
3. If some rows parse and others do not, warn with the offending line numbers rather than dropping them silently.

Two smaller alternatives, either of which would also have surfaced this:

- **Parse by header name instead of position.** `computeRunStats()` in the same `stats.mjs` already does this for `scan-runs.tsv`, with an explicit comment that positional slicing "would silently miscount". The same reasoning applies here, and it would make the file order-insensitive rather than merely loud about being wrong.
- **Have the two consumers report rows they could not parse**, instead of `continue`-ing. Even a one-line stderr note would have made this self-diagnosing.

Happy to open a PR — I have the `verify-pipeline.mjs` check written and tested locally and will link it here shortly. Also worth noting: the docstring above `computeFollowupStats()` says it parses "the same table shape followup-cadence parses", which is true of the code but reads as though the shape is guaranteed somewhere. It is not.

### Area

Other

IaroslavMazur:
Closes #2971.

## Problem

`stats.mjs` (`computeFollowupStats`) and `followup-cadence.mjs` (`parseTrackerContent`) both read `data/follow-ups.md` positionally, in the shape `modes/followup.md` documents:

```
| num | appNum | date | company | role | channel | contact | notes |
```

Both `continue` when `parts[1]` or `parts[2]` fails `parseInt`. A table written with any other column order is therefore dropped row by row and reports as **zero follow-ups** in both tools — no error, no warning, nothing distinguishing it from a file where nothing has been logged yet.

That silence is the actual bug. Follow-up compliance is the number a user consults to decide whether their follow-ups are working, so a silent zero tells them their logging is broken when the data is sitting right there in the file.

`verify-pipeline.mjs` already validates `applications.md` thoroughly (canonical statuses, duplicate rows, report links, score formats, row shape, column swaps). `follow-ups.md` had no equivalent check despite feeding two reporting surfaces and being equally hand-edited.

## What this adds

**`stats.mjs` — `checkFollowupsSchema(content)`**, exported next to the parser it protects. Returns the counts needed to separate six states that currently all look alike: absent, empty, header-only, missing-delimiter, malformed, valid. Its row guard mirrors both consumers exactly, with a comment saying so — if it diverges it stops predicting what they will actually do with the file. Reporting only: no throw, no writes, so the consumers keep degrading gracefully.

**`verify-pipeline.mjs` — Check 14.** Error when no data row parses, warning with line numbers when only some do, `ok` otherwise. Follows the existing numbered-check pattern and delegates to an exported function the way Check 13 delegates to `checkTrackerSync()`, so the logic is testable without running the script. Resolves the path the same way the two consumers do (`CAREER_OPS/data/follow-ups.md`) rather than from `APPS_FILE`, because the check exists to predict their behaviour and so has to read the file they read.

**`test-all.mjs` — 4 assertions**, added to the existing `stats.mjs` block: the documented order passes; the real-world 6-column shape that regressed is flagged *and* is pinned as reading zero through `computeFollowupStats`; a partially-parseable table reports the offending line number; and the four empty-ish states stay distinct.

No new files, so nothing to register in `SYSTEM_PATHS`.

## Behaviour

```
### no follow-ups.md
✅ No follow-ups.md yet — nothing to schema-check

### wrong column order (the regression)
❌ follow-ups.md: none of its 1 row(s) parse, so stats.mjs and followup-cadence.mjs
   will both report zero follow-ups — expected column order
   | num | appNum | date | company | role | channel | contact | notes | (see modes/followup.md)
📊 Pipeline Health: 1 errors, 0 warnings

### documented order
✅ follow-ups.md schema valid (1 logged follow-up)

### one bad row among good ones
⚠️  follow-ups.md: 1 of 2 rows will be skipped by stats.mjs and followup-cadence.mjs
   (line 6) — expected | num | appNum | date | company | role | channel | contact | notes |
📊 Pipeline Health: 0 errors, 1 warnings
```

`followup-seed.mjs`'s `- next #N …` pin lines sit outside the table and are ignored.

## Tests

`node test-all.mjs` → **3941 passed, 0 failed**, up from 3937 on `main` at `22cbe88e`. The 2 warnings are pre-existing (upstream provider HTTP 503s), unchanged by this branch.

## Notes for review

Two alternatives are described in #2971 and deliberately not taken here:

- **Parsing by header name** instead of position would make the file order-insensitive rather than merely loud about being wrong. `computeRunStats()` in the same file already does this for `scan-runs.tsv`, with a comment that positional slicing "would silently miscount". It is the better long-term shape but changes behaviour in two consumers, so it felt wrong to fold into a check-adding PR. Happy to do it as a follow-up if you'd prefer that direction.
- **Having the consumers report unparsed rows themselves** would make them self-diagnosing, at the cost of noise in every stats run.

Also: the docstring above `computeFollowupStats()` says it parses "the same table shape followup-cadence parses". True of the code, but it reads as though the shape is enforced somewhere. It wasn't. Left as-is rather than reworded in this PR — say the word and I'll adjust it.

## Summary by CodeRabbit

- **New Features**
  - Added validation for the follow-up Markdown table, including its structure, headers, data rows, and parseable entries.
  - Verification now reports clear success, warning, and error results for missing, incomplete, or invalid follow-up data.

- **Bug Fixes**
  - Improved detection of incorrectly ordered columns and partially or fully unparseable follow-up rows.

- **Tests**
  - Added regression coverage for valid, incomplete, malformed, and absent follow-up tables.


santifer:
This is the good kind of diagnostic, @IaroslavMazur: it doesn't invent its own idea of valid, it mirrors what the two consumers will actually do, and it says so in a comment that tells the next person the mirror IS the contract. I checked that claim rather than take it: `stats.mjs` and `followup-cadence.mjs:405` both guard on `parts.length < 8` plus `parseInt(parts[1])` and `parseInt(parts[2])`, which is exactly what `checkFollowupsSchema` replicates. The test that convinced me is the second one, because it asserts the invisible half too: same input, `computeFollowupStats` returns 0, and that zero is precisely what a user would have read as "my follow-ups are fine". Reporting rather than throwing is right for a file the tools must keep degrading on. Merged 🚀

Issue #2883, opened 2026-08-14
`discover-ats`: probes 4 of 74 providers, and reports definitive 404s as "re-run" transient errors


hernanflores:
## Summary

Two related problems in `discover-ats.mjs` that together make `discover` report false negatives and give the user no way to tell a false negative from a real one.

Found while resolving four LatAm/remote companies (Mercado Libre, Ualá, Clip, Deel) against `portals.yml`.

## 1. Vendor coverage is 4 of 74

`discover-ats.mjs:76` limits probing to three slug-resolvable vendors, plus Workday by explicit hint:

```js
const VENDOR_ORDER = ['gh', 'ashby', 'lever'];
```

Meanwhile `providers/` ships **74** non-internal provider modules. Several are slug-resolvable in exactly the same way as the three that are probed:

| Provider | Slug-derivable endpoint |
|---|---|
| `workable` | `https://apply.workable.com/api/v1/widget/accounts/{slug}` |
| `smartrecruiters` | `https://api.smartrecruiters.com/v1/companies/{slug}/postings` |
| `jobvite` | `https://jobs.jobvite.com/api/company/{slug}/jobs` |
| `teamtailor` | `https://{slug}.teamtailor.com/jobs.rss` |
| `breezy` | `https://{slug}.breezy.hr/json` |
| `recruitee` | `https://{slug}.recruitee.com/api/offers/` |
| `bamboohr` | `https://{slug}.bamboohr.com/careers/list` |

So `discover` reports "unresolved" for companies whose boards the scanner can already read. The mode doc frames unresolved as "JS-rendered portals or non-standard slugs", which is what a user will reasonably conclude — but a Workable or Recruitee company lands in that bucket purely because it was never probed.

This is invisible to the user: nothing in the output says "only 4 vendors were tried."

## 2. A 404 is reported as a transient error with "re-run" advice

`discover-ats.mjs:539-544`:

```js
: (errors.length && !coords)
  ? 'probe error(s) occurred — board status unknown, see errors[] and re-run'
```

with the comment at :541 reading `// Every probe errored (transient network/HTTP)`.

But an `HTTP 404 Not Found` from Greenhouse/Ashby/Lever is **definitive**: that board does not exist. It is not transient and re-running will never change it. Actual output for three of my four companies:

```
Mercado Libre: probe error(s) occurred — board status unknown, see errors[] and re-run
  err: {vendor: gh,    error: HTTP 404 Not Found}
[...]

PR #2975, opened 2026-08-17
fix(discover-ats): a definitive 404 was reported as a transient error with re-run advice (#2883)

Zoubeir23:
Addresses the **second half** of #2883 only — the part you said is "independently useful, smaller, and I'd merge it on its own".

Half 1 (vendor coverage: 4 of 74) is deliberately left alone. You raised two design questions on it — which vendors are meaningfully slug-resolvable, and how to bound the extra requests — and said you'd rather hear @hernanflores's view than decide alone. That conversation shouldn't be pre-empted by a PR.

## The defect

`discover-ats.mjs` reported every probe failure as transient:

```
: (errors.length && !coords)
  ? 'probe error(s) occurred — board status unknown, see errors[] and re-run'
```

But an HTTP 404 from Greenhouse/Ashby/Lever is an **answer**: the vendor was reached and said the board is not there. Re-running produces the same result, forever.

Measured on the reporter's case, before:

```
Mercado Libre: probe error(s) occurred — board status unknown, see errors[] and re-run
  err: {vendor: gh,    error: HTTP 404 Not Found}
  err: {vendor: ashby, error: HTTP 404 Not Found}
  err: {vendor: lever, error: HTTP 404 Not Found}
```

After:

```
no Greenhouse/Ashby/Lever board found. If this company uses Workday, add a hint — …
  errors: [{vendor: gh,    error: HTTP 404 Not Found, httpStatus: 404, definitive: true},
           {vendor: ashby, error: HTTP 404 Not Found, httpStatus: 404, definitive: true},
           {vendor: lever, error: HTTP 404 Not Found, httpStatus: 404, definitive: true}]
```

## The fix

`_http.mjs` already attaches the numeric status to the thrown error; `probeVendor` was discarding it. Keeping it is what lets the caller classify without re-parsing message text.

- `isDefinitiveAbsence()` — 404 and 410 establish absence. Everything else (5xx, timeout, DNS, a rejected redirect, or a candidate whose API URL could not be derived) leaves the question open, because nothing answered it.
- The unknown/re-run branch now requires that **not every** failure was definitive. One transient failure among 404s is enough to keep the outcome unknown: that vendor never answered, so absence is not established.
- Each error entry carries `httpStatus` and, when definitive, `definitive: true` — so the JSON says which of the three states the user is in rather than collapsing them.

## A judgement call worth flagging

The fallback message still ends with "add a Workday hint **and re-run**". I left that: it is a different instruction — the user changes their input first, so the retry can genuinely produce a different answer. What this PR removes is the unconditional retry attached to a status that will not change.

My first test asserted the reason contained no `re-run` at all, and it failed on that legitimate sentence. The assertion was too blunt, not the code; it now pins the exact string that had to go (`errors[] and re-run`).

## Test plan

Six cases in `discover-ats.test.mjs`, driven through the real `resolveCompany` with an injected `ctx` whose probes throw `err.status`.

- [x] 5 of the 6 verified red before the fix.
- [x] The guards pass before and after, which is what makes them guards: all-503 keeps the unknown wording, keeps the re-run advice, and is not marked definitive — and a **mixed** 404/503 run stays unknown, because one unanswered vendor is enough to leave absence unestablished.
- [x] 410 is covered alongside 404.
- [x] `node discover-ats.test.mjs` — 115 passed, 0 failed.
- [x] `node test-all.mjs` — 3967 passed, 0 failed (the single warning is the pre-existing `Dashboard build skipped — go compiler not in env`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved board availability detection for HTTP 404 and 410 responses.
  * Transient or mixed failures now remain marked as unresolved instead of being incorrectly reported as missing.
  * Error details now retain relevant HTTP status information and provide clearer retry guidance.

* **Tests**
  * Added coverage for confirmed missing boards, temporary service failures, and mixed failure scenarios.

### DIFF
--- a/discover-ats.mjs
+++ b/discover-ats.mjs
@@ -428,8 +428,29 @@ export async function probeVendor(company, candidate, ctx) {
     const jobCount = Array.isArray(jobs) ? jobs.length : 0;
     return { status: jobCount > 0 ? 'match' : 'empty', jobCount };
   } catch (err) {
-    return { status: 'error', jobCount: 0, error: err?.message || String(err) };
+    /** @type {any} */
+    const out = { status: 'error', jobCount: 0, error: err?.message || String(err) };
+    // _http.mjs attaches the numeric status to the thrown error; keeping it is
+    // what lets the caller tell a definitive 404 from a transient 5xx instead of
+    // re-parsing the message text (#2883).
+    if (Number.isInteger(err?.status)) out.httpStatus = err.status;
+    return out;
   }
+}
+
+/**
+ * Whether a probe failure ESTABLISHES that no board exists.
+ *
+ * 404 and 410 are answers: the vendor was reached and said the board is not
+ * there. Everything else — 5xx, a timeout, DNS, a rejected redirect, or a
+ * candidate whose API URL could not even be derived — leaves the question open,
+ * because nothing ever answered it.
+ *
+ * @param {{httpStatus?: number}} probeError
+ * @returns {boolean}
+ */
+export function isDefinitiveAbsence(probeError) {
+  return probeError?.httpStatus === 404 || probeError?.httpStatus === 410;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


santifer:
a definitive 404 was reported as a transient failure (#2975):A 404 that reads as 'try again later' sends the user retrying a board that will never exist. Classifying definitive answers as definitive is the point of error taxonomy. Merged! 🚀

PR #2920, opened 2026-08-15
fix(cli): a one-letter typo returned a false all-clear at exit 0 in 4 reporting CLIs

abankar1:
Fixes #2919.

Four reporting CLIs swallowed a mistyped flag and reported a result computed from inputs nobody asked for, at exit 0.

## The sharpest case

`rejection-latency.mjs` had **both** defects `lib/cli-flags.mjs` exists to end, on a script whose output is a blacklist suggestion — so its silent failure mode is a false negative the user then acts on.

Fixture: one company 217 days past the 30-day courtesy threshold.

```
$ node rejection-latency.mjs --tracker sb/applications.md --file sb/ai.md --today 2026-08-15 --summary
  SandboxCo               2026-01-10      217    courtesy   30d
  Suggested data/blacklist.md rows (copy manually — nothing is written for you):
```

One transposed letter:

```
$ node rejection-latency.mjs --traker sb/applications.md --file sb/ai.md --today 2026-08-15 --summary
  No post-interview silence exceeded the configured thresholds.
$ echo $?
0
```

Same exit code, opposite conclusion — `--tracker` fell back to `DEFAULT_TRACKER_PATH` and the tool answered from a file the caller never named. That is `doctor.mjs`'s `--targe` (#2874) with a worse payload.

The `=` form failed identically: `argValue` was `args.indexOf(flag)`-based, which returns -1 for `--tracker=path` and silently discards the value — defect #2401/#2402, in a script that never got the fix.

## Scope

| script | `--flag=value` | mistyped flag | fix |
|---|---|---|---|
| `rejection-latency.mjs` | ❌ own `indexOf` helper | ❌ ignored, exit 0 | `flagValue` + `hasFlag` + `validateFlags` |
| `weekly-digest.mjs` | ✅ own local copy | ❌ ignored, exit 0 | `validateFlags` + a `USAGE` block (it had neither) |
| `process-quality.mjs` | ✅ imports the helper | ❌ ignored, exit 0 | `validateFlags` |
| `detect-reposts.mjs` | ✅ imports the helper | ❌ ignored, exit 0 | `validateFlags` replacing a bare `--help` check |

Every value flag involved either redirects the script to a different data file or moves the threshold that decides what gets reported, so a dropped value changes the answer rather than producing an obvious error.

## Two things worth reviewing

[...]

Issue #2919, opened 2026-08-15
rejection-latency and 3 other CLIs: a one-letter typo returns a false all-clear at exit 0


abankar1:
## Summary

Four reporting CLIs still swallow a mistyped flag and report a result computed from inputs nobody asked for, at exit 0. `rejection-latency.mjs` is the worst of them: it has **both** defects `lib/cli-flags.mjs` was created to end, and its silent failure mode is a false all-clear on a script whose output is a blacklist suggestion.

`lib/cli-flags.mjs` already documents both:

> `args.indexOf('--flag')` returns -1 for `--flag=value`, so a lookup written that way silently DISCARDS the value the caller supplied and the script runs with its default instead (#2401/#2402)

> An unrecognized or mistyped flag (`--dryrun` instead of `--dry-run`) silently falls through to default/live behavior instead of failing fast (#1633/#1635, #2743/#2745, #2744/#2746, consolidated in #2775)

`validateFlags` has since been adopted by `doctor.mjs` (#2874), `check-table-freshness.mjs` (#2873) and `plugin-audit.mjs` (#2813). These four were not swept.

## Reproduce — a false all-clear from one typo

Fixture: a tracker with one `Interview` row and an interview round 217 days ago, well past the 30-day courtesy threshold.

```
$ node rejection-latency.mjs --tracker sb/data/applications.md --file sb/data/active-interviews.md --today 2026-08-15 --summary

  Company                 Last interview  Days   Tier       Threshold
  ----------------------------------------------------------------------
  SandboxCo               2026-01-10      217    courtesy   30d

  Suggested data/blacklist.md rows (copy manually — nothing is written for you):
```

Now misspell one flag — a single transposed letter:

```
$ node rejection-latency.mjs --traker sb/data/applications.md --file sb/data/active-interviews.md --today 2026-08-15 --summary

  No post-interview silence exceeded the configured thresholds.

$ echo $?
0
```

Same exit code, opposite conclusion. The typo'd flag was ignored, `--tracker` fell back to `DEFAULT_TRACKER_PATH`, and the tool reported a clean bill of health for a tracker the caller never named. This is the `--targe` failure #2874 fixed in `doctor.mjs`, except the user here acts on a **false negative** about whether an employer has gone silent.

The `=` form fails identically, because `argValue` is `args.indexOf(flag)`-based (rejection-latency.mjs:79):

```
[...]

PR #2983, opened 2026-08-17
fix(cli): the other half of #2919 — an unusable flag VALUE still defaults at exit 0

### LINKED ISSUE #2982: The other half of #2919: an unusable flag VALUE still falls through to the default at exit 0
## Summary

#2920 made the four reporting CLIs reject a mistyped flag NAME. It left the other half of the same defect open: an unusable flag VALUE still falls through to the default, and the command exits 0 with a report for parameters the caller never supplied.

CodeRabbit flagged this on #2920 as the merge risk and it was right; I did not act on it before merge. Six cases, measured on current `main`:

```
detect-reposts  --window --summary     -> exit 0, 90-day window
detect-reposts  --window abc           -> exit 0, 90-day window
detect-reposts  --window -5            -> exit 0, negative window accepted
process-quality --min-threshold abc    -> exit 0, threshold 1
process-quality --file                 -> exit 0, read the default path
weekly-digest   --dir                  -> exit 0, digested the default dir
```

This is exactly what #2919 described, arriving through the value instead of the name. `--windo 60` is now refused; `--window abc` is not, and both produce the same wrong report.

The family is already inconsistent with itself: `rejection-latency.mjs`, converted in the same PR, exits 2 on a missing or flag-like value because it carried its own `argValue` helper. The other three use bare `flagValue` and default.

## The part that needs a decision, not just a patch

Two existing tests assert the current behaviour **explicitly**:

```js
eq('--window abc falls back to 90', badWindowJson.metadata.windowDays, 90);       // detect-reposts.test.mjs
eq('--min-threshold abc falls back to 1', badThresholdJson.metadata.minThreshold, 1);  // process-quality.test.mjs
```

So the fallback was written down as intended, not merely unexamined. Reversing it is a behaviour change on a documented contract, and that is your call.

The case for reversing: it is the same behaviour you described on merging #2920 — *"worse than a crash: it answers a question nobody asked and nothing flags it"* — and `rejection-latency` already does the opposite in the same family.

The case against: a report with a default window is arguably still useful output, and `--window abc` is a rarer mistake than `--windo`.

I lean to reversing, but say so explicitly so it is not slipped in under a test update.

## Also in scope

Two review nits from #2920 that are now in `main`:

- **@Scott-Emberson**: `new RegExp(typo.replace(/^--/, '--'))` in `tests/cli-flag-validation.test.mjs:55` replaces `--` with itself — a no-op. CodeQL independently flagged it as *"Replacement of a substring with itself"*. It also compiles the typo into a regex, so a typo containing a metacharacter would match as a pattern rather than literally.
- **@Scott-Emberson**: only `rejection-latency` asserts on output; the other three rest on `--help` and an exit code, which cannot distinguish "the value was honoured" from "the value was dropped and the default produced the same exit status".

And **@artemtrofymenko**'s structural point from the same review: a `KNOWN_FLAGS` allow-list in a file makes every in-flight PR that adds a flag a silent coordination dependency — he showed #2702's `--min-span` merging without a textual conflict and then being rejected by its own file at runtime. He suggested a marker comment in each block so the next rebase catches it.

I have all of this working and can open the PR.

abankar1:
Fixes #2982. Follow-up to #2920, closing the review items that landed with it.

#2920 made the four reporting CLIs reject a mistyped flag **name**. It left the other half open: an unusable flag **value** still falls through to the default and the command exits 0 with a report for parameters nobody supplied. CodeRabbit flagged this on #2920 as the merge risk and was right; I did not act on it before merge.

Six cases on current `main`:

```
detect-reposts  --window --summary     -> exit 0, 90-day window
detect-reposts  --window abc           -> exit 0, 90-day window
detect-reposts  --window -5            -> exit 0, negative window accepted
process-quality --min-threshold abc    -> exit 0, threshold 1
process-quality --file                 -> exit 0, read the default path
weekly-digest   --dir                  -> exit 0, digested the default dir
```

All six now exit 2 naming the flag and what it expected. `--windo 60` was already refused; `--window abc` produced the same wrong report and was not.

## ⚠️ One thing needs your decision, not just review

Two existing tests asserted the old behaviour **explicitly**:

```js
eq('--window abc falls back to 90', ..., 90);           // detect-reposts.test.mjs
eq('--min-threshold abc falls back to 1', ..., 1);      // process-quality.test.mjs
```

So the fallback was written down as intended, not merely unexamined. This PR reverses a documented contract, and I would rather flag that loudly than slip it in under a test update.

My case for reversing: it is the behaviour you described on merging #2920 — *"worse than a crash: it answers a question nobody asked and nothing flags it"* — and `rejection-latency`, converted in the same PR, already exits 2 rather than defaulting, so the family was inconsistent with itself. The case against: a default-window report is arguably still useful, and `--window abc` is a rarer slip than `--windo`.

If you would rather keep the fallback, say so and I will drop the two source commits and keep only the test/nit fixes.

## Review nits from #2920, now addressed

**@Scott-Emberson (code owner, `tests/`)** raised two:

1. `new RegExp(typo.replace(/^--/, '--'))` replaced `--` with itself — a no-op, independently flagged by CodeQL as *"Replacement of a substring with itself"*. Worse than useless: it compiled the typo into a regex, so a typo containing a metacharacter would have matched as a pattern rather than literally. Now a plain substring check.
2. Only `rejection-latency` asserted on output. An exit code alone cannot distinguish "the value was honoured" from "the value was dropped and the default produced the same exit status" — the whole defect class. Each of the other three now reads its value back out of the report: detect-reposts' `window: 60 days` line, process-quality's `min threshold: 7` line, weekly-digest's `metadata.range`. The default is asserted alongside each, so a changed default fails loudly instead of silently shifting the baseline.

**@artemtrofymenko**'s structural point from the same review: a `KNOWN_FLAGS` allow-list makes every in-flight PR that adds a flag a silent coordination dependency — he showed #2702's `--min-span` merging without a textual conflict and then being rejected by its own file. Each of the four blocks now carries a marker comment pointing at `KNOWN_FLAGS`, `VALUE_FLAGS` and `USAGE` so the next rebase catches it.

## Design notes

`requireFlagValue` goes in `lib/cli-flags.mjs` rather than being copied three times — `rejection-latency` already had it as a local `argValue`, and leaving it private is what let the other three drift. That file now deletes its copy.

`hasFlag` is paired with `flagValue` deliberately: `flagValue` returns `undefined` for *both* an absent flag and one supplied without a value, so testing it alone would turn a trailing `--window` into a silent default — trading one silent default for another.

`--min-threshold 0` stays legal and has its own assertion: "report every company" is a meaningful threshold, so that flag validates **non-negative** while `--window` and `--courtesy-days` validate **positive**.

## Verification

Checked for discrimination: with the four source files and the shared helper reverted, **exactly the three new value tests fail and the other 17 pass**.

Legitimate usage unchanged — `--window 60`, `--window=60`, `--min-threshold 0`, `--summary`, `--self-test`, `--help`, `--from=`/`--to=` all still exit 0. Self-tests: detect-reposts 8, process-quality 21, weekly-digest 52, rejection-latency 34, all passing. `node test-all.mjs --quick`: 4030 pass / 0 fail (3 warnings pre-existing and environmental).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved command-line integer validation for thresholds and related options.
  * Invalid, fractional, negative, missing, or unsafe values now consistently use the documented fallback values.
  * Valid values, including zero and the largest supported safe integer, are preserved.

* **Tests**
  * Added coverage for valid and invalid threshold inputs to ensure reliable command-line behavior.

### DIFF
--- a/detect-reposts.mjs
+++ b/detect-reposts.mjs
@@ -65,7 +65,7 @@ import { fileURLToPath, pathToFileURL } from 'url';
 import * as yaml from 'js-yaml';
 
 import { normalizeCompanyName } from './invite-match.mjs';
-import { flagValue, validateFlags } from './lib/cli-flags.mjs';
+import { flagValue, validateFlags, safeIntFlag } from './lib/cli-flags.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
 const SCAN_HISTORY_PATH = join(CAREER_OPS, 'data/scan-history.tsv');
@@ -105,35 +105,11 @@ const summaryMode = args.includes('--summary');
 const selfTestMode = args.includes('--self-test');
 // Both numeric flags read through flagValue, so `--min-span=7` behaves exactly
 // like `--min-span 7` — the defect lib/cli-flags.mjs exists to keep the next
-// script from rediscovering. Anything that is not a plain non-negative integer
-// falls back to the default, which is the behaviour --window already had for an
-// unparseable value.
-//
-// The whole string must match, and it must not be negative. parseInt() alone
-// satisfies neither: it reads "7abc" as 7, and it reads "-5" as -5 — and a
-// negative floor silently disables the very guard --min-span exists to set,
-// reporting concurrent openings as reposts again with no indication that the
-// value was rejected. A negative --window likewise rejects every cluster.
-//
-// Tested against Number()/Number.isInteger() as well: Number('') is 0, so an
-// empty value (`--min-span=`) would read as a deliberate zero and disable the
-// floor. A total regex has no such hole.
-// The regex is not sufficient on its own: it happily accepts a 400-digit run of
-// nines, which Number() turns into Infinity. That reproduces this file's
-// original defect in a new place — an infinite floor rejects every cluster, an
-// infinite window is unbounded, and JSON.stringify writes Infinity as `null`,
-// so metadata reports a value that is not the one in effect. A merely UNSAFE
-// integer is the quiet version of the same thing: 9007199254740993 silently
-// becomes ...992. isSafeInteger rejects both.
-const intFlag = (flag, fallback) => {
-  const raw = flagValue(args, flag);
-  if (raw === undefined) return fallback;
-  const text = String(raw).trim();
-  if (!/^\d+$/.test(text)) return fallback;
-  const parsed = Number(text);
-  if (!Number.isSafeInteger(parsed)) return fallback;
-  return parsed;
-};
+// script from rediscovering. The rules and the reasoning now live in
+// lib/cli-flags.mjs's safeIntFlag() so process-quality.mjs stops being the
+// file that lacks them (#2982); the behaviour here is unchanged, including
+// the deliberate fall back to the default rather than an exit.
+const intFlag = (flag, fallback) => safeIntFlag(flagValue(args, flag), fallback);
 const windowDays = intFlag('--window', DEFAULT_WINDOW_DAYS);
 const minSpanDays = intFlag('--min-span', MIN_REPOST_SPAN_DAYS);
 
--- a/detect-reposts.mjs
+++ b/detect-reposts.mjs
@@ -28,7 +28,7 @@ import { fileURLToPath, pathToFileURL } from 'url';
 
 import { roleFuzzyMatch, roleTokens, BASELINE_TOKENS } from './role-matcher.mjs';
 import { normalizeCompanyName } from './invite-match.mjs';
-import { flagValue, validateFlags } from './lib/cli-flags.mjs';
+import { validateFlags, requireFlagValue, isPositiveInt } from './lib/cli-flags.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
 const SCAN_HISTORY_PATH = join(CAREER_OPS, 'data/scan-history.tsv');
@@ -36,6 +36,10 @@ const DEFAULT_WINDOW_DAYS = 90;
 
 // --- CLI args ---
 
+// ADDING A FLAG? Add it here too, and to VALUE_FLAGS if it takes a value.
+// This allow-list is why a flag added in another in-flight PR merges without a
+// textual conflict and is then rejected by its own file at runtime (#2920
+// review). The USAGE block below is the third place.
 const KNOWN_FLAGS = ['--window', '--summary', '--self-test', '--help', '-h'];
 const VALUE_FLAGS = ['--window'];
 
@@ -49,10 +53,15 @@ const USAGE = `Usage:
 const args = process.argv.slice(2);
 const summaryMode = args.includes('--summary');
 const selfTestMode = args.includes('--self-test');
-const windowValue = flagValue(args, '--window');
-const windowDays = windowValue !== undefined
-  ? (Number.isNaN(parseInt(windowValue, 10)) ? DEFAULT_WINDOW_DAYS : parseInt(windowValue, 10))
-  : DEFAULT_WINDOW_DAYS;
+// A present-but-unusable --window used to fall through to the default: no
+// value, `--summary` as the value, "abc", and even "-5" all reported on the
+// 90-day window at exit 0 (#2929). Rejecting the name but not the value left
+// half of #2919's defect in place.
+const windowValue = requireFlagValue(args, '--window', {
+  validate: isPositiveInt,
+  expects: 'a positive integer number of days',
+});
+const windowDays = windowValue !== undefined ? parseInt(windowValue, 10) : DEFAULT_WINDOW_DAYS;
 
 // --- Date helpers ---
 function parseDate(dateStr) {
--- a/detect-reposts.test.mjs
+++ b/detect-reposts.test.mjs
@@ -1179,21 +1179,24 @@ ok('default has metadata', 'metadata' in defaultJson);
 ok('default has clusters array', 'clusters' in defaultJson && Array.isArray(defaultJson.clusters));
 eq('default windowDays = 90', defaultJson.metadata.windowDays, 90);
 
-// Test --window with non-numeric value (falls back to default)
-const badWindowOut = execFileSync('node', [scriptPath, '--window', 'abc'], {
-  encoding: 'utf-8', timeout: 10000,
-  cwd: dirname(scriptPath),
-});
-const badWindowJson = JSON.parse(badWindowOut);
-eq('--window abc falls back to 90', badWindowJson.metadata.windowDays, 90);
-
-// Test --window with no value (falls back to default)
-const noWindowOut = execFileSync('node', [scriptPath, '--window'], {
-  encoding: 'utf-8', timeout: 10000,
-  cwd: dirname(scriptPath),
-});
-const noWindowJson = JSON.parse(noWindowOut);
-eq('--window without value falls back to 90', noWindowJson.metadata.windowDays, 90);
+// BEHAVIOUR CHANGE (#2929): an unusable --window is now REFUSED, where these
+// two assertions previously pinned the opposite — that it silently fell back
+// to 90 and produced a report for a window nobody asked for. That fallback is
+// the same defect #2920 removed on the flag NAME side ("--windo" was ignored),
+// arriving through the value instead; rejection-latency in this same family
+// already exited 2 rather than defaulting. Exit 2 = usage error.
+for (const bad of [['--window', 'abc'], ['--window', '-5'], ['--window']]) {
+  let code = 0;
+  let stderr = '';
+  try {
+    execFileSync('node', [scriptPath, ...bad], { encoding: 'utf-8', timeout: 10000, cwd: dirname(scriptPath) });
+  } catch (e) {
+    code = e.status;
+    stderr = e.stderr || '';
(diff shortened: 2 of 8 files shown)


abankar1:
Good catch — fixed in the latest push, and it was worse than the report described.

`'9'.repeat(400)` is all digits, so `/^\d+$/` passed it, and `parseInt()` then returned `Infinity`, which is `> 0`, so the positivity check passed too. End to end:

```
$ node detect-reposts.mjs --window 999…999   # 400 digits
  "windowDays": null
$ echo $?
0
```

`Infinity` does not survive JSON, so the caller got a **null window at exit 0** — a report for parameters nobody could have meant, which is precisely the class this helper exists to close. The bug was inside the fix.

Both validators now bound on `Number.isSafeInteger` rather than an arbitrary cap: past 2^53 the parsed value is no longer the number the caller typed, so honouring it would report on a different window than the one supplied — the same failure by a different route.

`--min-threshold 0` stays legal and keeps its own assertion. The 400-digit case joins the existing bad-value loop, so it is checked for the same properties as `abc`, `-5` and the no-value form: exit 2, and no default-window report.

20/20 in the file, 4030/0 in `test-all --quick`.

Scott-Emberson (approved):
Code owner for `tests/`. The new cases give the residual real teeth: each bad-value run (`abc`, `-5`, missing value, a 400-digit unsafe integer) asserts exit 2 alongside a valid-value control at exit 0, across detect-reposts, process-quality, and weekly-digest, so a silent-default regression flips 2 back to 0 and fails. The old no-op regex line is gone. One low nit, non-blocking: the three positive-control runs pass no `--file`/`--dir` fixture, so they read the real `data/` and `interview-prep/sessions/`, which contradicts the file's own "every path is a tmpdir fixture" header. Read-only and the asserted lines print regardless of ambient data, so it is a consistency nit, not a false-pass. Approving.

santifer:
This is ready, @abankar1: the isSafeInteger bound with the reasoning written down, the positive/non-negative split per consumer, and the old fallback-pinning tests replaced by refusal tests are all exactly right. One thing before merge: the code comments cite #2929 (~5 sites across the four CLIs + lib), but #2929 is an unrelated web test PR. The issue this closes is #2982. A quick s/2929/2982/ and this merges.

Solaris-star (commented):
I reviewed the exact head `202a7356c49e4b250821983e8b53fad09a2741c2` against the current `upstream/main` checkout.

Local verification on that SHA:

- `node --test tests/cli-flag-validation.test.mjs` — 20 passed, 0 failed
- `node detect-reposts.mjs --self-test` — 8 passed, 0 failed
- `node process-quality.mjs --self-test` — 21 passed, 0 failed
- `node weekly-digest.mjs --self-test` — 52 passed, 0 failed
- `node rejection-latency.mjs --self-test` — 34 passed, 0 failed
- `node test-all.mjs --quick` — 4030 passed, 0 failed, 3 repository warnings
- tracked/untracked secret scan against `upstream/main` — 0 new hits

The GitHub API currently reports all 12 check runs successful, including CodeQL, the platform test jobs, and GitGuardian. One merge-readiness limitation remains: the PR API currently reports `mergeable=false` / `mergeable_state=dirty`; I did not attempt to rebase or modify your branch. The maintainer’s explicit remaining request is still the comment-only `#2929` → `#2982` correction. After that correction and a clean rebase/update, the verification above should be rerun on the new head; this is not a blanket approval for a changed commit.

santifer:
The conflict is mine, @abankar1: I merged #2702 today (detect-reposts concurrent-openings fix) and it landed in the same file your value-validation touches. An update-branch from my side failed on it, so this needs your rebase. The substance of this PR is already reviewed-positive (the six measured cases and closing what CodeRabbit flagged on #2920 is exactly right), and #3071 is going in as soon as its re-run finishes: this one merges right behind it once rebased.

abankar1:
Rebased — but the substance changed underneath this PR while it sat, and most of it is now yours already. Rescoped rather than force-fitted, and the policy argument is withdrawn.

**What landed on main and superseded what:**

| this PR's claim | status on main now |
|---|---|
| missing operand silently dropped (`--window --summary`, bare `--dir`, …) | fixed — #3087's `requireOperand` |
| unsafe/400-digit integers accepted (`Infinity` → `"windowDays": null`) | fixed in `detect-reposts` — the `intFlag` helper from #2702, same `isSafeInteger` reasoning I'd used |
| an unusable value should **exit 2** rather than fall back | **you decided against this**, explicitly |

That last row is the important one. #2702's comment says it in as many words: *"Anything that is not a plain non-negative integer falls back to the default, which is the behaviour `--window` already had for an unparseable value."* My earlier commit flipped two tests that pinned exactly that fallback. **I've dropped that.** Erroring is no longer this PR's position — it argued for a policy you've since settled the other way, with the reasoning written down.

**What is still genuinely broken, and is now the whole PR:**

`process-quality.mjs` has no shape check at all — so unlike `detect-reposts` it doesn't even fall back. It *accepts* the value and reports it as the threshold in effect:

```
--min-threshold 999999999999999999999  ->  "minThreshold": 1e+21
--min-threshold 3.5                    ->  "minThreshold": 3
--min-threshold 9007199254740993       ->  ...992
```

That first line is exactly the failure `detect-reposts.mjs` now documents `isSafeInteger` as existing to prevent — *"the metadata reports a value that is not the one in effect"* — one file over, with the guard absent rather than differently worded. And it reaches `aggregateProcessQuality`, so the report is built on it.

`abc` and `-5` already landed on 1, but via the `>= 0` clamp rather than any rule about the value's shape.

**Shape of the fix:** `safeIntFlag` moves to `lib/cli-flags.mjs` carrying #2702's reasoning verbatim, `detect-reposts` delegates to it (one-line change, no behaviour change — verified `60`→60 and `abc`/`-5`/21-digit still →90), and `process-quality` gains the guard with the **same fallback policy**, not an exit.

Three commits, ~40 lines of source. The old five-commit version is gone.

**On the `#2929` → `#2982` correction** you gated merge on: those comment sites no longer exist — the commits carrying them are dropped. Every reference in the rescoped branch cites #2982. Worth a grep on your side rather than my word for it.

**@Scott-Emberson's nit** is resolved by the rescope rather than by me: the three positive-control runs that read the real `data/` and `interview-prep/sessions/` are gone with those tests. The new assertions run `process-quality` with an explicit `--min-threshold` and read only `metadata.minThreshold`, which is ambient-data-independent.

Verified: with `process-quality.mjs` reverted, exactly the four genuinely-broken cases fail and the honoured-value controls (`0`, `7`, `9007199254740991`) pass. `process-quality.test.mjs` 91/91, `detect-reposts.test.mjs` 255/255, self-tests 21/21 and 40/40, `test-all --quick` 5308 pass / 0 fail.

If you'd rather this were just the `process-quality` two-line guard without moving `safeIntFlag` into `lib/`, say so — I extracted it because the alternative is a second copy of reasoning that took a whole comment block to justify, but the copy would be smaller than the move.

artemtrofymenko:
Checked the delegation against the original rather than reading it, since it is my code being moved and "no behaviour change" is the claim that matters most here.

**Behaviour-neutral on every case the original's comment names as load-bearing**, not just the three you spot-checked:

```
                       intFlag (#2702)   safeIntFlag
""            empty        90                90     <- Number('') is 0, the hole
"   "                      90                90
"-5"                       90                90     <- a negative floor disables the guard
"7abc"                     90                90     <- parseInt would read 7
"3.5"                      90                90
"9007199254740993"         90                90     <- unsafe: silently ...992
"9"x400                    90                90     <- Infinity, JSON writes null
"1e3"                      90                90
"+60"                      90                90
"007"                       7                 7
" 60 " / "60\n" / 60       60                60
undefined / null / true    90                90
                                        19/19 identical
```

**And the reasoning moved with it**, which is the part whose loss would have been the real regression. All four hazards are in the new docblock, the fallback-rather-exit policy is stated, and the boundary with `requireOperand` — presence is its job, shape is this one's — is drawn explicitly. A one-line alias with a pointer stays behind in `detect-reposts.mjs`, so the call sites read the same.

On your open question to @santifer: **the move, not the copy.** The repo already answered this twice — `compileKeyword` came out to `title-keywords.mjs` because a second path filtered titles, and the copies had drifted three ways before anyone noticed; `hermeticGitEnv` came out to `tests/helpers.mjs` for the same reason, with the merge comment naming why: a regression case that hand-rolls its own version keeps passing when the pin is dropped from the original. A guard whose justification takes a twenty-line comment is exactly the thing that drifts silently, because the second copy inherits the code and not the argument.

The rescope reads well beyond that. Withdrawing the exit-2 policy by quoting the comment that settled it against you, and dropping the two commits that flipped the tests pinning it, is the same move @IaroslavMazur made on #3103 this week — and @santifer merged that decision citing the retraction as what settled it.

One thing I did not check: `process-quality.mjs`'s own behaviour, or the 91/91 and 5308-pass runs. Those are yours and outside what I can add to.

santifer:
This one earned its merge twice, @abankar1: once for the fix and once for the honest rescope after main moved underneath it — withdrawing the exit-2 policy argument instead of force-fitting it is exactly the right call. And thanks @artemtrofymenko for verifying behaviour-neutrality against your own original case by case: that review is what made this a same-day merge once rebased. Merged. 🚀

PR #2986, opened 2026-08-17
fix(portals): block German student and support postings in the example title filter

KehanLiu:
## What does this PR do?

The example `title_filter.negative` excludes `Junior` and `Intern` but not their German
equivalents, while `title_filter.positive` already carries German AI keywords (`KI`,
`Künstliche Intelligenz`). A scan of one large DACH employer therefore returns mostly
student postings.

Verified against the shipped template with `buildTitleFilter`. These all PASS today and
REJECT after the change:

- `Werkstudent AI (m/w/d)`
- `Praktikum Künstliche Intelligenz`
- `Masterarbeit: KI-Agenten`
- `Abschlussarbeit KI`
- `PreMaster Programm KI`
- `Master Thesis LLM Agents`
- `Bachelor Thesis AI`
- `Customer Care Agent`
- `Kundenbetreuung KI`

Nothing wanted is lost — `Speech Synthesis Engineer`, `Applied AI Architect`,
`Forward Deployed Engineer`, `Senior AI Engineer`, `Conversational AI Engineer` and
`KI Engineer` all still pass, before and after.

One deliberate choice, documented in a comment next to the entries: `"Master Thesis"` /
`"Bachelor Thesis"` rather than a bare `"Thesis"`. Keywords longer than three characters
match as plain substrings, so `"Thesis"` would also reject `"Speech Synthesis Engineer"` —
and `"Speech"` is a positive keyword in the same file. I checked both variants before
picking.

A companion PR adds SAP, Bosch and Delivery Hero to the same file; those are the employers
where these titles arrive in volume, which is what surfaced this.

## Related issue

None — bug fix, which CONTRIBUTING.md exempts from issue-first.

## Type of change

- [x] Bug fix

## Checklist

- [x] I ran `node test-all.mjs` and all tests pass (4053 passed, 0 failed)

## Summary by CodeRabbit

* **New Features**
  * Added German title filters to exclude irrelevant student, apprenticeship, thesis, and customer-support postings from search results.
  * Improved matching accuracy for roles using the “Agent” keyword.

### DIFF
--- a/templates/portals.example.yml
+++ b/templates/portals.example.yml
@@ -349,6 +349,29 @@ title_filter:
     - "Oracle EBS"
     - "Mainframe"
     - "COBOL"
+    # -- German student / apprenticeship postings (DACH market) --
+    # Large DACH employers publish these in volume, and the German AI keywords
+    # in title_filter.positive ("KI", "Künstliche Intelligenz") match them, so
+    # without these a scan of one big employer returns mostly student ads.
+    - "Werkstudent"
+    - "Praktikant"
+    - "Praktikum"
+    - "Masterarbeit"
+    - "Bachelorarbeit"
+    - "Abschlussarbeit"
+    - "PreMaster"
+    - "Duales Studium"
+    - "Ausbildung"
+    # Qualified rather than a bare "Thesis": keywords longer than three
+    # characters match as plain substrings, so "Thesis" would also reject
+    # "Speech Synthesis Engineer" — and "Speech" is a positive keyword above.
+    - "Master Thesis"
+    - "Bachelor Thesis"
+    # -- Support roles that match on "Agent" --
+    # "Customer Care Agent" and its German equivalent otherwise pass the
+    # "AI Agent" / "Agent" side of the positive list.
+    - "Customer Care"
+    - "Kundenbetreuung"
   seniority_boost:
     # These prefixes add relevance but are not required
     - "Senior"


artemtrofymenko:
Ran this against a real scan corpus rather than reading it — 2357 unique observed titles, verdict before your change vs after. Five change, and three are exactly what you aimed at:

```
true -> false   Praktikant Marketing Automation mit KI (m/w/d)…        <- Praktikant
true -> false   Werkstudent Gebäudeautomation (m/w/d)                  <- Werkstudent
true -> false   Werkstudent:in GTM Engineering & AI Automation (m/w/d) <- Werkstudent
```

The other two are the finding:

```
true -> false   AI Support Delivery Manager: Customer Care             <- Customer Care
true -> false   AI Agents Manager: Customer Care                       <- Customer Care
```

Both are senior AI roles **for** a customer-care organisation, not support-agent jobs. `Customer Care` there is the domain after the colon, not the role. Anchoring would not help — it is already two whole words in both — so the distinguishing edit is to negate the role rather than the domain: `Customer Care Agent`, `Kundenbetreuer`, `Customer Care Representative`. That still catches the two you named in the comment and leaves these.

I also ran the dead-positive check from #2970 across the merged result — a positive that the negative list matches inside its own text can never be satisfied by any title. **Your 13 new negatives kill none of the 37 positives.** Worth saying explicitly, because `Agent` and `Agentic` are positives and negating the support roles by domain instead of by keyword is what kept that clean.

## On the `Thesis` workaround

> Qualified rather than a bare `Thesis`: keywords longer than three characters match as plain substrings, so `Thesis` would also reject `Speech Synthesis Engineer` — and `Speech` is a positive keyword above.

You found the same defect I hit from the other side: `Intern` rejecting every `Internal` and `International`, and killing the `Internal Tools` positive outright. #2970 adds a `word:` prefix so a keyword can ask for whole-word matching at any length, which would let you write `word:Thesis` and drop the two-line workaround. Not a request to change anything here — just so you know it is coming, and that your comment is the clearest independent evidence that the substring rule bites more than one person.

## Coordination

#2970 also edits `templates/portals.example.yml`, so whichever of us lands second will need a trivial rebase. Different regions of the negative list, so no textual conflict either way.

*Scope: one personal scan corpus, N=2357, titles only — the five above are observed, not constructed. I have not run the full suite against this branch.*

santifer:
Welcome to career-ops, @KehanLiu! The qualified-thesis detail is what sold this: knowing that keywords match as plain substrings, so a bare "Thesis" would reject a Speech Synthesis Engineer — with "Speech" sitting right there in the positive list — is exactly the reasoning this filter needs. Verified titles on both sides of the change made it a one-read review. Merged — your first contribution. Glad to have you here. 🚀

Issue #2993, opened 2026-08-17
Applying to a LinkedIn posting is blocked by the authwall


EliRobinson:
## The problem

A LinkedIn posting cannot be applied to from the web Apply flow, and cannot be evaluated well either.

The apply session browses in a fresh, cookie-less context. That is the right default for a public ATS form and exactly wrong for LinkedIn: it serves its authwall, and the flow dies. LinkedIn's guest endpoint will tell you *whether* a job applies offsite, but not *where*, so the employer's real application URL has to be reconstructed from what the posting exposes.

Since LinkedIn is where a large share of postings are found, "paste the link and let career-ops take it from here" currently has a hole in it precisely where users spend the most time.

## Why this is on the core path

The core path is discover → evaluate → tailor → apply → track. This is the apply step failing for one very common source. It needs no hosted service, no scraping infrastructure and no auto-apply: it is a URL resolution problem, solved locally, with the human still reviewing and submitting.

## Proposed shape

A zero-token, no-auth, no-browser resolver:

1. Read company + title from LinkedIn's public guest page.
2. Find the employer's board with the existing `discover-ats.mjs`.
3. Match the posting with the existing `jd-similarity.mjs`.

The design decision I would most like a second opinion on: **it should refuse rather than guess.** Resolve only on a unique high-confidence title match; treat a seniority difference as disqualifying; return ranked candidates for the user to choose from in every other case. Sending an application to the wrong requisition is worse than asking a question. That means the UI has three outcomes to handle, not one: resolved, ambiguous, unresolved.

A confident resolve (or a user's pick) would be recorded on the report so the next apply for that offer skips resolution entirely. That needs somewhere to put it. My suggestion is an optional `**Apply URL:**` header line, with `**URL:**` keeping its current meaning as the canonical link the tracker records, never overwritten.

**Easy Apply is deliberately out of scope for the resolver.** Its form lives on LinkedIn, so there is no external URL to reconstruct. That case needs a different answer (an opt-in, signed-in browser profile that the user signs into themselves, never their everyday Chrome profile, with career-ops never handling the password), and it is reasonable to treat it as a separate decision from this one.

## Questions for the maintainers

1. Is the `**Apply URL:**` header line the right place to record a resolved link, or would you rather it lived somewhere else (the tracker, the Machine Summary block)?
2. Is refuse-over-guess the behaviour you want, given it means more user prompts?
3. Does the Easy Apply / signed-in-profile half belong in core at all, or is that plugin territory?

[...]

PR #2997, opened 2026-08-17
feat: resolve LinkedIn postings to the employer's real application URL

EliRobinson:
## Related issue

Addresses #2993.

## What does this PR do?

Resolves a LinkedIn posting to the employer's real ATS application URL, so the web Apply flow can reach a form instead of an authwall.

`linkedin-apply.mjs` reads company + title from LinkedIn's public guest page, finds the board via the existing `discover-ats.mjs`, and matches the posting via the existing `jd-similarity.mjs`. Zero tokens, no auth, no browser.

This started as one worked example so #2993 had something concrete to react to, and it is still open to a different structure. Per @santifer's note above, URL resolution to the employer's real application page sits on the welcome side of the boundary written down in #2998 (the system resolves, captures, drafts and pre-fills; only the human submits), so it has graduated from draft. The three questions in #2993 are still genuine, and answering them may still change this code.

## The design decision worth reviewing

**It refuses rather than guesses.** It resolves only on a unique high-confidence title match; a seniority difference disqualifies outright; everything else returns ranked candidates for the user to pick from. Sending an application to the wrong requisition is worse than asking a question.

That is why the UI handles three outcomes rather than one: resolved (go), ambiguous (pick from candidates), unresolved (paste the link yourself). If you would rather it guessed more aggressively and asked less, that is a one-line policy change plus a lot less UI.

## Where the resolved link is recorded

A confident resolve, or a user's pick, is written to the report as an optional `**Apply URL:**` header line, so the next apply for that offer skips resolution entirely. `**URL:**` keeps its meaning as the canonical link the tracker records and is never overwritten.

This is question 1 in the issue. If you want it in the tracker or the Machine Summary instead, the writer is one function.

## Easy Apply

Out of scope for the resolver: the form lives on LinkedIn, so there is no external URL to reconstruct. Included here is an opt-in signed-in browser profile (`apply.signed_in_profile`, toggled from web Config) which runs the apply session in a dedicated profile the user signs into themselves. Never the user's everyday Chrome profile, and career-ops never sees, types or stores the password.

**This half is the most separable part of the PR.** It is question 3 in the issue, and if you would rather it were a plugin, or simply not now, I will strip it out and the resolver stands on its own.

## Rebase notes

Rebased onto current `main`. Four things adapted rather than taken as-is:

- `apply-button.tsx` picked up `n` / `from` on `apply.open()` from main, so the resolve flow passes them through; the Apply page still marks the row Applied and returns the user where they were.
- `report-view.tsx` picked up `pdfReadyFromIndex` and `companyPresentation()` from main; `ApplyButton` now takes main's `pdfReady` alongside the new `applyUrl`.
- `update-system.mjs`: main moved `jd-similarity.test.mjs` under `tests/`, so `SYSTEM_PATHS` keeps main's path and adds `linkedin-apply.mjs` beside it.
- `linkedin-apply.mjs` adopts `isMainModule(import.meta.url)` from `lib/is-main-module.mjs` rather than hand-rolling the entry-path comparison, per the convention `tests/main-guard-convention.test.mjs` now enforces (#3170).

`js-yaml` is imported as `import * as yaml`, per main's guard against the default-import form.

## Tests

- `node linkedin-apply.mjs --self-test` → 58 passed, 0 failed
- `node test-all.mjs` → 7691 passed, 0 failed (13 pre-existing warnings, all unrelated: missing local user data and `santifer.io` strings in `dashboard/`, `funding.json`, `HIRED.md`)
- `web`: `pnpm test` → 463 passed, 0 failed; `tsc --noEmit` and `next build` clean

## Type of change

- [ ] Bug fix
- [x] New feature

## Checklist

- [x] I opened an issue first (#2993)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -96,6 +96,7 @@ AI-powered, CLI-agnostic job search automation: pipeline tracking, offer evaluat
 | `scan.mjs` | Zero-token portal scanner (Greenhouse/Ashby/Lever APIs, zero LLM cost) |
 | `scan-ats-full.mjs` | Reverse-ATS keyword-first scanner over full public ATS datasets (Greenhouse/Lever/Ashby/Workday/iCIMS), filtered by portals.yml `title_filter`/`location_filter` — no company list needed; checkpoints every 500 companies, `--resume` continues an interrupted sweep |
 | `scan-interamt.mjs` | Playwright browser scanner for Interamt.de (German public sector portal — Apache Wicket, no REST API) |
+| `linkedin-apply.mjs` | Resolves a LinkedIn posting to the employer's real ATS application URL (LinkedIn's guest endpoint says whether a job applies offsite but not where). Reads company + title from the guest page, finds the board via `discover-ats.mjs`, matches the posting via `jd-similarity.mjs`. Zero tokens, no auth, no browser. Refuses rather than guesses: resolves only on a unique high-confidence title match, a seniority difference disqualifies outright, everything else returns ranked candidates (JSON or `--summary`) |
 | `check-liveness.mjs` / `liveness-core.mjs` | Job posting liveness checker + shared logic (expired signals win over generic Apply text) |
 | `set-status.mjs` | Canonical tracker-row update: `node set-status.mjs <report#\|company> <State> [--note] [--force]` — strict states.yml validation, report-link mismatch guard, shared lock, atomic write |
 | `invite-match.mjs` | Fuzzy-match a pasted interview invite (company, date, req ID) against the tracker, ranking candidates when a company has multiple entries (JSON or `--summary`) |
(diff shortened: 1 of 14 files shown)


santifer:
Core-half note, @EliRobinson: URL resolution to the employer's real application page is machine work on the welcome side of the boundary just written down in #2998 (the system resolves, captures, drafts and pre-fills; only the human submits). No core objection to this example graduating from draft.

EliRobinson:
@coderabbitai review

santifer:
Thanks for taking this out of draft, @EliRobinson. Where it stands from my side:

- The branch is **conflicting** with `main` right now (a lot landed since 21-aug), so it needs a rebase before anyone can review the diff for real.
- It has two halves. Everything under `web/` is reviewed by the web track, so I'm adding `area:web`. The core half (`linkedin-apply.mjs` as a new top-level script, the `AGENTS.md` section, the `test-all.mjs` and `update-system.mjs` entries) I review myself once it's rebased: I can see you already added the SYSTEM_PATHS entry for the new script, which is the thing new top-level files usually miss.
- Direction unchanged since my note on 21-aug: resolve, capture, draft and pre-fill are machine work; the submit stays with the person.

Ping me when the rebase is up and I'll take the core half the same day.

Issue #2995, opened 2026-08-17
No way to add a job URL you found yourself; LinkedIn URLs evaluate the authwall


EliRobinson:
## The problem

There is no way to hand the web app a posting you found yourself.

The Pipeline page works only through what the scanner already put in the inbox. `quick-evaluate.tsx` does fire the evaluate worker, but it is gated behind an in-between lifecycle phase, so an established install never sees it. A user who finds a job on LinkedIn, in a newsletter, or from a friend has to drop to the CLI or edit `data/pipeline.md` by hand.

## Why this is on the core path

This is the discover step. The scanner covers boards career-ops knows about; a human covers everything else, and right now the human has no door into the same pipeline.

## Proposed shape

Surface the existing quick-evaluate path and add an "Add job URL" entry point on the Pipeline page. Extending what is there rather than building a parallel component.

**The part that needs a real decision is LinkedIn.** A `/jobs/view/` URL serves a login wall to a headless agent, so evaluating one today produces a confident report about an authwall rather than about the job. Normalizing the URL to LinkedIn's public guest endpoint fixes what the agent reads, but then the URL it reads is not the URL a human should click.

My suggestion: give the evaluate prompt an optional `fetchUrl`, so the agent READS the mirror and RECORDS the canonical link. A tracker full of guest-API links would be useless. Normalization happens server-side, before any tokens are spent on a URL that was never going to work.

That clause has to sit in step 1 of the prompt, next to the "use WebFetch to read the posting" instruction. Appending it at the tail puts it after the text saying nothing should follow the final VERDICT line, which is the least-attended position available.

Because #2185 freezes the evaluate prompt as a value, the no-mirror case must stay byte-identical: the clause returns "" when there is no mirror, and the suite should pin that both for an absent `fetchUrl` and for `fetchUrl === input`.

## Questions for the maintainers

1. Is a mirror-read/canonical-record split the right model, or would you rather LinkedIn URLs were refused with a clear message until a resolver exists?
2. Should "add a URL" write into `data/pipeline.md` (so the CLI sees it identically) or go straight to an evaluate run?
3. Is extending `quick-evaluate.tsx` the right move, or has that component's gating a purpose I have misread?

## Example implementation

[...]

PR #2999, opened 2026-08-17
feat(web): add a job URL from the pipeline, with LinkedIn authwall handling

EliRobinson:
## Related issue

Addresses #2995.

## ⚠️ This is an example, not a proposal that the structure is right

**I am opening this as one worked example of how #2995 could be resolved, so the discussion has something concrete to react to. It is explicitly open to a better structure.** If you would rather this were shaped differently, split up, or replaced with another approach, say so and I will redo it. The three questions at the bottom of #2995 are genuine, and answering them may well change this code.

## What does this PR do?

Adds an "Add job URL" entry point on the Pipeline page and surfaces the existing quick-evaluate path, so a posting the user found themselves can enter the same pipeline the scanner feeds. It extends `quick-evaluate.tsx` (which already fires the evaluate worker but was gated behind an in-between lifecycle phase) rather than building a parallel component. That is question 3 in the issue.

## The part that needs a real decision: LinkedIn

A `/jobs/view/` URL serves a login wall to a headless agent, so evaluating one today produces a confident report about an authwall rather than about the job.

A pasted URL is normalized and LinkedIn is resolved to its public guest endpoint, and the evaluate prompt takes an optional `fetchUrl`: the agent READS the mirror but RECORDS the canonical link. A tracker full of guest-API links would be useless to click. Normalization happens server-side, before tokens are spent on a URL that was never going to work.

The clause sits in step 1 of the prompt, next to the "use WebFetch to read the posting" instruction it belongs to. It carries the one rule the whole design depends on, and appending it at the tail would put it after the text saying nothing should follow the final VERDICT line.

This is question 1 in the issue. If you would rather LinkedIn URLs were simply refused with a clear message until a resolver exists, that is a much smaller change and I will make it instead.

## The #2185 freeze is respected

`mirrorClause()` returns `""` when there is no mirror, so an ordinary posting's prompt is byte-identical to before. The suite pins that both ways: absent `fetchUrl` and `fetchUrl === input` both equal the plain call.

## One adjacent fix, happy to split it out

Found while building this: evaluate runs got a 285s kill timer while pdf already got 600s, and an evaluation doing web research routinely exceeded it. The route's `maxDuration` has room, and unlike pdf, evaluate has no post-agent render phase.

That timer made a second bug reachable. When it landed after the report had been written and the tracker row merged, the honesty gate reported "this evaluation didn't save a report", which was false. That branch is split so a cut-off run that DID persist says so and tells the user to reload. It is still surfaced as an error: a half-finished evaluation must never be banked as a confident score.

**If you would rather review that on its own, say the word and I will lift it into a separate PR.**

## Responding to the review (commit 2)

Per the decision in this thread, LinkedIn URL parsing must not be a private copy in `web/`. The shared importable module has not been extracted yet: `liveness-api.mjs` still keeps the rung inside its `ATS_PROVIDERS` table and imports `./user-agent.mjs`, so there is nothing the client bundle can depend on. Both consumers here are client-reachable (`job-store.tsx` normalizes a pasted URL before it fires a worker), so this adopts the pattern the repo already uses for exactly that problem instead of waiting:

`web/src/lib/core/linkedin-url.mjs` is an algorithm mirror with a parity test, the same shape as `url-key.mjs` and `normalize-text-key.mjs`. `web/tests/lib/linkedin-url.test.mjs` loads the real `liveness-api.mjs` through `resolveAtsApi()` and compares both sides on every recognized shape and every rejected one, so the build fails if they drift. When the extraction lands, deleting the mirror and importing the core is a one-liner and the parity test is what proves the swap changed nothing.

**The mirror is byte-aligned with core, which makes it looser in one place and stricter in two:**

- the slug form accepts 1+ digits, not 6+ (`some-title-2` resolves to id `2`)
- the path is anchored, so `/jobs/view/{id}/extra` no longer resolves
- the path is case-sensitive, so `/JOBS/VIEW/{id}` no longer resolves

The 6-digit floor was mine and I still think it is right, but a mirror that is stricter than its core is drift wearing a better hat: the two disagree on real inputs while the parity test passes on whatever cases someone wrote down. If the floor is worth having it belongs in `liveness-api.mjs`, where the liveness ladder gets it too. Happy to send that separately.

**Second half, found while doing the first.** `postingKey()` was a second canonical URL key sitting alongside `core/url-key.mjs`'s `normalizeUrl`, with its own tracking-param denylist and its own idea of canonical shape. `normalizeUrl` is already the repo's one identity key (parity-tested against the root `url-key.mjs`) and is what `explore-ai.ts`'s `canon()` and scan-history dedup use. Two keys for one question means the same posting can read "new" on Explore and "in your pipeline" on Today, off the same card.

`postingKey` is now two stages with one key at the end: `normalizeJobUrl` collapses LinkedIn's three spellings of one posting (`normalizeUrl` cannot, being host-agnostic by design), then `normalizeUrl` produces the key. `normalizeJobUrl` keeps its own denylist because it does a different job, cleaning a link a human will click in the tracker; the header says so in as many words so the two are not re-merged later. It does not return `normalizeUrl`'s `''` sentinel: `''` means NO KEY and must never match another `''`, which is exactly what the Set membership tests calling this would do.

Behavior change worth naming: the key no longer carries a trailing slash, since url-key owns that shape now. Nothing persists a `postingKey` (`job.input` is `normalizeJobUrl`'s record form, and the two comments that claimed otherwise are fixed), so this only moves in-memory comparisons. `registry.ts` double-applies `postingKey` through `ctx.jobForUrl`, which was invisibly safe before and is now pinned by an idempotence test: if it ever breaks, the duplicate-spend guard silently re-fires an evaluation the user already paid for.

## Rebase notes

Rebased onto current `main`, merging with the changes that landed on the same lines rather than overwriting them:

- `buildPrompt` now takes main's `postedAt` and `lang` alongside this branch's `fetchUrl`. `postedAt` is still looked up by the URL the client sent, not by the rewrite, since the inbox and scan-date map are keyed by the canonical URL.
- The checkout precondition keeps main's language-aware behavior: `evaluate` is gated on the mode file the prompt will actually read (`lang.evalModeFile`), not always `modes/oferta.md`, and main's `turbopackIgnore` annotations are preserved. The `KINDS` row still names the default; the configured market replaces it where the config is read.
- The agent kill timer is main's `killMsForKind()` in `run-cli-support.mjs`, not a column in `KINDS`. Both this branch and main independently lengthened that timer; main's version is newer, tested in its own module, and carries the #3124 reasoning, so the table's duplicate was removed rather than kept as a second source of truth.
- Main's `stderrErrorSnippet` detail survives the honesty-gate extraction: `evaluateRunOutcome` takes an optional `stderrSnippet`, appended only on the "hit an error before finishing" arm and only when no structured error already sent its own message.
- Report detection keeps main's `reportEntries()` + `hasNewCompletedReport()` (which checks for a *completed* report, not just a count) while adopting this branch's registry-driven `k.persistsReport` flag.
- Internal planning docs from the original branch are not included.

**One overlap worth a reviewer's eye:** main now short-circuits on `killedByTimeout` *before* the honesty gate, and its message says the run "isn't recorded as a result". This branch's arm 3 in `run-outcome.mjs` handles the case where a report *did* land before the cut-off. Main's earlier check now wins for timeouts, so arm 3 only fires for non-timeout kills. Both are correct as written, but if you would rather the timeout message also distinguished "cut off, but a report landed", say so and I will fold arm 3's wording into it.

## Tests

- `node test-all.mjs` → 7697 passed, 0 failed (13 pre-existing warnings, all unrelated: missing local user data, and `santifer.io` / `Santiago` strings in `funding.json`, `HIRED.md` and `dashboard/`)
- `node --test "tests/**/*.test.mjs"` in `web/` → 517 passed, 0 failed
- `tsc --noEmit` clean

## Type of change

- [ ] Bug fix
- [x] New feature

## Checklist

- [x] I opened an issue first (#2995)

### DIFF
--- a/web/src/app/actions/registry.ts
+++ b/web/src/app/actions/registry.ts
@@ -33,8 +33,9 @@ export type StartJobInput = {
   batchId?: string;
 };
 
-// Same shape job-store.tsx's startEvaluate takes — a raw URL in, a canonical
-// postingKey out as the job's `input`. See its header for why this exists.
+// Same shape job-store.tsx's startEvaluate takes — a raw URL in, normalizeJobUrl's
+// canonical url out as the job's `input`. See its header for why this exists, and
+// job-url.mjs's for why that canonical url is not the same thing as postingKey.
 export type StartEvaluateInput = {
   url: string;
   title?: string;
--- a/web/src/app/actions/registry.ts
+++ b/web/src/app/actions/registry.ts
@@ -11,6 +11,7 @@
 import type { Application, InboxJob } from "@/lib/career-ops";
 import type { Job } from "@/components/jobs/job-store";
 import { normalizeTextKey } from "@/lib/core/normalize-text-key.mjs";
+import { postingKey } from "@/lib/job-url.mjs";
 
 export const AUTO_FIRE_MAX = 3; // fire ≤3 evaluations silently; confirm above that
 export const BATCH_CAP = 12; // hard ceiling on a single fan-out
@@ -32,10 +33,21 @@ export type StartJobInput = {
   batchId?: string;
 };
 
+// Same shape job-store.tsx's startEvaluate takes — a raw URL in, a canonical
+// postingKey out as the job's `input`. See its header for why this exists.
+export type StartEvaluateInput = {
+  url: string;
+  title?: string;
+  subtitle?: string;
+  page?: string;
+  batchId?: string;
+};
+
 export type ActionCtx = {
   push: (path: string) => void; // router.push — section/detail change
   replace: (path: string) => void; // router.replace — incremental filter tweak
   startJob: (opts: StartJobInput) => string | null;
+  startEvaluate: (opts: StartEvaluateInput) => string | null;
   inbox: InboxJob[];
   applications: Application[]; // tracker snapshot — resolve #n → company/role for confirms
   jobForUrl: (url: string) => Job | undefined; // skip-if-done / retry logic
@@ -143,14 +155,20 @@ const ACTIONS: Record<string, ActionDef> = {
     sideEffect: "spend",
     run: (raw, ctx) => {
       const url = raw.url;
-      if (!isStr(url) || !/^https?:\/\//i.test(url)) return { status: "ignored", note: "invalid url" };
-      const ex = ctx.jobForUrl(url);
+      // The ad-hoc http(s) sniff used to live here; postingKey (via startEvaluate)
+      // now owns URL validity, so a malformed url surfaces as an errored job in
+      // the Workers tray instead of being silently ignored pre-dispatch.
+      if (!isStr(url)) return { status: "ignored", note: "invalid url" };
+      // jobForUrl compares against job.input, which is always canonical now that
+      // every launch site routes through startEvaluate — so the lookup key must
+      // be canonical too, or a pipeline URL still carrying tracking noise would
+      // never match the worker already running on it.
+      const ex = ctx.jobForUrl(postingKey(url));
       if (ex && ex.status !== "error" && !raw.rerun) return { status: "ignored", note: "already evaluated" };
-      const id = ctx.startJob({
-        title: isStr(raw.title) ? String(raw.title) : "Evaluate",
+      const id = ctx.startEvaluate({
+        title: isStr(raw.title) ? String(raw.title) : undefined,
         subtitle: isStr(raw.subtitle) ? String(raw.subtitle) : undefined,
-        kind: "evaluate",
-        input: url,
+        url,
         page: "/pipeline",
       });
       return { status: "done", jobIds: id ? [id] : [] };
@@ -175,7 +193,8 @@ const ACTIONS: Record<string, ActionDef> = {
       const pending = matches
         .filter((j) => {
           if (rerun) return true;
-          const ex = ctx.jobForUrl(j.url);
+          // Same canonical-key reasoning as the `evaluate` action above.
+          const ex = ctx.jobForUrl(postingKey(j.url));
           return !ex || ex.status === "error";
         })
         .slice(0, cap);
@@ -191,11 +210,10 @@ const ACTIONS: Record<string, ActionDef> = {
         const batchId = pending.length > 1 ? genBatchId() : undefined;
         const ids = pending
           .map((j) =>
-            ctx.startJob({
+            ctx.startEvaluate({
               title: `Evaluate · ${j.company}`,
               subtitle: j.role,
-              kind: "evaluate",
-              input: j.url,
+              url: j.url,
               page: "/pipeline",
               batchId,
             }),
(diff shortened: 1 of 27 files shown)


santifer:
Thanks for the screenshots, @EliRobinson, and sorry this sat five days without a word: you did the right thing opening the issue first and keeping this a draft asking for shape, and silence was a poor return on that.

Three things from my side:

1. **web/ is owned by the career-ops-ui agent**, so this is `area:web` and goes to their queue with context, not to mine. I route and coordinate; they make the technical call.
2. **Do not rebase yet.** `web/src/lib/job-url.mjs` re-implements what #3180 landed in core today (`liveness-api.mjs` already parses `/jobs/view/{id}`, the slug form and `?currentJobId=`, and builds the guest endpoint). @rubicon offered to extract that into a shared module. Rebasing 24 files before we decide where that parser lives means doing the conflict twice. I will confirm here once the shared module has an owner.
3. Your premise held up while you waited: the guest endpoint went from workaround to repo policy in core. That was the open question in #2995, and it is answered.

I would rather you spent your next hour on the split I described in #2993 than on a rebase that will move under you.

santifer:
Update with the decision you were waiting on, @EliRobinson: the LinkedIn URL parsing will live canonically in the core (`liveness-api.mjs`, extracted into a shared importable module), and the web imports it — same pattern as the entity decoder and the text-key normalizer. One body of the parser, three consumers, zero drift.

For this PR that means: when you rebase, drop `web/src/lib/job-url.mjs` and import the shared module instead of carrying your own copy. Your authwall handling and the pipeline wiring stay yours — it's only the URL-parsing core that moves. The shared module isn't extracted yet; I'll ping this thread the moment it lands so you don't rebase twice. Nothing for you to do until then.

EliRobinson:
@coderabbitai review

santifer:
Answered your three questions in full at #2995 — https://github.com/career-ops-hq/career-ops/issues/2995#issuecomment-5543259745 — since you said the answers may change this code, and they do.

Short version so it's here too: **(1)** mirror-read / canonical-record is right and is already our apply-side invariant, with one condition — the report must declare it read a mirror, reusing the existing `Verification: unconfirmed` vocabulary. **(2)** write to `data/pipeline.md`, but through `scan.mjs`'s own `appendToPipeline`/`formatPipelineOffer` plus `pipeline-lock.mjs`, not a row shape hand-rolled in `web/`. **(3)** the gating is real (`page.tsx:21` → `phase === "in-between"`), so don't unhide it on the home — the Pipeline entry point you proposed is the right door.

And one that shrinks this: the adjacent 285s/kill-budget fix is already in main via #3124. Real bug, someone else's fix landed first — queue latency on our side, not a wasted diagnosis.

I'm not merging this as one 27-file change, which is the answer your own framing asks for rather than a knock on the code. Proposed split is in the issue comment. Marking it parked until you tell me which way you want to take it — and if you'd rather not carry three PRs, say so and I'll take the first one and hand the surface back to you.

Issue #2996, opened 2026-08-17
Pipeline always opens on INBOX, and the tab list is duplicated three times


EliRobinson:
## The problem

The Pipeline page always opens on INBOX. For a user whose daily habit is checking what is in flight, that is one extra click every single visit; someone living in APPLIED or INTERVIEW never gets the view they actually want by default.

While looking at this, a smaller structural issue turned up: the tab list is written out three times. The tab strip in `pipeline-view.tsx` has one copy, `actions/registry.ts` has another as `TAB_VALUES` for the assistant's `filterPipeline` action, and `test-all.mjs` §55.3b reads both as literal arrays. Adding a tab means editing several places, and nothing fails if you miss one; the assistant would simply reject a tab the UI happily renders.

## Why this is on the core path

Track is the last step of discover → evaluate → tailor → apply → track, and the Pipeline page is where tracking is done. This is a small preference, so I want to be explicit about the maintenance question in CONTRIBUTING's parallel-feature test: it adds one optional key to `config/profile.yml` and no new data file, and the centralization it requires removes duplication rather than adding surface.

## Proposed shape

1. `web/src/lib/pipeline-tabs.mjs`: the canonical tab list plus `normalizePipelineTab()` for untrusted values (a URL query, a `profile.yml` key, an assistant intent are all untrusted in the same way).
2. `pipeline-view.tsx` and `actions/registry.ts` alias that list instead of keeping their own copies, so §55.3b reads the shared module rather than two literal arrays.
3. A `web:` section in `config/profile.yml` holding `default_pipeline_tab`, set from Config in the web app, defaulting to INBOX when absent.

An explicit `?tab=` in the URL should still win, so a shared or bookmarked link keeps meaning what it says.

## Questions for the maintainers

1. Is `config/profile.yml` the right home for a web view preference, or would you rather web-only prefs lived in their own file? This is the question I am least sure about: `profile.yml` is user-layer and mostly about the person, not the UI, and a `web:` block sits slightly oddly next to comp targets and archetypes.
2. If a separate file is preferred, should it be gitignored (machine-local) or tracked?
3. Is a per-tab default enough, or would you want the whole filter state (tab plus sort plus min score) remembered?

## Example implementation

I have a working implementation and have opened it as a draft PR so there is something concrete to react to. It is an example of one way to resolve this, not a claim that the structure is right; the storage-location question above in particular would change its shape, and I am happy to redo it.

EliRobinson:
[...]

PR #3000, opened 2026-08-17
feat(web): choose which tab the Pipeline page opens on

EliRobinson:
## Related issue

Addresses #2996.

## ⚠️ This is an example, not a proposal that the structure is right

**I am opening this as one worked example of how #2996 could be resolved, so the discussion has something concrete to react to. It is explicitly open to a better structure.** If you would rather this were shaped differently, split up, or replaced with another approach, say so and I will redo it. Question 1 in #2996 in particular (where a web view preference should live) would change this code's shape, and I am happy to redo it either way.

## What does this PR do?

Two things, one of which is worth keeping regardless of the other.

**The centralization.** `web/src/lib/pipeline-tabs.mjs` becomes the canonical tab list plus `normalizePipelineTab()` for untrusted values. The tab strip in `pipeline-view.tsx` and `TAB_VALUES` in `actions/registry.ts` now alias it instead of keeping their own copies, so `test-all.mjs` §55.3b reads the shared module rather than two literal arrays. **If the preference below is rejected, this half still stands on its own** and I would happily submit it alone.

**The preference.** A `web:` section in `config/profile.yml` holding `default_pipeline_tab`, set from Config in the web app, defaulting to INBOX when the key is absent. An explicit `?tab=` in the URL still wins, so a shared or bookmarked link keeps meaning what it says.

## The question I am least sure about

`config/profile.yml` is user-layer and mostly describes the person, not the UI. A `web:` block sits slightly oddly next to comp targets and archetypes. A separate web-prefs file (tracked or gitignored) may be the better answer, and that is question 1 in the issue. The read/write path is one small module, so moving it is cheap; I did not want to invent a new config file unilaterally.

`normalizePipelineTab()` treats the URL query, the `profile.yml` value and an assistant intent as untrusted in the same way, which is the property I would want preserved wherever the value ends up living.

## Tests

- `node test-all.mjs` → 4081 passed, 0 failed (2 pre-existing warnings, unrelated: missing local user data, and a `santifer.io` string in `dashboard/`)
- `node --test "tests/**/*.test.mjs"` in `web/` → 259 passed, 0 failed
- `tsc --noEmit` clean

Rebased onto current `main`; `web-prefs.ts` uses the `import * as yaml` namespace form per main's js-yaml guard.

## Type of change

- [ ] Bug fix
- [x] New feature

## Checklist

- [x] I opened an issue first (#2996)

### DIFF
--- a/config/profile.example.yml
+++ b/config/profile.example.yml
@@ -290,3 +290,15 @@ cover_letter:
 #     cross_role_bucket: "all_EM_roles"
 #     applied_to: ["Senior Software Engineer"]
 #     last_apply_date: "2026-01-01"
+
+# ---------------------------------------------------------------------------
+# WEB APP VIEW PREFERENCES (optional, set from Config in the web app)
+# ---------------------------------------------------------------------------
+#
+#   default_pipeline_tab: which tab the Pipeline page opens on when the URL
+#                         names none. One of INBOX, ALL, EVALUATED, APPLIED,
+#                         RESPONDED, INTERVIEW, OFFER, HIRED, REJECTED,
+#                         DISCARDED, SKIP. Default (key absent): INBOX.
+#
+# web:
+#   default_pipeline_tab: INBOX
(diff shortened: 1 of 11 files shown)


santifer:
Routing note, @EliRobinson: this one is pure web surface (labeled area:web), so its substantive review belongs to the web maintainer, alongside your #2996 issue which is already in their queue.

PR #3001, opened 2026-08-17
fix(reply-matcher): the '?' unknown-employer placeholder matched almost every reply, at high confidence

Zoubeir23:
No issue — found while auditing `reply-matcher.mjs`, which ships with **no test file and no self-test**, the least-covered substantive module left. `CONTRIBUTING.md:22` welcomes a direct PR for a bug fix.

## Why the company check is load-bearing

`isCompanyMatch` is worth 2 points, it is the corroboration that lets a **partial** role match count at all (#2671), and next to a post-application keyword it raises the verdict to `high`. `reply-watch` then proposes tracker updates from that verdict. A false company signal is therefore a wrong status on a real application.

## Defect 1 — `?` matched almost every reply

`?` is the **documented** placeholder for an unknown end employer (#1596, `AGENTS.md`). `checkCompanyMatch` opened with a bare `text.includes(company)`, so any text containing a question mark matched it. Replies ask questions.

Measured end to end on `main` — two applications, a mail belonging to neither:

```js
apps = [ {num: 1, company: '?',    role: 'Developpeur Fullstack'},
         {num: 2, company: 'Acme', role: 'Backend Engineer'} ]

from: rh@societe-sans-rapport.com
body: "Seriez-vous disponible pour un interview la semaine prochaine ?"
```

```json
{ "application_num": 1, "company_hint": "?", "confidence": "high",
  "signals": ["company-name", "post-application-keyword"] }
```

**High confidence, on the strength of a question mark.** The same applies to the tracker's other no-data sentinels (`—`, `-`) when one reaches the Company column.

## Defect 2 — a short name matched inside a longer word

The length floor (`cNorm.length > 2`) guarded only the third, normalized check. The two substring checks above it had no floor and no word boundary:

| Company | Text | `main` |
|---|---|---|
| `HP` | "un profil **PHP** senior" | matched |
| `#` | "Poste en C**#** et .NET" | matched |

## The fix

- **`isPlaceholderCompany`** — a value carrying no letter and no digit is a placeholder, not a name, and never matches.
- **Short names are decided by a word boundary.** Three alphanumeric characters or fewer go through `matchesOnWordBoundary` and return there, rather than falling through to the substring checks that would reinstate the match they just refused. The boundary is `\p{L}\p{N}`-aware, so it does not split on accents or CJK.

The names that are genuinely short had to keep working, and the tests pin them: **HP**, **3M** and **IBM** all still match when they appear as words.

## Test plan

New `tests/reply-matcher-company-boundary.test.mjs` (14 cases), picked up by `test-all.mjs` auto-discovery — the module's first tests.

- [x] 6 of the 14 verified red before the fix.
- [x] The 8 guards pass before **and** after, which is what makes them guards: HP/3M/IBM as words, a normal name matching and correctly failing, case-insensitivity, and an end-to-end reply that genuinely names company and role still landing on it with `high`. Without that last one, the end-to-end case could have passed by matching nothing at all.
- [x] Verified the placeholder guard discriminates: removing that single line turns 3 cases red again.
- [x] `node test-all.mjs` — 4094 passed, 0 failed (the two warnings are the pre-existing `Dashboard build skipped — go compiler not in env` and the `santifer.io` personal-data notice in `dashboard/`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved company matching to ignore punctuation-only or otherwise invalid placeholder values.
  * Prevented short company names from matching inside longer words.
  * Preserved valid matching for multilingual names, including CJK text, and high-confidence company replies.

* **Tests**
  * Added comprehensive coverage for placeholder values, word boundaries, Unicode names, valid short names, and end-to-end matching scenarios.

### DIFF
--- a/reply-matcher.mjs
+++ b/reply-matcher.mjs
@@ -21,8 +21,38 @@ export function normalizeChinese(s) {
     .trim();
 }
 
+// A company value that carries no letter and no digit is a PLACEHOLDER, not a
+// name: `?` is the documented marker for an unknown end employer (#1596), and a
+// hand-edited row can hold the tracker's other no-data sentinels (`—`, `-`).
+// Substring-matching those turns punctuation into a company signal — and since
+// replies ask questions, `?` matched almost every mail, scoring 2, corroborating
+// partial role matches, and reaching confidence `high` next to any
+// post-application keyword.
+function isPlaceholderCompany(company) {
+  return !/[\p{L}\p{N}]/u.test(company);
+}
+
+// Short names must land on a word boundary. The normalized check further down
+// has always required more than two characters, but the two substring checks
+// above it had no floor at all, so `HP` matched the word `PHP`. A boundary
+// keeps the short names that are real — HP, 3M, IBM — while refusing the ones
+// that merely occur inside a longer word.
+const SHORT_NAME_MAX = 3;
+
+function matchesOnWordBoundary(text, company) {
+  const escaped = company.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+  return new RegExp(`(?<![\\p{L}\\p{N}])${escaped}(?![\\p{L}\\p{N}])`, 'iu').test(text);
+}
+
 export function checkCompanyMatch(text, company) {
   if (!company || !text) return false;
+  if (isPlaceholderCompany(company)) return false;
+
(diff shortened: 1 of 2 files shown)


santifer:
A placeholder that matches everything silently misclassifies replies across every tracked application: the quiet-wrong-answer class this repo hunts. Merged! 🚀

PR #3005, opened 2026-08-17
fix(process-quality): three undisclosed employers were reported as one company with an averaged friction rate

Zoubeir23:
No issue — found by following the `?` placeholder class out of #3001 into the analytics scripts. `CONTRIBUTING.md:22` welcomes a direct PR for a bug fix.

## The defect

`AGENTS.md` documents the convention: an undisclosed end employer is recorded as `?` in Company, with the agency in the tagged `via=` field (#1596). `merge-tracker` already honours it — its dedup carries an explicit cross-channel guard so `?` rows never fuzzy-merge across agencies (`tracker-columns-tests.mjs` case 12).

`aggregateProcessQuality` grouped on the Company cell alone. Every undisclosed employer therefore collapsed into one bucket. Measured on `main`, three unrelated employers reached through three different agencies:

```js
{ company: '?', via: 'Hays',         notes: '[process-friction] rescheduled twice' }
{ company: '?', via: 'Michael Page', notes: 'smooth process' }
{ company: '?', via: 'Robert Half',  notes: '[process-friction] ghosted after onsite' }
```

```json
{ "company": "?", "totalInterviews": 3, "frictionCount": 2, "frictionRate": 0.67 }
```

The script's own header calls this a "per-company recruiting-friction rate". That row is not a company, it is three — and `0.67` is arithmetic across employers the user never dealt with. The number they would act on ("this one wastes my time") is exactly the one that cannot be computed from these rows.

It also hides the real signal in the other direction: Michael Page had a clean process, and averaging buries that.

## The fix

- **`isPlaceholder`** — a cell with no letter and no digit is a marker, not a value. Covers `?` and the tracker's `—`/`-` no-data sentinels.
- **The channel becomes the bucket** for those rows: `? (via Hays)`. That is what actually distinguishes them, and it is the same axis merge-tracker already keys on.
- **A row naming neither employer nor channel is dropped.** It supports no per-company claim, so inventing a bucket for it would be the original bug in smaller form.

## Test plan

New `tests/process-quality-unknown-employer.test.mjs` (7 cases), the module's first dedicated tests.

- [x] 4 of the 7 verified red before the fix.
- [x] The 3 guards pass before **and** after: a named company unaffected, case-insensitive grouping of a named company still merging, and a pre-#1596 tracker with no Via column at all still aggregating normally.
- [x] One case pins the *direction* of the attribution, not just the split: Hays keeps `frictionRate: 1` and Michael Page `0`, so the friction lands on the channel that had it rather than being averaged.
- [x] Verified the guard discriminates: neutralising the `isPlaceholder` branch turns 4 cases red again.
- [x] `node process-quality.mjs --summary` still runs clean.
- [x] `node test-all.mjs` — 4087 passed, 0 failed (the two warnings are the pre-existing `Dashboard build skipped — go compiler not in env` and the `santifer.io` personal-data notice in `dashboard/`).

## Related

Same class as #3001, where `?` was matching almost every email reply because `checkCompanyMatch` treated it as a name. The convention is documented and one consumer honours it; the others were never audited against it. `analyze-patterns.mjs` and `company-history.mjs` group by company too — I have not measured them here and am not claiming they are affected, only that they are the next places I would look.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of undisclosed or placeholder employer information.
  * Separates unknown employers by available referral or agency channel.
  * Prevents records without usable employer or channel details from being incorrectly combined.
  * Preserves existing aggregation for named companies and older records without channel information.

* **Tests**
  * Added coverage for unknown-employer aggregation and channel attribution.

### DIFF
--- a/process-quality.mjs
+++ b/process-quality.mjs
@@ -149,6 +149,12 @@ export function extractFriction(row) {
   return { hasFriction: true, reason: (match[1] || '').trim() };
 }
 
+// A cell carrying no letter and no digit is a PLACEHOLDER, not a value: `?` for
+// an undisclosed employer (#1596), and the tracker's `—`/`-` no-data sentinels.
+function isPlaceholder(value) {
+  return !/[\p{L}\p{N}]/u.test(value);
+}
+
 // --- Core aggregation ---
 //
 // Groups rows by company (case-insensitive, trimmed), counts total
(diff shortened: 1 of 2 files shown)


santifer:
A report that says disclosed when it wasn't is worse than no report: the reader has no reason to doubt it. Clean fix with the test to pin it. Merged! 🚀

PR #3006, opened 2026-08-17
fix(tracker): treat Windows rename contention as contention, not death (#2777)

yaserselvam:
## The bug

`writeFileAtomic` in `tracker-utils.mjs` closes with a bare `renameSync`. On Windows, a rename whose **destination** is open by anyone else at that instant fails with `EPERM` (errno `-4048`). POSIX `rename(2)` replaces the destination atomically and cannot fail that way, which is why CI stays green and this only reproduces on Windows.

The handle holder is usually **not** another writer of ours, since the tracker lock serializes those. It is a transient reader: an antivirus scanner, the Windows Search indexer, or a concurrent `readFileSync` from a reporting script. The handle is released in milliseconds.

**This is a production bug, not only a test flake.** Any tracker write on Windows can fail with a raw stack trace and lose the write, and `data/applications.md` is the canonical store by the project's own doctrine ("Files are canonical, databases are derived").

## Why this is the same bug as #2777

`tracker-utils.mjs` already documents this exact portability gap, in its own header:

> pipeline-lock learned that Windows answers mkdir/rm with EPERM/EACCES/EBUSY under contention while this file still treated anything but EEXIST as fatal, killing a writer and losing its item.

`mkdir` was fixed, then `rm`/`stat` in `2743281`. `rename` is the third syscall in that family and the one still bare. I have followed the shape already established rather than inventing a new one.

## The fix

- **`isRenameContention(err)`** — one exported classifier, mirroring the exported `isMkdirContention` / `isRmContention`, so there is no second copy to drift.
- **`renameSyncWithRetry(tmp, path, rename = renameSync)`** — bounded backoff (`[1,2,5,10,25,50,100]`ms, ~193ms worst case, then rethrow the **original** error) on `EPERM`/`EACCES`/`EBUSY` only.
- Any other errno still fails on the first attempt. `writeFileAtomic` still cleans up its temp file on failure, and a failed write still leaves the original intact.
- The rename is injectable purely so the retry loop is testable off-Windows.

## Reproduction

Windows 11, Node v24.18.0:

| | Result |
|---|---|
| **Before** | `node test-all.mjs` failed **1-2 tests per run**, non-deterministically, in `tracker-writer-lock-tests.mjs` and `set-status-tests.mjs` |
| **After** | **4091 passed, 0 failed** |

Both suites pass in isolation (27/0 and 88/0, three runs each); only the full run manufactures enough concurrent readers to lose the race. Three consecutive clean full runs before the new test was added (4070/0), and green with it.

The actual failure:

```
Fatal: Error: EPERM: operation not permitted, rename
  'C:\...\Temp\career-ops-writer-lock-kJFY1z\.applications.md.15148.1786995213566.1a852d47-....tmp'
  -> 'C:\...\Temp\career-ops-writer-lock-kJFY1z\applications.md'
    at renameSync (node:fs:1013:11)
    at writeFileAtomic (tracker-utils.mjs:557:5)
    at Object.replace (tracker-utils.mjs:524:7)
    at updateTrackerStatuses (reply-watch.mjs:205:46)
  errno: -4048, code: 'EPERM', syscall: 'rename'
```

## Tests

`tests/rename-contention.test.mjs`, 21 assertions, glob-discovered so no registration needed. It pins:

1. The classifier agrees on the measured Windows codes **and only those** — `ENOENT`/`EXDEV`/`EISDIR`/`ENOSPC` still fail fast rather than stalling ~200ms behind a pointless retry loop.
2. A contended rename is retried until it lands; a permanently contended one gives up at the bound and rethrows the original `EPERM`.
3. `writeFileAtomic`'s contract holds: destination replaced, no `.tmp` left behind on success **or** failure, original intact after a failed write.
4. **No direct `renameSync()` call remains in the atomic path.** This is the guard that stops the bug being reintroduced one refactor from now, which is how the mkdir/rm half survived two earlier fixes. Same shape as the existing `no bare rmSync(lockDir) remains` guards in `tests/lock-rm-contention.test.mjs`.

## Deliberately left out

Six other bare `renameSync` atomic-replace sites carry the same latent bug. I have left them out rather than widening this PR:

- `discover-ats.mjs:852`
- `followup-seed.mjs:422`
- `paste-reply.mjs:134`
- `scan-ats-full.mjs:121`
- `merge-tracker.mjs:1228`
- `update-system.mjs:1296`

The last is the most consequential, since a failed rename there breaks an upgrade partway through. Happy to take them in a follow-up PR if you want them, either one-by-one or by exporting the helper for reuse.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of atomic tracker-file updates on Windows when files are temporarily locked or in use.
  * Automatically retries eligible rename conflicts before reporting an error.
  * Preserves original errors for persistent failures and immediately reports unrelated errors.
  * Ensures temporary files are cleaned up when an update cannot be completed.
  * Supports replacing existing destination files safely.

### DIFF
--- a/tests/rename-contention.test.mjs
+++ b/tests/rename-contention.test.mjs
@@ -1,0 +1,170 @@
+// tests/rename-contention.test.mjs
+//
+// #2777's missing third syscall. The mkdir and rm sides of both locks now treat
+// Windows' EPERM/EACCES/EBUSY answers as contention, but `writeFileAtomic`'s
+// closing `renameSync` was still bare — and on Windows a rename whose
+// DESTINATION is open by anyone else at that instant fails with EPERM
+// (errno -4048), killing the writer and losing the tracker write.
+//
+// POSIX `rename(2)` replaces the destination atomically and cannot fail that
+// way, which is why this never reproduces on Linux/macOS CI. Measured on
+// Windows 11 / Node v24.18.0: `node test-all.mjs` failed 1-2 tests per run,
+// non-deterministically, in tracker-writer-lock-tests.mjs and
+// set-status-tests.mjs. Both suites pass in isolation; only the full run
+// manufactures enough concurrent readers to lose the race.
+//
+// The holder is usually not another writer of ours (the tracker lock serializes
+// those) but a transient reader: an antivirus scanner, the Search indexer, or a
+// concurrent readFileSync from a reporting script.
+//
+// These tests pin four things:
+//   1. the classifier agrees on the measured Windows codes and ONLY those,
+//   2. a contended rename is retried and the write completes,
+//   3. a non-contention errno still fails fast, with the temp file cleaned up,
+//   4. no bare renameSync remains in the tracker's atomic-write path.
+
+import { readFileSync, existsSync, mkdtempSync, writeFileSync, readdirSync, rmSync } from 'fs';
+import { join } from 'path';
+import { tmpdir } from 'os';
+import { pass, fail, ROOT } from './helpers.mjs';
+import {
+  isRenameContention,
+  renameSyncWithRetry,
+  writeFileAtomic,
+  RENAME_RETRY_DELAYS_MS,
+} from '../tracker-utils.mjs';
+
+console.log('\n📝 atomic tracker writes: rename contention is contention, not death (#2777)');
+
+const ok = (cond, msg) => (cond ? pass(msg) : fail(msg));
+const mkErr = (code) => Object.assign(new Error(code), { code });
+
+// ── 1. Classifier table ──────────────────────────────────────────────────────
+// The codes come from measured Windows failures, not speculation. EPERM is the
+// one observed here; EACCES and EBUSY are the same family already recorded for
+// mkdir/rm in pipeline-lock.mjs.
+{
+  for (const code of ['EPERM', 'EACCES', 'EBUSY']) {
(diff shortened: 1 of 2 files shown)


Scott-Emberson (approved):
Code owner for `tests/`. `rename-contention.test.mjs` injects contention through the `rename` seam, so the Windows-only retry loop runs deterministically on all three runners rather than relying on a real race. Both directions are asserted with controls: `attempts === 4` after three EPERM throws proves the retry ran, `EXDEV` fails on the first attempt proves genuine fatal errors still fail fast, and the classifier table admits only the contention codes. The `directCalls === 0` static guard is non-vacuous. One process note, not a code point: the suite has not run in CI yet (first-time contributor gate), so the green is not CI-verified until the workflow is approved. Approving our area.

santifer:
Welcome to career-ops, @yaserselvam! This is a first contribution at the level we hope veterans reach: you found #2777's missing third syscall, measured it on real Windows instead of speculating (1-2 non-deterministic failures per full run, passing in isolation — the signature of a manufactured race), explained WHY it never reproduces on POSIX, and shipped the classifier exported the same way pipeline-lock exports its own, so there is one definition and nothing to drift. Atomics.wait for the sync backoff with the contract reason written down is exactly right. Merged — your first contribution. 🚀 A repo census shows 8 more files calling renameSync bare; filing the family audit as a follow-up issue with the list, and it has your name on it first if you want it.

PR #3007, opened 2026-08-17
fix(providers): senjob was the fifth private entity decoder, and the weakest — it emitted C0 controls

Zoubeir23:
Self-inflicted, and in the exact class this repo has spent four rounds consolidating. `providers/_html-entities.mjs` says it plainly:

> Each round has migrated whichever copies were in front of the contributor at the time, which is why there were four.

**#2962 added a fifth**, ten hours before #2902 landed the guard meant to stop them. `senjob.mjs` shipped with its own `NAMED_ENTITIES` table, its own `ENTITY_RE`, and its own `isValidScalar`.

## Why it matters: mine was weaker

My guard rejected surrogates and noncharacters. It did **not** reject the C0 controls:

```
visibleText('titre&#1;suite')  →  "titresuite"      ← before
                               →  "titre&#1;suite"        ← after (shared)
```

`_html-entities.mjs` explains exactly why that is not cosmetic: a decoded title is written to the scan history, the pipeline, the tracker, and every document generated downstream. A control character there produces ill-formed data that is tedious to trace back to one malformed entity weeks later. The shared decoder rejects it; my copy emitted it.

## Why the migration needed the shared table to grow

Swapping the decoder in wholesale would have **regressed** senjob: the shared `NAMED_ENTITIES` held only the XML five plus `nbsp`, so `Charg&eacute; de projet` came back literal. That is precisely the pressure that grows private tables in the first place, so the letters belong in the shared module rather than in the next European scraper:

- the Latin-1 letter entities (French, Portuguese, Spanish, German, Nordic), lower and upper case;
- the punctuation those same pages emit around titles (`&deg;`, `&hellip;`, `&laquo;`, `&ndash;`, `&euro;`, curly quotes).

Unknown names still pass through untouched, so the list is a floor, not a closed set.

## A bug found while adding them

The lookup was `NAMED_ENTITIES[body.toLowerCase()]`. That is right for the XML five — legacy pages really do write `&AMP;` — and **wrong for letters**, which are case-sensitive: `&Eacute;` is É, not é. Adding uppercase keys under a lowercasing lookup would have made every one of them unreachable *and* decoded `&Eacute;` to the lowercase letter.

Case-sensitive exact match first, with the case-insensitive fallback kept for the six names where it is correct. No existing behaviour changes: those six were the entire table before.

## One policy difference, stated rather than smoothed over

senjob's private guard rejected the U+FDD0–FDEF noncharacter block. The shared decoder admits it, because XML 1.0 §2.2 Char runs to U+FFFD and that block sits inside it. I did **not** change the shared policy: the module names a standard and implements it faithfully, and picking a different one is a design call, not a bug fix. The senjob test now pins the standard's real edge instead of a stricter rule one provider applied alone.

Worth a second opinion though: the range excludes U+FFFE and U+FFFF only as an accident of ending at FFFD, while admitting the other 32 noncharacters. If the intent is "safe to put in a job title" rather than "legal XML", those 32 are on the wrong side of the line.

## Test plan

- [x] `&#1;` verified emitted before the migration and literal after — the defect that motivated it.
- [x] Live re-parse of five real senjob listing pages: **208 postings**, identical to before.
- [x] `node test-all.mjs --only _html-entities` — 25 passed, the shared module's existing suite unaffected by the table growth.
- [x] `node test-all.mjs --only providers/senjob` — 27 passed. Two assertions were updated, both because the shared behaviour is now correct where mine was narrower, and both explain why in place.
- [x] `node test-all.mjs` — 4080 passed, 0 failed.

## Follow-up, deliberately not in this PR

The guard from #2902 could not have caught this, for two independent reasons: it skips any provider that does **not** import the shared module, and its name pattern looks for `decodeEntities`/`decodeXmlEntities`/`fromCodePoint` — senjob declared `decodeEntity`, singular. Widening it flags exactly one other file, `jobvite.mjs`, whose private copy the module's own header already describes as upstreamed. That is a separate change and I would rather send it separately than bundle a guard rewrite with a migration.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved decoding of HTML and XML entities, including Latin-1 characters and common punctuation.
  * Preserved correct case handling for entity names and left unsupported entities unchanged.
  * Improved text cleanup for invalid XML characters and special Unicode ranges.
  * Standardized entity decoding for provider content.

### DIFF
--- a/providers/_html-entities.mjs
+++ b/providers/_html-entities.mjs
@@ -37,7 +37,40 @@
 // so a decimal entity can never absorb trailing hex letters — "&#1a2;" no
 // longer silently parses as codepoint 1 and drops "a2"; it just fails to
 // match and passes through untouched, same as any other malformed entity.
-const NAMED_ENTITIES = { amp: '&', lt: '<', gt: '>', quot: '"', apos: "'", nbsp: ' ' };
+// The XML five plus nbsp, then the Latin-1 letter entities. The letters are not
+// decoration: a European board writes `D&eacute;veloppeur` and `Fran&ccedil;ais`
+// in its HTML, and leaving those literal puts `D&eacute;veloppeur` in a job
+// title, the tracker, and every document generated from it. Providers that
+// needed them grew private tables instead, which is the drift this module
+// exists to end — so they belong here rather than in the next scraper.
+//
+// Unknown names still pass through untouched (see decodeEntities), so this list
+// is a floor, not a closed set.
+const NAMED_ENTITIES = {
+  amp: '&', lt: '<', gt: '>', quot: '"', apos: "'", nbsp: ' ',
+  // French / Portuguese / Spanish / German / Nordic letters, lower and upper.
+  agrave: 'à', aacute: 'á', acirc: 'â', atilde: 'ã', auml: 'ä', aring: 'å', aelig: 'æ',
+  ccedil: 'ç',
+  egrave: 'è', eacute: 'é', ecirc: 'ê', euml: 'ë',
+  igrave: 'ì', iacute: 'í', icirc: 'î', iuml: 'ï',
+  ntilde: 'ñ',
+  ograve: 'ò', oacute: 'ó', ocirc: 'ô', otilde: 'õ', ouml: 'ö', oslash: 'ø',
+  ugrave: 'ù', uacute: 'ú', ucirc: 'û', uuml: 'ü',
+  yacute: 'ý', yuml: 'ÿ', szlig: 'ß',
+  Agrave: 'À', Aacute: 'Á', Acirc: 'Â', Atilde: 'Ã', Auml: 'Ä', Aring: 'Å', AElig: 'Æ',
+  Ccedil: 'Ç',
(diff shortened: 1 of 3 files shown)


Scott-Emberson (approved):
Code owner for `tests/`. The senjob test change pins the fix without a tautology. `visibleText('&#1;')` asserts the C0 control stays literal, and under the deleted private decoder `isValidScalar` admitted code 1, so that assertion would fail on the old code: it genuinely proves the shared decoder's rejection. The two retargeted assertions are honest adjustments to the shared policy (`&#xFDD0;` now decodes, `euro` is now a known entity), not weakenings. Hermetic, real parser, minimal. Approving.

santifer:
Merged 🚀 The case-sensitivity catch is the quality marker here, @Zoubeir23: a lowercased lookup would have made every uppercase entry unreachable and silently decoded &Eacute; to é — and you kept the case-insensitive floor exactly where legacy pages actually write &AMP;. Growing the shared table instead of the sixth private one is what #2150 started and this finishes; the header comment saying WHY the letters belong here will stop the next scraper from drifting again.

PR #3008, opened 2026-08-17
test(providers): the private-decoder guard reported clean over the copy it was written to catch

Zoubeir23:
> **Stacked on #3007.** Its commit is the first of the two here, because the widened guard fails on `senjob.mjs` until that migration lands. Merge #3007 first and this reduces to the second commit.

The guard from #2902 was written so "the next re-introduction fails on the commit that adds it instead of drifting quietly for a year". Ten hours earlier, #2962 had added one — and the guard reported a clean repository.

It could not have seen it, for two independent reasons.

## Blind spot 1 — non-importers are skipped

```js
if (!SHARED_IMPORT.test(src)) continue;
```

A provider that never imports the shared module and simply writes its own decoder is **exactly** the re-introduction being guarded against, and it was the one shape excluded from the scan. The check only ever asked "does an importer also keep a private copy?", which its own pass message states accurately — but that is narrower than the goal.

A provider that declares a decoder is now an offender whether or not it also imports the shared one, and the message distinguishes the two cases.

## Blind spot 2 — the name list missed the near misses

```js
(decodeEntities|decodeXmlEntities|fromCodePoint)
```

senjob declared `decodeEntity` — **singular** — next to its own `NAMED_ENTITIES` table. Neither name is in the list. A list of spellings has to carry the near misses, so it now also covers `decodeEntity`, `decodeXmlEntity`, `NAMED_ENTITIES`, `XML_ENTITIES` and `HTML_ENTITIES`.

## What widening it exposed: jobvite

Running the widened guard across all 74 providers flagged exactly two files: `senjob.mjs` (handled by #3007) and **`jobvite.mjs`**.

jobvite's copy is not a divergence — it is the *source* of the shared module's current strictness. `_html-entities.mjs` says so in its own header:

> the jobvite provider (#2623) grew its own copy that was STRICTER than this one … That strictness is now here, in isEmittableCodePoint below, which is the point of a shared decoder — the improvement should not have had to live in one provider.

The improvement was upstreamed; the provider was never migrated. This finishes it.

Verified equivalent on the references that feed really carries, which its own comment names as the reason the decoder exists there at all:

| input | shared decoder |
|---|---|
| `Manager &#8217;s role` | `Manager ’s role` |
| `Lead &#x2013; Platform` | `Lead – Platform` |
| `R&amp;D` | `R&D` |
| `&#0;` | `&#0;` (left literal) |

One behaviour change, in the safe direction: jobvite's private table had no `nbsp`, so `&nbsp;` used to survive literally into a title and now decodes to a space.

## Test plan

- [x] `node test-all.mjs --only jobvite` — 56 passed, 0 failed.
- [x] `node test-all.mjs --only rss-entity-decoding` — 17 passed, and **zero offenders across all 74 providers** once both migrations are in.
- [x] Both blind spots verified to discriminate, by planting each shape and watching the guard fail:
  - a private `decodeEntity` with **no** shared import → `senjob.mjs (declares decodeEntity, and does not import the shared decoder)`
  - a re-introduced `NAMED_ENTITIES` table → `senjob.mjs (declares NAMED_ENTITIES)`
- [x] `node test-all.mjs` — 4080 passed, 0 failed.

A guard that has never been shown to fail is a guard nobody has tested, which is how this one reported clean over a copy it was written to catch.

## Summary by CodeRabbit

- **New Features**
  - Expanded support for named HTML entities, including Latin-1 characters, punctuation, and currency symbols.
  - Improved case-sensitive handling for letter entities while preserving compatibility for standard XML entities and non-breaking spaces.
  - Correctly decodes additional supported entities, including `&euro;`.

- **Bug Fixes**
  - Standardized entity decoding across job listing and RSS content.
  - Improved handling and validation of numeric character references and unsupported characters.

### DIFF
--- a/providers/jobvite.mjs
+++ b/providers/jobvite.mjs
@@ -2,6 +2,12 @@
 /** @typedef {import('./_types.js').Provider} Provider */
 
 import { fetchTextWithRetry } from './_http.mjs';
+// Shared decoder, not a private copy. This provider's stricter XML 1.0 §2.2 Char
+// guard was upstreamed into _html-entities.mjs (#2623); keeping the local one
+// only risked the two drifting apart again — which is the drift that module
+// exists to end. The numeric references this feed really carries (`&#8217;`,
+// `&#x2013;`) decode identically there.
+import { decodeEntities } from './_html-entities.mjs';
 
 // Jobvite provider — per-tenant public jobs feed.
 // Used by ~3,000 companies across a wide range of industries.
@@ -353,45 +359,6 @@ function tagText(block, name) {
     value = value.slice('<![CDATA['.length, -']]>'.length).trim();
   }
   return value;
-}
-
-const XML_ENTITIES = { amp: '&', lt: '<', gt: '>', quot: '"', apos: "'" };
-
-/**
- * Decode the XML entities that appear in this feed.
- *
- * Numeric references matter here and are not decorative: Jobvite titles carry
- * typographic punctuation from the source ATS — `&#8217;` (right single quote)
- * and `&#x2013;` (en dash) both appear in real postings. Leaving them raw puts
- * literal `&#8217;` into a job title, which then reaches the tracker and any
- * generated document.
- *
- * @param {string} s
- */
-function decodeEntities(s) {
-  return String(s).replace(/&(#x[0-9a-fA-F]+|#\d+|amp|lt|gt|quot|apos);/g, (match, body) => {
-    if (body[0] === '#') {
-      const code = body[1] === 'x' || body[1] === 'X'
-        ? Number.parseInt(body.slice(2), 16)
-        : Number.parseInt(body.slice(1), 10);
-      // XML 1.0 §2.2 Char — only these ranges are legal, and a bare
-      // `code <= 0x10ffff` bound is not the same test. It admits NUL, the C0
-      // controls and lone surrogates, all of which String.fromCodePoint will
-      // happily emit; an unpaired surrogate then travels into a job title, the
-      // tracker, and every document generated from it. Anything illegal is left
-      // as written rather than decoded — the raw reference is visible and inert,
-      // which is the better failure.
-      const legalXmlChar = code === 0x9 || code === 0xa || code === 0xd
-        || (code >= 0x20 && code <= 0xd7ff)
-        || (code >= 0xe000 && code <= 0xfffd)
-        || (code >= 0x10000 && code <= 0x10ffff);
-      // NaN and Infinity fail every comparison above, so they need no separate
-      // guard, and fromCodePoint cannot throw on what survives.
-      if (!legalXmlChar) return match;
-      return String.fromCodePoint(code);
-    }
-    return XML_ENTITIES[/** @type {keyof typeof XML_ENTITIES} */ (body)] ?? match;
-  });
 }
 
 /**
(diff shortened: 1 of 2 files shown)


Scott-Emberson (changes requested):
Code owner for `tests/`. The widening does the real work: the dynamic sweep no longer skips non-importers, so a provider that drops the shared import and keeps a private decoder (the senjob escape) is now caught, and `_html-entities.mjs` is correctly excluded so the shared table does not self-trip. CI is green. One thing holds it up, and it is the same thing the #2902 review asked for.

The guard still has no positive control: the block runs `PRIVATE_DECL` against the real files and asserts `offenders.length === 0`, but nothing plants a known-offending source and asserts the guard fires. If the regex or the two-branch sweep later regresses to match nothing, every run stays green over a real private copy, which is exactly the failure this PR exists to prevent. The PR body says the discrimination was verified manually by planting each shape and watching it fail: encode that as one in-file assertion (`PRIVATE_DECL.test('const decodeEntity = (m, r) => m;')` is true, and the no-import branch yields the "does not import the shared decoder" message) and this is good to go. The senjob test changes themselves are fine.

Zoubeir23:
Right on both counts, and the second one is the criticism I should have made of my own work.

**The positive control is in `830ccd2`.**

You put your finger on exactly what was wrong: I *did* plant each shape and watch the guard fail — and then wrote that in the PR description, which is to say nowhere at all once this merges. A guard whose discrimination lives in prose is a guard nobody can re-verify.

The classification now lives in one `classifyProvider(file, src)` function, so the control **runs it** on synthetic sources rather than restating its logic — otherwise the control would only prove that two copies of the same regex agree.

It plants the four shapes that have actually escaped:

- `const decodeEntity = (m, r) => m;` with no shared import — the senjob escape, singular name
- the same name as a function declaration
- a re-introduced `NAMED_ENTITIES` table rather than a function
- a private copy *alongside* the shared import — the original #2902 shape

and asserts the exact offender line each produces, including the `does not import the shared decoder` wording you named.

**I added a negative control you did not ask for**, because widening a regex has a failure mode in the other direction: `const { decodeEntities } = …` plus a call site is correct usage and must stay clean. Without it, a pattern that matched *everything* would pass the positive control and turn the guard into noise.

Both directions verified by regression rather than assertion alone: shrinking `PRIVATE_DECL` back to its old three names fails the control, and so does restoring the `if (!SHARED_IMPORT.test(src)) continue;` early return.

**On #3007** — thanks for checking the `&#1;` assertion against the deleted decoder rather than taking the claim. CodeRabbit then caught something we both passed on: the comment there *stated* the XML 1.0 §2.2 Char boundary but the assertions only tested one side of it, so a decoder that rejected the whole U+FDD0–FDEF block or emitted U+FFFF still went green. Both edges are pinned now (`56277f2`), and I confirmed each fails when the corresponding range in `isEmittableCodePoint` is broken.

Worth flagging while it is fresh: my first two sabotage attempts changed nothing, because the patterns I patched did not match the real source. The tests looked discriminating and were not. That is the same failure as a green guard over a private copy, one level up.

santifer:
Flattened the stack maintainer-side: #3007 merged (squash), so your branch's base commits were patch-identical to main under different SHAs and a plain rebase conflicted. Rebuilt as a single commit of this PR's own delta over today's main — your guard changes and the jobvite migration, byte-for-byte from your branch tip. Full suite green locally (4269/0). The positive-control design deserves the merge on its own: planting every shape that actually escaped (the singular `decodeEntity` above all) is what turns "the guard is wider now" into something a future refactor can't silently undo. CI runs on the real base; merging on green.

santifer:
Merged 🚀 Three for three today, @Zoubeir23. The Scott-Emberson principle you quoted in the test comment — a guard whose discrimination is only demonstrated in the PR description is demonstrated nowhere after the merge — is now enforced by the positive controls themselves: every shape that actually escaped is planted and must fire, and the legitimate-usage negative keeps the widened pattern honest. Plus jobvite off its local table. The decoder family is now genuinely shut: shared table, migrated callers, and a guard that can't quietly go blind again.

Issue #2573, opened 2026-08-07
Umbrella: restore the anti-fabrication RULES in the 16 localized _shared.md files (#2395)


santifer:
Found and diagnosed by @Scott-Emberson in #2395: the two anti-fabrication rules in `modes/_shared.md` never made it into 16 of the 18 localized `_shared.md` files. The metrics rules crossed over; the authorship and reformulate rules did not. Sixteen independent translations dropped the same pair, which means this was a missing checklist, not sixteen careless people.

**What still protects everyone:** `verify-cv-facts.mjs` is language-agnostic (zero references to modes/locale/language) and its self-test is wired into `test-all.mjs`, so a fabricated employer or title is still blocked at generation time in every language. What is missing is the instruction that stops the model from attempting it in the first place. Last line of defence intact, first line absent.

## The canonical source (English, `modes/_shared.md:32-33`) — translate these two, verbatim in meaning

```
**RULE: NEVER claim the user authored a project, repo, library, tool, framework, or open-source artefact unless explicitly attributed to them in cv.md or article-digest.md.** Tool-of-trade conflation (user uses X → user built X) is the most common fabrication pattern and is forbidden.
**RULE: Keywords get reformulated, never fabricated.** Reorder, reframe, emphasise — but never invent. If a claim isn't backed by an in-scope file, ask the user. If no answer, omit. Silence on a topic beats manufactured detail.
```

Place them where the existing `**RULE:` blocks live in that language's file (right after the metrics rules, before the North Star table), matching the local rule keyword already in use (`REGEL`, `REGLA`, `RÈGLE`, `REGOLA`, `ПРАВИЛО`, …).

**Two reference translations already in the tree, use them as your model:**
- `modes/nl/_shared.md` — six `**REGEL:` blocks, both rules translated idiomatically.
- `modes/ja/_shared.md` — both present, keeping the second rule's English wording where translating it would have weakened it.

## Rules of engagement

- **One language per PR.** A rule mistranslated is worse than a rule missing: a native speaker per language beats one big sweep.
- **Meaning over literalness.** "Tool-of-trade conflation" has no useful literal translation in most languages; say the thing it means (using a tool is not building it).
- **Do not paraphrase the scope.** `cv.md` and `article-digest.md` are file names: they stay in English, exactly as written.
- Comment on this issue with the language you're taking and it gets assigned to you on the spot.

## Languages

- [ ] `ar` (Arabic)
- [ ] `da` (Danish)
[...]

PR #3009, opened 2026-08-17
docs(i18n): restore the two anti-fabrication RULES in the French _shared.md (#2573)

Zoubeir23:
Part of #2573 — French only, per the one-language-per-PR convention (#2426).

## What was missing

`modes/fr/_shared.md` carried two `**REGLE`  blocks, both about metrics. The English source has six, and the two the umbrella targets — authorship and reformulate-never-fabricate — were absent:

```
modes/_shared.md:32   NEVER claim the user authored a project, repo, library…
modes/_shared.md:33   Keywords get reformulated, never fabricated.
```

As the issue puts it: last line of defence intact (`verify-cv-facts.mjs` is language-agnostic and still blocks a fabricated employer at generation time), first line absent. This restores the first line for French.

Placed immediately after the existing metrics rules, which is where the umbrella asks for them and where the English original keeps them.

## Two details that decide whether a translation reads as native

**The file is written in unaccented ASCII.** All of `modes/fr/_shared.md` is — `metriques`, `priorite`, `evaluation`, zero accented characters in the whole file. Adding correctly-accented French would have been *more* correct in isolation and *less* correct here. The addition keeps zero accented characters, verified rather than assumed.

**It says `le candidat`, not `l'utilisateur`.** The English rule says "the user"; this file already uses "le candidat" throughout, so the translation follows the file rather than the source.

Same for `--` as the dash, which is the file's existing convention.

## Scope

French only, deliberately. These are safety rules: a translation I cannot verify to the same standard is worse than an honest gap, because it *looks* like coverage. Fifteen localized `_shared.md` files still need them, and they are visible with:

```bash
grep -c '^\*\*R' modes/*/_shared.md
```

I did not add a source-level guard for the same reason the umbrella exists: any check strong enough to be useful would fail on the fifteen files still waiting, so it belongs with the last translation rather than the first.

## Test plan

- [x] `node test-all.mjs` — 4195 passed, 0 failed.
- [x] Verified the addition introduces no accented character, matching the file's existing convention.
- [x] Verified the rules sit with the other `**REGLE` blocks, not appended elsewhere in the file.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance to avoid attributing authorship without explicit supporting evidence.
  * Clarified that keywords must be based on verified source material, with unsupported claims omitted or confirmed.

### DIFF
--- a/modes/fr/_shared.md
+++ b/modes/fr/_shared.md
@@ -21,6 +21,8 @@
 
 **REGLE : Ne JAMAIS coder en dur des metriques issues des proof points.** Les lire depuis `cv.md` et `article-digest.md` au moment de l'evaluation.
 **REGLE : Pour les metriques d'articles/projets, `article-digest.md` a priorite sur `cv.md`** (`cv.md` peut contenir des chiffres plus anciens).
+**REGLE : Ne JAMAIS affirmer que le candidat est l'auteur d'un projet, d'un depot, d'une bibliotheque, d'un outil, d'un framework ou d'un artefact open source sans attribution explicite dans `cv.md` ou `article-digest.md`.** La confusion outil-de-travail (le candidat utilise X -> le candidat a construit X) est le schema de fabrication le plus frequent, et il est interdit.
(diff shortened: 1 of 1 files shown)


santifer:
Merged 🚀 Checked line by line against modes/_shared.md :32-:33 before merging: both restored RULES mirror the canonical ones faithfully (tool-of-trade conflation and keywords-never-fabricated), in the file's existing style. These two are the core anti-fabrication guardrails of the product, @Zoubeir23 — French users were running without them, and that's exactly the kind of quiet gap the #2573 umbrella exists to close. Ru and Ua landed last week; thanks for keeping the sweep going.

PR #3022, opened 2026-08-18
fix(web): keep child stderr out of the status response on the crash path

### LINKED ISSUE #3020: /api/status returns child stderr to the client on the exit-1 crash path
`web/src/app/api/status/route.ts` guards child stderr on one failure path and not the other.

The spawn-failure branch logs it and returns a fixed message, with the reason stated in its comment:

```ts
// Child stderr is a Node stack trace carrying absolute server paths. It
// belongs in the server log, never in the response body.
console.error(`/api/status: set-status.mjs failed to run: ${stderr.trim()}`);
return NextResponse.json({ error: "status update failed to run" }, { status: 500 });
```

The exit-1 branch then puts that same content in the body (line 194 on `main`):

```ts
const error = typeof parsed?.error === "string" ? parsed.error : stderr.trim() || "status update failed";
```

The fallback fires only when the CLI printed no JSON document, and the allow-list comment a few lines above already says what that case usually is: a crash that printed no JSON at all. So the disclosure is open on the path *most* likely to produce a stack trace, while the path less likely to produce one is guarded. Guarding one of a pair and not the other is the whole defect here.

What reaches the caller is the absolute path of the checkout and the file names inside it, e.g. `file:///Users/<name>/.../career-ops/set-status.mjs:41`.

### Suggested fix

Pass through the CLI's own `error` string when it supplied one, use a fixed message otherwise, and log stderr at the call site so the operator still keeps it.

A test should assert the **absence** — no absolute path and no server file name in the returned message — rather than only comparing against the expected string. A later fallback that leaks a differently-shaped value would still pass an equality-only test.

Found while addressing the CodeRabbit threads on #2901. That PR merged before this could go on it, so it is separate.

rubicon:
Closes #3020

### The defect

`route.ts` guarded child stderr on the spawn-failure path and not on the exit-1 path:

```ts
// guarded, with the reason stated
console.error(`/api/status: set-status.mjs failed to run: ${stderr.trim()}`);
return NextResponse.json({ error: "status update failed to run" }, { status: 500 });

// not guarded
const error = typeof parsed?.error === "string" ? parsed.error : stderr.trim() || "status update failed";
```

The unguarded fallback fires only when the CLI printed no JSON document, which the allow-list comment a few lines above already identifies as most likely a crash. So the disclosure sat on the path *more* likely to produce a stack trace, while the less likely one was covered.

### The change

`clientErrorMessage(parsed, stderr)` now decides the body text: the CLI's own `error` string when it supplied one, a fixed message otherwise. stderr is logged at the call site, so the operator keeps it and the caller never sees it. It lives beside `parseCliJson` and `trackerRowArg` in `web/src/lib/status-cli.mjs`, so it is testable without spawning anything.

### Tests

Three added, and the negative one asserts the **absence** rather than only matching the expected string:

```js
assert.equal(msg, "status update failed");
assert.doesNotMatch(msg, new RegExp(root), "no absolute path reaches the client");
assert.doesNotMatch(msg, /set-status\.mjs/, "no server file name reaches the client");
```

An equality-only assertion would pass for any future fallback that leaked a differently shaped value. These fail for anything carrying the path or the file name.

The stack-trace fixture builds its path at run time rather than writing it out, because `test-all.mjs` step 7 greps tracked sources for an absolute-path literal and would flag the file. The value is still a real absolute path when the assertion runs; only the source form differs, and the comment says why.

Verified by mutation, not by inspection: restoring `stderr.trim()` as the fallback fails `a crash that printed no JSON never puts stderr in the response body`, and only that test.

`node test-all.mjs` 4198 passed, 0 failed, 2 warnings (both pre-existing and unrelated: a Polish CV section-order notice and a `santifer.io` string in `dashboard/internal/ui/screens/stats.go`). Web suite 264 passed, `npm run typecheck` exit 0.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved status error messages by showing clear, safe client-facing errors.
  * Prevented technical stack traces, server paths, and stderr details from appearing in responses.
  * Added a consistent fallback message when no usable error information is available.

* **Tests**
  * Added coverage for error propagation, fallback handling, and suppression of sensitive diagnostic details.


santifer (approved):
This is a privacy fix on my surface and it's correct: the exit-1 path echoed child stderr (a Node stack trace with absolute server paths) while the spawn-failure path already guarded it — guarding one and not the other left the disclosure open on the path most likely to produce a trace. Extracting clientErrorMessage() to status-cli.mjs makes it testable, and the test is non-vacuous: it feeds a real stack trace and asserts doesNotMatch on both the absolute path and the server filename. Same doctrine as the origin-guard work — a stack trace never reaches the client. Merging.

santifer:
Thanks, @rubicon — this is a privacy fix that matters: the exit-1 path echoed a Node stack trace with absolute server paths while the spawn-failure path already refused to. Extracting clientErrorMessage() so it's testable, and asserting doesNotMatch on both the absolute path and the server filename with a real stack-trace fixture, is exactly the shape this deserves. Same doctrine as the API-guard work: a stack trace never reaches the client. Merged 🚀

PR #3023, opened 2026-08-18
fix(web): methodology link was a 16px tap target with no new-tab cue

### LINKED ISSUE #3021: Web (alpha): "Full methodology" link is a 16px tall tap target on the report page
### Code of Conduct

### What happened?

On the report detail page, the "Full methodology" link at the bottom of the "How career-ops scored this" disclosure is a 16px tall tap target. WCAG 2.5.8 (Target Size (Minimum), AA) asks for at least 24x24 CSS pixels.

Measured on a 390x844 viewport with Playwright, after expanding the disclosure:

```
a[href="https://career-ops.org/methodology"]  ->  113.44 x 16
computed min-height: 0px
```

Every other tap target in the web UI already handles this. `report-view.tsx`, `generate-pdf-button.tsx`, `config-form.tsx`, `discovery-card.tsx` and others carry `max-sm:min-h-[44px]`, and the sibling "posting" link in `report-view.tsx` is the same shape as this one but does have it. `score-methodology.tsx` looks like it was just missed.

Second, smaller thing on the same element: it opens in a new tab (`target="_blank" rel="noreferrer"`) and nothing announces that. A screen reader user hears "Full methodology, link" and gets moved to a new tab with no warning. WCAG technique G201 covers this, and the accessible name is the natural place to put it.

To be clear about what is not wrong here, since a target-size audit can easily read it that way: the link does have an accessible name. It computes to "Full methodology" from its visible text. Reading `innerText` while the `<details>` is collapsed returns an empty string because the content is not rendered yet, which can look like a missing name in an automated sweep. It is not one.

### Steps to reproduce

1. Open any report detail page in the web UI, for example `/pipeline/32`.
2. Set the viewport to 390x844.
3. Expand "How career-ops scored this and why it's for you".
4. Measure the "Full methodology" link at the bottom of the panel: `document.querySelector('a[href="https://career-ops.org/methodology"]').getBoundingClientRect()`.

It returns 113.44 x 16. Height needs to be 24 or more.

### Version / channel

web 0.6.1 alpha

### Screen / route

/pipeline/[id]

### CLI tool

Claude Code

### OS / Browser

macOS 15 / Chromium 141 (Playwright)

### Expected

The link is at least 24x24, matching the 44px small screen minimum the rest of the app already uses, and its accessible name says it opens in a new tab.

rubicon:
## What does this PR do?

Makes the "Full methodology" link at the bottom of the score methodology disclosure a large enough tap target, and has it announce that it opens in a new tab. It was rendering 16px tall, under the 24x24 minimum in WCAG 2.5.8 (Target Size (Minimum), AA).

## Related issue

Closes #3021

## The change

Two attributes on one anchor in `web/src/components/score-methodology.tsx`:

```diff
-          className="inline-flex items-center gap-1 text-xs text-brand transition-colors hover:underline"
+          aria-label="Full methodology (opens in a new tab)"
+          className="inline-flex min-h-[24px] items-center gap-1 text-xs text-brand transition-colors hover:underline max-sm:min-h-[44px]"
```

The `max-sm:min-h-[44px]` is the pattern the web UI already uses everywhere else for this (`report-view.tsx`, `generate-pdf-button.tsx`, `config-form.tsx`, `discovery-card.tsx`, `pipeline-view.tsx` and more). The sibling "posting" link in `report-view.tsx` is the same inline-flex external link and already carries it, so this one looks like it was simply missed rather than deliberately different. I added a plain `min-h-[24px]` alongside it because 2.5.8 applies at every viewport, not only below the `sm` breakpoint, and without it the link stays 16px on desktop.

For the new tab, the `aria-label` keeps "Full methodology" as its prefix, so the visible text is still contained in the accessible name and 2.5.3 (Label in Name) holds. The lucide icon is marked `aria-hidden` so it cannot leak into the name if the library changes its defaults.

## Verification

Measured with Playwright on a report page, expanding the disclosure first:

| viewport | before | after |
|---|---|---|
| 390x844 | 113.44 x 16, computed `min-height: 0px` | 113.44 x 44, computed `min-height: 44px` |
| 1280x900 | 113.44 x 16, computed `min-height: 0px` | 113.44 x 24, computed `min-height: 24px` |

Accessible name in the Chromium accessibility tree goes from `link "Full methodology"` to `link "Full methodology (opens in a new tab)"`.

`node test-all.mjs` gives 4198 passed, 0 failed. The two warnings it reports are pre-existing on a clean checkout and unrelated to this change (`cv-sync-check.mjs` has no user data to check, and a `santifer.io` string in `dashboard/internal/ui/screens/stats.go`). `npm --prefix web run typecheck` is clean and `npm --prefix web test` is 261 passed.

One note on testing: the web suite covers `web/src/lib` only, and there is no component render harness in the repo. Adding one to cover a two-attribute presentational change felt like the wrong trade, so the browser measurements above are the evidence instead. Happy to add a harness in a separate PR if you want that coverage.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

## Summary by CodeRabbit

* **Accessibility**
  * Improved the methodology link’s screen-reader labeling.
  * Marked the external-link icon as decorative for assistive technologies.

* **Style**
  * Added responsive sizing and a minimum height for improved usability on mobile devices.


santifer (approved):
Correct a11y fix on my surface: 16px was under WCAG 2.5.8's 24px, and `max-sm:min-h-[44px]` gets the touch target right too. The aria-label announcing the new tab + aria-hidden on the decorative icon is the complete version, not just the height. Merging.

santifer:
Thanks, @rubicon — 16px was under WCAG 2.5.8, and doing it completely (the touch-target min-height AND the aria-label announcing the new tab AND aria-hidden on the decorative icon) is the right way to fix an a11y nit rather than just bumping the height. Merged 🚀

PR #3026, opened 2026-08-18
fix(web): keep the run stream alive during silent agent phases

### LINKED ISSUE #3024: web: pdf runs die with "Connection error" because the run stream goes silent for the whole cv-html envelope
## What happens

Running the dashboard's **pdf** mode ends in `Connection error` in the job card, even though the agent finished normally and the tailored CV was rendered correctly to `output/`.

## Why

`/api/run` streams NDJSON events as the agent produces them. In `kind: "pdf"`, `createCvEnvelopeFilter()` deliberately holds back the whole 15-25 KB `<<cv-html>>` envelope so it doesn't flood the run log (#2185). That is the right call for the log — but it means the HTTP response emits **nothing at all** for as long as the model takes to write the CV.

Nothing downstream can distinguish a long silence from a hung request, so the connection is dropped and the client falls into its `catch` and reports `Connection error`.

## Measurements

Instrumented on a real pdf run (report 63, Monumental), timestamping each stream event:

| | before |
|---|---|
| stream events | 64 |
| **longest idle gap** | **149 s** (and a second at 64 s) |
| outcome | connection dropped, `Connection error`, PDF present on disk |

The run itself was healthy: 329 s wall clock, 107 k tokens, `{"type":"done"}` emitted server-side.

## Proposed fix

A 10-second `setInterval` emitting `{"type":"keepalive"}` for the life of the stream, cleared in `close()` (which every exit path already calls). Unknown event types fall through the client's `switch`, so tabs on an older build are unaffected.

After the fix, same job shape: **max idle gap 10 s** over 55 events, 22 keepalives, job completes in the UI.

Worth noting for whoever picks this up: the keepalive **must** be a timer. My first attempt hooked it into `sendAgentText()` and it still left a 149 s gap — piggy-backing on agent output cannot fire during exactly the silences it needs to cover.

PR follows.

Lukas-drz:
Closes #3024

`/api/run` emits nothing for as long as the agent takes to write the 15-25 KB `<<cv-html>>` envelope, because `createCvEnvelopeFilter()` holds every byte back from the stream. The connection gets dropped and the client reports `Connection error` while the agent has finished and the PDF has rendered.

**Measured on a real pdf run** (report 63): idle gaps of **149 s** and 64 s across 64 events, connection dropped. After: **max gap 10 s** over 55 events, 22 keepalives, job completes.

The interval is cleared in `close()`, which every exit path already calls. Unknown event types are ignored by the client's `switch`, so an open tab on the previous build is unaffected.

Note for review: it has to be a timer, not a hook on `sendAgentText()`. I tried the latter first and it still left the 149 s gap — output-driven keepalives cannot fire during the silences they exist to cover.

`tsc --noEmit` passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of long-running operations by keeping the response connection active during processing.
  * Prevented connection timeouts while PDF CV content is being filtered.

### DIFF
--- a/web/src/app/api/run/route.ts
+++ b/web/src/app/api/run/route.ts
@@ -213,9 +213,20 @@ export async function POST(req: Request) {
         if (closed) return;
         try { controller.enqueue(enc.encode(JSON.stringify(obj) + "\n")); } catch { closed = true; }
       };
+      // Time-based keepalive. The stream is silent whenever the agent is thinking
+      // or inside a long tool call, and in pdf mode it is silent for the whole
+      // 15-25 KB <<cv-html>> envelope (cvFilter swallows every byte). Measured
+      // idle gaps on a real pdf run reached 149s — long enough for the browser or
+      // a proxy to drop the connection, after which the client reports
+      // "Connection error" even though the agent finished and the PDF rendered.
+      // It must be a timer, not a hook on incoming text: piggy-backing on agent
+      // output cannot fire during exactly the silences it needs to cover.
+      // Unknown event types are ignored by the client's switch, so old tabs are safe.
+      const heartbeat = setInterval(() => send({ type: "keepalive" }), 10_000);
       const close = () => {
         if (!closed) {
           closed = true;
+          clearInterval(heartbeat);
           if (killer) clearTimeout(killer);
           releaseWriteTokenOnce();
           try { controller.close(); } catch { /* */ }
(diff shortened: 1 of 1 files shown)


santifer (approved):
Real fix, and the reasoning is right where it counts. A keepalive must be a timer, not a hook on agent output — piggy-backing on text can't fire during exactly the silences (the 15-25KB cvFilter-swallowed envelope) it needs to cover. I verified the additive-safety claim myself: the run stream's consumer (job-store.tsx:177-194) is an if/else-if chain with no throwing default, so `type: "keepalive"` falls through and is ignored — old tabs are safe. Clearing the heartbeat in send() on client-disconnect (not waiting for close()) is the right call given the child can run for minutes. Merging.

santifer:
Thanks, @Lukas-drz — and welcome; this is a clean first contribution. The insight that a keepalive has to be a timer rather than a hook on agent output is the crux: piggy-backing on text can't fire during exactly the silences (the 15-25KB cvFilter-swallowed envelope) it needs to cover. Measuring the real 149s idle gap, and clearing the heartbeat on client-disconnect so an abandoned run doesn't leak a live timer, are the details that make it solid. The unknown-event-type is safely ignored by the client's switch, so older tabs are fine. Merged 🚀

PR #3027, opened 2026-08-18
fix(web): serve the tailored CV for the requested report, not the newest one

### LINKED ISSUE #3025: web: /api/cv-pdf serves the newest CV for a company, so two roles at the same employer resolve to the same PDF
## What happens

With two evaluated roles at the **same employer**, "View tailored CV" opens the **same** PDF for both.

Reproduction from my own tracker:

| report | role | tailored CV generated |
|---|---|---|
| 80 | ANYbotics — Field Deployment Technician (Barcelona) | `cv-…-anybotics-field-deployment-technician-…pdf` (07:59) |
| 81 | ANYbotics — Robotics Test Engineer (Zurich) | `cv-…-anybotics-robotics-test-engineer-…pdf` (07:55) |

Both rows opened the **Barcelona** CV, because it was written four minutes later.

## Why

`/api/cv-pdf` takes only `?company=`, lists `output/*.pdf`, filters by company slug and sorts by mtime:

```js
files.sort((a, b) => fs.statSync(...b).mtimeMs - fs.statSync(...a).mtimeMs);
const file = path.join(dir, files[0]);
```

Company is not a unique key for a tailored CV. One CV per *offer* is the whole point of the pdf mode.

## The mapping already exists

`generate-pdf.mjs` writes `data/pdf-index.tsv` on every render:

```
80	output/cv-…-field-deployment-technician-….pdf	.career-ops-web/pdf-tmp/cv-web-80.html	a4	2026-08-12
81	output/cv-…-robotics-test-engineer-….pdf	.career-ops-web/pdf-tmp/cv-web-81.html	a4	2026-08-12
```

Nothing reads it. The index is correct; only the lookup is wrong.

## Proposed fix

`resolveCvByReport(report)` reads `pdf-index.tsv` (last row wins, so a regenerated CV supersedes; resolved path checked to stay inside the project root). `/api/cv-pdf` prefers it and `GeneratePdfButton` passes `report=`. Company matching stays as the fallback for rows generated before the index existed.

## Related, not fixed here

`resolveTailoredCv()` in `web/src/lib/apply/cv.ts` has the same defect on the **apply** path, where it decides which file is *attached to a real application* — a wrong match sends the wrong CV rather than merely displaying it. I left it alone because the apply `Session` carries no report number and threading it through the provider is a larger change than this fix. Happy to open a separate issue if you want it tracked.

PR follows.

Lukas-drz:
Closes #3025

`/api/cv-pdf` matched `output/` by company slug and served the most recently modified match, so two evaluated roles at the same employer both resolved to the same PDF — ANYbotics report 80 (Barcelona) and report 81 (Zurich) both opened the Barcelona CV.

`data/pdf-index.tsv` already maps report number to PDF path and nothing read it. `resolveCvByReport()` reads it (last row wins so a regenerated CV supersedes; the resolved path is checked to stay inside the project root), the button passes `report=`, and company matching stays as the fallback for rows generated before the index existed.

**Not fixed here:** `resolveTailoredCv()` has the same defect on the apply path, where it picks the file *attached to a real application*. The apply `Session` carries no report number, so that needs the number threaded through the provider — flagged in #3025, happy to take it as a follow-up if you want it.

`tsc --noEmit` passes.

## Summary by CodeRabbit

* **New Features**
  * Tailored CV PDF requests can now use a specific report identifier for more accurate document retrieval.
  * PDFs open directly in the browser with updated caching behavior.
* **Bug Fixes**
  * Improved handling of missing, invalid, or unavailable report-linked PDFs.
  * Company-based PDF lookup remains available when no matching report entry is found.
  * Requests now return clearer errors when an indexed PDF is missing or cannot be read.

### DIFF
--- a/web/src/app/api/cv-pdf/route.ts
+++ b/web/src/app/api/cv-pdf/route.ts
@@ -16,14 +16,20 @@ export async function GET(req: NextRequest) {
   // the same file (Zurich vs Barcelona at ANYbotics). Company stays as the
   // fallback for rows generated before the index existed.
   const byReport = resolveCvByReport(req.nextUrl.searchParams.get("report") ?? undefined);
-  if (byReport) {
+  // An indexed report whose file has since been deleted must 404 rather than
+  // fall back: company matching would hand back another role's CV at the same
+  // employer, which is the bug this endpoint is being fixed for.
+  if (byReport.status === "missing-file") {
+    return new Response("the tailored CV for this report is indexed but missing from output/", { status: 404 });
+  }
+  if (byReport.status === "ok") {
     try {
-      const buf = fs.readFileSync(byReport);
+      const buf = fs.readFileSync(byReport.path);
       return new Response(new Uint8Array(buf), {
         status: 200,
         headers: {
           "Content-Type": "application/pdf",
-          "Content-Disposition": `inline; filename="${path.basename(byReport)}"`,
+          "Content-Disposition": `inline; filename="${path.basename(byReport.path)}"`,
           "Cache-Control": "no-store",
         },
       });
--- a/web/src/app/api/cv-pdf/route.ts
+++ b/web/src/app/api/cv-pdf/route.ts
@@ -2,6 +2,7 @@ import { NextRequest } from "next/server";
 import fs from "node:fs";
 import path from "node:path";
 import { careerOpsRoot } from "@/lib/career-ops";
+import { resolveCvByReport } from "@/lib/apply/cv";
 
 export const runtime = "nodejs";
 export const dynamic = "force-dynamic";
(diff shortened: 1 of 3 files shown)


Lukas-drz:
Closing this as a duplicate of #2599, which I should have found before opening it — my apologies for the noise.

@gregaro's PR predates mine by eleven days and is a superset in every respect:

- it fixes the **apply** path as well, which is the dangerous one (the CV actually attached to a real application, not merely displayed);
- it consolidates both routes onto a single `resolveTailoredCv(company, application)` instead of leaving two parallel resolution paths;
- it ships `cv-selection.mjs` with tests, where mine has none;
- and it resolves from the tracker application number through the report cell, which is the sounder entry point than my `report=` param.

`reportNumberFromCell()` requiring the full `[010](` inline-link form — so prose like `see [010] before [011](...)` can't resolve to the wrong report — is a case I hadn't considered, and the reasoning in that comment is right: a wrong CV attached silently is worse than no CV found.

I'll leave one small review note on #2599 about an edge case I ran into, and the reproduction from my issue (#3025 — ANYbotics reports 80 and 81 both opening the Barcelona CV) is available there if it's useful as a test case.

#3026 is unrelated and stays open.

PR #3030, opened 2026-08-18
feat(providers): add yourator job board provider

HeiTang:
## What

Adds `providers/yourator.mjs` — [Yourator](https://www.yourator.co), a Taiwanese job board for startup and digital roles. This is the first Taiwan-based source in the registry.

The endpoint is public, zero-auth, and needs no cookie or `Referer`:

```
GET https://www.yourator.co/api/v4/jobs?page=N
→ { payload: { hasMore, currentPage, nextPage, jobs: [ … ] } }
```

## Source Indexing Policy

Sampled live on **2026-08-18**: 88 pages walked, **1,760 rows, 1,760 unique ids, 151 distinct employers**.

**Rule 1 — real listings, identifiable employer, free for candidates.** Verified: every row carries `company.brand` (151 distinct employers, real named companies — HourLoop, 國泰金控, PChome, 7-11, GE Aerospace, Ogilvy…). No candidate-side paywall or registration: the feed and the postings are readable logged-out, and applying is free.

**Rule 2 — canonical URL is the employer's.** The provider emits `thirdPartyUrl` (the employer's own ATS) and falls back to the Yourator posting page only when a row has none. Verified across the full 1,760-row walk:

- 636 rows (36.1%) carry a `thirdPartyUrl`; **all 636 are `https:`, zero point back at yourator.co**
- host distribution is ordinary employer infrastructure: `teamdoor.io` per-tenant subdomains (548 — Teamdoor is a Taiwanese ATS), `recruit.cathayholdings.com` (57, self-hosted), `grnh.se` (9), `jobs.lever.co` (7), `job-boards.greenhouse.io` + `boards.greenhouse.io` (6), `amazingtalker.breezy.hr` (5), `17live.bamboohr.com` (3), `sinyi.surveycake.biz` (1) — 40 distinct hosts
- **Disclosed:** the 9 `grnh.se` rows are Greenhouse's own shortlink, i.e. one redirect hop rather than the final board URL. That is what the source exposes for those rows; flagging it rather than silently treating it as the shortest path.
- the remaining 1,124 rows (63.9%) expose no employer URL at all, so they travel with the board page as the policy allows

`utm_*` parameters are stripped from the emitted URL (5 rows carry `utm_source=yourator&utm_medium=ads&utm_campaign=…`). Two reasons: a role cross-listed here and scanned directly through Breezy/Greenhouse must resolve to the same dedup key, and rule 3 says paid placement does not reach the candidate — which should include not forwarding its attribution. Only `utm_*` keys are removed; functional params survive.

**Rule 3 — complete inventory, no paid placement.** The API exposes no free-text search parameter — `q`, `keyword`, `search`, `term`, `query` and `title` are all accepted and silently ignored, each returning the unfiltered board — and unlike remotli there is no coverage flag to set: the default view **is** the complete inventory (88 pages, `hasMore` false at the end, all ids unique). The provider walks all of it.

That is load-bearing here, not boilerplate. The board's ordering is not purely chronological, and the ad-carrying rows cluster at the front:

| pages | rows | rows with `utm_medium=ads` |
|---|---|---|
| 1–18 | 360 | **5** |
| 19–36 | 360 | 0 |
| 37–54 | 360 | 0 |
| 55–72 | 360 | 0 |
| 73–90 | 320 | 0 |

An earlier revision of this provider defaulted to 10 pages. That would have captured **100% of the promoted rows and 11% of the board** — exactly the response bias rule 3 exists to prevent. `max_pages` (default 120) is now a safety bound above the observed 88, not a coverage setting.

No `promoted`/`sponsored`/`featured` field exists in the payload; the only placement signal is the UTM tagging above.

**Rule 4 — operator declared.** Operator is **友睿資訊股份有限公司** (Yourator), named on the board's own [privacy page](https://www.yourator.co/privacy). I am **not affiliated with Yourator** — this is a third-party submission. Declared in the [supported boards table](docs/SUPPORTED_JOB_BOARDS.md) row per the policy.

**Rule 5 — aggregation stays in core.** The provider reads its own source and returns everything; `title_filter` / `content_filter` / location filtering and all ranking stay in `scan.mjs`.

## Also checked

- **Zero-auth**: a plain `curl` with no key, cookie, session or `Referer` returns the feed.
- **`robots.txt`**: only `Disallow: /r/*` — the API path and job pages are not disallowed.
- **SSRF**: `assertYouratorUrl` host-locks the API URL this provider requests and every page fetch passes `redirect: 'error'`. The emitted URL is deliberately **not** host-pinned (following `remotli.mjs` and `jobvite.mjs`) since it is display-only and never fetched here.
- **Path escapes**: `path` values shaped `//evil.example/x` or `/\evil.example/x` resolve away from `SITE_ORIGIN` under the WHATWG URL parser, so they are rejected before parsing as well as re-checked after. Covered by a test.
- **No `postedAt`**: the payload publishes only `lastActiveAt`, a localized relative string ("一天內更新"), with no absolute timestamp anywhere. Omitted per the `Job` contract rather than synthesized, which would corrupt `scan-ats-full.mjs` recency filtering.

## Tests

`tests/providers/yourator.test.mjs` — 13 assertions, auto-discovered, no registration.

```
node test-all.mjs --quick
📊 Results: 4210 passed, 0 failed, 2 warnings
```

Both warnings are pre-existing and unrelated (`cv-sync-check.mjs` without user data; a `santifer.io` string in `dashboard/internal/ui/screens/stats.go`).

## portals.yml

```yaml
job_boards:
  - name: Yourator (Taiwan startup board)
    provider: yourator
    careers_url: https://www.yourator.co/jobs
    enabled: true
```

## Summary by CodeRabbit

- **New Features**
  - Added Yourator as a supported job board.
  - Added job retrieval from Yourator’s public listings, including pagination and title filtering.
  - Added employer ATS links when available, with cleaned tracking parameters and safe fallback posting URLs.
  - Added company attribution and standardized job details.
  - Skips invalid or incomplete listings and omits unavailable publication timestamps.

- **Documentation**
  - Documented Yourator’s supported features, configuration, pagination behavior, limits, and URL handling.


tests/providers/yourator.test.mjs
```diff
@@ -0,0 +1,179 @@
...
+  } else {
+    fail(`normalizeYouratorJob drops = ${JSON.stringify(drops)}`);
+  }
+
+  // --- fetch() -------------------------------------------------------------
+
+  const mk = (i) => ({ name: `Role ${i}`, path: `/companies/co/jobs/${i}`, location: '臺北市', company: { brand: `Co ${i}` } });
+  const pages = {
+    1: { payload: { hasMore: true, jobs: Array.from({ length: 20 }, (_, i) => mk(i)) } },
+    2: { payload: { hasMore: false, jobs: [mk(20), mk(21), { name: '', path: '/companies/co/jobs/bad' }] } },
```
Scott-Emberson: Nice sentinel work on this test: the page-3 "must never be requested" fixture means an over-running loop fails on both the request count and the job count, not silently. One gap in this specific case, though.

The pass message names `stops on hasMore:false`, but this page-2 fixture is 3 rows, which is also `< PER_PAGE` (20). The provider breaks on `hasMore !== true || jobs.length < PER_PAGE`, so here both operands are true at once and the test can't tell which one fired. Drop the `hasMore` check from the provider entirely and this test still passes: page 1 has 20 rows so no early break, page 2 has 3 so it stops, same two requests, same 22 jobs. The assertion that's meant to guard the "walk until hasMore turns false" contract doesn't actually guard it.

One-line fix: make page 2 a full 20-row page (fold the bad row in among the 20) while keeping `hasMore:false`, so only the `hasMore` operand can cause the stop. Then the test pins the mechanism its label promises. Not blocking, the provider is correct as written, but rule-3 completeness is the whole point of this provider and it's worth having the test enforce it.

HeiTang: Good catch — addressed in 68fa0f4.

The final-page fixture now contains 20 raw rows (19 valid jobs and one invalid row) while keeping hasMore: false. That means the test can no longer pass through a short-page heuristic; it must stop on hasMore.

node tests/providers/yourator.test.mjs passes.

providers/yourator.mjs
```diff
@@ -0,0 +1,205 @@
...
+        throw new Error(
+          `yourator: unexpected API response on page ${page} — expected { payload: { jobs: [...] } }, got keys: [${json ? Object.keys(json).join(', ') : 'null'}]`,
+        );
+      }
+      for (const j of jobs) {
+        const normalized = normalizeYouratorJob(j, fallbackCompany);
+        if (normalized) out.push(normalized);
+      }
+      // Past the last page the API answers with an empty array and hasMore:false.
+      if (json.payload.hasMore !== true || jobs.length < PER_PAGE) break;
```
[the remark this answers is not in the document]
HeiTang: Addressed in 07945e3 and 68fa0f4.

Pagination now stops solely when payload.hasMore is not true, and the regression test covers a short intermediate page with hasMore: true. The existing documentation already describes this contract. The final-page fixture was also hardened so hasMore: false is isolated as the stop signal.

node tests/providers/yourator.test.mjs passes.

santifer:
Welcome to career-ops, @HeiTang — and what a first contribution. The response-bias measurement is the best policy application we've seen from a provider author: proving the API ignores search params, walking all 88 pages, and then showing the five ad-carrying rows all sat in pages 1-18 — so a shallow default would have captured 100% of the promoted rows and 11% of the board. That's understanding WHY rule 3 exists, not checking a box. Omitting postedAt rather than synthesizing it from a localized relative string was equally right (a guessed date silently corrupts recency filtering). First Taiwan source in the registry, and the bar for the next provider PR just went up. Merged — your first contribution. 🚀

Scott-Emberson (approved):
Re-reviewed at the current head. The fixture concern I raised earlier is resolved in 68fa0f4: the final page now carries 20 rows (19 valid plus one bad) with `hasMore:false`, so the "stops on hasMore:false" assertion can no longer be satisfied by a short-page heuristic, and the separate short-intermediate-page regression covers the other direction. I re-ran `tests/providers/yourator.test.mjs` and it is 14/14. The sentinels are real: page 3 must never be requested, the `ctx.maxPages` probe proves the health bound wins, and a malformed payload must throw.

Approving from the /tests/ owner side. The provider module itself is outside my area.

PR #3032, opened 2026-08-18
feat(doctor): emit the Gemini Node 20+ verdict instead of leaving it in prose

santifer:
## What does this PR do?

Makes `doctor` emit the "the Gemini CLI integration needs Node 20+" verdict, so the fact stops living only in prose.

## Why

The Gemini integration needs Node 20+ while the rest of the project needs 18+. Until now that difference was documented in `docs/SETUP.md` and nowhere else that a user actually reads on the way in:

- the docs site said "Node.js 18+ (22.5+ recommended)" and offered Gemini as an install path **without the note**, so someone on Node 18 who picked that path hit the requirement at runtime with no warning;
- the Spanish and French pages had drifted the *other* way and asserted "Node 20 LTS" as a **general** requirement, which is false, while their own troubleshooting sections still said 18+.

Prose copies of a fact drift. A runtime verdict does not: `doctor` already is the executable truth for prerequisites, and consumers (the web dashboard reads `/api/doctor` rather than restating the list) derive from it. This puts the Gemini floor in that same channel.

## What's in it

- `lib/gemini-node-floor.mjs` — the floor stated once, as a pure function.
- Scoped to the CLI actually in use: every other CLI gets no verdict, so a Claude Code user on Node 18 isn't nagged about a requirement that isn't theirs.
- Fails closed on an unreadable version instead of reporting all-clear.
- `tests/gemini-node-floor.test.mjs` — 23 assertions covering scope, both branches and the boundary (19 fails, 20 passes). The failing branch is the one that protects the user, and it can't be exercised through `doctor.mjs` on a machine already running a recent Node, which is why the logic is a separate pure module.
- Registered in `SYSTEM_PATHS`; coverage validator green.

## Type of change

- [x] New feature (a check that did not exist)

## Checklist

- [x] Full suite green: 4220 passed, 0 failed
- [x] `validate-system-paths-coverage.mjs`: OK, 1100 tracked files covered

## Summary by CodeRabbit

- **New Features**
  - Added diagnostics for the Gemini CLI’s Node.js version requirement.
  - Node.js 20 or newer is supported; older or unreadable versions now include upgrade guidance.

- **Bug Fixes**
  - Diagnostics now omit checks that do not apply to the selected CLI.

- **Maintenance**
  - Included the new Gemini compatibility check in system updates.

### DIFF
--- a/doctor.mjs
+++ b/doctor.mjs
@@ -15,6 +15,7 @@ import { discoverPlugins, pluginRoots, pluginStatus } from './plugins/_engine.mj
 import { resolveExtractorMode } from './browser-extract.mjs';
 import { parseConfigByExtension } from './jsonc-parse.mjs';
 import { validateFlags } from './lib/cli-flags.mjs';
+import { geminiNodeFloor } from './lib/gemini-node-floor.mjs';
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const argv = process.argv.slice(2);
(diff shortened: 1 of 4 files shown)


santifer:
Self-authored; merging on green.

PR #3033, opened 2026-08-18
fix(locks): the other two copies of the protocol had the same Windows bug

santifer:
## What does this PR do?

Fixes the same Windows contention bug in the two copies of the lock protocol that #2984 missed, and replaces the test's hand-written file list with one derived from the repo.

## Why

#2984 patched `pipeline-lock.mjs` and `tracker-utils.mjs` and stated "one definition, no sibling drift". There were **four** implementations of that protocol. `followup-seed.mjs` and `portal-health-lock.mjs` were still carrying all three faces of #2777:

- a non-`EEXIST` mkdir answer treated as fatal — Windows reports a mid-flight directory as `EPERM`/`EACCES`, which killed the writer and lost its write;
- `lockCanRecover`'s stat catch answering "recoverable" to **every** failure, so "could not look" became "nothing to recover" and a live lock got deleted;
- bare `rmSync` of lock artifacts, which throws under the same contention.

Both now import the classifiers from `pipeline-lock.mjs` instead of restating the rules.

## The test change is the point

The previous test listed the two files the fix had touched, so it could only confirm what was already known — it could never have found copies three and four. It now **asks the repo**: any `.mjs` containing a `recoverGuardDir` (the protocol's signature, used nowhere else) must import the classifiers, must not restate the EEXIST-only rule, and must not carry a bare artifact `rmSync`.

Verified in both directions: dropping a fifth copy into the tree gets it named in four separate assertions, and removing it goes green.

## Type of change

- [x] Bug fix

## Checklist

- [x] Full suite green: 4218 passed, 0 failed
- [x] Fifth-copy detection verified by experiment, not by reasoning

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved lock recovery across supported platforms, including Windows contention scenarios.
  * Prevented unrecoverable lock states when lock metadata cannot be read.
  * Added safer retry and backoff behavior during lock and recovery cleanup.
  * Improved handling of lock creation, guard creation, and release failures.

* **Tests**
  * Expanded coverage to validate lock behavior across all detected implementations.
  * Added checks for consistent contention detection, recovery handling, and artifact cleanup.

### DIFF
--- a/followup-seed.mjs
+++ b/followup-seed.mjs
@@ -59,6 +59,12 @@ import { readFileSync, writeFileSync, existsSync, mkdirSync, renameSync, rmSync,
 import { join, dirname, basename, resolve, isAbsolute, relative, sep } from 'path';
 import { fileURLToPath, pathToFileURL } from 'url';
 import { createHash, randomUUID } from 'crypto';
+// Third copy of the directory-lock protocol in this repo, and the one #2984
+// missed: that fix declared "one definition, no sibling drift" while patching
+// pipeline-lock.mjs and tracker-utils.mjs, and this file was carrying all three
+// faces of #2777 the whole time. The classifiers come from pipeline-lock now,
+// so the next Windows-contention finding lands in one place instead of three.
+import { isMkdirContention, isRmContention, rmLockArtifactSync } from './pipeline-lock.mjs';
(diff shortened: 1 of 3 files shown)


santifer:
Self-authored; merging on green.

PR #3044, opened 2026-08-18
fix(cli): validate fix-slugs flags

### LINKED ISSUE #2980: fix(cli): fix-slugs.mjs silently ignores mistyped flags — delegate to lib/cli-flags.mjs

`fix-slugs.mjs` writes to `portals.yml` (dry-run by default), which makes silent flag typos worse than usual: `--aply` is ignored, the run looks like a dry-run that found nothing to do, and the user thinks the fix happened. Same class as #2919, fixed in 4 CLIs already.

**What to do**

Delegate flag validation to the shared helper, exactly like `doctor.mjs` does since #2874:

1. Import `validateFlags` from `./lib/cli-flags.mjs` and declare the real flag set: `--apply`, `--dry-run`, `--file`, `--fix` (plus `--help`/`-h`).
2. Reject unrecognized flags BEFORE touching any file, with the usage block.
3. Add a small test following `tests/cli-flags.test.mjs` (see #2780 / #2801 for scope).

**Acceptance**

- `node fix-slugs.mjs --no-such-flag` exits non-zero naming the bad flag and listing valid ones — without reading or writing `portals.yml`.
- `--help` prints usage and exits 0.
- `node test-all.mjs` green.

HariomPCode:
## Summary

Fix `fix-slugs.mjs` silently ignoring mistyped CLI flags by delegating validation to the shared `validateFlags` helper in `lib/cli-flags.mjs`, following the existing `doctor.mjs` pattern.

### Changes

* Added shared CLI flag validation for:

  * `--apply`
  * `--dry-run`
  * `--file`
  * `--fix`
  * `--help`
  * `-h`
* Unknown/mistyped flags are now rejected before `portals.yml` is checked, read, or written.
* Added `--help` / `-h` usage handling.
* Switched `--file` parsing to the shared `flagValue()` helper, supporting both `--file <path>` and `--file=<path>`.
* Added regression tests for unknown flags, help handling, validation ordering, and `--file` syntax.
* Added `fix-slugs.mjs` CLI validation checks to the full test suite.

### Testing

* `node --test tests/cli-flag-validation.test.mjs` — **10/10 passed**
* `node test-all.mjs` — **3931/3931 passed**
* `git diff --check` — **clean**

Closes #2980

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved command-line flag validation for the slug-fixing tool.
  * Added support for both `--file <path>` and `--file=<path>` formats.
  * Invalid, unknown, missing, or incomplete flags now produce clear errors before file access.

* **Documentation**
  * Added consistent help output for `--help` and `-h`.

* **Tests**
  * Added coverage for valid flags, invalid flags, help behavior, validation order, and file argument handling.

### DIFF
--- a/tests/cli-flag-validation.test.mjs
+++ b/tests/cli-flag-validation.test.mjs
@@ -1,37 +1,21 @@
-// tests/cli-flag-validation.test.mjs — the CLIs that must reject a mistyped
-// flag instead of answering from their defaults (#2919).
+// tests/cli-flag-validation.test.mjs — CLIs must reject a mistyped flag
+// instead of answering from their defaults (#2980).
 //
 // The failure class lib/cli-flags.mjs exists to end: an unrecognized flag is
 // ignored, the value flag it was meant to be falls back to its default, and
 // the script reports a result for inputs nobody asked for at exit 0. Already
 // fixed in scan-ats-full.mjs (#1633/#1635), reply-watch.mjs (#2743/#2745),
 // dedup-tracker.mjs (#2744/#2746), scan.mjs (#2270), doctor.mjs (#2874),
-// check-table-freshness.mjs (#2873), plugin-audit.mjs (#2813) and
-// upskill.mjs and archive-posting.mjs (#2977).
+// and fix-slugs.mjs (#2980).
 //
-// archive-posting.mjs is the one that does not merely report: a dropped
-// --report writes a capture with no report prefix, findable only by rebuilding
-// the filename from that day's date and the scraped company and role — so it
-// stops resolving the next day, which is the exact failure --report exists to
-// prevent. A dropped --dry-run is worse than a wrong answer: it is a live
-// navigation and a real file, from a run the user asked to be a preview.
-//
-// rejection-latency.mjs is the sharpest case and gets its own fixture: its
-// output is a blacklist suggestion, so the silent failure is a FALSE ALL-CLEAR
-// the user then acts on. The fixture below has one company 217 days past the
-// 30-day courtesy threshold, so "no post-interview silence exceeded" is proof
-// the flag was dropped and a different tracker was read.
-//
-// HERMETIC: every path is a tmpdir fixture; nothing reads or writes the real
-// data/ directory. Each run asserts the subprocess actually ran (no spawn
-// error, no signal), so a timeout cannot pass as a silent success.
+// HERMETIC: paths use tmpdir fixtures; nothing reads or writes the real data.
 import { test } from 'node:test';
 import assert from 'node:assert/strict';
 import { spawnSync } from 'node:child_process';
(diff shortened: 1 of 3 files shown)


tests/cli-flag-validation.test.mjs
```diff
@@ -0,0 +1,108 @@
...
+
+for (const [script, typo] of SCRIPTS) {
+  test(`${script} rejects ${typo} instead of falling back to its default`, () => {
+    const r = runScript(script, typo, 'some-value');
+    assert.equal(r.status, 1, `${script} ${typo} exited ${r.status}, want 1`);
+    assert.match(r.all, /unrecognized flag/i, `${script} did not name the unrecognized flag`);
+    assert.match(r.all, new RegExp(typo.replace(/^--/, '--')), `${script} did not echo ${typo} back`);
+  });
+
+  test(`${script} --help exits 0 and prints usage`, () => {
```
Scott-Emberson: Small cleanup on this file, not a correctness issue (11/11 pass at head). The `--help`, `-h`, and `--help --bogus` tests inside this loop do not reference the `typo` parameter, so with two `SCRIPTS` entries they run twice against identical arguments (in the run, `ok 2-4` duplicate `ok 6-8`, three extra `node` spawns doing nothing new). Only the typo-rejection case actually varies per script, so it is the one that belongs in the loop. Pulling these three out into a single parametrized pass keeps the coverage and drops the redundant subprocesses.

santifer:
Heads-up, @HariomPCode: this went `CONFLICTING` because of my merges, not because of anything you did. I landed several PRs touching `test-all.mjs` tonight, and yours picks up the collision.

A rebase onto current `main` should be all it needs. One thing that will save you a round trip when you do: six of these `fix(cli): validate ... flags` migrations landed today, and the ones that merged cleanly all did the same two things — routed through the shared helper in `lib/cli-flags.mjs` rather than adding their own parsing, and kept a case for each direction (a bad flag refused, and an ordinary value that merely looks flag-like still accepted).

Also worth knowing before you push: your CI shows as absent right now. That gate is a repo setting for first-time contributors, and every push re-arms it, so ping me here once you've rebased and I'll go approve the run rather than leave you waiting on it.

HariomPCode:
Rebased onto the latest main and force-pushed the updated branch. The conflicts should be resolved now. Could you please approve the CI run when you get a chance? Thanks!

santifer:
Rebase delivered exactly as asked, @HariomPCode, and the conflict was mine to cause in the first place. fix-slugs joining the shared validateFlags family matters more than most in this batch: it is a tool that WRITES portals.yml, so a mistyped --fix silently becoming a dry-run (or worse, the reverse) was the expensive failure shape. Usage synopsis in the file, entry in the shared typo matrix, all three layers. Merged! 🚀

PR #3049, opened 2026-08-18
fix(cover-letter): preserve --out subdirectories inside output/

### LINKED ISSUE #2940: generate-cover-letter --out silently flattens nested paths into output/ and exits 0
### What happened?

`generate-cover-letter.mjs --out` silently discards any directory component of the requested path and writes the PDF to `output/<basename>` instead. It exits **0** and prints a success line naming the rewritten path, so nothing signals that the request was not honoured.

`safeOutputPath()` (generate-cover-letter.mjs:26-30) is the cause:

```js
function safeOutputPath(raw) {
  // Derive a sanitized filename from raw string (strip path separators and dots)
  const filename = basename(raw).replace(/[^a-zA-Z0-9._-]/g, "-").replace(/\.{2,}/g, "-");
  return join(OUTPUT_ROOT, filename);
}
```

The path-traversal guard is legitimate, but `basename()` also flattens *legitimate* subdirectories under `output/`, and the rewrite is never reported.

This matters because it contradicts the application-scoped bundle layout that `modes/pdf.md` documents (`npm run application:init`, artifacts under `output/{NNN}-{company}-{role}/cv/tailored/vNNN/`). `generate-pdf.mjs` supports those nested paths — it calls `mkdirSync(dirname(outputPath), { recursive: true })` at line 601 — so a CV can live in a bundle but its cover letter can never join it. Anyone organising deliverables per application hits this.

### Steps to reproduce

Minimal sanitized payload (`repro-cover.json`):

```json
{
  "candidate": { "name": "Repro Tester" },
  "letter": {
    "company": "Acme",
    "role_title": "Engineer",
    "opening": "Testing the --out path handling.",
    "profile_intro": "Minimal payload used only to reproduce the output-path behaviour."
  }
}
```

```bash
mkdir -p output/_repro/v001
node generate-cover-letter.mjs --payload repro-cover.json --out output/_repro/v001/cover.pdf --format a4
echo "exit=$?"
ls output/_repro/v001/     # empty
ls output/cover.pdf        # the PDF landed here
```

Observed output:

```
✅ PDF generated: <project>/output/cover.pdf
📊 Pages: 1
📦 Size: 33.9 KB
🔗 Manifest: data/pdf-index.tsv updated (no --report given)

Cover letter PDF: <project>/output/cover.pdf
exit=0
```

The target directory existed beforehand and stayed empty.

### Expected behavior

One of, in decreasing order of preference:

1. **Honour paths that stay inside `output/`.** Resolve the requested path, confirm the result is still contained in `OUTPUT_ROOT` (reject `..` escapes and absolute paths outside it), then `mkdirSync(dirname(...), { recursive: true })` before writing — matching what `generate-pdf.mjs` already does. This makes cover letters usable inside application bundles.
2. **At minimum, stop failing silently.** If the path must be rewritten, print a warning naming both the requested and the actual path, and exit non-zero — so a caller organising artifacts per application can detect it rather than discovering an empty folder later.

Happy to open a PR for option 1 if that direction is agreeable.

### CLI tool

Claude Code

### OS and Node.js version

Windows 11 / Node 25.2.1 / career-ops 1.26.0

tonycoder-hub:
## What does this PR do?

`generate-cover-letter.mjs --out` used `basename()` and always wrote `output/<file>`, so a legitimate bundle path such as `output/012-acme-vp-marketing/cover-letter/v2/carta.pdf` was flattened to `output/carta.pdf` and the process still exited 0. Honour paths that stay inside `output/` (mkdir is already handled by `generate-pdf.mjs`) and refuse escapes instead of silently rewriting them.

Tests cover both directions from #2940: keep a subdirectory under `output/`, and reject `..` / absolute escapes rather than flattening them.

## Related issue

Fixes #2940

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cover-letter output path handling.
  * Preserved valid nested output folders while blocking traversal and external paths.
  * Invalid output paths are now reported before rendering begins.

* **Tests**
  * Added coverage for default, nested, absolute, and invalid output paths.

### DIFF
--- a/generate-cover-letter.mjs
+++ b/generate-cover-letter.mjs
@@ -9,24 +9,65 @@
  * Fills templates/cover-letter-template.html with the payload, then renders
  * it to PDF via the same Playwright pipeline used for CVs (generate-pdf.mjs).
  *
- * `buildHtml` is exported as a pure function so the template can be tested
- * without loading Playwright (renderHtmlToPdf is imported lazily inside main).
+ * `buildHtml` and `safeOutputPath` are exported as pure functions so the
+ * template and --out path guard can be tested without loading Playwright
+ * (renderHtmlToPdf is imported lazily inside main).
  */
 
 import { readFileSync, existsSync, mkdirSync } from "fs";
-import { dirname, resolve, basename, join } from "path";
+import { dirname, resolve, join, relative, isAbsolute } from "path";
 import { fileURLToPath, pathToFileURL } from "url";
 import { parseArgs } from "util";
 import { assertFacts } from "./verify-cv-facts.mjs";
 import { resolveTemplate } from "./cv-templates.mjs";
 
-const OUTPUT_ROOT = resolve("output");
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const OUTPUT_ROOT = resolve(__dirname, "output");
 
-/** Sanitize a requested output filename and keep it under the output directory. */
-function safeOutputPath(raw) {
-  // Derive a sanitized filename from raw string (strip path separators and dots)
-  const filename = basename(raw).replace(/[^a-zA-Z0-9._-]/g, "-").replace(/\.{2,}/g, "-");
-  return join(OUTPUT_ROOT, filename);
+/**
+ * Resolve a requested cover-letter output path.
+ *
+ * Paths that stay inside `output/` keep their relative subdirectory (the
+ * application-bundle layout `generate-pdf.mjs` already supports). Paths that
+ * would escape `output/` — `..` traversal or an absolute path outside it —
+ * are rejected instead of being silently flattened to `output/<basename>`.
+ *
+ * @param {string} raw - Caller-supplied --out / payload.output_path value.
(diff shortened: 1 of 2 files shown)


Scott-Emberson (approved):
Verified at head. `tests/cover-letter-output-path.test.mjs` drives the exported pure `safeOutputPath`, 10/10 pass, hermetic (no fs writes). The traversal control is present and deliberate: five escape variants (parent, nested ../../, absolute-outside, /tmp, bare dir) must throw, with an independent `underOutput()` containment oracle on the honoured nested case. Two non-blocking notes: it unit-tests the resolver, not the on-disk nested write, and the positive cases assume cwd is the repo root. Approving from the /tests/ owner side.

santifer:
Welcome to career-ops, @tonycoder-hub! safeOutputPath exported pure with the traversal guard testable without Playwright, and OUTPUT_ROOT anchored to the module instead of whatever cwd the caller had — both halves right. Merged — your first contribution here 🚀

PR #3051, opened 2026-08-18
fix(portals.example): name the support ROLE, not the whole domain

artemtrofymenko:
Follow-up to #2986, and a continuation of the review I left there. @KehanLiu — this is not a complaint about your change; the reasoning in it is right and the `Thesis` qualification is the same idea applied one keyword earlier.

`title_filter` negatives are plain substrings, so a **domain** word used as a negative also catches every seniority above it. Measured on a 2,399-title corpus, six titles contain the phrase:

```
Customer Care Onboarding - Spanish Market            rejected either way
Customer Care Onboarding Officer (KYCB) - French     rejected either way
Customer Care Operations (PDX)                       rejected either way
Customer Care Operations (MSP)                       rejected either way
AI Support Delivery Manager: Customer Care           passed  ->  REJECTED
AI Agents Manager: Customer Care                     passed  ->  REJECTED
```

The four support titles the negative was aimed at carry **no positive keyword at all**, so they were already excluded and the negative bought nothing for them. The two it does change are AI roles *for* a support organisation — "Customer Care" is the domain after a colon, not the job.

Narrowing to `Customer Care Agent`, the exact title the original comment names, keeps what it earned and drops what it cost. Over the same corpus **exactly two verdicts change, both back to passing**, while `Customer Care Agent` and `Senior Customer Care Agent` stay rejected.

`Kundenbetreuung` is deliberately left alone. It has the same shape, but no title carrying it appears in the corpus, so there is nothing to measure and changing it would be a guess rather than a fix.

The regression is a discovered test file (`tests/example-config-support-negatives.test.mjs`), and it asserts the distinguishing property directly — that no bare domain word sits in the negative list — not just the two verdicts. Seen red by restoring `"Customer Care"`:

```
❌ domain-qualified senior roles rejected: ["AI Agents Manager: Customer Care", …]
❌ "Customer Care" is back as a bare negative — it vetoes senior roles in that domain
```

*Scope: one personal scan corpus, N=2,399, titles only. The six above are observed, not constructed. Touches the negative list only; #2970 edits a different region of the same file and the two merge cleanly either way.*

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined title filtering to reject only `Customer Care Agent` roles while retaining unrelated senior or AI roles that mention customer care.
  * Improved support-role exclusions in the example configuration.

* **Tests**
  * Added regression coverage to verify the updated filtering behavior and prevent future misclassification.


Scott-Emberson (approved):
Verified at head against the real `templates/portals.example.yml` through the production `buildTitleFilter`, 3/3 pass. I confirmed the false-pass control by swapping in main's old config: assertions 2 and 3 fail there and pass on the new config, with assertion 1 as a still-rejected positive control, so it cannot pass as a no-op. Approving from the /tests/ owner side; the example config itself is outside my area.

santifer:
Measured, and the measurement is the argument, @artemtrofymenko. I re-ran your case against the shipped template before merging and it reproduces exactly: `Customer Care Onboarding` and `Customer Care Operations` come back false both before and after, so the broad negative was buying nothing for them; the agent titles stay rejected; and the two senior AI roles flip false to true. Of the three assertions, the one that will still be earning its keep in a year is the third: the bare domain word cannot come back without a red test. Merged 🚀

Issue #3035, opened 2026-08-18
agent-inbox: an item was lost on macOS while every writer reported success (not the #2777 shape)


santifer:
On macOS, a concurrent `agent-inbox add` lost an item **while every process reported success**. That is a different failure from #2777, and the current explanation does not cover it.

### The evidence

`test (macos-latest)`, run on an unrelated PR (no lock code in the diff):

```
7. concurrent adds do not lose items (append, not rewrite)
  ✅ every concurrent add exited cleanly
  ❌ all 30 concurrently queued items survive — kept=29 of 30
  ❌ no item is duplicated or truncated — actual=item-4, item-0, … (29 items)
```

The first line is the one that matters. Every one of the 30 processes exited 0. Nobody timed out, nobody crashed, nobody hit `LockTimeoutError`. And one item is simply not in the file.

### Why the existing explanations do not fit

- **#2777 (Windows EPERM)** — a writer *dies*. Here nothing died, and this is macOS.
- **Lock starvation** (the retry-lottery caveat in `pipeline-lock.mjs`) — a starving writer ends in `LockTimeoutError` and a non-zero exit. Here every exit was clean.
- **POSIX append atomicity** — not the mechanism anymore: since #2777 the append runs inside `withPipelineLock`, so the lock is what serialises it, not `O_APPEND`.

That leaves a write that was performed, acknowledged, and then lost — which points at the lock not actually excluding, or at the write being overwritten by something that rewrote the file. Both are worse than a starving writer, because a starving writer is at least loud.

### Where I would look first

1. **Lock stolen mid-critical-section.** `lockCanRecover` reclaims an ownerless lock after `OWNERLESS_GRACE_MS` (1s floor). A lock directory exists for a moment before its `owner.json` is written; under a 30-way burst on a slow runner, that window is real. If a second holder reclaims it there, two writers are inside at once — and each of them exits cleanly.
2. **The `.gitignore` self-heal path**, which runs on first add and writes files while other adds are in flight.
3. Whether the test's own harness can drop an item (e.g. a spawn that reports success before its write is flushed).

### Why it is worth chasing rather than re-running

The inbox is a queue other tools append to. This shape means **a request can be accepted, confirmed, and silently never executed** — the exact failure #2614 set out to remove, except now it survives the fix that was supposed to close it. It is also intermittent, so re-running goes green and buys nothing: the last three merges all passed the same test.

[...]

PR #3053, opened 2026-08-18
fix(tests): keep the end of a failing script's output, not just the head

LayerC0de:
`formatRunFailure()` is read at exactly one moment — when a run goes red — and it was discarding the half that says why.

## The defect

```js
return t.length > maxChars ? `${t.slice(0, maxChars)}\n    ... (${t.length - maxChars} more chars)` : t;
```

It keeps the **head**. The suites here print a tick per assertion, so for an over-long stream the head is the setup that *passed*, and the tail carries the `Results: N passed, M failed` line and the most recently added cases. Keeping the head means **the newer a case is, the more certain it is to be truncated away** — backwards for output that exists only to diagnose failures.

## Observed, not hypothetical

`agent-inbox-tests.mjs` grew past the 2000-char cap. A `windows-latest` failure of its §7 ([job 95877342568](https://github.com/santifer/career-ops/actions/runs/32172069946)) rendered:

```
8. the lock is never held by two processes at once
  ✅ all 30 holder processes exited cleanly
  ✅ all 30 holders recorded an interval
  ✅ all 30 holders acquired the l
    ... (102 more chars)
```

Cut mid-word, one assertion short of §8's verdict — `no two holders were inside the critical section at the same time`, added specifically to attribute that class of failure (#3035). The run reproduced the bug, the experiment ran, and the harness threw the result away. It's in the raw job log too, so it isn't GitHub's rendering.

## The change

Keep **both ends**, weighted to the tail (35/65), with the dropped-character count between them.

Deliberately not tail-only: an early stack trace is the other common shape, and dropping the head would invert the defect rather than fix it. The cap still holds — both ends together stay within `maxChars`.

## Test

Case 72, driven through a real failing `run()` rather than by reaching into the module's private `lastFailure`, so it covers the path CI actually takes. Against unmodified `tests/helpers.mjs`:

```
❌ formatRunFailure dropped the tail — a red run loses its newest assertions and its Results line
✅ formatRunFailure still keeps the head (an early stack trace survives too)
❌ formatRunFailure elides silently — a reader cannot tell output is missing
✅ formatRunFailure still caps the stream rather than printing it whole
```

Two fail on `main`; the other two **pass on main by design** — they pin the properties the fix must not break. Flagging that so four green ticks aren't read as four independent repros.

## Notes

- Local: `test-all.mjs` **4315 passed / 0 failed**. The 4 warnings are pre-existing (cv-sync, personal-data scan, CV section order, tracker merge fixtures) and unrelated.
- **Touches `tests/helpers.mjs`, which #2907 also touches.** Different region — #2907 adds `DEFAULT_SCRIPT_TIMEOUT_MS` near the top; this changes `formatRunFailure`'s `clip()`. They should merge cleanly in either order, but flagging it since #2907 is already approved and I'd rather you knew than found out at merge time.
- Kept separate from #2905 rather than folded in, per your preference for one reviewable idea per PR — and separate from #2907 so it doesn't dismiss that approval.

Until this lands, every future §7 red on either platform truncates §8's verdict away, and #3035 stays unattributable for harness reasons rather than for want of evidence.

Refs #3035

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved failure messages for commands with very long diagnostic output.
  - Failure details now retain both the beginning and end of the output, indicate omitted content, and remain within the configured length limit.
  - Short and empty output continue to display as expected, while very small limits are handled safely.

- **Tests**
  - Added regression coverage for long-output formatting, omission reporting, boundary retention, and length limits.


LayerC0de:
Good catch from CodeRabbit, and it was right — fixed in `d1279e1`.

`head + tail` already consumed `maxChars` and the marker was appended on top, so every clipped stream ran over its documented cap. The pre-PR version had the same property, but this commit is what documents `maxChars` as a cap, so it should honour it.

The marker's width now comes out of the budget. One wrinkle worth naming: the dropped count is printed *inside* the marker, so the marker's width depends on the budget and the budget depends on its width. Broken by reserving against the widest that count can ever be (`t.length`) — that can only over-reserve, never under. A cap narrower than the marker itself degrades to plain truncation rather than emitting a marker that alone overruns it.

**The more useful half: my cap assertion was too weak to catch this.** It read

```js
if (shown.length < BIG)   // BIG = 8000, cap ≈ 2000
```

which any cap whatsoever passes — it tested that clipping happened at all, not that the cap held. Replaced with the real bound, `kept + marker <= maxChars`, measured from the reported dropped-count rather than off the rendered string (the caller re-indents newlines and prefixes ` (exit N)\n stdout: `, so rendered length is not clipped length). Verified against the unbudgeted version:

```
❌ formatRunFailure overruns its own cap: 1234 chars returned for maxChars=1200 (marker not budgeted)
```

One more thing found while tightening it, unprompted: the fixture did `process.stdout.write(payload); process.exit(1)`. stdout to a pipe is async and `process.exit()` discards whatever hasn't flushed — so the payload could silently shrink and the cap assertion would pass for the wrong reason. Now exits from the write callback. Same class of bug as the one this PR exists to fix, in the test for it.

`test-all.mjs`: 4315 passed / 0 failed.

LayerC0de:
Second CodeRabbit finding, also correct — fixed in `e39c4d2`.

The comment sitting directly above the line promised "never zero head". The line didn't deliver it:

```
budget= 1  head=0  tail=1   <-- head dropped entirely
budget= 2  head=0  tail=2   <-- head dropped entirely
budget= 3  head=1  tail=2
```

`Math.floor(budget * 0.35)` rounds to 0 below budget 3, handing the whole allowance to the tail — quietly reinstating the tail-only behaviour this function exists to avoid, at exactly the sizes where losing an end matters most. Now floored at one character whenever there's room for two; a one-character budget stays single-sided, since there's no way to keep both ends of a string in one character.

**Reachability, stated plainly rather than left implied:** `budget` is `maxChars` minus the marker (~34 chars), so hitting this needs `maxChars` ≈ 36. Nothing calls it with anything near that — the default is 2000 and the one explicit call site passes 1200. So this is the code matching its own documented invariant, **not a live bug**. I'd rather say that outright than let a tidy "fixed" imply I'd found something reachable in CI.

The new assertion uses its own fixture with `<` and `>` as boundaries — characters that appear nowhere else in the rendered output (not in ` (exit 1)`, not in `stdout: `, not in the marker) — so "both ends survived" is asserted precisely rather than by matching a letter that could have come from anywhere. Against the unfloored version:

```
❌ formatRunFailure drops an end at a two-character budget (head=false, tail=true)
```

`test-all.mjs`: 4316 passed / 0 failed.

Both CodeRabbit findings on this PR were in code I'd written in the same sitting, and both were real. Worth noting since the PR is itself about a diagnostic that had been quietly wrong for a while — the cheap lesson being that an invariant in a comment is worth exactly as much as the assertion that pins it, which is why each fix here arrived with one.

Scott-Emberson (approved):
Verified at head. The change is scoped to the `clip` closure inside `formatRunFailure`, now keeping head plus tail with the cap counting the elision marker's own width. The exported surface is byte-for-byte compatible and no caller parses the marker text, so nothing downstream breaks. Ran three suites that drive `run()`/`lastRunFailure()` (run-failure-diagnostics 10/10, shell-discovery 6/6, jd-skill-gap-json-output 8/8) plus a direct check that the clipped length is exactly the cap at 2000 and 1200 with both ends retained. CodeRabbit's earlier over-cap concern is resolved at this head. One unreachable nit: a maxChars of 1-2 would floor head to zero, but every caller uses 2000. Approving from the /tests/ owner side.

santifer:
The line that makes the case is buried in your comment, @LayerC0de: "the later a case was added, the more certain it was to be cut". A truncation that always eats the newest assertions is the worst possible bias for a suite that grows, and windows-latest cutting agent-inbox-tests.mjs one assertion short of the §8 verdict is a sharp illustration, given that verdict exists to attribute that exact failure. Keeping both ends rather than flipping to tail-only is right, and you say why instead of leaving it implicit. Two details I'd have got wrong: budgeting the elision marker's own width out of the cap rather than on top of it, and exiting from the write callback because stdout to a pipe is async and `process.exit()` would have silently shrunk the fixture, making the cap assertion pass for the wrong reason. That second one is a test that would have lied. Merged 🚀

Issue #3048, opened 2026-08-18
update-system.mjs apply: false-positive SAFETY VIOLATION on pre-existing dirty user file blocks every apply


alexkons:
### Summary

`node update-system.mjs apply` aborts every time with a false-positive `SAFETY VIOLATION: User file was modified` on a *pre-existing* dirty user-layer file, even though the update never touched it. The abort path then also throws a wall of `pathspec ... did not match` errors while trying to revert brand-new upstream files that don't exist in local `HEAD`. Net effect: `apply` never succeeds, and it leaves ~200 `*.bak` sidecar files scattered through the repo on every attempt.

### Environment

- Local version: `1.25.0`
- Remote version at time of repro: `1.27.0`
- `node -v`: (run in a stock macOS Homebrew Node install)
- Repo state: clean `git status` except one user-layer file (`interview-prep/story-bank.md`) that already had local, uncommitted edits *before* running the updater (accumulated STAR+R story additions — a normal, expected state for that file per `DATA_CONTRACT.md`).

### Steps to reproduce

1. Have local, uncommitted edits in a `USER_PATHS`-covered file, e.g. `interview-prep/story-bank.md` (any file matching an `interview-prep/` prefix works).
2. Confirm `git status --porcelain` shows only that file as dirty.
3. Run `node update-system.mjs check` → `update-available`.
4. Run `node update-system.mjs apply`.

### Expected behavior

Per the code's own stated intent (`update-system.mjs:1090-1108`), the check should **skip** any file that was already dirty before the run:

```js
if (initialStatusPaths.has(file)) continue;
```

`initialStatusPaths` is captured via `gitStatusEntries()` at the very top of `apply()` (line 911), before any checkout happens, so `interview-prep/story-bank.md` should already be in that set and the loop should `continue` past it without flagging.

### Actual behavior

The run instead prints:

```
201 system file(s) differ from upstream because THIS install changed them:
  .claude-plugin/marketplace.json  (local copy saved: .claude-plugin/marketplace.json.bak)
  ... (all ~200 system files, .bak sidecar written for every one)
Keeping your versions. They will NOT receive upstream changes.
Re-run with `node update-system.mjs apply --force` to take the upstream version instead.

.gitignore: appended 58 missing rule(s): ...

SAFETY VIOLATION: User file was modified: interview-prep/story-bank.md
Aborting: user files were touched. Rolling back system files...
error: pathspec 'voice-dna.template.md' did not match any file(s) known to git
[...]

PR #3054, opened 2026-08-18
fix(updater): parse git status paths from untrimmed output

BigBrown10:
Addresses the primary abort in #3048: `apply` reporting a false `SAFETY VIOLATION: User file was modified` on a pre-existing dirty user-layer file (e.g. `interview-prep/story-bank.md`), which blocks every update on installs that have an uncommitted file before the updater runs.

## Root cause

`gitStatusEntries()` read status through `git()` → `gitIn()`, which `.trim()`s the whole buffer. `git status --porcelain` emits worktree-modified entries as ` M path` — with a leading space — and when that line is the FIRST line of the buffer, the trim rewrote it into `M path`. The parser then slices `path: line.slice(3)`, which drops the first character: ` M interview-prep/story-bank.md` parsed as `nterview-prep/story-bank.md`.

`initialStatusPaths` (the Set captured by `apply()` before any checkout) is built from those parsed paths, so the guard `if (initialStatusPaths.has(file)) continue;` never matched the real path — `apply()` treated a pre-existing user edit as if the update had touched it and aborted. A hidden-dir path on the first line was mangled the same way (`.antigravitycli/…/SKILL.md` → `antigravitycli/…`).

## Fix

`gitStatusEntries()` now reads via `gitRawIn()` (untrimmed output), keeping the leading space on the first porcelain line so `code: line.slice(0, 2)` and `path: line.slice(3)` land on the exact bytes git emitted. The function is exported with a `root` parameter (default `ROOT`) so the parse can be driven against throwaway repos in tests through the same seam production uses.

## Tests

New `tests/updater-status-paths.test.mjs` (7 cases, all passing) pins the whitespace-sensitive parsing on the exact corruption shapes, including the #3048 repro: a dirty `interview-prep/story-bank.md` as the only pre-existing change, asserting the `initialStatusPaths` guard's `has()` now matches the full path. Cases cover first-line worktree mod (` M`), staged-first-line (`M `), hidden-dir leading-dot path, mixed worktree/staged/untracked entries, clean tree, and Windows CRLF robustness.

Full suite: `node test-all.mjs --quick` → **4316 passed, 0 failed** (4 expected warnings).

## Scope

Addresses the false-positive `SAFETY VIOLATION` abort (the issue title's symptom). Two other symptoms described in #3048 — the whole-set "system files differ from upstream" comparison on a clean checkout, and the rollback `pathspec did not match` errors for net-new upstream files — are separate and not addressed here; they'd be better tracked as their own issues/linked PRs.

Related but intentionally untouched: porcelain paths containing a space arrive C-quoted (`"my notes.md"`) and the parser keeps the quotes — a pre-existing, distinct limitation from the leading-space trim bug under test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved update status detection for files in hidden directories and repositories with mixed staged and unstaged changes.
  * Prevented file paths from being misread when Git output contains leading spaces or trailing carriage returns.
  * Preserved accurate handling of pre-existing local changes, helping avoid incorrect update warnings or actions.

* **Tests**
  * Added coverage for clean repositories, varied Git status combinations, and edge-case path parsing.

### DIFF
--- a/tests/updater-status-paths.test.mjs
+++ b/tests/updater-status-paths.test.mjs
@@ -1,0 +1,226 @@
+/**
+ * updater-status-paths.test.mjs — `git status --porcelain` paths must survive
+ * parsing byte-for-byte.
+ *
+ * gitStatusEntries() feeds the updater's "did we touch a user file?" safety
+ * check: the Set of paths that were dirty BEFORE the update, the list of paths
+ * the update `changed`, and the commit-failure guard all compare against the
+ * parsed `path` field. A mistyped path is a blind spot — a user file whose
+ * path is mangled no longer matches its real counterpart, so a violation can
+ * slip past the abort.
+ *
+ * The corruption vector is `gitIn()`'s `.trim()`: `--porcelain` lines for
+ * worktree/index changes begin with a space (` M path`), and trimming the WHOLE
+ * buffer strips the first line's leading space, turning it into `M path`. The
+ * parser's `path: line.slice(3)` then drops the path's first character — for
+ * ` M data/user-file.md` the resulting `ata/user-file.md` matches nothing.
+ *
+ * These tests drive gitStatusEntries() against throwaway repos through the
+ * same seam the production code uses (gitRawIn under a root), pinning the
+ * whitespace-sensitive behaviour on both code shapes (` M` and `M `) and on the
+ * first-line position that the old `gitIn`-based implementation corrupted.
+ */
+
+import { mkdtempSync, mkdirSync, writeFileSync, rmSync } from 'fs';
+import { tmpdir } from 'os';
+import { join } from 'path';
+import { pass, fail } from './helpers.mjs';
+import { gitIn, gitStatusEntries } from '../update-system.mjs';
+
+function makeRepo() {
+  const dir = mkdtempSync(join(tmpdir(), 'co-status-paths-'));
+  const g = (...args) => gitIn(dir, ...args);
+  g('init', '-q', '-b', 'main', '.');
+  g('config', 'user.email', 'test@example.com');
+  g('config', 'user.name', 'Test');
+  g('config', 'core.quotepath', 'false');
+  return { dir, g };
+}
+
+// Every parsed entry must name a real file exactly, with the two status
+// characters that git actually emitted.
+function assertRoundTrip(label, dir, expected) {
+  const entries = gitStatusEntries(dir);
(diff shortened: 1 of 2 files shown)


tests/updater-status-paths.test.mjs
```diff
@@ -0,0 +1,226 @@
...
+  const { dir, g } = makeRepo();
+  try {
+    writeFileSync(join(dir, 'b.mjs'), 'x\n');
+    g('add', '-A');
+    g('commit', '-qm', 'base');
+    writeFileSync(join(dir, 'b.mjs'), 'y\n');
+
+    const entries = gitStatusEntries(dir);
+    if (entries.length === 1 && entries[0].path === 'b.mjs' && entries[0].code === ' M') {
+      pass('parsed paths carry no trailing carriage return');
```
Scott-Emberson: Solid suite, and the fix is well proven: I swapped `gitRawIn` back to the old `gitIn` locally and 5 of these 7 fail, so this is a real regression test, not a tautology. One nit on this case: it is titled CRLF / Windows line-ending robustness, but it writes LF files and never puts a CR into the porcelain buffer, so it only passes because git emits LF. The parser (`line.slice(3)` over `split('\n')`) does not strip `\r`, so the robustness the name promises is not actually exercised. Either rename it to something like "porcelain output is LF-delimited", or make it real by running a synthetic `\r\n`-terminated buffer through the split/slice and asserting no trailing `\r`. Non-blocking, since tests 1/3/4/7 already carry the proof.

BigBrown10: Fair catch, thanks. I made it real instead of renaming it. The parsing now lives in parsePorcelainStatus(), which strips a trailing \r from each line before slicing the path (git can't put a CR inside a path). Test 6 feeds a literal CRLF buffer (' M b.mjs\r\n?? run-now.mjs\r\n') straight through the parser and asserts 'b.mjs' comes back with no trailing CR. I also ran the old slice-on-\n-only logic against the new assertion and it fails, so the test is genuinely driving the fix now. Green locally: updater-status-paths 8/8 and updater-commit-file-modes both pass.

santifer:
Welcome to career-ops, @BigBrown10! This one matters more than its size: gitStatusEntries feeds the did-we-touch-a-user-file safety check, and a path mangled by trim() is a violation that slips past the abort. The byte-for-byte test suite you shipped with it is what keeps this class closed. Merged — your first contribution here 🚀

Issue #19, opened 2026-04-06
examples/ doesn't cover hybrid/dual-track careers (engineer + instructor, IC + manager, etc.)


OthmanAdi:
### Problem

The current examples in `examples/` (and the `profile.example.yml` schema) assume a candidate optimizing for one primary archetype. Real careers often have two equally serious tracks running in parallel:

- engineer + technical instructor (bootcamps, university, internal enablement)
- backend + SRE
- ML engineer + research
- IC + manager
- engineer + product manager

A user with a hybrid career has no template to follow today. They have to make three decisions the single-track flow does not surface:

1. **Which side leads the Summary for THIS specific JD.** Lead with the wrong side and the recruiter discards in 8 seconds.
2. **How to handle "wait, why teaching?" / "wait, why engineering?" objections** in the interview. Both kill applications when answered badly.
3. **How to set TWO different salary ranges** when the two tracks pay differently (often 20-40% apart). A single `target_range:` either prices out of one side or undersells the other.

The existing examples are silent on all three.

### Proposed solution

Add `examples/dual-track-engineer-instructor/` with three files:

- `cv.md` — fictional persona ("Sam Rivera"), showing how a Professional Summary can name both tracks in the first sentence without diluting either, and how experience entries can include both engineering and teaching bullets per role.
- `profile.yml` — config with two `fit: primary` archetypes (Senior AI Engineer + Senior Technical Instructor), plus proposed additive fields:
  - `track:` per archetype (so the evaluator knows which side it is on)
  - `sell_when:` per archetype (deterministic rule for which framing to pick per JD)
  - `compensation.alternate_ranges` (block for the second salary band, keyed by track)
- `README.md` — explains when to use dual-track (3 prerequisites), when NOT to (4 anti-patterns), an interview-objection playbook with answer templates for the three forms I have actually had to handle, and the over/underqualified failure modes table.

The proposed `track:`, `sell_when:`, and `alternate_ranges` fields are net-new but strictly additive. The existing evaluator code paths that read `archetypes[].name`, `archetypes[].fit`, and `compensation.target_range` keep working unchanged. Open question for you on whether to merge them into `profile.example.yml` as canonical optional fields — happy to do that as a follow-up PR.

Honest scope:
- Direct value for a subset of users (engineers who teach, IC+manager hybrids, DevRel, training architects, FDEs)
[...]

PR #3056, opened 2026-08-18
fix(update): refuse non-canonical paths in the local declaration file

sdkkds:
## Summary

A local-paths declaration written as `./merge-tracker.mjs` is accepted today and **protects nothing**.

`localUserPaths()` refuses a declaration that collides with the system layer, because honouring it would silently stop a system file from updating. That collision check compares strings exactly:

```js
const collision = SYSTEM_PATHS.find((sys) =>
  sys.endsWith('/') ? path.startsWith(sys) : path === sys,
);
```

`./merge-tracker.mjs !== 'merge-tracker.mjs'`, so the declaration is waved through. `userLayerViolations()` then compares declared paths against git's changed-path format, which is always canonical — so it does not match there either.

Net effect: the declaration looks accepted, raises no error, and leaves the updater free to overwrite the very file it names. That is the data loss #2421 exists to prevent, reachable from a plausible typo.

## Reproduction (on `main`, before the fix)

```bash
T=$(mktemp -d) && mkdir -p "$T/config"
echo './merge-tracker.mjs' > "$T/config/local-paths.txt"
node -e "import('./update-system.mjs').then(m => console.log(m.localUserPaths(process.argv[1])))" "$T"
```

**Observed:** `[ './merge-tracker.mjs' ]` — no throw, despite `merge-tracker.mjs` being a `SYSTEM_PATHS` entry.
**Expected:** refused and named, exactly as the plainly-spelled `merge-tracker.mjs` already is.

## Fix

Canonical spelling is required **before** the collision check: no backslash separators (git reports forward slashes), no empty segment from a repeated separator, and no `.` segment. The existing `..` and absolute-path checks are unchanged.

**Refusing rather than normalising is deliberate.** Normalising would silently accept several spellings for one path and leave `config/local-paths.txt` disagreeing with what git reports; refusing keeps one path to one spelling and names the offender, consistent with every other rejection in that function.

The documented directory-prefix form is preserved: a single trailing slash is dropped before the segment check, so `fixtures/` still parses. Dots inside a filename (`.toolrc`, `notes-v2.md`) are characters, not segments, and are untouched.

## Tests

- **#19** pins all five non-canonical shapes (`./x`, `a/./b`, `a//b`, `a\b`, `.`). Fails on `main`, passes with the fix.
- **#20** pins the canonical forms the guard must **not** break — including the trailing-slash directory, a dotfile, and a nested path — so the check cannot later be tightened into rejecting valid vocabulary.

`node test-all.mjs`: **4320 passed, 0 failed.**

## Credit

The hole was spotted by CodeRabbit during review of #2793 and rated Major. It was out of scope for that PR, which was already gated on a contract question, so it is sent separately here. The suggested diff on that review was treated as a starting point and re-derived against the current code rather than adopted as-is.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation for local path declarations.
  - Non-canonical paths—including backslashes, repeated separators, dot segments, and bare `.`—are now rejected with clear errors.
  - Directory paths may retain a single trailing slash.
  - Valid canonical file, directory, nested, dotfile, and dotted-name paths continue to be accepted.


sdkkds:
`test (windows-latest)` is red on this PR, and I don't think it's this diff — evidence below rather than an assertion. I can't re-run the job myself (no admin), so a re-run would settle it either way.

**The failure:**

```
❌ tracker-writer-lock-tests.mjs crashed (exit null, signal SIGTERM)
```

`exit null` + `SIGTERM` is the child being killed, not an assertion failing — the shape `formatRunFailure()` documents as "a child killed on timeout".

**Why I don't think it's this PR:**

- The diff is 86 insertions across two files (`update-system.mjs`, `tests/updater-local-paths.test.mjs`), 0 deletions. It touches `localUserPaths()` validation only — not `tracker-writer-lock-tests.mjs`, not `tracker-utils.mjs`, not the lock modules that suite exercises.
- The same commit passes `test (ubuntu-latest)` and `test (macos-latest)`.
- `node tracker-writer-lock-tests.mjs` on this branch, on Windows 11 / Node 24: **27 passed, 0 failed**.
- Timeouts here are per-child (`run()` defaults to `timeout: 30000` in `tests/helpers.mjs`), so a new test file cannot consume another suite's budget. The added tests take 0.9s in their own child.

**The margin, in case it's useful beyond this PR:** that suite takes **~16.5s** on my machine against the 30s default — a little over half the budget for a deliberately concurrency-heavy suite that self-reports `recover guard observed in 2/8 guard-watched cases`. On a loaded Windows runner that looks like a thin margin, which would make this intermittent for anyone rather than specific to this branch. I'm not proposing a change to it here; just recording the measurement while I had it in front of me. Happy to open a separate issue if you'd like it tracked.

Could you re-run the Windows job? If it fails again on the same suite I'll dig further rather than leave it on you.

Scott-Emberson (approved):
Verified at head. `tests/updater-local-paths.test.mjs` adds an accept/reject pair: five non-canonical spellings (leading `./`, interior `.`, empty segment, backslash, bare `.`) must throw and name the offending entry, and five canonical forms (dotfile, dotted filename, trailing-slash dir, nested path) must parse and round-trip byte-identical, so the guard neither under-rejects dotfiles nor mangles the directory form. The reject case collides with a real `SYSTEM_PATHS` entry (`batch/batch-runner.sh`), exercising the exact silent-overwrite the feature prevents. I confirmed the false-pass control: reverting the guard to a no-op turns the reject block red while the accept block stays green. 22/22 pass. Approving from the /tests/ owner side. The `update-system.mjs` change is a red line I do not own, so this approval covers the test only; the updater still needs santifer.

santifer:
This closes a quiet data-loss path, @sdkkds: a declaration spelled ./merge-tracker.mjs passed the exact-string collision check, never matched git's canonical changed-path format, and left the updater free to overwrite the very file it named. Refusing instead of normalising (one path, one spelling, and the error says so) is the right call for the one file whose job is protecting user data, and the five-shape test with the error required to NAME the path makes the contract explicit. Merged! 🚀

Issue #2994, opened 2026-08-17
Application answers are trapped in the apply session and never reach the report


EliRobinson:
## The problem

Application questions can only be answered inside the live apply form-proxy session. Two consequences:

- **The answers are lost when the tab closes.** They live in session state, so a form the user spent real thought on is gone if the session expires or they step away.
- **The `apply` mode cannot reuse them.** A user who answered "why us?" in the web UI answers it again from scratch on the CLI side, because nothing wrote it anywhere both halves can read.

`application-answers.mjs` already writes a `## Application Answers` section into the report. Nothing reads it back, so the report is a write-only destination and the UI keeps its own copy.

## Why this is on the core path

This is the apply step of discover → evaluate → tailor → apply → track. It adds no new data file and no new mode: it makes an artifact that already exists (the report section) round-trip, so the two existing surfaces stop disagreeing.

## Proposed shape

1. Give `application-answers.mjs` a read path alongside its write path: `parseApplicationAnswersSection` for the section it already writes, and `parseDraftAnswersBlockH` for the evaluation mode's `## H) Draft Application Answers`. One file owning that section in both directions is the point; two parsers in two places would drift.
2. Add an Application Questions panel on the job page, backed by the report rather than by session state.
3. Reuse the prompt-building and planner-spawning that `api/apply/prefill` currently does inline, instead of growing a second copy in a new route.

Sensitive fields keep their current treatment: legal, visa, work authorization, salary and demographic questions are never auto-filled and come back flagged for the user. Nothing submits anything.

## Questions for the maintainers

1. Is the report the right home for these answers, or would you rather they lived in their own file under `data/`?
2. Extracting the planner into a shared module touches a route that already works. Is that refactor welcome in the same change, or would you prefer it split into its own PR first?
3. Should the job-page panel and the apply form-proxy converge on one component eventually, or is keeping them separate the intent?

## Example implementation

I have a working implementation and have opened it as a draft PR so there is something concrete to react to. It is an example of one way to resolve this, not a claim that the structure is right; happy to restructure or split it.

EliRobinson:
Draft PR #2998 is now open as a worked example of one way to resolve this. It is deliberately a starting point for discussion, not a claim that the structure is right: the questions above are genuine, and I am happy to restructure, split, or drop it depending on how you want this routed.

santifer:
Same apology as on the others, @EliRobinson: five days on a proposal with three open questions is too long.

[...]

PR #3059, opened 2026-08-19
feat: version persisted application answers

gregbond:
## Summary

- add a content-addressed `answer_version_hash` for the existing `## Application Answers` report section
- add strict, metadata-only `--verify` readback without printing answer values
- seal submitted answer versions against mutation or lifecycle downgrade
- require the apply workflow to verify and retain the exact persisted version before reuse

## Why

This addresses the report round-trip and version-binding portion of #2994. The report remains the single answer store; the change adds no service, database, or parallel data file.

## User impact

Career-Ops can prove which exact answer snapshot was persisted and reused. Changed answers must be saved as `filled` before that exact version can become `submitted`, while idempotent replay remains safe.

Sensitive answer values remain out of CLI verification output, and this change adds no form submission behavior.

## Validation

- `node tests/application-answers.test.mjs`
- strict write/verify/copy readback with matching version and counts
- submitted-version immutability, replay, downgrade, and changed-content coverage

The published commit has the same Git tree as the tested local commit; only its author email metadata was changed to GitHub's noreply address so account email-privacy protection remains enabled.

### DIFF
--- a/application-answers.mjs
+++ b/application-answers.mjs
@@ -1,6 +1,7 @@
 #!/usr/bin/env node
 
 import { readFileSync, writeFileSync } from 'fs';
+import { createHash } from 'crypto';
 import { resolve } from 'path';
 import { fileURLToPath } from 'url';
 
@@ -316,6 +317,64 @@ export function parseApplicationAnswersSection(reportText, { strict = false } = 
   return snapshot;
 }
 
+/**
+ * Derive a content-addressed version for one application-answer snapshot.
+ *
+ * The hash deliberately excludes lifecycle metadata (`date` and `state`). A
+ * filled snapshot can therefore become submitted without changing the answer
+ * version, while any answer, selection, field value, file path, or file
+ * version change produces a new version. Formatting and parsing once first
+ * makes the hash independent of the input aliases accepted by the formatter.
+ */
+export function applicationAnswersVersion(snapshot = {}) {
+  const canonical = parseApplicationAnswersSection(
+    formatApplicationAnswersSection(snapshot),
+    { strict: true },
+  );
+  const content = {
+    freeText: canonical.freeText,
+    selections: canonical.selections,
+    fieldValues: canonical.fieldValues,
+    files: canonical.files,
+  };
+  return `sha256:${createHash('sha256').update(JSON.stringify(content)).digest('hex')}`;
+}
+
+function metadataForSnapshot(report, snapshot) {
+  return {
+    report,
+    present: true,
+    date: snapshot.date,
+    state: snapshot.state,
+    answer_version_hash: applicationAnswersVersion(snapshot),
+    counts: {
+      free_text: snapshot.freeText.length,
+      selections: snapshot.selections.length,
+      field_values: snapshot.fieldValues.length,
+      files: snapshot.files.length,
+    },
+  };
+}
+
+// The library parser names unreadable lines so an interactive caller can
+// repair its own Markdown. The CLI is also used for metadata-only verification,
+// where echoing such a line could disclose an answer. Preserve the structural
+// count while keeping report content out of stdout/stderr.
+function parseApplicationAnswersForCli(reportText) {
+  try {
+    return parseApplicationAnswersSection(reportText, { strict: true });
+  } catch (err) {
+    const unreadable = /^Application Answers section has (\d+) unreadable/.exec(err.message);
+    if (unreadable) {
+      throw new Error(
+        `Application Answers section failed strict verification ` +
+        `(${unreadable[1]} unreadable ${unreadable[1] === '1' ? 'entry' : 'entries'})`,
+      );
+    }
+    throw err;
+  }
+}
+
 export function upsertApplicationAnswersSection(reportText, snapshot = {}) {
   const report = String(reportText ?? '').replace(/\r\n/g, '\n');
   const section = formatApplicationAnswersSection(snapshot).trimEnd();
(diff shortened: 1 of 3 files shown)


santifer:
Sorry for the slow reply, @gregbond: four days on a first contribution is too long, and it's worse here because of what I have to tell you. Thanks for your patience.

**Someone else is building the same file.** #2998 by @EliRobinson also creates `application-answers.mjs` from scratch, also for #2994, and is also a draft. You couldn't have known: both are drafts, neither shows up in a search for the other, and #2994 is a busy thread.

That's my failure to catch, not yours, and I'd rather say it on day four than let you finish.

Where it stands, and I'm quoting myself in [#2994](https://github.com/santifer/career-ops/issues/2994) from yesterday rather than deciding something new now that your PR exists:

> The **read path in `application-answers.mjs`** is core and it is the good half. It can go in on its own.

EliRobinson opened #2994, diagnosed that the report section is written and never read, and filed #2998 two days before this PR. That half is theirs on first-come, and I'd said so publicly before I saw your work.

**But I don't think your PR is the same half, and that's the interesting part.** #2994 is about answers being trapped in the apply session and never reaching the report. Yours is about proving *which exact snapshot* was persisted and reused: content-addressed `answer_version_hash`, metadata-only `--verify` readback, sealing submitted versions against mutation and lifecycle downgrade. That's integrity of the artifact, not access to it. It sits on top of the read path rather than competing with it.

So the honest question, and I'm asking rather than assuming:

**What does a user lose today without the version binding?** Concretely: what goes wrong for someone applying to jobs, that sealing a submitted version would have prevented? If the answer is a real failure someone hits, that's an issue worth opening and this PR is the answer to it. If it's mainly correctness insurance for a file nothing reads yet, then the honest sequencing is read-path first, integrity when there's something whose integrity matters.

I ask because #3059 has no issue behind it and adds ~300 lines, and I'd rather find out I'm wrong from you than close on an assumption. You've clearly thought about the lifecycle carefully: the `filled` → `submitted` gate and idempotent replay are not things you get right by accident.

What I'd suggest, and you get to disagree:

1. Say what the version binding buys, here or in a new issue. If it's real, it gets tracked and you have a clear path.
2. Leave this as a draft while that settles. Don't rebase and don't polish it: the base is moving and I don't want you paying twice.
3. If the answer is "later, after the read path," this stays open and I'll come back to it when #2998's core half lands.

Either way, this doesn't get closed quietly. And **thank you for the note about the author email**: matching trees with only the noreply address changed is exactly the kind of thing that saves a reviewer ten confused minutes.

gregbond:
Thanks — and no apology needed; catching the duplicate on day four is worth more to me than a fast reply would have been.

The read path is Eli's. #2998 was first, you'd already said so publicly in #2994, and I'm not going to relitigate that. I'll drop the read-path portion from this PR rather than have you diff two versions of the same file.

**What a user loses today without the version binding:**

You apply to a job with answer set A. A week later you improve those answers — which is the whole point of the tool, you iterate on your material as you learn what lands. The stored answers become B.

Nothing recorded that you sent A.

Right now that's invisible, because nothing reads the file — which is exactly your point, and you're right. But the moment #2998 lands, the report *does* read it, and it shows B as what you submitted. You then prep for the interview from B. You never sent B. The recruiter is holding A.

The second-order version is worse for anyone trying to learn from outcomes: "which answers correlate with callbacks" gets computed over text that has since been edited, so the feedback loop trains on things that were never actually sent. Silently, and in the direction of your most recent edits.

There's a narrower case too: without the `filled` → `submitted` gate, a re-run of the apply flow can regenerate answers over a submitted record. LLM output isn't byte-stable, so replay quietly rewrites history rather than being a no-op.

**So I think your sequencing is right, and I'd put it more strongly than you did:** version binding isn't just "later," it's specifically *load-bearing the moment the read path exists*. Before #2998 it protects an artifact nobody reads. After #2998 it's the difference between a report that shows what you sent and one that shows what you currently think.

What I'd suggest, and I'm happy to be told otherwise:

1. I'll open an issue describing the failure above, so this PR has something behind it rather than 300 lines and a vibe.
2. This stays a draft, unrebased and unpolished, until #2998's core lands. I'd rather pay once.
3. When it lands, I'll rebase onto the read path and cut this down to just the integrity layer — the hash, the `--verify` readback, and the lifecycle gate. It should be considerably smaller once it isn't carrying its own reader.

And if you look at the issue and conclude the failure is too rare to carry 300 lines, close it — I'd rather that than have it sit open out of politeness.

chipoto69:
github-steward autopilot tick 2026-08-28 03:21 CEST: new career-ops merge-conflict surface found and reproduced.

Fresh GitHub state:
- PR #3059 is OPEN and GitHub reports mergeStateStatus=DIRTY.
- Head: gregbond:codex/2525-answer-version.
- Base: santifer/career-ops main.

Bounded conflict probe:
- Workspace: /Users/rudlord/ORGANIZED/_agent-workspaces/github-steward/career-ops-pr3059-conflict-20260828-0321/repo
- No primary worktree edited; dirty primary /Users/rudlord/ORGANIZED/JOBSEARCH/career-ops left untouched.
- Fetched origin/main and PR head into local branch pr-3059.
- Ran `git merge --no-commit --no-ff origin/main` only to reproduce conflicts, then aborted the merge.

Conflict files reproduced:
- application-answers.mjs
- modes/apply.md
- tests/application-answers.test.mjs

Risk note:
- The merge also stages a very large upstream/main delta across web, providers, tests, docs, workflows, package files, and generated/support scripts. This is not a safe mechanical autopilot resolution.

Decision:
- No force-push, no overwrite of contributor branch, no merge/self-approval.
- Safe next path: maintainer/owner decides whether #3059 should be salvaged. If yes, resolve in a dedicated branch/worktree, keep the answer-version behavior intentional in application-answers + mode docs + tests, then run at minimum `git diff --check`, `npm run lint`, focused application-answers tests, and the project test gate before requesting review.

santifer:
Direction settled in #3430 today, @gregbond: the seal moves from the `submitted` marker to the write itself (previous version kept, your content hash identifies it), after @darkpandawarrior showed that a marker-based guard misses the common path. This PR is the natural home for it if you want it; the details, and the two questions it answers, are in that thread. Hold the rebase until you've read it.

PR #3061, opened 2026-08-19
fix(i18n): preserve safety guardrails in localized modes

luochen211:
Rebuilds #2860 from the current main branch so the PR is reviewable and mergeable.

Scope is intentionally limited to the 18 localized modes/_shared.md files plus test-all.mjs. The localized guardrails are preserved, and test-all.mjs copies pipeline-lock.mjs into its isolated cadence fixture for current-main compatibility.

Validation: git diff --check; node --check test-all.mjs; node test-all.mjs --quick (4300 passed, 0 failed, 2 warnings).

Supersedes #2860.

## Summary by CodeRabbit

- **Documentation**
  - Added consistent guidance across supported languages to ensure authorship claims are evidence-based and unsupported claims are never fabricated.
  - Restricted factual claims to approved source materials, while treating external job-related content as context only.
  - Clarified that application materials may be drafted and prepared, but sending or submitting requires explicit user review and approval.

- **Tests**
  - Added coverage to verify required safeguards are present and correctly structured across all localized modes.


santifer:
The localized half of this is work I want, @luochen211, and the guardrail markers are the right instrument: anchoring on `<!-- guardrail:authorship -->` rather than on translated prose means the check survives the translation instead of policing it. A human-approval rule that exists only in English is a real gap, and eighteen files is the unglamorous way to close it.

I can't merge it as it stands, though, and the reason is not in the part you were working on.

**The `test-all.mjs` side of this diff is a revert.** It is +56 / **-389**, and CI is green precisely because deleted assertions cannot fail:

```
$ gh pr diff 3061 --name-only            # 18 localized files + test-all.mjs
$ # balance within test-all.mjs only:  added 56, deleted 389
```

Among the 38 assertions it removes:

- the `assessment-log.mjs` CLI contract, three assertions (#2797)
- the liveness frame-poll suite, five assertions
- `verify-portals` ASCII-fold, three assertions
- the merge-tracker custom-layout suite, six assertions, including *"report link preserved in Notes when the layout has no Report column"*
- `hermeticGitEnv keeps an ambient GIT_CONFIG_PARAMETERS / GIT_CONFIG out of git`

It also drops `negotiation-roi.mjs --self-test` from the script list, so a shipped feature quietly stops being exercised, and restores an older version of the header comment about where new tests go.

Those last two date the base precisely: `negotiation-roi`'s self-test entry landed 17-ago in #2950, and the header block 18-ago in #2879. So this isn't a conflict resolved the wrong way, it's an older snapshot of the file re-applied over a newer one. Rebuilding a branch by copying the old file in does exactly this, and it is invisible in review unless someone counts the minus signs, which is the part worth remembering: **green is not evidence here, because the thing being removed IS the evidence.**

**The ask.** Keep the eighteen `modes/*/_shared.md` files and the new guardrail check. Drop every other `test-all.mjs` change: rebase onto current `main` and re-apply only your own addition, rather than carrying your copy of the file. If the cadence-fixture compatibility change you mention in the description is still needed after the rebase, keep that hunk too and I'll look at it on its own.

Worth being explicit since your description says the scope is "the 18 localized files plus test-all.mjs": the description is accurate about which FILES are touched, so nothing was hidden. What it doesn't say, and what I'd want to see stated in any PR that touches this file, is what the change does to the assertion count. I'll take that as a note to myself as much as to you.

Ping me when it's rebased and I'll pick it straight up. This is the last thing standing between the human-approval rule and eighteen languages that don't have it.

santifer:
Nice mechanism, @luochen211: the `<!-- guardrail:X -->` markers plus a test-all freeze is exactly the missing checklist that #2395 diagnosed (sixteen independent translations dropped the same pair because nothing was checking). I want that mechanism in. But as written this collides with the live umbrella #2573, which deliberately chose per-language translations over an English sweep (see its Rules of engagement: one language per PR, a native speaker per language) and has 13 languages open that contributors are picking up one at a time. Three adjustments:

1. Don't delete existing translations: this diff removes the French pair that #3009 shipped three days ago (and the ru/ua equivalents from #2621/#2620 if they're touched the same way). Where a locale already carries a translated rule, keep it.
2. Re-scope the test to assert **marker presence** (all 4 markers in all 18 locales) rather than exact English rule text. That keeps the freeze while letting each locale hold its translation between markers, so the umbrella's open tasks stay valid as translation upgrades.
3. The two new guardrails (source-exclusivity, human-approval) are a real addition #2573 doesn't cover: shipping them in English as a stopgap is fine under the marker-presence test, and locales can translate them later through the same umbrella flow.

Happy to merge once it composes with #2573 instead of replacing it.

Solaris-star (commented):
I reviewed the current head `517e7b78a8401920bda826f52e584fcd8cf447b3` after the maintainer's latest scope request and checked the full 19-file diff locally.

The rebase/rebuild problem is fixed: `test-all.mjs` is now `+52/-0` against current `main`, and the full suite passes. One requested change still remains at this head, though:

- `requiredGuardrails` still stores `[marker, exact English rule]` pairs and tests `source.includes(`${marker}\n${rule}`)`. Replacing the English stopgap with a real translation while preserving the marker therefore fails the guard. This freezes the prose, not the marker contract.
- `modes/fr/_shared.md` still removes the translated authorship and no-fabrication rules already present on `main` and replaces them with the English copies. By contrast, `ru` and `ua` retain their translated pairs below the new English block, so the locales are inconsistent.

That means the current head still does not compose with the per-language work in #2573. The minimal correction is the one described in the maintainer comment: assert the four marker strings only (ideally exactly once), preserve every existing translated rule, and add only the two genuinely new source-exclusivity/human-approval stopgaps where needed.

Local verification on this exact head:

- `node test-all.mjs` — 5,074 passed, 0 failed, 3 warnings
- `git diff --check upstream/main...HEAD` — passed
- tracked/untracked secret scan vs `upstream/main` — 1,157 tracked files, `blocking_new_hits=[]`
- GitHub checks API — all currently reported checks pass

I did not modify or push the contributor's branch. The three suite warnings and remote CI remain as reported; I did not reinterpret them as failures.

Solaris-star (approved):
Reviewed the updated head `c917a6f69a724f29eee6203c03b7a9c6c249fb3b` against current `main` and the owner's three requested adjustments.

The rework now composes with #2573: the PR diff is additive (`254 insertions, 0 deletions`), so the existing French/Russian/Ukrainian translations are preserved; `test-all.mjs` checks the four marker strings rather than freezing English prose; and every one of the 18 localized `_shared.md` files carries each marker exactly once. The two new source-exclusivity and human-approval rules remain English stopgaps as explicitly allowed by the maintainer.

Validation on this exact head:

- `node --check test-all.mjs` — passed
- `npm run lint` — 516 `.mjs` files passed syntax checking
- `node test-all.mjs --quick` — 5,182 passed, 0 failed, 3 existing warnings
- direct 18-locale × 4-marker uniqueness check — passed
- `git diff --check origin/main...HEAD` — passed
- full tracked secret scan — 1,166 files; 14 existing fixture/example matches, 0 new diff hits

Limitations: `--quick` intentionally skips the dashboard build, and I did not claim native-language semantic review for all 18 locales. The changed contract is marker presence, and the previously shipped translated rules remain byte-present rather than being replaced.

luochen211:
Synced current main in commit 2d6ad96. The 18 localized `_shared.md` files retain their existing translated rules and now each carry all four guardrail markers; the test checks marker presence rather than English text. Local full suite: 5312 passed, 0 failed, 3 warnings; the Gemini smoke test was skipped locally because the external fetch is unavailable here. Please re-review after CI.

luochen211:
Refreshed checks are now green across Linux, macOS, Windows, upgrade regression, CodeQL, dependency review, and security scans. @santifer this is ready for maintainer re-review.

santifer:

santifer:
Two adjustments before this lands, @luochen211, one structural and one that makes the check worth having:

1. **The new validation lives in an inline test-all.mjs section, and the file's own header (test-all.mjs:21-33) forbids exactly that**: new tests go in `tests/**/*.test.mjs`, auto-discovered, so six people adding tests do not rebase each other. It also means `--only` targeted runs can never select your check. Please move it to its own file under tests/.

2. **Marker-only checking guards the comment, not the rule.** After c917a6f6 the check asserts only that `<!-- guardrail:* -->` exists: a localized _shared.md could carry the marker with the rule under it weakened, mistranslated or deleted, and stay green. I understand why the strict version failed (the rules are translated, so a verbatim English match can never pass): the middle ground is to assert that each marker is immediately followed by a non-empty bold `**RULE/...**` line. That catches deletion and displacement without demanding a specific translation.

The restoration work across the 18 files is the valuable part here and it reads carefully done: these two changes make sure CI actually defends it. Happy to merge with both in.

Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/localized-guardrails.test.mjs`; the 18 `modes/*/_shared.md` files are santifer's red-line and outside my verdict.

The test asserts each localized shared-mode file carries all four guardrail markers (authorship, no-fabrication, source-exclusivity, human-approval), each immediately followed by its `**RULE...**` line. It reddens if any localized safety marker or its rule is removed or displaced, or a mode file goes missing, which is exactly the drift a localization PR could introduce. Hermetic, fs-only. Green on CI. Approving the test; the modes text is santifer's.

santifer:
Eighteen locales brought back in line in one reviewable move, @luochen211: rebuilding #2860 from current main instead of rebasing a stale branch was the right call, and the cadence-fixture compatibility note shows the test side was thought through, not patched. Merged 🚀

PR #3071, opened 2026-08-19
fix(dates): seven more today-gates resolve in UTC — a cooldown opens early, a follow-up comes due early, CI fails early

### LINKED ISSUE #3070: Six more places resolve today in UTC — a cooldown opens early, a follow-up comes due early, CI fails a day early
## Summary

#2765 fixed "today" in `followup-seed.mjs`; #2932 fixed the two in `set-status.mjs`. Six more places still resolve the current calendar day as the **UTC** day, and each of them gates a decision rather than merely stamping a filename.

`new Date().toISOString().slice(0, 10)` is the UTC day, so west of Greenwich an evening run answers "today" with tomorrow:

```
instant 2026-08-18T01:30:00Z  (UTC day 2026-08-18)

  Pacific/Auckland     local 2026-08-18   agrees
  Europe/Berlin        local 2026-08-18   agrees
  America/New_York     local 2026-08-17   UTC day is TOMORROW here
  America/Los_Angeles  local 2026-08-17   UTC day is TOMORROW here
```

## Where it decides something

| file | line | what "today" decides |
|---|---|---|
| `scan.mjs` | 1032 | `today < cooldownUntil` — a **cooldown gate**. A day early, and a posting the user deliberately silenced resurfaces. `daysBetweenIsoDates(firstSeen, today)` drives the recheck window the same way. |
| `followup-cadence.mjs` | 134 | `daysSinceApp = daysBetween(appDate, now)` — the follow-up **due** decision. A day early, and the user is told to chase an employer sooner than the cadence they configured. |
| `check-table-freshness.mjs` | 574 | `expired` vs `review-due`, and `expired` **exits 1** — a CI gate. A day early it fails a run that should pass; a day late it passes a table that has gone stale. |
| `funnel-velocity.mjs` | 696 | the "waiting" figure — how long a row has sat. |
| `company-history.mjs` | 193 | `today()` is the `now` for all age math in the report. |
| `assessment-log.mjs` | 150 | the date written into the appended `data/assessments.tsv` row — a user record dated a day that has not happened. |

The scan cooldown is the one I would call user-visible first: the whole point of a cooldown is that the user asked not to see something until a date, and the gate opens early for everyone west of Greenwich for part of every day.

## What must NOT change

The same line #2765 drew and #2932 kept: only *"what day is it here"* moves. Date **arithmetic** stays on UTC-midnight parsing — `parseDate('2026-06-20')` round-tripping through `toISOString()` is internally consistent and deliberate, and every one of these files depends on it.

Concretely, the fix is `new Date(localToday())` rather than `new Date(utcToday())`: the calendar day becomes local, the midnight anchor stays UTC, and every comparison downstream keeps working unchanged.

`lib/local-today.mjs` already exists (landed with #2932) and takes an injectable `now`, so each of these can be tested at a pinned instant rather than depending on when CI runs.

abankar1:
Fixes #3070.

#2765 fixed "today" in `followup-seed.mjs`; #2932 fixed the two in `set-status.mjs`. Seven call sites across six files still asked `toISOString()` what day it is — which is the **UTC** day, so west of Greenwich an evening run answers "today" with tomorrow.

Each of these **gates a decision** rather than stamping a filename:

| file | what "today" decides |
|---|---|
| `scan.mjs` | `today < cooldownUntil` — a posting the user deliberately silenced until a date **resurfaced a day early**. Same value is the `firstSeen` date written into `scan-history.tsv`, so the row it wrote then read a day old on the next recheck. Both the production call site and the exported default parameter. |
| `followup-cadence.mjs` | `daysSinceApp` — the follow-up **due** decision. The user was told to chase an employer a day before their configured cadence. |
| `check-table-freshness.mjs` | `expired`, which **exits 1**. A CI gate: a day early it fails a run that should pass; a day late it passes a table that has gone stale. |
| `funnel-velocity.mjs` | the "waiting" figure, one day high. |
| `company-history.mjs` | the `now` every age in the report is measured against. |
| `assessment-log.mjs` | the date written into the appended `assessments.tsv` row — a user record dated a day that had not happened. |

```
instant 2026-08-18T01:30:00Z  (UTC day 2026-08-18)

  Pacific/Auckland     local 2026-08-18   agrees
  Europe/Berlin        local 2026-08-18   agrees
  America/New_York     local 2026-08-17   UTC day is TOMORROW here
  America/Los_Angeles  local 2026-08-17   UTC day is TOMORROW here
```

The scan cooldown is the one I'd rank first: the entire point of a cooldown is that the user asked not to see something until a date, and the gate opened early for everyone west of Greenwich for part of every day.

## What deliberately does not change

The line #2765 drew and #2932 kept: only *"what day is it here"* moves. Date **arithmetic** stays on UTC-midnight parsing, which every one of these files depends on. Concretely the fix is `new Date(localToday())` rather than `new Date(utcToday())` — the calendar day becomes local, the midnight anchor stays UTC, every comparison downstream is untouched.

`scan.mjs`'s exported default parameter changes, not its signature; callers passing `today` explicitly are unaffected, and there's a test for that.

## The test caught me first

The window where the UTC day and the local day disagree exists for only *part* of the UTC day. A test that reads `new Date()` therefore agrees with the buggy value most of the time and passes whether the fix is present or not.

**My first draft of the test file did exactly that** — it passed with `scan.mjs` and `company-history.mjs` reverted. I'm flagging it rather than quietly fixing it, because it is precisely the non-discriminating-test failure the file exists to prevent, and it would have shipped looking like coverage.

The tests now replace `Date` in the child process before importing the module under test, pinned to `2026-08-18T01:30:00Z` under `TZ=America/New_York`. Two guards keep that honest: one asserts the freeze actually took effect, one asserts the premise that the two days differ at that instant — so a broken freeze fails loudly instead of making everything else vacuous.

Also pinned as must-not-change: `company-history`'s `today()` keeps its **UTC-midnight anchor** while moving which day it is, and `check-table-freshness --today` still overrides (that flag is the deterministic escape hatch CI pins rely on).

## Verification

Checked for discrimination: with `scan.mjs` and `company-history.mjs` reverted, **exactly the two default-reading tests fail** and the six premise/explicit-value tests still pass.

`node test-all.mjs --quick`: 4319 pass / 0 fail (3 warnings pre-existing and environmental). One test-harness change was needed: the `followup-cadence` e2e fixture copies the script into a tmpdir, so it now stages `lib/local-today.mjs` alongside it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Assessment logs, scans, freshness checks, follow-ups, and reports now use the local calendar date instead of UTC.
  * Scan history deduplication and saved scan dates now align with the local day.
  * Explicitly supplied dates continue to override automatic date detection.

* **Tests**
  * Added timezone-aware coverage for local-date behavior, scan decisions, date anchoring, and freshness checks.

### DIFF
--- a/scan.mjs
+++ b/scan.mjs
@@ -54,6 +54,7 @@ import { normalizeCompanyName } from './invite-match.mjs';
 import { withPipelineLock } from './pipeline-lock.mjs';
 import { flagValue, hasFlag, validateFlags } from './lib/cli-flags.mjs';
 import { withPortalHealthLock } from './portal-health-lock.mjs';
+import { localToday } from './lib/local-today.mjs';
 
 try {
   const { config } = await import('dotenv');
@@ -1047,7 +1048,13 @@ function daysBetweenIsoDates(start, end) {
   return Math.floor((endDate - startDate) / (1000 * 60 * 60 * 24));
 }
 
-export function shouldDedupScanHistoryRow({ firstSeen, status = 'added' }, { recheckAfterDays = null, today = new Date().toISOString().slice(0, 10) } = {}) {
+// `today` defaults to the LOCAL calendar day, not the UTC one. This function
+// gates a COOLDOWN (`today < cooldownUntil`) — the user asked not to see a
+// posting until a date — and the UTC day is tomorrow for a west-of-Greenwich
+// evening run, so the cooldown opened a day early (#3070). The recheck window
+// below reads one day high the same way. Callers may still pass `today`
+// explicitly; only the default moves.
+export function shouldDedupScanHistoryRow({ firstSeen, status = 'added' }, { recheckAfterDays = null, today = localToday() } = {}) {
   if (PERMANENT_SCAN_HISTORY_STATUSES.has(status)) return true;
   if (status.startsWith('cooldown:')) {
     const parts = status.split(':');
@@ -2534,7 +2541,12 @@ async function main() {
   const seenCompanyRoles = dedupSnapshot.seenCompanyRoles;
 
   // 5. Fetch from each target
-  const date = new Date().toISOString().slice(0, 10);
+  // LOCAL day. This one value does two things that both care which day it is:
+  // it is the `today` buildCooldownFilter compares against, and it is the
+  // firstSeen date written into scan-history.tsv. On the UTC day a
+  // west-of-Greenwich evening scan opened cooldowns early AND stamped history
+  // rows with tomorrow, which then read one day old on the next recheck (#3070).
+  const date = localToday();
   const windows = loadReApplyWindows();
   const cooldownFilter = buildCooldownFilter(windows, date);
   let totalFilteredCooldown = 0;
--- a/assessment-log.mjs
+++ b/assessment-log.mjs
@@ -29,6 +29,7 @@
 import { readFileSync, existsSync, appendFileSync, mkdirSync } from 'fs';
 import { join, dirname } from 'path';
 import { fileURLToPath, pathToFileURL } from 'url';
+import { localToday } from './lib/local-today.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
 const LOG_PATH = join(CAREER_OPS, 'data/assessments.tsv');
@@ -147,7 +148,10 @@ function addEntry(args) {
     const m = args[i].match(/^--(company|report|platform|subject|threshold|score|stale)$/);
     if (m) { fields[m[1]] = args[i + 1] ?? ''; i++; }
   }
-  const today = new Date().toISOString().slice(0, 10);
+  // LOCAL day: this is the date written into the appended assessments.tsv row,
+  // and the UTC day stamped a user record with a day that had not happened yet
+  // for anyone west of Greenwich (#3070).
+  const today = localToday();
   let row;
   try {
     row = buildRow(fields, today);
--- a/check-table-freshness.mjs
+++ b/check-table-freshness.mjs
@@ -55,6 +55,7 @@ import { join, dirname } from 'path';
 import { fileURLToPath, pathToFileURL } from 'url';
 import * as yaml from 'js-yaml';
 import { flagValue, validateFlags } from './lib/cli-flags.mjs';
+import { localToday } from './lib/local-today.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
 const TEMPLATES_DIR = join(CAREER_OPS, 'templates');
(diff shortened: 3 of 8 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope here is the new `tests/local-today-gates.test.mjs`; the six script changes and the `test-all.mjs` fixture line are outside my area and I read them only for context.

This is the rare date test that pins the boundary instead of the refactor, and it says why in its own header. The frozen-clock sentinel matters most: without it these assertions compare whatever `new Date()` returns during the run, agree with the UTC day for most of the day, and pass with the fix reverted, which is how the earlier drafts survived. Pinning the instant to `01:30 UTC` is what makes the local/UTC disagreement observable. From there the default-clock scan case and the cooldown/recheck cases each flip if the default goes back to UTC, and `company-history today()` asserts both the local day and the surviving `T00:00:00.000Z` anchor, so a later change cannot move the day and quietly drop the anchor. Cooldown and recheck arithmetic (`scan.mjs:1034-1043`) is unchanged, so the test is fixing the boundary, not the logic. Hermetic throughout: TZ-pinned frozen-clock children, no network.

Green on all three OS legs. Approving from my side. `mergeable` still reads UNKNOWN from the API, so worth a glance that GitHub finishes computing the merge before it lands.

scan.mjs
```diff
@@ -2516,7 +2523,12 @@ async function main() {
   const seenCompanyRoles = dedupSnapshot.seenCompanyRoles;
 
   // 5. Fetch from each target
-  const date = new Date().toISOString().slice(0, 10);
+  // LOCAL day. This one value does two things that both care which day it is:
+  // it is the `today` buildCooldownFilter compares against, and it is the
+  // firstSeen date written into scan-history.tsv. On the UTC day a
+  // west-of-Greenwich evening scan opened cooldowns early AND stamped history
+  // rows with tomorrow, which then read one day old on the next recheck (#3070).
+  const date = localToday();
```
Scott-Emberson, line 2549: Seconding CodeRabbit's note here with the exact mechanism. `date` on line 2531 is a second, independent `localToday()` read. `loadDedupSnapshot` on line 2521 already resolved its own `localToday()` through `shouldDedupScanHistoryRow`'s default `today` (`scan.mjs:1039`), because `historyPolicy` carries only `recheckAfterDays` and no `today`. So the recheck window and the cooldown filter plus `firstSeen` stamp read from two separate clock reads. In practice the gap is microseconds and harmless, but a scan that crosses local midnight between the two would recheck against one day and stamp against the next, which is the same one-day class of bug #3070 is closing. Capturing the day once above line 2521 and threading it in (`{ ...historyPolicy, today: date }`) collapses it to a single read. Minor, and it does not affect my approval.

santifer:
The table of what each today DECIDES is what makes this reviewable at a glance, @abankar1: a cooldown the user deliberately set answering with tomorrow's date west of Greenwich is a real broken promise, not a cosmetic drift. Third installment of the localToday family (#2765, #2932) and each site gates a decision rather than stamping a filename. Merged! 🚀

PR #3073, opened 2026-08-19
fix(ashby): retries a permanent 404 three times and ignores Retry-After

### LINKED ISSUE #3072: ashby retries a permanent 404 three times and ignores Retry-After — the one provider still outside the shared retry
## Summary

`providers/ashby.mjs` hand-rolls its retry loop and its catch is unconditional:

```js
for (let attempt = 0; attempt <= ASHBY_RETRIES; attempt++) {
  if (attempt > 0) {
    const backoff = 1000 * 2 ** (attempt - 1) + Math.floor(Math.random() * 500);
    await sleep(backoff, ctx);
  }
  try {
    const json = await ctx.fetchJson(apiUrl, { … });
    …
  } catch (e) {
    lastErr = e;          // ← every error, including permanent ones
  }
}
```

So a board that is **gone** is asked three times:

```
error                    shared isRetryableError()   ashby.mjs
404 board gone           false                       retries
401 unauthorized         false                       retries
410 gone                 false                       retries
429 rate limited         true                        retries
503 server error         true                        retries
network timeout          true                        retries
```

Every provider that uses `fetchJsonWithRetry` gets `isRetryableError` for free and stops after the first permanent failure. ashby is one of the two providers still outside it (`meituan` is the other, but its loop is an *empty-response* retry, not an error retry — that one is fine as written).

## Why it costs something

Two dead Ashby boards in `tracked_companies` cost ~3 seconds and 4 extra requests per scan that can only ever fail. #2840 measured **684 of 3,161** Ashby boards as permanently 404 — 22% — and its whole argument is that re-probing known-dead boards is the traffic that provokes the single-host throttle in #2839. Retrying each of those three times instead of once makes exactly that problem worse, and it is the same request pattern in both cases.

It also **ignores `Retry-After` entirely**. On a 429, Ashby's own answer about when to come back is discarded in favour of a fixed 1s/2s backoff. `_http.mjs` honours the header but CLAMPS it, so a hostile or misconfigured `Retry-After: 86400` cannot stall a sweep — the local loop has neither half.

Minor, but from the same cause: the local backoff has no ceiling. With `ASHBY_RETRIES` small today that is bounded in practice, so it is a latent issue rather than a live one.

## Proposed fix

Route it through `fetchJsonWithRetry(ctx, url, opts, policy)`, which is the same shape `workday.mjs`, `oraclecloud.mjs` and `a16z-speedrun-talent.mjs` already use. The existing per-attempt timeout and `redirect: 'error'` carry over as `opts`; `ASHBY_RETRIES` becomes the policy's `retries` so the current attempt budget is preserved exactly.

abankar1:
Fixes #3072.

`providers/ashby.mjs` hand-rolled its retry loop with an unconditional catch:

```js
} catch (e) {
  lastErr = e;        // every error, including permanent ones
}
```

so a board that is **gone** was asked three times. Measured against the provider directly, counting requests:

| error | before | after |
|---|---|---|
| 404 board gone | 3 | **1** |
| 401 unauthorized | 3 | **1** |
| 410 gone | 3 | **1** |
| 429 rate limited | 3 | 3 |
| 503 server error | 3 | 3 |
| network timeout | 3 | 3 |

## Why it costs something

#2840 measured **684 of 3,161** Ashby boards as permanently 404 — 22% — and its argument is that re-probing known-dead boards is the traffic that provokes the single-host throttle in #2839. Asking each of those three times instead of once is that same problem, tripled.

The local loop also **discarded Ashby's own `Retry-After`** on a 429 in favour of a fixed backoff. `_http.mjs` honours the header but CLAMPS it, so a hostile or misconfigured `Retry-After: 86400` still cannot stall a sweep — the local loop had neither half.

## Scope

Routed through `fetchJsonWithRetry`, the same shape `workday.mjs`, `oraclecloud.mjs` and `a16z-speedrun-talent.mjs` already use. ashby was one of only two providers outside it.

**`meituan` is the other, and I left it alone deliberately** — its loop is an *empty-response* retry, not an error retry (errors either rethrow or return what was collected), so it is correct as written and converting it would change different behaviour.

## Timing is preserved, not defaulted

`baseDelayMs` stays at ashby's own **1000ms**, not the shared 500ms default. The longer backoff is why this provider had a bespoke loop in the first place — the file header explains Ashby rate-limits repeated unauthenticated hits — so only *which* errors are retried changes. The bespoke 30s timeout travels through as `opts` for the same reason.

That mattered in practice: the existing `tests/providers/ashby.test.mjs` asserts the first backoff is `>= 1000`, and it caught the shared default when I first passed only `retries`. I preserved the provider's timing rather than relaxing its test.

## Verification

The test counts **requests**, not exit codes — the call fails either way, so only the request count distinguishes the fix from the bug. Both directions are pinned, because narrowing which errors retry could easily have become retrying less:

- 404/401/410/403/400 must cost exactly one request.
- 429/500/502/503 and a status-less network error must still get the full three-attempt budget.
- A transient failure that then succeeds must still return the jobs; a clean fetch must still parse and make exactly one request.

Checked for discrimination: with `providers/ashby.mjs` reverted, **exactly the permanent-failure test fails** and the four others pass — the intended split, since the old loop retried everything and so satisfied every "still retries" assertion.

`node test-all.mjs --quick`: 4319 pass / 0 fail (3 warnings pre-existing and environmental). The existing ashby provider suite is unchanged and still passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Ashby job loading reliability by automatically retrying temporary HTTP and network failures.
  * Honors server-provided retry timing when available.
  * Avoids retrying permanent failures while preserving existing timeout and redirect behavior.

* **Tests**
  * Added coverage for transient failures, permanent errors, successful responses, and recovery after a failed attempt.

### DIFF
--- a/providers/ashby.mjs
+++ b/providers/ashby.mjs
@@ -11,8 +11,11 @@
 // Ashby a longer timeout plus a backoff+jitter retry (the backoff spaces
 // requests out to dodge rate-limiting).
 // See .planning/codebase/ashby-scan-abort-diagnosis.md.
+import { fetchJsonWithRetry } from './_http.mjs';
+
 const ASHBY_TIMEOUT_MS = 30_000;
 const ASHBY_RETRIES = 2;
+const ASHBY_BACKOFF_BASE_MS = 1_000;
 
 // Annualization multipliers for different compensation intervals
 const INTERVAL_MULTIPLIERS = {
@@ -106,11 +109,6 @@ function resolveApiUrl(entry) {
   return `https://api.ashbyhq.com/posting-api/job-board/${match[1]}?includeCompensation=true`;
 }
 
-function sleep(ms, ctx) {
-  if (typeof ctx?.sleep === 'function') return ctx.sleep(ms);
-  return new Promise((r) => setTimeout(r, ms));
-}
-
 // NaN-safe Date.parse — `|| undefined` would also coerce a valid epoch 0.
 function toEpochMs(value) {
   if (!value) return undefined;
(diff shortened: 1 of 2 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope here is the new `tests/providers/ashby-retry.test.mjs`; the provider change is outside my area and I read it only for context.

The test does what it claims and it has a real control. The permanent-failure case (want 1 request) and the transient case (want 3) bracket the fix from both sides: revert to the old catch-all loop and the permanent case goes to 3, disable retry and the transient case drops to 1. Either regression reddens it, so a no-op fix cannot pass. I traced the request counts against `providers/_http.mjs`: `isRetryableError` (`_http.mjs:136`) rejects 404/401/410/403/400 on the first attempt and retries 429/5xx/status-less errors to the full `retries: 2` budget, which is what the assertions expect. Hermetic too: `ctx.sleep` stubbed, no network, `pathToFileURL` for the Windows leg, and auto-discovered so it runs in CI rather than only under a direct `node --test`.

One optional note, not blocking. CodeRabbit suggested adding `ctx.sleep` timing and Retry-After clamp assertions here. I would leave them out: the clamp and Retry-After handling live in `withRetry` (`_http.mjs:197-200`), and this PR does not touch `_http.mjs`, so asserting that behavior through the ashby test would exercise the shared helper from the wrong file. The clamp belongs in `_http`'s own test when someone adds it.

Green on all three OS legs. Approving from my side; it needs a rebase off the behind state before it lands.

santifer:
Right call and, more to the point, the right *reason*, @abankar1: the local loop caught every error, so a board that is simply gone got asked three times, and with 22% of Ashby boards permanently 404 that re-probing is itself the traffic that provokes the throttle. I checked the two things the comment promises and both hold: `sleep()` in `_http.mjs` honours `ctx.sleep` exactly like the loop you deleted, so nothing wall-clock waits, and `isRetryableError` short-circuits 404/401/410 while keeping 429 and 5xx. One honest footnote: the jitter window narrows from 500 ms to the shared 250 ms, base spacing unchanged, which I read as converging on the shared policy rather than a regression. Merged 🚀

Issue #12, opened 2026-04-05
Pull requests I suggested not merged


CharlieGreenman:
I decided to a rebuild a much much better one from scratch here: https://github.com/CharlieGreenman/JobForge

Thank you for your attention to this matter.

PR #3077, opened 2026-08-19
fix(set-status): --report cannot find a row on the customized layout merge-tracker just learned to read

### LINKED ISSUE #3075: set-status --report cannot find a row on the customized layout merge-tracker just learned to read
## Summary

8668ac1 taught `merge-tracker` to find a report link in the **Notes** cell, for customized trackers that have no dedicated Report column:

```js
function extractReportNum(reportStr, notesStr = '') {
  const m = String(reportStr ?? '').match(/\[(\d+)\]/);
  if (m) return parseInt(m[1]);
  const n = String(notesStr ?? '').match(/\[(\d+)\]\([^)]*reports\/[^)]+\)/);
  return n ? parseInt(n[1]) : null;
}
```

`set-status.mjs` — the canonical write path — was not taught the same thing. It calls `extractTrackerReportNumbers(r.report)` at all four sites and never looks at Notes, so on exactly the layout that fix added support for, `--report N` cannot find a row that `merge-tracker` links happily.

## Reproduce

A customized tracker in the shape 8668ac1 names (`… | Materials | Apply Link | Follow-up | Notes`, no Report column):

```
| # | Date | Company | Role | Score | Status | Materials | Apply Link | Follow-up | Notes |
|---|------|---------|------|-------|--------|-----------|------------|-----------|-------|
| 7 | 2026-06-01 | Acme | Staff Eng | 4.2/5 | Applied | CV v3 | https://acme.example/jobs/7 | 2026-06-14 | strong fit; report [12](reports/012-acme-2026-06-01.md) |
```

```
$ node set-status.mjs --report 12 Interview --dry-run
❌ No tracker row links report #12. (Report IDs and tracker row IDs differ — use --row N to select by tracker #.)

$ node set-status.mjs --row 7 Interview --dry-run
✅ #7 Acme — Staff Eng: would set Applied → Interview
```

The row is fine; only the `--report` lookup cannot see it. And `merge-tracker` resolves the same cell to `12`.

The advice in the error is actively misleading here: it says report and row IDs differ and to use `--row N`, when the real reason is that this layout keeps its report link somewhere `set-status` does not read.

## Not a missing capability — an unpassed argument

`extractTrackerReportNumbers` already finds it when handed the text:

```
extractTrackerReportNumbers('strong fit; report [12](reports/012-acme-2026-06-01.md)')  ->  [12]
```

So the fix is passing the Notes cell as a fallback, exactly as `merge-tracker` does.

## One thing to get right

The Notes cell is free-form prose, so the fallback has to be **scoped**, which is why merge-tracker's pattern requires a literal `reports/`. Handing the current function raw Notes is slightly too loose:

```
'applied via https://acme.example/jobs/7'          ->  []     ✔ absolute URLs already rejected
'req JR-10423; also mentioned #99 in passing'      ->  []     ✔
'#154 Sr PM is already live in the same ATS'       ->  []     ✔
'see [9](../notes/9-thing.md)'                     ->  [9]    ✘ not a report
```

That last one is harmless for the Report *cell*, which by contract holds a report link, but scanning prose needs the `reports/` requirement merge-tracker already applies.

## Proposed fix

Give `extractTrackerReportNumbers` an optional `notesStr` fallback that requires a `reports/` path segment, and pass `r.notes` at the four `set-status` sites. Both readers then share one function rather than each carrying its own idea of where a report link lives.

Converging `merge-tracker`'s own `extractReportNum` onto it is the obvious follow-up, but its report-cell branch is looser (`/\[(\d+)\]/`, no target validation) and unifying that is a behaviour change on a separate path — better done on its own than mixed in here.

abankar1:
Fixes #3075.

8668ac1 taught `merge-tracker` that a customized tracker with no dedicated Report column keeps its report link in **Notes** prose. `set-status.mjs` — the canonical write path — was not taught the same thing, so on exactly the layout that fix added support for, `--report N` fails on a row merge-tracker links happily.

```
| # | Date | Company | Role | Score | Status | Materials | Apply Link | Follow-up | Notes |
| 7 | 2026-06-01 | Acme | Staff Eng | 4.2/5 | Applied | CV v3 | https://acme.example/jobs/7 | 2026-06-14 | strong fit; report [12](reports/012-acme-2026-06-01.md) |
```

```
$ node set-status.mjs --report 12 Interview --dry-run
❌ No tracker row links report #12. (Report IDs and tracker row IDs differ — use --row N to select by tracker #.)

$ node set-status.mjs --row 7 Interview --dry-run
✅ #7 Acme — Staff Eng: would set Applied → Interview
```

The row is fine; only the `--report` lookup cannot see it. The error is misleading on top of being wrong — it blames the row/report counter divergence, when the real reason is that this layout keeps its link somewhere `set-status` never reads.

## Not a missing capability, an unpassed argument

`extractTrackerReportNumbers` already finds it when handed the text. All three call sites now pass the Notes cell, so both readers share one function instead of each carrying its own idea of where a report link lives.

## The part worth reviewing

Notes is **free-form prose**, so the fallback has to be scoped — which is why merge-tracker's own pattern requires a literal `reports/`. Handing the current matcher raw Notes is slightly too loose. Measured:

```
'applied via https://acme.example/jobs/7'      -> []    ✔ absolute URLs already rejected
'req JR-10423; also mentioned #99 in passing'  -> []    ✔
'#154 Sr PM is already live in the same ATS'   -> []    ✔
'see [9](../notes/9-thing.md)'                 -> [9]   ✘ not a report
```

That last case is harmless for the Report *cell*, which holds a report link by contract, and wrong for prose. So the new notes scan requires a `reports/` path segment: an unrelated markdown link, or a posting URL in the same sentence, cannot claim to be a report number.

Notes is a **fallback only** — a Report cell that yields a number is never overridden. An *empty* Report cell also falls through, so a customized tracker behaves the same whether its Report column is absent or merely blank.

## Verification

Three of the four new assertions pin what must **not** change, because a lookup that matched more freely would satisfy the fix while breaking the guard rails:

- an unlinked report number is still not found
- a non-report markdown link in Notes does not claim a number
- a populated Report cell still wins over Notes

Checked for discrimination: with `tracker-parse.mjs` and `set-status.mjs` reverted, **exactly the custom-layout assertion fails** and the three guards pass.

`set-status-tests.mjs`: 96 passed / 0 failed. `node test-all.mjs --quick`: 4318 pass / 0 fail (3 warnings pre-existing and environmental).

## Follow-up, deliberately not here

Converging `merge-tracker`'s own `extractReportNum` onto the shared function is the obvious next step. Its report-cell branch is looser (`/\[(\d+)\]/`, no target validation), so unifying that is a behaviour change on a separate path — better on its own than mixed into this.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved report detection when report links are stored in tracker notes instead of a dedicated report field.
  * Ignored unrelated external URLs and markdown links during report matching.
  * Preserved dedicated report-field precedence when conflicting links appear in notes.
  * Prevented unlinked report numbers from being incorrectly resolved.

* **Tests**
  * Added regression coverage for report resolution, fallback behavior, link filtering, and precedence rules.

### DIFF
--- a/set-status-tests.mjs
+++ b/set-status-tests.mjs
@@ -1351,5 +1351,68 @@ for (const bad of ['correction', 'backfill', 'cell-edit', 'nonsense']) {
   }
 }
 
+// ── #3075: --report resolves a report link kept in Notes ────────────────────
+// merge-tracker learned this layout in 8668ac1 — a customized tracker with no
+// dedicated Report column keeps the link in Notes prose. set-status read only
+// the Report cell, so --report could not find a row merge-tracker linked
+// happily, and the error even advised --row for the wrong reason.
+{
+  const CUSTOM = `# Applications Tracker
+
+| # | Date | Company | Role | Score | Status | Materials | Apply Link | Follow-up | Notes |
+|---|------|---------|------|-------|--------|-----------|------------|-----------|-------|
+| 7 | 2026-06-01 | Acme | Staff Eng | 4.2/5 | Applied | CV v3 | https://acme.example/jobs/7 | 2026-06-14 | strong fit; report [12](reports/012-acme-2026-06-01.md) |
+`;
+
+  {
+    const sb = makeSandbox(CUSTOM);
+    try {
+      const r = runSetStatus(['--report', '12', 'Interview', '--dry-run'], sb);
+      if (r.code === 0) pass('#3075: --report resolves a report link kept in the Notes cell');
+      else fail(`#3075: --report 12 failed on a custom layout (code=${r.code}) ${(r.stderr || r.stdout).split('\n')[0]}`);
+    } finally { rmSync(sb.dir, { recursive: true, force: true }); }
+  }
+
+  // MUST NOT CHANGE: a report nobody links is still not found. Without this,
+  // making the lookup match anything would pass the assertion above.
+  {
+    const sb = makeSandbox(CUSTOM);
+    try {
+      const r = runSetStatus(['--report', '99', 'Interview', '--dry-run'], sb);
+      if (r.code !== 0) pass('#3075: an unlinked report number is still not found');
+      else fail('#3075: --report 99 resolved a row nothing links');
+    } finally { rmSync(sb.dir, { recursive: true, force: true }); }
+  }
+
+  // MUST NOT CHANGE: Notes is prose, so the fallback is scoped to reports/.
+  // A job-posting URL and an unrelated markdown link both sit in Notes here.
+  {
+    const NOISY = CUSTOM.replace(
+      'strong fit; report [12](reports/012-acme-2026-06-01.md)',
+      'applied via https://acme.example/jobs/9 — see [9](../notes/9-thing.md)',
+    );
+    const sb = makeSandbox(NOISY);
+    try {
+      const r = runSetStatus(['--report', '9', 'Interview', '--dry-run'], sb);
+      if (r.code !== 0) pass('#3075: a non-report link in Notes does not claim a report number');
+      else fail('#3075: --report 9 matched a non-report markdown link in Notes');
+    } finally { rmSync(sb.dir, { recursive: true, force: true }); }
+  }
+
+  // MUST NOT CHANGE: a real Report cell still wins over anything in Notes.
+  {
+    const CONFLICT = TRACKER_9.replace(
+      '| 2 | 2026-06-02 | Globex | Platform Engineer | 4.0/5 | Evaluated | ✅ | [2](../reports/002-globex-2026-06-02.md) | — |',
+      '| 2 | 2026-06-02 | Globex | Platform Engineer | 4.0/5 | Evaluated | ✅ | [2](../reports/002-globex-2026-06-02.md) | also [77](reports/077-other.md) |',
+    );
+    const sb = makeSandbox(CONFLICT);
+    try {
+      const r = runSetStatus(['--report', '77', 'Interview', '--dry-run'], sb);
+      if (r.code !== 0) pass('#3075: Notes is a fallback only — a populated Report cell wins');
+      else fail('#3075: a Notes link overrode a populated Report cell');
+    } finally { rmSync(sb.dir, { recursive: true, force: true }); }
+  }
+}
+
 console.log(`\n${passed} passed, ${failed} failed`);
 process.exit(failed > 0 ? 1 : 0);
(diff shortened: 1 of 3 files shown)


santifer:
Direction is right, @abankar1, and the three MUST-NOT-CHANGE cases are the part I'd defend hardest: scoping the Notes scan to `reports/` is exactly what keeps a posting URL or a `[9](../notes/9-thing.md)` from claiming a report number. One gap before I merge it, and it's the same asymmetry this PR exists to remove, one layer down.

The Report-cell scan deliberately yields **two** numbers per link: the numeric label and the number in the path (`tracker-parse.mjs:327-332`), so a link whose label has drifted from its filename stays findable by either. I checked it rather than assumed:

```
extractTrackerReportNumbers("[12](reports/013-acme-2026-06-01.md)")  ->  [12, 13]
extractTrackerReportNumbers("[12](reports/012-acme-2026-06-01.md)")  ->  [12]
```

`scanNotesForReportNumbers` adds only the path number, so with the customized layout:

```
| ... | strong fit; report [12](reports/013-acme-2026-06-01.md) |
```

`--report 13` resolves the row and `--report 12` reports "No tracker row links report #12", while that identical link sitting in a Report cell answers to both. The drift shape isn't hypothetical: the Report branch carries that label handling on purpose, precisely because labels and filenames do come apart.

The ask is small: after the target has been scoped to `reports/`, add the numeric link label as well, mirroring the Report-cell branch. A fifth case pinning the drift shape would keep it from coming back.

Two things I'd like to name that are NOT asks, so they don't get lost:

- The optional second parameter is the right call. Every other caller keeps working unchanged, which is what made this reviewable in one sitting.
- `merge-tracker.mjs` has its own `extractReportNum(reportStr, notesStr)` doing this same job, and your comment says so plainly instead of quietly re-deriving it. Unifying the two is a separate PR, not this one, and I'd rather land yours first.

Happy to merge once the label is in.

santifer:
And the set-status half lands too, @abankar1. The three MUST-NOT-CHANGE assertions with the discrimination check (revert the fix and exactly the new assertion fails while the guards keep passing) pin the behavior, not the happy path. Making the Notes scan deliberately stricter than the Report-cell scan by requiring the reports/ path segment is the right asymmetry. Please do open that follow-up issue for the merge-tracker extractReportNum convergence so it does not evaporate. Merged! 🚀

PR #3081, opened 2026-08-19
fix(web-pipeline): first_seen was stamped with the UTC day, not the local one — the web/ site #3071 doesn't reach (#3070)

Zoubeir23:
#3070/#3071 catalog and fix six core `.mjs` call sites that resolve "today" as the UTC calendar day instead of the local one — each one *gates a decision* rather than merely stamping a filename, per that issue's own scope line.

There is a seventh site, in `web/`, that the audit didn't reach: `web/src/lib/core/pipeline.ts`'s "Add to pipeline" writer (`addOffersToPipeline`). It spawns a short-lived child process that calls the core's own `appendToScanHistory(offers, date, "added")`, and computes `date` itself with `new Date().toISOString().slice(0, 10)`.

**Why #3071 doesn't cover this.** #3071 fixes `scan.mjs`'s *default* for this same parameter. `pipeline.ts` always passes an explicit `date` argument, which overrides the default — so the core fix lands and this call site is still wrong.

**Impact.** West of Greenwich, an evening "Add to pipeline" click in the web dashboard stamps `data/scan-history.tsv`'s `first_seen` one calendar day ahead of the user's own clock. That date then drives `scan.mjs`'s `shouldDedupScanHistoryRow` recheck/cooldown window for that row — same mechanism, same consequence #3070 describes for the core sites, just reached from the web UI instead of a CLI scan.

**Verified by execution**, not reasoning — frozen clock at `2026-08-18T01:30:00Z` under `TZ=America/New_York` (same instant #3070/#3071 use):

```
before: new Date().toISOString().slice(0, 10)  -> 2026-08-18  (UTC day — tomorrow, from here)
after:  localToday()                           -> 2026-08-17  (correct local day)
```

**The fix** imports `lib/local-today.mjs`'s `localToday()` into the spawned child the same way `scanUrl` is already injected — an absolute `file://` URL built from `careerOpsRoot()`, since the child resolves imports independently of any relative path — and uses it in place of the UTC computation. Three-commit shape: wire the import, flip the one line, then the test.

**Test.** `pipeline.ts` is TypeScript, imports via the `@/` alias (unresolvable by plain `node --test`), and computes the date inside a template-literal string executed in a *separate* spawned process — so, same reasoning already documented in `tests/lib/core-writer-await.test.mjs`, it can't be imported and exercised directly from a test. `web/tests/lib/pipeline-local-today.test.mjs` follows that same source-text-extraction convention: it isolates the child-process template, asserts the legacy UTC pattern is gone, and asserts `localToday()` is both called and imported via the expected absolute-URL pattern. Confirmed the extractor's UTC-pattern regex matches the pre-fix line, so it would have caught this regression.

Full suite: `node test-all.mjs` — 4321 passed / 0 failed (2 pre-existing, unrelated warnings: the Go dashboard build skip and a PII-heuristic false positive on santifer.io in untouched Go source). `web`: `npm test` 269/269, `npm run typecheck` clean.

## Test plan
- [x] `node test-all.mjs` — 4321 passed, 0 failed
- [x] `cd web && npm test` — 269 passed, 0 failed
- [x] Fix verified by execution under a frozen clock (`TZ=America/New_York`, instant `2026-08-18T01:30:00Z`) before writing the regression test
- [x] New test confirmed to catch the pre-fix line (regex matches the old source text)

## Summary by CodeRabbit

* **Bug Fixes**
  * Scan history now records the correct local calendar date instead of potentially using the previous or next day due to UTC conversion.

* **Tests**
  * Added regression coverage to verify local-date handling remains correct when recording scan history.

### DIFF
--- a/web/src/lib/core/pipeline.ts
+++ b/web/src/lib/core/pipeline.ts
@@ -51,7 +51,10 @@ process.stdin.on("data", (d) => { input += d; });
 process.stdin.on("end", async () => {
   try {
     const offers = JSON.parse(input);
-    const date = new Date().toISOString().slice(0, 10);
+    // LOCAL calendar day, not the UTC one — west of Greenwich, an evening
+    // add would otherwise stamp scan-history.tsv's first_seen a day ahead,
+    // opening scan.mjs's recheck/cooldown gate a day late for this row (#3070).
+    const date = localToday();
     await appendToPipeline(offers);
     await appendToScanHistory(offers, date, "added");
     process.stdout.write(JSON.stringify({ added: offers.length }));
--- a/web/src/lib/core/pipeline.ts
+++ b/web/src/lib/core/pipeline.ts
@@ -41,8 +41,10 @@ export function addOffersToPipeline(offers: DiscoveredOffer[]): Promise<AddResul
   }
 
   const scanUrl = pathToFileURL(rootScript("scan")).href;
+  const localTodayUrl = pathToFileURL(path.join(careerOpsRoot(), "lib", "local-today.mjs")).href;
   const code = `
 import { appendToPipeline, appendToScanHistory } from ${JSON.stringify(scanUrl)};
+import { localToday } from ${JSON.stringify(localTodayUrl)};
 let input = "";
 process.stdin.setEncoding("utf8");
 process.stdin.on("data", (d) => { input += d; });
--- a/web/src/lib/core/pipeline.ts
+++ b/web/src/lib/core/pipeline.ts
@@ -1,5 +1,6 @@
 import { spawn } from "node:child_process";
 import fs from "node:fs";
+import path from "node:path";
 import { pathToFileURL } from "node:url";
 import { careerOpsRoot, rootScript } from "@/lib/career-ops";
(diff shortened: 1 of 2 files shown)


santifer (approved):
Correct, and it's the right cure: importing the core's localToday() rather than reimplementing the local-day calc — I verified lib/local-today.mjs exports it on main, so the runtime import resolves. This is the seventh site of the #3070/#3071 UTC-vs-local family, and the one #3071 couldn't reach because pipeline.ts passes an explicit date that overrides scan.mjs's fixed default. The source-shape test is the right level here (the date is computed inside a spawned template-literal child, same untestable-by-import case as core-writer-await), and you note the execution-verified frozen-clock check. Merging.

santifer:
Thanks, @Zoubeir23 — the seventh site of the UTC-vs-local family, and the one #3071 couldn't reach because pipeline.ts passes an explicit date that overrides scan.mjs's fixed default. Importing the core's localToday() instead of reimplementing the local-day calc is the right cure — one source of truth, not an eighth copy to drift. Merged 🚀

PR #3082, opened 2026-08-19
fix(web-explore): canon() discarded the entire query string, silently collapsing distinct postings into one dedup key

Zoubeir23:
`web/src/lib/explore-ai.ts`'s `canon()` — the URL-dedup key for the AI \"Discover\" feature — keyed on `host + pathname` only, unconditionally dropping the query string:

```ts
export function canon(u: string): string {
  try {
    const x = new URL(u);
    return (x.host + x.pathname).toLowerCase().replace(/\/$/, "");
  } catch { ... }
}
```

`canon()` builds the server-side "already known" `Set` (`web/src/lib/core/discover.ts`'s `assembleDedupContext`, from `data/scan-history.tsv` + inbox + applications) **and** the client-side per-offer key the AI stream parser checks each candidate against (`makeAiStreamParser`'s `feed()`). Any host+path collision — real or coincidental — silently drops the offer: `if (seen.has(key) || known.has(key)) continue;`, no log, no visible "duplicate" marker.

**The concrete failure case.** Some ATS boards embed the job identity in the query string rather than the path — Greenhouse's `?gh_jid=` on corporate-hosted boards is the one already documented in this repo (see `url-key.mjs`'s own comment). Two genuinely different openings at the same employer, same `careers_url` page, different `gh_jid`:

```
https://boards.greenhouse.io/acme/jobs/apply?gh_jid=4471829005
https://boards.greenhouse.io/acme/jobs/apply?gh_jid=5501203417
```

both keyed to `boards.greenhouse.io/acme/jobs/apply`. Once either has ever appeared in `scan-history.tsv`/inbox/applications, **every subsequent distinct opening at that host+path is silently absent from AI Discover results** — not shown as a duplicate, just gone.

**Why this is a fresh regression of an already-fixed bug class.** The core already solved this exact problem for the tracker: `url-key.mjs`'s `normalizeUrl` exists specifically because over-normalizing a posting URL causes silent data loss, and its docstring states the design goal directly — *"UNDER-STRIP ON PURPOSE... over-normalizing collapses two genuinely different postings into one key → a SILENT merge / data loss."* It strips only a narrow tracking-param denylist and deliberately keeps functional params like `gh_jid`. `explore-ai.ts`'s `canon()` reimplemented URL keying independently on the web side and reintroduced exactly the failure mode `url-key.mjs` was written to prevent.

**Why not just import `url-key.mjs` directly.** `canon()` runs in the browser (`explore-provider.tsx` is a client component; the file's own header says "Client-safe (NO node imports)"), and the SAME function has to run server-side too (`discover.ts`, building the known-URL `Set`) — the client's per-offer key and the server-built `Set`'s keys must be byte-identical or `known.has(key)` silently stops matching anything. `url-key.mjs` itself has zero Node dependencies (only the global `URL`), so this repo's own established pattern for exactly this constraint — see `normalize-text-key.mjs` next to `text-key.ts` (#2369/#2666) — is a parity-tested static mirror under `web/src/lib/core/`. `web/src/lib/core/url-key.mjs` is that mirror here, body byte-for-byte aligned with the root `url-key.mjs`, guarded by a parity test that imports both side-by-side.

**Secondary fix.** `normalizeUrl` returns `''` for an unparseable-but-regex-passing URL (explicit "NO KEY IS NOT A KEY" contract — a placeholder must never collide with another placeholder). The stream parser's dedup check didn't have this guard yet (`canon()` never used to return `''`), so two such offers would have been wrongly treated as duplicates of each other. Fixed alongside: `if (key && (seen.has(key) || known.has(key))) continue;`.

**Verified by execution**, not reasoning — both from the parity test (imports the real root `url-key.mjs` and the mirror side-by-side) and from exercising the real `makeAiStreamParser`/`canon()` end-to-end (`explore-ai.ts` has no `@/` alias and no Node-only deps, so unlike `web/src/lib/core/pipeline.ts` it can be imported directly in a test rather than needing source-text extraction):

```
core normalizeUrl:  jobA -> .../apply?gh_jid=4471829005
                     jobB -> .../apply?gh_jid=5501203417   (different key — correct)

pre-fix canon():    jobA -> boards.greenhouse.io/acme/jobs/apply
                     jobB -> boards.greenhouse.io/acme/jobs/apply   (SAME key — the bug)
```

**Severity:** high — silent, user-invisible data loss in a user-facing discovery feature. Query-string job ids are not a corner case; they're how one common ATS integration identifies postings.

**Not already tracked** — searched issues/PRs for `canon`, `gh_jid`, `dedup discover`: no hits. #2369/#2666 are the historical issues that built `normalizeUrl`'s asymmetric-strip design and the `normalize-text-key.mjs` mirror pattern in the first place; this is a fresh instance of the same class on the untouched `explore-ai.ts` path.

Five-commit shape: add the mirror → parity-test it → wire `canon()` to it → fix the empty-key guard → exercise the real stream parser end-to-end.

## Test plan
- [x] `node test-all.mjs` — 4321 passed, 0 failed (2 pre-existing unrelated warnings: Go dashboard build skip, PII-heuristic false positive on santifer.io in untouched Go source)
- [x] `cd web && npm test` — 272 passed, 0 failed
- [x] Parity test imports the real root `url-key.mjs` and the new mirror side-by-side and asserts they agree, including the concrete `gh_jid` case
- [x] `explore-ai-dedup.test.mjs` exercises the real `makeAiStreamParser`/`canon()`: two distinct postings sharing host+path are both kept, a true duplicate still dedups, and `knownUrls` filters by exact key rather than host+path

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved posting URL normalization for consistent HTTPS handling, hostname casing, trailing slashes, and tracking parameters.
  * Preserved meaningful query parameters so distinct postings are no longer incorrectly merged.
  * Invalid or unsupported URLs are safely ignored during deduplication.

* **Tests**
  * Added coverage for URL normalization, tracking removal, posting uniqueness, and invalid inputs.

### DIFF
--- a/web/src/lib/core/url-key.mjs
+++ b/web/src/lib/core/url-key.mjs
@@ -1,0 +1,79 @@
+/**
+ * Canonical posting-URL key — algorithm mirror of the core's `normalizeUrl`
+ * in url-key.mjs (root).
+ *
+ * Plain .mjs (same pattern as normalize-text-key.mjs) so node:test and client
+ * bundles can import it without a TS runner or Node-only deps — this file has
+ * none, same as the core it mirrors (only the global `URL`).
+ *
+ * WHY A COPY EXISTS AT ALL: the live core lives in the user's career-ops
+ * checkout and is resolved at runtime via careerOpsRoot() — unavailable to the
+ * client bundle. Unlike normalize-text-key.mjs's split (a Node-only live
+ * loader server-side, this mirror client-side), the URL key here is used to
+ * build ONE Set server-side (assembleDedupContext, from data/scan-history.tsv)
+ * that the CLIENT then does membership lookups against on each AI-streamed
+ * offer (explore-ai.ts's canon()). Server and client MUST key with the exact
+ * same function or `known.has(key)` silently stops matching anything — so both
+ * sides import this one mirror rather than the server preferring a live-loaded
+ * copy that could drift from what the client bundle was built with.
+ *
+ * Keep the body byte-for-byte aligned with url-key.mjs `normalizeUrl`. The
+ * parity test in tests/lib/url-key.test.mjs fails the build if they drift.
+ *
+ * UNDER-STRIP ON PURPOSE (see the root file's docstring for the full RFC 3986
+ * rationale). Never add path/query lowercasing or a broader strip list here:
+ * that is exactly the over-normalization that collapsed two different
+ * Greenhouse postings (same host+path, distinct `?gh_jid=`) into one dedup
+ * key and silently dropped every opening after the first at that employer —
+ * the bug this mirror exists to stop reintroducing.
+ */
+
+// Query params that identify a click/campaign, never the posting itself. Keep
+// this list literal and board-specific, identical to the core's denylist.
+const TRACKING_PARAMS = [
+  /^utm_/i, /^gh_src$/i, /^fbclid$/i, /^gclid$/i,
+  /^mc_cid$/i, /^mc_eid$/i, /^igshid$/i, /^_hsenc$/i, /^_hsmi$/i, /^trk$/i, /^trackingid$/i,
+];
+
+/**
+ * Reduce a posting URL to a stable comparison key.
+ *
+ * @param {string} raw - A posting URL (or any string).
+ * @returns {string} A normalized key, or '' when there is nothing to key on.
+ *   '' means NO KEY — callers must treat it as unknown, never as a value that
+ *   can match another ''.
+ */
+export function normalizeUrl(raw) {
+  if (typeof raw !== "string") return "";
+  const s = raw.trim();
+  if (!s) return "";
+
+  let u;
+  try {
+    u = new URL(s);
+  } catch {
+    return "";
+  }
+
+  if (u.protocol !== "http:" && u.protocol !== "https:") return "";
+
+  u.protocol = "https:";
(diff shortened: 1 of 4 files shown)


santifer (approved):
The strongest of the batch. canon() keying on host+pathname only collapsed distinct Greenhouse postings (same path, different ?gh_jid=) into one dedup key and silently dropped every opening after the first at that employer. The fix is right AND it installs the guard against the class reappearing: I verified the parity test actually imports BOTH webKey (the mirror) and coreKey (root url-key.mjs, dynamic import) and asserts webKey(x)===coreKey(x) across cases — so the mirror can't drift from core without failing the build. That's assert-agreement-by-executing-both, exactly the cure for a client/server key that must match or known.has() silently stops matching. Under-strip-on-purpose documented. Merging.

santifer:
Thanks, @Zoubeir23 — the strongest of the batch. Keying dedup on host+pathname only collapsed distinct Greenhouse openings (same path, different ?gh_jid=) into one key and silently dropped every posting after the first at an employer. What earns the merge isn't just the fix — it's the parity test that imports BOTH the mirror and the core normalizeUrl and asserts they agree across cases, so the copy can't drift without failing the build. A client/server key that must match or known.has() stops matching silently is exactly where a drift guard belongs. Merged 🚀

PR #3085, opened 2026-08-20
fix(funnel-velocity): validate CLI flags

### LINKED ISSUE #2853: funnel-velocity.mjs has no --help and silently ignores mistyped flags

MUKUL-PRASAD-SIGH:
## Summary

- add usage output for `--help` and `-h`
- reject unknown flags before report processing
- validate the value-taking `--benchmarks` flag
- use the shared `flagValue` helper, supporting both argument forms
- preserve normal report output byte-for-byte
- add focused CLI regression tests

## Testing

- `node --test tests/funnel-velocity-flags.test.mjs` — 8 passed
- `node funnel-velocity.mjs --self-test` — passed
- `node test-all.mjs` — 4,312 passed, 0 failed

Fixes #2853

## Summary by CodeRabbit

* **New Features**
  * Centralized command-line option handling for more consistent behavior.
  * Added updated usage guidance for supported options.
  * Added validation for unknown options and benchmark arguments.

* **Bug Fixes**
  * Invalid command-line input is now rejected before execution.
  * Preserved JSON output when running without a tracker.

* **Tests**
  * Added coverage for help handling, argument validation, successful execution, and temporary test environment cleanup.

### DIFF
--- a/funnel-velocity.mjs
+++ b/funnel-velocity.mjs
@@ -42,13 +42,20 @@ import { computeFunnel, computeTrackerStats } from './stats.mjs';
 import { resolveColumns, parseTrackerRow } from './tracker-parse.mjs';
 import { resolveTrackerPath, loadCanonicalStates, resolveCanonicalState } from './tracker-utils.mjs';
 import { parseAppliedDate, normalizeStatus } from './followup-cadence.mjs';
+import { flagValue, validateFlags } from './lib/cli-flags.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
 const STATES_FILE = join(CAREER_OPS, 'templates/states.yml');
 
-const args = process.argv.slice(2);
-const summaryMode = args.includes('--summary');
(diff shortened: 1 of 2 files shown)


santifer:
Welcome to career-ops, @MUKUL-PRASAD-SIGH! Double win in one small diff: mistyped flags now fail loudly, and your migration deleted a private flagValue copy in favor of the shared one — exactly the direction the whole CLI family is moving. Merged — your first contribution here 🚀

PR #3057, opened 2026-08-18
fix(cli): validate upskill.mjs flags via lib/cli-flags.mjs

obielin:
## What does this PR do?
upskill.mjs parsed flags positionally, so a mistyped flag (--sumary, --min-report) was silently ignored and the script ran on its aggregate defaults with exit 0. This delegates flag validation to validateFlags in lib/cli-flags.mjs, matching doctor.mjs (#2874), so it rejects unrecognised flags before reading any tracker or report file.

## Related issue
Closes #3002

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation / translation
- [ ] Refactor (no behavior change)

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/santifer/career-ops/blob/main/CONTRIBUTING.md)
- [x] If this is a new feature or architecture change, I opened an issue first (bug fixes, providers, docs & translations are exempt — send those straight in)
- [x] My PR does not include personal data (CV, email, real names, scan results, or pipeline data)
- [x] I ran `node test-all.mjs` and all tests pass
- [x] My changes respect the [Data Contract](https://github.com/santifer/career-ops/blob/main/DATA_CONTRACT.md) (no modifications to user-layer files)
- [x] My changes align with the [project roadmap](https://github.com/santifer/career-ops/discussions/156)

---
Questions? [Join the Discord](https://discord.gg/8pRpHETxa4) for faster feedback.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added command-line flag validation to reject unknown or malformed options before processing commands.
  * Corrected handling of the `--min-report` option.
  * Updated validation coverage to include `upskill.mjs`.

PR #3092, opened 2026-08-20
fix(cli): close the missing-operand hole for value-taking flags (#3087)

### LINKED ISSUE #3087: fix(cli): a value-taking flag with no value is silently dropped — doctor.mjs diagnoses a directory named "--json"
Found while working #2977. Same defect class through a different door, and it
is not archive-posting-specific — it is the shared helper's default.

`validateFlags(args, KNOWN, USAGE, { valueFlags })` rejects an *unrecognized*
flag, but a **recognized** value-taking flag with no value falls straight
through. `requireOperand` closes it and defaults to `false`. Eleven of the 15 call sites
pass `valueFlags`, and exactly one — `batch-tailor.mjs` — opts in. Scripts with
no guard of their own then run on partial arguments at exit 0 — the same
silent-wrong-answer shape #2775 exists to end.

### Confirmed, on `8f3bcf3`

**`doctor.mjs` — the sharpest case, because it is the onboarding entrypoint.**
The next flag is swallowed as the path, and the doctor diagnoses a directory
that does not exist:

```console
$ node doctor.mjs --target --json
{"onboardingNeeded":true,"missing":[...,"modes/_profile.md",...],"plugins":[],...}
                                                                  ^^^^^^^^^^^^
$ node doctor.mjs --json          # the same checkout, actually diagnosed
{"onboardingNeeded":true,"missing":[...],"plugins":[{"id":"apify",...},...],...}
```

Exit 0 both times. The first run reports on a path literally named `--json`, so
a user following onboarding is told to create files that already exist. A
trailing `--cli` is the milder version: exit 0, silently falls back to the
default CLI.

**`archive-posting.mjs`** — a capture named after a flag:

```console
$ node archive-posting.mjs --dry-run "https://…/eng" --company --pipeline
   Company: --pipeline
    local:jds/2026-08-20_pipeline_job.pdf
```

`--company` swallows `--pipeline`, so pipeline mode never turns on *and* the
company slug becomes the flag. A trailing `--company` is the milder version:
`Company: unknown`, exit 0.

### Not affected

`rejection-latency.mjs --summary --tracker` exits **2** with `--tracker requires
a value.` — its own `hasFlag`/`flagValue` pairing already catches this, as does
`scan.mjs`'s `requireValue()`. That is the pattern the fix should generalize,
not replace: `lib/cli-flags.mjs` already argues (at the `requireOperand` block)
that callers with a better message should keep it.

### What to do

Audit the 10 callers passing `valueFlags` without `requireOperand` and split
them in two:

1. **Has its own guard** (`scan.mjs`, `rejection-latency.mjs`, …) — leave alone,
   the tailored message says more.
2. **Has none** (`doctor.mjs`, `archive-posting.mjs`, and whichever of
   `contacts.mjs` / `weekly-digest.mjs` / `scan-ats-full.mjs` /
   `check-table-freshness.mjs` / `process-quality.mjs` / `detect-reposts.mjs`
   the audit confirms) — pass `requireOperand: true`.

Worth considering whether `requireOperand` should default to **true**, with the
handful of callers that have a better message opting *out*. Safe-by-default
matches the direction of #2775, #2919 and #2961; the current default means every
new caller re-opens the hole unless the author knows to close it.

### Acceptance

- `node doctor.mjs --target --json` exits non-zero instead of diagnosing a
  directory named `--json`.
- `node archive-posting.mjs <url> --company --pipeline` exits non-zero instead of
  writing `…_pipeline_job.pdf`.
- Callers with their own message keep it — `rejection-latency.mjs --summary
  --tracker` still exits 2 with `--tracker requires a value.`
- Coverage in `tests/cli-flag-validation.test.mjs`, alongside the
  unrecognized-flag cases.
- `node test-all.mjs` green.

Searched open and closed issues; #2977 is the closest and covers only the
*unrecognized*-flag half. Happy to take this one if it is wanted.

kovalancik06-max:
## Summary

Fixes #3087. Completes the audit the issue asked for: a value-taking CLI flag that's *recognized* but given with no value (nothing follows it, or the next token is itself another `--flag`) silently fell through instead of erroring — `validateFlags`'s unrecognized-flag check doesn't catch this, because the flag itself is spelled correctly.

The sharpest case, from the issue: `doctor.mjs`'s `--target`/`--cli` resolution (`argv.indexOf('--target')` + `argv[idx+1]`) has no adjacency check, so `node doctor.mjs --target --json` reads `--json` as the target path and silently diagnoses a directory that doesn't exist, at exit 0 — telling a user following onboarding to create files that already exist.

`lib/cli-flags.mjs`'s `validateFlags()` already had an opt-in `requireOperand` option (added for `batch-tailor.mjs`, per its own comment: *"This script has nothing more specific to say about a missing operand than the shared message, which is exactly when opting in is right"*), but most callers with value flags never passed it.

## Audit

Went through every caller of `validateFlags` that passes `valueFlags`:

**Needed `requireOperand: true` (no guard against a missing operand):**
- `doctor.mjs` — `--target`/`--cli`, the issue's headline example.
- `detect-reposts.mjs` — `--window`: swallowed value fails `parseInt`, falls back to the 90-day default silently.
- `process-quality.mjs` — `--file`/`--min-threshold`: same silent-fallback shape.
- `weekly-digest.mjs` — `--from`/`--to` happen to reject a swallowed flag indirectly (fails `isValidDateStr`), but `--dir` has no validation at all and would silently point the digest at a directory literally named `--from`.

**Already correct, left alone** (matching the issue's own "has its own guard, leave alone" criterion):
- `contacts.mjs` — `--vcf`'s value is optional by design (`node contacts.mjs --vcf [path]`) and already checks `.startsWith('--')` to fall back to `null` rather than swallow a flag.
- `scan-ats-full.mjs` — has its own adjacency-checked `valueOf()` helper.
- `check-table-freshness.mjs` — already parses and validates both `--max-age-months` and `--today` with `process.exit(1)` and a specific message.
- `rejection-latency.mjs`, `scan.mjs`, `batch-tailor.mjs` — the issue's own "not affected"/"already opts in" list, unchanged.

**Not applicable** — `invite-match.mjs`, `classify-tier.mjs`, `reply-watch.mjs`, `dedup-tracker.mjs` don't take any value flags.

**`archive-posting.mjs`** doesn't use `validateFlags` at all — it hand-rolls its own argv loop. Its `--company`/`--role` handling (`arg === '--company' && args[i + 1]`) has the identical defect through a different door: any truthy next token, including another flag, gets accepted as the value. `node archive-posting.mjs <url> --company --pipeline` used to set the company slug to `"--pipeline"` and silently leave pipeline mode off. Added the same "next token must not start with `--`" adjacency check directly, matching the convention `cli-flags.mjs` already documents.

## Test plan

- [x] Reproduced and fixed both headline examples from the issue directly:
  ```
  $ node doctor.mjs --target --json
  Error: --target requires a value
  $ node archive-posting.mjs https://example.com/job --company --pipeline
  --company requires a value.
  ```
- [x] Confirmed the valid cases still work (`doctor.mjs --json`, `detect-reposts.mjs --window 60 --summary`) and that callers with their own message still keep it (`rejection-latency.mjs --summary --tracker` still exits 2 with `--tracker requires a value.`, unchanged).
- [x] Found and fixed a stale assertion in `detect-reposts.test.mjs` that encoded the *old* behavior as correct — `--window` with no value was expected to silently fall back to the 90-day default. It now expects the exit-1 usage error, which is what this PR intentionally changes.
- [x] Added 7 new cases to `tests/cli-flag-validation.test.mjs` covering all five fixed spots, following the file's existing hermetic/no-fixture pattern for cases that fail inside `validateFlags` before any `data/` access.
- [x] `node --test tests/cli-flag-validation.test.mjs` → 24/24 passed.
- [x] `node test-all.mjs` → **4320 passed, 0 failed** (3 pre-existing warnings, unrelated to these files — present before this change too).

## Acceptance criteria from the issue

- [x] `node doctor.mjs --target --json` exits non-zero instead of diagnosing a directory named `--json`.
- [x] `node archive-posting.mjs <url> --company --pipeline` exits non-zero instead of writing `…_pipeline_job.pdf`.
- [x] Callers with their own message keep it (`rejection-latency.mjs --summary --tracker` still exits 2 with `--tracker requires a value.`).
- [x] Coverage in `tests/cli-flag-validation.test.mjs`, alongside the unrecognized-flag cases.
- [x] `node test-all.mjs` green.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved command-line validation across archive, repost detection, diagnostics, quality processing, and digest tools.
  - Value-based options now require an explicit value and reject another option used accidentally as that value.
  - Prevented missing inputs from being silently ignored or replaced with default settings.
  - Error messages now clearly identify options that require a value.
- **Tests**
  - Added regression coverage for missing values across supported command-line tools.
  - Updated expectations to confirm invalid commands exit with an error.

### DIFF
--- a/archive-posting.mjs
+++ b/archive-posting.mjs
@@ -101,11 +101,24 @@ function parseCliArgs(args) {
       dryRun = true;
     } else if (arg.startsWith('--company=')) {
       overrideCompany = arg.slice('--company='.length).trim();
-    } else if (arg === '--company' && args[i + 1]) {
+    } else if (arg === '--company') {
+      // The old `args[i + 1]` truthiness check consumed the NEXT flag as the
+      // company name whenever --company was given with no value (e.g.
+      // `--company --pipeline` set overrideCompany to "--pipeline" and left
+      // pipeline mode off, silently, at exit 0 (#3087)). A value that starts
+      // with `--` is another flag, not a company name.
+      if (args[i + 1] === undefined || args[i + 1].startsWith('--')) {
+        console.error('--company requires a value.');
+        process.exit(1);
+      }
       overrideCompany = args[++i].trim();
     } else if (arg.startsWith('--role=')) {
       overrideRole = arg.slice('--role='.length).trim();
-    } else if (arg === '--role' && args[i + 1]) {
+    } else if (arg === '--role') {
+      if (args[i + 1] === undefined || args[i + 1].startsWith('--')) {
+        console.error('--role requires a value.');
+        process.exit(1);
+      }
       overrideRole = args[++i].trim();
     } else if (arg.startsWith('--report=')) {
       reportNum = arg.slice('--report='.length).trim();
--- a/detect-reposts.mjs
+++ b/detect-reposts.mjs
@@ -502,7 +502,12 @@ if (process.argv[1] && import.meta.url === pathToFileURL(process.argv[1]).href) 
   // Inside the main-module guard, not at import time: company-history.mjs
   // imports detectReposts/parseScanHistory from here, so a top-level check
   // would judge the IMPORTER's argv.
-  validateFlags(args, KNOWN_FLAGS, USAGE, { valueFlags: VALUE_FLAGS });
+  // requireOperand: without it, `--window --summary` reads --summary as the
+  // window value; flagValue() has no adjacency check, parseInt('--summary')
+  // is NaN, and windowDays silently falls back to DEFAULT_WINDOW_DAYS at exit
+  // 0 instead of reporting the malformed flag (#3087). Nothing more specific
+  // to say than the shared message.
+  validateFlags(args, KNOWN_FLAGS, USAGE, { valueFlags: VALUE_FLAGS, requireOperand: true });
 
   if (selfTestMode) {
     runSelfTest();
--- a/detect-reposts.test.mjs
+++ b/detect-reposts.test.mjs
@@ -1187,13 +1187,22 @@ const badWindowOut = execFileSync('node', [scriptPath, '--window', 'abc'], {
 const badWindowJson = JSON.parse(badWindowOut);
 eq('--window abc falls back to 90', badWindowJson.metadata.windowDays, 90);
 
-// Test --window with no value (falls back to default)
-const noWindowOut = execFileSync('node', [scriptPath, '--window'], {
-  encoding: 'utf-8', timeout: 10000,
-  cwd: dirname(scriptPath),
-});
-const noWindowJson = JSON.parse(noWindowOut);
-eq('--window without value falls back to 90', noWindowJson.metadata.windowDays, 90);
+// Test --window with no value: a usage error, not a silent fallback (#3087).
+// This used to fall back to the 90-day default silently at exit 0 — the same
+// shape as `--window abc` above, except there `abc` is a value the caller
+// actually typed (just an invalid one), while a bare trailing `--window` has
+// no value at all. requireOperand distinguishes the two: a missing operand is
+// now a validateFlags usage error, before the script ever computes a window.
+try {
+  execFileSync('node', [scriptPath, '--window'], {
+    encoding: 'utf-8', timeout: 10000,
+    cwd: dirname(scriptPath),
+  });
+  ok('--window without value exits non-zero', false);
+} catch (e) {
+  ok('--window without value exits non-zero', e.status === 1);
+  ok('--window without value reports the missing operand', /--window requires a value/.test(e.stderr || ''));
+}
 
 // Test --help flag
 const helpOut = execFileSync('node', [scriptPath, '--help'], {
--- a/doctor.mjs
+++ b/doctor.mjs
@@ -43,7 +43,13 @@ const USAGE = `Usage:
 
 CLIs: ${VALID_CLIS.join(', ')}`;
 
-validateFlags(argv, KNOWN_FLAGS, USAGE, { valueFlags: VALUE_FLAGS });
+// requireOperand: without it, `--target --json` reads --json as the target
+// path (argv[targetIdx + 1] below has no adjacency check of its own), and the
+// doctor silently diagnoses a directory literally named "--json" at exit 0
+// (#3087) — this is the onboarding entrypoint, so that's a user told to
+// create files that already exist. Nothing more specific to say than the
+// shared message.
+validateFlags(argv, KNOWN_FLAGS, USAGE, { valueFlags: VALUE_FLAGS, requireOperand: true });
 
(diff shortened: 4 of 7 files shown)


santifer:
Substantive look done, @kovalancik06-max, and the direction is right: closing the missing-operand hole in the shared helper (per @dalenys's #3087 analysis) protects all fifteen call sites at once instead of one script at a time. Sequencing note before this merges: two sibling PRs landed today around you — #3088 took archive-posting.mjs (merged shortly) and #3057/#3085 extended the shared validation test file — and #2983 (the value-side of the same defect family, four CLIs) is one typo away from merging and touches three of your files. Rather than have you rebase twice: hold until #2983 lands, then rebase once over the lot; your helper-level change survives all of them intact, and I'll ping you here the moment the path is clear. The audit table in your PR body is exactly the right way to present this class of change.

Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/cli-flag-validation.test.mjs`; the script changes and the root `detect-reposts.test.mjs` (not under `tests/`) are outside my area and I read them for context.

The seven new cases assert that a value-taking flag followed by another flag now exits 1 with `--X requires a value`, across doctor, detect-reposts, process-quality and weekly-digest through `validateFlags`'s `requireOperand`, and archive-posting through its own adjacency check. Each rejection fires inside `validateFlags` before any `data/` access, so no fixture is needed, and the exit-code assertion is the sentinel: without the wiring these scripts exit 0 on the old silent fallback and the test fails. The message match pins the right rejection rather than any non-zero exit.

santifer:
Welcome to career-ops, @kovalancik06-max! The caller-by-caller audit in your PR body is the model: you didn't just fix the four unguarded sites, you documented why scan-ats-full, contacts and check-table-freshness stay as they are, using the criteria from cli-flags.mjs's own comments. And catching the assertion in detect-reposts.test.mjs that enshrined the silent fallback as correct is finer work than adding new tests. Merged, your first contribution. Glad to have you here. 🚀

kovalancik06-max:
Thanks so much, Santiago! Really glad the audit approach landed well and
appreciate the warm welcome.

I found another bug while poking around the codebase: parseAmount() in
salary-gap.mjs treats "." as a decimal point everywhere, so period-grouped
salaries (Spain, Germany, Italy, NL, Brazil, etc.) like "35.000" silently
fold to 35 instead of 35000 -- reachable via data/salary-observations.tsv
at the highest actual-comp trust tier. Filed a fix as PR #3197 (self-test +
full test-all.mjs green, 5304 passed / 0 failed). Would appreciate a look
when you have time.

PR #3100, opened 2026-08-20
unify(trust-validator): consume lib/ascii-fold, with punctuation as an explicit option

### LINKED ISSUE #3040: unify: _trust-validator should consume lib/ascii-fold instead of carrying a second folding table
#2927 landed an accent fold inside `providers/_trust-validator.mjs` (`asciiFoldForHostname`), and #2931 landed the shared primitive `lib/ascii-fold.mjs` minutes later with the same NFD + non-decomposing table. Both are correct; there are now two copies of the folding table in the tree, and two copies is how #2777's lock-protocol drift happened: they agree today and diverge the first time someone patches one of them.

The migration: `_trust-validator.mjs` imports `asciiFold` from `lib/ascii-fold.mjs`, keeps its own space-collapse/trim contract on top (the shared fold already lowercases, strips marks, applies the table, and collapses whitespace — check the two functions' final regex classes line up: `[^a-z0-9 ]` there vs `[^a-z0-9\s]` here), and deletes its local `NON_DECOMPOSING_LATIN` copy. `test-trust-validator.mjs` should not change: the behavior is identical, which is the point.

Claiming it: I said "on my side" in #2931's merge note, so I'll take it unless someone beats me to it — if you want it, comment and it's yours.

abankar1:
Fixes #3040. Claimed in-thread; this is the follow-up I promised in both #2927 and #2931, so the second copy is mine to remove.

`providers/_trust-validator.mjs` now imports `asciiFold` from `lib/ascii-fold.mjs` and its local `NON_DECOMPOSING_LATIN` copy is gone.

## Your design note was the whole job

The two final classes differ in behaviour: `'delete'` removes residual punctuation, `'space'` turns it into a separator. I went with the options flag rather than the documented loosening.

Working through why the `>= 3` word filter does **not** make it converge: the bare `o` from `o brien` is indeed dropped, so `O'Brien` converges on the slug either way. But punctuation-as-separator can *create* words that were never there:

```
Smith&Jones   space  -> "smith jones"   words: [smith, jones]
              delete -> "smithjones"    words: [smithjones]
```

`smith` then substring-matches `smithfield.com`, and the mismatch flag that should fire doesn't. Direction is "fewer false penalties", as you said — but on a legitimacy validator the flag *not* firing is the failure mode that costs something, so I'd rather not take it as a silent freebie.

So: `asciiFold(value, { punctuation: 'delete' })`, with `'space'` remaining the default so every existing caller is untouched.

## A second difference the note didn't mention

The delete class is `[^a-z0-9 ]` with a **literal space**, not `\s`. So a tab or newline is *removed* rather than collapsed:

```
'a\tb'   space -> "a b"    delete -> "ab"
```

That's behaviour on prose input, so it's preserved exactly and pinned by a test.

## Identical, measured rather than asserted

I inlined the pre-migration implementation verbatim and diffed it against the new one across a 43-input corpus — accents, non-decomposing Latin (`ø æ œ ß đ ł þ ð ħ ı ŋ ŧ ĸ ſ`), CJK/Cyrillic/Greek, punctuation-heavy brand names (`AT&T`, `Ben & Jerry's`, `L'Oréal`, `Booking.com`), full-width, tabs, empty, `null`, `undefined`:

```
corpus: 43 inputs, 0 differences
```

`test-trust-validator.mjs` is deliberately untouched and still passes 107/107 — the proof you asked for.

## But that proof isn't sufficient on its own, so I added the one that is

Worth flagging plainly: **with the validator changed to take the shared `'space'` default — the blind swap this migration invites — `test-trust-validator.mjs` still passes 107/107.** The existing suite cannot see the loosening.

So `tests/ascii-fold-punctuation.test.mjs` demonstrates it instead, and its decisive assertion is end-to-end rather than on the fold alone:

```js
assert.equal(companyMatchesHostname('Smith&Jones', 'smithfield.com'), false);
assert.equal(companyMatchesHostname('Smith&Jones', 'smithjones.com'), true);
```

Verified it discriminates: with the validator taking the shared default, exactly that assertion fails and the other four pass.

`node test-all.mjs --quick`: 4402 pass / 0 fail (3 warnings pre-existing and environmental).

## Summary by CodeRabbit

* **New Features**
  * Added configurable punctuation handling for ASCII folding, including preservation or removal of punctuation.
  * Improved hostname normalization for consistent matching behavior.

* **Bug Fixes**
  * Tabs and newlines are handled correctly when punctuation is removed.
  * Names without Latin characters now produce empty normalized results where appropriate.

* **Tests**
  * Added coverage for punctuation modes, whitespace handling, and hostname matching.


santifer:
This is exactly what the issue asked for and better argued than the issue itself, @abankar1. Making the punctuation behaviour an explicit option instead of harmonizing it silently is the right call, and the reason you wrote down is the whole point: 'Smith&Jones' folding to two words means 'smith' substring-matches smithfield.com, so a mismatch that should be flagged silently isn't. Slug callers converge either way, word-level callers don't — so the caller states which it needs. Second copy of the table gone. Merged 🚀

Issue #1169, opened 2026-06-22
Title filter matches short keywords as substrings ("CTO" matches inside "Director")


gtyree:
## Summary

`buildTitleFilter` in `scan.mjs` matches each keyword with `String.prototype.includes`, which is a substring match. Short keywords therefore match *inside* longer words. The most damaging case: the keyword **`CTO`** matches the substring in **`direCTOr`**.

## Impact

Any `portals.yml` whose `title_filter.positive` contains `CTO` (the shipped example does) will pass **every** Director-level title, regardless of department. During a scan, non-engineering roles get added to the pipeline:

- `Director of Analytics`
- `Director, FP&A - Growth`
- `Director, Global Provider GTM`
- `Director, Provider Product Marketing`
- `Senior Director of Lifecycle Strategy`

All five match purely because of `cto` ⊂ `direСtor`. This pollutes the pipeline on every run and undermines the title filter's purpose. It affects all users with `CTO` (or any other short keyword that is a substring of common words) in their positive list.

## Reproduction

```js
import { buildTitleFilter } from './scan.mjs';
const f = buildTitleFilter({ positive: ['CTO', 'Director of Engineering'] });
f('Director of Analytics'); // => true  (expected false)
```

## Root cause

```js
const hasPositive = positive.some(k => lower.includes(k)); // substring, not word-aware
```

## Suggested fix

Match keywords on word boundaries (not flanked by other letters) so surrounding spaces/punctuation still match but substrings inside words don't:

```js
new RegExp(`(?<![a-z])${escaped}(?![a-z])`, 'i')
```

PR with the fix + regression tests follows.

santifer:
Fixed by #1102 (just merged) — short acronyms like `CTO` now match on word boundaries, so `direCTOr` no longer slips through a CTO filter. Thanks for the precise diagnosis, @gtyree; your example is exactly what made the fix clear. 🙏

Issue #1101, opened 2026-06-18
Title filter: short acronyms (CFO, COO, SDR) should match on word boundaries, not as substrings


eliador90:
### Problem
The scan title filter matches keywords as plain substrings (`lower.includes(kw)` in `buildTitleFilter`). For short acronyms this causes false matches:

- positive `COO` matches `Coordinator` (e.g. `Marketing Coordinator`)
- `SDR` / `BDR` / `GSI` match mid-word
- as a **negative** keyword, `COO` wrongly knocks out legitimate `Coordinator` titles

### Proposed change
Compile each lowercased keyword into a matcher:

- short all-letter acronyms (2–3 chars: `cfo`, `coo`, `sdr`, …) match on **word boundaries** (`\bkw\b`)
- multi-word phrases and keywords containing non-letters (`.NET`, `SAP `, `L&D`) keep fast, permissive **substring** matching

This changes core title-matching semantics for every user's `title_filter`, so it's handled in its own focused PR with tests rather than bundled with provider additions (split out from #825 per review feedback).

PR #3104, opened 2026-08-20
fix(scan): anchor title keywords on word edges, not keyword length

### LINKED ISSUE #3103: title_filter anchors word boundaries by keyword LENGTH, so "Rust Engineer" matches "Zero Trust Engineer" and "DeFi" matches "Software Defined"
## Problem

#1101 and #1169 anchored 2-3 letter acronyms on word boundaries, so `COO` stopped matching `Coordinator`. The guard is keyed on keyword *length*:

```js
export function compileKeyword(kw) {
  if (/^[a-z]{2,3}$/.test(kw)) {           // <- length, not word-ness
    const re = new RegExp(`\\b${kw}\\b`);
    return (lower) => re.test(lower);
  }
  return (lower) => lower.includes(kw);    // everything else: substring
}
```

Length was never the property that made a substring match wrong, though - being a word was. Every keyword of 4 characters or more still matches mid-word, and the failure is identical, just less obvious because the examples are less famous:

|Keyword|Matches|
|---|---|
|`Rust Engineer`|`Sr. Zero Trust Engineer III (6794)`|
|`DeFi`|`Senior Engineer System Software, Software Defined Networking`|
|`Solana`|`Genomic Breeder (Solanaceae)`|
|`Fuel`|`Diesel Mechanic Assistant/Fueler`, `Cleaner Fueller Shunter Supervisor`|
|`DeFi`|`Auxiliar Administrativo, Vaga Afirmativa para Pessoas com Deficiência`|

Every row above is a real posting written to a live `data/pipeline.md` by `scan-ats-full.mjs`. `T-rust Engineer` is my favourite, because the keyword is two words long and still lands mid-word.

## Blast radius

Measured over the 488 distinct titles in my `scan-history.tsv`: **68 are false positives attributable purely to substring bleed**, i.e., they match no keyword once boundaries are applied. On a `tracked_companies` scan that is noise you triage by hand. On a reverse ATS sweep across ~39k boards it is the dominant failure mode, since the corpus contains every industry and words like `fuel`, `anchor` and `defi` are common English.

## The trap in the obvious fix

Anchoring naively - `\bkeyword\b` for anything word-edged - drops real roles, because keywords are written in the singular and boards post both forms:

```
"smart contract" vs "Senior Software Engineer, Blockchain (Smart Contracts)"   -> NO MATCH
```

I hit exactly this while measuring: the naive rule removed 75 titles, and one of them was a genuine smart-contract role. Given #2544's point about silent title drops, a fix that trades false positives for silent false negatives is not an improvement.

## Proposed Solution

Anchor on **word-edged-ness** rather than length, and accept a trailing plural:

1. **Keyword begins and ends with a word character** (`rust engineer`, `defi`, `solana`, `cfo`): compile to `\bkeyword(?:e?s)?\b`.
2. **Keyword has a non-word character at either edge** (`.NET`, `SAP `, `L&D`, `ink!`, `C++`): keep substring matching. This is not a nicety - `\b` is defined against word characters, so anchoring `c++` would demand a word character after `+` and the keyword could never match anything.

This subsumes the `/^[a-z]{2,3}$/` branch, so #1101's behaviour is preserved rather than replaced.

Re-measured with the plural exception: 68 titles dropped, **0 real roles lost, 0 newly matching**. The three that my crypto-detector flagged as possible losses were `Solanaceae`, and the two `Zero Trust Engineer` postings - my detector regex had the same substring bug the fix addresses.

PR attached, with the sweep's real-world examples as test cases.

## Alternatives considered

1. **Raise the length threshold** (anchor 2-6 chars instead of 2-3): arbitrary, and it would not fix `Rust Engineer`, which is 13 characters.
2. **Anchor only the start of the keyword**: cheaper, but it does not fix the motivating cases. `\bdefi` still matches `Defiance`, since `Defiance` starts on a word boundary.
3. **Let users opt in per keyword** (e.g. a `\b` prefix in `portals.yml`): pushes a regex detail into user config, and every user hits this before they learn the syntax exists.
4. **Full stemming / a plural library**: a dependency and a large behaviour change for one suffix. `(?:e?s)?` covers the case that actually appears in job titles.

Related: #1101 and #1169 (the short-acronym fix this generalises), #2544 (silent title drops), #3089 (reverse-sweep precision, where this is one of two root causes).

Would you rather see the plural exception as `(?:e?s)?`, or keep anchoring strict and treat plural spellings as the user's job to add?

IaroslavMazur:
Closes #3103.

## Problem

`compileKeyword` gates word-boundary matching on keyword *length*:

```js
if (/^[a-z]{2,3}$/.test(kw)) { /* \b anchored */ }
return (lower) => lower.includes(kw);   // 4+ chars: substring
```

Length was never the property that made a substring match wrong - being a word was. Every 4+ character keyword still matches mid-word, on real postings:

|Keyword|Matched|
|---|---|
|`Rust Engineer`|`Sr. Zero Trust Engineer III (6794)`|
|`DeFi`|`Senior Engineer System Software, Software Defined Networking`|
|`Solana`|`Genomic Breeder (Solanaceae)`|
|`Fuel`|`Diesel Mechanic Assistant/Fueler`|

## Change

`compileKeyword` now anchors any keyword that **begins and ends with a word character**, which subsumes the `/^[a-z]{2,3}$/` branch rather than replacing it, so #1101 and #1169 behaviour is preserved.

Two deliberate details:

1. **A trailing plural is accepted** (`(?:e?s)?`). Keywords are written singular and boards post both, so strict anchoring makes `"Smart Contract"` miss `Senior Software Engineer, Blockchain (Smart Contracts)`. I hit that during measurement - the naive rule dropped 75 titles and one was a genuine smart-contract role. Per #2544, trading false positives for silent false negatives is not an improvement.
2. **Keywords with a non-word character at either edge keep substring matching** (`.NET`, `SAP `, `L&D`, `ink!`, `C++`). This is a correctness requirement, not a preference: `\b` is defined against word characters, so anchoring `c++` would demand a word character after `+` and the keyword could never match anything.

## Measured impact

Over the 488 distinct titles in my `scan-history.tsv`, against my own `title_filter`:

|  | naive `\b` | with the plural exception |
|---|---|---|
| titles no longer matching | 75 | **68** |
| real roles lost | 1 | **0** |
| newly matching | 0 | **0** |

The three the measurement initially flagged as possible losses were `Solanaceae` and the two `Zero Trust Engineer` postings. My own detector regex had the same substring bug the change fixes.

## Tests

`tests/title-keyword-word-boundaries.test.mjs`, 17 assertions: #1101's acronym cases still pass, each real-world bleed above is refused, the same keywords still match their own titles, plurals resolve, `.NET`/`ink!`/`C++` keep matching, and one end-to-end pass through `buildTitleFilter` covering a plain keyword and an AND-group.

`node test-all.mjs`: 4433 passed, 0 failed, 3 warnings (all environmental - no user data in a clean checkout, no Go compiler).

This is a behaviour change to a matcher every scan path shares, so it is worth a close look at the plural exception specifically. I am happy to drop it and make plural spellings the user's job, if that is the call - the tradeoff is discussed in #3103.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved keyword matching to avoid unintended matches within larger words.
  - Added support for valid singular, plural, and trailing “e” keyword variations.
  - Preserved matching for keywords containing punctuation.
  - Special characters in keywords are now handled safely.

- **Tests**
  - Expanded coverage for acronym compatibility, standalone matches, plural forms, punctuation, and combined title filters.

### DIFF
--- a/scan.mjs
+++ b/scan.mjs
@@ -97,14 +97,31 @@ const CONCURRENCY = 10;
 
 // ── Title filter ────────────────────────────────────────────────────
 
-// Compile a lowercased keyword into a matcher. Short all-letter acronyms
-// (2-3 chars: cfo, coo, sdr, bdr, gsi…) match on WORD BOUNDARIES so "COO" no
-// longer matches "Coordinator", "SDR" no longer matches anything mid-word, etc.
-// Multi-word phrases and keywords containing non-letters (".NET", "SAP ",
-// "L&D") keep fast, permissive substring matching.
+// Compile a lowercased keyword into a matcher. Any keyword that BEGINS and
+// ENDS with a word character matches on WORD BOUNDARIES: "COO" no longer
+// matches "Coordinator" (#1101), and by the same rule "Rust Engineer" no
+// longer matches "Zero Trust Engineer", "DeFi" no longer matches "Software
+// Defined Networking", "Solana" no longer matches "Genomic Breeder
+// (Solanaceae)", and "Fuel" no longer matches "Diesel Mechanic
+// Assistant/Fueler". Length was never the property that made a substring
+// match wrong - being a word was.
+//
+// A trailing plural is still accepted, because a keyword is written in the
+// singular and boards post both: "Smart Contract" has to keep matching
+// "Senior Software Engineer, Blockchain (Smart Contracts)". Anchoring
+// without this exception silently drops real roles, which is a worse failure
+// than the false positives it fixes (see #2544 on silent title drops).
+//
+// Keywords containing non-word characters at either end (".NET", "SAP ",
+// "ink!", "C++", "L&D") keep permissive substring matching: \b is defined
+// against word characters, so anchoring "c++" would demand a word character
+// after "+" and the keyword would never match anything.
+const WORD_EDGED = /^\w[\s\S]*\w$|^\w$/;
+const REGEX_SPECIAL = /[.*+?^${}()|[\]\\]/g;
+
 export function compileKeyword(kw) {
-  if (/^[a-z]{2,3}$/.test(kw)) {
-    const re = new RegExp(`\\b${kw}\\b`);
+  if (WORD_EDGED.test(kw)) {
+    const re = new RegExp(`\\b${kw.replace(REGEX_SPECIAL, '\\$&')}(?:e?s)?\\b`);
     return (lower) => re.test(lower);
   }
   return (lower) => lower.includes(kw);
--- a/tests/title-keyword-word-boundaries.test.mjs
+++ b/tests/title-keyword-word-boundaries.test.mjs
@@ -1,0 +1,62 @@
+// tests/title-keyword-word-boundaries.test.mjs — compileKeyword anchors any
+// word-edged keyword on word boundaries, not just 2-3 letter acronyms (#1101),
+// while keeping trailing plurals and punctuation-bearing keywords working.
+import { join } from 'path';
+import { pathToFileURL } from 'url';
+import { pass, fail, ROOT } from './helpers.mjs';
+
+console.log('\ntitle_filter — word-boundary keyword matching');
+
+const { compileKeyword, buildTitleFilter } = await import(pathToFileURL(join(ROOT, 'scan.mjs')).href);
+
+const check = (kw, title, expected, why) => {
+  const got = compileKeyword(kw)(title);
+  if (got === expected) pass(why);
+  else fail(`${why} — "${kw}" vs "${title}": expected ${expected}, got ${got}`);
+};
+
+// #1101 behaviour is preserved exactly.
+check('cfo', 'group cfo, emea', true, 'short acronym still matches as a word');
+check('cfo', 'cfom', false, 'short acronym still refuses a mid-word hit');
+
+// The same rule now applies to longer keywords. Each of these was a real
+// posting written to a live pipeline by a reverse ATS sweep.
+check('rust engineer', 'sr. zero trust engineer iii (6794)', false, '"Rust Engineer" does not match "Trust Engineer"');
+check('defi', 'senior engineer system software, software defined networking', false, '"DeFi" does not match "Software Defined"');
+check('solana', 'genomic breeder (solanaceae)', false, '"Solana" does not match "Solanaceae"');
+check('fuel', 'diesel mechanic assistant/fueler', false, '"Fuel" does not match "Fueler"');
+check('near', 'praktikum im bereich near patient care biostatistics', true, '"NEAR" still matches a standalone word');
+
+// Real matches must survive.
+check('rust engineer', 'senior rust engineer', true, '"Rust Engineer" still matches its own title');
+check('solana', 'solana program engineer', true, '"Solana" still matches');
+check('fuel', 'fuel network protocol engineer', true, '"Fuel" still matches as a word');
+
+// Trailing plural: keywords are written singular, boards post both. Anchoring
+// without this drops real roles silently, which is worse than the noise fixed.
+check('smart contract', 'senior software engineer, blockchain (smart contracts)', true, 'singular keyword matches a plural title');
+check('canister', 'rust canisters engineer', true, 'plural tolerated on a one-word keyword');
+check('smart contract', 'smart contract engineer', true, 'singular keyword still matches singular title');
+
+// Punctuation-bearing keywords keep substring matching: \b is defined against
+// word characters, so anchoring "c++" would require a word char after "+" and
+// the keyword could never match.
+check('c++', 'principal backend development engineer(c++)', true, '"C++" still matches');
+check('ink!', 'ink! smart contract developer', true, '"ink!" still matches');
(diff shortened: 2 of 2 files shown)


Solaris-star (commented):
Reviewed head `5b9685967e8b0aa61c10630c82a7ef8350a5948c`.

The focused regression command passed: `node tests/title-keyword-word-boundaries.test.mjs` (17/17 assertions). I also ran a direct probe against this head:

```text
compileKeyword('llmops')('llmopses')       -> true
compileKeyword('business systems')('business systemses') -> true
compileKeyword('analysis')('analysises')   -> true
compileKeyword('ai')('aies')                -> true
```

The unconditional `(?:e?s)?` suffix is applied even when the configured keyword already ends in `s`, so malformed variants can pass the title filter. This affects real shipped keywords such as `LLMOps`, `MLOps`, `RevOps`, `Business Systems`, `Master Thesis`, and `Bachelor Thesis`, not just synthetic examples. Could the plural expansion be guarded for keywords already ending in `s`, or narrowed to the intended morphology, with a regression case for an already-plural keyword? The motivating singular/plural cases still pass; this is a correctness edge I would prefer to close before merge.

I did not claim the full `node test-all.mjs` suite from this review checkout; the evidence above is limited to the focused PR test and the direct matcher probe.

santifer:
Real analysis here, @IaroslavMazur, and the examples are the strongest part: "Rust Engineer" matching "Zero Trust Engineer" and "Solana" matching "Genomic Breeder (Solanaceae)" are exactly the failures length-gating was never going to catch. "Length was never the property that made a substring match wrong - being a word was" is the right sentence.

Before this can move, something you couldn't have known: **#2970 (@artemtrofymenko, opened three days ago) changes the same function**, and it lands on a different answer. It makes anchoring **opt-in per keyword** rather than the default, on this argument: "crypto" is meant to catch "Cryptocurrency" and "fellows" to catch "Fellowship", so anchoring the whole list by default would silently stop matching things existing configs rely on.

Both arguments are real, and they cannot both be the default. That is a product-direction call about matching semantics for every existing config, so it goes to Santiago rather than being settled by whichever PR I merge first. I'll bring the answer straight back here.

Nothing about your work is in question: keep it as is, don't rebase yet, and thanks for filing #3103 with the measurements — that issue is what makes the decision decidable at all.

Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is the new `tests/title-keyword-word-boundaries.test.mjs`; the `scan.mjs` change is outside my area and I read it for context.

The test tracks the new word-edge rule and the plural tolerance through both `compileKeyword` and `buildTitleFilter`. I traced each case: the acronym behaviour is preserved, `rust engineer` no longer matches `trust engineer`, `defi` does not match `defined`, `solana` does not match `solanaceae`, trailing plurals still match, and punctuation-bearing keywords fall through to substring. The sentinel is real: under the old two-to-three-letter rule `rust engineer` matches `trust engineer`, so the test reddens if the fix is reverted.

One small aside, and it is on `scan.mjs` so it stays off this thread and out of my remit: the header lists "L&D" as an example of a keyword that keeps substring matching, but `l&d` begins and ends with word characters, so it takes the anchored regex path, not substring. The behaviour is fine either way; only the example in the comment is misfiled.

IaroslavMazur:
@santifer, a measurement update while the call is open, so it is in front of you when the answer comes back.

@artemtrofymenko's numbers on #3103 hold up, and I reproduced them on my own corpus rather than take them: my `(?:e?s)?` exception is too narrow, and the `0 real roles lost` in my commit message does not survive a 628-title corpus. `DeFi + Engineer` drops `Engineering Lead - Crypto and DeFi`, which is a real role in my own config, so the claim is mine to correct.

The fix is one group in the regex, and I measured the two candidates instead of picking one: `(?:e?s|ing)?` recovers everything `(?:e?s|ing|ure)?` recovers on both configs, while `ure` puts back a false positive of exactly the kind this PR removes (`contract` matching "Contracture Management Specialist"). Numbers, the four-worktree setup, and the dictionary sweep are on #3103.

No rebase and no push from me while the call is open, as you asked. Whichever default Santiago lands on, though, the group should read `(?:e?s|ing)?` rather than what I pushed.

santifer:
Closing with the decision made, @IaroslavMazur, and with the record straight about what decided it: **your own measurement did.** The 628-title corpus you ran against your own PR — and reported honestly when it went against you — showed anchor-by-default loses real roles ("DeFi + Engineer" dropping "Engineering Lead - Crypto and DeFi"), and that settled the default on broad-with-opt-in: #2970 is the accepted design and is landing now (its CI is re-running on today's base after this week's merges; it merges on green). Reporting evidence against your own position is the most valuable thing a contributor can do here, and it will not be forgotten.

Two standing invitations: (1) your corpus deserves to live in the repo as a recall regression test — a PR adding it (titles + expected matches under the default matcher) lands fast and carries your name on exactly the guarantee your measurement created; (2) if you later find an exception set that survives the corpus with zero real-role loss, the anchored-default conversation reopens on that evidence.

PR #3118, opened 2026-08-20
fix(agent-inbox): seed the queue under the lock, not before it

LayerC0de:
## The bug

`ensureFile()` ran **outside** the queue lock:

```js
if (existsSync(PATH)) return;                    // (1)
writeFileSync(PATH, HEADER, { flag: 'wx' });     // (2) open(O_CREAT|O_EXCL) … then write()
```

`wx` closed the case it was added for — two writers both passing (1), the second's
default-`w` write truncating the first's item. But `wx` makes the **create** atomic, not
the **initialisation**. `writeFileSync` is `open()` *then* `write()`, and between those
two syscalls the file **exists and is zero bytes**.

That admits a third participant the exclusive flag cannot see: a writer arriving *inside*
the window finds `existsSync === true`, skips creation, takes the lock **correctly**, and
appends — into a file the creator is about to overwrite from offset 0. When the 479-byte
header lands, that line is simply gone.

Every writer exits 0, and the queue comes out perfectly well-formed and one item shorter.
That is the `kept=29 of 30` `agent-inbox-tests` §7 has been reporting on windows-latest —
a real data-loss bug, not a flake. Full investigation in
santifer/career-ops#2905 (comment).

## Evidence

The window is directly observable — a second process polling `existsSync` and stat-ing
the moment the file appeared saw it at **0 bytes in 303 of 400 rounds** on Windows.

Reproduced end to end by pinning the process tree to 2 cores, so the creator only has to
lose its timeslice between `open()` and `write()`:

```
=== ROUND 15 FAILED — missing=[item-2,item-12,item-16,item-28] nonzeroExits=0
  writers that won the wx create: item-19

  item-2 : existsSync seen=true → preLock size=0 → appended before=0   after=34
  item-16:                                          appended before=104 after=139
           → postCheck mine=FALSE      ← watched its own line vanish

    BREAK item-23: 479 -> 514   [expected before=139]
```

Four writers walked the window, each **correctly serialised by the lock** (every `before`
equals the previous `after`). Then `item-19`'s `write(HEADER)` landed at offset 0 and
replaced all 139 bytes. `479` is `Buffer.byteLength(HEADER)` to the byte.

For completeness: `appendFileSync` is atomic across processes on Windows — 25 rounds ×
30 concurrent writers with **no lock at all**, zero losses. The append and the lock are
both innocent.

## The change

Two lines of behaviour — `ensureFile()` moves inside `withPipelineLock`. It reuses the
lock already taken one line later, adds no second mechanism, and the critical section
stays sub-millisecond. `'wx'` stays as the guard against writers that are not this
function.

This makes the window **unreachable** rather than narrow: no writer can observe the file
until the creator has released, and the creator writes the header before it releases.

A/B at 30 concurrent writers pinned to 2 cores, arms alternated round by round, **400
rounds each**:

|                            | shipped | fixed |
|----------------------------|---------|-------|
| appends into a 0-byte file | 4       | **0** |
| silent loss (all exited 0) | 3       | **0** |

## Test

The regression test is **deterministic, not a race**: it holds the lock itself and
asserts the queue file does not appear while held, so a seed happening before acquisition
is visible without needing a loaded multi-core runner.

- with the fix: 23 passed, 0 failed
- against unfixed `agent-inbox.mjs`: 22 passed, 1 failed — `❌ queue file is NOT created while another process holds the lock`

Full suite on an idle machine: **4866 passed, 0 failed**. Fresh-install path checked
explicitly, since `ensureFile()` no longer pre-creates `data/` — `acquirePipelineLock`
already does `mkdirSync(dirname(lockDir), { recursive: true })`.

## Scope — please read before expecting §7 to go quiet

This removes the **silent** loss. It does **not** address a separate defect in
`pipeline-lock.mjs`: `lockCanRecover()`'s `catch` returns `true` for "vanished" and the
caller acts on it by `rmSync`-ing rather than retrying, which can delete a lock another
writer legitimately created in the meantime. Traced across three processes in a 540 µs
window. §7 can still go red with a **non-zero exit** until that is fixed separately.

Also left alone: a process killed between `open()` and `write()` while holding the lock
still leaves a 0-byte file, after which adds append with no header. Cosmetic, no data
loss, and fixing it means widening `ensureFile()`'s contract from "exists" to "exists and
is initialised" — beyond this bug.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved queue initialization to prevent incomplete or corrupted queue files during concurrent operations.
  * Ensured new queue items are added only after required queue setup is complete.
  * Added safeguards for operations running while another queue process holds the lock.

* **Tests**
  * Added coverage verifying safe queue initialization, header creation, and item placement under locking conditions.

### DIFF
--- a/agent-inbox-tests.mjs
+++ b/agent-inbox-tests.mjs
@@ -14,16 +14,20 @@
  *      the personal queue isn't accidentally tracked.
  *   7. Concurrent `add` calls all survive — the queue is appended to, never
  *      rewritten, so simultaneous writers cannot clobber each other.
+ *   8. The queue file is SEEDED under the lock too, not merely appended to under
+ *      it. Creating it is open() then write(), and a writer that observed the
+ *      gap appended into a zero-byte file and lost its item to the header.
  *
  * Provisions a throwaway queue via CAREER_OPS_INBOX and a temp CWD; never
  * touches real user data.
  */
 
 import { execFileSync, spawn } from 'child_process';
-import { readFileSync, writeFileSync, mkdtempSync } from 'fs';
+import { readFileSync, writeFileSync, mkdtempSync, existsSync } from 'fs';
 import { join, dirname } from 'path';
 import { tmpdir } from 'os';
 import { fileURLToPath } from 'url';
+import { acquirePipelineLock } from './pipeline-lock.mjs';
 
 const ROOT = dirname(fileURLToPath(import.meta.url));
 const NODE = process.execPath;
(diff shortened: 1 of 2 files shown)


santifer:
Another face of the same family, @LayerC0de: `wx` closes the two-writers-create case but not the read-then-write window before the lock, and you keep finding these where they actually live rather than where they show up. Merged 🚀

PR #3122, opened 2026-08-20
fix(add-entry): delegate flag validation to shared validateFlags

### LINKED ISSUE #3112: fix(add-entry): a mistyped flag is silently ignored — delegate to the shared validateFlags
**Time:** ~25 min · **Difficulty:** Beginner

`add-entry.mjs` declares `KNOWN_FLAGS` but checks flags by hand with `args.includes(...)`: a mistyped `--dry-rum` is silently ignored and the script runs for real when the user asked for a preview. Six scripts were migrated to the shared helper this week — mirror #3057 (upskill.mjs) exactly: import `validateFlags` from `./lib/cli-flags.mjs`, call it inside `main()` (NOT module scope — this file is imported for its exports, see the comment pattern in #3088), and add an `['add-entry.mjs', '--dry-rum']` row to `tests/cli-flag-validation.test.mjs`.

**Acceptance:** `node add-entry.mjs --dry-rum` exits non-zero naming the bad flag; the shared validation test covers it; valid flags behave exactly as before.

vaishaldsouza:
## What does this PR do?

Replaces add-entry.mjs’s manual args.includes(...) flag checking with the shared validateFlags helper, so a mistyped flag (e.g. --dry-rum) is rejected instead of silently ignored. Also updates the pre-existing add-entry unknown-flag assertion in test-all.mjs: it checked for a "Usage:" string in stderr, which matched the old hand-rolled validation but not validateFlags’s actual output (validateFlags only prints Usage: for --help, not for an unrecognized flag). Updated to check for the actual /unrecognized flag/ pattern, matching the same assertion shape already used for reply-watch.mjs’s migration (#2743).

## Related issue

Closes #3112

## Type of change

- [ ] Bug fix
- [ ] New feature

## Checklist

- [x] I ran `node test-all.mjs` — 4913/4914 passed. The one remaining
      failure (`live archive`) is a pre-existing Playwright/network
      timeout against a live Greenhouse URL, unrelated to this change
      and present before it.

---

## Bug Fixes

- Improved command-line option validation for unknown or mistyped flags.
- Help and usage behavior now works consistently when invalid options are provided.
- Invalid options are rejected before any payload data is read or written.
- Added coverage for invalid flags used alone or alongside other options.

### DIFF
--- a/add-entry.mjs
+++ b/add-entry.mjs
@@ -42,6 +42,7 @@ import { readFileSync, writeFileSync, existsSync } from 'fs';
 import { join, dirname } from 'path';
 import { fileURLToPath, pathToFileURL } from 'url';
 import { normalizeTextKey } from './tracker-parse.mjs';
+import { validateFlags } from './lib/cli-flags.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
 
@@ -216,17 +217,9 @@ async function readStdin() {
 async function main() {
   const args = process.argv.slice(2);
 
-  if (args.includes('--help') || args.includes('-h')) {
-    console.log(USAGE);
-    process.exit(0);
-  }
-
-  const unknownFlags = args.filter(a => a.startsWith('-') && !KNOWN_FLAGS.includes(a));
-  if (unknownFlags.length) {
-    console.error(`add-entry: unrecognized flag(s): ${unknownFlags.join(', ')}. Valid flags: ${KNOWN_FLAGS.join(', ')}`);
-    console.error(USAGE);
-    process.exit(1);
-  }
+  // Migrated to shared validateFlags to reject mistyped flags (#3112).
+  // Inside the main-module guard so importers are unaffected (#3088).
+  validateFlags(args, KNOWN_FLAGS, USAGE);
 
   const dryRun = args.includes('--dry-run');
   const useStdin = args.includes('--stdin');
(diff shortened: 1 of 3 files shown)


Scott-Emberson (changes requested):
Reviewed as code owner of `tests/`. Scope is `tests/cli-flag-validation.test.mjs`; `add-entry.mjs` is outside my area and I read it for context.

One thing to fix, and it is failing CI on all three OS legs. The new row `['add-entry.mjs', '--dry-rum', /Valid flags:/]` puts `/Valid flags:/` in the third slot, which the loop consumes as the `--help` usage-header pattern (`usage` at line 63, asserted at line 74). But `add-entry.mjs --help` prints only its USAGE block, which reads `Usage: node add-entry.mjs ...` and never contains "Valid flags:". That phrase comes from `validateFlags`'s error line on the exit-1 path, not from `--help`. So the `--help exits 0 and prints usage` case fails:

```
❌ tests/cli-flag-validation.test.mjs — node:test suite failed (exit 1)
        expected: /Valid flags:/
```

The typo case itself is fine; it is only the usage-header pattern that is wrong. Drop the third element so it falls back to the default `/Usage:/i`, which add-entry's USAGE satisfies, or set it explicitly to `/Usage:/i`. One-line change and CI goes green.

Heads-up unrelated to the fix: this touches the same file as #3092, so the two will need a rebase against each other whichever lands second.

Scott-Emberson (approved):
Correcting myself: I posted request-changes a moment ago against an earlier head, and you had already pushed the fix before it landed. Re-reviewed at `d5d49e4d75`.

The third element is dropped, so `['add-entry.mjs', '--dry-rum']` now falls back to the default `/Usage:/i` for the `--help` case, which add-entry's USAGE block satisfies, and the typo case still asserts the unrecognized-flag rejection. Green on CI across all three OS. Approving. Apologies for the churn, that request-changes was stale the moment it posted.

santifer:
Another clean one in the #2919 family, @vaishaldsouza: the migration sits inside the main-module guard so importers stay unaffected, and add-entry joins the shared typo matrix with all three layers covered. Merged! 🚀

PR #3123, opened 2026-08-20
fix(locks): one recovery judgment, imported by all three copies

LayerC0de:
> **Stacked on #3120** — it exports the verdict this PR imports. Please merge that first;
> this diff is against its branch, so review it as the three call-site conversions.

## Why this exists

#3120 fixes `lockCanRecover()` in `pipeline-lock.mjs`: an ENOENT from `statSync` answered
"recoverable", and the caller acts on that by **deleting** whatever is at the path now —
which may be a lock a rival acquirer legitimately created in between, whose winner then
dies with `ENOENT … open '<path>.lock/owner.json'`.

That helper is copied into **three** other files, and all three had the same defect:

- `followup-seed.mjs`
- `portal-health-lock.mjs`
- `tracker-utils.mjs`

I originally flagged two of these; `tracker-utils.mjs` turned up when I checked the guard's
own `protocolImplementors()` list. Fixing two of three would have left the drift in place.

## The change

All three now import `lockRecoveryVerdict` from `pipeline-lock` and delete on
`RECOVER_STALE` only, at **both** call sites — the lock and the recover guard. The guard
site had the same defect: evicting on "already gone" deletes a guard another caller has
just taken, putting two callers inside the decide-then-delete window it exists to
serialize.

The local `processIsAlive()` copies go with the judgment. `readLockOwner()` stays in each —
still used by `release()` for the token check.

### A second fix these files get for free

Their copies **predated #2984**, so an owner stamp that could not be *read* fell through to
the age rule and could condemn a live lock. That is the second face of #2777, fixed in
`pipeline-lock` some time ago and still open in all three of these. Importing the judgment
closes it without a separate patch.

### OWNERLESS_GRACE_MS

Re-exported rather than redeclared. The floor is applied inside `lockRecoveryVerdict` now,
so a local copy of the number would be a constant the file no longer enforces — free to
drift from the one that decides. Nothing imported it from these modules and the value is
unchanged at `1_000`.

## The guard got stricter, not looser

`tests/lock-rm-contention.test.mjs` used to require every implementor to spell the ENOENT
discrimination correctly — which **four correct copies would satisfy**. It now requires
exactly one definition and an import in everyone else, and asserts the discrimination once,
where it is decided.

Four copies in agreement were never the goal. #2984 asked for one definition, and a repo
that merely keeps its copies in step is one patch away from the drift that produced all
three faces of #2777.

## Tests

Full suite: **4898 passed, 0 failed.** Includes `tests/portal-health-lock.test.mjs`,
`tests/with-followups-lock.test.mjs`, `tracker-writer-lock-tests.mjs` and
`test/pipeline-lock.test.mjs`, all of which exercise these acquisition paths directly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved lock recovery to prevent deletion when lock state is uncertain or the lock has already disappeared.
  * Standardized recovery behavior across health checks, follow-up processing, and tracking workflows.
  * Reduced the risk of removing newly recreated or still-active locks during contention.

* **Tests**
  * Added validation ensuring only confirmed stale locks are removed and access errors remain recoverable.

### DIFF
--- a/followup-seed.mjs
+++ b/followup-seed.mjs
@@ -66,7 +66,7 @@ import { createHash, randomUUID } from 'crypto';
 // so the next Windows-contention finding lands in one place instead of three.
 import {
   isMkdirContention, isRmContention, rmLockArtifactSync, createLockWaitPolicy,
-  sameLockDirectory,
+  sameLockDirectory, lockRecoveryVerdict, RECOVER_STALE,
 } from './pipeline-lock.mjs';
 import { tmpdir } from 'os';
 import { resolveColumns, parseTrackerRow } from './tracker-parse.mjs';
@@ -94,9 +94,12 @@ const FOLLOWUPS_LOCK_PREFIX = 'career-ops-followups-';
 
 /**
  * Minimum age before directory age alone may condemn an ownerless lock.
- * See `lockCanRecover` for why the age check needs a floor.
+ * See `lockRecoveryVerdict` for why the age check needs a floor.
+ *
+ * Re-exported rather than redeclared: the floor is applied inside that function
+ * now, so a local copy would be a constant this file no longer enforces.
  */
-export const OWNERLESS_GRACE_MS = 1_000;
+export { OWNERLESS_GRACE_MS } from './pipeline-lock.mjs';
 
 /** Structured error carrying an exit-code-mapping `code`. */
 export class SeedError extends Error {
@@ -284,16 +287,6 @@ function sleep(ms) {
   return new Promise(res => setTimeout(res, ms));
 }
 
-function processIsAlive(pid) {
-  if (!Number.isInteger(pid) || pid <= 0) return false;
-  try {
-    process.kill(pid, 0);
-    return true;
-  } catch (err) {
-    return err?.code === 'EPERM';
-  }
-}
-
 function readLockOwner(lockDir) {
   try {
     return JSON.parse(readFileSync(join(lockDir, 'owner.json'), 'utf-8'));
@@ -302,36 +295,12 @@ function readLockOwner(lockDir) {
   }
 }
 
-/**
- * Decide whether an existing lock can be safely recovered. Owner-PID liveness
- * wins; directory age is only consulted when the metadata is missing.
- *
- * That age fallback needs a floor. A lock is ownerless by construction — not
- * by accident — for the instant between its `mkdirSync` and its `owner.json`
- * write. Judging it on `age > staleMs` alone lets a caller with an aggressive
- * staleMs delete a lock created microseconds ago, stealing it from a winner
- * still inside its acquisition window and putting two writers into the
- * read-check-append critical section. OWNERLESS_GRACE_MS is a lower bound on
- * that patience, never a cap: a larger caller staleMs still wins, and a
- * genuinely abandoned lock still ages out.
- *
- * @param {string} lockDir - Directory that represents the active lock.
- * @param {number} staleMs - Age threshold for metadata-free lock recovery, floored at OWNERLESS_GRACE_MS.
- * @returns {boolean} True when the caller may remove and recreate the lock.
- */
-function lockCanRecover(lockDir, staleMs) {
-  const owner = readLockOwner(lockDir);
-  if (owner?.pid) return !processIsAlive(owner.pid);
-  try {
-    return Date.now() - statSync(lockDir).mtimeMs > Math.max(staleMs, OWNERLESS_GRACE_MS);
-  } catch (err) {
-    // Only a genuinely vanished directory means "nothing to recover". Any other
-    // stat failure — a Windows EPERM/EBUSY while the directory is mid-flight —
-    // is "could not look", and answering "recoverable" to that hands the caller
-    // an rmSync of a LIVE lock created microseconds ago (#2777, third face).
-    return err?.code === 'ENOENT';
-  }
-}
+// The recovery judgment comes from pipeline-lock rather than a fourth copy of
+// it, for the reason stated at the import: a finding lands once. This file's
+// copy had drifted twice over — it still answered a bare boolean, so "the
+// directory was gone when I looked" reached the caller as a licence to DELETE
+// whatever is at that path now, and it predated #2984, so an unreadable owner
+// stamp fell through to the age rule and could condemn a live lock.
 
 /**
  * Acquire an exclusive filesystem lock covering the read-check-append
@@ -402,7 +371,7 @@ async function acquireFollowupsLock(lockDir, followupsPath, options = {}) {
           if (!sameLockDirectory(before, after)) return; // swapped underneath us
           // Best-effort: ownership is verified, so a contended rm (Windows
           // EPERM/EBUSY while another process stats it) must not kill a caller
-          // whose work already succeeded. The orphan ages out via lockCanRecover.
+          // whose work already succeeded. The orphan ages out via lockRecoveryVerdict.
           rmLockArtifactSync(lockDir);
         },
       };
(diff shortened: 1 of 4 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/lock-rm-contention.test.mjs`; the lock sources are outside my area and I read them for context.

The test moves with the refactor, which is the right call. Section 3 now asserts each implementor imports `lockRecoveryVerdict` and defines no local copy, and it correctly excludes `pipeline-lock.mjs` from that no-copy check (`filter(f => f !== 'pipeline-lock.mjs')`), so the definition is not held to the rule meant for its importers. Section 3b asserts the verdict form on `pipeline-lock` directly, `? RECOVER_VANISHED : RECOVER_LIVE`, and that the bare-boolean spelling is gone. Green on CI.

One thing to flag rather than to fix: this supersedes #3120, which I approved earlier. Both rewrite this same file and both touch `pipeline-lock.mjs`, so only one can land, and this is the fuller design, one definition rather than four kept in agreement. If this is the one that goes in, #3120 should be closed behind it.

santifer:
#3120 is in, @LayerC0de, so this one is unblocked: a rebase onto `main` should mostly be dropping the commits that just landed underneath it.

Worth saying why I merged the base rather than waiting to take both together. In #3120 the conformance regex now accepts either spelling, and the honest reading is that it never guarded the vanished case to begin with: the old strict pattern matched `return err?.code === 'ENOENT';`, which IS the buggy line. So nothing was newly blessed. What did change is that there is no longer any pressure on the copies to move, and that pressure was the only thing keeping four implementations of the same protocol in step. That makes this PR the half that matters, not a tidy-up.

The detail I'd have missed and you didn't: `tracker-utils.mjs` turning up from the guard's own `protocolImplementors()` list rather than from memory. Fixing three of four would have left exactly the kind of drift the guard exists to prevent, and it would have looked finished.

Ping me when it's green and I'll take it straight away.

LayerC0de:
One property worth naming before this merges. It makes the three copies inherit whatever
`pipeline-lock` currently has, and what `pipeline-lock` currently has is half of the fix.

#3120 landed the `VANISHED` case, where the verdict came from a directory that was absent
when the caller looked. `STALE` has the same shape and is still open. It sits in #3139, which
is green and does not depend on this PR.

Measured in the regime these files actually run in, 30 processes each taking the lock once
and exiting:

|---|---|---|
| before #3120 | 21 | 14 of 60 |
| main today | 2 | 2 of 60 |
| with #3139 | 0 | 0 of 120 |

Merging this first is still a clear win rather than a regression. Right now those three files
carry a worse judgment than the one they would import. Theirs is a boolean that deletes on
"the directory was gone when I looked", and it predates #2984, so an owner stamp they could
not read falls through to the age rule and can condemn a live lock. Importing hands them both
of those fixes today. It does not hand them the `STALE` one until #3139 lands, and once it
does they get it with no further change here.

Merge order does not affect correctness in either direction. #3139 sooner is better.

santifer:
The conflict is mine again, @LayerC0de, and this time twice over: after your rebase yesterday I merged your own #3139 and #3143 today, and both landed in the files this PR touches. Before you rebase a third time, one thing worth checking: #3143 shared the WAIT half of the protocol across the three copies, and this PR shares the RECOVERY judgment — the two are complementary, but the ground under this one has now shifted enough that part of its diff may already be on main. If a chunk of this PR dissolved into what already landed, a smaller rebased diff (or even a your-call close if it fully dissolved) are both fine outcomes: the credit for the recovery-dedup design is yours either way.

LayerC0de:
Rebased onto `main` (`a9a0949`). Answering your question directly: **it did not dissolve — but one commit did, and the rebase was not clean.**

## What dissolved

The base commit (`7fc2166`, the #3120 half) is gone — skipped, because `main` now carries a *better* version than it. Your #3139 landed the identity check, so replaying my commit would have replaced

```js
const judged = statOrNull(lockDir);
if (judged && lockRecoveryVerdict(...) === RECOVER_STALE
  && sameLockDirectory(judged, statOrNull(lockDir) ?? {})) {
```

with the plainer `if (lockRecoveryVerdict(...) === RECOVER_STALE) {` — a regression. Same for this PR's `release()` rewrite in `followup-seed.mjs`: `main`'s stat-either-side version supersedes mine, so I took `main`'s.

## What did not

All three copies on `main` today still have their own `lockCanRecover`, still ending in the bare boolean:

```
followup-seed.mjs:332       return err?.code === 'ENOENT';
portal-health-lock.mjs:106  return err?.code === 'ENOENT';
tracker-utils.mjs:309       return err?.code === 'ENOENT';
```

None import `lockRecoveryVerdict`. #3143 shared the **wait** half; the **recovery judgment** is untouched, so this PR is still the whole of its own point. Net **−64 lines**, three definitions down to zero.

## A bug the rebase introduced, and how it surfaced

Worth reporting rather than quietly fixing, because it is a nasty shape.

My commit removed `statSync` from the `fs` import in `followup-seed.mjs` — correct against the old base, where the `lockCanRecover` this PR deletes was its only consumer. Since then `main`'s new `release()` calls `statSync` twice. So the rebase produced a file referencing an undefined binding.

`node --check` passes on it. The module *imports* fine. It only fails when `release()` actually runs — and then it throws, the lock is never removed, and the next acquirer times out:

```
❌ tests\with-followups-lock.test.mjs — LOCK_TIMEOUT: Timed out waiting for follow-ups lock
     at acquireFollowupsLock (followup-seed.mjs:434)
```

I caught it because I ran the suite on the rebased tree and then checked the failure against `main` rather than assuming it was environmental:

| suite | main | branch (pre-fix) |
|---|---|---|
| `with-followups-lock` | PASS | **FAIL** |
| `plugin-symlink-discovery` | FAIL | FAIL |

The second is this machine's symlink EPERM and fails on `main` too. The first was mine. Fixed by restoring the import; checked the other two files for the same hazard — both already import everything they use.

The general shape is worth a note for the next person rebasing across these files: **a deletion that was right against the old base can be wrong against the new one, and `fs` imports fail deferred rather than at parse time**, so a green `node --check` and a clean rebase prove nothing here.

## State

`test-all.mjs`: 5054 passed, 1 failed — the symlink EPERM above, which fails identically on `main`. CI will be the real check since it doesn't have that limitation. `tests/lock-rm-contention.test.mjs` is fully green, including the four "defines no second copy of the recovery judgment" assertions.

Scott's note stands: #3120 already merged, so nothing to close behind this.

LayerC0de:
Green on all three now — the earlier macOS red was a flake, and worth logging rather than just re-running past.

It failed three `generate-pdf` page-budget assertions with

```
Refusing to write the PDF outside the tracker workspace:
input escapes the tracker workspace:
/Users/runner/work/career-ops/career-ops/output/page-budget-test-YGYpkw/three-pages.html
```

I did not assume it was unrelated, because there is a real link: `generate-pdf.mjs` imports `resolveWorkspaceRoot` from `tracker-utils.mjs`, which this PR touches. What ruled it out:

- my diff to `tracker-utils.mjs` is confined to the lock-recovery region — it touches no path, realpath or workspace code
- a module-load failure would take out every importer, not three page-budget assertions
- macOS was green on all five other open PRs, so it is not a broken `main`
- byte-identical re-trigger (same tree `f4b020a`, new SHA) → **pass in 3m13s**

`/Users/...` vs `/private/Users/...` is macOS's realpath divergence, so a containment check that canonicalises one side and not the other is exactly the shape that fails intermittently there. I have not chased it further since it is outside this PR, but if it recurs it is worth its own issue — it is a **security guard** failing open-or-closed nondeterministically, which is a bad property for one, and the run id is `32482756373` if you want the log before it ages out.

Final state of this PR: rebased, conflicts resolved, `mergeable`, three platforms green, `tests/lock-rm-contention.test.mjs` fully green including the four "defines no second copy of the recovery judgment" assertions. Still Scott's approval from before the rebase — the diff shrank rather than grew, and the only substantive change since is restoring the `statSync` import described above.

santifer:
Your flake report earned its keep: a second contributor hit the identical failure the same day, and the pattern you both logged is now #3162, with a companion fix (#3163) that makes the guard name WHICH event fired (real escape vs failed realpath, with errno) so the next CI hit diagnoses itself. Exactly the outcome logging-instead-of-rerunning exists for. Your PR itself stays unaffected: the flake was in main's test, not your branch.

santifer:
Sharp work, @LayerC0de. Finding the third copy by consulting the guard's own protocolImplementors() list when only two were flagged is what made this complete: fixing 2 of 3 would have left the drift alive. And hardening the test from four-correct-copies to one-definition-plus-imports is the right direction; as your PR puts it, a repo that merely keeps its copies in step is one patch away from the drift that produced all three faces of #2777. Merged! 🚀

PR #3125, opened 2026-08-20
test: retry cleanup rmSync in the six remaining standalone suites

### LINKED ISSUE #3114: test: six standalone suites still call rmSync bare in cleanup — Windows EPERM aborts a green run

#3066 taught `test-all.mjs` to tolerate Windows EPERM during fixture cleanup (a just-exited child keeps a handle open briefly; antivirus widens the window), so a green run stops dying with a stack trace over an unlucky `rmSync`. Six standalone suites still call `rmSync` bare in their cleanup paths: `tests/application-artifacts.test.mjs`, `tests/batch-evaluate.test.mjs`, `tests/batch-tailor-flags.test.mjs`, `tests/batch-runner-score-delimiter.test.mjs`, `tests/browser-extract.test.mjs`, `tests/batch-runner-jd-prefetch.test.mjs`.

Mirror #3066's wrapper exactly (same semantics: EPERM/EBUSY tolerated on cleanup ONLY, every other error still thrown, assertions unaffected). One PR covering all six is fine.

**Acceptance:** no bare rmSync remains in those six files' cleanup paths; the suites still fail loudly on real assertion errors.

MannXo:
## What does this PR do?

Gives the six remaining standalone suites the same retrying `rmSync` that #3066 gave `test-all.mjs`, so a Windows EPERM during fixture cleanup stops aborting a run whose assertions all passed. The wrapper is exported once from `tests/helpers.mjs` rather than pasted into each file.

## Related issue

Fixes #3114

## Type of change

- [ ] Bug fix
- [ ] New feature

## Checklist

---

### One definition instead of six

The issue says to mirror #3066's wrapper exactly, and the wrapper here is byte-identical to it apart from the `export` keyword:

```
test-all.mjs:52      const rmSync = (target, opts = {}) => _rmSync(target, { maxRetries: 10, retryDelay: 100, ...opts });
tests/helpers.mjs:22 export const rmSync = (target, opts = {}) => _rmSync(target, { maxRetries: 10, retryDelay: 100, ...opts });
```

What I did not do is paste it into all six files. Five of the six already import `pass`/`fail` from `./helpers.mjs`, and six copies of a helper plus its rationale comment is the drift this repo already has an issue class for. Happy to switch to the literal per-file copy if you would rather keep the wrapper local, it is a small change.

One consequence worth naming: `test-all.mjs` keeps its own inline copy, so there are two definitions rather than one. Folding it into the shared export is a two-line change but it touches the core file and is outside what this issue asks, so I left it. Say the word for a follow-up.

### Semantics

`maxRetries`/`retryDelay` are defaults spread before `...opts`, so nothing is swallowed. Verified against the shared helper directly:

```
ok  non-EPERM still throws: ENOENT
ok  explicit opts pass through: force suppressed ENOENT
ok  recursive cleanup removes the tree
```

### Still fails loudly

Injecting a deliberate `fail()` into `browser-extract.test.mjs`:

```
📊 Results: 0 passed, 1 failed, 0 warnings
🔴 TESTS FAILED — do NOT push/merge until fixed
exit=1
```

Reverted, and that suite is back to 21 passed, 0 failed.

### Testing

`node test-all.mjs --quick` on this branch and on `main` at `61f8850`, same base, same flags:

```
branch  📊 Results: 4918 passed, 0 failed, 2 warnings
main    📊 Results: 4918 passed, 0 failed, 2 warnings
```

Full `node test-all.mjs` on the branch is 4919 passed, 0 failed, 2 warnings, exit 0. The extra assertion over the `--quick` run is `✅ Dashboard compiles`, which `--quick` skips.

Both warnings are pre-existing on `main` and unrelated: `cv-sync-check.mjs exited with error (expected without user data)` and `Possible personal data in dashboard/internal/ui/screens/stats.go: "santifer.io"`.

Also green: `cd dashboard && go test ./...` (all packages ok) and `node upgrade-tests.mjs --pr-gate` (GREEN).

### One thing I noticed, not touched here

`tests/batch-evaluate.test.mjs` ends with a bare `run();` on an async function that is never awaited, so in a full run its five assertions land roughly 1300 lines after its own heading, interleaved into whatever suite is running by then. They are counted, so nothing is lost, but `--only batch-evaluate` reports `0 passed, 0 failed` because the summary prints before the promise settles. Unrelated to this change and left alone.

## Summary by CodeRabbit

* **Tests**
  * Improved test cleanup reliability with automatic retries when removing temporary files and directories.
  * Standardized cleanup behavior across multiple test scenarios.
  * No changes to application functionality or test coverage.

### DIFF
--- a/tests/application-artifacts.test.mjs
+++ b/tests/application-artifacts.test.mjs
@@ -1,8 +1,9 @@
-import { existsSync, mkdtempSync, readFileSync, rmSync } from 'fs';
+import { existsSync, mkdtempSync, readFileSync } from 'fs';
 import { join } from 'path';
 import { tmpdir } from 'os';
 import { spawnSync } from 'child_process';
 import { fileURLToPath } from 'url';
+import { rmSync } from './helpers.mjs';
 import { applicationArtifactPaths, ensureApplicationArtifactDirs, slugifySegment, writeReuseDecision } from '../application-artifacts.mjs';
 import { repoRelativeManifestPath, workspaceRelativeManifestPath } from '../generate-pdf.mjs';
 
--- a/tests/batch-evaluate.test.mjs
+++ b/tests/batch-evaluate.test.mjs
@@ -1,6 +1,6 @@
-import { pass, fail, ROOT } from './helpers.mjs';
+import { pass, fail, rmSync, ROOT } from './helpers.mjs';
 import { processPipelineBatch, processOffer, PATHS } from '../batch-evaluate-gemini.mjs';
-import { mkdtempSync, mkdirSync, readFileSync, readdirSync, rmSync } from 'fs';
+import { mkdtempSync, mkdirSync, readFileSync, readdirSync } from 'fs';
 import { join } from 'path';
 import { tmpdir } from 'os';
 
--- a/tests/batch-runner-jd-prefetch.test.mjs
+++ b/tests/batch-runner-jd-prefetch.test.mjs
@@ -17,9 +17,9 @@
 //
 // Tests extract the real bash snippets from batch-runner.sh so the tests and
 // the implementation can never drift apart.
-import { pass, fail, getBash } from './helpers.mjs';
+import { pass, fail, rmSync, getBash } from './helpers.mjs';
 import { execFileSync, spawnSync } from 'node:child_process';
-import { readFileSync, writeFileSync, mkdtempSync, mkdirSync, existsSync, rmSync, mkdtempSync as _mdt } from 'node:fs';
+import { readFileSync, writeFileSync, mkdtempSync, mkdirSync, existsSync, mkdtempSync as _mdt } from 'node:fs';
 import { join, dirname } from 'node:path';
 import { tmpdir } from 'node:os';
 import { fileURLToPath } from 'node:url';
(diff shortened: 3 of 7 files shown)


santifer:
Clean sweep of #3114, @MannXo. Two things I liked on the read: the wrapper is byte-identical to the one test-all.mjs already applies to its own call sites, and it leans on Node's own retry class rather than swallowing errors, so a removal that keeps failing still throws and a real assertion failure is untouched. Exporting it from helpers.mjs is the right home: the six suites can't drift from it now. Merged 🚀

Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is all seven changed files, which are all under `tests/`.

Straightforward and correct. `helpers.mjs` gains a shared `rmSync` that wraps the `fs` one with `maxRetries: 10, retryDelay: 100`, and the six suites import it in place of the raw call. The wrapper spreads caller options last, so an explicit `force`/`recursive` still wins and only the retry defaults are added, and a removal that keeps failing still throws. Same shape #3066 gave test-all.mjs, now shared so the discovered suites cannot drift from it. Import-only at each call site, green on CI.

PR #3126, opened 2026-08-20
fix(update): stop misreading upstream's own content as a local edit (#3094)

### LINKED ISSUE #3094: update-system.mjs apply: the local-edit check baselines on `merge-base`, which the updater never advances, so from the second update on upstream's own changes are misread as local edits

### Code of Conduct

### Existing issues

- [X] I searched existing issues and this hasn't been reported yet.

Same **symptom** as #2817 (closed) and #2881 (open), but a different root cause. Both of those were diagnosed as CRLF/`.gitattributes` and fixed by #2820; that fix is present in the code below and the files still get flagged (ruled out in detail under *Not CRLF*). #3048's "Suspected root cause (1)" is this bug, left undiagnosed there (*"possibly comparing against the wrong base ref"*), though its main subject is a different failure. #3052 is unrelated (FETCH_HEAD resolving to a tag). #2337/#2448 are the feature this lives in.

### What happened?

`locallyModifiedSystemFiles()` decides whether a system file was edited locally by diffing it against `merge-base(HEAD, FETCH_HEAD)` (`update-system.mjs:910`), documented as *"the last commit this install shares with upstream, i.e. the baseline it was last synced to"* (`:859-860`).

`apply()` never performs an operation that advances that merge-base. It fetches, does a raw `git checkout FETCH_HEAD -- <paths>` (`:851-852`, *"a raw checkout, not a merge"*, and `:854`, *"NOT a merge, by design"*), and records the result with a plain `git commit` (`:1829`). A plain commit creates no ancestry link to the fetched commit, so `merge-base(HEAD, upstream)` stays pinned to the commit the install was **cloned** at, forever.

From the second update onward the baseline describes a state the install left long ago, and every file upstream changed in between reads as a local edit. Those files are preserved, a `.bak` is written for each, and they never receive the update.

Because `VERSION` is itself a system file, it gets preserved too. `apply()` re-reads it after the checkout (`:1778`) and reports the version it just failed to install:

```
Update complete: v1.25.0 → v1.25.0
```

Exit code 0. `check()` reads the same stale `VERSION`, so it keeps offering the update and `apply` never clears it. `update-system.mjs` is itself replaced normally: `apply` checks out the fetched updater and re-executes it (`:1447-1448`) *before* the local-edit check runs, so by the time the check happens the file already matches upstream and can never be flagged.

`--force` is the only way through, and it takes the upstream version of every flagged file, including any genuine local fix now mixed in with the false positives.

#### Not CRLF

In the reproduction below, on the run that flagged 83 files:

- the updater doing the flagging is v1.27.0 and **contains #2820's `--ignore-cr-at-eol`**
- `core.autocrlf` is unset; Linux, `git 2.43.0`
- **83 of the 83 flagged files are byte-identical to the upstream blobs the previous update installed.** For example `AGENTS.md` hashes to `c4a4195f65424c5bd248ce5f872115a152bc4b93` both on disk and at `ae1a92d:AGENTS.md`

These are real content differences against a stale baseline, not line-ending noise.

### Steps to reproduce

A local bare mirror stands in for the canonical repo so "upstream" can be advanced on demand; a repo-local `insteadOf` rule redirects the updater's hardcoded fetch URL to it. No source file is modified. Inputs are three public release commits.

```bash
#!/usr/bin/env bash
set -euo pipefail
WORK="${1:-$(mktemp -d)}"
UPSTREAM_URL="https://github.com/santifer/career-ops.git"
V124=35faacc885efef508e24677fd85990430250a436   # release 1.24.0
V125=ae1a92dd1a4d299e637ce5d96f18e79f743a50ba   # release 1.25.0
V127=fdda56138988873022bf0e1b42baa1fdef665494   # release 1.27.0

mkdir -p "$WORK"; cd "$WORK"
git clone -q --bare "$UPSTREAM_URL" source.git
git init -q --bare mirror.git
git --git-dir=source.git push -q mirror.git "$V124:refs/heads/main"

git clone -q mirror.git install && cd install
git config user.name "Repro User"; git config user.email "repro@example.com"
git config url."file://$WORK/mirror.git".insteadOf "$UPSTREAM_URL"
export CAREER_OPS_NPM_INSTALL_TIMEOUT_MS=1 \
       CAREER_OPS_PLAYWRIGHT_INSTALL_TIMEOUT_MS=1 \
       CAREER_OPS_DASHBOARD_REBUILD_TIMEOUT_MS=1

# update 1
git --git-dir="$WORK/source.git" push -qf "$WORK/mirror.git" "$V125:refs/heads/main"
node update-system.mjs apply

# update 2, same repo, still never edited
git --git-dir="$WORK/source.git" push -qf "$WORK/mirror.git" "$V127:refs/heads/main"
node update-system.mjs apply

git merge-base HEAD FETCH_HEAD          # -> 35faacc (clone commit)
git rev-parse --short "$V125"           # -> ae1a92d (what update 1 installed)
```

Output:

```
installed: 1.24.0        local edits: 0

########## UPDATE 1: 1.24.0 -> 1.25.0 ##########
Update complete: v1.24.0 → v1.25.0
Updated 282 system paths.
VERSION now: 1.25.0

########## UPDATE 2: 1.25.0 -> 1.27.0 ##########
local commits since clone (the updater's own, nothing else):
  4103dbb chore: auto-update system files to v1.25.0
uncommitted changes: 0

83 system file(s) differ from upstream because THIS install changed them:
  .claude-plugin/marketplace.json  (local copy saved: .claude-plugin/marketplace.json.bak)
  AGENTS.md  (local copy saved: AGENTS.md.bak)
  CHANGELOG.md  (local copy saved: CHANGELOG.md.bak)
  DATA_CONTRACT.md  (local copy saved: DATA_CONTRACT.md.bak)
  README.md  (local copy saved: README.md.bak)
  VERSION  (local copy saved: VERSION.bak)
  ... (83 total)
Update complete: v1.25.0 → v1.25.0
VERSION now: 1.25.0

########## VERDICT ##########
merge-base(HEAD, upstream) : 35faacc  <- frozen at the clone commit
expected (last synced)     : ae1a92d  <- what update 1 actually installed
.bak files written         : 83
VERSION on disk            : 1.25.0
VERSION upstream ships     : 1.27.0
```

The install was cloned and then updated twice. It has exactly one local commit, the updater's own, and a clean working tree.

#### Why it is gradual rather than immediate

The second condition at `:917` (`differs from FETCH_HEAD`) filters the flagged set down to files the update was actually going to change, which keeps the number plausible early on. It also means **every flagged file is one the update needed to install**, so the skip list is the update's payload.

Replaying the same arithmetic over public release history for an install that never edits anything (comparing blobs at each release; derived from git history, not a run of the tool) gives:

| update to | system files needing the update | of those, skipped |
|---|---|---|
| 1.17.0 | 89 | 0 |
| 1.18.0 | 186 | 36 |
| 1.20.0 | 84 | 76 |
| 1.24.0 | 264 | 189 |
| 1.27.0 | 384 | 294 |

The window between the frozen baseline and upstream only widens, so the longer an install has been updating, the larger the share of each update it silently drops.

### Expected behavior

A file that was never locally modified should never be reported as *"THIS install changed them"*, and `VERSION` should never be preserved as a local edit.

Whatever the mechanism, the baseline needs to be the commit the install was **last synced to**, rather than the last commit it happens to share ancestry with.

#### Possible directions

Two that seem workable, though there may well be better ones:

1. **Record the sync point explicitly.** On a successful `apply`, store the applied `FETCH_HEAD` SHA (a ref such as `refs/career-ops/last-sync`, or a small file) and baseline on that, falling back to `merge-base` when it's absent so existing installs keep today's behaviour on their next update. `rollback` would need to restore the previous value too, otherwise the next `apply` compares against a baseline that is too new and genuine local edits get overwritten silently.
2. **Give the update commit upstream as a second parent**, which makes `merge-base` accurate by construction and leaves `locallyModifiedSystemFiles()` untouched. Less code, but it changes the topology of the user's branch, possibly unwelcome for forks, and it has the same rollback caveat.

Worth noting that neither repairs an install that is already stuck: both only take effect from the next successful update onward, and for those installs the current update is still the broken one.

I may be missing a constraint that rules either of these out. Happy to implement whichever direction you prefer. The reproduction above works as a harness to verify a fix against, and I'll open a PR shortly unless you'd rather take a different approach.

### CLI tool

Not applicable. Reproduced with plain `node` against a clean clone; no CLI is involved in the failing path.

### OS and Node.js version

Ubuntu 24.04.4 LTS / Node v22.23.2 / git 2.43.0

### Output of npm run doctor

Omitted as not diagnostic here: the failing path is `apply()`'s git logic, and the reproduction runs against a throwaway clone with no personalization, so `doctor` there only reports the expected `onboardingNeeded: true` with `cv.md`, `config/profile.yml` and `portals.yml` missing. Happy to paste the full output if it would help.

brunosandrini98:
## What does this PR do?

`locallyModifiedSystemFiles()` decides a system file was edited locally by diffing it against `merge-base(HEAD, FETCH_HEAD)`, documented as the baseline the install was last synced to. `apply()` never advances that merge-base: it installs an update with a raw checkout plus an ordinary commit, and neither creates ancestry to the fetched commit, so it stays pinned to the commit the install was *cloned* at.

From the second update on, every file upstream changed since the clone differs from that baseline and is reported as a local edit. The update's own payload is preserved into `.bak` files and silently dropped — `VERSION` with it, so the run reports the version it just failed to install and `check()` keeps offering the same update forever.

This adds a third condition to the at-risk test: **a file is not at risk if its current content is content upstream itself published at that path.** Upstream's own bytes cannot have been authored here.

## Related issue

Fixes #3094

## Type of change

- [x] Bug fix
- [ ] New feature

## Why per file, and not by recording the sync point

Recording the applied `FETCH_HEAD` (a ref, a file, or a second commit parent) was the first thing I tried to design, and two findings ruled it out:

1. **It cannot repair any install that is already affected.** No install has such a record yet, so the first run after the fix still falls back to the merge-base and stays stuck. The self-re-exec at `:1447-1448` means a fix merged here reaches every install on its next `apply` — but only if it works on that first run, inside an install that is already in the broken state.
2. **A stuck install is a mixture of versions, so no single commit describes it.** Measured in the reproduction after the failed update:

   ```
   system files present in both releases : 786
     content == v1.25.0 only :  83   <- left behind
     content == v1.27.0 only : 122   <- updated
     identical in both       : 574
   ```

   A sync point would also be a lie about exactly the preserved files: they are excluded from the checkout by design, so "this install is now at X" is false for them, and the next update would compare a genuine local edit against a baseline that already holds upstream's newer version — the #2337 loss mode.

Baselining on the last auto-update commit has the same flaw from the other direction: preserved files are excluded from that commit, so a *committed* local edit that survives one update diffs clean against it and is silently overwritten on the next.

## Evidence

Reproduction as described in #3094: an install cloned at 1.24.0, updated to 1.25.0, then updated again. One stuck install, copied, with a genuine local edit to `merge-tracker.mjs` added before the final update. One copy updated to plain `main`, the other to `main` + this patch.

| | flagged as local edits | `VERSION` after | new `.bak` | local fix kept |
|---|---|---|---|---|
| control (`main`) | 122 | 1.25.0 — unchanged | 122 | yes |
| this PR | 1 | 1.28.0 | 1 | yes |

Control reports `Update complete: v1.25.0 → v1.25.0`. With the patch the single flagged file is `merge-tracker.mjs`, the one that was actually edited.

## Cost

Two `git` calls, run only on the files that survive the existing two conditions.

| | flagged | added |
|---|---|---|
| healthy install, nothing edited | 0 | none measurable — the query never runs |
| stuck install, 418 commits in the window | 122 → 1 | +868 ms |

## Behaviour changes worth stating

1. **A file deliberately reverted to an older upstream version is no longer reported**, and receives no `.bak`. Its content is upstream's own by definition, so it remains recoverable from history, and the Data Contract calls the system layer system-owned — but it is a real change and I would rather name it than have it found in review.
2. **The dropped files are silent**, matching the existing rule that a file upstream independently adopted is not worth warning about ("warning about it would be noise").
3. **This compensates for the stale baseline rather than moving it.** The merge-base stays frozen; the fix stops the wrong answer from mattering. The record of what was installed was never written down and cannot be recovered retroactively, so removing the need for one is the only option that helps existing installs. A sync point could be added later as a *speed* improvement — the search window grows slowly as an install ages — but not as the correctness fix.

## Tests

Three cases added to `tests/updater-local-system-edits.test.mjs`, in the existing style:

- **14** — content installed by a previous update is not a local edit. Fails on `main` with `["modes/pdf.md"]`.
- **15** — a genuine local fix is still reported after a second update, so the new condition cannot quietly cost us the warning the detector exists for. Fails on `main` with the false positive alongside the real one.
- **16** — unreadable upstream history (shallow clone) degrades the filter and still reports what was already known, matching the contract case 7 pins for a bad ref.

`node test-all.mjs`: 4921 passed, 0 failed. The 3 warnings are present on unmodified `main` in the same environment (no `cv.md` in a fresh clone, no Go compiler, `santifer.io` in `dashboard/internal/ui/screens/stats.go`).

## Checklist

- [x] If this is a new feature or architecture change, I opened an issue first (bug fixes exempt — this is a bug fix, and #3094 exists anyway)
- [x] My changes respect the Data Contract (no user-layer files touched)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of locally modified system files during updates.
  * Prevents previously installed upstream content from being incorrectly reported as local edits.
  * Preserves genuine local-change warnings.
  * Updates continue safely when upstream history cannot be read.

### DIFF
--- a/tests/updater-local-system-edits.test.mjs
+++ b/tests/updater-local-system-edits.test.mjs
@@ -55,6 +55,16 @@ function upstreamChange(repo, file, content) {
   repo.g('checkout', '-q', 'main');
 }
 
+/**
+ * Replay what apply() does with an update: check the content out of the
+ * upstream ref and record it with an ordinary commit. NOT a merge — that is
+ * the whole point of case 14, so this must stay a raw checkout.
+ */
+function replayUpdate(repo, version) {
+  repo.g('checkout', 'upstream', '--', ...PATHS);
+  repo.g('commit', '-qm', `chore: auto-update system files to v${version}`);
+}
+
 const PATHS = ['modes/', 'generate-cover-letter.mjs'];
 
 // ── 1. The reported case: a committed local fix upstream has not adopted ──
@@ -362,4 +372,80 @@ const PATHS = ['modes/', 'generate-cover-letter.mjs'];
     fail(`#13 expected ['generate-cover-letter.mjs'], got ${JSON.stringify(atRisk)}`);
   }
   rmSync(dir, { recursive: true, force: true });
+}
+
+// ── 14. The SECOND update: upstream's own last release is not a local edit ──
+//    Every case above models an install at its FIRST update, where the
+//    merge-base is still the commit the install was cloned at and therefore
+//    still describes it. apply() never advances that merge-base: it installs
+//    updates with a raw checkout plus an ordinary commit, neither of which
+//    creates ancestry to the fetched commit. So from the second update on, the
+//    baseline describes a state the install left behind, and every file
+//    upstream changed in between reads as a local edit — preserved, backed up
+//    to `.bak`, and never updated. `VERSION` is a system file too, so it is
+//    preserved along with the rest and the run reports the version it just
+//    failed to install (#3094).
+{
+  const repo = makeRepo();
+  upstreamChange(repo, 'modes/pdf.md', 'shipped pdf v2\n');
+  replayUpdate(repo, '2');
+  upstreamChange(repo, 'modes/pdf.md', 'shipped pdf v3\n');
+
+  const atRisk = locallyModifiedSystemFiles(PATHS, 'upstream', repo.ctx);
+  if (atRisk.length === 0) {
+    pass('content installed by a previous update is not a local edit (#3094)');
+  } else {
+    fail(`#14 expected [], got ${JSON.stringify(atRisk)}`);
+  }
+}
+
+// ── 15. ...and the fix must not cost us the warning it exists for ──
+//    Case 14 removes files from atRisk, so pin that it removes ONLY the ones
+//    upstream itself installed. A genuine local fix in a twice-updated install
+//    is exactly the #2337 case, and it has to survive the second update too.
+{
+  const repo = makeRepo();
+  upstreamChange(repo, 'modes/pdf.md', 'shipped pdf v2\n');
+  replayUpdate(repo, '2');
+  upstreamChange(repo, 'modes/pdf.md', 'shipped pdf v3\n');
+  writeFileSync(join(repo.dir, 'generate-cover-letter.mjs'), 'local linkedin fix\n');
+  repo.g('commit', '-qam', 'local fix');
+
+  const atRisk = locallyModifiedSystemFiles(PATHS, 'upstream', repo.ctx);
+  if (atRisk.length === 1 && atRisk[0] === 'generate-cover-letter.mjs') {
+    pass('a real local fix is still reported after a second update (#2337)');
+  } else {
+    fail(`#15 expected ['generate-cover-letter.mjs'], got ${JSON.stringify(atRisk)}`);
+  }
+}
+
+// ── 16. History we cannot read degrades the warning, never the update ──
+//    Same contract as case 7, one level down: the #3094 filter reads upstream
+//    history, and a shallow clone does not have it. A history query that fails
+//    must leave the detector reporting what it already knew rather than
+//    throwing — a warning we cannot compute must never abort the checkout.
+{
+  const repo = makeRepo();
+  upstreamChange(repo, 'modes/pdf.md', 'shipped pdf v2\n');
+  writeFileSync(join(repo.dir, 'generate-cover-letter.mjs'), 'local linkedin fix\n');
+  const blind = {
+    root: repo.dir,
+    git: (...args) => {
+      if (args[0] === 'log') throw new Error('shallow clone: no history here');
+      return repo.g(...args);
+    },
+  };
+
+  let threw = false;
+  let atRisk = null;
+  try {
+    atRisk = locallyModifiedSystemFiles(PATHS, 'upstream', blind);
+  } catch {
+    threw = true;
+  }
+  if (!threw && Array.isArray(atRisk) && atRisk.includes('generate-cover-letter.mjs')) {
+    pass('unreadable upstream history degrades the filter, not the update');
+  } else {
+    fail(`#16 threw=${threw} atRisk=${JSON.stringify(atRisk)}`);
+  }
 }
--- a/update-system.mjs
+++ b/update-system.mjs
@@ -1038,7 +1038,20 @@ export function prepareMaterializedSkillEntrypointsForStage(paths, root = ROOT) 
  *   2. it differs from the upstream ref. A local fix upstream has since adopted
  *      independently is byte-identical there, so the checkout costs nothing and
  *      warning about it would be noise — the exact case the #2337 reporter
- *      isolated when one of their two fixes survived an update.
+ *      isolated when one of their two fixes survived an update;
+ *   3. its content is not content upstream itself published at that path
+ *      (#3094). Conditions 1 and 2 assume the merge-base still describes this
+ *      install. It does not: apply() installs an update with a raw checkout
+ *      plus an ordinary commit, and neither creates ancestry to the fetched
+ *      commit, so the merge-base stays pinned to the commit the install was
+ *      CLONED at. From the second update on, everything upstream changed since
+ *      then differs from that baseline and reads as a local edit, so the
+ *      update's own payload is preserved into `.bak` files and silently
+ *      dropped — `VERSION` with it, which is why such a run reports the version
+ *      it just failed to install. Content upstream published cannot have been
+ *      authored here, and the question is asked per file because a stuck
+ *      install is a MIXTURE of versions: no single baseline commit describes
+ *      it, which is what rules out recording one.
  *
  * @param {string[]} paths - manifest entries (files or `dir/` prefixes).
  * @param {string} upstreamRef - ref being checked out, normally FETCH_HEAD.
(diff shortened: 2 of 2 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/updater-local-system-edits.test.mjs`; `update-system.mjs` is the updater, which is not mine to own, so I read it for signal and my approval speaks only to the test.

The three new cases are well-formed. Case 14 models the second update and asserts upstream's own previously-installed content is not reported as a local edit. Case 15 is the guard that matters most: a genuine local fix in a twice-updated install must still surface, so the new filter cannot be an over-broad silence. Case 16 pins the fail-open contract, an unreadable upstream history degrades the warning and never aborts the checkout. `replayUpdate` is defined above `const PATHS` but only reads it when called, after initialisation, so there is no temporal-dead-zone trap. Green on CI.

The decision on the updater change itself belongs with santifer; this is a green light on the test, not on the fix landing.

santifer:
Welcome to career-ops, @brunosandrini98, and this is a hell of an opening move. The diagnosis in #3094 is the hard part and you got it exactly: apply() installs with a raw checkout plus an ordinary commit, neither of which creates ancestry to the fetched commit, so the merge-base stays pinned at the clone and from the second update on the update's own payload reads as a local edit, gets preserved into .bak, and never lands. VERSION with it, which is why such a run reports the version it just failed to install.

Three things I checked and want on the record. Content-addressing is the right instrument: "has upstream ever shipped exactly these bytes at this path" is a set lookup over blob IDs, two git calls no matter how many files. Hashing the WORKTREE copy rather than the committed one is the detail I'd have missed, and it quietly kills the CRLF false-positive family too, since hash-object applies the same filters git would on the way in. And every failure path returns an empty set, so a shallow clone degrades the warning and never the update: case 16 pins that, which is the assertion I'd have asked for if it weren't already there.

Now the trade-off this buys, stated plainly because it belongs in the public record rather than in my head. If someone's local edit IS a revert to an older upstream version of that file, its blob is in the window, so it now gets filtered out of atRisk and overwritten with no warning and no .bak. That is a real regression of the #2337 protection for a narrow case, and I am merging anyway: today every install that updates twice is broken outright, and the alternative is leaving that in place while we design the discriminator. I've opened a follow-up for it. Merged 🚀

PR #3127, opened 2026-08-20
fix(providers/lever): fold categories.allLocations into location

hernanflores:
## What does this PR do?

`providers/lever.mjs` reads only `categories.location`, which Lever populates with a **single primary** city. Multi-location postings expose the full set in `categories.allLocations`, which the provider ignores — so a requisition open in both Barcelona and Montevideo looks Barcelona-only to `scan.mjs`'s `location_filter` and is silently dropped.

This folds `allLocations` into the location string, mirroring the `resolveLocation()` helper that `providers/remotli.mjs` already uses for the same problem.

## Related issue

None — bug fix, sent straight in per CONTRIBUTING.

## Why it matters

The failure mode leaves no trace. A filtered-out posting never surfaces, so there is nothing to notice as missing; the scan just quietly returns fewer roles than the board actually offers.

Measured against one live public Lever board (`api.lever.co/v0/postings/dlocal`, 53 open postings at the time of writing), postings whose location string contains "Montevideo" go from **2 to 29**. Among the ones currently invisible: Staff Engineer - Java, Head of Platform Engineering, Principal Security Engineer, DevOps Engineer CI/CD, Cloud Architect. Every one is LatAm-eligible while appearing to the scanner as Barcelona/Madrid/Spain only.

Reproducible against any Lever board with multi-location requisitions; nothing about it is specific to that company.

## Implementation

```js
function resolveLocation(categories) {
  const primary = typeof categories?.location === 'string' ? categories.location.trim() : '';
  const all = Array.isArray(categories?.allLocations)
    ? categories.allLocations.filter(l => typeof l === 'string' && l.trim()).map(l => l.trim())
    : [];
  const merged = [];
  for (const l of [primary, ...all]) {
    if (l && !merged.some(m => m.toLowerCase() === l.toLowerCase())) merged.push(l);
  }
  return merged.join('; ');
}
```

Same `; ` join, same case-insensitive dedup, and same tolerance for blank/non-string entries as `remotli.mjs` — deliberately copied rather than re-invented, so the two stay comparable.

Backward compatible: a posting with no `allLocations` yields exactly the previous single-location string.

## Tests

Four assertions added to `tests/providers/lever.test.mjs`, extending the existing `fetch()` normalization fixture:

- folds `allLocations` into `location` for a multi-location posting
- drops blank/non-string entries and case-insensitive duplicates of the primary
- resolves `allLocations` when `categories.location` is absent
- ignores a non-array `allLocations` and preserves the primary

`node test-all.mjs` on this branch: **4922 passed, 0 failed**.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Lever job listings now display all available locations, including additional locations provided for a posting.
  - Duplicate and invalid location entries are filtered, and multiple locations are presented in a consistent format.

- **Bug Fixes**
  - Improved location display for postings without a primary location or with incomplete location data.

- **Tests**
  - Added coverage for multi-location, duplicate, invalid, and missing-location scenarios.

### DIFF
--- a/providers/lever.mjs
+++ b/providers/lever.mjs
@@ -43,6 +43,25 @@ function resolveApiUrl(entry) {
   return `https://api.${host[1]}/v0/postings/${slug}`;
 }
 
+/** Fold `categories.location` together with any extra `categories.allLocations`
+ *  into one string. Lever puts a SINGLE primary city in `location`, and exposes
+ *  the full set on multi-location postings in `allLocations` — reading only the
+ *  former silently hides every other eligible location from scan.mjs's
+ *  location_filter (e.g. a req open in Barcelona AND Montevideo looks
+ *  Barcelona-only). Mirrors resolveLocation() in providers/remotli.mjs.
+ *  @param {any} categories */
+function resolveLocation(categories) {
+  const primary = typeof categories?.location === 'string' ? categories.location.trim() : '';
+  const all = Array.isArray(categories?.allLocations)
+    ? categories.allLocations.filter(l => typeof l === 'string' && l.trim()).map(l => l.trim())
+    : [];
+  const merged = [];
+  for (const l of [primary, ...all]) {
+    if (l && !merged.some(m => m.toLowerCase() === l.toLowerCase())) merged.push(l);
+  }
+  return merged.join('; ');
+}
+
 /** @type {Provider} */
 export default {
   id: 'lever',
@@ -66,7 +85,7 @@ export default {
       title: j.text || '',
       url: j.hostedUrl || '',
       company: entry.name,
-      location: j.categories?.location || '',
+      location: resolveLocation(j.categories),
       // Lever's v0 postings list ships the full description for free (same
       // payload, no per-job request) — enables scan.mjs content_filter.
       description: typeof j.descriptionPlain === 'string' ? j.descriptionPlain : '',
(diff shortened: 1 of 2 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/providers/lever.test.mjs`; `providers/lever.mjs` is outside my area and I read it for context.

The four new fixtures pin `resolveLocation` where it matters. I traced each: the multi-location row folds to `Barcelona; Madrid; Montevideo (Hybrid)`, the junk row drops blanks, non-strings and a case-variant duplicate of the primary to `Madrid; Sao Paulo`, an absent primary still resolves from `allLocations`, and a non-array `allLocations` is ignored with the primary preserved. The row count moves 3 to 7 so no earlier assertion silently shifts. The sentinel is real: the old `categories.location` behaviour returns `Barcelona` for row 3 and the test reddens. Green on CI.

santifer:
Welcome to career-ops, @hernanflores, and what a first PR to arrive with. The bug is the quiet kind: a req open in Barcelona AND Montevideo reported as Barcelona-only, so `location_filter` dropped it for a Montevideo candidate and nothing anywhere said a job had been hidden. I reproduced it rather than take your word for it: with `allow: ["Madrid"]`, `"Barcelona"` returns false and `"Barcelona; Madrid; Montevideo (Hybrid)"` returns true, which is exactly the difference your change makes. Two things I want to call out because they're what makes this mergeable in one pass: you mirrored the existing `resolveLocation()` in `remotli.mjs` instead of inventing a second shape, and your fixtures include the junk (non-strings, blanks, a case-variant dupe) rather than only the happy path. Merged, and glad to have you here 🚀

PR #3128, opened 2026-08-20
fix(templates): keep a project's tech line with the description it belongs to

MdRaf1:
## What does this PR do?

Every CV template deliberately lets a `.project` block split across a page boundary — cv-template.html says so in its own comment: *"Sections and multi-bullet roles are allowed to flow across a page boundary so content packs tight instead of jumping wholesale and leaving large bottom gaps."* That's the right default, but it's only safe while the worst split is forbidden, and one wasn't.

`.project-desc` and `.project-tech` are sibling blocks, so the gap between them is a legal break opportunity. When a project's description happens to end near the bottom of a page, Chromium takes it — and page 2 opens with a bare `TypeScript | github.com/user/repo` with no title or description above it to say which project it belongs to.

The symmetric case was already handled: `.project-title` carries `break-after: avoid` so a title is never orphaned at the *bottom* of a page. This adds the other half.

## Related issue

None — bug fix, sending it straight in per CONTRIBUTING.md.

## Type of change

- [x] Bug fix
- [ ] New feature

## The fix

```css
/* Keep a project's tech line attached to the description above it. */
.project-tech {
  break-before: avoid;
  page-break-before: avoid;
}
```

Applied to the seven templates `listTemplates('cv')` discovers. `resume-template.html` is **not** touched: it sets `break-inside: avoid` on `.project`, so that block already moves as a unit and the break opportunity never exists there.

### Why `break-before` on the tech line rather than `break-after` on the description

`break-after: avoid` on `.project-desc` fixes the same symptom, and I started there — but `<!--TECH_BLOCK-->` is conditional. For a project with a description and no tech line, a trailing `break-after: avoid` propagates to the end edge of the whole `.project`, gluing it to the next project's title. That over-constrains a case that was never broken. `break-before: avoid` on `.project-tech` forbids exactly the one offending opportunity and is inert when there is no tech line.

## Before / after

Reproduced with a synthetic payload (no personal data) tuned so the last project's description ends just at the page-1 boundary — first non-empty line of page 2, via `pdftotext -f 2 -l 2`:

| template | before | after |
|---|---|---|
| `cv-template.html` | `TypeScript \| github.com/janesmith/project-five` | `than.` → then the tech line |
| `cv-template.zh-minimal.html` | `TypeScript \| github.com/janesmith/project-five` | `than.` → then the tech line |

Chromium now breaks *inside* the description instead, so the tech line arrives with the tail of the prose it describes. **The page count does not change** (2 before, 2 after) — the templates' documented preference for packing content tight over relocating whole blocks is preserved. The other five render clean with no regression.

The orphan window is narrow: it took growing the final description one word at a time to land on it. That is exactly why it went unnoticed — it depends on a CV being an unlucky length, so it shows up on a real person's PDF and not on any fixture.

## Regression test

`tests/template-page-breaks.test.mjs` asserts the invariant over `listTemplates('cv')` rather than per file, so a template added later is covered without touching the suite. Six of the seven shipped without the guard; a new variant copied from any of them would have shipped without it too.

The invariant, not the diff: *if a template lets `.project` split, it must forbid a break immediately before `.project-tech`.* A template that keeps `.project` atomic passes without needing the rule, which is why `resume-template.html` staying unchanged is a pass and not an exemption.

Verified in both directions — against the templates as they were, `test-all.mjs` reports **7 failed, exit 1**; with the fix, **4923 passed, 0 failed**.

It's a CSS assertion rather than a render, because suites under `tests/` must run on a bare clone with only Node (#1440) and reproducing the orphan needs Chromium plus a payload tuned to one template's exact geometry. It also guards the two adjacent things an edit here could quietly undo: the paired legacy `page-break-*` alias, and the `break-after: avoid` on `.project-title`.

I left the `version:` in the template metadata headers alone — nothing in the repo's history bumps it, so I didn't want to invent a convention.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved printed CV pagination so project technology details stay with their preceding descriptions instead of appearing alone at the top of a new page.
  * Applied the pagination improvement consistently across all CV templates.

* **Tests**
  * Added regression coverage to verify page-break behavior across every HTML CV template.

### DIFF
--- a/templates/cv-template.compact.html
+++ b/templates/cv-template.compact.html
@@ -287,6 +287,13 @@ version: 1.0.0
     page-break-after: avoid;
   }
 
+  /* Keep a project's tech line attached to the description above it, so a bare
+     "Python | github.com/..." line is never stranded atop the next page. */
+  .project-tech {
+    break-before: avoid;
+    page-break-before: avoid;
+  }
+
   /* === RTL (Arabic) support === */
   html[lang="ar"] body {
     direction: rtl;
--- a/templates/cv-template.executive.html
+++ b/templates/cv-template.executive.html
@@ -278,6 +278,13 @@ version: 1.0.0
     page-break-after: avoid;
   }
 
+  /* Keep a project's tech line attached to the description above it, so a bare
+     "Python | github.com/..." line is never stranded atop the next page. */
+  .project-tech {
+    break-before: avoid;
+    page-break-before: avoid;
+  }
+
   /* === RTL (Arabic) support === */
   html[lang="ar"] body {
     direction: rtl;
--- a/templates/cv-template.html
+++ b/templates/cv-template.html
@@ -442,6 +442,15 @@
     page-break-after: avoid;
   }
 
+  /* Keep a project's tech line attached to the description above it.  A project
+     may still split across pages, but not at the one opportunity that strands a
+     bare "Python | github.com/..." line at the top of the next page with no
+     title or description to identify it. */
+  .project-tech {
+    break-before: avoid;
+    page-break-before: avoid;
+  }
+
   /* === RTL (Arabic) Support === */
   html[lang="ar"] body {
     direction: rtl;
--- a/templates/cv-template.jake.html
+++ b/templates/cv-template.jake.html
@@ -271,6 +271,13 @@ version: 1.0.0
     page-break-after: avoid;
   }
 
+  /* Keep a project's tech line attached to the description above it, so a bare
+     "Python | github.com/..." line is never stranded atop the next page. */
+  .project-tech {
+    break-before: avoid;
+    page-break-before: avoid;
+  }
+
   /* === RTL (Arabic) support === */
   html[lang="ar"] body {
     direction: rtl;
--- a/templates/cv-template.leadership.html
+++ b/templates/cv-template.leadership.html
@@ -288,6 +288,13 @@ version: 1.0.0
     page-break-after: avoid;
   }
 
+  /* Keep a project's tech line attached to the description above it, so a bare
+     "Python | github.com/..." line is never stranded atop the next page. */
+  .project-tech {
+    break-before: avoid;
+    page-break-before: avoid;
+  }
+
   /* === RTL (Arabic) support === */
   html[lang="ar"] body {
     direction: rtl;
--- a/templates/cv-template.modern.html
+++ b/templates/cv-template.modern.html
@@ -321,6 +321,13 @@ version: 1.0.0
     page-break-after: avoid;
   }
 
+  /* Keep a project's tech line attached to the description above it, so a bare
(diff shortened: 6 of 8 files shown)


MdRaf1:
Both review findings were real; fixed in 71705db.

**The `.project-tech`-not-styled skip was a vacuous pass.** I had reasoned "a template that styles no tech line cannot orphan one," which is wrong — the markup isn't the template's to opt out of. It comes from the shared `templates/sections/projects.html`, so every template emits that div whenever the payload carries `tech`, styled or not, and an unstyled block is still a block-level sibling a page break can land in front of. A future variant that emitted the line without styling it would have passed the guard while still carrying the bug. Skip removed; every discovered template is now held to the rule.

**`rules()` claimed declarations never contain braces. They don't, except they do.** `max-width: {{PAGE_WIDTH}};` sits inside the style block of every template. I checked what it actually cost before changing anything: the `.page` rule was dropped entirely and the first rule's selector absorbed `— CV</title> <style> :root`. Both are near the top of the file, the `.project-*` rules are ~200 lines below and parsed cleanly, which is why the suite still failed pre-fix and passed post-fix. So the assertions were sound, but the comment was false and the helper was one template edit away from mattering. Placeholders are now stripped before parsing.

Both directions re-verified after the change: **7 failed / exit 1** against the templates as they were, **4923 passed / 0 failed** with the fix.

I kept the `atomicProject` branch even though nothing in `listTemplates('cv')` currently takes it. Keeping a whole `.project` atomic is the other legitimate way to satisfy the invariant — it's how `resume-template.html` already avoids this bug — so a future variant choosing that route should pass on its merits rather than trip a rule written for the splitting case.

Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is the new `tests/template-page-breaks.test.mjs`; the seven template files are outside my area and I read them for context.

Good instinct to pin this as an invariant over every discovered template rather than seven copies of one assertion, since the failure is invisible until a CV happens to be the wrong length. The check is honest about being a CSS contract rather than a render, and it says why (the suites must run on a bare clone). It strips comments before matching so a commented-out rule cannot satisfy it, accepts either an atomic `.project` or a forbidden break before `.project-tech`, pairs the modern property with its legacy alias so half the pair cannot be dropped quietly, and keeps a regression guard on the existing `.project-title` protection. Six of the seven templates shipped without the guard, so the test reddens if the CSS is reverted. Green on CI.

santifer:
The invariant-over-discovered-templates shape is the part that makes this stick, @MdRaf1: six of seven templates shipped without the guard and any new variant would have copied the gap, so pinning it via listTemplates() closes the class, not the instance. And stripping comments so a commented-out rule can never satisfy the assertion is exactly the blind-guardian lesson applied. Merged! 🚀

PR #3134, opened 2026-08-20
fix(web-logo): companyCacheKey's [^a-z0-9] strip collided distinct non-ASCII company names

Zoubeir23:
`web/src/lib/core/logo-cache-key.mjs`'s `companyCacheKey` normalized with `.toLowerCase().replace(/[^a-z0-9]/g, "")` before hashing — the exact ASCII-only pattern `normalize-text-key.mjs`'s own header comment names by number and warns against reinstating: *"Never reinstate `[^a-z0-9]`: that deleted every non-ASCII letter ('Škoda'→'koda', '日本電産'→''), which both suppressed real offers and resurfaced evaluated ones"* (#2369/#2666). This file reintroduced the same defect independently, in a different feature (logo caching, not tracker/offer dedup).

**Why it's not just cosmetic here.** This module's own docstring states requirement 2 in as many words: *"It must not collide. Two different companies sharing a key means one of them wears the other's logo permanently, with no expiry to correct it."* The logo cache is write-once, never revalidated — a hit or a miss, once written, is served forever.

**Verified by execution:**

```
companyCacheKey("Škoda")      -> co_v2_koda_74b8796bbe
companyCacheKey("Koda")       -> co_v2_koda_74b8796bbe
same key: true

companyCacheKey("日本電産")    -> null   (CJK erased to nothing — never cacheable at all)
```

Two real, distinct companies (Škoda / Koda) share one on-disk cache entry — whichever resolves first silently wins the logo for both, permanently, per the module's own no-expiry design. CJK company names weren't even miscacheable, they were **uncacheable**: `normalized` came back `""`, `companyCacheKey` returned `null`, and `route.ts` treats that as "bad company" (never reaches the cache at all).

## The fix

Delegates to `normalizeTextKey` (`./normalize-text-key.mjs`), already sitting in the same directory, already parity-tested against the core, already the sanctioned fix for exactly this bug class — rather than inventing a second lossy normalizer. `companyCacheKey` only consumes this server-side (`app/api/logo/route.ts`, `runtime = "nodejs"`), so there's no client-bundle constraint pulling toward a hand-rolled ASCII filter here either.

**Filesystem safety is unaffected**, just checked differently: `normalizeTextKey`'s `\p{L}\p{M}\p{N}` allowlist excludes `.`, `/`, `\`, spaces, and NUL just as completely as `[^a-z0-9]` did — traversal characters were never in the letter/mark/number Unicode categories to begin with. The existing `logo-cache-key.test.mjs` hostile-input test (`café/../../root`, `..\..\windows\system32`, embedded NUL, etc.) still passes; I widened its regex from `[a-z0-9]` to `\p{L}\p{N}` since a real accented/CJK company name now legitimately produces a Unicode prefix, and the test's actual safety property was never "ASCII-only," it was "no traversal/separator/control characters" — which `\p{L}\p{N}` guarantees regardless of script.

**Also bumps `COMPANY_KEY_VERSION` to `v3`.** The file's own requirement 1 is explicit about why: *"It must change when candidate generation changes. Otherwise a machine that cached a miss under the old resolver keeps serving that miss forever."* Any install with a warm cache may currently have a `co_v2_koda_...` entry poisoned by whichever of Škoda/Koda resolved first. Without a version bump, "Koda" (no diacritics, so its v2 and v3 normalized forms are identical) would keep reading that same key path and keep serving whatever's there. Bumping discards it cleanly rather than leaving it reachable.

Two-commit shape: the fix, then the tests (widened the existing safety-regex assertion, added direct collision + CJK-cacheability coverage).

## Test plan
- [x] `cd web && npm test` — 331 passed, 0 failed
- [x] All 8 pre-existing `logo-cache-key.test.mjs` tests still pass, including the widened filesystem-safety regex against the same hostile-input corpus
- [x] New tests confirm Škoda/Koda and two distinct CJK names no longer collide, and CJK names are cacheable (non-null) again

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved company logo caching for names containing accented, non-ASCII, CJK, and other Unicode characters.
  * Preserved distinctions between international company names that previously could produce identical cache entries.
  * Ensured long or complex company names remain safely and consistently cacheable.
  * Continued blocking unsafe characters, path separators, spaces, and control characters in cache keys.

### DIFF
--- a/web/src/lib/core/logo-cache-key.mjs
+++ b/web/src/lib/core/logo-cache-key.mjs
@@ -1,4 +1,5 @@
 import { createHash } from "node:crypto";
+import { normalizeTextKey } from "./normalize-text-key.mjs";
 
 // Cache identity for name-resolved company logos (see app/api/logo/route.ts).
 //
@@ -13,10 +14,19 @@ import { createHash } from "node:crypto";
 //     them wears the other's logo permanently, with no expiry to correct it.
 
 /** Bump whenever `companyDomains()` in app/api/logo/route.ts changes which
- *  domains it tries, or in what order. The version is part of every key, so
- *  bumping it makes entries written by the previous resolver unreachable.
- *  v2: curated brand domains (notion.so, zoom.us, …) now precede slug guesses. */
-export const COMPANY_KEY_VERSION = "v2";
+ *  domains it tries, or in what order — OR whenever the key derivation itself
+ *  changes, per requirement 1 above: an unchanged version leaves any entry
+ *  poisoned under the old derivation permanently reachable, with nothing to
+ *  correct it. The version is part of every key, so bumping it makes entries
+ *  written by the previous resolver unreachable.
+ *  v2: curated brand domains (notion.so, zoom.us, …) now precede slug guesses.
+ *  v3: normalization switched from `[^a-z0-9]` to the Unicode-safe
+ *  normalizeTextKey (#2369/#2666's own lesson, reintroduced here independently)
+ *  — v2 stripped every non-ASCII letter, so "Škoda" and "Koda" hashed to the
+ *  IDENTICAL key and one silently wore the other's logo forever (requirement 2,
+ *  violated). Bumping discards any v2 entry poisoned this way rather than
+ *  leaving it reachable under an unchanged "koda" key. */
+export const COMPANY_KEY_VERSION = "v3";
 
 /** Cache key for a company name, or null if the name carries nothing to key on.
  *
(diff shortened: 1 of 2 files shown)


santifer:

Zoubeir23:

santifer (approved):
Approving — and you respected both invariants #2942 baked into this file, which most contributors would have missed one of. The `[^a-z0-9]` strip collapsed every non-Latin name to the same empty-ish key, so two distinct CJK/Cyrillic employers shared one cached logo permanently. Importing the core's `normalizeTextKey` instead of hand-stripping is the right fix (one normalizer, and it's Unicode-aware), and bumping `COMPANY_KEY_VERSION` v2→v3 is the part that makes it safe: a changed key derivation with an unchanged version would leave every entry cached under the old strip permanently reachable — poisoned, with no expiry to correct it. You bumped it and said why in the comment. Merging when I have the field.

santifer:
Thanks, @Zoubeir23 — you respected both invariants this file carries, which is the hard part. The `[^a-z0-9]` strip collapsed every non-Latin name to the same empty-ish key, so two distinct CJK/Cyrillic employers shared one cached logo forever. Importing the core's Unicode-aware `normalizeTextKey` instead of hand-stripping is the right fix, and bumping COMPANY_KEY_VERSION v2→v3 is what makes it safe — a changed derivation under an unchanged version would leave every old-strip entry permanently reachable. You did both and said why. Merged 🚀

PR #3135, opened 2026-08-20
fix(web): regenerate the nested web/ lockfile on postcss bumps

### LINKED ISSUE #3110: renovate: nested web/ lockfile is never regenerated — every web dependency bump fails npm ci with EUSAGE
Diagnosed on #2654: renovate bumped `web/package.json` (next 16.3.1, shaders 0.0.80) but its `renovate/artifacts` step failed ("Artifact file update failure") and never regenerated `web/package-lock.json`. The `web typecheck + build` check then dies in `npm ci` with EUSAGE ("package.json and package-lock.json not in sync") before any code runs — so the red is infrastructure, not the dependency. Every future web/ bump will fail identically until this is fixed.

Repo config today: `renovate.json` is minimal (config:recommended + groups), nothing about the nested npm workspace under `web/`.

Things to evaluate for the fix (whoever takes it, verify against renovate docs rather than trusting this list):
- `constraints` (web/package.json declares `node >=22`; if renovate's runner resolves an older node, the nested install can fail exactly like this)
- whether the repo needs `web/` registered so the artifact step knows a second package file + lockfile pair exists
- `postUpdateOptions` relevance for npm lockfile regeneration

Immediate mitigation applied on #2654: lockfile regenerated manually with `npm install --package-lock-only` and pushed to the renovate branch. That unblocks one PR; this issue is about not needing to do it again. The real acceptance test is the NEXT renovate web/ bump going green on its own.

obielin:
## What does this PR do?
`web/package.json`'s `overrides.postcss` (`"^8.5.18"`) duplicated the direct `devDependencies.postcss` version as a plain string. npm's `EOVERRIDE` check requires those to match exactly or use the self-reference syntax, so any bump the override didn't mirror broke `npm ci` — which is what happened in #2654 and is blocking every future `web/` dependency bump. This changes the override to `"$postcss"`, npm's built-in self-reference syntax, so it can never drift out of sync again.

## Related issue
Closes #3110

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

---

## Root cause
@santifer already diagnosed this in a comment on #2654 — `web/package.json:41` pinned `"postcss": "^8.5.18"` in `overrides` while `postcss` is also a direct `devDependency`. npm rejects that combination unless the override references the dependency, so `npm install --package-lock-only` fails with `EOVERRIDE` and the lockfile never regenerates. The `web typecheck + build` CI check then dies in `npm ci` with `EUSAGE` before any code runs.

## Fix
One line: `"postcss": "^8.5.18"` → `"postcss": "$postcss"` in `web/package.json`'s `overrides`, plus the regenerated `web/package-lock.json`.

## Why the issue's original three leads don't apply
This issue's body still lists three things to evaluate (`constraints`, `web/` registration, `postUpdateOptions`) from before the root cause above was pinned down on #2654. Checked each against the actual failure:

- **`constraints`**: the observed error is `EOVERRIDE`, an npm dependency-resolution conflict — nothing to do with Node-version/toolchain resolution, so this wasn't the cause here.
- **`web/` registration**: not needed — Renovate already discovers and proposes bumps for `web/`-only packages (six of them, in [#2654](https://github.com/santifer/career-ops/pull/2654)'s PR body), proving the npm manager already finds `web/package.json` + its lockfile without extra config.
- **`postUpdateOptions`**: checked the full option enum (`npmDedupe`, `npmInstallTwice`, `pnpmDedupe`, etc.) — none address discovering or regenerating a second lockfile; not applicable to this failure class.

No `renovate.json` changes in this PR — nothing in it is implicated.

## Verification
- Reproduced the exact failure locally: bumped `postcss` in `devDependencies` while leaving the override stale, got the identical `EOVERRIDE` error npm produced in [#2654](https://github.com/santifer/career-ops/pull/2654).
- Applied the fix, reran `npm install --package-lock-only` — clean install, 0 vulnerabilities, no error.
- `npm ci` (the exact command CI runs and the one that was failing) — clean.
- `npm run typecheck` — clean, no errors.
- `npm run build` — compiles successfully, full route table generated. (5 pre-existing Turbopack warnings about dynamic filesystem tracing in unrelated route handlers — not touched by this change.)
- `node test-all.mjs` (full suite) — 4320 passed, 0 failed.
- Verified end-to-end on my own fork: applied the fix, then let Renovate itself process the same `postcss` bump that broke [#2654](https://github.com/santifer/career-ops/pull/2654). It succeeded — both `web/package.json` and `web/package-lock.json` changed (versus [#2654](https://github.com/santifer/career-ops/pull/2654)'s `package.json`-only diff), with no "Artifact update problem" comment, and `postcss` resolved to `8.5.26` in the lockfile.

## Scope
Only `web/package.json`'s `overrides.postcss` entry and the regenerated `web/package-lock.json`. No `renovate.json` changes, no other dependency version changes.

## Chores

* Updated PostCSS version alignment to match the declared project dependency.


santifer:

santifer:
Merged. It kills the drift between the devDependency and the override, which were pinned twice to the same value and could quietly diverge on the next bump. `$postcss` changes where the value comes from, not the value itself, which is why the lockfile stays untouched: I confirmed that after updating your branch onto current main. Reviewed by the web owner, executed by me, and sorry for the wait.

PR #3137, opened 2026-08-20
fix(providers): retire echojobs — feed is gone behind bot protection

### LINKED ISSUE #2976: providers/echojobs.mjs: EchoJobs closed the undocumented /api/jobs endpoint the provider relied on — no replacement, no announcement
> ⚠️ **Privacy first.** career-ops runs locally. Never paste your CV, scan results, pipeline data, job listings, or other personal info — issues are public forever. Keep your data on your machine; share only sanitized snippets.

### What happened?

The `echojobs` provider (`providers/echojobs.mjs`) fetches `https://echojobs.io/api/jobs` as a paginated JSON feed. That endpoint no longer works: EchoJobs rebuilt their frontend (Next.js/RSC) and their `robots.txt` now explicitly reads:

```
User-Agent: *
Allow: /
Disallow: /auth
Disallow: /settings
Disallow: /api
Disallow: /unsub_job_alert
```

A request to the feed URL now returns the site's HTML shell (200 OK) instead of `{ jobs: [...] }`, which makes `ctx.fetchJson` throw:

```
echojobs: unexpected API response on page 1 — expected { jobs: [...] }, got keys: [...]
```

**This isn't a versioned-API breakage — there's nothing to migrate to.** I checked for any notice or replacement before filing:
- Their blog (`sitemap-blog.xml`) is all salary-guide/career-advice content — no changelog, no engineering notes, nothing about API changes.
- Their `/about` page's only mention of "API" is the founder's origin story about building a *personal* tool for himself ("why not apply my engineering skills to create an API that would return only jobs that matched my interests") — there's no developer documentation and no stated third-party/programmatic access policy.
- No alternate JSON endpoint found (`/api/v1/jobs`, `/api/jobs.json` both 404; sitemap has nothing else relevant).

So `echojobs.io/api/jobs` was most likely always an **undocumented internal endpoint** their old frontend happened to call, which the provider was built against — not a published public API that got deprecated with notice. `robots.txt` disallowing `/api` now reads as them locking down internals during the rebuild, not an announcement aimed at API consumers. Worth naming that distinction precisely, since it changes what "fixing" this could even mean — there's no versioned replacement to port to, unlike prior cases (see below).

**Related but distinct precedent:**
- `providers/jobstreet.mjs` and `providers/arbeitsagentur.mjs` (`a713ed5d fix(arbeitsagentur): move to the v6 Jobsuche API`, #2494) both had their APIs break and got migrated to a replacement endpoint. That's the established playbook for "API version changed" — it doesn't apply here since no replacement exists.
- #1744 (closed/shipped) is what's actually generating the "unreachable for 3+ runs" streak warning via `data/portal-health.tsv` — its own design doc names `verify-portals.mjs` / re-slugging as *the* fix. That's the gap this issue hits: `verify-portals.mjs` only sweeps `tracked_companies` (Greenhouse/Ashby/Lever slugs), not `job_boards` aggregator entries, so a user following the suggested next step for an EchoJobs-class failure hits a dead end.
- #2644 (open) proposes turning portal-health streak data into prune recommendations ("re-find the slug or remove the entry"). This issue is a concrete instance of the failure class that feature would eventually need to handle for `job_boards` entries too, not just `tracked_companies` — worth a cross-reference for whoever picks either issue up.

Checked this against the current release (v1.26.0, compared against v1.25.0 via the GitHub compare API) before filing: `providers/echojobs.mjs`, `verify-portals.mjs`, and `portal-health-lock.mjs` are all untouched between the two versions, and the `scan.mjs`/`templates/portals.example.yml` changes in v1.26.0 are unrelated (Unicode dedup fixes, CLI flag parsing, and new provider examples for join.com/Getro/Consider/joinup.ch). This is still current.

### Steps to reproduce

1. Enable the `echojobs` job board entry (commented out by default in `templates/portals.example.yml`, along with the rest of the `job_boards:` block — so this requires a user to deliberately opt in).
2. Run `node scan.mjs`.
3. Observe the provider errors out / gets flagged unreachable after repeated runs.
4. Run the suggested `node verify-portals.mjs` — it doesn't check `job_boards` entries at all, so it reports nothing about EchoJobs.

Direct repro without the scanner:
```bash
curl -s "https://echojobs.io/api/jobs?per_page=1&page=1"   # returns HTML, not JSON
curl -s https://echojobs.io/robots.txt                      # Disallow: /api
```

### Expected behavior

Either the provider works, or the failure is surfaced clearly and users aren't pointed at a diagnostic tool (`verify-portals.mjs`) that can't actually see this class of entry. Suggestions:

1. Mark `providers/echojobs.mjs` deprecated (or have `scan.mjs` detect a `robots.txt`-style HTML-instead-of-JSON response and report "provider appears discontinued" rather than a generic network-error retry warning) — and note the deprecation next to the commented-out example in `templates/portals.example.yml` so someone weighing whether to uncomment it isn't caught by surprise.
2. Extend `verify-portals.mjs` to also sanity-check `job_boards` entries (see #1744/#2644 above), so the tool `scan.mjs` already recommends is actually useful for this failure class.

Not proposing scraping around the `Disallow: /api` directive — the right fix is to stop relying on an endpoint that was never a supported public API, not find a workaround.

### CLI tool

Claude Code

### OS and Node.js version

macOS 26.6.1 / Node v26.0.0

### Output of `npm run doctor`

N/A — this is a provider/data-source issue, not a local setup issue.

clede:
## Summary
- `echojobs.io/api/jobs` (and its `robots.txt`) now answer with a Vercel bot-protection checkpoint (HTTP 429). Per the discussion in #2976, this is not a provider to repair — career-ops does not work around bot protection, and either the checkpoint or the disallowed `/api` path settles it.
- `providers/echojobs.mjs`'s `fetch()` now throws immediately with a message naming the cause, instead of hitting the checkpoint and surfacing a raw `expected { jobs: [...] }` parse error that gives a user no way to tell "the site changed" from "I configured it wrong".
- `detect()` still matches `provider: echojobs` on purpose, so anyone with a stale `portals.yml` entry gets this explanation rather than a silent "no provider matched".
- `docs/SUPPORTED_JOB_BOARDS.md` marks the board **Retired (#2976)** with the same reason.
- `normalizeEchojobsJob()` and its unit tests are kept as-is — a record of the feed's shape — even though `fetch()` no longer reaches it. The pagination-specific tests are replaced with one asserting `fetch()` throws the retirement message without making any network call.

Closes #2976.

## Test plan
- [x] `node tests/providers/echojobs.test.mjs` — all pass
- [x] `node test-all.mjs` on a clean worktree off `upstream/main` — 4986 passed, 0 failed, 3 pre-existing unrelated warnings

## Summary by CodeRabbit

* **Bug Fixes**
  * Retired the EchoJobs job board integration because its feed is unavailable due to bot protection.
  * Configurations using EchoJobs now fail immediately with a clear explanatory message instead of attempting requests.

* **Documentation**
  * Updated supported job board documentation to mark EchoJobs as retired and instruct users to remove it from `portals.yml`.


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/providers/echojobs.test.mjs`; the provider is outside my area and I read it for context.

The retirement is tested where it matters: `fetch()` throws, the message names the cause (`#2976`), and the mocked transport asserts zero network calls, so a stubbed `fetchJson`/`fetchText` that got called would fail the check. `normalizeEchojobsJob` is kept and still unit-tested as a record of the feed shape. The sentinel is real, a `fetch()` that still paginated would trip the network counter or miss the message. Green on CI.

santifer:
Clean retirement, @clede: verified the 429 checkpoint myself before merging, and keeping detect() matching so a stale portals.yml gets an explanation instead of a silent no-match is exactly the right degradation. Merged! 🚀

PR #3138, opened 2026-08-20
fix(web-tracker-lock): withTrackerLock reported a filesystem failure as transient lock contention

Zoubeir23:
`web/src/lib/core/tracker-lock.ts`'s `withTrackerLock` had an unconditional acquire-failure catch:

```ts
try {
  lock = await acquire(lockDirFor(trackerFile), { ... });
} catch {
  throw new TrackerBusyError();
}
```

`TrackerBusyError` means "tracker is being written by another process" — a route answers 409 and the caller is expected to retry. But this catch converted **every** error from `acquire()` into that message, not just genuine contention.

The core it wraps, `tracker-utils.mjs`'s `acquireTrackerLock`, deliberately tags only the real timeout case:

```js
// Tag the timeout so callers can tell "lock is busy, retry later" apart
// from filesystem/configuration failures rethrown out of the loop above.
const timeoutErr = new Error(`Timed out waiting for tracker lock at ${lockDir}`);
timeoutErr.code = 'LOCK_TIMEOUT';
throw timeoutErr;
```

Anything else — an `ENOENT` from a missing lock-dir parent, `EACCES`, a broken checkout — is rethrown untagged. `tracker-lock.ts` wasn't checking the tag, so a real, permanent failure got reported as "retry shortly," which will never resolve by retrying.

**Verified by execution** against the real core function:

```
$ node --input-type=module -e "
import { acquireTrackerLock } from './tracker-utils.mjs';
try {
  await acquireTrackerLock('/nonexistent-parent-dir-xyz-123/lockdir', { timeoutMs: 2000, retryMs: 50 });
} catch (err) {
  console.log('threw:', err.code, '|', err.message);
  console.log('is LOCK_TIMEOUT?', err.code === 'LOCK_TIMEOUT');
}"
threw: ENOENT | ENOENT: no such file or directory, mkdir '/nonexistent-parent-dir-xyz-123/lockdir'
is LOCK_TIMEOUT? false
```

**Why this survived review:** its sibling `web/src/lib/core/followups-lock.ts` — same architecture, same era, explicitly cross-referenced in both files' own header comments as "same failure family, same cure" — already gets this right:

```ts
const err = e as { name?: string; code?: string } | null;
if (err && err.name === "SeedError" && err.code === "LOCK_TIMEOUT") {
  throw new FollowupsBusyError();
}
throw e;
```

`tracker-lock.ts` just never picked up the same discrimination. The fix matches that pattern, adapted for `tracker-utils.mjs`'s plain `Error` + `.code` tag (no `SeedError` name to check here).

**Current reachability, stated plainly rather than left implicit:** `withTrackerLock` is currently unreachable from any live route. #2903 introduced it for `POST /api/status`, but that route was since refactored (see its own header comment) to delegate to `set-status.mjs` via `child_process` instead, which holds the tracker lock core-side — it no longer imports `tracker-lock.ts` at all. I grepped every route under `web/src/app/api` that touches the tracker (`status`, `tracker/delete`, `run`, `report/shape`, `assistant`, `whats-new`) and all delegate to a core CLI subprocess rather than calling `withTrackerLock` directly. So this is a latent-bug fix in an exported, documented, still-referenced-by-its-sibling's-comments utility — not a live-traffic bug today. Worth fixing anyway: it's a landmine for whoever wires this back up (the file's own extensive JSDoc reads like it's meant to be used again), and the asymmetry with its actively-used sibling is exactly the kind of thing a future contributor would copy the wrong half of.

## Test plan
- [x] `cd web && npm test` — 332 passed, 0 failed
- [x] New test file source-text-extracts the catch block (`tracker-lock.ts` imports the `@/` alias, so — same constraint and same workaround as `pipeline-local-today.test.mjs` — it can't be imported directly) and confirms the bare-catch-all shape is gone, the `LOCK_TIMEOUT` discriminator is present, and non-timeout errors propagate
- [x] A live test against the real `acquireTrackerLock` pins the underlying contract this fix depends on: a genuine `ENOENT` is never tagged `LOCK_TIMEOUT`
- [x] Confirmed the regression-guard regex matches the pre-fix source (would have caught this)

## Summary by CodeRabbit

* **Bug Fixes**
  * Lock acquisition now reports only genuine timeout conditions as tracker-busy errors.
  * Filesystem, configuration, and other lock acquisition failures retain their original error details.

* **Tests**
  * Added coverage for accurate lock-error classification, propagation, and permission-failure handling.

### DIFF
--- a/web/src/lib/core/tracker-lock.ts
+++ b/web/src/lib/core/tracker-lock.ts
@@ -85,10 +85,23 @@ export async function withTrackerLock<T>(trackerFile: string, fn: () => T | Prom
     // The core tags ONLY the genuine-contention timeout with code LOCK_TIMEOUT
     // (tracker-utils.mjs's own comment: "so callers can tell 'lock is busy,
     // retry later' apart from filesystem/configuration failures rethrown out
-    // of the loop above"). A bare catch here converted EVERY error — ENOENT
-    // from a missing lock-dir parent, EACCES, a broken checkout — into
+    // of the loop above"). A bare catch here converted EVERY error — including
+    // ENOENT from a missing lock-dir parent, a broken checkout — into
     // "tracker is being written by another process, retry", which is false:
     // those never resolve by retrying. Anything not tagged propagates as-is.
+    //
+    // EACCES/EPERM is its own case, not "propagates as-is" like ENOENT: the
+    // core's isMkdirContention (pipeline-lock.mjs) treats EACCES/EPERM as
+    // contention too, because on Windows a lock dir mid-create/mid-remove by
+    // another process answers with exactly those codes (#2777) — there is no
+    // way to tell that apart from a genuine permission problem by code alone.
+    // So a PERSISTENT EACCES/EPERM (not another writer, just no permission)
+    // still retries inside the core's loop and, after timeoutMs, surfaces as
+    // LOCK_TIMEOUT — which this catch then reports as TrackerBusyError, same
+    // as real contention. That is a known, accepted characteristic of the
+    // shared lock machinery (see the PR discussion), not a gap in this catch:
+    // fixing it means changing isMkdirContention itself, which every lock in
+    // the repo shares, not something scoped to this wrapper.
     const err = e as { code?: string } | null;
     if (err && err.code === "LOCK_TIMEOUT") {
       throw new TrackerBusyError();
--- a/web/src/lib/core/tracker-lock.ts
+++ b/web/src/lib/core/tracker-lock.ts
@@ -81,8 +81,19 @@ export async function withTrackerLock<T>(trackerFile: string, fn: () => T | Prom
       retryMs: 50,
       tracker: trackerFile,
     });
-  } catch {
-    throw new TrackerBusyError();
+  } catch (e) {
+    // The core tags ONLY the genuine-contention timeout with code LOCK_TIMEOUT
+    // (tracker-utils.mjs's own comment: "so callers can tell 'lock is busy,
+    // retry later' apart from filesystem/configuration failures rethrown out
+    // of the loop above"). A bare catch here converted EVERY error — ENOENT
+    // from a missing lock-dir parent, EACCES, a broken checkout — into
+    // "tracker is being written by another process, retry", which is false:
+    // those never resolve by retrying. Anything not tagged propagates as-is.
+    const err = e as { code?: string } | null;
+    if (err && err.code === "LOCK_TIMEOUT") {
+      throw new TrackerBusyError();
+    }
+    throw e;
   }
   try {
     return await fn();
--- a/web/tests/lib/tracker-lock-error-classification.test.mjs
+++ b/web/tests/lib/tracker-lock-error-classification.test.mjs
@@ -104,3 +104,51 @@ test("the core: a genuine ENOENT from acquireTrackerLock is NOT tagged LOCK_TIME
     },
   );
 });
+
+// Two environments cannot produce a persistent mkdir refusal, and in both a
+// green result would mean nothing (same measured skip as
+// tests/pipeline-lock-mkdir-eperm.test.mjs, which this mirrors):
+//   - root: permission bits do not apply, mkdir simply succeeds.
+//   - win32: a POSIX chmod mode maps onto the read-only attribute there, which
+//     does not deny mkdir inside the directory — acquisition would succeed and
+//     no refusal would ever happen. Measured on windows-latest, not assumed.
+const cannotRefuse =
+  typeof process.getuid === "function" && process.getuid() === 0
+    ? "running as root, permission bits do not apply"
+    : process.platform === "win32"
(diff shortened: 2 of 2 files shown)


Zoubeir23:
On the Windows EPERM/EACCES point CodeRabbit raised above — it's accurate as a description, and I'm deliberately **not** fixing it here. Reasoning, checked rather than assumed.

**The behavior is real.** `isMkdirContention` (`pipeline-lock.mjs:117-119`) treats `EEXIST`, `EPERM`, and `EACCES` all as contention, unconditionally — there's no split between "another process is mid-create/mid-remove right now" and "this process genuinely lacks permission to write here at all." So on Windows, a persistent (non-transient) permission problem keeps the core's retry loop spinning until `timeoutMs`, then surfaces as `LOCK_TIMEOUT` — same as real contention would. This PR's fix then does exactly what it's supposed to: converts a `LOCK_TIMEOUT`-tagged error to `TrackerBusyError`. It just can't tell that this particular `LOCK_TIMEOUT` traces back to a permission problem, because the core itself already lost that distinction before tagging it.

**Why it's not this PR's fix to make.** `isMkdirContention` is shared, not specific to the tracker lock: `tracker-utils.mjs:447` and `followup-seed.mjs:381` both import the identical function from `pipeline-lock.mjs`. So `followups-lock.ts` — this PR's own reference implementation for "the correct pattern" — has the exact same characteristic today. Narrowing the EPERM/EACCES treatment would mean changing `isMkdirContention` itself, which is shared by every lock user in the repo, not something scoped to `tracker-lock.ts`'s catch block.

**Why it was built this way on purpose.** The comment right above the check in `tracker-utils.mjs:441-444` cites the reason: "Windows reports a lock directory that is mid-create or mid-remove by another process as EPERM/EACCES. That is contention, not failure — treating it as fatal is how a concurrent writer dies and its write is lost (#2777, measured on windows-latest)." Windows doesn't give a reliable way to tell "someone else has this open right now" from "I lack permission" via error code alone, and #2777's own CI evidence was concurrent writers dying on exactly this ambiguity when it was treated as fatal instead. Loosening it back toward "fatal" to fix this edge case risks reopening #2777.

If it's worth tightening — say, a persistent-vs-transient signal via repeated-failure count or elapsed time within the retry loop — that's a `pipeline-lock.mjs` change with its own blast radius across every lock in the repo, and deserves its own issue with #2777's history in view rather than a fix bolted onto this one wrapper. Happy to open that issue with this write-up if useful.

Zoubeir23:
Pushed a follow-up on the EPERM/EACCES point: the catch's own comment conflated EACCES with ENOENT ("propagates as-is like ENOENT"), which isn't accurate — EACCES/EPERM goes through the core's contention classification and can surface as LOCK_TIMEOUT, as described above. Fixed the comment to say so explicitly, and added executable coverage rather than leaving it as prose: a read-only parent directory forces a real EACCES, confirming `acquireTrackerLock` retries and surfaces `LOCK_TIMEOUT` rather than throwing the raw errno — same technique and same measured root/win32 skip conditions as `tests/pipeline-lock-mkdir-eperm.test.mjs`.

One thing worth flagging on the way: writing that test caught a `node:test` gotcha — passing `skip: null` (instead of `false` or a string reason) runs the test body but still reports it as SKIP, discarding the result silently. First version of this test passed its assertion and still showed 0 in the pass count. Fixed to `false`; confirmed 6/6 now run and pass.

santifer:

Zoubeir23:

santifer (approved):
This fixes my own bug from #2903, and it's exactly the failure-direction doctrine — approving. My acquire-failure catch was unconditional (`catch { throw new TrackerBusyError(); }`), so a missing lock-dir parent (ENOENT) or a broken checkout got dressed up as "the tracker is being written by another process, retry" — a 409 the caller retries forever, because those never resolve by retrying. Gating on the core's `LOCK_TIMEOUT` tag (verified: tracker-utils.mjs sets `code = 'LOCK_TIMEOUT'` only on the genuine-contention timeout, everything else rethrows out of the loop untagged) is the right cut: contention retries, an fs failure surfaces as itself. And the comment being honest about the residual EACCES/EPERM case — that isMkdirContention treats it as contention on Windows (#2777), so a persistent permission error still times out as LOCK_TIMEOUT, and fixing THAT means touching the shared lock, not this wrapper — is the diagnosis I want to see: it names the limit instead of hiding it. Merging when I have the field.

santifer:
Thanks, @Zoubeir23 — this fixes a bug I put in (#2903) and it's the right cut. My unconditional `catch { throw TrackerBusyError }` dressed every acquire error — a missing lock-dir parent, a broken checkout — as "another process is writing, retry", which those never resolve into. Gating on the core's LOCK_TIMEOUT tag so only genuine contention retries, and everything else surfaces as itself, is exactly right. And naming the residual EACCES/EPERM case honestly (it's the shared isMkdirContention's, not this wrapper's) is the diagnosis I want to see. Merged 🚀

PR #3139, opened 2026-08-20
fix(pipeline-lock): carry the judged directory's identity to the rm

LayerC0de:
Follow-up to #3120, which merged before I finished measuring it. This is the second half.

## What #3120 left open

#3120 stopped a caller acting on a `VANISHED` verdict, where the directory was *absent*
when it looked. `STALE` has the same structure and was untouched: it says the directory we
**read** was abandoned, and the rm acts on whatever is at the path a moment later. Reaching
that verdict costs a `readFileSync` and a `process.kill`, which is room enough to be
descheduled, and in that gap a stale lock can be reclaimed by someone else and replaced by a
live one.

I found this by running #2905's §9 as a control on the merged fix, which is the one check in
the repo that measures mutual exclusion. It was not clean.

## Measured

Logging the lock directory's inode either side of the verdict, the single stale reclaim in a
2,400-acquisition run:

```
idBefore=5910974513639709  ownerBefore=39392
idAfter =6192449490350378  ownerAfter =27256  alive=true
```

A different directory, owned by a live process, deleted anyway.

30 processes each taking the lock once and exiting, which is what every career-ops CLI does.
Hold intervals recorded per process to its own file, parent looks for overlapping
`[enter, exit]` pairs:

|---|---|---|
| before #3120 | 21 | 14 of 60 |
| #3120 as merged | 2 | 2 of 60 |
| with this PR | **0** | **0 of 120** (3,600 acquisitions) |

The high-turnover regime is unchanged at 0 in 60,000 acquisitions, so this does not trade one
regime for another.

## The change

The identity of the judged directory is carried to the rm and rechecked. A path that changed
underneath us is left alone and we compete for it normally.

The second half is a backstop for the symmetric case a caller cannot control: another process
doing this to *us*. That one is silent, because the delete lands after our `owner.json` write
succeeds and nothing in the acquisition path notices. `main` already recovers when the delete
lands *before* the stamp (the write fails ENOENT and the loop competes again); reading our own
stamp back extends the same recovery to a delete landing just after it.

## No test, and why

The window is synchronous. The replacement has to land between two statements, so a
deterministic test would have to monkey-patch `fs` rather than drive the lock. I wrote one
that did not discriminate — it failed identically with and without the fix, because a crashed
holder with `staleMs: 1` is legitimately reclaimed — and removed it rather than ship a green
tick that proves nothing. The evidence is the measurement above.

Suite: 4980 passed, 0 failed.

Related: #3123 converts the three copies of this lock to import the judgment, and will pick
this up when it rebases.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved lock handling in concurrent operations.
  * Prevented stale or replaced locks from being incorrectly reclaimed.
  * Reduced the risk of multiple processes entering protected operations simultaneously.
  * Added safer recovery when lock state changes during acquisition.

### DIFF
--- a/pipeline-lock.mjs
+++ b/pipeline-lock.mjs
@@ -89,6 +89,17 @@ function readLockOwner(lockDir) {
   }
 }
 
+// stat, or null when the path is gone or unreadable. Callers that are deciding
+// whether to DELETE something need "I could not establish this" and "it is not
+// there" to arrive as the same cautious answer.
+function statOrNull(dir) {
+  try {
+    return statSync(dir);
+  } catch {
+    return null;
+  }
+}
+
 // Identity of a directory, so a lock that was removed and recreated by another
 // process is never mistaken for the one this caller created.
 function sameLockDirectory(left, right) {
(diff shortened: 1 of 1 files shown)


santifer:
Second half of #3120 landed, @LayerC0de: the inode evidence (one reclaim in 2,400 acquisitions, different directory, live owner, deleted anyway) is exactly the kind of measurement that turns a plausible race into a fixed one, and the stamp read-back closes the symmetric case where the deleted party is us. Using #2905's §9 as the control on your own merged fix is the standard the rest of us should copy. Merged! 🚀

PR #3141, opened 2026-08-20
fix(plugins): discover a plugin directory that is symlinked into a plugin root

### LINKED ISSUE #3140: bug(plugins): discoverPlugins() silently skips a symlinked plugin directory, so plugins.local/ checkouts never load
### Code of Conduct

### Existing issues

### What happened?

`discoverPlugins()` in `plugins/_engine.mjs` silently skips any plugin directory that is a **symlink**.

The filter is:

```js
const dirs = entries
  .filter(e => e.isDirectory() && !e.name.startsWith('_') && !e.name.startsWith('.'))
```

`readdirSync(..., { withFileTypes: true })` does not follow links, so a symlinked entry reports `Dirent.isDirectory() === false` (it reports `isSymbolicLink() === true` instead). The entry is dropped before the manifest is ever read.

The failure is completely silent. `warnSkip()` is never reached, so there is no warning, no error, and no diagnostic — the plugin simply never appears in `plugins.mjs list`, and its hooks never fire, even when `config/plugins.yml` enables it and `doctor.mjs` is otherwise clean. From the user's side it looks like the plugin config is being ignored.

This matters most for `plugins.local/`, which exists so a developer can work on a plugin from its own checkout. Symlinking that checkout in is the natural way to do it, and it is the one arrangement that does not work.

### Steps to reproduce

From a clean checkout:

```bash
mkdir -p /tmp/cops-repro/checkouts/career-ops-plugin-demo /tmp/cops-repro/plugins.local
cd /tmp/cops-repro/checkouts/career-ops-plugin-demo
cat > manifest.json <<'JSON'
{"id":"demo","apiVersion":1,"description":"demo","hooks":["ingest"],"requiredEnv":[],"allowedHosts":[],"humanInTheLoop":true}
JSON
echo 'export default {};' > index.mjs
ln -s /tmp/cops-repro/checkouts/career-ops-plugin-demo /tmp/cops-repro/plugins.local/demo
```

Then discover against that root:

```js
import { discoverPlugins, pluginRoots } from './plugins/_engine.mjs';
console.log(discoverPlugins(pluginRoots('/tmp/cops-repro')).map(p => p.id));
```

Observed: `[]` — and no warning of any kind.

Replace the symlink with a real directory containing the same two files and the same call returns `['demo']`.

### Expected behavior

A symlink pointing at a directory should be treated as a plugin directory. Resolving the link with `statSync` gives the right answer, and a dangling symlink should be skipped quietly rather than throwing — an unguarded resolve would take down discovery for **every** plugin in that root, not just the dead link.

### CLI tool

Claude Code

### OS and Node.js version

macOS 26.5 / Node v22.22.3

### Output of npm run doctor

Not included deliberately: `doctor` reports clean here (`onboardingNeeded: false`, no missing files), and its output carries the local plugin roster without adding anything diagnostic. The repro above is standalone and does not depend on setup state.

rubicon:
## What does this PR do?

`discoverPlugins()` dropped any plugin directory that was a symlink, because `readdirSync` does not follow links and `Dirent.isDirectory()` therefore reports `false` for one. The entry was filtered out before its manifest was read, and because `warnSkip()` was never reached the skip was completely silent: the plugin never appeared in `plugins.mjs list` even with `config/plugins.yml` enabling it. This resolves the link with `statSync` instead, and adds a regression test.

## Related issue

Closes #3140

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

---

### Why it matters

`plugins.local/` exists so a developer can work on a plugin from its own checkout. Symlinking that checkout in is the natural way to do it, and it was the one arrangement that silently did not work. From the user's side there is no signal at all: no warning, no error, `doctor.mjs` clean, and the plugin's hooks simply never fire. It reads as the plugin config being ignored.

### On the dangling-link guard

`statSync` throws on a broken symlink. That resolve is wrapped because an unguarded throw propagates out of the `for` loop over the root's entries, so a single dead link would take down discovery for **every** plugin in that root, not just itself. A dangling link is treated as not-a-directory and skipped, which is what the old filter did with it anyway.

### Test

`tests/plugin-symlink-discovery.test.mjs`, 6 assertions, auto-discovered under `tests/` so no `test-all.mjs` section was needed.

Mutation-checked in both directions rather than assumed:

- Reverting the filter to bare `e.isDirectory()` reddens 3 named assertions: `a symlinked plugin directory is discovered`, `the symlinked plugin resolves to its real checkout directory`, and `a dangling symlink does not suppress the other plugins in its root`.
- Removing the `try`/`catch` reddens `a dangling symlink in a plugin root does not throw` with `ENOENT: no such file or directory, stat '.../plugins.local/dangling'`, plus the two assertions that depend on discovery surviving the dead link.

The manifest id in the fixture matches the **link** name rather than the target directory name, which is what a developer linking `career-ops-plugin-demo` in as `demo` actually gets.

`node test-all.mjs`: 4996 passed, 0 failed, 2 warnings (both pre-existing and environmental, unrelated to this branch).

One note for Windows CI: `symlinkSync` needs the privilege, and the test creates real symlinks rather than junctions. If the runner rejects it I will convert the fixture to a skip-with-warning in the same shape as the existing symlink-privilege handling.

## Summary by CodeRabbit

* **Bug Fixes**
  * Plugin directories linked via symbolic links are now discovered correctly.
  * Broken symbolic links are safely ignored without preventing other plugins from loading.

* **Tests**
  * Added coverage for valid and dangling symbolic links during plugin discovery.

### DIFF
--- a/plugins/_engine.mjs
+++ b/plugins/_engine.mjs
@@ -24,7 +24,7 @@
  * all reuse it with no prod-vs-test drift.
  */
 
-import { existsSync, readdirSync, readFileSync, realpathSync } from 'fs';
+import { existsSync, readdirSync, readFileSync, realpathSync, statSync } from 'fs';
 import path from 'path';
 import { pathToFileURL } from 'url';
 import { resolveAndValidate } from './_net.mjs';
@@ -253,8 +253,24 @@ export function discoverPlugins(roots, overrideIds = new Set()) {
       warnSkip(root, `unreadable — ${err.message}`);
       continue;
     }
+    // A symlink to a plugin repo is a directory for our purposes: plugins.local/
+    // exists so a developer can work on a plugin from its own checkout, and
+    // linking it in is the natural way to do that. Dirent.isDirectory() is false
+    // for a symlink, so those were silently skipped -- no warning, the plugin
+    // simply never appeared in `plugins.mjs list`. statSync resolves the link;
+    // a broken one throws and is treated as not-a-directory rather than crashing
+    // discovery for every other plugin.
+    const isDirLike = (e) => {
+      if (e.isDirectory()) return true;
+      if (!e.isSymbolicLink()) return false;
+      try {
+        return statSync(path.join(root, e.name)).isDirectory();
+      } catch {
+        return false;
+      }
+    };
     const dirs = entries
-      .filter(e => e.isDirectory() && !e.name.startsWith('_') && !e.name.startsWith('.'))
+      .filter(e => isDirLike(e) && !e.name.startsWith('_') && !e.name.startsWith('.'))
       .map(e => e.name)
       .sort();
     for (const name of dirs) {
--- a/tests/plugin-symlink-discovery.test.mjs
+++ b/tests/plugin-symlink-discovery.test.mjs
@@ -1,0 +1,107 @@
+// tests/plugin-symlink-discovery.test.mjs — discoverPlugins() must treat a
+// symlinked plugin directory as a plugin directory (#3140).
+//
+// plugins.local/ exists so a developer can work on a plugin from its own
+// checkout, and linking that checkout in is the natural way to do it.
+// readdirSync does not follow links, so a symlinked entry reports
+// isDirectory() === false and a bare isDirectory() filter drops it with no
+// warning at all: the plugin never appears in `plugins.mjs list` even though
+// config/plugins.yml enables it.
+import { pass, fail, ROOT } from './helpers.mjs';
+import { mkdtempSync, mkdirSync, writeFileSync, symlinkSync, rmSync } from 'fs';
+import { join } from 'path';
+import { tmpdir } from 'os';
+import { pathToFileURL } from 'url';
+
+const { discoverPlugins, pluginRoots } = await import(pathToFileURL(join(ROOT, 'plugins/_engine.mjs')).href);
+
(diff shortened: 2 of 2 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/plugin-symlink-discovery.test.mjs`; `plugins/_engine.mjs` is the plugin engine, which is not mine to own, so I read it for signal and my approval speaks only to the test.

The test is the most thorough of the ones addressing this bug. It does not stop at discovery: it drives `loadPlugins` end to end through the symlink, asserts `lockGate` pins the tree behind the link with a stable `sha256-` integrity and the expected file set, classifies the source as `local` rather than bundled, re-loads to hit the pinned branch with the same integrity, and flips the config to disabled as a sentinel so the enabled-path assertions cannot be satisfied by an empty array. Green on CI.

Two things to flag rather than fix. The engine change itself is santifer's call, not mine. And there are three open PRs for this one bug, #3141, #3146 and #3147, all carrying the same `_engine.mjs` fix; this one has the strongest test, so if the maintainers consolidate, this is the test worth keeping.

santifer:
Issue and fix in one motion, @rubicon: Dirent.isDirectory() being false for symlinks is exactly the kind of silent skip that never errors, and guarding statSync so one dangling link cannot abort discovery for every other plugin is the half most fixes forget. Merged! 🚀

PR #3142, opened 2026-08-20
fix(web-apply): extractJsonObject's truncation salvage reused one global pad for every candidate, losing already-drafted fields

Zoubeir23:
`apply/prefill/route.ts`'s `extractJsonObject` is meant to survive the planner CLI getting killed mid-output on a big form — its own docstring: *"we salvage the largest valid prefix so the fields that DID finish still come through."* The truncation-salvage path didn't do that reliably.

```ts
const frag = s.slice(start);
const open = (frag.match(/{/g) || []).length;
const close = (frag.match(/}/g) || []).length;
const pad = "}".repeat(Math.max(0, open - close));   // computed ONCE, from the WHOLE fragment
for (let tryEnd = frag.length; tryEnd > 1; ) {
  const cand = frag.slice(0, tryEnd).replace(/,\s*$/, "") + pad;   // same pad, every candidate
  try {
    return { obj: JSON.parse(cand), truncated: true };
  } catch {
    const prevComma = frag.lastIndexOf(",", tryEnd - 1);
    if (prevComma <= start) break;
    tryEnd = prevComma;
  }
}
```

`pad` is computed once from the brace count of the entire (broken) fragment, then reused unchanged as the loop backtracks through earlier commas trying shorter candidates. Each candidate is a *different, shorter* JSON prefix — it needs however many `}` **it itself** is missing, not however many the full truncated tail was missing. Unless a candidate happens to need exactly the same nesting depth as the whole fragment, appending the wrong number of `}` fails to parse, so the loop exhausts every comma and returns `null` — discarding every field that DID finish along with the one that didn't.

**Verified by execution.** A form with a fully-answered `name` field followed by a free-text field (the kind likely to mention config/code, i.e. contain stray braces) that gets killed mid-string:

```
truncated = {"name": {"value": "Jane Doe", "needs_confirmation": false}, "about_you": {"value": "I build systems using patterns like {config: {nested: true

pre-fix:  extractJsonObject(truncated) -> { obj: null, truncated: true }
post-fix: extractJsonObject(truncated) -> { obj: { name: { value: "Jane Doe", needs_confirmation: false } }, truncated: true }
```

The `name` field is complete and would `JSON.parse` fine entirely on its own — it was lost purely because the global pad (sized for the broken `about_you` tail) didn't fit it.

**Why this isn't an edge case.** Free-text fields — cover letters, "tell us about a project you built," "why us" — are exactly the fields most likely to be both long (more likely to hit a truncation point) and to mention brace-containing content, and exactly the kind of field a planner drafts *last* on a form (identity fields first, free text last). So the failure mode concentrates precisely where truncation is most likely to happen, and it doesn't just lose the broken field — it loses every field that came before it too. The user gets a generic "couldn't parse the planner's answer as JSON" error and re-drafts the whole form, on forms whose planner call can run for minutes (`killMs` scales up to 300s).

## The fix

Moves the `pad` computation inside the loop, scoped to each candidate's own body:

```ts
for (let tryEnd = frag.length; tryEnd > 1; ) {
  const body = frag.slice(0, tryEnd).replace(/,\s*$/, "");
  const open = (body.match(/{/g) || []).length;
  const close = (body.match(/}/g) || []).length;
  const pad = "}".repeat(Math.max(0, open - close));
  try {
    return { obj: JSON.parse(body + pad), truncated: true };
  } catch { ... }
}
```

Also factors `extractJsonObject` out of the route into its own dependency-free `web/src/lib/extract-json-object.mjs` — same pattern as `pdf-paths.mjs`/`clean-chips.mjs`, noted in-file. It has no `@/` alias and no Node-only deps, which is what makes it importable directly by `node --test`; inline inside the route it couldn't be exercised at all, which is very likely why this went uncaught. No behavior change from the move itself, only from the pad fix.

Three-commit shape: the fix (as a new module) → the route now importing it → the tests.

## Test plan
- [x] `cd web && npm test` — 334 passed, 0 failed
- [x] New tests confirm the reported repro (early complete field survives a later, deeper truncation), multiple fields at varying depths all recover, single-field truncation still correctly returns null (unrecoverable), and existing behavior (complete objects, code-fence stripping, no-JSON input) is unchanged
- [x] Confirmed the pre-fix (global-pad) implementation returns `null` on the repro — this test would have caught the regression

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of JSON responses during application prefilling.
  - Added recovery for incomplete or malformed JSON, including truncated responses.
  - Supports JSON wrapped in code fences and nested objects more reliably.
  - Prevents failures when valid content appears alongside incomplete or unmatched JSON structures.

- **Tests**
  - Added coverage for valid, incomplete, malformed, nested, and unrecoverable JSON inputs.
  - Expanded validation for quoted strings, leading text, and multiple recoverable fields.

### DIFF
--- a/web/src/app/api/apply/prefill/route.ts
+++ b/web/src/app/api/apply/prefill/route.ts
@@ -4,68 +4,11 @@ import path from "node:path";
 import { resolveCli } from "@/lib/clis";
 import { careerOpsRoot, readMemory } from "@/lib/career-ops";
 import { getSession } from "@/lib/apply/session";
+import { extractJsonObject } from "@/lib/extract-json-object.mjs";
 
 export const runtime = "nodejs";
 export const dynamic = "force-dynamic";
 export const maxDuration = 320;
-
-/**
- * Pull a JSON object out of an LLM's text answer, tolerating code fences,
- * trailing prose, and — crucially — TRUNCATION (the planner getting killed
- * mid-output on a big form). When the object is incomplete we salvage the
- * largest valid prefix so the fields that DID finish still come through.
- */
-function extractJsonObject(text: string): { obj: Record<string, unknown> | null; truncated: boolean } {
-  const s = text.replace(/```(?:json)?/gi, "");
-  const start = s.indexOf("{");
-  if (start === -1) return { obj: null, truncated: false };
-
-  let depth = 0;
-  let inStr = false;
-  let esc = false;
-  let end = -1;
-  for (let i = start; i < s.length; i++) {
-    const c = s[i];
-    if (inStr) {
-      if (esc) esc = false;
-      else if (c === "\\") esc = true;
-      else if (c === '"') inStr = false;
-    } else if (c === '"') inStr = true;
-    else if (c === "{") depth++;
-    else if (c === "}") {
-      depth--;
-      if (depth === 0) {
-        end = i;
-        break;
-      }
-    }
-  }
-  if (end !== -1) {
-    try {
-      return { obj: JSON.parse(s.slice(start, end + 1)) as Record<string, unknown>, truncated: false };
-    } catch {
-      /* malformed even though balanced — fall through to salvage */
-    }
-  }
-
-  // Truncated / unbalanced: walk back from successive commas, close the JSON,
-  // and parse the largest prefix that is valid.
-  const frag = s.slice(start);
-  const open = (frag.match(/{/g) || []).length;
-  const close = (frag.match(/}/g) || []).length;
-  const pad = "}".repeat(Math.max(0, open - close));
-  for (let tryEnd = frag.length; tryEnd > 1; ) {
-    const cand = frag.slice(0, tryEnd).replace(/,\s*$/, "") + pad;
-    try {
-      return { obj: JSON.parse(cand) as Record<string, unknown>, truncated: true };
-    } catch {
-      const prevComma = frag.lastIndexOf(",", tryEnd - 1);
-      if (prevComma <= start) break;
-      tryEnd = prevComma;
-    }
-  }
-  return { obj: null, truncated: true };
-}
 
 // AI pre-fill (STREAMING NDJSON). The user's BYO CLI (read-only PLANNER — no
 // browser access) drafts an answer per field from cv.md / profile / the job's
--- a/web/src/lib/extract-json-object.mjs
+++ b/web/src/lib/extract-json-object.mjs
@@ -1,0 +1,84 @@
+/**
+ * extract-json-object.mjs — pull a JSON object out of an LLM's text answer,
+ * tolerating code fences, trailing prose, and — crucially — TRUNCATION (the
+ * planner getting killed mid-output on a big form). When the object is
+ * incomplete we salvage the largest valid prefix so the fields that DID
+ * finish still come through (apply/prefill/route.ts).
+ *
+ * Plain .mjs (same pattern as pdf-paths.mjs / clean-chips.mjs) so this can be
+ * unit-tested with `node --test`, no TypeScript build step — it has no `@/`
+ * dependency, so factoring it out of the route is what makes it testable at
+ * all.
+ */
+
+/**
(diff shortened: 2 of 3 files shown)


Zoubeir23:
Both Major findings were valid — verified both by execution against the real function before touching anything.

**Backtrack index mismatch.** \`prevComma\` is relative to \`frag\` (0-based from where the JSON starts); it was compared against \`start\`, an offset into the *full* string including any leading prose. Confirmed with 200 chars of prose before the object: \`start\` (222) exceeded \`frag.length\` (96), so \`prevComma <= start\` held on the very first backtrack attempt regardless of content, and a fully recoverable earlier field (\`name\`) came back \`null\`. Your suggested \`prevComma <= 0\` is directionally right; I went one step further and replaced the whole backtrack search with one that only considers *structural* commas (below), so I didn't end up using that exact line.

**Naive brace counting inside strings.** Confirmed separately: a completed field whose string value contains a literal unmatched \`{\` (real LLM output — free text mentioning code/config is exactly the kind of content this route drafts) got miscounted by the regex-based \`open\`/\`close\` tally, producing a pad with one extra \`}\` and making an otherwise-valid, already-complete field unparseable.

Both trace to the same root cause — neither brace-counting nor comma-finding tracked string state — so I fixed them together with one string-aware structural scan (same \`inStr\`/\`esc\` pattern the balanced-parse pass above this loop already uses), rather than patching the index comparison and the regex separately. Added your two suggested regression cases plus verified both against the pre-fix code first.

Pushed. \`node test-all.mjs\`: 4990/0 (one earlier run showed \`scan-flag-forms.test.mjs\` failing, unrelated to this diff — reran it standalone (10/10 clean) and reran the full suite (also clean), confirms a flake under load, not a regression). \`npm test\`: 336/0. \`npm run typecheck\`: clean.

Zoubeir23:
Valid, and confirmed by execution before touching anything:

\`\`\`
extractJsonObject('{"a":1,"b":{"value":"x","needs_confirmation":')
pre-fix:  { obj: { a: 1, b: { value: "x" } }, truncated: true }   ← "b" fabricated, missing needs_confirmation silently dropped
\`\`\`

\`lastStructuralComma\` accepted a comma at any nesting depth as a candidate boundary. A comma one level deeper than the outer object separates two properties *within* one field's own value object — not two top-level fields — so backtracking to it closes that field before it actually finished. The caller gets back something that looks like a complete answer but silently dropped whatever the planner hadn't written yet, which is worse than omitting the field: there's no way downstream to tell "short but complete" from "cut off mid-write."

Applied essentially your proposed fix — depth-tracked during the same structural scan, candidate boundaries restricted to depth 1 (directly inside the outer object). Same input now correctly returns \`{ a: 1 }\`, omitting the field that never finished, rather than fabricating it. Added the case you gave as the regression test, verified against the pre-fix code first.

Pushed. \`node test-all.mjs\`: 4990/0. \`npm test\`: 337/0. \`npm run typecheck\`: clean.

santifer:

Zoubeir23:

Zoubeir23:

santifer (approved):
Approving. The bug is subtle and you named it exactly: the salvage computed the closing pad ONCE from the whole truncated fragment's brace count, then reused it for every candidate prefix — but each prefix ends at its own nesting depth, so the pad over- or under-closed for all but the full tail. Tracking depth per position (the structural/depths pass) so each candidate closes to its own depth is the right fix, and extracting extractJsonObject to its own .mjs with a test that walks the truncation path is the shape this deserves. Merging when I have the field.

santifer:
Thanks, @Zoubeir23 — subtle bug, named exactly. The salvage computed the closing pad ONCE from the whole truncated fragment's brace count and reused it for every candidate prefix, but each prefix ends at its own nesting depth — so the pad over/under-closed for all but the full tail. Tracking depth per position so each candidate closes to its own depth is the right fix, and pulling extractJsonObject into its own tested .mjs is the shape it deserved. Third clean fix in 48h. Merged 🚀

PR #3143, opened 2026-08-20
fix(locks): share the waiting half of the protocol with the three copies

LayerC0de:
#3123 removes the duplicated recovery judgment from the three copies of this lock. This
removes the duplicated **wait loop**, which had drifted further and far less visibly.

`followup-seed.mjs`, `portal-health-lock.mjs` and `tracker-utils.mjs` each slept a fixed
`retryMs` and bounded acquisition on plain elapsed time. Both defects `pipeline-lock` had
already had removed were still live in all three:

| | pipeline-lock | followup-seed | portal-health-lock | tracker-utils |
|---|---|---|---|---|
| retry jitter (#2506) | yes | **no** | **no** | **no** |
| progress rule (#2835) | yes | **no** | **no** | **no** |
| release identity check | yes | **no** | yes | yes |

**Jitter.** A fixed retry wakes every waiter at the same instant to re-race, which is the
coupon-collector problem: serving N waiters takes about N·H(N) rounds, and the starving one
times out and loses its write. Measured when it was fixed in `pipeline-lock`, a fixed retry
lost an item in 12 of 20 runs against 2 of 20 jittered.

**The progress rule.** An expired deadline only means "give up" when the lock has *not*
changed hands since we last looked. Without it, a caller waiting on a healthy lock being
handed round briskly is killed for being unlucky rather than for anything being stuck.

The repo's guard in `tests/lock-rm-contention.test.mjs` checks the contention classifiers and
the recovery judgment. It does not look at the wait loop, so none of this drift showed up as
drift.

## The change

`createLockWaitPolicy()` holds the jitter, the progress rule and the absolute ceiling, and
`pipeline-lock` uses it too rather than keeping a private copy.

One thing worth calling out, because getting it wrong would have made this a downgrade:
`holderStillWedged()` **re-arms** whenever the lock changes hands, so on a briskly churning
lock the progress rule alone never terminates. `pipeline-lock` always checked `hardDeadline`
at the top of its loop; the copies had no ceiling at all, having never needed one. Adopting
the progress rule without it would have traded a premature timeout for an unbounded wait. The
policy exposes `ceilingReached()` and every caller now bounds itself with both.

## The third divergence

`followup-seed`'s `release()` verified the owner **token** but not the directory **identity**.
Between reading the stamp and the rm, a reclaimer can delete the lock and a new holder can
create one at the same path, and the rm then lands on theirs. `pipeline-lock`,
`portal-health-lock` and `tracker-utils` all bracket the rm with a stat either side and
compare; `followup-seed` was the one that did not. `sameLockDirectory` is exported rather than
inlined a fourth time.

## Tests

Suite: **4980 passed, 0 failed**, including all five lock suites
(`test/pipeline-lock.test.mjs`, `tests/lock-rm-contention.test.mjs`,
`tests/portal-health-lock.test.mjs`, `tests/with-followups-lock.test.mjs`,
`tracker-writer-lock-tests.mjs`). `pipeline-lock`'s own 21 cases cover the progress rule and
the ceiling directly, and they now exercise the shared policy rather than a private copy.

## Order

Independent of #3123 and #3139, but it touches the same three files as #3123, so whichever
lands second will need a rebase. Happy to do that either way round.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved lock handling with jittered retry timing and clearer progress tracking.
  * Added hard wait limits to prevent indefinite waiting during contention.
  * Prevented accidental removal of replacement lock artifacts.
  * Standardized lock recovery behavior across pipeline, portal health, follow-up, and tracker operations.


santifer:
The drift table in your PR body is the whole argument, @LayerC0de: pipeline-lock had jitter and the progress rule while the three copies silently kept both defects. Sharing the wait loop instead of re-fixing it three times is the same move as #3123 on the recovery side, and between the two of them the protocol finally exists in one place. Merged! 🚀

PR #3144, opened 2026-08-21
fix(cv): honor Korean profile font theme

etehofk1-ops:
## Summary

- Make the Korean CV body and heading rules honor the profile `--font-family` theme variable.
- Remove the duplicate Korean body selector.
- Add a regression assertion covering the Korean body/heading cascade.

The release template used a fixed Korean font stack after the profile theme block, so `style.font_family` was not applied to the rendered Korean CV. This patch preserves the existing fallback stack while allowing the configured profile font to win.

## Validation

- Latest `origin/main` base audited: `61f8850`
- Focused suite: 30 tests passed, 0 failed
- Korean theme regression: passed
- Korean synthetic CV HTML build: passed
- CV factual verification: passed
- Negative metric mutation: correctly rejected with a non-zero exit
- A4 strict PDF: 1 page
- `pypdf` extraction: Korean text recovered
- `pdftotext -enc UTF-8 -layout`: Korean text and ATS markers recovered; replacement characters: 0
- Visual render check: no missing Korean glyphs, overlap, clipping, or overflow observed

## Scope and privacy

This PR contains only:

- `templates/cv-template.html`
- `tests/theme-style.test.mjs`

No CV, profile, fixture, generated PDF/HTML, credentials, or personal data is included.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Korean CV typography by applying the configured font family with appropriate system-font fallbacks.
  * Removed duplicate Korean styling rules for more consistent rendering.

* **Tests**
  * Added regression coverage to verify Korean body and heading typography settings.


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/theme-style.test.mjs`; the template is outside my area and I read it for context.

The regression block reads the real `cv-template.html`, extracts the `html[lang="ko"] body` and heading rules, and asserts both use `font-family: var(--font-family, …)` while the duplicated `body, body` selector is gone. Both halves are the fix, and both are checked, so a revert to the fixed-only stack or the doubled selector reddens it. Green on CI.

santifer:
The follow-up issue is up, @etehofk1-ops: #3154 (same pattern for the Japanese and Chinese blocks, test shape included). First right of refusal is yours; comment /assign there if you want it.

santifer:
Welcome to career-ops, @etehofk1-ops! Sharp first find: the duplicated body selector plus the fixed stack meant a Korean user's profile font choice was silently ignored, and your regression test pins BOTH failure modes (the var() fallback and the duplicate selector that would quietly defeat it). Merged — your first contribution. #3154 has the Japanese/Chinese follow-up with your name on it if you want it.

santifer:
Small follow-up on the offer above, @etehofk1-ops: that follow-up issue got claimed through the /assign mechanism two minutes after it was created, before my comment here even landed, and its PR is already in review. That is the pool moving fast, not your chance being taken. If you want a next one, the good-first-issue label has open picks, and if you tell me what you enjoyed about this first fix (the CSS side, the test side, the i18n side) I will point you at the best fit.

PR #3145, opened 2026-08-21
fix(providers): validate the resolved address before connecting (DNS rebinding)

### LINKED ISSUE #3096: providers: no fetch validates the resolved IP — a public hostname can point at a private address (DNS rebinding)
Surfaced by CodeRabbit on #2812 and measured by @Zoubeir23 there; re-verified on today's main. Every provider's host guard (e.g. `resolveOrigin()` in `providers/consider.mjs`) pins the *hostname* to a public registrable domain, but nothing checks where that name *resolves*:

```console
$ grep -rn "169.254|isPrivate|127.0" providers/_http.mjs providers/_dns-cache.mjs
(no matches, exit 1)
```

A hostname with an A record pointing at `127.0.0.1`, `169.254.169.254` or another private range gets connected to by native `fetch()`; `redirect:'error'` blocks redirect-based hops but not a direct malicious record. Exposure requires a hostile entry in the user's own `portals.yml` (or a compromised board DNS), so this is hardening rather than a remotely triggerable hole: local-first means the config author is the user.

The right fix is one shared choke point, not 30 per-provider patches: resolve-then-validate in `providers/_http.mjs` (reject private/link-local/loopback ranges for IPv4 and IPv6, including IPv4-mapped IPv6), so every provider inherits it. `_dns-cache.mjs` already centralizes lookups and is a natural place to attach the check.

Scoping notes for whoever takes it:
- The check belongs at connection time (validate the address actually dialed), not just a pre-flight lookup, or the rebinding window survives.
- Node's `fetch` has no connect hook; an `undici` Agent with a custom `connect` (or `lookup`) can enforce it without changing provider code.
- Tests: a name resolving to 127.0.0.1 must be rejected; public addresses must pass; IPv6 `::1` and `::ffff:127.0.0.1` covered.

abankar1:
Fixes #3096. Design posted in-thread before writing this, because the obvious placement turned out to be ruled out.

## The constraint that shaped it

`_dns-cache.mjs` is the natural home — its header explains it patches `dns.lookup` on the `node:dns` module object specifically to avoid a dependency, and `net.connect` reads it at call time, which is exactly the connect-time property this needs.

But that patch is **process-global**, and Node calls `dns.lookup` even for a numeric host. Measured:

```
fetch('http://127.0.0.1:PORT')  -> dns.lookup calls: 1
fetch('http://localhost:PORT')  -> dns.lookup calls: 2
```

So an unconditional private-range rejection there rejects every loopback connection anything in the process makes — already 10+ test files in this repo stand up a local server and fetch it. An env opt-out would have "fixed" the suite by disabling the guard precisely where it is most likely to regress.

`undici` is also not a dependency (`package.json` carries four), and `_dns-cache.mjs` exists in its current shape to keep it that way, so the Agent route would have added one for this.

**So the guard is scoped, not global:** `_http.mjs` marks its own requests with an `AsyncLocalStorage`, and the patched lookup validates only inside that context. Connect-time validation is preserved, loopback keeps working everywhere else, no new dependency.

## Measured end to end, no DNS stubbing

```
provider fetch     -> localhost   blocked (ECAREEROPS_BLOCKED_ADDRESS)
NON-provider fetch -> localhost   still works
cache pre-warmed outside, then
provider fetch     -> localhost   blocked
```

That third line is the hole I went looking for after writing the first two. The guard is applied at callback **entry**, so it covers a fresh resolver answer, a **cache hit**, and a coalesced waiter alike. Guarding only the resolver path would leave the cache as the bypass: a hostname resolved outside a provider fetch is memoized unvalidated, and the next provider fetch is served the private address from memory. There's a test for exactly that sequence.

## Scope

A literal private IP in a URL is untouched, and doesn't need to be — the per-provider host guards reject it before any fetch happens:

```
ashby.detect({careers_url:'http://169.254.169.254/acme'})  -> null
ashby.detect({careers_url:'http://127.0.0.1/acme'})        -> null
```

This closes the remaining path: a **permitted** hostname that resolves somewhere private.

## Details worth a look

- **IPv4-mapped IPv6 is unwrapped and judged as IPv4.** `::ffff:127.0.0.1` is loopback however it's spelled; treating the spellings differently is the bypass.
- **Octets must be canonical decimal.** `0177.0.0.1` is octal for 127.0.0.1 in some parsers, and a form this function reads as public while a connector reads as loopback defeats the point. `dns.lookup` never emits a non-canonical quad, so it's defence in depth — but I had it wrong first (`/^\d{1,3}$/` accepted `01.2.3.4`) and the test caught it.
- **The error is deliberately NOT in `RESOLVER_FAILURE_CODES`.** This is a verdict about one hostname, not resolver health — counting it as the latter would negative-cache it and could feed the resolver-outage signal from #2229 over a config problem the user should be told about instead.

## A wrong turn worth recording

My first test stubbed `dns.lookup` to make `evil.example` resolve to loopback, and reported a leak. The guard was fine — a stub that answers directly never reaches the patched lookup, so it tests nothing and looks like a failure. The committed tests use `localhost` against a local server instead: a real name, a real loopback address, still fully offline.

## Verification

Checked for discrimination: with the two wiring changes reverted, **exactly the two end-to-end assertions fail** and the six classification/scoping ones pass.

Classification covers both families, IPv4-mapped, the cloud-metadata address, and the boundaries either side of `172.16/12`, `100.64/10` and `192.168/16`, so a mask off by one bit fails rather than passing quietly.

`node test-all.mjs --quick`: 4402 pass / 0 fail (3 warnings pre-existing and environmental). The SYSTEM_PATHS coverage guard is satisfied — `providers/` is covered by a directory rule, so no manifest entry was needed.

## Summary by CodeRabbit

* **Security**
  * Provider requests now block connections to private, loopback, link-local, cloud metadata, multicast, and other reserved IP addresses.
  * Protection applies to fresh DNS lookups, cached results, and concurrent requests.
  * DNS rebinding attempts and malformed addresses are rejected.
* **Bug Fixes**
  * Non-provider requests to local addresses continue to work as expected.

### DIFF
--- a/providers/_dns-cache.mjs
+++ b/providers/_dns-cache.mjs
@@ -50,6 +50,7 @@
  */
 
 import dns from 'node:dns';
+import { inProviderFetch, isBlockedAddress, blockedAddressError } from './_ip-guard.mjs';
 
 /**
  * DNS failures that mean *the resolver itself refused or failed*, as opposed
@@ -244,11 +245,40 @@ export function createCachedLookup(realLookup, options = {}) {
   /** @type {Map<string, Function[]>} */
   const inflight = new Map();
 
+  /**
+   * Wrap a lookup callback so a non-public address never reaches the connector
+   * (#3096).
+   *
+   * Applied at ENTRY, so it covers all three ways a result is delivered: a
+   * fresh resolver answer, a cache HIT, and a coalesced waiter. Validating
+   * only the resolver path would leave the cache as the hole — a hostname
+   * resolved outside a provider fetch is cached unvalidated, and the next
+   * provider fetch for that name would be served the private address from
+   * memory without ever reaching the check.
+   *
+   * Only inside a provider request: this lookup is patched process-wide, and
+   * loopback has to keep working for everything else (see _ip-guard.mjs).
+   */
+  function guarded(hostname, callback) {
+    if (!inProviderFetch()) return callback;
+    return (err, ...rest) => {
+      if (err) return callback(err, ...rest);
+      // all:true yields one array of {address, family}; otherwise (address, family).
+      const addresses = Array.isArray(rest[0])
+        ? rest[0].map((entry) => entry && entry.address)
+        : [rest[0]];
+      const bad = addresses.find((address) => isBlockedAddress(address));
+      if (bad !== undefined) return callback(blockedAddressError(hostname, bad));
+      return callback(err, ...rest);
+    };
+  }
+
   function cachedLookup(hostname, options, callback) {
     if (typeof options === 'function') {
       callback = options;
       options = {};
     }
+    callback = guarded(hostname, callback);
     // dns.lookup accepts a bare family number in place of an options object.
     const opts = typeof options === 'number' ? { family: options } : (options ?? {});
     const key = `${hostname}|${opts.family ?? 0}|${opts.all ? 1 : 0}|${opts.hints ?? 0}|${opts.verbatim ?? ''}`;
--- a/providers/_http.mjs
+++ b/providers/_http.mjs
@@ -3,12 +3,27 @@
 
 import './_dns-cache.mjs'; // memoize dns.lookup process-wide (see that file)
 import { DEFAULT_USER_AGENT, BROWSER_LIKE_USER_AGENT } from '../user-agent.mjs';
+import { providerFetchContext } from './_ip-guard.mjs';
 
 export { BROWSER_LIKE_USER_AGENT };
 
 const DEFAULT_TIMEOUT_MS = 10_000;
 
-async function fetchWithTimeout(url, { timeoutMs = DEFAULT_TIMEOUT_MS, headers = {}, method = 'GET', body = null, redirect = 'follow' } = {}, consume) {
+async function fetchWithTimeout(url, opts = {}, consume) {
+  // Mark this request as provider traffic for the whole of its async life, so
+  // the patched dns.lookup validates the addresses it resolves (#3096). The
+  // guard is scoped rather than global because _dns-cache.mjs patches
+  // node:dns process-wide, and loopback has to keep working for everything
+  // that is not a provider fetch — see providers/_ip-guard.mjs.
+  //
+  // AsyncLocalStorage.run wraps the ENTIRE fetch, not just the call that
+  // starts it: the DNS lookup happens inside connect, well after the
+  // synchronous part of fetch() has returned, and the context has to still be
+  // entered when it does.
+  return providerFetchContext.run({ url: String(url) }, () => fetchInContext(url, opts, consume));
+}
+
+async function fetchInContext(url, { timeoutMs = DEFAULT_TIMEOUT_MS, headers = {}, method = 'GET', body = null, redirect = 'follow' } = {}, consume) {
   const controller = new AbortController();
   const timer = setTimeout(() => controller.abort(), timeoutMs);
   try {
--- a/providers/_ip-guard.mjs
+++ b/providers/_ip-guard.mjs
@@ -1,0 +1,150 @@
+// Reject a connection whose resolved address is not a public one (#3096).
+//
+// Every provider's host guard pins the HOSTNAME to a public registrable
+// domain, but nothing checked where that name RESOLVES. A hostname with an A
+// record pointing at 127.0.0.1, 169.254.169.254 (cloud metadata) or an RFC1918
+// range was connected to normally. `redirect: 'error'` blocks redirect-based
+// hops; it does not block a direct malicious record.
+//
+// WHY THE CHECK IS SCOPED, NOT GLOBAL. _dns-cache.mjs patches `dns.lookup` on
+// the node:dns module object, which is process-wide. Rejecting private ranges
+// there unconditionally would reject every loopback connection anything in the
+// process makes — measured, Node calls dns.lookup even for a numeric host:
+//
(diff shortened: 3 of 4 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/providers/private-address-guard.test.mjs`; `_ip-guard.mjs`, `_dns-cache.mjs` and `_http.mjs` are outside my area and I read them for context.

This is a security test with real teeth. Classification covers loopback, RFC1918, CGNAT, link-local including the `169.254.169.254` metadata address, IPv4-mapped IPv6, and the off-by-one boundaries (`172.15`/`172.32`, `100.63`, `192.169`), and it blocks non-canonical input like `01.2.3.4` rather than guessing. The end-to-end cases use a real loopback server with no DNS stubbing, which is the right call and the header says why, and they prove the three things that matter: a provider fetch to a name resolving to loopback is refused with `ECAREEROPS_BLOCKED_ADDRESS`, a non-provider fetch to the same address still works (the scoping guarantee), and a pre-warmed cache entry is still guarded. Green on CI.

Not blocking and not mine to gate: the guard logic in `_ip-guard.mjs` is the kind of SSRF-sensitive code that is worth a maintainer's eye on the classification itself, not just the test. Flagging it for santifer, the test I am happy with.

santifer:
This is the right architecture, @abankar1: the ALS-scoped context instead of a global rejection (your two-lookup measurement settles that), validation at lookup entry so cache hits and coalesced waiters are covered, and the blocked-address error deliberately kept out of resolver-health accounting. Three concrete gaps before it can merge, each verified by running the branch:

**1. Non-canonical IPv6 forms pass the literal checks** (`_ip-guard.mjs`, the `::`/`::1` comparisons):

```
$ node --input-type=module -e "import {isBlockedAddress} from './providers/_ip-guard.mjs'; ..."
0:0:0:0:0:0:0:1        false   # expanded ::1
0:0:0:0:0:0:0:0        false   # expanded ::
::ffff:7f00:1          false   # v4-mapped in hex form, no dots so the unwrap never fires
::1                    true
```

The dotted `::ffff:a.b.c.d` unwrap only triggers on `tail.includes('.')`, and the loopback/unspecified checks are string equality. Canonicalize before judging: `net.isIPv6()` + expanding/compressing to one canonical form (or parsing the groups numerically) closes all three rows. The fail-closed default you already chose ("neither v4 nor v6 - refuse to guess") is the right template; the gap is only that these ARE valid v6 spellings, so they reach the lenient branch.

**2. `CAREER_OPS_NO_DNS_CACHE=1` silently disables the guard.** The patched lookup is only installed inside the cache setup, and line 392 skips that entirely under the env var. That variable is documented as a cache/pacing opt-out; having it also switch off SSRF protection is exactly the "env opt-out fixes the suite by disabling the guard" failure your own header warns about. Install the guard wrapper unconditionally and let only the memoization be conditional.

**3. One provider dials outside the guarded helper.** `providers/consider.mjs:95` does a bare global `fetch(`${origin}/jobs`)` for its CSRF handshake, so `inProviderFetch()` is false for that request and the lookup guard never runs on it. Either route it through `_http.mjs` or export a small `runInProviderContext()` so direct calls can enter the same scope: worth a quick grep for other bare `fetch(` in `providers/` while at it.

None of this changes the design: 1 is a parser fix, 2 is moving one hook outside a conditional, 3 is one call site plus a sweep. Happy to merge on the next push.

abankar1:
The earlier red `test (macos-latest)` was a flake, not this branch. Recording the check rather than just re-running it:

The three failures were all in `generate-pdf`'s page-budget test:

```
❌ generate-pdf default page budget regressed: Refusing to write the PDF outside
   the tracker workspace: input escapes the tracker workspace:
   /Users/runner/work/career-ops/career-ops/output/page-budget-test-GL7K8n/three-pages.html
```

That message comes from the workspace guard in d3dbd63, and this branch touches only `providers/_ip-guard.mjs`, `providers/_dns-cache.mjs` and `providers/_http.mjs` — nothing in the PDF path, and nothing that resolves a tracker workspace.

What I checked before assuming:

- the failing run's head SHA was the current head, so it was not a stale run;
- the same commit passes the whole suite locally on macOS — 5073 passed, 0 failed, that test included;
- no other open PR's macOS leg was failing;
- my added test file sets no env var and never chdirs, so it cannot leak state into a later test — the one way adding a discovered test file could plausibly disturb an unrelated one.

Re-ran it: macOS and Ubuntu both green. Flagging it because if that test is intermittent on the macOS runner it will bite someone else with a much less obvious diff, and it looked like a real regression until the SHA and the local run ruled it out.

santifer:
Your flake report earned its keep: a second contributor hit the identical failure the same day, and the pattern you both logged is now #3162, with a companion fix (#3163) that makes the guard name WHICH event fired (real escape vs failed realpath, with errno) so the next CI hit diagnoses itself. Exactly the outcome logging-instead-of-rerunning exists for. Your PR itself stays unaffected: the flake was in main's test, not your branch.

santifer:
Excellent close on #3096, @abankar1. The pre-warmed-cache test is the one that impressed me: going looking for the bypass after the first two asserts passed is how security fixes get proven. Documenting the wrong turn (the stubbed dns.lookup test that reported a false leak) makes the next person's job easier, and keeping ECAREEROPS_BLOCKED_ADDRESS out of RESOLVER_FAILURE_CODES so it cannot poison the outage signal from #2229 sealed it. Merged! 🚀

PR #3146, opened 2026-08-21
fix(plugins): discoverPlugins() silently skips a symlinked plugin directory

### LINKED ISSUE #3140: bug(plugins): discoverPlugins() silently skips a symlinked plugin directory, so plugins.local/ checkouts never load

abankar1:
Fixes #3140. Reported and diagnosed by @rubicon; I commented offering to drop it if they were already on it.

Reproduced on main with their fixture — same two files, same call, only the symlink differs:

```
symlinked plugin dir -> []
real plugin dir      -> ["demo"]
```

`readdirSync(withFileTypes)` does not follow links, so a symlinked plugin directory reports `isDirectory() === false` and was filtered out **before** the manifest was read and before `warnSkip()` could fire. The plugin never appeared in `plugins.mjs list`, its hooks never ran, and nothing said why — with `config/plugins.yml` enabling it and `doctor.mjs` clean.

`plugins.local/` exists so a developer can work on a plugin from its own checkout, and symlinking that checkout in is the natural way to do it. The one arrangement the directory is *for* was the one that didn't work.

## The resolve, and the two ways to get it wrong

`statSync` is the right call because it **follows** the link, which is the question actually being asked: *does this entry lead to a directory?*

- A bare "symlinks are fine" pass would be wrong — a link to a **file** is not a plugin directory. Pinned by a test.
- The `try/catch` is load-bearing rather than defensive habit: `statSync` throws on a dangling link, and unguarded that would abort discovery for **every** plugin in the root rather than skipping the one dead entry. That's your expected-behaviour note, and it's the assertion I care most about:

```
symlink + real + DANGLING link -> ["demo","second"]
```

The `_`/`.` prefix filter still runs first and is unchanged, so `_engine.mjs` and friends stay excluded whether or not they're links.

## Verification

Three of the five assertions guard against over-correcting, since "treat symlinks as directories" has a wrong shape on either side of the right one: the file-symlink case, the dangling-link case, and the prefix-exclusion case. Plus the bug itself and a plain real directory, so an inverted filter can't pass.

Hermetic — every root is a tmpdir built in-test, so nothing depends on the repo's own `plugins/` or on `plugins.local/` existing.

Checked for discrimination: with `plugins/_engine.mjs` reverted, **exactly the symlink and dangling-link assertions fail** and the three guards pass.

`node test-all.mjs --quick`: 4988 pass / 0 fail (3 warnings pre-existing and environmental).

## Summary by CodeRabbit

* **Bug Fixes**
  * Plugin discovery now supports valid symlinked directories.
  * Invalid, dangling, unreadable, and file-targeting symlinks are safely skipped without interrupting discovery.
  * Existing filtering and manifest validation continue to work for discovered plugins.

* **Tests**
  * Added coverage for valid directories, symlinked directories, dangling symlinks, file symlinks, and excluded entries.

### DIFF
--- a/plugins/_engine.mjs
+++ b/plugins/_engine.mjs
@@ -24,7 +24,7 @@
  * all reuse it with no prod-vs-test drift.
  */
 
-import { existsSync, readdirSync, readFileSync, realpathSync } from 'fs';
+import { existsSync, readdirSync, readFileSync, realpathSync, statSync } from 'fs';
 import path from 'path';
 import { pathToFileURL } from 'url';
 import { resolveAndValidate } from './_net.mjs';
@@ -254,7 +254,27 @@ export function discoverPlugins(roots, overrideIds = new Set()) {
       continue;
     }
     const dirs = entries
-      .filter(e => e.isDirectory() && !e.name.startsWith('_') && !e.name.startsWith('.'))
+      .filter(e => !e.name.startsWith('_') && !e.name.startsWith('.'))
+      // readdirSync(withFileTypes) does NOT follow links, so a symlinked
+      // plugin directory reports isDirectory() === false and was dropped here
+      // — before the manifest was read and before warnSkip() could fire, so
+      // the plugin simply never appeared and nothing said why (#3140).
+      // plugins.local/ exists so a developer can work from their own checkout,
+      // and symlinking it in is the natural way to do that.
+      //
+      // statSync FOLLOWS the link, which is the question being asked: does
+      // this entry lead to a directory? Wrapped, because it throws on a
+      // dangling link — unguarded, one dead symlink would abort discovery for
+      // EVERY plugin in the root rather than skipping the one that is broken.
+      .filter((e) => {
+        if (e.isDirectory()) return true;
+        if (!e.isSymbolicLink()) return false;
+        try {
+          return statSync(path.join(root, e.name)).isDirectory();
+        } catch {
+          return false; // dangling or unreadable link — skip it, keep the rest
+        }
+      })
       .map(e => e.name)
       .sort();
     for (const name of dirs) {
--- a/tests/plugins-symlink-discovery.test.mjs
+++ b/tests/plugins-symlink-discovery.test.mjs
@@ -1,0 +1,102 @@
+// tests/plugins-symlink-discovery.test.mjs — a symlinked plugin directory must
+// be discovered (#3140).
+//
+// readdirSync(withFileTypes) does NOT follow links, so a symlinked plugin dir
+// reports isDirectory() === false and was filtered out before the manifest was
+// read — and before warnSkip() could fire, so the plugin never appeared and
+// nothing said why. plugins.local/ exists so a developer can work from their
+// own checkout, and symlinking it in is the natural way to do that, so the one
+// arrangement the directory is for was the one that did not work.
+//
+// The dangling-link case is the reason the resolve is guarded: statSync throws
+// on a broken link, and unguarded that would abort discovery for EVERY plugin
+// in the root rather than skipping the one dead entry.
+//
+// Run:  node --test tests/plugins-symlink-discovery.test.mjs
+
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { mkdtempSync, mkdirSync, writeFileSync, symlinkSync, rmSync } from 'node:fs';
+import { dirname, join } from 'node:path';
+import { tmpdir } from 'node:os';
+import { fileURLToPath, pathToFileURL } from 'node:url';
+
+const ROOT = dirname(dirname(fileURLToPath(import.meta.url)));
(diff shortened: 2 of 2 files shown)


Scott-Emberson (commented):
Reviewed as code owner of `tests/`. Scope is `tests/plugins-symlink-discovery.test.mjs`; `plugins/_engine.mjs` is the plugin engine and not mine to own, so I read it for signal.

No defect in the test, it is correct: symlinked dir discovered, filter not inverted, dangling link skipped, a file-symlink rejected, and the underscore/dot prefix filter still applied. I am commenting rather than approving because this is one of three open PRs for the same bug (#3140), and this test is discovery-only, a strict subset of #3141, which also exercises the load, lock-pinning and disabled paths. The engine fix is identical across all three and is santifer's to decide. My suggestion is to consolidate on #3141 and close the duplicates behind it.

santifer:
Closing as the duplicate of the pair, @abankar1, and only on timing: #3141 shipped the same fix (identical approach, down to the statSync-with-dangling-guard) two hours earlier, alongside the issue itself. Your independent repro on #3140 was genuinely useful: two people hitting the same walls separately is what made the diagnosis certain, and the comment about one dead symlink aborting discovery for every plugin is the sharpest line in either PR.

Heads-up for #3147: it builds on this branch's test file, so once #3141 is in you will want to rebase it onto main and drop the engine/test hunks that #3141 now carries. The suppression fix itself stays fully yours.

abankar1:
No argument at all — #3141 got there first and the approach is the same one, so closing the duplicate is right. Thanks for saying which half was still useful; that is more than a duplicate usually gets.

Acted on the heads-up: **#3147 is rebased onto main** and now carries only the three whats-new commits. The `plugins/_engine.mjs` and `tests/plugins-symlink-discovery.test.mjs` hunks are gone — #3141 carries that ground now. Diff is three files:

```
web/src/app/api/whats-new/route.ts           | 10 +--
web/src/lib/whats-new-suppression.mjs        | 76 +++++++++
web/tests/lib/whats-new-suppression.test.mjs | 91 +++++++++
```

web 344 pass / 0 fail, `test-all --quick` 5072 pass / 0 fail on the rebased branch.

One thing worth noting from re-running it against #3141's version: my test file asserted a symlink to a **file** must not be discovered, and that a **dangling** link must not abort discovery for the rest of the root. Both pass against #3141's implementation, so nothing is lost — but if you ever want those two cases pinned, they are in the closed branch's `tests/plugins-symlink-discovery.test.mjs` and I am happy to send them as a test-only PR on top of #3141.

PR #3147, opened 2026-08-21
fix(web): "new matches" hides an employer's entire board after one evaluation

### LINKED ISSUE #3131: web: "new matches" suppresses by company name, hiding an employer's entire board after one evaluation
### What happened?

`web/src/app/api/whats-new/route.ts` hides an offer when its company already
appears in `data/applications.md`:

```js
// Companies already evaluated → don't resurface as "new".
const evaluated = new Set(readApplications().map((a) => norm(a.company)).filter(Boolean));
...
if (company && evaluated.has(norm(company))) return null;
```

The suppression key is the **company name**, so evaluating a single role at an
employer removes that employer's entire board from Explore permanently,
including postings created later by different teams.

For a small company with one opening this is the intended behaviour. For a large
employer it is not. In my own use an employer with 20+ live roles in my pipeline
became completely invisible in Explore after I evaluated one of them, and I only
found the remaining roles by querying the ATS API directly, bypassing the UI.

The scanner is working correctly throughout: the roles are already present and
correct in `data/scan-history.tsv`. Only the view hides them. Because the hiding
is silent, "no new matches" is indistinguishable from "matches exist but are
suppressed", which is what makes this hard to notice.

### Steps to reproduce

1. Have two or more live roles at the same employer surfaced into
   `data/scan-history.tsv` by a scan.
2. Evaluate exactly one of them so a row for that company lands in
   `data/applications.md`.
3. Open the web UI → Explore ("fresh matches this week").
4. The other roles at that employer are gone, with no indication they were
   filtered.

### Expected behavior

Only the role that was actually evaluated is hidden. The employer's other
postings continue to surface.

### Version / channel
1.28.0

### Screen / route
`/explore` (and the "new matches this week" panel on `/`)

### CLI tool
Claude Code

### Possible direction (happy to be routed)

Key the suppression on the job URL rather than the company name, normalised
through the existing `normalizeUrlForDedup` in `scan.mjs`.

The obstacle is that `applications.md` rows do not carry the posting URL. Report
files do (`**URL:**` is mandatory per AGENTS.md), so the mapping would have to be
resolved via the report link column, or the tracker would need a URL column.
Both are bigger decisions than a bug fix, which is why I am opening an issue
rather than sending a PR: CONTRIBUTING says "Failing one of these is a routing,
not a rejection", and I would rather hear which door this belongs to first.

It is also possible the current behaviour is deliberate for the common case and
you would rather scope any change narrowly (for example, only stop suppressing
once an employer has more than N distinct postings in history). Happy to
implement whichever shape you prefer.

### Note on scope

#3130 covers a different defect in the same function: `status` is not
read at all, so `Rejected` and `Discarded` rows suppress as well. That one is a
straightforward bug (the code contradicts its own comment) and is independent of
this issue. Filing them separately so they do not get conflated.

abankar1:
Fixes #3131. Reported by @KehanLiu, who traced it to the exact line; I commented offering to drop it if they wanted to fix it themselves.

`whats-new/route.ts` keyed suppression on the **company alone**:

```js
const evaluated = new Set(readApplications().map((a) => norm(a.company)).filter(Boolean));
...
if (company && evaluated.has(norm(company))) return null;
```

So evaluating **one** role at an employer removed that employer's entire board from Explore — permanently, including postings first seen afterwards. The reporter had an employer with 20+ live roles go completely invisible after evaluating one of them, and found the rest only by querying the ATS API directly.

The scanner is correct throughout: the roles are in `scan-history.tsv`, only the view hides them. And because the hiding is silent, *"no new matches"* is indistinguishable from *"matches exist but are suppressed"* — which is what makes it hard to notice at all.

Now keyed on **company AND role**, through the same core `normalizeTextKey` the route already resolves, so the pair stays consistent with CLI dedup and no second key helper appears.

## The trade, stated plainly

If the tracker's role text and the scan-history title drift apart (`"Staff SWE"` vs `"Staff Software Engineer"`), an already-evaluated role can reappear in Explore once — a **visible** false positive the user can dismiss. What it replaces is a **silent false negative**: a board the user cannot see and gets no signal about. That's the direction #2544 argues for.

A tracker row with no role keys as `company|`, so a backfilled row (#1799) suppresses only a scan row that also has no title, instead of the whole employer.

## Structure

The decision moved to `web/src/lib/whats-new-suppression.mjs`, because it lived inline in a TypeScript route where no `node --test` unit test can reach it — web's script is `node --test tests/**/*.test.mjs` with no TS runner. Same pattern as `funnel-tiles.mjs` / `status-alias.mjs` / `clean-chips.mjs`.

The normalizer is **injected**, not imported: the route resolves the core's `normalizeTextKey` from the user's own checkout so web dedup matches CLI dedup, and importing a key helper into this file would be exactly the local reimplementation #2666 removed. The tests inject the Unicode-safe mirror the web already ships as its fallback, rather than a toy normalizer that wouldn't exist in production.

## Verification

The decisive pair, because a "fix" that simply stopped suppressing would satisfy the headline assertion and silently break the feature:

- a sibling role at an evaluated employer must **not** be suppressed (the bug);
- the evaluated role itself must **still** be suppressed on a repost.

Also pinned: `日本電産` / `Škoda` key distinctly from a sibling role and from `Koda` (so #2666 stays fixed under the new key), and a company that normalizes to nothing suppresses nothing — otherwise one unkeyable row would hide every scan row whose company is also unkeyable, which would relocate #2666 rather than fix it.

**One assertion records a non-contract deliberately.** I first wrote `"Acme, Inc."` as matching `"acme"`; the test caught me. The core key keeps punctuation as a separator, so it keys as `"acme inc"` and does not match — unchanged from the company-only key this replaces. The legal-suffix question is orthogonal to this issue, so pinning it either way would invent a contract the key never had.

Checked for discrimination: with the key reverted to company-only, **exactly the three assertions whose answers the change alters fail**, and the five pinning unchanged behaviour pass.

`web`: 335 pass / 0 fail. `node test-all.mjs --quick`: 4988 pass / 0 fail (3 warnings pre-existing and environmental).

## Summary by CodeRabbit

* **Bug Fixes**
  * New match filtering now suppresses only roles that have already been evaluated, rather than hiding all roles from the same company.
  * Reposted roles and variations are handled more consistently.
  * Unrelated employers, distinct roles, and non-Latin names remain correctly visible.

* **Tests**
  * Added coverage for role-level filtering, missing information, reposts, and name collisions.

### DIFF
--- a/web/src/app/api/whats-new/route.ts
+++ b/web/src/app/api/whats-new/route.ts
@@ -2,6 +2,7 @@ import fs from "node:fs";
 import path from "node:path";
 import { careerOpsRoot, readApplications } from "@/lib/career-ops";
 import { getNormalizeTextKey } from "@/lib/core/text-key";
+import { evaluatedKeys, isEvaluated } from "@/lib/whats-new-suppression.mjs";
 import type { DiscoveredOffer } from "@/lib/explore";
 import { collectWhatsNew, resolveOfferLimit } from "@/lib/whats-new.mjs";
 
@@ -33,16 +34,17 @@ export async function GET(req: Request) {
     return Response.json({ offers: [], count: 0 });
   }
 
-  // Companies already evaluated → don't resurface as "new".
+  // Roles already evaluated → don't resurface as "new". Keyed on company AND
+  // role, not company alone: suppressing by employer removed that employer's
+  // entire board after one evaluation (#3131). See lib/whats-new-suppression.
   const normalizeTextKey = await getNormalizeTextKey();
-  const norm = (s: string) => normalizeTextKey(s, " ");
-  const evaluated = new Set(readApplications().map((a) => norm(a.company)).filter(Boolean));
+  const evaluated = evaluatedKeys(readApplications(), normalizeTextKey);
 
   const toOffer = (c: string[]): DiscoveredOffer | null => {
     const [url, firstSeen, portal, title, company, status, location] = c;
     if (!url || !/^https?:\/\//i.test(url)) return null;
     if (status && /skipped|expired/i.test(status)) return null;
-    if (company && evaluated.has(norm(company))) return null;
+    if (isEvaluated(evaluated, normalizeTextKey, company, title)) return null;
     return {
       url,
       company: (company || "").trim(),
--- a/web/src/lib/whats-new-suppression.mjs
+++ b/web/src/lib/whats-new-suppression.mjs
@@ -1,0 +1,76 @@
+// Which discovered offers count as "already evaluated" for the Explore /
+// "new matches" supply loop (api/whats-new).
+//
+// Pure JS (no TS types) so it can be imported by the route AND by a
+// `node --test` unit test, matching funnel-tiles.mjs / status-alias.mjs /
+// clean-chips.mjs — web's test script is `node --test tests/**/*.test.mjs`
+// and there is no TS runner.
+//
+// THE KEY IS COMPANY **AND** ROLE (#3131). Suppressing by employer alone meant
+// one evaluated role removed that employer's entire board from Explore,
+// permanently, including postings first seen afterwards. An employer with 20+
+// live roles went completely invisible after evaluating one of them, and the
+// scanner was working throughout — the rows were in scan-history.tsv, only the
+// view hid them. "No new matches" then reads identically to "matches exist but
+// are suppressed", which is what makes it hard to notice.
+//
+// The trade is deliberate and in the safer direction. If the tracker's role
+// text and the scan-history title drift apart ("Staff SWE" vs "Staff Software
+// Engineer"), an already-evaluated role can reappear in Explore once: a
+// VISIBLE false positive the user can dismiss. What it replaces is a silent
+// false negative — a board the user cannot see and gets no signal about.
+//
+// The normalizer is injected rather than imported, because the route resolves
+// the CORE's normalizeTextKey from the user's own checkout (lib/core/text-key)
+// so web dedup always matches CLI dedup. A second key helper here is exactly
+// the local reimplementation #2666 removed.
+
+/**
+ * Suppression key for one company/role pair.
+ *
+ * A tracker row with no role yields "company|", which matches only a scan row
+ * that also has no title — so a backfilled, role-less row (#1799) suppresses
+ * that one shape rather than the whole employer.
+ *
+ * @param {(value: unknown, separator?: string) => string} norm - Core key fn.
+ * @param {string} company
+ * @param {string} role
+ * @returns {string}
+ */
+export function suppressionKey(norm, company, role) {
+  return `${norm(company, " ")}|${norm(role, " ")}`;
+}
+
+/**
+ * Keys for every evaluated application.
+ *
+ * Rows whose company normalizes to nothing are dropped: they cannot identify
+ * an employer, and keeping them would let one blank row suppress every scan
+ * row whose company is also unkeyable.
+ *
+ * @param {{company: string, role: string}[]} applications
+ * @param {(value: unknown, separator?: string) => string} norm
+ * @returns {Set<string>}
+ */
+export function evaluatedKeys(applications, norm) {
+  const keys = new Set();
+  for (const app of applications ?? []) {
+    if (!norm(app?.company, " ")) continue;
+    keys.add(suppressionKey(norm, app?.company ?? "", app?.role ?? ""));
+  }
+  return keys;
+}
+
+/**
+ * Has this discovered offer already been evaluated?
+ *
+ * @param {Set<string>} keys - From evaluatedKeys().
+ * @param {(value: unknown, separator?: string) => string} norm
+ * @param {string} company - Scan-history company cell.
+ * @param {string} title - Scan-history title cell.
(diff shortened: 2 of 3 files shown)


Scott-Emberson (commented):
Reviewed as code owner of `tests/`. Scope is `tests/plugins-symlink-discovery.test.mjs`; the rest of the diff is outside my area.

The plugin-discovery test here is byte-identical to #3146's, and the `plugins/_engine.mjs` fix is the same one carried by #3146 and #3141, so this is the third copy of the same #3140 fix. #3141 covers the bug most thoroughly (discovery plus load, lock-pinning and disabled paths), so my suggestion is to consolidate the plugin fix on #3141 and close the duplicate here. Separately, this PR bundles changes unrelated to its stated purpose; splitting them would let each land on its own merits. The engine fix itself is santifer's call.

santifer:
Before you spend time here, @abankar1: #3141 merged the symlink-discovery fix (it won on timing over your #3146, which carried the identical change), so this branch's `plugins/_engine.mjs` and discovery-test hunks are now redundant with main. A rebase dropping those leaves just the whats-new suppression work, which is the part this PR is actually about: smaller diff, cleaner review. The web half will get its look from the web maintainer per the usual split.

santifer:
Thanks @abankar1: solid diagnosis and the sibling-visible / repost-still-hidden test pair is exactly the right guard. This one is entirely `web/` surface (route.ts + lib + tests), so per our subsystem split it goes to the web maintainer's queue: labeling `area:web` and handing it over for the technical call. One trade they'll want to weigh explicitly: with company+role keying, a tracker-vs-scan title divergence can resurface an already-evaluated role once (visible false positive replacing a silent false negative, which #2544 argues is the right direction).

abankar1:
Both done, and rebased again onto current main just now.

**The redundant hunks are gone.** #3141 carries the symlink-discovery fix, so `plugins/_engine.mjs` and `tests/plugins-symlink-discovery.test.mjs` are dropped from this branch — it was cut from #3146's, which is how they got here. The diff is three files, all `web/`:

```
web/src/app/api/whats-new/route.ts
web/src/lib/whats-new-suppression.mjs
web/tests/lib/whats-new-suppression.test.mjs
```

@Scott-Emberson — that removes the third copy of the #3140 test you flagged, so nothing in your area is left in this PR.

**On the trade you want the web maintainer to weigh explicitly**, so it is stated in one place rather than inferred from the diff:

Company+role keying means a tracker-vs-scan title divergence (`"Staff SWE"` vs `"Staff Software Engineer"`) can resurface an already-evaluated role in Explore once. That is a **visible** false positive the user dismisses. What it replaces is a **silent** false negative — an employer's entire board invisible with no signal, which is what the reporter hit and could only diagnose by querying the ATS API directly.

Two things that bound the downside:

- A tracker row with no role keys as `company|`, so a backfilled row (#1799) suppresses only a title-less scan row rather than the whole employer — strictly less suppression than today, never more.
- The key comes from the core's `normalizeTextKey` the route already resolves, so it folds case and whitespace the same way CLI dedup does. It does **not** fold punctuation (`"Acme, Inc."` keys as `acme inc`, not `acme`) — unchanged from the company-only key, and I have a test recording that as a non-contract rather than pinning a legal-suffix rule this key never had.

If the web maintainer would rather not accept the resurfacing at all, the alternative I would reach for is keying on the posting URL where the tracker row has one and falling back to company+role otherwise — `merge-tracker` already tiers its dedup that way. I did not build it because the web tracker parser does not currently expose the URL column, so it is a larger change than the bug warrants; happy to do it if that is the call.

web 344 pass / 0 fail, `test-all --quick` 5308 pass / 0 fail on the rebased head.

santifer (approved):
Reviewed on my surface, and it's a clean fix on the doctrine I care about — approving.

The bug (#3131) is real: keying suppression on company alone meant evaluating one role wiped that employer's *entire* board from Explore, so live sibling roles went silently invisible. Company AND role is the correct key, and the "sibling visible / repost hidden" test pair proves exactly that.

Three things I checked because they're load-bearing:
1. **It injects the core's `normalizeTextKey` rather than reimplementing it** — the module takes the normalizer as a parameter so web dedup keys with the same function the CLI does. A second key helper here is precisely the drift #2666 exists to stop, and you avoided it. Right call.
2. **The trade is in the safe direction** and you documented it honestly: a tracker-vs-scan title/punctuation divergence ("Acme, Inc." vs "Acme", which the core key deliberately does NOT fold) can resurface an evaluated role in Explore *once* — a visible false positive, not a silent false negative. That's the #2544 direction (choose the visible failure), and it self-heals on the next evaluation.
3. The test asserts the core key's actual folding behavior (case/space folded, company punctuation kept) instead of assuming it — so it can't drift from what the core key really does.

Solid. Merges on my next field turn (or say the word and I'll take it).

santifer:
Thanks for the patience on this one, @abankar1 — the rebases were on us. The final shape is exactly right: suppression keyed by the core normalizeTextKey (one body of the normalizer, no drift) and the trade lands on the safe side — a punctuation divergence resurfaces a role visibly instead of hiding it silently, and it self-heals on the next evaluation. Test in web/tests following the established pattern. Merged. 🚀

PR #1827, opened 2026-07-13
feat: expand job discovery and application workflows

Marquis2002:
## What does this PR do?

Adds LinkedIn and Danish job-board providers with deterministic validation fixtures, introduces an evidence-based upskill planning mode, and expands LaTeX CV and cover-letter generation using reusable ai-job-search templates.

Also adds regression tests and privacy defaults that keep personal profiles, resumes, reports, generated outputs, and local application data out of Git.

## Related issue

Closes #<issue-number>

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation / translation
- [ ] Refactor (no behavior change)

## Checklist

- [ ] I have read [CONTRIBUTING.md](https://github.com/santifer/career-ops/blob/main/CONTRIBUTING.md)
- [ ] If this is a new feature or architecture change, I opened an issue first
- [x] My PR does not include personal data
- [ ] I ran `node test-all.mjs` and all tests pass
- [x] My changes respect the Data Contract
- [ ] My changes align with the project roadmap

## Summary by CodeRabbit

* **New Features**
  * Added an upskilling mode with skill-gap heatmaps and targeted learning plans.
  * Added job-board search coverage for LinkedIn, Jobbank, Jobdanmark, Jobindex, and Jobnet.
  * Added LaTeX cover-letter generation alongside PDF output.
  * Added enhanced LaTeX checks for page counts, compilation, and ATS-readable text.

* **Documentation**
  * Updated usage guidance, supported job-board documentation, templates, and configuration examples.

* **Tests**
  * Added coverage for new job-board parsers and LaTeX validation workflows.

PR #3150, opened 2026-08-21
feat(providers): jobbankca — Canada's federal Job Bank (jobbank.gc.ca), part of #230

Zoubeir23:
Part of #230.

Adds `Job Bank` (jobbank.gc.ca) — Canada's federal job-search service, run by Employment and Social Development Canada. A **national employment service**, the same category #230 names alongside Arbeitsagentur (Germany) and VDAB (Flanders): "high-volume public APIs that no commercial aggregator covers well."

## The source

Public, zero-auth Atom feed: `https://www.jobbank.gc.ca/jobsearch/feed/jobSearchRSSfeed`. Despite "RSS" in the path it's Atom (`<feed><entry>`, not `<rss><item>`) — confirmed by fetching it directly, not assumed from the URL. No login, no API key, no session cookie.

```
$ curl -s "https://www.jobbank.gc.ca/jobsearch/feed/jobSearchRSSfeed?searchstring=engineer&locationstring=" | head -20
<?xml version="1.0" encoding="UTF-8"?>
<feed xmlns="http://www.w3.org/2005/Atom" xml:lang="en">
	<title><![CDATA[engineer - Job Bank]]></title>
	...
	<entry>
		<title type="html"><![CDATA[devops engineer]]></title>
		<link rel="alternate" type="text/html" href="https://www.jobbank.gc.ca/jobsearch/jobposting/50123456"/>
		<id>https://www.jobbank.gc.ca/jobsearch/jobSearchRSSfeed?id=10236614247</id>
		<updated>2026-08-20T17:00:00Z</updated>
		<summary type="html"><![CDATA[<strong>Job number:</strong> 10236614247<br /><strong>Location:</strong> Vancouver (BC)  <br /><strong>Employer:</strong> Telescope Innovations Corp<br /><strong>Salary:</strong> $100,000.00 to $140,000.00 annually]]></summary>
	</entry>
```

`robots.txt` (`User-agent: *` / `Crawl-delay: 5`) does not disallow this path — `INTER_REQUEST_DELAY_MS = 5000` honors that delay between every request the provider makes, including the first.

## Design decisions, each checked rather than assumed

**Keyword-driven, not board-wide.** `locationstring` does not reliably filter — measured: `locationstring=Toronto` on a real query still returned Vancouver/Winnipeg/Québec hits. So this follows `arbeitsagentur.mjs`'s recall-first shape: `searchstring` per configured keyword, no location filtering at the source, `scan.mjs`'s own `location_filter` runs afterward.

**Falls back to `config/profile.yml`'s target_roles** when `jobbankca.keywords[]` is empty — the newer convention `vdab.mjs` established, so a user who already onboarded with target roles doesn't have to duplicate them into every keyword-required provider.

**Page size (100) measured directly**, not assumed from any doc — confirmed across three different searches. A page short of 100 ends that keyword's pagination.

**A transient hiccup, found and reported rather than hidden:** one query (`manager`) answered page 1 with 0 entries once, then 100 entries on an immediate retry seconds later. This isn't specially handled — a dry page still ends that keyword's pagination, matching `senjob.mjs`'s convention — but it's the same recall-first tradeoff `arbeitsagentur.mjs` already accepts for a single failed keyword request, and it's documented in the file rather than left as a silent assumption.

**Naming**: `jobbankca`, not `jobbank`. There's an unrelated, unmerged, long-stalled PR (#1827, +2140 lines across 40 files, flagged by @santifer for a rebase weeks ago) that happens to include a `providers/jobbank.mjs` targeting what looks like a Danish board (sibling files `jobdanmark.mjs`, `jobnet.mjs` in the same PR). Different site, different provider — `jobbankca` avoids any naming collision or reviewer confusion either way.

## Implementation

Follows `weworkremotely.mjs`'s tiny-regex XML-tag-extraction approach (no XML dependency) crossed with `senjob.mjs`'s pagination/dedup/max_pages shape, adapted for Atom's `<link href="...">` attribute (vs RSS's `<link>text</link>`) and the summary field's HTML-embedded Location/Employer/Salary labels. Uses the shared `decodeEntities` (never a private copy — this repo has a documented history of that drift, see `_html-entities.mjs`'s own header) and `_profile-keywords.mjs`'s `resolveProfileKeywords`.

## Test plan
- [x] `node test-all.mjs` — 5013 passed, 0 failed (2 pre-existing unrelated warnings: Go dashboard build skip, PII-heuristic false positive on santifer.io in untouched Go source)
- [x] Real feed fetched and inspected directly (not assumed) before writing the parser; fixture in the test file is a trimmed capture of that real response
- [x] `providers/_registry.mjs`'s `loadProviders()` confirmed to load `jobbankca` cleanly alongside the other 76 providers
- [x] Config-fallback tests run in an isolated tmp cwd (never this checkout's own `config/profile.yml`), same pattern as `tests/providers/vdab.test.mjs` — caught and fixed a hermeticity bug in my own first draft, where the test silently passed against this dev checkout's real profile instead of the isolated one
- [x] `docs/SUPPORTED_JOB_BOARDS.md` and `templates/portals.example.yml` updated

## Summary by CodeRabbit

* **New Features**
  * Added Job Bank Canada as a supported job board.
  * Fetches Canadian job listings through Atom feeds with keyword search, pagination, deduplication, configurable page limits, and crawl-delay handling.
  * Added configuration guidance for explicit provider selection and location filtering.

* **Documentation**
  * Documented the Atom feed and available provider settings.

* **Tests**
  * Added coverage for feed parsing, configuration, pagination, filtering, request handling, and failure scenarios.


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/providers/jobbankca.test.mjs`; the provider, docs and portals example are outside my area and I read them for context.

The test is thorough and grounded in a real captured feed rather than a hand-built one. The parse cases keep exactly two of five entries, dropping the missing-link, missing-title and off-host ones, and that off-host drop is the one worth calling out: an `example.com` link is rejected against the `www.jobbank.gc.ca` host pin, so the feed cannot smuggle a URL onto another host. It also pins title, company and location out of the summary CDATA, maps the Atom `<link href>` to url rather than inner text, parses `<updated>` to `postedAt`, decodes entities (`R&amp;D` to `R&D`), and returns an empty array for empty, null or undefined input without crashing.

The `fetch` cases are just as concrete: the `config/profile.yml` `target_roles` fallback runs in an isolated tmp cwd and restores it in `finally`, a clear throw when no keyword source exists, pagination that stops on a short page (two requests, 101 jobs), a `slept >= 2 x 5000ms` assertion that pins the robots.txt `Crawl-delay: 5`, `ctx.maxPages` capping `entry.max_pages` (one request despite `max_pages: 3`), recall-first behaviour where a failing keyword does not abort the others, a total-outage throw, and request hygiene checking `redirect: 'error'` and a User-Agent header. Values and counts are pinned rather than checked for existence, so a broken parse or an inverted maxPages precedence reddens it, and it is hermetic throughout: inline fixtures, a mocked transport, no network.

I grounded the assertions against the provider rather than trusting the suite alone: `parseJobBankFeed` is exported, the off-host link resolves to an empty string and is dropped, and the two error strings the test matches (`no jobbankca.keywords`, `all N keyword request(s) failed`) are the ones the provider actually throws.

Validation: the three OS test legs pass. One thing worth a glance that is not mine to gate and not blocking from my side: CodeQL failed in about three seconds, which reads like a scan-config or infra issue rather than anything in this diff. Flagging it for whoever owns that workflow. The test I am happy with.

Zoubeir23:
All four findings confirmed by execution before fixing, pushed:

**Doc wording ("required").** Left over from before I added the `config/profile.yml` fallback — the row still said the keywords block was required. Reworded to match VDAB's row.

**`linkHref()` took the first `<link>`, not `rel="alternate"`.** Reproduced: an entry with a `self` link before its `alternate` link returned the feed's self-referencing URL instead of the posting page. Now prefers `rel="alternate"` explicitly.

**`<entry>` match required no attributes.** Reproduced: `<entry xml:lang="en">` (legal Atom, not observed on this feed today but not excluded by it either) matched nothing — silent zero, no error. Fixed to tolerate attributes, matching `tagText()`'s existing convention elsewhere in the file.

**CodeQL: incomplete multi-character sanitization.** Reproduced the named concern — `<<script>script>alert(1)` left `script>alert(1)` behind after one regex pass. Fixed with a fixed-point loop (repeat until a pass changes nothing), but that alone doesn't fully close it: a malformed/nested construct can still leave a *lone* unmatched `<` or `>` even at fixed point (verified: `<<a>b>` left `b>` behind — there's never a complete `<...>` span for the loop to catch, since the leading bracket was already consumed by an earlier match). Neither `Location` nor `Employer` is meant to carry markup at all, so I went past the loop to a stricter invariant: any surviving bracket, paired or not, is stripped. Added both probes as regression tests, plus the two you flagged with no direct coverage yet — `max_pages` clamping to `MAX_PAGES_CAP` even when every page is full, and `assertJobBankUrl()`'s host/scheme guards exercised directly.

`node test-all.mjs`: 5022/0 (2 pre-existing unrelated warnings: Go dashboard build skip, PII-heuristic false positive on santifer.io in untouched Go source). New/updated tests: 31/31 in `tests/providers/jobbankca.test.mjs`.

Zoubeir23:
Both valid, confirmed and fixed:

**Doc: RSS → Atom.** The row's own description text already said Atom; the type column didn't match it. Fixed.

**`linkHref()` assumed double-quoted, no-space attributes.** Reproduced the exact scenario: a `self` link in the feed's usual double-quoted form, followed by an `alternate` link using the equally-valid single-quoted XML form, fell through to list order instead of being recognized as alternate. Applied your suggested fix (quote-agnostic, whitespace-tolerant regex for both `rel` and `href`), added the mixed-quote-style case as a regression test.

Pushed. `node test-all.mjs`: 5023/0 (2 pre-existing unrelated warnings). `tests/providers/jobbankca.test.mjs`: 33/33.

Zoubeir23:
Three more, all confirmed by execution before fixing:

**Doc header didn't list Atom.** Added it — `| Board | Type (API / Atom / RSS / parser) | Notes |`.

**`linkHref()` matched `<link-extension>` and `data-rel`/`xlink:href`-style attributes.** `\b` is a *word* boundary, not an element/attribute-name boundary — `-` and `:` are already non-word characters, so `/<link\b/` matches `<link-extension`, and `/\brel=/` matches `data-rel=` the same way. Reproduced both shapes before fixing: a standalone `<link-extension data-rel="alternate" data-href=...>` element, and a genuine `<link>` carrying its real `rel="self"` alongside a smuggled `data-rel="alternate" data-href=...` pair — both picked the wrong (same-host, so untouched by the trusted-host check) URL. Fixed by anchoring precisely: `(?=[\s/>])` after `link` requires the tag name to end exactly there; `(?<=\s)` before `rel`/`href` requires the attribute name to start exactly there — true for a real attribute (always preceded by whitespace), never true for a hyphen/colon-prefixed one.

**Missing whitespace-around-`=` coverage.** Added — the mixed-quote fixture had none, this one does (`rel = 'alternate'`).

Pushed. `node test-all.mjs`: 5026/0 (2 pre-existing unrelated warnings). `tests/providers/jobbankca.test.mjs`: 36/36.

Zoubeir23:
Good catch — and a sharper one than the previous round.

**The previous fix (`\b`→`(?<=\s)`/`(?=[\s/>])` anchoring) solved the hyphen/colon-prefixed-name attack (`link-extension`, `data-rel=`) but not this one.** That fix still worked by *scanning the entire tag text* for a `rel=`/`href=` pattern. Your repro —

```
<link data=" rel='alternate' href='https://www.jobbank.gc.ca/jobsearch/jobposting/WRONG'" rel="self" href="REAL"/>
```

— has `rel='alternate' href='...'` sitting *inside* the `data="..."` value, genuinely preceded by whitespace (just one quote level deeper). No amount of boundary-anchoring on `rel=`/`href=` themselves can distinguish that from a real sibling attribute, because the pattern has no concept of "am I currently inside another attribute's quoted value." Reproduced before fixing — the old code returned `WRONG`.

**Fix (`providers/jobbankca.mjs:145-157`):** replaced the substring-scan with sequential attribute tokenization (`attrsOf`) — a global regex (`/([a-zA-Z_:][-\w:.]*)\s*=\s*(["'])((?:(?!\2)[\s\S])*)\2/g`) walked via `exec()` in a loop. Each match consumes a full `name="value"` pair before `lastIndex` advances past it, so a quoted value's contents are never re-scanned as a sibling attribute — `data="...rel='alternate'..."` is captured whole as `data`'s value, and only the real `rel="self"` / `href="REAL"` pair afterward is tokenized as attributes. `linkHref()` now picks the `<link>` whose parsed `rel` attribute equals `alternate` (case-insensitively), same selection semantics as before.

Verified against your exact repro plus all previously-fixed shapes (`link-extension`, `data-rel`/`data-href`, mixed single/double quotes, spaced `=`) before committing — all pick the correct URL now. Added a regression test (`tests/providers/jobbankca.test.mjs`, "ignores rel=/href= text embedded inside another attribute's quoted value").

`node test-all.mjs`: 5027/0 (2 pre-existing unrelated warnings, +1 vs before from the new test). `tests/providers/jobbankca.test.mjs`: 37/37.

Pushed as `8c548fc`.

santifer:
Textbook provider work, @Zoubeir23: robots.txt fetched and its Crawl-delay honored from the first request, the Atom-not-RSS reality confirmed by fetching instead of assumed from the path, and the locationstring finding (does not reliably filter, measured) turned into the same recall-first choice arbeitsagentur already made. The rel=alternate attribute parsing that refuses to trust whichever link comes first is the detail most feed parsers skip. Merged! 🚀

PR #3152, opened 2026-08-21
fix(updater): settle same-version SHA mismatch on tree content, not SHAs

blayyd:
## Summary

`check` reported `update-available` / `system-files-changed` on every session after any successful `apply` — forever, with no way to clear it. This PR settles same-version SHA mismatches on **tree content** instead of commit SHAs.

## Root cause

- `apply()` never fast-forwards HEAD to upstream main: it checks out upstream system files from `FETCH_HEAD` and commits them as a **new local commit** on the install's own history.
- `check()` (#2630) read `localCommit !== remoteCommit` directly as system-tree drift.
- A freshly created local commit can never share a SHA with upstream's commit (different parents), so after *any* successful update every subsequent check reported drift — and re-running `apply` just creates another diverged commit. Observed on a real install: two consecutive "auto-update system files to v1.28.0" commits, still flagged on every session start.

## Fix

When SHAs differ, verify content before claiming drift:

1. `git fetch --quiet <canonical repo> main` — the same fetch `apply()` already performs.
2. New exported helper `systemTreeDiffers(SYSTEM_PATHS, 'FETCH_HEAD')`: `git diff --quiet --ignore-cr-at-eol FETCH_HEAD HEAD -- <SYSTEM_PATHS>`.
3. Only real committed-content drift reports `system-files-changed`; identical content → `up-to-date`.

Design decisions:

- **Committed state (`HEAD`), not the working tree** — uncommitted local edits to system files are the preserved-edit case `apply()` already handles with `.bak` + messaging (#2337), not an update waiting to happen.
- **`--ignore-cr-at-eol`** — installs that last synced before `.gitattributes` carry CRLF-only blob differences (#2817 rationale; same flag `locallyModifiedSystemFiles` uses).
- **Conservative failure** — an unreadable ref or git error reads as drift, preserving pre-fix behavior whenever content cannot be verified.
- **Computed after the offline early-return** — a machine with no network never pays for a doomed fetch.
- Helper follows the existing `ctx.git` injection seam used by `locallyModifiedSystemFiles`, so tests drive real throwaway repos.

Statuses and reason strings are unchanged; only the false positive disappears.

## Verification

- New `tests/updater-drift-detection.test.mjs` (8 assertions): post-apply divergence ≠ drift; genuine upstream change = drift; user-layer-only differences ignored; uncommitted edits ignored; CRLF-only ignored; unreadable ref fails conservative; empty pathspec short-circuits without invoking git.
- Full `node test-all.mjs`: **4997 passed, 0 failed**, 3 warnings (pre-existing environment noise in a fresh clone: no user data, no Go toolchain, upstream fixture string).
- Live end-to-end against the production GitHub API:
  - Empty commit atop upstream main at the same version (pure post-apply state), fixed checker → `{"status":"up-to-date","local_commit":"85dbd6b…","remote_commit":"051059d…"}` ✅ (old code flagged this exact state)
  - A commit that actually changes a system file → still correctly reported `system-files-changed` ✅

## Notes for reviewers

- The content check adds one incremental `git fetch` to `check()` when SHAs differ — which is the steady state of every healthy post-apply install. `apply()` performs the same fetch on every update, so no new transport or auth surface is introduced.
- The false positive is easy to reproduce on any install that has ever run `apply`: run `node update-system.mjs check` twice with no changes in between — both runs report `update-available`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Update checks now detect actual changes to system files instead of relying solely on commit differences.
  * Line-ending-only changes are ignored to prevent unnecessary update warnings.
  * Fetch or comparison errors are handled conservatively to avoid overlooking potential system drift.

* **Tests**
  * Added regression coverage for system changes, user-only changes, uncommitted edits, unavailable references, and empty path selections.


santifer:
Welcome to career-ops, @blayyd, and what a first contribution: you found why every healthy install reported update-available forever (apply commits upstream content as a NEW local commit, so SHA inequality is the steady state, not drift) and settled it on the only thing that can settle it: content. The four-quadrant test driving real throwaway repos through the production seam, and the conservative every-throw-reads-as-drift choice, are exactly how updater code has to think. Merged — your first contribution. 🚀

PR #3156, opened 2026-08-21
fix(verify-portals): decode the HTML entities a board page serves in its title

### LINKED ISSUE #3155: bug(verify-portals): boardTitleOwner keeps HTML entities, so any company with an ampersand fails its own board identity check
### Code of Conduct

### Existing issues

### What happened?

`boardTitleOwner()` (`verify-portals.mjs:285`, added in #2937 / #2938) reads a board page `<title>` as literal text. A title is markup, so a company whose name contains an ampersand is served as `Hims &amp; Hers`, and an apostrophe often as `&#39;`. The owner name it returns keeps the entity.

`boardIdentityMatches()` then compares that against the company name from `portals.yml`, which holds the real character. They never agree.

The result is a **false negative in the guard #2937 added**: a board that genuinely does belong to the company is judged not to, so the slug repair the check exists to permit is refused. Every company whose name contains `&` is affected, which is a common shape.

### Steps to reproduce

Against `main` at `764f20f`:

```js
import { boardTitleOwner, boardIdentityMatches } from './verify-portals.mjs';

for (const [served, company] of [
  ['Hims &amp; Hers',       'Hims & Hers'],
  ['AT&amp;T',              'AT&T'],
  ['Smith &amp; Nephew',    'Smith & Nephew'],
  ['Johnson &amp; Johnson', 'Johnson & Johnson'],
  ['Ben &amp; Jerry&#39;s', "Ben & Jerry's"],
  ['deepset',               'deepset'],
  ['Diabolocom',            'Diabolocom'],
]) {
  const owner = boardTitleOwner(`<title>${served} Jobs</title>`);
  console.log(served, '->', owner, '| match:', boardIdentityMatches(company, owner));
}
```

Observed:

```
Hims &amp; Hers       -> Hims &amp; Hers       | match: false
AT&amp;T              -> AT&amp;T              | match: false
Smith &amp; Nephew    -> Smith &amp; Nephew    | match: false
Johnson &amp; Johnson -> Johnson &amp; Johnson | match: false
Ben &amp; Jerry&#39;s -> Ben &amp; Jerry&#39;s | match: false
deepset               -> deepset               | match: true
Diabolocom            -> Diabolocom            | match: true
```

**5 of 7 legitimate owners fail to match.** The two entity-free rows still match, so this is specific to entity decoding and not a broader breakage of the comparison.

### Expected behavior

The extracted owner should carry the characters the page actually means: `Hims &amp; Hers` -> `Hims & Hers`.

Two details worth pinning, because both have a wrong way to do them:

1. **Decode exactly one level.** Handling `&amp;` in its own earlier pass turns `&amp;lt;` into `<`, inventing markup the page never served. A single `replace` pass consumes each match where it is found and does not rescan the substituted text, which gives one level naturally.
2. **Decode before collapsing whitespace.** `&nbsp;` is a space once decoded; collapsing first leaves it as literal text inside a name.

An unknown entity name should be left as written. A title reading `Foo &widget; Bar` more likely contains a literal ampersand than an entity worth guessing at.

rubicon:
## What does this PR do?

`boardTitleOwner()` read a board page `<title>` as literal text, so a company whose name contains an ampersand kept the entity (`Hims &amp; Hers`) and never matched the real name in `portals.yml`. This decodes the entities, which closes a false negative in the identity guard added by #2937.

## Related issue

Closes #3155

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

---

### Why it matters

The guard exists to stop the slug repair path adopting a stranger's board. Keeping the entity does not weaken that direction, it breaks the other one: a board that genuinely belongs to the company is judged not to, so a legitimate repair is refused. Measured on `main` at `764f20f`, **five of seven sampled owners containing an ampersand failed their own identity check**, while the two entity-free controls matched.

### Two decisions worth calling out, because each has a wrong way to do it

**One level of decoding, in a single pass.** Handling `&amp;` in its own earlier pass turns `&amp;lt;` into `<`, inventing markup the page never served. A single `replace` consumes each match where it is found and does not rescan the substituted text, so one level falls out naturally. Pinned by a test.

**Decode before collapsing whitespace.** `&nbsp;` is a space once decoded; collapsing first would leave it as literal text in the middle of a name.

An unknown entity name is left as written: a title reading `Foo &widget; Bar` more likely contains a literal ampersand than an entity worth guessing at. Numeric escapes out of range or in the surrogate block also fall back to the raw text rather than throwing or emitting a lone surrogate.

### Test

`tests/board-title-entities.test.mjs`, 14 assertions, auto-discovered under `tests/` so no `test-all.mjs` section was needed.

Written test-first: on unfixed code 10 assertions fail and the 4 controls pass. Then mutation-checked in both directions rather than assumed:

- **Removing the decode call** reddens 10 assertions, led by `a named ampersand entity is decoded (got "Hims &amp; Hers")`.
- **Moving `&amp;` into a separate earlier pass** reddens exactly one: `decoding runs one level only, so &amp;lt; becomes &lt; and not a bracket (got "A < B")` — the invented-markup case.

Negative controls are deliberate, so the suite cannot pass on a `boardTitleOwner` that returned the company name unconditionally or decoded so aggressively it matched anything: an entity-free title is unchanged, the jobs suffix still strips, a title-less page still yields `null`, and `Mercury Systems` still fails against a `Mercury &amp; Co` board.

`node test-all.mjs`: 5080 passed, 0 failed, 2 warnings (both pre-existing and environmental, unrelated to this branch).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved board title processing by decoding named, decimal, and hexadecimal HTML entities.
  * Preserved unknown, invalid, and unsafe entities while maintaining accurate owner-name matching.
  * Improved handling of encoded ampersands, apostrophes, whitespace, and title suffixes.

* **Tests**
  * Added coverage for encoded titles, identity matches, unchanged titles, missing titles, and mismatched employers.

### DIFF
--- a/tests/board-title-entities.test.mjs
+++ b/tests/board-title-entities.test.mjs
@@ -1,0 +1,107 @@
+// tests/board-title-entities.test.mjs — boardTitleOwner() must decode the HTML
+// entities a board page serves in its <title>.
+//
+// Ashby and Lever title the board after its owner, and a title is HTML: an
+// ampersand in a company name arrives as `&amp;`, an apostrophe often as
+// `&#39;`. boardTitleOwner reads the title as literal text, so the owner name it
+// returns still carries the entity. boardIdentityMatches then compares that
+// against the company name from portals.yml, which contains the real character,
+// and the two never agree.
+//
+// The failure is a FALSE NEGATIVE in the guard #2937 added: a board that really
+// does belong to the company is judged not to, so the slug repair it was meant
+// to allow is refused. Every company whose name contains an ampersand is
+// affected, which is not a rare shape.
+
+import { pass, fail, ROOT } from './helpers.mjs';
+import { join } from 'path';
+import { pathToFileURL } from 'url';
+
+const { boardTitleOwner, boardIdentityMatches } = await import(
+  pathToFileURL(join(ROOT, 'verify-portals.mjs')).href
+);
+
+console.log('\nverify-portals.mjs — HTML entities in a board page title (#3155)');
+
+const check = (desc, condition, details = '') => {
+  if (condition) pass(desc);
+  else fail(`${desc}${details ? ` (${details})` : ''}`);
+};
+
+// --- The entity is decoded out of the extracted owner name -------------------
+
+const owner = (title) => boardTitleOwner(`<title>${title}</title>`);
+
+check(
+  'a named ampersand entity is decoded',
+  owner('Hims &amp; Hers Jobs') === 'Hims & Hers',
+  `got ${JSON.stringify(owner('Hims &amp; Hers Jobs'))}`,
+);
+check(
+  'a numeric apostrophe entity is decoded',
+  owner('Ben &amp; Jerry&#39;s Jobs') === "Ben & Jerry's",
+  `got ${JSON.stringify(owner('Ben &amp; Jerry&#39;s Jobs'))}`,
+);
+check(
+  'a hexadecimal numeric entity is decoded',
+  owner('Ben &amp; Jerry&#x27;s Jobs') === "Ben & Jerry's",
+  `got ${JSON.stringify(owner('Ben &amp; Jerry&#x27;s Jobs'))}`,
+);
+
+// A double-encoded ampersand must decode exactly one level. Decoding `&amp;`
+// before the other named entities would turn `&amp;lt;` into `<`, inventing
+// markup the page never served.
+check(
+  'decoding runs one level only, so &amp;lt; becomes &lt; and not a bracket',
+  owner('A &amp;lt; B Jobs') === 'A &lt; B',
+  `got ${JSON.stringify(owner('A &amp;lt; B Jobs'))}`,
+);
+
+// --- The identity comparison the decode exists to serve ----------------------
+
+const IDENTITY_CASES = [
+  ['Hims &amp; Hers', 'Hims & Hers'],
+  ['AT&amp;T', 'AT&T'],
+  ['Smith &amp; Nephew', 'Smith & Nephew'],
+  ['Johnson &amp; Johnson', 'Johnson & Johnson'],
+  ['Ben &amp; Jerry&#39;s', "Ben & Jerry's"],
+];
+
+for (const [served, company] of IDENTITY_CASES) {
+  const extracted = owner(`${served} Jobs`);
+  check(
+    `${company} matches its own board title`,
+    boardIdentityMatches(company, extracted),
+    `extracted ${JSON.stringify(extracted)}`,
+  );
+}
+
+// --- Negative controls -------------------------------------------------------
+// Without these the suite would pass on a boardTitleOwner that returned the
+// company name unconditionally, or that decoded so aggressively it matched
+// anything.
+
+check(
+  'a title with no entities is unchanged',
+  owner('deepset Jobs') === 'deepset',
+  `got ${JSON.stringify(owner('deepset Jobs'))}`,
+);
+check(
+  'the jobs suffix is still stripped case-insensitively with surrounding space',
+  owner('  Lever Demo 2 jobs ') === 'Lever Demo 2',
+  `got ${JSON.stringify(owner('  Lever Demo 2 jobs '))}`,
+);
+check(
+  'a page with no title still yields null',
+  boardTitleOwner('<html><body>no title</body></html>') === null,
+);
+check(
+  'a different employer still fails to match',
+  !boardIdentityMatches('Mercury Systems', owner('Mercury &amp; Co Jobs')),
+  `extracted ${JSON.stringify(owner('Mercury &amp; Co Jobs'))}`,
+);
+check(
+  'an entity-only title does not decode into an empty owner treated as a match',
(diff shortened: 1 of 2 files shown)


santifer:
The diagnosis is right and the test file is exactly the evidence this needed, @rubicon: the #2937 guard refusing legitimate repairs for every ampersand-carrying company is a real false negative, and your single-pass reasoning (`&amp;lt;` must become `&lt;`, not a bracket) is correct.

One structural ask before merge: this adds a sixth private entity decoder, and the repo just finished consolidating the fifth (#3007 merged `providers/_html-entities.mjs` as the shared decoder precisely to end this drift). Import `decodeEntities` from there instead of the local `decodeHtmlEntities`: it is already single-pass (same CodeQL #2962 rationale you wrote), passes unknown names through untouched (same `&widget;` behavior), guards surrogates and out-of-range code points, and carries the full Latin-1 letter table your six-name map does not: a board titled `Soci&eacute;t&eacute; G&eacute;n&eacute;rale Careers` still hits your false negative with the local decoder, and decodes correctly with the shared one. Your test survives verbatim; worth adding one `&eacute;` case to pin the widened coverage.

Merges on that push: the substance is done.

Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is the new `tests/board-title-entities.test.mjs`; `verify-portals.mjs` is outside my area and I read it for context.

The test pins the decode where it matters. `boardTitleOwner` is checked against a named ampersand (`&amp;`), a decimal apostrophe (`&#39;`) and its hex form (`&#x27;`), and then the case worth calling out: `&amp;lt;` decodes to `&lt;`, not to a bracket, so a double-encoded title cannot invent markup the page never served. `boardIdentityMatches` is then exercised on real names that carry the character (Hims & Hers, AT&T, Smith & Nephew, Johnson & Johnson, Ben & Jerry's), which is the false-negative this fix removes: a board that really does belong to the company was being judged not to.

The negative controls are what make it honest: a title with no entities is returned unchanged, the `jobs` suffix is still stripped case-insensitively, and a page with no title still yields null. Without those a `boardTitleOwner` that returned the company name unconditionally, or decoded so aggressively it matched anything, would pass. Both exports are present, and it is hermetic (pure functions, no network).

rubicon:
Done, @santifer, and you were right about the drift: the local decoder is gone and this now imports `decodeEntities` from `providers/_html-entities.mjs`. The `verify-portals.mjs` side of the diff went from +41/-1 to **+2/-38**.

Three assertions added for the widened coverage:

- `a Latin-1 letter entity is decoded` — `Soci&eacute;t&eacute; G&eacute;n&eacute;rale Jobs` -> `Société Générale`
- `Société Générale matches its own board title` — the identity match that decode unblocks
- `an uppercase letter entity keeps its case` — `&Eacute;ditions Gallimard` -> `Éditions Gallimard`

That last one is a bug the swap fixed rather than just coverage: my local map looked names up lowercased, so `&Eacute;` would have decoded to a lowercase `é` and quietly altered a company's name. Your case-sensitive lookup is correct and the comment explaining why is what made me go check.

The existing 14 assertions survived verbatim. Removing the decode call still reddens 13 named assertions. `node test-all.mjs`: 5134 passed, 0 failed, 2 warnings.

### On the macOS red, which I do not believe is this PR

Head `d679180` is green on ubuntu, windows and the other 10 checks. macOS reports 3 failures, all in the same place:

```
❌ an accepted page budget changed the rendered PDF instead of acting only as a post-render decision
❌ generate-pdf default page budget regressed: Refusing to write the PDF outside the tracker workspace:
   input escapes the tracker workspace: .../output/page-budget-test-6B3bOj/three-pages.html
❌ generate-pdf strict page budget regressed: ... two-pages.html
```

What I checked before saying that:

- This PR touches two files, `verify-portals.mjs` and a new test. `generate-pdf.mjs` on this branch is **byte-identical** to `main` (`git rev-parse HEAD:generate-pdf.mjs` equals `upstream/main:generate-pdf.mjs`).
- The branch is rebased onto `7d97d14` and **already contains `e4a21ba`**, so the canonicalization-vs-real-escape split is in this run.
- Reproduced locally on macOS: rendering an HTML file from `output/<subdir>/` succeeds, so it does not reproduce off the runner.
- The `Possible personal data in dashboard/internal/ui/screens/stats.go: "santifer.io"` line appears on all three platforms including the two that pass, so it is not the cause.

The path being rejected is inside `output/`, which is the shape `e4a21ba` was aimed at, so this may be a remaining macOS-only case of the same thing rather than anything new. I cannot re-run the job to check whether it is flaky (no admin), so I have not established whether it reproduces.

Happy to open a separate issue with the above if it is useful, but I did not want to file noise against work you may already have in flight.

santifer:
The redo is the textbook outcome, @rubicon: your verify-portals side went from +41/-1 to +2/-38 by importing the shared decoder, the eacute case that the six-name map left open is now covered, and the false negative on every ampersand-carrying company is dead. Third merge of yours today, which is this repo's per-author ceiling — a pacing rule, not a hint. Merged! 🚀

PR #3158, opened 2026-08-21
fix(cli): validate browser-extract flags and honour both value forms

### LINKED ISSUE #3004: fix(cli): browser-extract.mjs silently ignores mistyped flags — delegate to lib/cli-flags.mjs
**Time:** ~40 min · **Difficulty:** Beginner

`browser-extract.mjs` takes value-carrying flags (`--mode`, `--max-chars`, `--timeout`, `--max`) but never rejects an unknown one: `--max-char` (singular) is silently ignored and the extraction runs with the default cap, which is exactly the silent-wrong-answer class of #2919.

**What to do**

Delegate flag validation to the shared helper, like `doctor.mjs` since #2874 — note this one has flags WITH values, so use the helper's value-flag support (see how `batch-tailor.mjs` declares `--min-score` after #2961):

1. Import from `./lib/cli-flags.mjs` and declare: `--mode`, `--max-chars`, `--timeout`, `--max` (with values) plus `--help`/`-h`.
2. Reject unrecognized flags BEFORE launching any browser work.
3. Add a small test following `tests/cli-flags.test.mjs`.

**Acceptance**

- `node browser-extract.mjs --max-char 100` (typo) exits non-zero naming the bad flag.
- `--help` prints usage and exits 0 without launching a browser.
- Valid-flag behavior unchanged; `node test-all.mjs` green.

IaroslavMazur:
Closes #3004.

## What does this PR do?

Routes `browser-extract.mjs`'s argv handling through `lib/cli-flags.mjs`: `validateFlags` in front of any browser work, and `flagValue`/`hasFlag` for the value reads, so a mistyped flag is refused by name and both accepted flag forms actually reach the extractor.

## Problem

The script hand-rolled its parsing and matched tokens **_exactly_** against a local `FLAGS` set. Measured on `764f20f8`, that left three separate failures, not one:

1. **`--max-chars=50000` was dropped.** It is not in `FLAGS`, and it starts with `--`, so it was not taken as the URL either. The run proceeded at the 12000 default and returned a JD truncated at the tail, exit 0. This is the `--from=…` class from #2401/#2402, and it is the half a validation gate alone cannot fix - the flag clears the gate and is then discarded anyway.

2. **A typo adjacent to a value stole the URL.** `--max-char 5000 https://example.com/job` skipped the unknown token, took `5000` as the URL and discarded the real one:

   ```
   {"error":"invalid URL","code":"invalid_url"}
   ```

   The one thing the caller got right is the thing it names.

3. **A typo away from a value was fully silent.** `<url> --bogus` was skipped entirely and the scan ran to completion, exit 0. `--help` exited 1 with a `no_url` error instead of printing usage, and `-h` has a single dash, so it was read as the URL and failed with `invalid URL`.

## Change

- **`validateFlags(args, KNOWN_FLAGS, USAGE, { valueFlags: VALUE_FLAGS, requireOperand: true })`**, before the URL check and before Playwright loads. `requireOperand` because this script has nothing more specific to say about a missing operand than the shared message, and without it `--max-chars --help` prints usage and exits 0 with the malformed flag never reported - the ordering CodeRabbit caught on #2745/#2746 and #2961.
- **`parseArgs` reads values via `flagValue`/`hasFlag`**, so `--max-chars=50000` and `--max-chars 50000` are the same request. The URL is still found positionally, with an adjacency rule mirroring `validateFlags`' own so a flag's space-separated value is never mistaken for it.
- **One `USAGE_SYNOPSIS` constant** behind both `--help` and the `no_url` error. Those had already drifted: the error's copy omitted `--timeout`.

## Compatibility

Behaviour deliberately preserved, each asserted:

- the URL is found on either side of the flags;
- `--max 0` is honoured rather than replaced by the default, and out-of-range or non-numeric values still fall back instead of propagating `NaN`;
- an unknown `--mode` **_value_** still reaches the extractor's own `bad_mode` error rather than being refused by the gate, since `validateFlags` only knows flag names;
- the module's exports are unchanged, so `doctor.mjs`'s `resolveExtractorMode` read is untouched.

The one intended behaviour change beyond the fixes: previously-ignored garbage now exits 1. That is the point of the issue, and it is the same break every prior adopter of this helper took (#1635, #2745, #2746, #2961).

## Tests

`tests/browser-extract-flags.test.mjs`, 11 assertions, run against the real binary as a subprocess because the defect is in what the CLI does end to end. Every error path exits before Playwright launches, so none of them needs a browser or a network.

Seen red on `764f20f8`: **8 of the 11 fail** there, including `--bogus` actually launching a browser and fetching the page. The 3 that pass on both sides are the regression guards above, which is what they are for.

`node test-all.mjs`: 5076 passed, 0 failed, 3 warnings. All 3 are pre-existing and environmental - `cv-sync-check` without user data, no Go compiler for the dashboard build, and a `santifer.io` string in `dashboard/internal/ui/screens/stats.go`. Same 3 on an unpatched checkout, which reports 5065 passed.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for both `--flag=value` and space-separated CLI option formats.
  * Improved URL detection regardless of argument position.
  * Added clearer help and usage guidance.

* **Bug Fixes**
  * Added validation for unknown flags, missing values, and numeric option ranges.
  * Improved handling of invalid command-line input before browser launch.

* **Tests**
  * Added end-to-end coverage for CLI parsing, validation, help behavior, and option handling.

### DIFF
--- a/browser-extract.mjs
+++ b/browser-extract.mjs
@@ -36,6 +36,7 @@ import { join, dirname } from 'path';
 import { fileURLToPath, pathToFileURL } from 'url';
 import * as yaml from 'js-yaml';
 import { LIVENESS_CONTEXT_OPTIONS, rejectPrivateOrInvalid } from './liveness-browser.mjs';
+import { flagValue, hasFlag, validateFlags } from './lib/cli-flags.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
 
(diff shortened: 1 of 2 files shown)


santifer:
Clean instance of the #2919 family, @IaroslavMazur, with the two touches that make it more than mechanical: documenting the exact silent failure (a --max-chars=50000 caller silently truncated at 12000) and unifying the synopsis so --help and the no_url error can never drift apart again: you even caught that the error's copy already omitted --timeout. Merged! 🚀

PR #3157, opened 2026-08-21
fix: honor custom font family for CJK templates

alinaqvi1129:
## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation / translation
- [ ] Refactor (no behavior change)

## Checklist

- [ ] I have read [CONTRIBUTING.md](https://github.com/santifer/career-ops/blob/main/CONTRIBUTING.md)
- [ ] If this is a new feature or architecture change, I opened an issue first (bug fixes, providers, docs & translations are exempt — send those straight in)
- [ ] My PR does not include personal data (CV, email, real names, scan results, or pipeline data)
- [ ] I ran `node test-all.mjs` and all tests pass
- [ ] My changes respect the [Data Contract](https://github.com/santifer/career-ops/blob/main/DATA_CONTRACT.md) (no modifications to user-layer files)
- [ ] My changes align with the [project roadmap](https://github.com/santifer/career-ops/discussions/156)

---
Questions? [Join the Discord](https://discord.gg/8pRpHETxa4) for faster feedback.

## Summary by CodeRabbit

* **Bug Fixes**
  * Japanese and Simplified Chinese CV templates now honor custom font settings.
  * Preserved appropriate locale-specific fallback fonts for consistent typography.

* **Tests**
  * Added regression coverage for localized font overrides, fallback behavior, and duplicate selector prevention.

PR #3161, opened 2026-08-21
fix(web): pluralize the Analytics evaluation count

tasmiarahmantanjin:
The Analytics page subtitle is hardcoded to the plural, so a user with one tracked role sees:

> Across **1 tracked evaluations.**

That's the state every user is in immediately after their first evaluation, so it's the first thing a new user reads on that page.

## Fix

`web/src/app/analytics/page.tsx`

```diff
-<p className="mt-1 text-sm text-muted">Across {total} tracked evaluations.</p>
+<p className="mt-1 text-sm text-muted">Across {total} tracked evaluation{total === 1 ? "" : "s"}.</p>
```

This follows the pattern already used elsewhere in the app rather than introducing a new one:

- `components/home/today-dashboard.tsx` — `new match{newThisWeek === 1 ? "" : "es"}`, `follow-up{overdue === 1 ? "" : "s"} due`
- `components/portals-view.tsx` — `{broken.length === 1 ? "company silently drops" : "companies silently drop"}`
- `components/explore/results-list.tsx` — `fresh role${offers.length === 1 ? "" : "s"}`
- `components/explore/explorer-view.tsx` — `compan${companiesScanned === 1 ? "y" : "ies"}`

`0` and `2+` still read "evaluations", as before.

## Testing

- `npm run typecheck` — clean
- `npm test` — 336/336 pass
- `npm run build` — succeeds
- Verified in the dev server with one tracked role: the subtitle now reads "Across 1 tracked evaluation."

## Summary

The Analytics page now displays singular grammar for one tracked evaluation: `Across 1 tracked evaluation.` It keeps `evaluations` for zero or multiple evaluations (`web/src/app/analytics/page.tsx:54`).

Only this page changes. `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/` are not changed. No user-visible behavior breaks.


tasmiarahmantanjin:
`test (macos-latest)` failed here on 3 assertions, all in `tests/generate-pdf-page-budget.test.mjs`:

```
❌ an accepted page budget changed the rendered PDF instead of acting only as a post-render decision
❌ generate-pdf default page budget regressed: Refusing to write the PDF outside the tracker workspace:
   input escapes the tracker workspace: .../output/page-budget-test-T88bzT/three-pages.html
❌ generate-pdf strict page budget regressed: ... two-pages.html
```

That's the `assertInsideWorkspace` containment guard in `generate-pdf.mjs`. The test builds its sandbox with `mkdtempSync(join(ROOT, 'output', 'page-budget-test-'))` and points `CAREER_OPS_TRACKER` at `<sandbox>/applications.md`, so the input is inside the workspace it's being checked against. The guard canonicalises through `realpathSync`, which is where macOS differs from the other two runners.

I don't think this is related to the change here — it's a one-line JSX edit in `web/src/app/analytics/page.tsx` and can't reach `generate-pdf.mjs`. Supporting evidence:

- `test (ubuntu-latest)` and `test (windows-latest)` passed on the same commit; only macOS failed
- running the full suite locally on macOS gives `5086 passed, 0 failed` (CI reported `5069 passed, 3 failed`)
- #3157 and #3158, both merged today, passed `test (macos-latest)`
- every other check here is green, and the PR shows as mergeable

Could someone with write access re-run that job? I tried and got `Must have admin rights to Repository`, which is expected from a fork.

Also happy to rebase — this branch is based on `5291cc7` and `main` has moved on since, so a fresh run on current `main` may just pass. Just say the word.

santifer:

santifer:
Sorry for the long wait, @tasmiarahmantanjin: I said in August this was in the web queue, and then it sat there for weeks. That one is on us, not on you.

The fix is right, and the way you argued it is why it needs no second round: you matched the pattern the app already uses in today-dashboard, portals-view and the explore views instead of inventing a new one, and 0 and 2+ still read exactly as before.

The macOS failure you flagged back in August was an unrelated flake in the PDF page-budget suite, nothing to do with this line. Checks are green on your current commit.

Bringing the branch up to date with main and re-running CI is on me, and then it goes in. Thanks for the patience.

santifer (approved):
Approach and change reviewed: the rendered count and its noun now agree, and it follows the pluralization pattern already used across the app. Merging once the required checks finish on this commit.

Issue #3162, opened 2026-08-21
flaky (macOS CI): page-budget test fails with 'input escapes the tracker workspace' on a path inside its own sandbox


santifer:
## The signal

Two contributors hit the same intermittent failure on macOS CI the same day, in unrelated PRs, and both logged it instead of re-running past it (#3123, #3145 — thank you both):

```
❌ generate-pdf default page budget regressed: Refusing to write the PDF outside
   the tracker workspace: input escapes the tracker workspace:
   /Users/runner/work/career-ops/career-ops/output/page-budget-test-XXXX/three-pages.html
```

The path is INSIDE the sandbox the test itself created. A re-run goes green.

## What has been ruled out

- **Not reproducible in isolation**: 30 consecutive local runs of `node --test tests/generate-pdf-page-budget.test.mjs` on macOS (same OS family as the runner), 0 failures.
- **Not cross-suite env contamination**: the two suites that mutate `process.env.CAREER_OPS_TRACKER` (tracker-utils, batch-evaluate) run as separate subprocesses; their mutations die with them.
- **Not a cross-suite race on `output/`**: test-all runs node:test suites sequentially (the only `Promise.all` pool is the read-only syntax check).

## The blind spot that keeps this undiagnosable

`assertInsideWorkspace()` in `generate-pdf.mjs` is fail-closed (correct), but its catch collapses two different events into one message: a path that genuinely escapes, and a `realpathSync` that failed for any reason (raced away, EPERM, ENOENT). On CI we cannot tell which one fired. Until the message distinguishes them, every recurrence is un-attributable.

A companion PR makes the catch name the canonicalization failure and its `err.code`. After it lands, the next CI hit will tell us which branch fired and carry the errno: that turns this from a shrug into a diagnosis.

## If you hit this

Please paste the full assertion output (post-companion-PR it will include the cause) plus the platform, rather than only re-running. Two data points found the pattern; the third will likely find the cause.

santifer:
Diagnostic update, from the flake reproducing on #3163's own CI (fourth occurrence, third day):

[...]

PR #3163, opened 2026-08-21
fix(pdf): name canonicalization failure separately from a real workspace escape

santifer:
Part of #3162's diagnosis path.

`assertInsideWorkspace`'s catch threw the same *escapes the tracker workspace* message as the lexical containment check, so the intermittent macOS CI failure in #3162 could not be attributed: was it a real escape or a failed `realpathSync`? The catch now says which event fired and carries `err.code` plus the probe path.

- Fail-closed semantics unchanged: any canonicalization failure still throws.
- The only test asserting the message (`generate-pdf-batch.test.mjs:312`) exercises the lexical branch, which keeps its wording.
- Both generate-pdf suites green locally (exit 0).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messages for PDF generation failures by identifying the specific failure code and affected paths.
  * Prevented canonicalization errors from being incorrectly reported as workspace path violations.

### DIFF
--- a/generate-pdf.mjs
+++ b/generate-pdf.mjs
@@ -81,11 +81,17 @@ function assertInsideWorkspace(absPath, label) {
   let canonical;
   try {
     canonical = existsSync(probe) ? resolve(realpathSync(probe), ...tail) : absPath;
-  } catch {
+  } catch (err) {
     // Canonicalization failed (realpath raced away, permission error): containment
     // is unprovable, so fail closed rather than fall back to a lexical form that a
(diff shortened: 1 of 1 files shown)


santifer:
Maintainer fix on the #3162 diagnosis path: the workspace guard's catch collapsed a failed realpath and a real escape into one message; now each names itself, with errno and the probe path, so the next flaky CI hit identifies its own cause. The retry itself proved the point: the recurring failure exercises the lexical branch, not the catch.

PR #3167, opened 2026-08-21
feat(providers): mycareersfuture — Singapore's national job bank (MyCareersFuture), part of #230

Zoubeir23:
## Summary

Adds a `mycareersfuture` provider for [MyCareersFuture](https://www.mycareersfuture.gov.sg) — Singapore's national job bank, run by Workforce Singapore (WSG), a statutory board under the Ministry of Manpower. Same class as Arbeitsagentur/VDAB/Job Bank Canada: a high-volume public employment service with no login and no API key, explicitly called out in #230 as an especially welcome pattern (public employment services + Southeast Asia coverage).

Part of #230.

## Details

- **Public API, no auth.** `POST https://api.mycareersfuture.gov.sg/v2/search` — confirmed live with a cold, unauthenticated request. `www.mycareersfuture.gov.sg/robots.txt` allows everything and sets no `Crawl-delay`, so unlike Job Bank Canada's provider there is no forced inter-request delay here.
- **Pagination bug I nearly shipped, caught by testing it both ways:** the search endpoint accepts a `page` field in both the URL query string and the JSON body. Only the **query string** actually advances pages — I confirmed live that varying the body's `page` while holding the query string fixed returns identical results every time. The provider builds the URL with `?limit=N&page=N`, and a regression test specifically asserts pages advance via the URL, not the body (a test that only varied `body.page` would have passed even on a provider that never changed the URL at all).
- `limit` (page size) is server-validated at `<= 100` — confirmed live, a `>100` request 400s with a clear message. `size` config is clamped to that ceiling.
- **Keyword fallback** to `config/profile.yml`'s `target_roles` when `mycareersfuture.keywords[]` is empty on the entry — same convention as `vdab.mjs`/`jobbankca.mjs`.
- **Host-locked job URLs.** `cleanUrl()` validates every `jobDetailsUrl` from the API response is `https://www.mycareersfuture.gov.sg/...` before it's trusted as the job's `url` (dedup key) — defense in depth, same discipline `jobbankca.mjs` applies to its Atom `<link href>`.
- `company` prefers `hiringCompany` (the real employer) over `postedCompany` (the poster, which is a recruitment agency when a listing is posted on behalf of someone else) — live samples never populated `hiringCompany`, but the schema clearly reserves it for exactly this case.
- Self-identifying `DEFAULT_USER_AGENT` (from `_http.mjs`) is left as-is rather than swapped for a browser-like one: a live request carrying it verbatim returned 200 with no WAF friction, so there's no reason to impersonate a browser against a government API that already answers an honest bot identifier.
- No location/geo filter option — results carry only coarse district names (e.g. "D01 Marina, Raffles Place"), so precision is left to `location_filter` downstream, same recall-first approach VDAB and Job Bank Canada use.

## Files

- `providers/mycareersfuture.mjs` — the provider (`parseConfig`, `cleanUrl`, `normalizeJob`, `detect`, `fetch`)
- `tests/providers/mycareersfuture.test.mjs` — 38 assertions, no network, hermetic `config/profile.yml` fallback test
- `docs/SUPPORTED_JOB_BOARDS.md` — one row
- `templates/portals.example.yml` — added to the explicit-provider-required list + a copy-paste example block

## Test plan

- [x] `node test-all.mjs` — 5127 passed, 0 failed (2 pre-existing unrelated warnings)
- [x] `tests/providers/mycareersfuture.test.mjs` standalone — 38/38
- [x] `npm run lint` (syntax check) — clean
- [x] Live smoke test against the real API during development (not part of the committed test suite — no network in tests, per the repo's contribution guide)

## Summary by CodeRabbit

* **New Features**
  * Added MyCareersFuture as a supported Singapore job board.
  * Supports keyword searches, profile-based keyword fallback, pagination, configurable page limits, and custom page sizes.
  * Standardizes results and removes duplicate job listings.

* **Documentation**
  * Added MyCareersFuture configuration guidance and example settings.

* **Tests**
  * Added comprehensive coverage for configuration, searching, pagination, errors, validation, and result handling.


Zoubeir23:
Good catch — fixed.

**`cleanUrl()` (`providers/mycareersfuture.mjs:76-99`) checked `.protocol` and `.hostname` only.** `.hostname` alone already can't be fooled by the classic `https://TRUSTED_JOB_HOST@evil.example/` userinfo trick — `URL.hostname` extracts only the real host (`evil.example` there), so that specific phishing shape was never a bypass here. But a URL carrying a **non-default port** or **`user:pass@` userinfo on the REAL trusted host** — `https://www.mycareersfuture.gov.sg:9999/job/evil` or `https://user:pass@www.mycareersfuture.gov.sg/job/x` — still passed the `.hostname`-only check. Reproduced before fixing: the old code returned both unchanged.

Neither has a legitimate reason to appear in this feed (the real site only serves 443, and there's no reason a government job posting URL would carry HTTP Basic Auth credentials), so both are now rejected outright: `parsed.port === ''`, `parsed.username === ''`, `parsed.password === ''`. `.port` is already normalized to `''` for the default HTTPS port by `URL` parsing (confirmed by execution), so an explicit `:443` still passes unchanged.

Added 4 regression tests (`tests/providers/mycareersfuture.test.mjs`) — non-default port, `user:pass@`, bare username, and the explicit-`:443`-still-works case.

`node test-all.mjs`: 5131/0 (2 pre-existing unrelated warnings). `tests/providers/mycareersfuture.test.mjs`: 42/42.

Pushed as `a39a95d`.

santifer:
Another one for the map, @Zoubeir23: Singapore's national job bank, with the same rigor as your jobbankca work. The pagination regression test is the highlight: catching that the server ignores body.page and only honors the query string, then writing the test that would fail for a provider that only varies the body, protects the real invariant instead of the current code. Also appreciated: keeping the honest default User-Agent because the API answers a self-identified bot just fine. Merged! 🚀

Zoubeir23:
Thanks! Glad the pagination test landed — that one felt worth writing carefully since a shallow version would've passed on broken code.

PR #3171, opened 2026-08-21
feat(detect-reposts): skip multi-employer aggregator boards, and say the check was skipped

### LINKED ISSUE #2703: feat(detect-reposts): aggregator companies in tracked_companies inflate postingChurn — mark them in config
### Code of Conduct

### Existing issues

Closest prior art I found, none of which covers this: #1750 (`company_aliases`, closed/completed — scan-time dedup, and its config docs explicitly scope it *out* of repost classification), #1711 (the original `company-history.mjs` evidence cards), #2383 (title-grouping performance).

### Problem

When a `tracked_companies` entry is a **multi-employer aggregator** rather than a single employer, `detect-reposts.mjs` treats every posting on its board as belonging to one company. Different employers' roles that happen to share a title then cluster together, and `company-history.mjs` reports `postingChurn: reposts-detected` for a "company" that never reposted anything.

This is not an exotic setup — `templates/portals.example.yml` ships aggregators inside `tracked_companies` (e.g. `joinup.ch`, `provider: joinup`, described in the file itself as a "Swiss startup job platform"), so a default install can hit it.

Measured on a real 2341-row `scan-history.tsv`, after #2702 removes the concurrent-openings and sibling-requisition defects: **24 of the 68 remaining clusters (35%) come from a single aggregator**, including individual clusters of 23 and 22 sightings. Every non-aggregator cluster in the same corpus has fewer than 5.

The reason it can't be solved in the scanner is worth stating: career-ops already has the right concept for this — `job_boards:` entries record the *true underlying employer* per posting (`providers/remotive.mjs` reads `company_name` and only falls back to the board name). An aggregator reached through a generic ATS provider can't use that path, because the ATS attributes every posting to the aggregator's own account and exposes no field for the partner employer. The information is simply not in the feed.

### Proposed solution

Mark aggregators in config and have the post-tracking layer skip them — mirroring the `company_aliases` precedent from #1750: an optional key, absent by default, no behaviour change for anyone who doesn't set it.

```yaml
tracked_companies:
  - name: SomeAggregator
    careers_url: https://jobs.lever.co/someaggregator
    provider: lever
    aggregator: true          # multi-employer board: identical titles are different employers
```

`detect-reposts.mjs` skips companies flagged this way; `company-history.mjs` then reports `none-detected` for them instead of a fabricated churn signal. Everything else is untouched.

This lands squarely in the layer `portals.example.yml` already delineates — *"It does not classify repost patterns or produce `postingChurn` evidence; detect-reposts.mjs and company-history.mjs own that post-tracking layer."*

Open questions for you, which is why this is an issue and not a PR:

1. **Where should the flag live** — per-entry in `tracked_companies` (above), or a separate list, or a system-layer default set of known aggregator names that users extend?
2. **Skip or annotate?** Suppressing the clusters is the clean read. Alternatively the clusters could carry an `aggregator: true` marker and let the consumer decide, which keeps the evidence visible.
3. Should `scan.mjs` share the flag for anything, or is this purely post-tracking?

### What I already ruled out

A config-free heuristic. I expected "many sightings across few dates = a board, not a repost" to separate them, and measured it before building rather than after:

| | sightings (median) | distinct dates | max URLs sharing one date |
|---|---|---|---|
| aggregator clusters | 2 | 2 | 1 |
| all other clusters | 2 | 2–3 | 1 |

They are statistically identical; only the extreme tail differs. The best threshold I could find (≥4 URLs sharing one date) removes 10 of the 24 aggregator clusters and 0 legitimate ones — it leaves **more than half** the noise in place while introducing a magic number. Recording it here so nobody re-derives it.

Happy to send the PR once you've picked a shape for (1) and (2).

### Area

Other — the post-tracking analytics layer (`detect-reposts.mjs` / `company-history.mjs`), configured from `portals.yml`.

artemtrofymenko:
Closes #2703.

A company marked `aggregator: true` in `portals.yml` is a multi-employer board. Its "same title twice" is many different employers, not one employer re-listing, so every cluster the detector builds for it is fabricated. The flag skips it, and the report says the check was **skipped** rather than that it found nothing.

## Why a per-entry flag rather than a derived signal

Your question on the issue was whether this could be derived — *the same posting appearing under N distinct employer names* — and I measured both readings on a 2,470-row scan-history rather than answer from taste.

**By title identity:** of the aggregator's 66 distinct titles, 11 also appear under another company. It catches 17% and misclassifies ordinary employers, because titles under many companies are simply common titles — `Account Manager` appears at 26.

**By fingerprint,** the fairer reading: the column is populated on **69 of 2,470 rows (2%)**. There is not enough of that field to derive anything from.

The concern behind the question is right, though, and it is worth separating: a *system-shipped list that must track an external world* goes stale silently, because the world moves and the list does not. A per-entry flag is not that — it is one key on an entry the user typed themselves, next to the board it describes. It tracks nothing external, so there is nothing to fall behind. `portals.example.yml` ships it set on the two multi-employer boards it already lists, so a default install is correct without anyone maintaining a registry.

## Surfaced, not suppressed

`postingChurn: aggregator-not-evaluated`, not folded into `none-detected`. The enum already separates *could not run* (`no-scan-data`) from *ran and found nothing*, and a skipped company belongs on the first side: reporting "no reposts" would assert a negative result from a check that never happened.

That half needed a second fix, found by checking the claim instead of asserting it. The JSON card was right; `--summary` was not. A flagged board has no tracker rows and now no clusters either, so it fell out of the card key set and **vanished from the report entirely** — which is the silence the label exists to prevent, and a regression besides: before the flag, the company at least appeared. `buildCompanyCards` now seeds its key set from the flagged companies as well.

## Tests, and a correction

On the issue I wrote that the vanish fix had "four checks pinned on it". **It did not.** Those ran as a throwaway during that session and were never committed — nothing in this branch or its history contains them. The claim was wrong on a public thread and I would rather say so than let it stand while quietly making it true.

It is true now. `tests/aggregator-companies.test.mjs`, a discovered file per the pattern #2879 set and #3050 followed, 8 assertions. Each is paired against the same input **without** the flag, because the failure mode is a false negative: rows that never clustered anyway would let "no clusters for the aggregator" pass on a build where the flag does nothing.

Seen red under four deliberate regressions, each printing proof that it applied:

| reverted | failure |
|---|---|
| `detectReposts` ignores the aggregator set | `flag did not decide the cluster: without=1 with=1` |
| `loadAggregatorCompanies` stops reading the flag | `aggregator flag misread: board=true acme=true explicitFalse=true` |
| `computePostingChurn` drops the separate label | `churn labels wrong: aggregator=none-detected clean=none-detected` |
| `buildCompanyCards` stops seeding keys | `the aggregator vanished from the report: ["Acme"]` |

Rows are built through `parseScanHistory` from TSV text rather than hand-shaped objects, so the fixture cannot drift from the format the parser accepts.

## Verification

`--only aggregator` 8, `--only scan` 95, `detect-reposts.test.mjs` 254, and the SYSTEM_PATHS coverage validator over 1,161 tracked files. Rebased onto `main` after #2702 landed; the three commits below are only the aggregator work, with #2702's own commits now upstream.

*Scope: one personal `scan-history.tsv`, 2,470 rows at the time of the derivation measurement. The 17% and the 2% are observed. The fixture in the test is synthetic — a real corpus cannot exercise "the same rows with and without the flag".*

## Summary by CodeRabbit

* **New Features**
  * Added support for identifying job-board and portfolio aggregators.
  * Aggregator companies remain visible in reports, even without tracking data.
  * Added a clear `aggregator-not-evaluated` status and explanation.
  * Repost detection now excludes flagged aggregator companies.
  * Added optional aggregator configuration examples.

* **Documentation**
  * Documented the new aggregator status and configuration.

* **Tests**
  * Added coverage for aggregator detection, filtering, reporting, and fallback behavior.


artemtrofymenko:
Both verified and both fixed. The first one is the serious one, and it invalidated a claim in this PR's own description.

## `job_boards` — you are right, and the shipped default was inert

The two entries `templates/portals.example.yml` flags are at lines 2381 and 2393, both under `job_boards`. The loader read only `tracked_companies`. Measured against the template itself:

```
before   0 aggregators
after    2 — founderful <- "Founderful (portfolio)",  joinup ch <- "joinup.ch"
```

The PR description said *"portals.example.yml ships it set on the two multi-employer boards it already lists, so a default install is correct"*. **That was false as shipped** — the flag was visible in the config and did nothing. I have left the description as written and am correcting it here rather than editing it quietly.

My tests did not catch it because every fixture used `tracked_companies`. That is the same blind spot in miniature: a synthetic case cannot tell you the shipped default is wired wrong. So the assertion I added is against the **template itself**, not another fixture —

```js
const shipped = loadAggregatorCompanies(join(ROOT, 'templates/portals.example.yml'));
if (shipped.size >= 2) …
```

— which returns 0 under the old loader while every synthetic case still passes. There is a `job_boards` fixture too, but the template assertion is the one that would have caught this.

## Temp fixtures — fixed, and it found a stale label

`co-aggr-<pid>.yml` in the shared temp dir is guessable, and `writeFileSync` follows a symlink left there. All three fixtures now live in one `mkdtempSync` directory removed by a single `rmSync`.

Doing that surfaced something the finding did not mention: one assertion read *"a portals.yml with no tracked_companies yields no aggregators"*, and once `job_boards` is read that stopped being the reason it passes. It still passed — on the new true reason, which is that its entries carry no flag. A green assertion whose label no longer describes what it tests is a comment that will mislead the next reader, so it now says what it checks.

## Verification

Reverting the loader to a single collection turns it red in both places, with the mutation printing proof it applied: 3 failures in `tests/aggregator-companies.test.mjs` (`job_boards flag not honoured: size=0`, `the shipped template loads 0 aggregators — its flags are inert`, `raw names look normalized rather than as written: []`) and 2 in `--self-test`.

Green at head: `--only aggregator` 11, `--only scan` 95, `detect-reposts.test.mjs` 254, `--self-test` 40, SYSTEM_PATHS over 1,162 tracked files.

## The macOS run

Unrelated to this change and already filed as #3162 — the `generate-pdf` page-budget test rejecting a path that is inside the workspace. Windows, Ubuntu and the upgrade gate are green on the same commit, and this PR touches no PDF code.

santifer:
Reviewed and approved on substance, @artemtrofymenko: an aggregator's repeated titles are many employers, not one re-listing, so every cluster built from it is fabricated — and reporting the check as SKIPPED instead of silently green is exactly the house rule (a skip that says nothing is how a guard stops guarding). One process note, in your favor: you are at this repo's three-merges-per-author-per-day ceiling (#2885, #3050, #2702 all landed today), a pacing rule that protects review quality, not a comment on this PR. It merges tomorrow morning as the first in line.

artemtrofymenko:
Right, and worth stating plainly: `shipped.size >= 2` is a **count**, and a count is not the claim. I spent this PR arguing that a synthetic fixture cannot tell you the shipped default is wired wrong, then wrote the assertion that checks the shipped default with a number.

Measured against a template with the two required boards unflagged and two unrelated entries flagged instead:

```
size = 2
current assertion  size >= 2     -> true
current assertion  /[.A-Z]/ raw  -> true      ("Totally Unrelated A." has both)
raw values: ["Totally Unrelated A.","Unrelated B"]
```

Both pass while protecting nothing. Fixed in `1097eb0` — membership and exact stored value for `Founderful (portfolio)` and `joinup.ch` by name. Seen red under that same mutation, and the failure now names the boards instead of reporting a number:

```
❌ the shipped template no longer flags: ["Founderful (portfolio)","joinup.ch"] (loaded 2)
❌ raw names lost or rewritten: ["Founderful (portfolio)","joinup.ch"]
     not in ["Totally Unrelated A.","Unrelated B"]
```

@santifer — understood on the pacing, and no objection to it: a ceiling that protects review depth is worth more to me than a fourth merge today, particularly on a PR whose whole argument is that a check which says nothing is a check that stopped working. This push is a test-only tightening, so it should not disturb tomorrow's ordering.

`--only aggregator` 11, `--only scan` 95, `--self-test` 40.

Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is the new `tests/aggregator-companies.test.mjs`; `detect-reposts.mjs` and `company-history.mjs` are outside my area and I read them for context.

Both sides of the sentinel are covered, which is what makes this hold up. The same fixture is run with the flag off and on, and it asserts the cluster exists without the flag and is gone with it, so the flag is proven to be what removes the aggregator cluster rather than rows that never clustered. The negative side is there too: a normal single-employer company alongside the aggregator board still clusters with the flag on, and the label path reads `none-detected` for the non-aggregator case against `aggregator-not-evaluated` for the board. Hermetic (throwaway dir, in-memory scan history, `rmSync` cleanup in `finally`), reads the shipped `portals.example.yml` to pin the two boards by name, and the failure messages interpolate the observed counts.

santifer:
Kept as promised, @artemtrofymenko: first in line this morning. The part that makes this one better than a plain skip is that it says WHY a board was skipped in the output instead of silently dropping it, so a user auditing their scan never mistakes a policy decision for a bug. Merged! 🚀

Issue #3170, opened 2026-08-21
Every .mjs entrypoint's main-guard compares a realpathed import.meta.url against a lexical argv[1] — a symlinked checkout makes the CLI a silent no-op with exit 0


vliggio:
Found while fixing the test-harness half of #3165. It is not a test problem — `generate-pdf.mjs` invoked through a symlinked path prints nothing and **exits 0**, so every caller reads it as success.

## Reproduction

```bash
ln -s /path/to/career-ops /tmp/co

node /path/to/career-ops/generate-pdf.mjs --help   # prints usage
node /tmp/co/generate-pdf.mjs --help               # prints NOTHING, exit 0
```

## Cause

`generate-pdf.mjs:1662`:

```js
const isMain = process.argv[1] && fileURLToPath(import.meta.url) === resolve(process.argv[1]);
```

Node resolves the ESM entry through **realpath**; `process.argv[1]` keeps whatever spelling the caller typed. Through a symlink the two never match, `isMain` is false, the CLI tail is skipped, and the process falls off the end of the module — exit 0, no output, no PDF.

`upskill.mjs:843-848` already documents this exact mechanism, and consciously accepts it:

> invoking this file through a SYMLINK reads as "not main" and prints nothing […] Every caller (test-all.mjs, the modes, package scripts) uses the real path, and matching the repo convention is worth more here than covering a path nothing takes.

I think that premise no longer holds, and the reasoning is worth revisiting rather than the individual line:

- **The path does get taken.** Any checkout reached through a symlink hits it: `/opt/career-ops` linked into a home directory, a checkout on an external volume linked into `~`, a shared install symlinked per-user, or a `~/bin` shim. None of those are exotic, and none of them look like they involve symlinks from the user's side.
- **`tmpdir()` is a symlink on macOS** (`/var/folders` → `/private/var`), which is why `contacts.test.mjs:354-357` already realpaths its temp root to stop the copied script's main-guard from silently defeating itself. That is the same bug, worked around per-test rather than fixed.
- **It fails silently and successfully**, which is the worst available failure mode. A CLI that exits 1 with "not main" would be a puzzle; one that exits 0 with no output tells `apply`, `pdf` and the batch runner that a PDF was produced.

`reserve-report-num.mjs:353-359` already contains the correct implementation:

```js
function isDirectInvocation() {
  if (!process.argv[1]) return false;
  try {
[...]

Issue #3169, opened 2026-08-21
generate-pdf: a symlinked data/ relocates workspaceRoot outside the repo, silently disabling the #2533 section-order guard and #1837 theming


vliggio:
Split out of #3165, where this surfaced as two test failures. Filing separately because — unlike the other two causes in that issue — this one is not a test artifact: it silently disables two guards at runtime.

## What happens

`generate-pdf.mjs:45-47`:

```js
const trackerPath = resolveTrackerPath(__dirname);
const workspaceRoot = resolveWorkspaceRoot(trackerPath);
```

`resolveTrackerPath` ends in `canonicalizeTrackerPath`, which **realpaths** — deliberately, and correctly, because the lock key needs one canonical spelling per tracker. `resolveWorkspaceRoot` then derives the workspace root *lexically* from that already-canonicalized path:

```js
export function resolveWorkspaceRoot(trackerPath) {
  const trackerDir = dirname(trackerPath);
  return basename(trackerDir) === 'data' ? dirname(trackerDir) : trackerDir;
}
```

Each half is defensible alone. Composed, they walk through a symlinked `data/` and land in a directory that is not the career-ops workspace:

```
data -> ../job-data/data

trackerPath   = <external>/job-data/data/applications.md
workspaceRoot = <external>/job-data          # not the repo
```

Nothing checks that the derived root looks like a workspace, so the mistake is silent.

## Why it matters at runtime

`workspaceRoot` is where `generate-pdf.mjs` reads the user's profile and CV:

- `cv.md` — `:1149`
- `config/profile.yml` — `:1160` (`readCvSectionOrder`) and `:1180` (`readStyleTokens`)

Someone who symlinks only `data/` — the smallest and most natural workaround for #524, and the one people reach for first because `data/` is where the PII actually accumulates — has neither file at the derived root. `:1148-1152` swallows the ENOENT:

```js
let cvMarkdown = '';
try {
  cvMarkdown = await readFile(resolve(workspaceRoot, 'cv.md'), 'utf-8');
} catch (err) {
  if (err?.code !== 'ENOENT') throw err;
}
```

`cvMarkdown` stays `''`, so:

- `validateCvSectionOrder(html, cvMarkdown, …)` compares against an empty document and no-ops — the CV section-order guard from #2533 is **off**, with no warning.
- `readCvSectionOrder` / `readStyleTokens` fall back to defaults against a `config/profile.yml` that isn't there — #1837 dynamic theming silently stops applying.

[...]

Issue #524, opened 2026-04-28
Feature Request: Support for external data directory (separate user data from code)


andruwa13:
### Code of Conduct

- [x] I agree to follow the [Code of Conduct](https://github.com/santifer/career-ops/blob/main/CODE_OF_CONDUCT.md)

### Existing issues

- [x] I searched existing issues and this hasn't been requested yet

### Problem

Currently, all user data is stored inside the repository in the following directories:

data/
reports/
output/
jds/
interview-prep/

Even though these folders are gitignored, they still live inside the project folder. This creates several inconveniences:

It's hard to safely switch between branches or different forks
Testing experimental changes in the code becomes risky
Difficult to maintain multiple separate "databases" (e.g. for different countries or job search strategies)
Mixing code and personal data is not ideal for long-term usage

### Proposed solution

Introduce support for an external data directory via an environment variable.
Main variable:
Bashexport CAREER_OPS_DATA_DIR=~/career-ops-data
# or
export CAREER_OPS_DATA_DIR=~/.local/share/career-ops
If CAREER_OPS_DATA_DIR is not set, the tool should fall back to the current behavior (./data, ./reports, ./output, etc.) for backward compatibility.
Benefits

Clear separation between code (repository) and user data
Safe to git switch, git pull, or work with multiple forks
Easy to have several independent job search profiles
Follows best practices for CLI tools
Makes testing and development much cleaner

What needs to be done

Add support for CAREER_OPS_DATA_DIR in all JavaScript/TypeScript .mjs scripts
Support the same variable in the Go dashboard
Add optional --data-dir CLI flag
Update doctor.mjs to check and create folders in the external directory
Update documentation (README.md, CLAUDE.md, DATA_CONTRACT.md)
(Optional) Add a config file or .env support for the data path

Alternative variable names (open for discussion)

CAREER_OPS_DATA_DIR ← my preference
CAREER_DATA_DIR
CAREER_OPS_HOME

Would love to hear the maintainer's thoughts on the best name and implementation approach.

### Area

Other

santifer:
Sorry for the slow reply here, @andruwa13 — and good request, the pain is real (switching branches/forks without your data getting in the way).

**You can actually do this today, no code change**, because the updater is symlink-safe. Move your user data out and symlink it back:

```bash
mv data ~/career-ops-data/data && ln -s ~/career-ops-data/data data
# repeat for reports/ output/ jds/ interview-prep/ as you like
```

[...]

PR #3172, opened 2026-08-21
fix(tests): make the suite symlink-aware for externalized user-layer dirs

### LINKED ISSUE #3165: test-all.mjs: 23 failures on a checkout with symlinked user-layer dirs (the #524 workaround) — two test-harness causes
`test-all.mjs` reports **25 failures / 5051 passes** at v1.28.0 on a checkout where the user-layer directories are symlinked out of the repo — the layout people adopt today as the manual workaround for #524 (external data directory).

None of the 25 is a defect in the behaviour being asserted. Three distinct causes; **this issue covers the two that are test-harness fragility (23 failures)**. The remaining 2 are a real `generate-pdf.mjs` bug that a symlinked `data/` triggers silently, split out to #3169 because it needs an architecture call rather than a test fix.

Because these failures are indistinguishable from real ones in the output, they mask genuine regressions for anyone on this layout, and `test-all.mjs` is the gate the contributing guide points at before every PR.

## Layout

```
career-ops/
  data          -> ../job-data/data
  jds           -> ../job-data/jds
  output        -> ../job-data/output
  reports       -> ../job-data/reports
  interview-prep -> ../job-data/interview-prep
```

`node -v` v26.5.0, `git --version` 2.55.0, macOS. Every path below is generalized; nothing here depends on where the external dir lives.

Minimal reproduction from a clean checkout:

```bash
mkdir -p ../job-data
mv data ../job-data/data && ln -s ../job-data/data data
mv output ../job-data/output && ln -s ../job-data/output output
mv reports ../job-data/reports && ln -s ../job-data/reports reports
node test-all.mjs
```

---

## Cause 1 — PDF sandboxes are built under `output/`, so `node_modules` resolution escapes the repo (20 failures)

Five sandboxes are created with `mkdtempSync` anchored at `ROOT/output`:

- `tests/generate-pdf-page-budget.test.mjs:17` and `:18`
- `tests/generate-pdf-batch.test.mjs:31`
- `tests/cv-section-order.test.mjs:736` and `:856`

Each copies `generate-pdf.mjs` plus `theme-style.mjs` into the sandbox and stubs `node_modules/playwright` inside it. `theme-style.mjs` also does `import * as yaml from 'js-yaml'` (`theme-style.mjs:29`), and *that* one is expected to resolve by walking up from the sandbox into the repo's own `node_modules`.

Node resolves ESM specifiers against the importer's **realpath** (`--preserve-symlinks` is off by default). When `output/` is a symlink, the walk-up starts outside the repo and never reaches `ROOT/node_modules`:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'js-yaml' imported from
  <external>/job-data/output/page-budget-test-XXXXXX/theme-style.mjs
    at Object.getPackageJSONURL (node:internal/modules/package_json_reader:301:9)
```

Every spawned `generate-pdf.mjs` dies before parsing argv, so all 20 assertions in those three files report as behaviour regressions — `generate-pdf did not reject --max-pages=nope before rendering`, `single vs batch render diverged`, `containment guard regressed`, `generate-pdf followed an output symlink outside its workspace`, and so on.

Note the sandbox location is load-bearing in a way that isn't obvious: moving these to `tmpdir()` does **not** fix it — `tmpdir()` is outside the repo too, and they only work today because `ROOT/output` happens to be physically inside the repo.

**Suggested fix — two parts, both needed.** The missing package is only the first thing that goes wrong; fixing it alone moves the failures rather than removing them.

1. **Make the sandbox self-sufficient.** It already stubs `playwright` in `sandbox/node_modules`; link the real `js-yaml` in beside it, and the sandbox no longer cares where it physically sits. (`'junction'` as the link type: ignored on POSIX, and the one link type Windows grants without elevation — cf. #2828.)

2. **Realpath the sandbox root**, i.e. `realpathSync(mkdtempSync(...))`. With only part 1 applied, the spawned `generate-pdf.mjs` loads fine and then does *nothing*: its main-guard at `generate-pdf.mjs:1662` compares a realpathed `import.meta.url` against a lexical `process.argv[1]`, those disagree under a symlinked `output/`, `isMain` is false, and the process exits 0 with no output. Every assertion then fails against an empty string instead of an ERR_MODULE_NOT_FOUND — same red suite, less legible cause.

Part 2 is a workaround for a real product bug, now filed separately as #3170 (the guard affects ~40 entrypoints, and a symlinked checkout makes any of them a silent no-op). The realpath is worth keeping in the tests either way: it makes them agree with themselves regardless of how #3170 is resolved.

---

## Cause 2 — `git check-ignore` fails hard beyond a symlink, and the bare `catch` scores that as "not ignored" (3 failures)

`tests/user-layer-gitignored.test.mjs:34-46` probes each declared user-layer path:

```js
try {
  execFileSync('git', ['check-ignore', '-q', '--no-index', probe], { cwd: ROOT });
} catch {
  ignored = false;
}
```

Git refuses to evaluate *any* pathspec that traverses a symlink, and exits **128**, not 1:

```
$ git check-ignore --no-index data/__gitignore_probe__.md
fatal: pathspec 'data/__gitignore_probe__.md' is beyond a symbolic link
$ echo $?
128
```

`--no-index` does not avoid this. The catch-all conflates "git answered no" (exit 1) with "git could not answer" (exit 128), so `data/`, `output/` and `reports/` are reported as:

> `data/ is declared user-layer in AGENTS.md but is NOT git-ignored — personal data could be committed`

which is a false negative on a security-flavoured assertion — the most expensive kind to get wrong, in both directions.

Minimal repro, no career-ops involved:

```bash
mkdir -p /tmp/gi/repo /tmp/gi/ext/data && cd /tmp/gi/repo && git init -q .
printf '/data\n' > .gitignore && ln -s ../ext/data data
git check-ignore --no-index -v data/x.md   # fatal: ... is beyond a symbolic link; exit 128
git check-ignore --no-index -v data        # .gitignore:1:/data  data; exit 0
```

**Suggested fix:** two parts.

1. Separate the exit codes. `status === 1` means not ignored; anything else is an error the test should surface as such rather than silently convert into a failing assertion.
2. When the declared path is a symlink, probe **the link entry itself** rather than a path through it — that is the question that actually matters, and git answers it fine (exit 0 above).

Part 2 is not merely cosmetic, and it turns up something real. Stock `.gitignore` ignores `data/*`, which does not match `data` itself, so on this layout the symlink entry is genuinely stageable:

```bash
$ git status --porcelain
?? data
```

Staging it commits the link target string, not the user's tracker — a smaller leak than the current message claims, but a leak of a local filesystem path all the same, and one the current probe cannot see. A rule that covers both shapes (`/data` alongside `data/*`) would close it; that may belong with #2277, which is already about ignore rules not reaching existing installs.

---

## Failure attribution

| Cause | Failures | Test-only? |
|---|---|---|
| 1 — sandbox under `output/` | 20 | yes |
| 2 — `check-ignore` beyond symlink | 3 | yes (but see the real `?? data` finding) |
| 3 — `workspaceRoot` escapes the repo | 2 | **no** — split to #3169 |

With causes 1 and 2 fixed the suite goes 25 → 2 on this layout, the remaining 2 being #3169. The symlink layout itself is deliberate and stays as it is.

PR incoming for both; #3169 is waiting on your call about which fix shape you want.

vliggio:
Closes #3165.

On a checkout whose user-layer directories are symlinked out of the repo — the layout people adopt as the manual workaround for #524 — `test-all.mjs` reported **25 failures**, 23 of them harness artifacts rather than defects in the behaviour under assertion. They are indistinguishable from real failures in the output, so they mask genuine regressions for everyone on that layout.

**Verified both ways:**

| Layout | Before | After |
|---|---|---|
| symlinked user layer | 25 failed / 5050 passed | **2 failed / 5073 passed** |
| stock checkout (this branch, on current `main`) | — | the four affected suites all green |

The 2 that remain are #3169, a runtime bug in `generate-pdf.mjs` rather than a test artifact. Untouched here, and awaiting your call on its fix shape.

---

## 1. PDF sandboxes could not resolve `js-yaml` (20 failures)

The sandboxes are created under `ROOT/output` and run a copy of `generate-pdf.mjs`, whose siblings `theme-style.mjs` and `tracker-utils.mjs` both `import * as yaml from 'js-yaml'`. That specifier resolves by walking up into `ROOT/node_modules`, and the walk starts from the importer's **realpath**, so a symlinked `output/` puts it out of reach. Every spawned script died with `ERR_MODULE_NOT_FOUND` before parsing an argument, failing 20 assertions that never ran.

`tests/helpers.mjs` gains `linkRepoPackage()`, which links the package into the sandbox's own `node_modules` beside the `playwright` stub each sandbox already writes there. The sandbox stops depending on where it physically sits.

Worth noting, since it was my first instinct too: **relocating these to `tmpdir()` would not have worked** — tmpdir is outside the repo as well. They pass today only because `ROOT/output` happens to be physically inside it, and that assumption is the thing worth removing.

The sandbox roots are also realpathed. That is required on top of the above, not belt-and-braces: with only the package linked, the spawned `generate-pdf.mjs` imports cleanly and then **exits 0 having done nothing**, because its main-guard compares a realpathed `import.meta.url` against a lexical `argv[1]`. I filed that separately as #3170 — it affects ~40 entrypoints and deserves its own decision. Realpathing the sandbox root makes these tests agree with themselves whichever way #3170 goes, so this PR does not depend on it.

## 2. `git check-ignore` beyond a symlink was scored as "not ignored" (3 failures)

`tests/user-layer-gitignored.test.mjs` probed each declared user-layer path inside a bare `try`/`catch`. Git refuses any pathspec that crosses a symlink and exits **128, not 1**, so the catch scored "git could not answer" as "not ignored" — a false negative on the one assertion in the file whose entire purpose is to be trustworthy about PII. `--no-index` does not avoid it.

```
$ git check-ignore --no-index data/__gitignore_probe__.md
fatal: pathspec 'data/__gitignore_probe__.md' is beyond a symbolic link
$ echo $?
128
```

`checkIgnore()` now separates exit 0 / exit 1 / anything else. When the declared path is itself a symlink it probes the link entry, which is the question that actually governs what `git add .` stages once the contents live outside the repository. All three probe loops in the file shared the same bare catch, so all three are converted.

**One judgment call worth your attention.** When only the *link entry* is unignored, this `warn()`s rather than `fail()`s. Stock `data/*` does not match `data`, so a symlinked layout genuinely leaves `?? data` stageable — but staging a symlink commits its target path, not the user's CV or tracker, and failing there would hand exactly the affected users a permanently red suite over a much smaller leak. The warning names the fix (`/data` alongside `data/*`). If you would rather it be a hard failure, or would rather close the gap in `.gitignore` instead (possibly alongside #2277), both are small changes and I am happy to make either.

---

No production code is touched; this is `tests/` only.

---

## Summary by CodeRabbit

* **Tests**
  * Improved test reliability when running in temporary and symlinked directories.
  * Enhanced dependency setup for isolated test environments.
  * Added fallback handling for environments that do not support directory junctions.
  * Improved Git ignore checks, including clearer handling of symlinks and pathspec errors.

### DIFF
--- a/tests/cv-section-order.test.mjs
+++ b/tests/cv-section-order.test.mjs
@@ -7,11 +7,11 @@
 // Assertions run against rendered HTML rather than source patterns: the reorder
 // has to survive nested markup, comments, absent optional sections and a second
 // application, and none of that is observable from the source text.
-import { pass, fail, ROOT, NODE } from './helpers.mjs';
+import { pass, fail, linkRepoPackage, ROOT, NODE } from './helpers.mjs';
 import { spawnSync } from 'child_process';
 import { join } from 'path';
 import { pathToFileURL } from 'url';
-import { copyFileSync, existsSync, mkdirSync, readFileSync, writeFileSync, mkdtempSync, rmSync } from 'fs';
+import { copyFileSync, existsSync, mkdirSync, readFileSync, realpathSync, writeFileSync, mkdtempSync, rmSync } from 'fs';
 import { tmpdir } from 'os';
 
 console.log('\nCV section order from config/profile.yml (#2533)');
@@ -733,7 +733,10 @@ try {
 {
   const outputRoot = join(ROOT, 'output');
   mkdirSync(outputRoot, { recursive: true });
-  const sandbox = mkdtempSync(join(outputRoot, 'section-order-batch-'));
+  // realpathSync: see tests/generate-pdf-page-budget.test.mjs -- argv[1] keeps the
+  // caller's spelling while import.meta.url is realpathed, so a symlinked
+  // output/ makes generate-pdf.mjs's isMain guard false and the spawn a no-op (#3165).
+  const sandbox = realpathSync(mkdtempSync(join(outputRoot, 'section-order-batch-')));
   try {
     const script = join(sandbox, 'generate-pdf.mjs');
     for (const f of [
@@ -742,6 +745,12 @@ try {
     ]) {
       copyFileSync(join(ROOT, f), join(sandbox, f));
     }
+
+    // theme-style.mjs and tracker-utils.mjs both `import * as yaml from
+    // 'js-yaml'`, resolved by walking up into the repo's node_modules -- from
+    // the sandbox's REALPATH, so a checkout with a symlinked output/ never
+    // reaches it and the spawned generate-pdf dies before parsing argv (#3165).
+    linkRepoPackage(sandbox, 'js-yaml');
     mkdirSync(join(sandbox, 'data'), { recursive: true });
     writeFileSync(join(sandbox, 'data', 'pdf-index.tsv'), '', 'utf-8');
 
@@ -853,7 +862,10 @@ export const chromium = {
 {
   const outputRoot = join(ROOT, 'output');
   mkdirSync(outputRoot, { recursive: true });
-  const sandbox = mkdtempSync(join(outputRoot, 'section-order-anchor-'));
+  // realpathSync: see tests/generate-pdf-page-budget.test.mjs -- argv[1] keeps the
+  // caller's spelling while import.meta.url is realpathed, so a symlinked
+  // output/ makes generate-pdf.mjs's isMain guard false and the spawn a no-op (#3165).
+  const sandbox = realpathSync(mkdtempSync(join(outputRoot, 'section-order-anchor-')));
   try {
     const script = join(sandbox, 'generate-pdf.mjs');
     for (const f of [
@@ -862,6 +874,12 @@ export const chromium = {
     ]) {
       copyFileSync(join(ROOT, f), join(sandbox, f));
     }
+
+    // theme-style.mjs and tracker-utils.mjs both `import * as yaml from
+    // 'js-yaml'`, resolved by walking up into the repo's node_modules -- from
+    // the sandbox's REALPATH, so a checkout with a symlinked output/ never
+    // reaches it and the spawned generate-pdf dies before parsing argv (#3165).
+    linkRepoPackage(sandbox, 'js-yaml');
 
     // The external workspace: tracker, profile, CV and documents all live here,
     // and NOT beside the script. This is the shape CAREER_OPS_TRACKER creates.
--- a/tests/generate-pdf-batch.test.mjs
+++ b/tests/generate-pdf-batch.test.mjs
@@ -18,17 +18,24 @@ import {
   existsSync,
   mkdirSync,
   mkdtempSync,
+  realpathSync,
   readFileSync,
   rmSync,
   writeFileSync,
 } from 'fs';
 import { join } from 'path';
 import { tmpdir } from 'os';
-import { pass, fail, ROOT, NODE } from './helpers.mjs';
+import { pass, fail, linkRepoPackage, ROOT, NODE } from './helpers.mjs';
 
 const outputRoot = join(ROOT, 'output');
 mkdirSync(outputRoot, { recursive: true });
-const sandbox = mkdtempSync(join(outputRoot, 'batch-test-'));
+// realpathSync, not the raw mkdtemp path: Node resolves both `import.meta.url`
+// and a module's node_modules walk from a file's REALPATH, while
+// `process.argv[1]` keeps whatever spelling the caller used. On a checkout with
+// a symlinked output/ the two disagree, so generate-pdf.mjs's `isMain` guard is
+// false and the spawned script exits 0 having done nothing at all -- assertions
+// then fail against empty output rather than against behaviour (#3165).
+const sandbox = realpathSync(mkdtempSync(join(outputRoot, 'batch-test-')));
 const script = join(sandbox, 'generate-pdf.mjs');
 const launchesFile = join(sandbox, '.launches');
 const pageClosesFile = join(sandbox, '.pagecloses');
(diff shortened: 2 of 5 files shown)


artemtrofymenko:
Reproduced your fix independently on Windows before reading it closely, because the interesting part of this one is portability and my box happens to be a hostile environment for it.

## The fix works, and on a link type you may not have been able to test

This machine refuses ordinary symlinks:

```
symlink dir   -> EPERM
symlink file  -> EPERM
junction      -> ok
```

So `symlinkSync(source, dest, 'junction')` is not a nicety here, it is the only thing that works. Recreating your layout with a junctioned `output/` and running the page-budget suite on the merge-base versus this branch:

```
base      exit=1   0 passed, 10 failed   ERR_MODULE_NOT_FOUND x27
#3172     exit=0  10 passed,  0 failed   ERR_MODULE_NOT_FOUND x0
```

And end to end, which is the thing that actually matters: a script spawned inside the sandbox resolves the specifier.

```
linkRepoPackage -> <sandbox>\node_modules\js-yaml
package.json readable through it : true
yaml.load works from the sandbox : true
```

Your note that relocating to `tmpdir()` would not have helped is right for the same reason it is right here — the resolution walk starts at the importer's realpath, and tmpdir is no more inside the repo than an externalised `output/` is.

## One regression, in a layout that works today

`linkRepoPackage` throws when `ROOT/node_modules/<pkg>` is absent. Node's own resolution does not require that — it walks **up** — so a `git worktree` checkout has no `node_modules` of its own and resolves everything from the parent repo. That works on `main` and stops working here:

```
                             base                 #3172
--only generate-pdf-page-budget    10 passed, 0 failed   0 passed, 1 failed
--only cv-section-order            47 passed, 0 failed  45 passed, 2 failed
```

Both failures are `linkRepoPackage: js-yaml is not installed at <worktree>/node_modules -- run npm install`, and `npm install` is not the answer: the package is installed, one directory up, exactly where node looks for it.

Worth flagging because of what this PR is: the point is to stop the sandbox depending on where it physically sits, and `existsSync(join(ROOT, 'node_modules', pkgName))` is a new assumption about where the *repo* physically sits.

Asking node instead of the filesystem covers both layouts. Measured from the worktree where the `existsSync` check fails:

```
existsSync(ROOT/node_modules/js-yaml)     false
createRequire(...).resolve('js-yaml')  -> C:\...\career-ops\node_modules\js-yaml   (exists)
```

Something like resolving the specifier and walking up to its `package.json` directory, with the current path as the fast path. I have not written it as a diff — the shape is yours to pick, and the `cpSync` fallback and the `existsSync(dest)` short-circuit both stay as they are.

## Scope

Windows 11, Node 22, one machine. The before/after and the worktree pair are observed, not reasoned. I could not test the full externalised user layer — only `output/` — because this box cannot create the ordinary symlinks that layout uses, which is also why the junction detail above is worth having from here. The two remaining failures you mention (#3169) I did not look at.

santifer:
This is bigger than its title, @vliggio: besides un-masking 23 harness artifacts for everyone on the symlinked-user-layer workaround (closing #3165), your realpathSync-the-sandbox-first fix hardens exactly the failure class behind #3162, the intermittent macOS CI hit where a canonicalized path and its spelled form disagree mid-test. If the occurrences stop after this lands, you found the cause; either way the test now stands on canonical ground. And @artemtrofymenko reproducing your fix independently on Windows before reading it is the review culture this repo hopes for. Merged! 🚀

Issue #3175, opened 2026-08-21
Populate job.description on Greenhouse boards so visa/content/country filters stop passing blind


BigBrown10:
### Code of Conduct

- [x] I agree to follow the [Code of Conduct](https://github.com/santifer/career-ops/blob/main/CODE_OF_CONDUCT.md)

### Existing issues

- [x] I searched existing issues and this hasn't been requested yet

### Problem

Three scan-time filters read `job.description`: `content_filter`, `country_eligibility`, and `visa_filter` (the sponsorship screen from #2057). All three degrade to pass-through on any provider whose list response carries no description, which is currently most hosted-ATS boards.

The biggest gap is Greenhouse: `providers/greenhouse.mjs` calls `/v1/boards/{slug}/jobs` with no query params, so a user who sets `visa_filter: { enabled: true }` gets zero signal from every Greenhouse board, even for postings whose text explicitly says "we are unable to sponsor". Same blind spot applies to the content and country filters.

For reference, providers already populating `job.description` at list time: lever, agentic-jobs, gem, workable, vdab, oraclecloud, alibaba, meituan, tencent, cryptocurrencyjobs, remotli. The remaining hole is concentrated in the large hosted ATS platforms.

### Proposed solution

Phased, one provider family at a time:

**Phase 1 — Greenhouse (single request, one param).** The public boards API accepts `content=true`, which embeds each posting's full content in the same list response. Verified live against a public board: the `content` field is present with the param and absent without it. Treatment: decode the HTML entities, strip tags to plain text (same approach as agentic-jobs' `stripHtml`), set `job.description`.

Cost disclosure, measured: `content=true` grew that board's response from ~57 KB to ~829 KB across 83 jobs (~10 KB/posting of raw HTML). Mitigation options, in my personal order of preference but very open to pushback:

1. Cap the stripped plain-text length, the way alibaba.mjs already caps its unusually long JDs.
2. Opt-in per `job_boards:` entry (e.g. an `include_descriptions: true` flag), default off.
3. Auto-enable only when a description-consuming filter (`visa_filter`, `content_filter`, country eligibility) is actually configured.

If the maintainer prefers option 2 or 3, say the word and the Phase 1 PR will ship that way.

[...]

PR #3177, opened 2026-08-21
fix(greenhouse): populate job.description via content=true (#3175)

BigBrown10:
## What does this PR do?

Greenhouse boards now ship each posting's full body as plain-text `job.description`, fetched in the same single list request via the boards API's `content=true` param. Until now `providers/greenhouse.mjs` requested titles only, so `visa_filter`, `content_filter`, and the country-eligibility filter all passed every Greenhouse board blind, including postings that explicitly refuse sponsorship.

## Related issue

Part of #3175 (Phase 1: Greenhouse; the detail-fetch tier for Recruitee/SmartRecruiters/Ashby is Phase 2, discussed there).

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] If this is a new feature or architecture change, I opened an issue first (bug fixes, providers, docs & translations are exempt — send those straight in) — issue #3175 was opened anyway for the multi-phase coordination
- [x] I ran `node test-all.mjs` and all tests pass — 5100 passed; 1 contained failure is `tests/plugin-symlink-discovery.test.mjs` dying on `symlinkSync` EPERM, a Windows symlink-privilege limitation of this machine (no Developer Mode), unrelated to this change and pre-existing here. The greenhouse suite (40 assertions) and the shared-decoder guard suite both pass.

## Implementation notes

- **Double-decoded HTML:** with `content=true`, bodies arrive entity-escaped (`&lt;p&gt;`), so one pass reveals tags and a second pass after stripping resolves text-level entities (`&amp;`, `&#39;`). Verified against the live API.
- **Payload cost, measured:** `content=true` grew a public board's response from ~57 KB to ~829 KB across 83 jobs (~10 KB/posting). Mitigated the same way alibaba caps its full-text JDs: stripped text is sliced at 4000 chars. Happy to switch to opt-in per board entry if the maintainer prefers that tradeoff — one-line change at the `searchParams.set` call site.
- **Shape discipline:** the `description` key is omitted entirely when a posting has no body (same convention as cryptocurrencyjobs/remotli), so downstream "no signal" stays distinguishable from empty string. The raw `content` field never leaks into the normalized job.
- **Guard chain:** the final URL (with params) is re-validated through `assertGreenhouseUrl` before the request; `searchParams.set` is idempotent so an `entry.api` already pinning `content=true` can't be duplicated.
- Live smoke test against a public 83-job board: 83/83 jobs carry clean text descriptions, cap enforced, no HTML residue.

## How to test

1. `node tests/providers/greenhouse.test.mjs` — 40 pass, including the new double-decode, omission, idempotent-param, and 4000-char-cap cases
2. `node tests/providers/rss-entity-decoding.test.mjs` — confirms no private decoder re-entered alongside the shared one
3. Optional live check: any Greenhouse board now yields descriptions where it previously yielded none

## Summary by CodeRabbit

* **New Features**
  * Job listings now include an optional plain-text description when posting content is available.
  * Descriptions are cleaned for readability, with scripts and styles removed and content limited to 4,000 characters.
  * Greenhouse content is fetched more reliably, including support for existing content parameters.

* **Bug Fixes**
  * Prevented duplicate content parameters in Greenhouse requests.
  * Excluded empty descriptions and raw HTML from normalized job listings.

### DIFF
--- a/providers/greenhouse.mjs
+++ b/providers/greenhouse.mjs
@@ -3,6 +3,12 @@
 
 // Greenhouse provider — hits the public boards-api JSON endpoint.
 // Handles both explicit `api:` URLs and auto-detection from `careers_url`.
+// Requests the board with `content=true` (#3175) so each posting carries its
+// full body as plain-text `description`; scan.mjs's content_filter,
+// country_eligibility filter and visa_filter all read that field, and without
+// it every Greenhouse board passed those filters blind.
+
+import { decodeEntities } from './_html-entities.mjs';
 
 const ALLOWED_GREENHOUSE_HOSTS = new Set([
   'boards-api.greenhouse.io',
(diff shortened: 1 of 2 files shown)


santifer:
Another solid one, @BigBrown10. Re-validating the FINAL URL with assertGreenhouseUrl after adding the param keeps the anti-SSRF chain on the exact string that goes over the wire, and reusing the shared _html-entities.mjs decoder instead of introducing a fifth private copy is exactly the discipline rss-entity-decoding.test.mjs exists to protect. Your opt-in offer for the payload is noted: if a bandwidth complaint ever lands, we take the one-line switch. Merged! 🚀

PR #3180, opened 2026-08-21
fix(liveness): add a LinkedIn rung so LinkedIn postings get a real verdict

### LINKED ISSUE #3179: bug(liveness): no LinkedIn rung, so a LinkedIn posting can only be judged from a redirect that yields a false expired
### Code of Conduct

### Existing issues

### What happened?

`liveness-api.mjs` carries API rungs for Greenhouse, Lever, Ashby and Workday. It has **no LinkedIn rung** — the string does not appear in the file.

So a LinkedIn posting can only be adjudicated by the Playwright rung. But `linkedin.com/jobs/view/{id}` fetched without a session redirects to a generic job search, and that page has neither an expired marker nor the original posting on it. Reading it produces a **false `expired`** for a posting that is still open. Avoiding that means leaving LinkedIn URLs `uncertain`, which is what happens in practice, so LinkedIn postings effectively never get a verdict.

That matters more than it looks, because `liveness-api.mjs` states the tradeoff itself:

> CONSERVATIVE BY DESIGN: a false "expired" is worse than the status quo (the user misses a real job).

A rung that cannot answer is safe but useless; a browser rung pointed at a redirect is actively wrong. Neither is what a user wants for the source many of their tracked roles come from.

### Steps to reproduce

Against `main`:

```bash
grep -c linkedin liveness-api.mjs        # -> 0
grep -coE 'greenhouse|lever|ashby|workday' liveness-api.mjs   # -> 25, so the grep is sound
```

Then check a known-closed LinkedIn posting through `check-liveness.mjs`. It cannot reach an API rung, falls to the browser path, and the page it lands on is the generic search rather than the posting.

### Expected behavior

A LinkedIn rung using the public guest endpoint, which needs no auth and no browser:

`https://www.linkedin.com/jobs-guest/jobs/api/jobPosting/{id}`

Two signals, which must **both** be consulted:

- **expired** iff the body contains `No longer accepting applications` (element `<figcaption class="closed-job__flavor--closed">`)
- **live** iff that marker is absent **and** an apply control is present (`public_jobs_apply-link-onsite` or `job-details-topcard-apply-modal`)
- anything else -> `uncertain`

Requiring both is the point. A marker check alone cannot distinguish "open" from "page did not render"; an apply check alone cannot distinguish "closed" from "apply control moved". Either one on its own can emit a confident wrong answer, and per the module's own invariant the wrong answer to avoid is a false `expired`.

Verified in both directions on six real postings, zero signal overlap: three live (apply control present, marker absent), three closed (marker present, apply control absent). A nonexistent id returns 404 with an empty body, so the existing 404-to-expired rule stays correct.

The id needs parsing from three shapes: `/jobs/view/{id}`, the slugged `/jobs/view/{title-slug}-{id}`, and `?currentJobId=` on search and collections pages.

### Related

#2995 and #2999 (@EliRobinson) normalize LinkedIn URLs to this same guest endpoint for the **evaluate** path, so the endpoint is already accepted as the legitimate mirror. This is the liveness half of the same idea and touches a different file; the two should not collide.

### CLI tool

Claude Code

### OS and Node.js version

macOS 26.5 / Node v22.22.3

### Output of npm run doctor

Not included: `doctor` reports clean here and its output adds nothing diagnostic. The gap is structural and visible from the two greps above.

rubicon:
## What does this PR do?

Adds a LinkedIn rung to `liveness-api.mjs` using the public guest endpoint, so a LinkedIn posting gets a real verdict instead of being left `uncertain` or judged from the redirect the browser rung lands on.

## Related issue

Closes #3179

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

---

### The rule, and why both signals are required

`https://www.linkedin.com/jobs-guest/jobs/api/jobPosting/{id}`, no auth and no browser.

- **expired** iff the body carries `No longer accepting applications` (`<figcaption class="closed-job__flavor--closed">`)
- **live** iff that marker is absent **and** an apply control is present (`public_jobs_apply-link-onsite` or `job-details-topcard-apply-modal`)
- anything else -> `uncertain`

Either signal alone can emit a confident wrong answer. A marker check on its own cannot tell "open" from "page did not render"; an apply check on its own cannot tell "closed" from "apply control moved". This follows the module's own stated invariant that a false `expired` is the worse failure, so ambiguity resolves to `uncertain` rather than to a guess.

Ids are parsed from `/jobs/view/{id}`, the slugged `/jobs/view/{title-slug}-{id}`, and `?currentJobId=` on search and collections pages.

### Two small pieces of provider plumbing

`throttleMs` (LinkedIn: 3500ms) and `accept` (LinkedIn: `text/html`), both threaded through `resolveAtsApi` and `checkLivenessViaApi`. The throttle is awaited **before** the timeout is armed, so waiting out the interval cannot consume the request's own budget. `api404Authoritative` is untouched and still flows through; LinkedIn does not set it and keeps the default behavior.

The `accept` override is semantic correctness rather than load-bearing: the endpoint serves the same HTML under the project's existing `DEFAULT_USER_AGENT`. I am not claiming it as a fix.

### Test

`tests/liveness-api-linkedin.test.mjs`, 36 assertions, auto-discovered under `tests/`, no `test-all.mjs` section and no `update-system.mjs` registration needed. Fixtures are synthetic.

Mutation-checked in three directions, and the third is the one that matters:

- **closed-marker signal forced off** reddens 5 named assertions, incl. `the closed marker on its own never reads as live`
- **apply-control signal forced off** reddens 6, incl. `checkLivenessViaApi returns the live verdict for a LinkedIn posting`
- **classifier forced to always answer `expired`** — the harmful checker — reddens 9, including the negative control `neither signal present is uncertain, not expired`

That last one is deliberate. A rung that can only say "expired" would pass a suite full of expired fixtures while being the worst possible outcome, so the suite is built so it cannot.

Also pinned: a lookalike host does not resolve to the rung (loosening the host guard to a substring match reddens two assertions), and a non-numeric job id does not resolve either.

`node test-all.mjs`: 5171 passed, 0 failed, 2 warnings (both pre-existing and unrelated).

### Docs

`docs/SCRIPTS.md` and `check-liveness.mjs`'s header both listed the rungs as Greenhouse/Lever and were already stale before this change; corrected to match.

## Summary by CodeRabbit

* **New Features**
  * Added LinkedIn posting checks using guest API responses.
  * Detects active, expired, and uncertain postings using multiple signals.
  * Supports regional LinkedIn hosts, request pacing, and appropriate response handling.
* **Bug Fixes**
  * Improved liveness checks with a generic public API check before browser-based fallback.
* **Documentation**
  * Updated liveness documentation to describe LinkedIn checks, outcomes, and request pacing.

### DIFF
--- a/check-liveness.mjs
+++ b/check-liveness.mjs
@@ -5,8 +5,8 @@
  *
  * Tests whether job posting URLs are still active or have expired.
  * Uses the same detection logic as scan.md step 7.5.
- * Zero Claude API tokens. Two rungs: a free ATS API check first
- * (Greenhouse/Lever — no browser), then Playwright for everything else.
+ * Zero Claude API tokens. Two rungs: a free public-API check first
+ * (liveness-api.mjs, no browser), then Playwright for everything else.
  *
  * Usage:
  *   node check-liveness.mjs <url1> [url2] ...
--- a/docs/SCRIPTS.md
+++ b/docs/SCRIPTS.md
@@ -526,7 +526,9 @@ npm run rollback
 
 ## liveness
 
-Tests whether job posting URLs are still live. Two rungs: a zero-token ATS API check first (`liveness-api.mjs` — Greenhouse, Lever, Ashby, Workday), falling back to headless Chromium (`liveness-browser.mjs`) for non-ATS pages or when the API is inconclusive. The browser rung detects expired patterns (e.g. "job no longer available"), HTTP 404/410, ATS redirect patterns, and apply-button presence, and supports multi-language expired patterns (English, German, French).
+Tests whether job posting URLs are still live. Two rungs: a zero-token API check first (`liveness-api.mjs` — Greenhouse, Lever, Ashby, Workday, LinkedIn), falling back to headless Chromium (`liveness-browser.mjs`) for everything else or when the API is inconclusive. The browser rung detects expired patterns (e.g. "job no longer available"), HTTP 404/410, ATS redirect patterns, and apply-button presence, and supports multi-language expired patterns (English, German, French).
+
+The LinkedIn rung reads the guest posting endpoint, which returns the rendered posting as HTML and answers HTTP 200 for closed postings as well as live ones. Liveness therefore comes from two independent signals in the body — the "No longer accepting applications" banner and the apply control — and the rung only concludes when they agree: banner without apply control is expired, apply control without banner is live, a body carrying both or neither is `uncertain`. That `uncertain` is final rather than a fall-through, because a headless fetch of `linkedin.com/jobs/view/{id}` lands on a generic search page rather than the posting, so the browser rung has nothing better to offer. The endpoint is unauthenticated and rate-limited, so the rung spaces its own requests.
 
 Per-job ATS endpoints (Greenhouse, Lever, Workday) treat a 200 as proof the posting is live; Ashby's public API is org-level (the whole job board), so that rung parses the board and confirms the specific job id is still listed. A definitive 404/410 from any ATS API is authoritative and short-circuits the browser check entirely — zero tokens, no browser launch.
 
--- a/liveness-api.mjs
+++ b/liveness-api.mjs
@@ -13,7 +13,7 @@
  * anything ambiguous (unknown ATS, redirect, 429/5xx, network/timeout) we return
  * `null` (→ caller falls back to Playwright).
  *
- * Two endpoint shapes:
+ * Three endpoint shapes:
  *   - Per-job (Greenhouse, Lever, Workday): the URL maps to a single-job endpoint,
  *     so a 200 is itself proof the posting is live.
  *   - Org-level (Ashby): the URL maps to the org's whole job board. A 200 only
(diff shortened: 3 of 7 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/liveness-api-linkedin.test.mjs` and its three fixtures; `liveness-api.mjs` and `check-liveness.mjs` are outside my area and I read them for context.

The test imports the live classifier and reads the fixtures off disk, stubbing `globalThis.fetch` so it stays hermetic with no network. The verdicts are right per fixture: the closed page reads `expired`, the on-site and modal pages read active, and the both-signals and neither cases read `uncertain`. The sentinels are genuine, not vacuous: the active assertions redden if the apply-control signal is dropped and the expired ones if the closed marker is dropped, so neither direction passes alone. The fixtures are worth the space, roughly twenty lines each of byte-realistic LinkedIn markup carrying the exact signals the classifier keys on, with their provenance documented in the header, and the both-signals case is built by concatenating real fixtures rather than a stub.

santifer:
Another strong one, @rubicon. The reserve-then-wait throttle is the detail I want to point at: bumping nextRequestAt BEFORE sleeping so back-to-back callers queue instead of all reading the same timestamp and firing together, and waiting before arming the timeout so the spacing never eats the request budget. And the negative control in the suite (forcing the classifier to always answer expired turns 9 assertions red, including neither-signal-is-uncertain) means the harmful checker cannot pass green. Also appreciated the drive-by fix to the stale SCRIPTS.md rung list. Merged! 🚀

PR #3181, opened 2026-08-21
fix(classify-tier): Associate + an academic rank is not entry-level

### LINKED ISSUE #3178: classify-tier: Associate Professor / Dean / Provost / General Counsel classify as entry, so skip_tiers silently drops them
### Code of Conduct

### Existing issues

### What happened?

`classify-tier.mjs:105-111` resolves `associate` + a senior noun to `senior`, so the `associate` prefix qualifies a senior role's band instead of demoting it:

```js
if (/\b(director|vice\s+president|vp|principal|partner|chief|head\s+of)\b/i.test(afterAssociate)) {
  return 'senior';
}
```

That noun list is closed, and it is missing the academic ladder. Every academic title where `associate` names a *rank* rather than a junior variant falls through to the `associate` matcher at `:72` (`tier: 'entry'`):

```
entry   Associate Professor
entry   Associate Dean
entry   Associate Dean of Students
entry   Associate Provost
entry   Associate Chancellor
entry   Associate General Counsel

senior  Associate Director            ← the existing guard, working
senior  Associate Creative Director
senior  Associate Vice President
senior  Associate Principal
senior  Associate Partner
```

`Associate Professor` is the rank *above* Assistant Professor and is normally the tenured one. `Associate Dean` / `Associate Provost` / `Associate Chancellor` are senior administration. None of them is entry-level under any reading.

### Why it costs a real user something

This is silent, and it is reachable from the shipped defaults alone:

1. `templates/portals.example.yml:277` suggests exactly this config:
   ```yaml
   # skip_tiers: [intern, entry]  # optional: skip listings by seniority tier
   ```
2. The same template ships a HigherEdJobs board (`:2270-2273`), whose entire catalogue is academic titles.
3. `scan.mjs:2606` drops the posting without naming it:
   ```js
   if (classifyTier && skipTiers.includes(classifyTier(job.title))) {
   ```

So a candidate scanning academic boards with the documented example config loses every Associate Professor posting and is never told. That is the same failure mode the leftmost-marker comment at `:133-136` already calls out for the junior-candidate case — *"scan.mjs drops a posting whose tier is in `skip_tiers` without naming it"* — just pointed at a different population.

### Steps to reproduce

```js
// repro.mjs — run from the repo root: node repro.mjs
import { classifyTier } from './classify-tier.mjs';

for (const t of ['Associate Professor', 'Associate Dean', 'Associate Provost',
                 'Associate General Counsel', 'Associate Director']) {
  console.log(classifyTier(t).padEnd(8), t);
}
```

Output on `main` @ `7d97d14`:

```
entry    Associate Professor
entry    Associate Dean
entry    Associate Provost
entry    Associate General Counsel
senior   Associate Director
```

### Expected behavior

The academic ranks and the `general counsel` compound stop classifying as `entry`.

Worth stating explicitly: **for the harm above, `mid` and `senior` are equally good** — the documented config skips `intern` and `entry`, so anything other than `entry` stops the silent drop. The choice between them is a refinement, not the fix, and it's your call:

- `senior` is what the existing guard already returns, so it needs no new machinery.
- `mid` is arguably truer for `Associate Professor` specifically, since the academic ladder runs Assistant → Associate → Full, making it the middle rung rather than the top.

`Associate Dean` / `Associate Provost` / `Associate Chancellor` / `Associate General Counsel` read as `senior` either way.

### Deliberately NOT proposed

The counterexamples matter more than the examples here — in plenty of fields `associate` genuinely *does* mark the junior variant, and all of these are correctly `entry` today:

```
entry   Associate Counsel        entry   Associate Editor
entry   Associate Attorney       entry   Associate Producer
entry   Associate Manager        entry   Associate Consultant
```

Note `Associate Counsel` vs `Associate General Counsel`: in a law firm an "associate" is a junior lawyer, so bare `counsel` must stay out of the list. Only the `general counsel` compound moves. Widening the noun list generically — or reaching for an "associate never means junior" rule — would break all six of the titles above, which is presumably why the list was closed to begin with.

So the proposal is narrow: add `professor`, `dean`, `provost`, `chancellor` and the `general\s+counsel` compound, and nothing else.

### Environment

- career-ops version: `main` @ `7d97d14`
- Node.js: v22
- OS: macOS

### Additional context

Pre-existing rather than a regression — the `associate` → `entry` matcher dates to #1361, and the guard added in #2737 is what makes the asymmetry visible (some `Associate X` titles now resolve senior, these don't).

Happy to send the PR with a regression test once you've picked `senior` vs `mid` for the professor case.

FReptar0:
Fixes #3178

## The bug

`classify-tier.mjs` guard (a) already resolves `associate` + a senior noun to `senior` — the `associate` prefix qualifies a senior role's band rather than demoting it. The noun list just never covered the academic ladder, so these fell through to the `associate` → `entry` matcher at `:72`:

```
entry   Associate Professor        entry   Associate Chancellor
entry   Associate Dean             entry   Associate General Counsel
entry   Associate Provost

senior  Associate Director         ← the existing guard, working
senior  Associate Creative Director
```

`scan.mjs` drops a posting whose tier is in `skip_tiers` and never names it — it prints a `Filtered by tier: N removed` count, not the title. So an academic candidate running the documented example config (`templates/portals.example.yml` suggests `skip_tiers: [intern, entry]` and ships a HigherEdJobs board, both commented out) lost every Associate Professor posting without seeing which.

## The change

Adds `professor` / `dean` / `provost` / `chancellor` / `superintendent` and the `general counsel` compound.

The criterion is **institution-level head, not office-level deputy** — which is why `Registrar`, `Bursar` and `Librarian` stay off, and `Rector` too, since a parish's associate rector genuinely *is* the junior one. The list has to stay closed: in plenty of fields `associate` really does mark the junior variant, and `Associate Attorney` / `Editor` / `Producer` / `Manager` / `Consultant` must keep resolving `entry`.

**Two bounds come with it, because widening the list alone made the guard worse than it was.** An adversarial pass over the first draft found both, and both are reachable from ordinary higher-ed titles:

**1. Guard (a) is skipped when an intern/entry marker leads the title.** It returns before the position loop, so:

```
Intern, Associate Dean of Student Life           →  senior   (before)
Student Intern, Office of the Associate Provost  →  senior   (before)
```

That inverts this fix's own harm on the same board — the junior candidate loses the dean's-office internships instead. It's the same doctrine the position loop already applies to "Summer Intern, Director of Product".

**2. The window is whitespace-only and at most two words.** It used to slice to end-of-string, so any senior noun anywhere to the right matched:

```
Administrative Associate, Office of the Dean     →  senior   (before)
Research Associate - Professor Smith Laboratory  →  senior   (before)
Associate Consultant - Dean & Company            →  senior   (before)
Associate Manager, Dean Foods                    →  senior   (before)
```

A comma or dash after `associate` means `associate` is the *role* and what follows is a separate clause. The two-word cap then stops "Office of the X" and real employers named for a noun on the list — Dean & Company (whose entry title is literally "Associate Consultant"), Dean Foods, Dean Witter, Provost Umphrey.

That second bound is also what makes the counsel split **real rather than asserted**. Bare `counsel` is deliberately off the list, but unbounded:

```
Associate Counsel                                 →  entry
Associate Counsel, Office of the General Counsel  →  senior   (before)
```

Same junior in-house lawyer, opposite tier, decided by the department name.

Both bounds are pre-existing in `director` — this change surfaces them because the academic nouns land in titles where they fire constantly.

## Test

Extends `tests/classify-tier-position.test.mjs` (79 assertions, 0 failed). Reverting `classify-tier.mjs` fails exactly 9 of them.

The counterexample blocks are the load-bearing half: they pin the titles where `associate` really is junior, so the noun list can't quietly become a generic "associate never demotes" rule.

## Scoped out, deliberately

- **The top of each ladder sits at `mid`.** `Professor`, `Dean`, `Provost`, `General Counsel` with no prefix hit no matcher and fall to the `mid` default, so `Associate Professor` now outranks `Professor`. That's odd, but fixing it means bare `dean`/`professor` matchers, which need the same windowing (otherwise every "Dean & Company" title goes senior) and would flip `Assistant Professor` too. Separate change, happy to take it if you want it.
- **`Staff Associate` → `senior`**, pre-existing and unrelated: the `staff` matcher fires at index 0. Narrowing that risks `Staff Engineer`, so I left it alone.
- **`senior` vs `mid` for `Associate Professor`.** #3178 asked; either fixes the harm, since the documented config skips `intern` and `entry`. This uses `senior` because it needs no new machinery — flipping it is a one-line change if you'd rather.

## Verification

- `node test-all.mjs --quick` → **5152 passed, 0 failed** (2 pre-existing warnings)
- `node validate-system-paths-coverage.mjs` → clean; `classify-tier.mjs` already in `SYSTEM_PATHS`, no new files
- Blast radius is exactly `skip_tiers` — `scan.mjs:2606` is the only caller; `role-matcher.mjs`, `web/src/lib/inbox.ts` and `dashboard/internal/data/stats.go` do their own string matching and never call `classifyTier`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved classification of associate-level titles, including academic and institutional senior roles.
  * Preserved junior classification when intern or entry-level markers appear first.
  * Prevented incorrect matches caused by punctuation, unrelated organization names, or overly broad title text.

* **Tests**
  * Added coverage for senior, junior, punctuation, spacing, and employer-name classification scenarios.

### DIFF
--- a/classify-tier.mjs
+++ b/classify-tier.mjs
@@ -96,17 +96,49 @@ export function classifyTier(title) {
     }
   ];
 
-  // Guard (a): "Associate [*] Director/VP/Chief/Principal/Partner/Head-of" resolves
-  // to senior. The `associate` prefix qualifies the seniority band of a senior role;
-  // it does not demote it to entry-level. Works for both a direct compound
-  // ("Associate Director") and a broad one ("Associate Creative Director"). Checked
-  // before the position loop because `associate` always precedes the senior noun,
-  // so the leftmost-marker rule would fire on `associate` at index 0 and return entry.
+  // Guard (a): "Associate <senior noun>" resolves to senior. The `associate`
+  // prefix qualifies the seniority band of a senior role; it does not demote it
+  // to entry-level. Checked before the position loop because `associate` always
+  // precedes the senior noun, so the leftmost-marker rule would fire on
+  // `associate` at index 0 and return entry.
+  //
+  // The noun list is CLOSED, and has to stay that way: in plenty of fields
+  // `associate` genuinely does mark the junior variant, and those must keep
+  // resolving to entry — Associate Attorney, Associate Editor, Associate
+  // Producer, Associate Manager, Associate Consultant. A generic "associate
+  // never demotes" rule breaks every one of them.
+  //
+  // The academic ranks are on the list because `associate` names a RANK there
+  // rather than a junior variant: Associate Professor is the rung above
+  // Assistant Professor, and Dean/Provost/Chancellor/Superintendent head an
+  // institution (#3178). The criterion is institution-level head, not
+  // office-level deputy — which is why Registrar, Bursar and Librarian stay
+  // off, along with Rector (a parish `associate rector` IS the junior one).
   const associateAt = cleanTitle.search(/\bassociate\b/i);
   if (associateAt >= 0) {
-    const afterAssociate = cleanTitle.slice(associateAt + 'associate'.length);
-    if (/\b(director|vice\s+president|vp|principal|partner|chief|head\s+of)\b/i.test(afterAssociate)) {
-      return 'senior';
+    // A junior marker that LEADS the title still decides it. "Intern, Associate
+    // Dean of Student Life" is an internship in a dean's office, not a
+    // deanship — the same doctrine the position loop below applies to
+    // "Summer Intern, Director of Product". Without this the guard returns
+    // early and inverts the very harm #3178 is about, on the same board.
+    const juniorAt = cleanTitle.search(/\b(?:intern(?:ship)?|trainee|co-op|graduate|junior|entry(?:-level)?)\b/i);
+    if (juniorAt < 0 || juniorAt > associateAt) {
+      const afterAssociate = cleanTitle.slice(associateAt + 'associate'.length);
+      // WHITESPACE only, and at most two words of gap. Both bounds carry
+      // weight. Any comma or dash after `associate` means `associate` is the
+      // role and what follows is a separate clause: "Research Associate -
+      // Professor Smith Laboratory" and "Administrative Associate, Office of
+      // the Dean" are junior roles that a to-end-of-string search reads as
+      // senior. The two-word cap then stops "Office of the Dean" and real
+      // employers named for a noun on this list — Dean & Company (whose entry
+      // title is literally "Associate Consultant"), Dean Foods, Dean Witter,
+      // Provost Umphrey. It also keeps the legal pair honest: bare `counsel`
+      // is off the list because a firm's associate IS the junior lawyer, and
+      // without the bound "Associate Counsel, Office of the General Counsel"
+      // would match `general counsel` from the department name.
+      if (/^\s+(?:[a-z]+\s+){0,2}(director|vice\s+president|vp|principal|partner|chief|head\s+of|professor|dean|provost|chancellor|superintendent|general\s+counsel)\b/i.test(afterAssociate)) {
+        return 'senior';
+      }
     }
   }
 
--- a/tests/classify-tier-position.test.mjs
+++ b/tests/classify-tier-position.test.mjs
@@ -108,6 +108,74 @@ try {
   check('Intern Program Coordinator', 'intern');
   check('Graduate Scheme Analyst', 'intern');
   check('Trainee Program Engineer', 'intern');
+
+  // ── Guard (a): `associate` names a RANK, not a junior variant (#3178) ──
+  // These fell through the closed noun list to the `associate` matcher at
+  // classify-tier.mjs:72 and classified `entry`. scan.mjs drops a posting whose
+  // tier is in `skip_tiers` and never names it (it prints a "Filtered by tier: N
+  // removed" count, not the title), so an academic candidate running the
+  // documented example config — templates/portals.example.yml suggests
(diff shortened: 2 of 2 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/classify-tier-position.test.mjs`; `classify-tier.mjs` is tier-classification, which is santifer's scoring red line, so I read it for context and my approval covers only the test.

The test pins the new behavior with real sentinels rather than restating the fix. `Associate Professor`, `Associate Dean`, `Associate Provost` all resolved to `entry` under the old noun list and now assert `senior`, so a revert reddens them. The window bound is pinned the same way: `Administrative Associate, Office of the Dean` would read `senior` under an unbounded search but the two-word cap keeps it `entry`. The junior-leading cases pin the early skip that stops the widened list from inverting #3178. Hermetic (pure function, no I/O), reuses the file's own `check()` helper.

Green on CI. Approving the test from my side; the `classify-tier.mjs` change is santifer's to weigh.

santifer:
Merged, @FReptar0, and the process here deserves as much credit as the fix: running an adversarial pass over your own report and correcting reachable-from-defaults to reachable-from-the-documented-example before anyone acted on it is the discipline that makes reports actionable without re-verification. The fix itself is surgical: the academic ladder joins the senior-noun guard, and Associate Professor stops being entry under any reading. 🚀

Issue #1245, opened 2026-06-26
Updater self-reexec breaks on 1.12.0 → 1.13.0: ERR_MODULE_NOT_FOUND for scaffolder/bin/skill-entrypoints.mjs


mihail-sirbu:
## Summary

`node update-system.mjs apply` upgrading **1.12.0 → 1.13.0** crashes during the updater self-reexec with `ERR_MODULE_NOT_FOUND` for `scaffolder/bin/skill-entrypoints.mjs`. The update aborts mid-way (VERSION still 1.12.0, only `update-system.mjs` swapped + staged). A clean machine on 1.12.0 cannot self-update to 1.13.0 without manual intervention.

## Environment

- From: career-ops 1.12.0 → To: 1.13.0
- Node v26.3.0, macOS (darwin)
- Clean working tree (only user-layer/untracked files)

## Error

```
Fetching latest from upstream...
 * branch            main       -> FETCH_HEAD
node:internal/modules/esm/resolve:271
Error [ERR_MODULE_NOT_FOUND]: Cannot find module
'/.../career-ops/scaffolder/bin/skill-entrypoints.mjs'
imported from /.../career-ops/update-system.mjs
  code: 'ERR_MODULE_NOT_FOUND',
Updater self-reexec failed: Command failed: node update-system.mjs apply
```

## Root cause

Two-stage self-reexec ordering bug.

1. The **1.12.0** updater's self-reexec stage checks out **only** `update-system.mjs` from `FETCH_HEAD`, then immediately re-execs it:

   ```js
   // update-system.mjs @ 1.12.0, ~line 534
   git('checkout', 'FETCH_HEAD', '--', 'update-system.mjs');
   execFileSync(process.execPath, ['update-system.mjs', 'apply'], { ... });
   ```

2. The **1.13.0** `update-system.mjs` adds a new **top-level import of a new file**:

   ```js
   // update-system.mjs @ 1.13.0, ~line 22-25
   import {
     ...
   } from './scaffolder/bin/skill-entrypoints.mjs';
   ```

3. `scaffolder/bin/skill-entrypoints.mjs` does not exist on disk yet (it is only created later, via `BOOTSTRAP_PATHS` checkout, which never runs because the re-exec crashes at ESM import resolution first). Result: `ERR_MODULE_NOT_FOUND` before any update work happens.

The 1.13.0 updater already self-corrects for the *next* hop (it checks out both files before re-exec):

```js
[...]

PR #3182, opened 2026-08-21
fix(rename-sync): migrate 5 vulnerable call sites to renameSyncWithRetry

### LINKED ISSUE #3046: unify: five more bare renameSync calls carry the Windows contention class #3006 just fixed in the tracker
#3006 fixed Windows rename contention in the canonical spot (`tracker-utils.mjs`'s `writeFileAtomic`) and exported the machinery: `isRenameContention` + `renameSyncWithRetry`, one definition, injectable, tested. A repo census shows the same bare-`renameSync` pattern in 7 more files. Classified by exposure:

**The vulnerable shape — write-tmp-then-rename over a LIVE destination** (Windows answers EPERM/EACCES/EBUSY when anything holds the destination handle; the writer dies and the write is lost):

- `followup-seed.mjs:447` — `data/follow-ups.md` (inside its lock, same as the tracker was)
- `scan-ats-full.mjs:135` — scan checkpoint (the comment says "atomic: a crash mid-write can't corrupt the checkpoint", which is true on POSIX and half-true on Windows: no corruption, but the write can die)
- `discover-ats.mjs:885` — `portals.yml`
- `paste-reply.mjs:134` — candidates file
- `update-system.mjs:1302` — system file updates

**Lower exposure, check rather than assume:**
- `merge-tracker.mjs:1239` — moves files between directories (destination usually doesn't pre-exist; confirm)
- `plugin-install.mjs:66` — already wrapped in a tolerant catch with a documented fallback

**The task:** migrate the vulnerable five to `renameSyncWithRetry` from `tracker-utils.mjs` (import, swap the call — the signature matches `renameSync`). Do NOT re-derive the classifier or the backoff: one definition is the point (#2984 patched 2 of 4 lock copies and the drift carried three bug faces for weeks; the derived-list test in `tests/lock-rm-contention.test.mjs` exists because of it). A test in the same shape as `tests/rename-contention.test.mjs` §4 — no bare renameSync in the migrated files' atomic paths — would pin the family shut.

@yaserselvam found and fixed the class in #3006 (measured on real Windows), so this has their name on it first. If unclaimed in a few days it's open to anyone — comment to take it.

yaserselvam:
Closes #3046.

Same write-tmp-then-rename-over-a-live-destination shape #3006 fixed in `tracker-utils.mjs`'s `writeFileAtomic`, present in five more files. Migrated all five (import + swap the call, signature matches `renameSync`):

- `followup-seed.mjs:447` (`data/follow-ups.md`, inside its lock)
- `scan-ats-full.mjs:135` (scan checkpoint)
- `discover-ats.mjs:1027` (`portals.yml`)
- `paste-reply.mjs:134` (candidates file)
- `update-system.mjs:1302` (`.gitignore` reconciliation)

`update-system.mjs` needed extra care: it must stay self-loading (#1706) — a static top-level import of `tracker-utils.mjs` crashes a pre-#1245 client's old→new re-exec. `writeGitignoreAtomic` is now `async` and pulls `renameSyncWithRetry` in lazily via `await import(...)`; both call sites now `await` it.

Left alone, per the issue's own classification, confirmed by reading both:
- `merge-tracker.mjs:1326` — moves a freshly processed TSV into `merged/`, a destination `mkdirSync`'d immediately above and unique per run, not a live file.
- `plugin-install.mjs:66` — already wrapped in a tolerant catch with a documented in-place fallback.

Added `tests/rename-contention-family.test.mjs`, same shape as `tests/rename-contention.test.mjs` §4: no bare `renameSync` may remain in any of the five migrated files' atomic paths, the renames route through `renameSyncWithRetry`, and the helper is imported from the one canonical definition (static, or `update-system.mjs`'s lazy form) rather than reimplemented.

`test-all.mjs`: 4309/20 failing → 4324/5 failing. Confirmed via `git stash` against the same baseline that the remaining 5 are pre-existing/environment (`node:sqlite` unavailable on this Node 20 runtime, plus two unrelated coverage-gap failures) — none touch rename paths.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when replacing files during concurrent or temporarily locked filesystem operations.
  * Preserved atomic updates for portal files, checkpoints, candidate data, and configuration files.

* **Tests**
  * Added regression coverage to verify retry-enabled file replacement across key workflows.

### DIFF
--- a/discover-ats.mjs
+++ b/discover-ats.mjs
@@ -31,10 +31,11 @@
  * Issue #1864 — github.com/santifer/career-ops
  */
 
-import { readFileSync, existsSync, writeFileSync, renameSync } from 'fs';
+import { readFileSync, existsSync, writeFileSync } from 'fs';
 import { dirname, join, resolve } from 'path';
 import { fileURLToPath, pathToFileURL } from 'url';
 import * as yaml from 'js-yaml';
+import { renameSyncWithRetry } from './tracker-utils.mjs';
 
 import { makeHttpCtx } from './providers/_http.mjs';
 import greenhouse from './providers/greenhouse.mjs';
@@ -1024,7 +1025,7 @@ async function main() {
     // mid-write can't leave the user's portals.yml truncated.
     const tmpPath = `${PORTALS_PATH}.tmp-${process.pid}`;
     writeFileSync(tmpPath, insertIntoTrackedCompanies(current, snippets), 'utf-8');
-    renameSync(tmpPath, PORTALS_PATH);
+    renameSyncWithRetry(tmpPath, PORTALS_PATH);
     written = true;
   } else if (opts.write && fresh.length && !existsSync(PORTALS_PATH)) {
     warnings.push(`--write given but portals.yml not found at ${PORTALS_PATH} — printing entries instead`);
--- a/followup-seed.mjs
+++ b/followup-seed.mjs
@@ -55,7 +55,7 @@
  *   CAREER_OPS_FOLLOWUPS_LOCK_STALE_MS     stale-lock recovery threshold
  */
 
-import { readFileSync, writeFileSync, existsSync, mkdirSync, renameSync, rmSync, statSync, realpathSync } from 'fs';
+import { readFileSync, writeFileSync, existsSync, mkdirSync, rmSync, statSync, realpathSync } from 'fs';
 import { join, dirname, basename, resolve, isAbsolute, relative, sep } from 'path';
 import { fileURLToPath, pathToFileURL } from 'url';
 import { createHash, randomUUID } from 'crypto';
@@ -68,6 +68,7 @@ import {
   isMkdirContention, isRmContention, rmLockArtifactSync, createLockWaitPolicy,
   sameLockDirectory, lockRecoveryVerdict, RECOVER_STALE,
 } from './pipeline-lock.mjs';
+import { renameSyncWithRetry } from './tracker-utils.mjs';
 import { tmpdir } from 'os';
 import { resolveColumns, parseTrackerRow } from './tracker-parse.mjs';
 import { localToday } from './lib/local-today.mjs';
@@ -440,7 +441,7 @@ function writeFileAtomic(filePath, content) {
   const tmpPath = join(dirname(filePath), `.${basename(filePath)}.${process.pid}.${Date.now()}.${randomUUID()}.tmp`);
   try {
     writeFileSync(tmpPath, content);
-    renameSync(tmpPath, filePath);
+    renameSyncWithRetry(tmpPath, filePath);
   } catch (err) {
     rmSync(tmpPath, { force: true });
     throw err;
--- a/paste-reply.mjs
+++ b/paste-reply.mjs
@@ -46,6 +46,7 @@ import fs from 'node:fs';
 import path from 'node:path';
 import readline from 'node:readline';
 import { fileURLToPath, pathToFileURL } from 'node:url';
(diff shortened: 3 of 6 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/rename-contention-family.test.mjs`; `update-system.mjs` is the red-line updater and the other four scripts are outside my area, all read for context.

The test is a source-inspection guard and it is built correctly. It derives the five migrated files rather than hardcoding a count, and runs three assertions per file. The direct-call regex catches a bare `renameSync(` but not `renameSyncWithRetry(`, so the safe seam is not flagged, and it asserts zero direct calls. The false-pass control is there and named: a file could pass the no-direct-calls check by deleting the write entirely, so it also asserts the seam is called at least once, and it pins the helper to the canonical `tracker-utils.mjs` so a local reimplementation cannot slip through. Hermetic, reads only from `ROOT`, same derived-list shape as `lock-rm-contention.test.mjs`.

Green on CI. Approving the test from my side. The `update-system.mjs` change itself is the updater, which is santifer's call, not mine.

santifer:
Exactly what #3046 asked for and a bit more, @yaserselvam. The update-system.mjs handling is the part worth naming: instead of a static import that would break the old-to-new re-exec for pre-#1245 clients, you made writeGitignoreAtomic async with a lazy import and awaited both callers. And verifying the two exclusions by reading them (merge-tracker's per-run mkdir, plugin-install's tolerant catch) instead of trusting the issue's classification is the standard. The rename-contention family now has one definition and five importers. Merged! 🚀

yaserselvam:
Thanks for the merge. Two things:

1. CodeRabbit flagged `parsePorcelainStatus` (update-system.mjs:772-785) outside this diff — `git status --porcelain=v1` quotes special paths and encodes renames as two paths, and the current `clean.slice(3)` parse doesn't account for either, so the user-layer safety check can miss changed local paths. Real bug, unrelated to #3046, nobody's claimed it. Want me to open an issue and take it, or is it already on your radar?

2. More generally — anything open you'd point me at next? update-system.mjs seems to be where I have the most context right now (3201/3149/3129/3052/3048/2957 are all there), but happy to go wherever's most useful.

santifer:
Open it and it's yours, @yaserselvam. I read `parsePorcelainStatus` against your claim before answering: correct on both counts — `clean.slice(3)` hands back a quoted path *with* its quotes and escapes for any path git decides to quote, and a rename line carries `old -> new` as a single string, so the user-layer safety check compares against a path that doesn't exist on disk in both cases. That's a real hole in a safety check, unrelated to the rename-retry work this PR closed. File the issue with those two cases as the acceptance list and take it — porcelain v1's quoting rules are documented in `git-status(1)`, and a fixture-driven test (quoted path, rename, CRLF) would pin all three forms at once.

yaserselvam:
Filed as #3417, with the quoted-path / rename / CRLF acceptance list you called out. Went ahead and took it — #3418: `gitStatusEntries()` now parses `git status --porcelain -z`, which drops quoting, splits renames into two fields, and retires the CRLF/leading-space workarounds in one move. Fixture cases for all three forms.

santifer:
You didn't wait for anyone, @yaserselvam: filed #3417 with the acceptance list, then took it yourself in #3418. That's the shortest distance between a review comment and a fix I've seen here in a while.

And `git status --porcelain -z` is the right move rather than the convenient one: it retires the CRLF and leading-space workarounds instead of adding a third special case on top of them. Fixtures for all three forms is what makes that safe to believe.

PR #3184, opened 2026-08-21
fix(pdf): name both sides when the lexical containment check fires

santifer:
Part of #3162. Occurrence 5 (logged by @FReptar0 on #3181's run) reproduced on a tree that **carries** the #3172 canonicalization fix: `20ca092` is a descendant of `7d97d14` (verified with `git merge-base --is-ancestor`), so that hypothesis is ruled out as the cure.

Since #3163 split the two failure paths, the surviving `escapes the tracker workspace` message can only come from the **lexical branch after a successful `realpathSync`**. This PR makes that branch print `workspaceRoot`, the canonicalized path and their `relative()` form, so the next occurrence names the disagreeing ancestor outright (both are runner-local temp paths, fine to surface in a test failure).

Verified: the only assertion pinning this message (`tests/generate-pdf-batch.test.mjs:326`) is a prefix regex, unaffected by the appended detail. Full `test-all.mjs --quick` on this branch: 5,183 passed / 0 failed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workspace escape error messages by including the workspace root, canonical path, and relative path for easier troubleshooting.

### DIFF
--- a/generate-pdf.mjs
+++ b/generate-pdf.mjs
@@ -95,7 +95,14 @@ function assertInsideWorkspace(absPath, label) {
   }
   const rel = relative(__workspaceRoot, canonical);
   if (rel === '' || rel.startsWith('..') || isAbsolute(rel)) {
-    throw new Error(`${label} escapes the tracker workspace: ${absPath}`);
+    // #3162: an intermittent macOS-CI-only hit of this branch happens on paths
+    // that are lexically inside the sandbox, and canonicalization SUCCEEDS
+    // before it. Print both sides so the next occurrence names the disagreeing
+    // ancestor outright instead of asking a reader to reconstruct it.
+    throw new Error(
+      `${label} escapes the tracker workspace: ${absPath}`
+      + ` (workspaceRoot=${__workspaceRoot} canonical=${canonical} rel=${rel})`,
+    );
   }
   return absPath;
 }


Solaris-star (approved):
Reviewed commit `9557835d22d09a9aec5012e0f5e85fd60a0d127d` against `main` `ae70a410bc77552b6a52ad15ab1ecb78f41f2fa3`.

I exercised the lexical containment branch directly with an external batch manifest. It still exited 1 before reading/writing the manifest and now reported all three diagnostic values (`workspaceRoot`, `canonical`, and `rel`), while preserving the existing `batch manifest escapes the tracker workspace` prefix.

Local verification on a clean detached worktree:

- `node --test tests/generate-pdf-batch.test.mjs` — passed (1 suite; all 8 batch cases reported green)
- `node test-all.mjs --quick` — 5,181 passed, 0 failed, 3 existing warnings
- `git diff --check upstream/main...HEAD` — passed
- tracked-secret scan — 1,166 tracked files checked; 32 baseline fixture/config-pattern hits, 0 findings in added lines

I did not reproduce the intermittent macOS CI failure itself; this review verifies the new evidence emitted when the already-observed lexical branch fires. The one-file change is scoped to diagnostics, and the current GitHub matrix is green across Ubuntu, macOS, Windows, CodeQL, dependency review, guard, GitGuardian, and CodeRabbit.

PR #135, opened 2026-04-09
Fix expired job links that still look active

lawrence3699:
Fixes #129.

WebSearch results can stay cached long after a posting has closed, so the liveness check is supposed to filter dead links before they reach the pipeline. The current check treats any `Apply` text in the page body as a strong active signal, which means expired pages with a generic careers header or footer CTA can still slip through as active.

Before this change, a page like this was classified as active:

- body contains `Apply`
- body also contains `The job you are looking for is no longer open`
- result: `active`

After this change, the checker only counts visible apply controls from the main page content and ignores generic header/footer navigation. Explicit expired signals still win, while real job pages with an actual apply control continue to classify as active.

I pulled the liveness decision into a small shared classifier so the edge case is regression-tested without needing a live ATS page, and updated `scan.md` to match the implemented rule.

Checks run:

- `node test-all.mjs --quick`
- `node check-liveness.mjs file:///tmp/career-ops-expired-fixture.html file:///tmp/career-ops-active-fixture.html`

Notes:

- `node test-all.mjs` still hits the existing dashboard build dependency fetch in this environment because `go build` needs to download modules. I didn't touch the dashboard in this change.

PR #3185, opened 2026-08-21
fix(tests): assert the flag echo with includes instead of an unescaped RegExp

santifer:
Fixes CodeQL alert #135 (`js/identity-replacement`, medium): `typo.replace(/^--/, '--')` replaces `--` with `--`, a no-op. The intent was to keep the RegExp safe for the flag text, and it worked only because the current fixtures carry no regex metacharacters. `String.prototype.includes` states the intent directly and removes the footgun for future fixtures.

Behavior-neutral for the existing table: all fixtures still pass, `test-all.mjs --quick` 5,183 / 0 on a tree carrying this change.

## Summary by CodeRabbit

* **Tests**
  * Improved validation of command-line typo error messages by checking for the exact reported text.

### DIFF
--- a/tests/cli-flag-validation.test.mjs
+++ b/tests/cli-flag-validation.test.mjs
@@ -41,7 +41,7 @@ for (const [script, typo] of SCRIPTS) {
     const r = runScript(script, typo, 'some-value');
     assert.equal(r.status, 1, `${script} ${typo} exited ${r.status}, want 1`);
     assert.match(r.all, /unrecognized flag/i, `${script} did not name the unrecognized flag`);
-    assert.match(r.all, new RegExp(typo.replace(/^--/, '--')), `${script} did not echo ${typo} back`);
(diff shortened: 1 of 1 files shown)


Scott-Emberson (commented):
Reviewed as code owner of `tests/`. Scope is the single changed line in `tests/cli-flag-validation.test.mjs`.

The change is correct: swapping `new RegExp(typo.replace(/^--/, '--'))` for `r.all.includes(typo)` is the right assertion for an echoed flag. That replace was a no-op, so the flag string was fed to `RegExp` unescaped and only held because no current flag carries a metacharacter, and a substring is what "the flag was echoed back" actually means.

Commenting rather than approving because this is the same one-line change, on the same line, as #3132, which I already approved. Only one of the two should land. My suggestion is to close this in favour of #3132, or if you would rather keep this one, close #3132 instead, but not both. (#3092 and #3122 also touch this file, on different assertions, so those do not collide with either.)

PR #3186, opened 2026-08-21
feat(providers): job.description for recruitee, ashby, and opt-in smartrecruiters (#3175)

### LINKED ISSUE #3175: Populate job.description on Greenhouse boards so visa/content/country filters stop passing blind

BigBrown10:
Phase 2 of #3175: extends `job.description` coverage past greenhouse to the three providers whose boards still pass content_filter/visa_filter blind.

## What changed

**Recruitee + Ashby needed no new requests at all.** Verified against live boards before writing code:

- Recruitee's `/api/offers/` list payload already embeds each offer's full HTML body. It now runs through a plain-text strip (`htmlToText`), so descriptions arrive with zero extra requests.
- Ashby's posting-api list already ships `descriptionPlain` — mapped verbatim, mirroring lever.

**SmartRecruiters is the one true N+1**: its list payload carries no body. Rather than taxing every SR board by default, detail enrichment is opt-in per tracked_companies entry (mirrors vdab):

```yaml
smartrecruiters:
  fetchDetails: true   # fetch each posting's detail JSON for descriptions
  detailLimit: 25      # max detail calls per sweep (default 25, range 1-100)
```

Details run in batches of 5, are skipped entirely while verify-portals probes (`ctx.maxPages`, same rule as vdab), tolerate per-job failures silently, join the four known `jobAd.sections` in fixed order, and the internal posting id is stripped before rows reach the pipeline.

## Shared pipeline

The stripping logic is extracted from greenhouse into `providers/_html-to-text.mjs` (`htmlToText`, cap 4000 like alibaba); greenhouse's tested `contentToText` export now delegates to it instead of keeping a private copy — the same consolidation rationale that produced `_html-entities`.

## Verification

- New/extended suites green: `providers/_html-to-text` (9), `providers/recruitee` (16), `providers/smartrecruiters` (32), `providers/ashby` (31)
- `npm run lint`: clean, 518 files
- `node test-all.mjs`: **5184 passed, 1 failed** — the failure is `tests/plugin-symlink-discovery.test.mjs` dying on `symlinkSync` EPERM (Windows without Developer Mode). Pre-existing/environmental, unrelated to this change; happy to re-run on CI where symlinks are available.
- Live smoke through the actual provider code:
  - Hostaway (Recruitee): 13/13 jobs with clean plain-text descriptions
  - Linear (Ashby): 32/32 with `descriptionPlain`
  - Wise (SmartRecruiters): 426 listed, exactly 5 enriched at `detailLimit=5`, id stripped, and one enriched body surfaced a real sponsorship statement ("unable to sponsor OR transfer Visa's...") — precisely what visa_filter was missing

Closes #3175 (phases 1-2; Workday/SuccessFactors were declared out of scope in the issue).

## Summary by CodeRabbit

* **New Features**
  * Job listings from Ashby, Recruitee, Greenhouse, and SmartRecruiters now include cleaner plain-text descriptions.
  * SmartRecruiters listings can optionally fetch enriched descriptions, with configurable request limits.
  * Descriptions are safely cleaned, normalized, and capped at 4,000 characters.

* **Documentation**
  * Updated provider documentation to describe description sources and enrichment behavior.

### DIFF
--- a/providers/README.md
+++ b/providers/README.md
@@ -70,7 +70,8 @@ There is no index file — discovery is filesystem-convention-based
 
 Underscore-prefixed files are shared helpers, never loaded as providers:
 `_types.js` (contract typedefs), `_registry.mjs` (loader/router),
-`_http.mjs` (HTTP transport), `_html-entities.mjs`, `_trust-validator.mjs`.
+`_http.mjs` (HTTP transport), `_html-entities.mjs`, `_html-to-text.mjs`
+(description HTML → plain text), `_config-utils.mjs`, `_trust-validator.mjs`.
 
 ## Security conventions
 
--- a/providers/README.md
+++ b/providers/README.md
@@ -44,7 +44,11 @@ export default {
 - `url` — required, absolute; used as the dedup key.
 - `company`, `location` — strings, may be empty.
 - `description` — optional; populate ONLY when the list payload carries it
-  for free (no extra per-job request — the scanner is zero-token).
+  for free (no extra per-job request — the scanner is zero-token). The one
+  exception is opt-in, bounded detail enrichment: `fetchDetails: true` plus
+  `detailLimit` in the portals entry (currently vdab and smartrecruiters)
+  fetches per-posting detail JSON to populate `description`, capped at
+  `detailLimit` calls and skipped entirely while a health probe is running.
 - `postedAt` — optional epoch ms; omit when the source has no usable date.
 
 ### Context (`ctx`)
--- a/providers/_html-to-text.mjs
+++ b/providers/_html-to-text.mjs
@@ -1,0 +1,33 @@
+// @ts-check
+// Shared HTML → plain-text pipeline for providers whose payloads embed
+// description markup. Greenhouse's contentToText was the first instance;
+// this is the extracted form so later providers cannot grow a divergent
+// copy — same rationale that produced _html-entities when entity decoders
+// drifted across four files (#1555/#1639/#2623).
+import { decodeEntities } from './_html-entities.mjs';
+
+// Capped like greenhouse/alibaba full-text JDs: a 10 KB/posting body is
+// normal on these boards, and scan payloads must stay sane.
+export const DESCRIPTION_CAP = 4000;
+
+/**
+ * Entity-decoded markup → stripped plain text.
+ *
+ * Double-decode: the payload often carries entity-escaped tags (`&lt;p&gt;`),
+ * so the first pass reveals real tags, and text-level entities (`&amp;`,
+ * `&#39;`) only become decodable once those tags are gone. Plain text is what
+ * the description-consuming filters match against — substring matching over
+ * raw HTML misses keywords split by a tag and pads matches into attribute
+ * soup.
+ *
+ * Exported for tests.
+ *
+ * @param {unknown} content
+ * @returns {string}
+ */
+export function htmlToText(content) {
+  if (typeof content !== 'string' || !content) return '';
+  const html = decodeEntities(content);
+  const noMedia = html.replace(/<(script|style)[^>]*>[\s\S]*?<\/\1>/gi, ' ');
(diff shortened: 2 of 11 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is the four provider test files; the provider sources are outside my area and I read them for context.

The tests genuinely exercise the new `job.description` behavior, and each feature carries a sentinel that fails if it regresses to a no-op. `_html-to-text` pins the double-decode and strips a `<script>evil()</script>` case, which reddens if raw-text contents leak. `ashby` maps `descriptionPlain` and guards the type-narrowing with a `descriptionPlain: 42` case that would leak the number if the `typeof === 'string'` check were dropped. `recruitee` checks the key is omitted rather than set empty. `smartrecruiters` is the strongest of the four: section-order join, default zero detail calls, the `ctx.maxPages` probe skip, and a `detailLimit` cap measured at 40 to 10. All four are hermetic (mocked `fetchJson`, in-memory parse, no network) and match the neighbour shape.

santifer:
Phase 2 lands the same week as phase 1, @BigBrown10, and it's better than the spec: verifying against live boards BEFORE writing code that Recruitee and Ashby already carry the body in the list payload turned two of three providers from planned N+1 into zero extra requests. The Wise smoke (426 listed, exactly 5 enriched at detailLimit=5, including a real no-sponsorship paragraph that visa_filter was missing) is the kind of receipt that makes a merge easy. And extracting _html-to-text.mjs while keeping greenhouse's tested export as a wrapper kept every existing assertion honest. Merged! 🚀

PR #3197, opened 2026-08-22
fix(salary-gap): treat a period as thousands grouping, not a decimal point

### LINKED ISSUE #3174: fix(salary-gap): parseAmount reads a period as a decimal point, so a period-grouped salary folds to 1/1000th
- [x] I searched existing issues and this hasn't been reported yet (nearest prior art: #2190/#2200 on the same function, and open draft #2466 — neither covers the separator)

## What happened?

`parseAmount()` in `salary-gap.mjs` treats `,` as a thousands separator but `.` as a decimal point. In the markets that group with a period — Spain, Germany, Italy, the Netherlands, Brazil, and others this repo ships mode sets for — a salary written the local way is silently divided by 1000 and flows into the gap math as if it were correct.

On a clean `v1.28.0` checkout:

```
"€35.000 - €45.000"  ->  min=35    max=45    mid=40
"35.000 - 55.000"    ->  min=35    max=55    mid=45
"40.000"             ->  min=40    max=40    mid=40
```

This does not return `null` and does not appear in the `quality.unparseable` warning list. Every other parse failure in this script fails closed; this one fails open, which is why I think it deserves separating from the general parser-flexibility work in #2466.

### It is reachable today via `data/salary-observations.tsv`

This is not only a latent problem behind the parser's prose handling. The observation log feeds the same `parseAmount`, so a user logging a real signed offer the way their market writes it gets a confident wrong answer at the **highest trust tier**:

```
data/salary-observations.tsv:
001  2026-08-21  actual  35.000  EUR  contract  signed
```

```
$ node salary-gap.mjs --summary

  #001 Acme — ML Engineer
      desired    70k EUR (profile, 0000-00-00)
      actual     35 EUR (contract, 2026-08-21)
      gap: advertised→actual —, desired→actual -100.0%
```

A €35.000 contract is reported as `35 EUR` and a **-100.0%** gap against the target, sourced from `contract`, with no data-quality warning. That number is the one a user would carry into a negotiation.

### On the report path it is currently masked, not guarded

`modes/oferta.md:631` requires the JD figure be recorded **verbatim** ("the JD's own salary figure **verbatim** … never estimated"), so an evaluator writing `"€35.000 - €45.000 brutos anuales"` is behaving correctly. Today the trailing prose makes the anchored `^…$` match reject the whole string, so the value is dropped rather than mangled. That is luck, not a guard — and #2466 is deliberately widening that stripping. I've noted the ordering there; it seems worth fixing the separator before or alongside it.

### Cause

`salary-gap.mjs:62-65`:

```js
const toNum = (numStr, kFlag) => {
  const n = parseFloat(numStr.replace(/,/g, ''));
  return Number.isNaN(n) ? null : (kFlag ? n * 1000 : n);
};
```

Commas are stripped as grouping. Periods fall through to `parseFloat`, which reads `"35.000"` as `35`.

## Steps to reproduce

```bash
git clone --depth 1 https://github.com/santifer/career-ops.git
cd career-ops && npm install
node -e 'import("./salary-gap.mjs").then(m =>
  console.log(m.parseAmount("€35.000 - €45.000")))'
# => { min: 35, max: 45, mid: 40 }
```

For the end-to-end path, add the single TSV line shown above against any tracked report and run `node salary-gap.mjs --summary`.

## Expected behavior

`"€35.000 - €45.000"` resolves to `min: 35000, max: 45000`, or fails closed into `quality.unparseable`. Either is defensible; silently returning `40` is not.

Ordinary decimals must not regress — both are already asserted in `--self-test`:

```
"82.5k"    -> 82500
"$150,000" -> 150000
```

### Suggested approach

#2722 settled this exact class in `verify-cv-facts.mjs` and is merged: a separator counts as grouping only when followed by **exactly three digits not followed by another digit**, which leaves genuine decimals alone at every precision these values realistically carry.

```js
.replace(/(\d)[,.\s  ](?=\d{3}(?!\d))/g, '$1')
```

That one normalizes for string comparison rather than numeric parsing, so it isn't a drop-in — but the three-digit lookahead is the transferable part, and applying the same window inside `toNum` before `parseFloat` keeps the two scripts consistent and inherits a tradeoff you've already accepted in review.

The residual ambiguity is the same one #2722 documents: a three-place decimal like `"1.250"` can't be disambiguated without knowing the document's locale, and under this rule reads as `1250`. For a salary field I'd argue that's the safer of the two readings, but it's a real tradeoff and belongs in a code comment rather than buried in an issue.

Happy to send this as a PR with self-test coverage — a period-grouped range, a period-grouped single value, the TSV path, and guards asserting `82.5k` and `$150,000` are unaffected — if you'd like it in that form.

## CLI tool

Claude Code

## OS and Node.js version

Ubuntu 24.04.4 LTS / Node v22.23.2

## Output of npm run doctor

```
✓ Node.js >= 22.5 (v22.23.2)
✓ Dependencies installed
✓ cv.md found
✓ config/profile.yml found
✓ modes/_profile.md found
✓ portals.yml found
✓ data/ directory ready
✓ reports/ directory ready

Result: All checks passed (1 warning — see above).
```

kovalancik06-max:
## Summary

Fixes #3174.

`parseAmount()` in `salary-gap.mjs` stripped `,` as a thousands separator but left `.` to fall through to `parseFloat`, so a period-grouped salary (Spain, Germany, Italy, the Netherlands, Brazil, ...) like `"35.000"` silently folded to `35` — a confident wrong answer that doesn't return `null` and never lands in `quality.unparseable`. This is reachable via `data/salary-observations.tsv` at the highest actual-comp trust tier (`contract`), so a real signed offer can produce a `-100%` gap with no data-quality warning.

## Fix

A period now counts as thousands grouping only when followed by exactly three digits and not a fourth (`"35.000"` → `35000`), leaving genuine decimals like `"82.5"` untouched. A three-place decimal (`"1.250"`) is inherently ambiguous without knowing the document's locale — this reads it as grouped, which is the safer default for a salary field (documented in a comment at the call site, per the issue's suggestion).

## Testing

- `node salary-gap.mjs --self-test` passes, with 3 new assertions covering the issue's repro cases (period-grouped range with and without currency symbol, period-grouped single value) plus the existing decimal/comma regression guards.
- Manually verified every case from the issue body, including the `data/salary-observations.tsv` end-to-end path.
- `node test-all.mjs` → 5304 passed, 0 failed (no regressions).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parsing of monetary values that use periods as thousands separators.
  * Preserved accurate handling of decimal values and currency-formatted amounts.
  * Fixed parsing for both individual amounts and value ranges.

### DIFF
--- a/salary-gap.mjs
+++ b/salary-gap.mjs
@@ -60,7 +60,16 @@ export function parseAmount(raw) {
   // its last three letters.
   s = s.replace(/[€$£¥]/g, '').replace(/\s*[A-Za-z]{3}\s*$/, '').trim();
   const toNum = (numStr, kFlag) => {
-    const n = parseFloat(numStr.replace(/,/g, ''));
+    // A period counts as thousands grouping only when it's followed by
+    // exactly three digits and not a fourth ("35.000" -> 35000), so markets
+    // that group with a period (Spain, Germany, Italy, the Netherlands,
+    // Brazil, ...) don't get folded to 1/1000th. Genuine decimals like
+    // "82.5" are untouched since they carry one digit after the point. A
+    // three-place decimal ("1.250") is ambiguous without knowing the
+    // document's locale; reading it as grouped is the safer default for a
+    // salary field (per #3174).
+    const normalized = numStr.replace(/,/g, '').replace(/(\d)\.(?=\d{3}(?!\d))/g, '$1');
+    const n = parseFloat(normalized);
     return Number.isNaN(n) ? null : (kFlag ? n * 1000 : n);
   };
   const range = s.match(/^([\d.,]+)\s*(k)?\s*[-–—]\s*([\d.,]+)\s*(k)?$/i);
@@ -377,6 +386,9 @@ function selfTest() {
   assert(parseAmount('$123,684-$254,644 USD')?.mid === 189164, 'US range, symbol on both bounds, hyphen');
   assert(parseAmount('€80,000-€90,000')?.min === 80000, 'EUR range, symbol on both bounds');
   assert(parseAmount('$150,000')?.mid === 150000, 'single value with symbol still works');
+  assert(parseAmount('€35.000 - €45.000')?.min === 35000 && parseAmount('€35.000 - €45.000')?.max === 45000, 'period-grouped range (#3174)');
+  assert(parseAmount('40.000')?.mid === 40000, 'period-grouped single value (#3174)');
+  assert(parseAmount('35.000 - 55.000')?.mid === 45000, 'period-grouped range, no currency symbol (#3174)');
 
   // parseObservations
   const obs = parseObservations(OBS_FIXTURE);


santifer:
Two for two on your first day, @kovalancik06-max. The comment block documenting WHY a three-place decimal reads as grouped (ambiguous without locale, and grouped is the safer default for a salary field) is exactly the kind of reasoning that makes a one-line regex reviewable in thirty seconds. And the lookahead guard so 82.5 stays a decimal while 35.000 groups is precise work. Merged! 🚀

PR #3199, opened 2026-08-22
fix(followup-cadence): validate CLI flags via the shared helper

### LINKED ISSUE #3196: fix(followup-cadence): mistyped/malformed flags silently fall through to defaults
## Summary

`followup-cadence.mjs` is the same class of defect as #2919: it declares `KNOWN_FLAGS` (`--summary`, `--overdue-only`, `--applied-days`, `--help`, `-h`) but never actually validates argv against it — the list is dead code. It was not part of the original #2919 sweep (that covered `rejection-latency`/`detect-reposts`/`process-quality`/`weekly-digest`) and has no open issue or PR referencing it.

Two failures, measured on current `main`:

```
$ node followup-cadence.mjs --summry
# typo silently ignored; falls through to full JSON output at exit 0, not the summary dashboard

$ node followup-cadence.mjs --applied-days=10
# indexOf('--applied-days') can't see the `=` form, so the override is silently
# discarded and the default (7 days) is used instead - the #2401/#2402 defect class

$ node followup-cadence.mjs --applied-days abc
# parseInt('abc') -> NaN -> positiveInteger() rejects it -> silently falls back
# to the default instead of failing fast
```

Same wrong-answer-at-exit-0 shape the shared helper (`lib/cli-flags.mjs`) exists to close, already applied to `funnel-velocity.mjs` (#3085), `upskill.mjs` (#3057), `fix-slugs.mjs`, `browser-extract.mjs`, and others.

## Proposed fix

- Delegate to `validateFlags`/`flagValue` from `./lib/cli-flags.mjs`, called from inside the existing CLI guard (`if (process.argv[1] && import.meta.url === ...)`) — not at module top level, since this module's `CADENCE` singleton is computed at import time and several tests (`test-all.mjs` section 12) dynamic-`import()` it in-process.
- Switch `--applied-days` lookup from `indexOf` to `flagValue` so both `--applied-days 10` and `--applied-days=10` are honored.
- Reject a non-numeric/negative `--applied-days` value explicitly instead of silently defaulting.
- Add a flags test mirroring `tests/funnel-velocity-flags.test.mjs`.

I have this working locally and would like to open a PR.

AdilAzhari:
Closes #3196

## Summary

`followup-cadence.mjs` declared `KNOWN_FLAGS` (`--summary`, `--overdue-only`, `--applied-days`, `--help`, `-h`) but never actually validated argv against it — a mistyped flag silently fell through to the default (full JSON) behavior at exit 0 instead of failing fast. `--applied-days` had the value-side defect too: `indexOf()` can't see `--applied-days=10`, so the equals form was silently discarded, and a non-numeric/negative value was swallowed by `positiveInteger()` into the same silent default.

Same shape already fixed in `funnel-velocity.mjs` (#3085), `upskill.mjs`, and others in the `lib/cli-flags.mjs` family (#2919/#2775).

## Reproduce (before this PR)

```
$ node followup-cadence.mjs --summry
# typo ignored; prints full JSON instead of the summary dashboard, exit 0

$ node followup-cadence.mjs --applied-days=10
# the override is silently discarded, default cadence used instead

$ node followup-cadence.mjs --applied-days abc
# NaN swallowed by positiveInteger(), silently falls back to the default
```

## Changes

- `followup-cadence.mjs`: delegate to `validateFlags`/`flagValue` from `./lib/cli-flags.mjs`, called from inside the existing CLI guard (`if (process.argv[1] && import.meta.url === ...)`) rather than at module top level — `CADENCE` is a module-level singleton built at import time, and `test-all.mjs` section 12 dynamic-imports this module in-process to read it, so validating there would false-positive on the test runner's own argv.
- Switched `--applied-days` lookup from `indexOf` to `flagValue` so both `--applied-days 10` and `--applied-days=10` work.
- A supplied-but-invalid `--applied-days` value now exits 1 with a clear message instead of silently defaulting.
- `test-all.mjs`: the section-12 e2e fixture copies `followup-cadence.mjs` plus its exact dependency list into an isolated sandbox and runs it as a real subprocess — added the new `lib/cli-flags.mjs` dependency to that copy list (same pattern as the existing `lib/local-today.mjs` line), otherwise the sandboxed run throws `ERR_MODULE_NOT_FOUND`.
- Added `tests/followup-cadence-flags.test.mjs`, mirroring `tests/funnel-velocity-flags.test.mjs`. Note: unlike `funnel-velocity.mjs`, this script's tracker path has no env-var override, so the tests assert only on flag-validation behavior (all of which happens before `analyze()` ever touches disk), not on analysis output.

## Test plan

- [x] `node test-all.mjs` — 5303 passed, 0 failed (confirmed the fixture fix resolves the one failure this change introduced before the `test-all.mjs` update)
- [x] `node --test tests/followup-cadence-flags.test.mjs` — 8/8 passed
- [x] `node scripts/check-syntax.mjs` — 522 files pass
- [x] Manually reproduced all three bugs above against `main`, confirmed each is now a clear exit 1

## Summary by CodeRabbit

* **New Features**
  * Added support for specifying applied days with the `--applied-days=10` format.
  * Added clear handling for help requests and unknown command-line options.

* **Bug Fixes**
  * Invalid, negative, fractional, nonnumeric, or suffixed applied-day values are now rejected.
  * Valid non-negative integer values correctly adjust the cadence.
  * Invalid values preserve the default cadence and provide consistent error messages and exit statuses.

### DIFF
--- a/followup-cadence.mjs
+++ b/followup-cadence.mjs
@@ -36,7 +36,17 @@ const overdueOnly = args.includes('--overdue-only');
 // can't see the `=` form and used to silently discard it, falling back to the
 // default cadence for a value the caller did supply (#2401/#2402 defect class).
 const appliedDaysRaw = flagValue(args, '--applied-days');
-const appliedDaysOverride = appliedDaysRaw !== undefined ? parseInt(appliedDaysRaw, 10) : null;
+
+// Whole-string match, not a bare parseInt: parseInt('1.5', 10) is 1 and
+// parseInt('10days', 10) is 10 — both truncate a bad value into a plausible
+// one instead of rejecting it, which is the exact silent-wrong-answer shape
+// this file is being fixed to close. Exported so a value's actual numeric
+// effect can be asserted directly, rather than only "the CLI didn't error".
+const APPLIED_DAYS_RE = /^\d+$/;
+export function parseAppliedDaysOverride(raw) {
+  return raw !== undefined && APPLIED_DAYS_RE.test(raw) ? parseInt(raw, 10) : null;
+}
+const appliedDaysOverride = parseAppliedDaysOverride(appliedDaysRaw);
 
 // --- Cadence config ---
 export const DEFAULT_CADENCE = {
--- a/followup-cadence.mjs
+++ b/followup-cadence.mjs
@@ -18,6 +18,7 @@ import * as yaml from 'js-yaml';
 import { loadCanonicalStates, foldStatusInput } from './tracker-utils.mjs';
 import { resolveColumns, parseTrackerRow } from './tracker-parse.mjs';
 import { localToday } from './lib/local-today.mjs';
+import { flagValue, validateFlags } from './lib/cli-flags.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
 const APPS_FILE = existsSync(join(CAREER_OPS, 'data/applications.md'))
@@ -31,8 +32,11 @@ const PROFILE_FILE = process.env.CAREER_OPS_PROFILE || join(CAREER_OPS, 'config/
 const args = process.argv.slice(2);
 const summaryMode = args.includes('--summary');
 const overdueOnly = args.includes('--overdue-only');
-const appliedDaysIdx = args.indexOf('--applied-days');
-const appliedDaysOverride = appliedDaysIdx !== -1 ? parseInt(args[appliedDaysIdx + 1], 10) : null;
+// flagValue (not indexOf) so `--applied-days=10` is honored too — indexOf()
+// can't see the `=` form and used to silently discard it, falling back to the
+// default cadence for a value the caller did supply (#2401/#2402 defect class).
+const appliedDaysRaw = flagValue(args, '--applied-days');
+const appliedDaysOverride = appliedDaysRaw !== undefined ? parseInt(appliedDaysRaw, 10) : null;
 
 // --- Cadence config ---
 export const DEFAULT_CADENCE = {
@@ -923,6 +927,7 @@ function printSummary(result) {
 // ── CLI flags + help ────────────────────────────────────────────────
 
 const KNOWN_FLAGS = ['--summary', '--overdue-only', '--applied-days', '--help', '-h'];
+const VALUE_FLAGS = ['--applied-days'];
 
 const USAGE = `Usage:
   node followup-cadence.mjs                    # full JSON analysis to stdout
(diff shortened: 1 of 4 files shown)


AdilAzhari:
@coderabbitai Both findings addressed in 6086111:

1. `--applied-days` now validates the whole string (`/^\d+$/`) instead of a bare `parseInt`, so `1.5` and `10days` are rejected instead of silently truncating to `1`/`10`.
2. Extracted the validation into an exported `parseAppliedDaysOverride()` and added unit tests (in `followup-cadence.test.mjs`) that assert the parsed value's actual numeric effect through `resolveCadenceConfig` — not just that the CLI didn't error.

Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/followup-cadence-flags.test.mjs`; `followup-cadence.mjs` and the shared `cli-flags.mjs` are outside my area and I read them for context.

The test asserts behavior, not the absence of a crash. The unrecognized-flag strings match the shared `validateFlags` output exactly, the help-plus-bogus case pins the ordering (the unknown check runs before `--help`, so no usage prints), the missing-operand case exercises `requireOperand`, and the bad-value cases (`abc`, `-5`, `1.5`, `10days`) each land on the `requires a non-negative integer` guard, with `-5` reaching it because value-flag adjacency consumes the single dash rather than misreading it as an unknown flag. The `=`-form case is honest, it asserts acceptance only and documents that it does not check the analysis result.

Real sentinels throughout (status plus stderr regex), neighbour-faithful to `cli-flag-validation.test.mjs`, and it lives in its own file per the repo rule. Green on CI. Approving from my side.

santifer:
Welcome to career-ops, @AdilAzhari, and this is a model first contribution: you measured the failures on current main before filing #3196, then fixed them yourself the same day.

Three things worth naming, because they are the difference between closing a ticket and closing a bug class. You used `flagValue` instead of `indexOf`, so the `=` form stops being silently discarded. You rejected `1.5` and `10days` with a whole-string match instead of letting `parseInt` truncate them into plausible-looking values, which is the same silent-wrong-answer shape one layer deeper. And you put `validateFlags` inside the `isMain` guard with the reason written down: test-all dynamic-imports this module in-process, so validating at module top level would have false-positived on the test runner's own flags. That is a trap several sweeps have walked into.

`requireOperand: true` was the right call too. Merged, your first contribution. Glad to have you here. 🚀

AdilAzhari:
@all-contributors please add @AdilAzhari for code, test

PR #3204, opened 2026-08-22
fix(pdf): derive the workspace root at use time, not at import (#3162)

### LINKED ISSUE #3162: flaky (macOS CI): page-budget test fails with 'input escapes the tracker workspace' on a path inside its own sandbox

santifer:
Closes #3162.

**Root cause**, named by the instrumentation from #3184 on occurrence 6 (macOS run of #3195):

```
workspaceRoot=/private/var/folders/df/.../T/cops-batcheval-N5TfQ3
canonical=/Users/runner/work/career-ops/career-ops/output/page-budget-test-7eO0Km/three-pages.html
```

The input was never escaping anything. The guard was comparing a valid repo path against **another test's temp directory**.

`tests/batch-evaluate.test.mjs:23` sets `CAREER_OPS_TRACKER` for its own fixture and **does** restore it in a `finally`. But this module resolved `trackerPath` → `workspaceRoot` → the canonical root as module-level `const`s that read that variable, so an import landing inside that window froze the root to the temp dir for the lifetime of the process. A later `delete` cannot un-read an env var.

Every property of the flake follows: only under the full suite (it needs the sibling to have run first), intermittent (import ordering), macOS-shaped (the `/var` → `/private/var` symlink made the relative path spectacular), and untouched by #3172, which was faithfully canonicalizing a correct input against the wrong root. Same defect class as #3159.

**Fix**: the root is memoized on the env variable's own value, so it costs the same as the const while nothing changes and self-corrects the moment it does. Two accessors keep each call site's existing semantics: canonical for the containment guard (a symlinked ancestor must not read as an escape), lexical for the manifest helpers.

**Verification**
- New `tests/generate-pdf-workspace-root.test.mjs` poisons the env before import and restores it exactly as the sibling does, then asserts a repo path still resolves against the real workspace.
- Discrimination checked: the test **fails on main** (exit 1, `repo path resolved to ""`) and passes here (exit 0).
- Full `test-all.mjs --quick`: **5309 passed, 0 failed**.

Note for reviewers: my first attempt fixed only the containment guard and the new test caught that the manifest helpers were still reading the frozen const, which is why both accessors exist.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed workspace path validation after temporary environment changes, preventing valid repository paths from being incorrectly rejected.
  * Ensured workspace-related path checks consistently use the current workspace location.

* **Tests**
  * Added regression coverage for workspace resolution and environment restoration scenarios.
  * Improved test cleanup by waiting for asynchronous operations to complete.

### DIFF
--- a/generate-pdf.mjs
+++ b/generate-pdf.mjs
@@ -51,7 +51,33 @@ const PDF_PAGE_MARGIN = '0.6in';
 // /var -> /private/var) is compared like-for-like by assertInsideWorkspace.
 // When CAREER_OPS_TRACKER points at another workspace, that workspace—not the
 // installed script directory—is the safe boundary for input and output.
-const __workspaceRoot = realpathSync(workspaceRoot);
+//
+// Derived at USE time, not frozen at import (#3162, root-caused occurrence 6):
+// the root reads CAREER_OPS_TRACKER, and a sibling test that legitimately sets
+// that variable for its own fixture — and correctly restores it afterwards —
+// still poisoned this module for the whole process if the import happened to
+// land inside that window. A `const` cannot un-read an env var, so the guard
+// then compared perfectly valid repo paths against another test's temp dir and
+// refused to write, intermittently and only under the full suite. Memoized on
+// the variable's own value: same cost as the const while nothing changes, and
+// self-correcting the moment it does. Same defect class as #3159.
+let __rootCache = { key: null, root: null, canonical: null };
+function refreshRootCache() {
+  const key = process.env.CAREER_OPS_TRACKER || '';
+  if (__rootCache.key !== key) {
+    // Always re-derive: falling back to the import-time const when the variable
+    // is unset would hand back the very value the poisoned import froze.
+    const root = resolveWorkspaceRoot(resolveTrackerPath(__dirname));
+    __rootCache = { key, root, canonical: realpathSync(root) };
+  }
+  return __rootCache;
+}
+// Two accessors on purpose, so each call site keeps the exact semantics it had:
+// the containment guard compares canonical forms (a symlinked ancestor must not
+// read as an escape), while the manifest/output helpers work in the lexical form
+// the rest of the module and its callers use.
+function currentWorkspaceRoot() { return refreshRootCache().root; }
+function canonicalWorkspaceRoot() { return refreshRootCache().canonical; }
 
 /**
  * Assert that an already-resolved absolute path stays inside the tracker workspace,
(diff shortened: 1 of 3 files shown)


santifer:
Pushed the actual cure, and updating the framing of this PR: **the lazy root was defence in depth, not the fix.** This branch's own CI proved it by reproducing the failure, and my new regression test failed there too, for a reason my model of the bug said was impossible.

The real mechanism is a race, not a frozen constant. `tests/batch-evaluate.test.mjs` ended with a floating `run()`, so its import resolved immediately while the async fixture kept running alongside later sections of test-all, with `CAREER_OPS_TRACKER` still pointing at its temp directory. The `finally` restores it correctly, but only once that work finishes. During that window the environment genuinely says the workspace is the temp dir, so no amount of lazy re-derivation helps: the module reads the env and gets what the env says.

`await run()` at top level closes it: the import does not resolve until the environment is clean again.

Both commits stay. A module that freezes its containment root at import time is fragile regardless of who poisons the environment, and the regression test now guards the composition of the two.

Full suite on this branch: **5309 passed, 0 failed.**

PR #3217, opened 2026-08-23
fix(reconcile-pipeline): validate CLI flags via the shared helper

### LINKED ISSUE #3216: fix(reconcile-pipeline): a typo'd --dry-run silently falls through to a live write
## Summary

`reconcile-pipeline.mjs` is the same class of defect as #2919/#3196: it has zero unrecognized-flag validation, so a mistyped `--dry-run` (e.g. `--dryrun`) is silently ignored and the script proceeds to a **live run that overwrites `data/pipeline.md`** — the caller believed they were in dry-run mode and were not. Worse than the reporting-CLI family this pattern was first fixed in, since this script actually mutates a user data file (with a `.pre-reconcile.bak` safety copy, but still not what the caller asked for).

`--pipeline`/`--state` have the value-side defect too: they're read via a hand-rolled `argValue()` using `indexOf`, so `--pipeline=path.md` silently falls back to the default pipeline file instead of the one the caller specified (the #2401/#2402 defect class).

Measured on current `main`:

```
$ node reconcile-pipeline.mjs --dryrun
# typo silently ignored; if batch-state.tsv has completed entries, this WRITES
# pipeline.md for real, exit 0 — no indication the flag was never applied

$ node reconcile-pipeline.mjs --pipeline=custom-pipeline.md --dry-run
# indexOf('--pipeline') can't see the `=` form; PIPELINE_FILE silently falls
# back to data/pipeline.md instead of the caller's custom-pipeline.md
```

No open issue or PR currently touches this file.

## Proposed fix

Delegate to `validateFlags`/`flagValue` from `./lib/cli-flags.mjs`, same shape as `funnel-velocity.mjs` (#3085) and `followup-cadence.mjs` (#3196/#3199). `resolveInsideRepo`'s repo-boundary sandboxing is unrelated and untouched — only the flag-parsing layer changes.

I have this working locally and would like to open a PR.

AdilAzhari:
Closes #3216

## Summary

`reconcile-pipeline.mjs` had zero unrecognized-flag validation — a mistyped `--dry-run` (e.g. `--dryrun`) was silently ignored and the script proceeded to a **live run that overwrites `data/pipeline.md`**, with the caller believing they were in dry-run mode. `--pipeline`/`--state` had the value-side defect too: the hand-rolled `argValue()` used `indexOf`, so `--pipeline=path.md` silently fell back to the default file instead of the caller's own path.

Same shape already fixed in `funnel-velocity.mjs` (#3085) and `followup-cadence.mjs` (#3196/#3199).

## Reproduce (before this PR)

```
$ node reconcile-pipeline.mjs --dryrun
# typo silently ignored; if batch-state.tsv has completed entries, this WRITES
# pipeline.md for real, exit 0 — no indication the flag was never applied

$ node reconcile-pipeline.mjs --pipeline=custom-pipeline.md --dry-run
# indexOf('--pipeline') can't see the `=` form; PIPELINE_FILE silently falls
# back to data/pipeline.md instead of the caller's custom-pipeline.md
```

## Changes

- `reconcile-pipeline.mjs`: delegate to `validateFlags`/`flagValue` from `./lib/cli-flags.mjs`. `resolveInsideRepo`'s repo-boundary sandboxing (symlink-safe, rejects paths escaping the repo) is unrelated and untouched — only the flag-parsing layer changes.
- Added `tests/reconcile-pipeline-flags.test.mjs` (9 tests): help, unrecognized flags (including the `--dryrun` typo itself), missing operands for both value flags, the `=`-form fix, and the existing repo-escape guard. Every invocation passes `--dry-run` explicitly so the suite can never write to this repo's own tracker data regardless of what `batch/batch-state.tsv` contains on the machine running it.

## Test plan

- [x] `node test-all.mjs` — 5309 passed, 0 failed
- [x] `node --test tests/reconcile-pipeline-flags.test.mjs` — 9/9 passed
- [x] `node scripts/check-syntax.mjs` — all files pass
- [x] Manually reproduced both bugs above against `main`, confirmed each is now a clear exit 1 / correctly-honored value

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved command-line argument handling for pipeline and state paths.
  - Unknown, mistyped, or incomplete options now produce clear errors.
  - Added validation for repository boundaries and required option values.
  - Added support for equals-form pipeline arguments.

- **Documentation**
  - Help and usage information is now centralized for more consistent guidance.

- **Tests**
  - Added coverage for help output, invalid flags, missing values, validation, and exit-status behavior.

### DIFF
--- a/reconcile-pipeline.mjs
+++ b/reconcile-pipeline.mjs
@@ -25,20 +25,22 @@ import { readFileSync, writeFileSync, existsSync, readdirSync, copyFileSync, rea
 import { join, dirname, resolve, relative, isAbsolute } from 'path';
 import { fileURLToPath } from 'url';
 import { normalizeReportLink } from './tracker-links.mjs';
+import { flagValue, validateFlags } from './lib/cli-flags.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
-const DRY_RUN = process.argv.includes('--dry-run');
 
-if (process.argv.includes('-h') || process.argv.includes('--help')) {
-  console.log('Usage: node reconcile-pipeline.mjs [--dry-run] [--state <path>] [--pipeline <path>]');
-  console.log('  Moves batch-processed offers out of pipeline.md "Pendientes" into "Procesadas".');
-  process.exit(0);
-}
+const KNOWN_FLAGS = ['--dry-run', '--pipeline', '--state', '--help', '-h'];
+const VALUE_FLAGS = ['--pipeline', '--state'];
+const USAGE = `Usage: node reconcile-pipeline.mjs [--dry-run] [--state <path>] [--pipeline <path>]
+  Moves batch-processed offers out of pipeline.md "Pendientes" into "Procesadas".`;
 
-function argValue(flag) {
-  const i = process.argv.indexOf(flag);
-  return i >= 0 && i + 1 < process.argv.length ? process.argv[i + 1] : null;
-}
+// An unrecognized flag (typo'd --dry-run, say) used to be silently ignored
+// and fall through to a LIVE run that writes pipeline.md — the caller
+// believed they were in dry-run mode and were not. Fail fast instead.
+const args = process.argv.slice(2);
+validateFlags(args, KNOWN_FLAGS, USAGE, { valueFlags: VALUE_FLAGS, requireOperand: true });
+
+const DRY_RUN = args.includes('--dry-run');
 
 // Constrain user-supplied --state/--pipeline paths to the repository tree, so a
 // crafted path cannot read from or overwrite files outside the project.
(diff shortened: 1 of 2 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/reconcile-pipeline-flags.test.mjs`; `reconcile-pipeline.mjs` is outside my area and I read it for context.

Every case asserts an exit code plus a stderr or stdout regex, and the runner checks `result.error`/`result.signal` so a timeout cannot pass as success. The messages match source exactly: the unrecognized-flag and `Valid flags:` strings from the shared `validateFlags`, the `requires a value` branch from `requireOperand`, and the `path must stay inside the repository` guard from `resolveInsideRepo`. It avoids the #3122 trap, the `--help` case asserts `/Usage:/` and `/node reconcile-pipeline\.mjs/`, both of which the real `USAGE` prints, and the ordering guard is there: `--help --bogus` exits 1 with no usage. Hermetic (every invocation carries `--dry-run`), neighbour-identical to `cli-flag-validation.test.mjs`.

PR #3229, opened 2026-08-23
feat(providers): add ITviec provider - first Vietnamese job source

vnk8071:
## Summary

- Adds `providers/itviec.mjs` — a zero-auth scanner provider for **ITviec** (itviec.com), Vietnam's largest IT job board and the project's first Vietnamese source.
- Adds `tests/providers/itviec.test.mjs` (30 assertions, auto-discovered, no registration) and rows in `docs/SUPPORTED_JOB_BOARDS.md` + `templates/portals.example.yml`.

## Why ITviec

Measured on 2026-08-23: itviec.com serves its listing pages **fully server-rendered over plain HTTPS** — no interstitial, no auth, no JS needed to see the cards. That makes it the right zero-auth entry point for the Vietnamese market; the other major boards (TopCV, VietnamWorks) are Vue/Cloudflare-rendered and need a browser, which the provider contract excludes.

`robots.txt` (fetched 2026-08-23) disallows only `/subscriptions/new`, a candidate-facing page this provider never requests — per the Source Indexing Policy.

## Implementation notes

- **Parsing anchors** follow the senjob.mjs philosophy (markup-based extraction rots): the per-card Stimulus attribute `data-search--job-selection-job-slug-value`, the posting URL shape `/it-jobs/{slug}`, and the title anchor inside `<h3 data-search--job-selection-target='jobTitle'>`. No layout classes matched.
- Emits `title`, `url`, `company` (first `/companies/` link with text — the first link wraps only the logo picture), `location` (map-pin div's `title` attribute) and `postedAt` (from the relative "Posted … ago" label; omitted when unparseable rather than fabricated).
- **Configuration**: explicit `provider: itviec`; optional `searchKeywords` / `searchLocation` (Ho Chi Minh / Hanoi / Da Nang) narrow via the path segments the board's own search form generates. Unrecognized cities are dropped, never guessed into the URL; slashes in keywords can't forge extra path segments.
- **Pagination**: `?page=N`, dedup by posting URL, stops when a page adds nothing new; `max_pages` default 10, hard cap 50; honours `ctx.maxPages` for the health probe.
- **Security**: host-pinned to `itviec.com` (HTTPS-only) + `redirect: 'error'` on every request, matching the SSRF convention in `ats-ssrf-hardening.test.mjs`.
- **Pacing**: 750ms between pages of the same board — HTTP 429 was *observed* under back-to-back full sweeps during live probing, so this is measured politeness, not a guess.
- **Silent-zero guard** (`assertParsedSomething`): if page 1 still contains posting links but parses to none, the provider throws so a markup change surfaces as a broken board instead of an empty market.

## Testing

- `node test-all.mjs --only providers/itviec` → 30/30 pass
- Full suite: `node test-all.mjs --quick` → **5340 passed, 0 failed** (2 pre-existing warnings, unrelated)
- Live verification against the real board:
  - bare board sweep → 200 jobs / 10 pages
  - `backend` + Hanoi → 50 jobs
  - `golang` → 37 jobs

## Summary by CodeRabbit

* **New Features**
  * Added ITviec as a supported job board for discovering Vietnamese job listings.
  * Added configurable keyword and location filters, pagination, request pacing, and duplicate prevention.
  * Added publication-date handling and safeguards for incomplete or unavailable listing results.
* **Documentation**
  * Documented ITviec support and added a sample configuration for enabling it.


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/providers/itviec.test.mjs`; `providers/itviec.mjs` is outside my area and I read it for context.

It is a faithful contract test. The detect and URL-building cases trace to source, including a prototype-name guard (`constructor`/`toString` fall through to a bare path via `Object.hasOwn`) and slash handling. The parse cases pin field values, not just counts: an entity decoded exactly once, an empty-logo link skipped for the text link, a location from the map-pin title attribute, and a cross-card leak guard. The silent-zero guard throws when a job link is present but nothing parsed. On fetch it pins pagination and dedup, the inter-page pacing, the `MAX_PAGES_CAP`, the entry-vs-ctx max-pages precedence, and an SSRF pin that every request carries `redirect: 'error'` and a User-Agent, which is the pair a refactor could quietly drop. Hermetic (mocked `ctx.sleep`/`fetchText`, no network), neighbour-style.

One small note, not blocking: `assertItviecUrl`'s off-host rejection is only covered indirectly, since the function is not exported. Green on CI. Approving from my side.

PR #3236, opened 2026-08-23
fix(story-provenance): the context matcher is ASCII-only, so a non-Latin CV verifies nothing (#3235)

### LINKED ISSUE #3235: story-provenance-check: the context matcher is ASCII-only, so a non-Latin CV can never verify anything (and an accented one verifies the wrong thing)
## Problem

`story-provenance-check.mjs:326` is the **last `[^a-z0-9]` strip left in the repo** — the shape #2393/#2429/#2569/#2666 replaced everywhere else with `normalizeTextKey`'s `\p{L}\p{M}\p{N}`:

```js
const snippet = body.slice(start, end).toLowerCase().replace(/[^a-z0-9\s]/g, ' ');
```

```
$ grep -rn "replace(/\[^a-z0-9" --include="*.mjs" --include="*.ts" .    # nothing
$ grep -n "a-z0-9" story-provenance-check.mjs
326:  const snippet = body.slice(start, end).toLowerCase().replace(/[^a-z0-9\s]/g, ' ');
```

`contextWords()` is the only context source this checker has, and **both** consumers read its result as a list — `hasScopedNumberMatch` does `claimWords.includes(w)`, `hasContextOverlap` does `words.some(...)`. An **empty** list makes both unconditionally false. So on any CV that is not written in ASCII, the `existing` and `supportedByResume` buckets are structurally unreachable and every claim the patterns extract falls to `derived-unverified`.

## Reproduction

Same person, same facts, same story — once in English, once in Russian:

```
########## ENGLISH ##########
Claims checked: 2
  ✅ existing (2)
     - [Scaling the platform team] "15 people" (scale-noun)
     - [Scaling the platform team] "40%" (percent)
  ⚠️ derived-unverified (0)

########## RUSSIAN ##########
Claims checked: 1
  ✅ existing (0)
  ⚠️ derived-unverified (1)
     - [Масштабирование платформенной команды] "40%" (percent)
```

`40%` sits verbatim in that `cv.md`, in the same sentence about infrastructure costs. It is reported as unverified anyway.

## Two failure modes, not one

```js
const strip = (s) => s.toLowerCase().replace(/[^a-z0-9\s]/g,' ').split(/\s+/).filter(w=>w.length>=4);

strip('Иван Петров')          // []
strip('παράδειγμα')           // []
strip('株式会社データ')          // []
strip('Größe')                // []   ← short accented Latin vanishes too
strip('naïve')                // []
strip('évaluation du risque') // ['valuation', 'risque']   ← NOT a deletion
```

The last line is the one I'd flag as the more dangerous half. The strip does not just drop `é`, it **re-cuts the word into a different real English word**. A French story about an `évaluation` can read as supported by an English finance CV that legitimately says `valuation` — an overlap on a token the story never contained.

## Why this bucket in particular

`derived-unverified` is not an internal label. It is what AGENTS.md's **Confirmation UX invariant** hands to the user to confirm or deny, and that section names the exact risk:

> a confirmed guess is worse than an honest unknown because it launders the guess into a "verified" fact

A checker that reports 100% unverified for a Russian or Japanese CV produces a confirmation list where every single entry is noise. That is how a user learns to confirm without reading — which is the failure this whole mechanism exists to prevent, arriving through the front door.

## Adjacent, and deliberately not in scope

`CLAIM_PATTERNS` is keyed on an English noun lexicon (`students|employees|people|staff|…`, `person|member`, `hours|hrs`), which is why the Russian run above extracted **one** claim instead of two — the `15 человек` claim is never checked at all. Internationalising the lexicon is a much larger change and a separate question; the `percent` pattern is already language-neutral, which is what makes the repro above a clean isolation of the context bug on its own.

One smaller thing I noticed while writing the tests: the percent pattern requires the sign to *follow* the number, so Turkish/German-style `%40` extracts nothing. Also separate.

## Fix

Adopt the alphabet the repo already settled on — `\p{L}\p{M}\p{N}`, with `normalizeTextKey`'s NFKC and U+0307 steps, since `\p{M}` would otherwise keep the combining dot that lowercasing a Turkish `İ` leaves behind. Mirrored rather than imported: `normalizeTextKey` builds a solid identity key where this needs word boundaries preserved, and importing `tracker-parse.mjs` would make a story-bank check fail hard on a missing `tracker-aliases.json` it has no other use for.

PR follows, with tests that fail on the current implementation plus guards pinning that the scoping checks (`15-person team` vs `15 years`) still fire — "keep every letter" would also be satisfied by a change that says yes to everything, and a false `existing` is a fabricated fact.

abankar1:
Fixes #3235.

`story-provenance-check.mjs:326` was the last `[^a-z0-9]` strip in the repo — the shape #2393/#2429/#2569/#2666 replaced everywhere else.

`contextWords()` is the checker's only context source, and both consumers read its result as a list (`claimWords.includes(w)`, `words.some(...)`), so an **empty** list makes both unconditionally false. On a Cyrillic/Greek/Hebrew/Arabic/CJK CV the strip empties every window, `existing` and `supportedByResume` become unreachable, and every extracted claim lands in `derived-unverified` — including a figure sitting verbatim in `cv.md`.

Accented Latin fails differently and worse: `évaluation` → `" valuation"`, a **different real English word**, so a French story can overlap an English finance CV on a token it never contained.

## Change

Keep letters, marks and digits of any script — `normalizeTextKey`'s alphabet, plus its NFKC and U+0307 steps (`\p{M}` is kept for Indic matras, and lowercasing a Turkish `İ` leaves a bare combining dot that `\p{M}` would otherwise preserve). Mirrored rather than imported: `normalizeTextKey` builds a solid identity key where this needs word boundaries, and importing `tracker-parse.mjs` would make a story-bank check fail hard on a missing `tracker-aliases.json` it never uses.

Two commits: the fix, then the tests.

## Before / after

Same person, same facts, same story:

| | before | after |
|---|---|---|
| English CV, `40%` in cv.md | `existing` | `existing` |
| Russian CV, `40%` in cv.md | **`derived-unverified`** | `existing` |
| Greek CV, `40%` in cv.md | **`derived-unverified`** | `existing` |
| Russian CV, `40` present but unrelated | `derived-unverified` | `derived-unverified` |
| French story `évaluation` vs English CV `valuation` | **`supportedByResume`** | `derived-unverified` |

## Verification

Three of the six new tests fail on the pre-fix implementation; the other three are guards that must hold on both sides. Checked by reverting only the function body:

```
--- new tests against the OLD implementation ---
✖ a Cyrillic CV verifies a claim its own prose supports
✖ a Greek CV verifies a claim its own prose supports
✔ a bare digit coincidence in a Cyrillic CV is NOT verified
✖ an accented word is no longer re-cut into a different real word
✔ the English behaviour the checker shipped with is unchanged
✔ a Turkish dotted capital folds to the same word as its plain i
ℹ pass 3  ℹ fail 3
```

The guards are the point of the pairing: "keep every letter" would also be satisfied by a change that says yes to everything, and a false `existing` is a fabricated fact rather than a noisy one.

`node story-provenance-check.mjs --self-test` → 24 passed, 0 failed.
`node test-all.mjs --quick` → 5309 passed, 0 failed (3 pre-existing warnings).

## Not in scope

`CLAIM_PATTERNS` is keyed on an English noun lexicon, so the Russian run extracts one claim instead of two — the `15 человек` claim is never checked at all. Internationalising the lexicon is a much larger change; the `percent` pattern is already language-neutral, which is what isolates the context bug cleanly here. Same for `%40` (Turkish/German percent-before), which the extractor does not see. Both noted in the issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved provenance classification for Cyrillic, Greek, accented, Turkish, and other non-ASCII characters.
  * Preserved matching evidence across international scripts while avoiding false matches from accented-word normalization.
  * Correctly handled non-ASCII numeric context and retained existing English-text behavior.

* **Tests**
  * Added coverage for international text, numeric matches, normalization, and existing classification behavior.

### DIFF
--- a/story-provenance-check.mjs
+++ b/story-provenance-check.mjs
@@ -352,20 +352,36 @@ function buildCvNumberContexts(cvText) {
  * @param {number} windowChars
  * @returns {string[]}
  */
-function contextWords(body, matchIndex, matchLength, windowChars = 90) {
-  const start = Math.max(0, matchIndex - windowChars);
-  const end = Math.min(body.length, matchIndex + matchLength + windowChars);
-  const snippet = body
-    .slice(start, end)
-    // NFKC before the strip so a full-width "４０％" keys like "40%", the same
+/**
+ * NFKC, lowercased, with the combining dot a lowercased Turkish `İ` leaves
+ * behind removed.
+ *
+ * Extracted so the two sides of every comparison below are folded IDENTICALLY.
+ * contextWords() folds the claim window; hasContextOverlap() searches cv.md.
+ * Folding only one of them is a bug that hides: a cv.md written "İstanbul"
+ * lowercases to `i` + U+0307 and stops matching a claim word the strip has
+ * already reduced to plain "istanbul".
+ *
+ * @param {string} text
+ * @returns {string}
+ */
+function foldForContext(text) {
+  return String(text ?? '')
+    // NFKC before the strip so a full-width "４０％" folds like "40%", the same
     // normalization order normalizeTextKey uses.
     .normalize('NFKC')
     .toLowerCase()
     // Lowercasing a Turkish dotted capital leaves a bare combining dot
-    // (U+0307) behind, and \p{M} below would keep it — so "İstanbul" and
-    // "Istanbul" would not compare equal. Same one-character strip, and the
-    // same reason, as normalizeTextKey's.
-    .replace(/\u0307/gu, '')
+    // (U+0307) behind, and \p{M} in the strip below keeps it — so "İstanbul"
+    // and "Istanbul" would not compare equal. Same one-character strip, and
+    // the same reason, as normalizeTextKey's.
+    .replace(/\u0307/gu, '');
+}
+
+function contextWords(body, matchIndex, matchLength, windowChars = 90) {
+  const start = Math.max(0, matchIndex - windowChars);
+  const end = Math.min(body.length, matchIndex + matchLength + windowChars);
+  const snippet = foldForContext(body.slice(start, end))
     .replace(/[^\p{L}\p{M}\p{N}\s]/gu, ' ');
   const words = snippet.split(/\s+/).filter(Boolean);
   // `\p{N}` not `\d`: a token of Devanagari or full-width digits is just as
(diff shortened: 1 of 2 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is the new `tests/story-provenance-non-ascii.test.mjs`; `story-provenance-check.mjs` is outside my area and I read it for context.

The sentinel is two-sided and stays within one script, which is what makes it trustworthy. The positive case is a Cyrillic claim with matching context: under the old `[^a-z0-9\s]` strip the context words were deleted entirely, so both the scoped-number match and the context-overlap check read false and the claim fell to `derivedUnverified`, and the new `\p{L}` strip keeps the letters so it verifies. The negative case is same-script on purpose, a bare digit coincidence in a Cyrillic CV with disjoint context must still come back unverified, which proves keeping the letters did not also loosen the scoping guard. The accented-Latin case pins the re-cut false positive shut from the other direction. The bucket names asserted match the real keys, and the verdict helper asserts exactly one bucket and one claim.

Hermetic, in-memory strings only, neighbour-style `node:test`. Green on CI. Approving from my side.

PR #3241, opened 2026-08-23
test(scan-history): make the writer guard an allowlist over discovered writers (#3240)

### LINKED ISSUE #3240: scan-history.tsv has four writers and two calendars: #3070 moved scan.mjs to the local day, the other three still stamp UTC
## Problem

`data/scan-history.tsv` is appended to by **four** scripts, all through the same exported helper (`appendToScanHistory` in `scan.mjs`). They do not agree about what day it is:

```
$ grep -n "localToday()\|new Date().toISOString()" scan.mjs scan-ats-full.mjs scan-hn.mjs scan-interamt.mjs

scan.mjs:2485          const date = localToday();                              # local day
scan-ats-full.mjs:687  const date = new Date().toISOString().slice(0, 10);     # UTC day
scan-hn.mjs:116        appendToScanHistory(newOffers, new Date().toISOString().slice(0, 10), 'added');
scan-interamt.mjs:260  const date = new Date().toISOString().slice(0, 10);     # UTC day
```

#3070 moved `scan.mjs` to the local day and its comment states both reasons:

> LOCAL day. This one value does two things that both care which day it is: it is the `today` `buildCooldownFilter` compares against, and it is the firstSeen date written into scan-history.tsv.

The three other writers of that same file were left behind. So west of Greenwich an evening run writes `2026-08-22` from `scan.mjs` and `2026-08-23` from `scan-ats-full.mjs` **for the same evening**; east of Greenwich the first hours of the morning go the other way — which is where `scan-interamt.mjs` runs by construction, it being a German public-sector portal.

## Why `first_seen` is not a cosmetic stamp

**1. The recheck window.** `shouldDedupScanHistoryRow` measures `firstSeen` against `localToday()`. A row stamped a day ahead reads as −1 days old, so its `recheck_after_days` counts from a day that has not happened yet.

**2. `detect-reposts` groups on it** — and its `--min-span` guard exists for precisely this, per its own header:

> `--min-span` raises it for anyone whose sweeps straddle midnight and land one company on two dates.

Mixed stamping makes that happen on **every evening run**, not only at literal midnight. And the second ingredient is already documented in the repo: `normalizeUrlForDedup`'s docstring describes the same Workday posting being reached with different path casing via the curated `portals.yml` entry versus the reverse-ATS dataset. Distinct URLs + distinct dates + span 1 is a complete cluster:

```js
// one posting, one evening, two scanners
const tsv = [
  'url\tfirst_seen\tportal\ttitle\tcompany\tstatus\tlocation',
  '…/en-US/Careers/job/Senior-Backend-Engineer_R123\t2026-08-22\tworkday\tSenior Backend Engineer\tAcme\tadded\tRemote',
  '…/en-US/careers/job/senior-backend-engineer_R123\t2026-08-23\tworkday\tSenior Backend Engineer\tAcme\tadded\tRemote',
].join('\n');

detectReposts(parseScanHistory(tsv), 90, 1, null)
```

```json
[{ "company": "Acme", "role": "Senior Backend Engineer",
   "repostCount": 2, "firstSeen": "2026-08-22", "lastSeen": "2026-08-23", "daysSpan": 1 }]
```

A fabricated repost signal is not a cosmetic bug — #18ff33d ("concurrent openings were reported as reposts") went in to stop exactly this class of false positive, and this reintroduces it through the date column instead of the URL column.

**3.** `scan-ats-full.mjs` also names its markdown digest `{date}.md` and headers it with the same value, so an evening run files the digest under tomorrow.

## Fix

Move the three remaining writers to `localToday()`. `new Date(iso).toISOString()` round-trips of a value that is *already* a calendar day stay UTC — that is arithmetic, the line `lib/local-today.mjs`'s docstring draws and #2765 drew before it.

PR follows, with a **derived-list** census (greps for `appendToScanHistory` callers rather than naming the four files, so a fifth scanner is covered the day it is written — the shape `tests/lock-rm-contention.test.mjs` needed before its family stopped being reintroduced), plus a test pinning the consequence: the two rows above are a cluster one day apart and are not a cluster on one date.

Disclosure: #3070 was mine, and this is the half of the census I missed — I checked the day-*gates* and not the writers of the file they gate on.

abankar1:
Fixes #3240. Completes the census #3070 started — which was mine, and I checked the day-*gates* without checking the writers of the file they gate on.

`data/scan-history.tsv` has four writers going through one exported helper. #3070 moved `scan.mjs` to `localToday()`; `scan-ats-full.mjs`, `scan-hn.mjs` and `scan-interamt.mjs` still stamp `new Date().toISOString().slice(0, 10)`. West of Greenwich an evening run writes two different dates for the same evening; east of Greenwich the first hours of the morning go the other way — which is where `scan-interamt.mjs` runs by construction.

## Why the stamp is a gate

- `shouldDedupScanHistoryRow` measures `firstSeen` against `localToday()`, so a row stamped a day ahead reads as −1 days old and its recheck window counts from a day that has not happened.
- `detect-reposts` groups on it, and its `--min-span` guard exists — its own header says so — "for anyone whose sweeps straddle midnight and land one company on two dates". Mixed stamping does that on **every** evening run. Two scanners reaching one Workday posting by two URL spellings (the case `normalizeUrlForDedup`'s docstring describes) then completes a cluster: distinct URLs, distinct dates, span 1. That is the false-positive class #18ff33d went in to stop, arriving through the date column instead of the URL column.
- `scan-ats-full.mjs` names its markdown digest `{date}.md` and headers it with the same value.

## Change

Three one-line migrations plus imports. Deliberately untouched: `new Date(iso).toISOString()` round-trips of a value that is already a calendar day — arithmetic, which stays UTC per `lib/local-today.mjs`'s docstring and #2765.

## Tests

Extends `tests/local-today-gates.test.mjs` (the file #3070 added) with two:

1. **A derived-list census.** It greps for `appendToScanHistory` *callers* rather than naming the four files, so a fifth scanner is covered the day it is written. That is the shape `tests/lock-rm-contention.test.mjs` needed before its own family stopped being reintroduced (#2777, #2984).
2. **The consequence.** The two rows above are a repost cluster one day apart, and are *not* a cluster on one date. Without this the census asserts a coding convention; with it, it asserts a bug.

## Verification

Reverting the three scanners and keeping the tests:

```
✖ every scan-history writer stamps the local day
  AssertionError: scan-history writers disagreeing about what day it is:
ℹ pass 9  ℹ fail 1
```

`node --test tests/local-today-gates.test.mjs` → 10 passed.
`node test-all.mjs --quick` → 5308 passed, 0 failed (3 pre-existing warnings).

## Summary by CodeRabbit

* **Bug Fixes**
  * Scan history now records dates based on the local calendar day instead of UTC.
  * Scan reports and newly added offers use consistent local dates, preventing incorrect day assignments around timezone boundaries.

* **Tests**
  * Added coverage to verify local-date handling and accurate repost grouping across scan records.


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/local-today-gates.test.mjs`; the three scan-history writers are outside my area and I read them for context.

The added source-census test is a good shape for this: rather than pin each writer by hand, it scans the writers that append to scan-history and asserts every one stamps `localToday()` and carries no `new Date().toISOString().slice(0,10)` offender, so a future `appendToScanHistory` caller is covered automatically. At head all three writers import and call `localToday()`, so the census passes with zero offenders. The `detectReposts` unit test pins two dates to one cluster and one date to none.

One honest caveat, not blocking: these are static source scans plus a pure in-memory unit test, not the frozen-clock TZ sentinels the neighbouring #3071 cases use. That is defensible here, since the stampers cannot be run hermetically, and the census is deterministic, but it is a different guarantee than a failing-clock gate, it proves the call is present rather than that the boundary behaves.

One merge note: #3244 also appends cases to the end of this file, so expect a textual conflict at the tail. Disjoint content, keep-both, trivial to resolve.

PR #3244, opened 2026-08-23
fix(dates): weekly-digest defaults to a week that has not started; rejection-latency suggests a blacklist row a day early (#3243)

### LINKED ISSUE #3243: weekly-digest defaults to a week that has not started, and rejection-latency suggests a blacklist row a day early — both still read the UTC day
## Problem

Two reporting scripts still resolve "now" as the UTC calendar day. Both use it as a **gate** — the class #2765/#2932/#3070 fixed elsewhere — not as a display stamp.

### 1. `weekly-digest.mjs` — the wrong week, and it looks like a quiet one

```js
export function computeDefaultRange(now = new Date()) {
  const d = new Date(Date.UTC(now.getUTCFullYear(), now.getUTCMonth(), now.getUTCDate()));
```

```
$ TZ=America/Los_Angeles node -e "…computeDefaultRange(new Date('2026-08-24T02:00:00Z'))"
instant (UTC) : 2026-08-24T02:00:00.000Z      # Sunday 2026-08-23, 19:00 PDT
week          : {"from":"2026-08-24","to":"2026-08-30"}
```

Sunday evening in the Americas, `weekly-digest` defaults to the week that **has not started**. Every `interview-prep/sessions/*.md` the user logged Mon–Sun falls outside `inRange()`, and the digest comes back empty — which reads as "a quiet week", not as "wrong window". Sunday evening is a fairly likely time to run a weekly digest.

### 2. `rejection-latency.mjs` — a blacklist suggestion a day early

```js
const today = opts.today instanceof Date && !Number.isNaN(opts.today.getTime())
  ? opts.today
  : new Date();
```

`daysBetween()` reduces both operands to their UTC date, so a bare clock read counts one extra day all evening west of Greenwich. The gate is `daysAll <= courtesyDays` (fires at 31+), so an interview that is exactly 30 local days old reads as 31 and **flags**. The output is not advisory phrasing — it is a ready-to-copy `data/blacklist.md` row:

```
| Acme Corp | 2026-08-18 | company | 31 days post-interview silence exceeds the 30-day courtesy threshold |
```

…for a courtesy window that has not elapsed, stamped via `isoDay()` with tomorrow's date. For a suggestion the user is invited to paste into a do-not-apply list, early-by-a-day is the direction that costs something.

## Fix

Read *which* day from the local calendar; keep the arithmetic anchored at UTC midnight. That is the split `lib/local-today.mjs`'s docstring draws, and both halves matter — `daysBetween` expects the UTC-midnight anchor, only the day it anchors on moves.

## Note on the existing weekly-digest assertions

Its three `computeDefaultRange` self-test assertions are built from `Z` instants, which name a different calendar day depending on where the suite runs — once the function reads the local day they pass or fail by timezone rather than by behaviour. The PR rebuilds them on local noon of a named date (that date everywhere) and adds the two instants where the reading moves a whole **week** rather than a day: late on the local Sunday, early on the local Monday. Verified across UTC, `America/Los_Angeles`, `Asia/Tokyo` and `Pacific/Kiritimati`.

Those two are no-ops in UTC by construction, so `tests/local-today-gates.test.mjs` pins the same property under an explicit `TZ` for CI. `inFrozenTz` there gains an optional instant: the file's existing `INSTANT` is a day-boundary pair *inside one ISO week*, which cannot see a week move.

PR follows.

abankar1:
Fixes #3243. Two reporting gates that still resolve "now" as the UTC day — the class #2765/#2932/#3070 closed elsewhere. Both *gate a decision*; neither is a display stamp.

## 1. `weekly-digest` — the wrong week, and it looks like a quiet one

`computeDefaultRange` read `now.getUTCDate()`, so at 19:00 Pacific on Sunday 2026-08-23 it returned `2026-08-24 … 2026-08-30` — the week that has not started. Every session logged Mon–Sun falls outside `inRange()` and the digest comes back empty, which reads as a quiet week rather than a wrong window. Sunday evening is a likely time to run a weekly digest.

## 2. `rejection-latency` — a blacklist suggestion a day early

The default `today` was a bare `new Date()`, and `daysBetween()` reduces both operands to their UTC date, so west of Greenwich it counts one extra day all evening. The gate is `daysAll <= courtesyDays` (fires at 31+), so an interview exactly 30 local days old reads as 31 and flags — producing a ready-to-copy `data/blacklist.md` row, stamped with tomorrow's date, for a courtesy window that has not elapsed. For something the user is invited to paste into a do-not-apply list, early-by-a-day is the direction that costs something.

## Change

Read *which* day from the local calendar; keep the arithmetic anchored at UTC midnight. `daysBetween` expects that anchor and it is preserved — only the day it anchors on moves. `--today` and an injected `opts.today` are untouched.

Three commits: each fix, then the tests.

## The weekly-digest self-test needed rebuilding

Its three existing `computeDefaultRange` assertions are built from `Z` instants, which name a different calendar day depending on where the suite runs — once the function reads the local day they pass or fail by timezone rather than by behaviour. Rebuilt on local noon of a named date (that date everywhere), plus the two instants where the reading moves a whole **week** rather than a day: late on the local Sunday, early on the local Monday.

```
UTC                  55 passed, 0 failed
America/Los_Angeles  55 passed, 0 failed
Asia/Tokyo           55 passed, 0 failed
Pacific/Kiritimati   55 passed, 0 failed
```

…and against the pre-fix implementation:

```
UTC                  55 passed, 0 failed          # no-op in UTC by construction
America/Los_Angeles  FAIL: …own day (Sunday 23:30 local)
Asia/Tokyo           FAIL: …own day (Monday 00:30 local)
Pacific/Kiritimati   FAIL: Monday itself as the range start · FAIL: …own day (Monday 00:30 local)
```

Because those are no-ops in UTC, `tests/local-today-gates.test.mjs` pins the same property under an explicit `TZ` for CI. `inFrozenTz` gains an optional instant — the file's `INSTANT` is a day-boundary pair *inside one ISO week*, which cannot see a week move, so `WEEK_INSTANT` is a Monday 01:30 UTC that is still the Sunday before in New York.

`rejection-latency` is pinned on **both** sides of its own boundary: an interview exactly `courtesyDays` old on the local day must not flag, one 32 days old must, and its blacklist suggestion must carry the local date. Without the second, "never flag at all" would satisfy the first.

## Verification

Reverting both fixes and keeping the tests:

```
✖ weekly-digest: "this week" is the week the caller is actually in
✖ weekly-digest: an explicit `now` is resolved locally too
✖ rejection-latency: the courtesy threshold is measured from the local day
✖ rejection-latency still flags once the local window HAS elapsed
ℹ pass 9  ℹ fail 4
```

`node --test tests/local-today-gates.test.mjs` → 13 passed.
`node test-all.mjs --quick` → 5308 passed, 0 failed (3 pre-existing warnings).

## Summary by CodeRabbit

* **Bug Fixes**
  * Date-sensitive evaluations now use the local calendar date by default.
  * Weekly digest ranges correctly reflect local Sunday/Monday boundaries.
  * Rejection-latency thresholds and blacklist dates now remain accurate across time zones.
* **Tests**
  * Added coverage for local-time weekly calculations and rejection-latency behavior before and after expiry.

### DIFF
--- a/rejection-latency.mjs
+++ b/rejection-latency.mjs
@@ -54,6 +54,7 @@ import { parseActiveInterviews } from './process-quality.mjs';
 import { resolveColumns, parseTrackerRow } from './tracker-parse.mjs';
 import { roleFuzzyMatch } from './role-matcher.mjs';
 import { flagValue, hasFlag, validateFlags } from './lib/cli-flags.mjs';
+import { localToday } from './lib/local-today.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
 const DEFAULT_ACTIVE_INTERVIEWS_PATH = existsSync(join(CAREER_OPS, 'data/active-interviews.md'))
@@ -221,9 +222,16 @@ export function buildBlacklistSuggestion(company, todayStr, reason) {
  * @returns {{ flags: object[], warnings: string[], companiesChecked: number }}
  */
 export function computeRejectionLatency(interviewRows, trackerByCompany, opts = {}) {
+  // The default is the LOCAL calendar day at UTC midnight, not `new Date()`.
+  // daysBetween() reduces both operands to their UTC date, so a bare clock read
+  // west of Greenwich counts one extra day all evening: a company crosses the
+  // courtesy threshold a day early and gets a ready-to-copy blacklist row —
+  // stamped, via isoDay(), with tomorrow's date. The UTC-midnight anchor is
+  // what daysBetween expects and is deliberately preserved; only WHICH day it
+  // anchors on moves (#2765 drew the same line).
   const today = opts.today instanceof Date && !Number.isNaN(opts.today.getTime())
     ? opts.today
-    : new Date();
+    : parseDate(localToday());
   const courtesyDays = Number.isFinite(opts.courtesyDays) && opts.courtesyDays > 0
     ? opts.courtesyDays
     : DEFAULT_COURTESY_DAYS;
--- a/tests/local-today-gates.test.mjs
+++ b/tests/local-today-gates.test.mjs
@@ -53,11 +53,15 @@ const NY_DAY = '2026-08-17';
  * `Date` is replaced before the module under test is imported, and the default
  * parameters being tested read the clock at CALL time, so the freeze reaches
  * them.
+ *
+ * `instant` defaults to INSTANT, the day-boundary pair every gate above needs.
+ * A caller pinning a WEEK boundary passes its own — a day apart is not enough
+ * to move an ISO week.
  */
-function inFrozenTz(tz, expr) {
+function inFrozenTz(tz, expr, instant = INSTANT) {
   const preamble =
     `const RealDate = Date;` +
-    `const FROZEN = new RealDate('${INSTANT}');` +
+    `const FROZEN = new RealDate('${instant}');` +
     `globalThis.Date = class extends RealDate {` +
     `  constructor(...a) { if (a.length === 0) { super(FROZEN.getTime()); } else { super(...a); } }` +
     `  static now() { return FROZEN.getTime(); }` +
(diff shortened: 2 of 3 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/local-today-gates.test.mjs`; `weekly-digest.mjs` and `rejection-latency.mjs` are outside my area and I read them for context.

The new cases are proper failing-clock gates in the #3071 style. The weekly-digest case pins `WEEK_INSTANT` at `2026-08-17T01:30:00Z`, Monday in UTC and Sunday in New York, so under the old `getUTCDate()` it resolved to the wrong ISO week and the `{from:'2026-08-10',to:'2026-08-16'}` assertion fails, passing only once the day is read locally. The rejection-latency pair is two-sided: an interview at exactly 30 local days is skipped while 31 UTC days would flag it, and the companion case at 32 days is flagged so "never flag" cannot pass either. All the helpers exist and are exported at head, and the change to `inFrozenTz` is a defaulted `instant` parameter so the day-boundary cases keep the old instant while the week cases pass the new one.

One note for whoever lands second: #3241 also appends cases to the tail of this same file, so the two will textually conflict at the end of the file. The content is disjoint and it is a keep-both resolution, not a semantic clash.

PR #3246, opened 2026-08-23
fix(web): a failed atomic write leaves a full copy of cv.md behind, unignored (#3245)

### LINKED ISSUE #3245: web: a failed atomic write leaves a full copy of cv.md behind, and that copy is not gitignored
## Problem

`web/src/lib/core/safe-write.ts` is, per its own header, **"THE one place every user-layer write goes through"** — `cv.md`, `config/profile.yml`, `portals.yml`, `data/follow-ups.md`, via seven route handlers.

```ts
export function atomicWrite(file: string, content: string): void {
  fs.mkdirSync(path.dirname(file), { recursive: true });
  const tmp = `${file}.tmp-${process.pid}-${randomUUID()}`;
  fs.writeFileSync(tmp, content, "utf8");
  fs.renameSync(tmp, file);          // ← no try, no cleanup, no retry
}
```

Compare the core's `writeFileAtomic` (`tracker-utils.mjs:595`), which this is a reimplementation of:

```js
try {
  writeFileSync(tmpPath, content);
  renameSyncWithRetry(tmpPath, path);
} catch (err) {
  rmSync(tmpPath, { force: true });
  throw err;
}
```

Three properties were lost. Two of them compound into a PII leak.

### 1. No cleanup on failure

The temp file holds a **byte-for-byte copy** of what is about to become the user's CV. That is fine while the rename succeeds and it lives for microseconds. A rename that *throws* is where it stops being fine — and it does throw: Windows refuses a rename whose destination is open by anyone else at that instant and answers EPERM/EACCES/EBUSY (the class #3006 fixed and #3046 catalogued across five more files, with the web left out), plus ENOSPC and read-only mounts. Nothing else ever removes the file, so it is one leaked copy per failure, indefinitely.

### 2. That leaked copy is not gitignored

`.gitignore` already carries `*.bak*`, and its comment names this very file and this very reason:

> `{file}.bak-{ISO timestamp}` (web/src/lib/core/safe-write.ts). A plain `*.bak` glob misses the timestamped form, which for cv.md or portals.yml leaves a file holding the same PII as the original sitting untracked in the repo root.

The sibling from the same writer was never added:

```
$ git check-ignore -q <path> && echo IGNORED || echo NOT IGNORED

cv.md                                              IGNORED
cv.md.bak-2026-08-05T16-55-08-641Z                 IGNORED
cv.md.tmp-4821-<uuid>                              NOT IGNORED     ←
config/profile.yml.tmp-4821-<uuid>                 NOT IGNORED     ←
portals.yml.tmp-4821-<uuid>                        NOT IGNORED     ←
```

`cv.md` is gitignored precisely because it is the user's CV. A leaked temp file is that same CV under a second name, one `git add .` from being committed — and career-ops is distributed by **forking**, so the repo a user runs this in is usually one they push. `tests/user-layer-gitignored.test.mjs` exists specifically to make user-layer content unstageable; it probes the `.bak-` shapes and not this one.

The core's own temp shape (`.{name}.{pid}.{ts}.{uuid}.tmp`) is not ignored either — but the core deletes it in a `catch`, so it never survives to matter.

### 3. No rename retry — real, and deliberately not fixed here

`renameSyncWithRetry` is the #3046 family's seam, and the web is the sixth site with the same write-tmp-then-rename-over-a-live-destination shape. I am **not** proposing a straight adoption: its backoff is a blocking `Atomics.wait`, which is right for a CLI holding a lock and wrong inside a Next request handler, where it would stall the event loop for every other request. Doing it properly wants an async `atomicWrite` (7 call sites, all already in async handlers). That is a design call for whoever owns `web/`, so the PR documents it in place rather than guessing.

## Fix (PR follows)

1. `try` / `rmSync(tmp, { force: true })` / rethrow, matching `writeFileAtomic`. Rethrowing matters as much as the cleanup — swallowing would let the route answer 200 for a write that never landed.
2. `*.tmp*` in `.gitignore`, alongside `*.bak*`, plus probes in `tests/user-layer-gitignored.test.mjs` for **both** temp shapes (a checkout predating the fix can still have one lying around). Defence in depth, not a substitute for (1).
3. Temp name aligned to the core's shape, so there is one ignorable shape instead of two.
4. A source-shape guard in `web/tests/lib/` — `safe-write.ts` is TypeScript and `node --test` has no TS runner, so this follows `core-writer-await.test.mjs`'s approach for `pipeline.ts`. It pins the cleanup, the rethrow, and the `.tmp` name that the ignore rule depends on, since neither file can see that drift alone.

abankar1:
Fixes #3245.

`atomicWrite` in `web/src/lib/core/safe-write.ts` — "THE one place every user-layer write goes through", per its own header — is a reimplementation of the core's `writeFileAtomic` that lost its `try/rmSync/rethrow`. The temp file it renames into place holds a byte-for-byte copy of what is about to become `cv.md` / `config/profile.yml` / `portals.yml`, so a rename that throws (Windows destination-handle contention — the #3006/#3046 class — ENOSPC, a read-only mount) leaves that copy behind. Nothing else removes it: one leaked copy of the user's CV per failure, indefinitely.

And it is not gitignored, while `cv.md` and its `.bak-` sibling both are:

```
cv.md                                IGNORED
cv.md.bak-2026-08-05T16-55-08-641Z   IGNORED
cv.md.tmp-4821-<uuid>                NOT IGNORED
config/profile.yml.tmp-4821-<uuid>   NOT IGNORED
portals.yml.tmp-4821-<uuid>          NOT IGNORED
```

`.gitignore`'s `*.bak*` comment already names this file and this exact reasoning — the temp sibling from the same writer was just never added. career-ops is distributed by forking, so the repo this runs in is usually one the user pushes.

## Change (three commits)

1. **`try` / `rmSync(tmp, { force: true })` / rethrow**, matching `writeFileAtomic`. Rethrowing matters as much as the cleanup: swallowing would let the route answer 200 for a write that never landed. Temp name aligned to the core's shape while here, so there is one ignorable shape instead of two.
2. **`*.tmp*` in `.gitignore`** beside `*.bak*`, plus probes in `tests/user-layer-gitignored.test.mjs` for **both** temp shapes — a checkout predating this can still have one lying around. Defence in depth, not a substitute for (1). No tracked file matches (`git ls-files | grep .tmp` → empty).
3. **A source-shape guard** in `web/tests/lib/`. `safe-write.ts` is TypeScript and `node --test` has no TS runner, so this follows `core-writer-await.test.mjs`'s approach for `pipeline.ts`; the extractor self-checks so a rename fails loudly rather than matching nothing.

## Deliberately NOT fixed: the missing rename retry

The web is the sixth site with the `#3046` write-tmp-then-rename-over-a-live-destination shape, and it still calls bare `renameSync`. I did **not** adopt `renameSyncWithRetry`, because a straight mirror would be wrong here: its backoff is a blocking `Atomics.wait`, correct for a CLI holding a lock and wrong inside a Next request handler, where it would stall the event loop for every other request. Doing it properly wants an async `atomicWrite` — 7 call sites, all already in async handlers.

That is a design call for whoever owns `web/`, so the PR documents it in place instead of guessing. Happy to send it as a follow-up in whichever direction you prefer. The cleanup above is worth having either way, and reduces how often the retry would be load-bearing.

## Verification

Reverting `safe-write.ts` and `.gitignore`, keeping the tests:

```
web/tests/lib/safe-write-temp-cleanup.test.mjs
  ✖ the rename is inside a try that removes the temp file
  ✖ a failed write is rethrown, never swallowed
  ℹ pass 2  ℹ fail 2

tests/user-layer-gitignored.test.mjs
  ❌ .cv.md.4821.….tmp is NOT git-ignored — a leaked atomic-write temp file holds the same PII as the original
  ❌ cv.md.tmp-4821-… is NOT git-ignored
  ❌ config/profile.yml.tmp-4821-… is NOT git-ignored
  ❌ portals.yml.tmp-4821-… is NOT git-ignored
```

`cd web && npm test` → 340 passed, 0 failed.
`node test-all.mjs --quick` → 5313 passed, 0 failed (3 pre-existing warnings).

## Summary

- `web/src/lib/core/safe-write.ts:15`: Uses the core-aligned `.tmp` filename format.
- `web/src/lib/core/safe-write.ts:18`: Removes temporary files after failed writes or renames.
- `web/src/lib/core/safe-write.ts:23`: Rethrows the original error, even if cleanup fails.
- `.gitignore:39`: Ignores `*.tmp*` files.
- `tests/user-layer-gitignored.test.mjs:178`: Checks current and legacy temporary filename patterns.
- `web/tests/lib/safe-write-temp-cleanup.test.mjs:1`: Guards cleanup, error propagation, filename, and ignore-rule behavior.

Users no longer retain sensitive temporary copies after a failed save. Rename retry remains out of scope because synchronous retry can block the Next.js event loop.


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/user-layer-gitignored.test.mjs`; the `web/` files and `.gitignore` are outside my area and I read them for context.

The added atomic-write temp-probe block is a real sentinel, not a rubber stamp. It runs `git check-ignore` on literal string probes and asserts they are ignored, and I confirmed both directions: without the new `*.tmp*` rule the probes report not-ignored and the test fails, with it they flip to ignored and pass. So a missing `.gitignore` entry reddens it, which is exactly the leak this guards against, an atomic-write temp file holding the same PII as the user-layer file it shadows. It covers both temp-name shapes, the new dot-prefixed form and the legacy `{file}.tmp-{pid}-{uuid}` a stale checkout could still hold. Hermetic (string probes via `git check-ignore --no-index`, no writes, no network), and it mirrors the neighbouring timestamped-backup probe pattern exactly.

Scott-Emberson (approved):
Re-reviewed at head fa7bf7cdaa (the atomic-write temp-cleanup work). The owned test `tests/user-layer-gitignored.test.mjs` holds up to the bar I care about: the new probe block resolves each candidate temp name through a real three-way `checkIgnore` verdict (git exit codes 0/1/128), passing only on `ignored` and failing on both `not-ignored` and `unanswerable`. So it's pinning actual gitignore state, not a throw-free run. Drop the new `*.tmp*` rule and every probe flips to `not-ignored` and the test reddens, which is exactly the guarantee it should carry.

One thing I like: the probe names are deliberately distinct from `cv.md` / `profile.yml` themselves, so no user-layer filename rule masks them, and the unanswerable branch fails closed. The "does safe-write still emit these shapes" coupling lives in the companion `web/tests/lib/safe-write-temp-cleanup.test.mjs`, which is the right split.

Owned test verified sound, CI green. Approving from the tests/ ownership side.

Scott-Emberson (commented):
Re-checked at head c681864b2b (moved past the earlier approval on a large repo-move merge). The owned test tests/user-layer-gitignored.test.mjs carries forward cleanly: the merge diff on it is purely additive, a new atomic-write-temp-probe block, and it does not revert the derived-index probes (#3469) or the symlink-crossing handling (#3598) that landed after my earlier approval. The new probes use the same git check-ignore exit-code contract as the rest of the file (0 ignored passes, 1 not-ignored fails, 128 fails closed), so a leaked atomic-write temp holding the same PII as the original reddens rather than slipping through.

Framing this correctly under the area rule: the actual fix here is in web/src/lib/core/safe-write.ts, which is outside our owned paths, so this is a test-axis review, not a gate. The owned test addition is sound and the leak guard is genuinely pinned, but the merge decision and any routing on the web change are the maintainer's, not something my approval gates. Nothing blocking from the tests/ side.

Scott-Emberson (commented):
Test-axis only. The substance is in `web/src/lib/core/safe-write.ts` and `.gitignore`, which is the maintainer's routing call; I am reviewing the owned test `tests/user-layer-gitignored.test.mjs`, and since this guards user PII (a leaked atomic-write copy of cv.md) I verified it strictly.

It is sound and not a tautology. I ran the mutation directly: built a repo with the PR's exact `.gitignore` and ran `git check-ignore` on all five temp probes, and all five are ignored; removing only the `*.tmp*` line flips all five to tracked, so the pass depends solely on the rule under test. The probes are real filename shapes from both writers, the new dot-prefixed `.{name}.{pid}.{ts}.{uuid}.tmp` and the legacy `{file}.tmp-{pid}-{uuid}`, and neither the exact-name rules nor `*.bak*` catch them (confirmed by the tracked result), so the `*.tmp*` rule genuinely does the work. The check yields three verdicts (ignored / not-ignored / error) and fails on not-ignored with a PII-leak message, so a regression reddens rather than silently passing.

santifer (approved):

santifer:
Merged. Thank you, @abankar1: **a failed atomic write in the web app no longer leaves a full copy of cv.md behind**. The temp file is removed when the write or the rename fails, the original error still surfaces, and `.gitignore` now covers those temp names, so a leftover from a killed process cannot slip into a fork's commit. The guard test that keeps the rename inside that try, with the Windows EPERM/EBUSY case written out in its header, makes the reason easy to keep.

That is merge 95 from you here. Sorry this waited as long as it did after the review.

PR #3043, opened 2026-08-18
Web UI: close CLI/web feature gap (contacto/deep/cover/email/outcome/training) + market-mode fix + contact persistence

code-harmony0:
Closes #3045.

## Summary

The web UI had a full workflow for `evaluate`/`pdf`/`apply`/`scan`/`pipeline`, but six CLI modes had no web equivalent: `contacto`, `deep`, `cover`, `email`, `outcome`, `training`. This adds all six as report-page actions, plus a few real bugs found and fixed while building them.

## New: 6 report-page actions

- **Find contacts** (`contacto`) — one primary target + ready-to-send message, brief alternates as one-liners (matches the mode's own "select primary target" design rather than drafting every persona in full)
- **Company research** (`deep`) — actually researches and answers the 6 axes, rather than just emitting a prompt for another tool
- **Cover letter** / **Application email** (`cover`/`email`) — draft-only, headless variants that skip the CLI's interactive confirmation checkpoints
- **Record outcome** (`outcome`) — wraps the real `outcome.mjs` script directly (deterministic tracker mutation, not an agent job)
- **Should I take a course or cert?** (`training`) — collapsed secondary input on the Pipeline page

The four generative actions (contacto/deep/cover/email) are consolidated behind one **AI actions** menu instead of four separate always-visible buttons, each showing running/done state individually.

## Bug fixes found along the way

- **`language.modes_dir` was write-only.** Settings saved the selected market mode (e.g. India → `modes/hi`) to `profile.yml`, but the web's evaluate flow always hardcoded `modes/oferta.md` regardless. A web-triggered evaluation silently ignored the configured market. Now resolves the correct market file (`angebot.md`, `fursah.md`, `naukri.md`, ...) per `language.modes_dir`, and injects the `language.output` vs `modes_dir` composition rule from AGENTS.md into every prompt (the CLI picks this up by reading AGENTS.md interactively; a one-shot headless prompt needs it stated explicitly).
- **Outreach messages read as AI-generated.** The web prompt asked the agent to draft a full message for every contact type found; `modes/contacto.md`'s own design is one primary + brief alternates. Tightened the web prompt to match, and added explicit rules to the mode file itself: no "Saw the [role] at [company]" opener (the most recognizable mass-outreach tell), no resume-bullet phrasing mid-sentence, end on a genuine question instead of a flat statement, vary sentence rhythm.
[...]

PR #3254, opened 2026-08-23
fix(web): stop evaluation tables collapsing to one word per line

code-harmony0:
Second one from the list on #3043. Bug fix, so no issue per CONTRIBUTING.md.

`report-view.tsx` is `max-w-3xl` and passes no `components` prop to any of its six `ReactMarkdown` calls, so tables render unwrapped. That is fine until you hit the STAR+R grid in an A-F report, which is eight columns of prose. At 768px every cell breaks down to about one word per line. "Re-architect" hyphenates in the middle of the word and "39% (53.9k to 33k lines)" runs over six lines.

Two changes here. The page container goes to `max-w-4xl`, and tables get their own horizontal scroll container so a wide one can overflow without taking the page with it.

The part worth reviewing is where the minimum width lives. I first put it on the table, which fixes the eight-column grid and immediately breaks everything else, because the same report has `Field | Value` and `Signal | Status` tables and they inherit that minimum and scroll for nothing. Putting the floor on the cells instead lets it scale with the column count by itself. Measured on report 037 at 1440px, container width in brackets:

| columns | table | container | scrolls |
|--------:|------:|----------:|---------|
| 8 | 1024px | 806px | yes |
| 5 | 812px | 814px | no |
| 3 | 846px | 848px | no |
| 3 | 812px | 814px | no |
| 2 | 846px | 848px | no |
| 2 | 812px | 814px | no |

So only the grid that needs the room takes it, and the small tables are left alone.

I stopped at `4xl` instead of going wider on purpose. The page carries the report's own prose as well, and widening enough to fit eight prose columns would push that past a comfortable measure. That just swaps a table problem for a paragraph problem.

One cost I should be upfront about: the floor is uniform, so a narrow leading column like `#` ends up wider than it needs to be. That is what you pay for a rule you can state in one line, and it only shows on the tables that scroll anyway. Happy to make it smarter if you would rather.

## On testing

There is no unit test here, and I do not think a fake one would help. Whether a Tailwind class makes a table readable is not something an assertion can answer, and a test that greps the source for a class name would pass on a comment, which is the trap the header of `run-prompts.test.mjs` already warns about.

What is mechanically checkable is that nothing escapes horizontally, so I checked that at 390px: `document.documentElement.scrollWidth` equals `clientWidth`. The column measurements in the table above came from the same run.

Suites: web 336 pass / 0 fail, `tsc --noEmit` clean, `next build` clean, `node test-all.mjs` 5310 pass / 0 fail.

Happy to attach before/after screenshots if useful, just say where you want them.

## Summary by CodeRabbit

* **Style**
  * Improved readability of wide report tables by ensuring cells in tables with six or more columns have sufficient minimum width.

### DIFF
--- a/web/src/app/globals.css
+++ b/web/src/app/globals.css
@@ -213,6 +213,32 @@ html {
      still prevents overflow from long unbreakable tokens (URLs). */
   overflow-wrap: break-word;
 }
+/* A minimum column width, but only on the wide evaluation grids.
+
+   #3160 removed the cause of the mid-syllable collapse (`word-break` made every
+   character a soft-wrap opportunity) and gave the table its own scroller. What
+   it does not give a wide table is a floor: `max-width: 100%` still squeezes
+   the columns back down to share the reading column, so the eight-column STAR+R
+   grid settles at roughly 1.5 words per line. Readable, but only just, and rows
+   run ~239px tall because every cell wraps that hard.
+
+   The floor is on the CELLS, not the table: a table-level `min-width` has to be
+   sized for the worst case, and then the "Field | Value" tables inherit it and
+   scroll for nothing. A per-cell minimum scales with the column count by itself.
+
+   The `:has()` scope is what keeps it off the narrow tables. An unscoped floor
+   pads short columns too -- a `#` index goes from 9px to the full floor -- which
+   pushed three- and five-column tables into scrolling that they did not need.
+   Restricting it to tables that actually have a sixth column leaves every
+   smaller table byte-identical to today. Where `:has()` is unsupported the rule
+   is simply dropped and the table behaves as it does now.
+
+   Measured on report 047, first STAR+R row, at 390px and 1440px:
+     floor    grid scrollWidth   row height   avg words/line
+     none            965px          239px          1.53
+     200px          1600px          166px          2.48
+   Document scrollWidth stays equal to clientWidth at 390px throughout. */
+.report-prose table:has(tr > *:nth-child(6)) :is(td, th) { min-width: 200px; }
 .report-prose tbody tr:hover { background: color-mix(in srgb, var(--fg) 3%, transparent); }
 .report-prose img { border-radius: 0.5rem; }
 


santifer (changes requested):
Heads-up before you spend more time on this: **#3160 just landed and fixes the same symptom from the CSS side**, so this PR needs to shrink rather than be merged as-is. Your diagnosis was right and I want to keep part of what's here — details below so you can decide what's still worth carrying.

### What landed instead, and why it went first

#3160 changed two things in `web/src/app/globals.css`:

```diff
-.report-prose td { word-break: break-word; }
+.report-prose td { overflow-wrap: break-word; }
```

plus a `display:block; width:max-content; max-width:100%; overflow-x:auto` rule on `.report-prose table`.

That first line is the **cause** of what you both saw: `word-break: break-word` makes every character a soft-wrap opportunity during min-content sizing, so the columns collapse to their narrowest and only then do words split mid-syllable. Neither your PR nor #3160's table rule removes that on its own — #3160 does.

The second reason it went first is reach. I checked where report markdown actually renders, and it's **four** components, not two: `report-view.tsx`, `jobs/[id]/page.tsx`, `assistant-console.tsx` and `cv-editor.tsx` all carry `.report-prose`. A `globals.css` fix covers all four; `markdownComponents` covers the two you wired it into, so a table in an assistant reply or a CV preview would still have collapsed.

### The direct collision

Your `table` override wraps in `<div className="... overflow-x-auto ...">`, and #3160 now makes `.report-prose table` itself `overflow-x: auto`. Landing both nests two horizontal scrollers on the same content — an outer one that scrolls and an inner one that never quite triggers. That's worse than either alone, which is why this couldn't just go in behind it.

### What's genuinely yours and worth keeping

Two things in here are better thinking than anything in #3160, and I'd take them:

1. **The floor on cells, not the table.** Your reasoning is right and I hadn't seen it put that cleanly — a table-level `min-width` has to be sized for the worst case, and then the two-column `Field | Value` tables inherit it and scroll for nothing, whereas a per-cell minimum scales with column count by itself. If `min-w-32` on `td`/`th` still improves the eight-column STAR+R grid **on top of** `width: max-content`, that's a real addition. Worth measuring rather than assuming: `max-content` already sizes columns to their content, so the floor may or may not still bite.

2. **One definition instead of two.** "The two places that used to hold this independently could drift, so they no longer hold it independently" is the right instinct, and it's the same reasoning as `report-sections.mjs`. It matters less now that the fix is in CSS, but if the per-cell floor survives, it should live in one place.

### One thing to pull out regardless

This PR also widens the reading column in three places:

```diff
-<div className="mx-auto max-w-3xl px-6 py-8">
+<div className="mx-auto max-w-4xl px-6 py-8">
```

That's a 25% change to the report's reading measure, and it's a design decision rather than a bug fix — it belongs in its own PR where it can be judged on its own. Your own comment argues against it, in fact: *"Widening far enough to fit eight prose columns would push the report's own prose past a comfortable measure, trading one readability problem for another."* I agree with your comment; the diff does the thing the comment warns about.

### What I'd like from you

Rebase onto current main and reduce this to whatever still measurably helps after #3160 — most likely just the per-cell floor, and only if it changes the rendering. If it turns out `max-content` already handles it, say so and I'll close this with the credit on the record; a PR that correctly identifies a bug someone else fixes first is still work that counted.

Open the `max-w-4xl` question separately if you want it, and I'll take it as a design call rather than a fix.

Issue #2582, opened 2026-08-07
feat: fetch-jd.mjs — fetch a JD via its ATS provider API instead of the full HTML page


darkpandawarrior:
### Problem

Batch and headless workers currently `WebFetch` the full JD page to read a job description. That is 10–30k tokens of nav and markup per job, and it is unreliable on SPA-rendered boards where the JD is not in the served HTML.

The irony is that `scan.mjs`'s own providers already know how to talk to those boards' per-job JSON endpoints — the JD text is available cheaply, we just don't use that path when fetching a single JD.

### Proposed solution

`fetch-jd.mjs`: hit the same Greenhouse / Ashby / Lever / Workday per-job JSON endpoints the providers already use, strip to clean text, print to stdout.

- **exits non-zero with empty stdout on a miss**, so the caller falls back to `WebFetch` exactly as today — a miss is visibly a miss, never a fabricated JD
- reuses `providers/_http.mjs`'s existing fetch wrapper (timeout, UA, error shape) rather than adding a second HTTP client
- no behaviour change at all on providers it doesn't cover

An 80–90% token reduction on the covered providers, which is the largest single line-item in the batch pipeline. Zero-token in the sense that matters: plain HTTP GETs against public per-job APIs, no LLM anywhere.

`htmlToText()` / `decodeEntities()` / `detectProvider()` are already pure exported functions; I'd add unit tests for them before the PR.

### Area

Scripts / providers

santifer:
Thanks for this, @darkpandawarrior.

fetch-jd.mjs pulls the job description from the per-listing JSON endpoint of Greenhouse/Ashby/Lever/Workday that the scan.mjs providers already know how to talk to, instead of WebFetching the full page: 10-30K tokens of nav and markup per listing, and unreliable on SPA boards where the JD isn't even in the served HTML. The detail that makes it safe and worth crediting: it exits non-zero with empty stdout when it doesn't cover the board, so the caller falls back to WebFetch exactly as today and a failure is visibly a failure, never an invented JD. It reuses the providers/_http.mjs wrapper (timeout, UA, error shape) rather than adding a second HTTP client, and estimates 80-90% fewer tokens on covered providers with no LLM in the loop.

Where it stands: triaged and in the queue. If you want to build it yourself, say so here and it is yours: a PR from the person who proposed it beats one from me.

PR #3257, opened 2026-08-23
fix(browser-extract): read Workday JDs from the CXS API, and fail loudly on an empty JD

### LINKED ISSUE #3256: bug(browser-extract): Workday JDs extract as empty text and exit 0 — the documented silent fallback never fires
### Code of Conduct

### Existing issues

### What happened?

`browser-extract.mjs --mode jd` returns an **empty `text`** for every Workday-hosted posting (`*.myworkdayjobs.com`), while exiting **0** with well-formed JSON.

That combination is the actual bug. A caller cannot tell "extraction failed" from "this posting has no content", so the documented fallback in `modes/pipeline.md` —

> Use its `text` as the JD. **Fall back silently** to `browser_navigate` + `browser_snapshot` if it errors or is missing.

— never fires. The empty JD is passed downstream and evaluated as if it were the posting.

Root cause: Workday hydrates the JD client-side into a virtualized DOM. `readDom()`'s `main, [role="main"], article` → `innerText` read runs before/around that and comes back with nothing.

Workday is not a niche host — it is one of the ATSes `scan.mjs` already covers, so postings from it land in `data/pipeline.md` routinely.

Verified on v1.28.0 (`10a569b1`).

### Steps to reproduce

```bash
node browser-extract.mjs "https://umusic.wd5.myworkdayjobs.com/umguk/job/Kings-Cross-London/AI-Product-Manager_UMG-27187" --max-chars 12000
node browser-extract.mjs "https://spgi.wd5.myworkdayjobs.com/spgi_careers/job/London-UK/Lead-Product-Manager-AIOPs_329276-2" --max-chars 12000
```

Both print `{"url":…,"title":…,"text":""}` (or near-empty) and exit 0.

### Expected behavior

Two separate things, and I think both are worth fixing:

**1. Workday JDs should extract.** Workday exposes a public per-job CXS JSON endpoint carrying the full description plus structured metadata:

```
https://{tenant}.{shard}.myworkdayjobs.com/wday/cxs/{tenant}/{site}/job/{job-path}
```

This is the same API family `providers/workday.mjs` already uses for board listings, and — conveniently — `liveness-api.mjs`'s `workday` provider **already derives exactly this per-job URL**, including its SSRF segment validation. So `resolveAtsApi()` can be reused as-is rather than growing a second copy of the tenant/shard/site parsing.

The payload's `jobPostingInfo` gives `title`, `jobDescription` (HTML), `location` + `additionalLocations`, `timeType`, `postedOn`, `jobReqId` — and `canApply`, which is a useful liveness signal on a posting that is still served but closed to applications. `jobReqId` in particular feeds the tracker's same-title disambiguation rule from `AGENTS.md`.

Verified against both URLs above: HTTP 200, full description, ~6.5k and ~7.9k chars of clean text.

**2. An empty jd extraction should be a non-zero exit, not a success.** Independent of Workday: whatever the host, a posting page that yields no main text is a render we missed (SPA shell, consent wall, bot interstitial), never a real JD. `browser-extract.mjs` should exit 1 with `{"error": …, "code": "empty_text"}` so the silent-fallback path the modes already document actually triggers.

This guard should apply to **jd mode only** — an empty `jobs` array in `listing` mode is a legitimate result (a company with no open roles).

### Notes

- Related to #2582 (`fetch-jd.mjs` — fetch a JD via its ATS provider API). This is narrower: the same idea applied inside the existing extractor, for the one provider where the scraped path is not merely expensive but silently *wrong*. If #2582 lands first, the Workday half here folds into it; the `empty_text` exit is orthogonal to both.
- I have a working patch against `main` (Workday CXS branch reusing `resolveAtsApi`, HTML→text that preserves block structure, the `empty_text` guard, plus 12 unit tests in `tests/browser-extract.test.mjs`). Both repro URLs now extract, full suite green. Happy to open a PR against this issue if the approach looks right.

### CLI tool

Claude Code

### OS and Node.js version

Windows 11 / Node 25.1.0

Dreamkeeper:
Fixes #3256

## The bug

`browser-extract.mjs --mode jd` returns an **empty `text`** for every Workday posting (`*.myworkdayjobs.com`) while exiting **0** with well-formed JSON. Workday hydrates the JD client-side into a virtualized DOM, so `readDom()`'s `main, [role="main"], article` → `innerText` read comes back with nothing.

The exit code is the part that actually hurts. A caller cannot tell "extraction failed" from "this posting has no content", so the fallback `modes/pipeline.md` documents —

> Use its `text` as the JD. **Fall back silently** to `browser_navigate` + `browser_snapshot` if it errors or is missing.

— never fires, and the empty JD goes downstream to be evaluated as if it were the posting.

## Two independent fixes

**1. Workday reads from the CXS API.** jd mode short-circuits `*.myworkdayjobs.com` to Workday's public per-job endpoint (`/wday/cxs/{tenant}/{site}/job/{path}`) — the same API family `providers/workday.mjs` already uses for boards.

The URL derivation is **not** reimplemented. `liveness-api.mjs`'s `workday` provider already builds exactly this URL, so `resolveAtsApi()` is reused, SSRF segment validation included (charset-validated, `..`-rejected path segments before the fixed host template). One copy, no drift.

Every inconclusive outcome — non-200, redirect, unparseable body, missing `jobPostingInfo`, timeout — returns `null` and falls through to Playwright unchanged, so a removed posting still yields the real "no longer available" page rather than a hard error.

The metadata header prepended to `text` carries what the rendered page shows and the description body does not: location(s), time type, posted, `jobReqId` (the tracker's same-title disambiguation key from `AGENTS.md`), and `canApply: false` as a liveness signal on a posting still served but closed to applications.

**2. An empty jd extraction exits non-zero.** Independent of Workday: whatever the host, a posting page yielding under 200 chars of main text is a render we missed (SPA shell, consent wall, bot interstitial), never a real JD. It now exits 1 with `{"error": …, "code": "empty_text", "url": …}`, so the silent-fallback path already documented in the modes actually triggers.

- A `--max-chars` below the floor is honored rather than made unsatisfiable — ask for 50 chars, get 50.
- **`listing` mode is deliberately untouched:** an empty `jobs` array is a legitimate result for a company with no open roles, so the same guard there would be wrong.

## Why a local `jdHtmlToText`

`providers/_html-to-text.mjs`'s `htmlToText` is tuned for scan payloads: it hard-caps at 4000 chars and collapses *all* whitespace including newlines. A JD read for evaluation wants the full body up to `--max-chars` with its paragraph and bullet breaks intact. `decodeEntities` **is** shared, so the two cannot drift on the thing that has actually drifted historically (#1555/#1639/#2623).

## Verification

Against two live tenants, where `main` returns empty:

```bash
node browser-extract.mjs "https://umusic.wd5.myworkdayjobs.com/umguk/job/Kings-Cross-London/AI-Product-Manager_UMG-27187" --max-chars 12000
# {"url":…,"title":"AI Product Manager","text":"Location: Kings Cross, London\nJob type: Full time\n…"}  6461 chars, exit 0

node browser-extract.mjs "https://spgi.wd5.myworkdayjobs.com/spgi_careers/job/London-UK/Lead-Product-Manager-AIOPs_329276-2" --max-chars 12000
# {"url":…,"title":"Lead Product Manager AIOPs","text":"Location: London, UK | Gurugram, Haryana | …\nReq ID: 329276\n…"}  7913 chars, exit 0
```

And the `empty_text` guard on a thin non-Workday page:

```bash
node browser-extract.mjs "https://example.com"
# {"error":"extracted 125 chars of JD text (minimum 200) — the page most likely renders its content client-side","code":"empty_text","url":"https://example.com/"}
# exit 1
```

12 new unit tests in `tests/browser-extract.test.mjs` cover the CXS derivation (optional locale segment, board-vs-posting URLs, `..` traversal rejection), the HTML→text pipeline, and the payload normalizer. `node test-all.mjs --only browser-extract` → 43 passed, 0 failed.

## Relation to #2582

#2582 (`fetch-jd.mjs`) proposes the general version of idea 1 across all ATS providers. This is narrower on purpose: the one provider where the scraped path is not merely expensive but silently *wrong*. If #2582 lands, the Workday half here folds into it; the `empty_text` exit is orthogonal to both.

## Notes

- Behaviour is unchanged for every non-Workday host that extracts a real JD, and for `listing` mode entirely.
- No new HTTP client — plain `fetch` with `redirect: 'error'`, `DEFAULT_USER_AGENT`, and an `AbortController` timeout.

---

## Summary by CodeRabbit

* **New Features**
  * Added Workday job-description extraction through its public endpoint.
  * Added fallback browser extraction when API responses are inconclusive.
  * Added validation for missing or unusually short job-description text.
  * Preserved existing listing output behavior.

* **Bug Fixes**
  * Improved handling of invalid URLs, timeouts, metadata, application status, and HTML content.

* **Documentation**
  * Updated extraction command documentation, including empty-description error behavior.

### DIFF
--- a/browser-extract.mjs
+++ b/browser-extract.mjs
@@ -25,9 +25,20 @@
  *                  anchors that look like individual postings, deduped. For scan
  *                  Level 1 (reading a company's open roles).
  *
+ * Workday (`*.myworkdayjobs.com`) is read through its public CXS JSON endpoint
+ * instead of the rendered page: Workday hydrates the JD into a virtualized DOM
+ * that readDom() below cannot see, so scraping it returned a well-formed result
+ * with an EMPTY `text` — indistinguishable, to a caller, from a posting that
+ * genuinely has no content. Same API family scan.mjs already uses for Workday
+ * boards (providers/workday.mjs); the per-job URL derivation is reused from
+ * liveness-api.mjs so the two cannot drift.
+ *
  * Output: compact JSON to stdout. Exit 0 on success; exit 1 on a hard error,
  * printing `{ "error": "...", "code": "..." }` (so a caller/mode can fall back
- * to the MCP path silently). Reuses liveness-browser.mjs's SSRF host guard and
+ * to the MCP path silently). An empty/near-empty jd-mode extraction is one of
+ * those hard errors (`code: "empty_text"`) rather than a successful-looking
+ * empty JD — the documented silent fallback only fires if the tool actually
+ * reports failure. Reuses liveness-browser.mjs's SSRF host guard and
  * realistic-UA context so it isn't instantly bot-walled.
  */
 
@@ -36,6 +47,9 @@ import { join, dirname } from 'path';
 import { fileURLToPath, pathToFileURL } from 'url';
 import * as yaml from 'js-yaml';
 import { LIVENESS_CONTEXT_OPTIONS, rejectPrivateOrInvalid } from './liveness-browser.mjs';
+import { resolveAtsApi } from './liveness-api.mjs';
+import { decodeEntities } from './providers/_html-entities.mjs';
+import { DEFAULT_USER_AGENT } from './user-agent.mjs';
 import { flagValue, hasFlag, validateFlags } from './lib/cli-flags.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
@@ -44,6 +58,16 @@ const DEFAULT_TIMEOUT_MS = 15_000;
 const HYDRATION_WAIT_MS = 2_000;
 const JD_TEXT_CAP = 12_000;     // plenty for a JD; a fraction of a full snapshot
 const DEFAULT_LISTING_MAX = 200;
+const WORKDAY_TIMEOUT_MS = 10_000;
+
+// Floor below which a jd-mode extraction is treated as failure, not content.
+// A posting page whose main text is a couple of sentences is a render we
+// missed (SPA shell, consent wall, bot interstitial), never a real JD; the
+// cost of being wrong is one silent fallback to the MCP path, whereas the
+// cost of NOT failing is an empty JD evaluated as if it were the posting.
+// Only jd mode gets this guard: an empty `listing` result is legitimate — a
+// company with no open roles.
+const MIN_JD_TEXT_CHARS = 200;
 
 // Anchor labels that are navigation chrome, not job postings. Kept small and
 // lowercase; matched against the trimmed label.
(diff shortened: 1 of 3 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/browser-extract.test.mjs`; `browser-extract.mjs` is outside my area and I read it for context.

The eight new assertions exercise three pure exported normalizers, each with a real sentinel: dropping the closed-posting liveness signal reddens the `canApply:false` case, breaking entity or bullet handling reddens `jdHtmlToText`, and the `../../evil` case guards the traversal reject. Hermetic pure functions, with the live fetch and emit paths correctly left uncovered rather than stubbed live. Green on CI. Approving.

PR #3261, opened 2026-08-24
feat(explorer): free banded title-vs-profile fit hint on discovery cards

### LINKED ISSUE #3260: Explorer (scan): free banded title-vs-profile fit hint on discovery cards
## Problem

Scan-mode discovery is free, but its output is a flat list. The only per-card signal today is `matched <chip>` (`discovery-card.tsx`), produced by `firstMatch()` in `lib/core/scan.ts`: a case-insensitive **substring** test of the current positive filter chips against the title. That has two real gaps:

1. **Multi-word chips break on word order and punctuation.** Chip `platform engineer` never matches the perfectly good title `Staff Engineer, Platform`, because the words are not contiguous.
2. **No cheap way to triage 150 free results into "which handful deserve a paid evaluation."** The honest scoring (A-F, blocks A-H) costs tokens per offer, so the decision of what to evaluate is currently made by eyeballing titles.

## Proposal

Attach a deterministic, zero-token **band** to each scan-mode offer, computed from data career-ops already has:

- Compare the posting **title** against the user's `config/profile.yml` target roles (`target_roles.primary[]` + `archetypes[].name`, read through the existing tested `profile-keywords.mjs` helper).
- Token-overlap ratio: the share of a target role's meaningful tokens (seniority words such as senior/staff/principal excluded, they say nothing about which role it is) that appear in the title. Best-matching role wins.
- Band the result: `strong` (>= 0.6), `related` (>= 0.34), otherwise `weak`.
- Render as a small passive chip next to the existing `matched` note, with a tooltip that says exactly what it is and is not: a free keyword-level estimate, not an evaluation - evaluating still gives the real A-F fit score.

Example: profile has `platform engineer`; titles `Senior Platform Engineer` (strong), `Engineer, Platform Infrastructure` (strong), `Data Engineer` (related), `Product Designer` (weak).

## House rules this respects

- **Recommendation layer only** (same contract as `jd-similarity.mjs`): it never filters, hides, reorders, or auto-adds anything. The reason it can be trusted next to the "NOT a spray-and-pray tool" philosophy is that it informs the evaluate decision without making it.
- **Zero tokens, offline**: pure string math on data already loaded server-side.
- **Does not touch the holistic evaluation**: the 1-5 judgment and A-F remain the only quality verdicts; the chip explicitly points to them.
- **AI-search offers unchanged**: they already carry `why` + `confidence` judgments; the band is added only on the deterministic scan paths where no judgment exists.

## Implementation sketch

- New `web/src/lib/title-fit.mjs`: plain `.mjs` like `profile-keywords.mjs` (Turbopack pins the root to `web/`, so core helpers must be mirrored, not imported) exporting a pure, self-tested `titleFit(title, targets)`.
- `DiscoveredOffer` gains an optional `fit?: { band, score }` field, so the stream grammar and AI offers are untouched.
- Wired in `lib/core/scan.ts` at both offer-construction sites (legacy stdout parse and `--json` parse), with `config/profile.yml` read tolerantly once per scan (missing file = no chips = no band, same posture as `seedExploreFilters`).
- New `tests/title-fit.test.mjs` (auto-discovered by `test-all.mjs`) covering tokenization, seniority exclusion, word-order insensitivity, multi-target best-match, and the empty-profile no-op.

Happy to implement if this fits the roadmap - the diff is small and fully covered by tests.

BigBrown10:
Closes #3260

## Summary

Adds a free, zero-token triage signal to scan-mode discovery cards, as proposed in #3260: a banded estimate of how well a posting's title overlaps the user's profile target roles, computed at discovery time from data already loaded server-side.

- **New `web/src/lib/title-fit.mjs`** - pure, plain-`.mjs` module (Turbopack pins the root to `web/`, so core helpers are mirrored rather than imported, same as `profile-keywords.mjs`). Token-overlap ratio between the title and each of the user's `config/profile.yml` target roles (seniority words excluded from the ratio), best role wins, banded `strong` / `related` / `weak`.
- **`DiscoveredOffer.fit?`** - optional field only; the stream grammar and AI offers (which already carry `why` + `confidence`) are untouched. Kept truly absent (spread-in) so offer objects stay JSON-clean.
- **`lib/core/scan.ts`** - reads `config/profile.yml` tolerantly ONCE per scan (missing file = no chips, same posture as `seedExploreFilters`) and annotates at both offer-construction sites (`--json` and legacy stdout parse).
- **`discovery-card.tsx`** - small passive chip next to the existing `matched` note; `strong` gets a quiet green tint, others stay faint. Tooltip states exactly what it is and is not: a free keyword-level estimate, NOT an evaluation - Evaluate still gives the real A-F.

## House rules

- Recommendation layer ONLY: nothing is filtered, hidden, reordered, or auto-added; summary counts unchanged (#3260 contract).
- Zero tokens, fully offline: string math on data already read server-side.
- The holistic evaluation (1-5 / A-F) remains the only quality verdict; the chip explicitly points to it.

## Testing

`node test-all.mjs --only title-fit` - 8 assertions, all passing: word-order/punctuation titles (the case substring matching misses), seniority-word exclusion, band thresholds pinned at 0.67/0.5/0.33, multi-target best-match, empty-profile no-op posture (null, chip omitted), single-token target behavior, determinism.

Full suite + web typecheck left to CI (no `node_modules` in this environment). Happy to adjust thresholds, naming, or extend to AI-mode offers if reviewers prefer.

## Summary by CodeRabbit

* **New Features**
  * Discovery cards now show an optional title-fit badge for strong, related, or weak matches.
  * Offers can include profile-based title-fit scores when available.
  * Scores are advisory keyword-level estimates and do not replace overall evaluation.

* **Tests**
  * Added coverage for title matching, scoring bands, punctuation and word order, invalid inputs, deterministic results, and error reporting.

### DIFF
--- a/tests/title-fit.test.mjs
+++ b/tests/title-fit.test.mjs
@@ -1,0 +1,88 @@
+// tests/title-fit.test.mjs — free banded title-vs-profile fit estimator (#3260).
+// Pure string math: no network, no tokens, no filesystem. The band annotates
+// discovery cards; the assertions pin the CONTRACT (recommendation layer,
+// bands not percentages in the UI) as much as the arithmetic.
+import { pass, fail, ROOT } from './helpers.mjs';
+import { join } from 'path';
+import { pathToFileURL } from 'url';
+
+console.log('\ntitle-fit.mjs — free banded title/profile overlap hint (#3260)');
+try {
+  const m = await import(pathToFileURL(join(ROOT, 'web', 'src', 'lib', 'title-fit.mjs')).href);
+  const { titleFit } = m;
+
+  // Word order + punctuation insensitivity — exactly what firstMatch()'s
+  // substring test misses ("platform engineer" chip vs "Engineer, Platform").
+  const r = titleFit('Staff Engineer, Platform', ['platform engineer']);
+  if (r && r.band === 'strong' && r.score === 1) {
+    pass('word-order/punctuation titles still band strong (substring matching misses these)');
+  } else {
+    fail(`expected strong/1 for 'Staff Engineer, Platform': ${JSON.stringify(r)}`);
+  }
+
+  // Seniority words must not dilute the ratio.
+  const rs = titleFit('Senior Staff Platform Engineer', ['platform engineer']);
+  if (rs && rs.band === 'strong') {
+    pass('seniority words (senior/staff) excluded from the ratio');
+  } else {
+    fail(`seniority dilution: ${JSON.stringify(rs)}`);
+  }
+
+  // Partial overlap → related; unrelated → weak (still returned, never hidden).
+  const rp = titleFit('Data Engineer', ['platform engineer']);
+  const rw = titleFit('Product Designer', ['platform engineer']);
+  if (rp && rp.band === 'related' && rw && rw.band === 'weak') {
+    pass('partial overlap → related, no overlap → weak');
+  } else {
+    fail(`banding wrong: partial=${JSON.stringify(rp)} none=${JSON.stringify(rw)}`);
+  }
+
+  // Multi-target profiles take the BEST role match.
+  const rm = titleFit('Data Engineer', ['product designer', 'data engineer']);
+  if (rm && rm.band === 'strong') {
+    pass('best-matching target role wins across profile targets');
+  } else {
+    fail(`multi-target should pick data engineer: ${JSON.stringify(rm)}`);
+  }
+
+  // Band thresholds pinned: 2/3 hits ≈ 0.67 strong, 1/3 ≈ 0.33 weak, 1/2 related.
+  const t3 = ['alpha beta gamma'];
+  const b67 = titleFit('x alpha beta', t3), b33 = titleFit('x gamma', t3), b50 = titleFit('x alpha', ['alpha beta']);
+  if (b67?.band === 'strong' && b33?.band === 'weak' && b50?.band === 'related'
+      && b67.score === 0.67 && b33.score === 0.33 && b50.score === 0.5) {
+    pass('thresholds: >=0.6 strong, >=0.34 related, else weak (scores rounded to 2dp)');
+  } else {
+    fail(`thresholds drifted: ${JSON.stringify([b67, b50, b33])}`);
+  }
+
+  // Empty/no-profile posture: null (chip omitted entirely), never a throw.
+  if (titleFit('Anything At All', []) === null
+      && titleFit('', ['engineer']) === null
+      && titleFit(null, ['engineer']) === null
+      && titleFit('Engineer', null) === null
+      && titleFit('Engineer', 'not an array') === null) {
+    pass('empty title/targets/non-array targets → null (caller omits the chip)');
+  } else {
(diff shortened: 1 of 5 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/title-fit.test.mjs`; `web/src/lib/title-fit.mjs` is outside my area and its banding thresholds are scoring-adjacent, so my verdict covers only the test.

All eight assertions trace to source with the arithmetic checked: seniority stripping on both sides, the 2/2 strong, 2/3 strong, 1/3 weak and 1/2 related boundaries, and the five null postures (empty targets, non-array guard). Real reddening via the failure counter, hermetic pure string math. Green on CI. Approving the test; the banding thresholds are santifer's to weigh.

Scott-Emberson (approved):
Re-reviewed at `916aeca5`. My approval sat on `70b972fa`, and `916aeca5` changed `tests/title-fit.test.mjs`, so it needed a fresh look. Only the test file is mine here; `web/src/lib/title-fit.mjs` is not, and I have not judged it.

The terminal-period fix is right and the assertion that covers it is real. I checked by removing the strip and re-running:

```
# .map((t) => t.replace(/\.+$/, ""))  ->  .map((t) => t)
❌ terminal-period handling: [{"band":"related","score":0.5},{"band":"strong","score":1},{"band":"strong","score":1}]
```

`Platform Engineer.` drops from 1/strong to 0.5/related the moment the strip goes, so that part of the check is load bearing. Suite at `916aeca5`: 9 checks, 0 failed. CI green on all fourteen.

**One nit, and it is the half CodeRabbit was pointing at.** `tests/title-fit.test.mjs:82-93`, minor.

The check is named "trailing periods stripped; node.js / .net stay whole", but it only proves the first clause. Both sides of the comparison run through the same `tokens()`, so any symmetric mangling of dots keeps the score at 1. I replaced the terminal strip with one that removes every dot (`t.split(".").join("")`, so `node.js` becomes `nodejs` and `.net` becomes `net`) and re-ran:

```
✅ trailing periods stripped; node.js / .net stay whole
```

Green, on an implementation that does the opposite of what the check's own name and failure message promise. That is the shape worth avoiding: the message tells the next reader the dotted-name behaviour is pinned, so they will not think to check it.

A discriminating case needs the dots to differ across the two sides. Measured against the current implementation:

```
"Node Engineer"     vs ["node.js engineer"]  => {"band":"related","score":0.5}
"NET Developer"     vs [".net developer"]    => {"band":"related","score":0.5}
"Node.js Engineer"  vs ["nodejs engineer"]   => {"band":"related","score":0.5}
```

All three go to 1/strong under the strip-everything version, so any of them added to the pass condition would have caught it.

One correction to the bot on the way past, since it matters if anyone applies the suggestion literally: CodeRabbit asks for `score 0` and `band weak` on those cases. That is not what happens. `engineer` and `developer` still overlap, so the real answer is `0.5` / `related`, as measured above. Pin the measured values, not the suggested ones.

Approving. The nit is not blocking, the fix it sits on is correct, and a follow-up assertion would tighten a check that currently claims more than it tests.

PR #3270, opened 2026-08-24
fix(update-system): never prune a file the update just preserved, or an asset it references

euskode:
## Summary

During `node update-system.mjs apply`, two independent checks run over the same file set:

1. **Local-modification detection** (`locallyModifiedSystemFiles`) — flags a system file this install changed, backs it up to `.bak`, and keeps it.
2. **Stale-file pruning** (`staleSystemFiles`) — deletes any local file matching a system path that upstream no longer ships.

A file can be **both**: locally modified *and* absent from the new upstream tree (e.g. a user's own custom `templates/cv-template.*.html`, which was never an upstream file to begin with). Check 1 correctly detects it, backs it up, and prints "Keeping your versions" — then check 2 deletes it a few lines later in the same run, because it only knows "not present upstream," not "was just preserved."

The same run also deleted font files that preserved template's `@font-face` rules pointed at, once upstream stopped shipping those specific fonts. The fonts themselves were never locally modified, so nothing flagged them as at-risk — but they were still load-bearing for the template check 1 had just decided to keep.

Net effect: a locally-customized CV template and its fonts silently vanish on update, with no error — just two unrelated-looking log lines ("Keeping your versions" followed later by "Pruned stale system file").

## Fix

- `staleSystemFiles()`'s call site now passes `preservedPaths` into the existing `userPaths` exclusion parameter — same mechanism that already protects user-layer paths, just widened to also cover this-run's preserved files.
- New `isReferencedByPreservedFile()` skips pruning a stale candidate when any preserved HTML/CSS file's on-disk content still references its basename (e.g. a font `url(...)` in a kept template).

## Testing

- [x] New `tests/updater-preserved-file-prune.test.mjs` reproduces the scenario (a template with no upstream counterpart, plus a font it references) — 7 assertions, all passing
- [x] `node updater-migration-tests.mjs` — 355 passed, 0 failed (no regressions)
- [x] All existing `tests/updater-*.test.mjs` still green
- [x] `node scripts/check-syntax.mjs` — 525 files pass

Found via a real update run that silently deleted a customized CV template plus 4 font files it depended on.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cleanup of outdated system files while protecting locally preserved files.
  - Retained stale assets referenced by preserved HTML or CSS files.
  - Handled unreadable preserved files safely without interrupting cleanup.
  - Continued pruning unrelated stale files that are not protected or referenced.

- **Tests**
  - Added coverage for preserved-file protection, referenced assets, unsupported file types, and safe handling of unreadable files.

### DIFF
--- a/tests/updater-preserved-file-prune.test.mjs
+++ b/tests/updater-preserved-file-prune.test.mjs
@@ -1,0 +1,98 @@
+/**
+ * updater-preserved-file-prune.test.mjs — a file `apply()` just KEPT because
+ * this install modified it (locallyModifiedSystemFiles → preservedPaths) must
+ * never also be deleted by the separate stale-file prune step, and an asset a
+ * kept file still references (e.g. a custom cv-template's font) must survive
+ * too even when upstream no longer ships it.
+ *
+ * Reproduces the real-world failure: a user's own custom CV template (no
+ * upstream counterpart at all) was backed up to .bak as "kept", then unlinked
+ * by the stale-file prune step in the same `apply()` run, because the two
+ * checks ran independently. The fonts that template's @font-face rules
+ * pointed at were pruned right alongside it, for the same underlying reason.
+ */
+
+import { pass, fail } from './helpers.mjs';
+import { staleSystemFiles, isReferencedByPreservedFile } from '../update-system.mjs';
+
+console.log('\n🧪 Testing preserved-file vs. stale-file prune interaction...');
+
+// ── 1. a preserved (locally-modified) file must not also be pruned as stale ──
+{
+  // Mirrors the real incident: a user's custom template has no upstream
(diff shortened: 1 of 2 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/updater-preserved-file-prune.test.mjs`; `update-system.mjs` is the red-line updater and outside my verdict.

Both imported symbols exist at head and the assertions trace to the real `apply()` path. The never-touch-user-data property is asserted rather than assumed: a custom template and a referenced font both survive the prune while an unrelated stale file is still removed, and an unreadable preserved file fails closed. The repro sentinel proves the un-widened check flags before the fix excludes. Hermetic (injected readFile seam, no git, fs or network). Green on CI. Approving the test; the updater change is santifer's.

Issue #3259, opened 2026-08-23
tests/plugin-symlink-discovery.test.mjs regresses the #2828 pattern — unguarded symlinkSync fails with EPERM on non-elevated Windows


jaybee2184:
## What happens

Running `node test-all.mjs` on Windows without Developer Mode enabled (the default Windows configuration, non-elevated shell):

```
plugins/_engine.mjs — symlinked plugin discovery (#3140)
  ❌ tests\plugin-symlink-discovery.test.mjs — suite threw and was contained (EPERM): EPERM: operation not permitted, symlink 'C:\Users\...\checkouts\career-ops-plugin-demo' -> 'C:\Users\...\plugins.local\demo'
      at symlinkSync (node:fs:1855:11)
      at file:///.../tests/plugin-symlink-discovery.test.mjs:68:3
```

`tests/plugin-symlink-discovery.test.mjs` (added for #3140, merged 2026-08-21) calls `symlinkSync` **unguarded, three times** (lines 68, 94, 130) with no capability probe and no try/skip around any of them.

## This is the #2828 shape, in a newer file

#2828 diagnosed and fixed this exact pattern in `tests/intake.test.mjs`: creating a symlink on Windows needs `SeCreateSymbolicLinkPrivilege`, held by Administrators or by everyone once Developer Mode is on — a standard non-elevated shell gets `EPERM`. The maintainer's own framing on that issue:

> Developer Mode off in a non-elevated shell is not an exotic setup, it is the Windows default. So the person this hits is a contributor on a normal Windows box.

The fix landed in two halves:
- **The class fix (#2867):** `test-all.mjs`'s discovered-suite loop now catches an uncaught throw from a discovered suite and reports it as one named failed suite instead of taking down the whole run. This part is working correctly here — the suite is "contained" rather than crashing `test-all.mjs`, per the log line above.
- **The instance fix (in `intake.test.mjs` itself):** probe once for symlink capability, skip the symlink-dependent cases with a printed reason when it's absent, and let the rest of the file run.

`plugin-symlink-discovery.test.mjs` was added after that fix landed, for a different feature (#3140), and never picked up the instance-level guard — so every one of its 12 checks is lost as a single opaque `EPERM` failure on any contributor's default Windows box, even though the plugin-discovery logic it's actually testing may be completely correct.

## Suggested fix

[...]

PR #3297, opened 2026-08-25
docs(title-filter): the side `stem:` cannot close, and what `word:` costs to close it

artemtrofymenko:
Follow-up to #3225, writing down the bound @santifer asked for on #3103:

> `stem:` fixes the Zero Trust half of your eleven and not the Solanaceae half — there the keyword genuinely starts the word, so those need `word:`. Your three-row table already tells a user which case they're in; saying it out loud saves the next person the experiment.

#3225 merged before that landed, so this is the paragraph, plus the assertion that keeps it true.

## What was missing

The template documented **one** side `stem:` does not cover — a keyword that ENDS a compound, `Automation` inside `Gebäudeautomation`, served by the plain form. It did not document the mirror, which is the side a reader is likelier to hit: a keyword that **starts** the unwanted word. `stem:` asks for the left boundary only, so `Solanaceae`, `Neutronics` and `Defined` match on purpose, and only `word:` closes them.

That is derivable from the three-row table — `stem:agent` still matching `Agentforce` says it — but deriving it costs the reader an experiment, which is the thing you asked to spare them.

## Re-measured rather than quoted

Numbers taken again on merged `main` through `buildTitleFilter`, not through a second copy of the rule, over the 675 observed titles in `tests/fixtures/title-recall-corpus.json`:

```
stem: closes        Zero Trust Engineer x2, NASTAR x2                 4 titles
only word: closes   Solanaceae, Neutronics, Software Defined x3,
                    Product Definer, Product Definition               7 titles
word: also drops    Smart Contracts x3, Engineering Lead - Crypto
                    and DeFi                                         4 titles
                    ...real roles, which is why it is not the default
```

4 + 7 is the eleven; the last row is the four the earlier table charged `word:` with, so the two halves reconcile against the fixture rather than against my own notes.

## The assertion

Section 2 of `tests/title-stem-prefix.test.mjs` pinned the compound side rather than leaving a reader to find it. This pins the other side the same way, and asserts **both halves per title** — `stem:` still matches, `word:` does not — so an implementation that anchored both sides fails it instead of passing half of it.

Verified by mutation: making `stemPattern` two-sided turns the new check red and names both titles.

```
❌ the start-of-word case differs from what the template documents:
   [["solana","Genomic Breeder (Solanaceae)"],
    ["neutron","Senior Neutronics Engineer - Isotope Production"]]
```

Both titles are observed corpus entries, so assertion 4's synthetic-marking rule is unaffected.

## Verification

No behaviour change — the shipped filter is untouched and still uses no prefix, which the template's own last check confirms.

```
tests/title-stem-prefix.test.mjs   9/9
templates/portals.example.yml      parses; positive=37 negative=35
node test-all.mjs                  5336 passed, 1 failed
```

That one failure is `tests/plugin-symlink-discovery.test.mjs` dying with EPERM, which is #3259 and is unrelated to this change — it is the default-Windows bug #3267 fixes, and this branch is off `main`, so it carries it. On a machine that can create symlinks the run is clean.

## Summary by CodeRabbit

* **Documentation**
  * Added examples and guidance explaining how `stem:` and `word:` title filters handle mirror-case matching, suffixes, and compound words.

* **Tests**
  * Added coverage for keywords appearing at the start of unwanted words, validating the expected differences between `stem:` and `word:` matching.

### DIFF
--- a/templates/portals.example.yml
+++ b/templates/portals.example.yml
@@ -314,6 +314,24 @@ title_filter:
   # plain form. That side has no prefix, deliberately — the plain default already
   # serves it, and nothing has been measured that the pair cannot express.
   #
+  # Nor does it close the MIRROR case, which is worth knowing before reaching for
+  # it: a keyword that genuinely STARTS the unwanted word. "Solana" does begin
+  # "Solanaceae", "Neutron" begins "Neutronics", "DeFi" begins "Defined" — `stem:`
+  # asks for the left boundary only, so it matches all three on purpose, and only
+  # `word:` closes them. Measured over the 675 observed titles in
+  # tests/fixtures/title-recall-corpus.json, that is where the two prefixes divide:
+  #
+  #   stem: closes      Zero Trust Engineer, NASTAR Course Setter        4 titles
+  #   only word: closes Solanaceae, Neutronics, Software Defined (x3),
+  #                     Product Definer, Product Definition              7 titles
+  #
+  # And `word:` is not free on that same corpus: it also drops 4 titles you
+  # probably wanted — "Engineering Manager, Smart Contracts" and "Engineering
+  # Lead - Crypto and DeFi" among them — because it forbids the suffix too. So
+  # reach for `stem:` when a keyword lands MID-word, and for `word:` when it lands
+  # at the START of a word you did not mean, knowing that costs the plural and
+  # compound forms as well.
+  #
   # Use " + " (with spaces) to require SEVERAL terms in any order:
   #
   #   - "director + engineering"   matches "Director of Engineering",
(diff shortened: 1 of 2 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/title-stem-prefix.test.mjs`; `title-keywords.mjs` is outside my area, read for context.

The new block is a genuine two-sided sentinel per title: `stem:` must match AND `word:` must not, which discriminates the left-boundary-only stem pattern from the both-anchored word pattern rather than just asserting a match. Hermetic pure strings, mirroring the adjacent compound-case block. The corpus-membership line is doc-only, not load-bearing. Green on CI. Approving.

PR #3298, opened 2026-08-25
fix(followup): stop truncating plus-addressed emails at the plus

darkpandawarrior:
`EMAIL_RE` uses `\w` for the local part, and `\w` is `[A-Za-z0-9_]` — so a `+` terminates the match and the address is silently **truncated**:

```
recruiter+a1b2c3d4e5f60718293a4b5c6d7e8f90@reply.cutshort.io
       ↓ captured as
       a1b2c3d4e5f60718293a4b5c6d7e8f90@reply.cutshort.io
```

That's a different mailbox. It would bounce. And it still looks like a perfectly valid address, so nothing downstream can notice — the tracker displays a contact that can never be reached, and a follow-up sent to it fails silently.

## Why this one matters more than it looks

Recruiting platforms route replies through exactly this form. CutShort, Greenhouse, Lever and Workable all use `name+token@reply.domain`. So **the addresses being corrupted are precisely the ones that reach a human** rather than a `no-reply` box — the follow-up module's whole purpose.

Found live on a real tracker: the highest-scoring row in the pipeline had an open screening questionnaire, its only repliable address was a CutShort relay, and the stored contact was the truncated form. The one contact that mattered most was the one that was wrong.

## Test

`tests/contacts-plus-addressing.test.mjs` asserts the **broken** output specifically, not just the correct one. A regression here produces something email-shaped, so an assertion that only checked "looks like an email" would pass straight through it.

```
node test-all.mjs   5361 passed
```

The 3 failures (`PDF manifest`, `application-artifacts`, `repo path resolved to ""`) are pre-existing on `main` and unrelated.

## Summary by CodeRabbit

- **Bug Fixes**
  - Email extraction now preserves addresses containing `+` or `%` in the local part.
  - Prevents truncation of plus-addressed mailboxes, including relay-format and punctuation-based addresses.

- **Tests**
  - Added coverage for plus-addressed emails, ordinary addresses, relay formats, punctuation, and truncation prevention.

### DIFF
--- a/followup-cadence.mjs
+++ b/followup-cadence.mjs
@@ -556,7 +556,13 @@ export function isRetired(cleared, lastFollowupDate) {
 // Emitted shape is `{ name, email, channel }`. `email` stays first-class (and
 // remains non-null for email contacts) so existing consumers keep working;
 // `channel` is additive.
-const EMAIL_RE = /[\w.-]+@[\w.-]+\.\w+/g;
+// `+` must be in the local part. \w is [A-Za-z0-9_], so the previous class silently TRUNCATED a
+// plus-addressed address at the plus — `mayank+6a88…@reply.cutshort.io` was captured as
+// `6a88…@reply.cutshort.io`, a different mailbox that would bounce. Recruiting platforms route
+// replies through exactly this form (CutShort, Greenhouse, Lever, Workable all use
+// `name+token@reply.domain`), so the addresses most worth capturing were the ones being corrupted
+// — and corrupted into something that still looks like a valid address, so nothing notices.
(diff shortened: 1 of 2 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/contacts-plus-addressing.test.mjs`; `followup-cadence.mjs` is outside my area and I read it for context.

The test imports the real `extractContacts` and asserts the whole email address per case, and the load-bearing part is the negative control: it `fail()`s on the truncated `a1b2@reply.cutshort.io` form and on undefined/empty, so it cannot pass on nothing. That reddens if the plus-address truncation returns. Hermetic, neighbour-style. Green on CI. Approving.

PR #3301, opened 2026-08-25
fix(scan): three scanners stamp first_seen in UTC, the reader measures it in local

darkpandawarrior:
#3070 moved `shouldDedupScanHistoryRow` to the local calendar day, because the UTC day is tomorrow for a west-of-Greenwich evening run and a cooldown opened early. The comment it left at `scan.mjs:987` says so.

That fixed the reader. It didn't fix the writers — and `first_seen` is the other operand in the comparison the reader performs.

Four scanners append to `data/scan-history.tsv`. One stamps `first_seen` with `localToday()`; three stamp it with the UTC day:

| writer | `first_seen` | |
|---|---|---|
| `scan.mjs:2485` | `localToday()` | correct |
| `scan-ats-full.mjs:687` | `new Date().toISOString().slice(0, 10)` | UTC |
| `scan-hn.mjs:116` | `new Date().toISOString().slice(0, 10)` | UTC |
| `scan-interamt.mjs:260` | `new Date().toISOString().slice(0, 10)` | UTC |

So one file now carries rows written on two different clocks, and the reader measures all of them against the local day.

## What it does

`shouldDedupScanHistoryRow` computes `daysBetweenIsoDates(firstSeen, today)` and compares it to `recheck_after_days`. With the two sides on different clocks the age is off by one, in opposite directions depending on which side of Greenwich you're on:

- **West** (New York, 20:00 local): UTC has already rolled over, so `first_seen` is stamped **tomorrow**. The row's age reads one low and the posting stays suppressed a day longer than `recheck_after_days` asked.
- **East** (UTC+5:30, 03:00 local): UTC hasn't rolled over yet, so `first_seen` is stamped **yesterday**. The age reads one high and the row is rechecked a day early.

Same for the `cooldown:` branch — a posting the user silenced until a date, whose row was written by one of the three UTC scanners.

## Why this shape

`localToday`'s docstring is careful that only "what day is it here" moves, and that UTC-midnight round-tripping stays as it is. This change respects that line: it touches only the three `first_seen` stamps, and leaves every `toISOString()` used for date arithmetic or validation alone.

The invariant isn't new either — `web/tests/lib/pipeline-local-today.test.mjs` already asserts that the dashboard's spawned scan child does exactly this:

> expected `const date = localToday();` … the local-day value must reach the writer, not just get computed and discarded

That check covers the web child only. The three engine-side scanners were never covered.

## The guard

Two tests appended to `tests/local-today-gates.test.mjs`, whose header already frames this file as the place where "what day is it" is pinned. It drew its line at readers — things that "GATE a decision rather than stamping a filename". These are the case in between: writers of the value a gate reads.

Source-level, deliberately: the value is a local `const` inside each scanner's `main()`, reachable only by running a real scan against a live board. What is cheap to check is that no call site hands the writer a UTC-derived day, which is the entire defect.

**The first version of that guard was wrong**, and I'd rather say so than let it read as a clean first draft. It captured the argument list with `appendToScanHistory\(([^)]*)\)` — and `[^)]*` stops at the first `)`, which in `appendToScanHistory(offers, new Date().toISOString().slice(0, 10), 'added')` is the one closing `new Date()`. The argument list read as `offers, new Date(`, no `toISOString` in it, and the inline form — the very shape in `scan-hn.mjs` — passed. Only the missing-import assertion caught that file. It now scans from the call to end of line, and there's a comment saying why rather than an inscrutable regex.

## Verified

Reverting only the three source files (keeping the tests) makes the guard name all three, each with file, line and the offending source:

```
scan-ats-full.mjs:687 — const date = new Date().toISOString().slice(0, 10);
scan-hn.mjs:116       — await appendToScanHistory(newOffers, new Date().toISOString()...
scan-interamt.mjs:260 — const date = new Date().toISOString().slice(0, 10);
```

Restored: `tests/local-today-gates.test.mjs` 10/10, `scripts/check-syntax.mjs` 534 files clean, full `test-all.mjs` green.

## One deliberate side effect, worth your call

In `scan-ats-full.mjs` the same `date` const also names the markdown digest — `${date}.md` — and appears in its log header. Those move to the local day too.

I think that's right and not merely acceptable: a 20:00 New York run currently writes a digest filename dated **tomorrow**. But it is a user-visible filename change, and if you'd rather keep the artifact name on UTC I'll split it into two variables instead.

`scan-interamt.mjs` and `scan-hn.mjs` have no such reuse — their `date` reaches only `appendToScanHistory` and a log line.

## Not in scope

`grep` finds 17 other `new Date().toISOString()` "today" stamps outside tests (`tracker.mjs:305`, `outcome.mjs:56`, `archive-posting.mjs:181`, `stats.mjs:516`, the eval runners' filename dates). Most stamp filenames or metadata rather than feeding a gate, and `tracker.mjs`'s `status_events.date` currently has no reader outside tests. They're a separate judgement call and I've left them alone rather than widen this PR past the one file with a demonstrated mixed-clock comparison.

## Summary by CodeRabbit

* **Bug Fixes**
  * Scan history and digest dates now use the local calendar day instead of UTC, preventing incorrect dates around midnight.

* **Tests**
  * Added coverage to verify all scanners consistently record and display local dates.


Scott-Emberson (commented):
Reviewed as code owner of `tests/`. Scope is `tests/local-today-gates.test.mjs`; the three scanner files are outside my area, read for context.

No defect in the test. The two new cases are a real source-census sentinel: a balanced-paren scan of every `appendToScanHistory` call site rejects any `toISOString`-derived day, and the second asserts each of the four writers imports `localToday`. It reddens under the exact UTC-stamped first_seen bug, and it is hermetic (four fs reads, no clock).

I am commenting rather than approving because this appears to duplicate open PR #3241, which touches the identical four files with the same three scanner fixes and the same additions to this test file (#3244 already merged the base). Whichever lands second conflicts on all four. My suggestion is to consolidate #3301 and #3241 before either merges; the close call is a maintainer's, not mine.

Issue #3247, opened 2026-08-23
test: two test/ suites are never run — test-all.mjs registration is manual and was forgotten twice


vliggio:
## Problem

`test-all.mjs` auto-discovers `tests/**/*.test.mjs` — that was #1440, and the comment on the discovery function is explicit about why:

```js
// Deterministic: recursive readdirSync ... No glob library, no
// registration list. Discovery is limited to tests/ so root-level
// standalone *.test.mjs files are never picked up.
```

But five node:test suites live in `test/` (singular), and those are included the old way: an explicit `run(NODE, ['--test', ...])` written inline in `test-all.mjs`. Registration is manual, so it can be forgotten, and twice it was:

| suite | assertions | registered |
|---|---|---|
| `test/cv-templates.test.mjs` | 25 | §59 |
| `test/pipeline-lock.test.mjs` | 21 | §59b |
| `test/cover-resolver.test.mjs` | 4 | yes |
| `test/profile-photo.test.mjs` | 8 | **no** |
| `test/zh-minimal-template.test.mjs` | 4 | **no** |

Twelve assertions, both files added 2026-07-21, and nothing has run either since. `grep` finds them referenced only in `update-system.mjs`'s file manifest, which is a list of what to ship, not a runner.

Both suites still pass today, so nothing is broken — that's rather the point. They'd pass identically if the code under them had rotted, because no one would know either way.

## A second, quieter effect

A registered suite reports as **one** assertion regardless of size:

```js
const unit = run(NODE, ['--test', 'test/cv-templates.test.mjs']);
if (unit !== null) pass('cv-templates.mjs unit tests pass');
```

`node --test test/cv-templates.test.mjs` says `tests 25`. `test-all.mjs` says `1`. So adding tests to a registered suite doesn't move the reported total at all, and "did my new tests actually run in CI?" can't be answered by looking at the count before and after. I got this wrong myself while working on #3202 — I read a flat total as proof the suite was orphaned, when it was registered and simply aggregated.

Worth noting the two failure modes point opposite ways: an unregistered suite is silently skipped, and a registered one is silently collapsed. Neither shows up as a red test.

## Options

[...]

PR #3303, opened 2026-08-25
test(harness): jd-similarity.test.mjs ran nowhere, and nothing would have said so

artemtrofymenko:
A sibling of #3247, found while measuring it, and deliberately not a fix for it.

## `jd-similarity.test.mjs` runs nowhere

20 assertions, added alongside the script they cover, named in no runner at all:

```
grep -rn "jd-similarity" --include=*.mjs --include=*.md . | grep -v node_modules
  CHANGELOG.md:204,205,412     history
  jd-similarity.mjs:3,9,143    the script's own usage block
  (nothing in test-all.mjs)
```

`test-all.mjs` discovers `tests/**/*.test.mjs` and stops there on purpose — the comment on `discoverTests` says so — and the nine root-level suites are named one by one in the `scripts` list instead. This one never made the list.

It passes, and always did:

```
node jd-similarity.test.mjs
jd-similarity: 20 passed, 0 failed
```

That is the quiet half of this failure. An unrun suite reports nothing, so the repo read exactly as green without it as with it. No count, no summary line and no CI signal distinguishes "20 assertions passing" from "20 assertions absent" — which is the same observation §5 of the `scripts` list already makes in its own comment about the coverage check: *"which is how five unregistered files shipped."*

## The one-line fix is not the point

Registering it is one entry. The guard is what stops the next one:

`tests/root-suite-registration.test.mjs` asserts that every root-level `*.test.mjs` is named **somewhere** in `test-all.mjs`. It checks for the filename rather than parsing the `scripts` array, because the question is "does anything run this", not "which section does" — a suite reached by a future glob, or by an inline `node --test`, is fine by this guard and should be.

Assertion 1 pins the other edge: an empty root listing would satisfy assertion 2 forever while guarding nothing.

Verified by mutation — removing the registration line I just added:

```
❌ 1 root-level suite(s) are never run — nothing in test-all.mjs names them:
     jd-similarity.test.mjs
   Add an entry to the `scripts` list: { name: '<file>', expectExit: 0 }
```

## What this deliberately does not do

`test/` (singular) has the same disease — that is #3247, and I have left it alone. Which mechanism should cover those five files is an open design question in that thread, its author has said they intend to send the fix, and a guard written now would pin whichever answer I happened to prefer. I have posted a measurement there that bears on the choice, and nothing more.

## Verification

```
node test-all.mjs         5338 passed, 1 failed
                          (5336 before; +2 is this guard's own assertions)
harness section 5:        ✅ jd-similarity.test.mjs runs OK
```

The single failure is `tests/plugin-symlink-discovery.test.mjs` dying with EPERM. That is #3259, fixed in #3267, unrelated to this branch and present on `main` for anyone on a default Windows box.

## Summary by CodeRabbit

- **Tests**
  - Added the similarity test suite to the complete test run.
  - Added automated validation to detect root-level test suites that are not registered with the test harness.
  - Improved safeguards against incomplete test execution and overlooked test files.

### DIFF
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -318,6 +318,7 @@ const scripts = [
   { name: 'company-funded.mjs --self-test', expectExit: 0 },
   { name: 'invite-match.mjs --self-test', expectExit: 0 },
   { name: 'invite-match.test.mjs', expectExit: 0 },
+  { name: 'jd-similarity.test.mjs', expectExit: 0 },
   { name: 'tracker-sync-check.mjs --self-test', expectExit: 0 },
   { name: 'updater-migration-tests.mjs', expectExit: 0 },
   { name: 'tracker-columns-tests.mjs', expectExit: 0 },
--- a/tests/root-suite-registration.test.mjs
+++ b/tests/root-suite-registration.test.mjs
@@ -1,0 +1,51 @@
+// tests/root-suite-registration.test.mjs — every root-level *.test.mjs must be
+// registered in test-all.mjs, because nothing discovers them.
+//
+// test-all.mjs auto-discovers tests/**/*.test.mjs (#1440) and the comment on
+// that function is explicit that discovery stops there: "root-level standalone
+// *.test.mjs files are never picked up". Nine such suites live at the repo root
+// and are named one by one in the `scripts` list instead. A list is a thing you
+// can forget, and it was forgotten: jd-similarity.test.mjs was added with 20
+// assertions and appeared in no runner at all — grep found it only in
+// CHANGELOG.md and in the script it tests. It passed the whole time, which is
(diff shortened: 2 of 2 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/root-suite-registration.test.mjs`; the `test-all.mjs` registration line is the shared harness and not mine to comment on, read for context.

This closes a real vacuous-pass class and proves it: as a discovered suite it uses the shared pass/fail and never calls finish(), and dropping the `jd-similarity` registration reddens it (exit 1, the orphan named). The `isRegistered` narrowing strips comments and requires quoting so a prose mention cannot fake registration, and assertion 1 guards the degenerate empty-list case. Hermetic, fs-only. Green on CI. Approving.

PR #3304, opened 2026-08-25
feat(scan-ats-full): gate boards on domain relevance before title-filtering

### LINKED ISSUE #3105: Reverse sweep filters on title, but the discriminator is the company: gate boards on domain relevance before title-filtering
Follow-up to #3089, which I said I would open separately if there was appetite. #3089 and #3103 fix the two mechanical causes of reverse-sweep noise. This one is about the axis the filter runs on, so it is a design question rather than a bug.

## Problem

`scan-ats-full.mjs` decides relevance from the **job title** alone. That works for `scan.mjs`, because `tracked_companies` has already established that the employer is in the user's domain, and the title only has to answer "is this the right role". The reverse sweep deletes that premise and asks the title to answer both questions at once.

A title cannot answer the first one. `Senior Backend Engineer` is character-for-character identical at a Solana infrastructure company and at a supermarket chain. So the user is forced into a choice with no good branch:

- keep broad keywords, and every board in the dataset matches
- drop them, and the sweep goes blind to exactly the roles it exists to discover at companies not yet in `portals.yml`

#3089 makes that choice expressible per scanner, which is worth having, but it does not remove it.

## Proposed Solution

Gate on the **company**, using data the sweep has already paid for.

The sweep fetches every posting on a board before it filters. So, before applying `title_filter` to a board's jobs, ask a cheaper question of the board as a whole:

```
does ANY posting on this board match a strong domain keyword?
   no  -> skip the board entirely, do not title-filter it at all
   yes -> the company is in-domain; apply title_filter normally, broad keywords included
```

The "strong domain keyword" list is a second, deliberately small set - the terms that identify an industry rather than a role (`solana`, `solidity`, `smart contract`, `blockchain`, for a crypto search; the equivalent for any other vertical). Users already know how to write that list, because it is the unambiguous half of the `title_filter` they have.

This reconstructs per run, automatically, the property `tracked_companies` provides by hand. A company qualifies by **what it posts**, not by being on a list someone remembered to update, which is the one thing a curated list can never do.

## Why this is worth more than a stricter title filter

It is the only version of the sweep that keeps broad keywords usable. Once a board is established as in-domain, `Backend Engineer` on it is a real lead again, and the discovery function the sweep exists for is fully restored rather than traded away.

On one sweep of mine, of 194 distinct companies in the output, exactly one had a name that looked crypto-adjacent, and it was a gas-station chain. None of the others post a single crypto role anywhere on their board, so a company gate would have removed them wholesale, before a single title was tested.

## Cost

- **No extra HTTP requests.** The board is already fully in memory when filtering starts.
- One extra pass over each board's job array, against a small keyword list.
- Skipped boards get *cheaper*, since their titles are never filtered and their postings never dated (`enrichDate` currently issues per-job detail requests for undated providers like iCIMS *after* the title check, so a board rejected up front skips those too).

## Open questions, and why this is an issue rather than a PR

1. **Where does the keyword list live?** A new `domain_filter` key, or reuse of a subset of `title_filter.positive`? I lean towards explicit, since silently reusing part of another key is how #3089 happened in the first place.
2. **What is the right threshold?** One matching posting on the board, or N? One risks a single stray posting admitting a whole irrelevant board; N risks excluding a small company with one relevant opening, which is exactly the kind of find the sweep is for. I would start at one and measure.
3. **Does it apply to `--seeds` too**, or only to the full directory sweep? Seeds are already a curated corpus, so gating them would be redundant at best.
4. **Opt-in or default?** Absent a domain list it has to be a no-op, but it is a large enough behaviour change that I would want it opt-in for at least one release.

I have the sweep data to measure any of these, and I am happy to implement it once the shape is agreed. I did not want to send a PR of this size against unsettled answers to 1 and 4.

Does the company-level gate look like the right direction, and would you want it behind its own config key?

IaroslavMazur:
## What does this PR do?

Adds an opt-in `domain_filter` to `portals.yml` that gates a whole ATS board on industry relevance before any title is filtered: if no posting anywhere on a board matches a domain keyword, the board is skipped entire. This restores the discovery function of the reverse sweep, since broad title keywords become usable again once the employer is established.

## Related issue

Closes #3105.

## Type of change

- [ ] Bug fix
- [x] New feature

## Checklist

- [x] If this is a new feature or architecture change, I opened an issue first (#3105, with the four design questions settled in the thread before any code)

---
Implements the company-level gate with the four answers from your decision comment: explicit `domain_filter` key, threshold of one, `--seeds` left ungated, opt-in and off by default. Built on merged `main`, so `stem:` from #3225 is available to it.

### What it does

Before any title is filtered, the sweep asks one question of the board as a whole: does any posting on it match a domain keyword? If none does, the board is skipped entire - no title filtering, no location filtering, no date enrichment.

```yaml
domain_filter:
  - "solana"
  - "defi"
  - "stem:smart contract"
```

The gate sits ahead of the truncation retry and of `processJobs`, which is what makes a skipped board **_cheaper_** rather than merely quieter: `provider.enrichDate()` issues a per-job detail request for undated providers, and a board dropped here never pays for one.

### One correction to my own measurement

My comment on #3105 said the domain list is "a list of stems by nature", so `stem:` is the matching mode it wants. That is right about the plurals and wrong about the junk, and the junk is the larger half. I checked it before writing the tests rather than after:

```
                                  plain   stem:   word:
junk   Analista de Deficiência     match   match     .
junk   Defiance Ohio Plant Mgr     match   match     .
junk   Software Defined Networking match   match     .
junk   Product Definition Lead     match   match     .
junk   Contrato Indefinido         match     .       .
junk   Solanaceae Genetics         match   match     .
junk   Cryptography Engineer       match   match     .
real   Crypto and DeFi             match   match   match
real   Digital Assets Backend Eng  match   match     .
real   Engineering Mgr, Smart Contracts  match match .
real   Core Blockchain Engineer    match   match   match
real   Cryptocurrency Analyst      match   match     .
```

`stem:` closes one of the seven false positives, not six: in all the others the domain term **_starts_** the unwanted word, and only the right-hand anchor closes those. That is the same 4/7 split @artemtrofymenko re-measured on #3297, arriving from the title side.

So, a plain `domain_filter` entry means `word:`, and the asymmetry is what settles the direction: a false positive admits an entire catalogue while a false negative costs one board, and the run summary counts skipped boards, so the loose direction is the one that fails in silence.

The cost is real and stated rather than hidden: `word:` drops `Digital AssetS`, `Smart ContractS` and `CryptoCURRENCY`, which are genuine finds. An entry buys the longer form back with `stem:`, one entry at a time - and `stem:digital asset` carries no junk risk, since no unrelated word begins with it. The example config ships the list already split that way.

### Evidence it works end to end

Twelve Greenhouse boards, `--dry-run`, one domain term:

```
domain_filter: ["stem:robotic"]

  3 boards gated, 36 postings never filtered
  1 board admitted (10beauty, on "Principal Robotics Software Engineer")
  kept: Technical Support Engineer, Principal Embedded Software Engineer, ...
```

The two kept titles carry no domain term at all. That is the point of the feature rather than a leak: once the employer is established, `Engineer` on that board is a real lead again, which is the recall a stricter `title_filter` cannot give back.

Control run, same twelve boards with no `domain_filter`: 17 postings kept, `domainFilterActive: false`, both counters zero. Byte-identical to today.

### Opt-in

The key is the opt-in. `domain_filter` exists in no `portals.yml` written before this, `buildDomainFilter` returns `null` for an absent or empty list, and the example config ships the block commented out, so new installs are off too.

### Reporting, because a missing term costs a board

```
Domain-gated boards: 4 skipped, 48 postings never filtered (--verbose names them)
```

`--verbose` names each skipped board; `--json` carries `domainFilterActive`, `domainGatedBoards` and `domainGatedPostings`. `domainFilterActive` is not derivable from the counts: zero skipped boards is equally what an absent list and a permissive one produce.

### Known limit

The aggregator case from the issue is documented in the code, in `docs/SCRIPTS.md` and in the example config, in the wording you asked for: the gate assumes one board is one employer, `jobgether` passes at every threshold under every matcher, no threshold fixes it, and there the gate degrades to the ungated behaviour.

### The alternative I did not take

An anchored form with an optional plural suffix - `(?<!\w)term(?:e?s)?(?!\w)` - beats both on the numbers I have. On #3103's 675-title corpus it removed all 11 false positives against `word:`'s 11 while losing 1 real role against `word:`'s 4, and here it would close the seven false positives above **_and_** keep `Digital Assets`, `Smart Contracts`, `Blockchains` without a per-entry decision.

I left it out for two reasons. It is a fourth matching mode in a vocabulary #3225 deliberately stopped at three, and it needs its own measurement on the domain corpus rather than an inference from the title one. The per-entry `stem:` gets the same outcome with no new machinery, at the price of the user making the call.

Worth a follow-up issue, or is three modes the right place to stop?

### Changes

- `title-keywords.mjs`: `compileDomainKeyword` / `buildDomainFilter`; the `" + "` group split extracted so both keyword kinds share it, and the defensive normalizer hoisted for the same reason
- `scan-ats-full.mjs`: `boardInDomain`, the gate call site, two counters carried through the checkpoint, the summary line and three `--json` fields
- `validate-portals.mjs`: shape check for `domain_filter`
- `templates/portals.example.yml`, `docs/SCRIPTS.md`, `AGENTS.md`
- `tests/domain-filter-board-gate.test.mjs` (20 assertions), `tests/validate-portals-domain-filter.test.mjs` (4)

## Summary

* **New feature**: Add `domain_filter` to `portals.yml` to skip ATS boards with no domain-matching posting before title filtering (`scan-ats-full.mjs:833`).
* **Matching**: Plain entries use whole-word matching. `word:` and `stem:` provide explicit modes (`title-keywords.mjs:217`).
* **Truncated boards**: The scan defers uncertain boards and admits boards that remain truncated after retry (`scan-ats-full.mjs:531`).
* **Reporting**: Summaries, verbose output, and JSON include gated board and posting counts.
* **Seeds**: `--seeds` scans are not gated.
* **Validation**: `domain_filter` must be a keyword list with valid string entries (`validate-portals.mjs:177`).
* **Compatibility**: Existing configurations continue to work when `domain_filter` is absent or empty. Invalid new configuration values fail validation.
* **Documentation and tests**: Added usage guidance and an example configuration in `docs/SCRIPTS.md:678` and `templates/portals.example.yml:628`. Added focused board-gating and validation tests.
* **System files**: `AGENTS.md:121` changed. `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/` were not changed.


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/domain-filter-board-gate.test.mjs` and `tests/validate-portals-domain-filter.test.mjs`; `scan-ats-full.mjs` is outside my area, read for context.

Two-sided sentinel: a board carrying one domain-matching posting admits the whole board, a board with none is gated out, plus the empty and ragged-array guards. The whole-word default is proven against the measured junk (Deficiência / Defiance) with `stem:` and `word:` recovery. One suite is pure imports, the other spawns validate-portals in a tmpdir; both hermetic and neighbour-style. Green on CI. Approving.

Scott-Emberson (commented):
Re-reviewed at head ae36f67437 (moved since my earlier approval). The delta is a main merge (external-data-dir, isMainModule, providers, hired-wall) plus a maintainer-side conflict-fix on scan-ats-full.mjs where two imports collided on one line (isMainModule + buildDomainFilter, both kept). Neither owned test file (tests/domain-filter-board-gate.test.mjs, tests/validate-portals-domain-filter.test.mjs) appears in the delta, so their assertions are byte-identical to what I approved, and the board-gate contract is unchanged.

Owned tests verified unchanged and sound. Holding formal approval until the branch is finalized, nothing blocking from the tests/ side.

Scott-Emberson (approved):

artemtrofymenko (commented):
Verified at `93b8bde5`. The rebase is clean from my side and the domain-keyword compiler does what its docblock says — I checked all four of its claims by calling it rather than reading it. One thing before @santifer's full read, because your rebase note and CodeRabbit's Major finding say opposite things about the same six lines, and yours went up four minutes after the bot's so you probably had not seen it yet.

## The truncation ordering

Your note:

> a gated board is neither marked dead nor queued for the sequential retry — **nothing downstream would have kept its postings anyway**

The code, `scan-ats-full.mjs:1004-1010`:

```js
if (domainFilter && !boardInDomain(jobs, domainFilter)) {
  domainGatedBoards++;
  domainGatedPostings += jobs.length;
  …
  return;                                    // ← exits here
}
if (jobs.workdayTruncated) truncated.push(entry);   // ← never reached
```

The gate therefore judges the **truncated first page**, and the last clause of your note is the part that does not hold in general. It is true exactly when the fuller retry result would also fail the gate — and a truncated board is, by definition, one whose remaining postings nobody has looked at yet.

Measured on this head by calling the real functions, with `domain_filter: [defi, crypto]`:

```js
page1 = [{title:'Office Manager'}, {title:'Executive Assistant'}]   // workdayTruncated
full  = page1 + [{title:'DeFi Protocol Engineer'}]                  // what the retry returns

boardInDomain(page1, df) → false     // gated, returns, never pushed to truncated[]
boardInDomain(full,  df) → true      // the board was real
```

So the board is dropped on the strength of a page that was explicitly incomplete, and the run summary counts it under `domainGatedBoards` — a bucket that reads as "correctly excluded", not "not yet known". That is the direction your own docblock argues is the dangerous one: *"A false positive admits an entire catalogue while a false negative costs one board, and the run summary counts skipped boards, so the loose direction is the one that fails silently."* This is the mirror — a false negative that also fails silently, because it lands in a counter that means something else.

The cheap-gate reasoning above it still holds for every non-truncated board, and that is the overwhelming majority; the fix is not to move the gate, only to let a `workdayTruncated` result through to the retry and gate the retry's result instead. I am not proposing the diff — where exactly that lands (and whether `domainGatedPostings` should count the truncated page's rows at all) is your call.

**What I did not verify:** that a real Workday board exists whose only domain-bearing posting sits past the truncation point. I have shown the code path and that the two results differ, not a live board where it bites. If you think page ordering makes that unreachable in practice, that is worth one line in the comment — it would close the question rather than leave the next reader to re-derive it.

## The compiler, since it is my old ground

`title-keywords.mjs` is where `word:`/`stem:` came from (#2970, #3225) and the doc paragraph you cite (#3297), so I ran its claims rather than nodding at them:

```
digital asset        → misses "digital assets platform", hits "digital asset custody"
stem:digital asset   → hits both
defi                 → misses Deficiência, Software Defined, Indefinido; hits "defi protocol"
stem:defi            → hits Deficiência and Software Defined; misses Indefinido
```

Which is exactly what the docblock asserts, including the sentence I expected to be wrong — `stem:` really does close only `Indefinido`, because every other junk case *starts* with the term. Sharing the `" + "` split through `compileAndGroup` rather than reimplementing it for domain keywords is the right call, and the comment explaining why is the kind that stops the drift before it starts.

Scott-Emberson (commented):
Test-axis only. The substance is `scan-ats-full.mjs`, `validate-portals.mjs`, and `title-keywords.mjs` plus a red-line `AGENTS.md` touch, which is the maintainer's routing call; I am reviewing the two owned tests.

They are sound. The opt-in default-off invariant is pinned both directions: `buildDomainFilter` returns null for absent, empty, and all-malformed configs and a real function for a genuine list, and the explicit function assertion defeats the always-returns-null tautology. The whole-word default has real mutation coverage rather than a tautology: a junk set (Deficiência, Defiance, Software Defined, Cryptography, Solanaceae) must all be rejected while real terms match, so a stem or left-anchor default reddens on the junk half, and the Unicode-boundary case reddens an ASCII-only anchor. The threshold-of-one gate is exercised both ways (`boardInDomain` true with one relevant posting among three, false once it is dropped, false for empty and all-malformed boards), and the validate-portals test pins malformed-rejected and valid-accepted with an absent-stays-valid contrast that proves the null results are real rejections rather than harness crashes.

One bounded coverage note, not a false-pass: the tests exercise the gate primitive (`boardInDomain`) and the predicate, but the `main()` wiring that places the early return ahead of `processJobs` (the "gate before title-filtering" ordering that is this PR's headline) is not directly exercised, so deleting that early-return block would not redden these tests. The decision function is genuinely pinned; only the placement rests on unit-level trust. An integration test that drives the scan path and asserts a domain-failing board is skipped before any title is tested would close that gap.

No test-axis concern beyond that recommendation. The branch is behind main and the substance is red-line, so routing and merge are the maintainer's call.

artemtrofymenko (commented):
@Scott-Emberson's coverage note checks out, and it matters more than a normal gap because of what my earlier comment found. Measured at `93b8bde5`, deleting the whole gate rather than tweaking it:

```js
- if (domainFilter && !boardInDomain(jobs, domainFilter)) { …; return; }
+ // the PR's headline behaviour, gone
```

```
--only domain-filter-board-gate        20 passed, 0 failed
--only validate-portals-domain-filter   4 passed, 0 failed
--only scan-ats-full-resume            26 passed, 0 failed
--only scan-ats-full                   47 passed, 0 failed
npm run lint                           687 files, clean
```

97 assertions, none of them red with the feature removed. (Scope: those four suites plus lint, not the full run.)

## The part worth flagging before anyone writes that test

The recommendation is an integration test asserting *"a domain-failing board is skipped before any title is tested"*. Written against the placement as it stands, that test **pins the truncation behaviour from my earlier comment** — the gate judging an explicitly incomplete first page and returning before `truncated.push(entry)`.

Which puts two sound recommendations in this thread on a collision course: one says record the current ordering in a test, the other says the current ordering is wrong for one class of board. Whichever lands first decides it, and a test written around the defect is worse than no test — it turns a fixable ordering into one that reddens CI when someone tries to fix it.

So the integration test wants **two** boards, not one:

- a complete board with no domain-bearing posting → skipped ahead of any title work (the property this PR is for), and
- a `workdayTruncated` board with no domain-bearing posting *on the truncated page* → reaches the sequential retry, and is gated on the retry's result.

The second case fails today, which is the point: it is the assertion that makes the fix safe to make, and it costs nothing to write now while the placement is still being decided. If @IaroslavMazur concludes the truncated case is unreachable in practice, then the second board becomes a comment saying so rather than a test — either way the next reader does not have to re-derive it.

Nothing further from me on this PR; the substance and the merge are the maintainer's call.

artemtrofymenko (commented):
Verified at `c15995ee` on Windows 11. The fix does what your note says, and both of your mutation counts reproduce exactly:

```
delete the sweep's gate block             29 passed, 2 failed
move truncated.push below the gate        30 passed, 1 failed
```

Both reds are the source-order guards, as you said.

## The place a deferred board is now decided is not pinned at all

Deferral moves the real verdict for an unmatched truncated board into the retry loop at `scan-ats-full.mjs:1128`. That line is an inline `boardInDomain`, not `boardGateDecision`, and no test reads it:

```diff
- if (domainFilter && !boardInDomain(jobs, domainFilter)) {
+ if (false) {        // the retry never gates, so every deferred board is processed ungated
```

```
--only domain-filter-board-gate          31 passed, 0 failed
--only scan-ats-full                     47 passed, 0 failed
--only scan-ats-full-resume              26 passed, 0 failed
--only validate-portals-domain-filter     4 passed, 0 failed
```

These are the same four suites as last time: 108 assertions, none red. The mutation admits every deferred board without a gate, which is the loose direction the new comment at `:1048-1051` says the deferral exists to close. It is the same shape as the 97, one hop later.

## CodeRabbit's finding is the same gap, and it holds on this head

I read these from the code and did not execute them: the retry runs only inside `main()` against a provider.

1. **A deferred board whose retry is truncated again.** `:1128` gates it and increments `domainGatedBoards`/`domainGatedPostings`, then `:1135` also counts it as an error. The gate has judged an incomplete page again, and the counter records it as correctly excluded. That is the harm `boardGateDecision`'s own docblock describes at `:508-510`. There is no third fetch to defer to, so something has to be chosen here: count the board as unknown, or admit it ungated, the way #3367 treats a capped iCIMS board. What it should not do is increment the gated counter.
2. **A board that matched on its truncated page.** `:1058` processes it and `:1035` queues it for the retry. The docblock at `:514` and the comment at `:1125` rely on the retry returning a superset, but `:1135` exists because the retry can itself come back truncated, cut at a different place. If that cut misses the domain posting, `:1128` counts as gated a board whose matches were already banked. No postings are lost, but the summary counts one board as both processed and gated.

Both cases go away if the retry also branches on a callable decision. Its inputs would be the sweep's verdict (did the partial page already match) and the retry's own `workdayTruncated`; its output would be `process` / `gate` / `unknown`. That function is also what would give those 108 assertions something to read.

## Not verified

- Neither case end to end. I did not run `main()` with a provider that truncates twice, so both cases rest on the code at the lines cited above.

IaroslavMazur:
@artemtrofymenko, both cases hold, and so does the mutation. Pushed as `585068cc`.

> Both cases go away if the retry also branches on a callable decision.

That is the shape now. The sweep records the boards it deferred, and the retry branches on an exported `retryGateDecision(jobs, domainFilter, wasDeferred)`:

1. **Admitted by the sweep**: always `process`. You are right that the "superset" premise does not survive a retry truncated at a different place, so the retry no longer re-judges a board whose matches are already banked. The `boardGateDecision` docblock and `docs/SCRIPTS.md` no longer claim a superset either.
2. **Deferred, retry truncated again**: `process`, i.e., admitted ungated, still counted as an error for staying truncated. I went with admit over a separate "unknown" bucket for the reason @chipoto69 gave on #3367 for a capped iCIMS board: when the gate cannot judge a board, it falls back to the ungated behaviour, the same way it already degrades on an aggregator. The gated counter now only ever means a board the gate could actually judge.
3. **Deferred, complete retry, nothing domain-bearing**: `gate`, the only path to the counter.

Your mutations, re-run on the new head against `tests/domain-filter-board-gate.test.mjs`:

```
retry call site -> if (false)                        1 red
retryGateDecision always returns process             1 red
ignore wasDeferred                                    1 red
drop deferred.add(entry) in the sweep                 1 red
count a twice-truncated board as gated (old logic)    1 red
```

The call site is still pinned by a source-order guard rather than an integration test, same as the sweep: the guard asserts the retry branches on `retryGateDecision(...)` ahead of `processJobs`, and that no inline `boardInDomain(` is left in the retry.

Locally: `npm run lint`, then `domain-filter-board-gate` 41 passed (up from 31), `scan-ats-full` 47, `scan-ats-full-resume` 26, `validate-portals-domain-filter` 4, all 0 failed, and `node test-all.mjs --quick` at 8672 passed, 0 failed.

@santifer, the branch is still `BEHIND` and I have left it that way, since I did not hear back on whether you are mid-read. Happy to update it onto `main` whenever you say.

artemtrofymenko (commented):
Verified at `585068cc` on Windows 11. Every number in your note reproduces.

```
baseline                                              41 passed, 0 failed
retry call site -> if (false)                         40 passed, 1 failed
retryGateDecision always returns process              40 passed, 1 failed
ignore wasDeferred                                    40 passed, 1 failed
drop deferred.add(entry) in the sweep                 40 passed, 1 failed
twice-truncated board counted as gated (old logic)    40 passed, 1 failed

scan-ats-full 47 · scan-ats-full-resume 26 · validate-portals-domain-filter 4, all 0 failed · lint clean
```

Each mutation reddens a different, correctly named assertion, so none of them is caught only by accident.

`retryGateDecision` reads the way the three cases need. A board the sweep admitted returns `process` before anything is judged. A deferred board that comes back truncated again maps `defer` to `process`. Only a complete, off-domain deferred result reaches `gate`. So the gated counter now means only what its name says, which was the point of both findings. Admit over a separate "unknown" bucket is a reasonable call, and it matches how the gate already degrades elsewhere.

Nothing further from me on this PR.

PR #3311, opened 2026-08-25
fix(workday): recover offset-clamped tenants with a facet split

### LINKED ISSUE #3310: workday: CXS pagination clamps at offset 2000 and reports total=2000, silently truncating large tenants
### Summary

Some Workday tenants' CXS backend refuses to paginate past **offset 2000**, and reports `total` as exactly `2000` while doing it. `providers/workday.mjs` trusts that `total`, so the board is silently truncated to at most 2,000 postings — and because `DEFAULT_MAX_PAGES * PAGE_SIZE` is also 2,000, the existing cap warning fires with advice (`raise max_pages`) that cannot help: offsets past the ceiling return the *same postings as offset 0*.

The provider already suspects this — `providers/workday.mjs` carries the comment `total may be Workday-capped, not real` and names `dickssportinggoods` — but it only tags the warning; there is no recovery.

### Reproduction (live, 2026-08-25)

`dickssportinggoods|wd1|dsg`, `POST https://dickssportinggoods.wd1.myworkdayjobs.com/wday/cxs/dickssportinggoods/dsg/jobs`:

| offset | reported `total` | first posting |
|--------|------------------|---------------|
| 0      | 2000             | Specialist Footwear |
| 1980   | 0                | Teammate Footwear |
| 2000   | 2000             | Specialist Golf |
| 4000   | 2000             | Specialist Golf |

`offset=2000` and `offset=4000` return **identical** results — pagination is clamped, not exhausted.

The same page-0 response proves the real board is far bigger, because **the facet counts are not clamped**:

```
FACET CF_-_Location_Type__EEB__Extended (Brand)  7813 + 403 + 188 + 11  = 8415
FACET timeType (Time Type)                       7044 + 1307            = 8351
FACET workerSubType (Job Type)                   8316 + 109             = 8425
FACET jobFamily (Job Family)                     24 values, max 6564
```

So the tenant reports 2,000 and actually has ~8,400. The public careers page listed 7,120+ at the time the existing code comment was written.

Not every large tenant is affected — `cvshealth|wd1|cvs_health_careers` reports `total=19173` and returns genuinely different results at offset 4000. The clamp is per-tenant, which is why a blanket `max_pages` raise is the wrong lever.

### Impact

Worst on `scan-ats-full.mjs`'s reverse sweep: synthesized entries have no `portals.yml` row, so `max_pages` cannot be raised per tenant even where raising it would help, and the sweep just logs a warning and moves on. Large enterprise tenants are exactly where senior/director-level reqs live, so the postings lost are not a random sample.

### Proposed fix — facet split

The same response that reveals the clamp also supplies the way around it. Re-issue the query once per value of a facet that partitions the board (`appliedFacets: { jobFamily: [<valueId>] }`), so each slice is under the ceiling and paginates honestly. Verified live: `CF_-_Location_Type__EEB__Extended = Public Lands` (count 11) returns a real `total=11` and works.

A slice can itself be clamped (`jobFamily = Field - Teammate`, count 6564, still reports `total=2000`), so the split has to recurse onto a second, not-yet-applied facet — bounded, since a tenant could otherwise report a clamp at every level.

Sketch:

1. `trueTotalFromFacets(facets)` — largest sum of a facet's value counts; detects the clamp by comparing against the reported `total`.
2. `chooseSplitFacet(facets, { exclude })` — pick the facet with the smallest largest-slice, skipping facets with fewer than two usable values and id-less group headers (`locationMainGroup` ships `id: null`).
3. On detection, stop unfaceted pagination immediately (every page past the ceiling is a duplicate of page 0) and recurse into slices, deduping the union on posting URL.
4. Bound depth and total slices; tag `jobs.workdayTruncated` when the split still could not cover the board, so a partial result is never reported as complete.

Facet parameters are tenant-specific (`CF_-_Location_Type__EEB__Extended` is DSG's own), so the split must read them from the response rather than hardcoding names.

I have this implemented with tests and will open a PR against this issue.

bigguy6883:
Closes #3310.

### The bug

Some Workday tenants' CXS backend refuses to paginate past **offset 2000** and reports `total` as exactly `2000` while doing it. The provider trusts that `total`, so the board is silently truncated — and since `DEFAULT_MAX_PAGES * PAGE_SIZE` is also 2000, the existing cap warning fires with advice (`raise max_pages`) that cannot help: offsets past the ceiling return the *same postings as offset 0*.

Verified live on `dickssportinggoods|wd1|dsg` (2026-08-25): `offset=2000` and `offset=4000` both return `Specialist Golf` as the first posting. The same page-0 response proves the real size, because **the facet counts are not clamped**:

```
workerSubType   8314 + 109              = 8423
Brand           7821 + 403 + 188 + 11   = 8423
jobFamily       24 values               = 8423
```

The clamp is per-tenant — `cvshealth|wd1|cvs_health_careers` reports `total=19173` and returns genuinely different results at offset 4000 — so a blanket `max_pages` raise is the wrong lever.

### The fix

Detect the clamp by comparing the reported `total` against what the facets add up to, then re-issue the query once per value of a facet that partitions the board, deduping the union on posting URL. Facet parameters are tenant-specific (`CF_-_Location_Type__EEB__Extended` is DSG's own), so they are read from the response, never hardcoded.

Two properties matter more than the split itself, and I only learned them by running against the real board — **every mock test passed while the implementation was strictly worse than doing nothing**:

- **The split is additive.** My first version skipped the unfaceted crawl on detecting the clamp, reasoning that pages past the ceiling are duplicates. Live result: **917 postings recovered, where the unpatched code returns 2,000.** Everything below the ceiling is real and reachable; the slices build on that floor.
- **A clamped slice is still paginated to the ceiling**, for the same reason. Taking only its page 0 was the other half of the 917.

### What it does and does not promise

It recovers coverage; it does not promise completeness. Real boards are skewed: DSG holds **6,564 of 8,423 postings in one `jobFamily` value**, and *inside that slice* every remaining facet is skewed the same way (Brand 6562/6564, timeType 6483/6499, Location Type 6118/6564). The dominant mass never splits below the ceiling, so a board the split could not finish keeps the existing `workdayTruncated` tag rather than being reported as complete.

### Live result

Same tenant, same script, before and after:

| | postings | requests |
|---|---|---|
| before | 2,000 (reported as the whole board) | 100 |
| after | **4,234 unique** (tagged `workdayTruncated`) | 501 |

2.1x coverage, honestly labelled as still short of the ~8,423 the facets describe.

### Bounds

One pathological tenant must not eat a sweep, so the split is bounded on three axes: `MAX_SPLIT_DEPTH` (2), `MAX_SPLIT_SLICES` (100), and a per-tenant page budget (`max_pages * SPLIT_PAGE_BUDGET_FACTOR`) shared by the unfaceted crawl and every slice. The live run stopped at exactly 501 requests. Hitting any bound sets `workdayTruncated`.

### Unclamped tenants are untouched

No faceted request, no dedup, no extra page — the existing path runs exactly as before. A `ctx.maxPages` probe (verify-portals, discover-ats) never splits: it asked for one page. The pre-existing `workday.test.mjs` suite passes unchanged, 61/61, including the `total may be Workday-capped` warning test — that fixture ships no facets, so it takes the old path.

### New API

- `trueTotalFromFacets(facets)` — board size per the facet counts, or null.
- `chooseSplitFacet(facets, { exclude })` — the facet to split on, skipping single-valued facets and id-less group headers (`locationMainGroup` ships `id: null`).

Both exported for the test suite, which pins the live DSG numbers.

### Testing

- New: `tests/providers/workday-facet-split.test.mjs`, 18 assertions. Includes an explicit regression guard for the additive property, named after the 917-vs-2000 failure.
- Existing `tests/providers/workday.test.mjs`: 61/61, unchanged.
- `tests/providers/ats-ssrf-hardening` (3), `tests/scan-ats-full-resume` (26), `tests/scan-ats-full-outage-checkpoint` (5), `tests/scan-ats-full-title-filter-full` (5) — all green.
- `npm run lint` clean.

`node test-all.mjs` OOMs on my Raspberry Pi, so I ran the affected suites individually rather than the full runner; CI will cover the rest.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Workday job listing recovery when pagination limits hide available postings.
  - Added facet-based recovery, including nested splits, to find additional postings.
  - Prevented duplicate postings when results overlap across recovery slices.
  - Added safeguards to limit recovery depth, requests, and processing time.
  - Clearly identifies results when recovery remains incomplete.
  - Preserved existing retry behavior, delays, date filtering, no-date handling, and result limits.
  - Supports filtered searches during recovery while retaining complete results from unaffected listings.

### DIFF
--- a/providers/workday.mjs
+++ b/providers/workday.mjs
@@ -485,8 +485,24 @@ export default {
           slicesSpent++;
           await sleep(INTER_PAGE_DELAY_MS, ctx);
           const nextApplied = { ...applied, [facet.facetParameter]: [value.id] };
+          // runQuery()'s page-0 fetch is unguarded — fine for the one page-0 of
+          // an ordinary board, but here it runs once per slice against a tenant
+          // that is by definition large, which is where a WAF or rate limiter
+          // lives. Letting it throw would abandon the whole tenant including
+          // the unfaceted crawl already absorbed into `out`, so a dead slice
+          // becomes an incomplete split and the rest of the partition is still
+          // tried. Same accounting as a slice that died mid-pagination.
+          let sliceResult;
+          try {
+            sliceResult = await runQuery(nextApplied);
+          } catch (err) {
+            const attempts = err.attempts ?? RETRY_POLICY.retries + 1;
+            console.error(`⚠️  workday: ${entry.name} slice ${facet.facetParameter}=${value.id} failed on its first page after ${attempts} attempts: ${err.message}`);
+            splitIncomplete = true;
+            continue;
+          }
           await split(
-            await runQuery(nextApplied),
+            sliceResult,
             nextApplied,
             depth + 1,
             [...excluded, facet.facetParameter],
--- a/providers/workday.mjs
+++ b/providers/workday.mjs
@@ -456,6 +456,23 @@ export default {
        */
       const split = async (result, applied, depth, excluded) => {
         absorb(result.jobs);
+        // A slice that stopped early is not a slice that finished. `clamped` is
+        // only ever true for stopReason 'complete', so without this the
+        // `!result.clamped` return below absorbs a slice's partial jobs and
+        // reports the board recovered — the one thing this path exists to
+        // avoid. 'early-stop' (and 'no-date-skip', which only drops postings
+        // the sweep would discard anyway) stay exempt: those slices are
+        // genuinely done for this sweep's purposes.
+        //
+        // 'cap' only counts against an UNCLAMPED query, matching the entry-cap
+        // warning below. A clamped query reports total at the ceiling, which is
+        // exactly maxPages * PAGE_SIZE, so it always ends at the cap — that is
+        // the clamp being detected, not pages going unread, and it is what the
+        // split then recovers. Tagging it would put "(still incomplete)" on
+        // every clamped board and say nothing.
+        if (result.stopReason === 'fetch-error' || (result.stopReason === 'cap' && !result.clamped)) {
+          splitIncomplete = true;
+        }
         if (!result.clamped) return;
 
         if (depth >= MAX_SPLIT_DEPTH) { splitIncomplete = true; return; }
--- a/providers/workday.mjs
+++ b/providers/workday.mjs
@@ -43,6 +43,28 @@ const RETRY_POLICY = { retries: 3 };
 // page 1 pay this; no-date-skip and early-stopped tenants never do.
 const INTER_PAGE_DELAY_MS = 250;
 
+// Offset past which some tenants' CXS backend stops paginating: it reports
+// `total` as exactly 2000 and answers offset=2000/4000 with the same postings
+// as offset=0. Raising max_pages buys duplicates, not coverage — see the facet
+// split below for the way around it.
+const WORKDAY_OFFSET_CEILING = 2000;
+
+// How many times a slice may itself be split. Two levels turn a clamped board
+// into (values of facet A) x (values of facet B) queries, which cleared every
+// clamped tenant observed; the bound exists because a tenant that reports a
+// clamp at every level would otherwise recurse until it runs out of facets.
+const MAX_SPLIT_DEPTH = 2;
+
+// Total slice queries one tenant may spend. A pathological facet (hundreds of
+// values, each still clamped) must not turn one board into an unbounded crawl.
+const MAX_SPLIT_SLICES = 100;
+
+// Page budget for a whole tenant, as a multiple of max_pages. A clamped board
+// is crawled once unfaceted and then once per slice, and slices overlap, so the
+// page count is not bounded by the board size — this is what stops one
+// pathological tenant from eating a sweep.
+const SPLIT_PAGE_BUDGET_FACTOR = 5;
+
 // Workday returns postings newest-first, so pagination can stop once a
 // page's oldest *dated* posting is well past --since — no point paying for
 // (and rate-limit-risking) pages that are entirely stale. Only unambiguous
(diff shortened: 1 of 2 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/providers/workday-facet-split.test.mjs`; `providers/workday.mjs` is outside my area, read for context.

The sentinel is the clamped tenant: total 2000 against faceted counts summing to 2700, which a naive paginator stops at the ceiling and drops the overflow. The test verifies the split is additive, dedups overlapping slices, re-splits still-clamped slices on a fresh facet, bounds depth and page budget, tags `workdayTruncated`, and never facets a healthy tenant. Hermetic (mocked fetchJson, no-op sleep, fetchText throws). Green on CI. Approving.

santifer (changes requested):
This is a genuinely good piece of provider engineering, @bigguy6883: the facet-split with hard budgets (depth, slices, page factor), the strictly-additive framing, and especially the honesty about skew (the DSG 6,564-in-one-jobFamily observation) show you understood the problem rather than patching the symptom. One precise thing before merge, and it's about the PR's own honesty standard:

**A slice that dies mid-fetch is counted as complete.** In `split()`, the early return is `if (!result.clamped) return;` — but `clamped` is only ever true when `stopReason === 'complete'`. A slice whose pagination hit a WAF/fetch failure comes back with `stopReason: 'fetch-error'` and `clamped: false`, so `split()` absorbs its partial jobs and returns without setting `splitIncomplete`. Net effect: a board that lost a slice mid-fetch can print its recovery line *without* the `(still incomplete)` tag: exactly the "reported as complete when it wasn't" failure the PR exists to prevent.

The fix is one condition: after `absorb(result.jobs)`, treat `result.stopReason === 'fetch-error'` (and arguably `'cap'`) as `splitIncomplete = true` before the `!result.clamped` return. `early-stop` should stay exempt: stopping past the `--since` window is the slice being genuinely done for the sweep's purposes.

With that condition and a test case pinning it (a slice stub returning `stopReason: 'fetch-error'` → recovery line carries the tag), this merges. Happy to do the final pass same-day when it's in.

Scott-Emberson (approved):
Re-reviewed at `416b49b5`. My approval sat on `80a32872`, and both commits since land inside `tests/providers/workday-facet-split.test.mjs`, so it needed a fresh look rather than carrying over.

The delta does what santifer asked. `providers/workday.mjs:473` now sets `splitIncomplete` on `stopReason === 'fetch-error'`, and on `'cap'` only when the query was not clamped. The comment gives the reason for that second half: a clamped query reports `total` at the ceiling, so it always ends at the cap, and tagging it would put "(still incomplete)" on every clamped board and mean nothing.

My part is the test half, and it holds up. Four new checks, none of them shape checks. I mutation tested them by replacing the new condition at `providers/workday.mjs:473` with `if (false) { splitIncomplete = true; }` and re-running the suite:

```
❌ slice that failed mid-pagination produced recovery line "⚠️  workday: Acme offset-clamped at 2000 — recovered 2040 jobs via 2 facet slices", expected it to carry "(still incomplete)"
❌ a split whose slice hit fetch-error should tag jobs.workdayTruncated, not report the board complete
❌ a split whose slice hit the page cap should tag jobs.workdayTruncated
✅ workday.fetch() stops a slice at the first page past the --since window instead of paginating its full total
✅ workday.fetch() leaves a slice that early-stopped past the --since window untagged
```

Three fail, and the two early-stop checks stay green. That is the right split: they pin the exemption, which the new condition deliberately does not touch. The assertions are load bearing.

The request count at `tests/providers/workday-facet-split.test.mjs:419`, `earlyStopCalls.length === 4`, is the piece I would have asked for had it been missing. Without it, "leaves an early-stopped slice untagged" passes just as happily when early-stop never fires and the slices simply run out of pages, which is the classic false pass. Counting requests turns it into proof that the exempt path was actually reached. Same for the note on the `Posted 300 Days Ago` fixture: pinning why the value is bounded (`parsePostedOn` reads "300+ Days Ago" as undated, which would leave the page dateless and early-stop unreachable) saves the next reader from rediscovering it.

Ran the full suite at `416b49b5`: 23 checks, 0 failed, exit 0, no network. CI is green on all ten checks.

Approving from my side. One thing I am explicitly not judging: CodeRabbit's `providers/workday.mjs:110-119` comment about overlapping facet counts summing above a healthy `total` of exactly 2000. That is a `providers/` question rather than a test one, so it is not mine to call, but it is worth a glance before merge.

Scott-Emberson (commented):
Re-reviewed at head d7bc2503b3 (moved since my earlier approval). The delta is purely additive: one new "slice fails on page 0" block appended to tests/providers/workday-facet-split.test.mjs (+51 lines, zero deletions), all previously-reviewed assertions unchanged. The new block throws on a `jobFamily=a` page-0 query and asserts the unfaceted crawl still survives (>= 2000 postings), the remaining `jobFamily=b` slice is still tried, `workdayTruncated === true`, and the recovery line carries "(still incomplete)". Those tie to the real production guard (providers/workday.mjs wraps the slice runQuery in try/catch and sets the incomplete flag), so it's not tautological.

Scott-Emberson (approved):

PR #3317, opened 2026-08-25
test: add tests for providers/local-parser.mjs

### LINKED ISSUE #3292: test: add tests for providers/local-parser.mjs, the last untested provider
**Time:** ~1-2h · **Difficulty:** Beginner (Node, no framework)

Every provider under `providers/` has a test file in `tests/providers/` except `local-parser.mjs`. It parses locally saved job pages, so its tests need no network: fixtures in, parsed postings out.

**The exact pattern to mirror:** any existing `tests/providers/*.test.mjs` (e.g. `recruitee.test.mjs`) — plain Node, `pass`/`fail` helpers from `tests/helpers.mjs`, auto-discovered by `test-all.mjs` with no registration. `docs/local-parser-cookbook.md` describes the expected behavior to assert.

**Acceptance:** a `tests/providers/local-parser.test.mjs` covering the exported entry point with at least one realistic fixture and one negative control, `node test-all.mjs` green.

Comment `/assign` below to claim this issue.

Monu01123:
Fixes #3292

Adds the missing test suite for `providers/local-parser.mjs`, bringing it in line with the other providers.

### Changes made
- Added `tests/providers/_fixture-local-parser.mjs` as a realistic local-parser dummy script to validate execution, argument interpolation, and stdout handling.
- Added `tests/providers/local-parser.test.mjs` to comprehensively test `detect()` and `fetch()`.

### Acceptance Criteria met
- Covers the exported entry point with realistic fixtures and negative controls (escaped paths, malformed JSON, flag injection).
- Asserts expected behavior documented in `docs/local-parser-cookbook.md`.
- `node test-all.mjs` runs these successfully.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for the local parser provider.
  * Verified configuration handling, fallback behavior, job extraction, normalization, and argument interpolation.
  * Added checks for injection prevention, wrapped job responses, relative URLs, invalid JSON, and failure reporting.
  * Added test fixtures covering valid, invalid, and incomplete job data.

### DIFF
--- a/tests/providers/_fixture-local-parser.mjs
+++ b/tests/providers/_fixture-local-parser.mjs
@@ -1,0 +1,28 @@
+// Mock script for testing local-parser.mjs
+
+const inputArgs = process.argv.slice(2);
+
+// If the first argument is "echo", return a job containing all args
+if (inputArgs[0] === 'echo') {
+  console.log(JSON.stringify([{ title: inputArgs.join(' '), url: 'https://example.com/job' }]));
+  process.exit(0);
+}
+
+// If the first argument is "envelope-jobs", test { jobs: [...] } wrapper
+if (inputArgs[0] === 'envelope-jobs') {
+  console.log(JSON.stringify({ jobs: [{ title: 'Envelope Job', url: '/job2' }] }));
+  process.exit(0);
+}
+
+// If the first argument is "invalid", print invalid JSON
+if (inputArgs[0] === 'invalid') {
+  console.log("NOT JSON");
+  process.exit(0);
+}
+
+// Default payload
+console.log(JSON.stringify([
+  { title: 'Standard Job', url: 'https://example.com/job1', location: ['Remote', 'NY'] },
+  { title: 'No URL' }, // Should be dropped
+  { url: 'https://example.com/notitle' } // Should be dropped
+]));
--- a/tests/providers/local-parser.test.mjs
+++ b/tests/providers/local-parser.test.mjs
@@ -1,0 +1,152 @@
+import { pass, fail, ROOT } from '../helpers.mjs';
+import { join } from 'path';
+import { pathToFileURL } from 'url';
+
+console.log('\nProvider — local-parser');
+
+try {
+  const localParserModule = await import(pathToFileURL(join(ROOT, 'providers/local-parser.mjs')).href);
+  const localParser = localParserModule.default;
+
+  // 1. Identity
+  if (localParser.id === 'local-parser') pass('localParser.id is "local-parser"');
+  else fail(`localParser.id is ${JSON.stringify(localParser.id)}`);
+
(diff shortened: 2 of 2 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/providers/local-parser.test.mjs` and its fixture; `providers/local-parser.mjs` is outside my area, read for context.

Every assertion traces to real behavior at head: identity, detect happy/fallback and three negatives, drop-invalid on fetch, location normalization, arg interpolation, both injection guards, the envelope plus relative-URL case, and invalid JSON. The sentinels are behaviour-specific (exact returns and exact error substrings), so each reddens on regression rather than merely passing, and the error-message assertions match source verbatim. Hermetic (in-repo node fixture, no network), neighbour-style. Green on CI. Approving.

BigBrown10 (commented):
Reviewed the diff against `providers/local-parser.mjs` on main, assertion by assertion - this is genuinely good work for a first suite, @Monu01123. Welcome!

**What I verified against the source:**

- The interpreter lane contract is pinned correctly: `args[0] !== scriptPath` rejection, in-repo script resolution, and the `detect()` null postures (no command, interpreter without script) all trace to `resolveInvocation()`.
- The echo-interpolation expectation accounts for the script path occupying argv slots 0-1, so the joined title matches exactly what the child sees.
- Both injection cases throw *before* spawn (expansion happens eagerly in `buildParserArgs`), so these are true pre-execution rejection tests, not crash-after-spawn tests. That distinction matters and your structure captures it.
- The envelope test also pins relative-URL resolution against a trailing-slash base (`https://example.com/base/`) - subtle and correct.

CI green on all three OS lanes, including Windows where `realpathSync` + PATH-resolved interpreters are the risky bits. Nice.

**One suggestion that would sharpen the escape test** (optional, not blocking):

`detect({ parser: { command: '../escaped' } })` currently passes for a conflated reason: `realpathSync` throws ENOENT because the path doesn't exist, which lands in the same `catch` as a genuine root-escape rejection. To pin the actual escape class, point at something that *exists* outside the root:

```js
if (localParser.detect({ parser: { command: '../package.json' } }) === null) {
  pass('localParser.detect() returns null for an existing path outside the project root');
}
```

That distinguishes "rejected because it escapes" from "rejected because it is missing" - two failure classes the guard deliberately separates.

**Two small coverage notes for a follow-up if you feel like it** (no need here):

1. The `{ results: [...] }` envelope variant of the JSON contract is untested (only `jobs[]`).
2. `parser.timeout_ms` / `max_buffer_bytes` config paths have no assertions yet.

Either way: this closes #3292's acceptance criteria as written - realistic fixture, negative controls, house style with `pass`/`fail`, auto-discovered. Nice first contribution.

Issue #3223, opened 2026-08-23
Native-speaker check wanted: the human-in-the-loop row in 11 languages (a940028a)


santifer:
I merged `a940028a` today, which restored the absolute human-in-the-loop wording in eleven translated READMEs. **I don't speak nine of those languages**, so I want the wording checked by people who do.

### What happened

Every README's features table carries the project's central promise: the system never submits an application. Eleven translations had weakened it to "never submits **automatically**", which is a different claim — it leaves room for the system to submit some other way. The drift appeared independently in eight languages and always in the same direction, because adding the adverb reads naturally to a translator.

Arabic was worse: `النظام لا يقوم بالتقديم التلقائي أبدًا دون إذنك` said "never submits automatically **without your permission**", turning an absolute guarantee into a conditional one.

### Why I could fix it without inventing anything

**Each README's own FAQ already stated the guarantee correctly** — only the table row had drifted, which suggests the two surfaces were translated at different times. So in seven languages the fix was deleting one word, and in the three that needed a clause added (`de`, `ru`, `ar`) I copied it verbatim from that same file's FAQ. Every word is its original translator's.

### What I'm asking for

Deleting a word can't change meaning, but it **can** leave a sentence that reads stiffly, and copying a clause from a FAQ into a table row can land in the wrong register. That's what I can't check.

If you read any of these, please look at the human-in-the-loop row and say whether it sounds natural. A comment is enough; a PR only if you'd rather.

| language | file | changed |
|---|---|---|
| العربية | `README.ar.md` | clause replaced from its own FAQ |
| Deutsch | `README.de.md` | clause replaced from its own FAQ |
| Русский | `README.ru.md` | clause replaced from its own FAQ |
| Français | `README.fr.md` | one word deleted |
| 日本語 | `README.ja.md` | one word deleted |
| 한국어 | `README.ko-KR.md` | one word deleted |
| 简体中文 | `README.cn.md` | one word deleted |
| 繁體中文 | `README.zh-TW.md` | one word deleted |
| தமிழ் | `README.ta.md` | one word deleted |
| Português (BR) | `README.pt-BR.md` | one word deleted |
| Español | `README.es.md` | `aplicación` → `candidatura`, accent fixed |

**Arabic is the one I'd most like a second pair of eyes on**, because it needed a real replacement rather than a deletion and it's the language where I'd least notice an awkward result.

### The rule this row now carries

Each row ends with an invisible marker that states the constraint at the point of editing:

```html
[...]

PR #3321, opened 2026-08-25
fix(fact-gate): a non-English CV inflating headcount returns 'pass' — say so instead (#3320)

### LINKED ISSUE #3320: The fabrication gate checks counts only in English, so a non-English CV inflating headcount returns 'pass'
## Problem

`verify-cv-facts.mjs` is the last check before a generated document goes out — `generate-cover-letter.mjs` calls `assertFacts` on every cover letter, and the CLI gates CVs. Its count extractor is keyed on `METRIC_NOUNS`, an English word list, and `COUNT_CLAIM_RE`'s modifier window is `[A-Za-z]`.

Percentages, currency and multipliers are language-neutral and still checked everywhere. **A count is checked only in English.** Same inflation, six languages, sources saying 20 and 2 in every case:

```
EN | claims: ["45 staff","3 facilities"]  | invented: ["45 staff","3 facilities"]
ES | claims: []                           | invented: []
DE | claims: []                           | invented: []
JA | claims: []                           | invented: []
TR | claims: []                           | invented: []
FR | claims: ["3 sites"]                  | invented: ["3 sites"]
```

French catches one only because *sites* is spelled the same in both languages — which is a good illustration that the coverage is coincidental rather than partial.

## Why this is the sharp end

Counts are the class this file's own `METRIC_NOUNS` comment singles out, in the note added when non-software English CVs were found slipping through:

> "Managed 45 staff" against a source saying 20 passed the gate silently, which is the exact fabrication class this script exists to catch.

For a CV in one of the market languages the project ships modes for, that sentence is true of **every** count, not just the ones outside the list. And this is not an edge case for this project — AGENTS.md makes non-English output first-class:

> `language.output` controls **human-facing output**: reports, tracker notes, PDFs, cover letters, outreach, interview prep, form answers, any user-visible prose.

with market mode sets shipped for de / fr / ar / ja / tr / hi.

## Reproduction

```js
import { auditClaims, metricClaims } from './verify-cv-facts.mjs';
auditClaims('Gestioné 45 empleados en 3 instalaciones.',
            'Gestioné 20 empleados en 2 instalaciones.');
// → { invented: [], forbidden: [] }        verdict: pass
```

## What I think the fix is — and what it isn't

The real answer is a lexicon those languages are in. I have **not** proposed that here: it is 70-odd nouns × 6+ languages, it needs native speakers (cf. #3223), and guessing at it is exactly the kind of change that looks done and is not.

What is unambiguous is that the gate should not return a confident `pass` on a document it could not read. The repo already has this pattern twice — `jd-skill-gap.mjs`'s `diagnoseExtraction()` and `story-provenance-check.mjs`'s `diagnose()`, the latter documented as existing so *"an empty/near-empty result isn't misread as 'scanned and clean'"*.

So the PR adds `diagnoseCoverage()`: when a document contains two or more count-shaped spans and the extractor recognised **no** count at all, the result carries a `coverage` explanation and the verdict becomes `warn` rather than `pass`. It never creates a block and never masks one — blocking instead would fail every non-English document, trading a silent gap for a wall.

**It is a partial answer and the PR says so.** It closes the silent pass for space-delimited languages (de, es, tr, pt, it, pl, ru, …). It does *not* cover ja/zh: those put digits flush against surrounding text (`3拠点で45名`), so the second number is preceded by a letter and the detector cannot locate it; relaxing that would match digits inside Latin identifiers, so it needs script-aware segmentation rather than a looser regex. Both that and the French coincidence are asserted by tests so they are decisions on record rather than surprises.

Happy to follow up with a lexicon PR for whichever languages you want, if you'd rather go straight at the root — I just didn't want to invent vocabulary for six markets unreviewed.

abankar1:
Fixes #3320.

`verify-cv-facts.mjs` is the last check before a generated document goes out (`generate-cover-letter.mjs` calls `assertFacts` on every cover letter; the CLI gates CVs). Its count extractor is keyed on `METRIC_NOUNS`, an English word list. Percentages, currency and multipliers are language-neutral and still checked everywhere — **a count is checked only in English.**

Same inflation, sources saying 20 and 2 in every case:

| | verdict before | verdict after |
|---|---|---|
| `Managed 45 staff across 3 facilities.` | block | block |
| `Gestioné 45 empleados en 3 instalaciones.` | **pass** | warn + coverage |
| `Leitete 45 Mitarbeiter an 3 Standorten.` | **pass** | warn + coverage |
| `3 tesiste 45 çalışanı yönetti.` | **pass** | warn + coverage |
| `Geri 45 funcionários em 3 unidades.` | **pass** | warn + coverage |

Counts are the class this file's own `METRIC_NOUNS` comment singles out — *"'Managed 45 staff' against a source saying 20 passed the gate silently, which is the exact fabrication class this script exists to catch."* For a non-English CV that is true of every count, not only the ones outside the list.

## What this does — and deliberately does not

The real answer is a lexicon those languages are in. I have **not** attempted that: it is ~70 nouns × 6+ languages and needs native speakers (cf. #3223). Inventing that vocabulary unreviewed would look done and not be.

What is unambiguous is that the gate should not return a confident `pass` on a document it could not read. `diagnoseCoverage()` reports the gap instead — the same choice `jd-skill-gap.mjs`'s `diagnoseExtraction()` and `story-provenance-check.mjs`'s `diagnose()` make, the latter documented as existing so *"an empty/near-empty result isn't misread as 'scanned and clean'"*.

It only ever turns a would-be `pass` into `warn`. It never creates a block and never masks one — blocking would fail every non-English document, trading a silent gap for a wall.

## Calibration, because this runs on every generated document

Fires only when there are **two or more** count-shaped spans and **no** recognised count at all. Spans the language-neutral patterns already own are excluded, derived from `SIMPLE_CLAIM_PATTERNS` so the two cannot drift.

That exclusion exists because the **existing suite caught a false positive** before this shipped:

```
❌ source-backed currency metrics were blocked:
   coverage.spans: ["120k and closed", "90,000 deal"]
```

`$120k and closed a $90,000 deal` is currency followed by prose — it reads as two counts to a detector with no lexicon, but those amounts *are* checked. Years are excluded for the same reason: `Led the 2024 migration` is the shape of a count and none of the meaning, and without that filter it fires on essentially every document.

## Two blind spots, asserted rather than described

- **Coincidental coverage.** French `3 sites` matches the English noun, so one recognised count silences the warning for a French CV whose other counts are invisible.
- **CJK.** `3拠点で45名` puts digits flush against the text, so the second number is preceded by a letter and the detector cannot locate it. Relaxing that would match digits inside Latin identifiers, so it needs script-aware segmentation, not a looser regex.

Both have tests. If a later change makes either fire, that is an improvement and the test should be updated deliberately — not something someone discovers by surprise.

## Verification

Reverting only the wiring, with tests kept:

```
✖ the same inflation in a space-delimited language is reported as unchecked
✖ the signal never creates or masks a block
ℹ pass 5  ℹ fail 2
```

`node verify-cv-facts.mjs --self-test` → 62 passed, 0 failed.
`node test-all.mjs --quick` → 5378 passed, 0 failed (3 pre-existing warnings).

Happy to follow up with a lexicon PR for whichever languages you want if you'd rather go at the root directly.

## Summary by CodeRabbit

- **New Features**
  - Added broader detection for count-based claims across languages.
  - Added coverage diagnostics to identify claims that cannot be fully verified.
  - CLI and JSON results now include coverage warnings when verification is incomplete.

- **Bug Fixes**
  - Unsupported-language counts now produce warnings instead of being incorrectly treated as verified.
  - Coverage warnings no longer hide or override blocking verification failures.

- **Tests**
  - Added coverage for multilingual counts, clean metrics and years, coincidental matches, and known CJK detection gaps.

### DIFF
--- a/tests/fact-gate-language-coverage.test.mjs
+++ b/tests/fact-gate-language-coverage.test.mjs
@@ -1,0 +1,108 @@
+// tests/fact-gate-language-coverage.test.mjs — the fabrication gate must not
+// report a confident pass on a document it could not read.
+//
+// verify-cv-facts.mjs is the last check before a generated CV or cover letter
+// goes out (generate-cover-letter.mjs calls assertFacts on every one). Its
+// count extractor is keyed on METRIC_NOUNS, an English word list, and
+// COUNT_CLAIM_RE's modifier window is `[A-Za-z]`. Percentages, currency and
+// multipliers are language-neutral and still checked everywhere — a COUNT is
+// checked only in English.
+//
+// That is not an edge case for this project. AGENTS.md's `language.output`
+// governs "reports, tracker notes, PDFs, cover letters ... any user-visible
+// prose", and the repo ships market modes for de/fr/ar/ja/tr/hi. And counts are
+// the class the file's own METRIC_NOUNS comment singles out:
+//
+//   "Managed 45 staff against a source saying 20 passed the gate silently,
+//    which is the exact fabrication class this script exists to catch."
+//
+// Run:  node --test tests/fact-gate-language-coverage.test.mjs
+
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { verifyFacts, diagnoseCoverage } from '../verify-cv-facts.mjs';
+
+// No sources and no config: this asks what the gate SEES, not what it allows.
+const BARE = { sourcePaths: [], configPath: 'config/__no_such_config__.json' };
+const verdictOf = (text) => verifyFacts(text, BARE).verdict;
+
+test('an English count inflation is caught, as it always was', () => {
+  const r = verifyFacts('Managed 45 staff across 3 facilities.', BARE);
+  assert.equal(r.verdict, 'block');
+  assert.deepEqual([...r.invented].sort(), ['3 facilities', '45 staff']);
+  assert.equal(r.coverage, null, 'a document the extractor read needs no coverage warning');
+});
+
+test('the same inflation in a space-delimited language is reported as unchecked', () => {
+  for (const [lang, text] of [
+    ['es', 'Gestioné 45 empleados en 3 instalaciones.'],
+    ['de', 'Leitete 45 Mitarbeiter an 3 Standorten.'],
+    ['tr', '3 tesiste 45 çalışanı yönetti.'],
+    ['pt', 'Geri 45 funcionários em 3 unidades.'],
+  ]) {
+    const r = verifyFacts(text, BARE);
+    assert.ok(r.coverage, `${lang}: no coverage signal — this document reads as scanned and clean`);
+    assert.equal(r.coverage.reason, 'no-count-claims-recognized');
+    assert.ok(r.coverage.spans.length >= 2, `${lang}: expected the count spans to be located`);
+    assert.equal(r.verdict, 'warn', `${lang}: a document the gate could not read must not verdict 'pass'`);
+  }
+});
+
+test('a clean English document is untouched — no new noise', () => {
+  // The signal is added to a gate every generated document already runs, so a
+  // false fire is a real cost. These must all stay silent.
+  for (const text of [
+    'Senior Engineer at Acme since 2019. Led the 2024 platform migration.',
+    'Cut p99 latency by 30% and infrastructure spend by $1.2M.',
+    'Mentored 6 engineers and ran 4 hiring loops.',
+    'Managed 20 staff across 2 facilities.',
+    // Currency followed by prose reads as "digits, then a word" to a detector
+    // that has no lexicon — but these ARE checked, in every language, so
+    // reporting them as unread is a false alarm. Caught by the existing suite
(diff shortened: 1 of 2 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/fact-gate-language-coverage.test.mjs`; `verify-cv-facts.mjs` is fabrication-detection logic outside my area, read carefully for context.

Two-sided sentinel: it catches the non-English inflation (es/de/tr/pt drive the coverage signal to a warn verdict with the spans located) AND it does not falsely flag clean English, with years excluded and the language-neutral block still firing. The known blind spots (a coincidental French "sites", CJK) are pinned as tests so a future change surfaces there rather than silently. Worth noting because it is a source-of-truth guard: the coverage signal only ever downgrades pass to warn, it never masks or creates a block, and that is asserted and confirmed in source. Hermetic, module-import only. Green on CI. Approving.

Issue #3315, opened 2026-08-25
feat: /calibrate — check your evaluation scores against your real outcomes (the learning loop's feedback half)


santifer:
Part of #1724. The last unbuilt piece of the learning loop.

## Problem

career-ops now records what happened to every application (`outcome.mjs`, #1722) and detects rejection patterns (`analyze-patterns.mjs`). But nothing answers the question the whole loop exists for: **does the A-H evaluation actually predict success *for you*?** A user six weeks into a search has a dataset (scores on one side, interviews/offers/silence on the other) and no tool that reads it.

## Design (deterministic, advisory, local)

A new `calibrate.mjs` (surfaced as `/calibrate`) that reads only local data — `data/outcomes/`, the tracker, and linked reports — and produces a calibration report:

1. **Calibration curve**: your applications bucketed by global score band (e.g. <3.5 / 3.5-4.0 / 4.0-4.5 / 4.5+) vs what actually happened (response rate, interview rate, offer rate). Honest minimum-n floors like `analyze-patterns` `--min-threshold`: no percentages on 3 data points.
2. **Dimension check**: which report dimensions (CV match, red flags, comp fit…) separated your interviews from your silences, and which carried no signal for you.
3. **A suggestion, never an action**: if your 4.0+ evaluations convert no better than your 3.7s, the report says so and *suggests* what to consider — it never edits scoring rules, thresholds, or `modes/_shared.md`. The scoring system stays global; what's personal is the evidence about how it performs on your search.
4. **Diff vs your previous calibration**: rerun after 20 more applications and see what moved.

## Non-goals

- No changes to `modes/_shared.md` or any scoring rule (that would need an RFC; this feature is deliberately scoped to stay advisory).
- No network, no keys, no LLM calls: pure deterministic parsing, same class as `analyze-patterns.mjs`.
- No auto-tuning. Human reads, human decides. As everywhere else in career-ops.

## Credit

The outcome-mining idea was sharpened by studying [MadsLorentzen/ai-job-search](https://github.com/MadsLorentzen/ai-job-search)'s setup flow, which mines recorded outcomes on re-runs. Good ideas travel.

Implementation PR incoming from the maintainer side; review and better ideas welcome.

PR #3323, opened 2026-08-25
fix(calibrate): reads 7 of the 14 outcome types outcome.mjs writes (#3322)

### LINKED ISSUE #3322: calibrate reads 7 of the 14 outcome types outcome.mjs writes — a declined offer is dropped, and a journal reports its happiest entry
## Problem

`outcome.mjs` accepts **fourteen** spellings of an outcome type and writes whichever one was typed verbatim into the journal:

```js
// outcome.mjs
const OUTCOME_MAP = {
  interview_progress, stage_reached, interview, offer_received, offer, hired,
  accepted, offer_declined, declined, rejected, rejection, no_response,
  ghosted, interview_only
};
...
- **Outcome Type**: ${normalizedOutcomeKey}
```

`calibrate.mjs` (#3315, merged yesterday) reads that journal against a private **seven**-entry map. The other seven resolve to nothing:

```js
parseOutcomeJournal('- **Outcome Type**: declined')  // → latestType: null
parseOutcomeJournal('- **Outcome Type**: ghosted')   // → latestType: null
```

## Two consequences, and the second is the one that matters

**1. A declined offer is dropped from calibration entirely.**

`declined` produces no verdict, so `computeCalibration` falls back to the tracker status — which `outcome.mjs` set to `Discarded`, and calibrate reads `Discarded` as *"never applied — outside calibration's population"* and `continue`s. The row is not counted as resolved, not counted as in-flight, not counted at all.

A declined offer is the strongest evidence a search can produce that the scoring picked well. It vanishes. Same for `ghosted`, which is a real negative data point.

**2. The journal reports its happiest historical moment — which the docstring forbids.**

`latestType` only advances when the type is recognized, so an unrecognized final entry leaves the previous one standing:

```
interview_progress → ghosted     reports  interview_progress
interview_progress → rejection   reports  interview_progress
offer_received     → declined    reports  offer_received
```

From `parseOutcomeJournal`'s own docstring:

> The journal is append-only (outcome.mjs), so the LAST `## Entry:` block is the current truth — an application that went interview_progress and later rejected must read as rejected, **not as its happiest historical moment**.

For half the vocabulary, it reads as exactly that. An application that reached interviews and was then rejected is scored as still progressing — inflating precisely the interview-conversion rate the tool exists to measure.

## Why the copy exists at all

Not carelessness: `outcome.mjs` is a top-level CLI that **exits on import**, so a reader cannot reuse its map and has to write a second one.

```
$ node -e "await import('./outcome.mjs')"
Usage: node outcome.mjs <report#|company> <outcome_type> [options]
$ echo $?
1
```

Which is why the fix is to give the vocabulary one home rather than to extend the copy.

## Fix

`lib/outcome-types.mjs` holds `OUTCOME_MAP` (moved verbatim), the canonical seven that `outcome.mjs`'s own USAGE advertises, and `canonicalOutcome()`. `outcome.mjs` imports it; `calibrate.mjs` resolves through it.

PR follows, with a census derived from `OUTCOME_MAP` — so a synonym added to the writer is covered the day it is added — asserted in both directions, plus a guard that the widening does not become "accept anything".

One detail worth flagging for anyone adding a lib file: `lib/outcome-types.mjs` needs a `SYSTEM_PATHS` entry or `apply` ships the updated `outcome.mjs` to an existing install without the module it now imports. `validate-system-paths-coverage.mjs` does catch it, but only once the file is tracked — it reported `OK: 1198 tracked files covered` while the file was still untracked, then flagged it correctly after `git add`.

abankar1:
Fixes #3322.

`outcome.mjs` accepts **14** spellings of an outcome type and writes whichever was typed verbatim into the journal. `calibrate.mjs` (#3315) reads it against a private **7**-entry map.

## Two consequences

**A declined offer is dropped from calibration entirely.** `declined` resolves to nothing, so the row falls back to its tracker status — `Discarded`, which calibrate reads as *"never applied — outside calibration's population"* and skips. Not resolved, not in-flight, not counted. A declined offer is the strongest evidence a search can produce that the scoring picked well.

**The journal reports its happiest historical moment.** `latestType` only advanced on a recognized type, so an unrecognized final entry left the previous one standing:

| journal | before | after |
|---|---|---|
| `interview_progress` → `ghosted` | `interview_progress` | `no_response` |
| `interview_progress` → `rejection` | `interview_progress` | `rejected` |
| `offer_received` → `declined` | `offer_received` | `offer_declined` |
| `declined` | `null` | `offer_declined` |
| `ghosted` | `null` | `no_response` |

`parseOutcomeJournal`'s own docstring forbids exactly this:

> the LAST `## Entry:` block is the current truth — an application that went interview_progress and later rejected must read as rejected, **not as its happiest historical moment**.

For half the vocabulary it read as exactly that, inflating the interview-conversion rate the tool exists to measure.

## Why the copy existed

`outcome.mjs` is a top-level CLI that **exits on import** (`node -e "await import('./outcome.mjs')"` → prints usage, exit 1), so a reader cannot reuse its map. The fix gives the vocabulary one home rather than extending the copy.

## Change (three commits)

1. **`lib/outcome-types.mjs`** — `OUTCOME_MAP` moved verbatim, plus the canonical seven `outcome.mjs`'s USAGE advertises and `canonicalOutcome()`. `outcome.mjs` imports it; usage line and error text unchanged.
2. **`calibrate.mjs`** resolves through it, and no longer leaves a stale `latestType` standing.
3. **The census**, derived from `OUTCOME_MAP` so a synonym added to the writer is covered the day it is added.

The alias table is written out rather than derived from the `{state, defaultNote}` pairs. Deriving looks tighter and is wrong: `interview` shares a state with `interview_progress` but carries its own note, and `offer_declined`, `no_response` and their synonyms all share the state `Discarded` while meaning opposite things about the outcome. **State is a tracker position, not a result.** A load-time coverage check plus the drift test is what prevents an omission.

## Verification

Reverting only the lookup in `parseOutcomeJournal`, tests kept:

```
✖ a journal written with an alias resolves to its canonical meaning
✖ the LAST entry wins even when it is an alias
✖ a hand-edited journal spelling still resolves
✖ a declined offer counts as reaching an offer, whichever spelling recorded it
ℹ pass 3  ℹ fail 4
```

The three that pass either way are the guards: the census, the exact-canonical-set assertion, and *an unrecognized type is still unrecognized* — the widening must not become "accept anything".

`node calibrate.mjs --self-test` → all checks passed.
`node outcome.mjs --help` / invalid-type error → unchanged, all 14 types still listed.
`node test-all.mjs --quick` → 5380 passed, 0 failed (3 pre-existing warnings).

## Updater registration

`lib/outcome-types.mjs` is added to `SYSTEM_PATHS` — without it, `apply` ships the updated `outcome.mjs` to an existing install without the module it now imports. Confirmed by removing the entry and re-running `validate-system-paths-coverage.mjs`:

```
Coverage gap — tracked files not in SYSTEM_PATHS or USER_PATHS:
  lib/outcome-types.mjs
```

Worth knowing for the next person adding a lib file: that guard reads `git ls-files`, so it reported `OK: 1198 tracked files covered` while the file was still untracked and only flagged it after `git add`. Not a bug — but it means the check is worth re-running *after* staging, not before.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved outcome recognition across calibration and validation workflows.
  * Added consistent handling for aliases, capitalization differences, and unknown outcomes.
  * Ensured the most recent journal entry determines the final outcome.
  * Prevented stale outcomes from being retained when the final entry is unrecognized.
  * Correctly calibrates declined offers.

* **Chores**
  * Included updated outcome definitions in system updates.

### DIFF
--- a/calibrate.mjs
+++ b/calibrate.mjs
@@ -94,15 +94,15 @@ export function parseOutcomeJournal(text) {
   const feedback = [];
   for (const entry of entries) {
     const typeMatch = entry.match(/^- \*\*Outcome Type\*\*: *(\S+)/m);
-    // Resolved through the shared vocabulary, not looked up directly. An
-    // UNRECOGNIZED type must not leave `latestType` at its previous value
-    // either: that is precisely how a journal ending in `ghosted` reported the
-    // `interview_progress` above it — the happiest historical moment this
+    // Assigned UNCONDITIONALLY when the field is present, including when it
+    // resolves to null. The last entry is the truth, so a final entry this
+    // vocabulary cannot read must not leave the previous one standing —
+    // `interview_progress` then `withdrawn_by_employer` is not an application
+    // still progressing through interviews. Clearing it falls through to the
+    // tracker status, which is the honest answer for an outcome we cannot read;
+    // keeping the earlier one is the happiest-historical-moment behaviour this
     // function's contract forbids.
-    if (typeMatch) {
-      const canonical = canonicalOutcome(typeMatch[1]);
-      if (canonical) latestType = canonical;
-    }
+    if (typeMatch) latestType = canonicalOutcome(typeMatch[1]);
     // Feedback is a blockquote under "Verbatim Feedback"; "None recorded" is
     // the explicit empty marker outcome.mjs writes, not user content.
     const fbMatch = entry.match(/- \*\*Verbatim Feedback\*\*:\n((?:> .*\n?)+)/);
--- a/calibrate.mjs
+++ b/calibrate.mjs
@@ -35,6 +35,7 @@ import { fileURLToPath } from 'url';
 import { dirname } from 'path';
 import { parseTrackerRow, resolveColumns, isSeparatorRow, isHeaderRow } from './tracker-parse.mjs';
 import { resolveTrackerPath, resolveWorkspaceRoot } from './tracker-utils.mjs';
+import { canonicalOutcome } from './lib/outcome-types.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
 
@@ -45,7 +46,11 @@ const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
 // scoring picked a matchable role), and an offer is tier 2. `rejected` and
 // `no_response` are terminal negatives. `interview_only` means the process
 // ended after interviews with no offer — it still reached tier 1.
-const JOURNAL_OUTCOMES = {
+// Keyed by CANONICAL type only. outcome.mjs accepts fourteen spellings and
+// writes whichever one was typed straight into the journal, so every read goes
+// through canonicalOutcome() first — a private list of the seven canonical
+// names silently ignored the other seven (see lib/outcome-types.mjs).
+export const JOURNAL_OUTCOMES = {
   hired: { reachedInterview: true, reachedOffer: true, terminal: true },
   offer_received: { reachedInterview: true, reachedOffer: true, terminal: true },
   offer_declined: { reachedInterview: true, reachedOffer: true, terminal: true },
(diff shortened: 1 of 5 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/outcome-vocabulary-drift.test.mjs`; `calibrate.mjs` is outside my area, read for context.

The sentinel is exactly the drift this guards against: it pins calibrate to precisely the seven canonical journal outcomes, and the census test derives from the full fourteen-spelling writer map, so dropping any state or synonym on either side reddens. The alias, last-entry-wins and clears-earlier cases trace to the unconditional `canonicalOutcome` assignment, and the journal-over-tracker override backs the declined-offer band. Hermetic (in-memory journal strings, pure computeCalibration, no filesystem), its own file per the repo rule. Green on CI. Approving.

PR #3325, opened 2026-08-25
fix(outcome): a tracker edit forks the append-only journal, and directory order decides which half counts (#3324)

### LINKED ISSUE #3324: outcome.mjs rebuilds the journal directory from the current role text, so a tracker edit splits an append-only journal in half
## Problem

`outcome.mjs` writes an application's outcome journal and artifacts to `data/outcomes/{num}_{company_slug}_{role_slug}/`, and rebuilds that name from the tracker row's **current** text on every call:

```js
const companySlug = slugify(matchedRow.company);
const roleSlug = slugify(matchedRow.role);
const outcomeDir = join(repoRoot, 'data', 'outcomes', `${matchedRow.num}_${companySlug}_${roleSlug}`);
```

So an ordinary tracker edit between two recordings splits the journal. Reproduced against the real CLI:

```
$ node outcome.mjs 1 interview_progress --stage "Tech Screen"
✅ Recorded ... in data/outcomes/1_acme_senior-backend-engineer

# user normalises the title in the tracker: "Senior Backend Engineer" → "Sr. Backend Engineer"

$ node outcome.mjs 1 rejected
✅ Recorded ... in data/outcomes/1_acme_sr-backend-engineer

$ ls data/outcomes
1_acme_senior-backend-engineer     # interview_progress
1_acme_sr-backend-engineer         # rejected
```

One application, two half-journals.

## Why that is worse than untidy

The journal is **append-only**, and its readers take the last entry as the truth — `parseOutcomeJournal`'s docstring:

> The journal is append-only (outcome.mjs), so the LAST `## Entry:` block is the current truth

But every reader keys on the leading `{num}_`, and none merges directories. `calibrate.mjs`'s `loadJournals` iterates entries and calls `journals.set(num, …)` per directory, so **the last one read wins — by alphabetical order.**

On the pair above:

```
band >=4.5: { n: 1, interviews: 0, offers: 0 }
```

The `interview_progress` entry is discarded, so an application with a recorded Tech Screen is scored as never having reached an interview — in the tool whose entire purpose is measuring interview conversion. And it goes the other way just as easily: with the edit reversed (`Sr.` → `Senior`), the *newest* entry is the one discarded and a stale outcome is reported.

Which half survives depends on how the slugs happen to sort, not on when anything happened.

Recorded artifacts split the same way — `submitted_cv.md`, `posting.pdf`, the cover letter — so the CV that was actually sent can end up in the directory nobody reads.

## Fix

The tracker row **number** is the identity; the slugs are a human-readable label on it. Resolve by number: reuse the row's existing directory when it has one, and only build a name from the current slugs on a first write.

A split that already exists is not repaired automatically — moving a user's recorded artifacts is not that command's job — but it should be reported, with the directories named, and both the writer and the reader should pick the most recently written journal rather than whichever sorts first.

PR follows. The first test drives the real CLI twice with a tracker edit in between and fails on current `main`; the ordering tests are each written so sort order and the right answer disagree, and `1_` is checked against `12_`/`100_` so a numeric prefix can't merge two unrelated applications.

abankar1:
Fixes #3324.

`outcome.mjs` rebuilt the journal directory name from the tracker row's **current** text on every call, so an ordinary tracker edit between two recordings split an append-only journal in half. Reproduced against the real CLI:

```
outcome.mjs 1 interview_progress   ->  1_acme_senior-backend-engineer
(user normalises "Senior Backend Engineer" to "Sr. Backend Engineer")
outcome.mjs 1 rejected             ->  1_acme_sr-backend-engineer
```

Every reader keys on the leading `{num}_` and none merges directories, so which half is seen comes down to alphabetical order. `calibrate.mjs` reported:

```
band >=4.5: { n: 1, interviews: 0, offers: 0 }
```

— an application with a recorded Tech Screen scored as never having reached an interview, in the tool whose purpose is measuring interview conversion. Reverse the edit (`Sr.` → `Senior`) and it discards the *newest* entry instead, reporting a stale outcome. Artifacts (`submitted_cv.md`, `posting.pdf`, the cover letter) split the same way, so the CV actually sent can land in the directory nobody reads.

## Change (three commits)

1. **`lib/outcome-dir.mjs`** — the row **number** is the identity, the slugs are a label on it. `resolveOutcomeDir()` reuses an existing directory and only builds a name from the current slugs on a first write. Registered in `SYSTEM_PATHS`, or `apply` ships an `outcome.mjs` importing a module the install doesn't have.
2. **`calibrate.mjs`** resolves one directory per tracker number via `outcomeDirsFor()`, ordered by journal mtime, instead of `journals.set()` per directory.
3. **Tests.**

A pre-existing split is **not** repaired automatically — moving a user's recorded artifacts is not this command's job — but it is now said out loud with the directories named, and the append goes to the most recently written one:

```
⚠ #1 has 2 outcome directories, so its journal is split: 1_acme_sr-…, 1_acme_senior-….
  Appending to 1_acme_sr-… (most recently written). Merge the entries into one
  directory to get a single history.
```

## Verification

After the fix, the same two commands:

```
$ ls data/outcomes
1_acme_senior-backend-engineer
$ grep -o 'Outcome Type\*\*: .*' data/outcomes/*/outcome.md
Outcome Type**: interview_progress
Outcome Type**: rejected
```

Reverting only the directory resolution, tests kept:

```
✖ a role edit between two recordings does not fork the journal
ℹ pass 5  ℹ fail 1
```

The first test drives the real CLI twice and asserts both the directory count **and** that the surviving journal holds both entries in order — a count of 1 with the first entry overwritten would satisfy the count and none of the point.

The ordering tests are each written so sort order and the right answer **disagree**: the first-sorting directory is made the older one, an artifacts-only directory is made the newer one, and `1_` is checked against `12_`/`100_` so a numeric prefix cannot merge two unrelated applications' histories.

`node calibrate.mjs --self-test` → passed · `node validate-system-paths-coverage.mjs` → OK, 1200 covered · `node test-all.mjs --quick` → 5380 passed, 0 failed (3 pre-existing warnings).

## Relationship to #3323

Both touch `outcome.mjs` and `calibrate.mjs`, in different regions (that one moves the outcome-type vocabulary; this one changes directory resolution). They should merge in either order — happy to rebase whichever lands second.

## Summary by CodeRabbit

- **Bug Fixes**
  - Outcome journals remain associated with the correct application when company or role details change.
  - When multiple journal directories exist, the most recently updated journal is selected reliably.
  - Artifact-only directories no longer override valid journals.
  - Split journals retain all matching locations for accurate reporting, with a warning instead of silently selecting one.
  - Missing outcome directories and whole-number application matches are handled correctly.


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/outcome-journal-fork.test.mjs`; `outcome.mjs` is outside my area, read for context.

The load-bearing case reddens on the real regression: before the fix a rebuild forked the application into two directories, and the test asserts exactly one (`assert.deepEqual(dirs, ['1_acme_senior-backend-engineer'])`). The journal regex, the slug derivation and the mtime-first resolution all trace to source. Hermetic (throwaway dirs, `CAREER_OPS_TRACKER` override, `rmSync` cleanup), and the tracker fixture uses the correct nine-column order with score before status. Green on CI. Approving.

PR #3326, opened 2026-08-25
fix(jobbankca): count raw entries for pagination

### LINKED ISSUE #3249: jobbankca.mjs: pagination's short-page stop counts normalized rows, not raw feed entries, and can silently drop later pages
In providers/jobbankca.mjs's fetch(), the end-of-results check at jobbankca.mjs:288 is:

```js
const parsed = parseJobBankFeed(xml);
for (const job of parsed) { ... }
if (parsed.length < PAGE_SIZE) break; // short page - end of this keyword's results
```

parsed.length is the count after parseJobBankFeed drops entries with no title or with a link that doesn't resolve to a trusted https://www.jobbank.gc.ca URL (cleanUrl()/linkHref()). It's not the raw entry count on the page.

So a page with a full 100 raw `<entry>` elements where some get filtered during normalization reads as "short" and pagination stops, silently dropping every later page for that keyword.

Repro with a synthetic 100-entry Atom page where 30 entries point at a non-jobbank.gc.ca host, so cleanUrl() returns '' and parseJobBankFeed drops them:

```
raw feed entries this page: 100
normalized rows returned  : 70
parsed.length < PAGE_SIZE fires and stops pagination: true
```

The existing pagination test ("pagination stops on a short page" in tests/providers/jobbankca.test.mjs) only covers the case where all 100 raw entries normalize cleanly, so this path isn't exercised.

Related in pattern only: #2547 covers a short-page break preempting total_pages on a16z-speedrun-talent; jobbankca has no total field, so the fix here is different.

Suggested direction: base the short-page decision on the raw entry count from the feed (count entries inside parseJobBankFeed, or return it alongside the parsed rows) rather than the post-filter parsed.length.

scttbnsn:
Fixes #3249.

Changes:
- Base the short-page stop on raw Atom entry count.
- Keep invalid-row filtering unchanged.
- Add the confirmed 100-entry regression with one titleless row and a valid second page.

Verified with the focused Job Bank suite and the full test suite: 5,386 passed, 0 failed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved job search pagination so results continue loading when a full page includes entries that cannot be displayed.
  - Prevented valid jobs from being omitted when some listings lack required details.

- **Tests**
  - Added coverage for pagination with incomplete job entries to ensure subsequent pages are requested correctly.

### DIFF
--- a/providers/jobbankca.mjs
+++ b/providers/jobbankca.mjs
@@ -281,11 +281,12 @@ export default {
           break;
         }
 
+        const rawEntryCount = (String(xml ?? '').match(/<entry\b[^>]*>[\s\S]*?<\/entry>/gi) || []).length;
         const parsed = parseJobBankFeed(xml);
         for (const job of parsed) {
           if (!byUrl.has(job.url)) byUrl.set(job.url, job);
         }
-        if (parsed.length < PAGE_SIZE) break; // short page — end of this keyword's results
+        if (rawEntryCount < PAGE_SIZE) break; // short feed page — end of this keyword's results
       }
       if (!keywordFailed) succeeded++;
     }
--- a/tests/providers/jobbankca.test.mjs
+++ b/tests/providers/jobbankca.test.mjs
@@ -413,6 +413,50 @@ try {
     else fail(`jobbankca.fetch() only slept ${slept}ms total for 2 requests`);
   }
 
+  // A page can contain the full 100 raw Atom entries even when one is
+  // filtered during parsing. Pagination is determined by the upstream page
+  // size, not by how many rows survive local validation.
+  {
+    const entry = (id, title = `role ${id}`) => `<entry>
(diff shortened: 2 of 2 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/providers/jobbankca.test.mjs`; `providers/jobbankca.mjs` is outside my area, read for context.

The new block has a precise sentinel: a page of 99 valid plus one titleless entry is 100 raw, so counting raw entries paginates to a second request, where the old `parsed.length` path filtered the titleless row to 99 and broke early at one request. That is the exact bug, and the `requested.length === 2` assertion reddens if it returns. The second assertion confirms the invalid row is still dropped while a later-page job survives. Hermetic (mocked fetchText and sleep, no network). Green on CI. Approving.

PR #3329, opened 2026-08-25
feat(providers/torre): add Torre opportunity-search provider

### LINKED ISSUE #3328: Add a Torre provider — LatAm postings never reach the Greenhouse/Lever/Ashby boards the scanner covers
Opening this ahead of a PR, per CONTRIBUTING.

## Problem

The zero-token scanner covers Greenhouse, Lever, Ashby, Workday, BambooHR, Teamtailor and Breezy. That set skews heavily US/EU. For a candidate based in LatAm and filtering for LatAm-reachable roles, a large share of the relevant market is simply not reachable by `scan.mjs` — the postings exist, but not on any board the scanner speaks to.

Concretely, on a 52-company scan from a Montevideo-based profile: 7,063 jobs found, and after title and location filters exactly **1** LatAm-reachable role survived. The scan then handed 10 companies off to the WebSearch fallback (Level 3), which is the expensive, index-lagged path the provider layer exists to avoid — and in that run it produced nothing live.

## Proposal

Add [Torre](https://torre.ai) as a provider. It is a pan-LatAm talent marketplace with strong AR/CO/MX/UY coverage and a meaningful share of remote roles that never appear on the boards already supported.

It exposes a public, zero-auth search endpoint (`POST search.torre.co/opportunities/_search`), so it fits the existing zero-token provider contract with no credentials.

Three API behaviours shape a correct implementation, and are worth recording here because each is a trap:

- **Unrecognised filter keys are silently ignored** and return the full ~300k catalogue with a `200`. A config typo would otherwise quietly walk the entire board. Only filters observed to actually move `total` should be sent.
- **`size` caps at 20.** Larger values return an empty array rather than clamping.
- **No pagination form works.** `offset`, `page`, `from`, and a body-level `offset` all return the identical first 20 rows.

So each entry issues exactly one request for at most 20 rows, with no `max_pages`. Breadth comes from multiple entries with different `search` terms; precision comes from the scanner’s own title/location filters.

Happy to send the PR — I have it implemented and tested against current `main`.

hernanflores:
Closes #3328.

Adds [Torre](https://torre.ai) as a zero-token provider. Torre is a pan-LatAm talent marketplace with strong AR/CO/MX/UY coverage and remote roles that never reach Greenhouse/Lever/Ashby — the boards the scanner already handles skew US/EU, leaving LatAm-reachable postings undiscovered. See the issue for the measured gap.

Reads the public zero-auth `POST search.torre.co/opportunities/_search`. No credentials, so it fits the existing zero-token contract.

## Three API behaviours that shape the implementation

- **Unrecognised filter keys are silently ignored**, returning the full ~300k catalogue with a `200`. Only filters observed to actually move `total` are sent, so a config typo cannot quietly walk the entire board.
- **`size` caps at 20** — larger values return an empty array instead of clamping.
- **No pagination form works.** `offset`, `page`, `from` and a body `offset` all return the identical first 20 rows.

Each entry therefore issues exactly one request for at most 20 rows, with no `max_pages`. Breadth comes from multiple entries with different `search` terms, and precision from the scanner’s own title/location filters. `experience` is a validated enum sent alongside `skill/role.text` because the API 500s when the text filter arrives without it. Posting URLs are built from the opportunity id rather than taken from the payload.

## Safety

- `redirect: "error"` passed to `fetchJson` (SSRF guard), covered by a test.
- Rows that are closed, untitled, id-less, or carry a malformed id are dropped.
- Repeated opportunities within a page are deduped.

## Tests

209 lines of new coverage in `tests/providers/torre.test.mjs`, including the one-request-per-entry invariant (holds regardless of `max_pages` / `ctx.maxPages`), the SSRF guard, row filtering, permalink construction, `created` → `postedAt` mapping, and the unexpected-response-shape throw.

`node test-all.mjs` on this branch against current `main`: **5414 passed, 0 failed**.

https://claude.ai/code/session_01XDPadS5zML44E5UKwjj232

## Summary by CodeRabbit

* **New Features**
  * Added Torre as a supported job board.
  * Supports role, experience, and remote-only search filters.
  * Displays validated, open opportunities with company details, posting dates, and direct links.
  * Removes duplicate or incomplete listings for more relevant results.

* **Documentation**
  * Documented Torre as a supported job board.


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/providers/torre.test.mjs`; `providers/torre.mjs` is outside my area, read for context.

Real sentinels throughout: the query build pins the skill/role plus required-experience pairing and throws on a bad experience, fetch pins the endpoint and POST body, drops closed, untitled, id-less and traversal-id rows to keep exactly two, builds the permalink from the id, dedups within a page, and makes exactly one request under the page cap. The SSRF guard is pinned, every request carries `redirect: 'error'`. Hermetic (mocked fetchJson, captured url and opts, no network), neighbour-style. Green on CI. Approving.

PR #3331, opened 2026-08-25
feat(providers/getonbrd): scan multiple categories per entry

### LINKED ISSUE #3330: providers/getonbrd is pinned to the programming category, so leadership and ML/data roles are never fetched
Opening this ahead of a PR, per CONTRIBUTING.

## Problem

`providers/getonbrd.mjs` is hardcoded to the `programming` category feed.

Get on Board files roles under several separate categories, and engineering-leadership titles are not among the ones `programming` covers. An Engineering Manager / Tech Lead search through this provider therefore misses most of its actual matches — `operations-management` and `machine-learning-ai` are never fetched at all.

This is easy to miss because the provider does not fail: it returns a plausible, non-empty result set drawn from a single category, so the gap looks like "Get on Board has few leadership roles" rather than "we only asked for one category".

Real example — all three of these are live Get on Board postings for the same company, and only the first is reachable today:

| Role | Category | Fetched today? |
|---|---|---|
| Tech Lead Payments (Full-Stack) | `programming` | yes |
| Engineering Manager | `operations-management` | no |
| Engineering Director | `programming` (`.com.mx` variant) | no |

## Proposal

Let a portal entry name the categories to scan, defaulting to `programming` so existing configs are unaffected.

Two things worth getting right:

- **Slug validation.** The slug lands in a feed URL, so it should be validated against an anchored pattern — otherwise a config typo can inject a path segment or a query string. The existing host assert is a good second gate but should not be the only one.
- **A cap and cross-category dedup.** Each category costs up to `max_pages` requests, so the count needs a ceiling; and a posting can be filed under more than one category, so results need deduping by URL.

Happy to send the PR — implemented and tested against current `main`.

hernanflores:
Closes #3330.

The provider was pinned to the `programming` category feed. Get on Board files engineering-leadership and ML/data roles under separate categories, so an Engineering Manager / Tech Lead search missed most of its matches: `operations-management` and `machine-learning-ai` were never fetched. See the issue for a live three-posting example where only one is reachable today.

## Change

Adds `category: <slug>` (single) and `categories: [<slug>, ...]` (several) on the portal entry.

- **Slug validation.** Slugs are checked against an anchored pattern so a config typo cannot inject a path segment or query into the feed URL. The existing host assert stays as the second gate.
- **Cap.** Categories are capped at 12, since each costs up to `max_pages` requests.
- **Dedup.** Jobs are deduped by URL across categories, because a posting can be filed under more than one.

## Backward compatibility

Defaults to `programming`, so existing entries fetch **byte-identical URLs**. There is a test asserting exactly that.

## Tests

79 lines of new coverage in `tests/providers/getonbrd.test.mjs`: single and multi-category forms, the unchanged-default-URL guarantee, slug rejection (path-segment and query injection), the 12-category cap, and cross-category URL dedup.

`node test-all.mjs` on this branch against current `main`: **5399 passed, 0 failed**.

https://claude.ai/code/session_01XDPadS5zML44E5UKwjj232

## Summary by CodeRabbit

* **New Features**
  * Added configurable category-based job searching, with support for multiple categories.
  * Automatically removes duplicate categories and job postings.
  * Supports up to 12 categories per search, defaulting to programming roles.
  * Added coverage for leadership and machine learning/data-related categories.

* **Documentation**
  * Updated provider documentation with category configuration, supported limits, and deduplication behavior.


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/providers/getonbrd.test.mjs`; `providers/getonbrd.mjs` is outside my area, read for context.

The `resolveCategories` sentinels pin the default, trimming, dedup-and-order, the `categories:` precedence, null-as-unset, path-injection rejection (`../../admin` and friends), and the empty-list and twelve-cap throws, each tracing to the slug regex and the resolver. The fetch case pins one feed per category in order, URL dedup across categories, and a pre-network throw on a bad category. Hermetic (all fetchJson mocked, no network), neighbour-style. Green on CI. Approving.

BigBrown10 (commented):
Reviewed the diff against `providers/getonbrd.mjs` on main — this is clean, well-scoped work, @hernanflores.

**What I verified:**

- `resolveCategories()` correctly handles all config shapes: missing key → default, single string → trimmed, array → deduped, `categories:` wins over `category:`, bare YAML `null` treated as unset. The slug regex (`/^[a-z0-9]+(?:-[a-z0-9]+)*$/`) rejects every injection/typo vector in the test matrix — path traversal, uppercase, double slashes, empty string, trailing hyphens, non-string types. Solid.
- The `fetch()` loop iterates categories in config order, and the `seen` Set deduplicates by normalized URL across categories. A posting listed under two categories appears exactly once — the test confirms this with a shared mock entry served by both category mocks.
- `assertGetonbrdUrl` is now called per-category (correct, since the URL varies), and the error message includes the category name for debugging.
- Backward compatibility is preserved: no config → single default category → byte-identical to the previous code path. The docs update matches the code's actual semantics (cap 12, deduped, slug format).

**One optional suggestion** (not blocking):

The `resolveCategories` tests don't cover `categories: null` (a bare `categories:` key in YAML parses as null). It currently falls through to the default, which is correct, but pinning it in a test would catch accidental regressions:

```js
if (JSON.stringify(resolveCategories({ categories: null })) === '["programming"]')
  pass('resolveCategories() treats a null categories: (bare YAML key) as unset');
```

Same for `categories: [null]` — the `typeof c !== 'string'` guard catches it, but an explicit assertion documents the behavior.

CI green across all three OS lanes. Good first provider PR — the multi-category feature addresses a real gap (EM/Tech Lead searches missing leadership roles outside `programming`). LGTM.

Issue #3306, opened 2026-08-25
test(harness): three homes for a test suite, two of them hand-registered


vliggio:
## Problem

A test suite can live in three places, and which one it lands in is not governed by a rule:

| location | suites | how it runs |
|---|---:|---|
| root, next to what it tests | 9 | explicit `run(NODE, ['--test', ...])` in `test-all.mjs` |
| `test/` | 5 | explicit `run(NODE, ['--test', ...])` in `test-all.mjs` |
| `tests/` | 216 | auto-discovered (#1440) |

`ARCHITECTURE.md` documents one convention — *"`*.test.mjs` sits next to what it tests"* — and that convention is the minority even within its own scope. Counting root-level scripts specifically:

- **23** root scripts have their tests in `tests/` (`intake.mjs` → `tests/intake.test.mjs`, `merge-tracker.mjs` → `tests/merge-tracker.test.mjs`, `linkedin-join.mjs` → `tests/linkedin-join.test.mjs`, and twenty more)
- **9** root scripts have their tests adjacent (`contacts.mjs` → `contacts.test.mjs`)

Both conventions took a new file on 2026-08-25. This is not a migration that stalled — it is two live conventions covering the same case, where the outcome depends on who wrote the test.

The cost is not tidiness. Manual registration is the shared root cause of two bugs found in the last week: #3247 (`test/profile-photo.test.mjs` and `test/zh-minimal-template.test.mjs`, added in July, never ran) and #3303 (`jd-similarity.test.mjs`, 26 checks, in no runner). Fourteen suites are registered by hand and can be forgotten; the 216 in `tests/` cannot, because nobody has to remember them.

`jd-similarity.test.mjs` is the sharpest version: it is listed in `update-system.mjs`'s `SYSTEM_PATHS`, so it ships to every user on update, and it is named nowhere in `test-all.mjs`, so it has never run.

## Not proposing a root reorganization

#1386 settled that and the reasoning holds. `ARCHITECTURE.md` § "Why the flat root" records it: path stability is a feature, `SYSTEM_PATHS` is the most bug-prone surface in the repo, and thousands of users have `node scan.mjs` in muscle memory. **No `.mjs` script moves. No entry point moves.**

What follows is only about where *test files* live, and the distinction matters: a test path is not an entry point. Checking every root-level `*.test.mjs` for external references turns up `test-all.mjs` and `update-system.mjs` and nothing else — no doc, no plugin, no workflow, and nothing a user types.

[...]

PR #3335, opened 2026-08-25
test(harness): retire test/ into tests/, one discovery mechanism (#3247)

### LINKED ISSUE #3247: test: two test/ suites are never run — test-all.mjs registration is manual and was forgotten twice

vliggio:
## What does this PR do?

Retires `test/` into `tests/`. Five node:test suites move into auto-discovery, `test/cv-visual/` moves alongside them, five registration blocks come out of `test-all.mjs`, and the directory is gone.

## Related issue

Closes #3247. This is the option @santifer picked in the thread — *"move the five suites into `tests/` and delete the manual registrations — one discovery mechanism, zero registration to forget, and the guard is structural instead of procedural"* — rather than the (a) I originally proposed and had opened #3305 for. **#3305 is superseded and I'll close it.**

## Why the move is cheap

`runDiscovered` already handles node:test suites: it branches on the `node:test` import and child-processes them, folding the real exit code into the counters (#2828, covering 16 such suites in `tests/` today). All five `test/` files are node:test style, so they run under discovery with no conversion. That measurement is @artemtrofymenko's, on #3247.

## `test/cv-visual/` moves rather than converts

It is a Playwright spec. Most of it would port — `page.goto`, `emulateMedia`, `evaluate`, `page.pdf` are plain `playwright` library calls and the package is already a direct dependency; the matchers map onto `node:assert`. One line does not, and it is the line the suite exists for:

```js
await expect(page).toHaveScreenshot(`${artifactBase}.png`, { maxDiffPixelRatio: 0.04 });
```

That brings four committed baselines, `snapshotPathTemplate`, per-platform naming, pixel diffing and the `--update-snapshots` path wired to `npm run test:cv-visual:update`. Porting it means reimplementing image diffing to win a directory move.

It does not need a directory of its own either. `discoverTests` filters on `.test.mjs`, so a `.spec.mjs` under `tests/` is invisible to `test-all.mjs`, and the naming convention that already separates the two runners keeps separating them. There are no other `.spec.mjs` files under `tests/`. `playwright.cv.config.mjs`'s `testDir` and the spec's hardcoded `baselines.json` path follow it across.

## Registrations removed whole

Headers and `fail()` message strings, not just the `run()` calls. @artemtrofymenko flagged this on #3305's branch: leaving a filename behind in a section header or an error string keeps any name-based check green while nothing invokes the suite. That halfway state is reachable by splitting the work across two commits, so this is one commit, and `grep` finds no `test/` filename left anywhere.

§59 keeps its CLI checks — those exercise `cv-templates.mjs`'s command surface, which discovery does not cover. §59b and §60 were nothing but the registration and are gone entirely.

## Verified by what executes, not by what is named

```
✅ tests/cv-templates.test.mjs — node:test suite passed (25 tests)
✅ tests/pipeline-lock.test.mjs — node:test suite passed (21 tests)
✅ tests/cover-resolver.test.mjs — node:test suite passed (4 tests)
✅ tests/profile-photo.test.mjs — node:test suite passed (8 tests)
✅ tests/zh-minimal-template.test.mjs — node:test suite passed (4 tests)
```

`node test-all.mjs` → **5390 passed, 0 failed**, 2 warnings, both pre-existing (baseline on the same head: 5388, same 2 warnings). The +2 is exactly the two suites that had never run.

`npx playwright test --config=playwright.cv.config.mjs --list` → 4 tests at the new path. Running it locally fails only on `pdftotext ENOENT`, a binary this machine lacks; `toHaveScreenshot` executes *before* that call in the spec, so the moved baselines resolving is what lets it get that far. CI installs `poppler-utils` in the `cv-visual` job, which needs no change since it invokes `npm run test:cv-visual`.

Six `SYSTEM_PATHS` entries repointed; the coverage guard is green. Worth noting it caught a half-staged move during development — with the deletions unstaged, `git ls-files` still listed `test/` and both that guard and the js-yaml sweep failed loudly. Exactly the failure mode the single-commit discipline above is for.

## Not in scope

The nine root-level `*.test.mjs` are hand-registered too, and `tests/` already holds the tests for 23 root scripts — so root-adjacent is the minority convention even within its own scope. Whether those follow is a separate call, raised as Part 2 of #3306 and left for @santifer. #3303 is the live instance of that class.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] Issue opened first (#3247)

## Summary by CodeRabbit

- **Tests**
  - Expanded coverage for CV and cover-letter template selection, validation, formatting, and fallback behavior.
  - Added regression checks for profile photos, pipeline locking, Chinese Minimal templates, path safety, and error handling.
  - Added visual regression fixtures and page-count baselines for English and Chinese CV variations, including short and detailed profiles with or without photos.

- **Chores**
  - Consolidated automated tests under a unified test directory and updated test discovery and visual-test configuration accordingly.


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is the twelve files moved under `tests/`; `test-all.mjs` is the shared harness and `update-system.mjs` is the updater, both santifer's, read for context.

This is a clean retire of the singular `test/` directory into `tests/`, and the thing I checked hardest is that no coverage is lost. It is: the five unit suites move at 100% similarity, cv-visual at 98% (one path line), and because discovery globs `tests/**/*.test.mjs`, each moved suite now runs via the harness. The three inline blocks removed from `test-all.mjs` are replaced by that discovery, net neutral, and two suites that previously sat undiscovered in `test/` (profile-photo, zh-minimal) now actually start running, which is a gain. No `test/` singular `--test` reference remains, and cv-visual stays wired through the updated Playwright config rather than orphaned. Green on CI. Approving.

PR #3337, opened 2026-08-25
refactor(reporters): adopt isMainModule for the read-only CLIs (#3170)

vliggio:
## What does this PR do?

Converts the 18 read-only reporters — the JSON / `--summary` read-models over
the tracker, the reports and the story bank — from hand-rolled main-guards to
`isMainModule()` from #3278.

Through a symlinked checkout each of these printed nothing and exited 0. For a
reporter that's a particularly quiet failure: an empty answer reads as "no
findings" rather than "never ran", so a silenced `rejection-latency` or
`detect-reposts` looks like good news.

Three carry more than the one-line swap:

- **`classify-tier.mjs`** — its guard was the last user of `path` and
  `fileURLToPath`; both imports go.
- **`upskill.mjs`** — the comment above its guard consciously accepted this
  defect ("matching the repo convention is worth more here than covering a path
  nothing takes"). Rewritten to record why that premise didn't hold.
- **`invite-match.mjs`** — two guards, both converted.

`PENDING` drops 58 → 40.

## Note on the conflict resolution

Six of these files changed on `main` while the series was in review. Rather than
replay a stale patch, the conflicted files were re-derived: take `main`'s
current version, re-apply the conversion. So the diff here is against today's
`main`, not a three-week-old snapshot.

## Part of the #3170 series

Batch 2 of 6. #3278 landed the shared helper, the convention test and its

ratchet.

## Checklist

## Type of change

- [ ] Bug fix
- [ ] New feature

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved command-line execution detection across utility scripts, including support for symlinked or path-shimmed invocations.
  * Preserved existing argument validation, processing, self-tests, and output behavior.
  * Standardized command-line startup behavior for more reliable script execution.

* **Tests**
  * Updated entry-point convention checks to reflect newly standardized scripts and continue tracking remaining conversions.


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/main-guard-convention.test.mjs`; the eighteen converted reporter scripts are outside my area, read for context.

The invariant is enforced on both sides: the offenders test requires the just-converted files to be clean of a raw `process.argv[1]`, and the dead-entry test requires every file remaining in the PENDING ratchet to still reference the entry path, so the ratchet must shrink in lockstep with each conversion. The sentinel is real behaviour, not a spelling pin: it spawns `generate-pdf.mjs --help` through a real junction symlink and asserts non-empty output with status parity, which is exactly the silent no-op #3170 was, and a self-test confirms the detector matches the defect. Hermetic (tmpdir plus junction, skips where symlinks are unavailable).

One merge note: #3338, #3339 and #3340 all ratchet this same PENDING list for their own scripts, so expect a keep-both conflict at the tail whenever one lands after another. The content is disjoint. Green on CI. Approving.

PR #3338, opened 2026-08-25
refactor(cv): adopt isMainModule for the document generators (#3170)

vliggio:
## What does this PR do?

Converts the 10 CV, cover-letter and artifact generators to `isMainModule()`
from #3278.

This is where the silent-success failure mode was most expensive. Through a
symlinked checkout `generate-latex`, `img-to-pdf` and `generate-cover-letter`
produced no file and exited 0 — so `apply` and `pdf` recorded an artifact that
was never written, and the tracker's PDF column said ✅ for a file that does not
exist.

`PENDING` drops by 10.

## One ordering note

`cv-templates.mjs` is also rewritten by `feature/template-packs` (parseMeta /
resolveTemplate). Whichever lands second needs a trivial rebase — the change
here is an import plus one line.

## Part of the #3170 series

Batch 3 of 6. #3278 landed the shared helper, the convention test and its

ratchet.

## Checklist

## Type of change

- [ ] Bug fix
- [ ] New feature

## Summary by CodeRabbit

- **Refactor**
  - Standardized command-line entry-point detection across multiple tools.
  - Improved consistency when scripts are run directly or imported as modules.
  - Preserved existing command-line behavior and exit-code handling.

- **Tests**
  - Updated convention checks to reflect the reduced number of remaining scripts requiring conversion.

### DIFF
--- a/application-answers.mjs
+++ b/application-answers.mjs
@@ -2,7 +2,7 @@
 
 import { readFileSync, writeFileSync } from 'fs';
 import { resolve } from 'path';
-import { fileURLToPath } from 'url';
+import { isMainModule } from './lib/is-main-module.mjs';
 
 export const APPLICATION_ANSWERS_HEADING = '## Application Answers';
 
@@ -424,7 +424,7 @@ async function main() {
   console.log(JSON.stringify({ report: reportPath, date: normalized.date, state: normalized.state }, null, 2));
 }
 
-if (process.argv[1] && resolve(process.argv[1]) === fileURLToPath(import.meta.url)) {
+if (isMainModule(import.meta.url)) {
   main().catch((err) => {
     console.error(err.message);
     process.exitCode = 1;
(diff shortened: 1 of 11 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/main-guard-convention.test.mjs`; the nine converted scripts (including the cv scripts) are outside my area, read for context.

Same invariant, held on both sides: the converted files are removed from the PENDING skip set and the offenders test then requires them clean of `process.argv[1]`, while the dead-entry test keeps every remaining PENDING file honest. The behaviour sentinel spawns a CLI through a real junction symlink and asserts it actually ran with status parity, catching the #3170 silent no-op, and a detector self-test proves it matches the defect. Hermetic (mkdtemp, skips when symlinks are unavailable).

Merge note: this shares the PENDING ratchet with #3337, #3339 and #3340, so whichever lands after the others rebases a keep-both tail conflict; the content is disjoint. Green on CI. Approving.

PR #3339, opened 2026-08-25
refactor(scan): adopt isMainModule for the scanners and discovery CLIs (#3170)

vliggio:
## What does this PR do?

Converts the 10 portal scanners, ATS discovery tools and batch runners to
`isMainModule()` from #3278.

A scanner that exits 0 having scanned nothing is indistinguishable from one that
found no new postings, so this batch's failure mode was a search that quietly
stopped surfacing roles — on a symlinked checkout, indefinitely.

`batch/aggregate-tokens.mjs` is included deliberately: `batch/` is gitignored
except its scripts, and the convention test's skip-list covers user-layer trees
but **not** `batch/` — its tracked scripts are entrypoints like any other and
stay under enforcement.

`PENDING` drops by 10.

## Part of the #3170 series

Batch 4 of 6. #3278 landed the shared helper, the convention test and its

ratchet.

## Checklist

## Type of change

- [ ] Bug fix
- [ ] New feature

## Summary by CodeRabbit

- **Refactor**
  - Standardized command-line entry-point detection across evaluation, scanning, discovery, and verification tools.
  - Improved consistency when running scripts directly or importing them as modules.
- **Tests**
  - Updated entry-point convention checks to reflect the scripts now using the shared execution behavior.


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/main-guard-convention.test.mjs`; the converted scan scripts are outside my area, read for context.

The scan scripts adopt `isMainModule(import.meta.url)` with no `process.argv[1]` reference, so the convention test's no-entry-path assertion passes and the dead-entry check holds for each removed slug. The behaviour sentinel is the real one: it spawns `generate-pdf.mjs` through a real junction symlink and requires output equal to the direct invocation, exactly the #3170 silent no-op. Hermetic (tmpdir plus cleanup, skips when symlinks are unavailable).

Merge note: shares the PENDING ratchet with #3337, #3338 and #3340, so expect a keep-both tail conflict for whichever lands later; disjoint content. Green on CI. Approving.

Issue #1428, opened 2026-07-02
set-status.mjs: canonical CLI to update a tracker row's status/note (no more hand-editing the table)


santifer:
Surfaced by a real apply-run field test (~12 applications → ~12 fragile hand-edits).

## Problem

`modes/apply.md` Step 8 instructs *"update status to Applied"*, but the only mechanism is hand-editing the `data/applications.md` table (or ad-hoc node one-liners with hardcoded column indexes — which is what the field run ended up doing, 12 times).

## Proposal

`node set-status.mjs <report#|company> <CanonicalState> [--note "..."]`
- Resolves the row by report number or company name (fuzzy via the existing role-matcher patterns)
- Validates the state against `templates/states.yml` (reject non-canonical)
- Column-mapped by header name (the #1293 tolerance pattern), atomic write

## Why it matters beyond ergonomics

`data/applications.md` is a shared surface with multiple readers — ONE canonical write-path CLI is safer than N agents hand-editing markdown. Composes with #1430-style follow-up seeding as a natural hook point.

PR #3344, opened 2026-08-25
fix(stats,calibrate): ledger-aware funnel + calibration (#1428, #1724)

alechemenway:
## What

`stats.mjs` (the ever-stage funnel) and `calibrate.mjs` (score-vs-outcome calibration) both read only the **current** tracker status. A row that reached Offer or Interview and then landed on a terminal status — `Discarded` after a declined offer, `Rejected` after interviews — was undercounted or dropped from the population entirely. Now that `data/status-log.tsv` records transitions, both can fold that history and report what actually happened.

Completes two documented follow-ups:
- **#1428** — `computeFunnel`'s own comment: *"middle stages are lower bounds ... until status-transition logging exists."* The log now exists.
- **#1724** — the learning loop (calibrate reading real outcomes).

## Changes

**`stats.mjs`**
- Add `parseStatusLogStages` + `computeFunnelWithHistory`: `everX` counts distinct rows that reached stage X per the ledger's `from`/`to` history (a declined offer now in `Discarded` still counts into `everOffer`; a `Rejected`-after-interview counts into `everInterview`).
- `computeFunnel(byStatus)` is **unchanged** — `funnel-velocity.mjs` and the existing tests depend on it. The ledger path is used only when `status-log.tsv` is present; otherwise the summary is byte-for-byte the previous status-only funnel.
- Summary line gains a `· ledger-adjusted` marker and the JSON gains `basis: 'ledger'`, so the source is explicit.

**`calibrate.mjs`**
- Add `loadLedgerReached` + a `reached` option on `computeCalibration`.
- Evidence precedence is now **journal > ledger > status**: a declined offer (now `Discarded`) counts as `reachedOffer`; an interview that ended in `Rejected` counts as `reachedInterview`. A `Discarded` row with no interview/offer history stays outside the population, as before.

## Backward compatibility

- `computeFunnel` signature and output are unchanged.
- Both scripts behave exactly as before when `data/status-log.tsv` is absent (the new arg defaults to an empty map / the file read returns an empty map).
- No scoring rules, thresholds, or user files are touched — `calibrate` stays strictly advisory.

## Tests

- `tests/stats.test.mjs`: new `computeFunnelWithHistory` case (declined offer counts into `everOffer`; distinct-row counting; torn ledger line skipped). Existing `computeFunnel` cases unchanged and green.
- `calibrate.mjs --self-test`: new cases — ledger resolves a declined offer as `reachedOffer`; a history-less `Discarded` stays excluded; a journal outranks the ledger. All pass.

## Summary by CodeRabbit

- **New Features**
  - Funnel reporting now reflects historical stage progression, including applications that later reached interviews or offers.
  - Calibration prioritizes outcome journals and stage history for more accurate results.
  - Reports indicate when historical status data contributes to funnel calculations.

- **Bug Fixes**
  - Terminal applications without supporting stage history are excluded from historical-stage counts.
  - Declined offers and rejected interviews retain their previously reached stages.
  - Missing or malformed history data safely falls back to current-status reporting.

### DIFF
--- a/calibrate.mjs
+++ b/calibrate.mjs
@@ -24,9 +24,12 @@
  *     It is excluded from every rate and reported separately as in-flight.
  *   - No percentage is printed on a band below the sample floor. "2 of 3
  *     got interviews" is an anecdote wearing a rate's clothes.
- *   - The outcome journal (data/outcomes/) outranks the tracker status: the
- *     journal records what actually happened (outcome.mjs), the status is a
- *     workflow position. Where both exist, the journal wins.
+ *   - Evidence precedence: the outcome journal (data/outcomes/) wins, then the
+ *     transition ledger (data/status-log.tsv), then the tracker status. The
+ *     journal records what actually happened (outcome.mjs); the ledger records
+ *     the stages a row passed through (set-status.mjs), so a declined offer now
+ *     sitting in Discarded still counts as an offer; the current status is only
+ *     a workflow snapshot and loses that history.
  */
 
 import { readFileSync, readdirSync, existsSync } from 'fs';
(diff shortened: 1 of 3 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/stats.test.mjs`; `stats.mjs` and `calibrate.mjs` are outside my area, read for context.

The added block pins the ledger-aware funnel with a real sentinel: a row that now reads Discarded but whose status-log ledger records Interview then Offer then Discarded folds back to the offer rank, so `everOffer` is 1, and without the ledger fold that declined offer would read rank 0 and `everOffer` would be 0. The test reddens on exactly that regression, and `everApplied=4`, `everInterview=3`, `everOffer=1`, `parsed=3`, `basis='ledger'` all recompute against the fixture, with the torn ledger line correctly skipped. Hermetic (in-memory maps and strings, no filesystem), neighbour-style. Green on CI. Approving.

PR #3345, opened 2026-08-25
fix: resolve local-today gate conflict for PR 3301

chipoto69:
## Summary
- Resolves the PR #3301 merge conflict against current `origin/main`.
- Keeps both local-today scanner writer guards from #3301 and the newer reporting-window/rejection-latency local-day tests from main.
- No primary `/Users/rudlord/ORGANIZED/JOBSEARCH/career-ops` worktree touched.

## Test plan
- `node --test tests/local-today-gates.test.mjs` (15 passed)

## Steward context
- Bridge branch for santifer/career-ops#3301 because the original PR is `mergeStateStatus=DIRTY` and `reviewDecision=REVIEW_REQUIRED`.
- Normal push to `chipoto69:steward/pr3301-conflict-probe`; no force-push.

## Summary by CodeRabbit

* **Bug Fixes**
  * Scan dates now use the local calendar day instead of UTC, ensuring scan history, summaries, and digest filenames show the correct date.
  * Applied consistent local-date handling across all supported scanners.

* **Tests**
  * Added coverage to verify scanners use local dates and avoid UTC-derived date values.


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/local-today-gates.test.mjs`; the scanner and reporting sources are outside my area, read for context.

This resolves the merge conflict two sibling PRs left at the tail of this file, and the resolution is clean. Both appended sets survive once each, with no duplication and no conflict markers: the writer-census set (a balanced-paren `toISOString` scan of the four scan-history call sites plus a per-writer `localToday` import check) and the reporting-windows set (frozen-clock `WEEK_INSTANT`/`INSTANT` under `TZ=America/New_York`, asserting the local day over the UTC day for weekly-digest and rejection-latency). Each keeps its real sentinel that reddens under the UTC-day bug, every dynamic-import target exists at head, and the `scan.mjs` UTC-midnight anchor at line 776 is correctly not flagged as an offender. No test case was lost, dropped or duplicated. Green on CI. Approving.

PR #3350, opened 2026-08-26
fix: reject malformed status ledger rows

chipoto69:
## Summary
- Addresses CodeRabbit review comment on PR #3344.
- Rejects malformed `data/status-log.tsv` rows unless `num` is a complete numeric token and `num`, `date`, `from`, and `to` are all non-empty.
- Adds regression coverage for partial numeric identifiers and missing ledger fields.

## Test plan
- `node tests/stats.test.mjs`
- `npm run lint`
- `node calibrate.mjs --self-test`

Bridge for #3344; original contributor branch push was denied to chipoto69, so this is a non-force handoff PR.

## Summary by CodeRabbit

- **New Features**
  - Funnel metrics now incorporate application history, preserving interview and offer progress after applications become terminal.
  - Calibration results combine historical stage progression, current statuses, and recorded outcomes.
  - Reports indicate when funnel results use historical activity.

- **Bug Fixes**
  - Improved handling of declined offers, rejected or discarded applications, incomplete records, malformed history, and partial identifiers.
  - Invalid or incomplete status transitions are now excluded from metrics.
  - Explicit recorded outcomes take precedence over inferred historical stages.
  - Metrics fall back to current-status calculations when historical data is unavailable.


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/stats.test.mjs`; `stats.mjs` is outside my area, read for context.

Two-sided sentinel: four malformed ledger rows are rejected (a missing destination, a missing date, a non-numeric num, a torn line with no num) via the `^\d+$` plus date/from/to guard, while three valid rows still fold into the funnel, and the non-numeric-id case specifically proves a torn id does not inflate its row. `everApplied=4`, `everInterview=3`, `everOffer=1` and `ledgerParsed.length===3` all recompute against the fixture. Hermetic (inline TSV, no filesystem), neighbour-style.

One rebase note: this adds its block to the same region of the file as #3344's ledger-funnel block, so the two will conflict textually whichever lands second; the content is additive, not contradictory. Green on CI. Approving.

PR #1728, opened 2026-07-09
fix(pdf): render markdown bold in CV summary as <strong>

gidwyz:
## Summary
- Convert markdown bold (`**text**`) in CV HTML to `<strong>` tags during ATS normalization in `generate-pdf.mjs`
- Add `.summary-text strong` CSS so tailored summary emphasis renders correctly in PDF output instead of literal asterisks

## Test plan
- [ ] Run `node generate-pdf.mjs` on HTML with `**bold**` in the summary section
- [ ] Confirm the generated PDF shows bold text, not literal `**` asterisks

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * Bold text in CV content is now preserved and displayed more consistently in generated PDFs.
  * Markup-style bold formatting is converted into proper bold styling, improving readability in the final document.
  * Text normalization continues to work as before, with an added count for this formatting conversion.

PR #3352, opened 2026-08-26
fix(build-cv-latex): render markdown bold in bullets as \textbf

### LINKED ISSUE #3351: build-cv-latex: markdown bold in bullets renders literally — #1728 fixed only the HTML path
### Code of Conduct

### Existing issues

### What happened?

Markdown bold (**text**) in payload bullets renders as bold through the HTML → PDF path, but reaches the LaTeX output as literal asterisks. Same payload, two different results.

The HTML path already works — including bullets. [#1728](https://github.com/santifer/career-ops/issues/1728) added a **…** → <strong> conversion to normalizeTextForATS() (generate-pdf.mjs:163). The PR title says "CV summary", but the conversion runs over every text node in the document, so it already covers experience[].bullets and projects[].bullets. templates/cv-template.html:235 already ships .job li strong { font-weight: 600 } to style exactly that case.

The LaTeX twin has no equivalent step. generate-latex.mjs does no markdown normalization at all, and build-cv-latex.mjs:36 / :48 pass bullets through escapeLatex(), which leaves * untouched (it is not a LaTeX special character). The asterisks survive verbatim into the .tex.

This is the same twin-divergence shape as #2915 (#2641 fixed only the HTML mirror) and #2170 (HTML template missing a field the LaTeX one had).

Secondary: the payload schema tables in modes/pdf.md and modes/latex.md never document that **bold** is supported, so an agent building the payload has no reason to emit it. The HTML-side capability is effectively undiscoverable.

### Steps to reproduce

Synthetic payload, no personal data:
```
cat > /tmp/co-payload.json <<'JSON'
{
  "lang": "en",
  "page_format": "letter",
  "candidate": { "name": "Test Candidate", "email": "t@example.com", "location": "City" },
  "summary": "Backend engineer who cut cold start to **1018 KB**.",
  "competencies": ["Cloud Architecture"],
  "experience": [{
    "company": "Test Corp", "role": "Test Engineer", "location": "Remote", "dates": "2024 - Present",
    "bullets": ["Reduced bundle from 3.2 MB to **1018 KB** across 4 & 5 services"]
  }]
}
JSON

node build-cv-html.mjs  /tmp/co-payload.json /tmp/co-out.html
node build-cv-latex.mjs /tmp/co-payload.json /tmp/co-out.tex
```

HTML path — apply the normalization generate-pdf.mjs runs before rendering:

```
node --input-type=module -e "
import { readFile } from 'fs/promises';
import { normalizeTextForATS } from './generate-pdf.mjs';
const r = normalizeTextForATS(await readFile('/tmp/co-out.html', 'utf-8'));
console.log(JSON.stringify(r.replacements));
for (const l of r.html.split('\n')) if (l.includes('1018')) console.log(l.trim());
"
```

Output — bold applied in both the summary and the bullet:

```
{"em-dash":1,"markdown-bold":2}
<div class="summary-text">Backend engineer who cut cold start to <strong>1018 KB</strong>.</div>
<li>Reduced bundle from 3.2 MB to <strong>1018 KB</strong> across 4 & 5 services</li>
```

LaTeX path:
`grep 1018 /tmp/co-out.tex`

Output — literal asterisks in the .tex:
```
            \resumeItem{Reduced bundle from 3.2 MB to **1018 KB** across 4 \& 5 services}
```

### Expected behavior

One payload should render the same in both twins: build-cv-latex.mjs should restore **text** to \textbf{text} after escapeLatex() runs, mirroring what generate-pdf.mjs does on the HTML side.

Escaping order is what keeps the HTML path safe today, and the fix should preserve it. escapeHtml() runs first in the builder; the ** conversion then wraps already-escaped text. A probe bullet of **<script>alert(1)</script>** comes out as <strong><script>alert(1)</script></strong> — inert. The LaTeX fix should keep the same order (escape first, restore second), so escapeLatex() still owns every special character and only the ** markers are reinterpreted.

Worth documenting **bold** in the payload schema tables in modes/pdf.md and modes/latex.md at the same time, since the existing HTML support is currently invisible to the agent writing the payload.

Happy to send a PR for both if this framing looks right.

### CLI tool

Claude Code

### OS and Node.js version

macOS 26.5.2 / Node v23.7.0 / career-ops v1.28.0

### Output of npm run doctor

```shell
N/A — reproducible from the synthetic payload above; not a setup issue.
```

yuki-uix:
## What does this PR do?

Teaches `build-cv-latex.mjs` to render `**text**` as `\textbf{text}`, the LaTeX half of the support #1728 added to the HTML path. Before this, the same payload emphasised correctly in the HTML PDF and printed literal `**1018 KB**` in the LaTeX one, with the builder reporting `"valid": true` and exiting 0.

## Related issue

Closes #3351

## Type of change

- [x] Bug fix
- [ ] New feature

## The fix

`escapeLatex()` leaves `*` alone because it is not a LaTeX special character, so the markers survived into the `.tex`. The restore runs **after** escaping:

```js
const MARKDOWN_BOLD_RE = /\*\*([^*]+?)\*\*/g;   // same regex as the HTML path

function escapeLatexBullet(text) {
  return escapeLatex(text).replace(MARKDOWN_BOLD_RE, (_, inner) => `\\textbf{${inner}}`);
}
```

**That order is the safety property.** `escapeLatex()` has already turned every backslash into `\textbackslash{}` and every brace into `\{`/`\}`, so nothing the candidate wrote can come back as a control sequence — this pass only reinterprets the `**` markers, which escaping deliberately left untouched. A bullet containing a hand-typed `\textbf{this}` comes out as `\textbackslash{}textbf\{this\}`, inert.

The replacer is a **function, not a string**, for the same reason the render path below it documents: escaped text is full of `\$` and `\&`, and a string replacement would reinterpret `$&` as a match reference.

### Scope: every `\resumeItem` exit, not every field named `bullets`

The gate covers all three places this builder emits a `\resumeItem` — `experience[].bullets`, `projects[].bullets`, and the `education[].coursework` line. Coursework carries no `bullets` key but renders as a bullet, and a bullet whose emphasis prints as `**` is the bug being fixed. The shape of the output decides membership, not the payload field name.

Bold stays bullet-scoped on this side. `projects[].name`, `projects[].context`, `awards[].title` and the `skills[]` fields still print `**` literally while the HTML path bolds them, and this builder has no `summary` field at all. Both mode docs now say exactly that, field by field, instead of claiming the two formats emphasise identically — happy to widen the scope if you'd rather close that divergence in one go.

## Verification

**Payloads that contain no `**` produce a byte-identical `.tex`** — `cmp` clean against the pre-change builder. Existing users see no change at all.

New `tests/cv-latex-bullet-bold.test.mjs` (modeled on `tests/cv-latex-numeric-scalars.test.mjs` from #2916, the same twin-divergence shape), 9 assertions end-to-end through the real builder:

- experience / project / coursework bullets render `\textbf{…}`
- LaTeX specials inside a bold span keep their escaping: `\textbf{\$2M budget \& 99.9\% uptime}`
- a hand-typed `\textbf` stays inert
- single asterisks and unmatched `**` stay literal
- a bold span containing a `*` (`**a *b* c**`) matches nothing — pinned as a known limit rather than fixed, since it is the HTML path's behaviour too and changing it here alone would re-open the divergence. Documented in both mode files.

Mutation-checked: reverting any one call site to `escapeLatex`, or inverting the escape order, makes assertions fail. The tests exercise the real path rather than re-deriving the transform.

`node test-all.mjs`: **4910 passed, 1 failed**. The failure is `analyze() appDateSource end-to-end check` — `followup-cadence.mjs` copied to a temp dir cannot resolve `js-yaml`. It reproduces identically on a clean tree with this branch stashed (baseline 4901 passed, 1 failed), so it is a local environment artifact, unrelated to this change.

### Two things I could not fully verify

**`pdflatex` is not installed on my machine**, so the `.tex` was never compiled to a PDF. Verification stops at the `.tex` content.

**The issue body says `stripMarkup` in `verify-cv-facts.mjs` already handles `\textbf` — that is only half right, and I want to correct it here rather than leave it standing.** It holds for a standalone `\textbf{X}`, but its brace matching is `[^}]*`, not balanced, so a `\textbf{}` *nested inside* `\resumeItem{}` leaks a bare `textbf` token and misplaces the brace boundary. This is **pre-existing** — the shipped template's `\resumeItem{\textbf{Coursework:} …}` leaks it with an unmodified builder and no `**` in the payload at all. No verdict flip showed up in realistic before/after runs (all words survive; only the braces move), but this PR does raise how often it is hit, from once per CV to once per bolded span. Happy to open a separate issue for it; `lib/latex-content.mjs` already has a balanced `findMatchingBrace` that could back a fix.

## Checklist

- [x] Bug fix — issue #3351 opened first anyway
- [x] My PR does not include personal data (all fixtures are synthetic)
- [x] I ran `node test-all.mjs` — 4910 passed, 1 pre-existing env failure documented above
- [x] My changes respect the [Data Contract](https://github.com/santifer/career-ops/blob/main/DATA_CONTRACT.md) — system layer only

## Summary by CodeRabbit

- **New Features**
  - Markdown bold syntax (`**text**`) now renders as bold text in generated LaTeX CVs.
  - Supported in coursework, experience bullets, and project bullets.
  - Text remains safely escaped while preserving literal content and handling invalid markers predictably.

- **Documentation**
  - Updated LaTeX and PDF mode documentation with supported syntax, field coverage, escaping behavior, and usage restrictions.

- **Tests**
  - Added coverage for bold rendering, escaping, unmatched markers, and nested asterisk patterns.

### DIFF
--- a/build-cv-latex.mjs
+++ b/build-cv-latex.mjs
@@ -13,6 +13,50 @@ const __dirname = dirname(fileURLToPath(import.meta.url));
 const TEMPLATE_PATH = resolve(__dirname, 'templates', 'cv-template.tex');
 const PLACEHOLDER_RE = /\{\{[A-Z_]+\}\}/g;
 
+// Markdown bold inside bullets — the LaTeX half of #1728, which taught the HTML
+// path to render `**text**` as <strong> (normalizeTextForATS in generate-pdf.mjs).
+// escapeLatex() leaves `*` alone because it is not a LaTeX special character, so
+// the markers reached the .tex verbatim and printed as literal asterisks (#3351).
+//
+// Order is the safety property, and it mirrors the HTML twin: escapeLatex() runs
+// FIRST, so every backslash and brace in the payload is already neutralized
+// (`\` becomes \textbackslash{}, braces become \{ \}). Nothing the candidate wrote
+// can survive as a real control sequence — this pass only reinterprets the `**`
+// markers, which escaping deliberately left untouched. Same regex as the HTML
+// path so the two twins agree on what counts as bold.
+//
+// The gate covers every field this builder emits inside a \resumeItem: experience
+// bullets, project bullets, and the education coursework line. Coursework does not
+// carry the payload key `bullets`, but it renders as a bullet, and a bullet whose
+// emphasis silently prints as `**` is the bug being fixed — the shape of the
+// output decides what goes through the gate, not the name of the payload field.
+const MARKDOWN_BOLD_RE = /\*\*([^*]+?)\*\*/g;
+
+/**
+ * Escape bullet text, then restore markdown bold as \textbf.
+ *
+ * Use this for every value that ends up inside a \resumeItem; use escapeLatex
+ * directly everywhere else, where `**` is meant to stay literal.
+ *
+ * @param {string} text raw payload text, not yet escaped
+ * @returns {string} LaTeX-safe text with `**…**` spans rendered as \textbf{…}
+ */
+function escapeLatexBullet(text) {
+  // Replacer FUNCTION, not a string: escaped text is full of `\$` and `\&`, and a
+  // string replacement would reinterpret `$&` and friends as match references
+  // (same trap the render path documents below).
+  return escapeLatex(text).replace(MARKDOWN_BOLD_RE, (_, inner) => `\\textbf{${inner}}`);
+}
+
+/**
+ * Render the Education section as \resumeSubheading blocks.
+ *
+ * An entry's optional `coursework` becomes a single \resumeItem line, which is
+ * why it goes through escapeLatexBullet rather than escapeLatex.
+ *
+ * @param {Array<object>} entries `education[]` from the payload
+ * @returns {string} LaTeX for the section body, or '' when there is nothing to render
+ */
 function buildEducation(entries) {
   if (!Array.isArray(entries) || entries.length === 0) return '';
   const blocks = [];
@@ -20,7 +64,7 @@ function buildEducation(entries) {
     if (!e) continue;
     let block = `    \\resumeSubheading\n      {${escapeLatex(e.institution)}}{${escapeLatex(e.location)}}\n      {${escapeLatex(e.degree)}}{${escapeLatex(e.dates)}}`;
     if (Array.isArray(e.coursework) && e.coursework.length > 0) {
-      const courses = e.coursework.map(c => escapeLatex(c)).join(', ');
+      const courses = e.coursework.map(c => escapeLatexBullet(c)).join(', ');
       block += `\n        \\resumeItemListStart\n            \\resumeItem{\\textbf{Coursework:} ${courses}}\n        \\resumeItemListEnd`;
     }
     blocks.push(block);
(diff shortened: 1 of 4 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/cv-latex-bullet-bold.test.mjs`; `build-cv-latex.mjs` is outside my area, read for context.

The test holds up with a real two-sided sentinel. Bold markdown becomes `\textbf` in all three bullet emitters (experience, projects, coursework), and the escape-first ordering is preserved and pinned: `\textbf{\$2M budget \& 99.9\% uptime}` keeps its LaTeX escaping, a literal `\textbf` stays inert as `\textbackslash{}textbf\{this\}`, single asterisks are left alone, and an unmatched `**` stays literal. The NOT_COVERED block makes the non-bold side explicit rather than implied, which is what keeps a future change to the regex honest. Hermetic (mkdtemp with `finally` cleanup), neighbour-style.

One coordination note: #3353 lands the same feature and test as a bridge against main. This one carries the fuller test and reads mergeable on its own, so it is the one worth taking; I am recommending #3353 close in its favour. Green on CI. Approving.

PR #3359, opened 2026-08-26
fix: resolve cv generator main-guard conflict

chipoto69:
## Summary
- Resolves the PR #3338 merge conflict against current `origin/main` without touching Rudy's dirty/divergent primary career-ops worktree.
- Keeps the PR #3338 `isMainModule(import.meta.url)` conversions for the CV/document generator entrypoints.
- Preserves current main's `validateFlags` guard in `application-artifacts.mjs` and shrinks the `PENDING` ratchet only for now-converted entrypoints.

## Test Plan
- `git diff --check`
- `node --test tests/main-guard-convention.test.mjs` — 7/7 passing
- `npm run lint` — 562 .mjs syntax clean
- `npm install --no-package-lock --ignore-scripts` to install local deps without adding a lockfile
- `node test-all.mjs --quick` — 5579 passed / 0 failed / 2 warnings

Context: bridge replacement for santifer/career-ops PR #3338, which is approved but `DIRTY/CONFLICTING` and whose contributor branch rejected direct push from chipoto69.

## Summary by CodeRabbit

* **Tests**
  * Added coverage tracking the remaining entrypoints that use the legacy main-module convention.
  * Established a ratchet to ensure the tracked set only decreases as conversions are completed.


Scott-Emberson (commented):
Reviewed as code owner of `tests/`. Scope is `tests/main-guard-convention.test.mjs`; the converted cv-generator scripts are outside my area, read for context.

On the test itself I have no concern. This is a clean keep-both resolution: the PENDING ratchet shrinks to exactly the four still-unconverted files (`plugin-audit.mjs`, `plugins.mjs`, `update-system.mjs`, `validate-plugin-registry.mjs`), the ten dropped entries are each converted to `isMainModule()` in the same diff so they satisfy the offenders and dead-entry checks, there are no conflict markers, the file parses and is discovered, and the behaviour sentinels (the junction-symlink CLI run, both list checks, the detector self-test) are intact. #3338 is the conflicting sibling, so this is a legitimate bridge rather than a stale duplicate, and nothing is lost in the merge.

I am holding the approval rather than posting it, for one reason that is not about the test: the OS `test` legs have not run on this head. The Tests workflow is awaiting a maintainer to approve the run (first run for this contributor), so only CodeRabbit, GitGuardian and the labeler have executed. I do not sign off on a green suite I have not seen. Once a maintainer approves the run and it passes, this is good to go from my side with nothing further to change.

Scott-Emberson (commented):
Reviewed at `cba23b2e` as code owner of `/tests/`. This PR is one file and that file is mine, so this is the whole of it.

I do not think this should merge as it stands, for two independent reasons. Neither is about the quality of the work.

**1. The 22 lines it adds do nothing.** The change is a single `const PENDING = new Set([...])` at `tests/main-guard-convention.test.mjs:257`, and nothing in the file reads it:

```
$ grep -n PENDING tests/main-guard-convention.test.mjs
255:// When the last entry goes, delete PENDING, its uses, and this comment. A
257:const PENDING = new Set([
```

Only the declaration and one mention inside its own comment. I deleted the whole block and re-ran the suite:

```
with the block:     # pass 9, # fail 0
without the block:  # pass 9, # fail 0
```

Identical. The PR's entire diff is `1 file changed, 22 insertions(+)`, so removing all of it changes no behaviour at all.

That matters more than dead code usually does, because of what the comment above it promises: "the dead-entry check below FAILS if a listed file no longer references the entry path, so a conversion cannot land without shrinking it". There is no such check in the file at this head. Anyone reading `plugin-audit.mjs`, `plugins.mjs`, `update-system.mjs` and `validate-plugin-registry.mjs` sitting in that list will believe those four are ratcheted and will be caught when converted. They are not. A comment that describes a guard the code does not implement is worse than no comment, because it stops the next person from checking.

`main` currently has no `PENDING` in this file at all, so this is reintroducing the scaffolding of a ratchet without the ratchet.

**2. The PR it bridges has already merged.** #3338 is `MERGED` (head `034ce6c3`). The bridge was opened because that branch was `DIRTY/CONFLICTING` and would not take a push, and that situation has since resolved itself.

So my read is that this can simply be closed rather than fixed. If some part of the `#3170` ratchet is genuinely still wanted on top of what merged, it needs the check that reads `PENDING` to come with it, otherwise the list is documentation pretending to be a test.

Leaving this as a comment rather than a block: whether to close it is santifer's call, not mine, and there is nothing here that is wrong so much as nothing here that does anything. Suite is green either way, 9 checks, 0 failed.

PR #3361, opened 2026-08-26
fix: resolve scan-history local-day conflict

chipoto69:
## Summary
- Bridges approved conflicting PR #3241 onto current `main` without touching Rudy's dirty primary `career-ops` worktree.
- Keeps the local-day scan-history writer changes for HN and Interamt.
- Preserves current main's scan-history local-day rationale in `scan-ats-full.mjs` and resolves duplicate/import drift in the expanded gate test.
- Refreshed once after `main` advanced during the tick; latest pushed head is `94b83832b0ee16a297114b2c68eeb1d0dabfb6a3`.

## Test Plan
- `git diff --check`
- `npm run lint` → 563 `.mjs` syntax checks passed
- `node --test tests/local-today-gates.test.mjs` → 19/19 passed
- `node test-all.mjs --quick` → 5616 passed, 0 failed, 2 warnings

Context: steward bridge for santifer/career-ops#3241 (`fix(scan-history): four writers, two calendars`). No force-push; conflict resolved in isolated workspace `/Users/rudlord/ORGANIZED/_agent-workspaces/github-steward/career-ops-pr3241-conflict-20260826-1229`.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that scan history uses the local calendar day for recording and date comparisons.
  * Documented timezone handling to ensure scans remain consistent across regions.

* **Tests**
  * Expanded validation of local-day handling throughout scan-history workflows.
  * Improved test-source analysis to avoid false detections and ensure accurate test discovery.


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/local-today-gates.test.mjs`; the scanner and reporting sources are outside my area, read for context.

This resolves the recurring conflict at the tail of this file cleanly. No conflict markers, the file parses and is discovered, and every prior sibling gate family survives exactly once across the nineteen test titles: the frozen-clock TZ premise and localToday check, the scan/company-history/check-table-freshness cases, weekly-digest, rejection-latency, and the scan-history census, with the pre-existing scan-history tests coexisting rather than being duplicated. The real sentinels are preserved, the writer census reddens on the UTC-day bug and the two-dates-one-posting case reddens on mixed stamping, and `detectReposts`/`parseScanHistory` are real exports at head. #3345 is closed and unmerged, so this is the surviving resolution, not a competing duplicate. Nothing lost in the merge. Green on CI. Approving.

Scott-Emberson (approved):
Re-approving at `681de6ff`. My earlier approval sat on `0e0d8588`, and `c22c034e` ("address local-day gate review nits") landed inside `tests/local-today-gates.test.mjs` afterwards, so it needed looking at rather than carrying over. Everything else since is main merges.

Two changes, both good:

**The header comment now matches the code.** It used to say regex literals are not distinguished from division, and that a `/.../ ` argument would make the gate fail loudly. That stopped being true once `startsRegex()` landed: `tests/local-today-gates.test.mjs:430` calls it and masks the literal's contents as DATA. A comment that describes the opposite of what the function does is the kind of thing that stops the next reader from checking, so correcting it is worth more than it looks.

**`sourceFiles()` now skips `tests`.** Right call for a gate whose whole job is to find production writers. I checked it does not open a hole, by planting a deliberately bad writer in each place and running the suite:

```
# a UTC-stamped appendToScanHistory() call in ./mutant-writer.mjs
not ok 17 - every scan-history writer stamps the local day
# pass 17, fail 1

# the same file under tests/scratchdir/
ok 17 - every scan-history writer stamps the local day
# pass 18, fail 0
```

So the gate still catches a real production writer that stamps a UTC day, and no longer trips over a fixture. That is exactly the split the comment claims.

Suite at `681de6ff`: 18 tests, 18 pass, 0 fail. CI green across the board including both `test` runners.

Nothing further from my side.

PR #3367, opened 2026-08-26
fix: resolve domain-filter board gate conflict

chipoto69:
## Summary
- Bridge/handoff PR for santifer/career-ops PR #3304, which remains merge-conflicted/dirty on the original contributor branch.
- Preserves the domain_filter board-gate work and steward conflict-resolution fixes on a branch Rudy can review without overwriting `IaroslavMazur:feat/domain-filter-board-gate`.
- Refreshed against current `origin/main` (`ca6ef860955b`) by normal merge commit; no force-push and no dirty primary worktree edits.

## Test plan
- `npm run lint`
- `node tests/domain-filter-board-gate.test.mjs`
- `node tests/validate-portals-domain-filter.test.mjs`
- `node tests/scan-ats-full-resume.test.mjs`
- `node test-all.mjs` — 5671 passed, 0 failed, 6 warnings (warnings are pre-existing/expected local data-surface warnings, including santifer.io personal-data scanner hits in hired-wall files)

## Steward notes
- Original PR: https://github.com/santifer/career-ops/pull/3304
- Steward workspace: `/Users/rudlord/ORGANIZED/_agent-workspaces/github-steward/career-ops-pr3304-conflict-20260826-1500`
- Head SHA: `6751ab088c34804c2cb192c545c8547c3aac4f6d`
- This PR is a bridge, not a self-approval. Review-required gate remains.

## Summary by CodeRabbit

- **New Features**
  - Added an optional domain filter for full ATS scans.
  - Boards are admitted when at least one posting matches configured domain keywords before title filtering.
  - Supports whole-word, `stem:`, `word:`, acronym, and grouped keyword matching.
  - Scan summaries and JSON output report skipped boards and postings.
  - Resume checkpoints prevent continuation with different domain-filter settings.

- **Bug Fixes**
  - Invalid domain-filter configurations are detected during validation.

- **Documentation**
  - Updated configuration guidance and examples for domain filtering.


Scott-Emberson (commented):
Reviewed as code owner of `tests/` at head `3016e48c35`. Scope is `tests/domain-filter-board-gate.test.mjs`, `tests/scan-ats-full-resume.test.mjs` and `tests/validate-portals-domain-filter.test.mjs`; `AGENTS.md` is santifer's and the scan/title/validate sources are outside my area, all read for context.

On the tests I have no concern. All three are clean: no conflict markers, each parses and is discovered, and the resume suite is strictly additive with its pre-existing cases preserved. The latest commit sharpens it in a good way, splitting the legacy-checkpoint case into an absent-fingerprint one and an explicit-ungated one, both still asserting rejection, which is a finer sentinel than the single case it replaced. The two-sided sentinels trace to head: the board gate admits a board with a domain-bearing posting and skips one without via `boardInDomain`, and the resume suite rejects a mismatched and both flavours of ungated fingerprint via `domainFilterFingerprint`. #3304 is open but conflicting, so this is its clean bridge onto main rather than a competing duplicate, and nothing is lost.

I am holding the approval rather than posting it, for one reason that is not about the test: the OS `test` legs have not run on this head. The Tests workflow is awaiting a maintainer to approve the run (first run for this contributor), so only CodeRabbit, GitGuardian and the labeler have executed, and I do not sign off on a green suite I have not seen. Once a maintainer approves the run and it passes, this is good to go from my side with nothing further to change.

Scott-Emberson (changes requested):
Reviewed at `68d80875` as code owner of `/tests/`. Three owned files: `tests/domain-filter-board-gate.test.mjs`, `tests/validate-portals-domain-filter.test.mjs`, `tests/scan-ats-full-resume.test.mjs`. `AGENTS.md`, `scan-ats-full.mjs`, `validate-portals.mjs`, `title-keywords.mjs` and the docs are outside my area; I have read them for context and judged none of them.

Worth saying first that this is not a pure replay. Two of the three owned files are byte identical to #3304 at `d1c1f816`, but `tests/scan-ats-full-resume.test.mjs` is not: the bridge adds four checks around `domainFilterFingerprint` and checkpoint compatibility that were not in the PR I approved. So they get a first review here rather than inheriting one.

All three suites are green at `68d80875`: 14, 4, and 19 checks, 0 failed. CI is green on every check.

The board-gate and validate-portals suites are good. `domain-filter-board-gate` covers the prefix forms (`stem:`, `word:`, acronyms, AND-groups) on both the matching and the non-matching side, and it pins the documented `stem:` over-match rather than pretending it does not exist, which is the right way to handle a known edge. The ragged-array and all-malformed-rows cases are the ones that usually get skipped, and they are here.

**One gap in the new checkpoint work, and it is the case that matters most in practice.** `tests/scan-ats-full-resume.test.mjs:116-128`, major, CONFIRMED.

The four new checks are all negatives: a mismatched gate is rejected, a legacy checkpoint is rejected, an explicitly ungated checkpoint is rejected. Nothing asserts that a run with `domain_filter` set can resume against a checkpoint carrying the *same* fingerprint, which is the entire happy path of the feature.

That is not theoretical. I replaced the comparison at `scan-ats-full.mjs:138` with one that refuses to resume whenever a gate is configured at all:

```
    && (opts.domainFilterFingerprint ?? null) === null;
```

and re-ran the suite:

```
✅ domainFilterFingerprint normalizes case, whitespace, and duplicates
✅ domain_filter mismatch rejected for resume
✅ new domain_filter rejects legacy absent-fingerprint checkpoint
✅ new domain_filter rejects explicit ungated checkpoint
```

Four for four, on an implementation where `--resume` is permanently broken for every user who configures `domain_filter`. A multi-hour sweep would restart from zero each time and nothing in the suite would say so.

The existing positive control at line 104 does not cover this, because `cp` and `opts` both carry `domainFilterFingerprint: null` there, so it only proves the ungated path still resumes. One line closes it:

```js
if (checkpointCompatible({ ...cp, domainFilterFingerprint: solanaGate }, { ...opts, domainFilterFingerprint: solanaGate })) pass('a matching domain_filter still resumes');
else fail('a matching domain_filter blocked resume');
```

With that in, the four negatives mean what they say, because something proves the positive is still reachable.

Requesting changes on that one assertion. Everything else in my area is sound, and I will turn it round quickly.

Scott-Emberson (approved):
Re-reviewed at head 230507beaa. My earlier request-changes is addressed: the only delta from the previous head is the +3 lines in `tests/scan-ats-full-resume.test.mjs` adding the positive-direction assertion ("matching domain_filter resumes from checkpoint"). That closes the dead-positive gap I flagged, where the resume test asserted only the mismatch-rejected direction and could have passed vacuously if `checkpointCompatible` always rejected on the fingerprint. That's exactly the fix I was after.

The three owned tests hold to the bar: `domain-filter-board-gate.test.mjs` asserts real `boardInDomain` true/false outcomes (threshold-of-one, whole-word rejection of Deficiência / Defiance / Software Defined vs stem:/word: recovery), `scan-ats-full-resume.test.mjs` now pins `checkpointCompatible` in both directions plus legacy-absent-fingerprint rejection, and `validate-portals-domain-filter.test.mjs` exercises real CLI exit codes on scalar/bad/valid/absent fixtures. A regression in the board-gate behavior reddens these.

ubuntu and macos legs are green; the windows leg is still running (not blocked). Approving from the tests/ side; the merge stays gated on CI finishing green as usual.

PR #3370, opened 2026-08-26
fix: resolve scan import side-effect conflict

chipoto69:
## Summary
- Replays PR #3190 onto current `main` as a clean bridge branch.
- Preserves the original fix: `scan.mjs` no longer creates `data/` at import time, and scan-run summaries create their parent directory at write time.
- Resolves the current `scan.mjs` conflict by keeping main's scan-runs header drift guard and adding the PR's lazy `mkdirSync(path.dirname(filePath), { recursive: true })` before the guard.

## Test Plan
- conflict-marker scan on `scan.mjs`
- `npm run lint`
- `node tests/scan-output-paths.test.mjs`
- `node test-all.mjs --quick` was also attempted; it exercised thousands of tests but failed on pre-existing local missing `playwright` dependency paths (`generate-pdf.mjs`, `archive-posting.mjs`, `img-to-pdf.mjs`) plus `tests/zh-minimal-template.test.mjs`, so not counted as a green gate for this bridge.

## Steward context
Original PR #3190 is approved and green but still `CONFLICTING` against `main`. This bridge is intentionally isolated from Rudy's dirty/divergent primary worktree at `/Users/rudlord/ORGANIZED/JOBSEARCH/career-ops`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented unnecessary creation of empty data directories when scanning functionality is loaded.
  - Scan results now reliably create required parent directories before being saved, including nested output paths.
  - Improved handling of scan output paths without affecting existing command-line and reporting behavior.

- **Tests**
  - Added end-to-end coverage for nested output directories and import-only behavior.
  - Expanded validation across scanning, parsing, deduplication, plugin checks, batch execution, and statistics.


Scott-Emberson (commented):
Reviewed as code owner of `tests/`. Scope is `tests/scan-output-paths.test.mjs`; `scan.mjs` is outside my area, read for context.

On the test I have no concern. Clean keep-both: cases one through five are untouched, case six is appended only, no conflict markers, it parses and is discovered. The both-sides sentinel survives at head: the import-time `mkdirSync('data')` is gone from scan.mjs so an import-only spawn creates no `data/`, and all three write paths create their directory at write time. Case six pins exactly that removed import side-effect (#3159). #3190 is open but conflicting, so this is its clean bridge rather than a competing duplicate, and nothing is lost.

I am holding the approval because the OS `test` legs have not run on this head (Tests workflow awaiting a maintainer to approve the run for this contributor); only CodeRabbit, GitGuardian and the labeler have executed. I do not sign off on a green suite I have not seen. Once the test legs run and pass, this is good to go from my side.

Scott-Emberson (dismissed):
Reviewed at `b8a3a798` as code owner of `/tests/`. `tests/scan-output-paths.test.mjs` is the only file here that is mine. `test-all.mjs` is deliberately not, and I have not commented on it or on the rest of the diff.

The owned test file is byte identical to the version I approved on #3190, which is what a replay bridge should look like. The thing worth checking on a bridge is not the test, it is whether the conflict resolution in `scan.mjs` kept the behaviour the test exists to pin, since the merge is where that gets lost quietly.

It did. Suite at `b8a3a798`: 6 checks, 0 failed, including `importing scan.mjs does not create data/ as a side effect (#3159)`. I confirmed that check still bites by hoisting a top-level `mkdirSync('data', { recursive: true })` back into `scan.mjs` and re-running:

```
❌ importing scan.mjs created a data/ directory with no write ever requested (#3159)
```

So the guard is live against the merged `scan.mjs`, not just against the version it was written for. The lazy `mkdirSync(path.dirname(filePath), { recursive: true })` sits at `scan.mjs:2023` and `scan.mjs:2069`, at the write sites, which is the shape #3190 asked for.

CI is green across the board.

Approving from my side.

Issue #3016, opened 2026-08-18
merge-tracker writes rows narrower than a customized header, making them unparseable by set-status


andrgavrilenko:
## Summary

`buildRow()` in `merge-tracker.mjs` writes tracker rows from a **hardcoded column list** (`num / date / company / [via] / role / [location] / score / status / pdf / report / notes / [url]`), ignoring the actual width of the `applications.md` header. On any customized layout that carries columns career-ops has no field for, every merged row comes out **narrower than the header** — and header-driven readers (`set-status.mjs`, everything built on `tracker-parse.mjs`) then refuse to parse those rows. The application is "in the tracker" but its status can never again be changed through the supported write path.

Verified present on v1.27.0 (current `main`, `fdda561`).

## Background: readers are header-aware, the writer is not

Since #946/#954 the tracker layout is explicitly customizable: `tracker-parse.mjs` maps columns **by header name** (`detectColumns` / `headerSchemaMap`), and `parseTrackerRow` enforces a dynamic width guard — a row must span the full header width, otherwise a missing interior cell would silently shift every later column one position left:

```js
// tracker-parse.mjs
const width = Math.max(...Object.values(colmap)) + (line.trimEnd().endsWith('|') ? 2 : 1);
if (parts.length < width) return null;
```

That guard is correct — but it means the *writer* must produce rows of exactly the header's width. `buildRow()` doesn't: it appends a fixed tail (`score, status, pdf, report, notes, [url]`) after the optional Via/Location columns and never looks at the header at all. The reader half of the customizable-layout contract was implemented; the writer half was not.

## Concrete failure

Take a 10-column tracker (a real-world layout: PDFs and cover letters tracked as "Materials", plus an apply link and a follow-up date):

```
| # | Date | Company | Role | Score | Status | Materials | Apply Link | Follow-up | Notes |
```

`merge-tracker.mjs` happily detects the header (all five required fields present), then writes **9-cell rows** under the 10-column header. Two distinct corruptions result:

1. **Width mismatch** — the row is one cell short, so `parseTrackerRow`'s width guard rejects it. `set-status.mjs` reports the row as unparseable/not found; the merged application becomes unaddressable. Because nothing fails at merge time, the damage is only discovered later, when a status update silently can't be applied.
[...]

PR #3381, opened 2026-08-26
fix(merge-tracker): --backfill-urls leaves unfillable rows one cell s…

kevind589:
hort

## What does this PR do?

`--backfill-urls` rebuilds the tracker rows it can fill and hands back the rows it cannot — unchanged, and therefore one cell narrower than the header. Every reader built on `tracker-parse.mjs` then refuses those rows permanently, so `set-status.mjs` reports `No tracker row with #N` for a row that is plainly there. This routes all three of those branches through the same header-aware `buildRow` that #3016 introduced.

## Related issue

No issue — bug fix. Directly complementary to **#3016** / `8668ac1`, which made `buildRow` write at the header's actual width; this fixes the call sites that were bypassing it.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

---

## Summary

`--backfill-urls` rebuilds the rows it *can* fill and returns the rows it *cannot* unchanged. On a tracker whose rows predate the URL column, an unfillable row therefore stays **one cell narrower than the header** — and every reader built on `tracker-parse.mjs` then refuses it for good.

This is the same failure class as **#3016**, from the other direction. That issue established that the reader's width guard is *deliberate and correct*:

> `parseTrackerRow` enforces a dynamic width guard — a row must span the full header width, otherwise a missing interior cell would silently shift every later column one position left.

**This PR agrees with that guard.** It fixes the writer that was producing rows the guard correctly rejects.

## The defect

The backfill is the one place this can be repaired, because of a second asymmetry: `parseAppLine` guards with `parts.length <= maxIdx`, while `parseTrackerRow` guards with `parts.length < width` (`maxIdx + 2` for a pipe-terminated row). **The backfill is one cell more permissive than the readers — so it can still see, and repair, precisely the rows they cannot read.** It sees them and hands them back unchanged.

```js
if ((app.url || '').trim()) { already++; return line; }             // ← original width
const resolved = resolveReportUrl(app.report);
if (resolved.reason === 'no-report') { noReport++; return line; }   // ← original width
if (resolved.reason === 'no-url')    { noUrl++;    return line; }   // ← original width
filled++;
return buildRow({ ...app, url: resolved.url });                     // ← header width
```

**Three** of the four branches return `line`, so the row keeps its old width forever:

* the row has no report link at all;
* the linked report genuinely has no `**URL:**` — a referral, a warm-contact lead, a confidential pre-posting enquiry;
* **the row already has a URL.** This one still matters: on a layout carrying user-owned columns *after* `URL`, a row can hold a URL and *still* be short of the header. `--backfill-urls` reports it as `1 already set` and leaves it unreadable.

## Reproduction — clean install, no local state

Run from a checkout of `main`:

```bash
mkdir -p /tmp/repro/reports

cat > /tmp/repro/applications.md <<'EOF'
# Applications Tracker

| # | Date | Company | Role | Score | Status | PDF | Report | Notes | URL |
|---|------|---------|------|-------|--------|-----|--------|-------|-----|
| 1 | 2026-08-01 | Acme | Director, Test | 4.5/5 | Evaluated | ✅ | [1](reports/1-acme.md) | warm contact, no public posting |
EOF

cat > /tmp/repro/reports/1-acme.md <<'EOF'
# Evaluation: Acme — Director, Test

**URL:** none — confidential pre-posting enquiry, no public JD
EOF

CAREER_OPS_TRACKER=/tmp/repro/applications.md node merge-tracker.mjs --backfill-urls
CAREER_OPS_TRACKER=/tmp/repro/applications.md node set-status.mjs --row 1 Applied --dry-run
```

**Before this PR:**

```
✅ Backfill URLs: 0 filled, 0 already set, 0 no/missing report, 1 report has no **URL:**.
   row is 11 fields; header is 12
❌ Tracker at /tmp/repro/applications.md has no data rows
```

**After:**

```
✅ Backfill URLs: 0 filled, 0 already set, 0 no/missing report, 1 report has no **URL:**.
   row is 12 fields
✅ #1 Acme — Director, Test: would set Evaluated → Applied
```

The backfill's own counters are unchanged in both runs — it still correctly reports the row as unfilled. Only the width changes.

### The `already set` branch

A row can hold a real URL and *still* be short of the header, when the layout carries a user-owned column **after** `URL` — a hand-kept `Follow-up` date, say. The backfill sees the URL, counts the row as `already set`, and returns it untouched at its old width.

```bash
mkdir -p /tmp/repro2/reports

cat > /tmp/repro2/applications.md <<'EOF'
# Applications Tracker

| # | Date | Company | Role | Score | Status | PDF | Report | Notes | URL | Follow-up |
|---|------|---------|------|-------|--------|-----|--------|-------|-----|-----------|
| 1 | 2026-08-01 | Acme | Director, Test | 4.5/5 | Evaluated | ✅ | [1](reports/1-acme.md) | referred by a colleague | https://acme.example/jobs/1 |
EOF

cat > /tmp/repro2/reports/1-acme.md <<'EOF'
# Evaluation: Acme — Director, Test

**URL:** https://acme.example/jobs/1
EOF

CAREER_OPS_TRACKER=/tmp/repro2/applications.md node merge-tracker.mjs --backfill-urls
CAREER_OPS_TRACKER=/tmp/repro2/applications.md node set-status.mjs --row 1 Applied --dry-run
```

**Before this PR:**

```
🧭 Detected URL column (deterministic dedup active).
✅ Backfill URLs: 0 filled, 1 already set, 0 no/missing report, 0 report has no **URL:**.
   row is 12 fields; header is 13
❌ Tracker at /tmp/repro2/applications.md has no data rows
```

**After:**

```
🧭 Detected URL column (deterministic dedup active).
✅ Backfill URLs: 0 filled, 1 already set, 0 no/missing report, 0 report has no **URL:**.
   row is 13 fields; URL cell still https://acme.example/jobs/1
✅ #1 Acme — Director, Test: would set Evaluated → Applied
```

The URL is preserved, not blanked — the row is rebuilt from its own cells. Padding a row and losing its URL would be a worse defect than the one being fixed.

## Why it is worth fixing

**The affected rows are not a random subset.** A row has no URL when the lead did not come from a job board — which is the definition of a referral, a warm-contact lead, or a confidential pre-posting enquiry. So the rows the canonical write path cannot address are systematically the *best-sourced* ones a user has.

The failure is also permanent rather than transient. `--backfill-urls` is the documented repair for a tracker that predates the URL column, so a user who hits this runs the one tool built to fix it, sees a success line, and still cannot change those rows' status. There is no second pass that would catch them.

**The diagnostics point away from the cause.** `set-status.mjs` reports `No tracker row with #N` (or, as above, `has no data rows`), which reads as *the row does not exist* rather than *the row is one cell short*.

## The fix

Rebuild on all three branches. Unfillable rows get an empty URL cell; the already-set row is rebuilt from its own cells and keeps its URL.

**The delimiter is added, never a value.** `buildRow` writes `put('url', cell(o.url) || '')`, and an empty URL falls back to the report-number and fuzzy company+role dedup tiers exactly as it does today — no dedup behaviour changes. Nothing is fabricated.

`buildRow` is already the function this code path uses for filled rows. With `o.raw` set it copies the row's existing cells verbatim before overwriting only the fields career-ops owns, so user-owned columns survive and the report link is written back unchanged (`normalizeReportLink` runs on other paths, not here).

## Code paths this change creates or alters

| Path | Covered by the existing suite? | Covered here |
|---|---|---|
| `no-url` — report exists, has no `**URL:**` | ❌ no | ✅ padded, readable, empty cell |
| `no-report` — no report link at all | ❌ no | ✅ padded, readable |
| `already set` — row already has a URL, layout wider than `URL` | ❌ no | ✅ padded, readable, URL preserved |
| fillable — report carries `**URL:**` | ✅ yes | ✅ re-asserted as a control |
| other cells survive the rebuild | ❌ no | ✅ every mapped field compared |
| idempotence on re-run | partially (fill path) | ✅ asserted for the padded row |
| counters still report unfilled | ❌ no | ✅ `0 filled … 1 report has no **URL:**` |

Per **#2879**, the tests live in their own auto-discovered file — `tests/merge-tracker-backfill-row-width.test.mjs` — not appended to `test-all.mjs`.

**Verified the tests fail without the fix**: restoring `merge-tracker.mjs` from pristine `main` and keeping the test file fails **6 of 9** cases. The three that still pass are the fillable control, the counter assertion (width-independent), and idempotence (vacuous while the row stays short) — stated plainly rather than counted as evidence.

## Testing

```
node test-all.mjs   →  5633 passed, 0 failed, 2 warnings
```

Full suite, not `--only`, on a clean checkout of `main` plus this branch.

## Notes

Written with Claude Code, which this project uses itself. Every figure above is measured on the branch rather than estimated: both reproductions were run on a clean tree in both directions, the suite count is a full run, and the failure-without-fix check was run by restoring `merge-tracker.mjs` from pristine `main` (`cd378fb`) and re-running the test — stashing only the uncommitted hunk undercounted it at 2 of 9.

One thing worth surfacing about the tests themselves: an earlier draft asserted on `parsed.url` from `parseTrackerRow`, which exposes no `url` field — so `(parsed.url ?? '') === ''` passed whether or not the cell was empty. The assertions now read the cell out of the row directly. It was the third branch's failure that exposed it.

The diff is 4 functional lines plus explanation; the rest is the test file.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved URL backfilling to consistently preserve existing row data.
  * Ensured tracker rows match the header width, including rows with missing or extra cells.
  * Maintained accurate counts for filled, unresolved, and already populated URLs.
  * Repeated backfill runs now produce consistent results without altering valid data.

* **Tests**
  * Added coverage for fillable rows, missing URLs or reports, existing URLs, varying header widths, and repeated runs.

### DIFF
--- a/merge-tracker.mjs
+++ b/merge-tracker.mjs
@@ -864,13 +864,33 @@ if (BACKFILL_URLS) {
     if (!line.startsWith('|')) return line;
     const app = parseAppLine(line);
     if (!app) return line;
-    if ((app.url || '').trim()) { already++; return line; }
+    // Rebuilt, not returned verbatim — a row can carry a URL and STILL be short
+    // of the header on a layout with user-owned columns after `URL`. Returning
+    // `line` there leaves the same unreadable row this fix exists to prevent.
+    if ((app.url || '').trim()) { already++; return buildRow({ ...app }); }
     // Shared derivation with the merge loop — one place that knows how to turn a
     // report link into the posting URL. Tracker links may be root-relative
     // (`reports/...`) or data-relative (`../reports/...`); both resolve.
     const resolved = resolveReportUrl(app.report);
-    if (resolved.reason === 'no-report') { noReport++; return line; }
-    if (resolved.reason === 'no-url') { noUrl++; return line; }
+    // A row we cannot fill is still rebuilt, with an empty URL cell.
+    //
+    // Returning the original line leaves a pre-URL-column row one cell short of
+    // the header, and #3016 established that a short row is exactly what
+    // parseTrackerRow must reject: "a row must span the full header width,
+    // otherwise a missing interior cell would silently shift every later column
+    // one position left". That guard is correct; this writer was producing rows
+    // it correctly refuses. Every reader built on tracker-parse.mjs then skips
+    // the row for good — set-status.mjs reports "No tracker row with #N".
+    //
+    // parseAppLine's guard is `parts.length <= maxIdx`, one cell looser than
+    // parseTrackerRow's `parts.length < width`, which is why the backfill can
+    // still see and repair precisely the rows the readers cannot.
+    //
+    // The DELIMITER is added, never a value: buildRow writes `put('url', '')`,
+    // and an empty URL falls back to the report-number and fuzzy company+role
+    // dedup tiers exactly as it does today. Nothing is fabricated.
+    if (resolved.reason === 'no-report') { noReport++; return buildRow({ ...app, url: '' }); }
+    if (resolved.reason === 'no-url') { noUrl++; return buildRow({ ...app, url: '' }); }
     filled++;
     return buildRow({ ...app, url: resolved.url });
   });
--- a/tests/merge-tracker-backfill-row-width.test.mjs
+++ b/tests/merge-tracker-backfill-row-width.test.mjs
@@ -1,0 +1,186 @@
+// tests/merge-tracker-backfill-row-width.test.mjs — `--backfill-urls` must leave
+// every row it touches at the header's full width, including rows it cannot fill.
+//
+// A row written before the URL column existed is one cell short of the header.
+// #3016 established that a short row is exactly what parseTrackerRow must
+// reject — "a row must span the full header width, otherwise a missing interior
+// cell would silently shift every later column one position left". The guard is
+// correct; the backfill was returning such rows unchanged, so they stayed
+// permanently unreadable to every reader built on tracker-parse.mjs.
+//
+// Driven as a CLI integration test through the CAREER_OPS_TRACKER /
+// CAREER_OPS_ADDITIONS overrides, matching tests/merge-tracker.test.mjs:
+// importing merge-tracker.mjs runs the CLI at import time.
+import { pass, fail, NODE, ROOT } from './helpers.mjs';
+import { join } from 'path';
+import { execFileSync } from 'child_process';
+import { mkdtempSync, mkdirSync, writeFileSync, readFileSync, rmSync } from 'fs';
+import { tmpdir } from 'os';
+import { resolveColumns, parseTrackerRow } from '../tracker-parse.mjs';
+
+console.log('\nmerge-tracker.mjs — --backfill-urls row width');
+
+const HEADER = '| # | Date | Company | Role | Score | Status | PDF | Report | Notes | URL |';
+const SEP = '|---|------|---------|------|-------|--------|-----|--------|-------|-----|';
+
+// A row as written before the URL column existed: nine cells, no URL delimiter.
+const SHORT_ROW = (n, note) =>
+  `| ${n} | 2026-08-01 | Acme | Director, Test | 4.5/5 | Evaluated | ✅ | [${n}](reports/${n}-acme.md) | ${note} |`;
+
+/** One --backfill-urls run in an isolated workspace. Returns the tracker text. */
+function runBackfill({ rows, reports }) {
+  const dir = mkdtempSync(join(tmpdir(), 'backfill-width-'));
+  try {
+    mkdirSync(join(dir, 'reports'), { recursive: true });
+    for (const [name, body] of Object.entries(reports || {})) {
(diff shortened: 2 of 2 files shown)


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/merge-tracker-backfill-row-width.test.mjs`; `merge-tracker.mjs` is outside my area, read for context.

The sentinel is real: the test drives the actual `merge-tracker --backfill-urls` CLI and asserts each backfilled row survives `parseTrackerRow` at full header width. Before the fix the writer returned short rows verbatim, which that guard rejects, and the fix rebuilds them through `buildRow`. All three unfillable branches are covered, plus an empty-cell case that proves no fabrication, verbatim preservation of an already-set row, idempotence, and the counter. Hermetic (throwaway dirs, an isolated fixture tracker, `CAREER_OPS_TRACKER` and additions overrides, `rmSync` cleanup), neighbour-style. Green on CI across all three OS. Approving.

PR #3386, opened 2026-08-26
feat: add dashboard command center

chipoto69:
## Summary
- Replays the preserved local command-center dashboard commits onto current upstream main.
- Adds dashboard context/readiness data plus a command-center home screen.
- Wires home navigation to pipeline/progress and documents the new dashboard entry point.

## Steward decision for t_5da949d9
- Upstreamable: the 5 ahead dashboard command-center commits.
- Not replayed: dirty batch/batch-prompt.md, batch/batch-runner.sh, scan.mjs, test-all.mjs changes because current upstream has materially evolved versions of those areas and the preserved dirty hunks are a mixed local/student batch lane.
- Local-only: .omc/, batch/batch-input.full.tsv, portals.student.yml from the snapshot; they contain operator/profile-specific batch config/data and should stay out of the public system layer.
- Primary worktree was not modified.

## Test plan
- npm install
- npm run lint
- go test ./... (from dashboard/)
- node test-all.mjs (5648 passed, 0 failed, 6 warnings)

Board: t_5da949d9

## Summary by CodeRabbit

- **New Features**
  - Added a Career Command Center with readiness, next-action guidance, journey stages, momentum metrics, and pipeline/report/analytics summaries.
  - Added navigation between command center, pipeline, progress, home, and other dashboard views.
  - Added guidance for empty workspaces and missing career materials.
  - Added automatic dashboard refreshes after pipeline updates.

- **Bug Fixes**
  - Pressing `Esc` from the pipeline without an active search now returns to the command center.

- **Documentation**
  - Updated the README with the new dashboard experience and onboarding guidance.


Scott-Emberson (changes requested):
Reviewed as code owner of `dashboard/`. The command-center screen is a nice addition and the tests earn their keep: `app_test.go` pins the fresh-repo start on `viewCommand` with the "Add your CV" next action and the three routing transitions, and `career_test.go` proves the empty and populated data paths (`RequiredReady`, `PendingURLs==2`, `ReportCount==1`, metrics computed). The data wiring reads only local files, and the prior Stats feature coexists cleanly. Go CI is green on all three OS.

One thing to fix before it lands, one line.

**Stats renders all zeros on first view.** `newAppModel` (`main.go:100`) does not initialize `statsMetrics` or `evaluatedCount`. The old `main()` computed both eagerly; now they are only set in `reloadPipelineData`, which is not called at construction. So the sequence launch, `p`, `S` reaches the Stats screen before any refresh and it shows zeros against a populated tracker. It self-heals after `r` or any status edit, so it is not a crash, but it is a broken first impression on a screen whose whole job is to show numbers, and there is no test covering it.

The fix is to initialize `statsMetrics` and `evaluatedCount` in `newAppModel` (or call `reloadPipelineData` once at construction so the same path that populates them on refresh also runs at startup), and a small test asserting Stats is non-zero on a populated repo before any refresh would keep it from regressing.

Everything else is good to go from my side. Make that one change and this is ready.

Scott-Emberson (approved):
Re-reviewed at head `689e248c72` after the fix. This addresses the one thing I asked for, and it addresses it well.

`newAppModel` now hydrates the Stats state at construction, before any refresh: `statsMetrics: data.ComputeStatsMetrics(apps)` and `evaluatedCount: countEvaluated(apps)` (`main.go:129-130`), plus `enrichArchetypes(...)` so the report-derived breakdown is primed too. Better than a bare init, the same `countEvaluated` helper feeds both the constructor and `reloadPipelineData`, so the two paths cannot drift apart later, which is the failure mode a one-off fix would have left open.

And there is a real regression test for it: `TestNewAppModelInitializesStatsOnPopulatedRepo` writes a populated tracker and report to a temp dir, constructs the model with no refresh, and asserts `evaluatedCount == 2`, a non-empty `ScoreTiers`, `QualityBarPct > 0` against the 4.0 row, and the report-derived archetype (`AI Platform & LLMOps`, count 1). Those are tied to the fixture, not smoke checks, so the Stats-on-first-view path now reddens if it regresses.

Go CI is green on all three OS. Clearing my earlier request. Approving.

PR #3388, opened 2026-08-26
test(harness): move the nine root suites into tests/ (#3306)

vliggio:
## What does this PR do?

Moves the nine root-adjacent `*.test.mjs` suites into `tests/`, deletes the last hand-registered list in `test-all.mjs`, and corrects `ARCHITECTURE.md`'s testing section. With #3335 already landed, this leaves one home for tests and no registration to forget.

## Related issue

Closes #3303. Part 2 of #3306, which @santifer agreed the direction on — *"one home (`tests/`, auto-discovered), and the two hand-registered lists in `test-all.mjs` exist only as migration debt"* — and sequenced behind #3335, now merged.

## This is not the `git mv` that #3335 was

Worth stating plainly, because I costed it wrong myself before looking.

**Eight of the nine are not node:test suites.** Each carries a private harness — its own `passed`/`failed` counters, a local `ok()`/`eq()`, a printed summary, and `process.exit(1)`. Discovery refuses any suite containing `process.exit(` at `test-all.mjs:138`, and that check sits *before* the node:test branch at :154, so a straight move turns CI red with *"calls process.exit() — discovered suites must use pass/fail"*. All eight now report through `pass`/`fail` from `tests/helpers.mjs` and return; `eq()` keeps its expected/actual diagnostic in the failure message. Only `reply-matcher.test.mjs` moved untouched.

**Every one located the script under test as a sibling**, through `join(dirname(fileURLToPath(import.meta.url)), 'x.mjs')` or `new URL('./x.mjs', import.meta.url)`. Both idioms resolve one directory up now. The `cwd: dirname(scriptPath)` call sites — 17 in `detect-reposts.test.mjs` alone — needed no change: they derive from the corrected path and still land on the repo root.

## Two latent bugs surfaced

Both invisible while these ran in their own process, both real for any importer, and neither caused by the move.

**`contacts.mjs` validated `process.argv` at module scope**, outside its `isMainModule` guard. Importing it from a process started with any unrecognized flag died at import — `node test-all.mjs --quick` hit exactly that and the run stopped dead. Same family as #3159. Args are now read only when it is the entry point; CLI behaviour is unchanged (`--self-test` passes, `--bogus` still rejected). This is the only production file touched.

**`followup-cadence.test.mjs` pinned `CAREER_OPS_PROFILE` and never restored it.** Its comment explains the pin (#2268) and was correct for a standalone process. Discovered suites share one process, so it leaked forward — and `providers/_profile-keywords.mjs:25` reads that variable at module scope, so three provider suites read the cadence fixture instead of the `config/profile.yml` their own tmpdir had just written. They passed in isolation and failed in the full run, which is how it announced itself. Restored at suite end.

## The guard is deleted, not repointed

`tests/root-suite-registration.test.mjs` goes with the list it guarded. That is @artemtrofymenko's own argument, and their sign-off in advance on #3306: *"a name-based check over a list that no longer exists is worse than nothing, because it reads as protection."* Their first assertion was built to fail the moment root emptied, so it asked for exactly this.

Registrations came out whole — section headers and `fail()` message strings included, not just the entries — for the reason they measured on #3305: a filename surviving in a comment keeps a name-based check green while nothing invokes the suite. `grep` finds no root-suite filename left in `test-all.mjs`.

`test-trust-validator.mjs` and `test-salary-filter.mjs` stay in the `SCRIPTS` list. They are named `test-*.mjs`, which discovery does not match, and the comment now says so.

## ARCHITECTURE.md

It documented *"`*.test.mjs` sits next to what it tests"* — a convention covering 9 suites out of 230. It now describes `tests/`, the discovery contract (`pass`/`fail`, never `process.exit()`/`finish()`, node:test child-processed per #2828), how the split arose, and the Playwright exception. Folded in here as @santifer asked, rather than promised separately.

## Testing

`node test-all.mjs` → **6488 passed, 0 failed**, 6 warnings, all pre-existing (baseline on the same head: 5647, same 6 warnings). The **+841** is the nine suites reporting their real assertions instead of nine "script ran, exit 0" checks. `node scripts/check-syntax.mjs` clean.

Nine `SYSTEM_PATHS` entries repointed; the coverage guard is green.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] Issue opened first (#3306, #3303)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved command-line argument handling so importing the contacts functionality no longer unintentionally processes the host application’s arguments.

* **Documentation**
  * Clarified how automated tests are discovered, executed, and reported, including the exception for specially named suites.

* **Tests**
  * Consolidated test discovery and standardized pass/fail reporting across the test suite.
  * Added validation to ensure root-level test suites remain registered.
  * Improved test isolation and environment cleanup for more reliable results.


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is the ten owned files; `test-all.mjs` is the shared harness and `update-system.mjs` is the updater, both santifer's, read for context.

The thing to check on a move like this is that no coverage goes silently missing, and it does not. The nine suites move as git renames at 87 to 99 percent similarity, and each one drops its old `process.exit()`/RESULTS block and now reports through `pass`/`fail`, which is exactly what discovery requires, and the harness actively reddens any discovered suite that still calls `process.exit()` or `finish()`. The inline entries removed from the harness (the two script entries and the seven blocks) are net neutral because all nine now run via the recursive `tests/**` discovery, confirmed present under `tests/` with zero root-level `*.test.mjs` left, and `SYSTEM_PATHS` was repointed so they still ship. The two in-process side-effects are handled: the contacts import-time flag guard and the followup-cadence env restore.

One residual worth naming, not blocking. `root-suite-registration.test.mjs` is deleted rather than updated, which is correct here since it guarded root suites and self-obsoletes at zero of them, but it does mean nothing now catches a future stray root `*.test.mjs`. Given the documented doctrine that tests live under `tests/`, that is an acceptable trade, just worth being deliberate about. Green on CI. Approving.

PR #3384, opened 2026-08-26
feat: wire reply-watch into status-log.tsv (#1582)

Shaamak:
## Wire reply-watch into status-log.tsv

Closes #1582.

### What changed

- **`funnel-velocity.mjs`** — `reply-watch` added to `VALID_SOURCES` and `DAY_MATH_SOURCES`; dedicated self-test fixture asserts both parse recognition and that the hop reaches velocity figures
- **`set-status.mjs`** — `reply-watch` added to `WRITER_SOURCES` to satisfy the source allow-list boundary check
- **`reply-watch.mjs`** — appends a ledger row inside the tracker lock callback (atomic with the tracker write); failure is a non-fatal `console.warn` so a disk error never aborts the update; `--yes`/`-y` flag added for non-interactive use
- **`tests/reply-watch.test.mjs`** *(new)* — 5 tests: happy path, idempotency (no duplicate on re-run), resilience (append failure warns but tracker still updates), noise filtering, empty input

### Invariants held

- `status-log.tsv` format unchanged — new source value only, zero column/format changes
- Downstream readers unaffected; the vocabulary assertion in the suite is updated in this same PR so the diff explains itself

## Summary by CodeRabbit

- **New Features**
  - Added noninteractive confirmation mode for reply-watch updates.
  - Status changes can now be recorded with the local date in the status log.
  - Added support for tracking reply-watch updates in velocity calculations.
  - Added reply-watch as a recognized status-update source.

- **Bug Fixes**
  - Improved status recommendations by excluding irrelevant “Noise” entries.
  - Added safer handling when status-log updates cannot be written.
  - Prevented duplicate status-log entries during repeated updates.

PR #3393, opened 2026-08-27
fix(reply-watch): keep status updates interactive

chipoto69:
## Summary
- Supersedes PR #3384 after maintainer requested removing the unplanned `--yes`/`-y` path.
- Removes non-interactive approval flags from `reply-watch.mjs` so status changes remain explicitly user-confirmed.
- Updates reply-watch tests to send interactive confirmation over stdin and adds a regression that `--yes` is rejected.

## Test Plan
- `npm run lint`
- `node --test tests/reply-watch.test.mjs`
- `node test-all.mjs --quick` (5384 passed, 0 failed, 2 known warnings)

## Steward notes
- Original PR #3384 branch is contributor-owned; direct maintainer edit push to `Shaamak/feat/track-employer-replies-1582` was denied for chipoto69 despite maintainerCanModify readback.
- No dirty/divergent primary worktree was touched.
- Workspace: /Users/rudlord/ORGANIZED/_agent-workspaces/github-steward/career-ops-pr3384-drop-yes-20260827-0302

## Summary by CodeRabbit

* **New Features**
  * Added logging for confirmed application status changes, including application number, date, previous and new statuses, and update source.
  * Added `reply-watch` as a recognized source for status updates and funnel velocity tracking.

* **Bug Fixes**
  * Tracker updates now complete successfully even when activity-log writing fails.

* **Tests**
  * Added coverage for logging, duplicate prevention, empty results, unsupported options, and noise classification.


Scott-Emberson (approved):
Reviewed as code owner of `tests/`. Scope is `tests/reply-watch.test.mjs`; `reply-watch.mjs`, `set-status.mjs` and `funnel-velocity.mjs` are outside my area, read for context.

This pins the invariant that matters most on this path, that a status update stays interactive and is never auto-applied, and it pins it from two sides. The happy path only writes when a `y` is fed on stdin, and the negative test proves there is no non-interactive bypass, `--yes` is rejected as an unrecognized flag, so if reply-watch ever grew an auto-approve flag that assertion reddens. That is exactly the never-auto-submit rule this project holds, so the test is worth having. It is a clean additive follow-up to #3384 with no overlap, hermetic (throwaway dirs, `CAREER_OPS_TRACKER` override, `rmSync` cleanup, no ambient tracker or ledger touched), and neighbour-style.

One follow-up worth naming, not blocking. There is no explicit case that feeds an `n` and asserts the tracker is unchanged, the no-op path leans on the already-current branch rather than the prompt itself, so a full removal of the prompt would not redden this suite. A one-line `n` case would close that, and given how much rides on this gate it is worth adding when you next touch the file. Green on CI. Approving.

PR #3400, opened 2026-08-27
fix(scan): opt-in word:/stem: anchoring for content_filter keywords

### LINKED ISSUE #3274: bug(scan): single-token content_filter negatives match inside other words ("java" rejects every JavaScript posting)
### What

`buildContentFilter()` in `scan.mjs` matches negative keywords with plain `String.includes()`. Single tokens therefore reject postings that merely *mention* them inside another word:

```js
buildContentFilter({ negative: ['java'] })('We ship TypeScript/JavaScript')
// → false — a Node/TS role discarded because of one substring hit
```

The same collision hits `ios`: a negative `ios` rejects any posting containing the word "curiosity".

Combined with #2695 (discards leave no record), these postings vanish silently.

### Proposed fix

Anchor single tokens at word boundaries (same lookaround discipline `compileLocationKeyword()` already uses, so the two filters can't drift); multi-word phrases like `security clearance` keep exact-substring semantics since those are intended to match anywhere. Genuine standalone mentions still reject: `java` still blocks an actual Java shop, and a real `ios` negative still blocks iOS-only roles.

A tested implementation exists locally (unit tests covering both the anchoring and phrase passthrough); happy to share the diff if a PR isn't wanted from my side.

AdilAzhari:
## What

`buildContentFilter()` in `scan.mjs` matches every `content_filter` keyword with plain `String.includes()`. Single-token negatives therefore reject postings that merely *mention* them inside another word:

```js
buildContentFilter({ negative: ['java'] })('We ship TypeScript/JavaScript') // → false
buildContentFilter({ negative: ['ios'] })('We value curiosity and rigor')   // → false
```

Combined with discards leaving no record, those postings vanish silently.

## Approach

Per the direction in #3274 (and the #3103 / #3225 precedent for `title_filter`): **opt-in, per entry, default untouched.** A `word:` or `stem:` prefix asks for boundary-anchored matching on that one keyword; every unprefixed entry keeps plain substring behaviour byte-for-byte, so no configured `portals.yml` changes meaning.

- `word:java` — whole word only: keeps "JavaScript", still blocks a Java shop
- `stem:ios` — left boundary only: keeps "curiosity", still blocks "iOS"

Semantics are identical to `title_filter`'s prefixes. The shared `word:`/`stem:` handling is factored out of `compileKeyword` into `compilePrefixedKeyword` so there is still one definition. `compileContentKeyword` reuses it but **skips the title filter's 2–3 letter auto-anchor** — a short run inside a paragraph of description prose ("aws", "sql", "go") is routinely intended, unlike "COO" inside "Coordinator" in a title.

## Changes

| File | Change |
|---|---|
| `title-keywords.mjs` | Extract `compilePrefixedKeyword`; add `compileContentKeyword` |
| `scan.mjs` | `buildContentFilter` compiles `positive`, `negative`, and every `by_title_keyword` rule list through it |
| `templates/portals.example.yml` | Document the prefixes in the `content_filter` section |
| `tests/content-filter-word-prefix.test.mjs` | New suite (auto-discovered) |

`scan-ats-full.mjs` imports `buildContentFilter` from `scan.mjs`, so reverse sweeps get this for free.

## Tests

New suite (11 assertions) covers: the `plain ⊇ stem: ⊇ word:` ordering on the negative side, the unchanged substring default (incl. a full prefix-free legacy config), the absent short-acronym anchor, positive-side + `by_title_keyword` coverage, and the bare-prefix / regex-metacharacter plumbing.

- `node test-all.mjs --only prefix` → 43/43
- Full discovered suite green (one unrelated pre-existing failure on my local checkout: `SKILL.md: missing cwd-independent routing rule`).

Closes #3274

## Summary by CodeRabbit

* **New Features**
  * Added optional `word:` matching for whole-word keyword detection.
  * Added optional `stem:` matching for left-boundary keyword detection.
  * Existing unprefixed keywords continue using substring matching.
  * Matching is supported across global content filters and title-based rules.

* **Documentation**
  * Added configuration examples explaining the new keyword matching options.

* **Bug Fixes**
  * Improved content filtering accuracy by preventing unintended partial matches when boundary matching is requested.

### DIFF
--- a/scan.mjs
+++ b/scan.mjs
@@ -52,7 +52,7 @@ import { resolveColumns, parseTrackerRow, normalizeTextKey } from './tracker-par
 import { normalizeCompany } from './tracker-utils.mjs';
 import { normalizeCompanyName } from './invite-match.mjs';
 import { withPipelineLock } from './pipeline-lock.mjs';
-import { compileKeyword, compilePositiveKeyword, buildTitleFilter } from './title-keywords.mjs';
+import { compileKeyword, compilePositiveKeyword, compileContentKeyword, buildTitleFilter } from './title-keywords.mjs';
 import { flagValue, hasFlag, validateFlags } from './lib/cli-flags.mjs';
 import { withPortalHealthLock } from './portal-health-lock.mjs';
 import { localToday } from './lib/local-today.mjs';
@@ -100,13 +100,15 @@ const CONCURRENCY = 10;
 
 // ── Title filter ────────────────────────────────────────────────────
 
-// How a title_filter matches a title lives in title-keywords.mjs, because
+// How a keyword matches text lives in title-keywords.mjs, because
 // openrouter-runner.mjs filters titles too and cannot import this file (scan.mjs
 // creates data/ at import time). It called a second, hand-kept copy of this
 // logic until the two drifted; there is now one implementation and this file
 // re-exports it, so existing importers — scan-ats-full.mjs and test-all.mjs's
 // sections 11b and 44 among them — keep resolving it from here.
-export { compileKeyword, compilePositiveKeyword, buildTitleFilter };
+// compileContentKeyword shares the `word:`/`stem:` prefix machinery but skips
+// the title filter's short-acronym auto-anchor (#3274).
+export { compileKeyword, compilePositiveKeyword, compileContentKeyword, buildTitleFilter };
 
 // Compiled-matcher cache for matchedTitleKeywords(), keyed by the
 // `title_filter.positive` array reference. The scan loop calls this once per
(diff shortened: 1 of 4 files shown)


Scott-Emberson (approved):
Reviewed at head ffb96293b6. The owned test `tests/content-filter-word-prefix.test.mjs` carries the guarantee I want on an anchoring change like this: it doesn't just check the happy path, it pins the near-misses that a no-op or substring-only regression would flip.

- Section 1 uses an exact-triple compare (`['java', '...JavaScript', true, true, false]`) so `word:` must NOT match the substring inside "JavaScript" while plain and `stem:` still do; that assertion reddens the moment whole-word stops being enforced.
- The bare-keyword default is held steady: unprefixed `java` still rejects "JavaScript" and `go` still hits inside "MongoDB", so there's no accidental auto-anchoring (test:55-56, 77-78).
- `word:go` pins both a real match ("everything in go") and a near-miss ("mongo and django" → false), and `word:.net` matches ".net 8" but not "anet internal tool" — that last one guards metacharacter literalness/boundary, which is exactly where a naive regex build would break.

Wiring checks out: `compileContentKeyword = compilePrefixedKeyword(kw) ?? substring` (title-keywords.mjs) and scan.mjs routing positive/negative/by_title_keyword through the compiled matchers is what the test exercises. Auto-discovered under tests/, uses pass/fail, so it genuinely runs in CI. Green across the board. Approving from the tests/ side.

Issue #1386, opened 2026-07-01
Reorganize root: group scripts by function, co-locate tests, consolidate data dirs


mkoelle:
### Code of Conduct

- [x] I agree to follow the [Code of Conduct](https://github.com/santifer/career-ops/blob/main/CODE_OF_CONDUCT.md)

### Existing issues

- [x] I searched existing issues and this hasn't been requested yet

### Problem

**Time:** ~8-12 h total (4 independent PRs, ~2-3 h each) · **Difficulty:** Intermediate-Advanced

Idea
----

career-ops has **46 `.mjs` scripts and 34 markdown files at root** with no grouping. Every script — CLI entry point, internal library module, test, build helper — lives at the same level. The `providers/` and `plugins/` directories already demonstrate the right model (named dir, public entry points, `_`-prefixed internals); the rest of the codebase hasn't caught up. Meanwhile the five user-data directories (`data/`, `reports/`, `output/`, `jds/`, `interview-prep/`) sit as peers at root with no signal about whether they hold input, mutable state, or generated output.

This is a contributor-experience tax that compounds as the codebase grows. A new contributor picking up a `good first issue` should be able to look at the structure and know immediately where their file belongs.

The testing problem
-------------------

`test-all.mjs` is currently **8,255 lines** — a single file that owns every test assertion in the project. That design made sense as a bootstrap, but it creates a set of structural problems that get worse as the codebase grows:

**Ownership is invisible.** When a test fails at line 4,800, there's no filesystem signal about which feature or script is responsible. You trace it manually. With co-located tests, a red CI run on `scripts/tracker/merge-tracker.test.mjs` tells you exactly what broke and who should fix it.

**Every PR touches the same file.** A contributor adding a scan provider and a contributor fixing a tracker bug both have to insert into `test-all.mjs`. On an active project with concurrent contributors, that's a recurring merge conflict at the worst possible location — the test suite.

**Coverage gaps are structurally invisible.** If `liveness-api.mjs` has no tests in `test-all.mjs`, nothing in the directory listing signals that. With co-located tests, a directory scan makes the gap obvious: `liveness-api.mjs` exists, `liveness-api.test.mjs` does not. A CI lint rule or a reviewer can catch it at a glance.

[...]

PR #3411, opened 2026-08-27
fix(tests): scope the profile pin, and move the last two hand-registered suites

vliggio:
## What does this PR do?

Follow-up to #3388. Two commits: fixes the two real CodeRabbit findings that PR merged without, and finishes the job by moving the last two hand-registered suites into `tests/`.

After this, `test-all.mjs`'s `scripts` list holds no test suite at all.

---

## 1. The `CAREER_OPS_PROFILE` restore was skippable (🟠 Major)

#3388 fixed a leak: `tests/followup-cadence.test.mjs` pins `CAREER_OPS_PROFILE` so `followup-cadence.mjs`'s module-level `CADENCE = resolveCadenceConfig()` reads a fixture rather than the user's profile (#2268). Discovered suites share one process, and `providers/_profile-keywords.mjs:25` reads that variable at module scope, so leaving it set made three provider suites read the cadence fixture instead of the profile their own tmpdir had just written.

The restore landed at the **end of the file**, which is only good enough on the happy path. A failed assertion is fine — `fail()` counts, it does not throw — but an uncaught error skips the restore, and discovery *contains* the throw and runs the next suite anyway. So the leak came back on precisely the run that was already going wrong, which is the worst time to start silently swapping profiles underneath other suites.

Measured, not argued. Injecting a throw mid-suite and inspecting what the next importer sees:

```
merged #3388   CAREER_OPS_PROFILE after: ".../tests/fixtures/profile-default-cadence.yml"   LEAKED
this branch    CAREER_OPS_PROFILE after: undefined                                          NO LEAK
```

**Scoped to the import rather than wrapping the whole body.** CodeRabbit suggested a `try/finally` around the entire suite; that works but reindents ~250 lines. The pin exists for one statement — the dynamic import — because that is when `followup-cadence.mjs` reads the variable. Nothing below needs it: every later call passes `profilePath` explicitly (`loadProfileCadence(DEFAULT_CADENCE_PROFILE)`, `resolveCadenceConfig({ profilePath: ... })`). So the `try/finally` goes around the import alone. Exposure is one `await` instead of the whole file, and the diff stays readable.

## 2. ARCHITECTURE.md contradicted the harness (🟡 Minor)

I wrote *"Every suite lives in `tests/`"* in #3388 while two suites sat at the repo root, registered by hand — and `test-all.mjs`'s own comment in the same PR said exactly that. The doc and the harness disagreed on one commit.

My first fix named them as an exception that *"predates the convention"*. That was worse: it dressed an accident up as a decision. Which leads to the second commit.

## 3. The last two hand-registered suites move (@vliggio's catch)

`test-salary-filter.mjs` and `test-trust-validator.mjs` are test suites — their own headers say *"Comprehensive test suite for ..."* — and they are the same shape as the nine moved in #3388: private `passed`/`failed` counters, a local `assert()`, a printed summary, `process.exit(1)`. The only thing keeping them out of `tests/` was the filename. `test-*.mjs` does not match the discovery glob; `*.test.mjs` does. That is an accident of naming, not a design decision, and documenting it as one was the wrong call.

They are now `tests/salary-filter.test.mjs` and `tests/trust-validator.test.mjs`, converted the same way as the nine.

**231 assertions** stop reporting as two "script ran, exit 0" checks. Counts verified identical in both directions — 124 and 107 standalone before the move, 124 and 107 again under `node test-all.mjs --only salary-filter` / `--only trust-validator` after it — so nothing was dropped in the conversion.

**`test-all.mjs` itself stays at the root**, to be explicit about it. It is the runner, not a suite: the entry point CI invokes, named in the PR template, `CONTRIBUTING.md` and `AGENTS.md`, and it is what performs discovery. #1386's path-stability argument covers it fully.

ARCHITECTURE.md's exception paragraph is deleted rather than reworded, and the section now says outright that no suite is registered by hand any more — the `scripts` list holds scripts and their `--self-test` entry points, nothing else.

This also removes a carve-out from @artemtrofymenko's forthcoming stray-root guard: with these two gone, "no `*.test.mjs` at the repo root" needs no exception list.

## Not taking: provenance markers on the doc prose

The third CodeRabbit thread asks for primary-file references or provenance markers on the counts in `ARCHITECTURE.md` and the "three provider suites" figure in the cadence comment. That convention is `AGENTS.md`'s **Source-of-Truth Boundary**, which governs claims about the *user* in generated user-facing content — CVs, cover letters, outreach. It does not apply to a repo doc counting its own suites, and those numbers are checkable by reading the tree. Applying it here would be cargo-culting the rule past its scope.

## Testing

`node test-all.mjs` → **6769 passed, 0 failed**, 6 warnings, all pre-existing. `node scripts/check-syntax.mjs` clean.

The env fix is verified on the *failure* path with the throw-injection probe above, in both directions — not only by checking that the happy path still passes.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] Bug fix / follow-up to a merged PR — exempt from issue-first

## Summary by CodeRabbit

* **Documentation**
  * Clarified test suite discovery, execution behavior, migration details, and supported Playwright exceptions.
  * Documented shared test counters and requirements for reporting pass/fail results.

* **Tests**
  * Migrated salary filter and trust validator tests to the standard test layout and conventions.
  * Improved environment handling so follow-up cadence tests restore settings immediately after setup.
  * Updated test manifests and runner configuration to use the standardized test locations.


Scott-Emberson (approved):
Reviewed at head a95a708b79. The owned test `tests/followup-cadence.test.mjs` holds to the bar: it pins concrete computed dates, not throw-free runs. The responded-branch case asserts `computeNextFollowupDate('responded', APP, null, 0)` equals `APP + responded_initial` (1 day), which reddens if that branch regressed to the subsequent cadence (3 days) since the two defaults genuinely differ. The subsequent branch, the applied regression case, the exact-shape `parseFollowups`, cold-classification, and full-object `resolveCadenceConfig` equality are all concrete-value assertions.

The change itself is a safe hardening: it moves the `CAREER_OPS_PROFILE` restore into a try/finally around the dynamic import so a throw can't leak the env override into later tests, with no assertion values altered. Green across the board. Approving from the tests/ side.

Scott-Emberson (approved):
Re-reviewed at head 07680cf32b — the PR grew since my earlier approval, so I treated the new scope as a fresh review. It now also physically moves `test-salary-filter.mjs` and `test-trust-validator.mjs` into `tests/` as `*.test.mjs`, drops their hand-registration in `test-all.mjs`, updates the two `SYSTEM_PATHS` entries, and rewrites the ARCHITECTURE.md tests section.

The move preserves coverage properly, which is the thing that matters here: both suites are now matched by the recursive `discoverTests()` (test-all.mjs:109-115), they import `{pass, fail}` from `helpers.mjs` and route their existing `assert()` calls through them, and the old standalone `process.exit` summary block was removed from each — required, since the discovery guard (test-all.mjs:139/174) fails any discovered suite that calls `process.exit()`/`finish()`. So a broken move can't silently go green: the shared fail counter governs the verdict. No assertions were dropped in the rename (full `parseCompensation`/`buildSalaryFilter` and `buildTrustValidator`/`companyMatchesHostname` bodies intact), and the followup-cadence profile pin is now wrapped in try/import/finally so it can't leak into later provider suites on a throw.

Green. Approving from the tests/ side. One cosmetic nit, non-blocking: the two moved files still carry their stale `Run: node test-…` header comments / shebangs — worth stripping on a later pass, but harmless.

PR #3418, opened 2026-08-27
fix(updater): parse git status with -z so quoted paths and renames match real files

### LINKED ISSUE #3417: parsePorcelainStatus: quoted paths and rename lines produce a path that matches no file, blinding the user-layer safety check
## Code of Conduct / existing issues

- [x] I searched existing issues — this is the follow-up @santifer asked for in https://github.com/santifer/career-ops/pull/3182#issuecomment (PR review comment), split out from the rename-retry work that PR closed.

## Summary

`parsePorcelainStatus()` (`update-system.mjs:772-786`) parses each `git status --porcelain` line as `{ code: clean.slice(0, 2), path: clean.slice(3) }`. That is correct only for unquoted, non-rename lines. Two porcelain v1 shapes it does not handle both produce a `path` string that exists nowhere on disk, so the updater's "did we touch a user file?" safety check compares a mangled path and a real modification can slip past the abort — the same bug class as the leading-space trim and CRLF cases already fixed (`#3048` and the CRLF note in the same file).

### 1. C-quoted paths

When a path contains a space, a `"`, a control char, or (with `core.quotepath` at its git default of `true`) any non-ASCII byte, porcelain v1 wraps it in double quotes and C-style-escapes it. `clean.slice(3)` keeps the quotes and escapes verbatim:

```
 M "data/my notes.md"            ->  path = "data/my notes.md"   (literal quotes)
 M "data/caf\303\251.md"         ->  path = "data/caf\303\251.md" (quotes + octal escapes)
```

Neither string matches the on-disk path (`data/my notes.md` / `data/café.md`), so `initialStatusPaths.has(...)`, the `changed` list, and the commit-failure guard all miss the entry.

### 2. Rename / copy lines

Porcelain v1 renders a rename on one line as `R  <old> -> <new>` (copy: `C  <old> -> <new>`). `clean.slice(3)` returns the whole `old -> new` blob as a single `path`:

```
R  data/plain.md -> data/renamed.md   ->  path = "data/plain.md -> data/renamed.md"
```

Neither side is a real path. If the updater ever renames a user-layer file, both the pre-update `initialStatusPaths` set and the post-update `changed` list carry this blob instead of the two real paths, so the safety check cannot match either side of the rename. When a renamed path also contains a space the line additionally carries quotes on each side (`R  "old name.md" -> "new name.md"`), stacking both bugs.

## Steps to reproduce

```sh
D=$(mktemp -d); cd "$D"; git init -q -b main .
git config user.email t@e.com; git config user.name T
mkdir data; printf 'v1\n' > "data/my notes.md"; printf 'v1\n' > data/plain.md
git add -A; git commit -qm base
printf 'v2\n' > "data/my notes.md"; git mv data/plain.md data/renamed.md
git status --porcelain
#  M "data/my notes.md"
# R  data/plain.md -> data/renamed.md
node -e 'import("./update-system.mjs").then(m => console.log(m.parsePorcelainStatus(` M "data/my notes.md"\nR  data/plain.md -> data/renamed.md`)))'
# [ { code: ' M', path: '"data/my notes.md"' },
#   { code: 'R ', path: 'data/plain.md -> data/renamed.md' } ]
```

Consumers affected: `gitStatusEntries()` -> `initialStatusPaths` (`update-system.mjs:1573`), the `changed` list (`~1909`), the commit-scope guard (`~2040`).

## Expected behavior

`parsePorcelainStatus()` yields the real, on-disk repo-relative path for every entry, byte-for-byte — quoted names dequoted, rename/copy lines split into their `old` and `new` paths (both surfaced, since the safety check cares about both).

## Suggested fix

Switch the source to `git status --porcelain -z`, which is already the house pattern for exactly this reason (`update-system.mjs:1140`, `:1220` — "-z: NUL-delimited, unquoted output"). `-z` never quotes, splits fields on NUL, and emits a rename as `XY <new>\0<old>\0` (two NUL fields), so no dequoting or ` -> ` splitting is needed. Keep `parsePorcelainStatus()` as the pure, unit-testable seam; feed it the `-z` buffer. Fallback option if the seam must stay newline-based: dequote per `git-status(1)` porcelain v1 rules (strip wrapping `"`, un-escape C sequences) and split rename/copy lines on ` -> `.

## Acceptance list

One fixture-driven test in `tests/updater-status-paths.test.mjs` (its case 4 note already flags the quoted-path gap) covering all three forms at once:

1. **Quoted path** — ` M "data/my notes.md"` parses to `path: 'data/my notes.md'` (no quotes); a non-ASCII name under default `core.quotepath` parses to the real UTF-8 path.
2. **Rename** — `R  a.md -> b.md` yields the new path `b.md` (and the old path `a.md`) as real strings, not one blob; a rename with a space on either side still does.
3. **CRLF** — a CRLF-terminated buffer through `parsePorcelainStatus()` directly (Windows git EOL), asserting no trailing `\r` — already covered by case 6, keep it green under the new shape.

yaserselvam:
## What does this PR do?

`parsePorcelainStatus()` split `git status --porcelain` on newlines and sliced `path` at offset 3. Two porcelain v1 shapes came out mangled, and a mangled path is a hole in the user-layer safety check — it compares the parsed `path` against files on disk, so a mismatch lets a touched user file slip past the abort (same class as #3048):

- **C-quoted paths** — a path with a space, a quote, or (under git's default `core.quotepath`) a non-ASCII byte arrives as `` M "data/my notes.md"`` / `` M "data/caf\303\251.md"``. The slice kept the quotes and octal escapes, so the string matched no file.
- **Renames / copies** — `R  old -> new` is one line, so `path` became the blob `old -> new`, neither side a real path.

Fix: `gitStatusEntries()` now reads `git status --porcelain -z`, already the house pattern for this exact reason (`removeAdditionsNotInHead`, the staged-diff guard). `-z` is NUL-delimited, never quoted, and not line-ending translated, so it also retires the leading-space-trim and CRLF workarounds. Renames/copies come as two NUL fields (destination then origin); `parsePorcelainStatus()` emits both as their own entry, since the safety check only reads `.path`.

## Related issue

Closes #3417 (filed at @santifer's request in #3182).

## Type of change

- [x] Bug fix
- [ ] New feature

## Acceptance / tests

`tests/updater-status-paths.test.mjs`: replaced the CRLF direct-buffer case with synthetic `-z` buffers (plain, CR-as-real-byte, rename, copy) and added real-git cases for a spaced path, a non-ASCII path under default `core.quotepath`, and a `git mv` rename. All updater suites + `npm run lint` pass.

## Checklist

- [x] Bug fix — issue-first exempt (issue filed anyway: #3417)
- [x] No personal data
- [x] Ran `node test-all.mjs` — updater suites and lint pass. 6 unrelated failures are pre-existing on this machine's Node 20 (repo requires ≥22: `node:sqlite`, `node --test` web suites, sqlite-backed tracker tests); identical on the parent commit.
- [x] Respects the Data Contract — no user-layer file changes; this hardens the check that protects them
- [x] Aligns with roadmap

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Git status parsing for paths containing spaces, non-ASCII characters, and embedded carriage returns.
  * Correctly handles renamed and copied files by reporting both original and destination paths.
  * Improved reliability when processing empty or NUL-delimited status output.

### DIFF
--- a/tests/updater-status-paths.test.mjs
+++ b/tests/updater-status-paths.test.mjs
@@ -9,21 +9,27 @@
  * path is mangled no longer matches its real counterpart, so a violation can
  * slip past the abort.
  *
- * The corruption vector is `gitIn()`'s `.trim()`: `--porcelain` lines for
- * worktree/index changes begin with a space (` M path`), and trimming the WHOLE
- * buffer strips the first line's leading space, turning it into `M path`. The
- * parser's `path: line.slice(3)` then drops the path's first character — for
- * ` M data/user-file.md` the resulting `ata/user-file.md` matches nothing.
+ * Historical corruption vectors, all now closed by parsing `--porcelain -z`
+ * (NUL-delimited) instead of the newline form:
+ *   - `gitIn()`'s `.trim()`: worktree/index lines begin with a space (` M path`),
+ *     and trimming the whole buffer stripped the first line's leading space, so
+ *     `path: line.slice(3)` dropped the path's first character.
+ *   - Windows CRLF line endings gave every path a trailing `\r`; `-z` suppresses
+ *     line-ending translation, so there is nothing to strip.
+ *   - C-quoting: the newline form wraps any path with a space, a quote, or a
+ *     non-ASCII byte (under git's default core.quotepath) in `"…"` with octal
+ *     escapes; the safety check then compared a quoted string to a real file.
+ *   - renames: the newline form writes `R  old -> new` on ONE line, so the
+ *     "path" was the blob `old -> new`; `-z` writes destination and origin as
+ *     two NUL fields, and parsePorcelainStatus surfaces both as their own entry.
  *
- * The parser lives in parsePorcelainStatus() so its second disguise — a
- * Windows git whose CRLF line endings would give every path a trailing `\r` —
- * can be driven with a synthetic CRLF buffer instead of trusting whatever EOL
- * the local git happens to emit.
- *
- * These tests drive gitStatusEntries() against throwaway repos through the
- * same seam the production code uses (gitRawIn under a root), pinning the
- * whitespace-sensitive behaviour on both code shapes (` M` and `M `) and on the
- * first-line position that the old `gitIn`-based implementation corrupted.
+ * A mistyped path is a blind spot — gitStatusEntries() feeds the "did we touch
+ * a user file?" safety check (the pre-update dirty Set, the `changed` list, the
+ * commit-failure guard), all of which compare against the parsed `path`. These
+ * tests drive gitStatusEntries() against throwaway repos through the same seam
+ * production uses (gitRawIn under a root), plus a few synthetic `-z` buffers
+ * straight through parsePorcelainStatus() for the shapes a local git may not
+ * emit on demand.
  */
 
 import { mkdtempSync, mkdirSync, writeFileSync, rmSync } from 'fs';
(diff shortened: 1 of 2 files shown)


Scott-Emberson (approved):
Reviewed at head 61121e6fcd. The owned test `tests/updater-status-paths.test.mjs` is exactly the kind of test I want on a `git status` parsing change — it pins the cases where the old newline/space parser silently does the wrong thing:

- A spaced path (`data/my notes.md`) dirtied in a real repo, asserting the result is the unquoted real name — the newline parser would hand back the C-quoted `"data/my notes.md"` and it reddens.
- A non-ASCII path (`data/café.md`) under `core.quotepath: true` (git's dangerous default), asserting raw UTF-8 — the old form yields octal-escaped `"data/caf\303\251.md"` and reddens.
- A real `git mv` through `gitStatusEntries()` asserting the Set is exactly `{data/old.md, data/new.md}` (size 2) — the single-line `R old -> new` slice yields the arrow-blob and fails.
- Direct `parsePorcelainStatus` calls with NUL buffers for both rename and copy, asserting the origin is emitted as its own entry — a `split('\n')` regression collapses these.

All four are coupled to the `-z` NUL-delimited parsing, not smoke runs. Green, mergeable. Approving from the tests/ side. (Trivial non-blocker: file has no trailing newline.)

PR #3404, opened 2026-08-27
feat(providers): JobsDB Hong Kong via the existing jobstreet provider

kwjho:
## What

Adds `hk.jobsdb.com` to `ALLOWED_JOBSTREET_HOSTS` and documents the `HK-Main` site key. Two files, 29 insertions, no behaviour change for existing markets.

## Why

SEEK runs Hong Kong under the **JobsDB** brand rather than Jobstreet, but it is the same platform behind the same v5 endpoint the provider already speaks. Verified live:

```
GET https://hk.jobsdb.com/api/jobsearch/v5/search
    ?siteKey=HK-Main&sourcesystem=houston&keywords=business+intelligence

→ HTTP 200 · totalCount 5872
→ items carry advertiser.description and jobLocation.label
  e.g. "Business Intelligence Manager / Senior Business Intelligence Analyst"
       — IT Channel (Asia) Limited — Central, Central and Western District
```

Because the host was not allowlisted, every Hong Kong entry failed the SSRF guard with `jobstreet: untrusted hostname "hk.jobsdb.com"`, so HK had no zero-token path at all.

The WebSearch fallback is not an adequate substitute for this board. Three `site:hk.jobsdb.com` queries returned only category landing pages (`/business-intelligence-jobs`, `/power-bi-analyst-jobs`) — Google indexes JobsDB's facets, not its listings — so zero addressable postings reached `pipeline.md`.

More broadly, HK corporates are poorly served by the ATS providers: probing 58 endpoints across Greenhouse / Lever / Ashby / SmartRecruiters / Workday for major Hong Kong employers produced only 4 live boards. JobsDB is the practical entry point for the market.

## Why not a new provider

Same platform, same endpoint, same response shape, same parser. `parseJobstreetItem` handles the payload unchanged. A separate module would be a copy of this one with one string altered.

## Test

Adds `jobstreet.fetch() accepts hk.jobsdb.com (HK-Main)`, mirroring the existing untrusted-host case: asserts the request is issued against the `hk.jobsdb.com` origin and parses cleanly. Full `tests/providers/jobstreet.test.mjs` passes (16/16).

## Suite status

`node test-all.mjs` has one failure on this branch, `web/tests/lib/explore-ai-dedup.test.mjs` — **pre-existing and unrelated**. Reproduced on the unmodified base with these changes stashed (362 pass / 1 fail / 1 skipped in `web/tests/lib`). Nothing in this PR touches `web/`.

## Routing note

[...]

PR #3436, opened 2026-08-28
fix(providers): build SEEK job links per market host + opt-in appendWorkType

glenelkin:
## Summary

- `parseJobstreetItem` hard-coded the job-detail path as `{origin}/id/job/{id}`. That is the **Indonesian locale path**: on `www.seek.com.au`, `www.seek.co.nz`, `my.jobstreet.com` and `sg.jobstreet.com` it returns **404** — every SEEK AU/NZ/MY/SG result lands in `pipeline.md` with a dead link. Those hosts serve `/job/{id}` (verified against live ids on 2026-08-28; `id.jobstreet.com` still needs `/id/job/{id}`, which is why the global switch in #2210 couldn't land).
- Fix: the path is now keyed on the origin host — `/id/job/` for `id.jobstreet.com` / `jobstreet.co.id`, `/job/` for every other allow-listed host. Indonesia is byte-for-byte unchanged.
- Adds an **opt-in** `appendWorkType: true` portal-entry flag that suffixes the title with the listing's `workTypes` (`"Strategy Consultant [Part time]"`). The v5 API already returns this field; surfacing it lets `title_filter` and triage see employment type at scan time. Off by default, so no existing install changes behaviour.
- Docs: SEEK Australia/NZ configuration (`api` + `siteKey: AU-Main` / `NZ-Main`) in `SUPPORTED_JOB_BOARDS.md` and a commented example in `templates/portals.example.yml`. The provider header comment listed only ID/SG/MY site keys although the AU/NZ hosts were already allow-listed.

Compatible with #3404 (JobsDB HK): `hk.jobsdb.com` falls into the `/job/` branch, which matches JobsDB's URL scheme.

## Test plan

- [x] `node tests/providers/jobstreet.test.mjs` — 20/20 (15 existing + 5 new: AU and SG path, suffix on/off/absent, fetch-level flag)
- [x] `npm run lint`
- [x] Live: `jobstreet.fetch()` against `https://www.seek.com.au/api/jobsearch/v5/search` (`AU-Main`, "part time strategy", Melbourne VIC) returns `https://www.seek.com.au/job/<id>` links that resolve, with `[Part time]` / `[Contract/Temp]` suffixes when the flag is set
- [x] Existing Indonesian assertion (`https://id.jobstreet.com/id/job/92996157`) unchanged and passing

## Summary

Users can configure SEEK Australia and New Zealand with `siteKey: AU-Main` or `NZ-Main` in `templates/portals.example.yml:1346-1356`.

Job links now use market-specific paths: Indonesian hosts use `/id/job/{id}`. Other supported hosts use `/job/{id}` (`providers/jobstreet.mjs:63-73`).

Users can set `appendWorkType: true` to add work types to titles (`providers/jobstreet.mjs:137-147`). The default behavior does not change.

Updated files: `providers/jobstreet.mjs`, `tests/providers/jobstreet.test.mjs`, `docs/SUPPORTED_JOB_BOARDS.md`, and `templates/portals.example.yml`.

System files requested for review: `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/`. Only `providers/` is identified as changed. The supplied results show no change evidence for the other system files.


santifer:
@glenelkin: one thing I owe you before this lands. We pushed to your branch and didn't tell you.

You were told on 4 September that the conflict was ours and that we would take it if you would rather not, and then two weeks passed. Yesterday we resolved it and published the result on your branch, and CI came back green on all twelve checks including Windows. Moving someone's branch without a word is not how we want to keep a promise. Sorry.

Nothing of yours changed: the host-aware detail path and the opt-in work-type suffix are as you wrote them, Indonesia stays byte for byte identical, and your commit stays the parent, so the credit is yours.

CodeRabbit raised one thing on that commit, encoding the job id before it goes into the URL. We checked it against main: it is already there, so it is not yours and it does not block this. It goes out as a good first issue with the two existing precedents attached.

Next, and it is on us: the branch had fallen behind three merges today, so we have just updated it. CI runs once more and then we merge.

santifer (approved):
Host-aware SEEK detail paths, with Indonesia unchanged and the work-type suffix opt-in. Reviewed across earlier rounds and re-verified on this head; the conflict resolution on this branch is ours and is explained in the thread.

PR #3442, opened 2026-08-28
feat(batch): sync pipeline pending entries into batch-input.tsv

BigBrown10:
## Problem

Scan results land in `data/pipeline.md` Pending section, but batch processing reads from `batch/batch-input.tsv`. These two formats are unrelated â€” there is no built-in way to feed scan-discovered offers into the batch pipeline without manually copying URLs into a TSV. The batch runner already has `reconcile-pipeline.mjs` for the reverse direction (batch->pipeline after evaluation), but the forward path (pipeline->batch) is missing.

## Proposed solution

A `sync-batch-input.mjs` script that:

1. Reads `data/pipeline.md` Pending section (accepting EN/ES/DE/FR localized headers, same as `reconcile-pipeline.mjs`)
2. For each `- [ ]` entry, parses URL, company, role, and labeled segments (`posted:`, `trust:`, `note:`, `rank:`)
3. Writes to `batch/batch-input.tsv` in the format `url<tab>pipeline<tab>{company}<tab>{notes}`
4. Deduplicates against existing `batch-input.tsv` entries and `batch-state.tsv` (already-processed URLs)
5. Supports `--dry-run` to preview what would be added

### Format mapping

- pipeline.md URL (positional) -> batch-input.tsv `url`
- pipeline.md Company (positional, 2nd cell) -> batch-input.tsv `topic`
- pipeline.md Labeled segments (`note:`, `trust:`, `posted:`, `rank:`) -> batch-input.tsv `notes` (concatenated)
- batch-input.tsv `source` -> hardcoded `pipeline`

### Example

```
# pipeline.md
- [ ] https://jobs.ashbyhq.com/acme/791 | Acme Corp | Staff PM | note: curated shortlist
- [ ] https://boards.greenhouse.io/acme/jobs/792 | Acme Corp | Backend Engineer | trust: 80 missing_apply_url

# becomes batch-input.tsv
https://jobs.ashbyhq.com/acme/791	pipeline	Acme Corp	curated shortlist
https://boards.greenhouse.io/acme/jobs/792	pipeline	Acme Corp	trust: 80 missing_apply_url
```

## Why not just paste URLs manually?

The pipeline is the primary inbox. Scans write there, the pipeline mode processes from there, and reconcile writes back there. The batch system is a parallel evaluation path that currently requires a separate manual step to feed. Closing this gap makes batch mode a drop-in alternative to interactive pipeline processing: run a scan, sync to batch, run batch, reconcile back.

## Alternatives considered

- CLI flag on batch-runner.sh: Would work but doesn't compose with the existing pipeline workflow. A standalone script is more composable (can be called from batch-runner.sh, from npm scripts, or manually).
- Modify batch-runner.sh to read pipeline.md directly: Too invasive. Better to keep the sync as a separate pre-step.

## Testing

- Unit tests for the pipeline parser (variable-width columns, labeled segments, localized headers)
- Idempotency test (running twice produces no duplicates)
- Skip test (URLs already in batch-state.tsv are not re-added)

## Summary by CodeRabbit

- **New Features**
  - Added a tool to convert unchecked “Pending” pipeline entries into batch input records.
  - Supports duplicate detection, dry-run previews, configurable paths, and localized “Pending” headers.
  - Automatically extracts URLs, sources, topics, and notes for each record.
  - Filters unsafe or non-public HTTP(S) URLs and keeps generated files within the repository.

- **Maintenance**
  - Included the tool in system-file update and cleanup processes.

- **Tests**
  - Added coverage for parsing, extraction, formatting, localization, deduplication, validation, blank entries, and error reporting.


Scott-Emberson (commented):
Reviewed at head a92c096378. sync-batch-input.test.mjs uses concrete fixture-based assertions: it pins the exact 4-column TSV output row for a known pending entry (not just a length), asserts dedup against both batch-input.tsv and batch-state.tsv (4 pending with one queued + one state URL yields 3 rows), and excludes non-pending/processed entries (the `## Procesadas` / `- [x]` rows are not synced). Blank-line skipping and the 4-column invariant after tab/newline sanitization are checked. All nine imported symbols are real exports in the diffed module, so there are no silent undefined-import skips. This is a net-new forward-sync script and its first test.

Scott-Emberson (approved):

Scott-Emberson (commented):
Re-reviewed at head b11377c6a3 (moved since my earlier approval). Two commits, both sound and both worth having.

The row-contract fix is the important one: sync-batch-input.mjs was writing `url\tsource\ttopic\tnotes`, but the batch runner reads `id\turl\tsource\tnotes` and requires a positive integer in the id column, so every synced row was silently skipped and source landed in the wrong field. planBatchRows now allocates sequential numeric ids (a new maxTsvId helper scans column 0 for the largest existing id across batch-input.tsv and batch-state.tsv), drops the invented topic column, and reads the URL from column 1. The test pins this concretely: `rows[0].row === '6\thttps://new.example/1\tpipeline\t'` with the id sequential after the max existing, every id a positive integer, and the four-column invariant, so the runner's `^[0-9]+$` requirement is actually encoded now. The earlier dedup and non-pending-exclusion assertions are retained.

The second commit is a genuine SSRF hardening (CodeRabbit): `new URL()` normalizes `http://[::127.0.0.1]/` to a host whose empty first hextet slipped past the private-host check, so isSafePublicHttpUrl was accepting loopback and the unspecified `::` address, and parsePending would queue an internal fetch target. The fix rejects `::`/`:` and decodes IPv4-compatible IPv6 (`::h1:h2` / `::a.b.c.d`) through its low 32 bits to catch the mapped loopback/private destinations. The regression tests exercise the full isSafePublicHttpUrl and parsePending path (not the private helper directly), asserting `http://[::127.0.0.1]/`, its normalized form, `[::10.0.0.1]`, and `[::]` are all rejected and the loopback entry is dropped from queued targets, so a guard regression reddens.

Owned tests verified sound, including the SSRF regression coverage. Holding formal approval until the branch is finalized, nothing blocking from the tests/ side.

Scott-Emberson (approved):

PR #3444, opened 2026-08-28
fix(cv-sync-check): resolve prompt files from CODE_ROOT, not an undefined projectRoot

WojciechJelen:
## Problem

`node cv-sync-check.mjs` crashes on every invocation, before running a single check:

```
ReferenceError: projectRoot is not defined
    at file:///.../cv-sync-check.mjs:54:16
```

`filesToCheck` (lines 54–56) still builds its paths from `projectRoot`, which stopped existing when the script gained `CODE_ROOT` / `DATA_ROOT` for the `CAREER_OPS_ROOT` override (02daaf1c). Reproduces on current `main` (bb46c66) and on the v1.30.0 release.

`modes/_shared.md` → ALWAYS rule 1b tells the agent to run this on the first evaluation of every session, so the crash surfaces every session that honours it.

## Fix

Use `CODE_ROOT` for the three files scanned for hardcoded metrics (`modes/_shared.md`, `modes/_writing.md`, `batch/batch-prompt.md`). They're system-layer prompt files that ship with the code, so `CODE_ROOT` is correct: `DATA_ROOT` would silently skip the check for anyone with an external data directory, since those files wouldn't exist there.

3-line change, no behaviour change beyond "runs at all".

## Test

`tests/cv-sync-check-code-root.test.mjs` spawns the script against a minimal temp `CAREER_OPS_ROOT` (cv.md + profile.yml) and asserts it reaches its own `=== career-ops sync check ===` banner and exits 0, with no `ReferenceError` on stderr. Fails on `main`, passes with the fix. Auto-discovered by `test-all.mjs` via `tests/**/*.test.mjs`, no registration needed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a crash in the CV sync check that prevented it from loading successfully.
  * Ensured system prompt files are resolved from the correct code location.
  * Confirmed the sync check completes successfully and reports its results as expected.

* **Tests**
  * Added coverage to verify correct path resolution and successful execution.

### DIFF
--- a/cv-sync-check.mjs
+++ b/cv-sync-check.mjs
@@ -51,9 +51,9 @@ if (!existsSync(profilePath)) {
 
 // 3. Check for hardcoded metrics in prompt files
 const filesToCheck = [
-  { path: join(projectRoot, 'modes', '_shared.md'), name: '_shared.md' },
-  { path: join(projectRoot, 'modes', '_writing.md'), name: '_writing.md' },
-  { path: join(projectRoot, 'batch', 'batch-prompt.md'), name: 'batch-prompt.md' },
+  { path: join(CODE_ROOT, 'modes', '_shared.md'), name: '_shared.md' },
+  { path: join(CODE_ROOT, 'modes', '_writing.md'), name: '_writing.md' },
+  { path: join(CODE_ROOT, 'batch', 'batch-prompt.md'), name: 'batch-prompt.md' },
 ];
 
 // Pattern: numbers that look like hardcoded metrics (e.g., "170+ hours", "90% self-service")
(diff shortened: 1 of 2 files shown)


Scott-Emberson (commented):
Reviewed at head c3284fb4ac. cv-sync-check-code-root.test.mjs is a real regression test for the shipped defect: the buggy `join(projectRoot, ...)` runs at the top of the file, before the `=== career-ops sync check ===` header, so under the undefined-projectRoot bug the ReferenceError fires first, the header never prints, and exit is non-zero. The test asserts stderr does not match `/ReferenceError/` or `/projectRoot/`, stdout matches the header, and status is 0, all four of which redden under the bug.

One coverage note, since the commit message frames this as resolving from CODE_ROOT: the test proves "not the undefined projectRoot" firmly, but not "from CODE_ROOT specifically". The loop skips missing prompt files silently and only exits non-zero on errors (mismatches are warnings), so a hypothetical wrong-but-defined root (say DATA_ROOT) would leave prompts unresolved in a temp dir, skip quietly, and still exit 0 with the header, staying green. Not a blocker for this fix (the actual defect was the crash), just worth an assertion pinning provenance if you extend it.

Owned test verified sound for the fixed defect. Holding formal approval until the branch is finalized, nothing blocking from the tests/ side.

Scott-Emberson (approved):

Issue #3364, opened 2026-08-26
test-all is red on default Windows: git checks out the 7 skill-entrypoint symlinks as text, and no CI runner can reproduce it


artemtrofymenko:
## What happens

`node test-all.mjs` on Windows 11, non-elevated shell, Developer Mode off — the default configuration — on a clean checkout of current `main`:

```
❌ .antigravitycli\skills\career-ops\SKILL.md: missing cwd-independent routing rule |
   .claude\skills\career-ops\SKILL.md: missing cwd-independent routing rule |
   .cursor\skills\career-ops\SKILL.md: missing cwd-independent routing rule |
   .grok\... | .kimi\... | .opencode\... | .qwen\...
```

Seven of the eight entrypoints `tests/skill-project-root.test.mjs` checks, and the suite is red for every contributor on that configuration.

## Why the files look empty of the rule

They are not files. Seven of them are symlinks, and they are the only seven in the repository:

```
git ls-files -s -- '.*/skills/career-ops/SKILL.md'
  100644  .agents/skills/career-ops/SKILL.md      <- the real document
  120000  .claude/skills/career-ops/SKILL.md      <- symlink
  120000  .cursor/  .grok/  .kimi/  .opencode/  .qwen/  .antigravitycli/
```

Creating a symlink on Windows needs `SeCreateSymbolicLinkPrivilege`, so git records its own capability at clone time:

```
git config --show-origin core.symlinks
  file:.../.git/config    false
```

With that set, git materializes each symlink as an **ordinary text file containing the link target**:

```
head -c 70 .claude/skills/career-ops/SKILL.md
  ../../../.agents/skills/career-ops/SKILL.md
```

So `readFileSync` returns 43 bytes of path, the test looks for `Resolve every path in this router` inside it, and fails. On Linux and macOS the same call follows a real symlink and returns the document, which is why all three CI jobs are green.

## Why CI cannot catch this

#3355 introduced both the assertion and the rule text, touching exactly two files — `.agents/skills/career-ops/SKILL.md` and the new test — and merged with `test (ubuntu-latest)`, `test (macos-latest)` and `test (windows-latest)` all SUCCESS. The Windows runner passes because it can create symlinks; an ordinary Windows box cannot, and `core.symlinks` is decided per clone rather than per platform. No CI configuration reproduces it.

The other seven mirrors were last touched in #572 (May) and #2115 (July) and never contained the rule, so the assertion has been false since the moment it was written — just not anywhere it could be seen.

[...]

PR #3445, opened 2026-08-28
test(harness): a suite at the repo root is a suite nothing runs

artemtrofymenko:
The stray-root successor @vliggio said yes to on #3388, closing the last thread of #3306.

## What changed about the question

`tests/root-suite-registration.test.mjs` asked *"is every root-level suite registered in `test-all.mjs`?"* — one name at a time, because there was a list to guard. #3388 deleted the list, so that guard went with it: a name-based check over a list that no longer exists reads as protection while protecting nothing.

The property that survives the move is a **location**, not a registration. So this asks *"is there a root suite at all?"* — nothing to keep in sync, nothing to forget, and it encodes what ARCHITECTURE.md now states rather than the set of files that currently satisfy it.

```js
const strays = readdirSync(ROOT, { withFileTypes: true })
  .filter((e) => e.isFile() && e.name.endsWith('.test.mjs'))
```

## Verified both ways

Copying #3303's orphan back to the root — the file that sat there for weeks with 20 assertions, passing every one of them, while nothing ran it:

```
❌ 1 suite(s) at the repo root, where discovery does not reach and nothing runs them:
     jd-similarity.test.mjs
   Move the file into tests/ — discovery picks it up with no registration.
```

## Why there is a second assertion

The real check only ever reports an **absence**, and a wrong `ROOT` produces exactly that reading while measuring nothing — a silent pass, which is the same shape as the bug the file exists to prevent. `test-all.mjs` is the cheapest sentinel: it is the harness itself and cannot move without the premise moving with it. Hiding it separates the two outcomes:

```
❌ ROOT does not contain test-all.mjs (...) — this guard is looking in the wrong
   place and would pass on any tree
✅ no test suite sits at the repo root — tests/ is the only home
```

That second line is the vacuous pass the first one catches.

## The pattern is deliberately narrow

`*.test.mjs`, not "anything test-shaped" — @vliggio's warning on #3388, and it is load-bearing: `test-salary-filter.mjs` and `test-trust-validator.mjs` are at the root and correctly registered, so a looser match would redden on two files that are fine. #3411 moves them into `tests/`, after which this reads the same either way, which is the point of keying on the discovery pattern rather than on a naming convention.

## Verification

```
node test-all.mjs        6565 passed, 1 failed, 15 warnings
```

That failure is #3364 — the seven skill-entrypoint symlinks that a default Windows clone checks out as pointer text — and it reproduces on unmodified `main`. Unrelated to this branch; assigned to @gowrishacv with me verifying on the box.

Run on Windows 11, non-elevated, Developer Mode off.

## Summary by CodeRabbit

* **Tests**
  * Added checks to ensure test suites are stored in the supported test directory.
  * Added clear failure guidance when a test suite is placed at the repository root.

### DIFF
--- a/tests/no-root-suites.test.mjs
+++ b/tests/no-root-suites.test.mjs
@@ -1,0 +1,58 @@
+// tests/no-root-suites.test.mjs — no test suite may sit at the repo root,
+// because nothing there runs it.
+//
+// test-all.mjs discovers tests/**/*.test.mjs (#1440) and the comment on that
+// function is explicit that discovery stops there: root-level standalone
+// *.test.mjs files are never picked up. Until #3306 the nine that lived at the
+// root were named one by one in a `scripts` list, and a list is a thing you can
+// forget — jd-similarity.test.mjs was added with 20 assertions, appeared in no
+// runner at all, and passed the whole time it was not running (#3303).
+//
+// #3388 moved all nine into tests/ and deleted that list, which retired the
+// guard over it: a name-based check over a list that no longer exists reads as
+// protection while protecting nothing. This is the successor, and it is a
+// different kind of check. The old one asked "is every root suite registered?"
+// — procedural, one entry per file, drifting the moment someone forgets. This
+// one asks "is there a root suite at all?" — a location, with nothing to keep
+// in sync, and it encodes the doctrine ARCHITECTURE.md now states rather than a
+// list of the files that happen to satisfy it.
+//
+// `*.test.mjs` specifically, NOT "anything test-shaped". test-salary-filter.mjs
+// and test-trust-validator.mjs sit at the root and are correctly registered in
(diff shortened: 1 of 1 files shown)


Scott-Emberson (commented):
Reviewed at head e0fd1e2e50. This is a real guard, not a hardcoded list: it readdir's the repo root and fails on any `*.test.mjs` found on disk, and the premise gate (`statSync('test-all.mjs')` must be a file, else fail) kills the vacuous-pass path a mis-resolved root would otherwise give. It's auto-discovered by test-all.mjs and its `fail()` reddens the aggregate suite, so a stray root suite genuinely turns CI red. Matching the discovery pattern (`*.test.mjs`) is the right scope, so an intentional `test-foo.mjs` at root is correctly allowed.

Scott-Emberson (approved):

PR #3446, opened 2026-08-28
fix(scan-ats-full): collapse a Workday requisition served under sever…

### LINKED ISSUE #3439: One Workday requisition served under several tenant sites is queued two or three times
## The problem

A Workday tenant can publish the same requisition on more than one site — its own careers page plus the Indeed or Glassdoor feeds it powers. The URLs differ, so URL-based dedup in `scan-ats-full.mjs` treats them as separate openings, and the same job lands in the queue two or three times.

The requisition id is right there in every one of those URLs, and it is the thing that actually identifies the opening. Normalising the URL cannot reach it, because the difference is the host and the path shape, not the query string.

## Proposed change

An optional `dedupKey(job)` on the provider contract, returning a tenant-scoped key when the provider can derive one. `scan-ats-full.mjs` prefers it and falls back to the normalised URL when a provider does not define the method, or defines it but cannot derive a key from a given URL.

Opt-in per provider and backward compatible: providers without the method behave exactly as they do now. Workday would be the first implementation; the same shape fits any ATS that exposes a stable requisition id.

Tenant scoping matters — two different tenants can legitimately reuse the same requisition number, so the key has to carry both.

## Status

Implemented locally with tests covering collapsing across sites, tenant scoping, multi-underscore requisition ids, and both null paths. This touches the provider contract, so raising it as an issue first per CONTRIBUTING rather than opening the PR — happy to send it if the direction fits.

Bekon0700:
…al sites

Add an optional dedupKey(job) to the provider contract, and prefer it over normalizeUrlForDedup(job.url) at every dedup site in scan-ats-full.mjs (main sweep, checkpoint-resume reseed, seed pass). workday.mjs implements it: same tenant+instance host, requisition ID taken from everything after the first underscore in the URL's last path segment (Workday title slugs are hyphen-only, so this holds even when the requisition ID itself contains underscores).

The checkpoint-resume reseed loop previously reseeded seenUrls by URL only, so a resumed run could let a cross-site duplicate back in even though the pre-checkpoint sweep had already collapsed it via dedupKey — fixed by routing it through the same dedupTokenFor() helper as the main sweep.

Fixes #3439

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [ ] I have read [CONTRIBUTING.md](https://github.com/santifer/career-ops/blob/main/CONTRIBUTING.md)
- [ ] My PR does not include personal data (CV, email, real names, scan results, or pipeline data)
- [ ] My changes respect the [Data Contract](https://github.com/santifer/career-ops/blob/main/DATA_CONTRACT.md) (no modifications to user-layer files)

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved duplicate detection for Workday postings shared across multiple site URLs.
  - Preserved separation between different Workday tenants and requisitions.
  - Maintained URL-based detection for other providers and unsupported cases.
  - Improved deduplication when resuming scans from previous history.
  - Added safer handling for malformed or incomplete job URLs.

- **Tests**
  - Added coverage for cross-site deduplication, requisition formats, malformed URLs, and fallback behavior.


Scott-Emberson (commented):
Reviewed at head 0220228d62. scan-ats-full-dedup-key.test.mjs meets the bar: it first independently reproduces the bug (asserts the two site URLs normalize differently, so the fixture genuinely needs the fix), then asserts the same requisition served under two site paths collapses to one dedup token (if the collapse regressed, dedupKey returns null, falls back to URL norm, and the tokens diverge, which fails). The over-collapse guard is there too (same req string under a different tenant stays distinct), workday.test.mjs pins the exact keys, and a legacy scan-history row dedupes a fresh site-B offer so the cross-session/resume gap is covered, not just intra-run.

Scott-Emberson (approved):

providers/workday.mjs
```diff
@@ -128,6 +128,50 @@ function locationFromPath(externalPath) {
...
+  // bogus workday: key just because their last path segment happens to
+  // contain an underscore (e.g. a Lever/Greenhouse job whose slug does) —
+  // reported by CodeRabbit against this exact function.
+  if (!parsed.hostname.toLowerCase().endsWith('.myworkdayjobs.com')) return null;
+  const segments = parsed.pathname.split('/').filter(Boolean);
+  const lastSegment = segments[segments.length - 1];
+  if (!lastSegment) return null;
+  const underscoreIdx = lastSegment.indexOf('_');
+  if (underscoreIdx === -1) return null; // no title/requisition-ID separator — nothing to key on
+  const reqId = lastSegment.slice(underscoreIdx + 1).toLowerCase().replace(/-\d+$/, '');
```
ronanime-arch: Cross-posting from #3439 (issue-comment-5470960651) since this PR is approved and mergeable — this line is the one edge neither test set covers, and `-N` **is** a real part of the requisition ID on a number of tenants.

Checked `/-\d+$/` against the Workday IDs in my own scan history and ATS cache:

| ID (real, from my cache) | key after strip | |
|---|---|---|
| `R-2593225`, `R-2592964` (Walmart) | `…:r` | every posting on the tenant → one key |
| `R-058589` (Red Hat), `R-101976` (XPEL), `R-4253` (Lytx), `R-103502` (Job Duck) | `…:r` | same |
| `JR26-39350`, `JR26-42996` | `…:jr26` | two distinct reqs collapse |
| `R2026-00707`, `-01237`, `-01334`, `-01355` | `…:r2026` | four collapse |
| `R11312`, `R11312-2`, `R11312-3` | `…:r11312` | strip is correct here |
| `R26007842-1`, `R260022205-2` | base | correct here too |

The `R-NNNNNNN` family is the damaging one: the hyphen-digits *are* the whole ID, so the strip leaves a bare `r`, `reqId` is non-empty, and the scanner silently keeps one job per tenant. That is a bigger loss than the duplicate this PR fixes — Walmart alone is a few hundred postings per sweep in my data.

The split in the data is clean: a disambiguator is 1–2 digits hanging off an already-long base; an ID tail is 5+ digits and often zero-padded. Gating on both ends:

```js
const raw = lastSegment.slice(underscoreIdx + 1).toLowerCase();
// A trailing -N is a site disambiguator only when it is short AND the base is
// still a requisition ID on its own — otherwise the hyphen-digits are the ID
// itself (R-2593225, JR26-39350, R2026-01237) and stripping empties it.
const m = raw.match(/^(.*?)-(\d{1,2})$/);
const reqId = m && /^[a-z]*\d{3,}$/.test(m[1]) ? m[1] : raw;
```

Ran it over every shape in my cache: `R11312-2/-3`, `R260022205-2`, `R26007842-1`, `R266069-1`, `R53113-2`, `JR1126610-1`, `R2000678390-1` still strip to their base; `R-2593225`, `JR-8327`, `JR26-39350`, `R2026-00707` stay intact. The `R26_05710` case is unaffected — that one splits on the underscore, not the hyphen.

Suggest two extra cases in `tests/providers/workday.test.mjs`: one `R-NNNNNNN` tenant (two different reqs must **not** collapse) and one `JR26-NNNNN`.

Scott-Emberson (commented):
Re-reviewed at head 135c1a5f59 (moved past the earlier approval, so this wants a fresh approval). The new commit is a real behavior change (don't strip hyphenated requisition IDs in dedupKey), and the added coverage has teeth in both directions. The mustNotStrip fixtures (R-058589, R-2593225, R2026-01355) assert the dedup key keeps the hyphen and that R-2593225 versus R-2592964 stay distinct, so an over-strip that wrongly merges distinct reqs reddens. The mustCollapse fixtures (R11312 with -2 / -3, R26007842-1, JR1126610-1) assert real "-N" disambiguators still collapse to one key, so a lazy "disable all stripping" fix also reddens. The provider guard (strip a -N tail only when the base is itself req-ID-shaped) matches all fixtures, and the earlier cross-site JR00123 collapse and tenant-scoping assertions are untouched (the patch is additive).

Owned test verified sound, bidirectional coverage. Holding formal approval until the branch is finalized (fresh approval needed since the head moved), nothing blocking from the tests/ side.

Scott-Emberson (approved):

PR #3456, opened 2026-08-28
test: handle Windows materialized symlink entrypoints

### LINKED ISSUE #3364: test-all is red on default Windows: git checks out the 7 skill-entrypoint symlinks as text, and no CI runner can reproduce it

gowrishacv:
## What does this PR do?

This PR fixes the skill project-root test for Windows checkouts where Git materializes symlinks as pointer-text files when `core.symlinks=false`.

The test now checks Git index mode `120000`, resolves the pointer target relative to the entrypoint, and verifies the real `SKILL.md`. A cross-platform fixture was added so CI covers this behavior without requiring Windows symlink privileges.

## Related issue

Fixes #3364

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [ ] I have read [CONTRIBUTING.md](https://github.com/santifer/career-ops/blob/main/CONTRIBUTING.md)
- [x] Bug fix; issue-first requirement is not applicable

## Tests

- `node test-all.mjs --only skill-project-root`
- `node scripts/check-syntax.mjs`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved project-root detection for skill entrypoints represented as Git symlinks.
  * Ensured symlink-based entrypoints correctly resolve to their target paths.
  * Improved validation of routing rules associated with resolved skill entrypoints.
* **Tests**
  * Added coverage for materialized symlink pointers in temporary repositories.
  * Strengthened validation and cleanup during test execution to improve reliability.


Scott-Emberson (commented):
Reviewed at head 7012fbb587. The Windows materialized-symlink handling keeps both original assertions live against the resolved target: project-root resolution (`findProjectRoot(entrypoint.path) === ROOT`) and the routing-rule content check both still fire. There's no platform branch that skips the assertion, detection rides on the git index mode 120000 which is platform-agnostic, and every mismatch path (multiline content, pointer outside root, wrong mode) returns the original pointer text, so a genuine "rule removed from target" regression fails rather than being masked. The new fixture positively exercises the resolve path.

Scott-Emberson (approved):

PR #3466, opened 2026-08-28
fix(merge-tracker): two '?' rows with no Via silently merge, overwriting one (#3410)

### LINKED ISSUE #3410: merge-tracker.mjs: two '?' (confidential-employer) rows with no via silently merge instead of staying distinct
## Summary

`merge-tracker.mjs`'s Tier-2 (company + role fuzzy match) dedup has a cross-channel guard specifically to stop two different agency-mediated `?` (confidential end-employer) rows with a similarly-worded role title from being merged into one row — the code comment at the guard site describes this exact hazard. The guard only fires when the two rows' `via` values differ. When **both** rows have an empty/unset `via` (e.g. a `?` row added before Via tracking existed, or a TSV author who — as I did — puts the `via=` tag inside the notes column instead of as its own trailing field, so it's silently dropped by the parser), `normalizeVia('') === normalizeVia('')` is `true`, the guard does not fire, and the two unrelated rows are treated as a duplicate and merged. This silently overwrites one row's company/role/score/report-link/notes with the other's.

## What I hit

Two genuinely unrelated evaluations both landed as company `?` with role fuzzy-matching to "Program Manager":
- An existing row (a fictional "Acme Corp" role sourced via a fictional staffing agency "Blue Oak Staffing", an internal grad/rotational program eval)
- A new row (an undisclosed application-security client sourced via a fictional contingent-workforce platform "Contractly")

Running `node merge-tracker.mjs` to add the second one instead overwrote the first row's data (score, report link, notes) — no warning, no rejection, just a silent `🔄 Update`. A comment already left on that tracker row shows this exact failure mode happened once before as well (a different pair of unrelated `?` rows, a few weeks earlier), so this isn't a one-off.

In my case the immediate trigger was my own formatting mistake (I put `via=Contractly` inside the notes text instead of as its own tab-delimited trailing field, so the parser's `parts.slice(9)` never saw it as a `via=` tag and `addition.via` came back empty) — so this issue isn't "the guard is broken," it's "the guard's failure mode when via is unavailable on either side is silent merge, when arguably it should be silent *rejection*." A `?` company row with no known via carries no positive evidence that it's the same posting as another `?` row with no known via — the current behavior treats "unknown vs. unknown" the same as "confirmed same," when it's really closer to "confirmed different" (two independent draws from the empty-via bucket).

## Suggested fix

In the Tier-2 cross-channel guard (`merge-tracker.mjs`, around the `companiesMatch(app.company, addition.company)` / `roleFuzzyMatch` block for the `'?'` company case): when *either* side's `via` is empty/unset, don't fall through to a match — treat it the same as a confirmed via mismatch (return `false` / not-a-duplicate) rather than treating two empty vias as equal. This would make an ambiguous case fail safe (append as a new row, possibly with a console warning like "two `?` rows with no via — treating as distinct, verify manually") instead of failing by silently clobbering existing data.

Happy to be more specific about the exact line numbers if useful — I found the guard site while investigating today (search for the `#1596` cross-channel comment in `merge-tracker.mjs`).

## Why this matters

The tracker (`data/applications.md`) is gitignored with no `.bak` on this code path (per an existing comment in the file, `#2803`), so a silent overwrite here is not recoverable from version control — the only way I caught it was manually cross-checking the `🔄 Update` line against the report on disk, which the merge output doesn't prompt anyone to do.

abankar1:
Fixes #3410.

`?` marks an undisclosed end employer, so every such row normalizes to the **same empty company key**, and the Via column is what tells two of them apart. The tier-2 cross-channel guard exists for that (#1596) — but it rejected only when the two vias *differed*, so two empty ones compared equal and fell through to the fuzzy title match.

## What it costs

Existing `? / Program Manager` (4.2, Applied, report [1]) + an unrelated `? / Senior Program Manager` addition, on `main`:

```
BEFORE: | 1 | 2026-01-05 | ? | Program Manager | 4.2/5 | Applied | ✅ | [1](../reports/001-a-…) | 2K Games via Insight Global |
AFTER : | 1 | 2026-02-10 | ? | Program Manager | 3.6/5 | Applied | ❌ | [2](../reports/002-b-…) | 2K Games via Insight Global. Re-eval … Superseded report [1] (was 4.2/5): different agency, different employer |
rows: 1
```

One row where there were two. The survivor keeps the first application's `Applied` status while its date, score, PDF flag and **report link** now describe the other posting — and the note calls report [1] "Superseded", which it is not. The tracker is gitignored and no `.bak` is written, so nothing recovers it.

@jaybee2184's report used a fuzzy title pair; note the seniority-prefixed variant merges too, so the surviving row's *role title* can be rewritten as well.

## Fix

Both vias must be **present and equal**, not merely equal. This is the same three-valued logic Pass 0 states twenty lines above — an absent key is UNKNOWN, never "equal" — landing the other way here on purpose: there, company and role still carry identity, so an unknown URL must not block a tier that has other evidence; here the company key is empty *by construction* and a fuzzy role match is not identity. A visible duplicate row is recoverable; a silent overwrite of a row whose report link now points at a different job is not.

**Gated on the tracker actually having a Via column.** Without one, every row parses with `via=''` and the addition's own tag is cleared deliberately, so empty-vs-empty is the *normal* state for a genuine same-agency re-blast rather than a missing signal. `tracker-columns-tests.mjs` §12b asserts exactly that case and **caught the ungated first version of this fix** — good catch by whoever wrote it. On that layout the pre-existing behaviour stands, and `--migrate-via` is the way to get the guard.

## What does not move

| case | before | after |
|---|---|---|
| neither has Via (**the bug**) | merged | **stays distinct** |
| same agency both sides | merged | merged |
| different agencies | distinct | distinct |
| re-eval carrying the same report number | merged | merged |
| legacy tracker, no Via column | merged | merged |

The re-eval row is what keeps this from costing anything real: tier 1 matches on the report number — provable identity — and runs *before* this guard, so re-running an evaluation of the same confidential posting still updates in place, Via or no Via.

## Verification

Two of the six new tests fail on `main`:

```
✖ two `?` rows with no Via stay distinct instead of merging
✖ an empty Via is unknown, not a matching value
ℹ pass 4  ℹ fail 2
```

The other four are the cost side, which is where the risk actually was. And the first test asserts the surviving row is **intact**, not merely present — a merge that kept two rows while overwriting one would satisfy a row count and none of the point, so it checks the original score, report link and date survive and that the untouched report is not marked "Superseded".

`node test-all.mjs --quick` → green (including `tracker-columns-tests.mjs`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved duplicate detection for confidential applications using unknown company markers.
  * Prevented unrelated roles with missing or different agency information from being merged.
  * Preserved correct merging when agency details match.
  * Maintained existing behavior for legacy trackers without agency columns.

* **Tests**
  * Added coverage for missing, matching, and differing agency values, including same-report updates.

### DIFF
--- a/merge-tracker.mjs
+++ b/merge-tracker.mjs
@@ -1148,7 +1148,44 @@ for (const file of tsvFiles) {
       // the same channel (the agency re-blasting one listing) is a duplicate.
       // Via comparison is Unicode-aware (#1603): normalizeCompany() would
       // collapse distinct non-Latin agency names to the same empty key.
-      if ((String(addition.company).trim() === '?' || String(app.company).trim() === '?')
+      //
+      // ON A TRACKER THAT HAS A VIA COLUMN, both vias must be PRESENT and equal,
+      // not merely equal (#3410). Rejecting only on a difference let two empty
+      // ones compare equal, so two unrelated `?` rows with no Via fell through
+      // to the fuzzy title match and merged. Measured: an existing
+      // `? / Program Manager` (4.2, Applied, report [1]) and an unrelated
+      // `? / Senior Program Manager` became ONE row carrying the addition's
+      // date, score, PDF flag and report link, with the surviving note calling
+      // report [1] "Superseded" — which it was not. The tracker is gitignored
+      // and no .bak is written, so nothing recovers it.
+      //
+      // This is the same three-valued logic Pass 0 states above — an absent key
+      // is UNKNOWN, never "equal" — landing the other way here, on purpose.
+      // There, company and role still carry identity, so an unknown URL must not
+      // block a tier that has other evidence. Here the company key is empty BY
+      // CONSTRUCTION: `?` means "employer not disclosed", so a fuzzy role match
+      // is all that is left, and it is not identity. With nothing to tell the
+      // channels apart, not merging is the recoverable answer — a visible
+      // duplicate row the user can fix, rather than a silent overwrite of a row
+      // whose report link now points at a different job.
+      //
+      // GATED ON THE COLUMN EXISTING, because without it every row parses with
+      // via='' and the addition's own tag is cleared on purpose (see the
+      // --migrate-via warning above): empty-vs-empty is then the NORMAL state
+      // for a genuine same-agency re-blast, not a missing signal, and requiring
+      // a value would turn every legacy re-blast into a duplicate row. On that
+      // layout there is nothing to compare, so the pre-existing behaviour
+      // stands and `--migrate-via` is the way to get the guard.
+      if (COLMAP.via != null
+          && (String(addition.company).trim() === '?' || String(app.company).trim() === '?')) {
+        const additionVia = normalizeVia(addition.via || '');
+        const appVia = normalizeVia(app.via || '');
+        if (!additionVia || !appVia || additionVia !== appVia) return false;
+      }
+      // Legacy layout (no Via column): the original difference-only guard, which
+      // is all the information available there.
+      if (COLMAP.via == null
+          && (String(addition.company).trim() === '?' || String(app.company).trim() === '?')
           && normalizeVia(addition.via || '') !== normalizeVia(app.via || '')) return false;
       // Req/job-number guard (#1524): a similarly-worded title at the same
       // company can still be a genuinely distinct posting when a req/job
--- a/tests/merge-tracker-confidential-via.test.mjs
+++ b/tests/merge-tracker-confidential-via.test.mjs
@@ -1,0 +1,167 @@
+// tests/merge-tracker-confidential-via.test.mjs — two undisclosed employers are
+// not one employer (#3410).
+//
+// `?` is the marker for "end employer not disclosed", so every such row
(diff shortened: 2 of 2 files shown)


Scott-Emberson (commented):
Reviewed at head 6617c93913. merge-tracker-confidential-via.test.mjs is a genuine regression test for #3410: two `?`-company rows with empty Via and different roles must stay 2 rows (on main, empty-vs-empty compares equal, falls to fuzzy title match, and merges to 1, which reddens), and beyond the count it asserts row 1 keeps its score, report link, and date and does not gain a "Superseded" note, killing the "kept 2 rows but overwrote one" false-pass. The other direction is covered too: same-agency rows still merge to 1, two different agencies stay 2, and the report-number tier-1 re-eval still updates in place. The new `COLMAP.via != null` branch requires both vias present and equal, and the test drives the real merge path (symbols confirmed at head).

Scott-Emberson (approved):

Issue #918, opened 2026-06-11
RFC: SQLite storage layer for the tracker — applications.md becomes a rendered view


Schlaflied:
> **Status (2026-06-11):** RFC [endorsed](https://github.com/santifer/career-ops/issues/918#issuecomment-4678262108) with the rollout order inverted — the derived index ships first, the source-of-truth flip becomes an explicit per-user opt-in. This text reflects the agreed order. Phase 1 is implemented in #919.

## Roadmap

- [x] **Phase 1 — DB as derived index** → #919 (in review). `applications.md` stays the source of truth; the SQLite DB is rebuilt from it; tools read through the index. Users notice nothing except fewer bugs.
- [ ] **Phase 2 — opt-in flip.** A user runs an explicit migrate command; for them the DB becomes source of truth and `applications.md` becomes the rendered view. Gated on:
  - [ ] Phase 1 merged, released, and lived with in the wild
  - [ ] Round-trip tests (md → db → md lossless) holding in CI — shipped early in #919
  - [ ] `migrate` ships with its inverse (export back to markdown, documented rollback)
  - [ ] Never the default — opt-in only, reversible, documented
  - [ ] Point modes/docs at `tracker.mjs query`; teach `merge-tracker.mjs` to write through the DB

## Problem

`data/applications.md` is a markdown table acting as a database. At small scale it works; at real scale it degrades in ways that are structural, not user error. Evidence from my own tracker (934 entries, accumulated over ~2 months of heavy use):

`node verify-pipeline.mjs` reports **422 errors**, all of them artifacts of many-writers-plus-no-schema:

- **Encoding corruption** — dozens of rows where an em-dash became `鈥?` (UTF-8 read as GBK at some write point); once it is in, every later read propagates it
- **Column drift** — 80+ rows with a score (`3.5/5`) sitting in the *status* column; markdown tables silently accept any cell content
- **Delimiter fragility** — any `|` inside a company name or note shifts every column after it
- **No referential integrity** — report links point at files that do not exist

None of these are *possible* in a real store: columns have types, statuses can be constrained to `templates/states.yml`, links can be validated on write.

## The deeper problem: data correctness currently depends on model intelligence

[...]

PR #3469, opened 2026-08-28
fix(gitignore): the tracker's derived SQLite index is unignored outside data/ (#3468)

### LINKED ISSUE #3468: The tracker's derived SQLite index is not gitignored outside data/, and every test run leaves one in the repo root
## Problem

`applications.db` is `tracker.mjs`'s SQLite index of `data/applications.md` (#918) — the whole tracker: company, role, score, status, notes for every application. It is **not gitignored** when it lands in the repo root:

```
$ git check-ignore -q <path> && echo IGNORED || echo NOT IGNORED

data/applications.db     IGNORED          # covered by data/*
applications.db          NOT IGNORED      ←
applications.db-wal      NOT IGNORED      ←
applications.db-shm      NOT IGNORED      ←
```

## When it lands in the root

The index sits beside the tracker markdown, and `resolveTrackerPath()` falls back to `<root>/applications.md` when `data/applications.md` is absent:

```js
const raw = env ? env
  : existsSync(join(rootDir, 'data/applications.md'))
    ? join(rootDir, 'data/applications.md')
    : join(rootDir, 'applications.md');
```

Two places that happens:

1. **The legacy install layout** — a tracker in the repo root is still a supported read path, and the index follows it there.
2. **Every fresh clone.** `node test-all.mjs` leaves a 36 KB `applications.db` in the root, reproducibly:

```
$ rm -f applications.db && node test-all.mjs --quick >/dev/null 2>&1 && ls -la applications.db
-rw-r--r--  36864  applications.db
$ node -e "…sqlite_master…"
tables: applications, status_events, sqlite_sequence, meta
```

Empty in this repo, since no tracker ships — but `test-all.mjs`'s own `--dry-run` comment already flags the case that matters: *"risky for end users who run tests inside their active career-ops workspace."* On that machine the file is populated.

## Scope, stated honestly

For an end user on the standard `data/` layout this never fires — the existing rules cover it. What it actually costs is (a) contributors, who get an untracked 36 KB binary after every test run that `git add .` will stage, and (b) legacy-layout installs, where it is the user's real tracker. I would not call this an active leak for most users; it is the same *class* as the timestamped backups the `*.bak*` rule was added for, reached through a layout the rules never covered.

## Fix

`*.db` (plus the `-wal`/`-shm` sidecars) beside the existing `*.bak*` rule, whose comment already states this reasoning for backups: a file holding the same PII as the original, under a name the other rules do not match. No tracked file matches — `git ls-files | grep '\.db$'` is empty.

PR follows, with probes in `tests/user-layer-gitignored.test.mjs` in the same block shape as the backup probes. Removing the rule reports all three root-layout shapes as not-ignored while `data/applications.db` stays ignored either way, which is exactly the gap.

abankar1:
Fixes #3468.

`applications.db` is `tracker.mjs`'s SQLite index of `data/applications.md` (#918) — company, role, score, status and notes for every application. It's not gitignored when it lands in the repo root:

```
data/applications.db     IGNORED          # covered by data/*
applications.db          NOT IGNORED      ←
applications.db-wal      NOT IGNORED      ←
applications.db-shm      NOT IGNORED      ←
```

The index sits beside the tracker markdown, and `resolveTrackerPath()` falls back to `<root>/applications.md` when `data/applications.md` is absent — so it lands in the root on the **legacy install layout**, and on **every fresh clone**:

```
$ rm -f applications.db && node test-all.mjs --quick >/dev/null 2>&1 && ls -la applications.db
-rw-r--r--  36864  applications.db
tables: applications, status_events, sqlite_sequence, meta
```

## Change

`*.db` and its `-wal`/`-shm` sidecars, beside the existing `*.bak*` rule whose comment already states this reasoning for the timestamped backups: a file holding the same PII as the original, under a name the other rules don't match. No tracked file matches (`git ls-files | grep '\.db$'` → empty).

Probes added to `tests/user-layer-gitignored.test.mjs` in the same block shape as the backup probes. Removing the rule:

```
❌ applications.db is NOT git-ignored — the derived index holds the same PII as the tracker
❌ applications.db-wal is NOT git-ignored — …
❌ applications.db-shm is NOT git-ignored — …
```

while `data/applications.db` stays ignored either way — which is exactly the gap: the existing rules only ever covered the standard layout.

## Scope, stated honestly

For an end user on the standard `data/` layout this never fires. What it actually costs is contributors — an untracked 36 KB binary after every test run that `git add .` will stage — and legacy-layout installs, where it is the user's real tracker. I wouldn't call it an active leak for most users; it's the same *class* as the backups, reached through a layout the rules never covered. Cheap to close either way.

`node test-all.mjs --quick` → green, and the `applications.db` the run itself leaves is now ignored.

## Note on ordering

This touches `.gitignore` a few lines from #3246's `*.tmp*` rule (still open). Different lines, so they should merge in either order — happy to rebase whichever lands second.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented derived tracker database and temporary sidecar files from appearing as untracked repository changes.

* **Tests**
  * Added coverage verifying that tracker index files in supported locations are correctly ignored by version control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,6 +104,21 @@ reports/*-RESERVED.md
 # original sitting untracked in the repo root. Trailing `*` covers both.
 *.bak*
 
+# The tracker's derived SQLite index (tracker.mjs, #918). It sits beside the
+# tracker markdown, so on the standard layout it lands in data/ and is already
+# covered by the rules above — but resolveTrackerPath() falls back to
+# `<root>/applications.md` when data/applications.md is absent, and the index
+# follows it there. That is the legacy install layout, and it is also every
+# fresh clone: `node test-all.mjs` leaves a 36KB applications.db in the repo
+# root, which `git add .` then stages.
+#
+# Its content is the whole tracker — company, role, score, status, notes — so it
+# carries the same PII as the file it indexes, under a name none of the rules
+# above match. SQLite's -wal/-shm sidecars travel with it.
+*.db
+*.db-wal
+*.db-shm
+
 # OpenRouter runner — generated data (personal, never commit)
 data/model-blacklist.json
 
--- a/.gitignore
+++ b/.gitignore
@@ -115,6 +115,15 @@ reports/*-RESERVED.md
 # Its content is the whole tracker — company, role, score, status, notes — so it
 # carries the same PII as the file it indexes, under a name none of the rules
 # above match. SQLite's -wal/-shm sidecars travel with it.
+#
+# Deliberately not root-anchored. The index follows the tracker markdown, and
+# getCareerOpsRoot() resolves a relative CAREER_OPS_ROOT (or .career-ops-data
+# marker) against the codebase directory — so a data root configured as
+# `career-data` puts both inside the checkout, in neither of the two places an
+# anchored rule would cover. Nothing tracked in this repo ends in .db, and an
+# ignore rule cannot untrack a file that is already tracked, so the cost of the
+# wider glob is a `git add -f` on a .db fixture nobody has added yet; the cost
+# of the narrower one is this leak, for every non-default data root.
 *.db
 *.db-wal
 *.db-shm
--- a/tests/user-layer-gitignored.test.mjs
+++ b/tests/user-layer-gitignored.test.mjs
@@ -151,14 +151,21 @@ for (const path of timestampedBackupProbes) {
 // in a subdirectory of the checkout. `data/applications.db` cannot settle this
 // on its own: it is already covered by the blanket `data/*` rule at the top of
 // the file, so it passes either way.
-const derivedIndexProbes = [
-  'applications.db',
-  'applications.db-wal',
-  'applications.db-shm',
-  'data/applications.db',
-  'career-data/applications.db',
-  'career-data/applications.db-wal',
+//
+// Built as a cross-product rather than hand-listed. tracker.mjs derives every
+// one of these from a single path — DB_PATH is resolveTrackerPath() with .md
+// swapped for .db, and SQLite appends the sidecar suffixes to that — so any
+// directory that can hold the index can hold all three names. Enumerating them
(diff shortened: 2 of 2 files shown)


Scott-Emberson (commented):
Reviewed at head 1772d6151e. The added derived-index case has teeth: it resolves each probe through a real `git check-ignore` verdict (exit-code based, not a string match), and the probe paths match where the index actually lives (`DB_PATH` = `resolveTrackerPath()` minus `.md` plus `.db`, with `-wal`/`-shm` sidecars). The root probe `applications.db` and the `career-data/applications.db` subdir probe are covered only by the new un-anchored `*.db` rule, so deleting that rule reddens both. The `data/applications.db` probe is redundant with `data/*` (as the comment notes), but the out-of-data/ probes are the load-bearing ones.

Scott-Emberson (approved):

Scott-Emberson (commented):
Re-reviewed at head 0a22823939. Heads-up: this PR was already approved at the earlier head, and a new commit has moved it, so that approval is now stale against the old commit and it will want a fresh approval. The change is test-only and a genuine strengthening: CodeRabbit spotted that the hand-written probe list was asymmetric (it covered .db/-wal/-shm in the repo root but only .db/-wal one directory down, so the -shm shared-memory sidecar went unprobed in exactly the career-data location the fix added). The rework derives the probes as a locations-by-suffixes cross-product (nine probes), matching the single-path shape the code actually uses, so it cannot go asymmetric again.

The discrimination still holds: each probe resolves through a real `git check-ignore` exit-code verdict (not a string match), not-ignored and unanswerable both fail, root-anchoring the rule reddens the three career-data probes, removing it entirely reddens six, and data/ stays covered by the blanket data/* rule. No .gitignore or product code changed since the approval.

Owned test verified sound. Holding formal approval until the branch is finalized (this one needs a fresh approval since the head moved), nothing blocking from the tests/ side.

Scott-Emberson (approved):

Issue #1430, opened 2026-07-02
Marking a row Applied should seed data/follow-ups.md with the default cadence


santifer:
Surfaced by a real apply-run field test (~12 applications, zero follow-ups seeded).

## Problem

The follow-up cadence is born dead: applying updates the tracker, but `data/follow-ups.md` only gets entries when the `followup` mode is run by hand. Nobody remembers to — so `followup-cadence.mjs` (and #1352's fix) protect a file that stays empty.

## Proposal

When a row transitions to `Applied`, seed `data/follow-ups.md` with the first follow-up date from `followup-cadence.mjs` defaults. Hook points: `modes/apply.md` Step 8 instruction + the future `set-status.mjs` CLI (one write-path). User-layer file, additive append only.

PR #3470, opened 2026-08-28
fix(set-status): a transition into Applied schedules nothing outside the agent path (#3459)

### LINKED ISSUE #3459: A transition into Applied only seeds a follow-up on the agent path — the web UI and the CLI schedule nothing
### What happened?

#1430 ("Marking a row Applied should seed data/follow-ups.md with the default cadence") is closed as completed, but the behaviour only happens on the agent path. A user who records an application from the web UI gets no follow-up scheduled, silently.

`set-status.mjs` announces the transition and stops there:

- `set-status.mjs:58` — `"followupSeedCandidate": true` — the hook point for seeding
- `set-status.mjs:576` — `...(statusChanged && newStatus === 'Applied' ? { followupSeedCandidate: true } : {})`

Nothing consumes that flag. `grep -n "followupSeedCandidate\\|seedFollowup" set-status.mjs` returns only those two lines. The only callers of `followup-seed.mjs` are `modes/apply.md` and `modes/followup.md` — agent instructions.

Since #2735 gave the Apply page a way to record that you applied, and #2901 converged `/api/status` onto `set-status.mjs`, the web path now writes the tracker and the transition ledger correctly and still schedules nothing.

This is the same shape as #2714 — a consumer the convergence has not reached yet — for a different artifact. #2901 fixed it for the status log because `set-status.mjs` already wrote that file itself. Follow-ups are the remaining case, because the seeding lives behind a flag instead of in the writer.

### Steps to reproduce

1. Have a row in `Evaluated`.
2. From the web UI, mark it Applied.
3. `grep` that row number in `data/follow-ups.md` — no pin.
4. Do the same transition with `node set-status.mjs <n> Applied` — also no pin. The CLI is affected too; only an agent following `modes/apply.md` gets one.

### Expected behavior

A transition into `Applied` seeds the follow-up regardless of which entry point performed it, which is what #1430 asked for.

The natural place is `set-status.mjs` itself, for the same reason #2901 moved the tracker write there: it is the one writer every other caller already delegates to, and it already holds the tracker lock at that moment. `followup-seed.mjs` already exports what is needed — `seedFollowup(appNum, options)` at `:536`, taking `{ trackerPath, date }`.

Three details that matter, from carrying this in a fork for a while:

1. **Seed before releasing the tracker lock**, so the seeder reads the row the process just wrote. Ordering is safe — nothing acquires the follow-ups lock before the tracker lock, so the two cannot deadlock against each other.
2. **Only pass `date` when `--on` actually named one.** Left to itself the seeder resolves the applied date from the row's own history, which is the better answer when nobody has said otherwise, and the wrong one when someone just did — a backdated transition would otherwise get a follow-up clock anchored on today.
3. **A seed failure must not fail the status write.** Catch it and report it in the JSON rather than letting it propagate; the status change has already committed by then.

I have this running as a roughly 20-line change and am happy to send it as a PR. Flagging it as an issue first since it moves `set-status.mjs` from announcing a candidate to performing the seed, which is a change in what that script is responsible for — your call whether that is the right home or whether the consumer belongs elsewhere.

### CLI tool

Not CLI-specific — affects the web UI and `set-status.mjs` directly.

### OS and Node.js version

macOS 26.5.2, Node v22.22.3

### Output of npm run doctor

Not included — reproducible from a clean checkout without any local profile or data.

abankar1:
Fixes #3459.

#1430 is closed as completed, but the behaviour only ever happened on the agent path. `set-status.mjs` *announced* the transition — `followupSeedCandidate: true` — and nothing consumed the flag. The only callers of `followup-seed.mjs` were `modes/apply.md` and `modes/followup.md`, both agent instructions.

So recording an application from the web UI, or from this CLI directly, wrote the tracker and the transition ledger correctly and scheduled nothing, silently.

## Where the fix goes

In `set-status.mjs`, not in each caller — which is what makes it one fix instead of three. #2901 converged `/api/status` onto this script, so the web path inherits it, and so does every future caller that delegates here rather than editing the table.

```
$ node set-status.mjs --row 7 Applied
📅 Follow-up seeded for #7: next 2026-02-08
✅ #7 Acme — Backend Engineer: set Evaluated → Applied
```

## Four decisions worth flagging

- **After the tracker lock is released.** `seedFollowup()` re-reads the tracker to resolve the applied date and must see the row this run just wrote. It takes only its own followups lock, never the tracker lock, so there's no ordering to get wrong.
- **`followupsPath` derived from the tracker's own directory.** `followup-seed`'s default is the *repo's* `data/follow-ups.md`, so an install whose data lives outside the checkout — the whole point of `CAREER_OPS_TRACKER`, and how the web configures its root — would have had the status written to one tracker and the follow-up seeded beside another. The status-log append derives its path the same way. **The first version of this fix got that wrong and wrote into the repo during development**, which is why there's a test asserting the repo stays untouched.
- **A seeding failure never fails the status change.** The write has already committed and the exit code is about that write — same policy and same wording as the status-log append directly above.
- **`force` on a dry run only.** The tracker is deliberately unwritten there, so the row still reads `Evaluated` and the preview would otherwise report a failure for the one thing the real run is about to do.

The "consider seeding follow-ups" advisory now prints only when seeding *didn't* happen — it asked the user to do by hand what now runs for them, and would otherwise sit directly under "Follow-up seeded".

## Verification

Six of the seven tests fail on `main`:

```
✖ a transition into Applied seeds the first follow-up
✖ it seeds beside the tracker it just wrote, not beside the repo
✖ re-running on an already-Applied row does not stack a second pin
✖ --dry-run previews the seed and writes nothing
✖ the JSON result reports what actually happened
✖ a seeding failure never fails the status change
ℹ pass 1  ℹ fail 6
```

The seventh — *a transition to any other status seeds nothing* — passes either way and is the guard: "seed on every transition" would satisfy all the others.

The failure-path test makes the follow-ups path a **directory** rather than mocking, so it exercises the real throw. `followupSeedCandidate` is kept in the JSON for anything already reading it; the outcome now travels beside it as `followupSeeded`.

`node set-status-tests.mjs` → 96 passed, 0 failed · `node test-all.mjs --quick` → green.

## Summary by CodeRabbit

- **New Features**
  - Status changes to **Applied** now automatically seed the first follow-up.
  - Dry-run mode previews follow-up seeding without writing changes.
  - Results report whether follow-up seeding occurred and its outcome.
  - Manual seeding guidance appears only when automatic seeding is unavailable.

- **Bug Fixes**
  - Follow-up seeding failures no longer prevent a completed status change.

- **Tests**
  - Added coverage for automatic seeding, idempotency, dry runs, status filtering, result reporting, and failure handling.


Scott-Emberson (commented):
Reviewed at head 07b0e57e75. set-status-seeds-followup.test.mjs drives the real `set-status.mjs` via spawnSync against a sandboxed tracker (the non-agent CLI path this PR fixes), with no mocking, and checks the real follow-ups.md. The positive (a direct `set-status Applied` writes exactly 1 pin) reddens to 0 if the seeding block is reverted, and the negative (an Interview transition seeds 0 pins) reddens if seeding fires unconditionally. Idempotency (re-run stays 1 pin), the `assumeApplied` preview path, correct-root seeding, and "seeding failure never fails the status change" are all asserted on real filesystem state.

Scott-Emberson (approved):

PR #3479, opened 2026-08-28
fix(agent-inbox): lock resolve rewrites

### LINKED ISSUE #3478: fix(agent-inbox): resolve can overwrite a concurrently queued item
## Bug

`agent-inbox.mjs add` serializes its append through `withPipelineLock(PATH, ...)`, but `resolve` reads the whole queue and rewrites it without taking that lock.

That makes this interleaving possible:

1. `resolve` reads the old queue.
2. `add` acquires the lock, appends a request, prints `Queued`, and exits 0.
3. `resolve` writes its stale snapshot and silently removes the accepted request.

This is separate from #3035's add-only CI failure: it is a deterministic missing lock on the other writer.

## Measured on current main

I acquired the queue's own pipeline lock, then spawned `agent-inbox.mjs resolve 1` against that queue. It exited 0 in 43 ms while the lock was still held:

```json
{"code":0,"elapsedMs":43,"out":"Resolved #1: 2026-01-01 00:00 — alpha"}
```

A writer honoring the queue lock cannot complete until the holder releases.

## Proposed scope

- Run the entire resolve read/modify/write transaction under `withPipelineLock(PATH, ...)`.
- Keep parsing and numbering inside the critical section so `list` → `resolve N` semantics use the locked snapshot.
- Add a regression that holds the lock, starts `resolve`, proves it waits, releases, and verifies only the requested item changes.
- Preserve current CLI output and error behavior.

I checked open and closed issues/PRs for `agent-inbox resolve`; no existing item covers this path.

tarang-tj:
Closes #3478.

## What changed

- serialize `resolve` through the same queue lock already used by `add`
- keep pending-item selection and the full read/modify/write in one critical section
- defer CLI failure until after lock release, avoiding a stranded lock on an invalid item number
- add a process-level regression that holds the lock, proves `resolve` waits, then verifies the selected and sibling rows

## Why

Previously, `resolve` could read an old queue, a concurrent `add` could append and report success, and `resolve` could then overwrite that accepted request with its stale snapshot. Holding the queue lock externally showed `resolve` still exiting 0 in 43 ms on current main.

## Verification

- `node agent-inbox-tests.mjs` — 42 passed, 0 failed
- `npm run lint` — 585 `.mjs` files passed
- `node test-all.mjs --quick` — 7288 passed, 0 failed, 11 pre-existing/environment warnings
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved inbox resolution reliability when multiple actions occur at the same time.
  * Prevented newly added items from being overwritten during resolution.
  * Ensured resolution waits for ongoing inbox updates to finish.
  * Improved handling and reporting when the selected item is missing.

* **Tests**
  * Added regression coverage for concurrent inbox operations and preservation of pending items.

### DIFF
--- a/agent-inbox-tests.mjs
+++ b/agent-inbox-tests.mjs
@@ -9,6 +9,8 @@
  *   3. `list` shows pending only; `list --all` shows resolved items too.
  *   4. `resolve N` ticks the N-th *pending* item and appends a one-line result,
  *      so `list` then `resolve N` line up.
+ *   4b. resolve holds the same lock as add for its full read/modify/write, so a
+ *       request accepted concurrently cannot be overwritten by a stale view.
  *   5. An empty `add` fails loudly (exit 1) rather than queuing a blank line.
  *   6. On the default path, a first `add` self-heals .gitignore (idempotent) so
  *      the personal queue isn't accidentally tracked.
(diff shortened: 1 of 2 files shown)


luochen211 (approved):
Reviewed at `789c1842f0ba`. The fix puts the complete resolve read/modify/write transaction under the same file-scoped lock used by `add`: pending-item numbering, the line snapshot, and the rewrite are now derived from one locked view, so an acknowledged concurrent append cannot be overwritten by stale contents. Returning an error from the callback and calling `fail()` only after `withPipelineLock()` completes also preserves its `finally`-based release path.

I ran `node agent-inbox-tests.mjs` locally (42 passed, including the real child-process lock wait and sibling-item survival assertions), ran `git diff --check origin/main...HEAD`, and verified the GitHub matrix/CodeQL checks are green. No blocking issue found.

PR #3482, opened 2026-08-28
fix(web): reject malformed follow-up IDs

### LINKED ISSUE #3480: fix(web): reject malformed follow-up record identifiers
## Bug

The follow-up write APIs parse identifiers with `Number.parseInt`, so a numeric prefix is treated as a real record identifier and zero is accepted.

On current main:

- POST `/api/followups/log` with `appNum: "42junk"` returns 200 and persists application 42.
- POST `/api/followups/override` with `appNum: "0"` returns 200 and persists `#0`.
- DELETE `/api/followups/log` with `num: "1junk"` returns 200 and deletes valid row 1.

This contradicts the core contract in `followup-seed.mjs`, which requires an integer greater than zero. It is a data-integrity bug: malformed client input can target a different valid record.

## Reproduction

Against the real Next routes on current main:

```text
POST log {appNum:"42junk", ...}  -> 200, persisted | 1 | 42 | ... |
POST override {appNum:"0", ...} -> 200, persisted - next #0 ...
DELETE log {num:"1junk"}        -> 200, removed valid row #1
```

## Proposed scope

- Add one shared parser for trimmed positive safe integers.
- Use it for log POST `appNum`, log DELETE `num`, override POST `appNum`, and override DELETE `appNum`.
- Preserve positive JSON numbers and digit-only strings.
- Reject prefixes, zero, negatives, decimals, exponent forms, objects/arrays, and unsafe integers before any write.
- Add pure boundary tests plus integration coverage proving all four handlers use the parser.

No tracker-existence validation or stored-table repair in this change. I searched open/closed issues and PRs for these routes, `parseInt`, numeric-prefix IDs, and application zero; no existing report or open implementation covers it.

tarang-tj:
Closes #3480.

## What changed

- add one strict parser for follow-up and application identifiers
- use it in all four log/override mutation handlers
- reject coercible prefixes, zero/negative/decimal values, and unsafe integers
- cover accepted and rejected boundaries plus route wiring

## Why

`Number.parseInt` accepted a valid-looking prefix and discarded the rest. In the real routes, `42junk` could write application 42 and `1junk` could delete follow-up row 1; application `0` was also accepted by both POST handlers.

The shared parser now accepts only positive safe-integer numbers or canonical digit strings, so every mutation endpoint has the same contract as the core follow-up tooling.

## Verification

- `node --test web/tests/lib/followup-id.test.mjs` — 3 passed
- `cd web && npm test` — 388 passed, 0 failed
- `cd web && npm run typecheck` — passed
- `npm run lint` — 587 syntax checks passed
- `node test-all.mjs --quick` — 7290 passed, 0 failed, 11 existing warnings
- `git diff --check` — passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of follow-up application and log identifiers.
  * Invalid, unsafe, or incorrectly formatted identifiers are now rejected consistently.
  * Existing error handling and follow-up actions remain unchanged.

* **Tests**
  * Added coverage for valid and invalid identifier formats, including boundary cases.

### DIFF
--- a/web/src/app/api/followups/log/route.ts
+++ b/web/src/app/api/followups/log/route.ts
@@ -2,6 +2,7 @@ import fs from "node:fs";
 import path from "node:path";
 import { atomicWrite } from "@/lib/core/safe-write";
 import { CHANNELS, isRealISODate, localISODate } from "@/lib/followups";
+import { parseFollowupId } from "@/lib/followup-id.mjs";
 import { followupsLogPath, withFollowupsWrite, followupsWriteError } from "@/lib/followups-server";
 
 export const runtime = "nodejs";
@@ -44,8 +45,8 @@ export async function POST(req: Request) {
     return Response.json({ error: "bad json" }, { status: 400 });
   }
 
-  const appNum = Number.parseInt(String(body.appNum ?? body.num ?? ""), 10);
-  if (!Number.isInteger(appNum) || appNum < 0) {
+  const appNum = parseFollowupId(body.appNum ?? body.num);
+  if (appNum === null) {
     return Response.json({ error: "appNum (application #) required" }, { status: 400 });
   }
   const company = cell(body.company, 80);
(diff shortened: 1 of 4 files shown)


luochen211 (approved):
Reviewed at `939a899b0596`. All four mutation boundaries now call the same strict parser before constructing a row, matching a stored row, or building the override regex. The parser accepts only positive safe integers (numeric or canonical decimal strings) and rejects the previous prefix-coercion cases such as `42junk`, exponent/decimal forms, leading-zero strings, non-positive values, and unsafe integers.

I traced the values through both POST/DELETE routes, ran `node --test web/tests/lib/followup*.test.mjs` (12 passed; 2 unrelated lock tests self-skipped because this web-only checkout lacks core dependencies), and ran `git diff --check origin/main...HEAD`. Current CI is green. No blocking issue found.

PR #3483, opened 2026-08-28
fix(dashboard): bound company fallback matches

### LINKED ISSUE #3481: fix(dashboard): company-prefix fallback can select the wrong CV
## Bug

The terminal dashboard's legacy CV fallback uses raw substring matching for company slugs of three or more characters. As a result, a `Meta` application matches a `Metabase` artifact.

This reaches destructive/user-facing actions:

- lower-case `d` opens the first resolved candidate, so a Meta row opens Metabase's CV;
- upper-case `D`, with an unkeyed manifest row, can emit regeneration paths for Metabase's HTML/PDF while the selected application is Meta.

The web resolver already enforces token boundaries and explicitly documents the `Meta` / `Metabase` case; the dashboard diverges from that contract.

## Reproduction on current main

With only `output/cv-jane-doe-metabase-2026-08-29.pdf` present:

```text
ResolvePDFs(Meta) -> [output/cv-jane-doe-metabase-2026-08-29.pdf]
d                 -> PipelineOpenPDFMsg{Path:"...metabase...pdf"}
D + unkeyed index -> PipelineGeneratePDFMsg{HTMLPath:"...metabase...html", PDFPath:"...metabase...pdf"}
```

## Proposed scope

- Require a hyphen-delimited company slug boundary for fallback filenames, including slugs of three or more runes.
- Add resolver coverage for `Meta` versus `Metabase`.
- Add UI regressions proving `d` and unkeyed-manifest `D` emit no command for a prefix-only artifact.
- Do not redesign the manifest, picker, or web resolver.

I searched open/closed issues and PRs for `Meta Metabase`, `matchesCompanySlug`, wrong dashboard PDF, company-prefix collisions, and token-boundary PDF matching. No active report, owner, or implementation covers this case.

tarang-tj:
Closes #3481.

## What changed

- require the company slug to occupy a complete hyphen-delimited filename segment
- preserve legacy role-variant and multiword-company fallback matches
- add resolver coverage for the `Meta` / `Metabase` prefix collision
- add dashboard regressions for both `d` (open PDF) and `D` (regenerate from manifest source)

## Why

The pre-manifest fallback used raw substring matching for company slugs of three or more characters. That made a selected `Meta` application resolve `Metabase` artifacts, so `d` could open the wrong CV and `D` could regenerate from the wrong HTML when a manifest entry was unkeyed.

Generated CV names already delimit company slugs with hyphens. Matching `-<slug>-` keeps intended files such as `-anthropic-staff-ui-...` and multiword slugs while preventing one company from matching a longer company name.

## Verification

- `cd dashboard && go test ./... -count=1` — all packages passed
- focused resolver, `d`, and `D` regressions — passed
- `gofmt` — clean
- `git diff --check` — passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CV PDF matching to prevent company-name prefixes from being treated as valid matches.
  * Prevented incorrect PDFs from opening when company names are similar, such as “Meta” and “Metabase.”
  * Prevented regeneration attempts when no exact matching source document is available.
  * Added clearer feedback when a matching PDF or source document cannot be found.

### DIFF
--- a/dashboard/internal/data/pdf.go
+++ b/dashboard/internal/data/pdf.go
@@ -172,7 +172,8 @@ var rePDFDate = regexp.MustCompile(`(\d{4}-\d{2}-\d{2})\.pdf$`)
 //  1. Manifest entry for the application's report number, when the file
 //     still exists. This is exact — generate-pdf.mjs recorded the linkage.
 //  2. Filename match: output/cv-*.pdf whose name contains the kebab-cased
-//     company. This covers every PDF generated before the manifest existed.
+//     company at hyphen boundaries. This covers every PDF generated before
+//     the manifest existed without matching a longer company's slug.
 //     Multiple matches are all returned (newest first) so the caller can
 //     offer a picker instead of guessing — one company can have several
 //     role-variant CVs from the same day.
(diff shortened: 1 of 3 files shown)


Scott-Emberson (commented):
Reviewed at head 5efad6f2d3. The company-fallback bound has real teeth on both layers. Data resolver: `TestResolvePDFsDoesNotMatchCompanyPrefix` asserts Meta against a metabase PDF returns zero (under the old substring logic "meta" matched "metabase", so this reddens), while the multi-word (Monarch Money) and role-variant (anthropic / anthropic-staff-ui) positives still resolve, so the bound isn't just "reject everything". TUI: `pipeline_pdf_test.go` pairs the same negative (no open cmd, a flash instead) with a positive regenerate driven from a real manifest row. Both the positive and negative cases flip red if the segment bound reverts to substring containment.

Owned test verified sound. Holding formal approval until the branch is finalized, nothing blocking from the dashboard/ side.

Scott-Emberson (approved):
Approving from the dashboard/ side. The owned test is verified sound at this head (evidence in my review note above), CI is green, and the branch is mergeable. Good to go from our end. As always this routes the code-owner review, it does not merge.

PR #3489, opened 2026-08-28
test: relax dashboard lock wait on Windows runners

chipoto69:
## Summary
- Raises the dashboard shared-lock test post-release wait from 2s to 10s.
- Targets the Windows CI flake seen on PR #3463 where TestUpdateApplicationStatusWaitsForSharedLock timed out at 2.16s after lock release.
- Keeps the lock-bypass assertion intact; this only gives slow Windows runners enough time to observe the released lock.

## Test Plan
- git diff --check
- npm run lint
- cd dashboard && go test ./internal/data
- cd dashboard && go test ./...

Related: #3463

## Summary by CodeRabbit

* **Tests**
  * Increased the wait time allowed for a dashboard update test to complete after a shared lock is released, reducing potential failures caused by timing delays.


Scott-Emberson (commented):
Reviewed at head 5fb4db82a1. The Windows lock-wait relaxation is safe: it only widens the failure deadline (time.After 2s to 10s), it doesn't touch the assertions that give this test teeth. The 150ms select that proves the shared lock held (career_test.go:63-70, which trips t.Fatal if an update bypasses the lock) is unchanged, the success path still requires <-done with err==nil, and the post-resume checks still assert the row shows | Interview | and the concurrent row survives. A longer deadline can't turn a broken lock or a non-resuming update into a pass.

Owned test verified sound. Per our current process I'm holding formal approval until the branch is finalized, but nothing is blocking from the dashboard/ side.

Scott-Emberson (approved):
Approving from the dashboard/ side. The owned test is verified sound at this head (evidence in my review note above), CI is green, and the branch is mergeable. Good to go from our end. As always this routes the code-owner review, it does not merge.

PR #3494, opened 2026-08-29
fix(funnel-velocity): fold status-log history into calibration rates (#3493)

### LINKED ISSUE #3493: stats.mjs: ever* funnel undercounts interviews and responses — terminal Rejected erases the furthest stage reached
### Summary

`computeFunnel()` in `stats.mjs` derives the `ever*` funnel from each row's **current** status only. Because `Rejected` and `Discarded` are terminal states that replace whatever stage the candidate actually reached, every application that reached an interview and was then rejected is counted as "applied, never interviewed".

The result is that `interviewRate` and `responseRate` systematically undercount, and `funnel-velocity.mjs` inherits the error because it reads `funnel.interviewRate` directly (line ~273).

### Where

```js
// stats.mjs
const everApplied   = n('Applied') + n('Responded') + n('Interview') + n('Offer') + n('Hired') + n('Rejected');
const everResponded = n('Responded') + n('Interview') + n('Offer') + n('Hired');   // omits Rejected
const everInterview = n('Interview') + n('Offer') + n('Hired');                    // omits Rejected
```

`everApplied` correctly includes `Rejected` (the comment notes "a rejection proves a submission"). The same reasoning is not applied one line down: a rejection *email* proves a response, and a rejection *after a screen* proves an interview.

### Reproduce

1. `node set-status.mjs --report N Applied`
2. `node set-status.mjs --report N Interview`
3. `node set-status.mjs --report N Rejected`
4. `node stats.mjs --summary`

`everInterview` does not count row N. The `status-log.tsv` ledger records the `Interview` transition correctly, but `computeFunnel()` never reads it.

### Impact

On a real tracker with 64 ever-applied rows, 8 of which demonstrably reached a recruiter screen, tech screen, or video round (all documented in the Notes column and now in `status-log.tsv`):

```
reported:  responded 3 (4.7%)  ->  interview 1 (1.6%)
actual:    responded >= 8      ->  interview 8 (~12%)
```

`funnel-velocity.mjs` then reports "Interview rate: 1.1% vs 2-4% typical band - below the typical band" and advises checking follow-up compliance or the score threshold. The true rate is above the band, and the real problem is conversion *after* the interview. The advice points at the wrong stage.

This is the failure mode the metric is most likely to be trusted for: a user reading "below typical band" concludes they have a targeting or CV problem, when they have an interview-performance problem.

### Suggested fix

Derive furthest-stage-reached per row from `data/status-log.tsv` when the row has dated transitions, falling back to current status when it does not (rows predating the ledger). Something like:

```js
const reached = maxStage(ledgerStagesFor(row) ?? [row.currentStatus]);
```

A cheaper interim fix: keep the current computation but state the limitation in the `--summary` output, so the number is not read as "ever reached" when it means "currently sitting in or past".

### Note

Backfilling `Interview` transitions into `status-log.tsv` does **not** work around this, since `computeFunnel()` does not read the ledger. The row must end in a terminal state, and terminal states are exactly what erases the signal.

### Environment

career-ops v1.29.0

AdilAzhariOmsan:
Fixes #3493.

## Problem

`analyze()` in `funnel-velocity.mjs` derives the funnel from `computeFunnel(stats.byStatus)` — each row's **current** tracker status only. Because `Rejected` and `Discarded` are terminal states that replace whatever stage a row actually reached, every application that got an interview and was then rejected is counted as "applied, never interviewed". So `responseRate` and `interviewRate` systematically undercount.

The downstream effect is the one the issue calls out: `funnel-velocity --summary` tells a user their interview rate is "below the typical band" and points them at follow-up compliance or their score threshold, when the real problem is post-interview conversion — the failure mode the metric is most likely to be trusted for.

## Fix

`stats.mjs` already solved this for `node stats.mjs` in #1428 with `computeFunnelWithHistory()`, which folds `data/status-log.tsv` so a row now sitting in a terminal snapshot still counts for the stages it passed through. `funnel-velocity.mjs` never adopted it.

- `analyze()` uses `computeFunnelWithHistory(trackerStatusByNum(trackerContent), parseStatusLogStages(logContent))` when the transition log is present, and falls back to the snapshot `computeFunnel()` when it is not — rows predating the ledger stay lower bounds, exactly as `stats.mjs` does. Same canonical `ever*` definition, no third copy.
- `computeCalibration()` now reports `basis: 'ledger' | 'snapshot'`.
- `renderSummary()` prints a one-line note when the rates fold history, so a folded number is never read as a raw snapshot.

Scope is `funnel-velocity.mjs` only. `computeFunnel` is still imported and still used for the no-ledger path.

## Tests

`node funnel-velocity.mjs --self-test` — added a block (`#3493`): a 25-row fixture with three rows currently `Rejected` whose ledger shows they reached `Responded`/`Interview`.

- snapshot basis → `interviewRate` 0%
- ledger basis → `everApplied` still 25, `interviewRate` 8% (2/25), `responseRate` 12% (3/25), summary carries the fold note

Full suite: `node test-all.mjs --quick` → `7285 passed, 1 failed`. The one failure is the pre-existing Windows symlink-materialization issue (#3364) and is identical with and without this change.

Manual before/after on a synthetic tracker with one interviewed-then-rejected row: interview rate `0% → 50%`.

https://claude.ai/code/session_01HK9dmjbU8R6QCjQAGXjJZc

## Summary by CodeRabbit

- **New Features**
  - Calibration funnel results now account for historical transitions, including items that ended in terminal states.
  - Results identify whether calculations are based on transition history or the latest snapshot.
- **Bug Fixes**
  - Improved stage counts and rate calculations when records moved through intermediate stages before being rejected or discarded.

### DIFF
--- a/funnel-velocity.mjs
+++ b/funnel-velocity.mjs
@@ -4,7 +4,9 @@
  *
  * Three payloads, decreasing availability:
  *   1. calibration — own funnel rates (canonical ever* definition imported from
- *      stats.mjs) vs candidate-side market benchmark ranges. Works day one.
+ *      stats.mjs) vs candidate-side market benchmark ranges. Works day one;
+ *      folds status-log history when the ledger is present so a row rejected
+ *      after an interview still counts as having reached Interview (#3493).
  *   2. waiting — in-flight Applied rows and elapsed days vs the typical
  *      first-response window. Per-row factual reporting, not an aggregate claim.
  *   3. velocity — median/p75 days per stage hop, folded from the append-only
@@ -38,7 +40,7 @@ import { readFileSync, existsSync } from 'fs';
 import { join, dirname } from 'path';
 import { fileURLToPath } from 'url';
 import * as yaml from 'js-yaml';
-import { computeFunnel, computeTrackerStats } from './stats.mjs';
+import { computeFunnel, computeFunnelWithHistory, parseStatusLogStages, trackerStatusByNum, computeTrackerStats } from './stats.mjs';
 import { resolveColumns, parseTrackerRow } from './tracker-parse.mjs';
 import { resolveTrackerPath, loadCanonicalStates, resolveCanonicalState } from './tracker-utils.mjs';
 import { parseAppliedDate, normalizeStatus } from './followup-cadence.mjs';
@@ -272,6 +274,9 @@ export function computeCalibration(funnel, benchmarks) {
     everApplied: funnel.everApplied,
     smallSample,
     claimMinN: CLAIM_MIN_N,
+    // 'ledger' when the ever* counts fold status-log history (rejected-after-
+    // interview rows counted at the stage they reached), 'snapshot' otherwise.
+    basis: funnel.basis === 'ledger' ? 'ledger' : 'snapshot',
     responseRate: classify(funnel.everApplied > 0 ? funnel.responseRate : null, benchmarks.response_rate),
     interviewRate: classify(funnel.everApplied > 0 ? funnel.interviewRate : null, benchmarks.application_to_interview),
   };
@@ -335,7 +340,14 @@ export function parseTrackerRows(content) {
 export function analyze({ trackerContent, logContent, benchmarks, states, todayStr }) {
   const rows = parseTrackerRows(trackerContent);
   const stats = computeTrackerStats(trackerContent);
-  const funnel = computeFunnel(stats.byStatus);
+  // Ledger-aware funnel when the transition log is present: a row now sitting in
+  // a terminal snapshot (Rejected/Discarded) still counts for the middle stages
+  // it passed through, so interviewRate/responseRate stop undercounting rejected-
+  // after-interview rows (#3493). Same canonical definition `node stats.mjs`
+  // uses; falls back to the snapshot funnel when there is no log.
+  const funnel = logContent && logContent.trim()
+    ? computeFunnelWithHistory(trackerStatusByNum(trackerContent), parseStatusLogStages(logContent))
+    : computeFunnel(stats.byStatus);
   const { observations, unparseable, unknownSources } = parseStatusLog(logContent, states);
   const timelines = foldObservations(observations);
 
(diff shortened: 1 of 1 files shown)


AdilAzhariOmsan:
Rebased on `main` (was 34 behind) and pushed a follow-up commit (ec2cf92) covering your double-count question and the two CodeRabbit findings.

**Double-counting a repeated transition — can't happen, now pinned by tests.**
`computeFunnelWithHistory` keys `reached` by tracker row number and only ever bumps a row to the *max* stage rank it held; the `everX` totals then count distinct rows over `reached.values()`, never transitions. So a duplicated ledger line (`bump` is idempotent), a torn-then-rewritten append, and a row that genuinely bounces `Responded→Interview→Responded→Interview` all land the row once at its deepest stage. New self-test cases:
- every `ledgerLog` line doubled → identical `everApplied` / `interviewRate` / `responseRate`
- a bouncing row → counted once into `everInterview` (1/25 = 4%)

**CodeRabbit — `basis` on a history-less log (valid, fixed in ec2cf92).**
`logContent.trim()` was truthy for a header-only / comment-only / all-torn log, so `analyze()` took the ledger path, stamped `basis:'ledger'`, and printed the "rates fold history" note over a funnel identical to the snapshot. Now branches on `parseStatusLogStages(logContent).length`. Test covers header-only, comment-only, torn-rows-only, and whitespace logs.

**CodeRabbit — no-history `Discarded` and `everApplied` (working as intended, test added).**
A current `Discarded` row with no ledger history gets rank 0 and stays out of `everApplied` — this matches canonical `computeFunnel`, which counts `Rejected` into `everApplied` but not `Discarded`. That asymmetry is deliberate: a rejection proves a submission, a discard doesn't (a row can go `Evaluated → Discarded` without ever applying). A `Discarded` row that *does* carry ledger history still folds normally (`Offer → Discarded` counts into `everOffer`). Added a test pinning that the snapshot and ledger paths agree on the no-history case so they can't silently diverge; happy to change the semantics if you'd rather `Discarded` count, but that's a change to the canonical funnel in `stats.mjs`, out of scope for this PR.

`node funnel-velocity.mjs --self-test` green; full `test-all.mjs` — 7568 passed, 0 failed.

PR #3496, opened 2026-08-29
fix(calibrate): resolve the tracker colmap from the line array, not the header line

### LINKED ISSUE #3495: calibrate.mjs reads the tracker with LEGACY_COLMAP on trackers with a Via column, silently dropping every row
## Summary

`calibrate.mjs` reads the tracker with the wrong column map on any tracker that carries the optional `Via` column (#1596). Every row silently drops out of the population, and the report renders `0 resolved / 0 in-flight` with an "insufficient data" verdict on a tracker full of recorded outcomes.

## Cause

`loadTrackerRows()` (calibrate.mjs):

```js
const headerLine = lines.find((l) => isHeaderRow(l));
const colmap = headerLine ? resolveColumns(headerLine) : undefined;
```

`resolveColumns()` takes the **line array** and locates the header itself:

```js
export function resolveColumns(lines) {
  return detectColumns(lines) || LEGACY_COLMAP;
}
```

Handed a single string, `detectColumns()` iterates it character by character, matches no header, and falls back to `LEGACY_COLMAP` — the 9-column order with no `Via` column.

`Via` sits between `Company` and `Role`, so under `LEGACY_COLMAP` every field from `role` onward reads one column to the left: `status` lands on the score cell (`"4.5/5"`), no canonical status matches, and the row is dropped.

Every other `resolveColumns()` caller in the repo (`stats.mjs`, `set-status.mjs`, `analyze-patterns.mjs`, `outcome.mjs`, `funnel-velocity.mjs`, `tracker.mjs`, … 25 in total) passes `lines`. `calibrate.mjs` is the only one that does not.

## Reproduce

```
| # | Date | Company | Via | Role | Score | Status | PDF | Report | Notes |
|---|------|---------|-----|------|-------|--------|-----|--------|-------|
| 1 | 2026-01-02 | Acme | Hays | Backend Dev | 4.5/5 | Applied | ✅ | [1](reports/1.md) | note |
```

`node calibrate.mjs` on a tracker in that shape:

```
Resolved outcomes: 0 · in flight (not counted): 0 · unscored: 0
Verdict: Not enough resolved outcomes to compare score bands yet (0 resolved, floor 5 per band).
```

Same tracker, colmap resolved from `lines`: `10 resolved · 9 in flight`.

## Why it matters

The failure is silent. `/calibrate` does not error — it reports "not enough history yet", which reads as a normal state for a young search. A user with months of recorded outcomes is told the tool has nothing to say, and has no signal that it read nothing at all.

## Fix

One line, plus a regression suite. `calibrate.mjs --self-test` covers `computeCalibration()` against in-memory rows, so nothing exercised the tracker-loading step — hence the gap. PR follows.

thibmonier:
Fixes #3495.

## The bug

`loadTrackerRows()` in `calibrate.mjs` built its column map from a single header string:

```js
const headerLine = lines.find((l) => isHeaderRow(l));
const colmap = headerLine ? resolveColumns(headerLine) : undefined;
```

`resolveColumns()` takes the **line array** — it locates the header itself. Given a string, `detectColumns()` iterates it character by character, matches no header, and falls back to `LEGACY_COLMAP`, the 9-column order with no `Via` column.

`Via` sits between `Company` and `Role`, so on a tracker carrying it (#1596) every field from `role` onward reads one column to the left: `status` lands on the score cell (`"4.5/5"`), no canonical status matches, and the row is dropped from the population.

Result on a real tracker with 19 tracked applications:

```
Resolved outcomes: 0 · in flight (not counted): 0 · unscored: 0
Verdict: Not enough resolved outcomes to compare score bands yet (0 resolved, floor 5 per band).
```

After the fix, same tracker: `10 resolved · 9 in flight`.

The failure mode is what makes it worth fixing rather than documenting: it does not error. It reports "not enough history yet", which is indistinguishable from the normal state of a young search — so the user has no signal that the tool read nothing.

## The fix

One line: pass `lines`, as all 25 other `resolveColumns()` callers in the repo do (`stats.mjs`, `set-status.mjs`, `analyze-patterns.mjs`, `outcome.mjs`, `funnel-velocity.mjs`, `tracker.mjs`, …). `calibrate.mjs` was the only caller passing a string.

## Test coverage

`calibrate.mjs --self-test` exercises `computeCalibration()` against in-memory rows, so nothing covered the step that turns `applications.md` into those rows — which is why this survived.

`tests/calibrate-tracker-colmap.test.mjs` covers both tracker shapes: with `Via` (status, score and company each read from their own column) and the legacy 9-column layout (unchanged). `loadTrackerRows` is exported so the external suite can reach it, following the `company-history.mjs` precedent. The suite is auto-discovered, uses `pass`/`fail` from `tests/helpers.mjs`, and calls neither `process.exit()` nor `finish()`.

Verified: **fails before the fix** (the two column-position assertions go red), **passes after**.

## Suite

`node test-all.mjs --quick` → `7297 passed, 0 failed, 11 warnings` (baseline on `main`: `7288 passed, 0 failed, 11 warnings`).

One note in case it is useful: on the first run the LaTeX assertion `localized section titles validate` went red with `null`, and passed on re-run. `latexValidate()` returns `null` whenever the `generate-latex.mjs` child exceeds its 30 s timeout, so under load the assertion reports as a validation failure rather than as a timeout. Unrelated to this change, and I have not touched it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed tracker calibration so records with an optional Via column are read correctly.
  * Prevented valid tracker rows from being silently excluded from calibration.

* **Tests**
  * Added regression coverage for trackers both with and without the Via column.

### DIFF
--- a/calibrate.mjs
+++ b/calibrate.mjs
@@ -212,10 +212,18 @@ export function computeCalibration(rows, journals, opts = {}) {
 
 // --- Filesystem assembly --------------------------------------------------
 
-function loadTrackerRows(appsFile) {
+export function loadTrackerRows(appsFile) {
   const lines = readFileSync(appsFile, 'utf-8').split(/\r?\n/);
-  const headerLine = lines.find((l) => isHeaderRow(l));
-  const colmap = headerLine ? resolveColumns(headerLine) : undefined;
+  // resolveColumns() takes the LINE ARRAY and locates the header itself.
+  // Passing it a single header string made detectColumns() iterate that
+  // string character by character, find no header, and fall back to
+  // LEGACY_COLMAP — the 9-column order with no Via column. On a tracker that
+  // HAS Via (#1596), where the column sits between Company and Role, every
+  // field from `role` onward then reads one column to the left: `status`
+  // reads the score cell ("4.5/5"), no status matches, and every row silently
+  // drops out of the population. The report renders 0 resolved / 0 in-flight
+  // and a false "insufficient data" verdict on a tracker full of outcomes.
+  const colmap = resolveColumns(lines);
   const rows = [];
   for (const line of lines) {
     if (!line.trim().startsWith('|') || isHeaderRow(line) || isSeparatorRow(line)) continue;
--- a/tests/calibrate-tracker-colmap.test.mjs
+++ b/tests/calibrate-tracker-colmap.test.mjs
@@ -1,0 +1,76 @@
+/**
+ * calibrate-tracker-colmap.test.mjs — regression suite for calibrate.mjs's
+ * tracker loader.
+ *
+ * calibrate.mjs's `--self-test` covers computeCalibration() against in-memory
+ * rows, so it never exercises the step that turns applications.md into those
+ * rows. That gap hid a column-map bug: loadTrackerRows() called
+ * `resolveColumns(headerLine)` with a single string, but resolveColumns()
+ * takes the LINE ARRAY and finds the header itself. detectColumns() iterated
+ * the string character by character, matched no header, and fell back to
+ * LEGACY_COLMAP — the 9-column order with no Via column.
+ *
+ * On a tracker carrying the optional Via column (#1596), which sits between
+ * Company and Role, every field from `role` onward then reads one column to
+ * the left: `status` reads the score cell ("4.5/5"), no canonical status
+ * matches, and every row drops out of the population. The failure is silent —
+ * the report renders "0 resolved / 0 in-flight" and an "insufficient data"
+ * verdict rather than an error, so a user reads it as "not enough history yet"
+ * on a tracker full of recorded outcomes.
+ *
+ * Every other resolveColumns() caller in the repo passes `lines`.
+ *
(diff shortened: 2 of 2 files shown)


Scott-Emberson (commented):
Reviewed at head 79103ff6c1. `calibrate-tracker-colmap.test.mjs` builds a real Via-column tracker and asserts on concrete column values that only resolve when the colmap is built from the parsed line array. The fix (`resolveColumns(lines)` instead of `resolveColumns(headerLine)`) matters because `detectColumns` iterates its argument with for..of: a bare header string iterates char by char, matches no header, and falls back to LEGACY_COLMAP. The `status === Applied/Rejected` and `score === 4.5/2.2` assertions are the discriminating ones, both redden under the legacy fallback (status maps to the score cell, score maps to a role string and goes NaN). `viaRows.length === 2` alone doesn't discriminate, but the status/score checks do the real work.

Scott-Emberson (approved):

PR #3502, opened 2026-08-29
fix(web): identify the verdict block by its heading, not by the letter F (#3416)

### LINKED ISSUE #3416: Web UI: report block lettered F is always shown as a 'Verdict' callout, even when it's the Interview Plan table

### What happened?

The web report viewer (`web/src/components/report-view.tsx`) assumes the core always writes a `## F) Verdict` block, and promotes whatever section is lettered F into a small highlighted "Verdict" callout styled for a single sentence:

```js
const verdict = sections.find((s) => s.letter === "F");
```

The current core report schema (`modes/oferta.md`) has no "Verdict" block. Blocks are:

```
A) Role Summary
B) Match with CV
C) Level and Strategy
D) Comp and Demand
E) Customization Plan
F) Interview Plan
G) Posting Legitimacy
H) Draft Application Answers
```

F is always "Interview Plan" — an 8-column STAR+R table (`#`, JD Requirement, STAR+R Story, S, T, A, R, Reflection). That table gets forced into the narrow Verdict callout instead of the normal collapsible-section treatment the other lettered blocks get.

This compounds with a separate table-styling issue: `.report-prose table` in `globals.css` is forced to `width: 100%` with `word-break: break-word` and no horizontal scroll. Any table with more than ~4 columns (the Interview Plan table above is the worst case, but any dense table hits this) gets its columns squeezed to a few characters wide, with words breaking mid-token. Together these make the Interview Plan section close to unreadable.

### Steps to reproduce

1. Generate any evaluation report — block F (Interview Plan) always contains the STAR+R table.
2. Open that report in the web UI at `/pipeline/{id}`.
3. Observe: the table renders inside a highlighted "Verdict" callout near the top of the page (not in its normal position between Customization Plan and Posting Legitimacy), with columns squeezed unreadably narrow and words breaking mid-token.

### Expected behavior

- Nothing in the A–H schema is a short verdict paragraph, so nothing should be specially promoted as "the verdict." The header already shows the score plus a Recommended / Below-the-apply-line badge, which covers the at-a-glance verdict.
- Wide tables should stay readable: horizontal scroll with a sane per-column minimum width, instead of forcing everything into the container width.

I have a fix for both parts, verified locally against a real report (screenshots on request) — removes the `letter === "F"` special case so F flows through the normal collapsible-section rendering like C/D/E/G, and changes `.report-prose table` to `display: block; overflow-x: auto` with per-cell `min-width`/`max-width` instead of forced `width: 100%` + `word-break`.

### Version / channel

0.7.1-alpha · e47599c (commit immediately after syncing this checkout to core v1.30.0, i.e. stock/unmodified web code)

### Screen / route

`/pipeline/{id}` — Interview Plan (block F) section specifically

### CLI tool

Claude Code

### OS / Browser

macOS (Darwin) — reproduced via Playwright/Chromium; layout-only bug, not browser-specific

maxmilian:
## What does this PR do?

Stops the report viewer from treating "whatever block is lettered F" as the Verdict. The callout is now selected by the heading itself, so the Interview Plan table is no longer rendered into a box built for one sentence.

## Related issue

Fixes #3416

## Two things worth adding to the issue's account

**The callout has been wrong since the day it landed, not since a later schema change.** #3416 describes the current schema as having no Verdict block. It is stronger than that: `30fa1d1` (#1535), the commit that introduced the callout, changed only `web/src/components/report-view.tsx`, and `git show 30fa1d1:modes/oferta.md` already read `## F) Interview Plan`. There was never a Verdict block for it to find.

**It was never English-only.** All 19 localized modes that write lettered blocks put Interview Plan at F — `## F) Plan rozmów kwalifikacyjnych`, `## F) План співбесід`, `## F) 面試準備計畫`, `## F) 면접 준비 계획`, and so on. A letter-based rule is wrong in every language simultaneously, which is what makes the letter the wrong key rather than the mapping being out of date.

## How it is identified now

Not by the word, and not by the letter — by the authoring marker `cleanHeading` has always stripped: a trailing `(lead)` / `(verdict)`. That is the signal the core actually promises, and it is language-independent, so a Spanish `## F) Veredicto (lead)` reads the same as an English one. A block plainly titled `Verdict` is accepted too, with its letter stripped first, while a heading that merely mentions the word (`D) Verdict rationale and caveats`) is not.

Since no mode writes a verdict block today, the practical result is that there is no callout and every block renders through the normal progressive disclosure — which is the correct rendering for the current schema. If the core ever adds one, at any letter and in any language, it is picked up without another change here.

The predicate lives in `report-sections.mjs` rather than in the component, for the reason that file already gives for splitting and heading cleanup: a rule duplicated in the viewer is what let `H) Draft Application Answers` drift in #2324.

## Verification

Red/green against the rule itself, not just a green run: reverting `isVerdictHeading` to the old `authorLetter(heading) === "F"` turns the three new tests red (13 pass / 3 fail); restoring it gives 16 pass / 0 fail.

- `node --test tests/lib/report-sections.test.mjs` — 16 passed
- `npm run typecheck` (web) — clean
- `node test-all.mjs` — **7289 passed, 0 failed**, 11 warnings (all pre-existing maintainer-name matches in `funding.json` / `HIRED.md` / `stats.go`, none from this diff)

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] If this is a new feature or architecture change, I opened an issue first (bug fix — exempt)
- [x] My changes respect the Data Contract (no user-layer files touched)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved verdict callout detection using section headings rather than fixed section letters.
  * Recognizes verdict headings across localized content and supported marker formats.
  * Prevents interview plan and rationale sections from being incorrectly highlighted as verdicts.
  * Reports without a verdict now display all sections below the detected content.

* **Tests**
  * Added coverage for localized headings, marker-based verdict titles, plain verdict titles, and false-positive cases.

### DIFF
--- a/web/src/components/report-view.tsx
+++ b/web/src/components/report-view.tsx
@@ -148,8 +148,8 @@ export function ReportView({
             //
             // Identified by the heading, never by the letter: reading "whatever is
             // lettered F" as the verdict rendered the Interview Plan table into a
-            // callout built for one sentence, in every mode and every language
-            // (#3416). No mode writes a Verdict block today, so today there is
+            // callout built for one sentence — in the canonical mode and in all
+            // eight localized `oferta.md` files alike (#3416). No mode writes a Verdict block today, so today there is
             // simply no callout and every block renders below.
             const verdict = sections.find((s) => isVerdictHeading(s.heading));
             const rest = sections.filter((s) => s !== verdict);
--- a/web/src/components/report-view.tsx
+++ b/web/src/components/report-view.tsx
@@ -5,7 +5,7 @@ import remarkGfm from "remark-gfm";
 import type { Application } from "@/lib/career-ops";
 import { Badge } from "@/components/ui/badge";
 import { scoreTone, scoreNum, legitimacyTone, parseReport } from "@/lib/format";
-import { cleanHeading, splitSections } from "@/lib/report-sections.mjs";
+import { cleanHeading, isVerdictHeading, splitSections } from "@/lib/report-sections.mjs";
 import { StatusSelect } from "@/components/status-select";
 import { CompanyLogo } from "@/components/company-logo";
 import { ScoreMethodology } from "@/components/score-methodology";
@@ -15,7 +15,7 @@ import { DeleteFromTracker } from "@/components/delete-from-tracker";
 import { companyPresentation } from "@/lib/company-presentation.mjs";
 
 // Progressive disclosure of the report. The core writes prose blocks
-// "## F) Verdict (lead)", "## A) Role Summary", "## B) Match with CV", then
+// "## A) Role Summary", "## B) Match with CV", then
 // the remaining lettered blocks + machine artifacts (Machine Summary YAML,
 // Application Answers, submit log). A mainstream user deciding "should I
 // apply?" needs the verdict + fit; the rest is depth-on-demand. We lead with
@@ -140,11 +140,18 @@ export function ReportView({
                 </article>
               );
             }
-            // Verdict (F) leads as a highlighted callout with no competing heading —
-            // it's THE answer. A/B stay expanded (fit detail); C–G collapse as
-            // content (with a 1-line preview); machine artifacts drop to a dimmer
-            // "Technical" tier so the CLI-DNA is present-but-clearly-secondary.
-            const verdict = sections.find((s) => s.letter === "F");
+            // A verdict block leads as a highlighted callout with no competing
+            // heading — it's THE answer. A/B stay expanded (fit detail); the rest
+            // collapse as content (with a 1-line preview); machine artifacts drop
+            // to a dimmer "Technical" tier so the CLI-DNA is present-but-clearly-
+            // secondary.
+            //
+            // Identified by the heading, never by the letter: reading "whatever is
+            // lettered F" as the verdict rendered the Interview Plan table into a
+            // callout built for one sentence, in every mode and every language
+            // (#3416). No mode writes a Verdict block today, so today there is
+            // simply no callout and every block renders below.
+            const verdict = sections.find((s) => isVerdictHeading(s.heading));
             const rest = sections.filter((s) => s !== verdict);
             const machine = rest.filter((s) => isMachine(s.heading));
             const mainSections = rest.filter((s) => !isMachine(s.heading));
--- a/web/src/lib/report-sections.mjs
+++ b/web/src/lib/report-sections.mjs
@@ -82,8 +82,9 @@ export function authorLetter(heading) {
  * lettered F" as the verdict, but no mode has ever written a Verdict block:
  * the callout landed in #1535 when `modes/oferta.md` already read
  * `## F) Interview Plan`, so every report has rendered the Interview Plan
- * table into a callout built for one sentence (#3416). All 19 localized modes
- * write Interview Plan at F too, so this was never an English-only slip.
+ * table into a callout built for one sentence (#3416). All eight localized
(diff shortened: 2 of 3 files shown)


santifer (changes requested):
This is the right fix, @maxmilian, and it just settled a six-PR pile-up on this file. One factual correction to make before I merge — details at the end, it is a two-minute edit.

**Why this one wins.** `report-view.tsx` currently has six open PRs on it, and two of them fix *this same bug* by different means. The other is #2323, which picks the lead block by matching heading names in a preference order: `Recommendation`, then `Verdict`, then `Risk Summary`. That works, until you look at what the modes actually write.

I checked all eight localized `oferta.md` files. Every one writes Interview Plan at F, in its own language — so your central claim holds and a letter-based rule was wrong everywhere at once:

```
modes/da/oferta.md:138     ## F) Samtaleplan
modes/es/oferta.md:138     ## F) Plan de entrevistas
modes/oferta.md:613        ## F) Interview Plan
modes/pl/oferta.md:138     ## F) Plan rozmów kwalifikacyjnych
modes/pt/oferta.md:138     ## F) Plano de Entrevistas
modes/ru/oferta.md:137     ## F) План собеседований
modes/ua/oferta.md:153     ## F) План співбесід
modes/zh-TW/oferta.md:265  ## F) 面試準備計畫
modes/zh/oferta.md:249     ## F) 面试备考计划
```

But the same check is what decides between the two fixes. A name-matching rule inherits whatever each translation happened to leave in English: `zh` and `zh-TW` keep `## Risk Summary` untranslated, so those reports would get a callout, while `es`, `da`, `pl`, `pt`, `ru` and `ua` don't write a Risk Summary block at all and would get none. That is a selector whose behaviour depends on translation accidents rather than on anything the core promises.

Your marker doesn't have that problem, because `(lead)` is not a word — `cleanHeading` has stripped it since before either PR, and `## F) Veredicto (lead)` reads exactly like `## F) Verdict (lead)`. So: **the marker is the mechanism, and the remaining question — which block should carry it — belongs in the modes, not in the viewer.** I am taking that to the core side rather than asking either of you to guess it in TypeScript.

And I want to be explicit that I am merging a change whose visible effect today is *no callout at all*, because your reasoning for that is correct and worth quoting: a confident callout labelled Verdict over content that is not the verdict is worse than no callout. Removing it is the fix, not a side effect of it.

## The one change

"All 19 localized modes" is not right, three times over (the doc comment in `report-sections.mjs`, the comment in `report-view.tsx`, and the test comment). There are **eight** localized `oferta.md` files, listed above. The 19 is presumably the locale directory count — there are 18 of those, and ten of them (`ar`, `de`, `fr`, `hi`, `id`, `it`, `ja`, `ko`, `nl`, `tr`) ship no `oferta.md` at all and fall back to the canonical English mode, so they never had an F block of their own to get wrong.

Wording that keeps your point intact:

> All eight localized `oferta.md` files write Interview Plan at F too, so this was never an English-only slip.

Push that and I merge — nothing else is outstanding, and the behaviour is not in question.

santifer (approved):
Approving: the correction landed in all three places, and the count in it is now mine to fix rather than yours — see the comment above. The verdict is unchanged: `(lead)` is a marker, not a word, so it survives translation where a heading-name selector does not.

Verified before approving: all 19 mode files carrying `## F)` (18 localized + canonical) write the Interview Plan there, so removing the letter-based verdict callout is correct in every locale, not just English.

PR #3503, opened 2026-08-29
fix(data-root): resolve templates/states.yml from the codebase, not the data root (#3500)

### LINKED ISSUE #3500: bug(data-root): normalize-statuses resolves templates/states.yml through the DATA_ROOT, silently normalizing every status to unknown
## Summary

`normalize-statuses.mjs` resolves `templates/states.yml` — a **System Layer** file — through the *data* root. Setting a data directory therefore makes the canonical states unreadable, and because the load is wrapped in a `catch` that falls back to an empty list, every status silently becomes `unknown` instead of raising.

`DATA_CONTRACT.md` is explicit that this should not happen:

> System Layer files continue to be resolved relative to the codebase/repository root, keeping code and personal data completely separate.

This is the one thing that stopped me migrating a symlinked user layer onto the supported `.career-ops-data` mechanism — the migration otherwise worked.

## Repro

```bash
echo "/some/data/dir" > .career-ops-data
node -e "import('./normalize-statuses.mjs').then(m =>
  console.log(m.normalizeStatus('Accepted')))"
```

| | `normalizeStatus('Accepted')` | `('Contratado')` | `('Evaluada')` |
|---|---|---|---|
| no marker | `{status: "Hired"}` | `{status: "Hired"}` | `{status: "Evaluated"}` |
| marker set | `{status: null, unknown: true}` | `{status: null, unknown: true}` | `{status: null, unknown: true}` |

No error, no warning. Same binary, same tracker — only the data root differs.

## Cause

`normalize-statuses.mjs:25-38`:

```js
const CAREER_OPS = getCareerOpsRoot();
const APPS_FILE = resolveTrackerPath(CAREER_OPS);   // correct — tracker is User Layer
...
    statesCache = loadCanonicalStates(join(CAREER_OPS, 'templates', 'states.yml'));  // wrong — System Layer
  } catch {
    statesCache = []; // broken install: fall through to "unknown", never a stale copy
  }
```

One variable serves both roles. `resolveTrackerPath(CAREER_OPS)` is right — the tracker *is* user data. The `templates/states.yml` join on the next lines is not, and there is nothing to catch it because a data root that legitimately has no `templates/` is indistinguishable from the "broken install" the `catch` was written for.

That `catch` is what turns a path bug into a silent one. Its reasoning is sound in isolation — preferring `unknown` over a stale copy — but it was written for a missing file in a broken checkout, not a file being looked up in the wrong tree entirely.

Worth noting the fix is narrow: `tracker-sync-check.mjs:105` joins the same `templates/states.yml` onto `dirname(fileURLToPath(import.meta.url))` and is correct. So is `check-table-freshness.mjs`, and so are the `modes/`/`templates/` lookups in `openai-eval.mjs`, `openai-tailor.mjs`, `ollama-eval.mjs` and `batch-evaluate-gemini.mjs` — all of them bind `ROOT` to `dirname()`. I checked every `join(ROOT, '<system-dir>')` call site I could find; `normalize-statuses.mjs` is the only one bound to `getCareerOpsRoot()`.

## Suggested fix

Resolve the system file from the module's own location, leaving the data root for user data:

```js
const CODEBASE_ROOT = dirname(fileURLToPath(import.meta.url));
const CAREER_OPS = getCareerOpsRoot();
const APPS_FILE = resolveTrackerPath(CAREER_OPS);
...
statesCache = loadCanonicalStates(join(CODEBASE_ROOT, 'templates', 'states.yml'));
```

That matches what `tracker-sync-check.mjs` already does for the same file.

Two things that might be worth considering beyond the one-line fix, though they're your call:

- The `catch` could distinguish "file absent" from "file unparseable". Silently returning `[]` when `states.yml` is missing means any future path regression here degrades invisibly again.
- A test asserting `normalizeStatus('Accepted') === 'Hired'` **with a data root configured** would pin the layer boundary. The current suite passes because it runs with the default root, so this class of leak is invisible to CI.

## Impact

Any install using `CAREER_OPS_ROOT` / `CAREER_OPS_DATA_DIR` / `.career-ops-data`. Status normalization is not cosmetic — `normalize-statuses.mjs` is the documented repair path for the tracker, and under a data root it would rewrite every recognized status to unknown rather than canonicalizing it. Localized trackers lose the most: all 156 non-English spellings stop resolving, so a Spanish or Turkish tracker normalizes to nothing at all.

On a full `test-all.mjs` run in my checkout, configuring a data root took the suite from 4 failures to 19, and ~15 of those trace to this single line.

## Environment

- career-ops v1.30.0
- Data root configured via a `.career-ops-data` marker

vliggio:
## What does this PR do?

`normalize-statuses.mjs` resolved `templates/states.yml` — a System Layer file — through the *data* root, so any install using `CAREER_OPS_ROOT` / `CAREER_OPS_DATA_DIR` / `.career-ops-data` silently normalized every status to unknown. This resolves it from the codebase instead, fixes the same defect in `followup-cadence.mjs`, makes both fallbacks report themselves on stderr, and adds a suite that pins the layer boundary.

## Related issue

Closes #3500

## Type of change

- [x] Bug fix
- [ ] New feature

## Detail

**The bug.** One variable served both roles. `resolveTrackerPath(CAREER_OPS)` is right — the tracker *is* user data — but the next lines joined `templates/states.yml` onto the same root, where it does not exist. The surrounding `catch` then turned a path bug into a silent one: its reasoning is sound in isolation (prefer `unknown` over a stale hardcoded copy), but it was written for a missing file in a broken checkout, and a data root that legitimately has no `templates/` is indistinguishable from that.

| | `normalizeStatus('Accepted')` | `('Contratado')` | `('Evaluada')` |
|---|---|---|---|
| no marker | `{status: "Hired"}` | `{status: "Hired"}` | `{status: "Evaluated"}` |
| marker set (before) | `{status: null, unknown: true}` | `{status: null, unknown: true}` | `{status: null, unknown: true}` |
| marker set (after) | `{status: "Hired"}` | `{status: "Hired"}` | `{status: "Evaluated"}` |

This is not cosmetic: `normalize-statuses.mjs` is the documented tracker repair path, so under a data root it rewrote recognized statuses to unknown rather than canonicalizing them. Localized trackers lost the most — all 156 non-English spellings stopped resolving.

**A second site.** `followup-cadence.mjs` carried the identical join, inline in `statusAliasMap()`, where an empty alias map drops every localized row out of the `ACTIONABLE`/`ADVANCED` sets. It escaped the audit in #3500 because it has no hoisted `STATES_FILE` constant to grep for.

`set-status.mjs`, `funnel-velocity.mjs`, `tracker-sync-check.mjs` and `check-table-freshness.mjs` were already correct — they bind their root to `dirname(fileURLToPath(import.meta.url))`.

**Both catches now report to stderr** before degrading. Fallback behavior is unchanged (still an empty list, never a resurrected hardcoded table); what changed is that it says so. Degrading silently is what made this invisible.

## Test

`tests/system-layer-under-data-root.test.mjs` (auto-discovered), five assertions:

1. `normalize-statuses` canonicalizes `Accepted` / `Contratado` / `Evaluada` / `Rechazado` under `CAREER_OPS_ROOT` **and** `CAREER_OPS_DATA_DIR` — the table above, run with a data root configured.
2. `followup-cadence` resolves Spanish aliases to `hired` / `applied` / `interview` under a data root.
3. A source-level guard against the whole class: no root-level `.mjs` may join a `getCareerOpsRoot()`-bound variable onto `templates/` or `modes/`. `modes/_profile.md`, `_custom.md` and `_brief.md` are whitelisted — those specific files *are* user data that happens to live under `modes/`, which is why `gemini-eval.mjs` and `ollama-eval.mjs` legitimately resolve them through the data root.
4. The `catch` in `normalize-statuses` emits a diagnostic rather than a bare `[]`.

Mutation-checked: reverting either join fails both the behavioral assertion and the source guard.

The existing suite could not see any of this, because it runs with the default root — where the data root and the codebase root are the same directory.

## Checklist

- [x] I ran `node test-all.mjs` and all tests pass (7294 passed, 0 failed, 12 pre-existing warnings)
- [x] My changes respect the [Data Contract](https://github.com/santifer/career-ops/blob/main/DATA_CONTRACT.md) (no modifications to user-layer files) — restoring its System/User Layer split is the point of the change

## Noticed while verifying, not fixed here

Running the full suite creates a stray `applications.db` in the repo root. `tests/tracker-busy-timeout.test.mjs` sets `CAREER_OPS_TRACKER_DB` before importing `tracker.mjs`, but by then an inline section of `test-all.mjs` (`removeRowByNum`, ~line 5035) has already imported that module into the same process, so `DB_PATH` was resolved at first import and the pin is a no-op. `openDb()` then creates the schema at the unpinned path. It passes either way, so nothing flags it. Happy to send that separately if you'd like it filed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected status and follow-up processing to consistently use the application’s built-in configuration.
  * Improved error reporting when status definitions cannot be loaded, making fallback behavior clear.

* **Tests**
  * Added coverage to verify configuration resolution, localized status handling, and diagnostics for unreadable status definitions.

### DIFF
--- a/followup-cadence.mjs
+++ b/followup-cadence.mjs
@@ -22,6 +22,9 @@ import { localToday } from './lib/local-today.mjs';
 import { flagValue, validateFlags } from './lib/cli-flags.mjs';
 import { isMainModule } from './lib/is-main-module.mjs';
 
+// templates/states.yml is System Layer — resolved from the codebase, not from
+// the user's data root (#3500).
+const CODEBASE_ROOT = dirname(fileURLToPath(import.meta.url));
 const CAREER_OPS = getCareerOpsRoot();
 const APPS_FILE = resolveTrackerPath(CAREER_OPS);
 
@@ -120,16 +123,19 @@ function statusAliasMap() {
   if (aliasMapCache) return aliasMapCache;
   const map = new Map();
   try {
-    for (const st of loadCanonicalStates(join(CAREER_OPS, 'templates', 'states.yml'))) {
+    for (const st of loadCanonicalStates(join(CODEBASE_ROOT, 'templates', 'states.yml'))) {
       const id = st.id.toLowerCase();
       map.set(foldStatusInput(id), id);
       if (st.label) map.set(foldStatusInput(st.label), id);
       for (const a of st.aliases) map.set(foldStatusInput(a), id);
     }
-  } catch {
+  } catch (err) {
     // A missing/malformed states.yml is a broken install. Degrade to
     // identity-normalization rather than resurrecting a hardcoded table: a
     // fallback copy is the same copy in disguise and drifts the same way.
+    // Report it, though — degrading silently is what hid #3500, and here it
+    // costs every localized status its place in the funnel.
+    console.error(`[followup-cadence] cannot read canonical states from templates/states.yml: ${err.message}`);
     return (aliasMapCache = new Map());
   }
   return (aliasMapCache = map);
--- a/normalize-statuses.mjs
+++ b/normalize-statuses.mjs
@@ -22,6 +22,12 @@ import {
 import { resolveColumns, parseTrackerRow } from './tracker-parse.mjs';
 import { isMainModule } from './lib/is-main-module.mjs';
 
+// System Layer files (templates/, modes/, scripts) live in the codebase and are
+// resolved from this module's own directory; only User Layer data follows the
+// configurable data root. Binding both to getCareerOpsRoot() made states.yml
+// unreadable under CAREER_OPS_ROOT / .career-ops-data, and the catch below
+// turned that into a silent "everything is unknown" (#3500).
+const CODEBASE_ROOT = dirname(fileURLToPath(import.meta.url));
 const CAREER_OPS = getCareerOpsRoot();
 const APPS_FILE = resolveTrackerPath(CAREER_OPS);
 const DRY_RUN = process.argv.includes('--dry-run');
@@ -34,10 +40,17 @@ let statesCache = null;
 /** Canonical states from templates/states.yml, read once per CLI run. */
 function canonicalStates() {
   if (statesCache) return statesCache;
+  const statesFile = join(CODEBASE_ROOT, 'templates', 'states.yml');
   try {
-    statesCache = loadCanonicalStates(join(CAREER_OPS, 'templates', 'states.yml'));
-  } catch {
-    statesCache = []; // broken install: fall through to "unknown", never a stale copy
+    statesCache = loadCanonicalStates(statesFile);
+  } catch (err) {
+    // Broken install: fall through to "unknown", never a stale copy — but say
+    // so. The silent [] is what made #3500 invisible: a states.yml looked up in
+    // the wrong tree is indistinguishable from one that is genuinely missing,
+    // and the whole tracker normalized to unknown without a word.
+    console.error(`[normalize-statuses] cannot read canonical states from ${statesFile}: ${err.message}`);
+    console.error('[normalize-statuses] every status will be reported as unknown until this is fixed.');
+    statesCache = [];
   }
   return statesCache;
 }
--- a/tests/system-layer-under-data-root.test.mjs
+++ b/tests/system-layer-under-data-root.test.mjs
@@ -1,0 +1,99 @@
+// tests/system-layer-under-data-root.test.mjs — System Layer files must resolve
+// from the codebase, never from the configurable data root (#3500).
+//
+// DATA_CONTRACT.md: "System Layer files continue to be resolved relative to the
+// codebase/repository root, keeping code and personal data completely separate."
+//
+// normalize-statuses.mjs broke that for templates/states.yml: one variable served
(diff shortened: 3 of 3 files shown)


Scott-Emberson (commented):
Reviewed at head ff9c64db7b. `system-layer-under-data-root.test.mjs` configures a real separate data root (a temp dir with only `data/`, no `templates/`) via CAREER_OPS_ROOT and CAREER_OPS_DATA_DIR, both of which genuinely divert the data root, and asserts states.yml still resolves from the codebase. A regression to joining `templates/states.yml` off the data root reddens because the temp root has no `templates/` (normalize-statuses and the followup-cadence alias map both come back empty and mismatch). There's also a static guard that flags a literal `join(<datavar>, 'templates'|'modes', ...)` revert, plus a check that `canonicalStates()` now logs on an unreadable states.yml (the silent-`[]` that hid #3500). Two independent checks fail on regression.

Scott-Emberson (approved):

Issue #2271, opened 2026-07-29
scan.mjs: portals/profile paths are env-overridable but pipeline.md and scan-history.tsv are hardcoded, so parallel search lanes collide in one inbox


ringo380:
## Use case

A single person often runs **two parallel searches** with different targeting: the primary career lane, and a second lane with a different rubric — a bridge/immediate-income search, a career-change track, a contract-vs-FTE split, or a partner's search sharing the same checkout.

`scan.mjs` already supports most of this. Both inputs are overridable:

```js
const PORTALS_PATH = process.env.CAREER_OPS_PORTALS || 'portals.yml';   // scan.mjs:63
const PROFILE_PATH = process.env.CAREER_OPS_PROFILE || 'config/profile.yml'; // :64
```

So a second lane can define its own `portals.<lane>.yml` with a different `title_filter` and its own `search_queries`, and scan with:

```bash
CAREER_OPS_PORTALS=portals.bridge.yml node scan.mjs
```

That part works well. The problem is the **outputs**:

```js
const SCAN_HISTORY_PATH = 'data/scan-history.tsv';   // scan.mjs:65
const PIPELINE_PATH     = 'data/pipeline.md';        // scan.mjs:66
```

Both are hardcoded. Every lane writes into the same inbox and the same dedup history.

## Consequences

1. **The lanes become indistinguishable in the inbox.** A pending row is `- [ ] {url} | {company} | {title} | {location}` with no field identifying which config produced it. After one bridge scan, entry-level support roles sit interleaved with staff/architect roles and nothing marks the difference — so the reader (or the `pipeline` mode) cannot tell which rubric to evaluate a given row against. Since the two lanes score on genuinely different criteria, scoring a row under the wrong one produces a confidently wrong number.
2. **Shared dedup silently suppresses cross-lane hits.** `scan-history.tsv` is the dedup source. A posting seen once in lane A is `Duplicates: skipped` in lane B — even when it is a legitimate result for both, and even though lane B has never surfaced it to the user. This one is quiet: the counter increments and no row appears.
3. **Stats and analytics conflate the lanes.** `stats.mjs`, `analyze-patterns.mjs`, and `detect-reposts.mjs` all read these files. A bridge lane with different volume and a different advance rate skews the career funnel it is pooled with.

## Suggested shape

Make the outputs overridable the same way the inputs already are — smallest change, consistent with existing convention:

```js
[...]

Issue #2283, opened 2026-07-29
a resolver-outage stop deletes its own checkpoint, so --resume can't continue


bigguy6883:
# Issue: a resolver-outage stop deletes its own checkpoint, so `--resume` can't continue

**Type:** Bug
**Component:** `scan-ats-full.mjs`

## Summary

When the resolver breaker added in #2266 stops a sweep, the run deletes its
checkpoint on the way out. `--resume` therefore cannot continue the interrupted
sweep — the exact recovery the breaker exists to enable.

## What happens

`scan-ats-full.mjs` on `main`:

1. `:779` — 50 consecutive resolver failures set `resolverOutage = true`.
2. `:829-836` — the source loop logs the stop and `break`s. **No final
   `writeCheckpoint`.** The newest checkpoint on disk is whatever the periodic
   writer last produced at `:789` (`CHECKPOINT_EVERY = 500` companies).
3. `:933` — on the way out, unconditionally:
   ```js
   if (!opts.dryRun && existsSync(CHECKPOINT_PATH)) unlinkSync(CHECKPOINT_PATH);
   ```
   That line is meant to clear the checkpoint after a *successful* sweep. It
   does not distinguish an interrupted one, so the resume point is deleted.

Net effect: the operator fixes their resolver, runs `--resume`, and the sweep
starts over from zero.

## Why this looks unintended

The outage branch already goes out of its way to preserve resumability — the
comment at `:830` reads:

> Deliberately before completedSources/checkpoint: this source did NOT finish,
> and marking it done would make `--resume` skip the rest of it.

So the intent is clearly that an outage-stopped sweep stays resumable. `:933`
then removes the file that would have made it so.

## Impact

Worst on the runs that cost the most. A full Workday sweep is 12,884 tenants
and takes hours; iCIMS adds ~10,108. A resolver blip at 80% means restarting
from 0% rather than from the last 500-company checkpoint.

Not a data-loss bug — nothing already written to `pipeline.md` is affected —
but it removes the recovery path that #2138 (checkpoint/`--resume`) and #2266
(the breaker) were built to provide together.

## Suggested fix

Two small changes:

1. Write a final checkpoint on the outage path before `break`, carrying the
   latest `{ done, resumeAt }` observed by the progress callback at `:785`, so
   the resume point is current rather than up to 499 companies stale.
2. Make the `:933` cleanup conditional on the run having finished — e.g. an
   `interrupted` flag set on the outage break — so an interrupted sweep keeps
[...]

PR #3511, opened 2026-08-29
fix(scan): resolve scanner user data against the data root, not the cwd (#3510)

### LINKED ISSUE #3510: bug(scan): blacklist, scan-runs, portal-health and the ATS checkpoint resolve against the cwd, not the data root
## Summary

The scanners resolve part of their **User Layer** data against `process.cwd()` while resolving the rest against `DATA_ROOT`. One command therefore writes one dataset to two places, and under a configured data root the user's own files are ignored in favour of whatever happens to sit in the current directory.

Unlike #3500 / #3508 nothing fails loudly here — a wrong-but-writable path is silently accepted.

## Repro

Two directories: a real data root, and a "decoy" that is merely the shell's cwd.

```bash
mkdir -p /tmp/root4/data /tmp/decoy/data
printf '| Company | Since | Scope | Reason |\n|---|---|---|---|\n| RealCo | 2026-01-01 | all | real |\n'   > /tmp/root4/data/blacklist.md
printf '| Company | Since | Scope | Reason |\n|---|---|---|---|\n| DecoyCo | 2026-01-01 | all | decoy |\n' > /tmp/decoy/data/blacklist.md

cd /tmp/decoy
CAREER_OPS_ROOT=/tmp/root4 node /path/to/career-ops/scan.mjs   # or import loadBlacklist directly
```

Observed:

| call | reads/writes | should be |
|---|---|---|
| `loadBlacklist()` | `/tmp/decoy/data/blacklist.md` (`decoyco`) | `/tmp/root4/data/blacklist.md` (`realco`) |
| `appendScanRunSummary(...)` | `/tmp/decoy/data/scan-runs.tsv` | `/tmp/root4/data/scan-runs.tsv` |
| `loadCheckpoint()` (scan-ats-full) | `/tmp/decoy/data/cache/ats-full-checkpoint.json` | the data root's |

Meanwhile the *same run* writes `data/pipeline.md` and `data/scan-history.tsv` to `/tmp/root4`, because those constants are anchored.

**The blacklist case is the one that matters.** `AGENTS.md` documents `data/blacklist.md` as respected by `scan.mjs`; a user with a data root gets an empty blacklist and their do-not-apply companies come back in the results, with no indication anything was skipped.

The scan-runs case is a writer/reader split: `scan.mjs` writes `data/scan-runs.tsv` relative to cwd, `stats.mjs:36` reads `join(DATA_ROOT, 'data', 'scan-runs.tsv')`. Same for `data/portal-health.tsv` (`stats.mjs:39`).

The checkpoint case corrupts `--resume`: a multi-hour sweep resumed from a different directory finds no checkpoint and restarts from zero, or resumes from an unrelated one.

## Cause

Bare relative strings, resolved against the cwd:

| file | line | constant |
|---|---|---|
| `scan.mjs` | 1988 | `BLACKLIST_PATH = 'data/blacklist.md'` |
| `scan.mjs` | 2037 | `SCAN_RUNS_PATH = 'data/scan-runs.tsv'` |
| `scan.mjs` | 2110 | `PORTAL_HEALTH_PATH = 'data/portal-health.tsv'` |
| `scan-ats-full.mjs` | 60, 61, 62, 95 | `PORTALS_PATH` fallback, `PIPELINE_PATH`, `CACHE_DIR`, `CHECKPOINT_PATH` |
| `scan-interamt.mjs` | 28, 29, 207, 258 | `PORTALS_PATH`, `SCAN_HISTORY`, `mkdirSync('output')`, `mkdirSync('data')` |
| `scan-hn.mjs` | 27 | `PORTALS_PATH` |

`scan.mjs:75-92` shows the established pattern the rest of the file already follows:

```js
const DATA_ROOT = getCareerOpsRoot();
const PORTALS_PATH = process.env.CAREER_OPS_PORTALS || path.join(DATA_ROOT, 'portals.yml');
const SCAN_HISTORY_PATH = process.env.CAREER_OPS_SCAN_HISTORY || path.join(DATA_ROOT, 'data/scan-history.tsv');
const PIPELINE_PATH = process.env.CAREER_OPS_PIPELINE || path.join(DATA_ROOT, 'data/pipeline.md');
```

The sibling scanners are worse off than a plain inconsistency: they `import { appendToPipeline, appendToScanHistory } from './scan.mjs'`, so they *append* through `scan.mjs`'s anchored paths while checking existence, creating, and reading through their own cwd-relative ones. `scan-ats-full.mjs:1017` can create an empty `data/pipeline.md` in the cwd and then append the actual results somewhere else entirely. They also silently ignore `CAREER_OPS_PORTALS` / `CAREER_OPS_SCAN_HISTORY`, which #2271 added precisely so a second search lane does not poison the first one's dedup history.

## The part that needs a decision

`PORTAL_HEALTH_PATH` is cwd-relative **on purpose**. Commit `96c578b` (2026-07-30, "resolve PORTAL_HEALTH_PATH from cwd") changed it from `dirname(fileURLToPath(import.meta.url))` — the script's own directory — and `tests/portal-health-path.test.mjs` locks that in, asserting the script directory is "left completely alone".

That fix was solving a real problem: a sandboxed run wrote fixture rows into the live `data/portal-health.tsv` of whatever checkout owns `scan.mjs`. But it swapped one unanchored path for another. `getCareerOpsRoot()` falls back to the codebase root when nothing is configured, so "anchor to DATA_ROOT" and "anchor to the script dir" are the same directory on a default install — which is why simply making it consistent with its siblings would trip that test.

Note the test's own premise has gone stale: it says "every sibling data path in scan.mjs (SCAN_HISTORY_PATH, PIPELINE_PATH, APPLICATIONS_PATH, BLACKLIST_PATH, SCAN_RUNS_PATH) is a bare cwd-relative string", but `02daaf1` (2026-07-02) had already anchored the first three to `DATA_ROOT`. Only the last two still match that description.

Two ways forward:

**A — anchor everything to `DATA_ROOT`, isolate tests with `CAREER_OPS_ROOT`.** One rule for the whole file, matching the pattern its siblings have used since the data-root feature landed, and matching every reader (`stats.mjs`). `tests/portal-health-path.test.mjs` keeps its guarantee — a test run must never touch the checkout's live data — but enforces it by pointing `CAREER_OPS_ROOT` at a temp dir, the mechanism the rest of the suite already uses, instead of by relying on cwd.

**B — explicit-root-wins, cwd fallback.** Use `DATA_ROOT` only when a root is *explicitly* configured (env var or `.career-ops-data`), else stay cwd-relative. Preserves today's behavior exactly for unconfigured installs, at the cost of a third resolution mode and a new `path-resolver` helper to distinguish "configured" from "defaulted".

I'd go with **A**: B leaves `scan.mjs` with two path rules and makes the answer to "where does this file live?" depend on how the install is configured. But A reverses the *mechanism* of a deliberate four-week-old fix (not its intent), so it is your call.

## Impact

Anyone using `CAREER_OPS_ROOT` / `CAREER_OPS_DATA_DIR` / `.career-ops-data`, and anyone invoking the scanners from outside the checkout (cron, launchd, a wrapper, an editor task). Concretely: a silently empty blacklist, `stats.mjs` under-reporting scan totals it cannot find, `--resume` restarting a multi-hour sweep from zero, and `data/` + `output/` directories appearing in whatever directory the user happened to be in.

## Environment

- career-ops v1.30.0
- Node.js v26.5.0 on macOS (darwin 25.6.0)
- Found while fixing #3500 / #3506 / #3508

vliggio:
## What does this PR do?

The scanners resolved part of their User Layer data against `process.cwd()` and the rest against `DATA_ROOT`, so one command wrote one dataset to two places. This anchors all of it to the data root, gives the sibling scanners one shared resolution to import instead of their own copies, and adds a suite that separates the data root from the cwd so the difference is actually observable.

## Related issue

Closes #3510 — which has the full analysis, including the one decision I'd like a second opinion on (below).

## Type of change

- [x] Bug fix
- [ ] New feature

## Detail

**What broke.** Nothing failed loudly — a wrong-but-writable path is silently accepted, which is why this outlived #3500 and #3508.

| call | read/wrote | should have |
|---|---|---|
| `loadBlacklist()` | the cwd's `data/blacklist.md` | the data root's |
| `appendScanRunSummary()` | the cwd's `data/scan-runs.tsv` | the data root's, where `stats.mjs:36` reads |
| `appendPortalHealth()` | the cwd's `data/portal-health.tsv` | the data root's, where `stats.mjs:39` reads |
| `loadCheckpoint()` (scan-ats-full) | the cwd's checkpoint | the data root's |

The blacklist is the one that matters: `AGENTS.md` documents `data/blacklist.md` as respected by `scan.mjs`, so a user with a data root got an empty list and their do-not-apply companies came back in the results, with nothing to indicate anything had been let through.

**The siblings were worse than inconsistent.** They import `appendToPipeline` / `appendToScanHistory` from `scan.mjs`, so they *appended* through anchored paths while checking existence, creating and reading through their own cwd-relative copies. `scan-ats-full.mjs:1017` could create an empty `data/pipeline.md` in the cwd and then write the actual matches somewhere else; `scan-interamt.mjs` read one scan-history to compute its last-scan date and appended to another. Both silently ignored `CAREER_OPS_PORTALS` / `CAREER_OPS_SCAN_HISTORY`, which #2271 added precisely so a second search lane cannot poison the first one's dedup history. `scan.mjs` now exports `PORTALS_PATH`, `PIPELINE_PATH` and `SCAN_HISTORY_PATH`; the siblings import them.

## The part worth a second opinion

`PORTAL_HEALTH_PATH` was cwd-relative **deliberately**. `96c578b` changed it from `dirname(fileURLToPath(import.meta.url))` because sandboxed runs were writing fixture rows into the live `data/portal-health.tsv` of whatever checkout owned `scan.mjs`, and `tests/portal-health-path.test.mjs` locked that in.

That problem was real. But the fix swapped one unanchored path for another and left the file with two rules for where user data lives — and `getCareerOpsRoot()` falls back to the codebase root, so "anchor to DATA_ROOT" and "anchor to the script dir" are the same directory on a default install, which is why making it consistent trips that test as written.

I kept the guarantee and changed the mechanism: the test still asserts the checkout's own data directory is never touched, but isolates with `CAREER_OPS_ROOT` — what the rest of the suite already uses — instead of relying on cwd. It now pins the data root and the cwd to **different** temp dirs and asserts the row lands under the root and *not* under the cwd, which the original could not distinguish.

Worth noting that test's premise had gone stale: it says "every sibling data path in scan.mjs (SCAN_HISTORY_PATH, PIPELINE_PATH, APPLICATIONS_PATH, BLACKLIST_PATH, SCAN_RUNS_PATH) is a bare cwd-relative string", but `02daaf1` had already anchored the first three.

#3510 lays out the alternative (use `DATA_ROOT` only when a root is *explicitly* configured, else stay cwd-relative). It preserves today's behavior exactly for unconfigured installs at the cost of a third resolution mode. I went with consistency; say the word if you prefer the other and I'll rework it.

## One extra change, deliberately included

`scan-interamt.mjs` calls `main()` at module scope — no `isMainModule` guard, unlike `scan-hn.mjs:122` and `scan-ats-full.mjs:1115`. Importing the module therefore drove a **live browser scan of interamt.de** and appended 118 offers to the pipeline. I found this the hard way, by importing it to check my own change; the artifacts were cleaned up.

It is in scope because the file could not otherwise be imported by a test at all, and because it is the same family of accident: a script doing real work somewhere nobody asked it to. Happy to split it out if you'd rather.

## Test

`tests/scan-data-paths-under-data-root.test.mjs` runs each module in a child process with the data root and the cwd pointed at **different** directories, each holding a decoy copy of the file under test — so a path that follows the shell picks the decoy and is caught. Six assertions: blacklist, scan-runs, ATS checkpoint, the exported shared paths, the interamt main guard, and a source guard against a new bare-relative user-data constant.

All six fail against the pre-fix code:

```
❌ loadBlacklist() read the wrong file — saw [decoyco], expected [realco]
❌ scan-runs.tsv landed wrong — root:false cwd:true
❌ loadCheckpoint() read the cwd checkpoint — a resume would restart or resume the wrong sweep
❌ scan.mjs does not export pipeline/portals paths under the data root: {}
❌ scan-interamt.mjs calls main() at module scope — importing it starts a live scan
❌ cwd-relative user-data path(s) reintroduced: ...9 constants...
```

`tests/scan-ats-full-outage-checkpoint.test.mjs` also needed updating: it built a sandbox **cwd** holding `portals.yml` and the dataset cache. That layout is already exactly a data root, so it now hands the directory to the child as `CAREER_OPS_ROOT` and keeps the cwd pinned there too.

## Checklist

- [x] I ran `node test-all.mjs` and all tests pass (7296 passed, 0 failed, 12 pre-existing warnings)
- [x] My changes respect the [Data Contract](https://github.com/santifer/career-ops/blob/main/DATA_CONTRACT.md) (no modifications to user-layer files) — restoring the User Layer to the configured data root is the point of the change

Independent of #3503, #3507 and #3509 — different files, branched from `main`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Scanner configuration, history, pipeline results, health reports, and checkpoints now consistently use the configured data location.
  * Resuming a paused ATS scan from another directory now correctly finds its checkpoint instead of restarting from the beginning.
  * Prevented scan results and related files from being accidentally created in the current working directory.
  * Improved consistency across supported portal scanners and safer module imports.

* **Tests**
  * Added regression coverage for data-path handling, checkpoint recovery, and scanner isolation.

### DIFF
--- a/scan-ats-full.mjs
+++ b/scan-ats-full.mjs
@@ -48,18 +48,24 @@ import lever from './providers/lever.mjs';
 import ashby from './providers/ashby.mjs';
 import workday from './providers/workday.mjs';
 import icims from './providers/icims.mjs';
-import { buildTitleFilter, buildLocationFilter, buildContentFilter, matchedTitleKeywords, loadSeenUrls, normalizeUrlForDedup, appendToPipeline, appendToScanHistory, loadBlacklist, parseSinceDays } from './scan.mjs';
+import { buildTitleFilter, buildLocationFilter, buildContentFilter, matchedTitleKeywords, loadSeenUrls, normalizeUrlForDedup, appendToPipeline, appendToScanHistory, loadBlacklist, parseSinceDays, PORTALS_PATH, PIPELINE_PATH } from './scan.mjs';
 import { localToday } from './lib/local-today.mjs';
 import { SEED_SOURCES, toPortalEntry } from './seeds/vc-portfolios.mjs';
 import { normalizeCompany } from './tracker-utils.mjs';
 import { validateFlags } from './lib/cli-flags.mjs';
 import { isMainModule } from './lib/is-main-module.mjs';
+import { getCareerOpsRoot } from './path-resolver.mjs';
 
 // ── Config ──────────────────────────────────────────────────────────
 
-const PORTALS_PATH = process.env.CAREER_OPS_PORTALS || 'portals.yml';
-const PIPELINE_PATH = 'data/pipeline.md';
-const CACHE_DIR = 'data/cache/ats-companies';
+// PORTALS_PATH and PIPELINE_PATH are imported from scan.mjs rather than
+// re-derived here (#3510). This file appends results through scan.mjs's
+// appendToPipeline, so its own bare-relative copy meant it could create an empty
+// data/pipeline.md in the cwd and then write the actual matches somewhere else.
+// Its portals fallback had the same split: it honored CAREER_OPS_PORTALS but
+// otherwise looked in the cwd instead of the data root.
+const DATA_ROOT = getCareerOpsRoot();
+const CACHE_DIR = path.join(DATA_ROOT, 'data/cache/ats-companies');
 const CACHE_TTL_HOURS = 24;
 // Tracks `main` deliberately: the dataset's value is freshness (new boards
 // appear weekly), so pinning a commit would defeat the purpose. Integrity rests
@@ -92,7 +98,10 @@ const RESOLVER_FAILURE_LIMIT = 50;
 // Crash insurance for multi-hour directory sweeps: progress + matches are
 // checkpointed every CHECKPOINT_EVERY companies so --resume can continue a
 // dead run (with its ORIGINAL date window) instead of restarting from zero.
-const CHECKPOINT_PATH = 'data/cache/ats-full-checkpoint.json';
+// Anchored too (#3510): a sweep resumed from a different directory found no
+// checkpoint and silently restarted a multi-hour run from zero — the one failure
+// mode --resume exists to prevent.
+const CHECKPOINT_PATH = path.join(DATA_ROOT, 'data/cache/ats-full-checkpoint.json');
 const CHECKPOINT_EVERY = 500;
 
 export function loadCheckpoint(file = CHECKPOINT_PATH) {
--- a/scan-hn.mjs
+++ b/scan-hn.mjs
@@ -16,7 +16,7 @@ try {
 import { readFileSync, existsSync } from 'fs';
 import * as yaml from 'js-yaml';
 import { GoogleGenerativeAI } from '@google/generative-ai';
-import { appendToPipeline, appendToScanHistory, loadSeenUrls } from './scan.mjs';
+import { appendToPipeline, appendToScanHistory, loadSeenUrls, PORTALS_PATH } from './scan.mjs';
 import { localToday } from './lib/local-today.mjs';
 
 // Import the deterministic provider
@@ -24,7 +24,7 @@ import hnProvider from './providers/hackernews.mjs';
 import { isMainModule } from './lib/is-main-module.mjs';
 
 // ── Configuration ────────────────────────────────────────────────────
-const PORTALS_PATH = 'portals.yml';
+// Imported from scan.mjs so it honors CAREER_OPS_PORTALS and the data root (#3510).
 
 function loadKeywords() {
   const defaultKeywords = ["Software Engineer"];
--- a/scan-interamt.mjs
+++ b/scan-interamt.mjs
@@ -19,14 +19,23 @@
 
 import { chromium } from 'playwright';
 import { readFileSync, existsSync, mkdirSync } from 'fs';
+import { join } from 'path';
 import * as yaml from 'js-yaml';
-import { appendToPipeline, appendToScanHistory, loadSeenUrls } from './scan.mjs';
+import { appendToPipeline, appendToScanHistory, loadSeenUrls, PORTALS_PATH, SCAN_HISTORY_PATH } from './scan.mjs';
+import { getCareerOpsRoot } from './path-resolver.mjs';
 import { localToday } from './lib/local-today.mjs';
+import { isMainModule } from './lib/is-main-module.mjs';
 
 // ── Config ───────────────────────────────────────────────────────────
 
-const PORTALS_PATH    = 'portals.yml';
-const SCAN_HISTORY    = 'data/scan-history.tsv';
+// Both imported from scan.mjs (#3510). This scanner appends through
+// scan.mjs's appendToScanHistory — anchored to the data root — while reading its
+// own bare-relative copy to work out the last scan date, so within a single run
+// it could write one history file and read another. Its portals copy also
+// ignored CAREER_OPS_PORTALS, which #2271 added so a second search lane does not
+// poison the first one's dedup history.
+const SCAN_HISTORY    = SCAN_HISTORY_PATH;
+const DATA_ROOT       = getCareerOpsRoot();
 const INTERAMT_HOME   = 'https://interamt.de/koop/app/';
 // Direct offer URL — constructed from StellenangebotId.
 // Wicket adds a session version number (?28&id=...) during live navigation,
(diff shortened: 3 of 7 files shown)


Scott-Emberson (commented):
Reviewed at head b58acfdad0. `scan-data-paths-under-data-root.test.mjs` is exactly the shape this fix needs: it runs the real modules in child processes with cwd (a decoy dir) and CAREER_OPS_ROOT (the data root) pointed at different temp dirs, each holding a decoy copy, so a cwd-relative resolution picks the decoy and reddens. It asserts the blacklist read sees `realco` not `decoyco`, and that scan-runs and the checkpoint land under the data root not cwd. The other two files still pin their behavior: portal-health is strengthened to assert data-root and not-cwd, and the outage-checkpoint suite keeps its #2283 end-to-end check with only its isolation mechanism moved to CAREER_OPS_ROOT.

Scott-Emberson (approved):

PR #3513, opened 2026-08-29
fix(providers): a lone surrogate in one id/slug no longer aborts the whole page

ilyakanevskiy:
## Summary

`encodeURIComponent` throws `URIError` on a lone UTF-16 surrogate, and a JSON string can carry one (a `"\uD800"` escape survives `JSON.parse`). A provider that builds a job URL from a host-controlled `id`/`slug` inside a `.map()` / `for` loop and hits one such value throws out of the loop and loses every posting already parsed on that page — `scan.mjs`'s per-company `catch` then drops the lot. Thirteen JSON-API providers had this shape; `company-funded.mjs`'s Hacker News evidence loop had it too. This adds a shared `safeEncodeURIComponent` that returns `null` instead of throwing, wired so each caller drops just the bad posting — the same thing it already does for a posting with no `id`. It also fixes the mirror case on the decode side: `rheinmetall.mjs`'s slug-fallback `decodeURIComponent` was unguarded against a malformed percent-escape, where the three sibling scrapers that do the same thing guard theirs. And it corrects the one spot where a config-derived `encodeURIComponent` sat *inside* a per-posting loop — `tkms.mjs`'s `cfg.locale` — hoisting it out so a broken locale fails loud once instead of silently emptying the page.

## Why null, not a fallback

A lone surrogate — or a `U+FFFD` substitute for one, e.g. via `String.prototype.toWellFormed()` — in a job URL flows on into `data/scan-history.tsv`, the tracker, and every generated document, where it serialises to ill-formed UTF-8. For `arbeitsagentur` and `vdab` the encoded value is also the dedup key, where a degraded-but-kept value would collide distinct bad postings onto one key. `providers/_html-entities.mjs` already makes the same call on the decode side (reject the bad code point rather than emit it). So the helper returns `null` and the caller drops the posting.

## Changes

- **`providers/_safe-url.mjs`** (new) — `safeEncodeURIComponent(value)`: `encodeURIComponent(String(value))`, returning `null` instead of throwing `URIError` on a lone surrogate, and re-throwing anything else. The header documents the scope: a host-controlled value from an API response becoming a job-URL path segment inside a loop — not config-derived values, not calls already inside their own try/catch, not values a regex already restricts to safe characters.

- **Thirteen providers** route their per-posting `id`/`slug` through the helper and drop the posting on `null`: `alibaba`, `arbeitsagentur`, `bamboohr`, `feishu-jobs`, `garena`, `jibeapply`, `manfred`, `meituan`, `mokahr`, `phenom`, `thehub`, `tkms`, `vdab`. `bamboohr` gained a trailing `.filter(Boolean)` (its `.map()` had none); the rest already drop `null` through an existing `.filter(Boolean)`, `if (job)` guard, or `continue`. `feishu-jobs` and `mokahr` (merged after this PR opened, within #3491) and `garena` (merged after this PR opened, within #3567) have the same host-controlled-`id`-in-a-loop shape (`parseFeishuJobsResponse` / `parseMokaHrJobs` / `parseGarenaResponse`) and are converted here too. `garena` routes `id` through the helper via a null-returning `idUrlSegment` and skips the one posting; a `.`/`..` `id` now drops the posting there too rather than throwing, matching the surrogate case and the existing `!id` guard, while `garena`'s config-derived `office` segment still throws loudly.

- **`providers/tkms.mjs`** — besides routing `id` through the helper like the others, `parseQuery` built its URL with `encodeURIComponent(cfg.locale)` *inside* the per-posting loop. `cfg.locale` is a trusted `portals.yml` segment (`tkms.locale`, the only attested value being the default `"en"`), so it keeps a plain `encodeURIComponent` — a structural `/`, `?` or `#` stays escaped so the URL is well-formed — but the call is hoisted to a single `localeSeg` above the loop: a genuinely broken locale (a lone surrogate) throws there once, as the loud config error it is, instead of silently dropping every posting on the page one at a time. Same split as `garena`'s `urlSegment` (config, throws) vs `idUrlSegment` (response field, drops).

- **`company-funded.mjs`** — the Hacker News evidence loop built a fallback `news.ycombinator.com/item?id=` URL from `hit.objectID` (an Algolia response field) with a bare `encodeURIComponent`, inside `for (const hit of hits)`. Now routed through the helper, with no synthetic fallback URL when it can't encode (`hit.url` is tried first anyway).

- **`providers/rheinmetall.mjs`** — `parseVacancies`'s slug-fallback title path, taken only when the headline `<div>` is absent (a markup shift), called `decodeURIComponent` on a scraped href segment with no guard; a malformed percent-escape threw `URIError` out of the whole page's parse. Now wrapped in try/catch → raw slug, matching `workday.mjs`, `successfactors.mjs`, and `scan.mjs`, which all guard the same operation.

- **Tests** — five files:
  - `tests/providers/_safe-url.test.mjs` (new) — the helper in isolation: lone high / low / mid-string surrogate → `null`; a valid surrogate pair is preserved; ordinary string, empty string, and non-string coercion match `encodeURIComponent`.
  - `tests/providers/url-encoding-surrogate.test.mjs` (new) — one behavioural case per converted provider (a batch with one lone-surrogate posting and one clean one → no throw, clean posting kept, bad one dropped), plus a source-level guard: the converted set stays wired to the helper, `safeEncodeURIComponent` is never referenced without importing it, and no provider builds a `url:` line with a bare `encodeURIComponent` (allowlist: `csod.mjs`, `4dayweek.mjs`, each with a stated reason), with positive and negative controls on the guard regex. The guard is line-scoped and would not have flagged `feishu-jobs` (the `url:` assignment and the `encodeURIComponent` call are on different lines), `mokahr` (the encode is a separate `const`), or `garena` (the `url:` line calls a `urlSegment` wrapper, not `encodeURIComponent` directly), so their behavioural cases are the coverage that matters.
  - `tests/providers/rheinmetall.test.mjs` — added a slug-fallback case: a card with no headline `<div>` and a malformed `%ZZ` in its href → no throw, the sibling card survives.
  - `tests/providers/tkms.test.mjs` — two `parseQuery` cases for the configured locale: a structural character in it lands percent-encoded in the URL, and a lone-surrogate locale throws instead of yielding an empty page.
  - `tests/providers/garena.test.mjs` — a lone-surrogate `office` (config) throws, mirroring the tkms locale case; the existing `.`/`..` `id` cases were retargeted from "throws" to "drops the posting, keeps the sibling" to match the surrogate contract.

## Not in this PR

- Config-derived `encodeURIComponent` calls that sit outside any per-posting loop — `csod` corpName, `careerviet` / `itviec` keywords, `thehub` countryCode, `rippling` slug. These take a `portals.yml` value, not a host-controlled API field, and a bad one already fails loud without emptying a page, so they keep a plain `encodeURIComponent`. (`tkms`'s `cfg.locale` was the one config-derived call that *was* in a loop; it is fixed above.)
- `beesite.mjs`'s `encodeURIComponent(JSON.stringify(data))` — `JSON.stringify` (ES2019+) emits a lone surrogate as a `\udXXX` escape, so its output can't make `encodeURIComponent` throw; the input is config-derived anyway.
- `smartrecruiters.mjs` and `vdab.mjs`'s detail-fetch `encodeURIComponent` — already inside a try/catch (enrichment, fails open per detail).
- `builtin.mjs` (merged after this PR opened, within #3645) — its only `encodeURIComponent` calls are on `queries` / `categories`, which come from the `portals.yml` entry, not from a response; same class as the `careerviet` / `itviec` keywords above. Its per-posting URLs are parsed straight out of the page, not built from an encoded id.

## Verification

- `node test-all.mjs` — 8128 passed, 0 failed (12 pre-existing environment warnings only, unrelated: personal-data scan on the repo's own fixtures, CV-section-order fixtures, font-path fixtures).
- `npm run lint` — 657 `.mjs` files pass the syntax check.
- `node test-all.mjs --only providers/` — 2137 passed.

## Bug Fixes
- Prevented malformed job IDs and slugs from stopping entire job-result pages from loading.
- Invalid postings are now skipped while valid postings continue to appear.
- Improved fallback title handling when job URLs contain malformed encoded text.
- Corrected job URL generation for locale values.
- Improved URL encoding for special characters and malformed values across providers.

## Tests
- Added coverage for malformed URL values, surrogate characters, locale encoding, and fallback title parsing across providers.


Scott-Emberson (commented):
Reviewed at head 03190b04a3 (now out of draft). The three owned tests hold up well. url-encoding-surrogate.test.mjs feeds each provider one `\uD800` lone-surrogate id plus one clean id and asserts exactly the clean job survives with the right URL, so if the URIError-aborts-the-whole-page bug returned, the bad item throws out of the map/normalize call, gets caught, and the case reddens. Its source-level guard carries both a positive control (must fire on a planted bare `encodeURIComponent` url line) and a negative control (must not flag the helper), which kills the regex-slip false pass. _safe-url.test.mjs pins the helper contract with real assertions: lone high/low/mid surrogate to null, a valid surrogate pair preserved, ordinary/empty/reserved identical to encodeURIComponent, and a URIError from the value's own toString propagates rather than being swallowed. rheinmetall.test.mjs keeps its existing assertions (id/title pairing, host guards, pagination, the #1639 out-of-range-entity regression) and only adds a malformed-slug fallback case, so it is not weakened.

Scott-Emberson (approved):

Scott-Emberson (commented):
Re-reviewed at head 593c7c8fd7 (moved past the earlier approval, largely a repo-move main merge). On a two-dot diff, _safe-url.test.mjs and rheinmetall.test.mjs are byte-identical to the approved state, so their assertions (the safe-encode contract, and rheinmetall's id/title/SSRF/pagination plus the slug-fallback case) stand unchanged. The one genuine change is in url-encoding-surrogate.test.mjs, and it is purely additive: it adds feishu-jobs and mokahr cases each asserting that a lone-surrogate id drops just that one posting while the clean sibling survives (jobs length 1), the same contract as the existing providers, and it bumps the CONVERTED provider list from 10 to 12 with the static guard still requiring the shared import plus the helper use per provider (the pass string is now self-counting). The pre-existing positive and negative source controls are untouched.

Owned tests verified sound, the delta only extends the surrogate-safe coverage to the two new providers without weakening anything. Holding formal approval until the branch is finalized (fresh approval needed since the head moved), nothing blocking from the tests/ side.

Scott-Emberson (approved):

Scott-Emberson (commented):
Re-checked at head bf46f8b186 (moved past the earlier approval on a large repo-move merge). On a blob compare, tests/providers/_safe-url.test.mjs and tests/providers/rheinmetall.test.mjs are byte-identical to what I approved, and the only genuine change is in tests/providers/url-encoding-surrogate.test.mjs: a purely additive garena.parseGarenaResponse case (a lone-surrogate id drops just that row while the clean sibling survives) plus garena.mjs added to the converted-providers list. No earlier assertion was touched, reordered, or weakened, so the safe-encode contract, the page-abort behavior, and rheinmetall's id/title/SSRF/pagination/slug-fallback all stand.

Owned tests verified unchanged and sound, the one addition follows the same drop-one/keep-sibling contract. This is in our area and up to date (the branch is not behind, only blocked on our review), so it is ready to approve on your go-ahead.

Scott-Emberson (commented):
Re-checked at head 963f57a467. The head moved but this was a rebase, not new work in our area: on a blob-SHA compare the four owned test files (_safe-url.test.mjs, url-encoding-surrogate.test.mjs, garena.test.mjs, rheinmetall.test.mjs) are byte-identical to the version I last reviewed at bf46f8b186 (the branch replayed its 7 commits over 5 new main commits into fresh SHAs, but the owned content did not change). So the coverage stands: safeEncodeURIComponent returns null on a lone/unpaired surrogate and preserves a valid pair with the URIError propagating, the drop-one-posting-not-the-page contract is exercised across the 13 routed providers (including feishu-jobs, mokahr, garena), and the static guard asserts each routed provider imports _safe-url and calls the helper with positive and negative controls. The one changed garena assertion (a bad id now drops that posting and keeps the sibling, rather than throwing) is the intended strengthened contract, and rheinmetall only gained an additive slug-fallback guard.

Owned tests verified unchanged and sound. CI is green (including the upgrade gate) and the branch is up to date, only blocked on our review, so this is ready to approve on your go-ahead.

Scott-Emberson (commented):
Re-checked at head aa6c8dc19e. On a blob compare the subject test tests/providers/url-encoding-surrogate.test.mjs is byte-identical to what I last reviewed, so this head move is a pure rebase with no content change in our area. The drop-one-posting-not-the-page contract across the routed providers and the static import-and-use guard stand.

Owned tests verified unchanged and sound. The only thing holding it is that the branch is behind main. Rebase to current main and I will re-verify and clear it to approve.

Scott-Emberson (commented):
Re-verified at the current head after the rebase and the follow-up fix commits. This is real new owned work since my last look, not a pure rebase: `_safe-url.mjs` now coerces `String(value)` before the try (so a URIError from the value's own `toString` propagates as a caller bug instead of being swallowed to null), `garena.mjs` routes the remote id through a new `idUrlSegment` that drops on `.`/`..`/surrogate while config `office` still throws, `tkms.mjs` hoists the locale re-encode out of the row loop to fail loud, `rheinmetall.mjs` wraps its slug-fallback decode in try/catch, and the garena/rheinmetall/tkms owned tests gained assertions.

The owned tests are sound and reddening. I mutation-tested by reverting the helper to plain `encodeURIComponent(String(value))` (the pre-fix throwing shape) and both `_safe-url.test.mjs` and `url-encoding-surrogate.test.mjs` (13 provider checks) go red, so the guard genuinely catches a regression back to aborting the page. On success the helper returns `encodeURIComponent`'s output so all separators and control chars are percent-encoded, and on a lone surrogate it returns null and the caller drops just that one posting; the host is a per-provider constant or an independently-validated origin, never derived from the id, so there is no new SSRF surface. The config-vs-remote distinction (fail loud for config, drop-one for a scraped id) is correctly asserted on both sides for garena, tkms, and rheinmetall.

Owned tests verified sound, CI green, mergeable and not behind. This is GATE-tier (`providers/_safe-url.mjs` and the provider owned tests are the substance). Holding formal approval until Scott gives the go-ahead, and surfacing it as ready-to-approve.

Scott-Emberson (commented):
Re-verified at the current head. The owned tests `tests/providers/_safe-url.test.mjs` and `tests/providers/url-encoding-surrogate.test.mjs` are byte-identical to the head I last verified sound (the three-dot compare listing them as added is merge-base noise from a large main merge, which the direct blob compare settles), so my prior verification carries: reverting the `_safe-url` helper to its throwing shape reddens both suites, the helper returns encoded output or a null-drop on a lone surrogate, the host is never derived from the id so there is no SSRF surface, and the config-fail-loud versus remote-drop distinction is asserted.

Owned tests verified sound, CI green, mergeable and not behind. This is GATE-tier (`providers/_safe-url` and the provider owned tests). Holding formal approval until Scott gives the go-ahead; surfacing it as ready-to-approve.

FReptar0 (approved):
@santifer handed this one to me, so this is an independent pass rather than a second opinion on his.

Note this was rebased on 09-05 after my first read, so everything below is **re-verified against the current head (`5b8f1f0`) and today's `main`**, not carried over.

**The helper, in isolation — seven shapes:**
```
lone high (\uD800)   → null        valid pair (😀)   → "%F0%9F%98%80"
lone low  (\uDC00)   → null        ordinary "a/b?c"  → "a%2Fb%3Fc"
mid-string           → null        ""                → ""    · 42 → "42"
```
It also re-throws a broken `toString` (`TypeError: boom`) rather than swallowing it. Coercing outside the `try` is what makes `null` mean *lone surrogate* and nothing else — the detail a reviewer skims past, and the one that keeps a caller bug from being silently reclassified as bad host data.

**The fix end-to-end on a real provider,** one page carrying one bad `id`:
```
── bamboohr on origin/main (today) ──
  THREW: URIError                                ← all three postings lost

── bamboohr with this PR ──
  no throw · jobs = 2 → Engineer A, Engineer C   ← only the bad one drops
```

**All 13 named providers are actually wired to the helper** — I checked each rather than trusting the list: `alibaba, arbeitsagentur, bamboohr, feishu-jobs, garena, jibeapply, manfred, meituan, mokahr, phenom, thehub, tkms, vdab`, 13/13 importing `safeEncodeURIComponent`.

**The source-level guard is non-vacuous.** Reverting `manfred.mjs` to a bare `encodeURIComponent` fires it on behaviour:
```
❌ manfred.normalizeManfredOffer: threw on a lone-surrogate id (URIError: URI malformed)
```
That guard is the part with the longest half-life. The entity-decoder class needed four rounds (#1555 → #1639 → #2623 → #2790) before a source-level assertion stopped it recurring; shipping one here means this class doesn't need the same four.

**The "why null" argument holds.** `toWellFormed()` would keep the posting but push a U+FFFD into `scan-history.tsv`, the tracker and every generated document — and for `arbeitsagentur`/`vdab` the encoded value is the dedup key, so substitution would collide distinct bad postings onto one. Dropping is the same call `_html-entities.mjs` already makes on the decode side. Consistent with what's there, and argued rather than assumed.

**On the earlier approvals:** @Scott-Emberson's two (08-31, 09-02) predate the 09-05 rebase, so they were dismissed and `reviewDecision` reads `REVIEW_REQUIRED`. Flagging it because it also means his eyes never saw the tkms locale hoist that landed last.

Approving. It's `BEHIND`, so I'll run the Update-branch cycle before merging — unlike @santifer I don't bypass branch protection, so that's a separate step rather than a formality.

Nice work @ilyakanevskiy. The "Not in this PR" section is what made 21 files reviewable in one sitting: every `encodeURIComponent` you *didn't* touch has a stated reason, so the scope is checkable instead of having to be taken on faith. That's the part I'd point other contributors at.

PR #3537, opened 2026-08-30
fix(skill-extract): recognize marketing/GTM skills, and stop reporting tracked employers as gaps

rebeccasims-ai:
## The bug

`extractSkills` is a closed allowlist, and every token in it is an engineering tool or a delivery credential. A marketing-operations user is therefore structurally invisible to `upskill`: their gap map can only ever report the handful of engineering words that happen to appear in their reports.

This is the same hole the certification block (#2603) was added to close, one discipline further over, and it fails the same way — silently, producing a plausible-looking gap map that simply omits the recurring gap.

Measured on a real marketing-operations corpus:

- **"ABM" was an explicit `soft_gap` in several linked reports** — twice flagged *in the report text itself* as a recurring gap — and never once reached the output.
- The **top-ranked entry was a data-platform vendor's NAME**, matched out of the provenance asides inside those very ABM sentences. A report logging a recurring gap names the sibling companies where it was logged before (`"4th log: <company>, <company>, now here"`), and the extractor cannot tell that from a JD asking for the tool of the same name.
- So the tool ranked *"learn `<vendor>`"* above the gap those sentences were recording.
- On the known-skills side the same blindness left a marketing `cv.md` yielding a known-skills set of **five**, so almost nothing could be excluded as already-held.

Worth stressing that this is **not** a multi-word or casing problem — multi-word tokens like `React Native` and `Lean Six Sigma` already work. The vocabulary, not the scoring, was the constraint, exactly as the #2603 comment records.

## The fix

**`skill-extract.mjs`** gains a marketing/GTM vocabulary block: ABM, programmatic/DSP, media agency management, the demand/lifecycle/paid disciplines, and the platforms a marketing CV lists. Longest-first within each family, matching the existing `React Native`-before-`React` convention.

Two deliberate boundaries, both tested:

- **ABM is NOT aliased to Demandbase.** The platform is not the program. Aliasing them would put ABM into the known-skills set of anyone whose CV lists the tool and re-bury the gap as *"no gap found"* — worse than the silence being fixed, because it reads as a verified absence. Same rule as the existing no-umbrella-aliases policy (`"cloud"` must never count as knowing AWS).
- **`CRO`, `GEO`, bare `Programmatic`, `Segment` and `Iterable` are omitted**, on the asymmetry the `CSM` note already sets out. Each collides with everyday or adjacent prose — *"reports to the CRO"*, *"the EMEA geo"*, *"programmatic access to the API"*, *"each segment"*, *"any iterable"* — and a false token mints a phantom gap on every posting that uses the word. `Iterable` matters especially because this module is shared with `jd-skill-gap`, which runs over full engineering JDs. Guard tests pin all five.

**`upskill.mjs`** excludes skill tokens that name a company in the user's own tracker, checked *after* known-skills so a genuine known skill keeps its bucket. Reported as `excludedAsCompanyName` rather than dropped — this module's own history (see the known-skills header) is that a silent suppression is indistinguishable from *"never appeared"*, which is what made the original suppression bug expensive. Someone who genuinely needs the tool and once applied to the company of the same name sees it listed rather than losing it without a trace.

`SCHEMA_VERSION` 1 → 2, since a v1 and a v2 gap list are not comparable: a gap "appearing" in v2 may be a word v1 could not see, and one "closing" may be an employer name v1 mistook for a skill.

## Result

On the corpus above, same data, before and after:

| | before | after |
|---|---|---|
| top gap | `<vendor>` (a company name) | **ABM** — Critical, 6 reports |
| marketing gaps found | none | ABM, DSP, Programmatic, Media Agency Management, Demand Generation, Paid Media, … |
| `knownSkillCount` | 5 | 24 |
| company names as gaps | 2 | 0 (reported under `excludedAsCompanyName`) |

## Tests

- `tests/skill-extract.test.mjs` — 27 assertions: table-driven over the whole new vocabulary asserting **both** halves (extraction from lowercase prose *and* canonicalization to the display form, the drift class this module exists to prevent), the five omission guards, longest-first precedence, the `SEM`/`SEMrush` boundary, and the Demandbase/ABM boundary.
- `tests/upskill-company-names.test.mjs` (new) — the exclusion, that it is reported and not silent, bucket precedence vs known-skills, case/whitespace-insensitivity, and backward compatibility when no company set is passed (`aggregateGaps` keeps its 2-arg behaviour exactly).

All green, plus `upskill.mjs --self-test`. Full suite shows no new failures against a clean `main` tree.

Fixtures are synthetic throughout — the corpus that surfaced this is a real job search, so no company names, report contents or CV lines from it appear here.

## Summary by CodeRabbit

- **New Features**
  - Expanded skill recognition and normalization for marketing, GTM, revenue operations, analytics, and marketing platforms.
  - Reports now distinguish tracked company names from genuine skill gaps.
  - Added company-excluded skill counts and tracked-company totals to report summaries.

- **Bug Fixes**
  - Improved handling of aliases, spelling variations, case, and whitespace.
  - Preserved valid ABM gaps when related companies such as Demandbase are tracked.
  - Updated report schema to reflect non-comparable results after these changes.

### DIFF
--- a/skill-extract.mjs
+++ b/skill-extract.mjs
@@ -91,6 +91,75 @@ export const SKILL_TOKENS = [
   //
   // 'SAFe' is NOT here either — it is handled case-sensitively below, for the
   // same reason 'Go' is: 'safe' is an everyday English word.
+
+  // ── Marketing / GTM / revenue operations (added 2026-08-30) ──────────────
+  // The certification block above records that the vocabulary, not the scoring,
+  // was the constraint for delivery and program managers. The same hole ran one
+  // discipline further over: every token above this line is an engineering tool
+  // or a delivery credential, so a marketing-operations candidate was
+  // structurally invisible to the gap map.
+  //
+  // Measured on a real marketing-operations corpus: "ABM" was logged as an
+  // explicit soft_gap in several linked reports — twice flagged in the report
+  // text itself as a recurring gap — and never once reached the output. The
+  // map's top-ranked entry was instead an observability vendor's name, matched
+  // out of the provenance asides inside those very ABM sentences (a report
+  // logging a recurring gap tends to cite the sibling companies where it was
+  // logged before). The same blindness ran on the known-skills side: a
+  // marketing cv.md yielded a known-skills set of FIVE, so almost nothing could
+  // be excluded as already-held.
+  //
+  // Longest-first within each family, matching the 'React Native'-before-'React'
+  // and 'Lean Six Sigma'-before-'Six Sigma' convention above.
+  'Account-Based Marketing', 'Account Based Marketing', 'ABM',
+  'Demand-Side Platform', 'Demand Side Platform', 'DSPs', 'DSP',
+  // Only the MEDIA senses of "programmatic" — see the omission note below.
+  'Programmatic Advertising', 'Programmatic Display', 'Programmatic Media',
+  'Programmatic Buying',
+  'Media Agency Management', 'Media Agencies', 'Media Agency',
+  'Demand Generation', 'Demand Gen',
+  'Marketing Automation', 'Marketing Operations',
+  'Revenue Operations', 'RevOps',
+  'Lifecycle Marketing', 'Growth Marketing', 'Product Marketing',
+  'Partner Marketing', 'Performance Marketing', 'Field Marketing',
+  'Content Marketing', 'Email Marketing',
+  'Paid Media', 'Paid Social', 'Paid Search',
+  'Conversion Rate Optimization',
+  'Marketing Mix Modeling', 'Media Mix Modeling', 'Incrementality',
+  'Lead Scoring', 'Marketing Attribution',
+  // 'SEMrush' precedes 'SEM' by convention; the (?!\w) boundary already stops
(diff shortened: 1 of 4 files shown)


Scott-Emberson (commented):
Reviewed at head 0bf9f7d12c. Both owned tests pair positive extraction with a concrete negative, so neither can pass by luck. skill-extract.test.mjs asserts across 50-plus GTM tokens (ABM, DSP, programmatic, HubSpot, GA4, ...) that each extracts AND canonicalizes to its display form, while the negatives reject real non-skills and collisions (`programmatic access` is not `Programmatic`, CRO/GEO/Segment/Iterable collisions rejected, `demandbase` maps to Demandbase not ABM), so a recognizer or canonical-map regression reddens. upskill-company-names.test.mjs asserts a tracked employer (Snowflake) is excluded from gaps and recorded as `excludedAsCompanyName` while a genuine gap (ABM, Kubernetes) still surfaces, and pins precedence (known-skill bucket wins over company) plus the 2-arg backward-compat path, so dropping or over-applying the employer filter reddens.

Scott-Emberson (approved):

PR #3540, opened 2026-08-30
fix: keep MokaHR escaped markup inert

chipoto69:
## Summary
- Builds on renlon:codex/china-providers-pr-prep from PR #3491.
- Fixes the CodeQL findings in `providers/mokahr.mjs` by stripping real HTML tags, removing leftover literal angle brackets, and keeping escaped `<`, `>`, and `&` sequences inert text.
- Adds regression coverage for encoded markup/XSS-shaped payloads and malformed leftover angle-bracket payloads in MokaHR job descriptions.

## Test plan
- `git diff --check`
- `node --check providers/mokahr.mjs`
- `node tests/providers/mokahr.test.mjs`
- `npm run lint`

Context: bridge for https://github.com/santifer/career-ops/pull/3491 because the original PR is blocked by high-severity CodeQL alerts. Supersedes #3539; v2 was opened instead of force-pushing the first bridge branch.

## Summary by CodeRabbit

* **New Features**
  * Added support for Feishu Jobs and MokaHR career sites, including job search, filtering, pagination, and job detail extraction.
  * Added recognition for Chinese application controls such as “申请职位” and “投递.”
* **Bug Fixes**
  * Improved duplicate detection for MokaHR job links.
  * Prevented unrelated URLs from contributing location hints.
* **Documentation**
  * Updated supported and evaluated job board documentation.
* **Tests**
  * Added comprehensive coverage for new providers, application detection, URL handling, pagination, and failure recovery.


Scott-Emberson (commented):
Reviewed at head 7d280c5ee9. The MokaHR escaped-markup fix is well tested. The encoded-xss case feeds `&lt;script&gt;`-style and `&amp;lt;img`-style input and asserts both directions: the output does not contain the live `<script>` or `` tags, and it does keep the escaped text form, so un-escaping into live markup reddens all four checks rather than a one-sided negation. That lines up with providers/mokahr.mjs stripHtml(), which only strips literal `<...>` spans and deliberately does not decode `&lt;`/`&amp;` (there is an explicit comment to that effect). The malformed-xss case (`<<script>`) asserts residual-bracket removal too, and the liveness-core change is purely additive (two new Chinese apply-control cases, nothing deleted or loosened).

One non-blocking aside from reading the wider diff, out of this fix's scope and not a regression it introduces: providers/feishu-jobs.mjs parseFeishuJobsResponse concatenates `description`/`requirement` raw, without the stripHtml that mokahr now uses, so if Feishu JD fields ever carry HTML they pass through unstripped. Worth a follow-up on the Feishu path, but it is pre-existing and separate from the MokaHR markup this PR makes inert.

Scott-Emberson (approved):

PR #3557, opened 2026-08-31
feat(resume): add governed JobBot render contract

mannan24:
## Summary by CodeRabbit

* **New Features**
  * Added a guided resume workflow to identify opportunities, prepare tailored resume requests, validate results, and generate ATS-friendly PDFs.
  * Added versioned receipts and safeguards to keep resume outputs traceable and based on unchanged source information.
  * Added an optional staging location for generating PDFs without publishing them to the standard manifest.

* **Improvements**
  * Updated resume contact details to display consistently in one row.

* **Tests**
  * Added coverage for resume generation, validation, source integrity, PDF rendering, and security safeguards.


Scott-Emberson (commented):
Reviewed at head 91acc6946a. The owned test tests/jobbot-resume-contract.test.mjs holds this net-new render path to a real governed contract, and I especially like that it pins the source-of-truth property with teeth: a tampered candidate.name, even with a re-computed valid manifest hash, still throws `/verified profile/` and the renderer never runs (calls===0), because renderResume re-derives the candidate from the verified profile before any command. Beyond that, the receipt shape is pinned exactly (version, artifact_key, mime_type, and it asserts the JSON is path-free), the candidates contract is deepEqual'd, and the refusal paths all redden with the renderer never invoked: extra field, wrong manifest_hash, changed cv.md, changed modes/pdf.md rules, symlinked report that escapes, and the over-two-pages policy.

One non-blocking nit: the fact-gate mock always returns a pass verdict, so the branch where a fail verdict should block the render is not pinned by a reddening negative (the module does enforce the pass/warn set, it just is not exercised). Worth a case that feeds a fail verdict and asserts no render. Everything else, including drop-a-field and wrong-shape, reddens.

This is a net-new feature, so it is more than a tests/ change, but from the tests/ side the coverage is sound. Holding formal approval until the branch is finalized, nothing blocking from the tests/ side.

Scott-Emberson (commented):
Re-reviewed at head f82c75a8e4. This closes the nit I raised. The delta (past a release chore and a main merge) adds one test, "render refuses a failed fact gate before invoking the PDF renderer": it feeds a non-pass fact-gate verdict (`{"verdict":"fail"}`), has the builder emit an unverified-claim body, then asserts renderResume throws `/fact gate/` AND that generate-pdf.mjs never appears in the recorded calls. The companion happy-path test (verdict pass) still renders a real receipt, so the throw is genuinely driven by the fail verdict rather than an unconditional refusal, which rules out a false pass. The merge left the earlier assertions intact: the source-of-truth candidate check (`/verified profile/`, renderer never invoked), the path-free receipt shape, and every refusal path (changed cv, changed rules, symlink escape, render policy).

Owned test verified sound, and the fact-gate rejection is now pinned. Holding formal approval until the branch is finalized, nothing blocking from the tests/ side.

Scott-Emberson (approved):

Issue #3559, opened 2026-08-31
[BUG] Ensure tracker, plugin, and ATS tooling consistently use canonical states and configured data root


hardik2004gupta:
### Code of Conduct

- [x] I agree to follow the [Code of Conduct](https://github.com/santifer/career-ops/blob/main/CODE_OF_CONDUCT.md)

### Existing issues

- [x] I searched existing issues and this hasn't been reported yet

## Description

Five correctness bugs were discovered during a codebase review. They are independent fixes but share enough context to triage together. Each issue causes incorrect validation, duplicate data, empty snapshots, broken tracker links, or incorrect file I/O when `CAREER_OPS_ROOT` is configured.

---

## 1. `merge-tracker.mjs` and `verify-pipeline.mjs` hardcode canonical states instead of loading `templates/states.yml`

**Files:**
- `merge-tracker.mjs:141`
- `verify-pipeline.mjs:52–75`
- `tracker-utils.mjs:668`

`templates/states.yml` is the declared single source of truth for canonical states according to `DATA_CONTRACT.md`. `tracker-utils.mjs` already exports `loadCanonicalStates()` at line 668, and `set-status.mjs` already uses it correctly.

However, both `merge-tracker.mjs` and `verify-pipeline.mjs` maintain their own hardcoded copies:

```js
// merge-tracker.mjs
const CANONICAL_STATES = [
  'Evaluated',
  'Applied',
  'Responded',
  'Interview',
  'Offer',
  'Hired',
  'Rejected',
  'Discarded',
  'SKIP'
];

// verify-pipeline.mjs
const CANONICAL_STATUSES = [
  'evaluated',
  'applied',
  // ...
];

const ALIASES = {
  // hardcoded Spanish/English aliases
};
```

If a state or alias is added to `states.yml`, `merge-tracker.mjs` can reject valid batch rows while `verify-pipeline.mjs` reports false validation errors. There is currently no CI coverage that guarantees these scripts remain synchronized with the canonical state configuration.

### Expected behavior

Both scripts should call:

```js
loadCanonicalStates(join(ROOT, 'templates/states.yml'))
```

and derive their canonical state/status lists and aliases from the returned configuration rather than maintaining independent hardcoded copies.

### Reproduction

1. Add a new state to `templates/states.yml`, for example:

   ```yaml
   - id: pending
     label: Pending
     aliases: [waiting]
   ```

2. Add a tracker row with:

   ```text
   Status: Pending
   ```

3. Run:

[...]

PR #3561, opened 2026-08-31
fix: canonical states, data-root paths, pipeline parser, and link regex (#3559)

hardik2004gupta:
## Summary

Fixes five independent bugs reported in #3559 that each silently produced wrong
behaviour without surfacing any error:

- **Hardcoded canonical states** (`merge-tracker.mjs`, `verify-pipeline.mjs`) —
  both files maintained a hand-rolled status list and a duplicate Spanish-alias
  table that drifted from `templates/states.yml` whenever a state or alias was
  added there. Both now load exclusively through the existing
  `loadCanonicalStates()` helper. `verify-pipeline.mjs` also switches
  `CANONICAL_STATUSES` from `Array` to `Set` (`.includes()` → `.has()`).

- **`plugins.mjs` ignored `CAREER_OPS_ROOT`** — `APPLICATIONS_PATH` and
  `PIPELINE_PATH` were anchored to the code directory (`path.dirname(import.meta.url)`)
  instead of the user-configured data root. Plugin import/export/status hooks
  silently read and wrote the wrong files for any user whose data lives outside
  the repo root. Both paths now derive from `getCareerOpsRoot()`; `ROOT` (used
  for plugin loading) is unchanged.

- **`buildSnapshot()` used a table parser on a checklist file** — `data/pipeline.md`
  is a Markdown checklist (`- [ ] url`), not a table. Passing it to
  `parseMarkdownTable()` always returned `[]`, so `snapshot.pipeline` was
  permanently empty for every plugin hook that consumed it. The fix uses the
  correct checklist regex and freezes each matched entry to match the snapshot's
  immutability contract.

- **Link regex truncated filenames with parentheses** (`tracker-links.mjs`) —
  `/\]\(([^)]+)\)/g` stops at the first `)`, silently truncating report slugs
  such as `042-acme-(berlin)-2024-01-15.md`. Replaced with a nested-paren-aware
  pattern that handles one level of parentheses, covering all real report slugs.

- **`scan-ats-full.mjs` used bare relative paths** — `PORTALS_PATH`,
  `PIPELINE_PATH`, `CACHE_DIR`, and `CHECKPOINT_PATH` were bare strings resolved
  against the process working directory. Running the scanner from any directory
  other than the repo root, or with `CAREER_OPS_ROOT` set to a separate data
  directory, caused it to read the wrong portals config, write cache and
  checkpoint files to the wrong location, and fail to resume an interrupted
  sweep. All four constants now anchor to `getCareerOpsRoot()`; the
  `CAREER_OPS_PORTALS` env override is preserved and takes priority.

## Test plan

- [x] `tests/states-canonical-load.test.mjs` (new, 6 assertions) — verifies
  `loadCanonicalStates()` returns all 9 states, resolves Spanish aliases, handles
  a custom states file, and throws clearly on malformed input
- [x] `tests/plugins-data-root.test.mjs` (new, 5 assertions) — verifies
  `CAREER_OPS_ROOT` override is respected, checklist regex matches correctly,
  and the old `parseMarkdownTable` approach genuinely returns 0 on checklist
  input (proving the bug was real)
- [x] `tests/tracker-links-parens.test.mjs` (new, 4 assertions) — verifies
  plain filenames are preserved, filenames with parentheses are preserved, non-
  report links are left untouched, and multiple links in one field all normalize
  correctly
- [x] `tests/scan-ats-full-data-root.test.mjs` (new, 7 assertions) — verifies
  all four path constants anchor to the data root, `CAREER_OPS_PORTALS` override
  takes priority, and `loadCheckpoint` is exported and handles a missing file
- [x] `tests/scan-ats-full-outage-checkpoint.test.mjs` (existing, updated) —
  `sweep()` now passes `CAREER_OPS_ROOT: dir` in the child process env so the
  sandbox resolves correctly after the Bug 5 path fix; all 3 assertions still
  pass
- [x] `node test-all.mjs --quick` — 7303 passed, no new failures introduced

## Summary by CodeRabbit

- **New Features**
  - Canonical application statuses and aliases load from configurable state definitions.
  - Data paths support a configurable career operations root.
  - Pipeline checklist entries are recognized when building snapshots.
  - Report links containing parentheses are normalized correctly.

- **Bug Fixes**
  - Improved status validation, pipeline verification, and ATS data-location handling.
  - Preserved links and paths containing nested parentheses.

- **Tests**
  - Added regression coverage for status loading, data-root resolution, pipeline parsing, outage checkpoints, and link handling.

### DIFF
--- a/merge-tracker.mjs
+++ b/merge-tracker.mjs
@@ -23,7 +23,7 @@ import { getCareerOpsRoot } from './path-resolver.mjs';
 import { roleFuzzyMatch } from './role-matcher.mjs';
 import { parsePdfIndex } from './find.mjs';
 import { LEGACY_COLMAP, detectColumns, isHeaderRow, resolveScoreStatus, normalizeVia, SEPARATOR_ROW_RE } from './tracker-parse.mjs';
-import { resolveTrackerPath, resolveWorkspaceRoot, resolvePdfIndexPath, trackerLockDirFor, acquireTrackerLock, writeFileAtomic, normalizeCompany, cell } from './tracker-utils.mjs';
+import { resolveTrackerPath, resolveWorkspaceRoot, resolvePdfIndexPath, trackerLockDirFor, acquireTrackerLock, writeFileAtomic, normalizeCompany, cell, loadCanonicalStates } from './tracker-utils.mjs';
 // Canonical posting-URL key. Kept in its own module so scan.mjs / scan-history
 // can adopt the same key later without the definitions drifting.
 import { normalizeUrl } from './url-key.mjs';
@@ -137,8 +137,16 @@ try {
   process.exit(1);
 }
 
-// Canonical states and aliases
-const CANONICAL_STATES = ['Evaluated', 'Applied', 'Responded', 'Interview', 'Offer', 'Hired', 'Rejected', 'Discarded', 'SKIP'];
+// Canonical states — loaded from templates/states.yml, the single source of truth.
+// Adding a state or alias there is all that is needed; no code change required here.
+const _CODE_ROOT = dirname(fileURLToPath(import.meta.url));
+const _STATES_FILE = existsSync(join(_CODE_ROOT, 'templates/states.yml'))
+  ? join(_CODE_ROOT, 'templates/states.yml')
+  : join(_CODE_ROOT, 'states.yml');
+const _canonicalStates = loadCanonicalStates(_STATES_FILE);
+const _aliasMap = Object.fromEntries(
+  _canonicalStates.flatMap(s => s.aliases.map(a => [a.toLowerCase(), s.label]))
+);
 
 /**
  * Convert raw addition status text into one canonical tracker state.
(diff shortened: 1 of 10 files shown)


Scott-Emberson (commented):
Reviewed at head e548bd4f99. Three of the five owned tests are solid and I am happy with them: tracker-links-parens.test.mjs imports normalizeReportLink directly and a `(senior)` target reddens under the old `[^)]+` regex (which would truncate to `042-acme-(senior`); states-canonical-load.test.mjs imports the real loadCanonicalStates and its custom-yaml-parse and malformed-throws arms are the true discriminators that it reads states.yml rather than a hardcoded list; and scan-ats-full-outage-checkpoint.test.mjs is a full child-process run that keeps the checkpoint on EAI_AGAIN, resumes to completion without double-counting, and deletes on ENOTFOUND, exercising the data-root path anchoring end-to-end.

Two of the five, though, do not actually guard the fix they are named for, so I cannot call the set ready:

1. tests/plugins-data-root.test.mjs never imports plugins.mjs. It exercises path-resolver's getCareerOpsRoot and an inline copy of the checklist regex, so reverting plugins.mjs (APPLICATIONS_PATH/PIPELINE_PATH back to join(ROOT, ...), or buildSnapshot back to the old parseMarkdownTable) would not redden this test. The "old parser is broken" arm re-implements a `|`-filter rather than calling the real parser.

2. tests/scan-ats-full-data-root.test.mjs blocks 1-2 assert against an inline "Simulates:" copy of the constant rather than scan-ats-full.mjs's own DATA_ROOT, and block 3 only asserts the loadCheckpoint export plus null-on-missing. The genuine path coverage for that module comes from the outage-checkpoint test, not this file.

The fixes themselves look correct, and each is verified somewhere, but for these two the owned test would stay green if the module regressed, which is the one thing I hold the line on. Could you bind them to the real modules: import plugins.mjs and assert its resolved APPLICATIONS_PATH/PIPELINE_PATH and buildSnapshot output under a temp data root, and have scan-ats-full-data-root import the module's DATA_ROOT-derived paths rather than an inline copy. Once those two redden on a real regression this is good to go from my side. Holding formal approval until then.

Scott-Emberson (commented):
Re-reviewed at head d81c728747. This closes the two false-passes I flagged. plugins-data-root.test.mjs now dynamic-imports the real plugins.mjs (cache-busted) after setting CAREER_OPS_ROOT and asserts the module's own resolved APPLICATIONS_PATH / PIPELINE_PATH plus the real buildSnapshot output, so reverting plugins.mjs to join(ROOT, ...) (temp not equal to code dir) or reverting buildSnapshot to the old parseMarkdownTable (returns zero checklist URLs) now reddens. scan-ats-full-data-root.test.mjs likewise imports the module's real DATA_ROOT-derived PORTALS_PATH / PIPELINE_PATH / CACHE_DIR / CHECKPOINT_PATH instead of the inline "Simulates" copy, so a revert to the cwd-relative path reddens. The other three owned tests (tracker-links-parens, states-canonical-load, outage-checkpoint) were already sound and are untouched.

Both owned tests now bind to the real production modules, so the earlier inline-copy blind spot is gone. Owned tests verified sound. Holding formal approval until the branch is finalized, nothing blocking from the tests/ side.

Scott-Emberson (approved):

PR #1017, opened 2026-06-14
feat(scan): persist compensation to pipeline.md

pragalbh-dev:
## What does this PR do?

Implements **B3** from #586. Providers like Ashby already parse compensation into an annualized `{min,max,currency}` on `offer.salary` (the salary filter consumes it), but it was never written out. This persists it to `data/pipeline.md` as a 5th pipe-delimited column:

```
- [ ] {url} | {company} | {title} | {location} | {compensation}
```

rendered as a sanitized range like `180000-220000 USD`. **Persistence only — no scoring logic** (as scoped in the issue).

Built on the current `formatPipelineOffer` sanitized path (post-#1098): compensation is escaped like every other field. Because the columns are positional, a present compensation forces the (possibly empty) location cell so comp always lands in column 5. Offers without comp keep the 1–4 column behavior. Non-positive bounds are dropped (a `0` min/max is meaningless comp data, not "$0").

Also gates the location column on an actual string, so malformed provider data (a number/object) degrades to the documented 3-column form instead of stringifying into a spurious column — this addresses CodeRabbit's location-gating note from #1015.

Note: the issue assumed the field was named `compensation`; it's actually already wired as `offer.salary` by the Ashby provider, so this reuses that.

Rebased onto `main` now that B1 (#1015) has merged — this is a single self-contained commit (no longer stacked). `node test-all.mjs` passes (526/0), including the existing pipeline injection/sanitization test.

## Related issue

Refs #586 (item B3)

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation / translation
- [ ] Refactor (no behavior change)

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/santifer/career-ops/blob/main/CONTRIBUTING.md)
- [x] I linked a related issue above (required for features and architecture changes)
- [x] My PR does not include personal data (CV, email, real names)
- [x] I ran `node test-all.mjs` and all tests pass
- [x] My changes respect the [Data Contract](https://github.com/santifer/career-ops/blob/main/DATA_CONTRACT.md) (no modifications to user-layer files)
- [x] My changes align with the [project roadmap](https://github.com/santifer/career-ops/discussions/156)

## Summary by CodeRabbit

* **New Features**
  * Pipeline entries now support variable-width pipe-delimited rows, including optional location and optional compensation (range format) while staying compatible with older 3-column rows.
* **Bug Fixes**
[...]

PR #1015, opened 2026-06-14
feat(scan): preserve job location in pipeline.md

pragalbh-dev:
## What does this PR do?

Implements **B1** from #586. The scanner already parses each offer's `location`, but `appendToPipeline` discarded it. This emits it as a 4th pipe-delimited column in `data/pipeline.md` — `- [ ] {url} | {company} | {title} | {location}` — when the ATS exposes it. Offers with no location keep the original 3-column form, which downstream readers treat as an empty location.

One correction to the issue's proposed plan: **no `pipeline.md` parser needs a column-count guard.** `scan.mjs`'s URL dedup (`loadSeenUrls`) matches the URL by regex and ignores trailing columns, and `verify-pipeline.mjs` / `dedup-tracker.mjs` parse `applications.md`, not `pipeline.md`. So the only writer (`appendToPipeline`, shared with `scan-ats-full.mjs` via import) changed.

The row formatting is extracted into a small exported `formatPipelineEntry()` helper (also removes a duplicated inline template) to give the change a unit-test seam.

## Related issue

Refs #586 (item B1)

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation / translation
- [ ] Refactor (no behavior change)

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/santifer/career-ops/blob/main/CONTRIBUTING.md)
- [x] I linked a related issue above (required for features and architecture changes)
- [x] My PR does not include personal data (CV, email, real names)
- [x] I ran `node test-all.mjs` and all tests pass
- [x] My changes respect the [Data Contract](https://github.com/santifer/career-ops/blob/main/DATA_CONTRACT.md) (no modifications to user-layer files)
- [x] My changes align with the [project roadmap](https://github.com/santifer/career-ops/discussions/156)

## Summary by CodeRabbit

* **New Features**
  * Pipeline entries now include an optional location column when location is available, appended in a new 4th field for better organization and filtering.
  * If location is missing or blank, entries automatically fall back to the original 3-column format for backward compatibility.

* **Documentation**
  * Updated pipeline documentation to explicitly define the pending-line format and describe how location-aware vs. legacy lines are handled.

* **Tests**
  * Expanded pipeline formatting and location-filter coverage to verify correct 3- vs 4-column output and safe row shaping.

PR #3566, opened 2026-08-31
fix(web): readInbox returns an empty inbox on a CRLF data/pipeline.md

andrgavrilenko:
## What does this PR do?

`readInbox()` in `web/src/lib/career-ops.ts` parses **zero** rows out of a `data/pipeline.md` that uses CRLF line endings. It does not throw and it does not warn — `GET /api/pipeline` simply answers `"inbox": []` for a file full of pending jobs, so the web app shows an empty pipeline that is not empty.

This PR splits career-ops data files on the line break (`/\r?\n/`) instead of on `"\n"`, and adds a CRLF/LF parity regression test for the pipeline read path.

Same failure mode as #2540 (`BULLET_LINE_RE` extracting zero skills from CRLF JDs), one reader over. That one was in the core; this one is in `web/`.

## Related issue

No issue — the template exempts bug fixes, so I am sending it straight in. Happy to open one first if you would rather have it tracked.

---

### Symptom

On a CRLF `data/pipeline.md`:

- `GET /api/pipeline` → `"inbox": []`
- `rootExists` → `true`
- `applications` → parsed correctly, full list
- no error, no warning, no low-confidence marker anywhere

The dashboard renders as "nothing pending". Nothing in the response distinguishes this from a genuinely empty pipeline.

### Root cause

`readInbox()` split the file on `"\n"` and matched each line with:

```js
/^\s*-\s*\[([ xX])\]\s*(.+)$/
```

Splitting on `"\n"` leaves a trailing `\r` on every line of a CRLF file. Two facts then combine, exactly as in #2540:

1. JavaScript counts `\r` as a **line terminator**, and `.` never matches a line terminator. So `(.+)` consumes the row body and then **stops before the `\r`**.
2. The regex has no `m` flag, so a bare `$` asserts **end of string**, not end of line — it will not match in front of a `\r`.

The match therefore has to succeed at a position that still has a `\r` after it, which `$` refuses. The regex fails on **every** row, `readInbox()` falls through to `return jobs` with `jobs` still empty, and the caller gets `[]`.

Nothing downstream can tell the difference: an empty array is the same shape as a legitimately empty inbox.

### Measured

On a real 106-line `pipeline.md` that had been written with CRLF:

| file | rows parsed by `readInbox()` |
|---|---|
| as found (CRLF) | **0** |
| same bytes, CRLF → LF | **62** |

Also verified end-to-end against `main` at `5ad6133` by importing the real `readInbox()` and pointing `CAREER_OPS_ROOT` at a two-row fixture: LF → 2 rows, CRLF → 0 rows. After this patch: 2 and 2.

### Why this is worse than a normal parse bug

**1. The failure is silent and it is plausible.**
`readApplications()` is unaffected — `parseApplications` in `tracker-table.mjs` does `raw.trim()` on every line, so it is CRLF-tolerant by accident. `rootExists` stays `true`. So the API answers with a healthy-looking payload in which only the inbox is empty. The natural reading is "I have no pending jobs", not "the parser died". A user has no thread to pull on.

**2. The writer entrenches it into something even more convincing.**
`appendToPipeline` in `scan.mjs` writes its rows with `"\n"`. Append to a CRLF file and you get mixed endings: the old rows still end `\r\n`, the new ones end `\n`. `readInbox()` then matches **only the freshly appended rows**. The inbox is no longer empty — it shows exactly the jobs from the last scan and hides everything older. That looks like correct, working software right up until you go looking for a job you saved last week.

**3. Git does not normalize this file, and the mixed state is observable.**
The repo-wide policy is real and correct:

```
* text=auto eol=lf
```

but it governs **tracked** files, and `.gitignore` excludes the whole data layer (`data/*`, minus the `.gitkeep`). So `data/pipeline.md` is never normalized by git; its line endings are whatever wrote it last.

I want to be careful about how far that generalizes, because @artemtrofymenko checked a real Windows install with `core.autocrlf=true` and found a 2827-line `pipeline.md` that is pure LF, parsing identically before and after this patch. **I am not claiming career-ops writes CRLF, and I am not claiming this is the default Windows experience** — his install is good evidence against both.

What I can show is that the mixed-ending state is real rather than hypothetical. Byte-exact census of the data layer on the install this bug came from:

| file | lines | CRLF | LF-only | lone CR |
|---|---|---|---|---|
| `data/applications.md` | 53 | 52 | 1 | 0 |
| `data/pdf-index.tsv` | 15 | 7 | 8 | 0 |
| `data/scan-runs.tsv` | 4 | 3 | 1 | 0 |
| `data/status-log.tsv` | 18 | 6 | 12 | 0 |
| `data/pipeline.md` | 112 | 0 | 112 | 0 |
| `data/scan-history.tsv` | 55 | 0 | 55 | 0 |

No lone CRs anywhere, so this is CRLF against LF and not a stray-CR artifact. In three of the four mixed files the LF-only lines run contiguously to the end of the file — `status-log.tsv` from line 7 of 18, `pdf-index.tsv` from line 9 of 15, `scan-runs.tsv` on line 4 of 4 — which is precisely the "CRLF body, LF-written tail" shape that case 3 of the test suite pins.

Two honest caveats on that table:

- `pipeline.md` reads as pure LF there because **I converted it myself before filing this**, to unblock my own run. That is why I can quote 0 rows against 62 for it but cannot show its original bytes; the 106-line figure above is from before the conversion.
- I have **not** determined where the initial CRLF came from. `data/` is untracked, so there is no history to consult.

So the scope of this fix is: a state that demonstrably occurs, at a frequency nobody in this thread can put a number on. A parser that silently returns an empty list for a file full of rows should survive the line ending either way, and `/\r?\n/` is the cheap end of that.

### Reproduce

Take any populated `data/pipeline.md` and flip it between the two encodings, changing nothing else:

```bash
# LF
node -e "const fs=require('fs'),p='data/pipeline.md';fs.writeFileSync(p,fs.readFileSync(p,'utf8').replace(/\r\n/g,'\n'))"
curl -s localhost:3000/api/pipeline | node -e "let s='';process.stdin.on('data',d=>s+=d).on('end',()=>console.log('inbox rows:',JSON.parse(s).inbox.length))"
```

```bash
# same bytes, CRLF
node -e "const fs=require('fs'),p='data/pipeline.md';const lf=fs.readFileSync(p,'utf8').replace(/\r\n/g,'\n');fs.writeFileSync(p,lf.replace(/\n/g,'\r\n'))"
curl -s localhost:3000/api/pipeline | node -e "let s='';process.stdin.on('data',d=>s+=d).on('end',()=>console.log('inbox rows:',JSON.parse(s).inbox.length))"
```

Observed on `main`:

```
inbox rows: 62
inbox rows: 0
```

Identical content, identical rows — only the line endings differ.

---

### Shape

One shared helper, splitting on the line break rather than on the line-feed byte:

```js
export function splitLines(text) {
  return text.split(/\r?\n/);
}
```

used by both `readInbox()` and `readScanDates()`. **The row grammar is untouched** — same checkbox regex, same positional columns, same labeled-segment handling, same `posted:` extraction. The only behavioral change is that a `\r` no longer reaches the matcher.

### Why the parser moved into a `.mjs`

`web/src/lib/pipeline-table.mjs` is new, and it is plain `.mjs` for exactly the reason `tracker-table.mjs` states in its own header:

> Plain .mjs (same pattern as clean-chips.mjs) so tracker-columns-tests.mjs can import it directly under Node and regression-test the REAL alias chain.

`career-ops.ts` imports `@/lib/core/safe-write`, `js-yaml` and other Next-resolved specifiers, so `node --test` — which `web/package.json` runs over `tests/**/*.test.mjs` — cannot import it at all. **That is why this read path had no test in the first place**, and why a bug this total could sit in it. Moving the pure text→rows function into a sibling `.mjs`, right next to the existing `tracker-table.mjs` that does the same job for `applications.md`, makes the parser testable with the runner you already have and no new tooling.

`career-ops.ts` keeps the `InboxJob` type and the fs/root resolution; it now reads:

```ts
export function readInbox(): InboxJob[] {
  const md = read("data/pipeline.md");
  if (!md) return [];
  return parseInbox(md);
}
```

Net effect on `career-ops.ts`: 7 insertions, 43 deletions — the deleted lines are the parser body, moved verbatim.

### Audit of the neighbouring readers

I checked the rest of the read path rather than patching only the one that broke, so the next person writing a reader here has a convention to copy:

| reader | line handling | status |
|---|---|---|
| `readInbox` (`career-ops.ts`) | `split("\n")` + `$`-anchored row regex | **broken** — fixed here |
| `readScanDates` (`career-ops.ts`) | `split("\n")`, survives only because the field is `.trim()`ed later | tolerant **by accident** — moved to `splitLines` so it stays tolerant if that `trim` ever moves |
| `parseApplications` (`tracker-table.mjs`) | `split("\n")` + `raw.trim()` per line | genuinely tolerant — **left alone** |
| `resolveEvalModeFile` (`career-ops.ts`) | `split("\n", 1)[0]` + `EVAL_MODE_TITLE_RE` | tolerant, the regex is unanchored — **left alone** |

Core-side, `PIPELINE_CHECKBOX_RE` in `scan.mjs` is a prefix match with no `$`, so it is not exposed to this class of bug. I did not touch the core.

### Alternative, if you want a smaller diff

The surgical equivalent is `\r?$` in the row regex — the exact shape #2540 landed with in `jd-skill-gap.mjs`. Say the word and I will reduce this PR to that one-line change.

I went the other way deliberately: `\r?$` fixes this one regex and leaves the same trap for the next reader that splits on `"\n"`, and it does not make the read path testable, so the regression test here would have to be dropped along with it. Your call — I will follow whichever you prefer.

---

## Tests

New suite: `web/tests/lib/pipeline-crlf.test.mjs`, importing straight from `pipeline-table.mjs` so it exercises the production parser.

| test | what it pins |
|---|---|
| `CRLF pipeline.md parses to the same rows as LF` | `deepEqual` parity on a 3-row fixture — the headline invariant |
| `no field carries a stray carriage return` | a `\r` must not survive into `url` / `company` / `role` / … even if a future matcher does start accepting it |
| `mixed endings parse — an LF-written row appended to a CRLF file` | CRLF body plus one appended LF row → all 4 rows, not just the new one. The mechanism is `appendToPipeline` emitting `"\n"`; the state itself is the one measured in the census above |
| `row fields survive CRLF: last column and labels are not truncated` | a 3-column row's last cell, the 4th column (#1015), the 5th column (#1017), the `posted:` label, and the `[x]` done flag |
| `splitLines handles LF, CRLF and a lone CR-less tail` | the helper itself, including the trailing-newline and empty-string edges |

**The suite is not vacuous:** reverting `splitLines` to `split("\n")` and re-running fails 4 of the 5 tests. I checked that explicitly rather than assuming it.

### Verification runs

| check | result |
|---|---|
| `npm test` (in `web/`) | 419 tests, 0 fail (the pass/skip split moves between runs — I have seen 418/1 and 413/6 — but nothing fails) |
| `npm run typecheck` (in `web/`) | clean |
| `node test-all.mjs` (repo root) | 7287 pass, **1 fail** — see below |
| end-to-end `readInbox()` against a CRLF fixture | 0 rows before, 2 rows after |

The one `test-all.mjs` failure is `SKILL.md: missing cwd-independent routing rule` across the seven agent skill directories. It reproduces **identically on pristine `main`** in this same Windows checkout — same 7287 / 1 counts, same message — so it predates this branch and is unrelated. I did not touch it.

For the web discovery contract (#2360): the new suite is picked up by the existing `tests/**/*.test.mjs` glob, and `test-all.mjs`'s "all N web suites" assertion passes with the added file.

---

## Type of change

- [x] Bug fix
- [ ] New feature

The extraction into `pipeline-table.mjs` is a move, not a rewrite — the parser body is unchanged apart from dropping the TypeScript annotations.

## Checklist

- [x] Bug fix — exempt from issue-first per the template
- [x] My PR does not include personal data (CV, email, real names, scan results, or pipeline data). The test fixture is invented (`example.com`, Acme / Globex / Initech / Umbrella); the 106-line / 62-row figures above are counts only, no content from that file appears anywhere in the diff.
- [x] I ran `node test-all.mjs` — everything passes except the one pre-existing failure documented above, which also fails on `main`
- [x] My changes respect the [Data Contract](https://github.com/santifer/career-ops/blob/main/DATA_CONTRACT.md) — this is a read path only. No user-layer file is written, and the fix deliberately does **not** normalize the user's `pipeline.md` on disk.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pipeline data parsing for files with Windows-style and mixed line endings.
  * Prevented stray carriage-return characters from appearing in parsed job details.
  * Improved handling of labeled fields such as posting dates.
  * Ensured valid entries are consistently detected when new rows are added.


artemtrofymenko (commented):
Reviewed by reproducing rather than reading, on Windows 11 with `core.autocrlf=true` — and I have one confirmation and one correction to offer.

## The test discriminates, and against the right baseline

Reverting `splitLines` to `text.split("\n")`, which is exactly what `readInbox` does on `main` (`career-ops.ts`, `for (const line of md.split("\n"))`):

```
not ok 1 - CRLF pipeline.md parses to the same rows as LF
not ok 3 - mixed endings parse — the writer appends LF rows to a CRLF file
not ok 4 - row fields survive CRLF: last column and labels are not truncated
not ok 5 - splitLines handles LF, CRLF and a lone CR-less tail
# pass 1  # fail 4
```

Four of five red, and the one that stays green is the LF-parity case, which should pass either way — so the suite has its own control built in. The mechanism is sound by construction too: your row pattern ends `(.+)$`, `.` excludes `\r`, and `$` without the `m` flag will not match with a carriage return still ahead of it, so a CRLF row is skipped whole rather than parsed with a dirty last field. Zero rows is the correct description of the symptom.

## The correction: I could not reproduce it from a real install, and I think the severity claim needs narrowing

I went looking for the bug in the wild rather than in a fixture, on the configuration where you would most expect it — Windows, `core.autocrlf=true`, a 2827-line `data/pipeline.md` from months of real use:

```
CR bytes in the file:  0
LF bytes:              2826
checkbox rows parsed:  2772   (identical before and after your fix)
```

The file is pure LF. `.gitattributes` is not what saves it — `* text=auto eol=lf` governs tracked files and `data/pipeline.md` is untracked user data — it is that career-ops writes the file itself, and its writers emit `\n`.

I got this wrong on the first pass and want to flag the shape of the mistake, since it is the one this PR is about. My first probe reported "2826 of 2826 lines carry CR", which I nearly put in this comment. The pattern had collapsed to empty, so it matched every line: the CR count came out exactly equal to the line count, and *that* equality was the tell I ignored.

So the open question is where a CRLF `data/pipeline.md` comes from, and it is worth a line in the PR body because it decides who this protects. A user editing the file in a CRLF editor, a plugin or an external writer, a Windows tool round-tripping it? Your test name for case 3 — *"the writer appends LF rows to a CRLF file"* — suggests you have seen the mixed-ending state in the wild, which would be the strongest thing you could put in the description.

None of that argues against merging. A parser that silently returns an empty list for a file full of rows should be robust to the line ending whether or not it is common, and `/\r?\n/` is the cheap end of that. It is the *"every Windows user sees an empty pipeline"* reading that I cannot support from here, and I would rather say so than let my earlier bad number support it.

## Small note on the refactor

Moving the parse out of `career-ops.ts` into `pipeline-table.mjs` means the `.mjs` is now the single definition and the `.ts` imports it. That is the same shape as `lib/mjs-files.mjs` in the harness, and for the same reason — two copies of one rule is the drift that produces bugs like this one. Worth keeping the comment at the top of the new file saying so; it is the part a future reader needs.

PR #3567, opened 2026-08-31
feat(providers): add Garena careers provider

kuan0808:
## What does this PR do?

Adds a single-company provider for Garena's public careers API
(`careers.garena.com`), following the shape of `providers/dassault.mjs` /
`providers/ibm.mjs`. It's zero-auth and public: `POST
/api/job/list?office=<office>` with an empty JSON body, no login, returning
job id/title/location tags/description.

Verified live before opening this PR: the endpoint 200s with `{}` (curl),
returns 98 postings today, and — unusually among the single-company
providers — `office` does not filter the result set (every value I tried,
including a made-up one, returned the identical list). It's still sent
because it's the real request shape, and it IS meaningful for the per-job
page URL (`/<office>/careers/<id>`), confirmed by diffing the SSR'd
`<title>` tag between a real job id and a bogus one.

`detect()` is a fixed-host check on `careers.garena.com`; `fetch()` goes
through the shared `ctx.fetchJson` with `redirect:'error'` (SSRF guard),
same as every other provider.

## Test coverage

`tests/providers/garena.test.mjs`: id, `detect()` (positive/host-spoof
rejection/non-string input), `parseGarenaResponse()` (happy path, HTML→text
description, multi-location join, drops title-less/id-less entries, throws
on a shape that doesn't carry `jobs[]`), and `fetch()` against a mock ctx
(request shape, default vs. custom `office` reaching both the API query
string and the job URL).

## Open question

Does a single-employer provider like this need a `docs/SOURCE_INDEXING_LOG.md`
entry? My reading of `CONTRIBUTING.md`'s Source Indexing Policy is that it's
scoped to third-party-operated boards/aggregators (the two logged entries
are both that), and no existing single-employer provider (Dassault, IBM,
Rheinmetall, Heckler & Koch) has a log entry — but I couldn't find a
maintainer ruling on this exact case, so raising it here rather than
deciding silently.

## Type of change

- [x] New feature

## Checklist

- [x] New zero-auth scanner provider — exempt from issue-first per the PR template
- [x] No personal data
- [x] `node test-all.mjs` passes
- [x] No user-layer files touched (`portals.yml` untouched; only
      `templates/portals.example.yml`)

## Summary by CodeRabbit

* **New Features**
  * Added support for collecting job listings from Garena Careers.
  * Garena listings include job titles, links, company, location, and available descriptions.
  * Added configuration examples, including optional office selection with a default of `global`.
* **Documentation**
  * Updated the supported job boards documentation with Garena setup and provider details.


Scott-Emberson (commented):
Reviewed at head 05bf55c1d2. The Garena provider test is concrete, not a smoke check, and the SSRF surface is handled the codebase-standard way. detect() matches only on `new URL(url).host === HOST`, and the test pins that with real spoof rejections: a path-spoof (`evil.com/careers.garena.com`), a suffix-spoof (`careers.garena.com.evil.com`), a non-string, and a missing url are all rejected. The provider fetches a hardcoded API_URL/HOST (never user-derived, the office only lands in the query string and job-URL path) and sets `redirect: 'error'`, and the fetch mock asserts the exact POST url, empty `{}` body, the redirect setting, and office propagation. Parsing is pinned with exact values: two-job extraction, the built URL for default and custom office, company/location mapping with the multi-location join, HTML strip, description-omit-on-empty, dropping title/id/whitespace-only rows, and a throw on a missing jobs array. A parsing regression reddens it. The shared ctx.fetchJson IP guard is not re-exercised here, which is correct since it lives in the shared helper, not this provider.

This is a net-new provider, so it is more than a tests/ change, but from the tests/ side the coverage is sound. Holding formal approval until the branch is finalized, nothing blocking from the tests/ side.

Scott-Emberson (commented):
Re-checked at head d906e0ff5b. My earlier verification stands: tests/providers/garena.test.mjs is byte-identical to what I reviewed (it only reappeared in the compare as merge noise from the large main merge), so the detect() spoof rejections (path-spoof, suffix-spoof, non-string), the exact parse values, the redirect-is-error plus empty-JSON POST body, and the drift-throw are all the assertions I already confirmed.

For the maintainer: this and PR #3586 (chipoto69) both add the same two new files, providers/garena.mjs and tests/providers/garena.test.mjs. #3586 is a superset of this PR, it adds path-segment hardening (rejecting dot and dot-dot in office/id and URL-encoding those segments) with three matching tests. Both are independently mergeable, so only one can land (add/add collision). #3586 strictly dominates on security; the clean options are to merge #3586 over this one, or fold its hardening into this PR. This owned test is sound either way. Holding formal approval until the branch is finalized.

Scott-Emberson (commented):
Re-reviewed at head 4e2b0e026f. Good news on the competing-duplicate front: this PR has now folded in the path-segment hardening that was #3586's only advantage. providers/garena.mjs gains a urlSegment() helper that rejects `.`/`..` outright and URL-encodes everything else, applied to office and id in the URL build and to office in the fetch query string, and the test asserts it with exact percent-encoded output plus dot-loop throw cases that redden if the escaping is dropped. The earlier-verified assertions (detect spoof rejections, exact parse, redirect-is-error, drift-throw) are intact, and this version actually adds two extra fetch-layer escaping assertions, so it is marginally more thorough than #3586, never less.

So the two Garena PRs are now behaviorally equivalent on our owned test and the provider (this one folds the same guard into a single helper; #3586 uses a separate assert plus inline encode). I have already approved #3586 from the tests/ side, so I am not going to also approve this one, since only one of the pair can land (they add the same two files). Both are sound now, and the pick between them is purely which author's version you want. If you would rather land this one, say so and I will move the approval here. Holding formal approval until you pick.

Scott-Emberson (approved):
Approving from the tests/ side. Between the two competing Garena PRs, this is the one to land: the owned test is verified sound at this head (detect spoof rejections, exact parse, redirect-is-error, drift-throw, plus the folded-in path-segment hardening that rejects and encodes unsafe office/id segments), CI is green, and it is mergeable. It is the original author's PR and carries the same security hardening as #3586 with slightly broader fetch-layer coverage, so it is the better one to keep. I am moving my approval here and retracting it from #3586. This routes the code-owner review; it does not merge, and closing the duplicate is the maintainer's step.

PR #3569, opened 2026-08-31
fix(reserve-report-num): handle --help/-h explicitly

Wizarck:
## Summary
- `reserve-report-num.mjs`'s CLI dispatch only special-cases `--release`, `--gc`, and `--count`; any other argument (including `--help`/`-h`) falls through to the default reserve-1-slot path.
- Found this the hard way: a parallel-fanout worker ran `node reserve-report-num.mjs --help` expecting usage text, and instead it silently claimed a real report-number reservation.
- Adds an explicit `--help`/`-h` branch that prints usage and returns `0` before touching any reports/tracker path.

## Test plan
- [x] `node reserve-report-num.mjs --help` / `-h` print usage, exit 0, create no `*-RESERVED.md` sentinel (manually verified)
- [x] New regression test in `tests/reserve-report-num.test.mjs` spawns the CLI with both flags and asserts on stdout + no sentinel
- [x] `node tests/reserve-report-num.test.mjs` — all cases pass
- [x] `node --check` on both changed files

## Summary by CodeRabbit

* **New Features**
  * Added `--help` and `-h` options to display CLI usage information.
  * Help requests now complete successfully without performing other actions.

* **Bug Fixes**
  * Prevented help commands from creating reservation marker files.

### DIFF
--- a/reserve-report-num.mjs
+++ b/reserve-report-num.mjs
@@ -301,6 +301,19 @@ async function runCli() {
   const [,, cmd, arg] = process.argv;
   const options = {};
 
+  if (cmd === '--help' || cmd === '-h') {
+    process.stdout.write([
+      'Usage: node reserve-report-num.mjs [--count <1-N>] [--release <NNN>[-<MMM>]] [--gc]',
+      '',
+      '  (no flags)                Reserve 1 report number (default)',
+      `  --count <1-${MAX_COUNT}>            Reserve N report numbers, printed as a range`,
+      '  --release <NNN>[-<MMM>]  Release a previously reserved number or range',
+      '  --gc                      Garbage-collect stale reservation sentinels',
+      '',
+    ].join('\n'));
+    return 0;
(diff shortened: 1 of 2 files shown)


Scott-Emberson (commented):
Reviewed at head 7a26dce413. The owned test pins the one thing that actually matters for this fix: that --help does not consume a report slot. It loops both --help and -h, spawns the real reserve-report-num.mjs in an empty scratch cwd, and requires all three of exit 0, a `Usage: node reserve-report-num.mjs` line in stdout, and zero `-RESERVED.md` sentinels in reports/. Since the fix short-circuits --help/-h with an early return before the reserve path, a regression that fell through to reserving number 1 would leave a sentinel and trip the count-zero check, so it reddens on exactly the failure this guards. The normal reserve path stays covered by the existing occupancy cases.

Scott-Emberson (approved):

PR #3573, opened 2026-08-31
Fix/build dashboard cli flags

### LINKED ISSUE #3565: feat(cli): four documented scripts have no --help — add the usage block (one script per PR)
**Time:** ~20 min · **Difficulty:** Beginner

Several user-facing scripts documented in `docs/SCRIPTS.md` still have no `--help`, so running one with a wrong flag gives you no way to discover the right one. We already fixed this for `stats.mjs`, `followup-cadence.mjs`, `detect-reposts.mjs` and `check-liveness.mjs`; these four are the ones left:

- [ ] `build-dashboard.mjs`
- [ ] `cv-sync-check.mjs`
- [ ] `normalize-statuses.mjs`
- [ ] `plugin-install.mjs`

### What to do
1. Pick ONE script from the list above.
2. Add a `USAGE` string and a `KNOWN_FLAGS` array, and print `USAGE` and exit 0 when `--help` or `-h` is passed.
3. Unknown flags should exit non-zero with the usage block, rather than being silently ignored.
4. Document the real flags the script already accepts: read the argument handling first, do not invent options.

### Acceptance
- `node <script>.mjs --help` prints the usage block and exits 0.
- `node <script>.mjs --nonsense` prints the usage block and exits non-zero.
- No behaviour change when the script runs normally.
- One script per PR: that keeps review easy and lets four different people take this issue at the same time.

### Pattern to mirror
`stats.mjs` (its `KNOWN_FLAGS` / `USAGE` block near the bottom): that is exactly the shape to copy.

ayushi-wq:
## What does this PR do?
Adds `--help`/`-h` support to `build-dashboard.mjs`, printing a usage block and exiting 0. Unrecognized flags now print the usage block and exit non-zero instead of being silently ignored. No change to normal build behavior.

## Related issue
Closes #3565 

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

---

## Summary by CodeRabbit

* **New Features**
  * Added command-line help output and validation for unsupported options.
  * Scan result files now create missing directories automatically and initialize new files with the correct header.

* **Bug Fixes**
  * Prevented scan module imports from creating a `data/` directory unless a write operation occurs.

* **Tests**
  * Added coverage for nested output directory creation, file initialization, and command-line helper behavior.


Scott-Emberson (changes requested):
Requesting changes. Two blocking issues, one of which is a repeat of a defect we have hit before on this file.

1. scan.mjs does not parse at this head. There are literal git conflict markers left in it (a `=======` at line 90 and another around line 2082, with both sides of the conflict kept), so `node --check scan.mjs` fails with a SyntaxError and the module also ends up redeclaring SCAN_HISTORY_PATH / PIPELINE_PATH / APPLICATIONS_PATH. That is why all three test legs (ubuntu, macos, windows) fail: any suite that imports scan.mjs throws at load. This is a half-resolved merge that needs finishing, delete the conflict markers and confirm `node --check scan.mjs` plus a green test-all before re-pushing.

2. The owned test's new check 6 re-ships the exact blind spot from #3190 / #3370. It spawns `import(scan.mjs)` with `cwd: dir` and asserts `!existsSync(join(dir, 'data'))`, but it does not set `CAREER_OPS_ROOT=dir`. scan.mjs resolves its data paths under DATA_ROOT, which falls back to the repo root when that env is unset, so a real import-time `mkdirSync` would create `<repo>/data` while this assertion only inspects the temp `dir/data` and never sees it. It happens to go red in this PR, but only because the syntax error above makes the spawned import throw into the catch, not because the guard actually detects a side effect. Once the conflict markers are gone, the check reverts to its weak form and can false-pass again. Please set `env: { ...process.env, CAREER_OPS_ROOT: dir }` on that spawn (and ideally also assert on the repo-relative `data/` path) so a stray import-time mkdir lands where the assertion looks.

The build-dashboard.mjs validateFlags change itself looks fine, but it should not merge riding on a syntax-broken scan.mjs. Happy to re-review once scan.mjs parses and check 6 pins the root.

PR #3580, opened 2026-08-31
docs(security): sync GitHub policy scope

dajiaohuang:
## Summary by CodeRabbit

- **Documentation**
  - Expanded the security policy to cover the web dashboard.
  - Clarified that cross-origin requests, same-network hosts, and command injection through the API are in scope.
  - Updated out-of-scope guidance to explain that locally hosted features may still be reachable and reportable.
  - Confirmed that issues reproducible through normal browsing remain within the security policy’s scope.

### DIFF
--- a/.github/SECURITY.md
+++ b/.github/SECURITY.md
@@ -19,6 +19,7 @@ Security issues in the following are in scope:
 
 - **Scripts** (`*.mjs`) — command injection, path traversal, SSRF
 - **Dashboard** (`dashboard/`) — any Go binary vulnerabilities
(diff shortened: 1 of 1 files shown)


luochen211 (approved):
Reviewed the full one-file diff against the current policy structure. The change correctly distinguishes attacks on nonexistent hosted infrastructure from attacks that reach the locally running web server, and it explicitly covers both drive-by cross-origin and same-network access without weakening the existing physical-access or third-party exclusions. `git diff --check` is clean and the full CI matrix is green. No blocking issues found.

PR #3586, opened 2026-08-31
fix: resolve Garena provider portal conflict

chipoto69:
## Summary
- Bridge for contributor PR #3567 because the original branch is currently merge-conflicted (`DIRTY`) against `main`.
- Preserves the Garena provider/docs/tests from `fm/careerops-garena-provider`.
- Resolves the only conflict in `templates/portals.example.yml` by keeping current main examples and adding the Garena portal without duplicating existing entries.

## Test Plan
- `git diff --cached --check`
- `npm run lint`
- `node test-all.mjs --only providers/garena.test.mjs`

## Steward context
- Original PR: https://github.com/santifer/career-ops/pull/3567
- Primary dirty worktree was not touched.
- No force-push used.

## Summary by CodeRabbit

* **New Features**
  * Added Garena as a supported careers provider.
  * Fetches and displays Garena job listings with titles, locations, descriptions, and links.
  * Supports optional office configuration, defaulting to the global office.

* **Documentation**
  * Added Garena to supported provider documentation and configuration examples.


Scott-Emberson (commented):
Reviewed at head 7a7a294f92. The owned test is sound: detect() spoof rejections (path-spoof, suffix-spoof, non-string / missing all to null), exact parse values (the global careers URL, empty description omitted, office honored), the redirect-is-error plus empty-JSON POST body, and a drift-throw when the jobs array is missing. Assertions are exact string equality and require real throws, no tautology, fetch mocked at ctx.fetchJson which is the right unit boundary. This PR also adds the path-segment hardening that its title implies: assertSafePathSegment rejects dot and dot-dot in office/id and the URL builder wraps both in encodeURIComponent, with three matching tests (unsafe office/id URL-encoded, dot-only office rejected, dot-only id rejected).

For the maintainer: this is a competing duplicate of PR #3567 (kuan0808), which adds the same provider and test path and is still cleanly mergeable, so this is not a conflict-resolution bridge for a broken original. The substantive difference is exactly the path-segment hardening above, so this PR strictly dominates #3567 on security. The clean choices are to merge this over #3567, or ask #3567 to fold in the hardening. This owned test is sound. Holding formal approval until the branch is finalized.

Scott-Emberson (approved):
Approving from the tests/ side. The owned test is verified sound at this head (evidence in my review note above), CI is green, and the branch is mergeable. Between this PR and its competing duplicate I am endorsing this one because it is the strict superset on security. Good to go from our end; this routes the code-owner review, it does not merge, and the maintainer picks which of the pair actually lands.

Scott-Emberson (commented):
Withdrawing my endorsement of this PR in favor of #3567. The two are competing duplicates that add the same Garena provider and test, and now that #3567 has folded in the same office/id path-segment hardening this PR introduced, they are behaviorally equivalent on our owned surface (with #3567 carrying slightly broader fetch-layer coverage). Only one can land, so I have moved my approval to #3567 as the original author's version. No defect here, purely the duplicate-resolution call, and thanks for the hardening work, it is what pushed #3567 to fold the same guard in. Recommend closing this in favor of #3567.

Note: I tried to formally dismiss my earlier approval via the API but that is 404ing after the repository move, so the stale approval marker may persist until this PR is closed. The owned test itself remains sound.

PR #3589, opened 2026-08-31
test(scan): pin that stats.mjs reads the scan runs scan.mjs writes

abankar1:
**Rescoped.** This opened as a fix for the blacklist and scan-runs paths; #3511 landed those first and went further. The code changes here were redundant, so I reset the branch to main and kept only the one guarantee that is still missing.

## What #3511 covers, and the half it cannot

#3511 pins this well, and its decoy-in-both-directories arrangement is a better test than the one I had written. Its scan-runs check asserts where the **writer** puts the file:

```js
existsSync(runsAtRoot) && !existsSync(runsAtCwd)
  ? pass('appendScanRunSummary() writes under the data root, where stats.mjs reads')
```

The comment says "where stats.mjs reads", but nothing asks stats.mjs. Move its `SCAN_RUNS_FILE` and the suite stays green:

```
stats.mjs:  const SCAN_RUNS_FILE = join(DATA_ROOT, 'data', 'scan-runs-MOVED.tsv');

  ✅ loadBlacklist() reads the data root's blacklist, not the cwd's
  ✅ appendScanRunSummary() writes under the data root, where stats.mjs reads
  ✅ loadCheckpoint() resumes from the data root's checkpoint
  ✅ scan.mjs exports pipeline/portals paths anchored to the data root
  ✅ scan-interamt.mjs runs its scan behind a main guard, like its siblings
  ✅ no scanner declares a bare cwd-relative user-data path
  -> 0 failures
```

Reasonably so — from the writer's side nothing changed. But the two constants are resolved independently in different files, so either can drift without the other noticing, and when they disagree the failure is invisible: an absent counter file is indistinguishable from never having scanned, so `stats` reports empty scan-run trends exactly as it would for a fresh install.

## The test

A round trip through both public entry points on their **default** paths — `appendScanRunSummary()` then `computeAllStats()` — rather than a comparison of two module constants. The constants are private, the defaults are what can drift, and a round trip survives either side being rewritten. cwd and data root are pointed at different directories, matching #3511's arrangement.

Coverage is complementary, measured both ways:

| | this suite | #3511's |
|---|---|---|
| reader-side drift (`stats.mjs`) | **fails** | passes |
| writer-side drift (`scan.mjs`) | fails | fails |

One test, 82 lines, no production change. Full suite green at 8037.

Credit where it is due: #3511 fixed more than I had found — `scan-ats-full`, `scan-hn`, `scan-interamt`, portal-health, and a structural guard against the next bare-relative path. Its check 6 in particular is the thing I would have asked for.

If you would rather fold this single assertion into `tests/scan-data-paths-under-data-root.test.mjs` so the scan path coverage stays in one file, say so and I will move it there instead — or close this, which is a perfectly fine outcome now that the bug itself is fixed.

### DIFF
--- a/tests/scan-runs-reader-agreement.test.mjs
+++ b/tests/scan-runs-reader-agreement.test.mjs
@@ -1,0 +1,82 @@
+// tests/scan-runs-reader-agreement.test.mjs — data/scan-runs.tsv has one writer
+// and one reader, and the pair has to agree (#1604, #3510).
+//
+// #3511 anchored the scanners' user data to the data root and pinned it well,
+// including that appendScanRunSummary() writes under the data root. This adds
+// the half that check cannot make: it asserts where the WRITER puts the file,
+// never asking stats.mjs where it actually looks. Move stats.mjs's own constant
+// and tests/scan-data-paths-under-data-root.test.mjs stays green — verified —
+// because from the writer's side nothing has changed.
+//
+// That is worth its own assertion because the two sides are resolved
+// independently, in different files, and the failure is invisible: an absent
+// counter file is indistinguishable from never having scanned, so `stats`
+// reports empty scan-run trends exactly as it would for a fresh install.
+//
+// Written as a ROUND TRIP through both public entry points on their DEFAULT
+// paths, so it keeps holding if either side's resolution is rewritten again.
+//
+// Run:  node --test tests/scan-runs-reader-agreement.test.mjs
+
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { spawnSync } from 'node:child_process';
+import { mkdtempSync, mkdirSync, rmSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { dirname, join } from 'node:path';
+import { fileURLToPath, pathToFileURL } from 'node:url';
+
+const ROOT = dirname(dirname(fileURLToPath(import.meta.url)));
+const url = (rel) => JSON.stringify(pathToFileURL(join(ROOT, rel)).href);
+
+const COUNTERS = {
+  timestamp: '2026-09-02T09:00:00.000Z', status: 'completed',
+  companies: 12, boards: 12, found: 40,
+  filteredTitle: 3, filteredTier: 1, filteredLocation: 2, filteredPostingAge: 0,
+  filteredSalary: 0, filteredContent: 1, filteredCooldown: 0,
+  dupes: 5, newAdded: 28, errors: 0,
+  filteredBlacklist: 0, filteredVisa: 0, filteredPostedDate: 0,
+  filteredCountryEligibility: 0,
+};
+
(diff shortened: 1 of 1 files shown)


Scott-Emberson (commented):
Reviewed at head b5b346588e. This is a real fail-open fix and the tests pin it where it matters. BLACKLIST_PATH and SCAN_RUNS_PATH are now anchored to DATA_ROOT (getCareerOpsRoot(), which honors CAREER_OPS_ROOT) rather than a cwd-relative string. scan-blacklist-path.test.mjs spawns a child from a foreign temp cwd with CAREER_OPS_ROOT pointed at a separate sandbox holding a blacklisted "Acme Corp", calls loadBlacklist() on its default path, and asserts size is 1, so a regression back to cwd-relative resolution makes existsSync false from the foreign cwd, yields an empty Map, and reddens. That empty Map is exactly the fail-open (do-not-apply companies silently scanned), so this is the right thing to guard. scan-runs-path-agreement.test.mjs round-trips the real append/read entry points and additionally asserts the file lands under DATA_ROOT and not under cwd, which kills the degenerate "both sides agree on the wrong cwd file" pass.

Owned tests verified sound, including the security-relevant fail-open guard. Holding formal approval until the branch is finalized, nothing blocking from the tests/ side.

Scott-Emberson (approved):

luochen211 (approved):
Reviewed the code-side path separately from the existing tests/ approval. Both values now derive from the same resolved `DATA_ROOT` used by their consumers: `loadBlacklist()` can no longer silently read a cwd-local absence, and `appendScanRunSummary()` writes the file `stats.mjs` reads. The deliberate cwd-relative `PORTAL_HEALTH_PATH` remains untouched, so the change does not broaden writes back into the checkout. The focused tests exercise distinct cwd/data-root directories, `git diff --check` is clean, and the full CI matrix is green. No blocking code issue found.

Scott-Emberson (commented):
Re-reviewed at head d8ccbb91ec (moved past the earlier approval, so this wants a fresh approval). The PR was reworked, sensibly, around what landed in main. Main now carries #3510's consolidated scan-data-paths-under-data-root.test.mjs, which already holds the blacklist-from-a-foreign-cwd and writer-location checks I approved here (in fact stronger, it uses a cwd decoy), so this PR is narrowed to the one assertion that suite does not make: a real reader round trip proving stats.computeAllStats() sees what scan.appendScanRunSummary() writes, run from a decoy cwd on the default paths through the real modules. No coverage was lost, the earlier blacklist and writer checks live on in base main, and there is no false-pass: the reader test alone could pass if both sides were cwd-relative, but that exact agree-on-the-wrong-file gap is independently closed by the consolidated check in base main (its header names this).

Owned test verified sound, the rework preserves coverage and adds the reader-agreement half. Holding formal approval until the branch is finalized (fresh approval needed since the head moved), nothing blocking from the tests/ side.

PR #3593, opened 2026-08-31
fix(plugins): a YAML typo in plugins.yml destroyed the whole config (#3592)

### LINKED ISSUE #3592: plugins.mjs: a YAML typo in config/plugins.yml destroys the whole file
`setEnabled` states its contract in the comment directly above it:

> Write enabled:true/false into config/plugins.yml, **merging (never clobbering the user's other plugins or non-secret settings)**.

The read is wrapped in a bare `catch {}`:

```js
plugins.mjs:204-213
function setEnabled(id, on, settings) {
  const file = path.join(ROOT, 'config', 'plugins.yml');
  let cfg = {};
  if (existsSync(file)) { try { cfg = yaml.load(readFileSync(file, 'utf8')) || {}; } catch {} }
  if (!cfg.plugins || typeof cfg.plugins !== 'object') cfg.plugins = {};
  ...
  writeFileSync(file, '# career-ops plugin activation ...\n' + yaml.dump(cfg), 'utf8');
}
```

On a parse error `cfg` stays `{}`, and the write puts that empty object back over the file.

### Repro

`config/plugins.yml` with two plugins, a stored setting, and one malformed line:

```yaml
plugins:
  apify:
    enabled: true
    settings:
      dataset: KEEP-ME
  teamtailor:
    enabled: true
  : this line is malformed
```

After `node plugins.mjs enable <anything> --confirm`:

```yaml
plugins:
  newplugin:
    enabled: true
```

Both plugins gone, the stored setting gone, nothing printed.

### Why this is expensive rather than annoying

- The trigger is a **YAML typo** — the most likely reason a hand-edited config does not parse, and precisely when the file most needs not to be replaced.
- `config/plugins.yml` is a **USER path** in `update-system.mjs:447` (never auto-updated), so there is no copy to restore from.

### The same swallow is in doctor.mjs

```js
doctor.mjs:558
try { return yaml.load(readFileSync(cfgPath, 'utf8')) || {}; } catch { return {}; }
```

An unreadable config and a config with nothing enabled produce the same empty object, so `doctor` — the tool whose job is to say what is wrong, and the one the agent runs on the first message of every session — reports every enabled plugin as "off" and says nothing about why.

### Separate finding, not fixed here

`plugins.mjs:34` is `const ROOT = path.dirname(fileURLToPath(import.meta.url))` — the **code** root. `doctor.mjs` resolves the same file via `getCareerOpsRoot()`. On a non-default data root the two never see the same `config/plugins.yml`. That is a path/migration question rather than a data-loss one, so I have left it out of the fix; happy to file it separately.

PR to follow.

abankar1:
Fixes #3592.

`setEnabled`'s own comment states the contract — *"merging (never clobbering the user's other plugins or non-secret settings)"* — and the code did the opposite. The read was wrapped in a bare `catch {}`, so on a parse error `cfg` stayed `{}` and the write two lines later put that empty object back over the file.

```
BEFORE                        AFTER `plugins.mjs enable <new> --confirm`
  apify:                        plugins:
    enabled: true                 newplugin:
    settings:                       enabled: true
      dataset: KEEP-ME
  teamtailor:
    enabled: true
```

Both plugins gone, the stored setting gone, nothing printed.

Two things make it expensive rather than annoying. The trigger is a **YAML typo** — the most likely reason a hand-edited config does not parse, and precisely when the file most needs not to be replaced. And `config/plugins.yml` is a **USER path** in `update-system.mjs:447`, so there is no copy to restore from.

## Absent is not unreadable

No file still means a first enable — that is the whole point of the function — and an empty or comment-only file is the same. Only an existing file that will not parse is refused.

A scalar or a list is refused too. Those parse cleanly and are still not a config, so spreading one into the write would discard it exactly as silently as the empty object did.

## `cmdRemove` reports instead of swallowing

`try { setEnabled(id, false); } catch {}` used to catch nothing worth seeing. It can now catch a real, actionable error, and a user told "disabled" while the config still says `enabled: true` has been told the wrong thing. The removal itself still cannot be failed by a config problem — the files and the lock entry are already gone by then.

## The second commit: doctor said "nothing enabled"

`doctor.mjs:558` carried the same swallow, so an unreadable config and a config with nothing enabled produced the same empty object. The one tool whose job is to say what is wrong reported every enabled plugin as "off", with no mention of why — and it is the tool the agent runs on the first message of every session.

So the two halves pair up: `plugins.mjs` declines to overwrite a config it cannot parse, and `doctor` tells you which config and what is wrong with it.

`pluginConfigError` is added to `--json` **only when it happened**, so a healthy run carries no new key and the field's presence means one thing. The `--json` builder was the second caller of `readPluginConfigSync` and is updated for the new return shape — missing it would have passed the wrapper object where the config was expected, and every plugin would have read as off for the opposite reason.

## Tests

The guard is split into an exported `parsePluginConfig` so it is testable without driving the CLI at a real config file: `setEnabled` resolves its path from the **script's** directory, so a CLI-level test would have to write into the real checkout to exercise anything. (I found that the hard way — an earlier draft of these tests set `CAREER_OPS_ROOT` and was passing vacuously, because `enable` needs `--confirm` and nothing was being written at all.)

Eight tests: the refusal and its message, a direct pin that `{}` is no longer a possible return for unparseable input, the absent/empty/comment-only paths, non-mapping YAML, and the two doctor cases. Reverting either half reddens only its own tests — 2 of 8 for the `plugins.mjs` guard, 1 of 8 for the doctor half.

Full suite green at 7852.

## Separate finding, deliberately not fixed here

`plugins.mjs:34` is `const ROOT = path.dirname(fileURLToPath(import.meta.url))` — the **code** root — while `doctor.mjs` resolves the same file via `getCareerOpsRoot()`. On a non-default data root the two never see the same `config/plugins.yml`. That is a path/migration question rather than a data-loss one, and mixing it in would change where the file lives in a PR about not destroying it. Happy to file it separately.

## Summary by CodeRabbit

- **Bug Fixes**
  - Plugin configuration errors are now clearly reported instead of being treated as empty settings.
  - Invalid configuration files are protected from accidental overwrites during plugin changes.
  - Plugin removal now reports update failures and only confirms success when changes are saved.
  - Diagnostic JSON output identifies configuration parsing errors while remaining unchanged for valid configurations.

- **Tests**
  - Added coverage for malformed, empty, absent, and valid plugin configuration files.

### DIFF
--- a/doctor.mjs
+++ b/doctor.mjs
@@ -552,10 +552,20 @@ function checkPipelineFile() {
 
 // Discover plugins + their non-secret config block, synchronously. Used by both
 // the human check and the --json onboarding state.
+// A parse failure is REPORTED, not folded into {}. An unreadable config and a
+// config with nothing enabled produced the same empty object, so doctor — the
+// one tool whose job is to say what is wrong — answered "off" for every plugin
+// the user had actually switched on, and said nothing about why.
+//
+// Returns the config plus the parse error, so callers can tell the two apart.
 function readPluginConfigSync(root) {
   const cfgPath = join(root, 'config', 'plugins.yml');
-  if (!existsSync(cfgPath)) return {};
-  try { return yaml.load(readFileSync(cfgPath, 'utf8')) || {}; } catch { return {}; }
+  if (!existsSync(cfgPath)) return { cfg: {}, error: null };
+  try {
+    return { cfg: yaml.load(readFileSync(cfgPath, 'utf8')) || {}, error: null };
+  } catch (err) {
+    return { cfg: {}, error: String(err.message).split('\n')[0] };
+  }
 }
 
 // Plugin layer health: list discovered plugins + whether each enabled one's keys
@@ -564,7 +574,17 @@ function checkPlugins(root) {
   let manifests;
   try { manifests = discoverPlugins(pluginRoots(root)); } catch { return { pass: true, label: 'Plugins: none' }; }
   if (manifests.length === 0) return { pass: true, label: 'Plugins: none installed' };
-  const cfg = readPluginConfigSync(root);
+  const { cfg, error: cfgError } = readPluginConfigSync(root);
+  // Reported before the per-plugin lines, because when the config did not parse
+  // every one of those lines is derived from an empty object and says "off"
+  // regardless of what the user configured.
+  if (cfgError) {
+    return {
+      warn: true,
+      label: `Plugins: config/plugins.yml did not parse (${cfgError}) — every plugin below reads as off`,
+      fix: ['Fix the YAML in config/plugins.yml. Until then `plugins.mjs enable/disable` will refuse to write to it.'],
+    };
+  }
   const lines = [];
   const fixes = [];
   for (const m of manifests) {
@@ -770,8 +790,12 @@ function onboardingState(root) {
     : {};
 
   let plugins = [];
+  let pluginConfigError = null;
   try {
-    const cfg = readPluginConfigSync(root);
+    const { cfg, error } = readPluginConfigSync(root);
+    // Travels in the JSON so a consumer can tell "nothing enabled" from "the
+    // config did not parse" — the two used to be the same empty list.
+    pluginConfigError = error;
     plugins = discoverPlugins(pluginRoots(root)).map((m) => {
       const s = pluginStatus(m, cfg);
       return { id: m.id, hooks: m.hooks, enabled: s.enabled, missingEnv: s.missingEnv };
@@ -788,6 +812,9 @@ function onboardingState(root) {
     warnings,
     autoCopied,
     plugins,
+    // Only present when it happened, so existing consumers see no new key on a
+    // healthy run and a broken config is impossible to read as "none enabled".
+    ...(pluginConfigError ? { pluginConfigError } : {}),
     playwright_mcp: playwrightMcp,
     active_cli: activeCli,
     cli_source: cliSource,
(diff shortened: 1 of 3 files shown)


luochen211 (changes requested):
Two current-head blockers remain at `ac699e4d9e95`:

1. `parsePluginConfig()` still treats an explicit YAML null document as an empty file. I checked the repository's declared major with isolated `js-yaml@5.4.1`: both `yaml.load('null')` and `yaml.load('~')` return `null`. The `if (cfg == null) return {}` branch therefore lets the next enable/disable rewrite an existing, non-empty `config/plugins.yml`, which is the same data-loss class this PR is intended to close. Only absent, blank, or comment-only input should take the empty-config path; explicit null documents should be refused and covered by regression tests.

2. `cmdRemove()` passes the CLI-supplied ID through `path.join(ROOT, 'plugins.local', id)` and then into recursive `rmSync()` before validating containment. A pure path reproduction gives `../victim -> /install/victim` and `../../outside -> /outside`; the command dispatcher exposes this directly as `plugins.mjs remove <id>`. Please require one valid plugin ID and prove the resolved/real path remains inside `plugins.local` before deletion, including traversal and symlink-escape coverage.

I could not run the new suite in this minimal checkout because the root dependency tree is absent; I did run `git diff --check`, reproduced the YAML behavior against isolated 5.4.1, and verified the path escape with Node's own `path.join`/`relative`. The closed bridge #3595 contains prior work for these cases, but the fixes need to land on this author branch now that the maintainer selected #3593.

PR #3595, opened 2026-08-31
fix: harden plugin config parsing

chipoto69:
Bridge/follow-up for PR #3593 and issue #3592.

Summary:
- Preserves PR #3593's config-clobber guard.
- Rejects malformed YAML that js-yaml accepts as a null-key mapping, so `: malformed` cannot silently become `plugins.null`.
- Reports scalar/list/null-key plugin configs through `doctor --json` without dropping `pluginConfigError` when plugin discovery fails.
- Treats absent, empty, and comment-only configs as first-enable empty state.
- Blocks `plugins.mjs remove` path traversal before recursive deletion.

Test plan:
- `node --test tests/plugins-config-preservation.test.mjs` PASS (10/10)
- `node test-all.mjs --only plugins-config-preservation` PASS
- `npm run lint` PASS (615 .mjs syntax clean)
- `node test-all.mjs --quick` PASS: 7572 passed / 0 failed / 13 warnings

Steward note: direct normal push to abankar1:fix/plugins-config-clobber was denied despite maintainerCanModify=true, so this is a non-force bridge PR from chipoto69's fork. No dirty primary worktree touched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented enabling or disabling plugins from overwriting malformed configuration files.
  * Added validation for invalid YAML, unsupported configuration formats, and unsafe plugin paths.
  * Improved feedback when plugin configuration updates fail.
  * Added safer handling of malformed or missing plugin settings.

* **New Features**
  * JSON onboarding status now reports plugin configuration errors when applicable.
  * Empty or comment-only configuration files remain supported during initial plugin setup.

### DIFF
--- a/doctor.mjs
+++ b/doctor.mjs
@@ -552,10 +552,20 @@ function checkPipelineFile() {
 
 // Discover plugins + their non-secret config block, synchronously. Used by both
 // the human check and the --json onboarding state.
+// A parse failure is REPORTED, not folded into {}. An unreadable config and a
+// config with nothing enabled produced the same empty object, so doctor — the
+// one tool whose job is to say what is wrong — answered "off" for every plugin
+// the user had actually switched on, and said nothing about why.
+//
+// Returns the config plus the parse error, so callers can tell the two apart.
 function readPluginConfigSync(root) {
   const cfgPath = join(root, 'config', 'plugins.yml');
-  if (!existsSync(cfgPath)) return {};
-  try { return yaml.load(readFileSync(cfgPath, 'utf8')) || {}; } catch { return {}; }
+  if (!existsSync(cfgPath)) return { cfg: {}, error: null };
+  try {
+    return { cfg: yaml.load(readFileSync(cfgPath, 'utf8')) || {}, error: null };
+  } catch (err) {
+    return { cfg: {}, error: String(err.message).split('\n')[0] };
+  }
 }
 
 // Plugin layer health: list discovered plugins + whether each enabled one's keys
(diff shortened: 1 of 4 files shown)


Scott-Emberson (commented):
Reviewed at head 1319404adb. The owned test is sound: it asserts parsePluginConfig throws on unparseable input with a message that names the file and says it is refusing to overwrite, pins the "returning empty is what destroyed the file" regression directly, and confirms a well-formed config round-trips with its settings intact. It also spawns the real doctor.mjs and asserts pluginConfigError present on unparseable and absent on healthy, and exercises the plugin-removal path's traversal / absolute / symlink-escape refusals, so this PR's extra hardening beyond a single-typo fix is actually covered.

For the maintainer: this overlaps PR #3593 (abankar1), which fixes the same #3592 issue with the same throw-instead-of-swallow approach but touches only the test file, plugins.mjs, and doctor.mjs. This PR (#3595) is the broader of the two, it also hardens plugins/_engine.mjs and adds the removal-path security. Both are sound; the choice between them is yours. One narrow scope note on this file: the tests exercise parsePluginConfig at the seam, not setEnabled writing a real file, so a future regression that re-wrapped setEnabled's write in a try/catch would stay green here, worth a follow-up integration assertion. Holding formal approval until the branch is finalized.

Scott-Emberson (approved):
Approving from the tests/ side. The owned test is verified sound at this head (evidence in my review note above), CI is green, and the branch is mergeable. Between this PR and its competing duplicate I am endorsing this one because it is the strict superset on security. Good to go from our end; this routes the code-owner review, it does not merge, and the maintainer picks which of the pair actually lands.

PR #3598, opened 2026-09-01
fix(tests): answer gitignore probes that cross a user-layer symlink

vliggio:
## What

`tests/user-layer-gitignored.test.mjs` has three probe lists. The AGENTS.md user-layer loop handles `git check-ignore`'s exit 128 (`fatal: pathspec '...' is beyond a symbolic link`); the `derivedIndexProbes` list added alongside the new `*.db` / `*.db-wal` / `*.db-shm` rules does not, and fails hard on it.

That exit-128 case is the one #3165 was about, fixed in #3172. The fallback was written inline in the first loop, so the probe list added afterwards didn't inherit it.

## Impact

On the symlinked-user-layer layout (the manual workaround for #524, where `data/` points outside the repo), this puts three assertions permanently red on a checkout where the rules are present and correct:

```
❌ data/applications.db: git check-ignore could not answer — fatal: pathspec 'data/applications.db' is beyond a symbolic link
❌ data/applications.db-wal: ...
❌ data/applications.db-shm: ...
```

Test-only. The `.gitignore` rules work; the index lives outside the repository where git can't reach it either way. But it's the same false negative #3165 called out: noise in the one suite whose whole job is to be trustworthy about PII.

## Change

Lifts the fallback into `checkIgnoreThroughSymlinks()`. When git refuses a pathspec, it asks about the outermost symlink on the way down instead, since that entry is what actually governs whether `git add .` can stage anything through it.

It mirrors the existing loop's policy rather than inventing one: `warn` (not `fail`) when only the link entry is unignored, because staging a link commits its target path, not the user's data. `data/*` doesn't match `data`, so stock rules land there.

The helper is deliberately shared so the next probe list can opt in with one call rather than rediscovering exit 128.

## Verification

Reproduced the layout by pointing `data/` at a directory outside the repo:

| | failures |
|---|---|
| `origin/main` | 3 |
| this branch | 0 |

Full `node test-all.mjs` on the standard layout: **7570 passed, 0 failed** (unchanged).

Related: #3165, #3172, #524.

## Summary by CodeRabbit

* **Tests**
  * Improved validation of ignore rules for files accessed through symbolic links.
  * Added clearer checks and diagnostics for cases where ignore status cannot be determined.
  * Distinguishes successful, warning, and failure outcomes when validating derived indexes.

### DIFF
--- a/tests/user-layer-gitignored.test.mjs
+++ b/tests/user-layer-gitignored.test.mjs
@@ -55,6 +55,50 @@ function isSymlink(rel) {
   }
 }
 
+/**
+ * The outermost symlink on the way down to `rel`, if any.
+ *
+ * git stops at the FIRST symlink it meets walking a pathspec, so that entry --
+ * not the leaf -- is the one it can still answer for. Returns the repo-relative
+ * path of that ancestor (or `rel` itself when the leaf is the symlink), else null.
+ *
+ * @param {string} rel - Repo-relative path.
+ * @returns {string|null}
+ */
+function symlinkedAncestor(rel) {
+  const parts = rel.split('/').filter(Boolean);
+  for (let i = 1; i <= parts.length; i += 1) {
+    const prefix = parts.slice(0, i).join('/');
+    if (isSymlink(prefix)) return prefix;
+  }
+  return null;
+}
+
+/**
+ * checkIgnore, but able to answer for probes that cross a user-layer symlink.
+ *
+ * The loop over the AGENTS.md paths below has resolved this since #3165: when
+ * git refuses a pathspec, ask about the symlinked entry instead, because that
+ * entry is what actually governs whether `git add .` can stage anything through
+ * it -- the contents live outside the repository. That fallback was written
+ * inline, so the derived-index probe list added later did not inherit it and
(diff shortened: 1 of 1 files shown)


Scott-Emberson (commented):
Reviewed at head 5474b70fff. The symlink handling is added without weakening the guard, which is the thing I care about on this file. checkIgnoreThroughSymlinks passes a direct git check-ignore verdict through unchanged (via null) for exit 0 and exit 1, and only falls back to the ancestor walk when the direct probe is unanswerable (exit 128, git refusing because the path crosses a symlink). On the consumer side, a not-ignored result only downgrades to warn when via is set (a real symlinked ancestor exists and the only stageable thing is the link target, mirroring the accepted #3165 loop), a bare not-ignored with via null still fails, and the unanswerable / else branch still fails under any via value. The derived index-probe cross-product (locations by .db / -wal / -shm) is preserved verbatim, so the earlier coverage is intact.

Owned test verified sound, exit-code discrimination and the fail-closed branches preserved. Holding formal approval until the branch is finalized, nothing blocking from the tests/ side.

Scott-Emberson (approved):

PR #3599, opened 2026-09-01
fix(tests): stop the photo suite writing cv-preview.html into the user's data root

vliggio:
## What

`build-cv-html.mjs --preview` ignores its output argument and resolves the destination from the **data root**:

https://github.com/santifer/career-ops/blob/main/build-cv-html.mjs#L740

That line is the file's only use of `DATA_ROOT`. On the default layout the data root is the checkout, so this is invisible. On the external-data-root layout (#524, and the symlink workaround people use for it) it is the candidate's own `output/` directory, somewhere else on the machine entirely.

`tests/profile-photo.test.mjs` calls `--preview` directly, so **running the test suite writes a "Test Candidate" fixture into a real user's CV output directory** — from any checkout, on every run.

I hit this from a throwaway worktree under `/tmp`, which had no business touching the user's data at all: a `cv-preview.html` appeared in their `output/` alongside 64 real tailored CVs, timestamped to that suite run.

## The assertion passes by luck

The test writes via `DATA_ROOT` and reads back via `ROOT`:

```js
readFileSync(preview ? join(ROOT, 'output', 'cv-preview.html') : output, 'utf8')
```

Those are the same file only when the data root *is* the repo. On any other layout the write escapes the sandbox and the read still succeeds against whatever happens to be sitting there — so the test stays green while doing the wrong thing in both directions.

## Change

Point `CAREER_OPS_ROOT` at the render's own `mkdtemp` dir for the preview spawn. The artifact lands there, is read back from there, and the two paths are equal by construction rather than by coincidence.

Scoped to the `preview` branch so the non-preview renders are untouched. Photo resolution uses `__dirname` (line 146) and is unaffected regardless.

Test-only — no change to `--preview`'s real behaviour, which is arguably correct as-is: `output/*` is user layer per the Data Contract, so writing there is the point. The bug is that the *test* uses the real path.

## Verification

With `CAREER_OPS_ROOT` pointed at a stand-in data root:

| | leaked into data root | suite |
|---|---|---|
| `origin/main` | `cv-preview.html` | 8/8 pass |
| this branch | *(empty)* | 8/8 pass |

Full `node test-all.mjs`: **7570 passed, 0 failed** (unchanged).

Related: #524.

## Summary by CodeRabbit

* **Tests**
  * Improved profile photo preview test isolation by using a temporary data directory.
  * Prevented test-generated preview files from being written to the project’s output directory.

### DIFF
--- a/tests/profile-photo.test.mjs
+++ b/tests/profile-photo.test.mjs
@@ -1,7 +1,7 @@
 import { test } from 'node:test';
 import assert from 'node:assert/strict';
 import { execFileSync, spawnSync } from 'node:child_process';
-import { mkdtempSync, readFileSync, writeFileSync, existsSync } from 'node:fs';
+import { mkdirSync, mkdtempSync, readFileSync, writeFileSync, existsSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { dirname, join, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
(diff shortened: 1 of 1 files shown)


Scott-Emberson (commented):
Reviewed at head b13304275a. Good test-hygiene fix. The photo suite now sets CAREER_OPS_ROOT to a temp data-root and reads cv-preview.html back from that same temp path, so it no longer writes into the checkout's real output/ directory. build-cv-html resolves the preview destination under DATA_ROOT from getCareerOpsRoot(), so the redirect is genuinely honored and write and read-back land in the same temp dir. The isolation is self-checking: if the builder ignored the env and wrote to the real root, the temp read-back would throw and redden rather than silently pass. The substantive assertions (image inlining as a data URI, the photo class, square aspect ratio, error-path exit codes, the preview-ready marker and DOCTYPE) are byte-identical, only the render helper changed.

Owned test verified sound, isolation added without gutting any assertion. Holding formal approval until the branch is finalized, nothing blocking from the tests/ side.

Scott-Emberson (approved):

PR #3605, opened 2026-09-01
test: tolerate a missing web/ subproject in the mjs-files syntax gate

### LINKED ISSUE #3563: tests/mjs-files.test.mjs hardcodes 'web/ must be inside the gate', fails on any checkout without the optional web/ subproject
### What happened?

`tests/mjs-files.test.mjs` (added/touched in the v1.31.0 release, #3419) hardcodes an assertion that the syntax-check gate must include files under `web/`:

```js
assert.ok(files.some((f) => f.startsWith('web/')), 'web/ must be inside the gate');
```

`web/` is documented elsewhere in this same repo as an **optional** subproject that may not exist in a given checkout — see #2360 and the passing `test-all.mjs` case right next to this failure:

```
web/ test discovery contract (#2360)
  ✅ web/ is not present in this checkout — discovery contract not applicable
```

So on any install without the `web/` directory, `test-all.mjs` fails here even though nothing is actually wrong:

```
✖ the syntax gate reaches past the repository root (19.8ms)
  AssertionError [ERR_ASSERTION]: web/ must be inside the gate
      at tests/mjs-files.test.mjs:85:10
```

### Steps to reproduce

1. Use a career-ops checkout that has no `web/` directory (any install that never opted into the web UI, or one predating it).
2. `node --test tests/mjs-files.test.mjs`, or `node test-all.mjs`.
3. `tests/mjs-files.test.mjs — node:test suite failed (exit 1)`, specifically the `'the syntax gate reaches past the repository root'` test, `web/ must be inside the gate`.

### Expected behavior

The assertion should tolerate `web/` being absent, the same way the adjacent `web/ test discovery contract (#2360)` check already does — e.g. skip that specific `startsWith('web/')` assertion (or downgrade it to a no-op) when `existsSync(join(ROOT, 'web'))` is false, while still asserting it when `web/` is present. `tests/`, `providers/`, and `lib/` are presumably safe to assert unconditionally since those ship with every install; `web/` is the one directory in that list that's opt-in.

### Environment

- career-ops v1.31.0
- `node -v`: v26.7.0

gunjanjaswal:
Fixes #3563.

## What

`tests/mjs-files.test.mjs`, in the `the syntax gate reaches past the repository root` test, asserts that `web/` is inside the gate with no guard:

```js
assert.ok(files.some((f) => f.startsWith('web/')), 'web/ must be inside the gate');
```

But `web/` is an opt-in subproject (#2360), not something every checkout has. The file right next to this one already knows that: `tests/web-test-layout.test.mjs` runs its whole discovery contract behind `existsSync(WEB_PKG)` and prints `web/ is not present in this checkout — discovery contract not applicable` when it's missing. So on any install that never took the web UI, `node test-all.mjs` goes red here for a directory that was never supposed to be mandatory.

## Fix

Guard the one `web/` assertion behind `existsSync(join(ROOT, 'web'))`, the same conditional the adjacent contract uses. `tests/`, `providers/` and `lib/` stay unconditional, since those ship with every install and are the real invariant. When `web/` exists the assertion runs exactly as before; when it doesn't, it's vacuously true instead of a false failure.

## Evidence

`web/` is present in a full clone, so I parked it (`mv web ../web-parked`) to reproduce a web-less checkout.

Before the fix, web-less checkout — reproduces the report:

```
✖ the syntax gate reaches past the repository root
  AssertionError [ERR_ASSERTION]: web/ must be inside the gate
      at tests/mjs-files.test.mjs:85:10
ℹ pass 4
ℹ fail 1
```

After the fix:

```
# web/ present (assertion still runs)
✔ the syntax gate reaches past the repository root
ℹ pass 5
ℹ fail 0

# web/ absent (parked)
✔ the syntax gate reaches past the repository root
ℹ pass 5
ℹ fail 0
```

The pre-fix failure is proof the guarded assertion still fires when `web/` is there, so nothing is weakened for the common case, it just stops treating an optional directory as required.

Tested on Node v26.4.0.

## Summary by CodeRabbit

* **Tests**
  * Updated coverage checks to account for environments where the `web/` directory is not present.
  * Preserved validation for the `lib/` directory and existing web test discovery behavior.

### DIFF
--- a/tests/mjs-files.test.mjs
+++ b/tests/mjs-files.test.mjs
@@ -22,7 +22,7 @@
 
 import { test } from 'node:test';
 import assert from 'node:assert/strict';
-import { mkdtempSync, mkdirSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
+import { existsSync, mkdtempSync, mkdirSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { dirname, join } from 'node:path';
 import { fileURLToPath } from 'node:url';
@@ -82,8 +82,16 @@ test('the syntax gate reaches past the repository root', () => {
     `gate must cover far more than the root: ${files.length} total vs ${rootOnly.length} at root`);
   assert.ok(files.some((f) => f.startsWith('tests/')), 'tests/ must be inside the gate');
   assert.ok(files.some((f) => f.startsWith('providers/')), 'providers/ must be inside the gate');
-  assert.ok(files.some((f) => f.startsWith('web/')), 'web/ must be inside the gate');
   assert.ok(files.some((f) => f.startsWith('lib/')), 'lib/ must be inside the gate');
+
+  // web/ is the one opt-in subproject in this list (#2360): tests/, providers/
+  // and lib/ ship with every install, but a checkout that never took the web UI
+  // has no web/ on disk. Assert it's inside the gate when it exists; when it
+  // doesn't, the invariant is vacuously true — the same conditional the adjacent
+  // 'web/ test discovery contract' check already uses instead of hardcoding it.
+  if (existsSync(join(ROOT, 'web'))) {
+    assert.ok(files.some((f) => f.startsWith('web/')), 'web/ must be inside the gate when present');
+  }
 });
 
 test('both syntax checkers derive their file list from the shared collector', () => {


Scott-Emberson (commented):
Reviewed at head 3e647adf71. The tolerance is narrowly scoped, which is the right way to do this. The web/ expectation is now guarded by an existsSync check on the web/ path, so it skips only when the web/ directory is genuinely absent on disk; when web/ is present the "some file starts with web/" assertion still fires. The tests/, providers/, lib/ checks and the coverage-ratio assertion stay unconditional, and collectMjsFiles still recurses over every present .mjs, so a real syntax error in any shipped file still reddens the section-1 gate. There is no blanket try/catch swallowing failures.

Owned test verified sound, the missing-directory tolerance does not mask real syntax errors. Holding formal approval until the branch is finalized, nothing blocking from the tests/ side.

Scott-Emberson (approved):

PR #3607, opened 2026-09-01
fix: widen ownerless lock grace for Windows

chipoto69:
Bridge for PR #3602 Windows CI failure.\n\nSummary:\n- keeps the follow-ups/pipeline ownerless lock recovery grace above Windows directory mtime granularity\n- prevents a just-created ownerless lock from being reclaimed when staleMs is aggressively small on windows-latest\n- leaves genuinely old ownerless locks recoverable\n\nEvidence:\n- PR #3602 Windows Tests failed in followup-seed-tests.mjs test 16: ownerless lock inside grace period was stolen/removed; run 33470224628 job 99738298210\n- local: node followup-seed-tests.mjs -> 49 passed, 0 failed\n- local: node --test tests/pipeline-lock.test.mjs tests/pipeline-lock-mkdir-eperm.test.mjs tests/lock-rm-contention.test.mjs tests/portal-health-lock.test.mjs tests/with-followups-lock.test.mjs -> 25 passed, 0 failed\n- local: npm run lint -> 614 .mjs files passed syntax check\n- local: git diff --check -> pass\n\nNo primary Rudy worktree touched. No force-push. Contributor PR #3602 remains canonical unless maintainer prefers this bridge.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved status reporting to correctly process formatted, multi-line results and diagnostics containing braces.
  - Reduced the risk of newly created pipeline locks being incorrectly treated as stale on Windows.

- **Tests**
  - Added coverage for formatted success responses and brace-containing diagnostic output.


artemtrofymenko (commented):
The stated mechanism does not reproduce on real Windows, and I think that matters for what this fix should say rather than for whether it lands.

The comment justifies `OWNERLESS_GRACE_MS 1_000 → 5_000` with *"Windows directory mtimes can round/coalesce"*. Measured on Windows 11, NTFS, non-elevated:

```
40 writes spaced 25ms apart
  distinct directory mtimes:  40 / 40
  steps between them:         23.8–28.6ms, tracking the spacing exactly
  samples where the dir mtime lagged its write:  0 / 40

200 writes with no delay, 56ms elapsed
  distinct directory mtimes:  53
  step between them:          0.33–1.67ms
  total mtime span:           55.7ms
```

Directory mtime here resolves at roughly a millisecond and does not coalesce beyond that — three orders of magnitude below the 1000ms grace that was already in place. On this hardware, no rounding or coalescing of directory mtimes can explain an ownerless lock being reclaimed inside a one-second window.

**What I cannot test, and what I think this turns into.** `windows-latest` is an Azure VM with its own storage layer, and I have no way to measure directory mtime behaviour there. So the claim may well be true *on the runner* — but then the comment wants to say "on `windows-latest`" rather than "Windows", and the evidence for it wants to come from the runner rather than from the failure it is inferred from. As written, the next person to read that line will believe something about Windows that is not true of Windows.

The alternative worth ruling out before widening the grace: on a loaded CI VM the gap between "lock created" and "staleness checked" is dominated by scheduling, not by filesystem timestamps. A run that fails at `staleMs` aggressively small under load looks identical to one that fails on mtime granularity, and a five-second grace makes both go away — which is why it is worth knowing which one it was. The failing job (`33470224628`) should be able to tell you: if the lock's own recorded timestamp and the reclaim decision are more than a second apart in wall-clock, that is scheduling, not granularity.

None of this argues the grace is wrong. One second is tight for anything crossing a virtualised filesystem, and five is cheap. It is the sentence in the source I would not want to inherit — a comment that names a mechanism is the thing a future reader trusts instead of re-measuring, which is exactly what I did here and why I could tell it apart.

Measured on the configuration CI does not run: Windows 11, non-elevated shell, Developer Mode off.

PR #3609, opened 2026-09-01
fix: four defects invisible in the maintainer's own checkout

### LINKED ISSUE #3608: Four defects that are invisible in the maintainer's own checkout (cwd, inherited stderr, optional web/, live cv.md)
Four independent defects that share one property: **each is invisible in the environment where the code was written**, and only appears once someone runs it from a different directory, reads the test log, clones without an optional subproject, or has different data in their CV. Found while running the suite on two separate career-ops instances belonging to two different people.

Happy to split into four PRs if you'd rather review them separately.

---

### 1. `scan-interamt.mjs` reads config relative to the cwd but writes relative to the data root

`PORTALS_PATH = 'portals.yml'` and `SCAN_HISTORY = 'data/scan-history.tsv'` are bare relative paths, but the script writes through `scan.mjs`'s `appendToPipeline` / `appendToScanHistory`, which resolve against `getCareerOpsRoot()`. Started from anywhere but the repo root, the two halves disagree: it reads an absent (or a different) config while still appending to the real files.

Both failure modes are silent:

- no `interamt_searches` found → falls back to the six built-in keywords, **with no title filter at all**
- no scan-history found → dedup sees an empty corpus and **re-reports every posting as new**
- `mkdirSync('data')` creates a stray directory beside the cwd that nothing writes to

`localToday()` was already fixed here (thanks); this is the remaining half.

### 2. `applyRejectionStatus()` leaks the child process's stderr

`execFileSync` inherits stderr by default. `set-status.mjs`'s `failWith` writes the machine payload to stdout **and** a human `❌ {message}` line to stderr, so that line prints straight through `applyRejectionStatus()` — a function that by contract returns a structured result and never speaks for itself.

Visible in the suite today: the nonexistent-row test deliberately passes `#999`, **passes**, and emits

```
❌ No tracker row with #999
```

into the log, indistinguishable from a real failure to anyone reading or grepping it.

### 3. `tests/mjs-files.test.mjs` asserts an untracked subproject is present

```js
assert.ok(files.some((f) => f.startsWith('web/')), 'web/ must be inside the gate');
```

`web/` is untracked — `.gitignore` names it ("e.g. `web/` before it graduates to main"), `git ls-files web` is empty, and no `web/` path is in `update-system.mjs`'s `SYSTEM_PATHS`. **A plain clone fails this test.** It also contradicts its own sibling assertion in the same file that `SKIP_DIRS` keeps the count checkout-independent. `test-all.mjs` already handles the same directory correctly ("web/ not present in this checkout — skipping").

### 4. The `verify-cv-facts` fixture is validated against the live `cv.md`

The visible-metric fixture plants `"Improved onboarding for 500 users."` and asserts it is flagged as unsupported. But `verify-cv-facts.mjs` validates against the **live source files**, so the assertion inverts for any user whose CV contains that number.

Observed on a real checkout whose `cv.md` read `~500 active users/month`: `verify-cv-facts` correctly found the metric supported, correctly declined to flag it, and the test reported that **correct behaviour as a miss**. The test passes or fails depending on whose CV is checked out — a coin flip for contributors, and a false alarm in CI on any fork carrying real data.

---

Branch with all four, each its own commit, suite green (`7567 passed, 0 failed`):
https://github.com/Wizarck/career-ops-fork/tree/fix/path-anchoring-and-test-isolation

Wizarck:
Closes #3608.

Two independent fixes, one commit each, sharing a property: **each is invisible in the environment where the code was written.** They surface only when someone reads the test log, or when the checked-out `cv.md` contains different numbers.

| Commit | Fix |
|---|---|
| `ee190306` | `invite-match.mjs` — pipe the child's stderr so `applyRejectionStatus()` stops printing `❌ …` from a passing test |
| `4878d5f9` | `test-all.mjs` — use a fixture metric no real CV can support, so the assertion tests `verify-cv-facts` and not the CV that happens to be present |

Two further fixes from #3608 were part of this PR originally and have since landed on main independently, so they were dropped in the rebase rather than re-applied:

| Was | Superseded by |
|---|---|
| `scan-interamt.mjs` — anchor `portals.yml` / `scan-history.tsv` to the data root, matching where the script already **writes** | #3511 (`7d7b0f97`) — imports `PORTALS_PATH` / `SCAN_HISTORY_PATH` from `scan.mjs` |
| `tests/mjs-files.test.mjs` — guard the `web/` assertion behind an existence check, as `test-all.mjs` already does | #3605 (`ec8ee3c4`) — same `existsSync` guard |

Full reasoning and reproduction for each is in #3608; each commit message carries its own rationale.

### Verification

`node test-all.mjs` on a clean `origin/main` worktree with both applied:

```
📊 Results: 8085 passed, 0 failed, 14 warnings
```

Equal to main's own count, which is the expected shape for two fixes that only remove noise: the first stops a stray `❌ No tracker row with #999` line from appearing in a passing run, the second keeps an existing assertion from inverting on a checkout whose `cv.md` mentions 500 users (which is how it was found).

### Notes for review

- No behavioural change to any passing path.
- The `invite-match` fix loses nothing: the suppressed stderr string is already the JSON `error` field the CLI prints as `Apply FAILED: …`. It additionally folds a captured stderr into the returned error for `failUsage`'s non-JSON branch, where it had been dropped entirely.
- The `test-all` fix is test-only.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error reporting when invite matching fails by surfacing meaningful child-process error details when available.
  - Preserved structured error information while providing a clearer fallback message.

- **Tests**
  - Updated CV verification test data to avoid results depending on the contents of the checked-out CV.
  - Added documentation clarifying the test fixture requirements and preventing false-positive passes.


Scott-Emberson (commented):
Reviewed at head c70f72c3c9. The owned test change (tests/mjs-files.test.mjs) is sound and narrowly scoped: it guards the web/ gate assertion behind an existsSync check on the web/ path, which is correct since web/ is gitignored / not in SYSTEM_PATHS and a plain checkout lacks it (the old unconditional assert contradicted this file's own checkout-independent-count sibling test). The web/ assertion still fires when web/ is present, and every load-bearing assertion stays unconditional: the tests/ / providers/ / lib/ presence checks, the ratio invariant, the no-false-pass ENOENT case ("a gate that checks nothing must never read as passed"), and the readdir ban. The other three fixes in the PR are real and outside our area but clean (invite-match piping child stderr so a passing test stops emitting a stray failure mark, scan-interamt anchoring its paths to getCareerOpsRoot so reads match writes, and a test-all fixture metric hardened so a live cv.md cannot invert the assertion).

On the one red check: the Upgrade regression gate is failing at its canary step on the updater's self-commit ("Update commit failed", "Updater self-reexec failed"), which is a repo-move environment break, not this PR. It is red identically on the other post-move PRs (3614, 3613, 3610) and was green on the pre-move ones (3605, 3604, 3603), and nothing here touches update-system.mjs or upgrade-tests.mjs. The three real test legs (ubuntu/macos/windows) are green.

Owned test verified sound; the lone red is a system-layer gate break from the repository move, not this change. Holding formal approval until the branch is finalized, nothing blocking from the tests/ side.

Scott-Emberson (commented):
Re-reviewed at head 4878d5f90d. Two things resolved: the Upgrade regression gate that was the only red is now green (all 12 checks pass, the repo-move infra break is fixed), and the owned test tests/mjs-files.test.mjs change is the same sound pattern used elsewhere, it scopes the web/ assertion behind existsSync so an absent web/ makes that one opt-in invariant vacuously true while the load-bearing tests/, providers/, lib/ gate assertions and the past-root count stay unconditional, no skip, no early return. The four defect fixes I noted before still hold.

Verified sound and CI is green now. The only remaining item is that the branch is behind main, which is outside our owned-test scope. Rebase to current main and I will re-verify at the new head and clear it to approve. Good from the tests/ side.

PR #3610, opened 2026-09-01
feat(portals): audit board content, not just reachability

Wizarck:
`verify-portals.mjs` proves a board **answers**. It never proves the board is **worth having** — and an entry can rot into uselessness in two ways a reachability check reports as healthy.

### The two failure modes

**1. No provider claims the `careers_url`.** `scan.mjs` skips the entry on every run, silently, while it still reads as coverage in `portals.yml`. In one real checkout twelve entries were in this state; one of them was a company whose real board existed and was **a single character off** the slug in the file. Nothing surfaced it — the scan summary counts what it scanned, not what it declined to.

**2. It points at a real, healthy board belonging to the wrong entity** — a parent company, a regional subsidiary, an unrelated same-named tenant. The board answers, returns hundreds of jobs, and none of them are the company you thought you were tracking.

### What this adds

`audit-portals.mjs` fetches each enabled board through the same `providers/` modules `scan.mjs` uses, and reports provider + posting count + sample titles/locations per entry, worst verdict first: `no-provider`, `error`, `empty`, `small`, `ok`. `--baseline prev.json` diffs against an earlier `--json` run and flags boards that lost ≥50% of their postings — the shape an ATS migration takes, since a migrated board collapses toward zero rather than 404ing.

**Failure mode 2 is deliberately not given a verdict.** A parent company's board is full of real jobs, and no heuristic separates it from the right one. The tool prints the sample titles and locations a human needs to judge, rather than pretending to a certainty it does not have. `--json`, `--summary` and `--strict` for CI.

`verify-pipeline.mjs` gains **Check 15**, running only the *offline* half. Provider resolution is pure config matching, so it costs no network and keeps the health check as fast as it is now:

- entry names an **unknown provider** → error (it can never scan)
- entry that **no provider claims** → warning
- `portals.yml` absent → not a finding (it is user-layer and gitignored)

The live half stays a separate command because it needs one fetch per board.

Both files are registered in `SYSTEM_PATHS`, so `update-system.mjs` actually ships them.

### Docs (second commit)

The script is documented where the others are: a row in the `AGENTS.md` Main Files table, item 6 under Pipeline Integrity (coverage as a health axis distinct from reachability, and when to run the audit), and a `docs/SCRIPTS.md` quick-reference row plus section with the verdict table, flags and exit codes.

### Verification

```
📊 Results: 8111 passed, 0 failed, 14 warnings
```

on a clean `origin/main` worktree — 26 above main's 8085, which are `tests/audit-portals.test.mjs`. `node upgrade-tests.mjs --pr-gate` on this head: `GREEN`, both legs, both new system paths present after upgrade.

Check 15 found a real dead entry on first run against an unrelated user's config:

```
⚠️  portals.yml: "Vinted" is enabled but no provider claims https://careers.vinted.com
    — scan.mjs skips it on every run without naming it (run node audit-portals.mjs)
```

### Note

Independent of #3609; either can land first. Happy to split the CLI and the `verify-pipeline` check into two PRs if you would rather take them separately.

## Summary

Users can audit portal content with `audit-portals.mjs`, not only portal reachability. The audit reports provider coverage, posting counts, sample titles and locations, and verdicts: `no-provider`, `error`, `empty`, `small`, or `ok` (`audit-portals.mjs:1`).

Users can compare audit results with a baseline. The tool flags posting-count drops of at least 50% and supports JSON, summary, filtering, and strict-mode output (`audit-portals.mjs:45`).

`verify-pipeline.mjs` now validates portal entries against providers in offline Check 15. Unknown providers are errors. Enabled entries without provider claims are warnings. Missing `portals.yml` is accepted (`verify-pipeline.mjs:196`).

The update manifest includes the audit script and its tests (`update-system.mjs:27`). Documentation covers the audit workflow, pipeline checks, and the limitation that entity mismatches require human review (`AGENTS.md:84`, `docs/SCRIPTS.md:118`).

Tests cover classification, provider failures, baseline drops, disabled entries, unclaimed entries, and nameless enabled entries (`tests/audit-portals.test.mjs:1`, `tests/verify-pipeline-check15.test.mjs:1`).

## System files touched

The changed system files are: `AGENTS.md`, `update-system.mjs`, `audit-portals.mjs`, `verify-pipeline.mjs`, `docs/SCRIPTS.md`, `tests/audit-portals.test.mjs`, and `tests/verify-pipeline-check15.test.mjs`.

No changes were made under `modes/`, `providers/`, `.github/`, or to `DATA_CONTRACT.md`.


Scott-Emberson (commented):
Reviewed at head 129c6907bf. The owned test (tests/audit-portals.test.mjs) genuinely asserts board content, not just reachability, which is the point of the feature. classifyBoard on a reachable workday board that returns an empty jobs array must come back empty (count 0), so a board that answers but has no postings is flagged rather than passed; a board with real jobs above the threshold passes with a concrete verdict and count, and a thin one is flagged as small. diffAgainstBaseline pins the migration/wrong-entity case the script exists for: Booking.com dropping 145 to 2 must yield a single lost:143 while grown boards stay quiet. It imports the real module and exercises classifyBoard / sampleJobs / diffAgainstBaseline / auditCompanies / findUnclaimedEntries with exact expected outputs, wrapped so a missing export or throw fails, no vacuous pass.

On the one red check: the Upgrade regression gate is failing system-wide from the repository move, not this PR. The canary leg dies on the updater's self-commit / a js-yaml resolution error in a temp install, the same failure shows on the other post-move PRs (3614, 3613, 3609) and was green pre-move (3605), and the FAIL lines mentioning audit-portals paths are downstream of that crashed apply (files never got copied because apply died), not a defect in this PR. The real test legs (ubuntu/macos/windows), where audit-portals.test.mjs actually runs, are green.

This is a net-new feature, so more than a tests/ change, but from the tests/ side the coverage is sound and the lone red is a system-layer gate break. Holding formal approval until the branch is finalized, nothing blocking from the tests/ side.

Scott-Emberson (commented):
Re-checked at head af725547a7. Two things resolved since my last note. First, CI is now fully green, including the Upgrade regression gate that was the only thing holding this before, so the repo-move infra break is fixed and this is no longer gate-blocked on CI. Second, the owned test tests/audit-portals.test.mjs is byte-identical to what I verified (the head move is a rebase onto the fixed main, not a test edit), so it still asserts board content rather than reachability: classifyBoard on a reachable-but-empty workday board must come back empty with count 0, a healthy board above the threshold passes, and diffAgainstBaseline catches the Booking.com 143-to-0 migration drop while staying quiet on a new company, a zero-to-zero, and churn. Every case reddens through an explicit fail(), no tautologies.

Owned test verified sound and CI is green now. The only remaining block is the review gate. Ready to approve on your go-ahead.

luochen211 (changes requested):
Two blocking gaps remain on the current head:

1. `getCareerOpsRoot()` is the data root, but both `audit-portals.mjs` and the new `verify-pipeline.mjs` check load providers from `join(dataRoot, 'providers')`. With `CAREER_OPS_ROOT` / `.career-ops-data` pointing at a separate data directory, that directory has no provider modules, so the audit reports every entry as `no-provider`. Provider code needs to come from the checkout/code root; only `portals.yml` belongs under the data root.
2. `makeHttpCtx({ maxPages: AUDIT_MAX_PAGES })` does not apply the advertised one-page budget: `makeHttpCtx()` accepts no arguments and returns no `maxPages`. A direct injected-provider repro sees `ctx.maxPages === undefined`, so a live audit can fully paginate every board.

Please split code/data roots, and either extend `makeHttpCtx` or construct `{ ...makeHttpCtx(), maxPages: 1 }`. Add regressions that run with a separate data root and assert the default context's page cap; the current tests inject `httpCtx` and miss both production paths.

Scott-Emberson (commented):
Test-axis only. The substance is `audit-portals.mjs` and `verify-pipeline.mjs` plus `AGENTS.md` and `update-system.mjs` (red-line), which is the maintainer's routing call; I am reviewing the owned test `tests/audit-portals.test.mjs`.

It is sound and genuinely pins content-vs-reachability, not a tautology. `classifyBoard` on a reachable-but-empty board (`{provider:'workday', jobs:[]}`) must return empty with count 0, `many(2)` to small, `many(threshold+1)` to ok, so a regression to reachability-only (classifying by provider presence alone) reads an empty board as ok and reddens the empty and small asserts, and the count is asserted, not just the verdict. Config-defect is separated from content (`provider:null` to `no-provider`, which outranks a fetch error), and the baseline-diff cases carry hard numeric asserts (Booking 145 to 2 flagged with `lost === 143`, while growth, new-entry, 0 to 0, and 10% churn are each asserted not flagged). The fetch-count-classify path runs end to end against a real provider mock (a throwing provider degrades to error, disabled entries skipped), so it defends against the fetch path being stubbed out.

Scott-Emberson (commented):
Test-axis only. The substance is `audit-portals.mjs` and `verify-pipeline.mjs` plus `AGENTS.md` and `update-system.mjs` (red-line), which is the maintainer's routing call; I am reviewing the two owned tests.

`tests/audit-portals.test.mjs` is sound and the delta did not weaken it (the change is +9/-0, a nameless-entry fixture plus an assertion that it is skipped into no bucket). Content-vs-reachability is still pinned: a provider-present-but-empty board asserts `verdict === 'empty'` with count 0, distinct from `no-provider`, so reachability alone cannot produce `empty`, and the baseline-diff numeric asserts hold (`lost === 143` flagged, 10% churn not flagged, zero-baseline skipped, a new company not counted as a drop). The added nameless entry also makes the pre-existing `silent.length === 1` asserts fail-red if a nameless entry ever leaked, so the change strengthens it.

The new `tests/verify-pipeline-check15.test.mjs` is a real end-to-end test, not a smoke test: it spawns the real `verify-pipeline.mjs` against a fixture and dual-asserts that the malformed-entry warning is emitted AND the success line is withheld (`warned && !claimedResolved`), so a regression that counts the malformed entry as resolved reddens, with a control run pinning the clean case both directions.

PR #3622, opened 2026-09-01
fix: ignore preserved excludes when staging updater paths

chipoto69:
## Summary
- Fixes the upgrade regression gate failure where preserved `:(exclude)...` pathspecs were passed into `git --literal-pathspecs add` and treated as literal filenames.
- Keeps preserved-file exclusions for checkout/commit scoping, but filters them out of the staging list.
- Unblocks PRs whose upgrade canary hits local preserved system-file paths (for example PR #3614/#3613 class failures).

## Test Plan
- `node --check update-system.mjs`
- `node --check upgrade-tests.mjs`
- `node tests/updater-add-paths.test.mjs`
- `node upgrade-tests.mjs --canary`
- `node upgrade-tests.mjs --pr-gate`
- `npm run lint`
- `npm install` to materialize Playwright dependency in isolated steward workspace
- `node test-all.mjs --quick` => 7570 passed / 0 failed / 13 warnings

Board context: github-steward autopilot tick 2026-09-01 15:42 CEST. Primary dirty career-ops worktree was not touched.

### DIFF
--- a/update-system.mjs
+++ b/update-system.mjs
@@ -2287,7 +2287,8 @@ async function apply() {
       // unstaged edit to a tracked file under a system directory, or an ignored
       // file that an earlier buggy run stranded in the index. Closing that is a
       // separate change to the commit call; this one only closes the add.
-      addPaths(expandToShippedFiles(pathsToStage));
+      const stageablePaths = pathsToStage.filter((spec) => !spec.startsWith(EXCLUDE_PATHSPEC_PREFIX));
+      addPaths(expandToShippedFiles(stageablePaths));
       // Scope the commit to only the staged update paths (#915 bug 2).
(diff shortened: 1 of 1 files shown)


santifer (changes requested):
Thanks @chipoto69 for **keeping the preserved-path excludes out of the `git add` call when the updater stages paths**. To get this landed we need a few concrete things; this is the complete list, not a first round:

1. Rebase on current `main`: `update-system.mjs` has moved since and conflicts.
2. Add a test that stages an update with a preserved exclude and checks it never reaches `git add`: failing without this change, passing with it. The updater runs on every install, so we want this pinned.
3. One line in the PR body on what a user saw before this fix, so the symptom sits next to the change.

PR #3627, opened 2026-09-01
fix(web): reject private URLs in Apply session

mbrow190:
## What does this PR do?

`POST /api/apply/session` Playwright-opened whatever `http(s)` URL the client sent. An attacker (or a poisoned job URL) could make the local Next server browse loopback, RFC1918, link-local, or `file:`.

This reuses the existing egress guard from `liveness-browser.mjs` (`rejectPrivateOrInvalid` + `validateUrlSecurity`) rather than forking a second allowlist:

- **Before Playwright** — literal-host 400 from the route, plus both layers inside `openSession` / `newDrivePage` so a private target never launches Chrome.
- **On every request** — context-wide `**/*` route interceptor (main document, redirect hops, subresources), same wiring as `archive-posting.mjs`.
- **On the landed URL** — re-check after `goto`, and do not retry a blocked hop.

Blocked responses are a generic 400 (`This application URL is not allowed`). They do not include the host, IP, or core reason string.

`fill` / `prefill` / `drive` do not open a new client URL; drive clicks on an already-open session are covered by the context interceptor.

## Related issue

Bug fix — no issue opened first.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] Focused tests pass: `node --test web/tests/lib/apply-url-guard.test.mjs` (13/13). No Playwright, no network (DNS resolver injected).


luochen211 (changes requested):
The route/interceptor wiring and the 13 focused tests pass locally, but the shared address classifier leaves two IPv6 private ranges open, so this Apply guard can still browse internal IPv6 services. `PRIVATE_HOST_PATTERNS` matches only `fc..:` for ULA and only the single `fe80:` prefix for link-local. It misses the other half of RFC 4193 ULA (`fd00::/8`) and most of RFC 4291 link-local (`fe80::/10`, i.e. fe80–febf). I reproduced both paths on this head: `rejectPrivateOrInvalid("http://[fd00::1]/")` and `rejectPrivateOrInvalid("http://[fe90::1]/")` return null; with the injected resolver returning either address, `assertSafeApplyUrl("https://attacker.example/apply")` resolves successfully instead of throwing. Please fix the shared `liveness-browser.mjs` classifier (so literal and DNS-resolved addresses agree) and add route/interceptor regressions for at least `fd00::1` and a non-fe80 link-local address such as `fe90::1`.

luochen211 (approved):
Re-reviewed current head 308657f. The shared classifier now covers the full RFC 4193 ULA range (fc00::/7) and RFC 4291 link-local range (fe80::/10), with literal and DNS-resolved regressions for fd00::1 and fe90::1. All required checks are green. Approved from the SSRF/security review axis.

PR #3629, opened 2026-09-01
fix(web): deny Write/Bash when evaluating untrusted JDs

mbrow190:
## What does this PR do?

Job postings are data, never instructions (`AGENTS.md` Untrusted External Content). `/api/run` evaluate still granted Claude `Write`/`Bash` (pdf already did not). A JD could try to induce file writes or a shell.

Evaluate now shares pdf's read-only tool arm. The agent emits the report in a `<<report-md>>` envelope; the backend persists via `reserve-report-num.mjs` / `merge-tracker.mjs`. `fix-portal` keeps Write/Bash. `oferta` / `auto-pipeline` are pinned to the same read-only scope so a future `/api/run` kind of that name cannot inherit persisting tools.

## Related issue

Untrusted-JD tool lockdown on Claude evaluate. Leftover: **#2507** — non-Claude `/api/run` still gets no tool flags (`spec.args(prompt)` only). That is per-CLI sandboxing (open PR #2941); not boiled in here.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] Focused tests pass (`node --test` on `claude-invocation`, `run-prompts`, `report-envelope`, `report-persist` — 65 pass). `node test-all.mjs` not re-run in this worktree (web `node_modules` absent).

## Tests

- `UNTRUSTED_JD_KINDS` (`pdf`, `evaluate`, `oferta`, `auto-pipeline`) allow no Write/Bash and pass `--strict-mcp-config`
- `fix-portal` still allows Write/Bash
- evaluate prompt forbids saving; backend persist writes report + TSV + merge
- `test-all.mjs` §55.6 freeze extended to untrusted-JD kinds

## Residual risk

- Non-Claude CLIs remain unrestricted on this route (**#2507** / #2941).
- Persistence now depends on the agent emitting a well-formed envelope with Machine Summary; a malformed report fails closed instead of writing via the agent.

### DIFF
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -15329,6 +15329,8 @@ try {
     const required = {
       'claude-invocation.mjs': join(webLib, 'claude-invocation.mjs'),
       'cv-envelope.mjs': join(webLib, 'cv-envelope.mjs'),
+      'report-envelope.mjs': join(webLib, 'report-envelope.mjs'),
+      'report-persist.mjs': join(webLib, 'report-persist.mjs'),
       'run-prompts.mjs': join(webLib, 'run-prompts.mjs'),
       'api/run/route.ts': runRoutePath,
     };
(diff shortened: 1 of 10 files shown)


luochen211 (changes requested):
The Claude tool lockdown and the four focused suites look good (65/65 pass locally), but the new backend persistence boundary writes agent-controlled metadata into TSV without validating TSV control characters. The report is model output derived from an untrusted JD; `yamlScalar()` accepts literal tabs in `company`, `role`, or `via`, and `trackerTsvLine()` joins those values verbatim. On this head, a Machine Summary containing `company: Acme<TAB>Injected` passes `parseReportMeta()` and produces 11 columns instead of the documented 10. That can shift status/score/report fields or cause `merge-tracker.mjs` to skip/corrupt the row while the report has already been written. Please fail closed on `\\t`, `\\r`, and `\\n` in every model-derived TSV scalar (or normalize them through one explicit, tested policy) before reserving/writing, and add a regression asserting hostile metadata cannot change the TSV column count. The fixed report filename is path-safe; this finding is specifically about the tracker integrity boundary introduced by this PR.

luochen211 (approved):
Re-reviewed current head e25eabe. Model-derived company, role, score, and via fields now fail closed on tab/CR/LF before reservation or persistence, and the regression verifies hostile metadata cannot create a shifted TSV row. Four focused web suites pass locally (68/68). My prior blocking finding is resolved.

PR #3634, opened 2026-09-01
fix: honor CAREER_OPS_DATA_DIR in web and scan writes

mbrow190:
## Why

Setting only `CAREER_OPS_DATA_DIR` (the documented core data-root override) made web and some scan writes use a different tree than `scan` / `doctor` / `stats`.

## What

- Web `careerOpsRoot()` now uses `CAREER_OPS_ROOT || CAREER_OPS_DATA_DIR` (absolute, or resolved against cwd). No-env default stays `cwd/..` because Next runs from `web/` — it does **not** call core `getCareerOpsRoot()`, whose default is path-resolver `__dirname`.
- `scan.mjs` joins `BLACKLIST_PATH`, `SCAN_RUNS_PATH`, and `PORTAL_HEALTH_PATH` to `DATA_ROOT`, matching pipeline/scan-history. Explicit `filePath` overrides (`loadBlacklist(filePath = BLACKLIST_PATH)`, etc.) still work.

## Tests

- `web/tests/lib/career-ops-root.test.mjs` — DATA_DIR alone, ROOT wins, relative resolve, cwd/.. default
- `tests/scan-data-root-paths.test.mjs` — blacklist + scan-runs under DATA_ROOT when cwd differs; explicit blacklist path still honored
- `tests/portal-health-path.test.mjs` — same DATA_ROOT contract for portal-health (stats.mjs already reads it from DATA_ROOT)

No user-layer files.

### DIFF
--- a/scan.mjs
+++ b/scan.mjs
@@ -1985,7 +1985,7 @@ export async function appendToScanHistory(offers, date, status = 'added') {
 
 // ── Company blacklist (#1742) ───────────────────────────────────────
 
-const BLACKLIST_PATH = 'data/blacklist.md';
+const BLACKLIST_PATH = path.join(DATA_ROOT, 'data/blacklist.md');
 
 /**
  * Parse the user's do-not-apply list (data/blacklist.md, user layer, opt-in).
@@ -2034,7 +2034,7 @@ export function loadBlacklist(filePath = BLACKLIST_PATH) {
 
 // ── Scan-run persistence (#1604) ────────────────────────────────────
 
-const SCAN_RUNS_PATH = 'data/scan-runs.tsv';
+const SCAN_RUNS_PATH = path.join(DATA_ROOT, 'data/scan-runs.tsv');
 
 // One row of run counters per non-dry scan — today these numbers are printed
 // once in the summary and lost when the terminal scrolls. Full ISO timestamp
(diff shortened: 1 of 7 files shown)


Scott-Emberson (commented):
Reviewed at head 9746695181. The owned tests are sound. scan-data-root-paths.test.mjs spawns a child with CAREER_OPS_DATA_DIR set to a data root and cwd set to a different sandbox, deletes the scanner path vars, and asserts loadBlacklist() default keys come from DATA_ROOT (datarootco), an explicit filePath override is still honored, and appendScanRunSummary lands in dataRoot/data/scan-runs.tsv with a negative check that cwd stays empty. portal-health-path.test.mjs pins the row under DATA_ROOT, not under cwd, and not under the script dir. Both redden if CAREER_OPS_DATA_DIR were ignored, no false pass.

Two things for the maintainer. First, the one red leg (test windows-latest) is not this PR: it is the pre-existing churnLock Windows flake in tests/pipeline-lock.test.mjs (the -5 ceiling case giving up after 74ms), a file this PR does not touch, and the non-Windows legs plus the upgrade gate are green. Second, this is a competing duplicate of PR #3635 (chipoto69): the two shared owned test files are byte-identical between the PRs, and #3635 is the superset, it also rewrites pipeline-lock.test.mjs to fix the exact flake that is reddening this PR's Windows CI. So #3635 carries this PR's changes verbatim plus the flake fix, while this one stays red on that flake until a rebase. My recommendation is to land #3635 and close this one; the DATA_ROOT fix here is sound and will ship either way. Holding formal approval, and note the red CI is a known flake, not a blocker from our side.

PR #3635, opened 2026-09-01
test: harden pipeline lock ceiling test on Windows

chipoto69:
## Summary\n- Hardens the PR #3634 pipeline-lock maxWaitMs negative/-Infinity regression test by replacing the churning-lock fixture with a single live holder.\n- Keeps the assertion focused on sanitization: bad negative ceilings must not become an immediate zero ceiling.\n\n## Test Plan\n- node --test tests/pipeline-lock.test.mjs (twice, passed 21/21 each)\n- node test-all.mjs --quick (local macOS run reached unrelated missing Playwright/template failures; pipeline-lock suite passed)\n\nContext: steward follow-up for career-ops-hq/career-ops#3634 Windows-only failure in tests/pipeline-lock.test.mjs.

### DIFF
--- a/scan.mjs
+++ b/scan.mjs
@@ -1985,7 +1985,7 @@ export async function appendToScanHistory(offers, date, status = 'added') {
 
 // ── Company blacklist (#1742) ───────────────────────────────────────
 
-const BLACKLIST_PATH = 'data/blacklist.md';
+const BLACKLIST_PATH = path.join(DATA_ROOT, 'data/blacklist.md');
 
 /**
  * Parse the user's do-not-apply list (data/blacklist.md, user layer, opt-in).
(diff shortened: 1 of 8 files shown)


Scott-Emberson (commented):
Reviewed at head 4dd91e2c08. All three owned tests are sound. The pipeline-lock Windows hardening is a flakiness fix, not a weakening: the negative / -Infinity ceiling test swaps the churnLock holder (whose directory-fingerprint progress-extension could push the wait past the settle cap on Windows and spuriously fail) for a single live lock held through the finally, with the assertion unchanged (LockTimeoutError AND elapsed >= 300). A zero-clamp regression still times out below the timeout and reddens, and there is no platform branch or skip. portal-health-path.test.mjs pins CAREER_OPS_DATA_DIR against a foreign cwd and asserts the row lands under DATA_ROOT, not under cwd, and not under the script dir. scan-data-root-paths.test.mjs asserts loadBlacklist() default reads DATA_ROOT (datarootco), an explicit filePath override is still honored (overrideco), and scan-runs is written under DATA_ROOT with a negative check that cwd stays empty. Both path tests redden on a cwd-relative regression in both the landed and leaked directions.

One thing for the maintainer: this and PR #3634 (mbrow190) are the same DATA_ROOT fix by different authors, and the two shared owned test files (portal-health-path and scan-data-root-paths) are byte-identical between the two PRs. This PR is the superset, it additionally rewrites pipeline-lock.test.mjs to drop the flaky churnLock holder, which is the exact test currently reddening #3634's Windows CI. So landing this one carries #3634's owned changes verbatim plus the flake fix, whereas #3634 alone stays red on that Windows flake until a rebase. My recommendation is to land this and close #3634. Owned tests verified sound. Holding formal approval until the branch is finalized, nothing blocking from the tests/ side.

Scott-Emberson (approved):

PR #3645, opened 2026-09-01
feat(providers): add Built In aggregator provider (market hosts + card enrichment)

pwilinchery:
## What does this PR do?

Adds a `builtin` scanner provider for [Built In](https://builtin.com), a zero-auth US tech aggregator. It parses two payloads out of the same fetched page — the schema.org `ItemList` blob as the spine, and the rendered job card as optional enrichment — joined 1:1 on the numeric job id, so `company`, `location`, `salary` and `postedAt` are populated and the existing zero-token gates have something real to judge.

Built In is a board-wide aggregator across the many employers that post to one board, in the same shape as the existing arbeitnow, flowxtra, jobicy and 4dayweek providers — employers post to Built In directly and it does not index other job boards, so this reads one source and leaves cross-source aggregation in core, per Source Indexing Policy rule 5. It spans many employers rather than one tracked company, so there is no `detect()`; it is reached only through an explicit `provider: builtin` entry.

Details:

- **Market hosts.** `host:` picks one of the nine Built In market hosts against an allowlist, defaulting to `builtin.com`. All nine serve byte-identical markup. The market host is the only location scoping this board offers: on four identical queries, one page each, the national host returned 89 rows spread across many metros while a single market host returned 64 rows that were nearly all in-metro or US-remote. Only 8 job ids overlapped, and a market host carries that market's remote inventory too, so it is not "onsite only". `scope: remote` restricts to `/jobs/remote`.
- **The spine is the only required half.** If card parsing yields nothing the provider still emits title, url and description rather than dropping jobs.
- **Nothing is guessed.** Card fields are anchored on the FontAwesome icon classes rather than field order. Every enriched field is optional and every unresolvable one is left empty: per the Job contract an empty location PASSES `location_filter`, so a wrong guess silently rejects real jobs while an empty one simply leaves the field unset. A multi-location card whose tooltip cannot be read therefore yields `''`, not a placeholder like `"3 Locations"` that names no city. Hourly bands are skipped (Built In renders some as `"115K-130K Hourly"`, their own data error) and salary currency is left unset because the card never states one.
- **Drift guard.** Card-parsing failure is silent by construction: every field comes back empty, an empty location passes the filter, and the entry keeps returning jobs. So `fetch()` warns loudly when a page yields rows but zero cards, or cards but almost no locations, naming the portals entry.
- **SSRF.** The host comes from config, so every request URL passes an allowlist
+ HTTPS check immediately before it is fetched.
- **No default queries.** An entry must supply `queries:` and/or `categories:`, otherwise `fetch()` returns `[]`, so no search terms ship in shared code.

On the Source Indexing Policy: listings are real, employer-attributed, and free for candidates to read and apply to (rule 1). On rule 2, the listing page exposes only its own `/job/` and `/company/` links — the employer's ATS URL appears on the detail page, not in the listing payload — so the builtin.com job URL is the shortest verifiable path this source publishes at this level.

Config:

```yaml
- name: Built In
  provider: builtin
  builtin:
    queries: ["platform engineer"]   # required: queries and/or categories
    categories: [dev-engineering]
    host: www.builtinseattle.com     # optional, default builtin.com, allowlisted
    scope: remote                    # optional -> /jobs/remote/...
    max_pages: 3
```

Flat `queries` / `categories` / `max_pages` keys are also honoured; nested wins. `templates/portals.example.yml` carries the same stanza, since a provider with no `detect()` is only reachable through config.

New suite `tests/providers/builtin.test.mjs`, 43 assertions, all synthetic fixtures and no network. `node test-all.mjs` on this branch: **7731 passed, 0 failed**, against `main` at 1696bec (7686 passed, 0 failed on its own).

## Related issue

None — new zero-auth scanner provider, which CONTRIBUTING and the template both exempt from issue-first.

## Type of change

- [ ] Bug fix
- [x] New feature

## Checklist


Scott-Emberson (commented):
Reviewed at head cec45dc6bd. The Built In provider handles its fetch surface the codebase-standard way, and the test covers it concretely. resolveHost rejects off-allowlist inputs (evil.com, builtin.com.evil.com, notbuiltin.com, a number) to null, and fetch against an off-allowlist host returns an empty list with zero requests (no silent fallback to the canonical host). assertHost re-checks HTTPS-only and the canonical allowlist before every fetch, the fetch URLs are hardcoded https with the allowlisted host, and each request pins redirect:'error' and a UA with the requested host asserted to be builtin.com, plus a scope path-traversal (../../etc) is rejected to empty.

The point I most wanted to confirm on an aggregator with card enrichment: enrichment is not extra fetches. It parses the same already-fetched listing HTML (the rendered cards joined to the ItemList spine on the numeric job id), so there is no per-card URL fetch and no attacker-influenced secondary target. The only fetch target is the allowlisted host with a slug-validated scope and encodeURIComponent'd query. Parsing itself is pinned with exact values: company-join-by-id, mode and place compose, multi-location tooltip, unresolvable to empty, annual salary range, hourly dropped, numeric postedAt, and a drift guard that reddens on rows-but-no-cards.

This is a net-new provider, so more than a tests/ change, but from the tests/ side the coverage is sound and there is no SSRF surface introduced. Holding formal approval until the branch is finalized, nothing blocking from the tests/ side.

Scott-Emberson (approved):

PR #3654, opened 2026-09-02
feat(browser-extract): read Greenhouse, Lever and Ashby JDs from their public APIs

### LINKED ISSUE #2582: feat: fetch-jd.mjs — fetch a JD via its ATS provider API instead of the full HTML page

pwilinchery:
## What does this PR do?

Extends `jd` mode's API-first read from Workday to the other three ATSs whose public endpoint ships the JD body — Greenhouse, Lever and Ashby — behind one dispatch point, and adds `fetch-jd.mjs`, the bash-callable front door #2582 asks for: JD text on stdout and exit 0 on a hit, exit 1 with empty stdout otherwise so the caller's existing `WebFetch` fallback is the next step.

Today a headless caller fetches the posting page and strips HTML. On Ashby and on Greenhouse's embedded boards that returns an unrendered SPA shell, correctly detected as thin, so the caller falls through to `WebFetch` — which hits the same shell and fails the same way. Meanwhile the ATS's own public endpoint, the one `resolveAtsApi()` already points at for liveness, returns the full description for free.

## What to look at

`update-system.mjs` is in this diff for **one line**: a `'fetch-jd.mjs'` entry in `SYSTEM_PATHS`, added directly after the existing `'browser-extract.mjs'`. There is no other change to that file, and no change to `USER_PATHS` or to any updater logic. The entry is required or `validate-system-paths-coverage.mjs` fails on the new script as an orphan, and installs would never receive it on `apply`.

That registration is verified end to end rather than by inspection — `node upgrade-tests.mjs --pr-gate` upgrades a real `career-ops-v1.31.0` install onto this commit and asserts the new script lands:

```
PASS [career-ops-v1.31.0 dismiss=tracked] new system path present after upgrade (#1998 class): fetch-jd.mjs
```

`AGENTS.md` is one added table row describing `fetch-jd.mjs`, in the script table next to `check-liveness.mjs`. No prompt or behavioral text is touched.

## Related issue

Closes #2582.

Two notes on how this maps to that issue. **Workday is already done:** #3257 landed the Workday half and scoped itself deliberately, noting that #2582 proposes the general version and that "if #2582 lands, the Workday half here folds into it." This is that generalization, so the remaining three providers plus the script complete the issue's scope.

**One deviation from the issue text:** #2582 suggests reusing `providers/_http.mjs`'s fetch wrapper. These fetchers instead follow `fetchWorkdayJd` next door — plain `fetch` with `DEFAULT_USER_AGENT`, `redirect: 'error'` and an `AbortController` timeout — so all four ATS paths in this file stay identical. Happy to move all of them onto `_http.mjs` in a follow-up if you'd rather converge the other way.

**On the claim.** #2582 was proposed by @darkpandawarrior, and the offer on that thread was theirs to take first. It has been unassigned since, and CONTRIBUTING says "a PR straight onto any unassigned issue is always welcome" — so this is sent in that spirit, not to cut ahead of them. @darkpandawarrior, if you were still planning to build it, say so and I will happily close this in favour of yours; the design notes above are yours to use either way.

The `batch-runner.sh` consumer side (an `--extraction api|browser` filter) is out of scope here; it adds a new CLI flag, so it wants its own issue first.

## Type of change

- [ ] Bug fix
- [x] New feature

## Checklist

- [x] If this is a new feature or architecture change, I opened an issue first (bug fixes, providers, docs & translations are exempt — send those straight in) — see the note under **Related issue**: the issue is #2582, opened by someone else.

---

## Implementation notes

**What is added**

- `fetchAshbyJd` / `fetchGreenhouseJd` / `fetchLeverJd`, each following `fetchWorkdayJd` exactly: `rejectPrivateOrInvalid` before the request, `AbortController` timeout, `DEFAULT_USER_AGENT`, `redirect: 'error'`, and `null` on every inconclusive outcome (non-200, redirect, unparseable body, job not on the board, timeout) so the caller falls through to the browser path unchanged.
- `fetchJdViaKnownApi()` — one dispatch point over `JD_TEXT_API_ATS`, so the interactive extractor and any headless caller cannot drift on which ATS is API-fetchable.
- `JD_TEXT_API_ATS` in `liveness-api.mjs`: the four ATS ids whose API returns a JD body. Microsoft and LinkedIn are on `ATS_PROVIDERS` for liveness only — their public endpoints answer search/status, never body text — so they are excluded and stay browser-required.
- `fetch-jd.mjs`, registered in `SYSTEM_PATHS` so the updater propagates it and the coverage guard stays green.

No new HTTP client and no new dependency.

**Per-provider details that aren't obvious from the endpoint shape**

| Provider | Detail |
|---|---|
| **Ashby** | The public API is **org-level** (the whole board), not per-job, so the posting is selected by a case-insensitive `id` compare — mirroring `classifyAshbyBoard`'s match logic without importing it, since that returns a liveness verdict, not a job record. |
| **Lever** | `lists` carries the labeled sections (Requirements, Nice to have) as separate HTML blocks; dropping them loses half the JD. Lever also still **serves** `state: 'closed'` postings, so those are rejected rather than evaluated as live. |
| **Greenhouse** | `content=true` is required or the payload omits the description entirely (the same query param `providers/greenhouse.mjs` uses). `content` is HTML, often entity-escaped, so it goes through `jdHtmlToText` like Workday's `jobDescription`. |

**Testing**

Each fetcher delegates its field mapping to a pure exported normalizer — the shape #3257 established with `normalizeWorkdayJob` — so the mapping is unit tested without a network round trip. The live fetch and emit paths are left uncovered rather than stubbed, same as #3257.

13 new assertions in `tests/browser-extract.test.mjs`. Ten cover the three normalizers: metadata header, entity-escaped `content`, Lever's labeled `lists` (including a block with no heading), the `isListed: false` and `state: 'closed'` signals, `null` on a bodyless payload, and the text cap on every path. The other three are the ones worth a second look, because nothing else in the suite reddens if they go:

- **Routing drift.** `fetchJdViaKnownApi` switches on `resolveAtsApi(url).ats` with `JD_TEXT_API_ATS` as the gate in front of it. Rename an id on either side and every posting for that ATS falls through to the browser silently — no error, no empty JD, just the slow path and a token bill. The test asserts each gated id is the id `resolveAtsApi` actually emits for a representative URL of that ATS.
- **Ashby's `parts.jobId`.** Ashby's API is org-level, so `fetchAshbyJd` picks the posting out of the whole board by that field. If it stops being populated the lookup matches nothing and returns `null`, which reads as "no JD here" rather than as a bug.
- **The dispatch gate's no-network `null`** for a host outside `JD_TEXT_API_ATS`, which is what lets a caller fall through rather than hang on an unknown host.

Each was checked against its own sentinel: renaming an id in `JD_TEXT_API_ATS`, dropping the Lever `lists` append, and making the Ashby id compare case-sensitive each redden the corresponding assertion and nothing else.

```
npm run lint                               ->  625 .mjs files pass (Node 18 syntax gate)
node test-all.mjs --only browser-extract   ->  56 passed, 0 failed   (was 43)
node test-all.mjs                          ->  7700 passed, 0 failed, 13 warnings
node upgrade-tests.mjs --pr-gate           ->  GREEN
node upgrade-tests.mjs --canary            ->  GREEN
node upgrade-tests.mjs --local-paths       ->  GREEN
node validate-system-paths-coverage.mjs    ->  OK: 1312 tracked files covered
```


Scott-Emberson (commented):
Reviewed at head 3e7025e508. The ATS-API extraction handles its fetch surface safely and the test pins both the routing and the parsing. Each provider's match() anchors the input hostname (anchored regexes for greenhouse.io, the exact jobs.ashbyhq.com host, the myworkdayjobs pattern) and returns null otherwise, so a spoofed suffix like greenhouse.io.evil.com does not resolve, and api() builds the fetch URL on fixed hosts (boards-api.greenhouse.io, api.ashbyhq.com) or reconstructs the workday host from validated captures, never a free-form echo of the input. There is an allowlist gate (greenhouse/lever/ashby/workday only), and every fetcher calls rejectPrivateOrInvalid (http(s)-only, blocks private and loopback) then fetches with redirect:'error'. The owned test asserts that example.com, linkedin.com (present in the provider list but excluded from the JD-text allowlist), a non-URL, and an empty string all return null with no network call, pins the four allowlisted route ids, and asserts concrete parsed fields (location, req id, decoded entities, bullet formatting, closed-state to null, listing and apply flags, the text cap) that redden on a parsing regression.

One honest coverage note, not a blocker: the four network fetchers hold the SSRF guards (redirect:'error', rejectPrivateOrInvalid) but those guards are exercised by liveness-api's own suite, not by assertions in this owned file, so a regression deleting a guard from one fetcher would not redden this test directly. The routing rejections and the parsing coverage here are solid; the guard coverage lives one module over.

This is a net-new feature, so more than a tests/ change, but from the tests/ side the coverage is sound and there is no SSRF surface introduced. Holding formal approval until the branch is finalized, nothing blocking from the tests/ side.

luochen211 (approved):
Reviewed the network/data path at `3e7025e508ad`.

The caller cannot select an arbitrary fetch host: `resolveAtsApi()` first matches an anchored ATS hostname/path shape, validates every captured value with the restricted segment rules, and reconstructs the request on the provider's fixed API host (or the validated Workday subdomain under `myworkdayjobs.com`). `fetchJdViaKnownApi()` also gates the resolved provider through `JD_TEXT_API_ATS`; LinkedIn and unknown hosts fall through without a fetch. Each of the four fetchers rechecks the constructed URL, refuses redirects, applies an abort signal, treats non-OK/unparseable responses as inconclusive, and only emits capped normalized text.

I ran `node tests/browser-extract.test.mjs` locally (45 checks passed), traced all four fetch call sites, ran `git diff --check origin/main...HEAD`, and verified the current GitHub matrix/CodeQL/dependency checks are green. No blocking issue found.

browser-extract.mjs
```diff
@@ -257,6 +257,266 @@ async function fetchWorkdayJd(apiUrl, postingUrl, textCap, timeoutMs) {
...
+    ? json.lists
+        .map((l) => {
+          const heading = str(l?.text);
+          const content = jdHtmlToText(l?.content);
+          return content ? `${heading ? `${heading}\n` : ''}${content}` : '';
+        })
+        .filter(Boolean)
+        .join('\n\n')
+    : '';
+  const body = [str(json.descriptionPlain), lists].filter(Boolean).join('\n\n');
```
nikitacometa: `additionalPlain` (2,107 chars on a live Palantir posting: benefits, accommodation, comp) is dropped here; the rendered page shows it after the lists.

browser-extract.mjs
```diff
@@ -257,6 +257,266 @@ async function fetchWorkdayJd(apiUrl, postingUrl, textCap, timeoutMs) {
...
+ * rendered page shows them, or the JD loses its requirements entirely. And a
+ * `state: 'closed'` posting is still served by the API, so it is rejected here
+ * rather than evaluated as live.
+ *
+ * @param {any} json - parsed postings-api response body
+ * @param {string} postingUrl
+ * @param {number} [textCap]
+ */
+export function normalizeLeverJob(json, postingUrl, textCap = JD_TEXT_CAP) {
+  if (!json || typeof json !== 'object' || json.state === 'closed') return null;
```
nikitacometa: Public v0 has no `state` field (measured 2026-09-02); closed postings 404. Inert branch.

browser-extract.mjs
```diff
@@ -257,6 +257,266 @@ async function fetchWorkdayJd(apiUrl, postingUrl, textCap, timeoutMs) {
...
+ * @param {string} postingUrl
+ * @param {number} [textCap]
+ */
+export function normalizeGreenhouseJob(json, postingUrl, textCap = JD_TEXT_CAP) {
+  const body = jdHtmlToText(json?.content);
+  if (!body) return null;
+
+  const str = (v) => (typeof v === 'string' && v.trim() ? v.trim() : '');
+  const meta = [];
+  const offices = Array.isArray(json?.offices) ? json.offices.map((o) => str(o?.name)).filter(Boolean) : [];
```
nikitacometa, line 386: `location.name` is the primary field per providers/greenhouse.mjs:55; `offices` is often empty.

browser-extract.mjs
```diff
@@ -257,6 +257,266 @@ async function fetchWorkdayJd(apiUrl, postingUrl, textCap, timeoutMs) {
...
+  const resolved = resolveAtsApi(url);
+  if (!resolved || !JD_TEXT_API_ATS.has(resolved.ats)) return null;
+
+  let result = null;
+  switch (resolved.ats) {
+    case 'workday':
+      result = await fetchWorkdayJd(resolved.apiUrl, url, textCap, timeoutMs);
+      break;
+    case 'ashby':
+      result = await fetchAshbyJd(resolved.apiUrl, resolved.parts.jobId, url, textCap, timeoutMs);
```
nikitacometa: `resolved.timeoutMs` (Ashby: 20 s) is ignored; this call gets the Workday default.

browser-extract.mjs
```diff
@@ -257,6 +257,266 @@ async function fetchWorkdayJd(apiUrl, postingUrl, textCap, timeoutMs) {
...
+    return null;
+  } finally {
+    clearTimeout(timer);
+  }
+}
+
+/**
+ * Try every known JD-text-capable ATS API for one posting URL before anyone
+ * reaches for a browser. This is the single dispatch point both this script's
+ * own CLI (jd mode) and `jd-api-fetch.mjs` (the batch-runner.sh prefetch hook)
```
nikitacometa: `jd-api-fetch.mjs` does not exist; the script is `fetch-jd.mjs`, and batch-runner is out of scope.

nikitacometa (commented):
Routing, SSRF guards and the normalizers check out at 3e7025e5, and `validate-system-paths-coverage` is green with the new entry. Measured one thing live and read the rest against the providers that already handle these payloads:

1. Lever drops `additionalPlain`. Live `api.lever.co/v0/postings/palantir` (2026-09-02): `descriptionPlain` is 1,228 chars (it already equals `openingPlain` + `descriptionBodyPlain`), `lists` is three blocks, and `additionalPlain` is 2,107 chars of "Life at …", accommodation and comp text that the rendered page shows after the lists. The argument the PR makes for `lists` applies to it.
2. The same payload has no `state` key (keys: additionalPlain, additional, categories, createdAt, descriptionPlain, description, id, lists, text, country, workplaceType, opening, openingPlain, descriptionBody, descriptionBodyPlain, hostedUrl, applyUrl). The public v0 endpoint 404s closed postings, per liveness-api.mjs:91. The `state === 'closed'` branch is inert and "Lever still serves closed postings" is not what the public API does. Harmless, but the test pins a shape the API never emits.
3. Greenhouse reads `offices[]` only. providers/greenhouse.mjs:55 documents `location.name` as the primary field and offices as enrichment; a `{location:{name:"Remote"}, offices:[]}` job gets no `Location:` line.
4. `resolveAtsApi` gives Ashby `timeoutMs: 20_000` (liveness-api.mjs:113); `fetchJdViaKnownApi` ignores `resolved.timeoutMs` and uses the Workday default, so a slow Ashby board that liveness waits for falls through here. providers/ashby.mjs also reads `workplaceType` / `isRemote`, which `normalizeAshbyJob` does not.
5. browser-extract.mjs:485 names `jd-api-fetch.mjs` and a batch-runner prefetch hook; the script is `fetch-jd.mjs` and batch-runner is out of scope per the description.
6. The routing-drift test compares ids to `resolveAtsApi`; deleting `case 'lever'` from the dispatcher leaves it green. A map of ats → fetcher, asserted equal to `JD_TEXT_API_ATS`, would go red on that.

Scott-Emberson (commented):
Re-checked at head d4a21bddbe. The head move is a large repo-move main merge; on a blob compare my subject test tests/browser-extract.test.mjs is byte-identical to what I reviewed, so the earlier verification stands: the ATS-API extraction routes only to the anchored known hosts, rejects spoofed and non-ATS URLs with no network call, uses rejectPrivateOrInvalid plus redirect:'error' on every fetcher, and pins the parsed JD fields concretely. Nothing changed in our subject file.

The substance is in browser-extract.mjs, outside our owned paths, so this stays a test-axis review: the owned test is sound and unchanged, the merge is the maintainer's call. Nothing blocking from the tests/ side.

browser-extract.mjs
pwilinchery: Add additionalPlain behind lists keeping the same order as the webpage.

browser-extract.mjs
pwilinchery: removed

luochen211 (approved):
Re-reviewed current head 08ae445. The follow-up covers the identified ATS extraction gaps: Lever additionalPlain ordering, Greenhouse location precedence, Ashby work-model and timeout handling, and dispatcher-key coverage. All required checks are green. Approved from the extraction/test/security review axis.

Scott-Emberson (commented):
Re-reviewed at head 08ae4457ae (the review-feedback commit on the ATS JD-API path). The delta is a substantive normalizer rework (Greenhouse location.name primary over offices, Ashby workplaceType over isRemote, Lever additionalPlain with an order guarantee, and the dispatch switch replaced by an exported JD_FETCHERS map), and every test change tracks a matching implementation change. It actually strengthens the safety-relevant coverage: the new JD_FETCHERS set-equality test asserts the fetcher map exactly matches the gated allowlist and that each value is a function, so a dropped fetcher now reddens, which the prior routing-only check could not catch. The earlier network-safety assertions are untouched: the anchored no-network gate for a non-covered URL, and the per-fetcher rejectPrivateOrInvalid plus redirect:'error'. The one removed assertion (Lever state:closed) tracks a deliberate implementation removal, not masked coverage, liveness for Lever moved to the fetcher's 404 guard while the analogous Workday and Ashby signals remain tested.

The substance is browser-extract.mjs, outside our owned paths, so this stays a test-axis review: the owned test is sound and the rework does not weaken the safety assertions, but the merge and scope are the maintainer's call. Nothing blocking from the tests/ side.

nikitacometa (approved):
Re-checked at 08ae4457. All six points from my pass are in: `additionalPlain` lands after the lists and the test pins the order by index rather than by `includes`; the inert `state: 'closed'` branch is gone and the docblock says why (404 on closed, no `state` key on live); `location.name` is primary, with `offices[]` joined only for a work-model-only name through the provider's own `isWorkModelOnly`; `resolved.timeoutMs` wins over the caller's default and the test captures the abort timer (Ashby 20 s, the other three 15 s) with fetch stubbed and both globals restored in `finally`; `workplaceType` over `isRemote` with the fallback kept; `JD_FETCHERS` exported as a map and asserted set-equal to `JD_TEXT_API_ATS`, so a deleted fetcher reddens. The docblock names `fetch-jd.mjs` now. Nothing left from my side.

Issue #1411, opened 2026-07-02
Feature proposal: a fail-closed CV faithfulness gate (enforce "keywords reformulated, never fabricated")


kellymmcgrath:
## Summary

Add an optional, fail-closed **CV faithfulness gate** that enforces the project's founding rule — *"Keywords get reformulated, never fabricated"* (quoted verbatim in `CLAUDE.md`/`AGENTS.md`) — on generated CVs. Today that rule is prose guidance the agent is trusted to follow; nothing checks it, so a tailored CV can silently drift from `cv.md` (reworded bullets, invented metrics, fabricated scope). This proposal makes the rule executable.

## Motivation

The whole system is built to reformulate `cv.md` content for each JD **without fabricating**. That's a correctness property, and correctness properties decay when they live only as instructions. A deterministic gate that runs after render — and stops a non-compliant CV — turns the founding rule into an enforced invariant.

## Proposal

A `verify-cv.mjs` gate. Every experience bullet/header in the rendered CV must be one of:

1. **Verbatim** from `cv.md`;
2. a **declared surgical term-swap** — `swaps: [[cvTerm, jdTerm], …]`. The gate *reverse-applies* the swap and requires the result to still match `cv.md`, so a swap can change a keyword but **cannot hide a rewrite or fabricated tail**;
3. a **declared, justified, optionally-expiring net-new addition** — `allow: [[prefix, reason, expiry?], …]`.

Plus configurable checks: no pronouns (résumés are third-person), no em-dashes / banned AI-tell phrases in *authored* text only (never the candidate's own verbatim words), one page, Areas-of-Expertise ≤ 2 lines. Everything is config-driven — no hardcoded identity or paths.

## Design (follows established practice)

- **Policy-as-code, fail-closed** — Open Policy Agent model (deny-by-default, versioned, auditable).
- **Justified, expiring waivers** — Snyk `.snyk` (`reason` + `expiry`); ESLint `require-description` + `reportUnusedDisableDirectives` (flag stale waivers).
- **Golden-master / approval testing** — `cv.md` is the golden source; deviations are approved deltas.
- **Provenance** — W3C PROV framing (each derived line traces to a source line + transform).
- **Two layers** — deterministic string/diff gate + an *optional* LLM faithfulness judge (RAGAS-style groundedness) for the semantic checks a string gate can't do (advisory).

[...]

PR #3657, opened 2026-09-02
fix(cv): block delegated authorship escalation

luochen211:
## Summary

- Block the concrete #1411 case where a source attributes implementation to a vendor but generated text claims the candidate wrote or built it.
- Keep the detector narrow: require an explicit delegation verb, an explicit third-party role, a direct implementation verb in the rewrite, and at least two shared content tokens.
- Fail open when the source statement mixes delegated and direct work, and prefer separate explicit direct-work evidence over a nearby delegated project.
- Add regressions for the reported WebGL kiosk example, attribution-preserving rewrites, unrelated direct work, and ambiguous/mixed source evidence.

Part of #1411.

## Validation

- `node tests/nonmetric-fact-gate.test.mjs`
- `node tests/cover-fact-gate.test.mjs`
- `node verify-cv-facts.mjs --self-test` — 70 passed, 0 failed
- `env -u GEMINI_API_KEY node test-all.mjs --quick` — 7,693 passed, 0 failed, 13 existing environment/content warnings


Scott-Emberson (commented):
Reviewed at head 9eeb8269fb. On an authorship surface this is the right kind of gate, and the owned test pins it in both directions. The negative case: a source attributes a WebGL build to a named vendor (Acme Interactive), and a rewrite that escalates it to "wrote the WebGL implementation" is asserted verdict block, with the claim showing up in both delegatedAuthorshipClaims and unsupportedFacts, so it reddens if detection regresses. The positive case: a rewrite that keeps the vendor attribution passes with no authorship flag, so it does not over-block ordinary attributed language. The detector requires a delegation verb plus a named third-party role plus at least two shared tokens, skips statements that retain attribution, and folds the finding into unsupportedFacts to drive the block.

One thing worth naming, since this is anti-fabrication: the gate is deliberately narrow. It requires an explicitly named vendor or agency plus a delegation verb and fails open on mixed or ambiguous source, so escalations phrased as vague leadership prose or unnamed delegation ("led the effort that built X" with no named party) are not caught. That is the documented design scope rather than a defect, the test asserts exactly the implemented narrow behavior, but it is worth knowing the gate closes one specific escalation pattern, not all of them.

Owned test verified sound, blocks the named-delegation escalation and does not over-block legitimate attribution. Holding formal approval until the branch is finalized, nothing blocking from the tests/ side.

chipoto69 (changes requested):
Reviewed PR #3657 at head 9eeb8269fb.

Verdict: changes requested.

What passed:
- GitHub checks are green: Analyze(go), Analyze(javascript-typescript), CodeQL, Upgrade regression gate, cv-visual, dependency-review, guard, label, macOS/Ubuntu/Windows tests, welcome.
- Local steward verification passed:
  - /Users/rudlord/.bun/bin/bun run lint
  - node tests/nonmetric-fact-gate.test.mjs
  - node verify-cv-facts.mjs --self-test

Blocking issue:
The new authorship detector fails open on a common delegated-authorship wording where the direct-work verb belongs grammatically to the third party:

- source: "Managed vendor Acme Interactive, which built the WebGL implementation for an in-store kiosk."
- target: "Wrote the WebGL implementation for an in-store kiosk."
- observed result: delegatedAuthorshipClaims returns [] and verifyFacts verdict is pass.

Same with: "Oversaw contractors who developed the onboarding automation in Node.js." -> "Developed the onboarding automation in Node.js."

Root cause: delegatedSources filters out every source statement containing DIRECT_AUTHORSHIP_SIGNAL_RE, even when that verb is attached to vendor/contractor/agency, not the candidate. For this anti-fabrication gate, that is too permissive: contractor/vendor-built is exactly delegated work and should not allow first-person direct authorship in generated CV prose.

Suggested fix:
Do not globally drop delegated source statements just because they contain built/developed/etc. Distinguish direct candidate evidence from third-party execution evidence, or at minimum add tests for:

1. "Managed vendor X, which built Y" blocks "Wrote/Built Y".
2. "Oversaw contractors who developed Y" blocks "Developed Y".
3. Separate explicit candidate direct evidence still wins when present.

No code pushed by steward. Primary worktree untouched; review performed in dedicated workspace:
/Users/rudlord/ORGANIZED/_agent-workspaces/github-steward/career-ops-pr3657-review-20260902/repo

Scott-Emberson (commented):
Re-reviewed at head 729ddf53bc (moved since my last comment, largely a repo-move main merge plus a real "catch third-party implementation authorship" commit). The genuine change to the owned test is a pure insertion: it adds a relative-clause delegated-authorship block, where a source that frames execution as delegated ("Managed vendor Acme, which built Y", "Oversaw contractors who developed Y") followed by a rewrite claiming "Wrote Y" or "Developed Y" must produce an authorship claim AND verdict block. That strengthens coverage in the block direction and cannot admit fabrication. The earlier named-vendor escalation test is upstream and untouched, and the attribution-kept (pass) and no-over-block assertions remain.

One benign hygiene nit, not a blocker: the new block overwrites the shared source fixture, so two downstream pass-direction tests now run against the relative-clause context rather than the original delegated one. That only shifts the backing context for pass-direction assertions (dilution there is the opposite of letting fabrication through), and no block-direction test reads the shared file after the mutation. Worth a separate fixture on a later pass.

The code substance here is verify-cv-facts.mjs, a source-of-truth surface outside our owned paths, so this is a test-axis review: the owned test is sound and the anti-fabrication coverage is genuinely extended, but the scope and routing of the change itself is the maintainer's call, not something my approval gates. Holding formal approval until the branch is finalized, nothing blocking from the tests/ side.

verify-cv-facts.mjs
```diff
@@ -522,7 +606,7 @@ export function verifyFacts(targetText, {
   const invented = [...targetClaims].filter(claim => !allowed.has(claim));
   const sourceNormalized = normalizeFact(stripMarkup(sourceText));
   const allowedFacts = new Set(config.allow_facts.map(normalizeFact));
-  const unsupportedFacts = factClaims(targetText)
+  const unsupportedFacts = [...factClaims(targetText), ...delegatedAuthorshipClaims(targetText, sourceText)]
```
nikitacometa, line 609: `sourceContainsFact` drops an authorship claim whose words sit in the vendor clause: `Built WebGL implementation for kiosk.` vs `…vendor Acme, which built WebGL implementation for kiosk.` → helper flags, verdict pass.

verify-cv-facts.mjs
```diff
@@ -312,6 +324,78 @@ export function factClaims(text) {
...
+    .map(statement => ({
+      statement,
+      tokens: new Set(attributionTokens(statement)),
+    }));
+  if (!delegatedSources.length) return [];
+
+  const claims = [];
+  for (const statement of factStatements(targetText)) {
+    // Keeping the third-party attribution is not an authorship escalation.
+    if (DELEGATED_PARTY_RE.test(statement)) continue;
```
nikitacometa, line 375: Any role noun anywhere in the sentence skips it: `Built the WebGL implementation for the kiosk while contractors handled catering.` passes.

verify-cv-facts.mjs
```diff
@@ -312,6 +324,78 @@ export function factClaims(text) {
   return claims;
 }
 
+/** Split generated/source documents into bounded statements for attribution checks. */
+function factStatements(text) {
+  const withLineBoundaries = String(text ?? '').replace(/\r?\n+/g, '. ');
```
nikitacometa, line 329: Single newlines are soft wraps in cv.md. Treating them as sentence ends splits the vendor from the project and the two-token link fails.

tests/nonmetric-fact-gate.test.mjs
```diff
@@ -116,6 +116,86 @@ try {
...
+
+  const relativeClauseSource = [
+    'Managed vendor Acme Interactive, which built the WebGL implementation for an in-store kiosk.',
+    'Oversaw contractors who developed the onboarding automation in Node.js.',
+  ].join('\n');
+  const relativeClauseCases = [
+    ['Wrote the WebGL implementation for an in-store kiosk.', 'vendor relative clause is treated as delegated execution'],
+    ['Developed the onboarding automation in Node.js.', 'contractor relative clause is treated as delegated execution'],
+  ];
+  writeFileSync(source, relativeClauseSource);
```
nikitacometa, line 147: After this write, the "unrelated source-backed direct work" case below has no direct-work source left; it passes on token count alone.

nikitacometa (changes requested):
Confirmed at 729ddf53 that @chipoto69's two relative-clause cases block now, including the same-verb `Oversaw contractors who developed … → Developed …`. Then probed around the detector; five results are worth fixing before this ships as an anti-fabrication gate.

1. Same verb, no articles, passes end to end. Source `Managed vendor Acme, which built WebGL implementation for kiosk.`, target `Built WebGL implementation for kiosk.`: `delegatedAuthorshipClaims` flags it, then verify-cv-facts.mjs:609-611 runs the claim through `sourceContainsFact`, finds the vendor's words in the source and drops it. Verdict `pass`. The article-bearing cases block only because the normalizer strips `the` from the claim and not from the source. Skip the substring filter for `authorship` claims; the helper already did the attribution check.

2. Any third-party noun in the target skips the whole statement (line 375). `Built the WebGL implementation for the kiosk while contractors handled catering.` → no claim, pass.

3. The comma-spanning relative clause blocks a mixed statement. Source `Managed vendor Acme, which supplied the visual assets, and built the WebGL implementation for an in-store kiosk.` (the candidate built it), target `Wrote the WebGL implementation…` → block. The description says mixed statements fail open; `THIRD_PARTY_EXECUTION_RE` reaches 120 chars past `which` and picks up the candidate's verb.

4. Soft-wrapped markdown loses the evidence. `factStatements` turns every newline into a sentence end (line 329), so `Managed vendor Acme\nthrough the WebGL implementation for the kiosk.` leaves `Acme` alone in one statement and the project in another; `Wrote the WebGL implementation for the kiosk.` passes. cv.md is markdown.

5. Dotted names truncate the claim. `DIRECT_AUTHORSHIP_CLAIM_RE` stops at `.`, so `Built the Node.js service implementation.` captures `Node` and never reaches two tokens; passes against `Directed vendor Acme through the Node.js service implementation.`

Test file: line 147 overwrites the source with `relativeClauseSource`, so the later "unrelated source-backed direct work" case runs with no pipeline evidence in the source and passes on token count alone. Write the source back before that assertion.

PR #3662, opened 2026-09-02
fix(web): never type a generated answer into a visa or salary field

EliRobinson:
## What does this PR do?

Makes the "never auto-fill legal / visa / work-authorization / salary / demographic fields" rule something the code enforces, on the path that types into a real employer's form.

Today that rule rests on two things, and both of them are the planner agreeing to refuse:

- one bullet in `buildAnswerPrompt` telling it never to fill those fields
- `needs_confirmation` on its reply

Neither is an enforcement point. Trace what happens when a planner declines to refuse and returns a fluent, confident, entirely invented sentence about the candidate's immigration status with `needs_confirmation: false`:

1. `api/apply/prefill` emits the planner's **raw** object (it does not normalise it)
2. `apply-provider` reads `value` out of it into the answers map
3. `session.ts` `fillFields` fills anything whose value is not the empty string

So the only thing standing between an invented visa answer and a real application is that the planner happened to return `""`. An invented answer is not an empty one.

## The guard that already existed was right, just too narrow

`session.ts` had one deterministic check, on consent-worded **checkbox** labels, and its comment states the reasoning exactly:

> Defense-in-depth: NEVER auto-tick a legal consent/agreement checkbox. (The planner already flags these `needs_confirmation`; this guarantees it even if it slips.)

That reasoning was never checkbox-specific. This PR generalises it: one tested predicate, applied to every field type, before any value is entered. A match sends the field back to the human instead of filling it.

`sensitive-questions.mjs` covers six categories: visa, work-authorization, salary, demographic, legal, consent.

## Notes on the shape

- **The consent category is a strict superset of the regex it replaces.** A checkbox labelled only `Terms *` was caught by that regex's bare `terms`, so the word stays. A lookbehind excludes `in terms of`, so an ordinary question ("In terms of impact, what are you proudest of?") is not swept up. Both directions are pinned by tests.
- **No pattern carries the `g` flag, and a test says so.** A shared module-level `g` regex carries `lastIndex` between `.test()` calls and starts matching every *other* field, silently and intermittently, in the one direction that lets a sensitive field through.
- **The list leans wide on purpose.** A false positive costs the candidate one answer they were going to write themselves. A false negative is a fabricated answer about their immigration status sent to an employer, and it looks exactly like a real one until after it has been sent.
- **Plain `.mjs` with no imports**, following `extract-json-object.mjs` (#3142): `web/`'s runner is `node --test tests/**/*.test.mjs`, so a `.ts` file or one reaching through the `@/` alias cannot be tested at all. `session.ts` imports `playwright-core`, which is exactly why the predicate cannot live there and be tested.

Nothing here submits anything. The human still reviews every field, and now also answers the ones that were always theirs.

## Related issues

Refs #2994.

Found while reviewing **#2998**, which adds an Application Questions panel and its route. That PR carries an identical copy of `web/src/lib/apply/sensitive-questions.mjs` and applies it to the panel and `/api/answers`; this PR applies the same module to the apply flow. **The two are independent and can merge in either order** — the file is byte-identical on both branches, so git resolves the add/add without a conflict. Whichever lands second contributes only its own call site.

Opened separately rather than folded into #2998 on the reasoning in that PR's thread: the apply flow is a different surface from the report panel, and this is the more dangerous of the two (a report is a local file; a form is an application). It deserves its own review rather than riding along with a UI feature.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] I ran `node test-all.mjs` and all tests pass (7701 passed, 0 failed)
- [x] `web`: `pnpm test` 471 passed, 0 failed · `tsc --noEmit` clean · `pnpm build` clean
- [x] My changes respect the [Data Contract](https://github.com/santifer/career-ops/blob/main/DATA_CONTRACT.md) (system-layer files only)

### DIFF
--- a/web/src/lib/apply/sensitive-questions.mjs
+++ b/web/src/lib/apply/sensitive-questions.mjs
@@ -1,0 +1,120 @@
+/**
+ * sensitive-questions.mjs - the questions this system never answers for the user.
+ *
+ * Legal, visa, work-authorization, salary and demographic questions are the
+ * candidate's own to answer. That has always been the rule, and until now it
+ * rested entirely on the planner honouring one bullet in its prompt and setting
+ * needs_confirmation on those fields. That is the planner agreeing to refuse.
+ *
+ * A planner that returns needs_confirmation:false instead - a refusal arriving
+ * as a fluent, confident, entirely invented sentence about the candidate's
+ * immigration status or pay - had its value used: typed into a real employer's
+ * form, or written into a report and re-read before the next application. A
+ * prompt instruction is not an enforcement point.
+ *
+ * So the rule is a predicate, tested, applied at every point where a generated
+ * answer could reach a field. It is the same defense-in-depth the apply flow
+ * already applied to legal consent checkboxes, generalized from one bespoke
+ * regex at one call site to one policy every caller shares. Callers must not
+ * re-implement it; a second copy is a second thing to forget.
+ *
+ * The list leans deliberately wide. A false positive costs the candidate one
+ * answer they were going to write themselves anyway. A false negative is an
+ * invented answer about their immigration status sent to an employer, and it
+ * looks exactly like a real one until after it has been sent.
+ *
+ * Plain .mjs and no imports, same reason as extract-json-object.mjs: `web/`'s
+ * test runner is `node --test tests/**\/*.test.mjs`, so a .ts file or one
+ * reaching through the `@/` alias cannot be tested at all.
+ */
+
+/**
+ * @typedef {"visa" | "work-authorization" | "salary" | "demographic" | "legal" | "consent"} SensitiveCategory
+ */
+
+/**
+ * Category patterns, in the order they are tried.
+ *
+ * No `g` flag on any of them, deliberately: a `g` regex carries `lastIndex`
+ * between `.test()` calls, so a shared module-level pattern would start matching
+ * every other question. That failure is silent and intermittent, which is the
+ * worst possible shape for a guard whose whole job is to never miss one.
+ *
+ * @type {Array<{category: SensitiveCategory, pattern: RegExp}>}
+ */
+export const SENSITIVE_PATTERNS = [
+  {
+    // Immigration status. "sponsor" carries the near-universal phrasing "will you
+    // now or in the future require sponsorship for employment".
+    category: "visa",
+    pattern:
+      /\b(visas?|sponsorship|sponsor|immigration|work permits?|h-?1-?b|tn visa|green card|citizenship|citizen of|nationality|residency status|permanent resident)\b/i,
+  },
+  {
+    // Right to work. Written as prefixes rather than whole words because the
+    // forms spell it a dozen ways ("authorised", "authorization", "authorized").
+    category: "work-authorization",
+    pattern:
+      /(\bwork authoris|\bwork authoriz|\bemployment authoris|\bemployment authoriz|\bauthoris(?:ed|ation) to work|\bauthoriz(?:ed|ation) to work|\blegally (?:able|entitled|authoris|authoriz)|\bright to work\b|\beligib(?:le|ility) to work)/i,
+  },
+  {
+    // Pay. "compensation" is included knowing it also catches a question about
+    // designing someone else's comp plan; that answer is one the candidate can
+    // write, and the alternative is an invented salary expectation.
+    category: "salary",
+    pattern:
+      /\b(salary|salaries|compensation|remuneration|current pay|expected pay|desired pay|pay expectation|pay range|expected rate|desired rate|rate expectation|hourly rate|day rate|equity expectation|bonus expectation)\b/i,
+  },
+  {
+    // Protected characteristics, plus the self-identification blocks that carry
+    // them. Accommodation and medical questions sit here for the same reason.
+    category: "demographic",
+    pattern:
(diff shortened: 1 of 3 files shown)


luochen211 (changes requested):
The predicate itself and its focused tests look sound, but the enforcement point still trusts client-supplied metadata. `POST /api/apply/fill` accepts `fields` from the request body and `fillSession()` builds `byId` from that array, so the sensitive decision is made from `meta.label` supplied by the caller rather than the authoritative `s.fields` captured when the session opened. A caller can send the real sensitive field id with a benign label (for example, relabel `Visa sponsorship` as `Portfolio`) and the generated value reaches the locator. The session already stores the extracted fields, so please make the refusal decision from the server-side session record (and ideally use that record for type/locator metadata generally), treating client fields only as display/input state if they remain necessary. Please add a regression that posts or passes a relabeled sensitive field and proves it is still refused. I ran `node --test web/tests/lib/sensitive-questions.test.mjs`; all 7 current tests pass, but they exercise only the predicate and therefore cannot catch this boundary bypass.

web/src/lib/apply/session.ts
```diff
@@ -417,11 +418,23 @@ export async function fillSession(
...
+    // immigration status sets needs_confirmation false, and the only thing then
+    // standing between that value and a real employer's form was the
+    // `value === ""` check above. An invented answer is not an empty one.
+    //
+    // Was a regex matching consent-worded CHECKBOX labels only. The same
+    // reasoning covers every field type, and the shared predicate states it once
+    // so the policy has one home instead of one per call site. Consent wording
+    // is still covered: it is the predicate's `consent` category, unchanged in
+    // meaning.
+    if (isSensitiveQuestion(meta.label || "")) {
```
nikitacometa, line 436: This also refuses values the human typed in the UI for `needs_confirmation` fields (apply-view.tsx:567/592, same `answers` map), and a refusal here is an `ok: false` step that trips the `agentFill` escalation in apply-provider.tsx:335 with the refused value still in the payload. Gate the planner's output at prefill instead.

web/src/lib/apply/sensitive-questions.mjs
```diff
@@ -0,0 +1,120 @@
...
+    // consent-checkbox regex covered, and it must stay a superset of it: a
+    // checkbox labelled only "Terms *" was caught by that regex's bare `terms`.
+    //
+    // The lookbehind is what makes keeping that word affordable. This predicate
+    // now runs on free-text questions too, and "In terms of impact, what are you
+    // proudest of?" is an ordinary question that must stay draftable. Excluding
+    // the one idiom keeps the label coverage without eating the question.
+    category: "consent",
+    pattern:
+      /\b(i (?:have )?read|i agree|i consent|i accept|consent to|privacy (?:notice|policy|statement)|(?<!\bin )terms\b|gdpr|data protection)\b/i,
```
nikitacometa, line 93: `I acknowledge the changes in terms and conditions` → `null` here, `true` under the regex this replaces. Add `terms and conditions` explicitly.

web/src/lib/apply/sensitive-questions.mjs
```diff
@@ -0,0 +1,120 @@
...
+    // designing someone else's comp plan; that answer is one the candidate can
+    // write, and the alternative is an invented salary expectation.
+    category: "salary",
+    pattern:
+      /\b(salary|salaries|compensation|remuneration|current pay|expected pay|desired pay|pay expectation|pay range|expected rate|desired rate|rate expectation|hourly rate|day rate|equity expectation|bonus expectation)\b/i,
+  },
+  {
+    // Protected characteristics, plus the self-identification blocks that carry
+    // them. Accommodation and medical questions sit here for the same reason.
+    category: "demographic",
```
nikitacometa, line 71: `\brace\b` catches `Describe a race condition you debugged`, and `Are you Hispanic or Latino?` is missed. Swap the bare `race` for `racial|your race|race and ethnicity|hispanic|latino`.

nikitacometa (changes requested):
@luochen211's boundary point holds at 882b5ad2: fill/route.ts:20 takes `fields` from the body, session.ts:370 builds `byId` from it, and `Session` (session.ts:114) already carries the extracted `fields`. Three more things, traced on this branch.

1. The refusal feeds the agent path. A refused field is a step with `ok: false`. apply-provider.tsx:335 escalates when `okCount === 0` (or on a mismatch with fewer than half ok) by calling `agentFill`, which posts every non-empty answer with its label to `/api/apply/drive` with `goal: "full"` (apply-provider.tsx:353-357). The drive route has no sensitive check. A form with one visa question and a planner that answered "No": fill refuses, okCount is 0, the agent types "No". The guard makes that path more likely, because refusals count as failures.

2. It refuses the human's own answers. apply-view.tsx:567/592 lets the user type into a field marked "You fill this one."; that value rides the same `answers` map to fill, and session.ts:436 refuses it by label. The one field the UI told the user to answer is the one it will not type. The gate belongs at prefill, server-side from `s.fields`: blank the planner's value on a sensitive label and keep `needs_confirmation`, then fill types whatever the human confirmed. That also settles the boundary question, since fill no longer has to decide.

3. The predicate. At this head, `Are you Hispanic or Latino?` (the standard US EEO wording), `Are you legally permitted to work in the United States?`, `What hourly wage are you seeking?` and `Have you ever pleaded guilty to an offence?` all return `null`. `I acknowledge the changes in terms and conditions` was caught by the old checkbox regex and is not now (the `in ` lookbehind), so consent is not a superset. And `race` as a whole word refuses `Describe a race condition you debugged.`, a normal engineering-form question.

Suggested shape: enforce in `/api/apply/prefill` and `/api/apply/drive` from `Session.fields`; make a policy refusal a `skipped` step so it cannot trigger escalation; add `hispanic|latino|wage|permitted to work|pleaded|terms and conditions`, and scope `race` to `racial|your race|race and ethnicity`.

Issue #2570, opened 2026-08-07
companyMatch() partial-name matching never fires for non-Latin company names


barkaaa:
**Title:** `companyMatch()` partial-name matching never fires for non-Latin company names

## Summary

`companyMatch()` in `scan.mjs` has two branches: exact equality on a punctuation-stripped key, and a containment fallback that lets a short window name match a longer official name (`Acme` vs `Acme Corp`). After #2569 made the keys Unicode-aware, the exact branch works for any script — but the containment fallback cannot fire for names in scripts that are not space-delimited.

**Scoping correction (this issue was originally filed as "non-Latin", which was too broad).** #2569 replaced `\b` with Unicode lookarounds over `[\p{L}\p{M}\p{N}]`, which fixed containment for **accented Latin** names — `Nestlé Deutschland` vs `Nestlé`, `Ørsted Energy` vs `Ørsted`. Those work now. What remains is the set no anchor rule can serve: Japanese, Chinese and Korean write without spaces, so there is no boundary between a corporate form and the name it qualifies.

## Why

After #2569 the fallback anchors on `(?<![\p{L}\p{M}\p{N}])` / `(?![\p{L}\p{M}\p{N}])`. For a space-delimited language that is exactly right. For Japanese there is no boundary to assert at all — the character before `メルカリ` in `株式会社メルカリ` is 社, a letter:

```js
// what #2569 ships
new RegExp('(?<![\\p{L}\\p{M}\\p{N}])メルカリ(?![\\p{L}\\p{M}\\p{N}])','u')
  .test('株式会社メルカリ')                                    // false

// and no other anchor rule helps
new RegExp('\\bメルカリ\\b').test('株式会社メルカリ')             // false (\b is ASCII-only)
new RegExp('(?:^|\\s)メルカリ(?:\\s|$)','u').test('株式会社メルカリ')  // false (no spaces)
'株式会社メルカリ'.includes('メルカリ')                          // true  (unbounded — see below)
```

## Impact

At the only call site (`buildCooldownFilter`, fed by `re_apply_windows` in `config/profile.yml`), a user has to write the company name exactly as the ATS reports it. `メルカリ` will not match a board whose org name is `株式会社メルカリ`, so the re-apply cooldown silently doesn't apply.

[...]

PR #3666, opened 2026-09-02
fix(merge-tracker): a corporate-suffix variant is the same employer (#3665)

### LINKED ISSUE #3665: merge-tracker: a corporate-suffix variant of one employer name never matches, so the same opening becomes two tracker rows
### Code of Conduct

### Existing issues

### What happened?

One employer entering the tracker under two slightly different display names never matches itself, so a routine cross-board evaluation appends a duplicate row even when the role title is identical.

`normalizeCompany()` is `normalizeTextKey()` from `tracker-parse.mjs`. It case-folds, NFKC-normalizes and strips everything outside `\p{L}\p{M}\p{N}`, but it does not fold corporate suffixes, and every company comparison in the tracker pipeline is exact equality on that key:

```js
> normalizeTextKey('Acme')
'acme'
> normalizeTextKey('Acme Technologies')
'acmetechnologies'
> normalizeTextKey('Acme Inc.')
'acmeinc'
```

Boards disagree about the legal form all the time. The same employer is `Acme` on one ATS, `Acme Technologies` on the next, `Acme Inc.` in a recruiter's posting. Three consumers compare that key directly:

`merge-tracker.mjs`, `companiesMatch()` at line 308, reached from the fuzzy company plus role tier at line 1214. The tier never fires for a suffix variant, so a second row is written.

`dedup-tracker.mjs` at line 333 groups candidates on `normalizeCompany(entry.company)`, so the two rows are never in the same group and the duplicate survives a cleanup run.

`verify-pipeline.mjs` Check 2 at line 157 keys on `normalizeTextKey(company) + '::' + normalizeTextKey(role)` and prints "No exact duplicates found" on a tracker that holds the pair. That is the part I would most want your read on, since it is a false green on exactly the case someone runs the health check to catch.

This is the second half of what I found while working [#3652](https://github.com/career-ops-hq/career-ops/issues/3652), flagged there in [this comment](https://github.com/career-ops-hq/career-ops/issues/3652#issuecomment-5503051782) rather than folded into [#3653](https://github.com/career-ops-hq/career-ops/pull/3653), which stays one behavior change. The two are independent: [#3653](https://github.com/career-ops-hq/career-ops/pull/3653) stops an aggregator URL from blocking the fuzzy tier, and this stops the company key from failing once that tier is allowed to run. Fixing either alone leaves the pair split.

Related but not the same defect: [#2570](https://github.com/career-ops-hq/career-ops/issues/2570) covers `companyMatch()` in `scan.mjs` and the CJK corporate forms, where there is no token boundary to anchor on. This issue is about the space-delimited case in the tracker merge.

### Steps to reproduce

Tracker starts with the employer as the ATS spelled it:

```
| 1 | 2026-06-01 | Acme Technologies | Director of Marketing | 4.0/5 | Evaluated | ❌ | [1](reports/1-acme.md) | n |
```

Addition `batch/tracker-additions/2-acme.tsv` (tab separated), same opening, second board, short form of the name, different row and report numbers so only the fuzzy tier is left:

```
2	2026-06-03	Acme	Director of Marketing	Evaluated	4.1/5	❌	[2](reports/2-acme.md)	n
```

Run `node merge-tracker.mjs`. Observed: two rows. Then `node verify-pipeline.mjs` reports no duplicates, and `node dedup-tracker.mjs` finds nothing to merge.

### Expected behavior

The pair should merge, and a pair that only looks similar should not. Suffix folding is the kind of change that over-merges if it is written loosely, and over-merging is worse here than the duplicate it prevents: `dedup-tracker.mjs` promotes the most advanced status and merges notes onto the keeper and then drops the losing line, so a wrong grouping costs that row's score, date, report link, PDF flag and URL, in a file that is gitignored with no backup.

So I would rather put the shape in front of you explicitly. Here is what I propose, and I am happy to build a different one if you would rather.

**Keep the existing exact-match tier untouched and add a narrower second tier.** The suffix-insensitive comparison would run only inside the fuzzy company plus role tier in `merge-tracker.mjs`, where the role title already fuzzy-matches, no requisition ID mismatch and no employer-board URL mismatch has proved the rows distinct, and the `?` blind-employer channel guard still applies. Tiers 1 and 2 (report number, entry number) keep exact company equality.

**Match on a token prefix, not on a stripped key.** Tokenize both names with `normalizeTextKey(name, ' ')`, which already carries the Unicode and Turkish dotted-I handling from #2445 and #2736, then treat them as the same employer only when the shorter token list is a prefix of the longer one and every remaining token in the longer one is a known corporate-form word. `Acme` and `Acme Technologies Inc.` match. `Acme Robotics` and `Acme Financial` do not, because the extra tokens are not corporate forms. `Acme Solutions` and `Acme Technologies` do not, because neither is a prefix of the other.

That last case is why I am not just calling `normalizeCompanyName()` from `invite-match.mjs` directly, even though it already does chained suffix stripping. It maps both of those to the same key:

```js
> normalizeCompanyName('Acme Technologies')
'acme'
> normalizeCompanyName('Acme Solutions')
'acme'
```

For matching a loosely quoted recruiter email against a tracker row that is the right trade. For deciding whether to fold two rows together it is not, so the prefix rule keeps the vocabulary and tightens the rule.

**Reuse the vocabulary rather than adding a second list.** `invite-match.mjs` already holds `LEGAL_SUFFIXES` (incorporated, inc, corporation, corp, company, co, limited, ltd, llc, llp, lp, plc) and `GENERIC_DESCRIPTORS` (group, holdings, technologies, technology, solutions, canada, international). They are currently module-private. Exporting them and importing both here keeps one definition, which is the point #2445 and #2570 both make about identity drift between scripts.

**Refuse a short stem.** If the shared prefix is shorter than three characters, fall through to today's behavior. `HP` and `HP Inc.` would then stay two rows, which is a miss, but a two-character stem is weak identity and the cost of refusing is only the duplicate row you already had.

Three things I would like your call on:

1. Scope. I would send `merge-tracker.mjs` alone first, since that is prevention and the other two are cleanup and reporting. `dedup-tracker.mjs` deletes a row, so I would rather it move second and separately. `verify-pipeline.mjs` needs a decision about what it should say, since a suffix variant is a weaker signal than an exact key collision and probably wants its own wording rather than being folded into the existing warning. Do you want all three at once instead?
2. Where the helper lives. Local to `merge-tracker.mjs` while it has one caller, or in `tracker-utils.mjs` next to `normalizeCompany` from the start if you already know you want the other two consumers.
3. Vocabulary. The current list is English-only, and the repo ships DACH, FR, JA, AR, TR and HI market modes. Adding gmbh, ag, sa, bv, nv, oy, ab, aps, spa, srl and pty would help those markets, but it changes `invite-match.mjs` behavior too, so I have left it out. Say the word and I will send it as its own change.

I am building the merge-tracker half now, with tests, and will open the PR shortly so there is something concrete to react to rather than a design sketch. The three questions above are genuine, and I am happy to rework or drop it if you want a different shape.

### CLI tool

Claude Code

### OS and Node.js version

macOS 26.6 / Node 26.8.1

### Output of npm run doctor

Not included on purpose. This is not a setup problem, and the doctor output carries local paths. The reproduction above runs against a throwaway tracker via `CAREER_OPS_TRACKER` and needs nothing from a configured workspace.

rubicon:
## What does this PR do?

Teaches the fuzzy company plus role tier in `merge-tracker.mjs` that "Acme" and "Acme Technologies Inc." are one employer, so an opening that reaches the tracker from two boards under two spellings updates its row instead of adding a second one. The exact-match tier and the two tiers above it are unchanged.

## Related issue

Closes #3665

## Type of change

- [x] Bug fix
- [ ] New feature

## The change

`normalizeCompany()` is `normalizeTextKey()`, which folds case, punctuation and width but not corporate forms, so `Acme` keys to `acme` and `Acme Technologies` keys to `acmetechnologies`. Boards disagree about the legal form constantly, so the tier that exists precisely to catch a re-listing never fired for that pair.

The new tier runs only inside the fuzzy match, after `urlBlocksHeuristic` and before the req-number guard, so it sees a pair whose role title already matches and where neither an employer-board URL nor a requisition ID has proved the rows distinct. The report-number and entry-number tiers keep exact company equality on purpose: those match on an identity claim of their own, and sequence drift between report files and tracker rows is common enough that a wider company match there would let one employer's addition write to another's row.

The rule is a token prefix plus a corporate-form tail, not a stripped key:

```js
normalizeTextKey('Acme Solutions', ' ')     // 'acme solutions'
normalizeTextKey('Acme Technologies', ' ')  // 'acme technologies'
```

Stripping the descriptor before comparing maps both of those to `acme`. That folds two genuinely different employers, and folding is unrecoverable here: `dedup-tracker.mjs` promotes the most advanced status onto the keeper and then drops the losing line, taking its score, date, report link, PDF flag and URL with it, out of a file that is gitignored with no backup. Requiring the shorter name to be a prefix of the longer makes that case structurally impossible, since neither is a prefix of the other. An extra token that is not a corporate form (`Acme Robotics`) also still reads as a different employer, and a stem shorter than three characters falls through to today's behavior, which loses `AB` against `AB Inc.` and is the direction I would rather miss in.

Tokens come from `normalizeTextKey(name, ' ')` rather than a private strip, so the NFKC, any-script and Turkish dotted-I handling from #2445 and #2736 is shared. Scripts written without spaces produce a single token and never reach the prefix rule, so CJK corporate forms are untouched and stay #2570's subject.

The vocabulary is imported from `invite-match.mjs` (`LEGAL_SUFFIXES` and `GENERIC_DESCRIPTORS`, now exported) rather than copied. A second private list is the identity drift #2445 set out to remove. Taking the union of the two lists is safe here in a way chaining them is not inside `normalizeCompanyName()`, because the prefix rule never lets two names disagree on a substantive token.

`normalizeCompanyName()` itself is unchanged, and so is every existing caller of it.

## Interaction with #3653

None in behavior, and they are complementary. #3653 stops an aggregator URL from short-circuiting the fuzzy tier; this stops the company key from failing once that tier is allowed to run. A pair that is both cross-listed on an aggregator and spelled differently needs both.

This branch is cut from `main` at `2e33ea86` and does not contain #3653. The two touch `merge-tracker.mjs` in different places (#3653 at `urlDiffers`, around line 1150, plus `url-key.mjs`; this one at the import block, at `companiesMatch`, and at the company check inside the fuzzy tier). The only overlap is the import block at the top of the file, where #3653 adds `isAggregatorUrl` to the `url-key.mjs` line and this adds `normalizeTextKey` to the `tracker-parse.mjs` line two lines above it. Whichever merges second may want a one-line rebase there. Happy to rebase this onto #3653 if you would rather land them in that order.

## How it was verified

New suite `tests/merge-tracker-company-suffix.test.mjs`, auto-discovered under `tests/`, driving the real `merge-tracker.mjs` CLI against a temp tracker through the `CAREER_OPS_TRACKER` and `CAREER_OPS_ADDITIONS` hooks, in the shape `tests/merge-tracker-url-dedup.test.mjs` already uses. Every fixture is a placeholder name.

Against unfixed `main` the two duplicate cases fail and the guards pass:

```
merge-tracker — corporate-suffix company variants (#3665)
  ❌ THE BUG: a generic-descriptor variant of one employer merges into its row — expected the row to be UPDATED, got 2 rows (duplicate)
  ❌ THE BUG: a legal-suffix variant merges too, and chains — expected 1 row, got 2
  ✅ THE OVER-MERGE GUARD: two employers sharing a stem stay two rows
  ✅ THE OVER-MERGE GUARD: an extra word that is not a corporate form stays distinct
  ✅ THE OVER-MERGE GUARD: a stem under three characters is refused
  ✅ req-number mismatch still proves two postings apart (#1524 precedence)
  ✅ an employer-board URL mismatch still proves two postings apart
  ✅ a role that does not fuzzy-match is unaffected by the wider company match
  ✅ CONTROL: exact company match still merges, unchanged
  ✅ CONTROL: unrelated employers still stay apart
  ✅ CONTROL: a blind-employer row cannot reach the tier (empty stem)
```

Guards that pass before and after prove nothing on their own, so I also ran them against the obvious wrong implementation, `companiesMatchIgnoringCorporateForm` replaced by an equality test on `normalizeCompanyName()`. That version passes both bug cases and fails two guards:

```
  ✅ THE BUG: a generic-descriptor variant of one employer merges into its row
  ✅ THE BUG: a legal-suffix variant merges too, and chains
  ❌ THE OVER-MERGE GUARD: two employers sharing a stem stay two rows — a shared stem is not a shared employer
  ❌ THE OVER-MERGE GUARD: a stem under three characters is refused — short stems fall through to the previous behavior
```

Full `node test-all.mjs`: 7735 passed, 0 failed, 13 warnings, all of which are the pre-existing ones on `main` (funding.json and HIRED.md name matches, the intentional fixture warnings in the merge and font suites). Also smoke-tested `merge-tracker.mjs` with no flags, `--migrate` and `--backfill-urls` after the new import, since `invite-match.mjs` has top-level argument handling; it is behind `isMainModule`, so nothing changes.

## Not in this PR, on purpose

The same key is compared for exact equality in two more places, and both are described in #3665: `dedup-tracker.mjs` line 333 cannot group the pair, and `verify-pipeline.mjs` Check 2 reports "No exact duplicates found" on a tracker that holds it. I left them out because this one is prevention while those are cleanup and reporting, because `dedup-tracker.mjs` deletes a row and deserves its own review, and because what the health check should say about a suffix variant is a wording decision rather than a code one. Say the word on the issue and I will send either or both, and move the helper into `tracker-utils.mjs` next to `normalizeCompany` if you want more than one caller.

## Checklist

### DIFF
--- a/merge-tracker.mjs
+++ b/merge-tracker.mjs
@@ -22,7 +22,11 @@ import { normalizeReportLink as normalizeLink } from './tracker-links.mjs';
 import { getCareerOpsRoot } from './path-resolver.mjs';
 import { roleFuzzyMatch } from './role-matcher.mjs';
 import { parsePdfIndex } from './find.mjs';
-import { LEGACY_COLMAP, detectColumns, isHeaderRow, resolveScoreStatus, normalizeVia, SEPARATOR_ROW_RE } from './tracker-parse.mjs';
+import { LEGACY_COLMAP, detectColumns, isHeaderRow, resolveScoreStatus, normalizeVia, normalizeTextKey, SEPARATOR_ROW_RE } from './tracker-parse.mjs';
+// Corporate-form vocabulary, shared with invite-match.mjs rather than copied,
+// for the same reason normalizeCompany lives in tracker-utils: a second private
+// list is how company identity drifts between scripts (#2445, #3665).
+import { LEGAL_SUFFIXES, GENERIC_DESCRIPTORS } from './invite-match.mjs';
 import { resolveTrackerPath, resolveWorkspaceRoot, resolvePdfIndexPath, trackerLockDirFor, acquireTrackerLock, writeFileAtomic, normalizeCompany, cell } from './tracker-utils.mjs';
 // Canonical posting-URL key. Kept in its own module so scan.mjs / scan-history
 // can adopt the same key later without the definitions drifting.
@@ -309,6 +313,65 @@ function companiesMatch(a, b) {
   const key = normalizeCompany(String(a));
   if (key !== normalizeCompany(String(b))) return false;
   return key !== '' || String(a).trim() === String(b).trim();
+}
+
+// Words that name a legal form or a generic business descriptor rather than the
+// employer. Both lists come from invite-match.mjs; the union is safe here in a
+// way chaining them there is not, because the prefix rule below never lets two
+// names disagree on a substantive token.
+const CORPORATE_FORM_WORDS = new Set([...LEGAL_SUFFIXES, ...GENERIC_DESCRIPTORS]);
+
+// A one-token stem carries almost no identity, and a two-letter one is usually
+// an initialism that several unrelated employers share. Refusing them costs the
+// duplicate row that exists today; accepting them risks deleting a real one.
+const MIN_STEM_CHARS = 3;
+
+/**
+ * True when two company cells are the same employer written with a different
+ * corporate suffix ("Acme" vs "Acme Technologies Inc.").
+ *
+ * normalizeCompany() folds case, punctuation and width but not corporate
+ * forms, so a suffix variant of one employer never reaches the fuzzy
+ * company+role tier and the same opening is written twice (#3665). This is the
+ * narrow second tier that closes that, used ONLY where the role title already
+ * fuzzy-matches and neither a req-number nor an employer-board URL has proved
+ * the rows distinct. The exact tier above is untouched.
+ *
+ * The rule is a token PREFIX plus a corporate-form tail, not a stripped key.
+ * Stripping suffixes before comparing looks equivalent and is not: it maps
+ * "Acme Solutions" and "Acme Technologies" to the same "acme", which folds two
+ * genuinely different employers and deletes one row. Requiring the shorter name
+ * to be a prefix of the longer makes that structurally impossible, since
+ * neither of those is a prefix of the other. It also means an extra token that
+ * is NOT a corporate form ("Acme Robotics") still reads as a different
+ * employer, which is the conservative direction: over-merging is unrecoverable
+ * here (dedup-tracker.mjs drops the losing line, and applications.md is
+ * gitignored with no backup), while under-merging leaves a duplicate row the
+ * user can already see.
+ *
+ * Tokens come from normalizeTextKey(name, ' ') so this shares the Unicode,
+ * NFKC and Turkish dotted-I handling every other identity key uses (#2445,
+ * #2736) instead of a private strip. Scripts written without spaces produce a
+ * single token and never match here, so CJK corporate forms are unaffected and
+ * remain #2570's subject.
+ *
+ * @param {string} a - Company cell from one side of the comparison.
+ * @param {string} b - Company cell from the other side.
+ * @returns {boolean} True when the two cells name the same employer under a
+ *   different corporate form.
+ */
+function companiesMatchIgnoringCorporateForm(a, b) {
+  const ta = normalizeTextKey(a, ' ').split(' ').filter(Boolean);
+  const tb = normalizeTextKey(b, ' ').split(' ').filter(Boolean);
+  const [stem, full] = ta.length <= tb.length ? [ta, tb] : [tb, ta];
+  // Equal lengths are either the exact tier's business or a genuine
+  // disagreement on a substantive token. An empty stem is the blind-employer
+  // `?` marker and every other punctuation-only cell, which must never match a
+  // named employer.
+  if (stem.length === 0 || stem.length === full.length) return false;
+  if (stem.join('').length < MIN_STEM_CHARS) return false;
+  if (stem.some((token, i) => token !== full[i])) return false;
+  return full.slice(stem.length).every(token => CORPORATE_FORM_WORDS.has(token));
 }
 
 /**
@@ -1211,7 +1274,14 @@ for (const file of tsvFiles) {
       // the #1524 req-number guard, and the tier where an unkeyed addition is
       // also held back from claiming a row whose posting is known.
       if (urlBlocksHeuristic(app)) return false;
-      if (!companiesMatch(app.company, addition.company)) return false;
+      // Corporate-form variants of one employer are accepted HERE and nowhere
+      // else (#3665). The tiers above match on a report or entry number, which
+      // is an identity claim on its own, so a wider company match there would
+      // let sequence drift across two employers write to the wrong row. This
+      // tier already requires a fuzzy role match, and the req-number and URL
+      // guards below and above still get to prove the rows distinct.
+      if (!companiesMatch(app.company, addition.company)
+          && !companiesMatchIgnoringCorporateForm(app.company, addition.company)) return false;
       if (!roleFuzzyMatch(addition.role, app.role)) return false;
       // Cross-channel guard (#1596): unknown-employer rows (`?`) all normalize
       // to the same empty company key, but the same role via two DIFFERENT
--- a/invite-match.mjs
+++ b/invite-match.mjs
@@ -140,7 +140,10 @@ function normalizeStatusKey(status) {
 // True legal-entity suffixes, stripped repeatedly (chained) since a name can
 // legitimately carry more than one ("Acme Holdings Inc." → "acme holdings").
 // These are unambiguous enough that removing several in a row is safe.
-const LEGAL_SUFFIXES = [
+// Exported so merge-tracker.mjs can share the vocabulary rather than keep a
+// second copy of it: a private list there is exactly the identity drift #2445
+// set out to remove.
+export const LEGAL_SUFFIXES = [
   'incorporated', 'inc', 'corporation', 'corp', 'company', 'co',
   'limited', 'ltd', 'llc', 'llp', 'lp', 'plc',
 ];
@@ -151,7 +154,7 @@ const LEGAL_SUFFIXES = [
 // chaining their removal risks collapsing two different companies to the
 // same key. Stripped at most once, and only after legal suffixes are gone —
 // never chained with each other or with LEGAL_SUFFIXES.
-const GENERIC_DESCRIPTORS = [
+export const GENERIC_DESCRIPTORS = [
   'group', 'holdings', 'technologies', 'technology', 'solutions',
   'canada', 'international',
 ];
--- a/tests/merge-tracker-company-suffix.test.mjs
+++ b/tests/merge-tracker-company-suffix.test.mjs
@@ -1,0 +1,170 @@
+// tests/merge-tracker-company-suffix.test.mjs — corporate-suffix company variants.
+//
+// One employer reaches the tracker under two spellings ("Acme" from one board,
+// "Acme Technologies" from the next). normalizeCompany() folds case and
+// punctuation but not corporate suffixes, so the fuzzy company+role tier never
+// fired and a duplicate row was written (#3665).
+//
+// Every case drives the REAL merge-tracker.mjs CLI against a temp tracker via
+// the CAREER_OPS_TRACKER / CAREER_OPS_ADDITIONS env hooks: the merge path is
+// where the bug lives, so the resulting rows are what proves the fix. The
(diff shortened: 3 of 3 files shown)


Scott-Emberson (commented):
Reviewed at head 6a7cd3fd08. This treats a corporate-suffix variant as the same employer without over-merging, and the test pins both directions. The merge case ("Acme Technologies" plus "Acme") asserts one row with the 4.1 score written through, so the merge updates rather than drops or overwrites. The over-merge guards then hold in every direction that matters: equal-length stems with neither a prefix of the other ("Acme Solutions" vs "Acme Technologies") stay two rows, a short stem ("AB" vs "AB Inc.", stem under three chars) stays two, and the distinct-employer signals (req-number mismatch, employer-board URL mismatch, role non-match) all stay two. The implementation matches what the test targets: companiesMatchIgnoringCorporateForm uses a token-prefix plus corporate-form-tail rule rather than a stripped key, and it rejects an empty stem, equal-length stems, and stems under the minimum length, so over-merge is structurally prevented. The shared legal-suffix and generic-descriptor lists are exported from invite-match.mjs so there is no drift between the two.

Owned test verified sound, data-integrity safe with no over-merge. Holding formal approval until the branch is finalized, nothing blocking from the tests/ side.

merge-tracker.mjs
```diff
@@ -1211,7 +1274,14 @@ for (const file of tsvFiles) {
...
       if (urlBlocksHeuristic(app)) return false;
-      if (!companiesMatch(app.company, addition.company)) return false;
+      // Corporate-form variants of one employer are accepted HERE and nowhere
+      // else (#3665). The tiers above match on a report or entry number, which
+      // is an identity claim on its own, so a wider company match there would
+      // let sequence drift across two employers write to the wrong row. This
+      // tier already requires a fuzzy role match, and the req-number and URL
+      // guards below and above still get to prove the rows distinct.
+      if (!companiesMatch(app.company, addition.company)
+          && !companiesMatchIgnoringCorporateForm(app.company, addition.company)) return false;
```
nikitacometa, line 1284: One `find()` for both predicates: a corporate-form row earlier in the file wins over an exact match later. Probe in the body (#1 `Acme Technologies` + #5 `Acme`, addition `Acme` → #1 updated and renamed).

merge-tracker.mjs
```diff
@@ -311,6 +315,65 @@ function companiesMatch(a, b) {
   return key !== '' || String(a).trim() === String(b).trim();
 }
 
+// Words that name a legal form or a generic business descriptor rather than the
+// employer. Both lists come from invite-match.mjs; the union is safe here in a
+// way chaining them there is not, because the prefix rule below never lets two
+// names disagree on a substantive token.
+const CORPORATE_FORM_WORDS = new Set([...LEGAL_SUFFIXES, ...GENERIC_DESCRIPTORS]);
```
nikitacometa, line 322: `Data` vs `Data Solutions` and `Acme` vs `Acme Canada` merge through this set, and `buildRow` then renames the row to the addition's spelling. Legal suffixes are safe here; the generic descriptors are not.

nikitacometa (changes requested):
Ran the suite (green), then drove `merge-tracker.mjs` through `CAREER_OPS_TRACKER` with pairs the tests do not have. Three things, two of them destructive.

1. The merge renames the row. `buildRow` writes `company: addition.company` (merge-tracker.mjs:1371) while `role` keeps the existing value in the fuzzy tiers because "the pairing is only a guess". Company needs the same rule now. Existing `Acme Technologies Inc.` (Applied) + addition `Acme` → the row becomes `Acme`. Existing `Acme` + addition `Acme Canada` → the row becomes `Acme Canada` (`canada` is in GENERIC_DESCRIPTORS). Every company-keyed consumer (`company-history.mjs`, `blacklist.md`, `linkedin-join.mjs`, follow-ups) loses the join on the spelling that was there. Keep `duplicate.company` on a corporate-form match.

2. First match shadows the exact match. Both predicates sit in one `existingApps.find()`, so file order decides. Rows `#1 Acme Technologies` and `#5 Acme` (the pre-existing duplicate this PR exists to prevent), addition `Acme` → merged into #1, #1 renamed to `Acme`, #5 untouched. The tracker now holds two literal `Acme` rows with the same role. Two passes: exact `companiesMatch` over all rows first, corporate-form only when nothing matched.

3. Generic descriptors as a droppable tail. `Data` + `Data Solutions` → one row, renamed `Data Solutions`. invite-match.mjs:151 already says these words can be the name. Legal suffixes are safe to drop; for the generic list I would require a legal suffix in the same tail (`Acme Technologies Inc` vs `Acme`) or keep it out of this tier.

Smaller: `Acme Inc` vs `Acme LLC` stay two rows (equal length) and `Acme L.L.C.` tokenizes to `l l c` and never matches; both are the stated direction, worth a line in the docblock. And merge-tracker.mjs now imports invite-match.mjs, whose `--file` / `--id` checks run at import time (invite-match.mjs:95-112, `process.exit(1)`). scan.mjs already does this, but the vocabulary would sit better in tracker-utils.

Issue #2274, opened 2026-07-29
bug(tracker): a fully localized header defeats detectColumns — ES tracker with a Location column reads `Remote` as the score


LayerC0de:
### Code of Conduct

- [x] I agree to follow the [Code of Conduct](https://github.com/santifer/career-ops/blob/main/CODE_OF_CONDUCT.md)

### Existing issues

- [x] I searched existing issues and this hasn't been reported yet — closest are #946 (which introduced header-name mapping and explicitly asked for "the existing `empresa`/`puesto` ES headers"), #1291 and #1955 (other scripts still using fixed indices). This is a different half of #946 that never landed.

### What happened?

`detectColumns()` maps columns by header name, but gates on **literal** `company` and `role` cells before consulting the alias table:

```js
if (!cells.includes('company') || !cells.includes('role')) continue;
const map = {};
cells.forEach((c, i) => { if (HEADER_ALIASES[c] != null) map[HEADER_ALIASES[c]] = i; });
```

`tracker-aliases.json` ships `"empresa": "company"` and `"puesto": "role"` for exactly this case, but a **fully** localized header never reaches them — the pre-filter rejects it first. `detectColumns()` returns `null` and every caller falls back to `LEGACY_COLMAP`, i.e. the fixed positions #946 replaced.

On a plain 9-column ES tracker the fallback happens to line up, so the bug is invisible. Add the Location column from #946's own use case and the columns shift:

```
EN + Location → mapped, score at index 6
ES + Location → NULL → LEGACY_COLMAP, score read at index 5

ES tracker reads its score cell as: "Remote"  (should be "4.2/5")
```

### Steps to reproduce

Against `origin/main` (`06da05c`), no local changes:

```js
import { detectColumns, LEGACY_COLMAP } from './tracker-parse.mjs';

const es  = '| # | Fecha | Empresa | Puesto | Location | Score | Status | PDF | Report | Notes |';
const row = '| 1 | 2026-01-05 | Acme Corp | Director | Remote | 4.2/5 | Evaluated | ❌ | — | note |';

const map = detectColumns([es]) || LEGACY_COLMAP;   // → LEGACY_COLMAP
row.split('|').map(s => s.trim())[map.score];        // → "Remote"
```

Swap `Fecha/Empresa/Puesto` for `Date/Company/Role` and it maps correctly.

### Expected behavior

[...]

PR #3676, opened 2026-09-02
fix(calibrate): detect tracker columns from the file lines, not the header string

nikitacometa:
## Summary

`calibrate.mjs` reads the tracker through `resolveColumns()`, which takes the file's **lines** and detects the header layout. `loadTrackerRows()` handed it the header **string**: iterated character by character, no character is a header, so it always fell back to the legacy fixed layout. Any tracker whose columns are not the fixed 9 in the fixed order (a `Location` column, the #2274 layout; a localized header with columns moved) reads Score from the wrong cell — the row's status becomes the score cell, the score becomes `null`, and calibrate reports `0 resolved` with exit 0. Nothing looked wrong on the plain 9-column tracker, which is why it went unnoticed.

Fix: pass `lines` (and drop the now-unused `headerLine` lookup — `resolveColumns()` already falls back to the legacy map when no header matches). `loadTrackerRows` is now exported so the regression test can target it directly.

## Test

`tests/calibrate-columns.test.mjs` (auto-discovered): a tracker with `Location` inserted after `Role` parses to `score 4.4 / status Rejected`; the plain 9-column tracker still parses. Reverting the fix turns the first assertion red — `{"score":null,"status":"4.4/5"}` — while the second stays green. `node calibrate.mjs --self-test` unchanged; `node test-all.mjs --quick` passes.

### DIFF
--- a/calibrate.mjs
+++ b/calibrate.mjs
@@ -212,10 +212,9 @@ export function computeCalibration(rows, journals, opts = {}) {
 
 // --- Filesystem assembly --------------------------------------------------
 
-function loadTrackerRows(appsFile) {
+export function loadTrackerRows(appsFile) {
   const lines = readFileSync(appsFile, 'utf-8').split(/\r?\n/);
-  const headerLine = lines.find((l) => isHeaderRow(l));
-  const colmap = headerLine ? resolveColumns(headerLine) : undefined;
+  const colmap = resolveColumns(lines);
   const rows = [];
   for (const line of lines) {
     if (!line.trim().startsWith('|') || isHeaderRow(line) || isSeparatorRow(line)) continue;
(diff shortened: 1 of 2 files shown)


Scott-Emberson (commented):
Reviewed at head 9070c44914. The column-detection fix is pinned properly. loadTrackerRows now builds the column map from the parsed line array rather than the raw header string, and the test asserts a positive mapping on a tracker with Location inserted (row 1 score 4.4 and status Rejected, row 2 3.1 and Interview) plus the legacy nine-column layout (num 7, score 3.8, status Offer). Those reddens on a revert: header-string splitting finds no header, falls back to the fixed legacy layout, and with Location at index 4 it reads Score from the Remote/Berlin cell (NaN, not 4.4) and Status one column off, so the equality assertions fail. Same approach as the earlier #3496 fix but self-contained and correct on its own.

Issue #3419, opened 2026-08-27
test-all.mjs's syntax gate reads only the repo root, and narrows silently when files move


vliggio:
## Problem

`test-all.mjs`'s section 1 presents itself as the suite's syntax gate. It reads one directory:

```js
// test-all.mjs:225
const mjsFiles = readdirSync(ROOT).filter(f => f.endsWith('.mjs'));
```

`readdirSync` does not recurse, so the gate covers **121** of the repo's **596** `.mjs` files. What it never sees:

| directory | `.mjs` files |
|---|---:|
| `tests/` | 263 |
| `providers/` | 90 |
| `web/` | 67 |
| `plugins/` | 12 |
| `lib/` | 12 |
| `scripts/`, `scaffolder/`, `utils/`, `seeds/`, `batch/` | 7 |

## This is redundancy, not a hole

Worth stating up front so nobody treats it as urgent. `scripts/check-syntax.mjs` walks recursively and covers all 596, and `.github/workflows/test.yml:34` runs it as `npm run lint` in the same required job, *before* `test-all.mjs`. So CI does syntax-check everything. A broken file under `tests/` would also surface as a contained import throw once discovery reached it.

The problem is that the gate is **silently partial and quietly shrinking**, in a way its output actively conceals. It prints `{file} syntax OK` per file, so a reader watching 121 green lines scroll past has no way to tell the directory holding 263 files was never opened. And someone running `node test-all.mjs` locally without `npm run lint` gets a fraction of the coverage the section appears to promise.

## How it surfaced

Measured while accounting for a two-check discrepancy in #3411. Moving `test-salary-filter.mjs` and `test-trust-validator.mjs` into `tests/` should have changed the total by −2 registrations +231 assertions = **+229**; the actual delta was **+227**. Diffing the pass labels between the two runs found the missing pair:

```
test-salary-filter.mjs syntax OK
test-trust-validator.mjs syntax OK
```

present before the move, absent after, with nothing gaining a replacement — `syntax OK` went 123 → 121. The two files did not stop being syntax-checked by CI; they stopped being visible to *this* gate, because they left the root.

That is the shape of the bug: the gate narrows every time a file moves out of the root, and never says so. #3306's series moved eleven suites out, so eleven files left this gate silently. The same happened, unnoticed, with #3388's nine.

## Options

[...]

PR #3681, opened 2026-09-02
fix(tests): skip nested checkouts by marker, not by the name .git (#3499)

### LINKED ISSUE #3499: bug(tests): a git worktree nested under the repo root is walked as source — .git is skipped by name, but a worktree marks itself with a .git file
## Summary

Every repo-walking helper here skips a directory **named** `.git`. A linked git worktree has no `.git` directory — it has a `.git` **file** pointing at the parent's gitdir. So the one exclusion meant to keep git's own storage out of the walk slides straight past a worktree, and the walkers descend into a complete second copy of the repository at a different commit.

`lib/mjs-files.mjs` — "the one definition of *every `.mjs` file in this repository*" — returns **1097 files in a 576-file checkout** when one worktree is nested under the root. 90% inflation, and the extra 521 files are a stale copy of the same source.

Consequences vary by consumer, which is the uncomfortable part:

- `scripts/check-syntax.mjs` and `test-all.mjs` section 1 silently syntax-check ~2× the files, reporting a count that depends on whether the developer happens to have a worktree open.
- `tests/local-today-gates.test.mjs` **fails outright**, naming correct files as violations.
- `tests/main-guard-convention.test.mjs` passes — but only because the stale copy happens to satisfy that particular convention.

So whether a nested worktree breaks the suite depends on how old the worktree is and which conventions changed since. A contributor gets a red suite from a commit they didn't write, or a green one that checked the wrong tree.

## Repro

```bash
git worktree add .wt HEAD
node -e "import('./lib/mjs-files.mjs').then(m => console.log(m.collectMjsFiles(process.cwd()).length))"
git worktree remove .wt
```

The count roughly doubles. Any worktree path under the repo root reproduces it; `.claude/worktrees/<name>` is where Claude Code puts them by default, which is how I hit it.

## The failure it produced

On a v1.28.0 worktree nested under a v1.30.0 checkout:

```
AssertionError: scan-history writers disagreeing about what day it is:
  actual: [
    '.claude/worktrees/<name>/scan-ats-full.mjs: stamps `date`, which is not localToday()',
    '.claude/worktrees/<name>/scan-hn.mjs: stamps `new Date().toISOString().slice(0, 10)`, ...',
    '.claude/worktrees/<name>/scan-interamt.mjs: stamps `date`, ...',
    ...
  ]
  expected: []
```

All three files in the actual checkout use `localToday()` correctly. The only violators are the stale copies. The suite prints `🔴 TESTS FAILED — do NOT push/merge until fixed` for source that is already correct, and the paths in the message look enough like repo files to send someone editing them.

## Why this looks like an oversight rather than a design choice

Both walkers already articulate this exact hazard.

`tests/local-today-gates.test.mjs:509`:

```js
// Vendored code, build output, and the suite's own scratch copy of the repo
// (test-all.mjs mkdtemps it under ROOT) would otherwise be walked.
```

That is the same failure mode — a second copy of the repo under ROOT — defended against for the copy the *test suite* creates, and left open for the copy *git* creates.

`lib/mjs-files.mjs`'s `SKIP_DIRS` rationale:

> including them would make the result depend on what a given checkout happens to have run — a clean clone and a working install would disagree about how many files were checked

A nested worktree is precisely two checkouts disagreeing about the file count. And the header's whole subject is the gate silently covering the wrong file set (#3306, #3388, #3411, #3419) — 121 of ~575 then, 1097 of 576 now.

## Scope

Ten walkers recurse from the repo root with a `node_modules` exclusion; none tests for a worktree marker:

```
agent-inbox-tests.mjs          lib/mjs-files.mjs
doctor.mjs                     plugin-audit.mjs
test-all.mjs                   tests/helpers.mjs
tests/local-today-gates.test.mjs
tests/main-guard-convention.test.mjs
tests/mjs-files.test.mjs
tests/generate-pdf-page-budget.test.mjs
```

I confirmed the wrong-count behavior in `lib/mjs-files.mjs` and the hard failure in `local-today-gates`. I did not verify the other eight individually, so treat that list as "shares the pattern," not "eight confirmed bugs."

## Suggested fix

Detect the marker rather than the name — a directory containing a `.git` **file** is a linked worktree and is not this repository's source:

```js
import { statSync } from 'fs';

function isLinkedWorktree(dir) {
  try { return statSync(join(dir, '.git')).isFile(); }
  catch { return false; }
}
```

Skipping on that is narrower than blanket-ignoring `.claude/`, and it holds for a worktree placed anywhere, not just the Claude Code default path.

`lib/mjs-files.mjs` exists specifically so two walkers can't drift about what "every file" means, so it's the natural home — though `local-today-gates` and the other suites carry their own private walkers, so a complete fix either routes them through the shared helper or exports the predicate for them to use. Which of those you'd prefer is a call I'd rather leave to you than guess at in a PR.

Happy to send the patch if you name the shape.

## Environment

- career-ops v1.30.0
- One linked worktree at `.claude/worktrees/<name>`, parent checkout at v1.30.0

vliggio:
## What does this PR do?

Repo-walking helpers here skip a directory *named* `.git`. A linked worktree has no `.git` directory — it marks itself with a `.git` **file** pointing at the parent's gitdir — so the one exclusion meant to keep git's storage out of the walk slides straight past a worktree, and the walkers descend into a complete second copy of the repository at whatever commit it happens to sit on. This detects the marker instead of the name.

`lib/mjs-files.mjs` reported **1252 files in a 626-file checkout**. What made it hard to see is that the consequence differed per consumer:

- the syntax gates silently checked ~2× the files, reporting a count that depended on whether the developer had a worktree open;
- `tests/local-today-gates.test.mjs` **failed outright**, naming *correct* source as violations because the stale copies predated the convention;
- `tests/main-guard-convention.test.mjs` **passed** — only because the stale copy happened to satisfy that particular convention.

So a contributor got a red suite for a commit they didn't write, or a green one that graded the wrong tree. That is the same "gate silently covering a file set nobody chose" failure `lib/mjs-files.mjs` was written to remove (#3306, #3388, #3411, #3419).

### The fix

`isNestedCheckout()` in `lib/mjs-files.mjs`, consulted by the walk. It tests for a `.git` entry of **either type**: a file is a linked worktree or submodule, a directory below the root is a nested independent clone. All are somebody else's source tree sitting inside this one.

**Never applied to the walk ROOT.** This repo is frequently checked out *as* a worktree, whose own `.git` is a file — testing the root would return `[]`, and the syntax gate would report "0 .mjs files" and pass, having checked nothing. That is the exact #3419 failure shape and strictly worse than the bug being fixed. Pinned by its own test.

**Shape chosen:** the three private walkers import the *predicate*, not the collector. Each filters differently (`.test.mjs` exclusion, dot-dir skip, divergent `SKIP_DIRS` sets), so they can't share the walk — but a fourth hand-rolled `.git` rule is exactly the drift this module exists to prevent, so a test now pins that all three consult it.

`test-all.mjs`'s scratch-copy walker gets the same skip; it was copying an entire second checkout into the throwaway tree on every run.

### Scope

The issue listed ten walkers; **four needed the fix.** `doctor.mjs`, `agent-inbox-tests.mjs`, `tests/helpers.mjs` and `tests/generate-pdf-page-budget.test.mjs` don't recurse from the repo root. `plugin-audit.mjs` walks a user-named plugin directory, where the predicate would be actively **wrong** — a plugin installed as a git clone should still be audited. The tenth, `tests/mjs-files.test.mjs`, is the collector's own test file and carries the new cases.

### Note on #3563 (no longer part of this PR)

This originally also carried the `web/` fix for #3563. That landed independently on `main` while this PR was open, so the rebase takes `main`'s version and this branch no longer touches it. #3563 is closed; nothing here is needed for it.

## Related issue

Fixes #3499

**Follow-up, deliberately not in this PR:** #3762 — `test-all.mjs`'s `discoverTests()` walks a nested worktree placed under `tests/` and *executes* its suites. Found by @luochen211 in review here. It anchors at `ROOT/tests` rather than `ROOT`, so it falls outside this PR's set, and fixing it properly means auditing the whole subtree-anchored walker class plus reworking the by-name drift gate.

## Type of change

- [x] Bug fix
- [ ] New feature

## Verification

| | before | after |
|---|---|---|
| `collectMjsFiles`, worktree present | **2× the tree** | 653 |
| `collectMjsFiles`, no worktree | 653 | 653 |

(Measured 1252 vs 626 at the commit this was written against; `main` has grown since.)

The count is now checkout-independent, which is the invariant the module's header claims.

Rebased onto `main` at `55ef89b` (65 commits) and re-verified there. Full `node test-all.mjs` **with a nested worktree present: 8088 passed, 0 failed, 14 warnings, exit 0.** `npm run lint` reports 653 files with and without a worktree present.

One note for the reviewer: an earlier full run showed `tests/pipeline-lock.test.mjs:542` failing on a wall-clock `elapsed >= 300` assertion. It passes standalone (21/21) and passed on both full runs since — a load-sensitive flake, unrelated to this change and not addressed here.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Recursive source discovery now skips nested Git checkouts, linked worktrees, and submodules.
  * The repository root continues to be scanned correctly, including when its Git marker is a file.

* **Tests**
  * Added coverage for nested checkout detection and scanning behavior.
  * Updated validation checks to avoid treating nested repository contents as part of the current project.

### DIFF
--- a/tests/mjs-files.test.mjs
+++ b/tests/mjs-files.test.mjs
@@ -26,7 +26,7 @@ import { existsSync, mkdtempSync, mkdirSync, readFileSync, rmSync, writeFileSync
 import { tmpdir } from 'node:os';
 import { dirname, join } from 'node:path';
 import { fileURLToPath } from 'node:url';
-import { collectMjsFiles, SKIP_DIRS } from '../lib/mjs-files.mjs';
+import { collectMjsFiles, isNestedCheckout, SKIP_DIRS } from '../lib/mjs-files.mjs';
 
 const ROOT = dirname(dirname(fileURLToPath(import.meta.url)));
 
@@ -91,6 +91,86 @@ test('the syntax gate reaches past the repository root', () => {
   // 'web/ test discovery contract' check already uses instead of hardcoding it.
   if (existsSync(join(ROOT, 'web'))) {
     assert.ok(files.some((f) => f.startsWith('web/')), 'web/ must be inside the gate when present');
+  }
+});
+
+test('a nested checkout is not walked as this repository\u2019s source', () => {
+  const dir = mkdtempSync(join(tmpdir(), 'co-mjs-files-'));
+  try {
+    writeFileSync(join(dir, 'real.mjs'), '');
+
+    // A linked worktree, exactly as git writes one: a `.git` FILE holding a
+    // gitdir pointer. The `.git` entry in SKIP_DIRS matches a NAME, so it never
+    // fires here, and the walk used to descend into the whole second checkout —
+    // 1097 files reported in a 576-file repo (#3499).
+    mkdirSync(join(dir, 'wt'));
+    writeFileSync(join(dir, 'wt', '.git'), 'gitdir: /elsewhere/.git/worktrees/wt\n');
+    writeFileSync(join(dir, 'wt', 'stale.mjs'), '');
+    mkdirSync(join(dir, 'wt', 'tests'));
+    writeFileSync(join(dir, 'wt', 'tests', 'deep.mjs'), '');
+
+    // A nested independent clone marks itself with a `.git` DIRECTORY. SKIP_DIRS
+    // drops git's storage there but not the working tree beside it, so the same
+    // second-copy hazard applies.
+    mkdirSync(join(dir, 'clone', '.git'), { recursive: true });
+    writeFileSync(join(dir, 'clone', 'other.mjs'), '');
+
+    const rel = collectMjsFiles(dir).map((f) => f.slice(dir.length + 1).replace(/\\/g, '/'));
+
+    assert.deepEqual(rel, ['real.mjs'],
+      `only this checkout's source is walked, got: ${rel.join(', ')}`);
+  } finally {
+    rmSync(dir, { recursive: true, force: true });
+  }
+});
+
+test('the walk root is exempt, so running from inside a worktree still checks it', () => {
+  const dir = mkdtempSync(join(tmpdir(), 'co-mjs-files-'));
+  try {
+    // The root's own `.git` is what makes it the repository, and in a linked
+    // worktree it is a file — the same marker the predicate skips on below the
+    // root. Applying it to the root would return [] and the syntax gate would
+    // report "0 .mjs files" and pass, having checked nothing: strictly worse
+    // than the bug it fixes, and the same shape as #3419.
+    writeFileSync(join(dir, '.git'), 'gitdir: /elsewhere/.git/worktrees/self\n');
+    writeFileSync(join(dir, 'source.mjs'), '');
+    mkdirSync(join(dir, 'lib'));
+    writeFileSync(join(dir, 'lib', 'nested.mjs'), '');
+
+    const rel = collectMjsFiles(dir).map((f) => f.slice(dir.length + 1).replace(/\\/g, '/'));
+
+    assert.deepEqual(rel, ['lib/nested.mjs', 'source.mjs']);
+  } finally {
+    rmSync(dir, { recursive: true, force: true });
+  }
+});
+
+test('isNestedCheckout detects the marker, of either type, and nothing else', () => {
+  const dir = mkdtempSync(join(tmpdir(), 'co-mjs-files-'));
+  try {
+    mkdirSync(join(dir, 'worktree'));
+    writeFileSync(join(dir, 'worktree', '.git'), 'gitdir: /elsewhere\n');
+    mkdirSync(join(dir, 'clone', '.git'), { recursive: true });
+    mkdirSync(join(dir, 'plain'));
+
+    assert.equal(isNestedCheckout(join(dir, 'worktree')), true, 'a .git file is a linked worktree or submodule');
+    assert.equal(isNestedCheckout(join(dir, 'clone')), true, 'a .git directory is an independent clone');
+    assert.equal(isNestedCheckout(join(dir, 'plain')), false, 'an ordinary subdirectory is source');
+    assert.equal(isNestedCheckout(join(dir, 'does-not-exist')), false, 'a missing directory is not a checkout');
+  } finally {
+    rmSync(dir, { recursive: true, force: true });
+  }
+});
+
+test('the private repo walkers consult the shared predicate, not their own rule', () => {
+  // lib/mjs-files.mjs exists so two walkers cannot drift about what "every
+  // file" means; the same reasoning applies to what "not our source" means.
+  // These three suites keep private walkers because each filters differently
+  // (.test.mjs, dot-dirs, SKIP_DIRS sets), so they import the predicate rather
+  // than the collector — but a fourth hand-rolled `.git` rule is the drift.
+  for (const caller of ['tests/local-today-gates.test.mjs', 'tests/main-guard-convention.test.mjs', 'test-all.mjs']) {
+    const src = readFileSync(join(ROOT, caller), 'utf-8');
+    assert.match(src, /isNestedCheckout\(/, `${caller} must skip nested checkouts via lib/mjs-files.mjs (#3499)`);
   }
 });
 
--- a/tests/mjs-files.test.mjs
+++ b/tests/mjs-files.test.mjs
@@ -170,7 +170,20 @@ test('the private repo walkers consult the shared predicate, not their own rule'
   // than the collector — but a fourth hand-rolled `.git` rule is the drift.
   for (const caller of ['tests/local-today-gates.test.mjs', 'tests/main-guard-convention.test.mjs', 'test-all.mjs']) {
     const src = readFileSync(join(ROOT, caller), 'utf-8');
-    assert.match(src, /isNestedCheckout\(/, `${caller} must skip nested checkouts via lib/mjs-files.mjs (#3499)`);
+
+    // The IMPORT is the assertion, not the call. Matching `isNestedCheckout(`
+    // anywhere in the file is satisfied by a local `const isNestedCheckout =
+    // () => false` — a hand-rolled re-implementation wearing the shared name,
+    // which is precisely the drift this test exists to catch, passing as proof
+    // against itself. Pinning the import binds the name to the one definition.
+    assert.match(
+      src,
+      /import\s*\{[^}]*\bisNestedCheckout\b[^}]*\}\s*from\s*'\.{1,2}\/lib\/mjs-files\.mjs'/,
+      `${caller} must import isNestedCheckout FROM lib/mjs-files.mjs, not re-implement it (#3499)`,
+    );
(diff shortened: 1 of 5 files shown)


Scott-Emberson (commented):
Reviewed at head 10972fe6a6. The marker-based skip is scoped correctly, so no gate is neutered. isNestedCheckout(dir) checks for a .git marker and the skip is applied only to child directories in each walk, never to the root, which stays exempt, so collectMjsFiles run from inside a worktree still returns the branch's own source (guarding the "0 files, passes" regression). main-guard-convention and local-today-gates each add the skip on the child path only, with their assertion logic untouched, so they still redden on a real guard violation or rule break in present source. The new mjs-files cases prove both a nested-worktree .git file and a nested-clone .git directory are skipped while real.mjs is still walked, that the root .git marker is exempt, that isNestedCheckout is true for both marker types and false for a plain or missing dir, and there is a drift gate asserting the three private walkers import the shared predicate. The web/ assertion is made conditional on existsSync (#3563) without becoming an unconditional skip when web/ is present.

Owned tests verified sound, only nested checkouts are dropped and no real source is silently skipped. Holding formal approval until the branch is finalized, nothing blocking from the tests/ side.

tests/mjs-files.test.mjs
```diff
@@ -82,8 +82,98 @@ test('the syntax gate reaches past the repository root', () => {
...
+
+test('the private repo walkers consult the shared predicate, not their own rule', () => {
+  // lib/mjs-files.mjs exists so two walkers cannot drift about what "every
+  // file" means; the same reasoning applies to what "not our source" means.
+  // These three suites keep private walkers because each filters differently
+  // (.test.mjs, dot-dirs, SKIP_DIRS sets), so they import the predicate rather
+  // than the collector — but a fourth hand-rolled `.git` rule is the drift.
+  for (const caller of ['tests/local-today-gates.test.mjs', 'tests/main-guard-convention.test.mjs', 'test-all.mjs']) {
+    const src = readFileSync(join(ROOT, caller), 'utf-8');
+    assert.match(src, /isNestedCheckout\(/, `${caller} must skip nested checkouts via lib/mjs-files.mjs (#3499)`);
```
nikitacometa: A local `const isNestedCheckout = () => false` satisfies this. Match the import from `lib/mjs-files.mjs` instead of the call.

nikitacometa (approved):
Verified at 10972fe6. Enumerated every recursive walker on main that starts at ROOT: `lib/mjs-files.mjs`, `tests/local-today-gates.test.mjs`, `tests/main-guard-convention.test.mjs`, and `test-all.mjs`'s scratch copy. The rest of the issue's list starts elsewhere (`mojibake-canary` walks `templates/` and `modes/`, `seed-fixture` walks `test-fixtures/`, `helpers.walkFiles` takes a dir), so four is the full set. Root exemption is pinned, the symlink skip stays ahead of the marker check, and all three suites pass here.

One nit: the drift test at tests/mjs-files.test.mjs:175 matches the text `isNestedCheckout(` anywhere in the file, so a local `const isNestedCheckout = () => false` keeps it green. Asserting the import line (`from '../lib/mjs-files.mjs'` / `./lib/mjs-files.mjs`) closes that.

tests/mjs-files.test.mjs
vliggio: Valid, and worse than a nit — thank you. The test is named "the private walkers consult the shared predicate, **not their own rule**," and a local `const isNestedCheckout = () => false` *is* that rule. The assertion was passing as proof against itself.

Fixed in 3567e36. The import is now the assertion:

```js
assert.match(
  src,
  /import\s*\{[^}]*\bisNestedCheckout\b[^}]*\}\s*from\s*'\.{1,2}\/lib\/mjs-files\.mjs'/,
  `${caller} must import isNestedCheckout FROM lib/mjs-files.mjs, not re-implement it (#3499)`,
);
assert.match(src, /isNestedCheckout\(/, `${caller} must actually call isNestedCheckout (#3499)`);
```

Kept the call check as a second assertion — the import alone is satisfied by an unused import, which would leave the walk descending into every nested checkout exactly as before.

Mutation-checked rather than assumed. Swapping the import for a local shadow in `main-guard-convention`:

```
mutated: shared import -> local shadow
✖ the private repo walkers consult the shared predicate, not their own rule
ℹ pass 8 / ℹ fail 1
```

Green before the change, red after. Thanks also for independently enumerating the ROOT-anchored walkers — that matches what I found, including `mojibake-canary` and `seed-fixture` starting elsewhere.

---

Unrelated correction for @santifer and @Scott-Emberson, since you all reviewed at `10972fe` (pre-rebase): **this PR no longer closes #3563.** That was fixed independently by #3605, which is one of the merges that caused the conflict. The rebase takes main's version of the `web/` conditional and this branch doesn't touch that block any more — I've dropped `Fixes #3563` from the description. #3499 is unaffected; everything else in the reviews still stands.

Current head is `3567e36`. Rebased onto main at `55ef89b`; full `node test-all.mjs` with a nested worktree present: 8088 passed, 0 failed. `npm run lint` reports 653 files with and without a worktree, which is the checkout-independence this is all for.

Scott-Emberson (commented):
Re-reviewed at head 3567e36ab1 (moved past my earlier comment on a large repo-move merge). On a blob compare only tests/mjs-files.test.mjs genuinely changed (main-guard-convention and local-today-gates are byte-identical). The change strengthens the drift gate: it now asserts each walker actually imports isNestedCheckout from ./lib/mjs-files.mjs, not merely that it calls a same-named symbol, so a hand-rolled `const isNestedCheckout = () => false` in one of the walkers now reddens the gate, while the existing call assertion is retained. The callers all still import and call the shared predicate, and the earlier-verified skip semantics (marker-based, root never masked via the src !== ROOT guard, no silent skip of real source) are intact.

Owned test verified sound, the delta hardens the shared-import contract. This is in our area, ready to approve on your go-ahead.

Scott-Emberson (approved):
Approving from the tests/ side. The owned test is verified sound at this head (evidence in my review note above), CI is green, and the branch is mergeable and up to date. Good to go from our end. As always this routes the code-owner review, it does not merge.

luochen211 (changes requested):
One recursive gate is still unprotected: `test-all.mjs`'s `discoverTests(dir)` descends into every directory under `tests/` without calling `isNestedCheckout`.

Repro on the current-main synthetic merge: add a linked worktree at `tests/nested-wt`, then run `node test-all.mjs --only nested-wt/tests/mjs-files`. The runner discovers and executes `tests/nested-wt/tests/mjs-files.test.mjs` (5 tests), so a second checkout is still graded when it sits under the discovery root. This contradicts the issue's any-path guarantee.

Please guard directories in `discoverTests` too and extend the convention/regression test so this fourth recursive walker cannot be omitted again.

PR #3688, opened 2026-09-02
fix(updater): scope commits to shipped files

### LINKED ISSUE #3504: update-system: apply() commits directory pathspecs, so an unstaged edit to a tracked system file is swept into the update commit
### Code of Conduct

### Existing issues

- [X] I searched existing issues and this hasn't been reported yet

### What happened?

Split out of #2531 at @luochen211's request ([comment](https://github.com/santifer/career-ops/pull/2531#issuecomment-5299757731)): *"the two pre-existing concerns called out above (`git commit -- <dir>` capturing unstaged/index-stranded files and the directory-vs-file status mismatch) should either be fixed in this PR or explicitly split into a follow-up with an issue link."*

#2531 closes the **staging** half of this seam. It does not close the **commit** half, and the two are now visibly asymmetric in `apply()`:

```js
// update-system.mjs
2290:      addPaths(expandToShippedFiles(pathsToStage));   // staging: expanded, files only
2326:      git('commit', ..., '--', ...pathsToStage);      // commit:  unexpanded, directories included
```

`pathsToStage` is the manifest, and the manifest carries directory entries (`docs/`, `tests/`, `templates/`, `.github/`, `plugins/`, `modes/*/` — 53 of them as counted in #2531). So `apply()` stages a file list and then commits a pathspec list that still contains directories.

`git commit -- <dir>` takes the **working-tree** content of matching tracked files, bypassing the index. So an unstaged local edit to a tracked system file under one of those directories is committed by the updater even though nothing staged it.

This is narrower than the `git add -f` hazard #2531 fixed — a pathspec commit will not add untracked or ignored content, so it cannot sweep user data into the commit. The exposure is limited to tracked files the updater did not itself write.

### Steps to reproduce

In a scratch repo: one tracked file with an unstaged edit, one untracked file, one ignored file, all under the same directory pathspec.

```console
$ git init -q -b main . && mkdir docs && echo v1 > docs/README.md
$ printf 'docs/secret.env\n' > .gitignore && git add -A && git commit -qm base

$ echo v2     > docs/README.md      # TRACKED, edited, never staged
$ echo untrkd > docs/new.txt        # UNTRACKED
$ echo SECRET > docs/secret.env     # IGNORED

$ git status --porcelain
 M docs/README.md
?? docs/new.txt

$ git commit -qm update -- docs/
$ git show --name-only --format= HEAD
docs/README.md
```

`docs/README.md` is committed without ever having been staged. The untracked and ignored files are not — which is the limit of the hazard, and worth stating so this is not read as a repeat of the `git add -f` problem #2531 fixed.

### Expected behavior

The commit should be scoped to the same expanded, file-only list that staging uses, so `apply()` commits exactly what it staged and nothing else.

### Additional context

Not a one-line change, which is why it is split rather than folded into #2531:

`pathsToStage` also carries `:(exclude)` pathspecs (`EXCLUDE_PATHSPEC_PREFIX`, defined at `update-system.mjs:1469`, filtered at 2317 for the `ownedPaths` check). `expandToShippedFiles` has no handling for a negative pathspec — it would either take it verbatim, which git would then interpret against the expanded list in a way nobody has specified, or drop it, which silently widens the commit. Deciding what expansion means for an exclude, and testing it, is the actual work here.

`rollback()` at 2468 commits the same way (`'--', ...rollbackPaths`), so whatever shape the fix takes should cover both call sites — that symmetry is the bug #2531 already had to fix once on the staging side.

The second concern @luochen211 raised (directory-vs-file status mismatch) is not covered here; happy to open it separately if it is still live after this one.

luochen211:
## Summary

- scope apply() and rollback() pathspec commits to the same target/backup-tree file list used for staging
- preserve `:(exclude)` pathspecs for locally preserved files
- add regression coverage for retired tracked files, preserved exclusions, and rollback files introduced after the backup

## Validation

- `node tests/updater-add-paths.test.mjs`
- `node updater-migration-tests.mjs` (372 passed)
- `node upgrade-tests.mjs --pr-gate` (GREEN)
- `env -u GEMINI_API_KEY node test-all.mjs --quick` (7761 passed, 0 failed, 13 warnings)

#3622 remains a separate staging-side change; this PR covers the commit-side seam and is based on the current canonical `main`.

Closes #3504


Scott-Emberson (commented):
Reviewed at head fbe3dec73c. On the owned test the data-safety property holds: a user file (docs/RETIRED.md) present in the worktree but retired from the target tree is proven not in the updater's commit and left dirty and unstaged, the preserve path (a :(exclude) entry) survives expansion and is likewise not swept, and the rollback path mirrors the same guarantee for a later local edit. The tests drive the real expandToShippedFiles export against a throwaway repo, so this is behavioral, not pattern-matched, and each has a real fail() branch that reddens if a user or retired file lands in the commit.

One honest scope note: the owned tests construct the commit pathspec themselves, so they verify the mechanism (expansion omits the file, committing the expanded list leaves user edits dirty) rather than calling apply()/rollback() end to end. The wiring that apply() actually uses the expanded list is enforced by the sibling migration test in the same PR, so the two together redden on a scoping regression.

The code change itself is in update-system.mjs, which is a red-line file, so I am reviewing it for signal only, the decision on the updater change defers to santifer. From the tests/ side the owned test is sound and the no-user-file-swept guarantee is genuinely pinned. Holding formal approval until the branch is finalized, nothing blocking from our area.

update-system.mjs
```diff
@@ -2316,20 +2316,20 @@ async function apply() {
     // recovery command. Declared outside the try because the catch reads it.
     let usedIndexCommit = false;
 
+    // The staging and scoped-commit paths must use the same file-only list.
+    // Passing a manifest directory to `git commit -- <dir>` reads matching
+    // tracked files from the working tree, including files the target tree no
+    // longer ships. That can sweep a user's unstaged edit into the updater
+    // commit even though staging never touched it (#3504). Exclusion
+    // pathspecs remain in the expanded list so preserved files stay out.
+    const expandedPathsToStage = expandToShippedFiles(pathsToStage);
```
nikitacometa, line 2325: `expandToShippedFiles` passes `:(exclude)…` through and `addPaths` runs with `--literal-pathspecs`, so this list cannot be staged whenever `preservedPaths` is non-empty. Probe in the review body.

update-system.mjs
```diff
@@ -2365,22 +2365,24 @@ async function apply() {
       if (usedIndexCommit) {
         git('commit', '-m', `chore: auto-update system files to v${remote}`);
       } else {
-        git('commit', '-m', `chore: auto-update system files to v${remote}`, '--', ...pathsToStage);
+        git('commit', '-m', `chore: auto-update system files to v${remote}`, '--', ...expandedPathsToStage);
```
nikitacometa, line 2368: Single unbatchable call: 29,845 chars at this manifest, Windows cap 32,767. `--pathspec-from-file` keeps it out of argv.

update-system.mjs
```diff
@@ -2503,11 +2505,15 @@ function rollback() {
...
+    // A directory pathspec would otherwise include tracked files still present
+    // in the worktree but absent from the backup tree (#3504).
+    const expandedRollbackPaths = expandToShippedFiles(rollbackPaths, latest);
     try {
       // Scope the commit to the rollback paths (#915 bug 2). A bare
       // `git commit` would sweep unrelated staged files into the rollback.
-      if (rollbackPaths.length > 0) {
-        git('commit', '-m', `chore: rollback system files from ${latest}`, '--', ...rollbackPaths);
+      if (expandedRollbackPaths.length > 0) {
+        git('commit', '-m', `chore: rollback system files from ${latest}`, '--', ...expandedRollbackPaths);
```
nikitacometa, line 2516: Always the scoped form, same argv size. Rollback also swallows the failure and prints "Rollback complete".

tests/updater-add-paths.test.mjs
```diff
@@ -307,6 +307,84 @@ console.log('\n🧪 Testing updater staging behavior (ignored + never-tracked pa
...
+  writeFileSync(join(dir, 'docs/README.md'), 'v1');
+  writeFileSync(join(dir, 'docs/KEEP.md'), 'v1');
+  g('add', '-A');
+  g('commit', '-qm', 'base');
+
+  writeFileSync(join(dir, 'docs/README.md'), 'updated');
+  writeFileSync(join(dir, 'docs/KEEP.md'), 'the preserved local edit');
+  const paths = ['docs/', ':(exclude)docs/KEEP.md'];
+  const expanded = expandToShippedFiles(paths, 'HEAD', ctx);
+  g('add', '--', 'docs/README.md');
```
nikitacometa, line 368: Staging here bypasses `addPaths`. Run `addPaths(expanded, ctx)` and this case goes red with `pathspec ':(exclude)docs/KEEP.md' did not match any files`.

nikitacometa (changes requested):
Ran the branch's tests (green), then measured the two seams the PR touches. Two blockers, one of them pre-existing but claimed as handled here, plus two notes.

1. Windows command line. `addPaths` batches at 8,000 bytes because CreateProcess caps a command line at 32,767 chars. The scoped commits are single calls that cannot batch. At this branch's manifest: 325 entries, 56 directories, 1,047 files after expansion; `git commit -m "chore: auto-update system files to v1.31.0" -- <files>` is 29,845 chars. Headroom is 2,922 chars, about 103 more shipped files. `rollback()` always takes the scoped form; `apply()` takes it whenever anything unrelated is staged. The recovery message then prints the same 30 KB pathspec as a command to paste, and cmd.exe stops at 8,191. Fix: `git commit --pathspec-from-file=<tmp> --pathspec-file-nul` (git 2.25+) for both commits and for the printed recovery; `addPaths` can drop its batching the same way.

2. `:(exclude)` never survives staging. `addPaths` runs `git --literal-pathspecs add -f -- ...`, and `--literal-pathspecs` turns pathspec magic off, so `:(exclude)docs/KEEP.md` is a literal filename. Probe at this head, through `addPaths` itself: `expandToShippedFiles(['docs/', ':(exclude)docs/KEEP.md'], 'HEAD')` returns `["docs/KEEP.md","docs/README.md",":(exclude)docs/KEEP.md"]`, and `addPaths` throws `fatal: pathspec ':(exclude)docs/KEEP.md' did not match any files`. The expansion also lists `docs/KEEP.md` itself, so the preserved file would be force-added even without the throw. Same result on main, so `apply` without `--force` on any locally modified system file has been aborting at staging before either commit. Test 6c passes because it stages with a raw `g('add', '--', 'docs/README.md')` instead of `addPaths(expanded)`. Fix: resolve preservation in JS (expand, subtract `preservedPaths`, hand literal filenames to add and commit) and route 6c through `addPaths`.

Notes: the commit calls run without `--literal-pathspecs`, so a shipped name with `[`, `*` or `?` would be read as a glob at commit time (none in the manifest today). And the new tests call `expandToShippedFiles` themselves, so `expandToShippedFiles(pathsToStage, 'HEAD')` in `apply()` leaves them green; the migration regex only sees the variable name.

PR #3689, opened 2026-09-02
fix(web): read set-status.mjs's pretty --json document, not one line of it

dankadolba:
This was opened by mistake while I was using my AI coding tool.

It was not meant to be submitted to this repository, so please disregard it. Sorry for the noise and the extra work.

### DIFF
--- a/web/tests/lib/status-cli.test.mjs
+++ b/web/tests/lib/status-cli.test.mjs
@@ -84,3 +84,54 @@ test("a non-string error field is not passed through as one", () => {
   assert.equal(clientErrorMessage({ error: { nested: true } }), "status update failed");
   assert.equal(clientErrorMessage({}), "status update failed");
 });
+
+// ── the shape set-status.mjs actually prints ─────────────────────────────────
+// Every test above feeds a COMPACT object, and the CLI prints none. Its --json
+// path ends in `JSON.stringify(result, null, 2)`, so the document is pretty and
+// its first line is a bare `{`. Reading one line at a time therefore matched
+// nothing on the success path: /api/status answered 500 "status update returned
+// no result" for a tracker row it had just written, and the suite stayed green
+// because no case here had ever been pretty. So these build their document the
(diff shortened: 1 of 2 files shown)


web/tests/lib/status-cli.test.mjs
```diff
@@ -84,3 +84,54 @@ test("a non-string error field is not passed through as one", () => {
...
+  tracker: "data/applications.md",
+};
+
+test("a pretty-printed document is read, because that is the only shape the CLI prints", () => {
+  const stdout = `${JSON.stringify(SET_STATUS_RESULT, null, 2)}
+`;
+  assert.deepEqual(parseCliJson(stdout), SET_STATUS_RESULT);
+});
+
+test("a diagnostic ahead of a pretty document still does not shadow it", () => {
```
nikitacometa, line 115: This warning is printed with `console.error` in set-status.mjs, so it never shares stdout with the document. A generic preamble in the fixture would stop the next reader from taking it for the ledger path.

nikitacometa (approved):
Verified at 455f0908. The four new cases go 9 pass / 4 fail against main's parser and 13/13 here. set-status.mjs:582 is the only success-path print and it is `JSON.stringify(result, null, 2)`, so the one-line reader could never match it. The slice scan is sound by construction: any `{`-opening line that is not the root is an array element, its slice ends in `]`/`}` and fails to parse, so the scan walks on to the outer `{`. CRLF is fine too, `\r` is JSON whitespace.

Two nits on the description, not the code:

- The retry consequence is overstated. set-status.mjs appends the ledger row only `if (statusChanged && !flags.dryRun)`, so a retry of `Evaluated → SKIP` is a `SKIP → SKIP` no-op and writes nothing. The cost of the 500 is the misleading UI, not duplicate transitions.
- The "diagnostic ahead of the document" fixture is synthetic: the ledger warning goes out via `console.error` (set-status.mjs:557) and the route reads stdout alone. It still pins the #2901 property, just not against a line the CLI can print on stdout.

PR #3711, opened 2026-09-02
fix(doctor): detect project-scoped Playwright MCP plugin installs (#3698)

### LINKED ISSUE #3698: doctor.mjs false negative for project-scoped Playwright MCP plugin installs
### Code of Conduct

### Existing issues

### What happened?

doctor.mjs's isPlaywrightMcpFromPlugin() (around line 280) only reads enabledPlugins from the user-level config (~/.claude/settings.json):

js
const enabled = readConfigIfPresent(join(configDir, 'settings.json'))?.enabledPlugins;
if (!enabled || typeof enabled !== 'object') return false;

When /plugin install playwright@claude-plugins-official is run with project scope, the enable flag lands in the project config instead (career-ops/.claude/settings.json), not the user config. Doctor never checks there, so it reports "Playwright MCP not detected" even though the plugin is installed and the browser_* tools are live in the session.

This is misleading beyond cosmetics: the warning tells users that mandatory offer-liveness verification isn't available when it actually is, and pushes them to reinstall a plugin that's already working. Same failure class as #2752, one layer up (that one fixed where the server is declared; this one is where the plugin is enabled).

### Steps to reproduce

Suggested fix: merge enabledPlugins from all three sources Claude Code itself uses, project-first (a plain union would let a user-level enable silently override a project that deliberately disables the plugin):

js
function isPlaywrightMcpFromPlugin(root) {
  const configDir = claudeConfigDir();
  const enabled = [
    readConfigIfPresent(join(root, '.claude', 'settings.json'))?.enabledPlugins,
    readConfigIfPresent(join(root, '.claude', 'settings.local.json'))?.enabledPlugins,
    readConfigIfPresent(join(configDir, 'settings.json'))?.enabledPlugins,
  ].filter((e) => e && typeof e === 'object')
    .reduce((acc, e) => Object.assign(acc, e), {});
  if (!Object.keys(enabled).length) return false;
  // ...unchanged
}

(plus threading root through the call site at doctor.mjs:317)

Repro:

/plugin install playwright@claude-plugins-official with project scope
career-ops doctor → still shows ⚠️ Playwright MCP not detected, even though browser_* tools work fine in the session

I don't have a fork/PR set up, just reporting — happy to share more details if useful.

### Expected behavior

_No response_

### CLI tool

_No response_

### OS and Node.js version

_No response_

### Output of npm run doctor

```shell

```

OBOneProtocol:
Closes #3698

### Context & Problem
When Claude Code plugins are installed with project scope (e.g., `/plugin install playwright@claude-plugins-official`), the plugin enable flag lands in `<project>/.claude/settings.json` or `<project>/.claude/settings.local.json` rather than the global user configuration (`~/.claude/settings.json`). 

Previously, `isPlaywrightMcpFromPlugin` in `doctor.mjs` only inspected the global user-level `settings.json`, causing `doctor.mjs` to emit a false-negative `Playwright MCP tools not detected` warning on working project-scoped setups.

### Solution
1. **Multi-tier Config Merging**: Merged `enabledPlugins` across user config (`~/.claude/settings.json`), project shared config (`.claude/settings.json`), and project local config (`.claude/settings.local.json`), preserving Claude Code's project-over-user precedence.
2. **Callsite Threading**: Threaded `root` through `isPlaywrightMcpConfigured` into `isPlaywrightMcpFromPlugin(root)`.
3. **Automated Test Coverage**: Added 4 new test cases to `tests/playwright-mcp-detection.test.mjs` (covering project-scoped enabled plugins, project-local enabled plugins, project-level disables overriding user-level enables, and project-local overrides). Full test suite (including `test-all.mjs`) passes cleanly.

## Summary

`doctor.mjs` now detects Playwright MCP plugins enabled at project scope.

Users can enable the plugin in project or project-local Claude settings without receiving a false “Playwright MCP not detected” warning. Project settings override user settings. Project-local settings override shared project settings.

### Changes

- `doctor.mjs:280`: Merges `enabledPlugins` from user, project, and project-local Claude settings.
- `doctor.mjs:317`: Passes the project root to plugin detection.
- `tests/playwright-mcp-detection.test.mjs`: Adds four regression tests for project enables and override behavior.

### Files touched

- `doctor.mjs`
- `tests/playwright-mcp-detection.test.mjs`

The following system files or directories are not touched: `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, `.github/`.


Scott-Emberson (commented):
Reviewed at head bc0e53f8cb. The owned test tests/playwright-mcp-detection.test.mjs pins the project-scoped detection in both directions: a project .claude/settings.json (and separately settings.local.json) that enables the plugin with the user config empty is detected (playwright_mcp.claude true, no warning), and a project config that disables it after the user enabled it comes back false with the warning present, which guards the override direction and against over-detection. The local-over-shared precedence case is asserted too. Fixtures use temp plugin homes and per-scenario config-dir pinning with cleanup, so it is hermetic. Concrete true/false on the resolved flag plus warning presence discriminates detect from absent, so a detection regression or an over-detect both redden.

The substance of the fix is in doctor.mjs, which is outside our owned paths, so this is a test-axis review: the owned test is sound, but the scope and merge of the doctor change are the maintainer's, not something my approval gates. Nothing blocking from the tests/ side.

doctor.mjs
```diff
@@ -274,14 +274,23 @@ function claudeConfigDir() {
...
+// walking plugins/cache, so a large cache costs nothing and disabled plugins
+// are never read.
+function isPlaywrightMcpFromPlugin(root) {
   const configDir = claudeConfigDir();
 
-  const enabled = readConfigIfPresent(join(configDir, 'settings.json'))?.enabledPlugins;
-  if (!enabled || typeof enabled !== 'object') return false;
+  const enabled = [
+    readConfigIfPresent(join(configDir, 'settings.json'))?.enabledPlugins,
+    root ? readConfigIfPresent(join(root, '.claude', 'settings.json'))?.enabledPlugins : null,
```
nikitacometa, line 289: `root` is always truthy here; the only callsite (326) passes a resolved project root. The ternaries can go.

doctor.mjs
```diff
@@ -314,7 +323,7 @@ function isPlaywrightMcpConfigured(root, activeCli) {
   if (inProject) return true;
   // Gated behind the project scan, so an already-configured project pays no
```
nikitacometa, line 324: `.claude/settings.json` and `.claude/settings.local.json` are now parsed twice on the fallback path (319-322 for `mcpServers`, 289-290 for `enabledPlugins`). Reword this comment or hoist the parsed objects.

nikitacometa (approved):
Verified at bc0e53f8. Merge order `[user, project, project-local]` under `reduce(Object.assign)` gives local over project over user, matching Claude Code's settings chain; the issue's suggested snippet had that order reversed, and this fixes it. Detection still requires `installed_plugins.json` plus the plugin's own `.mcp.json`, so a project file alone cannot turn it on, and `--json` still exposes only a boolean.

Tests have teeth: reverting doctor.mjs to main and keeping the tests gives 26 passed, 4 failed, with `playwright_mcp.claude` flipping in both directions. `--only doctor` 21/0.

Separate issue worth filing: a user-scoped server install is still a false negative. With `<configDir>/mcp.json` = `{"mcpServers":{"playwright":{"args":["@playwright/mcp@latest"]}}}` and an empty project, this head returns `{"playwright_mcp":{"claude":false}}`; doctor.mjs:229 lists project-relative files only, and the warning at 388 tells the user to add the server to their claude config, which they did.

Scott-Emberson (commented):
Test-axis only. The substance is in `doctor.mjs`, which is the maintainer's routing call; I am reviewing the owned test `tests/playwright-mcp-detection.test.mjs`.

It is sound. It runs the real `doctor.mjs --json` subprocess and parses actual output (no mock), and each case gates `pass()` behind a concrete condition, failing otherwise. The positive cases (project `.claude/settings.json` and `.claude/settings.local.json`) pin `playwright_mcp.claude === true` and the absence of a `/playwright mcp/i` warning, with the user `settings.json` cleared to `{}` so detection depends solely on the project-scoped file the fix added, so on pre-fix code they read false with the warning present and genuinely redden. The absence and override cases pin `claude === false` and the warning present, exercising the last-wins merge order, so a regression that dropped the override flips them to fail. Detection-present and detection-absent are both asserted with correlated warning checks.

PR #3713, opened 2026-09-02
fix(rejection-latency): confidential-employer applications vanished from the signal (#3712)

### LINKED ISSUE #3712: rejection-latency.mjs: confidential-employer applications are silently dropped from the signal
`rejection-latency.mjs` flags applications still in `Interview` whose silence exceeds a courtesy threshold. Any row whose employer is the `?` placeholder never reaches that check.

```js
rejection-latency.mjs:169  export function companyKey(name) {
                             return String(name || '').normalize('NFKC').toLowerCase()
                                      .replace(/[^\p{L}\p{N}]/gu, '');
                           }
rejection-latency.mjs:203    const key = companyKey(row.company);
rejection-latency.mjs:204    if (!key) continue;        // <- every `?` row leaves here, silently
```

`?` contains no letter and no digit, so it normalizes to `""`. The `—` and `-` no-data sentinels (#1799) go the same way.

Measured against the real parser — three `Interview` rows, all confidential:

```
companyKey("?")             = ""
groups found                = [ 'acme' ]
confidential rows retained  = 0 of 3
```

### Why this blind spot in particular

This check exists to notice applications that have gone quiet. Agency-brokered roles — precisely the ones carrying `?` plus a `via=` field — are where the candidate has the least visibility into what is happening and where ghosting is most common. A report that omits them without saying so reads as *"nothing is overdue"*, not *"I did not look at these"*.

### The helper already exists twice

Two sibling readers of the same two files recognise placeholders explicitly:

| File | Helper | Behaviour |
|---|---|---|
| `reply-matcher.mjs:31` | `isPlaceholderCompany()` | recognises `?`/`—`/`-` (added by #3001, where `?` matched almost every reply) |
| `process-quality.mjs:160` | `isPlaceholder()` | identical body, identical rationale |
| `rejection-latency.mjs` | — | **silently drops instead** |

Same body, written independently, and the third reader has neither. That is the shape of the bug as much as the `if (!key)` is.

### What a fix should not do

Joining `?` rows to `active-interviews.md` by company is not available: every `?` row says `?` on both sides, so they would all match each other. A confident wrong answer would be worse than the current silence. Naming the gap is the honest result until the rows carry something identifiable (the `via=` agency, or a req ID in notes).

PR to follow.

abankar1:
Fixes #3712.

`companyKey()` strips everything that is not a letter or a digit, so `?` — the documented marker for an undisclosed end employer (#1596) — normalizes to the empty string and `parseTrackerInterviewRows`' `if (!key) continue` discarded the row. The `—`/`-` no-data sentinels went the same way.

```
companyKey("?")             = ""
groups found                = [ 'acme' ]
confidential rows retained  = 0 of 3
```

This is the wrong blind spot for this check in particular. It exists to flag applications that have gone quiet, and agency-brokered roles — the ones carrying `?` plus a `via=` field — are where the candidate has the least visibility and where ghosting is most common. A report that omits them silently reads as *"nothing is overdue"*, not *"I did not look at these"*.

## Reported, not guessed

Every `?` row says `?` on both sides, so joining them to `active-interviews.md` by company would match each against all the others — a confident wrong answer in place of a stated gap. So the rows are counted into `metadata.placeholderApplicationsExcluded` and named in a warning that says what would make them identifiable:

```
3 Interview-state applications (#2, #3, #4) excluded: the employer cell is a
placeholder, so there is nothing to group or attribute by. Put the agency in a
via= field or the req ID in notes to make them identifiable.
```

When the tracker holds no placeholder rows, nothing changes.

## Two commits, because the duplication is half the bug

**1 — one definition.** `reply-matcher.mjs` and `process-quality.mjs` had grown the same helper independently, same body, same rationale, while this file — the third reader of the same two files — had neither. `lib/placeholder-cell.mjs` now holds it and all three import it.

In `lib/` rather than `tracker-parse.mjs` deliberately: `reply-matcher.mjs` has **no imports at all** today and `tracker-parse.mjs` reads the filesystem. Pointing a pure matcher at an fs-importing module to share one regex is the wrong trade. The new file imports nothing, like `lib/local-today.mjs` and `lib/ascii-fold.mjs`, and is registered in `SYSTEM_PATHS` — without which `apply` ships the importers and not the module.

`process-quality` keeps its local `isPlaceholder` name as an alias so its call sites are untouched, and `reply-matcher` keeps the comment above its call site, which records the incident that bought the rule.

**2 — the fix.** `placeholderInterviewRows` is a separate export rather than folded into `parseTrackerInterviewRows`' return, so that Map contract is untouched: two internal callers and `tests/local-today-gates.test.mjs` consume it as a plain Map, and widening it to an object would have meant editing a test in `/tests/` for no gain here.

## Tests

Five, including one end to end — the whole value of this fix is the user being told, so a count that never reaches the report would be the same silence. Two guards that must stay green: named companies still group (the fix must not start excluding real employers), and only `Interview`-state rows count (a placeholder in another state is not something this check would have looked at, so counting it would overstate the gap).

Reverting the check reddens three of five; both guards stay green. Full suite green at 8038.

## Summary

`rejection-latency.mjs:174` now handles `Interview` applications with employer placeholders: `?`, `—`, and `-`.

Users can:

- See placeholder applications grouped by `via=` agency.
- See warnings and `metadata.placeholderApplicationsExcluded` for applications without attribution.
- See a partial-assessment message instead of a false all-clear.
- Continue using named-employer matching without behavior changes.

Placeholder applications are not joined by employer. This prevents incorrect matches between confidential applications.

`lib/placeholder-cell.mjs:44` provides the shared predicate. `process-quality.mjs:161` and `reply-matcher.mjs:70` use it.

Tests cover agency grouping, placeholder rows, named employers, non-`Interview` rows, malformed input, and CLI reporting.

## System files touched

- `update-system.mjs:235`: Adds `lib/placeholder-cell.mjs` to `SYSTEM_PATHS`.
- `AGENTS.md`, `modes/`, `DATA_CONTRACT.md`, `providers/`, `.github/`: Not touched.

## Breaking changes

No breaking changes are expected for named employers or existing `Interview` matching. Placeholder applications now produce explicit warnings and changed summary output instead of being silently excluded.


Scott-Emberson (commented):
Reviewed at head f150ef3be3. The owned test tests/rejection-latency-placeholder-rows.test.mjs pins the fix against its actual contract. The three placeholder rows (a ?, an em dash, and a hyphen company, all in Interview state) are asserted to be surfaced, with the excluded count equal to 3 and the row numbers [2,3,4], so dropping placeholder rows again reddens. An Applied-state placeholder is asserted not counted, so genuinely not-due rows stay excluded and the gap is not overstated. The end-to-end CLI asserts metadata.placeholderApplicationsExcluded equals 3 and a warning naming #2, so the count actually reaches the user rather than sitting in an internal array. A named company still groups (byCompany has acme, size 1) and empty or malformed input returns an empty array without throwing.

One thing worth naming: the fix surfaces confidential-employer rows through a reported and warned exclusion (they cannot be attributed well enough to group by), not by adding them to the latency flags array, and the test correctly asserts against that visibility contract with concrete counts and the #2 id. The substance is in rejection-latency.mjs, outside our owned paths, so this is a test-axis review: the owned test is sound, the merge is the maintainer's call. Nothing blocking from the tests/ side.

luochen211 (changes requested):
The new warning recommends a remediation the implementation does not consume. Fixture row #2 already has `via=Hays`, yet `placeholderInterviewRows()` still counts it as excluded and the CLI still tells the user to "Put the agency in a via= field ... to make them identifiable." The function looks only at status and placeholder company; neither `via=` nor a req ID affects grouping or exclusion.

Please either implement the promised identifier path end-to-end, or change the warning to an action the current code can actually use (for example, replace the placeholder employer once known). Add a regression showing that following the displayed advice changes the result; the existing fixture currently demonstrates the contradiction while still passing.

rejection-latency.mjs
```diff
@@ -587,17 +631,32 @@ if (isMainModule(import.meta.url)) {
...
   const result = computeRejectionLatency(interviewRows, trackerByCompany, {
     today, courtesyDays,
   });
 
+  if (excluded.length) {
+    const nums = excluded.map(r => `#${r.num}`).join(', ');
+    result.warnings.push(
+      `${excluded.length} Interview-state application${excluded.length === 1 ? '' : 's'} (${nums}) `
+      + 'excluded: the employer cell is a placeholder, so there is nothing to group or attribute by. '
+      + 'Put the agency in a via= field or the req ID in notes to make them identifiable.',
```
nikitacometa: Advice no code consumes: `via` and req IDs are never read here. process-quality.mjs:196-203 in this diff shows the consuming version. Probe with both remediations applied returns the identical warning.

rejection-latency.mjs
```diff
@@ -208,6 +209,49 @@ export function parseTrackerInterviewRows(content) {
...
+ * can identify them (a `via=` agency, or a req ID in notes).
+ *
+ * Separate from parseTrackerInterviewRows rather than folded into its return so
+ * that function's Map contract is untouched — two internal callers and
+ * tests/local-today-gates.test.mjs consume it as a plain Map.
+ *
+ * @param {string} content - Raw tracker markdown.
+ * @returns {object[]} The excluded rows, in file order.
+ */
+export function placeholderInterviewRows(content) {
```
nikitacometa, line 273: Counts tracker rows only; 308 keeps the same silence for `active-interviews.md`, where the flags actually come from. A `?` round there gives `placeholderApplicationsExcluded=0` and no warning.

process-quality.mjs
```diff
@@ -155,11 +156,9 @@ export function extractFriction(row) {
...
 }
 
-// A cell carrying no letter and no digit is a PLACEHOLDER, not a value: `?` for
-// an undisclosed employer (#1596), and the tracker's `—`/`-` no-data sentinels.
-function isPlaceholder(value) {
-  return !/[\p{L}\p{N}]/u.test(value);
-}
+// Local alias so the call sites below read unchanged; the definition moved to
+// lib/placeholder-cell.mjs, which was one of two identical copies.
+const isPlaceholder = isPlaceholderCompany;
```
nikitacometa, line 161: The alias hides that `undefined` now answers `true` (old bodies stringified it to "undefined", not a placeholder). Inert today since both call sites guard first; name it in the lib header.

nikitacometa (changes requested):
The counter works (probe: two `?` Interview rows give `⚠ 2 Interview-state applications (#2, #3) excluded` and `placeholderApplicationsExcluded: 2`; reverting rejection-latency.mjs to main fails 4 of the 5 new tests; `validate-system-paths-coverage` sees the new lib file). Three things.

The warning at rejection-latency.mjs:649 tells the user to add `via=` or a req ID; neither is read. `placeholderInterviewRows()` branches on status plus `isPlaceholderCompany(row.company)` only. Probe: rows `| 2 | ? | … | Interview | via=Hays |` and `| 3 | ? | … | Interview | job id JR-991 |`, both remediations applied, still return the same warning. process-quality.mjs:196-203 in this same diff already implements the consuming version (`? (via Hays)` bucket); mirror it or reword to an action the code honours.

rejection-latency.mjs:308 keeps the same silence on the side that produces the flags: a `?` round in `data/active-interviews.md` is dropped by `if (!cKey) continue` with no counter. One named tracker row plus one `?` interview round (`via=Hays, agency-brokered`) gives `placeholderApplicationsExcluded=0, companiesChecked=0, flagged=0, warnings=[]`. The counter reports tracker rows; the flags come from the other file.

And with `--summary` a tracker holding only placeholder Interview rows still prints "No post-interview silence exceeded the configured thresholds." above the warning (measured at this head), an all-clear for exactly the rows the change says cannot be assessed. Gate that line on `placeholderApplicationsExcluded === 0`.

Scott-Emberson (commented):
Test-axis only. The substance is in `rejection-latency.mjs`, `process-quality.mjs`, `reply-matcher.mjs`, and `lib/placeholder-cell.mjs`, plus a red-line `update-system.mjs` touch, which is the maintainer's routing call; I am reviewing the owned test `tests/rejection-latency-placeholder-rows.test.mjs`.

It is sound and pins the bug both directions. It pins inclusion of the confidential-employer rows: the `?`/`—`/`-` Interview rows are surfaced (`placeholderInterviewRows` returns [2,3,4]) and the end-to-end run spawns the real CLI and asserts `metadata.placeholderApplicationsExcluded === 3` with a warning naming the row, so a placeholder row reaches the report instead of vanishing. It pins that a normal named row still groups (`byCompany.size === 1`, no placeholder leakage). I mutation-checked it: forcing `placeholderInterviewRows` back to empty (the silent-drop regression) reddens five tests including the e2e, disabling the via-inclusion reddens three, and breaking the named grouping reddens seven including the normal-row test, so the normal-row guard is load-bearing rather than tautological. The tests parse real tracker markdown and exercise the CLI via spawn, not a re-implementation.

PR #3715, opened 2026-09-02
fix(analysis): funnel-velocity, calibrate and weekly-digest ignore the configured data root (#3714)

### LINKED ISSUE #3714: funnel-velocity, calibrate and weekly-digest ignore CAREER_OPS_ROOT and the .career-ops-data marker
#3511 fixed this class for the scanners. These three have a different spelling of the same bug, and the structural guard it added does not see it.

```js
funnel-velocity.mjs:51   const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
calibrate.mjs:44         const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
weekly-digest.mjs:51     const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
```

Not a bare relative literal — a constant named as though it were the project root, holding the **code** root, with user-layer paths joined onto it:

```js
funnel-velocity.mjs:812  const trackerPath = resolveTrackerPath(CAREER_OPS);
calibrate.mjs:443        const appsFile = resolveTrackerPath(CAREER_OPS);
weekly-digest.mjs:52     const DEFAULT_SESSIONS_DIR = join(CAREER_OPS, 'interview-prep', 'sessions');
```

`getCareerOpsRoot()` is the only thing that honours `CAREER_OPS_ROOT`, `CAREER_OPS_DATA_DIR` and the `.career-ops-data` marker, and none of these three calls it. `stats.mjs:28-33` is the reference implementation.

### Measured

With a real tracker and a session file in a configured data root, cwd elsewhere:

```
CAREER_OPS_ROOT=<root> funnel-velocity --summary
  -> "No tracker found at <CHECKOUT>/applications.md — nothing to calibrate yet."

CAREER_OPS_ROOT=<root> calibrate --summary
  -> same

CAREER_OPS_ROOT=<root> weekly-digest --summary --from 2026-08-24 --to 2026-08-30
  -> "No interviews recorded in this range (no session files fall inside this range)."
     --dir <root>/interview-prep/sessions  ->  "Sessions: 1 in range across 1 companies"
```

The `.career-ops-data` marker is ignored the same way; only `CAREER_OPS_TRACKER` reaches the first two, because it bypasses root resolution entirely.

### Why these are worse than "reads the wrong file"

None of them errors. `calibrate` exists to answer *"are my scores predicting my real outcomes"*, and `nothing to calibrate yet` is a legitimate answer for a new user — given to someone with a full tracker it is a wrong answer that looks like a right one.

`weekly-digest` is sharper still: `(no session files fall inside this range)` asserts that sessions exist and none match the dates. It sends the user to check their date range for a directory that was never opened.

### Note for a fix

`funnel-velocity` needs **two** roots, not one substitution. It reads `templates/states.yml` and `templates/benchmarks.yml`, which ship with the code and are not in a user's data root; `config/benchmarks.yml` is a user override by the file's own header. A blanket move to the data root would break the templates.

PR to follow.

abankar1:
Fixes #3714.

#3511 fixed this class for the scanners. These three have a different spelling of the same bug, and the structural guard it added does not see it — not a bare relative literal, but a constant named as though it were the project root, holding the **code** root, with user-layer paths joined onto it.

```js
const CAREER_OPS = dirname(fileURLToPath(import.meta.url));   // the CODE root
...
resolveTrackerPath(CAREER_OPS)                                 // user data
join(CAREER_OPS, 'interview-prep', 'sessions')                 // user data
```

`getCareerOpsRoot()` is the only thing that honours `CAREER_OPS_ROOT`, `CAREER_OPS_DATA_DIR` and the `.career-ops-data` marker, and none of the three called it.

## What the user saw

```
funnel-velocity   "No tracker found at <CHECKOUT>/applications.md — nothing to calibrate yet."
calibrate          same
weekly-digest     "No interviews recorded in this range (no session files fall inside this range)."
                   --dir <same directory>  ->  "Sessions: 1 in range across 1 companies"
```

None of them errors, which is the problem. `calibrate` exists to answer *"are my scores predicting my real outcomes"*, and `nothing to calibrate yet` is a legitimate answer for a new user — given to someone with a full tracker it is a wrong answer that looks like a right one. `weekly-digest` is sharper still: the parenthetical asserts that sessions exist and none match the dates, sending the user to check their range for a directory that was never opened.

## funnel-velocity needed two roots, not a substitution

It genuinely reads both layers, so a blanket move to `DATA_ROOT` would have broken the shipped templates:

| Path | Root | Why |
|---|---|---|
| `templates/states.yml` | CODE | ships with the code, absent from a user's data root |
| `templates/benchmarks.yml` | CODE | the shipped default |
| `config/benchmarks.yml` | DATA | the file's own header calls it a user override, and every other reader of `config/` resolves it that way |
| the tracker | DATA | |

One name for both is what let this survive — the wrong lookups sat beside correct ones and read identically. `data/outcomes/` in `calibrate` follows automatically, since `resolveWorkspaceRoot()` derives it from the tracker path.

## Tests

Six checks, each in a child with the data root and the cwd pointed at **different** directories — same reasoning as #3511's suite: if those were one directory the test could not tell the two resolutions apart, which is how the drift survived.

Two guards beyond the three fixes:

- **The other half of the split** — `templates/states.yml` and `templates/benchmarks.yml` must keep resolving from the code root, so a blanket `DATA_ROOT` move reddens it.
- **A structural check for this spelling**, in the spirit of #3511's check 6: no user-layer path may be joined onto a code-root constant. It covers five analysis scripts, not just the three being fixed, so the next one fails closed.

Reverting any one file's root reddens exactly its own check (5 pass / 1 fail, three times). Full suite green at 8370.

One thing found while building this and left alone: `browser-extract.mjs:91` and `followup-cadence.mjs:32` still resolve `config/profile.yml` from their own directory, the same way. Out of scope here — happy to file it separately.

### DIFF
--- a/calibrate.mjs
+++ b/calibrate.mjs
@@ -38,10 +38,16 @@ import { fileURLToPath } from 'url';
 import { dirname } from 'path';
 import { parseTrackerRow, resolveColumns, isSeparatorRow, isHeaderRow } from './tracker-parse.mjs';
 import { resolveTrackerPath, resolveWorkspaceRoot } from './tracker-utils.mjs';
+import { getCareerOpsRoot } from './path-resolver.mjs';
 import { canonicalOutcome } from './lib/outcome-types.mjs';
 import { outcomeDirsFor } from './lib/outcome-dir.mjs';
 
-const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
+// The USER's data root, not this file's directory. It was __dirname under the
+// name CAREER_OPS, so resolveTrackerPath() looked inside the checkout and a user
+// with CAREER_OPS_ROOT (or a .career-ops-data marker) got "No tracker found ...
+// nothing to calibrate yet" — which reads as "you have no outcome data",
+// exactly the thing this advisory is meant to answer.
+const DATA_ROOT = getCareerOpsRoot();
 
 // --- Outcome semantics ---------------------------------------------------
 //
@@ -440,7 +446,7 @@ if (isMainModule(import.meta.url)) {
     console.error('--min-band-n must be a positive integer');
     process.exit(2);
   }
-  const appsFile = resolveTrackerPath(CAREER_OPS);
+  const appsFile = resolveTrackerPath(DATA_ROOT);
   if (!existsSync(appsFile)) {
     console.error(`No tracker found at ${appsFile} — nothing to calibrate yet.`);
     process.exit(1);
--- a/tests/analysis-scripts-data-root.test.mjs
+++ b/tests/analysis-scripts-data-root.test.mjs
@@ -1,0 +1,143 @@
+// tests/analysis-scripts-data-root.test.mjs — the analysis scripts read the
+// USER's data root, not the directory they happen to live in (#3510's family).
+//
+// #3511 fixed this for the scanners and added a structural guard against a new
+// bare cwd-relative literal. These three had a different spelling of the same
+// bug and that guard does not see it: the constant was named CAREER_OPS and
+// assigned `dirname(fileURLToPath(import.meta.url))`, so it reads as the
+// project root and is in fact the CODE root. Both documented overrides —
+// CAREER_OPS_ROOT / CAREER_OPS_DATA_DIR and the .career-ops-data marker — were
+// ignored, and getCareerOpsRoot() is the only thing that honours them.
+//
+// The failures were quiet in the way that matters:
+//
+//   funnel-velocity   "No tracker found at <CHECKOUT>/applications.md"
+//   calibrate         "... nothing to calibrate yet"   <- reads as "you have no data"
+//   weekly-digest     "no session files fall inside this range"  <- blames the DATES
+//
+// Each runs in a child with the data root and the cwd pointed at DIFFERENT
+// directories. A path following the cwd or the checkout finds nothing; one
+// following the data root finds the fixture. If those were the same directory
+// this suite could not tell them apart, which is exactly how the drift survived.
+//
+// Run:  node --test tests/analysis-scripts-data-root.test.mjs
+
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { spawnSync } from 'node:child_process';
+import { mkdtempSync, mkdirSync, writeFileSync, rmSync, readFileSync, readdirSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const ROOT = dirname(dirname(fileURLToPath(import.meta.url)));
+
+function fixture() {
+  const dataRoot = mkdtempSync(join(tmpdir(), 'career-ops-analysisroot-'));
+  const decoyCwd = mkdtempSync(join(tmpdir(), 'career-ops-analysiscwd-'));
+  mkdirSync(join(dataRoot, 'data'), { recursive: true });
+  mkdirSync(join(dataRoot, 'interview-prep', 'sessions'), { recursive: true });
+  writeFileSync(join(dataRoot, 'data', 'applications.md'), [
+    '# Applications Tracker',
+    '',
+    '| # | Date | Company | Role | Score | Status | PDF | Report | Notes |',
+    '|---|---|---|---|---|---|---|---|---|',
+    '| 1 | 2026-01-05 | Acme | Backend Engineer | 4.2/5 | Applied | ✅ | [1](r1.md) | n |',
+    '| 2 | 2026-02-05 | Globex | ML Engineer | 4.4/5 | Interview | ✅ | [2](r2.md) | n |',
+    '',
+  ].join('\n'));
+  writeFileSync(join(dataRoot, 'interview-prep', 'sessions', 'acme-backend-r1.md'), [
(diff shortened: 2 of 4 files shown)


Scott-Emberson (commented):
Reviewed at head ccf1f004ee. The owned test tests/analysis-scripts-data-root.test.mjs covers all three scripts and the discriminator is genuine. Each script is spawned from the repo root with cwd set to a decoy temp dir but CAREER_OPS_ROOT pointed at a separate temp data root holding the fixture (and CAREER_OPS_DATA_DIR empty, which the resolver ignores), so resolution has to follow the configured data root rather than cwd or the checkout. Positive per script: funnel-velocity prints its Funnel Calibration report and not "No tracker found", calibrate likewise avoids "No tracker found", and weekly-digest reports one session in range. Negative: the decoy cwd is asserted to stay empty for all three. The reddening is real, not masked by a shipped tracker, because the checkout has no data/applications.md and the sessions dir holds only placeholders, so the buggy __dirname-rooted resolution would find nothing and flip each assertion. One minor note, not a defect: calibrate's runtime check is absence-only rather than asserting a specific line, but that message is the bug's signature so it still reddens.

The substance is in funnel-velocity.mjs / calibrate.mjs / weekly-digest.mjs, outside our owned paths, so this is a test-axis review: the owned test is sound and covers all three, the merge is the maintainer's call. Nothing blocking from the tests/ side.

luochen211 (approved):
Reviewed the configured data-root changes across funnel velocity, calibration, and weekly digest. The current head merges cleanly with current main, `git diff --check` is clean, and `tests/analysis-scripts-data-root.test.mjs` passes on that synthetic merge. Code assets remain checkout-relative while user data follows the configured root; no blocking issue found.

tests/analysis-scripts-data-root.test.mjs
```diff
@@ -0,0 +1,143 @@
...
+    assert.equal(r.status, 0, `exited ${r.status}: ${r.all.slice(0, 400)}`);
+  } finally { cleanup(f); }
+});
+
+test('no analysis script derives a data path from its own directory', () => {
+  // The structural half, in the spirit of #3511's check 6. That one greps for a
+  // bare relative string literal; this spelling assigns __dirname to a constant
+  // and joins user-layer paths onto it, which reads as correct and is not.
+  const offenders = [];
+  const USER_LAYER = /join\(\s*(CAREER_OPS|CODE_ROOT)\s*,\s*'(data|interview-prep|reports|output|jds|documents)[/']/;
```
nikitacometa, line 135: Misses `resolveTrackerPath(CONST)`, the spelling this PR fixed in two files; re-breaking them under the `DATA_ROOT` name keeps this green.

tests/analysis-scripts-data-root.test.mjs
```diff
@@ -0,0 +1,143 @@
...
+  for (const file of ['funnel-velocity.mjs', 'calibrate.mjs', 'weekly-digest.mjs', 'stats.mjs', 'company-history.mjs']) {
```
nikitacometa, line 136: Five hardcoded names; `lib/mjs-files.mjs` would enumerate the domain, and six more scripts on main have the same class (set-status.mjs:263 among them).

calibrate.mjs
```diff
@@ -38,10 +38,16 @@ import { fileURLToPath } from 'url';
...
 import { canonicalOutcome } from './lib/outcome-types.mjs';
 import { outcomeDirsFor } from './lib/outcome-dir.mjs';
 
-const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
+// The USER's data root, not this file's directory. It was __dirname under the
+// name CAREER_OPS, so resolveTrackerPath() looked inside the checkout and a user
+// with CAREER_OPS_ROOT (or a .career-ops-data marker) got "No tracker found ...
+// nothing to calibrate yet" — which reads as "you have no outcome data",
+// exactly the thing this advisory is meant to answer.
+const DATA_ROOT = getCareerOpsRoot();
```
nikitacometa, line 50: With `CAREER_OPS` gone, the `fileURLToPath` and `dirname` imports above are unused; same in weekly-digest.mjs:41-42.

nikitacometa (approved):
Fix verified at head: reverting the three files reddens all three runtime assertions, at head the 6-case suite passes, `funnel-velocity --self-test` still resolves `templates/` from the code root, and nothing writes into the cwd.

One note for a follow-up, not blocking. tests/analysis-scripts-data-root.test.mjs:135 matches only `join(CAREER_OPS|CODE_ROOT, 'data…')`, so it does not see `resolveTrackerPath(<const>)`, the spelling in two of the three files this PR fixes. Re-breaking calibrate.mjs and funnel-velocity.mjs while keeping the `DATA_ROOT` name leaves that check green (the two runtime tests catch it, the structural one does not). The same class is still on main in six files: set-status.mjs:263, sync-pdf-flags.mjs:23, outcome.mjs:145, mark-pdf-ready.mjs:105, negotiation-roi.mjs:70, plugins.mjs:35. Probe: `CAREER_OPS_ROOT=$D node set-status.mjs 1 Interview` answers "No tracker found at <checkout>/applications.md", and set-status.mjs is the canonical tracker write path. Worth an issue.

Also dead imports at calibrate.mjs:37-38 and weekly-digest.mjs:41-42.

PR #3716, opened 2026-09-02
fix(web): bind the dashboard to loopback by default

nikolaysm:
## What does this PR do?

`next dev` / `next start` with no `-H` reach `server.listen(port, undefined)`, which binds every network interface. The dashboard is a local-first tool with no login, so the interface it listens on is what decides who can reach it — it should default to this machine only.

The bind is derived from `CAREER_OPS_WEB_ALLOWED_HOSTS`, the variable `origin-guard.mjs` already reads, so the socket and the request guard are driven by one value instead of two independent settings.

Review (thanks @nikitacometa) established that the bind alone does not close the exposure — a DNS-rebinding page reaches the ungated page routes from the user's own browser regardless of what the socket is bound to. That is fixed here too; see *The guard now runs on page routes* below.

## Related issue

None — bug fix, sent straight in per the template's note.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] If this is a new feature or architecture change, I opened an issue first (bug fixes are exempt)
- [x] I ran `node test-all.mjs` — green in CI on all three OS legs. `web` suite: 546/546, `tsc --noEmit` clean.
- [x] My changes respect the [Data Contract](https://github.com/career-ops-hq/career-ops/blob/main/DATA_CONTRACT.md) (system layer only)

---

### Design notes

**One variable, both layers.** Unset (or blank) binds `127.0.0.1`. Naming a host that loopback cannot already serve opens the socket, with the guard still answering for loopback and the hosts named. Naming *only* loopback (`localhost`, `127.0.0.1`, `::1`) is deliberately **not** an opt-in — widening for it would open every interface while granting nothing the guard did not already admit.

**The bind is as narrow as the opt-in allows.** A single named IP literal binds *that address*, not `0.0.0.0`: opting in one LAN box should not also publish the dashboard on a VPN tunnel or a phone hotspot. One socket carries one address, so loopback stops answering — the startup line says so, since the symptom otherwise looks like a server that failed to start, and naming a loopback host alongside is the documented way back to `0.0.0.0`. A hostname (which needs a resolver the launcher deliberately does not have) or two addresses (which cannot share a socket) still fall back to `0.0.0.0`.

**The variable is validated, not just parsed.** Every token the loopback predicate rejected used to count as an opt-in, so `CAREER_OPS_WEB_ALLOWED_HOSTS=off` bound every interface for someone trying to turn exposure *off* — and `off`, `false`, `no` and `0` are all syntactically valid hostnames, so only an explicit check catches them. Values are now validated as hostnames or IP literals, with the run aborted rather than widened. Refused by name:

| value | why |
|---|---|
| `off`, `false`, `no`, `0`, `none`, … | switch-like words — valid hostnames, so silence was not an option |
| `http://192.168.1.50` | host normalisation cuts at the first colon, so a URL used to arrive as the bare host `http` — valid, non-loopback, and an opt-in to everything |
| `0.0.0.0`, `::`, `0::0`, `0:0:0:0:0:0:0:0` | wildcards are bind targets, not host identities. Accepting one opened the socket while the guard went on matching `Host` literally — honest clients got `403`, and only a client spelling `Host: 0.0.0.0` was let in |
| `fe80::…` | link-local: ambiguous across interfaces, unbindable without a zone index |
| `224.0.0.1`, `ff02::1`, `255.255.255.255` | multicast and limited-broadcast — TCP is point-to-point, so these can never be listened on |
| `2001:db8:1`, `192.168.1`, `010.0.0.1` | incomplete or leading-zero literals |

The last row is why the hand-rolled parsers went: `isIPv6Literal` counted hex groups and `::` runs, so `2001:db8:1` passed, became a single-address bind, and reached `next -H`, where `listen()` refused to start the server. Leniency in that predicate cannot fail safe, because whatever it admits becomes the bind target. `node:net`'s `isIP` is the parser to use, and it covers the leading-zero IPv4 case for free. The module keeps its no-I/O, no-process-state property — it now imports `node:net` for one pure predicate, and only `server.mjs` and the tests import it (`proxy.ts` imports `origin-guard.mjs`, which stays free of node imports for the edge runtime).

**The guard now runs on page routes.** The matcher was `/api/:path*`, so no page route was filtered on any bind. A page that re-resolves its own domain to `127.0.0.1` fetches `/pipeline` same-origin from the user's own browser, and `force-dynamic` pages read the tracker off disk per request — a loopback bind does not close that, because the browser making the request is already on the machine. What does close it is the Host filter: a browser sets `Host` from the URL and will not let script forge it, so the rebound request carries the attacker's domain and is refused — but only if the guard runs. Verified against a live server: that request goes 200 → 403, while normal navigation and same-origin fetches are unaffected. Build assets stay excluded; they are identical for every user, and a caller who cannot read a page has no use for the chunk that renders it.

**`-H` is still yours, and the launcher stops guessing what it means.** The bind is injected only when no hostname flag was supplied. When one was, `next` owns the decision and the notice says exactly that, rather than naming an address the launcher would have to derive by reimplementing commander's resolution — a bet this launcher already lost once, when a missed spelling of the attached short form (`-H0.0.0.0`) had it announce "loopback only" while every interface was open. The cost is deliberate conservatism: an `-H` that *narrows* to loopback is also reported as next's to choose. A warning that is never false beats one that is precise until the parser drifts.

**Why a launcher.** npm scripts cannot branch on an env var portably given the Windows CI leg and no `cross-env`, so `web/server.mjs` resolves the bind and spawns `next` through this Node rather than `node_modules/.bin/next`, which is a `.cmd` shim on Windows that `spawn()` cannot exec without a shell. It follows `build-dashboard.mjs`'s shape and prints the interface it is about to listen on *before* resolving next, so a broken install still reports what the run would have exposed.

**Policy is pure and tested on values.** `server.mjs` only prints and spawns. `web/tests/lib/server-launcher.test.mjs` runs the real launcher against a stub `next` in a temp dir that echoes its argv, so the tests assert the argv `next` actually receives rather than the launcher's source text. The matcher is extracted from `proxy.ts` and exercised in the guard's own suite, since a guard that never runs is indistinguishable from one that always allows.

### Rebased onto main after #3597

#3597 (@andrgavrilenko) landed in `origin-guard.mjs` while this was open, which put the branch in conflict. Rebased; two files needed resolution, and two consequences are worth calling out rather than leaving to be discovered.

**The `checkRequest` hunk was adjacent, the header hunk was not.** #3597's header says *"Host layer (F2): only answer on a loopback Host by default"* — retracting exactly that is what this PR is for, since `Host` is client-chosen and a LAN request spelling `Host: localhost` satisfies it. Resolved by taking either side wholesale, that claim comes back. The merged header keeps the F2 correction and folds the allowlist paragraph into F1; `allowedOrigins` is passed through unchanged.

**The origin allowlist now reaches past `/api/*`.** #3597 shipped gating the API; this matcher spans every route but build assets, so an opted-in `chrome-extension://` origin now reaches page routes too. Measured against a running server:

| | allowlisted origin | `evil.test` |
|---|---|---|
| `/pipeline`, `/cv` | 200 | 403 |
| `/api/*` | passes guard | 403 |

Build assets stay outside the matcher — they load with no `Origin`; the 403 a cross-origin request gets for them in dev is Next's own `allowedDevOrigins` block, not this guard (plain `Unauthorized`, where this guard returns JSON). Both invariants #3597 pinned still hold: the host layer runs first and independently, and `null` is still refused. The widening looks right — an extension that can read the API can already read what the pages render — but it is a scope change that PR did not ask for.

One README line needed fixing on top: #3597's paragraph opened *"`/api` is gated by the same-origin + loopback guard"*, accurate when written and made false by this matcher. It now reads "Every route but build assets is gated…", squashed into the matcher commit so no commit contradicts itself.

### Deliberately out of scope

- `CAREER_OPS_WEB_ALLOWED_HOSTS` recognises the loopback spellings `isLoopbackHost` already knows; unusual ones (`127.1`, `localhost.`, `::ffff:127.0.0.1`) still widen. No worse than before this PR, and narrowing that predicate would change the request guard's behaviour, which this diff should not do. The IPv4-mapped form (`::ffff:224.0.0.1`) slips past the non-unicast check for the same reason — fixing it in `bind-host.mjs` alone would make the two predicates disagree about what an address is, which is the failure this PR exists to prevent.
- A subnet-directed broadcast (`192.168.1.255`) is not classified: whether it is one depends on the interface prefix the machine holds, which this pure planner does not inspect.
- `origin-guard.mjs`'s predicates are untouched — only its threat-model header changed, to record that the bind and the guard answer different attackers: the socket answers callers who connect directly, the guard answers callers arriving through a browser the user already trusts.
- There is still no login, and on an opened socket `Host` remains client-chosen — so the allow-list filters honest clients rather than hostile ones. The bind stays the real access control. Documented in `web/README.md` rather than papered over.

### DIFF
--- a/web/README.md
+++ b/web/README.md
@@ -22,6 +22,9 @@ npm run dev
 Open http://localhost:3000. The app reads the career-ops checkout it lives in
 (the parent directory) — your existing CV, pipeline and reports appear as-is.
 
+The server binds IPv4 loopback, so reach it as `localhost` or `127.0.0.1`;
+`http://[::1]:3000` has nothing listening on it.
+
 ## What works today
 
 - **Pipeline** — your tracker as a sortable, filterable table; status changes
@@ -37,6 +40,10 @@ Open http://localhost:3000. The app reads the career-ops checkout it lives in
 
 - **Local-first:** the local web app runs entirely on your machine — no cloud,
   no account needed. Your CV and data stay in your own files.
+- **Loopback by default:** `npm run dev` and `npm start` bind `127.0.0.1`, so
+  nothing else on your network can reach the dashboard. It has no login — the
+  bind is the access control. Opt out deliberately with
+  `CAREER_OPS_WEB_ALLOWED_HOSTS` ([below](#exposing-it-beyond-this-machine)).
 - **Never auto-submits:** the apply flow drafts and prefills; submitting is
   always a human action.
 - **CV generation never asks the agent to write:** the `pdf` worker tailors your
@@ -74,11 +81,57 @@ from a `chrome-extension://` origin, which Fetch Metadata always reports as
 `cross-site`, so the guard refuses it unless the id is named here. The host
 layer still applies to an allowlisted origin.
 
+Pass one-off flags through npm's `--`: `npm run dev -- -p 4000`.
+
+### Exposing it beyond this machine
+
+`CAREER_OPS_WEB_ALLOWED_HOSTS` is the switch. Unset (or blank) the server binds
+`127.0.0.1`; name a host it cannot already reach on loopback and the server binds
+`0.0.0.0`, with the API guard answering for loopback and the hosts you named:
+
+```bash
+CAREER_OPS_WEB_ALLOWED_HOSTS="192.168.1.50" npm run dev
+```
+
+Naming only loopback (`localhost`, `127.0.0.1`, `::1`) is not an opt-in and
+changes nothing — the guard already answers for those, so widening the socket
+would expose every interface while granting no host that wasn't already allowed.
+
+Set through `CAREER_OPS_WEB_ALLOWED_HOSTS`, neither layer can widen without the
+other — the bind and the allow-list are derived from that one value. They are not
+equivalent, and two things move them apart: the guard also answers for loopback,
+which no bind address changes, and a `-H` of your own moves the bind while the
+allow-list stays as it was (see below).
+
+Understand what you are turning on. The host allow-list covers `/api/*` only
+(`src/proxy.ts`'s matcher), so once the socket is open **every page the
+dashboard renders — your CV, pipeline and reports — is readable by anyone who
+can reach that port**, and there is no login. Two further caveats: a client that
+can reach the port can also supply its own `Host` header, so on a widened bind
+the allow-list filters honest clients rather than hostile ones; and `0.0.0.0` is
+IPv4-only, so an opted-in IPv6 address will pass the filter with nothing
+listening for it.
+
+You can also override the bind directly — `npm run dev -- -H 0.0.0.0` — since
+Next keeps the last value of a repeated option. **Prefer the env var.** An `-H`
+on its own opens the socket while the allow-list stays empty, which is the one
+combination with no honest reading: the guard then refuses LAN clients that
+identify themselves truthfully and admits any that spell `Host: localhost`. The
+launcher says so on startup rather than doing it quietly.
+
+Use it on a network you trust, and prefer an SSH tunnel
+(`ssh -L 3000:127.0.0.1:3000 you@box`) where you can — that needs no env var at
+all and keeps the bind on loopback.
+
 ### Tests
 
 Suites live in `web/tests/`, mirroring the path of what they test under
 `web/src/` — so `src/lib/clean-chips.mjs` is tested by
 `tests/lib/clean-chips.test.mjs`. Name the file `{module}.test.mjs`.
+
+Root-level scripts are the one exception: `server.mjs` is tested by
+`tests/lib/server-launcher.test.mjs`, because `test-all.mjs`'s parity check gates
+only `web/tests/lib/` and a suite outside it would never run in the required check.
 
 `npm test` discovers them with a glob (`tests/**/*.test.mjs`), so a new suite
 needs **no registration** — just add the file. **Requires Node ≥ 22**: earlier
--- a/web/README.md
+++ b/web/README.md
@@ -85,13 +85,31 @@ Pass one-off flags through npm's `--`: `npm run dev -- -p 4000`.
 
 ### Exposing it beyond this machine
 
-`CAREER_OPS_WEB_ALLOWED_HOSTS` is the switch. Unset (or blank) the server binds
-`127.0.0.1`; name a host it cannot already reach on loopback and the server binds
-`0.0.0.0`, with the API guard answering for loopback and the hosts you named:
+`CAREER_OPS_WEB_ALLOWED_HOSTS` is the list of hosts allowed to reach the
+dashboard. Unset (or blank) the server binds `127.0.0.1`; name a host it cannot
+already reach on loopback and the socket opens, with the guard answering for
+loopback and the hosts you named:
 
 ```bash
 CAREER_OPS_WEB_ALLOWED_HOSTS="192.168.1.50" npm run dev
 ```
+
+It is not an on/off switch, and it is not read as one: `off`, `false`, `no` and
+`0` are all valid hostnames, so treating them as words would have opened the
+socket to the whole network for someone typing the word for "closed". They are
+refused, as is anything else that cannot be read as a hostname or an IP address.
+To stay on loopback, unset the variable.
+
+**How the bind is chosen.** Name a single IP address and the socket binds exactly
+that address — `192.168.1.50` and nothing else. Binding `0.0.0.0` for it would
+also publish the dashboard on every other interface the machine has, a VPN tunnel
+or a phone hotspot included, none of which you asked for. One socket carries one
+address, so `http://localhost:3000` stops answering; open it at the address you
+named. If you want loopback too, name a loopback host alongside
+(`"192.168.1.50, localhost"`) — that binds `0.0.0.0` and accepts the wider
+exposure deliberately. Anything else — a hostname, which needs a resolver the
+launcher does not have, or two addresses, which cannot share a socket — binds
+`0.0.0.0`. The startup line says which of these happened.
 
 Naming only loopback (`localhost`, `127.0.0.1`, `::1`) is not an opt-in and
 changes nothing — the guard already answers for those, so widening the socket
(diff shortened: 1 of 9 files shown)


web/README.md
```diff
@@ -64,12 +71,58 @@ npm run build        # production build
...
+changes nothing — the guard already answers for those, so widening the socket
+would expose every interface while granting no host that wasn't already allowed.
+
+Set through `CAREER_OPS_WEB_ALLOWED_HOSTS`, neither layer can widen without the
+other — the bind and the allow-list are derived from that one value. They are not
+equivalent, and two things move them apart: the guard also answers for loopback,
+which no bind address changes, and a `-H` of your own moves the bind while the
+allow-list stays as it was (see below).
+
+Understand what you are turning on. The host allow-list covers `/api/*` only
```
nikitacometa: Pages are not gated by the socket. `src/proxy.ts:25` matches `/api/:path*` only, so a DNS-rebinding page reads `/pipeline` and `/cv` on a 127.0.0.1 bind too. Reword, or widen the matcher.

web/src/lib/bind-host.mjs
```diff
@@ -0,0 +1,161 @@
...
+ * interface while granting no host that was not already allowed — full LAN
+ * exposure from a value a user could reasonably read as a no-op, or even as a
+ * tightening.
+ *
+ * @param {string|undefined} envValue  CAREER_OPS_WEB_ALLOWED_HOSTS
+ * @returns {"0.0.0.0"|"127.0.0.1"}
+ */
+export function resolveBindHost(envValue) {
+  for (const host of parseAllowedHosts(envValue)) {
+    if (!isLoopbackHost(host)) return ALL_INTERFACES_BIND;
```
nikitacometa: Widens on any token `isLoopbackHost` rejects, so `CAREER_OPS_WEB_ALLOWED_HOSTS=off` gives 0.0.0.0 (measured). Validate as a hostname or IP literal and exit 1 on garbage. And when every named host is an IP literal, bind that address: `192.168.1.50` currently also exposes VPN and hotspot interfaces.

web/src/lib/bind-host.mjs
```diff
@@ -0,0 +1,161 @@
...
+  // socket. Naming it here keeps the notice readable and, more importantly,
+  // keeps it truthful — reporting the empty string as the bind would be
+  // accurate and useless, and treating it as absent would hide a real exposure.
+  const effectiveBindHost =
+    override === null ? bindHost : override.trim() === "" ? ALL_INTERFACES_BIND : override;
+
+  const allowedHosts = [...parseAllowedHosts(envValue)];
+  return {
+    ok: true,
+    argv: [command, "-H", bindHost, ...extra],
```
nikitacometa: `[command, "-H", bindHost, ...extra]` makes commander's last-wins load-bearing, and it is only tested against a stub. Inject `-H` when `overriddenHost(extra) === null` and the parser goes away.

web/README.md
```diff
@@ -64,12 +71,58 @@ npm run build        # production build
...
+listening for it.
+
+You can also override the bind directly — `npm run dev -- -H 0.0.0.0` — since
+Next keeps the last value of a repeated option. **Prefer the env var.** An `-H`
+on its own opens the socket while the allow-list stays empty, which is the one
+combination with no honest reading: the guard then refuses LAN clients that
+identify themselves truthfully and admits any that spell `Host: localhost`. The
+launcher says so on startup rather than doing it quietly.
+
+Use it on a network you trust, and prefer an SSH tunnel
```
nikitacometa, line 168: The container case is missing: inside a namespace the bind has to be 0.0.0.0 (`CAREER_OPS_WEB_ALLOWED_HOSTS=0.0.0.0` does it), and a forwarded port still sends `Host: localhost`, which the guard admits.

web/src/lib/origin-guard.mjs
```diff
@@ -6,17 +6,24 @@
...
 //
 //   Origin layer (F1): trust Sec-Fetch-Site when the browser sends it, fall
 //     back to comparing Origin against Host for older/non-browser clients.
-//   Host layer (F2): only answer on a loopback Host by default; extra hosts
-//     require an explicit opt-in (CAREER_OPS_WEB_ALLOWED_HOSTS).
+//   Host filter: answer only for loopback OR a host that was opted in via
+//     CAREER_OPS_WEB_ALLOWED_HOSTS. Note the loopback half is unconditional and
+//     is NOT derived from that variable.
+//
+// Nothing here closes F2, and nothing here can: Host is chosen by the client,
```
nikitacometa, line 36: True for a direct client, not for a browser: rebinding reaches the ungated page routes on loopback. Since this file is the threat model, one clause here.

nikitacometa (commented):
The bind change is right and the launcher tests hold at 239cd749: reverting `"dev"` to `"next dev"` turns two tests red, 52/52 pass locally. `dashboard/` is a TUI with no listener, so nothing to change there.

One correction to the new text. web/README.md:96 says pages are readable "once the socket is open". `src/proxy.ts:25` is `matcher: "/api/:path*"` and `checkRequest` has one caller, so no page route is filtered on any bind. A page that re-resolves its own domain to 127.0.0.1 fetches `/pipeline` same-origin, `pipeline/page.tsx:7` runs `pipelineSummary()` under `force-dynamic`, and the tracker comes back. Loopback does not close that, and the rewritten threat-model header reads as if it does. Widen the matcher to everything but `_next/*` and `favicon.ico`, or state the caveat.

Three smaller ones: `CAREER_OPS_WEB_ALLOWED_HOSTS=off` binds 0.0.0.0 (probe: `bind: "0.0.0.0", widened: true`); naming one LAN IP opens every interface, VPN and hotspot included, rather than that address; and the `-H` last-wins premise is asserted against a stub, never against Next 16.3.3. Injecting `-H` only when the caller supplied none removes that assumption and the ~24 lines of flag parsing with it. The container case (bind must be 0.0.0.0 inside a namespace, forwarded port still sends `Host: localhost`) is one line under README:112.

PR #3718, opened 2026-09-02
fix(gitignore): the tracker itself is unignored in the legacy root layout (#3717)

### LINKED ISSUE #3717: The tracker and follow-ups file are unignored in the legacy root layout
`resolveTrackerPath()` documents its own fallback chain, and the third entry is a supported install shape:

```
CAREER_OPS_TRACKER  >  <root>/data/applications.md  >  <root>/applications.md
```

`data/` is covered by the blanket rule at the top of `.gitignore`. The root spelling is covered by nothing:

```
  applications.md         NOT IGNORED       data/applications.md    ignored
  follow-ups.md           NOT IGNORED       data/follow-ups.md      ignored
  applications.db         ignored (#3469)
```

The tracker holds the user's entire job search — company, role, score, status, notes. `cv.md`, `portals.yml`, `article-digest.md` and `voice-dna.md` are all ignored at the root already.

The sharp version: **#3469 ignores the derived SQLite index on the argument that it "holds the same PII as the tracker"** — and it does. The tracker was sitting right beside it, unignored, one `git add .` from a commit.

### Shape

Unanchored is right here for the same reason it was for the `.db` rule: `getCareerOpsRoot()` resolves a relative `CAREER_OPS_ROOT` (or `.career-ops-data` marker) against the codebase directory, so a data root configured as `career-data` puts these inside the checkout, where `/applications.md` would not reach them.

Unlike the `.db` case there *are* tracked files with these names — the upgrade fixtures under `test-fixtures/`. They stay tracked through the existing `!test-fixtures/**` rule, which is a later pattern and wins, so the new rule has to sit above it. Worth pinning, because the ordering is invisible.

PR to follow.

abankar1:
Fixes #3717.

`resolveTrackerPath()`'s own documented fallback chain ends at `<root>/applications.md`, a supported install shape. `data/` is covered by the blanket rule at the top of `.gitignore`; the root spelling was covered by nothing.

```
  applications.md    NOT IGNORED        data/applications.md   ignored
  follow-ups.md      NOT IGNORED        data/follow-ups.md     ignored
```

The tracker holds the user's whole job search — company, role, score, status, notes. `cv.md`, `portals.yml`, `article-digest.md` and `voice-dna.md` are all ignored at the root already.

The sharp version: **#3469 ignores the derived SQLite index on the argument that it "holds the same PII as the tracker."** It does. The tracker was sitting right beside it, unignored.

## Shape, and the ordering trap

Unanchored, for the same reason as that `.db` rule: `getCareerOpsRoot()` resolves a relative `CAREER_OPS_ROOT` (or `.career-ops-data` marker) against the codebase directory, so a data root configured as `career-data` puts these inside the checkout, where `/applications.md` would not reach them. The `career-data/` probes are what decide that — an anchored rule passes every root probe.

Unlike the `.db` case, these names **are** carried by tracked files: the upgrade fixtures under `test-fixtures/`. They stay tracked through the existing `!test-fixtures/**` rule, which is a later pattern and therefore wins — so the new rule has to sit **above** it. That ordering is invisible in the file, so it is pinned in both directions:

| change | reddens |
|---|---|
| drop the rule | the 5 leak probes |
| move it below `!test-fixtures/**` | the 3 fixture probes |

One of the fixture probes is for `state-v1.20`, which does not exist yet — the next fixture must land the same way as the two that do.

Also verified directly that no currently tracked file became ignored:

```
git ls-files | git check-ignore --stdin   ->   empty
```

Full suite green at 8376.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,6 +104,26 @@ reports/*-RESERVED.md
 # original sitting untracked in the repo root. Trailing `*` covers both.
 *.bak*
 
+# The tracker and its follow-ups file in the LEGACY ROOT LAYOUT. resolveTrackerPath()
+# documents its own fallback chain — CAREER_OPS_TRACKER > <root>/data/applications.md
+# > <root>/applications.md — and that third entry is a supported install shape, not a
+# mistake. data/ is covered by the blanket rule at the top of this file; the root
+# spelling was not covered by anything.
+#
+# The content is the user's whole job search: company, role, score, status, notes.
+# cv.md, portals.yml, article-digest.md and voice-dna.md are all ignored at the root
+# already, and #3469 ignores the derived .db index on the argument that it "holds the
+# same PII as the tracker" — while the tracker itself sat beside it, unignored.
+#
+# Unanchored, like the .db rule below and for the same reason: getCareerOpsRoot()
+# resolves a relative CAREER_OPS_ROOT (or .career-ops-data marker) against the
+# codebase directory, so a data root configured as `career-data` puts these inside
+# the checkout, where an anchored rule would not reach them. The upgrade fixtures
+# that share these names stay tracked via the `!test-fixtures/**` rule further down,
+# which is a later pattern and therefore wins.
+applications.md
+follow-ups.md
+
 # The tracker's derived SQLite index (tracker.mjs, #918). It sits beside the
 # tracker markdown, so on the standard layout it lands in data/ and is already
 # covered by the rules above — but resolveTrackerPath() falls back to
--- a/tests/user-layer-gitignored.test.mjs
+++ b/tests/user-layer-gitignored.test.mjs
@@ -202,6 +202,53 @@ for (const path of timestampedBackupProbes) {
 // directory that can hold the index can hold all three names. Enumerating them
 // by hand is how the -shm sidecar ended up covered in the repo root and missed
 // one directory down.
+// The tracker and its follow-ups file in the LEGACY ROOT LAYOUT — the third
+// entry in resolveTrackerPath()'s own documented fallback chain
+// (CAREER_OPS_TRACKER > <root>/data/applications.md > <root>/applications.md),
+// and a supported install shape rather than a mistake.
+//
+// data/ is covered by the blanket rule at the top of .gitignore; the root
+// spelling was covered by nothing, so the file holding the user's entire job
+// search — company, role, score, status, notes — sat untracked and unignored,
+// one `git add .` from a commit. The derived .db index beside it was ignored
+// first, on the argument that it carries the same PII as the tracker.
+//
+// The career-data/ probes decide the SHAPE, exactly as they do for the index: a
+// relative CAREER_OPS_ROOT resolves against the codebase directory, so an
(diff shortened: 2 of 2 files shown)


Scott-Emberson (commented):
Reviewed at head 9fb502c2ab. This is a user-data protection fix and the owned test pins it in both directions. The .gitignore change adds bare applications.md and follow-ups.md above the test-fixtures negation, so the root-layout tracker (user PII, the same argument that already ignores the .db index) is ignored, and the test asserts it via real git check-ignore verdicts: the rootLayoutTrackerProbes require applications.md / follow-ups.md to come back ignored (five leak probes redden if the rule is dropped), while the trackedFixtureProbes require test-fixtures/upgrade/.../applications.md to stay not-ignored (three probes redden if the rule is moved below the negation). The hunk is purely additive, and the existing derived-index, symlink, user-layer and timestamped-backup probes are untouched.

Owned test verified sound, the PII-ignore rule is pinned bidirectionally. This one sits squarely in our area (a gitignore rule plus its owned test, no out-of-area feature code), so it is ready to approve on your go-ahead.

luochen211 (approved):
Reviewed the legacy-root ignore reconciliation and user-layer safety path. The current head merges cleanly with current main, `git diff --check` is clean, and `tests/user-layer-gitignored.test.mjs` passes on that synthetic merge. The change restores the tracked template while keeping the user-owned tracker ignored; no blocking issue found.

.gitignore
```diff
@@ -104,6 +104,26 @@ reports/*-RESERVED.md
...
+# already, and #3469 ignores the derived .db index on the argument that it "holds the
+# same PII as the tracker" — while the tracker itself sat beside it, unignored.
+#
+# Unanchored, like the .db rule below and for the same reason: getCareerOpsRoot()
+# resolves a relative CAREER_OPS_ROOT (or .career-ops-data marker) against the
+# codebase directory, so a data root configured as `career-data` puts these inside
+# the checkout, where an anchored rule would not reach them. The upgrade fixtures
+# that share these names stay tracked via the `!test-fixtures/**` rule further down,
+# which is a later pattern and therefore wins.
+applications.md
```
nikitacometa, line 124: Unanchored by design. One clause here that the name is now reserved at any depth (`docs/applications.md` would be ignored) and that re-inclusion needs an explicit `!` rule; same for `follow-ups.md` below.

nikitacometa (approved):
Verified by `git check-ignore -v` on both layouts: at origin/main `applications.md` and `follow-ups.md` come back NOT-IGNORED, at head both resolve to .gitignore:124/125, `career-data/applications.md` included. `git ls-files | git check-ignore --stdin` is empty at head, and `test-fixtures/upgrade/state-v1.16/data/applications.md` stays visible via the `!test-fixtures/**` rule at 193. Dropping the two lines reddens exactly the 5 leak probes (30 passed, 5 failed).

One note: the rules are unanchored, so `docs/applications.md` is ignored too now. That is the intended trade for a relative `CAREER_OPS_ROOT`, and the PII argument wins, but the comment block is the only place a future contributor learns it; one clause saying the names are reserved at every depth.

PR #3720, opened 2026-09-02
docs(agents): drop a script reference that never existed, and guard the rest (#3719)

### LINKED ISSUE #3719: AGENTS.md references scan-apify.mjs, which does not exist
`AGENTS.md:423`, in the jds/ capture-filename table:

```
| `plugins/apify/index.mjs`, `scan-apify.mjs` | `{company}-{role}-{sha1(url)[0:10]}.md` |
```

There is no `scan-apify.mjs` anywhere in the repo, and no commit in the history ever added or removed one — so it was wrong when written rather than left behind by a rename.

`plugins/apify/index.mjs` does write that shape (`index.mjs:127-135`), so the row is correct once the second name is dropped.

Small, but `AGENTS.md` is loaded into every agent session, which makes a path in it an instruction rather than prose. An agent that goes looking spends a tool call and cannot tell "this was renamed" from "I am in the wrong directory".

A census of backticked `*.mjs` references across `AGENTS.md`, `CLAUDE.md`, `CODEX.md`, `OPENCODE.md`, `GEMINI.md`, `KIMI.md` and `DATA_CONTRACT.md` finds exactly this one, so a guard would start green.

PR to follow.

abankar1:
Fixes #3719.

`AGENTS.md:423` listed `scan-apify.mjs` beside `plugins/apify/index.mjs` in the jds/ capture table. No commit in the history ever added or removed it — it was wrong when written, not left behind by a rename. The plugin does write that filename shape (`index.mjs:127-135`), so the row is correct once the second name is gone.

Small on its own. The reason it is worth a check rather than a one-line edit: `AGENTS.md` is loaded into every agent session, which makes a path in it an instruction an agent will act on, not prose. A name that resolves to nothing costs a tool call every time one goes looking, and the agent cannot tell *"this was renamed"* from *"I am in the wrong directory"*.

## The guard

Covers the seven documents a CLI loads automatically (`AGENTS.md`, `CLAUDE.md`, `CODEX.md`, `OPENCODE.md`, `GEMINI.md`, `KIMI.md`, `DATA_CONTRACT.md`), and is scoped to `.mjs` **deliberately** — these files also name templates, data files and user-layer paths that legitimately do not exist in a fresh checkout. A script does.

Backticked references only, so prose and example invocations are left alone. Files absent from a given fork are skipped rather than failed.

Exactly one ghost across all seven today, so it starts green. Restoring the reference reddens it. Full suite green at 8370.

### DIFF
--- a/tests/agent-docs-script-refs.test.mjs
+++ b/tests/agent-docs-script-refs.test.mjs
@@ -1,0 +1,55 @@
+// tests/agent-docs-script-refs.test.mjs — every `*.mjs` a top-level agent
+// document names has to exist.
+//
+// AGENTS.md is loaded into every agent session, so a path in it is not prose:
+// it is an instruction an agent will act on. A name that resolves to nothing
+// costs a tool call every time one goes looking, and the agent has no way to
+// tell "this was renamed" from "I am in the wrong directory".
+//
+// The one this was written for is `scan-apify.mjs`, listed in the jds/ capture
+// table beside plugins/apify/index.mjs. It never existed — no commit in the
+// history ever added or removed it — so it was wrong from the day it was
+// written rather than stale, which is the kind of error only a check like this
+// finds.
+//
+// Scoped to .mjs on purpose. These documents also name templates, data files
+// and paths that are created at runtime or belong to the user layer, none of
+// which need to exist in a fresh checkout. A script does.
+//
+// Run:  node --test tests/agent-docs-script-refs.test.mjs
+
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { readFileSync, existsSync } from 'node:fs';
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const ROOT = dirname(dirname(fileURLToPath(import.meta.url)));
+
+// The documents a CLI loads automatically. DATA_CONTRACT.md is included because
+// the agent instructions point at it as the authority on file ownership.
+const AGENT_DOCS = [
+  'AGENTS.md', 'CLAUDE.md', 'CODEX.md', 'OPENCODE.md',
+  'GEMINI.md', 'KIMI.md', 'DATA_CONTRACT.md',
+];
+
+// Backticked only. Prose can discuss a script that a user might write, and a
+// code fence can hold an example invocation; a backticked path is a reference.
+const MJS_REF = /`([A-Za-z0-9_./-]+\.mjs)`/g;
+
+for (const doc of AGENT_DOCS) {
+  test(`${doc} names no script that does not exist`, () => {
+    const path = join(ROOT, doc);
+    if (!existsSync(path)) return;   // not every CLI's file ships in every fork
+    const src = readFileSync(path, 'utf-8');
+    const missing = [...new Set([...src.matchAll(MJS_REF)].map((m) => m[1]))]
+      .filter((ref) => !existsSync(join(ROOT, ref)))
+      .sort();
+    assert.deepEqual(
+      missing,
+      [],
+      `${doc} names ${missing.length} script(s) that do not exist: ${missing.join(', ')}. `
+      + 'An agent reading this will spend a tool call looking for each one.',
+    );
+  });
+}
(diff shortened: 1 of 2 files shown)


Scott-Emberson (commented):
Reviewed at head 91edb330f4. The owned guard tests/agent-docs-script-refs.test.mjs does exactly what a drift guard should: it scans the docs' own backticked .mjs references and runs each through existsSync against the repo root, so a reference resolving to nothing lands in the missing list and reddens. There is no hardcoded allowlist, it reads the real referenced paths, it covers AGENTS.md plus the sibling instruction files (CLAUDE / CODEX / OPENCODE / GEMINI / KIMI / DATA_CONTRACT), it only skips a doc that is absent rather than skipping the check, and the backtick-scoped regex avoids false-flagging prose or code-fence examples. The fix is real: scan-apify.mjs returns 404 at head, and the diff removes it from the jds capture table, so the guard passes now and would redden if it or any listed script were dropped or renamed again.

The substance is AGENTS.md, a red-line agent-instruction file, so the doc change itself defers to santifer; from the tests/ side the guard is sound and genuinely catches doc drift on real disk state. Nothing blocking from our area.

luochen211 (changes requested):
The guard misses the most common form of an agent-doc script reference: a command. `MJS_REF = /\`([A-Za-z0-9_./-]+\\.mjs)\`/g` only matches an inline code span whose entire contents are the path. A span such as `node missing-script.mjs --flag` yields no match, so a nonexistent executable instruction stays green.

Please first extract inline-code spans and then extract `.mjs` path tokens inside them (while keeping any deliberate fence/example exclusions), and add command-form regressions. Agent docs already use backticked `node ...mjs ...` invocations, so checking only path-only spans does not enforce the stated guarantee.

tests/agent-docs-script-refs.test.mjs
```diff
@@ -0,0 +1,55 @@
...
+// The documents a CLI loads automatically. DATA_CONTRACT.md is included because
+// the agent instructions point at it as the authority on file ownership.
+const AGENT_DOCS = [
+  'AGENTS.md', 'CLAUDE.md', 'CODEX.md', 'OPENCODE.md',
+  'GEMINI.md', 'KIMI.md', 'DATA_CONTRACT.md',
+];
+
+// Backticked only. Prose can discuss a script that a user might write, and a
+// code fence can hold an example invocation; a backticked path is a reference.
+const MJS_REF = /`([A-Za-z0-9_./-]+\.mjs)`/g;
```
nikitacometa, line 38: Command form is missed: a span with a space in it never matches. Measured: a ghost `node ghost-runner-does-not-exist.mjs --flag` appended to AGENTS.md keeps the suite at pass 7 / fail 0.

tests/agent-docs-script-refs.test.mjs
```diff
@@ -0,0 +1,55 @@
...
+import assert from 'node:assert/strict';
+import { readFileSync, existsSync } from 'node:fs';
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const ROOT = dirname(dirname(fileURLToPath(import.meta.url)));
+
+// The documents a CLI loads automatically. DATA_CONTRACT.md is included because
+// the agent instructions point at it as the authority on file ownership.
+const AGENT_DOCS = [
```
nikitacometa, line 31: The list stops at the 7 top-level docs; `modes/*.md` are loaded per mode and carry the same class (modes/scan.md:345 names `workday.mjs`, resolvable only as `providers/workday.mjs`), so a wider list needs a basename fallback.

nikitacometa (commented):
The doc fix is right: `git log --all -S 'scan-apify' -- '*.mjs'` returns nothing before this commit, and plugins/apify/index.mjs:127-135 writes the shape the row now claims.

The guard's promise is wider than its regex. tests/agent-docs-script-refs.test.mjs:38 only matches a span whose entire contents are the path, so command form is invisible. Appending ``Run `node ghost-runner-does-not-exist.mjs --flag`.`` to AGENTS.md leaves the suite at pass 7 / fail 0; the bare-path form reddens it (pass 6 / fail 1). AGENTS.md holds about a dozen command-form spans today, `node set-status.mjs` (124) and `node merge-tracker.mjs` (443) among them, which are the instructions an agent is most likely to run. Extract spans first, then `.mjs` tokens inside them, or narrow the header's claim to path-only spans. `modes/*.md` are loaded the same way and carry the same class (modes/scan.md names `workday.mjs`, which resolves only as `providers/workday.mjs`), so extending the guard there needs a basename fallback.

PR #3722, opened 2026-09-02
fix(doctor): --json leaked git's "not a git repository" onto stderr (#3721)

### LINKED ISSUE #3721: doctor --json leaks git's "not a git repository" onto stderr
`checkTrackedBakFiles` (`doctor.mjs:168`) shells out to `git ls-files`, and `execFileSync`'s default hands the child the parent's stderr. Run anywhere that is not a git checkout — which is every user whose data root lives outside the repo — and git prints before the `catch` that handles exactly that case ever runs:

```
$ node doctor.mjs --json --target /tmp   > out  2> err
  stdout: 547 bytes (valid JSON)
  stderr:  69 bytes -> "fatal: not a git repository (or any of the parent directories): .git"
```

Nothing breaks: the JSON is valid and the check reports a clean skip. But `AGENTS.md` has every agent run `node doctor.mjs --json` on the **first message of every session**, so this is the hottest path in the repo, and a `fatal:` line is the first thing an agent sees.

### Note for a fix

Piping stderr rather than discarding it keeps the failure readable. The `catch` currently maps *every* failure to `skipped (not a git checkout)` — git missing, a permissions problem or a timeout would be reported as "not a checkout", which claims a check ran that did not.

PR to follow.

abankar1:
Fixes #3721.

`checkTrackedBakFiles` shells out to `git ls-files`, and `execFileSync`'s default hands the child our own stderr. Outside a checkout git printed before the `catch` that handles exactly that case ever ran:

```
                 stdout            stderr
  before         547 bytes (JSON)   69 bytes -> "fatal: not a git repository ..."
  after          547 bytes (JSON)    0 bytes
```

Nothing broke — the JSON stayed valid and the check still reported a clean skip. It just told every agent, every session, that something was fatally wrong when nothing was. `AGENTS.md` has every agent run `doctor --json` on the **first message of each session**, so this is the hottest path in the repo and it is the first thing an agent sees.

## Piped, not discarded

Muting stderr would have fixed the symptom and hidden a real failure. The `catch` mapped *every* git failure to `skipped (not a git checkout)`, so git missing, a permissions problem or a timeout all reported as "not a checkout" — claiming a check ran that did not.

Now the reason is ours to read: not-a-checkout (and `ENOENT`) stays a clean skip; anything else surfaces as a warning naming what happened.

## Tests

Stderr is measured in **bytes**, not matched against the message — the guarantee is that nothing reaches stderr on a clean run, so a future subprocess leaking a different string reddens this too.

Two guards beside it, both aimed at the ways this fix could cheat:

- the JSON envelope stays complete (a clean stderr bought by suppressing the output would fail),
- inside a real checkout the `.bak` check still runs rather than reporting itself broken (a clean stderr bought by making the check a permanent skip would fail).

Reverting the `stdio` option reddens the first test only. Full suite green at 8370.

### DIFF
--- a/doctor.mjs
+++ b/doctor.mjs
@@ -166,11 +166,29 @@ function checkTrackedBakFiles(root) {
   let raw;
   try {
     raw = execFileSync('git', ['ls-files', '-z', '--', '*.bak*'], {
-      cwd: root, encoding: 'utf-8', timeout: 5000,
+      cwd: root,
+      encoding: 'utf-8',
+      timeout: 5000,
+      // stderr PIPED, not inherited. execFileSync's default hands the child our
+      // own stderr, so outside a checkout git printed
+      //   fatal: not a git repository (or any of the parent directories): .git
+      // before this catch ever ran — ahead of the JSON on `doctor --json`, which
+      // AGENTS.md has every agent run on the first message of every session. The
+      // catch below already handles that case; git's own message added nothing
+      // but the appearance of something being broken.
+      stdio: ['ignore', 'pipe', 'pipe'],
     });
-  } catch {
-    // Not a git checkout (or git unavailable) — nothing to check.
-    return { pass: true, label: 'Tracked .bak files: skipped (not a git checkout)' };
+  } catch (err) {
+    // Piping means the reason is ours to read rather than the terminal's. Not
+    // being a checkout is the expected case and stays a clean skip; anything
+    // else — git missing, a permissions problem, a timeout — is reported, since
+    // silently skipping those would claim a check ran that did not.
+    const stderr = String(err?.stderr ?? '');
+    if (/not a git repository/i.test(stderr) || err?.code === 'ENOENT') {
+      return { pass: true, label: 'Tracked .bak files: skipped (not a git checkout)' };
+    }
+    const detail = stderr.trim().split('\n')[0] || err?.message || 'unknown error';
+    return { warn: true, label: `Tracked .bak files: check could not run (${detail})` };
   }
   const paths = raw.split('\0').filter(Boolean);
   if (paths.length === 0) {
--- a/tests/doctor-clean-stderr.test.mjs
+++ b/tests/doctor-clean-stderr.test.mjs
@@ -1,0 +1,82 @@
+// tests/doctor-clean-stderr.test.mjs — `doctor --json` must not leak a
+// subprocess's diagnostics onto stderr.
+//
+// AGENTS.md has every agent run `node doctor.mjs --json` on the FIRST MESSAGE of
+// every session, which makes this the hottest path in the repo. checkTrackedBakFiles
+// shells out to `git ls-files`, and execFileSync's default hands the child our own
+// stderr — so outside a git checkout git printed
+//
+//     fatal: not a git repository (or any of the parent directories): .git
+//
+// before the catch that handles exactly that case ever ran. The JSON on stdout
+// stayed valid, so nothing broke; it just told every agent, every session, that
+// something was fatally wrong when nothing was.
+//
+// Measured in bytes rather than by matching the message, because the point is
+// that NOTHING reaches stderr on a clean run — a future subprocess leaking a
+// different string should redden this too.
+//
+// Run:  node --test tests/doctor-clean-stderr.test.mjs
+
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { spawnSync } from 'node:child_process';
+import { mkdtempSync, rmSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const ROOT = dirname(dirname(fileURLToPath(import.meta.url)));
+
+function doctor(cwd, target) {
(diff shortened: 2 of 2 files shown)


Scott-Emberson (commented):
Reviewed at head ec021afd8d. The owned test tests/doctor-clean-stderr.test.mjs pins the leak fix with a content assertion, not just an exit code. Run in a fresh temp dir outside any checkout (the exact arrangement where the old fatal: not a git repository leaked from the tracked-.bak check), it asserts stderr is exactly the empty string, so any returned byte including the old message reddens, and it parses stdout from the first brace as JSON asserting onboardingNeeded plus the missing/warnings arrays, so a clean stderr bought by suppressing all output would fail the JSON check. A third case run inside a real repo asserts stderr stays clean and no "check could not run" warning appears, guarding against the fix turning the .bak check into a permanent skip. The three assertions map one to one to the doctor.mjs branches (pipe git stderr, keep the clean skip for not-a-git-repo/ENOENT, warn otherwise).

The substance of the fix is in doctor.mjs, outside our owned paths, so this is a test-axis review: the owned test is sound, the merge and scope of the doctor change are the maintainer's call. Nothing blocking from the tests/ side.

luochen211 (changes requested):
The catch conflates a missing Git executable with a non-Git directory. For `execFileSync('git', ...)`, `err.code === 'ENOENT'` is exactly what occurs when Git cannot be launched; an ordinary directory outside a checkout instead returns Git's status 128 plus "not a git repository" stderr.

Repro: invoke `doctor.mjs --json` with an absolute Node path and `PATH=/definitely-missing`. It exits 0 with `warnings: []`, silently claiming the check was skipped as a normal non-checkout even though the comment says "git missing ... is reported." Please reserve the clean skip for the explicit non-repository stderr case, warn on ENOENT, and add the missing-Git regression alongside the clean-stderr tests.

doctor.mjs
```diff
@@ -166,11 +166,29 @@ function checkTrackedBakFiles(root) {
...
-  } catch {
-    // Not a git checkout (or git unavailable) — nothing to check.
-    return { pass: true, label: 'Tracked .bak files: skipped (not a git checkout)' };
+  } catch (err) {
+    // Piping means the reason is ours to read rather than the terminal's. Not
+    // being a checkout is the expected case and stays a clean skip; anything
+    // else — git missing, a permissions problem, a timeout — is reported, since
+    // silently skipping those would claim a check ran that did not.
+    const stderr = String(err?.stderr ?? '');
+    if (/not a git repository/i.test(stderr) || err?.code === 'ENOENT') {
```
nikitacometa, line 187: `ENOENT` is git-cannot-be-launched, not not-a-checkout; probed with `PATH=/definitely-missing`, it reports `skipped (not a git checkout)`. Also: the `not a git repository` text is localized when `LANG` is set; pass `LC_ALL=C` to the child so the match holds.

doctor.mjs
```diff
@@ -166,11 +166,29 @@ function checkTrackedBakFiles(root) {
...
+      return { pass: true, label: 'Tracked .bak files: skipped (not a git checkout)' };
+    }
+    const detail = stderr.trim().split('\n')[0] || err?.message || 'unknown error';
+    return { warn: true, label: `Tracked .bak files: check could not run (${detail})` };
```
nikitacometa, line 191: No test reaches this branch; a PATH-stripped spawn covers it.

tests/doctor-clean-stderr.test.mjs
```diff
@@ -0,0 +1,82 @@
...
+import { mkdtempSync, rmSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const ROOT = dirname(dirname(fileURLToPath(import.meta.url)));
+
+function doctor(cwd, target) {
+  const r = spawnSync(process.execPath, [join(ROOT, 'doctor.mjs'), '--json', '--target', target], {
+    cwd, encoding: 'utf-8', timeout: 60_000,
```
nikitacometa, line 33: The helper already sets a custom `env`, so a `PATH`-stripped variant with an absolute node path costs one case.

nikitacometa (changes requested):
Leak fix verified: from a temp dir origin/main writes 69 bytes to stderr (`fatal: not a git repository …`), this head writes 0 with the JSON intact, and reverting the `stdio` option reddens exactly the first test. doctor.mjs:168 is the only subprocess call in the file, so the class is closed.

doctor.mjs:187 puts `ENOENT` in the clean-skip branch, and `ENOENT` is what `execFileSync` throws when git cannot be launched. Probe: `env PATH=/definitely-missing $(which node) doctor.mjs --json --target $D` exits 0 with no git warning, and `--summary` prints `✓ Tracked .bak files: skipped (not a git checkout)`. That is the "claiming a check ran that did not" the comment three lines above argues against. Gate the skip on the `not a git repository` stderr only, warn on ENOENT with "git not found", and add a PATH-stripped regression: the warn branch at 191 has no test at all.

One more on the same line: git localizes that message when `LANG`/`LC_MESSAGES` is set, so on a localized install a plain non-checkout would not match `/not a git repository/i` and would surface as the new warning instead of a skip. Run the child with `LC_ALL=C` in its env and the match is stable.

PR #3728, opened 2026-09-03
fix(icims): read every jobLocation entry when picking the location

### LINKED ISSUE #3727: icims: pickLocation only reads the first jobLocation entry
\`providers/icims.mjs\` \`pickLocation()\` takes \`node.jobLocation[0]\` and ignores the rest of the array. When that first entry has no usable address (iCIMS writes the literal \`UNAVAILABLE\` into fields it has no value for) and a later entry does, the posting comes out with no location, so \`location_filter\` never sees the country or region that was actually in the feed.

Repro against \`ffb49be1\`, a single JSON-LD node with two locations:

\`\`\`js
pickLocation([{ jobLocation: [
  { address: { addressCountry: 'UNAVAILABLE' } },
  { address: { addressLocality: 'Toronto', addressRegion: 'ON', addressCountry: 'CA' } },
]}]);
// returns null, expected "Toronto, ON, Canada"
\`\`\`

Suggested fix: iterate every \`jobLocation\` entry and take the first one that yields any address parts, keeping the existing \`UNAVAILABLE\` cleanup. Found while re-vendoring the providers into CareerRat; happy to send a PR if you'd rather.

scttbnsn:
## What does this PR do?

pickLocation() only looked at node.jobLocation[0]. When that entry is all UNAVAILABLE and a later one has a real address, the posting ends up with no location and location_filter can't apply. Now it walks every entry and takes the first one that yields any address parts. Same UNAVAILABLE cleanup and country mapping as before. Fixes #3727.

## Related issue

Fixes #3727

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

---

## Summary

`providers/icims.mjs:pickLocation()` now checks every `jobLocation` entry. It skips entries with only `UNAVAILABLE` or empty fields and returns the first usable address.

Country and region mapping remain available for `location_filter`.

`tests/providers/icims.test.mjs` adds regression coverage for multiple locations and preserves single-location behavior.

Touched system files: `providers/` and `tests/providers/`. No changes affect `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, or `.github/`.

### DIFF
--- a/providers/icims.mjs
+++ b/providers/icims.mjs
@@ -179,27 +179,32 @@ export default {
 const COUNTRY_NAMES = { US: 'United States', CA: 'Canada' };
 
 // From flattened JSON-LD nodes, build "Locality, Region, Country" out of the
-// first JobPosting's first jobLocation address. Partial addresses are kept:
-// the country alone is already enough for location_filter to decide.
+// first jobLocation entry (across all JobPosting nodes) that yields any usable
+// parts. Some tenants emit an all-UNAVAILABLE entry ahead of the real address,
+// so reading only the first entry would return no location even though a
+// later one has one. Partial addresses are kept: the country alone is already
+// enough for location_filter to decide.
 function pickLocation(nodes) {
   for (const node of nodes) {
     if (!node || typeof node !== 'object' || !node.jobLocation) continue;
-    const place = Array.isArray(node.jobLocation) ? node.jobLocation[0] : node.jobLocation;
-    const addr = place?.address;
-    if (!addr || typeof addr !== 'object') continue;
-    const clean = v => {
-      const s = String(v ?? '').trim();
-      // iCIMS writes the literal string UNAVAILABLE into fields it has no value
-      // for, so an unchecked join yields "UNAVAILABLE, MD, United States".
-      return !s || /^unavailable$/i.test(s) ? '' : s;
-    };
-    const country = clean(addr.addressCountry);
-    const parts = [
-      clean(addr.addressLocality),
-      clean(addr.addressRegion),
-      COUNTRY_NAMES[country.toUpperCase()] || country,
-    ].filter(Boolean);
-    if (parts.length) return parts.join(', ');
+    const places = Array.isArray(node.jobLocation) ? node.jobLocation : [node.jobLocation];
+    for (const place of places) {
+      const addr = place?.address;
+      if (!addr || typeof addr !== 'object') continue;
+      const clean = v => {
+        const s = String(v ?? '').trim();
+        // iCIMS writes the literal string UNAVAILABLE into fields it has no value
+        // for, so an unchecked join yields "UNAVAILABLE, MD, United States".
+        return !s || /^unavailable$/i.test(s) ? '' : s;
+      };
+      const country = clean(addr.addressCountry);
+      const parts = [
+        clean(addr.addressLocality),
+        clean(addr.addressRegion),
+        COUNTRY_NAMES[country.toUpperCase()] || country,
+      ].filter(Boolean);
+      if (parts.length) return parts.join(', ');
+    }
   }
   return null;
 }
--- a/tests/providers/icims.test.mjs
+++ b/tests/providers/icims.test.mjs
@@ -244,6 +244,24 @@ const mkCtx = (pages) => ({
   if (job.location === 'MD, United States') pass('enrichDate drops UNAVAILABLE address parts');
   else fail(`location: ${job.location}`);
 }
+// Regression for #3727: the first jobLocation entry is entirely UNAVAILABLE
+// and the real address lives in a later entry. Reading only jobLocation[0]
+// returned no location at all even though a usable one was in the array.
+{
+  const detail = `<script type="application/ld+json">{"@type":"JobPosting","jobLocation":[{"address":{"addressCountry":"UNAVAILABLE"}},{"address":{"addressLocality":"Toronto","addressRegion":"ON","addressCountry":"CA"}}]}</script>`;
+  const job = { title: 'X', url: `${ORIGIN}/jobs/1234/x/job`, company: 'acmefreight', location: '' };
+  await icims.enrichDate(job, { fetchText: async () => detail });
+  if (job.location === 'Toronto, ON, Canada') pass('enrichDate walks past an all-UNAVAILABLE jobLocation entry to a later one');
+  else fail(`location: ${job.location}`);
+}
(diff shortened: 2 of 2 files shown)


Scott-Emberson (commented):
Reviewed at head 44c44559e5. In our gated area (providers/ plus tests/providers/). The fix makes pickLocation walk every jobLocation entry (Array-normalized, with an inner loop returning the first entry that yields parts) instead of reading only jobLocation[0], and the owned test pins it: a job whose first jobLocation entry is all-UNAVAILABLE and whose second is a real Toronto / ON / CA address is asserted to resolve to "Toronto, ON, Canada", which reddens on a regression to reading only entry zero (entry zero cleans to empty, yields no parts, returns null, and the location stays blank). A second added case locks the single-entry behavior ("Denver, CO, United States"). The change is additive, the pre-existing assertions (enrichDate dropping UNAVAILABLE parts, list-page location being authoritative over enrichment) are untouched, and only pickLocation changed so there is no new fetch, host, or SSRF surface.

Owned test verified sound. The branch is behind main but mergeable with CI green. This is in our area, ready to approve on your go-ahead (a rebase before merge would clear the behind-main state).

luochen211 (approved):
Reviewed multi-entry iCIMS location extraction, fallback behavior, and malformed-input handling. The current head merges cleanly with current main, `git diff --check` is clean, and `tests/providers/icims.test.mjs` passes on that synthetic merge. No blocking issue found.

Scott-Emberson (commented):
Re-checked at head 2e910d6ef4. On a blob compare the owned test tests/providers/icims.test.mjs is byte-identical to what I reviewed, so the head move is a rebase without a content change in our area. The earlier verification stands: pickLocation walks every jobLocation entry, the added case asserts an all-UNAVAILABLE first entry plus a real Toronto second yields the Toronto location (reddening on a [0]-only regression), a single-entry case locks the unchanged behavior, and no fetch or host surface changed.

Owned test verified unchanged and sound. The only thing holding it is that the branch is behind main. Rebase to current main and I will re-verify and clear it to approve.

Scott-Emberson (commented):
Re-checked at head 71244b74c8. On a blob compare the owned test tests/providers/icims.test.mjs is byte-identical to what I reviewed, so this head move is a pure rebase with no content change in our area. The earlier verification stands: pickLocation walks every jobLocation entry, and the case asserting an all-UNAVAILABLE first entry plus a real Toronto second yields the Toronto location (reddening on a first-entry-only regression) is intact, with no fetch or host surface change.

Owned test verified unchanged and sound. CI is green and the branch is up to date (blocked only on our review), so this is ready to approve on your go-ahead.

providers/icims.mjs
```diff
@@ -179,27 +179,32 @@ export default {
...
-    };
-    const country = clean(addr.addressCountry);
-    const parts = [
-      clean(addr.addressLocality),
-      clean(addr.addressRegion),
-      COUNTRY_NAMES[country.toUpperCase()] || country,
-    ].filter(Boolean);
-    if (parts.length) return parts.join(', ');
+    const places = Array.isArray(node.jobLocation) ? node.jobLocation : [node.jobLocation];
+    for (const place of places) {
```
nikitacometa, line 191: Mutating this to `[...places].reverse()` (last usable wins) leaves `--only icims` at 38 passed, 0 failed. The added fixture has one usable entry, so the ordering the docblock promises is asserted nowhere.

providers/icims.mjs
```diff
@@ -179,27 +179,32 @@ export default {
 const COUNTRY_NAMES = { US: 'United States', CA: 'Canada' };
 
 // From flattened JSON-LD nodes, build "Locality, Region, Country" out of the
-// first JobPosting's first jobLocation address. Partial addresses are kept:
-// the country alone is already enough for location_filter to decide.
+// first jobLocation entry (across all JobPosting nodes) that yields any usable
```
nikitacometa, line 182: "yields any usable parts" is weaker than "first real address": at this head `[{address:{addressLocality:'Remote',addressCountry:'UNAVAILABLE'}}, Toronto, Denver]` returns `"Remote"`. Same on main, but worth saying here that only the literal `UNAVAILABLE` sentinel is skipped.

tests/providers/icims.test.mjs
```diff
@@ -244,6 +244,24 @@ const mkCtx = (pages) => ({
...
   else fail(`location: ${job.location}`);
 }
+// Regression for #3727: the first jobLocation entry is entirely UNAVAILABLE
+// and the real address lives in a later entry. Reading only jobLocation[0]
+// returned no location at all even though a usable one was in the array.
+{
+  const detail = `<script type="application/ld+json">{"@type":"JobPosting","jobLocation":[{"address":{"addressCountry":"UNAVAILABLE"}},{"address":{"addressLocality":"Toronto","addressRegion":"ON","addressCountry":"CA"}}]}</script>`;
+  const job = { title: 'X', url: `${ORIGIN}/jobs/1234/x/job`, company: 'acmefreight', location: '' };
+  await icims.enrichDate(job, { fetchText: async () => detail });
+  if (job.location === 'Toronto, ON, Canada') pass('enrichDate walks past an all-UNAVAILABLE jobLocation entry to a later one');
```
nikitacometa, line 254: Live check on `careers-en-nortal.icims.com` (3 postings) shows `jobLocation` as a 1-element array with `UNAVAILABLE` inside fields, not as a whole leading entry, so this fixture stays synthetic. Not an objection; a real tenant named here would help the next reader.

nikitacometa (approved):
Verified at 71244b74. `--only icims` 38/0; reverting providers/icims.mjs to the base gives 37/1. Probed `pickLocation` through `enrichDate` on both heads: 0 entries, 1 entry and the bare object shape are byte-identical, `[UNAVAILABLE, Toronto, Denver]` goes from `""` to `"Toronto, ON, Canada"`, and an entry whose `address` is a string instead of an object no longer suppresses the rest. No input resolves worse than before.

One test gap. icims.mjs:182 says "the first entry that yields any usable parts", but the fixture has only one usable entry, so order is untested: mutating icims.mjs:191 to `for (const place of [...places].reverse())` leaves the suite at 38/0. A three-entry fixture with two usable entries pins it.

Unrelated, from the live probe on `careers-en-nortal.icims.com`: posting 6345 resolves to `"OM"` and 6360 to `"Kraków, PL"`, since `COUNTRY_NAMES` maps only US and CA, so a `location_filter` spelled "Oman" or "Poland" matches neither. Separate issue.

PR #3742, opened 2026-09-03
fix(update): resolve preserved files before staging so :(exclude) never reaches git add

### LINKED ISSUE #3740: Updater aborts staging when a preserved file's :(exclude) pathspec reaches git add under --literal-pathspecs
### Summary

`node update-system.mjs apply` can abort the update commit **half-done** ("files may be staged but not committed") whenever this install has a **preserved** system file — a system-layer file the user edited locally, which the updater deliberately leaves alone (#2337).

### Repro

1. Locally edit a tracked system file the updater preserves (in my case `Dockerfile` / `docker-compose.yml`, customized for a Docker + Claude Code sandbox).
2. Run an update that touches other system files.
3. The commit aborts partway; the update is left staged but uncommitted.

### Root cause

`apply()` builds its staging list as `[...updated, ...preserveSpecs]`, where `preserveSpecs` are `:(exclude)<path>` git pathspecs. That list is handed to `addPaths()`, which force-adds under `--literal-pathspecs`:

```
git --literal-pathspecs add -f -- :(exclude)Dockerfile
fatal: pathspec ':(exclude)Dockerfile' did not match any files
```

`--literal-pathspecs` disables pathspec magic, so `:(exclude)Dockerfile` is read as a literal (nonexistent) filename. One fatal pathspec takes the whole batch down.

Two individually-correct flags that are lethal together: `--literal-pathspecs` guards against filenames containing glob characters, while `:(exclude)` needs pathspec magic **on**. The exclude specs also never performed their other job during staging — a preserved file under a positive directory entry (`providers/` over a preserved `providers/acme.mjs`) is pulled back in by the directory expansion, and an appended `:(exclude)` spec the expansion ignores does not remove it.

### Fix (PR to follow)

Resolve preservation into a concrete file list **before** anything reaches `git add`: expand only the positive specs against the target tree, then subtract the preserved files — no `:(exclude)` spec reaches `addPaths`, and no preserved file is staged. Extracted as `stagingFileList()`, with a regression test covering both a standalone preserved file (the Dockerfile shape) and one under an updated directory.

Environment: macOS, running the updater inside the repo's own Docker container.

coolchigi:
Fixes #3740.

## What breaks

Run the self-updater with a locally-edited system file (I hit it with a customized `Dockerfile`) and the update commits halfway, then aborts: "files may be staged but not committed."

The updater protects your local edits by marking them with a `:(exclude)<path>` pathspec. That spec gets passed to `git add --literal-pathspecs`, which switches off pathspec magic and reads `:(exclude)Dockerfile` as a literal filename that doesn't exist:

```
fatal: pathspec ':(exclude)Dockerfile' did not match any files
```

One bad pathspec kills the whole commit.

## The fix

Work out which files to keep before staging: expand the real paths, subtract the preserved ones, pass git a plain list. New `stagingFileList()`, so no `:(exclude)` ever reaches `git add`. Also fixes a quieter version where a preserved file inside an updated directory got swept back in.

## Tests

Regression test in `updater-add-paths.test.mjs` covering a standalone preserved file and one under an updated directory. All `updater-*` suites pass.

### DIFF
--- a/tests/updater-add-paths.test.mjs
+++ b/tests/updater-add-paths.test.mjs
@@ -25,7 +25,7 @@
 import { writeFileSync, mkdirSync, rmSync, unlinkSync } from 'fs';
 import { join } from 'path';
 import { pass, fail, makeUpdaterRepo } from './helpers.mjs';
-import { gitIn, addPaths, isTracked, expandToShippedFiles } from '../update-system.mjs';
+import { gitIn, addPaths, isTracked, expandToShippedFiles, stagingFileList } from '../update-system.mjs';
 
 // Shared with updater-is-tracked.test.mjs so the git-isolation pins live in one
 // body: dropping one has to redden both suites, not leave this one quietly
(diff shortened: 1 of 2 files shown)


Scott-Emberson (commented):
Re-reviewed at head f635abaed0. Thank you for splitting this, it is now exactly the focused change it should be: two files, update-system.mjs and its owned test tests/updater-add-paths.test.mjs, with the unrelated 295-file dashboard churn gone. The fix resolves preserved files before staging: stagingFileList filters out the :(exclude) specs, expands the positive paths, then subtracts the preserved set, and it is wired on apply()'s real staging path in place of the old expand call that passed the :(exclude) token through to git add. The owned test asserts no :(exclude) token survives, that both a standalone preserved file (Dockerfile) and the harder under-a-directory case (providers/acme.mjs, which a providers/ positive spec would otherwise re-pull) are subtracted, and that running the derived list through addPaths does not throw while staging the right files and leaving the preserved ones unstaged and on disk. That subtraction, not the ignored exclude spec, is what protects the under-directory file.

The substance is update-system.mjs, a red-line file, so the decision on the updater change defers to santifer, this is a test-axis review. From that axis the owned test is sound and the no-accidental-stage guarantee for preserved files is genuinely pinned. Nothing blocking from our owned area, and the split makes it reviewable.

update-system.mjs
```diff
@@ -2413,13 +2447,16 @@ async function apply() {
...
-    // commit even though staging never touched it (#3504). Exclusion
-    // pathspecs remain in the expanded list so preserved files stay out.
-    const expandedPathsToStage = expandToShippedFiles(pathsToStage);
+    // longer ships, which can sweep a user's unstaged edit into the updater
+    // commit even though staging never touched it (#3504). stagingFileList
+    // expands the positive specs and subtracts the preserved files, so no
+    // `:(exclude)` spec reaches addPaths (where --literal-pathspecs would read
+    // it as a literal filename and abort the commit) and no preserved file is
+    // staged. preservedSet is the same Set built at the top of apply().
+    const expandedPathsToStage = stagingFileList(pathsToStage, preservedSet);
```
nikitacometa, line 2459: `apply()` is not exported, so no test reaches this line; grep finds `stagingFileList` only in the unit suite. Dropping `preservedSet` here compiles and reinstates the under-directory sweep. updater-migration-tests.mjs already source-pins `pathsToStage = [...updated, ...preserveSpecs]`; one more pattern for this call.

update-system.mjs
```diff
@@ -1565,6 +1565,40 @@ function rejectDirectories(paths, root) {
...
+ * expansion and has to be removed here, not merely appended as a spec the
+ * expansion ignores.
+ *
+ * @param {string[]} pathsToStage - positive specs + `:(exclude)<path>` specs.
+ * @param {string[]|Set<string>} [preserved] - exact preserved file paths.
+ * @param {string} [ref] - tree the directory entries resolve against.
+ * @param {{git?: Function}} [ctx] - test seam; defaults to the ROOT-bound runner.
+ * @returns {string[]} concrete file paths, preserved files removed, no exclusions.
+ */
+export function stagingFileList(pathsToStage, preserved = [], ref = 'FETCH_HEAD', ctx = {}) {
```
nikitacometa, line 1596: `ref` and `ctx` are reachable only from tests; `apply()` passes two arguments. Worth one line in the JSDoc that the defaults are the production path.

tests/updater-add-paths.test.mjs
```diff
@@ -865,3 +865,67 @@ console.log('\n🧪 Testing updater staging behavior (ignored + never-tracked pa
...
+  // (standalone) and providers/acme.mjs (under an updated directory), so both
+  // are preserved: kept on disk with the user's content and out of the commit.
+  writeFileSync(join(dir, 'AGENTS.md'), 'v2');
+  writeFileSync(join(dir, 'Dockerfile'), "the user's local sandbox edit");
+  writeFileSync(join(dir, 'providers/core.mjs'), 'v2');
+  writeFileSync(join(dir, 'providers/acme.mjs'), "the user's local edit");
+
+  const pathsToStage = ['AGENTS.md', 'providers/', ':(exclude)Dockerfile', ':(exclude)providers/acme.mjs'];
+  const preserved = ['Dockerfile', 'providers/acme.mjs'];
+  const files = stagingFileList(pathsToStage, preserved, 'HEAD', ctx);
```
nikitacometa, line 899: The fixture passes `'HEAD'`; `apply()` uses the `'FETCH_HEAD'` default, so that half of the signature is only exercised in test form.

nikitacometa (approved):
Verified at head f635abae. Probe on a throwaway repo with the #3740/#3779 shape: on main the staging list is `["modes/pdf.md", ":(exclude)AGENTS.md"]` and addPaths dies with `fatal: pathspec ':(exclude)AGENTS.md' did not match any files`; the same probe at this head returns `["modes/pdf.md"]` and stages clean. A compiling mutant with `stagingFileList` reverted to a pass-through reddens exactly the three new assertions (39 passed, 3 failed). `node test-all.mjs --only updater` at head: 173 passed, 0 failed.

Two notes, neither blocking. update-system.mjs:2510 still filters `:(exclude)` out of `expandedPathsToStage`; that list can no longer contain one, so it is dead code asserting an invariant this PR removed. And `apply()` is not exported, so nothing pins the call at update-system.mjs:2459; updater-migration-tests.mjs source-pins its neighbours, and dropping the `preservedSet` argument there compiles and reinstates the under-directory sweep. One pattern closes it.

For the record: #3781 also says `Fixes #3779`, but at its head the identical fatal still reproduces; this is the PR that closes it. The scoped commit line itself stays at 30,962 chars of argv against the 32,767 Windows cap (measured with this PR's SYSTEM_PATHS, 1080 files), which neither PR changes.

PR #3747, opened 2026-09-03
feat(scan): one machine-findable marker line above every scan summary (#3560)

### LINKED ISSUE #3560: Scan output has no stable summary boundary — every consumer duplicates five banner literals
### Problem

Every scanner prints a human-readable summary at the end of a run, but there is no stable, machine-findable boundary where that summary begins. Anything that wants only the summary — a wrapper script, a status line, a log slicer, or an agent reading a backgrounded run's output — has to hardcode each scanner's banner string, and the banners are neither uniform nor safe to match loosely.

What exists today (line numbers against `main` as of this edit):

| script | summary boundary |
|---|---|
| `scan.mjs:2825` | `Portal Scan — ${date}` |
| `scan-ats-full.mjs:978` | `Reverse ATS Scan — ${date}` |
| `scan-interamt.mjs:334` | `Interamt Scan — ${date}` |
| `scan-hn.mjs:118` | none — ends with `🎉 Success: N offers added.` |
| wrapper scripts around the above | none of their own |

Three sharp edges:

1. **Case-sensitivity trap.** `scan-ats-full.mjs:676` prints `Reverse ATS scan — ...` at the *start* of a run (lowercase `s`), and `:978` prints `Reverse ATS Scan — ...` for the summary (capital `S`). A case-insensitive match on the obvious string swallows the entire log — for a full sweep that is thousands of progress lines. Nothing in the code signals that the two are meant to be distinguished by case alone. A third copy of the summary literal is written into the `--md-out` digest at `:1035`, so the string is maintained in three places.
2. **`scan-hn.mjs` has no banner at all.** Matches print inline as `  ✅ Match: {company}` and the run ends with one `🎉 Success:` line, so a banner-based slice returns zero lines for it.
3. **Nothing marks the boundary as an interface.** These are presentation strings. Rewording one is a cosmetic change upstream and a silent breakage downstream — the slice quietly returns the wrong region, or nothing, and the caller has no way to tell that from "the scan found nothing".

### Why it matters

I originally wrote that the counters and warnings exist only in stdout. That is too broad, and the accurate version is more useful: **coverage exists, but no two scanners have the same amount of it.**

| scanner | run row in `scan-runs.tsv` | `--json` mode | left in stdout only |
|---|---|---|---|
| `scan.mjs` | yes (`:3017`) | no | cross-listing warnings (`:2774` → printed `:2872`), agent/WebSearch handoff list (`:2496` → printed `:2913`) |
| `scan-ats-full.mjs` | no | yes (`:1060`) | every counter it prints, when run without `--json` |
| `scan-interamt.mjs` | no | no | everything, always |
| `scan-hn.mjs` | no | no | everything, always — and no banner to slice on either |

`scan.mjs` is the only scanner that appends to `data/scan-runs.tsv` (`SCAN_RUNS_HEADER`, 19 columns, `:2046`); nothing else in the tree writes that file. Trust is durable per-offer (`trust_score`/`trust_flags` in `scan-history.tsv`, `:1977`) and target failures land in `portal-health.tsv` (`:2111`) — for that one scanner. The other three have neither, so a consumer that wants their counters has no option but the log, which brings it straight back to the banner literals above.

**A second, smaller defect in the same family: a run's own rows do not carry one identity.** `portal-health.tsv` rows are stamped from `nowStr`, computed once at `:2929`. The `scan-runs.tsv` row takes a *second*, independent `new Date().toISOString()` at `:3018`. The only statement between them that can consume time is `await appendPortalHealth(healthRecords)` at `:3016` — so the two timestamps for one run differ by however long that write took, and a consumer joining the files on the timestamp gets zero rows and reads it as "no unhealthy portals this run".

Provenance, since this one is latent rather than observed upstream: no consumer in the tree performs that join today — `stats.mjs` aggregates `scan-runs.tsv` by date (`:521`, `timestamp.slice(0, 10)`) and reads `portal-health.tsv` as whole-history streaks, so neither hits it. I found it writing a local consumer against these two files, and measured the drift twice on my own runs: 14 ms and 11 ms (`2026-08-31T05:57:04.763Z` in `scan-runs.tsv` vs `...04.752Z` in `portal-health.tsv`). The mechanism, though, is readable from upstream alone — two `new Date()` calls with a file write between them.

### Proposal

All additive; no existing consumer changes behaviour. Smallest and most certain first.

1. **Reuse `nowStr` for the run row.** Pass `timestamp: nowStr` at the `appendScanRunSummary` call site (`:3018`) instead of calling `new Date().toISOString()` again. One line, no schema change, and one run's rows across the two files become joinable exactly. This is the whole of the demonstrated defect above.
2. **Give `scan-ats-full.mjs`, `scan-interamt.mjs` and `scan-hn.mjs` a `data/scan-runs.tsv` row.** Follows a convention `scan.mjs` already established, needs no new interface, and removes most of the reason to read a log at all. `stats.mjs` picks up three more scanners for free.
3. **Extend the `--json` *convention* — not the schema — to the scanners that lack it.** `scan-ats-full.mjs:1060` already establishes the useful half: under `--json`, stdout carries exactly one JSON object and nothing else, with progress on stderr. The payload itself is necessarily per-scanner (ats-full reports `datasetStatus`, `capHit`, `stoppedByOutage`; `scan.mjs` would report its 19 filter counters plus the two warnings above), so the shared part is the stdout guarantee, not the field names. `scan-interamt.mjs` and `scan-hn.mjs` are small enough to be one change; `scan.mjs` is 3000+ lines and probably wants its own PR.
4. **A run id, if (2) lands.** Once several scanners append to the same files from separate processes, reusing a local variable stops being enough and a per-run id written into every file a run touches is the general form of (1). Not worth doing on its own today.
5. **Fallback if none of the above is wanted:** one machine-readable sentinel printed immediately before the summary, identical across every scanner (e.g. `::summary::`), plus an exported `SUMMARY_BANNER` constant so the human title can be reworded without breaking a consumer — and the `━━━` + `{Name} Scan — {date}` block for `scan-hn.mjs`, which is the only one without it.

Related in spirit rather than in code: on #3438 you noted that a change here should be additive by default rather than silently altering existing behaviour. Everything above is written to that constraint — (1) is invisible to any current reader, and (2), (3) and (5) only add output that nothing consumes yet.

Happy to open a PR for any of these, in whatever order and split you prefer.

ronanime-arch:
Closes #3560, in the scope you named there: one shared marker line emitted by every scanner right before its summary, with the existing banners left as they are underneath.

### What changed

- **New `lib/scan-summary-marker.mjs`** — `SCAN_SUMMARY_MARKER` (`::career-ops:scan-summary::`) plus `scanSummaryHeaderLines()` / `printScanSummaryHeader()`.
- **`scan.mjs`, `scan-ats-full.mjs`, `scan-interamt.mjs`** print the marker above their existing banner. Banner text, rule width, and the blank line above it are unchanged.
- **`scan-hn.mjs`** gets the banner block it never had, printed unconditionally — see below.
- `lib/scan-summary-marker.mjs` registered in `SYSTEM_PATHS`.

### The matching contract

Consumers match the **whole trimmed line** against the constant — equality, not a prefix. Nothing is appended to the token (no scanner id, no counts), so adding one later would be a knowingly breaking change rather than a formatting tweak. That is stated in the module header and asserted in the test.

### `--json` stdout is untouched

`scan-ats-full.mjs` routes human output to stderr under `--json` (`log`, `:691`) so stdout carries exactly one JSON object. The marker goes through that same `log`, so under `--json` it lands on stderr with the summary it belongs to. The JSON emit path (`:1152`) is unchanged.

### The one behaviour change, flagged deliberately: `scan-hn.mjs`

Today `scan-hn.mjs` prints nothing at all when nothing matched — the run just ends after `⚠️ No AI key...`. A consumer cannot tell that from a run that died at the fetch, and a marker printed only on the happy path would carry the same ambiguity, so the block is unconditional and reports `New offers: 0`. This is the one place where the PR changes what an existing scanner prints rather than only adding a line. Item 5 of the issue named the hn block, but it is the part of this PR most worth a second look, so I am not burying it.

The `🎉 Success:` line is unchanged and still prints only when something was added.

### Left alone deliberately

`scan-ats-full.mjs:1128` writes `# Reverse ATS Scan — ${date}` into the `--md-out` digest. That is the third copy of the literal the issue called out, but it is a markdown heading in a file, not a boundary in a log stream, so it gets no marker.

### Tests

`tests/scan-summary-marker.test.mjs` (auto-discovered, no registration). It asserts the marker is alone on its line, and that the printed block is byte-identical to what the scanners emitted before the marker existed — a golden string rather than a claim in this description.

Full suite locally on Windows/Node 22: **8084 passed, 0 failed, 18 warnings**; a clean checkout of `main` at 398edb1 gives 8081 passed and the same 18 warnings, so nothing new is warned about.

### Not in this PR

Items 1-4 of the issue (reusing `nowStr` for the run row, `scan-runs.tsv` rows for the other three scanners, extending the `--json` convention, a run id) are untouched here — happy to send any of them separately, in whatever order suits you.

### DIFF
--- a/lib/scan-summary-marker.mjs
+++ b/lib/scan-summary-marker.mjs
@@ -1,0 +1,61 @@
+/**
+ * scan-summary-marker.mjs — the machine-findable boundary where a scan's
+ * summary begins (#3560).
+ *
+ * Every scanner ends a run with a human-readable summary, but until now the
+ * only way to find where that summary starts was to hardcode the scanner's
+ * banner text. Those banners are presentation strings: rewording one is a
+ * cosmetic change here and a silent breakage in whatever slices the log —
+ * the slice quietly returns the wrong region, or nothing, and the caller
+ * cannot tell that from "the scan found nothing".
+ *
+ * They are also not safe to match loosely. `scan-ats-full.mjs` prints
+ * `Reverse ATS scan — ...` at the START of a run and `Reverse ATS Scan — ...`
+ * for the summary: a case-insensitive match on the obvious string swallows
+ * the whole log, which for a full sweep is thousands of progress lines.
+ *
+ * So the banners stay exactly as they are, and one marker line goes above
+ * them. The marker is the interface; the banner is decoration.
+ *
+ * MATCHING CONTRACT: consumers match the whole trimmed line against
+ * SCAN_SUMMARY_MARKER — equality, not a prefix. Nothing is appended to it
+ * (no scanner id, no counts), and adding anything later would be a breaking
+ * change to be made knowingly, not a formatting tweak.
+ *
+ * Stream: the marker follows its scanner's summary. `scan-ats-full.mjs`
+ * routes every human line to stderr under `--json` so that stdout carries
+ * exactly one JSON object; passing that scanner's `log` as `emit` keeps the
+ * marker on stderr with the summary it belongs to, and stdout uncontaminated.
+ */
+
+/**
+ * The line printed immediately above a scan summary. Deliberately not a word
+ * any scanner would print for another reason.
+ */
+export const SCAN_SUMMARY_MARKER = '::career-ops:scan-summary::';
+
+/**
+ * The marker plus the banner block that follows it.
+ *
+ * @param {string} title - Human banner title, e.g. 'Portal Scan'.
+ * @param {string} date - Run date, already formatted (YYYY-MM-DD).
+ * @returns {string[]} Lines to print in order. The leading '' preserves the
+ *   blank line the scanners printed before their banner.
+ */
+export function scanSummaryHeaderLines(title, date) {
+  const rule = '━'.repeat(45);
+  return ['', SCAN_SUMMARY_MARKER, rule, `${title} — ${date}`, rule];
+}
+
+/**
+ * Print the marker and banner block.
+ *
+ * @param {string} title - Human banner title, e.g. 'Portal Scan'.
+ * @param {string} date - Run date, already formatted (YYYY-MM-DD).
+ * @param {(line: string) => void} [emit] - Where to write; defaults to
+ *   console.log. Pass the scanner's own logger when it redirects output
+ *   (see the stream note above).
+ */
+export function printScanSummaryHeader(title, date, emit = console.log) {
+  for (const line of scanSummaryHeaderLines(title, date)) emit(line);
+}
--- a/scan-hn.mjs
+++ b/scan-hn.mjs
@@ -22,6 +22,7 @@ import { localToday } from './lib/local-today.mjs';
 // Import the deterministic provider
 import hnProvider from './providers/hackernews.mjs';
 import { isMainModule } from './lib/is-main-module.mjs';
+import { printScanSummaryHeader } from './lib/scan-summary-marker.mjs';
 
 // ── Configuration ────────────────────────────────────────────────────
 // Imported from scan.mjs so it honors CAREER_OPS_PORTALS and the data root (#3510).
@@ -115,8 +116,15 @@ async function main() {
   if (newOffers.length > 0) {
     await appendToPipeline(newOffers);
     await appendToScanHistory(newOffers, localToday(), 'added');
-    console.log(`\n🎉 Success: ${newOffers.length} offers added.`);
   }
+
+  // Printed on every run, including the zero-match one. This scanner used to
+  // end in silence when nothing matched, which reads identically to a run that
+  // died at the fetch — the summary is what tells those apart (#3560).
+  printScanSummaryHeader('HN Scan', localToday());
+  console.log(`Postings fetched:   ${rawJobs.length}`);
+  console.log(`New offers:         ${newOffers.length}`);
+  if (newOffers.length > 0) console.log(`\n🎉 Success: ${newOffers.length} offers added.`);
 }
 
 if (isMainModule(import.meta.url)) {
--- a/scan-ats-full.mjs
+++ b/scan-ats-full.mjs
@@ -50,6 +50,7 @@ import workday from './providers/workday.mjs';
 import icims from './providers/icims.mjs';
 import { buildTitleFilter, buildLocationFilter, buildContentFilter, matchedTitleKeywords, loadSeenUrls, normalizeUrlForDedup, appendToPipeline, appendToScanHistory, loadBlacklist, parseSinceDays, PORTALS_PATH, PIPELINE_PATH } from './scan.mjs';
 import { localToday } from './lib/local-today.mjs';
+import { printScanSummaryHeader } from './lib/scan-summary-marker.mjs';
 import { SEED_SOURCES, toPortalEntry } from './seeds/vc-portfolios.mjs';
 import { normalizeCompany } from './tracker-utils.mjs';
 import { validateFlags } from './lib/cli-flags.mjs';
@@ -1066,9 +1067,7 @@ async function main() {
   if (offers.length && opts.liveness) offers = await filterLive(offers);
   offers.sort((a, b) => (b.postedAt || 0) - (a.postedAt || 0));
 
-  log(`\n${'━'.repeat(45)}`);
-  log(`Reverse ATS Scan — ${date}`);
-  log(`${'━'.repeat(45)}`);
+  printScanSummaryHeader('Reverse ATS Scan', date, log);
   log(`Companies scanned:  ${totalCompaniesScanned}${capHit ? ` of ${totalCompaniesAvailable} (capped)` : ''}`);
   log(`Unreachable boards: ${totalErrors}`);
   if (cappedBoards) log(`Page-capped boards: ${cappedBoards} (partial coverage — later postings not scanned)`);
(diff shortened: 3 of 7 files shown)


Scott-Emberson (commented):
Reviewed at head a6d3800d50. The owned test tests/scan-summary-marker.test.mjs pins the marker precisely: a full-array deepEqual asserts the lines are exactly empty, the marker, the rule, the title-and-date line, the rule, so the marker sits at index one immediately above the banner exactly once and a missing, duplicated, or misplaced marker reddens, and a second assertion confirms nothing is appended. There is also a real wiring guard that greps each scanner (scan.mjs, scan-ats-full.mjs, scan-interamt.mjs, scan-hn.mjs) for both the import and the printScanSummaryHeader call, so a dropped call in any one scanner is caught rather than smoke-passing. One minor note, not a gap: the marker is compared against the imported constant rather than the literal string, so a rename of the constant would not redden, though the banner strings are literal goldens.

On the one red check: the Upgrade regression gate failure is system-layer, not this PR. The --local-paths leg of upgrade-tests.mjs calls apply without --confirm, which now bails at the confirmation gate that landed on main in #2866, and this branch predates that, so the leg fails before reaching the local-paths evaluation. This PR's only update-system.mjs change is one benign SYSTEM_PATHS registry line for the new module, it does not touch upgrade-tests.mjs or the confirm logic, and the three test legs that run the owned test are green. A rebase onto current main clears the gate.

The substance is scan.mjs summary output, outside our owned paths, so this is a test-axis review: the owned test is sound and the lone red is a stale system-layer gate. Nothing blocking from the tests/ side.

Scott-Emberson (commented):
Re-reviewed at head d1ba6aa30d. This closes the nit I raised: a new test asserts the marker token as a literal, SCAN_SUMMARY_MARKER === '::career-ops:scan-summary::' and the header's second line equals that same literal, so a rename or re-spell of the exported constant now reddens (before, the comparison was against the constant itself and would not have caught a rename). The change is purely additive (nine lines, no deletions) and the earlier assertions are intact: the full-array deepEqual pinning the marker at index one above the banner, the one-call-per-line and exactly-once guards, and the per-scanner import-and-printScanSummaryHeader wiring guard.

The one red check is the system-layer Upgrade regression gate (the --local-paths leg trips on the updater confirm change that this branch predates), not this PR, and a rebase onto current main clears it. The substance is scan.mjs and the lib marker module, outside our owned paths, so this stays a test-axis review: the owned test is sound and the nit is closed. Nothing blocking from the tests/ side.

lib/scan-summary-marker.mjs
```diff
@@ -0,0 +1,61 @@
...
+ * the whole log, which for a full sweep is thousands of progress lines.
+ *
+ * So the banners stay exactly as they are, and one marker line goes above
+ * them. The marker is the interface; the banner is decoration.
+ *
+ * MATCHING CONTRACT: consumers match the whole trimmed line against
+ * SCAN_SUMMARY_MARKER — equality, not a prefix. Nothing is appended to it
+ * (no scanner id, no counts), and adding anything later would be a breaking
+ * change to be made knowingly, not a formatting tweak.
+ *
```
nikitacometa, line 24: Covers a successful run. Add that a run which dies before its summary emits no marker at all (scan-hn.mjs:130, scan.mjs:3267), so a consumer reads absence as meaningful.

scan-hn.mjs
```diff
@@ -115,8 +116,15 @@ async function main() {
   if (newOffers.length > 0) {
     await appendToPipeline(newOffers);
     await appendToScanHistory(newOffers, localToday(), 'added');
-    console.log(`\n🎉 Success: ${newOffers.length} offers added.`);
   }
+
+  // Printed on every run, including the zero-match one. This scanner used to
+  // end in silence when nothing matched, which reads identically to a run that
+  // died at the fetch — the summary is what tells those apart (#3560).
```
nikitacometa, line 123: The description says the "No AI key" run "just ends"; it is an `else` branch at 99, not an early return, so it already reached the summary. Narrower change than the text suggests.

nikitacometa (approved):
Probed at d1ba6aa3. The marker prints on the normal, `--dry-run` and zero-match paths of scan.mjs against the local fixture board (count 1 each); origin/main on the same lane prints 0. `scan-runs.tsv` is unaffected, `stats.mjs --summary` still reports `Runs: 1 recorded`, and `--json` stdout stays clean since the marker rides the stderr logger. scan.mjs `main()` has no body-level early return between arg validation and 3012, so there is no summary path without the marker, and no plugin prints a banner, so there is no fifth emit site. The wiring guard is real: replacing the call in scan-interamt.mjs:348 with a plain `console.log` leaves the file compiling and reddens exactly that test (5 pass, 1 fail). CI is green now, including the Upgrade regression gate the earlier reviews expected to need a rebase.

Only gap is documentation. Nothing outside `lib/` and `tests/` mentions the token: docs/AUTOMATION.md:45 is the cron recipe that accumulates many runs in one log, the case the marker exists for, and a reader there still slices on `Portal Scan —`. modes/scan.md:404-408 shows the pre-marker block (and was already wrong about the rule placement). And the module header states the contract for a successful run only; a run that dies before its summary emits no marker (scan-hn.mjs:130, scan.mjs:3267), which is the right behaviour and worth one sentence.

PR #3748, opened 2026-09-03
feat(providers): OCC Mundial — Mexico's dominant job board

VictorMeneses77:
## What

Adds a zero-auth scanner provider for **[occ.com.mx](https://www.occ.com.mx)** (OCC Mundial), the dominant general job board in Mexico, plus 18 offline tests.

Per CONTRIBUTING, new zero-auth scanner providers go straight to a PR without an issue first. Happy to open one if you'd rather discuss direction.

## Why

A large share of Mexican employers never appear on Greenhouse, Lever, Ashby, Workday or SuccessFactors. Mid-size industrials and domestic groups post to OCC instead, so the scanner has a structural hole in Mexico that no amount of ATS probing closes.

I hit this concretely: I resolved 30+ ATS tenants by hand (Workday, SuccessFactors, SmartRecruiters) to widen Mexico coverage, and the Mexican mid-market still did not show up. Checking a few directly — Xignux, Deacero — they publish on OCC, not on any international ATS.

A live scan across 8 keywords × 6 states returned **1,136 postings, 95 of which survived `title_filter`, from 69 companies no ATS sweep had ever returned**. A sample:

| Company | Role | Location |
|---|---|---|
| Tecnoap | Ingeniero en Automatización | San Nicolás de los Garza, NL |
| AWL Automation | Service PLC Engineer | Colón, Querétaro |
| IMVISION | Ingeniero de Aplicaciones — Visión Artificial | Querétaro |
| SSS CARRIER | Técnico de mantenimiento PLC (Allen Bradley) | Santa Catarina, NL |
| Kimberly Clark de México | Ing. procesos Mecatrónico/Robótico | Cuautitlán Izcalli |

## How

There is no public JSON API — `/api/*` answers `403`. The search pages are server-rendered, so the provider parses the job cards out of the markup with the same tiny-tag-extractor approach already used by `weworkremotely.mjs` and `successfactors.mjs`. **No new dependencies.**

Config shape:

```yaml
  - name: OCC Mundial
    provider: occ
    queries: [automatizacion, robotica, mecatronica]
    states: [nuevo-leon, jalisco]   # optional; omit for nationwide
    max_pages: 3                    # optional, default 3, hard cap 10
```

## Two behaviours worth flagging

**1. Pagination is a slug suffix, not a query parameter.** `?page=2`, `?pagina=2`, `?p=2`, `?start=20`, `?offset=20` and `/2/` all return `HTTP 200` with a full page of cards — from page 1. The failure is completely silent, so a naive implementation re-reads page 1 forever and looks like it is working. The only shape that advances is:

```
https://www.occ.com.mx/empleos/de-{keyword}-pagina-{N}/
```

Because OCC also answers an out-of-range page with page 1 rather than an empty result, the fetch loop additionally stops as soon as a page yields no new ids. Both are covered by tests.

**2. `"Empresa confidencial"` is OCC's own placeholder, not an employer name.** It is normalized to the locale-invariant `?` marker (#1596) so the Spanish string never reaches the tracker or a report.

## Tests

`tests/providers/occ.test.mjs` — 18 assertions, **all offline** against a fixture of the real card markup, so CI never depends on occ.com.mx being reachable.

Covers: provider shape · page-1 URL has no suffix · pagination uses the `-pagina-N` slug · accent folding and the `en-{state}` segment · URLs cannot be steered off the `occ.com.mx` origin · entity decoding in title and location · posting URL built from `data-id` · malformed cards dropped · `"Empresa confidencial"` → `?` · empty input → `[]` · dedup when page 1 is re-served · loop stops on no new ids · normalized output shape · transport errors do not kill the board · `max_pages` capped at 10.

```
$ node test-all.mjs --only providers/occ
📊 Results: 18 passed, 0 failed, 0 warnings
```

## Follow-up, not in this PR

Computrabajo is the other portal with the same footprint in Mexico. Same argument, separate provider — happy to do it if this shape looks right to you.

https://claude.ai/code/session_01Y8Udned64LVbEUQZNW8obB

## User impact

- Users can scan `occ.com.mx` with the new `occ` provider: `providers/occ.mjs:1`.
- Configuration must include `queries: [...]`. Default queries are removed: `providers/occ.mjs:1`.
- Users can filter by Mexican state and limit pagination.
- The provider deduplicates postings and normalizes confidential or invalid employers to `?`.
- Cloudflare challenges and complete transport outages return errors instead of empty results.
- Requests use retries and 750 ms pacing.

## Files touched

- `providers/occ.mjs:1`: Adds OCC search, parsing, validation, pagination, and failure handling.
- `tests/providers/occ.test.mjs:1`: Adds offline provider tests.
- `docs/SUPPORTED_JOB_BOARDS.md:1`: Documents OCC requirements, Cloudflare behavior, pacing, and permitted paths.
- `templates/portals.example.yml:1`: Adds an OCC configuration example.

`AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, and `.github/` are not touched by this change.


Scott-Emberson (commented):
Reviewed at head cc3ceec43e. In our gated area (providers/ plus tests/providers/). The OCC Mundial fetch surface is well constrained despite OCC being a hosted board: the origin is a fixed HTTPS constant (https://www.occ.com.mx), never derived from config, so no host surface is exposed, queries and states only feed slugify (stripped to [a-z0-9-] with accent folding) into path segments, and fetches pass redirect:'error'. The test asserts spoof states (../../evil and an https://evil.example value) stay on the fixed origin, and pins real parsing concretely: exact detail URLs from the data-id, company and location extraction, Empresa confidencial mapping to the ? confidential marker, the pagination slug shape (-pagina-3, not a query param), id-dedup on a re-served first page, early stop on zero-fresh, and the max_pages hard cap.

On the one red check: the Upgrade regression gate failure is system-layer, not this PR. It is the upgrade-tests.mjs --local-paths leg, which fails because this branch predates the recent updater and confirmation changes on main (the leg trips on an offending entry unrelated to occ), and this PR touches only providers/occ.mjs and tests/providers/occ.test.mjs, not update-system.mjs or upgrade-tests.mjs. The three test legs that actually run the provider suite (ubuntu, macos, windows) are all green. A rebase onto current main (which has the gate passing again) clears it.

Owned test verified sound, no SSRF surface, the lone red is a stale system-layer gate. Rebase to current main and I will re-verify and clear it to approve.

luochen211 (commented):
Reviewed at head cc3ceec and replayed it cleanly onto current main (5c8b02a); all 18 submitted OCC assertions pass. I also exercised the provider with `ctx.maxPages = 1`: an entry with `max_pages: 10` still made 10 list requests, so the current tests miss the health-probe contract.

The main issues I found:

1. `fetch()` ignores `ctx.maxPages`, and its broad catch also swallows every `ctx.fetchText` rejection while probing. In `verify-portals`, that can swallow `ProbePageBudgetReached` and continue through other query/state combinations instead of returning the sentinel to the probe. Please cap the walk with the context hint and rethrow fetch errors unwrapped when probing.

2. Pagination uses bare `ctx.fetchText` back-to-back. The provider contract calls for the shared bounded retry helper and an inter-page `sleep(ms, ctx)`; this board can otherwise receive up to queries × states × pages requests in a burst.

3. A complete transport outage is converted to `[]` (the submitted test explicitly blesses that), and arbitrary card-less HTML is also converted to `[]`. Both make a broken source or markup change look like a healthy empty board. Keep recall-first partial results after at least one successful page if desired, but throw when every request fails, and distinguish the site's real empty-board shape from a first-page selector miss.

4. The new provider is absent from `docs/SUPPORTED_JOB_BOARDS.md` and `templates/portals.example.yml`, both required by the provider checklist.

5. If company extraction fails, line 158 substitutes the board entry name (normally "OCC Mundial") as the employer. For an aggregator, that misattributes the posting to the board and conflicts with the source-indexing requirement that listings remain employer-attributed. Drop that malformed row or preserve an explicit unknown marker instead of naming OCC as the employer.

The fixed-origin and redirect-refusal work is sound; these are lifecycle/coverage failures around it rather than an SSRF concern.

providers/occ.mjs
```diff
@@ -0,0 +1,169 @@
...
+          // signal. Without this the loop would re-add page 1 max_pages times.
+          let fresh = 0;
+          for (const c of cards) {
+            if (seen.has(c.id)) continue;
+            seen.add(c.id);
+            fresh++;
+            jobs.push({
+              title: c.title,
+              url: c.url,
+              company: c.company || (entry.name || 'OCC'),
```
nikitacometa: Naming the board as the employer breaks rule 1. Probe: a card with no `h-[21px]` row yields `"company":"OCC Mundial"`. Use `'?'` (already used at 109) or drop the row.

providers/occ.mjs
```diff
@@ -0,0 +1,169 @@
...
+    const jobs = [];
+
+    for (const query of queries) {
+      for (const state of states) {
+        for (let page = 1; page <= maxPages; page++) {
+          const url = buildSearchUrl(query, state, page);
+          let html;
+          try {
+            html = await ctx.fetchText(url, { redirect: 'error' });
+          } catch (err) {
```
nikitacometa, line 221: This catch swallows `ProbePageBudgetReached` and turns a site-wide 403 into `[]`. Rethrow unwrapped while `ctx.maxPages` is set, and throw when every request for an entry failed.

providers/occ.mjs
```diff
@@ -0,0 +1,169 @@
...
+/** @type {Provider} */
+export default {
+  id: 'occ',
+
+  async fetch(entry, ctx) {
+    const queries = Array.isArray(entry.queries) && entry.queries.length
+      ? entry.queries : DEFAULT_QUERIES;
+    const states = Array.isArray(entry.states) && entry.states.length
+      ? entry.states : [null];
+    const maxPages = Math.min(
```
nikitacometa: `ctx.maxPages` is not consulted. Measured: `maxPages:1` plus `max_pages:10` gives 10 requests.

providers/occ.mjs
```diff
@@ -0,0 +1,169 @@
...
+    const seen = new Set();
+    const jobs = [];
+
+    for (const query of queries) {
+      for (const state of states) {
+        for (let page = 1; page <= maxPages; page++) {
+          const url = buildSearchUrl(query, state, page);
+          let html;
+          try {
+            html = await ctx.fetchText(url, { redirect: 'error' });
```
nikitacometa: `fetchTextWithRetry` and `sleep` from `_http.mjs` (166, 296); pages past the first need pacing, and a 429/503 gets no retry here.

providers/occ.mjs
```diff
@@ -0,0 +1,169 @@
...
+//     provider: occ
+//     queries: ["automatizacion", "robotica", "mecatronica"]
+//     states: ["nuevo-leon", "jalisco"]      # optional; omit for nationwide
+//     max_pages: 3                            # optional, default 3, hard cap 10
+
+import { decodeEntities } from './_html-entities.mjs';
+
+const ORIGIN = 'https://www.occ.com.mx';
+const DEFAULT_QUERIES = ['automatizacion', 'robotica', 'mecatronica', 'sistemas-embebidos', 'control'];
+const DEFAULT_MAX_PAGES = 3;
```
nikitacometa, line 73: 3 pages by default, 10 hard cap, over a relevance-sorted view with `is-highlighted` cards. How does this meet rule 3's complete-inventory requirement?

tests/providers/occ.test.mjs
```diff
@@ -0,0 +1,151 @@
...
+
+  if (jobs.every(j => j.title && j.url && typeof j.company === 'string' && typeof j.location === 'string')) {
+    pass('fetch() returns the normalized {title, url, company, location} shape');
+  } else {
+    fail(`fetch() shape = ${JSON.stringify(jobs)}`);
+  }
+
+  // A transport failure on one keyword must not take down the whole board.
+  const failingCtx = { async fetchText() { throw new Error('boom'); } };
+  const resilient = await occ.fetch({ name: 'OCC', queries: ['a', 'b'], max_pages: 2 }, failingCtx);
```
nikitacometa: This asserts a total transport outage returns `[]`. Live, occ.com.mx answers a Node fetch with a Cloudflare 403 challenge, so this blesses the silent-failure path.

nikitacometa (changes requested):
Reproduced luochen211's findings at cc3ceec4 with probes. `ctx.maxPages=1` with `max_pages:10` still issues 10 requests, and a ctx throwing `ProbePageBudgetReached` returns `[]` after 4 attempts, so verify-portals loses the sentinel (occ.mjs:125-143). A card with no company row emits `{"company":"OCC Mundial"}` (occ.mjs:158), which breaks source-policy rule 1; the `?` marker already exists at 109. No inter-page `sleep` and no `fetchTextWithRetry` (139), so a default run is 15 unpaced requests and a configured one up to 480. `SUPPORTED_JOB_BOARDS.md` and `templates/portals.example.yml` have no occ entry, and with no `detect()` the explicit-`provider:` list needs it too. The suite itself has teeth (the confidential-company mutant reddens the named test).

Separately, measured 2026-09-04: `https://www.occ.com.mx/empleos/de-automatizacion/` returns HTTP 403 with `cf-mitigated: challenge` and "Just a moment..." from two independent non-Mexican egresses, and `/robots.txt` returns the same challenge, so the crawl policy is not readable either. Node's `fetch` cannot clear a JS challenge, and occ.mjs:140-142 turns every rejection into `[]`, which reads as a healthy empty board; tests/providers/occ.test.mjs:133 pins that silent path. Please say whether the challenge is geo-scoped and how the 1,136-posting scan was run, since that decides whether the source is indexable at all.

This is a job board, so the Source Indexing Policy applies in full: the employer-attributed share could not be measured behind the 403, and rule 3 is open (`DEFAULT_MAX_PAGES = 3`, hard cap 10, over a relevance-sorted view whose own fixture card is `is-highlighted`). A source proposal issue with those numbers is the usual route. The CI red is #3755 on main; a rebase clears it.

santifer (changes requested):
Welcome to career-ops, @VictorMeneses77, and thanks for the thorough header comment: the pagination trap (`-pagina-N` on the keyword segment, every other shape silently re-serving page 1) is exactly the kind of note that saves the next person a day. Two things before I merge:

- Providers ship with their docs: please add OCC to `docs/SUPPORTED_JOB_BOARDS.md` and a commented example entry to `templates/portals.example.yml`. That is how `builtin` and `garena` landed.
- `DEFAULT_QUERIES` (automatizacion, robotica, mecatronica, …) is one profile's search, not a board default. Do what `wttj` does: require `queries` and throw with a clear message when it is missing, so nobody scans robotics jobs by accident.

Per CONTRIBUTING, zero-auth providers do not need an issue first, so no need to open one. Happy to merge once these are in.

Scott-Emberson (commented):
Verified the owned test `tests/providers/occ.test.mjs` and the provider. The shipped code is SSRF-safe by construction: `ORIGIN = 'https://www.occ.com.mx'` is a hardcoded prefix and every request URL is `ORIGIN + '/empleos/' + slugify(...)`, with all user input (queries, states) folded through `slugify()` to `[a-z0-9-]`, so the host is never derived from input and there is no userinfo or suffix bypass surface. The shared transport also passes `redirect` to fetch and the process-wide DNS/IP guard (#3096) validates resolved addresses on every request. The owned test pins real behavior, not a smoke test: `parseCards` drops missing-data-id and empty-title cards, decodes entities, builds the posting URL from the data id, normalizes "Empresa confidencial" to `?`, dedups by id and stops on the first no-new-ids page, throws on a total outage, tolerates partial failure, honors `ctx.maxPages`, propagates the probe sentinel unwrapped, and caps `max_pages` at 10. `slugify`'s accent-fold and pagination output are pinned, so removing it reddens the test.

Two owned-test quality notes, both non-blocking (the code is SSRF-safe structurally and at the transport layer, so neither is a hole):

1. The "cannot be steered off origin" assertion is tautological. It checks `new URL(u).origin === 'https://www.occ.com.mx'` for hostile inputs, but because ORIGIN is a fixed prefix that assertion passes even with `slugify` removed (a raw `.../en-https://evil.example/` still parses its origin as occ.com.mx). It gives false confidence as an SSRF guard. Reword it to exercise slugify's neutralization directly (assert the hostile input is slugified into the path segment), or drop the SSRF framing.
2. `redirect:'error'` is set in the provider but not pinned by this owned test, because the stub `fetchText(url)` ignores opts. Removing `redirect:'error'` from occ.mjs reddens nothing here (real enforcement and its regression test live in the transport layer). A stub that records opts and asserts `redirect === 'error'` would close it.

The provider is SSRF-safe and the substantive behavior is pinned, so this is GATE-tier and gate-ready, CI green and not behind. Holding formal approval until Scott gives the go-ahead; surfacing it as ready-to-approve, with the two test-quality notes above worth folding in either before or right after merge.

Scott-Emberson (changes requested):
Thanks for the `isEmptyResultPage` fix (a 200-served challenge no longer counting as an answered page is a real improvement), and for tightening the page-cap assertion to an exact `=== 10`. Those are good.

I am requesting changes on the two owned-test gaps I raised last time, because they are still open and both are assertions that claim to guard something they do not, so a regression would pass silently. The provider code is SSRF-safe (hardcoded `ORIGIN`, `slugify` to `[a-z0-9-]`, `redirect:'error'` present), so neither is a live hole; these are owned-test quality defects that should not ride into merge on a GATE provider.

1. The "cannot be steered off origin" assertion is tautological. It checks `new URL(u).origin === 'https://www.occ.com.mx'` for hostile inputs, but because `ORIGIN` is a hardcoded prefix, that origin is fixed regardless of the path, so the assertion passes even with `slugify` neutered to identity (I confirmed: `.../en-https://evil.example/` still resolves origin to occ.com.mx). It exercises nothing about slugify's neutralization. Fix: assert on the neutralized path instead, for example that `slugify`'s output matches `[a-z0-9-]*`, or that the built pathname contains no `..` or `//`, so removing slugify reddens it.

2. `redirect:'error'` is not pinned by the owned test. The provider passes `redirect: 'error'` to `fetchTextWithRetry`, but the stub `ctx.fetchText(url)` takes only the url and ignores opts, so removing `redirect: 'error'` from occ.mjs would not redden anything here. Real enforcement and its regression test live in the transport layer, but the provider's own contract is unpinned. Fix: have the stub record the opts and assert `redirect === 'error'` on the recorded call.

The substantive behavior (parse, dedup, pagination, slugify output, outage and challenge handling, the exact page cap) is soundly pinned, so once these two assertions actually redden on their regressions this is gate-ready on our axis. The branch is also behind main and will need a rebase.

santifer:
Hi @VictorMeneses77, and sorry for the long wait on a first contribution — that one's on us.

I read the provider properly. The OCC work is solid: the empty-result guard for a 200 that comes back with no cards, and the fact that you honour `ctx.maxPages` so the `verify-portals` health probe only pulls a single page, are both things several existing providers get wrong. The test file covers the parser and the failure paths.

One thing blocks it, and it's small: the branch now conflicts with `main` in `templates/portals.example.yml`. It's only the comment line listing the built-in providers — `main` added `telegram-channel` and reflowed it while your branch added `occ`. Merge `main` in and keep both names in that list; `docs/SUPPORTED_JOB_BOARDS.md` merges cleanly by itself.

Once that's rebased and CI is green again, this is next in line from my side.

PR #3751, opened 2026-09-03
fix(scan): opt-in location-aware company+role dedupe

### LINKED ISSUE #3750: scan: company+role dedupe drops the eligible city when one employer posts the same title in two countries
### What happened?

`scan.mjs` dedupes scanned jobs on `companyRoleDedupKey(company, role)` (`canonicalize(company) + '::' + normalizeRoleForDedup(role)`). The key has **no location component**, so when one employer posts the *same title in several cities*, only the first-seen posting survives; the rest are counted in `totalDupes` and dropped.

That collapse is deliberate and documented above `collectSeenCompanyRoles` (a company opening one req per city would otherwise leak a city variant into the pipeline on every scan), and it is the right default. The problem is what happens when the user's eligibility is location-bound: **the survivor is arbitrary** — whichever twin the ATS happened to return first — and it may be the city the user cannot legally work in, while the one they can is discarded.

`location_filter` cannot rescue this, and that is the part worth being explicit about: for a candidate whose allow-list contains both cities, the filter passes both, so it never discriminates and dedupe then picks at random. Ordering location filtering before dedupe would not change anything.

### Steps to reproduce

Anthropic's public Greenhouse board carries twin postings with adjacent IDs:

```
curl -s https://boards-api.greenhouse.io/v1/boards/anthropic/jobs
```

```
5101173008  Staff Software Engineer, AI Reliability Engineering  London, UK
5101169008  Staff Software Engineer, AI Reliability Engineering  Dublin, IE
5097742008  Staff Software Engineer, Inference                   London, UK
5150472008  Staff Software Engineer, Inference                   Dublin, IE
```

With an EU-shaped `location_filter` that allows both `London` and `Dublin`:

```js
buildLocationFilter(cfg.location_filter)('London, UK', url, title)  // true
buildLocationFilter(cfg.location_filter)('Dublin, IE', url, title)  // true

companyRoleDedupKey('Anthropic', 'Staff Software Engineer, Inference')
// 'anthropic::staff software engineer inference'  — identical for both cities
```

Scanning that board yields **2** pipeline entries instead of 4, one arbitrary city per role. A real scan history from an EU-based user shows 9 Anthropic rows, every one of them London and not a single Dublin — i.e. the four eligible EU postings had been silently dropped in favour of postings requiring UK work authorization.

### Expected behavior

The default should stay exactly as it is — the one-entry-per-role collapse is right for most people and must not regress.

But a location-constrained user should be able to opt in to keeping the city in the comparison, so both twins reach the pipeline and they choose. Proposal:

```yaml
scan_history:
  dedup_include_location: true
```

Absent (the default) = today's behaviour, byte for byte. Set, the normalized location joins the company+role key. Sources that record no location (a tracker without a Location column, a report-led processed pipeline row that puts a score where a pending row puts the city) must keep seeding a key that matches *every* city, so a role already applied to can never resurface city by city.

Happy to send the PR — it is written and green against the full suite.

### CLI tool

Claude Code

### OS and Node.js version

macOS 15 / Node 24

sebastienfi:
## What does this PR do?

Adds `scan_history.dedup_include_location` (opt-in, default off) so the posting location can join the company+role dedupe key. Without it, an employer that posts one title in two countries reaches the pipeline as a single arbitrary city — and for a location-constrained candidate that city may be the one they cannot legally work in.

## Related issue

Fixes #3750

## Type of change

- [x] Bug fix
- [ ] New feature

## The bug

`companyRoleDedupKey(company, role)` is `canonicalize(company) + '::' + normalizeRoleForDedup(role)` — no location. Twins collapse, the loser is counted in `totalDupes` and dropped.

Reproduced against a live public board (Anthropic's Greenhouse), not hypothetically:

```
5101173008  Staff Software Engineer, AI Reliability Engineering  London, UK
5101169008  Staff Software Engineer, AI Reliability Engineering  Dublin, IE
5097742008  Staff Software Engineer, Inference                   London, UK
5150472008  Staff Software Engineer, Inference                   Dublin, IE
```

Scanning that board through a local-parser fixture with an EU-shaped `location_filter`:

| | pipeline entries |
|---|---|
| before | 2 of 4 — one arbitrary city per role |
| after, flag on | 4 of 4, and a re-run adds 0 |
| after, flag absent | 2 of 4 — unchanged |

**`location_filter` is not the cause and reordering it would not fix this.** For a candidate whose allow-list holds both cities the filter passes both, so it cannot discriminate and dedupe then picks by feed order. Verified directly:

```js
buildLocationFilter(cfg.location_filter)('London, UK', url, title)  // true
buildLocationFilter(cfg.location_filter)('Dublin, IE', url, title)  // true
companyRoleDedupKey('Anthropic', 'Staff Software Engineer, Inference')
// 'anthropic::staff software engineer inference' — the same string for both
```

A real EU-based scan history held 9 Anthropic rows, every one London and not one Dublin: the postings requiring UK work authorization survived, the eligible ones were discarded.

## Why opt-in rather than always-on

The collapse is deliberate and the long comment above `collectSeenCompanyRoles` says why — one req per city would otherwise leak a variant per scan. That default is right for most people and this PR does not touch it. `dedup_include_location` is strict-boolean `true` only, so a typo reads as the default rather than as a pipeline full of city variants nobody asked for.

## Design note: a key with no location is a wildcard, not an empty component

This is the part that makes the flag safe, and the reason it is not a one-line concatenation.

The three seed sources do not all record a location. `applications.md` has no Location column in the default layout, and `pipeline.md`'s column after the role is positional: it is the city in the shape `appendToPipeline` writes, but the **score** in a report-led processed row (`#143 | {url} | … | 4.2/5 | PDF ✅`), and a labeled `posted:`/`trust:`/`note:` segment slides into it when an offer had neither location nor compensation.

So a bare key has to keep matching *every* city, otherwise turning the flag on would resurface a role the user had already applied to, once per city. Concretely:

- `companyRoleDedupKey(co, role)` with no location returns the identical string it always did; with one it returns `company::role@@location`.
- Seeding: a source that records a location seeds the located key; one that does not seeds the bare key.
- Checking: the scan tests the located key **and** the bare key. When the flag is off, or the provider gave no location, the two are the same string and the second lookup collapses away — the default path is unchanged.
- `pipeline.md` is read for a location only when `urlIndex === 0`, and never from a labeled segment.

`@@` rather than a third `::`, because `normalizeTextKey` strips punctuation from both the role and the location, so no component can contain the separator and a company alias cannot forge a located key out of a bare one.

## Tests

New file `tests/scan-dedup-location-aware.test.mjs` (auto-discovered, no registration), 15 assertions:

- the bare key shape is byte-identical, and blank/malformed locations degrade to it
- opt-in seeding: a London scan-history row seeds London only, leaving the Dublin twin eligible
- a tracker row with no Location column still seeds the wildcard
- every `pipeline.md` shape: the pending row's city is read; a report-led score cell, a `posted:` segment and a comp-forced empty cell are not
- **e2e, flag on**: two real `scan.mjs` runs over the existing three-city fixture — 3 entries after run 1, still 3 after run 2
- **e2e, flag absent**: the same board yields 1 and 1, gating the default against regression

Both e2e legs were written first and failed before the fix.

`node test-all.mjs` — **7315 passed, 0 failed**, 10 warnings, all pre-existing and unrelated (`Possible personal data` hits on `funding.json` / `HIRED.md` / `dashboard/` and fixture warnings in unrelated suites). `node validate-portals.mjs --file templates/portals.example.yml` and `--self-test` both clean.

## Docs

`templates/portals.example.yml` gains a commented block in the house style next to `recheck_after_days`, and `docs/SCRIPTS.md` a paragraph under `## scan`. No user-layer file is touched.

## Checklist

https://claude.ai/code/session_019SChXRvLUbh2yvpxUQbXjH

## Summary

Users can set `scan_history.dedup_include_location: true` to keep postings with the same company and role separate when normalized locations differ.

The default remains unchanged: one posting per company and role.

Location values use sorted, deduplicated location sets. Locationless records match postings in any city, so processed roles do not reappear by location.

Pipeline parsing reads locations only from valid pending-row positions. It ignores scores and recognized metadata segments.

Greenhouse multi-office locations now use deterministic ordering.

## User impact

Users with location constraints can receive all eligible city variants.

Users without the setting keep the existing deduplication behavior.

No existing configuration is required. Existing single-location keys remain compatible.

## Documentation and tests

The setting is documented in `templates/portals.example.yml:216` and `docs/SCRIPTS.md:326`.

Tests cover location normalization, wildcard behavior, history and tracker seeding, pipeline parsing, reordered locations, distinct locations, CLI behavior, Greenhouse ordering, and default behavior.

## Files touched

: `scan.mjs`
: `providers/greenhouse.mjs`
: `templates/portals.example.yml`
: `docs/SCRIPTS.md`
: `tests/scan-dedup-location-aware.test.mjs`
: `tests/providers/greenhouse.test.mjs`
: `tests/fixtures/reordering-multi-city-board.mjs`
: `tests/fixtures/locationless-relist-board.mjs`

The requested system files and directories were not touched: `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/` was touched through `providers/greenhouse.mjs`, and `.github/`.


Scott-Emberson (commented):
Reviewed at head c68d2dfc02. The owned test tests/scan-dedup-location-aware.test.mjs pins the opt-in behavior in both directions, and it is genuinely opt-in. With the flag on, a three-city board with one role keeps all three cities distinct on the first run and re-adds none on the second (same-location still dedupes), and with the flag absent the legacy one-per-role collapse to a single entry is unchanged, which is the default-off regression gate. The distinct-city key cases and the location-less wildcard case confirm the safety property that an already-applied role suppresses all cities. The flag resolver is a strict === true and the flag-off path is byte-identical, so nothing changes unless enabled. A flag-on collapse would fail the count-3 assertion and a flag-off leak would fail the count-1 assertion, so it reddens in both directions.

The substance is the scan dedup logic outside our owned paths, so this is a test-axis review: the owned test is sound and the opt-in is real, the merge is the maintainer's call. Nothing blocking from the tests/ side.

Scott-Emberson (commented):
Re-reviewed at head 2772bc5908. The change since my last look is additive hardening: a new section keys multi-place location values as a canonical set (order-independent across the common separators, with commas kept intra-place, and distinct sets staying distinct), a round-trip case confirms a re-ordered scan-history row still dedupes, and the labeled-cell detection is now restricted to the four real labels (posted, trust, note, rank) so a free-text Remote location is not misparsed as a label. The opt-in bidirectional behavior (flag on keeps distinct cities, flag off unchanged) is preserved and strengthened.

The substance is the scan dedup logic outside our owned paths, so this is a test-axis review: the owned test is sound and hardened, the merge is the maintainer's call. Nothing blocking from the tests/ side.

scan.mjs
```diff
@@ -1744,41 +1896,65 @@ export function companyRoleDedupKey(company, role, canonicalize = defaultCompany
...
+  // Location column; the default layout has none, so an applied role seeds the
+  // wildcard and no city variant of it can resurface.
   if (applicationsText) {
     const lines = applicationsText.split('\n');
     const colmap = resolveColumns(lines);
     for (const line of lines) {
       const row = parseTrackerRow(line, colmap);
       if (!row) continue;
-      add(row.company, row.role);
+      add(row.company, row.role, row.location);
```
nikitacometa, line 1945: `row.location` makes a tracker with a Location column a strict match, not the wildcard the docs promise. Probe: an Applied role re-added when the cell says `London` and the board says `London, UK`. The tracker is the "never show me this again" source; it needs no city granularity.

scan.mjs
```diff
@@ -1696,20 +1765,103 @@ export function normalizeRoleForDedup(role) {
...
+ * extra cities into the string themselves, `' · '`-joined, in whatever order the
+ * upstream array happened to arrive in.
+ *
+ * `,` is deliberately NOT a separator. It is the city/region delimiter INSIDE a
+ * place ("London, UK"), so splitting on it would shatter every ordinary location
+ * into fragments and make "London, UK" and "Dublin, UK" share the fragment `uk`.
+ *
+ * Used only by {@link normalizeLocationForDedup}; nothing else parses the field.
+ */
+const LOCATION_LIST_SEPARATOR_RE = /\s*(?:[;|\u00b7/]|\bor\b)\s*/iu;
```
nikitacometa, line 1787: `\bor\b` is case-insensitive: `Portland, OR` becomes `portland`, `Portland, OR, USA` becomes `portland+usa`, and `OR` alone becomes `''`, the wildcard. CodeRabbit's `/\s*[;|·/]\s*|\s+or\s+/iu` on this line fixes it.

docs/SCRIPTS.md
```diff
@@ -553,6 +553,10 @@ Zero-token portal scanner. Runs configured local parsers for SSR/static career p
...
+
+The location component is the canonical **set** of the places a posting names, not the provider's display string. That field is free text and is often not one place: live Greenhouse boards pack several into one value with `;`, `|`, `/` or the word `or`, sometimes mixing two separators in the same value, and several providers here (greenhouse, ashby, eightfold, gem, ibm, echojobs) fold a multi-site role's extra cities into the string themselves in whatever order the upstream array arrived. Keying that string verbatim is stable only while the order holds, so a re-ordered list would read as a new posting and re-enter the pipeline. Splitting on those separators, normalizing each place, deduplicating and sorting makes the key depend on which places a posting names rather than the order it names them in. `,` is not a separator - it delimits city from region inside one place.
```
nikitacometa, line 558: Order-independence holds only for the four listed separators. `Remote - US/Canada` vs `Remote - Canada/US`, `London – Dublin` vs `Dublin – London` key differently, so a re-list at a new URL comes back.

templates/portals.example.yml
```diff
@@ -31,6 +31,30 @@
...
+#
+# Set dedup_include_location: true to keep the city in that comparison, so
+# "Staff Engineer — London" and "Staff Engineer — Dublin" are two entries rather
+# than whichever one the ATS happened to return first. Turn it on when your
+# eligibility is location-bound (work authorization, relocation, an office you
+# must be near): location_filter cannot choose for you when it allows both
+# cities, and the arbitrary survivor may be the one you cannot take.
+# Expect more entries per scan at large multi-country employers.
+#
+# Rows that record no location (a tracker without a Location column, a processed
```
nikitacometa, line 47: "a tracker without a Location column" is the only wildcard case today; say what happens when the column exists, or make the code match this sentence.

providers/greenhouse.mjs
```diff
@@ -196,7 +196,14 @@ export default {
       let location = j.location?.name || '';
       if (officeMap && isWorkModelOnly(location)) {
         const offices = officeMap.get(j.id);
-        if (offices && offices.size > 0) location = [location, ...offices].join(' · ');
+        // Sorted, not in /offices traversal order. The set is built by walking
```
nikitacometa, line 199: Sorting changes the `location` string written to `scan-history.tsv` for existing multi-office rows. URL dedupe still covers them; one line saying so.

nikitacometa (changes requested):
Verified at 2772bc59: default off (a typo in the key reads as off), `scan-history.tsv` format unchanged, an upgrade with the flag then on adds only the Dublin twin and a re-run skips both, `--only scan` 162/0, and the new suite reddens with `scan.mjs` at base (9 passed, 3 failed) and with `resolveDedupIncludeLocation` forced to false (38/3). Two blockers.

scan.mjs:1945 passes `row.location` from `applications.md`. A Location column is a supported layout (tracker-parse.mjs:216; merge-tracker.mjs:942 prints "Detected Location column"), and its cell is user-typed, so it will not match the ATS string. Tracker row `Fixture Defense | Strategic Finance Manager | London | Applied` and a board serving the same role at `London, UK`: flag off gives `Duplicates: 1 skipped / New offers added: 0`, flag on gives `New offers added: 1`. The applied role re-enters the pipeline. The description, docs/SCRIPTS.md:556 and portals.example.yml:47 say a tracker row seeds a wildcard; true only without that column. Simplest fix: do not pass `row.location`.

scan.mjs:1787 still splits on `\bor\b`, case-insensitively, so it eats the state code: `Portland, OR` keys as `portland` and collides with `Portland`; a location of exactly `OR` becomes `''`, the wildcard. CodeRabbit's `/\s*[;|·/]\s*|\s+or\s+/iu` on that line is unanswered.

Smaller: docs/SCRIPTS.md:558 overstates order-independence (`Remote - US/Canada` and `Remote - Canada/US` key differently, as do `London – Dublin` and `Dublin – London`).

PR #3754, opened 2026-09-03
docs(providers): fill gaps in the provider-authoring docs

ilyakanevskiy:
## What does this PR do?

Follow-up to #3528, which introduced `providers/ADDING_A_PROVIDER.md`. Fills gaps in the guide surfaced by later provider reviews, adds a missing rule to the Source Indexing Policy it defers to, plus small consistency fixes in files the guide points at. Documentation only — no provider behaviour changes.

### `providers/ADDING_A_PROVIDER.md`

**`detect()` and the URL a user actually has (§1).** `detect()` must claim the URL a real entry point hands the user — a careers-page link, an aggregator, a search hit — which is usually the ATS landing/root. Where a listing is served at several paths, match the shallowest; requiring a deeper canonical path the user would have to build by hand defeats the point of a URL-pattern `detect()` and the entry silently lands in `verify-portals` `skipped`.

**The `enrichDate` hook (§1, new subsection).** The guide never mentioned it, yet §4 sends HTML-scraper authors to `icims.mjs`, which has it — so it gets copied without context. Documents what it is, that its only caller is `scan-ats-full.mjs` (which imports a fixed module set, so the hook is inert until the provider is wired in there), that `_registry.mjs` never calls it, and that it overlaps the `fetchDetails` opt-in — pick one, don't ship both.

**`fetchDetails` is opt-in *only* (§1).** Gate the whole detail loop on the entry config asking for it **and** on not being in a probe (`smartrecruiters.mjs` — `if (fetchDetails && !probing)`; `vdab.mjs`). A provider that fetches per-posting detail on every scan with no config switch defeats the opt-in. `detailLimit` is the cap on that loop, never the thing that turns it on.

**Single-company vs aggregator header comment (§1).** Point at `ibm.mjs` / `dassault.mjs` for a single-company `tracked_companies:` adapter alongside the existing `remotli.mjs` / `yourator.mjs` aggregator references.

**`robots.txt` in "Before the code" (§ eligibility, new bullet).** A source-eligibility bullet pointing at Source Indexing Policy rule 6: check `robots.txt` per path — a `User-agent: *` disallow covering the listing paths the provider reads is a stop (precedent: Lagou), a disallow on routes the parser never touches (`/api`, a subscribe form) is not (ITviec, CareerViet), and a rule that only names an AI/agent crawler is a maintainer call rather than an automatic no.

**Wildcard-tenant SSRF allowlisting (§2, new bullet).** A `*.vendor.com` tenant host can't enumerate a `Set` — match it with an anchored full-host regex (`bamboohr.mjs`, `breezy.mjs`) or `hostname.endsWith('.vendor.com')` (`workday.mjs`, `icims.mjs`), with the anchoring pitfalls spelled out: keep the trailing `$`, escape the dots, keep the leading dot on a suffix test.

**Redirects (§2, new bullet).** `redirect: 'error'` is the default and stays it even when the source's own bootstrap issues a legitimate same-origin 3xx. A provider that needs to *inspect* a redirect without following it passes `redirect: 'manual'` (`jobvite.mjs`, `telegram-channel.mjs` — a structured error carrying `.status` / `.location`). Following a hop is out: target the settled URL directly (`deutschebahn.mjs`) or rewrite a known static 3xx up front (`builtin.mjs`). The one exception is a session-establishing bootstrap 3xx whose target is not knowable ahead of time — follow it manually, re-validating **every** hop's `Location` against the tenant origin and bounding the chain with a hop cap. That is stricter than the single up-front check, not a relaxation of the guard.

**Empty board vs broken parser (§2).** §2 stated this only for a JSON envelope (`parseIbmResponse`); the two §4 HTML references disagreed — `join.mjs` throws on a structural mismatch and returns `[]` only for a genuinely empty list, `icims.mjs` returns `[]` either way. Reworked into three outcomes (`[]` / descriptive `throw` / per-row `continue`) with a `parseIbmResponse`-shaped code sample that also covers a bare non-object body, and an explicit HTML/SSR-scraper paragraph citing `join.mjs` (`!Array.isArray(items)` throws, `items.length === 0` returns `[]`); notes that `icims.mjs` deliberately does not draw this line and a new scraper should not copy that.

**Short-page stop counts the source's rows (§2).** The loop-termination count must be what the *source* returned for that page — never a count the provider already narrowed, whether by dropping malformed rows or by de-duping against a shared `seen` set across query/category passes. Covers the fan-out case where an earlier pass covers a later pass's *first* page while that pass's deeper pages still hold unique postings, so a "nothing new" stop ends it pages early. References `oraclecloud.mjs`, `jobbankca.mjs` (`rawEntryCount`).

**Config-derived URL segments (§2).** A config-derived path segment (`locale`, `office`) is handled the opposite way from a host-controlled field: not `safeEncodeURIComponent` + drop — dropping real postings over a config typo is the wrong trade — but `encodeURIComponent`'d **once outside the row loop**, so a structural `/` `?` `#` stays escaped and a genuinely broken value throws there as the loud config error it is. `garena.mjs` splits the two explicitly (`urlSegment()` throws for config, `idUrlSegment()` drops for the response `id`); `tkms.mjs` hoists its locale encode out of the loop. A value a regex already restricted (`4dayweek.mjs`'s `SLUG_RE`) needs nothing further.

**Pacing a `fetchDetails` loop (§ Pacing and retry).** A non-paginating provider that fires `detailLimit` sequential detail fetches is the same WAF-burst risk as a page walk — pace it with the same inter-request `sleep` and wrap each fetch in `fetchTextWithRetry` / `fetchJsonWithRetry`; a failed detail fetch is caught, the listing row kept, and the run never fails. Also: import `sleep` from `_http.mjs`, don't call `ctx.sleep(ms)` directly — `ctx.sleep` is `Optional` in the §1 contract, so a ctx with no clock turns a bare `ctx.sleep(...)` into a `TypeError` out of `fetch()`.

**Stop-reason drives both the warning and the marker (§ Pacing and retry).** Track *why* the walk stopped — ceiling, fetch error, unrecognised page — as one value and drive both the "raise `max_pages`" warning and the incomplete marker's `reason` from it (`workday.mjs`'s `stopReason`; `phenom.mjs` is the compact version). Keep it independent of whether the source ever reported a `total` — deriving the marker only from a `total`-based "did we reach the count" check drops the signal on exactly the tenants that omit the count.

**Page budget counts result pages (§ Pacing and retry).** `max_pages` / `ctx.maxPages` count pages *of results*. A one-shot bootstrap that carries no listings — a token/cookie primer, a board-id lookup — is free and sits outside the count (`csod.mjs`). But a first request that does return the first batch of postings is page one: issued before the pagination loop, the loop bound is `maxPages - 1` (floored at 0), and `ctx.maxPages: 1` then means that first request and nothing else — the "exactly one list request" probe assertion in §3 depends on it.

**A short final page beats the `max_pages` ceiling (§ Pacing and retry).** "Hit the ceiling" means the budget ran out *and* the last fetched page was full, by the source's own raw row count. A shorter-than-page-size final page is a natural end — stop reason `complete`, not `cap`, and nothing warns. Gating the incomplete marker on the page index alone (`page === lastAllowedPage`) mislabels a naturally-short final page as truncated. References `eightfold.mjs`, `jobbankca.mjs` (`rawEntryCount`).

**Source-reported counts are not proof of completeness (§ Pacing and retry).** A `total` can be silently clamped (already noted via `workday.mjs`'s facet split, #3310) and a results count can just as easily *overstate* what pagination will serve — observed live across every transport a source exposed. A page count is computed differently and held up as more reliable in the same checks, so bound the walk by it when given one — but never treat either number, pages or results, as *proof* the walk is complete. Only the walk's own natural end (an empty or short page) is that proof.

**Hard result-window ceilings (§ Pacing and retry).** A paginating JSON API can hit a source-side result-window cap that looks nothing like a short page: every page up to a round number (often a multiple of the page size — 100 pages of 100 is a signature) comes back full, then every page past it is a clean, well-formed, structurally *empty* response. That is the default max-result-window on an Elasticsearch/Solr-style backend, not a walk bug, and usually has no client-side workaround unless the source exposes a documented alternate path. Treat it as any other natural stop (`complete`); don't let `max_pages` / `max_jobs` take the blame.

**`Crawl-delay` as an inter-page floor (§ Pacing and retry).** When a source states a `robots.txt` `Crawl-delay` (rule 6), it is a reasonable floor for `INTER_PAGE_DELAY_MS` directly — no need to observe throttling first. It governs spacing *between* requests, not a mandatory pause before the first one, and is not part of RFC 9309 (`jobbankca.mjs` delays before page 1 too, but that is a per-provider choice, not a rule-6 requirement).

**Test requirements (§3).** `redirect: 'error'` assertions extended to every request the provider issues — list, pagination, detail-enrichment alike — with the sole exception of a deliberate `redirect: 'manual'` inspection request (`jobvite.mjs`). The SSRF-guard test must show the fetch stub was **never entered** (call count `0`, or a stub that fails on entry) — a stub that only throws passes whether the guard fired or a fetch was wrongly attempted. New required cases: scraper empty-vs-broken fixtures (`join.mjs`), a shared-`seen` pass that still walks its deeper pages and returns their unique postings, a failed detail fetch that leaves the listing row intact and the run green, and a `detailLimit` cap test (`smartrecruiters.test.mjs` runs 40 postings at `detailLimit: 10` and asserts exactly 10 detail calls).

**Checklist + reference table.** New checklist lines for the config-derived-segment rule, the `fetchDetails` pacing rule, the `enrichDate` constraint, and the `robots.txt` check. `SUPPORTED_JOB_BOARDS.md` sort clarified as case-insensitive (a lowercase name sorts among its letter's rows, not after `Z`). New reference-table rows for empty-board-vs-broken-selector (`join.mjs`) and wildcard-tenant allowlisting (`bamboohr.mjs`, `workday.mjs`).

**Dropped the hand-maintained no-`detect()` name list.** `portals.example.yml` carried an inline list of provider ids that omit `detect()`, duplicating what the commented stanzas already encode — each carries its own `provider:` line. Removed both the list and the checklist step that told authors to append to it; the stanzas are now the authoritative list.

### `CONTRIBUTING.md`

**Source Indexing Policy — new rule 6, `robots.txt`.** The policy had five rules and none about `robots.txt`, yet a source's `robots.txt` has been a de-facto eligibility gate in the `SUPPORTED_JOB_BOARDS.md` "Evaluated, not supported" table (Lagou, TopDev, CareerLink) with no written rule behind it — and that table conflated a `User-agent: *` path block, which binds the scanner, with an AI-crawler-only block, which does not. Rule 6 makes it explicit: the check is per-path under the group that binds career-ops's own client; a `Disallow` covering the listing paths the provider reads is a stop and goes in the Evaluated table, a `Disallow` on routes the parser never requests (`/api`, a subscribe form) is not; a rule scoped to a named AI/agent crawler (`ClaudeBot`, `anthropic-ai`) doesn't name the scanner — a distinct, bounded `career-ops` fetcher — but since a scan can be the first step toward a model reading the posting in a later `pipeline` / `triage` run, whether the operator's intent reaches it is a maintainer call on a source proposal rather than an automatic disqualification. Bumps the two "five rules" references to "six".

### `.github/ISSUE_TEMPLATE/source-proposal.yml`

Adds a **Rule 6 — robots.txt** field so a source proposal states the `robots.txt` situation (which `User-agent` group applies, whether any `Disallow` covers the listing paths, any AI-crawler rule) up front, and updates the "five rules" wording to "six".

### `providers/consider.mjs`

Corrected an inaccurate comment on the CSRF handshake. It claimed `redirect: 'manual'` was unusable because undici returns an opaque status-0 response with empty headers — but on Node's fetch a manual 3xx is a real non-ok response with a readable `Location` header (that is exactly what `jobvite.mjs` uses to classify a redirect target). The handshake stays on `'error'` because it only needs to *refuse* the redirect, not read it — the simpler choice, not the only workable one.

### `templates/portals.example.yml`

Replaced the hand-maintained inline list of explicit-`provider:` modules with a pointer to the commented stanzas as the authoritative source — the same change as the `ADDING_A_PROVIDER.md` checklist edit above. Two of those modules (`torre`, `yourator`) turned out to have no stanza at all, so the "authoritative" claim wasn't true yet — added both, plus `tests/providers-explicit-only-stanza.test.mjs`, which loads every provider, filters to the ones with no `detect()`, and asserts each has a `provider: <id>` stanza in this file, so the list can't silently drift again.

## Related issue

None — documentation.

## Type of change

- [ ] Bug fix
- [ ] New feature

## Depends on

#3513 — §2's id/slug guidance describes the `safeEncodeURIComponent` per-row-drop behaviour as already on `main`; it lands with #3513.

## Checklist

- [x] If this is a new feature or architecture change, I opened an issue first (docs are exempt — this is docs)


luochen211 (commented):
Reviewed at head dfb2b28 and replayed it cleanly onto current main (5c8b02a). I cross-checked the new claims against the current tree: `scan-ats-full.mjs` is the only `enrichDate` caller, it uses the fixed five-provider set described here, `_registry.mjs` does not call that hook, and the cited `join.mjs` structural-empty distinction is accurate. `git diff --check` is clean. No blocking finding from this pass; maintainer approval/merge remains the project decision.

Scott-Emberson (commented):
Reviewed at head dfb2b2880f. This is a docs-only change to providers/ADDING_A_PROVIDER.md, filling gaps in the provider-authoring guide, with no code or test in the diff, so there is nothing to verify on the test axis, it is a documentation edit in our providers/ area. It reads accurately against how the provider contract and the test conventions actually work. The branch is behind main, so a rebase before merge would be worth it, but nothing here is blocking from our side.

providers/ADDING_A_PROVIDER.md
```diff
@@ -306,6 +364,11 @@ Two mechanisms, both expected once a provider paginates:
   `{ retries: 2, baseDelayMs: 500, maxDelayMs: 8_000 }`; pass a `policy` 4th
   argument for a different cadence (`workday.mjs`, `oraclecloud.mjs` use
   `{ retries: 3 }` because their API is WAF-fronted).
+- **A `fetchDetails` loop counts too.** A provider that does not paginate
+  but issues up to `detailLimit` sequential detail fetches under
+  `fetchDetails: true` is the same burst against the same WAF — pace it
+  with the same inter-request `sleep`, and wrap each detail fetch in
+  `fetchTextWithRetry` / `fetchJsonWithRetry`.
```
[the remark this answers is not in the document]
ilyakanevskiy: The best-effort behavior is now spelled out in the §2 "Pacing and retry" `fetchDetails` bullet: a detail fetch that exhausts retry is caught, the listing row is kept, and the sweep continues — never fatal to the target (`smartrecruiters.mjs`, `vdab.mjs`). §3 gains a matching test line: an exhausted detail-enrichment fetch leaves the listing intact.

providers/ADDING_A_PROVIDER.md
```diff
@@ -372,9 +438,17 @@ went through them. Must cover:
...
 - `fetch()`: normalization of the source's real response shape; rows missing
   a required field are filtered; **`redirect: 'error'` is passed on every
-  request** — assert `opts.redirect === 'error'`, not just that the call
-  happened; the allowlist guard throws **before** `fetchJson` / `fetchText`
-  is called.
+  request the provider issues** — list, pagination, and detail-enrichment
+  fetches alike — assert `opts.redirect === 'error'`, not just that the call
+  happened; the sole exception is a deliberate redirect-*inspection* request
+  (`redirect: 'manual'`, to read `.location` off the thrown 3xx without
+  following it — `jobvite.mjs`), and `redirect: 'follow'` is never used. For
```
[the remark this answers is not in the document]
ilyakanevskiy: `providers/ADDING_A_PROVIDER.md` is the provider guide, not a provider module -- its own header ("This file is the checklist and the set of requirements") already establishes that. The matching-provider-test convention is for `providers/*.mjs` adapters; adding a "no test required" disclaimer to the guide to satisfy a per-path rule would just be noise. No change here.

Scott-Emberson (commented):
Re-checked at head 4ec84e5111. Still a docs-only change to providers/ADDING_A_PROVIDER.md with no code or test, so nothing to verify on the test axis, and it reads accurately against the provider contract. The branch is behind main, worth a rebase before merge, but nothing here is blocking from our side.

providers/ADDING_A_PROVIDER.md
```diff
@@ -287,12 +407,18 @@ or a probe-budget cut-off.
 
 A full walk of a large board is 100+ sequential requests (`workday`,
 `radancy`), and several sources sit behind a WAF that rate-limits in bursts.
-Two mechanisms, both expected once a provider paginates:
+Two mechanisms, both expected once a provider paginates — or fans out
+per-posting detail fetches:
 
 - **Inter-page delay.** A module constant applied only to pages past the
   first: `if (page > 0) await sleep(INTER_PAGE_DELAY_MS, ctx)`. Import
   `sleep` from [`_http.mjs`](_http.mjs) (it honours a ctx-supplied test
```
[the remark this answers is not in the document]
ilyakanevskiy: The `if (page > 0)` snippet here isn't copied from a referenced loop — this bullet cites providers only for delay *magnitudes* (`careerviet` / `itviec` at 750 ms, `agentic-jobs` at 2100 ms). `4dayweek.mjs` appears in this guide exactly once, at line 316, about `SLUG_RE` — an unrelated context. There is no one-based loop the example is out of step with.

`if (page > 0)` is the zero-based idiom and the majority one in the tree: `avature`, `eightfold`, `getro`, `icims`, `oraclecloud`, `deutschebahn`, `phenom`, `tkms` all gate the inter-page delay that way, and `avature.mjs`'s constant comment reads "only boards that paginate past page 0 pay it." The snippet is internally consistent — zero-based `page`, `page > 0` skips the first request — and the prose ("applied only to pages past the first") states the intent independently of the counter base; a reader whose loop starts at 1 adapts this guard the same way they adapt every other detail to their own loop.

Leaving the example as-is. The sibling redirect (`never followed automatically`) and empty-response findings were valid and are fixed in 352e9d9.

Scott-Emberson (commented):
Re-checked at head c0699f790d. Still a docs-only change to providers/ADDING_A_PROVIDER.md, now rebased and up to date (blocked only on review). It reads accurately against the provider contract and the test conventions, and there is no code or test to verify on the test axis. As a pure doc in our providers/ area it is really the maintainer's to land, nothing blocking from our side.

luochen211 (changes requested):
Re-reviewed current head c0699f7 after the documentation follow-ups. The new source-proposal Issue Form is invalid YAML: `.github/ISSUE_TEMPLATE/source-proposal.yml` line 72 adds an unquoted plain scalar containing `User-agent: *`, and the colon-space terminates the mapping value. Parsing with the repository's own dependency fails at 72:103 with `YAMLException: bad indentation of a mapping entry`. GitHub therefore cannot load the form.

Repro:
`node --input-type=module -e "import {readFileSync} from 'node:fs'; import yaml from 'js-yaml'; yaml.load(readFileSync('.github/ISSUE_TEMPLATE/source-proposal.yml', 'utf8'))"`

Please quote that full description or make it a folded/block scalar, and add/extend a syntax gate for Issue Form YAML so this cannot pass CI again. `providers/consider` remains green locally (24/24); this finding is isolated to the newly edited form.

providers/ADDING_A_PROVIDER.md
```diff
@@ -191,11 +311,21 @@ UTF-8, and where the encoded value is also the dedup key (`arbeitsagentur`,
...
 
-Scope is exactly this case — a host-controlled API field becoming a URL path
-segment in a loop. Config-derived values (a company slug from `portals.yml`,
-a keyword, a locale), calls already inside their own try/catch, and values
-already checked against a slug charset are left alone; dropping a real
-posting over a bad character in config is the wrong trade.
+Scope is exactly this case — a **host-controlled** API field becoming a URL
+path segment in a loop. A **config-derived** segment (a company slug from
+`portals.yml`, a locale) is handled the other way: not through
+`safeEncodeURIComponent` + drop — dropping real postings over a typo in
```
nikitacometa, line 341: A second reference to `safeEncodeURIComponent`, but `providers/_safe-url.mjs` does not exist in the tree; the import at 295 fails with ERR_MODULE_NOT_FOUND. In scope here since the PR is about the guide matching the code.

providers/ADDING_A_PROVIDER.md
```diff
@@ -414,6 +617,8 @@ Dev loop: `node test-all.mjs --only providers/{name}`. Before a PR: the full
 | Pagination honoring `ctx.maxPages` | `providers/workday.mjs` |
 | HTML scraping with the shared `decodeEntities` | `providers/icims.mjs` |
 | SSR JSON inside HTML (`__NEXT_DATA__`) | `providers/join.mjs` |
+| Empty board vs a broken selector, told apart | `providers/join.mjs` |
+| Wildcard-tenant host allowlisting (anchored, not a `Set`) | `providers/bamboohr.mjs`, `providers/workday.mjs` |
 | RSS parsed in-process | `providers/larajobs.mjs` |
 | `job.url` from a host-controlled `id` / `slug` via `safeEncodeURIComponent` | `providers/phenom.mjs`, `providers/bamboohr.mjs` |
```
nikitacometa, line 694: phenom.mjs:143 and bamboohr.mjs:106 both use bare `encodeURIComponent`, not `safeEncodeURIComponent`, so this row cites them for a rule they do not follow.

templates/portals.example.yml
```diff
@@ -924,9 +924,10 @@ search_queries:
 #
 # Some providers are job boards / aggregators with no per-company careers
 # URL to detect, so they must be selected with an explicit `provider:`
-# field: 4dayweek, agentic-jobs, arbeitnow, arbeitsagentur, flowxtra, getonbrd, glints,
-# himalayas, ibm, jobbankca, jobicy, jobstreet, justjoin, landingjobs, mycareersfuture, nofluffjobs, remoteok, remotive, senjob,
-# manfred, thehub, vdab, weworkremotely, workingnomads, wttj, a16z-speedrun-talent. Examples for every built-in provider are below.
+# field. Those are the provider modules that omit detect(); the commented
+# stanzas below are the authoritative list of them — each one carries its
+# own `provider:` line. (A provider with a URL-pattern detect() is
```
nikitacometa, line 989: Checked all 85 provider modules: 17 have no `detect()`, and `torre` and `yourator` have no stanza in this file, so the stanzas are not yet the authoritative list. Add both, plus a test that enumerates `providers/` and asserts a stanza per explicit-only id.

.github/ISSUE_TEMPLATE/source-proposal.yml
```diff
@@ -65,11 +65,18 @@ body:
       placeholder: "Featured listings exist; the API marks them with featured: true and returns all listings regardless"
     validations:
       required: true
+  - type: textarea
+    id: rule6
+    attributes:
+      label: "Rule 6 — robots.txt"
+      description: "Paste the source's robots.txt. Which group applies to a generic client (`User-agent: *`)? Does any `Disallow` cover the listing paths a provider would read, or only routes it never requests (`/api`, a subscribe form)? Note any rule that names an AI/agent crawler (`ClaudeBot`, `anthropic-ai`)."
```
nikitacometa, line 72: Quoting fixes the parse. The syntax gate from the same review is not here; #3802's tracked-YAML test covers this directory once it lands.

nikitacometa (commented):
The Issue Form YAML is valid at 2ec424f2 (js-yaml loads it), so the standing blocker is resolved, and all 14 inline threads are addressed or withdrawn. The gate the same review asked for is not in this PR, but #3802 (open) adds a tracked-YAML parse test that covers `.github/ISSUE_TEMPLATE/*.yml`; once it lands, that half is done. Verified the `consider.mjs` correction: Node 24.19 with `redirect:'manual'` gives status 302 and a readable `Location`, so the comment this replaces was wrong. Merges clean against main and extends the guide that landed in #3528 rather than conflicting with it.

Two doc-versus-code gaps, the first pre-existing but in scope for a PR whose job is making the guide match the code. ADDING_A_PROVIDER.md:295 tells authors to `import { safeEncodeURIComponent } from './_safe-url.mjs'`; that module does not exist (ERR_MODULE_NOT_FOUND from `providers/`), nothing exports that name, and the cited exemplars phenom.mjs:143 and bamboohr.mjs:106 use bare `encodeURIComponent`. Main has 7 references; this PR adds 2 more (317, 671). Every provider PR this week hit the same wall. Either land the module or rewrite those paragraphs around `encodeURIComponent` plus the explicit null/drop check the rule wants.

portals.example.yml:927 now says the stanzas are the authoritative list of no-`detect()` providers. Of 85 modules, 17 have no `detect()`, and `torre` and `yourator` have no stanza; SUPPORTED_JOB_BOARDS.md:94 sends users to `provider: yourator`. The removed list also omitted both, but it was a fallback, and the new sentence is an invariant nothing enforces. Add the two stanzas and a test that makes the sentence true by construction.

providers/ADDING_A_PROVIDER.md
ilyakanevskiy: This is the known sequencing gap already flagged in https://github.com/career-ops-hq/career-ops/pull/3528#issuecomment-5525302587 — the guide's `safeEncodeURIComponent` copy describes a helper that lands with #3513, which hasn't merged yet. Proposing no action here; this PR is about the guide matching the code, but the fix belongs to #3513 landing, not to this one.

providers/ADDING_A_PROVIDER.md
ilyakanevskiy: Same root cause as the thread above — https://github.com/career-ops-hq/career-ops/pull/3528#issuecomment-5525302587. This row's citation is only wrong because it's describing the post-#3513 state ahead of #3513 landing. Proposing no action here for the same reason.

templates/portals.example.yml
ilyakanevskiy: Fixed in c2ec040 — added stanzas for `torre` and `yourator` to templates/portals.example.yml, and added tests/providers-explicit-only-stanza.test.mjs, which loads every provider via `loadProviders()`, filters to the ones with no `detect()`, and asserts each has a `provider: <id>` stanza in the example file. That's checked against the real provider directory rather than the file's own hand-written claim, so this can't silently drift again.

.github/ISSUE_TEMPLATE/source-proposal.yml
ilyakanevskiy: Confirmed — the quoting is already in this PR (86efc33), and I'm leaving the syntax gate itself out of scope here since #3802 covers it once that lands.

luochen211 (approved):
Re-reviewed current head f8783475. The quoted rule 6 description now parses with the repository's js-yaml; source-proposal.yml and portals.example.yml both load successfully. The consider provider suite passes 24/24 and the explicit-only provider stanza checks pass 2/2. I also confirmed the separate YAML syntax gate in #3802 is merged. My previously reported YAML blocker is resolved; approved from this review scope.

santifer:
Sorry for the wait, @ilyakanevskiy — I promised this read two weeks ago and you'd already done your part.

The check I said I'd make comes out clean: #3513 and #3802 are merged, so `safeEncodeURIComponent` is on `main` and `tests/yaml-syntax.test.mjs` is already here. @nikitacometa's four points hold against the current tree too — `providers/_safe-url.mjs` exists, and `phenom.mjs` and `bamboohr.mjs` both import and use `safeEncodeURIComponent`.

I followed the code references rather than trusting them, including `_ip-guard.mjs`: `_http.mjs` wraps every `ctx.fetch*` in `providerFetchContext`, so a branded-host provider is covered without its own allowlist.

Your new test is load-bearing — I ran it with your branch merged into current `main` and it passes. It's the enumeration @nikitacometa asked for, so torre and yourator are covered by the assertion, and it fails on zero explicit-only providers instead of passing vacuously.

On rule 6: that last clause was a maintainer call, not an edit to ask you for, and it's now decided in favour of what you wrote. That was what held this up. One note: deleting the explicit-provider name list touches the same line #3748 edits to add `occ`, so whichever lands first invalidates the other's hunk; your test stays green either way.

What's left is on our side: a branch update and fresh CI on the merged tree.

PR #3757, opened 2026-09-03
feat(wttj): server-side Algolia filters so a scan can be exhaustive

nicolasey:
## The problem

A keyword query cannot narrow the Welcome to the Jungle board. It is Algolia's relevance ranking — not the scanner's config — that decides which `max_hits` results come back, and `title_filter` / `location_filter` only run *afterwards*, on what already arrived. Anything past the cap is invisible no matter how well it matches the user's own filters.

The provider caps at `MAX_HITS_CAP = 200` per query, and the documented example config uses broad keywords. Measured against the live index on 2026-09-03:

| Query / filter | Hits | Seen |
|---|---|---|
| `"product manager"`, unfiltered | 13 981 | 200 (**1.4 %**) |
| `"senior product manager"` | 4 295 | 200 |
| `"lead product manager"` | 4 093 | 200 |
| `"head of product"` | 895 | 200 |
| `"director of product"` | 1 230 | 200 |
| `"group product manager"` | 692 | 200 |
| `"chef de produit digital"` | 263 | 200 |

Those seven queries return **1 226 distinct jobs in total** — a sample of a board that holds far more matching roles.

This is not theoretical. A real posting (Paris, product management, `full_time`, published *inside* the scan window) sat past rank 200 on `"product manager"` and was never captured across six scan runs, despite passing every configured `title_filter` and `location_filter` rule.

## The fix

Add a `wttj.filters` passthrough for Algolia's filter expression, applied to every configured query. A filter shrinks the result set **server-side** instead of re-ranking it, which is what makes a scan exhaustive rather than a sample:

| Filter | Hits | Requests to exhaust |
|---|---|---|
| `offices.country_code:FR` | 5 091 | 26 |
| `+ contract_type:full_time` | 4 384 | 22 |
| `+ new_profession.sub_category_reference:"product-management-wNjYw"` | **450** | **1** |

Because a filter bounds the result set, the per-query cap rises from 200 to Algolia's per-request ceiling for this index (1000) **when and only when** a filter is configured. Unfiltered queries keep the 200 cap, since they can still match the whole board.

```yaml
  - name: Welcome to the Jungle
    provider: wttj
    wttj:
      filters: 'new_profession.sub_category_reference:"product-management-wNjYw" AND offices.country_code:FR AND contract_type:full_time'
      max_hits: 1000
    enabled: true
```

`queries` becomes optional when `filters` is set — the empty query means "everything that passes the filter", which is the useful default for a taxonomy filter. With neither, the provider still throws rather than scanning the board, so the existing guard is preserved.

Useful faceted attributes on this index, documented in the provider header: `offices.country_code`, `offices.state`, `offices.city`, `contract_type`, `remote`, `experience_level_minimum`, `salary_yearly_minimum`, `organization.name`, and WTTJ's own job taxonomy `new_profession.sub_category_reference`.

## Measured effect

On one real config, replacing the seven truncated keyword queries with a single taxonomy-filtered request: **450 jobs found, 120 new** (previously unseen) in one request. A full scan run went from 32 new offers to 167.

## Safety

- The filter string is parsed by Algolia as a filter expression and **never reaches a URL or host**, so the existing `assertHost` guard still covers every request target.
- Length is bounded (`FILTERS_MAX_LEN = 1000`) so a config typo cannot build an unbounded request body.
- A blank/whitespace `filters` value is treated as absent: the cap stays 200 and no `filters` param is sent.

## Tests

Seven cases added to `tests/providers/wttj.test.mjs`, all following the existing discovered-test layout:

- passes `wttj.filters` through to Algolia
- uses the empty query when only filters are configured
- applies the filter to **every** configured query, not just the first
- still caps `max_hits` at 200 when no filters are configured
- raises the cap to 1000 when filters narrow the board
- treats a blank `filters` as absent (cap stays 200, no param sent)
- rejects an over-long filters expression

Full suite: `7719 passed, 4 failed` with this change vs `7712 passed, 4 failed` on `main` at the same commit — same four pre-existing failures (an invalid Gemini API key in the environment, a locale-dependent number format assertion, and unregistered root-level `*.test.mjs` files in the `SYSTEM_PATHS` coverage guard). No new failures; +7 passing.

## Backward compatibility

Fully additive. A config without `filters` behaves exactly as before: same 200 cap, same mandatory `queries`, same request shape.

https://claude.ai/code/session_01BBjnSErr9wEDmcdqYbeyHW

### DIFF
--- a/providers/wttj.mjs
+++ b/providers/wttj.mjs
@@ -8,16 +8,33 @@
 // hardcoded. The key is referer-locked, so every Algolia request sends a
 // welcometothejungle.com Referer header.
 //
-// The board is global and enormous, so a `wttj:` config block with explicit
-// search queries is REQUIRED — without one the provider throws rather than
-// silently scanning an arbitrary slice:
+// The board is global and enormous, so a `wttj:` config block that narrows it is
+// REQUIRED — without one the provider throws rather than silently scanning an
+// arbitrary slice. Narrow it with `filters`, with `queries`, or with both:
 //
 //   - name: Welcome to the Jungle
 //     provider: wttj
 //     wttj:
+//       # Algolia filter expression, applied server-side (recommended).
+//       filters: 'offices.country_code:FR AND contract_type:full_time'
 //       queries: ["finops", "data platform engineer", "snowflake"]
-//       max_hits: 100        # optional, per query, capped at 200
+//       max_hits: 100        # optional, per query; capped at 200, or 1000 with filters
 //     enabled: true
+//
+// Prefer `filters` over broad keyword queries. A keyword alone cannot narrow a
+// global board: it is Algolia's relevance ranking that decides which `max_hits`
+// results come back, so anything past the cap is invisible no matter how well it
+// matches the scanner's own title/location filters (those run afterwards, on what
+// already came back). Filtering server-side shrinks the result set instead of
+// re-ranking it, which is what makes a scan exhaustive rather than a sample.
+//
+// Useful faceted attributes on this index: offices.country_code, offices.state,
+// offices.city, contract_type, remote, experience_level_minimum,
+// salary_yearly_minimum, organization.name, and WTTJ's own job taxonomy
+// new_profession.sub_category_reference (e.g. "product-management-wNjYw").
+//
+// When `filters` is set, `queries` becomes optional — the empty query means
+// "everything that passes the filter", which is usually what you want.
 //
 // Each hit maps to the normalized Job shape; salary_yearly_minimum (when
 // present) is attached as `salary: {min, max, currency}` so scan.mjs's
@@ -28,6 +45,14 @@ const SITE_ORIGIN = 'https://www.welcometothejungle.com';
 const INDEX = 'wttj_jobs_production_en';
 const DEFAULT_MAX_HITS = 100;
 const MAX_HITS_CAP = 200;
+// An unfiltered keyword query matches a large slice of a global board — "product
+// manager" alone returns ~14k hits — so Algolia's own relevance ranking, not the
+// scanner's filters, decides which 200 are seen. A server-side `filters`
+// expression cuts the result set to something a single request can actually
+// exhaust (e.g. product-management + France + full_time is ~450), so the cap is
+// raised to Algolia's per-request ceiling for this index when one is configured.
+const FILTERED_MAX_HITS_CAP = 1000;
+const FILTERS_MAX_LEN = 1000;
 
 /** Pin a URL to an expected https host. */
 function assertHost(url, host, label) {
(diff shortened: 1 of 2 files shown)


luochen211 (commented):
Reviewed at head adad7f2 and replayed it cleanly onto current main (5c8b02a); the focused WTTJ test passes.

Two contract gaps still look blocking:

1. `resolveConfig()` keeps `DEFAULT_MAX_HITS` at 100 when `filters` is present and `max_hits` is omitted. That means the filter-only path described in this PR can still be silently partial: the documented ~450-hit example asks Algolia for only 100 rows. Raising the allowed cap to 1000 does not make the default filtered scan exhaustive. The tests cover an explicit `max_hits: 5000`, but the filter-only test never asserts `hitsPerPage`. Please either use 1000 as the filtered default, paginate using the response's count/page fields, or detect and clearly surface truncation; add a regression for the filter-only default.

2. The observable config contract is not synced. `docs/SUPPORTED_JOB_BOARDS.md` still says queries are required and the cap is 200, and `templates/portals.example.yml` does not document `filters` or the filtered cap. The provider guide explicitly requires those surfaces to stay in sync when an existing provider's config/defaults change.

The current Upgrade regression failure is stale-branch infrastructure: current main contains the confirmation-harness fix, and the local replay was clean. That is separate from the two gaps above.

Scott-Emberson (commented):
Reviewed at head adad7f21bd. In our gated area (providers/ plus tests/providers/). The WTTJ fetch surface stays safe: both request targets are host-pinned via assertHost with redirect:'error', the /api/env bootstrap on www.welcometothejungle.com and the Algolia POST on the derived https ${appId}-dsn.algolia.net host, and parseEnvPayload rejects a non-alphanumeric app id so the derived hostname cannot be injected. The new filters value is a POST-body Algolia DSL param, never a URL or host component, so it opens no new host surface. The test pins the filter pass-through, the empty-query default, per-query application, the cap contract (200 without a filter, raised to 1000 with one), a blank filter falling back to cap 200 with filters null, and an over-long filter throwing.

Two notes. First, the one red check is the system-layer Upgrade regression gate, not this PR: it fails on the updater confirm-flow (it expects apply to name the declaration file and the offending run-nightly.ps1 entry, while this branch still prints the older generic confirmation message), and this PR touches only providers/wttj.mjs and tests/providers/wttj.test.mjs, not update-system.mjs or upgrade-tests.mjs. The three test legs that run the provider suite are green, and a rebase onto current main clears the gate. Second, a design caveat worth flagging, not a test-soundness issue: "exhaustive" here is a single-request cap raise to 1000 relying on server-side filters to shrink the set, not a page-walk, so a filtered result set exceeding 1000 still truncates at Algolia's per-request ceiling with no pagination and no warning.

Owned test verified sound, no SSRF surface. Rebase to current main to clear the stale gate and I will re-verify and clear it to approve. The over-1000 truncation is worth a follow-up but is not blocking from the tests/ side.

luochen211 (changes requested):
Following the additional review, I am formalizing the two unresolved items from my earlier comment as blocking changes on the current head:

1. A filters-only config still defaults to `DEFAULT_MAX_HITS = 100`. The 1000 value is only a cap when `max_hits` is explicitly supplied; the existing test proves `max_hits: 5000` is capped to 1000 but never asserts the filters-only default. Therefore the documented ~450-hit filtered example still returns only 100 rows without any truncation signal.
2. The user-facing config contract remains unsynchronized: `docs/SUPPORTED_JOB_BOARDS.md` still requires queries and documents the old cap, while `templates/portals.example.yml` does not show `filters` or its cap/default behavior.

The latest test-axis review also notes that result sets above 1000 still truncate without pagination or warning. At minimum, fix the filters-only default and documentation; preferably detect/report any remaining server-side truncation. Please add a regression asserting `hitsPerPage` for `{ filters: FILTER }` with no `max_hits`.

providers/wttj.mjs
```diff
@@ -8,17 +8,34 @@
...
 //
-// The board is global and enormous, so a `wttj:` config block with explicit
-// search queries is REQUIRED — without one the provider throws rather than
-// silently scanning an arbitrary slice:
+// The board is global and enormous, so a `wttj:` config block that narrows it is
+// REQUIRED — without one the provider throws rather than silently scanning an
+// arbitrary slice. Narrow it with `filters`, with `queries`, or with both:
 //
 //   - name: Welcome to the Jungle
 //     provider: wttj
```
nikitacometa, line 16: Measured 2026-09-04: this filter matches `nbHits 52783` and, with no `max_hits`, the provider returns 100 jobs. As the header's example it teaches the worst case; the taxonomy filter (453 of 453) belongs here.

providers/wttj.mjs
```diff
@@ -28,6 +45,14 @@ const SITE_ORIGIN = 'https://www.welcometothejungle.com';
 const INDEX = 'wttj_jobs_production_en';
```
nikitacometa, line 45: `DEFAULT_MAX_HITS` stays 100 when a filter is set, so the filters-only path asks for 100 rows. Default to `cap` when `filters` is present, with a regression asserting `hitsPerPage` for `{ filters }` and no `max_hits`.

nikitacometa (changes requested):
Measured the live index on 2026-09-04 through the provider at adad7f21. The filter in the module header (16, `offices.country_code:FR AND contract_type:full_time`) matches `nbHits 52783`, not the 4,384 in the description's table, and `offices.country_code:FR` matches 66,858, not 5,091. Run through the provider, that documented example returns 100 jobs, and 1000 with `max_hits: 1000`. The provider sends one request with no `page` param and never reads `nbHits` or `nbPages`, and Algolia answers `nbPages: 1`, so a filtered scan is still the top N by relevance ranking: the cap moved from 200 to 1000 and the sampling did not change. CONTRIBUTING rule 3 asks providers to traverse the complete inventory. The taxonomy filter measures 453 of 453 and does work as described; the header should show that one.

`json.nbHits` is already in the response (238) and unused: warn when it exceeds `hits.length`, which turns a silent sample into a visible one and answers the standing review's truncation item. `DEFAULT_MAX_HITS` (45) stays 100 when a filter is set, so a filters-only config without `max_hits` asks for 100 rows; default to `cap` there and add the regression. docs/SUPPORTED_JOB_BOARDS.md:90 and templates/portals.example.yml:1360-1363 still say queries are required and the cap is 200.

The CI red is the stale `--confirm` leg fixed on main by #3755 (e979d4e0), absent from this branch; `upgrade-tests.mjs --pr-gate` is GREEN here. A rebase clears it.

PR #3759, opened 2026-09-03
fix(web): separate evaluation and application dates

theyv:
## Summary

- relabel the tracker Date column as **Eval date**
- add a separately sortable **Applied date** column to every tracker view
- resolve submission dates from the first transition into `Applied` in `data/status-log.tsv`
- fall back to an explicit `Applied YYYY-MM-DD` note for pre-ledger installations, and show an em dash when the date is unknown

## Why

The web pipeline previously labeled and sorted the tracker’s evaluation date as `date`, including on the Applied tab. This made offers appear to have been submitted on the day they were evaluated.

## Validation

- `npm test` in `web/`: 461 passed, 6 skipped, 0 failed
- `npm run typecheck` in `web/`
- `npm run build` in `web/`
- `node test-all.mjs`: 8096 passed, 0 failed

### DIFF
--- a/web/src/lib/application-dates.mjs
+++ b/web/src/lib/application-dates.mjs
@@ -1,0 +1,41 @@
+/**
+ * Read the first transition into Applied for every tracker row.
+ * The status ledger records event dates; the tracker Date column records evaluation dates.
+ *
+ * @param {string | null | undefined} tsv data/status-log.tsv content
+ * @returns {Map<string, string>} tracker number to YYYY-MM-DD application date
+ */
+export function parseApplicationDatesFromStatusLog(tsv) {
+  const applicationDates = new Map();
+  for (const rawLine of String(tsv ?? "").split("\n")) {
+    if (!rawLine || rawLine.startsWith("#")) continue;
+    const [trackerNumber, eventDate, , toStatus] = rawLine.split("\t").map((cell) => cell.trim());
+    if (!/^\d+$/.test(trackerNumber) || !isRealApplicationDate(eventDate)) continue;
+    if (toStatus.toLowerCase() !== "applied" || applicationDates.has(trackerNumber)) continue;
+    applicationDates.set(trackerNumber, eventDate);
+  }
+  return applicationDates;
+}
+
+/**
+ * Resolve an application's submission date without falling back to its evaluation date.
+ * The status ledger wins; notes support installations whose status history predates the ledger.
+ *
+ * @param {{n: string, notes?: string}} application tracker application
+ * @param {Map<string, string>} statusLogDates parsed status-ledger dates
+ * @returns {string} YYYY-MM-DD or an empty string when the submission date is unknown
+ */
+export function resolveApplicationAppliedDate(application, statusLogDates) {
+  const fromStatusLog = statusLogDates.get(application.n);
+  if (fromStatusLog) return fromStatusLog;
+
+  const match = String(application.notes ?? "").match(/\bapplied\s+~?(\d{4}-\d{2}-\d{2})(?![\w-])/i);
+  return match && isRealApplicationDate(match[1]) ? match[1] : "";
+}
+
+/** True when a value is a real YYYY-MM-DD application date. */
+function isRealApplicationDate(value) {
+  if (!/^\d{4}-\d{2}-\d{2}$/.test(value)) return false;
+  const parsed = new Date(value);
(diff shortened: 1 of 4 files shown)


web/src/lib/application-dates.mjs
```diff
@@ -0,0 +1,41 @@
...
+ * @param {string | null | undefined} tsv data/status-log.tsv content
+ * @returns {Map<string, string>} tracker number to YYYY-MM-DD application date
+ */
+export function parseApplicationDatesFromStatusLog(tsv) {
+  const applicationDates = new Map();
+  for (const rawLine of String(tsv ?? "").split("\n")) {
+    if (!rawLine || rawLine.startsWith("#")) continue;
+    const [trackerNumber, eventDate, , toStatus] = rawLine.split("\t").map((cell) => cell.trim());
+    if (!/^\d+$/.test(trackerNumber) || !isRealApplicationDate(eventDate)) continue;
+    if (toStatus.toLowerCase() !== "applied" || applicationDates.has(trackerNumber)) continue;
```
nikitacometa, line 14: Three things on this line. `toStatus` can be undefined: a line with fewer than 4 tabs passes the guard above and throws (funnel-velocity.mjs:115 rejects `cells.length < 5`). First-wins ignores `source`: per DATA_CONTRACT.md:47 a later `correction` replaces the date and `to: "-"` retracts it, both dropped here. And the literal `"applied"` misses the states.yml:25 aliases (`sent`, `aplicado`, `enviada`).

web/src/lib/application-dates.mjs
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @param {{n: string, notes?: string}} application tracker application
+ * @param {Map<string, string>} statusLogDates parsed status-ledger dates
+ * @returns {string} YYYY-MM-DD or an empty string when the submission date is unknown
+ */
+export function resolveApplicationAppliedDate(application, statusLogDates) {
+  const fromStatusLog = statusLogDates.get(application.n);
+  if (fromStatusLog) return fromStatusLog;
+
+  const match = String(application.notes ?? "").match(/\bapplied\s+~?(\d{4}-\d{2}-\d{2})(?![\w-])/i);
```
nikitacometa, line 32: Same regex as followup-cadence.mjs:219 minus `isCrossReferencedMention` (#2607): a sibling row's date in the notes is reported as this row's, and `Re-applied 2026-08-27` matches too. Import `parseAppliedDate`.

web/src/lib/application-dates.mjs
```diff
@@ -0,0 +1,41 @@
...
+  return match && isRealApplicationDate(match[1]) ? match[1] : "";
+}
+
+/** True when a value is a real YYYY-MM-DD application date. */
+function isRealApplicationDate(value) {
```
nikitacometa, line 37: Identical to the exported `isRealISODate` in followups.ts:127.

nikitacometa (changes requested):
Right shape: no tracker column, the status-log already answers the question, and the mutant that returns the eval date on a notes miss reddens the test. Five things in `application-dates.mjs`, all measured at 37773cdc.

Line 14 dereferences `toStatus` after line 13 guards only cells 1-2. A ledger line `40\t2026-09-02\tEvaluated` throws `TypeError: Cannot read properties of undefined (reading 'toLowerCase')`, and `readApplications()` has no catch, so every page that reads the tracker 500s. The ledger is hand-appendable (DATA_CONTRACT.md:47); funnel-velocity.mjs:115 rejects `cells.length < 5` instead.

Line 14 is first-wins and ignores `source`. DATA_CONTRACT.md:47 and `foldObservations` define a later `correction` as replacing the date and `to: "-"` as retracting it. Measured: a correction to 2026-08-15 still displays 2026-09-02, and a retraction still displays it.

Line 14 compares to the literal `applied`; templates/states.yml:25 declares aliases (`sent`, `aplicado`, `enviada`, `aplicada`, `başvuruldu`). `sent` gives an empty cell here and Applied in the funnel.

Line 32 duplicates followup-cadence.mjs:219 without `isCrossReferencedMention` (#2607): a note `#154 Sr PM M&A is already live in the same ATS (applied 2026-08-04)` displays 2026-08-04 for this row while follow-up-card.tsx:72 shows the eval-date fallback for the same row. Import `parseAppliedDate`.

Line 37 duplicates the exported `isRealISODate` in followups.ts:127. Reusing `parseStatusLog` + `foldObservations` from the core would close the first three at once.

PR #3761, opened 2026-09-03
Fix merge-tracker help and external data hooks

theyv:
## Summary

- make `merge-tracker.mjs --help` and `-h` print usage and exit before any tracker, queue, archive, or lock mutation
- separate the career-ops code root from the external data root when resolving post-merge executables
- make `sync-pdf-flags.mjs` resolve tracker data through the standard DATA_ROOT contract
- add CLI regression coverage for non-mutating help and post-hook execution with an external DATA_ROOT

## Root cause

`merge-tracker.mjs` used the result of `getCareerOpsRoot()` for both user data and executable scripts. With `.career-ops-data` or an environment override, that made the post-hook path point into the external data directory. The PDF sync script also resolved its tracker from its own install directory rather than through `getCareerOpsRoot()`.

The help flags were never handled, so they fell through to the normal merge path.

## Tests

- `node --test tests/merge-tracker-cli-roots.test.mjs`
- `node test-all.mjs --only merge-tracker-cli-roots`
- `npm run lint` (655 files)
- `node test-all.mjs` (8096 passed, 0 failed; 13 existing diagnostic warnings)

### DIFF
--- a/merge-tracker.mjs
+++ b/merge-tracker.mjs
@@ -11,7 +11,7 @@
  * If duplicate with higher score → update in-place, update report link
  * Validates status against states.yml (rejects non-canonical, logs warning)
  *
- * Run: node career-ops/merge-tracker.mjs [--dry-run] [--verify]
+ * Run: node merge-tracker.mjs [--dry-run] [--verify]
  */
 
 import { readFileSync, readdirSync, mkdirSync, renameSync, existsSync } from 'fs';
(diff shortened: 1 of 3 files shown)


Scott-Emberson (commented):
Reviewed at head e5996e190a. The owned test covers both fixed behaviors and the data-integrity property holds. The --help and -h handling is placed above the DATA_ROOT and mkdirSync and merge path and exits 0 before any write, and the test asserts help exits 0 with a Usage line while the tracker and the addition stay byte-identical and the merged/ directory is not created, so help cannot accidentally merge or create dirs. The external-data-hook fix is exercised with CAREER_OPS_ROOT pointed at a temp data root: it seeds a tracker with a pending flag plus an external pdf-index, asserts no "Failed to sync PDF flags" warning, and confirms the flag flips, which reddens on the old behavior where the sync script was resolved under the data root and hit ENOENT. The env is fully isolated to the temp data root.

The substance is merge-tracker.mjs, outside our owned paths, so this is a test-axis review: the owned test is sound and data-safe (help short-circuits before any write, the split code and data roots read and write the configured tracker), the merge is the maintainer's call. Nothing blocking from the tests/ side.

tests/merge-tracker-cli-roots.test.mjs
```diff
@@ -0,0 +1,88 @@
...
+    writeFileSync(
+      tracker,
+      `${TRACKER_HEADER}\n| 7 | 2026-01-04 | Acme | Engineer | 4.2/5 | Evaluated | ❌ | [12](../reports/012-acme.md) | seeded |\n`,
+    );
+    writeFileSync(
+      join(dataRoot, 'data', 'pdf-index.tsv'),
+      '# report\tpdf\thtml\tformat\tdate\n012\toutput/cv-acme.pdf\toutput/cv-acme.html\tletter\t2026-01-04\n',
+    );
+
+    const result = runMergeTrackerCli(dataRoot);
```
nikitacometa, line 79: With no pending TSV the run exits at merge-tracker.mjs:1094, before the post-hook at 1589; both assertions below hold on the base commit (measured). Write one addition into `<dataRoot>/batch/tracker-additions/` first.

tests/merge-tracker-cli-roots.test.mjs
```diff
@@ -0,0 +1,88 @@
...
+    const trackerAfter = readFileSync(tracker, 'utf-8');
+
+    assert.equal(result.status, 0, result.output);
+    assert.doesNotMatch(result.output, /Failed to sync PDF flags/);
+    assert.match(trackerAfter, /\| ✅ \| \[12\]\(\.\.\/reports\/012-acme\.md\) \|/);
```
nikitacometa, line 84: This ✅ comes from the inline sync at merge-tracker.mjs:503, not the spawned sync-pdf-flags.mjs, so it cannot tell the two roots apart.

merge-tracker.mjs
```diff
@@ -32,22 +32,39 @@ import { resolveTrackerPath, resolveWorkspaceRoot, resolvePdfIndexPath, trackerL
...
+  --migrate        Rewrite legacy report links relative to the tracker
+  --migrate-via    Add the Via column to a legacy tracker
+  --backfill-urls  Add the URL column and populate it from report metadata
+  -h, --help       Show this help and exit`);
+  process.exit(0);
+}
+
+// Executable hooks live beside this script even when user data is redirected
+// through CAREER_OPS_ROOT / CAREER_OPS_DATA_DIR / .career-ops-data.
+const CAREER_OPS_CODE_ROOT = dirname(fileURLToPath(import.meta.url));
```
nikitacometa, line 51: Same split still missing at invite-match.mjs:763, which joins `set-status.mjs` onto `getCareerOpsRoot()`; probe under `CAREER_OPS_ROOT` throws `Cannot find module <root>/set-status.mjs`.

merge-tracker.mjs
```diff
@@ -32,22 +32,39 @@ import { resolveTrackerPath, resolveWorkspaceRoot, resolvePdfIndexPath, trackerL
...
+  console.log(`Usage: node merge-tracker.mjs [options]
+
+Options:
+  --dry-run        Preview the merge without writing files
+  --verify         Run pipeline verification after a successful merge
+  --migrate        Rewrite legacy report links relative to the tracker
+  --migrate-via    Add the Via column to a legacy tracker
+  --backfill-urls  Add the URL column and populate it from report metadata
+  -h, --help       Show this help and exit`);
+  process.exit(0);
```
nikitacometa, line 46: An `Environment:` block for `CAREER_OPS_ROOT`, `CAREER_OPS_DATA_DIR`, `.career-ops-data`, `CAREER_OPS_TRACKER`, `CAREER_OPS_ADDITIONS`, `CAREER_OPS_BATCH_STATE` would make the help say what the title promises.

nikitacometa (changes requested):
The fix is right and measured at e5996e19: with one pending TSV under an external `CAREER_OPS_ROOT`, base emits `⚠️ Failed to sync PDF flags … MODULE_NOT_FOUND`, head does not, and `--help` now exits before the lock (base printed "Tracker merge lock acquired" on `--help`).

The second test does not cover it. tests/merge-tracker-cli-roots.test.mjs:65 leaves `batch/tracker-additions` empty, so the run exits at merge-tracker.mjs:1094 ("No pending additions to merge"), about 490 lines before the `execFileSync` at 1589. The ❌ to ✅ it asserts comes from the inline sync at merge-tracker.mjs:503, and `doesNotMatch(/Failed to sync PDF flags/)` is vacuous because the warning site is never reached. Reverting both product files to the base: the help test fails, the external-root test passes. Seed one addition TSV and it reddens.

invite-match.mjs:763 still joins `set-status.mjs` onto `getCareerOpsRoot()` (45); under `CAREER_OPS_ROOT` it throws `Cannot find module <root>/set-status.mjs`. #3715's review lists six more scripts with the same class (set-status.mjs:263 among them), and sync-pdf-flags.mjs is one this PR fixes, so the two overlap; one pass over the list would close it.

The help block omits the inputs that decide which tracker is rewritten (`CAREER_OPS_ROOT`, `CAREER_OPS_DATA_DIR`, `.career-ops-data`, `CAREER_OPS_TRACKER`, `CAREER_OPS_ADDITIONS`, `CAREER_OPS_BATCH_STATE`), and the sync-pdf-flags.mjs change has no test of its own (its 7 pass identically before and after).

PR #3763, opened 2026-09-03
feat(cv): optional professional-title headline slot under the name

Zoubeir23:
## Problem

The CV template has no place for a job title. `config/profile.yml` carries `candidate.full_name` but nothing for the role, and **none of the 8 shipped templates render one**. An ATS reads the title first to place a candidate — a CV that shows only a name forces the reader to infer the role from the summary prose. Every mainstream résumé template (moderncv, the common HTML resume kits) has this line.

## Change

Adds an **optional** `candidate.title` field to the render payload. Fully backward-compatible: absent or blank → no element, existing payloads render byte-identical.

- **`templates/cv-template.html`** — a `{{TITLE_BLOCK}}` slot right after `<h1>{{NAME}}</h1>`, a `.header-title` rule (accent colour, 16px / ~55% of the name, uppercase, tracked), and `.header-title` added to every `html[lang=…]` CJK/RTL font-override group so a localized title gets the right face.
- **`build-cv-html.mjs`** — `buildTitle()` trims, HTML-escapes, and emits nothing for an absent/blank/whitespace-only title. Filled next to the `{{PHOTO}}` slot, before the unresolved-placeholder guard.
- **The 6 alternate templates** (compact / executive / jake / leadership / modern / zh-minimal) each get the same slot plus a `.header-title` rule scaled to that template's own header style (its accent colour, case, tracking).
- **`modes/pdf.md`** — documents the field in the JSON example and the field-reference table: read from `config/profile.yml`, tailor to the JD's title wording only when the candidate's real level supports it, never inflate.

## Tests — `tests/cv-title.test.mjs`

- a `candidate.title` renders a `.header-title` element directly after the name
- absent / empty / whitespace-only → no element, no unresolved `{{TITLE_BLOCK}}`
- `<script>` / `&` / `"` in the title are escaped (no markup injection through the headline)
- all 8 built-in templates carry and resolve the slot

## Test plan

- [x] `node build-cv-html.mjs --test`
- [x] `node --test tests/cv-title.test.mjs` — 4/4
- [x] `node --test` on the CV/template suites (section-order, optional-sections, named-templates, template-packs, cover-unresolved-placeholders, profile-photo, dollar-pattern, doctor-unfilled-templates) — all green
- [x] `node test-all.mjs` — 8485 passed, 0 failed **on the change**. One unrelated red on this machine: `update-system.mjs check` gets SIGTERM'd on the harness's 30s budget under a slow link to GitHub (the command itself exits 0 in ~3s standalone); this PR touches nothing it calls.
- [ ] `test:cv-visual` — not runnable locally (`pdftotext` absent). Titleless fixtures render identically (the empty slot collapses to whitespace, the unused `.header-title` rule has no effect), so the pixel baselines are unaffected.

## Summary by CodeRabbit

* **User feature**: Set `candidate.title` for PDF CVs. The title appears below the candidate name in every built-in template, including ATS output.
* **Rendering**: Missing, empty, or whitespace-only titles produce no markup and preserve byte-identical output. Other titles are trimmed and HTML-escaped (`build-cv-html.mjs:634-689`).
* **Templates**: Added `{{TITLE_BLOCK}}`, `.header-title`, and localized font support across all CV templates (`templates/cv-template.html:87,672`).
* **Documentation and examples**: Added title guidance and a profile example (`modes/pdf.md:198`, `config/profile.example.yml:8`). LaTeX output does not support `candidate.title` (`modes/latex.md:83`).
* **Tests**: Added coverage for rendering, escaping, empty values, placeholder removal, and all built-in templates (`tests/cv-title.test.mjs:50-88`).
* **Documentation issue**: `modes/pdf.md:198` says to read `candidate.title` from `config/profile.yml`. The profile field is `candidate.headline`; only the render payload uses `candidate.title`.
* **System files**: No changes affect `AGENTS.md`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, or `.github/`. The only system documentation change is under `modes/`.


Scott-Emberson (commented):
Reviewed at head d75df17cdf. Since this adds a headline slot to the CV, I checked the source-of-truth boundary and it is clean: the title comes only from candidate.title (trimmed, with no default or fallback), and an empty value produces no element, so there is no fabricated or invented title. The owned test asserts a configured title renders the header-title div in the right slot (immediately after the name heading), while undefined, empty, and whitespace-only values all produce no headline and leave no unresolved {{TITLE_BLOCK}} placeholder, across every built-in template, and an escaping test blocks markup injection through the title. A fabricated default would trip the empty cases and a misplaced or unresolved slot would trip the ordering and placeholder assertions.

The substance is the CV builder and templates, outside our owned paths, so this is a test-axis review: the owned test is sound and the no-fabrication boundary is genuinely pinned, the merge is the maintainer's call. Nothing blocking from the tests/ side.

tests/cv-title.test.mjs
```diff
@@ -0,0 +1,66 @@
...
+import { tmpdir } from 'node:os';
+import { dirname, join, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const ROOT = resolve(dirname(fileURLToPath(import.meta.url)), '..');
+const TEMPLATE_DIR = join(ROOT, 'templates');
+const BASE_TEMPLATE = join(TEMPLATE_DIR, 'cv-template.html');
+
+const ALL_TEMPLATES = readdirSync(TEMPLATE_DIR)
+  .filter((f) => /^cv-template(\.[a-z-]+)?\.html$/.test(f))
```
nikitacometa, line 14: This regex matches 7 files, not 8: `resume-template.html` is excluded, and so is the ats pack. Enumerate the way `cv-templates.mjs` does, or hardcode the paths like tests/theme-style.test.mjs in #3789.

build-cv-html.mjs
```diff
@@ -665,6 +674,7 @@ function renderHtml(template, payload, templatePath) {
   // no <img>), so they are rebuilt as whole blocks before placeholder fill.
   let html = template.replace(CONTACT_ROW_RE, () => buildContactRow(candidate));
   html = html.replace(/\{\{PHOTO\}\}/g, () => buildPhoto(candidate, candidate.name));
+  html = html.replace(/\{\{TITLE_BLOCK\}\}/g, () => buildTitle(candidate));
```
nikitacometa, line 677: An absent slot is silent; only a leftover `{{…}}` trips the guard. `resume-template.html` has no `{{TITLE_BLOCK}}`, so the title is dropped with exit 0.

modes/pdf.md
```diff
@@ -193,6 +194,7 @@ Write a JSON file with this structure, then run `node build-cv-html.mjs <input.j
 | `lang` | string | CV language code (`en`, `es`, `zh-CN`, `ja`, `ar`). Drives language-specific CSS: `zh-CN` enables Simplified Chinese fonts and strict CJK line breaking; `ja` enables a Japanese CJK font fallback; `ar` enables RTL + Arabic fonts. Defaults to `en`. |
 | `page_format` | string | `letter` → `8.5in` page width, `a4` → `210mm`. Defaults to `letter`. Pass the SAME value to `generate-pdf.mjs --format`. |
 | `candidate.name` | string | From `profile.yml`. |
+| `candidate.title` | string | Optional professional headline shown directly under the name (e.g. "Senior Backend Engineer"). Read it from `candidate.title` in `config/profile.yml`. An ATS reads this first to place the candidate; a CV with no title forces the reader to infer the role. Omit or leave empty → no title element, byte-identical to before. Tailor it to the JD's own title wording when the candidate's real level supports it (never inflate). |
```
nikitacometa, line 197: config/profile.example.yml documents no `candidate.title`; add a commented key there, or the agent infers the level from cv.md prose and nothing gates it.

templates/cv-template.zh-minimal.html
```diff
@@ -73,6 +73,17 @@
     line-height: 1.1;
   }
 
+  .header-title {
+    font-family: 'Liberation Sans', 'Helvetica Neue', Arial, 'DejaVu Sans', sans-serif;
```
nikitacometa, line 77: Latin-only stack, and the `html[lang="ar"]` selectors at 456-462 cover `.header h1` but not `.header-title`, so a non-Latin headline loses the locale font the h1 keeps.

nikitacometa (changes requested):
Escape and slot mutants both redden `test-all.mjs --only cv-title`, absent and blank titles leave no element and no leftover placeholder, and the fact gate passes a rendered file. Four things at d75df17c.

`templates/resume-template.html` gets no `{{TITLE_BLOCK}}`. Rendering `{"candidate":{"title":"Senior Backend Engineer"}}` through it exits 0 with `grep -c header-title` = 0, no warning (an absent slot is silent at build-cv-html.mjs:677; only a leftover `{{…}}` trips the guard); the same payload through `cv-template.html` gives 8 hits. It is a maintained template (asserted in cv-optional-sections, cv-section-order, profile-photo and several test-all blocks), and #3776 and #3789 both patch it this round. tests/cv-title.test.mjs:14 filters `/^cv-template(\.[a-z-]+)?\.html$/`, which is 7 files, so the suite asserts "all 8" over 7 and cannot see this; it also skips the ats pack.

modes/pdf.md says to read `candidate.title` from `config/profile.yml`, but config/profile.example.yml:5-15 adds no `title` key, and `verify-cv-facts.mjs` does not gate a title (a headline absent from cv.md passes). With no user-layer key to read, the agent has to infer a seniority label from cv.md prose, and "never inflate" has nothing behind it. A commented `# title:` under `candidate:` in the example fixes it.

cv-template.zh-minimal.html:77 gives `.header-title` a Latin-only `font-family`, and the `html[lang="ar"]` block at 456-462 lists `.header h1` and the other header classes but not this one, so a Chinese, Japanese or Arabic headline in that template loses the locale font stack the h1 keeps.

`build-cv-latex.mjs` has no title concept, so `cv.output_format: latex` users get nothing; worth a line in the description.

Issue #3736, opened 2026-09-03
test(harness): move five of the eight root -tests.mjs suites into tests/ (#3306 Part 2 residual)


vliggio:
## Problem

#3306 Part 2 moved the nine root-adjacent `*.test.mjs` suites into `tests/` and deleted the registration list that named them. Eight suites remain at the root under a different naming convention, `*-tests.mjs`, and seven of them are still hand-registered in the `scripts` list at `test-all.mjs:356-361` and `:376`.

That convention was never in scope for #3306 — its table counted `*.test.mjs` only. Five of the eight can follow the nine into `tests/`; three cannot, and the reasons are worth recording so this is not re-costed a third time.

## The five that can move

`agent-inbox-tests.mjs` · `followup-seed-tests.mjs` · `paste-reply-tests.mjs` · `set-status-tests.mjs` · `tracker-columns-tests.mjs`

I expected the `scripts` section's throwaway repo copy (`scriptTmp`, `test-all.mjs:439`) to be load-bearing for these — it hands each script a clean `data/applications.md` — and it is not. All five build their own `mkdtempSync` sandboxes and point the CLI under test at them via `CAREER_OPS_TRACKER`. The copy is belt-and-braces, so in-process discovery does not take anything away.

The work per file is the #3388 refactor, not a `git mv`:

- Drop the private counter/`process.exit(1)` harness for `pass`/`fail` from `tests/helpers.mjs`. This is mandatory, not stylistic: `test-all.mjs:140` refuses to import any discovered suite containing `process.exit(` — it would terminate test-all mid-run with a forged exit code. All five trip it today.
- Repoint sibling paths to `..`, the same two path idioms #3388 dealt with.
- Delete the `scripts` entry.
- Drop the now-redundant `SYSTEM_PATHS` entries — `update-system.mjs:270`, `:335`, `:353`, `:364`, `:351` — since `'tests/'` at `:294` already covers the directory.

Suggested order: `paste-reply-tests.mjs` first. At 209 lines with no self-references it is the cheapest pilot, and it proves the sandbox reasoning above before the 1418-line `set-status-tests.mjs` is touched.

## The three that should stay

[...]

PR #3764, opened 2026-09-03
test(harness): guard the *-tests.mjs root convention against an unrun suite

### LINKED ISSUE #3735: test(harness): the -tests.mjs root convention is a hand-registered list that no guard covers (#3306 residual)
## Problem

`tests/no-root-suites.test.mjs` (#3388) guards the doctrine that no test suite sits at the repo root. It matches `*.test.mjs` — and that is the right pattern for what it was written to catch.

Eight suites at the root end in `-tests.mjs` instead, and no guard matches them:

| file | added |
|---|---|
| `updater-migration-tests.mjs` | 2026-06-11 (#833) |
| `tracker-columns-tests.mjs` | 2026-06-13 (#954) |
| `agent-inbox-tests.mjs` | 2026-07-03 (#1472) |
| `followup-seed-tests.mjs` | 2026-07-04 (#1468) |
| `set-status-tests.mjs` | 2026-07-07 (#1460) |
| `paste-reply-tests.mjs` | 2026-07-13 (#1803) |
| `tracker-writer-lock-tests.mjs` | 2026-07-21 (#1912) |
| `upgrade-tests.mjs` | 2026-08-02 (#2358) |

Seven of them are registered by hand, one entry per file, in the `scripts` list at `test-all.mjs:356-361` and `:376`. That is the shape #3306 set out to remove — "a list is a thing you can forget" — and it survived the series intact, because the `scripts` list could not be deleted wholesale: it also carries ~40 `--self-test` CLI invocations that have nothing to do with this.

So today a contributor who adds `foo-tests.mjs` at the root and forgets the `scripts` entry gets exactly #3303: a suite that is distributed and never executed, reporting nothing, indistinguishable from green. `tests/no-root-suites.test.mjs` stays green through it, because the filename does not end in `.test.mjs`.

## This is not an oversight in #3388

The guard's header reasons about the boundary explicitly:

> `*.test.mjs` specifically, NOT "anything test-shaped". test-salary-filter.mjs and test-trust-validator.mjs sit at the root and are correctly registered in test-all.mjs; a looser pattern would redden on two files that are fine. #3411 moves them into tests/, after which this check reads the same either way.

That reasoning is about the `test-*.mjs` **prefix**, and it resolved correctly — #3411 moved both, and `test-all.mjs` is the only `test-*.mjs` left at the root. The `-tests.mjs` **suffix** is a third convention that was never in scope for that discussion, and it is the one still carrying a hand-maintained list.

## Why not just widen the existing guard

Because these eight are not strays. They are correctly registered and they run — I confirmed each one against `test-all.mjs`. Three of them also cannot move (see the companion issue): `upgrade-tests.mjs` is a flag-driven CI harness, `updater-migration-tests.mjs` asserts on its own filename, `tracker-writer-lock-tests.mjs` needs the `timeoutMs: 180_000` that only the `scripts` list can express. Widening `no-root-suites.test.mjs` to `-tests.mjs` would redden on files that are fine — the precise failure its header rejects.

The property to check is different, and it is the one the deleted `tests/root-suite-registration.test.mjs` was checking before #3388 retired it: **is every root-level `*-tests.mjs` reachable by something that runs it?**

## Proposal

A guard over the residual convention, in the shape of the old one but scoped to what actually remains:

1. Glob `*-tests.mjs` at ROOT (`isFile() || isSymbolicLink()`, per #3140/#3364).
2. Assert the list is non-empty, so the guard fails loudly if the convention is ever fully retired rather than passing vacuously — the assertion-1 pattern @artemtrofymenko argued for on #3306, which is what would tell a future mover to delete this file.
3. Assert each is named in executable code in `test-all.mjs` — comments stripped, name required to sit inside quotes. Both narrowings already exist in the deleted guard and were added for the CodeRabbit finding on #3303/#3305: a filename surviving in a comment after its invocation is gone must not read as registered.
4. Exempt `upgrade-tests.mjs`, which is invoked by `.github/workflows/test.yml:89-93` rather than by `test-all.mjs`, with the reason inline.

`git show f9139b2 -- tests/root-suite-registration.test.mjs` recovers the original; steps 1-3 are that file with the pattern changed.

## Alternative, if this is judged the wrong direction

Retiring the convention entirely is the other consistent answer — but it does not work here. The three files named above have concrete reasons to stay at the root, so `-tests.mjs` is a permanent residual rather than migration debt, and a permanent hand-maintained list is the thing #3306 was about.

## Not blocked by, and does not block, the move

The companion issue proposes moving five of the eight into `tests/`. That shrinks the guarded set from seven to two but does not empty it, so this guard is worth having either way, and the two can land in either order.

**Area:** Testing / CI

vliggio:
## What does this PR do?

Adds `tests/root-tests-registration.test.mjs`, which asserts that every root-level `*-tests.mjs` is reachable by something that runs it — quoted in `test-all.mjs`, or invoked by a workflow. Corrects the `ARCHITECTURE.md` sentence claiming no suite is registered by hand any more.

## Related issue

Fixes #3735

## The gap

#3388 emptied the `*.test.mjs` half of the root and deleted the list naming those nine suites; `tests/no-root-suites.test.mjs` now keeps the root empty of that pattern.

The `*-tests.mjs` half was never in scope for that series. Eight such suites remain at the root, seven of them named one by one in `test-all.mjs`'s `scripts` list — the hand-maintained list #3306 set out to remove, which survived because `scripts` also carries ~40 `--self-test` CLI entry points and could not be deleted wholesale. A new `foo-tests.mjs` forgotten from it runs nowhere and reddens nothing: #3303 exactly, in a spelling no guard matches.

## Why a second guard rather than widening `no-root-suites.test.mjs`

That file asks *"is there a root suite at all?"*, and for `*-tests.mjs` the answer is a permanent yes. Three of the eight cannot move:

| file | why it stays |
|---|---|
| `upgrade-tests.mjs` | flag-driven CI harness; bare invocation prints usage and exits 1. Run three ways from `.github/workflows/test.yml:89-93`, deliberately outside `test-all.mjs` |
| `updater-migration-tests.mjs` | `git add`s itself as a fixture and asserts its own filename is in `SYSTEM_PATHS`/`BOOTSTRAP_PATHS` |
| `tracker-writer-lock-tests.mjs` | `timeoutMs: 180_000` (#2906, SIGTERM'd on windows-latest at the default 30s). The discovery path has no timeout mechanism at all |

So a pattern widened to `-tests.mjs` would redden on files that are fine — the precise failure `no-root-suites.test.mjs`'s own header rejects. The property here is reachability, not location.

## No exemption list

The first draft of the issue exempted `upgrade-tests.mjs` by name. @artemtrofymenko's review is the reason it does not: an exemption list is a hand-maintained list of one, which is the shape being removed. Both surfaces that run things are read instead.

The match rule differs per surface, because the surfaces differ syntactically:

| surface | rule |
|---|---|
| `test-all.mjs` | name appears **quoted** in non-comment code — both narrowings from the deleted `tests/root-suite-registration.test.mjs`, verbatim |
| `.github/workflows/*.yml` | name is the argument of a **`node` invocation** |

A quoted rule cannot work on the second surface: in YAML the filename is a bare shell token (`run: node upgrade-tests.mjs --pr-gate`), so it would match nothing and every workflow-run suite would read as unreachable. Requiring the `node` verb is *tighter* than quoting rather than looser — a trailing `# see upgrade-tests.mjs` cannot satisfy it — so the mention-vs-invocation narrowing survives with no quote-aware YAML comment stripping.

`.github/` is in `SYSTEM_PATHS` (`update-system.mjs:432`), so the second surface exists in user installs too. An unreadable workflows dir is reported as a broken premise rather than read as "no invocations", which would fail a workflow-run suite for the wrong reason.

## Verified by making it red

```
as it stands
  ✅ ROOT is the repo root — test-all.mjs is a file there, so an empty result means empty
  ✅ 8 root-level *-tests.mjs found to check
  ✅ 19 workflow file(s) read as the second run surface
  ✅ every root-level *-tests.mjs is reachable — quoted in test-all.mjs, or invoked by a workflow (8 checked)
```

1. **new root suite registered nowhere** — `orphan-tests.mjs` added → ❌
2. **name present only in a `test-all.mjs` comment** → ❌
   Not hypothetical: reviewing #3735, a plain `grep -q` over each filename reported eight of eight registered; the eighth was `test-all.mjs:6549`, a comment.
3. **workflow surface removed** — `test.yml` hidden → ❌ on `upgrade-tests.mjs`
   This is the exemption's replacement doing the work: `upgrade-tests.mjs` is quoted nowhere in executable code in `test-all.mjs`, so it passes on the workflow surface alone.
4. **convention fully retired** — the eight moved out → ❌ `no root-level *-tests.mjs remains … DELETE this file; do not repoint it at another pattern.`

Assertion 2 is the vacuity guard, carrying the instruction for a future mover in the file rather than in a thread.

## Stated limitation

An indirect invocation — npm script, composite action, shell wrapper — matches neither rule and reports unreachable. Left as the safe direction and documented in the header: a false red is read and resolved by whoever added it, a false green is the bug this file prevents. If an indirect invocation ever lands, widening beats exempting.

## ARCHITECTURE.md

`:34` read *"No suite is registered by hand any more — `test-all.mjs`'s `SCRIPTS` list holds scripts and their `--self-test` entry points, not test files."* Seven `-tests.mjs` suites are registered by hand in exactly that list — the same overstatement shape #3411 corrected once already, and load-bearing here because it is the sentence that would tell a contributor this convention does not exist. Rewritten to describe the residual, why the three are permanent rather than migration debt, and which guard covers which pattern. No test asserts on that prose.

## Not in this PR

#3736 proposes moving five of the eight into `tests/`. That shrinks the set this guard covers from seven to two without emptying it, so the two are independent and can land in either order.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] I ran `node test-all.mjs` and all tests pass — 8093 passed, 0 failed, 14 warnings (all pre-existing: no Go compiler in env, maintainer domain in shipped files, opt-in live tests skipped)

## Summary

- Added `tests/root-tests-registration.test.mjs:1` to verify that every root-level `*-tests.mjs` suite is reachable.
- The guard checks executable registrations in `test-all.mjs` and qualifying workflow commands.
- It ignores comments, regex literals, indirect commands, and later commands in workflow steps.
- Updated `ARCHITECTURE.md:1` to document the remaining root suites and their guards.

## User impact

- A new root-level `*-tests.mjs` file must be registered or CI will fail.
- Existing test execution remains unchanged.
- No changes were made to `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, or `.github/`.


Scott-Emberson (commented):
Reviewed at head 43231d2d75. In our gated area (tests/). This guard is real, not a hardcoded list: it readdir's the repo root for -tests.mjs files and fails if any is neither registered in test-all.mjs nor invoked by a workflow, so dropping an unregistered root suite reddens. The registration check reads test-all.mjs from disk, strips block and whole-line comments, and requires the suite name to appear quoted rather than as a bare grep (the header notes a plain grep once false-passed on a comment), and the workflow check reads .github/workflows/*.yml for a node invocation. There is no by-name exemption list, and the silent-pass paths are all closed: a wrong ROOT, an unreadable readdir, an empty suite list, or an unreadable workflow dir each fail loudly rather than passing. It is correctly scoped to the -tests.mjs convention, distinct from and non-overlapping with the .test.mjs no-root-suites guard, and it degrades cleanly alongside the #3765 move (it still checks the residual root suites and fails loudly if the convention is ever fully retired).

Owned test verified sound. The only thing holding it is that the branch is behind main. Rebase to current main and I will re-verify and clear it to approve.

artemtrofymenko (commented):
The two-surface mechanism is right and the per-surface rules are right. One hole in the argument for skipping YAML comment stripping, measured on your branch at `43231d2`.

Your claim is half true. A trailing `# see upgrade-tests.mjs` genuinely cannot satisfy `node\s+<name>` — there is no `node` verb in front of it. But a **commented-out invocation** has one, and that is the likelier real-world form: someone disables a CI step for an afternoon and leaves the line in place.

```
.github/workflows/test.yml
  -      run: node upgrade-tests.mjs --pr-gate      (and --canary, --local-paths)
  +      # run: node upgrade-tests.mjs --pr-gate

live invocations: 0    commented: 3

node tests/root-tests-registration.test.mjs
  ✅ every root-level *-tests.mjs is reachable ... (8 checked)
```

Green, with the suite running nowhere. That is the mention-vs-invocation failure the harness surface strips comments to prevent, one surface over — and it lands on precisely the file the exemption was dropped for, so the guard would be silent about the case it exists to cover.

## The fix, verified both directions

```js
const workflowCode = workflowText
  .replace(/^[ \t]*#.*$/gm, ' ')   // whole-line comments
  .replace(/\s#.*$/gm, ' ');       // trailing comments
```
and match against `workflowCode`.

```
commented-out invocations →  ❌ 1 root-level suite(s) are never run ... upgrade-tests.mjs
unmodified tree           →  0 failures, no false red
```

Same two narrowings as the harness surface, which is a nicer symmetry than the asymmetry it replaces: both surfaces now strip comments, and the second additionally requires the verb.

One caveat I did not test: naive `#` stripping eats a `#` inside a quoted YAML string. For finding `node <file>` that can only produce a false red, which is the direction your header already chooses deliberately.

## Two things worth keeping as they are

**Assertion 1 carrying the deletion instruction in the file** is better than what I asked for on #3306 — I said the instruction should exist, you put it where the person who trips it is already looking.

**Requiring the `node` verb rather than quoting** is the right call for YAML and the reasoning is sound; the only correction is that it does not remove the need for comment stripping, it removes the need for *quote-aware* stripping. Plain stripping still earns its place.

Nothing else from me. On #3736 — yes, I will take it, and I will argue the three that stay rather than inherit them.

tests/root-tests-registration.test.mjs
```diff
@@ -0,0 +1,195 @@
...
+
+  if (suites && suites.length > 0) {
+    // ── Surface 1: the scripts list in test-all.mjs ──────────────────────────
+    // Deliberately NOT a parse of the `scripts` array. A suite reached by any
+    // mechanism in this file — that list, an inline run(), a future glob —
+    // names the file, and the question is "does anything run this", not "which
+    // section does".
+    const code = readFileSync(join(ROOT, 'test-all.mjs'), 'utf-8')
+      .replace(/\/\*[\s\S]*?\*\//g, ' ') // block comments
+      .replace(/^\s*\/\/.*$/gm, ' '); // whole-line comments
```
nikitacometa: Strips whole-line `//` only, so `const x = 1; // was 'zzz-tests.mjs'` makes an orphan `zzz-tests.mjs` read as registered (4 passed, 0 failed, 9 checked), the case the header at 44-51 says is closed.

tests/root-tests-registration.test.mjs
```diff
@@ -0,0 +1,195 @@
...
+    if (workflowErr) {
+      warnOrFailWorkflows(workflowErr);
+    } else {
+      pass(`${workflowsRead} workflow file(s) read as the second run surface`);
+    }
+
+    /** True when `name` is the target of a `node` invocation in any workflow. */
+    const invokedByWorkflow = (name) => {
+      const escaped = name.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+      return new RegExp(`node\\s+(?:\\.[\\\\/])?${escaped}(?=\\s|$)`, 'm').test(workflowText);
```
nikitacometa: Matches over the raw concatenation of every workflow file, so YAML comments and `echo` arguments count. Commenting out test.yml:89/91/93 leaves `upgrade-tests.mjs` green with nothing running it.

tests/root-tests-registration.test.mjs
```diff
@@ -0,0 +1,195 @@
...
+//     comment after its invocation is gone must not read as registered. That
+//     is not hypothetical here. Reviewing #3735, a plain `grep -q` over each
+//     filename reported eight of eight registered; the eighth was
+//     test-all.mjs:6549, a comment.
+//
+//   - workflows: the name must appear as the argument of a `node` invocation.
+//     In YAML it is a bare shell token (`run: node upgrade-tests.mjs
+//     --pr-gate`), so the quoted rule would match nothing at all and every
+//     workflow-run suite would read as unreachable. Requiring the `node` verb
+//     is also TIGHTER than quoting rather than looser: a trailing
```
nikitacometa: "TIGHTER than quoting" holds for `# see upgrade-tests.mjs`, not for `# node upgrade-tests.mjs`, which is what a disabled step looks like.

ARCHITECTURE.md
```diff
@@ -31,7 +31,7 @@ Tests are the one thing the flat root does not hold. Suites live in `tests/`, na
...
...m ([#3306](https://github.com/career-ops-hq/career-ops/issues/3306)). Almost no suite is registered by hand any more. The residual is the `*-tests.mjs` naming convention: eight such suites remain at the root, seven of them named one by one in `test-all.mjs`'s `scripts` list alongside the `--self-test` entry points. Three cannot move — a flag-driven CI harness (`upgrade-tests.mjs`, run from `.github/workflows/test.yml`), one that asserts on its own filename (`updater-migration-tests.mjs`), and one carrying a per-script timeout discovery cannot express (`tracker-writer-lock-tests.mjs`, [#2906](https://github.com/career-ops-hq/career-ops/issues/2906)) — so the convention is permanent rather than migration debt, and `tests/root-tests-registration.test.mjs` guards it: every root `*-tests.mjs` must be quoted in `test-all.mjs` or invoked by a workflow ([#3735](https://github.com/career-ops-hq/career-ops/issues/3735)). `tests/no-root-suites.test.mjs` separately keeps the root empty of `*.test.mjs`, where discovery does not reach. Playwright specs are the deliberate exception: `tests/cv-visual/` is named `*.spec.mjs`, which discovery does not match, and runs from `playwright.cv.config.mjs`.
```
nikitacometa, line 34: #3765 rewrites this same sentence; `git merge-tree --write-tree pr-3764 pr-3765` exits 1 with a content conflict here.

nikitacometa (dismissed):
Core property holds at 43231d2d: an orphan `zzz-tests.mjs` at the root gives exit 1 and "1 root-level suite(s) are never run", the ROOT sentinel and vacuity guard are right, and test-all.mjs:6549 reproduces as the comment-only mention. The two match rules do not hold. Four inputs read as reachable when nothing runs the file:

- comment out `.github/workflows/test.yml:89`, `:91`, `:93`, the three real `run: node upgrade-tests.mjs` steps, and nothing invokes it anywhere. Result: "every root-level *-tests.mjs is reachable (8 checked)", 4 passed, 0 failed, exit 0. That is a CI step disabled the normal way.
- a trailing `//` comment in test-all.mjs quoting the name (129 strips whole-line `//` only).
- a YAML comment `# node x-tests.mjs`.
- `run: echo "node x-tests.mjs"`.

All four go red on #3765's head, which is this same file after six more commits (+274/-47). Please cherry-pick that hardening here rather than shipping the superseded rule, or adopt its final form (`node <file>` must be the first command of a `run:` step, plus trailing-`//` stripping). A quoted name anywhere in executable code also counts as registration (133-141), so `const legacy = 'agent-inbox-tests.mjs'` keeps a removed `scripts` entry green.

Separately, both PRs rewrite ARCHITECTURE.md:34; `git merge-tree --write-tree pr-3764 pr-3765` exits 1 with a content conflict there and no main commit has touched the file since the fork point, so "either order" is not accurate: #3765 rebases after this lands.

tests/root-tests-registration.test.mjs
vliggio: Confirmed and fixed in 08de25b. Reproduced end to end on this branch first — orphan `zzz-tests.mjs` plus `const x = 1; // retired: 'zzz-tests.mjs'` gave "every root-level *-tests.mjs is reachable (9 checked)", green with nothing running it.

The strip-then-require-quotes approach is replaced by a scanner that collects string literals, so a comment holds no literals by construction and the whole-line-vs-trailing distinction stops mattering. It skips regex literals too, which is load-bearing here: `test-all.mjs` contains `/from ['"]node:test['"]/`, and treating that quote as a string opener would corrupt every answer below it.

tests/root-tests-registration.test.mjs
vliggio: Confirmed and fixed in 08de25b. Reproduced exactly as you and @artemtrofymenko described — commenting out test.yml:89/91/93 left "every root-level *-tests.mjs is reachable (8 checked)" with `upgrade-tests.mjs` running nowhere.

The surface now parses each workflow with js-yaml and reads only `jobs.<id>.steps[].run`, so YAML comments are dropped by the parser and an action input that happens to be named `run` is not a command. Within a step, `node <file>` must be its FIRST command, which also disposes of `echo "node x-tests.mjs"`. All four of your inputs redden now, and the baseline stays green with `upgrade-tests.mjs` still reachable through the workflow surface alone.

tests/root-tests-registration.test.mjs
vliggio: You and @artemtrofymenko are both right, and the claim was mine. "Tighter than quoting" holds for `# see upgrade-tests.mjs` — no `node` verb — and fails for `# node upgrade-tests.mjs`, which is what a step disabled for an afternoon looks like. I generalised from the weaker example.

The header no longer makes that argument: comments are dropped by the YAML parser rather than by a claim about verbs, and the surviving text says why the first command is the only position that cannot be shell data. Fixed in 08de25b.

ARCHITECTURE.md
vliggio: Confirmed — `git merge-tree` does report a content conflict there, and I do not think it is a defect to fix. That sentence states how many suites sit at the root, which is genuinely eight on this branch and three after #3765 moves five into `tests/`. Forcing the two into agreement would mean one of them stating a number that is wrong for its own tree.

It resolves by merge order instead: #3764 lands, #3765 rebases onto it, and #3765's version wins because by then the move has happened. I have kept this branch's copy accurate for this branch (eight root suites, seven registered) rather than pre-empting the other PR.

Flagging one related thing for whoever sequences these: #3765 currently sits on `main` and carries a rebased copy of this branch's original guard commit, so it will need a restack once this merges. That is expected for the stack, not a new problem.

artemtrofymenko (commented):
Re-ran this on default non-elevated Windows 11 at `08de25b`, since I flagged the trailing-comment hole and owe you the check on the platform CI cannot reproduce. **Both hardened rules hold here**, and one new finding below that is a misdiagnosis rather than a gate hole.

## The hardening holds, and the greens are discriminating

Baseline: `5 passed, 0 failed`, `20 workflow file(s) read as the second run surface (30 run: scripts)`.

| input | result |
| --- | --- |
| comment out `test.yml:89/91/93`, the three real `run: node upgrade-tests.mjs` steps | ❌ `1 root-level suite(s) are never run` |
| orphan `zzz-tests.mjs`, name hidden in a whole-line `// retired: 'zzz-tests.mjs'` | ❌ still red |
| same, hidden in a **trailing** comment: `const _x = 1; // retired: 'zzz-tests.mjs'` | ❌ still red |

That third row is the one I reported, and the literal scanner closes it — a comment holds no literals by construction, so the whole-line-vs-trailing distinction genuinely stops existing. Skipping regex literals is the right call too; `/from ['"]node:test['"]/` in `test-all.mjs` would otherwise have opened a string at that quote and corrupted every answer after it.

Three reds in a row would look identical if the guard had simply stopped matching anything, so I ran the control in both directions:

| control | result |
| --- | --- |
| orphan + a real literal `{ name: 'zzz-tests.mjs', expectExit: 0 }` | ✅ `reachable (9 checked)` |
| orphan + a real workflow step `- name: Orphan probe / run: node zzz-tests.mjs` | ✅ `reachable (9 checked, 31 run: scripts)` |

Both surfaces still say yes when the answer is yes.

## Finding: one unparseable workflow makes the guard accuse a suite by filename sort order

`tests/root-tests-registration.test.mjs:164-168` breaks out of the directory loop on the first YAML parse error:

```js
} catch (err) {
  workflowErr = `${entry.name}: ${err.message}`;
  break;
}
```

Every workflow after the broken one is then never read, but `unreachable` is still computed against the partial `runScripts` and reports as if it had been.

The same file, same content, two names:

```
zz-dup.yml   →  ❌ .github/workflows is unreadable (…)
                ✅ every root-level *-tests.mjs is reachable (8 checked)

aa-broken.yml → ❌ .github/workflows is unreadable (…)
                ❌ 1 root-level suite(s) are never run — nothing in test-all.mjs
                   or .github/workflows names them:
                       upgrade-tests.mjs
                   Add an entry to the `scripts` list: { name: '<file>', expectExit: 0 }
```

`upgrade-tests.mjs` is invoked three times in `test.yml`. It reads as unreachable only because `aa-` sorts before `test.yml` and `zz-` sorts after it.

Two things make this worth fixing rather than noting:

- **The remedy it offers is wrong and would stick.** Following it — adding `upgrade-tests.mjs` to the `scripts` list — turns the second assertion green while the broken workflow stays broken, and quietly gives a flag-driven harness a bare-invocation registration it was deliberately kept out of. Your ARCHITECTURE.md paragraph names that as one of the three that cannot move.
- **The trigger is not exotic.** My fixture is a duplicate `steps:` key, which **GitHub Actions accepts** (last wins) and js-yaml rejects outright. So a workflow that runs fine in CI can redden this suite locally and name the wrong file while doing it.

Fail-closed does hold — the run is red either way, so nothing ships unguarded, and `warnOrFailWorkflows` says plainly that the surface was not checked. This is purely about which file the developer is sent to. Two ways out, both small: `continue` instead of `break` so one bad file costs only its own contribution, or skip the `unreachable` assertion entirely when `workflowErr` is set, since the surface is known incomplete and any verdict from it is unsound.

I lean to the second — a partial surface can still produce a *false green* (a suite invoked only in the broken file, where another file happens to name it), and suppressing the verdict says what is true: this run cannot answer the question.

## Not a finding, recorded so nobody re-derives it

My first control-probe attempt inserted `run: node zzz-tests.mjs` as a second `run:` key inside an existing step, which is a duplicate key — js-yaml threw, and I briefly read the result as the guard failing on a valid workflow. It was my fixture. The well-formed version is the ✅ row above. Mentioning it because it is how I found the sort-order behaviour, and because "my probe was malformed" and "the code under test is broken" look identical from the exit code alone.

artemtrofymenko (commented):
CodeRabbit's finding on the fixture loop is real, and it is reachable — measured at `08de25b` rather than argued.

Dropping **one clause** from the live rule at `:144`, leaving the fixture copy at `:399` untouched:

```js
- literals.some((v) => v === name || v.startsWith(`${name} `) || v.endsWith(`/${name}`));
+ literals.some((v) => v === name || v.startsWith(`${name} `));
```

```
✅ ROOT is the repo root …
✅ 8 root-level *-tests.mjs found to check
✅ 20 workflow file(s) read as the second run surface (30 run: scripts)
✅ every root-level *-tests.mjs is reachable … (8 checked)
✅ both match rules hold against 24 fixtures
📊 5 passed, 0 failed
```

`HARNESS_CASES` carries the case that exists to cover precisely that clause —

```js
["run(NODE, ['./x-tests.mjs']);", true, 'a path-qualified invocation'],
```

— and it passes while the rule it is meant to pin no longer recognises the form. Nothing else in the suite notices, because no real registration is path-qualified today. So the fixture block reports on a rule that stopped being the rule.

## The distinction worth writing down, because this repo has a precedent pulling the other way

`tests/providers/_http.test.mjs:41-44` deliberately does the opposite, and says so:

> Hardcoded here rather than imported — the whole point of pinning it is to catch a typo/drift in the production constant, not compare it to itself.

Both are right, and they are not about the same thing:

- **A pinned VALUE is hardcoded.** `'unexpected redirect'` is undici's wording; the test exists to redden when a runtime bump changes it. Importing the constant would compare it to itself and pin nothing.
- **A RULE is called, never restated.** A reimplemented predicate agrees with itself by construction and says nothing about the code that decides. That is what `:399` currently does.

The fixture block here is doing both jobs at once, which is why it reads as fine. The seven source snippets and their expected verdicts are pinned *values* and should stay exactly as they are — they are the specification. The predicate applied to them is a *rule* and should be the live one. CodeRabbit's `harnessMatches(literals, name)` split gets that line in the right place; after it, the mutation above reddens `both match rules hold against 24 fixtures`, which is what that assertion's name has been claiming all along.

Worth doing the same pass on the workflow half while you are there: `WORKFLOW_CASES` is the larger fixture set and the rule it exercises is the one that took five review rounds, so a silent divergence there costs more.

## Unrelated, and only because I nearly published it as a finding

My first attempt at this mutation went through `perl` and the `${name}` interpolation was eaten, so what actually ran was `v.startsWith(' ')` — a far bigger change than I intended, still 5 passed / 0 failed. Same green, different question answered. I redid it as the single-clause edit above before believing it. Flagging because "the rule was gutted and stayed green" and "one clause was removed and it stayed green" are very different claims, and only the second one is mine.

Scott-Emberson (commented):
Re-verified at the current head after the substantive rewrite landed (the "adopt the hardened reachability rules" commit; the other commits in the compare are main merges that do not touch this test's logic). The owned test `tests/root-tests-registration.test.mjs` is sound, and I mutation-tested it empirically at head:

- an unregistered `orphan-tests.mjs` reddens and names the file;
- registering it as a string literal in test-all.mjs greens it (so the guard is not tautological, registration flips it);
- a real `run: node www-tests.mjs` workflow step greens; an `echo node eee-tests.mjs` (data, not a command) reddens, so command-position discrimination holds;
- a trailing-comment mention (`// retired: 'zzz-tests.mjs'`) stays red, so the #3765 comment hole is closed;
- a heredoc body carrying the node line reddens, so the first-command-line masking holds end to end;
- and the reported hole reproduces as fixed: commenting out the three `run: node upgrade-tests.mjs` steps reddens naming `upgrade-tests.mjs`, so the old by-name exemption is gone and a disabled CI step is now caught.

The tokenizer skips `//`, `/* */`, and regex literals, and the workflow surface parses `jobs.*.steps[].run` requiring `node <file>` as the first command line. No tautology, no exemption list. `ARCHITECTURE.md` prose now matches the guard.

Owned test verified sound. This is GATE-tier (test harness), but the branch is behind main, so I am holding formal approval until it is rebased and green, then I will re-confirm at the final head.

Scott-Emberson (commented):
Re-verified at the current head. The owned test `tests/root-tests-registration.test.mjs` is sound, and the delta since my last look strengthened it: the "one definition of the harness match rule, used by both callers" commit deduplicated `harnessMatches` so the fixture loop now calls the live rule instead of restating it (the commit confirmed that the `endsWith('/name')`-drop mutation now fails the fixture block, where before it left all 24 green over a dead rule). The workflow reachability tokenizer (first-command-only `invokesNode` on `firstCommandLine`) was not touched.

The guard still bites: an orphan root `*-tests.mjs` that is neither a string literal in test-all.mjs nor a first-command `node` call in a workflow lands in `unreachable` and fails naming it, with anti-no-op guards (missing test-all.mjs, empty suite list) preventing a silent green, and echo/comment/heredoc mentions correctly not counting as registration.

One cross-PR coordination note, not a fault in this PR: #3765 (move five root suites into tests/) also adds this same file from scratch, so the two will conflict textually on it, and #3765 reduces the root suite count from eight to three, which changes the "eight/seven" premise this PR is written around. From the commit history these read as a stacked pair rather than rival duplicates (the plan is this guard lands first and #3765 restacks on top), so the resolution is a merge ordering (land #3764, then #3765 rebases and updates the counts) rather than picking one. Flagging so whoever merges sequences them.

Owned test verified sound. This is GATE-tier, but the branch is behind main, so I am holding formal approval until it is rebased and green, then I will re-confirm at the final head.

artemtrofymenko (commented):
Re-ran my own mutation against `a64b26d` on Windows 11, rather than take the fix on the commit message. It closes.

Same single-clause edit, now applied to the one shared `harnessMatches` at `:362`:

```js
- return literals.some((v) => v === name || v.startsWith(`${name} `) || v.endsWith(`/${name}`));
+ return literals.some((v) => v === name || v.startsWith(`${name} `));
```

```
❌ 1 match-rule fixture(s) failed:
    harness rule: a path-qualified invocation → false, want true
📊 4 passed, 1 failed
```

Before the dedup the identical mutation gave `5 passed, 0 failed` with `both match rules hold against 24 fixtures` still green. The fixture that exists to cover that clause now actually covers it, and it names itself in the failure — which is the property that makes the block worth having rather than just present.

Worth saying which half of the change did the work, because it is the transferable part: the fixture *inputs* stayed hardcoded (`["run(NODE, ['./x-tests.mjs']);", true, …]` and its six siblings), and only the *predicate* moved to the live definition. Those are two different jobs that look alike in a fixture table — a pinned value belongs to something outside your control and must be restated, while a rule must be called, because a reimplementation agrees with itself by construction. Keeping the values literal and the rule shared is exactly the line, and it is why this fix strengthens the block rather than just deduplicating it.

Nothing further from me on this one. Not re-testing the workflow surface: `invokesNode`/`firstCommandLine` are untouched in this delta, and I verified them at `08de25b` — four hostile inputs red, two real registrations green, on both surfaces.

Issue #3704, opened 2026-09-02
tracker-parse: localized tracker headers shipped by modes/{de,pl,pt,da,id} resolve to null and fall back to the legacy fixed layout


vliggio:
Split out of #3517 (finding 3 of the column-order grep), where the maintainer asked for these to be filed separately.

Five localized evaluation modes tell users to write a tracker header in their own language. `tracker-aliases.json` is English plus `empresa`/`puesto`, and `headerSchemaMap` requires all of num/company/role/score/status to resolve — so those headers return `null` and **every** reader falls back to `LEGACY_COLMAP`. On a plain nine-column table the fallback lines up by luck. Insert a column and it is #2274 verbatim, still live.

## Where the headers are shipped

- `modes/de/angebot.md:164` — `| # | Datum | Firma | Rolle | Score | Status | PDF | Report |`
- `modes/pl/oferta.md:165`
- `modes/pt/oferta.md:165`
- `modes/da/oferta.md:165`
- `modes/id/lowongan.md:165`

## Reproduction

```
DE detect: null
```

Add the Location column from #946's own use case (`| # | Datum | Firma | Ort | Rolle | Score | Status | PDF | Report |`) and the fallback shifts every field one left:

```
DE row: {"date":"2026-01-01","company":"Acme","role":"Berlin","score":"Engineer","status":"4.2/5","pdf":"Applied","report":"❌","notes":"[1](r.md)"}
EN row: {"date":"2026-01-01","company":"Acme","location":"Berlin","role":"Engineer","score":"4.2/5","status":"Applied","pdf":"❌"}
```

Role reads the city, Score reads the role title, Status reads the score.

## Why this is the weak link, not a corner case

For these users the default column order is not merely assumed — it is the only thing holding the parse together, and the repo's own modes are what put them there. The failure is silent by construction: `detectColumns` returning `null` is the documented signal to use the legacy map, so nothing warns.

It also sits on the critical path for anything that adds a second name-resolution surface. #3517 gave tracker-addition files a header row resolved through this same alias table; the table's English-only coverage is now load-bearing in two places instead of one.

## Options

1. Extend `tracker-aliases.json` with the labels this repo's own modes ship (de/pl/pt/da/id, and any other market set that prints a header). Smallest fix, and it is the table's stated purpose — `tracker-parse.mjs:106-112` already cites a fully localized header as the case the full-schema check exists to serve.
[...]

Issue #3517, opened 2026-08-30
tracker: the TSV ingest format transposes score/status relative to applications.md, and a content-sniffing guard exists to survive it


vliggio:
The TSV ingest format and `data/applications.md` disagree on the order of two adjacent columns, and the disagreement is load-bearing enough that a content-sniffing guard exists to survive it.

`AGENTS.md:454`:

> **Column order (IMPORTANT -- status BEFORE score):** 1 `num` · 2 `date` · 3 `company` · 4 `role` · 5 `status` · 6 `score` · 7 `pdf` · 8 `report` · 9 `notes`

`AGENTS.md:456`:

> **Note:** In applications.md, score comes BEFORE status; `merge-tracker.mjs` handles the swap automatically.

So the same two fields are adjacent and transposed between the format written by evaluators and the file it merges into.

## Why this is worth fixing rather than documenting

The tracker side is robust: `tracker-parse.mjs` resolves columns by header name through `HEADER_ALIASES`, `detectColumns()` reassigns `COLMAP` to the detected layout, and writes follow the real header width. A user can reorder their tracker's columns freely and every reader and writer follows. I verified this end to end — reordered header, ran `stats.mjs`, `verify-pipeline.mjs`, and a `merge-tracker.mjs` insert, all correct.

The TSV side has none of that. It is fixed positional, with no header row and no names. So the two representations can only be reconciled by a positional swap that has to be remembered at exactly one point in the pipeline.

And it is genuinely hazardous, which the codebase already acknowledges: `tracker-parse.mjs:61` carries `looksLikeScoreCell()`, which identifies the score column by *content pattern* rather than position:

```js
export const SCORE_CELL_RE = /^\d+(?:\.\d+)?\/5$/;
export function looksLikeScoreCell(v) {
  const t = String(v ?? '').replace(/\*\*/g, '').trim();
  return SCORE_CELL_RE.test(t) || t === 'N/A' || t === 'DUP' || t === '—' || t === '-';
}
```

That guard was added under #1427 specifically to stop a silent score/status swap, and `AGENTS.md` documents the consequence for authors: a backfilled row must use one of the recognized sentinels (`N/A`, `—`, `-`), never blank and never another placeholder, or the row becomes ambiguous and is skipped with a warning (#1799).

[...]

PR #3767, opened 2026-09-03
fix(tracker): map "fecha" to the date column (#3705)

### LINKED ISSUE #3705: tracker-aliases.json: "fecha" is missing, so the Spanish header resolves with no date column instead of failing
Split out of #3517 (finding 4 of the column-order grep), where the maintainer asked for these to be filed separately.

`tracker-parse.mjs:106-112` cites `| # | Fecha | Empresa | Puesto | … |` as the exact case the full-schema header check exists to serve. `fecha` is not in `tracker-aliases.json`.

`date` is not in `REQUIRED_HEADER_FIELDS`, so the header still resolves — it just resolves without a `date` column, and every row comes back with an empty date:

```
ES detect: { num: 1, company: 3, role: 4, score: 5, status: 6, pdf: 7, report: 8, notes: 9 }   # no date
ES row:    {"num":1,"date":"","company":"Acme","role":"Engineer","score":"4.2/5","status":"Applied", …}
```

Note this is not a fallback to the legacy layout, which would at least read position 2 and get the right value by accident. The date is resolved as absent, and consumers that key off it (follow-up cadence, funnel velocity, anything sorting or ageing rows) see a row with no date rather than a parse failure.

The alias table already carries `empresa` and `puesto`, so the Spanish header is half-supported: two of its three localized labels resolve and the third silently does not.

## Fix

Add `fecha` to `tracker-aliases.json`. One line, in the file that is explicitly the single place aliases belong (`tracker-parse.mjs` says to add them in the JSON, not in code), and the web reader picks it up from the same file.

Worth checking in the same pass whether other localized labels this repo ships are missing — that overlaps with the localized-header issue, which is about headers that fail to resolve *at all* rather than one that resolves incompletely.

## Environment

- Node.js on macOS

AdilAzhariOmsan:
Fixes #3705.

## Problem

`tracker-parse.mjs` resolves tracker table columns by header name through `tracker-aliases.json`. The Spanish label **`Fecha`** is missing from that table, even though `empresa` → company and `puesto` → role are present — so the Spanish header is only half-supported.

Because `date` is not in `REQUIRED_HEADER_FIELDS`, a localized header like `| # | Fecha | Empresa | Puesto | … |` still *resolves* — it just resolves **with no date column**. Every row then comes back with an empty date, and consumers keyed off it (follow-up cadence, funnel velocity, anything sorting or ageing rows) see a dateless row rather than a parse failure. This is not a fallback to the legacy fixed layout, which would at least read position 2 by accident.

## Fix

- Add `"fecha": "date"` to `tracker-aliases.json` — one line, in the single file aliases are meant to live in (`tracker-parse.mjs` says to add them there, not in code). The web reader loads the same file, so it picks this up too (`tracker-columns-tests.mjs` Test 8 already guards that the two tables stay byte-identical). `fecha` is already an established localized date alias elsewhere in the codebase (`web/src/lib/format.ts`).
- Strengthen the existing "fully localized header maps through the alias table" test (#2274). It checked `company` / `role` / `location` / `score` positions but **not** `date` — which is exactly why this gap survived. It now asserts the `Fecha` column resolves.

Scope kept to `fecha` per the maintainer's split on #3517. The separate matter of `modes/es/oferta.md` shipping `Rol` / `Estado` (labels that fail to resolve *at all*) belongs to #3704.

## Testing

- `node tracker-columns-tests.mjs` — 46/46 pass
- `node test-all.mjs` — the targeted assertion `a fully localized header maps through the alias table (#2274, #3705)` passes; full suite 8090 pass (one unrelated `update-system.mjs check` network-timeout flake under full-suite load, passes in isolation)
- Confirmed the repro: without the alias, `detectColumns` returns a map with no `date` key; with it, `date` resolves to column 2

### DIFF
--- a/tracker-aliases.json
+++ b/tracker-aliases.json
@@ -2,6 +2,7 @@
   "#": "num",
   "num": "num",
   "date": "date",
+  "fecha": "date",
   "company": "company",
   "empresa": "company",
   "via": "via",
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -12274,11 +12274,15 @@ try {
     // to LEGACY_COLMAP (#2274). On a plain 9-column table the fallback happens
     // to line up and hides the bug; with a Location column inserted, the Score
     // cell is read from Location instead — an ES tracker scored "Remote".
+    // `date` is not in REQUIRED_HEADER_FIELDS, so a missing `fecha` alias does
+    // not fail the header — it resolves with no date column and every row comes
+    // back with an empty date (#3705). Assert the Fecha column is actually
+    // resolved, not silently absent.
     const trackerParse = await import(pathToFileURL(join(ROOT, 'tracker-parse.mjs')).href);
     const esHeader = '| # | Fecha | Empresa | Puesto | Location | Score | Status | PDF | Report | Notes |';
     const esMap = trackerParse.detectColumns([esHeader]);
-    if (esMap && esMap.score === 6 && esMap.company === 3 && esMap.role === 4 && esMap.location === 5) {
-      pass('a fully localized header maps through the alias table (#2274)');
+    if (esMap && esMap.date === 2 && esMap.score === 6 && esMap.company === 3 && esMap.role === 4 && esMap.location === 5) {
+      pass('a fully localized header maps through the alias table (#2274, #3705)');
     } else {
       fail(`localized header did not map: ${JSON.stringify(esMap)}`);
     }


tracker-aliases.json
```diff
@@ -2,6 +2,7 @@
   "#": "num",
   "num": "num",
   "date": "date",
+  "fecha": "date",
```
nikitacometa, line 5: Correct and in the right file. The same one-line gap remains for `datum`, `data`, `dato`, `tanggal`, all shipped in `modes/*/`. A follow-up issue rather than scope creep here.

nikitacometa (approved):
Verified at e206ce7e. With the alias reverted, `detectColumns` on the #3705 header returns no `date` key and `parseTrackerRow` yields `"date": ""`; with it, `date=2` and the row carries the value. `tracker-columns-tests.mjs` passes 46/46 either way, so the guard is the assertion at test-all.mjs:12284, which does go red without the alias; worth citing that one in the description instead of the columns-test count. `looksLikeScoreCell` and the score/status swap are unchanged on both heads.

Two things this does not cover, both fine as follow-ups: modes/es/oferta.md:165 ships `Rol` and `Estado`, so that header still fails `isHeaderRow` and falls back to LEGACY_COLMAP before and after (#3704 owns it); and `Datum` (de), `Data` (pl, pt, it), `Dato` (da), `Tanggal` (id) lose the date the same silent way. #3679's FIELD_KEYS already lists `data` and `dato`.

PR #3771, opened 2026-09-03
docs(types): document the optional Job.salary field

ilyakanevskiy:
`providers/_types.js`'s `Job` typedef gains an optional `[salary]` entry — `{min?: number, max?: number, currency?: string}`.

`job.salary` is already part of the contract in practice: `ashby`, `wttj`, `manfred`, `remotli`, `getro`, `agentic-jobs` and `builtin` attach it, and `scan.mjs` both filters on it (`buildSalaryFilter` / `salary_filter`) and renders it into `data/pipeline.md`'s compensation column (`formatCompensation`). The typedef just never listed it, so `// @ts-check` provider files got no hint for the field and a new provider handling comp data had no sanctioned shape to mirror — `ADDING_A_PROVIDER.md` §1 points authors at `_types.js` as the type catalog.

All three sub-fields are optional. Most providers (`ashby`, `getro`) normalize both bounds, filling a one-sided range from the other; `agentic-jobs` deliberately omits the absent bound rather than coerce it (`normalizeAgenticSalary` returns `{min}` or `{max}` alone), and `scan.mjs`'s `salary_filter` reads `salary.min ?? salary.max`, so a one-sided object is valid input. Declaring both bounds required would make that a `// @ts-check` error via `Provider.fetch`'s `Promise<Job[]>` return type.

`currency` is the source's currency string, upper-cased by most providers (`ashby`, `agentic-jobs`) or passed through with only a `.trim()` (`getro`); `scan.mjs` compares it case-insensitively and never validates it as an ISO code. Absent or `''` when the source gives none.

JSDoc only. `_types.js` is documentation-only (per its own header) — nothing executes it, no runtime or behaviour change.

Tested: `node test-all.mjs` green (8092 passed, 0 failed); `npm run lint` clean.

## Summary

Documents the optional `Job.salary` field in `providers/_types.js`:13.

The field accepts optional `min` and `max` numeric bounds and an optional `currency` string. Missing bounds use the other bound. Currency is compared case-insensitively and is not validated.

This change helps providers and consumers use documented salary filtering and compensation rendering. It changes JSDoc only. It does not change runtime behavior.

## User impact

Users can now rely on the documented salary shape and currency behavior. No existing behavior breaks.

## Files changed

`providers/_types.js`: updated the `Job` typedef.

## Validation

`node test-all.mjs` and `npm run lint` pass.

### DIFF
--- a/providers/_types.js
+++ b/providers/_types.js
@@ -35,19 +35,25 @@
  *                               usable date. scan.mjs ignores it; consumers
  *                               like scan-ats-full.mjs use it for recency
  *                               filtering.
- * @property {{min: number, max: number, currency?: string}} [salary]
+ * @property {{min?: number, max?: number, currency?: string}} [salary]
  *                               Annualized compensation, attached ONLY when the
  *                               source exposes real figures — never inferred
  *                               (`ashby.mjs` is the reference shape; most
- *                               providers omit it). `min` and `max` are both
- *                               present — a one-sided range fills the missing
- *                               bound from the other.
- *                               `currency` is an ISO code when the source gives
- *                               one, and `''` or absent otherwise. Consumed by
- *                               scan.mjs's salary_filter and rendered into
- *                               pipeline.md's compensation column via
- *                               formatCompensation(); an empty/absent value
- *                               always passes the filter.
+ *                               providers omit it). At least one of `min` / `max`
+ *                               is present; most providers normalize both bounds
+ *                               (filling a one-sided range from the other), but
+ *                               some — e.g. `agentic-jobs.mjs` — leave the absent
+ *                               bound off rather than coerce it. scan.mjs's
+ *                               salary_filter reads `min ?? max` and tolerates
+ *                               either bound alone.
+ *                               `currency` is the source's currency string,
+ *                               upper-cased by most providers (`''` or absent
+ *                               when the source gives none); it is not validated
+ *                               as an ISO code, and the filter compares it
+ *                               case-insensitively. Consumed by scan.mjs's
+ *                               salary_filter and rendered into pipeline.md's
+ *                               compensation column via formatCompensation(); an
+ *                               empty/absent value always passes the filter.
  * @property {number} [trustScore] 0-100 trust score from _trust-validator.mjs.
  * @property {string[]} [trustFlags] Flags raised by trust validation (e.g.
  *                                   'invalid_url', 'suspicious_domain').
--- a/providers/_types.js
+++ b/providers/_types.js
@@ -35,6 +35,19 @@
  *                               usable date. scan.mjs ignores it; consumers
  *                               like scan-ats-full.mjs use it for recency
  *                               filtering.
+ * @property {{min: number, max: number, currency?: string}} [salary]
+ *                               Annualized compensation, attached ONLY when the
+ *                               source exposes real figures — never inferred
+ *                               (`ashby.mjs` is the reference shape; most
+ *                               providers omit it). `min` and `max` are both
+ *                               present — a one-sided range fills the missing
+ *                               bound from the other.
+ *                               `currency` is an ISO code when the source gives
+ *                               one, and `''` or absent otherwise. Consumed by
+ *                               scan.mjs's salary_filter and rendered into
+ *                               pipeline.md's compensation column via
+ *                               formatCompensation(); an empty/absent value
+ *                               always passes the filter.
  * @property {number} [trustScore] 0-100 trust score from _trust-validator.mjs.
  * @property {string[]} [trustFlags] Flags raised by trust validation (e.g.
  *                                   'invalid_url', 'suspicious_domain').


Scott-Emberson (commented):
Reviewed at head 30be716e38. This is a docs-only change in our providers/ area: it documents the optional Job.salary field in providers/_types.js (a JSDoc typedef comment), with no runtime code or test in the diff, so there is nothing to verify on the test axis. It matches how the salary field is actually used by the providers, so the type doc is accurate. CI is green and the branch is up to date (blocked only on review). From our side there is nothing blocking, though as a pure doc it is really the maintainer's to land.

providers/_types.js
```diff
@@ -35,6 +35,19 @@
  *                               usable date. scan.mjs ignores it; consumers
  *                               like scan-ats-full.mjs use it for recency
  *                               filtering.
+ * @property {{min: number, max: number, currency?: string}} [salary]
```
nikitacometa: `min` and `max` should be optional. agentic-jobs.mjs:144 assigns a `{min?, max?, currency?}` object here; under `// @ts-check`, which that file enables, this typedef makes the assignment a type error.

providers/_types.js
```diff
@@ -35,6 +35,19 @@
...
+ *                               Annualized compensation, attached ONLY when the
+ *                               source exposes real figures — never inferred
+ *                               (`ashby.mjs` is the reference shape; most
+ *                               providers omit it). `min` and `max` are both
+ *                               present — a one-sided range fills the missing
```
nikitacometa: "`min` and `max` are both present" is false for agentic-jobs, which omits the missing bound by design (agentic-jobs.mjs:97-99). Probed: `normalizeAgenticSalary({salaryMin: 120000})` gives `{"min":120000}`.

providers/_types.js
```diff
@@ -35,6 +35,19 @@
...
+ *                               bound from the other.
+ *                               `currency` is an ISO code when the source gives
+ *                               one, and `''` or absent otherwise. Consumed by
```
nikitacometa: "ISO code" overstates: getro.mjs:179 and ashby only `.trim()` the source string. `buildSalaryFilter` compares case-insensitively, so "the source's code, compared case-insensitively" matches the code.

nikitacometa (changes requested):
providers/_types.js:38 declares `min` and `max` required, and 43-45 says both are always present. providers/agentic-jobs.mjs, listed in this PR's own description as an emitter, contradicts that. At this head `normalizeAgenticSalary({salaryMin: 120000})` returns `{"min":120000}` and `normalizeAgenticSalary({salaryMax: 90000, salaryCurrency: 'usd'})` returns `{"max":90000,"currency":"USD"}`; its local typedef (agentic-jobs.mjs:107, 129) is `{min?, max?, currency?}` and 97-99 says an absent bound is omitted, never coerced. That file is `// @ts-check` and imports this typedef, so the assignment at 144 becomes a type error.

scan.mjs:874-875 (`salary.min ?? salary.max`) already tolerates one-sided input, so the runtime contract is wider than the doc, and since `_types.js` is the catalog provider authors mirror, the narrower doc pushes a new provider to invent a bound. `{{min?: number, max?: number, currency?: string}}` plus a line that at least one bound is present and most providers normalize both matches the code. "ISO code" also overstates: getro.mjs:179 and ashby only `.trim()` the source string; the filter compares case-insensitively, so say that.

providers/_types.js
ilyakanevskiy: Done in 0dda2d7 — `salary` is now `{min?: number, max?: number, currency?: string}`. Small correction on the mechanism: agentic-jobs.mjs imports only `Provider`, so line 144 assigns against that file's own permissive local typedef and type-checks there. The `// @ts-check` error surfaces one level up, through `Provider.fetch`'s `Promise<Job[]>` return contract. The fix is the same either way.

providers/_types.js
ilyakanevskiy: Fixed in 0dda2d7. The prose now says at least one of `min` / `max` is present, most providers normalize both, and `agentic-jobs` omits the absent bound by design — `normalizeAgenticSalary` returns `{min}` or `{max}` alone.

providers/_types.js
ilyakanevskiy: Fixed in 0dda2d7 — now described as the source's currency string, upper-cased by most providers (`.trim()`-only in `getro`), compared case-insensitively by `scan.mjs` and never validated as an ISO code.

Scott-Emberson (commented):
Verified the change to `providers/_types.js`. It is a documentation-only JSDoc addition: it documents the optional `Job.salary` field (`{min?, max?, currency?}`), attached only when the source exposes real figures and never inferred, with `ashby.mjs` as the reference shape. The doc accurately describes the field's real contract, that at least one of `min`/`max` is present, that some providers leave a one-sided bound off rather than coerce it, and that `scan.mjs`'s salary_filter reads `min ?? max` and compares currency case-insensitively with an empty/absent value passing. This matches the `Job.salary` typedef and consumer behavior already in the tree.

No executable change and no behavioral surface, so there is no owned test to exercise; the typedef is accurate as written. This is a GATE-tier file (`providers/`), verified sound as documentation. Holding formal approval until Scott gives the go-ahead, and surfacing it as ready-to-approve.

PR #3774, opened 2026-09-03
fix(update): run verify-pipeline.mjs alongside doctor.mjs after apply

beerock:
## What

`modes/update.md` Step 6 currently validates a completed update with `node doctor.mjs` only. This adds `node verify-pipeline.mjs` alongside it.

## Why

`doctor.mjs` checks the **setup layer**: do `cv.md`, `config/profile.yml`, `modes/_profile.md`, `portals.yml` exist and carry real (non-template) content. It does not check whether existing **data** still matches the shape the just-updated code expects.

Reproduced going from v1.26.0 to v1.31.0: `#2971` added a column-order check for `data/follow-ups.md`. A pre-existing `follow-ups.md` written in the prior column order made `parseFollowups()` skip every row silently — `stats.mjs` and `followup-cadence.mjs` both reported **zero follow-ups** instead of erroring, on an install that actually had several overdue. `doctor.mjs` reported `onboardingNeeded: false` throughout; `verify-pipeline.mjs` was the only check in the repo that caught the break (it already validates `data/follow-ups.md`'s schema, among other tracker/report checks — this PR doesn't add a new check, it just wires the existing one into the update flow).

The general failure shape isn't specific to this one release: any future change to a data file's expected shape can pass `doctor.mjs` cleanly while every downstream reader of that file degrades silently. `verify-pipeline.mjs` is the tool built to catch exactly that class of break, so it belongs in the same place `doctor.mjs` already is.

## Change

Minimal, three lines in one file:
- Step 6 runs `verify-pipeline.mjs` right after `doctor.mjs`, with the same non-zero-exit → stop-and-offer-rollback handling already used for `doctor.mjs`.
- The "do not run X on a partially-applied update" note in Step 5 now also names `verify-pipeline`.
- The final success message mentions `verify-pipeline.mjs` alongside `doctor.mjs`.

No issue was filed first — this is a small, self-contained doc fix discovered and reproduced during normal use, with the repro steps above standing in for one. Happy to file an issue too if that's preferred process for this repo.

https://claude.ai/code/session_015d3wzPAaxLVcFnuiiLbBqJ

`modes/update.md:Step 4–6` now runs `verify-pipeline.mjs` after `doctor.mjs` for completed updates.

From the user’s point of view:

: Updates validate existing data schemas and setup files.
: `doctor.mjs` failures still stop the update and use rollback handling.
: `verify-pipeline.mjs` reports data issues without stopping the update or triggering rollback.
: Warning and success messages mention `verify-pipeline.mjs`.

System files touched: `modes/update.md` only.

### DIFF
--- a/modes/update.md
+++ b/modes/update.md
@@ -78,9 +78,9 @@ If yes:
 5. Now check the exit code captured in step 3:
    - If non-zero, treat apply as failed. Show the captured output and offer:
      > "⚠️ Update apply failed. Want me to show the full error, or try `/career-ops update rollback`?"
-   - Stop the flow here if apply failed — do not run doctor or reconciliation on a partially-applied update.
-6. Run `node doctor.mjs` to validate the installation
-   - If the command exits with a non-zero code, treat validation as failed. Show the captured output and offer:
+   - Stop the flow here if apply failed — do not run doctor, verify-pipeline, or reconciliation on a partially-applied update.
+6. Run `node doctor.mjs` to validate the installation, then `node verify-pipeline.mjs` to validate the data layer. They check different things: doctor confirms the setup files exist and are personalized, not that existing data still matches what the updated code expects — a release that changes a data file's expected shape can pass doctor cleanly while breaking every downstream reader of that file silently (e.g. a tracker/log schema change that makes existing rows fail to parse, so stats and cadence tools report zero instead of erroring).
+   - If either command exits with a non-zero code, treat validation as failed. Show the captured output and offer:
      > "⚠️ Validation failed after update. Want me to show the full error, or roll back with `/career-ops update rollback`?"
    - Stop the flow here if validation failed — do not run reconciliation or show the success message.
 7. If Step 3 flagged archetype/scoring changes, reconcile `modes/_profile.md` against the new `modes/_shared.md`:
@@ -97,7 +97,7 @@ If yes:
      - For removals:
        > "Your _profile.md references archetype '{old_name}' which was removed in the new _shared.md. Want me to delete the reference or replace it with another archetype?"
 8. Show final status:
-   > "✅ Updated to v{version}. Run `node doctor.mjs` anytime to verify setup."
+   > "✅ Updated to v{version}. Run `node doctor.mjs` anytime to verify setup, or `node verify-pipeline.mjs` to check your data."
 
    If the updater's output ended with its note about the CareerOps Manifesto, relay it once (do not drop it when summarizing):
    > "One more thing: this project ships with the CareerOps Manifesto — a new way of job searching is taking shape, and you are already practicing it. Run `npm run manifesto` to read it and sign it if you want to help. No action needed."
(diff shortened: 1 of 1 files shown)


modes/update.md
```diff
@@ -78,9 +78,9 @@ If yes:
 5. Now check the exit code captured in step 3:
    - If non-zero, treat apply as failed. Show the captured output and offer:
      > "⚠️ Update apply failed. Want me to show the full error, or try `/career-ops update rollback`?"
-   - Stop the flow here if apply failed — do not run doctor or reconciliation on a partially-applied update.
-6. Run `node doctor.mjs` to validate the installation
-   - If the command exits with a non-zero code, treat validation as failed. Show the captured output and offer:
+   - Stop the flow here if apply failed — do not run doctor, verify-pipeline, or reconciliation on a partially-applied update.
+6. Run `node doctor.mjs` to validate the installation, then `node verify-pipeline.mjs` to validate the data layer. They check different things: doctor confirms the setup files exist and are personalized, not that existing data still matches what the updated code expects — a release that changes a data file's expected shape can pass doctor cleanly while breaking every downstream reader of that file silently (e.g. a tracker/log schema change that makes existing rows fail to parse, so stats and cadence tools report zero instead of erroring).
```
nikitacometa: `verify-pipeline` exits 1 on pre-existing tracker state (132 non-canonical status, 181 missing report, 431 duplicate row number). Probed: a tracker with one such row gives EXIT=1. Gating here reports the user's own data as an update failure; capture it before the apply and gate on the transition. Also a 90-word sentence in prompt text the agent re-reads each update; split instruction from rationale.

modes/update.md
```diff
@@ -78,9 +78,9 @@ If yes:
...
+   - If either command exits with a non-zero code, treat validation as failed. Show the captured output and offer:
```
nikitacometa: The rollback offer is wrong for the common case here: rolling back the release will not fix a non-canonical status or a missing report file. A separate message when verify-pipeline is the failing command.

modes/update.md
```diff
@@ -78,9 +78,9 @@ If yes:
...
      > "⚠️ Validation failed after update. Want me to show the full error, or roll back with `/career-ops update rollback`?"
    - Stop the flow here if validation failed — do not run reconciliation or show the success message.
```
nikitacometa: This stop clause now skips Step 7 reconciliation for any user whose tracker has a pre-existing error.

nikitacometa (changes requested):
Direction is right, and the two failure modes I expected are absent: verify-pipeline.mjs:78-81 exits 0 on a first-run repo with no tracker, and 521 exits non-zero only on errors, not warnings. `upgrade-tests.mjs --pr-gate` is green at c3589559. `update-system.mjs apply` never invokes either script itself, so modes/update.md is the right place.

One problem. `verify-pipeline` exits 1 on conditions that predate the release: a non-canonical status (132), a report file the user moved (181), a bad score cell (191), a duplicate row number (431). Probed with a tracker holding one such row (`| 1 | … | Waiting | ✅ | [1](reports/001-acme-2026-08-01.md) |`): `EXIT=1`, 2 errors. Step 6 then reports "Validation failed after update" and offers a rollback that fixes neither, and its stop clause skips Step 7 archetype reconciliation for every user in that state, on the releases where reconciliation matters most. Before this PR that data never touched the update flow.

Suggest running `verify-pipeline` before the apply too and gating only on a clean-to-dirty transition, which also matches your own repro (clean at v1.26.0, broken at v1.31.0). The cheaper alternative is to keep the run informational: show the output, do not gate Steps 7-8 on it, and drop the rollback offer for this check.

PR #3781, opened 2026-09-03
refactor(update-system): extract the checkout skip-check into pathFullyPreserved()

### LINKED ISSUE #3779: update-system.mjs apply fails with exclude-pathspec cancel-out on a preserved root file (AGENTS.md), leaving a partial staged commit
> ⚠️ No personal data included — repro is generic filenames and command output only.

**Code of Conduct:** ✅ agreed
**Existing issues:** ✅ searched (`pathspec`, `update-system`, `apply --confirm`, `commit failed`, `did not match any files`, `exclude pathspec`) — nothing open covers this.

## What happened?

Running `node update-system.mjs apply --confirm` (v1.30.0 → v1.32.0) correctly detected several locally-customized system files and said it would preserve them, then failed:

```
fatal: pathspec ':(exclude)AGENTS.md' did not match any files
...
Update commit failed (files may be staged but not committed).
    Error: Command failed: git --literal-pathspecs add -f -- <hundreds of files>
```

This left the working tree with dozens of system files staged but uncommitted. `node update-system.mjs rollback` mostly recovered (restored 322 paths from the pre-update backup branch) but also logged 6 `pathspec '...' did not match any file(s) known to git` errors for new upstream files that weren't in the backup branch — it did still remove them per its own summary line ("removed 6 path(s) added after the backup"). After rollback, ~48 untracked new system-only files (docs/tests/provider stubs, no user data) plus a `.gitignore` diff were still left over and needed manual cleanup to fully return to the pre-update state.

No user-layer data (`cv.md`, `config/profile.yml`, `data/`, `reports/`, `modes/_profile.md`, `modes/_custom.md`) was touched at any point — confirmed via `git status` throughout.

This looks like a distinct edge case in the exclude-cancel-out skip-check at `update-system.mjs:1811-1838` (the code that builds `git checkout FETCH_HEAD -- <path> :(exclude)<preserved-file>...` per path, with a check meant to skip a path entirely when every upstream file under it is preserved — since `<path> :(exclude)<path>` cancels out to nothing). That skip-check didn't catch the `AGENTS.md` case here.

Worth noting: this file has already been patched once for a similar-looking pathspec fragility class — `update-system.mjs:1841-1847` pipes stderr specifically to avoid a confusing `pathspec ... did not match any file(s)` message from an unrelated, expected skip case (fixed per #1998, pre-1.11.0 client upgrades silently skipping new upstream paths). That fix doesn't cover this failure mode, since it's a different code path (exclude-cancel-out on a *preserved* file during a normal same-generation upgrade, not a stale/absent-upstream path on an old client).

## Steps to reproduce

1. Have local edits to a handful of tracked system files that also changed upstream between versions. In this case: `AGENTS.md`, 4 files under `fonts/`, `modes/pdf.md`, 3 `templates/cv-template.*.html` files, `test-all.mjs`, one file under `tests/`, and `verify-cv-facts.mjs`.
2. Run `node update-system.mjs check` (reports an update available).
3. Run `node update-system.mjs apply --confirm`.
4. Observe the `fatal: pathspec ':(exclude)AGENTS.md' did not match any files` error and the subsequent "Update commit failed" message.

## Expected behavior

The update should either apply cleanly while preserving the locally-customized files (per its own stated behavior), or fail atomically without leaving a half-staged working tree — and `rollback` should fully restore the pre-update state without needing manual cleanup of leftover untracked files.

## CLI tool

Claude Code

## OS and Node.js version

macOS 26.6.2 / Node v24.3.0

## Output of `npm run doctor`

```json
{"onboardingNeeded":false,"missing":[],"unpersonalized":[],"warnings":[],"autoCopied":[]}
```

(Repo was on commit `424952ba`, VERSION 1.30.0, at time of the failed update.)

euskode:
## Summary

Refs #3779 — **this does not fix it**, see below.

`apply()`'s checkout loop built `git checkout FETCH_HEAD -- <path> :(exclude)<path>` per SYSTEM_PATHS entry and skipped entries whose exclusions would cancel the pathspec out, via an inline `ls-tree` round-trip against upstream. For a single-file entry that exactly matches a preserved path, that lookup is redundant: the string match already proves the whole file is preserved.

This extracts the check into an exported `pathFullyPreserved()`, ctx-injectable the same way `locallyModifiedSystemFiles()` already is, so it is directly unit-testable instead of living inline inside the ROOT-bound `apply()`, and short-circuits before any git call when the match is a single file. The directory case, which genuinely needs the `ls-tree` to confirm every file under it is preserved, is unchanged.

## Why this is `Refs`, not `Fixes` (corrected after review)

I originally filed this as `Fixes #3779` on a wrong reading of that issue, and @nikitacometa was right to catch it. Re-traced:

#3779's failure is `fatal: pathspec ':(exclude)AGENTS.md' did not match any files` from `git --literal-pathspecs add -f -- <hundreds of files>`. That command is `addPaths()`, fed by `expandToShippedFiles()`, which passes `:(exclude)AGENTS.md` through unexpanded because it doesn't end in `/`. Under `--literal-pathspecs` git reads it as a literal filename that doesn't exist, and the whole staging step dies.

That is a different step from this checkout skip-check, which already returned `true` for `AGENTS.md` before this PR — so it never failed and was never the cause. #3779 is the same root cause as #3740, and **#3742 fixes the real site**.

## Known limits (both documented on the function)

1. **Single-file divergence.** For a preserved file *absent* from FETCH_HEAD the pre-extraction inline check fell through to the real checkout, which failed and put the path in apply()'s "Skipped N path(s) absent upstream" summary; this returns `true` and skips it silently. Unreachable today — preserved paths come from `locallyModifiedSystemFiles`, which only reports files that exist upstream (it gates each candidate on `cat-file -e <ref>:<file>`) — but it is a real divergence, not a behaviour-preserving one.

2. **The directory cancel-out is NOT closed.** A throwing `ls-tree` still returns `false` and falls through to `git checkout FETCH_HEAD -- modes/ :(exclude)modes/pdf.md`, which aborts the update when that directory is fully preserved. That is the pre-extraction behaviour carried over unchanged — this PR makes the check testable and drops a redundant lookup; it does not fix the directory case. Closing it needs the tri-state result @nikitacometa suggested, whose "unknown" makes the checkout's "did not match any file(s)" benign at the call site. Tracked separately rather than folded in here, because it means matching on git's stderr text — a fragility this file has deliberately avoided before (#1998 chose stderr piping over string matching).

## Scope changes after review

- The two `rollback()` edits (the `gitQuiet` swap and the `ls-tree` backup-existence check) are **removed from this PR**. They were unrelated to the title, on the recovery path, and untested; the `ls-tree` one also threw from inside a `catch`, which would end rollback mid-restore and surface the lookup error instead of the original. They move to #3782, which already rewrites that function.
- This PR is now purely the extraction: `update-system.mjs` + its test.

## Test plan

- [x] `node --test tests/updater-local-system-edits.test.mjs` — passes, including the 5 `pathFullyPreserved()` cases (single-file skip with zero git calls, unrelated path with zero git calls, fully-preserved directory retaining the ls-tree path, partially-preserved directory not skipped, unreadable ls-tree degrading to `false`)
- [x] `node updater-migration-tests.mjs` — 382 passed, 0 failed
- [x] `node test-all.mjs --only=update` — 174 passed, 0 failed
- [x] `npm run lint` — 654 files, syntax OK

## Summary

Users can run `update-system.mjs apply --confirm` when root files such as `AGENTS.md` are preserved. Fully preserved paths no longer cause an empty Git pathspec error.

`rollback()` now suppresses expected checkout errors for paths absent from the backup branch.

### Changes

- `update-system.mjs`: Adds and uses `pathFullyPreserved()` to skip fully preserved files and directories. Uses `gitQuiet()` for expected rollback checkout failures.
- `tests/updater-local-system-edits.test.mjs`: Adds coverage for preserved files, fully preserved directories, partial preservation, unrelated paths, and unreadable directories.
- Affected system files: `AGENTS.md`, `modes/`, `DATA_CONTRACT.md`, `providers/`, and `.github/`.
- Orphaned-file cleanup after rollback remains out of scope.

### DIFF
--- a/tests/updater-local-system-edits.test.mjs
+++ b/tests/updater-local-system-edits.test.mjs
@@ -13,7 +13,7 @@ import { mkdtempSync, mkdirSync, writeFileSync, readFileSync, rmSync, existsSync
 import { tmpdir } from 'os';
 import { join } from 'path';
 import { pass, fail } from './helpers.mjs';
-import { gitIn, locallyModifiedSystemFiles } from '../update-system.mjs';
+import { gitIn, locallyModifiedSystemFiles, pathFullyPreserved } from '../update-system.mjs';
 
 // A repo with an `upstream` branch standing in for FETCH_HEAD, and `main` as
 // the install. Both start from a shared base commit, which is what gives
@@ -263,6 +263,108 @@ const PATHS = ['modes/', 'generate-cover-letter.mjs'];
     pass('a fully-excluded pathspec errors — hence the skip in apply() (#2337)');
   } else {
     fail(`#9 errored=${errored} content=${JSON.stringify(content)}`);
+  }
+}
+
+// ── 9b. pathFullyPreserved: a single-file match skips WITHOUT an upstream
+//    lookup ── The reported bug: every one of the reporter's 12 preserved
+//    files (AGENTS.md, fonts/*.ttf, cv-template.*.html, ...) is a single
+//    SYSTEM_PATHS entry, not a directory. The old code still ran an
+//    unnecessary `ls-tree` round-trip to "confirm" what the string match (`f
+//    === path`) had already proven, and when that lookup failed to return the
+//    expected result, execution fell through into the exact cancel-out
+//    checkout case 9 shows fails. Assert no git call happens at all for this
+//    case, so a future regression that reintroduces the lookup is caught even
+//    if the lookup itself would have "worked" in a normal test repo.
+{
+  let gitCalls = 0;
+  const spyCtx = { git: (...args) => { gitCalls++; throw new Error(`unexpected git call: ${args.join(' ')}`); } };
+
+  const preservedPaths = ['AGENTS.md'];
+  const preservedSet = new Set(preservedPaths);
+  const result = pathFullyPreserved('AGENTS.md', preservedPaths, preservedSet, spyCtx);
+
+  if (result === true && gitCalls === 0) {
+    pass('a single-file preserved path is skipped without any upstream lookup');
+  } else {
+    fail(`#9b expected true/0 calls, got result=${result} gitCalls=${gitCalls}`);
+  }
+}
+
+// ── 9c. pathFullyPreserved: a path unrelated to any preserved file is never
+//    skipped, and costs no git call either ──
+{
+  let gitCalls = 0;
+  const spyCtx = { git: (...args) => { gitCalls++; throw new Error(`unexpected git call: ${args.join(' ')}`); } };
+
+  const preservedPaths = ['AGENTS.md'];
+  const preservedSet = new Set(preservedPaths);
+  const result = pathFullyPreserved('modes/pdf.md', preservedPaths, preservedSet, spyCtx);
+
+  if (result === false && gitCalls === 0) {
+    pass('an unrelated path is never skipped, and needs no upstream lookup');
+  } else {
+    fail(`#9c expected false/0 calls, got result=${result} gitCalls=${gitCalls}`);
+  }
+}
+
+// ── 9d. pathFullyPreserved: a directory entirely made of preserved files IS
+//    skipped — the pre-existing behaviour this fix must not regress ──
+{
+  const repo = makeRepo();
+  upstreamChange(repo, 'modes/pdf.md', 'shipped pdf v2\n');
+  // pathFullyPreserved's ls-tree branch reads FETCH_HEAD, same ref apply()
+  // fetches into for real. Populate it here by fetching the local `upstream`
+  // branch into this repo's own FETCH_HEAD.
+  repo.g('fetch', '.', 'upstream');
+  const preservedPaths = ['modes/pdf.md', 'modes/cover.md'];
+  const preservedSet = new Set(preservedPaths);
+  const ctx = { git: (...args) => gitIn(repo.dir, ...args) };
+
+  const result = pathFullyPreserved('modes/', preservedPaths, preservedSet, ctx);
+  if (result === true) {
+    pass('a directory whose entire upstream content is preserved is skipped (ls-tree path)');
+  } else {
+    fail(`#9d expected true, got ${result}`);
+  }
+}
+
+// ── 9e. pathFullyPreserved: a directory only PARTLY preserved is NOT skipped
+//    — the rest of the directory still needs a real checkout ──
+{
+  const repo = makeRepo();
+  upstreamChange(repo, 'modes/pdf.md', 'shipped pdf v2\n');
+  repo.g('fetch', '.', 'upstream'); // populate FETCH_HEAD, see #9d
+  const preservedPaths = ['modes/cover.md']; // pdf.md is NOT preserved here
+  const preservedSet = new Set(preservedPaths);
+  const ctx = { git: (...args) => gitIn(repo.dir, ...args) };
+
+  const result = pathFullyPreserved('modes/', preservedPaths, preservedSet, ctx);
+  if (result === false) {
+    pass('a directory only partly preserved is not skipped');
+  } else {
+    fail(`#9e expected false, got ${result}`);
+  }
+}
+
+// ── 9f. pathFullyPreserved: an unreadable upstream lookup degrades to "not
+//    fully preserved" for a directory, never throws ──
+{
+  const preservedPaths = ['modes/pdf.md'];
+  const preservedSet = new Set(preservedPaths);
+  const throwingCtx = { git: () => { throw new Error('unreadable ref'); } };
+
+  let threw = false;
+  let result = null;
+  try {
+    result = pathFullyPreserved('modes/', preservedPaths, preservedSet, throwingCtx);
+  } catch {
+    threw = true;
+  }
+  if (!threw && result === false) {
+    pass('an unreadable directory lookup degrades to "not fully preserved" instead of throwing');
+  } else {
+    fail(`#9f threw=${threw} result=${result}`);
   }
 }
 
--- a/update-system.mjs
+++ b/update-system.mjs
@@ -1263,6 +1263,72 @@ export function locallyModifiedSystemFiles(paths, upstreamRef = 'FETCH_HEAD', ct
 }
 
 /**
+ * True when checking out `path` from upstream with `preservedPaths` excluded
+ * would leave nothing to check out — i.e. `path` itself is (for a single
+ * file) or entirely consists of (for a `dir/`-suffixed directory) preserved
+ * content. apply()'s checkout loop uses this to skip such an entry outright:
+ * `git checkout FETCH_HEAD -- <path> :(exclude)<path>` errors with "did not
+ * match any file(s)" when the exclusions cancel the whole pathspec, and that
+ * error is indistinguishable from a genuine checkout failure at the call
+ * site, so it would abort the entire update over a file the user asked to
+ * keep.
+ *
+ * A single-file `path` that exactly matches a preserved entry is fully
+ * preserved by definition — the match IS the file's only content, so no
+ * upstream lookup can add information. Only a directory `path` needs the
+ * upstream ls-tree lookup, to confirm EVERY file it would check out is
+ * preserved; an unreadable lookup degrades to "not fully preserved" so the
+ * real checkout runs and reports its own diagnostics, same contract as
+ * `locallyModifiedSystemFiles`.
+ *
+ * Two limits are deliberate, both raised in review of #3781:
+ *
+ * 1. The single-file shortcut diverges from the pre-extraction inline check
+ *    for a preserved file ABSENT from FETCH_HEAD. That check fell through to
+ *    the real checkout, which failed and put the path in apply()'s "Skipped
+ *    N path(s) absent upstream" summary; this returns true and skips it
+ *    silently. Unreachable while preserved paths come from
+ *    `locallyModifiedSystemFiles`, which only reports files that exist
+ *    upstream (it gates each candidate on `cat-file -e <ref>:<file>`), so
+ *    nothing today can construct the case — but it is a real divergence, not
+ *    a behaviour-preserving one, and a future caller sourcing preservedPaths
+ *    some other way would hit it.
+ *
+ * 2. The directory branch's `catch → false` does NOT close the cancel-out
+ *    abort for directories. A throwing ls-tree still falls through to
+ *    `git checkout FETCH_HEAD -- modes/ :(exclude)modes/pdf.md`, which
+ *    aborts the update when the directory happens to be fully preserved.
+ *    That is exactly the pre-extraction behaviour, carried over unchanged —
+ *    this function makes the check testable and drops a redundant lookup for
+ *    the single-file case; it does not fix the directory case. Closing it
+ *    needs a tri-state result whose "unknown" makes the checkout's "did not
+ *    match any file(s)" benign at the call site; tracked separately rather
+ *    than folded in here, since that means matching on git's stderr text.
+ *
+ * @param {string} path - a SYSTEM_PATHS entry, file or `dir/`-suffixed directory.
+ * @param {string[]} preservedPaths - files this run is keeping local content for.
+ * @param {Set<string>} preservedSet - the same paths, as a Set, for lookup.
+ * @param {{git?: Function}} [ctx] - injection point for tests; defaults to gitQuiet.
+ * @returns {boolean}
+ */
+export function pathFullyPreserved(path, preservedPaths, preservedSet, ctx = {}) {
+  if (preservedSet.size === 0) return false;
+  const runGitQuiet = ctx.git || gitQuiet;
+  const isDirectory = path.endsWith('/');
+  const preservedHere = preservedPaths.filter((f) => (isDirectory ? f.startsWith(path) : f === path));
+  if (preservedHere.length === 0) return false;
+  if (!isDirectory) return true;
+  let upstreamFiles = [];
+  try {
+    upstreamFiles = runGitQuiet('ls-tree', '-r', '--name-only', 'FETCH_HEAD', '--', path)
+      .split('\n').map((f) => f.trim()).filter(Boolean);
+  } catch {
+    return false;
+  }
+  return upstreamFiles.length > 0 && upstreamFiles.every((f) => preservedSet.has(f));
+}
+
+/**
  * Preserve byte-for-byte copies of system files before an unavoidable
  * overwrite. The self-bootstrap stage cannot use the normal "keep local"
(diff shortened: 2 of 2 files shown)


update-system.mjs
```diff
@@ -2213,13 +2241,19 @@ function rollback() {
...
+        // the same misleading-stderr class #1998 already fixed for apply()'s
+        // analogous checkout.
+        gitQuiet('checkout', latest, '--', path);
         restored.push(path);
       } catch (err) {
         const pathspec = path.endsWith('/') ? path.slice(0, -1) : path;
         let existedInBackup = true;
         try {
-          git('cat-file', '-e', `${latest}:${pathspec}`);
+          gitQuiet('cat-file', '-e', `${latest}:${pathspec}`);
```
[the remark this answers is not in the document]
euskode: Confirmed and fixed in 9782f52. `cat-file -e` conflated "absent from backup" with "lookup failed for any other reason" (timeout, corrupt object, unreadable ref), so a transient failure could have deleted a path rollback should have restored. Swapped to `ls-tree`, which returns an empty, successful result for an absent path and only throws for a genuine lookup failure — verified manually that `git ls-tree <ref> -- <path>` exits 0 with empty stdout for a non-existent path and non-empty for an existing one, in both cases without throwing. A thrown error now propagates (rethrows the original checkout error) instead of being treated as a green light to delete.

tests/updater-local-system-edits.test.mjs
```diff
@@ -266,6 +266,108 @@ const PATHS = ['modes/', 'generate-cover-letter.mjs'];
   }
 }
 
+// ── 9b. pathFullyPreserved: a single-file match skips WITHOUT an upstream
```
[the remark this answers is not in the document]
euskode: I looked for this guideline before moving anything and can't find it: not in `.coderabbit.yaml`, not in `CONTRIBUTING.md`, not in any `.md`/`.yaml`/`.json` in the repo. The "named `{module}.test.mjs`" rule that *does* exist (CONTRIBUTING.md:186-187) is scoped explicitly to `web/tests/` (the Next.js app), not root `tests/`.

The actual, established convention for root `tests/` coverage of `update-system.mjs` is what I followed: ten existing sibling files (`tests/updater-commit-file-modes.test.mjs`, `updater-drift-detection.test.mjs`, `updater-local-paths.test.mjs`, `updater-nested-checkout-guard.test.mjs`, `updater-preserved-file-prune.test.mjs`, `updater-rollback-behavior.test.mjs`, `updater-status-paths.test.mjs`, `updater-upgrade-safety.test.mjs`, `updater-extract-array-syntax.test.mjs`, plus this PR's `updater-local-system-edits.test.mjs`), all importing from `../update-system.mjs`, none consolidated into a single `update-system.test.mjs`. Adding the new `pathFullyPreserved()` cases to the existing, topically-adjacent `updater-local-system-edits.test.mjs` (which already covers the sibling `locallyModifiedSystemFiles()`/exclude-pathspec behavior this fix depends on) matches that established pattern.

Leaving this as-is unless there's a guideline I'm missing — happy to move it if you can point me to where it's written.

Scott-Emberson (commented):
Re-reviewed at the resolved head 2f50ad4ba1 (the conflict is now cleared). The owned test holds up: pathFullyPreserved on a single-file preserved path returns true with zero git calls (the spy ctx throws on any git call, so a reintroduced ls-tree lookup would trip it), and the directory over-skip guard still reddens on an unpreserved file, proven non-vacuous by a companion fully-preserved-dir case that shows the ls-tree is non-empty. The single-file shortcut is correct because the one preserved file is the checkout's only content, so excluding it cancels the pathspec regardless of upstream. The diff only touches the import line and inserts the new 9b through 9f cases, so the local-edit detection and byte-integrity cases (including CRLF, second-update, and revert) are byte-unchanged.

The substance is update-system.mjs, a red-line file, so the decision defers to santifer, this is a test-axis review. From that axis the owned test is sound and the optimization is behavior-preserving. Nothing blocking from our owned area.

update-system.mjs
```diff
@@ -1262,6 +1262,48 @@ export function locallyModifiedSystemFiles(paths, upstreamRef = 'FETCH_HEAD', ct
...
+  const runGitQuiet = ctx.git || gitQuiet;
+  const isDirectory = path.endsWith('/');
+  const preservedHere = preservedPaths.filter((f) => (isDirectory ? f.startsWith(path) : f === path));
+  if (preservedHere.length === 0) return false;
+  if (!isDirectory) return true;
+  let upstreamFiles = [];
+  try {
+    upstreamFiles = runGitQuiet('ls-tree', '-r', '--name-only', 'FETCH_HEAD', '--', path)
+      .split('\n').map((f) => f.trim()).filter(Boolean);
+  } catch {
```
nikitacometa, line 1325: For `dir/` entries this keeps the exact fall-through the description says it closes: a throwing lookup returns false, and `git checkout FETCH_HEAD -- modes/ :(exclude)modes/pdf.md` then aborts the update (measured). A tri-state return, with the checkout's "did not match any file(s)" treated as benign on unknown, closes both branches.

update-system.mjs
```diff
@@ -1262,6 +1262,48 @@ export function locallyModifiedSystemFiles(paths, upstreamRef = 'FETCH_HEAD', ct
...
+ * @param {{git?: Function}} [ctx] - injection point for tests; defaults to gitQuiet.
+ * @returns {boolean}
+ */
+export function pathFullyPreserved(path, preservedPaths, preservedSet, ctx = {}) {
+  if (preservedSet.size === 0) return false;
+  const runGitQuiet = ctx.git || gitQuiet;
+  const isDirectory = path.endsWith('/');
+  const preservedHere = preservedPaths.filter((f) => (isDirectory ? f.startsWith(path) : f === path));
+  if (preservedHere.length === 0) return false;
+  if (!isDirectory) return true;
```
nikitacometa, line 1320: Differs from main for a preserved single file absent from FETCH_HEAD: main returns false and the path lands in the "Skipped N path(s) absent upstream" summary at 2219; this skips it silently. Unreachable today, since preserved paths come from `locallyModifiedSystemFiles`, but not behaviour-preserving.

update-system.mjs
```diff
@@ -2567,17 +2595,25 @@ function rollback() {
...
-        if (existedInBackup) {
+        // ls-tree, not cat-file -e: cat-file -e throws identically whether
+        // the path is genuinely absent from the backup OR the lookup itself
+        // failed (timeout, a corrupt object, an unreadable ref) — the old
+        // catch-all collapsed both to "absent," which could delete a path
+        // rollback should instead have restored. ls-tree returns an empty,
+        // SUCCESSFUL result for a path absent from the tree and only throws
+        // for a genuine lookup failure, so letting that throw propagate
+        // here (rather than degrading it to "absent") is correct.
+        if (gitQuiet('ls-tree', latest, '--', pathspec).length > 0) {
```
nikitacometa: Un-try'd inside a `catch`: a transient lookup failure ends rollback mid-restore with paths checked out and nothing committed, and the message printed is the ls-tree failure, not `err`. Also drops `--literal-pathspecs`, which `addPaths` / `isTracked` / `expandToShippedFiles` carry. Wrap it, chain `{ cause: err }`, add a case, or fold it into #3782.

update-system.mjs
```diff
@@ -2567,17 +2595,25 @@ function rollback() {
...
     for (const path of SYSTEM_PATHS) {
       try {
-        git('checkout', latest, '--', path);
+        // gitQuiet, not git: failure mode (a) below is an EXPECTED, handled
+        // outcome (the path simply didn't exist in the backup), and printing
+        // git's raw "pathspec did not match" stderr for it reads as a
+        // rollback failure even though the catch handles it correctly —
+        // the same misleading-stderr class #1998 already fixed for apply()'s
+        // analogous checkout.
+        gitQuiet('checkout', latest, '--', path);
```
nikitacometa: Unrelated to the title and to #3779, on the recovery path, with no test. #3782 already rewrites this function; it belongs there.

nikitacometa (changes requested):
This does not fix #3779, and `Fixes #3779` closes it on merge.

#3779 quotes `git --literal-pathspecs add -f -- <hundreds of files>` and `fatal: pathspec ':(exclude)AGENTS.md' did not match any files`. That is `addPaths` (update-system.mjs:1528) fed by `expandToShippedFiles` at 2450, and neither is touched here. Probe at head 2f50ad4b with the issue's repo shape: staging list `["modes/pdf.md", ":(exclude)AGENTS.md"]`, addPaths throws the identical fatal; the same probe on main is byte-identical. The reporter guessed at the checkout skip-check, but main's inline check already returns true for `AGENTS.md` (measured), so that step never failed. #3742 changes the real site.

update-system.mjs:1301 keeps the fall-through the description says it closes, for directories: a throwing ls-tree returns false, and `git checkout FETCH_HEAD -- modes/ :(exclude)modes/pdf.md` then aborts the update. Test 9f pins that as intended.

The rollback edits at 2604 and 2616 are unrelated to the title and untested: 2616 throws inside a `catch`, so a transient lookup failure ends rollback mid-restore with nothing committed and prints the ls-tree message instead of `err`. #3782 rewrites that function already.

The extraction itself is fine (mutant on 1296 reddens 9b; `--only updater` 174/0). Please change to `Refs #3779` and move the rollback change to #3782.

update-system.mjs
euskode: Agreed, and both rollback edits are now out of #3781 — not just the `ls-tree` one at 2616 but this `gitQuiet` swap too, since it's the same recovery-path scope problem. #3781 is now the extraction and its test, nothing else (head `d55a66d4`, 2 files).

They're re-landed in #3782 as `a29708e1`, which already rewrites this function.

update-system.mjs
euskode: Confirmed on all three counts, and this was the most serious of them — thank you. Moved to #3782 (`a29708e1`) with each fixed:

- **Wrapped.** The probe runs inside the checkout's own `catch`, so an escaping throw ended `rollback()` mid-restore with paths checked out and nothing committed.
- **`{ cause: err }` chained.** On an indeterminate lookup it now throws an error naming the path, with the original checkout error as `cause`, instead of surfacing the ls-tree failure and losing the error that started the recovery.
- **`--literal-pathspecs` added**, matching `addPaths` / `expandToShippedFiles`.

Also added two source-pattern cases to `updater-migration-tests.mjs` (the ls-tree probe, and the wrap + cause chain), pinned the same way `rollback()`'s other internals are. Both verified non-vacuous by mutation: reverting the probe reddens the first, dropping `{ cause: err }` reddens the second.

update-system.mjs
euskode: Confirmed — you're right that it isn't behaviour-preserving, and I've stopped claiming it is. Traced it the same way: main falls through to the checkout, which fails and puts the path in the "Skipped N path(s) absent upstream" summary; the shortcut returns `true` and skips it silently.

Unreachable today for the reason you gave — `locallyModifiedSystemFiles` gates each candidate on `cat-file -e <ref>:<file>`, so a preserved path absent upstream can't be constructed — but that's a property of the current caller, not of this function, and a future caller sourcing `preservedPaths` differently would hit it.

Rather than paper over it I documented it on the function itself, since making it exact would mean reinstating the upstream lookup this change exists to remove.

update-system.mjs
euskode: Correct, and the description was wrong rather than the code — the PR body claimed to close a fall-through that, for directories, it carries over from the inline version unchanged (`catch → false` is behaviourally identical to main's `catch {}` plus the `length > 0` guard). Rewrote the body so it only claims what it does: drops a redundant lookup for the single-file case and makes the check testable. The directory limit is now documented on the function too, and test 9f pins it as the current contract rather than as a fix.

On the tri-state — I agree it's the right shape, and filed #3824 for it rather than folding it in here, because the interesting part isn't the tri-state itself but how the call site recognizes the benign cancel-out. Doing it by matching git's "did not match any file(s)" stderr is something this file has twice chosen not to do (#1998 piped stderr instead of matching it; `rollback()`'s catch calls error-string narrowing fragile and uses a structural check). That tradeoff seemed worth settling on its own rather than inside a PR you'd already asked to narrow — happy to take it either way if you'd rather it land as part of this one.

update-system.mjs
```diff
@@ -2146,22 +2212,8 @@ async function apply() {
...
-              .split('\n').map((f) => f.trim()).filter(Boolean);
-          } catch {
-            // Unreadable entry — fall through to the normal checkout, which
-            // reports the real failure with its own diagnostics.
-          }
-          if (upstreamFiles.length > 0 && upstreamFiles.every((f) => preservedSet.has(f))) continue;
-        }
-      }
+      // nothing would be left to check out (see pathFullyPreserved).
+      if (pathFullyPreserved(path, preservedPaths, preservedSet)) continue;
```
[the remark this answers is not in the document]
euskode: Correct diagnosis, but declining it here — this is #3742's change, not this PR's.

You've identified the real #3779 failure exactly: `pathsToStage` keeps `:(exclude)<path>`, `expandToShippedFiles()` passes it through unexpanded, and `addPaths()` hands it to `git add --literal-pathspecs`, which reads it as a literal filename that doesn't exist. I independently traced the same thing earlier today and corrected my own issue report on #3779 to say so.

#3742 ("resolve preserved files before staging so `:(exclude)` never reaches git add") already implements precisely your suggested fix, via a new `stagingFileList()`. Implementing it here too would duplicate that work and conflict with it.

This PR was also just re-scoped down at @nikitacometa's request — it had been carrying unrelated `rollback()` edits, which moved to #3782. It is now only the `pathFullyPreserved()` extraction plus its test, and re-widening it to fix a different step would undo that.

Scott-Emberson (commented):
Test-axis only. The substance is in `update-system.mjs`, which is a red-line file and the maintainer's call; I am reviewing the owned test `tests/updater-local-system-edits.test.mjs`.

The delta since my last look is real (the extract-`pathFullyPreserved()` commit landed, touching the updater and this owned test), and the test is sound. It has five new cases and I mutation-checked the extracted predicate in both directions: gutting it to always-`false` (never preserve, so a real local edit would be clobbered by the cancel-out checkout) reddens the single-file and fully-preserved-directory cases, and gutting it to always-`true` (always skip, so an upstream update is silently dropped) reddens the unrelated-path and partly-preserved-directory cases. The fully-preserved single-file case asserts `gitCalls === 0` via a throwing git spy, the directory cases run against a real populated FETCH_HEAD via ls-tree, and a partly-preserved directory is proven genuine by a sibling case sharing the same fetch setup (a swallowed throw would have reddened that sibling). The degrade-not-throw contract for an unreadable directory lookup is pinned too.

Issue #1837, opened 2026-07-13
feat(pdf): support dynamic CSS custom properties (colors, fonts) via config/profile.yml


SparshGarg999:
### Code of Conduct

- [x] I agree to follow the [Code of Conduct](https://github.com/santifer/career-ops/blob/main/CODE_OF_CONDUCT.md)

### Existing issues

- [x] I searched existing issues and this hasn't been requested yet

### Problem

Currently, `career-ops` CV and Cover Letter templates are styled statically. If a user wants to customize the visual theme (e.g. change the accent color, adjust margins, or select a different font family), they must modify the underlying HTML templates (`cv-template.html` / `cover-letter-template.html`) or install third-party template packs.

This introduces two issues:
1. High barrier to simple design tweaks (requires HTML/CSS editing).
2. Upgrades or template improvements overwrite user modifications.

### Proposed solution

Allow style tokens to be declared in `config/profile.yml` under a `style` block, which the PDF rendering pipeline dynamically injects as CSS custom properties (variables) into the HTML document:

1. In `config/profile.yml`:
   ```yaml
   style:
     accent_color: "#2563eb"     # sleek blue
     font_family: "Outfit, Inter, sans-serif"
     font_size: "10pt"
     margin: "0.5in"
   ```
2. Update `generate-pdf.mjs` and `generate-cover-letter.mjs` to parse this block and prepend a dynamic style block to the HTML before rendering:
   ```html
   <style id="career-ops-dynamic-theme">
     :root {
       --accent-color: #2563eb;
       --font-family: Outfit, Inter, sans-serif;
       --font-size: 10pt;
       --page-margin: 0.5in;
     }
   </style>
   ```
3. Update the default templates to use these CSS variables (e.g., `color: var(--accent-color); font-family: var(--font-family);`).

This allows users to visually align and customize their CV and cover letters with zero HTML modification.

### Area

PDF Generation

santifer:
PR #1862 implements this (dynamic CSS theming via config). Review queued after the current batch.

santifer:
Shipped in #1862 (@abankar1), merged today. `config/profile.yml` now takes a `style:` block with `accent_color`, `font_family`, `font_size` and `margin`, injected as CSS custom properties before the HTML reaches the PDF pipeline.

Both problems you named are covered: no HTML editing needed, and because the templates read `var(--x, <default>)`, a profile with no `style:` block renders byte-identical output, so an upgrade can't overwrite your theme.

[...]

Issue #1145, opened 2026-06-20
Template: break-inside on whole sections inflates CV PDFs to 5 or 6 pages


rubicon:
### Code of Conduct

- [x] I agree to follow the [Code of Conduct](https://github.com/santifer/career-ops/blob/main/CODE_OF_CONDUCT.md)

### Existing issues

- [x] I searched existing issues and this hasn't been reported yet

### What happened?

`templates/cv-template.html` applies `break-inside: avoid` (and `page-break-inside: avoid`) to whole sections (`.avoid-break`) and whole role blocks (`.job`, `.project`). When one of those blocks does not fit the space left on the current page, the browser moves the entire block to the next page, leaving a large gap. Across a full CV this compounds: a resume with roughly 3.3 pages of content renders to 5 or 6 physical pages, most of which is whitespace.

### Steps to reproduce

1. Take any cv.md whose content exceeds one page (several multi-bullet roles, plus Education, Certifications, Skills).
2. Render it through `generate-pdf.mjs`.
3. Sections and roles jump wholesale at page boundaries; page count is far higher than the content volume (e.g., 5 to 6 pages for ~3.3 pages of content).

### Expected behavior

Content should flow across page boundaries and pack normally; only atomic units (header, individual cert/education rows, a single competency or skill) are kept from splitting, headings staying with their content. A representative CV should land on ~2 balanced pages.

### Output of npm run doctor

```shell

```

PR #3789, opened 2026-09-04
fix(templates): make job_break_inside an opt-in style token

### LINKED ISSUE #3242: Make "a work entry never splits across a page break" an opt-in style token
### Problem

Two shipped templates disagree on whether a work entry may split across a page break, and neither is configurable:

- `templates/cv-template.html` — `.job` is **not** in the `break-inside: avoid` list, so a role splits mid-bullets.
- `templates/resume-template.html` — `.job` **is** in the `break-inside: avoid` list, so a role never splits.

The six template packs (`compact`, `executive`, `jake`, `leadership`, `modern`, `zh-minimal`) all follow `cv-template.html`. So the split behavior is 7-to-1, but the odd one out is a shipped, supported template — a user switching templates silently changes pagination semantics.

#1146 was right about the symptom it fixed: broad `break-inside: avoid` inflated a ~3.3-page CV to 5–6 mostly-blank pages. But it removed the only lever, and since then the repo has been patching the resulting symptoms one at a time:

- #1366 — orphaned role titles at page breaks
- #1903 — role header separated from its bullets when a location line is present
- current release — `.project-tech { break-before: avoid }` so a bare tech line doesn't strand at the top of a page

Each is a narrow fix for "*this particular* split looks wrong". For some users the underlying want is different in kind: a work entry is one unit, and a bottom gap is preferable to a reader seeing two half-jobs. At senior/leadership levels a role with 5–6 bullets split mid-list genuinely reads as two separate positions when someone skims the PDF.

Right now the only way to get that is to edit `templates/cv-template.html`, which is a SYSTEM_PATHS file — `update-system.mjs apply` preserves the fork but then the template stops receiving upstream fixes, and `apply --force` wipes it.

### Proposed solution

One new recognized token in the existing #1837 theming mechanism, so it lives in the user layer (`config/profile.yml`) and survives updates:

```yaml
style:
  job_break_inside: avoid    # default: unchanged per template
```

`theme-style.mjs`:

```js
export const STYLE_VAR_MAP = {
  accent_color:     '--accent-color',
  font_family:      '--font-family',
  font_size:        '--font-size',
  margin:           '--page-margin',
  job_break_inside: '--job-break-inside',   // new
};
```

Template side — **each template keeps its current behavior as the `var()` fallback**, so this is strictly additive and no default changes for anyone:

```css
/* cv-template.html + the 6 packs — default stays `auto`, #1145 packing preserved */
.job {
  break-inside: var(--job-break-inside, auto);
  page-break-inside: var(--job-break-inside, auto);
}
```

```css
/* resume-template.html — default stays `avoid`, current behavior preserved */
.job {
  break-inside: var(--job-break-inside, avoid);
  page-break-inside: var(--job-break-inside, avoid);
}
```

No change to the value-coercion logic: the value is a raw CSS keyword, exactly like the existing `accent_color` / `font_family` tokens, and the existing `[;{}<>]` sanitizer already covers it.

### Scope

- `theme-style.mjs` — one line in `STYLE_VAR_MAP`
- the 8 templates carrying a `.job {` CSS rule (`templates/sections/experience.html` is markup-only, no `<style>` — out of scope)
- tests: extend the `styleTokensFrom` coverage for the new token, plus a template-contract assertion that every `.job` rule reads the var
- docs: one line noting the trade-off — with `avoid`, an entry taller than a full page still can't be honored and will split anyway, after a gap

### Why this shape

It doesn't relitigate #1146: the default stays exactly where that PR put it. It converges the `resume-template.html` inconsistency by making the difference explicit and declared rather than by picking a winner and regressing one group of users. And it uses the extension point already built for this class of customization instead of adding a new one.

Happy to open the PR if the direction is right.

Wizarck:
## Summary
- Adds `job_break_inside` to the existing #1837 theming mechanism (`theme-style.mjs` `STYLE_VAR_MAP`), so a user can set `style.job_break_inside: avoid` in `config/profile.yml` to keep every Work Experience entry whole, without forking a `SYSTEM_PATHS` template file.
- Each of the 9 templates carrying a `.job` rule keeps its **current** pagination behavior as the `var()` fallback (`auto` for `cv-template.html` + its 6 packs, matching #1145; `avoid` for `resume-template.html` and `templates/ats/cv-template.ats.html`, unchanged) — strictly additive, no default repaginates anyone's CV on update.
- `templates/sections/experience.html` is markup-only (no `<style>`), out of scope per the issue.

Fixes #3242.

## Test plan
- [x] Extended `tests/theme-style.test.mjs`: `styleTokensFrom` recognizes the new token; a new template-contract block asserts every `.job` rule in all 9 templates reads `var(--job-break-inside, <its own current default>)` with no leftover hardcoded `break-inside`/`page-break-inside` value.
- [x] `node tests/theme-style.test.mjs` — all pass (25/25, including the new assertions)
- [x] `node tests/template-page-breaks.test.mjs` — unaffected, still passes
- [x] `node test-all.mjs` — 7576 passed, 0 failed (14 pre-existing, unrelated warnings: personal-data-leak checks on maintainer's own docs, opt-in live-API test skips, deliberate negative-path fixtures)

## Summary

Users can set `style.job_break_inside: avoid` in `config/profile.yml` without copying system templates. The token maps to `--job-break-inside` in `theme-style.mjs:13`.

The `.job` rules in all shipped templates now use this variable for `break-inside` and `page-break-inside`:

: `auto` remains the default for CV templates and template packs.
: `avoid` remains the default for resume and ATS templates.

Existing users keep the current pagination behavior. Users who select `avoid` can keep a complete work entry together when space permits. Entries taller than one page can still split.

Tests cover token mapping and `.job` template contracts in `tests/theme-style.test.mjs`.

Changed system files: `theme-style.mjs` and `tests/theme-style.test.mjs`. No changes were made to `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, or `.github/`.

### DIFF
--- a/templates/cv-template.html
+++ b/templates/cv-template.html
@@ -16,6 +16,10 @@
        comes from the later @page rule injectPrintPageCss appends (which reads
        this same variable), so this default must match it exactly. */
     --page-margin: 0.6in;
+    /* #3242: unchanged default — a role may still flow across a page break to
+       keep the CV compact (see #1145). Override via config/profile.yml
+       `style.job_break_inside: avoid` for "a role is one unit, never split". */
+    --job-break-inside: auto;
   }
   @page { margin: var(--page-margin); }
 
@@ -444,6 +448,12 @@
   .skill-item {
     break-inside: avoid;
     page-break-inside: avoid;
+  }
+
+  /* #3242: opt-in — see the --job-break-inside default above. */
+  .job {
+    break-inside: var(--job-break-inside, auto);
+    page-break-inside: var(--job-break-inside, auto);
   }
 
   /* Keep a heading attached to the content that follows it, so headings are not
--- a/tests/theme-style.test.mjs
+++ b/tests/theme-style.test.mjs
@@ -15,10 +15,10 @@ try {
   } = await import(pathToFileURL(join(ROOT, 'theme-style.mjs')).href);
 
   // styleTokensFrom: recognized keys → css vars; ignore unknown/non-string/missing
-  const t = styleTokensFrom({ accent_color: '#2563eb', font_family: 'Outfit, sans-serif', font_size: '10pt', margin: '0.5in', nope: 'x', font_weight: 700 });
+  const t = styleTokensFrom({ accent_color: '#2563eb', font_family: 'Outfit, sans-serif', font_size: '10pt', margin: '0.5in', job_break_inside: 'avoid', nope: 'x', font_weight: 700 });
   if (t['--accent-color'] === '#2563eb' && t['--font-family'] === 'Outfit, sans-serif' && t['--font-size'] === '10pt' && t['--page-margin'] === '0.5in'
-      && !('--font-weight' in t) && Object.keys(t).length === 4) {
-    pass('styleTokensFrom maps the 4 recognized keys and ignores unknown/non-string');
+      && t['--job-break-inside'] === 'avoid' && !('--font-weight' in t) && Object.keys(t).length === 5) {
+    pass('styleTokensFrom maps the 5 recognized keys and ignores unknown/non-string');
   } else {
     fail(`styleTokensFrom => ${JSON.stringify(t)}`);
   }
@@ -86,6 +86,36 @@ try {
     if (hasRoot && usesVars && !circular) pass(`${tpl} declares :root theme defaults and reads them via var() (no circular refs)`);
     else fail(`${tpl}: hasRoot=${hasRoot} usesVars=${usesVars} circular=${circular}`);
   }
+  // Template contract (#3242): every shipped template's .job rule reads
+  // --job-break-inside with ITS OWN current pagination behavior as the var()
+  // fallback, so the opt-in token changes nobody's default. resume-template.html
+  // and templates/ats/cv-template.ats.html have always kept a role whole
+  // (avoid); the rest let a role flow across a page break (auto, per #1145).
+  {
+    const expected = {
+      'templates/cv-template.html': 'auto',
+      'templates/cv-template.compact.html': 'auto',
+      'templates/cv-template.executive.html': 'auto',
+      'templates/cv-template.jake.html': 'auto',
+      'templates/cv-template.leadership.html': 'auto',
+      'templates/cv-template.modern.html': 'auto',
+      'templates/cv-template.zh-minimal.html': 'auto',
+      'templates/resume-template.html': 'avoid',
+      'templates/ats/cv-template.ats.html': 'avoid',
+    };
+    for (const [tpl, fallback] of Object.entries(expected)) {
+      const src = readFileSync(join(ROOT, tpl), 'utf-8');
+      const jobRule = src.match(/\.job\s*\{[^}]*break-inside:[^}]*\}/s)?.[0] || '';
+      const usesVar = new RegExp(`break-inside:\\s*var\\(--job-break-inside,\\s*${fallback}\\)`).test(jobRule);
+      const noHardcoded = !/break-inside:\s*(avoid|auto)\s*;/.test(jobRule.replace(/var\([^)]*\)/g, ''));
+      if (usesVar && noHardcoded) {
+        pass(`${tpl}: .job reads --job-break-inside with its own default (${fallback}) as the fallback`);
+      } else {
+        fail(`${tpl}: .job break-inside contract broken => ${jobRule}`);
+      }
+    }
+  }
+
   // Regression: localized CJK body font stacks must honor the profile
   // --font-family override while keeping their curated fallback stacks.
   {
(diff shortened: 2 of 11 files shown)


theme-style.mjs
```diff
@@ -31,10 +32,18 @@ import * as yaml from 'js-yaml';
...
+  font_size:        '--font-size',
+  margin:           '--page-margin',
+  // #3242: whether a Work Experience entry may split across a page break.
+  // Each template supplies its OWN current behavior as the var() fallback, so
+  // this is purely additive — no shipped template's default pagination
+  // changes for anyone who doesn't set this. Raw CSS keyword (e.g. "avoid" or
+  // "auto"), same contract as the other tokens. Trade-off worth knowing:
+  // "avoid" cannot help an entry taller than a full page — it still splits,
+  // just after leaving a bottom gap on the page before it.
+  job_break_inside: '--job-break-inside',
```
nikitacometa, line 46: config/profile.example.yml:122 says "Only these four keys are recognized", four listed at 124-127. It is the only user-facing doc for style tokens and system layer; bump the count and add `job_break_inside`.

templates/cv-template.html
```diff
@@ -16,6 +16,10 @@
        comes from the later @page rule injectPrintPageCss appends (which reads
        this same variable), so this default must match it exactly. */
     --page-margin: 0.6in;
+    /* #3242: unchanged default — a role may still flow across a page break to
+       keep the CV compact (see #1145). Override via config/profile.yml
+       `style.job_break_inside: avoid` for "a role is one unit, never split". */
+    --job-break-inside: auto;
```
nikitacometa, line 22: This `:root` declaration, not the `var()` fallback, is the effective default, and nothing asserts it. Flipping it to `avoid` repaginates every default CV with `theme-style` 35/35 green.

tests/theme-style.test.mjs
```diff
@@ -86,6 +86,36 @@ try {
...
+      'templates/cv-template.leadership.html': 'auto',
+      'templates/cv-template.modern.html': 'auto',
+      'templates/cv-template.zh-minimal.html': 'auto',
+      'templates/resume-template.html': 'avoid',
+      'templates/ats/cv-template.ats.html': 'avoid',
+    };
+    for (const [tpl, fallback] of Object.entries(expected)) {
+      const src = readFileSync(join(ROOT, tpl), 'utf-8');
+      const jobRule = src.match(/\.job\s*\{[^}]*break-inside:[^}]*\}/s)?.[0] || '';
+      const usesVar = new RegExp(`break-inside:\\s*var\\(--job-break-inside,\\s*${fallback}\\)`).test(jobRule);
```
nikitacometa, line 109: Asserts the fallback only; for the 7 templates that carry a `:root` declaration, check that too, or drop those declarations.

tests/theme-style.test.mjs
```diff
@@ -86,6 +86,36 @@ try {
...
+      const noHardcoded = !/break-inside:\s*(avoid|auto)\s*;/.test(jobRule.replace(/var\([^)]*\)/g, ''));
```
nikitacometa, line 110: `break-inside:` also matches `page-break-inside:`. Deleting templates/cv-template.html:455 leaves this green (measured). Anchor the property.

nikitacometa (changes requested):
The wiring is right (`--only theme-style` 35/0; removing the token from `STYLE_VAR_MAP` or flipping resume-template's fallback reddens one test each), and the closed domain is covered: 7 `cv-template*.html`, `resume-template.html`, `ats/cv-template.ats.html`; `cover-letter-template.html` has no `.job`. Three things.

`git merge-tree origin/main pr-3789` conflicts in tests/theme-style.test.mjs; main moved that file 53 lines since base 3a067ee5. Rebase on 71044b4c.

config/profile.example.yml:122 says "Only these four keys are recognized" and lists four at 124-127. This makes five and the file is untouched. It is the only place style tokens are documented (`accent_color` appears nowhere else in *.md or *.yml) and it is system layer, so the wrong count ships.

The contract test asserts the `var()` fallback, but 7 of 9 templates declare the value in `:root` (cv-template.html:22, `.compact`:37, `.executive`:31, `.jake`:39, `.leadership`:38, `.modern`:34, ats:17) and that is what resolves. Flipping cv-template.html:22 to `avoid` repaginates every default CV with `theme-style` 35/35 and `template-page-breaks` 17/17 still green, which is the regression the description promises cannot happen. Assert `:root`, or drop those lines and let the fallback carry it, as zh-minimal and resume-template do. CodeRabbit's open comment is also right: `break-inside:` at 110 matches the tail of `page-break-inside:`, so deleting cv-template.html:455 leaves 35/35 green.

PR #3791, opened 2026-09-04
fix(tests): skip nested checkouts in subtree walkers

### LINKED ISSUE #3762: bug(tests): test-all.mjs discoverTests() walks a nested worktree under tests/ and executes its suites
## Summary

#3499 fixed the walkers anchored at the **repository root**: they now skip a directory carrying its own `.git` marker, so a nested worktree is no longer graded as this checkout's source. Four walkers were in that set, and #3681 covers all four.

`test-all.mjs`'s `discoverTests(dir)` is a fifth recursive walker with the same defect and a **different anchor** — it starts at `ROOT/tests` rather than `ROOT`, which is why it fell outside that enumeration. A worktree placed under `tests/` is discovered and its suites are **executed**.

Reported by @luochen211 in review of #3681.

## Repro

```bash
git worktree add tests/nested-wt HEAD
node test-all.mjs --only nested-wt/tests/mjs-files
git worktree remove tests/nested-wt --force
```

```
  ✅ tests/nested-wt/tests/mjs-files.test.mjs — node:test suite passed (5 tests)

📊 Results: 1 passed, 0 failed, 0 warnings
🟢 All tests passed — safe to push/merge
```

Discovery counts on a checkout with one worktree at `tests/nested-wt`:

```
discovered total:        644
inside nested worktree:  347
```

## Why this is worse than the root-anchored case, and why it is also narrower

**Worse:** every walker fixed in #3499 *reads* files — they syntax-check or pattern-match them. `discoverTests` feeds the runner, so the files it finds are **executed**. A stale checkout's suites run against the current tree, and `test-all.mjs` prints `🟢 All tests passed — safe to push/merge` for them. That is the #3499 failure mode with an execution step bolted on.

**Narrower:** it needs a worktree under `tests/` specifically. The default location — Claude Code's `.claude/worktrees/` — does not reproduce it, which is why #3499's reporter, three reviewers and the fix all missed it. `git worktree add` accepts any path, so nothing prevents it.

## Scope — this is a class, not one function

Several walkers anchor at a subtree and none guards the marker:

| Walker | Anchor | Consumes how |
|---|---|---|
| `test-all.mjs` `discoverTests` | `tests/` | **executes** |
| `test-all.mjs` `walkMjs` (test-all.mjs:15005) | `plugins/` | reads (deny-list scan) |
| `test-all.mjs` web-suite parity walk | `web/tests/` | reads |
| `tests/mojibake-canary.test.mjs` `walkAndCheck` | `templates/`, `modes/` | reads |
| `seed-fixture.mjs` | `test-fixtures/` | reads |

`discoverTests` is the one with teeth. The others are worth auditing in the same pass rather than one issue each — a worktree is unlikely under `templates/`, but "unlikely" is the assumption that produced this issue.

## Suggested fix

`isNestedCheckout` already exists and is exported from `lib/mjs-files.mjs` (#3681). `discoverTests` needs the same guard applied to child directories:

```js
if (entry.isDirectory()) {
  if (isNestedCheckout(full)) continue;
  out.push(...discoverTests(full));
}
```

The root exemption that matters for `collectMjsFiles` does not apply here — `TESTS_DIR` is never itself a checkout root — but the predicate should still only be tested against children, for the same reason.

**Also extend the convention test.** `tests/mjs-files.test.mjs`'s drift gate currently pins three callers by name. A fourth walker was omitted precisely because nothing enumerated it. The durable form asserts the property — every recursive walker in these files consults the predicate — rather than listing the walkers known today. If that is not practical, adding `discoverTests` to the list is the floor, and the list should say what it is a list *of*.

## Environment

- career-ops `main` @ `55ef89b`, plus #3681

luochen211:
## Summary

- skip child directories with their own `.git` marker during test discovery
- apply the same boundary to plugin, web-test, mojibake, and fixture subtree walkers
- add a behavioral regression proving nested checkout suites are not executed, plus per-walker convention checks

Closes #3762

## Test plan

- `node --test tests/mjs-files.test.mjs` (11 passed)
- `node test-all.mjs --only mojibake-canary` (8 passed)
- `node seed-fixture.mjs --self-test`
- `npm run lint` (654 files passed)
- `node test-all.mjs` (8092 passed; 2 unrelated live Gemini checks failed with external `fetch failed`)

## Summary

Nested checkout directories are now excluded from recursive test and fixture scans.

Users can now place a nested worktree below `tests/`, `plugins/`, `web/tests/`, `templates/`, `modes/`, or `test-fixtures/` without causing its tests or files to be discovered, executed, or checked.

### Changes

- `test-all.mjs`: Guards test, plugin, and web-test walkers with `isNestedCheckout`.
- `seed-fixture.mjs`: Skips nested checkouts during fixture traversal and state discovery.
- `tests/mojibake-canary.test.mjs`: Skips nested checkouts during `templates/` and `modes/` scans.
- `tests/mjs-files.test.mjs`: Adds regression coverage and convention checks for all affected walkers.

### DIFF
--- a/seed-fixture.mjs
+++ b/seed-fixture.mjs
@@ -16,6 +16,7 @@ import { tmpdir } from 'os';
 import { join, dirname, relative, sep } from 'path';
 import { fileURLToPath } from 'url';
 import { isMainModule } from './lib/is-main-module.mjs';
+import { isNestedCheckout } from './lib/mjs-files.mjs';
 
 const ROOT = dirname(fileURLToPath(import.meta.url));
 const FIXTURES = join(ROOT, 'test-fixtures', 'upgrade');
@@ -26,7 +27,10 @@ function walk(dir, base = dir, out = []) {
   // platforms (readdirSync returns filesystem order, which varies).
   for (const name of readdirSync(dir).sort()) {
     const p = join(dir, name);
-    if (statSync(p).isDirectory()) walk(p, base, out);
+    if (statSync(p).isDirectory()) {
+      if (isNestedCheckout(p)) continue;
+      walk(p, base, out);
+    }
     else out.push(relative(base, p).split(sep).join('/'));
   }
   return out;
@@ -36,7 +40,10 @@ export function listStates() {
   if (!existsSync(FIXTURES)) return [];
   // Sort for deterministic state ordering across platforms (readdirSync
   // returns filesystem order, which varies).
-  return readdirSync(FIXTURES).sort().filter((n) => statSync(join(FIXTURES, n)).isDirectory());
+  return readdirSync(FIXTURES).sort().filter((n) => {
+    const path = join(FIXTURES, n);
+    return statSync(path).isDirectory() && !isNestedCheckout(path);
+  });
 }
 
 // Strict allowlist: a state must be one of the real fixture subdirectories.
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -112,7 +112,10 @@ function discoverTests(dir) {
   const entries = readdirSync(dir, { withFileTypes: true }).sort((a, b) => (a.name < b.name ? -1 : a.name > b.name ? 1 : 0));
   for (const entry of entries) {
     const full = join(dir, entry.name);
-    if (entry.isDirectory()) out.push(...discoverTests(full));
+    if (entry.isDirectory()) {
+      if (isNestedCheckout(full)) continue;
+      out.push(...discoverTests(full));
+    }
     else if (entry.name.endsWith('.test.mjs')) out.push(full);
   }
   return out;
@@ -14998,7 +15001,10 @@ try {
   const walkMjs = (d) => {
     for (const e of readdirSync(d, { withFileTypes: true })) {
       const fp = join(d, e.name);
-      if (e.isDirectory()) walkMjs(fp);
+      if (e.isDirectory()) {
+        if (isNestedCheckout(fp)) continue;
+        walkMjs(fp);
+      }
       else if (e.name.endsWith('.mjs')) allPluginMjs.push(fp);
     }
   };
@@ -15850,7 +15856,7 @@ try {
         const webTestsRoot = join(ROOT, 'web', 'tests');
         const walk = (dir) => readdirSync(dir, { withFileTypes: true }).flatMap((e) => {
           const p = join(dir, e.name);
-          if (e.isDirectory()) return walk(p);
+          if (e.isDirectory()) return isNestedCheckout(p) ? [] : walk(p);
           return e.isFile() && e.name.endsWith('.test.mjs') ? [p] : [];
         });
         if (existsSync(webTestsRoot)) {
--- a/tests/mjs-files.test.mjs
+++ b/tests/mjs-files.test.mjs
@@ -22,6 +22,7 @@
 
 import { test } from 'node:test';
 import assert from 'node:assert/strict';
+import { spawnSync } from 'node:child_process';
 import { existsSync, mkdtempSync, mkdirSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { dirname, join } from 'node:path';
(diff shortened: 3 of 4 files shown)


Scott-Emberson (commented):
Verified the owned test at this head, and it is sound: the `discovery does not execute suites from a nested checkout` test writes a real `.git` marker plus a suite that throws on execution, runs `test-all.mjs --only`, and asserts exit 1, `no test files matched`, and that the throwing suite never runs. That is a genuine runtime guarantee, not a substring check.

Two things worth weighing before this merges, though, because #3792 targets the same #3762 fix:

1. Security regression in the implementation. This PR adds the nested-checkout guard to `walkMjs`, which is the plugin deny-list scan. Guarding that walk means a plugin can drop a `.git` file next to its sources to make the deny-list scanner skip it, which is opting out of a security scan. That is a real hole rather than a style difference. #3792 deliberately leaves that walk unguarded for exactly this reason.

2. Completeness. This PR pins each walker with a hand-written regex of its current code shape, which is only as complete as the list and brittle to a refactor. #3792 builds a detector that discovers every recursive walker in the repo and asserts each is guarded or explicitly exempt, which is the property #3762 was really about (the walker nobody remembered to add to the list). #3792 also guards `tests/helpers.mjs` `walkFiles` (#3499), which this PR does not touch.

The owned test itself is fine. The concern is the `walkMjs` guard and that #3792 supersedes this on coverage. Deferring the dedup decision to the maintainer.

test-all.mjs
```diff
@@ -14998,7 +15001,10 @@ try {
   const walkMjs = (d) => {
     for (const e of readdirSync(d, { withFileTypes: true })) {
       const fp = join(d, e.name);
-      if (e.isDirectory()) walkMjs(fp);
+      if (e.isDirectory()) {
+        if (isNestedCheckout(fp)) continue;
```
nikitacometa, line 15005: Deny-list opt-out. `walkMjs` feeds the `child_process`/`playwright` scan at 15012 and the revenue firewall at 15020. With `plugins/notion/vendor/.git` + `evil.mjs` using `await import('node:child_process')`, this head PASSES and main FAILS. Remove the guard and keep the reason in a comment, as `plugins/_lock.mjs` does.

tests/mjs-files.test.mjs
```diff
@@ -187,6 +214,44 @@ test('the private repo walkers consult the shared predicate, not their own rule'
...
+  // hide another unguarded one. Pin each recursive edge named in #3762.
+  const assertions = [
+    {
+      caller: 'test-all.mjs',
+      walker: 'test discovery',
+      pattern: /if \(entry\.isDirectory\(\)\) \{\s*if \(isNestedCheckout\(full\)\) continue;\s*out\.push\(\.\.\.discoverTests\(full\)\);/,
+    },
+    {
+      caller: 'test-all.mjs',
+      walker: 'plugin deny-list scan',
```
nikitacometa, line 229: This entry pins the deny-list scan as a walker that MUST be guarded, so it locks the regression above in place. Delete it with the `walkMjs` hunk.

tests/mjs-files.test.mjs
```diff
@@ -187,6 +214,44 @@ test('the private repo walkers consult the shared predicate, not their own rule'
...
 
+test('every subtree-anchored recursive edge checks the nested-checkout boundary', () => {
+  // Checking only that a FILE calls isNestedCheckout is too weak for test-all:
+  // it owns several independent recursive walkers, so one guarded walk can
+  // hide another unguarded one. Pin each recursive edge named in #3762.
+  const assertions = [
+    {
+      caller: 'test-all.mjs',
+      walker: 'test discovery',
+      pattern: /if \(entry\.isDirectory\(\)\) \{\s*if \(isNestedCheckout\(full\)\) continue;\s*out\.push\(\.\.\.discoverTests\(full\)\);/,
```
nikitacometa, line 225: `\s*` between statements means a comment inserted above the guard fails this while behaviour is unchanged (`pass 10 / fail 1`, "must guard its own recursive edge"). Assert the property, not the source shape.

tests/mjs-files.test.mjs
```diff
@@ -162,13 +163,39 @@ test('isNestedCheckout detects the marker, of either type, and nothing else', ()
...
+});
+
 test('the private repo walkers consult the shared predicate, not their own rule', () => {
   // lib/mjs-files.mjs exists so two walkers cannot drift about what "every
   // file" means; the same reasoning applies to what "not our source" means.
   // These three suites keep private walkers because each filters differently
   // (.test.mjs, dot-dirs, SKIP_DIRS sets), so they import the predicate rather
   // than the collector — but a fourth hand-rolled `.git` rule is the drift.
-  for (const caller of ['tests/local-today-gates.test.mjs', 'tests/main-guard-convention.test.mjs', 'test-all.mjs']) {
+  for (const caller of [
```
nikitacometa, line 192: The extended by-name list still cannot reach tests/helpers.mjs `walkFiles`, which recurses under `web/` with a name-matching `skipDirs`. `--only web-test-layout` fails on this head with a worktree at `web/nested-wt`.

seed-fixture.mjs
```diff
@@ -36,7 +40,10 @@ export function listStates() {
   if (!existsSync(FIXTURES)) return [];
   // Sort for deterministic state ordering across platforms (readdirSync
   // returns filesystem order, which varies).
-  return readdirSync(FIXTURES).sort().filter((n) => statSync(join(FIXTURES, n)).isDirectory());
+  return readdirSync(FIXTURES).sort().filter((n) => {
+    const path = join(FIXTURES, n);
+    return statSync(path).isDirectory() && !isNestedCheckout(path);
```
nikitacometa, line 45: `listStates` is a flat listing, not a recursive walker, and it is the allowlist `seedState()` validates against, so this silently drops a fixture state carrying `.git`. No test covers the dropped-state path.

tests/mjs-files.test.mjs
```diff
@@ -162,13 +163,39 @@ test('isNestedCheckout detects the marker, of either type, and nothing else', ()
   }
 });
 
+test('test discovery does not execute suites from a nested checkout', () => {
+  const dir = mkdtempSync(join(ROOT, 'tests', 'nested-checkout-'));
+  const filter = dir.slice(join(ROOT, 'tests').length + 1);
+  try {
+    writeFileSync(join(dir, '.git'), 'gitdir: /elsewhere/.git/worktrees/nested\n');
```
nikitacometa, line 170: Keep this line in whichever PR merges: writing the suite at the fixture's top level is what kills a guard applied to the parent path instead of the child. #3792's fixture nests it one level down and survives that mutant.

nikitacometa (changes requested):
Probed this head (0452dda1), #3792 and main 71044b4c with a linked worktree under `tests/` and `web/` and a shared clone under `templates/` and `plugins/`.

Blocking: test-all.mjs:15005. `walkMjs` is the plugin deny-list and firewall scan over shipped third-party code. With `plugins/notion/vendor/.git` (a plain file is enough) plus an `evil.mjs` containing `await import('node:child_process')`, the gate prints "PASS: no bundled plugin imports child_process/playwright" on this head and "FAIL: bundled plugins import forbidden modules: plugins/notion/vendor/evil.mjs" on main. A plugin opts out of the scan by shipping a marker. Drop that hunk and the matching entry at tests/mjs-files.test.mjs:229; `plugins/_lock.mjs` and `plugin-audit.mjs` already carry the reason in a comment.

Also: tests/helpers.mjs:316 `walkFiles` stays unguarded, so `--only web-test-layout` still fails with a worktree at `web/nested-wt` ("web suites outside the discovered layout (#2360): nested-wt/..."), same as main. And the per-edge regexes redden on a whitespace-only edit: a comment inserted above the guard at test-all.mjs:116 gives `pass 10 / fail 1` with "must guard its own recursive edge" while the edge is guarded.

Two things here beat #3792 and should survive whichever merges: the fixture writes the suite at the checkout's top level (tests/mjs-files.test.mjs:170), which kills a parent/child mutant #3792's deeper fixture survives; and `listStates` is guarded. On the enumeration (12 recursive walkers repo-wide) #3792 covers `walkFiles` and discovers walkers by pattern rather than by name, so it is the one to merge, with those two lines ported.

Issue #3506, opened 2026-08-29
bug(tests): full-suite run writes a stray applications.db — CAREER_OPS_TRACKER_DB pin is a no-op once tracker.mjs is in the module cache


vliggio:
## Summary

A full `node test-all.mjs` run leaves a stray `applications.db` in the repository root. It is written by `tests/tracker-busy-timeout.test.mjs`, whose own comment states the opposite:

```js
// openDb reads DB_PATH once at import time, so point it at a throwaway file
// before importing tracker.mjs — the test never touches a real applications.db.
const work = mkdtempSync(join(tmpdir(), 'cops-busy-'));
process.env.CAREER_OPS_TRACKER_DB = join(work, 'applications.db');
```

The comment correctly identifies that `DB_PATH` is resolved once at import time — and that is exactly why the pin does not hold. Discovered suites are *imported into `test-all.mjs`'s own process*, and an inline section at `test-all.mjs:5035` has already imported `tracker.mjs`:

```js
const { removeRowByNum } = await import(pathToFileURL(join(ROOT, 'tracker.mjs')).href);
```

By the time the suite sets the env var, `tracker.mjs` is in the module cache, module scope has already run, and `DB_PATH` (`tracker.mjs:51`) is bound to the unpinned path. `openDb()` then runs its `CREATE TABLE IF NOT EXISTS` schema against it.

The test passes either way — `PRAGMA busy_timeout` reads back 5000 no matter which file was opened — so nothing flags it.

## Repro

```bash
rm -f applications.db && node test-all.mjs >/dev/null 2>&1 && ls -l applications.db
```

Three ways to confirm the mechanism:

| | result |
|---|---|
| `node test-all.mjs --only .test.mjs` (discovered suites alone) | no file — the pin works, nothing imported `tracker.mjs` first |
| `node test-all.mjs` (full run) | `applications.db` created in the repo root |
| full run with `applications.db` pre-created as a **directory** | exactly one failure: `tracker busy_timeout test crashed: unable to open database file` |

The third names the culprit outright.

## Cause

`tracker.mjs:49-52` resolves both paths at module scope:

```js
const CAREER_OPS = getCareerOpsRoot();
const MD_PATH = resolveTrackerPath(CAREER_OPS);
const DB_PATH = process.env.CAREER_OPS_TRACKER_DB
  || (MD_PATH.endsWith('.md') ? MD_PATH.slice(0, -3) + '.db' : MD_PATH + '.db');
```

[...]

Issue #946, opened 2026-06-11
Tracker parsers break on customized column layouts — detect columns by header name


Narash25:
## Problem

`merge-tracker.mjs`, `verify-pipeline.mjs`, and the Go dashboard (`dashboard/internal/data/career.go`) parse `applications.md` by *fixed column position* (`num=1, date=2, … notes=9`). If a column is added or reordered — for example a **Location** column after Role — every later index shifts by one: Location gets read as Score, Score as Status, and so on. `verify-pipeline.mjs` then reports false "bad row" errors and `merge-tracker.mjs` writes malformed rows.

## Use case

I track job location (it matters a lot when applying across Singapore / Malaysia / remote), so I added a Location column to my tracker. That immediately broke the pipeline scripts, even though the table was still valid markdown.

## Proposal

Map columns by **header name** instead of position: scan for the header row, build a name→index map (with aliases, including the existing `empresa`/`puesto` ES headers), and fall back to the current fixed 9-column layout when no recognizable header is found. Fully backward-compatible — existing trackers are unaffected, and a Location column becomes optional.

## Status

I've implemented this across `merge-tracker.mjs`, `verify-pipeline.mjs`, and `career.go` and have been running it on my own fork. Follows existing code style, adds no dependencies, degrades gracefully. Happy to open a PR referencing this issue if you're open to the direction.

santifer:
Good catch, @Narash25 — and well diagnosed: the fixed-position parsing is real fragility, and your Location use case is legitimate (location filtering shipped for the scanner in #570, so tracking it per-application is a natural next step).

Two paths, both valid:

1. **Short-term (yours if you want it):** header-name column mapping in `merge-tracker.mjs` + `verify-pipeline.mjs`, exactly as you propose — scan the header row once, build a name→index map, tolerate unknown columns. Scope it to the two .mjs parsers first (the Go dashboard can follow as its own PR). With a regression test for an inserted column, that merges fast.

2. **Structural (context worth knowing):** this fragility class is precisely why RFC #918 (SQLite as a derived index over the tracker, endorsed this week) exists — once tools read through a typed layer, column order stops mattering entirely. Your fix and that work are complementary, not competing: header-mapping makes the markdown side robust meanwhile.

Labeling as enhancement. If you open the PR, link it here. Thanks for the clear writeup — repro, cause, and proposal all in one. 🙏

PR #3794, opened 2026-09-04
fix(tracker): carry the tracker layout through the export round-trip (#3703)

### LINKED ISSUE #3703: tracker.mjs: export writes by fixed position what sync read by header name, silently dropping columns
Split out of #3517 (finding 1 of the column-order grep), where the maintainer asked for these to be filed separately.

`tracker.mjs sync` reads the tracker by header NAME and `tracker.mjs export` writes it back by fixed POSITION. So the round-trip is lossy for exactly the layouts the name-based readers were built to support (#946/#954), and it reports success while doing it.

## Reproduction

`data/applications.md` with the Location column from #946's own use case:

```
| # | Date | Company | Location | Role | Score | Status | PDF | Report | Notes |
|---|------|---------|----------|------|-------|--------|-----|--------|-------|
| 1 | 2026-01-01 | Acme | Berlin | Engineer | 4.2/5 | Applied | ❌ | [1](../reports/1.md) | note one |
```

`node tracker.mjs sync`:

```
Indexed 1 applications from …/applications.md into …/applications.db
No corruption detected — index matches the markdown cleanly.
```

`node tracker.mjs export`:

```
| # | Date | Company | Role | Score | Status | PDF | Report | Notes |
|---|------|---------|------|-------|--------|-----|--------|-------|
| 1 | 2026-01-01 | Acme | Engineer | 4.2/5 | Applied | ❌ | [1](../reports/1.md) | note one |
```

Every value is correct — the read side did its job through `resolveColumns` (`tracker.mjs:229`). The COLUMN is gone, and the layout is silently normalized to the default order. Same for `Via` and `URL`: neither exists in the sqlite schema (`tracker.mjs:141-152`), which is a fixed nine columns.

## Why it matters more than a display quirk

`export --out data/applications.md` writes that result back over the user's tracker. It backs up to `.bak` first (`tracker.mjs:544`), so nothing is unrecoverable, but the user is not told anything was dropped — `sync` said the index matched cleanly, and the header comment at `tracker.mjs:503` sells the round-trip as "lossless for clean input", where *clean* silently means *default order*.

The offer is "a repaired copy the user can review and adopt". Adopting it costs a user their Location, Via and URL columns — and losing the URL column is the one that disables merge-tracker's deterministic dedup pass, which is not visible in the file either.

`tracker.mjs query` prints the same fixed `HEADER` (`tracker.mjs:104-105`), but that is display only and harmless.

## Options

1. Carry the layout through the round-trip: persist the source header (a `meta` row already exists) plus the unmapped cells, and have `exportMd` rebuild the table it read rather than the table it assumes. Fully lossless, largest change.
2. Widen the schema to the columns the readers already know (`location`, `via`, `url`) and emit the header from the detected layout, the way `merge-tracker.mjs buildHeaderRows()` does. Covers every column career-ops itself defines; still drops a user's own extra column.
3. At minimum, refuse to be silent: if `detectColumns` finds a layout `exportMd` cannot reproduce, say which columns will be dropped and make `--out` over an existing tracker require confirmation.

(3) is worth doing regardless of which of (1)/(2) lands, since it is the part that turns data loss into a decision.

## Environment

- Node.js on macOS

vliggio:
## What does this PR do?

`tracker.mjs sync` read the tracker by header NAME and `export` wrote it back by fixed POSITION, so the round-trip was lossy for exactly the layouts the name-based readers exist to support (#946/#954) — and it reported success while doing it. Export now rebuilds the table it *read* (option 1 of the issue), and refuses to write over an existing tracker when something genuinely cannot be placed (option 3).

## Related issue

Fixes #3703

## The failure

Every VALUE round-tripped correctly; the COLUMN was gone.

```
$ node tracker.mjs sync
No corruption detected — index matches the markdown cleanly.

$ node tracker.mjs export
| # | Date | Company | Role | Score | Status | PDF | Report | Notes |
```

…from a tracker whose header was `| # | Date | Company | Location | Role | Score | Status | PDF | Report | Notes | URL |`. Because `export --out data/applications.md` writes that back over the user's own tracker, adopting the offered "repaired copy" cost them Location, Via and URL — and losing URL silently disables `merge-tracker.mjs`'s deterministic dedup pass, which is not visible in the file either.

## Option 1 — carry the layout through

- `sync` records the source header and separator lines in `meta`, and every cell no schema field claims goes into a new `applications.extras` column, keyed by its column index.
- `export` rebuilds the header row it saw and re-materializes unmapped cells in their own columns.
- The sqlite schema stays the canonical nine fields — **option 2 (widening the schema) was deliberately not taken**. Extra columns ride along positionally: preserved by `export`, not queryable via `query`. That keeps a user's own column working identically to `location`/`via`/`url`, rather than covering only the three career-ops happens to name.
- `schema_version` joins the content hash in the freshness check. Without it, an index written before `extras` existed matches the md hash while missing everything an export needs — the same silent drop, arriving from a stale schema. A version mismatch forces the rebuild.
- Found along the way: the expected row width came from the highest mapped column index rather than the header, so a trailing user column (`| … | Notes | Priority |`) made every row look one cell too wide and the stray-pipe fold swallowed it into Notes. Width now comes from the header.

Round-trip on the issue's own fixture is byte-identical.

## Option 3 — refuse to be silent

Cells the declared header cannot hold are named on stderr (`column 11 (URL)`), and `--out` over an existing file refuses to write until re-run with `--force`. With option 1 in place the drop list is empty for every layout career-ops can produce; this is the guard that keeps it that way, and it turns the remaining edge cases into a decision rather than a `.bak` line nobody was told to read.

The refusal returns through the `finally` instead of `process.exit` — exiting there would skip the cleanup and hold the tracker lock dir until its 10-minute stale timeout.

## Type of change

- [x] Bug fix
- [ ] New feature

## Tests

Nine regression tests in `tracker-columns-tests.mjs` (46 → 55 passing): byte-identical round-trips for the Location/URL, unknown-column and legacy 9-column layouts; adoption over the tracker via `--out` with a non-canonical status still repaired; the warn → refuse → `--force` path; and the pre-#3703 stale-schema rebuild.

`runScript` in that file now captures stderr separately (tracker.mjs puts diagnostics there so stdout stays pipeable) and pins `CAREER_OPS_TRACKER_DB` into the sandbox, so a test run can never leave a derived index beside real data (#3506).

`node test-all.mjs`: **8094 passed, 0 failed**, 13 pre-existing unrelated warnings.

## Checklist

## Summary

`tracker.mjs` now preserves tracker layouts during `sync` and `export`.

- Stores layout metadata and unmapped cells in SQLite.
- Rebuilds custom, legacy, localized, indented, and CRLF layouts.
- Preserves surrounding content and line endings.
- Rebuilds indexes when the schema version is stale.
- Reports unreproducible cells and content.
- Requires `--force` for lossy overwrites.
- Releases the tracker lock after overwrite refusal.

`docs/SCRIPTS.md:704-719` documents layout preservation, loss reporting, and `--force`.

`tracker-columns-tests.mjs` adds regression coverage. The full suite passes: 8,094 tests.

## User impact

Users can preserve custom columns and tracker structure through `sync` and `export`.

If export cannot reproduce content, it reports the loss. If the output file exists, export refuses the overwrite unless the user supplies `--force`.

The canonical nine-field SQLite schema remains unchanged. Extra columns remain positional and are not queryable.

## Files changed

- `tracker.mjs`
- `tracker-columns-tests.mjs`
- `docs/SCRIPTS.md`


tracker.mjs
```diff
@@ -524,22 +693,40 @@ async function exportMd(args) {
...
-      SEPARATOR,
-      ...rows.map(rowToMarkdown),
-      '',
-    ].join('\n');
+    const meta = db.prepare('SELECT key, value FROM meta').all()
+      .reduce((acc, m) => Object.assign(acc, { [m.key]: m.value }), {});
+    const { header, separator, body, dropped } = renderTable(rows, {
+      header: meta.md_header || '', separator: meta.md_separator || '',
+    });
+    const out = ['# Applications Tracker', '', header, separator, ...body, ''].join('\n');
```
nikitacometa: Fixed skeleton: title, preamble and any second table are dropped here, and `dropped` cannot see it, so `--out` over the tracker exits 0 with no warning (measured with a localized title, a legend line and an archive section).

tracker.mjs
```diff
@@ -499,13 +600,81 @@ async function history(args) {
...
+    }
+    return `| ${cells.join(' | ')} |`;
+  });
+
+  const separator = layout.separator
+    || (layout.header
+      ? `|${labels.map(l => '-'.repeat(Math.max(3, l.length + 2))).join('|')}|`
+      : SEPARATOR);
+
+  const dropped = [...unplaceable].sort((a, b) => a - b)
```
nikitacometa: `dropped` comes only from extras indices past the header width; that is why the `--force` gate at 721 misses non-table content loss.

tracker.mjs
```diff
@@ -221,30 +251,85 @@ function repairPlaceholder(cell) {
...
+ * vanished on export — silently, over the user's own tracker with `--out`.
+ *
+ * @param {string[]} lines - All lines of the tracker markdown.
+ * @returns {{header: string, separator: string|null, labels: string[]}|null}
+ *   null when the file has no recognizable header row (legacy positional table).
+ */
+function detectLayout(lines) {
+  for (let i = 0; i < lines.length; i++) {
+    const t = lines[i].trim();
+    if (!isHeaderRow(t)) continue;
```
nikitacometa: `isHeaderRow` requires the full schema through HEADER_ALIASES. `puntuación` and `estado` are unmapped, so a Spanish header records `md_header: ''` and exports as the English header, exit 0. Record header and separator whenever a table is found; keep LEGACY_COLMAP for field placement.

tracker.mjs
```diff
@@ -221,30 +251,85 @@ function repairPlaceholder(cell) {
...
+
 function parseMarkdownRows(text, diag) {
   const lines = text.split('\n');
   // Map columns by header name (tracker-parse.mjs, #954) so a customized layout
   // (e.g. an inserted Location column) can't shift Score into Status. Falls back
   // to the legacy fixed 9-column layout when no header row is found.
   const colmap = resolveColumns(lines);
-  // Expected `split('|')` width: highest mapped index + the trailing empty cell.
-  const width = Math.max(...Object.values(colmap)) + 2;
+  const layout = detectLayout(lines);
```
nikitacometa, line 333: `resolveColumns` gets raw lines (it needs `startsWith('|')`), `detectLayout` trims them. An indented header gives LEGACY on read and the real map on write; probe: `Berlin` moved one column and `Applied` landed in PDF.

docs/SCRIPTS.md
```diff
@@ -699,15 +699,20 @@ node tracker.mjs sync --check             # diagnose corruption only, no write (
...
 
-`export` is the inverse of `sync` (round-trip `md → db → md` is lossless for clean input — enforced by `test-all.mjs`). It writes to stdout by default and never touches `applications.md` unless you explicitly pass it as `--out`. Phase 2 of #918 (DB becomes source of truth, markdown becomes a rendered view) is a separate, explicit per-user opt-in — not part of this script yet.
+`export` is the inverse of `sync` (round-trip `md → db → md` is lossless — enforced by `tracker-columns-tests.mjs`). It writes to stdout by default and never touches `applications.md` unless you explicitly pass it as `--out`. Phase 2 of #918 (DB becomes source of truth, markdown becomes a rendered view) is a separate, explicit per-user opt-in — not part of this script yet.
```
nikitacometa: "lossless" also fails for CRLF (rewritten to LF) and column-aligned tables (padding collapses). Scope the claim to the table's values and column set.

tracker-columns-tests.mjs
```diff
@@ -702,5 +713,157 @@ if (!HAS_WEB) {
...
+  runTracker(['sync'], sb);
+  const dbFile = join(sb.dir, 'applications.db');
+  const { DatabaseSync } = await import('node:sqlite');
+  const db = new DatabaseSync(dbFile);
+  db.exec("DELETE FROM meta WHERE key IN ('schema_version', 'md_header', 'md_separator')");
+  db.exec("UPDATE applications SET extras = '{}'");
+  db.close();
+  const exported = runTracker(['export'], sb);
+  if (exported.code === 0 && exported.stdout === CUSTOM) {
+    pass('tracker.mjs export: a pre-#3703 index is resynced instead of exporting a lossy table');
```
nikitacometa: This assertion passes with `tracker.mjs` at base, so it does not pin the stale-schema rebuild. Assert the exported header, not only exit 0. (Also: #3765 moves this file into tests/, so one of the two rebases.)

nikitacometa (changes requested):
The layout carry-through works at 754b57e8: legacy 9-column, Location, Via+URL, an unknown `Priority` column and unicode/`N/A` rows all round-trip byte-identically, the refusal path exits 1 and `--force` exits 0, and six of the nine new assertions fail with `tracker.mjs` at base. Two gaps.

tracker.mjs:701 emits a fixed skeleton (`# Applications Tracker`, header, separator, body), so `export --out data/applications.md` deletes the title, any preamble and any second table. `dropped` (tracker.mjs:670) counts only extras past the header width, so the `--force` gate never sees it: a tracker with `# Seguimiento de candidaturas`, a legend line and an `## Archive (2025)` section came back as `# Applications Tracker` plus one table, the archive row re-emitted against the first table's 11-column layout with every cell shifted, exit 0, no warning. docs/SCRIPTS.md:715 ("data loss is a decision you make") states the opposite. Scope the claim to the table, or route the non-table diff through the same gate.

`detectLayout` (tracker.mjs:269) records the header only when `isHeaderRow` resolves the full schema. `puntuación` and `estado` are not in tracker-aliases.json, so a Spanish header (`| # | Fecha | Empresa | Puesto | Puntuación | Estado | PDF | Informe | Notas |`) records `md_header: ''` and exports as the English one, silently, exit 0: the class #3703 filed. Record header and separator whenever a table is found; field placement can keep falling back to LEGACY_COLMAP. #3767 adds `fecha` only, so it does not close this.

Smaller: `resolveColumns` gets raw lines while `detectLayout` trims them, so an indented header reads with LEGACY_COLMAP and writes with the real map (probe loses two cells); "byte-identical" also fails for CRLF and column-aligned tables; and the drop message prints `column 11`, not `column 11 (URL)` as the description says (`labels[i-1]` is undefined there).

docs/SCRIPTS.md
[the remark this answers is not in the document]
vliggio: Scoped in docs/SCRIPTS.md: the round-trip is now stated as byte-identical for a single-table file (title, preamble, header, values, column set, trailing lines, CRLF) and explicitly NOT for cell whitespace or content between table rows — both named on stderr and gated behind `--force`.

docs/SCRIPTS.md
```diff
@@ -699,15 +699,20 @@ node tracker.mjs sync --check             # diagnose corruption only, no write (
...
+
+**Lossless includes the layout, not only the values (#3703).** `sync` maps columns by header NAME, so a customized tracker (a `Location`, `Via` or `URL` column, or one of your own) indexes correctly — but `export` used to write nine fixed columns in a fixed order, so adopting its output cost you those columns with no warning, right after `sync` reported a clean index. Losing the `URL` column in particular disables `merge-tracker.mjs`'s deterministic dedup pass, which is not visible in the file either. `export` now rebuilds the header row it read and puts unmapped cells back in their own columns. The schema itself is still the canonical nine fields — extra columns ride along by position, so they are preserved by `export` but not queryable via `query`.
```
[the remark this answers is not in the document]
vliggio: Not done: `career-ops-docs` is a separate repository, so it is out of scope for this PR. Flagging it for a follow-up there rather than silently skipping.

tracker-columns-tests.mjs
```diff
@@ -41,24 +41,35 @@ function pass(m) { console.log(`PASS ${m}`); passed++; }
...
     });
-    return { code: 0, stdout };
+    if (res.error) throw res.error;
+    const stdout = res.stdout || '';
+    const stderr = res.stderr || '';
+    if (res.status === 0) return { code: 0, stdout, stderr };
+    return { code: res.status ?? 1, stdout: `${stdout}${stderr}`, stderr };
   } catch (e) {
-    return { code: e.status ?? 1, stdout: `${e.stdout || ''}${e.stderr || ''}` };
+    return { code: e.status ?? 1, stdout: `${e.stdout || ''}${e.stderr || ''}`, stderr: `${e.stderr || ''}` };
```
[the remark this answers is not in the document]
vliggio: Applied. `spawnSync` fills both streams on a timeout and `res.error` carries no `.status`/`.stdout`, so rethrowing produced 'code 1, no output'; the reason (plus signal) and both streams are now returned.

tracker.mjs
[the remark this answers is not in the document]
vliggio: Fixed at 0849f91: `sync` records the lines before and after the table and `export` replays them verbatim (title, legend, trailing note, and the file's line endings). Content between the first and last table row goes through the same warning and `--force` gate — replaying a second table's heading and header would frame a row shifted into the first table's columns as intact. docs/SCRIPTS.md scopes the claim accordingly.

vliggio: Reproduced exactly as described and fixed at 0849f91. Prologue and epilogue are recorded and replayed; the archive section is now named on stderr and `--out` exits 1 with the file untouched. Covered by two new tests (non-table content round-trip, archive-section refusal).

tracker.mjs
vliggio: Right — `dropped` was column-only. The gate now takes a combined `losses` list: cells past the header width plus every line between the first and last table row.

tracker.mjs
vliggio: Fixed at 0849f91. `detectLayout` falls back to the markdown shape (the line above the first separator row) when `isHeaderRow` cannot resolve the schema, so the header is recorded verbatim either way; placement still comes from `detectColumns`/LEGACY_COLMAP, exactly as you suggested, so an unresolvable header is never trusted as a column map. The Spanish fixture now round-trips byte-identically.

tracker.mjs
vliggio: Confirmed, and worse than one column: at base the probe gave `Berlin` in Role, `Applied` in PDF and the note folded away. Both now read one trimmed view of the lines; `removeRowByNum` had the same raw-vs-trimmed split and is fixed with it.

docs/SCRIPTS.md
vliggio: Both scoped in docs/SCRIPTS.md. CRLF is now preserved rather than documented as a limit — the ending is recorded and restored (new test). Alignment padding and indentation stay normalized by design, since export renders values instead of replaying raw lines, and that is stated explicitly.

tracker-columns-tests.mjs
vliggio: Rebuilt at 0849f91: the stale index is now constructed with the pre-#3703 schema by hand instead of mutating a current one — against base it aborted on 'no such column' rather than failing — and it asserts the exported header carries `Location`. 14 of the 16 new assertions now fail at base. Noted on #3765; whichever lands second rebases.

tracker.mjs
```diff
@@ -221,30 +251,159 @@ function repairPlaceholder(cell) {
...
+  // Line endings are part of the file, not of the table: splitting on '\n' and
+  // joining the export back with '\n' silently rewrote a CRLF tracker to LF
+  // (PR #3794 review). Remember the dominant ending and restore it on export.
+  const eol = /\r\n/.test(text) ? '\r\n' : '\n';
+  // ONE trimmed view, used by every consumer below. resolveColumns needs
+  // `startsWith('|')` and detectLayout used to trim on its own, so an indented
+  // table gave the legacy map on READ and the real map on WRITE — cells landed
+  // a column off (Berlin in Role, Applied in PDF) on a layout the reader had
+  // never accepted in the first place.
+  const lines = text.split('\n').map(l => l.replace(/\r$/, '').trim());
```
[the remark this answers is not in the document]
vliggio: Fixed at 6f309b6. There are now two views of the file: trimmed for detection/parsing, raw for what is replayed — so prologue and epilogue keep their own indentation and trailing spaces. docs/SCRIPTS.md narrows the claim to what export actually preserves: prose whitespace included, table-cell whitespace excluded (values are rendered, not replayed). New test covers an indented prose line and a trailing indented line.

tracker.mjs
```diff
@@ -221,30 +251,159 @@ function repairPlaceholder(cell) {
...
+    const extras = {};
+    for (let i = 1; i <= lastCell; i++) {
+      if (schemaIndices.has(i)) continue;
+      if (parts[i]) extras[i] = parts[i]; // empty cells re-materialize as '' on export
+    }
+    consumed.add(index);
+    rows.push({
+      cells: [at('num'), at('date'), at('company'), at('role'), at('score'), at('status'), at('pdf'), at('report'), at('notes')],
+      extras,
+    });
```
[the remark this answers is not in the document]
vliggio: Confirmed — a forced export moved the `Oldco` row into the active table with an invented empty URL cell. Fixed at 6f309b6: a second table can only start with a fresh separator row, so that marks where the rebuilt table ends (a blank line does not, since it doesn't change the columns). Those rows stay in the index — `sync`/`query` semantics unchanged — but export names them and a forced export no longer emits them. Added the forced-export assertion you asked for.

tracker.mjs
```diff
@@ -524,22 +807,46 @@ async function exportMd(args) {
...
+    const { header, separator, body, dropped } = renderTable(rows, layout);
+    // The lines around the table are replayed verbatim; the ones inside it that
+    // no single rebuilt table can hold are reported instead, never invented.
+    const out = [...layout.prologue, header, separator, ...body, ...layout.epilogue]
+      .join(layout.eol);
+
+    const losses = [
+      ...dropped,
+      ...layout.interior.map(line => `line between table rows: ${line === '' ? '(blank)' : `"${line}"`}`),
+    ];
```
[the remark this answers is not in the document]
vliggio: Confirmed: `| note | extra |` overwrote the tracker at exit 0 with the value silently changed to `note │ extra`. Fixed at 6f309b6 — `renderTable` returns each rewritten cell with its before/after and they join `losses`, so `--out` now refuses without `--force`.

docs/SCRIPTS.md
```diff
@@ -699,15 +699,33 @@ node tracker.mjs sync --check             # diagnose corruption only, no write (
...
+
+**What "lossless" covers, exactly** — the round-trip `md → db → md` is byte-identical for a file holding ONE table: the title, preamble and trailing lines *with their own whitespace* (they are copied, not re-rendered), the header and separator, every row's values and column set, and CRLF vs LF. Enforced by `tracker-columns-tests.mjs`, including localized headers career-ops cannot name, unknown user columns, indented tables and indented prose.
```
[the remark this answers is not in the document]
vliggio: Correct — a one-table tracker whose Notes cell holds a stray pipe is rewritten, so the claim was too broad. Qualified at 5944d98: byte identity now holds "whenever `export` reports no losses", with a note that a reported loss is exactly the signal it does not hold for that file.

docs/SCRIPTS.md
```diff
@@ -699,15 +699,33 @@ node tracker.mjs sync --check             # diagnose corruption only, no write (
...
+| Reported | Why it cannot be rebuilt |
+|----------|--------------------------|
+| A cell outside the columns the header declares | The header has no column to put it in |
+| A line between the first and last table row (a heading, a second table's header) | A single rebuilt table has nowhere to put it |
+| A row belonging to a later table (an `## Archive (2025)` section) | It was indexed against the **first** table's header, so re-emitting it there would move archived data into the active table under columns it never had |
+| A cell whose value had to be rewritten to survive as a cell (a stray `\|` folded into Notes comes back as `│`) | The value itself changed |
+
+Data loss is a decision you make, not a side effect of adopting a repaired copy.
 
 **Exit codes:** `0` success, `1` validation error, missing prerequisites (Node < 22.5, no `applications.md` to index), or corruption found by `sync --check`.
```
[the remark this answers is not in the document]
vliggio: Added at 5944d98. The exit-code line now lists `export --out` refusing a lossy overwrite, plus the fact that nothing is written in that case — true of every exit-1 path in export, since the write is the last step and atomic, so `1` from `export --out` always means the target is untouched.

docs/SCRIPTS.md
```diff
@@ -699,17 +699,35 @@ node tracker.mjs sync --check             # diagnose corruption only, no write (
...
+
+**What "lossless" covers, exactly** — the round-trip `md → db → md` is byte-identical for a file holding ONE table **whenever `export` reports no losses**: the title, preamble and trailing lines *with their own whitespace* (they are copied, not re-rendered), the header and separator, every row's values and column set, and CRLF vs LF. Enforced by `tracker-columns-tests.mjs`, including localized headers career-ops cannot name, unknown user columns, indented tables and indented prose. A reported loss is exactly the signal that byte identity does **not** hold for that file — a one-table tracker whose Notes cell holds a stray `|` is rewritten, and says so.
```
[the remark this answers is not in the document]
vliggio: Valid, and sharper than described: `aplicado` is a recognized *alias* in `templates/states.yml`, so `sync --check` prints "No corruption detected" and exits 0 while the export still rewrites the cell — tying the guarantee to the exit code would have been wrong too.

Not fixed by reporting canonicalization as a loss: `export` is offered as "a repaired copy you can review and adopt", so the repair is the feature, and gating it behind `--force` would gate the primary use case. Fixed at 3321e2b by narrowing the claim instead — the guarantee is now structural (layout, and every value export does not deliberately repair), with byte identity claimed only for a one-table file already in canonical form and reporting no losses. The three things that change bytes without being losses are listed, and readers wanting to know what a run would change are pointed at `node tracker.mjs export | diff data/applications.md -` rather than an exit code.

docs/SCRIPTS.md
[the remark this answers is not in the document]
vliggio: Follow-up filed: career-ops-hq/career-ops-docs#61 — covers both pages, plus the `--out`/`--force` and loss-warning behavior and the byte-identity-vs-repairs distinction.

Worth noting one correction to the original finding: the blog page's actual inaccuracy is the sentence "The columns are fixed so the tooling and the dashboard agree on the shape" (line 42), which has been wrong since #946/#954 independently of this PR — the schema is fixed, the layout is not.

Issue #3796, opened 2026-09-04
Four evaluators carry private tsvSafe/normalizedTrackerScore copies, and gemini-eval's has drifted into writing an unparseable score cell


vliggio:
Split out of #3707, which converted `openai-eval.mjs` and `ollama-eval.mjs` onto the tracker-addition path and, following the sibling evaluators' pattern, gave each its own local copy of the helpers.

Four evaluators now carry private copies of `tsvSafe` / `normalizedTrackerScore` (and three of `slugifyCompany`):

- `gemini-eval.mjs:269-284`
- `batch-evaluate-gemini.mjs:103-115`
- `openai-eval.mjs:215-253` (added in #3707)
- `ollama-eval.mjs:154-192` (added in #3707)

## The copies have already drifted, and one of them is wrong

`normalizedTrackerScore` exists in two different versions:

```js
// gemini-eval.mjs:280
function normalizedTrackerScore(value) {
  const clean = tsvSafe(value);
  if (!clean || clean === '?') return 'N/A';
  return /\/5$/i.test(clean) ? clean : `${clean}/5`;   // appends to whatever it got
}

// batch-evaluate-gemini.mjs:111
function normalizedTrackerScore(value) {
  const clean = tsvSafe(value);
  if (!clean || clean === '?' || /n\/?a/i.test(clean) || isNaN(parseFloat(clean))) return 'N/A';
  return /\/5$/i.test(clean) ? clean : parseFloat(clean) + '/5';   // parses first
}
```

`gemini-eval.mjs`'s version concatenates rather than parses, so anything the model puts in `SCORE:` beyond the bare number is carried into the score cell. A model answering `SCORE: 4.2 (strong fit)` writes:

```
4.2 (strong fit)/5
```

which is not a score cell:

```
$ node --input-type=module -e "import {looksLikeScoreCell} from './tracker-parse.mjs'; \
    console.log(looksLikeScoreCell('4.2 (strong fit)/5'))"
false
```

`SCORE_CELL_RE` is `/^\d+(?:\.\d+)?\/5$/` and the sentinel list is `N/A` / `DUP` / `—` / `-`, so cell 6 reads as neither a score nor a sentinel. That is the ambiguous row `merge-tracker.mjs` refuses with *"cannot tell score from status in columns 5–6"* — the evaluation is skipped, loudly but wholly. `batch-evaluate-gemini.mjs`'s version has been immune the whole time, for the same input. Neither file records that the other exists.

[...]

Issue #3795, opened 2026-09-04
gemini-eval / openai-eval / ollama-eval: reports carry no `**URL:**` header, so their rows can never reach the URL dedup key


vliggio:
Split out of #3707, which converted `openai-eval.mjs` and `ollama-eval.mjs` onto the tracker-addition path and left this next to it.

`AGENTS.md` Pipeline Integrity rule 3: *"All reports MUST include `**URL:**` in the header (between Score and PDF)."* Three of the five evaluators don't.

| Evaluator | `**URL:**` in report header |
|---|---|
| `batch-evaluate-gemini.mjs:283` | ✅ `**URL:** ${url}` |
| `openrouter-runner.mjs:709` | ✅ `**URL:** ${input \|\| '(pasted)'}` |
| `gemini-eval.mjs:477` | ❌ |
| `openai-eval.mjs:474` | ❌ |
| `ollama-eval.mjs:414` | ❌ |

All three write the same header block — Date / Archetype / Score / Legitimacy / PDF / Tool — with the URL line simply absent:

```js
const reportContent = `# Evaluation: ${company} — ${role}

**Date:** ${today}
**Archetype:** ${archetype}
**Score:** ${score}/5
**Legitimacy:** ${legitimacy}
**PDF:** pending
**Tool:** Gemini (${modelName})
```

## Why it isn't only a missing line

None of the three *has* a URL to write. They accept JD text or `--file` and nothing else — `openai-eval`'s `--url` is the OpenAI-compatible endpoint base, not the posting. So the fix needs an input (a `--posting-url` flag, or reading a URL out of the pipeline entry that dispatched the evaluation), not just a template edit. Where the JD really was pasted, `openrouter-runner.mjs`'s `(pasted)` is the honest value and keeps the field present.

## What the omission costs

The report is where the posting URL survives after the posting dies, and it is also the only place `merge-tracker.mjs --backfill-urls` looks: `resolveReportUrl()` (`merge-tracker.mjs:233`) reads the tracker's report link, opens that report, and pulls `**URL:**` out of it. A report without the line yields `reason: 'no-url'`, so the row can never be backfilled and stays outside the deterministic URL dedup key — the tier `merge-tracker` tries *first*, and the one that can prove two same-title rows are different openings. Every evaluation from these three evaluators falls back to report-number / fuzzy company+role matching, permanently.

`verify-pipeline.mjs` does not check for the header, so nothing flags it today.

## Suggested fix

[...]

PR #3797, opened 2026-09-04
fix(eval): persist openai/ollama evaluations as tracker additions (#3707)

### LINKED ISSUE #3707: openai-eval / ollama-eval: print a hand-paste tracker row that no reader can parse, instead of writing a TSV addition
Split out of #3517 (finding 2 of the column-order grep). The other three are #3703, #3704 and #3705.

`openai-eval.mjs:453-454` and `ollama-eval.mjs:393-394` finish an evaluation by printing a markdown row for the user to paste:

```js
console.log(`\n📊  Tracker entry (add to data/applications.md):`);
console.log(`    | ${num} | ${today} | ${company} | ${role} | ${score}/5 | Evaluated | ❌ | [${num}](reports/${filename}) |`);
```

Three things are wrong with it, and they compound.

## 1. It tells the user to do the one thing the pipeline forbids

`AGENTS.md` Pipeline Integrity rule 1: *"NEVER edit applications.md to ADD new entries — write TSV in `batch/tracker-additions/` and let `merge-tracker.mjs` merge."* The other three evaluators comply — `gemini-eval.mjs`, `batch-evaluate-gemini.mjs` and `openrouter-runner.mjs` all write an addition file. These two are the exception, and they instruct the user out of the dedup, status validation, report-link normalization and locking that merging exists to provide.

## 2. The row it prints is silently invisible

It is 8 cells; the default tracker is 9 columns. A row missing an interior cell is exactly what `parseTrackerRow`'s width guard rejects — so every reader drops it, and nothing says so.

Pasted into a stock tracker:

```
| 1 | 2026-01-01 | Acme | Engineer | 4.0/5 | Applied | ✅ | — | seed row |
| 2 | 2026-02-02 | Globex | Manager | 4.5/5 | Evaluated | ❌ | [2](reports/2-globex-2026-02-02.md) |
```

```
$ node stats.mjs --summary
Tracker:    1 total | 1 active | …          # the second row is not there

$ node verify-pipeline.mjs
✅ All rows properly formatted
📊 Pipeline Health: 0 errors, 0 warnings
🟢 Pipeline is clean!
```

The row is skipped as a non-data line, so the format check passes it rather than flagging it. The user pasted an evaluation into their tracker, saw a clean bill of health, and has no row.

## 3. It is positional, and assumes the default layout

Column order is user-reorderable and resolved by name everywhere else (#946/#954). This literal hard-codes score-before-status, and carries no `Via`, `Location` or `URL` cell — so even at the right width it would land shifted in a tracker with any of those columns, and would sit outside the deterministic URL dedup for the same reason the web prompt was fixed in #1298.

## Suggested fix

Make these two evaluators write a tracker addition like the other three, and print "run `node merge-tracker.mjs`" instead of "add to data/applications.md". Post-#3517 that means the headed form, from the shared header constant — see #3706, which converts the other three.

If a paste-able line is still wanted as a convenience, it must be built from the tracker's *detected* layout rather than a literal, and it should be labelled as a preview, not as the write path.

Worth noting `test-all.mjs:9457` already treats these four files as one family ("all headless evaluators use the shared atomic report allocator"). The same list would make a natural place to assert that every evaluator persists through the TSV path, which is what would have caught this.

## Environment

- Node.js on macOS

vliggio:
## What does this PR do?

`openai-eval.mjs` and `ollama-eval.mjs` now persist an evaluation as a tracker addition in `batch/tracker-additions/`, the way `gemini-eval.mjs`, `batch-evaluate-gemini.mjs` and `openrouter-runner.mjs` already do, and print `run node merge-tracker.mjs` instead of a markdown row to paste into `data/applications.md`.

## Related issue

Fixes #3707.

## Why the printed row was worse than it looked

The three problems in the issue compound: the paste instruction contradicts Pipeline Integrity rule 1, the row it printed was 8 cells against a 9-column tracker, and it hard-coded score-before-status positionally.

The second one is what makes it a silent failure. An 8-cell row is exactly what `parseTrackerRow`'s width guard rejects, so every reader skips it as a non-data line — including `verify-pipeline.mjs`, which reports the tracker clean rather than flagging it. The user pastes an evaluation into their tracker, gets a green bill of health, and has no row.

## What changed

**Both evaluators** write the nine-field TSV their siblings write — `#`, date, company, role, **status**, score, PDF, report link, notes — status before score per the TSV contract, with `merge-tracker.mjs` performing the swap into the tracker's own name-resolved column order. That puts them inside dedup, status validation, report-link normalization and the tracker lock, which the paste instruction routed around.

Notes carry the tool and model (`OpenAI-compatible evaluation (gpt-4o-mini)`), matching `gemini-eval.mjs`'s `Gemini evaluation`.

**No preview line.** The issue allows one if it is built from the tracker's *detected* layout and labelled as a preview. A literal cannot be, so keeping one would have preserved problem 3 with a friendlier label.

**Two small correctness fixes in the same block:** an unparseable or missing score becomes the documented `N/A` sentinel rather than an empty cell, since a blank score column makes the row ambiguous to the column-swap guard and gets it skipped with a warning (#1799); and `companySlug` falls back to `unknown` instead of producing an empty slug (`042--2026-09-03.md`) for an all-punctuation company name. Both mirror the helpers already in `gemini-eval.mjs`.

**Posting URL** (added in review). Both evaluators take `--posting-url` and write `**URL:** ${postingUrl || '(pasted)'}` between Score and PDF, as Pipeline Integrity rule 3 requires — `openai-eval`'s existing `--url` is the endpoint base, so the posting needed its own flag. A real URL is also appended as the optional tenth TSV field (detected by its `http(s)` prefix, `merge-tracker.mjs:697`), so the row reaches the URL dedup tier at merge time instead of waiting for `--backfill-urls`. `(pasted)` cannot poison the column: `normalizeUrl` yields no key for it, so `resolveReportUrl` reports `no-url` rather than handing every pasted row the same key. A `--posting-url` that is not `http(s)` is refused.

## Tests

The issue notes that `test-all.mjs:9587` already treats these four files as one family, and that the same list is where this would have been caught. This PR adds that assertion next to it:

- every headless evaluator persists through `batch/tracker-additions` and none dictates a hand-pasted row (all four checked, guarded against both the old wording and a printed pipe row);
- a field-order pin for the two converted writers, so the nine fields cannot be reordered back into a transposition.

Both fail against the parent commit and pass here — verified by running the regexes against `git show HEAD~1:openai-eval.mjs`.

End-to-end in a temp workspace, the emitted line merges, lands score and status in their own columns, and gets its report link rewritten to `../reports/…`; `stats.mjs` then counts the row the pasted form left invisible.

`node test-all.mjs`: **8096 passed, 0 failed**, 13 warnings — all pre-existing (personal-data scan on `funding.json` / `HIRED.md`), none in touched files. `npm run lint` clean.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

## Not in this PR

Two adjacent gaps, filed separately:

- **#3795** — the `**URL:**` report header. **Addressed here for these two files** after review showed this PR is what puts the affected rows in the tracker; `gemini-eval.mjs` still lacks it, so the issue stays open for that one.
- **#3796** — `tsvSafe` / `normalizedTrackerScore` / `slugifyCompany` now exist in four copies, and they have already drifted. `gemini-eval.mjs`'s copy concatenates instead of parsing (fixed separately in #3799), and `batch-evaluate-gemini.mjs`'s still reinterprets a wrong denominator (`8/10` → `8/5`). Review here produced the strictest version of the three, and #3796 has been amended to name it as the survivor.

Both are better done after #3706, which rewrites the same emitters onto a shared header constant. This PR touches only `openai-eval.mjs`, `ollama-eval.mjs` and `test-all.mjs` to stay clear of it.

## User-visible changes

- `openai-eval.mjs` and `ollama-eval.mjs`: Evaluations now write headed TSV files to `batch/tracker-additions/`.
- Run `node merge-tracker.mjs` to update `data/applications.md`.
- Manual tracker-row pasting is no longer supported.
- Tracker additions use status-before-score order.
- Missing or invalid scores become `N/A`.
- Missing company slugs become `unknown`.
- `--posting-url` accepts HTTP(S) URLs and records them in the report and tracker addition.
- `CAREER_OPS_ADDITIONS` controls the output location.

## Tests

- `test-all.mjs`: Verifies TSV persistence, headed output, field order, write sinks, and removal of manual tracker-edit instructions.
- `test-all.mjs`: Verifies the URL-header exemption is no longer required for OpenAI and Ollama evaluators.

## System files


ollama-eval.mjs
```diff
@@ -150,6 +151,47 @@ function readFile(path, label) {
...
+ * `N/A`, and a blank or unrecognized placeholder makes the row ambiguous and
+ * gets it skipped with a warning (#1799).
+ *
+ * @param {string} value - Score as extracted from the model's summary block.
+ * @returns {string} `X.X/5` or `N/A`.
+ */
+function normalizedTrackerScore(value) {
+  const clean = tsvSafe(value);
+  if (!clean || clean === '?' || /n\/?a/i.test(clean) || isNaN(parseFloat(clean))) return 'N/A';
+  return /\/5$/i.test(clean) ? clean : `${parseFloat(clean)}/5`;
```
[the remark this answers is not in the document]
vliggio: Valid, and reachable in a worse form than the example: neither evaluator has a `validateReport` (unlike `gemini-eval.mjs`), so nothing else in the path range-checks the value — `8/10` became `8/5`, `80%` became `80/5`, and a bare `7` became `7/5`, all of which satisfy `SCORE_CELL_RE` and merge as genuine scores into `stats.mjs` averages. Both files now match a leading number with an optional denominator and return `N/A` unless the denominator is 5 and the value is within 0..5; the trailing-prose case models actually produce (`4.2 (strong fit)`) still normalizes to `4.2/5`. Fixed in f15f110.

test-all.mjs
```diff
@@ -9597,6 +9597,48 @@ try {
...
+  // evaluation as a TSV in batch/tracker-additions, so merge-tracker.mjs applies
+  // dedup, status validation, report-link normalization and the tracker lock
+  // (AGENTS.md Pipeline Integrity rule 1). openai-eval/ollama-eval used to print
+  // a markdown row for the user to paste into data/applications.md instead --
+  // 8 cells against a 9-column tracker, which every reader's width guard drops
+  // as a non-data line, so the pasted evaluation vanished and verify-pipeline
+  // still reported the tracker clean.
+  const pastedRowRe = /Tracker entry \(add to|console\.log\(`[^`]*\|\s*\$\{num\}\s*\|/;
+  const unpersistedEvaluators = evaluatorSources
+    .filter(([, source]) => !/tracker-additions/.test(source) || pastedRowRe.test(source))
```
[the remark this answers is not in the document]
vliggio: Agreed — the checks pinned the `trackerFields` literal but not the write, so a source that built the row and never wrote it, or wrote something else, passed both. The field-order check now also requires `trackerFields.join('\t')` to reach `writeFileSync(join(PATHS.trackerAdditions, ...))`. Verified by mutation: replacing the written content with a placeholder makes it fail. Fixed in f15f110.

openai-eval.mjs
```diff
@@ -450,8 +503,29 @@ ${evaluationText.replace(/---SCORE_SUMMARY---[\s\S]*?---END_SUMMARY---/, '').tri
     writeFileSync(reportPath, reportContent, 'utf-8');
```
nikitacometa, line 548: The report has no `**URL:**` header, so `resolveReportUrl` (merge-tracker.mjs:233) returns `no-url` and this row never reaches the URL dedup tier. Probe: 7/7 rows `report has no **URL:**`, and a re-eval under a changed title added a duplicate. `**URL:** ${postingUrl || '(pasted)'}` between Score and PDF, as openrouter-runner.mjs:709 does; same in ollama-eval.mjs:443.

openai-eval.mjs
```diff
@@ -59,6 +59,7 @@ const PATHS = {
   cv:        join(DATA_ROOT, 'cv.md'),
   profileYml: join(DATA_ROOT, 'config', 'profile.yml'),
   reports:    join(DATA_ROOT, 'reports'),
+  trackerAdditions: join(DATA_ROOT, 'batch', 'tracker-additions'),
```
nikitacometa: merge-tracker.mjs:43 reads `CAREER_OPS_ADDITIONS` when set; this always writes under `DATA_ROOT`. With the override the instructed merge finds no pending TSV. Resolve from the same variable (and in ollama-eval.mjs).

openai-eval.mjs
```diff
@@ -211,6 +212,58 @@ function readFile(path, label) {
...
+ * @returns {string} `X.X/5` or `N/A`.
+ */
+function normalizedTrackerScore(value) {
+  const clean = tsvSafe(value);
+  // Leading number, with an optional explicit denominator. Trailing prose is
+  // tolerated because models add it ("4.2 (strong fit)"), but a denominator that
+  // is not 5 is refused rather than reinterpreted: parseFloat alone read `8/10`
+  // as 8 and wrote `8/5`, which satisfies SCORE_CELL_RE and merges as a genuine
+  // score. Nothing else range-checks this -- unlike gemini-eval, these two have
+  // no validateReport -- so an out-of-range value would reach stats.mjs averages
```
nikitacometa: `validateReport` does not exist in the repo; the function is `validateEvaluationShape` (gemini-eval.mjs:228, called at 425). Claim right, name wrong; same in ollama-eval.mjs:196.

test-all.mjs
```diff
@@ -9597,6 +9597,54 @@ try {
...
```
nikitacometa: `/tracker-additions/.test(source)` matches a comment. gemini-eval.mjs and openrouter-runner.mjs are not covered by `writeSinkRe` (9638, literal list), so deleting their write keeps both checks green. Run the write-sink check over `evaluatorSources`.

openai-eval.mjs
```diff
@@ -211,6 +212,58 @@ function readFile(path, label) {
...
+  // and score gates as fact. N/A is the honest cell for anything unreadable.
+  const parsed = clean.match(/^(\d+(?:\.\d+)?)\s*(?:\/\s*(\d+(?:\.\d+)?))?/);
+  if (!parsed) return 'N/A';
+  const score = parseFloat(parsed[1]);
+  const denominator = parsed[2] === undefined ? 5 : parseFloat(parsed[2]);
```
nikitacometa: This is the strongest of the four `normalizedTrackerScore` copies; #3796 proposes converging on batch-evaluate-gemini.mjs:111 instead, which returns `8/5`, `80/5`, `7/5` for the same inputs. Amend #3796 to name this one as the survivor.

nikitacometa (dismissed):
The conversion is correct at f15f110d and the reverted run proves both new assertions have teeth (`8095 passed, 2 failed` with the two evaluators at base; head 8097/0). Seven rows built from the head helpers merged with status and score in their named columns, the slug fallback yields `005-unknown.tsv` and a matching report name, and the report number comes from the allocator. The score normalizer added in f15f110 is the strongest of the four copies: `8/10`, `80%`, `7`, `4.2/10` all return `N/A`, where batch-evaluate-gemini.mjs:111 still returns `8/5`, `80/5`, `7/5`; worth amending #3796 to name this version as the survivor rather than the batch one.

Two asks, both one line each.

openai-eval.mjs:493 and ollama-eval.mjs:433 still write no `**URL:**` header, and this PR is what puts those rows in the tracker. merge-tracker.mjs:233 mints the URL key by reading that header, so all 7 rows in my probe came back `Backfill URLs: 0 filled … 7 report has no **URL:**`, and re-evaluating one posting under a slightly different title added a second row. openrouter-runner.mjs:709 shows the honest value with no posting-URL input: `**URL:** ${input || '(pasted)'}`, between Score and PDF.

openai-eval.mjs:62 (and the ollama twin) resolves `trackerAdditions` as `join(DATA_ROOT, 'batch', 'tracker-additions')`, while merge-tracker.mjs:43 reads `CAREER_OPS_ADDITIONS` when set. Under that override the evaluator writes where the merge does not look, the instructed `node merge-tracker.mjs` finds nothing, and the evaluation never reaches the tracker. Resolve the directory from the same override.

Smaller: test-all.mjs:9610 passes on the substring `tracker-additions`, so gemini-eval.mjs and openrouter-runner.mjs are guarded by a comment, not a write (the `writeSinkRe` teeth apply only to the two literal names at 9638); `validateReport` in the new comments does not exist, the function is `validateEvaluationShape` (gemini-eval.mjs:228); and the report header writes `**Score:** ${score}/5` from the raw extraction, so `SCORE: 8/10` renders as `8/10/5` next to a tracker cell of `N/A`.

openai-eval.mjs
vliggio: Fixed in f39a057 — you're right that this PR is what makes it live. Both evaluators now take `--posting-url` and write `**URL:** ${postingUrl || '(pasted)'}` between Score and PDF; a real URL is also appended as the optional tenth TSV field (merge-tracker.mjs:697) so the row reaches the URL tier at merge time rather than waiting for a backfill. `(pasted)` cannot poison the column — `normalizeUrl('(pasted)')` returns nothing, so resolveReportUrl reports `no-url` instead of minting a shared key — and a non-http(s) `--posting-url` is refused. Reproduced your probe: the same posting re-evaluated under a drifted title now updates row #2 instead of adding a second row.

openai-eval.mjs
vliggio: Fixed in f39a057. Both now resolve the additions dir from `CAREER_OPS_ADDITIONS` when set, mirroring merge-tracker.mjs:43, so the merge they instruct looks where they wrote. Verified end to end with the override pointing at a directory outside DATA_ROOT.

openai-eval.mjs
vliggio: Correct — the function is `validateEvaluationShape` (gemini-eval.mjs:228, called at 409). Comment corrected in both files in f39a057; the claim it supports is unchanged.

test-all.mjs
vliggio: Both right, and the comment case was self-inflicted — my own comment contained the string. Dropped that presence check entirely: the write sink is now declared per evaluator and run across `evaluatorSources`, with the table asserted against that list so a fifth evaluator or a rename fails here rather than dropping out of coverage. Mutation-verified in f39a057 by deleting the write in each of the four files in turn; each one fails.

openai-eval.mjs
vliggio: Agreed and done — #3796's suggested fix now names this version as the survivor, with a table showing `8/10`, `4.2/10`, `80%` and `7` returning `N/A` here against `8/5`, `4.2/5`, `80/5`, `7/5` from batch-evaluate-gemini.mjs:111. #3799 fixes gemini-eval's copy but to the looser semantics, so that branch needs the same tightening before the copies can converge.

ollama-eval.mjs
```diff
@@ -132,6 +143,15 @@ if (!jdText) {
   process.exit(1);
 }
 
+// A posting URL is the tracker's deterministic dedup key, so it is taken only in
+// a form that can actually become one. normalizeUrl yields no key for anything
+// else, and a placeholder written into the column would hand every such row the
+// same key.
+if (postingUrl && !/^https?:\/\//i.test(postingUrl)) {
+  console.error(`❌  --posting-url must be an http(s) URL: "${postingUrl}"`);
+  process.exit(1);
```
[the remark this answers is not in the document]
vliggio: Fixed in ee638ab. `isPostingUrl` now parses with `new URL()` and requires an http/https protocol and a non-empty host, so `https://` (which throws) and `ftp://x.co` are refused before anything is written. Confirmed the concern was real: `normalizeUrl('https://')` returns empty, so the value would have sat in the URL column looking like a dedup key while matching nothing.

test-all.mjs
```diff
@@ -9597,6 +9597,64 @@ try {
...
+  };
+  const uncovered = evaluatorSources.map(([name]) => name).filter(name => !writeSinks[name]);
+  const orphaned  = Object.keys(writeSinks).filter(name => !evaluatorSources.some(([n]) => n === name));
+  const pastedRowRe = /Tracker entry \(add to|console\.log\(`[^`]*\|\s*\$\{num\}\s*\|/;
+  const unpersistedEvaluators = evaluatorSources
+    .filter(([name, source]) => !writeSinks[name]?.test(source) || pastedRowRe.test(source))
+    .map(([name]) => name);
+  if (uncovered.length > 0 || orphaned.length > 0) {
+    fail(`write-sink table out of step with the evaluator family: uncovered=${uncovered.join(', ') || 'none'} orphaned=${orphaned.join(', ') || 'none'}`);
+  } else if (unpersistedEvaluators.length === 0) {
```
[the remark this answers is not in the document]
vliggio: No change — the `else { fail(...) }` is already there, at test-all.mjs:9630, immediately after the `unpersistedEvaluators.length === 0` branch. It is what printed `headless evaluators bypass the tracker-addition path: <file>` in each of the four write-sink mutation runs reported on f39a057, so a broken sink does reach `fail()`. Happy to look again if you were reading an earlier commit.

test-all.mjs
```diff
@@ -9808,12 +9886,13 @@ try {
...
+  // Empty since #3797 landed the header in openai-eval.mjs and ollama-eval.mjs,
+  // the two files it exempted. The list is asserted to be EXACTLY the files still
+  // missing the header, so leaving the names here would fail as a stale
+  // exemption -- which is what that guard is for. Kept rather than deleted: the
+  // next evaluator added without a URL header has somewhere to be named
+  // deliberately, instead of the check being quietly weakened to accommodate it.
+  const pendingUrlHeader = [];
   const missingUrlHeader = evaluatorSources
     .filter(([, source]) => !urlHeaderRe.test(source) || !urlSourceRe.test(source))
     .map(([name]) => name);
```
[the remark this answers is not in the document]
vliggio: Fixed in 788eeb1 — valid, and the same class as the write-sink finding in the block directly below, which this check (inherited from #3876 in yesterday's merge) had escaped. It now runs over the comment-stripped source that check already builds.

On the second half: the header must carry an interpolated value, not a bare literal, which is what ties it to the report write path — every writer substitutes one today (`**URL:** ${postingUrl || '(pasted)'}`, `**URL:** ${url}`), so a hardcoded line with nothing substituted in is a report that cannot carry the posting's URL. Mutation-verified for all four evaluators: commenting the header out, and replacing the interpolation with a hardcoded string, both fail. The evaluator-family section run standalone passes all four of its checks.

PR #3799, opened 2026-09-04
fix(eval): parse the score in gemini-eval instead of concatenating it

vliggio:
## What does this PR do?

`gemini-eval.mjs`'s `normalizedTrackerScore` appended `/5` to whatever the summary block carried instead of parsing a number out of it, so trailing prose on the `SCORE:` line went straight into the tracker's score column and got the whole addition refused at merge time.

## Related issue

No issue — bug fix, which CONTRIBUTING exempts. Found while writing #3796 (the helper is duplicated across four evaluators, and this is the copy that drifted). Not a regression from #3707, and independent of it: this branch is off `main` and touches neither file that PR touches.

## The two ends of the pipe disagree on what a score is

`validateReport` checks only the **numeric prefix** of the SCORE line:

```js
// gemini-eval.mjs:257
const score = summaryBlock.match(/^\s*SCORE:\s*([0-9]+(?:\.[0-9]+)?)/mi);
```

`extract('SCORE')` returns the **whole rest of the line**:

```js
// gemini-eval.mjs:446
if (trimmed.startsWith(prefix)) return trimmed.slice(prefix.length).trim();
```

So a model answering `SCORE: 4.2 (strong fit)` validates as 4.2, reaches the tracker builder intact, and becomes:

```
4.2 (strong fit)/5
```

`SCORE_CELL_RE` is `/^\d+(?:\.\d+)?\/5$/` and the sentinel list is `N/A` / `DUP` / `—` / `-`, so the cell reads as neither:

```
$ node --input-type=module -e "import {looksLikeScoreCell} from './tracker-parse.mjs'; \
    console.log(looksLikeScoreCell('4.2 (strong fit)/5'))"
false
```

That is exactly the row `merge-tracker.mjs` refuses with *"cannot tell score from status in columns 5–6 — refusing to merge a possible column swap"*. A completed evaluation — model call paid for, report written — is dropped at the merge. `N/A`, `n/a` and `unknown` took the same path and came out as `N/A/5`.

`batch-evaluate-gemini.mjs:111` parses the float first and has never been affected by any of this. Nothing held the two copies to one behavior.

## The fix

Parse the value instead of pattern-matching the string. Review found two further defects in the guard this PR originally copied from `batch-evaluate-gemini.mjs:111`, pointing in opposite directions:

- **`/n\/?a/i` was unanchored, with an optional slash**, so it matched the bare substring `na`. `4.2 (final)`, `4.2 (internal)` and `4.5 - strong signal` were all recorded as `N/A` — a completed evaluation filed as unscored, the same trailing-prose class this PR set out to fix. `validateEvaluationShape` (`gemini-eval.mjs:228`) already guarantees a leading 0..5 number by the time the value arrives, so the branch could only ever destroy a valid score.
- **No denominator or range check**, and the `/5` early return kept the whole string. `4.2/10` became `4.2/5` and merged as a genuine score, where it previously produced `4.2/10/5` and was refused loudly. A silent wrong number is worse than a loud refusal.

Both copies now use the version reviewed into #3797:

```js
const parsed = clean.match(/^(\d+(?:\.\d+)?)\s*(?:\/\s*(\d+(?:\.\d+)?))?/);
if (!parsed) return 'N/A';
const score = parseFloat(parsed[1]);
const denominator = parsed[2] === undefined ? 5 : parseFloat(parsed[2]);
if (!Number.isFinite(score) || denominator !== 5 || score < 0 || score > 5) return 'N/A';
return `${score}/5`;
```

`batch-evaluate-gemini.mjs` is included rather than left for #3796, because a shared test cannot assert agreement between copies that disagree. That leaves #3796 as deleting three copies rather than deciding semantics.

## Tests

Moved to `tests/evaluator-score-cell.test.mjs` — the repo's own instruction routes numbered sections there, and inline meant `node test-all.mjs --only …` could never reach it. It now does (`--only evaluator-score-cell` → 1 passed).

The evaluators run on import (arg parse + network), so each copy is lifted out of its source and evaluated standalone. Discovery is `readdirSync` over root `.mjs` filtered by the definition of the helper — the earlier filename list quietly excluded any future evaluator — and the suite fails outright if fewer than two copies are found, rather than passing vacuously if the helper moves. It reports 2 here and will report 4 once #3797 lands.

21 inputs are pinned, including every case raised in review: `4.2 (final)`, `4.2 (internal)`, `4.5 - strong signal`, `4.2 (N/A noted)`, `4.2 (strong fit)/5`, `8/10`, `4.2/10`, `7`, `-1`, `80%`. Each result is also asserted against `looksLikeScoreCell`, the property that actually matters downstream.

Restoring either old helper fails 12 assertions, e.g.:

```
❌ batch-evaluate-gemini.mjs: "4.2 (final)" -> "N/A", expected "4.2/5"
   batch-evaluate-gemini.mjs: "8/10" -> "8/5", expected "N/A"
   batch-evaluate-gemini.mjs: "4.2 (strong fit)/5" -> "4.2 (strong fit)/5", which looksLikeScoreCell rejects
```

`node test-all.mjs`: **8096 passed, 0 failed**, 13 warnings — all pre-existing, none in touched files. `npm run lint` clean.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

## Scope

Behavior only, now across both Gemini-side copies. Collapsing the four into one implementation stays #3796 — amended to name this version as the survivor.

## Summary

Evaluator scores now produce valid tracker cells.

- `gemini-eval.mjs:290`: Parses numeric prefixes and accepts trailing prose.
- `batch-evaluate-gemini.mjs:111`: Applies the same normalization rules.
- Both helpers return `N/A` for invalid denominators, out-of-range values, non-score values, and unrelated fractions.
- `tests/evaluator-score-cell.test.mjs:32`: Covers valid scores, trailing prose, invalid values, and tracker-cell validity.

From the user’s point of view: `4.2 (strong fit)` becomes `4.2/5`. Values such as `4.2/10`, `8/10`, `unknown`, `N/A`, and `4.2 (fit 3/4 axes)/5` become `N/A`.

System files touched: `gemini-eval.mjs`, `batch-evaluate-gemini.mjs`, and `tests/evaluator-score-cell.test.mjs`. `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/` are not touched.


gemini-eval.mjs
```diff
@@ -277,10 +277,26 @@ function tsvSafe(value) {
...
+ * a made-up cell -- an unrecognized placeholder gets the row skipped (#1799).
+ *
+ * @param {string} value - Score as extracted from the model's summary block.
+ * @returns {string} `X.X/5` or `N/A`.
+ */
 function normalizedTrackerScore(value) {
   const clean = tsvSafe(value);
-  if (!clean || clean === '?') return 'N/A';
-  return /\/5$/i.test(clean) ? clean : `${clean}/5`;
+  if (!clean || clean === '?' || /n\/?a/i.test(clean) || isNaN(parseFloat(clean))) return 'N/A';
```
nikitacometa: `/n\/?a/i` is unanchored with an optional slash, so it matches bare `na`: `4.2 (final)`, `4.5 - strong signal`, `4.2 (internal)` all become `N/A`. `validateEvaluationShape` (228) guarantees a leading 0..5 number here, so this branch can only destroy a valid score. Drop it; `isNaN(parseFloat(clean))` already covers `N/A`, `n/a`, `unknown`, `?`.

gemini-eval.mjs
```diff
@@ -277,10 +277,26 @@ function tsvSafe(value) {
...
+  return /\/5$/i.test(clean) ? clean : `${parseFloat(clean)}/5`;
```
nikitacometa: No denominator or range check: `4.2/10` becomes `4.2/5` and merges as a genuine score (previously `4.2/10/5`, refused loudly), and `4.2 (strong fit)/5` returns the whole string, still refused. openai-eval.mjs:259-264 (#3797) handles both.

test-all.mjs
```diff
@@ -14257,6 +14257,75 @@ try {
...
+// held the copies to one behavior.
+//
+// The evaluators run on import (arg parse + fetch), so lift each copy out of the
+// source and exercise it directly. Discovery is by definition, not by a fixed
+// list: a new evaluator that grows its own copy is tested the day it lands.
+console.log('\n44g. evaluators — normalizedTrackerScore emits a parseable score cell');
+try {
+  const { looksLikeScoreCell } = await import(pathToFileURL(join(ROOT, 'tracker-parse.mjs')).href);
+  const helperRe = (name) => new RegExp(`^function ${name}\\([\\s\\S]*?\\n\\}`, 'm');
+  const evaluators = ['gemini-eval.mjs', 'batch-evaluate-gemini.mjs', 'openai-eval.mjs', 'ollama-eval.mjs']
```
nikitacometa: Hardcoded four-name array, so the "discovery by definition" claim is not delivered; the head run prints `all 2 normalizedTrackerScore copies`. `readdirSync` plus the definition filter would.

test-all.mjs
```diff
@@ -14257,6 +14257,75 @@ try {
...
+    .filter(name => existsSync(join(ROOT, name)))
+    .map(name => [name, readFile(name)])
+    .filter(([, src]) => helperRe('normalizedTrackerScore').test(src));
+
+  // A score line the model wrote with trailing prose, plus the sentinels every
+  // copy has to agree on. Expectations are the behavior, not one file's code.
+  const cases = [
```
nikitacometa: Both prose cases avoid the `na` substring, so this suite is green against the 298 defect. Add `['4.2 (final)', '4.2/5']`, `['8/10', 'N/A']`, `['7', 'N/A']`; after #3797 lands those three split the four copies apart.

test-all.mjs
```diff
@@ -14257,6 +14257,75 @@ try {
   fail(`gemini-eval systemInstruction source test crashed: ${e.message}`);
 }
 
+// ── 44g. evaluators — the score cell is parsed, never concatenated ────
```
nikitacometa: New numbered inline section; the repo's own instruction routes these to `tests/`, and inline means `--only` never runs it. The helper lifts from source with no import, so it moves as-is.

nikitacometa (changes requested):
The `4.2 (strong fit)` fix is right and the reverted run confirms it (8095 passed, 1 failed with gemini-eval.mjs at base; head 8096/0). Two problems in the guard copied from batch-evaluate-gemini.mjs:111.

gemini-eval.mjs:298: `/n\/?a/i` is unanchored and the slash is optional, so it matches the bare substring `na`. `4.2 (final)` gives `N/A`, `4.5 - strong signal` gives `N/A`, `4.2 (internal)` gives `N/A`: the same trailing-prose class this PR fixes. `validateEvaluationShape` (228) already guarantees a leading number in 0..5 here, so this branch can never catch a real `N/A`; it can only destroy a valid score, and the tracker records `N/A` for a completed 4.2. `isNaN(parseFloat(clean))` already covers `N/A`, `n/a`, `unknown` and `?` (verified with the branch removed).

gemini-eval.mjs:299: no denominator or range check, and the `/5` early return keeps the whole string. `4.2/10` becomes `4.2/5`, accepted by `looksLikeScoreCell` and merged as a real score, where before it produced `4.2/10/5` and was refused loudly; `4.2 (strong fit)/5` still comes back verbatim and is refused. openai-eval.mjs:259-264 in #3797 rejects both (same author, same day, after the same CodeRabbit finding); adopt that form here.

test-all.mjs:14277 is a hardcoded four-name array narrowed by a regex, so "discovery by definition" is not delivered (the head run prints `all 2 copies`), and both prose cases avoid the `na` substring, so the suite is green against the defect above; after #3797 lands the same list certifies four implementations that disagree on `4.2 (final)`, `8/10`, `80%` and `7`. #3796 should name #3797's version as the survivor, not this one. The description's `validateReport` is `validateEvaluationShape`, and line 446 is blank at head (the `extract` return is 461).

gemini-eval.mjs
vliggio: Fixed in 04b9901 — and the failure was worse than an edge case: `4.2 (final)`, `4.2 (internal)` and `4.5 - strong signal` all contain the bare substring `na`, so a completed 4.2 was filed as unscored. You're right that `validateEvaluationShape` (gemini-eval.mjs:228) guarantees a leading 0..5 number by this point, so the branch could only destroy a valid score. Dropped; parsing covers `N/A`, `n/a`, `unknown` and `?`.

gemini-eval.mjs
vliggio: Fixed in 04b9901 by adopting openai-eval.mjs:259-264 from #3797 verbatim: a leading number with an optional denominator, refused unless the denominator is 5 and the value is within 0..5. `4.2/10` and `8/10` now return `N/A` rather than merging as `4.2/5` / `8/5`, and `4.2 (strong fit)/5` normalizes to `4.2/5` instead of coming back verbatim. `batch-evaluate-gemini.mjs` is included in the same commit, so all copies now agree and #3796 is down to deleting three of them.

test-all.mjs
vliggio: Fixed in 04b9901. Discovery is now `readdirSync` over root `.mjs` filtered by the definition of the helper, so the claim matches the code; it also fails outright if fewer than two copies are found, rather than passing vacuously if the helper moves. It still reports 2 on this branch because openai-eval and ollama-eval grow their copies in #3797 — it will report 4 once that lands.

test-all.mjs
vliggio: Added in 04b9901, all three, plus CodeRabbit's `4.2 (strong fit)/5` and `4.2 (N/A noted)`. You're right that they split the copies apart: restoring either old helper now fails 12 assertions across the two files, including `8/10 -> 8/5` and `4.2 (final) -> N/A`.

test-all.mjs
vliggio: Moved in 04b9901 to `tests/evaluator-score-cell.test.mjs`, and you were right that it lifts from source with no import so it moved unchanged. Confirmed `node test-all.mjs --only evaluator-score-cell` now reaches it (1 passed), which it could not while the section was inline.

gemini-eval.mjs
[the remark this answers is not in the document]
vliggio: Fixed in 04b9901, taking the direction but not the exact diff: `parseFloat` + a finite check alone would still turn `8/10` into `8/5`, which `looksLikeScoreCell` accepts and `merge-tracker` then merges as a genuine score. The version now in both files parses a leading number with an optional denominator and returns `N/A` unless the denominator is 5 and the value is within 0..5. Both of your cases are pinned in the new suite: `4.2 (strong fit)/5` and `4.2 (N/A noted)` each normalize to `4.2/5`.

test-all.mjs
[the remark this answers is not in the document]
vliggio: Fixed in 04b9901 — moved to `tests/evaluator-score-cell.test.mjs`, section 44g removed from the harness. It is auto-discovered (`tests/**/*.test.mjs`) and `node test-all.mjs --only evaluator-score-cell` now reaches it, which was impossible inline.

test-all.mjs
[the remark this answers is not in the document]
vliggio: Fixed in 04b9901. The filename list is gone: the suite now runs `readdirSync` over root `.mjs` files and keeps those that define `normalizedTrackerScore`, so the coverage matches the claim, and it fails rather than passing vacuously if fewer than two copies are found.

gemini-eval.mjs
```diff
@@ -277,10 +277,31 @@ function tsvSafe(value) {
...
-  if (!clean || clean === '?') return 'N/A';
-  return /\/5$/i.test(clean) ? clean : `${clean}/5`;
+  // Parse, do not pattern-match the string. Two bugs lived in the old guard:
+  // `/n\/?a/i` was unanchored with an optional slash, so bare `na` matched and a
+  // real score with trailing prose -- `4.2 (final)`, `4.2 (internal)`,
+  // `4.5 - strong signal` -- was recorded as `N/A`; and the `/5` early return kept
+  // the whole string, so `4.2/10` became `4.2/5` and merged as a genuine score.
+  // Trailing prose is tolerated because models produce it; a denominator that is
+  // not 5, or a value outside 0..5, is refused rather than reinterpreted.
+  const parsed = clean.match(/^(\d+(?:\.\d+)?)\s*(?:\/\s*(\d+(?:\.\d+)?))?/);
```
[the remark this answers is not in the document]
vliggio: Fixed in 6a0eeca — confirmed `4.2 (strong fit)/10` returned `4.2/5`, the same wrong number `8/10` used to produce by a different route. The first denominator anywhere in the cell is now taken and must be 5, so annotations no longer hide it. A cell whose first fraction is unrelated (`4.2 (fit 3/4 axes)`) is refused rather than guessed at — N/A is recoverable, a wrong score is not. `4.2 (strong fit)/10 -> N/A` added to the suite.

gemini-eval.mjs
```diff
@@ -277,10 +277,31 @@ function tsvSafe(value) {
...
+  if (!parsed) return 'N/A';
+  const score = parseFloat(parsed[1]);
+  const denominator = parsed[2] === undefined ? 5 : parseFloat(parsed[2]);
+  if (!Number.isFinite(score) || denominator !== 5 || score < 0 || score > 5) return 'N/A';
```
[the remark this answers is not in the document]
vliggio: Not adopted, and the reason is repo evidence rather than preference: `validateEvaluationShape` in the same file (gemini-eval.mjs:259) accepts `0..5`, so rejecting 0 here would mean a report its own validator passed lands as `N/A` in the tracker. Two guards in one file disagreeing about what a valid score is, is the exact bug class this PR chain exists to close (the prefix-vs-whole-line split at :257/:446). `looksLikeScoreCell('0/5')` is also true, so the cell is legible today. You're right that modes/_shared.md:77 documents 1-5; aligning the validator and the normalizer on that is a change to what counts as a valid report, and I'd rather it be its own PR than a side effect of a score-cell fix. Happy to open one if a maintainer wants it.

tests/evaluator-score-cell.test.mjs
```diff
@@ -0,0 +1,86 @@
+// tests/evaluator-score-cell.test.mjs — every evaluator's normalizedTrackerScore
```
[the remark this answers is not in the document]
vliggio: Not adopted. CONTRIBUTING.md:189 scopes `{module}.test.mjs` to `web/` suites mirroring `web/src/` paths; root `tests/` is behaviour-named in practice — 39 of 199 suites match a root script name, and siblings like `add-entry-dedup.test.mjs`, `analyze-patterns-outcomes.test.mjs` and `calibrate-tracker-colmap.test.mjs` follow the same style as this one. More to the point, the suite's subject is agreement BETWEEN copies discovered by definition; a per-module split cannot express that and would duplicate the case table in every file.

tests/evaluator-score-cell.test.mjs
```diff
@@ -0,0 +1,87 @@
...
+// `4.2 (final)`, `4.2 (internal)` and `4.5 - strong signal` all contain the bare
+// substring `na`, which an unanchored /n\/?a/i guard turned into `N/A` — a
+// completed evaluation recorded as unscored. `8/10` and `4.2/10` are the mirror
+// image: a wrong denominator reinterpreted as `8/5` merges as a genuine score.
+const cases = [
+  ['4.2',                 '4.2/5'],
+  ['4.2/5',               '4.2/5'],
+  ['4.2 (strong fit)',    '4.2/5'],
+  ['4.2 (strong fit)/5',  '4.2/5'],
+  ['4.2 (strong fit)/10', 'N/A'],
```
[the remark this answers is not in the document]
vliggio: Added in bd51326, plus the bare `4.2 (fit 3/4 axes)` alongside it. Worth pinning: the refusal was documented in the helper's comment but untested, which is how a documented trade-off quietly turns into an accidental one. Confirmed both return `N/A` — the later `/5` does not mask the earlier `3/4`.

Scott-Emberson (commented):
Test-axis only. The substance is in `gemini-eval.mjs` and `batch-evaluate-gemini.mjs`, which is the maintainer's routing call; I am reviewing the owned test `tests/evaluator-score-cell.test.mjs`.

It is sound. It lifts each evaluator's actual `normalizedTrackerScore` out of source and runs it against hardcoded literal expectations, so it is not a tautology (the expected cells are independent constants). It pins the parsed numeric form and rejects the concatenated form that was the bug: `'4.2 (strong fit)'` to `'4.2/5'` (the old concatenation produced the malformed `'4.2 (strong fit)/5'`), `'8/10'` to `N/A`, `'80%'` to `N/A`, `'4.2/10'` to `N/A`, so each reddens under a regression back to string concatenation. It is double-guarded: every returned cell must also pass `looksLikeScoreCell`, tying the output to what `tracker-parse.mjs` actually accepts, and wrong-denominator or out-of-range inputs are pinned to `N/A` so a wrong number is never laundered into a genuine `X.X/5` score.

PR #3800, opened 2026-09-04
fix: add MiniMax-M3 and MiniMax-M2.7 token rates

octo-patch:
Reason: Add current pricing for the two supported MiniMax evaluator model IDs so cost estimates no longer use a fallback rate.

Adds input, cached-input, and output rates for MiniMax-M3 and MiniMax-M2.7 in the shared token tracker.

Adds regression coverage that calculates a mixed cached and uncached request for both models.

Checks:
- `node tests/token-tracker.test.mjs`
- `npm run lint -- --quiet`
- Target-rate validation against the configured model facts
- `node test-all.mjs --quick` (8,092 checks passed; one unrelated integration check failed because `cv.md` was absent and its external API request was unavailable)

## Summary

Adds accurate token cost estimates for `MiniMax-M3` and `MiniMax-M2.7`.

## User impact

Users receive separate rates for uncached input, cached input, and output tokens. Mixed cached and uncached requests no longer use fallback rates.

## Changes

- `utils/token-tracker.mjs:5-8`: Adds both MiniMax models to `RATES`.
- `tests/token-tracker.test.mjs:83-90`: Adds regression coverage for mixed cached and uncached usage.

### DIFF
--- a/utils/token-tracker.mjs
+++ b/utils/token-tracker.mjs
@@ -3,6 +3,10 @@
  */
 
 export const RATES = {
+  // MiniMax models
+  'MiniMax-M3': { input: 0.60 / 1000000, output: 2.40 / 1000000, cachedInput: 0.12 / 1000000 },
+  'MiniMax-M2.7': { input: 0.30 / 1000000, output: 1.20 / 1000000, cachedInput: 0.06 / 1000000 },
+
   // OpenAI models
   'gpt-4o-mini': { input: 0.150 / 1000000, output: 0.600 / 1000000 },
   'gpt-4o': { input: 2.50 / 1000000, output: 10.00 / 1000000 },
--- a/tests/token-tracker.test.mjs
+++ b/tests/token-tracker.test.mjs
@@ -79,6 +79,21 @@ try {
   } else {
     fail(`formatBreakdown zero-token step rendering failed:\n${breakdownZero}`);
   }
+
+  // 8. MiniMax model rates include uncached input, cached input, and output
+  const minimaxUsage = { prompt_tokens: 1000000, completion_tokens: 1000000, cached_tokens: 500000 };
+  const minimaxCases = [
+    ['MiniMax-M3', 2.76],
+    ['MiniMax-M2.7', 1.38],
+  ];
+  for (const [model, expected] of minimaxCases) {
+    const actual = estimateCost(model, minimaxUsage, 'minimax');
+    if (actual !== null && Math.abs(actual - expected) < 1e-9) {
+      pass(`estimateCost for ${model} includes its cached-input rate`);
+    } else {
+      fail(`estimateCost for ${model} failed: expected ${expected}, got ${actual}`);
+    }
+  }
 } catch (e) {
   fail(`token-tracker tests crashed: ${e.message}`);
 }


Scott-Emberson (commented):
Test-axis only. The substance here is in `utils/token-tracker.mjs`, which is the maintainer's routing call; I am reviewing the owned test.

`tests/token-tracker.test.mjs` is sound. It pins hardcoded expected totals for both new models (`MiniMax-M3` to 2.76, `MiniMax-M2.7` to 1.38) as literals independent of the RATES table, so it is not a tautology, and the usage fixture (1M prompt, 1M completion, 500k cached) exercises all three rate fields through `estimateCost`. A wrong input, output, or cachedInput rate reddens (dropping cachedInput falls back to `input*0.5` giving 2.775, which fails the 1e-9 tolerance), and removing either model key makes `estimateCost` return null so the assertion fails. It reddens on a wrong number or a dropped key.

nikitacometa (commented):
Arithmetic checks out (M3: 0.30 + 0.06 + 2.40 = 2.76 for the fixture) and the test reddens on a wrong rate or a dropped key (dropping `cachedInput` falls back to `input * 0.5` = 2.775, outside the tolerance). Two things before this is mergeable.

Sourcing: utils/token-tracker.mjs:6 is `// MiniMax models`, while the Gemini block at 14 names the tier and the pricing basis. These constants are checkable only against a vendor page and go stale silently, and `formatBreakdown` prints the result as a dollar figure. Add the price page URL and an as-of date, and answer the open inline comment about 512k tiering: if MiniMax bills tiered, one flat rate misprices every request past the threshold and the pinned 2.76 is wrong.

Reachability: nothing in the tree names `MiniMax-M3` or `MiniMax-M2.7` (`grep -rni minimax` finds prose at openai-eval.mjs:6 and the OpenRouter slug prefix at openrouter-runner.mjs:74). `estimateCost` matches `RATES[model]` and then `model.includes(k)` case-sensitively (61-64), so a lowercase OpenRouter slug returns null and the user sees `est. cost n/a`. Say which invocation produces these exact IDs, or add lowercase aliases.

utils/token-tracker.mjs
```diff
@@ -3,6 +3,10 @@
  */
 
 export const RATES = {
+  // MiniMax models
```
nikitacometa, line 6: No source named; the Gemini block below cites its tier and basis. Add the price page URL and an as-of date.

utils/token-tracker.mjs
```diff
@@ -3,6 +3,10 @@
...
+  'MiniMax-M3': { input: 0.60 / 1000000, output: 2.40 / 1000000, cachedInput: 0.12 / 1000000 },
```
nikitacometa, line 7: Flat-rate, or tiered at 512k input as the open comment says? If tiered, one rate misprices every request past the threshold and the pinned 2.76 is wrong.

utils/token-tracker.mjs
```diff
@@ -3,6 +3,10 @@
...
+  'MiniMax-M2.7': { input: 0.30 / 1000000, output: 1.20 / 1000000, cachedInput: 0.06 / 1000000 },
```
nikitacometa, line 8: Which call path emits these exact IDs? Matching is case-sensitive (61-64), so an OpenRouter `minimax/...` slug returns null.

PR #3802, opened 2026-09-04
test(yaml): fail CI on an unparseable tracked YAML file

ilyakanevskiy:
## What

Adds `tests/yaml-syntax.test.mjs`: enumerate every tracked `.yml`/`.yaml` file with `git ls-files` and parse each one with the `js-yaml` the repo already depends on. A file that does not parse fails the check.

## Why

Nothing validated YAML. `npm run lint` runs `node --check` on `.mjs` files only, and no CI job touched YAML at all. A GitHub Issue Form or workflow that is invalid YAML does not fail a build — the form just silently stops rendering on github.com. That happened in review of #3754: an unquoted `description:` value containing `User-agent: *` opened a phantom nested mapping and made `.github/ISSUE_TEMPLATE/source-proposal.yml` unloadable.

## Scope

Parse-only, not schema validation. "Does it parse" catches the whole class of failure that actually reaches `main` (a value that silently restructures the document) without coupling the test to GitHub's issue-form schema. Non-`.github/` files are parsed with `yaml.load` — the call every runtime consumer uses (`scan.mjs`: `const parseYaml = yaml.load`) — so a multi-document file that would throw in the runtime fails here too. `.github/` YAML is parsed with `loadAll`, since GitHub's own issue-form / workflow parsers read those, not this repo's code.

Files are discovered, never listed — a new YAML file is covered the moment it is tracked, and the extension is matched case-insensitively so a `.YAML` / `.YML` file is not missed. The check is auto-discovered by `test-all.mjs` (`tests/**/*.test.mjs`), so it runs on every PR with no wiring. Fail-closed: an empty sweep, an unreadable file, a non-regular entry, or a discovery error is a failure, not a skip, and a guard-the-guard assertion proves the parser still rejects the exact shape from #3754, accepts the quoted form, and that `load()` rejects a multi-document stream.

Currently green over all 50 tracked YAML files.

Adds `tests/yaml-syntax.test.mjs` to validate every tracked `.yml` and `.yaml` file.

Users now receive test failures for invalid, empty, unreadable, symlinked, or non-regular YAML entries. The test supports multi-document YAML with `js-yaml.loadAll` and checks the quoted and unquoted scalar failure pattern: `tests/yaml-syntax.test.mjs:51-73`.

The test runs through existing `test-all.mjs` discovery: `tests/yaml-syntax.test.mjs:81-113`.

System-file scope: `.github/` YAML files are checked. No changes were made to `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, or `providers/`.

### DIFF
--- a/tests/yaml-syntax.test.mjs
+++ b/tests/yaml-syntax.test.mjs
@@ -1,16 +1,20 @@
 // tests/yaml-syntax.test.mjs — every tracked YAML file must parse.
 //
-// Enumerates *.yml/*.yaml via `git ls-files` and parses each one with the
-// repo's js-yaml. A file that does not parse fails the check. This is the only
-// YAML validation in the suite — `npm run lint` runs `node --check` on .mjs
-// files only.
+// Matches *.yml/*.yaml case-insensitively over `git ls-files` and parses each
+// one with the repo's js-yaml. A file that does not parse fails the check. This
+// is the only YAML validation in the suite — `npm run lint` runs `node --check`
+// on .mjs files only.
 //
 // Parse-only, not schema validation: it catches a value that silently
 // restructures the document (an unquoted scalar with an embedded `: `, a bad
 // indent) without coupling the test to GitHub's issue-form or workflow schemas.
 //
-// loadAll, not load: load() throws on a legitimately multi-document file
-// ("expected a single document"), which must not read as a syntax error.
+// Parsed with yaml.load — the same call every non-.github/ consumer uses
+// (scan.mjs: `const parseYaml = yaml.load`), so a multi-document file that would
+// throw in the runtime ("expected a single document") fails here too. .github/
+// YAML is read by GitHub's own issue-form / workflow parsers, not by this repo's
+// code, so it is parsed with loadAll instead — a multi-document stream there is
+// not a syntax error this gate should invent.
 //
 // Files are discovered, not listed, so a new YAML file is covered as soon as it
 // is tracked. `git ls-files` is the set that ships, needs no skip-list, and
@@ -33,14 +37,20 @@ console.log('\nevery tracked YAML file must parse');
 function yamlFiles() {
   try {
     // -z: NUL-separated, so a path containing a newline or quote cannot split a
-    // record and silently drop a file from the sweep.
-    const out = execFileSync('git', ['-C', ROOT, 'ls-files', '-z', '--', '*.yml', '*.yaml'], {
+    // record and silently drop a file from the sweep. No pathspec: list every
+    // tracked path and match the extension below, so this one filter is the
+    // single source of truth for what counts as YAML — a `.YAML`/`.YML` file
+    // (git pathspecs are case-sensitive) is still swept.
+    const out = execFileSync('git', ['-C', ROOT, 'ls-files', '-z'], {
       encoding: 'utf-8',
       maxBuffer: 32 * 1024 * 1024,
     });
     const files = out
       .split('\0')
-      .filter((p) => p && (p.endsWith('.yml') || p.endsWith('.yaml')))
+      .filter((p) => {
+        const lower = p.toLowerCase();
+        return lower.endsWith('.yml') || lower.endsWith('.yaml');
+      })
       .map((p) => join(ROOT, p));
     return { files, error: null };
   } catch (err) {
@@ -50,10 +60,12 @@ function yamlFiles() {
 
 // Guard the guard: a parse step that silently swallowed errors, or a `load` that
 // had been swapped for something inert, would report a clean sweep forever.
-// Prove it rejects an unquoted value with an embedded `: ` and accepts the
-// quoted form.
+// Prove it rejects an unquoted value with an embedded `: `, accepts the quoted
+// form, and that yaml.load (the non-.github/ path) rejects a multi-document
+// stream that loadAll would wave through.
 const BROKEN_FIXTURE = "attributes:\n  description: A generic client (`User-agent: *`)?\n";
 const FIXED_FIXTURE = 'attributes:\n  description: "A generic client (`User-agent: *`)?"\n';
+const MULTIDOC_FIXTURE = 'a: 1\n---\nb: 2\n';
 let detectorRejects = false;
 try {
   yaml.loadAll(BROKEN_FIXTURE);
@@ -66,10 +78,18 @@ try {
 } catch {
   detectorAccepts = false;
 }
-if (detectorRejects && detectorAccepts) {
-  pass('parser rejects an unquoted value with an embedded `: ` and accepts the quoted form');
+let loadRejectsMultiDoc = false;
+try {
+  yaml.load(MULTIDOC_FIXTURE);
+} catch {
+  loadRejectsMultiDoc = true;
+}
+if (detectorRejects && detectorAccepts && loadRejectsMultiDoc) {
+  pass('parser rejects an unquoted embedded `: `, accepts the quoted form, and load() rejects a multi-document stream');
 } else {
-  fail(`detector broken: rejects=${detectorRejects} accepts=${detectorAccepts} — it would pass the sweep regardless of the files`);
+  fail(
+    `detector broken: rejects=${detectorRejects} accepts=${detectorAccepts} loadRejectsMultiDoc=${loadRejectsMultiDoc} — it would pass the sweep regardless of the files`,
+  );
 }
 
 const offenders = [];
@@ -80,6 +100,7 @@ const unreadable = [];
 
 const { files, error: discoveryError } = yamlFiles();
 for (const file of files) {
+  const rel = relative(ROOT, file);
   let text;
   try {
     // lstat before read, and reject anything that is not a regular file.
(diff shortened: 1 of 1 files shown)


tests/yaml-syntax.test.mjs
```diff
@@ -0,0 +1,107 @@
...
+// A file that cannot be read is not a file that passes — report it as its own
+// failure rather than skip it, so the sweep cannot go green while covering less
+// of the tree than it claims.
+const unreadable = [];
+
+const { files, error: discoveryError } = yamlFiles();
+for (const file of files) {
+  let text;
+  try {
+    text = readFileSync(file, 'utf-8');
```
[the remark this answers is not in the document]
ilyakanevskiy: All 50 tracked `.yml`/`.yaml` are mode `100644`, and `git ls-files` only ever returns blobs, never directories — so the only non-regular case reachable here is a symlink, and there are none matching the glob today. If one were added, `readFileSync` already covers it: a symlink to a readable YAML file gets validated, and a dangling or non-file target throws, is caught, and lands in `unreadable` → `fail()` with the path named. That fail-closed reject is the behaviour a gate wants — a tracked YAML path that can't be validated should turn CI red, not be silently skipped, which is what an `lstatSync(...).isFile()` pre-check would do here. The sibling sweeps (`js-yaml-import-form.test.mjs`, `js-yaml-version-floor.test.mjs`) treat unreadable-on-read as a failure the same way, with no stat guard. Leaving as is.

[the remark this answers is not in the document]
ilyakanevskiy: Reversed — you're right that `/dev/zero` and a FIFO never reach the catch: `readFileSync` reads the char device unboundedly and blocks on the FIFO, so my earlier "non-regular target throws" only held for dangling symlinks and directories. Applied in 3200ec0: `lstatSync(file).isFile()` before the read, and a non-regular entry goes to `unreadable` (a hard failure), not skipped. The two sibling sweeps carry the same gap; folding the `git ls-files` enumeration + guard + read into one shared helper is a separate follow-up so this PR stays scoped to the YAML gate.

Scott-Emberson (commented):
Verified the owned test `tests/yaml-syntax.test.mjs`. It is sound and the guard has teeth:

- Enumeration is `git ls-files` over `*.yml`/`*.yaml` under ROOT (NUL-split, extension-filtered), so it is the real tracked set with no hardcoded list, and a newly-tracked YAML is covered the moment it ships. An empty result fails rather than silently passing, and a git failure fails via `discoveryError`.
- The failing assertion parses each file with `yaml.loadAll` in try/catch, collects offenders, and fails if any exist, so a committed broken YAML reddens (the catch records the failure, it does not swallow it into a pass). Using `loadAll` avoids false-failing legitimate multi-doc files.
- The detector is itself guarded: a live self-test parses a known-broken and a known-good fixture and asserts it rejects the first and accepts the second, so if js-yaml were swapped inert the sweep would go green over everything and this self-test fails first.

Owned test verified sound. The branch is behind main, so I am holding formal approval until it is rebased and green, then I will re-confirm at the final head and surface it as ready-to-approve.

tests/yaml-syntax.test.mjs
```diff
@@ -0,0 +1,116 @@
...
+  try {
+    // -z: NUL-separated, so a path containing a newline or quote cannot split a
+    // record and silently drop a file from the sweep.
+    const out = execFileSync('git', ['-C', ROOT, 'ls-files', '-z', '--', '*.yml', '*.yaml'], {
+      encoding: 'utf-8',
+      maxBuffer: 32 * 1024 * 1024,
+    });
+    const files = out
+      .split('\0')
+      .filter((p) => p && (p.endsWith('.yml') || p.endsWith('.yaml')))
```
nikitacometa: Case-sensitive, as is the pathspec at 37. With a broken `broken.YAML` staged the sweep stayed green at 50 files. `p.toLowerCase().endsWith(…)` plus an unrestricted `ls-files` closes it and makes this line the single source of truth for the extension set.

tests/yaml-syntax.test.mjs
```diff
@@ -0,0 +1,116 @@
...
+      unreadable.push(`${relative(ROOT, file)} (not a regular file)`);
+      continue;
+    }
+    text = readFileSync(file, 'utf-8');
+  } catch (err) {
+    unreadable.push(`${relative(ROOT, file)} (${err.code || err.message})`);
+    continue;
+  }
+  try {
+    yaml.loadAll(text);
```
nikitacometa: `loadAll` accepts what the runtime rejects. scan.mjs:87 is `const parseYaml = yaml.load`; a multi-document `templates/states.yml` passes here and throws there. Assert one document outside `.github/`.

nikitacometa (approved):
Verified at 3200ec00. Enumeration is tracked-only, so a gitignored `portals.yml` is never read (`git ls-files portals.yml` is empty, .gitignore:65). The gate has teeth: appending the #3754 shape to `.github/ISSUE_TEMPLATE/source-proposal.yml` gives 1 passed, 1 failed naming file and position; a removed tracked file reports ENOENT and fails the sweep. The lstat thread is resolved at 90, and `execFileSync('git', …)` matches the three sibling suites that already do it.

Two gaps, neither blocking. `loadAll` at 100 is more permissive than every runtime consumer, which uses `load` (scan.mjs:87 `const parseYaml = yaml.load`): appending `---\nextra: doc\n` to `templates/states.yml` leaves the suite green while `yaml.load` on that file throws "expected a single document in the stream". And a tracked `broken.YAML` is invisible: 37 and 43 are lowercase-only, so staged it passed unseen (50 files, 2 passed).

tests/yaml-syntax.test.mjs
ilyakanevskiy: Fixed in 9d2fc22. Dropped the `*.yml`/`*.yaml` pathspec — `git ls-files -z` is unrestricted now, and the JS filter lowercases before matching, so it's the single source of truth for the extension set.

Verified: staging an unparseable `brokenupper.YAML` now goes red and names the file (it was invisible before).

tests/yaml-syntax.test.mjs
ilyakanevskiy: Fixed in 9d2fc22 — worth flagging this reverses a choice the PR made deliberately, not just closes a gap: the header comment and the PR body both argued for `loadAll` specifically to tolerate 'a legitimately multi-document file arriving later.'

You're right that argument was speculative — no runtime consumer uses `loadAll` (every one, including scan.mjs's `parseYaml = yaml.load`, is single-document-only), and there are zero multi-doc files in the tree today. Non-`.github/` files now parse with `yaml.load`; `.github/` stays on `loadAll` since GitHub's own parsers read those, not this repo's code.

Verified: a multi-doc `templates/states.yml` now reddens with 'expected a single document in the stream', the same shape under `.github/` stays green. Added a `load()`-rejects-multi-doc case to the guard-the-guard self-test.

tests/yaml-syntax.test.mjs
```diff
@@ -0,0 +1,142 @@
...
+if (discoveryError !== null) {
+  fail(`could not list tracked YAML files, so the sweep ran against nothing: ${discoveryError}`);
+} else if (files.length === 0) {
+  fail('git ls-files produced no *.yml/*.yaml — the YAML syntax sweep scanned nothing');
+} else if (unreadable.length > 0) {
+  fail(`could not read ${unreadable.length} tracked YAML file(s), so the sweep is incomplete: ${unreadable.join(', ')}`);
+} else if (offenders.length === 0) {
+  pass(`all ${files.length} tracked YAML files parse`);
+} else {
+  fail(`YAML file(s) that do not parse — a GitHub Issue Form or workflow in this set will not load: ${offenders.join(' · ')}`);
```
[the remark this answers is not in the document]
ilyakanevskiy: Applied in b7120dc — dropped the Issue-Form/workflow-specific wording as suggested.

Scott-Emberson (commented):
Re-verified at the rebased head `b7120dc3` (now up to date with main, CI green). The owned test changed since my prior look and the edit only strengthens it:

- enumeration is still `git ls-files -z` over `*.yml`/`*.yaml` with a case-insensitive extension filter (no hardcoded list, NUL-separated so odd paths cannot drop out), and both an empty sweep and a discovery error hard-fail;
- a committed broken YAML still reddens (the per-file catch records an offender and `offenders.length > 0` fails), and unreadable or non-regular entries hard-fail rather than being skipped;
- the guard-the-guard self-test still bites, now requiring the detector to reject a broken embedded doc, accept a quoted-good one, and reject a multi-doc stream, so the parser cannot be swapped inert;
- the four new commits add an lstat symlink/FIFO hang-guard (a new hard-fail path), case-insensitive `.YML`/`.YAML` coverage, and parsing non-`.github/` files via `load()` (which rejects the multi-doc streams `loadAll` would wave through), so strictly more errors are caught.

This is a GATE-tier PR (the owned test is the whole substance), it is verified sound, CI is green, and it is mergeable and no longer behind. I am holding formal approval until Scott gives the go-ahead, and surfacing it to him as ready-to-approve.

PR #3803, opened 2026-09-04
fix(pdf): recognize Chinese section titles in SECTION_ALIASES (#3658)

### LINKED ISSUE #3658: generate-pdf.mjs: SECTION_ALIASES has English + Polish entries only — cv.sections is a silent no-op for Chinese CVs
## Summary

`cv.sections` in `config/profile.yml` reorders CV sections by mapping rendered section titles to canonical keys through `SECTION_ALIASES` (`generate-pdf.mjs:241`). The table carries English and Polish vocabularies only — the Polish entries were added to match `modes/pl`'s documented vocabulary. The repo also ships `modes/zh-TW/` and `modes/zh/` as supported markets, but there are no Chinese entries, so for a Chinese CV every name in `cv.sections` matches nothing and the setting is a no-op.

The failure is detected but not solved: the warning at `generate-pdf.mjs:862` correctly reports "matched no section" and even points at adding entries to `SECTION_ALIASES` — which a user can't do durably, since `update-system.mjs` restores core scripts on update.

Meanwhile the section-order guard compares rendered titles against `cv.md` titles literally, which works fine in Chinese. So on a zh CV whose `cv.md` order differs from the template's, the guard blocks the render and `cv.sections` — the documented way to resolve exactly that conflict — can't take effect. The two mechanisms' i18n coverage is asymmetric.

## Repro

With a zh-TW CV (`cv.md` sections `技能` before `學歷`, template `cv-template.zh-minimal.html`) and:

```yaml
cv:
  sections: [skills, education, certifications, awards]
```

`node generate-pdf.mjs …` prints:

```
⚠️  config/profile.yml cv.sections names "skills", "education", "certifications", "awards", which matched no section in this CV. The CV also renders 6 section title(s) the section-order alias table doesn't recognize (專業摘要, 核心能力, 工作經歷, 專案, 學歷, 技能) …
```

then the order guard fails the run (`rendered 工作經歷 → 專案 → 學歷 → 技能; cv.md 工作經歷 → 專案 → 技能 → 學歷`).

## Suggested fix

Follow the Polish precedent — add the zh vocabulary the shipped market modes already imply:

```js
// Chinese (Traditional/Simplified) — matching modes/zh-TW and modes/zh
['專業摘要', 'summary'], ['摘要', 'summary'],
['核心能力', 'competencies'],
['工作經歷', 'experience'], ['工作经历', 'experience'],
['專案', 'projects'], ['项目', 'projects'], ['專案成就', 'projects'],
['學歷', 'education'], ['学历', 'education'],
['證照', 'certifications'], ['資格認證', 'certifications'],
['技能', 'skills'], ['專長', 'skills'], ['语言能力', 'skills'],
```

(The `语言能力` → `skills` mapping reflects how our CV folds languages into the skills section; upstream may prefer a dedicated key — maintainer's call.)

Longer-term, the same market-aware idea as #3601 applies: let `language.modes_dir` contribute per-market alias vocabulary so each shipped mode brings its own list instead of growing one central table.

Workarounds meanwhile: reorder `cv.md` to match the template (what we did), or pass `--allow-reorder`.

Related: #3601 (same i18n gap class, in `jd-skill-gap.mjs`'s requirement-header vocabulary).

## Environment

career-ops v1.31.0 # x-release-please-version @ 3a067ee, Node v22, macOS

maxmilian:
Closes #3658.

## The problem

`SECTION_ALIASES` in `generate-pdf.mjs` held English and Polish spellings only. This repo ships modes for two Chinese markets — `modes/zh-TW` and `modes/zh`, rendered through `templates/cv-template.zh-minimal.html` — and every title those CVs render fell through `sectionKey()`'s "return the normalized title" fallback, matching no canonical key.

The two mechanisms that depend on the table then broke **asymmetrically**:

- `validateCvSectionOrder()` compares titles verbatim, so it *does* work for Chinese: a Chinese CV whose order diverges from `cv.md` is rejected.
- `reorderCvSections()` — `cv.sections` in `config/profile.yml`, the documented way to declare that order on purpose — resolved fewer than two blocks and silently returned the document untouched.

So a Chinese CV could be blocked by the guard with no working way to fix it. `generate-pdf.mjs` already prints the right diagnosis ("its title needs an entry in `SECTION_ALIASES`"), but that advice is not actionable for users: `update-system.mjs` restores the core scripts on update.

The sharpest illustration is in this repo rather than in a user's CV. `tests/zh-minimal-template.test.mjs` renders a Simplified fixture whose `sections` override is `个人简介 / 核心能力 / 工作经历 / 精选项目 / 教育经历 / 认证 / 技术栈` — **before this change, none of those seven titles resolved to a canonical key.**

## The change

56 Traditional/Simplified entries covering all nine canonical keys, added after the Polish block in the same style, with a comment stating where the vocabulary comes from.

The vocabulary is the repo's own rather than invented:

- the `sections` payload in `tests/zh-minimal-template.test.mjs` (the seven titles above), and
- the modes' own wording (e.g. 專業摘要 in `modes/zh-TW/oferta.md`),

plus the everyday synonyms of each. Section titles here come from a user-supplied `sections` override rather than from `DEFAULT_SECTION_TITLES`, so they have no single canonical spelling — which is exactly why an alias table is the right place for them.

Both scripts are listed against every key because a CV written in either renders through this one table. Note these are genuine cross-strait pairs, not character conversion: Traditional CVs say 專案 where Simplified ones say 项目, and 專業證照 pairs with 专业证书.

`foldDiacritics()` is a no-op on Han characters — NFD leaves them undecomposed and there are no combining marks to strip — so the entries survive the fold unchanged. The tests assert that rather than assume it.

No User Layer file is touched: only `generate-pdf.mjs` and tests.

## Tests

Following the shape the Polish entries already use:

- `test-all.mjs` §7e — alias presence in both scripts, `sectionKey()` spellings, and the guard rejecting a Traditional CV with Education hoisted above 工作經歷 while accepting a Simplified CV in `cv.md` order.
- `tests/cv-section-order.test.mjs` — end-to-end for both scripts against the same English `cv.md` (the source of truth in every mode): a scrambled Chinese CV is rejected by the guard, `cv.sections` then permutes it (the document is no longer byte-identical — the observable symptom of the no-op disappearing — every section body survives exactly once, and the "matched no section" warning is gone), and the guard accepts the result. Plus every title the shipped zh-minimal fixture renders, and an explicit assertion that the NFD fold does not disturb Han characters.

`node test-all.mjs --quick`: 8108 passed, 0 failed. The 14 warnings are the pre-existing ones on `main`.

## Summary

Chinese CVs can now use `cv.sections` to reorder sections.

`generate-pdf.mjs:241` adds 56 Traditional and Simplified Chinese aliases for all nine canonical section keys. The aliases match headings used by the `zh-TW` and `zh` modes.

This fixes the previous silent no-op when Chinese section names matched fewer than two aliases. Section-order validation continues to detect incorrect order and now works with configured reordering.

## Tests

Coverage now includes Chinese alias presence, `sectionKey` resolution, section reordering, order validation, fixture titles, warning behavior, body preservation, and Han-character preservation during diacritic folding.

The quick test suite passes with 8,108 tests.

## Files changed

: `generate-pdf.mjs`
: `test-all.mjs`
: `tests/cv-section-order.test.mjs`

### DIFF
--- a/generate-pdf.mjs
+++ b/generate-pdf.mjs
@@ -231,10 +231,11 @@ function foldDiacritics(text) {
 /**
  * Heading spelling -> canonical section key.
  *
- * Polish (modes/pl) is here because without these aliases the rendered Polish
- * titles match nothing derived from the English cv.md: validateCvSectionOrder()
- * finds fewer than two comparable sections and silently returns, leaving the
- * section-order guard disabled on every CV rendered in that mode.
+ * Polish (modes/pl) and Chinese (modes/zh-TW, modes/zh) are here because without
+ * these aliases the rendered non-English titles match nothing derived from the
+ * English cv.md: validateCvSectionOrder() finds fewer than two comparable
+ * sections and silently returns, leaving the section-order guard disabled on
+ * every CV rendered in those modes, and cv.sections resolves no block at all.
  *
  * Keys are folded on construction so authored diacritics match stripped input.
  */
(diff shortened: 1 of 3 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is in `generate-pdf.mjs`, which is a red-line file and the maintainer's call; I am reviewing the owned test `tests/cv-section-order.test.mjs`.

It is sound. Ordering is pinned concretely: `reorderCvSections` on a scrambled Traditional-Chinese CV must yield exactly `['專業摘要','工作經歷','專案','技能','學歷','證照']`, so a wrong order or a no-op reorder fails, and any `matched no section` warning fails too (same for Simplified with its own vocabulary). Mapping is pinned per-title rather than by substring: roughly thirty `sectionKey(title) === canonical` pairs against `SECTION_ALIASES`, which is an exact Map lookup, so a substring cannot pass and a fallthrough returns the Chinese string instead of the canonical key and fails. The asserted titles cross-check against alias entries actually added in the diff, and the end-to-end guard has teeth: the scrambled CV must be rejected before the reordered one is asserted accepted, so a disabled guard fails rather than silently passing.

Issue #3788, opened 2026-09-04
discover-ats.mjs: bamboohr probe crashes the Node process with a Windows libuv assertion (UV_HANDLE_CLOSING)


Schlaflied:
## Summary

Running `discover-ats.mjs` with `--vendors bamboohr` against even a small company list reliably crashes the Node process on Windows, rather than the target company simply landing in the "unresolved (probe error)" bucket.

## Environment

- Node v25.5.0
- Windows 11
- Reproduced with `DEFAULT_CONCURRENCY = 8` (unchanged from the file's default)

## Reproduction

```bash
node discover-ats.mjs "Rocscience" "MaRS Discovery District" --vendors bamboohr --summary
```

Output:

```
==============================================================================
  ATS Discovery — career-ops
  resolved: 1 | unresolved: 1 | duplicates skipped: 1
==============================================================================

  Company                 Vendor      Jobs   Board
  ------------------------------------------------------------------------------------------
  Rocscience              bamboohr    2      https://rocscience.bamboohr.com

  Unresolved (manual follow-up):
    - MaRS Discovery District: probe error(s) occurred — board status unknown, see errors[] and re-run

Assertion failed: !(handle->flags & UV_HANDLE_CLOSING), file src\win\async.c, line 76
```

Exit code 127. Also reproduced at larger scale: a 62-company BambooHR batch (`--in companies.yml --vendors bamboohr --write`) resolved only 14/62 — every other company came back as `probe error(s) occurred`, and re-running the exact same batch twice produced the identical 14 resolved / 48 failed split both times (not flaky/transient).

## Expected vs actual

**Expected:** an unreachable/failing BambooHR board should land the company in the "unresolved" bucket with a normal error message (same as it does for the vendors that report "no supported ATS board found").

**Actual:** the whole Node process crashes with a libuv-level assertion failure, taking down the rest of the batch with it (any company not yet probed at crash time never gets a result at all).

## My guess, not confirmed

[...]

PR #3804, opened 2026-09-04
fix(discover-ats): a refused redirect is an answer, not a status to re-run

artemtrofymenko:
## What does this PR do?

Closes the reproducible half of #3788: `discover-ats.mjs` reported 48 of one user's 62 BambooHR companies as an unknown board status with advice to re-run, when the vendor had in fact answered on the first request.

**BambooHR does not 404 an unknown tenant.** It redirects:

```
mars-discovery-district.bamboohr.com   302 → https://www.bamboohr.com   text/html
definitely-not-a-real-tenant-xyz123    302 → https://www.bamboohr.com   text/html
rocscience.bamboohr.com                200                              application/json
marsdd.bamboohr.com                    200                              application/json
```

`redirect: 'error'` (mandatory, #1440) turns that 302 into a bare `TypeError: fetch failed` — byte-for-byte the shape of a timeout or a DNS failure. The only thing separating them is `err.cause.message === 'unexpected redirect'`, and `probeVendor` kept `err.message` only:

```json
"errors": [ { "vendor": "bamboohr", "error": "fetch failed" } ]
```

With the cause gone, `isDefinitiveAbsence` (404/410 only) classified it as transient, and the user got *"probe error(s) occurred — board status unknown, see errors[] and re-run"*. The same probe redirects again, forever.

### The repo already disagreed with that verdict

`providers/_http.mjs:197` tests for this exact cause string and returns `false` from `isRetryableError`, under a comment that says why:

> it's deterministic and will never succeed on retry

Two files, one error object, opposite conclusions — and the one the user reads was the wrong one. That is the actual defect here; the BambooHR behaviour is just what exposed it.

## The change

- **`providers/_http.mjs`** — export `isRefusedRedirectError()`. The constant and the test were already there; this gives the verdict a second consumer rather than a second copy, and `isRetryableError` now calls it instead of inlining the same three conditions. (Same reasoning as sharing one walker between `npm run lint` and section 1 in #3411: two independent definitions of one predicate are free to re-diverge.)
- **`discover-ats.mjs`** — `probeVendor` keeps the cause (`"fetch failed (unexpected redirect)"`) and sets `refusedRedirect: true`, alongside the existing `httpStatus` discriminator and for the same reason: keep the signal on the object rather than re-parsing prose downstream (#2883).
- **`discover-ats.mjs`** — a refused redirect becomes *settled* (no re-run advice) but **not absence**, with its own reason:

```
MaRS Discovery District: vendor(s) redirected off-tenant (bamboohr) — the slug is
wrong, or no board exists under it. The same probe will redirect again, so set an
explicit `slug` and re-run.
```

`isDefinitiveAbsence` is left alone. A 302 to a marketing page is not the vendor saying "no board" — `mars-discovery-district` redirects while `marsdd` serves three open jobs, so folding this into absence would be as wrong as calling it transient.

## Verification

**Live, before and after, same company:**

```
before:  probe error(s) occurred — board status unknown, see errors[] and re-run
after:   vendor(s) redirected off-tenant (bamboohr) — … set an explicit `slug` and re-run
```

Then following the new advice:

```
Company                 Vendor      Jobs   Board
MaRS Discovery Distric  bamboohr    3      https://marsdd.bamboohr.com
```

The advice the tool now gives, followed literally, turns an unresolved company into three open jobs.

**Suites:** `--only discover-ats` 149 passed / 0 failed; `--only _http` 16 passed / 0 failed. Full `test-all.mjs` 8069 passed / 5 failed — the same 5 as a pristine checkout of this branch's base (8055 passed / 5 failed), all from this being a fresh worktree without `node_modules` (`linkRepoPackage: js-yaml is not installed`), none from this change.

**Mutation, each with proof it applied:**

| mutation | applied | result |
| --- | --- | --- |
| drop the cause enrichment in `probeVendor` | grep 1 → 0 | ❌ `the dropped cause survives into errors[]` |
| revert the guard to `isDefinitiveAbsence` (pre-fix behaviour) | `every(isSettled)` 1 → 0 | ❌ 4 assertions |
| let a refused redirect count as definitive absence | 2 occurrences | ❌ `error entry is not marked definitive` |
| drop the cause check from the predicate, keep the TypeError shape | 1 occurrence | ❌ 3 assertions, including the pre-existing old-Node fallback |

The last one is the one I care about: a predicate that trusted the `TypeError` shape alone would pass every other assertion, and a DNS failure would start reporting as a wrong slug. `isRefusedRedirectError({ status: 404 })` is asserted `false` next to `isRetryableError({ status: 404 })` being `false` too, so a predicate defined as "not retryable" also fails.

## What this does NOT do

The `UV_HANDLE_CLOSING` assertion in #3788's title is untouched. I could not reproduce it in 27 runs across Node 22.22.0, 25.5.0 (the reporter's exact version) and 25.9.0 on Windows 11, at both their two-company scale and 62 companies — including a batch where 61 probes really were refused redirects rather than DNS misses, which I checked. Measurements and the one repo-side trigger I did find (`process.exit(0)` firing with a live `TCPSocketWrap`, at a cost of 1 ms to remove) are in [my comment on the issue](https://github.com/career-ops-hq/career-ops/issues/3788#issuecomment-5534113765). That half needs the reporter's environment, so it stays a separate change.

## Summary

- Refused redirects now count as settled vendor responses.
- BambooHR and Workday redirect errors preserve their causes and set `refusedRedirect`.
- Discovery reports when a vendor redirects off-tenant.
- Discovery advises users to set an explicit slug.
- DNS failures remain transient and retain re-run guidance.
- Definitive `404` and `410` responses remain separate from refused redirects.

## User impact

Users can identify off-tenant vendor redirects and configure an explicit slug. Refused redirects no longer appear as generic `fetch failed` errors or as definitive slug absence.

## Files changed

- `discover-ats.mjs`: Classifies refused redirects, preserves redirect causes, and produces slug-focused results.
- `providers/_http.mjs`: Adds and exports `isRefusedRedirectError()`. HTTP retry logic reuses this predicate.
- `tests/discover-ats.test.mjs`: Covers redirect, DNS, transient, Workday, and definitive absence cases.
- `tests/providers/_http.test.mjs`: Covers redirect detection and retry behavior.

### DIFF
--- a/discover-ats.mjs
+++ b/discover-ats.mjs
@@ -568,6 +568,7 @@ export async function resolveWorkday(company, coords, ctx) {
   const tried = [];
   let emptyUrl = null;
   let lastError;
+  let lastRefusedRedirect = false;
 
   for (const candidate of candidates) {
     tried.push(candidate.careers_url);
@@ -593,11 +594,22 @@ export async function resolveWorkday(company, coords, ctx) {
       if (!emptyUrl) emptyUrl = candidate.careers_url;
     } catch (err) {
       lastError = err?.message || String(err);
+      // The file's second `.fetch(` site, and workday.mjs passes
+      // redirect:'error' too — so a Workday host that redirects arrives here as
+      // the same bare "fetch failed" probeVendor used to report. Keeping the
+      // cause costs nothing and stops the two error paths from describing the
+      // same failure differently (#3788).
+      if (isRefusedRedirectError(err)) {
+        lastError = `${lastError} (${err.cause.message})`;
+        lastRefusedRedirect = true;
+      } else {
+        lastRefusedRedirect = false;
+      }
     }
   }
   return emptyUrl
     ? { status: 'empty', tried, careers_url: emptyUrl }
-    : { status: 'error', tried, detail: lastError };
+    : { status: 'error', tried, detail: lastError, refusedRedirect: lastRefusedRedirect };
 }
 
 /**
@@ -650,7 +662,15 @@ export async function resolveCompany(company, { vendors = VENDOR_ORDER, ctx, inc
       // Use the host resolveWorkday actually confirmed empty, not always wd1.
       emptyBoards.push({ vendor: 'workday', careers_url: wd.careers_url });
     } else if (wd.detail) {
-      errors.push({ vendor: 'workday', error: wd.detail });
+      /** @type {any} */
+      const wdEntry = { vendor: 'workday', error: wd.detail };
+      // Carried so errors[] describes a Workday refusal the same way it
+      // describes a BambooHR one. It cannot reach the reason ladder — the
+      // refused-redirect branch is guarded by !coords and a Workday probe only
+      // runs WITH coords — but a machine reading errors[] should not have to
+      // know that (#3788).
+      if (wd.refusedRedirect) wdEntry.refusedRedirect = true;
+      errors.push(wdEntry);
     }
   }
 
--- a/discover-ats.mjs
+++ b/discover-ats.mjs
@@ -37,7 +37,7 @@ import { fileURLToPath } from 'url';
 import * as yaml from 'js-yaml';
 import { renameSyncWithRetry } from './tracker-utils.mjs';
 
-import { makeHttpCtx } from './providers/_http.mjs';
+import { makeHttpCtx, isRefusedRedirectError } from './providers/_http.mjs';
 import greenhouse from './providers/greenhouse.mjs';
 import ashby from './providers/ashby.mjs';
 import lever from './providers/lever.mjs';
@@ -522,6 +522,18 @@ export async function probeVendor(company, candidate, ctx) {
     // what lets the caller tell a definitive 404 from a transient 5xx instead of
     // re-parsing the message text (#2883).
     if (Number.isInteger(err?.status)) out.httpStatus = err.status;
+    // A refused redirect has no status at all: it is a bare `TypeError: fetch
+    // failed`, and the vendor's actual answer lives one level down in
+    // err.cause. Recording only err.message discarded that answer, so a
+    // BambooHR tenant that does not exist — answered with `302 →
+    // www.bamboohr.com`, not a 404 — reached the user as the single word
+    // "fetch failed" under advice to re-run it. Same reasoning as the
+    // httpStatus line above: keep the discriminator on the object rather than
+    // re-parsing prose downstream (#3788).
+    if (isRefusedRedirectError(err)) {
+      out.refusedRedirect = true;
+      out.error = `${out.error} (${err.cause.message})`;
+    }
     return out;
   }
 }
@@ -621,6 +633,7 @@ export async function resolveCompany(company, { vendors = VENDOR_ORDER, ctx, inc
       /** @type {any} */
       const entry = { vendor: candidate.vendor, error: result.error };
       if (Number.isInteger(result.httpStatus)) entry.httpStatus = result.httpStatus;
+      if (result.refusedRedirect) entry.refusedRedirect = true;
       if (isDefinitiveAbsence(result)) entry.definitive = true;
       errors.push(entry);
     }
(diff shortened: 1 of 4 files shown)


Scott-Emberson (commented):
Verified the change to the shared provider HTTP layer and its owned tests. The SSRF posture is not weakened:

- The new `isRefusedRedirectError(err)` is behaviorally identical to the previous inline guard (`status === undefined && err instanceof TypeError && err.cause.message === REDIRECT_REFUSAL_CAUSE_MESSAGE`), still classifying a refused redirect as non-retryable. `isRetryableError` just delegates to it, no logic change.
- The redirect policy itself is untouched: `fetchInContext` still passes `redirect` through, the `redirect:'error'` contract and `REDIRECT_REFUSAL_CAUSE_MESSAGE` are unchanged, and nothing follows a redirect that was previously blocked. The change only classifies an already-refused error so the CLI stops advising a blind re-run.
- `tests/providers/_http.test.mjs` pins the classifier to the exact TypeError-plus-cause shape and false for every lookalike (non-TypeError, old-Node `cause===undefined`, a DNS-shaped TypeError, a 404), so a refused redirect stays distinguished from other failures and this reddens if the classification regresses. `tests/discover-ats.test.mjs` pins that a refused redirect is surfaced as still-unknown (not re-run, absence not over-claimed), including the `refused redirect + 503 → still unknown` case.

One honest limitation, pre-existing and not introduced here: the `_http` cases assert classification on synthetic error fixtures, not a live 3xx, so they would not redden if someone flipped the production `redirect` value to `follow`. This PR does not touch that policy, so the guard is exactly as strong as before.

Owned tests verified sound. The branch is behind main, so I am holding formal approval until it is rebased and green, then I will re-confirm at the final head.

providers/_http.mjs
```diff
@@ -180,6 +180,28 @@ export function parseRetryAfterMs(value) {
...
+ * "board status unknown — re-run", and 48 of one user's 62 companies were
+ * BambooHR answering "no such tenant" with a 302 (#3788).
+ *
+ * @param {any} err
+ * @returns {boolean}
+ */
+export function isRefusedRedirectError(err) {
+  return err?.status === undefined
+    && err instanceof TypeError
+    && err?.cause?.message === REDIRECT_REFUSAL_CAUSE_MESSAGE;
```
nikitacometa, line 202: Widening this to `cause?.message !== undefined` reddens `--only _http` only; `--only discover-ats` stays 149/0. One case there driving a `getaddrinfo ENOTFOUND` cause and asserting `/status unknown/` would close it, since that is the direction where a hiccup reaches a user as a wrong slug.

discover-ats.mjs
```diff
@@ -522,6 +522,18 @@ export async function probeVendor(company, candidate, ctx) {
...
     if (Number.isInteger(err?.status)) out.httpStatus = err.status;
+    // A refused redirect has no status at all: it is a bare `TypeError: fetch
+    // failed`, and the vendor's actual answer lives one level down in
+    // err.cause. Recording only err.message discarded that answer, so a
+    // BambooHR tenant that does not exist — answered with `302 →
+    // www.bamboohr.com`, not a 404 — reached the user as the single word
+    // "fetch failed" under advice to re-run it. Same reasoning as the
+    // httpStatus line above: keep the discriminator on the object rather than
+    // re-parsing prose downstream (#3788).
+    if (isRefusedRedirectError(err)) {
```
nikitacometa, line 533: Covers `probeVendor`, the first of the file's two `.fetch(` sites. `resolveWorkday` at 595 keeps only `err?.message`, so a refusal on a Workday host still arrives as bare "fetch failed". workday.mjs:366 passes `redirect:'error'`; a live bogus tenant answered 422, so reachable but not the common case.

discover-ats.mjs
```diff
@@ -649,36 +662,59 @@ export async function resolveCompany(company, { vendors = VENDOR_ORDER, ctx, inc
...
-          : (!candidates.length && unsupported.length)
-            ? `slug "${company.slug || deriveSlug(company.name)}" is not a valid board slug for any probed vendor `
-              + `(${unsupported.join(', ')}) — nothing was probed; fix the \`slug\` field and re-run.`
-            // VENDOR_ORDER covers eleven vendors now, so name none of them here.
-            : 'no supported ATS board found. If this company uses Workday, add a hint — '
-              + 'a full careers URL (workday: https://<tenant>.wd<N>.myworkdayjobs.com/<site>) or '
-              + 'workday: {tenant, site} — and re-run; discover-ats will confirm and add it.';
+      // Settled, but by a redirect rather than a 404: the actionable fix is the
+      // slug, not a retry and not a Workday hint.
+      : (errors.length && !coords && errors.some((e) => e.refusedRedirect))
```
nikitacometa, line 717: This branch sits above `workdayHintProvided && !coords`, so a malformed `workday` hint plus a refused redirect now shows the redirect message. Main showed "status unknown" in the same case, so not a regression, but the order is load-bearing.

nikitacometa (approved):
Verified at 0e1d3fdb. Live on Node v24.19.0, `mars-discovery-district.bamboohr.com` under `redirect:'error'` gives `TypeError: fetch failed` with `cause.message "unexpected redirect"` and no status. Through `resolveCompany` with the real http ctx: main says "board status unknown … re-run" with `error: "fetch failed"`, this head says "vendor(s) redirected off-tenant (bamboohr)" with `refusedRedirect: true`. A 404, a 503 and a timeout-shaped TypeError come out identical on both heads. `--only discover-ats` 149/0, `--only _http` 16/0; reverting discover-ats.mjs to the base gives 143/6, reverting providers/_http.mjs gives 7/1.

Three notes, none blocking. discover-ats.mjs:595 in `resolveWorkday` is the file's other `.fetch(` site and still keeps only `err.message`, so a refusal there lands as a bare "fetch failed"; workday.mjs:366 passes `redirect:'error'`, so the class is reachable. Relaxing _http.mjs:202 to `cause?.message !== undefined` reddens `--only _http` but leaves `--only discover-ats` at 149/0, so nothing there pins that a DNS failure still reads "status unknown". And `redirect:'error'` throws before `Location` is read, so this cannot tell a parking page from a board behind SSO or a canonical same-origin redirect; the message hedges, which is right, but `_http.mjs:52-57` already exposes `err.location` under `redirect:'manual'` (jobvite uses it), so the discrimination is available if wanted.

Scott-Emberson (commented):
Re-verified after the head moved (`0e1d3fdb` to `3af13f33`). The delta is one test-hardening commit. The owned `providers/_http.mjs` and `tests/providers/_http.test.mjs` are byte-identical to what I reviewed (absent from the compare), so the `redirect:'error'/'manual'` policy and the `isRefusedRedirectError` predicate are untouched and my earlier read stands. `tests/discover-ats.test.mjs` gained pure additions (no deletions) that harden classification: a DNS-shaped TypeError (cause `getaddrinfo ENOTFOUND`) must stay status-unknown and must not be marked `refusedRedirect` or blame the slug, and the workday refused-redirect path is pinned with the cause surviving into detail plus the discriminator flag. The recorded mutation shows widening the predicate to accept any cause reddens the `_http` run.

The source change to `discover-ats.mjs` (not our lane) is labeling-only: it keeps `err.cause.message` in detail and sets a `refusedRedirect` flag propagated into `errors[]`. No fetch option changed, no `redirect:'follow'` introduced, and a refused redirect is still a terminal non-retried answer, so no redirect that was previously blocked is now followed.

Owned tests verified sound, SSRF posture unchanged. This is GATE-tier (`providers/_http`), but the branch is behind main, so I am holding formal approval until it is rebased and green, then I will re-confirm at the final head.

PR #3805, opened 2026-09-04
fix(update-system): make apply() verify the target it is about to install

### LINKED ISSUE #3052: update-system apply can resolve FETCH_HEAD to an older auto-followed tag and silently downgrade system files on Windows
### Code of Conduct

### Existing issues

## Summary

`node update-system.mjs check` correctly reported an available upgrade:

- local: `1.26.0`
- remote/main: `1.27.0`

But `node update-system.mjs apply` subsequently resolved `FETCH_HEAD`
to the auto-followed tag `career-ops-v1.25.0` instead of the fetched
`main` branch and partially downgraded the installation.

The command exited `0` and reported:

```text
Update complete: v1.26.0 → v1.26.0

### Steps to reproduce

1. Start from a Career-Ops `1.26.0` installation on Windows.

2. Run:

   `node update-system.mjs check`

   The updater reports that `1.27.0` is available.

3. Run:

   `node update-system.mjs apply`

4. The updater performs `git fetch https://github.com/santifer/career-ops.git main`.

5. After that fetch, `.git/FETCH_HEAD` points to the auto-followed older tag `career-ops-v1.25.0`:

   `ae1a92dd1a4d299e637ce5d96f18e79f743a50ba`

   instead of the actual upstream `main` commit:

   `8668ac1a0d37676a8540be54724333b57240ec6d`

6. The updater then uses that incorrect `FETCH_HEAD` as its checkout source and partially replaces v1.26.0 system files with v1.25.0 content.

7. The command still exits with code `0` and reports:

   `Update complete: v1.26.0 → v1.26.0`

### Expected behavior

`update-system.mjs apply` should resolve the requested upstream `main` branch to an explicit immutable commit SHA and use that SHA throughout the update.

Before modifying files, it should verify that:

1. the resolved target corresponds to the requested upstream branch;
2. the target VERSION is not older than the currently installed VERSION;
3. the target VERSION matches the version reported by `update-system.mjs check`.

After applying, the installed VERSION should also match the expected target VERSION.

If any of these checks fail, the updater should fail closed, avoid applying the update, and return a non-zero exit code instead of reporting success.

### CLI tool

Claude Code

### OS and Node.js version

Windows · Git 2.43.0.windows.1 · Node.js v24.19.0

### Output of npm run doctor

```shell
Not run immediately after the failed update because the update had materially failed and the installation was in a mixed state.

After restoring the exact pre-update 1.26.0 baseline, `node doctor.mjs` reported:

All checks passed (1 warning)

The only warning was:

Playwright MCP tools not detected (active CLI: claude)

This warning was pre-existing and unrelated to the updater failure.
```

maxmilian:
Closes #3052.

Thanks to @florzeta for the original report and to @Solaris-star for the analysis that pinned down where in `apply()` the target goes unverified.

## Scope

To be explicit about what this PR does and does not claim: the auto-followed-tag mechanism from the original report has **not** been reproduced, and this PR does not assert it. What it fixes is a structural property readable straight off `apply()` on `main` today — **`apply()` never verifies the target it installs** — independently of how a wrong target could come to be resolved.

## Two gaps

**1. No version comparison on the install path.** `compareVersions()` (`update-system.mjs:677`) has exactly two call sites, `:1771` and `:1795`, and both live inside `check()` — which only decides whether to *notify*. `apply()` (`:1969`–`:2534`) contains zero references to it. So the install path checked out whatever the fetch produced and printed `Update complete: vX → vY` regardless of which direction the version moved.

**2. The target is a moving pseudo-ref, read a dozen times.** `FETCH_HEAD` is re-resolved on every read, and `apply()` read it 13 times: the bootstrap checkout (`:2039`, `:2040`, `:2054`), the target manifest read (`:2095`), the at-risk scan (`:2112`), the per-path checkout loop (`:2156`, `:2172`, `:2182`), the stale-file prune (`:2201`), and the `.gitignore` reconciliation (`:2244`) — plus two indirect ones through `expandToShippedFiles()` and `missingFromTargetManifest()`, which resolved it themselves. Nothing tied those reads to one tree, so a single `apply` run had no guarantee that the tree it inspected was the tree it wrote.

## Changes

- **`pinRefToCommit()`** resolves the fetched tip **once** to an immutable 40-hex SHA; every later step addresses that SHA. All 13 sites plus the two indirect ones now take the pinned ref.
- The re-exec'd child re-fetches, so it would otherwise pin its own target and install a tree different from the one the parent verified and bootstrapped from. It now inherits the parent's SHA via `CAREER_OPS_UPDATE_TARGET_SHA` — a hint, not an authorization (#2866): read only in the authenticated re-exec, validated with `rev-parse` against the local object store, and falling back to a fresh pin if unusable.
- **`versionAtRef()` + `downgradeRefusal()`** gate the update before a single file is written, reusing the existing `compareVersions()`. An older target is refused with `target VERSION x is older than installed y, refusing`. Fail closed: a target whose `VERSION` cannot be read is refused too, since an unverifiable target is not a verified one. Same-version re-applies stay allowed (the #1998 repair path).

## The escape hatch — a deliberate non-decision

I did **not** add a `--allow-downgrade` flag. `--force` already exists with a different meaning (overwrite *my locally modified system files*), and overloading it with a version-direction meaning would be worse than having no flag. The refusal message instead points at `node update-system.mjs rollback`, which is the existing mechanism for returning to an earlier state — and it restores this installation's own backup branch rather than overwriting it with an older upstream tree.

If maintainers want an explicit downgrade path, an opt-in flag is a small follow-up; I would rather you choose its name and semantics than invent one here.

## Tests

New `tests/updater-apply-target-verification.test.mjs`, same shape as `updater-upgrade-safety.test.mjs` (throwaway git repo through the ctx git seam; no network, no `apply()` call):

- **pin**: after `FETCH_HEAD` moves, a read through the pinned SHA still sees the original tree — with a control assertion showing the same read through `FETCH_HEAD` *does* follow the move, so the pin is demonstrably load-bearing; an unresolvable ref throws.
- **guard**: older / newer / same-version / patch-level / major-level, plus the fail-closed empty-`VERSION` case.
- **`apply()` itself** is not exported, so the two properties that live only in its body are asserted against its source: no unpinned ref reads remain, and the guard runs before the first write.

`updater-migration-tests.mjs` pins both new call sites; its existing `FETCH_HEAD` source assertions (and one in `tests/updater-self-bootstrap-backup.test.mjs`) move to the pinned form.

`node test-all.mjs --quick`: **8109 passed, 0 failed, 14 warnings** (baseline on `main`: 8093 / 0 / 14 — no new warnings).

System Layer only; no user-layer file is touched.

## Summary

`update-system.mjs:2280` now installs one immutable commit for each `apply()` run. It verifies the commit against upstream `main` at `update-system.mjs:2398` and rejects older, malformed, or unverifiable versions before changing files.

Users can reapply the installed version. The updater no longer follows tags during fetches at `update-system.mjs:2359`. Verification failures return a non-zero result and do not report success.

Tests cover pinned references, target identity, version validation, bootstrap, staging, manifest checks, and migration behavior.

System files touched: `update-system.mjs`.

System files not touched: `AGENTS.md`, `modes/`, `DATA_CONTRACT.md`, `providers/`, `.github/`.

### DIFF
--- a/tests/updater-apply-target-verification.test.mjs
+++ b/tests/updater-apply-target-verification.test.mjs
@@ -34,6 +34,7 @@ import { pass, fail } from './helpers.mjs';
 import {
   gitIn, pinRefToCommit, versionAtRef, downgradeRefusal,
   isComparableVersion, targetIdentityRefusal, refusalMessage,
+  authoritativeShaFromRefBody,
 } from '../update-system.mjs';
 
 const REPO_ROOT = join(dirname(fileURLToPath(import.meta.url)), '..');
@@ -308,6 +309,84 @@ console.log('\n🧪 Testing that apply() verifies its update target (#3052)...')
   }
 
   rmSync(dir, { recursive: true, force: true, maxRetries: 10, retryDelay: 100 });
+}
+
+// ── 2c-bis. a ref-API body that parses but says nothing must not be quoted ──
+// The authoritative SHA is a string the far side chose. A body can parse as JSON
+// and still carry no commit — a proxied error page, a truncated response, a
+// captive portal. Before the shape check, `object.sha` went to
+// targetIdentityRefusal() verbatim; git cannot resolve it, throws, and the catch
+// reports "does not descend from upstream main". That is a claim about the
+// TARGET made on the strength of a value that described nothing, and it sends
+// the user to look at a tree that was never actually examined. The truthful
+// outcome is the no-claim one that offline and rate-limited runs already take.
+{
+  const bodies = [
+    ['not-a-sha', 'a plain non-SHA string'],
+    ['<html><body>403 Forbidden</body></html>', 'an HTML error page in the sha field'],
+    ['abc123', 'a short hex string that is not a full object name'],
+    ['0'.repeat(39), 'a 39-hex near-miss'],
+    ['0'.repeat(41), 'a 41-hex near-miss'],
+    ['A'.repeat(40), 'uppercase hex, which is not what git rev-parse emits'],
+    ['  ', 'whitespace'],
+  ];
+
+  const leaked = bodies.filter(
+    ([sha]) => authoritativeShaFromRefBody(JSON.stringify({ object: { sha } })) !== '',
+  );
+  if (leaked.length === 0) {
+    pass(`every ref-API body carrying a non-SHA reads as "no answer" (${bodies.length} shapes)`);
+  } else {
+    fail(`non-SHA object.sha values survived the shape check: ${leaked.map(([s]) => JSON.stringify(s)).join(', ')}`);
+  }
+
+  // The two properties the fix is actually for, asserted end to end through the
+  // consumer: the verdict is the no-claim one, and git is never reached at all —
+  // so the guard is proven to stop the value BEFORE merge-base sees it, rather
+  // than rewording a refusal after the fact.
+  const calls = [];
+  const spy = { git: (...args) => { calls.push(args); throw new Error('git should never have been called'); } };
+  const target = 'a'.repeat(40);
+  const parsedButUseless = authoritativeShaFromRefBody(JSON.stringify({ object: { sha: 'not-a-sha' } }));
+  const verdict = targetIdentityRefusal(target, parsedButUseless, spy);
+
+  if (verdict === null) {
+    pass('a body that parses but carries no commit yields no identity verdict, not a false "does not descend"');
+  } else {
+    fail(`a non-SHA authoritative value produced a verdict about the target: ${JSON.stringify(verdict)}`);
+  }
+  if (calls.length === 0) {
+    pass('merge-base is never invoked for a non-SHA authoritative value');
+  } else {
+    fail(`git was called with an unresolvable value: ${JSON.stringify(calls)}`);
+  }
+
+  // The regression this replaces, stated as the control: feeding the RAW field
+  // (what the code did before the shape check) does reach merge-base and does
+  // produce the untrue sentence. Without this, the two assertions above would
+  // still pass if the whole identity guard were deleted.
+  const rawCalls = [];
+  const rawSpy = { git: (...args) => { rawCalls.push(args); throw new Error('fatal: Not a valid object name'); } };
+  const rawVerdict = targetIdentityRefusal(target, 'not-a-sha', rawSpy);
+  if (rawCalls.length === 1 && /does not descend from upstream main/.test(rawVerdict || '')) {
+    pass('control: the unchecked value would have reached merge-base and produced the untrue refusal');
+  } else {
+    fail(`control failed: the pre-fix path no longer misbehaves (calls=${rawCalls.length}, verdict=${JSON.stringify(rawVerdict)})`);
+  }
+
+  // Filter, not blanket rejection: a real ref-API body still yields its commit,
+  // and a body that does not parse at all keeps its existing '' answer.
+  const real = 'd'.repeat(40);
+  if (authoritativeShaFromRefBody(JSON.stringify({ object: { sha: real, type: 'commit' } })) === real) {
+    pass('control: a genuine ref-API body still yields its commit SHA');
+  } else {
+    fail('the shape check rejects a legitimate ref-API body');
+  }
+  if (authoritativeShaFromRefBody('<html>502</html>') === '' && authoritativeShaFromRefBody('{}') === '') {
+    pass("an unparseable body and a body with no object still read as '' (unchanged)");
+  } else {
+    fail('the malformed-body path regressed');
+  }
 }
 
 // ── 2d. refusalMessage tells the truth about what is on disk ──
--- a/tests/updater-apply-target-verification.test.mjs
+++ b/tests/updater-apply-target-verification.test.mjs
@@ -1,0 +1,412 @@
+/**
+ * updater-apply-target-verification.test.mjs — apply() must verify its target (#3052).
+ *
+ * Two structural gaps in `update-system.mjs apply()`, both readable straight
+ * off the code before this change:
+ *
+ *   1. apply() never compared versions. `compareVersions()` has exactly two
+ *      call sites and both live inside `check()`, which only decides whether to
+ *      NOTIFY. The install path checked out whatever the fetch produced and
+ *      printed "Update complete: vX → vY" regardless of which direction the
+ *      version moved. `downgradeRefusal()` is the pure predicate apply() now
+ *      consults before it writes anything.
+ *
+ *   2. apply() re-read `FETCH_HEAD` more than a dozen times. It is a pseudo-ref,
+ *      re-resolved on every read, so the bootstrap checkout, the manifest read,
+ *      the per-path checkout loop, the stale-file prune, the .gitignore
+ *      reconciliation and the staging expansion were a dozen independent
+ *      questions about a ref free to move between them. `pinRefToCommit()`
+ *      resolves it once to an immutable SHA that the rest of the run addresses.
+ *
+ * Same shape as updater-upgrade-safety.test.mjs: the pure exports are driven
+ * against a throwaway git repo through the ctx git seam — no network, no
+ * apply() call. apply() itself is not exported, so the two properties that only
+ * exist in its body (no unpinned ref reads; the guard runs before the first
+ * write) are asserted against its source text.
+ */
+
+import { execFileSync } from 'child_process';
+import { mkdtempSync, mkdirSync, writeFileSync, readFileSync, rmSync } from 'fs';
+import { tmpdir } from 'os';
+import { join, dirname } from 'path';
+import { fileURLToPath } from 'url';
+import { pass, fail } from './helpers.mjs';
+import {
+  gitIn, pinRefToCommit, versionAtRef, downgradeRefusal,
+  isComparableVersion, targetIdentityRefusal, refusalMessage,
+} from '../update-system.mjs';
+
+const REPO_ROOT = join(dirname(fileURLToPath(import.meta.url)), '..');
+
+function makeRepo(prefix) {
+  const dir = mkdtempSync(join(tmpdir(), prefix));
+  const g = (...args) => gitIn(dir, ...args);
+  g('init', '-q', '-b', 'main', '.');
+  g('config', 'core.autocrlf', 'false');
+  g('config', 'user.email', 'test@example.com');
+  g('config', 'user.name', 'Test');
+  // A second runner with stderr piped, for the calls whose git failure is the
+  // expected outcome — gitIn inherits stderr, so those would print git's own
+  // `fatal:` line into an otherwise passing suite.
+  const quiet = (...args) => execFileSync('git', args, { cwd: dir, encoding: 'utf-8', stdio: ['pipe', 'pipe', 'pipe'] }).trim();
+  return { dir, g, ctx: { git: g }, quietCtx: { git: quiet } };
+}
+
+function writeFixture(dir, rel, text) {
+  const abs = join(dir, rel);
+  mkdirSync(dirname(abs), { recursive: true });
+  writeFileSync(abs, text, 'utf-8');
+}
+
+// A real FETCH_HEAD, written the way `git fetch` writes it, so rev-parse
+// resolves it exactly as it would after a fetch from the canonical repo.
+function setFetchHead(dir, sha) {
+  writeFileSync(join(dir, '.git', 'FETCH_HEAD'), `${sha}\t\tbranch 'main' of https://example.invalid/repo\n`, 'utf-8');
+}
+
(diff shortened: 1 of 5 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is in `update-system.mjs`, which is a red-line file and the maintainer's call; I am reviewing the two owned tests.

Both are sound and mutation-sensitive:

`tests/updater-apply-target-verification.test.mjs`
- The `pinRefToCommit`/`versionAtRef` section drives the real exports against a throwaway git repo, and the TOCTOU case is two-sided: it moves FETCH_HEAD to an older SHA, asserts the pinned read still sees the intended version, and has a control asserting the unpinned read does follow the move, so the pin is shown to be load-bearing. Gutting `pinRefToCommit` reddens it.
- The downgrade-refusal section is a two-sided predicate (older/patch/major/empty-VERSION refused, newer/same allowed), fail-closed on an empty target.
- The apply() section pins both the call-site (every FETCH_HEAD read is inside `pinRefToCommit`) and the ordering (the verify-and-throw runs before the first checkout/unlink), with comments stripped before matching so a mention cannot spoof it. Deleting the verify block from apply() reddens the ordering guard.

`tests/updater-self-bootstrap-backup.test.mjs` pins the order backup-before-checkout; moving the checkout ahead of backup, or removing the backup call, reddens it. It verifies the call is present and correctly ordered rather than the backup's runtime effect (a different function, out of this diff's scope), which is a reasonable boundary, not a false-pass.

update-system.mjs
```diff
@@ -2029,15 +2099,54 @@ async function apply() {
...
+    // The re-exec'd child re-fetches, so without the inherited SHA it would pin
+    // its own target — and then install a tree different from the one the parent
+    // verified and bootstrapped from. The env value is a hint, not an
+    // authorization (#2866): it is only read in the authenticated re-exec, it is
+    // validated by rev-parse against the local object store, and an unusable
+    // value falls back to a fresh pin rather than aborting.
+    const inheritedTarget = isReexec ? (process.env.CAREER_OPS_UPDATE_TARGET_SHA || '') : '';
+    let targetRef;
+    try {
+      targetRef = inheritedTarget ? pinRefToCommit(inheritedTarget) : pinRefToCommit('FETCH_HEAD');
```
nikitacometa: Pins the ref, never checks it is upstream main. Probe: FETCH_HEAD naming a side-branch tag with VERSION 9.9.9 against installed 1.26.0, `downgradeRefusal` returns null and the rogue tree installs. `check()` resolves `refs/heads/main` to `remoteCommit` at 1795-1801; consume it here, or fetch `+refs/heads/main:refs/career-ops-update/main --no-tags` and pin that ref.

update-system.mjs
```diff
@@ -2029,15 +2099,54 @@ async function apply() {
...
+    } catch {
+      targetRef = pinRefToCommit('FETCH_HEAD');
```
nikitacometa: Fails open: this catch also swallows a failure to pin `inheritedTarget`, so the child re-pins its own `FETCH_HEAD` while running bootstrap files from the parent's target (2163). When `isReexec && inheritedTarget`, throw.

update-system.mjs
```diff
@@ -2029,15 +2099,54 @@ async function apply() {
...
+    }
+
+    // 2b. Verify the target before touching a single file. apply() never did:
+    // compareVersions() is called only from check(), which decides whether to
+    // NOTIFY, so the install path checked out whatever the fetch produced and
+    // printed a success banner regardless of which direction the version moved.
+    // Refuse an older target instead, and refuse a target whose VERSION cannot
+    // be read at all — an unverifiable target is not a verified one.
+    const targetVersion = versionAtRef(targetRef);
```
nikitacometa, line 2435: Nothing pins this to `targetRef`: `versionAtRef('HEAD')` makes the guard compare local against local and return null, and 184 updater tests plus 384 migration checks stay green. Add the call shape to updater-migration-tests.mjs.

update-system.mjs
```diff
@@ -2029,15 +2099,54 @@ async function apply() {
...
+    const refusal = downgradeRefusal(local, targetVersion);
+    if (refusal) {
+      throw new Error(
+        `Refusing to update from ${targetRef.slice(0, 12)}: ${refusal}.\n` +
+        `    No system files were changed. To go back to an earlier state deliberately,\n` +
```
nikitacometa: False on the re-exec path: the parent already ran `git('checkout', targetRef, '--', ...reexecFiles)` at 2163 before the child reaches this guard. Branch the wording on `isReexec`, or the user skips `rollback` and keeps a half-updated tree.

update-system.mjs
```diff
@@ -684,6 +684,75 @@ function compareVersions(a, b) {
...
+ *
+ * Fail closed: an unreadable target VERSION is not evidence of a newer target,
+ * so it is refused rather than assumed benign. Same-version targets stay allowed
+ * — re-applying the current release is a legitimate repair path (#1998).
+ *
+ * @param {string} installedVersion - VERSION on disk before the update.
+ * @param {string} targetVersion - VERSION shipped by the pinned target ref.
+ * @returns {string|null} Refusal reason, or null when the target is acceptable.
+ */
+export function downgradeRefusal(installedVersion, targetVersion) {
```
nikitacometa, line 785: Fails open on a malformed value: `compareVersions` coerces with `|| 0` and ignores components past the third, so `999.bad.0`, `999.0.0-rc` and `1.32.0.1` all return null against installed 1.32.0 (measured at this head). Refuse anything that is not `^\d+\.\d+\.\d+$` before comparing.

tests/updater-apply-target-verification.test.mjs
```diff
@@ -0,0 +1,226 @@
...
+
+    const reads = body.match(/FETCH_HEAD/g) || [];
+    const pinned = body.match(/pinRefToCommit\('FETCH_HEAD'\)/g) || [];
+    if (reads.length > 0 && reads.length === pinned.length) {
+      pass(`every FETCH_HEAD read in apply() goes through pinRefToCommit (${pinned.length})`);
+    } else {
+      fail(`apply() reads FETCH_HEAD outside the pin: ${reads.length} occurrence(s), ${pinned.length} pinned`);
+    }
+
+    const guardAt = body.indexOf('downgradeRefusal(');
```
nikitacometa: Asserts the position of the `downgradeRefusal(` call, not that a refusal aborts: `if (false && refusal) {` keeps all 15 green including this one; only updater-migration-tests.mjs:282 reddens. Match `if (refusal) {` followed by the throw here too.

nikitacometa (changes requested):
The pin and the downgrade guard are sound at de57f11f (`--only updater` 184/0, migration checks 384/0, gutting `pinRefToCommit` reddens the new suite). Four items before this leaves draft.

1. It verifies version direction, not target identity. update-system.mjs:2119 pins whatever `FETCH_HEAD` names, and 2100 is still `git('fetch', CANONICAL_REPO, 'main')` with no `--no-tags` and no explicit refspec, so a target that is not upstream main but ships a VERSION at or above the installed one passes 2131. Probe against the PR's own exports: FETCH_HEAD naming a side-branch tag with VERSION 9.9.9, installed 1.26.0, `downgradeRefusal` returns null and apply proceeds with the rogue tree. `check()` already resolves the authoritative SHA (`remoteCommit`, 1795-1801); `apply()` never consumes it. That is #3052's expected-behaviour item 1.
2. `downgradeRefusal` fails open on a malformed VERSION: `compareVersions` coerces with `|| 0` and stops at three components, so `999.bad.0`, `999.0.0-rc` and `1.32.0.1` all return null against installed 1.32.0 (measured). Validate the shape before comparing.
3. update-system.mjs:2121 fails open: the catch also swallows a failure to pin `inheritedTarget`, so a child that cannot resolve the parent's SHA re-pins its own `FETCH_HEAD` while running bootstrap files the parent checked out at 2163. That is the mixed state #3052 reports. When `isReexec && inheritedTarget`, throw.
4. Nothing pins the guard to the pinned target: mutant `versionAtRef('HEAD')` at 2130 makes it compare the installed tree with itself, always null, and survives 184 updater tests plus 384 migration checks. One call-shape pattern in updater-migration-tests.mjs closes it. The ordering assertion at tests/updater-apply-target-verification.test.mjs:215 has the same gap: `if (false && refusal)` keeps all 15 green.

Smaller: 2135 says "No system files were changed" on the re-exec path after the parent already checked out `reexecFiles`; and `remote = localVersion()` at 2497 re-reads disk, so a preserved VERSION still prints #3052's own `v1.26.0 → v1.26.0`. The branch is behind main and conflicts with #3742, #3781 and #3782 in `apply()`.

Scott-Emberson (commented):
Test-axis only. The substance is `update-system.mjs`, a red-line file and the maintainer's call; I am reviewing the owned tests, and since this is updater data-integrity I re-verified from scratch after two of the three owned tests changed.

They are sound and strengthened. I ran the suite through the real discovery harness at head (apply-target 31/0, local-system 25/0, self-bootstrap 4/0) and mutated `update-system.mjs`: deleting the verify-block from `apply()` reddens four assertions (guard-ordering, the `versionAtRef(targetRef)` argument, the identity cross-check, and the pin-fallback), making the downgrade guard inert reddens one, unpinning one `pinRefToCommit('FETCH_HEAD')` reddens the two-sided TOCTOU pin, and making the identity guard read `HEAD` instead of `targetRef` reddens one. The changes are strengthening, not weakening: the only removed assertion is the old weaker `indexOf` ordering check, replaced by an exact full-statement regex (`versionAtRef(targetRef)` then `downgradeRefusal(local, ...)` then `throw`) that defeats a `false && refusal` inert-guard mutant, plus new blocks for an unparseable-version refusal, a target-identity refusal (the target must descend from authoritative main), and refusal-message truthfulness, and the local-system additions are purely additive coverage of the new `pathFullyPreserved` helper.

Scott-Emberson (commented):
Test-axis only. The substance is in `update-system.mjs` (the updater's target-verification path), a data-loss red-line and the maintainer's call; I am reviewing the three owned tests, and since this is the updater I verified strictly, specifically for the #3955 shape where a test hand-reconstructs apply()'s probe logic instead of driving it.

This PR does not repeat that shape. The fix extracts the guard logic into real exported pure functions (`pinRefToCommit`, `versionAtRef`, `downgradeRefusal`, `targetIdentityRefusal`, `refusalMessage`, `authoritativeShaFromRefBody`, and the re-signatured `pathFullyPreserved`), and apply() genuinely calls those same functions rather than inline copies. The tests import and drive those exports against throwaway temp repos through the `ctx.git` seam. All three are sound and mutation-sensitive: for the new apply-target-verification test, neutering `downgradeRefusal` and swapping apply()'s `versionAtRef(targetRef)` to `'HEAD'` each redden; for local-system-edits, flipping the non-directory branch reddens; for self-bootstrap-backup, swapping apply()'s checkout target reddens. Root pinning is clean (temp repos via mkdtempSync, explicit `ctx.git`, no import-time stray writes), so it is not the #3370 shape.

One caveat worth the maintainer's eye, not a false-pass: two of the checks (all of the self-bootstrap test, and one section of the new test) are source-text regex assertions on apply()'s body rather than executed calls, because apply() is not exported and cannot run without a network, git remote, and re-exec. That is the correct countermeasure to the #3955 problem (it proves apply() literally names the real functions and orders the guard before the first write, and it even pins the exact `versionAtRef('HEAD')` mutant), but text assertions can go stale silently under a future rename or refactor of apply(). The bulk of the coverage is genuine execution against the real exports, so this is a limitation of two checks, not a false pass.

Owned tests sound; branch behind main. SIGNAL (update-system.mjs red-line), so not surfacing as ready-to-approve; routing and merge are the maintainer's.

santifer:
Sorry for the wait, @maxmilian: you answered my one line on the 6th and I left it sitting for two weeks.

Re-read at 1e1130d, and the two things I wanted to be sure of both hold: inside apply() the only live `FETCH_HEAD` is the single `pinRefToCommit('FETCH_HEAD')`, everything else addresses the pinned SHA; and `stagingFileList` and `resolveReexecCheckout` really do take a ref, so threading the target through them is not an argument landing nowhere. CI is green on this head, windows included, and it merges onto current main cleanly.

`authoritativeShaFromRefBody()` is the shape I asked for, and your reading of it is the right one: not a new refusal, just us no longer asserting something about the target that was never checked.

Agreed on leaving CodeRabbit's second note as a follow-up: it is about that file's assertion windows rather than about this fix.

@nikitacometa: re-review request on 1e1130d. Your four items and the two smaller ones were addressed, I verified them at 07aa962c, and the last piece landed in this head. If you would rather not take another pass, say so and I will dismiss the stale request with that note.

One limit from my side: I have not re-run the updater suites locally, so this is a code read plus CI on this head. The branch update and the merge stay my call, @maxmilian, so nothing is blocked on you.

PR #3808, opened 2026-09-04
fix(analyze-patterns): parse report headers written in French

### LINKED ISSUE #3807: analyze-patterns: report headers written in French parse as absent (URL, Archetype)
parseReport() reads the `URL` and `Archetype` header fields with patterns that require an ASCII colon directly after the label, and no accent:

```js
const urlMatch = plain.match(/^URL:\s*(https?:\/\/\S+)/im);
const headerArchRegex = /^(?:Archetype|Arquetipo):\s*(.+?)$/im;
```

French typography puts a space before a colon, and the modes write the report header in the user's `language.output`. A French report therefore carries:

```
**URL :** https://job-boards.greenhouse.io/acme/jobs/1
**Archétype :** Delivery Manager
```

Neither pattern matches. Both fields parse as absent.

### Why it is hard to notice

A missing header field is a legitimate state for reports predating the field, so nothing warns and nothing throws. The run exits 0 and answers from a fraction of the corpus.

Measured on a 224-report corpus written in French:

| Field | Parsed | Lost silently |
|---|---|---|
| `URL` | 96 | **128** |
| `Archetype` | 116 | **61** |

The ATS vendor is derived from that URL, so the vendor breakdown spoke for **1 submitted application out of 22**, reporting `coveragePct: 5` — which reads like sparse data rather than a parser miss.

This is the same shape as #3495: a parse that empties the analysed population instead of breaking it.

### Suggested fix

Allow optional whitespace before the colon (`\s` also covers U+00A0, which French editors insert automatically) and accept the accented spelling, keeping the accent-less and Spanish `Arquetipo` variants so existing reports are unaffected.

### Out of scope, noted for later

32 reports in the same corpus use a `**Archétype visé :**` label. That is a label drift produced by the modes rather than a locale issue, so it is not covered here.

PR follows.

thibmonier:
Fixes #3807.

`parseReport()` read the `URL` and `Archetype` header fields with patterns requiring an ASCII colon directly after the label and no accent. French typography puts a space before a colon, and the modes write the header in the user's `language.output`, so `**URL :**` and `**Archétype :**` never matched and both fields parsed as absent.

The failure is silent by construction: a missing header field is a legitimate state for older reports, so nothing warns and nothing throws. The run exits 0 and answers from a fraction of the corpus.

On a 224-report French corpus: **128 reports lost their URL**, 61 lost their archetype. Since the ATS vendor is derived from that URL, the vendor breakdown spoke for 1 submitted application out of 22 while reporting `coveragePct: 5` — indistinguishable from genuinely sparse data.

Same shape as #3495.

### What changed

- The two patterns move to module scope as `REPORT_URL_RE` and `REPORT_ARCHETYPE_RE`, so the behaviour is directly testable rather than buried in a 200-line function.
- Both accept optional whitespace before the colon. `\s` covers U+00A0, which French editors insert automatically.
- `Arch[eé]type` accepts the accented spelling. The accent-less form and the Spanish `Arquetipo` variant are unchanged, so existing reports keep parsing.

### Tests

Self-test coverage for both fields across the EN, FR and ES spellings, including the non-breaking-space form, plus a negative case asserting that a line carrying no URL field still yields `null` — so the fix cannot turn "field absent" into a false positive.

Confirmed the tests are meaningful: the FR cases fail against the previous patterns.

### Verification

- `node analyze-patterns.mjs --self-test` passes.
- `node test-all.mjs` shows **no change against baseline** on the same working tree (8053 passed / 6 failed before and after; the 6 are pre-existing and unrelated — SYSTEM_PATHS coverage, user-layer root resolution, root-level suite discovery).
- End to end: a French report carrying a Greenhouse URL is now detected and appears in the vendor breakdown, where it was previously bucketed as `unknown`.

### Not covered

32 reports in the same corpus use a `**Archétype visé :**` label. That is label drift from the modes rather than a locale issue, so it is left out to keep this diff to one well-defined problem.

## Summary

French reports now retain their URL and archetype fields when parsed.

- `analyze-patterns.mjs:859-860`: Header patterns accept spaces and non-breaking spaces before colons.
- `analyze-patterns.mjs:860`: Archetype parsing supports `Archétype`, `Archetype`, and `Arquetipo`.
- `analyze-patterns.mjs:534-548`: Self-tests cover supported languages, whitespace forms, split-line headers, and absent URLs.
- `analyze-patterns.mjs:928-938`: `parseReport()` uses the shared patterns.
- `Archétype visé` remains out of scope.

System files touched: `analyze-patterns.mjs` only. `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/` are unchanged.

### DIFF
--- a/analyze-patterns.mjs
+++ b/analyze-patterns.mjs
@@ -543,6 +543,41 @@ requirement_importance:
     if (got !== expected) failures.push(`detectVendor(${JSON.stringify(url)}) → ${JSON.stringify(got)}, expected ${JSON.stringify(expected)}`);
   }
 
+  // Report header fields survive the locale they were written in. French output
+  // puts a space before the colon and accents "Archétype"; both used to make the
+  // field parse as absent, silently. parseReport strips `**` first, so the cases
+  // below are given in that post-strip form. Call the pattern rather than the
+  // string: `subject.match(re)` makes CodeQL read these URL fixtures as patterns
+  // (js/incomplete-hostname-regexp) because of their unescaped host dots.
+  const headerCases = [
+    ['URL: https://job-boards.greenhouse.io/acme/jobs/1', REPORT_URL_RE, 'https://job-boards.greenhouse.io/acme/jobs/1'],
+    ['URL : https://job-boards.greenhouse.io/acme/jobs/1', REPORT_URL_RE, 'https://job-boards.greenhouse.io/acme/jobs/1'],
+    ['URL\u00a0: https://jobs.lever.co/acme/x', REPORT_URL_RE, 'https://jobs.lever.co/acme/x'],
+    ['Archetype: Delivery Manager', REPORT_ARCHETYPE_RE, 'Delivery Manager'],
+    ['Archétype : Delivery Manager', REPORT_ARCHETYPE_RE, 'Delivery Manager'],
+    ['Arquetipo: Delivery Manager', REPORT_ARCHETYPE_RE, 'Delivery Manager'],
+  ];
+  for (const [line, re, expected] of headerCases) {
+    const got = re.exec(line)?.[1] ?? null;
+    if (got !== expected) failures.push(`report header ${JSON.stringify(line)} → ${JSON.stringify(got)}, expected ${JSON.stringify(expected)}`);
+  }
+
+  // A header that is genuinely absent must still yield null, so the fix cannot
+  // turn "no field" into a false positive.
+  if (REPORT_URL_RE.test('Score: 4.4/5')) failures.push('REPORT_URL_RE matched a line with no URL field');
+
+  // A header field is single-line. A label split from its colon, or a value pushed to
+  // the next line, must not parse — otherwise the capture picks up whatever URL happens
+  // to follow in the document.
+  const splitHeaderCases = [
+    ['URL\n: https://elsewhere.example.com/x', REPORT_URL_RE],
+    ['URL :\nhttps://elsewhere.example.com/y', REPORT_URL_RE],
+    ['Archétype\n: Delivery Manager', REPORT_ARCHETYPE_RE],
+  ];
+  for (const [text, re] of splitHeaderCases) {
+    if (re.test(text)) failures.push(`split-line header parsed as valid: ${JSON.stringify(text)}`);
+  }
+
   // Via channel analysis (#1596): agency vs direct yield, normalized buckets.
   const viaRows = [
     { via: 'Hays', normalizedStatus: 'interview' },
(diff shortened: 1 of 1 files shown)


analyze-patterns.mjs
```diff
@@ -531,6 +531,27 @@ requirement_importance:
     if (got !== expected) failures.push(`detectVendor(${JSON.stringify(url)}) → ${JSON.stringify(got)}, expected ${JSON.stringify(expected)}`);
   }
 
+  // Report header fields survive the locale they were written in. French output
+  // puts a space before the colon and accents "Archétype"; both used to make the
+  // field parse as absent, silently. parseReport strips `**` first, so the cases
+  // below are given in that post-strip form.
+  const headerCases = [
+    ['URL: https://job-boards.greenhouse.io/acme/jobs/1', REPORT_URL_RE, 'https://job-boards.greenhouse.io/acme/jobs/1'],
```
nikitacometa, line 553: Source of the 2 CodeQL alerts blocking merge: this tuple holds a string and a RegExp, so `re` at 546 can be this string. One array per regex, or objects with the regex outside the tuple, clears it.

analyze-patterns.mjs
```diff
@@ -819,6 +840,25 @@ function readTextIfExists(path) {
...
+// The failure is invisible and it empties the analysis rather than breaking it:
+// on a 224-report corpus, 128 reports lost their URL (so their ATS vendor fell
+// into the `unknown` bucket, leaving the vendor breakdown to speak for 1 report
+// out of 22 submitted) and 61 lost their archetype. Same shape as #3495.
+//
+// `\s*` before the colon covers both the ASCII space and U+00A0, which French
+// editors insert automatically. Keep the accent-less spellings: reports written
+// before this fix, and the `Arquetipo` Spanish variant, must keep parsing.
+const REPORT_URL_RE = /^URL\s*:\s*(https?:\/\/\S+)/im;
+const REPORT_ARCHETYPE_RE = /^(?:Arch[eé]type|Arquetipo)\s*:\s*(.+?)$/im;
```
nikitacometa: `\s*` matches `\n`: an empty `Archétype :` field captures the following `Score : 4.2/5` line (measured on six reports: `[('Score : 4.2/5', 6)]`). Use `[^\S\r\n]*` and add a split-line negative case. Same line misses the labels shipped in modes/de, pl, it, pt, id, tr, da, ru, ua; #3679's FIELD_KEYS has the list.

analyze-patterns.mjs
```diff
@@ -531,6 +531,27 @@ requirement_importance:
...
+  // puts a space before the colon and accents "Archétype"; both used to make the
+  // field parse as absent, silently. parseReport strips `**` first, so the cases
+  // below are given in that post-strip form.
+  const headerCases = [
+    ['URL: https://job-boards.greenhouse.io/acme/jobs/1', REPORT_URL_RE, 'https://job-boards.greenhouse.io/acme/jobs/1'],
+    ['URL : https://job-boards.greenhouse.io/acme/jobs/1', REPORT_URL_RE, 'https://job-boards.greenhouse.io/acme/jobs/1'],
+    ['URL\u00a0: https://jobs.lever.co/acme/x', REPORT_URL_RE, 'https://jobs.lever.co/acme/x'],
+    ['Archetype: Delivery Manager', REPORT_ARCHETYPE_RE, 'Delivery Manager'],
+    ['Archétype : Delivery Manager', REPORT_ARCHETYPE_RE, 'Delivery Manager'],
+    ['Arquetipo: Delivery Manager', REPORT_ARCHETYPE_RE, 'Delivery Manager'],
```
nikitacometa, line 558: Block A at 921 keeps `(?:Archetype|Arquetipo)` and runs before the header fallback, so a localized Block A table stays unreadable even with this fix.

nikitacometa (changes requested):
CI is red from this PR's own lines, not flake. CodeQL raises 2 high alerts on analyze-patterns.mjs:539-540: `headerCases` mixes strings and RegExps in one array literal, and 546 destructures `re` into `line.match(re)`, so the URL strings reach a regexp position (`js/incomplete-hostname-regexp`). `vendorCases` at 517 holds the same greenhouse URLs and stays clean because it has no RegExp element. Split the array per regex, or key the cases by name.

860: `\s*` also matches newlines. Six French reports with an empty `**Archétype :**` field followed by `**Score :** 4.2/5` give `archetypeBreakdown: [('Score : 4.2/5', 6)]` at this head against `[('Unknown', 6)]` on main, and that value feeds `bestArchetype` (1374), so the recommendation becomes "double down on Score : 4.2/5". Empty header fields are a shipped shape (modes/oferta.md:666). `[^\S\r\n]*` on both patterns plus a split-line negative case fixes it.

The fix itself holds: self-test at head passes, the mutant with main's regexes and the new cases fails on exactly the named cases, and `\s` does cover U+00A0. Coverage on a 12-locale corpus goes URL 10/12 to 12/12 and archetype 3/12 to 5/12; `Archetyp` (de, pl), `Archetipo` (it), `Arquétipo` (pt), `Arketipe` (id), `Arketip` (tr), `Arketype` (da) and `Архетип` (ru, ua) still read as Unknown, and Block A at 921 keeps the English-only pattern. #3679 carries that label list for the web header; one map for both would end the class.

Issue #2679, opened 2026-08-11
feat: network matching ("do I know anyone here?") from the user's own connections export


henryjrobinson:
### Code of Conduct

- [x] I agree to follow the [Code of Conduct](https://github.com/santifer/career-ops/blob/main/CODE_OF_CONDUCT.md)

### Existing issues

- [x] I searched existing issues and this hasn't been requested yet

### Problem

When a scan surfaces a job, career-ops tells me how well the role matches my CV. It cannot tell me whether I have a human path into the company. For anyone with an established network, that path is usually the biggest single factor in whether an application gets read at all, and the answer is sitting in data I already own.

The user story: as a job searcher, I want to know who works or has worked at a company I am pursuing, up to two degrees away, so I can reach out for a warm introduction.

Checking this by hand costs a few minutes per company. That is exactly enough that it gets skipped when I am working through a batch of scans, and the batch is when it matters most. Two things make the manual path worse than it looks:

- Cold outreach is rate-limited. InMail quotas cap how many people outside your network you can contact, so a mutual connection is often not the better route but the only one available.
- The check means opening LinkedIn, and LinkedIn is built to make sure you don't leave. That is a real tax on a batch workflow.

So this is a throughput feature. It automates a step that everyone skips under volume.

### Proposed solution

Add a `network` field to the job record: a list of people from the user's own network who work, or have worked, at the posting's company, with profile links.

#### What each data source can and cannot answer

Every major professional network prohibits automated access in its terms of service, and there is no sanctioned API that returns your connections' employment histories. The design assumes no scraping and splits by what the user legitimately possesses:

| Question | Source | Availability |
|---|---|---|
| Who works there now | Your own connections export (the self-service data download) | Complete, offline, free |
| Who worked there before | Not in any export. Optionally derivable from your own mailbox (correspondence domains over time), or purchasable from an enrichment API | Partial |
| Who knows someone there | Exists only inside the platform UI | Not exportable at all |

The connections export gives name, profile URL, current company, current title, and connection date. It does not give employment history or second-degree edges. That gap is deliberate on the platform's side, and it defines the shape of the feature.

#### Design: precompute what the user owns, link out for the rest

The precomputed part costs nothing per job: career-ops parses the connections export once into a company-to-people index, and every scored job gets a `network` field listing matching connections.

[...]

PR #3811, opened 2026-09-04
fix(linkedin-join): use the real people-search route

JOduMonT:
`secondDegreeSearchUrl()` builds `https://www.linkedin.com/search/people/?…`. That path is not a LinkedIn route — it renders a "Page doesn't exist" error — so every second-degree link the warm-intro finder prints is dead on arrival.

Found the ordinary way: a user clicked one of the links `linkedin-join.mjs --summary` had just printed and got the error page.

The working route is `/search/results/people/`. The query string was already correct; only the path was wrong.

### Why the test didn't catch it

```js
assert.ok(url.startsWith('https://www.linkedin.com/search/people/?'));
```

The assertion encoded the same wrong path as the implementation, so it passed against the bug. Corrected alongside, and the inline `--self-test` check had the same issue.

### Verification

```
node --test tests/linkedin-join.test.mjs   → 22 pass, 0 fail
node linkedin-join.mjs --self-test         → PASS 43/43
```

Behaviour is otherwise unchanged: still constructed and never fetched, so the #2679 criterion-3 property (career-ops builds the string, the user opens it) still holds.

https://claude.ai/code/session_0118pmJ5jiXf3ggjd7bHQCWu

## Summary

Updated `secondDegreeSearchUrl()` in `linkedin-join.mjs` to use LinkedIn’s working `/search/results/people/` route. The query string remains unchanged.

Updated the matching test and inline self-test in `tests/linkedin-join.test.mjs` and `linkedin-join.mjs`.

Users receive a valid LinkedIn people-search URL instead of the previous “Page doesn’t exist” route. The URL is constructed but not fetched.

## Verification

22 unit tests and 43 self-tests pass.

## Files changed

: `linkedin-join.mjs`
: `tests/linkedin-join.test.mjs`

The following system files were not changed: `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, `.github/`.

### DIFF
--- a/linkedin-join.mjs
+++ b/linkedin-join.mjs
@@ -496,7 +496,7 @@ export function parseKnownContacts(content) {
  */
 export function secondDegreeSearchUrl(company) {
   const q = encodeURIComponent(String(company || '').trim());
-  return `https://www.linkedin.com/search/people/?keywords=${q}&network=%5B%22S%22%5D`;
+  return `https://www.linkedin.com/search/results/people/?keywords=${q}&network=%5B%22S%22%5D`;
 }
 
 // --- Join ------------------------------------------------------------------
@@ -790,7 +790,7 @@ function selfTest() {
   check('2nd-degree url encodes the company', url.includes('keywords=Acme%20%26%20Co'));
   check('2nd-degree url filters to 2nd degree', url.includes('network=%5B%22S%22%5D'));
   check('2nd-degree url is linkedin people search',
-    url.startsWith('https://www.linkedin.com/search/people/?'));
+    url.startsWith('https://www.linkedin.com/search/results/people/?'));
 
   const placeholders = parseTrackerTargets([
     '| # | Date | Company | Role | Score | Status | PDF | Report | Notes |',
--- a/tests/linkedin-join.test.mjs
+++ b/tests/linkedin-join.test.mjs
@@ -155,7 +155,7 @@ test('people already in the phonebook are marked, not re-suggested', () => {
 
 test('every target carries a second-degree link the user opens themselves (#2679 criterion 3)', () => {
   const url = secondDegreeSearchUrl('Acme & Co');
-  assert.ok(url.startsWith('https://www.linkedin.com/search/people/?'));
+  assert.ok(url.startsWith('https://www.linkedin.com/search/results/people/?'));
   assert.ok(url.includes('keywords=Acme%20%26%20Co'));
   assert.ok(url.includes('network=%5B%22S%22%5D'), 'must filter to 2nd degree');
 });


Scott-Emberson (commented):
Test-axis only. The substance is in `linkedin-join.mjs`, which is the maintainer's routing call; I am reviewing the owned test.

`tests/linkedin-join.test.mjs` is sound. It pins the new people-search route with a concrete prefix assertion, `url.startsWith('https://www.linkedin.com/search/results/people/?')`, so if the fix reverted to the old `/search/people/?` route the `startsWith` fails and the test reddens. It also pins the query (`keywords=Acme%20%26%20Co` exact encoding, `network=%5B%22S%22%5D` for the 2nd-degree param), and the URL builder is synchronous string construction with no network call, consistent with the zero-token/offline/read-only contract.

linkedin-join.mjs
```diff
@@ -496,7 +496,7 @@ export function parseKnownContacts(content) {
  */
 export function secondDegreeSearchUrl(company) {
   const q = encodeURIComponent(String(company || '').trim());
-  return `https://www.linkedin.com/search/people/?keywords=${q}&network=%5B%22S%22%5D`;
+  return `https://www.linkedin.com/search/results/people/?keywords=${q}&network=%5B%22S%22%5D`;
```
nikitacometa, line 499: Correct. docs/LINKEDIN_JOIN.md lines 52, 57, 64, 69 and 105 still carry the old `/search/people/` route in sample output; update them here so the docs match the helper.

tests/linkedin-join.test.mjs
```diff
@@ -155,7 +155,7 @@ test('people already in the phonebook are marked, not re-suggested', () => {
 
 test('every target carries a second-degree link the user opens themselves (#2679 criterion 3)', () => {
   const url = secondDegreeSearchUrl('Acme & Co');
-  assert.ok(url.startsWith('https://www.linkedin.com/search/people/?'));
+  assert.ok(url.startsWith('https://www.linkedin.com/search/results/people/?'));
```
nikitacometa, line 158: Includes the trailing `?`, so a mutant appending to the path reddens. An assertion that the doc file contains no `linkedin.com/search/people/` would pin the other copy of this string.

nikitacometa (commented):
Path fix is right and the assertions now pin it at the `?` boundary, so a regression reddens. Reproduced locally at head: 22/22 and self-test 43/43. No network call added; the URL is still string-only at linkedin-join.mjs:564 and 710, so the #2679 boundary holds.

One gap: docs/LINKEDIN_JOIN.md lines 52, 57, 64, 69 and 105 still print the dead `/search/people/` route in sample output. Those are the copy-pasteable lines a user reads, and they now disagree with the tool. One `sed` over that file closes it, and an assertion that the doc contains no `linkedin.com/search/people/` keeps the two copies from drifting again. Approve once that lands.

PR #3857, opened 2026-09-05
Added Plugin

### LINKED ISSUE #3565: feat(cli): four documented scripts have no --help — add the usage block (one script per PR)

Abhishek84313:
## What does this PR do?

Adds a `--help`/`-h` usage block to `plugin-install.mjs` and makes it reject
unrecognized flags instead of ignoring them. Before this, every direct
invocation — `--help`, `-h`, a typo, a repo argument — printed nothing and
exited 0, so there was no way to discover that the install/scaffold commands
actually live on `plugins.mjs`.

`plugin-install.mjs` only, per the one-script-per-PR rule on the issue.

## Related issue

Part of #3565 (the `plugin-install.mjs` slot — assigned to me in santifer's comment).

Deliberately **not** `Fixes #3565`: that issue covers four scripts and three are
still open, so this PR must not auto-close it.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] If this is a new feature or architecture change, I opened an issue first …
- [x] My PR does not include personal data …
- [x] My changes respect the Data Contract …

## Summary

Users can run `plugin-install.mjs --help` or `plugin-install.mjs -h` to view usage and supported flags.

The script rejects unknown flags and stray arguments with usage output and exit status 1. Normal execution remains unchanged. Module imports remain side-effect free.

## Tests

Added CLI tests for help output, invalid arguments, validation order, and import behavior: `tests/plugin-install-cli-flags.test.mjs:1`.

## Files changed

- `plugin-install.mjs:1`: Added usage output, known-flag validation, and direct-execution handling.
- `docs/SCRIPTS.md:...`: Documented the `--help` option.
- `tests/plugin-install-cli-flags.test.mjs:1`: Added CLI behavior tests.


Scott-Emberson (commented):
Test-axis only. The substance is in `plugin-install.mjs`, the plugin engine, which is a red-line file and the maintainer's call; I am reviewing the owned test `tests/plugin-install-cli-flags.test.mjs`.

It is sound and mutation-proven. It runs the real script via `spawnSync` and asserts on actual exit codes and output, not a re-implementation. An unknown flag is rejected with exit 1 and a `/unrecognized flag/i` message that names the flag (the message assertion is load-bearing, since it distinguishes the `validateFlags` path from the operand-reject branch), `--help`/`-h` prints `Usage:` and exits 0, and the ordering case `--help --bogus` exits 1 with the unrecognized-flag message and no `Usage:` on stdout, which is the reverse-order bug and can only pass because validation runs before the help short-circuit. There is also an import-safety case pinning that the CLI tail stays behind `isMainModule` so it never fires when `plugins.mjs` imports this module. I mutation-checked it: removing the `validateFlags` call makes all four cases fall to the "takes no arguments" operand branch and fail on both exit code and message.

I confirmed the change is CLI-flag and help handling only: the three files are a docs row, `plugin-install.mjs` (two added imports plus an `isMainModule`-guarded CLI tail), and the new test. No change to `GITHUB_URL_RE`/`NAME_RE`, manifest validation, safe-clone, hashing, the audit, SHA-pinning, or the allowlist, so the engine security logic is untouched.

No test-axis concern. The branch is behind main and the substance is the red-line plugin engine, so routing and merge are the maintainer's call.

Scott-Emberson (commented):
Re-verified at the rebased head. The two owned files (`plugin-install.mjs` and `tests/plugin-install-cli-flags.test.mjs`) are byte-identical to the head I last verified sound, so the delta is a pure rebase of unrelated main traffic and my prior verification carries: the owned test spawns the real script and is mutation-proven (removing `validateFlags` makes the unknown-flag, `--help`, and reverse-order cases fall to the operand branch and fail on exit code and message), and the plugin-install engine security logic (allowlist, SHA-pinning, manifest validation) is untouched.

No test-axis concern. The branch is now up to date with main, but the substance is the red-line plugin engine, so routing and merge remain the maintainer's call.

PR #3862, opened 2026-09-05
fix(followup-cadence): accept --json, the flag both web follow-up routes pass

nikitacometa:
## What does this PR do?

`followup-cadence.mjs` rejects `--json`, the exact flag both web follow-up routes pass it, so both features have been returning nothing since the flag validation landed. This adds `--json` to the script's `KNOWN_FLAGS` (JSON stays the default output, behaviour is unchanged) and adds a contract test that puts every argv the web spawns to the core script it spawns.

## Related issue

None. Bug fix, sent straight in per CONTRIBUTING.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

## How it works

`ec800a2c` (#3196) moved `followup-cadence.mjs` onto the shared `validateFlags()` helper, which rejects any flag not on the script's own list. The list it introduced (`followup-cadence.mjs:989` before this change) is `['--summary', '--overdue-only', '--applied-days', '--help', '-h']`. `--json` is missing from it, and `--json` is what the web passes:

- `web/src/app/api/followups/route.ts:19`, `execFile("node", [script, "--json"], …)`, which resolves `""` on error
- `web/src/app/api/followups/cadence/route.ts:42`, `execFile("node", [script, "--json"], …)`, which ignores the error argument entirely

Measured on `main` (fad070fe):

```
$ node followup-cadence.mjs --json
Error: unrecognized flag(s): --json. Valid flags: --summary, --overdue-only, --applied-days, --help, -h
exit 1
```

Both routes discard that failure and hand the UI an empty answer, so `today-dashboard.tsx` renders "You're all caught up" from a script that never ran. The flag is not a mode switch (with no flag at all the script already prints JSON), so accepting it is a one-line list change with no behaviour change.

Two smaller things in the same file:

- `USAGE` now names `--json` as the default output form, so `--help` no longer implies the flag is invalid.
- `--summary` plus `--json` together: `--summary` wins, which is what the existing code already does, since `summaryMode` is checked first. Now written down in `USAGE` and in a comment rather than left to be rediscovered.

**Sibling scripts.** Since this is a class of bug and not a one-off, I enumerated the closed set: every `execFile`/`spawn` of a root script under `web/src`, and the flags each one passes, checked against that script's own accepted flags.

| Web call site | argv | Script accepts it? |
|---|---|---|
| `web/src/app/api/followups/route.ts:19` | `followup-cadence.mjs --json` | **no, fixed here** |
| `web/src/app/api/followups/cadence/route.ts:42` | `followup-cadence.mjs --json` | **no, fixed here** |
| `web/src/app/api/doctor/route.ts:18` | `doctor.mjs --json` | yes (`doctor.mjs:33`) |
| `web/src/app/api/status/route.ts:160` | `set-status.mjs --row N <State> --source web --json` | yes (`set-status.mjs:138`, `:173`) |
| `web/src/app/api/portals/verify/route.ts:32` | `verify-portals.mjs` (no flags) | yes |
| `web/src/app/api/tracker/delete/route.ts:58` | `tracker.mjs delete --num N [--dry-run]` | yes (`tracker.mjs:562`, `:571`) |
| `web/src/lib/core/scan.ts:88` | `scan-ats-full.mjs --dry-run --since N --ats X --limit N [--json]` | yes (`scan-ats-full.mjs:255`) |
| `web/src/lib/core/pipeline.ts:68` | `node --input-type=module -e <code>`, no root script in argv | n/a |

`followup-cadence.mjs` was the only mismatch, so this PR is a one-line fix plus the test that keeps it one.

## Tests

New file `tests/web-core-argv-contract.test.mjs` (188 lines, `pass`/`fail` from `tests/helpers.mjs`). It holds the table above and puts each argv to the real script, in a hermetic temp data root (`CAREER_OPS_ROOT` plus a one-row fixture tracker with a fictional company), no network:

- 7 argv probes. Six run the exact web argv and require exit 0, with `JSON.parse(stdout)` where the argv carries `--json`. `scan-ats-full.mjs` would sweep live job boards, so its probe appends `--help` instead: `validateFlags()` checks unrecognized flags *before* `--help`, so a rejected flag still exits 1 naming the flag, and exit 0 with usage means every flag was accepted. `verify-portals.mjs` and `set-status.mjs` are pointed at the sandbox with `CAREER_OPS_PORTALS` and `CAREER_OPS_TRACKER`, because both resolve from the codebase root rather than from `CAREER_OPS_ROOT`.
- 1 static check that every `"--flag"` literal in a listed web source is covered by a probe. Best effort by design: it sees the argv literals the routes write inline, not a flag assembled at runtime from a variable or a template string.
- 2 static checks that the table and the sources cannot drift apart: every `web/src` file that both calls `rootScript()` and spawns must be listed, and every listed source must still spawn.

Results on this branch:

```
$ node test-all.mjs --only web-core-argv
web → core argv contract
  ✅ followup-cadence.mjs --json — exit 0, stdout parses as JSON (web/src/app/api/followups/route.ts)
  ✅ followup-cadence.mjs --json — exit 0, stdout parses as JSON (web/src/app/api/followups/cadence/route.ts)
  ✅ doctor.mjs --json — exit 0, stdout parses as JSON (web/src/app/api/doctor/route.ts)
  ✅ set-status.mjs --row 1 Responded --source web --json — exit 0, stdout parses as JSON (…/status/route.ts)
  ✅ verify-portals.mjs — exit 0 (web/src/app/api/portals/verify/route.ts)
  ✅ tracker.mjs delete --num 1 --dry-run — exit 0 (web/src/app/api/tracker/delete/route.ts)
  ✅ scan-ats-full.mjs --dry-run --since 7 --ats greenhouse --limit 150 --json --help — exit 0 (web/src/lib/core/scan.ts)
  ✅ every --flag literal in the listed web sources is covered by a probe
  ✅ all 8 web sources that spawn a core script are listed here
  ✅ no stale entries — every listed source still spawns a core script
📊 Results: 10 passed, 0 failed, 0 warnings
```

`node test-all.mjs --quick` gives **8104 passed, 0 failed, 13 warnings** (exit 0), against **8093 passed, 0 failed, 13 warnings** measured on `main`: the 10 new assertions plus one more from the syntax section, which prints a line per `.mjs` file and so counts the new file. The 13 warnings are pre-existing and unrelated to this change. `npm run lint` passes on 655 files.

**Mutant checks** (both restored afterwards; in each case the file kept parsing and only the named assertions went red):

1. Removed `--json` from `KNOWN_FLAGS`. The two `followup-cadence.mjs --json` probes went red naming the rejection (`8 passed, 2 failed`), everything else stayed green. That is the bug reproduced through the test.
2. Added `"--verbose"` to the doctor route's argv literal. The static half went red with `web/src/app/api/doctor/route.ts passes --verbose, which no probe above covers` (`9 passed, 1 failed`). That half is not dead code.

The existing `test-all.mjs:15573` assertion titled "followup-cadence --json emits…" could not catch this: it calls `analyzeFromContent()` in-process and never spawns the CLI, so the flag list is never consulted.

## Out of scope

- **The routes' error handling.** `followups/route.ts` resolves `""` on error and `followups/cadence/route.ts` ignores the error argument entirely, which is why a script exiting 1 reads as "no follow-ups" instead of as a failure. That is a separate class of fix and there is already work in flight on it.
- **Any other `--json` semantics.** `--json` stays a no-op alias for the default output; this PR does not make it suppress anything or change an exit code.
- **`set-status.mjs` and `verify-portals.mjs` resolving paths from the codebase root rather than from `CAREER_OPS_ROOT`.** The new test works around it with the env overrides those scripts already support; whether that resolution is right is a separate question.

## Summary

Users can now pass `--json` to `followup-cadence.mjs` without an argument error. JSON remains the default output. If users pass both `--json` and `--summary`, `--summary` takes precedence.

The usage documentation now describes these options: `followup-cadence.mjs`.

A new contract test checks that web routes pass valid arguments to core scripts: `tests/web-core-argv-contract.test.mjs`. It also detects uncovered flags and stale route entries.

## System files touched

### DIFF
--- a/followup-cadence.mjs
+++ b/followup-cadence.mjs
@@ -986,12 +986,20 @@ function printSummary(result) {
 
 // ── CLI flags + help ────────────────────────────────────────────────
 
-const KNOWN_FLAGS = ['--summary', '--overdue-only', '--applied-days', '--help', '-h'];
+// --json is the DEFAULT output form, not a mode switch: it names what the
+// script already does with no flag at all. It is listed here because callers
+// pass it explicitly — web/src/app/api/followups/route.ts and
+// web/src/app/api/followups/cadence/route.ts both spawn `[script, '--json']`
+// — and validateFlags() rejects any flag not on this list, so omitting it
+// made both routes exit 1 and read as "no follow-ups" (#3196 added the
+// validation without the flag the web already passed).
+const KNOWN_FLAGS = ['--summary', '--json', '--overdue-only', '--applied-days', '--help', '-h'];
 const VALUE_FLAGS = ['--applied-days'];
 
 const USAGE = `Usage:
   node followup-cadence.mjs                    # full JSON analysis to stdout
-  node followup-cadence.mjs --summary          # human-readable dashboard
+  node followup-cadence.mjs --json             # same as above, JSON is the default
+  node followup-cadence.mjs --summary          # human-readable dashboard (wins over --json)
   node followup-cadence.mjs --overdue-only     # only show overdue/urgent entries
   node followup-cadence.mjs --applied-days 10  # override applied_first cadence (days)
   node followup-cadence.mjs --help|-h          # print this usage block and exit`;
@@ -1016,6 +1024,8 @@ if (isMainModule(import.meta.url)) {
 
   const result = analyze();
 
+  // --summary wins when both are given: it is the flag that asks for something
+  // other than the default, so it is the one carrying an intention.
   if (summaryMode) {
     printSummary(result);
   } else {
--- a/tests/web-core-argv-contract.test.mjs
+++ b/tests/web-core-argv-contract.test.mjs
@@ -1,0 +1,188 @@
+// tests/web-core-argv-contract.test.mjs — the web app spawns core scripts with
+// a fixed argv, and the core validates its own flags. Nothing linked the two.
+//
+// followup-cadence.mjs gained validateFlags() without `--json` on its
+// KNOWN_FLAGS list, while both web follow-up routes had been spawning
+// `[script, '--json']` all along. The script started exiting 1 with
+// "unrecognized flag(s): --json"; both routes discard the error and resolve to
+// "", so the UI rendered an empty follow-up list and told the user they were
+// caught up. No test could see it: the suite exercised the cadence analysis
+// in-process and never spawned the CLI with the web's argv.
+//
+// This file is that missing link. Every web call site that spawns a root
+// script is listed below with the argv it passes, and the argv is put to the
+// real script.
+import { pass, fail, ROOT, NODE, rmSync, walkFiles } from './helpers.mjs';
+import { spawnSync } from 'child_process';
+import { mkdtempSync, mkdirSync, readFileSync, writeFileSync } from 'fs';
+import { tmpdir } from 'os';
+import { join, relative, sep } from 'path';
+
+console.log('\nweb → core argv contract');
+
+// Probe kinds:
+//   'run'        — spawn the exact argv the web passes and require exit 0.
+//   'flags-only' — the real run would sweep live job boards, so append --help
+//                  instead. validateFlags() checks unrecognized flags BEFORE
+//                  --help, so an argv the script rejects still exits 1 naming
+//                  the flag, and exit 0 with usage means every flag was
+//                  accepted. Same verdict on the flags, no network.
+//   'none'       — the call site spawns node with an inline module rather than
+//                  a root script with flags; listed so the enumeration at the
+//                  bottom stays complete.
+const CALL_SITES = [
+  {
+    source: 'web/src/app/api/followups/route.ts',
+    script: 'followup-cadence.mjs',
+    args: ['--json'],
+    probe: 'run',
+  },
+  {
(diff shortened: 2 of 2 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is in `followup-cadence.mjs`, which is the maintainer's routing call; I am reviewing the owned test `tests/web-core-argv-contract.test.mjs`.

It is sound. It spawns the real `followup-cadence.mjs` with the exact `['--json']` argv both web follow-up routes pass, requires exit 0, then `JSON.parse(stdout)`, so it pins both halves of the fix (flag acceptance and JSON emission), not a mock. It reddens on regression: reverting `--json` off the script's `KNOWN_FLAGS` makes `validateFlags` print `unrecognized flag(s): --json` and exit 1 before `analyze()`, so the probe sees a nonzero status and fails. The argv-contract half is real too: it greps each web source for `"--flag"` literals and fails on any not in the probe's args (flag drift), and enumerates every web spawner of the root script, so the routes' passed argv is tied back to what the script actually accepts. The sandbox is hermetic (temp root, fixture tracker, portals pointed at a nonexistent path), so exit 0 reflects flag acceptance rather than environment luck.

PR #2234, opened 2026-07-28
test(context-budget): enforce every _shared.md heading is classified in SECTION_PRIORITY

KD-one:
## What does this PR do?

Adds a reverse check in test #11 of context-budget.test.mjs: every ## heading in _shared.md must have an explicit entry in SECTION_PRIORITY. Prevents the quiet failure mode where a new section silently defaults to P2 (compressible) because no one remembered to classify it.

## Related issue

Closes #2230

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation / translation
- [ ] Refactor (no behavior change)

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/santifer/career-ops/blob/main/CONTRIBUTING.md)
- [x] If this is a new feature or architecture change, I opened an issue first (bug fixes, providers, docs & translations are exempt — send those straight in)
- [x] My PR does not include personal data (CV, email, real names, scan results, or pipeline data)
- [x] I ran `node test-all.mjs` and all tests pass
- [x] My changes respect the [Data Contract](https://github.com/santifer/career-ops/blob/main/DATA_CONTRACT.md) (no modifications to user-layer files)
- [x] My changes align with the [project roadmap](https://github.com/santifer/career-ops/discussions/156)

## Summary by CodeRabbit

* **Tests**
  * Added integrity checks to ensure all shared content sections have explicit priority settings.
  * Improved failure messages to identify sections that may otherwise receive the default compression priority.

PR #3863, opened 2026-09-05
test: move context-budget.test.mjs under tests/ so the harness discovers it, and pin that no suite lives outside tests/

nikitacometa:
## What does this PR do?

`lib/context-budget.test.mjs` was the only `*.test.mjs` outside `tests/`, so `test-all.mjs` never discovered it and its 54 assertions have never run in CI. This moves it to `tests/context-budget.test.mjs`, adds a guard so a suite cannot be written outside a discovered tree again, and keeps the retired `SYSTEM_PATHS` entry one release so the updater prunes the old file on existing installs.

## Related issue

None. Bug fix in the test layer, sent straight in per CONTRIBUTING. Related context: #2234 (the guard that had been dark), #3735 / #3736 (the root-suite convention).

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

## How it works

`test-all.mjs:108-122` walks `tests/` and nothing else, on purpose: it keeps standalone root-level files out of the run. The cost is that a suite written anywhere else is invisible, and invisible quietly. Measured on `main` (fad070fe):

```
$ node test-all.mjs --only context-budget
  no test files matched --only "context-budget" under tests/
exit 1

$ node lib/context-budget.test.mjs
  Results: 54 passed, 0 failed
exit 0
```

Among those 54 is the #2234 guard that every `## ` heading in `modes/_shared.md` has an entry in `SECTION_PRIORITY`; an unlisted section falls through to `DEFAULT_PRIORITY` and becomes compressible under budget pressure. It had never protected a PR.

1. `git mv lib/context-budget.test.mjs tests/context-budget.test.mjs` (a rename, 93% similarity). The counters, and only the counters, switch to `tests/helpers.mjs`: `test-all.mjs:140` refuses to import a discovered suite containing `process.exit(` and `:175` one that calls `finish()`, and the old file ended with both. Every assertion expression is byte-identical; the diff touches the header comment, the imports, the two helper bodies, the `ROOT` line and the trailing summary.
2. `tests/core-test-layout.test.mjs`: every `*.test.mjs` in the repo, enumerated through `lib/mjs-files.mjs` (which already skips `node_modules`, `output/`, `data/` and nested checkouts), must live under `tests/` or `web/tests/`, the two trees a runner discovers. It is the core half of `tests/web-test-layout.test.mjs`, in `tests/` rather than inline so `--only` can reach it. An empty enumeration fails rather than passes.
3. `update-system.mjs`: the `'lib/context-budget.test.mjs'` entry stays in `SYSTEM_PATHS` with a dated comment. `staleSystemFiles()` (`update-system.mjs:981`) prunes a local file only when it is absent from the new remote tree and still matched by some entry, and no directory-level `lib/` entry exists, so dropping the entry would leave the old file on every upgraded install and the new guard would then fail there. Keeping it one release is the updater's own mechanism for a retired path; the comment says when to drop it. `node validate-system-paths-coverage.mjs` and `node upgrade-tests.mjs --pr-gate` are both green with it.

## Tests

```
$ node test-all.mjs --only context-budget
Results: 54 passed, 0 failed, 0 warnings        # was: no test files matched, exit 1

$ node test-all.mjs --only core-test-layout
  all 349 *.test.mjs files live under tests/ or web/tests/
Results: 1 passed, 0 failed, 0 warnings
```

`node test-all.mjs --quick`: green (8093 on `main` plus the 54 assertions that now run, the guard, and one syntax-section line for the new file). `npm run lint`, `node validate-system-paths-coverage.mjs` (`OK: 1351 tracked files covered`) and `node upgrade-tests.mjs --pr-gate` (GREEN) all pass.

Mutant checks (restored afterwards, the file kept parsing, only the named assertion went red):

1. A stray `lib/stray-mutant.test.mjs`: `core-test-layout` goes red naming it.
2. `'spend tier': 1` deleted from `SECTION_PRIORITY`: the moved suite goes red on `all _shared.md headings classified in SECTION_PRIORITY` (53 passed, 1 failed). The same mutation on `main` leaves `test-all.mjs` fully green, which is the point of the move.

## Out of scope

- The contents of the moved suite: no assertion was added, removed or reworded.
- The `*-tests.mjs` files at the repository root (`upgrade-tests.mjs`, `set-status-tests.mjs` and others): they do not match the discovery glob by design and are invoked deliberately; the guard matches `*.test.mjs` only.
- The nine root suites #3388 moved into `tests/` changed their `SYSTEM_PATHS` entries in place rather than keeping the old ones, so an install that has not upgraded since before that migration may still carry them at the root and would trip this guard. That is a wider audit than this PR.

## Summary

- Moved `lib/context-budget.test.mjs` to `tests/context-budget.test.mjs` so `test-all.mjs` runs its 54 assertions.
- Added `tests/core-test-layout.test.mjs` to reject test files outside `tests/` and `web/tests/`.
- Retained the retired path in `SYSTEM_PATHS` for one release. Upgrades can remove the old file: `update-system.mjs:...`.
- Added the migration allowlist entry for the retired path: `updater-migration-tests.mjs:...`.

## User impact

Users do not need to change test commands. Existing installations remove the retired file during updates. New test files must reside under `tests/` or `web/tests/`, or the layout check fails.

## Files touched

- `tests/context-budget.test.mjs`
- `tests/core-test-layout.test.mjs`
- `update-system.mjs`
- `updater-migration-tests.mjs`

`AGENTS.md`, `modes/`, `DATA_CONTRACT.md`, `providers/`, and `.github/` are not touched.

## Validation

Context-budget, test-layout, quick test, lint, system-path coverage, updater migration, and upgrade-gate checks pass.


Scott-Emberson (commented):
Verified the owned tests. This is GATE-tier on the test-layout axis (the change moves a suite under `tests/` and adds a layout guard), with a red-line `update-system.mjs` path-list touch alongside.

The moved `tests/context-budget.test.mjs` preserves its assertions: it is a 93 percent rename whose only hunks rewire the counters to the shared `helpers.mjs` `pass`/`fail`, repoint imports, and import `ROOT`, with every assertion body (including the #2234 P0-heading guard) untouched and its own summary and `process.exit` correctly dropped so a discovered suite does not exit early. The new `tests/core-test-layout.test.mjs` genuinely enforces the layout on the real tree: it walks via `collectMjsFiles(ROOT)` (a real recursive readdir, not a hardcoded list), filters to `*.test.mjs`, and fails naming any file not under `tests/` or `web/tests/`, with an empty-walk guard so a broken walk cannot false-green. A stray `foo.test.mjs` at root reddens it. The `update-system.mjs` change is a two-line retirement comment above the existing entry, no logic touched.

One honest note, not a false-pass: `web/tests/` is an allowed root by design (web has its own discovery glob and a sibling layout test), and the new `tests/context-budget.test.mjs` path is not added to `SYSTEM_PATHS` (only the old one is retained for orphan-pruning).

Owned tests verified sound. This is GATE-tier for the layout guard, but the branch is behind main and it carries a red-line `update-system.mjs` touch, so I am holding formal approval and leaving that to the maintainer; I will re-confirm at the final head once rebased and green.

Scott-Emberson (commented):
Re-verified at the current head. The two owned test files are byte-identical to the head I last verified sound (`tests/core-test-layout.test.mjs` and the moved `tests/context-budget.test.mjs`), so my prior verification carries: the layout guard walks the real tree via `collectMjsFiles(ROOT)` and fails naming any `*.test.mjs` outside `tests/` or `web/tests/`, with an empty-walk guard, and the moved suite keeps its assertions (including the #2234 P0-heading guard).

The delta since then is the correct paired bookkeeping for the `lib/ to tests/` move and nothing more: `update-system.mjs` keeps `lib/context-budget.test.mjs` in `SYSTEM_PATHS` with a retirement comment so the orphan is pruned on upgrade, and `updater-migration-tests.mjs` adds it to `ALLOWED_MISSING_ENTRIES` so the manifest-consistency check tolerates the moved suite for one release. No test logic or scoring touched.

Owned tests verified sound, CI green, mergeable and not behind. The layout-guard substance is GATE-tier and ours, but the `update-system.mjs`/`updater-migration-tests.mjs` bookkeeping is red-line, so I am leaving that piece to the maintainer. Surfacing it as ready-to-approve on the owned-test axis, holding formal approval for Scott's go-ahead.

PR #3864, opened 2026-09-05
fix(invite-match): extract company names in any script, not just ASCII

### LINKED ISSUE #3861: invite-match: extractCompany() is ASCII-only, so an invite from Nestlé or 株式会社 yields no company and zero candidates
## What happens

`extractCompany()` in `invite-match.mjs` returns `null` for any company name containing a non-ASCII letter, and for any name in a script without capital letters. Because `matchInvite()` has no company to match on, the invite produces **zero candidates** even when the company is sitting in the tracker.

Measured on `main` (fad070fe), each accented case paired with its ASCII control so the cause is unambiguous:

```
node --input-type=module -e "
const { extractCompany } = await import('./invite-match.mjs');
..."

null                  <- Your interview with Nestlé
"Acme"                <- Your interview with Acme

null                  <- Interview at Nestlé for Senior Engineer
"Acme"                <- Interview at Acme for Senior Engineer

null                  <- Phone screen - Nestlé
"Acme"                <- Phone screen - Acme

null                  <- Schedule your interview - Nestlé opportunity
null                  <- Your interview with München Analytics
null                  <- Your interview with Яндекс
null                  <- Your interview with 株式会社日立製作所

"Nestlé"              <- Company: Nestlé
```

Three of the four `COMPANY_LINE_PATTERNS` are affected. The fourth (`Company: X`) captures `(.+)` and works, which is why this is easy to miss: paste an invite that happens to spell out `Company:` and everything looks fine.

## Why

`invite-match.mjs:262-267`:

```js
/interview(?:ing)?\s+(?:with|at)\s+([A-Z][\w.,&' -]{1,60}?)(?:[.,\n]|\s+for\s|\s+regarding\s|$)/i
```

Two separate ASCII assumptions in one character class:

1. **`\w` is ASCII-only in JavaScript** (`[A-Za-z0-9_]`). The name body is a *lazy* `{1,60}?`, so it expands only far enough to reach a terminator. It cannot expand across the `é` in `Nestlé`, the terminator never arrives, and the pattern fails outright rather than matching a truncated name. A mid-name character is enough: `München Analytics` fails on the `ü`.
2. **The leading `[A-Z]`** refuses any first letter outside A-Z, so `Яндекс` and `株式会社日立製作所` never start matching at all. Japanese and Chinese names have no case distinction to satisfy it with.

## Relation to #2517

`normalizeCompanyName()` was made script-preserving in the #2517 lineage. `invite-match.mjs:195` now reads:

```js
.replace(/[^\p{L}\p{M}\p{N} ]/gu, ' ')
```

with a comment naming exactly this failure mode: "the Latin-only class DELETED every non-Latin name, so アクメ株式会社 and Яндекс both produced `''`, and matchInvite() bails on an empty key, so pasting an invite from any company in the ja/ko/zh/zh-TW/ru/ua/ar/hi markets that modes/ ships returned ZERO candidates even when the row was right there."

That is the same user-visible outcome, one function later in the chain. Normalization was fixed; extraction was not, so the same invite still returns zero candidates. It just fails earlier now.

## Impact

career-ops ships modes for ja, ko, zh, zh-TW, ru, ua, ar, hi, tr, de and fr markets. For a user in any of them, and for any Latin-script name carrying an accent (Nestlé, Zürich Insurance, Société Générale, Ørsted), pasting an interview invite reports no match against a tracker row that exists. The user's only signal is an empty candidate list, which reads as "this application is not tracked".

## Suggested fix

Give the three affected patterns the `u` flag and Unicode classes, reusing the class `normalizeCompanyName()` already settled on, and keep `_` in the body class since `\w` admitted it and dropping it reproduces this same failure for an ASCII name that contains one:

```js
([\p{Lu}\p{Lo}][\p{L}\p{M}\p{N}_.,&' -]{1,60}?)
```

`\p{Lo}` covers scripts with no case distinction. The terminators stay as they are.

## Not part of this

The trailing-clause weakness is separate and unchanged by the above: `extractCompany('Interview at Acme Inc on Monday')` returns `"Acme Inc on Monday"`, because `on Monday` is not one of the pattern's terminators. That is a heuristic-quality question, not a script question, and it behaves identically before and after.

Also separate: `Interview invitation from <Company> for <Role>` returns `null` for **any** company, ASCII included (`Interview invitation from Globex Corporation for Senior Engineer` → `null`), because no pattern covers the `invitation from` phrasing. That is a missing pattern, not an encoding bug.

nikitacometa:
## What does this PR do?

`extractCompany()` in `invite-match.mjs` could not read a company name containing a non-ASCII letter, or one written in a script without capitals, so pasting an invite from Nestlé, München Analytics, Яндекс or 株式会社日立製作所 produced no company and therefore zero candidates against a tracker row that exists. This gives the three affected patterns Unicode character classes and the `u` flag.

## Related issue

Closes #3861

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

## How it works

`COMPANY_LINE_PATTERNS` (`invite-match.mjs:260-265` on `main`) built each name as `[A-Z][\w.,&' -]{1,60}?`. Two ASCII assumptions in one class:

- **`\w` is ASCII-only in JavaScript.** The body is a *lazy* `{1,60}?`, so it expands only far enough to reach a terminator. It cannot expand across the `é` in `Nestlé`, the terminator never arrives, and the pattern fails outright rather than matching a truncated name. One accented character anywhere in the name is enough: `München Analytics` fails on the `ü`.
- **The leading `[A-Z]`** refuses any non-Latin first letter, so `Яндекс` and `株式会社日立製作所` never started matching. Names in Japanese and Chinese have no case distinction to satisfy it with.

Measured on `main` (fad070fe), each accented case beside its ASCII control:

```
null                  <- Your interview with Nestlé
"Acme"                <- Your interview with Acme
null                  <- Interview at Nestlé for Senior Engineer
"Acme"                <- Interview at Acme for Senior Engineer
null                  <- Phone screen - Nestlé
"Acme"                <- Phone screen - Acme
null                  <- Schedule your interview - Nestlé opportunity
null                  <- Your interview with München Analytics
null                  <- Your interview with Яндекс
null                  <- Your interview with 株式会社日立製作所
"Nestlé"              <- Company: Nestlé
```

The fourth pattern (`Company: X`) captures `(.+)` and was never script-bound, which is what made this easy to miss: an invite that spells out `Company:` works fine.

The fix, on the three affected patterns only:

```js
([\p{Lu}\p{Lo}][\p{L}\p{M}\p{N}_.,&' -]{1,60}?)   // + the u flag
```

`\p{Lo}` is the "other letter" category, which is where scripts without a case distinction live. The body class is the one `normalizeCompanyName()` already settled on at `invite-match.mjs:195` (`[^\p{L}\p{M}\p{N} ]/gu`, from the #2517 lineage), plus `_` (see below) and the punctuation company names carry. The terminators are untouched.

This closes the second half of #2517. That fix made `normalizeCompanyName()` script-preserving, with a comment naming this exact outcome: a non-Latin name normalized to `''`, `matchInvite()` bailed on the empty key, and the invite returned zero candidates. Extraction was left ASCII-only, so the same invite still returns zero candidates; it just fails one function earlier.

**`_` is carried over into the new class deliberately.** `[\w…]` admitted `_`; `[\p{L}\p{M}\p{N}…]` on its own does not. An earlier draft dropped it and that is a real regression, not a no-op: for a company like `Acme_Corp`, the lazy body class has no reachable terminator right after the `_` (it is not `.`, `,`, a newline, ` for `, ` regarding`, or end of string), so the whole pattern fails and returns `null` instead of extracting a truncated or full name. That is the exact same failure mode this PR fixes for `é` and non-Latin scripts, reintroduced for one ASCII character. The class here is `[\p{L}\p{M}\p{N}_.,&' -]`, keeping `_` so extraction still succeeds on an ASCII name that has one; `normalizeCompanyName()` strips it again during matching, so the final match key is unaffected either way.

**No backtracking risk.** #3463 fixed quadratic backtracking in the sibling `extractReqId`, so I measured this one: `Interview with ` + 50,000 class characters with no terminator returns in 0.1 ms, in both the Unicode and ASCII forms. The `{1,60}` cap bounds the work regardless of input length.

## Tests

20 assertions added to the existing `tests/invite-match.test.mjs`, which goes from **103 passed** to **123 passed**:

- 7 for the affected phrasings: Nestlé in all three patterns, a mid-name umlaut, a Cyrillic name, a Japanese name.
- 10 pinning the ASCII behaviour as unchanged: the pre-existing 7 (including the known trailing-clause weakness, `Interview at Acme Inc on Monday` still returns `"Acme Inc on Monday"`, so this PR is on record as not touching it), plus 3 for a company name containing `_` (`tests/invite-match.test.mjs:398-400`), pinning the fix at `invite-match.mjs:275-277` described above.
- 3 end to end through `analyzeInvite()`: the tracker holds a `Nestlé` row, and the invite must now produce exactly that one candidate. This is the assertion that describes what the user actually lost, since a missing company name is only visible to them as an empty candidate list.

All fixture data is fictional except the company names, which are public brands used as encoding samples.

```
$ node test-all.mjs --only invite-match
📊 Results: 123 passed, 0 failed, 0 warnings          # 103 on main
```

`node test-all.mjs --quick` gives **8113 passed, 0 failed, 13 warnings** (exit 0), against **8093 passed, 0 failed, 13 warnings** measured on `main` in the same worktree: exactly the 20 new assertions, no collateral. `npm run lint` passes on 654 files.

**Mutant checks** (four, each restored afterwards; in every case the file kept parsing):

1. Dropped the `u` flag from pattern 2, the mutation the fix is named for. `110 passed, 13 failed`: every new assertion that reaches that pattern went red, including the ASCII ones, since without `u` the `\p{…}` classes decay into literal characters.
2. Reverted pattern 2's body class to `[\w.,&' -]`, keeping the `u` flag. `115 passed, 8 failed`: the accented, Cyrillic and Japanese cases went red and **the ASCII cases, including the `_` ones, stayed green**. The new tests are not simply red against any change.
3. Dropped `\p{Lo}` from pattern 2's leading class. `122 passed, 1 failed`, and the one failure is `a Japanese name is extracted (no case distinction to lead with)`. Exactly the assertion that names the property, and nothing else.
4. Dropped `_` from all three body classes (the mutation named above). `120 passed, 3 failed`, and the three failures are exactly the `Acme_Corp` assertions; every other assertion, including all the other ASCII and Unicode ones, stayed green.

## Out of scope

- **The trailing-clause weakness.** `extractCompany('Interview at Acme Inc on Monday')` returns `"Acme Inc on Monday"` because `on Monday` is not one of the pattern's terminators. It behaves identically before and after this change and is now pinned by a test so a future fix has a starting point.
- **The `Interview invitation from <Company> for <Role>` phrasing**, which returns `null` for **any** company, ASCII included (`Interview invitation from Globex Corporation for Senior Engineer` → `null`). No pattern covers `invitation from`. That is a missing pattern rather than an encoding bug, and adding phrasings is a wider change than this fix.
- **The `Company: X` pattern**, which captures `(.+)` and was never script-bound. Left alone.

### DIFF
--- a/invite-match.mjs
+++ b/invite-match.mjs
@@ -257,11 +257,24 @@ export function companySimilarity(a, b) {
 // too much for one regex to be authoritative, so this is a best-effort
 // extraction — the fuzzy match against the tracker is what actually decides
 // the result, not this heuristic alone.
+//
+// The name class is Unicode, not `[A-Z][\w…]`: `\w` is ASCII-only in JS, so a
+// lazy `[\w…]{1,60}?` cannot cross the `é` in Nestlé and the terminator it is
+// looking for never arrives — the whole pattern fails and the invite yields no
+// company at all. `[A-Z]` refused a non-Latin first letter for the same reason,
+// so Яндекс and 株式会社 never even started matching. `\p{Lo}` covers
+// scripts with no case distinction; the body class is the one
+// normalizeCompanyName() already uses (#2517), plus the punctuation company
+// names carry, plus `_`: `\w` included it, and dropping it reproduces the
+// exact same failure mode this fix is for (a lazy class with no reachable
+// terminator fails the whole pattern, not just the one character) for any
+// ASCII company name that happens to contain an underscore. The first
+// pattern captures `(.+)` and was never script-bound.
 const COMPANY_LINE_PATTERNS = [
   /(?:^|\n)\s*company\s*[:\-]\s*(.+)/i,
-  /interview(?:ing)?\s+(?:with|at)\s+([A-Z][\w.,&' -]{1,60}?)(?:[.,\n]|\s+for\s|\s+regarding\s|$)/i,
-  /(?:phone screen|screening|interview)\s*[-–—:]\s*([A-Z][\w.,&' -]{1,60}?)(?:\s+opportunity)?(?:[.,\n]|$)/i,
-  /schedule your (?:phone screen|interview)\s*(?:[-–—:]\s*)?([A-Z][\w.,&' -]{1,60}?)\s*opportunity/i,
+  /interview(?:ing)?\s+(?:with|at)\s+([\p{Lu}\p{Lo}][\p{L}\p{M}\p{N}_.,&' -]{1,60}?)(?:[.,\n]|\s+for\s|\s+regarding\s|$)/iu,
+  /(?:phone screen|screening|interview)\s*[-–—:]\s*([\p{Lu}\p{Lo}][\p{L}\p{M}\p{N}_.,&' -]{1,60}?)(?:\s+opportunity)?(?:[.,\n]|$)/iu,
+  /schedule your (?:phone screen|interview)\s*(?:[-–—:]\s*)?([\p{Lu}\p{Lo}][\p{L}\p{M}\p{N}_.,&' -]{1,60}?)\s*opportunity/iu,
 ];
 
 /**
--- a/tests/invite-match.test.mjs
+++ b/tests/invite-match.test.mjs
@@ -15,7 +15,7 @@ import { join, dirname } from 'path';
 import { fileURLToPath } from 'url';
 import { tmpdir } from 'os';
 import { execFileSync } from 'child_process';
-import { matchInvite, normalizeCompanyName, extractPlatform, isAIInterviewerPlatform, classifyEmail, analyzeInvite, applyRejectionStatus, selectApplyTarget } from '../invite-match.mjs';
+import { matchInvite, normalizeCompanyName, extractCompany, extractPlatform, isAIInterviewerPlatform, classifyEmail, analyzeInvite, applyRejectionStatus, selectApplyTarget } from '../invite-match.mjs';
 import { pass, fail } from './helpers.mjs';
 
 console.log('\ninvite-match.mjs — interview invite matching');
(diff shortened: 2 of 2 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is in `invite-match.mjs`, which is the maintainer's routing call; I am reviewing the owned test `tests/invite-match.test.mjs`.

It is sound. It imports the real `extractCompany`, which loops over the `COMPANY_LINE_PATTERNS` regexes the fix changes from `[A-Z][\w...]` to the Unicode-property classes (`\p{Lu}\p{Lo}`, `\p{L}\p{M}\p{N}`), so it exercises the patched path, not a stub. The fixtures are concrete non-ASCII literals with hardcoded expecteds (`Nestlé`, `München Analytics`, `Яндекс`, `株式会社日立製作所`), and the expected is a literal rather than the function's own output, so it is not a tautology. It reddens on the old ASCII-only class: `é` breaks the lazy match's terminator and `[A-Z]` refuses the Cyrillic and CJK first letters, so each would return null and fail. The end-to-end guard flows through the same patched path (`analyzeInvite` yields `signals.company === 'Nestlé'` and matches the tracker row), covering the real user-visible failure.

PR #3873, opened 2026-09-05
feat(gmail): parse LinkedIn and Indeed digests into titled leads

### LINKED ISSUE #3731: gmail ingest: LinkedIn digests and Indeed alerts land as untitled URL-only leads
Follow-up to #3649, which fixed one class of junk in `plugins/gmail/_helpers.mjs`. Working on that file surfaced a second gap, and per CONTRIBUTING I'm opening an issue first because this is a feature, not a bug fix.

## The gap

The generic path (`extractUrls().filter(isCleanUrl)`) treats every alert email the same: it harvests URLs and files each one as `Job lead (email)` with no title, no company, no location. For the two highest-volume alert sources that is the wrong shape:

**LinkedIn job digests.** These are roughly 90% navigation and tracking chrome — `/comm/feed`, `/comm/premium`, `jobs/search-results`, promos, `/in/` profiles, plus pixels on `licdn.com` — and each real job appears as 4+ tracker-wrapped variants of the same `/jobs/view/<id>`. So the ingest produces many rows per digest, most of them not jobs, and the ones that are jobs are duplicated and untitled. Meanwhile the plain-text part of the same email lays out title / company / location cleanly, one block per job.

**Boards that bot-wall direct scraping.** Indeed and Glassdoor return 403, 104.com.tw and 1111.com.tw return 400. For those, the alert email is the only ingest path there is — but every link in it is a click-tracker redirect, so the raw URL is neither stable nor deduplicable.

## What I've been running

I patched this locally on 2026-07-18 and it has been in daily use since. Measured over 47 days on one install:

| | rows | outcome |
|---|---|---|
| LinkedIn `/jobs/view/` | 223 | **189 (84%)** carry a real title, company and location instead of `Job lead (email)` |
| Indeed | 17 | rebuilt to canonical `viewjob?jk=<id>`, deduped |

The approach is deliberately narrow:

- **`parseLinkedInJobs(plainText)`** — reads the text/plain part. The three core lines always sit directly above a blank separator, while the variable junk (an "actively hiring" proof line, an "Apply with resume" CTA, a skill tag) sits below it. So it walks up past whatever junk to the blank, then reads the three lines above. That avoids enumerating every junk variant LinkedIn ships, which was the part that kept breaking when I tried to match forward.
- **`extractLinkedInJobUrls(body)`** — bare-URL fallback for jobs the parser missed. Keeps only the numeric id, rebuilds a canonical tracking-free URL, dedupes by id.
- **`extractBoardJobUrls(body)`** — same idea for the walled boards: ignore the wrapper, pull the stable job id, rebuild the canonical link.
- **`isCleanUrl()`** then drops raw `linkedin.com` / `licdn.com` and raw board URLs on the generic path, since the real jobs now come from the functions above and the rest is chrome.

## What I'd want a maintainer to weigh in on before I send a PR

1. **Does this belong in the bundled `gmail` plugin, or in a plugin of its own?** It is per-sender parsing logic, and the bundled plugin is otherwise sender-agnostic. If you'd rather keep `_helpers.mjs` generic, a sender-dispatch shape (a small table of `{ matches(headers), parse(body) }`) would be a cleaner place to grow this, and I'm happy to build it that way instead.

2. **Only two of the four board patterns are evidence-backed.** Indeed has 17 real rows behind it. Glassdoor, 104 and 1111 were written from each board's documented URL shape *before* I had a real alert email in hand, and none has matched anything since — so I'd rather ship Indeed alone than ship three untested regexes. Would you prefer them dropped, or kept behind a comment saying they are unverified?

3. **Plain-text layout is a moving target.** `parseLinkedInJobs` reads a visual layout, and LinkedIn can change it. It fails closed (no clean separator → fall through to the URL fallback, which still yields a deduped canonical link), so a layout change degrades to today's behaviour rather than breaking. Is that failure mode acceptable to you, or would you want a canary test that flags when the enrichment rate drops?

Happy to send this as one PR or split it (LinkedIn first, Indeed second) — whichever you'd rather review. Tests would follow the section-74 pattern from #3649: fixtures for the layout, a guard for the fallback path, and a regression guard on `isCleanUrl`.

Felix-Lai-12:
Closes #3731.

A job-alert digest states the title, company and location of every posting in its
own body. The generic `extractUrls().filter(isCleanUrl)` path sees only links, so
each of those postings reached the pipeline as `Job lead (email)` with an empty
company — unrankable, undedupable against the same posting found by a scanner,
and unreadable without opening it. On the mailbox this was written against that
was 189 of 223 LinkedIn leads, and every Indeed lead.

Adds a per-sender parser layer in `plugins/gmail/_helpers.mjs`, dispatched on the
domain Gmail's own DMARC check verified, plus quoted-printable decoding shared by
the parser and the existing sweep. The commit message has the reasoning for each
piece; the two constraints from the issue thread are held to:

- **Fixtures, not a live mailbox.** `tests/fixtures/` carries one real LinkedIn
  digest and one real Indeed alert (a Traditional Chinese one, which is why the
  quoted-printable path is covered) with every account-identifying token
  replaced. 64 assertions, including an end-to-end pass through `ingest()`.
- **Nothing inferred.** A field the digest does not state stays `''`, and a block
  is read only when its shape is exactly the one the parser knows — bounded by a
  blank line or the start of the message, containing no second anchor. Anything
  else falls through to today's URL path and arrives untitled, which is the
  intended degradation rather than a gap.

### Security fix folded in

`isAuthenticEmail` accepted *any* `Authentication-Results` header. RFC 8601 §5
only obliges a receiver to strip headers claiming its **own** authserv-id, so a
header the sender wrote under any other id — `Authentication-Results: evil.test;
dmarc=pass header.from=linkedin.com` — is delivered untouched, and was enough to
pass the anti-spoof gate whenever Gmail itself reported no `dmarc=pass` (a
failure, or a domain with no DMARC record). This predates the PR; it became
load-bearing here, so it is fixed here rather than built on. Both the gate and
the new dispatch now read only the header whose authserv-id is `mx.google.com`,
and read the `dmarc=` clause whole.

One visible behaviour change comes with it: a digest that reaches the mailbox
through an ARC-signed forward carries Gmail's own `dmarc=fail` next to an
`arc=pass` whose comment quotes the forwarder's `dmarc=pass`. The old substring
search accepted that; it is now treated as spoofed and skipped. Reading the
`arc=pass` clause — also Gmail's own verdict — would restore the case, and is
left out until someone with such a mailbox can test it. Happy to add it here if
you'd rather it not regress.

### Deliberately not included

Glassdoor, 104 and 1111. Patterns for them were written from each board's
documented URL shape without a real alert on hand; shipping a parser nobody has
run against real mail would break the second constraint.

### Known gaps, if you want them in scope

- Dedupe is first-writer-wins within a run; the same posting in two messages
  keeps whichever body was read first, not the richer one.
- `canonicalizeJobUrl` does not handle LinkedIn's slug-form URLs, and truncates a
  `jk` longer than 20 hex characters.
- A three-part Indeed company line (`Acme - Taipei - Hybrid`) splits on the last
  separator, so `Taipei` lands in the company field.
- Indeed's label line (`Promoted`, an estimated-salary row) is handled by
  skipping the block, not by reading past it.
- The false-negative rate is unmeasured: I can say what fraction of leads arrive
  titled on one mailbox, not what fraction of postings the parsers skip.

## Summary

Gmail now converts verified LinkedIn and Indeed alerts into structured, deduplicated job leads.

### User-visible changes

- LinkedIn leads include title, company, location, and canonical URL.
- Indeed tracking links become canonical `viewjob?jk=` URLs.
- Navigation, footer, tracking, and non-job links are filtered.
- Unknown layouts fall back to generic URL processing.
- Parser dispatch uses Gmail’s verified DMARC domain, not the `From` header (`plugins/gmail/_helpers.mjs:511-554`).
- Messages with Gmail-authored `dmarc=fail` or forged authentication data are skipped (`plugins/gmail/_helpers.mjs:171-218`).
- ARC-forwarded messages with Gmail’s own DMARC failure may no longer be ingested.

### Files changed

- `plugins/gmail/_helpers.mjs:49-554`: URL filtering, authentication checks, quoted-printable decoding, canonicalization, and digest parsers.
- `plugins/gmail/index.mjs:105-145`: parser-first ingestion and fallback deduplication.
- `tests/fixtures/linkedin-job-alert.txt`: LinkedIn alert fixture.
- `tests/fixtures/indeed-job-alert.txt`: Indeed alert fixture.
- `tests/gmail-digest-parsers.test.mjs:400-470`: end-to-end parser and deduplication coverage.

No files in `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, or `.github/` changed.

### DIFF
--- a/plugins/gmail/_helpers.mjs
+++ b/plugins/gmail/_helpers.mjs
@@ -33,6 +33,19 @@ const ASSET_HOST_PREFIXES = ['cdn.', 'static.', 'assets.', 'img.', 'images.', 'm
  * @param {string} url
  * @returns {boolean}
  */
+/**
+ * Is `host` this domain or a subdomain of it? Compared label-wise, because a
+ * substring test reads `evillinkedin.com` as LinkedIn — and for a rewrite that
+ * is worse than a missed match: the foreign URL is replaced by a plausible board
+ * URL and takes the real posting's place in the dedupe key.
+ * @param {string} host lowercased hostname
+ * @param {string} domain registrable domain, e.g. 'linkedin.com'
+ * @returns {boolean}
+ */
+function isHostOf(host, domain) {
+  return host === domain || host.endsWith(`.${domain}`);
+}
+
 export function isCleanUrl(url) {
   try {
     const u = new URL(url);
@@ -42,6 +55,7 @@ export function isCleanUrl(url) {
       'newsletter', 'subscribe', 'w3.org', 'doubleclick', 'googlesyndication',
       'googleadservices', 'mailgun', 'mandrill', 'mjml', 'github.com/login',
       'linkedin.com/legal', 'linkedin.com/help', 'linkedin.com/settings',
+      'linkedin.com/comm/psettings', 'linkedin.com/comm/jobs/search',
     ];
     if (badKeywords.some(kw => lowerUrl.includes(kw))) return false;
     // Page assets, not postings. Job-alert emails embed one company logo per job,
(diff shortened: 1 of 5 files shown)


Scott-Emberson (commented):
Test-axis only, plus a routing flag. The substance is in `plugins/gmail/`, which CODEOWNERS puts with the maintainer (our ownership is `/tests/`), so this is a routing call for @santifer before our review counts; I am reviewing the owned test `tests/gmail-digest-parsers.test.mjs`.

Routing: this is a `feat(...)` against a bundled plugin, and CONTRIBUTING.md routes bundled plugins as stable reference seeds that take only security and compatibility fixes, with feature work published as a separate `career-ops-plugin-<id>`. There is real tension here, because part of the diff is genuine security hardening (DMARC authserv-id gating so a sender cannot forge `Authentication-Results`, plus the quoted-printable ghost-URL fix) that would qualify under the security-fix carve-out, while the digest-parsing capability is a feature. So the plugin-vs-core decision is the maintainer's: accept the security slice into the seed, or send the parser feature to a published plugin. I am not surfacing this as ready-to-approve.

The owned test is sound on its own merits. It runs the parsers against redacted real fixtures with concrete, non-tautological assertions: exact title/company/location per LinkedIn posting, the correct count of three distinct leads with the footer "View job" entry skipped, canonical `/jobs/view/<id>` URLs with tracking stripped, and CJK company extraction through quoted-printable decode on the Indeed side. It has real regression teeth: a field the digest omits stays empty rather than inheriting the neighbour's value, and four-line and shared-separator blocks are asserted to be skipped rather than mislabeled, with an end-to-end `ingest()` pass pinning dedup ordering and the no-ghost-URL result. On untrusted-content safety it is strong: the parsers are pure string and regex over the decoded body with no eval or dynamic import, dispatch keys on Gmail's own DMARC-verified `header.from` rather than the attacker-chosen From, and the adversarial surface (forged Authentication-Results, From-header spoofing, lookalike domains, the ghost URL) is explicitly asserted to fail closed.

So the owned test verifies sound, but hold on the routing flag: the plugin-vs-core call is the maintainer's, and the branch is behind main and needs a rebase regardless.

Scott-Emberson (commented):
Re-verified the owned test at the current head. It changed since my last look, but only to strengthen: the host checks were tightened from substring to exact match via a new `hostOf()` helper (removing a substring-match weakness in the test itself), lookalike-host regression teeth were added in both directions (`evillinkedin.com`, `myindeed.com`, `linkedin.com.evil.test`, `linkedin.com/feed#jobs/view` are left untouched by `canonicalizeJobUrl`, while a genuine `uk2.indeed.com` still canonicalizes), footer-navigation paths are asserted never to become leads, and `isWellFormed` was swapped for a portable `wellFormed()` shim with the assertion unchanged. Nothing was weakened.

The untrusted-content safety holds on the new version: no eval, Function, child_process, or dynamic import; concrete assertions against the real redacted fixtures; and dispatch keys on the DMARC-verified domain via the `mx.google.com` authserv-id and the whole `header.from` clause, failing closed on `dmarc=fail`, on `dmarc=pass` with no `header.from`, and on the foreign-authserv injection, header-in-comment, semicolon-in-reason, and forged-clause-in-quoted-local-part cases.

The owned test verifies sound. My routing flag from last time still stands, though: this is a `feat` against the bundled plugin `plugins/gmail` with a load-bearing DMARC hardening riding along, so the plugin-vs-core decision (accept the security slice into the seed, or send the parser feature to a published plugin) is the maintainer's, and the branch is behind main and needs a rebase regardless.

PR #3874, opened 2026-09-05
fix(workday): tag a split whose chosen facet under-covers the board

bigguy6883:
This is the second half of your review on #3311 — the finding about the split keeping its own honesty promise. #3311 merged with the page-0 containment landed and this one still open, so it comes back on its own.

It also closes CodeRabbit's remaining actionable on `providers/workday.mjs`, which asked for exactly this comparison plus a regression proving an under-covering facet reports `workdayTruncated`.

## The gap

The clamp is detected against the **largest** facet sum, but the split runs on whichever facet partitions most finely. Postings outside the chosen facet's values are never requested by any slice. So a facet that covers materially less than the board can finish every slice cleanly, leave the board short, and get logged as recovered with no `(still incomplete)` — the exact failure the recovery path exists to avoid.

## Why the obvious one-liner doesn't work

The bare `chosen < trueTotal` you suggested, and my own first counter-proposal, both fail — and it's worth writing down why so it isn't re-derived later.

`trueTotalFromFacets()` returns the **maximum** facet coverage, and the chosen facet is one of the counted facets. So for chosen coverage `c`:

```
gap    = trueTotal - c = max - c
spread = max - min
c >= min                    (c is one of the counted facets)
=>  gap <= spread,  always
```

Two consequences:

- **A bare comparison fires on noise.** Real facets disagree by a point or two — a posting missing a facet value is absent from that facet's counts, which is what your own comment above `trueTotalFromFacets` predicts. The chosen facet therefore sits just under the max on essentially every tenant. `workdayTruncated` would come to mean "some facet did not sum to the maximum" rather than "coverage was lost" — the tag-that-says-nothing case `'cap'` already had to avoid.
- **A spread bar computed over *all* counted facets is unsatisfiable.** `gap > spread` can never be true. Worse, it inverts: a genuinely under-covering facet *is* the minimum, so it sets `spread = max - c`, exactly the gap it is being judged against. The rule goes quietest precisely where it should fire hardest.

## The rule that works

Compute the spread over the **other** counted facets, excluding the chosen one. The bar then measures ordinary facet disagreement — which is what it was always meant to measure — and the chosen facet is judged against it instead of contributing to it.

## Evidence

Live boards, page 0, measured 2026-08-28:

| board | trueTotal | chosen facet | gap | spread (others) | tags? |
|---|---|---|---|---|---|
| `dickssportinggoods\|wd1\|dsg` | 8367 | `jobFamily` 8366 | 1 | 77 | no — noise, correct |
| `cvshealth` | 19436 | `timeType` 19434 | 2 | 2 | no — noise, correct |
| synthetic under-coverer | 2700 | `tier` 600 | 2100 | 0 | **yes** |

Three tests: two positive assertions that an under-covering split tags `workdayTruncated` and says `(still incomplete)`, plus the discriminator — the DSG-shaped 1-posting gap that must *not* tag.

Mutation-tested against today's `main`: replacing the spread bar with the bare comparison flips **exactly one** test, the discriminator, and nothing else. That is the whole behavioural difference between the two shapes, and it lands on the flagship tenant this feature was built for.

## If you'd still rather have the bare comparison

It's one line — drop the `spread` computation and compare `trueTotal - chosenCoverage > 0`. I've built it that way and it's green too, minus the discriminator. I went with the materiality bar because I don't think a tag that fires on every healthy board is one the sweep can act on, but it's your invariant and I'll switch it on request.

## Not in this PR

`locationMainGroup` is a live 2x trap in the same function — filed as #3875 rather than bundled here, since it is latent rather than a bug today.

## Summary

Workday facet splitting now detects material under-coverage.

When the selected facet covers only part of the board, the provider sets `workdayTruncated` and reports “(still incomplete)”. Small facet-count differences do not trigger this result.

### Files changed

- `providers/workday.mjs: coverage validation`
- `tests/providers/workday-facet-split.test.mjs: regression tests`

No changes affect `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, or `.github/`.

### DIFF
--- a/providers/workday.mjs
+++ b/providers/workday.mjs
@@ -555,6 +555,37 @@ export default {
         const facet = chooseSplitFacet(result.facets, { exclude: excluded });
         if (!facet) { splitIncomplete = true; return; }
 
+        // The clamp is detected against the LARGEST facet sum, but the split
+        // runs on whichever facet partitions most finely. Postings outside the
+        // chosen facet's values are never requested by any slice, so a facet
+        // that covers materially less than the board can finish every slice
+        // cleanly and still leave the board short — reported recovered, which
+        // is the failure this path exists to avoid.
+        //
+        // Materiality matters here, and the bar comes from the response. Real
+        // facets disagree by a point or two (a posting missing a facet value is
+        // absent from that facet's counts), so the chosen facet sits just under
+        // the max on essentially every board — DSG: trueTotal 8367, chosen
+        // jobFamily 8366. A bare `chosen < trueTotal` would tag every one of
+        // them, the tag-that-says-nothing case 'cap' already had to avoid above.
+        // The spread across the OTHER counted facets measures that ordinary
+        // disagreement (77 on DSG, 2 on cvshealth); a gap wider than it is real
+        // undercoverage. The chosen facet is excluded from the spread because a
+        // badly under-covering facet is itself the minimum, and leaving it in
+        // would inflate the bar to exactly the gap it should be judged against.
+        const chosenCoverage = facet.values.reduce((sum, v) => sum + v.count, 0);
+        const trueTotal = trueTotalFromFacets(result.facets);
+        if (trueTotal !== null) {
+          const others = [];
+          for (const f of Array.isArray(result.facets) ? result.facets : []) {
+            if (f?.facetParameter === facet.facetParameter) continue;
+            const coverage = facetCoverage(f);
+            if (coverage !== null) others.push(coverage);
+          }
+          const spread = others.length > 0 ? Math.max(...others) - Math.min(...others) : 0;
+          if (trueTotal - chosenCoverage > spread) splitIncomplete = true;
+        }
+
         for (const value of facet.values) {
           if (slicesSpent >= MAX_SPLIT_SLICES) { splitIncomplete = true; break; }
           if (pagesSpent >= pageBudget) { splitIncomplete = true; break; }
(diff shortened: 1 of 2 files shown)


Scott-Emberson (commented):
Verified the owned test `tests/providers/workday-facet-split.test.mjs`. The change tags a facet split whose chosen facet under-covers the board, and the test pins it concretely. The undercovering case picks the finest partition (`tier`, largest slice 300 below jobFamily's 1500) covering only 600 of a 2700 board, and asserts both `workdayTruncated === true` and that the clamp line includes "(still incomplete)"; since both slices complete cleanly, the only source of that truncation flag is the new under-coverage detector (`trueTotal - chosenCoverage > spread`), so removing the provider block makes both assertions fail. The over-tag guard is pinned too: a jobFamily split covering 2366 of 2367 (one short against a 77-wide spread) asserts `workdayTruncated === undefined`, so a naive `chosen < trueTotal` would fire but the spread-based materiality bar correctly suppresses it, which is what keeps the tag from degrading into noise.

The diff to the test file is additions only, so the #3851 config-location-priority assertions are intact, and the provider change is facet arithmetic only (reads `facet.values[].count`, sums, sets a boolean) with no URL or host construction, so no SSRF surface.

Owned test verified sound, CI green, mergeable and not behind. This is GATE-tier and clean. Holding formal approval until Scott gives the go-ahead; surfacing it as ready-to-approve.

PR #3898, opened 2026-09-05
Fix/sync pdf flags prune 3893

### LINKED ISSUE #3893: bug(sync-pdf-flags): a PDF flag can never be cleared and manifest rows outlive the files they name, so a correction does not stick
### What happened?

Two defects that compound, and neither is fixable alone:

1. **`sync-pdf-flags.mjs` is monotonic.** It sets the tracker's PDF cell to ✅ when the manifest names a PDF, and has no path that ever takes it back.
2. **`data/pdf-index.tsv` rows are never reconciled against the filesystem,** so a row keeps naming a PDF after the file is deleted.

Together, a deleted artifact leaves the tracker permanently asserting that a tailored CV exists for that application. That assertion is load-bearing — it is what the apply flow and `outcome.mjs` consult.

Fixing one without the other does not stick: clearing the tracker flag while the stale manifest row remains just gets it re-flagged on the next run, so a user who notices the wrong ✅ and corrects it watches the correction get reverted.

Verified against `main`: `sync-pdf-flags.mjs` has no clear, unset, downgrade or prune path.

### Relationship to #3887

**Not a duplicate.** #3887 is about the manifest *key* — one row per report number, so a cover letter evicts the CV row. This issue is about the flag being write-only and rows outliving the files they name. Different defects on the same file.

There is a sequencing dependency worth flagging, though. The natural way to express "this ✅ is stale" is "there is no CV-kind row for this report", which is not sayable until #3887 introduces a way to distinguish artifact kinds. **So this issue is scoped to the kind-agnostic half only** — pruning rows whose PDF is not on disk, which is true regardless of what kind of artifact the row names. The flag-downgrade half is better filed as a follow-on once #3887 lands.

### Steps to reproduce

1. Generate a CV for report N. Tracker cell → ✅, manifest row written.
2. Delete the PDF from `output/`.
3. `node sync-pdf-flags.mjs`.

The tracker still shows ✅ and the manifest still names the missing file.

### Expected behavior

A prune mode that drops manifest rows whose PDF is not on disk — kind-agnostic, since a cover letter whose PDF is gone is exactly as stale as a CV's. Dry-run by default seems the right shape given it edits the manifest.

Shaamak:
### Summary
Fixes #3893 by adding a `--prune` mode to `sync-pdf-flags.mjs` to reconcile `data/pdf-index.tsv` against the filesystem.

- **Reconciles missing files**: Drops manifest rows pointing to PDFs that no longer exist on disk in `output/`.
- **Dry-run by default**: `node sync-pdf-flags.mjs --prune` previews stale rows; `--prune --write` updates `data/pdf-index.tsv`.
- **Kind-agnostic**: Works across all PDF entry types as scoped in #3893.
- **Unit tests**: Added test suite covering `--prune` and `--prune --write` modes.

### Testing
- `node sync-pdf-flags.mjs --prune` → Previews removed stale rows without writing.
- `node sync-pdf-flags.mjs --prune --write` → Safely updates manifest file.
- `node --test tests/sync-pdf-flags.test.mjs` (16/16 Passed)

## Summary

- Added `--prune` to `sync-pdf-flags.mjs`: reports manifest rows for missing or out-of-workspace PDFs without changing files by default.
- Use `--prune --write` to remove stale rows. `--dry-run` takes precedence over `--write`.
- Added validation for absolute paths and `..` traversal.
- Added `update-system.mjs status` to report the installed version and short Git SHA.
- Documented the new command in `docs/SCRIPTS.md`.
- Added regression tests for prune behavior and nested Git installations.

System files touched: `update-system.mjs`. `AGENTS.md`, `modes/`, `DATA_CONTRACT.md`, `providers/`, and `.github/` were not changed.

### DIFF
--- a/docs/SCRIPTS.md
+++ b/docs/SCRIPTS.md
@@ -24,6 +24,7 @@ All scripts live in the project root as `.mjs` modules. Most are exposed via
 | `npm run update:check` | `update-system.mjs check` | Check for upstream updates |
 | `npm run update` | `update-system.mjs apply --confirm` | Apply upstream update |
 | `npm run rollback` | `update-system.mjs rollback` | Rollback last update |
+| `node update-system.mjs status` | `update-system.mjs status` | Print installed version + short SHA |
 | `npm run liveness` | `check-liveness.mjs` | Test if job URLs are still active |
 | `npm run extract` | `browser-extract.mjs` | Headless read-only page extractor (opt-in `scan.extractor: cli`) — compact JSON for scan/JD; Greenhouse, Lever, Ashby and Workday postings are read from their public JSON endpoints instead of the client-rendered page, and an empty jd extraction exits 1 with `code: empty_text` |
 | `node fetch-jd.mjs <url>` | `fetch-jd.mjs` | JD text on stdout from a known ATS API (Greenhouse/Lever/Ashby/Workday) — exit 1 with empty stdout when the host has no JD-bearing API, so a caller falls back to its browser/WebFetch path |
@@ -556,6 +557,32 @@ Possible JSON responses:
 | `update-available` | Newer version exists (includes `local`, `remote`, `changelog`) |
 | `dismissed` | User dismissed the update prompt |
 | `offline` | Could not reach GitHub |
+
+The `local` field in the JSON output is the formatted installed version, including the short commit SHA when the install is a git checkout — for example `"1.32.0 (ae919b6f)"`. For tarball installs without git metadata it is just the version string (`"1.32.0"`). This lets a bug report identify the exact tree under test, not just the release name (two installs pulled days apart can share a version string while running different code — see #3203).
+
+**Exit codes:** `0` always.
+
+---
+
+## status
+
+Prints the installed version to stdout — a quick human-readable alternative to parsing `check` JSON.
+
+```bash
+node update-system.mjs status
+```
+
+Example output:
+
+```
+career-ops v1.32.0 (ae919b6f)
+```
+
+On a tarball install with no git metadata the short SHA is omitted:
+
+```
+career-ops v1.32.0
+```
 
 **Exit codes:** `0` always.
 
--- a/sync-pdf-flags.mjs
+++ b/sync-pdf-flags.mjs
@@ -9,14 +9,21 @@
  *
  * Runs under the shared tracker lock and replaces the file atomically.
  *
+ * Prune mode (--prune) drops manifest rows whose PDF file is no longer on disk.
+ * A deleted artifact otherwise keeps the manifest row alive, which causes the next
+ * normal sync to re-assert ✅ on the tracker even though no file backs it. Prune
+ * is kind-agnostic: it drops any row (CV or cover letter) whose pdf column points
+ * at a missing file. It is a dry run by default — add --write to commit changes.
+ *
  * Usage:
  *   node sync-pdf-flags.mjs [--dry-run] [--json]
+ *   node sync-pdf-flags.mjs --prune [--write] [--json]
  */
 
 import { readFileSync, existsSync } from 'fs';
-import { join } from 'path';
+import { join, resolve, dirname } from 'path';
 import { extractTrackerReportNumbers, resolveColumns, parseTrackerRow } from './tracker-parse.mjs';
-import { rebuildRow, resolveTrackerPath, resolvePdfIndexPath, openTrackerTransaction } from './tracker-utils.mjs';
+import { rebuildRow, resolveTrackerPath, resolvePdfIndexPath, openTrackerTransaction, writeFileAtomic, resolveWorkspaceRoot, pathIsInsideCanonical } from './tracker-utils.mjs';
 import { getCareerOpsRoot } from './path-resolver.mjs';
 
 const DATA_ROOT = getCareerOpsRoot();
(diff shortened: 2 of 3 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is `sync-pdf-flags.mjs` plus a red-line `update-system.mjs` touch, which is the maintainer's routing call; I am reviewing the owned tests.

`tests/sync-pdf-flags.test.mjs` pins the prune and sync behavior with content-level assertions, not exit-code-only checks: the dry-run case asserts `pruned===1, kept===1, dryRun===true` and that the manifest is byte-for-byte unchanged (reddens if dry-run ever writes); the `--write` case asserts the stale row is gone while the live row and the header are kept (reddens if prune removes the wrong row, keeps the stale one, or drops the header); the all-present no-op, the `--write` without `--prune` flag-scoping rejection, and the no-manifest no-op are all pinned. `tests/updater-version-sha.test.mjs` likewise pins the exact `X (sha)` format and the bare-version no-git path. Both are non-tautological.

One structural flag for the maintainer: this PR is stacked on #3897, containing #3897's two commits verbatim (identical OIDs) plus the sync-pdf-flags prune commit on top, and the shared files are byte-identical between the two heads. So this is a superset of #3897, not a competing implementation. The clean path is to land #3897 first, then rebase this branch down to just the sync-pdf-flags commit; merging both independently risks a double-merge on the red-line `update-system.mjs` if #3897 is squash-merged (its OIDs would change).

Scott-Emberson (commented):
Test-axis only. The substance is now purely `sync-pdf-flags.mjs` and its doc, the maintainer's call; I am reviewing the owned test `tests/sync-pdf-flags.test.mjs`.

It is sound. The test spawns the real `sync-pdf-flags.mjs` CLI over per-test fixture manifests (via CAREER_OPS_TRACKER / CAREER_OPS_PDF_INDEX redirection), not a hand-reconstructed pruner, and its assertions are content-level rather than exit-code-only: parsed pruned/kept/dryRun counts plus direct manifest checks (the stale row removed, the live row and the comment header kept, out-of-bounds paths pruned via the inside-canonical guard). It passes (20 assertions via the real harness). I mutation-checked it: forcing missing PDFs to never prune reddens four assertions, and dropping the write guard so a dry run writes reddens the two dry-run-does-not-mutate assertions.

One useful update for the record: this PR is no longer stacked on #3897. It now carries a single commit touching only sync-pdf-flags.mjs, its test, and docs/SCRIPTS.md, and does not touch update-system.mjs, so the earlier double-merge concern on that red-line file is gone. The only remaining merge-state item is that the branch is behind main.

Owned test sound. SIGNAL (the substance is a standalone core script, out of our owned paths), so not surfacing as ready-to-approve; routing and merge are the maintainer's.

rubicon (changes requested):
Thanks for this, and the prune mode is the right shape for the half of #3893 I scoped it to. I reproduced the original bug on main and confirmed your branch fixes it: after `--prune --write`, a hand-corrected tracker flag survives the next sync instead of being reverted from the stale row. That part works.

There is a data-loss path I need to flag before this lands.

`sync-pdf-flags.mjs:129` treats `existsSync(absPath) === false` as proof the PDF was deleted. It is not. `existsSync` returns false for any stat failure, not just ENOENT. If `output/` is unreadable or simply not present, every row looks deleted at once.

```
# all four artifacts intact on disk, output/ at mode 000
$ node sync-pdf-flags.mjs --prune --write --json
{ "pruned": 2, "kept": 0, "dryRun": false }
$ echo $?
0
$ cat data/pdf-index.tsv
# report	pdf	html
```

The manifest is reduced to its header, the exit code is 0, and nothing warns. An unmounted external volume, or a fresh clone where `output/` has not been created yet, produces the same result.

The containment check above it has the same inversion. `pathIsInsideCanonical` is documented in `tracker-utils.mjs:229-242` as a fail-closed deny guard that returns false whenever canonicalization fails for any reason, and `outcome.mjs` uses it to refuse an operation. At `sync-pdf-flags.mjs:119` that same false becomes a reason to delete the record, so "I could not resolve this path" is treated as "this row is garbage."

Three things would close it:

1. Distinguish ENOENT from other stat errors. `statSync` in a try/catch, prune only on ENOENT, keep-and-warn on anything else.
2. Do the same on the containment check. A path you cannot canonicalize should be reported and kept, not dropped.
3. Add a blast-radius cap. Pruning every row in the manifest is currently indistinguishable from pruning one, and it is the case most likely to be an environment problem rather than real staleness.

Two smaller things.

Prune drops the whole row when only the `pdf` column is stale, which takes the `html` column with it. That column is the only per-report record of the source HTML, and `modes/pdf/hm-audit.md` step 4.2 falls back to a newest-file heuristic that the docs themselves call worse than asking. So the remedy for a deleted PDF removes the means of regenerating it.

Five of the new assertions pass against unfixed main, because main rejects `--prune` and `--write` as unknown options and exits 1 having written nothing, which makes "the manifest was not modified" trivially true. The affected ones are the dry-run no-write case, the comment-header preservation case, the nothing-pruned case, the `--dry-run --write` case, and `rejects --write outside of --prune`. That last one is the only coverage of your new guard at `:43-45`, so that branch is currently unproven. The other eight assertions do fail on unfixed source, so the suite as a whole is not vacuous.

Also worth checking: the `docs/SCRIPTS.md` hunk documents `update-system.mjs status`, which does not exist. The dispatch switch at `update-system.mjs:2756` has only check, apply, rollback and dismiss, and running `status` prints usage and exits 1. Meanwhile `--prune` itself is not documented anywhere. `sync-pdf-flags.mjs:10` still says the script runs under the shared tracker lock, which is no longer true in prune mode since that branch returns at `:169` before the transaction opens at `:202`. And `dirname` is imported at `:24` and never used.

One heads-up on the test at `tests/sync-pdf-flags.test.mjs:390-435`: it asserts that a row failing the containment check gets pruned, so fixing the guard the way I suggested means changing that test. I would rather the test move than the behavior stay.

Last thing, on verification rather than the code. The PR body cites `node --test tests/sync-pdf-flags.test.mjs` as 16/16 passing, but these are console.log helpers rather than node:test assertions, so that command exits 0 even when they fail. I had a run print eight failures and still exit 0. `node test-all.mjs` is the gate that actually aggregates them.

santifer:
Sorry for the wait, @Shaamak: your 7 Sep reply answered all three points and then sat here twelve days. On the current head `--dry-run` wins over `--write`, containment goes through `pathIsInsideCanonical` against the workspace root like generate-pdf does, and the branch is already down to one commit.

Only one thing still ties this to #3897: the `docs/SCRIPTS.md` hunk documenting `node update-system.mjs status` and the `local` field. That command isn't in main, so landing this first would publish docs for something nobody can run, and #3897 edits the same file. Drop the hunk and #3898 stops depending on #3897, which is worth doing since #3897 is conflicting right now and can't land as-is.

Two small ones while you're there. The module docblock still says the script runs under the shared tracker lock, but prune exits before any transaction opens. No lock is needed: `updatePDFManifest` in generate-pdf.mjs is the manifest's only writer and uses a plain `writeFileSync`, so it's the sentence to fix, not the code. And `--dry-run --write` without `--prune` now swallows `--write` silently while `--write` alone still exits 1; moving the precedence line below the unknown-option check restores what the comment there promises.

CodeRabbit's note about the traversal test not covering an existing escaped file is stale: your test does write the external PDF. CI is green on `632ad11c`. Once the docs hunk is gone, updating the branch and merging is on us.

Shaamak:
All done @santifer! I’ve updated the branch and resolved the remaining items:

1. **Docs decoupled:** Dropped the `docs/SCRIPTS.md` hunk entirely, so this no longer depends on #3897 to land first. 
2. **Docblock fixed:** Clarified the lock comment in `sync-pdf-flags.mjs` to specify that it only applies to the normal sync, not `--prune`.
3. **Flag precedence fixed:** Moved the `--write` precedence check so it properly surfaces an error if someone passes `--write` by itself outside of `--prune` mode.

The branch is clean, up to date, and ready whenever you are to pull the trigger. Thanks for the thorough reviews!

PR #3901, opened 2026-09-05
fix(verify-cv-facts): tell a disabled fact gate apart from one that ran clean

### LINKED ISSUE #3894: bug(verify-cv-facts): a missing config/cv-facts.json disables the fact gate silently and reads as a clean pass
### What happened?

`loadConfig()` in `verify-cv-facts.mjs` returns `{allow_metrics: [], allow_facts: [], forbidden_phrases: [], warn_phrases: []}` when the config file does not exist.

That is a reasonable default and a terrible silent one. An empty `forbidden_phrases` is the only phrase-level guard the module has, so a gate that is **disabled** is indistinguishable from a gate that **ran and found nothing**. Both print the same clean result, and the clean result is the one a user reads as "my CV was checked".

The failure mode is not hypothetical. Downstream, this let a claim that had been deliberately quarantined ship in **57 rendered CVs** before anyone noticed — every render reported the fact gate as passing, because there was no config for it to fail against.

Verified against `main`: no `loadFactConfig` or equivalent missing-config signal anywhere in the tree.

### Steps to reproduce

1. Delete (or never create) `config/cv-facts.json`.
2. Render a CV containing a phrase that would be listed in `forbidden_phrases`.
3. The gate reports clean.

### Expected behavior

Callers can tell "no config" from "config found nothing". A minimal shape: export `loadFactConfig(path)` returning `{missing: boolean, config}`, and have the render paths surface it — a single warning line is enough.

Deliberately **not** a hard failure by default: a user who has never written a `cv-facts.json` is in a normal state. They just should not be left believing a gate is protecting them when none is loaded.

rubicon:
## What does this PR do?

`loadConfig()` returned the empty shape when `config/cv-facts.json` was absent, and `forbidden_phrases`/`warn_phrases` are the only phrase-level guard the module has — so a **disabled** gate printed exactly what a gate that **ran and found nothing** prints. `loadFactConfig()` now returns `{missing, config}`, `verifyFacts()` carries `configMissing` through to callers, and the three render paths (the CLI, `generate-pdf.mjs`, `generate-cover-letter.mjs`) each emit one warning line when it is true. Deliberately a warning, not a hard failure: a user who has never written a `cv-facts.json` is in a normal state, so verdicts and exit codes are unchanged.

## Related issue

Closes #3894

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

---

### How it was verified

`tests/fact-gate-missing-config.test.mjs` asserts the distinction that did not exist before this change. On unfixed code it reports **2 passed, 7 failed** — the two greens are the "must not change" assertions (verdict stays `pass`, CLI still exits 0), which is what keeps this a warning rather than a gate failure. With the fix: **11 passed, 0 failed**.

The suite is namespace-imported on purpose: a named import of a not-yet-existing export is a link-time `SyntaxError` that kills the whole file before a single assertion runs, reporting one opaque failure instead of the real set.

Mutation-checked — each mutation applied alone, then reverted, and the file confirmed byte-identical to the fixed version afterwards:

| Mutation | Result |
|---|---|
| `loadFactConfig` returns `missing: false` for an absent file (the original silent shape) | 🔴 5 failed — `loadFactConfig did not flag an absent config`, identical-results, `configMissing` on both sides, CLI output identical, `--json` |
| `configMissing` dropped from the `verifyFacts` return | 🔴 4 failed — identical-results, `configMissing` on both sides, CLI output identical, `--json` |
| CLI warning line removed | 🔴 1 failed — `the CLI printed the same output with and without a fact-gate config` |
| `generate-pdf.mjs` warning removed | 🔴 1 failed — `generate-pdf.mjs never surfaces configMissing` |
| `generate-cover-letter.mjs` warning removed | 🔴 1 failed — `generate-cover-letter.mjs never surfaces configMissing` |
| non-array key validation removed from `loadFactConfig` | 🔴 1 failed — `loadFactConfig accepted a non-array forbidden_phrases` (proves that assertion is not vacuous across the shape change) |

No mutation left the suite green.

Full `node test-all.mjs`: **8580 passed, 0 failed, 16 warnings**. All 16 warnings are pre-existing and name none of the files touched here (`cv-sync-check.mjs` without user data, the personal-data scanner on `funding.json`/`HIRED.md`/`hired-wall-build.mjs`/`dashboard/…/stats.go`/`.codex-plugin/plugin.json` and their tests, and two live-API suites skipped behind env flags).

The two document render paths are asserted at source level rather than by rendering: both gate calls sit immediately before a Playwright import, so exercising them would mean producing a real PDF to test one console line. Both source assertions are mutation-checked above.

## Summary

The CV fact gate now distinguishes a missing `config/cv-facts.json` from a configured gate that finds no issues.

### User impact

When `config/cv-facts.json` is missing:

- `verifyFacts()` exposes `configMissing`.
- The CLI, PDF generator, and cover-letter generator print one warning.
- Verdicts and exit codes do not change.
- JSON output remains parseable.
- Fact checks do not run, so a clean result does not imply that the CV was checked.

Malformed configurations still fail validation.

### Files changed

- `verify-cv-facts.mjs`: Exports `loadFactConfig(path)` with `{ missing, config }`. Adds `configMissing` to verification results and CLI warnings.
- `generate-pdf.mjs`: Warns when the fact configuration is missing.
- `generate-cover-letter.mjs`: Warns when the fact configuration is missing.
- `tests/fact-gate-missing-config.test.mjs`: Adds regression coverage for loader state, verification results, CLI output, document-generation warnings, verdicts, and exit codes.

### System files

### Validation

The new suite reports 11 passing assertions. The full suite reports 8,580 passed and 0 failed.

### DIFF
--- a/generate-cover-letter.mjs
+++ b/generate-cover-letter.mjs
@@ -325,6 +325,11 @@ Usage:
     // validator before importing Playwright or writing a PDF so a failed gate
     // cannot leave behind a misleading artifact.
     const factCheck = assertFacts(html, { label: "cover letter" });
+    // Ahead of the verdict, because it qualifies it: with no config the phrase
+    // lists are empty, so a silent gate here covers metrics and facts only.
+    if (factCheck.configMissing) {
+      console.error("No config/cv-facts.json — forbidden/advisory phrase checks did not run.");
+    }
     if (factCheck.verdict === "warn") {
       console.error(`CV fact check warning: cover letter`);
       for (const phrase of factCheck.warnings) {
--- a/generate-pdf.mjs
+++ b/generate-pdf.mjs
@@ -1354,6 +1354,11 @@ async function generatePDF() {
     // fails, which is the correct direction to fail for a fact gate.
     const { assertFacts } = await import('./verify-cv-facts.mjs');
     const factCheck = assertFacts(html, { label: basename(inputPath) });
+    // Ahead of the verdict, because it qualifies it: with no config the phrase
+    // lists are empty, so a "passed" below covers metrics and facts only.
+    if (factCheck.configMissing) {
+      console.warn('⚠️  No config/cv-facts.json — forbidden/advisory phrase checks did not run.');
+    }
     if (factCheck.verdict === 'warn') {
       console.warn(`⚠️  CV fact check warning: ${basename(inputPath)}`);
       for (const phrase of factCheck.warnings) console.warn(`  - advisory phrase: ${phrase}`);
--- a/tests/fact-gate-missing-config.test.mjs
+++ b/tests/fact-gate-missing-config.test.mjs
@@ -1,0 +1,142 @@
+// tests/fact-gate-missing-config.test.mjs — a fact gate with no config loaded
+// must not look like a fact gate that ran and found nothing.
+//
+// config/cv-facts.json is optional and absent from a fresh clone, and its
+// forbidden_phrases/warn_phrases are the only phrase-level guard the module
+// has. loadConfig() returned the empty shape for a missing file, so the two
+// states produced a byte-identical clean result and every render reported the
+// gate as passing — including the renders of a claim that had been
+// deliberately quarantined (#3894).
+//
+// Deliberately a warning, never a failure: a user who never wrote a
+// cv-facts.json is in a normal state. The assertions below pin the
+// distinction, not a verdict change.
+//
+// Run:  node test-all.mjs --only fact-gate-missing-config
+
+import { pass, fail, ROOT, NODE } from './helpers.mjs';
+import { mkdtempSync, writeFileSync, rmSync, readFileSync } from 'fs';
+import { spawnSync } from 'child_process';
+import { join } from 'path';
+import { tmpdir } from 'os';
+// Namespace import on purpose: a named import of a not-yet-existing export is a
+// SyntaxError at link time, which kills the whole suite before a single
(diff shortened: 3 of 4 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is `verify-cv-facts.mjs`, `generate-cover-letter.mjs`, and `generate-pdf.mjs` (the last red-line), which is the maintainer's routing call; I am reviewing the owned test `tests/fact-gate-missing-config.test.mjs`, and since this guards against unverified CV facts shipping I verified strictly.

It is sound. It pins the disabled-versus-clean distinction at three layers: `loadFactConfig` (`missing:true` for an absent config versus `missing:false` for present), `verifyFacts` (asserts `disabled.configMissing === true` and `ranClean.configMissing === false` and that the two result objects are non-identical), and the CLI, `--json`, and render paths. I mutation-checked it: flipping the fix to treat a missing config as a clean pass reddened five assertions (did not flag an absent config, returned identical results, did not surface `configMissing`, same CLI output, missing `--json` flag), so a disabled gate cannot masquerade as a passed fact-check. The flags it asserts on are real, and the config-mismatch assertions are safe-directional (a broken flag would fail, not false-pass). Correctly, a missing config stays a warning with the verdict `pass`, so this surfaces the disabled state without hard-failing.

No test-axis concern. The substance includes a red-line file, so routing and merge are the maintainer's call.

PR #3902, opened 2026-09-05
fix(tracker): strip invisible control characters at the cell sanitizer

### LINKED ISSUE #3892: bug(tracker): a control byte in a tracker cell is invisible in every view and corrupts every positional parse downstream
### What happened?

`data/applications.md` is parsed positionally on `|` boundaries. A control character pasted into a cell — most often from copying a job title out of a rendered posting — survives the write and is **invisible in every rendered view of the table**, including GitHub's and the web dashboard's.

Downstream it shifts or truncates the positional parse, so a row silently reads as a different row, or drops out of a count entirely. Nothing on the write path rejects it and nothing on the verify path detects it, so the corruption surfaces only as an unrelated arithmetic discrepancy much later, with no trail back to the cause.

The asymmetry is what makes it worth guarding: the byte is trivially cheap to reject at the write chokepoint, and very expensive to find once it is in the file, because every tool that displays the table hides it.

Verified against `main`: no control-character handling in `tracker-utils.mjs` or `tracker-parse.mjs`.

### Steps to reproduce

1. Write a tracker row whose Role cell contains a control byte in the ` `–`` range (excluding `\t` and `\n`) — pasting from a rendered posting is the realistic path.
2. View `data/applications.md` in any markdown renderer. The cell looks correct.
3. Run any positional consumer over the tracker.

### Expected behavior

Two halves, deliberately:

- **Strip (or reject) control characters at the tracker write path** — stops new ones entering.
- **Add a `verify-pipeline.mjs` check that fails on one already in the file** — finds the ones already written, which stripping alone cannot.

Same "invisible in every view, corrupts the parse" shape that `tests/source-no-nul-bytes.test.mjs` already guards for source files, applied to data.

rubicon:
## What does this PR do?

Strips invisible control characters at `cell()` in `tracker-utils.mjs` — the one sanitizer every tracker writer passes through — and adds `verify-pipeline.mjs` Check 16 for the bytes already sitting in a tracker file, which stripping alone cannot reach. Both halves share one exported `CONTROL_CHARS`, so the write path and the detector cannot drift apart about what counts.

## Related issue

Closes #3892

## Type of change

- [x] Bug fix
- [ ] New feature

## Why the sanitizer and not each writer

`cell()` is the chokepoint: `merge-tracker.mjs`'s `buildRow` runs company, role, location, notes and url through it — and the web dictates its rows through that same merge path — while `set-status.mjs` runs its note through it. A guard added per writer is a guard the next writer is free to reintroduce the bug around, so the strip lands once, where all of them already route.

The tests assert `cell()` directly, never a caller. A writer added later either inherits the guard or bypasses `cell()` visibly; it cannot quietly opt out of a caller-shaped test that was never written about it.

## What the class covers

C0, DEL and C1, minus the three whitespace controls. `\t` is ordinary whitespace inside a cell and `\r`/`\n` are folded to a single space by the replace that already runs ahead of this one — stripping any of the three would glue words together instead of separating them. Same range, same name, as the existing display sanitizer in `plugins/h1b-sponsor/`.

They are **deleted, not replaced with a space**. The byte renders as nothing in markdown, on GitHub and in the dashboard, so substituting a space would change text a human has already read.

## Check 16

Reads the raw lines rather than the parsed entries, so a byte in a column this file has no field for is caught too, and reports the file line number — a shifted parse is exactly the case where a row's own `#` cell is the thing not to trust. It names the code point (`U+0001`) so the byte can be found, and it `error`s rather than `warn`s: the exit code is what CI and cron wrappers read, and a finding that printed while exiting 0 would be indistinguishable from a clean tracker.

Columns that do not pass through `cell()` (`num`, `date`, `score`, `status`, `pdf`, `report`) are left as they are — their own existing checks already reject a malformed value, and Check 16 backstops every column regardless of how the byte arrived.

## Verification

TDD, both halves red first on unfixed code:

- `cell('Senior\x01 Engineer')` returned `'Senior\x01 Engineer'` unchanged.
- `verify-pipeline.mjs` on a tracker whose Role cell held `U+0001` printed `0 errors, 0 warnings` and exited 0 — the silent-corruption shape the issue describes.

Mutation-checked, each one reverted after:

| Mutation | Test that went red |
|---|---|
| Drop `.replace(CONTROL_CHARS, '')` from `cell()` | `cell strips invisible control characters (#3892)` |
| Narrow the class to `\x02-\x08` (drop NUL/SOH) | same |
| Drop `\x7F-\x9F` (DEL and C1) | same |
| Drop `\x0C` (the delete-don't-substitute case) | same |
| Widen the class to `\x00-\x1F` so it swallows `\t` | same |
| Check 16 `error()` → `warn()` | `a control byte ... is reported as an error and exits 1` |
| Drop the code points from the finding | `the finding names the code point` |
| Change the clean-run success line | `control: the same row without the byte reports clean` |

`node test-all.mjs` (full, not `--quick`): **8572 passed, 0 failed, 16 warnings**. None of the 16 warnings name a file in this PR — they are the pre-existing set (maintainer name/domain matches in `funding.json`/`HIRED.md`, missing-font probes, fixture merge notices, skipped opt-in live-API tests, portal config notes).

## Checklist

## Summary

`tracker-utils.mjs:cell()` removes invisible C0, DEL, and C1 control characters. Tabs remain valid. Newlines keep their existing folding behavior. Pipes remain sanitized.

`verify-pipeline.mjs` adds Check 16. It scans tracker lines and reports the line number and Unicode code point for each control character. The check fails verification.

## User impact

New tracker writes cannot add these control characters through `cell()`. Existing corrupted tracker files fail verification with a precise location and code point.

Tests cover sanitization, detection, error output, and clean tracker files. Full verification passed: 8,572 tests, 0 failures, and 16 pre-existing warnings.

## Files touched

- `tracker-utils.mjs`
- `verify-pipeline.mjs`
- `tests/tracker-utils.test.mjs`
- `tests/verify-pipeline-control-bytes.test.mjs`

The requested system files were not touched: `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/`.

### DIFF
--- a/tests/tracker-utils.test.mjs
+++ b/tests/tracker-utils.test.mjs
@@ -37,6 +37,31 @@ test('cell neutralizes table-breaking characters without changing ordinary text'
   assert.equal(cell('Senior Engineer'), 'Senior Engineer');
 });
 
+// Asserted on the sanitizer itself, never through a writer. cell() is the one
+// chokepoint every tracker writer passes through — merge-tracker's buildRow for
+// company/role/location/notes/url (and so the web, which dictates rows through
+// the same merge path), set-status for its note. Testing a caller instead would
+// leave the guard un-asserted the moment a new writer is added; testing the
+// chokepoint means a new writer inherits it or bypasses cell() visibly.
+test('cell strips invisible control characters (#3892)', () => {
+  // The realistic path: a role title copied out of a rendered posting carries a
+  // C0 byte. It is invisible in markdown, on GitHub and in the web dashboard,
+  // and it shifts or truncates the positional `split('|')` parse downstream.
+  assert.equal(cell('Senior\x01 Engineer'), 'Senior Engineer');
+  assert.equal(cell('Head\x7f of Growth'), 'Head of Growth');
+  assert.equal(cell('Zeta\x9dCorp'), 'ZetaCorp');
+  assert.equal(cell('\x00\x1b'), '');
+
+  // Deleted, not replaced with a space: the byte renders as nothing, so a
+  // replacement would change the text a human already sees.
+  assert.equal(cell('Data\x0cEngineer'), 'DataEngineer');
+
+  // Whitespace that legitimately separates words is left alone — a tab is
+  // ordinary whitespace in a cell, and a newline still folds to one space.
+  assert.equal(cell('Staff\tEngineer'), 'Staff\tEngineer');
+  assert.equal(cell('Acme\r\nRemote'), 'Acme Remote');
+});
+
 test('normalizeCompany preserves meaningful non-Latin company names', () => {
   assert.equal(normalizeCompany('Acme, Inc. (Remote)'), 'acmeincremote');
   assert.equal(normalizeCompany('株式会社ゼータ'), '株式会社ゼータ');
--- a/tests/verify-pipeline-control-bytes.test.mjs
+++ b/tests/verify-pipeline-control-bytes.test.mjs
@@ -1,0 +1,72 @@
+// tests/verify-pipeline-control-bytes.test.mjs — Check 16 (#3892), end to end
+// through the real verify-pipeline.mjs process.
+//
+// Stripping at the cell sanitizer stops NEW control bytes entering the tracker;
+// it can do nothing about the ones already written. This is the other half:
+// the only place a byte that is invisible in every rendered view of the table
+// becomes visible again. Asserted through the process because the exit code is
+// the part CI and cron wrappers read — a finding that printed but exited 0
+// would be indistinguishable from a clean tracker.
+//
+// Only the tracker and reports dir point at a fixture, the same way
+// tests/verify-pipeline-check15.test.mjs arranges them, so no user data is read.
+import { execFileSync } from 'child_process';
+import { mkdirSync, mkdtempSync, writeFileSync } from 'fs';
+import { tmpdir } from 'os';
+import { join } from 'path';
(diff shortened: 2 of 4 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is `tracker-utils.mjs` and `verify-pipeline.mjs`, which is the maintainer's routing call; I am reviewing the owned tests, and since this is tracker data-integrity I verified strictly.

Both are sound. `CONTROL_CHARS` is `/[\x00-\x08\x0B\x0C\x0E-\x1F\x7F-\x9F]/g`, deliberately excluding tab, LF, and CR. `tests/tracker-utils.test.mjs` pins the strip with concrete dirty-to-clean pairs (`Senior\x01 Engineer` to `Senior Engineer`, DEL, C1, `\x00\x1b` to empty), and removing the replace from `cell()` reddens every one; the over-strip guard is asserted rather than assumed (a tab-containing value must survive verbatim, so widening the range to grab tab reddens, and a form-feed is deleted rather than space-substituted), and Unicode and emoji sit above the range so it cannot touch them. `tests/verify-pipeline-control-bytes.test.mjs` runs the real process, reads stdout off the thrown error, and requires both a `/control character/i` finding and exit 1 for a dirty tracker (a finding that printed but exited 0 fails), plus the clean-control half requires exit 0 and the all-clear message and the absence of the finding string, ruling out a false negative. The shared `CONTROL_CHARS` export prevents write-path and detector drift.

PR #3903, opened 2026-09-05
refactor(lock): clamp the retry backoff inside the policy that owns it

### LINKED ISSUE #3895: refactor(lock): retry backoff is clamped at three call sites instead of inside the policy that owns it
### What happened?

The lock retry backoff is clamped by each caller rather than by the policy that owns it — three call sites each re-derive the same bound.

This is the standard duplicated-invariant shape: the copies are correct until one is edited, and **nothing fails when they diverge**. A wrong clamp just produces slightly different retry timing, which no test asserts and no user reports, so the drift is silent and permanent once it happens.

Verified against `main`: no shared clamp helper in `tracker-utils.mjs`, `pipeline-lock.mjs` or `followup-seed.mjs`.

### Steps to reproduce

Read the three call sites; each applies its own bound to the value the policy supplies.

### Expected behavior

Clamp inside the policy, so callers receive an already-valid interval and cannot individually get it wrong.

### Severity

Low, and filed as such. Nothing is currently broken — this is a small consolidation that removes a class of future drift rather than a defect anyone is hitting. Close it if it is not worth the churn; I would rather raise it than have the fourth call site copy the pattern.

rubicon:
## What does this PR do?

Moves the lock retry backoff's clamp inside `createLockWaitPolicy`, so callers receive an already-valid interval instead of each deriving the bound themselves. `hardDeadline` becomes optional: omit it and the policy applies its own `DEFAULT_MAX_WAIT_FACTOR x timeoutMs` ceiling. Three callers stop bounding, and two assertions are added that make a future divergence fail loudly.

## Related issue

Closes #3895

## Type of change

- [ ] Bug fix
- [ ] New feature

---

### Which three callers

The issue names `tracker-utils.mjs`, `pipeline-lock.mjs` and `followup-seed.mjs`. On `main` the modules that actually re-derive the bound are:

| Module | Line on `main` |
|---|---|
| `tracker-utils.mjs` | `hardDeadline: Date.now() + timeoutMs * 10` |
| `followup-seed.mjs` | `hardDeadline: Date.now() + timeoutMs * 10` |
| `portal-health-lock.mjs` | `hardDeadline: Date.now() + timeoutMs * 10` |

`portal-health-lock.mjs` adopted the shared policy in `39f87cd7` ("share the waiting half of the protocol with the three lock copies") and carries the same copy — it is the third of the "three lock copies" that commit names, so it is included here. Leaving it out would have left exactly one copy of the bound alive, which is the fourth-call-site outcome the issue was filed to prevent.

`acquirePipelineLock` still passes an explicit `hardDeadline`, and deliberately so: it has a real `maxWaitMs` option whose `0` ("never wait past now") and `Infinity` ("no ceiling") readings are load-bearing, and its factor constant already lives in the same module as the policy. It is a caller with a knob, not a copy of the default.

### Why a refactor alone would not have been a fix

The issue's own framing is the important half: **nothing currently fails when the copies diverge.** A wrong ceiling only changes retry timing, which no test asserts and no user reports, so the drift is silent and permanent.

That is not a claim I took on trust. On this branch, drifting one caller's copy from `timeoutMs * 10` to `timeoutMs * 2` leaves the entire 27-case `tracker-writer-lock-tests.mjs` suite green — 27 passed, 0 failed. Moving the clamp without adding an assertion would have relocated a silent invariant rather than removed one.

Two assertions close it:

1. **The policy refuses a ceiling that is not a number.** This failure mode is invisible by construction: `Math.min(x, NaN)` is `NaN`, `Math.max(0, NaN)` is `NaN`, and `setTimeout(NaN)` fires immediately — so a caller that got the ceiling wrong would spin hot through the retry loop instead of raising anything. `Infinity` is explicitly kept, because it is a deliberate "no ceiling" that `acquirePipelineLock` relies on; `NaN` and non-numbers are mistakes and are refused where they are made. `null` is treated as "not supplied" and takes the default, pinned by a test because `??` is what draws that line.

2. **A guard fails if any module other than the policy names `hardDeadline`.** That is precisely the shape a fourth call site copying the pattern would take. Comments are stripped before the check, so a module may explain why it does *not* pass one — a guard that fires on prose gets silenced rather than obeyed. The scan refuses to pass on an empty or implausibly small file list, so it cannot report success when it failed to look.

### Test evidence

`tests/pipeline-lock.test.mjs` — 4 new cases, 25 passing in that file.

Written test-first, and every assertion mutation-checked:

| Mutation | Result |
|---|---|
| Remove the default derivation (`ceiling = hardDeadline`) | RED — 3 of the 4 new cases; `backoffMs()` returns `NaN` and `ceilingReached()` never fires |
| Drift the factor to `timeoutMs * 5` | RED — the ceiling-value case (~1000ms observed, ~2000ms expected) |
| Disable the type guard | RED — the refusal case |
| Tighten the guard to `Number.isFinite`, rejecting `Infinity` | RED — the refusal case. Worth noting: the pre-existing `maxWaitMs` cases all stayed green, so nothing else covered that path |
| Re-add `hardDeadline: Date.now() + timeoutMs * 10` at a caller | RED — the drift guard only; `tracker-writer-lock-tests.mjs` stayed 27/27 green |
| Diverge a caller's copy to `timeoutMs * 2` | RED — the drift guard only; `tracker-writer-lock-tests.mjs` stayed 27/27 green |

The last two are the point of the PR: the existing suite cannot tell a diverged copy from a correct one, and now something can.

The ceiling value is asserted as a literal `2000ms` rather than by importing the constant — a test that reads the constant back out of the module it is testing cannot notice that constant moving.

## Checklist

- [x] I ran `node test-all.mjs` and all tests pass — 8568 passed, 0 failed, 16 warnings, none of which name any file this PR touches (they are the pre-existing template/fixture and font-probe warnings on `main`)

## Summary

Centralizes lock retry backoff limits in `createLockWaitPolicy`.

### User impact

- Lock callers no longer calculate their own retry ceiling.
- The policy derives the default ceiling from `DEFAULT_MAX_WAIT_FACTOR × timeoutMs`.
- `acquirePipelineLock` keeps explicit `maxWaitMs` behavior, including `0` and `Infinity`.
- Invalid non-number or `NaN` ceilings now raise `TypeError`.
- No user-facing behavior changes are expected beyond consistent retry timing.

### Files changed

- `pipeline-lock.mjs`: Adds optional `hardDeadline` handling and validation.
- `followup-seed.mjs`: Removes local ceiling calculation.
- `portal-health-lock.mjs`: Removes local ceiling calculation.
- `tracker-utils.mjs`: Removes local ceiling calculation.
- `tests/pipeline-lock.test.mjs`: Adds coverage for default ceilings, expiration, invalid values, and duplicated call-site calculations.

### Verification

- `tests/pipeline-lock.test.mjs`: 25 tests pass.
- Full test suite: 8,568 tests pass with no failures.

### DIFF
--- a/followup-seed.mjs
+++ b/followup-seed.mjs
@@ -326,10 +326,12 @@ async function acquireFollowupsLock(lockDir, followupsPath, options = {}) {
   // from the definition: waiters woke in lockstep and re-raced, and a caller
   // waiting on a healthy lock being handed round briskly was killed anyway.
   //
-  // There is no separate maxWaitMs knob here, so the ceiling is the same
-  // multiple of timeoutMs the definition defaults to.
+  // There is no separate maxWaitMs knob here, so no hardDeadline is passed and
+  // the policy applies its own ceiling. Writing one out here would put a fourth
+  // copy of that bound in the tree, and a copy that drifts changes retry timing
+  // silently — nothing fails, so nothing reports it (#3895).
   const { backoffMs, holderStillWedged, noteWaiting, ceilingReached } = createLockWaitPolicy(lockDir, {
-    timeoutMs, retryMs, deadline: Date.now() + timeoutMs, hardDeadline: Date.now() + timeoutMs * 10,
+    timeoutMs, retryMs, deadline: Date.now() + timeoutMs,
   });
   for (;;) {
     if (holderStillWedged() || ceilingReached()) break;
--- a/pipeline-lock.mjs
+++ b/pipeline-lock.mjs
@@ -246,21 +246,43 @@ export function lockRecoveryVerdict(lockDir, staleMs) {
  * previous fingerprint and a re-armable window.
  *
  * @param {string} lockDir - The lock directory this caller is waiting on.
- * @param {{timeoutMs: number, retryMs: number, deadline: number, hardDeadline: number}} timing
+ * @param {{timeoutMs: number, retryMs: number, deadline: number, hardDeadline?: number}} timing
+ *   `hardDeadline` is OPTIONAL: omit it and the policy applies its own default
+ *   ceiling, DEFAULT_MAX_WAIT_FACTOR x timeoutMs from now. Pass one only when
+ *   the caller has a real knob of its own (acquirePipelineLock's `maxWaitMs`).
  * @returns {{backoffMs: () => number, holderStillWedged: () => boolean}}
  */
 export function createLockWaitPolicy(lockDir, { timeoutMs, retryMs, deadline, hardDeadline }) {
   let perHolderDeadline = deadline;
   let lastFingerprint;
 
+  // The ceiling is the policy's, not each caller's (#3895). Three lock modules
+  // used to write `Date.now() + timeoutMs * 10` at their own call sites, so the
+  // bound the clamp below applies existed in four places. Copies of an
+  // invariant stay correct only until one is edited, and a wrong ceiling fails
+  // silently — it changes retry timing, which nothing asserts and nobody
+  // reports. Deriving it here leaves one copy for them all.
+  const ceiling = hardDeadline ?? Date.now() + timeoutMs * DEFAULT_MAX_WAIT_FACTOR;
+  // A ceiling that is not a number is not a ceiling, and it fails INVISIBLY:
+  // Math.min(x, NaN) is NaN, Math.max(0, NaN) is NaN, and setTimeout(NaN) fires
+  // immediately — so a caller that got this wrong would spin hot through the
+  // retry loop rather than raise anything. Infinity is a deliberate, legible
+  // "no ceiling" (acquirePipelineLock passes it for maxWaitMs: Infinity) and is
+  // kept; NaN and non-numbers are mistakes and are refused where they are made.
+  if (typeof ceiling !== 'number' || Number.isNaN(ceiling)) {
+    throw new TypeError(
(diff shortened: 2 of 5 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is `pipeline-lock.mjs`, `portal-health-lock.mjs`, and the lock-policy callers, which is the maintainer's routing call; I am reviewing the owned test `tests/pipeline-lock.test.mjs`.

It is sound and mutation-tested. Removing the moved clamp (so `backoffMs` returns the bare unbounded value) reddens tests 22 and 23: test 22 pins the clamped value at about 2000ms via an external literal (not read back from the module, so a moved constant cannot hide the regression) and reddens because unbounded backoff returns hundreds of thousands of ms, and test 23 sleeps past the ceiling and asserts the ceiling is reached and `backoffMs()` is 0. Removing the policy-owned default derivation (falling back to the old caller-supplies model) reddens tests 22, 23, and 24, confirming the default ceiling is what is pinned rather than a caller value. The acquire/release axis is covered by the pre-existing suite (live-holder blocks and times out, stale-reclaim serialization, `maxWaitMs`-bounded churn, release without double-delete) running through the refactored policy, and a grep guard pins that no module re-derives the hard deadline. Assertions use external literals and behavioral checks, not values echoed from the module.

PR #3911, opened 2026-09-05
fix: Indonesian liveness/eval-mode/model-id gaps + Kalibrr provider

Binis-code:
## Summary

Four fixes found while running a 39-offer Indonesian batch through this system. Each one failed *silently* rather than erroring, so they're grouped by that shared symptom rather than by file.

## 1. Indonesian apply-control patterns (`liveness-core.mjs`)

`APPLY_PATTERNS` had entries for English, Spanish, German, French, Polish and Chinese, but no Indonesian. A live Jobstreet Indonesia posting's apply control reads `"Lamaran Cepat Kirim lamaran sebagai <role> di <company>"`, which matched nothing — so `check-liveness.mjs` classified every live Indonesian posting as `no_apply_control` and dropped it. Measured 2026-09-04: **40 of 40** verified offers were dropped in one scan, 0 survived.

Worse, that verdict is written to `data/scan-history.tsv` as `skipped_no_apply_control`, which dedup-blocks the URL from every later scan — so the false negative was also permanent.

Verified on two real pages before patching:
- live posting (`id.jobstreet.com/id/job/94396926`) → HTTP 200, control `"Lamaran Cepat Kirim lamaran sebagai…"`, body 4120 chars
- removed posting (`id.jobstreet.com/id/job/93338944`) → HTTP 200, controls only `"Cari lowongan kerja lain"`, body carries `"Lowongan kerja ini tidak lagi diiklankan"`

Added `/lamaran cepat/i`, `/kirim lamaran/i`, `/lamar sekarang/i`, `/^lamar$/i` to `APPLY_PATTERNS` (narrow phrases, not a bare `/lamar/`, which also fires on the `"Lamaran kerja"` footer link present on both live and dead pages), and `/lowongan kerja ini tidak lagi diiklankan/i` to `HARD_EXPIRED_PATTERNS` so a removed posting resolves as `expired` instead of falling through to `uncertain`.

## 2. `modes/id` was unreachable (`gemini-eval.mjs`, `batch-evaluate-gemini.mjs`)

`modes/id` ships its evaluation mode as `lowongan.md`, but `gemini-eval.mjs` resolves `language.modes_dir` against a `candidateFiles` list containing `oferta.md`, `angebot.md`, `offre.md`, `kyujin.md`, `is-ilani.md`, `naukri.md` — not `lowongan.md`. So an Indonesian profile always hit the "No matching evaluation file found" fallback and evaluated against the English rubric. Added `lowongan.md` to the list.

`batch-evaluate-gemini.mjs` had the same gap and no warning at all: it hardcoded `modes/_shared.md` and `modes/oferta.md` and never read `language.modes_dir`. All 39 offers in the batch that surfaced this were evaluated against the wrong mode files. It now resolves the directory the same way `gemini-eval.mjs` does, refuses a value that would escape the project root, falls back per-file (a market directory may localise only some modes), and shares the same evaluation-filename list so the two entry points can't drift apart again.

## 3. Gemini model ids were dead (`batch-evaluate-gemini.mjs`)

The spend-tier → model map still named the 2.5 family: `gemini-2.5-flash` (deprecated 2026-06-17) and `gemini-2.5-flash-lite` (deprecated 2026-07-22). Both now answer `404 … no longer available to new users`, so every offer in a batch failed. Updated the map to `gemini-3.5-flash-lite` / `gemini-3.5-flash` / `gemini-3.1-pro-preview`.

## 4. Fact-check gate was English-only (`verify-cv-facts.mjs`)

`METRIC_NOUNS` had no Indonesian entries, so no count in an Indonesian-language CV was ever extracted, and the fact-check gate reported "not checked" for every quantified claim instead of actually verifying them. Confirmed on a real CV before and after: `"Mengelola 5000 dokumen arsip dan 250 vendor"` passed silently before this change and is now correctly flagged as an unverified metric claim. Added the Indonesian nouns an admin/procurement/logistics CV typically counts (`dokumen`, `pegawai`, `vendor`, `unit`, `cabang`, etc.), and corrected the advisory message, which attributed every "not checked" case to the English-only noun list even when the actual cause was an identifier (a GPA, a test score) rather than a missing translation.

## 5. New provider: Kalibrr (`providers/kalibrr.mjs`)

Not a fix — Indonesian coverage in `providers/` was a single source (`jobstreet.mjs`), which is thin for a market this size. Kalibrr is one of the largest job boards in Indonesia and the Philippines; its public job-board search API needs no auth and returns `description`/`qualifications` inline, so the scan stays zero-token (no per-job request needed for `content_filter`). Postings resolve to identifiable employers and are free for the candidate, so the source clears the Source Indexing Policy.

Measured 2026-09-06: `GET /kjs/job_board/search?limit=30&offset=0&country=Indonesia&text=admin` returns HTTP 200 with 18 matching jobs out of 1057 Indonesian postings; job URLs build as `/c/{company-code}/jobs/{id}/{slug}` and all were well-formed, one confirmed live by `check-liveness.mjs`.

One thing deliberately **not** used: the API also accepts a `location` param, but it doesn't filter — `location=Karawang` returned 1169 results spread across Riau, Banten and Jakarta and *also cancelled the `text` filter* (18 results became 1169). Region narrowing is left to `portals.yml`'s existing `location_filter`, which reads the `"City, Region"` string this provider emits from `google_location.address_components`. Postings past `application_end_date` are dropped at parse time so a closed listing never reaches the pipeline.

## Testing

`node --test tests/liveness-core.test.mjs tests/fact-gate-language-coverage.test.mjs tests/nonmetric-fact-gate.test.mjs tests/cover-fact-gate.test.mjs` → 10/10 pass.

`node validate-portals.mjs` and `node verify-pipeline.mjs` → 0 errors on a live installation running this branch since 2026-09-04, with all five fixes exercised against real Jobstreet/Kalibrr traffic (not just the unit tests above).

## Notes for reviewers

- No config files, secrets, or personal data in this diff — checked before pushing.
- Happy to split into separate PRs per fix if that's easier to review; grouped them here because they were all found in the same debugging session and share the "silent failure" pattern.

## Summary

This change improves Indonesian job processing and adds Kalibrr support.

### User impact

- Indonesian Jobstreet postings now detect active apply controls and expired listings: `liveness-core.mjs:29-31`, `liveness-core.mjs:100-102`.
- Indonesian Gemini evaluations can load `modes/id/lowongan.md`: `gemini-eval.mjs:106-135`.
- Batch evaluation resolves market mode files from `modes_dir` and applies safe fallbacks: `batch-evaluate-gemini.mjs:34-67`.
- Gemini uses current model IDs: `batch-evaluate-gemini.mjs:98-105`.
- CV verification recognizes Indonesian count nouns and gives clearer coverage guidance: `verify-cv-facts.mjs:41-138`, `verify-cv-facts.mjs:546`.
- Users can configure Kalibrr searches for Indonesian and Philippine listings: `providers/kalibrr.mjs:151-194`.
- Kalibrr filters expired or malformed postings and suppresses duplicates. Auto-detection remains disabled.

### System files touched

- `modes/`: Indonesian evaluation mode resolution uses `modes/id/lowongan.md`: `gemini-eval.mjs:106-135`.
- `providers/`: Adds `providers/kalibrr.mjs`.
- `AGENTS.md`: Documents `language.modes_dir`, but no change is shown.
- `update-system.mjs`, `DATA_CONTRACT.md`, and `.github/`: No change is shown.

### DIFF
--- a/batch-evaluate-gemini.mjs
+++ b/batch-evaluate-gemini.mjs
@@ -31,9 +31,42 @@ import { isMainModule } from './lib/is-main-module.mjs';
 
 const ROOT = dirname(fileURLToPath(import.meta.url));
 const DATA_ROOT = getCareerOpsRoot();
+// Resolve language.modes_dir from config/profile.yml the way gemini-eval.mjs
+// does. Without this the batch evaluator always loaded the English modes/
+// files, so a profile selecting modes/id (or any other market) silently got
+// the generic rubric - the localisation was configured and never applied.
+// Measured 2026-09-04: 39 Indonesian offers all evaluated against modes/oferta.md.
+function resolveModesDir() {
+  try {
+    const ymlPath = join(DATA_ROOT, 'config', 'profile.yml');
+    if (!existsSync(ymlPath)) return join(ROOT, 'modes');
+    const m = readFileSync(ymlPath, 'utf-8').match(/^[ 	]*modes_dir[ 	]*:[ 	]*(.+)$/m);
+    if (!m) return join(ROOT, 'modes');
+    const rel = m[1].trim().replace(/^["']|["']$/g, '');
+    const candidate = join(ROOT, rel);
+    // Never let a profile value escape the project root.
+    if (!candidate.startsWith(ROOT)) return join(ROOT, 'modes');
+    return existsSync(candidate) ? candidate : join(ROOT, 'modes');
+  } catch { return join(ROOT, 'modes'); }
+}
+
+const MODES_DIR = resolveModesDir();
+// Fall back per file: a market dir may localise only some of the modes.
+const modeFile = (name) =>
+  existsSync(join(MODES_DIR, name)) ? join(MODES_DIR, name) : join(ROOT, 'modes', name);
+
+// The evaluation mode is named per market - oferta.md in Spanish, lowongan.md
+// in Indonesian, and so on. Same list gemini-eval.mjs resolves against; keep
+// the two in step or a market localises for one entry point and not the other.
+const EVAL_MODE_FILES = ['oferta.md', 'lowongan.md', 'angebot.md', 'offre.md', 'kyujin.md', 'is-ilani.md', 'naukri.md'];
+const evalModeFile = () => {
+  const found = EVAL_MODE_FILES.find((f) => existsSync(join(MODES_DIR, f)));
+  return found ? join(MODES_DIR, found) : join(ROOT, 'modes', 'oferta.md');
+};
+
 export const PATHS = {
-  shared:      join(ROOT, 'modes', '_shared.md'),
-  oferta:      join(ROOT, 'modes', 'oferta.md'),
+  shared:      modeFile('_shared.md'),
+  oferta:      evalModeFile(),
   cv:          join(DATA_ROOT, 'cv.md'),
   profile:     join(ROOT, 'modes', '_profile.md'),
   profileYml:  join(DATA_ROOT, 'config', 'profile.yml'),
@@ -59,13 +92,17 @@ function readSpendTier() {
   return 'standard';
 }
 
+// Model ids per spend tier. The 2.5 family this used to name was deprecated
+// (2.5-flash 2026-06-17, 2.5-flash-lite 2026-07-22) and now answers HTTP 404
+// "no longer available to new users", so every offer in a batch failed before
+// this was updated. Keep in step with the deprecation table in gemini-eval.mjs.
 function spendTierToModel(tier) {
   switch (tier) {
-    case 'economy': return 'gemini-2.5-flash-lite';
-    case 'premium': return 'gemini-2.5-pro';
+    case 'economy': return 'gemini-3.5-flash-lite';
+    case 'premium': return 'gemini-3.1-pro-preview';
     case 'standard':
     default:
-      return 'gemini-2.5-flash';
+      return 'gemini-3.5-flash';
   }
 }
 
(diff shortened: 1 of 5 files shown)


Scott-Emberson (commented):
Two things on the provider, both within the owned-path lane since this adds `providers/kalibrr.mjs`.

First, the new provider ships with no owned test. Nearly every provider in the repo carries a `tests/providers/<name>.test.mjs`, and this one adds a 200-line provider with SSRF-guarding logic and an HTML/JSON parser but no `tests/providers/kalibrr.test.mjs`. On a code read the SSRF guards are correct: `ALLOWED_KALIBRR_HOSTS` is an exact-hostname `Set` matched via `.has(parsed.hostname)` (not substring or suffix), `assertKalibrrUrl` enforces HTTPS-only, every fetch passes `redirect:'error'`, and only the operator-configured `entry.api` can set the base (validated before use, and `searchParams.set` cannot mutate the host). So the guard is correct today, but nothing pins it, so a later edit that loosened the allowlist or dropped `redirect:'error'` would ship green. Please add a kalibrr owned test asserting `assertKalibrrUrl` rejects a non-kalibrr host, a non-HTTPS URL, and a redirect, plus the parser's basic shape, the same pattern the other providers follow.

Second, a scope note for the maintainer: this PR bundles the net-new provider with four unrelated Indonesian-market changes across core files (eval-mode/model-id in `batch-evaluate-gemini.mjs` and `gemini-eval.mjs`, liveness patterns in `liveness-core.mjs`, and metric-nouns in `verify-cv-facts.mjs`). Each is individually small, but they share only the "Indonesian" theme, not a code path, so the provider in particular would be cleaner as its own PR with its own test. Recommending a split.

The branch is also behind main and will need a rebase. I did not find an SSRF hole on the code read, but the missing test is a real gap for a security-guarding provider.

PR #3922, opened 2026-09-06
feat: add control to ollama ctx size and implement streaming from olla…

### LINKED ISSUE #3920: ollama-eval: num_ctx and the prompt budget are hardcoded to 32768, so the recommended 128k models lose 75% of their window and no full A-G report fits
### Code of Conduct

### Existing issues

### What happened?

What happened?
==============

## Summary

`ollama-eval.mjs` pins the context window to **32,768 tokens in two places** — the prompt
budget and the request itself. Neither is configurable. On every real JD the A-G prompt
exceeds the resulting budget, three context sections are silently dropped, and what survives
leaves under 4k tokens for the model to actually write the report in.

The cap contradicts the script's own recommended-model list, 15 lines above it: three of the
four models it names ship 128k windows.

## The code

`ollama-eval.mjs:220` — the prompt budget:

```js
const { contextBody, budgetReport } = buildBudgetedPrompt({
  ...
  maxTokens: 32_768, // matches options.num_ctx below
});
```

`ollama-eval.mjs:292` — the request:

```js
      options: { temperature: 0.4, num_ctx: 32768 },
```

`buildBudgetedPrompt` subtracts an 8,192-token safety margin, so the usable budget is
**24,576 tokens**. Every evaluation I have run overshoots it:

```
📊  Token budget: 31540 → 29378 tokens (saved 2162)
    Trimmed sections: Sources of Truth (EXCLUSIVE), Spend Tier (Model Routing),
                      Company Type and Compensation Reliability
    ⚠️  Still 4802 tokens over budget after compression
```

Two consequences, and the second is the one that ruins the output:

1. **Three context sections are dropped from every run**, including *Sources of Truth
   (EXCLUSIVE)* — and the prompt is still ~4.8k over budget afterwards.
2. **There is no room left to generate.** A ~29k prompt in a 32,768 window leaves under 4k
   tokens for the report. A-G reports run 10–12k. Measured: a local run produced a 15.9 KB
   report where the same rubric and the same inputs produce 42.8 KB.

## The cap contradicts the docstring above it

`ollama-eval.mjs:21`:

```js
 *   Recommended models (32K+ context): llama3.3, mistral-nemo, qwen2.5, gemma3
```

Of those four, only `qwen2.5` is actually 32,768. Verified with `ollama show`:

| model | context length | usable under the current cap |
|---|---|---|
| gemma4:31b | 262144 | 12% |
| gemma3 / llama3.3 / mistral-nemo | 131072 | 25% |
| qwen2.5:32b | 32768 | 100% |

So for three of the four recommended models, and every newer one, the code discards most of
the window it told the user to go get.

## Reproduction

1. `ollama pull gemma4:31b` (or any 128k model from the recommended list)
2. `OLLAMA_MODEL=gemma4:31b node ollama-eval.mjs --file jds/<any-real-jd>.md`

**Expected:** the prompt fits, no sections are trimmed, and the model has room to write a
full A-G report.

**Actual:** the budget warning above, three sections dropped, and a short report.

Reproduces on any real JD — the prompt is ~29–31k before the JD is even large.

## Proposed fix

Read the window from the environment, defaulting to today's value so nothing changes for
anyone who sets nothing:

```js
// beside the other env reads, near baseUrl
const numCtx = parseInt(process.env.OLLAMA_NUM_CTX || '32768', 10);

// :220
  maxTokens: numCtx,

// :292
      options: { temperature: 0.4, num_ctx: numCtx },
```

Measured with `OLLAMA_NUM_CTX=65536` on gemma4:31b: the budget warning disappears entirely,
no sections are trimmed, prompt 31.2k with ~35k of headroom, and the report comes out at the
expected length. Memory cost is modest — 20 GB resident, 100% GPU, on a 48 GB machine.

Two things worth documenting next to the recommended-model list if this lands:

- **The model's own ceiling still applies.** `OLLAMA_NUM_CTX` above what `ollama show` reports
  as `context length` does not buy a bigger window; qwen2.5's 32768 is a hard cap.
- **KV cache grows with the window**, so the guidance should pair a window with a RAM figure
  rather than just naming models.

Happy to open the PR.

## Why this isn't already covered

**#3534** (*docs(budget): a verified quick-start for running a local model*, closed) documents
how to get a local model running — but a user who follows it and picks any of the three 128k
models still cannot produce a complete A-G report, because the cap is in the code, not the
docs. This is the code-side half of that issue.

**#3707** and **#3795** touch `ollama-eval` output formatting (tracker rows, missing `**URL:**`
header). Neither concerns the context window.

## Related, filed separately

Raising the window alone is not sufficient to get a working run: a longer report then takes
over five minutes to generate, and the request dies as `fetch failed` at exactly 5:01.
`stream: false` (`:285`) means Ollama sends no response headers until generation completes, and
Node's undici client applies its own 300s `headersTimeout` that neither `OLLAMA_TIMEOUT_MS`
nor `AbortSignal.timeout` controls. Separate defect, separate issue.

### Steps to reproduce

1. `ollama pull gemma4:31b`
2. `OLLAMA_MODEL=gemma4:31b node ollama-eval.mjs --file jds/<any-real-jd>.md`
3. Read the "Token budget" lines printed before the model call.

### Expected behavior

The context window should follow the selected model. A 128k or 256k model should be given a
window large enough for the full A-G prompt plus a 10–12k report, with the 32768 default
preserved for anyone who does not opt in.

### CLI tool

ollama-eval.mjs (career-ops v1.32.0), ollama 0.33.2

### OS and Node.js version

macOS 26.6.2 (Apple Silicon, 48 GB), Node v26.8.1

### Output of npm run doctor

```shell
Result: All checks passed. You're ready to go!
```

### LINKED ISSUE #3921: ollama-eval: with stream:false the response carries no headers until generation ends, so undici's 300s headersTimeout kills any run over 5 minutes as a bare "fetch failed"
### Code of Conduct

### Existing issues

## Summary

`ollama-eval.mjs` posts with `stream: false`, so Ollama buffers the entire report and sends
**no response headers until generation has finished**. Node's undici client applies its own
`headersTimeout`, which defaults to **300,000 ms**, and aborts the request when nothing has
arrived by then.

The result is a hard five-minute ceiling on report generation that no setting controls.
`OLLAMA_TIMEOUT_MS` does not affect it. It surfaces as `❌ Ollama API call failed: fetch
failed`, which reads as an Ollama or network fault rather than a client-side deadline.

This is exactly the failure class already accepted and fixed in **#3024** ("the run stream
goes silent for the whole cv-html envelope… nothing downstream can distinguish a long silence
from a hung request"), one layer down: there the silence was in `/api/run`, here it is in the
Ollama call itself.

## The code

`ollama-eval.mjs:285` — the request buffers:

```js
      stream: false,
```

`ollama-eval.mjs:266` — the only timeout the script knows about:

```js
const timeoutMs = parseInt(process.env.OLLAMA_TIMEOUT_MS || '300000', 10);
```

…applied at `:294` as `signal: AbortSignal.timeout(timeoutMs)`. That governs the *whole
request*. It does not govern undici's separate `headersTimeout`, and there is no way to reach
that from here — `AbortSignal` cannot extend it, and `undici` is not importable in this
context (`ERR_MODULE_NOT_FOUND`) so a custom `Agent({ headersTimeout: 0 })` dispatcher is not
available without adding a dependency.

## Measured

Same machine, same rubric, same JD shapes. The only variable is whether the response streams.

| run | model | streams? | duration | outcome |
|---|---|---|---|---|
| Cortex (cold model) | qwen2.5:32b | no | **5:01.19** | `fetch failed` |
| Cortex (warm model) | qwen2.5:32b | no | **4:54.56** | succeeded |
| UHG | gemma4:31b | no | **5:01.16** | `fetch failed` |
| Cortex | gemma4:31b | yes | 11:44.62 | succeeded |
| Shopify | gemma4:31b | yes | 11:05.46 | succeeded |
| UHG | gemma4:31b | yes | 11:32.32 | succeeded |

Two things this shows:

1. **The boundary is 5:00, not the model.** The 4:54 run is the same model on the same prompt
   as the 5:01 run — it succeeded only because the model was already resident and it squeaked
   under the deadline by six seconds.
2. **Real runs exceed it routinely.** Every streamed run took 10–12 minutes. A full A-G report
   is 10–12k tokens, and no local model on consumer hardware writes that in under five
   minutes. So the ceiling is not an edge case; it is the normal path.

`OLLAMA_TIMEOUT_MS=900000` was verified present in the environment for the failing runs. The
proof it was not the script's own abort: `:319` has a dedicated branch for that —

```js
  if (err.name === 'TimeoutError') {
    console.error(`❌  Request timed out after ${Math.round(timeoutMs / 1000)}s.`);
    console.error(`    Try a smaller/faster model, or increase OLLAMA_TIMEOUT_MS.`);
  } else {
    console.error(`❌  Ollama API call failed: ${err.message}`);
  }
```

…and that branch never fired. Every failure took the generic `else`.

## Reproduction

1. Pull a model large enough that a full report takes over five minutes — `gemma4:31b` or
   `qwen2.5:32b` on Apple Silicon both do.
2. `OLLAMA_TIMEOUT_MS=900000 OLLAMA_MODEL=gemma4:31b node ollama-eval.mjs --file jds/<a-real-jd>.md`
3. Wait.

**Expected:** the run completes, or fails with the script's own timeout message after 900s.

**Actual:** `❌ Ollama API call failed: fetch failed` at 5:01, with `OLLAMA_TIMEOUT_MS` ignored
and no indication that a client-side deadline was responsible.

## Proposed fix

Switch to `stream: true` so headers arrive with the first token, which removes the deadline
entirely regardless of how long generation runs.

This is not a one-line change: `:304` does `await res.json()`, which cannot read an NDJSON
stream. The response body has to be accumulated:

```js
      stream: true,
```

```js
  // Streamed /api/chat is newline-delimited JSON: one object per token, the last carrying
  // done:true and the token counts.
  let acc = '', buf = '', promptCount = 0, evalCount = 0;
  const decoder = new TextDecoder();
  for await (const chunk of res.body) {
    buf += decoder.decode(chunk, { stream: true });
    let nl;
    while ((nl = buf.indexOf('\n')) !== -1) {
      const line = buf.slice(0, nl).trim();
      buf = buf.slice(nl + 1);
      if (!line) continue;
      let obj;
      try { obj = JSON.parse(line); } catch { continue; }
      if (obj.error) { console.error(`❌  Ollama error: ${obj.error}`); process.exit(1); }
      if (obj.message?.content) acc += obj.message.content;
      if (obj.done) { promptCount = obj.prompt_eval_count ?? 0; evalCount = obj.eval_count ?? 0; }
    }
  }
  evaluationText = acc.trim();
```

The token counts move with it — `prompt_eval_count` and `eval_count` arrive on the `done:true`
chunk rather than on a single JSON body, so the `normalizeOpenAIUsage` call at `:307` reads
those accumulators instead of `data.*`. Everything downstream is unchanged.

Verified response shape against ollama 0.33.2:

```
{"model":"...","message":{"role":"assistant","content":"."},"done":false}
{"model":"...","message":{"role":"assistant","content":""},"done":true,"done_reason":"stop",
 "prompt_eval_count":31,"eval_count":3, ...}
```

A side benefit: with `stream: true` the report can be printed as it is produced, so a
ten-minute run stops looking indistinguishable from a hang.

Happy to open the PR.

## Why this isn't already covered

**#3024** (closed) is the same defect one layer up and was accepted on exactly this reasoning:
a response that emits nothing while the model works is indistinguishable from a hung request,
so the client drops it. That fix went into `/api/run`; the Ollama call never got it.

`docs/RUNNING_ON_A_BUDGET.md` §5 directs local users to **32B+ or 70B+ models** on 48 GB of
VRAM. Those are precisely the models that take more than five minutes to write an A-G report,
so the documented configuration is the one that cannot complete a run. The `docs(budget)`
series — **#2978**, **#3534**, **#3881** — maintains that path; this is a code-side blocker on it.

**#3707** and **#3795** touch `ollama-eval` output formatting. **#3796** notes the four
evaluators have already drifted apart in private helper copies. Neither concerns transport.

## Scope

`ollama-eval.mjs` is the only evaluator affected — `openai-eval.mjs`, `gemini-eval.mjs` and
`openrouter-runner.mjs` contain no `stream: false` and go through SDKs with their own
transport handling.

## Related, filed separately

The same file hardcodes `num_ctx` and the prompt budget to 32768, which caps the report length
independently of this. Filed as its own issue; the two together are what make a full A-G
report possible on a local model, but they are independent defects and either can land alone.

### Steps to reproduce

1. `ollama pull gemma4:31b`
2. `OLLAMA_TIMEOUT_MS=900000 OLLAMA_MODEL=gemma4:31b node ollama-eval.mjs --file jds/<a-real-jd>.md`
3. Wait. The run dies at 5:01 with `fetch failed`, regardless of OLLAMA_TIMEOUT_MS.

### Expected behavior

A run should be bounded by `OLLAMA_TIMEOUT_MS`, not by an undocumented 300s client deadline —
and when a timeout does fire it should report as the script's own timeout, naming the setting
that controls it, rather than as a generic `fetch failed`.

### CLI tool

ollama-eval.mjs (career-ops v1.32.0), ollama 0.33.2

### OS and Node.js version

macOS 26.6.2 (Apple Silicon, 48 GB), Node v26.8.1

### Output of npm run doctor

```shell
Result: All checks passed. You're ready to go!
```

EvgeniGordeev:
## What does this PR do?

Makes `ollama-eval`'s context window configurable via `OLLAMA_NUM_CTX` (defaulting to today's
32768, so it is a no-op unless set) and switches the chat request to `stream: true`. Together
these are what a local model needs to finish a full A-G report: the window fix stops the prompt
overflowing its budget, and streaming stops undici's 300s `headersTimeout` killing every run
that generates for more than five minutes.

### 1. Context window follows the model — #3920

`num_ctx` and the prompt budget were hardcoded to 32768 in two places. Both now read one value:

```js
const numCtx = parseInt(process.env.OLLAMA_NUM_CTX || '32768', 10);
```

The cap contradicted the script's own recommended-model list at `:21` — three of those four
models ship 128k windows — and `docs/RUNNING_ON_A_BUDGET.md` §5, which tells local users to run
a 32B+/70B+ model for reliable scoring. On every real JD the prompt overshot the resulting
24576-token budget (32768 minus the 8192 safety margin): three context sections were dropped,
the prompt was still ~4.8k over, and under 4k tokens remained to write a report that runs 10–12k.

```
📊  Token budget: 31540 → 29378 tokens (saved 2162)
    Trimmed sections: Sources of Truth (EXCLUSIVE), Spend Tier (Model Routing),
                      Company Type and Compensation Reliability
    ⚠️  Still 4802 tokens over budget after compression
```

With `OLLAMA_NUM_CTX=65536` on a 256k model: no warning, no sections trimmed, 31.2k prompt with
~35k of headroom.

### 2. The response streams — #3921

`stream: false` meant Ollama sent no response headers until generation finished, and Node's
undici client aborts at its own 300s `headersTimeout` — a deadline neither `OLLAMA_TIMEOUT_MS`
nor `AbortSignal.timeout` controls. Long runs died as a bare `fetch failed`, taking the generic
error branch while the purpose-built `TimeoutError` branch at `:319` never fired, so the symptom
pointed at Ollama rather than at the client.

`await res.json()` cannot read an NDJSON stream, so the body is accumulated line by line.
`prompt_eval_count` / `eval_count` arrive on the `done: true` chunk and feed
`normalizeOpenAIUsage` as before; nothing downstream changes.

Measured on one machine, same rubric, same JD shapes:

| run | model | streams? | window | duration | outcome |
|---|---|---|---|---|---|
| A (cold) | qwen2.5:32b | no | 32768 | **5:01.19** | `fetch failed` |
| A (warm) | qwen2.5:32b | no | 32768 | **4:54.56** | succeeded |
| B | gemma4:31b | no | 32768 | **5:01.16** | `fetch failed` |
| A | gemma4:31b | yes | 65536 | 11:44.62 | succeeded |
| C | gemma4:31b | yes | 65536 | 11:05.46 | succeeded |
| B | gemma4:31b | yes | 65536 | 11:32.32 | succeeded |

The 4:54 row is the load-bearing one: same model and prompt as the 5:01 failure above it,
surviving only because the model was already resident. The boundary is 5:00, not the model.

### Tests

New `tests/ollama-eval-streaming-context.test.mjs` (own file per CONTRIBUTING, auto-discovered)
drives the real CLI against a mock Ollama and asserts: `stream: true` is sent, NDJSON chunks
reassemble in order, the token counts are read off the `done:true` chunk, `OLLAMA_NUM_CTX=65536`
reaches `options.num_ctx`, an unset value keeps the 32768 default, and a final chunk with no
trailing newline is still flushed.

That last case is a regression test for a bug the existing suite caught during development: the
first cut of the parser only flushed on `\n`, so a body ending mid-line was silently dropped.
Ollama terminates every chunk with a newline, which is why six live runs never surfaced it.

`tests/ollama-eval-endpoint.test.mjs` is updated in the same commit-range, for two reasons that
follow from the change: its mock now speaks NDJSON, and it pins `OLLAMA_NUM_CTX` explicitly so
the assertion tests the request plumbing rather than whatever the contributor's `.env` happens
to set. Without that pin the test fails on any machine where that variable is exported.

## Related issue

Closes #3920
Closes #3921

## Type of change

- [x] Bug fix

## Summary

Users can configure the Ollama context window with `OLLAMA_NUM_CTX`. The default remains `32768`. The value controls prompt budgeting and `options.num_ctx` in `ollama-eval.mjs`.

Ollama responses now stream as NDJSON. The evaluator preserves content order, token counts, streamed errors, and final response lines without trailing newlines. Long generations can now run under `OLLAMA_TIMEOUT_MS` instead of undici’s 300-second headers timeout.

Tests cover streaming, chunk ordering, token counts, context configuration, defaults, endpoint behavior, and unterminated final lines.

## Files changed

- `ollama-eval.mjs`: Adds context configuration and streamed response handling.
- `tests/ollama-eval-endpoint.test.mjs`: Updates endpoint mocks for NDJSON streaming.
- `tests/ollama-eval.test.mjs`: Adds integration coverage for the new behavior.

### DIFF
--- a/ollama-eval.mjs
+++ b/ollama-eval.mjs
@@ -104,6 +104,10 @@ let baseUrl   = (process.env.OLLAMA_BASE_URL || 'http://localhost:11434').replac
 // 32768, so behaviour is unchanged unless OLLAMA_NUM_CTX is set. Raise it for a model with a
 // bigger window; `ollama show` reports each model's ceiling (qwen2.5 caps at 32768).
 const numCtx = parseInt(process.env.OLLAMA_NUM_CTX || '32768', 10);
+if (Number.isNaN(numCtx) || numCtx <= 0) {
+  console.error(`❌  Invalid OLLAMA_NUM_CTX: "${process.env.OLLAMA_NUM_CTX}" — must be a positive integer (tokens).`);
+  process.exit(1);
+}
 let saveReport = true;
 
 for (let i = 0; i < args.length; i++) {
@@ -339,6 +343,10 @@ try {
   if (tail) {
     try {
       const obj = JSON.parse(tail);
+      if (obj.error) {
+        console.error(`❌  Ollama error: ${obj.error}`);
+        process.exit(1);
+      }
       if (obj.message?.content) acc += obj.message.content;
       if (obj.done) {
         promptCount = obj.prompt_eval_count ?? promptCount;
--- a/ollama-eval.mjs
+++ b/ollama-eval.mjs
@@ -100,6 +100,10 @@ if (args.length === 0 || args[0] === '--help' || args[0] === '-h') {
 let jdText    = '';
 let modelName = process.env.OLLAMA_MODEL || 'llama3.3';
 let baseUrl   = (process.env.OLLAMA_BASE_URL || 'http://localhost:11434').replace(/\/$/, '');
+// Context window for the request AND the prompt budget. Defaults to the previous hardcoded
+// 32768, so behaviour is unchanged unless OLLAMA_NUM_CTX is set. Raise it for a model with a
+// bigger window; `ollama show` reports each model's ceiling (qwen2.5 caps at 32768).
+const numCtx = parseInt(process.env.OLLAMA_NUM_CTX || '32768', 10);
 let saveReport = true;
 
 for (let i = 0; i < args.length; i++) {
@@ -217,7 +221,7 @@ const { contextBody, budgetReport } = buildBudgetedPrompt({
   profileYml,
   profileContent,
   jdText,
-  maxTokens: 32_768, // matches options.num_ctx below
+  maxTokens: numCtx, // matches options.num_ctx below
 });
 
 if (budgetReport.compressed) {
@@ -282,14 +286,18 @@ try {
         { role: 'system', content: systemPrompt },
         { role: 'user',   content: `JOB DESCRIPTION TO EVALUATE:\n\n${jdText}` },
       ],
-      stream: false,
+      // stream:true so response headers arrive with the FIRST token. With stream:false
+      // Ollama sends nothing until the whole report is generated, and Node's undici client
+      // gives up at its own 300s headersTimeout — a deadline neither OLLAMA_TIMEOUT_MS nor
+      // AbortSignal.timeout controls, which surfaced as a bare "fetch failed" at 5:01.
+      stream: true,
       // Ollama's native /api/chat reads generation params from `options` only.
       // This call targets that endpoint (NOT the OpenAI-compatible /v1 route,
       // which ignores `options` and has no num_ctx equivalent), so both the
       // deterministic temperature and the enlarged context window actually take
       // effect. Without num_ctx here Ollama defaults to a 2048-token context and
       // silently truncates the prompt; without temperature it runs at 0.8.
-      options: { temperature: 0.4, num_ctx: 32768 },
+      options: { temperature: 0.4, num_ctx: numCtx },
     }),
     signal: AbortSignal.timeout(timeoutMs),
   });
@@ -301,14 +309,50 @@ try {
     process.exit(1);
   }
 
-  const data = await res.json();
-  evaluationText = data.message?.content?.trim();
+  // Streamed /api/chat is newline-delimited JSON: one object per token, the last carrying
+  // done:true and the token counts.
+  let acc = '', buf = '', promptCount = 0, evalCount = 0;
+  const decoder = new TextDecoder();
+  for await (const chunk of res.body) {
+    buf += decoder.decode(chunk, { stream: true });
+    let nl;
+    while ((nl = buf.indexOf('\n')) !== -1) {
+      const line = buf.slice(0, nl).trim();
+      buf = buf.slice(nl + 1);
+      if (!line) continue;
+      let obj;
+      try { obj = JSON.parse(line); } catch { continue; }
+      if (obj.error) {
+        console.error(`❌  Ollama error: ${obj.error}`);
+        process.exit(1);
+      }
+      if (obj.message?.content) acc += obj.message.content;
+      if (obj.done) {
+        promptCount = obj.prompt_eval_count ?? 0;
+        evalCount = obj.eval_count ?? 0;
+      }
+    }
+  }
+  // Flush a final line that arrived without a trailing newline. Ollama terminates every
+  // chunk with one, but a body that ends mid-line would otherwise be dropped silently.
+  const tail = buf.trim();
+  if (tail) {
+    try {
+      const obj = JSON.parse(tail);
+      if (obj.message?.content) acc += obj.message.content;
+      if (obj.done) {
+        promptCount = obj.prompt_eval_count ?? promptCount;
+        evalCount = obj.eval_count ?? evalCount;
+      }
+    } catch { /* a truncated final line is not recoverable; the empty-response check below reports it */ }
+  }
+  evaluationText = acc.trim();
   // Native /api/chat reports tokens as prompt_eval_count / eval_count, not an
   // OpenAI-shaped `usage` object; map them through the shared normalizer.
   const usage = normalizeOpenAIUsage({
-    prompt_tokens: data.prompt_eval_count,
-    completion_tokens: data.eval_count,
-    total_tokens: (data.prompt_eval_count ?? 0) + (data.eval_count ?? 0),
+    prompt_tokens: promptCount,
+    completion_tokens: evalCount,
+    total_tokens: promptCount + evalCount,
   });
   tracker.record('evaluation', usage);
   if (!evaluationText) {
--- a/tests/ollama-eval-endpoint.test.mjs
+++ b/tests/ollama-eval-endpoint.test.mjs
@@ -3,7 +3,7 @@
 //
 // The request was POSTed to the OpenAI-compatible /v1/chat/completions route,
 // which ignores the `options` object and has no num_ctx equivalent, so NEITHER
-// temperature: 0.4 NOR num_ctx: 32768 reached the model — every eval ran at the
+// temperature: 0.4 NOR num_ctx reached the model — every eval ran at the
 // server default temperature and a 2048-token context that silently truncated
 // the prompt. This drives the real CLI against a mock Ollama that captures the
 // request, asserting the endpoint and the params, and that the native response
@@ -22,15 +22,13 @@ const server = createServer((req, res) => {
   req.on('end', () => {
     captured.path = req.url;
     try { captured.body = JSON.parse(raw); } catch { captured.body = null; }
-    // Native /api/chat shape (NOT OpenAI choices[]).
-    res.writeHead(200, { 'Content-Type': 'application/json' });
-    res.end(JSON.stringify({
-      model: 'test',
-      message: { role: 'assistant', content: 'Evaluation.\nVERDICT: 4/5 — solid fit' },
-      done: true,
-      prompt_eval_count: 1234,
-      eval_count: 56,
-    }));
+    // Native /api/chat shape (NOT OpenAI choices[]). Since #3921 the request sets
+    // stream:true, so the reply is newline-delimited JSON rather than one body.
+    res.writeHead(200, { 'Content-Type': 'application/x-ndjson' });
+    res.end([
+      { model: 'test', message: { role: 'assistant', content: 'Evaluation.\nVERDICT: 4/5 — solid fit' }, done: false },
+      { model: 'test', message: { role: 'assistant', content: '' }, done: true, prompt_eval_count: 1234, eval_count: 56 },
+    ].map((c) => JSON.stringify(c)).join('\n') + '\n');
   });
 });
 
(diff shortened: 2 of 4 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is `ollama-eval.mjs`, which is the maintainer's routing call; I am reviewing the owned tests.

Both are sound and mutation-sensitive. The ctx-size passthrough is pinned non-tautologically: the test drives the real CLI against a mock that captures the actual POST body and asserts `body.options.num_ctx === 65536` for a non-default value (so reverting to a hardcoded default reddens) plus the `32768` default when the env is unset. Streaming is pinned across three real behaviors: an NDJSON reply split across three content chunks must reassemble in order, the token counts that arrive only on the `done:true` chunk are asserted via the footer (would show zero if not read off the final chunk), and a partial/no-trailing-newline case exercises the tail-flush path the original parser dropped. The endpoint is safe: `${baseUrl}/api/chat` where baseUrl is operator-configured, validated via `new URL()`, and gated by a loopback guard (localhost only, remote needs an explicit `OLLAMA_ALLOW_REMOTE=1`), not driven by untrusted JD content. The assertions hit the captured real request and real stdout, not mock echoes.

PR #3923, opened 2026-09-06
fix(updater): do not commit the auto-update onto a non-default branch

### LINKED ISSUE #3846: update-system.mjs commits the auto-update onto whatever branch you are on, including a contributor's feature branch
**Time:** ~2-3 h · **Difficulty:** Intermediate · **Area:** `update-system.mjs`

## What happens

`update-system.mjs` commits the files it refreshes onto **whatever branch HEAD is on**, with no check of which branch that is (`update-system.mjs:2464`):

```js
git('commit', '-m', `chore: auto-update system files to v${remote}`);
```

There is a lot of care in that function about *which paths* get staged (owned vs preserved vs excluded). There is none about *which branch* receives the commit.

For someone who only uses career-ops, that is correct and invisible. For someone who **uses it from the same checkout they contribute from**, it means the updater silently drops a full system snapshot onto their feature branch.

## What it costs

Real case, #3648. The contributor wrote a good 39-line change to the Go dashboard, with a test:

```
aa48053 feat(dashboard): add posted-date sort mode
  dashboard/internal/i18n/catalog.go         |  6 ++
  dashboard/internal/ui/screens/pipeline.go  | 11 +-
  dashboard/internal/ui/screens/sort_test.go | 24 +-
  3 files changed, 39 insertions(+), 2 deletions(-)
```

The PR arrived as **+173,092 / -13,920**, because the updater had run three times on that branch first:

```
b1af5c4a chore: auto-update system files to v1.30.0
b9e117dc chore: auto-update system files to v1.26.0
e4dfe896 chore: auto-update system files to v1.24.0
```

Consequences, all of them silent to the author: the PR goes CONFLICTING, the diff is unreviewable, and every automated size/scope signal points at "close this". Nothing is wrong with the code. It is the first contribution of someone with no merges yet, which is exactly the population that does not come back after one confusing round.

To be clear about blast radius: **none of these commits has ever reached `main`** (squash-merge collapses them), so no user is affected. The damage is entirely to the contribution path.

## Repro

1. Clone the repo, check out a feature branch, make a small commit.
2. Run the updater while a newer version exists.
3. `git log --oneline` shows `chore: auto-update system files to vX.Y.Z` sitting on your feature branch.

## Acceptance

- Running the updater on a branch that is not the default branch does **not** silently commit onto it.
- The user is told what is happening and can choose: the exact behaviour is open, and worth discussing in the issue before coding. Options include refusing with an explanation, committing only on the default branch, or asking. Whatever is chosen must keep the plain-user path (on the default branch, no local commits) exactly as it is today: that is the overwhelming majority of runs and it must not grow a prompt.
- A test covers the new branch check.

## Notes for whoever picks this up

- The mirror-image failure is worse than the current one: an updater that refuses too eagerly leaves users on stale system files, and staleness is invisible until something breaks. Bias toward "warn and proceed" over "refuse", unless the discussion concludes otherwise.
- `--force`, `CAREER_OPS_UPDATE_BACKUP_BRANCH` and the backup-branch machinery around `update-system.mjs:84-115` are adjacent; read them before adding a fourth state.

yuki-uix:
The updater committed its refreshed system files onto whatever branch HEAD pointed at. On a non-default branch it now applies the update but withholds the commit, tells the user, and leaves the choice to them.

Closes #3846

## Behaviour

Following the issue's "bias toward warn and proceed over refuse": the files are still refreshed and staged, so nobody is left on stale system files. Only the commit is withheld.

```
Update applied but NOT committed: you are on 'feat/posted-date-sort', not 'main'.
The refreshed system files are staged in your working tree. Committing them here
would put a full system snapshot on your branch, which is how a small pull request
turns into an unreviewable one (#3846).

Pick one:
  1. Keep the update on this branch:
       git commit -m "chore: auto-update system files to v1.33.0"
  2. Move it to 'main' (git 2.35+ for --staged):
       git stash push --staged -m "career-ops v1.33.0"
       git switch main && git stash pop && git commit -m "chore: auto-update system files to v1.33.0"
  3. Undo the update entirely:
       node update-system.mjs rollback

Always want it committed here? Re-run with --commit-on-branch, or set
CAREER_OPS_UPDATE_COMMIT_ON_BRANCH=1.
```

**The plain-user path is untouched** — on the default branch nothing changes, no new output and no prompt. Every uncertain case also commits exactly as before: detached HEAD, and a checkout where the default branch cannot be established (no `origin/HEAD`, no local `main`/`master`, no git at all). Guessing wrong there would withhold the commit from precisely the users the plain path exists for.

`--commit-on-branch` / `CAREER_OPS_UPDATE_COMMIT_ON_BRANCH=1` restores the old behaviour. It is a separate switch rather than a fourth meaning on `--force`, as the issue asked, and is forwarded across the self-reexec because the child process is the one that commits.

## The second-order failure this also had to fix

A withheld update has to survive the *next* run. With its snapshot staged but uncommitted there is no `chore: auto-update system files` commit to serve as `locallyModifiedSystemFiles`' baseline, so the next run read the previous run's own files as edits this install made: it preserved them, wrote `.bak` copies, and skipped exactly the delta it came to install. Measured against a throwaway repo (updater commit at v1, uncommitted v2 tree, upstream at v3), the file landed in the at-risk set — the invisible-staleness failure the issue warns about, arriving one run later. #3648's author ran the updater three times, so this is the same population.

A local ref, `refs/career-ops/staged-update`, records the upstream commit the staged files came from and is used as the baseline while it exists; it is cleared when a commit does land and on rollback. A ref rather than a marker file: it never shows in `git status`, needs no gitignore entry, and cannot leak into a contributor's PR — the same reasoning as the existing `refs/backup-pre-update-wip/*`.

## Changes

- `update-system.mjs` — branch resolution (`currentBranchIn`, `defaultBranchIn` via `origin/HEAD` with a `main`/`master` fallback), the decision (`updateCommitBranchDecision` / `resolveUpdateCommitBranch`), the notice, the opt-in, and the staged-update baseline ref. `updateCommitCommand` is extracted so the command handed to the user is byte-identical to the one the updater would have run — index form vs pathspec form are not interchangeable (the pathspec form drops staged mode bits where `core.fileMode` is false).
- `modes/update.md` — a relay rule, so the `update` mode does not summarize the notice away. It is the only place the user learns their tree is staged-but-uncommitted.
- `tests/updater-branch-guard.test.mjs` — 13 behavioural tests over throwaway repos, following `tests/updater-commit-file-modes.test.mjs`. The repo-facing ones drive `resolveUpdateCommitBranch`, the exact composition `apply()` runs, so a caller wired up wrong cannot pass. The baseline regression carries a negative control, and a companion test pins that a genuine local edit is still flagged (#2337 protection must not become a blanket amnesty).

## The third-order failure, after review

@santifer caught that the ref recording a withheld update was believed on sight, while the state it describes is transient: discarding the update by hand (`git reset --hard`, `git restore`, or the stash-and-switch this guard's own notice suggests) left the ref behind, and every system file then read as a local edit against a snapshot that was no longer there. Reproduced before fixing (`sys.mjs` at risk with the working tree untouched), fixed in ae8fea3.

The ref now points at the snapshot that was actually staged rather than at the upstream commit, and `stagedUpdateBaseline()` asks whether that snapshot is still here — of both the index and the working tree, since each check alone has a blind spot with a real case behind it — before trusting it. A failed check distrusts the ref without deleting it; deleting makes the damage permanent across a `stash push --staged` / `stash pop` round trip, which is measured in scenario 16 below.

## Testing

`node test-all.mjs`: 8,644 passed, 0 failed.

`tests/updater-branch-guard.test.mjs` — behavioural, over throwaway git repos, driving the same functions `apply()` calls:

| # | Scenario | Setup | Expected | What it pins |
|---|---|---|---|---|
| 1 | Feature branch | `origin/HEAD`→`main`, HEAD on `feat/…` | `commit: false`, `non-default-branch` | The bug itself |
| 2 | Default branch | HEAD on `main` | `commit: true`, `on-default-branch` | Plain path unchanged; positive control against over-withholding |
| 3 | Non-`main` default | `origin/HEAD`→`trunk`, then HEAD on `main` | trunk commits, `main` withheld | Default branch is read, not assumed |
| 4 | Detached HEAD | `checkout --detach` | `commit: true`, `detached-head` | Uncertainty must not withhold |
| 5 | No discoverable default | no `origin/HEAD`, no `main`/`master` | `commit: true`, `unknown-default-branch` | Same, for exotic checkouts |
| 6 | `origin/HEAD` absent | local `main` / local `master` | falls back to the one that exists | Fallback answers only with a real branch |
| 7 | Opt-in | `optIn: true` on a feature branch | `commit: true`, `opt-in` | Escape hatch works |
| 8 | Opt-in sources | flag / env / neither / `=0` | true, true, false, false | No accidental activation |
| 9 | `--force` | `--force --confirm`, feature branch | still withheld | No fourth meaning on `--force`; #2337 stays separate |
| 10 | Suggested command | index form vs pathspec form + quoting | matches what the updater would have run | Recovery must not reintroduce the mode-bit loss |
| 11 | Notice content | withheld on `feat/x` | names both branches, keep / move / undo, opt-in | Told, and given a choice |
| 12 | Second run, snapshot pending | v1 committed, v2 staged, upstream v3 | **negative control:** flagged without the snapshot, clean with it | A withheld update is not a local edit |
| 13 | Real local edit, update pending | snapshot recorded, then the user edits the file | still flagged | #2337 protection must not become a blanket amnesty |
| 14a | Discarded by hand | `git reset --hard` | not flagged | Durable ref, transient state |
| 14b | Discarded by hand | `git restore --staged --worktree` | not flagged | Same, second discard shape |
| 15 | Unrelated staged work | contributor stages their own file | snapshot still trusted | The check is scoped to the queried paths |
| 16 | Stash round trip | `stash push --staged` → switch → updater run → `stash pop` | not flagged at either point | Why a failed check distrusts rather than deletes |

Scenarios 12, 14 and 16 each correspond to a defect reproduced with a probe before it was fixed, not to a hypothetical. 2, 4 and 5 are the reverse insurance: they fail if the guard ever starts withholding too eagerly, which is the worse of the two failure directions.

## Left out, deliberately

`rollback` still commits onto the current branch. With this guard in place a feature branch should rarely have an updater commit to roll back, and rolling back a withheld (uncommitted) update produces an empty commit that the existing broad catch already tolerates — but it is the same shape of problem and probably deserves its own issue rather than a widened diff here. Happy to take it if you want it in this PR.

There is also no end-to-end `apply()` test: that needs network and a real remote, so the branch check is covered at the seam `apply()` itself calls. Same boundary as the existing updater suites, noted so it is not mistaken for full coverage.

## Summary

- `update-system.mjs:2180`: Prevents automatic commits on non-default branches unless users opt in.
- `update-system.mjs:2357`: Records and validates the staged update snapshot.
- `update-system.mjs:2389`: Shows commands to keep, move, or roll back the staged update.
- `modes/update.md:105`: Requires the updater notice and recovery guidance.
- `tests/updater-branch-guard.test.mjs:70`: Covers branch checks, recovery flows, snapshot validation, and Git identity handling.

## User impact

On a feature branch, the updater still refreshes and stages system files. It does not commit them automatically. Users can commit the update on the current branch, move it to the default branch, or run `node update-system.mjs rollback`.

Use `--commit-on-branch` or `CAREER_OPS_UPDATE_COMMIT_ON_BRANCH=1` to restore automatic commits. Default-branch and uncertain-case behavior remain unchanged. `--force` does not enable branch commits.

Tests report 8,644 passes and zero failures.

## Touched system files

`AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/`.

### DIFF
--- a/modes/update.md
+++ b/modes/update.md
@@ -102,6 +102,15 @@ If yes:
 8. Show final status:
    > "✅ Updated to v{version}. Run `node doctor.mjs` anytime to verify setup, or `node verify-pipeline.mjs` to check your data."
 
+   If the updater printed `Update applied but NOT committed` (you were on a branch that is not the
+   default one — #3846), relay that block instead of the plain success line, and do not summarize the
+   three options away:
+   > "✅ Updated to v{version}, but you're on branch `{branch}`, not `{default}` — so the updater staged the
+   > refreshed files instead of committing them there. Committing a full system snapshot onto a feature
+   > branch is what turns a small PR into an unreviewable one. You can commit it here, move it to
+   > `{default}`, or undo it with `node update-system.mjs rollback` — the updater printed the exact
+   > commands. Re-run with `--commit-on-branch` if you always want it committed wherever you are."
+
    If the updater's output ended with its note about the CareerOps Manifesto, relay it once (do not drop it when summarizing):
    > "One more thing: this project ships with the CareerOps Manifesto — a new way of job searching is taking shape, and you are already practicing it. Run `npm run manifesto` to read it and sign it if you want to help. No action needed."
 
@@ -122,5 +131,8 @@ If the user says "rollback" or runs `/career-ops update rollback`:
 - Exception: `modes/_profile.md` may be edited **only** in Step 4.7, and **only** after the user explicitly confirms each individual rename/removal. Never batch-edit without per-change consent.
 - User-specific customizations (archetypes, scoring weights, narrative) belong in `modes/_profile.md` or `config/profile.yml`, never in `modes/_shared.md`
 - CLAUDE.md's local additions (everything after the two-line `@AGENTS.md` header) MUST be saved before apply and restored immediately after — on both the success AND failure path (Step 4.2, Step 4.4). `update-system.mjs apply` resets CLAUDE.md before it can fail partway through, so a failed apply still needs the restore. `apply` has no awareness of this content and will silently discard it otherwise.
+- The branch notice (`Update applied but NOT committed`) is not optional output to trim: it is the only
+  place the user learns their tree is staged-but-uncommitted, and dropping it recreates the silence #3846
+  fixed. Relay it whenever it appears, with all three options intact.
 - If anything goes wrong, tell the user to run `node update-system.mjs rollback`
 - Keep the output concise — users don't want walls of text during an update
--- a/tests/updater-branch-guard.test.mjs
+++ b/tests/updater-branch-guard.test.mjs
@@ -21,7 +21,7 @@
 import { mkdtempSync, writeFileSync, rmSync } from 'fs';
 import { tmpdir } from 'os';
 import { join } from 'path';
-import { pass, fail } from './helpers.mjs';
+import { pass, fail, warn } from './helpers.mjs';
 import {
   COMMIT_ON_BRANCH_FLAG,
   commitOnBranchOptIn,
@@ -36,6 +36,7 @@ import {
   updateCommitBranchDecision,
   updateCommitCommand,
   gitIn,
+  gitQuietIn,
 } from '../update-system.mjs';
 
 // A throwaway repo with one commit on `main`, no remote. Remotes are added per
@@ -416,6 +417,56 @@ function stashStagedArgs(g) {
     pass('the contributor\'s own staged work leaves the snapshot trusted');
   } else {
     fail('unrelated staged work invalidated the withheld snapshot');
+  }
+}
+
+// ── 17. The snapshot must be recordable without a git identity ─────────
+{
+  // `commit-tree` needs an author and a committer. On an install with no git
+  // identity — a container, a fresh machine, a harness that isolates the global
+  // config — it fails, recordStagedUpdate()'s catch swallows that, and the next
+  // run falls back to an older baseline and reads this update's own files as
+  // the user's edits: the second-order bug, back through a side door.
+  const { dir, g, gitAt } = repoWithWithheldUpdate();
+  g('config', '--unset', 'user.email');
+  g('config', '--unset', 'user.name');
+
+  const noConfig = join(dir, 'no-such-gitconfig');   // absent file reads as empty, on every platform
+  const restore = { global: process.env.GIT_CONFIG_GLOBAL, system: process.env.GIT_CONFIG_SYSTEM };
+  process.env.GIT_CONFIG_GLOBAL = noConfig;
+  process.env.GIT_CONFIG_SYSTEM = noConfig;
+  try {
+    // NEGATIVE CONTROL: the same call without a pinned identity is what fails.
+    let bareCommitTreeFailed = false;
+    try {
+      // gitQuietIn: this call is EXPECTED to fail, so its stderr is noise.
+      gitQuietIn(dir, 'commit-tree', gitAt('write-tree'), '-p', gitAt('rev-parse', 'HEAD'), '-m', 'probe');
+    } catch {
+      bareCommitTreeFailed = true;
+    }
+
+    recordStagedUpdate('2.0.0', { git: gitAt });
+    let recorded = false;
+    try {
+      recorded = Boolean(gitAt('rev-parse', '--verify', '--quiet', `${STAGED_UPDATE_REF}^{commit}`));
+    } catch {
+      recorded = false;
+    }
+
+    if (!bareCommitTreeFailed) {
+      // This git resolved an identity anyway, so the case cannot be exercised
+      // here — say so rather than passing on a control that never fired.
+      warn('no-identity case not reachable on this git; snapshot recording untested here');
+    } else if (recorded && !atRiskIn(dir, gitAt).includes('sys.mjs')) {
+      pass('the snapshot is recorded even with no git identity configured');
+    } else {
+      fail(`no-identity: recorded=${recorded}, atRisk=${JSON.stringify(atRiskIn(dir, gitAt))}`);
+    }
+  } finally {
+    if (restore.global === undefined) delete process.env.GIT_CONFIG_GLOBAL;
+    else process.env.GIT_CONFIG_GLOBAL = restore.global;
+    if (restore.system === undefined) delete process.env.GIT_CONFIG_SYSTEM;
+    else process.env.GIT_CONFIG_SYSTEM = restore.system;
   }
 }
 
--- a/tests/updater-branch-guard.test.mjs
+++ b/tests/updater-branch-guard.test.mjs
@@ -29,6 +29,7 @@ import {
   defaultBranchIn,
   STAGED_UPDATE_REF,
   locallyModifiedSystemFiles,
+  recordStagedUpdate,
   resolveUpdateCommitBranch,
   skippedBranchCommitNotice,
   updateCommitBranchDecision,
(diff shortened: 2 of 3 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is `update-system.mjs` and `modes/update.md`, both red-line and the maintainer's call; I am reviewing the owned test `tests/updater-branch-guard.test.mjs`.

It is sound. It drives the real `resolveUpdateCommitBranch()`, the exact function `apply()` gates its commit on, against throwaway git repos rather than a stub. Both directions are pinned: on a non-default branch it asserts `commit === false` with `reason === 'non-default-branch'`, and on the default branch it asserts `commit === true` with `reason === 'on-default-branch'` as a positive control against over-withholding. I mutation-checked it: gutting the guard so a non-default branch returns `commit: true` reddens four assertions while the default-branch case correctly stays green, so both sides of the guard are load-bearing. `gitIn` is exported and the suite runs green.

Scott-Emberson (commented):
Test-axis only; the substance is update-system.mjs and modes/update.md (red-line, the maintainer's call). Re-verified at the current head: tests/updater-branch-guard.test.mjs is byte-identical to the head I last verified sound, so this is a rebase that does not touch it and my verdict carries: it drives the real resolveUpdateCommitBranch(), pins both directions (non-default refuses the commit, default proceeds), and reddens on gutting the guard. The branch is now up to date with main. Routing and merge remain the maintainer's call.

Scott-Emberson (commented):
Test-axis only. The substance is `update-system.mjs` and `modes/update.md` (red-line, the maintainer's call); I am reviewing the owned test `tests/updater-branch-guard.test.mjs`.

Re-verified at the current head. The change since my last look is a real edit (+229/-38, not a rename): the test grew identity-snapshot cases, shell-quoting cases, and a git-2.35 stash fallback alongside the updater fixes, with the implementation co-evolving. It still drives the real gate: tests 1 and 2 call `resolveUpdateCommitBranch()` against throwaway repos through the gitIn seam, the exact function `apply()` gates its commit on, so a mis-wired caller cannot pass. Both directions are pinned (non-default branch refuses the commit, default proceeds), plus opt-in, detached-HEAD, unknown-default, origin/HEAD-over-name, and that `--force` does not imply opt-in. I mutation-checked it: gutting the guard to always-commit reddens four assertions and gutting it to withhold-on-default reddens two, and restoring greens them. Nineteen of nineteen pass at head (one by-design soft warn on a Windows-git-unreachable path).

Scott-Emberson (commented):
Test-axis only. The substance is in `update-system.mjs` (the updater), a red-line file and the maintainer's call; I am reviewing the owned test `tests/updater-branch-guard.test.mjs`, and since this guards against committing an auto-update onto the wrong branch I verified it strictly.

It is sound, and notably it drives the real path through an exported seam rather than reconstructing the check in the test body. The test imports and calls the real `resolveUpdateCommitBranch`, and `apply()` calls the same function and gates the commit on its result, so the test exercises the exact entry point the caller uses. (The module comment states this intent explicitly: exported so a test drives the same entry point rather than re-deriving the branch reads.) I mutation-checked the guard: reverting it so an update would commit onto a non-default branch flips the suite from 21 passed to 17 passed / 4 failed, including "feature branch: update commit is withheld". Not a false-pass.

Root pinning is clean, not the #3370 shape: every case builds a throwaway repo with `mkdtempSync` and passes that explicit dir to the branch and staging helpers, with no bare ROOT-default call or reliance on the live repo HEAD.

Scott-Emberson (commented):
Re-verified after the head moved. This is a rebase: the two red-line files, `tests/updater-branch-guard.test.mjs` and `update-system.mjs`, are byte-identical to the previously-verified head, so the earlier verdict carries. The only new content is a doc-only change to `modes/update.md` (already on main), adding relay instructions for the "update applied but not committed" branch notice plus a do-not-trim invariant; it documents and reinforces the guard's user-facing contract and touches no code.

I re-confirmed the owned test at the new head: `node test-all.mjs --only updater-branch-guard` passes 21 of 21 (one environmental skip warning). The exported-seam design still holds (the test drives the real `resolveUpdateCommitBranch`, and apply() calls the same function), root pinning via mkdtempSync temp repos is intact, and reverting the guard so an update would commit onto a non-default branch reddens 4 assertions.

Owned test sound, not behind. SIGNAL (update-system.mjs red-line plus the modes/update.md doc), so not surfacing as ready-to-approve; routing and merge are the maintainer's.

PR #3930, opened 2026-09-06
fix(scan): handle Unicode location boundaries and Workday hints

### LINKED ISSUE #3431: Location filter silently drops in-region roles: ASCII-only word boundaries + job title read as location
Two independent defects in `scan.mjs`'s location filter. Both **silently discard in-region postings** — the scanner simply never writes the row, so there is no `skipped_*` status and no counter, and the loss is invisible in `data/scan-history.tsv`.

Found by re-running an ~1,800-row `scan-history.tsv` ledger back through `buildLocationFilter` and asking which surviving rows the filter would now refuse.

### 1. `compileLocationKeyword` anchors with ASCII-only lookarounds

```js
const prefix = startsWord ? '(?<![a-z0-9])' : '';
const suffix = endsWord   ? '(?![a-z0-9])'  : '';
```

`[a-z0-9]` contains no accented letter, so every accented character reads as a word boundary and a keyword can anchor **mid-word**:

```js
buildLocationFilter({ allow: ['United States', 'al,'] })('montréal, quebec, can', '', '')
// => true   ("al," matches inside "montré|al,"; the "é" is outside [a-z0-9])
```

This fails in the dangerous direction: a stray match on an **`allow`** keyword admits a location the config does not allow. Every non-ASCII place name is exposed — Montréal, Bogotá, Málaga, München, São Paulo.

It is easy to miss because `block` is evaluated before `allow`, so a config that also blocks the country rejects the string for the right reason and hides it.

### 2. `locationHintFromUrl` reads a job title as a location

Workday emits two URL shapes, and only one carries a location after `/job/`:

| shape | segment after `/job/` |
|---|---|
| `/job/{Location}/{Title}_{ReqId}` | the location |
| `/job/{Title}_{ReqId}` | the **title** |

On the second shape the hint becomes the job title:

```js
locationHintFromUrl('https://acme.wd12.myworkdayjobs.com/careers/job/Scrum-Master---Technical-Project-Manager_R0073509')
// => "scrum master technical project manager r0073509"
```

A job title matches no `allow` keyword, but it **is** a non-empty hint, so it defeats `buildLocationFilter`'s escape hatch:

```js
if (lower === '' && hint === '') return true;   // never taken
```

A posting whose location field is empty is then rejected on the strength of its own title. In the sample ledger this accounted for several in-region roles from tenants that use the title-only shape.

### Suggested fixes

1. Use Unicode-aware boundaries — `\p{L}\p{N}` with the `u` flag.
2. Return `''` when the segment after `/job/` is the last segment (the location shape always has the title after it).

PR with both fixes and a regression test that fails on `main` in three places to follow.

nikitacometa:
## What does this PR do?

Prevents location aliases such as `al,` and the built-in Alabama abbreviation from matching inside `Montréal`. Unicode letters, combining marks, and numbers now form word boundaries consistently across all location tiers.

Workday title-only `/job/{Title}_{ReqId}` URLs no longer manufacture a location hint, preserving the existing missing-location fallback.

Fixes #3431.

## Validation

- `node test-all.mjs`: 8709 passed, 0 failed; the same 16 warnings as the clean-upstream baseline.
- 82 focused checks cover Unicode, tier precedence, real US-state controls, and Workday path shapes. Restoring each old matcher/URL guard makes regressions fail.

## Checklist

- [x] Full test suite passes.


Scott-Emberson (commented):
Test-axis only. The substance is in `scan.mjs`, a red-line file and the maintainer's call; I am reviewing the owned test `tests/location-filter-unicode.test.mjs`.

It is sound and mutation-proven. I ran the test against the pre-fix ASCII-only version of the location filter (the natural regression, `[a-z0-9]` lookarounds without the `/u` flag) and it reddens 35 assertions, including the headline false match (`al` must not admit `Montréal`), the decomposed `Montréal`, CJK `東京都`/`新東京`, the astral `𐐀`, an Arabic-Indic digit, and the title-only Workday hint cases; against the PR's fixed version (`\p{L}\p{M}\p{N}` plus `/u`) there are zero failures. It is not a tautology: the assertions go through the public `buildLocationFilter`/`locationHintFromUrl` with hardcoded literal expected values, pinning both directions (a real Unicode city matches standalone and does not false-match embedded across a Unicode boundary) across all four filter tiers and the USPS-abbreviation matcher.

Scott-Emberson (commented):
Test-axis only; the substance is scan.mjs (red-line, the maintainer's call). Re-verified at the current head: tests/location-filter-unicode.test.mjs is byte-identical to the head I last verified sound, so this is a rebase that does not touch it and my verdict carries: it is mutation-proven (the pre-fix ASCII-only boundary reddens 35 assertions incl. the Montréal false-match, decomposed/CJK/astral/Arabic-Indic and the Workday hints; the fixed version passes clean), non-tautological via the public filter API with literal expecteds. The branch is now up to date with main. Routing and merge remain the maintainer's call.

PR #3931, opened 2026-09-06
fix(tracker): recognize localized headers across Node, web, and Go

### LINKED ISSUE #3704: tracker-parse: localized tracker headers shipped by modes/{de,pl,pt,da,id} resolve to null and fall back to the legacy fixed layout

nikitacometa:
## What does this PR do?

Recognizes the tracker headers shipped by the German, Polish, Portuguese, Danish, and Indonesian modes, so inserted or reordered columns stay aligned in Node, web, and Go.

Synchronizes Go's aliases with the shared JSON and prevents reordered headers or Markdown separators from appearing as applications. Separators work with or without a closing pipe.

Fixes #3704.

## Validation

- `node test-all.mjs`: 8652 passed, 0 failed; the same 16 warnings as clean upstream.
- Web: 505 tests, typecheck, and production build pass. Go suite and race detector pass.
- Real merge/status roundtrips across five locales, full alias parity, and targeted mutants cover the regressions.

## Checklist

- [x] Full test suite passes.


Scott-Emberson (commented):
Verified the owned tests. This makes localized tracker headers resolve consistently across the Node parser, the web layer, and the Go dashboard, and both owned tests pin it soundly.

On the dashboard side (our owned area), `dashboard/internal/data/tracker_aliases_test.go` does a `reflect.DeepEqual` of the Go alias literal against `tracker-aliases.json`, so any Go/JSON drift (a dropped, added, or remapped alias) reddens, and I confirmed the two are byte-for-byte the same 33-entry set, so Go and Node/web provably agree. The semantic pinning is in `TestLocalizedTrackerHeaders`: per-locale fixtures (de, pl, pt, da, id) assert `detectTrackerColumns` is non-nil (no fixed-position fallback) plus exact field values, and the reordered Company/Status-first layout is the real mutation-killer since a mis-mapped or dropped alias yields nil columns or a misaligned Company. The write path round-trips and asserts only Status and Notes changed, and counter-tests block the trivial treat-everything-as-a-header fix.

The also-owned `tests/tracker-localized-headers.test.mjs` mirrors the same per-locale matrix against `tracker-parse.mjs` and drives the real `merge-tracker.mjs` and `set-status.mjs` write paths, turning any throw into a failure with no silent skips, and its fixtures self-guard by asserting the header matches the shipped modes verbatim. One minor non-blocking gap: the `applylink` and `followup` aliases are covered only by the parity `DeepEqual`, not a semantic per-locale assertion, but parity guards them across all three sides.

Owned tests verified sound, CI green, mergeable and not behind. The dashboard portion is GATE-tier. Holding formal approval for Scott's go-ahead; surfacing it as ready-to-approve.

Scott-Emberson (commented):
Re-verified at the current head. The apparent owned-test change is a pure rename (tests/tracker-localized-headers.test.mjs to tests/tracker-parse-localized-headers.test.mjs, zero additions or deletions, body byte-identical), so my prior verification carries: the Node test still resolves per-locale headers through tracker-parse.mjs and drives the real merge-tracker.mjs and set-status.mjs write paths asserting only Status and Notes change, with the shipped-header self-guard live, and the Go DeepEqual parity between the alias literal and tracker-aliases.json still holds (both unchanged).

Owned tests verified sound at this head. The dashboard portion is GATE-tier, but the branch is now behind main, so I am holding formal approval until it is rebased and green, then I will re-confirm at the final head.

Scott-Emberson (commented):
Re-verified at the current head. Both owned tests are byte-identical to the head I last verified sound (blob-compared): the Go tracker_aliases_test.go DeepEqual parity between the alias literal and tracker-aliases.json, and the Node localized-headers test resolving per-locale headers through tracker-parse.mjs while driving the real merge-tracker and set-status write paths. This head move is a rebase that touches neither. CI is green and the branch is now up to date with main.

The dashboard portion is GATE-tier. Holding formal approval for Scott's go-ahead; surfacing it as ready-to-approve at this head.

PR #3932, opened 2026-09-06
fix(dashboard): target the correct tracker file, row, and status cell

nikitacometa:
## What does this PR do?

Makes TUI status updates use the same tracker path as the reader, including `CAREER_OPS_TRACKER`. Resolves one row through its Report column and changes only its mapped Status cell: a `[7]` reference in Notes or a stale `Applied` snapshot can no longer redirect the write.

Only canonical new status names and complete single-link Report cells are accepted. A target ID anywhere in a malformed Report cell blocks the update, including reversed link order. Duplicate/missing identities, malformed target cells, and notes containing table delimiters fail without changing the file. Existing locking and atomic replacement remain in place; tab-formatted rows and explicit empty Notes retain their delimiters.

## Validation

- `node test-all.mjs`: 8626 passed, 0 failed; the same 16 warnings as clean upstream.
- `go test ./...` and `go test -race ./internal/data` pass.
- Regression fixtures cover tracker overrides, decoy files/rows/cells, stale status, malformed input, empty/truncated Notes, permissive status aliases, and multi-link reports. Nine targeted mutants fail; legacy disk statuses and UI `Skip` remain supported.

## Checklist

- [x] Full test suite passes.


Scott-Emberson (commented):
Verified the owned Go test `dashboard/internal/data/status_target_test.go`. This is a dashboard change (our owned area) fixing a data-corruption bug (a status update hitting the wrong file, row, or cell), and the test pins all three failure modes, mutation-proven.

The reader and writer both go through `resolveTrackerPath`, and the test asserts full-file bytes plus untouched decoy files, so it is not a tautology. I confirmed each fix by reverting it: forcing `statusCellIndex` to return the first canonical cell anywhere reddens the stale-status-never-matches-another-cell test (the exact Company "Applied" to "Interview" corruption, across pipe, tabs, and status-last layouts); restoring the old `strings.Contains(line, "[N]")` match reddens the notes-reference case (a `[7]` in another row's Notes no longer isolates the real report-column row); and forcing the writer to the legacy root path (reader/writer divergence) reddens the reader-path test on both the canonical and decoy files. The table tests are real subtests over layouts and formats, and ambiguous, missing, or malformed identity and status inputs all fail without a write.

Owned test verified sound, CI green, mergeable and not behind. This is GATE-tier (dashboard/). Holding formal approval until Scott gives the go-ahead; surfacing it as ready-to-approve.

Scott-Emberson (commented):
Re-verified at the current head. The two commits since my last look strengthen the fix and weaken nothing: the write-path status validation moved from the permissive alias check to an exact canonical-name check (a status carrying `|`, CR, LF, or tab can no longer equal a canonical name, so that guard is subsumed), and the report-cell match moved from first-link-only to all-matches with a malformed guard, so a target `[N]` present in a cell that is not exactly one link now errors instead of silently redirecting the write. All three failure modes remain pinned: the wrong-cell stale-status case, the wrong-row notes-reference case, and the wrong-file reader-path case across six layouts with untouched decoys, plus a new malformed-cell-not-first guard (asserts no write) and a legacy-disk-read versus write-reject split. The owned test is real table-driven Go with whole-file byte comparisons; nothing was removed or loosened.

Owned test verified sound at this head. This is GATE-tier (dashboard/), but the branch is now behind main, so I am holding formal approval until it is rebased and green, then I will re-confirm at the final head and surface it as ready-to-approve.

Scott-Emberson (commented):
Re-verified at the current head. The owned Go test dashboard/internal/data/status_target_test.go is byte-identical to the head I last verified sound (blob-compared), so this head move is a rebase onto main that does not touch it, and my verdict carries: all three failure modes (wrong cell, wrong row, wrong file) stay pinned with real table-driven Go and whole-file byte comparisons, plus the malformed-report-cell and legacy-read-vs-write-reject guards. CI is green and the branch is now up to date with main.

This is GATE-tier (dashboard/). Holding formal approval for Scott's go-ahead; surfacing it as ready-to-approve at this head.

PR #3934, opened 2026-09-06
fix(updater): stop apply() deleting a fork's own files under system path prefixes

Meng991003:
## The bug

`config/local-paths.txt` (#2421) tells a fork "list anything you keep in your own fork ... `update-system.mjs apply` will never write to it." Three code paths in `apply()` never consulted it, so a fork's own provider, skill or doc **is deleted** by the stale-file prune.

The mechanism: a fork-local file under a SYSTEM_PATHS *directory* prefix (`providers/`, `.claude/skills/`, `docs/`, `plugins-registry/`) is absent from the fetched tree, so `staleSystemFiles()` reads it as a system file upstream has dropped and unlinks it.

What made it hard to spot is that such files are usually spared by accident. `locallyModifiedSystemFiles()` puts anything differing from the updater baseline into `preservedPaths`, and the prune excludes those. So a file with uncommitted edits survives — and **committing your work is what makes it deletable.**

### Reproduced on a real fork

16 files under `providers/`, `.claude/skills/`, `docs/`, `lib/` and `plugins-registry/`, all absent upstream, all authored in the fork. Running the real `staleSystemFiles(...)` against a real `FETCH_HEAD`:

| | files deleted by `apply` |
|---|---|
| before | **14** |
| after | **0** |

With `preservedPaths` forced empty (i.e. everything committed and clean — the state that removes the accidental protection), it is still 0 after the change.

## The fix

**1. `localUserPaths()` warns instead of refusing on a SYSTEM_PATHS collision.**

The refusal assumed a SYSTEM_PATHS match proves upstream ships the path. It doesn't:

- a directory entry is a wildcard — `providers/` covers upstream's `greenhouse.mjs` and a provider that exists only in a fork, identically;
- even an exact entry can be stale, naming a file the fetched tree no longer has. The checkout loop already tolerates precisely this, calling it "a path genuinely absent upstream".

So the refusal fired on exactly the files this mechanism exists for and left them undeclarable. Answering "does upstream ship this?" needs the fetched tree, which `localUserPaths()` can't reach — `validate-system-paths-coverage.mjs` calls it offline. The original rationale was about *silence*, not refusal, so a warning carries it: fork a genuinely shipped file and you're told it will stop updating.

**2. The prune reads `effectiveUserPaths()` instead of `USER_PATHS`** — the helper that already existed and was already used by `userLayerViolations()`.

**3. The checkout excludes declared paths too**, unconditionally and outside the `--force` branch. `--force` means "discard my local edits to system files", not "discard the files I told you are mine". This one only bites for a declared path upstream *does* ship; a path absent upstream is unreachable by the checkout anyway.

## Tests

`tests/updater-local-paths.test.mjs` #6 and #7 pinned the old refusal and are rewritten to the new contract (honored **and** warned, naming the path). Added **#7b**, the actual regression test: a declared fork-local file survives the prune while an undeclared stale one is still removed.

`node test-all.mjs` on this branch: **8626 passed, 0 failed.**

Happy to split the three changes into separate commits, or to add the `warn` line to the docs in `config/local-paths.example.txt`, if you'd prefer.

## Summary

`update-system.mjs apply` now protects declared fork-owned files under system paths.

- `update-system.mjs:577-662`: `localUserPaths()` warns about collisions with `SYSTEM_PATHS` and keeps the paths user-owned.
- `update-system.mjs:2276-2298`: Checkout skips declared user paths, including with `--force`.
- `update-system.mjs:2340-2346`: Stale-file pruning uses `effectiveUserPaths()`.
- `tests/updater-local-paths.test.mjs:122-169`: Tests warnings, preservation of declared files, and removal of undeclared stale files.

Users can declare fork-owned files under `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/` without `apply` deleting or overwriting them. Undeclared stale system files remain removable.

The test suite reports 8,626 passed tests and 0 failed tests.

### DIFF
--- a/tests/updater-local-paths.test.mjs
+++ b/tests/updater-local-paths.test.mjs
@@ -9,8 +9,8 @@
  * moves that statement OUT of the system layer.
  *
  * What is pinned here is the property that makes the feature safe to ship:
- * absent file changes nothing, and a declaration that would silently stop a
- * system file from updating is refused instead of honored.
+ * absent file changes nothing, a declaration is honored, and one that would
+ * stop a system file from updating says so out loud instead of doing it quietly.
  */
 
 import { mkdtempSync, mkdirSync, writeFileSync, copyFileSync, rmSync } from 'fs';
@@ -25,6 +25,7 @@ import {
   localUserPaths,
   effectiveUserPaths,
   userLayerViolations,
+  staleSystemFiles,
 } from '../update-system.mjs';
 
 /** A throwaway root with an optional declaration file already written. */
@@ -106,36 +107,66 @@ console.log('\n🧪 Local user-paths declaration file (#2421)\n');
   }
 }
 
-// ── 6. A SYSTEM_PATHS collision is refused, loudly, naming the path ──
-//    Honoring it would be worse than refusing: the user would stop receiving
-//    updates for a system file and get no signal that it happened.
-{
-  let threw = null;
+/** Run `fn` with console.error captured, returning [result, capturedText]. */
+function withStderr(fn) {
+  const original = console.error;
+  const lines = [];
+  console.error = (...args) => lines.push(args.join(' '));
   try {
-    localUserPaths(root('merge-tracker.mjs\n'));
-  } catch (err) {
-    threw = err;
-  }
-  if (threw && threw.message.includes('merge-tracker.mjs')) {
-    pass('declaring a SYSTEM_PATHS entry throws and names the path');
-  } else {
-    fail(`#6 expected a throw naming merge-tracker.mjs, got ${threw ? threw.message : 'no throw'}`);
+    return [fn(), lines.join('\n')];
+  } finally {
+    console.error = original;
   }
 }
 
-// ── 7. A collision inside a SYSTEM_PATHS *directory* is refused too ──
-//    'modes/' is a directory entry; modes/pdf.md is shipped by it.
+// ── 6. A SYSTEM_PATHS collision is HONORED, and warns, naming the path ──
+//    This refused until a fork hit the real failure: refusing assumed a
+//    SYSTEM_PATHS match proves upstream ships the path, and it does not. An
+//    exact entry can be stale (naming a file the fetched tree no longer has),
+//    so the refusal fired on a fork's OWN file and left it undeclarable — after
+//    which the stale-file prune in apply() deleted it as a dropped system file.
+//    The rationale was about silence, not refusal, so the warning carries it.
 {
-  let threw = null;
-  try {
-    localUserPaths(root('modes/pdf.md\n'));
-  } catch (err) {
-    threw = err;
-  }
-  if (threw && threw.message.includes('modes/pdf.md')) {
-    pass('a file inside a system directory is refused too');
+  const [declared, stderr] = withStderr(() => localUserPaths(root('merge-tracker.mjs\n')));
+  if (declared.includes('merge-tracker.mjs') && stderr.includes('merge-tracker.mjs')) {
+    pass('declaring a SYSTEM_PATHS entry is honored and warns, naming the path');
   } else {
-    fail(`#7 expected a throw naming modes/pdf.md, got ${threw ? threw.message : 'no throw'}`);
+    fail(`#6 expected it honored + warned, got declared=${JSON.stringify(declared)} stderr=${JSON.stringify(stderr)}`);
+  }
+}
+
+// ── 7. A path inside a SYSTEM_PATHS *directory* is honored and warns too ──
+//    'providers/' is the case that matters: it covers upstream's greenhouse.mjs
+//    and a provider that exists only in this fork, so the prefix cannot tell
+//    them apart. Honoring both is what keeps the fork's file alive; the warning
+//    is what stops a genuine fork-of-a-shipped-file being silent.
+{
+  const [declared, stderr] = withStderr(() => localUserPaths(root('providers/my-own-board.mjs\n')));
+  if (declared.includes('providers/my-own-board.mjs') && stderr.includes('providers/')) {
+    pass('a file inside a system directory is honored and warns');
+  } else {
+    fail(`#7 expected it honored + warned, got declared=${JSON.stringify(declared)} stderr=${JSON.stringify(stderr)}`);
+  }
+}
+
+// ── 7b. The prune must not delete a declared fork-local file ──
+//    The regression this whole change exists for: a fork's provider/skill/doc
+//    under an owned prefix is absent upstream, so staleSystemFiles() saw it as
+//    a system file upstream had dropped. Protection must not depend on the file
+//    having uncommitted edits, which is all that used to spare it.
+{
+  const declaredRoot = root('providers/my-own-board.mjs\n');
+  const [userPaths] = withStderr(() => effectiveUserPaths(declaredRoot));
+  const stale = staleSystemFiles(
+    ['providers/my-own-board.mjs', 'providers/dropped-upstream.mjs'],
+    ['providers/greenhouse.mjs'],          // upstream ships neither of the above
+    ['providers/'],
+    userPaths,
+  );
+  if (!stale.includes('providers/my-own-board.mjs') && stale.includes('providers/dropped-upstream.mjs')) {
+    pass('a declared fork-local file survives the prune; an undeclared stale one still goes');
+  } else {
+    fail(`#7b expected only dropped-upstream.mjs pruned, got ${JSON.stringify(stale)}`);
   }
 }
 
--- a/update-system.mjs
+++ b/update-system.mjs
@@ -582,6 +582,7 @@ export function localUserPaths(root = ROOT) {
   const reject = (path, why) => {
     throw new Error(`${LOCAL_PATHS_FILE}: refusing "${path}" — ${why}`);
   };
+  const underSystemPath = [];
 
   for (const path of declared) {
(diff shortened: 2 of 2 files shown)


Scott-Emberson (changes requested):
Test-axis only. The substance is `update-system.mjs`, a red-line file and the maintainer's call; I am flagging a defect in the owned test `tests/updater-local-paths.test.mjs`, and this is data-loss critical (apply() deleting a fork's own declared files), so it needs a real guard.

Requesting changes: the owned test does not exercise the fix, so the exact regression this PR exists to prevent would ship green.

The whole fix is one line in `apply()` (around line 2346): `mergePathLists(USER_PATHS, preservedPaths)` became `mergePathLists(effectiveUserPaths(), preservedPaths)`. `staleSystemFiles`'s body is unchanged from base. But the owned test (case 7b) never calls `apply()`. It calls the real `staleSystemFiles(...)` directly and hand-supplies the fourth argument as `effectiveUserPaths(declaredRoot)`, which is the very argument whose wiring in `apply()` is the bug.

I mutation-tested it. Gutting the `userPaths` exclusion filter inside the real `staleSystemFiles` reddens the case, so it does drive the real prune predicate (better than a hand-reconstructed filter). But reverting the actual fix (line 2346 back to the buggy `USER_PATHS`) leaves the test fully green, exit 0. So a future revert of that one-line wiring silently reintroduces the fork-file deletion, and the test would not catch it.

To make it load-bearing, drive `apply()` (or assert on the exact `userPaths` argument `apply()` hands to `staleSystemFiles`), so that reverting line 2346 to `USER_PATHS` reddens it. The predicate-level coverage you have (fork file survives, genuine-stale still pruned) is good and worth keeping; it just needs to sit on top of the real call site rather than a supplied argument.

santifer:
@Meng991003: twelve days on a data-loss bug without a word from us, on a first contribution. That one is mine.

The bug is real and your diagnosis holds, including the part that is easy to miss: committing your work is what makes the file deletable, because the only thing sparing it was landing in `preservedPaths` for having uncommitted edits. I reproduced that rather than take it on trust.

On the refusal you turned into a warning: you are right that it fires on exactly the files this mechanism exists for, and I am not asking you to put it back as it was. The rule the updater should keep is narrower. An exact `SYSTEM_PATHS` entry is a claim that upstream ships that file, so a declaration colliding with one should still fail closed. A directory prefix is a wildcard and proves nothing either way, so warning and carrying on is right there. Draw the line between those two and `upgrade-tests.mjs --local-paths`, which is why the Upgrade regression gate is red, follows you instead of blocking you.

Two more, both reproduced here:

- Scott is right. I reverted the `effectiveUserPaths()` line and your test still passes, so the exact regression this PR prevents would ship unguarded. Drive `apply()`, or assert on the argument it hands to `staleSystemFiles`.
- A declared `providers/`, the trailing-slash form your own test accepts as valid, now reaches the checkout as both a path and its own exclusion. `git checkout` exits 1 on that and the error is rethrown, so the whole update aborts.

Nothing else. The measurement and the writeup are better than most changes this file gets.

chipoto69:
github-steward follow-up for the local-paths/data-loss review gate.

I did not push to the contributor branch, approve, or merge. I worked in an isolated steward workspace and pushed a reference branch only:

- Reference branch: chipoto69:steward/pr3934-local-dir-commit-guard-20260920
- Commit: 51456b37e0e3c5f147c351cb1a71af39f32a675f
- Workspace: /Users/rudlord/ORGANIZED/_agent-workspaces/github-steward/career-ops-pr3934-reviewfix-20260920-1435/repo

Patch shape:
- `stagedPathsOutside()` now treats preserved `dir/` entries as prefixes, so a declared local directory such as `providers/` keeps staged child files out of the updater-owned index commit path.
- `apply()` passes `keptFromCheckout` to the staged-index guard instead of only `preservedPaths`, so `config/local-paths.txt` declarations participate in the commit-safety check.
- Added regression coverage in `tests/updater-local-paths.test.mjs` for staged children under a preserved local directory.

Local verification passed:
- `node --check update-system.mjs`
- `node --check tests/updater-local-paths.test.mjs`
- `node tests/updater-local-paths.test.mjs` -> 25 passed, 0 failed
- `git diff --check`
- `npm run lint -- --changed update-system.mjs tests/updater-local-paths.test.mjs` -> 685 .mjs files passed syntax check
- `node test-all.mjs --only updater-local-paths` -> 25 passed, 0 failed

Safe next action: maintainer/contributor can cherry-pick/apply `51456b37e0e3c5f147c351cb1a71af39f32a675f` or equivalent, then trigger the trusted CI matrix and complete normal review. I am leaving the PR owner gate intact.

santifer:
Thanks for the rework @Meng991003: the contract line is drawn where Santiago asked, and the Upgrade regression gate is green again.

Two of the three points are closed at this head. An exact `SYSTEM_PATHS` entry still fails closed, a directory prefix warns and carries on, and `pathFullyPreserved()` keeps a declared `providers/` from reaching the checkout as both a path and its own exclusion.

The third is still open, and it is Scott's point. I reverted your `effectiveUserPaths()` wiring back to `USER_PATHS` and ran the local-paths suite: case 7b still passed. So a later revert of that one line would ship the fork-file deletion again with a green suite. 7b calls `staleSystemFiles()` with an argument it builds itself, and 7c drives `pathFullyPreserved()` through a stub, so neither sits on the real call site. Driving `apply()`, or asserting the exact `userPaths` argument it passes, is what would make it bite.

One more, still open from CodeRabbit's first pass: `apply()` hands the staged-index guard only `preservedPaths`, not `keptFromCheckout`, and that guard matches preserved entries exactly. A staged file under a declared `providers/` is then treated as updater-owned and rides along in the auto-update commit. Same family as the bug you are fixing, so it is worth closing here.

Both next steps are yours. Nothing is approved or merged, and the branch is behind `main`, so the checks will want a rerun once you push.

Meng991003:
Thanks — both closed at `de02bfe`.

## Scott's point

You're right about `tests/updater-local-paths.test.mjs`: I reverted the wiring and ran it, and it is green, 0 failures. 7b supplies the argument and 7c stubs the git call, so neither sits on the call site. No argument from me.

One thing worth flagging before you re-test, though: the behavioural guard does exist, and it does bite. It's in the other suite with the same name.

| with `effectiveUserPaths()` reverted to `USER_PATHS` | |
|---|---|
| `node tests/updater-local-paths.test.mjs` | GREEN, 0 failures |
| `node upgrade-tests.mjs --local-paths` | **RED, 2 failures** |

```
FAIL [local-paths] second apply completes (exit 1)
FAIL [local-paths] the stale-file prune leaves the declared file alone (providers/my-own-board.mjs)
```

That's the leg I added last round. It drives `apply()` for real on an unpoisoned mirror, twice — the second apply is the one that matters, because the first writes the `chore: auto-update system files` commit that becomes the baseline, and only then does the fork file drop out of `preservedPaths` and meet the prune with nothing but the declaration between them. CI runs it as the "Upgrade regression gate" (`.github/workflows/test.yml:93`), so the revert would have failed the gate rather than shipping green.

Two suites called "local-paths" is a trap I set myself by naming the leg after the unit file. Rather than leave it as a thing you have to know, **7d** now pins the call site in the fast unit file too — a source-pattern assertion, the same technique `updater-migration-tests.mjs` already uses for the other `apply()` call sites. Your revert now fails it:

```
❌ #7d apply()'s staleSystemFiles call must read effectiveUserPaths(); wired=false usesBuiltinOnly=true
```

So it reddens whichever suite anyone happens to run. The expensive behavioural leg stays the real proof; 7d is the cheap tripwire.

## The staged-index guard

Fixed as you described. `stagedPathsOutside()` matched its preserved list exactly — correct for the list's usual source, since `git diff --name-only` and `git ls-files` never emit directories, but wrong once `apply()` folds `config/local-paths.txt` in, where a declaration may be a directory. A staged `providers/my-own-board.mjs` under a declared `providers/` matched nothing preserved, fell through to the owned lookups, and the `providers/` **owned** entry claimed it: bare index commit, fork's staged work committed under the updater's message.

- `stagedPathsOutside()`: preserved `dir/` entries match their children.
- `apply()`: hands the guard `keptFromCheckout`, not `preservedPaths`. A file the updater must not *write* is equally one it must not *commit*.
- Tests 12-13 in `updater-commit-file-modes.test.mjs`, with a negative control, next to the #2337 cases they mirror. Both verified red without the fix.

## Verified at this head

| | |
|---|---|
| `upgrade-tests.mjs --local-paths` | GREEN, 8/8 |
| `upgrade-tests.mjs --pr-gate` | GREEN |
| `upgrade-tests.mjs --canary` | GREEN |
| `updater-migration-tests.mjs` | 388 passed, 0 failed |
| `tests/updater-local-paths.test.mjs` | 25 passed, 0 failed |
| all `tests/updater-*.test.mjs` | clean |
| `validate-system-paths-coverage.mjs` | OK, 1381 files covered |

Full `test-all.mjs` not re-run on this head; that table is the exact scope.

## On the steward comment

For the record: I did not fetch, read or apply anything from `chipoto69:steward/pr3934-local-dir-commit-guard-20260920`. The patch above is written from CodeRabbit's `:2594` description and your comment, so any resemblance is both of us working from the same public description of the defect. Flagging it since provenance on a data-loss path seemed worth being explicit about.

Branch is still behind `main` — happy to rebase or merge it forward, whichever you prefer.

PR #3944, opened 2026-09-06
fix(pipeline): route CV output on cv.output_format instead of the PDF score gate

### LINKED ISSUE #3910: Pipeline ignores cv.output_format and independently renders HTML/PDF
## Code of Conduct

## Existing issues

- [x] I searched existing issues and this has not been reported yet

## What happened?

`/career-ops pipeline` does not honour `config/profile.yml` -> `cv.output_format`.

With:

```yaml
cv:
  output_format: text
```

pipeline independently applies its `auto_pdf_score_threshold` rule and, for a qualifying evaluation, creates tailored CV HTML and PDF artifacts. It can also mark the tracker row `PDF ✅`.

This conflicts with `modes/auto-pipeline.md`, whose Step 3 explicitly routes `text` to `modes/text.md` rather than `modes/pdf.md`. The public pipeline documentation says pipeline invokes auto-pipeline, but `modes/pipeline.md` duplicates a separate score-based PDF gate.

## Steps to reproduce

1. In `config/profile.yml`, set:

   ```yaml
   cv:
     output_format: text
   ```

2. Add a publicly accessible, evaluable posting to `data/pipeline.md`.
3. Run `/career-ops pipeline`.
4. Process an evaluation whose final score is at or above `auto_pdf_score_threshold` (the documented default is 3.0).
5. Observe an HTML/PDF artifact despite `output_format: text`.

## Expected behavior

Pipeline should delegate CV-output selection to auto-pipeline Step 3:

- `text` -> tailored Markdown/handoff only, no HTML or PDF
- `latex` -> LaTeX workflow
- default HTML mode -> existing conditional PDF workflow

`auto_pdf_score_threshold` should not override `cv.output_format: text`.

A regression test or instruction-contract test should cover a qualifying pipeline evaluation with `output_format: text`, asserting that it creates neither CV HTML nor PDF and records `PDF ❌` in the tracker.

## CLI tool

Codex CLI 0.153.4

## OS and Node.js version

macOS 15.7.1 / Node.js v26.3.0

## Output of npm run doctor

Not included: the fault is reproducible from the two mode instruction files and does not require local profile, CV, pipeline, or scan output.

aron-intframe:
## What does this PR do?

`modes/pipeline.md` says it "executes the full auto-pipeline", but it restated its own artifact rule: generate a PDF when the score clears `auto_pdf_score_threshold`. auto-pipeline Step 3 instead routes on `cv.output_format`, so a profile set to `text` still got HTML/PDF artifacts and a `PDF ✅` cell. This hands the artifact choice back to Step 3 and scopes the score gate to the default HTML route.

**Tests:** `tests/pipeline-output-format-parity.test.mjs` pins both files to the same three routes (`latex`→`latex.md`, `text`→`text.md`, default→`pdf.md`) and to the ordering that makes the fix work — the format resolves before the gate applies. On the unpatched tree it fails: `modes/pipeline.md ignores cv.output_format when choosing the CV artifact`. Full suite after the fix: 8691 passed, 0 failed.

**Disclosed gap:** `batch/batch-prompt.md` Step 4 (Path B) still gates on the threshold alone and never reads `cv.output_format`, so the same drift lives there. Left out to keep this to the file the issue names; happy to follow up. The 18 localized `modes/*/pipeline.md` files have drifted separately — that reads as an i18n re-sync.

## Related issue

Fixes #3910

## Type of change

- [x] Bug fix

## Checklist

- [x] Bug fix — exempt from issue-first (and an issue exists anyway)
- [x] My changes respect the Data Contract (`modes/` is system layer)

## Summary

- `modes/pipeline.md:45-48`: Routes CV output by `cv.output_format`.
  - `latex`: Uses `latex.md`.
  - `text`: Uses `text.md` and skips HTML/PDF generation.
  - Default: Uses `pdf.md` and applies `auto_pdf_score_threshold`.
- `tests/pipeline-output-format-parity.test.mjs:1-142`: Verifies route mappings, route ordering, and unconditional HTML/PDF exclusion for `latex` and `text`.
- Users with `cv.output_format: text` no longer receive CV HTML or PDF artifacts or `PDF ✅`.

Touched system files: `modes/` and `tests/`. No changes affect `AGENTS.md`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, or `.github/`.


Scott-Emberson (commented):
Test-axis only. The substance is `modes/pipeline.md`, a red-line file and the maintainer's call; I am reviewing the owned test `tests/pipeline-output-format-parity.test.mjs`.

It is sound. The fix routes CV output on `cv.output_format` (latex to latex.md, text to text.md, default to pdf.md) and demotes the `auto_pdf_score_threshold` gate to a default-route-only fallback. The test pins pipeline.md to contain `cv.output_format`, name all three route files, and place `cv.output_format` before `auto_pdf_score_threshold` (`formatAt < gateAt`), so a revert to score-gate-only prose drops the string and the mode names and loses the ordering, reddening all three assertions. The ordering assertion is the load-bearing one for the bug (a text profile must not be forced into PDF by the score gate), and it fails concretely if the gate precedes format resolution. The cross-file parity assertions (profile.example.yml declaring the formats, the auto-pipeline canonical routing rule) fail on absence, so they cannot silently green a missing fix, and the assertions are on real file content, not a self-referential fixture.

One minor, non-blocking note: the check asserts all three mode strings are present but not that each format maps to its specific file (latex to latex.md), though the score-gate revert still reddens it, so the signal holds.

Scott-Emberson (commented):
Re-verified after the head moved past our earlier test-axis comment. The new commit touches only the owned test `tests/pipeline-output-format-parity.test.mjs`; the substance `modes/pipeline.md` is byte-identical across both commits. There was no force-push, just one commit added on top.

The change strengthens the test rather than weakening it. The old assertions used `pipeline.includes(mode)` (the mode-file name appearing anywhere, which a route swap would slip past); the head version pairs each route cue ("latex", "text", "otherwise") with the first `modes/*.md` that follows it, scopes the pipeline.md check to the span between `cv.output_format` and `auto_pdf_score_threshold`, and adds an assertion that latex/text never produce a PDF whatever the score. I mutation-checked it: forcing all routes to `pdf.md` reddens, and swapping latex/text targets (the exact weakness the head commit closes) reddens where the old test would have stayed green.

One note: this suite uses the shared pass/fail counters, so a bare `node --test <file>` exits 0 even under a failing mutation; the authoritative gate is `test-all.mjs`, which imports it in-process, and my mutation verdicts were taken there.

Owned test sound and now more mutation-sensitive; not behind. SIGNAL (modes/ is the maintainer's red-line), so not surfacing as ready-to-approve; routing and merge are the maintainer's.

Scott-Emberson (commented):
Re-verified after the head moved again (`4f3827849e` to `71cfdc3f85`). The new commit touches only the owned test, and `modes/pipeline.md` is byte-identical to the last-reviewed head, so no maintainer-owned substance changed.

The commit tightens the owned test: the no-artifact guarantee now has to span both HTML and PDF rather than PDF alone. The rationale is sound: the `text` route is markdown-only, so a "never produce a PDF" promise still lets the default route's HTML preview leak to a `text` profile, and a bare "contains HTML" check would pass a "write the HTML but never a PDF" sentence, so the negation itself must cover both. The existing prose in modes/pipeline.md already reads "the latex and text routes never produce HTML or a PDF, whatever the score", so it satisfies the stricter assertion, and the run is green (5 assertions via the real harness). I mutation-checked it: forcing all routes to pdf.md reddens, swapping the latex/text targets reddens, and weakening the guarantee to PDF-only reddens the newly-tightened assertion specifically.

Owned test sound and now stricter, not behind. SIGNAL (modes/ is the maintainer's red-line), so not surfacing as ready-to-approve; routing and merge are the maintainer's.

Scott-Emberson (commented):
Re-verified after the head moved again. The new commit touches only the owned test; `modes/pipeline.md` is byte-identical to the last-reviewed head, so no maintainer-owned substance changed.

The commit strengthens the test further: the no-artifact promise now also has to be score-independent and carry no conditional qualifier (below/under/above/when/if/unless/only/threshold). That closes a gap where a promise like "never produce HTML or a PDF below the threshold" spanned both artifacts and passed the old check yet inverts the fix, since the artifacts return above the gate. The current prose satisfies the stricter assertion and the run is green (5 assertions via the real harness). I mutation-checked it: forcing all routes to pdf.md reddens, swapping the latex/text targets reddens, weakening the guarantee to PDF-only reddens, and making the promise score-conditional reddens the newly-added guard specifically (the old test would have passed it).

Owned test sound and now stricter; branch behind main. SIGNAL (modes/ red-line), so not surfacing as ready-to-approve; routing and merge are the maintainer's.

santifer:
Sorry for the wait, @aron-intframe. Four rounds of tightening on your own test, each one reproduced before you touched anything, and nobody had answered you since 8 Sep. That part is on us.

The diagnosis is right, and I checked it rather than taking it: on current `main`, `modes/pipeline.md` never mentions `cv.output_format`, while `modes/auto-pipeline.md` Step 3 routes on it. So the same offer gives a `text` profile markdown from a single-URL run and a PDF from the inbox path. Your change makes pipeline.md defer to Step 3 and demotes the score gate to what it always was: a narrowing of the default route.

Your suite runs 5 passed, 0 failed on a checkout of your branch over current `main`, and the full suite 9067 passed, 0 failed. Swapping in main's `modes/pipeline.md` turns three of your five rows red, so the test bites on the real bug rather than on its shape.

On CodeRabbit's last note, I am closing it without a change. The suggested regex wants `latex and text` as bare words, but your sentence writes both in backticks, so applying that patch would turn a correct file red. The four layers you already built cover the case it worries about.

The branch is behind `main`, so refreshing it and the merge are on me. Thanks for the mutation work: it made this review short.

santifer (approved):

aron-intframe:
No apology needed; the wait cost nothing here. I re-ran your three checks on the merge commit `1d4993d` rather than taking them, and all three hold.

`git grep -c 'cv.output_format' 1d4993d^ -- modes/pipeline.md modes/auto-pipeline.md` on the parent of the merge returns only

```
1d4993d^:modes/auto-pipeline.md:1
```

so pre-merge `modes/pipeline.md` had zero mentions, exactly as you read it.

Your CodeRabbit call is right, and the reason is the backticks: `latex\s+and\s+text` needs a bare space between `latex` and `and`, but the merged sentence reads `` `latex` and `text` routes ``, so the closing backtick breaks the match. Running its regex against the merged sentence:

```
coderabbit regex on sentence: false
```

And applying its committable suggestion to the test with `modes/pipeline.md` untouched turns the merged file red, as you predicted:

```
  ❌ modes/pipeline.md no longer says the latex/text routes skip both HTML and the PDF regardless of score, so auto_pdf_score_threshold can override cv.output_format (promise is conditional on the score, found instead: never produce HTML or a PDF, whatever the score)
```

Swapping in the pre-merge `modes/pipeline.md` reddens three of five rows on my side too:

```
  ❌ modes/pipeline.md ignores cv.output_format when choosing the CV artifact (missing cv.output_format; "latex" → nothing (expected modes/latex.md); "text" → nothing (expected modes/text.md); default → nothing (expected modes/pdf.md))
  ❌ modes/pipeline.md no longer says the latex/text routes skip both HTML and the PDF regardless of score, so auto_pdf_score_threshold can override cv.output_format (no unconditional promise found)
  ❌ modes/pipeline.md applies the auto_pdf_score_threshold gate before resolving cv.output_format, so a `text` profile can still be forced into a PDF
```

Full suite on `1d4993d`, Node v22.23.1:

```
📊 Results: 9079 passed, 0 failed, 18 warnings
```

(9079 rather than your 9067; the merge commit sits on later `main` than your checkout did. Zero failures either way.)

One small thing the mutation above exposed in the merged test, which I did not catch before: when the route-name check is the part that fails, the diagnostic still prints `promise is conditional on the score`, because the message only branches on `coversBoth`. Diagnostic wording only, the verdict is correct; I will fold a third branch into the follow-up rather than open a PR for one string.

Still open on current `main` (`d8e7407`): `batch/batch-prompt.md` Path B has 0 mentions of `cv.output_format` and gates on `auto_pdf_score_threshold` alone at lines 396 and 460, and 0 of the 18 `modes/*/pipeline.md` files route on it. I found no issue tracking the Path B drift. It is the same bug as #3910 in the inbox path, so it qualifies as a bug fix under CONTRIBUTING; I will open it as its own PR with the same parity test extended to Path B, unless you would rather it wait on the i18n re-sync.

PR #3945, opened 2026-09-06
test(cli-flags): assert every value-taking caller handles a missing operand

### LINKED ISSUE #3194: test: assert that every value-taking flag caller opts into requireOperand
Follow-up to #3087 / #3092 (merged today). `requireOperand` is opt-in: the four callers that lacked a guard now pass it, and the callers with their own better messages keep them. But the hole re-opens silently with every NEW script that calls `validateFlags` with `valueFlags` and forgets the option: nothing fails, the script just inherits the missing-operand bug.

**Proposal**: one repo-wide assertion (test-all style) that finds every `validateFlags(` call site carrying `valueFlags`, and requires each to either (a) pass `requireOperand: true`, or (b) appear in a short documented allowlist of callers with their own guard (currently: scan.mjs, rejection-latency.mjs, scan-ats-full.mjs, check-table-freshness.mjs, contacts.mjs, per the audit in #3092's body). A new caller outside both sets turns CI red with the file named.

**Acceptance**:
- Adding a fresh `validateFlags` caller with `valueFlags` and no `requireOperand` fails CI naming the file.
- The five documented own-guard callers stay green without changes.
- Zero new dependencies.

@Priyans-Lathiya this one is yours first if you want it (as promised in #2611): it's the same guard-the-guard shape as your zh-minimal find. It stays off the good-first-issue shelf for 48h so the offer is real this time; one comment here and it's assigned.

Priyans-Lathiya:
## What does this PR do?

1. **Adds a repo-wide assertion** (`tests/validate-flags-value-operand.test.mjs`, auto-discovered by `test-all.mjs`, zero new dependencies): it sweeps every `validateFlags(` call site in tracked non-test `.mjs` source and fails CI naming the file and line when one declares `valueFlags` without a missing-operand guard. `requireOperand` is opt-in by design — several callers run richer validation of their own — but opt-in re-opens the hole it was added to close, silently: a new script wires up `valueFlags`, forgets the option, nothing fails, and it just inherits the bug. Two callers were already in that state when the tree was last audited by hand.
   1. **The rule is "handled somewhere", not "everyone opts in":** each caller must pass `requireOperand: true` *or* appear in `OWN_GUARD` naming the validator it runs instead. `valueFlags` without `requireOperand` is reported as an observation, never a verdict — it can equally mean a richer local validator, an optional operand (`contacts.mjs --vcf`), or a real hole, and no text scan can tell those apart.
   2. **The allowlist can only shrink:** six entries, each carrying a one-line reason, capped by `MAX_OWN_GUARD` so a seventh needs a second edit visible in the diff, plus a staleness check that drops entries whose caller has since opted in.
   3. **Placement is deliberately not checked:** validating at module top level and inside the main-module guard are both correct, decided by whether anything imports the module. The classifier reads the call's arguments and nothing about its position; five fixtures pin that.
2. **Fixes the two callers the sweep caught.** `upskill.mjs`: `--min-reports --summary` read `--summary` as the count, `parseInt` gave `NaN`, and the aggregate branch silently fell back to 5; `--url-text --help` printed usage at exit 0 before the URL check could run. `audit-portals.mjs` (landed in #3610 mid-review): a trailing `--file`/`--company`/`--baseline` each fell through to a default at exit 0, and a manual `--help` check *before* `validateFlags` meant `--help --bogus` exited 0 without reporting `--bogus`.
3. **Allowlists `fix-slugs.mjs`**, which the #3092 audit missed — it has its own, stronger guard (it also rejects `--file=` and `--file -h`).
4. **Documents `requireOperand` in the `validateFlags` JSDoc**, which had never mentioned the option at all.
5. **Adds 7 behavioural regression tests** to `tests/cli-flag-validation.test.mjs` for the two fixed callers.
6. **Records what breaks the sweep** — in both the test header and next to the `validateFlags` signature — separating the changes that fail loudly (reordered parameters, renamed properties) from the ones that go silently green (shorthand `{ valueFlags }`, a wrapper helper), so the next person changing the helper knows which is which.

## Related issue
Closes #3194 

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

---

## Summary

- Added a repo-wide test to enforce missing-operand handling for `validateFlags` calls with `valueFlags`: `tests/validate-flags-value-operand.test.mjs:1`.
- Added `requireOperand` documentation: `lib/cli-flags.mjs:38`.
- Fixed missing-value handling in `upskill.mjs:68` and `audit-portals.mjs:92`.
- Added seven regression tests for missing operands and help/unknown-flag precedence: `tests/cli-flag-validation.test.mjs:1`.
- Allowlisted callers with documented guards through `OWN_GUARD`.

## User impact

Value-taking flags now reject missing operands instead of silently falling back or bypassing validation. Standalone `--help` still succeeds. New `validateFlags` callers with `valueFlags` must use `requireOperand: true` or a documented guard.

## System files

### DIFF
--- a/audit-portals.mjs
+++ b/audit-portals.mjs
@@ -328,11 +328,19 @@ const USAGE = `Usage:
 
 async function main() {
   const args = process.argv.slice(2);
-  if (hasFlag(args, '--help') || hasFlag(args, '-h')) {
-    console.log(USAGE);
-    return;
-  }
-  validateFlags(args, KNOWN_FLAGS, USAGE, { valueFlags: VALUE_FLAGS });
+  // validateFlags handles --help itself, and handles it LAST — after the
+  // unrecognized-flag and missing-operand checks. The manual pre-check that used
+  // to sit here inverted that order, which is the bug CodeRabbit caught on
+  // #2745/#2746: `--help --bogus` exited 0 having never looked at `--bogus`,
+  // and `--file --help` printed usage instead of reporting the missing operand.
+  //
+  // requireOperand: nothing here rejects a missing operand. A trailing `--file`
+  // fell back to the default portals.yml, a trailing `--company` audited every
+  // board instead of one, and a trailing `--baseline` skipped the comparison
+  // entirely — each of them silently, at exit 0, reporting on inputs nobody
+  // asked for. `--small-threshold` keeps its own SHAPE check below (a
+  // non-negative number); this only adds the presence check it lacked.
+  validateFlags(args, KNOWN_FLAGS, USAGE, { valueFlags: VALUE_FLAGS, requireOperand: true });
 
   const filePath = flagValue(args, '--file') || DEFAULT_PORTALS_PATH;
   const only = flagValue(args, '--company');
--- a/lib/cli-flags.mjs
+++ b/lib/cli-flags.mjs
@@ -77,14 +77,31 @@ export function hasFlag(args, flag) {
  * @param {string[]} knownFlags - Every flag name this script accepts,
  *   leading dashes included (e.g. `['--dry-run', '--help', '-h']`).
  * @param {string} usage - Usage text printed for `--help`/`-h`.
- * @param {{valueFlags?: string[]}} [options] - `valueFlags` lists flags whose
- *   space-separated value is the NEXT argv token (e.g. `--since 7`). Without
- *   this, a value that itself starts with `-` (e.g. `--since -5`, a negative
- *   day count) would be misread as an unrecognized flag rather than left for
- *   the caller's own value validation to reject with a clearer message.
- *   `--flag=value` is self-contained and never needs to be listed.
+ * @param {{valueFlags?: string[], requireOperand?: boolean}} [options] -
+ *   `valueFlags` lists flags whose space-separated value is the NEXT argv token
+ *   (e.g. `--since 7`). Without this, a value that itself starts with `-` (e.g.
+ *   `--since -5`, a negative day count) would be misread as an unrecognized
+ *   flag rather than left for the caller's own value validation to reject with
+ *   a clearer message. `--flag=value` is self-contained and never needs to be
+ *   listed.
+ *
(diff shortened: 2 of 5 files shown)


Scott-Emberson (commented):
Verified the two owned tests by execution and mutation. Together they enforce that every value-taking CLI flag rejects a missing operand, and neither is vacuous.

The enumeration is real, not a hardcoded subset: `validate-flags-value-operand.test.mjs` sweeps the whole tree via `git ls-files`, classifies every `validateFlags(` call, and flags any caller that declares `valueFlags` without `requireOperand: true` (and is not in the own-guard allowlist). I confirmed it by adding a brand-new tracked `zz-newcli.mjs` with `valueFlags` and no guard, which reddened immediately naming the file and line. The backstops are genuine: a `MIN_CALL_SITES` floor (23 matched today against a floor of 12) plus a staleness check on the known files defeat a scanner that silently stops matching, and 14 classifier fixtures plus placement and prose fixtures pin the parser, with opaque and unterminated cases failing closed.

Each caller actually rejects a missing operand: `cli-flag-validation.test.mjs` spawns the real scripts and asserts both `status === 1` and `/--X requires a value/`, with a help-alone case asserting exit 0 and usage to guard against over-rejection. I mutation-checked both layers: stripping `requireOperand` from `upskill.mjs` reddened the static sweep and the two upskill runtime tests, and neutering `requireOperand` in `lib/cli-flags.mjs` (the regression to consuming the next token) reddened fourteen behavioral tests across doctor, detect-reposts, process-quality, weekly-digest, upskill, and audit-portals. The two layers are complementary and honest about scope: the static sweep proves the guard is wired on every caller, the behavioral tests prove it actually rejects.

Owned tests verified sound, CI green, mergeable and not behind. This is GATE-tier (the test harness). Holding formal approval until Scott gives the go-ahead; surfacing it as ready-to-approve.

Priyans-Lathiya:
Hi, 
please let me know if any CodeRabbit suggestions need to be addressed or if there is anything pending from my side.

Issue #958, opened 2026-06-12
update-system.mjs apply silently drops newly-introduced files (uses old SYSTEM_PATHS manifest for self-update)


Narash25:
## Summary
Updating **1.8.1 → 1.10.0** via `node update-system.mjs apply` reported success (version bumped, `Updated 59 system paths`) but **silently failed to install 26 new system-layer files** introduced in 1.10.0. The result is a broken install: `merge-tracker.mjs` imports `./tracker-links.mjs`, which was never copied, so it crashes on every run.

## Root cause
`apply` checks out only the paths in the **local (pre-update) version's** hardcoded `SYSTEM_PATHS` array. Files newly added in the target version aren't in the old array, and the `BOOTSTRAP_PATHS` escape hatch doesn't cover them, so they're skipped. The new `update-system.mjs` *does* list them — but it isn't in effect until after the very apply that was supposed to install everything (self-updater chicken-and-egg).

**Suggested fix:** apply using the **new** version's manifest — e.g. re-exec `update-system.mjs` after it updates itself, or derive the file list from `FETCH_HEAD` rather than the local array.

## Files skipped (26 total, examples)
- `tracker-links.mjs` — imported by `merge-tracker.mjs`, so its absence crashes merge on every run
- `generate-cover-letter.mjs` + `modes/cover.md` — cover-letter feature
- `modes/ar/*` — Arabic / RTL feature
- `validate-portals.mjs`, `reserve-report-num.mjs`, `scan-ats-full.mjs`, `modes/interview.md`
- `Dockerfile`, `docker-compose.yml`, `.dockerignore`, `DOCKER.md`
- `scaffolder/*`, several README translations (`README.ar.md`, `README.fr.md`, `README.pl.md`)

## Repro
1. From a 1.8.x checkout: `node update-system.mjs apply` (to 1.10.0)
2. `node merge-tracker.mjs --dry-run` → crashes: `Cannot find module '.../tracker-links.mjs'`
3. List files present upstream but never copied:
   ```
   git ls-tree -r --name-only FETCH_HEAD | while read f; do [ -e "$f" ] || echo "$f"; done
   ```
   → returns the 26 missing paths.

## Also observed (separate, non-fatal)
apply logs: `error: pathspec '.gemini/commands/' did not match any file(s) known to git`

## Workaround
`git checkout FETCH_HEAD -- <the 26 paths>` completes the install.

santifer:
[...]

PR #3947, opened 2026-09-06
fix(update-system): refuse manifest entries that name the user layer

### LINKED ISSUE #3946: update-system: a manifest entry naming a user directory deletes an untracked user file, while reporting "your content was NOT overwritten"
A `SYSTEM_PATHS` entry naming a **directory** that covers user content makes `apply()` destroy an untracked user file and then report the opposite. This is the shape a manifest edit produces by broadening an existing entry — e.g. `documents/README.md` → `documents/`.

## What happens

1. `:2252` — `git checkout FETCH_HEAD -- <dir>` writes upstream's files over the user's.
2. `:2220` — the #2337 local-edit detector is diff-based, so an **untracked** user file whose path is absent from the install's git history is invisible to it. No `.bak`, not added to `preservedPaths`.
3. `:2410` — `userLayerViolations` then flags the file and the update aborts.
4. `:2443` — `revertPaths` deletes it as an addition `HEAD` lacks (#2015).
5. `:2452` — the run prints `User file(s) left as-is (your content was NOT overwritten)`, naming a file that no longer exists.

The intent at `:2445-2451` is explicit and correct — *"never `git checkout` the violated user paths back to HEAD... leaving them as-is is the safe choice"*. The untracked case defeats it: the content is gone before that decision is reached, and the message asserts otherwise.

Observed output:

```
SAFETY VIOLATION: User file was modified: documents/NOTES.md
Aborting: user files were touched. Rolling back system files...
User file(s) left as-is (your content was NOT overwritten):
  documents/NOTES.md
Update aborted: user files were touched.
### AFTER: cat: documents/NOTES.md: No such file or directory
```

## Preconditions

All three are required:

1. the manifest names a directory that covers user content
2. upstream ships a file inside it
3. the user has a file at that same path which is untracked **and** whose path is absent from the install's git history

If the path *is* in the install's history, `locallyModifiedSystemFiles` sees it, writes a `.bak` and preserves it — that case is already safe. The loss is specific to (3).

## Steps to reproduce

Run from a checkout of this repo (writes only inside the workdir you pass):

```bash
#!/usr/bin/env bash
set -euo pipefail
ROOT="${1:?usage: repro.sh <empty-workdir>}"
SRC="${SRC:-$PWD}"                 # a career-ops checkout
rm -rf "$ROOT"; mkdir -p "$ROOT"

# upstream: a clone carrying the manifest mistake
git clone -q --no-hardlinks "$SRC" "$ROOT/upstream"
cd "$ROOT/upstream"; git checkout -q -B main
git config user.email t@e.com; git config user.name T; git config commit.gpgsign false
git -C "$SRC" show origin/main:update-system.mjs > update-system.mjs
python3 - "$ROOT/upstream" <<'PY'
import sys, re
p='update-system.mjs'; s=open(p).read()
# THE MISTAKE: broaden a file entry to the user directory holding it
assert "  'documents/README.md',\n" in s
s = s.replace("  'documents/README.md',\n", "  'documents/',\n", 1)
# keep the self-bootstrap inside the sandbox instead of hitting real GitHub
s = re.sub(r"const CANONICAL_REPO = '[^']*';",
           "const CANONICAL_REPO = '%s';" % sys.argv[1], s, count=1)
open(p,'w').write(s)
PY
echo "9.9.9" > VERSION
git commit -qam "upstream: broaden documents/README.md -> documents/"

# install: an older release cloned from that upstream
git clone -q --no-hardlinks "$ROOT/upstream" "$ROOT/install"
cd "$ROOT/install"
git config user.email t@e.com; git config user.name T; git config commit.gpgsign false
echo "1.0.0" > VERSION; git commit -qam "install: older version"

# the user's own file: untracked, gitignored, unknown to this install's history
echo "MY PRIVATE NOTES - irreplaceable" > documents/NOTES.md
echo "### BEFORE: $(cat documents/NOTES.md)"

# upstream later ships a file at the SAME path
cd "$ROOT/upstream"; echo "upstream doc" > documents/NOTES.md
git add -f documents/NOTES.md; git commit -qm "upstream: ship documents/NOTES.md"

# the user updates
cd "$ROOT/install"; node update-system.mjs apply --confirm 2>&1 || true
echo "### AFTER:"; cat documents/NOTES.md 2>&1 || true
```

## Expected behavior

A manifest entry that covers user content should be refused before the checkout, and the user's file left intact. Failing that, the abort path should not report content as preserved when it was deleted.

## Severity, stated plainly

This has never fired and cannot be triggered by a third party — `CANONICAL_REPO` is a hardcoded const with no env override, so the only route in is a mistake in the manifest. It is worth fixing anyway because the manifest is edited constantly and has been got wrong before (#958, #1002, #2532), and because when it does fire the loss is unrecoverable: `.bak`, the backup branch (`:2059`) and `git stash create` (`:2050`) are all diff- or commit-based, and the user layer is gitignored and untracked.

## One thing worth knowing for any fix

`apply()` self-bootstraps — it checks the fetched `update-system.mjs` out and re-execs it before the merge — so the `SYSTEM_PATHS` constant in the running file is upstream's list, not the install's. A guard that filters only `remoteSystemPaths` therefore reads as protection while the identical entry walks in through the "local" half. I found this by writing that version first and watching the repro still destroy the file. Any guard has to wrap the merged manifest.

PR follows.

nikolaysm:
## What does this PR do?

Refuses system-manifest entries that name the user layer, before `apply()` checks anything out.

A manifest entry naming user territory makes the per-path `git checkout FETCH_HEAD -- <path>` write upstream's content over the user's own. When the target is an **untracked** file, nothing recovers it: the #2337 detector is diff-based and never sees it, so no `.bak` is written, `git stash create` captures nothing, and the backup branch holds only committed state. The abort path then deletes it as an addition `HEAD` lacks, while printing `your content was NOT overwritten`. Reproduction script in #3946.

## Related issue

Fixes #3946

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] If this is a new feature or architecture change, I opened an issue first
- [ ] I ran `node test-all.mjs` and all tests pass — ran it; left unticked deliberately because the suite is not green. The failures are pre-existing and unrelated; see "Tests" below.
- [x] My changes respect the [Data Contract](https://github.com/career-ops-hq/career-ops/blob/main/DATA_CONTRACT.md)

---

## What the guard refuses

Comparison is on path **segments**, treating a trailing slash as a spelling rather than a fact — `git checkout <ref> -- documents` and `-- documents/` name one tree.

- an entry equal to a declared user path, or an **ancestor** of one (`modes` would claim `modes/_profile.md`)
- a **subtree claim** inside user territory — open-ended, since upstream decides its contents in this release and every later one. Directory-ness is read from the tree being checked out, not from the slash.
- a **single file** inside a user directory that would land on a file this install does not track. Keyed to recoverability rather than intent, because `writing-samples/README.md` and `interview-prep/story-bank.md` are indistinguishable by shape, and untracked-and-present is exactly the state nothing can undo.
- **malformed spellings**: `.` / `..` / empty segments, backslashes, a leading `/`, git pathspec magic (a leading `:`, and the wildcards `*` `?` `[` that carry the same power without announcing it — a default pathspec wildcard matches `/`, so `modes/*` claims `modes/_profile.md`), and NUL bytes (which `child_process` rejects with `ERR_INVALID_ARG_VALUE` before git starts, killing `apply()` on an opaque error).

A **tracked** file inside a user directory stays allowed. Refusing those would stop new upstream docs from ever arriving — the #958 silent non-arrival. All 342 entries the real manifest ships pass unchanged.

Refusal is loud but non-fatal, and refused entries are subtracted from the completeness check — otherwise a single manifest mistake would make `apply()` exit 1 on every subsequent run, forever.

## Review notes

**The filter wraps the merged manifest, not `remoteSystemPaths`.** `apply()` self-bootstraps into the fetched updater before this point, so the local `SYSTEM_PATHS` constant is upstream's list too. I wrote the fetched-half-only version first and the repro still destroyed the file; filtering one half lets the identical entry in through the other.

**`manifestProbes()` was extracted rather than left inline.** Review asked instead for larger regexes proving the probes are consumed. A regex cannot do dataflow analysis, and I verified the gap rather than assuming it: gutting both probes to constants disabled the entire named-file half of the guard while every structural check still passed. The seam follows the same reasoning `userLayerViolations()` gives for being pure, and the probes' behaviour is now tested directly instead of pattern-matched.

**I relaxed a pre-existing assertion.** The #1998 structural check pinned `missingFromTargetManifest(remoteSystemPaths)` literally; the new filter wraps that argument, so it now pins `missingFromTargetManifest(\s*remoteSystemPaths` — still the property that matters (the *target* manifest is verified, not the local one), with the subtraction pinned by a separate check. Flagging it because it is someone else's guard, not mine.

**The canary, and what the fix costs.** `upgrade-tests.mjs --canary` went red here *because* the guard works: its poison adds `cv.md` to `SYSTEM_PATHS`, which is now refused, so the harness never sees a byte-level clobber. @chipoto69 diagnosed this and published a patch (`8efa988`), cherry-picked here with authorship preserved, so the canary greens on either detection path.

Worth stating plainly rather than leaving to be discovered: with the guard in place the refusal path **always** wins, so the `clobbered` branch is now unreachable and the `user file byte-identical` oracle is no longer exercised by anything. The gate still proves it can fail, and newly proves the guard fires — but the specific guarantee the canary was built for is gone. If you want it kept, the poison should also strip the guard from the mirrored updater, so both properties stay covered. That is your call on your own safety net, not something I wanted to decide by editing it.

**One suggestion skipped:** renaming the suite to `tests/update-system.test.mjs`. The cited rule (`{module}.test.mjs`) is CONTRIBUTING.md's **web** convention, for suites mirroring `web/src/` paths under `web/tests/`. Root-script suites follow `tests/<subject>.test.mjs` per `.coderabbit.yaml`, and 15 existing `updater-*.test.mjs` files all cover `update-system.mjs`. Renaming would add a 16th under a scheme none of the others use. Happy to change it if you would rather.

**Why the malformed-spelling check has to be the whole defence.** The checkout cannot pass `--literal-pathspecs`: it builds `:(exclude)` specs for preserved paths, so the flag would disable the very magic it depends on. There is no second line here — an entry carrying a wildcard past validation reaches git as a live pathspec. An earlier revision of this description called `--literal-pathspecs` a deliberate hot-path omission; that was wrong, and it is structurally unavailable.

## Tests

`node test-all.mjs` → **8708 passed, 1 failed**. The failure is the `web/` group (`apply-cv-resolver.test.mjs`, `explore-ai-dedup.test.mjs`), which I confirmed fails identically on a clean `origin/main` worktree — pre-existing and untouched by this change.

`node upgrade-tests.mjs --canary` → GREEN (via the refusal path, as described above).

The unit suite drives the real exported `rejectUserLayerPaths` and `manifestProbes` with explicit doubles, so it verifies the rule and the probe behaviour; the structural checks in `updater-migration-tests.mjs` pin the wiring to the real sources. Both refusal and acceptance are asserted, so neither a guard that keeps everything nor one that refuses everything can pass.

## Summary

`update-system` now refuses manifest entries that name or overlap the user layer before checkout. This protects declared user paths, user directories, and untracked files inside user directories.

The filter applies to the merged manifest used by `apply()`. Refused entries are reported and excluded from completeness checks. Tracked system files inside user directories remain allowed.

## User impact

Users can update without losing user-layer files when a manifest entry covers them. Malformed paths and pathspec-like entries are refused. The update continues with the remaining safe entries.

## Files changed

- `update-system.mjs`: Adds `manifestProbes()` and `rejectUserLayerPaths()`; filters the merged manifest.
- `tests/updater-remote-manifest-user-paths.test.mjs`: Adds coverage for overlaps, untracked files, malformed paths, and Git probes.
- `updater-migration-tests.mjs`: Verifies merged-manifest filtering and completeness handling.
- `upgrade-tests.mjs`: Accepts explicit refusal in the upgrade canary.

System files touched: `update-system.mjs`. `AGENTS.md`, `modes/`, `DATA_CONTRACT.md`, `providers/`, and `.github/` are unchanged.

### DIFF
--- a/tests/updater-remote-manifest-user-paths.test.mjs
+++ b/tests/updater-remote-manifest-user-paths.test.mjs
@@ -23,7 +23,7 @@
  */
 
 import { pass, fail } from './helpers.mjs';
-import { rejectUserLayerPaths } from '../update-system.mjs';
+import { rejectUserLayerPaths, manifestProbes } from '../update-system.mjs';
 
 // A stand-in for effectiveUserPaths() covering both declaration forms: a
 // trailing `/` is a directory, anything else is an exact file path.
@@ -266,5 +266,85 @@ console.log('\n🧪 Testing rejectUserLayerPaths (fetched manifest vs local user
     pass('an empty fetched manifest yields nothing on either side');
   } else {
     fail('an empty fetched manifest must yield nothing on either side');
+  }
+}
+
+// manifestProbes() is what apply() actually hands the rule. The blocks above use
+// doubles, so they verify the RULE; these verify the WIRING — that the probes read
+// the git output they are given. Left inline in apply(), this could only be checked
+// by pattern-matching the source, and a source pattern cannot tell
+// `trackedFiles.has(path)` from `() => true`.
+console.log('\n🧪 Testing manifestProbes (git output -> the rule\'s three questions)...');
+
+{
+  // Given: `ls-files -z` output naming two tracked files
+  const probes = manifestProbes({
+    trackedOutput: 'writing-samples/README.md\0documents/README.md\0',
+    upstreamOutput: '',
+  });
+
+  // When: the tracked probe is asked about a listed and an unlisted path
+  // Then: it answers from that output, not from the surrounding checkout
+  if (probes.tracked('writing-samples/README.md') && !probes.tracked('data/applications.md')) {
+    pass('tracked() answers from the ls-files output it was given');
+  } else {
+    fail('tracked() does not read the ls-files output');
+  }
+}
+
+{
+  // Given: a NUL-delimited name that a newline split would mangle. core.quotePath
+  // makes this the realistic shape, and it is why -z is used.
+  const probes = manifestProbes({
+    trackedOutput: 'documents/résumé notes.md\0documents/README.md\0',
+    upstreamOutput: '',
+  });
+
+  // When/Then: the whole name survives as one entry
+  if (probes.tracked('documents/résumé notes.md') && probes.tracked('documents/README.md')) {
+    pass('tracked() parses NUL-delimited names whole');
+  } else {
+    fail('tracked() mangles NUL-delimited names');
+  }
+}
+
+{
+  // Given: an upstream tree that ships files under data/outcomes but not under
+  // documents/GUIDE.md, which is a blob
+  const probes = manifestProbes({
+    trackedOutput: '',
+    upstreamOutput: 'data/outcomes/posting.md\0documents/GUIDE.md\0',
+  });
+
+  // When/Then: only the path with files beneath it is a subtree claim
+  if (probes.claimsSubtree('data/outcomes') && !probes.claimsSubtree('documents/GUIDE.md')) {
+    pass('claimsSubtree() distinguishes a tree from a blob using the upstream listing');
+  } else {
+    fail('claimsSubtree() does not read the upstream listing');
+  }
+}
+
+{
+  // Given: any probes, asked about a trailing-slash entry
+  const probes = manifestProbes({ trackedOutput: '', upstreamOutput: '' });
+
+  // When/Then: an explicit slash is a subtree claim without consulting the tree,
+  // so an entry upstream ships nothing under yet is still refused
+  if (probes.claimsSubtree('documents/')) {
+    pass('claimsSubtree() honours an explicit trailing slash with an empty tree');
+  } else {
+    fail('claimsSubtree() ignores an explicit trailing slash');
+  }
+}
+
+{
+  // Given: a root the exists probe should resolve against
+  const probes = manifestProbes({ trackedOutput: '', upstreamOutput: '', root: process.cwd() });
+
+  // When/Then: it reports on the real filesystem under that root
+  if (probes.exists('update-system.mjs') && !probes.exists('no-such-file-xyz.md')) {
+    pass('exists() resolves against the root it was given');
+  } else {
+    fail('exists() does not resolve against the given root');
   }
 }
--- a/tests/updater-remote-manifest-user-paths.test.mjs
+++ b/tests/updater-remote-manifest-user-paths.test.mjs
@@ -3,11 +3,12 @@
  * guard on the FETCHED manifest.
  *
  * apply() reads SYSTEM_PATHS out of the updater it just fetched and merges it
- * into updatePaths. That list decides what the checkout overwrites AND, through
- * `updatePaths.includes(file)` in userLayerViolations(), what is excused from
- * the user-layer safety check — so an upstream entry naming a user path both
- * causes the damage and waives the check written to catch it. The guard is
- * disabled by the same entry that triggers it.
+ * into updatePaths, the list the per-path `git checkout FETCH_HEAD -- <path>`
+ * walks. An entry naming user territory therefore writes upstream's content over
+ * the user's own, and an UNTRACKED local file is invisible to the diff-based
+ * #2337 detector — so it gets no .bak, is not preserved, and is then deleted by
+ * the abort path as an addition HEAD lacks, while the run prints "your content
+ * was NOT overwritten". Refusing the entry is what stops that sequence starting.
  *
  * rejectUserLayerPaths() is pure and exported for the same reason
  * userLayerViolations() is: apply() is ROOT-bound and full of side effects, so
@@ -36,6 +37,24 @@ const USER_PATHS = [
   'writing-samples/',
 ];
 
+// Deterministic stand-ins for the three local-state questions the rule asks.
+// Every block passes these: without them rejectUserLayerPaths falls back to a
+// real `git ls-files` and existsSync against whatever checkout the suite happens
+// to run in, which makes the outcome depend on the developer's tree and throws
+// outright where there is no git. The values below deliberately DISAGREE with
+// this repo — documents/GUIDE.md and data/outcomes/posting.md do not exist here —
+// so a regression to the real probes cannot keep these assertions green.
+const TRACKED = new Set(['writing-samples/README.md', 'documents/README.md',
+  'documents/.gitkeep', 'interview-prep/sessions/.gitkeep']);
+const ON_DISK = new Set([...TRACKED, 'data/applications.md', 'interview-prep/story-bank.md']);
+const UPSTREAM = ['data/outcomes/posting.md', 'documents/README.md', 'modes/pdf/hm-audit.md'];
+const probes = {
+  tracked: (p) => TRACKED.has(p),
+  exists: (p) => ON_DISK.has(p),
+  claimsSubtree: (p) => p.endsWith('/')
+    || UPSTREAM.some((f) => f.startsWith(`${p.replace(/\/$/, '')}/`)),
+};
+
 console.log('\n🧪 Testing rejectUserLayerPaths (fetched manifest vs local user layer)...');
 
 {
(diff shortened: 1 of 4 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is `update-system.mjs`, a red-line file and the maintainer's call; I am reviewing the owned test `tests/updater-remote-manifest-user-paths.test.mjs`, and since this is a security guard (a crafted upstream manifest naming a user-layer path) I verified it strictly.

It is sound. The test imports the real exported `rejectUserLayerPaths` and calls it directly, so it drives the actual refusal logic rather than a hand-reconstructed filter. The refusal is pinned four ways that each redden if the guard is gutted to all-kept: `cv.md`, `documents/`, the nested `data/outcomes/`, and `modes/` (which contains the user file `modes/_profile.md`), each asserting the path is refused and not kept. Acceptance is pinned too, so a lazy over-broad refusal also reddens: an ordinary manifest passes system paths through in order, and system-owned docs sitting inside user directories (`writing-samples/README.md`, `documents/GUIDE.md`) stay kept. Crucially the self-bootstrap trap is closed: the guard runs on the merged manifest (`rejectUserLayerPaths(mergePathLists(SYSTEM_PATHS, remoteSystemPaths, BOOTSTRAP_PATHS), effectiveUserPaths())`), not the fetched half alone, and that wiring is pinned separately by the migration structural check.

One honest note, not a false-pass: the pure-function test uses a synthetic user-paths stand-in rather than the real `effectiveUserPaths()`, but the apply()-side wiring to the real source is covered by the migration test, so the two together hold the red-line.

Scott-Emberson (commented):
Re-verified at this head. This carries new commits since my last look (close two bypasses, stop the guard bricking the updater, extract `manifestProbes` for testability, refuse non-canonical spellings), and I re-ran the whole thing rather than trusting the rebase.

The owned test imports the real `rejectUserLayerPaths` and `manifestProbes` and drives them directly, no hand-rebuilt filter. Refusal is pinned many independent ways: an exact user file (`cv.md`), a user directory (`documents/`), a nested user dir (`data/outcomes/`), an ancestor of a user file (`modes/` claiming `modes/_profile.md`), slashless spellings, and non-canonical spellings (`./data/`, `..\data`, `:(glob)data/**`, `data/../data/`). Acceptance is pinned too: ordinary system paths pass in order, and system-owned docs sitting inside user directories stay kept. Crucially the guard runs on the merged manifest (`rejectUserLayerPaths(mergePathLists(SYSTEM_PATHS, remoteSystemPaths, BOOTSTRAP_PATHS), effectiveUserPaths(), manifestProbes(...))`), not the fetched half alone, and the `manifestProbes` wiring to real git output is separately pinned. 20 assertions pass. Gutting the guard to all-kept reddens 8 refusal assertions while acceptance stays green, so it is not tautological.

Sound on the test axis. The branch is behind main, and `update-system.mjs` is the maintainer's call on the security substance.

Scott-Emberson (commented):
Re-verified at this head. New commits since my last look harden the guard further: refuse a manifest entry containing a NUL byte, refuse glob metacharacters (`* ? [`), and refuse any colon (covering pathspec magic like `:(glob)`, a Windows drive prefix such as `C:/x`, and an NTFS alternate-data-stream `file.txt:stream`). I re-ran rather than carrying the prior verdict.

The owned test imports the real `rejectUserLayerPaths` and `manifestProbes` and drives them on the merged manifest (`rejectUserLayerPaths(mergePathLists(SYSTEM_PATHS, remoteSystemPaths, BOOTSTRAP_PATHS), effectiveUserPaths(), manifestProbes(...))`), not the fetched half alone, with the probes fed real git output. Refusal is pinned many ways: exact user file, user directory, nested user dir, an ancestor of a user file, slashless spellings, and now the NUL, wildcard, colon/drive, and ADS cases; acceptance is pinned too (system-owned files inside user dirs kept, ordinary passthrough in order). 20 assertions pass, and neutering the guard to all-kept reddens 8 refusal assertions including the non-canonical-spellings one that pins every new rule, while acceptance stays green.

Owned test sound; branch behind main. SIGNAL (update-system.mjs red-line, security), so not surfacing as ready-to-approve; routing and merge are the maintainer's.

PR #1910, opened 2026-07-15
fix: unify atomic report number allocation

luochen211:
## Summary

- make `reserve-report-num.mjs` an import-safe shared allocator used by every headless evaluator
- reserve against report files, sentinels, tracker row IDs, and tracker report links under the shared tracker lock
- preserve free reserved IDs during tracker merge and warn only on real collisions
- block numeric `set-status` updates when the tracker row and report link disagree unless `--force` is explicit
- keep report numbering correct beyond 999

This carries forward the two-layer safeguard proposed in #1734 and consolidates the related allocator fixes in one non-conflicting change.

Closes #1733
Closes #1801
Closes #1835
Closes #1895
Supersedes #1734

## Root cause

Report files, tracker rows, and the four headless evaluators maintained independent counters. Out-of-order completion or an incomplete local `reports/` directory could therefore reuse or silently rewrite a reserved ID.

## Verification

- `node set-status-tests.mjs` (38 passed)
- `node test-all.mjs` (1742 passed, 0 failed; 2 expected environment warnings)
- dashboard compile included in the full suite

## Summary by CodeRabbit

- **New Features**
  - Report numbering now uses a shared, atomic reservation workflow across evaluators for safer parallel runs.
  - Report numbers are formatted consistently, including four-digit values.
  - Added `--force` to the status CLI, enabling override behavior when tracker/report numbers don’t align.

- **Bug Fixes**
  - Detects tracker/report-number mismatches and returns a structured error unless `--force` is used.
  - Improved collision handling and cleanup of stale reservations.

- **Documentation**
  - Updated CLI usage and headless/batch guidance for reservation behavior.

- **Tests**
  - Added regression coverage for mismatches and reservation fidelity.

Issue #1895, opened 2026-07-15
bug(evals): all four headless engines compute report numbers with max+1 — the #749 race reserve-report-num.mjs exists to prevent


abankar1:
## Problem

AGENTS.md is explicit about report numbering:

> **Parallel fan-outs — reserve report numbers first.** … Never let parallel workers compute `max+1` themselves — that is the #749 race.

`reserve-report-num.mjs` exists precisely for this (atomic `NNN-RESERVED.md` sentinel claims). But all **four headless eval engines** ship their own private `max+1`:

| Engine | Location | Implementation |
|---|---|---|
| `ollama-eval.mjs` | 146–152 | `readdirSync(reports/)` → `Math.max(...nums) + 1` |
| `openai-eval.mjs` | 201–207 | same, duplicated |
| `gemini-eval.mjs` | 156–163 | same, duplicated |
| `openrouter-runner.mjs` | ~532 | same, duplicated |

None of them create a sentinel or look at `reserve-report-num`'s reservations.

## Why this is not theoretical

- `openrouter-runner.mjs` is the **batch** runner for the no-CLI path — its whole point is processing many offers, and the free-model rotation makes runs long enough that users start a second one, or run it alongside an agent-driven `/career-ops pipeline` (whose parallel workers *do* use `reserve-report-num.mjs`).
- Two writers that pick the same `NNN` produce either a clobbered report file or two tracker rows pointing at each other's reports — the exact corruption #749 was about. Worse, the engines' `max+1` **ignores `NNN-RESERVED.md` sentinels**… actually it doesn't: `readdirSync` sees the sentinel files and counts them, which is only accidental protection in one direction — an *agent* reservation bumps the *engine's* number, but an engine's in-flight number is invisible to everyone else until its report is written (the un-reserved window is the whole model-inference time, minutes on ollama).
- Four duplicated copies of the same helper is also drift waiting to happen (cf. the tokenizer drift that produced #1851).

## Fix

1. Export the claim logic from `reserve-report-num.mjs` as a function (`export function reserveReportNums(count = 1)` wrapping the existing sentinel mechanics; CLI keeps calling it).
2. Replace the four `nextReportNumber()` copies with a claim → write report → release (or let the existing verify-pipeline GC handle stale sentinels, as documented).
[...]

Issue #1801, opened 2026-07-11
bug(reserve-report-num): reservation never cross-checks applications.md — can hand out an already-claimed number (follow-up to #1733/#1734)


Schlaflied:
## Follow-up to #1733/#1734 — same root cause, opposite end of the pipe

#1733 nailed the root cause in one line: *"report numbers come from `reserve-report-num.mjs` (counts off `reports/`) while `maxNum` comes from `applications.md`'s row count — two independent counters."* #1734 fixes the **merge-time** half (a reserved number no longer gets silently discarded when `<= maxNum`). This issue is the **reservation-time** half that #1734 doesn't touch: `reserve-report-num.mjs` itself never cross-checks `data/applications.md`, so it can hand out a number that's already claimed in the tracker.

## Problem

`reserve-report-num.mjs` computes the next number purely from `readdirSync(REPORTS_DIR)` — scanning the **local, gitignored** `reports/` directory. It has no visibility into `data/applications.md` (also gitignored, and the actual permanent record of every report number ever claimed via a tracker row's `Report` link).

When the local `reports/` folder is missing files that the tracker already references — which will happen on any environment where report `.md` files aren't 1:1 preserved locally (partial sync across machines, manual archiving/cleanup, a fresh-ish local checkout, etc.) — the script confidently hands out a number that **collides with an existing tracker row**, because from its point of view that number was never used.

**Concrete, reproduced case (not hypothetical):** on my machine right now, `reports/` contains only 4 real files (001–004) plus one long-stale `005-RESERVED.md` sentinel (well past the 4h TTL, never GC'd). `data/applications.md` references **270 distinct report numbers** up to **1433**. Numbers 5–62 happen to be an unused historical gap, but **#63 is already taken** — tracker row #63 (`Company A — Instructional Designer`) links `reports/063-company-a-instructional-designer-2026-04-17.md`. Run `reserve-report-num.mjs` enough times in sequence from this state and the 58th call hands out `063` — a number already live in the tracker.

## Why #1734 alone doesn't close this

[...]

PR #3948, opened 2026-09-06
fix(eval): reserve report IDs from the configured data root

moelomda:
## What does this PR do?

Use `DATA_ROOT`, rather than the installation's `ROOT`, when OpenAI and Ollama reserve report numbers. Reports already go to the configured data directory, but the old argument makes the shared allocator read and lock the installation's tracker instead. Tracker-only IDs in an external workspace can therefore be reused when the corresponding report files are absent.

The production change is two lines. The allocator, reservation ownership tokens, and merge logic are unchanged; Gemini already passes `DATA_ROOT` for this operation.

## Related issue

A surviving caller-path defect related to the shared allocator work in #1910 (which addressed #1801 and #1895), not a replacement for that work. No separate issue is required for a bug fix under the current contribution guidelines.

Rechecked upstream `fe06051ba57c78934b89355bebe2c67f3a5cdb14`, relevant historical work, and active PRs. No equivalent fix was found. This intentionally does not duplicate the tracker-addition helper consolidation in #3906.

## Reproduction and tests

The new `tests/evaluator-report-reservations.test.mjs` drives both real CLIs against local HTTP mocks with isolated code/data directories and synthetic inputs. In the regression fixture, the installation tracker reserves 7 while the external tracker has row 42 and report link 73, with no old report files on disk. Before the fix the evaluators allocate 008; the correct result is 074.

Coverage includes both root environment variables, relative roots, a relative marker file, legacy tracker placement, explicit tracker precedence, default-root behavior, first-run state, and existing report files. Each case checks the report/addition IDs and link, unchanged trackers, reservation cleanup, and output placement.

- Before the fix: `node test-all.mjs --only evaluator-report-reservations` -> **6 passed, 12 failed**.
- After the fix: the same command -> **18 passed, 0 failed**.
- `node test-all.mjs --only reserve-report-num` -> **6 passed, 0 failed**.
- `node validate-system-paths-coverage.mjs` -> **1390 tracked files covered**, including the new test.
- Syntax checks for all changed JavaScript files and `git diff --check` passed.

## Verification limitation / why draft

**The full suite is not verified on this patch.** A full `node test-all.mjs` run on unchanged upstream in this Windows environment recorded 19 failing assertions/suites, including subprocess timeouts/SIGTERM exits, batch checks, web unit suites, and CLI checks. That run was stopped before its final summary. Those observations are not being attributed to this patch or presented as confirmed new upstream bugs.

WSL could not start (`HCS_E_HYPERV_NOT_INSTALLED`), and Go was unavailable on PATH. Linux/macOS, the dashboard build, and live model calls were not verified locally. Targeted test passes are not a substitute for the full suite. This is a draft pending full-suite/CI verification and maintainer review.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] Issue-first requirement considered: this is an exempt bug fix, not a new feature or architecture change.
- [x] No personal data is included; all new fixtures are synthetic and temporary.
- [ ] I ran `node test-all.mjs` on this patch and all tests pass.
- [x] The change respects the Data Contract and does not edit user-layer files.
- [x] The change strengthens the existing local evaluation/tracking path.

Prepared with Codex assistance; the verification scope and limitations are listed above.

### DIFF
--- a/ollama-eval.mjs
+++ b/ollama-eval.mjs
@@ -515,7 +515,7 @@ if (saveReport) {
       mkdirSync(PATHS.reports, { recursive: true });
     }
 
-    reservedNumbers   = await reserveReportNumbers(1, { rootDir: ROOT, reportsDir: PATHS.reports });
+    reservedNumbers   = await reserveReportNumbers(1, { rootDir: DATA_ROOT, reportsDir: PATHS.reports });
     const num         = formatReportNumber(reservedNumbers[0]);
     const today       = new Date().toISOString().split('T')[0];
     const companySlug = slugifyCompany(company);
--- a/openai-eval.mjs
+++ b/openai-eval.mjs
@@ -523,7 +523,7 @@ if (saveReport) {
       mkdirSync(PATHS.reports, { recursive: true });
     }
 
-    reservedNumbers   = await reserveReportNumbers(1, { rootDir: ROOT, reportsDir: PATHS.reports });
+    reservedNumbers   = await reserveReportNumbers(1, { rootDir: DATA_ROOT, reportsDir: PATHS.reports });
     const num         = formatReportNumber(reservedNumbers[0]);
     const today       = new Date().toISOString().split('T')[0];
     const companySlug = slugifyCompany(company);
--- a/tests/evaluator-report-reservations.test.mjs
+++ b/tests/evaluator-report-reservations.test.mjs
@@ -1,0 +1,173 @@
+// Exercise the real save paths without models, credentials, or personal data.
+// A relocated tracker still owns its IDs when its old report files are absent.
+import assert from 'node:assert/strict';
+import { execFile } from 'node:child_process';
+import { copyFileSync, existsSync, mkdirSync, mkdtempSync, readFileSync, readdirSync, writeFileSync } from 'node:fs';
+import { createServer } from 'node:http';
+import { tmpdir } from 'node:os';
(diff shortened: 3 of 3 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is report-number reservation root resolution in `reserve-report-num.mjs`, which is eval-core routing and the maintainer's call; I am reviewing the owned test.

It is sound and specifically not a #3370-style root-not-pinned false-pass. The test seeds a code-root tracker with report 7 and an external `dataRoot` with report 73, then asserts the configured-root cases reserve 74 (73+1), so it reads from the data root rather than the code root. I mutation-checked the actual fix: reverting `rootDir` from `DATA_ROOT` back to `ROOT` makes the reservation read the code-root tracker and reserve 8, which reddens the `074-` and num-74 assertions. Both env axes are pinned separately, `CAREER_OPS_ROOT` and `CAREER_OPS_DATA_DIR`, each with a literal expected value rather than a value derived from the function under test, so there is no tautology. The first-run empty case expects 1, and the single-process path is covered against a double-reserve.

Because the assertions are against literal expecteds and the reserve path is driven through the real entry point with the environment explicitly pinned, the import-time-mkdir-in-the-wrong-root failure mode that defeated the earlier guard does not apply here.

No test-axis concern. The substance is eval-core routing, so routing and merge are the maintainer's call.

PR #3957, opened 2026-09-07
fix(scan): strip CJK/Korean corporate-form markers in companyMatch (#2570)

### LINKED ISSUE #2570: companyMatch() partial-name matching never fires for non-Latin company names

barkaaa:
Closes #2570.

`companyMatch('株式会社メルカリ', 'メルカリ')` returned `false`: the character before メルカリ in 株式会社メルカリ is 社, a letter, so no anchor rule can treat it as a boundary — and CJK/Korean don't delimit words with spaces the way the existing containment fallback (or `LEGAL_SUFFIXES` in `invite-match.mjs`) assumes.

Per the direction we aligned on in the issue thread: a `CORPORATE_FORMS` list local to `scan.mjs` (株式会社/合同会社/有限会社, 股份有限公司/有限公司, 주식회사), stripped as an unspaced prefix or suffix (at most one strip, longer forms checked first) before the equality/containment checks run — rather than loosening the anchors, which would let アカネ match アカネスタジオ.

`invite-match.mjs`'s `normalizeCompanyName`/`LEGAL_SUFFIXES` are untouched: that normalizer is deliberately stricter than `companyMatch` (matching loosely-quoted email text vs. defending an identity key that must never merge two different companies), so folding a CJK list in there risked loosening its callers too.

**Tests added** (`test-all.mjs`, section 45d):
- the confirmed repro pair (`株式会社メルカリ` / `メルカリ`)
- the suffix form (`メルカリ株式会社` / `メルカリ`)
- Chinese and Korean forms, including the longer-form-first case (股份有限公司 before 有限公司)
- unrelated companies stay unrelated after stripping (discrimination)
- a bare-marker-only name doesn't collapse into another bare-marker-only name
- a mid-name marker (not at an edge) is left alone
- Latin `companyMatch` behavior regression check

Full `test-all.mjs` suite: 7514 passed. The 4 failures in this checkout are pre-existing on `main` (unrelated: a missing `web/` dir, an untracked local batch artifact, and unregistered `tests/*.test.mjs` files in `SYSTEM_PATHS`) — confirmed by running the same suite on `main` before this branch's commit.

## Summary

`companyMatch` now matches Japanese, Chinese, and Korean company names when one supported corporate-form marker appears at the start or end of either name.

Users can match names such as `株式会社メルカリ` and `メルカリ` without adding spaces. The logic removes at most one marker and checks longer markers before shorter overlapping markers. It preserves safeguards for unrelated companies, mid-name markers, marker-only names, and existing Latin matching.

Tests were moved to `tests/company-match-corporate-forms.test.mjs` and cover supported forms, discrimination cases, and Latin regressions.

## Files changed

`scan.mjs`: Updated `companyMatch` normalization and comparison logic.

`tests/company-match-corporate-forms.test.mjs`: Added focused coverage for corporate-form matching.

## Validation

The supplied review reports 9,178 tests passed with no failures.


santifer:
Sorry, @barkaaa. You were told on #2570 to send it when ready, you sent it two days later, and then eleven days of silence from us. That is on us, not on you.

The change is right, and we ran it rather than read it. On top of current main the full suite is 9178 passed and 0 failed, and a targeted probe confirms both directions: the Japanese, Chinese and Korean pairs go from not matching against main to matching with your change, and the pairs that must stay apart stay apart, including two different Japanese companies sharing a prefix and two different brands sharing the same corporate form. That is the part worth checking, since this key has to split rather than merge. You built what was agreed: a separate list next to companyMatch, one strip at either edge, longest form first, and the stricter normalizer left alone.

One change and it goes in. Move section 45d out of test-all.mjs into a file of its own under tests/, for example tests/company-match-corporate-forms.test.mjs, importing companyMatch from ../scan.mjs and pass/fail from ./helpers.mjs. The assertions themselves do not change, and scan.mjs is fine as written.

Why that is worth another round trip rather than a follow-up issue: test-all.mjs is open in more than forty pull requests right now, and a hand-picked section number is the one edit that makes them collide. That is what the note at the top of the file is about. A file of its own is discovered automatically and lands in parallel with everything else.

Ping here when it is moved and we take it from there.

barkaaa:
Moved — and no apology needed on the timing.

Section 45d is out of `test-all.mjs`; the assertions now live in `tests/company-match-corporate-forms.test.mjs`, importing `companyMatch` from `../scan.mjs` and `pass`/`fail` from `./helpers.mjs`. `test-all.mjs` is byte-identical to this branch's base again, so the PR no longer touches it at all. I also rebased onto current `main` so the numbers line up with yours.

**But moving them turned up a bug in my own change, and I'd rather you hear it from me than find it.**

Stripping the form from each side independently makes the two sides equal whenever they differ *only* by form. Against `main`, old → new:

| pair | main | this branch (first commit) |
|---|---|---|
| `株式会社アカネ` vs `合同会社アカネ` | false | **true** |
| `株式会社アカネ` vs `有限会社アカネ` | false | **true** |
| `阿里有限公司` vs `阿里株式会社` | false | **true** |
| `小米股份有限公司` vs `小米有限公司` | false | **true** |

A KK and a GK sharing a trade name are two different legal entities, as are a PRC and a JP entity. That's a false merge in an identity key — the exact direction "splits, never merges" exists to forbid. It's also precisely the case your probe didn't cover: you checked two different brands sharing the same corporate form, which stays correct; this is the same brand with different forms.

Blast radius is bounded — `buildCooldownFilter` is the only production caller — but the symptom is silent: a job from a *different* company skipped as in-cooldown.

Note the asymmetry it created, too: Latin `Acme Inc` vs `Acme Ltd` was still false, while the CJK equivalent had become true.

**The fix** (third commit): `stripCorporateForm` returns `[form, remainder]` instead of a bare string, and `stripFormPair` declines to strip when both sides carry a form and the forms differ — two different explicit forms are positive evidence of two different entities, the same way a mismatched req number is (#1524). All four rows above are back to `false`; `株式会社メルカリ` vs `メルカリ`, the suffix form, and the zh/ko pairs all still match. The trade is that a company which converted 有限会社 → 株式会社 stops matching itself across the rename — the split direction, which I took to be the one you'd want.

Returning the form also retires the overload where `''` meant both "no form found" and "the name was only the form".

**Two test changes**, both mutation-verified as the only assertion that fails against the mutant it targets:

- same-name/different-form must stay apart. The discrimination case I originally shipped (`株式会社アカネ` vs `合同会社ゾロ`) varies *both* name and form, so it passes with or without the strip — it had no teeth.
- `companyMatch('株式会社','株式会社') === true`, the only possible witness for the stripped-to-empty fallback. Worth noting this one only *became* testable after the `[form, remainder]` refactor; while `''` carried both meanings, no mutant could isolate it.

Full suite on this branch: **9212 passed, 0 failed.**

I realise this is more than the one change you asked for, and you said `scan.mjs` was fine as written. If you'd rather keep this PR to exactly the test move and take the merge fix as its own PR against a fresh issue, say the word and I'll split it in a few minutes — the commits are already separate.

Two things I deliberately left out, happy to file as follow-ups: `CORPORATE_FORMS` still misses some common unspaced forms (`有限责任公司` / `有限責任公司` — note it doesn't end in `有限公司`, so the list doesn't reach it — plus `合名会社`, `合資会社`, `一般社団法人`, `유한회사`), and I didn't touch `invite-match.mjs`, as agreed on #2570.

santifer:
@barkaaa you moved what we asked, and then found a bug in your own change and told us before we could find it. That is the part worth saying out loud.

We ran it rather than take your word for it. Against main, the five Japanese, Chinese and Korean pairs go from not matching to matching; the four you flagged, same trade name with different legal forms, stay apart on the final commit and do merge on your second one. Your same-name/different-form assertion is the only one that fails against that mutant. Full suite here: 9213 passed, 0 failed.

No need to split it. The merge fix belongs with the change that introduced it, and the trade you took is the right way round: a company that converts from 有限会社 to 株式会社 no longer matching itself is the safe side, because this key has to split rather than merge.

Please do file the missing forms (有限责任公司, 合名会社, 合資会社, 一般社団法人, 유한회사) as their own issue. invite-match.mjs stays as agreed.

Approved. Next step is ours: refresh the branch on top of main, then merge.

santifer (approved):

PR #3958, opened 2026-09-07
fix(scan): recheck_after_days frees a row, then pipeline.md pins it again

Duyanh090205:
## The bug

`collectSeenUrls` builds the seen-set from three sources, but only the first consults the age policy. scan-history.tsv releases a row once it is past `scan_history.recheck_after_days`; pipeline.md and applications.md are then parsed with no policy at all — and every URL in scan-history that has ever been queued is also in pipeline.md. The row is re-pinned a few lines later, so the window frees nothing on any install past its first run.

The counter hid it. `recheckEligible` was incremented at the release site, before the two later sources could pin the row again, so it reported rows as eligible that the same call had already put back in the set.

Measured on one install: **252 of 252** scan-history URLs were also in pipeline.md, so a 30-day window "released" 244 rows, re-pinned every one of them, and printed `244 eligible`.

## Why the release has to be conditional

A `- [ ]` row under `## Pending` is a line the user can still pull from. Re-scanning it appends a **second copy** of a job already on the list, so an unconditional release would trade a silent no-op for a duplicate-entry bug.

Only finished work is safe to hand back to the TTL:

- `- [x]`, or anything under `## Processed` → released, but **only** if scan-history is what released it
- a pipeline-only URL → always pinned; nothing authorised un-pinning it
- applications.md → still pins unconditionally. An applied job must not be re-offered, and there is no checkbox there to tell finished from open.

The count moves to the end, against the finished set, so it reports what is actually rescannable rather than what was momentarily released.

## Tests

New suite `tests/scan-recheck-pipeline-repin.test.mjs`, 7 assertions. It **fails 4 of 7 on the unpatched collector** and passes all 7 with the fix; the three that pass either way are the over-release guards, which is deliberate — they protect the fix, not the bug.

`tests/scan-dedup-snapshot.test.mjs` is unaffected. Its recheck candidates (`jobs.lever.co/beta/2`, `jobs.example.com/eps/5`) are not in its pipeline fixture, so both the golden seen-set and the golden `recheckEligible: 2` come out byte-identical — the behaviour proof still holds.

Full suite on this branch: **756 tests, 752 pass, 0 fail, 4 skipped.** `check-syntax.mjs`: 694 files.

## Scope

One function. No signature change, no change to the `{ seen, recheckEligible }` shape, nothing else touched.

## Summary

Users can re-scan completed pipeline URLs after `scan_history.recheck_after_days` expires: `scan.mjs:1539-1621`.

Pending URLs, pipeline-only URLs, and application URLs remain pinned. Processed status continues under deeper headings and ends at same-level or shallower headings: `scan.mjs:1577-1606`.

Added regression coverage for release, re-pinning, heading handling, and final `recheckEligible` counting: `tests/scan-recheck-pipeline-repin.test.mjs:61-146`.

No function signatures or return shapes changed.

System files not touched: `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/`.

### DIFF
--- a/scan.mjs
+++ b/scan.mjs
@@ -1574,19 +1574,20 @@ export function collectSeenUrls(sources = {}, policy = {}, { extraTokensFor } = 
   // under `## Processed`, is finished work: no queue entry can be duplicated, so
   // the scan-history TTL is allowed to govern it alone. Release also requires the
   // URL to be a recheck candidate, so a pipeline-only URL is never un-pinned.
+  // Sections are `##` in PIPELINE_SKELETON — `# Pipeline` is the document title,
+  // `## Pending` and `## Processed` are the sections. So only a level-2 heading
+  // opens or closes one; anything deeper is a subdivision INSIDE the current
+  // section and must leave the section alone. Matching `Processed` at any depth
+  // breaks both ways: `### Processed leftovers` under `## Pending` would release
+  // rows that are still queued for work — the exact duplication this function
+  // exists to prevent — while `### August` under `## Processed` has to keep the
+  // section open.
+  const SECTION_LEVEL = 2;
   let inProcessed = false;
-  let processedHeadingLevel = 0;
   for (const line of pipelineText.split('\n')) {
     const heading = line.match(/^(#+)\s+(.*)$/);
-    if (heading) {
-      const level = heading[1].length;
-      if (/^processed\b/i.test(heading[2].trim())) {
-        inProcessed = true;
-        processedHeadingLevel = level;
-      } else if (inProcessed && level <= processedHeadingLevel) {
-        inProcessed = false;
-        processedHeadingLevel = 0;
-      }
+    if (heading && heading[1].length === SECTION_LEVEL) {
+      inProcessed = /^processed\b/i.test(heading[2].trim());
     }
     const url = extractPipelineUrl(line);
     if (!url) continue;
--- a/scan.mjs
+++ b/scan.mjs
@@ -1575,9 +1575,19 @@ export function collectSeenUrls(sources = {}, policy = {}, { extraTokensFor } = 
   // the scan-history TTL is allowed to govern it alone. Release also requires the
   // URL to be a recheck candidate, so a pipeline-only URL is never un-pinned.
   let inProcessed = false;
+  let processedHeadingLevel = 0;
   for (const line of pipelineText.split('\n')) {
-    const heading = line.match(/^#+\s+(.*)$/);
-    if (heading) inProcessed = /^processed\b/i.test(heading[1].trim());
+    const heading = line.match(/^(#+)\s+(.*)$/);
+    if (heading) {
+      const level = heading[1].length;
+      if (/^processed\b/i.test(heading[2].trim())) {
+        inProcessed = true;
+        processedHeadingLevel = level;
+      } else if (inProcessed && level <= processedHeadingLevel) {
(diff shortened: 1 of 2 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is in `scan.mjs`, a red-line file and the maintainer's call; I am reviewing the owned test `tests/scan-recheck-pipeline-repin.test.mjs`.

It is sound and mutation-proven. The test imports the real `collectSeenUrls` from scan.mjs and calls it with fixture sources and policy, feeding raw scan-history/pipeline/applications text and letting scan.mjs's own three-source merge, its TTL dedup, the `## Processed`/`- [x]` heading parse, and the end-of-function counter all run. It does not re-implement the date-freeing filter or hand-supply the released set. I mutation-checked the fix that matters by reverting the pipeline.md loop to unconditionally re-add the key (dropping the `if ((done || inProcessed) && recheckCandidates.has(key)) continue` release guard), and four assertions reddened (the `- [x]` and `## Processed` rows stayed pinned, `recheckEligible` came back 0 instead of 2, and the applications pin-leak count). Restoring returned it to green, so it is not a false-pass. The still-passing assertions (an actionable `- [ ]` row and a pipeline-only URL stay pinned) are correctly invariant under that mutation.

Owned test sound, not behind. SIGNAL (scan.mjs red-line), so not surfacing as ready-to-approve; routing and merge are the maintainer's.

Scott-Emberson (commented):
Test-axis only. The substance is in `scan.mjs` (the recheck pipeline repin, a red-line file), the maintainer's call; I am reviewing the owned test `tests/scan-recheck-pipeline-repin.test.mjs`.

It is sound. The test imports the real `collectSeenUrls` and drives the genuine three-source merge (scan-history, pipeline, applications), exercising the age-policy release, the `- [x]` and `## Processed` (including a nested child heading) release path, and the actionable `- [ ]` re-pin. Eight assertions pass. I mutation-checked it: reverting the repin guard so a freed row is never re-pinned reddens three assertions, the still-pinned actionable row, the recheckEligible counter (4 versus the expected 3), and the applications.md pin-leak count.

Owned test sound; branch behind main. SIGNAL (scan.mjs red-line), so not surfacing as ready-to-approve; routing and merge are the maintainer's.

santifer:
Thanks for your patience, @Duyanh090205, and for how you handled the three findings. Reproducing each one before changing anything, and pushing @chipoto69's fix as their commit instead of squashing it, is the right instinct on both counts.

What this fixes is bigger than the title suggests. `scan_history.recheck_after_days` was a no-op on every install past its first run: only the scan-history pass consulted the age policy, so a URL the window released was pinned again a few lines later by the pipeline.md pass. The counter was incremented at the release site too, so it reported rows as eligible that the same call had already put back. Both halves are gone.

I checked instead of reasoning: on a checkout of your branch over current `main`, your suite is 10 passed, 0 failed and the full suite 9072 passed, 0 failed. Swapping in main's `scan.mjs` turns five of your ten rows red, the counter one included.

One for later, not a blocker: `scan.mjs` still honours a legacy Spanish `## Procesadas` section elsewhere, and the new heading check only matches `Processed`. It fails closed, so those rows stay pinned exactly as they do today. Worth a follow-up if you want it.

The branch is behind `main`, so refreshing it and the merge are on me.

santifer (approved):

PR #3844, opened 2026-09-04
Closes #3670: i18n(da): re-sync modes/da/oferta.md (Danish) to the canonical A–H + Risk Summary structure

evgenyponomarev:
Closes #3670.

This change touches only documentation, and the sandbox that produced it recognises no build system in this repository, so nothing was compiled or tested.
The patch applies cleanly to the pinned commit (`71044b4c7b1f`).

Written by an AI coding agent (Sloppy) and opened under my account.

## Summary

- Re-syncs `modes/da/oferta.md` with the canonical A–H evaluation structure.
- Adds URL liveness, blacklist, bounded research, and untrusted-input gates.
- Updates Blocks B–D with structured requirements, evidence tiers, gap mitigation, compensation rules, and verification questions.
- Adds Block G for posting legitimacy and risk signals.
- Adds `Machine Summary`, `Risk Summary`, cover-letter, agency-handling, tracker, and salary-observation sections.
- Updates report metadata and report ordering to end with `G)` → `Risk Summary` → `H)` → keywords.

## User impact

Users receive a complete Danish offer evaluation workflow with legitimacy and risk analysis. Reports include machine-readable metadata and additional cover-letter and compensation guidance.

## Files changed

- `modes/da/oferta.md`: documentation-only changes.
- `AGENTS.md`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/`: not changed.

PR #3843, opened 2026-09-04
Closes #3675: i18n(ua): re-sync modes/ua/oferta.md (Ukrainian) to the canonical A–H + Risk Summary structure

evgenyponomarev:
Closes #3675.

This change touches only documentation, and the sandbox that produced it recognises no build system in this repository, so nothing was compiled or tested.
The patch applies cleanly to the pinned commit (`71044b4c7b1f`).

Written by an AI coding agent (Sloppy) and opened under my account.

## Summary

- Re-synced `modes/ua/oferta.md` with the canonical A–H workflow.
- Added the liveness, blacklist, and research-budget gates.
- Added posting-legitimacy analysis, risk summaries, cover-letter drafting, and machine-readable output.
- Updated report labels, tracker behavior, compensation rules, and ATS keyword extraction.
- Renamed the report draft-answers section to `## H)` and placed `## Risk Summary` between `## G)` and `## H)`.

## User impact

Users can evaluate Ukrainian job postings with the complete canonical workflow. The document now supports legitimacy and risk checks, structured application output, and machine-readable summaries.

Only `modes/ua/oferta.md` changed. No changes were made to `AGENTS.md`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, or `.github/`.

No runtime behavior changes. This is a documentation-only update.

PR #3814, opened 2026-09-04
Closes #3671: i18n(es): re-sync modes/es/oferta.md (Spanish) to the canonical A–H + Risk Summary structure

evgenyponomarev:
Closes #3671.

This change touches only documentation, and the sandbox that produced it recognises no build system in this repository, so nothing was compiled or tested.
The patch applies cleanly to the pinned commit (`71044b4c7b1f`).

Written by an AI coding agent (Sloppy) and opened under my account.

## Summary

Updated `modes/es/oferta.md` to match the canonical Spanish A–H evaluation format.

### User impact

- Spanish evaluations now apply URL, blacklist, and research-budget gates: `modes/es/oferta.md:7-39`.
- Blocks A–G now include current matching, compensation, legitimacy, and risk rules: `modes/es/oferta.md:48-304`.
- Reports now include `G` → `Risk Summary` → `H`, plus `Machine Summary`: `modes/es/oferta.md:370-402`.
- No runtime behavior breaks. The change affects the Spanish evaluation instructions only.
- `Verdict (lead)` remains excluded.

### Files changed

- `modes/es/oferta.md`: localization and documentation update.
- `AGENTS.md`: not changed.
- `update-system.mjs`: not changed.
- `DATA_CONTRACT.md`: not changed.
- `providers/`: not changed.
- `.github/`: not changed.

No public entities or build files changed. Test execution is not confirmed by the available evidence.

Issue #3672, opened 2026-09-02
i18n(pl): re-sync modes/pl/oferta.md (Polish) to the canonical A–H + Risk Summary structure


santifer:
Part of #3669 (umbrella: re-sync the six frozen localized `oferta.md` modes).

**Time:** ~3-4 h · **Difficulty:** Intermediate (translation + structure; no code) · **Language:** Polish

## What is wrong today

`modes/pl/oferta.md` is 166 lines against 748 in the canonical `modes/oferta.md` (main at `2e33ea86`). It predates Block G entirely. Its current headings:

```
    5:## Krok 0 -- Wykrycie archetypu
    12:## Blok A -- Podsumowanie roli
    23:## Blok B -- Dopasowanie do CV
    41:## Blok C -- Poziom i strategia
    47:## Blok D -- Wynagrodzenie i popyt
    64:## Blok E -- Plan personalizacji
    73:## Blok F -- Plan rozmów kwalifikacyjnych
    98:## Po ocenie
    123:## A) Podsumowanie roli
    126:## B) Dopasowanie do CV
    129:## C) Poziom i strategia
    132:## D) Wynagrodzenie i popyt
    135:## E) Plan personalizacji
    138:## F) Plan rozmów kwalifikacyjnych
    141:## G) Szkice odpowiedzi do aplikacji
    146:## Wyekstrahowane słowa kluczowe
```

Concretely, compared with the canonical file it is **missing**:

- `## Liveness gate (URL inputs)`, `## Blacklist gate (#1742)` and `## Bounded Research Budget` (the three gates before Step 0)
- `## Block G — Posting Legitimacy` (the whole block: signals 1–15, the agency-licensing, AI-screening-disclosure and pay-transparency sections)
- `## Risk Summary (after Block G)` and the `## Risk Summary` block-format section
- `## Cover Letter Draft` (both the trigger section and the format section)
- `## Machine Summary` (the YAML fence contract and the `advertised_comp` / `risk_summary` / `requirement_importance` rules)
- in the report format: `## G)` here is the **draft-answers** block, which is `## H)` in the canonical; the canonical `## G) Posting Legitimacy` and the `## Risk Summary` line between `G)` and `H)` are absent

Blocks A–F also drifted in content since the translation was written; diff each against the canonical rather than assuming they match.

## Acceptance

- [ ] every section of the canonical file is present, in the same order (list in #3669, rule 1)
[...]

PR #3960, opened 2026-09-07
feat(oferta): add the Verdict (lead) block the report contract has been missing

rubicon:
Part of #2318. **Draft on purpose: this must not merge before the #3669 re-sync batch lands.** Details in "Sequencing" below.

## What this is

The design we agreed on 2026-09-02: a derived `## Verdict (lead)` block, local noun with the invariant marker, positioned after Block G and before `## Risk Summary`, restating `final_decision` and never originating a different call. The mode text states that dependency out loud, so the next contributor cannot edit one without the other.

## Sequencing, and why this is a draft

Two reasons it cannot merge yet.

**The five frozen files are not in it.** "After Block G, before Risk Summary" has no meaning in `da`, `es`, `pl`, `pt` or `ua` until their re-syncs land, so they are deliberately untouched.

**And merging the test change early would break four open PRs.** If I added `## Verdict (lead)` to `REQUIRED_HEADINGS` in the parity check now, the re-sync PRs could never clear their files off `FROZEN_OFERTA`, because you excluded the lead block from those issues on purpose. So this branch does not touch `REQUIRED_HEADINGS` at all. That change belongs in the final commit here, once #3844, #3814, #3877, #3843 and #3672 are in.

I will add the five localized blocks and the `REQUIRED_HEADINGS` update as the batch clears, then mark this ready.

## One thing that changes the scope we agreed

The nine-file plan globbed `oferta.md`, and three other places carry the same report skeleton:

- `modes/ja/kyujin.md` and `modes/ar/fursah.md` are the Japanese and Arabic evaluation modes. Both already carry the canonical A to H plus Risk Summary, and both are invisible to the parity check in `test-all.mjs`, which walks `modes/*/oferta.md` only.
- `batch/batch-prompt.md` carries its own report block list at the "Then include:" step. Batch-generated reports would have had no lead block at all.

All three are included here. Worth deciding separately whether the parity check should walk the evaluation mode per locale rather than the filename `oferta.md`, since today `ja` and `ar` can drift without anything failing. I have not touched that; it is a wider change than this issue.

## A test fix that is not cosmetic

`test-all.mjs` bounded the Risk Summary slice with a whole-file `indexOf('Block format:')`. Any section added before Risk Summary that carries a block-format example pushes that end index before the start index, slices the section to empty, and fails the AI-screening disclosure row check (#2892) for a reason unrelated to the row. This branch hit exactly that. The end bound is now searched from the section start.

I checked the repaired assertion still guards rather than passing vacuously: deleting the AI-screening disclosure row from `modes/oferta.md` still fails it, and restoring the row passes it.

## Verification

`node test-all.mjs --quick`: 8686 passed, 0 failed, with the same 16 warnings `main` already has. Both governing assertions are green, including the parity check, which still reports `da, es, pl, pt, ua` as the frozen set with the allowlist untouched.

## On the translations

The `ru` block follows the file @nikitacometa re-synced in #3830. I wrote the `zh`, `zh-TW`, `ja` and `ar` blocks to match each file's existing register, and the heading nouns are the part most worth a native reader's eye: `结论`, `結論`, and English `Verdict` in `ja` and `ar`, whose report skeletons are already in English. Happy to take corrections on any of them.


rubicon:
Commit `4de87b57` closes a gap in this PR. The lead block joined the report contract here with nothing asserting it.

## What was measured

Eight mode files carry the full A to H skeleton plus `## Risk Summary`. This PR gave six of them the lead block. Two were left without it:

| file | `(lead)` heading |
|---|---|
| `modes/oferta.md` | yes |
| `modes/ar/fursah.md` | yes |
| `modes/ja/kyujin.md` | yes |
| `modes/ru/oferta.md` | yes |
| `modes/zh/oferta.md` | yes |
| `modes/zh-TW/oferta.md` | yes |
| `modes/es/oferta.md` | **no** |
| `modes/tr/is-ilani.md` | **no** |

Eleven other localized evaluation modes sit at the pre-Block-G shape. They stop at `## G)` with no `## H)` and no `## Risk Summary`, so the lead block has no position in them yet.

## Why this is a separate guard

The localized parity walk near line 4155 discovers `modes/*/oferta.md`. It reaches eight locales. `tr/is-ilani.md`, `ja/kyujin.md` and `ar/fursah.md` never enter that walk, so a `REQUIRED_HEADINGS` entry could never have caught the tr case.

`REQUIRED_HEADINGS` also matches literal substrings. The heading noun is translated per locale, and every translated mode quotes the English form in prose as an example of the convention. Measured: strip the `(lead)` marker from zh's real heading and `text.includes('## Verdict (lead)')` still returns true, because line 183 mentions the string inside a paragraph.

The new guard keys on the report skeleton and matches `(lead)` on a heading line, which is the marker `cleanHeading` in `web/src/lib/report-sections.mjs` already reads. A mode at the pre-Block-G shape drops out by carrying no `## H)`, so there is no allowlist to maintain.

## Order of work

The guard went in first and was observed failing:

```
❌ evaluation modes missing the lead verdict block: modes/es/oferta.md (no `(lead)` heading); modes/tr/is-ilani.md (no `(lead)` heading)
📊 Results: 9206 passed, 1 failed, 3 warnings
```

Then `modes/es/oferta.md` got `## Veredicto (lead)` and `modes/tr/is-ilani.md` got `## Karar (lead)`, each with the section prose and a report-template entry between Block G and Risk Summary.

```
✅ Verdict (lead) block present and positioned in all 8 full-skeleton evaluation modes
📊 Results: 9207 passed, 0 failed, 3 warnings
```

Warning count held at 3 across both runs.

## Mutations

Five, each with the edit proven to land before the guard ran:

| mutation | result |
|---|---|
| strip the marker from tr's headings | ❌ names `modes/tr/is-ilani.md` |
| strip the marker from es's headings | ❌ names `modes/es/oferta.md` |
| strip the marker from zh's heading, leave the prose mention | ❌ names `modes/zh/oferta.md` |
| move the canonical block below Risk Summary | ❌ order `G)` then lead then Risk Summary |
| rename `modes/tr/is-ilani.md` | ❌ walk found 7 full-skeleton modes, expected 8 or more |

The third one is the reason the check reads a heading line. A literal substring check passes that mutation.

rubicon:
Correcting a claim in `4de87b57`'s message and in my comment above.

The commit says a prose mention stops counting once the check matches a heading line. That is wrong about fenced examples. All eight of these files illustrate the block inside a fenced example, and those illustrations carry heading lines too.

`modes/tr/is-ilani.md` is where it bites. It is the one full-skeleton mode whose report template sits before its prose. So `lastIndexOf('## G)')` landed at line 140 and `lastIndexOf('## Risk Summary')` at line 282. The window swallowed the file.

Measured on `4de87b57`: delete the `## Karar (lead)` entry from tr's report template and the guard stays green. The fenced illustration at line 254 satisfies it.

My mutation table hid this. The tr row stripped all three of that file's markers at once. That cannot separate the template from the documentation around it.

What holds after `08e4c3e3`. The heading-line match still earns its place, since it stops an inline `` `## Verdict (lead)` `` in a sentence from counting. The window does the rest. The template is the one run where `## G)` is followed by `## Risk Summary`, so the lead heading has to sit inside it. Deleting the single in-window heading from each of the eight files in turn now reddens the guard eight times.

`08e4c3e3` also brings `batch/batch-prompt.md` inside the guard. On `4de87b57`, removing its `## Verdict (lead)` list entry left the suite at 9207 passed, 0 failed, exit 0. A second assertion now pins that entry between Block G and `## Risk Summary`.

`node test-all.mjs --quick` on `08e4c3e3`: 9208 passed, 0 failed, 3 warnings, exit 0. Base `de4d2853` measures 9206 passed with the same 3 warnings. The PR body carried stale figures and now matches these.

PR #3963, opened 2026-09-07
fix(verify-pipeline): merge provider plugins before the coverage check

JOduMonT:
## Problem

`verify-pipeline.mjs` loads only `providers/` when resolving `portals.yml` entries, while `scan.mjs` — the thing the coverage check exists to model — calls `mergeProviderPlugins()` first.

So any entry pointing at a **plugin-supplied** provider is reported as an unknown-provider **error**, even though the scan resolves it fine. Since unknown-provider is a hard error, `verify-pipeline` exits 1 and prints "fix before proceeding" over a portal that is not actually broken.

## Reproduction

With a provider plugin enabled in `config/plugins.yml` and a `portals.yml` entry naming it:

```
❌ portals.yml: "<plugin-id>" sets an unknown provider — unknown provider: <plugin-id>.
   The entry never scans (see providers/ for valid ids)
📊 Pipeline Health: 1 errors
```

Calling `resolveProvider()` on the same entry with and without the merge:

```
WITHOUT plugin merge (verify-pipeline path): {"error":"unknown provider: <plugin-id>"}
WITH plugin merge (scan.mjs path):           resolved -> <plugin-id>
```

## Fix

Call `mergeProviderPlugins(providers, { root: CAREER_OPS })` after `loadProviders()`, exactly as `scan.mjs` does.

`mergeProviderPlugins()` returns immediately when `config/plugins.yml` is absent and is internally fail-open, so a checkout with no plugins configured behaves byte-identically to before.

## Testing

Full suite on a checkout with a provider plugin enabled: **95 → 87 failures, 8433 → 8456 passing**. `verify-pipeline` goes from 1 error to 0 on the same tree.

## Summary

`verify-pipeline.mjs:8` now calls `mergeProviderPlugins()` after loading providers.

Users can configure `portals.yml` entries with providers supplied by enabled plugins. The coverage check now resolves these providers instead of reporting `unknown-provider` errors. Behavior is unchanged when no plugins are configured.

## Files changed

- `verify-pipeline.mjs:8`

## Validation

A checkout with the `job-digest` provider plugin reduced `verify-pipeline` errors from 1 to 0.

### DIFF
--- a/verify-pipeline.mjs
+++ b/verify-pipeline.mjs
@@ -531,6 +531,7 @@ if (!existsSync(PORTALS_FILE)) {
   try {
     const { findUnclaimedEntries } = await import('./audit-portals.mjs');
     const { loadProviders } = await import('./providers/_registry.mjs');
+    const { mergeProviderPlugins } = await import('./plugins/_engine.mjs');
     const yaml = await import('js-yaml');
 
     const cfg = yaml.load(readFileSync(PORTALS_FILE, 'utf-8')) || {};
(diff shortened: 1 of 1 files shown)


santifer:
Closing this one, @JOduMonT, and I want to be clear about why, because your diagnosis was right.

You spotted on 7 Sep that `verify-pipeline.mjs` built its provider map from `loadProviders()` alone while `scan.mjs` also folds in `mergeProviderPlugins()`. So a `portals.yml` entry pointing at a plugin-supplied provider came back as an unknown-provider ERROR, while the scan that check exists to model resolved the same entry fine. That is exactly the bug, and it is fixed in `main`.

It landed through #4046 earlier today. That change covers both verifiers instead of one, and roots them at the code root rather than the data root, which is what matters on an install where those two differ. Yours resolved from the data root, so the wider fix is the one that could go in.

I credited your diagnosis on that thread rather than letting it land silently, and it is worth saying here too: this was found first, and by you. Thank you.

Issue #3523, opened 2026-08-30
CV payloads with unrecognised section keys render empty blocks while the report says "valid": true


novaxertz:
## What happens

`build-cv-html.mjs` and `build-cv-latex.mjs` read each list section off fixed key names (`buildEducation()` reads `e.title` / `e.org` / `e.year` / `e.description`). A payload that supplies different key names is not rejected — every field resolves to `undefined`, the section renders as an empty `<div class="edu-item">` block (or a `\resumeSubheading` full of empty braces), and the JSON report still ends in `"valid": true`.

Nothing in the pipeline catches it. The unresolved-placeholder guard passes, because the placeholders *were* filled — with empty strings.

## Why it is easy to hit

The two builders do not share a schema, and neither the scripts nor the mode docs say so:

| Section | `build-cv-html.mjs` | `build-cv-latex.mjs` |
|---|---|---|
| education | `title`, `org`, `year`, `description` | `institution`, `degree`, `dates`, `coursework` |
| projects | `name`, `badge`, `tech`, `description` | `name`, `context`, `dates`, `bullets` |
| certifications | `title`, `org`, `year` | *(no section)* |

An agent that has recently read `modes/latex.md` and then builds an HTML payload writes `{institution, degree, dates}` for education — which is exactly the shape that renders nothing. `modes/pdf.md` and `modes/latex.md` each document their own table, but neither warns that the other exists.

## Impact

This is a silent, output-only failure: the exit code is 0, the report says valid, and it is only visible to a human looking at the finished PDF. In my case four applications went out with no education section at all.

The report's `counts` block does include `educationEntries`, but it counts payload entries, not rendered ones, so it reads `1` for an entry that rendered nothing.

## Reproduce

```bash
cat > /tmp/cv.json <<'JSON'
{"lang":"en","candidate":{"name":"A B","email":"a@b.co"},"summary":"S",
 "competencies":["X"],
 "experience":[{"company":"C","role":"R","dates":"2024","bullets":["b"]}],
 "projects":[],
 "education":[{"institution":"Test University","degree":"BSc CS","dates":"2024"}],
 "certifications":[],
 "skills":[{"category":"L","items":"Node"}]}
JSON
node build-cv-html.mjs /tmp/cv.json /tmp/cv.html
```

Exit 0, `"valid": true`, and `/tmp/cv.html` contains an education block with no text in it.

## Suggested fix

[...]

PR #3965, opened 2026-09-07
feat(pdf): multilingual CV rendering — running footer, engagements, skill notes

### LINKED ISSUE #3961: feat(pdf): multilingual CV rendering — running footer, engagements, skill notes
The HTML CV builder can't currently express four things a multilingual / consulting-shaped CV needs. Each is opt-in and additive — nothing changes for an existing payload.

- **Running footer** (opt-in `--running-footer`): name, a contact method and "N of TOTAL" on every rendered page, so a page separated from the rest is still traceable back to the candidate.
- **`experience[].engagements`**: a "Selected engagements" sub-list for a role that bundles several named client/project stints under one employer — common in consulting, and currently forced into ordinary bullets where it reads as unrelated achievements.
- **`experience[].keepTogether`**: opt-in, default `false`. Stops one bullet-heavy entry being split across a page break.
- **`skills[].note`**: an optional qualifier line rendered under a category's items — the honest place for "working familiarity" instead of silently implying equal depth across everything listed.
- **Multi-paragraph `summary`**: a literal `\n` becomes a line break and `\n\n` a paragraph gap, plus NBSP handling for French punctuation (a negative lookahead exempts NBSP directly before `:;!?`, which ATS parsers mangle).

Motivation: rendering a CV for a bilingual consulting profile currently means either flattening real client engagements into bullets, or hand-editing the generated HTML — which defeats the point of the payload being the source of truth.

Implementation is written and tested against current `main` (`fe06051b`), including `skills[]`/`experience[]` schema updates in `lib/cv-payload-schema.mjs` so the strict key validator from #3523 doesn't warn on the new keys. Happy to send it as a PR if the direction is right — and equally happy to be redirected if this belongs in a plugin instead of core.

JOduMonT:
Closes #3961.

## What

Five additions to the HTML CV builder, all opt-in and additive — an existing payload renders byte-identically.

| Feature | Shape |
|---|---|
| Running footer | `--running-footer` flag: name, a contact method, "N of TOTAL" on every page |
| Selected engagements | `experience[].engagements: [{ name, detail }]` |
| Keep entry whole | `experience[].keepTogether: boolean` (default `false`) |
| Skill qualifier | `skills[].note: string` |
| Multi-paragraph summary | literal `\n` → `<br>`, `\n\n` → paragraph gap |

## Why

Rendering a CV for a bilingual consulting profile currently means flattening real client engagements into ordinary bullets — where they read as unrelated achievements — or hand-editing generated HTML, which defeats the payload being the source of truth.

`skills[].note` exists for honesty: it's the place to say "working familiarity" instead of silently implying equal depth across everything in a category.

## French typography

Multi-paragraph summary support comes with NBSP handling: a negative lookahead exempts NBSP directly before `:;!?` (French spacing) while still normalising stray NBSP to a plain space for ATS safety, continuing #1728. Covered by `tests/nbsp-french-punctuation.test.mjs`.

## Second commit

`fix(cv-schema)` widens `ENTRY_FIELD_SPECS` for the three new keys. Without it the strict key validator from #3523 warns "a key no builder reads" on every payload using them — even though the builder renders them.

## Notes for review

- Both `templates/cv-template.html` and the ATS variant are updated; `keepTogether` and `engagements` are CSS-only opt-ins that no existing payload triggers.
- `--running-footer` composes with the existing `--skip-fact-check`; both are listed in the usage block.
- The `modes/pdf.md` schema table gains rows for each new key, alongside upstream's #3523 key-enforcement section.

## Testing

`node build-cv-html.mjs --test` renders valid with zero warnings. Full suite against current `main`: no failures attributable to this change.

## Summary

Users can create multilingual and consulting-focused CVs from one payload.

- Add `--running-footer` for candidate details and localized page numbers on multi-page PDFs: `generate-pdf.mjs`
- Add `experience[].engagements` and `experience[].keepTogether`: `build-cv-html.mjs`, `templates/sections/experience.html`, `templates/ats/sections/experience.html`
- Add `skills[].note` and multiline summaries: `build-cv-html.mjs`, `templates/sections/skills.html`
- Preserve French NBSP before `:;!?` during ATS normalization: `generate-pdf.mjs`, `tests/nbsp-french-punctuation.test.mjs`
- Accept the new fields in strict schema validation: `lib/cv-payload-schema.mjs`
- Document the options and newline behavior: `modes/pdf.md`
- Add `pdf-lib` for footer stamping: `package.json`

Existing payloads remain unchanged. No changes were made to `AGENTS.md`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, or `.github/`.

### DIFF
--- a/build-cv-html.mjs
+++ b/build-cv-html.mjs
@@ -61,12 +61,27 @@ const DEFAULT_SECTION_TITLES = {
   awards: 'Awards & Honors',
   interests: 'Interests',
   skills: 'Skills',
+  engagements: 'Selected engagements',
 };
 
 // Escape user text for HTML text/attribute context. Covers the five characters
 // that change meaning in markup so tailored bullets containing &, <, >, quotes
 // (e.g. "R&D", "scaled 10x < budget", 'the "north star" metric') render as
 // literal text instead of breaking the document or injecting tags.
+// Converts a literal newline in payload.summary into a <br> line break, so a
+// multi-paragraph summary can be written as plain text with a blank line
+// between paragraphs — "\n\n" becomes "<br><br>", a visible gap (#3961
+// follow-up). Escaping runs first, the same safety ordering already used for
+// the "**bold**" convention (see modes/pdf.md): only a literal newline the
+// agent typed is converted afterward, nothing already escaped is
+// re-interpreted.
+// Shared by SUMMARY_TEXT and skill-category items (#3961 follow-up, page-fit):
+// a literal "\n" becomes a line break, "\n\n" a paragraph gap, and escaping
+// runs first so a "<script>" typed into either field stays inert.
+function renderMultilineText(text) {
+  return escapeHtml(text).replace(/\n/g, '<br>');
+}
+
 function escapeHtml(text) {
   // Blank out only truly absent/structural values. A number or boolean scalar
   // (e.g. a payload with `year: 2024` instead of `"2024"`) must render its value,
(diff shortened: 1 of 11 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is in `generate-pdf.mjs`, a red-line file and the maintainer's call; I am reviewing the owned test `tests/nbsp-french-punctuation.test.mjs`.

It is sound. The test imports the real `normalizeTextForATS` from generate-pdf.mjs and exercises it over realistic HTML strings, driving the exact production line under test (the `/ (?![:;!?])/g` negative-lookahead from the #3961 follow-up), not a hand-rolled inserter. The assertions pin the correct codepoint: NBSP is verified as U+00A0 (not U+202F narrow NBSP, not a plain space), and it checks both that the literal char survives and that the replacement tally stays falsy, so it holds both directions. The negative cases (a mid-number `10 000`, and a word followed by a non-punctuation `:` later in the string) confirm the lookahead is anchored rather than "punctuation somewhere ahead". I mutation-checked it: dropping the lookahead so all NBSP flatten reddens eight assertions (11 passed to 3 passed / 8 failed via the real harness).

One scope note for the author: the fix targets U+00A0 before `:;!?` only; it does not handle guillemets or `/`, though the PR title mentions them. The test correctly matches the implementation, so this is a scope observation, not a test defect.

Owned test sound, not behind. SIGNAL (generate-pdf.mjs red-line), so not surfacing as ready-to-approve; routing and merge are the maintainer's.

Scott-Emberson (commented):
Re-verified after the head moved. This is a pure rebase: both `tests/nbsp-french-punctuation.test.mjs` and `generate-pdf.mjs` are byte-identical to the previously-verified head, so the earlier verdict carries. I re-ran the owned test at the new head to confirm: it passes 11 of 11 via the real harness, imports the real `normalizeTextForATS`, pins the U+00A0 codepoint (not U+202F, not a plain space), and reddens 8 assertions when the `(?![:;!?])` lookahead is dropped.

Owned test sound, not behind. SIGNAL (generate-pdf.mjs red-line), so not surfacing as ready-to-approve; routing and merge are the maintainer's.

PR #3968, opened 2026-09-07
fix(tests): a killed test run leaves no trap for the next one

### LINKED ISSUE #3940: test-all: a leftover .tmp-script-test-* scratch copy turns the next local run red (264 failures in the layout guard)
### What happened

A full `node test-all.mjs` on a clean `main` (1d4951f6) failed 264 checks locally while CI and a detached-worktree preview of the same commits were green (8626 passed, 0 failed). Every failure named a path under `.tmp-script-test-OP9Bzd/.tmp-script-test-tBjFgy/…`:

```
❌ .tmp-script-test-OP9Bzd/.tmp-script-test-tBjFgy/company-history.test.mjs is a *.test.mjs outside tests/ and web/tests/: no runner discovers it, move it under tests/
…
❌ tests/mjs-files.test.mjs — node:test suite failed (exit 1)
```

Moving the two directories out of the checkout and re-running: `9312 passed, 0 failed`.

### Why

- `test-all.mjs:409` copies the whole checkout into `mkdtempSync(join(ROOT, '.tmp-script-test-'))` for the script smoke tests and removes it at `:661`. A run killed between the two (timeout, Ctrl-C, a crash) leaves the copy behind; it is gitignored, so nothing notices. The two leftovers here were dated 7-aug and 15-aug.
- The copy excludes `.git` (`:415`, `EXCLUDE_AT_ANY_DEPTH`), so `isNestedCheckout()` in `lib/mjs-files.mjs:84` cannot recognise it, and `SKIP_DIRS` (`lib/mjs-files.mjs:40`) has no entry for the suite's own scratch prefix. Every walker then reads the stale copy as part of the repository. The layout guard from #3863 is the loudest consumer, but `tests/mjs-files.test.mjs` fails the same way.
- Copies compound: the copy loop does not exclude an existing `.tmp-script-test-*` either, which is how one scratch ended up nested inside the next.

### Fix shape

1. `lib/mjs-files.mjs`: skip any top-level directory whose name starts with `.tmp-script-test-`, and apply the same rule in the copy loop at `test-all.mjs:415` so a stale scratch is never copied into the next one.
2. `test-all.mjs`: sweep stale `.tmp-script-test-*` at startup, or register the removal on `process.on('exit')`, so a killed run does not poison the next one.

#3904 and #3908 touch the same walker; whoever lands first can carry (1). Repro without waiting for a crash: create `.tmp-script-test-x/test/` in the checkout, copy any `*.test.mjs` into it, run `node test-all.mjs`.

yuki-uix:
A `test-all.mjs` run that gets killed leaves a full copy of the checkout behind, and that copy turns the **next** run red: 264 failures locally against 0 in CI, on a clean `main`, every one of them naming a file that is in the right place.

Closes #3940

---

## Why it happens

Section 75 copies the whole checkout into a scratch directory so the script smoke tests can run against a tree they are free to damage, and removes it in a `finally`:

```js
const scriptTmp = mkdtempSync(join(ROOT, '.tmp-script-test-'));
try   { /* run the CLIs against the copy */ }
finally { rmSync(scriptTmp, { recursive: true, force: true }); }
```

A timeout, a Ctrl-C or a crash never reaches that `finally`, so the copy stays. Nothing then notices it, because both mechanisms that would have are blind to this particular directory:

| | why it misses the leftover |
|---|---|
| `git status` | `.gitignore` carries `.tmp-script-test-*/` |
| `isNestedCheckout()` | the copy excludes `.git`, so there is no marker to detect |

Every repo-walking guard therefore reads the copy as source. The loudest consumer is the layout guard, which requires each `*.test.mjs` to live under `tests/` — and a copied suite's path is `.tmp-script-test-OP9Bzd/tests/foo.test.mjs`, which does not start with `tests/`. One leftover, several hundred violations.

It also compounds: the copy loop did not exclude an existing scratch either, so each run wrapped the previous leftover inside its own — the nested `.tmp-script-test-OP9Bzd/.tmp-script-test-tBjFgy/…` in the report.

**Reproducing takes three commands, no crash required:**

```bash
mkdir -p .tmp-script-test-x/tests
cp tests/core-test-layout.test.mjs .tmp-script-test-x/tests/
node tests/core-test-layout.test.mjs
```

`all 377 *.test.mjs files live under tests/` becomes a red line per file, while `git status` stays empty throughout. That invisibility is most of the cost here: the run is red locally and green in CI, it points at correct files, nobody else can reproduce it, and the cause does not appear in any command you would think to run. For a first-time contributor it reads as "I changed nothing and the suite broke".

## The fix

`lib/scratch-dirs.mjs` (new) owns the prefix and the predicate. It was previously written twice — the `mkdtempSync` call and `.gitignore` — and had begun to drift: one of them knew the directory existed, the other did not.

Three consumers, with clearly separated jobs:

| layer | file | job |
|---|---|---|
| **1. Correctness** | `lib/mjs-files.mjs` | `collectMjsFiles` skips a scratch directory, so no guard ever reads one |
| **2. No compounding** | `test-all.mjs` (copy loop) | a stale scratch is never copied into the new one |
| **3. Housekeeping** | `test-all.mjs` (startup) | sweep old leftovers to reclaim the disk |

**Layer 1 is the fix.** The walk is the single entry point every guard shares, so excluding the directory once immunises all of them — a new guard inherits it instead of rediscovering the bug. Same shape as the nested worktree in #3499, different marker. This layer alone removes the 264 failures; the leftover may still sit on disk, but nothing grades it.

**Layer 3 only reclaims disk**, and that separation is what lets it be careful. Without care, a second `test-all.mjs` in the same checkout has its live scratch deleted underneath it — trading a false red for a real one. So the sweep asks two questions in order:

> **Is anybody using this?** A run writes its own pid into the scratch before filling it, and `kill(pid, 0)` answers. A directory with a living owner is not removed.
>
> **Failing that, how old is it?** For a leftover with no usable marker — everything predating this mechanism — an hour, against a full suite that takes ~4 minutes locally. The leftovers in the report were dated 7-aug and 15-aug.

Liveness comes first because age is only a proxy for it, and a wrong one in the direction that costs something: mtime stops advancing when the copy finishes, so a run that then blocks ages while it is alive. The pid claim is itself bounded at a week, since a pid freed by a crash can be reused and would otherwise protect a dead run's tree forever.

The sweep runs **before** the `--only` early exit, because the 264-failure run is exactly the one a developer then re-runs with `--only core-test-layout` to look closer.

### Files

| file | |
|---|---|
| `lib/scratch-dirs.mjs` | new — the prefix and predicate, the owner marker and its liveness check, the sweep |
| `lib/mjs-files.mjs` | +1 line in the walk |
| `test-all.mjs` | import, startup sweep, copy-loop exclusion, owner marker, `mkdtemp` uses the shared constant |
| `tests/scratch-dirs.test.mjs` | new — 11 assertions |
| `update-system.mjs` | +1 line registering the new library file in `SYSTEM_PATHS` |

## How it is verified

Full suite: **8701 passed, 0 failed**.

`main` measures 8687 in a fresh worktree, and the 14 between them is worth spelling out because two of the three parts are not this PR's assertions: 11 from the suite below, 2 from the syntax gate running over two added files, and 1 from `cv-sync-check.mjs`, which passes here and warns in a fresh worktree because that worktree has no user-layer data. The warning count moving 16 → 15 is the same event seen from the other side. A baseline taken in a different checkout is not a pinned comparison, so the number is only useful with that caveat attached.

| # | what it pins | negative control |
|---|---|---|
| 1 | `isScratchDir` matches the mkdtemp prefix and nothing adjacent to it | — |
| 2 | the walk skips a scratch copy and still returns the real tree | ✅ |
| 3 | the sweep removes every stale scratch and touches nothing else | — |
| 4 | a scratch that cannot be removed is reported, not thrown | — |
| 5 | `.gitignore` still ignores a directory named from the constant | ✅ |
| 6 | a concurrent run's fresh scratch is kept, the stale one swept | ✅ |
| 7 | a symlink named like a scratch is neither followed nor removed | ✅ |
| 8 | a FILE carrying the prefix stays in the walk; only the directory is skipped | ✅ |
| 9 | a live owner keeps its scratch past the age gate; a dead one does not | ✅ |
| 10 | a missing or unparseable owner marker falls through to the age gate | — |
| 11 | a living owner is believed, but not past the ceiling | ✅ |

Seven were checked against their own negative — the fix removed, the assertion watched to go red — rather than assumed to work. The two that matter most:

- **remove the walk's skip** → `collectMjsFiles returned [".tmp-script-test-OP9Bzd/tests/real.test.mjs","tests/real.test.mjs"]`, the source of the original 264
- **swap `entry.isDirectory()` for a link-following `statSync`** → `removed=[".tmp-script-test-symlink"]`

That last probe backdates the link's **target**, not just its parent: a fresh link clears the age gate, and a broken sweep would pass on that technicality instead of reaching the delete path. Measured blast radius, stated in the test rather than overclaimed — the link goes, its target survives, because `rmSync` unlinks a symlink rather than recursing through it.

Two deliberate choices in the test:

- **Ages are set with real `utimesSync`, not a stubbed clock.** The gate reads the filesystem; a stub would let an implementation reading the wrong field (`birthtime`, `ctime`) pass.
- **The `.gitignore` pairing asks git**, via `check-ignore`, rather than grepping for the rule's text. `.gitignore` cannot import the constant, so the two are a hand-maintained pair — and a grep would still pass on a rule that had stopped matching anything, which is the failure worth catching.

End to end, on the path a person actually takes: with a leftover present, `node test-all.mjs --only scratch-dirs` prints `🧹 removed a stale scratch copy from an interrupted run: …` and the directory is gone.

## Notes for review

1. **Overlap with #3908 and #3904 — bigger than I first said.** I originally wrote that the contended surface was a single line in `lib/mjs-files.mjs`. @artemtrofymenko measured it and that was wrong; I re-measured and got the same result:

   ```
   3968 x 3904 -> CONFLICT: tests/helpers.mjs, tests/mjs-files.test.mjs, tests/mojibake-canary.test.mjs
   3968 x 3908 -> CONFLICT: lib/mjs-files.mjs, test-all.mjs, update-system.mjs
   ```

   So #3908 overlaps in all three production files, not one line. Nothing here is a defect in any of the three branches — it is routine for work off one file — but it makes the ordering worth choosing deliberately rather than by arrival. I have no preference and will rebase onto whatever lands first; say the word and I will go after #3908 instead.

2. **The copy-loop exclusion has no direct automated test.** `copyDirSync` is inline in `test-all.mjs` and not importable, and extracting it felt like scope creep on a file two open PRs are already rewriting. It is also the second gate rather than the first — the startup sweep normally removes a leftover before the copy runs. Stating the gap rather than leaving it to be found.

3. **A flake in `update-system.mjs check`, now identified.** An earlier run reported one unexplained failure. It has since reproduced and named itself: `update-system.mjs check crashed (exit null, signal SIGTERM)` — the 30s script timeout firing on a check that reaches the network. It appeared three runs running during one window and has been green either side of it, on this branch and on `main`. Confirmed not to be from this change by running `main` and this branch alternately: `main` 8687/0, this branch 8701/0, same window. f4c28e6's message names the same flake, so it predates this PR.

4. **Windows.** The symlink section originally used a `dir` symlink, which needs a privilege a default non-elevated Windows box does not have — so it threw, took the four sections after it, and `windows-latest` never showed it. Fixed with a junction (28cc350), and link creation is no longer fatal to the rest of the file. That whole finding came from @artemtrofymenko running it on a real machine; CI could not have told me.

## One open question for the maintainer

The sweep believes a living-owner claim, but overrules it after seven days. Two review rounds asked for opposite things here and both were right about their own half:

Without the ceiling, kill(pid, 0) cannot tell a live owner from a recycled pid, so a crashed run's copy can be protected forever — an unbounded disk leak.
With the ceiling, a test-all.mjs run alive for more than seven days would lose its scratch to another run's sweep.

I kept the ceiling, because the first failure needs a crash plus a pid wrap (which happens on a busy machine over months) and the second needs a four-minute suite to stay alive for a week. Neither loses authored content; a scratch is a copy, so the cost is a failed run or reclaimable disk.

The principled fix is a non-reusable owner identity, and I could not find a cheap portable one: process start time has no cross-platform Node API, an open handle does not prevent unlink on POSIX, and a heartbeat stops when a debugger pauses the very run it is meant to protect. A platform-branching ps / tasklist probe would work — say the word if that is proportionate here and I will add it.

## Summary

This PR prevents interrupted `test-all.mjs` runs from causing false layout failures in later runs.

## User-visible changes

- Repository scans skip scratch directories: `lib/mjs-files.mjs:176`.
- New scratch copies skip existing scratch directories: `test-all.mjs:479`.
- Startup cleanup removes stale scratch directories and preserves active runs: `test-all.mjs:226`, `lib/scratch-dirs.mjs:194`.
- Cleanup handles invalid markers, symlinks, and removal failures without stopping the test run.
- Tests cover detection, walking, cleanup, age handling, concurrency, process liveness, `.gitignore`, and symlinks: `tests/scratch-dirs.test.mjs`.

## Files touched

- `lib/mjs-files.mjs`
- `lib/scratch-dirs.mjs`
- `test-all.mjs`
- `tests/scratch-dirs.test.mjs`
- `update-system.mjs:239`: adds `lib/scratch-dirs.mjs` to `SYSTEM_PATHS`.

The requested system files were not changed: `AGENTS.md`, `modes/`, `DATA_CONTRACT.md`, `providers/`, and `.github/`.

## Validation

Full suite: 8697 passed, 0 failed.

### DIFF
--- a/lib/mjs-files.mjs
+++ b/lib/mjs-files.mjs
@@ -23,6 +23,7 @@
 
 import { existsSync, readdirSync } from 'fs';
 import { join } from 'path';
+import { isScratchDir } from './scratch-dirs.mjs';
 
 /**
  * Directories excluded from the walk BY NAME.
@@ -158,6 +159,15 @@ export function collectMjsFiles(root) {
     }
     for (const entry of entries) {
       if (SKIP_DIRS.has(entry.name)) continue;
+      // A scratch copy left behind by an interrupted `test-all.mjs` is a second
+      // copy of this repository that neither exclusion above can see: it is
+      // named unpredictably, so SKIP_DIRS cannot list it, and it is copied
+      // without `.git`, so isNestedCheckout() below does not recognise it
+      // either. Every consumer of this walk then grades a stale duplicate as
+      // source — same shape as the nested worktree in #3499, different marker
+      // (#3940). Skipped here rather than per consumer so a new walker inherits
+      // the exclusion instead of rediscovering the bug.
+      if (isScratchDir(entry.name)) continue;
       // Symlinked directories can point outside the checkout or back into it;
       // neither should make the walk recurse unpredictably.
       if (entry.isSymbolicLink()) continue;
--- a/lib/mjs-files.mjs
+++ b/lib/mjs-files.mjs
@@ -167,7 +167,13 @@ export function collectMjsFiles(root) {
       // source — same shape as the nested worktree in #3499, different marker
       // (#3940). Skipped here rather than per consumer so a new walker inherits
       // the exclusion instead of rediscovering the bug.
-      if (isScratchDir(entry.name)) continue;
+      //
+      // Directories only. The name is a prefix, so a FILE could carry it too —
+      // and skipping `.tmp-script-test-something.mjs` would quietly drop it from
+      // the syntax gate, which is the "gate covering a file set nobody chose"
+      // failure this module exists to remove. Only a directory can hold a copy
+      // of the repository, so only a directory is what this excludes.
+      if (entry.isDirectory() && isScratchDir(entry.name)) continue;
       // Symlinked directories can point outside the checkout or back into it;
       // neither should make the walk recurse unpredictably.
       if (entry.isSymbolicLink()) continue;
--- a/lib/scratch-dirs.mjs
+++ b/lib/scratch-dirs.mjs
@@ -1,0 +1,175 @@
+/**
+ * scratch-dirs.mjs — the one definition of "a throwaway copy of this checkout".
+ *
+ * Section 75 of `test-all.mjs` copies the whole checkout into
+ * `mkdtempSync(join(ROOT, '.tmp-script-test-'))` so the script smoke tests can
+ * run against a tree they are free to damage, and removes it in a `finally`. A
+ * run killed between the two — a timeout, Ctrl-C, a crash — never reaches that
+ * `finally` and leaves the copy behind.
+ *
+ * Nothing then notices. `.gitignore` carries a directory rule for the prefix, so
+ * `git status` is empty; the copy excludes `.git`, so `isNestedCheckout()` in
+ * `mjs-files.mjs` cannot recognise it either. Every walker reads the stale copy
+ * as part of the repository, and the guards that grade file LAYOUT are the ones
+ * that break: a full second copy of `tests/` is, to
+ * `tests/core-test-layout.test.mjs`, several hundred suites sitting outside
+ * `tests/`. Measured on a clean `main`: 264 failures locally against 0 in CI,
+ * naming paths under `.tmp-script-test-OP9Bzd/.tmp-script-test-tBjFgy/…`
+ * (#3940). The nesting in that path is the second half of the bug — the copy
+ * loop did not exclude an existing scratch either, so each run wrapped the
+ * previous leftover inside its own.
+ *
+ * That failure shape is worse than a plain false red. It is unreproducible for
+ * anyone else, it points at correct files, it appears on a commit the developer
+ * did not write, and the state causing it is invisible to every command they
+ * would think to run.
+ *
+ * The prefix lives here rather than at its three use sites because it was
+ * already written twice — the `mkdtempSync` call and `.gitignore` — and had
+ * begun to drift: one of them knew the scratch directory existed and the other
+ * did not. A consumer that must recognise a scratch directory imports
+ * `isScratchDir`; it cannot be spelled slightly differently in a fourth place.
+ *
+ * `.gitignore` is the one copy that cannot import this, since git reads it. Its
+ * rule is the prefix followed by a glob and a directory slash, and
+ * `tests/scratch-dirs.test.mjs` pins the two against each other by asking git
+ * itself whether it ignores a directory named from this constant — a textual
+ * comparison would pass on a rule that no longer matches anything.
+ */
+
+import { readdirSync, rmSync, statSync } from 'fs';
+import { join } from 'path';
+
+/**
+ * The prefix `mkdtempSync` is called with, and therefore the only thing a
+ * scratch directory is guaranteed to have in common with the next one — the
+ * six random characters after it are the point of `mkdtemp`.
+ */
+export const SCRATCH_PREFIX = '.tmp-script-test-';
+
+/**
+ * Is this directory NAME a throwaway copy left by the script smoke tests?
+ *
+ * Takes a bare name, not a path, because both walkers that consult it are
+ * already iterating `readdirSync` entries and a path-taking predicate would
+ * invite a caller to pass a path and match a legitimate directory that merely
+ * sits inside a scratch tree.
+ *
+ * @param {string} name - A single path segment.
+ * @returns {boolean}
+ */
+export function isScratchDir(name) {
+  return name.startsWith(SCRATCH_PREFIX);
+}
+
+/**
+ * How long a scratch directory must have sat untouched before a sweep will
+ * delete it.
+ *
+ * The sweep is housekeeping, not correctness. What keeps a leftover from
+ * breaking a run is `isScratchDir` in the walkers — the guards never read one
+ * whether it is swept or not. All the sweep does is reclaim the disk, so it can
+ * afford to be timid, and being timid is the whole point here: a second
+ * `test-all.mjs` running in the same checkout has a scratch directory of its
+ * own, and deleting THAT crashes a run that was doing nothing wrong.
+ *
+ * An hour, against a full suite that takes ~4 minutes locally and well under 30
+ * on the slowest CI runner. The leftovers in #3940 were dated 7-aug and 15-aug,
+ * so nothing is being kept that anyone wanted.
+ */
+export const MIN_SCRATCH_AGE_MS = 60 * 60 * 1000;
+
+/**
+ * Remove every scratch directory directly under `root` that has been untouched
+ * for at least `minAgeMs`, and say which.
+ *
+ * Called at the start of a run rather than only at the end of one: the leftover
+ * exists precisely because the previous run's cleanup did not get to execute,
+ * so a second cleanup path scheduled the same way would not have run either.
+ * The next run's startup is the first moment that is guaranteed to happen.
+ *
+ * Age is read from the directory's own mtime, which stops advancing once the
+ * copy that filled it has finished — so it measures "when did a run last touch
+ * this", not "when was it created". A live run's scratch is minutes old at
+ * most; a leftover's is as old as the crash.
+ *
+ * Top level only. A scratch directory is created directly under ROOT, so that
+ * is where a stale one is; recursing would mean walking into a tree this
+ * function is about to delete, and a `.tmp-script-test-` name deeper in the
+ * repository belongs to whoever put it there.
+ *
+ * Symlinks are never followed, and never removed: `entry.isDirectory()` is
+ * false for one, so a link named like a scratch directory is skipped before any
+ * of this. That is load-bearing — this function authorises a recursive delete,
+ * and following a link would move that delete somewhere nobody asked for — so
+ * `tests/scratch-dirs.test.mjs` pins it rather than leaving it to be inferred
+ * from `withFileTypes`.
+ *
+ * Returns the names rather than printing them, so the caller decides whether a
+ * sweep is worth a line of output. Silence would be wrong — a run that quietly
+ * deleted a directory is its own small mystery — but that is the caller's call
+ * to make, not this function's.
+ *
+ * A directory that cannot be removed is reported, not thrown: a sweep is
+ * housekeeping in front of the real work, and failing the whole suite because a
+ * stale copy is momentarily locked would replace a false red with another one.
+ * The walkers skip it regardless, so the run stays correct either way.
+ *
+ * @param {string} root - Absolute path to sweep.
+ * @param {object} [options]
+ * @param {number} [options.minAgeMs] - Age floor; see MIN_SCRATCH_AGE_MS.
+ * @param {() => number} [options.now] - Clock seam, so the age gate is testable
+ *   without sleeping for an hour.
+ * @param {(dir: string) => void} [options.remove] - Removal seam, so a test can
+ *   drive the failure branch without arranging an undeletable directory.
+ * @returns {{removed: string[], kept: string[], failed: {name: string, error: string}[]}}
+ */
+export function sweepScratchDirs(root, options = {}) {
+  const {
+    minAgeMs = MIN_SCRATCH_AGE_MS,
+    now = Date.now,
+    remove = (dir) => rmSync(dir, { recursive: true, force: true }),
+  } = options;
+
+  let entries;
+  try {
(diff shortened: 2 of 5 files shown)


Scott-Emberson (commented):
Reviewing the owned test `tests/scratch-dirs.test.mjs`. The substance sits in the test harness (`test-all.mjs`, the new `lib/scratch-dirs.mjs`, `lib/mjs-files.mjs`, and a one-line `update-system.mjs` registration), which is harness and red-line territory, so I am treating routing and merge as the maintainer's call and verifying the owned test only.

It is sound. The test imports the real functions from `lib/scratch-dirs.mjs` (`isScratchDir`, `sweepScratchDirs`, `markScratchOwner`, `scratchOwnerAlive`, and the prefix/age constants) and the real `collectMjsFiles` from `lib/mjs-files.mjs`, not reconstructed copies. It reproduces the killed-run-then-next-run shape (#3940): a leftover scratch dir containing a second `tests/` copy, asserting the real walker returns only the real suite, plus a `.gitignore` check via real `git check-ignore` and age-gate-versus-pid-liveness cases. I mutation-checked it: removing the scratch skip in the walker reddens the trap-poisoning cases, and bypassing the liveness gate in the sweep reddens the live-owner case (while the age-gated no-owner case correctly stays green). Ten assertions pass via the real harness.

Owned test sound, not behind. Because the substance is harness and red-line (test-all.mjs, update-system.mjs), this is test-axis only, not surfaced as ready-to-approve; routing and merge are the maintainer's.

artemtrofymenko (commented):
Ran this on default non-elevated Windows 11 at `5264db8`. The fix works. **The new suite does not run here** — it loses 4 of its 10 assertions to an `EPERM`, and `test (windows-latest)` is green, so nothing announces it.

## First, the bug and the fix, both reproduced

@santifer's recipe on `origin/main` (`f20b94c`), one planted file rather than waiting for a crash:

```
mkdir -p .tmp-script-test-x/test && cp tests/mjs-files.test.mjs .tmp-script-test-x/test/planted.test.mjs
node test-all.mjs --only core-test-layout

❌ .tmp-script-test-x/test/planted.test.mjs is a *.test.mjs outside tests/ and web/tests/:
   no runner discovers it, move it under tests/
📊 0 passed, 1 failed
```

Same plant, same command, on this branch:

```
📊 1 passed, 0 failed
```

## The finding: `tests/scratch-dirs.test.mjs:207` cannot run on a default Windows box

```js
symlinkSync(precious, link, 'dir');
```

```
❌ tests\scratch-dirs.test.mjs — suite threw and was contained (EPERM):
   EPERM: operation not permitted, symlink '…\precious' -> '…\root\.tmp-script-test-symlink'
📊 6 passed, 1 failed
```

Creating a symlink on Windows needs `SeCreateSymbolicLinkPrivilege`, which a non-elevated shell without Developer Mode does not hold — the default configuration, and the one `windows-latest` does not reproduce. The suite is section-ordered, so the throw takes everything after it: **six assertions run, four never do**, and the file-prefix case, the live-owner case and the owner-marker case are among the lost ones. This is the same class as #2828 → #3259 → #3267 and #3362.

A **junction** needs no privilege, and the link target here is a directory (`precious/work/`), so the substitute applies:

```js
symlinkSync(precious, link, process.platform === 'win32' ? 'junction' : 'dir');
```

Measured with that one line changed, nothing else:

```
✅ a symlink named like a scratch directory is neither followed nor removed
📊 10 passed, 0 failed
```

The assertion is not weakened by the swap, which is the part worth checking rather than assuming (`coverage-illusions`: a substitute that makes the check vacuous is worse than the skip it replaced). `lstat()` on a junction reports `isSymbolicLink() === true`, so `sweepScratchDirs` takes the same branch it takes for a symlink, and the case still passes for its own reason — the target survived, and `removed`, `kept` and `failed` were all empty. Measured on this box:

```
symlinkSync(…, 'dir')       → EPERM
symlinkSync(…, 'junction')  → created; lstat isSymbolicLink=true isDirectory=false
```

The comment above the line is worth keeping as-is; only the third argument needs to change.

## Merge sequencing, since @santifer flagged it on the issue

> #3904 and #3908 touch the same walker; whoever lands first can carry (1)

Measured with `git merge-tree --write-tree` between the PR heads:

```
3968 × 3904 → CONFLICT: tests/helpers.mjs, tests/mjs-files.test.mjs, tests/mojibake-canary.test.mjs
3968 × 3908 → CONFLICT: lib/mjs-files.mjs, test-all.mjs, update-system.mjs
```

Not a defect in any of the three — routine for branches off one file — but the second pair overlaps in **all three production files**, so that is the expensive ordering to get wrong. Posting the file lists so whoever sequences them is not doing it from titles.

## What I could not reproduce

The issue says `tests/mjs-files.test.mjs` fails the same way. With my plant it does not: `node --test tests/mjs-files.test.mjs` on `main` with the scratch in place gives 12/12. The reported leftovers were nested full checkout copies (`.tmp-script-test-OP9Bzd/.tmp-script-test-tBjFgy/`), so a single planted file is a narrower fixture than the real thing — I would not read my green as contradicting the report, only as not covering that half.

## Otherwise

Two things in the test file that I would have got wrong. Asking `.gitignore` coverage as a **behaviour** question — `git check-ignore` rather than a grep for the rule's text — closes the case where the rule survives but stops matching. And the age gate in section 6, so a second concurrent run's fresh scratch is not swept by the first, is the failure that would have traded #3940's false red for a real one; it is the kind of thing that is only obvious after somebody has been bitten.

Scott-Emberson (commented):
Re-verified after the head moved (`b3cffb08b0` to `933a261ffd`). This is not a pure rebase: `lib/scratch-dirs.mjs`, `test-all.mjs`, and the owned test changed, carrying three new commits on top of a main merge. The substance changes add a `MAX_SCRATCH_AGE_MS` ceiling so a living-owner claim is overruled past seven days (a pid-reuse defence), reorder the harness copy-exclude for clarity with no behavior change, and add a Windows junction fallback plus a named non-Error-throw case in the test.

The owned test imports the real `collectMjsFiles` and `sweepScratchDirs` and drives them, not reconstructed copies. It passes (11 assertions via the real harness). I mutation-checked it: removing the scratch skip in the walker reddens two, bypassing the liveness gate reddens two, and dropping the new age ceiling reddens the section that pins exactly that logic. Not a false-pass.

Owned test sound, not behind. Because the substance is harness and red-line (test-all.mjs, lib/, a one-line update-system.mjs registration), this is test-axis only, not surfaced as ready-to-approve; routing and merge are the maintainer's.

Scott-Emberson (commented):
Re-verified after the head moved again (`933a261ffd` to `f68eb65b9f`). This one is a pure rebase: the owned test, `lib/scratch-dirs.mjs`, `lib/mjs-files.mjs`, and `test-all.mjs` are all byte-identical to the previously-verified head, so the earlier verdict carries. I re-ran the owned test at the new head to confirm: `node test-all.mjs --only scratch-dirs` passes 11 of 11.

Owned test sound, not behind. Still SIGNAL (the substance is harness and red-line, test-all.mjs / lib/ / a one-line update-system.mjs registration), so this stays test-axis only, not surfaced as ready-to-approve; routing and merge are the maintainer's.

PR #3969, opened 2026-09-07
fix(pipeline): preserve unknown scores when reconciling reports

nikitacometa:
An empty `**Score:**` followed by a URL containing `4521` became `4521/5`. Numbers in `N/A` explanations could become scores too, and an empty `**PDF:**` could mark a PDF as generated from the next header.

Read each header on its own line and accept only scores from 0 to 5 on the five-point scale. Valid batch scores take precedence; bare numbers, Markdown formatting and trailing explanations remain supported.

Validation:

- 54 CLI tests pass, including 45 new cases covering rejected scores, PDF flags, dry-run, backup and repeat runs. Six mutants fail as expected.
- Full suite: 8689 passed, 0 failed with `CAREER_OPS_GIT_TIMEOUT_MS=2000 CAREER_OPS_GIT_FETCH_TIMEOUT_MS=5000 node test-all.mjs`. These bound the existing updater network timeout in the smoke fixture; CI runs with defaults.
- `npm run lint` and `git diff --check` pass.

### DIFF
--- a/reconcile-pipeline.mjs
+++ b/reconcile-pipeline.mjs
@@ -151,7 +151,7 @@ function resolveScore(stateScore, reportFile) {
     const denominator = rep.match(/\/[ \t]*(\d+(?:\.\d+)?)/);
     if (denominator) {
       const tail = rep.slice(denominator.index + denominator[0].length);
-      if (Number(denominator[1]) !== 5 || (tail && !/^[\s*_`,;(—–-]/.test(tail))) return 'N/A';
+      if (Number(denominator[1]) !== 5 || (tail && !/^[\s*_`,;:.()\]—–-]/.test(tail))) return 'N/A';
     } else if (/^[*_`]*[ \t]*\//.test(rep.slice(score[0].length))) {
       return 'N/A'; // An explicit but unreadable scale is not a bare score.
     }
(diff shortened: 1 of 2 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is in `reconcile-pipeline.mjs` (pipeline core), the maintainer's call; I am reviewing the owned test `tests/reconcile-pipeline-scores.test.mjs`.

It is sound. The test spawns the real `reconcile-pipeline.mjs` CLI against synthetic pipeline.md, batch-state.tsv, and report fixtures, then asserts on the score written into the real processed pipeline row, not a hand-rolled reconciler. It covers the preserve-unknown-score behavior directly: empty Score before the URL/PDF line (the exact "borrow the next line" bug), missing Score, out-of-range values and denominators, and an out-of-range batch score falling back to the report, each asserted to survive as the N/A sentinel rather than being overwritten. Crucially it does not conflate the sentinel with a real zero: a genuine 0/5 is preserved as 0, while the dash/empty/N/A forms map to N/A. I mutation-checked it: reverting the field regex to the greedy form that borrows the next line reddens one assertion, and reverting `resolveScore` to the naive grab-first-number body reddens fifteen. Forty-five assertions pass via the real harness.

Owned test sound, not behind. SIGNAL (pipeline core), so not surfacing as ready-to-approve; routing and merge are the maintainer's.

PR #3973, opened 2026-09-07
fix(doctor): report malformed setup YAML

nikitacometa:
Doctor marks malformed `config/profile.yml` and `portals.yml` as found. A broken profile also produces a successful default-MCP extractor check.

Validate both YAML prerequisites and report the filename and line/column without exposing file contents. Human output fails; `--json` adds diagnostics to `warnings`, preserving its exit status and missing-file/onboarding fields. Skip the extractor check or `--strict` portal probe when its configuration is invalid. No YAML files are rewritten.

Real CLI tests cover both outputs, target selection, empty/missing files, and diagnostic privacy. Four regression mutants fail.

Validation: `CAREER_OPS_GIT_TIMEOUT_MS=2000 CAREER_OPS_GIT_FETCH_TIMEOUT_MS=5000 node test-all.mjs` (8689 passed, 0 failed, 16 warnings); `npm run lint`.

### DIFF
--- a/doctor.mjs
+++ b/doctor.mjs
@@ -466,11 +466,33 @@ function prereqPresent(root, path) {
   return existsSync(join(root, ...path.split('/')));
 }
 
-function checkPrereq({ path, fix }) {
-  if (prereqPresent(projectRoot, path)) {
-    return { pass: true, label: `${path} found` };
+function checkPrereq(root, { path, fix }) {
+  if (!prereqPresent(root, path)) {
+    return { warn: true, label: `${path} not found (user setup required)`, fix };
   }
-  return { warn: true, label: `${path} not found (user setup required)`, fix };
+  if (path.endsWith('.yml')) {
+    let source;
+    try {
+      source = readFileSync(join(root, ...path.split('/')), 'utf8');
+    } catch {
+      return { pass: false, label: `${path}: YAML file could not be read` };
+    }
+    try {
+      // load() rejects empty/comment-only input in js-yaml 5. Allow an empty
+      // configuration while keeping the runtime's single-document contract.
+      if (yaml.loadAll(source).length > 1) {
+        return { pass: false, label: `${path}: invalid YAML (expected a single document)` };
+      }
+    } catch (err) {
+      // Parser messages, reasons and snippets can expose profile values (even
+      // an unknown alias/tag name). Only report fixed text and numeric positions.
+      const { line, column } = err?.mark || {};
+      const position = Number.isInteger(line) && line >= 0 && Number.isInteger(column) && column >= 0
+        ? ` (line ${line + 1}, column ${column + 1})` : '';
+      return { pass: false, label: `${path}: invalid YAML${position}` };
+    }
+  }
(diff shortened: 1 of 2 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is in `doctor.mjs` (core cold-start checker), the maintainer's call; I am reviewing the owned test `tests/doctor-invalid-yaml.test.mjs`.

It is sound. The test spawns the real doctor CLI over temp-dir fixtures with malformed YAML and parses its real JSON and human output, not a hand-rolled YAML parser, driving the exact `checkPrereq` path the fix touches. It covers JSON and human reporting per file and combined, valid-YAML-still-accepted (including empty and comment-only), preserved missing-YAML onboarding semantics, a YAML-path-is-a-directory case, secret non-exposure (a synthetic private marker asserted absent from stdout), `--strict` not leaking the live-probe exception, multi-document rejection, and Data-Root pinning across CAREER_OPS_ROOT / CAREER_OPS_DATA_DIR / --target precedence, and it asserts no user files or cwd are mutated. I mutation-checked it: stripping the `.yml` validation block to restore presence-only behavior reddens four assertions, and the failing one is specific (it deep-equals the expected warning list "config/profile.yml: invalid YAML (line 4, column 1)" against an empty list), not a bare non-crash.

One caveat for the maintainer, not a test defect: under `test-all.mjs --only` this suite hit the harness's 30-second per-suite timeout on my Windows box because it runs many real doctor spawns and took about 51 seconds; under a direct `node --test` it passes 16 of 16 with an authoritative exit 0 (it is a node:test suite, so the counter-style exit caveat does not apply, and the mutation returning exit 1 confirms it). It should land under the budget on CI hardware, but it is worth a glance if it runs borderline there.

Owned test sound, not behind. SIGNAL (core doctor.mjs), so not surfacing as ready-to-approve; routing and merge are the maintainer's.

Scott-Emberson (commented):
Re-verified after the head moved. This is a rebase with one owned-test-only change: `doctor.mjs` is byte-identical to the previously-verified head, so the substance is unchanged. The change to `tests/doctor-invalid-yaml.test.mjs` is harness hardening only, pinning `PLAYWRIGHT_BROWSERS_PATH` to a nonexistent dir inside the fixture root so the YAML checks stay independent of installed browsers and the Windows cache; no assertion was weakened, added, or removed.

I re-confirmed the owned test at the new head: `node test-all.mjs --only doctor-invalid-yaml` passes (16 tests, exit 0), driving the real doctor CLI over malformed-YAML fixtures with the specific warning-list assertion, root pinning via --target and the env cases, and the privacy and no-file-mutation guards intact. Stripping doctor.mjs's `.yml` validation block reddens the suite with the exact warning-list AssertionError. A useful side effect of the harness change: the 30-second test-all timeout I flagged before did not recur (the run finished in about 11.6 seconds), so that CI concern is addressed.

Owned test sound; branch behind main. SIGNAL (doctor core), so not surfacing as ready-to-approve; routing and merge are the maintainer's.

Issue #3636, opened 2026-09-01
bug: update-system.mjs apply() prunes user-generated CVs in templates/ as stale system files


Schlaflied:
## Bug

`update-system.mjs apply()` treats *any* git-tracked file under `templates/` as a prunable system file, with no `USER_PATHS` carve-out for the directory. Since the project's own documented convention (`modes/_custom.md`'s house-rule pattern, mirrored in `AGENTS.md`'s CV-generation flow) tells users to save every per-application generated CV to `templates/cv-{candidate}-{company-slug}.html` (PDF goes to `output/`, which *is* covered by `USER_PATHS`), a normal `apply()` run can silently delete a user's entire history of generated CVs.

## Root cause

- `SYSTEM_PATHS` includes a directory-prefix entry: `'templates/'` (matches everything under it via `pathMatchesManifest`'s `startsWith` check).
- `USER_PATHS` has no `templates/` entry at all — it covers `output/`, `jds/`, `reports/`, `data/`, etc., but not `templates/`.
- `staleSystemFiles()` deletes any locally-tracked file that (a) matches a `SYSTEM_PATHS` entry, (b) is absent from the upstream tree, and (c) does **not** match a `USER_PATHS` entry.
- A user-generated `templates/cv-jane-doe-acme-corp.html` satisfies (a) and (b), and nothing exempts it under (c) — so on the next `apply()` it gets `unlinkSync`'d and logged as `Pruned stale system file: templates/cv-jane-doe-acme-corp.html`.

## Repro

1. Follow the documented CV-generation flow (`modes/pdf.md` / `modes/_custom.md`'s "HTML → `templates/`, PDF → `output/`" convention) to produce several `templates/cv-{candidate}-{slug}.html` files, and `git add`/commit them (or just leave them tracked from an earlier commit).
2. Run `node update-system.mjs apply` on a version bump.
3. Every one of those generated CV files that isn't also present in the upstream repo (which is all of them — they're per-user, per-application artifacts) gets deleted and logged as `Pruned stale system file: templates/cv-...html`.

In my case this deleted 211 previously-generated CV files in one `apply()` run (the commit step itself then failed for an unrelated reason, so nothing was committed — I was able to `git checkout` them back from HEAD before any commit made the deletion permanent). On an install where the commit succeeds, this is silent, permanent data loss of a user's own application history.

## Suggested fix

[...]

PR #3974, opened 2026-09-07
fix(update-system): never prune a file upstream has not shipped

### LINKED ISSUE #3971: bug: update-system.mjs apply() prunes fork-local code under directory-prefix SYSTEM_PATHS entries (distinct from #3636/#3696)
## Bug

`update-system.mjs apply()`'s stale-file prune deletes **fork-local code** — files added under a directory-prefix `SYSTEM_PATHS` entry that upstream has never shipped. On my fork this has recurred at v1.12.0, v1.28.0 and v1.32.0; the v1.32.0 run pruned 11 files in one `apply()`, and unlike #3636 the deletions **were** committed by the update's own commit step.

This is the same root cause @santifer named in #3636 — *"the discriminator has to be 'shipped by us' … versus 'under `templates/`' by convention, never the directory"* — on a third surface. I'm filing it separately because the fixes in flight can't reach it:

- **#3636 / PR #3638** exempt generated per-application CV/cover artifacts under `templates/`, by output naming shape.
- **#3696 / PR #3700** exempt authored named-template variants, also under `templates/`.
- **This case** is `providers/*.mjs`, `tests/**`, `plugins-registry/*.json` — files that are *genuinely system-layer*. There is no correct `USER_PATHS` carve-out (you can't declare `providers/` user-layer), and no filename shape to key an exemption off: a fork's `providers/vcstack.mjs` is spelled exactly like a shipped provider. It affects forks and contributors rather than end users, so it's lower severity than #3636 — nobody loses a CV — but it's silent and it recurs on every single update.

## Root cause

`staleSystemFiles()` (update-system.mjs) selects local files that are:

1. absent from the upstream tree (`!remote.has(file)`),
2. matched by a `SYSTEM_PATHS` entry, and
3. not matched by a `USER_PATHS` entry.

`pathMatchesManifest()` matches a directory entry by `startsWith`, so every file under the ~50 directory-prefix entries (`providers/`, `tests/`, `templates/`, `docs/`, `modes/*/`, `dashboard/`, …) is in scope.

Step 1 is the faulty inference. "Absent from upstream's *current* tree" is read as "upstream removed it", but for a fork-local addition the file was **never** upstream, so its absence means nothing. The two cases are indistinguishable from the current tree alone.

## Repro

1. Fork the repo. Add `providers/example.mjs` (any file under a directory-prefix `SYSTEM_PATHS` entry — `tests/foo.test.mjs` works too). Commit it.
2. Wait for an upstream version bump, then run `node update-system.mjs apply --confirm`.
3. `Pruned stale system file: providers/example.mjs`, and it is deleted. If the update's commit step succeeds, the deletion is committed.

## Expected

A file upstream has never shipped is left alone.

## Proposed fix

Discriminate on the **previous** upstream tree, which `apply()` can already reach — it fetches upstream and knows both the installed version and the target:

| in old upstream tree | in new upstream tree | verdict |
|---|---|---|
| yes | no | genuinely retired upstream → **prune** |
| no | no | never shipped → fork-local → **keep** |

This needs no new state file and no filename heuristics, and it closes the gap in **#2532** (retired files persisting forever) at the same time — the two issues are the same missing signal read in opposite directions, which is presumably why fixing either one alone feels like it breaks the other.

A useful check on the rule: in my v1.32.0 run, 10 of the 11 prunes were wrong, but one was **right** — `lib/context-budget.test.mjs`, which upstream had genuinely migrated to `tests/context-budget.test.mjs`. The table above gets that one right too (present in the old tree, absent in the new → prune), so the fix doesn't just blanket-disable the feature.

If the old tree isn't cheaply reachable at that point in `apply()`, a recorded manifest of what the last update actually wrote would serve the same purpose, and would additionally survive a shallow clone.

Happy to open a PR if you'd like it — say the word and I'll take it.

## Environment

- CLI: Claude Code · Node 24.19.0 · Windows 11
- Observed at v1.12.0 → v1.28.0 → v1.32.0 on a public fork

eliador90:
## What does this PR do?

Stops `update-system.mjs apply()`'s stale-file prune from deleting files upstream has **never shipped** — fork-local providers, tests and registry entries under the directory-prefix `SYSTEM_PATHS` entries. It discriminates on upstream's history (already fetched by `apply()`) instead of on the current tree alone, so genuinely retired files still prune.

## Related issue

Closes #3971

## Type of change

- [x] Bug fix
- [ ] New feature

## The bug

`staleSystemFiles()` selects local files that are absent from upstream's **current** tree, matched by a `SYSTEM_PATHS` entry, and not matched by a `USER_PATHS` entry. `pathMatchesManifest()` matches directory entries by `startsWith`, so everything under the ~50 directory-prefix entries (`providers/`, `tests/`, `templates/`, `docs/`, `modes/<lang>/`, `dashboard/`, ...) is in scope.

"Absent from the current tree" reads as "upstream removed it", but for a file upstream never carried the absence means nothing. The two are indistinguishable from the current tree alone, and the prune deleted both — silently, and committed by the update's own commit step. On my fork this recurred at v1.12.0, v1.28.0 and v1.32.0; the v1.32.0 run pruned 11 files.

This is the root cause @santifer described in #3636 — *"the discriminator has to be 'shipped by us' … never the directory"* — on a third surface. It can't be fixed the way #3638 and #3700 fix theirs: there's no correct `USER_PATHS` carve-out (the file genuinely **is** system-layer), and no filename shape to key on, since a fork's `providers/acme.mjs` is spelled exactly like a shipped provider.

## The fix

`wasEverShippedUpstream()` asks `git rev-list --max-count=1 <ref> -- <path>`. A path reachable from no commit in the fetched ref was never shipped, so its absence proves nothing → keep. A path in history but gone now is a real removal → prune.

Three properties worth checking in review:

- **#2532 does not reopen.** Retired files still prune. That issue and this one are the same missing signal read in opposite directions.
- **Moves still prune.** A file upstream relocated keeps its old path in history, so the stale copy is still removed. This is real, not hypothetical — v1.32.0 moved `lib/context-budget.test.mjs` to `tests/context-budget.test.mjs`, and I verified this fix still prunes the old path.
- **Fails safe.** Pruning requires positive proof the file was shipped upstream. A shallow clone returns empty and a broken ref throws; both mean "not proven", so the file is kept. Keeping a retired file is cosmetic; deleting a fork's source file isn't recoverable from the update itself. (The first push had this fallback inverted — caught in review, fixed in cd29f128, and the test that missed it now asserts through the call site's actual decision.)

Cost is one `rev-list` per prune candidate, and candidates are typically 0–15.

## Alternative considered

Version tags would give a cleaner "previous shipped tree", but tagging stopped at `v1.6.0` upstream, so no recent version is reachable that way. A recorded manifest of what each `apply()` wrote would also work and would survive shallow clones, at the cost of new on-disk state and a one-cycle bootstrap where nothing prunes. Happy to switch to that if you'd prefer it.

## Verification

- New suite `tests/updater-never-shipped-prune.test.mjs` (5 cases): never-shipped kept, retired still prunes, moved still prunes at the old path, errored rev walk keeps, empty path rejected.
- `node test-all.mjs` on this branch: **8680 passed, 0 failed**.
- Also exercised against real history rather than only the injected fake: `providers/vcstack.mjs` → false, `lib/context-budget.test.mjs` → true, `modes/_shared.md` → true.

## Checklist

## Summary

`update-system.mjs:1027` adds `wasEverShippedUpstream()`. `apply()` uses it at `update-system.mjs:2336` to keep files that upstream never shipped.

Previously shipped files still prune, including files moved to new upstream paths. Failed or shallow history checks keep the file. `--literal-pathspecs` prevents false matches for filenames with Git pathspec characters at `update-system.mjs:1035`.

`tests/updater-never-shipped-prune.test.mjs:51` adds regression coverage for these decisions.

From the user's point of view: updates no longer delete fork-local files under `providers/`, `tests/`, or `plugins-registry/`. Retired upstream files continue to prune.

Named system paths: `update-system.mjs` is touched. `AGENTS.md`, `modes/`, `DATA_CONTRACT.md`, `providers/`, and `.github/` are not changed.

### DIFF
--- a/tests/updater-never-shipped-prune.test.mjs
+++ b/tests/updater-never-shipped-prune.test.mjs
@@ -28,6 +28,13 @@ const historyContaining = (...paths) => (...args) => {
   return paths.includes(file) ? 'a1b2c3d4e5f6a1b2c3d4e5f6a1b2c3d4e5f6a1b2' : '';
 };
 
+// The decision `apply()` actually makes, mirroring its call site:
+//   if (!wasEverShippedUpstream(f, ref)) { keep; continue; }
+// Tests assert through this rather than the bare return value — a helper named
+// for the SHIPPED state is easy to read with the wrong polarity at the point
+// where it decides a DELETE.
+const wouldPrune = (file, revList) => wasEverShippedUpstream(file, 'FETCH_HEAD', revList);
+
 // ── 1. a file upstream never shipped is kept ────────────────────────────────
 {
   // The real incident: a fork's own provider, spelled exactly like a shipped
@@ -66,14 +73,29 @@ const historyContaining = (...paths) => (...args) => {
 
 // ── 4. an unusable history keeps the file (fail safe) ───────────────────────
 {
-  // Shallow clone, or a rev walk that errors. History cannot prove the file was
-  // never shipped, and deleting a fork's source is not recoverable from the
-  // update itself, while keeping a retired file is cosmetic.
+  // A rev walk that errors (broken ref, git unavailable) proves nothing. The
+  // caller prunes only on a TRUE return, so the safe answer is false: pruning
+  // requires positive proof the file was shipped. Asserted through
+  // `wouldPrune` rather than the raw return value, because reading the
+  // polarity backwards here is exactly how the inverted fallback got written
+  // and then confirmed by its own test.
   const revListThrows = () => { throw new Error('fatal: bad object FETCH_HEAD'); };
-  if (wasEverShippedUpstream('providers/acme.mjs', 'FETCH_HEAD', revListThrows)) {
-    pass('an errored rev walk keeps the file rather than pruning on no evidence');
+  if (wouldPrune('providers/acme.mjs', revListThrows)) {
+    fail('an errored rev walk pruned the file — a broken ref would delete fork-local code');
   } else {
-    fail('an errored rev walk pruned the file — a shallow clone would delete fork-local code');
+    pass('an errored rev walk keeps the file rather than pruning on no evidence');
+  }
+}
+
+// ── 4b. a shallow clone keeps the file too ──────────────────────────────────
+{
+  // Truncated history does not throw; the walk simply finds nothing. Same
+  // verdict as the error case, reached by a different route.
+  const revListEmpty = () => '';
+  if (wouldPrune('modes/retired-mode.md', revListEmpty)) {
+    fail('a shallow clone pruned a file whose history it cannot see');
+  } else {
+    pass('a shallow clone keeps the file rather than pruning on truncated history');
   }
 }
 
--- a/tests/updater-never-shipped-prune.test.mjs
+++ b/tests/updater-never-shipped-prune.test.mjs
@@ -28,6 +28,21 @@ const historyContaining = (...paths) => (...args) => {
   return paths.includes(file) ? 'a1b2c3d4e5f6a1b2c3d4e5f6a1b2c3d4e5f6a1b2' : '';
 };
 
+// A fake that models git's PATHSPEC behaviour: without --literal-pathspecs a
+// candidate containing glob metacharacters matches a different upstream path.
+// '*' is modelled as "matches any run of characters", which is enough to show
+// the difference the flag makes.
+const globbingHistory = (...paths) => (...args) => {
+  const file = args[args.length - 1];
+  const literal = args.includes('--literal-pathspecs');
+  const star = file.indexOf('*');
+  const matches = (p) => {
+    if (literal || star === -1) return p === file;
+    return p.startsWith(file.slice(0, star)) && p.endsWith(file.slice(star + 1));
+  };
+  return paths.some(matches) ? 'a1b2c3d4e5f6a1b2c3d4e5f6a1b2c3d4e5f6a1b2' : '';
+};
+
 // The decision `apply()` actually makes, mirroring its call site:
 //   if (!wasEverShippedUpstream(f, ref)) { keep; continue; }
 // Tests assert through this rather than the bare return value — a helper named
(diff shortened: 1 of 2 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is in `update-system.mjs` (the updater's prune decision), a data-loss red-line and the maintainer's call; I am reviewing the owned test `tests/updater-never-shipped-prune.test.mjs`, and since this is the data-loss surface I verified strictly.

It is sound and uses the right design, the exported-seam pattern rather than the #3955 reconstruction. The test imports the real exported `wasEverShippedUpstream` and drives it through its injected git seam, and apply() calls that same function to gate the prune two lines below, so the decision under test is the one the caller runs, not a copy. It covers the actual guarantee: a fork-local file absent from upstream history is kept rather than pruned, the new tests extend this to the glob case (a `modes/_share*.md` fork file must not match upstream's `modes/_shared.md` as a pathspec pattern and get deleted), and retired-upstream and moved-file cases still prune while a shallow or errored history fails safe. Root pinning is clean (pure dependency injection, no temp repo, no import-time mkdir). I mutation-checked it: removing `--literal-pathspecs` from the rev-list call reddens the glob test exactly ("a fork-local filename with a glob was matched as a pattern and would be pruned").

santifer (approved):

santifer:
Merged. Thank you, @eliador90: **the updater no longer deletes a file the project never shipped**. Before pruning under a system directory it now asks upstream history: a fork's own provider or a generated CV is kept, a retired or moved file is still removed, and if git cannot answer, the file stays, which is the right way to fail. The `--literal-pathspecs` fix for names with glob characters is the kind of edge case that would have bitten someone eventually. Sorry it waited this long.

That is merge 7 from you here. If you want a next one: nothing yet runs this check through `apply()` against a real git repo, and a test that does would keep it from being dropped by accident.

PR #3975, opened 2026-09-07
docs(workday): pin the locationMainGroup nested-facet trap with a fixture

### LINKED ISSUE #3875: workday: locationMainGroup double-counts the board 2x — a latent trap for any nested-facet refactor of the facet split
Not a bug today — a trap that a plausible-looking improvement to the facet split (#3310 / #3311) would spring. Filing it so the reasoning exists before someone reaches for that refactor.

## The shape

On `dickssportinggoods|wd1|dsg`, page 0 (measured 2026-08-28), the `locationMainGroup` facet carries **2 group parents** whose **938 nested children sum to 16,732** — almost exactly **2.00x** the board's ~8,366. It double-counts because a requisition open in several locations is counted once per location.

Every other counted facet on that board agrees within 0.9% and errs *downward*:

| facet | values | sum |
|---|---|---|
| `CF_-_Job_Posting_Location_Type__LRV__Extended` | 9 | 8367 |
| `workerSubType` | 2 | 8366 |
| `jobFamily` | 24 | 8366 |
| `CF_-_Location_Type__EEB__Extended` | 4 | 8365 |
| `CF_-_Job_Requisition_Job_Category__LRV__Extended` | 26 | 8365 |
| `timeType` | 2 | 8290 |
| **`locationMainGroup` (nested)** | **938** | **16732** |

## Why it's harmless right now

Both exclusions are accidents of shape rather than deliberate checks, and neither function descends into nested values:

- `facetCoverage()` requires `Number.isInteger(v?.count)` on the facet's own `values`. The group parents carry no `count`, so `counted` stays 0 and it returns `null`.
- `chooseSplitFacet()` filters values on `typeof v?.id === 'string' && v.id`. The parents carry no `id`, so the facet drops below `values.length < 2` and is skipped.

## Why that's fragile

Recursing into nested facet values is an obvious-looking improvement — more values, a finer partition, better coverage. It would instead:

1. Make `trueTotalFromFacets()` return **16,732** instead of 8,367, since it takes the maximum.
2. Make **every healthy board** compare its honest `total` against a doubled true total and classify as offset-clamped.
3. Fire the split on boards that were never clamped, and hand `workdayTruncated` to boards that were complete.

The failure is silent and looks like success — more facets, more slices, more "recovered" lines.

## Suggested fix

An explicit guard rather than the current accident: skip a facet whose values carry nested children, or whose coverage exceeds the response's own `total` by more than the other facets' spread, and pin it with a nested-facet fixture built from the `locationMainGroup` shape above. The fixture is the part that matters — it turns the trap into a test failure instead of a production surprise.

Happy to send a PR if you want it in that shape.

bigguy6883:
Closes #3875. Landed in the shape you scoped: a comment naming `locationMainGroup` and the 2.00x, plus a fixture asserting the nested facet is never chosen. **No behavior change.**

## The trap

On `dickssportinggoods|wd1|dsg` (page 0, 2026-08-28) `locationMainGroup` carries 2 id-less group parents whose 938 nested children sum to **16,732** against a board of ~8,366 — almost exactly **2.00x**, because a requisition open in several locations is counted once per location. Every other counted facet on that board agrees within 0.9% and errs downward.

Both functions exclude it today by accident of shape rather than by an explicit check: `facetCoverage()` finds no `count` on the parents, `chooseSplitFacet()` finds no `id`. Descending into the children looks like a free improvement — more values, a finer partition — and would instead double `trueTotalFromFacets()` (it takes the maximum), make every healthy board compare its honest `total` against the doubled figure and read as offset-clamped, and hand `workdayTruncated` to boards that were complete. Silent, and it looks like success.

## What's here

- **`providers/workday.mjs`** — the reasoning recorded on `chooseSplitFacet()`'s doc block, so the `id` filter reads as load-bearing rather than tidiness. One-line pointer on `facetCoverage()`, since the doubling reaches `trueTotalFromFacets()` through it.
- **`tests/providers/workday-facet-split.test.mjs`** — `DSG_NESTED_FACETS`, the same board with `locationMainGroup` in the shape a live tenant ships: id-less parents carrying children that *do* have ids and counts. Children collapsed from 938 to 4 with the 16,732 sum preserved, since the sum is the property under test. Two assertions, each in its existing section:
  - `trueTotalFromFacets(DSG_NESTED_FACETS) === 8425` — not 16,732.
  - `chooseSplitFacet(DSG_NESTED_FACETS)` still picks `jobFamily`.

## Verification

Beyond the suite passing, I simulated the refactor — flattened `values[].values` in both functions — and confirmed the fixture actually catches it:

```
❌ trueTotalFromFacets(DSG_NESTED) returned 16732, expected 8425 — 16732 means the nested children were summed (#3875)
❌ chooseSplitFacet(DSG_NESTED) chose "locationMainGroup", expected "jobFamily" (#3875)
```

Exactly those two fail and nothing else, so the fixture adds coverage rather than restating checks that already existed. Reverted before committing.

`workday-facet-split` 34 passing / 0 failing (was 32), `workday.test.mjs` 0 failing, `npm run lint` clean.

https://claude.ai/code/session_01XNd8V4nj7XJYoAYLCVfwtg

## User impact

Workday facet totals remain correct when `locationMainGroup` contains id-less nested groups. The nested child counts are ignored, so `trueTotalFromFacets()` returns `8425` instead of `16732`. `chooseSplitFacet()` continues to select `jobFamily`.

The change adds comments that explain why facet values must have IDs and why nested values are excluded. It adds the `DSG_NESTED_FACETS` regression fixture and related tests.

## Files changed

- `providers/workday.mjs`: Documents the existing facet exclusions.
- `tests/providers/workday-facet-split.test.mjs`: Adds nested-facet regression coverage.

No behavior changes. No changes were made to `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, or `.github/`.

### DIFF
--- a/providers/workday.mjs
+++ b/providers/workday.mjs
@@ -106,7 +106,12 @@ function resolveMaxPages(entry) {
 // unfaceted crawl, and a board it could not finish keeps the workdayTruncated
 // tag rather than being reported as complete.
 
-/** Sum one facet's value counts; null when none of its values carry a count. */
+/**
+ * Sum one facet's value counts; null when none of its values carry a count.
+ *
+ * Reads a facet's own `values` only. Nested children are deliberately not
+ * summed — see the trap documented on `chooseSplitFacet()` (#3875).
+ */
 function facetCoverage(facet) {
   const values = Array.isArray(facet?.values) ? facet.values : [];
   let sum = 0;
@@ -152,6 +157,20 @@ export function trueTotalFromFacets(facets) {
  *
  * `exclude` carries the facet parameters already applied further up the split,
  * without which re-splitting a slice would keep re-deriving the same partition.
+ *
+ * Descending into those id-less headers' nested children looks like a free
+ * improvement — more values, a finer partition — and is a trap. On
+ * dickssportinggoods|wd1|dsg (measured 2026-08-28) `locationMainGroup` carries
+ * 2 group parents whose 938 nested children sum to 16,732 against a board of
+ * ~8,366: almost exactly 2.00x, because a requisition open in several locations
+ * is counted once per location. Every other counted facet on that board agrees
+ * within 0.9% and errs downward. Since `trueTotalFromFacets()` takes the
+ * maximum, recursing would double the true total, make every healthy board
+ * compare its honest `total` against it and read as offset-clamped, and hand
+ * `workdayTruncated` to boards that were complete — a silent failure that looks
+ * like success. The `id` filter below is what keeps that shut; it is
+ * load-bearing, not tidiness. See #3875; pinned by the nested-shape fixture in
+ * tests/providers/workday-facet-split.test.mjs.
  *
  * Exported for the test suite.
  */
--- a/tests/providers/workday-facet-split.test.mjs
+++ b/tests/providers/workday-facet-split.test.mjs
@@ -58,6 +58,51 @@ const DSG_FACETS = [
   },
 ];
 
+// The same board with `locationMainGroup` in the fuller shape a live tenant
+// ships (page 0, 2026-08-28): the id-less group parents above, each carrying
+// nested children that *do* have their own ids and counts.
+//
+// Those children double-count. A requisition open in several locations is
+// counted once per location, so the live facet's 938 children summed to 16,732
+// against a board of ~8,366 — almost exactly 2.00x — while every other counted
+// facet on the board agreed within 0.9% and erred downward. The children below
+// are collapsed from 938 to 4 with that sum preserved, since the sum is the
+// property under test.
+//
+// Neither function should reach them: `facetCoverage()` sums a facet's own
+// `values`, and `chooseSplitFacet()` filters on a string `id`. Recursing into
+// `values[].values` looks like a free improvement — more values, a finer
+// partition — and would instead double the true total, so every healthy board
+// would compare its honest `total` against it and read as offset-clamped. The
+// two assertions below go red if that refactor ever lands. See #3875.
+const DSG_NESTED_FACETS = [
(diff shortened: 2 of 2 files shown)


Scott-Emberson (commented):
Verified the owned test by execution and mutation. Sound and gate-relevant (both files under providers/ and tests/providers/, our owned paths).

Worth noting up front: this is effectively a docs-plus-fixture change. The `providers/workday.mjs` diff is entirely comments (an expanded JSDoc on `facetCoverage()` and an explanatory block on `chooseSplitFacet()` documenting the `locationMainGroup` nested-facet double-count trap); the real code is byte-identical, so there is no behavior change and no request path to weaken. The test adds a `DSG_NESTED_FACETS` fixture (id-less group parents with counted children) plus two regression assertions.

The test imports the real `providers/workday.mjs` and drives the actual `trueTotalFromFacets` and `chooseSplitFacet`, not a reconstructed splitter. It passes (34 assertions via the real harness). I mutation-checked both new assertions: making `facetCoverage()` recurse into nested children reddens (the total doubles to 16732 vs the expected 8425), and making `chooseSplitFacet()` flatten id-less parents into split candidates reddens (it picks locationMainGroup instead of jobFamily). Both are genuinely mutation-sensitive.

For the maintainer's awareness: there is a concurrent Workday PR (#3954, multi-location enrichment). These two are independent, not stacked and not competing: they touch different functions in disjoint regions with separate test files, so a three-way merge is clean in either order. Whichever lands second just wants a routine CI re-run.

GATE-tier (providers/), owned test sound, CI green, not behind. Holding formal approval until Scott gives the go-ahead; surfacing it as ready-to-approve.

PR #3979, opened 2026-09-07
Verify pipeline

### LINKED ISSUE #3978: verify-pipeline: a SKIP row counts as a submission channel, so correctly resolving a cross-channel collision warns forever
## Summary

`verify-pipeline.mjs`'s cross-channel duplicate check (Check "Via channels consistent") counts a `SKIP` row as a submission channel. The result is that **correctly resolving** a cross-channel collision produces a permanent warning about the very risk the user just avoided.

## Why this matters

The check exists to catch double-submission risk: the same company+role reached through two agencies. But the canonical way to resolve that collision is exactly what `states.yml` provides — apply through one channel, mark the other `SKIP` ("Doesn't fit, don't apply").

Do that, and `verify-pipeline` warns forever:

```
⚠️  Cross-channel duplicate — {company} / {role} reached via {AgencyA} AND {AgencyB}
    (rows #19, #11) — double-submission risk, resolve by hand
```

There is no "resolve by hand" action that clears it, because the row is already resolved. The user's only options are to ignore the warning permanently or to falsify data (blank the `Via`, rename the company) to silence it. Both are bad: the first trains people to ignore this check, the second destroys the audit trail. A check that cannot be satisfied by correct behaviour is worse than no check.

## Reproduction

Tracker with two rows for the same company+role, reached via different agencies, where one was applied to and the other was correctly skipped:

```
| 11 | 2026-09-02 | ExampleCorp | AgencyA | Senior Test Automation Engineer | 3.8/5 | Responded | ... |
| 19 | 2026-09-02 | ExampleCorp | AgencyB | Senior Test Automation Engineer | 3.8/5 | SKIP     | ... |
```

`node verify-pipeline.mjs` → warns about double-submission risk, though only one application was ever sent.

## Cause

`verify-pipeline.mjs` (v1.32.0, ~line 400) filters only on company:

```js
for (const e of entries) {
  const company = String(e.company || '').trim();
  if (!company || company === '?') continue;
  // ... every row becomes a channel, regardless of status
```

Status is never consulted, so a row that explicitly records "nothing was sent here" is counted as a channel through which something was sent.

## Proposed fix

Skip **only** `skip`. Deliberately narrow:

- `discarded` is ambiguous (`"Discarded by candidate or offer closed"` — can follow a real application) → stays a channel
- `evaluated` is pre-decision, and warning *before* a second submission is this check's most valuable moment → stays
- `rejected` implies an application was sent → stays

```js
const isNeverSubmitted = (status) => {
  const clean = String(status || '').replace(/\*\*/g, '').trim().toLowerCase()
    .replace(/\s+\d{4}-\d{2}-\d{2}.*$/, '').trim();
  return (ALIASES[clean] || clean) === 'skip';
};

for (const e of entries) {
  const company = String(e.company || '').trim();
  if (!company || company === '?') continue;
  if (isNeverSubmitted(e.status)) continue;
  // ...
```

Alias normalization matches Check 1's existing pattern, so localized spellings (`no_aplicar`, `uygun değil`, …) resolve correctly.

## Verification

Tested locally against v1.32.0:

- **Control** — flipping the `SKIP` row to `Applied` still raises the warning (real double-submission is caught)
- **Fixed case** — with `SKIP`, the check reports `✅ Via channels consistent`
- **No regression** — `node test-all.mjs` gives an identical `8650 passed, 27 failed` before and after the change; none of the pre-existing failures touch this check. `tests/verify-pipeline-check15.test.mjs` and `tests/verify-pipeline-control-bytes.test.mjs` pass.

Happy to open a PR with a regression test covering both directions if the approach looks right.

Abhishek84313:
## What does this PR do?

Check 11 (Via channel consistency) counted a `SKIP` row as a submission channel, so resolving a cross-channel collision the canonical way — apply through one agency, mark the other `SKIP` — produced a permanent warning about the double submission the user had just avoided, with no "resolve by hand" action left that could clear it. This skips rows whose status normalizes to `skip` before the channel map is built; `discarded`, `rejected` and `evaluated` deliberately stay channels. Adds a regression test pinning both directions.

### Implementation note

Status is normalized through the shared, states.yml-driven `normalizeStatus()` from `followup-cadence.mjs` rather than `verify-pipeline.mjs`'s local `ALIASES` table. The local table carries none of the `skip` aliases `templates/states.yml` defines (`geo_blocker`, `uygun değil`, `uygun_değil`, `uygun degil`, `uygun_degil`), so a Turkish tracker would still have hit the permanent warning; the shared path also folds the U+0307 that JS's `toLowerCase()` leaves on `İ`. This mirrors what the file already does for column layout — `resolveColumns` was pulled out of a local copy for exactly this drift (see the comment above `COLMAP`). `followup-cadence.mjs` is main-guarded with no cycle back, so the import is side-effect free.

The exemption is deliberately narrow: `discarded` is ambiguous ("Discarded by candidate or offer closed") and can follow a real application, `rejected` implies one was sent, and `evaluated` is pre-decision — warning *before* a second submission is this check's most valuable moment.

### Test evidence

`tests/verify-pipeline-via-skip.test.mjs` (auto-discovered, drives the real process):

| Assertion | pre-fix | post-fix |
|---|---|---|
| resolved collision (Applied + SKIP) → no warning | ❌ | ✅ |
| control: two Applied rows via different agencies → still warns | ✅ | ✅ |
| localized skip alias `uygun değil` → no warning | ❌ | ✅ |
| `Discarded` → still counts as a channel | ✅ | ✅ |

The two control assertions pass in both directions on purpose: a change that deleted the check outright would satisfy the first assertion alone, and the fourth pins the exemption as skip-only so it can't quietly widen later.

## Related issue

Fixes #3978

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

---

## Summary

Users can mark one duplicate application row as `SKIP` without receiving a false cross-channel warning. `verify-pipeline.mjs:413` uses shared `normalizeStatus()` and excludes only `skip`; `Applied`, `Discarded`, `Rejected`, and `Evaluated` remain counted.

Draft-answer extraction now uses the locale-invariant marker `<!-- career-ops:draft-answers -->` (`application-answers.mjs:338`). Localized headings are supported, and archived job-description content cannot redirect extraction. Report templates in `modes/` now include the marker.

`plugin-install.mjs:150` now supports `--help` and `-h`, rejects unknown flags and positional arguments, and shows usage for bare invocation.

Regression coverage includes status handling, localized aliases, draft-answer discovery, report templates, and CLI validation (`tests/verify-pipeline-via-skip.test.mjs:62`, `test-all.mjs:3086`, `tests/plugin-install-cli-flags.test.mjs:31`).

## System files touched

- `modes/`
- `AGENTS.md`: not changed
- `update-system.mjs`: not changed
- `DATA_CONTRACT.md`: not changed
- `providers/`: not changed
- `.github/`: not changed


santifer:
Hi @Abhishek84313, this PR has waited 15 days without a reply from a maintainer. That is on us, not on you, and we are sorry for it.

Where it actually stands right now: **in conflict with main, needs a rebase**.

Abhishek84313:
Thanks @santifer — sorry for the delay

Rather than just rebasing, I've split this up, because the PR was doing three unrelated things and that's my fault for stacking them on one branch:

1. verify-pipeline SKIP fix (#3978) → moved to #NEW. Branched off current main, two files, no conflicts, and I've addressed the CodeRabbit comment on its test there.
2. draft-answers marker — application-answers.mjs + the modes/ templates. This is what this PR now is. Both conflicts (modes/es/oferta.md, modes/ru/oferta.md) and three of the review comments live here.
3. plugin-install --help/flag validation — unrelated to both; happy to drop it from here and send it separately if you'd prefer.

PR #3982, opened 2026-09-07
feat(validate-profile): config/profile.yml is never validated, so a misspelled key silently does nothing (#3981)

### LINKED ISSUE #3981: config/profile.yml is never validated — a misspelled key silently does nothing
`config/profile.yml` steers scoring targets, output language, spend tier, CV format and location policy. Nothing checks its shape.

`doctor.mjs` checks the file **exists** (`:416-418`). Every reader then does `profile?.language?.output` and takes the fallback when the key is absent — which is indistinguishable from the key being **misspelled**:

```yaml
langauge:            # parses cleanly, validates nowhere
  output: ja
spend_teir: premium
```

Measured: `doctor --json` reports `warnings: []` and no mention of either key. Output comes out in English at the default tier. The user's only signal is noticing the wrong language in finished work.

`portals.yml` gets `validate-portals.mjs`. The plugin registry gets `validate-plugin-registry.mjs`. The one file that decides how every evaluation is scored and what language the output is written in gets nothing.

### A related half

`profile-language.mjs:15-22` folds three situations into `en`:

```js
try { ... return normalizeOutputLanguage(profile?.language?.output); }
catch { return DEFAULT_OUTPUT_LANGUAGE; }
```

No profile, no language key, and *a profile that does not parse* are indistinguishable to the caller. The first two are correct; the third means a YAML typo anywhere in the file silently produces English.

### Note for a fix

The known-key set should be derived from `config/profile.example.yml`, not hardcoded — a hardcoded roster can only chase the example. Three keys need care: `rejection_latency`, `table_freshness` and `scan` are read by code and are **not** in the example, so a naive derivation would warn on a correct profile.

PR to follow.

abankar1:
Fixes #3981.

`config/profile.yml` steers scoring targets, output language, spend tier, CV format and location policy. `doctor.mjs` checked that it exists; every reader then does `profile?.language?.output` and takes the fallback when the key is absent — indistinguishable from the key being **misspelled**.

```yaml
langauge:            # parses cleanly, validates nowhere
  output: ja
spend_teir: premium
```

English output at the default tier, no signal anywhere. `portals.yml` has `validate-portals.mjs` and the plugin registry has `validate-plugin-registry.mjs`; this is the same family for the one config file that had none.

## Derived, not hardcoded

The known-key set comes from `config/profile.example.yml`, so a key added there is understood on the same commit — a hardcoded roster can only chase the example and drifts.

Three keys needed care: `rejection_latency`, `table_freshness` and `scan` are read by code and are **not** in the example, so a naive derivation would warn on a *correct* profile — the failure mode that teaches users to ignore a validator. They are listed explicitly, each naming its reader, and their absence from the example is reported as its own informational finding rather than papered over. A schema you cannot learn from the file you were told to copy is a real gap.

Unknown keys carry a did-you-mean, because a typo is the whole point: *"langauge is unknown"* still leaves the user to spot the transposition.

## WARN, never FAIL

An unknown key is a typo, not a broken install. The check returns `null` when the profile is absent, so the existing prereq check keeps sole ownership of "you have not set this up yet" and the two cannot double-report. Only actionable findings reach the doctor label — the undocumented-key finding is informational and would otherwise warn users about the repo's own doc gap.

## The English fallback now says why

`parseOutputLanguage` folded three situations into `en`: no profile, no language key, and *a profile that does not parse*. The first two are correct. The third means a YAML typo anywhere in the file silently produces English in every report, cover letter and outreach message.

`describeOutputLanguage` returns `{language, source}` — `configured` / `default` / `unparseable`. `parseOutputLanguage` delegates and keeps its exact contract, so no caller changes. `configured` only when the file asked for what it got: a value that normalizes away (over 64 chars, embedded newline) is a default, not a setting, and calling it configured would be the same false confidence one level down.

## Tests

Nine, and the load-bearing one is that **the shipped `config/profile.example.yml` validates clean against itself**. If the file a user is told to copy would warn, the check is wrong rather than the user, and every other assertion is worth less.

Also pinned: a key added to the example is accepted with no edit to the validator; empty/comment-only profiles are not errors; and a clean profile adds no doctor noise.

Full suite green at 8687.

## Summary

Users can validate `config/profile.yml` before misspelled keys silently trigger defaults.

- `validate-profile.mjs`: Validates profile keys against `config/profile.example.yml` and suggests corrections for unknown keys.
- `doctor.mjs`: Reports invalid profiles as warnings without failing checks.
- `profile-language.mjs`: Distinguishes configured, default, and unparseable output languages while preserving `parseOutputLanguage`.
- `tests/profile-schema-validation.test.mjs`: Covers validation, typo suggestions, malformed and empty profiles, doctor warnings, and clean profiles.
- `update-system.mjs`: Adds `validate-profile.mjs` to the system update manifest.

From the user’s point of view: misspelled keys no longer fail silently. Malformed profiles are distinct from missing language settings. Clean profiles continue to work without warnings.

System files touched from the requested list: `update-system.mjs`. No changes were made to `AGENTS.md`, `modes/`, `DATA_CONTRACT.md`, `providers/`, or `.github/`.

### DIFF
--- a/doctor.mjs
+++ b/doctor.mjs
@@ -14,6 +14,7 @@ import * as yaml from 'js-yaml';
 import dotenv from 'dotenv';
 import { discoverPlugins, pluginRoots, pluginStatus } from './plugins/_engine.mjs';
 import { getCareerOpsRoot } from './path-resolver.mjs';
+import { validateProfile, EXAMPLE_PATH } from './validate-profile.mjs';
 import { resolveExtractorMode } from './browser-extract.mjs';
 import { parseConfigByExtension } from './jsonc-parse.mjs';
 import { validateFlags } from './lib/cli-flags.mjs';
@@ -640,6 +641,34 @@ function checkPlugins(root) {
   return fixes.length ? { warn: true, label, fix: fixes } : { pass: true, label };
 }
 
+// profile.yml steers scoring targets, output language, spend tier, CV format and
+// location policy — and the existence check above is all that ever looked at it.
+// Every reader does `profile?.language?.output` and takes the fallback when the
+// key is missing, which is indistinguishable from the key being MISSPELLED. So
+// `langauge: {output: ja}` produces English output and no signal anywhere.
+//
+// WARN, never FAIL, like the plugin check below it: an unknown key is a typo,
+// not a broken install, and refusing to run would be a worse answer than naming
+// it.
+function checkProfileShape(root) {
+  const profilePath = process.env.CAREER_OPS_PROFILE || join(root, 'config', 'profile.yml');
+  if (!existsSync(profilePath)) return null;   // the prereq check owns "absent"
+  let findings;
+  try {
+    const example = existsSync(EXAMPLE_PATH) ? readFileSync(EXAMPLE_PATH, 'utf-8') : '';
+    findings = validateProfile(readFileSync(profilePath, 'utf-8'), example).findings;
+  } catch (err) {
+    return { warn: true, label: `config/profile.yml could not be read (${err.message})` };
+  }
+  const actionable = findings.filter((f) => f.level !== 'info');
+  if (actionable.length === 0) return { pass: true, label: 'config/profile.yml: shape OK' };
+  return {
+    warn: true,
+    label: `config/profile.yml: ${actionable.length} issue${actionable.length === 1 ? '' : 's'} — settings under an unrecognized key have no effect`,
+    fix: actionable.map((f) => f.message),
+  };
+}
+
 async function main() {
   console.log('\ncareer-ops doctor');
   console.log('================\n');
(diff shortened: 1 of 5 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is `validate-profile.mjs` plus a doctor.mjs integration (core), the maintainer's call; I am reviewing the owned test `tests/profile-schema-validation.test.mjs`.

It is sound and pins both directions. The test imports the real `validateProfile` and its helpers, and two cases spawn the real `doctor.mjs --target` which itself calls `validateProfile`, so it exercises the wired-in integration, not just the library. The known-key roster is derived from the shipped `config/profile.example.yml`, and the example is asserted to validate clean against itself. Invalid input is flagged with the specific error (a misspelled key yields an unknown-key finding with the correct suggestion; a tab-broken file yields unparseable; a scalar yields not-a-mapping; doctor surfaces the "did you mean" hint), and valid input is not flagged (clean example, code-read-but-undocumented keys never flagged, empty and comment-only produce no findings). I mutation-checked both: neutering the validator to accept everything reddens the typo cases, and rejecting everything reddens the clean cases. Root pinning is clean (temp dirs, explicit CAREER_OPS_ROOT, cleared CAREER_OPS_DATA_DIR).

Owned test sound, not behind. SIGNAL (core), so not surfacing as ready-to-approve; routing and merge are the maintainer's.

PR #3985, opened 2026-09-07
fix(story-bank): three readers, three roots — and a safety gate split across two (#3983)

### LINKED ISSUE #3983: negotiation-roi's safety gate reads story-bank.md and cv.md from different roots
`interview-prep/story-bank.md` is named in the Source-of-Truth Boundary as a derived-trust input for generated content. It is resolved three different ways:

```js
match-star.mjs:22               'interview-prep/story-bank.md'            -> cwd
story-provenance-check.mjs:162  'interview-prep/story-bank.md'            -> cwd
negotiation-roi.mjs:70          join(CAREER_OPS, 'interview-prep', ...)   -> CODE root
```

### The sharp one

`negotiation-roi.mjs` already defines **both** roots, and its v1 safety gate compares the two files — a quantified claim survives only if its number also appears verbatim in `cv.md`:

```js
const CAREER_OPS      = dirname(fileURLToPath(import.meta.url));
const DATA_ROOT       = getCareerOpsRoot();
const STORY_BANK_PATH = join(CAREER_OPS, 'interview-prep', 'story-bank.md');   // CODE
const CV_PATH         = join(DATA_ROOT, 'cv.md');                              // DATA
```

Two halves of one comparison, pointed at two different installs. For any configured data root the gate reads a story bank that is not there. Measured with a story bank and matching `cv.md` in a configured root, cwd elsewhere:

```
  Stories scanned: 0
  Quantified claims found: 0
  ⚠ No stories found in story-bank.md.
```

That reads as *"you have no quantified achievements"*, not as *"I looked in the wrong place"* — and this is the tool whose whole job is anchoring a salary ask in a verified number.

`story-provenance-check.mjs` has **both** of its inputs (`story-bank.md` and `cv.md`) relative to the cwd, so the same comparison misses both sides at once.

### Note for a fix

`--story-bank` and `--cv` are documented overrides on `story-provenance-check` and must keep working; only the defaults need anchoring.

PR to follow.

abankar1:
Fixes #3983.

`interview-prep/story-bank.md` is a Source-of-Truth Boundary file, resolved three different ways:

```js
match-star.mjs:22               'interview-prep/story-bank.md'            -> cwd
story-provenance-check.mjs:162  'interview-prep/story-bank.md'            -> cwd
negotiation-roi.mjs:70          join(CAREER_OPS, 'interview-prep', ...)   -> CODE root
```

## The one that matters

`negotiation-roi.mjs` already had **both** roots defined, and its v1 safety gate compares the two files — a claim survives only if its number also appears verbatim in `cv.md`:

```js
const STORY_BANK_PATH = join(CAREER_OPS, 'interview-prep', 'story-bank.md');   // CODE
const CV_PATH         = join(DATA_ROOT, 'cv.md');                              // DATA
```

Two halves of one comparison, pointed at two different installs. Measured with a story bank and a matching `cv.md` in a configured root:

```
before:  Stories scanned: 0    ⚠ No stories found in story-bank.md.
after:   Stories scanned: 1
```

Zero reads as *"you have no quantified achievements"* — from the tool whose entire job is anchoring a salary ask in a verified number.

`story-provenance-check.mjs` had **both** inputs relative to the cwd, so the same comparison missed both sides at once. `--story-bank` and `--cv` still override; only the defaults are anchored.

## One deliberate restraint in the tests

The negotiation-roi test asserts stories **scanned**, not a claim surviving the gate. Whether a given `**Result:**` line yields an extractable claim belongs to the claim parser — coupling a path test to it would make this redden for an unrelated reason later. What this test owns is that the file was found.

I found that boundary the useful way: my first fixture omitted `**Action:**`, which `parseStories` requires, so it reported 0 stories for a reason that had nothing to do with paths.

## Structural guard

Covers both spellings that caused this — a bare relative literal and a `join()` onto a `__dirname` constant — across all three files, so the next reader of these two files fails closed rather than joining the list. Same shape as #3511's check 6.

Six tests. Reverting any one file's resolution reddens its own check; reverting all three reddens five of six. Full suite green at 8686.

## Summary

The tools now resolve default `story-bank.md` and `cv.md` files from the configured data root.

Users can run `match-star.mjs`, `story-provenance-check.mjs`, and `negotiation-roi.mjs` from any working directory without reading the wrong installation. The `--story-bank` and `--cv` overrides remain available in `story-provenance-check.mjs`.

Tests cover data-root reads, safety-gate behavior, missing-file diagnostics, overrides, and invalid path patterns in `tests/story-bank-data-root.test.mjs`.

Changed files: `match-star.mjs`, `story-provenance-check.mjs`, `negotiation-roi.mjs`, and `tests/story-bank-data-root.test.mjs`.

The requested system files and directories are not part of this change: `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/`.

### DIFF
--- a/match-star.mjs
+++ b/match-star.mjs
@@ -16,10 +16,16 @@
 
 import { readFileSync, existsSync } from 'fs';
 import { isMainModule } from './lib/is-main-module.mjs';
+import { join } from 'path';
+import { getCareerOpsRoot } from './path-resolver.mjs';
 
 // ── Config ──────────────────────────────────────────────────────────
 
-const STORY_BANK_PATH = 'interview-prep/story-bank.md';
+// From the data root, not the cwd. interview-prep/ is USER_PATHS in
+// update-system.mjs, and a bare relative path resolves against wherever the
+// process was started — so this read the checkout's story bank (usually absent)
+// for anyone whose data lives elsewhere.
+const STORY_BANK_PATH = join(getCareerOpsRoot(), 'interview-prep', 'story-bank.md');
 
 const args       = process.argv.slice(2);
 const LIST_MODE  = args.includes('--list');
--- a/negotiation-roi.mjs
+++ b/negotiation-roi.mjs
@@ -58,7 +58,7 @@
  */
 
 import { readFileSync, existsSync } from 'fs';
-import { dirname, join } from 'path';
+import { join } from 'path';
 import { fileURLToPath } from 'url';
 import { spawnSync } from 'child_process';
 import { parseStories } from './match-star.mjs';
@@ -66,9 +66,14 @@ import { flagValue, hasFlag } from './lib/cli-flags.mjs';
 import { isMainModule } from './lib/is-main-module.mjs';
 import { getCareerOpsRoot } from './path-resolver.mjs';
 
-const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
-const STORY_BANK_PATH = join(CAREER_OPS, 'interview-prep', 'story-bank.md');
 const DATA_ROOT = getCareerOpsRoot();
+// BOTH from the data root, because the v1 safety gate compares them: a claim
+// survives only if its number also appears verbatim in cv.md. Resolving the
+// story bank from the CODE root and cv.md from the DATA root pointed the two
+// halves of one comparison at two different installs — so for any configured
+// data root the gate read a story bank that is not there, found no claims, and
+// produced an empty draft that looks like "you have no quantified stories".
+const STORY_BANK_PATH = join(DATA_ROOT, 'interview-prep', 'story-bank.md');
 const CV_PATH = join(DATA_ROOT, 'cv.md');
 
 // ── Frequency vocabulary ─────────────────────────────────────────────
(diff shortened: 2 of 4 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is three story-bank readers (`match-star.mjs`, `story-provenance-check.mjs`, `negotiation-roi.mjs`) resolving off the wrong root, source-of-truth tooling and the maintainer's call; I am reviewing the owned test `tests/story-bank-data-root.test.mjs`.

It is sound and pins the anti-#3370 shape rigorously. The test spawns the real CLIs and asserts on their output, not a reconstructed resolver, and adds a structural grep guarding the next reader. Root pinning is distinct-per-root: CAREER_OPS_ROOT points at a temp dataRoot, CAREER_OPS_DATA_DIR is cleared, cwd is a separate empty decoy, and only dataRoot holds the sentinel story and cv.md while the code root has neither, so the sentinels are reachable only by resolving to the configured dataRoot rather than cwd or the code root. The not-found case asserts the message names the resolved dataRoot. I mutation-checked it: reverting all three readers to their pre-fix resolution reddens 5 of 6 (only the explicit-override case stays green, correctly).

One note for the maintainer: `story-provenance-check.mjs`'s root resolution is also touched by #3972 (already reviewed), so if both are in flight they will want a glance for overlap on that file. Both are SIGNAL, so the routing is yours either way.

Owned test sound, not behind. SIGNAL (core, source-of-truth), so not surfacing as ready-to-approve; routing and merge are the maintainer's.

PR #3988, opened 2026-09-07
fix(set-status): the canonical tracker write path cannot find a configured tracker (#3987)

### LINKED ISSUE #3987: set-status.mjs — the canonical tracker write path — cannot find a configured tracker
#3715 fixed this class for the analysis scripts, which only read. Three tracker **writers** still have it, and one of them is the write path the whole system funnels through.

```js
set-status.mjs:103    const CAREER_OPS = dirname(fileURLToPath(import.meta.url));   // the CODE root
set-status.mjs:263    const APPS_FILE  = resolveTrackerPath(CAREER_OPS);
```

`getCareerOpsRoot()` — the only thing that honours `CAREER_OPS_ROOT`, `CAREER_OPS_DATA_DIR` and the `.career-ops-data` marker — is never called.

### Measured

Tracker at a configured data root, cwd elsewhere:

```
$ CAREER_OPS_ROOT=<root> node set-status.mjs 1 Interview --json
{"error":"No tracker found at <CHECKOUT>/applications.md","code":"no-tracker"}
```

It names a file the user never configured, while their real tracker sits untouched.

### Why this one is the sharp case

- `AGENTS.md` calls it *"the canonical (locked, validated, atomic) write path"*
- the Pipeline Integrity rules say status changes go through it and **not** through hand edits
- #2901 converged the web layer's `/api/status` onto it

So on any configured data root, the single supported way to change an application's status does not work — and `outcome.mjs`, which shells out to it, fails the same way.

### The full census

Every `resolveTrackerPath()` caller in the repo, by what it is handed:

```
OK  (data root)   16 scripts
BUG (code root)   set-status.mjs, mark-pdf-ready.mjs, outcome.mjs
```

### Note for a fix

`set-status.mjs` and `outcome.mjs` need **two** roots, not a substitution. `set-status` reads `templates/states.yml`, which ships with the code and is absent from a user's data root — moving everything to the data root would break every status validation. `outcome.mjs` shells out to sibling scripts and uses its own directory as their cwd.

PR to follow.

abankar1:
Fixes #3987. Completes #3715, which covered the readers.

```js
const CAREER_OPS = dirname(fileURLToPath(import.meta.url));   // the CODE root
const APPS_FILE  = resolveTrackerPath(CAREER_OPS);
```

Measured, tracker at a configured data root and cwd elsewhere:

```
$ CAREER_OPS_ROOT=<root> node set-status.mjs 1 Interview --json
{"error":"No tracker found at <CHECKOUT>/applications.md","code":"no-tracker"}
```

It names a file the user never configured while their real tracker sits untouched.

## Why this one is worse than the readers

`AGENTS.md` calls `set-status.mjs` *"the canonical (locked, validated, atomic) write path"*; the Pipeline Integrity rules say status changes go through it and **not** through hand edits; and #2901 converged the web layer's `/api/status` onto it. On any configured data root the single supported way to change a status did not work — and `outcome.mjs`, which shells out to it, failed the same way.

A wrong path in a reader produces an empty report. In a writer it reports success against a file the user never configured.

## The census

Every `resolveTrackerPath()` caller, by what it is handed:

```
OK  (data root)   16 scripts
BUG (code root)   set-status.mjs · mark-pdf-ready.mjs · outcome.mjs
```

All three fixed here. `set-status` and `outcome` needed **two** roots rather than a substitution — `set-status` reads `templates/states.yml`, which ships with the code and is absent from a user's data root, and `outcome` uses its own directory as the cwd for the sibling scripts it spawns. A blanket move would have broken status validation, which is the thing `set-status` exists to do.

## Tests

Seven, each separating the data root from the cwd — from the data root the two resolutions agree and the bug is invisible.

Three guards beyond the fixes:

- `data/status-log.tsv` must land **beside** the tracker it describes, or the record is split in two.
- `templates/states.yml` must keep resolving from the code root — so the easier blanket fix reddens. It also asserts a bogus state is still **rejected**, which `set-status` can only do by having read the template; without that, "no error about states.yml" would pass on a script that silently stopped validating.
- A structural sweep over every `resolveTrackerPath()` caller, so the next writer fails closed rather than joining a list of three.

That structural check earned its place immediately — it caught `outcome.mjs`, which I had fixed on a different branch and not on this one.

Reverting `set-status` alone reddens 4 of 7; either of the other two reddens 2. Full suite green at 8686.

## Summary

Tracker writers now resolve tracker files from the configured data root.

### User impact

When `CAREER_OPS_ROOT` or `CAREER_OPS_DATA_DIR` is configured, users can run status, outcome, and PDF-ready updates from another working directory. The scripts update the configured tracker and place the status log beside it.

Status validation still loads `templates/states.yml` from the code root. Sibling scripts also run from the code root. The change prevents writes to the launch directory or the writer module directory.

### Files changed

- `set-status.mjs`: Uses `getCareerOpsRoot()` for tracker data and retains code-root status validation.
- `outcome.mjs`: Uses separate code and data roots.
- `mark-pdf-ready.mjs`: Uses the configured data root.
- `tests/tracker-writers-data-root.test.mjs`: Adds regression coverage for separated roots, working directories, logs, validation, and tracker-path callers.

The requested system files were not part of this change: `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/`.

### DIFF
--- a/mark-pdf-ready.mjs
+++ b/mark-pdf-ready.mjs
@@ -41,8 +41,11 @@ import { extractTrackerReportNumbers, resolveColumns, parseTrackerRow } from './
 import {
   rebuildRow, resolveTrackerPath, writeFileAtomic, CLI_EXIT, makeCliFailWith, acquireTrackerLockForCli,
 } from './tracker-utils.mjs';
+import { getCareerOpsRoot } from './path-resolver.mjs';
 
-const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
+// The USER's data root, not this file's directory — same defect, same fix as
+// set-status.mjs above it in this change.
+const DATA_ROOT = getCareerOpsRoot();
 
 // LOCK_TIMEOUT is not destructured here — that exit path is raised inside
 // acquireTrackerLockForCli() itself (tracker-utils.mjs), via CLI_EXIT.LOCK_TIMEOUT.
@@ -102,7 +105,7 @@ const targetReportNum = parseInt(reportSelector, 10);
 
 // ── tracker access ───────────────────────────────────────────────
 
-const APPS_FILE = resolveTrackerPath(CAREER_OPS);
+const APPS_FILE = resolveTrackerPath(DATA_ROOT);
 if (!existsSync(APPS_FILE)) {
   failWith(EXIT_NOT_FOUND, 'no-tracker', `No tracker found at ${APPS_FILE}`);
 }
--- a/outcome.mjs
+++ b/outcome.mjs
@@ -36,14 +36,22 @@ import {
   resolveTrackerPath,
   resolveWorkspaceRoot,
 } from './tracker-utils.mjs';
+import { getCareerOpsRoot } from './path-resolver.mjs';
 import { resolveOutcomeDir } from './lib/outcome-dir.mjs';
 import { parsePdfIndex } from './find.mjs';
 import { findCaptureForReport } from './jd-capture.mjs';
 
-const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
+// Two roots. CODE_ROOT locates the sibling scripts this file shells out to and
+// is the cwd it runs them from; DATA_ROOT is where the user's tracker and
+// outcome journals live. One constant named CAREER_OPS did both, so
+// resolveTrackerPath() looked inside the checkout and `outcome.mjs 1 rejected`
+// answered "Tracker not found at <CHECKOUT>/applications.md" — a path the user
+// never configured.
+const CODE_ROOT = dirname(fileURLToPath(import.meta.url));
+const DATA_ROOT = getCareerOpsRoot();
 const NODE = process.execPath;
-const SET_STATUS_SCRIPT = join(CAREER_OPS, 'set-status.mjs');
-const ARCHIVE_POSTING_SCRIPT = join(CAREER_OPS, 'archive-posting.mjs');
+const SET_STATUS_SCRIPT = join(CODE_ROOT, 'set-status.mjs');
+const ARCHIVE_POSTING_SCRIPT = join(CODE_ROOT, 'archive-posting.mjs');
 
 const EXIT_OK = 0;
 const EXIT_USAGE = 1;
@@ -142,7 +150,7 @@ if (!outcomeConfig) {
   failExit(`Invalid outcome_type "${rawOutcomeType}". Valid types: ${validTypes}`, 'invalid-outcome', EXIT_USAGE);
 }
 
-const appsFile = resolveTrackerPath(CAREER_OPS);
+const appsFile = resolveTrackerPath(DATA_ROOT);
 if (!existsSync(appsFile)) {
   failExit(`Tracker not found at ${appsFile}`, 'tracker-not-found', EXIT_NOT_FOUND);
 }
@@ -422,7 +430,7 @@ if (!resolvedPostingPath) {
 if (!resolvedPostingPath && targetUrl) {
   try {
     execFileSync(NODE, [ARCHIVE_POSTING_SCRIPT, targetUrl, `--company=${matchedRow.company}`, `--role=${matchedRow.role}`, `--report=${matchedRow.num}`], {
-      cwd: CAREER_OPS,
+      cwd: CODE_ROOT,
       env: process.env,
       stdio: 'ignore',
       timeout: 45000,
@@ -503,7 +511,7 @@ if (matchedRow.role) {
 
 let setStatusResult = null;
 try {
-  const statusOutput = execFileSync(NODE, setStatusArgs, { cwd: CAREER_OPS, env: process.env, encoding: 'utf-8' });
+  const statusOutput = execFileSync(NODE, setStatusArgs, { cwd: CODE_ROOT, env: process.env, encoding: 'utf-8' });
   setStatusResult = JSON.parse(statusOutput);
 } catch (err) {
   failExit(`Tracker update via set-status.mjs failed: ${err.message}`, 'tracker-update-failed', 1);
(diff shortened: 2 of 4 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is `set-status.mjs` (the canonical tracker write path) plus outcome.mjs and mark-pdf-ready.mjs, core and the maintainer's call; I am reviewing the owned test `tests/tracker-writers-data-root.test.mjs`. Since this fixes a wrong-root write (a data-loss risk, #3987) I verified it strictly.

It is sound. The test spawns the real CLIs, and for outcome.mjs it exercises the real parent-locates-then-child-writes shell-out chain, not a reconstructed resolver. Root pinning is distinct-per-root: each case holds the real tracker in a temp dataRoot and spawns with cwd set to a separate empty decoy but CAREER_OPS_ROOT set to dataRoot, so cwd and the configured root deliberately disagree (the exact condition where the bug appears). It then asserts positively (the dataRoot tracker and its sibling status-log gained the transition) and negatively (the decoy cwd stays empty, no "No tracker found" in output), plus a structural scan for the offending pattern and a guard that states.yml still resolves from the code root. I mutation-checked it: reverting all three writers to resolve the tracker from the code root reddens 5 of 7 (the two survivors are the code-root guards, correctly unaffected).

santifer (changes requested):
Thanks for this, @abankar1: the split into two roots is right, and the reason each script needs both is the careful part. `set-status` keeping the code root for `templates/states.yml`, and `outcome` keeping it as the cwd for the siblings it spawns, is exactly what a blanket substitution would have broken.

I reverted the three writers to check rather than read: 5 of your 7 tests go red and the two guards stay green, which is the right shape. The reported failure reproduces exactly, naming the checkout path before and the configured root after.

I am holding the merge on one line, and it is in the test rather than in the fix. `run()` spawns the real writer CLIs with the whole environment inherited and clears only `CAREER_OPS_DATA_DIR`. `CAREER_OPS_TRACKER` outranks the resolved root in `path-resolver.mjs`, so a developer who has it exported does not just get a red suite. I put a tracker outside the fixture whose row 1 matches yours, ran the file, and it rewrote that tracker from `Applied` to `Rejected` and appended a note. In a change about writers finding the wrong root, the test can write to the wrong root.

Adding `CAREER_OPS_TRACKER: ''` to the cleared keys gave 7 of 7 green with the external file untouched. CodeRabbit flagged the same line; I am repeating it because I measured the write rather than the risk, and because CI never exports that variable, so CI will stay green either way.

Two smaller things, neither blocking. `mark-pdf-ready.mjs` still imports `dirname` and `fileURLToPath` with nothing using them. And the suite covers the `CAREER_OPS_ROOT` leg only: @Zoubeir23 offered a `CAREER_OPS_DATA_DIR` regression test when he closed #4032 in favour of this one, and the answer to that offer is yes, as a follow-up once this lands.

Push that line and I will take it. #3990 is stacked on this and cannot run its own test until it goes in.

abankar1:
You're right, and the irony is not lost on me: a change about writers finding the wrong root shipped a test that writes to the wrong root. Fixed and pushed.

I reproduced your scenario before fixing it rather than taking it on faith — a tracker outside the fixture, `CAREER_OPS_TRACKER` exported, then the suite:

```
external row before:  | 1 | ... | Applied  | ... | untouched |
external row after:   | 1 | ... | Rejected | ... | untouched; Application rejected |
```

Exactly what you measured. With `CAREER_OPS_TRACKER: ''` added to the cleared keys: 7 of 7 green and that file byte-identical. Reverting just that one line brings the write straight back (4 pass / 3 fail, external tracker modified), so the guard is load-bearing rather than decorative.

What I had wrong was the reasoning, not just the line. I cleared `CAREER_OPS_DATA_DIR` because it is the sibling of the variable I was setting, and never asked which variables outrank the one I was setting. `path-resolver.mjs` answers that plainly — `CAREER_OPS_TRACKER` sits above the resolved root — and I had read that function closely enough in this same PR to have known.

`mark-pdf-ready.mjs` — `dirname` and `fileURLToPath` both dropped; neither had a call site left after the split.

On the `CAREER_OPS_DATA_DIR` leg: yes, and thank you @Zoubeir23 for offering it. I'll open it as a follow-up once this lands rather than widen this PR.

One thing worth passing on, since it is the same class as your finding. Two suites already on main have the same gap: `tests/analysis-scripts-data-root.test.mjs` (#3715) and `tests/rejection-latency-placeholder-rows.test.mjs` (#3713) both clear only `CAREER_OPS_DATA_DIR`. Both spawn READERS, so the consequence is a confusing local failure rather than a write — materially less bad, which is presumably why nobody hit it. I'd rather name it than leave it for someone to rediscover; happy to fold both into the `CAREER_OPS_DATA_DIR` follow-up.

santifer:
Thanks, @abankar1: that's the line, and the reasoning you put around it is the part that matters. `CAREER_OPS_TRACKER` outranking the resolved root is what makes that guard load-bearing rather than decorative.

I read the pushed diff rather than the bot's marker: `run()` now clears it alongside `CAREER_OPS_DATA_DIR`, and `mark-pdf-ready.mjs` is down to `getCareerOpsRoot()` with both dead imports gone. CI is green on this head across ubuntu, macOS and windows. That closes the one thing I was holding this on, so it's a favourable review from my side.

One leftover, optional and not a reason to hold anything: CodeRabbit's second note still applies. The structural check uses `src.match()` without the global flag, so it only inspects the first `resolveTrackerPath()` call in each file. Fine to leave as is, or to fold into the `CAREER_OPS_DATA_DIR` follow-up along with the two suites you flagged on main: naming those was the right call.

The branch is behind main. There are no conflicts, I merged it against current main locally to check, so the remaining integration step is ours rather than yours. Nothing else for you to push.

PR #3990, opened 2026-09-07
fix(outcome,tracker): stamp the local calendar day, not the UTC one (#3989)

### LINKED ISSUE #3989: outcome.mjs and tracker.mjs stamp the UTC day, not the local one
`tests/local-today-gates.test.mjs` scoped itself to places that *"GATE a decision rather than stamping a filename"*, and included `assessment-log.mjs` for *"the date written into a user's assessments.tsv row"*. Two more write exactly that kind of date and are still on `new Date().toISOString()`:

```js
outcome.mjs:62    return new Date().toISOString().split('T')[0];
tracker.mjs:552   const today = new Date().toISOString().slice(0, 10);
```

**`outcome.mjs`** stamps the `## Entry:` header and `**Date**:` in the append-only journal under `data/outcomes/`, which `calibrate.mjs` and `funnel-velocity.mjs` then read. An outcome recorded on a Sunday evening anywhere in the Americas is journaled as Monday — moving it into the next week and the next funnel bucket.

**`tracker.mjs`** dates a row in the SQLite `status_events` table. Lower severity because the index is derived, except that this is the one part a resync does **not** rebuild identically — the comment right below it says events *"persist across rebuilds, keyed by id"*, so a wrong day is written once and sticks.

### Two notes for whoever tests this

The tracker case needs **two** syncs with a status change between them. `today` is only reached on a transition; a first sync dates the event from the row's own Date column (`DATE_RE.test(a.date) ? a.date : today`), so a single-sync test never touches the clock.

And freezing the clock needs `--import`, not a wrapper module that imports the target. `isMainModule()` compares against the process entry path, so a wrapper entry makes these scripts decide they are libraries and skip their CLI bodies — a silent exit 0 that looks like the command ran and did nothing.

PR to follow.

abankar1:
Fixes #3989.

> **Stacked on #3988.** The outcome test cannot run without it — `outcome.mjs` could not find a configured tracker at all. Review #3988 first; this branch shows only its own two commits once that lands.

`tests/local-today-gates.test.mjs` scoped itself to places that *"GATE a decision rather than stamping a filename"*, and included `assessment-log.mjs` for *"the date written into a user's assessments.tsv row"*. These two write exactly that kind of date:

```js
outcome.mjs:62    return new Date().toISOString().split('T')[0];
tracker.mjs:552   const today = new Date().toISOString().slice(0, 10);
```

An outcome recorded on a Sunday evening anywhere in the Americas was journaled as Monday, moving it into the next week and the next funnel bucket — and `calibrate.mjs` and `funnel-velocity.mjs` read that journal.

The tracker one sticks. The index is derived and rebuildable, except that `status_events` is the one part a resync does **not** rebuild identically — the comment right below says events *"persist across rebuilds, keyed by id"*.

## Two things the tests had to get right

Both found the useful way: by reverting the fix and watching the test still pass.

**The tracker case needs two syncs with a status change between them.** `today` is only reached on a *transition* — a first sync dates the event from the row's own Date column (`DATE_RE.test(a.date) ? a.date : today`) — so a single-sync test never touches the clock and measures nothing.

**The clock is frozen via `--import`, not a wrapper that imports the target.** `isMainModule()` compares against the process entry path, so a wrapper entry makes these scripts decide they are libraries and skip their CLI bodies entirely: a silent exit 0 that looks like the command ran and did nothing. `--import` runs the preload first and leaves the script as the entry point — which also keeps this file clear of `process.argv[1]`, the convention `tests/main-guard-convention.test.mjs` enforces. It caught my first attempt.

The instant is pinned rather than read from the wall clock, for the reason the gate file records: the window where the UTC day and the local day disagree exists for only part of the UTC day, so a test reading `new Date()` passes for most of the day whether the bug is there or not.

Reverting either fix reddens exactly its own check. Full suite green at 8688.

## Summary

Users now get local-calendar dates for outcome entries and tracker status events. Sunday-evening entries in the Americas no longer move into Monday’s weekly or funnel calculations.

The change also supports configured data roots:

- `outcome.mjs:62`: Uses the local date for journal headers and `Date` fields.
- `tracker.mjs:552`: Uses the local date for new `status_events` records.
- `outcome.mjs`, `set-status.mjs`, `mark-pdf-ready.mjs`: Resolve tracker data from the configured root.
- `outcome.mjs`, `set-status.mjs`: Continue loading templates and scripts from the code root.
- `tests/outcome-tracker-local-dates.test.mjs`: Adds frozen-clock regression tests.
- `tests/tracker-writers-data-root.test.mjs`: Verifies configured-root resolution, status-log placement, and template loading.

### DIFF
--- a/mark-pdf-ready.mjs
+++ b/mark-pdf-ready.mjs
@@ -41,8 +41,11 @@ import { extractTrackerReportNumbers, resolveColumns, parseTrackerRow } from './
 import {
   rebuildRow, resolveTrackerPath, writeFileAtomic, CLI_EXIT, makeCliFailWith, acquireTrackerLockForCli,
 } from './tracker-utils.mjs';
+import { getCareerOpsRoot } from './path-resolver.mjs';
 
-const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
+// The USER's data root, not this file's directory — same defect, same fix as
+// set-status.mjs above it in this change.
+const DATA_ROOT = getCareerOpsRoot();
 
 // LOCK_TIMEOUT is not destructured here — that exit path is raised inside
 // acquireTrackerLockForCli() itself (tracker-utils.mjs), via CLI_EXIT.LOCK_TIMEOUT.
@@ -102,7 +105,7 @@ const targetReportNum = parseInt(reportSelector, 10);
 
 // ── tracker access ───────────────────────────────────────────────
 
-const APPS_FILE = resolveTrackerPath(CAREER_OPS);
+const APPS_FILE = resolveTrackerPath(DATA_ROOT);
 if (!existsSync(APPS_FILE)) {
   failWith(EXIT_NOT_FOUND, 'no-tracker', `No tracker found at ${APPS_FILE}`);
 }
--- a/outcome.mjs
+++ b/outcome.mjs
@@ -36,14 +36,22 @@ import {
   resolveTrackerPath,
   resolveWorkspaceRoot,
 } from './tracker-utils.mjs';
+import { getCareerOpsRoot } from './path-resolver.mjs';
 import { resolveOutcomeDir } from './lib/outcome-dir.mjs';
 import { parsePdfIndex } from './find.mjs';
 import { findCaptureForReport } from './jd-capture.mjs';
 
-const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
+// Two roots. CODE_ROOT locates the sibling scripts this file shells out to and
+// is the cwd it runs them from; DATA_ROOT is where the user's tracker and
+// outcome journals live. One constant named CAREER_OPS did both, so
+// resolveTrackerPath() looked inside the checkout and `outcome.mjs 1 rejected`
+// answered "Tracker not found at <CHECKOUT>/applications.md" — a path the user
+// never configured.
+const CODE_ROOT = dirname(fileURLToPath(import.meta.url));
+const DATA_ROOT = getCareerOpsRoot();
 const NODE = process.execPath;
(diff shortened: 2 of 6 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is outcome.mjs and tracker.mjs stamping the local calendar day instead of the UTC one, core and the maintainer's call; I am reviewing the owned test `tests/outcome-tracker-local-dates.test.mjs` (the second owned file, `tracker-writers-data-root.test.mjs`, is inherited byte-identical from #3988, see the stack note below).

It is sound. The test spawns the real outcome.mjs and tracker.mjs entry points (a preload freezes the clock while the script stays argv[1] so the real CLI body runs) and asserts on the real outcome journal and the status_events rows read back from the built database, not a reconstructed stamper. It is timezone-controlled correctly: it freezes the instant 2026-08-18T01:30Z under America/New_York, where the local day (08-17) differs from the UTC day (08-18), with a guard asserting that premise, so it catches the bug every hour of the day rather than only during a disagreement window. The sqlite branch actually ran (Node 22.23), so the status_events assertion is live, and the suite correctly requires two syncs with a transition between them. I mutation-checked it: reverting both writers to the UTC stamp reddens 2 of 3 (the premise guard stays green).

For the maintainer: this PR is a clean stack on #3988. Its first three commits are exactly #3988, and its own commit adds only the local-date fix plus this test; the shared `tracker-writers-data-root.test.mjs` is byte-identical. The clean path is to land #3988 first, then this rebases down to just the local-date commit.

Owned test sound, not behind. SIGNAL (core), so not surfacing as ready-to-approve; routing and merge are the maintainer's.

Issue #3828, opened 2026-09-04
i18n: seven market evaluation modes (de, fr, hi, id, it, ko, nl) are frozen pre-Block-G, and the oferta.md parity check cannot see them


nikitacometa:
## What is wrong today

#3669 re-syncs the six frozen `modes/<lang>/oferta.md` files, and the parity check it added (`test-all.mjs`, the `FROZEN_OFERTA` block) walks exactly `modes/*/oferta.md`. Seven market mode sets keep their evaluation mode under a market-specific name, so the same freeze is invisible to that check. Measured on main at fad070fe:

| file | lines | mentions of "legitimacy" | `## Risk Summary` | `## G)` |
|---|---|---|---|---|
| `modes/de/angebot.md` | 165 | 0 | no | draft answers |
| `modes/fr/offre.md` | 166 | 0 | no | draft answers |
| `modes/hi/naukri.md` | 170 | 0 | no | draft answers |
| `modes/id/lowongan.md` | 166 | 0 | no | draft answers |
| `modes/it/annuncio.md` | 166 | 0 | no | draft answers |
| `modes/ko/gonggo.md` | 164 | 0 | no | draft answers |
| `modes/nl/vacature.md` | 167 | 0 | no | draft answers |
| `modes/ar/fursah.md` | 247 | 5 | yes | draft answers |
| `modes/ja/kyujin.md` | 341 | 7 | yes | draft answers |
| `modes/tr/is-ilani.md` | 285 | 2 | yes | draft answers |

The seven files at the top have the pre-Block-G shape #3669 describes: no Liveness / Blacklist gates, no Block G (Posting Legitimacy), no Risk Summary, no Machine Summary, and a `## G)` that means draft answers. `AGENTS.md` requires every report to carry `**Legitimacy:** {tier}` (Block G), so a user with `language.modes_dir: modes/de` (the documented DACH path) gets evaluations with no legitimacy screen at all, while a locale with no evaluation mode falls back to the canonical and gets it. The same "having the translation is worse than not having it" effect as #3669, one directory over. `ar`, `ja` and `tr` already carry the block and are listed only for completeness.

## Proposal

1. Extend the parity check so it cannot go blind on file names: walk each market directory's evaluation mode by the name the README table declares (or every `modes/<lang>/*.md` that contains a report-format `## A)` heading), with a `FROZEN` allowlist that shrinks as files land, exactly like `FROZEN_OFERTA`. I can send that as a small PR if the shape is acceptable.
[...]

Issue #3168, opened 2026-08-21
i18n: automated per-language drift check (unblocks full-set translations)


santifer:
## The problem, measured

Full-set translations go stale before they land: main merges 14-34 PRs a day, and #1666's Turkish full set already lagged its canonical files on arrival (email.md at 15 of 27 sections, scan.md at 352 of 464 lines) through no fault of the translator. #2573 tracks what that drift costs once it is in the tree. Today's standing policy is therefore: languages ship the core subset; full sets wait for this mechanism.

## The mechanism this issue asks for

An automated i18n drift check that compares each `modes/{lang}/X.md` against `modes/X.md` STRUCTURALLY (section headings present/absent, section count, optionally line-count ratio per section) and reports per-file coverage: exactly the measurement done by hand on #1666, as a script anyone can run and CI can gate on.

- Output: a per-language, per-file coverage table (sections covered / total, with missing section names).
- Non-goal: judging translation QUALITY. Structure only — presence, not prose.
- Wiring: standalone script first (`i18n-drift.mjs`); a CI warn (not fail) once the baseline is clean.

## Why it unlocks more than it checks

The day this exists, full-set translations become acceptable again for every language: the check keeps them honest afterwards. Several translators (Turkish is the live case, see #1666) have full-set work banked waiting on exactly this.

@d-ulker has first right to this one, having built the measurement's motivating case. If unclaimed in a while, it opens to everyone.

santifer:
Assigned, @alinaqvi1129, and **your `/assign` two days ago should have done this by itself.** It didn't, and the reason is a bug on our side rather than anything you did.

The claim bot only acts on issues labelled `good first issue` or `first-timers-only` (`gfi-claims.yml:49`). This one is labelled `enhancement`, so your comment matched nothing and the workflow returned without a word. **Silence, not a refusal** — which is the worst possible answer, because from where you sit "the bot is thinking" and "the bot ignored me" look identical.

You've hit the same wall twice this week from opposite sides: #3154 worked because it carries the label, this one didn't because it doesn't. Filing the gap now so the next person gets an answer instead of nothing.

The issue itself is a good one to be holding. An automated per-language drift check is the thing that makes the whole translation set maintainable rather than a growing debt — and today I merged a fix for eleven translated READMEs that had quietly weakened the project's central guarantee, which is precisely the class of drift a check like yours would have caught months earlier. Worth knowing that the ground under your issue just got firmer.

santifer:
[...]

Issue #1449, opened 2026-07-02
Migrate from Playwright MCP over to Playwright CLI, to reduce token consumption


dempseydata:
### Code of Conduct

- [x] I agree to follow the [Code of Conduct](https://github.com/santifer/career-ops/blob/main/CODE_OF_CONDUCT.md)

### Existing issues

- [x] I searched existing issues and this hasn't been requested yet

### Problem

Playwright MCP consumes tokens at a higher rate than the CLI version.  With a $20 claude subscription and sonnet 2.6 medium effort, the limit in a 5 hour session can easily be just 10-15 roles in the pipeline, without any applications.

### Proposed solution

Migrate fro Playwright MCP to the CLI version - https://github.com/microsoft/playwright-cli

### Area

Portal Scanner

santifer:
This direction is worth pursuing — token cost is the #1 community pain and the MCP layer's protocol overhead is real. Before a PR though, this needs a short proposal here in the issue: (1) measured numbers (same scan via MCP vs CLI — even rough ones), (2) how liveness checks and `apply` mode would work through the CLI path, and (3) whether it's a swap or an opt-in alternative (opt-in is far likelier to land — plenty of users have working MCP setups we shouldn't break). Bring numbers and a migration shape and this moves fast.

abankar1:
Would like to take a run at this, @santifer — here's a proposal addressing your three asks. (Context: I authored the Teamtailor provider #1376 and the recent /add, contacto-greeting, and merge-tracker column-order PRs.)

## The actual cost driver
The token burn isn't Playwright itself — it's the **agent-driven `browser_navigate` + `browser_snapshot`** in `scan.md` (Level 1, flagged in the mode as "the most expensive token-consuming method"), plus the same pair in `pipeline.md` / `oferta.md` / `auto-pipeline.md` for JD extraction. Each `browser_snapshot` streams the full accessibility tree of a careers SPA back to the model.

Crucially, **the repo already runs Playwright headlessly in Node** with zero agent tokens: `liveness-browser.mjs` (shared by `check-liveness.mjs` and `scan.mjs --verify`, with a realistic-UA context to clear WAFs) and `generate-pdf.mjs`. So the pattern and dependency already exist — the win is routing the *read/extract* step through that same lane instead of the MCP.

## (1) Numbers (rough, will measure for real)
- A `browser_snapshot` of an Ashby/Workday careers SPA is typically **~15k–40k tokens**; scanning N tracked companies multiplies it.
[...]

Issue #749, opened 2026-05-26
Race condition in report-number allocation across parallel interactive CLI windows


Jonny-boy9000:
### Code of Conduct

- [x] I agree to follow the [Code of Conduct](https://github.com/santifer/career-ops/blob/main/CODE_OF_CONDUCT.md)

### Existing issues

- [x] I searched existing issues and this hasn't been requested yet

## Summary
  When running multiple Claude Code (or any CLI-agent) windows in parallel on the same `career-ops` checkout, two workers can independently compute the same "next report number" (max existing + 1) and produce
  colliding artifacts.

  ## Repro
  1. Open two Claude Code windows in `C:/career_ops`.
  2. In each, paste a different JD URL roughly simultaneously to trigger `auto-pipeline`.
  3. Both windows scan `reports/`, see the same max (e.g., `034`), and both pick `035`.
  4. Each writes `reports/035-{differentSlug}-{date}.md` — both files persist (different slugs, no FS overwrite) but the slot number is duplicated.
  5. Each writes `batch/tracker-additions/035-{differentSlug}.tsv`. `merge-tracker.mjs` then appends two rows numbered `035` into `data/applications.md`.

  ## Impact
  - Duplicate sequential IDs in `data/applications.md` (breaks "app #N" references in memory / interview-prep links).
  - Tracker dedup script can mis-merge if two evaluations collide on slot + date.
  - Silent failure mode — no error is raised; the user only notices when reviewing the tracker.

  ## Root cause
  The "next num = max existing + 1" rule in `CLAUDE.md` / `AGENTS.md` / `modes/oferta.md` is computed independently in each process with no cross-process lock. `batch/batch-runner.sh` already solves this for
  the batch case (`reserve_report_num` under `run_with_state_lock`, lines ~235–315), but interactive auto-pipeline / `oferta` modes don't go through that path.

## Environment
  - Windows 11 / PowerShell
  - career-ops latest version.
  - Claude Code CLI, multiple parallel windows

## Suggested fixes (any one would close it)
  1. **Atomic reservation script** (e.g., `reserve-report-num.mjs`) that uses `fs.writeFileSync(..., {flag:'wx'})` on a sentinel `reports/{NNN}-RESERVED.md` — atomic on both Windows and POSIX, no extra deps.
  Mode files call it instead of computing max+1. Worker deletes the sentinel after writing the real report; `verify-pipeline.mjs` GCs stale sentinels.
[...]

PR #3991, opened 2026-09-07
fix(i18n): route zh/zh-TW pipeline report numbering through reserve-report-num.mjs

### LINKED ISSUE #3980: i18n(zh/zh-TW): pipeline.md tells the agent to compute max+1 for the report number — the #749 race, with 3+ parallel subagents recommended in the same file
**Area:** `modes/zh/pipeline.md`, `modes/zh-TW/pipeline.md` (and `modes/ar/pipeline.md`, see Scope)

## What happens

Both Chinese `pipeline.md` modes tell the agent to derive the next report number by hand:

> a. **计算新报告编号**：扫描 `reports/` 目录，找出当前最大的三位数字前缀并加 1，作为 `REPORT_NUM`。
>
> *("Compute the new report number: scan `reports/`, find the largest three-digit prefix and add 1.")*

`AGENTS.md` names this exact algorithm as a known race:

> **Never let parallel workers compute `max+1` themselves — that is the #749 race.**

The same file then recommends fanning out:

> 3. **并发加速**：如果待处理的 URL 达到 3 个及以上，建议以并发方式（在后台启动子 Agent）并行执行评估
>
> *("Concurrency: with 3+ pending URLs, launch subagents in the background to evaluate in parallel.")*

So three subagents start together, all read `reports/`, all see `041`, and all three claim `042`. Two evaluations are overwritten. Nothing errors — the run reports three successes.

## This was not drift; the English was already correct when the translation landed

`modes/zh/pipeline.md` was created in 207f9600 (2026-06-18, *feat(i18n): add Chinese language modes for China-market job seekers (#934) (#965)*). At that same commit the English mode already read:

```
:23  Claim the next sequential `REPORT_NUM` atomically by running `node reserve-report-num.mjs`
:65  Run `node reserve-report-num.mjs` to claim the next sequential number
```

The translation replaced an explicit command with a hand-rolled algorithm. Worth noting the English was internally consistent at the time: an unconditional 3+ fan-out is safe *because* the number is reserved atomically. Dropping the reservation while keeping the fan-out is what opens the race.

## Cross-language check

15 of the 18 `pipeline.md` files route through the reserver:

```
OK    da de es fr hi id it ja ko nl pl pt ru tr ua
BAD   ar  zh-TW  zh
```

## Scope

`zh` and `zh-TW` are what I can verify and write; a PR for those two follows. `modes/ar/pipeline.md:11` carries the same defect and I have left it for an Arabic speaker rather than machine-translating the replacement.

## What I am not claiming here

While checking this I noticed that the English `modes/pipeline.md:43` rule — browser-backed extraction must be processed one at a time, because workers sharing a Playwright session cross-contaminate and can evaluate the wrong posting — is absent from **all 17** localized `pipeline.md` files, not only the Chinese ones. That is a separate, systemic propagation gap of the same family as #3828 and #3168, so I am deliberately not folding it into this issue or the PR.

yuki-uix:
Routes the report-number step in both Chinese `pipeline.md` modes through `reserve-report-num.mjs` instead of a hand-rolled `max+1` scan.

Closes #3980

## What changed

- `modes/zh/pipeline.md:9` and `modes/zh-TW/pipeline.md:9` — claim the number with `node reserve-report-num.mjs`, release the sentinel with `--release <num>` after the report is written, mirroring the English `modes/pipeline.md:36`.
- Each line also states why, inline: `切勿自己扫描 reports/ 取最大值加 1 —— 并发时多个 worker 会算出同一个编号（#749）`. The surrounding mode files put the reason next to the instruction, and it is what stops the next translator reintroducing this.

Two lines total; no other content in either file is touched.

## Why this is a restoration, not a change

The English mode already read `node reserve-report-num.mjs` at 207f9600, the commit that introduced these translations, so this puts back what the source said rather than proposing new behaviour. The English was internally consistent at the time: an unconditional 3+ fan-out is safe *because* the number is reserved atomically. The translation dropped the reservation and kept the fan-out.

## Scope — what this PR deliberately does not do

- **`modes/ar/pipeline.md:11` carries the same defect and is left unfixed.** I do not write Arabic and would rather leave it to someone who does than machine-translate the replacement.
- **The Playwright-serial rule is still missing from these files.** English `modes/pipeline.md:43` requires browser-backed extraction to be processed one at a time, because workers sharing a Playwright session cross-contaminate. That rule is absent from **all 17** localized `pipeline.md` files, so after this PR both Chinese modes still recommend 3+ parallel subagents with Playwright as the preferred extractor. That is a separate, systemic propagation gap of the same family as #3828 and #3168, and folding it in here would mean writing the rule into 17 languages I mostly cannot check.

Flagging both explicitly so the merged file is not read as fully aligned with English.

## Suggested follow-up (not in this PR)

A parity assertion in `test-all.mjs` requiring every `modes/*/pipeline.md` to reference `reserve-report-num.mjs` — same shape as the existing `#1449 Phase 2b` extractor check at `test-all.mjs:4800`. Adding it now would go red on `ar`, which would amount to using CI to force a machine translation. It belongs in the PR that fixes `ar`.

## Verification

- Re-read `modes/pipeline.md` at `ref=207f9600` via the API to confirm the English was already correct when these files landed.
- Walked all 18 `modes/*/pipeline.md`: 15 already route through the reserver (`da de es fr hi id it ja ko nl pl pt ru tr ua`), 3 did not (`ar`, `zh-TW`, `zh`).
- Confirmed `--release` exists by running `node reserve-report-num.mjs --help`.
- Confirmed each Chinese file mentions report numbering in exactly one place, so one line per file is the whole fix. (The English `## Automatic numbering` section has no counterpart in either localized file — pre-existing, not introduced here.)
- `node test-all.mjs`: **8687 passed, 0 failed** on `origin/main` and with this change. The existing language-parity assertion (`all 18 language pipeline mirrors wire the opt-in extractor`) still passes.

## Summary

- `modes/zh/pipeline.md:9` and `modes/zh-TW/pipeline.md:9`: Reserve report numbers atomically with `node reserve-report-num.mjs`.
- Release each reservation after writing the report.
- Do not calculate `max+1`; parallel workers could select the same number and overwrite reports.

Only `modes/` changed. `AGENTS.md`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/` were not changed.

### DIFF
--- a/modes/zh-TW/pipeline.md
+++ b/modes/zh-TW/pipeline.md
@@ -6,7 +6,7 @@
 
 1. **讀取資料**：分析 `data/pipeline.md`，找出 "Pending" 區塊下所有標記為 `- [ ]` 的待處理項目。
 2. **逐一處理**：對每一個未處理的 URL：
-   a. **計算新報告編號**：掃描 `reports/` 目錄，找出目前最大的三位數字前綴並加 1，作為 `REPORT_NUM`。
+   a. **領取新報告編號**：執行 `node reserve-report-num.mjs` 原子地佔用下一個順序編號作為 `REPORT_NUM`；報告寫入後執行 `node reserve-report-num.mjs --release <num>` 釋放佔位。切勿自行掃描 `reports/` 取最大值加 1 —— 並行時多個 worker 會算出同一個編號（#749）。
    b. **擷取職缺描述 (JD)**：首選 Playwright（透過 `browser_navigate` + `browser_snapshot`）渲染網頁擷取；次選 `WebFetch` 抓取靜態文字；最後以 `WebSearch` 搜尋同名職缺的快照。
    c. **例外處理**：若連結因權限、失效等原因完全打不開，將該項標記為 `- [!]`，附上錯誤描述，然後繼續處理下一個。
    d. **執行一鍵管線評估**：跑 A–F 各維度評估 → 存成報告 `.md` → 依設定產生履歷 PDF（若評分達門檻）→ 自動登錄至 tracker。
--- a/modes/zh/pipeline.md
+++ b/modes/zh/pipeline.md
@@ -6,7 +6,7 @@
 
 1. **读取数据**：分析 `data/pipeline.md` 文件，找出 "Pending" 模块下所有标记为 `- [ ]` 的待处理项目。
 2. **循环遍历处理**：对每一个未处理的 URL：
-   a. **计算新报告编号**：扫描 `reports/` 目录，找出当前最大的三位数字前缀并加 1，作为 `REPORT_NUM`。
+   a. **领取新报告编号**：运行 `node reserve-report-num.mjs` 原子地占用下一个顺序编号作为 `REPORT_NUM`；报告写入后运行 `node reserve-report-num.mjs --release <num>` 释放占位。切勿自己扫描 `reports/` 取最大值加 1 —— 并发时多个 worker 会算出同一个编号（#749）。
    b. **抓取职位描述 (JD)**：首选 Playwright (通过 `browser_navigate` + `browser_snapshot`) 渲染网页抓取，备选 `WebFetch` 提取静态文本，最后使用 `WebSearch` 搜索同名岗位快照。
    c. **异常处理**：若链接因权限、失效等原因完全无法打开，将该项标记为 `- [!]`，附加错误描述，并继续处理下一个。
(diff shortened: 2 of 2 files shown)


santifer:
Hi @yuki-uix, this PR has waited 15 days without a reply from a maintainer. That is on us, not on you, and we are sorry for it.

PR #3993, opened 2026-09-07
docs(agents): twelve market mode sets ship and were undocumented (#3992)

### LINKED ISSUE #3992: Twelve market mode sets ship and are missing from AGENTS.md's market table
Eighteen complete market mode sets are on disk — each with `_shared.md`, an evaluation mode, an apply mode and `pipeline.md`:

```
ar  da  de  es  fr  hi  id  it  ja  ko  nl  pl  pt  ru  tr  ua  zh  zh-TW
```

`AGENTS.md`'s market table lists six: `de fr ar ja tr hi`.

The other **twelve** — `es pt it nl pl da ru ua zh zh-TW ko id` — are complete and undocumented.

### Why this is not cosmetic

`AGENTS.md` is loaded into every agent session, and its own rule for selecting a set is driven by that table:

> you detect a JD written in that language → *suggest* switching

A set the table does not name is a set the agent cannot route to, whatever is on disk. A Spanish JD cannot reach `modes/es/`; a Brazilian one cannot reach `modes/pt/`. Twelve full localizations, invisible to the system that would use them.

### Note for a fix

The "Local vocabulary (examples)" column should be verifiable, not written from general knowledge of each market — these files are read as instruction. Confirming each term appears in that market's own files (`grep -rF` over `modes/{dir}/`) keeps the column honest, and a short row beats a fuller invented one.

Worth a guard in both directions: a shipped set missing from the table, and a row naming a directory or mode file that does not exist. The second matters because a dead row costs a tool call every session an agent tries to follow it.

PR to follow.

abankar1:
Fixes #3992.

Eighteen complete market mode sets ship. The market table listed six.

```
on disk      ar da de es fr hi id it ja ko nl pl pt ru tr ua zh zh-TW   (18)
documented   ar de fr hi ja tr                                          (6)
```

`AGENTS.md` is loaded into every agent session, and its own rule for choosing a set — *"you detect a JD written in that language → suggest switching"* — is driven by that table. A set the table does not name is a set the agent cannot route to, whatever is on disk. A Spanish JD could not reach `modes/es/`, a Brazilian one could not reach `modes/pt/`, and so on for twelve complete localizations.

## Every cell is verified, not remembered

The directory and both mode filenames come from the filesystem. The **Local vocabulary** terms were each confirmed to appear in that market's own files (`grep -rF` over `modes/{dir}/`) rather than supplied from general knowledge of the market:

```
es     Contrato indefinido, convenio, pagas extra, Seguridad Social
pl     Umowa o pracę, B2B, ZUS, okres wypowiedzenia, urlop
ru     ТК РФ, оклад, испытательный срок, самозанятый, ДМС
zh     五险一金, 试用期, 年终奖, 劳动合同, 竞业
...
```

Where a term could not be confirmed in the files it is not listed. This table is read as instruction, so a short honest row beats a fuller invented one — and the repo's own rule is that keywords get reformulated, never fabricated.

I originally tried to derive the column mechanically from `_shared.md`, which turned out to be the translated scoring guidance rather than employment vocabulary; the market terms live in the evaluation modes and READMEs. Grepping the whole market directory is what made each term checkable.

## Guard, in both directions

- a shipped set missing from the table reddens
- a row naming a set **or a mode file** that does not exist reddens

The second direction matters because a dead row is not harmless: it costs the agent a tool call every session it tries to follow it. Verified by deleting the twelve rows (reddens) and by adding a `modes/tlh/` row (reddens).

Full suite green at 8686.

## Summary

- Added 12 market mode entries to `AGENTS.md: market modes table`.
- Agents can now route job descriptions for Spanish, Portuguese, Italian, Dutch, Polish, Danish, Russian, Ukrainian, Simplified Chinese, Traditional Chinese, Korean, and Indonesian.
- Added filesystem-verified directories, evaluation modes, apply modes, and local vocabulary examples.
- Added bidirectional validation in `tests/market-modes-documented.test.mjs:1-71`. The tests detect undocumented shipped sets and invalid table references.

## Files touched

- `AGENTS.md`
- `tests/market-modes-documented.test.mjs`

The following system files were not touched: `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -280,6 +280,18 @@ Default modes are in `modes/` (English). Market-specific mode sets (each include
 | Japanese (Japan) | `modes/ja/` | `kyujin` / `oubo` | 正社員, 賞与, みなし残業, 年俸制, 36協定 |
 | Turkish (Turkey) | `modes/tr/` | `is-ilani` / `basvuru` | SGK, kıdem tazminatı, brüt/net maaş, BES |
 | Hindi (India) | `modes/hi/` | `naukri` / `aavedan` | CTC vs. in-hand, PF/EPF, Notice period/buyout, ESOPs |
+| Spanish (ES/LatAm) | `modes/es/` | `oferta` / `aplicar` | Contrato indefinido, convenio, pagas extra, Seguridad Social |
+| Portuguese (BR/PT) | `modes/pt/` | `oferta` / `aplicar` | CLT, PJ, FGTS, 13º, férias, vale-refeição |
+| Italian (Italy) | `modes/it/` | `annuncio` / `candidarsi` | CCNL, tempo indeterminato, tredicesima |
+| Dutch (NL/BE) | `modes/nl/` | `vacature` / `solliciteren` | vakantiegeld, proeftijd, opzegtermijn, pensioen |
+| Polish (Poland) | `modes/pl/` | `oferta` / `aplikuj` | Umowa o pracę, B2B, ZUS, okres wypowiedzenia, urlop |
+| Danish (Denmark) | `modes/da/` | `oferta` / `apply` | løn, opsigelsesvarsel, ferie, overenskomst, A-kasse |
+| Russian | `modes/ru/` | `oferta` / `apply` | ТК РФ, оклад, испытательный срок, самозанятый, ДМС |
+| Ukrainian (Ukraine) | `modes/ua/` | `oferta` / `apply` | ФОП, КЗпП, оклад, випробувальний термін |
+| Chinese, Simplified | `modes/zh/` | `oferta` / `apply` | 五险一金, 试用期, 年终奖, 劳动合同, 竞业 |
+| Chinese, Traditional | `modes/zh-TW/` | `oferta` / `apply` | 勞保, 試用期, 年終獎金, 勞動契約, 特休 |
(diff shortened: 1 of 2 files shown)


Scott-Emberson (commented):
Test-axis only. The substance adds twelve rows to the market-mode table in `AGENTS.md`, a red-line doc and the maintainer's call; I am reviewing the owned test `tests/market-modes-documented.test.mjs`.

It is sound, a genuine drift guard rather than a tautology. One side reads the real filesystem (`readdirSync('modes')`, keeping every subdir that carries `_shared.md`, the actual on-disk market sets), the other regex-parses the real market table in AGENTS.md, and the two derived sets are compared, with no hardcoded name list on either side (confirmed live: 18 shipped, 18 documented). The guard is bidirectional (every shipped set is documented, and the table names no set that does not ship) and a third test asserts every eval/apply mode filename the table names exists on disk, with a row-count floor so a table-format change also reddens. I mutation-checked both directions: removing a set's `_shared.md` reddens the ghost-row check, and deleting its table row reddens the missing-set check.

One honest limit worth noting: the test checks directory and mode-file existence, not the correctness of the "Local vocabulary" cells, so a wrong vocabulary term would not be caught here. That verification lives in the human doc review, consistent with the PR's claim that the terms were grep-confirmed.

Owned test sound, not behind. SIGNAL (AGENTS.md red-line), so not surfacing as ready-to-approve; routing and merge are the maintainer's.

PR #3994, opened 2026-09-07
test(providers): cover intInRange, the last untested file under providers/

abankar1:
89 of the 90 files under `providers/` are referenced by a test. `_config-utils.mjs` was the one that was not, and three providers depend on it for the values that decide whether a scan works at all:

```
vdab.mjs:98-101            days 1..1000, size 1..100, detailLimit 1..100
mycareersfuture.mjs:69,72  size 1..MAX_PAGE_SIZE, maxPages 1..MAX_PAGES_CAP
smartrecruiters.mjs:38     detailLimit 1..100
```

Its own docstring states the stake: a stray `portals.yml` value *"can't produce an empty (e.g. size=0) or pathological query"*. A silent `size: 0` returns nothing and reads as "this board has no jobs"; a silent `maxPages: 100000` is a very long sweep against someone else's API. Neither errors.

## One thing worth your call

Writing this turned up behaviour the parameter name does not suggest, so the test **documents** it rather than assuming:

```js
Number(null)  // 0, not NaN
Number('')    // 0
Number([])    // 0
```

Those miss the `def` fallback and are clamped by the lower bound instead. So an empty YAML key:

```yaml
size:            # -> null -> 0 -> size = 1
```

gives **one result per page**, not the default 100. The scan still runs, so nothing errors — it just makes 100x the requests, on a board that looks slow rather than misconfigured. Arguably the "pathological query" the function exists to prevent, arrived at from the other side.

I have **not** changed it. Making `null`/`''` take the default instead of the minimum would change three providers' behaviour on existing configs, and that is not a test-only decision. The test pins current behaviour with a message that says so, so if you do change it the failure reads as *"this is now the default, and this test should say so"* rather than as a regression.

Happy to follow up with the change if you want it.

Eight tests, no production change. Full suite green at 8686.

## Summary

Adds eight tests for `intInRange` in `tests/provider-config-utils.test.mjs:1`.

Users gain documented coverage for provider configuration values used by `vdab`, `mycareersfuture`, and `smartrecruiters`. The tests cover validation, inclusive bounds, clamping, defaults, numeric coercion, zero-like values, infinities, and fraction truncation.

No production behavior changes. No system files are touched: `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/` remain unchanged.

### DIFF
--- a/tests/provider-config-utils.test.mjs
+++ b/tests/provider-config-utils.test.mjs
@@ -1,0 +1,96 @@
+// tests/provider-config-utils.test.mjs — intInRange is the only guard between a
+// hand-edited portals.yml and a provider's outbound query.
+//
+// It was the one file under providers/ that no test referenced (89 others are
+// covered), and three providers depend on it for exactly the values that decide
+// whether a scan works at all:
+//
+//   vdab.mjs:98-101            days 1..1000, size 1..100, detailLimit 1..100
+//   mycareersfuture.mjs:69,72  size 1..MAX_PAGE_SIZE, maxPages 1..MAX_PAGES_CAP
+//   smartrecruiters.mjs:38     detailLimit 1..100
+//
+// Its own docstring states the stake: a stray value "can't produce an empty
+// (e.g. size=0) or pathological query". A silent `size: 0` returns nothing and
+// reads as "this board has no jobs"; a silent `maxPages: 100000` is a very long
+// sweep against someone else's API. Neither errors.
+//
+// Run:  node --test tests/provider-config-utils.test.mjs
+
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { dirname, join } from 'node:path';
+import { fileURLToPath, pathToFileURL } from 'node:url';
+
+const ROOT = dirname(dirname(fileURLToPath(import.meta.url)));
+const { intInRange } = await import(pathToFileURL(join(ROOT, 'providers', '_config-utils.mjs')).href);
+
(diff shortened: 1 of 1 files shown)


Scott-Emberson (commented):
Verified the owned test by execution and mutation. Sound and gate-relevant: this is a test-only PR entirely under tests/, our owned path, covering `intInRange` in `providers/_config-utils.mjs` (the last untested file under providers/).

The test dynamically imports the real `intInRange` from `providers/_config-utils.mjs`, not a shadow copy, and asserts against literal expecteds across thorough boundaries: inclusive min and max, clamp below and above, NaN falling back to the default, the non-obvious zeroish path (null/empty/array coercing to 0 then to min, documented as real-but-surprising and deliberately pinned rather than changed in a test-only PR), numeric strings, fractional truncation-then-clamp, and Infinity to default, plus the real caller bounds sampled from the three importing providers. I mutation-checked it: flipping the lower clamp reddens the suite (8 pass to 3 pass / 5 fail via the real harness). Not tautological.

GATE-tier (tests/), owned test sound, CI green, not behind. Holding formal approval until Scott gives the go-ahead; surfacing it as ready-to-approve.

santifer (approved):
Covering the last untested file under providers/ is useful on its own, but what makes this easy to land is that the tests read as documentation: each assertion says what breaks and why. I reverted the clamp twice locally to check the suite catches the regressions it describes, and it does. You were also right to pin the zero-like behaviour rather than change it here.

PR #3995, opened 2026-09-07
fix(web): surface a successful render's warnings instead of dropping them

### LINKED ISSUE #3984: bug(web): a successful pdf render discards generate-pdf.mjs stderr, hiding fact-check advisories and the page-budget warning
## What happens

On the web `pdf` path, `generate-pdf.mjs` can exit 0 while printing warnings to stderr, and the web layer throws that stderr away. The user is shown a clean render.

In `web/src/lib/pdf-render.mjs`, `spawnGeneratePdf` collects stderr and returns it:

```js
child.stderr.on("data", (d) => { stderr += d.toString(); });
child.on("close", (code) => resolve({ ok: code === 0, stderr: stderr.trim() }));
```

and its only caller reads that field in one branch:

```js
const render = await spawnGeneratePdf({ ... });
if (!render.ok) {
  return { kind: "render-failed", error: render.stderr || "PDF rendering failed." };
}
```

So `render.stderr` is consulted only when the render failed. On success it is collected and dropped.

Two warnings travel that channel:

1. `enforcePageBudget` (generate-pdf.mjs:1019) defaults to `maxPages = 2`, and on overflow calls `console.warn` and continues (line 1038).
2. `assertFacts` runs by default (generate-pdf.mjs:1355) and reports advisory phrases through `console.warn` (line 1359).

The second is the one I care about. `pdf` mode is precisely where an agent rewrites CV content, so the fact checker is the guard against a model asserting something the source CV does not support. On the web path that guard still runs, still fires, and is then discarded before anyone sees it.

## Reproduce

Render a tailored CV that is longer than two pages:

```
node generate-pdf.mjs <tailored-cv.html> output/cv.pdf --format=letter --report=001 --allow-reorder
```

stderr carries the warning and the exit code is 0:

```
CV is 3 pages; the allowed maximum is 2 pages. Trim lower-priority bullets, older
roles, secondary projects, or the competencies strip, then regenerate. Continuing
because overflow is warning-only by default; use --strict-pages to reject it.
```

Run the same HTML through the web `pdf` worker and `renderAndMarkPdf` resolves `{ kind: "rendered", warnings: [] }`. The UI reports a successful render with no warning at all. The `warnings` array is populated from the defaulted page format and the tracker mark, never from the render's own stderr.

Seen on v1.32.0, web 0.10.0.

## Impact

A tailored CV can exceed the documented page budget, or trip the fact checker, and still present as a clean render. A CLI user sees both warnings; a web user sees neither, and the web UI is the surface where a less experienced user is least likely to re-read the PDF critically.

The fact-check half is the more serious one. On a lower-capability CLI, a tailoring run can assert a certification or a metric the source CV does not contain. That is exactly what `assertFacts` exists to catch, and on the web path its advisory is guaranteed to be thrown away.

## Suggested fix

Surface a successful render's stderr as warnings instead of discarding it. `renderAndMarkPdf` already builds a `warnings` array and the route already streams it to the client, so passing the stderr lines through when `ok` is true restores parity with the CLI without changing any exit-code behaviour.

Happy to open a PR if the approach looks right.

rubicon:
## What does this PR do?

`renderAndMarkPdf` now passes a successful render's stderr through as warnings instead of discarding it. `spawnGeneratePdf` resolves `{ ok, stderr }` for every run, but `render.stderr` was read only inside the `!render.ok` branch, so on a clean exit the collected stderr went nowhere.

## Related issue

Closes #3984

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

---

### What was being dropped

Two advisories ride that channel, and both are written by a run that exits 0:

| Source | Written by | Condition |
|---|---|---|
| `enforcePageBudget` (`generate-pdf.mjs:1019`) | `console.warn` at line 1038 | CV exceeds `maxPages` (default 2) and `--strict-pages` was not passed, so overflow is warning-only |
| `assertFacts` (`generate-pdf.mjs:1355`) | `console.warn` at line 1359, one line per advisory phrase | the fact gate returns `warn` |

The second is why this is worth fixing rather than logging. `pdf` mode is where an agent rewrites CV content, so the check that flags an unsupported claim is aimed at the person who is about to send that CV. On the web path it was collected into `render.stderr` and then never read.

### The fix

The stderr is split on newlines and pushed into the `warnings` array `renderAndMarkPdf` already returns; the route already streams that array through `sendWarnings`. Blank lines are skipped, so a clean render with empty stderr still produces zero warnings rather than one empty one.

Nothing about exit codes changes. A non-zero exit still returns `render-failed` with the same message, ahead of this block.

### Test evidence

Two cases added to `web/tests/lib/pdf-render.test.mjs`, one per direction, both mutation-checked:

| Mutation | Result |
|---|---|
| Revert the change (read stderr only on failure) | RED on the non-empty case (0 warnings, 3 expected) |
| Drop the blank-line guard (`warnings.push(text)` unconditionally) | RED on the empty case |

`node test-all.mjs`: 8687 passed, 0 failed. `web/`: 499 passed, 0 failed.

## Summary

Successful web PDF renders now show renderer advisories to users. `renderAndMarkPdf` splits non-empty `stderr` lines into warnings at `web/src/lib/pdf-render.mjs:201-210`.

Failed renders keep the existing `render-failed` behavior at `web/src/lib/pdf-render.mjs:198`. Empty `stderr` produces no warnings.

## Tests

`web/tests/lib/pdf-render.test.mjs:325-371` verifies:

- Successful renders expose each stderr line as a warning.
- Empty stderr does not create warnings.

## Files changed

- `web/src/lib/pdf-render.mjs`
- `web/tests/lib/pdf-render.test.mjs`

### DIFF
--- a/web/src/lib/pdf-render.mjs
+++ b/web/src/lib/pdf-render.mjs
@@ -172,7 +172,7 @@ export function cleanupPdfScratch(scratchDir, prefix) {
 /**
  * @typedef {Object} RenderedResult
  * @property {"rendered"} kind
- * @property {string[]} warnings - Non-fatal issues to surface to the user (e.g. a tracker row that was not marked).
+ * @property {string[]} warnings - Non-fatal issues to surface to the user (e.g. a renderer advisory, or a tracker row that was not marked).
  */
 /** @typedef {RenderFailedResult | RenderedResult} RenderResult */
 
@@ -196,6 +196,18 @@ export async function renderAndMarkPdf({ spawnFn, execPath, root, pdfPaths, form
 
   if (!render.ok) {
     return { kind: "render-failed", error: render.stderr || "PDF rendering failed." };
+  }
+
+  // A render that exits 0 can still have written advisories to stderr, and two
+  // of them matter to whoever asked for this PDF: the page-budget overflow
+  // (warning-only unless --strict-pages) and the CV fact check's advisory
+  // phrases. Reading render.stderr only in the failure branch above collected
+  // both and dropped them, so on the web path an unsupported claim the fact
+  // gate flagged reached nobody. One warning per line, blank lines skipped, so
+  // a clean render still reports none.
+  for (const line of render.stderr.split("\n")) {
+    const text = line.trim();
+    if (text) warnings.push(text);
   }
 
   // The PDF is the real deliverable and it already rendered successfully — a
--- a/web/tests/lib/pdf-render.test.mjs
+++ b/web/tests/lib/pdf-render.test.mjs
@@ -322,6 +322,58 @@ test("renderAndMarkPdf: render succeeds but mark-pdf-ready fails with no parseab
   }
 });
 
+test("renderAndMarkPdf: successful render with stderr -> that stderr is surfaced as warnings", async () => {
+  // Given generate-pdf.mjs exits cleanly but wrote advisories to stderr — the
+  // page-budget overflow and the CV fact check both warn on that channel
+  const dir = makeScratchDir();
+  const pdfPaths = makePdfPaths(dir, "6");
+  writeFileSync(pdfPaths.html, "<html></html>");
+  const stderr = [
+    "\u26a0\ufe0f  CV is 3 pages; the allowed maximum is 2 pages.",
+    "\u26a0\ufe0f  CV fact check warning: cv-web-6.html",
+    "  - advisory phrase: world-class",
+  ].join("\n");
+  const { spawnFn } = makeRouterSpawn({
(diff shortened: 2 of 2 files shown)


santifer (approved):

santifer:
Merged. Thank you, @rubicon: **when a PDF renders successfully, the web app now shows the render's warnings instead of dropping them**, such as an extra page or a sentence the fact check could not back. Those are exactly the notes a person needs before sending a CV, and the two tests pin both the warning and the clean case.

Thanks for making the web app say what the CLI already knew.

PR #3996, opened 2026-09-07
fix(analyze-patterns): an empty tracker is a new user, not a failed command

abankar1:
Eleven analysis scripts read the same tracker. With every required input present and empty:

```
exit 0   stats  upskill  salary-gap  process-quality  rejection-latency
         detect-reposts  company-history  calibrate  funnel-velocity
         tracker-sync-check
exit 1   analyze-patterns          <-- same condition
```

An empty tracker is the state of a new user — `AGENTS.md` tells them to run these before they have data. Exiting 1 breaks `&&` chaining and makes the batch runners treat a fresh install as a broken command.

## On the kind of error, not the message

`analyze()` marks the empty-tracker return with `noData: true`, and the exit stays non-zero for anything else. Matching on `'No applications found in tracker.'` would have tied the exit status to prose that is free to change.

## A scope correction I made along the way

My first measurement had `negotiation-roi` in this set as a second outlier. It was not: its exit 1 came from resolving `story-bank.md` off the **code** root and never finding the fixture at all — a path bug, fixed separately in #3985. My fixture had left it without a story bank, which is a *missing input*, and a missing input is a legitimate non-zero.

Re-measured with every required input present and empty, it exits 0 like the rest. The test excludes it on the honest grounds: it reads the story bank, not the tracker, so "empty tracker" is not a condition it has an opinion about.

## Tests

Asserted as **agreement** across the eleven rather than a hardcoded 0 for one script, so it keeps holding if the convention is ever revisited and a new analysis script is measured against its siblings instead of a number in a test.

Two guards: the shared answer must be *success* (agreement could also be reached by making all eleven exit 1), and an unreadable tracker must still fail — otherwise this would buy consistency by never reporting anything.

Reverting the one-line change reddens 2 of 3; the failure guard stays green. Full suite green at 8686.

## Summary

Users can run `analyze-patterns` with an empty tracker and receive a successful result. The command reports `noData: true` instead of failing because no applications exist.

Other tracker errors remain non-zero. Unreadable trackers still fail.

The new test checks consistent exit codes across the eleven tracker-based analysis scripts and verifies unreadable tracker handling: `tests/empty-tracker-exit-code.test.mjs:1`.

Changed files: `analyze-patterns.mjs`, `tests/empty-tracker-exit-code.test.mjs`.

### DIFF
--- a/analyze-patterns.mjs
+++ b/analyze-patterns.mjs
@@ -1250,7 +1250,9 @@ function analyze() {
   const entries = parseTracker();
 
   if (entries.length === 0) {
-    return { error: 'No applications found in tracker.' };
+    // noData marks this as the empty-tracker case rather than a failure, so the
+    // exit status below does not have to match on the message text.
+    return { error: 'No applications found in tracker.', noData: true };
   }
 
   // Enrich entries with report data and classification
@@ -1690,5 +1692,16 @@ if (isMainModule(import.meta.url)) {
     console.log(JSON.stringify(result, null, 2));
   }
 
-  if (result.error) process.exit(1);
+  // "No applications found" is the state of a NEW USER, not a failure. Every
+  // other analysis script over the same tracker — stats, upskill, salary-gap,
+  // process-quality, rejection-latency, detect-reposts, company-history,
+  // calibrate, funnel-velocity, tracker-sync-check — reports it and exits 0.
+  // This one exited 1, which breaks `&&` chaining and makes the batch runners
+  // treat an empty tracker as a broken command.
+  //
+  // Still non-zero for a genuine failure: the check is on the KIND of error, so
+  // a future `result.error` that is not "no data" keeps its exit 1. Written as
+  // an allowlist of no-data codes rather than a message match, so the exit
+  // status does not depend on prose.
+  if (result.error && !result.noData) process.exit(1);
 }
--- a/tests/empty-tracker-exit-code.test.mjs
+++ b/tests/empty-tracker-exit-code.test.mjs
@@ -1,0 +1,119 @@
+// tests/empty-tracker-exit-code.test.mjs — an empty tracker is the state of a
+// NEW USER, not a failed command.
+//
+// Eleven analysis scripts read the same tracker. Ten reported "no data yet" and
+// exited 0; analyze-patterns.mjs exited 1 for the same condition, which breaks
+// `&&` chaining and makes the batch runners treat a fresh install as a broken
+// command.
+//
+// Asserted as AGREEMENT across the set rather than against a hardcoded exit
+// code for one script, so this keeps holding if the convention itself is ever
+// revisited — and so a new analysis script joining the family is measured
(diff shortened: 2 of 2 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is `analyze-patterns.mjs` treating an empty tracker as a new-user state rather than a failure, core and the maintainer's call; I am reviewing the owned test `tests/empty-tracker-exit-code.test.mjs`.

It is sound and asserts the specific contract, not a bare non-crash. The test spawns the real scripts over an on-disk empty-tracker fixture (valid header, zero rows) under a temp CAREER_OPS_ROOT and reads the real exit code, not a reconstructed check. It asserts all eleven tracker-reading scripts return one shared exit code, that the shared code is 0 (explicitly including analyze-patterns.mjs, the load-bearing new-user assertion), and that a genuine failure (tracker path pointing at a directory) still exits non-zero, so the fix cannot buy consistency by swallowing real errors. I mutation-checked it: reverting the fix so an empty tracker exits 1 reddens 2 of 3 (the genuine-failure guard correctly stays green).

One note: the test asserts the exit code, not the friendly message string. That is deliberate, matching the fix's design of a `noData` flag so the exit status does not depend on prose, and the exit-code contract is the real behavioral guarantee.

Owned test sound, not behind. SIGNAL (analyze-patterns core), so not surfacing as ready-to-approve; routing and merge are the maintainer's.

santifer:
Thanks for this, @abankar1, and sorry it sat eleven days without a word from me.

The measurement is what makes it. Ten scripts reading the same tracker report an empty one and exit 0; this one exited 1 for the same condition. That is a disagreement, not a preference, and `&&` chaining is where a new user runs into it: a fresh install looks like a broken command.

Gating on the kind of error instead of matching the message text is the right call, and worth saying because the cheaper version of this fix would have tied the exit status to prose that is free to change tomorrow. A genuine failure still exits 1.

I ran your branch merged onto today's main rather than trusting the PR checks: full suite green, 9202 passed, 0 failed.

The scope correction you made yourself, pulling `negotiation-roi` back out once you found its exit 1 was a path bug and not this, is the kind of thing I would much rather read than not.

Taking it in. I am bringing the branch up to date with main now; once its checks report on that new head I will merge.

santifer (approved):

santifer:
Merging this, @abankar1.

One thing I measured while re-checking the head: `parseTracker()` returns an empty array both when the tracker is empty and when it is missing, so this also moves a missing tracker to exit 0. Your test comment draws the opposite line, that a missing input is a legitimate non-zero. So I ran the eleven scripts against a data root with no tracker at all: nine of the ten siblings already exit 0 there. The new behaviour is the family's, not an exception to it.

The one that disagrees is `calibrate.mjs`, which still exits 1 on a missing tracker. That is the remaining inconsistency, and it is ours to file rather than yours to fix here.

Thanks for the measurement that opened this: asserting agreement across the set rather than a number for one script is why it was cheap to check.

PR #3998, opened 2026-09-07
fix(cli): --help was discarded, so the script ran instead

abankar1:
Six scripts did not recognise the flag, and an unrecognised flag is simply discarded — so `--help` fell through to the work:

```
calibrate.mjs --help               printed a full calibration report
salary-gap.mjs --help              printed JSON
tracker-sync-check.mjs --help      printed JSON
story-provenance-check.mjs --help  printed JSON
```

Nothing errored, which is the problem. The user asked what the flags are and got output answering a different question, with no sign the flag was never read. On a script that writes, the same silent discard would run the write.

## Usage derived, not remembered

None of the six had a `USAGE` const, and only two documented usage in their header. The text is built from the flags each file **actually accepts**, because a usage block naming an option the script ignores is worse than none — it sends the user to write a flag that silently does nothing. A test asserts that correspondence rather than trusting me to have got it right.

## Two invariants per script

Either alone is satisfiable the wrong way:

- help must be **printed**
- the work must **not** have run

Exiting 0 with a usage block would still be wrong if the analysis also ran — the original bug produced output too.

## The other nineteen

The census that found these six turned up nineteen more CLIs with the same defect (`generate-pdf`, `tracker`, `plugins`, `intake`, …). That is far more than one change should touch, so it is a **shrinking allowlist**, in the style this repo already uses for coverage ratchets:

- a **new** CLI cannot join the list — it must handle `--help`
- a script that gets fixed must be **removed** from the list, or the test says so

Verified both directions: adding `calibrate.mjs` back to the list reddens with *"these now handle --help and should be removed"*, and reverting one fix reddens 4 of 20.

I'd rather hand you a ratchet that describes the real state than an assertion that nineteen scripts are fine when they are not. Happy to work the list down in follow-ups.

Full suite green at 8686.

## User impact

- Users can run `--help` or `-h` for six CLI scripts and receive usage information.
- Each script exits before normal analysis, checks, or report generation starts.
- Usage text lists the flags accepted by each script.
- Tests detect new CLIs that silently discard `--help`.
- The allowlist contains the remaining affected CLIs and shrinks when a CLI is fixed.

## Files changed

- `calibrate.mjs`: Adds usage output and early help handling.
- `jd-skill-gap.mjs`: Adds usage output and early help handling.
- `negotiation-roi.mjs`: Adds usage output and early help handling.
- `salary-gap.mjs`: Adds usage output and early help handling.
- `story-provenance-check.mjs`: Adds usage output and early help handling.
- `tracker-sync-check.mjs`: Adds usage output and early help handling.
- `tests/help-flag-handled.test.mjs`: Verifies help output, successful exit, prevented execution, flag correspondence, and the shrinking allowlist.

### DIFF
--- a/calibrate.mjs
+++ b/calibrate.mjs
@@ -436,7 +436,24 @@ function selfTest() {
 // against the real tracker inside the test process.
 import { isMainModule } from './lib/is-main-module.mjs';
 
+// Derived from the flags this file actually accepts, so `--help` cannot
+// describe an option that does not exist.
+const USAGE = `Usage:
+  node calibrate.mjs [--json] [--min-band-n <n>] [--self-test]
+
+  --json           full JSON to stdout
+  --min-band-n <n> per-band sample floor (default 5)
+  --self-test      run the built-in checks
+  --help, -h   print this and exit`;
+
 if (isMainModule(import.meta.url)) {
+  // BEFORE any work. Unhandled, `--help` fell through to the analysis: this
+  // script printed a full report for it, which is not what the flag asks for
+  // and hides that it was never recognised.
+  if (process.argv.slice(2).some((a) => a === '--help' || a === '-h')) {
+    console.log(USAGE);
+    process.exit(0);
+  }
   const argv = process.argv.slice(2);
   if (argv.includes('--self-test')) selfTest();
 
--- a/jd-skill-gap.mjs
+++ b/jd-skill-gap.mjs
@@ -814,7 +814,23 @@ Maintained the internal Fabrikam-SDK build.
 
 // ── Main ─────────────────────────────────────────────────────────────
 
+// Derived from the flags this file actually accepts, so `--help` cannot
+// describe an option that does not exist.
+const USAGE = `Usage:
+  node jd-skill-gap.mjs <jd-file> [--summary] [--self-test]
+
+  --summary    human-readable output instead of JSON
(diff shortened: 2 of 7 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is CLI --help handling across six root scripts plus a lib interaction, core and the maintainer's call; I am reviewing the owned test `tests/help-flag-handled.test.mjs`.

It is sound. The test spawns each real script with --help and -h and asserts on the real process status and stdout, not a reconstructed flag parser, enforcing two distinct invariants per script: help is printed (usage line, exit 0) and the work did not run (no leading JSON, under 40 lines of output). A third test cross-checks that every flag named in a usage block is actually referenced in that script's source. On the drift question it is layered correctly: the six spawn-tested scripts are this PR's changeset (a fair hardcoded set), while a separate census test enumerates every real CLI via isMainModule and enforces a shrinking allowlist, so a new CLI lacking --help handling fails rather than silently drifting. I mutation-checked it: reverting the guard in calibrate.mjs so --help falls through to the analysis reddens 4 assertions. There is no conflict with the lib/cli-flags requireOperand path (#3945): the help guard short-circuits ahead of any flag validation.

Owned test sound, not behind. SIGNAL (CLI core), so not surfacing as ready-to-approve; routing and merge are the maintainer's.

PR #4000, opened 2026-09-07
feat(providers): add BWI GmbH careers-site provider

duathron:
## What

Adds a zero-auth provider for **BWI GmbH** — Germany's armed-forces IT service provider, which runs one of the larger SOC / cyber-defense operations in the country (~270 open roles, a Cyber Defense Center in Cologne, security analyst and forensics roles posted continuously).

The careers site is self-hosted with no third-party ATS, so no feed-based provider matched it. Before this, the board was effectively invisible to the scanner apart from whatever leaked through job aggregators.

## How

Two server-rendered, zero-auth sources:

1. The listing page embeds one `application/ld+json` **CollectionPage** whose `mainEntity` ItemList carries **every** posting URL — not just the ~15 cards rendered above the fold. Items hold only a `url`, so titles are rebuilt from the slug (trailing `-m-w-d-69624` boilerplate stripped, `(m/w/d)` restored).
2. Each detail page embeds a proper **JobPosting** with title, `datePosted` and `jobLocation.address.addressLocality`.

Fetching ~270 detail pages per scan would be unreasonable, so every posting gets a slug-derived title and only security-adjacent slugs are enriched, capped at **45 detail requests**. Enrichment is best-effort: a failing detail page keeps the slug title and an empty location, which the central `location_filter` treats as pass. The candidate pattern is deliberately wider than any `title_filter` — it decides what to *enrich*, not what to *keep*, so a miss costs a location string and never a job.

## Hardening

Detail URLs originate in remote page content, so they pass an HTTPS + `www.bwi.de` host-allowlist guard before being fetched (following the `assertGreenhouseUrl` pattern), and every request uses `redirect: 'error'`. `detect()` matches the host exactly rather than by substring, so `www.bwi.de.attacker.example` is not claimed — that case has its own test.

## Tests

`tests/providers/bwi.test.mjs`, 15 assertions: id/detect contract including the lookalike-host rejection, both JSON-LD parsers (including skipping a malformed and an unrelated `ld+json` block), slug-to-title, `redirect: "error"` on every request, the enrichment budget spending exactly one detail request for one relevant posting, and resilience when a detail page fails.

```
node test-all.mjs --only providers/bwi
📊 Results: 15 passed, 0 failed, 0 warnings
```

Docs row added to `docs/SUPPORTED_JOB_BOARDS.md` in the same PR.

## Config

```yaml
- name: BWI GmbH
  careers_url: https://www.bwi.de/karriere/stellenangebote
  provider: bwi
  enabled: true
```

## Note

Many BWI roles require a German security clearance (Ü2/Ü3), which lengthens the hiring process. That is a property of the employer rather than of the provider — the scanner surfaces the postings and the user decides.

## User impact

Users can scan BWI GmbH jobs from `https://www.bwi.de/karriere/stellenangebote` with provider ID `bwi`.

The provider discovers postings from JSON-LD and enriches security-related jobs with title, date, and location data. It keeps postings when enrichment fails and limits detail requests to 45 pages: `providers/bwi.mjs:1-159`.

The provider accepts only HTTPS URLs on the exact `www.bwi.de` host and blocks redirects: `providers/bwi.mjs:100-117`.

Supported job board documentation now includes BWI GmbH: `docs/SUPPORTED_JOB_BOARDS.md`.

## System files touched

`providers/`: added the BWI provider.

`AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, and `.github/`: not changed.


Scott-Emberson (changes requested):
Requesting changes on the test axis. To be clear up front: the implementation looks safe, and the parse-side coverage is genuinely sound. The gap is that the SSRF fetch-path guard, the axis that matters most for a net-new scraper, has no test coverage, so a future regression that neutered it would merge with a green suite.

What is sound: the test imports the real `providers/bwi.mjs` (not a reconstructed copy), uses non-throwing recorder mocks, and its parse and routing assertions are mutation-sensitive. Breaking `slugToTitle`, flipping `redirect:'error'` to `'follow'`, and loosening the `detect()` host check from `===` to a superstring `.includes()` each redden. Fifteen assertions pass via the real harness.

The problem is what stays green. The shipped implementation does pin the fetch path (exact hostname equality in `assertBwiUrl`, HTTPS-only, `redirect:'error'`, and a literal host in `JOB_URL_RE`), but two mutations that gut those guards do not redden any test:

- Neutering `assertBwiUrl` to return the URL unconditionally: 15 of 15 still pass.
- Loosening the `JOB_URL_RE` host from the literal `www.bwi.de` to any host: 15 of 15 still pass.

So the only host-anchoring actually covered is `detect()` (a routing/claim check), not the SSRF fetch guard. No fixture ever presents an off-host URL inside the listing JSON-LD, so nothing asserts that a hostile detail URL taken from remote content is refused and never fetched. The recorder array is already in place, so closing this is roughly one fixture (an attacker-host URL in the listing) plus an assertion that the guard blocks it and no off-host detail fetch is recorded.

For a net-new scraper reaching a new external host, I would want that fetch-path spoof test present before this is gated, since it is exactly the guard whose silent regression would be the worst case. The substance (providers/bwi.mjs) routing and merge are the maintainer's call; this request-changes is specifically about the owned test not pinning the SSRF fetch guard.

PR #4001, opened 2026-09-07
feat(providers): support explicit personio tenant pin

duathron:
## Problem

`personio.mjs` resolves the tenant purely from `careers_url`, which must therefore already be on `<slug>.jobs.personio.de`. But a very common setup — especially across the DACH mid-market — is a **branded careers page that embeds the Personio tenant in an iframe**:

- `secuinfra.com/de/company/career/` → iframe → `secuinfra.jobs.personio.de`
- `perseus.de/karriere`, `glueckkanja.com/de/stellenanzeigen/`, `aware7.de/karriere/` — same shape

For those entries `resolveHost()` returns `null`, `detect()` declines, and the company is **silently skipped**. No error, no warning: the scan simply reports nothing for that employer, which is the failure mode that takes longest to notice. I lost four tracked companies this way without seeing a single log line.

Pointing `careers_url` at the raw Personio host works around it, but then the entry no longer documents the page a human would actually open.

## Fix

Honour an explicit `personio: <slug>` field on the entry, taking precedence over `careers_url`:

```yaml
- name: SECUINFRA
  careers_url: https://www.secuinfra.com/de/company/career/   # what a human opens
  provider: personio
  personio: secuinfra                                          # the tenant to scan
```

## Security

The slug is charset-restricted (`/^[a-z0-9][a-z0-9-]{0,62}$/i`) and the resulting URL still goes through the existing `assertPersonioUrl()` — HTTPS plus the `*.jobs.personio.(de|com)` host allowlist remain the only way a request URL is accepted. The pin selects a tenant, it cannot introduce a host.

A test covers the rejection: `personio: 'evil.example/../x'` yields `detect() === null`.

## Tests

Two assertions added to `tests/providers/personio.test.mjs` — the pin resolving to the right feed URL, and the malformed-slug rejection.

```
node test-all.mjs --only providers/personio
📊 Results: 32 passed, 0 failed, 0 warnings
```

Docs row in `docs/SUPPORTED_JOB_BOARDS.md` updated in the same PR.

## Summary

Users can set `personio: <slug>` to scan a Personio job board embedded in a branded careers page. The explicit tenant takes precedence over `careers_url` (`providers/personio.mjs:37-48`).

The slug accepts only letters, numbers, and hyphens. Generated HTTPS URLs still pass Personio host validation (`providers/personio.mjs:14,35-46,77-81`).

Tests cover explicit tenant resolution and malformed-slug rejection (`tests/providers/personio.test.mjs:25-50`). The supported job boards documentation now describes the field (`docs/SUPPORTED_JOB_BOARDS.md:69`).

The reported full suite passed with 9,202 tests and no failures.

## Files changed

- `providers/personio.mjs:35-48`: Adds validated tenant resolution.
- `tests/providers/personio.test.mjs:25-50`: Adds tenant resolution tests.
- `docs/SUPPORTED_JOB_BOARDS.md:69`: Documents the `personio` tenant field.

`providers/_types.js` does not yet declare the `personio` field on `PortalEntry` (`providers/_types.js:81-84`).


Scott-Emberson (commented):
Verified the owned test by execution and mutation. Sound and gate-relevant (both files under providers/ and tests/providers/).

The test imports the real `providers/personio.mjs` and drives `detect()` with an explicit tenant slug, exercising the real `resolveHost()` pin branch this PR adds, not a reconstructed copy. It passes (32 assertions via the real harness). SSRF is layered and pinned: an anchored host allowlist `^[a-z0-9][a-z0-9-]*\.jobs\.personio\.(de|com)$` anchored at both ends (the suffix-spoof case asserts null, so the anchoring itself is covered here), HTTPS-only in both the assert and the resolver, `redirect: 'error'` asserted via a non-throwing mock that captures options and checks the exact fallback call count, and a slug charset that admits no dots or slashes so the constructed host is always `<slug>.jobs.personio.de`, with `fetch()` re-running `assertPersonioUrl` as a second layer. I mutation-checked all three axes: dropping the slug validation reddens, resolving the pin to the wrong TLD reddens, and disabling the numeric-id filter in the parser reddens.

GATE-tier (providers/), owned test sound, CI green, not behind. Holding formal approval until Scott gives the go-ahead; surfacing it as ready-to-approve.

santifer:
Thanks for this, @duathron, and sorry for the eleven days of silence on it.

The failure mode you describe is the one that costs the most to notice: no error, no warning, the company simply stops appearing in the scan. A branded careers page with the tenant behind an iframe is common enough in the DACH mid-market that this has been quietly dropping boards, and the four examples in your description are exactly the shape of it.

I checked the slug guard instead of trusting it. Taking `PERSONIO_SLUG_RE` out makes `detect()` hand back a URL whose host is whatever the entry says, and the negative case you added goes red the moment it does, so that assertion carries weight rather than decorating the diff. `fetch()` stays protected on its own by `assertPersonioUrl` either way. Merged onto today's main, full suite green: 9202 passed, 0 failed.

One small thing I will add myself rather than send you round again for it: `personio` belongs in the `PortalEntry` typedef in `providers/_types.js`, alongside `offset_param`.

Taking it in. I am bringing the branch up to date with main now; once its checks report on that new head I will merge.

santifer (approved):

chipoto69:
github-steward merge gate check, 2026-09-18 21:39 CEST tick: PR is CLEAN/MERGEABLE, santifer approved at 2026-09-18T19:16Z, trusted checks are green (Tests ubuntu/macos/windows, CodeQL JS/Go, Dependency Review, No user data guard, cv-visual, Upgrade regression, CodeRabbit). Proceeding with squash merge; no branch edits or force-push.

chipoto69:
Correction/readback: the attempted squash merge was rejected by GitHub: chipoto69 lacks MergePullRequest permission on career-ops-hq/career-ops. The objective gate remains ready (approved, clean, trusted checks green), but final merge is owner:career-ops-maintainer only. Steward will not retry merge or push over the branch without maintainer permission.

PR #4023, opened 2026-09-08
Add JSON output support for set-status, with tests  issue #3855

### LINKED ISSUE #3855: set-status.mjs: the JSON result must be the last thing on stdout, and nothing pins it
## What

`set-status.mjs` prints its result as JSON on stdout and the web's `parseCliJson` (web/src/lib) reads it back. The whole contract rests on one unwritten rule: **the JSON document is the last thing printed**. Today that holds (the success result at `set-status.mjs:652`, the error object at `:238`), and nothing pins it.

## Why it matters

A single `console.log` added after the result (a farewell line, a hint, a debug print) makes every successful status write come back to the web as a 500, while the write itself lands in `applications.md` and `status-log.tsv`. That is exactly the shape of the bug #3602 just fixed on the parser side (pretty-printed success read as "no result"): the UI reverts the control, the user reloads, the change is there, and nobody reports it.

## What to do

Add a test under `tests/` that runs `set-status.mjs` on a temp data root (`CAREER_OPS_ROOT`) for one success and one failure, and asserts that the JSON document is the **final** stdout output in both cases (nothing after the closing brace but whitespace). A short comment above the two `console.log` sites naming the rule would help the next person too.

## Acceptance

- The test is red if a `console.log('done')` is added after line 652, green on main as it is.
- No change to `set-status.mjs` behaviour.

Found by the web owner while reviewing #3602.

chshiv:
## What does this PR do?

This PR adds regression test cases for issue #3855 to verify the set-status.mjs JSON stdout contract.

The tests verify that when --json is used:

    The successful command returns a valid JSON document on stdout.
    The failed command returns a valid JSON error on stdout.
    The JSON document is the final content written to stdout.
    Whitespace, tabs, and newlines after the JSON are allowed.
    Any non-whitespace output after the JSON causes the test to fail.
    The success case updates applications.md and creates status-log.tsv.

These tests ensure that future output such as console.log("done") after the JSON result cannot be introduced without the regression test catching it.

## Related issue

Fixes #3855

## Type of change

Regression tests for existing behavior

Add Test cases for issue #3855

## Checklist

---

## Summary

Adds regression coverage for the `set-status.mjs --json` stdout contract.

## User impact

The JSON result must be the final stdout content. Extra output can make the web client report an error after a status update succeeds.

## Changes

: Adds success and failure integration tests in `tests/set-status-json-output.test.mjs`.
: Verifies JSON fields, exit codes, tracker updates, `status-log.tsv` creation, and trailing-output rules.
: Documents the contract near `set-status.mjs:238` and `set-status.mjs:652`.
: Preserves existing `set-status.mjs` behavior.

## System files touched

Not touched: `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/`.

### DIFF
--- a/set-status.mjs
+++ b/set-status.mjs
@@ -235,7 +235,9 @@ const failWith = makeCliFailWith(flags.json);
 function failUsage(message) {
   const msg = message ?? 'Expected 2 arguments: <report#|company> <state>';
   if (rawArgs.includes('--json')) {
+     // JSON must be the last thing written to stdout; machine callers parse stdout as one JSON document.
     console.log(JSON.stringify({ error: msg, code: 'usage' }));
+    // consol.log("end")  # simulate failure case
     console.error(`❌ ${msg}`);
   } else {
     if (message) console.error(`❌ ${message}\n`);
@@ -649,7 +651,11 @@ const result = {
 };
 
 if (flags.json) {
+  // JSON must be the last thing written to stdout; machine callers parse stdout as one JSON document.
   console.log(JSON.stringify(result, null, 2));
+  //console.log("done")  //simulate failure case
+  //console.log("      ")     whitespace test case
+  //console.log("      ")     new line test case
 } else {
   const verb = flags.dryRun ? 'would set' : changed ? 'set' : 'already';
   console.log(`✅ #${target.num} ${target.company} — ${target.role}: ${verb} ${oldStatus} → ${newStatus}${note ? ` (note: ${note})` : ''}`);
--- a/set-status.mjs
+++ b/set-status.mjs
@@ -237,7 +237,6 @@ function failUsage(message) {
   if (rawArgs.includes('--json')) {
      // JSON must be the last thing written to stdout; machine callers parse stdout as one JSON document.
     console.log(JSON.stringify({ error: msg, code: 'usage' }));
-    // consol.log("end")  # simulate failure case
     console.error(`❌ ${msg}`);
   } else {
     if (message) console.error(`❌ ${message}\n`);
@@ -653,9 +652,6 @@ const result = {
 if (flags.json) {
   // JSON must be the last thing written to stdout; machine callers parse stdout as one JSON document.
   console.log(JSON.stringify(result, null, 2));
-  //console.log("done")  //simulate failure case
-  //console.log("      ")     whitespace test case
-  //console.log("      ")     new line test case
 } else {
   const verb = flags.dryRun ? 'would set' : changed ? 'set' : 'already';
   console.log(`✅ #${target.num} ${target.company} — ${target.role}: ${verb} ${oldStatus} → ${newStatus}${note ? ` (note: ${note})` : ''}`);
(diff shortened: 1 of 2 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is in `set-status.mjs` (the canonical locked, validated, atomic tracker write path, core), the maintainer's call; I am reviewing the owned test `tests/set-status-json-output.test.mjs`.

It is sound. The test spawns the real set-status.mjs with `--json` over a temp tracker fixture, parses stdout as JSON, and asserts on real fields (changed true, the new status, and on the failure path an error with the invalid-state code). It also reads the tracker back to confirm the row actually flipped and that status-log.tsv was written, so it verifies the mutation, not just the printed shape. Root pinning is correct (mkdtempSync, explicit CAREER_OPS_ROOT and CAREER_OPS_TRACKER). Importantly the JSON path does not bypass the safety machinery: the failure case drives a rejected status through the real states.yml validation and confirms it still refuses and emits a JSON error, and the lock, atomic write, and report-link guard are untouched. I mutation-checked it: a trailing non-JSON line after the document reddens ("stdout is not one JSON document"), and a wrong field value reddens. A normal non-JSON invocation still works, so the flag is additive.

Two small hygiene items for the author, not soundness problems: the test leaves its temp dir behind (the rmSync cleanup is commented out), and the production file carries a couple of dead commented-out console.log lines plus a missing trailing newline.

Scott-Emberson (commented):
Re-verified after the new commit, which addresses two of the three hygiene items from my last note: the dead commented-out console.log lines are removed from set-status.mjs, and the test's temp-dir cleanup is restored (the rmSync is now active in the finally). The trailing newline on the test file is still missing, which is cosmetic and non-blocking.

The owned test remains sound. It still spawns the real set-status.mjs with --json over a temp tracker fixture, parses stdout as one JSON document, asserts real fields, and reads the tracker back to confirm the row actually mutated and that status-log.tsv was written; the failure case drives a rejected status through the real states.yml validation and asserts the invalid-state code, so the JSON path does not bypass validation, the lock, or the report-link guard. It passes 8 of 8 via the real harness. I re-ran the mutations: a trailing non-JSON line reddens, and a wrong field value reddens. A non-JSON invocation still writes correctly, so the flag stays additive.

Owned test sound; branch behind main. SIGNAL (set-status.mjs core), so not surfacing as ready-to-approve; routing and merge are the maintainer's.

santifer:
Thanks for this, and welcome — sorry for the wait.

The test does the useful thing rather than the easy one: it spawns the real CLI, reads the tracker back to confirm the row actually flipped, and drives a rejected status through the real validation on the failure path. I checked that it bites by adding a `console.log` after the JSON — it goes red with exactly the message you'd want, which is the regression the issue was about.

Two bits of whitespace I've taken care of on my side rather than send you another round: `set-status.mjs` had lost its trailing newline, and the new comment was indented one space off. I've pushed that to your branch, so the head moved — nothing for you to do, and the work is yours.

Merging once CI comes back green on it.

santifer (approved):

PR #4039, opened 2026-09-09
fix(web): stop /api/apply/drive closing its stream twice on success

### LINKED ISSUE #3966: web/apply: /api/apply/drive can close ReadableStream controller twice, causing ERR_INVALID_STATE / 500
### Code of Conduct

### Existing issues

### What happened?

When using the Web UI to apply for a job and letting the AI fill the real application form, the flow fails.

The UI shows:
Image

```text
The agent couldn't fill the form: Failed to fetch.
```
At the same time, /api/apply/drive returns 500 and the server logs:
```
TypeError: Invalid state: Controller is already closed
code: 'ERR_INVALID_STATE'

Error: failed to pipe response
POST /api/apply/drive 500
```

The stack trace points to `controller.close()` in `web/src/app/api/apply/drive/route.ts`.

That route also appears to have multiple paths that call controller.close(), including an unconditional call in finally, so the same stream controller can be closed more than once.

### Steps to reproduce

**Steps to reproduce**
```markdown
1. Open the Career Ops Web UI.
2. Go to the Apply flow.
3. Provide a real job application form URL.
4. Choose the option to let the AI fill the application form.
5. Wait for the agent to start processing/filling the form.
6. The UI reports `Failed to fetch`.
7. The server logs `ERR_INVALID_STATE: Controller is already closed` and `/api/apply/drive` returns 500.

### Expected behavior

The AI should fill the application form, or report the underlying form/fetch failure cleanly.

`/api/apply/drive` should not throw an additional `Controller is already closed` error or return a 500 while cleaning up the response stream.

### CLI tool

N/A — reproduced through the Web UI

### OS and Node.js version

_No response_

### Output of npm run doctor

```shell

```

maxmilian:
## What does this PR do?

Removes two `controller.close()` calls from the success paths inside `start(controller)`'s `try` block in `/api/apply/drive`, so the `finally` is the only place the stream is closed.

## Related issue

Closes #3966

## Type of change

- [x] Bug fix
- [ ] New feature

## The mechanism

`start(controller)` wraps the whole drive in `try` / `catch` / `finally` and closes the controller in the `finally`. Two paths inside the `try` closed it and then returned:

```ts
if (goal === "full") {
  …
  emit({ t: "done", filled: result.reached, … });
  controller.close();   // ← first close
  return;               // ← runs the finally, which closes again
}
```

A `return` runs the `finally` on its way out, so both of those paths closed twice. The second close throws `TypeError: Invalid state: Controller is already closed` (`ERR_INVALID_STATE`) out of an async `start`, which Next.js reports as `failed to pipe response` and a 500 — after the body has already been streamed, which is why the browser sees `Failed to fetch` rather than an error the UI can render.

**It is the two success paths that were wrong.** The error path at the bottom of the `try` returns through the `finally` alone and was always correct. That asymmetry is what makes @dbaikov's report specific: it only appears once the agent actually reaches and fills a form (`goal === "full"`, and the `reached` + `finalizeDrivenSession` path), not when the drive fails.

I checked whether this is a pattern worth changing more broadly: the sibling route `apply/prefill/route.ts` has no `finally` and returns on every path, so it is not affected and nothing else needs touching. The fix stays at two deleted lines.

Nothing is lost by deferring to the `finally` — `emit()` has already written the `done` frame before the close in both paths.

## Tests

`web/tests/lib/drive-stream-close.test.mjs` (new).

`route.ts` is TypeScript inside a Next.js route and the `start` closure cannot be imported and exercised, so this reads the source and asserts the shape — the same approach as the existing `tests/lib/core-writer-await.test.mjs`, and with the same self-checking discipline: a `self-check` test asserts every anchor it relies on (`new ReadableStream`, `async start(controller)`, at least one `controller.close(`) still exists, so a rename fails loudly instead of matching nothing and passing.

Three tests:

1. **self-check** — the anchors exist.
2. **the real assertion** — `controller.close()` appears exactly once in the `finally`, and **zero** times before it. This is the one that catches #3966.
3. **the other way a fix could go wrong** — the two success paths still `return` early. Dropping the `return`s along with the closes would fall through into the "didn't reach a form" branch and emit an `error` frame after a `done`.

The `finally` block is located by matching `} finally {` rather than by counting braces: the body is full of nested braces and object literals, and a counter that drifted would silently redefine what "inside the try" means.

**Verification:**

- New test on the fix: **3 passed**.
- Red control — reverting only `route.ts` to `main` and re-running: **fails** with
  `controller.close() is called 2 time(s) before the finally; each one double-closes, because the return that follows runs the finally too.`
- All of `web/tests/lib/*.test.mjs`: **500 passed, 0 failed**.
- `node test-all.mjs --quick`: **8687 passed, 0 failed** (16 warnings, all pre-existing — the personal-data heuristic firing on the owner's own name and `cv-sync-check.mjs` needing user data).

**What I did not do:** I have not run the Next.js UI through a real apply flow against a live posting — that needs a browser session and a real job form. The 500 is reproduced and fixed at the stream level, and the UI recovery follows from it, but I am stating that as inference rather than something I watched happen.

## Checklist

https://claude.ai/code/session_01WgtWMywGTyu6twnk1Bu9nP

## Summary

`/api/apply/drive` no longer closes its `ReadableStream` controller twice. The route now closes the controller only in `finally` at `web/src/app/api/apply/drive/route.ts`. This prevents `ERR_INVALID_STATE` and related 500 responses during the application flow.

A source-shape test verifies the single close and preserves both early-return paths in `web/tests/lib/drive-stream-close.test.mjs`.

User-visible result: AI form filling can complete or report its underlying failure without an additional `Controller is already closed` error.

## Files changed

- `web/src/app/api/apply/drive/route.ts`
- `web/tests/lib/drive-stream-close.test.mjs`

### DIFF
--- a/web/src/app/api/apply/drive/route.ts
+++ b/web/src/app/api/apply/drive/route.ts
@@ -47,7 +47,6 @@ export async function POST(req: Request) {
           // to review + submit themselves. We never submit.
           if (result.reached) await handoffSession(s.id).catch(() => {});
           emit({ t: "done", filled: result.reached, turns: result.turns, reason: result.reason });
-          controller.close();
           return;
         }
 
@@ -55,7 +54,6 @@ export async function POST(req: Request) {
           const fin = await finalizeDrivenSession(s.id, cliId);
           if (fin) {
             emit({ t: "done", reached: true, turns: result.turns, title: fin.title, fields: fin.fields, issues: fin.issues });
-            controller.close();
             return;
           }
         }
--- a/web/tests/lib/drive-stream-close.test.mjs
+++ b/web/tests/lib/drive-stream-close.test.mjs
@@ -1,0 +1,98 @@
+// Guards the ReadableStream in src/app/api/apply/drive/route.ts (issue #3966).
+//
+// The stream's `start(controller)` wraps the whole drive in try/catch/finally
+// and closes the controller in the `finally`. Two success paths inside the
+// `try` also closed it and then returned -- and a `return` runs the `finally`,
+// so those paths closed twice. The second close throws
+// `TypeError: Invalid state: Controller is already closed` (ERR_INVALID_STATE)
+// out of an async `start`, which Next.js surfaces as `failed to pipe response`
+// and a 500 after the body has already been streamed.
+//
+// It was the two SUCCESS paths that were wrong -- the error path returned
+// through the `finally` alone and was fine -- which is why it only showed up
+// once the agent actually reached and filled a form.
+//
+// route.ts is TypeScript inside a Next.js route and its `start` closure cannot
+// be imported and exercised here, so this reads the source and asserts the
+// shape instead. Same approach, and the same self-checking discipline, as
+// tests/lib/core-writer-await.test.mjs: every anchor this test relies on is
+// asserted to exist first, so a rename or a refactor fails loudly instead of
+// matching nothing and passing.
+//
+// Run (from web/, as `npm test` does):  node --test tests/lib/drive-stream-close.test.mjs
+// From the repo root:                   node --test web/tests/lib/drive-stream-close.test.mjs
+
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { dirname, join } from "node:path";
+
+const SRC = join(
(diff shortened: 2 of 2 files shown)


santifer:
Hi @maxmilian, this PR has waited 14 days without a reply from a maintainer. That is on us, not on you, and we are sorry for it.

PR #4045, opened 2026-09-09
fix(analyze-patterns): validate threshold and vendor flags

### LINKED ISSUE #3113: fix(analyze-patterns): --min-threshold and --min-vendor-n values are never validated

`analyze-patterns.mjs` reads its flags positionally (`args.indexOf('--min-threshold')`) with no validation at all: a mistyped flag is ignored, and `--min-threshold abc` silently becomes the default — the same silent-wrong-answer family #2982 closed for four other CLIs. Mirror that work: `validateFlags` for the names, `requireFlagValue` with `isNonNegativeInt`/`isPositiveInt` (both already exported from `lib/cli-flags.mjs`) for the values, plus a row in `tests/cli-flag-validation.test.mjs`.

**Acceptance:** unknown flags and unusable values exit 2 naming the flag; valid invocations behave exactly as before. (Note: #2983 touches lib/cli-flags.mjs and merges shortly — rebase over it if you collide.)

Syed-Haseeb97:
Closes #3113

## Summary

* validate --min-threshold and --min-vendor-n as strict positive integers
* reject invalid and unsafe integer values with exit status 2
* preserve the existing analyze-patterns execution and self-test behavior

## Validation

* node analyze-patterns.mjs --self-test
* node test-all.mjs --only cli-flag-validation

## Summary

`analyze-patterns.mjs:83-91` now validates `--min-threshold` and `--min-vendor-n` before execution.

Users can provide positive integers in separated or equals-sign forms. Missing, duplicate, zero, negative, fractional, non-numeric, or unsafe values identify the flag and exit with status `2`.

Valid invocations and self-tests keep their existing behavior. Invalid flags now fail instead of being ignored or replaced with defaults.

## Files changed

: `analyze-patterns.mjs:70-91,1718`
: `tests/cli-flag-validation.test.mjs:155-209`

Tests cover invalid values, missing operands, duplicate flags, and unsafe integers.

`AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/` are present but unchanged.

### DIFF
--- a/analyze-patterns.mjs
+++ b/analyze-patterns.mjs
@@ -1678,6 +1678,37 @@ if (isMainModule(import.meta.url)) {
     requireOperand: true,
   });
 
+  // Strict validation for numeric value flags. Errors exit with code 2 to
+  // match recent conventions. Presence is detected via flagValue so both
+  // `--flag value` and `--flag=value` are accepted.
+  const rawMinThreshold = flagValue(args, '--min-threshold');
+  if (rawMinThreshold !== undefined) {
+    const text = String(rawMinThreshold).trim();
+    if (!/^\d+$/.test(text)) {
+      console.error(`Error: --min-threshold requires a non-negative integer, got "${rawMinThreshold}"`);
+      process.exit(2);
+    }
+    const parsed = Number(text);
+    if (!Number.isSafeInteger(parsed) || parsed < 0) {
+      console.error(`Error: --min-threshold requires a non-negative integer, got "${rawMinThreshold}"`);
+      process.exit(2);
+    }
+  }
+
+  const rawMinVendor = flagValue(args, '--min-vendor-n');
+  if (rawMinVendor !== undefined) {
+    const text = String(rawMinVendor).trim();
+    if (!/^\d+$/.test(text)) {
+      console.error(`Error: --min-vendor-n requires a positive integer, got "${rawMinVendor}"`);
+      process.exit(2);
+    }
+    const parsed = Number(text);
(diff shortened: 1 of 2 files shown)


Scott-Emberson (commented):
Verified the owned test by execution and mutation. It is sound, and the substance file is real this time (`analyze-patterns.mjs` is the full ~82 KB script, not a stub).

`tests/cli-flag-validation.test.mjs` spawns the real `analyze-patterns.mjs` via `spawnSync` and asserts the range and format guards: `--min-threshold abc` / `-5` / `1.5` / `9007199254740992` exit 2 with "requires a non-negative integer", and `--min-vendor-n 0` / `-1` exit 2 with "requires a positive integer" (the 2^53 case is correctly caught by the safe-integer guard). 38 tests pass. Deleting the numeric-validation block in `analyze-patterns.mjs` drops those to exit 1 and reddens the suite, so the guard is genuinely enforced. The change to the shared file is additive: it inserts the nine analyze-patterns subtests without touching the existing `SCRIPTS` array, `runScript`, or any prior assertion.

One thing worth adding while you are here: the new tests cover the invalid-value path (exit 2) but not the missing-operand path (a bare `--min-threshold` with no value). That behavior does exist through the shared `validateFlags(requireOperand: true)` and returns exit 1 with "requires a value", but mutating `requireOperand` to false does not redden this suite, so it is currently unguarded here. One added case would pin it. The exit-2-for-bad-values choice itself is deliberate and self-consistent with the code, so no issue there.

Sound on the test axis. `analyze-patterns.mjs` is the maintainer's call on the substance.

PR #4046, opened 2026-09-09
fix(verifiers): resolve provider plugins like the scanner (#4026)

### LINKED ISSUE #4026: verify-pipeline.mjs and verify-portals.mjs miss plugin providers — bundled apify reported as "unknown provider", cascades into 8 test-all failures
## Summary

`verify-pipeline.mjs` and `verify-portals.mjs` build their provider map with `loadProviders()` alone, while `scan.mjs` builds its with `loadProviders()` **and** `mergeProviderPlugins()`. As a result the health checkers cannot see plugin providers, and report every `provider: <plugin-id>` entry as an unknown provider that "never scans" — while the scanner resolves and scans it perfectly.

This affects the **bundled `apify` plugin**, so it does not require any third-party code to reproduce.

## Reproduction

On v1.32.0, add a bundled-plugin provider entry to `portals.yml`:

```yaml
tracked_companies:
  - name: "apify repro"
    provider: apify
    actor: misceres/indeed-scraper
    enabled: true
```

Then:

```console
$ node validate-portals.mjs
0 errors, 0 warnings

$ node verify-pipeline.mjs
❌ portals.yml: "apify repro" sets an unknown provider — unknown provider: apify.
   The entry never scans (see providers/ for valid ids)
📊 Pipeline Health: 4 errors, 0 warnings
```

The claim "The entry never scans" is false — with the plugin enabled and its env present, `node scan.mjs --dry-run --company "apify repro"` resolves the provider and returns postings.

## Root cause

| File | Line | Builds provider map with |
|---|---|---|
| `scan.mjs` | 2869 | `loadProviders()` **+ `mergeProviderPlugins()`** |
| `verify-pipeline.mjs` | 543 | `loadProviders()` only |
| `verify-portals.mjs` | 726 | `loadProviders()` only |

`mergeProviderPlugins()` (`plugins/_engine.mjs`) is what folds enabled plugin providers into the map. Without it, `resolveProvider()` sees only `providers/*.mjs` and rejects every plugin id.

`validate-portals.mjs` passes because it does not resolve providers at all — so the two validators disagree with each other as well as with the scanner.

## Impact

Beyond the false error, `verify-pipeline.mjs` exits non-zero, which cascades into `test-all.mjs`: four fixtures that shell out to it fail as collateral.

Enabling one plugin provider took a local `test-all.mjs` run from 23 failures to 31 — 8 extra, all from this one defect:

```
❌ verify-pipeline.mjs crashed (exit 1)
❌ verify-pipeline crashed on duplicate/orphan report fixture
❌ verify-pipeline crashed on the Turkish dotted-I fixture
❌ verify-pipeline crashed on non-Latin grouping fixture (#2393)
❌ verify-pipeline crashed on the orphan-reference fixture
```

So using a documented, supported feature makes the health check and the test suite red, with no way to tell the real failures from this artefact.

## Suggested fix

Have both verifiers mirror `scan.mjs`:

```js
const providers = await loadProviders(PROVIDERS_DIR);
await mergeProviderPlugins(providers, { root: CAREER_OPS });
```

`mergeProviderPlugins()` already returns early when `config/plugins.yml` is absent, so a plugin-free install is byte-identical to today. Its `inactiveProviderStub()` path also means an enabled-but-unconfigured plugin would be reported with an actionable message ("missing env APIFY_TOKEN") instead of "unknown provider".

Worth a regression test asserting the scanner and the verifiers agree on provider resolution — that divergence is the actual bug, and it can reappear anywhere a third caller loads providers.

## Environment

- career-ops v1.32.0
- Node v22.14.0, macOS (darwin 25.6.0)
- Reproduced with the bundled `apify` plugin and with a local `plugins.local/` provider

Zoubeir23:
Fixes #4026.

## The bug

`verify-pipeline.mjs` and `verify-portals.mjs` build their provider map with `loadProviders()` alone. `scan.mjs` also calls `mergeProviderPlugins()` (`scan.mjs:2869`). So the health checkers can't see plugin providers, and report every `provider: <plugin-id>` portals entry as an **unknown provider** that *"never scans"* — while the scanner resolves and scans it.

This hits the **bundled `apify` plugin**, so it needs no third-party code. And `verify-pipeline.mjs` exits non-zero on the false error, which cascades into `test-all.mjs`: the reporter measured 8 extra failures, all from this one defect, making real failures indistinguishable from the artefact.

## Verified before/after

`config/plugins.yml` enabling `apify`, `portals.yml` with `provider: apify`, `APIFY_TOKEN` unset:

| | before | after |
|---|---|---|
| `verify-pipeline.mjs` | `❌ ... unknown provider: apify. The entry never scans` → `1 errors` → exit 1 | `0 errors` → exit 0 |
| `verify-portals.mjs` | `unknown provider` | `❌ apify — plugin "apify" inactive: missing env APIFY_TOKEN — add to .env` (actionable) |

## The fix

Both verifiers now mirror `scan.mjs`: `loadProviders()` from `CODE_ROOT`, then `await mergeProviderPlugins(providers, { root: CODE_ROOT })`.

- `mergeProviderPlugins()` returns before any work (no discovery, no dotenv, no `process.env` read) when `config/plugins.yml` is absent, so a **plugin-free install is byte-identical**.
- An enabled-but-unconfigured plugin resolves to `inactiveProviderStub()` — an actionable *"missing env APIFY_TOKEN"* message rather than *"unknown provider"*.
- `verify-pipeline.mjs` also had `loadProviders(join(CAREER_OPS, 'providers'))` — the **data** root; `providers/` ships in the code layer, so it now uses `CODE_ROOT` to match `scan.mjs` (identical on a non-split install, correct on a split one).

## Tests — `tests/verifiers-plugin-providers.test.mjs`

1. **Structural:** `scan.mjs`, `verify-pipeline.mjs` and `verify-portals.mjs` all call `mergeProviderPlugins()` — the reporter's point that "the divergence is the bug, and it can reappear anywhere a third caller loads providers".
2. **Behavioral:** spawns the real `verify-pipeline.mjs` against an enabled `apify` portals entry and asserts stdout carries no `unknown provider: apify`. Writes `config/plugins.yml` into the checkout (gitignored, removed in `finally`) and **skips** if one already exists rather than clobbering a real user's config.

Both assertions redden on a straight revert of the fix.

## Test plan

- [x] `node tests/verifiers-plugin-providers.test.mjs` — 2/2
- [x] `node verify-pipeline.mjs` / `node verify-portals.mjs` on a plugin-free checkout — unchanged (0 errors)
- [x] `node test-all.mjs` — 8689 passed, 0 failed (1 pre-existing local flake: `update-system.mjs check` SIGTERM'd on this machine's slow link to GitHub — untouched by this PR)

## Summary

Users can configure bundled plugin providers, such as `apify`, without false `unknown provider` or `never scans` errors.

Changes:

- `verify-pipeline.mjs:551-552`: Loads providers from `CODE_ROOT/providers` and merges enabled plugins.
- `verify-portals.mjs:764-767`: Merges plugins before provider validation.
- `tests/verifiers-plugin-providers.test.mjs:61-141`: Adds structural and behavioral regression coverage.
- Tests isolate `CAREER_OPS_ROOT` and remove inherited `CAREER_OPS_REPORTS`.
- Inactive plugins report actionable errors, such as `missing env APIFY_TOKEN`.
- Plugin-free installations preserve existing behavior because plugin merging returns early without `config/plugins.yml`.

Validation: `test-all.mjs` reported 8692 passed and 0 failed.

System files not changed: `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/`.

### DIFF
--- a/tests/verifiers-plugin-providers.test.mjs
+++ b/tests/verifiers-plugin-providers.test.mjs
@@ -1,0 +1,74 @@
+// tests/verifiers-plugin-providers.test.mjs — the scanner and both health
+// checkers must resolve providers through the SAME path (#4026).
+//
+// scan.mjs folds enabled provider plugins into its map with
+// mergeProviderPlugins(). verify-pipeline.mjs and verify-portals.mjs did not,
+// so a supported `provider: <plugin-id>` portals entry was reported as an
+// "unknown provider" that "never scans" — while the scanner scanned it. The
+// divergence is the bug and it can reappear the moment a caller drifts, so it
+// is pinned structurally (three callers must agree) AND behaviorally (an
+// enabled stub plugin resolves in verify-pipeline's own load path).
+import { readFileSync, existsSync, mkdtempSync, mkdirSync, writeFileSync, rmSync, unlinkSync } from 'fs';
+import { execFileSync } from 'child_process';
+import { tmpdir } from 'os';
+import { join, dirname } from 'path';
+import { fileURLToPath } from 'url';
+import { pass, fail, warn, NODE } from './helpers.mjs';
+
+const ROOT = join(dirname(fileURLToPath(import.meta.url)), '..');
+
+console.log('\nverifiers resolve provider plugins like the scanner (#4026)');
+
+// ── 1. Structural: all three provider-map builders call mergeProviderPlugins ──
+{
+  const callers = ['scan.mjs', 'verify-pipeline.mjs', 'verify-portals.mjs'];
+  const missing = callers.filter((f) => !/mergeProviderPlugins\s*\(/.test(readFileSync(join(ROOT, f), 'utf8')));
+  if (missing.length === 0) {
+    pass('scan.mjs, verify-pipeline.mjs and verify-portals.mjs all call mergeProviderPlugins()');
+  } else {
+    fail(`these provider-map builders skip mergeProviderPlugins() and will disagree with the scanner: ${missing.join(', ')}`);
+  }
+}
+
+// ── 2. Behavioral: run the real verify-pipeline.mjs against a portals entry
+//    using the bundled `apify` plugin, enabled. Before the fix it printed
+//    "unknown provider: apify" and exited 1; after, apify resolves (to an
+//    actionable "missing env APIFY_TOKEN" stub here) and the false error is
+//    gone. verify-pipeline resolves plugins from its own CODE_ROOT, so
+//    config/plugins.yml has to live in the real checkout — skip rather than
+//    clobber a real user's config. ──
+{
+  const pluginsConfig = join(ROOT, 'config', 'plugins.yml');
+  if (existsSync(pluginsConfig)) {
+    warn('config/plugins.yml already exists — skipping the spawn test rather than overwriting it');
+  } else {
+    const tmp = mkdtempSync(join(tmpdir(), 'co-4026-'));
+    try {
+      writeFileSync(pluginsConfig, 'plugins:\n  apify: { enabled: true }\n');
+      const portals = join(tmp, 'portals.yml');
+      writeFileSync(portals,
+        'tracked_companies:\n  - name: "apify 4026"\n    provider: apify\n    actor: x/y\n    enabled: true\n');
+
+      let out = '';
+      let exitCode = 0;
+      try {
+        out = execFileSync(NODE, [join(ROOT, 'verify-pipeline.mjs')], {
+          cwd: ROOT, encoding: 'utf8', timeout: 60000, stdio: ['pipe', 'pipe', 'pipe'],
+          env: { ...process.env, CAREER_OPS_PORTALS: portals, APIFY_TOKEN: '' },
+        });
+      } catch (e) {
+        out = `${e.stdout || ''}${e.stderr || ''}`;
+        exitCode = e.status ?? 1;
+      }
+
+      if (!/unknown provider:\s*apify/i.test(out)) {
+        pass('verify-pipeline.mjs resolves an enabled `apify` plugin provider instead of "unknown provider"');
+      } else {
+        fail(`verify-pipeline.mjs still reports the enabled apify plugin as an unknown provider (exit ${exitCode})`);
+      }
+    } finally {
+      if (existsSync(pluginsConfig)) unlinkSync(pluginsConfig);
+      rmSync(tmp, { recursive: true, force: true });
+    }
+  }
+}
--- a/tests/verifiers-plugin-providers.test.mjs
+++ b/tests/verifiers-plugin-providers.test.mjs
@@ -6,69 +6,130 @@
 // so a supported `provider: <plugin-id>` portals entry was reported as an
 // "unknown provider" that "never scans" — while the scanner scanned it. The
 // divergence is the bug and it can reappear the moment a caller drifts, so it
-// is pinned structurally (three callers must agree) AND behaviorally (an
-// enabled stub plugin resolves in verify-pipeline's own load path).
-import { readFileSync, existsSync, mkdtempSync, mkdirSync, writeFileSync, rmSync, unlinkSync } from 'fs';
+// is pinned structurally (three callers must run the call) AND behaviorally
+// (the real verifiers resolve an enabled plugin instead of flagging it).
+import {
+  readFileSync, existsSync, mkdtempSync, mkdirSync, writeFileSync, rmSync, cpSync,
+  symlinkSync, readdirSync,
+} from 'fs';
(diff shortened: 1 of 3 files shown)


Scott-Emberson (changes requested):
The fix reads correctly: `mergeProviderPlugins(providersMap, { root })` is consumed the same way `scan.mjs` already does it, and all three callers pass `root: CODE_ROOT`, so an enabled provider-hook plugin resolves instead of being flagged. My concern is with the owned test, which currently passes even when that fix is reverted.

The behavioral case spawns `verify-pipeline.mjs`, but `verify-pipeline.mjs` exits 0 near line 85 when `data/applications.md` is absent, before it reaches the provider-resolution check the test is meant to exercise. `data/applications.md` is user-layer and gitignored, so it is absent in a clean checkout and in CI, and the test never creates it. I measured it through the authoritative runner:

1. Fix reverted (merge call removed) on a clean checkout, matching CI: `node test-all.mjs --only verifiers-plugin-providers` reports 2 passed, 0 failed, green. The assertion is vacuous because verify-pipeline exited before the code under test ran.
2. Same revert, but with a `data/applications.md` present: 1 passed, 1 failed, red ("still reports the enabled apify plugin as an unknown provider").

So the test only reddens on the real regression when an untracked user-layer file happens to be present, which it is not where it counts. There is a second, smaller version of the same problem: the structural check matches `mergeProviderPlugins(` as a bare token, and a code comment in verify-pipeline contains that string, so a call-only revert leaves the structural assertion green too (verify-portals is fine, its added comments do not contain the call form).

Two things to make it bite, both contained to the test:

**1.** Seed a minimal `data/applications.md` in a temp root (a throwaway `CAREER_OPS` / cwd, cleaned up after) so the provider-resolution check is actually reached before the spawn asserts. Without it the behavioral case cannot fail.

**2.** Add the negative direction: a genuinely unknown provider id should still be flagged. The code already behaves correctly here (only enabled provider-hook plugins merge), but nothing asserts it, so an accept-everything regression would pass. While you are at it, assert the real call rather than the token, or exclude comments, so the structural check does not pass on a commented-out mention.

The fix is right; it just needs the test to fail when the fix is gone. Once the behavioral case reaches Check 15 and the negative case is in, this holds.

Scott-Emberson (commented):
Re-checked after the new commit, which resolves the test-axis concern I raised. Previously the owned test passed even with the fix reverted, because on a clean checkout (and in CI) `data/applications.md` is absent, so verify-pipeline.mjs exited early and never reached the provider-resolution check the test was meant to exercise.

That is fixed. The behavioral case now runs the verifiers from an isolated copied code-root and seeds a minimal tracker via CAREER_OPS_TRACKER, so verify-pipeline gets past its no-tracker early exit and actually reaches Check 15. I confirmed it with the decisive mutation on a clean checkout with no `data/applications.md` present: reverting the real fix (removing the `mergeProviderPlugins` call from verify-pipeline.mjs) now reddens both the structural and the behavioral assertions, where before it stayed green. The other two points landed too: there is a negative case asserting a genuinely unknown provider is still flagged, and the structural check now strips comments and matches the call form so a commented-out mention no longer satisfies it.

The owned test now genuinely gates the fix. The substance in verify-pipeline.mjs and verify-portals.mjs is core, so routing and merge are the maintainer's; the branch is behind main.

santifer:
Sorry for the wait on this one, @Zoubeir23, and thanks for the before/after table: it made the review short.

The diagnosis holds and the scope is the right one. `scan.mjs` was the only provider-map builder folding plugins in, so both verifiers disagreed with the thing they exist to model. Resolving from `CODE_ROOT` rather than the data root is also the correct half of this: `providers/` and `plugins/` ship in the code layer, and #4240 landed the same correction for `audit-portals.mjs` earlier today.

The test is what makes this stick. I ran it against current `main` with only the two verifiers reverted and three of its four rows go red, so it bites rather than passing vacuously; the structural row is well paired with the one asserting a genuinely unknown id still gets flagged.

On a checkout of your branch over current `main`: 9061 passed, 0 failed, and the `--help` test that landed today stays green alongside your change.

One thing you should know: #3963 reported the `verify-pipeline` half of this on 7 Sep. Yours covers both verifiers, picks the right root and brings the regression test, so yours is the one landing, and I'll say so over there.

The review is favorable. The branch is behind `main`, so refreshing it and the merge are on me.

santifer (approved):
Both verifiers now resolve providers the way `scan.mjs` does, from the code root, and the regression test reddens on a revert. Verified on a checkout over current main: 9061 passed, 0 failed.

PR #4049, opened 2026-09-09
feat(pdf): add optional `context` line to CV experience entries

lesiashu:
## What

Adds an optional `context` field to CV experience entries. It renders as a short, **un-bulleted italic line directly under the role** — for background that isn't an achievement:

- company size / stage ("Early-stage startup; joined at growth stage.")
- an acquisition ("Product reached 500k MAU and was acquired by BigCo.")
- headcount owned, or the team the role sat in

## Why

Today that information has two bad homes: a **bullet** (where it competes with real achievements and dilutes the six-second scan), or **nowhere** (dropped in tailoring). Many source CVs already write a one-line company blurb under the title; the generated CV had no way to keep that shape.

## Changes

| File | Change |
|---|---|
| `lib/cv-payload-schema.mjs` | `context` added to the `html` experience `known` keys (optional). Unknown-key warning no longer fires for it. |
| `build-cv-html.mjs` | Rendered in **both** the partial and the fallback builder as `<div class="job-context">`; conditional — emitted only when present. Self-test extended (present → rendered, absent → not). |
| `templates/sections/experience.html` | `{{CONTEXT_BLOCK}}` conditional, mirroring the existing `{{LOCATION_BLOCK}}`. |
| `templates/cv-template*.html` (all 7 variants) | `.job-context` style — italic, muted, sits under the role — and added to the page-break-avoid selector list so it stays attached to its bullets. |
| `modes/pdf.md` | JSON schema example + field-reference table row. |

## Compatibility

Fully backward compatible. A payload with no `context` renders byte-identical to before — the `CONTEXT_BLOCK` conditional and the fallback `e.context ? … : ''` both collapse to nothing.

## Testing

- `node build-cv-html.mjs --test` passes (extended with present/absent assertions for the new block).
- Rendered against all 7 templates; `verify-cv-facts.mjs` unaffected.

Happy to open a tracking issue first if that's preferred per CONTRIBUTING — the change is small and additive so I went straight to a PR.

## Summary

Users can add an optional `context` field to experience entries. The field renders as a short, italic, muted, unbulleted line below the role.

Use `context` for company stage, acquisitions, headcount, or team context. Keep achievements in `bullets`: `modes/pdf.md:167,209`.

Entries without `context` render unchanged. Both HTML paths support the field: `build-cv-html.mjs:359-389` and `templates/sections/experience.html:22,28`.

## User impact

- The HTML schema accepts `context`: `lib/cv-payload-schema.mjs:24`.
- All seven CV templates style `.job-context` and apply page-break handling: `templates/cv-template*.html`.
- Self-tests cover present and absent values: `build-cv-html.mjs:975-1004`.

## System files

`AGENTS.md`, `DATA_CONTRACT.md`, `update-system.mjs`, `providers/`, and `.github/` are not changed.

### DIFF
--- a/build-cv-html.mjs
+++ b/build-cv-html.mjs
@@ -356,12 +356,15 @@ function buildExperience(entries, partial) {
       const location = e.location
         ? `\n    <div class="job-location">${escapeHtml(e.location)}</div>`
         : '';
+      const context = e.context
+        ? `\n    <div class="job-context">${escapeHtml(e.context)}</div>`
+        : '';
       return `<div class="job">
     <div class="job-header">
       <span class="job-company">${escapeHtml(e.company)}</span>
       <span class="job-period">${escapeHtml(e.dates || e.period || '')}</span>
     </div>
-    <div class="job-role">${escapeHtml(e.role)}</div>${location}
+    <div class="job-role">${escapeHtml(e.role)}</div>${location}${context}
     <ul>
 ${bullets}
     </ul>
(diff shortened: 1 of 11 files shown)


santifer:
Hi @lesiashu, this PR has waited 13 days without a reply from a maintainer. That is on us, not on you, and we are sorry for it.

santifer (changes requested):
Thanks @lesiashu for **an optional `context` line under each CV experience entry**. To get this landed we need a few concrete things; this is the complete list, not a first round:

- [ ] In `modes/pdf.md`, say that `context` only uses facts already in `cv.md`, and keep the example to company facts (stage, acquisition) rather than an achievement like "500k MAU": the agent must never invent a line in someone's CV
- [ ] Resolve or answer the 3 open CodeRabbit threads
- [ ] Check that every CV template picks up the `context` style: this PR styles seven, and the similar #3776 touched eight

The self-test that checks the line shows up when `context` is set and stays out when it is not is exactly the right guard. Welcome: there is no clock, and the PR stays yours.

PR #4064, opened 2026-09-10
fix(web): resolve the data root exactly as the core does

jaybarrett34:
## Problem

`web/`'s `careerOpsRoot()` honors only `CAREER_OPS_ROOT`, then falls back to the parent directory:

```ts
const env = process.env.CAREER_OPS_ROOT?.trim();
if (env) return env;
return path.resolve(process.cwd(), "..");
```

The core's `path-resolver.mjs` honors `CAREER_OPS_ROOT` -> `CAREER_OPS_DATA_DIR` -> a `.career-ops-data` marker -> its own directory, which `AGENTS.md` documents under **Path Resolution Override & Precedence**.

A user with a marker file has the CLI and the web app reading **two different data roots**, and nothing reports the disagreement.

## The subtler half: the resolution base

`path-resolver.mjs` resolves relative values against `__dirname` — the checkout. `web/` runs with `cwd = <root>/web`.

A marker containing `../shared-data` therefore means:

| | resolves to |
|---|---|
| CLI | `<root>/../shared-data` |
| web | `<root>/shared-data` |

Same string, two directories, no error. `resolveDataRoot()` takes `coreRoot` as an explicit argument rather than re-deriving it.

## Why not just import the core resolver

It would be DRYer, and it would be silently wrong. `path-resolver.mjs` derives `__dirname` from `import.meta.url`; bundled into the Next server output, that resolves relative values against `.next/server/...`.

Instead, `tests/lib/data-root.test.mjs` includes an **agreement test** that loads the core's real `path-resolver.mjs` from a fixture and compares both implementations across ten branches — the pattern that file's own header prescribes ("test fixtures copy this file standalone").

## Compatibility

Absolute env values and marker contents are returned unchanged by `path.resolve`, so **every existing install keeps its current root**. The behavior change is limited to relative values and to `CAREER_OPS_DATA_DIR` / marker files, which previously did nothing in `web/`.

## Verification

11 new tests. Negative control — reintroducing either bug fails the suite:

| injected bug | result |
|---|---|
| ignore `CAREER_OPS_DATA_DIR` | 2 failures, agreement test catches it |
| web-relative resolution base | 3 failures, agreement test catches it |
| (fixed) | 11 pass |

Note for reviewers: this suite imports only `.mjs`, so it runs on Node 22 without extra flags. Two neighbouring suites do not — that is a separate PR.

## Summary

Web data-root resolution now matches the core resolver.

From the user’s point of view:

: `CAREER_OPS_ROOT` takes priority.
: `CAREER_OPS_DATA_DIR` is used when `CAREER_OPS_ROOT` is not set.
: `.career-ops-data` is used when neither variable is set.
: The checkout directory is the final fallback.
: Relative paths resolve from the checkout directory.
: Absolute paths keep their existing roots.

The web app uses a local resolver because bundling the core resolver would change its `__dirname` base: `web/src/lib/career-ops.ts:12`.

The resolver is defined in `web/src/lib/core/data-root.mjs:1`. Tests cover precedence, marker files, relative paths, and agreement with the core resolver: `web/tests/lib/data-root.test.mjs:1`.

## Touched system files

`update-system.mjs` is included in the changed-file list.

No changes are reported for `AGENTS.md`, `modes/`, `DATA_CONTRACT.md`, `providers/`, or `.github/`.

### DIFF
--- a/web/src/lib/career-ops.ts
+++ b/web/src/lib/career-ops.ts
@@ -2,6 +2,7 @@ import fs from "node:fs";
 import path from "node:path";
 import * as yaml from "js-yaml";
 import { atomicWrite } from "@/lib/core/safe-write";
+import { resolveDataRoot } from "@/lib/core/data-root.mjs";
 import { parseApplications } from "@/lib/tracker-table.mjs";
 // One definition of the `{n}-RESERVED.md` convention, shared with
 // run-cli-support.mjs — see report-files.mjs for why it lives there.
@@ -20,9 +21,24 @@ import { pdfIndexEntryForReport } from "@/lib/apply/cv-selection.mjs";
  * checkout — see web/.env.local.
  */
 export function careerOpsRoot(): string {
-  const env = process.env.CAREER_OPS_ROOT?.trim();
-  if (env) return env;
-  return path.resolve(process.cwd(), "..");
+  // `process.cwd()` is `<core>/web` for `next dev`/`next start`, so its parent is
+  // the core checkout — the same directory `path-resolver.mjs` calls `__dirname`.
+  // resolveDataRoot() needs it explicitly because relative env values and marker
+  // contents resolve against it; see data-root.mjs for why that base matters.
+  const coreRoot = path.resolve(process.cwd(), "..");
+  return resolveDataRoot(
+    coreRoot,
+    (p) => {
+      try {
+        return fs.readFileSync(p, "utf8");
+      } catch {
+        return null; // absent, unreadable, or a directory — all mean "no marker"
+      }
+    },
+    process.env,
+    path.resolve,
+    path.join,
+  );
 }
 
 /**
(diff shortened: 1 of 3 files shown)


santifer:
Welcome to career-ops, @jaybarrett34, and thank you — this is a better piece of work than the title suggests. The bug you found is the quiet kind: a user with `CAREER_OPS_DATA_DIR` set, or a `.career-ops-data` marker, had the CLI and the web app reading two different data roots with nothing reporting the disagreement.

The half I want to name, because it is the part a reviewer skims past, is the resolution base rather than the precedence order. `path-resolver.mjs` resolves relative values against the core checkout; `web/` runs with its cwd one directory deeper, so the same marker string meant two different directories. Your comment explaining why you did not simply import the core module — `import.meta.url` would point at the Next bundle — is the note that stops someone "simplifying" this back into a bug later.

I have refreshed your branch onto current main. The one red job, `test (windows-latest)`, ran on a base seven days old, with ubuntu and macOS green on the same commit; its annotations only report exit code 1 without naming a test, so the fresh run will settle it. Locally on current main: your suite 11 passed, `test-all.mjs --quick` 9043 passed, web 582 passed, typecheck clean — 0 failures anywhere. Next step is mine.

santifer (approved):
Approving as maintainer. The web resolver now matches the core on both precedence and resolution base, the agreement test pins the two implementations together, and the full local run on current main is clean across the core suite, the web suite and typecheck.

Issue #3976, opened 2026-09-07
test(harness): a contained throw reports one failure, not the assertions it silently cost


artemtrofymenko:
### What happens

A discovered suite that throws part-way through loses every assertion after the throw, and the run reports that as **one** failure. Nothing says how much stopped existing.

Self-contained, on current `main`:

```js
// tests/zz-partial-probe.test.mjs
import { pass } from './helpers.mjs';
pass('assertion 1 of 5');
pass('assertion 2 of 5');
throw new Error('something the third assertion needed');
pass('assertion 3 of 5');   // and 4, and 5
```

```
✅ assertion 1 of 5
✅ assertion 2 of 5
❌ tests\zz-partial-probe.test.mjs — suite threw and was contained (Error): …
📊 Results: 2 passed, 1 failed, 0 warnings
```

Three assertions did not run. The summary calls that one failure, and the count it prints is the count of assertions that *happened*, so there is nothing to compare it against.

### Why it matters more than a diagnostics nit

The build is red either way, so nothing ships broken on the platform where the throw occurs. The cost lands on the platform where it does **not**.

Live case, this week. `tests/scratch-dirs.test.mjs` in #3968 called `symlinkSync(…, 'dir')`, which is `EPERM` on a non-elevated Windows shell without Developer Mode. On that box:

```
📊 6 passed, 1 failed        ← four assertions never ran
```

on CI, `test (windows-latest)`: **pass, 7m25s**. So the only signal that four assertions do not exist on the default Windows configuration was a local run that describes itself as *one* broken assertion. A reviewer reading "6 passed, 1 failed" reasonably concludes one thing is wrong; the true statement is "this file has 11 assertions and 4 of them cannot run here". I only found the real number by fixing the throw and re-running.

This is the shape #3303 was about, one level down: there, a suite ran nowhere and passed the whole time it was not running. Here, part of a suite runs nowhere and the harness reports a number that looks like coverage.

### Mechanism

`test-all.mjs:201` — a contained throw becomes a single `fail()`:

```js
} catch (err) {
  fail(`${rel} — suite threw and was contained (${err?.code ?? err?.name ?? 'Error'}): ${err?.message ?? err}`);
```

The comment two lines below already names the gap — *"a suite that dies mid-import leaves no other clue how far it got"* — and answers it with a stack trace, which tells you **where** it stopped, never **how much** was left.

[...]

PR #4065, opened 2026-09-10
fix: run the web test suite on the minimum supported Node

jaybarrett34:
## Problem

`web/package.json` declares `"engines": { "node": ">=22.0.0" }`, but two suites cannot run there at all:

```
TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts"
  for web/src/lib/apply/cv.ts
  for web/src/lib/explore-ai.ts
```

`apply-cv-resolver.test.mjs` and `explore-ai-dedup.test.mjs` import `.ts` modules directly. Node strips types automatically only from **23.6**; on 22.x it needs `--experimental-strip-types`, which neither call site passed.

## Why this reads as two known-failing suites

The failure is a **loader error, not an assertion**, so both suites abort before executing a single test. They surface as two red lines rather than as eleven tests that never ran — so it looks like two flaky suites instead of missing coverage.

## Two call sites

| | |
|---|---|
| `web/package.json` | the `npm test` script |
| `test-all.mjs` | the #2185 write-scope gate, which invokes the web suites directly and so was unaffected by the `package.json` script |

Fixing only the first leaves the core suite still failing, which is how the second site was found.

The flag remains valid on 23.6+ and 24, where type stripping is already the default. The failure message's reproduce hint is updated to match the invocation it now describes.

## Verification

On Node v22.14.0:

| | before | after |
|---|---|---|
| `npm test` in `web/` | 499 tests, 2 failing suites | **508 tests, 0 failures** |
| `test-all.mjs` web gate | fail | pass |

Eleven tests that were never executing now execute, and all pass.

## Summary

Web tests now support direct TypeScript module imports with Node.js 22.6.0 or later:

- `web/package.json:10`: Adds `--experimental-strip-types` to `npm test`.
- `test-all.mjs:76`: Adds the flag to the web test gate.
- `test-all.mjs:80`: Updates the reported command.
- `web/package.json`: Raises the minimum Node.js version to `>=22.6.0`.
- `web/README.md`: Documents the new Node.js requirement and corrected reproduction hint.

On older Node.js 22.x versions, web tests do not support this flag and may fail. On Node.js v22.14.0, the suite passes all 508 tests.

`AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/` are not changed.

### DIFF
--- a/test-all.mjs
+++ b/test-all.mjs
@@ -16169,13 +16169,13 @@ try {
         if (existsSync(join(ROOT, 'web', 'tests', 'lib'))) {
           fail('web/tests/lib contains no *.test.mjs — the #2185 unit suites are not being gated');
         }
-      } else if (run(NODE, ['--test', ...webUnits], { timeout: 180000 }) !== null) {
+      } else if (run(NODE, ['--experimental-strip-types', '--test', ...webUnits], { timeout: 180000 }) !== null) {
         pass('web pdf write-scope unit suites pass (#2185)');
       } else {
         // The signal distinguishes a timeout/kill from an assertion failure —
         // run()'s default 30s is short for six suites in one child process.
         const killed = lastRunFailure()?.signal;
-        fail(`web pdf write-scope unit suites failed${killed ? ` (killed: ${killed})` : ''} (run: node --test ${webUnits.join(' ')})`);
+        fail(`web pdf write-scope unit suites failed${killed ? ` (killed: ${killed})` : ''} (run: node --experimental-strip-types --test ${webUnits.join(' ')})`);
       }
 
       // Parity: everything web/package.json would run must be something we DO run.
--- a/web/README.md
+++ b/web/README.md
@@ -11,7 +11,7 @@ database, no server. If you never run it, nothing about your CLI workflow change
 
 ## Quick start
 
-Requires Node 22+ (see [Tests](#tests) — `npm test`'s glob discovery needs it).
+Requires Node 22.6+ (see [Tests](#tests) — `npm test` needs glob discovery, and `--experimental-strip-types` for the suites that import `.ts` modules directly; the flag landed in 22.6.0).
 
 ```bash
 cd web
--- a/web/package.json
+++ b/web/package.json
@@ -2,7 +2,7 @@
   "name": "@career-ops/web",
   "version": "0.10.0",
   "private": true,
-  "description": "Official web experience for career-ops \u2014 local-first.",
+  "description": "Official web experience for career-ops — local-first.",
   "scripts": {
     "dev": "next dev",
     "build": "next build",
--- a/web/package.json
+++ b/web/package.json
@@ -2,13 +2,13 @@
   "name": "@career-ops/web",
   "version": "0.10.0",
   "private": true,
-  "description": "Official web experience for career-ops — local-first.",
+  "description": "Official web experience for career-ops \u2014 local-first.",
   "scripts": {
     "dev": "next dev",
     "build": "next build",
     "start": "next start",
     "typecheck": "tsc --noEmit",
-    "test": "node --test \"tests/**/*.test.mjs\""
+    "test": "node --experimental-strip-types --test \"tests/**/*.test.mjs\""
   },
   "engines": {
     "node": ">=22.0.0"
--- a/web/package.json
+++ b/web/package.json
@@ -11,7 +11,7 @@
     "test": "node --experimental-strip-types --test \"tests/**/*.test.mjs\""
   },
   "engines": {
-    "node": ">=22.0.0"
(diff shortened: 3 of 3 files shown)


artemtrofymenko (commented):
Ran this across four Node versions on Windows 11, since the change is entirely about which Node you have. **The failure is real and the fix works — but the version boundary in the description is wrong, and getting it right makes a simpler fix available.**

Measured at PR head `6023de4` against base `da8c6f9`, both suites in one child process:

| Node | base, `node --test` (today) | PR head, `--experimental-strip-types --test` |
| --- | --- | --- |
| 20.11.1 | `ERR_UNKNOWN_FILE_EXTENSION` — `# tests 2 # pass 0 # fail 2` | **exit 9, `bad option: --experimental-strip-types`** |
| 22.22.0 | **11 pass, 0 fail** | 11 pass, 0 fail |
| 25.5.0 | 11 pass | 11 pass |
| 25.9.0 | 11 pass | 11 pass |

## The boundary is 22.18, not 23.6

> Node strips types automatically only from **23.6**; on 22.x it needs `--experimental-strip-types`

22.22.0 is 22.x and runs both suites clean with no flag. The unflagging was backported: the [22.18.0 release notes](https://nodejs.org/en/blog/release/v22.18.0) carry `(SEMVER-MINOR) module: unflag --experimental-strip-types` and say *"Node.js will be able to execute TypeScript files without additional configuration"*. 23.6.0 is where it landed on the **current** line; 22.18.0 is where the LTS line got it.

So the broken range is **22.0 – 22.17**, not all of 22.x. The bug is real for anyone sitting there, and `engines: ">=22.0.0"` does admit them.

## Which makes a smaller fix available

If the flag is only needed below 22.18, then `engines: ">=22.18.0"` with **no flag at all** covers the same ground: type stripping is automatic from there up, and the two call sites stay plain `node --test`. That is worth weighing against the current shape, because an `--experimental-` flag is a thing you then carry in two places until someone notices Node renamed or removed it — and the flag exists precisely to opt into behaviour that is now the default.

The trade is real in both directions: `>=22.18.0` cuts off 22.0–22.17 users who work today, while the flag keeps them. That is a support-window call, not a technical one, so it is the author's and the maintainer's to make. I am naming it because the description's 23.6 framing hides that the option exists.

## The flag makes the out-of-range failure worse

`engines` is advisory here — no `engine-strict` in `.npmrc` at either level — so nothing stops an older Node from running `npm test`. On 20.11.1, base gives two failing tests naming the exact `.ts` imports; the PR head gives `bad option`, exit 9, before a single file loads. Out of the declared range either way, but one of those tells you what to fix and the other tells you nothing. If the flag stays, `engines` is the only thing standing between a 20.x user and an unreadable exit, and it is not enforcing anything.

## The half of the description I would keep verbatim

> The failure is a **loader error, not an assertion**, so both suites abort before executing a single test. They surface as two red lines rather than as eleven tests that never ran.

Confirmed exactly, on 20.11.1: `# tests 2 # pass 0 # fail 2` where the file holds eleven. That is an independent instance of #3976 — a suite whose assertions stop existing while the run reports a small number that reads like a verdict — and it is a better example than the one I filed that issue with, because here the arithmetic is visible in the same output. Worth cross-linking whichever way this lands.

It is also why CI never caught it: every job in `.github/workflows/test.yml` pins `node-version: '24'`, which is past the unflagging on both lines.

## One adjacent thing, since this PR edits that file

On 25.x both suites emit:

```
[MODULE_TYPELESS_PACKAGE_JSON] Warning: Module type of …/web/src/lib/apply/cv.ts is not specified
and it doesn't parse as CommonJS. Reparsing as ES module … This incurs a performance overhead.
To eliminate this warning, add "type": "module" to …/web/package.json.
```

Node names the file this PR already changes and the one-line remedy. Not a blocker, and not what the PR is for — take it or leave it.

## What I did not verify

The 22.17/22.18 boundary itself is from the release notes, not from a run: I have 20.11.1, 22.22.0, 25.5.0 and 25.9.0 installed, and no 22.x below the unflagging. What I measured is that 22.22.0 needs no flag, which is enough to place the boundary below it but not to prove it is exactly 22.18.0.

santifer:
Hi @jaybarrett34, this PR has waited 12 days without a reply from a maintainer. That is on us, not on you, and we are sorry for it.

Where it actually stands right now: **ready for a maintainer decision**.

santifer (approved):

santifer:
Merged. Thank you, @jaybarrett34: your change is now in `main`.

If this was your first contribution here, welcome aboard. If you want to keep going, `good first issue` and `help wanted` are curated for exactly that, and you can ask on any of them before starting.

PR #4066, opened 2026-09-10
fix(set-status): add --help/-h and list the canonical states in usage

jduffey1990:
## Problem

`node set-status.mjs --help` fell through to the unknown-flag branch and exited 1:

```
❌ Unknown flag: --help
```

A bare invocation printed usage but exited 1 as a missing-operand error, so there was **no exit-0 path to the help text at all**. That breaks `cmd --help` in CI smoke checks and `|| true` idioms. This is the same class already fixed in `plugin-install.mjs` (#3857), `scan.mjs` (#2270) and `scan-ats-full.mjs` (#1633/#1635).

Separately, the usage block only *named* `templates/states.yml`:

```
<state>            Canonical state from templates/states.yml (aliases accepted)
```

So the one question a caller has at the prompt — *which states may I pass?* — was answerable only by opening another file, or by guessing wrong and reading the rejection. The list is already loaded at runtime to build that rejection message, so printing it up front costs nothing.

## What changed

`--help`/`-h` now prints usage plus the canonical states and exits 0:

```
Canonical states (aliases also accepted — see templates/states.yml):
  Evaluated  Offer evaluated with report, pending decision
  Applied    Application submitted
  Responded  Company has responded (not yet interview)
  Interview  Active interview process
  Offer      Offer received  (terminal)
  Rejected   Rejected by company  (terminal)
  Discarded  Discarded by candidate or offer closed  (terminal)
  SKIP       Doesn't fit, don't apply  (terminal)
  Hired      Offer accepted, job landed!  (terminal)

  A terminal state ends the application. Discarded is YOUR decision or a
  closed req; Rejected is theirs; SKIP means never applied for.
```

Design decisions worth reviewing:

- **Handled before any other parsing**, so help wins over a missing-operand error *and* over an invalid flag value. A caller reaching for `--help` after a failed run usually still has the bad arguments on the line; re-rejecting them instead of answering is the behavior this removes.
- **Detection skips value positions.** `--note "--help"` stays the existing "never consume a flag as a value" error rather than becoming a help screen that exits 0 — which would turn a refused write into a silent success for any script checking only the exit code.
- **The states section is generated from `states.yml`**, with each description and a terminal marker — not a hardcoded copy. A test asserts every state in that file appears, so a new state cannot silently go undocumented.
- **An unreadable or malformed `states.yml` degrades to the static pointer** instead of throwing. A broken states file is a real error, but it belongs to the run that tries to *write* a state, not to `--help`.
- **Bare invocation still exits 1.** Missing operands are a usage error, not a request for help.
- Adds a worked examples block.

`loadCanonicalStates()` now passes `description` and `terminal` through from `states.yml`. Additive — the other callers (`normalize-statuses`, `funnel-velocity`, `followup-cadence`, `check-jd-archive`, `test-all`) read only `id`/`label`/`aliases`.

## Testing

New `tests/set-status-help-flag.test.mjs` — 10 tests, all passing. Hermetic: every run pins `CAREER_OPS_TRACKER` at a nonexistent path, so `No tracker found` doubles as proof that help short-circuits before any tracker access. Each assertion also checks the subprocess actually ran (no spawn error, no signal), so a timeout cannot pass silently. Mirrors the structure of `tests/scan-help-flag.test.mjs`.

Coverage: exit 0 and usage printed; `-h` byte-identical to `--help`; every canonical state present; every description present; terminal states marked; tracker never consulted; help wins over invalid args; `--help` in a value position is *not* a help request; unknown flag still exits 1; bare invocation still exits 1.

Also verified clean: `tests/tracker-utils.test.mjs`, `tests/states-alias-coverage.test.mjs`, `tests/set-status-seeds-followup.test.mjs`, and direct runs of all four `loadCanonicalStates` consumers.

`node test-all.mjs --quick` on my machine reports 8666 passed / 10 failed. **The 10 failures are pre-existing and environmental, not from this change** — I confirmed by stashing the diff and re-running the failing suites, which fail identically on a clean baseline. They are `node:sqlite is not available. tracker.mjs needs Node >= 22.5 (you are on v20.19.6)` and the `web/` unit suites, whose dependencies I had not installed. I expect CI on Node >= 22.5 to be green.

## Summary

Users can run `node set-status.mjs --help` or `node set-status.mjs -h` to view usage, canonical states, descriptions, terminal markers, and examples. Help exits with code 0 and runs before tracker access: `set-status.mjs:141-151`.

Help detection skips option values. Therefore, `--note --help` keeps its existing missing-value error. Bare invocation remains a usage error with exit code 1: `set-status.mjs:180-198`.

If `states.yml` cannot be read or parsed, help prints a static pointer instead of failing: `set-status.mjs:141-151`. `loadCanonicalStates()` now returns `description` and `terminal` fields: `tracker-utils.mjs:710-734`.

The change adds tests for help output, state documentation, argument handling, tracker isolation, malformed state data, and exit codes: `tests/set-status-help-flag.test.mjs:34-110`.

The documentation adds the new help invocation and behavior: `docs/SCRIPTS.md:1069-1088`.

## System files

### DIFF
--- a/docs/SCRIPTS.md
+++ b/docs/SCRIPTS.md
@@ -1077,7 +1077,15 @@ node set-status.mjs --report N <state> [--note "..."]       # row whose Report c
 node set-status.mjs "Company Name" Applied --role "Role"    # narrow match by role fragment
 node set-status.mjs --row 12 Applied
 node set-status.mjs --report 345 Applied --on 2026-08-01
+node set-status.mjs --help                                  # usage + the canonical states, exits 0
 ```
+
+`--help`/`-h` prints the usage block followed by every canonical state with its
+one-line description, read from `templates/states.yml` rather than duplicated —
+so the states are answerable at the prompt instead of requiring another file.
+It short-circuits before any tracker access and exits 0. A bare invocation with
+no operands still prints usage and exits 1, because missing operands are a usage
+error rather than a request for help.
 
 A bare number or company name is convenient, but becomes ambiguous when multiple tracker rows exist for a company or when tracker row IDs and report IDs diverge. That divergence is permanent once it starts: `reserve-report-num.mjs` treats tracker row IDs as occupied when it allocates a report number, so a row that never got a report still consumes a number the report sequence then skips — the two counters leapfrog each other and never realign. On a diverged tracker "5" may mean tracker row #5 or report #5, which are different applications. Base selectors resolve the main target, while explicit selectors and filters disambiguate the target row:
 
--- a/set-status.mjs
+++ b/set-status.mjs
@@ -138,18 +138,10 @@ Examples:
   node set-status.mjs --row 7 Discarded --dry-run`;
 
 /**
- * Render the canonical states for `--help`.
+ * Render the canonical states from states.yml for `--help`.
  *
- * The states live in templates/states.yml, and before this the usage block only
- * NAMED that file — so the one question a caller actually has at the prompt
- * ("which states may I pass?") was answerable only by opening another file, or
- * by guessing wrong and reading the rejection. The list is already loaded at
- * runtime for that rejection message; printing it up front costs nothing.
- *
- * Help must never be the thing that fails, so an unreadable or malformed
- * states.yml degrades to the static pointer instead of throwing: a broken
- * states file is a real error, but it belongs to the run that tries to WRITE a
- * state, not to `--help`.
+ * A broken states.yml degrades to a pointer rather than throwing: that failure
+ * belongs to the run that tries to WRITE a state, not to `--help`.
  *
  * @returns {string} The states section, or a pointer line when unreadable.
  */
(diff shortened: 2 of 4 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is in `set-status.mjs` (the canonical tracker write path, core), the maintainer's call; I am reviewing the owned test `tests/set-status-help-flag.test.mjs`.

It is sound. The test spawns the real set-status.mjs with --help and -h and asserts the output contains every canonical states.yml label plus its description and the terminal marker (deriving the expected list from `loadCanonicalStates`, so a hardcoded copy would not satisfy it), and that both flags print identical usage. It confirms help short-circuits before any work: it pins CAREER_OPS_TRACKER to a nonexistent path and asserts the output does not contain "No tracker found", so the tracker, the states validation, the lock, and the write are never reached; a bare --help with no other args exits 0 with usage rather than a usage error, and --help wins over otherwise-invalid args while the value-position guard stays intact. I mutation-checked it: disabling the short-circuit so --help falls through reddens the suite, and a normal non-help invocation still writes the row correctly, so the handling is additive.

santifer:
Sorry for the wait — this has been ready since 11 September and the delay is entirely on our side.

The change is right, and the part worth calling out is the one not in the title: printing the canonical states rather than naming the file they live in. "Which states may I pass?" was answerable only by opening another file or by guessing wrong and reading the rejection, and the list was already loaded to build that rejection. Reading it from the states file instead of duplicating it means a new state stays a one-file change.

You had already addressed the own-property point before anyone got back to you. I re-ran it here: `set-status.mjs toString --help` prints help, and `--note --help` still exits 1 with a missing-value error, so the value-position guard you described holds. The help-flag suite, the tracker-utils suite and the set-status follow-up suite all pass on your head.

@Scott-Emberson reviewed the test file and routed the rest to us. That routing is what we were slow on, not anything you owed.

Taking it from here: the branch needs a refresh against main and a green run, then it goes in. Thanks — and welcome.

santifer (approved):

PR #4072, opened 2026-09-10
test(url-key): give url-key.mjs its own suite

### LINKED ISSUE #4070: test(url-key): url-key.mjs has no suite of its own, only cases inside merge-tracker-url-dedup
### Code of Conduct

### Existing issues

### What happened?

`tests/merge-tracker-url-dedup.test.mjs` covers two root scripts at once, which the documented suite convention does not allow.

`ARCHITECTURE.md:30`:

> Suites live in `tests/`, named `{module}.test.mjs` for the root script they cover

`CONTRIBUTING.md:189` says the same. The file imports and exercises both `url-key.mjs` and `merge-tracker.mjs`, and there is no `tests/url-key.test.mjs` at all, so `url-key.mjs`'s unit behaviour is reachable only through a suite named for a different script. Someone changing `normalizeUrl` or `isAggregatorUrl` has no obvious file to run.

### Steps to reproduce

Not a runtime defect. It is visible from the file:

```
$ grep -n '^import' tests/merge-tracker-url-dedup.test.mjs
14:import { normalizeUrl, isAggregatorUrl } from '../url-key.mjs';
```

and the suite also shells out to `merge-tracker.mjs` through `execFileSync` for its integration cases.

### Expected behavior

`url-key.mjs`'s unit tests live in `tests/url-key.test.mjs`, and `tests/merge-tracker-url-dedup.test.mjs` keeps the merge-behaviour cases.

The seam is clean, which is the reason to do it now rather than let it set. I classified all 32 test blocks in the file by which module each one actually exercises:

| exercises | blocks |
|---|---|
| `url-key.mjs` only | 11 |
| `merge-tracker.mjs` only | 20 |
| both | 1 |

So this is a move of 11 blocks plus one judgement call on the single mixed case, not a rewrite. No behaviour changes, and `test-all.mjs` discovers `tests/**/*.test.mjs` with no registration list, so the new file runs the moment it exists.

### Context

Raised by the review bot on #3653. I declined it there on the grounds that it was a large refactor competing for review attention on a PR already carrying two confirmed bugs. The first half of that was wrong: I had not measured it, and the size claim I gave was not accurate. The placement argument holds, so it is here as its own change instead.

### CLI tool

Claude Code

### OS and Node.js version

macOS 26.6.2 / Node 26.8.1

---

rubicon:
## What does this PR do?

Moves `normalizeUrl`'s unit tests out of `tests/merge-tracker-url-dedup.test.mjs` into a new `tests/url-key.test.mjs`, so `url-key.mjs` has a suite named for itself.

## Related issue

Closes #4070

## Type of change

- [ ] Bug fix
- [ ] New feature

## Checklist

---

## Why

`ARCHITECTURE.md:30` and `CONTRIBUTING.md:189` both specify `tests/{module}.test.mjs` for the root script a suite covers. `url-key.mjs` had no suite at all. Its unit cases sat inside a file named for `merge-tracker.mjs`, which means someone changing `normalizeUrl` had no obvious file to run, and `test-all.mjs --only url-key` matched nothing.

## It is a move, not a rewrite

The file was already sectioned by a divider comment at line 52, and the split follows that seam exactly:

| exercises | blocks | goes to |
|---|---|---|
| `normalizeUrl` | 6 | `tests/url-key.test.mjs` |
| drives the `merge-tracker.mjs` CLI | 16 | stays |
| both | 0 | n/a |

Nothing exercised both, so no case had to be duplicated or rewritten. The assertions are byte-identical to what they were; only their file changed. Both headers now point at each other, and the `normalizeUrl` import is dropped from the file that no longer uses it.

## Verification

`node tests/url-key.test.mjs`: 6 passed. `node tests/merge-tracker-url-dedup.test.mjs`: unchanged, all green. `node test-all.mjs --quick`: 8687 passed, 0 failed.

Checked for orphaned imports in the file the cases left, since that is the obvious way to break a move like this: every remaining import is still referenced.

## Origin

The review bot raised this on #3653. I declined it there on a cost I had not measured, and the size I quoted was wrong, so this is the corrected answer. It is a separate PR because #3653 is carrying two confirmed bug findings and a test reorganization would compete with them for review.

## Summary

- Added `tests/url-key.test.mjs:1` as the dedicated `normalizeUrl` test suite.
- Moved URL-key unit tests out of `tests/merge-tracker-url-dedup.test.mjs:14`.
- Kept merge-tracker integration tests in their existing suite.
- Preserved behavior and assertions.
- No changes affect `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, or `.github/`.

## Verification

- URL-key tests passed: 6 tests.
- Merge-tracker tests remained green.
- `test-all.mjs --quick` passed: 8,687 tests, 0 failures.

### DIFF
--- a/tests/merge-tracker-url-dedup.test.mjs
+++ b/tests/merge-tracker-url-dedup.test.mjs
@@ -1,9 +1,10 @@
 // tests/merge-tracker-url-dedup.test.mjs — URL-keyed deterministic dedup.
 //
-// Unit-tests normalizeUrl (pure), then drives the REAL merge-tracker.mjs CLI
-// end-to-end against a temp tracker via the CAREER_OPS_TRACKER /
-// CAREER_OPS_ADDITIONS env hooks — the merge path is where the bug lived, so
-// asserting on the resulting tracker rows is what actually proves the fix.
+// Drives the REAL merge-tracker.mjs CLI end-to-end against a temp tracker via
+// the CAREER_OPS_TRACKER / CAREER_OPS_ADDITIONS env hooks — the merge path is
+// where the bug lived, so asserting on the resulting tracker rows is what
+// actually proves the fix. normalizeUrl's own unit cases live in
+// tests/url-key.test.mjs.
 import { pass, fail } from './helpers.mjs';
 import assert from 'node:assert';
 import { mkdtempSync, mkdirSync, writeFileSync, readFileSync, rmSync } from 'node:fs';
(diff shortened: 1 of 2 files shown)


Scott-Emberson (commented):
Verified by execution and mutation. Sound (both files under tests/).

The new `tests/url-key.test.mjs` imports the real `normalizeUrl` from `url-key.mjs` and asserts each normalization rule against literal expecteds. I mutation-checked it rule by rule: emptying the tracking-param denylist, forcing non-https, dropping the fragment strip, dropping the trailing-slash strip, dropping the query sort, and the no-key fallback each redden. The one rule whose mutation stays green is the host-lowercase line, and that is not a coverage gap: the WHATWG URL parser already lowercases the hostname, so that line is defensive and redundant, while the test's assertion is real and would catch a genuine host-case regression if the platform behavior ever changed.

The change to `tests/merge-tracker-url-dedup.test.mjs` is a clean pure move: the six normalizeUrl unit blocks are byte-for-byte the ones now in the new suite, the now-unused import is dropped, and the sixteen merge-tracker integration cases (real CLI end to end) stay in place and still pass. No coverage is lost, and the dedup path is still exercised through the real merge.

GATE-tier (tests/), owned tests sound, CI green, not behind. Good to go.

santifer (approved):

santifer:
Merged. Thank you, @rubicon: **give url-key.mjs its own suite** is now in `main` and ships with the next release. That is merge 56 from you here; the project is what it is because of contributors who keep coming back.

If there is something you want to take on next, say so and we will point you at it.

Issue #3149, opened 2026-08-21
update-system.mjs check: false 'system-files-changed' drift on every install, forever, right after a clean apply


venunayar:
## Summary

`node update-system.mjs check` reports `system-files-changed` / `update-available` **immediately after** a fully successful `node update-system.mjs apply --force`, even when the installed version matches the release and every system file was just synced from upstream `main`. The false positive never clears — every subsequent `check` repeats it, forever, on every install.

## Steps to reproduce

```
node update-system.mjs apply --force
# ... "Update complete: v1.28.0 → v1.28.0", 300+ paths updated, no errors

node update-system.mjs check
# {"status":"update-available", ..., "reason":"system-files-changed",
#  "local_commit":"<some sha>", "remote_commit":"<some other sha>"}
```

`local` and `remote` versions are equal; only the drift reason fires.

## Root cause

`check()` (added for #2630, "same-version manifest/file drift") flags drift by comparing:

- `localCommit = git rev-parse HEAD` (this install's current commit)
- `remoteCommit` = the SHA GitHub's API reports for `heads/main` on `santifer/career-ops`

against each other, treating any inequality as drift:

```js
try { localCommit = gitQuiet('rev-parse', 'HEAD'); } catch { /* no git checkout */ }
...
const systemTreeDrift = Boolean(localCommit && remoteCommit && localCommit !== remoteCommit);
```

But `apply()` never merges, fast-forwards, or otherwise adopts a commit from upstream's history. It:

1. `git fetch`es upstream into `FETCH_HEAD`
2. `git checkout FETCH_HEAD -- <system paths>` for the paths in `SYSTEM_PATHS`/target manifest
3. Creates a **brand-new local commit** (`chore: auto-update system files to v${remote}`), whose parent is the install's *own* prior HEAD, not anything from upstream's graph

Step 3 produces a commit object that is, by construction, never present in upstream's commit graph — it has a different parent, a different tree in general (only the system-layer subset of paths, not the full upstream tree), and a fresh SHA. So `git rev-parse HEAD` **can never equal** `heads/main`'s tip SHA, no matter how current the install's files actually are. `systemTreeDrift` is therefore `true` unconditionally, on every install, forever, from the very first run of the code that added this check.

This also means the check is *useless* for its intended purpose (catching same-version file drift) — it always fires, so it carries no signal.

## Verified

Confirmed by inspecting the actual SHAs after a clean `apply --force`:

```
$ git rev-parse HEAD
[...]

PR #4074, opened 2026-09-11
fix(updater): drop SIGNATURES.md from SYSTEM_PATHS so a new signature stops reading as system drift

### LINKED ISSUE #4062: update-system.mjs: SIGNATURES.md in SYSTEM_PATHS causes perpetual system-files-changed notice
### What happened

`update-system.mjs check` reports `{"status":"update-available","reason":"system-files-changed"}` on every session, even immediately after running `apply --confirm`.

Root cause: `SIGNATURES.md` (manifesto signatures) is listed in `SYSTEM_PATHS` in `update-system.mjs`, which `systemTreeDiffers()` uses to diff the local install against upstream `main`. But `SIGNATURES.md` is an append-only, continuously-growing file — new signatures get committed to `main` independent of any release. So:

1. User runs `apply --confirm`, syncing `SIGNATURES.md` to that moment's content.
2. Someone signs the manifesto upstream, adding a new line to `SIGNATURES.md` on `main`.
3. Next `check` diffs local vs. upstream `main` and finds `SIGNATURES.md` differs → reports `system-files-changed` again.
4. Loop repeats daily forever, regardless of whether any actual system file (modes/, scripts, etc.) changed.

### Reproduction

```bash
git fetch https://github.com/career-ops-hq/career-ops.git main
git diff --quiet --ignore-cr-at-eol FETCH_HEAD HEAD -- $(node -e "
$(sed -n '/^const SYSTEM_PATHS = \[/,/^\];/p' update-system.mjs)
console.log(SYSTEM_PATHS.join(' '))
")
echo $?  # 1 = diff found
git diff --name-only --ignore-cr-at-eol FETCH_HEAD HEAD -- <same path list>
# → SIGNATURES.md
```

On a fully up-to-date install (`bff2f4c`, v1.32.0, applied same day as the release), the only file differing from upstream `main` was `SIGNATURES.md` — one new signature line appended after the local apply.

### Suggested fix

Remove `SIGNATURES.md` from `SYSTEM_PATHS` in `update-system.mjs`. It's not a versioned system-behavior file (no modes/scripts/config logic lives there) — it's an ever-growing community ledger that will never stay in sync between two independently-timed snapshots. If it needs to ship with `apply` at all, it shouldn't be part of the drift-detection set.

### Impact

Every user hits the "update available" prompt on every session indefinitely, even right after updating, training people to either ignore real updates or churn through unnecessary `apply` runs that touch nothing meaningful.

maxmilian:
Closes #4062

## The bug

`SIGNATURES.md` is an append-only ledger of who signed the manifesto — **49 commits in the last 30 days**, none of which change anything an install runs. While it sat in `SYSTEM_PATHS` it was inside the pathspec `check()` diffs:

```
check(): SHA mismatch
  -> systemTreeDiffers(SYSTEM_PATHS, 'FETCH_HEAD')                          update-system.mjs:1879
  -> git diff --quiet --ignore-cr-at-eol FETCH_HEAD HEAD -- <SYSTEM_PATHS>  :1118-1130
  -> ANY one pathspec differing => reason: 'system-files-changed'           :1894
```

So the moment anyone signs the manifesto, **every install in the world** starts reporting `update-available / system-files-changed` — for a file no install reads. Re-syncing clears it only until the next signature lands, which on current cadence is roughly every other day.

### Reproduction (against `main@da8c6f9`)

Taking `fe06051` as a stale install ref and diffing with the real 315-entry `SYSTEM_PATHS`:

- `git diff --quiet --ignore-cr-at-eol fe06051 HEAD -- <315 paths>` → **exit 1** (drift)
- the same diff with `--name-only` lists **`SIGNATURES.md` and nothing else**
- drop that one entry → **exit 0** (no drift)

## Why this is not "yet another drift fix"

This is **one of several independent causes** of the #3149 class of permanent "update available", and it is orthogonal to the two already fixed below. (A further cause — skill entrypoint symlinks materialised into regular files on `core.symlinks=false` checkouts, confirmed on #3149 — is separate and must **not** be fixed the way this entry is; see the code comment.)

| # | Cause | Fix |
|---|---|---|
| #2630 | `apply()` commits upstream content as a *new* local commit, so HEAD can never equal upstream main — SHA inequality was read as drift | settle same-version mismatches on **content** (`systemTreeDiffers`) |
| #2756 | `.gitignore` cannot be checked out raw (users co-own it), so it stayed permanently out of sync | `reconcileGitignore()` — append-if-missing, excluded from the manifest |
| **#4062 (this PR)** | a **high-churn, install-irrelevant document** sits inside the content diff, so the content check itself fires forever | remove it from the pathspec |

#2630 made the check correct; this PR makes its *input* correct. The check has no way to distinguish "the ledger grew by one name" from "a script changed" — the pathspec is the only place that distinction can live.

It is also orthogonal to #3805 (in review): that one is about a different stage of the same script and the two do not touch the same lines.

## The change

1. **`update-system.mjs`** — remove `'SIGNATURES.md'` from `SYSTEM_PATHS`, replaced by a comment explaining *why it is excluded*, in the same house style as the `.gitignore` rationale at `:1979`.
2. **`validate-system-paths-coverage.mjs`** — removing the entry would make the file an orphan and turn the coverage guard red, so it gets a `REPO_ONLY` group of its own (not folded into `EXCLUDES`, for the same reason `.gitignore` has its own group: these entries differ in *why* they never ship). Two self-test assertions, one of which pins the **mechanism** — `!SYSTEM_PATHS.includes('SIGNATURES.md')` — because `covered()` would keep answering `true` if the entry came back, silently restoring the false positive.
3. **`tests/updater-signature-ledger-drift.test.mjs`** — new, auto-discovered (no `test-all.mjs` edit).

### Why excluded rather than exempted

`manifesto.mjs` only ever parses `MANIFESTO.md`; nothing on an install opens `SIGNATURES.md`. It is a repository-side record of who signed, not a shipped artifact, so it stops shipping instead of shipping-but-ignored. Side effects checked explicitly:

- `staleSystemFiles()` (`:972`) only prunes files the **install has and upstream lacks** — a dropped manifest entry never triggers a delete, so nobody's copy is removed.
- `missingFromTargetManifest()` (`:864`) iterates the **remote** list — no false "missing" warning.
- the only other reference to the filename is `test-all.mjs:1973`, a privacy-scan allowlist, unaffected.

## Test

`tests/updater-signature-ledger-drift.test.mjs` drives the **real** manifest — read out of `update-system.mjs` with `extractArrayFromSource`, never a hard-coded copy, so the test pins shipped behaviour rather than a fixture — against throwaway repos through the same `ctx.git` seam production uses (the approach `tests/updater-drift-detection.test.mjs` established):

- upstream appends one signature, nothing else moves → **not** drift
- upstream edits a genuine system file → **still** drift (the fix must narrow the pathspec, not blunt the check)

**Red-armed both ways.** With `'SIGNATURES.md',` put back:

```
  ✅ real SYSTEM_PATHS read from update-system.mjs (315 entries, includes MANIFESTO.md)
  ✅ fixture: SIGNATURES.md is the only file differing from upstream
  ❌ a new signature alone reported as system drift — #4062 is back
  ✅ an upstream change to MANIFESTO.md still reports drift
📊 Results: 3 passed, 1 failed
FAIL: SIGNATURES.md must NOT re-enter SYSTEM_PATHS (#4062)   [coverage self-test]
```

With it removed:

```
  ✅ real SYSTEM_PATHS read from update-system.mjs (314 entries, includes MANIFESTO.md)
  ✅ fixture: SIGNATURES.md is the only file differing from upstream
  ✅ a new signature alone is NOT system drift
  ✅ an upstream change to MANIFESTO.md still reports drift
📊 Results: 4 passed, 0 failed
ALL SELF-TESTS PASSED
```

Full suite (not `--quick`): **8692 passed, 0 failed, 16 warnings**, exit 0. The 16 warnings are byte-identical to the pre-change baseline apart from one randomized probe filename.

### One note for future editors

`updater-migration-tests.mjs:60` parses `SYSTEM_PATHS` with a **comment-blind** regex, unlike the exported `extractArrayFromSource` which strips comments first. A straight quote or apostrophe anywhere in a comment inside that array becomes a phantom manifest entry and fails the suite with a wall of `SYSTEM_PATHS entry missing from tree`. The new comment is deliberately quote-free and says so in place. Happy to split a parser-convergence hardening into its own PR if you would prefer the two to agree.

## Alternative, if you would rather keep shipping the ledger

Keep the `SYSTEM_PATHS` entry (so installs continue to receive `SIGNATURES.md`) and introduce a separate `DRIFT_EXEMPT_PATHS`, subtracted from the pathspec only at the `systemTreeDiffers` call site in `check()`. That separates "what ships" from "what counts as drift" and generalizes to any future high-churn shipped document.

I went with removal because it is the smaller change and nothing on an install reads the file, so the extra concept buys nothing today — but the exempt-list version is a ~10-line diff and I am glad to switch if you would rather installs keep a current copy of the ledger. Happy to take either.

https://claude.ai/code/session_01XDR1wm73oH8J8cRWnKyv9m

## Summary

`SIGNATURES.md` no longer causes false `system-files-changed` notices. Users will not receive repeated update prompts when only the signature ledger changes upstream.

## Changes

- Removed `SIGNATURES.md` from `SYSTEM_PATHS` in `update-system.mjs:395-413`.
- Added repository-only coverage and assertions in `validate-system-paths-coverage.mjs:95-104,164-168`.
- Added regression tests for signature-only changes and real `MANIFESTO.md` drift in `tests/updater-signature-ledger-drift.test.mjs:93-141`.
- Kept symlinked skill entrypoints in `SYSTEM_PATHS`; drift comparison must handle them separately.

## User impact

`check` still detects genuine system-file changes. A signature-only update no longer reports drift or prompts users to apply an update.

The full suite reports 8,692 passed, 0 failed, with 16 warnings.

## System files touched

- `update-system.mjs`
- `validate-system-paths-coverage.mjs`
- `tests/updater-signature-ledger-drift.test.mjs`

`AGENTS.md`, `modes/`, `DATA_CONTRACT.md`, `providers/`, and `.github/` are unchanged.

### DIFF
--- a/tests/updater-signature-ledger-drift.test.mjs
+++ b/tests/updater-signature-ledger-drift.test.mjs
@@ -1,0 +1,147 @@
+/**
+ * updater-signature-ledger-drift.test.mjs — a new manifesto signature must not
+ * read as system drift on every install (#4062).
+ *
+ * SIGNATURES.md is an append-only ledger of who signed MANIFESTO.md. It used
+ * to sit in SYSTEM_PATHS, which put it inside the pathspec check() diffs:
+ *
+ *   systemTreeDiffers(SYSTEM_PATHS, 'FETCH_HEAD')
+ *     → git diff --quiet --ignore-cr-at-eol FETCH_HEAD HEAD -- <SYSTEM_PATHS>
+ *     → any one pathspec differing ⇒ reason: 'system-files-changed'
+ *
+ * The ledger takes dozens of commits a month while the code it ships beside
+ * does not, so the moment anyone signed, every install on the planet started
+ * reporting `update-available / system-files-changed` — the same permanent
+ * false positive as #3149, from a third independent cause (after the SHA-vs-
+ * content bug of #2630 and the ignore-rule route of #2756).
+ *
+ * These tests drive the REAL manifest — read out of update-system.mjs with
+ * extractArrayFromSource, never a hard-coded copy — against throwaway repos
+ * through the same ctx.git seam production uses. Hard-coding the path list
+ * here would pin the fixture instead of the shipped behaviour, and the
+ * regression would come back the first time someone re-added the entry.
+ *
+ *   - upstream appends a signature, nothing else moves → NOT drift
+ *   - upstream edits a genuine system file             → still drift
+ *     (the fix must narrow the pathspec, not blunt the check)
+ */
+
+import { mkdtempSync, mkdirSync, writeFileSync, appendFileSync, readFileSync } from 'fs';
+import { tmpdir } from 'os';
+import { join, dirname } from 'path';
+import { pass, fail, rmSync, ROOT } from './helpers.mjs';
+import { gitIn, systemTreeDiffers, extractArrayFromSource } from '../update-system.mjs';
+
+const SIGNATURES = 'SIGNATURES.md';
+// A second real system file, used as the control: whatever the manifest says
+// today, an upstream change to THIS one must still report drift.
+const CONTROL = 'MANIFESTO.md';
+const USER_PATH = 'data/applications.md';
+
+const realSystemPaths = extractArrayFromSource(
+  readFileSync(join(ROOT, 'update-system.mjs'), 'utf-8'),
+  'SYSTEM_PATHS',
+);
+
+function seed(dir, path, body) {
+  mkdirSync(join(dir, dirname(path)), { recursive: true });
+  writeFileSync(join(dir, path), body);
+}
+
+function makeOrigin() {
+  const dir = mkdtempSync(join(tmpdir(), 'co-sig-origin-'));
+  const g = (...args) => gitIn(dir, ...args);
+  g('init', '-q', '-b', 'main', '.');
+  g('config', 'user.email', 'test@example.com');
+  g('config', 'user.name', 'Test');
+  g('config', 'core.autocrlf', 'false');
+  g('config', 'commit.gpgsign', 'false');
+  seed(dir, SIGNATURES, '# Signatures\n\n- @first\n');
+  seed(dir, CONTROL, '# Manifesto\n\nv1\n');
+  seed(dir, USER_PATH, '| base | Acme | Intern |\n');
+  g('add', '-A');
+  g('commit', '-qm', 'base');
+  return { dir, g };
+}
+
+function cloneInstall(originDir) {
+  const dir = mkdtempSync(join(tmpdir(), 'co-sig-install-'));
+  gitIn(dir, 'clone', '-q', originDir, '.');
+  const g = (...args) => gitIn(dir, ...args);
+  g('config', 'user.email', 'test@example.com');
+  g('config', 'user.name', 'Test');
+  g('config', 'core.autocrlf', 'false');
+  g('config', 'commit.gpgsign', 'false');
+  return { dir, g };
+}
+
+function cleanup(...dirs) {
+  for (const d of dirs) rmSync(d, { recursive: true, force: true });
+}
+
+console.log('\n🧪 Testing signature-ledger drift (#4062)...');
+
+// ── 0. Fixture preconditions ────────────────────────────────────────────────
+// The two tests below are only meaningful if the real manifest is what it
+// claims to be: non-empty, and still shipping the control file.
+if (realSystemPaths.length > 0 && realSystemPaths.includes(CONTROL)) {
+  pass(`real SYSTEM_PATHS read from update-system.mjs (${realSystemPaths.length} entries, includes ${CONTROL})`);
+} else {
+  fail(`could not read a usable SYSTEM_PATHS from update-system.mjs (${realSystemPaths.length} entries)`);
+}
+
+// ── 1. THE REGRESSION ───────────────────────────────────────────────────────
+// Someone signs the manifesto upstream. Nothing an install runs has changed.
+// check() must not call that system-files-changed.
+{
+  const origin = makeOrigin();
+  const install = cloneInstall(origin.dir);
+  try {
+    appendFileSync(join(origin.dir, SIGNATURES), '- @second\n');
+    origin.g('add', '-A');
+    origin.g('commit', '-qm', 'docs(signatures): add @second');
+
+    install.g('fetch', '-q', origin.dir, 'main');
+
+    // Fixture guard: the ledger really did diverge between the two refs.
+    const changed = install.g('diff', '--name-only', 'FETCH_HEAD', 'HEAD').split('\n').filter(Boolean);
+    if (changed.length === 1 && changed[0] === SIGNATURES) {
+      pass(`fixture: ${SIGNATURES} is the only file differing from upstream`);
+    } else {
+      fail(`fixture: expected only ${SIGNATURES} to differ, got [${changed.join(', ')}]`);
+    }
+
+    const drift = systemTreeDiffers(realSystemPaths, 'FETCH_HEAD', { git: (...a) => gitIn(install.dir, ...a) });
+    if (drift === false) {
+      pass('a new signature alone is NOT system drift');
+    } else {
+      fail('a new signature alone reported as system drift — #4062 is back');
+    }
+  } finally {
+    cleanup(origin.dir, install.dir);
+  }
+}
+
+// ── 2. The check still works ────────────────────────────────────────────────
+// Narrowing the pathspec must not cost real drift detection: an upstream
+// change to a genuine system file still has to report.
+{
+  const origin = makeOrigin();
+  const install = cloneInstall(origin.dir);
+  try {
+    writeFileSync(join(origin.dir, CONTROL), '# Manifesto\n\nv2\n');
+    origin.g('add', '-A');
+    origin.g('commit', '-qm', 'docs: revise manifesto');
+
+    install.g('fetch', '-q', origin.dir, 'main');
+
+    const drift = systemTreeDiffers(realSystemPaths, 'FETCH_HEAD', { git: (...a) => gitIn(install.dir, ...a) });
+    if (drift === true) {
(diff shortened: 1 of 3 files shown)


Scott-Emberson (commented):
Test-axis only. The substance is in `update-system.mjs` (the updater signature-ledger, a red-line), the maintainer's call; I am reviewing the owned test `tests/updater-signature-ledger-drift.test.mjs`.

It is sound and drives real behavior, not a source grep. The test imports the real `gitIn`, `systemTreeDiffers`, and `extractArrayFromSource`, reads the live SYSTEM_PATHS array out of the shipped source, and drives it through `systemTreeDiffers`, the same function `check()` uses in production, against throwaway git repos. It asserts SIGNATURES.md's absence from SYSTEM_PATHS by consequence: an upstream that only appends a signature to SIGNATURES.md must report no system drift, and a control that edits MANIFESTO.md must still report drift (so the fix narrowed the pathspec rather than blunting the check). Root pinning is clean (mkdtempSync temp dirs, explicit git seams). I mutation-checked it: re-adding SIGNATURES.md to SYSTEM_PATHS reddens the exact regression assertion (a signature alone reported as drift, #4062 back).

artemtrofymenko (commented):
Ran it on default Windows 11 at `3db89ff`. **The fix works**, and the one thing I would change is a sentence rather than code — because as written it points the next person at a fix that would silently break Windows installs.

```
✅ real SYSTEM_PATHS read from update-system.mjs (314 entries, includes MANIFESTO.md)
✅ fixture: SIGNATURES.md is the only file differing from upstream
✅ a new signature alone is NOT system drift
✅ an upstream change to MANIFESTO.md still reports drift
📊 4 passed, 0 failed
```

The last assertion is the one that matters, and it is good that it is there: the drift check is narrowed by one path, not silenced.

## There is a fourth cause, confirmed and still open

`update-system.mjs:403` on this branch calls this *"the third independent cause of the #3149 class, after the SHA-vs-content bug of #2630 and the ignore-rule route of #2756"*. That comment ships, so it is what the next reader of the updater is told.

#3149 carries a second cause that @santifer labelled `confirmed` on 2026-09-05, and it is not fixed. Re-measured on current `main` (`bb641dc`) against a Windows install:

```
:120000 100644 82b4547 e2e5ea6 T   .claude/skills/career-ops/SKILL.md
:120000 100644 82b4547 e2e5ea6 T   .cursor/skills/career-ops/SKILL.md
… the same for .grok, .kimi, .opencode, .qwen, .antigravitycli
```

Upstream ships the seven skill entrypoints as 43-byte symlinks to `.agents/skills/career-ops/SKILL.md`. A clone with `core.symlinks=false` checks each out as a text file holding that path, and `apply()` then materialises it into the real 10 972-byte document and commits it. `systemTreeDiffers` compares committed trees, so all seven report as drift on every check. Same symptom as `SIGNATURES.md`; entirely different reason.

## Why this PR's mechanism must not be reused for it

Dropping a path from `SYSTEM_PATHS` works here because nothing on an install reads `SIGNATURES.md`. The obvious next move for the entrypoints is the same trick — and I measured what it would do, by calling the real `ensureSkillEntrypoints` on a temp tree:

```
A. first apply                              entry = "SKILL v1"
B. still in SYSTEM_PATHS, upstream ships v2  entry = "SKILL v2"
C. dropped from SYSTEM_PATHS, upstream v3    entry = "SKILL v2"   canonical = "SKILL v3"
```

Not broken — **frozen**. `scaffolder/bin/skill-entrypoints.mjs:115` only rewrites an entry whose content is still the pointer:

```js
if (content !== entry.pointer) continue;
```

What resets it to the pointer on every update is `apply()` checking the path out *because it is in the manifest*. Take it out and a materialised entry keeps whatever document it last received, while `.agents/` keeps moving. Every CLI on a symlink-less install would load a stale skill, with no error and no drift report — the report being the thing that disappeared.

It only bites where the entries are materialised. On a filesystem with symlinks, `ensureSkillEntrypoints` sees `isSymbolicLink()` and leaves the link alone, and the link tracks canonical by construction. So the population that would lose its updates is exactly the one no CI runner represents.

The entrypoint cause needs the opposite shape: **stay in the manifest, excluded from the drift comparison.** `ensureSkillEntrypoints` already returns the list it materialised, so the set does not need re-deriving.

## What I would change

One line, not a code change: either drop the word "third", or name #3149's second cause beside the other two and add that it must **not** be fixed by leaving `SYSTEM_PATHS`. Either stops the comment reading as a complete inventory. Not a blocker.

## What I did not verify

- A live `apply()`. Case B above writes the pointer by hand to model the checkout; that step comes from a fresh `core.symlinks=false` checkout of the upstream ref, which I measured on 2026-09-05 (`120000` kept in the index, 43 bytes on disk). The freeze in case C is the function's own behaviour, not a model.
- Whether any symlink-capable install ends up with materialised files anyway — for example a repo copied across filesystems. If so, it freezes the same way.

maxmilian:
@artemtrofymenko — thank you for running it on Windows, and for measuring case C rather than reasoning about it. You're right: santifer confirmed the symlink-entrypoint cause on #3149 on 09-05, so "the third independent cause" reads as a complete inventory when it isn't — and it is exactly the comment someone would follow to the wrong fix. Reworded in 790b18c: it now names the entrypoint cause beside #2630 and #2756, and says that one must stay in `SYSTEM_PATHS` and be excluded from the drift comparison instead, because `ensureSkillEntrypoints` only refreshes an entry that still holds the pointer. Comment-only; the manifest change and the test are untouched.

santifer (approved):

santifer:
Merged. Thank you, @maxmilian: your change is now in `main`.

If this was your first contribution here, welcome aboard. If you want to keep going, `good first issue` and `help wanted` are curated for exactly that, and you can ask on any of them before starting.

PR #4075, opened 2026-09-11
fix(dashboard): stats panels overflow their column, and "evaluated" labels three different numbers

TapiocaQAQ:
## What does this PR do?

Fixes two rendering bugs on the dashboard's Stats and Progress screens. The stats
panels sized their columns off the terminal width while being rendered into a box
half that wide, so rows wrapped and lost their indent; and the word "evaluated"
labelled three different numbers, two of which appear on adjacent screens.

## Related issue

None — these are bug fixes, which CONTRIBUTING explicitly exempts from
issue-first.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

> On `test-all.mjs`: 8673 passed, 1 failed — `update-system.mjs check crashed
> (exit null, signal SIGTERM)`. That check makes a network call to GitHub and was
> killed by the harness timeout; running `node update-system.mjs check` on its own
> exits 0. This PR touches only `dashboard/internal/**` (five Go files, no Node),
> so it cannot be the cause. Flagging it rather than quietly ticking the box.
>
> `go test ./...` in `dashboard/` is green: 220 passed across 6 packages.

---

## What I hit

I run career-ops on a job search in Taiwan — 179 rows in the tracker, mostly
Chinese-language postings — and I use the Go dashboard to look at it. Two things
were wrong every time I opened it.

**The archetype chart falls apart in a normal terminal.** At 120 columns the top
three bars wrap, the wrapped half starts at column 0 with the panel indent gone,
and it collides with the pie charts in the right-hand column:

```
  By Archetype
  AI & ML Engineering         44 (26%)
█████████████████████████    ★ 3.6/5          <- indent gone, under the panel
  Agentic & Automation        38 (22%)
█████████████████████    ★ 3.2/5
  AI Solutions & FDE          16 (9%)  █████████    ★
3.3/5                                                 <- score split in half
```

**Two screens one keypress apart disagree.** Progress says `179 evaluated`, Stats
says `171 evaluated`. Same tracker.

## Why

**The layout bug.** `renderBody` places the panels inside lipgloss boxes roughly
half the terminal wide, but `renderArchetypeChart`, `renderLabelCountTable` and
`renderPay` sized their label and bar cells off `m.width` — the terminal, not the
box. An archetype row comes to a fixed 77 columns; the box is 53–76. Every
terminal from 110 to 157 columns is affected, and 120 is the common case.
`renderInsights` already computed `m.width/2 - 6` for exactly this reason, so the
mechanism was known — it just never reached the bar charts.

At 60 columns the same arithmetic makes the row 61 wide, so the *whole screen*
renders one column wider than the terminal and every line wraps.

**Why it survived this long.** It cannot be seen by measuring `View()`. lipgloss
pads each line back out to the terminal width, so the assembled screen always
measures exactly right. The only way to catch it is to measure a panel *before*
it goes into its box — which is what the new test does.

**The number bug.** Progress renders `FunnelStages[0].Count`, which is
`len(apps)`; Stats counts rows carrying a score. Mine differ because 8 rows were
backfilled without an evaluation (#1799). `Evaluated` is a canonical status value
as well, so the word stood for three different things. The Progress header also
contradicted itself — it pairs that count with `AvgScore`, which divides by the
scored rows only.

`computeFunnel()` in `stats.mjs`, described in its own docstring as the canonical
funnel definition, has no evaluated stage at all, and warns:

> Keys are deliberately NOT bare status names ... the same word for two different
> numbers would read as a bug.

## What I changed

**Renamed, did not renumber.** The top funnel stage becomes `Tracked` and
`ProgressSummary` reads `%d tracked`. **No counting logic changes**, so the Go
funnel stays in step with `stats.mjs`. Stats keeps `evaluated` for the number
that really is the evaluated count.

**Budget the cells against the panel column.** `contentWidth()` /
`rightColumnWidth()` give a panel its real width, and `barWidth()` clamps the bar
to what is left after the fixed cells, floored at one glyph.

Three related fixes fell out of writing the test:

- The pie chart's 17-column disc, gap and legend come to **55 columns that
  nothing can shrink**, so the two-column layout never fit at its old 110
  threshold — the pie wrapped into the panel below. Raised to 114.
- Label truncation guarded on **display width** but cut by **rune count**, so a
  CJK label at two columns per rune passed the guard untouched and overflowed.
  Replaced both hand-rolled copies with `ansi.Truncate` from the `x/ansi` package
  this package already imports in `viewer.go` — that removes ~20 lines.
- The archetype count cell is one column wider than its reserved width once a
  share rounds to 100% (a single archetype, or a first-run tracker). Cells are
  measured now, and the fixed widths are named so the bar budget and the
  `lipgloss.Width` call that renders the cell cannot drift apart.

## Tests

`stats_width_test.go` follows the invariant `pipeline_width_test.go` already
establishes for the pipeline screen — a rendered row must fit the width it was
given — and extends it to all six stats panels across 12 terminal widths.

It asserts **row counts as well as widths**. That is the load-bearing half:
wrapping makes lines *narrower*, so a width-only check passes straight through
the defect. Panels whose fixed content has a hard floor (the pie at 55 columns,
the archetype row at 47) are skipped below it rather than silently excluded.

Both tests fail against `main` and pass here.

## Compatibility

Checked by rendering both revisions against the same 179-row tracker and diffing
the output:

| terminal | vs `main` |
|---|---|
| 70–109 | **byte-identical** |
| 200 | **byte-identical** (the old arithmetic already fit) |
| 60 | fixed — the screen was 1 column wider than the terminal |
| 110 | now single-column; the two-column layout could not fit the pie here |
| 120–158 | fixed |

## One thing to check

The Turkish and Spanish `ProgressSummary` strings are mine and I am not a native
speaker of either. Turkish reuses `takip`, the root already in that screen's own
title (`TAKİP İLERLEMESİ`); Spanish uses `en seguimiento`. Happy to take any
correction.

## Not in this PR

- The funnel stage labels in `data/career.go` are hardcoded English and never go
  through `i18n`, so the Progress funnel stays English under `--lang tr`. Separate
  issue; I did not want to mix it in here.
- `renderInsights` reserves 6 columns in the single-column branch and 4 in the
  two-column one. One of them is wrong by 2, but fixing it changes rendering
  beyond this bug.
- `screens/stats.go:316` and `internal/data/stats.go` are not `gofmt`-clean on
  `main`. I left them alone to keep this diff focused.

## User-visible changes

- Stats panels now size labels, counts, and bars to their assigned panel columns. Narrow terminals no longer wrap rows or lose indentation.
- Pie-chart thresholds, CJK label truncation, and archetype count-cell sizing now use the available display width.
- The Progress funnel now labels its top stage `Tracked`. Its summary separates tracked applications from the scored applications counted in Stats.
- Added width and row-count coverage for six stats panels across 12 terminal widths.

## Files changed

- `dashboard/internal/data/career.go: ComputeProgressMetrics`: Renames the top funnel stage to `Tracked`.
- `dashboard/internal/data/progress_metrics_test.go`: Verifies tracked-row counting and scored-row averages.
- `dashboard/internal/i18n/catalog.go`: Updates English, Turkish, and Spanish progress summaries.
- `dashboard/internal/ui/screens/stats.go`: Adds panel-width helpers and display-width-safe row sizing.
- `dashboard/internal/ui/screens/stats_width_test.go`: Adds terminal-width and row-layout tests.

### DIFF
--- a/dashboard/internal/data/career.go
+++ b/dashboard/internal/data/career.go
@@ -992,8 +992,11 @@ func ComputeProgressMetrics(apps []model.CareerApplication) model.ProgressMetric
 	interview := statusCounts["interview"] + statusCounts["offer"] + statusCounts["hired"]
 	offer := statusCounts["offer"] + statusCounts["hired"]
 
+	// Top stage counts every tracked row, including rows backfilled without a
+	// score (#1799) — hence "Tracked", not "Evaluated", which already means both
+	// a status value and the Stats screen's scored count.
 	pm.FunnelStages = []model.FunnelStage{
-		{Label: "Evaluated", Count: total, Pct: 100.0},
+		{Label: "Tracked", Count: total, Pct: 100.0},
 		{Label: "Applied", Count: applied, Pct: safePct(applied, total)},
 		{Label: "Responded", Count: responded, Pct: safePct(responded, applied)},
 		{Label: "Interview", Count: interview, Pct: safePct(interview, applied)},
--- a/dashboard/internal/data/progress_metrics_test.go
+++ b/dashboard/internal/data/progress_metrics_test.go
@@ -35,7 +35,7 @@ func TestComputeProgressMetricsCountsHiredInEveryStage(t *testing.T) {
 		label string
 		want  int
 	}{
-		{"Evaluated", 1},
+		{"Tracked", 1},
 		{"Applied", 1},
 		{"Responded", 1},
 		{"Interview", 1},
@@ -98,5 +98,30 @@ func TestComputeProgressMetricsFunnelIsMonotonic(t *testing.T) {
 	// offer = offer+hired = 2
 	if offer != 2 {
 		t.Errorf("Offer = %d, want 2", offer)
+	}
+}
+
+// The top funnel stage counts every tracked row, scored or not — a row
+// backfilled without an evaluation (#1799) sits there with no score. So it must
+// not borrow the word the Stats header uses for the scored count, nor the
+// canonical "Evaluated" status: computeFunnel() in stats.mjs already warns that
+// the same word standing for two different numbers reads as a bug, and this
+// label made it three.
+//
+// The AvgScore assertion is the other half of the same defect: the Progress
+// header renders this stage's count and AvgScore side by side, but AvgScore
+// divides by the scored rows only. Naming the count honestly is what stops the
+// two halves of that line from contradicting each other.
+func TestFunnelTopStageCountsUnscoredRowsAndIsNotNamedEvaluated(t *testing.T) {
+	apps := appsWithStatuses("Applied", "Applied")
+	apps[0].Score = 4.2 // one evaluated, one backfilled with no evaluation
+
+	pm := ComputeProgressMetrics(apps)
+
+	if got := stageCount(pm, "Tracked"); got != 2 {
+		t.Errorf("Tracked stage = %d, want 2 — both rows are tracked even though only one carries a score", got)
+	}
+	if pm.AvgScore != 4.2 {
+		t.Errorf("AvgScore = %v, want 4.2 — the average covers scored rows only, which is why the count beside it cannot be called 'evaluated'", pm.AvgScore)
 	}
 }
--- a/dashboard/internal/i18n/catalog.go
+++ b/dashboard/internal/i18n/catalog.go
@@ -328,7 +328,7 @@ var En = Catalog{
 
 	// Progress screen
 	ProgressTitle:   "SEARCH PROGRESS",
-	ProgressSummary: "%d evaluated | %.1f avg score",
+	ProgressSummary: "%d tracked | %.1f avg score",
 	FunnelTitle:     "Pipeline Funnel",
 	ScoresTitle:     "Score Distribution",
 	RatesTitle:      "Conversion Rates",
@@ -492,7 +492,7 @@ var Tr = Catalog{
 
 	// Progress screen
 	ProgressTitle:   "TAKİP İLERLEMESİ",
-	ProgressSummary: "%d değerlendirildi | %.1f ort. puan",
+	ProgressSummary: "%d takipte | %.1f ort. puan",
 	FunnelTitle:     "Pipeline Hunisi",
 	ScoresTitle:     "Puan Dağılımı",
 	RatesTitle:      "Dönüşüm Oranları",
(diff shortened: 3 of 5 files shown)


Scott-Emberson (commented):
Verified the owned tests by execution and mutation. Sound (all changed files are under dashboard/).

`stats_width_test.go` drives the real render path, not reconstructed width math: it calls the production `renderArchetypeChart`, `renderLabelCountTable`, `renderPay`, and `renderPieChart`, sizes each panel against the same `contentWidth()` / `rightColumnWidth()` the real `renderBody` uses, and references the production width constants directly rather than re-declaring them. The width budget is pinned two ways: every rendered line must be within its column, and the line count must be unchanged, which is the load-bearing half because a wrapped remainder is what collides with the neighbouring pie column. I mutation-checked it: removing the width clamp reddens with the exact overflow ("archetypes line measured 67 columns, but its column is only 55 wide"), and reverting the display-width truncation to a rune-count cut reddens the CJK-truncation test. The paired `progress_metrics_test.go` addition is a genuine behavioral assertion (top stage counts all tracked rows while the average covers scored rows only), and the i18n catalog change only updates values of an existing struct field for the Evaluated-to-Tracked rename, so no key can break; go build and the package tests are green.

GATE-tier (dashboard/), owned tests sound, CI green, not behind. Good to go.

santifer:
Welcome, @TapiocaQAQ, and sorry for the week of silence: that wait was on us, not on you.

This is a strong first PR. The part that convinced me is the half you called load-bearing: measuring a panel before it goes into its box, and asserting the row count as well as the width, because wrapping makes lines narrower and a width-only check sails straight past the defect. I removed the bar clamp locally and your test reddens with the exact overflow, so it has real teeth.

Renaming rather than renumbering was right too. The top stage counts rows a score never touched, so borrowing the word the Stats header uses would have left one word standing for three numbers.

On the strings you flagged: the Spanish "en seguimiento" reads correctly and is what I would have written. I cannot vouch for the Turkish, so it stays as you wrote it unless a native speaker says otherwise.

CodeRabbit's note about stacking the pie legend below 55 columns is a separate follow-up, not a condition here.

Updating the branch and re-running CI is on me, and then it goes in.

santifer (approved):
Reviewed on this commit: the stats panels are budgeted against their own column instead of the terminal, the width test reddens when the bar clamp is removed, and the funnel rename is presentation-only with no consumer branching on the label. Merging once the required checks finish here.

PR #4077, opened 2026-09-11
feat(providers): add four multi-tenant ATS platform providers

kellymmcgrath:
## Problem

Each of these is a career-site **platform**, not one employer's board, so a single provider unlocks every tenant running it. Today a `portals.yml` entry pointing at one of these hosts is claimed by no provider and skipped silently by `scan.mjs` — it reads as coverage while scanning nothing, which is exactly the failure mode `audit-portals.mjs`'s `no-provider` verdict exists to surface.

| Provider | Platform | Endpoint shape | Boards |
|---|---|---|---|
| `oracle` | Oracle Recruiting Cloud (Fusion HCM `recruitingCE`) | REST on the Fusion **pod** host | American Express |
| `j2w` | SAP SuccessFactors "Jobs2Web" (RMK) | SSR HTML, `?startrow=N` | Paramount, Lionsgate |
| `careers-home` | careers-home (Jibe, fronting iCIMS) | `GET /api/jobs` JSON | AMD, Rivian |
| `phenom-widgets` | Phenom canvas sites | `POST /widgets`, `ddoKey: refineSearch` | Activision Blizzard, Warner Bros |

`phenom-widgets` complements the existing `phenom` provider, which handles the older site shape.

## Explicit-config, not URL-sniffing

`detect()` fires only on the entry's own key (`oracle:` / `j2w:` / `careershome:` / `phenom_widgets:`), never by inspecting `careers_url`.

For `oracle` that is forced: the public API lives on the Fusion pod host (`egug.fa.us2.oraclecloud.com`), not the vanity careers domain, so there is nothing in `careers_url` to detect. For the other three it is the safer default — these platforms serve behind per-customer vanity domains, so a URL heuristic would start claiming other vendors' boards. The tests pin that contract, since a later "helpful" refactor to sniff `careers_url` would regress it silently.

## Conventions these now follow

Three gaps against the existing providers, all closed here:

- **`max_pages` on the entry**, bounded by each provider's `PAGE_CAP`. Without it a large board truncated at the hardcoded cap with no way to raise it.
- **`ctx.maxPages`**, which is how `verify-portals.mjs` probes a board. Bounding pages alone was not enough for the three that loop *queries × pages* — a health check still fired one request per configured query — so a probe also collapses the query list to one. Answering "does this host respond?" costs one request now, not five.
- **`redirect: 'error'` on `oracle`'s fetch.** Its three siblings already refused redirects; `oracle` did not, so a misconfigured or compromised Fusion pod could bounce the scanner to an internal address and have the response parsed as job data.

## One paging helper, not one per provider

`providers/_paging.mjs` exports `resolveMaxPages()` and `isProbing()`. Both rules were already being re-implemented per provider — **19 copies of the cap resolver and four of the probe check** — which is how a newly written provider ends up silently ignoring `max_pages` or walking a whole board for a health check.

These four import the shared pair rather than adding a fifth and sixth copy. The 19 existing copies are deliberately left alone: rewriting them is a separate refactor, not something to fold into a PR that adds providers. Happy to send that separately if you want it.

## Tests

`tests/providers/{oracle,j2w,careers-home,phenom-widgets}.test.mjs` (28), following the existing per-provider convention: `id`, `detect()` positive, `detect()` ignores `careers_url`, `detect()` null without config, `fetch()` callable, the redirect guard, and the one-request health probe.

The `ctx.maxPages` gap only surfaced because the probe assertion was written first and failed at 5 requests — the initial fix had bounded the wrong dimension.

## Docs

Four rows in `docs/SUPPORTED_JOB_BOARDS.md` with each platform's endpoint shape, config key and page cap.

Full suite: **8724 passed, 0 failed.**

## Summary

Users can configure four explicit ATS providers:

- `oracle`: Oracle Recruiting Cloud REST API (`providers/oracle.mjs`)
- `j2w`: SAP SuccessFactors Jobs2Web (`providers/j2w.mjs`)
- `careers-home`: Jibe/iCIMS (`providers/careers-home.mjs`)
- `phenom-widgets`: newer Phenom canvas sites (`providers/phenom-widgets.mjs`)

Each provider requires its matching configuration block and `host`. Existing `careers_url` entries do not activate these providers. Each provider supports bounded pagination and one-request health probes through `ctx.maxPages`. Outbound requests reject redirects.

The supported platforms are documented in `docs/SUPPORTED_JOB_BOARDS.md`.

## Files touched

- `providers/`: four providers and shared pagination helpers
- `tests/providers/`: provider contract tests
- `docs/SUPPORTED_JOB_BOARDS.md`: supported platform documentation

The change does not touch `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, or `.github/`.


Scott-Emberson (changes requested):
The provider code here is mostly in good shape: careers-home, j2w, and phenom-widgets each parse a well-formed job over a real fixture at runtime, the SSRF posture is sound for what these are (a full host taken from trusted portals.yml config rather than untrusted posting data, HTTPS-only, `redirect: 'error'` on every fetch, encodeURIComponent on the data-derived query, and response-derived job URLs stored rather than fetched), and j2w's scraped href is anchored to a leading-slash path so it cannot inject a host. Two things need fixing before this lands, both about the owned tests not catching what they should.

**1. oracle.mjs is non-functional at runtime, and its owned test hides it. `providers/oracle.mjs:58-78`, blocker.** `fetchQuery()` is a module-scope function that reads `pageCap` at lines 60 and 78, but `pageCap` is a local declared inside `fetch()` (line 96), not a parameter and not the module const `PAGE_CAP`. So every real, non-probe `fetch()` throws `ReferenceError: pageCap is not defined` before it makes a single request. I confirmed it: the fetch throws with `calls === 0`, so the parse code is never reached and the provider returns nothing to a scan. The owned test passes 7 of 7 anyway because its only fetch-driving assertion wraps the call in `.catch(() => [])` and then asserts `pagesRequested <= 1`, which is trivially true at zero. The fix is a one-liner, pass `pageCap` into `fetchQuery` (or read the module `PAGE_CAP`), but the test needs to change too so this cannot happen again (see 2).

**2. None of the four owned tests assert parsed output, so the parser is untested. `tests/providers/{careers-home,j2w,oracle,phenom-widgets}.test.mjs`, blocker.** All four are the same template that drives `provider.fetch` only for the maxPages probe with a mock returning an empty body, and none feeds a realistic payload through the parser and checks the jobs that come back. I verified the gap: breaking the core parse in careers-home (`json?.jobs` to a bogus key) and in j2w (the link regex) leaves both suites green. A single fixture-driven assertion per provider (feed a representative response, assert the extracted title/company/url) would close this, and it is exactly what would have caught the oracle ReferenceError.

Two smaller notes, not blocking. The redirect guard is currently pinned only by a source-text match on `redirect: 'error'`, so a future fetch path that keeps the literal but reintroduces a follow would slip; a behavioral spoof test (a non-throwing mock asserting the guarded host is never fetched) is the stronger form. And the test header comments describe these as multi-tenant SSRF-hardened, which over-claims, since the host is trusted config and there is no tenant-to-host construction or host allowlist here; wording it as trusted-config would be more accurate.

The shared `providers/_paging.mjs` is a clean additive pair of pure helpers; fine as is, though the min/cap arithmetic is only covered indirectly.

Fix the oracle scope bug and give each owned test a real parse assertion, and this is close.

kellymmcgrath:
Closing this — all four providers duplicate modules career-ops already ships. That is my mistake: I wrote them without checking upstream first.

| this PR | already upstream | same endpoint |
|---|---|---|
| `oracle` | `oraclecloud.mjs` (#1929) | `recruitingCEJobRequisitions` |
| `phenom-widgets` | `phenom.mjs` | `POST {origin}/widgets` refineSearch |
| `careers-home` | `jibeapply.mjs` (branded `api:`) | `/api/jobs` |
| `j2w` | `successfactors.mjs` | SuccessFactors RMK (ex-jobs2web) |

I ran both implementations against the five boards these serve and compared by **posting id**, not counts:

| company | this PR | upstream provider | postings only this PR found |
|---|---|---|---|
| American Express | 380 | `oraclecloud` 399 | 48 |
| Activision Blizzard | 58 | `phenom` 58 | 0 |
| AMD | 167 | `jibeapply` 1224 | 0 |
| Paramount | 232 | `successfactors` 295 | 0 |
| Lionsgate | 24 | `successfactors` 35 | 0 |

The upstream providers walk the whole board rather than a handful of keyword queries, so they are a superset. They also carry `PostedDate`, which my `oracle` never captured.

The Amex column of 48 was not a gap in `oraclecloud` coverage but a **bug in it**, which this comparison surfaced: it stops on a short middle page and drops 12% of that board. Fixed separately in #4296.

Two of the four also had defects of their own: `j2w` glued the location into titles ("Santa Monica Coordinator, Content Strategy … CA 90404") and left `&amp;` undecoded in URLs.

@Scott-Emberson — thanks for the review here. The `oracle` scope bug you caught (`pageCap` out of scope, provider made zero requests) is moot now, but it is what sent me to check the rest, so it paid off. My portals config now points at the four upstream providers.

PR #4078, opened 2026-09-11
feat(providers): add apple, atlassian and google (robots-compliant)

kellymmcgrath:
Reshaped from the original seven after a compliance pass on each source. **Depends on #4076** (the three providers populate `externalId`, which that PR adds to `Job`).

## Four providers left the branch, on their sources' own stated rules

| provider | why |
|---|---|
| **meta** | `metacareers.com/robots.txt` opens with *"Collection of data on Facebook through automated means is prohibited unless you have express written permission"*. There is a technically cleaner route (the declared `/jobsearch/sitemap.xml` plus schema.org JSON-LD on each posting, 990 postings and a real `datePosted`), but it obeys the robots RULES while still ignoring that NOTICE — so there is no compliant version of this provider to submit. |
| **uber** | `www.uber.com/api/*` is disallowed **and** now 301s to `jobs.uber.com`, so the shipped provider was already dead. The new host allows everything and publishes `/en/jobs/sitemap.xml` (517 postings with `lastmod`), but its posting pages are challenge-protected, so no titles are reachable — and `scan.mjs:3118` filters on `job.title`. |
| **whatnot** | The board is an Ashby board. It belongs in `providers/ashby.mjs`, not in a per-company provider. |
| **pcsx** | Eightfold's newer API, on the same `*.eightfold.ai` host `eightfold.mjs` already pins. It belongs there as a capability switch, not as a second provider for one vendor. |

## google is first-page-only now

@Kevin-Boyle was right: every URL the provider built carried the disallowed `page=` parameter, including the `verify-portals` probe. It now sends none, adds `sort_by=date` so the single page is the newest postings, and documents that `max_pages` does not apply. Same reason career-ops declines Lagou — the project does not route around a stated restriction.

A test asserts no request ever contains `page=`, and another asserts `max_pages: 25` still produces one request per query.

## apple, brought to `ADDING_A_PROVIDER.md`

- `max_pages` on the **entry** with a hard ceiling and a truncation warning (was `cfg.max_pages`, uncapped)
- honours `ctx.maxPages`: one request for a probe, not queries × teams
- a missing hydration blob **throws** instead of returning zero jobs forever
- `fetchTextWithRetry` + a 200 ms inter-page delay, replacing a bare fetch whose only error path was a silent `return`
- **no hardcoded Chrome User-Agent.** Measured: `jobs.apple.com` returns the same 313,692-byte page carrying the same blob to career-ops' own UA, to a bare `node`, and to that Chrome string. The masquerade bought nothing and cost the identity RFC 9110 §10.1.5 and RFC 9309 §2.2.1 expect a crawler to present.

## atlassian stops inventing a posting date

The feed's only date is `portalJobPost.updatedDate` — when the posting was last **edited**. `Job.postedAt` is the publication date (schema.org `datePosted`), and `scan.mjs:555` uses it for the max-age filter, so an old role edited yesterday read as new. The value is also zone-less (`"2026-09-11 01:00 AM"`), which `Date.parse` reads in local time: a 16-hour spread between `TZ=UTC` and `TZ=Asia/Tokyo` on the same string. `postedAt` is now omitted, as in `successfactors.mjs` — absent means "unknown", never "stale". A non-array feed body throws and names the keys it got.

## New: `tests/provider-user-agent-convention.test.mjs`

Bans a `Mozilla/5.0` literal anywhere under `providers/` (importing the shared `BROWSER_LIKE_USER_AGENT` stays fine — several providers legitimately need it behind a WAF). Same idiom as `tests/main-guard-convention.test.mjs` (#3170): ban the raw ingredient, not the recipe. `main` already passes it; only this branch's apple did not.

## Tests

Every provider assertion now drives `fetch()` against a recording context and checks literal fields, request counts and request options — the source greps are gone. Mutation-verified: **12 of 12 mutations go red** across the three providers, plus the UA convention test (checked by hardcoding a UA into apple and watching it fail).

Full suite: 9245 passed, 0 failed.

## Summary

Adds seven job-board providers in `providers/`: Apple and Google use HTML parsing. Uber, Meta, Atlassian, Whatnot, and PCSX use APIs.

Adds redirect protection, exact host detection, pagination limits, and 43 provider tests.

Adds normalized `externalId` and `requisitionId` fields for ATS results. See `providers/_types.js:57-62` and `providers/_ids.mjs:20-26`.

## User impact

Users can scan seven additional job-board sources.

Google scans one newest-results page per query because Google disallows the `page` parameter. See `providers/google.mjs:16-27`.

Provider consumers can use stable posting and requisition identifiers for deduplication and grouping.

## System files

The repository contains `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/`. The supplied changed-path output does not identify changes to `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, or `.github/`.

The reported changes are in `providers/`, tests, and `docs/SUPPORTED_JOB_BOARDS.md`.

## Verification

The supplied summary reports 8,745 passing tests. A Windows-only lock timing failure remains reported in `tests/portal-health-lock.test.mjs`.


Scott-Emberson (changes requested):
The provider code across these seven is mostly sound: apple, atlassian, uber, and whatnot pin a hardcoded host with an exact-compare guard, google and uber build from a compile-time constant, meta guards its tenant-derived origin with an ALLOWED_HOSTS allowlist in both resolveConfig and assertMetaUrl, and every provider sets `redirect: 'error'` and encodes its data-derived query. The problem is the owned tests, which do not exercise the two things that matter most for a net-new scraper, so a real regression in either would ship green.

**1. None of the seven tests drive fetch or parse. `tests/providers/{apple,atlassian,google,meta,pcsx,uber,whatnot}.test.mjs`, blocker.** Each test imports the real provider (good), but only calls `detect()`, checks `typeof provider.fetch === 'function'`, and greps the source for `redirect: 'error'`. There are no fixtures, no mock ctx, and no call into any parser. I mutation-checked the consequence: disabling `assertMetaUrl`'s host allowlist stays green, disabling `assertAppleUrl` stays green, and nulling the parsers (apple `parseHydration`, meta `parseLsdToken`, pcsx `toJob`) all stay green. So the fetch-time SSRF host guard and all parse logic are currently untested, and the `redirect` guard is pinned only by the text grep, which a new fetch path could keep while reintroducing the hole.

**2. meta is the one to prioritize. `providers/meta.mjs`, major (security).** Of the seven, meta and pcsx build the request host from tenant config (`entry.api` / `careers_url` for meta, `cfg.host` for pcsx), which is the actual SSRF surface; the other five use a hardcoded host and are low-risk. meta defends itself with the allowlist, but that guard has zero test coverage, so a silent removal of it would not be caught. A spoof test here matters more than anywhere else in the batch.

The fix is the same per provider and small: construct a mock ctx, call `fetch` with a spoofed off-allowlist host, and assert the guarded fetch is never made (a call count of zero via a non-throwing mock), then feed one realistic fixture through the exported parser and assert the extracted fields. For the hardcoded-host providers the spoof test is quick; for meta and pcsx it is the important one. Worth noting your own #4076 did exactly this (real fixtures, real parse, mutation-sensitive), so the pattern is already in the repo to copy.

Smaller note: the detect coverage is uneven. apple, atlassian, uber, and whatnot each assert a host-hijack rejection, but google, meta, and pcsx do not, and meta never tests its accept path (a `www.metacareers.com` careers_url) at all.

Add the mock-ctx fetch/parse tests (the meta host-guard spoof first) and this batch is in good shape; the provider code itself is close.

Kevin-Boyle:
One blocking issue on `providers/google.mjs`: every request it makes is disallowed by Google's `robots.txt`.

The `User-agent: *` group of https://www.google.com/robots.txt (currently lines 202–205) contains:

```
Disallow: /about/careers/applications/jobs/results?page=
Disallow: /about/careers/applications/jobs/results/?page=
Disallow: /about/careers/applications/jobs/results?*&page=
Disallow: /about/careers/applications/jobs/results/?*&page=
```

That group has no `Allow` rule under `/about/careers` that could take precedence.

At `3f23949`, `fetch()` builds

```js
`${BASE}?q=${encodeURIComponent(q)}&location=${encodeURIComponent(location)}&page=${page}`
```

with `BASE` ending in `/jobs/results/` and `page` starting at 1. Every URL therefore has the shape `/jobs/results/?q=…&location=…&page=N`, which the last rule matches. That includes the very first request of every query, and the single-query `verify-portals` probe.

The project already treats this as disqualifying. `docs/SUPPORTED_JOB_BOARDS.md` declines Lagou because its `robots.txt` disallows the query URLs it would need: "The project does not route around that stated restriction."

Because the rules only target `page=`, there's a compliant option: send no `page` parameter at all, so each query reads only the first results page (~20 postings). Adding `sort_by=date`, which the results page accepts, makes that first page the newest postings, which suits a scanner that runs regularly. Pages beyond the first aren't reachable without the disallowed parameter, so `PAGE_CAP` / `max_pages` wouldn't apply to this provider.

kellymmcgrath:
@Kevin-Boyle — the robots.txt block is fixed the way you proposed: no `page` parameter at all, `sort_by=date` so the one reachable page is the newest postings, and `max_pages` documented as not applying. A test now asserts no request ever carries `page=`.

@Scott-Emberson — the branch is reshaped: meta, uber, whatnot and pcsx are gone (each for a reason in the description, mostly their sources' own stated rules), apple and atlassian are brought to `ADDING_A_PROVIDER.md`, and the tests are behavioural rather than source greps, mutation-verified at 12/12. It is now based on #4076 because the three providers populate `externalId`.

Issue #3810, opened 2026-09-04
feat(web): onboarding never fills modes/_profile.md — proposal: confirm-gated personalization + a live progress panel


Rkrijn:
## Problem

`doctorState()` lists four user-layer prerequisites: `cv.md`, `config/profile.yml`, `portals.yml`, `modes/_profile.md`. The web's conversational onboarding can already produce three of them without the user touching YAML — `cv.md` through the CV editor, `config/profile.yml` + `portals.yml` through the confirm-gated `setProfile`. The fourth never gets filled from the UI.

`modes/_profile.md` only ever receives appended `<!-- co-web-notes -->` facts (`/api/memory`). Its eight structured sections — Target Roles (archetypes), Adaptive Framing, Exit Narrative, Cross-cutting Advantage, Portfolio / Demo, Comp Targets, Negotiation Scripts, Location Policy — stay byte-identical to `modes/_profile.template.md`. Two effects:

- `doctor.mjs` keeps warning *"modes/_profile.md still identical to the shipped template — evaluations score against the template author's targeting, not yours"*, and every evaluation really does score against the example archetypes.
- Once a single `remember` fact is appended, the file is no longer identical, so that warning silently disappears while the archetypes are still the template's. The onboarding "feels done" but the brief is still generic.

The CLI `intake` mode covers this (the agent edits the file directly), so web-only users are the ones left with a generic brief — and the web README's promise is *"no YAML to edit — answer in plain language and the assistant writes it for you."*

## Proposed solution

Close the gap in the same pattern `setProfile` uses — propose → confirm card → merge-safe write — plus a visible "profile that grows while you talk":

1. **`lib/personalization.mjs`** (pure, unit-tested): coerce a proposal, render each section to markdown, `mergeSections()` that replaces **only** the bodies of the sections proposed (keeps every other byte, including the `co-web-notes` block and hand edits; appends a missing heading before the notes), and `sectionState()` = filled / template / missing per section against the shipped template.
2. **`api/personalization`**: `GET` section state; `POST` section-safe `atomicWriteWithBackup`, seeded from the template on first create; `?dryRun=1` returns the exact merged bytes without touching disk.
3. **`setPersonalization` action** in `actions/registry.ts`: a confirm card like `setProfile`; nothing is written the user did not approve.
[...]

PR #4082, opened 2026-09-11
fix(web): stale inherited console kills child spawns on Windows — 4 execFile sites + honest failure handling

### LINKED ISSUE #2375: bug(web): npm-installed CLIs can't be spawned on Windows — every CLI-backed web feature fails with spawn EINVAL
Every CLI-backed web feature (evaluate, assistant, AI search, CV ingest) fails on a Windows machine whose CLI came from npm, with `error launching Claude Code: spawn EINVAL`.

## Root cause, two layers

**1. `.cmd`/`.bat` can no longer be spawned directly.** Since the CVE-2024-27980 patches (Node 18.20.2 / 20.12.2, and all of 20+ since), `child_process.spawn()` of a `.cmd` or `.bat` without `shell: true` throws `EINVAL`. The comment in `web/src/lib/clis.ts` `binCandidates()` says the opposite ("Only include extensions that `child_process.spawn()` can execute directly" while listing `.bat`/`.cmd`), which I verified on Node 22: `.cmd err: EINVAL`.

**2. The unspawnable candidate wins the search.** npm on Windows installs three shims in `%APPDATA%\npm`: `claude` (sh script), `claude.cmd`, `claude.ps1`. `binCandidates()` tries the bare name first, and `fs.accessSync(p, X_OK)` is a no-op check on Windows, so `findBin` returns the extensionless sh shim, which `spawn()` can't execute either. Even if it skipped that, the next hit is the `.cmd`.

Native installs (`claude.exe`) are unaffected, which is probably why this survived: it only bites npm-installed CLIs, and all of the `KNOWN` list except Antigravity commonly arrives via npm on Windows.

`batch-tailor.mjs` shares the root cause at its `spawnSync('claude', ...)` call (plus a cwd-relative `--append-system-prompt-file modes/pdf.md` while its state paths are `__dirname`-anchored, so it only works when run from the repo root).

## Fix options, in the order I'd consider them

1. **Resolve through the shim to the real entry.** For an npm shim, `%APPDATA%\npm\node_modules\<pkg>\<bin>.js` exists and can be spawned as `process.execPath [cli.js, ...args]` with no shell involved and no quoting problem. Most robust, slightly more code.
2. **Spawn `.cmd` via `{ shell: true }`** and escape args. The prompt text passed to these CLIs is arbitrary user text, so the escaping has to be airtight (cmd.exe quoting, `%VAR%` expansion). Easy to get subtly wrong.
3. **Take `cross-spawn` as a dependency** in web/. It solves exactly this and is battle-tested, but it's a new dependency and the escaping caveats for cmd shims still apply to argument content.

Happy to send a PR once there's a preferred direction. Flagging the analysis first since options 2 and 3 have injection-shaped edges and option 1 changes how every spawn site resolves its binary.

### LINKED ISSUE #3809: web: a stale inherited console kills every child spawn with 0xC0000142 — /api/doctor then reports "fully set up"
## Summary

On Windows, once a long-running `next dev` process's inherited console goes stale, **every** child it spawns dies at DLL initialization with exit `3221225794` = `0xC0000142` (`STATUS_DLL_INIT_FAILED`) in ~20 ms, with no stdout or stderr. That takes down every CLI-backed web feature — and, worse, `/api/doctor` reads the empty stdout as `{}` and reports `onboardingNeeded: false, missing: []`, so a user with **no `cv.md`** is told they are fully set up and never sees the onboarding banner.

This is distinct from #2375 (npm shims): it hits `node.exe`, `cmd.exe` and `where.exe` alike, and it hits `process.execPath` too.

## Environment

- Windows 11 (10.0.26200), Node v25.3.0, Next 16.3.3 (Turbopack), `web` 0.10.0 on `career-ops-v1.32.0-2-g5c8b02a`
- `next dev` started via `npm run dev` from PowerShell; process had been up ~21 h when measured

## Reproduction (measured from inside the Next process)

A temporary API route ran `spawn()` variants from within the affected server. Every child that inherited the parent's console failed identically; only the two that don't attach to it succeeded:

| spawn variant (from inside `next dev`) | result |
|---|---|
| `node -e …` (default) | `0xC0000142` |
| `node -e …` with `stdio: "ignore"` | `0xC0000142` |
| `node -e …` with `SystemRoot`/`windir` forced in env | `0xC0000142` |
| `node -e …` via `shell: true` | `0xC0000142` |
| `C:\WINDOWS\System32\cmd.exe /c echo HI` | `0xC0000142` |
| `C:\WINDOWS\System32\where.exe node` | `0xC0000142` |
| `node -e …` with **`windowsHide: true`** | **exit 0, `OK`** |
| `node -e …` with **`detached: true`** | **exit 0, `OK`** |

Same binaries, same env and cwd, run from a fresh shell: all succeed. The parent's environment was sane (`SystemRoot`, `ComSpec`, `PATHEXT` present; `NODE_OPTIONS` = `--max-old-space-size=8047 --enable-source-maps`); bare `node` resolved to `C:\Program Files\nodejs\node.exe`; env block ~10 KB. Two Turbopack pool workers spawned at server start (when the console was still fresh) were alive and fine.

## Root cause

A Windows child normally attaches to its parent's console. When that console handle has gone stale on a long-lived parent, the child's console initialization fails during process start → `STATUS_DLL_INIT_FAILED`, before any user code runs. `windowsHide: true` (`CREATE_NO_WINDOW`) and `detached: true` never attach, so they are immune. Nothing in `web/src` passes `windowsHide` today.

## Consequences observed

- `/api/doctor`: `execFile("node", [doctor, "--json"], …, (_err, out) => resolve(out || ""))` → `""` → `JSON.parse("{}")` → `{ available: true, onboardingNeeded: false, missing: [] }`. **"Doctor didn't answer" is reported as a clean bill of health**, and `OnboardingBanner` (which reads this route) hides. Related in spirit to #3401 / #3028.
- `/api/followups`, `/api/followups/cadence`: same pattern → empty data.
- `/api/assistant`, `/api/cv/ingest`, evaluate/AI search: "no output — is the CLI installed and authenticated?" although the CLI runs fine from a shell.

## Fix (verified in-process, on the still-affected server)

1. `spawnHeadlessCli` passes `windowsHide: true` on every spawn (no-op on POSIX). Nothing spawned by the web needs a console.
2. `api/doctor`, `api/followups`, `api/followups/cadence`: `process.execPath` instead of bare `"node"`, `windowsHide: true`, and an `err` or empty stdout surfaces as `available: false` — never as `onboardingNeeded: false`.

After (1)+(2), `/api/doctor` on the very same stale process returned `onboardingNeeded: true, missing: ["cv.md","config/profile.yml","portals.yml"]` on the first request. Web suite 483/0, typecheck and production build clean. Happy to open the PR (it composes with the #2375 shim resolution, which I also have implemented against the same spawn boundary).

## Workaround for users

Restart the dev server. It works until the console goes stale again.

Rkrijn:
Fixes #2375 (npm shims unspawnable) AND #3809 (stale console crash). Same spawn boundary, two independent bugs that were both blocking web features on Windows.

## Changes

1. **resolve-npm-shim.mjs + spawn-cli.mjs + tests** (#2375 resolution): npm-installed CLIs resolve to bare .sh/.cmd/.ps1 shims that spawn() cannot run. Shim → real binary path at the spawn boundary.

2. **Four execFile sites (doctor, followups, cadence, portals/verify) + honest failure** (#3809 resolution): 
   - `process.execPath` not bare `"node"`
   - `windowsHide: true` (never inherit a stale console on Windows)
   - Surface spawn error or empty output as `available: false` instead of fake success

Both patterns match existing code (e.g. apply/drive.ts uses --resume for session reuse). Tests: web 525/0, tsc, production build clean.

## Related

- Adopted #2626 shim pattern (co-authored)
- Fixes #3809 (this PR)
- Addresses nikitacometa's comment on #3809 (fourth execFile site)

## Summary

This change improves Windows reliability and adds persistent assistant onboarding.

### User-visible changes

- npm-installed CLIs now run on Windows without `spawn EINVAL`, `ENOENT`, or `EFTYPE`: `web/src/lib/resolve-npm-shim.mjs:1`, `web/src/lib/spawn-cli.mjs:1`
- CLI child processes use `process.execPath` and hidden Windows consoles: `web/src/app/api/doctor/route.ts:19`, `web/src/app/api/followups/route.ts:21`, `web/src/app/api/followups/cadence/route.ts:20`, `web/src/app/api/portals/verify/route.ts:24`
- Failed or empty CLI responses now report `available: false`. Doctor no longer reports successful setup when its child process fails: `web/src/app/api/doctor/route.ts:25`
- Users can save, rename, delete, and resume assistant chats: `web/src/app/api/chats/route.ts:1`, `web/src/app/api/chats/[id]/route.ts:1`, `web/src/components/assistant-console.tsx:1`
- The assistant can update structured personalization sections: `web/src/app/actions/registry.ts:1`, `web/src/lib/personalization.mjs:1`
- Onboarding now shows live progress for CV, configuration, portals, and personalization: `web/src/components/home/onboarding-progress.tsx:1`

### System files touched

The provided change list does not include `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, or `.github/`.


Rkrijn:
Closing in favour of #4253.

@coderabbitai's out-of-scope check was right: this branch was my fork's `main`, so it carried unrelated chat-history and personalization work alongside the Windows fixes. #4253 is the same two fixes on a dedicated branch off current `main` — 7 files, no feature changes. The feature work will come back separately against #3810 and an issue of its own.

Also carried over into #4253:
- @chipoto69's `path.win32` diagnosis and fix, credited in the commit — that was the Linux/macOS CI failure, and the reason only the Windows job passed.
- @nikitacometa's fourth `execFile` site (`portals/verify`), which calls `execFile` directly and so was not reached by the spawn-boundary fix.
- The review findings on the resolver: absolute-path normalisation before the directory-escape check, and an escape-test fixture that was passing for the wrong reason (its crafted target resolved one directory above the planted file, so `exists()` refused it before the guard ran).

One correction to this PR's description, which was mine and wrong: it said the shim pattern was adopted from #2626. It was not — #2626 takes a different approach (`cli-launch.mjs` called from each route) and I wrote this independently against the existing spawn boundary. #4253 states that relationship accurately.

PR #4088, opened 2026-09-11
fix(tracker): resolve tracker paths via resolveTrackerPath(DATA_ROOT) across CLI tools (#4087)

### LINKED ISSUE #4087: bug(scan, reply-watch): CLI tools bypass resolveTrackerPath() / CAREER_OPS_DATA_DIR
### Code of Conduct

### Existing issues

### What happened?

Several CLI scripts hardcode `data/applications.md` or pass `__dirname` to `resolveTrackerPath()`, which breaks setups using `CAREER_OPS_TRACKER`, `CAREER_OPS_DATA_DIR` / `CAREER_OPS_ROOT` (split checkouts), or the flat root `applications.md` layout:

1. **`scan.mjs:110`**:
   `SCAN_HISTORY_PATH` and `PIPELINE_PATH` respect environment overrides, but `APPLICATIONS_PATH` hardcodes `path.join(DATA_ROOT, 'data/applications.md')` instead of using `resolveTrackerPath(DATA_ROOT)`. Because `loadDedupSnapshot()` uses `APPLICATIONS_PATH`, deduplication against already-applied/tracked roles falls back to an empty tracker whenever `CAREER_OPS_TRACKER` is set or when using the flat root layout.

2. **`reply-watch.mjs:26-29`**:
   `reply-watch.mjs` passes `__dirname` (code repository) to `resolveTrackerPath()` instead of `getCareerOpsRoot()`. It also anchors `reply-candidates.json` and `follow-ups.md` to `__dirname`. Under a split checkout (`CAREER_OPS_DATA_DIR=...`), it looks in the code repository and fails to find the tracker.

3. **`paste-reply.mjs:53`**:
   Anchors `data/reply-candidates.json` to `__dirname` instead of `getCareerOpsRoot()`.

4. **`linkedin-join.mjs:86`**:
   Hardcodes `join(DATA_ROOT, 'data/applications.md')` instead of calling `resolveTrackerPath(DATA_ROOT)`.

5. **`rejection-latency.mjs:67`**:
   Uses an ad-hoc existence check for `data/applications.md` / `applications.md` that misses `CAREER_OPS_TRACKER` and canonicalization.

### Steps to reproduce

1. Set `CAREER_OPS_TRACKER=/path/to/custom-tracker.md` (or set `CAREER_OPS_DATA_DIR` to a directory containing a flat `applications.md`).
2. Inspect `APPLICATIONS_PATH` in `scan.mjs`.
3. Observed: `APPLICATIONS_PATH` ignores `CAREER_OPS_TRACKER` and points to `<root>/data/applications.md`.

### Expected behavior

All CLI tools should resolve the tracker path via `resolveTrackerPath(DATA_ROOT)` from `path-resolver.mjs` and anchor their user-data artifacts to `getCareerOpsRoot()`.

### CLI tool

Claude Code

### OS and Node.js version

Windows 11 / Node v22.22.2

---

I have a fix with regression tests ready and will open a PR shortly.

SparshGarg999:
## Problem

Several CLI tools hardcode `path.join(DATA_ROOT, 'data/applications.md')` or pass `__dirname` to `resolveTrackerPath()`, which causes them to ignore `CAREER_OPS_TRACKER` overrides and break under split-checkout (`CAREER_OPS_DATA_DIR` / `CAREER_OPS_ROOT`) or flat root (`applications.md`) layouts (#4087):

- **`scan.mjs`**: `APPLICATIONS_PATH` hardcodes `data/applications.md`. When a user configures `CAREER_OPS_TRACKER` or uses a flat layout, `loadDedupSnapshot()` falls back to an empty tracker string, causing job deduplication against existing applications to fail silently.
- **`reply-watch.mjs`**: passes `__dirname` (code repository) to `resolveTrackerPath()` instead of `DATA_ROOT`, and anchors `data/reply-candidates.json` and `data/follow-ups.md` to `__dirname`. Under `CAREER_OPS_DATA_DIR`, it fails to find the tracker.
- **`paste-reply.mjs`**: anchors `data/reply-candidates.json` to `__dirname` instead of `DATA_ROOT`.
- **`linkedin-join.mjs`**: hardcodes `data/applications.md` rather than using `resolveTrackerPath(DATA_ROOT)`.
- **`rejection-latency.mjs`**: hand-rolls an ad-hoc layout check that misses `CAREER_OPS_TRACKER` and canonicalization.

## Fix

1. In `scan.mjs`, resolve `export const APPLICATIONS_PATH = resolveTrackerPath(DATA_ROOT)`.
2. In `reply-watch.mjs`, resolve `APPS_FILE = resolveTrackerPath(DATA_ROOT)`, and anchor `DEFAULT_CANDIDATES_PATH` and `FOLLOWUPS_FILE` to `DATA_ROOT`.
3. In `paste-reply.mjs`, anchor `CANDIDATES_PATH` to `DATA_ROOT`.
4. In `linkedin-join.mjs`, resolve `TRACKER_PATH = resolveTrackerPath(DATA_ROOT)`.
5. In `rejection-latency.mjs`, resolve `DEFAULT_TRACKER_PATH = resolveTrackerPath(DATA_ROOT)`.
6. Add `tests/cli-tracker-path-resolution.test.mjs` asserting that:
   - `scan.mjs` respects `CAREER_OPS_TRACKER`.
   - `scan.mjs` resolves flat `applications.md` under `CAREER_OPS_DATA_DIR`.
   - `reply-watch.mjs` discovers the tracker under `CAREER_OPS_DATA_DIR` without `CAREER_OPS_TRACKER`.
   - `linkedin-join.mjs` passes self-test with `resolveTrackerPath`.

Closes #4087

## Summary

Users can use `CAREER_OPS_TRACKER`, `CAREER_OPS_DATA_DIR`, `CAREER_OPS_ROOT`, split checkouts, and flat `applications.md` layouts without tracker discovery failures.

The CLI tools now resolve trackers with `resolveTrackerPath(DATA_ROOT)` and user-data files with `getCareerOpsRoot()`:

- `scan.mjs:91-110`
- `reply-watch.mjs:21-30`
- `paste-reply.mjs:49-54`
- `linkedin-join.mjs:81-86`
- `rejection-latency.mjs:60-67`

Before this change, these configurations could make tools read the wrong tracker or fail to find user-data files.

Tests in `tests/cli-tracker-path-resolution.test.mjs` cover tracker overrides, flat layouts, split-checkout discovery, and `linkedin-join.mjs --self-test`.

No changes touch `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, or `.github/`.

### DIFF
--- a/linkedin-join.mjs
+++ b/linkedin-join.mjs
@@ -83,7 +83,7 @@ import { getCareerOpsRoot, resolveTrackerPath } from './path-resolver.mjs';
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
 const DATA_ROOT = getCareerOpsRoot();
 const DEFAULT_CSV = join(DATA_ROOT, 'data/Connections.csv');
-const TRACKER_PATH = resolveTrackerPath(DATA_ROOT);
+export const TRACKER_PATH = resolveTrackerPath(DATA_ROOT);
 const PORTALS_PATH = join(DATA_ROOT, 'portals.yml');
 const CONTACTS_PATH = join(DATA_ROOT, 'data/contacts.tsv');
 
--- a/linkedin-join.mjs
+++ b/linkedin-join.mjs
@@ -78,12 +78,12 @@ import { resolveColumns, parseTrackerRow, normalizeTextKey } from './tracker-par
 import { asciiFold } from './lib/ascii-fold.mjs';
 import { flagValue, hasFlag, validateFlags } from './lib/cli-flags.mjs';
 import { isMainModule } from './lib/is-main-module.mjs';
-import { getCareerOpsRoot } from './path-resolver.mjs';
+import { getCareerOpsRoot, resolveTrackerPath } from './path-resolver.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
 const DATA_ROOT = getCareerOpsRoot();
 const DEFAULT_CSV = join(DATA_ROOT, 'data/Connections.csv');
-const TRACKER_PATH = join(DATA_ROOT, 'data/applications.md');
+const TRACKER_PATH = resolveTrackerPath(DATA_ROOT);
 const PORTALS_PATH = join(DATA_ROOT, 'portals.yml');
 const CONTACTS_PATH = join(DATA_ROOT, 'data/contacts.tsv');
 
--- a/paste-reply.mjs
+++ b/paste-reply.mjs
@@ -46,12 +46,13 @@ import fs from 'node:fs';
 import path from 'node:path';
 import readline from 'node:readline';
 import { fileURLToPath, pathToFileURL } from 'node:url';
+import { getCareerOpsRoot } from './path-resolver.mjs';
 import { renameSyncWithRetry } from './tracker-utils.mjs';
 import { isMainModule } from './lib/is-main-module.mjs';
 
-const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const DATA_ROOT = getCareerOpsRoot();
 const CANDIDATES_PATH = process.env.CAREER_OPS_REPLY_CANDIDATES
-  || path.join(__dirname, 'data', 'reply-candidates.json');
+  || path.join(DATA_ROOT, 'data', 'reply-candidates.json');
 
 
 /**
--- a/paste-reply.mjs
+++ b/paste-reply.mjs
@@ -51,7 +51,7 @@ import { renameSyncWithRetry } from './tracker-utils.mjs';
 import { isMainModule } from './lib/is-main-module.mjs';
 
 const DATA_ROOT = getCareerOpsRoot();
-const CANDIDATES_PATH = process.env.CAREER_OPS_REPLY_CANDIDATES
+export const CANDIDATES_PATH = process.env.CAREER_OPS_REPLY_CANDIDATES
   || path.join(DATA_ROOT, 'data', 'reply-candidates.json');
 
 
--- a/rejection-latency.mjs
+++ b/rejection-latency.mjs
@@ -64,7 +64,7 @@ const DATA_ROOT = getCareerOpsRoot();
 const DEFAULT_ACTIVE_INTERVIEWS_PATH = existsSync(join(DATA_ROOT, 'data/active-interviews.md'))
   ? join(DATA_ROOT, 'data/active-interviews.md')
   : join(DATA_ROOT, 'active-interviews.md');
-const DEFAULT_TRACKER_PATH = resolveTrackerPath(DATA_ROOT);
+export const DEFAULT_TRACKER_PATH = resolveTrackerPath(DATA_ROOT);
 const PROFILE_FILE = process.env.CAREER_OPS_PROFILE || join(DATA_ROOT, 'config/profile.yml');
 
 export const DEFAULT_COURTESY_DAYS = 30;
--- a/rejection-latency.mjs
+++ b/rejection-latency.mjs
@@ -57,16 +57,14 @@ import { roleFuzzyMatch } from './role-matcher.mjs';
 import { flagValue, hasFlag, validateFlags } from './lib/cli-flags.mjs';
 import { localToday } from './lib/local-today.mjs';
 import { isMainModule } from './lib/is-main-module.mjs';
-import { getCareerOpsRoot } from './path-resolver.mjs';
+import { getCareerOpsRoot, resolveTrackerPath } from './path-resolver.mjs';
 
 const CAREER_OPS = dirname(fileURLToPath(import.meta.url));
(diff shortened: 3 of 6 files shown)


Scott-Emberson (changes requested):
The substance fix is legitimate: swapping the hardcoded `data/applications.md` and `__dirname` anchoring in the five CLIs for `resolveTrackerPath(DATA_ROOT)` / `getCareerOpsRoot()` is the right data-root resolution, and the scan.mjs part of the owned test proves it. My concern is that the owned test as shipped does not run under the real harness and is only partly load-bearing.

**1. The suite fails `test-all.mjs` outright, so it gives zero coverage in CI. `tests/cli-tracker-path-resolution.test.mjs:39,66,103`, blocker.** The harness has a static guard (#1916) that scans each discovered suite's source for `process.exit(` and refuses to import any file that contains it. This test embeds `process.exit(1)` inside the child-process `-e` strings it spawns, so the guard matches and counts the whole suite a hard failure: `node test-all.mjs --only cli-tracker-path-resolution` reports 0 passed, 1 failed with "calls process.exit() ... discovered suites must use pass/fail from tests/helpers.mjs and never exit". It only appears to pass when hand-run with bare node, which is the non-authoritative path. Restructure the child snippets so the literal `process.exit(` token does not appear in the suite source (for example return a nonzero code another way, or assert on the spawned result rather than embedding the exit), and route the top-level checks through the harness pass/fail rather than throwing.

**2. Two of the four named tools are only smoke-tested, and two changed tools have no test at all. Same file, major.** The scan checks (`CAREER_OPS_TRACKER` and flat `CAREER_OPS_DATA_DIR`) are strong and mutation-sensitive: reverting the fix to a hardcoded path reddens them with a concrete MISMATCH. But the reply-watch check only asserts exit 0, and because `CAREER_OPS_DATA_DIR` is set, a hardcoded `join(DATA_ROOT,'data','applications.md')` resolves to the same file, so the mutation does not move it; the linkedin-join `--self-test` never reaches `TRACKER_PATH` at all. So neither is mutation-sensitive. And `rejection-latency.mjs` and `paste-reply.mjs`, both changed by this PR, have no test. To make the coverage real, clear `CAREER_OPS_DATA_DIR` in the reply-watch case so the resolver genuinely differs and assert the resolved tracker path, drive linkedin-join down a path that reads the tracker, and add a case each for rejection-latency and paste-reply.

Fix the harness-guard issue so the suite actually runs, and give the other four tools a real path assertion, and this holds. The substance change itself is the maintainer's call on core.

Scott-Emberson (commented):
Re-checked at this head. Both things I raised earlier are resolved, so I am clearing my earlier request.

The first was the blocker: the owned test carried a literal `process.exit(`, which `test-all.mjs` statically scans for (line 149, `/\bprocess\.exit\s*\(/`), so the suite was refused and never actually ran, green only because it was skipped. That is gone. `tests/cli-tracker-path-resolution.test.mjs` now spawns its children through `execFileSync` and reads results via `JSON.parse`, nothing trips the guard, and `node test-all.mjs --only cli-tracker-path-resolution` genuinely executes: 10 passed, 0 failed. The suite is discovered and runs in a full CI pass now.

The second was that the test did not really constrain the fix. It does now. It imports the actual resolved constants (`scan.mjs` `APPLICATIONS_PATH`, `reply-watch.mjs` `APPS_FILE`, `linkedin-join.mjs` `TRACKER_PATH`, `rejection-latency.mjs` `DEFAULT_TRACKER_PATH`) that route through `resolveTrackerPath(DATA_ROOT)`, asserts the DATA_ROOT-anchored paths against literal expecteds, and spawns two real CLIs (`linkedin-join --tracker-only`, `rejection-latency`) to check actual behavior, covering both the `CAREER_OPS_TRACKER` override and the flat `CAREER_OPS_DATA_DIR` layout. I mutation-checked it: reverting `resolveTrackerPath` to the old code-root anchor reddens 8 of 10 and the run exits 1, and the two that stay green anchor via `getCareerOpsRoot` rather than `resolveTrackerPath`, so they are correctly outside the blast radius. Green again on revert.

Nothing blocking from my side now.

Scott-Emberson (approved):
Re-verified at this head, both earlier concerns are resolved.

The owned test runs in CI now: it no longer carries a literal `process.exit(`, so `test-all.mjs`'s static scan no longer refuses it, and `node test-all.mjs --only cli-tracker-path-resolution` genuinely executes, 10 passed, 0 failed. It also constrains the fix: it imports the real resolved constants that route through `resolveTrackerPath(DATA_ROOT)` and spawns two real CLIs, covering both the `CAREER_OPS_TRACKER` override and the flat `CAREER_OPS_DATA_DIR` layout. Reverting `resolveTrackerPath` to the old code-root anchor reddens 8 of 10 and the run exits 1, and the two that stay green anchor via `getCareerOpsRoot`, correctly outside the blast radius.

Good to go.

Issue #3732, opened 2026-09-03
updater-migration-tests.mjs runs its git fixture inside the real checkout on Windows


ringo380:
## What happens

`updater-migration-tests.mjs` builds its own git fixture, guarded by a toplevel check (lines 46-55):

```js
const cwd = process.cwd();
const toplevel = execFileSync('git', ['rev-parse', '--show-toplevel'], { cwd, encoding: 'utf8' }).trim();
if (toplevel !== cwd) {
  execFileSync('git', ['init', '-q'], { cwd });
  execFileSync('git', ['config', 'user.email', 'tests@example.invalid'], { cwd });
  execFileSync('git', ['config', 'user.name', 'career-ops tests'], { cwd });
  execFileSync('git', ['add', 'update-system.mjs', 'updater-migration-tests.mjs', 'AGENTS.md'], { cwd });
  execFileSync('git', ['commit', '-qm', 'test fixture'], { cwd });
}
```

On Windows the two strings never match, because git prints the toplevel with forward slashes and Node's `process.cwd()` uses backslashes:

```
toplevel= "C:/Users/ringo/git/career-ops"
cwd     = "C:\Users\ringo\git\career-ops"
equal? false
```

So running the suite standalone from the repo root (`node updater-migration-tests.mjs`, which the file's shebang invites) takes the "throwaway copy" branch inside the real checkout. It re-runs `git init`, writes a fake identity into the real `.git/config`, stages three tracked files, and tries to commit `test fixture` onto the user's current branch. In my run the commit only failed because nothing was staged; anything already staged would have been committed under the fake author. The config write always lands, and it replaces a repo-local identity if one is set (mine was, so the checkout silently switched to the fake author):

```
$ git config --local user.name
career-ops tests
$ git config --local user.email
tests@example.invalid
```

My next real commit in the checkout was authored as `career-ops tests <tests@example.invalid>` before I noticed. The suite itself reports the mutation only as `FAIL migration fixture git setup: Command failed: git commit -qm test fixture`, which reads like a fixture problem, not like "your repo was just modified".

## Repro

Windows 11, git 2.55.0.windows.4, node v24.19.0, main at 55ef89b (v1.31.0):

```
git config --local --get-regexp '^user\.'   # empty
node updater-migration-tests.mjs
git config --local --get-regexp '^user\.'   # user.email tests@example.invalid / user.name career-ops tests
```

[...]

Issue #3334, opened 2026-08-25
updater: nested-folder checkout breaks update flow (v1.7/v1.8-era reports) — verify against current update-system


santifer:
Originally reported by waqy (detailed report, May 29) and har27il (autocomplete missing post-update) on Discord #bugs — converting so the report isn't lost.

**Report (v1.7/v1.8 era):** with career-ops in a nested folder, the updater misbehaved (files landing at the wrong level / skill autocomplete disappearing after update). Suspected at the time to relate to the AGENTS.md migration follow-ups.

**Status: needs a current-version repro.** update-system.mjs has been substantially rewritten since (SYSTEM_PATHS contract, user-layer safety checks, porcelain parsing), so the first step is a repro attempt on today's main from a nested checkout. If it's gone, close with the fixing era named; if any variant survives, scope it here.

**Acceptance if it reproduces:** update from a nested checkout leaves system files where the CLIs expect them and skill autocomplete intact, with the failing path named in a test.

maxmilian:
Repro attempt on today's `main` (`aff8be8`, VERSION 1.29.0), as the issue asks. **A variant survives — but it is not the one the title describes.**

## Three layouts, same build

| layout | `git rev-parse --show-toplevel` from the install dir | `check` | `apply` |
|---|---|---|---|
| **A. `.git`-less copy inside another git repo** (unzip into an existing project) | **the outer repo** | `update-available` / `system-files-changed` | ❌ fails, and writes to the outer repo |
| B. real clone, nested arbitrarily deep | career-ops itself | `up-to-date` | ✅ `Update complete` / `Updated 318 system paths` |
| C. `.git`-less copy not inside any git repo | `fatal: not a git repository` | `up-to-date` | — |

**B is the one that matters for the framing.** A normal clone works no matter how deep it sits, so "nested folder" is not the discriminator. The trigger is **career-ops having no `.git` of its own while some ancestor directory does** — which is what you get from downloading the ZIP into an existing project. Worth correcting in the title, or the next person will try nesting a clone and conclude it's fixed.

## Mechanism

[...]

PR #4089, opened 2026-09-11
test(harness): remove the migration suite's fixture repo on exit

sebastienfi:
## What does this PR do?

`updater-migration-tests.mjs` gives the smoke matrix's shared throwaway copy a git repository of its own and never removes it. Every smoke script after it then runs inside that repository, including `update-system.mjs check`, which passes its nested-install guard (#3334) and `git fetch`es upstream `main` into the empty fixture repo: a full-history download from GitHub, inside the test suite. On a slow enough connection that crosses the shared 30s budget and the entry is killed as `update-system.mjs check crashed (exit null, signal SIGTERM)`.

This removes the fixture repository when the suite exits.

## Related issue

None - bug fix, sent straight in. Related: #3732 is the same ten-line fixture block (the Windows toplevel comparison). It is not fixed here, but the cleanup is deliberately keyed so it cannot make #3732 worse - details below.

This is also the failure I mentioned in #4036's test-suite note, where I put it down to a cold cache and offered a `timeoutMs` bump. That diagnosis was wrong, and a timeout bump would have been the wrong fix.

## Reproduction

The same sequence the smoke matrix runs, from a copy of the tree nested inside the checkout (`.git`, `node_modules`, `data/`, `reports/` excluded, as `copyDirSync` does):

| step | before | after |
|---|---|---|
| `node updater-migration-tests.mjs` | passes, leaves `.git` in the copy | passes (368/368), leaves nothing |
| `node update-system.mjs check` from the same copy | `rev-parse` finds the fixture repo, then `git fetch --quiet https://github.com/career-ops-hq/career-ops.git main` - 19s on one run here, ~48s on another, and past 30s inside the full suite | `{"status":"not-a-git-toplevel",...}` in 0s, no network |

CI stays green only because GitHub-hosted runners fetch fast, so every CI run of the smoke section currently downloads the repository once more.

## Why not raise the entry's `timeoutMs`

The time is not the script's own work: `check` from a nested copy should return in milliseconds without touching the network, and does once the copy is left alone. A bigger budget would make the matrix wait for a network clone instead of noticing that one is happening, and would still fail offline.

## The fix

- The `exit` handler is registered only when `.git` did **not** exist before the suite ran, and before `git init`, so a setup that throws halfway still cleans up. It runs on normal completion, on the suite's failing exit, and after an uncaught error.
- It is keyed on that pre-existence, never on the `toplevel !== cwd` comparison that guards `git init`. Per #3732 that comparison never matches on Windows, so the init branch runs inside a real checkout there; a cleanup keyed on it could delete the user's `.git`. A `.git` the suite did not create, a linked worktree's `.git` file included, is never touched.
- A failed removal prints `FAIL migration fixture cleanup: ...` and sets a non-zero exit code rather than passing quietly.

**Known limitation:** exit handlers do not run when the process is killed, so a suite that is itself SIGTERM'd for time still leaves the repository behind. That path already reports its own failure.

## Tests

New `tests/updater-migration-fixture-cleanup.test.mjs` (auto-discovered), 4 assertions, hermetic (`hermeticGitEnv`, a throwaway outer repo, and every proxy variable pointed at a closed local port so a regression fails as `offline` in milliseconds instead of downloading):

1. the suite really ran its `apply()` probes against a fixture repository (so the next check is not vacuous), and
2. no `.git` is left in the copy it ran from;
3. `check` run next from the same directory reports `not-a-git-toplevel`;
4. a copy that already was its own checkout keeps its `.git` and its `HEAD`.

Run against `main`'s unfixed `updater-migration-tests.mjs`, exactly 2 of the 4 fail - the leaked `.git`, and `check` returning `{"status":"offline"}` because it got past the guard and reached for the network. The non-vacuity and safety assertions pass either way, which is the point of having them.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] If this is a new feature or architecture change, I opened an issue first (bug fixes, providers, docs & translations are exempt - send those straight in)

`node test-all.mjs` here: 8693 passed, 0 failed, 15 warnings - the same 15 as a run without this change (only a random probe filename differs) - and the smoke matrix now reports `update-system.mjs check runs OK`, the entry that was being killed.

https://claude.ai/code/session_01Bzg79mEBiS4AQQeagJ1FUP

## Summary

The migration test harness now removes the temporary `.git` directory when it created one: `updater-migration-tests.mjs`. Cleanup runs after normal completion, test failure, and uncaught errors. Pre-existing repositories, including linked worktrees, remain unchanged.

A new test verifies fixture cleanup, subsequent `update-system.mjs check` behavior, and preservation of an existing repository: `tests/updater-migration-fixture-cleanup.test.mjs`.

System files touched: `updater-migration-tests.mjs` and `tests/updater-migration-fixture-cleanup.test.mjs`. No changes affect `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, or `.github/`.

The full suite passes: 8,693 tests passed with no failures.

### DIFF
--- a/tests/updater-migration-fixture-cleanup.test.mjs
+++ b/tests/updater-migration-fixture-cleanup.test.mjs
@@ -1,0 +1,129 @@
+/**
+ * updater-migration-fixture-cleanup.test.mjs - the migration suite's git
+ * fixture must not outlive the suite.
+ *
+ * updater-migration-tests.mjs gives a nested .git-less copy a tiny repository
+ * of its own, so update-system's nested-install guard (#3334) lets its apply()
+ * probes through. test-all runs every smoke script from ONE shared copy, and
+ * that repository used to stay behind: `update-system.mjs check`, later in the
+ * same matrix, then passed the guard it should have tripped and fetched
+ * upstream main into the empty fixture repo - a full clone over the network,
+ * killed at the 30s default on a slow connection as
+ * `update-system.mjs check crashed (exit null, signal SIGTERM)`.
+ *
+ * The sections run the real suite from a nested copy inside a throwaway outer
+ * repo (the shape of test-all's shared copy), then `check` from the SAME
+ * directory, which is the order the smoke matrix runs them in. Four files are
+ * enough: update-system.mjs is self-loading (#1706). The suite's manifest
+ * assertions fail against so sparse a copy, which is useful rather than noise:
+ * it sends the suite down its failing-exit path, and the cleanup has to hold
+ * there too.
+ */
+
+import { mkdtempSync, mkdirSync, copyFileSync, existsSync } from 'fs';
+import { spawnSync } from 'child_process';
+import { tmpdir } from 'os';
+import { join } from 'path';
+import { pass, fail, NODE, ROOT, rmSync, hermeticGitEnv } from './helpers.mjs';
+import { gitIn } from '../update-system.mjs';
+
+console.log('\n🧪 Testing updater-migration-tests.mjs fixture cleanup...');
+
+const SUITE_FILES = ['update-system.mjs', 'updater-migration-tests.mjs', 'AGENTS.md', 'VERSION'];
+
+// Every proxy variable curl and git read, pointed at a closed local port. With
+// the fixture repository leaked, `check` gets past its guard and reaches for
+// the network; this turns that into an immediate `offline` instead of a
+// download, so the regression fails fast and never depends on connectivity.
+const NO_NETWORK = {
+  ...Object.fromEntries(
+    ['HTTPS_PROXY', 'https_proxy', 'HTTP_PROXY', 'http_proxy', 'ALL_PROXY', 'all_proxy']
+      .map((name) => [name, 'http://127.0.0.1:9']),
+  ),
+  NO_PROXY: '',
+  no_proxy: '',
+};
+
+function makeCopy(dir) {
+  mkdirSync(dir, { recursive: true });
+  for (const file of SUITE_FILES) copyFileSync(join(ROOT, file), join(dir, file));
+}
+
+function spawnIn(cwd, fixtureDir, args) {
+  return spawnSync(NODE, args, {
+    cwd,
+    encoding: 'utf-8',
+    timeout: 60000,
+    env: { ...hermeticGitEnv(join(fixtureDir, 'hermetic-gitconfig')), ...NO_NETWORK },
+  });
+}
+
+// ── 1. The suite's fixture repository is gone once the suite exits ──
+{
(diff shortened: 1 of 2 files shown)


Scott-Emberson (commented):
Reviewing the owned test `tests/updater-migration-fixture-cleanup.test.mjs`. The substance change is in `updater-migration-tests.mjs`, a maintainer-owned harness for the updater, so routing is the maintainer's; the added test is ours and I verified it.

It is sound and it targets a real problem. The migration suite gave its throwaway nested copy its own .git so the updater's nested-install guard would let apply() probes through, but never removed it, and since test-all runs every smoke script from the same shared copy, the leaked .git turned that copy into a real checkout, so a later `update-system.mjs check` passed the guard it should have tripped and fetched upstream main into the fixture, which is the SIGTERM-at-30s hang. The fix registers an exit handler that removes the fixture .git, keyed on .git being absent before the suite ran rather than on the toplevel-versus-cwd comparison (which is unreliable on Windows and could delete a real checkout). The test spawns the real migration suite from a nested copy, runs the real check from the same dir with proxies pointed at a dead port so a leak fails fast as offline rather than a real fetch, and asserts the real state (the .git is gone, check returns not-a-git-toplevel, a pre-existing repo keeps its .git), with a non-vacuity guard that the fixture existed while the suite ran. Root is pinned via mkdtemp and cleaned in finally. Reverting the cleanup reddens two of the four assertions (the leftover .git, and check reaching the network).

Owned test sound, not behind. SIGNAL (the substance is the maintainer's updater harness), so not surfacing as ready-to-approve; routing and merge are the maintainer's.

santifer:
@sebastienfi: the diagnosis is the part worth naming here.

You went back to a call you had already made in #4036, said it was wrong, and replaced it with the real cause. A timeout bump would have hidden this for months. The leaked fixture repository turning the shared copy into a checkout of its own, so the updater's check passed the nested-install guard and fetched upstream inside the smoke matrix, is the actual thing.

We ran it here on top of current main rather than reading it: your test passes four of four, and reverting only the harness change turns two of them red and takes the suite to a failing exit. The migration suite itself stays at 369 passed.

The safety key is right too: cleanup fires on the fixture repository being absent beforehand, never on the toplevel comparison, so the Windows problem in #3732 cannot turn this into deleting a repository someone already had.

Favorable review, and sorry it took a week to reach you. The branch is updated, CI is running, and we merge when it comes back. Nothing needed from you.

santifer (approved):
Removes the migration suite's leaked fixture repository on exit, keyed on it being absent beforehand so a real checkout is never touched. Verified locally on current main: the added test is red without the fix and green with it, and the migration suite stays clean.

PR #4090, opened 2026-09-11
test(updater): seal git fixture environment

### LINKED ISSUE #3801: `makeUpdaterRepo` builds its fixture with the un-hermetic runner, so both updater staging suites are still defeatable by ambient `GIT_CONFIG_*`
Reporting this against my own PR: the fixture is one I added in #2531, and it does not use the hermetic environment this repo already built. Verified at `6c8c0f5f`.

**Related but distinct from #3732**, and worth reading together. That issue's second half proposes to "pin `core.excludesFile` in the fixture the way `tests/helpers.mjs` does". This issue is the reason that pin is not sufficient on its own: the pin `tests/helpers.mjs` uses is the *file* layer, and #2726 already established that the runtime layer overrides it. Fixing #3732 with a file-layer pin would reproduce the gap described below in a third fixture.

### The lineage this sits in

- **#2269 → #2567 → #2726** (all @ringo380, who reported it and then fixed it twice) — the index-mode fixture had no effective local pin; one was added; then a fixture-local pin was shown to be *still* insufficient, because runtime `GIT_CONFIG_COUNT`/`KEY_n`/`VALUE_n` apply after every config file. That produced `hermeticGitEnv`.
- **#3011** (@artemtrofymenko) — showed `GIT_CONFIG_PARAMETERS` and `GIT_CONFIG` walk past all three of that helper's pins, and closed them.

This is not that fixture reopening. It is the same bug class surviving in a different fixture — `makeUpdaterRepo` (`tests/helpers.mjs:576`), which configures itself exactly the way #2726 proved insufficient:

```js
// tests/helpers.mjs:587
g('config', 'core.excludesFile', emptyExcludes);
```

That pin sits 24 lines above `hermeticGitEnv` (line 611) in the same file, and the fixture never reaches for it.

### No fixture in the family is hermetic

I nearly wrote that `updater-nested-checkout-guard` is the protected one. It is protected at a different layer:

| suite | fixture setup | spawned updater |
|---|---|---|
| `tests/updater-nested-checkout-guard.test.mjs` | bare `gitIn` | ✅ `env: hermeticGitEnv(...)` (line 59) |
| `tests/updater-add-paths.test.mjs` | bare `gitIn` via `makeUpdaterRepo` (line 35) | n/a |
| `tests/updater-is-tracked.test.mjs` | bare `gitIn` via `makeUpdaterRepo` (line 40) | n/a |

The one hermetic use in the family guards the updater *subprocess*, while every fixture in it is still built with the un-hermetic runner — and the fixture is where this bug lands.

### Reproduction

At `6c8c0f5f`. One ambient-config injection, expressed through three environment variables:

```console
$ node tests/updater-add-paths.test.mjs                                   EXIT=0
$ node tests/updater-is-tracked.test.mjs                                  EXIT=0

$ GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=core.excludesFile \
  GIT_CONFIG_VALUE_0=<file containing "*.txt"> node tests/updater-add-paths.test.mjs
EXIT=1      # 5 assertions pass first, then:
      'nothing to commit (create/copy files and use "git add" to track)\n',

$ GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=core.excludesFile \
  GIT_CONFIG_VALUE_0=<file containing "*.txt"> node tests/updater-is-tracked.test.mjs
EXIT=1      # 0 assertions — the fixture commit precedes every one of them:

🧪 Testing isTracked (ignored-but-tracked vs never-tracked)...
node:internal/errors:983
  const err = new Error(message);
              ^
Error: Command failed: git commit -qm base
    at execFileSync (node:child_process:952:15)
    at gitRawIn (file:///.../update-system.mjs:766:12)
```

The ambient rule excludes the fixture's own seed file, `git add -A` stages nothing, and the base commit dies. Both fail as an **uncaught `execFileSync` throw carrying a git transcript**, not as a red assertion — so the output points at the updater, not at the environment. The two suites differ only in where the poisoned commit sits relative to their assertions.

(Re-measured after #3688 added 116 lines to `updater-add-paths.test.mjs`; the counts are unchanged. #3688 did not touch `tests/helpers.mjs` — the 60 lines that file gained since are #3514's unrelated `directoryDenyBinds` work, which sits after `hermeticGitEnv` and does not bear on this.)

### The fix does not need to touch production code

Correcting something I assumed before checking. `gitRawIn` (`update-system.mjs:763`) does call `execFileSync` with no `env` option — but that does not matter here, because **the runner is already injected**. `makeUpdaterRepo`'s own docstring says so:

> `gitIn` is injected rather than imported so this module keeps depending on nothing but Node builtins

and both suites pass it themselves — `makeUpdaterRepo(gitIn, { prefix: 'co-addpaths-', includeRoot: true })`. The same seam exists one level down: `addPaths(names, { git: counting })` (`updater-add-paths:510`) and `isTracked('.update-dismissed', { git: ... })` are already used for spies and fault injection in these very files.

So a hermetic fixture is reachable by handing `makeUpdaterRepo` an env-aware runner, with the production wrapper untouched.

### Options

1. **Pass an env-aware runner through the existing injection seam** — the fixture builds its own `hermeticGitEnv(...)` and wraps `execFileSync` with it, for the lifetime of the returned `g`. Closes every channel #2726 and #3011 documented, changes no production code, and keeps the suite family consistent.
2. Give `gitRawIn` an optional `env` parameter. Possible, not necessary — worth it only if production paths want the same control for their own reasons.
3. Docstring saying which layer `makeUpdaterRepo` does not close. That is acceptance, not a fix, and precondition-without-enforcement is the shape #2531 spent seven rounds replacing.

Not proposing a bare `GIT_CONFIG_COUNT=0` around the helper: `makeUpdaterRepo` returns `g`, which runs git *after* the helper returns, so restoring immediately would leave those calls exposed and not restoring would mutate process-global state with no cleanup lifecycle. It would also close only the `KEY_n`/`VALUE_n` channel, leaving the two #3011 found.

Happy to take (1). Following #3011's principle, the test would assert what git *saw* through the fixture rather than that the env object lacks a key — the latter passes on an implementation that deletes the variables after git already has them.

OBOneProtocol:
## What does this PR do?

Seals updater-related Git fixtures so ambient `GIT_CONFIG_*` settings from a contributor's shell cannot change what the fixture stages, commits, or writes. The shared updater fixture now builds its Git runner through `hermeticGitEnv()`, and the affected updater fixture-family suites use that sealed runner consistently.

## Related issue

Fixes #3801

Related: #3732. I did not change `updater-migration-tests.mjs` in this PR because `.github/copilot-instructions.md` lists it as a protected control file for agents.

## Type of change

- [x] Bug fix
- [ ] New feature

## Validation

- `npm run lint` -- 693 `.mjs` files passed syntax check.
- `node tests/hermetic-git-env.test.mjs && node tests/updater-add-paths.test.mjs && node tests/updater-is-tracked.test.mjs && node tests/updater-drift-detection.test.mjs && node tests/doctor-tracked-bak-files.test.mjs && node tests/updater-nested-checkout-guard.test.mjs` -- affected fixture suites passed.

## Checklist

- [x] If this is a new feature or architecture change, I opened an issue first (bug fixes, providers, docs & translations are exempt -- send those straight in)

## AI assistance

CLOUD, using Copilot SDK in VS Code, prepared this patch at the user's request. The user authorized commit, push, and PR submission after repository-guideline review and local validation.

## Human review

- [ ] Diff read
- [ ] Behavior validated
- [ ] Tests reviewed
- [ ] Public-code matches checked

### DIFF
--- a/tests/doctor-tracked-bak-files.test.mjs
+++ b/tests/doctor-tracked-bak-files.test.mjs
@@ -6,7 +6,7 @@
 // untracks a path git already has, so the update looks like it silently did
 // nothing (career-ops#2881) with no signal pointing at .bak. This pins the
 // doctor check that surfaces it instead of letting it stay silent.
-import { pass, fail, NODE, ROOT } from './helpers.mjs';
+import { pass, fail, NODE, ROOT, makeHermeticGitRunner } from './helpers.mjs';
 import { execFileSync } from 'child_process';
 import { mkdtempSync, writeFileSync, rmSync } from 'fs';
 import { tmpdir } from 'os';
@@ -29,24 +29,22 @@ function runDoctor(cwd) {
   }
 }
 
-function git(cwd, ...args) {
-  execFileSync('git', args, { cwd, stdio: 'ignore' });
-}
-
 function initRepo(dir) {
-  git(dir, 'init', '-q');
-  git(dir, 'config', 'user.email', 'test@example.com');
-  git(dir, 'config', 'user.name', 'Test');
+  const git = makeHermeticGitRunner(dir);
+  git('init', '-q');
+  git('config', 'user.email', 'test@example.com');
+  git('config', 'user.name', 'Test');
+  return git;
 }
 
 // 1. A tracked .bak file is reported as a warning naming the remedy.
 {
   const dir = mkdtempSync(join(tmpdir(), 'co-bak-2-'));
   try {
-    initRepo(dir);
+    const git = initRepo(dir);
     writeFileSync(join(dir, 'notes.md.bak'), 'stale backup\n', 'utf-8');
-    git(dir, 'add', 'notes.md.bak');
-    git(dir, 'commit', '-q', '-m', 'accidentally tracked backup');
+    git('add', 'notes.md.bak');
+    git('commit', '-q', '-m', 'accidentally tracked backup');
 
     const state = runDoctor(dir);
     if (state._error) {
@@ -73,10 +71,10 @@ function initRepo(dir) {
 {
   const dir = mkdtempSync(join(tmpdir(), 'co-bak-3-'));
   try {
-    initRepo(dir);
+    const git = initRepo(dir);
     writeFileSync(join(dir, 'cv.md'), '# CV\n', 'utf-8');
-    git(dir, 'add', 'cv.md');
-    git(dir, 'commit', '-q', '-m', 'add cv');
+    git('add', 'cv.md');
+    git('commit', '-q', '-m', 'add cv');
 
     const state = runDoctor(dir);
     if (state._error) {
--- a/tests/helpers.mjs
+++ b/tests/helpers.mjs
@@ -636,11 +636,6 @@ export async function captureConsoleErrors(fn) {
  * pins nothing. They are different fixtures that share a name, not copies of
  * this one.
  *
- * `gitIn` is injected rather than imported so this module keeps depending on
- * nothing but Node builtins — 57 of the 62 suites import it, and none of them
- * should pull in update-system.mjs as a side effect of asking for `pass`/`fail`.
- *
- * @param {(dir: string, ...args: string[]) => any} gitIn - Updater's git runner.
  * @param {object} [options]
  * @param {string} [options.prefix='co-updater-'] - mkdtemp prefix, so a leftover
  *   temp dir names the suite that made it.
@@ -650,9 +645,9 @@ export async function captureConsoleErrors(fn) {
  *   fixture. `isTracked` never reads it.
  * @returns {{dir: string, g: Function, ctx: {git: Function, root?: string}}}
  */
-export function makeUpdaterRepo(gitIn, { prefix = 'co-updater-', includeRoot = false } = {}) {
+export function makeUpdaterRepo({ prefix = 'co-updater-', includeRoot = false } = {}) {
   const dir = mkdtempSync(join(tmpdir(), prefix));
-  const g = (...args) => gitIn(dir, ...args);
+  const g = makeHermeticGitRunner(dir);
   g('init', '-q', '-b', 'main', '.');
   g('config', 'user.email', 'test@example.com');
   g('config', 'user.name', 'Test');
@@ -709,6 +704,29 @@ export function hermeticGitEnv(gitConfigPath, base = process.env) {
   delete env.GIT_CONFIG_PARAMETERS;
   delete env.GIT_CONFIG;
   return env;
+}
+
+/**
+ * Build a git runner whose every invocation uses hermeticGitEnv().
+ *
+ * The config path lives under .git so fixture `git add -A` calls cannot stage
+ * the pin itself. It may not exist for the initial `git init` or `git clone`;
+ * git treats a missing global/system config as empty, then the repository
+ * creation makes the parent directory available for later commands.
+ *
+ * @param {string} root - Fixture repository root.
+ * @param {object} [base=process.env] - Parent environment to seal.
+ * @returns {(...args: string[]) => string} Trimmed git stdout.
+ */
+export function makeHermeticGitRunner(root, base = process.env) {
+  const env = hermeticGitEnv(join(root, '.git', 'co-hermetic-gitconfig'), base);
+  return (...args) => execFileSync('git', args, {
+    cwd: root,
+    encoding: 'utf-8',
+    timeout: 30000,
+    env,
+    stdio: ['pipe', 'pipe', 'pipe'],
+  }).trim();
 }
 
 /**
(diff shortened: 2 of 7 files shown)


Scott-Emberson (commented):
Verified the owned tests by execution and mutation. Sound, and this is exactly the right fix for the recurring CRLF and ambient-gitconfig flakiness in the updater fixtures.

The seal in `hermeticGitEnv` neutralizes every ambient git config channel rather than pinning one setting: it points GIT_CONFIG_GLOBAL and GIT_CONFIG_SYSTEM at a path under the fixture's own .git (so the developer's ~/.gitconfig is ignored), sets GIT_CONFIG_COUNT to 0, and deletes GIT_CONFIG_PARAMETERS and GIT_CONFIG, so git falls back to its defaults (autocrlf off, no hooksPath, no gpgsign). The new `makeHermeticGitRunner` wraps every invocation with that env, and the pin file sits under .git so a fixture `git add -A` cannot stage it. The new `hermetic-git-env.test.mjs` proves the seal for real: it poisons the process env through every channel, builds a fixture, and asserts both that the read did not leak (user.name stays Test) and that the ambient file is byte-unchanged (the write was not redirected out of the fixture); reverting the seal to the ambient env reddens it with the real global gitconfig leaking in.

The `helpers.mjs` change is additive apart from one intentional signature change to `makeUpdaterRepo` (it now builds its runner internally rather than taking an injected gitIn), and all three in-repo callers were migrated in the same PR with no straggler on the old form. The five updater tests are mechanical swaps to the sealed runner with their assertions unchanged, and all pass (add-paths 42, drift-detection 8, is-tracked 3, nested-checkout-guard 9, doctor-tracked-bak-files 4).

GATE-tier (tests/), owned tests sound, not behind. Good to go.

Scott-Emberson (commented):
Re-verified after the head moved. The change is sound and correctly scoped, and the seal is genuinely tested.

Every fixture git runner now switches from the injected production `gitIn` (which reads ambient `process.env`, no sealing) to a test-local `makeHermeticGitRunner` built on the existing `hermeticGitEnv`, and `makeUpdaterRepo` drops its `gitIn` parameter to build its own hermetic runner. That is the right scoping: a real install must respect the user's git config, so only the fixtures get sealed, and the old tests were genuinely non-hermetic. The `helpers.mjs` change is additive (a new exported `makeHermeticGitRunner`, plus `makeUpdaterRepo` sealing its own env), and it removes ambient state rather than weakening anything: the seal deletes `GIT_CONFIG_PARAMETERS`/`GIT_CONFIG`, pins `GIT_CONFIG_GLOBAL`/`SYSTEM` to a fixture-local `.git/co-hermetic-gitconfig`, and sets `GIT_CONFIG_COUNT=0`. The config path sitting under `.git` means a fixture `git add -A` cannot stage the pin, and a missing global/system path at init/clone reads as empty, so both cases are handled.

The affected suites are green at this head: hermetic-git-env 3, updater-add-paths 42, updater-drift-detection 8, updater-is-tracked 3, updater-nested-checkout-guard 9, doctor-tracked-bak-files 4, all with 0 failed. The full run is 8684 passed with 2 failures that are pre-existing and unrelated (the `data/offers/.gitkeep` and `jds/.gitkeep` "tracked user-layer" checks, which reproduce on main and touch none of this PR's files), so the shared-helper change broke no siblings.

I mutation-checked the seal: unsealing `GIT_CONFIG_PARAMETERS` in `hermeticGitEnv` reddens both the existing unit assertion and the new `makeUpdaterRepo` regression assertion (1 passed, 2 failed), green again on revert. The new regression test poisons all six channels (`GIT_CONFIG`, `GIT_CONFIG_GLOBAL`/`SYSTEM`, the `GIT_CONFIG_COUNT`/`KEY_0`/`VALUE_0` triple for `core.excludesFile`, and `GIT_CONFIG_PARAMETERS`) and asserts the fixture `user.name` stays `Test`, so the seal is tested rather than green-on-drop. No version-tag pollution, no network, temp fixtures cleaned.

Verified sound at this head.

Scott-Emberson (approved):
Verified at this head. Every fixture git runner now seals its own environment through `hermeticGitEnv` instead of inheriting the ambient `process.env`, which is the right scoping: a real install still respects the user's git config, only the test fixtures are sealed. The `helpers.mjs` change is additive (a new `makeHermeticGitRunner`, and `makeUpdaterRepo` sealing its own env) and removes ambient state rather than weakening anything.

The seal is genuinely tested: unsealing `GIT_CONFIG_PARAMETERS` reddens both the existing unit assertion and the new `makeUpdaterRepo` regression assertion, which poisons all six config channels and asserts the fixture `user.name` stays `Test`. The affected suites are green at this head (hermetic-git-env, updater-add-paths, updater-drift-detection, updater-is-tracked, updater-nested-checkout-guard, doctor-tracked-bak-files, all 0 failed), and the full run's only two failures are the pre-existing `.gitkeep` tracked-file checks that reproduce on main and touch nothing here. No version-tag pollution, no network, fixtures cleaned up.

Good to go.

OBOneProtocol:
My apologies for closing this one early, team. The last branch before I closed it was verified and ready to merge with approval. The issues it fixed are still showing 'OPEN'
@santifer

PR #4095, opened 2026-09-11
docs(tests): describe the test-naming convention the tree actually follows

### LINKED ISSUE #4091: docs: the test-naming rule describes 19% of tests/ and 54% of web/tests/
### Code of Conduct

### Existing issues

### What happened?

The documented test-naming rule describes a minority of the test tree, in both places it is stated.

`ARCHITECTURE.md:30`:

> Suites live in `tests/`, named `{module}.test.mjs` for the root script they cover

`CONTRIBUTING.md:189`, for the web app:

> web suites live under `web/tests/`, mirroring the tested module's path below `web/src/` ... named `{module}.test.mjs`

Measured against `main`:

| tree | suites | named for the module | named for behaviour |
|---|---|---|---|
| `tests/*.test.mjs` | 226 | 41 | **185 (82%)** |
| `web/tests/` | 49 | 26 | **23 (47%)** |

A suite counts as module-named when a root script (or, for web, a file under the mirrored `web/src/` path) exists with the same stem. Everything else is named for what it exercises: `cover-unresolved-placeholders`, `applied-date-local-vs-utc`, `mojibake-canary`, and 182 more.

Counted on `main` at `da8c6f91`. The row is the flat `tests/` directory only. `tests/providers/` is excluded on purpose: its 101 suites follow a different rule that `CONTRIBUTING.md` states explicitly and that they do follow, so counting them either way would be misleading.

*Correction: I first posted this table as 227 suites and 42 module-named. Both were one too high. The 185 behaviour-named count was right, and so was the conclusion.*

### Why it matters

This is not a tidiness complaint. A rule that four out of five files ignore still gets cited in review as though it were binding, and the cost lands on contributors.

It happened on #3653: a suite was flagged for splitting on the strength of this rule, and again on #4043, where a new suite was asked to rename to match a module that already has 11 suites covering it. Following the rule there would have made the new file the only one of its group to do so. A contributor cannot tell from the text that the rule describes 19% of the tree.

The naming is also load-bearing in the opposite direction. `test-all.mjs` discovers `tests/**/*.test.mjs` with no registration list, so nothing enforces the rule and nothing breaks when it is ignored. `--only <substring>` matches on the filename, which is why behaviour names are useful: `--only mojibake` finds the canary, while a module name would not.

### Expected behavior

The docs describe what the repo does. Two options, and I do not think it is my call which:

1. **Narrow it.** Say that a suite covering exactly one root script is named `{module}.test.mjs`, and a suite covering a behaviour across several is named for the behaviour. That matches both trees as they stand and keeps the useful half of the guidance.
2. **Drop the naming sentence** and keep the part that matters, which is where suites live and how they are discovered.

I lean to 1, because the `{module}` form is genuinely right when it applies and 42 files follow it. Happy to open the PR either way once someone picks.

### CLI tool

Claude Code

### OS and Node.js version

macOS 26.6.2 / Node 26.8.1

---

rubicon:
## What does this PR do?

`ARCHITECTURE.md` states that suites in `tests/` are "named `{module}.test.mjs` for the root script they cover." Measured on `main` at `da8c6f91`, 41 of the 226 suites directly under `tests/` are named for a root script. The other 185 are named for the behaviour they pin. The sentence describes the exception and reads as binding.

This narrows the rule rather than dropping it, because the module form is genuinely right when a suite covers exactly one script:

- **Module name** when a suite covers exactly one root script.
- **Behaviour name** when it spans several, or is one of a group carving up a large module.
- Neither is enforced. Discovery is by glob, so the name buys legibility and nothing else.

`CONTRIBUTING.md`'s web paragraph carried the same overstatement — 26 of 49 `web/tests/` suites mirror a `web/src` module — plus one imprecision worth fixing while I was there: it says `tests/web-test-layout.test.mjs` "enforces it on every PR," but that guard freezes the **discovery** contract (the glob reaches every suite on disk, and `web/`'s engines floor can run a glob operand). It asserts nothing about names. Its own header says so: "ONE responsibility: web/'s suites must be reachable by what `npm test` actually runs."

Counts are written as ratios in the docs rather than exact numbers, so they do not go stale the next time someone adds a suite.

## Related issue

Closes #4091

## Why it is worth a patch

A rule that four out of five files ignore still gets cited in review as binding, and the cost lands on contributors. It came up on #3653, and again on #4043 where a new suite was asked to rename to match a module that already has 11 suites covering it — following the rule there would have made the new file the only one of its group to do so. Nothing in the text lets a contributor tell that the rule describes the minority.

`tests/README.md` already describes both shapes correctly, so this mostly brings the two larger docs in line with the one that was right.

## Type of change

- [ ] Bug fix
- [ ] New feature

## Measurement

A suite counts as module-named when a root script (or, for web, a file under the mirrored `web/src/` path) exists with the same stem.

| tree | suites | module-named | behaviour-named |
|---|---|---|---|
| `tests/*.test.mjs` | 226 | 41 | **185 (82%)** |
| `web/tests/**` | 49 | 26 | **23 (47%)** |

`tests/providers/` is excluded from that first row deliberately. Its 101 suites follow a different convention that the docs state and that they do follow: 91 are named for the provider they cover, and the remaining 10 are cross-provider behaviour suites such as `ats-ssrf-hardening.test.mjs`. Folding them into either column would have been misleading, which is why the new text calls that subtree out separately.

## Checklist

- [x] I ran `node test-all.mjs` and all tests pass (8687 passed, 0 failed)

## Summary

Contributors can choose module-based names for single-script suites or behaviour-based names for multi-script suites. Neither naming style is enforced: glob discovery still runs matching `.test.mjs` files (`ARCHITECTURE.md:30-32`).

Web contributors can use either naming style. `tests/web-test-layout.test.mjs` checks discovery and the engines floor, not filenames (`CONTRIBUTING.md:187-197`).

No system files are touched: `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/`.


santifer:
Hi @rubicon, this PR has waited 11 days without a reply from a maintainer. That is on us, not on you, and we are sorry for it.

santifer (approved):

santifer:
Merged. Thank you, @rubicon: **ARCHITECTURE and CONTRIBUTING now describe the test-naming convention the tree actually follows**: a test named for the behaviour it checks is no longer a deviation, and the web layout test is described as checking that tests are found, not what they are called. Docs that match the repo save every new contributor from chasing a rule nobody follows.

Thank you for your patience while this waited for a decision.

PR #4102, opened 2026-09-12
theme: make the competency-tag palette themeable

sundarshankar:
## Problem

`style:` in `config/profile.yml` can recolor the accents, but the competency-tag palette is three hardcoded literals in every template that draws tags:

- tag text `hsl(187, 74%, 28%)`
- tag and `.project-tech` fill `hsl(187, 40%, 95%)`
- tag border `hsl(187, 40%, 88%)`

So a profile that sets `accent_color` and `secondary_color` to a monochrome palette gets monochrome headings and company lines next to teal-tinted tags, and the only way to finish the job is to edit `templates/`. `theme-style.mjs`'s own header explains why that is the wrong place:

> a setting that lives in config/profile.yml survives `update-system.mjs apply`, while the same customization made in templates/cv-template.html (a SYSTEM_PATHS file) is reverted by every release.

Which is exactly what happens — the edit has to be re-made after every release, or given up.

## Change

Adds `--tag-color`, `--tag-bg` and `--tag-border` with `:root` defaults **byte-identical** to the literals they replace, plus the matching `tag_color` / `tag_bg` / `tag_border` keys in `STYLE_VAR_MAP`. A profile with no `style:` block renders exactly as before.

```yaml
style:
  accent_color:    "#2563eb"
  secondary_color: "#111827"
  tag_color:       "#0e7490"   # competency-tag text
  tag_bg:          "#ecfeff"   # competency-tag and project-tech fill
  tag_border:      "#a5f3fc"   # competency-tag border
```

Applied to all three templates carrying those literals — `cv-template.html`, `cv-template.zh-minimal.html`, `resume-template.html` — because a token only the default template reads is a `style:` key that silently does nothing on the others.

## Tests

`tests/theme-style.test.mjs` gains a per-template guard asserting that each `:root` default equals the exact color it replaced, that the template actually reads every token (an unreferenced `var()` makes an override inert), that no token resolves to itself, and that no palette literal is left behind. The token-count assertion moves 5 → 8.

`node test-all.mjs`: **8690 passed, 0 failed**.

## Summary

Users can customize competency-tag text, background, and border colors with `tag_color`, `tag_bg`, and `tag_border`.

The change updates all three templates and preserves existing colors by default. Tests verify token mappings, template usage, and removal of hardcoded colors.

Tests: 8690 passed, 0 failed.

System files touched: none. `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/` were not changed.

### DIFF
--- a/templates/cv-template.html
+++ b/templates/cv-template.html
@@ -11,6 +11,9 @@
   :root {
     --accent-color: hsl(187 74% 32%);
     --secondary-color: hsl(270, 70%, 45%);
+    --tag-color: hsl(187, 74%, 28%);
+    --tag-bg: hsl(187, 40%, 95%);
+    --tag-border: hsl(187, 40%, 88%);
     --font-family: "Liberation Sans", 'Helvetica Neue', Arial, 'DejaVu Sans', sans-serif;
     --font-size: 11px;
     /* Matches generate-pdf.mjs's PDF_PAGE_MARGIN — the actual PDF page margin
@@ -176,11 +179,11 @@
     font-family: var(--font-family);
     font-size: 10px;
     font-weight: 500;
-    color: hsl(187, 74%, 28%);
-    background: hsl(187, 40%, 95%);
+    color: var(--tag-color);
+    background: var(--tag-bg);
     padding: 4px 10px;
     border-radius: 3px;
-    border: 1px solid hsl(187, 40%, 88%);
+    border: 1px solid var(--tag-border);
   }
 
   /* === WORK EXPERIENCE === */
@@ -262,7 +265,7 @@
     font-size: 9px;
     font-weight: 500;
     color: var(--accent-color);
-    background: hsl(187, 40%, 95%);
+    background: var(--tag-bg);
     padding: 1px 6px;
     border-radius: 2px;
     margin-left: 6px;
--- a/templates/cv-template.zh-minimal.html
+++ b/templates/cv-template.zh-minimal.html
@@ -15,6 +15,9 @@ version: 1.0.0
      `style:`, which the PDF pipeline injects as a later :root block. */
   :root {
     --secondary-color: hsl(270, 70%, 45%);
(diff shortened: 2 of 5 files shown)


Scott-Emberson (commented):
I reviewed the owned test `tests/theme-style.test.mjs`. It is sound and mutation-sensitive.

The test imports the real `theme-style.mjs` (via `pathToFileURL`) and reads the actual template files, no reconstruction. It round-trips `styleTokensFrom({tag_color, tag_bg, tag_border, ...})` and asserts the exact CSS-var outputs (`--tag-color === '#0e7490'`, and so on) plus the token count. Then, per template (`cv-template.html`, `cv-template.zh-minimal.html`, `resume-template.html`), it asserts each `:root` default is byte-identical to the literal it replaced (`--tag-color: hsl(187, 74%, 28%);` etc.), that each token is actually referenced by a `var(--tag-*)` (an unreferenced var would make the override inert), that there is no self-referential `--tag-x: var(--tag-x)`, and that no stray hardcoded HSL literal is left behind. So it constrains that the tokens actually drive the palette, not just that a function ran.

I mutation-checked it. Dropping `tag_color` from `STYLE_VAR_MAP` reddens the token round-trip. Reverting `color: var(--tag-color)` back to the hardcoded `hsl(187, 74%, 28%)` in `cv-template.html` reddens that template's assertion while the sibling templates stay green, so it pinpoints rather than fails broadly. `node test-all.mjs --only theme-style` is 36 passed, 0 failed at this head.

One naming note, not blocking: the tag-token defaults live in each template's `:root`, so a template edited later that drops a `--tag-*` reference would go inert silently unless it is one of these three. The test covers the three that exist today, which is the right scope for this change.

Owned test sound, not a false-pass. The substance here is in `theme-style.mjs` and the templates, so the routing and merge decision is the maintainer's, not mine to gate.

PR #4104, opened 2026-09-12
fix(web): preserve report identity when generating tailored PDFs

mbrow190:
## What does this PR do?

Generating tailored CVs for two roles at the same company on the same day previously reused the same output filename. A tracker application linked to a differently numbered report could also tailor from the wrong report while recording its PDF under the application number.

This change derives the filename, scratch path, manifest key, and tailoring prompt from the resolved report. The prompt receives the exact report path, including unpadded filenames. The company/date suffix stays compatible with existing PDF readers; existing PDFs are not renamed.

Only web-generated filenames change. The CLI mode keeps its existing naming convention, so the terminal examples in the PDF mode documentation remain valid.

## Related issue

Independent PDF fix split from #4103. This PR contains only the six PDF-related files and can merge on its own.

## Type of change

- [x] Bug fix

## Validation

- `node test-all.mjs`: 8,686 passed, 0 failed, 17 warnings. Warnings cover absent user data, the unavailable Go compiler, stock author references, and two opt-in live API checks. The Go dashboard build was skipped.
- Web: 502 tests passed, type checking passed, production build passed.
- Also checked the web-only dependency layout used by CI: 496 passed, with six existing core parity checks skipped because core dependencies are absent.
- Regression tests cover same-company collisions, application/report number mismatches, padding, exact report paths, and compatibility with the existing PDF readers.
- The application lookup regression writes both report entries to the PDF manifest and calls the real resolver for applications 309 and 310, confirming each resolves its own PDF through linked reports 018 and 019.

## Checklist

- [x] Read CONTRIBUTING.md; these bug fixes are exempt from the issue-first requirement.
- [x] No personal data or user-layer files are included.
- [x] Changes preserve the Data Contract and existing local workflow.

## Summary

Tailored PDFs now preserve the resolved report identity.

Users can generate CVs for multiple roles at the same company on the same day without filename collisions. Tracker applications can use a different application number from the linked report. The backend now uses the resolved report number for filenames, scratch HTML, manifest keys, and PDF paths (`web/src/lib/pdf-paths.mjs:58-95`).

The tailoring prompt receives the exact resolved report path, including unpadded filenames (`web/src/app/api/run/route.ts:92-114`, `web/src/lib/run-prompts.mjs:78-94`). Padded and unpadded selectors resolve to the same report. Existing PDFs are not renamed, and CLI naming remains unchanged.

## Tests

Regression coverage includes same-company collisions, application/report number mismatches, padding, exact report paths, manifest integration, and existing PDF-reader compatibility. The PDF fixture cleanup also isolates web tests from core-only dependencies (`web/tests/lib/apply-cv-resolver.test.mjs`).

Validation passed: 8,686 repository tests, 502 web tests, type checking, and the production build.

## System files

### DIFF
--- a/web/src/app/api/run/route.ts
+++ b/web/src/app/api/run/route.ts
@@ -111,7 +111,7 @@ export async function POST(req: Request) {
     kind === "evaluate"
       ? readInbox().find((j) => j.url === input)?.postedAt ?? readScanDates().get(input)
       : undefined;
-  const prompt = buildPrompt({ kind, input, memory: readMemory(), today, postedAt, lang });
+  const prompt = buildPrompt({ kind, input, memory: readMemory(), today, postedAt, lang, reportFile: pdfPaths?.reportFile });
 
   const isClaude = cliId === "claude";
   // Which tools each kind gets, and the whole claude argv, live in
@@ -384,7 +384,7 @@ export async function POST(req: Request) {
             root: careerOpsRoot(),
             pdfPaths: paths,
             format,
-            reportNum: input,
+            reportNum: paths.reportNum,
           });
           if (result.kind === "render-failed") {
             send({ type: "error", msg: result.error.slice(0, 200) });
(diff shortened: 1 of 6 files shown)


santifer:

PR #4109, opened 2026-09-12
fix(web): recover interrupted application fills without losing answers

mbrow190:
## What does this PR do?

Full-agent application filling could remain in `filling` forever after `/api/apply/drive` returned HTTP 200 and then closed without a terminal result. The client also dropped final JSON records without a newline and did not handle HTTP errors before reading the stream.

Require a valid terminal fill result, parse the final record, and stop malformed responses from becoming success. On interruption or failure, return to the existing answer editor with a visible error and retain the drafted answers so the user can review the real form and explicitly retry. Responses from a replaced session cannot update the current one, and the reader is released after completion.

This fixes the full-fill client path only. The server route and automatic submission boundary are unchanged; successful fills still instruct the user to review every answer and submit manually.

## Type of change

- [x] Bug fix

## Validation

- `node test-all.mjs`: 8,688 passed, 0 failed, 17 warnings covering missing user data, the unavailable Go compiler, stock author references, and opt-in integration checks. The Go dashboard build was skipped.
- Web: 517 tests passed, type checking passed, production build passed.
- Twenty stream tests cover interrupted and empty responses, terminal records without a newline, malformed JSON, invalid completion payloads, HTTP errors, fragmented data, terminal precedence, reader cleanup, and stale sessions.
- Executed the actual transpiled ApplyProvider with mocked HTTP. Verified interruption returns to the editor with answers retained, explicit retry succeeds, errors remain visible, and resetting the session prevents stale updates. No real AI, employer browser session, or application submission was used.

## Checklist

- [x] Read CONTRIBUTING.md; this bug fix is exempt from the issue-first requirement.
- [x] No personal data or user-layer files are included.
- [x] Keeps the existing apply protocol and human review requirement.

## Summary

Full-agent form filling now recovers safely from interrupted or malformed `/api/apply/drive` streams.

## User impact

- Interrupted fills return to the answer editor.
- Drafted answers remain available for review and explicit retry.
- Errors explain when the agent did not confirm completion.
- Malformed responses cannot be treated as successful fills.
- Users must still review every answer and submit manually.
- Replaced sessions no longer update the current form.

## Files changed

- `web/src/components/apply/apply-provider.tsx:7`: Uses `readFillResult` and returns failed fills to `"ready"`.
- `web/src/lib/apply/fill-stream.mjs:10-68`: Validates terminal records, handles HTTP errors, parses final records without trailing newlines, supports split UTF-8 data, and releases the stream reader.
- `web/tests/lib/apply-fill-stream.test.mjs:22-161`: Adds stream recovery and retry coverage.

## System files

No changes: `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/`.

### DIFF
--- a/web/src/components/apply/apply-provider.tsx
+++ b/web/src/components/apply/apply-provider.tsx
@@ -4,6 +4,7 @@ import { createContext, useCallback, useContext, useEffect, useMemo, useRef, use
 import type { ApplyField } from "@/lib/apply/extract";
 import type { ApplyIssue, DriveStep } from "@/lib/apply/issue";
 import { resolveLateSession } from "@/lib/apply/exit.mjs";
+import { readFillResult } from "@/lib/apply/fill-stream.mjs";
 
 export type FillStep = { fieldId: string; label: string; ok: boolean; thumb?: string };
 type Meta = { needsConfirmation?: boolean };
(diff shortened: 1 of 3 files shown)


santifer:

PR #4111, opened 2026-09-12
fix(pdf): use workspace styling for batch and imported renders

mbrow190:
## What does this PR do?

Generating a CV in a batch, or calling the PDF renderer from cover-letter generation, could read `style:` from the terminal's working directory instead of the tracker workspace. A configured font, color, size, or margin could disappear or come from an unrelated profile, even though a single CV generated with the CLI used the correct settings.

Read the batch's style tokens once from its workspace profile and pass them to every render. Imported rendering also resolves its default profile inside its output workspace. Explicit `styleTokens` still win, including an empty map to disable theming.

## Related issue

Bug fix to existing profile theming. No new configuration keys or UI.

## Validation

`node test-all.mjs`: 8,690 passed, 0 failed, 17 existing warnings. Warnings cover absent user fixtures, the unavailable Go compiler, shipped reference strings, and opt-in external checks.

The regression fixture runs the actual CLI from a folder containing conflicting styling settings, then compares the HTML sent to the browser for single and batch renders. It also checks imported rendering with default, explicit, and disabled styling. The new tests reproduced two failures before the fix and pass afterward. Chromium is stubbed in these tests so no personal data or installed browser is needed.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] Changes align with the existing PDF workflow.

### DIFF
--- a/generate-pdf.mjs
+++ b/generate-pdf.mjs
@@ -1445,11 +1445,11 @@ async function runBatchFromManifest(manifestPath, globals) {
   } catch (err) {
     if (err?.code !== 'ENOENT') throw err;
   }
-  // One profile governs the whole batch, so the declared order is read once
-  // rather than per entry. Anchored to workspaceRoot for the same reason the
-  // single render is: it is the anchor readStyleTokens() and the cv.md read
-  // already use, so one profile.yml supplies every setting.
-  const cvSectionOrder = readCvSectionOrder(resolve(workspaceRoot, 'config', 'profile.yml'));
+  // Read one workspace profile for the batch. The working directory must not
+  // choose a different theme from the single-document render.
+  const profilePath = resolve(workspaceRoot, 'config', 'profile.yml');
+  const cvSectionOrder = readCvSectionOrder(profilePath);
+  const styleTokens = readStyleTokens(profilePath);
 
   for (let i = 0; i < manifest.length; i++) {
     const spec = manifest[i];
@@ -1502,6 +1502,7 @@ async function runBatchFromManifest(manifestPath, globals) {
         inputPath: entryInput,
         maxPages: globals.maxPages,
         strictPages: globals.strictPages,
+        styleTokens,
       });
     } catch (err) {
       console.error(`❌ Skipping batch entry ${i} (${spec?.output ?? '?'}): ${err.message}`);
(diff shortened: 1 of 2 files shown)


Scott-Emberson (commented):
I reviewed the owned test `tests/generate-pdf-batch.test.mjs`. It is sound and mutation-sensitive, and it proves the thing that matters here.

The test drives the real styling path, not a reconstruction: it copies the real `generate-pdf.mjs` into a sandbox and both spawns it (CLI single and `--batch`) and imports the real `renderHtmlToPdf`, so `readStyleTokens` and `injectThemeStyle` actually run. The clever part is how it checks what reached Chromium: the Playwright stub base64-encodes the rendered HTML into the stub PDF's `/Marker` field, and the test decodes it and asserts the exact workspace tokens are present (`--accent-color: #123456;`, `--font-family: Georgia, serif;`, `--font-size: 12pt;`, `--page-margin: 0.4in;`) and that the competing cwd profile's `#abcdef` is absent. It runs from a different working directory that has its own `config/profile.yml`, so it genuinely proves the workspace source wins over cwd, plus single equals batch equals batch-second, explicit-token-wins, and `{}`-disables-theme. Since the stub's bytes vary with the input HTML, a constant-blob vacuous pass is not possible, and there is no real-browser dependency to skip in CI.

I mutation-checked it. A full revert of the fix reddens 2 (workspace theme differs single vs batch, and imported default theme), both at exit status 0, which is the proof the assertions check content and not mere execution. Reverting just the `renderInPage` fallback to the old cwd `readStyleTokens()` reddens the imported-default case. Revert returns to 12 passed, 0 failed. Not a false-pass.

One coverage note, not a defect, worth knowing. Reverting only the batch's explicit `styleTokens` pass leaves the suite green, because `renderInPage`'s new fallback resolves workspace styling anyway (`outputRoot === workspaceRoot`), so batch entries still get the workspace tokens through the fallback. Batch correctness is genuinely protected here, by the fallback change and the single-vs-batch equality assertion, just not by the explicit batch line itself. If you want that line pinned independently, a case where the output root and the workspace root differ would isolate it. Belt-and-suspenders as written, so this is a nit, not a blocker.

Owned test sound, not a false-pass. The substance is in `generate-pdf.mjs`, so the routing and merge decision is the maintainer's, not mine to gate.

PR #4112, opened 2026-09-12
fix(templates): honor profile styles in Chinese Minimal CVs

mbrow190:
## What does this PR do?

Chinese Minimal silently ignored the existing profile font, size, and color settings because its CSS used fixed values and overrode the shared secondary-color setting later in the template.

Connect its typography and palette to `style.font_family`, `style.font_size`, `style.accent_color`, and `style.secondary_color`. Keep the original values as fallbacks, preserve the locale font stacks, and leave the default appearance unchanged.

## Related issue

Bug fix to existing profile theming. No new settings or template-management UI.

## Validation

`node test-all.mjs`: 8,687 passed, 0 failed, 17 existing warnings. The Chromium regression test ran without a skip. Independent review also passed.

A Chromium regression test renders a synthetic CV and checks default, custom, size-only, and disabled settings across English, Chinese, Traditional Chinese, Japanese, and Arabic locale rules. It fails on upstream when the selected font is ignored.

Separate rendering checks compared all computed styles and element boundaries with upstream across six locales. The default PDF's rendered image was byte-identical, and extracted text matched. A custom font, body size, and palette stayed on one page with English and Chinese text preserved. Both default and custom output were visually inspected for clipping and overlap.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] Changes align with the existing PDF workflow.

## Summary

Chinese Minimal CV templates now honor profile font, font size, accent color, and secondary color settings.

User impact:
: Custom profile styling now changes Chinese Minimal CV output across supported locales.
: Existing locale-specific font fallbacks remain active.
: Default rendering remains unchanged through CSS fallbacks.
: No user-facing behavior breaks are identified.

Files changed:
: `templates/cv-template.zh-minimal.html:47-73, 520-524`: Adds themeable typography and color variables with fallbacks.
: `tests/theme-style.test.mjs:118`: Moves Chinese Minimal coverage to its dedicated test suite.
: `tests/zh-minimal-template.test.mjs:24`: Accepts the accent-color fallback syntax.
: `tests/zh-minimal-theme.test.mjs:23-101`: Adds Chromium regression coverage for locales and custom settings.

System files:
: No changes were made to `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, or `.github/`.

Validation:
: The supplied results confirm the expected files and implementations.
: Test execution results and review severity counts were not supplied.

### DIFF
--- a/templates/cv-template.zh-minimal.html
+++ b/templates/cv-template.zh-minimal.html
@@ -9,13 +9,10 @@ version: 1.0.0
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
 <title>{{NAME}} — CV</title>
 <style>
-  /* Themeable tokens (#1837): only --secondary-color is wired in this template
-     so far (this template does not yet expose --accent-color; that gap is
-     pre-existing and out of scope here). Overridable via config/profile.yml
-     `style:`, which the PDF pipeline injects as a later :root block. */
-  :root {
-    --secondary-color: hsl(270, 70%, 45%);
-  }
+  /* Profile style overrides are injected later by the PDF pipeline (#1837).
+     Keep each surface's existing font and color as its fallback: this template
+     has different defaults from the standard CV, including within its own
+     typography and palette. With no profile override those defaults stay put. */
 
   /* ATS-safe typography (#font-extraction fix):
      The bundled variable woff2 fonts (Space Grotesk / DM Sans) render with
@@ -47,8 +44,8 @@ version: 1.0.0
   }
 
   body {
-    font-family: 'Liberation Sans', 'Helvetica Neue', Arial, 'DejaVu Sans', sans-serif;
-    font-size: 11px;
+    font-family: var(--font-family, 'Liberation Sans'), 'Helvetica Neue', Arial, 'DejaVu Sans', sans-serif;
+    font-size: var(--font-size, 11px);
     line-height: 1.5;
     color: #1a1a2e;
     background: #ffffff;
@@ -72,7 +69,7 @@ version: 1.0.0
   }
 
   .header h1 {
-    font-family: 'Liberation Sans', 'Helvetica Neue', Arial, 'DejaVu Sans', sans-serif;
+    font-family: var(--font-family, 'Liberation Sans'), 'Helvetica Neue', Arial, 'DejaVu Sans', sans-serif;
     font-size: 28px;
     font-weight: 700;
     color: #1a1a2e;
@@ -83,7 +80,7 @@ version: 1.0.0
 
   .header-gradient {
     height: 2px;
-    background: linear-gradient(to right, hsl(187, 74%, 32%), var(--secondary-color));
+    background: linear-gradient(to right, hsl(187, 74%, 32%), var(--secondary-color, hsl(270, 70%, 45%)));
     border-radius: 1px;
     margin-bottom: 10px;
   }
@@ -92,8 +89,8 @@ version: 1.0.0
     display: flex;
     flex-wrap: wrap;
     gap: 8px 14px;
-    font-family: 'Liberation Sans', 'Helvetica Neue', Arial, 'DejaVu Sans', sans-serif;
-    font-size: 10.5px;
+    font-family: var(--font-family, 'Liberation Sans'), 'Helvetica Neue', Arial, 'DejaVu Sans', sans-serif;
+    font-size: var(--font-size, 10.5px);
     line-height: 1.4;
     color: #555;
   }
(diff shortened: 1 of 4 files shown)


Scott-Emberson (commented):
I reviewed the three owned tests here (`tests/zh-minimal-theme.test.mjs`, `tests/zh-minimal-template.test.mjs`, `tests/theme-style.test.mjs`). All three are sound and mutation-sensitive, and they drive real code.

`zh-minimal-theme.test.mjs` is the strongest. It builds the real `templates/cv-template.zh-minimal.html` through `build-cv-html.mjs` and renders it in headless Chromium, then asserts the literal defaults (body 11px, job li 10.5px, heading 27px, section-title rgb(23,32,51), job-role rgb(23,74,126), award-org rgb(115,34,195)) and that a `Georgia / 10pt / #2563eb / #111827` profile override flows through across six locales with literal RGB and px expecteds, while a partial and an empty override leave unrelated surfaces untouched. I mutation-checked it: pinning `--zhm-ink` to a literal reddens the company-color assertion (rgb(23,32,51) vs the expected rgb(17,24,39)), pinning `--zhm-accent` reddens the job-role assertion, and both go green on revert.

`zh-minimal-template.test.mjs` reads the real template and asserts `--zhm-accent: var(--accent-color, #174a7e)` keeps its default literal as the fallback, the chip is de-styled, and no stray `hsl(270)` is left in the zh section. Pinning `--zhm-accent` to a literal reddens it with an exact regex mismatch on the input, green on revert, so it pinpoints.

`theme-style.test.mjs` is sound but orthogonal to this PR's substance: the only edit here removes zh-minimal from the standard-template `:root` guard loop, and that removal is correct. zh-minimal uses per-surface `var(--secondary-color, hsl(...))` fallbacks rather than a single `:root --secondary-color` default, so leaving it in the loop false-fails (hasRoot false, usesVar false, leftoverHardcoded true). The suite still guards the shared `theme-style.mjs` machinery: dropping `secondary_color` from `STYLE_VAR_MAP` reddens 2, green on revert. None of the three is a false-pass.

One thing worth passing to whoever routes this. This PR does not touch `theme-style.mjs` at all (it is byte-identical to main), so it is not a follow-on to the tag-palette token work in the way the summary I was working from implied. The tokens actually in play here are `--secondary-color` / `--accent-color` / `--font-family` / `--font-size` (via `--zhm-ink` / `--zhm-accent`), not the tag-palette vars, and the zh chip is de-styled rather than themed. The tests correctly target the font and secondary and accent tokens, so the tests match the code. The substance of this PR (the zh template edit) is covered entirely by the two dedicated zh tests, and both hold.

Owned tests sound, not a false-pass. The substance is in the template, so the routing and merge decision is the maintainer's, not mine to gate.

santifer:
Thanks for this, @mbrow190, and sorry it sat six days.

The part that convinced me is the one that is easiest to get wrong. I rendered both templates through `build-cv-html.mjs` with no profile and compared computed font, size, colour, background, weight and border across twenty-one surfaces in headless Chromium: identical to main. The change stays inert until someone opts in, which is the contract `theme-style.mjs` documents.

I also checked that your Chromium test runs rather than skipping quietly, because a skip guard is exactly how this kind of coverage disappears: `skipped 0`, and reverting the template to main turns it red on the font assertion.

Narrowing the three-template guard in `theme-style.test.mjs` to two is a fair trade, because what replaces it for this template is stronger than what it asserted. Saying so in the diff comment is what made that easy to check.

The review is favourable. What is left is mine: approving the current head and refreshing the branch onto current main, then it goes in once CI closes green on the refreshed head.

santifer (approved):
Favourable review: default rendering measured identical to main across twenty-one surfaces, and the new Chromium test verified to run rather than skip.

santifer:
Thanks @mbrow190 — the token wiring is consistent: every hardcoded value becomes `var(--token, <the value that was already there>)`, so a profile with no `style:` block renders byte-identical.

On CodeRabbit's note about `.contact-row` and the locale font stacks: it describes something real, but it isn't yours. That declaration already overrode the `zh`, `zh-CN`, `ja` and `ar` body stacks before this PR, with the same Latin fallback chain. The zh-minimal block resets `.header h1`, `.section-title` and `.competency-tag` to `font-family: inherit` and simply never did the same for `.contact-row` — which is also why `zh-TW` needed its own explicit rule. Changing it here would move rendering under the visual-regression gate for reasons unrelated to your change, so I'd rather track it separately than grow your PR.

The review of the approach is favourable. I've just refreshed the branch against `main`; once that CI run comes back green it can go in.

santifer:
@mbrow190 merging this now, and thanks for the patience: it sat approved while another PR went in ahead of it, and then needed a refresh on top of that.

The change does what it says. The zh-minimal template had font-family and font-size hard-coded, so it ignored the `:root` block theme-style injects from the profile. Wrapping each value in `var(--token, <the value that was already there>)` means a profile with no overrides renders exactly as before, and one with overrides is finally honoured. Your three tests cover both directions, and the suite ran green on this content.

Nothing further needed from you.

PR #4113, opened 2026-09-12
fix(web): retain tracker guard until canceled worker closes

mbrow190:
## What does this PR do?

Canceling an evaluation or PDF job released the tracker deletion guard as soon as SIGTERM was sent. The worker could still be writing the tracker when a row deletion became available. A failed stream enqueue could also leave the guard stuck after the worker had closed.

Release the guard once both the worker's `close` event and the run's completion have occurred. Cancellation during PDF rendering continues to wait for rendering and tracker marking to finish. Cleanup also runs when the stream has already closed.

## Related issue

Bug fix to the existing tracker deletion guard. Heartbeat cleanup belongs to the separate active work in #2941 and is outside this diff.

## Validation

`node test-all.mjs`: 8,688 passed, 0 failed, 17 existing warnings. Web validation passed with 503 tests, type checking, and a production build. Independent review found no blockers.

Six regression tests cover cancellation, exit before stdio closes, PDF follow-up work, process errors, repeated cleanup, independent runs, and a harmless real Node worker. An additional isolated check executes the actual route with mocked AI workers and PDF rendering to verify cancellation, stream failure, and delayed rendering.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

- [x] Changes align with the existing evaluation and PDF workflows.

### DIFF
--- a/web/src/app/api/run/route.ts
+++ b/web/src/app/api/run/route.ts
@@ -15,6 +15,7 @@ import { createCvEnvelopeFilter, type CvEnvelope } from "@/lib/cv-envelope.mjs";
 import { buildPrompt, isShellSafeCompanyName } from "@/lib/run-prompts.mjs";
 import { claudeCliArgs } from "@/lib/claude-invocation.mjs";
 import { acquireTrackerWrite, releaseTrackerWrite } from "@/lib/core/run-registry";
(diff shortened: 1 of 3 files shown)


santifer:

PR #4115, opened 2026-09-12
fix(generate-pdf): keep both a report's CV and cover letter in the PDF manifest (#3967)

### LINKED ISSUE #3967: generate-cover-letter.mjs --report clobbers the CV's pdf-index.tsv row for the same report number
## Summary

`updatePDFManifest()` in `generate-pdf.mjs` (used by both `generate-pdf.mjs` and, via the shared `renderHtmlToPdf` export, `generate-cover-letter.mjs`) dedupes `data/pdf-index.tsv` rows by report number alone:

```js
// generate-pdf.mjs, updatePDFManifest()
lines = readFileSync(manifestPath, 'utf-8').split('\n').filter((line) => {
  if (!line.trim() || line.startsWith('#')) return false;
  const fields = line.split('\t');
  if (fields[1] === relPDF) return false;
  if (reportNum && normKey(fields[0]) === normKey(reportNum)) return false;
  return true;
});
```

Both the CV and the cover letter for the same application share the same report number (e.g. `052`). Since the filter only keys on `reportNum`, generating the cover letter with `--report 052` after the CV was already indexed for `052` silently drops the CV's row instead of letting both rows coexist.

## Repro

```bash
node generate-pdf.mjs "output/My Resume.html" "output/My Resume.pdf" --format=letter --report=052
# pdf-index.tsv now has one row: 052  output/My Resume.pdf  output/My Resume.html  letter  <date>

node generate-cover-letter.mjs --payload /tmp/cover-payload.json --out "output/My Cover Letter.pdf" --format letter --report 052
# pdf-index.tsv now has ONLY: 052  output/My Cover Letter.pdf    letter  <date>
# The CV row for 052 is gone — the dashboard's CV<->report link for this report is lost.
```

## Impact

- `data/pdf-index.tsv` is the source the dashboard uses to open/regenerate the exact HTML/PDF pair for a report (per `AGENTS.md`'s `pdf-index.tsv` description). Losing the CV row breaks that lookup for the report silently — no warning is printed, and `verify-pipeline.mjs` doesn't currently catch it either.
- Any workflow that generates a CV and then a cover letter for the same report (the normal `pdf.md` -> `cover.md` flow, or `auto-pipeline`) will hit this every time `--report` is passed to both.

## Suggested fix

The dedup key needs to distinguish artifact type, not just report number — e.g. key on `(reportNum, isCoverLetter)` by checking whether `relPDF`/`payload.letter` marks the row as a cover letter (or simpler: only drop a prior row for the same report number when the new row's path looks like the same kind of artifact — same directory/basename pattern as the outgoing PDF, e.g. both CV-shaped or both cover-letter-shaped). A minimal fix: track whether the call came from `generate-cover-letter.mjs` (e.g. an explicit `kind: 'cv' | 'cover'` param threaded through `renderHtmlToPdf` -> `updatePDFManifest`) and only replace rows of the same `kind` for that report number.

## Workaround (manual, current)

After generating both artifacts for the same report, manually re-add the dropped row to `data/pdf-index.tsv` (it's a plain gitignored TSV, so this is safe but easy to miss).

## Environment

Found while running the `career-ops` `auto-pipeline` -> `pdf` -> `cover` flow for a single application (CV then cover letter, same report number, `--report` passed to both generators).

gunjanjaswal:
Fixes #3967.

`updatePDFManifest()` in `generate-pdf.mjs` deduped `data/pdf-index.tsv` rows by report number alone:

```js
if (reportNum && normKey(fields[0]) === normKey(reportNum)) return false;
```

A CV and its cover letter share the same report number, so generating the cover letter for report `052` after the CV was indexed silently dropped the CV's row. The dashboard uses that manifest to open and regenerate the exact HTML/PDF pair for a report, so the CV↔report link was lost with no warning, and `verify-pipeline.mjs` didn't catch it either.

### Fix
Thread an artifact `kind` (`'cv' | 'cover'`) from `renderHtmlToPdf` into `updatePDFManifest`, record it as a trailing column, and only drop a prior same-report row when it is the same kind. A report now keeps both its CV and its cover letter, and regenerating either still supersedes its own stale row. `generate-cover-letter.mjs` passes `kind: 'cover'`; the CV path defaults to `'cv'`.

The trailing column is backward compatible: the `pdf-index.tsv` readers (`find.mjs`, `outcome.mjs`, `sync-pdf-flags.mjs`, `merge-tracker.mjs`) key on the leading columns and ignore extras, and a legacy row with no kind reads as `'cv'`.

### Before / after
Before: `generate-pdf … --report 052` then `generate-cover-letter … --report 052` leaves one row (cover only); the CV row for 052 is gone.

After, both coexist:
```
052  output/052-cv.pdf     output/052-cv.html  letter  <date>  cv
052  output/052-cover.pdf                      letter  <date>  cover
```

### Tests
`tests/generate-pdf-manifest-kind.test.mjs` covers CV + cover coexistence, same-kind regeneration replacing its own row, and a legacy (no-kind) row surviving a cover write. It fails on `main`, passes here. `npm run lint` is clean, and the manifest-reader suites (`outcome`, `pdf-index-path`, `sync-pdf-flags`, `merge-tracker`, `generate-pdf-batch`) stay green.

## Summary

Preserves CV and cover-letter entries for the same report in `data/pdf-index.tsv`.

## User impact

Users can generate both artifacts without losing the CV dashboard lookup. Regeneration replaces only the matching artifact kind. Legacy rows remain CV entries. Cover-only reports do not resolve to a CV.

## Changes

- `generate-pdf.mjs:1164-1208`: Adds the trailing `kind` column and deduplicates by report and artifact kind.
- `generate-pdf.mjs:1650-1800`: Threads `kind` through PDF rendering. The default is `cv`.
- `generate-cover-letter.mjs:341-347`: Marks cover letters with `kind: "cover"`.
- `find.mjs`, `web/src/lib/apply/cv-selection.mjs`: Select CV rows and exclude cover-letter rows.
- `tests/generate-pdf-manifest-kind.test.mjs:38-77`, `test-all.mjs`, `web/tests/lib/cv-selection.test.mjs`: Cover coexistence, same-kind replacement, legacy rows, row ordering, and cover-only reports.

## System files

### DIFF
--- a/generate-cover-letter.mjs
+++ b/generate-cover-letter.mjs
@@ -344,6 +344,7 @@ Usage:
       format: args.format || "a4",
       reportNum: args.report,
       inputPath: payloadPath,
+      kind: "cover",
     });
     console.log(`\nCover letter PDF: ${payload.output_path}`);
   } catch (err) {
--- a/generate-pdf.mjs
+++ b/generate-pdf.mjs
@@ -1161,21 +1161,27 @@ export function injectPrintPageCss(html, format = 'a4') {
  * Record a generated PDF in data/pdf-index.tsv so tools can map a tracker
  * report number to the exact PDF (and its source HTML for regeneration).
  *
- * Columns: report \t pdf \t html \t format \t date — paths relative to the
- * tracker workspace with forward slashes. One row per PDF path; when a report
- * number is given, older rows for that report are dropped too (regenerated
- * CVs supersede stale entries). The file is gitignored: it references
- * gitignored output/ artifacts and is meaningless on another machine.
+ * Columns: report \t pdf \t html \t format \t date \t kind — paths relative to
+ * the tracker workspace with forward slashes; kind is 'cv' or 'cover'. One row
+ * per PDF path; when a report number is given, an older row for that report is
+ * dropped only when it is the same kind, so a report can hold both its CV and
+ * its cover letter at once (#3967). A regenerated artifact still supersedes its
+ * own stale row. Readers key on the leading columns and ignore kind, so the
+ * trailing column is backward compatible; a legacy row with no kind reads as
+ * 'cv'. The file is gitignored: it references gitignored output/ artifacts and
+ * is meaningless on another machine.
  */
-function updatePDFManifest(reportNum, pdfPath, htmlPath, format) {
+export function updatePDFManifest(reportNum, pdfPath, htmlPath, format, kind = 'cv') {
   const manifestPath = resolvePdfIndexPath(trackerPath);
   const toRel = (p) => relative(workspaceRoot, p).split(sep).join('/');
   const relPDF = toRel(pdfPath);
   const relHTML = workspaceRelativeManifestPath(htmlPath, workspaceRoot);
   const date = new Date().toISOString().slice(0, 10);
+  const rowKind = kind === 'cover' ? 'cover' : 'cv';
   // "008" and "8" are the same report — zero-padded report-link form vs
   // unpadded tracker-# form. Normalize so replacement rows match.
   const normKey = (s) => (s || '').trim().replace(/^0+(?=\d)/, '');
+  const kindOf = (field) => ((field || '').trim() === 'cover' ? 'cover' : 'cv');
 
   let lines = [];
   if (existsSync(manifestPath)) {
@@ -1183,17 +1189,23 @@ function updatePDFManifest(reportNum, pdfPath, htmlPath, format) {
       if (!line.trim() || line.startsWith('#')) return false;
       const fields = line.split('\t');
       if (fields[1] === relPDF) return false;
-      if (reportNum && normKey(fields[0]) === normKey(reportNum)) return false;
+      if (
+        reportNum &&
+        normKey(fields[0]) === normKey(reportNum) &&
+        kindOf(fields[5]) === rowKind
+      ) {
+        return false;
+      }
       return true;
     });
   }
 
-  lines.push([reportNum || '', relPDF, relHTML, format, date].join('\t'));
+  lines.push([reportNum || '', relPDF, relHTML, format, date, rowKind].join('\t'));
 
   mkdirSync(dirname(manifestPath), { recursive: true });
   writeFileSync(
     manifestPath,
-    '# report\tpdf\thtml\tformat\tdate — written by generate-pdf.mjs, do not edit\n' +
+    '# report\tpdf\thtml\tformat\tdate\tkind — written by generate-pdf.mjs, do not edit\n' +
       lines.join('\n') + '\n'
   );
   return relPDF;
@@ -1626,6 +1638,7 @@ export async function inlineLocalFonts(html) {
  *   format?: 'a4'|'letter',
  *   baseDir?: string,
  *   reportNum?: string,
+ *   kind?: 'cv'|'cover',
  *   inputPath?: string,
  *   workspaceRoot?: string,
  *   maxPages?: number,
@@ -1668,6 +1681,7 @@ export async function renderHtmlToPdf(html, outputPath, opts = {}) {
  *   format?: 'a4'|'letter',
  *   baseDir?: string,
  *   reportNum?: string,
+ *   kind?: 'cv'|'cover',
  *   inputPath?: string,
  *   maxPages?: number,
  *   strictPages?: boolean,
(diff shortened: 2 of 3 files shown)


Scott-Emberson (commented):
I reviewed the owned test `tests/generate-pdf-manifest-kind.test.mjs`. It is sound and mutation-sensitive, and it gates the #3967 fix directly.

The test drives the real code: it imports `updatePDFManifest` from `generate-pdf.mjs` and exercises the actual manifest-writing path, redirecting the manifest to a temp dir via `CAREER_OPS_PDF_INDEX` (which `resolvePdfIndexPath` honors first), so it runs hermetically. There is no Playwright on this path, `updatePDFManifest` is a pure fs function, so there is no missing-browser skip or vacuous pass to worry about.

It asserts both kinds survive with literal expecteds. Writing a cv row then a cover row for report 052 asserts `rows.length === 2` with exactly one `kind==='cv'` and one `kind==='cover'`, which is the direct #3967 regression of one clobbering the other. A second case confirms regenerating the same kind replaces only its own stale row (cover to v2 supersedes, cv untouched), and a third confirms a legacy 5-column row with no kind reads as cv and survives a cover write. The dedup key it pins is (normalized report number, kind), matching the fix at generate-pdf.mjs:1192-1198.

I mutation-checked it: reverting the dedup predicate to key on the report number alone (dropping the `kindOf(fields[5]) === rowKind` clause), which is the original bug, turns the suite red (0 passed, 1 failed, all three subtests fail, exit 1). Green again on revert. Not a false-pass.

Owned test sound, not a false-pass. The substance is in `generate-pdf.mjs`, so the routing and merge decision is the maintainer's, not mine to gate.

santifer:
Thanks for this, @gunjanjaswal, and sorry it sat without an answer. You're right that dedup by report number alone was dropping the CV, and the kind column with same-kind replacement is the right shape; the legacy five-column test is well chosen.

One thing needs changing before it can land, and it's on the reading side rather than the writing side. The manifest now holds two rows per report and no reader looks at `kind`. `parsePdfIndex` in `find.mjs` sets `report -> pdf` in a loop, so the last row wins, and `updatePDFManifest` appends: once a cover letter is generated, that map hands back the cover letter as the report's PDF. `outcome.mjs` and `merge-tracker.mjs` both consume it.

`pdfIndexEntryForReport` in `web/src/lib/apply/cv-selection.mjs` takes the first match instead, so the two disagree, and which one is wrong depends only on which artifact you generated first. I ran it: with the cover row ahead of the CV row it returns `output/052-cover.pdf` for report 52. That file's own comment says a wrong CV attached silently is worse than no CV found, and this is the apply path.

So: could the readers select by kind, treating a missing column as `cv`? Related, the `updatePDFManifest` doc comment says readers key on the leading columns and can ignore kind. The format is compatible, but the row count per report isn't, and that sentence is what hides this.

Once that's in I'll pick it up. Plainly though: the branch is behind, so the merge itself is my move and needs a call I don't own yet, same as #3191.

gunjanjaswal:
Good catch, and thanks for tracing it all the way to the apply path — you're right that the row-count-per-report change broke the readers even though the column format stayed compatible, and the "readers can ignore kind" line in my doc comment is exactly what papered over it.

Fixed in the latest push. Both readers now select by kind and skip cover rows, treating a missing kind column as cv so legacy single-row manifests read exactly as before:

- `parsePdfIndex` in find.mjs no longer lets a trailing cover row overwrite the CV (feeds outcome.mjs and merge-tracker.mjs).
- `pdfIndexEntryForReport` in cv-selection.mjs picks the CV instead of the first match, so it agrees with find.mjs regardless of which artifact was generated first. I reproduced your report-52 case first — it returned the cover; now it returns the CV.

A cover-only report resolves to "no CV" rather than the cover, since attaching the wrong CV silently is worse than none. Rewrote the misleading doc comment, and added tests: the cover-precedes-CV ordering and the cover-only case in cv-selection.test.mjs, plus a parsePdfIndex case in the find.mjs suite.

The branch is still behind main — I've left the update and the #3191/#4314 sequencing to you as you said.

santifer:
Thanks @gunjanjaswal: this and #3959 solve the same problem, and we are landing #3959. That is not a judgement on your work, only that two fixes cannot both go in.

Your effort is not lost: it is linked from #3959, and if part of your approach ends up in the final change you will be credited as co-author.

If you think #3959 misses something yours covers, say so here or on that thread: we read both before merging. And if you want a next thing to work on, `good first issue` and `help wanted` are curated for that.

gunjanjaswal:
Totally fair, @santifer — one of them had to win, and #3959's kind-parameter reader API is a cleaner shape than mine. No concerns about the call. I left one note on #3959 about a reader it doesn't cover yet; other than that, happy to see this solved either way.

PR #4116, opened 2026-09-12
docs(tests): the repo root is where a suite does not run, not where it is registered

rubicon:
## What does this PR do?

`tests/README.md` told contributors that standalone `*.test.mjs` files in the repo root are run by `test-all.mjs`'s inline script list. That was accurate when it was written in #1714, but #3388 moved all nine root suites into `tests/` and deleted the list, and `tests/no-root-suites.test.mjs` now fails the build if a `*.test.mjs` file shows up at the root at all. A contributor following the README lands a red required check on a file the README told them where to put.

This points the paragraph at the guard instead. `ARCHITECTURE.md:30` already states it correctly, so this brings the third doc into line rather than introducing a new claim.

## Related issue

Closes #4097

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Documentation / translation
- [ ] Refactor (no behavior change)

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/career-ops-hq/career-ops/blob/main/CONTRIBUTING.md)
- [x] If this is a new feature or architecture change, I opened an issue first (bug fixes, providers, docs & translations are exempt — send those straight in)
- [x] My PR does not include personal data (CV, email, real names, scan results, or pipeline data)
- [x] I ran `node test-all.mjs` and all tests pass
- [x] My changes respect the [Data Contract](https://github.com/career-ops-hq/career-ops/blob/main/DATA_CONTRACT.md) (no modifications to user-layer files)
- [x] My changes align with the [project roadmap](https://github.com/career-ops-hq/career-ops/discussions/156)

## Summary

Updated `tests/README.md:24-26` to match current test discovery behavior.

## User impact

Contributors must place `*.test.mjs` suites in `tests/`. Root-level suites are not discovered or executed. `tests/no-root-suites.test.mjs` rejects root-level suites.

## Files touched

`tests/README.md`: documentation only.

No changes affect `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, or `.github/`.

PR #4117, opened 2026-09-12
docs(tests): correct stale root-suite sentence in tests/README.md (#4097)

### LINKED ISSUE #4097: docs(tests): tests/README.md says root *.test.mjs files are run; a required check fails the build if one exists
### Code of Conduct

### Existing issues

### What happened?

`tests/README.md:24-26` tells a contributor that a root-level suite is registered rather than discovered, and therefore runs:

> Other `*.test.mjs` files at this level (e.g. `stats.test.mjs`) cover root scripts. Note: standalone `*.test.mjs` files in the repo root are run by `test-all.mjs`'s inline script list, not by this directory's discovery.

That was accurate when it was written in #1714 (2026-07-28), back when nine root suites were named one by one in `test-all.mjs`'s `scripts` list. #3388 moved all nine into `tests/` and deleted the list, and `tests/no-root-suites.test.mjs` now fails the build if a `*.test.mjs` file shows up at the root at all. The sentence outlived both changes.

Measured on `main` at `da8c6f91`:

- No root suite exists. `git ls-files '*.test.mjs' | grep -c '^[^/]*\.test\.mjs$'` returns `0`.
- Nothing in `test-all.mjs`'s `scripts` list ends in `.test.mjs`. The seven root entries there use the `*-tests.mjs` name instead (`set-status-tests.mjs`, `agent-inbox-tests.mjs`, and so on), which is a different convention with its own guard in `tests/root-tests-registration.test.mjs` (#3735).
- `tests/no-root-suites.test.mjs` fails on any root `*.test.mjs`. Its header explains the cost it exists to prevent: `jd-similarity.test.mjs` was added with 20 assertions, appeared in no runner, and passed the entire time it was not running (#3303).

`ARCHITECTURE.md:30` states the opposite and is the correct one:

> Write new ones there; `*.test.mjs` at the repo root is not discovered and will not run.

So the two docs a contributor is most likely to read before adding a test disagree, and the one named `tests/README.md` is the wrong one.

### Steps to reproduce

Follow `tests/README.md:24-26` and put a suite at the root:

```bash
printf 'import { pass } from "./tests/helpers.mjs";\npass("demo");\n' > demo-repro.test.mjs
node test-all.mjs --only no-root-suites
```

```
test-all.mjs — no suite outside discovery
  ✅ ROOT is the repo root — test-all.mjs is a file there, so an empty result means empty
  ❌ 1 suite(s) at the repo root, where discovery does not reach and nothing runs them:
    demo-repro.test.mjs
  Move the file into tests/ — discovery picks it up with no registration.

📊 Results: 1 passed, 1 failed, 0 warnings
🔴 TESTS FAILED — do NOT push/merge until fixed
```

`test-all.mjs` is a required check, so a contributor who followed the README lands a red PR on a file the README told them where to put.

### Expected behavior

The paragraph says what the repo root actually is for suites: not discovered, not a place a `*.test.mjs` file may live, with `tests/no-root-suites.test.mjs` as the guard that enforces it. Correcting it costs one sentence, and it turns a harmful pointer into the one that sends a curious contributor to the guard.

Related but separate: #4091 and #4095 cover the *naming* sentence in the other two docs that describe this layout, `ARCHITECTURE.md` and `CONTRIBUTING.md`. That is a different claim, and there is no file overlap, since #4095 does not touch `tests/README.md`. Worth noting only because it is the same trio of docs: #4095's body records that `tests/README.md` was already right about naming, and on discovery it is the only one of the three that is wrong.

### CLI tool

Claude Code

### OS and Node.js version

macOS 26.6.2 / Node 26.8.2

Rayan-and-beyond:
Fixes #4097.

## What does this PR do?

`tests/README.md` still tells contributors that standalone `*.test.mjs`
files in the repo root "are run by `test-all.mjs`'s inline script list".
That list was deleted in #3388, zero root suites exist, and
`tests/no-root-suites.test.mjs` fails the build if one shows up — so the
sentence directs contributors to do exactly what CI forbids. This corrects
it to match `ARCHITECTURE.md` ("Why the flat root"), which already states
the true rule.

## Type of change

## Checklist

- [x] New feature/architecture first: not applicable (docs fix; also fixes #4097)
- [x] No personal data (docs-only, one prose sentence)
- [x] I ran `node test-all.mjs` and all tests pass — 8686 passed, 0 failed
  (17 warnings, byte-identical count on pristine `main`, pre-existing)
- [x] Data Contract: untouched (no user-layer files)
- [x] Roadmap: untouched (docs fix, no behavior change)

## Verification beyond the checklist

- Guard green: `node test-all.mjs --only no-root-suites` — 2 passed.
- Guard proven both directions: with a decoy root `*.test.mjs` present it
  fails (`1 passed, 1 failed`); decoy removed, green again. Decoy deleted,
  tree holds only this edit.

## Summary

`tests/README.md:24-26` now states that root-level `*.test.mjs` files are not discovered and will not run. New test suites belong in `tests/`, where discovery finds them automatically. `tests/no-root-suites.test.mjs` fails the build if a root-level suite exists.

Only `tests/README.md` is changed. AGENTS.md, modes/, update-system.mjs, DATA_CONTRACT.md, providers/, and .github/ are not changed.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -22,8 +22,9 @@ Node.js (`tests/helpers.mjs`).
   Underscore-prefixed files (e.g. `_html-entities.test.mjs`) test shared
   helper modules.
 - Other `*.test.mjs` files at this level (e.g. `stats.test.mjs`) cover root
-  scripts. Note: standalone `*.test.mjs` files in the repo root are run by
-  `test-all.mjs`'s inline script list, not by this directory's discovery.
+  scripts. Note: standalone `*.test.mjs` files in the repo root are not
+  discovered and will not run — `tests/no-root-suites.test.mjs` fails the
+  build if one shows up (see "Why the flat root" in ARCHITECTURE.md).
 
 **Web tests do not live here.** `web/` runs its own `npm test` over
 `web/tests/**/*.test.mjs` (see [../web/README.md](../web/README.md)); this


Scott-Emberson (commented):
The correction is right. I checked the current tree: `test-all.mjs` discovers `tests/**/*.test.mjs` and its own comment says standalone root `*.test.mjs` files are never picked up, `tests/no-root-suites.test.mjs` exists and fails the build if a suite appears at the root, and ARCHITECTURE.md's "Why the flat root" says exactly what you point to ("`*.test.mjs` at the repo root is not discovered and will not run"). So the old sentence, which claimed root suites run via an inline script list, was stale, and this replaces it accurately. Using the full guard path and pointing at that ARCHITECTURE.md section is the right call, since it is the canonical statement of the rule.

One thing worth knowing: this overlaps #4116, which corrects the same sentence in the same spot with slightly different wording and issue references (#3303, #3388). Both are correct, so it is a pick-one between the two rather than a code question. No issue with the substance here.

PR #4120, opened 2026-09-12
chore(web): pin @types/node and typescript to exact versions

prasta1:
## Summary
- Pins `@types/node` and `typescript` in `web/package.json` (and lockfile) to exact versions instead of caret ranges, avoiding patch/minor drift across installs.

## Test plan
- No behavior change; devDependency-only version pin. `npm install` in `web/` resolves to the same versions already in use.

https://claude.ai/code/session_01NcUFuGKD8xcDvmnCkSPpJg

Pins `@types/node` to `25.9.5` and `typescript` to `7.0.2` in `web/package.json:33,38` and `web/package-lock.json:27,32`.

This prevents version drift across installs. The change affects development dependencies only. It does not change application behavior or exported APIs.

No system files were touched: `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/`.

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -24,12 +24,12 @@
       "devDependencies": {
         "@tailwindcss/postcss": "^4.2.2",
         "@types/js-yaml": "^4.0.9",
-        "@types/node": "^25.6.0",
+        "@types/node": "25.9.5",
         "@types/react": "^19.2.14",
         "@types/react-dom": "^19.2.3",
         "postcss": "^8.5.18",
         "tailwindcss": "^4.2.2",
-        "typescript": "^7.0.0"
+        "typescript": "7.0.2"
       },
       "engines": {
         "node": ">=22.0.0"
--- a/web/package.json
+++ b/web/package.json
@@ -30,12 +30,12 @@
   "devDependencies": {
     "@tailwindcss/postcss": "^4.2.2",
     "@types/js-yaml": "^4.0.9",
-    "@types/node": "^25.6.0",
+    "@types/node": "25.9.5",
     "@types/react": "^19.2.14",
     "@types/react-dom": "^19.2.3",
     "postcss": "^8.5.18",
     "tailwindcss": "^4.2.2",
-    "typescript": "^7.0.0"
+    "typescript": "7.0.2"
   },
   "overrides": {
     "postcss": "$postcss",


santifer:
Sorry for the six days of silence, @prasta1. One question before this goes either way, and it is a genuine one rather than a polite no.

Did you hit a concrete resolution problem, or is this precautionary? I ask because web/package-lock.json is committed and web-ci.yml installs with npm ci, so in CI those caret ranges already resolve to exactly the versions you pinned. The place where drift is real is the repository root, which tracks no lockfile and installs with npm install, and that is why playwright alone is pinned exactly there. So the convention is not "pin devDependencies", it is "pin where something concrete would otherwise move".

If you did hit real drift locally, say what moved and I will take the change. In that case I would rather do all eight web devDependencies than two, so the file carries one rule instead of a mixed one.

If it was precautionary, the honest answer is that the lockfile is already doing this job, and pinning typescript to a patch turns every future patch into a pull request.

Either answer is fine and this stays open until you have had the chance to give it.

prasta1:
Precautionary. I did not hit a resolution problem, and I attemped to convey that in the PR, "resolves to the same versions already in use." What actually prompted it was renovate's major bump of the web deps landing in my fork the day before; I saw versions move and pinned them reflexively. Nothing failed.

Your description of the convention is better than my patch, and I would rather have it written down than have the PR. "Pin where something concrete would otherwise move" explains the playwright pin at the root in a way that "pin devDependencies" never would.

One correction, offered because it changes the cost side of your tradeoff rather than to argue the PR back in: renovate.json extends group:allNonMajor and groups every npm update into a single npm dependencies PR on a Monday schedule. A pinned typescript patch joins that PR instead of opening one — it is a line in a grouped diff you already get weekly. fix(deps): update web npm dependencies (next 16.3.1, shaders 0.0.80) is two packages in one commit, so the grouping is already doing this.

If that makes it cheap enough to be worth reconsidering, the version I would actually defend is typescript alone, and not on drift grounds. web CI runs npx tsc --noEmit and next build, and ^7.0.0 admits minors, and TypeScript minors add type errors as a matter of routine rather than as a mistake. That is a property of the compiler, not of devDependencies in general — which is why I would argue against the all-eight version you offered. @types/js-yaml moving cannot break that build the way a compiler minor can, so pinning it buys nothing and the uniform rule ends up being the one with the weaker reason behind it. One file, one rule is a real virtue, but not when the rule has to be vague enough to cover both cases.

That said, it is still precautionary, and "the lockfile already does this job" is a fair answer to a precautionary PR. Close it if you prefer — no need to keep it open on my account! Thanks for looking into it!

PR #4122, opened 2026-09-12
fix(role-matcher): split one-sided level with two-sided vocabulary gap (#4058)

### LINKED ISSUE #4058: role-matcher: overlapping non-subset titles with a one-sided level false-match
> ⚠️ Sanitized reproduction below — no real tracker, CV, report, or application data is included.

## Code of Conduct

## Existing issues

- [x] I searched existing issues. This is a new boundary case in the existing `roleFuzzyMatch` false-positive family; the related closed issues below do not cover this token shape.

## What happened?

`roleFuzzyMatch()` returns `true` for two materially different same-employer roles when they share several reordered qualifiers, each title also has a unique token, and only one title states a numeric/roman level:

```js
roleFuzzyMatch(
  'Front Desk Assistant (Summer Housing)',
  'Administrative Assistant II (Housing Front Desk)',
); // true — expected false
```

When an incoming tracker-addition row has no usable posting URL and neither side carries a recognizable req/job/posting ID, `merge-tracker.mjs` reaches its company + fuzzy-role tier. The new opening is then treated as a duplicate of the existing row and enters the skip/update path instead of receiving its own tracker row. Depending on the scores, this can silently discard the new opening or write its date, score, PDF/report metadata, and re-evaluation note onto the wrong existing application.

This reproduces on upstream `main` at `7c6de77e46db1935b714e1d3978831051640ea20`.

## Steps to reproduce

```js
import { roleFuzzyMatch, roleTokens, extractLevels } from './role-matcher.mjs';

const a = 'Front Desk Assistant (Summer Housing)';
const b = 'Administrative Assistant II (Housing Front Desk)';

console.log(roleTokens(a));       // [front, desk, assistant, summer, housing]
console.log(roleTokens(b));       // [administrative, assistant, housing, front, desk]
console.log([...extractLevels(a)]); // []
console.log([...extractLevels(b)]); // [2]
console.log(roleFuzzyMatch(a, b));  // true
```

The titles share four of six union tokens (`front`, `desk`, `assistant`, `housing`), producing a Jaccard score above the 0.6 threshold. The existing specialization guard does not fire because neither token set is a strict subset of the other. The level guard does not fire because it intentionally compares levels only when both titles state one.

## Why the related fixes do not cover this

- #3650 / #3651 protects two titles that both state conflicting levels (`Specialist I` vs `Specialist II`). Here only one side states `II`.
- #1881 / #1886 protects a generic title whose token set is a strict subset of its specialized sibling. Here both titles have a unique non-baseline token (`summer` vs `administrative`).
- #2009 protects a lone sub-baseline seniority qualifier such as `Associate` or `Junior`. A numeric/roman level is handled separately and does not currently reject a one-sided level.

## Expected behavior

These two titles should not be considered sufficient evidence of one requisition. With no deterministic matching URL, report identity, or req/job/posting number, the conservative outcome is to add a separate row rather than mutate or suppress an existing application.

Any fix should preserve established behavior for:

- exact and punctuation-only reposts;
- a deliberately accepted loose rewrite such as a bare title versus the same title with a one-sided level;
- conflicting levels on both sides (#3650);
- strict base/specialization subsets (#1881); and
- deterministic URL or requisition-ID matches in `merge-tracker.mjs`.

Please add both a direct `roleFuzzyMatch` regression and a tracker-level test proving that the second opening becomes a new row rather than updating the first.

## CLI tool

Codex

## OS and Node.js version

Windows / Node v25.5.0

## Output of npm run doctor

Not included: the defect is reproduced by the pure `roleFuzzyMatch` call above and is independent of user configuration. No user-layer files are needed.

Rayan-and-beyond:
## What does this PR do?

`roleFuzzyMatch()` returns `true` for two materially different same-employer roles when they share reordered qualifiers, both titles have meaningful unique vocabulary, and only one side states a numeric/roman level:

`Front Desk Assistant (Summer Housing)` vs `Administrative Assistant II (Housing Front Desk)` — was `true`, now `false`.

The Jaccard score (4/6) clears 0.6, neither token set is a subset, and the level guard only compares levels when both sides state one — so merge-tracker reaches its company + fuzzy-role tier and folds the new opening into the wrong row.

A level on one side alone stays a loose rewrite when the titles tokenize identically (`Administrative Assistant` vs `Administrative Assistant II` still matches). It also stays a match when only one side contributes meaningful unique vocabulary. The split applies only when both sides contribute non-baseline unique vocabulary, so baseline-only differences do not turn a repost into a separate row.

## Related issue

Closes #4058

## Type of change

- [x] Bug fix

## Checklist

- [x] `node test-all.mjs --quick`: 8691 passed, 0 failed, 16 warnings
- [x] Matcher regression suite: 28 passed, 0 failed
- [x] Merge-tracker regression suite: 44 passed, 0 failed
- [x] Guard proven both ways: reverting only the two-sided meaningful-vocabulary requirement makes the new repost control fail (27 passed, 1 failed)

## Summary

`roleFuzzyMatch()` now separates materially different same-employer roles when one title has a numeric or Roman level and both titles contain meaningful unique vocabulary.

Valid reposts and identical-token titles still merge.

## Tests

Added matcher and merge-tracker regression coverage.

Reported validation passed: 28 matcher tests, 44 merge-tracker tests, and 8,691 quick-suite tests.

## System files

The repository contains `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/`. The supplied diff shows no changes to these paths.


Scott-Emberson (commented):
I reviewed the two owned tests here (`tests/role-matcher-level-tokens.test.mjs` and `tests/merge-tracker.test.mjs`). Both are sound and mutation-sensitive, and they cover the fix at the layer where the bug actually bit.

`tests/role-matcher-level-tokens.test.mjs` dynamically imports the real `role-matcher.mjs` and calls the exported `roleFuzzyMatch` in both argument orders. The new case asserts that a one-sided level with a unique word on each side splits ("Front Desk Assistant (Summer Housing)" against "Administrative Assistant II (Housing Front Desk)" must not match), and a paired control asserts identical-token roles still match ("Administrative Assistant" against "Administrative Assistant II"), so it guards against over-splitting too. Additive, 27 passed at this head.

`tests/merge-tracker.test.mjs` is touched for a real reason: `roleFuzzyMatch` feeds merge-tracker's fuzzy company+role dedup tier, and the #4058 damage was a silently collapsed tracker row, so pinning it at the integration layer catches the bug where it hurt. It spawns the real `merge-tracker.mjs` against a temp tracker via the `CAREER_OPS_TRACKER`/`CAREER_OPS_ADDITIONS` env override and asserts the #4058 pair stays two rows, plus a same-role repost control that stays one row with a triple check (`exitCode === 0 && killedBy === null && rows === 1`), where the `killedBy` guard stops a hung or killed merge from satisfying the assertion. Purely additive, no existing assertion relaxed, 151 passed at this head.

I mutation-checked both. Removing the new one-sided-level guard in `role-matcher.mjs` (restoring the #4058 fuzzy-merge) reddens the split assertion in the unit test and the "keeps as two rows" case in merge-tracker (collapses the pair to one row), while both over-split controls stay green because a true repost is unaffected. Green again on revert, so neither is a false-pass, and neither has a vacuous shape.

Owned tests sound, not a false-pass. The substance is in `role-matcher.mjs`, so the routing and merge decision is the maintainer's, not mine to gate.

santifer:
Welcome, and thanks for staying with this through four rounds of review — that's a lot to ask of a first contribution.

The final narrowing is the right call. Requiring meaningful vocabulary on *both* sides is what keeps a loose rewrite from turning into two tracker rows, and the control you added for the one-sided case is what makes that safe rather than just claimed.

I checked it the hard way instead of reading it: neutralizing the new guard reddens both the unit assertion and the merge-tracker row — the pair collapses back into one — while both over-split controls stay green. 28/28 and 44/44 against current main.

One thing I fixed on my side so it wouldn't cost you a round: `tests/merge-tracker.test.mjs` had lost its trailing newline. I've pushed that to your branch, so the head moved — nothing for you to do. Merging once CI is green on it.

santifer (approved):

PR #4132, opened 2026-09-12
fix(icims): stop doubling the careers- prefix on iCIMS board hosts

mcgalleg:
## What does this PR do?

`scan-ats-full.mjs` builds every iCIMS board as `careers-<entry>.icims.com`, but the public dataset stores many tenants as the full portal subdomain already (`careers-acme`, `uscareers-acme`, `acmecareers-west`). Those entries become hosts like `careers-careers-acme.icims.com`, answer 404, and get recorded as dead boards, so a large share of live iCIMS boards are never scanned. This PR builds the host from what the entry actually is, and lets `icims.fetch()` try the other host shape when the first one answers 404 on its first page.

**How big the gap is.** After a full sweep, iCIMS accounted for 7,469 of the 13,795 boards in `data/dead-boards.tsv`. Grouping the dataset's 10,108 iCIMS entries by shape, and checking a random sample of the dead ones at the corrected host:

| Entry shape | Entries | Recorded dead | Sampled | Live with postings at the corrected host |
|---|---|---|---|---|
| Starts with `careers-` | 3,255 | 100% | 80 | 32 (as-is) |
| Contains `careers` elsewhere (`acmecareers-west`) | 861 | 99% | 40 | 16 (as-is) |
| Other prefix (`jobs-`, `campus-`, `external-`…) | 505 | 100% | 40 | 12 (as-is) |
| Bare tenant | 5,486 | 52% | 80 | 13 (without the `careers-` prefix) |

Samples are small, but the direction is consistent: 73 of the 240 sampled dead boards (about 30%) are live at the corrected host.

**The change.**

- `icimsHostCandidates(entry)` (exported from `scan-ats-full.mjs`) returns the likelier host first and the other shape as a fallback. An entry that already starts with `careers-` is used as-is with no fallback. One that contains `careers` is tried as-is first. Any other entry keeps the canonical `careers-` host first, so `toEntry('acmefreight')` still yields `https://careers-acmefreight.icims.com/...` and the existing test is unchanged.
- `SOURCES.icims.toEntry` sets the primary as `careers_url` and the other as `fallback_urls`. The SSRF host check still applies to the primary, and `icims.fetch()` runs every fallback through the same `https` + `*.icims.com` gate.
- `icims.fetch()` moves to a fallback **only** on a 404 from the first search page. A throttle, timeout, DNS failure or later-page 404 is rethrown unchanged, so `dead-boards.mjs` still counts a miss only for a real 404. The page loop moved verbatim into `fetchPortal()` so a fallback host gets identical pagination and truncation handling.
- Entries without `fallback_urls`, which includes every `portals.yml` entry, behave exactly as before.

One side effect worth knowing: for `careers-*` entries the dead-board key (`careers_url`) changes, so their old `careers-careers-*` rows in `dead-boards.tsv` simply stop matching anything and are re-checked from zero.

**Tests.** Two new files, so nothing collides with other open PRs:

- `tests/scan-ats-full-icims-hosts.test.mjs`: host candidates for each entry shape, normalization (a leading dash in the dataset, uppercase), candidates never leaving `icims.com`, hostile input still rejected.
- `tests/providers/icims-host-fallback.test.mjs`: first-page 404 falls back; a live primary never requests the fallback; all hosts 404 throws the 404 after one request each; 429 and later-page 404 are rethrown without switching hosts; off-host fallback URLs are ignored; an entry without fallbacks is unchanged.

Full `node test-all.mjs`: 8,742 passed, 0 failed. Its warnings are all pre-existing and unrelated to iCIMS or the sweep; the dashboard build was skipped locally because there is no Go toolchain on this machine.

Docs: the iCIMS row in `docs/SUPPORTED_JOB_BOARDS.md` describes the host handling.

## Related issue

None. This is a bug fix, sent straight in per CONTRIBUTING.

## Type of change

- [x] Bug fix
- [ ] New feature

## Checklist

## Summary

iCIMS host construction now supports bare tenants and existing portal subdomains.

- `scan-ats-full.mjs`: Adds `icimsHostCandidates()` and creates `careers_url` with optional `fallback_urls`.
- `providers/icims.mjs`: Tries a fallback only after a first-page `404`. Other errors and later-page `404` responses keep their existing behavior.
- `docs/SUPPORTED_JOB_BOARDS.md`: Documents that only non-canonical entries use fallback hosts.
- Tests cover host normalization, HTTPS and `*.icims.com` validation, fallback requests, `429` handling, and error behavior.

From the user’s perspective: entries such as `careers-acme` no longer become `careers-careers-acme`. If the primary host returns a first-page `404`, scanning tries the alternate host.

System files touched: `providers/`, `docs/`, `tests/`, and `scan-ats-full.mjs`. `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, and `.github/` are unchanged.

Test results: The supplied context reports 9,111 full-suite tests passing. Targeted fallback tests also passed.


Scott-Emberson (commented):
I reviewed the two owned tests here (`tests/scan-ats-full-icims-hosts.test.mjs` and `tests/providers/icims-host-fallback.test.mjs`). Both are sound, drive real code, and pin the host safety that matters on a host-construction change.

`tests/scan-ats-full-icims-hosts.test.mjs` imports the real `scan-ats-full.mjs` (`SOURCES.icims.toEntry`, `icimsHostCandidates`), which is where the doubling actually lived, and asserts the fix with literal hostnames: a bare `acmefreight` builds `careers-acmefreight.icims.com` with a bare fallback, an already-prefixed `careers-acmefreight` is used as-is with no `fallback_urls` (the no `careers-careers-` case), `uscareers-acme` is tried as-is first, and a hostile slug like `evil/..%2f` resolves to null. It also sweeps every candidate over five inputs and asserts each ends with `.icims.com`, so the domain stays pinned.

`tests/providers/icims-host-fallback.test.mjs` imports the real `providers/icims.mjs` and drives `icims.fetch(entry, ctx)` with a mock fetchText (no network), covering the fallback path: a first-page 404 may try a fallback, 429 and later-page 404s rethrow, a live primary never touches the fallback, and crucially an `https://evil.example` fallback and an `http://...icims.com` fallback are both ignored because the fallback is re-gated through `resolveOrigin` (HTTPS plus `endsWith('.icims.com')`).

I mutation-checked the host safety on both paths. Reintroducing the doubling reddens the construction test, removing the `.icims.com` anchor in `resolveOrigin` reddens the fetch test ("followed an off-host fallback"), and widening the candidate suffix to `.evil.com` reddens the stays-on-icims.com sweep. So neither anchor is decorative, and neither test is vacuous.

One scope note, since `providers/icims.mjs` is in my area: the diff there carries changes beyond the host-prefix fix (an `enrichDate` location backfill and an `INTER_PAGE_DELAY_MS` bump from 150 to 250). They look benign, but they are unrelated to the stated fix and are not covered by the two owned tests, so it is worth either calling them out in the description or splitting them, so the next reader is not left guessing which change the tests guard.

Owned tests sound, host safety covered. I am speaking to the two tests and the iCIMS host logic here; the `docs/SUPPORTED_JOB_BOARDS.md` and `scan-ats-full.mjs` changes are outside what I own, so I have not passed judgment on those.

santifer:
Welcome to career-ops, @mcgalleg, and sorry this sat for five days — that wait is on me, not on you.

Good catch, and a well-scoped fix. Prefixing every dataset entry with `careers-` built hosts like `careers-careers-acme.icims.com`, and those boards were being recorded as dead rather than as misaddressed. Returning the likelier host first, with the other shape as a fallback tried only on a first-page 404, keeps a throttle or a timeout reading as "unknown" instead of "dead" — which is the part that matters for dead-board tracking.

I re-ran the full suite with your branch merged into current `main`: 9111 passed, 0 failed. One review note no longer applies — the `enrichDate` and page-delay changes it mentioned are already in `main`, so against today's `main` your diff is only the host fix.

I took the two remaining review points myself rather than send you another round: the docs line now says only non-canonical entries carry a fallback, and there's an extra case covering a throttle raised by the fallback host. Your authorship is kept.

Next step is mine: push that integration to your branch so CI runs on it, then approve and merge. Nothing needed from you. Thanks for a careful first contribution.

santifer (approved):
Verified on the published integration: full suite green on the merged tree, and CI is green on all three operating systems for this exact commit. The host-candidate logic keeps every fallback re-gated through `resolveOrigin`, so it cannot leave `*.icims.com`, and only a first-page 404 switches hosts.

Issue #3867, opened 2026-09-05
[BUG] Split-checkout data-root violations: set-status.mjs can't resolve the tracker, merge-tracker.mjs spawns system scripts from the data root, plugins.mjs config path + .env gate order


nessuarez:
### Code of Conduct

- [x] I agree to follow the [Code of Conduct](https://github.com/career-ops-hq/career-ops/blob/main/CODE_OF_CONDUCT.md)

### Existing issues

- [x] I searched existing issues and this hasn't been reported yet

Closest existing work, none of which covers the three findings below:

- #3559 — its bug 2 covers `plugins.mjs` tracker/pipeline **paths** (assigned); finding 3 here is the adjacent-but-distinct **config path + `.env` gate ordering** problem.
- #3500 (fixed) — same *inverse* pattern as finding 2: a system-layer file resolved through the data root.
- #3714 (fixed) — same *ignore* pattern as findings 1 and 3: a script that bypasses `getCareerOpsRoot()`.
- #524 — the external-data-directory feature that introduced the split-checkout layout.

All findings verified on v1.32.0 (`ba8c83a`), with a split checkout: code checkout at `~/career-ops`, data-only root at `~/career-ops-data` (has `cv.md`, `config/`, `data/`, `portals.yml`, `reports/`, …), `CAREER_OPS_DATA_DIR=~/career-ops-data` exported. No `CAREER_OPS_TRACKER`, no `.career-ops-data` marker. `path-resolver.mjs`'s documented precedence (env vars > marker > codebase root) is the contract these scripts violate — in **both directions**: two ignore the data root, one uses it for system files.

---

## 1. `set-status.mjs` resolves the tracker against the code repo, not the data root

**Files:** `set-status.mjs:103-104`, `set-status.mjs:263`

```js
const CAREER_OPS = dirname(fileURLToPath(import.meta.url));   // the code checkout
const STATES_FILE = join(CAREER_OPS, 'templates/states.yml');
...
const APPS_FILE = resolveTrackerPath(CAREER_OPS);
```

`resolveTrackerPath()` honors `CAREER_OPS_TRACKER`, but the root passed to it is the script's own directory, so under a split checkout:

```
❌ No tracker found at ~/career-ops/applications.md
```

[...]

PR #4138, opened 2026-09-13
fix(doctor): checkDependencies and checkTrackedBakFiles read the code root

Zoubeir23:
Fixes finding 6 of #3867.

`checkDependencies()` and `checkTrackedBakFiles()` both read `projectRoot`
(`getCareerOpsRoot()`), but `node_modules` and `.git` only ever live where
`doctor.mjs` itself sits. Under a split checkout (`CAREER_OPS_ROOT` /
`CAREER_OPS_DATA_DIR`, or the `.career-ops-data` marker) that data root is a
separate directory that never holds either.

jday013/maxmilian's proof on the issue: creating an empty `node_modules/`
inside the data root flips "Dependencies not installed" to "installed" even
though the real code checkout's own `node_modules` never moved. The inverse
holds for the `.bak` check — a data root with no `.git` silently skips a
warning that should have fired against the real code checkout.

## Fix

Introduces `codeRoot` alongside `projectRoot`:
- `--target <path>` keeps meaning "diagnose this whole other checkout" — both
  roots follow it, matching the existing coverage in
  `tests/doctor-tracked-bak-files.test.mjs`.
- Otherwise `codeRoot` falls back to `__dirname` (where `doctor.mjs` lives).

Two call sites needed it, not one:
- `main()`'s human-readable checks array.
- The separate `onboardingState()` used by `doctor.mjs --json` — the form
  `AGENTS.md` has every agent run on the first message of every session — had
  its **own, independently broken** `checkTrackedBakFiles(root)` call. Found
  this while writing the test; it's the same bug hiding in a second place.

## Tests

`tests/doctor-code-root-checks.test.mjs` (new): seeds a throwaway
`CAREER_OPS_ROOT` data root that is its own git repo with a tracked `.bak`
file and no `node_modules`, then asserts both `doctor.mjs` (plain) and
`doctor.mjs --json` report on the real code checkout instead. Mutation-checked
— reverting either call site independently reddens its assertion.

`node test-all.mjs`: 8729 passed, 0 failed.

## Summary

`doctor.mjs` now separates `codeRoot` from `projectRoot`.

- Dependency checks inspect `codeRoot/node_modules`.
- Tracked `.bak` checks inspect the code checkout.
- Plain-text and JSON output use the same corrected paths.
- Split-checkout users no longer receive false warnings from the data root.

`tests/doctor-code-root-checks.test.mjs:1-101` adds coverage for both output modes. Reported result: 8,729 passed and 0 failed.

Touched system files: `doctor.mjs` and `tests/doctor-code-root-checks.test.mjs`. No changes were reported for `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, or `.github/`.


Scott-Emberson (commented):
I reviewed the owned test `tests/doctor-code-root-checks.test.mjs`. It is sound and mutation-sensitive, and it drives the real code.

The test spawns the real `doctor.mjs` (`execFileSync` from the code root, with `CAREER_OPS_ROOT` pointed at a throwaway data root), so it exercises both driver paths: the plain-text `main()` checks and the separate `onboardingState()` path behind `--json`. No import of the functions, no reconstructed logic. It sets up the split scenario deliberately: the throwaway data root is its own git checkout carrying a tracked `stray.md.bak` and no `node_modules`, while the code root has `node_modules` and no such tracked file. Then it asserts, with literal expecteds, that the output does not say "Dependencies not installed" (so `checkDependencies` read the code root's node_modules), and that `stray.md.bak` appears in neither the plain output nor the parsed `--json` `warnings[]` (so `checkTrackedBakFiles` read the code root, not the data root). That discriminates code-root from data-root rather than just confirming the command ran.

I mutation-checked all three. Pointing `checkDependencies` at the project root reddens assertion 1, pointing `main()`'s `checkTrackedBakFiles` at the project root reddens assertion 2, and pointing `onboardingState()`'s call at the data root reddens assertion 3, each independently, green again on full revert. So both checks are covered in both callers and none of the three is a false-pass. The suite also has no `process.exit(` in it, so `test-all` runs it rather than refusing it, and it is hermetic (temp data root, isolated git config, cleaned in `finally`).

Owned test sound, not a false-pass. The substance is in `doctor.mjs`, so the routing and merge decision is the maintainer's, not mine to gate.

PR #4142, opened 2026-09-13
fix(update-system): keep an upstream negation ahead of newly appended rules

### LINKED ISSUE #4127: update-system: reconcileGitignore appends rules after an existing upstream negation, re-ignoring test-fixtures
### Code of Conduct

### Existing issues

### What happened?

`update-system.mjs apply` re-ignored the upgrade fixtures in my fork, and `test-all` then failed:

```
test-fixtures/upgrade/state-v1.16/data/applications.md became ignored — the upgrade fixtures cannot be committed
test-fixtures/upgrade/state-v1.18/data/follow-ups.md became ignored — the upgrade fixtures cannot be committed
test-fixtures/upgrade/state-v1.20/data/applications.md became ignored — the upgrade fixtures cannot be committed
```

Cause: my `.gitignore` already had `!test-fixtures/**` but not the newer unanchored `applications.md` / `follow-ups.md` rules.
`reconcileGitignore` skips the negation as already present and appends the two rules at the end of the file, after it.
Later lines win, so the fixtures become ignored.

I know the ordering caveat above `reconcileGitignore` says a user-authored negation can lose to an appended rule, and that this is the intended precedence.
This case is different: `!test-fixtures/**` is an upstream line, and upstream's own `.gitignore` puts it after `applications.md` precisely so it wins.
The reconciled file inverts upstream's intent and fails upstream's own suite.

### Steps to reproduce

1. Start from a `.gitignore` that contains `!test-fixtures/**` followed by other rules, and does not contain `applications.md` or `follow-ups.md` (any install from before those rules were added).
2. Run `node update-system.mjs apply`.
3. Run `git check-ignore -v test-fixtures/upgrade/state-v1.16/data/applications.md`: it now reports the appended `applications.md` rule.

### Expected behavior

After reconciliation, every negation keeps the precedence it has in upstream's `.gitignore`.

A fix that keeps the "never modify local lines" promise: after building the appended block, re-append any upstream negation line that (a) already exists locally and (b) comes after one of the newly appended patterns in the upstream file.
The duplicate negation is harmless, and nothing local is rewritten.

Workaround today: move `!test-fixtures/**` to the end of `.gitignore` by hand (the header comment on the appended block already says reordering is safe).

### CLI tool

Claude Code

### OS and Node.js version

macOS (Darwin 27.0.0), Node.js v24.14.0

### Output of npm run doctor

`node doctor.mjs --json`: `{"onboardingNeeded":false,"missing":[],"warnings":[]}`

L4XB:
Fixes #4127.

## Reproduced

Your case as a test, before any change — a local `.gitignore` that already has `!test-fixtures/**` but not the newer unanchored rules:

```
local:     node_modules/ , !test-fixtures/** , *.log
upstream:  node_modules/ , applications.md , follow-ups.md , !test-fixtures/**
```

```
❌ the negation ends up after applications.md
❌ and after follow-ups.md
```

The two rules land at the end of the file, after the user's copy of the negation, and later lines win.

## The change

Exactly the shape you proposed. After the appended block is built, an upstream negation is appended again when it (a) already exists locally and (b) sits after one of the newly appended patterns in upstream's own file:

```js
for (const [index, raw] of upstreamLines.entries()) {
  if (index <= firstAddedIndex) continue;
  const line = raw.trim();
  if (!line.startsWith('!') || !localLines.has(line)) continue;
  block.push(raw);
}
```

Two cases deliberately do nothing:

- A negation the local file **lacks** needs no restoration — the main loop already emitted it, in upstream's own order relative to the patterns around it. `localLines` is the pre-loop snapshot, so those are excluded rather than duplicated.
- A negation upstream places **before** the new rules is left alone. Precedence cuts both ways, and repeating it would hand it a win upstream never gave it. There is a test for that direction too.

The `never modify a local line` promise is intact: repeating a line is not rewriting one, the local text is still a verbatim prefix of the result, and the user's own copy of the negation stays at the index they put it. A duplicate negation is a no-op to git.

## Testing

`tests/gitignore-reconcile.test.mjs`:

- the reproduction above, asserting the negation now follows both appended rules, that `text.startsWith(local)` still holds, and that the user's copy has not moved;
- **idempotence** — reconciling the result again adds nothing and returns byte-identical text, so an update does not rewrite `.gitignore` forever (the property the existing self-consistency case protects);
- the before-the-new-rules direction, asserting the earlier negation is *not* repeated.

```
tests/gitignore-reconcile.test.mjs      38 passed, 0 failed
tests/user-layer-gitignored.test.mjs    35 passed, 0 failed
every other test importing update-system.mjs   green
```

Negative control: disarming the restoration loop reddens exactly the two new assertions and nothing else.

(`tests/doctor-tracked-bak-files.test.mjs` fails 3/3 here, on this branch and on a stashed clean tree alike — `doctor.mjs` crashes against a temp target in my environment. Unrelated to this change; mentioning it so the count is not a surprise.)

---

## Summary

Users can run `.gitignore` reconciliation without upstream negations losing precedence.

Before this change, newer rules such as `applications.md` and `follow-ups.md` could re-ignore files under `test-fixtures/`. `update-system.mjs:2014-2058` now repeats existing upstream negations after newly appended rules while preserving local lines.

Negations that precede new upstream rules are not repeated.

## Tests

`tests/gitignore-reconcile.test.mjs:134-171` covers precedence, unchanged local lines, idempotence, and rule ordering.

The relevant suite reports 43 passing tests. The full run reports 9,075 passing tests. A label workflow failure was caused by GitHub API errors, not these changes. Local `doctor.mjs` failures are unrelated.

## Files changed

Changed: `update-system.mjs`, `tests/gitignore-reconcile.test.mjs`.

Not changed: `AGENTS.md`, `modes/`, `DATA_CONTRACT.md`, `providers/`, `.github/`.


Scott-Emberson (commented):
I reviewed the owned test `tests/gitignore-reconcile.test.mjs`. It is sound and mutation-sensitive, and it drives the real reconcile.

The test imports the real `reconcileGitignore` from `update-system.mjs` and calls it directly, plus it carries a source-level guard that `update-system.mjs` reads the upstream blob untrimmed via `gitShowRaw`, matching the existing source-guard pattern in the file. The new block sets a local `['node_modules/', '!test-fixtures/**', '*.log']` against an upstream that places `!test-fixtures/**` after `applications.md`/`follow-ups.md`, then asserts with literal expecteds that `added` is exactly `applications.md,follow-ups.md`, that the negation's `lastIndexOf` lands after both new patterns (the whole point of the fix), and that a second pass is byte-identical. The mirror block proves a negation that upstream places before the new rules is not re-appended (`filter(=== '!keep/**').length === 1`), so the fix is not a blind always-duplicate.

I mutation-checked both halves. Neutralizing the restore append fails the two "negation ends up after" assertions, and removing the position guard fails the "earlier negation is not repeated" assertion (expected 1, got 2), each exiting 1, green again on revert (38 passed). No vacuous shape, no `process.exit(` so test-all runs it. User-layer safety is covered too: since `.gitignore` is a file users also write, the assertions pin `text.startsWith(local)` (the user's lines untouched byte-for-byte) and that the user's own negation stays at its original index, plus the file asserts CRLF and escaped-trailing-space byte-identity.

Owned test sound, not a false-pass. The substance is in `update-system.mjs`, which is a red-line for me, so the routing and merge decision there is the maintainer's; I am speaking only to the owned test.

santifer:
Thanks for this, @L4XB, and sorry for the wait. The diagnosis is exact: upstream puts `!test-fixtures/**` after `applications.md` on purpose, and appending new rules behind a negation the install already had inverted that.

One case was still open, the one CodeRabbit raised on 13 Sep, and I reproduced it before deciding anything. With upstream `*.log`, `!keep/**`, `keep/secret.md` and a local file holding only `!keep/**`, the negation was restored after both new rules instead of between them, and `git check-ignore` then stopped ignoring `keep/secret.md`. Several of these system rules exist to keep files with personal data out of commits, so that inversion leaves trackable exactly what upstream re-ignores.

I put the fix in rather than sending it back to you. The appended block is now built in upstream's own order, repeating a local negation at the point upstream lists it and only once something has been appended. Your two cases still hold, and I added a regression that asks git itself: 43 passed, 0 failed on that suite and 9075 passed, 0 failed overall, with the new rows going red without the fix.

The failing `label` check is GitHub's API, as you and @chipoto69 both said. It re-runs on the refreshed head.

The fix is yours; the extra case is on me, and so are the refresh and the merge.

santifer (approved):

PR #4145, opened 2026-09-13
fix(ci): check out base before running labeler

rajatvarna:
Fork PRs fail the Label PRs workflow with a bare HttpError (e.g. run 34748939531 on #4142 from L4XB). Without a checkout step, actions/labeler falls back to fetching .github/labeler.yml via the API, which 404s when the fork head lacks the file. Checking out the pull_request_target base puts the config on disk; no PR code is executed, only the config is read. Verified the workflow YAML still parses and both steps are present.

## Summary

`.github/workflows/labeler.yml:28-35` now checks out the base revision before `actions/labeler`. The workflow uses pinned `actions/checkout` v7.0.1 and disables credential persistence.

## User impact

Pull requests continue to receive automatic labels when intermittent Contents API failures occur. The job reads the base configuration only and does not execute pull request code.

## Files touched

- `.github/workflows/labeler.yml:28-35`

Not touched: `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, and `providers/`.


santifer:
Thanks for this, @rajatvarna, and sorry for the wait. The workflow really does fall over sometimes, and you're right about the fix: actions/labeler only calls the contents API when `.github/labeler.yml` isn't already in the workspace, so a checkout removes that call entirely.

One correction worth having in the file, because it changes what the comment should say. The API fallback always asks this repo at the base SHA, never the fork's head, so a fork missing the file was never the trigger. Looking at the run history, fork PRs pass the labeler the overwhelming majority of the time, and the failures — including the one you cited — hit base branches too. The call fails intermittently. Your change is still the right fix, because it removes the dependency instead of the symptom.

I adjusted two things while integrating: pinned `actions/checkout` to the v7.0.1 SHA already used in `test.yml`, and added `persist-credentials: false`. This job is `pull_request_target` with `pull-requests: write`, so every action in it is pinned and nothing downstream needs the token — that's the point CodeRabbit raised too. The comment now describes the real cause.

Nothing needed from you. The credit is yours and I'll take it from here.

santifer (approved):

PR #4146, opened 2026-09-13
chore(deps): playwright 1.62.1 -> 1.63.0 incl. Dockerfile pins

rajatvarna:
Supersedes #4135, whose package.json-only bump trips the Dockerfile pin-sync gate in test-all.mjs (test run 34732928892: same 3 failures on ubuntu, macos and windows). The gate requires the FROM tag, the --save-exact pin and the line-2 comment to match package.json, so this bumps all four together. Verified mcr.microsoft.com/playwright:v1.63.0-jammy is published and replicated the gate locally (all pins 1.63.0). No root lockfile is tracked, so none is included. Follow-up suggestion: teach dependabot or a workflow to bump the Dockerfile pins alongside package.json so every npm bump does not red the matrix again.

Updates Playwright from `1.62.1` to `1.63.0`.

- `package.json:101`: Updates the dependency pin.
- `Dockerfile:2,6,38`: Updates the browser image tag, version comment, and exact install pin.
- User impact: Playwright tests use the `1.63.0` package and `v1.63.0-jammy` image.
- No public API changes or known breaking changes are included.
- System files touched: `Dockerfile` and `package.json` only. `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, and `.github/` are unchanged.


santifer:
Five days of silence here was not a judgement on the change, @rajatvarna. Four of the workflows never ran: they are sitting at action_required waiting for a maintainer to approve them, and nobody pressed it. Sorry.

Your reading of why the dependabot bump could not pass is right, and I checked it rather than taking your word for it. The gate in test-all.mjs compares three things against the dependency: the FROM tag, the save-exact pin and the line-2 comment. Yours moves all three to 1.63.0 and main is consistently 1.62.1, so the sync holds either way. I also confirmed that playwright 1.63.0 is published and that the v1.63.0-jammy base image tag exists.

Next step is mine: releasing those held runs so the suite actually reports on your head. Once it is green I will take it in, and close the dependabot one against yours rather than the other way round.

Your follow-up suggestion about teaching dependabot to move the Dockerfile with the dependency is a good one and worth its own issue. Say the word and I will open it, or open it yourself if you would rather own it.

santifer (approved):

PR #4149, opened 2026-09-13
Scope startup checks to requested career work

woud420:
## Primary changes

Repository maintenance and explanation requests currently inherit a silent updater check and onboarding before any other work. Scope updater checks to update-related requests and run setup checks only when the requested product mode needs prerequisites that are not already known to be ready.

## Reviewer walkthrough

- Add a short task-scope rule distinguishing product work from maintenance and explanations.
- Example: correcting a documentation reference can finish without checking for updates or copying personalization templates. A requested evaluation still needs its candidate source data.
- Limit missing-prerequisite questions to the requested mode and continue independent work.

## Correctness and invariants

The Data Contract, privacy boundaries, updater apply confirmation, and candidate-source requirements remain in place. The doctor check's template-copy side effect is explicit. Unrelated missing files no longer block otherwise valid work.

## Testing and QA

- `node --test tests/agent-docs-script-refs.test.mjs`: **7 passed**.
- Executed the existing `test-all.mjs` routing and interview-prep contract blocks in a filesystem-only harness: **9 assertions passed**; full suite not run.
- `git diff --check`: passed; only `AGENTS.md` changed.
- No updater, doctor, onboarding, job search, or career pipeline was invoked.
- Contribution checklist: CONTRIBUTING.md read; no personal data or user-layer files changed. This is a workflow/documentation fix. The full runtime suite was not run for this instruction-only change.

## Risk / Rollout

This narrows setup and permission-sensitive startup behavior and remains a draft for review. A future runtime check cannot be assumed cached across sessions; the rule is limited to known readiness in the current session and rechecks when setup changes. Existing apply/submit approvals remain unchanged.

Resolves ENG-447


santifer:
Welcome, @woud420, and sorry for the silence: four days is too long for a first PR.

Your diagnosis holds, and I checked it. `AGENTS.md` does tell an agent to run `node doctor.mjs --json` on the first message of every session, and `doctor.mjs` copies personalization templates when it finds them missing. So opening a session in the repo to review a PR or read the docs can write files into someone's tree. That is a real problem and nobody had named it.

Where I am not settled is the shape of the answer. Relaxing the onboarding gate has a cost in the other direction: an evaluation that runs against an unpersonalized `_profile.md` scores an offer against a stranger's targeting, and the current wording exists to stop that. Getting both right changes the agent contract, so it is my call, and I would rather make it deliberately than quickly.

There is also a scope question between this and your #4150, which argues the same thing across 22 files. I would rather pick one vehicle than review both halves of the argument.

Nothing for you to do right now: this is marked draft and the next move is mine. I will come back on this thread with a direction.

woud420:
Apologies @santifer, totally agent driven PR that was meant for a local branch. I've closed it but feel free to ignore or refactor!

PR #4151, opened 2026-09-13
fix(web): persist the default CLI at any installed count, not only a sole one

jsmith7735:
## What does this PR do?

Persists the default CLI on machines with **two or more** AI CLIs installed. Today only a *sole* install is persisted, so every multi-CLI machine gets a Config page that looks configured while every AI feature silently does nothing.

## Related issue

None — bug fix, which CONTRIBUTING.md exempts from issue-first. Happy to open one if you'd prefer it tracked.

## Type of change

- [x] Bug fix
- [ ] New feature

---

## The bug

`config-form.tsx` currently persists the auto-selection only when exactly one CLI is installed. The multi-CLI path falls through into precisely the state the code's own comment warns about:

```js
// Highlight + persist the only installed CLI when Config was never saved.
// Highlight-only used to look configured while jobs still read empty localStorage.
const only = list.filter((c) => c.installed);
if (only.length !== 1) return list.find((c) => c.installed)?.id || "";   // ← highlights, persists nothing
if (!readSavedCliId()) persistCliId(only[0].id);
```

With two or more installed, the early return highlights the first CLI and writes nothing.

Every AI surface reads `career-ops:config` directly and sends it as `cliId`. So Config displays a selected CLI over an empty `localStorage`, `/api/explore/ai` answers `{"error":"query and cliId required"}`, and the assistant shows "no CLI configured" next to a Config page that looks configured.

Nothing prompts a Save, because nothing looks unset. The workaround a user lands on by accident: select a *different* CLI → Save → reload → select the original → Save → reload.

## Reproduction

`main` at 1.32.0, clean browser profile, four CLIs installed (`claude`, `codex`, `gemini`, `opencode`):

```
localStorage after visiting /config, no clicks: <EMPTY>
```

With this change:

```
localStorage after visiting /config, no clicks: {"mode":"cli","cliId":"claude"}
```

## The fix

Whatever Config renders as selected is what it must persist, so the pick becomes `pickDefaultInstalled()` — first installed, any count.

`pickSoleInstalled()` keeps its narrower meaning for `resolveCliId()`, which answers a different question (*"can I choose for the user unambiguously?"*) and is deliberately left alone.

## On the test move

Both helpers now live in `src/lib/cli-pick.mjs`, re-exported from `saved-cli.ts`.

That is what lets `tests/lib/saved-cli-pick.test.mjs` import the **real** functions. The suite previously carried a hand-copied mirror of `pickSoleInstalled`:

```js
// Mirror of pickSoleInstalled in src/lib/saved-cli.ts (TS; this suite is .mjs).
function pickSoleInstalled(clis) { ... }
```

A mirrored test cannot fail when the real function changes — plausibly why this gap went unnoticed. The extraction follows the `.mjs` pattern `web/README.md` prescribes for exactly this reason.

Mutation-checked: reverting `pickDefaultInstalled` to the sole-install behavior turns **2 tests red**.

## Tests

`web/` suite: **502 pass / 0 fail** (was 497). `npx tsc --noEmit` clean. `npm run build` clean.

## Checklist

- [x] Bug fix — exempt from issue-first
- [ ] I ran `node test-all.mjs` and all tests pass — **see note**
- [x] My changes respect the Data Contract (only `web/` system-layer files touched)

**On the test box:** `node test-all.mjs` reports 2 failures on this branch. Both are pre-existing and environmental — verified by stashing the changes and re-running on a pristine tree, which fails identically: `live archive: no PDF written to jds/` (fetches a live posting over the network), and `tests/mjs-files.test.mjs` flagging a gitignored file in my own checkout that won't exist in CI. Flagging rather than ticking a box that isn't strictly true.

## Summary

The Config page now saves the first installed AI CLI when no saved choice exists. AI features no longer behave as unconfigured on machines with multiple installed CLIs.

## Changes

- `web/src/components/config-form.tsx:15-82`: Use `pickDefaultInstalled` and persist the detected CLI unless a saved choice exists.
- `web/src/lib/cli-pick.mjs:6-20`: Add shared helpers for sole-CLI resolution and default selection.
- `web/src/lib/saved-cli.ts:1-39`: Reuse and re-export the shared helpers while preserving sole-install resolution.
- `web/tests/lib/saved-cli-pick.test.mjs:8-74`: Test the real helper implementations.
- `web/tests/lib/config-form-persist-wiring.test.mjs:21-55`: Verify persistence, saved-choice protection, and detection-effect wiring.

## Validation

- 507 web tests pass.
- TypeScript checks pass.
- The production build succeeds.

## System files


santifer:
Sorry for the three-day silence, @jsmith7735 — that wait was on us, not on you. And welcome: this is a strong first PR.

You diagnosed this exactly right. Config highlighted the first installed CLI but only wrote it to storage when there was precisely one, so anyone with, say, claude and codex installed saw a selected CLI sitting over an empty key — and every surface that reads that key quietly did nothing. Making the value we render the value we persist is the right fix, and pulling the helpers into a shared module so the suite imports the real implementation instead of a hand-copied mirror is a genuine improvement on top of it.

One thing before this can go in. I deleted the persistCliId call from config-form.tsx — the line this PR exists to add — and the whole web suite still passed. The new tests cover the helper, but not the wiring that actually carried the bug, so a future regression of this exact fix would slip through unnoticed. Could you add one assertion over config-form.tsx? tests/lib/first-run-copy.test.mjs and tests/lib/decision-card-cta.test.mjs show the pattern the suite already uses: read the component source and assert on it. No new tooling needed.

Once that lands, bringing the branch up to date with main and re-running CI is on me.

santifer:
That is exactly the assertion I was missing, @jsmith7735. I re-ran the check that started this: with the persistCliId call deleted, three of your tests now go red, so the wiring is covered and not just the picker. Pulling the helpers into cli-pick.mjs so the suite imports the real implementation instead of a hand-copied mirror was the right call on top of it.

One optional thing, and it is a single click if you want it: CodeRabbit's suggestion on the last test. I checked it and it holds up. If that guarded call ever moves out of the detection effect but stays above save(), the current assertions still pass. Scoping the match to the /api/clis slice closes that. Take it or leave it, it is not a condition.

Updating the branch against main and re-running CI is on me, and then this goes in. Thanks for a careful first PR.

santifer (approved):
Reviewed on this commit: Config now persists whatever it renders as selected, the write stays behind the saved-choice guard, and the wiring is covered by a test that reddens when the persist call is removed. Merging once the required checks finish here.

PR #4152, opened 2026-09-13
fix(web): let report tables use the screen on large displays

jsmith7735:
## What does this PR do?

Lets report tables use the available width on large displays. The report container is pinned to `max-w-3xl` at every viewport, so a dense Block B verdict table renders 720px wide on a 3440px screen.

## Related issue

None — bug fix, which CONTRIBUTING.md exempts from issue-first. Happy to open one if you'd prefer it tracked.

## Type of change

- [x] Bug fix
- [ ] New feature

---

## The problem

`report-view.tsx` sets `mx-auto max-w-3xl px-6 py-8` regardless of viewport. A Block B verdict table is three columns of long evidence text — JD requirement, evidence in `cv.md`, verdict — and at 768px each column gets roughly 250px. On a 34" display the table uses about a fifth of the screen while the rest sits empty, and the cells wrap into narrow ragged strips that are genuinely hard to read.

## The fix

The container steps up on large screens:

```
max-w-3xl  →  xl:max-w-5xl  →  2xl:max-w-[1600px]
```

That takes the same table from **720px to 1552px**.

Running text must **not** widen with it — a 1500px measure is worse to read, not better — so prose blocks are capped:

```css
.report-prose p,
.report-prose li,
.report-prose blockquote { max-width: 76ch; }
```

Tables get the density; paragraphs keep a comfortable measure.

## Why not a viewport-width breakout

The usual trick — let tables break out of the reading column with `width: 96vw` and a negative margin — does not work in this layout:

- The 240px sidebar means the content column is **not** centered on the viewport, so viewport-centered math lands off-center.
- `<main>` is `overflow-x: hidden`, so anything that overshoots is **clipped**, not scrollable.

Widening the container and capping the prose achieves the same result using only the space `main` actually has.

## Measurements

Playwright, `/pipeline/{id}` with a real report, no horizontal page scroll at any size:

| viewport | table | prose |
|---|---|---|
| 3440 | 1552 | 760 |
| 1920 | 1552 | 760 |
| 1440 | 934 | 760 |
| 1024 | 678 | 678 |
| 390 | 285 | 285 |

## Scope

Width only. The small-screen table overflow handling already in `globals.css` (`overflow-x: auto`, `max-width: 100%`, the `nth-child` min-widths) is untouched — I hit that bug independently on 1.26.0 and found it already fixed on current `main`.

## Tests

`web/` suite: **497 pass / 0 fail**. `npx tsc --noEmit` clean. `npm run build` clean.

## Checklist

- [x] Bug fix — exempt from issue-first
- [ ] I ran `node test-all.mjs` and all tests pass — **see note**
- [x] My changes respect the Data Contract (only `web/` system-layer files touched)

**On the test box:** `node test-all.mjs` reports 2 failures on this branch. Both are pre-existing and environmental — verified by stashing the changes and re-running on a pristine tree, which fails identically: `live archive: no PDF written to jds/` (fetches a live posting over the network), and `tests/mjs-files.test.mjs` flagging a gitignored file in my own checkout that won't exist in CI. Flagging rather than ticking a box that isn't strictly true.

## Summary

- `web/src/components/report-view.tsx:76`: Report containers expand on large displays. Users can view wider tables.
- `web/src/app/globals.css:160-166`: Paragraphs, list items, and blockquotes remain capped at `76ch`.
- Small-screen table overflow behavior remains unchanged. Page-level horizontal scrolling does not occur.
- No changes affect `AGENTS.md`, `modes/`, `update-system.mjs`, `DATA_CONTRACT.md`, `providers/`, or `.github/`.
- Validation passed: 497 web tests, TypeScript checks, and the production build. Two `node test-all.mjs` failures are pre-existing environmental issues.


santifer:
Thanks for this, @jsmith7735 — and welcome, this is a really good first PR. Sorry for the quiet few days.

The diagnosis is right and the fix is the one I'd have reached for. Widening the container and capping the measure separately is the correct split: tables get the room, running text doesn't stretch into a 1500px line. The note on why the viewport-breakout approach doesn't work in this layout was genuinely useful — it saved me the experiment.

One thing worth clearing up so you don't go chasing it: the automated review asked you to scope the 76ch cap so it doesn't reach the job description view. You don't need to. I measured it, and that panel's paragraphs are already narrower than your cap, so the rule never takes effect there. 76ch sits just above the current reading column, which is exactly why nothing below the large breakpoints shifts. Leave it as it is.

I've merged your branch with current main locally and run the web suite, the type check and a production build — all clean — and confirmed the widths behave as you described, including no horizontal scroll on a phone.

Nothing further is needed from you. What's left is on my side: your branch is a little behind main, so I'll bring it up to date and let CI run on it from there. I'll keep you posted here.

And thanks for flagging the two test failures instead of ticking the box — I checked, and you're right that both are environmental and pre-existing.

santifer (approved):
Reviewed against current main: web suite, type check and a production build all pass, and the width behaviour is as described. Approving the change itself; bringing the branch up to date is on me.

Issue #2991, opened 2026-08-17
config/local-paths.txt is gitignored, so #2793 cannot be used by any fork that runs test-all.mjs in CI


EliRobinson:
Follow-up to #2421 / #2793. The mechanism works exactly as designed for a plain local checkout; this is about the one population it cannot serve.

## What happens

`config/local-paths.txt` is gitignored (`.gitignore:64`, shipped by #2793). `validate-system-paths-coverage.mjs` reads it at runtime (`localUserPaths(ROOT)`, line 48), and `test-all.mjs` hard-fails the suite on a coverage gap:

```js
} else {
  fail(`SYSTEM_PATHS coverage gap — a new file is unregistered and update-system will not ship it:\n${...}`);
}
```

A fork that runs the repo's own suite in CI therefore cannot use the feature. CI checks out the repo *without* the ignored file, so a fork-only tracked file is an orphan there and every PR goes red, even though the same command is green on the contributor's machine.

Committing the declaration file so CI can see it does not work either: it becomes a tracked file that nothing covers, and `localUserPaths()` refuses to let it list itself.

The result is that a fork with CI is pushed back to editing `USER_PATHS` inside `update-system.mjs`, which is the permanent merge conflict #2421 set out to remove.

## Reproduction

Against `upstream/main` at `2743281`, in a clean worktree. No fork code involved.

**1. A fork-only tracked file, no declaration. This is what CI sees:**

```
$ printf 'export const x = 1;\n' > local-only.mjs && git add local-only.mjs
$ node validate-system-paths-coverage.mjs
Coverage gap — tracked files not in SYSTEM_PATHS or USER_PATHS:
  local-only.mjs
```

**2. Declare it the intended way. Green locally:**

```
$ printf 'local-only.mjs\n' > config/local-paths.txt
$ node validate-system-paths-coverage.mjs
OK: 1082 tracked files covered by SYSTEM_PATHS or USER_PATHS
```

**2b. But that file never reaches CI:**

```
$ git check-ignore -v config/local-paths.txt
.gitignore:64:config/local-paths.txt	config/local-paths.txt
```

So CI reproduces step 1 and the suite is red.

**3. Commit it so CI receives it:**

```
$ git add -f config/local-paths.txt
$ node validate-system-paths-coverage.mjs
Coverage gap — tracked files not in SYSTEM_PATHS or USER_PATHS:
  config/local-paths.txt
```

**4. Declare the declaration file:**

```
[...]

PR #4154, opened 2026-09-13
docs(local-paths): document the supported fork-CI route for the declaration file

### LINKED ISSUE #4128: Committing config/local-paths.txt is required by the coverage guard (#2991) but fails user-layer-untracked
### Code of Conduct

### Existing issues

### What happened?

A fork that commits `config/local-paths.txt` fails `test-all`, and a fork that does not commit it also fails `test-all` in CI.

- `validate-system-paths-coverage.mjs` excludes the file and says a fork that runs the suite in CI "has to commit it", or every declared path is an orphan there (#2991).
- `tests/user-layer-untracked.test.mjs` fails any User Layer file that is both gitignored and tracked, and `.gitignore` ignores `config/local-paths.txt`. So the committed file fails with `config/local-paths.txt is User Layer, git-ignored, and still tracked — remove it from the index`.
- `DATA_CONTRACT.md` (lines 62 and 79) and `config/local-paths.example.txt` describe the file as always gitignored.

### Steps to reproduce

1. Add a fork-only root script and declare it in `config/local-paths.txt`.
2. `git add -f config/local-paths.txt` and commit.
3. Run `node test-all.mjs --quick`: `user-layer-untracked` fails. Without step 2, the coverage guard fails in CI instead.

### Expected behavior

One supported way for a fork to keep its declaration visible to CI.

What works in my fork today: add `!config/local-paths.txt` at the end of `.gitignore` and commit the file. Both checks pass.
Either document that next to the #2991 comment and in `config/local-paths.example.txt`, or exempt `config/local-paths.txt` in `user-layer-untracked.test.mjs`.

### CLI tool

Claude Code

### OS and Node.js version

macOS (Darwin 27.0.0), Node.js v24.14.0

### Output of npm run doctor

`node doctor.mjs --json`: `{"onboardingNeeded":false,"missing":[],"warnings":[]}`

L4XB:
Fixes #4128.

## The bind

A fork that runs the suite in CI is caught between two guards:

- `validate-system-paths-coverage.mjs` needs `config/local-paths.txt` visible to that checkout, or every path declared in it is an orphan there (#2991).
- `tests/user-layer-untracked.test.mjs` refuses a User Layer file that is git-ignored **and** tracked — the signature of a late ignore rule hiding an already-committed personal file.

Committing it fails the second. Not committing it fails the first. `DATA_CONTRACT.md` (lines 62 and 79) and `config/local-paths.example.txt` both described the file as simply gitignored, so nothing pointed at a way out.

## The route, and why it needs no behaviour change

There is a supported state that satisfies both, and it is the one the reporter found: append `!config/local-paths.txt` to the fork's own `.gitignore` and commit the file. It is then **tracked and not ignored** — so the coverage guard sees it, and the untracked guard has nothing to flag, because the condition it tests is "ignored *and* tracked".

That is now written down at each place a reader lands: the shipped example carries the recipe (including that the negation must come after upstream's rule, since Git applies the last matching pattern), `DATA_CONTRACT.md` stops claiming the file is always gitignored, and the `#2991` assertion in the coverage validator names it.

**The alternative is deliberately not taken.** Exempting the file inside `user-layer-untracked.test.mjs` would accept exactly the ignored-and-tracked state that guard exists to catch, for a file that holds fork-local paths, and it would do so for every fork rather than the ones that opted in. The negation makes the fork's intent explicit in the fork's own config, where it belongs.

## Evidence

The route is pinned by a fixture, not by prose alone — with a control, so it cannot pass by testing nothing:

```
$ node tests/user-layer-untracked.test.mjs
  ✅ parsed 46 user-layer paths from DATA_CONTRACT.md
  ✅ no ignored User Layer file remains tracked
  ✅ parses a User Layer section at end of file
  ✅ root wildcard finds tracked personal files without flagging a negated scaffold
  ✅ a fork that un-ignores config/local-paths.txt may commit it (#4128)
  ✅ committing it while it stays ignored is still flagged (#4128 control)

$ node test-all.mjs
📊 Results: 8726 passed, 1 failed, 16 warnings
```

Test count goes up by the two new assertions (8724 → 8726). The single failure (`--status prerequisite/score handling wrong`) is pre-existing and unrelated: `main` at `bb641dc`, in a clean worktree with the same dependencies, reports `8724 passed, 1 failed`.

## User impact

Forks that run CI can now commit `config/local-paths.txt` safely.

To pass validation:

1. Add `!config/local-paths.txt` at the end of the fork’s `.gitignore`.
2. Commit `config/local-paths.txt`.

The coverage guard can then inspect declared paths. The user-layer validation also accepts the file because it is tracked and not ignored.

The regression fixture verifies both the accepted configuration and the failure when the file remains ignored.

## Files changed

- `DATA_CONTRACT.md:62,79`: Documents the tracked and un-ignored configuration.
- `config/local-paths.example.txt:1-16`: Documents the fork CI workflow.
- `validate-system-paths-coverage.mjs: self-test comment`: Explains the required `.gitignore` rule and commit.
- `tests/user-layer-untracked.test.mjs: new fixture`: Covers accepted and rejected configurations.

No validation logic changed.

The following requested system files were not changed: `AGENTS.md`, `modes/`, `update-system.mjs`, `providers/`, and `.github/`.


Scott-Emberson (commented):
I reviewed the owned test `tests/user-layer-untracked.test.mjs`. The new fixture block is sound and mutation-sensitive, and it drives real code.

The added block (the #4128 fork-CI regression fixture) builds isolated real git repos in temp dirs (`git init` + `git add`) and runs the same `parseUserLayerPaths` / `trackedIgnoredUserLayerFiles` helpers the always-on block uses against the real `DATA_CONTRACT.md` User Layer table, no mocks or reconstruction. It asserts the invariant with literal expecteds: the negated case pins `forkViolations.length === 0` (a fork may commit `config/local-paths.txt`), and the control case pins `stillTracked.length === 1 && stillTracked[0] === 'config/local-paths.txt'` (an ignored-and-tracked user-layer file is still flagged, which is the actual privacy-leak invariant). I mutation-checked both directions: dropping `--ignored` reddens the negated case, forcing an empty return reddens the #4128 control, green again on revert. No `process.exit(`, so test-all runs it, and the fixture is hermetic.

One thing worth knowing, and it is not introduced by this PR: the pre-existing always-on real-tree block in this same file is host-git-version dependent. On my machine (`git 2.55.0.windows.3`) it exits 1 flagging `data/offers/.gitkeep` and `jds/.gitkeep`, because `git ls-files --ignored` disagrees with `git check-ignore` on those directory-nested `.gitkeep` negations. Your CI is green on all three OSes and your new fixtures use single-file negation in isolated repos, so they are not affected. Flagging it only because it lives in the file you are touching: at some point the real-tree block would be more robust querying `git check-ignore` per candidate rather than leaning on `ls-files --ignored`, so it stops depending on the local git build.

Owned test sound, not a false-pass. The substance is `DATA_CONTRACT.md` and the coverage script, so the routing and merge decision is the maintainer's; I am speaking to the owned test.

L4XB:
Thanks for taking the owned test seriously enough to mutation-check it in both
directions — dropping `--ignored` and forcing an empty return is exactly the
pair that separates a fixture that drives code from one that decorates it.

On the pre-existing always-on block: that is a real finding and I would rather
it not sit only in this thread. Your reproduction is specific enough to act on —
`git 2.55.0.windows.3`, `git ls-files --ignored` disagreeing with
`git check-ignore` on directory-nested `.gitkeep` negations, flagging
`data/offers/.gitkeep` and `jds/.gitkeep`. It is invisible in CI precisely
because all three runners agree with each other, which is the shape of bug that
only ever bites a contributor on their own machine.

Moving that block to a per-candidate `git check-ignore` is a behaviour change to
a test this PR does not otherwise touch, so I have kept it out of here rather
than bundling it. Happy to open it as its own PR against the same file if you
want it — say the word and I will, with your git version named as the
reproduction.

